PR #5, opened 2015-03-03
init/main.c

segin:
Ignore this.

PR #6, opened 2015-03-03
Revert "init/main.c"

jbruchon:
Reverts jbruchon/elks#5


jbruchon:
Console code requires some of these to work correctly.

PR #7, opened 2015-03-03
Return/refactor console init code

segin:
Code removed in previous commit is intended to init the console before launching init process. Restored that code and refactored it into it's own subroutine.


jbruchon:
This seems unnecessary. I'm retaining the original code.

PR #20, opened 2016-04-29
Fixed support for vi editor (levee) to use termcap.

lithoxs:
BCC libc already provides support for termcap.

The termcap support under levee did not work for ELKS.
To make vi work, was originally compiled with fixed VT52 support.
The TERM environment points to an ansi term.

Anyway, the current kernel ansi/vt52 emulator does not provide
all functions needed by vi. It provides a small subset of ansi and vt52
at the same time.

With the fix of this pull request, will be possible to modify the console
driver to provide complete support to vi with the ansi emulator and reduce
code size by removing vt52 emulation, at the same time.

Let me know if there are more programs in elkscmd requiring VT52

### DIFF
--- a/elkscmd/levee/display.c
+++ b/elkscmd/levee/display.c
@@ -39,7 +39,9 @@ PROC
 mvcur(y,x)
 int y,x;
 {
+#if ANSI
     static char gt[30];
+#endif
    
     if (y == -1)
 	y = curpos.y;
@@ -77,8 +79,7 @@ int y,x;
 #endif
 
 #if TERMCAP
-    tgoto(gt,y,x);
-    strput(gt);
+    strput(tgoto(CM,x,y));
 #endif
 }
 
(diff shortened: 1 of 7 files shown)


mfld-fr:
Upgrading from VT52 to ANSI, and removing VT52 overhead would be a good thing, but don't you think a more rigorous check of the VT52 usage should be performed before removing its support, more than just asking to the community ?

lithoxs:
In this pull request the vt52 kernel support is kept in place.

Also, I ran greps for strings vt52, vt-52, VT52 and VT-52 in elkscmd directory and only levee and elvis gave matches. Elvis requires support for vt52 only in the atari computer.

For the future, my proposal is to provide support for vt52 OR ansi., not simply remove vt52 code.

mfld-fr:
OK, got it. Thanks !

PR #22, opened 2016-05-07
Fixed problem for building ktcp

lithoxs:
This problem was reported by Derek Johansen in the mailing list.
Building ktcp required macro **KERNEL**, despite
that ktcp is a userspace program. This causes
wreak havoc because userspace and kernel space
have different library support.
Here the need for that macro was reduced to a minimum
and fixed 2 kernel headers


lithoxs:
Ktcp now builds sucessfully. The changes in the headers will prevent further problems for programs using the select syscall.

PR #28, opened 2016-05-25
Fix for problem with ls command, reported in the mailing list

lithoxs:


jbruchon:
This seems to work until I tried running `ls -aR` at which point it goes haywire again with the same "wrong waitpt" message. `ls -aR` looks like it's following `..` and hitting an infinite loop. I'll commit this to fix the obvious problem though; I'm sure it's easy to fix `ls` but I'd like to know what is causing the "wrong waitpt" crash as that could indicate a problem with the kernel.

lithoxs:
The problem with ls -aR is from the kernel 0.1.3 era. Indeed, the program is following .. directories. In my system, eventually finishes and I don't get that panic. But if you hit ^C maybe the tty layer is handling signals wrong.

jbruchon:
I would not be surprised if that is the case. ^C seems to crash everything.

jbruchon:
By the way, I'm working on fixing up some of the `ls` issues.

mfld-fr:
Git bisect tells me that the problem occurs starting from commit
4286d710caee2e25a0446932104d99b6cec2fe39
file_utils: add usage to some programs, clean up futils.h, **ls fix**

Looks like the fix is right for the reported problem, but as Jody said, this could the tree that hides the forest...

PR #31, opened 2016-05-29
Make read and write for tty devices return the correct value.

lithoxs:
Improve comments in serial.c and fork.c.
Simplifications in irq.c and pipe.c.
Compiled with BCC, tested with QEMU.
Code size reduced by 16 bytes.


mfld-fr:
Is this related to the problem of CTRL-C effect reported on the mailing list ?

Issue #39, opened 2016-08-02
XT IDE hdd detection results in system crash


sparky4:
My XT has an XTIDE card in it and when elks trys to look for a HDD, it crashes.... ><
I am reporting this issue and i can also provide screen photos of the crash. ^^;

http://4ch.mooo.com/4/xtssd/ the card i have is this

mfld-fr:
Hello, thanks for the report, yes your screenshots would be appreciated to help.

jbruchon:
I'd also like to see some screenshots of this crash.

sparky4:
oh my i apologize for long response! ok i will photo take with latest version. damn it school!

sparky4:
ok photos soon going to compile and make xt boot
![img_20161007_150615471](https://cloud.githubusercontent.com/assets/8118902/19203998/0a8adada-8ca0-11e6-9ae8-0d79b038fbb8.jpg)
![img_20161007_150616747](https://cloud.githubusercontent.com/assets/8118902/19204001/0a8ce4ce-8ca0-11e6-936b-62c802dcc2d1.jpg)
![img_20161007_150616929](https://cloud.githubusercontent.com/assets/8118902/19204002/0a8f1992-8ca0-11e6-8275-b3c4f9fcd23b.jpg)
![img_20161007_150617111](https://cloud.githubusercontent.com/assets/8118902/19203999/0a8b9916-8ca0-11e6-9122-d7a4bf4db179.jpg)
![img_20161007_150617292](https://cloud.githubusercontent.com/assets/8118902/19204000/0a8c2944-8ca0-11e6-8643-584c9dd3f091.jpg)
![img_20161007_150645090](https://cloud.githubusercontent.com/assets/8118902/19203997/0a8a5ec0-8ca0-11e6-9c97-f64700683097.jpg)
![img_20161007_150647158](https://cloud.githubusercontent.com/assets/8118902/19204003/0aa0dc22-8ca0-11e6-880c-12b0547e6b5c.jpg)
![img_20161007_150648504](https://cloud.githubusercontent.com/assets/8118902/19204005/0aa3fd80-8ca0-11e6-9a91-abf2bd7a2fde.jpg)
![img_20161007_150654050](https://cloud.githubusercontent.com/assets/8118902/19204004/0aa2d676-8ca0-11e6-8fa4-4a8fce934253.jpg)
[...]

PR #44, opened 2016-12-09
Miscelaneous fixes to BIOS disk driver.

lithoxs:
Now honors the CONFIG_HW_USE_INT13_FOR_FLOPPY option.
Bypass above config option if computer lower than PC-AT.
Use startup detected geometry for floppies if autodetect
fails when opening block device.
Fix for wierd int 13 BIOS behavior.


mfld-fr:
Fixes #39

PR #52, opened 2017-02-08
New document describing the ELKS system for new users and developers.

georgp24:
I analysed the ELKS system and made my notes as an introduction document for ELKS. I think this can be very helpful for new users or developers.

Georg

https://github.com/jbruchon/elks/commit/94d29b0602bcb3903eed328bcc66418d4899b5f4

### DIFF
--- a/elks/Documentation/html/Introduction/Introduction_to_ELKS.html
+++ b/elks/Documentation/html/Introduction/Introduction_to_ELKS.html
@@ -1,0 +1,2931 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">
+<HTML>
+<HEAD>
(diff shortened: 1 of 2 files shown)


mfld-fr:
Should `hdrinfo` be pushed to `dev86` rather than to `elks` ?
IMHO, it looks like more a companion tool for `bcc`, `as`, `ar`, etc, than a tool directly related to `elks`.

georgp24:
You could do this. But if it is not mentioned anywhere in the documentation nobody will find and use it.

jbruchon:
`hdrinfo` duplicates the functionality in `size86` from the dev86 tools. Regarding the documentation, there needs to be collaboration with other developers. HTML format aside, it reads like the random contents of an engineer's notebook rather than coherent technical documentation to introduce unfamiliar people. I can't pull any of these changes as they currently stand.

PR #53, opened 2017-02-12
Add a script to elkscmd to generate a bootable 32MB hard disk image f…

georgp24:
…or ELKS. This script requires the byte0x01 file to patch the kernel to look for the root file system on the hard disk.


mfld-fr:
@georgp24 : I have one concern with that request : you are adding a new script `buildHD` while the existing build system is based on Makefile, thus duplicating the things. Could you please rather integrate that HD target as an option as the other image targets `full3`, etc ?

georgp24:
This script works just fine and is not duplicating things since it starts off with a full3 image made with the existing makefile. It even calls the makefile if it finds no full3 image. There is nothing duplicated since you have to partition the disk, format the specific partition and use a different bootblock and mbr sector.

I looked at the makefile and I cannot make a new target in there. Someone else would need to do that. 

I suggest to use this script until someone has modified the makefile. Then it may be removed. 

Frankly, if everything gets rejected what I submit you force me to work on a different project.

jbruchon:
# ಠ_ಠ

jbruchon:
I'll pull this script but I am doing so with the expectation that it will receive some needed improvements in the future. It also needs to start with `#!/bin/sh` and the caps in the file name should be dropped; `build_hd_image.sh` would be a name I'd choose for it. I don't have time to make these edits right now, but others might be interested. It might be a good idea to drop it beside my master build script and tweak any hard-coded path constructs accordingly.

PR #54, opened 2017-02-12
First version of ELKS driver for NE2K

mfld-fr:
This pull request contains the second step for the N2K driver : the previous tiny packet driver has been integrated into ELKS as a new character driver (MAJOR 9, /dev/eth). Previous code has been modified to accommodate the QEMU environment.

Test procedure:
- configure and build your kernel with the new CONFIG_ETH option (Drivers / Network)
- build the `elkscmd` part and the `full3` floppy image
- run QEMU with the provided script `qemu.sh` at the root of the project (*)
- in ELKS, log as root and execute the `/bin/eth` test program
- open the network dump `qemu-vlan0.pcap` with WireShark or TcpDump
- check the ARP request / reply is correct

(*) That `qemu.sh` script configure QEMU not only for the ISA NE2K device, but also to be as close as possible to a pure ISA machine for ELKS. This is why I included it in the distribution, even if it should be reworked a bit for production (select the right keyboard and remove the `-net dump` option).


mfld-fr:
Please hold on this request, there is a regression when building from a fresh clone.

mfld-fr:
Regression fixed.
Thanks to @georgp24 for the problem report.

PR #56, opened 2017-02-15
Add usage information to kill command

georgp24:


jbruchon:
When you converted `write()` to `printf()` you didn't remove the `write()` character count parameter. It is also not necessary to use K&R style function declarations since `bcc -ansi` will allow the use of ANSI C function declarations and we want to be using ANSI C wherever possible moving forward. Also, if `printf()` is not called in a program at all, the `printf()` function code is not linked into it; this is part of why `write()` is so heavily used in ELKS programs (to avoid `printf()` being pulled in) but if a single `printf()` has to be used for any reason then adding more isn't a problem.

Can you make some more edits with this new info and add the changes to this pull request?

georgp24:
Yes, I will change the code.

georgp24:
This is one title but three commits:
- modified kill command as requested
- Miniterm serial communications program ported to ELKS
- fix two regression problems in elks/fs which I need to be fixed for my next commit which enables unix domain sockets

mfld-fr:
I am afraid to get lost at this point : here we are dealing at the same time with a bug fix in `fs`, an improvement of the `kill` command, and an enhancement as the addition of the `miniterm` program.

Changes in `kill` and `miniterm` are at low risk, but not the ones in `fs`. Could we have a minimal explanation of the bug in `fs`, a description of the fix, and, last but not least, a test procedure / test report so that we could check it covers the impacted code ?

georgp24:
I am sorry when I hit "New Pull Request" on Github it adds all the commits I have done in the branch and sends them together. Otherwise I have to make a new branch for every commit I guess.

My echo server hangs without the change in namei.c and my echo client will not run without the change in inode.c. If you want to test that I can commit these programs. I would like to commit my fix to the unix domain sockets but as long as my fixes to the regression problems are not accepted that will not run. I think it makes good sense to finally enable unix domain sockets with ELKS.

mfld-fr:
Yes, sure, but could you explain how you fixed the regression in `fs` ? Did you roll the changes on the related files back to the 0.2.0 version ? Or did you analyse the regression and found a forward solution ? This is the point that needs clarification, I think.

I suggest, when delivering a bug fix, to include the corresponding unit test in the same commit or pull request, so that the other contributors can 1) reproduce the problem on their side to understand the bug, 2) assess the proposed solution and 3) verify the fix is effective.

georgp24:
First I rolled the files back to the 0.2.0 version. Then I slowly rolled parts of the code of the old version into the files of the new version till I found the lines in the code that caused the problem. Since the old code works I used that as the guideline to change the new code.

In inode.c I just had to change a "==" to "!=" as had been the case in the old code. If you analyse it you can see that a test for equal is apparently wrong. 

In namei.c the line:
`if (!iop || !(op = (*(int (**)())((char *)iop + offst))))` is completely different from what the old code intended to achieve and I replaced that line with the lines that worked before.

After these changes my applications worked again.

I will commit my unix domain socket code tomorrow so you can test it.

mfld-fr:
OK, thanks Georg for that clarification. It would be nice to identify the commit that caused the regression, so that we could loop with the author and get his opinion.

jbruchon:
In the future, make a new branch for each modification you want to make. Don't commit to your `master`; instead, pull the changes from my `master` once the pull request is accepted to keep it in sync and avoid the need to delete and remake your entire fork to continue issuing "clean" pull requests. If you want to maintain a branch for all of your local changes, create it and use e.g. `git checkout mystuff_branch` then `git pull some_branch_name` to merge your fix(es) into your own local "master" that can be remade as needed without touching `master`.

georgp24:
Marc-Francois,
the problems I observed were introduced with these commits:
namei.c was changed on 8th July 2016 with 939b8cf
inode.c was changed on 24th January 2017 with b007a52
Georg

mfld-fr:
Thanks again Georg for the commit references.

@lithoxs, you authored these commits: could we have your feedback to check we miss nothing and we would not fall in another problem that you could have previously identified ?

PR #57, opened 2017-02-18
Ud sockets

georgp24:
Enabling unix domain sockets required just a few changes. Took me quite a while to find these though. Added echo server and echo client applications for testing and as examples for using unix domain sockets.

### DIFF
--- a/elks/include/linuxmt/socket.h
+++ b/elks/include/linuxmt/socket.h
@@ -6,7 +6,7 @@
 
 struct sockaddr {
     unsigned short sa_family;
-    char sa_data[14];
+    char sa_data[110];
 };
 
 struct msghdr {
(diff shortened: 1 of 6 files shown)


mfld-fr:
Nice to see the AF_UNIX come back, and moreover, with a test program !

My two cents:
- move the `MAX_SOCK_ADDR` constant from socket.c to socket.h to avoid a literal size of 110 in `sockaddr` structure,
- move the client & server test programs out of `\inet` (because not related to internet stuff), to another folder with a name related to Unix domain socket (something like "/test/socket/unix" ?).

By the way, I think it would be a good thing to start grouping our unit tests in a common location (like "/test") so that we could chain them in a beginning of a continuous integration process.

georgp24:
The `MAX_SOCK_ADDR` constant had been in socket.c before, but I have no objections to move it to socket.h.

I am not sure if one should move `echo` to a different folder. I intend to extend these programs to be used with ethernet as well and the loopback driver I have planned. Currently I use a variant of these programs for testing ne2k.

PR #60, opened 2017-02-23
Miniterm: ~x to exit without reset, fix up bps display

transplier:
Exit without reset is handy if you want to leave the serial port set up for use by another program, especially if `stty` is making problems.

I also fixed the bps status display, higher baudrates were causing the signed int to overflow and display as negative.

### DIFF
--- a/elkscmd/misc_utils/miniterm.c
+++ b/elkscmd/misc_utils/miniterm.c
@@ -232,7 +232,7 @@ static speed_t convert_baudrate(unsigned int baudrate)
  * \param old Pointer to termios structure for storing previous settings (may be \c NULL)
  * \return File descriptor
  */
-static int serial_open(const char *device, int baudrate, bool rtscts, struct termios *old)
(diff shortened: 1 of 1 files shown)


mfld-fr:
Thanks for the fix, mister Giacomo !

transplier:
Sure. Feel free to suggest any changes.

Aside: This is a neat project. It's the only way I could find to make a block level backup of an ancient HDD I have kicking around. Basically I'm using `miniterm` to set up the serial port so I can then do `dd if=/dev/bda of=/dev/ttyS0`. Slow but steady :)

georgp24:
"especially if stty is making problems"

I also observed that stty is not working ok. Maybe you could look into that and fix it?

mfld-fr:
Too late for the fix on `stty` in this pull request :-)
Please help the project by creating an issue to track this one, with details on what you found.

jbruchon:
With a project that has suffered from inactivity as long as ELKS has, I tend to prefer to pull fixes aggressively so that more people can quickly check them out and test them.

PR #69, opened 2017-02-28
[ktcp] Multi-line ARP cache

mfld-fr:
**Jody: please wait for the end of the code review before merging this one.**

This commit replaces the single-line ARP cache by a multi-line one. This allows to simplify the code (no direct cache manipulation) and get a cleaner layering.

I also removed the unit tests from the default images, as discussed on the mailing list, plus some cosmetics.

I am asking for a code review, because a peer review is always a good thing, and in Github workflow, a pull request is a good place to discuss the proposed changes. And we do have time to discuss: this is an improvement, not a bug fix.

### DIFF
--- a/elkscmd/inet/nettools/.gitignore
+++ b/elkscmd/inet/nettools/.gitignore
@@ -1,0 +1,1 @@
+nslookup
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -39,12 +39,14 @@ typedef struct arp_cache_s arp_cache_t;
 static arp_cache_t _arp_cache [ARP_CACHE_MAX];
 
 
-static int arp_cache_init () {
+static int arp_cache_init (void)
+{
 	memset (_arp_cache, 0, ARP_CACHE_MAX * sizeof (arp_cache_t));
-	}
+}
 
 
-int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr) {
+int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr)
+{
 	int err = -1;
 
 	/* First pair is the more recent */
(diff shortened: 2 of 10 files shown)


jbruchon:
I don't think I've mentioned the code style I want to see before but now is as good a time as any.

I would like to see spacing in function calls, function-looking constructs such as `if` `while` `switch`, and between operators and operands. Instead of
```c
for(x=1;x==5;x++){
while(foo){
if(foo){
switch(foo){
function(a,b,c,d);
```
I'd like to see this
```c
for (x = 1; x == 5; x++) {
while (foo) {
if (foo) {
switch (foo) {
function (a, b, c, d);
```
But don't go too far like these examples
```c
for ( x = 1 ; x == 5 ; x++ )
  {
while ( foo )
  {
if ( foo )
  {
switch ( foo )
  {
function( a, b, c, d );
```

mfld-fr:
So you want to wake up the supreme troll of the C coding style :-D ?

jbruchon:
Well, if I make you guys do it my way then I don't have to work as hard to fix it later ;-)

georgp24:
My usual style is this:

```
for (x = 1 ; x == 5 ; x++) {
   while ( foo ) {
      if ( foo )  {
         switch ( foo )  {
            function( a, b, c, d ) {
               ...... ;
            }
         } // switch (foo)
       } // if (foo)
   } // while (foo)
} // for (x)
`

I guess I can get away with this.

Georg

jbruchon:
Actually, I'm going to revoke what I said before and adopt this document to be the ELKS coding style guidelines, with the exception being that lines longer than 80 characters are okay.

https://www.kernel.org/doc/html/latest/process/coding-style.html

Much easier than me typing it out. I'll also add it to the project somewhere.

EDIT: Added in last commit.

mfld-fr:
Better than duplicating the document from the "big brother", perhaps a link from the ELKS documentation would have been enough and easier to maintain ? Provisioning you accept HTML in place of TXT for the link to be possible... as that "big brother".

jbruchon:
There's no maintenance to it; it's easier to pull it in from Linux than to write it out myself and miss details. I don't intend to change the style document unless someone has an argument against any part of the contents. Anyone who works with the Linux kernel code will already be used to it and there needs to be some kind of standard. Simply referencing it doesn't work because it's a standard that needs to be distributed with the source code, especially since it is possible that contributors may not necessarily all have constant Internet access.

mfld-fr:
Okay, okay... back to the code review: as you and Georg (actually the two I expected) made it, maybe you could merge now ?

georgp24:
Well, I did not have a chance to test the code yet, but I guess Marc-F. can fix it should anything show up.

mfld-fr:
Goal of this code review was to keep you informed of what I am doing, as you asked in issue mfld-fr#11. But I would feel more comfortable if we were two to test, and test again, for sure !

PR #71, opened 2017-03-01
[ktcp] Add MAC filtering for promiscuous mode

mfld-fr:
To work with QEMU that implements badly the MAC unicast filtering, the Ethernet NE2K driver forces the promiscuous mode to bypass that filtering. This results in receiving all the packets from the network, that causes failure on a real one. So this commit adds a filter to retain only the packets of interest.

Tested with HTTP from the host, and with TELNET from the guest (yes, ELKS telnet client works !).

### DIFF
--- a/elks/Documentation/html/user/qemu.html
+++ b/elks/Documentation/html/user/qemu.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+
+<html>
+
+<head>
+<meta charset="utf-8">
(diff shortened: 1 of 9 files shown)


georgp24:
Maybe you could add instructions how to test telnet to the readme-ktcp-ne2k.txt file.

mfld-fr:
See also issue jbruchon#72.

PR #74, opened 2017-03-02
delete leftover, garbage and unused files

lithoxs:
Directory arch/i86/boot/tools was a very old copy of present arch/i86/tools.
Most files in arch/i86/lib are the runtime library of BCC. Files l*b.s implement 32 bit
arithmetic and logical operations for an unused BCC option (big endian representation of 32 bit
numbers). The other deleted files in this directory are userspace C runtime functions.
The last file deleted is kernel/stubs.c. Had 0 bytes size and has been so for many years.
All deleted files were unused.


jbruchon:
Can you make sure that any of the removed files are also removed from the master .gitignore as well? I haven't checked.

PR #75, opened 2017-03-02
Initialize register ES from the begining.

lithoxs:
Do not assume a 80186 processor or higher in  function printsp().
Fix minor bug in strace.c.
Other minor fixes.

### DIFF
--- a/elks/arch/i86/kernel/process.c
+++ b/elks/arch/i86/kernel/process.c
@@ -65,8 +65,8 @@ void kfork_proc(char *addr)
 
     t = find_empty_process();
 
-    t->t_regs.cs = kernel_cs;
-    t->t_regs.ds = t->t_regs.ss = kernel_ds; /* Run in kernel space */
+    t->t_regs.cs = kernel_cs;			/* Run in kernel space */
+    t->t_regs.ds = t->t_regs.es = t->t_regs.ss = kernel_ds;
     arch_build_stack(t, addr);
 }
 
@@ -92,11 +92,10 @@ unsigned get_ustack(register struct task_struct *t,int off)
  */
 void arch_setup_kernel_stack(register struct task_struct *t)
 {
-    put_ustack(t, -2, USER_FLAGS);			/* Flags */
-    put_ustack(t, -4, (int) current->t_regs.cs);	/* user CS */
-    put_ustack(t, -6, 0);				/* addr 0 */
+    put_ustack(t, -2, USER_FLAGS);		/* Flags */
+    put_ustack(t, -4, (int) t->t_regs.cs);	/* user CS */
+    put_ustack(t, -6, 0);			/* addr 0 */
     t->t_regs.sp -= 6;
-    t->t_kstackm = KSTACK_MAGIC;
 }
 
 /* Called by do_signal()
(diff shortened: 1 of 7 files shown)


lithoxs:
Userspace requires segment register ES to be set to the same value than segment register DS.
Syscalls fork() exec() and others in malloc.c and process.c did not initialized that register. Its value
was set at the end of the first syscall after fork() or exec().

elks/arch/i86/kernel/process.c
```diff
@@ -92,11 +92,10 @@ unsigned get_ustack(register struct task_struct *t,int off)
...
 void arch_setup_kernel_stack(register struct task_struct *t)
 {
-    put_ustack(t, -2, USER_FLAGS);			/* Flags */
-    put_ustack(t, -4, (int) current->t_regs.cs);	/* user CS */
-    put_ustack(t, -6, 0);				/* addr 0 */
+    put_ustack(t, -2, USER_FLAGS);		/* Flags */
+    put_ustack(t, -4, (int) t->t_regs.cs);	/* user CS */
+    put_ustack(t, -6, 0);			/* addr 0 */
     t->t_regs.sp -= 6;
-    t->t_kstackm = KSTACK_MAGIC;
```
mfld-fr, line 99: Why removing the magic number that helps to detect stack overflow ?

mfld-fr (commented):
Again Jody merged the changes without waiting for the review to complete, leaving some questions opened (and maybe regressions), and giving no chance for the contributors to learn more through an open discussion. :-1:

georgp24:
Although Juan does an amazing job working through the existing code, I am always afraid he may cause a regression problem this way. I detected three regression problems which took me a very long time to locate. This is compared to the many changes done by Juan only a few percent but then someone has to take the time to run through all the commits done in 18 months to find these regressions in the end.

They used to say: "never change a running program" and I prefer it that way. Weeding other peoples code is dangerous. Otherwise we would need a thorough or automated testing procedure.

Georg

elks/arch/i86/kernel/process.c
lithoxs: 1. Because that assignment **resets** the marker to the state of no error. If an stack overflow has already happened, the problem will go undetected because of that assignment.
2. Because this in not the place to initialize that marker. The right place is at the creation of a task, just as is currently done.
3.To remove ugliness. Setting up the USER stack has nothing to do with initialization of a task structure.

lithoxs:
Again a commit without waiting for the review to complete.
All changes were to fix bugs that "I detected and which took me a very long time to locate", including the change objected by mfld-fr. And none are regressions which can be blamed to me.

If you accept to "never change a running program" I see two courses of action:
1. Assume that ELKS is running properly and reject all pull requests, or
2. Recognize that ELKS is bloated, buggy and ugly. Much less now than 5 years ago, but enough to drop ELKS from the status of "running".

mfld-fr:
To make the things clear : I don't ask for review to complete because I don't trust Juan. I do trust Juan, because he demonstrated he knowns far better the ELKS code than me. So reviewing his changes and understanding them is a real good opportunity to learn from him.

I just ask him some time to help us for that understanding, to allow us to to contribute in a more effective manner. And if we spot regression, it would be a nice to have, but not the primary goal of the review. This is why I don't understand either why his pull request was reverted back.

We are more these times than the usual two guys that maintain ELKS, surely for different reasons and goals, probably not for a long time, but nevertheless, maybe we should spend a bit more time to work together to maximize our global efficiency ?

elks/arch/i86/kernel/process.c
mfld-fr: OK, got it, thanks Juan. The marker is already set in `find_empty_process()`. Maybe that task structure initialization should be moved out of this function ?

jbruchon:
**I'm not doing "code reviews."**

Bureaucracy is a complete waste of time. I'm managing code. I'm not managing people. I'm going to accept pulls often and if there's a problem it can always be reverted or fixed down the line.

If you want to spin wheels forever with people going...

> "We need to review all changes before acceptance!"
> "We didn't review long enough, there's a problem!"
> "We didn't review long[er] enough before correcting the problem because it's not actually a problem!"

...then fork ELKS and spin those wheels in your own sandbox.

What do you want me to do with this? I can revert the revert or I can leave things as-is. Come up with a final answer and do it fast, otherwise I'm going to put Juan's code back in place.

mfld-fr:
That's your project, do as you like, but don't forget you won't go alone as far as within a team. That's my final answer.

jbruchon:
That's not an answer to the question posed. Is the change in this pull request staying or going?

mfld-fr:
These changes are valuable, so I would take them.

jbruchon:
Revert reverted in 9c141e6015c4e7cddb31cf1c25f67e97278fd658

PR #80, opened 2017-03-04
Update existing and add new documentation.

georgp24:
Move the readme file from the ktcp directory to Documentation/text giving it the new name networking_guide.txt. Extend the file networking.txt. Add new descriptions how to setup a slip connection and how to write kernel mode device drivers. Update index.html with the new files.

### DIFF
--- a/elks/Documentation/html/technical/kernel_mode_device_drivers.html
+++ b/elks/Documentation/html/technical/kernel_mode_device_drivers.html
@@ -1,0 +1,270 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">
+<HTML>
+<HEAD>
+	<META HTTP-EQUIV="CONTENT-TYPE" CONTENT="text/html; charset=windows-1252">
+	<TITLE></TITLE>
+	<META NAME="GENERATOR" CONTENT="OpenOffice 4.1.2  (Win32)">
+	<META NAME="AUTHOR" CONTENT="Georg Potthast">
+	<META NAME="CREATED" CONTENT="20170105;19172240">
+	<META NAME="CHANGEDBY" CONTENT="Georg Potthast">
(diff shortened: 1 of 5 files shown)


mfld-fr:
Georg, you should first pull back the master branch from Jody's repository with `git pull upstream master`, and solve the merge conflict.

georgp24:
Well, I tried to do my best. The conflicting file just needs to be erased since it has moved to documentation/text. I did erase it locally before pushing it to github.

I did pull Jody's master before making this branch. Maybe Jody or you can easily erase the file and solve the conflict that way.

mfld-fr:
@georgp24 : just to be sure, did you use the `git rm` and `git mv` commands, or the native `rm` and `mv` ? Please have a look to the network graph, it shows that the conflicting file has been modified in two different paths, so a latest pull back is needed in that case. Then push your commit to the same branch, it will be appended automatically to this pull request until it is closed.

georgp24:
I use git gui and this showed that I added an erased file.

Now I made another new branch but that gave the same result. Github detected that I moved the file under a different name into a new directory. When I tried to make the pull request it says:
"Can’t automatically merge. Don’t worry, you can still create the pull request."
So I did not worry on my first pull request.

Shall I make a branch without erasing the readme file in ktcp? Erasing does not work for me.

jbruchon:
It is generally a good idea to `git pull` before you do a `git commit` to see if something changed so you can `git stash` your changes, pull, then re-apply the stashed changes. https://git-scm.com/book/en/v1/Git-Tools-Stashing

In this case you just need to move the file from the old name you're using to the new name and it should fix it up.

mfld-fr:
@georgp24: if you already tried, don't spend any more your time to redo in another branch. That's documentation, not code, so it is far to be critical. Please go on with this pull request, we will have time later to make diff's and patching if needed.

Anyway, thanks for this valuable documentation, and moreover, at the right place ;-).

PR #85, opened 2017-03-05
Add ftpget as inet application

georgp24:


mfld-fr:
This would help to download test programs without changing the Makefiles and is also a good step toward the binary repository you mentioned :+1: .

PR #87, opened 2017-03-05
fix issue 73 - frequent pauses during data download.

georgp24:
Allows for fast ftp and web page downloads.

### DIFF
--- a/elkscmd/ktcp/slip.h
+++ b/elkscmd/ktcp/slip.h
@@ -1,6 +1,6 @@
 #ifndef __SLIP_H__
 #define __SLIP_H__
 
-#define SLIP_MTU	296
+#define SLIP_MTU   1064
 
 #endif
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -71,7 +71,7 @@ struct iptcp_s {
 #define	TS_LAST_ACK	9
 #define	TS_TIME_WAIT	10
 
-#define CB_IN_BUF_SIZE	512
+#define CB_IN_BUF_SIZE	2048 /* 1024 too small */
 #define CB_BUF_USED(x)	((x)->buf_len)
 #define CB_BUF_SPACE(x)	(CB_IN_BUF_SIZE - CB_BUF_USED((x)))
 #define CB_BUF_TAIL(x)	(((x)->buf_head + (x)->buf_len) % (CB_IN_BUF_SIZE - 1))
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -18,6 +18,7 @@
 #include "tcp.h"
 #include <linuxmt/arpa/inet.h>
 #include "timer.h"
+#include "tcpdev.h"
 
 char buf[128];
 
@@ -332,6 +333,8 @@ void tcp_output(struct tcpcb_s *cb)
     len = (__u16)CB_BUF_SPACE(cb);
     if (len == 0)
 	len = 1;		/* Never advertise zero window size */
+    if (len>255) 
+        len=256; //use even len, reducing the len causes the servers to pause
     th->window = htons(len);
     th->urgpnt = 0;
     th->flags = cb->flags;	


mfld-fr:
I made a few tests with that fix. It changes the timing of the TCP traffic, so reducing the chance of the problem to occur, but it does not solve it. In Wireshark, there is still something weird in the window sizes returned by `ktcp`, causing the router to sometimes resend some packets after the TCP ACK timeout. So I would not close the issue jbruchon#73 because of that.

Moreover, you are changing the MTU for SLIP to 1064, that is too much (see standard MTUs in https://wiki.wireshark.org/MTU). It could work in emulated environment, but would break SLIP on real lines. You stated right earlier that MTU should be selected based on the actual network interface, and in that commit, you force it whatever the interface. Why ?

georgp24:
ktcp calculates how much space is free in its buffer and then instructs the sending side about the maximum size it can accept. So this size varied constantly. I observered that the sender paused for about five seconds as soon as the advertised size was reduced. This slowed down the transfer extensively. 

So I increased the buffer, which was selected small when slip was used, and changed the code so the maximum size advertised is limited to 256 bytes. The result is that the advertised size is almost always the same size of 256 bytes and this increased the download time dramatically. Now a file download of 90k is done in a few seconds while before I could go for a coffee.

So the fix improves the result a lot and therefore I committed it. I guess one could implement a different solution but I doubt that it will result in a significantly better transfer time.

The MTU of 1064 is also a valid value for SLIP, I looked that up here:
[http://osr507doc.xinuos.com/en/NetAdminG/slipN.advanced_config.html](http://osr507doc.xinuos.com/en/NetAdminG/slipN.advanced_config.html)

mfld-fr:
Georg, please, let us be serious. We cannot improve the speed at the cost of a degraded reliability. The current code contains the following lines:
```
   len = (__u16)CB_BUF_SPACE(cb);
     if (len == 0)
 	len = 1;		/* Never advertise zero window size */
```
How can it work without loosing byte, as it still advertises a window size of 1, even if the buffer is full ?

About MTU: yes you can increase it, as for Ethernet where you can use 64 KB jumbo frames on recent chips. But as long you don't implement a path discovery to compute the right MTU, changing the default could lead to dramatically reduced performance on real and routed networks.

As stated in the page you mentioned:
"*The default MTU value is 296. Increasing this value may increase throughput as packets can contain more data. However, it may also decrease throughput if packets are routed through systems with lower MTU values, as each packet is split (fragmented) into smaller packets for transmittal. You can experiment with this value to obtain the most desirable results*".

georgp24:
Marc-F.,

the code you cited (converting 0 byte to 1 byte) is not done by me but by Harry about fifteen years ago. I did the lines below that, limiting the advertised size to 256 although the buffer does have space for more. This limitation, which I added, does not compromise reliability. It just uses part of the free buffer, not the full available buffer.

I also do not consider converting 0 byte to 1 byte a problem. It causes the sender to wait for about five seconds and this is sufficient time for ktcp to free the entire buffer.

We read different parts of the web page I mentioned. I read this part:
`NOTE: The suggested value for the MTU is 40 plus some power of 2. For example, the default value, 296, is 40 plus 2^8. Other possible values include 552, 1064, and 2088. ` So I implemented just not the default any more.

I will test if slip still works with this MTU. If not I will use a different MTU for ethernet and slip.

My commit maybe not the final solution to this issue but it works much better than that which was available before. You can test a file download with ftpget and you will see that.

mfld-fr:
I agree the lines I mentioned are not from your own, and my statement has to be refined. So, I am just saying that, BEFORE improving the performance, we should FIRST solve the problem we detected for reliability, and not implement shortcuts like this one without understanding the whole picture. Why "shortcut", despite it looks like to work well at a first glance ? Here is the explanation:

In this commit, you actually deceive the congestion algorithm on the other side. It is absolutely NORMAL that the remote host makes a pause when the window size is reduced so far, because that algorithm begins to adapt its throughput to the acceptance rate of the receiver. It starts with a huge delay, true, but will tune that delay for a smooth traffic later. See RFC's about the TCP adaptive traffic for more details, that are implemented in the far more complex TCP/IP stack on Linux side.

By forcing window size several times below the real capacity of the buffer, you give no chance to that algorithm to adapt. More, your patch is only working because you changed the timing with smaller packets, and you take advantage of the high speed of your emulation for ELKS to pop the small packets before the host sends next ones.

On the real target, we are not working @ several GHz, we are working with a **real 8086 / 80186 @ few MHz**. So more packets, more processing, more latency, and less throughput at the end. Your patch is ineffective, because the target is too slow to pop at the same speed than the host.

Even if you don't want to take into account the use case of ELKS on real devices, and restrict yourself to emulators, please **at least** wrap this patch with somewhat a "#define EMULATOR" to not have such side effect on other environments, including mine !

georgp24:
I only limit the size to 256 bytes, if ktcp has less free buffer size available this will be used instead. The only problem could result from converting 0 to 1 as Harry has done. However, the sender will pause then and this will free at least one byte in the buffer so it will not be lost.

What I observed is that the sender cannot adapt since ktcp specifies a different sized buffer each time. So this results in these five second pauses. My change avoids this.

If you are concerned how it works on real 8086 hardware we should test it on real 8086 hardware. I think it will work reliable with that hardware, just that these pauses become more frequent again.

mfld-fr:
I am testing on both emulated and real HW (this one: http://downloadt.advantech.com/download/downloadlit.aspx?LIT_ID=1-17BKQM), this latest directly connected onto my physical router, with about 10 Ethernet stations that generate noise, plus a real Linux TCP/IP stack as the host, not the QEMU simplified implementation through SLiRP. So, yes, I am concerned, and I observe a different behavior than the one you describe.

georgp24:
If you tell the sender to only send 256 bytes while you have space for 1.400 bytes that cannot cause a data loss. 

The only loss could be the change from zero to 1 byte implemented by Harry. As I said I consider this to be very unlikely. But you will find out. Maybe Harry did reserve some space for this byte?

mfld-fr:
*If you tell the sender to only send 256 bytes while you have space for 1.400 bytes that cannot cause a data loss* : TRUE, but you spend more CPU because you have many more little packets. It has a higher cost than processing only one big packet. So the most efficient window size is the one that reflects the actual free space in the receive buffer.

*The only loss could be the change from zero to 1 byte implemented by Harry. As I said I consider this to be very unlikely. But you will find out. Maybe Harry did reserve some space for this byte?* : this is the question we have to answer. If not reserved, there is a loss opportunity here, and I cannot accept it, because it will probably cause data corruption on low speed devices, as mine.

georgp24:
I guess we could replace 
`if (len == 0)
len = 1;`
with
`if (len < 65 && len != 0)
len = len/2;`
`if (len == 0)
len = 1;`
This way we reserve space for 5 slow transfers. Each will cause the sender to pause for 5 seconds since the size is gradually reduced. So we get a delay of 25 seconds total. This should be safe even for your hardware.

jbruchon:
Since bcc is stupid I'd rather see `len = len >> 1` to use an arithmetic right shift instead of a costly divide.

Can't we make the buffer larger? If this is going to operate on a real network of any sort there is going to be a lot of traffic anyway and anything that reduces the overhead of processing that traffic will be helpful. Retries and unnecessarily small packets both hurt performance.

Why is the max MTU when using the NE2K not 1500? Or maybe I've remembered some details wrong (I am on my mobile atm)

mfld-fr:
@georgp24: no, it looks like you did not understand well what I wrote. I don't need 25 seconds for my HW. I need a window size that maximizes the packet size, to save CPU. So `window size = place left in buffer`. Not less, not more.

The constant 5 seconds you observe on QEMU looks like a "silly" algorithm, not the real adaptive algorithm implemented in the Linux stack for real networks.

Therefore, the solution I am asking for, after this discussion, is to wrap these lines with a `#define` to allow the QEMU user to boost the performance, and to allow the others (including me) to save CPU on real HW.

mfld-fr:
@jbruchon: before increasing the size of the buffers, we should remove promiscuous mode that we enabled to bypass the tricky QEMU NE2K emulation, and left only the unicast and broadcast filtering. The asked `#define` would be useful for that.

About NE2K MTU: yes, the standard MTU for Ethernet is 1500, but the NE2K uses 6 blocks of 256 bytes in the ring buffer, so the possible packet size could be 6 * 256 - 4 (NE2K specific header) - 4 (appended CRC) = 1528. So I defined the `MAX_PACKET_ETH` constant to size the buffers accordingly and avoid overflow.

jbruchon:
> wrap these lines with a #define to allow the QEMU user to boost the performance, and to allow the others (including me) to save CPU on real HW.
This may be fine as a stop-gap during development but I do not think it is a sufficiently robust solution. The OS needs to be adaptable to different hardware. There needs to be a self-tuning element that changes the behavior of the algorithm on-the-fly when performance issues due to incorrect tuning crop up. This code could be run on a real-world 8086, a real-world 80386, or an emulated 80486 running at a virtual 5000 MHz because the emulator is so astonishingly fast compared to the 8086. One chunk of the same code needs to be effective on all platforms and "expand to fill its container."

The real-world 8086 is the lowest common denominator and the ultimate "golden standard" for the code, but the 8086 accommodations don't have to happen at the expense of taking advantage of better performance.

Regarding the network buffers, I can see now why you made that decision.

mfld-fr:
@jbruchon: agree the principle, but... how could you detect we are running on QEMU or not, in the userspace where `ktcp` runs ? Or do you suggest to add our own adaptive algorithm in the TCP layer to maximize performance ?

georgp24:
As I wrote `window size = place left in buffer` results in very, very poor performance according to my tests. You write that is absolutely NORMAL but I prefer to write the code so that I have a fast download, pretty much the same speed that I have on my current desktop PC.

At the moment, Marc-F., you assume that the code will result in bad performance on the board you are targetting. If this should be the case, you can modify the code. But please determine that there is really a problem and not ask to have the code changed because you assume there may be a problem.

I was also surprised that a small packet size resulted in fast performance, but that is what happened.

mfld-fr:
No, sorry Georg, I am not assuming. YOU are assuming, and you build your rational on a tricky behaviour of the QEMU emulator, not the behaviour on real networks, as already explained. Your performance booster works for this particular case, but will NOT in general case.

I already observed that direct relation between the performance and the average size of the packet, not only on my SBC while testing the packet driver as a stand-alone code, but also in a professional project some years ago, where we maximized the throughput on a 1 GB/s link with direct control of the Intel chip.

We achieved 95 MB/s with large frames, not small ones, to transfer medical images from an acquisition device to a review workstation (HP 8600 with Linux). Simply because the performance killer is the packet overhead, and one must maximize the data payload / minimize the protocol overhead.

This is by the way the main reason why I wrote the NE2K driver in 8086 assembly, rather than in C: to minimize the time to process not only the payload, but the Ethernet overhead.

You could also read some serious studies that lead to the same conclusion. For example:  https://pdfs.semanticscholar.org/23b6/90b7188d2075a400264636a092c21f689d1e.pdf

I don't understand why you still like a limpet to his rock on that topic, refusing to consider the overall sea ?..

PR #89, opened 2017-03-08
Reduction of kernel stack size: first part

lithoxs:
BSS size reduced by 208 bytes.
Code size reduced by 48 bytes.


lithoxs:
The t_regs field in struct task is mostly unused. The idea is to put this field above the tops of the kernel
stack, and order the fields of the struct __registers to match the order required by syscalls. Also, match the order to pop them from the stack upon returning to userspace.
This efectively makes the t_regs field part of the kernel stack.

mfld-fr:
I like the idea. I did the same matching in MON86 (see https://github.com/jbruchon/dev86/tree/master/mon86). It costs a little more space on the stacks, but simplifies the whole.

PR #91, opened 2017-03-09
Reduction of kernel stack size: second part.

lithoxs:
Syscalls and interrupts now have the same stack layout.
Also share the code to save/restore registers.
Code size reduced by 64 bytes and BSS reduced in 32 bytes.


lithoxs:
Also the *pt_regs field in interrupts functions now actually have a pointer to the registers. Before, it had a near pointer to a far pointer to the registers, undocumented.

Also, the source code for the interrupt handler had a comment:

"!	We do all of this to avoid per process interrupt stacks and
 !	related nonsense. This way we need only one dedicted int stack"

However, all interrupts are taken using the current kernel stack and the above promise has not been fulfilled so far. I will fix this later.

PR #107, opened 2017-03-29
Claim 61Kb of unused memory for user processes.

lithoxs:
Done by relocating the kernel code and data segments to the lowest possible location.
Tested with the new telnetd daemon and a standard linux telnet client.
The vi editor now start on the client side, but complains of 'no TERM'.
Code size increased by 96 bytes.
The increase in usable memory was determined running the meminfo command before
and after the modification.

### DIFF
--- a/elks/arch/i86/boot/crt0.S
+++ b/elks/arch/i86/boot/crt0.S
@@ -30,6 +30,76 @@ _main:
 	pop	dx
 #endif
 
+#ifdef REL_SYS
+! Relocate kernel code and data to lower memory
+! First set the stack next to the kernel data segment
+
+	mov	ax,si
+	add	ax,dx
+	rcr	ax,#1
+	shr	ax,#1
+	shr	ax,#1
+	shr	ax,#1
+	add	ax,cx
+	mov	ss,ax
+	mov	sp,#768
+	push	dx		! Save registers
+	push	si
+	push	cx
+	push	bx
+
+! Relocate setup data
+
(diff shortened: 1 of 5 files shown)


lithoxs:
Memory map before relocation:

From      To     Size     Use

00000 -> 003FF   1024     Real mode interrupt vector table
00400 -> 004FF    256     Bios Data Area
00500 -> 00FFF   2815     Unused Ram
01000 -> 011FF    512     Setup Data (Relocated Boot Loader)
01200 -> 01880   1664     Unused Ram (Setup program)
01880 -> 07BFF  25472     Unused Ram
07C00 -> 07DFF    512     Unused Ram (Bios Boot Loader)
07E00 -> 07FFF    512     Unused Ram
08000 -> 081FF    512     BUFSEG
08200 -> 0FFFF  32255     Unused Ram
10000 -> 1001F     32     Unused Ram (Kernel Minix Header)
10020 -> ?????            Kernel code, data and bss

Memory map after relocation

From      To     Size     Use

00000 -> 003FF   1024     Real mode interrupt vector table
00400 -> 004FF    256     Bios Data Area
00500 -> 005FF    256     Additional Bios Data Area
00600 -> 007FF    512     BUFSEG
00800 -> 009FF    512     Setup Data
00A00 -> ?????            Kernel code, data and bss

lithoxs:
BUFSEG is a 512 byte block guaranteed to not cross a 64K boundary. Used by doshd.c.

PR #111, opened 2017-04-08
Add FAT filesystem support to ELKS

georgp24:
This is the first version of the FAT filesystem support for ELKS. This allows to use both the minix and the FAT filesystem with ELKS. 

Please see the file `fat_fs.txt` in the Documentation directory for further details.


lithoxs:
What is the reason for this change?:

 @@ -381,7 +381,7 @@ int sys_open(char *filename, int flags, int mode)
  	 * an incomplete fd to other processes which may share
  	 * the same file table with us.
  	 */
   \-	      if ((error = get_unused_fd(filp)) > -1) goto exit_open;
   \+	      if ((error = get_unused_fd(filp)) != -1) goto exit_open;
  	    close_filp(pinode, filp);
  	}
  	iput(pinode);
 
If successful, function get_unused_fd() returns a valid file descriptor (>= 0). On failure returns -EMFILE (-24). With the above change, failure of said function will go unnoticed.

georgp24:
Thank you for pointing out this error. I tried to work around a bug in bcc.

mfld-fr:
You tried to work around a bug in bcc ?
See https://github.com/jbruchon/elks/pull/117.

PR #114, opened 2017-04-14
Enable mkdir for the FAT filesystem. 

georgp24:
Get cp to work within FAT disks too.

### DIFF
--- a/elks/Documentation/text/fat_fs.txt
+++ b/elks/Documentation/text/fat_fs.txt
@@ -31,11 +31,8 @@ You can also access USB flash disks from Qemu and read files from that. I format
 flash disk with a 480 MB partition as number one and could read a file from that using
 ELKS and the FAT32 support.
 
-Not supported is the mkdir command. Also the mv command will only work if one of the 
-specified files is on a minix disk. You cannot move a file just on a FAT disk.
-
-You have to execute the „sync“ command to write the buffers to the disk to effectively 
-remove the file after the rm command though.
+The mv command will only work if one of the specified files is on a minix disk. You 
+cannot move a file just on a FAT disk. Also rm and ln do not work.
 
 Currently, FAT16 is not supported. The reason is that you cannot list the files
 on the FAT16 disk since the readdir command will hang in map_buffer() in fs/buffer.c. 
@@ -57,7 +54,6 @@ A FAT12 floppy image can be mounted with: "mount -t msdos /dev/fd1 /mnt". In the
 script you have to specify the floppy image e.g. with:
 "qemu-system-i386 -fda elkscmd/full3 -fdb ./fddos"
 
-The mkdir command does not work since the code still calls minix_mkdir() and not msdos_mkdir().
 The code contains long file name support but "ls" will still report the files as "xxx~1" and
 you have to use this file name to open the files.
 
--- a/elks/fs/namei.c
+++ b/elks/fs/namei.c
@@ -416,7 +416,9 @@ int do_mknod(char *pathname, size_t offst, int mode, dev_t dev)
     struct inode *dir;
     char *basename;
     size_t namelen;
-    int (*op) ();
+    /* declare function pointer with varying arguments as specified by () 
+     * a pair of parentheses to bind the * to the name instead of the return type */
+    int (*op) (); 
     int error = dir_namei(pathname, &namelen, &basename, NULL, &dir);
 
     if (!error) {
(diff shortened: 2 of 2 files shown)


lithoxs:
In 38b979, georgp24 credits himself with fixing two regression problems, one of them in file fs/namei.c.
Now, in this pull request he credits himself with enabling mkdir for FAT filesystems.

The problem is that in this pull request he undoes the changes that he did in 38b979 to file namei.c, uncovering the fact that there was no problem with fs/namei.c in the first place.

Examine function do_mknod() in fs/namei.c prior to 38b979 and compare it with the latest version and you will find that the only differences, besides a few comments, are insignificant code changes whose only purpose is to increase the code size.

This is totally unacceptable.

jbruchon:
What would you like me to do?

lithoxs:
Function do_mknod() in fs/namei.c should be left exactly as before 38b979. Also, I expect apologies from GeorgP24

jbruchon:
Do you object to this code fragment in `do_mknod()`? It was added in this commit. Otherwise my working copy has been made functionally identical to the version you specified (do you mean commit a38b979?) while preserving the style and whitespace cleanups I've been doing. I cleaned up the whitespace that got added in this commit as well.

```
if (offst != offsetof(struct inode_operations,mknod))
  error = op(dirp, basename, namelen, mode);
else
  error = op(dirp, basename, namelen, mode, dev);
```

I can't extract apologies from people, so there isn't much I can do for the personal side of this.

lithoxs:
suggest changing the following fragment from function do_mknod() in fs/namei.c
in current HEAD:

----------------------------------------------------------------------
        else if (!(error = permission(dirp, MAY_WRITE | MAY_EXEC))) {
              iop = dirp->i_op;
              /* cast function pointer in offst to op. Function address in offst is 
               * relative to iop structure address*/
              op = *(int (**)())((char *)iop + offst);
  #if 0
              printk("iop->mknod:%u,iop->mkdir:%u,iop->symlink:%u\n",\
                  &(iop->mknod),&(iop->mkdir),&(iop->symlink));
              printk("\niop:%u,offst:%u,iop+offst:%u,op:%u,dev:%d|\n",\
                  iop,offst,(char *)iop+offst,op,dev);
  #endif                
              if (!iop || !(op)) { 
                error = -EPERM; 
              } else {
                dirp->i_count++;
                down(&dirp->i_sem);

                if (offst != offsetof(struct inode_operations,mknod))
                    error = op(dirp, basename, namelen, mode);
                else
                    error = op(dirp, basename, namelen, mode, dev);
                  
                up(&dirp->i_sem);
-----------------------------------------------------------------------

with my original code plus style corrections (2e1b7a4)

-----------------------------------------------------------------------
        else if (!(error = permission(dirp, MAY_WRITE | MAY_EXEC))) {
              iop = dirp->i_op;
              if (!iop || !(op = (*(int (**)())((char *)iop + offst))))
                  error = -EPERM;
              else {
                  dirp->i_count++;
                  down(&dirp->i_sem);
                  error = (offst != offsetof(struct inode_operations,mknod)
                                  ? op(dirp, basename, namelen, mode)
                                  : op(dirp, basename, namelen, mode, dev)
                              );
                  up(&dirp->i_sem);
------------------------------------------------------------------------

georgp24:
I would not have complained about a possible regression problem if I would not have observed a difference. I can only assume now that other files have been changed that cause original code that I changed weeks ago to work now.

I found this while trying to find the cause why the mkdir command would not work in the FAT filesystem. It is not my intention to claim credit for other developers code. When I detect a problem in the current version of the code I try to fix it no matter whether I introduced the problem of someone else. In this case it turned out that the code I had put in did not work properly anymore.

While analysing this code section I changed the ternary operators to if/else statements to increase the readability and added comments and debugging statements to make the code easier to understand and allow for debugging should further changes become necessary. I do not think that the compiled code increases in size if you replace a ternary operator with if/else statements.

Therefore I do not think it is an advantage to replace if/else statements with ternary operators again. I read the recommendation in several places to replace ternary operators with if/else statements for improved  readability.

lithoxs:
First the easy part. Whether or not there is an increase in code size by replacing (?:) with if/else is not a matter of opinion or intuition. It can be measured.

So, I compiled the current HEAD and look at the size of fs/namei.o. It is 3294 bytes.
Compiling with my original code, the size of fs/namei.o is 3286 bytes. Thus, the replacement increases code size.

The reason for this is that BCC is very bad at optimizing. I think it generates two assignment instruction sequences instead of one. Of course, 8 bytes is not very much, but if you do things unfriendly to BCC in every opportunity, you easily end up with a kernel that cannot be compiled.

lithoxs:
Now, the theory that other files have changed that cause the original code to now work.

Function do_mknod() is called only from fs/namei.c, and the calls are from syscalls, so there are not other files involved. Modification to fs/namei.c were done by Georg, the first to introduce the problem with mkdir and the last is this pull request. Other than that, there are only style and space cleanups.

The only other place where do_mknod() is called is from net/unix/af_unix.c in function unix_bind(). This file was last modified by Georg by the time he  posted 38b979.

So, there are not files changed that can do the magic.

lithoxs:
The change posted by Georg in a38b979 is an older version of do_mknod(), namely, d9d6ece. In that version, a code is passed to do_mknod to select the operation as follows:
0, mknod
1, mkdir
2, symlink

The problem is that with a38b979, the functions were passing the offset of the needed function in the structure inode_ops instead of the function code. In that moment, syscalls mknod and mkdir did not work any more, even for minix filesystem, and nobody bothered to check. But unix_bind() started working, which perhaps Georg was trying to fix.

Now that the original code is restored, is obvious that unix_bind() will not work anymore. What is needed is to replace in file net/unix/af_unix.c, line 168:

    i = do_mknod(fname, 0, S_IFSOCK | S_IRWXUGO, 0);

with:

    i = do_mknod(fname, offsetof(struct inode_operations,mknod), S_IFSOCK | S_IRWXUGO , 0);

In short, a38b979 worsened things, because two important syscall stopped working. This pull request leaves things as before a38b979, I mean, unix domain sockets not working.

georgp24:
Yes, I also find that my change makes namei.o `8` bytes larger than Juan's version. However, the file `Image` in `elks/arch/i86/boot` does not change in size. Maybe the optimization is done later in the process.

So at the moment it has to be decided if my if/else shall stay because of better readability of the code or a ternary operator shall be used as requested by Juan. The code will have identical results.

I did observe that mkdir worked with the FAT file system after the change in my pull request. I did not test unix domain sockets. Currently I do not recall what I tried to fix in namei.c before, but I posted a working version of unix domain sockets the day after that so it may have to do with that.

lithoxs:
The reason for Image not changing is that the total size is rounded to be a multiple of 16 bytes. With a change of 8 bytes it is likely that we won't see a change in Image size. However many small savings are added up and the end result rounded.

There are not any optimizations later. Code sizes will be different, larger in if/else style.

The question here is, did you test mkdir and mknod after a38b979?.
Is it fair for you to say this is your pull request when in reality you are restoring someone else's code?

lithoxs:
I just got back to a38b979, compiled and run it. I ran commands mkdir y mknod and in both cases they produced a symbolic link.

This confirms that a38b979 actually worsened the code. Of course, restoring the original code makes these command to work again.

jbruchon:
I fixed the code up. Let me know if it's broken now.

lithoxs:
I just pulled the code, compiled the kernel and tested commands mknod, mkdir and ln -s. Everything is fine. Thanks.

jbruchon:
I also had success in my tests. Thanks for verifying.

PR #117, opened 2017-04-21
Fix bug in fs/open.c

georgp24:
Pull request fixes a bug in open.c located by Juan.

bcc apparently has a bug when comparing with `-1`. I added a test program to demonstrate the problem. Below is the output of the program which shows that bcc determines e.g. a `+10` as being smaller than `0`.

greater_minus_one


mfld-fr:
That test program has two defects, and demonstrates that BCC behaves correctly.

georgp24:
Marc,

then please explain why `if (10 > -1)`  should result in `10 is smaller than 0`

Georg

mfld-fr:
Just look at the type of the variable...

jbruchon:
Signed types use the high bit = 1 for negative numbers, so the implicit cast of signed -1 to unsigned will result in the unsigned number 0xFFFF, hence `(ANY_number > (uint16_t)-1)` will always return the same fail result.

georgp24:
Thank you, I understand it now. I had found a problem before in msdos/dir.c where there was a `if (__u16 > -1)` which did not work while in fs/namei.c this is a `if (int > -1)` instead.

lithoxs:
This pull request should be applied because the bug referenced in pull request #111 still holds and the modification here fixes the problem. The test program and Makefile were already removed.

PR #118, opened 2017-04-26
Simplifications to reduce code size.

lithoxs:
Fix to problem when removing inodes.
Pending fix in unix domain sockets.
code size reduced by 32 bytes.
Tested with Qemu.

### DIFF
--- a/elks/fs/minix/bitmap.c
+++ b/elks/fs/minix/bitmap.c
@@ -160,10 +160,10 @@ void minix_free_inode(register struct inode *inode)
 	s = "nonexistent imap\n";
     else {
 	map_buffer(bh);
-	clear_inode(inode);
 	if (!clear_bit((unsigned int) (inode->i_ino & 8191), bh->b_data)) {
 	    debug1("%s: bit %ld already cleared.\n",ino);
 	}
+	clear_inode(inode);
 	mark_buffer_dirty(bh, 1);
 	unmap_buffer(bh);
     }
--- a/elks/fs/minix/namei.c
+++ b/elks/fs/minix/namei.c
@@ -407,51 +407,53 @@ int minix_rmdir(register struct inode *dir, char *name, size_t len)
     inode = NULL;
     bh = minix_find_entry(dir, name, len, &de);
     retval = -ENOENT;
-    if (!bh) goto end_rmdir;
-    retval = -EPERM;
-    if (!(inode = iget(dir->i_sb, (ino_t) de->inode))) goto end_rmdir;
-    if ((dir->i_mode & S_ISVTX) && !suser() &&
-	current->euid != inode->i_uid && current->euid != dir->i_uid)
-	goto end_rmdir;
-    if (inode->i_dev != dir->i_dev) goto end_rmdir;
-    if (inode == dir)		/* we may not delete ".", but "../dir" is ok */
-	goto end_rmdir;
-    if (!S_ISDIR(inode->i_mode)) {
-	retval = -ENOTDIR;
-	goto end_rmdir;
-    }
-    if (!empty_dir(inode)) {
-	retval = -ENOTEMPTY;
-	goto end_rmdir;
-    }
-    if (de->inode != inode->i_ino) {
-	retval = -ENOENT;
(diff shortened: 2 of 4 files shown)


Mellvik:
I just tested the latest ELKS - built from source this morning - on a Compaq portable III (286/AT class). keyboard input seems to be broken:
- Works for 15 chars - echo and everything, then - on the 16th - system hang (hard, powert cycle needed).
- If one of the chars is esc, the behaviour is different. Echo stops and when the buffer is full, I get the usual beep per char forever. Nothing works except the beep, and the system can be ctrl-alt-delet booted.
- No virtual consoles (unless the magic console switch key is something other than the usual ctrl-Fx og ctrl-alt-Fx).

I will be looking at the sources, but I’m not really qualified to do debugging at this point.


Helge S

lithoxs:
You description leads me to conclude that you are using the BIOS console driver option.
Using that option I was able to replicate the behavior described by you using the PCE emulator.
A quick fix to your problem is to use the direct console driver. The problem has nothing to do with this pull request.
If you still want to look at the sources, see file arch/i86/drivers/char/bioscon.c

lithoxs:
I just posted a pull request that may solve the problem for Mellvik. It is necessary to make a test to confirm that the problem is solved for him.
Regarding the problem of virtual consoles, in the bios console driver do not (and never have) work(ed).

Mellvik:
Thanks - and yes, I was using the bios console driver.
The direct driver works as far as I/O is concerned, but does not accomodate keyboard functions like capslock, scrollock, numlock with lights in the keys etc. They’re just dead.

So I’m heading back to the bios console, where your most recent change fixes the problem. That gave me the opportunity to discover a new one. Control-keys don’t work. Numeric keypad keys work if num lock is on, otherwise not. Function keys don’t work. All have the same effect: After hitting one of them once, the keyboard starts beeping at any keypush. No more input, just beeps - until reboot. Soft reboot works, and keyboard lights (numlock etc.) goes on and off as normal.

Further, I couldn’t get mount -t msdos to work and ps only prints a heading, but I’ll look more into those later.


Helge

Mellvik:
A couple of things while we’re at it:

- Console keyboard: deleting text: The first backspace moves the cursor back, but doesn’t visibly delete the char. More backspaces delete (visibly) the char that the previous backspace should have deleted. So when start of line is reached, there is one char left - in the cursor position. The buffer content is correct.

- FAT FS: Enabling MSDOS file system support in config does nothing, unless ‘Full VFS’ is also enabled.
- FAT FS: I can mount a FAT FS but the system hangs when acessing it: Error message: "BUFMAP: buffer #51 waiting on L1 slot", and the system is in a hung state. Yes, I do have external buffering enabled. The buffer # varies. After the message, the system is soft-hung. (see attached photo).

- Networking: Maybe it would make sense to not include the /bin/ktcpd file when networking is not configured, so that network configuration is not attempted at boot…

- Config: Changing system type to PC/AT does not change the processor setting to 286. Does it matter?

—Helge (aka mellvik)

georgp24:
Helge,
in the file `elks/Documentation/text/fat_fs.txt` it is documented that you have to enable `Full VFS` too.

I observed the problem you describe with FAT16 only. It is documented that this does not work yet.

I would be grateful if litoxs would look into this problem. The function `msdos_get_entry()` in `msdos/misc.c` which is called in `msdos/dir.c` line 244 calls `msdos_sread()`  in `msdos/misc.c` which calls `map_buffer()` in `fs/buffer.c` and hangs with this message.

Georg

lithoxs:
I suggest continuing this discussion at other place. It may be the mailing list or you may open an issue. Your observations will have a greater audience and there will be greater chances for the problems to be solved.

All the problems mentioned are unrelated to the purpose of this discussion thread. Also, here we are only 4 participants.

mfld-fr:
Plus the ones that receive the notifications from the project ;-)

Mellvik:
Ok,
I wanted to keep the noise level down, assuming there might be things I had missed (turned out to be correct).
Anyway, I'll take you up on the advice and open issues instead... assuming that's more effective.

Helge

PR #119, opened 2017-04-27
Fix to BIOS console driver.

lithoxs:
This makes the bios console driver work under Qemu. With PCE emulator still don't work.
I cannot guess what will happen with real hardware.


Mellvik:
With real hardware it just does not work (like described yesterday).

This fix didn’t help the compaq. Now that you pointed me to the right place, I’ll see if I can trace it.

HS

Issue #121, opened 2017-04-28
FAT16 support hangs with message ""BUFMAP: buffer #.... waiting on L1 slot"


georgp24:
When you mount a FAT16 disk and use `ls` to list the files on this disk the message above will occur.

This is as far as I could trace it: 
in `msdos/dir.c` the function `msdos_get_entry()` is called in line 244. This function is implemented in `msdos/misc.c`. There it calls `msdos_sread()` which calls `map_buffer()` in `fs/buffer.c`. The function `map_buffer()` hangs with the message above. 

The `ls` command causes these calls to be repeated several times until `bread()` in `msdos_sread()` returns a `buffer_head ptr` that cannot be handled in `map_buffer()`.

You can generate a FAT16 disk image for testing by modifying lines 41 and 43 in
 `elkscmd/test/fat/build_fat32_hd_image.sh`. I.e uncomment line 43 and comment out line 41.

Georg

PR #122, opened 2017-05-02
Fix to issue #121.

lithoxs:
Only did a quick test. I suggest to make an extensive test.


georgp24:
Great! The `ls` command works fine now.

However, `cd /mnt` and e.g. `cp fatdisk1 testfil` does not work, so you cannot copy files on the FAT16 disk. 

You can copy files from the FAT16 disk to a minix disk but you cannot copy files from the minix disk to the FAT16 disk.
fat16-cp-problem

lithoxs:
Please test the above commit.
The file copy problem is unrelated to the ls command problem.
I guess that there are more problems in this filesystem.
It will be bettter to have the most possible complete list of problems.

PR #126, opened 2017-05-03
Fixes to BIOS console driver.

lithoxs:
Virtual consoles, control key, function and arrow keys, backspace.
Tested under Qemu. Thus, virtual consoles not fully tested because windowing
system intercepts Alt-Fn.
Arrow keys woks as an ANSI (or VT52, as configured) terminal.
Function keys send ESC 'a'  up to ESC 'l'.
Backspace now works as expected.

### DIFF
--- a/elks/arch/i86/drivers/char/bioscon.c
+++ b/elks/arch/i86/drivers/char/bioscon.c
@@ -105,6 +105,7 @@ int kraw = 0;
 #endif
 
 static void std_char(register Console *, char);
+static void Console_set_vc(unsigned int);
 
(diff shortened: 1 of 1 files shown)


lithoxs:
Bear in mind that the internal bios driver only knows about US keyboard keycodes.
The BIOS console driver in ELKS is intended to be used if the direct driver fails to work.
Currently, there is a driver for the PC-XT keyboard.
Recently, Georg added some support for PC-AT keyboard to the direct keyboard driver.
Perhaps Georg could write a separate driver for full PC-AT keyboard direct driver.

Issue #123, opened 2017-05-03
MSDOS fs buffer hang & more


Mellvik:
Mounting a FAT16 FS works (tested on physical and virtualbox), files can be copied in and out, some issues remain:
- deleting a file from the msdos fs appears to work (the shell prompt comes back), but the next cmd always soft-hangs the system, regardless of pwd. 
- ls -l on DOS files give wrong sizes and dates on most files (see top of attached photo)
- After working with the mounted FS for a dozen or so commands, the 'Buffer #x waiting for L1' error comes back and the system soft hangs (see photo)

![img_3661](https://cloud.githubusercontent.com/assets/3629880/25653999/aecf119a-2fef-11e7-8f9f-a3be61a56e22.JPG)

georgp24:
I tried to reproduce this and cannot. ELKS shows the right file sizes on my FAT16 image and I did not get your buffer error. 

Is the `wina20.38_` a compressed file or what type of file? Why is there an underscore at the end of the file extension?

Mellvik:
The trailing underscore was the ‘standard’ M$ distribution format for MSDOS on floppies. The files are compressed but the file system does (obviously) not know anything about that, so a dir listing on DOS lists the physical sizes. I can send you a listing of the entire directory, notice that some of the dates are wrong too.

The buffer problem is repeatable, possibly not occuring at exactly the same point in a chain of commands …

georgp24:
I can see on your photo that the file sizes are wrong. If your FAT16 hard disk image is smaller than 10 MB you may attach it to your post so I can test it.

Mellvik:
I have no way to get the disk image off the machine…
There is no network on the old klunker (at this time, I ordered an old NE2k card the other day, so maybe in a few weeks).

Here’s the list - the sizes and dates are frequently different from what I get from DOS and none of the files are actually directories, so there are several issues here. I’m attaching a photo of part of the dos listing for reference. I can get you a full dos listing too if it’s useful.

cat /mnt/floppy/xxx

tmp:
-rwxr-xr-x  1 root     root         6261  May 31 1994  ansi.sy_
-rwxr-xr-x  1 root     root         7467  May 31 1994  append.ex_
-rwxr-xr-x  1 root     root        11208  May 31 1994  attrib.exe
-rwxr-xr-x  1 root     root           38  May 31 1994  autoexec.bat
-r--r--r--  1 root     root     3523999830  Apr 09 2015  av.gr_
[...]

PR #128, opened 2017-05-15
Fix to access to inodes in FAT16 FS, issue #123.

lithoxs:
Tested only with Mellvik's dos286.img.
Uses 32 bit inode numbers when msdos-fs is enabled,
otherwise, uses 16 bit inode numbers.
There is an increase in code size of 480 bytes when msdos-fs enabled.
Also modified configuration to enable Full VFS when msdos-fs enabled,
without need to ask.

### DIFF
--- a/elks/fs/Config.in
+++ b/elks/fs/Config.in
@@ -8,11 +8,16 @@ mainmenu_option next_comment
 	bool 'All file systems are READ-ONLY'  CONFIG_FS_RO               n
 	bool 'Minix file system'               CONFIG_MINIX_FS            y
 	bool 'ROM file system'                 CONFIG_ROMFS_FS            n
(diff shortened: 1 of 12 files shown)


mfld-fr:
Thank you Juan to finish the job !

Mellvik:
Works on Compaq 286, and also on VirtualBox. The testing serves a reminder about the BUFMAP problem, which can consistently be repeated on both platforms: Mount the dos-partition, cd to /mnt, ls tmp, ls tmp1 (which is a copy of the entire tmp directory. See attached photos.
virtualbox_elks_16_05_2017_18_04_10

img_3707

mfld-fr:
Tracked by #121, fixed by #129.

PR #129, opened 2017-05-16
Fix to 'L1 buffer hang ..' described in issue #123.

lithoxs:
Tested only with Mellvik's dos286.img. This solves many problems, but other memory leaks may remain. Please test as extensively as possible.

### DIFF
--- a/elks/fs/msdos/inode.c
+++ b/elks/fs/msdos/inode.c
@@ -44,7 +44,7 @@ void msdos_put_inode(register struct inode *inode)
     if (!inode->i_nlink) {
 	inode->i_size = 0;
 	msdos_truncate(inode);
-	memset(inode,0,sizeof(struct inode));
+	clear_inode(inode);
     }
 }
 
--- a/elks/fs/msdos/namei.c
+++ b/elks/fs/msdos/namei.c
@@ -106,7 +106,7 @@ int msdos_lookup(register struct inode *dir,const char *name,int len,
 		iput(dir);
 		return res;
 	}
-	if (bh) brelse(bh);
+	if (bh) unmap_brelse(bh);
 /* printk("lookup: ino=%ld\r\n",ino); */
 	if (!(*result = iget(dir->i_sb,ino))) {
 		iput(dir);
@@ -144,7 +144,7 @@ static int msdos_create_entry(register struct inode *dir,char *name,int is_dir,
 	de->size = 0;
 	bh->b_dirty = 1;
 	if (*result = iget(dir->i_sb,ino)) msdos_read_inode(*result);
-	brelse(bh);
+	unmap_brelse(bh);
 	if (!*result) return -EIO;
 	(*result)->i_mtime = 
 	    CURRENT_TIME;
@@ -170,7 +170,7 @@ int msdos_create(register struct inode *dir,const char *name,int len,int mode,
 	lock_creation();
 	if (msdos_scan(dir,msdos_name,&bh,&de,&ino) >= 0) {
 		unlock_creation();
-		brelse(bh);
+		unmap_brelse(bh);
 		iput(dir);
 		return -EEXIST;
  	}
@@ -200,7 +200,7 @@ int msdos_mkdir(struct inode *dir,const char *name,int len,int mode)
 	if (msdos_scan(dir,msdos_name,&bh,&de,&ino) >= 0) {
 
 		unlock_creation();
-		brelse(bh);
+		unmap_brelse(bh);
 		iput(dir);
 		return -EEXIST;
  	}
@@ -270,7 +270,7 @@ int msdos_rmdir(register struct inode *dir,const char *name,int len)
 			    DELETED_FLAG && strncmp(dde->name,MSDOS_DOT,
 			    MSDOS_NAME) && strncmp(dde->name,MSDOS_DOTDOT,
 			    MSDOS_NAME)) goto rmdir_done; /* linux bug ??? */
-		if (dbh) brelse(dbh);
+		if (dbh) unmap_brelse(dbh);
 	}
 	inode->i_nlink = 0;
 	dir->i_mtime = CURRENT_TIME;
@@ -279,7 +279,7 @@ int msdos_rmdir(register struct inode *dir,const char *name,int len)
 	bh->b_dirty = 1;
 	res = 0;
 rmdir_done:
-	brelse(bh);
+	unmap_brelse(bh);
 	iput(dir);
 	iput(inode);
 	return res;
@@ -312,7 +312,7 @@ int msdos_unlink(register struct inode *dir,const char *name,int len)
 	de->name[0] = DELETED_FLAG;
 	bh->b_dirty = 1;
 unlink_done:
-	brelse(bh);
+	unmap_brelse(bh);
 	iput(inode);
 	iput(dir);
 	return res;


Mellvik:
Tested on physical 286 and Virtual box. Buffer hang is gone. System still hangs when deleting a file. If the system date is not set (ie. is default Jan 1st 1970), locally copied files on the dos partition gets wrong time stamp (Jan 1st, 2098).
ON a physical machine, disk access is extremely slow. ls -l takes loooong, cp tmp1/* tmp2 using the Mellvik DOS volume takes forever (15+ minutes). Looks like every block physically passes thru the floppy, floppy activity is continuously very high.

lithoxs:
As before, tested with Mellvik's dos286.img under Qemu.
I guess that the date stamp of copied files and copy speed are unrelated to each other and to problems already solved.
Also, the kernel configuration used in your tests will be of great help.

Mellvik:
config file enclosed.

lithoxs:
I could not find you config file.
Regarding the date stamp on msdos, the problem don't have solution. Under FAT filesystem, there are only 7 bits to represent the year, with all zeros meaning 1980 and all ones meaning 2107. It is impossible to have any year less than 1980.
While it is possible to replace any date before 1980 with the minimum FAT filesystem date, I will leave the code as is, in the interest of clarity.

Mellvik:
Sorry about the config file, find it pasted in below.

And thanks for clarifying the date issue. Not a problem. As a matter of fact, if the kernel had picked up the date from the BIOS (or initialized it to something closer to now than 0), the issue would likely never have popped up.

Helge

iMac21H:tmp helge$ cat config.txt 
#
# Automatically generated make config: don't edit
#

#
# Just accept the defaults unless you know what you are doing
#

#
# Advanced - for Developers and Hackers only
#
CONFIG_EXPERIMENTAL=y
# CONFIG_MODULES is not set

#
# Architecture
#
# CONFIG_ARCH_AUTO is not set
CONFIG_ARCH_IBMPC=y
# CONFIG_ARCH_SIBO is not set
# CONFIG_PC_AUTO is not set
# CONFIG_PC_XT is not set
CONFIG_PC_AT=y
# CONFIG_PC_MCA is not set
# CONFIG_ROMCODE is not set
CONFIG_SMALL_KERNEL=y

#
# Model-specific features
#
# CONFIG_IBMPC_CLONE is not set
CONFIG_IBMPC_COMPAQ=y
# CONFIG_COMPAQ_FAST is not set
# CONFIG_ROOT_READONLY is not set

#
# Memory manager
#
CONFIG_ADVANCED_MM=y
# CONFIG_SWAP is not set

#
# Hardware Facilities
#
# CONFIG_CPU_8086 is not set
# CONFIG_CPU_80186 is not set
CONFIG_CPU_80286=y
# CONFIG_CPU_80386 is not set
# CONFIG_CPU_80486 is not set

#
# System Hardware
#
CONFIG_HW_FPU=y
# CONFIG_HW_PS2_MOUSE is not set
CONFIG_HW_VGA=y

#
# Block Devices
#
CONFIG_HW_FLOPPY_DRIVE=y
CONFIG_HW_HARD_DRIVE=y

#
# Memory Configuration
#
CONFIG_MEM_PAGES=8
# CONFIG_XMS is not set
# CONFIG_XMS_INT15 is not set
CONFIG_XMS_SIZE=0

#
# BIOS Facilities
#
CONFIG_HW_KEYBOARD_BIOS=y
CONFIG_HW_VIDEO_HOC=y
# CONFIG_HW_259_USE_ORIGINAL_MASK is not set
CONFIG_BOGOMIPS=0
# CONFIG_HW_USE_INT13_FOR_FLOPPY is not set
# CONFIG_HW_USE_INT13_FOR_DISKPARMS is not set
CONFIG_APM=y

#
# 286 Protected Mode Support
#
# CONFIG_286PMODE is not set

#
# Driver Support
#

#
# Block device drivers
#
CONFIG_BLK_DEV_BIOS=y
CONFIG_BLK_DEV_BFD=y
CONFIG_BLK_DEV_BFD_HARD=y

#
# Currently ignored and assumed to be 2880k for both drives.
#
# CONFIG_BLK_DEV_BFD0_NONE is not set
# CONFIG_BLK_DEV_BFD0_360 is not set
# CONFIG_BLK_DEV_BFD0_720 is not set
# CONFIG_BLK_DEV_BFD0_1200 is not set
CONFIG_BLK_DEV_BFD0_1440=y
# CONFIG_BLK_DEV_BFD0_2880 is not set
CONFIG_BLK_DEV_BFD1_NONE=y
# CONFIG_BLK_DEV_BFD1_360 is not set
# CONFIG_BLK_DEV_BFD1_720 is not set
# CONFIG_BLK_DEV_BFD1_1200 is not set
# CONFIG_BLK_DEV_BFD1_1440 is not set
# CONFIG_BLK_DEV_BFD1_2880 is not set
CONFIG_BLK_DEV_BHD=y
# CONFIG_BLK_DEV_FD is not set
# CONFIG_BLK_DEV_HD is not set
# CONFIG_DMA is not set

#
# Additional block devices
#
# CONFIG_BLK_DEV_RAM is not set

#
# Block device options
#
CONFIG_BLK_DEV_CHAR=y

#
# Character device drivers
#
# CONFIG_CONSOLE_DIRECT is not set
CONFIG_CONSOLE_BIOS=y
# CONFIG_CONSOLE_SERIAL is not set
CONFIG_EMUL_ANSI=y
# CONFIG_EMUL_VT52 is not set
# CONFIG_EMUL_NONE is not set
# CONFIG_EMUL_ANSI_PRINTK is not set

#
# Other character devices
#
CONFIG_CHAR_DEV_RS=y
CONFIG_CHAR_DEV_LP=y
CONFIG_CHAR_DEV_MEM=y
CONFIG_PSEUDO_TTY=y
# CONFIG_DEV_META is not set

#
# Network device drivers
#
# CONFIG_ETH is not set

#
# Networking Support
#
# CONFIG_SOCKET is not set

#
# Filesystem Support
#
# CONFIG_FS_RO is not set
CONFIG_MINIX_FS=y
# CONFIG_ROMFS_FS is not set
CONFIG_MSDOS_FS=y

#
# General filesystem settings
#
CONFIG_FULL_VFS=y
CONFIG_32BIT_INODES=y
CONFIG_FS_EXTERNAL_BUFFER=y
CONFIG_PIPE=y

#
# Executable file formats
#
CONFIG_EXEC_ELKS=y

#
# Kernel hacking
#
CONFIG_SYS_VERSION=y
# CONFIG_STRACE is not set

lithoxs:
I was able to reproduce the behavior described by you, using the PCE emulator. It took 14 minutes to make
cp /mnt/tmp/* /mnt/tmp1
Using Qemu it takes nothing. With Dioscuri (which also emulates the speed for a 5 Mhz 8088) takes less than 1 minute the copy.
So far, the only clue is that the floppy shows a lot of activity.

jbruchon:
I think DOS etc. interleaved disk writes due to the slowness of the system because it took longer to push the data to the floppy controller than it took for a single sector to be written. It's been a long time since I read up on it but I think DOS would jump 3 sectors per track when dumping out a bunch of floppy sectors. I could be wrong.

Mellvik:
I don't think interleaving is part of this (and interleaving or skew is not mentioned in any Microsoft document I could find related to FAT). However, in one of the documents there is a hint that may be at least part of the problem:
"Because FAT directories are not sorted or indexed, it is a bad idea to create huge directories; otherwise, operations like creating a new entry (which requires every allocated directory entry to be checked to verify that the name doesn’t already exist in the directory) become very slow."
This is a fairly big directory for a FAT filesys, while most files are relatively small. Is the intense floppy activity possibly related to directory searches and file creation rather than file transfer?

lithoxs:
I don't think so. The file transfer is from hard disk to hard disk. The floppy has nothing to do in that operation.

georgp24:
It may be that ELKS, if booted from a floppy image, writes data to the floppy disk while executing the transfer. 

Wonder if it will work at all if you make the floppy read-only.

lithoxs:
I put a marker when accessing the floppy, use a minix-fs image mounted on /mnt1 and did these tests:

1.Keep CWD=/root and do "cp /mnt/tmp/* /mnt/tmp1". Copy time ~ 14:24 min. About 74 floppy reads.
2. Put CWD=/mnt and do "cp tmp/* tmp1". Copy time ~ 14:15 min. No floppy reads.
3. CWD=/root and do "cp /mnt/tmp/* /mnt1/tmp". Copy time ~ 9:45 min. About 74 floppy reads.
4. CWD=/mnt1 and "cp tmp/* tmp1". Copy time ~ 8:55 min. No floppy reads.

My conclusion is that the floppy access is to lookup the filename, which starts in the floppy in tests 1 and 3. In tests 2 and 4, the filename starts at CWD and its inode is kept in the task structure, thus no floppy access.
The directory has ~ 140 files. In test 1 there is a file transferred every ~ 6 seconds on average. The floppy motor don't have time to get turned off, so appears to have heavy activity. But actually it reads at most one block per file. No writes.
Msdos fs has a poor perforrmance under ELKS. Minixfs transfers files in 2/3 the time than Msdosfs, so there is room for improvement, but in the end we will hit the limits of ELKS itself.

PR #136, opened 2017-06-13
Make buffer.c and inode.c even more alike.

lithoxs:
Compiled with BCC and IA16-GCC. Tested
with Qemu and PCE. Code size reduced in
16 bytes.


lithoxs:
Also tested with Mentor's ia16-elf-gcc compiler mentioned in the mailing list. There is a code size reduction of 2880 bytes using this compiler compared with that produced by BCC.

mfld-fr:
This makes the compiler question still relevant :smile:

PR #138, opened 2017-06-20
Simplification in fs/namei.c.

lithoxs:
Code size reduced in 112 bytes.
Tested with Qemu and PCE. Compiled with
BCC and IA16-GCC.


jbruchon:
Your code cleanup work is great and I appreciate all of your effort!

PR #143, opened 2017-08-21
Optimizations to reduce kernel stack size.

lithoxs:
Task kernel stack reduced by 7 bytes.
DATA + BSS reduced by 92 bytes. Code size
unchanged.
Basically rewrote function numout() from scratch to avoid using a buffer allocated in the stack, plus some optimizations to keep the code size unchanged.
The savings in the kernel stack are 7 bytes, but only reduced it by 6 bytes to keep an even size for the struct task.

### DIFF
--- a/elks/include/linuxmt/sched.h
+++ b/elks/include/linuxmt/sched.h
@@ -4,7 +4,7 @@
 #define MAX_TASKS 15
 #define NGROUPS	13		/* Supplementary groups */
 #define NOGROUP 0xFFFF
-#define KSTACK_BYTES 1014	/* Size of kernel stacks */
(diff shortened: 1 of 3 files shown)


lithoxs:
The worst case size for the kernel stack is in the system call passing the file name (e.g. sys_open). Then you have recursive calls to dir_namei() -> follow_link() -> minix_follow_link -> open_namei() -> dir_namei() up to 5 times. In the last call to dir_namei(), you can have the worst case in dir_namei() -> lookup() -> minix_lookup() -> iget -> printk() -> numout().

I did a rough estimate for the stack size in this situation and it is about 1/2 Kbyte. I guess there was not any estimate for this and the method used was "Let's allocate 1 Kbytes and forget this issue".

PR #145, opened 2017-08-29
Fix waste of kernel data space in struct super_block.

lithoxs:
Allocation of space for inode and zone bitmaps
was for minix-V2 needs. But we only support
minix-V1, so now allocate space for current needs.
Data space reduced by 480 bytes. Unexpectedly, there
is also a code size reduction of 48 bytes, because
offsets in indexed addressing instructions are smaller.
Compiled with BCC. Tested with Qemu.


krzycz (approved):

jbruchon:
Sorry it took so long to merge!

PR #160, opened 2017-10-15
[romfs] Boot from /dev/rom

mfld-fr:
This is a next step in implementing the boot from ROM (both kernel and root file system in ROM). See mfld-fr#2 for the implementation plan.

- Add a new dummy memory block device `/dev/rom0`
- Force it as the root device in case kernel is started in ROM
- Add a configuration option for ROMFS segment base 
- Add low-level routines for direct access to filesystem in ROM
- Modify previous ROMFS high-level code to call low-level routines
- Plus minor fixes in existing kernel code

Tested under EMU86. ELKS fails to load the `init` program as it is not present in the tested filesystem, and runs `sh` as backup (actually `sash`). Then the `ls` and `cd` commands are used to display the filesystem contents:
```
Loading kernel image...
C
onsole: BIOS 80x25 emulating ANSI (3 virtual consoles)
PC/XT class machine, NEC V20 CPU
512K base RAM.
ELKS kernel (35392 text + 3948 data + 45164 bss)
Kernel text at e062:0000, data at 90:0000
461 K of memory for user processes.
ELKS network sockets
TCP/IP stack by Harry Kalogirou
ELKS version 0.2.0
VFS: Mounted root (romfs filesystem).
Loading init
sys_execve("/sbin/init", args, 18) => 2.
sys_execve("/etc/init", args, 18) => 2.
sys_execve("/bin/init", args, 18) => 2.
No init - running /bin/sh
Stand-alone shell (version 1.0)
> ls -l
ls -l
dr-xr-xr-x  1 root     root           46  Jan 01 1970  bin
dr-xr-xr-x  1 root     root           19  Jan 01 1970  boot
dr-xr-xr-x  1 root     root           38  Jan 01 1970  dev
dr-xr-xr-x  1 root     root           77  Jan 01 1970  etc
dr-xr-xr-x  1 root     root           19  Jan 01 1970  home
dr-xr-xr-x  1 root     root           34  Jan 01 1970  lib
dr-xr-xr-x  1 root     root           19  Jan 01 1970  mnt
dr-xr-xr-x  1 root     root           22  Jan 01 1970  root
dr-xr-xr-x  1 root     root           19  Jan 01 1970  sbin
dr-xr-xr-x  1 root     root           19  Jan 01 1970  tmp
dr-xr-xr-x  1 root     root           23  Jan 01 1970  usr
dr-xr-xr-x  1 root     root           17  Jan 01 1970  var
> cd bin
cd bin
> ls -l
ls -l
-r-xr-xr-x  1 root     root         4124  Jan 01 1970  cat
-r-xr-xr-x  1 root     root         6972  Jan 01 1970  du
-r-xr-xr-x  1 root     root        14844  Jan 01 1970  ls
-r-xr-xr-x  1 root     root         1836  Jan 01 1970  pwd
-r-xr-xr-x  1 root     root         9792  Jan 01 1970  sh
> 

```


lithoxs:
OK to merge

mfld-fr:
@lithoxs : Thank you for that first review. But let us keep that PR open, so that you could review the next commits I will push until the feature works.

mfld-fr:
To make it work up to try loading the super block (see the test result in the updated first comment), I had to hack the code here:
```
diff --git a/elks/fs/super.c b/elks/fs/super.c
index c36e0c1..9742a89 100644
--- a/elks/fs/super.c
+++ b/elks/fs/super.c
@@ -486,7 +486,9 @@ void mount_root(void)
     struct file *filp;
     int retval;
 
-    d_inode = new_inode(NULL, S_IFBLK);
+    /* TODO: allow mounting root from both character and block devices /*
+    /* d_inode = new_inode(NULL, S_IFBLK); */
+    d_inode = new_inode(NULL, S_IFCHR);
     d_inode->i_rdev = ROOT_DEV;
   retry_floppy:
     retval = open_filp(((root_mountflags & MS_RDONLY) ? 0 : 2), d_inode, &filp);
```

Because ELKS only accepts a block device as the root one. Could we consider modifying this part to accept also a character device, or should we rather enable virtual FS (i.e. not loaded from a device, like /proc, /sys and so on) ?

lithoxs:
It will be better if you sketch your general plan rather than providing information incrementally.

Regarding your last question. Function mount_root() does not complain if you try to mount a char device. If you try to remount, sys_mount() will not permit that, but I don't see any reason to remount.

Now, if you implement a virtual fs, you don't need any device, not even for mount_root. The other option is to use the VFS and then (I guess) you will need the block device.

mfld-fr:
Since I am not clear with this question of mounting either a block or a char device, I temporary hacked `mount_root()` with a `#ifdef CONFIG_ROMFS_FS` in super.c until I can figure a clean solution.

I added `inode` reading and `dentry` lookup, and now the boot fails later on loading the `sbin/init` program. Looks like the file reading operation was not implemented in the earlier ROMFS. Working on it...

jbruchon:
I'm going to wait on merging until you say so. Be sure to resolve merge conflicts with master at the end.

mfld-fr:
@lithoxs @jbruchon : the new ROMFS implementation is now working at least to allow the `ls` command to work (see the updated first comment for testing detail). Could you please review the whole for any mistake against the ELKS core design ?

elks/fs/super.c
```diff
@@ -486,14 +486,22 @@ void mount_root(void)
     struct file *filp;
     int retval;
 
+    /* TODO: allow mounting root from both character and block devices */
+    /* TEMP: hack to allow mounting from /dev/rom */
+#ifdef CONFIG_ROMFS_FS
+    d_inode = new_inode(NULL, S_IFCHR);
```
jbruchon: Shouldn't ROMFS work as a block device? If this is merged and ROMFS is enabled, doesn't it break non-ROMFS roots?

elks/tools/mkromfs.c
```diff
@@ -225,8 +228,25 @@ static int parse_dir (inode_build_t * grand_parent_inode,
 					if (err) break;
 					}
 
+				else if (S_ISCHR (child_stat.st_mode))
```
jbruchon: There is common code in these two `else if` sections; can that code be combined or must it be duplicated? Perhaps a good place to use a `goto`?

elks/tools/mkromfs.c
```diff
@@ -372,6 +381,17 @@ static int compile_dir (int fd, inode_build_t * inode)
 
 			size += count;
 
+			/* Entry inode index */
+
+			count = write (fd, &entry->inode->index, sizeof (word_t));
+			if (count != sizeof (word_t))
+				{
```
jbruchon, line 391: I prefer the starting brace for `if while for` etc. to be on the same line. Only functions should have the brace below.

elks/fs/romfs/romfs-low.S
```diff
@@ -0,0 +1,250 @@
...
+	mov   ax,#1
+	jmp   inode_get_exit
+
+inode_get_next1:
+
+	; Compute inode base from index
+
+	xor   dx,dx
+	mov   cx,[si+8]  ; inode size
+	mul   cx
```
jbruchon, line 82: Is this aligned in some way that would allow a faster left bit shift to be used instead of a multiply?

elks/fs/romfs/romfs-low.S
```diff
@@ -0,0 +1,250 @@
...
+	pop   ds
+	pop   bp
+	ret
+
+
+; Get inode from index
+; int romfs_inode_get (word_t index, byte_t *);
+; TODO: add argument with size of destination buffer
+
+_romfs_inode_get:
```
jbruchon, line 52: Should there be checks in these assembly routines, e.g. a bad value used for indirection doesn't cause us to go outside of the boundaries of the ROMFS area?

elks/tools/mkromfs.c
mfld-fr: Yes, a `goto` simplifies the code. Done for next commit.

elks/fs/romfs/romfs-low.S
mfld-fr: Well, one goal of the ROMFS is to take advantage of the flat nature of the ROM, to avoid wasting space with alignments like other filesystems that divide space in blocks. I only align the files on paragraphs for future in-place execution, as suggested by Juan on the mailing list.

So yes, I agree MUL is a costly operation, but it helps to have a minimal size of the inode, and that cost is balanced by direct access to the filesystem in ROM. I am not sure padding the inode from 6 to 8 bytes and replacing MUL by SHL would improve the overall performance a lot, at the cost of reducing by 25% the available space for inodes in the first data segment (so limited to 64 KB).

elks/fs/super.c
mfld-fr: No, ROMFS does not work neither with a block device nor with buffers, but accesses directly to the data in ROM. The root device `/dev/rom` is a dummy one, just because ELKS requires one to mount a filesystem.

But I agree the change in `mount_root()` is not clean and would break the other use cases. Furthermore, this design does not comply to the Linux standard: https://www.kernel.org/doc/html/v4.11/admin-guide/devices.html

I then propose to remove that hack in `super.c`, to change `/dev/rom` from character to block device, and to allocate the (major,minor) = (31, 0). Do you agree ?

jbruchon: It makes sense to me to make `/dev/rom` a real block device that consists of the contents of the computer's boot ROM and make ROMFS work on that `/dev/rom` block device without any special behavior. That would allow ROMFS to be used on other devices outside of `/dev/rom` if desired for some reason.

mfld-fr: OK, then, let me change `/dev/rom` to a block device. Working on it, will be part of the next commit.

mfld-fr:
I lately spotted a bug while using `cat` to display file content: it goes beyond the end of that file. Looks like I forgot to check EOF in file reading...

Edit: fixed in a next commit.

lithoxs:
I have read partially the changes, which are a lot. I have some remarks:

1. Please try to use assembly only below directory arch/i86. There are 2 places where this breaks: in file main.c, with instruction hlt, and in file fs/romfs/romfs-low.S. The instruction halt were not used so far because some emulators fail to work properly with this instruction, so its inclusion should be a configuration option. You could place the halt in a function in file arch/i86/system.c and place a macro in main.c, but other options exists. The romfs-low.S appears to have memory block copiers, you could use function fmemcpy(). If that function don't fit your needs, you could place your versions in arch/i86/mm/segment.S.

2. In fs/devices.c an array is initialized by hand with all NULL's. It can be more easily accomplished by simply not initializing anything, then the compiler will place the array in BSS, and there will be set to all zeroes at startup. This way you spare a lot of text to read and type. This is how several arrays are initialized, for example the file_array[], in file_table.c.

3. Please be sure to enclose all your changes in #ifdef CONFIG_ROMFS; #endif so we don't have unneeded/unwanted features if we don't have rom (or romfs).
I will continue reading the commits. Please be patient.

elks/fs/romfs/romfs-low.S
lithoxs: If you ever want to use the ia16-elf-gcc compiler, you have to keep in mind the differences in usage of segment register ES. In this compiler, ES is used as scratchpad and can have any value, and its contents must be preserved across function call. This is, you should have to save ES, load it with the kernel data segment and restore its value before return.

mfld-fr:
@lithoxs : thanks for this second review.

1a: I fully agree the halt instruction should be conditional. Moreover, it has nothing to do in this PR as saving the CPU power is another feature, not related to ROMFS. Will be removed in a next commit, and reintroduced in another story.

1b: Fully agree too, as I realized while coding the low-level routines that the same could be performed in C with the existing memory routines in a more generic way, at the cost of a little overhead but without significant impact on the performance. I keep these low-level routines for this step, but I will rewrite that part in the next step (created mfld-fr#26 to track that).

2: Yes, you are right, and I wonder now why I completed straightly the array in place of removing the first initializer to move it from `.data` to `.bss`. Will be fixed in a next commit.

3: Of course. I keep the point, and before requesting the merge, I will also test with ELKS in the default configuration to check that ROMFS is fully optionalized.

elks/fs/romfs/romfs-low.S
mfld-fr: Sure. Indeed I won't rework these routines at all, as @lithoxs highlighted the point that these are actually not needed and will be replaced by more generic code in a next step (see below).

elks/tools/mkromfs.c
mfld-fr: OK, I forgot again to apply the indentation rules of this project, sorry. Will not rewrite all the code, but will pay more attention to this point while editing that code.

elks/fs/romfs/romfs-low.S
lithoxs: The struct inode_operations was modified in last commit. The bmap field is used, with another name. You can leave it with NULL, but the space has to be used. See files in fs/minix.

lithoxs:
I read the commits. I agree with all.

mfld-fr:
Just tested ELKS in default configuration: no problem found at build from a clean working tree or under QEMU. Thanks to all for the review, it helped to mature the changes.

@jbruchon : I think it is now right time to merge this step into the mainstream.

PR #163, opened 2017-10-26
Code reduction by elimination of file fs/block_dev.c.

lithoxs:
Operation of functions in that file and those of
fs/minix/file.c were almost identical, so merging
them was easy. Code size reduced by 272 bytes and
data size increased by 20 bytes. Compiled with BCC,
tested with Qemu.
Changes were indicated by conditional compilation using
the macro USE_GETBLK, so that changes can be easily
evaluated. If accepted, I will submit changes for the final
version.

### DIFF
--- a/elks/fs/block_dev.c
+++ b/elks/fs/block_dev.c
@@ -18,6 +18,7 @@
 
 #include <arch/segment.h>
 #include <arch/system.h>
+#ifndef USE_GETBLK
 
 static int blk_rw(struct inode *inode, register struct file *filp,
 		  char *buf, size_t count, int wr)
@@ -95,4 +96,5 @@ size_t block_write(struct inode *inode, struct file *filp,
     return blk_rw(inode, filp, buf, count, BLOCK_WRITE);
 }
 
+#endif
 #endif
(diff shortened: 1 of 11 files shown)


mfld-fr:
Hello, I just have a question about this change: `fs/block_dev.c` looks like to be generic for all FS, and `fs/minix/file.c` is obviously Minix-specific. Not better to keep generic code and lighten the specific one ?

lithoxs:
On the contrary, fs/block_dev.c is specific for block devices (e.g. /dev/fd0, /dev/hda, /dev/bda2) and cannot be used for any FS. On Linux 2.2 (at least) and below, there was a generic_read() function for use of all FS code and not block devices, but it relies on x386 MMU.
The new code can be used by block devices and all FS, as long as the FS work with 1 Kb block sizes or an integer multiple of 1 Kb. This is not the case of FAT for example, where you can have cluster sizes of 512 bytes. In this regard, the new code is even more generic than before and the additional burden is insignificant.
The differences between code of minix_file_read() and block_read() is the use of minix_getblk() (original line 119) by the former instead of function getblk() (original line 42) by the later.

lithoxs:
I decided to check for inode->i_op->getblk != NULL as a safety measure, but it can be simplified if we require that the getblk field of struct inode_operations not to be null.

mfld-fr:
Thanks for your explanation.

PR #165, opened 2017-11-01
Normalize memory functions

mfld-fr:
This is a preliminary step before next ROMFS revision, to make the room clear. Now, memory functions have the same and consistent API. Also removed unused sister functions.

Sanity checked in default configuration on QEMU.


elks/include/linuxmt/mm.h
```diff
@@ -54,10 +54,9 @@ extern void pokew(__u16,__u16,__u16);
 extern __u8  peekb(__u16,__u16);
 extern __u16 peekw(__u16,__u16);
 
-extern void fmemcpy(__u16,__u16,__u16,__u16,__u16);
-extern void fmemset(char *,__u16,__u16,size_t);
-
-extern void far_memmove(unsigned,unsigned,unsigned,unsigned,unsigned);
+/* memory primitives - count in bytes */
+extern void fmemcpy (word_t dst_off, seg_t dst_seg, word_t src_off, seg_t src_seg, word_t count);
```
jbruchon: I like seeing these header files changed to have the variable names in them instead of a bunch of nameless type specifications.

elks/arch/i86/lib/memmove.c
```diff
@@ -1,29 +0,0 @@
...
- * must be <= 32767 ( ie. fit in an signed int ). That's because   *
- * of ( you guessed it... ) the braindead segmented architecture   *
- * Saku Airila 1996                                                *
- *******************************************************************/
-
-#include <linuxmt/mm.h>
-
-void blt_forth(unsigned,unsigned,unsigned,unsigned,unsigned);
-
-void far_memmove(unsigned sseg, unsigned soff, unsigned dseg, unsigned doff,
```
jbruchon, line 12: I assume that `fmemcpy` already does the same thing as this.

elks/arch/i86/mm/segment.S
```diff
@@ -1,11 +1,88 @@
...
+
+_fmemcpy:
+	push   bp
+	mov    bp,sp
+	push   es
+	push   ds
+	push   di
+	push   si
+	push   cx
+	les    di,[bp+4]           ; arg0+1: far destination pointer
```
jbruchon: Will this conflict with ia16-gcc?

jbruchon (commented):
Let me know when you're ready for actual merging.

elks/arch/i86/lib/memmove.c
mfld-fr: Not exactly, `fmemmove` manages the overlap, not `fmemcpy`. According to the documentation, the first was used in the console for scrolling, but is not used any more, so I removed it from the tree.

elks/arch/i86/mm/segment.S
mfld-fr: This is a portable implementation that works on all the 16-bits compilers (no more assumption on the register usage, except AX for the return value, that is de facto standard).

elks/include/linuxmt/mm.h
mfld-fr: You are welcome.

lithoxs:
I have several remarks.

1. Functions in arch/i86/lib/* are not included in the binary if they are not used. For example, function peekb() is implemented with a macro as (char)peekw() so peekb is not included in the final binary. This also happens with far_memmove and blt_forth and who knows what other functions, but we don't have to worry. In contrast, fstrlen() in user.c has to be commented out with #if 0 because other functions in that file are essential and the linker will put everything in that file in the binary, used or not. In short, it is bad idea to move peek/poke from their present location to segment.S.

2. I like the idea of a standardized ordering in the parameters for this functions. That said, I don't see improvements in the code size of peek/poke functions: differences are +/- 1 bytes compared with the present implementation, so, no gain. And present implementations of this functions are clearly faster. So, I don't see the point. It is easier and better to make some simple modifications to present functions.

3. Present implementation for fmemset is clearly more compact than the new one. The time lengthy part is the same, but the setup and cleanup parts are faster in the present version.

4. The fmemcpy() function has a hopeless situation. Aligned transfers can speed up the move operation significantly. But the source address can be misaligned, independently the destination address can be misaligned. To complicate matters, the number of bytes to transfer can be odd. I see that the common solution to this problem is simply to use the "rep movsb" instructions because the setup and cleanup code will have an unreasonably large size. In case I would be willing to pay the price, I would check the alignment of the source address (or destination) so that I guarantee the alignment of at least one block. This will lead to a byte transfer at the beginning, or the end or both. But you check if bytes to transfer is odd, so you don't guarantee anything.

mfld-fr:
@lithoxs : you are right for the point 4. I was focused on my x88 clone, and I forgot the misalignement cost on a x86 with a 16 bits bus. So let me rollback this optimization try. Rather than a complex routine that tries to realign, I think the simplest is to provide two versions, one with byte transfert, the other with word transfert. Up then to the caller to use the best one with the knowledge of the actually used structures.

About the point 1, I had the feeling that the grouped routines were used in almost all configurations, but I would more rely on your experience. Let me then move back these ones into the library.

I will discuss the other points later, because they are not so obvious when balancing the performance and the code portability.

mfld-fr:
@lithoxs : now about the remaining points. I agree the rewritten functions are slower with bigger prologue & epilogue, but they are now compiler-agnostic. That's the key point.

Previous versions were doing some assumptions about the registers usage by the caller. So the bit tricky `#ifdef USE_IA16`. IMHO, having to modify many low level routines each time we give a try to another compiler, just to save some CPU cycles, does not look a good strategy.

If one day we move from that silly BCC to a compiler with a really effective registry allocation, one would have to write these routines as proposed.

If the overhead of such prologues & epilogues becomes significant in a particular use case, it probably means that these routines are too often called and should rather be inlined in the caller. According to the history and the source, this was the story around the `fstrnlen` previously based on `peekb`.

My last argument would be that, having this set of memory primitives, we could later clean up other memory functions and make them more generic. Today, most of the `segment.S` content could be written in C, and some functions in `user.c` could be optimized. This gain would balance the loss in primitives.

lithoxs:
Whole kernel, including old peek, fmemset, etc. already run correctly when compiled with BCC and ia16-elf-gcc. In addition, as far as I know, they comply with the rules set forth by OpenWatcom for the 8086 target. From unrelated experiences, I guarantee that this functions will properly work with Microsoft C if we declare the C prototypes as _cdecl and remove the preprocessor statement in segment.S 

The #ifdef USE_IA16 in segment.S functions is an optimization for BCC, producing the elimination of a few instructions, not to circumvent some weird behavior. If we remove the #ifdef's and leave the  instructions, it will also work with BCC and ia16-elf-gcc, and I am quite sure also OpenWatcom.

Sadly, I don't have expectations on the emergence of new compilers for this platform to improve the produced code. Indeed, I have many times mentioned in the list that I have used ia-unknown-gcc and ia-elf-gcc to produce a working kernel, and nobody has shown any interest. So, expect the continued use of BCC.

jbruchon:
I'd be happy to add the ia16-gcc code to dev86.

lithoxs:
The repeated string instructions for the 8086/8088 do have small gains when using the sixteen versions over the 8 bit versions, even with operators misaligned. 

However, starting with the 80186/80188 that small gain vanished because once the instruction is started, all time is spent only in memory access. This is, on a 80186 a sequence "rep stosb"  and "rep stosw" to set a block of the same size to some value, will take the same time if operator is misaligned, regardless that stosw makes half iterations.

So, operator alignment has greater impact than choosing 16 bit string over 8 bit string. For example, on fmemset, if adddress is aligned and the count is odd, the setup will misalign the address and the whole operation will take more time on an 80186 (and above) than if simply use stosb, because of the setup time.

So, the setup must be effective (which is not in new version), and must take short time. Even if you succeed to meet this conditions, the additional work might make things worse. For example, if you are going to move a few bytes (as the typical use of memcpy_fromfs/tofs) it is faster to simply use movsb than check alignment, taking jumps (which forces to flush the prefetch queue), complete the transfer with additional movsb, to save fewer periods than those used in the complex setup because of the small size of the block to move.

It might be simpler to use movsb.

jbruchon:
`config` already allows CPU selection; why not change behavior based on the chosen target CPU?

mfld-fr:
@lithoxs : Thank you Juan for your deep explanations. I understand two things:

1- It is up to the caller of the memory functions to align the data and to select between byte or word transfer for the best performance. That functions should only perform linear execution for optimal prefetching.

2- If using BCC, there is no need to save the AX,BX,CX,DX and ES registers on function entry. Same if using GCC-IA16, except for ES register. Same as GCC-IA16 for any other 16 bits compiler you know.

Am I correct ?

For the point 2, I feel uncomfortable because if I call these routines from another assembly routine where I would optimize the registry allocation, then I must not forget to save the registers before the call (potential bug). But as one overall goal is to minimize the assembly code, this is an unlikely use case, so a minor concern.

About the GCC-IA16 topic, I opened an issue for that (jbruchon#166). I am not only interested, but I **DO** expect we would drop BCC and switch to GCC for ELKS mainstream, for its better C parsing and better generated code.

@jbruchon : GCC-IA16 code is already archived on GitHub. Please see the mentioned issue. Or maybe you meant porting the DEV86 libraries to that compiler for ELKS user space ?

lithoxs:
For the point 1, it may not always be up to the caller to have have data aligned in some way. For example, reading from a file, the kernel has no have way of influence the alignment of the user buffer. Nor the application programmer, because that is hidden below the library. However, may be we can do more. For example, the L2 cache is aligned, but for the L1 cache it is left up to the linker (am I correct?). Maybe we can find other examples. But in the general case the caller don't know the alignment. 

For the point 2, you may modify the contents of ax, bx, cx and dx on all compilers. For BCC, registers BP, SI and DI must be preserved, that is, if you modify any of them, must save on entry and restore on exit. It is assumed that ES == DS. For IA16-GCC, registers BP, SI and DI must be preserved, but in general ES has a value ES != DS. For OpenWatcom, I'm not sure what is the rule for BP, SI and DI, but it  must work if you save them when you modify them.
I understand you point when programming in assembly. When I did that, I liked to leave the burden of saving registers to the called function instead of the caller, and this produced very good code. But our main concern must be calling assembly code from C.

mfld-fr:
@lithoxs : Based on what we discussed, I reverted the full prologues & epilogues in the memory functions, and I tried to improved some. Could you please review these latest versions and tell me if you agree ?

lithoxs:
1. Agree with new peek?/poke?.
2. At the end of fmemcmpb/w you have:
#ifdef USE_IA16
￼ 	mov    es,bx
￼ #else
￼ 	mov    es,cx
￼ #endif

The mov es,cx is a bug. 
3. Original function fs_memcmp is only used at two places and compares strings. So, I don't foresee any use for fmemcmpw. But being in arch/i86/lib, its presence is not harmful because will never be included.
4. As with fmemcmp, function fmemsetb/w are mostly used in situations where the size of the block is not guaranteed to be even (or odd). May be is better to leave the original.
5. Function strlen_fromfs was used only in 2 places.  The new version is larger and leaves to the caller the burden of setting the user segment. May be is better to leave the original.
6. function memcmp is not used anywhere. As with fmemcmpw, being in arch/i86/lib, its presence is harmless.
7. Regarding files arch/i86/lib/string.S and lib/string.c, Check that the remaning functions there are really used because they will be included in the final binary. Also check include/arch/string.h because there are defined the string function that will not be comliled in string.c.
8. For the fmemcpy function, I think that the best compromise is to have a generic function using rep movsb and a complex one, to be used when we are sure that the block to transfer is large enough to justify the setup and cleanup cost, as in the buffer cache.
9. For the case of memcpy_fromfs/tofs, I think the best is to keep them as now, for speed reasons.

mfld-fr:
2. Are you sure ? If using BCC, ES is forgotten in line 17 by LES, and is restored as for DS in line 37 assuming that DS=ES=SS.
3. I need `fmemcmpw` for the next revision of ROMFS to check the header in ROM, that is aligned. Also for the next revision of `/dev/rom0` to check block in Flash after reprogramming it.
4. I need `fmemsetw` for upcoming RAMFS / TMPFS to erase a segment in RAM, also aligned.
5. I did not changed strlen_fromfs. I just added a TODO comment for my automatic tracking in Eclipse. Removed it.
6. Not sure I will use it in ROMFS. Until I decide, let us keep it for now, as no cost.
7. Checked. The C version of `memcmp` was already commented out in `string.c`.
8. I will use `fmemcpyw` for RAMFS / TMPFS to compact filesystem in RAM, again aligned. I am not interested right now in a smart `fmemcpy`, as I am not using any buffer in ROMFS / RAMFS / TMPFS, but we could work on that later.
9. OK, let us keep it for now, but "I will come back" on these 😈 .

lithoxs:
What is the use of functions _seg_?s in segment.S?. There are the variables kernel_cs and kernel_ds with self-explanatory contents. If needed by ROMFS then they should be optionalized

mfld-fr:
Yes, you are right. Completely useless. Just removed them and squashed the previous commit.

mfld-fr:
@jbruchon : I think the review is now over, as we discussed all the points. Could you please merge this PR ?

PR #167, opened 2017-11-05
Close mfld-fr#26 : no assembly in romfs

mfld-fr:
Replaced assembly low-level routines in ROMFS by memory primitives and little more C lines.
Also little cleanup in QEMU script and more optionalization in `fs.h`.
Tested in ROM configuration on EMU86.


lithoxs:
I'm affraid the ROMFS has not been properly optionalized. There is a general increase in memory usage:
The following compilation results are for the default configuration, except for no CONFIG_APM and no CONFIG_ETH. The default configuration is no CONFIG_ROMFS_FS.

BEFORE ROMFS patch:
System is 106544 B (48976 B code, 6440 B data and 51128 B bss)

AFTER ROMFS patch:
System is 106764 B (49008 B code, 6448 B data and 51308 B bss)

I understand that normalization of memory function can lead to some inefficiencies. But, what is the reason for the increase in data and bss?

mfld-fr:
Mmm... Maybe a diff between system.map files could help to spot the change ? I am first thinking about additional standard entries in the character memory devices, not related to ROMFS, that we could cleanup.

lithoxs:
Do you mean struct file_operations and struct inode_operations? They are initialized and will be in data. But the largest increase is in bss.

mfld-fr:
No, I mean the array of character memory devices, and the array of block devices. For the latest, we should optionalize the maximum of major number. Less without /dev/rom, and at least 31 (the Linux standard) with.

mfld-fr:
Tracked by jbruchon#169. Please continue the discussion in the issue, no more in this closed PR.

PR #168, opened 2017-11-06
Restore compilability using ia16-elf-gcc compiler

lithoxs:
Includes modification in Makefiles-rules to use ia16-elf-gcc instead of ia16-unknown-elks-gcc.
The kernel image is successfully built, fails to mount the root filesystem.

### DIFF
--- a/elks/kernel/printk.c
+++ b/elks/kernel/printk.c
@@ -30,6 +30,12 @@
 #include <linuxmt/mm.h>
 #include <stdarg.h>
 
+#ifdef __BCC__
+#define REGISTER register
+#else
+#define REGISTER
+#endif
+
 /*
  *	Just to make it work for now
  */
@@ -77,14 +83,14 @@ static void numout(__u32 v, int width, int base, int useSign,
     int c, vch;
     register char *i;
 
-    i = 10;
+    i = (char *)10;
     dvr = 0x3B9ACA00L;
     if (base > 10) {
-	i = 8;
+	i = (char *)8;
 	dvr = 0x10000000L;
     }
     if (base < 10) {
-	i = 11;
+	i = (char *)11;
 	dvr = 0x40000000L;
     }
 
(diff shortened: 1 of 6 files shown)


lithoxs:
Kernel compiled with ia16-unknown-elks-gcc now runs.
The default compiler still is BCC. You have to uncomment USEIA16 in Makefiles-rules to use ia16-elf-gcc or ia16-unknown-elks-gcc.
Kernel compiled using ia16-unknown-elks-gcc works fine.
Kernel compiled using ia16-elf-gcc hang after loading the init program.

I think is that ia16-elf-gcc optimizes too much. For example, function delay() in init/main.c is optimized to just the setup and cleanup of the stack frame, with nothing in between. But not smart enough to compile delay to a simple ret instruction.
There are places where variables should be declared volatile. Otherwise, the compiler will optimize assigning them to a register and will never be modified by interrupts.

elks/kernel/printk.c
```diff
@@ -77,14 +83,14 @@ static void numout(__u32 v, int width, int base, int useSign,
     int c, vch;
     register char *i;
 
-    i = 10;
+    i = (char *)10;
```
mfld-fr, line 86: I don't agree the fix: `i` was declared as a pointer, and always used as a scalar. I would rather declare it as a scalar and remove all these casts.

lithoxs: This is a BCC-ism, and early developers used this heavily to optimize code. I'm affraid we will have to live with this as long as we don't change to other compiler because otherwise the code size will grow hugely. So, I use it too.
The only way that BCC honors the register directive is when applied to pointers. This is the reason for declaring a pointer to a variable used as integer.

lithoxs: By the way, this is not the fix to the compilation problem. I put that change to reduce the noise.

PR #170, opened 2017-11-07
Restore functionality when compiled with ia16-elf-gcc compiler.

lithoxs:
The problem was that some assembly files with extension .s needs to be passed by the preprocessor, and the preprocessor is applied only to assembly files with extension .S (uppercase). The solution was to rename those file, plus other minor fixes.
Note that there are more assembly files with this problem when the CONFIG_ROMFS_FS is enabled.
Also note that the default compiler is ia16-elf-gcc (not BCC). If disagree, let me know to revert this.


jbruchon:
It'd be nice if `build.sh` detected the availability of `bcc` or `ia16-elf-gcc` and set the appropriate exports; that would take some of the guesswork out for anyone new to the project.

Issue #166, opened 2017-11-03
GCC-IA16 guidelines


mfld-fr:
@lithoxs : earlier this year, you announced that you started to compile the ELKS kernel with the latest GCC-IA6 (http://www.spinics.net/lists/linux-8086/msg00817.html).

As I can see in your PR flow, you still maintain the build with GCC-IA16, and it looks like, after tons of (more or less serious) posts to discuss that topic, that this compiler is today the best alternative to BCC.

Not only because of the work you already performed, but also because this GCC fork is still alive & able to be mainstreamed: https://github.com/tkchia/gcc-ia16

Even Alan is keeping an eye on this option: https://github.com/crtc-demos/gcc-ia16/issues/4

It would be nice if you could provide some guidelines to use that compiler in the /Documentation folder, so that we could be more to use it to improve the ELKS code.

lithoxs:
There are two compilers for 8086 based on GCC:
 ia16-unknown-elks-gcc, the original compiler by Rask, with a few improvements by me, and
 ia-elf-gcc, based on the work by Rask, with many improvements.

For the first, I posted here http://www.spinics.net/lists/linux-8086/msg00674.html instructions to build the compiler. It tested the building in Debian squezze 32 bits and Ubuntu 13.1 64 bits, producing a working cross-compiler. Trying to build in Debian jessie 64 bits fails.

This compiler can build applications for ELKS, but only some applications work well. The C library is based in newlib and is quite bloaty, and the linker likes to place everything in a library to the final binary. So, applications size are much larger than those produced by BCC.

The kernel can be built with this compiler. To avoid rewriting everything in assembly, I wrote an utility to convert ATT 8086 assembly to AS86, and rewrite the rule for compilation: first compiles with the S option, producing an assembly file, then convert it to as86 with the above utility, assemble it with AS86, and remove the garbage. Everything is already in the Makefile, but you need the conversion utility. It was necessary to write one function of libgcc.

The second compiler is newer, faster and produces smaller code. It also relies on newlib and has the same linker as before. So, the problems of first compiler in this regard are the same. As distributed, the compiler cannot produce ELKS binaries. To build the kernel, the Makefile needs further changes, incompatibles with  ia16-unknown-elks-gcc.

lithoxs:
INSTRUCTIONS TO COMPILE THE KERNEL USING ia16-unknown-elks-gcc.

[...]

PR #171, opened 2017-11-07
Move to recent GCC-IA16

mfld-fr:
GCC-IA16 sourced from GitHub and built as per @tkchia advice (see tkchia/gcc-ia16#1).

### DIFF
--- a/gcc/.gitignore
+++ b/gcc/.gitignore
@@ -1,0 +1,16 @@
+att2as86
+
+*.zip
+
+binutils-ia16/
+binutils.src
+binutils-build/
+binutils.build
+
+gcc-ia16/
+gcc.src
+gcc-build/
+gcc.build
+
+install/
+
--- a/gcc/Makefile
+++ b/gcc/Makefile
@@ -1,0 +1,51 @@
+.PHONY: all clean
+
+
+# ATT to AS86 conversion utility
+
+CFLAGS=-O2 -ansi -Wpedantic -Wall -Wextra
+
+all:: att2as86
+
+clean::
+	rm att2as86
+
+
+# BINUTILS and GCC for IA16
+# PREFIX must be set in environment
+
+BINUTILS-VER=c045395ece435fcb3bf6225e9c4734b3ebcc874b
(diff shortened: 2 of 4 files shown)


mfld-fr:
@jbruchon : GCC-IA16 build OK. I need to resynch on latest Juan's PR, so please merge this one (no impact on ELKS).

lithoxs:
Should this be placed in the ELKS tree?. Even now, the BCC source tree is put besides the ELKS tree. Perhaps the conversion utility could be placed in arch/i86/tools. The binutils package is not used in any way now.
But the most important points are some questions I asked in issue #166. Looks like we are moving without any goal. Right now, all we have are some tricks to fake the build system to think it is still using BCC.

jbruchon:
Am I correct that gcc doesn't build a working system yet?

lithoxs:
No. The kernel built using any of ia16-unknown-elks-gcc or ia16-elf-gcc are able to boot, bring up the shell, run commands without crashing, etc. We need to make tests with diverse configurations and load levels to be sure it is fully working.
But I would say that this compilers do produce working systems.

mfld-fr:
Today, for both BCC and GCC-IA16, user has to build manually the compiler, then link it to ELKS, and tune the Makefile globals. A first goal is build chain automation. I agree the built chain should be rather placed out of the tree, so I will move it later. But the automation scripts themselves have to be in the tree to have a common reference.

mfld-fr:
The conversion utility is part of the build chain, not of ELKS, so it looks more logical to place it there.

mfld-fr:
For the overall discussion about compiler, please continue the thread in jbruchon#166, not sparsely in many PR. Thanks.

jbruchon:
Should I go ahead and merge this?

mfld-fr:
Just sanity tested this recent GCC-IA16 version with default configuration and QEMU, no problem at a first glance. Please merge this one, it would set a common reference, even if everybody is free to use his own GCC-IA16 version. I will tune later this GCC-IA16 build & installation script after fixing more urgent regressions in non-default configuration.

PR #174, opened 2017-11-09
Tune build with GCC

mfld-fr:
Various fixes to extend GCC build:
* with ROMFS option
* with BIOS console option

Tested on EMU86 with ROM only configuration.


mfld-fr:
@jbruchon : ready to merge 😄

PR #191, opened 2017-11-30
Build using local /dev86

mfld-fr:
Today ELKS can be built only with an installed `dev86`. The goal is to build ELKS with a local `dev86`, to address one Jody's concern (see jbruchon#188). This PR fixes some ELKS Makefiles, and another PR (see jbruchon/dev86#11) fixes some `dev86` ones. Minix boot blocks have also been moved from `dev86` to ELKS.


elkscmd/levee/Makefile
```diff
@@ -12,7 +12,7 @@ include $(BASEDIR)/Make.rules
 
 ###############################################################################
 
-CFLAGS = -ansi -O -0 -DSIZE=16000L
+LOCALFLAGS=-DSIZE=16000L
 
 xCC=gcc
```
jbruchon: Shell variables must be in all caps.

mfld-fr: These `x*` are legacy and unused ones. Cleanup done for the next commit.

mfld-fr:
User is not required to install `dev86` on his machine any more. Now ELKS uses the local cross build chain (in folder `cross/`) from the main `build.sh`. I temporary disabled the previous build of GCC-IA16 to focus on the default configuration with BCC.

For the cross build chain folder, user is still free either to create it outside the ELKS tree and link it (as `dev86` before), or to create it directly inside the ELKS tree. He is also free either to build the cross chain himself, or to use the script `tools/build.sh` to automatically download and build it.

What remains to do next:
- Move the configuration scripts from kernel to top level
- Select the compiler at top level for both the kernel and the cross tools
- Enable the build of GCC-IA16 if it is selected
- Move the C library from DEV86 to ELKS
- Use GCC-IA16 to compile that C library
- Use GCC-IA16 to compile the applets

mfld-fr:
This PR has now a quite large number of modified files, so I think I should pause the development at this point and do the next steps later. Could you please review this one, and merge if you agree ? Thanks.

jbruchon:
What else needs to be done? If I merge this now, will anything break?

mfld-fr:
I stopped at a point where nothing should be broken. I tested the whole workflow from a fresh ELKS tree and an empty `cross/` folder, up to running the build result in QEMU, with the default configuration. All looks fine, except the boot where `ktcp` and `telnetd` are started without network support enabled in the kernel, and a FAT volume is tried to be mounted. These are minor errors already known before that PR.

I think we would have to revisit that boot sequence after moving the configuration script at top level: if user could select the applets with a config / menuconfig, then we could definitively setup the boot with the actually installed ones.

PR #194, opened 2018-04-14
Entend configuration to the whole source tree

mfld-fr:
To make the user land as configurable as the kernel, move the configuration script and entry to the top directory. Also remove the unused Tcl/Tk front-end.


jbruchon:
I'll merge this as soon as more people try it out and confirm it works OK with them.

mfld-fr:
Did you try it yourself ? Do you see some more improvement opportunities ?

mfld-fr:
Looks like we are only two to be active these days... did you give a try to that new configuration layout ?

jbruchon:
Unfortunately I am in quite a bit of arm pain due to Guillain-Barré and can't type a whole lot right now. Works fine for you?

mfld-fr:
Oh, so bad... I wish you then a complete recovery. But I know a GBS is quite long to recover, so I will avoid to stress you in the next months. Do you want me to help as a collaborator on this project to avoid disturbing you in the meantime ?

It looks like to work fine here. I checked the complete cleanup, the full build with the default configuration, and ran the full3 image in QEmu. There are still some standalone scripts from @georgp24, but they would be superseded when making the applets and the images also configurable.

jbruchon:
I lucked up and have the "best case" kind of GBS. Should be better in weeks rather than months. I reviewed the work and I am OK with it. I'll pull this now and if others have problems we'll know soon.

mfld-fr:
OK, thanks for the review and the merge !

PR #197, opened 2018-04-21
Optionalize commands and image format

mfld-fr:
This is the implementation of the design change proposed on the mailing list:
https://www.spinics.net/lists/linux-8086/msg00856.html

The main goal is to allow the user to customize the set of commands he wants to include in the target image without hacking the Makefiles.

Another goal is to prepare the move of the currently orphan scripts that build some FAT images back into the main configuration and build process.


mfld-fr:
Test result: MINIX images on floppy media are working well on QEMU, so there is no functional regression.

Indeed there is still work to do for a comprehensive implementation:
- ROM filesystem: see #198
- hard-disk medium: see #199
- FAT filesystem: see #200

PR #201, opened 2018-04-25
Add link support to ROM filesystem

### LINKED ISSUE #198: [romfs] Add support of symbolic link
From review and testing of PR #197:

The ROM filesystem currently does not support the symbolic link, resulting in an image build failure when packing the `target` folder with the `ROM` configuration option. In that folder, `/bin/sh` is a link to either `ash` or `sash`, depending on the configuration.

mfld-fr:
Closes issue #198.

### DIFF
--- a/elks/fs/romfs/romfs.c
+++ b/elks/fs/romfs/romfs.c
@@ -194,6 +194,11 @@ static int romfs_followlink (struct inode * dir, register struct inode * inode,
 		pds = &current->t_regs.ds;
 		user_ds = *pds;
 		*pds = inode->u.romfs.seg;
+
+		/* Trick: mkromfs appends null character to link data
+		 * to protect against null-terminated string comparison
+		 */
+
 		err = open_namei (0, flag, mode, res_inode, dir);
 		*pds = user_ds;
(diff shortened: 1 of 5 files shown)


elks/fs/romfs/romfs.c
```diff
@@ -163,130 +163,103 @@ static int romfs_lookup (struct inode * dir, char * name, size_t len1,
...
+	 */
+
+	if (!S_ISLNK (inode->i_mode)) {
+		*res_inode = inode;
+		err = 0;
+	} else {
+		pds = &current->t_regs.ds;
+		user_ds = *pds;
+		*pds = inode->u.romfs.seg;
+		err = open_namei (0, flag, mode, res_inode, dir);
```
mfld-fr: Ooop's ! The function `open_namei` expects a null-terminated string, while the link content is not... failure opportunity here. It can work only because of the paragraph alignment in ROMFS, and will fail in case the length of the link is multiple of 16.

mfld-fr: Tricky fixed by d5d70a7a.

PR #215, opened 2018-05-16
Bunch of non-critical commits

### LINKED ISSUE #207: 386 config fails to build
Setting processor type to 386 in the config cause the build to fail in bootsect.s (as86 -3):

	00557                                           	mov	sectors, ax
	***** illegal immediate mode...........................................^
	***** mismatched size………………………………………………^
	… and so on …

### LINKED ISSUE #213: FD arrays not reset in select() after timeout
See issue #133 reported by @Mellvik for an example : `ktcp` iterates quickly on `select()` despite the 1 second timeout, and moreover, the FD arrays are not reset upon a timeout, causing useless reads.

mfld-fr:
- add a precursor of the early `printk()` (normal one is not available until console initialization)
- more intuitive configuration layout for hardware description and kernel options
- remove support of 80386 and later (fixes #207)
- move documentation folder to top one
- select() behaviour compliant to POSIX after a timeout (fixes #213)


mfld-fr:
The stacked changes begin to be numerous, it is time to push them into master.

PR #220, opened 2018-06-11
[kernel] Move to full GNU toolchain

mfld-fr:
This is the next step in the compiler / toolchain migration, as discussed in #166.
Special thanks to @tkchia for his support on GCC-IA16 and related stuff.

Contains:
- Rewrite some assembly sources to ATT syntax
- Tune GCC options: small model, no relocations...
- Add LD scripts to output ELKS a.out
- Rework build scripts for full GNU
- Add RAW option for target image
- Warnings and glitches fixes
- Wipe out some BCC related stuff

Status:
- Tested under QEMU
- Tested in default and network configurations
- Kernel boots with both RAW and MINIX target images
- Successful login and walking the MINIX FS
- Successful telnetd


elks/arch/i86/mm/segment.S
```diff
@@ -4,85 +4,87 @@
...
-	mov	si,4[si]
+memcpy_tofs:
+	mov	%si,%ax
+	mov	%di,%dx
+	mov	%es,%bx
+	mov	current,%si
+	mov	TASK_USER_DS(%si),%es
+	mov	%sp,%si
+	mov	2(%si),%di
+	mov	6(%si),%cs
```
mfld-fr: Wrong register here ! Replace CS by right CX. Fixed in 248ae46.

mfld-fr:
First draft was compiled with default global optimization flag `-O0`:
`System is 116589 B (64224 B code, 5758 B data and 46607 B bss)`

Sizes with `-Os` flag in second draft:
`System is 91853 B (39856 B code, 5392 B data and 46605 B bss)`

PR #227, opened 2018-12-02
Fix: add needed permissions to files when populating disk image

tkchia:


mfld-fr:
Thank you for that fix !

ghaerr (commented):
@tkchia: Can you comment on whether this commit is still needed? I would like to delete it. The build process now use `mfs` and `mcopy` which no longer use the current user's UID/GID when creating filesystems. In addition, the filesystem is never mounted by the host OS kernel thus eliminating any problem of later having a `cd /dev` failure as indicated in your original commit.

PR #228, opened 2018-12-02
Fix: some build errors when using `config_emu86` settings

tkchia:


mfld-fr:
Oh, so I did not push my local fixes yet... sorry !

PR #229, opened 2019-01-19
Switch to newer version of binutils-ia16, and slightly fix documentation

tkchia:


mfld-fr:
Problem with old `binutils-ia16` reported in https://github.com/tkchia/binutils-ia16/issues/2

PR #231, opened 2019-02-16
Improve execution under EMU86

tkchia:


mfld-fr:
@tkchia : thank you for the improvement !

elks/arch/i86/boot/setup.S
```diff
@@ -451,7 +451,10 @@ size_ok:
 	int     $3               // break for debugger just before kernel
 #endif
 
-	ljmp     $SYSSEG+2,$0
+	mov	$SYSSEG+2,%bp
+	pushw	%bp
+	pushw	0x18	//; entry point
```
mfld-fr, line 456: **BUG:** At this time, the current data segment is no more the kernel executable header, but the INITSEG set in previous line 448. So the next RETF fails.

mfld-fr: Fixed in 46f929803270fa2cf2351f2e96d0cd19f0799efc

Issue #245, opened 2019-04-02
Mac OS X cross compilation build


ghaerr:
Hi,

Relating to my comments to getting nano-X running again in the "Nano-X running?" thread, I pulled down both gcc-ia16 and elks, and started the process of building on my Mac OS X system. After a bit, I got tools/build.sh to work, which required installating wget, and the gnu mpc, mpfr, and gmp libraries. The cross-compiler built and runs on OS X :)

Getting ELKS built is a problem, because OS X is running a case-insensitive file system.
This means that, say, setjmp.S and setjmp.s are the same file. One is overwritten by the next in the git clone. More of a problem is that elks/elks/Makefile-rules runs the C preprocessor "gcc -E" in order to handle #if's in .S files, which then writes the output to the same-named .s file, which overwrites the source on OS X. In addition, the make clean runs "rm -f" on the .s intermediate file, which also removes the original .S file.

I could fix all this, but wanted to ask what you feel a better source extension for c-preprocessing-required source files should be? Instead of .S, how about .ss, or perhaps .scpp or .Scpp? 

The fix would require that all .S files get renamed to .scpp, and the Makefiles modified with the renamed source files. Another option would be to keep .S and have the intermediate .s file be named something different, but this would be even stranger since bin/as accepts .s as default.

Finally, elks/libc/system/syscall.awk line 31 has an awk printf moving register %ax,
this needs to be changed to %%ax or awk fails with a bad printf.

I realize that ELKS was built for Linux, but given that the system is run and tested under QEMU, I thought that building for OS X would be useful. I regularly build multiple targets for Microwindows/Nano-X on OS X and would like to take a pass at this on ELKS.

Mellvik:
Gregory, 

The first thing to do is to change the setup of your Mac file system, to become case sensitive - as any ***x file system should be.

Then some practical advice: I gave up on ELKS and OSX almost immediately. You don’t need that pain when a Raspberry Pi will set you back only a few $ - and will make life a lot easier for you.

My 2c...

Helge

mfld-fr:
@ghaerr : I think you are right about the file extensions. As '.s' is the *de facto* standard for ASM files in the GNU world, I would change '.S' into something else like '.sc' to mean "ASM to be CPP", and be able to build on case-insensitive systems.

ghaerr:
I found a better way to solve the problem, after seeing the the gcc-ia16 build itself uses .S files:

[...]

Issue #242, opened 2019-04-01
Nano-X running?


ghaerr:
Hello everyone!

I just happened by ELKS and am glad to see there's a lot going on lately!

I noticed that dev86/bcc has been recently replaced by gcc-ia16 and that the Nano-X make still uses bcc. Has anyone tried compiling it up? Is there any interest in making it work on the current ELKS?

Also, elks/nano-X/LICENSE states that there wasn't a license file provided, but actually there was, buried in elks/nano-X/src/misc/LICENSE, which should be moved to replace the former.

Let me know if you need any help; I'm reading that the gcc-ia16 compiler generates smaller code, and Microwindows v0.86 was the last that was made for ELKS (20 years ago!) because it was getting too close to going over 64K when more features were being planned.

Greg

mfld-fr:
Hello @ghaerr,

Thanks for the clue about the license file, it will be fixed as you instructed.

About Nano-X in the context of ELKS: it has not been moved to GCC-IA16 yet, not because lack of interest, but because priority is today to stabilize the kernel and the core utilities on that new build chain.

So any help to complete the move for remaining parts like Nano-X is welcome !

ghaerr:
Thanks @mfld-fr. I'll pull the project down and build GCC-IA16 on my OS X system, and test it under QEMU. Any known issues on OSX cross-compilation?

I think there may be a .s file or two for the Nano-X drivers. Is the calling convention identical from bcc to gcc, along with the register use (save si/di, scratch ax/bx/cx/dx)? Does gcc support #asm (probably not).

Is there a particular video card/setup which should be supported (EGA/VGA 4 planes, 16 colors)?

mfld-fr:
The GCC-IA16 calling convention is quite the same as BCC, except that you must save ES and don't use it for other things than segmentation.

GCC has a better support of the 'asm' statement in a 'C' context than BCC, see:
https://gcc.gnu.org/onlinedocs/gcc-6.5.0/gcc/Using-Assembly-Language-with-C.html#Using-Assembly-Language-with-C.

No special requirement about the video HW, testing can be performed in QEMU in first intent with emulated VGA. If somebody needs to run it on an antic CGA or EGA, then he would have to request it through an issue :smile:.

mfld-fr:
LICENSE issue fixed in 7d0920b

ghaerr:
[...]

PR #251, opened 2019-04-12
Nano-X and OSX Compile

ghaerr:
These three commits implement more changes required for the OSX compile, and completion of the port of Nano-X to gcc-ia16.

The first OSX commit adds more .s/.S fixes. The second is dubious in that it eliminates a seemingly unneeded header in build.c (sys/macros.h not available on OSX), but also fixes the OSX libc build race error discussed separately by inserting a sleep 1 in libc/Makefile. This second commit may want to be done differently.

The third commit is the port of Nano-X to gcc-ia16.


ghaerr:
We still need the fix for the .s/.S files being removed when "make clean", discussed earlier. That fix is not include in this pull request. With this pull request, the make on OSX compiles all files, but still fails in the kernel link. Nano-X demos are compiled and linked.

mfld-fr:
Nice contribution ! Waiting the next week-end to have some time to play with it...

mfld-fr:
Implements partially issues #242 and #245.

PR #255, opened 2019-04-17
Last modifications required to fully build ELKS under OS X

ghaerr:
This patch allows ELKS kernel and cmds to be built with gcc-ia16 on OS X.
The following changes are made:

1) ia16-elf-ar must be used for all archives, OS X ar is broken and ia16-elf-ld can't read some .o symbol entries in resulting .a files.

2) Remove .s deletion in "make clean".

3) Remove .s: .S  and .o: .s rules in Makefiles (all unnecessary anyways).

4) Add test graphics adapter code to lib/vga (currently unused) and fix nano-X demo hang on startup.

5) Kluge fix for continuing libc.a build problem where race results in not all .o files added to archive: "sleep 1" added if $(shell uname) == "Darwin".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 # IDE local settings
 #
 .project
+.DS_Store
 
 #
 # Build ignores
--- a/elks/Makefile-rules
+++ b/elks/Makefile-rules
@@ -260,13 +260,13 @@ elks:
 
 doclean: $(CLEANDEP)
 	rm -fv *~ *.a *.bak *.lint *.map *.o *.tmp core $(CLEANME)
-	@for FILE in *.s ; do \
-		BASE=`basename "$$FILE" .s` ; \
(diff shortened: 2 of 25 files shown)


mfld-fr:
Implements more #242 and #245

tkchia:

The method of getting several independent subdirectories to update the same `libc.a` is rather flaky in general, I think.

As part of my effort to add support for building `elks-libc` [multilibs](https://www.airs.com/ian/configure/configure_8.html), I am working on a patch to get the `libc/` subtree to use a saner system for building up `libc.a`.  I hope to put up a push request for this soon.

Thank you!

mfld-fr:
Nice work. I missed the fact that we were still using the host program for the archiver and not the cross one, so the linking problem on OSX. Thanks to have fixed that dirty part and for more cleanup in the Makefiles !

Issue #252, opened 2019-04-14
Integrating elks-libc installation more closely with gcc-ia16 toolchain?


tkchia:
Hello all,

@ghaerr has [complained](https://github.com/jbruchon/elks/issues/242#issuecomment-481756600) --- and I quite agree --- that compiling a program for ELKS using the `gcc-ia16` toolchain is more complicated than it needs to be.

Yesterday I had made some tweaks to the [Newlib](https://github.com/tkchia/newlib-ia16/commit/14af178d4403d018db6a2bba96226568890ee3c4) and [GCC](https://github.com/tkchia/gcc-ia16/commit/dbbadd1b237963a87c58d2dab97f4d627cfac4a4) components of the toolchain, so that there is now a semi-working setup for compiling ELKS programs by just saying `ia16-elf-gcc -melks` ...&nbsp;.  However, this setup is based on Newlib, which [is quite bloated](https://github.com/jbruchon/elks/issues/166#issuecomment-341820826) and may not be the best choice of C runtime.

Therefore I hope to find a way to somehow integrate ELKS's own `libc` into the `gcc-ia16` toolchain installation.  Ideally, one can easily ask to link a program against `elks-libc` as an alternative to Newlib, or even as a default replacement for Newlib.

For secondary goals, I hope
  * ~~to add multilib support to `elks-libc` (automatically build and use different versions of `elks-libc` for different function call conventions, etc.), which Newlib already has; and~~ [update 20190623: OK, this should probably be a separate issue]
  * to incorporate the improved compiler helper routines (`__ashlsi3`, `__divsi3`, etc.) that are currently in `elks-libc`, into the GCC source tree itself.

Thank you!

mfld-fr:
Hello @tkchia !

I understand that adding 'multilib' support to ELKS 'libc' is a nice exercise, but I still wonder what you could do more after ? For kernel side, the default calling convention is assumed in low level routines, and supporting more would make the code bigger. For user side, I agree building a program could be simpler with only GCC as the front-end and no more direct LD invocations, but is it worth to make so many changes just because of that ?

Except that point I see no complexity in the current build, and I do not see any problem to implement the needed C helper routines into the ELKS 'libc', as it is fundamentally the "C runtime library". Am I missing something ?

tkchia:
Hello @mfld-fr,

[...]

PR #256, opened 2019-04-17
[libc] build up libc.a from separate dirs. in more reliable way

tkchia:
With this patch, the `libc/` subtree now creates a separate `out.a` file for each of its subdirectories, then merges them all together into a single `libc.a` file using `ia16-elf-ar -M`.

I hope to add full support for multilibs once I have updated the assembly source files to work with the different calling conventions now supported by `ia16-elf-gcc` (`cdecl`, `stdcall`, and `regparmcall`).

Thank you!

### DIFF
--- a/libc/Makefile
+++ b/libc/Makefile
@@ -32,21 +32,62 @@ SUBDIRS = \
 
 .PHONY: all $(SUBDIRS)
 
-all: $(LIBC) crt0.o
+ifdef MULTISUBDIR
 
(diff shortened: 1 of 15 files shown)


mfld-fr:
Just tested the full build, got an error on the missing `crt0.o` in `libc`, fixed in commit 1a42f7e2. Thank you for fixing the 'libc' build problem reported in #245 and #255 !

mfld-fr:
Also partial implementation of #252

PR #257, opened 2019-04-19
Build libc.a and crt0.o multilibs for different calling conventions

tkchia:
These patches build separate versions of `libc.a` and `crt0.o` for the various calling conventions (`stdcall`, `regparmcall`, and the default `cdecl`) currently supported by `ia16-elf-gcc`.

With this, one can compile an ELKS program with (e.g.) `-mregparmcall` and then link it with the appropriate version of `libc.a` and `crt0.o`, and the resulting program should run right.

Thank you!


mfld-fr:
Next step in implementation of #252

mfld-fr:
Sorry @tkchia, but this PR does not pass the sanity check (rebuilding all from scratch with `build.sh` then running with `qemu.sh`), and I don't see the faulty commit at first glance, so have to revert this one to a stable state :disappointed:

tkchia:
Hello @mfld-fr,

Argh!  Sorry for the problem.  I will look at it more closely once I have the time.

Any indication of what the problem looks like precisely?

Thank you!

mfld-fr:
No problem @tkchia, that's life. It looks like the ELKS boot hangs just after loading the user land 'init' process, so I guess there is something wrong in the way the 'libc' is built ?

JL2210:
Could you build an example program and give the output from a debugger?

mfld-fr:
Well... I hardly see how to run a debugger on ELKS. Could you please tell us how to proceed ?

tkchia:
Hello @mfld-fr, hello @JL2210,

I suspect that it may be caused by my changes to properly support the variadic syscall routines (e.g. `ioctl` can take 2 or 3 parameters).  But I need to look at the problem more closely.

Thank you!

JL2210:
I suspected that, but wasn't sure of it. Is it possible to get around this
(i.e. by making them all take the maximum number of arguments and finding
if those arguments are valid)?

PR #258, opened 2019-04-22
Revert "Build libc.a and crt0.o multilibs for different calling conventions"

mfld-fr:
Reverts jbruchon/elks#257, that causes a regression in the sanity test (full build from scratch and booting in QEMU).


JL2210:
Shouldn't it just be fixed instead of reverted? What's the regression?

mfld-fr:
Sure, it has to be reworked and pulled again. But any new code shall pass at least the sanity tests so we keep the master branch stable. Please see #257 for regression details.

Issue #253, opened 2019-04-14
Boot regression on FD360


mfld-fr:
Reported on the mailing list: boot is not working any more on an IBM PC/XT with 360K floppy disk: https://www.spinics.net/lists/linux-8086/msg00878.html

mfld-fr:
In the new boot sector code, the disk geometry is fixed and set through configuration. Currently only the parameters of a 3,5'' 1.44M disk are defined. Add the parameters for the 5.25'' 360K disk (CHS = 40 tracks / 2 heads / 9 sectors).

ghaerr:
Is this the reason that ELKS will boot with some emulators, like VirtualBox and 8086tiny, but not others, like DOSBox and fake86?

mfld-fr:
To answer that question, one needs to know the floppy drive geometry as presented by the emulator. I only know that QEMU by default presents a 3,5'' 1.44 MB geometry for the floppy drive. There is some commented code in the new boot sector to display the CHS parameters as reported by the BIOS, that could help to select them in the configuration.

ghaerr:
@mfld-fr: I'm still trying to figure out the reason a few emulators won't boot ELKS. I've been looking around at your rewrite of minix.c -> minix_first.S. Nice job getting rid of the need for the extra ELKS /boot/boot :) However, I still don't understand where one might setup the # sectors for a floppy boot, can you tell me where that's specified in the configuration?

I've attached a screenshot of DOSBox failed boot on a Linux build of the latest source producing images/fd1440.bin. Unfortunately, most of the good information is scrolled off the screen (any way to get that info?). The screenshot does show that minix.bin thinks that the 1.4M 3.5" floppy has 36 sectors when it should only have 18. I'm thinking this is the reason DOSBox and other emulators fail.

If you want to try this yourself, run DOSBox and type "boot /pathto/fd1440.bin"

Can you help me?
<img width="752" alt="Screen Shot 2019-04-18 at 5 04 55 PM" src="https://user-images.githubusercontent.com/11985637/56396672-2a8cc700-61fd-11e9-9e03-47317d81a004.png">

ghaerr:
[...]

PR #259, opened 2019-04-22
Fix OSX build on cp -u, set default floppy boot to use int13 disk params, and fix nano-X in master branch

ghaerr:
I separated out @tkchia's fix for the OSX build from the libc issues.

Also after looking hard at the floppy kernel driver doshd.c, decided simplest solution for now is to enable int13 disk params config so it matches new boot code from @mfld-fr. Added kernel printk to show sector count similar to older probing message.


ghaerr:
@mfld-fr : Sorry, I am getting myself confused with pull requests. This pull request now contains a third commit which was automatically added by Github. This third commit fixes the Nano-X port in the master branch, the nano-X demo programs now run on ELKS, and this also moves the elks/Nano-X over to elkscmd/nano-X where it should be.

mfld-fr:
@ghaerr A good practice is to work in dedicated branches (one for each topic), not the master one, and to dispatch your commits to the right branch, then opening one PR for each branch (so one PR for one topic).

Otherwise there is a risk that I have to revert many things in case of regression (like the #257 from @tkchia), and to lost the whole contributor's work just because of a one wrong commit.

One skilled integrator of a far bigger project even said : "man, you don't need a master branch in your fork, that's the upstream one". I know today he is 100% right :smile:.

mfld-fr:
Forcing the option to use the INT13h AH=08h breaks the things for PC, XT and first generation of AT. A better design would be to first query the BIOS, and if it returns CF=1 and AH=01h (invalid command), then to fall back to static configuration. To be fixed as part of the #253 funnel.

PR #260, opened 2019-04-22
branch for backwards testing of Nano-X and other applications

ghaerr:
Please consider adding this commit so I can continue moving things forward. I have made much progress, but need the ability to run both new and old versions of ELKS on OS X. This commit establishes a version number and branch for older userland applications to be built before gcc-ia16 without having to maintain two completely separate source trees which I'm finding too time-consuming.

I have nano-X now working (!) but, long story short, had to use the old dev86 tools because of the extensive assembly language and complexity of debugging a graphics system.  I am running all this on desktop emulators as well as using the nearly completed but not-released new userland minix filesystem tool on OSX. 

This commit allows the following now:

1) Establishes a 0.2.2 version number that completely works for dev86 on Linux and OSX, since the move to gcc-ia16 happened in between versions.

2) Allows comparing userland binaries between bcc/ld86 and gcc-ia16 to actually measure improyements in executable size (both for application code segments as well as libc routines).

3) Allows building older user programs (like nano-X) in the official tree, until they can be ported to the new compiler. I have already rewritten the main assembly language portions of nano-X, but more is needed, and I need the ability to debug new versus old.

4) This commit moves nano-X from elks/nano-X to where it belongs as an application, elkscmd/nano-X, and also simplifies the contents considerably. I plan on moving nano-X to the same location in the master branch, but don't want to yet, since I haven't been able to actually get it to work yet with gcc-ia16.

The 0.2.2-bcc branch could be deleted afterwards, but since there have been so many changes, and the switch done between versions, there is no easy way to speed app porting. If you don't like the 0.2.2-bcc branch name, call it 0.2.2 instead (branch, not label).

Thank you!


ghaerr:
I'm still a bit new at GitHub processes. What I tried to do with the above pull request is to create a branch called 0.2.2-bcc just after the June 2, 2018 commit fac34cbe2af0cfff091cb9b75afcd0290736ef88. If that branch is created at that point, this commit will automatically merge into that branch.

ghaerr:
Screen Shot 2019-04-22 at 5 19 55 PM

mfld-fr:
Hello @ghaerr, I agree a release should have been issued before starting the GCC-IA16 migration, so I like your idea to pull a branch dedicated to the 2.x line. I will manage to do that based on this PR with a bit of Git sorcery :wink:.

ghaerr:
@mfld-fr : It seems I am having some issues with GitHub on this pull request. The first commit is for creating the 0.2.2-bcc branch as described above. 

I then fixed nano-X on the master branch (finally), and pushed that commit. I was going to open a new pull request for that.  For some reason, Github added that push to this pull request. Thus, when you apply the commit, the first one is for a new branch, and the second is for the master branch.

ghaerr:
@mfld-fr : To clarify, both the commits in this pull request should be added to the new branch 0.2.2-bcc as discussed above. A seperate, but similar pull request deals with the nano-X port in the master branch.

mfld-fr:
I pulled your `0.2.2-bcc` branch directly from your fork and pushed it as `dev86` to this master repository, so that it could be used for reference before the migration to GCC-IA16, as you suggested. Closing this now useless PR.

PR #262, opened 2019-04-26
Add rand, srand and gmtime to libc so that nano-X demos fully build

ghaerr:
This pull request adds misc/rand.c and time/gmtime.c from dev86/libc to elks/libc so that more nano-X applications link. It also sets defconfig to default building nano-X in an ELKS build.

I was a bit concerned as to why gmtime and rand/srand were not included in the initial pull-over of dev86/libc sources, is there a reason they were removed, or perhaps forgotten? Perhaps the libc directories should be diff'd.


mfld-fr:
They were neither removed nor forgotten. ELKS 'libc' is not intended to be a general purpose library, but dedicated to ELKS user land. No more, no less. This is one reason why it is not a distinct project that could be reused elsewhere. One could add any function as needed, like you did, as long as the whole stills POSIX compliant (in best effort for a 16-bit platform).

mfld-fr:
Improves #242

mfld-fr:
Just tested `nclock` after merging this one:
test
Looks to work fine !

PR #263, opened 2019-04-27
[libc] indicate variadic syscall routines in syscall.{dat, awk}

tkchia:
This is part of my re-attempt to add support for non-`cdecl` function calling conventions in userland to `elks-libc`.

Since my previous attempt (https://github.com/jbruchon/elks/pull/257) was not very successful, I am now trying to add this support in stages, rather than all in one go (as before).

Thank you!

Partly for https://github.com/jbruchon/elks/issues/252 .

This change will come in useful when adding support for non-cdecl calling conventions (e.g. stdcall) in userland.

### DIFF
--- a/elks/arch/i86/kernel/syscall.dat
+++ b/elks/arch/i86/kernel/syscall.dat
@@ -9,6 +9,7 @@
 #	'-' = Obsolete/not required
 #	'@' = May be required later
 #	'=' = Depends on stated config variable
+#	'!' = Last argument is optional in libc routine (i.e. variadic routine)
 #
 #	An initial plus on the call number specifies that this call is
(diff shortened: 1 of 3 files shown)


mfld-fr:
Thanks for that retry ! One question : did you test through a full rebuild and by running `qemu.sh` ?

tkchia:
Hello @mfld-fr,

I could not get `qemu.sh` working for some reason, but I did test the build with
```
qemu-system-i386 -fda image/fd1440.bin
```
Thank you!

mfld-fr:
Just tested with `qemu.sh`, no regression during the boot. Also tested `telnet`, no problem. Looks like this change was not the faulty one in #257 failure...

PR #264, opened 2019-04-29
[libc] read ( ) & write ( ) should return ssize_t, not size_t

tkchia:
This fixes a build error that occurred when compiling `gcc-ia16`'s stack protector (`libssp`) together with `elks-libc`.

I am not sure why `read ( )` and `write ( )` were declared to return `size_t` in the first place --- they need to be able to return a negative value (`-1`) in case of an error, and `size_t` is unsigned.

I have tested the change with a run of `qemu-system-i386`.

Thank you!

### DIFF
--- a/libc/include/unistd.h
+++ b/libc/include/unistd.h
@@ -4,12 +4,15 @@
 #include <features.h>
 #include <sys/types.h>
 
+typedef int intptr_t;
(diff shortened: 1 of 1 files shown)


mfld-fr:
Yes, more compliant to the standard. Thank you for the fix !

Issue #261, opened 2019-04-26
Tools for building target file system


ghaerr:
I have a working version of a new tool, mfs, a user-mode minix file system utility that allows mkfs, cp, ls, mkdir, mknod, rm, rmdir, set boot block, and other operations to create and manipulate a Minix V1 system image.

This tool is required on OS X to build bootable ELKS images, but for a variety of reasons which I can detail later, may or may not want to be used as the main mechanism for creating ELKS images. It certainly makes the ELKS build process more portable and eliminates the need for dd, minix.mkfs, mount/umount, sudo, and others.

I have a couple questions:

1) Where should this tool live, and when be built? As part of the cross-compiler toolchain (that is, built before building the rest of ELKS but before elkscmd's "make image")? Since dev86 was thrown out and not replaced, there isn't an external place to store tools, and the gcc-ia16 build already occupies elks/cross and is tremendously complicated. Libc was removed from an external devkit and placed in elks/libc, so should we use elks/tools for this? Mfs is quite convenient and ultimately should be part of an external dev86 environment, along with other tools, a cross-compiler make facility (supporting either bcc for the old 0.2.2 branch testing and gcc-ia16) and libc, IMHO.

2) We need a mechanism in elks/image/Makefile for checking the OS environment, and determining whether to run mfs, or using the existing method, to create the ELKS image.

mfld-fr:
I fully agree that using such tool should **entirely** replace the current image generation. Now, 2 cases:
- either the tool can be sourced and built "as is", or needs only a few lines to be patched, so it can be added as another third-party in `tools/`,
- or it has to be ported with many changes, so one should place the modified sources (including the license) in `elks/tools/`.

tkchia:
Hello @ghaerr, hello @mfld-fr,

A compromise might be to include it as a Git "submodule" --- such that it has its own Git tree, but this main ELKS tree can refer to it, and arrange to check it out when needed.

The GCC IA-16 toolchain is itself already spread out across several Git trees --- `build-ia16`, `binutils-ia16`, `gcc-ia16`, and `newlib-ia16` --- even in Mr. Jenner's upstream version.  (The setup does not use Git submodules, though it probably can.)  So I think there should be no problem if you adopt a similar arrangement for the Minix filesystem tools and other ELKS tools.

Thank you!

jbruchon:
[...]

PR #266, opened 2019-05-01
Enhance mfs to create and copy a filesystem from a template in a single command

ghaerr:
I went ahead and added the capability to create and copy directories, regular files, character/block special files, and symlinks from a template host directory into a new MINIX image, using the new mfs utility.

I modified image/Makefile slightly for testing (under the min: Makefile entry point only), and this is now ready for someone to change the ELKS applications install to a template directory (suggest elkscmd/rootfs_template?). 

Use "mfs <image_file> genfs [options] <template_dir>" and the complete disk image will be generated. Defaults were preset to a 1440-block floppy for ease of use, but can be overridden when desired.

Mfs copies character and block special files as-is, but we may want to use special symlinks or another method so that sudo is not required to mknod the special files in the template in the first place, unless that would be a once-only operation.

Since I am not completely familiar with the application-installation process desired for ELKS, perhaps someone else should change the application installation mechanism to a template directory, create the special files in the template, remove the sudo's from elkscmd/*/Makefile and update image/Makefile.

@mfld-fr : My ELKS fork got de-synced somehow, and there is only one commit "377ee00..." which should be pushed to the image branch. The other three commits in this PR are unintended, I don't know how to remove them.

Thank you!


ghaerr:
Ok, it's all done: Complete ELKS MINIX filesystem build without any sudo requirement, ever.

The build mechanism now creates a filesystem template by installing all applications and the linux Image into elkscmd/rootfs_template without sudo. This template is then used by "mfs genfs" to create a  MINIX image and copy the files to the unmounted image. Then, "mfs mknod" is run creating all the char/block special files on the image without requiring root, since it's not using the Linux kernel to mount or write the image. Finally, "mfs boot" is run to add boot blocks.

The entire build system now works on Linux and OSX, and the issues #261 , #245 , and #242 can now be closed.

I didn't test the rom filesystem build, it should be easily converted to use the template directory to create romfs.bin without requiring sudo.

All elkscmd/* Makefiles were converted to use $(SUDO) instead of sudo, in case sudo is needed, say for the FAT image build. However, the image/Makefile sets SUDO= <blank> for the MINIX build.

Thank you and have fun!

ghaerr:
This last commit removes sudo from the romfs build (trivial).

The ROMFS is now built, but has the problem that the romfs /dev is empty. This is because creating actual character/block devices in the template (or mounted file system) requires root. The MINIX image build gets around this by including "Makefile.devices" which is essentially the old MAKEDEV sh script converted to run mfs mknod commands after the mfs genfs.

I propose the following: implement character and block device intentions in the template directory using symlinks. Symlinks would be created in the template/dev directory using the name as the device name and the link specifying the character/block and device major/minor. These symlinks would be specially interpreted by a trivially modified mfs and mkromfs such that when a symlink is found, if the link starts with "#", such as "#mknod c 1 1", it would mknod a character device 1,1 (mem in this case). Symlinks whose contents don't start with '#' would be treated normally (almost no files start with '#').

This solution is straightforward and would work with any older mechanisms by not adding the /dev symlinks until they are needed in the MINIX or ROMFS image building process. I could forward a PR that would accomplish this quite easily. This would then complete the problem of generating both MINIX and ROMFS using the new approach.

mfld-fr:
Mmm... your proposal looks a bit tricky but fine for the case of the device nodes, but I am afraid it would be not applicable for the case of any file that we would like to change its mode / owner / group on the target.

I think we actually try to implement the same functionality as the `install` command, but redirecting its action from a mounted filesystem (as the `root` user), to a filesystem image (as the current user).

This is why I am starting to consider a more comprehensive solution to substitute the `install` command by our version with the same name (earlier in the path) to keep the installation scripts as is, but that would in turn invoke the tools to operate on the filesystem image, selecting the right one depending on the image configuration, and translating the `install` options (`-o -g -m` and a new custom one for the device nodes).

Does it make sense ?

ghaerr:
> 
> Mmm... your proposal looks a bit tricky but fine for the case of the device nodes, but I am afraid it would be not applicable for the case of any file that we would like to change its mode / owner / group on the target
> 
> 

For which cases do we need to change owner/group on the target? It is not done at all at the moment other
than everything is 0,0.

Should this need to be done, a chmod/chown/chgrp is trivially added to mfs.

I suppose less easily for romfs, but does it matter whether a group id needs to be specified in romfs?


> I think we actually try to implement the same functionality as the install command, but redirecting its action from a mounted filesystem (as the root user), to a filesystem image (as the current user).
> 
> 

That’s exactly what my current PR does: installs to the rootfs_template as the current user. Have you tried it?

> This is why I am starting to consider a more comprehensive solution to substitute the install command by our version with the same name (earlier in the path) to keep the installation scripts as is, but that would in turn invoke the tools to operate on the filesystem image, selecting the right one depending on the image configuration, and translating the install options (-o -g -m and a new custom one for the device nodes).
> 
> 

You said earlier you didn’t want to change the workflow all projects use: install. Now you’re saying
that you want to rewrite install specially for ELKS? What for, so that some program can specify 
a new uid/gid? And thus any program to be added to ELKS would have to substitute a modified
installation script just to get on a boot diskette? Sounds complicated for no good reason.
> Does it make sense 
> 
> 

No, not really :(. 

My take is ELKS is far too complicated for its own good already. I have written a MINIX filesystem
utility that removes the need for sudo/root and works with the current install mechanism. If you
want to rewrite the installation mechanism to be more complicated for ELKS, so be it, but we
shouldn’t have to wait for such a low-priority enhancement as adding group id’s to the installation
mechanism versus removing root requirement in the meantime. My $0.02.

My whole approach has been to try to get ELKS “more usable” by more people. Getting it running
on OSX, removing root requirements, making it easier to understand, install and contribute.
There’s more but I hesitate to begin.

Since I’m already writing code into a non-master branch anyways, I ask that you accept the current PR into
the image branch. Then other people can use it and comment. Also, I can’t do any more programming without
GitHub automatically adding to the same PR.

jbruchon:
Once again, I think the commands that need to be different for different install scenarios should become variables `$IMAGE_CP $IMAGE_CP_PARAMS` etc. and those variables set appropriately and automatically for the user's unique situation. PATH trickery and command name overlaps sounds like a confusing mess waiting to happen. The variable approach allows rewriting of the commands and required parameters on-the-fly in a way that makes sense to a user that is learning about how our system works for the first time. If I am a new user picking up the project and I see `install` in a build script, it had better go to `/usr/bin/install` as I expect. Anything different is simply not acceptable.

ghaerr:
>> This is why I am starting to consider a more comprehensive solution to substitute the install command by our version with the same name (earlier in the path) to keep the installation scripts as is, but that would in turn invoke the tools to operate on the filesystem image, selecting the right one depending on the image configuration, and translating the install options (-o -g -m and a new custom one for the device nodes).
>> 
>> 
> 
> You said earlier you didn’t want to change the workflow all projects use: install. Now you’re saying
> that you want to rewrite install specially for ELKS? What for, so that some program can specify 
> a new uid/gid? And thus any program to be added to ELKS would have to substitute a modified
> installation script just to get on a boot diskette? Sounds complicated for no good reason.

@mfld-fr: Ok, I reread your response and now understand I was mistaken. Users would not be
required to rewrite application install scripts.Your rewritten install script would be the same
name as install, but earlier in the path, so that users wouldn’t be impacted and it would
all automatically work. 

Except it won’t: the complete process of building an installation disk has more issues than
just application installation. Device nodes, for instance don’t use an “install” script, so they
wouldn’t get the automatic configuration that your idea proposes, and would still have to be
specifically handled with the image-making script. The entire image/Makefile is necessary
to handle all these other cases than its single “MAKE -C $(ELKSCMD_DIR) install” command in the middle of it.

So, my conclusion is not changed much: while your idea of having two scripts both named
install (confusing) would nicely solve the problem of allowing user and group id’s to be specified
in the install script, it poses much more complexity for a very small benefit. And it still doesn’t
actually solve the problem of getting the job done since that script would have to record information for AFTER
the applications were installed on the template and the files copied. Otherwise an install
script would have be run sudo/root in order to set a different owner/group id. We definitely
don’t want that, and mfs and romfs automatically creates files owned by root in the target
filesystem, which are 100% of the current cases. Any file that truly needs as different uid/gid than
root could easily be changed by special-casing it and running an mfs chown or mkromfs chown
command after the filesystem is genfs’d.

I reiterate that I’m trying to keep things simple for the most-used cases, which was the suggestion for the
next change to creating the /dev devices, which needs to be done every time.

Post-thought: if you’re going to spend time rewriting application installation mechanisms, I propose
that you include more important abilities like which set of diskettes the application might want
to be installed on, application dependencies for e.g. TCP vs no-login vs graphical environments for the image
build, etc. In this way a set of diskettes could be produced that perform a useful function for a user,
while still having some space on them to do something useful with.

ghaerr:
> Once again, I think the commands that need to be different for different install scenarios should become variables $IMAGE_CP $IMAGE_CP_PARAMS etc. and those variables set appropriately and automatically for the user's unique situation. PATH trickery and command name overlaps sounds like a confusing mess waiting to happen. The variable approach allows rewriting of the commands and required parameters on-the-fly in a way that makes sense to a user that is learning about how our system works for the first time. If I am a new user picking up the project and I see install in a build script, it had better go to /usr/bin/install as I expect. Anything different is simply not acceptable.
> 
> 
I don’t quite fully understand Jody’s proposal for variables and exactly where they would be put,
but definitely agree that having multiple versions of install is a bad idea.

I’m not trying to take over the application installation mechanism, and don’t really want
to get in the middle of all of @mfld-fr’s or Jody’s desires of exactly how things should work.

This PR could be accepted into the image branch and then others, including the above,
can patch it and make it work for the non-mainline cases, like user/group id specifications
and possibly MSDOS disks. The PR completely works for MINIX with no changes, and
would work for ROMFS with the suggestion made earlier. The images built would be the
same as they are today. After that, of course feel free to modify and change it as the ELKS
maintainers desire. I have tried hard to contribute something that actually works and could
be used as a baseline from which to add all the more complex things desired later.

jbruchon:
Instead of `cp some/files /mnt/elks/bin/` you'd have something more like`$IMAGE_CP $IMAGE_CP_PARAMS` and for the old-style install you'd have `IMAGE_CP=cp; IMAGE_CP_PARAMS="/mnt/elks/bin"` or similar. I haven't thought it out too far. Now that I think of it, it'd make more sense to define a function:

```
#!/bin/sh
image_cp () {
    test '$INSTMODE" = "mount" && cp "$1" "/mnt/elks/$2"
    test "$INSTMODE" = "mfs" && mfs cp "$1" "$2"
}

image_cp elkscmd/vi/vi bin/vi
```

(As you can see, I've not tried mfs yet so the syntax is surely wrong.) The idea is that you replace the normal commands with something more intelligent that can change behavior depending on the way the files should be copied to the image. If `mfs` is present and executable, `export INSTMODE=mfs` and the commands change from traditional `cp` commands to `mfs` variants. It is entirely possible that we'll add some other install thing in the future like perhaps a `tar` package generator, and then we only have to modify the installation command functions to add the files to a list and drop a little glue somewhere to actually create the archive at the end of the process.

The install system needs more flexibility, so why not add that flexibility? If you'll have to touch every single one of these scripts regardless of the final decision, why not?

ghaerr:
> On May 4, 2019, at 12:59 PM, Jody Bruchon <notifications@github.com> wrote:
> 
> Instead of cp some/files /mnt/elks/bin/ you'd have something more like$IMAGE_CP $IMAGE_CP_PARAMS and for the old-style install you'd have IMAGE_CP=cp; IMAGE_CP_PARAMS="/mnt/elks/bin" or similar. I haven't thought it out too far. Now that I think of it, it'd make more sense to define a function:
> 
> #!/bin/sh
> image_cp () {
>     test '$INSTMODE" = "mount" && cp "$1" "/mnt/elks/$2"
>     test "$INSTMODE" = "mfs" && mfs cp "$1" "$2"
> }
> 
> image_cp elkscmd/vi/vi bin/vi
> (As you can see, I've not tried mfs yet so the syntax is surely wrong.) The idea is that you replace the normal commands with something more intelligent that can change behavior depending on the way the files should be copied to the image. If mfs is present and executable, export INSTMODE=mfs and the commands change from traditional cp commands to mfs variants. It is entirely possible that we'll add some other install thing in the future like perhaps a tar package generator, and then we only have to modify the installation command functions to add the files to a list and drop a little glue somewhere to actually create the archive at the end of the process.
> 
> The install system needs more flexibility, so why not add that flexibility? If you'll have to touch every single one of these scripts regardless of the final decision, why not?
> 
> 
Jody - 

Thanks for the clarification, I get it now.

The problem with that approach is that it changes basic “installation” mechanism so that
the image script has to know which applications to copy, which is NOT the case now.
The current build only does one thing with regards to applications: “MAKE -C $(ELKSCMD_DIR) install”.
That command goes to elkscmd/Makefile and iterates through the applications which have been
selected to build, which copies them to an installation directory, which was and is $(TARGET_MNT). (unchanged).

There’s no immediate need to add each application to be installed into the images/Makefile script
(or worse, each application’s install: entry point). Instead, the approach I’ve coded uses the exact
same installation mechanism, something that @mfld-fr I thought wanted, to install into a local directory.
AFTER THAT, the new image script copies the files to an unmounted target using mfs genfs (genfs does NOT
know the names of the applications to install, it copies an entire directory tree). There’st  no need to
code a separate copy step, as both MINIX and ROMFS are covered, and it increases maintenance
for whenever anybody adds an application into ELKS, which isn’t require now or before. If it is desired to build an MSDOS
distribution, then yes, we would need a more complex mechanism, but those disks aren’t bootable anways
because you can’t mknod special files. An MSDOS version of mfs would solve that problem better anyways,
otherwise root access is require again.

Do you see what I’m saying? Currently, there is no “cp elkscmd/vi/vi bin/vi” , instead "install $(PROGS) $(TARGET_MNT)/bin" is used.

I think the confusion here is talking about how application are installed, when there isn’t any change being
made, except installing to a local directory instead of /mnt. And it was agreed we didn’t want root/sudo priviledge for
image making, as well as didn’t want to change imported application’s install scripts.

I highly suggest that all interested actually look into and play with the mechanism proposed, so it is in fact fully understood.

ghaerr:
>> The install system needs more flexibility, so why not add that flexibility? If you'll have to touch every single one of these scripts regardless of the final decision, why not?
>> 
>> 
None of the application scripts were changed at all, except to remove sudo. That’s it.  That’s how
proper application installation is guaranteed, by not touching the installation mechanism.

    test '$INSTMODE" = "mount" && cp "$1" "/mnt/elks/$2"
In this case, you’re suggesting we go back to a mode requiring root access to build ELKS disks? Why?

Also, the application install scripts are more complicated than “cp”, they use “install”, and sometimes
have multiple invocations, symlinks, etc. You would have to go through each and every one and rewrite it.

ghaerr:
>  It is entirely possible that we'll add some other install thing in the future like perhaps a tar package generator, and then we only have to modify the installation command functions to add the files to a list and drop a little glue somewhere to actually create the archive at the end of the process.
> 
> 

The newly coded mechanism installs all the applications and the linux image 
into a local staging directory that already exists, rootfs_template. If you would
like to have a tar package generated now to create an elksfd1440bin.tar.gz right now,
you don’t need to change any of the installation command functions at all, instead
just execute a “tar zcvf …” from the template directory and we can distribute those as well.
That could very quickly be added to the end of the current process as you suggest.

What’s nice about the revised mechanism is that it creates the entire distribution in
a local directory, prior to running any mfs, cp, or mkromfs. In that way, the distribution
is identical regardless of format, and is only changed by the needs of the distribution filesystem format,
and for that we’re using tools rather than sudo mount and cp.

mfld-fr:
Even if it looks like we still have some work to tune that new image build, I would like to thank @ghaerr for that fruitful discussion and his contribution effort for that first working version. Let us play with it now. I keep the corresponding issues opened until I fully understand the outcomes and completed the non-regression testing, especially on ROMFS.

ghaerr:
> Even if it looks like we still have some work to tune that new image build, I would like to thank @ghaerr <https://github.com/ghaerr> for that fruitful discussion and his contribution effort for that first working version. Let us play with it now. I keep the corresponding issues opened until I fully understand the outcomes and completed the non-regression testing, especially on ROMFS.
> 
> 
@mfld-fr: thanks. Do you think I should make the symlink mods now to mfs and mkromfs so that
device files can be built into ROM images without sudo, or wait?

mfld-fr:
I think I should spend some time to play with the current version of the new image build, and try some ideas I got through our discussion, before resuming the integration. So please let us wait for the new code and our thoughts to mature a bit...

PR #267, opened 2019-05-05
elks-libc: allow `make install` if supported by gcc-ia16

tkchia:
Hello all,

As part of my to more closely integrate the installation of `elks-libc` and `gcc-ia16` (https://github.com/jbruchon/elks/issues/252), I had added support on the GCC end for a separate `elks-libc` library subdirectory (e.g. `/usr/ia16-elf/lib/elkslibc`).

These patches to the ELKS tree basically add to `libc/Makefile` a `make install` rule.  If `gcc-ia16` supports it and if `$(PREFIX)` is defined, the `install` rule will install the `elks-libc` library files --- `libc.a`, `crt0.o`, linker scripts, and include files --- outside the ELKS build tree and into the GCC installation tree.

I have tested the patches with a full build of the ELKS tree plus a `qemu-system-i386` run.  The patches should also work with older versions of `gcc-ia16`: in this case the makefile will simply not define a `make install` rule, but other than that, ELKS can still build as before.

Much of the complexity in the patches is in parsing the output of `ia16-elf-gcc -print-multi-lib`, as explained in the makefile.

Thank you!

### DIFF
--- a/libc/Makefile
+++ b/libc/Makefile
@@ -32,16 +32,88 @@ SUBDIRS = \
 
 .PHONY: all $(SUBDIRS)
 
-ifdef MULTISUBDIR
+# Stuff for handling multilibs (https://wiki.gentoo.org/wiki/Multilib),
+# which ia16-elf-gcc uses to support separate library files for multiple ABIs
+# for the same architecture and OS.
+#
+# Much of the complexity in the code below is for processing the output of
+# GCC's -print-multi-lib option.  `ia16-elf-gcc -print-multi-lib' prints
+# something like this:
+#
+#	.;
+#	...
+#	rtd;@mrtd
+#	...
+#	regparmcall/any_186/size;@mregparmcall@march=any_186@Os
+#	...
+#
+# The last line above means that the options `-mregparmcall -march=any_186
+# -Os' (listed after the `;') will make GCC look in the subdirectories
+# regparmcall/any_186/size/ within the library search path's directories,
+# when looking for library files.  `.' in the first line is the subdirectory
+# for default compiler options.
+#
+# So, in Apr 2019 I added an -melks-libc option to ia16-elf-gcc, to try to
+# support linking with elks-libc and not Newlib, when targeting ELKS.  Later
+# I added a separate multilib setting for -melks-libc.
+#
+# Thus depending on the version of ia16-elf-gcc, there are two possibilities:
+#   * gcc has separate -melks-libc multilibs (and supports -melks-libc).
+#   * gcc supports -melks, and possibly -melks-libc, but does not have
+#     separate -melks-libc multilibs.
+#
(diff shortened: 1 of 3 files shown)


mfld-fr:
Hello @tkchia, I merged your PR and made some tests. Sanity test in default configuration is OK, but unfortunately, it looks like the new version of GCC-IA16 breaks something in the ELKS network stack, as the `telnet` test fails. The very first error is `[ne2k] IRQ 9 request error: -16` during the kernel boot.

mfld-fr:
I made more testing to be sure: I checked out commit e20fa1cd (merge of #264), rebuilt the cross tools, changed the configuration to add the NE2K driver, rebuilt all ELKS, and ran QEMU. There is no more error during the kernel boot (`[eth] NE2K driver OK`), and I am now able to `telnet`.

Then I came back to the latest master and made the same, except I kept the previous GCC-IA16 version in `cross/`: no problem on `telnet`.

So I am sorry, but I have to revert a34d0553.

tkchia:
Hello @mfld-fr,

Argh!  Thanks again for the report.

I am not sure how to reproduce the problem you saw though (and this makes it harder for me to debug it).  When I run `qemu.sh` I only get a blank screen in the virtual machine, while the Linux terminal says

```
char device redirected to /dev/pts/9 (label chardev1)
WARNING: Image format was not specified for 'image/fd1440.bin' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.
QEMU 2.5.0 monitor - type 'help' for more information
(qemu) warning: TCG doesn't support requested feature: CPUID.01H:EDX.vme [bit 1]
Could not open option rom 'vgabios.bin': No such file or directory
```

If instead I use the command

```
qemu-system-i386 -net user -net nic,model=ne2k_isa -fda image/fd1440.bin
```

I get just this:

```
...
Starting network process: eth ktcp
Starting network services: ERROR: failed to open eth device /dev/eth
failed to open /dev/eth [-1]
 telnetd.
socket error:
...
```
I will see what I can do to figure this out.

Thank you!

mfld-fr:
The `qemu.sh` script configures QEMU to emulate an ISA machine with the `-machine isapc` option, so the VGA BIOS is not the same as for a standard PCI machine:
```
lrwxrwxrwx 1 root root      32  1 mai   03:50 vgabios.bin -> ../seavgabios/vgabios-isavga.bin
lrwxrwxrwx 1 root root      32  1 mai   03:50 vgabios-stdvga.bin -> ../seavgabios/vgabios-stdvga.bin
```
So I suggest you check your QEMU installation against the `seavgabios` package.

tkchia:
Hello @mfld-fr,

Thanks!  I think I have found out why the network driver was failing --- I will send another pull request soon.

Thank you!

Issue #241, opened 2019-03-13
Configurable process heap + stack size 


mfld-fr:
Postponed from #225:
- [ ] check data + bss + "heap-stack-env" < 64K - 1
- [ ] default "heap-stack-env" as configurable (currently fixed to 16K)
- [ ] rework the heap + stack space allocation in kernel `execve()`
- [x] restore the heap + stack size parameter (previous ld86 -H option) : **DONE** in #365
- [x] tool at ELKS runtime to adjust the data size : **DONE** in #337

tkchia:
Hello @mfld-fr,

I believe this can be done in the linker script: instead of saying `LONG (0)` for the chmem field, check whether a particular symbol is `DEFINE`'d and use that in an expression.  Then on the command line to GCC we can use `Wl,--defsym,`... to tell the linker script what to do.

Thank you!

mfld-fr:
@tkchia Good idea ! It would be more simple than a new command to hack the header after the linking.

tkchia:
Hello @mfld-fr,

I recently added a `-maout-total=`_t_ option to `gcc-ia16` to allow setting the heap + stack top address in the a.out header (with some co-operation from the linker scripts).

The option should both work when linking against Newlib (e.g. `-melks -mcmodel=small -maout-total=0xc000`) and against ELKS's `libc` (e.g. `-melks-libc -mcmodel=small -maout-total=0xc000`).

The `gcc-ia16` version currently downloaded by the ELKS build (https://github.com/jbruchon/elks/commit/0fb8d0a6b9807db2157a1e17d98c59a02c912bd6) should already include this option.

Thank you!

mfld-fr:
Hello @tkchia, thanks for your pro-active attitude ! Mmm... I am wondering about the right naming + semantic for this additional parameter when building a program... don't you think that in place of `aout-total`, it should be better to have a more intuitive name for the user to mean "minimum size for heap + stack + env" ?

tkchia:
Hello @mfld-fr,

> thanks for your pro-active attitude !

😃

> don't you think that in place of `aout-total`, it should be better to have a more intuitive name for the user to mean "minimum size for heap + stack + env" ?

Perhaps -- but at least for now, it seems that `total` is the least confusing name I can find, for this particular a.out field.

[...]

PR #269, opened 2019-05-10
LD scripts: set a.out's heap + stack size if `_total` defined

tkchia:
Partly addresses https://github.com/jbruchon/elks/issues/241 .

With this patch, if a user links an ELKS program with `-Wl,--defsym=_total=`_t_, the resulting a.out header will ask for _t_ bytes to be allocated for the whole data segment --- i.e. static data, heap, and stack.

The linker scripts will check and adjust _t_ to a sane offset value, in much the same way as `dev86`'s `ld86` (see https://github.com/jbruchon/dev86/blob/master/ld/writex86.c).


mfld-fr:
Thank you @tkchia for this implementation step !

PR #270, opened 2019-05-20
[libc] fix: `make install' forgot to install autoconf.h

tkchia:
This patch fixes a bug in the `install` rule I added to `libc/Makefile` to allow copying the `elks-libc` files into a `gcc-ia16` installation (part of https://github.com/jbruchon/elks/issues/252).

I found that `make install` forgot to install the `include/autoconf.h` file produced by `config/Configure` and friends, and the other `elks-libc` include files required `<autoconf.h>` to be present to work properly.

Thank you!


mfld-fr:
Thank you @tkchia for this fix !

PR #273, opened 2019-06-14
Toolchain build script changes.

Mintsuki:
Fix toolchain build scripts and Makefile to download GCC prerequisites and acknowledge MAKEFLAGS.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -31,7 +31,7 @@ $(DISTDIR)/$(BINUTILS_DIST).zip:
 	-rm -rf binutils-build
 	mkdir binutils-build
 	cd binutils-build && ../binutils-src/configure --target=ia16-elf --prefix="$(CROSSDIR)" --disable-gdb --disable-libdecnumber --disable-readline --disable-sim --disable-nls
-	make -C binutils-build
+	make -C binutils-build $(MAKEFLAGS)
 	make -C binutils-build install
 	touch .binutils.build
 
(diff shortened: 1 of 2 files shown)


mfld-fr (commented):
Hello @mintsuki, thanks for the proposal. I understand the modified line to download the GCC prerequisites, but could you please explain the lines about MAKEFLAGS ? I don't see where this environment variable is initially set when invoking `env.sh` ?

tkchia:
Hello @mfld-fr,

> could you please explain the lines about MAKEFLAGS ?

My understanding is `$(MAKEFLAGS)` is set by GNU `make` itself, when you pass options (say, `-n` or `-k`) to `make` ([q. v.](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html)).

I have not really tried to see what happens if I pass options to `make`, but it might be good for the makefile to pass these options to sub-`make` tasks.  One way to do this will be to manually write `make $(MAKEFLAGS)` as done here.  But it is also possible to do so by invoking `$(MAKE)` --- which should automatically include `$(MAKEFLAGS)` --- instead of `make`.

Thank you!

Mintsuki:
Hi, the idea was to invoke `tools/build.sh` with a set `MAKEFLAGS` variable so that one could do something such as (on GNU Bash):
```sh
MAKEFLAGS=-j8 tools/build.sh
```
I realised after making the PR that the `env.sh` file is sourced by the main `build.sh` file as well, and in this case it does not work and is overridden by it to use 4 threads. I haven't looked into it, but I think it would be a better idea to let the user choose their `MAKEFLAGS`, rather than defaulting to `-j4`.
If I can find and remove the override I'll update the PR.

Mintsuki:
I noticed the `$THREADS` variable in `build.sh` is autodetected, but it is also ignored and never used. Forcing a parallel build actually breaks the build process yielding this error:
```
make[2]: *** No rule to make target 'bltin/operators.c', needed by 'bltin/operators.o'.  Stop.
```

mfld-fr:
As the parallel build is broken, I would like rather to see it fixed, or totally removed, in this PR.

mfld-fr:
Hello @mintsuki, just to tell you that your PR is not forgotten, but its processing would be delayed because I am on holidays. Stay tuned !

mfld-fr:
Took the proposed changes, but still have to rework the things a bit, like using $(MAKE) as suggested by @tkchia, and wiping out the broken parallel build for ELKS parts (does not work because of dirty dependencies).

PR #281, opened 2019-12-15
[elksemu] auto-detect {call, defn}_tab.v locations, if not given

tkchia:
This patch causes the `elksemu` build procedure to try to look for the `call_tab.v` and `defn_tab.v` files installed from `elks-libc` --- with the help of the `ia16-elf-gcc` driver program --- if their location is not explicitly specified.


mfld-fr:
Thank you @tkchia for that improvement !

PR #287, opened 2020-01-25
[libc] address gcc-ia16 warnings for libc/misc/{crypt, getcwd, getopt, rand}.c, libc/stdio/stdio.c, and libc/string/string.c

tkchia:


mfld-fr:
Warnings in `libc` down from 106 to 64... thanks again @tkchia !

PR #291, opened 2020-01-30
Tools for building target file system

mfld-fr:
Common work with @ghaerr in staging branch `image` is now complete:
* added MFS tool to build the MINIX FS without using the driver in the host kernel,
* replaced the MKNOD & MKDEV by list of devices to fully build as normal user,
* made the `mkromfs` tool consistent with above changes,
* many cosmetics in the build scripts for easier understanding.

Sanity tested in default configuration (standard PC) under QEmu and in ROM only configuration under `emu86`.


ghaerr:
@mfld-fr: This looks great! Also thanks for cleaning up and testing my mkromfs patch.

There's one other small change, that can be done before or after this merge: in image/Makefile, the "include Make.devices" and the comment above it can be deleted, as the revised mechanism uses "make -f Make.devices" rather than including it.

mfld-fr:
Added a commit to cleanup the `image/Makefile` as requested.

PR #292, opened 2020-02-01
[boot] shrink bootblocks code via various tricks

tkchia:
These two commits reduce the size of the boot blocks' code, from `0x1ee` bytes in the first sector and `0x1de` bytes in the second sector, down to `0x1cc` bytes and `0x19e` bytes respectively, without reducing functionality.

The main things I did were
  * using the `regparmcall` parameter passing convention (which I had recently introduced into `gcc-ia16`) when interfacing with the C code, instead of the default stack-based `cdecl`;
  * making the diskette drive parameter table overlap unneeded parts of the boot code; and
  * reimplementing `seg_data ()` as an inline macro, so that it mostly compiles into just a 1-byte `push %ss`.

Thank you!

### DIFF
--- a/elkscmd/bootblocks/Makefile
+++ b/elkscmd/bootblocks/Makefile
@@ -1,6 +1,6 @@
 
 CC = ia16-elf-gcc
-CROSS_CFLAGS = -fno-inline -mcmodel=tiny -mno-segment-relocation-stuff -ffreestanding -mtune=i8086
+CROSS_CFLAGS = -mregparmcall -fno-inline -mcmodel=tiny -mno-segment-relocation-stuff -ffreestanding -mtune=i8086
 INCLUDES = -I $(TOPDIR)/include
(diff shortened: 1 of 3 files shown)


mfld-fr:
Oup's, we worked on the same files at the same time, and now this PR has one conflict... could you please rebase your branch on master ? I will wait for your PR to be merged before continuing...

tkchia:
Hello @mfld-fr,

I have rebased the code.  Thank you!

tkchia:
Hello @mfld-fr,

(I did not notice that you removed the commented-out `word_hex (...)` routine in `boot_sect.S`.  I have now removed it too.  Thank you!)

mfld-fr:
A bit too tricky for me, but I like it :smile:
Thank you for the demonstration !

tkchia:
Thanks @mfld-fr!

mfld-fr:
I gained two more bytes after fixing 2 typos (`sector_1` and `puts_exit`) :grin:

tkchia:
:-)

PR #293, opened 2020-02-02
[boot] further improvements to boot blocks code and data

tkchia:
With these patches:
  * In the boot sector, there is now a marker string `<eLkS>` (and a size indicator) before the block of disk parameters that give the disk geometry.  This allows other programs, possibly even the kernel, to read or write these disk parameters, by searching for this marker.
  * To the disk parameter block, I also added parameters to allow customizing the location and size (as recommended at https://github.com/elks-org/elks/issues/288#issuecomment-581140441) of the OS-specific boot payload.  Thus the payload no longer needs to be always on the disk's second sector.
  * Also, the OS-independent boot sector no longer needs to know the precise location of the OS-specific `load_prog ()` routine.  It now always does a `call` to offset `0x200`.  The onus is on the OS-specific code (e.g. `boot_minix.c`) to do a jump from there to its own `load_prog ()`.

As side effects:
  * The `disk_read (...)` code, originally used only for reading the ELKS kernel, is now also used to read the OS-specific payload.
  * I also enabled and updated the logic for retrying reads in the `disk_read (...)` routine.

Thank you!


elkscmd/bootblocks/Makefile
```diff
@@ -1,6 +1,6 @@
 
 CC = ia16-elf-gcc
-CROSS_CFLAGS = -mregparmcall -fno-inline -mcmodel=tiny -mno-segment-relocation-stuff -ffreestanding -mtune=i8086
+CROSS_CFLAGS = -mregparmcall -fno-toplevel-reorder -fno-inline -mcmodel=tiny -mno-segment-relocation-stuff -ffreestanding -mtune=i8086
```
mfld-fr: What is this `-fno-toplevel-reorder` flag ?

mfld-fr: Forget my comment, the answer is already given in another location.

ghaerr:
I have a couple questions and comments about this. Please forgive me, I've not studied the changing boot code in detail, so please explain if I'm going the wrong direction or grossly misunderstand.

1. Why use extra bytes for an <eLkS> marker when we could place the disk parameter block just before the magic boot signature mark 0xAA55 at 0x1FE? Any boot parameter modification programs could just look at a fixed offset rather than searching.

2. Why save the size of the disk parameter block in itself? The structure of the block would have to be known by outside programs anyways, there's no need to save its size in the boot block parameter copy, is there?

I don't yet understand why the disk parameter block is copied to floppy_table during the boot sequence, can someone explain that to me?

Are we hard-coding the disk parameter block instead of depending on the BIOS for this information at boot time?

On another issue: I am just about finished with the modifications to "mfs" that will allow my new after-build package management facility to select sets of applications and create 320k, 360k, 720k, 1440k minix volumes all in one pass with the applications installed that will fit each size. So I am concerned about how complicated we want to make having to change boot sector 1 for each floppy size. The package manager will immediately support creating all floppy sizes with a selected set of applications packed to fit the disk size limitations, rather than using .config reruns for each floppy size. (As an example of the need for this, the ELKS applications are too large for the default 1440k floppy build currently implemented, I have had to comment out 5 graphical nano-X demo applications since otherwise the build would fail on image writing. With package management, all applications are built by .config and then selected based on image size). Currently, it looks like the .config file is used to set the # heads, max sectors, etc through CONFIG_IMG_* options. The package manager will have to set different disk parameters in order to build 360k/720k/1440k floppies, and I assume will the modify the disk parameter block?

Since the mfs program is now mkfs'ing the volume and writing the boot block(s), do we still want to have all the disk parameter information set in the .config file (not bad for defaults, but always overwritten for every other floppy size)? I am not sure whether the boot code calls the BIOS to get the disk parameter table currently in use, or whether the new plan is to have all this hard-coded. It will be easy to have mfs write this data but I'm just asking about it now to understand the intent of the complexity being built into the boot process. I hope this long explanation makes sense.

tkchia:
Hello @ghaerr,

> 1. Why use extra bytes for an marker when we could place the disk parameter block just before the magic boot signature mark 0xAA55 at 0x1FE? Any boot parameter modification programs could just look at a fixed offset rather than searching.

> 2. Why save the size of the disk parameter block in itself? The structure of the block would have to be known by outside programs anyways, there's no need to save its size in the boot block parameter copy, is there?

Mainly I am thinking about the possibility of having to extend the parameter block with new fields in the future.

I hope to avoid the situation that happened with MS-DOS BPBs, where newer versions of MS-DOS successively added fields to the BPBs, and one needed all kinds of deep magic to decide which new fields could be "trusted" to be valid.

And, I am not sure at the moment which offsets from the end of the boot sector are really "safe" for use without, for example, clashing with how they might be interpreted on a FAT12 filesystem.  (E.g. reportedly some DOS versions and OS/2 store a drive number at offset `0x1fd` --- ugh.)

> I don't yet understand why the disk parameter block is copied to `floppy_table` during the boot sequence, can someone explain that to me?

The [diskette parameter table](http://www.delorie.com/djgpp/doc/rbinter/id/55/24.html) needs to be modified, so that the "sectors per track" is at least equal to the sectors per track for our boot disk's geometry --- otherwise reads may fail.

The default table is likely in ROM, so the boot code copies it to RAM, then makes the modification.

> Are we hard-coding the disk parameter block instead of depending on the BIOS for this information at boot time?

For a fixed disk, yes, it will be reasonable to consult the BIOS to get the fixed disk's geometry.  For floppy disks, though, the BIOS only gives the capabilities of the floppy _drive_, which may not be the same as the geometry of the _disk_ --- e.g. one can perfectly well insert a 360 KiB disk into a 1.2 MiB-capable drive.

So consulting the BIOS does not work very well for floppy drives, and we need some other way.

Thank you!

ghaerr:
Hello @tkchia,

Thank you very much for your explanation about disk parameter tables, I know understand the need to have the disk (not drive) parameters encoded into the boot block! I will make the modifications to mfs to write the disk geometry information in the locations that you and @mfld-fr ultimately decide.

> Mainly I am thinking about the possibility of having to extend the parameter block with new fields in the future.
> 
> I hope to avoid the situation that happened with MS-DOS BPBs, where newer versions of MS-DOS successively added fields to the BPBs, and one needed all kinds of deep magic to decide which new fields could be "trusted" to be valid.

I understand that, but ELKS is already completely dependent on compiling up the entire system each time from scratch (we don't support older ELKS versions for booting or running) - including the boot blocks code, kernel, mfs and apps - so it still seems a bit wasteful to encode a size field (and possibly a magic header, see below) to me, especially when the recent commits have been all about saving a few bytes here and there. We may need those bytes back for better boot debugging, for instance.


> And, I am not sure at the moment which offsets from the end of the boot sector are really "safe" for use without, for example, clashing with how they might be interpreted on a FAT12 filesystem. (E.g. reportedly some DOS versions and OS/2 store a drive number at offset `0x1fd` --- ugh.)

Yes, but those locations are only important when running MSDOS or OS/2 boot code, which we will never do. The FAT12 filesystem will never write the boot block. Thus, all locations below 0x01FE are safe for ELKS since we require running our own boot sector.

mfld-fr:
Maybe a good compromize for now would be to move the boot parameters to the end of the sector, just before the BIOS magic @ 1FEh, and to keep the 'eLkS' marker and the size of that parameter structure ?

ghaerr:
> Yes, but those locations are only important when running MSDOS or OS/2 boot code, which we will never do. The FAT12 filesystem will never write the boot block. Thus, all locations below 0x01FE are safe for ELKS since we require running our own boot sector.

While coding up the MSDOS FAT12 disk image package management routines, I studied the MSDOS filesystem structure in depth, and I've realized that all three of my statements above are incorrect: 1) The MSDOS FAT12/FAT16/FAT32 disk images, whether bootable or not, *always* use certain fixed fields in the boot sector, called the BPB (BIOS Parameter Block), in order to learn the disk geometry of the disk inserted into the floppy drive. 2) All FAT filesystem drivers and utilities will *ALWAYS* read (but not write) the boot block BPB fields. In reading the Microsoft FAT specification (http://read.pudn.com/downloads77/ebook/294884/FAT32%20Spec%20%28SDA%20Contribution%29.pdf, Section 3.1 and 3.2 BPB Structure in FAT12/FAT16/FAT32), all fields between offsets 11 and either 36 or 62 decimal are required, and 0xAA44 at offset 510, or MSDOS and most FAT reading/writing utilities will fail.

Thus, if we're seriously thinking of creating MSDOS bootable floppies, our current boot sector needs to be redesigned to allocate the specified BPB_* fields from 11 to 62 decimal as specified in the FAT document. If we want to have a common boot loader in sector 1, then the MINIX boot loader should also store the sectors per track and # heads in these locations, NOT the locations we're currently contemplating. Doing this would mean we need 62-11+1 = 52 bytes free in our common MINIX/MSDOS boot loader in sector 0. We may be able to use a smaller BPB with just (35-11+1 =) 25 bytes, I haven't tested that yet.

In testing this theory, I used mfs to write a minix boot block to a FAT volume in the new package manager, after having successfully created a 1440k FAT16 volume. After writing our non-compliant boot block, the mdir program failed to recognize the FAT volume because of the incompatible BPB in the boot block.

Thank you!

mfld-fr:
@ghaerr : you are right for a future version of the boot sector that would handle FAT, in addition to MINIX and RAW. But for now, the intent of this PR is just for more generic & more robustness. Let us plan the space for a FAT12 / FAT16 BPB in a next step (FAT32 is out of the scope of this project).

tkchia:

> But for now, the intent of this PR is just for more generic & more robustness. Let us plan the space for a FAT12 / FAT16 BPB in a next step (FAT32 is out of the scope of this project).

Hmm... on second thought, I guess I should really go back to the drawing board for a while, and try to come up with an ELKS boot parameter structure that dovetails more nicely with the MS-DOS BPB.

At least, I hope to get the structure reasonably right, before fixing it as part of a binary interface.

> After writing our non-compliant boot block, the mdir program failed to recognize the FAT volume because of the incompatible BPB in the boot block.

It seems you are right.  If I try to read a 720 KiB disk under MS-DOS 6 (within QEMU), the system erroneously treats it as a 360 KiB-formatted disk.  I had thought DOS would use the media descriptor byte (`0xf9`) to figure out the format, but that is not the case.

Thank you!

mfld-fr:
@tkchia : FAT BPB structure is not so complex (even with the stacking of different versions), but writing code to get correct disk information is a real nightmare, because of all the cases to consider to identify the actual version.

This is why I would like to move that complexity **out** of the boot code, with just the abstraction of the payload, and just the spacing for the BPB, to let external tools manage that complexity. No try to use some fields of that structure, just keeping the space reserved for it in the boot sector.

@ghaerr : it implies that installing our boot sector is not a simple overwrite, but a read-modify-write.

ghaerr:
> @tkchia : FAT BPB structure is not so complex (even with the stacking of different versions), but writing code to get correct disk information is a real nightmare, because of all the cases to consider to identify the actual version.

This boot sector problem is an easy adjustment - we just need to reserve around 25 bytes using the reserved locations specified in Section 3.1 of the MSFT FAT document from locations 11-36 and set them all to zero, except for our MINIX boot required heads and tracks values. We'll probably have enough space, thanks to all the work that @tkchia and @mfld-fr have recently done squeezing bytes out. We can test correctness and compatibility using the mformat and mdir programs on our MINIX boot sector.

I plan on adding a command line option to mfs that will allow the heads and tracks to be specified when writing the boot sector "mfs boot ..." that will apply the MINIX boot sector with the correct values in the BPB.

Then, using my package management PR, many different floppy format sizes can be produced extremely quickly after each build, and the images tested on the simulator or real hardware from our end users.

We can use the MSDOS FAT formatting tool used in the interim package manager to correctly format various sized FAT images and the boot sector, which removes the nightmare of worrying about which values are correct for the boot sector when the payload is MSDOS boot.

> 
> This is why I would like to move that complexity **out** of the boot code, with just the abstraction of the payload, and just the spacing for the BPB, to let external tools manage that complexity. No try to use some fields of that structure, just keeping the space reserved for it in the boot sector.

Exactly. We're almost there, a rearrangement of a few data fields in the current MINIX boot block and the mfs mod and we're good to go.


> 
> @ghaerr : it implies that installing our boot sector is not a simple overwrite, but a read-modify-write.

Yep.

tkchia:

OK, I have further tweaked the boot code to reserve some space for an MS-DOS 3.0 BPB, which spans offsets `0x3` to `0x1d`.

I stashed some information about parameters specific to the ELKS bootloader in the OEM name, which is rather ugly, but oh well.

Also, there are now only about 6 bytes of space left in the boot sector.  I guess this means we need to further crunch down the size of the boot code (probably starting from the `disk_read` routine).

Thank you!

mfld-fr:
@tkchia : Mmm... could you please push your latest commit in a distinct PR ? I am not sure it allows to simply install the boot sector over an existing one on an already formatted FAT, and it requires more testing. Thanks !

tkchia:
Hello @mfld-fr,

OK, you are right...  This pull request is getting quite messy; maybe I should close this PR first, and redo it as several distinct PRs.

Thank you!

ghaerr:
> OK, I have further tweaked the boot code to reserve some space for an MS-DOS 3.0 BPB, which spans offsets `0x3` to `0x1d`.

I think we will need space up through decimal 35, the last field being the 32-bit total sectors in the filesystem (BPB_TotSec32).

IMO, it would be nice to name the BPB fields from the MSFT FAT document and reserve then with .byte/.word, even though we night not use them. This will ensure understanding and compatibility when the same locations are modified by mfs or mformat utilities when creating boot sectors.


> I stashed some information about parameters specific to the ELKS bootloader in the OEM name, which is rather ugly, but oh well.

That includes our sec_max and head_max, right? These values need to be in the BPB_SecTrk and BPB_NumHeads fields, respectively, not in the OEM area, otherwise the boot sector isn't FAT compatible.

It appears our track_max isn't used internally, and FAT doesn't use it either, why not just delete it?


> Also, there are now only about 6 bytes of space left in the boot sector. I guess this means we need to further crunch down the size of the boot code (probably starting from the `disk_read` routine).

Nice work! LOL, I'm smiling over here: it seems that just after one removes more bytes and makes things smaller and more efficient, something always comes along and uses up all that space again :)

Issue #290, opened 2020-01-30
Binary packages for end users


mfld-fr:
@ghaerr wrote:

> (and open a way to setup an automatic deployment of the releases in the form of 'ready-to-use' binary packages for end users)

I have thought through a design for a lightweight, text-based ELKS package management mechanism using make and command line tools that would allow easy maintenance and addition of ELKS applications to be specified to be copied to MINIX or FAT images during or after each build, or automatically, is there interest in this contribution?

I am thinking of a Packages file in the elks/images directory that would specify which applications would be copied by the "mfs" (and upcoming "msfs") program from the template dir to various images, based on image names like fd360-minix, fd720-msdos, hd-minix, network-minix, etc that both ELKS maintainers, the CI system, and users could use to manage a variety of ELKS distribution or test images. All ELKS applications would be built in every build (unless explicitly turned off) which should help regression testing/rot and the package manager used to produce desired sets of distribution binaries, instead of the current mechanism of specifying each application wanted in the .config file.

mfld-fr:
Moved from #289

mfld-fr:
The selection of the applets to be included into the image **during** the build is already implemented in the configuration scripts, based on an old version of KConfig (BTW should be upgraded from Linux mainstream for better dependency management).

I would advise not to modify that design, inspired by [BuildRoot](https://buildroot.org), that fully decouples the selection of the applets and the selection of the image format. Earlier design was coupling the image formats and their contents, and proved to be unmanageable.

I would prefer a set of scripts that iterates on generating one `.config` for every wanted image, building & packaging & deploying that image, as a second layer based on the existing system.

Your suggestion to build all the applications in issue #261 has already been implemented in commit 4829877a8ba04910b8d7d1539408e283225f4b32.

For selection and inclusion **after** the build, I am fully open...

ghaerr:
> The selection of the applets to be included into the image **during** the build is already implemented in the configuration scripts, based on an old version of KConfig.
> 
> I would advise not to modify that design, inspired by [BuildRoot](https://buildroot.org), that fully decouples the selection of the applets and the selection of the image format. Earlier design was coupling the image formats and their contents, and proved to be unmanageable.

My package management idea was not intended to change the build process or .config/KConfig mechanism, and will improve the ability for selecting applications and image format *after* the build has completed.

[...]

PR #294, opened 2020-02-02
[image] Application Package Management for multiple disk images

ghaerr:
Here's the completed ELKS application package manager, a lightweight, text-based mechanism for specifying which applications are copied onto MINIX bootable images. This PR does not change any of the build process, and is run after the build by typing "make images" in the elks/image directory.

I looked at all ELKS applications and divided them into functional groups that just fit into four different floppy disk sizes for the time being: 320k, 360k, 720k, and 1440k. Running "make images" will produce fd320.bin, fd360.bin, fd720.bin and fd1440.bin (same as the build fd1440.bin) in the elks/image directory. Each of these disk images is packed almost full with as many functional apps as the disk image size will take. Run "mfs fd320.bin stat" or "msfsck -fv fd320.bin" to judge remaining size.

The new images/Package file contains the application tags for specifying the inclusion of various classes of application into groups. These can be easily updated by the ELKS maintainer when new applications are added, or changed by the end user after any build to specify their own disk images, without rebuilding ELKS nor requiring root permissions.

NOTE: The modifications to mfs have not been completed for changing the newly discussed and changed disk parameters in the minix boot block, so currently the only disk image that will boot will be the one that matches the .config file. When this other work is completed, there will not be a need to specify the disk image format in .config (or it can be ignored), as mfs will write the boot block with the correct disk parameter block without requiring a kernel or boot block recompile, as is now needed.

If this PR meets the maintainer's approval, it can be committed with no change to the build process, allowing easy testing by users and subsequent changes to mfs for the decided boot disk parameters mechanism. The applications selected for the various disk images can be observed and commented on for ease of use. 

It is my intention to add support for FAT12/FAT16 image creation support using the package manager in the future, allowing the same set of bootable disks to be created in MSDOS format.


ghaerr:
Just added the ability to create MSDOS FAT12 floppy images using the package management system. Run "make msdosfs" in elks/image to produce fd1440-msdos.bin (non-bootable, but compliant MSDOS FAT12 1440k image).

NOTE: Currently requires mtools-4.0.23 programs mformat, mmd, mcopy and mdir to be installed. I will be adding this package into the elks/tools/mtools directory in the future. Mtools source is available from GNU.

Also, writing the boot sector is not currently supported, as the mtools mdir program complains that our minix boot sector isn't MSDOS compatible when writing current minix.bin. See http://www.disc.ua.es/~gil/FAT12Description.pdf for apparently required boot sector fields layout.

The following files on the ELKS filesystem are longer than the FAT12 8.3 filename restriction, and thus have a VFAT long filename and modified compatible FAT12 filename created, which is not supported in our ELKS FAT file driver: /etc/resolve.conf, /etc/rc.d/rc.sysinit, /root/.profile, and /var/www/index.html. Also, symbolic links are copied as regular files, thus increasing the disk usage on FAT12 volumes. 

Enjoy!

mfld-fr:
I am seeing plenty of good bricks in this PR, mainly the matrix that lists the applications for each package, in a single file that could be easily edited, and the support of FAT image.

First thing that could be improved at low cost is to move the list of media, and the list of packages per medium out of the Makefile to another configuration file. Many users are not at ease with GNU Make, and it would be easier for them to edit a raw text file.

I understand that this PR allows to generate easily some 'ready to use' images for the next upcoming release, and that way answers the request from some end users, that are not able to build but would like to have fun by running ELKS on their old machines.

But I still don't like the big picture, because you are assuming in the proposed design that the kernel and the applications are the same for all the possible media, so you can build once and then package many. This is not true : kernel features, application switches and image format are linked to each other.

For example, if you don't select the networking applications, you can disable the networking and the Ethernet driver in the kernel, resulting in saving much space. And vice-versa : if you disable the Ethernet driver in the kernel, networking applications must be excluded from the image.

I suggest you to have a look to BuildRoot. This is, with other embedded development environments, the state of the art, and they all came to the same conclusion through inspiring each other : one menu for all (kernel, applications and image).

Moreover, this design adds a parallel way to configure the image contents, and would confuse the user : should he go through the configuration menu, or through editing the text files as above ?

So I would prefer what I suggested in #290 : design as an upper layer on the existing one, and iterate on several configurations. In each iteration, generate `.config` from an more easier textual description and build up to the image.

Each iteration could also be a parallel job for the upcoming CI, where we just have to reference the configuration in each job description, speeding the generation of the artifacts (= the images).

ghaerr:
> First thing that could be improved at low cost is to move the list of media, and the list of packages per medium out of the Makefile to another configuration file. Many users are not at ease with GNU Make, and it would be easier for them to edit a raw text file.

I was planning on that, but wanted to give you a look as requested,.


> I understand that this PR allows to generate easily some 'ready to use' images for the next upcoming release, and that way answers the request from some end users, that are not able to build but would like to have fun by running ELKS on their old machines.

This project should be made easier for users to try out and use ELKS, but he current developers need to be able to quickly create all supported disk image sizes for testing, especially when we're still developing boot sectors for six different floppy sizes and simultaneously adding support for direct MINIX and MSDOS filesystem generation for all disk sizes.


> But I still don't like the big picture, because you are assuming in the proposed design that the kernel and the applications are the same for all the possible media, so you can build once and then package many. This is not true : kernel features, application switches and image format are linked to each other.

That's fine. "When building tall buildings, cranes are used, which are removed after construction." I built this for the reasons described above, and am trying hard to move ELKS forward more quickly.

I urge you to commit this PR for the time being. It doesn't interfere with the build process at all (a method you were "fully open to"), and needs only be run by developers for now. I definitely need it in order to use my time effectively and contribute useful additions to ELKS, like MSDOS FAT building support and MINIX image boot testing. You can delete it all later (two Makefile entries and a Packages file), after the menu system is converted to Buildroot (see below).


> For example, if you don't select the networking applications, you can disable the networking and the Ethernet driver in the kernel, resulting in saving much space. And vice-versa : if you disable the Ethernet driver in the kernel, networking applications must be excluded from the image.

That's what I thought /etc/rc.d/rc.sysinit was for, but I agree and understand that a better final design is to do everything in .config, to save every byte.  I don't understand your statement about having to exclude applications without kernel support - does running them crash the system?


> I suggest you to have a look to BuildRoot. This is, with other embedded development environments, the state of the art, and they all came to the same conclusion through inspiring each other : one menu for all (kernel, applications and image).

I have been looking at this after you mentioned it. I am in the midst of reading their 300 page explanation.

However, I'm not going to be the one rewriting the menu system or BuildRoot. The current KConfig doesn't even run on OSX, I've never been able to configure anything using our (ancient) menu system!!!


> Moreover, this design adds a parallel way to configure the image contents, and would confuse the user : should he go through the configuration menu, or through editing the text files as above ?

Users need not be confused for now - they're complaining about not even getting any images!


> So I would prefer what I suggested in #290 : design as an upper layer on the existing one, and iterate on several configurations. In each iteration, generate `.config` from an more easier textual description and build up to the image.

Ok, that can be a project for after we finish getting MINIX boot sector compatibility and MSDOS FAT boot working on all six disk sizes - I certainly don't want to wait through six full ELKS builds while testing MINIX and MSDOS boot on software emulators.

Please consider committing this PR, for my ease of making contributions. It can be deleted later. There is lots more work to do on MSDOS support and we're just getting started.

mfld-fr:
> However, I'm not going to be the one rewriting the menu system or BuildRoot. The current KConfig doesn't even run on OSX, I've never been able to configure anything using our (ancient) menu system!!!

Oh :open_mouth: ... I was not aware of that. Now I understand the way you take. We should fix that problem quickly.

ghaerr:
> Oh 😮 ... I was not aware of that. Now I understand the way you take. We should fix that problem quickly.

Thank you for understanding... but it's not that important to me that it be fixed very soon. Lets get MSDOS boot working first! :)

PR #295, opened 2020-02-03
[boot] (take 2) put boot_minix.c's entry point at sector start

tkchia:
This means that the OS-independent boot code (`boot_sect.S`) does not need to know where precisely the OS-specific `load_prog()` routine is.  It can instead simply do a `call` to the start of the second sector.


mfld-fr:
Fully agree for this one 😄

PR #297, opened 2020-02-03
[image] Cleanup MSDOS disk image creation and format

ghaerr:
FAT12 image cleaned up to all zeros except boot block and ensure \linux is first dir entry
Confirmed only boot block offsets 11 through 62 are modified, nothing else

This PR cleans up previous submission where I found mformat has bugs and didn't create an all-zero filesystem, making it hard to confirm exactly which portions of the image are changed when mformat modifies the boot sector BPB.

I also confirmed that when using mformat for our MSDOS disk image creation, it will now *only* modify boot sector offsets 11 through 62. This means that the process to create MSDOS bootable images will be first to copy the ELKS boot sector (either one or two sectors) to the image start, *then* use mformat to update offsets 11 through 62 with the BPB with the disk format parameters and create the FAT12 filesystem. The initial format now ensures that \linux is the first directory entry, simplifying the boot loader. Finally, all applications and the real /linux will be copied, using the same first directory entry for \linux.

Updated elks/image/Make.packages documentation for mtools parameter used, for easier testing by others.

Also removes swapon from the disk images.

### DIFF
--- a/elks/tools/mfs/main.c
+++ b/elks/tools/mfs/main.c
@@ -40,6 +40,7 @@
  *	use ELKS defaults of -1 -n14 -i360 -s1440 for mkfs/genfs
  *	add genfs -k option to not copy 0 length (hidden) files starting with .
  *	add addfs option to add files/dirs specified in file from directory
+ *	add boot -B option to update BPB sector and head fields in boot block
  *
  * Bug fixes by ghaerr:
  * fix mkfs -1, -n overwriting -i, -n14
@@ -73,6 +74,7 @@ void usage(const char *name) {
 	"	-v	verbose\n"
 	"cmd:\n"
 	"	mkfs [-1|2] [-i<#inodes>] [-n<#direntlen>] [-s<#blocks>]\n"
+	"	boot [-B<sectors>,<heads>] [bootimage]\n"
 	"	genfs [-1|2] [-i<#inodes>] [-n<#direntlen>] [-s<#blocks>] [-k] <directory>\n"
 	"	addfs <file_of_filenames> <directory>\n"
 	"	[stat]\n"
--- a/elks/tools/mfs/mkfs.c
+++ b/elks/tools/mfs/mkfs.c
@@ -109,34 +109,67 @@ void cmd_mkfs(char *filename, int argc,char **argv) {
   close_fs(fs);
 }
 
+#define BPB_SecPerTrk	24		/* offset of sectors per track (short)*/
+#define BPB_NumHeads	26		/* offset of number of heads (short)*/
 /**
  * Write boot block to image file
  */
 void cmd_boot(char *filename, int argc,char **argv) {
-  FILE *ifp, *ofp;
+  FILE *ifp = NULL, *ofp;
   struct stat sb;
   int count;
+  int opt_new_bootblock = 0, opt_updatebpb = 0;
+  int SecPerTrk, NumHeads;
   u8 blk[MINIX_BOOT_BLOCKS * BLOCK_SIZE];
 
-  if (argc != 2) fatalmsg("Usage: %s [boot file]\n",argv[0]);
-  if (stat(argv[1],&sb)) die("stat(%s)",argv[1]);
-  if (!S_ISREG(sb.st_mode)) fatalmsg("%s: not a regular file\n",argv[1]);
-  if (sb.st_size > MINIX_BOOT_BLOCKS * BLOCK_SIZE)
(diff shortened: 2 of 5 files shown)


ghaerr:
Updated MSDOS image building to build 320k, 360k, 720k and 1440k FAT12 images for testing. Use "make dosimages" to make, or "make allimages" to make all 8 MINIX and MSDOS test images.

Prepared (but still commented out) MINIX boot BPB update option for "mfs boot -B<sectors>,<heads>" that will update MINIX BPB after applying boot sector, or just update the BPB if no boot file specified.

Question: For the ELKS 360k format, is that for a 5" or 3" floppy drive? I thought ELKS just supported the 3" drive family (360, 720, 1200, 1440, 1680). While looking through elks/elkscmd/bootblocks/boot_sect.S, it shows CONFIG_IMG_360 using 2 heads, which is correct for 5", but incorrect for 3". For 3" floppy drives, this should be 1 head. See (https://en.wikipedia.org/wiki/List_of_floppy_disk_formats) under IBM PC logical formats.

I was thinking the ELKS 320k and 360k formats were 1 head and 8 and 9 sectors, respectively.

ghaerr:
Added mfs boot option to update sector and head fields (shorts at offsets 24 and 26) in DOS BPB area of boot sector for MINIX boot images. All tools are now completed for transitioning to BPB format boot blocks and also supporting any variety of MINIX or MSDOS disk image formats.

Usage (commented out in package build for now):
mfs diskimage boot bootimage -- write bootimage sector(s) as is to diskimage
mfs diskimage boot -B9,2 bootimage -- write boot sector(s) and set BPB to 720k (9 secs, 2 heads)
mfs diskimage boot -B18,2 -- update BPB in diskimage to 1440k (18 secs, 2 heads)

Mellvik:
Gregory,

You must have misread the Wikipedia article about diskette types and densities (and heads).

ELKS always supported 3,5 and 5,25 inch floppies. And both types always had 2 heads. 
3,5" started at 720k doubled to 1.44, then 2.88 - the latter seems to be rarely used.
5,25" started at 320, then 360 (8 and 9 sectors respectively, as you point out - times 40trks x 2heads) and finally 1200 (18x80x2).

It seems the assumptions in ELKS about this are OK.


Mellvik

ghaerr:
> ELKS always supported 3,5 and 5,25 inch floppies. And both types always had 2 heads.

@Mellvik, thanks for your reply. I will update my boot block head information so that 2 heads are used for the 320k and 360k 5.25" floppies that ELKS supports.

> You must have misread the Wikipedia article about diskette types and densities (and heads). 
> 5,25" started at 320, then 360 (8 and 9 sectors respectively, as you point out - times 40trks x 2heads) 

I reread the Wikipedia article and the source of my confusion was that 3.5" floppies were also produced in 320k and 360k formats - these used 8 and 9 sectors and only 1 head. What I didn't realize is that 5.25" floppies with the same capacities used 2 heads, and those are the ones that ELKS supports.

On another note, now that we're using a tool to set the head and sectors in the boot block separately from compilation, we could support all four varieties of 320k and 360k disks, in both 3.5" and 5.25", if someone wanted that. Of course, they could also set this up using the menu system as well for direct compilation.

Mellvik:
I suggest continued focus on simplicity, thus as few (diskette) formats as makes sense. The starting point, baseline so to speak, for ELKS is the IBM PC, so going further back in history for compatibility does not make sense.

the 5.25in 320K format existed before the PC and before 1980 (it was supported by MS-DOS, as was a number of 8" formats, but that doesn't mean there were products using those formats).
the 3,5in 360k format is unique to an Epson portable drive from the early 80s, made for CP/M systems. It was never supported by MSDOS.
the Wikipedia article lists all possible formats, not just those used in practice (although there were many odd and proprietary systems around in the heydays of the late 70s).
—Mellvik

mfld-fr:
Fully agree with @Mellvik : our legacy platform is the IBM PC / XT / AT, and currently most of the end users are asking for 5.25'' 360K, 3.5'' 720K or 3.5'' 1440K floppy images only.

In repair, the latest mainstream FDD is used (so 3.5'' 1440K), and original HDD are replaced by CF-based that still compatible with old BIOS'es, so the need for a better support of HD images.

Current targets are the embedded systems, so SD or ROM images.

ghaerr:
Sounds good, I agree also. I had originally created 320k for testing and thought it might be interesting to find out how large a completely minimal bootable ELKS image might be on a MINIX filesystem - turns out that's just over 160k.

I'll prepare a PR that removes the 320k format generations, and properly uses 2 heads for the 360k format. The resulting test images built will just be 5.25" 360k, 3.5" 720k, and 3.5" 1440k. An SD or HD image can easily be added for testing at a later time.

PR #300, opened 2020-02-05
[boot] rewrite & unify code for reading stage 2 booter & kernel

tkchia:
The unified code recalculates the physical &lt;cylinder, head, sector&gt; for each BIOS read call.  This simplifies and shrinks the code, at the expense of slowing it down a bit.

I also changed the logic somewhat, to make it more robust:
  * Reads that straddle multiple 64 KiB segments are now supported.
  * The code now tries to make sure that each individual read does not overshoot the end of an IA-16 segment.
  * Whenever a multi-sector read needs to be retried, the code will now try to re-read only the first sector (before continuing with the remaining sectors).

The boot sector code now stands at `0x1c0` bytes.

### DIFF
--- a/elkscmd/bootblocks/boot_sect.S
+++ b/elkscmd/bootblocks/boot_sect.S
@@ -80,20 +80,20 @@ floppy_table:
 	xor %sp,%sp
 
 	push %ax
-	mov $_next1,%ax
-	push %ax
+	mov $_next1,%cl  // CH = 0
+	push %cx
 	retf
 
 _next1:
+
+	// Save boot drive as provided by BIOS
+
+	mov %dl,drive_num  // DX preserved above
 
 	// Print header
 
 	mov $msg_boot,%bx
 	call _puts
-
-	// Save boot drive as provided by BIOS
-
-	mov %dl,drive_num  // DX preserved above
 
 	// Set sector count in floppy parameter table
 
@@ -102,31 +102,12 @@ _next1:
 
 	// Load the second sector of the boot block
 
-	// TODO: move the retry to the 'disk_read' function
-	// and reuse the same function for all reading
-
-	mov $3,%si         // number of tries = 3
-
-_loop1:
-
-	xor %ah,%ah        // reset the floppy drive
-	int $0x13
-
-	mov $0x0201,%ax    // read 1 sector
-	mov $payload,%bx   // destination (ES already set)
-	mov $2,%cx         // track 0 - from sector 2 (base 1)
-	cwtd               // head 0 (dh = 0)
-	mov drive_num,%dl  // boot drive
-	int $0x13          // BIOS disk services
-	jnc _next2
-
-	dec %si
-	jnz _loop1
-
-	mov $ERR_PAYLOAD,%al
-	jmp _except
-
-_next2:
+	mov $1,%ax
+	mov %ax,%dx
+	mov $payload,%cx
+	push %ss
+	mov $ERR_PAYLOAD,%bl
+	call _disk_read
 
 	call payload
 	mov $ERR_NO_SYSTEM,%al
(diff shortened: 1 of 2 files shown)


elkscmd/bootblocks/boot_sect.S
```diff
@@ -217,134 +198,136 @@ drive_reset:
...
 
-	mov -2(%bp),%ax
-	add 2(%bp),%ax
-	cmp sect_max,%ax
-	ja dr_over
-	mov 2(%bp),%ax
-	jmp dr_next1
+	mov -6+1(%bp),%bl
+	neg %bl
+	shr %bl
```
mfld-fr, line 251: Is this the instruction SHR BL,1 ? 8086/88 or 186/188 ?

elkscmd/bootblocks/boot_sect.S
```diff
@@ -498,57 +482,66 @@ msg_reboot:
 drive_num:
 	.byte 0
```
mfld-fr: I understood that the byte @1FDh is reserved for the fixed or dynamic boot drive number, so we could move `drive_num` to that location for consistency and to save... 1 more byte 😄 !

elkscmd/bootblocks/boot_sect.S
```diff
@@ -498,57 +482,66 @@ msg_reboot:
...
 // Disk geometry (CHS)
 
 	.global sect_max
 	.global head_max
 	.global track_max
 
-// #undef CONFIG_IMG_FD1680
-// #undef CONFIG_IMG_FD1200
+// TODO: number of tracks is not used
+
```
mfld-fr, line 494: Not used in boot sector, so we can move it to the probing.

ghaerr:
@tkchia: I have reviewed this, and it looks great. I will update the mfs boot-block writer accordingly with a PR after committed. This will allow our package manager floppy test disks to be tested in each of the three supported sizes, thereby producing all MINIX images with our own tools. If the new MSDOS boot can fit below this BPB and reserve the MSDOS BPB space, so much the better!

To document this new ELKS BPB, I have the following at decimal fixed offsets in the boot sector:

503 word track_max
505 byte sec_max
506 byte head_max
507 byte BPB length
508 word 'eL' BPB signature
510 word 0xAA55 boot signature

Does that look right?

Thank you!

elkscmd/bootblocks/boot_sect.S
tkchia: Yes, this is an alternate way in AT&amp;T syntax to write `shr $1,%bl`.

elkscmd/bootblocks/boot_sect.S
tkchia: Done. :)

tkchia:
Hello @ghaerr,

> 503 word track_max
> 505 byte sec_max
> 506 byte head_max
> 507 byte BPB length
> 508 word 'eL' BPB signature
> 510 word 0xAA55 boot signature

> Does that look right?

Thank you!  Yes, this is correct.

tkchia:
Hello @mfld-fr,

> Not used in boot sector, so we can move it to the probing.

Indeed, the boot sector does not need to know the track count.  However, I think such a field will allow the kernel (and other programs) to quickly find out the number of tracks on a disk, without having to go through a trial-and-error probing.  I hope to later modify the kernel to do so.

(But perhaps I am overthinking this...)

Thank you!

elkscmd/bootblocks/boot_sect.S
mfld-fr: Okay, but as this is not 8086/88 code, it won't run on IBM XT, right ?

elkscmd/bootblocks/boot_sect.S
mfld-fr: Finally, I agree that it could help the kernel to keep it in the boot sector.

elkscmd/bootblocks/boot_sect.S
tkchia: It is the same instruction as the 8088 `shr $1,%bl` (opcode `0xd0 0xeb`).  I just happened to write it in a different syntax at the source level.  Sorry for the confusion.

mfld-fr: Okay, I confused SHR BL,1 (8086) and SHR BL,imm8 (80186) 😕.

mfld-fr:
I like the inversion of CHS with logical sector to simplify the counting 😄.
Thanks for the improvement !

PR #301, opened 2020-02-05
 [image] Remove 320k image builds from package manager

ghaerr:
Removes 320k MINIX and FAT image builds from package manager, now just 360k, 720k ad 1440k built.

Corrects head_max to 2 on 360k MINIX BPB boot image (but still commented out, waiting for head_max MINIX BPB location)


mfld-fr:
I would advise to not work in your master branch, but in a development branch, to avoid such merge commits that floods the version tree. Please let your master branch just follow the upstream one, and then start your development branch after the synch:
```
git checkout master
git pull upstream
git checkout -b <dev_branch>
git push origin
```
And push your development branch in the PR.
Thanks.

ghaerr:
@mfld-fr: Thanks. The reason for the mess this time was that I made a change to image/Makefile and committed it to my repo before realizing that you had also changed it yesterday. For that reason, I had to revert the commit, pull from elks-org, and start over.

I am a very basic user with git and will update from elks-org before making a local commit, in an effort to not have this happen again. Thanks for the git instructions :)

ghaerr:
> I would advise to not work in your master branch, but in a development branch, to avoid such merge commits that floods the version tree. Please let your master branch just follow the upstream one, and then start your development branch after the synch:
> 
> ```
> git checkout master
> git pull upstream
> git checkout -b <dev_branch>
> git push origin
> ```

@mfld-fr: Stupid git question: What are the git commands and procedure to use when I would like to prepare multiple PR's, that each should have their own name and not use the same branch so as to not automatically get added to the PR's waiting for commit approval? I understand how to create new branches etc, but in some cases would like my previous branches included in my own work before the PR's are approved?

Does one just create sub-branches underneath each waiting PR branch and submit PR's from there?

That is:
dev_branch ->    1st PR
    subdev2_branch -> keeps 1st PR changes while I work on new PR that only includes subdev2
      ... etc

It's definitely easier for me to write code than it is to use git! 😂

mfld-fr:
Yes, exactly, as soon as you consider the proposed branch to be stable and ready for pull, you should start another branch from the stable point with `git checkout -b <sub-branch>`. Then you have the choice either to rebase with `git fetch -tp upstream; git rebase upstream/master`, or to merge back from master `git pull upstream master`. I prefer the first option to avoid such 'merge back' commits.

PR #305, opened 2020-02-06
[kernel] FAT filesystem rework : devfs & fixups

ghaerr:
MSDOSFS implementation of /dev is automatic if /dev is first or second root dir entry
	Renamed CONFIG_UMSDOS_FS to CONFIG_MSDOS_DEV
	Fixed various initialization problems with fsdev
	Reformatted elks/fs/msdos source code with no compilation output changes
	Updated disk image package list for 360k filesystems

@mfld-fr: I have completed the work on getting an internal devfs working for MSDOS FAT disks. It is based on a rewritten CONFIG_UMSDOS_FS (renamed to CONFIG_MSDOS_DEV) and tested with CONFIG_32BIT_INODES=n for now. It also requires setting CONFIG_MSDOS_DEV=y. I am having serious problems getting my build to work correctly since KConfig is broken and I am manually editing .config and include/autoconf.h, and don't have the means to setup the config.in files since I can't test them. The build is acting strangely after hand-editing these files to various values for testing. I will be opening various bug reports for both KConfig on OSX as well as MSDOS FS issues in general (I have found many, this PR however works great).

Providing we can get the configuration system working with a couple of new .config entries, we now have all all that is needed to boot ELKS on FAT images except for the boot loader.

Have fun!

### DIFF
--- a/elks/fs/msdos/dir.c
+++ b/elks/fs/msdos/dir.c
@@ -25,8 +25,7 @@ static int msdos_dir_read(inode, filp, buf, count)
     return -EISDIR;
 }
 
-static int msdos_readdir(struct inode *inode,
-			 struct file *filp,
+static int msdos_readdir(struct inode *inode, struct file *filp,
 			 char *dirent, filldir_t filldir);
 
 /*@-type@*/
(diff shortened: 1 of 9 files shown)


ghaerr:
Added a 4th commit which fixes major MSDOS filesystem consistency bug, seen immediately with creating a file then unmounting. FAT filesystem is now finally working well with exception of umount device busy failure on rare occasions. Current testing only done on FAT12 filesystems (current package manager only building FAT12 disk images). Will be testing FAT16 and FAT32 filesystems shortly. Kernel printk messages left in but commented out for the time being, they'll be needed to chase down rarer filesystem bugs in the next while, then cleaned up and removed.

mfld-fr:
Nice work !

Just a comment : when committing, put a short but meaningful title, then an empty line, then the text body. Many tools (including Git itself) expect less than 40~50 characters for the title and an empty line between the title and the body. Also do not forget to reference the issue in the body, if both are linked.

Like:
```
[keyword] Wash the dirty code

To make it more readable.

Issue #NNN
```

PR #308, opened 2020-02-07
[kernel] use new "eL" parameter block to get floppy disk geometry

tkchia:


ghaerr:
@tkchia: This PR clears up some questions I have had about ELKS boot, floppy and HD geometries, but I still have the following questions (not yet fully understanding our boot/kernel method for determining floppy and HD drive configurations):

o Are we now getting rid of the CONFIG_HW_USE_INT13_FOR_DISKPARMS option since the new method is to have the boot sector specify them?  I guess that would imply that if the boot sector is stamped with the wrong geometry then the boot load will fail and the kernel never execute.

o We will almost always likely use your new boot sector, which means that the kernel probing currently left in will never get executed for boot disks, only for inserted disks, since your boot sector will always have an ELKS BPB, correct? Or does this try to read boot sectors of inserted non-boot disks also?

o This kernel code is in doshd.c, how are we determining the HD drive parameters for non-boot HD's that are later mounted, will that use the kernel probe method, or do all HD disks need an ELKS boot block (primarily used for ELKS BPB geometry)?

I like the idea of getting rid of (using boot sector instead) BIOS calls in the kernel to determine geometry, unless those calls might still be needed on disks inserted without boot sectors. There is also the issue that this block driver will be used to determine geometry of a FAT disk inserted, where there will be no ELKS BPB.

tkchia:
Hello @ghaerr,

> Or does this try to read boot sectors of inserted non-boot disks also?

My patch should work as well with non-boot floppy disks which do not have an ELKS parameter block.  If it cannot read the first sector, or if the first sector does not have an ELKS BPB, it will just fall back on a trial-and-error probe, which should still work.

In the case of a hard disk --- from what I see --- the disk geometry is obtained by `setupw(.)` through the `SETUP_DATA` segment.  `SETUP_DATA` is in turn presumably set up by `elks/arch/i86/boot/setup.S` at boot time; it invokes the BIOS to get the geometry.

> Are we now getting rid of the `CONFIG_HW_USE_INT13_FOR_DISKPARMS` option since the new method is to have the boot sector specify them?

Hmm... yes, it does look like `CONFIG_HW_USE_INT13_FOR_DISKPARMS` is now quite useless (unless there is a plan for obtaining hard disk geometries in `setup.S` by non-BIOS means?).

Thank you!

mfld-fr:
@tkchia : that change will reduce the probing failure rate that is experienced on weak BIOSes, thanks !

PR #309, opened 2020-02-07
[tools] Update mfs boot -B option with new ELKS BPB locations

ghaerr:
Also temporarily try mounting FAT filesystem on /dev/fd1 at boot for testing.

This has been tested on QEMU to work correctly on 1440k disks, but fails with 360 and 720 floppies as they will require @tkchia's recent kernel ELKS BPB commit to read the disk geometry using this new mechanism.

Note: The MFS -B option does NOT currently set max_tracks. Should the kernel start using track_max, then the -B parameters will have to include a track_max, and mfs updated as well.

### DIFF
--- a/elks/tools/mfs/mkfs.c
+++ b/elks/tools/mfs/mkfs.c
@@ -113,9 +113,6 @@ void cmd_mkfs(char *filename, int argc,char **argv) {
 #define ELKS_BPB_SecPerTrk	505		/* offset of sectors per track (byte)*/
 #define ELKS_BPB_NumHeads	506		/* offset of number of heads (byte)*/
 
-/* MSDOS FAT parameters (unused at present)*/
-#define FAT_BPB_SecPerTrk	24		/* offset of sectors per track (short)*/
-#define FAT_BPB_NumHeads	26		/* offset of number of heads (short)*/
 /**
  * Write boot block to image file
  */
--- a/elks/tools/mfs/mkfs.c
+++ b/elks/tools/mfs/mkfs.c
@@ -109,8 +109,13 @@ void cmd_mkfs(char *filename, int argc,char **argv) {
   close_fs(fs);
 }
 
-#define BPB_SecPerTrk	24		/* offset of sectors per track (short)*/
-#define BPB_NumHeads	26		/* offset of number of heads (short)*/
+/* ELKS BPB parameters*/
+#define ELKS_BPB_SecPerTrk	505		/* offset of sectors per track (byte)*/
+#define ELKS_BPB_NumHeads	506		/* offset of number of heads (byte)*/
+
+/* MSDOS FAT parameters (unused at present)*/
+#define FAT_BPB_SecPerTrk	24		/* offset of sectors per track (short)*/
+#define FAT_BPB_NumHeads	26		/* offset of number of heads (short)*/
 /**
  * Write boot block to image file
  */
@@ -153,8 +158,8 @@ void cmd_boot(char *filename, int argc,char **argv) {
 		fprintf(stderr, "%s warning: may not be valid boot block\n", argv[1]);
   
 	if (opt_updatebpb) {	/* update BPB before writing*/
-		blk[BPB_SecPerTrk] = (unsigned char)SecPerTrk;
-		blk[BPB_NumHeads] = (unsigned char)NumHeads;
+		blk[ELKS_BPB_SecPerTrk] = (unsigned char)SecPerTrk;
+		blk[ELKS_BPB_NumHeads] = (unsigned char)NumHeads;
 	}
 	if (fwrite(blk,1,count,ofp) != count) die("fwrite(%s)", argv[1]);
 	fclose(ofp);
@@ -165,8 +170,8 @@ void cmd_boot(char *filename, int argc,char **argv) {
 
 	count = fread(blk,1,512,ofp);
 	if (count != 512) die("fread(%s)", filename);
-	blk[BPB_SecPerTrk] = (unsigned char)SecPerTrk;
-	blk[BPB_NumHeads] = (unsigned char)NumHeads;
+	blk[ELKS_BPB_SecPerTrk] = (unsigned char)SecPerTrk;
+	blk[ELKS_BPB_NumHeads] = (unsigned char)NumHeads;
 	if (fseek(ofp, 0L, SEEK_SET) != 0)
 		die("fseek(%s)", filename);
 	if (fwrite(blk,1,512,ofp) != 512) die("fwrite(%s)", filename);
--- a/elkscmd/rootfs_template/etc/rc.d/rc.sysinit
+++ b/elkscmd/rootfs_template/etc/rc.d/rc.sysinit
@@ -42,11 +42,11 @@ then
 fi
 
 #try to mount FAT file system disk
-if test /dev/fd1
+if test /dev/bda1
 then
   echo
   echo -n "Mounting FAT file system: "
-  mount -t msdos /dev/fd1 /mnt
+  mount -t msdos /dev/bda1 /mnt
 fi
 
 # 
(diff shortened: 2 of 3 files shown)


elks/tools/mfs/mkfs.c
```diff
@@ -109,8 +109,13 @@ void cmd_mkfs(char *filename, int argc,char **argv) {
...
 
-#define BPB_SecPerTrk	24		/* offset of sectors per track (short)*/
-#define BPB_NumHeads	26		/* offset of number of heads (short)*/
+/* ELKS BPB parameters*/
+#define ELKS_BPB_SecPerTrk	505		/* offset of sectors per track (byte)*/
+#define ELKS_BPB_NumHeads	506		/* offset of number of heads (byte)*/
+
+/* MSDOS FAT parameters (unused at present)*/
+#define FAT_BPB_SecPerTrk	24		/* offset of sectors per track (short)*/
+#define FAT_BPB_NumHeads	26		/* offset of number of heads (short)*/
```
mfld-fr: These changes do not meet the acceptance criteria as defined in https://github.com/elks-org/elks/issues/296#issuecomment-582372862 : `mfs` is a tool for working with the MINIX filesystem, the ELKS specific code related to the boot sector has to be moved to another tool.

elks/tools/mfs/mkfs.c
```diff
@@ -109,8 +109,13 @@ void cmd_mkfs(char *filename, int argc,char **argv) {
   close_fs(fs);
 }
 
-#define BPB_SecPerTrk	24		/* offset of sectors per track (short)*/
-#define BPB_NumHeads	26		/* offset of number of heads (short)*/
+/* ELKS BPB parameters*/
+#define ELKS_BPB_SecPerTrk	505		/* offset of sectors per track (byte)*/
+#define ELKS_BPB_NumHeads	506		/* offset of number of heads (byte)*/
+
+/* MSDOS FAT parameters (unused at present)*/
```
mfld-fr: What FAT parameters are doing in a tool for working with the MINIX filesystem ?

mfld-fr (changes requested):

elks/tools/mfs/mkfs.c
ghaerr: @mfld-fr: We're using the mfs tools for writing the boot sector, and have been ever since we removed sudo from the build. The "mfs boot -B<sector_max><head_max> option was added during the discussion of adding BPB's to boot sectors, and mfs is always used to create the MINIX boot block for ELKS. While I understand what you're saying, I don't see the need to write a completely new tool just to stamp an ELKS BPB when mfs is used to write the boot image anyways.

elks/tools/mfs/mkfs.c
ghaerr: The FAT parameters were added back when were were talking about a combined MINIX and FAT ELKS BPB boot block. They are unused, would you like me to delete them?

elks/tools/mfs/mkfs.c
ghaerr: Another reason that mfs was chosen to modify the boot sector sectors and heads is that this is used in the package manager to quickly create the 360k, 720k and 1440k disk images without modifying the .config build images. That is, with a separate tool, one would have to copy the build boot block, modify it by the new tool, then replace the $(FDSECT) with the changed version, then use mfs boot to write it. It seems far simpler to just have mfs modify it while reading the .config boot image from $(DESTINATION).

mfld-fr: I accepted the previous changes linked to the generation of the images, because you explained to me it was blocking you from moving forward, and convinced me that such images are strongly expected from the end users (also because @jbruchon asked me to think a bit more to them as a testimonial), despite IMO you are taking the wrong path in term of long term sustainable design. Now that this 'urgent' need is fullfilled, it is time to come back to the right things, and to avoid shortcuts that caused us (= the maintainers) much reworking and cleanup in the last 5 years. Please follow the guidelines I already posted. Thanks !

ghaerr: @mfld-fr: You are right and I agree! (It's just not quite time yet considering test needs).

I did say that I wanted to have the package manager accepted because it would help me *quickly* get things moved forward. It has helped me immensely in getting the FAT filesystem code fixed very quickly in the last two days, and all the other work I've been committing. I always test on all formats for the current code submissions.

I did agree that the entire package manager concept could be temporary (I still think it is very valuable for developer and end user testing though).

I added the mfs boot -B option, as a favor for your and @tkchia's very-much-changing of the boot blocks during the last weeks of development, to enable me to test that work. Who else is going to test it for all the disk formats supported as we move rapidly? 

Manually running make menuconfig for each image type just to test the new daily boot BPB sector code is just too time consuming and menuconfig doesn't even run on my machine yet (thanks for your help on ncurses, I'll try that).

The mfs boot -B option is NOT used in the build, only in the package manager that allows quick generation of all disk images (currently six). No sense in writing special new tools for updating the boot block properly, etc. if the entire package manager will be thrown away at some point.

So - I agree with you, but we're still heavy in development adding fantastic new features to ELKS, and we need quick testing! Now is not the time to refuse small modifications that enable testing of what everyone is creating daily, the package manager isn't in fact one of the main points of the 0.3.1, although the MSDOS/FAT image building process will not work without it. I still want to have ELKS FAT image boot make it into 0.3.1.

I suppose, realistically, that the entire package manager and the newly created mfs options to make it work are really only tools that I care about, including the -B option, so that I can move ELKS forward rapidly. I have appreciated your enabling my contributions by including it.

Thank you for listening. I don't have the time to write temp tools for boot block code from scratch. I just removed all FAT code from mfs on this PR with an update. If you don't agree, just delete this PR and the package management system will still produce images, they just won't be bootable with the latest kernel commits, except for the image specified in .config (as always).

ghaerr: @mfld-fr and @tkchia:

What about the knl program for modifying the ELKS MINIX boot sector? I was reading it's manual page where it says:

```DESCRIPTION
       In a bootable image for the Linux kernel, there are several pairs of bytes
       which specify various kernel options, located at the following offsets:

               498  1F2  Root flags
               500  1F4  Reserved
               502  1F6  Unused (was Swap Device)
               504  1F8  RAM Disk Parameters
               506  1FA  VGA Mode
               508  1FC  Root Device
               510  1FE  Boot Signature

       knl will change or display these values.
```
Sounds like our version is quite out of date, (since we don't use any of these values from boot in our kernel, or am I wrong?). The elkscmd/sys_utils/knl.c utility could probably be easily converted to be built and run on the host system and updated to use our current boot sector, in which case it could be used in the package manager to modify the boot block, then use mfs boot to write the sector to the MINIX image.

This would involve running knl on the $TARGET boot sector, then copied by mfs boot to the disk image being built.

Onto another problem that possibly knl could also be used for: I was looking for the ability to compile in DEBUG statements into the FAT filesystem driver that could be turned on or off at runtime - isn't this what knl is really for? It's already on each image. The idea would be to produce an image that had full debug capability in FAT, but then only turn on the debug printk output when finding an image that exhibits the FAT crash bug, which could all be done without recompilation adding code, which usually stops the symptoms. I just need to know how knl would communicate with the running kernel to switch on a variable the FAT filesystem could read.

mfld-fr: The `knl` command relies on the (past) fact that many kernel parameters are located in the 'raw' boot sector, that is copied to the INITSEG. I am progressively moving from that deprecated design to the Linux current one, where the parameters are passed by the boot loader to the kernel command line. For ELKS, such textual command line would be too costly, so it will be as binary parameters on the 'setup' and 'system' initial stack.

elks/tools/mfs/mkfs.c
mfld-fr: Yes please.

mfld-fr:
Superseeded by #312.

PR #310, opened 2020-02-07
[busyelks] Bunch of commits from mailing list

mfld-fr:
These are 20 patches posted by Marcin to the ELKS mailing list.
Opened this PR in order for him to verify that I correctly committed these patches.


ghaerr:
This looks potentially fantastic! We need some better userland command implementations, I've been running into a number of issues already testing FAT with poor implementations of rm, ls, mount and umount.

Will the plan be to add each and every one of these into .config as an option for selection to be included in the image? This will be messy, but could work. There will definitely be immediate issues because we're already almost completely out of space with the "default" 1440k built image, and several nano-X applications have been removed from even compiling since we don't have application granularity in the .config file, only by also editing various elksmd/*/Makefiles.

Or would we want an updated package manager images/Packages list? Providing that busy elks completely installed into a separate, say /usr/local/bin directory on the template, then all of these commands could be built every time and easily selected for inclusion in a Packages file based on disk size or a differently named output disk image based on size. We would also want to ensure the shell PATH environment included the new directory, a small mod.

mfld-fr:
And now 16 more commits, again from Marcin. Wow !

marcin-laszewski (approved):

mfld-fr:
Thanks @marcin-laszewski !

marcin-laszewski (approved):

PR #311, opened 2020-02-07
[fat] More FAT filesystem cleanup

ghaerr:
Reduce kernel data size by cleaning up and standardizing printk messages
	Remove all forced kernel panics from badly formatted fat images, printk instead
	Clean up source and compiler output warnings
	Change rc.sysinit to check mount /dev/fd1 rather than /dev/bda1 for easier testing


mfld-fr:
Good cleanup, thanks !

PR #312, opened 2020-02-08
[tools] Create new setboot tool to modify ELKS boot block parameters

ghaerr:
see elks/tools/setboot/setboot.c for usage examples
	package manager updated to use setboot properly, mfs boot usage removed

After looking at both elkscmd/sys_utils/knl.c and rdev.c (which are ancient, outdated, and ugly), as well as reflecting on @mfld-fr's statements about getting our standard filesystem utilities out of ELKS-specific business, I just decided to write the dang tool for the sake of keeping the package manager building correct images. My previous PR #309 can be closed or deleted.

This PR updates the package manager to run setboot, which writes the proper sectors and heads in each of the 360, 720 and 1440k MINIX floppy disk builds, of which I have verified the boot sector is modified correctly. However, QEMU testing boot images on the 360 and 720 images fail with the kernel still thinking that /dev/fd0 has 18 sectors. I'll leave that debugging to those that are writing the boot sector and kernel floppy driver probe mods.

### DIFF
--- a/elks/tools/Makefile
+++ b/elks/tools/Makefile
@@ -30,6 +30,7 @@ all:
 	$(MAKE) -C mfs all
 	$(MAKE) -C mfsck all
 	$(MAKE) -C mkromfs all
+	$(MAKE) -C setboot all
 
 
 #########################################################################
--- a/elks/tools/setboot/Makefile
+++ b/elks/tools/setboot/Makefile
@@ -1,0 +1,51 @@
+# setboot - modify ELKS boot block parameters
+#
+#########################################################################
+#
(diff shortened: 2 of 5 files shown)


mfld-fr (approved):
Fine ! I am happy to see you are accepting a little more workload for a lesser mixed design :smiley: 

Just one remark before merging this one : option `boot` shall be kept in the `mfs` tool, because inserting a 'boot block', whatever the contents of this block, is specific to the MINIX filesystem. But without any 'stamping' of the 'disk sector' with disk parameters, because it has no knowledge of the 'boot block' contents. Just that the first block in MINIX is the 'boot' one...

PR #314, opened 2020-02-08
[config] Remove now-unused "Use INT 13h for disk parameters" option

tkchia:


mfld-fr (changes requested):
I am not sure it is a good idea to remove that option, even now unused.

In case of a PC AT, we can query the BIOS for FDD geometry capabilities, and keeping in mind that such capabilities are distinct from the actual disk geometry, it is a way to guess that latest if reading the boot sector returned no information, before falling back to default.

Let us take the example of a raw floppy, or unsupported format. No ELKS boot parameters, no FAT so no BPB. There is a big chance that the geometry used to write / format the disk is the same as the drive capabilities. If the user selects that option, it means "same disk geometry as the drive". If not, then the user would have to set the CHS parameters in the configuration.

At least that option is to be moved above the floppy & HD menus, because for a HD, some BIOSes (without a CHS menu) are known to report incorrect geometry, and again the user has to opt out and set the CHS in the configuration.

tkchia:
Hello @mfld-fr,

Well, it looks like we need to rethink the role of BIOS "get drive parameter" (`int $0x13, %ah = 8`) calls then.

May I know what you see as the role of such "get drive parameter" calls for floppy disks?  With my recent patches, the kernel will try to size up a floppy
  * by reading from its ELKS parameter block, if it can; then
  * by doing a trial-and-error probe at various track and sector numbers.

I guess some possibilities are to call `int $0x13` with `%ah = 8`,
  * if both the above methods fail to guess the correct geometry, or
  * whenever a disk has no ELKS parameter block (so we skip the trial-and-error probe entirely).

But perhaps you have other ideas.

As for hard disks --- from what I can tell, `elks/arch/i86/boot/setup.S` will always use BIOS calls to obtain hard disk geometries, with or without the "Use INT 13h for disk parameters" options.  Is it your intention to allow the user to override the BIOS's geometry at this point?

Do let me know your thoughts.

Thank you!

mfld-fr:
I don't like the trial-and-error probing, as we got several users reporting it was failing on their systems. ELKS has to be more robust on that point, so it is up to the user to describe his system (and the inserted removable media), or to select a **safe** system facility to get that description.

For HDD, yes, we have to make the INT13 AH=8 optional to solve some issues on weird BIOSes before the 'BIOS HDD CHS menu' became a standard.

mfld-fr:
After a bit more thinking, we could also optionalize the trial-and-error probing, and / or move it to the 'probing boot payload' to help to user to find out the correct CHS values.

ghaerr:
> Let us take the example of a raw floppy, or unsupported format. No ELKS boot parameters, no FAT so no BPB. There is a big chance that the geometry used to write / format the disk is the same as the drive capabilities. If the user selects that option, it means "same disk geometry as the drive". If not, then the user would have to set the CHS parameters in the configuration.

@tkchia and @mfld-fr: 

This brings up another issue coming up shortly: what are we going to do about FAT floppies (which by specification *must* have a properly formatted boot sector FAT BPB) inserted? These floppies won't have an ELKS BPB. Will the floppy driver try to read the disk geometry from the FAT BPB after verifying a FAT/BOOT signature, just as @tkchia is doing now for media that has a signature-verified ELKS BPB? It looks like the floppy driver may have to get in the business of knowing about both boot and filesystem types, the way this is going

Also, when we write the FAT boot sector, of course the FAT BPB will be reserved, for the mformat stamping tool to use when formatting. Will there also be an ELKS BPB at the top? If so, this brings open the opportunity for having mismatched head/sector info, unless also requiring double stamping, which I think is a bad idea because of duplicate fields.

If the kernel is going to get into the business of reading ELKS BPB information, I think it also needs to be prepared to read FAT BPB as well. I agree with @mfld-fr and don't like probing, since it varies tremendously per emulator and hardware. Supporting both ELKS AND FAT BPBs in the kernel would allow using probing as a last resort.

ghaerr:
Another quick comment: 

Putting an ELKS BPB in our FAT boot sector code is a bad idea, since that then makes a difference between FAT floppies from the wild, which will always have a valid FAT BPB, and ELKS-bootable FAT floppies, which would have an ELKS BPB and FAT BPB. If a previous ELKS-bootable floppy were modified/formatted by a standard FAT filesystem tool by a user, it would update the FAT BPB but not the ELKS BPB. 

Thus, it would seem that if the kernel will be reading ELKS BPB/signatures, it will also have to read FAT BPB/signatures, in order to most safely determine geometry, likely as the first option in @tkchia's proposed process above, before probing and then possibly INT 13.

mfld-fr:
@ghaerr : no design problem actually, because a filesystem driver is able to set back some parameters in a disk / volume driver. FAT driver does not need the CHS to be correct when requesting to read sector 0.

This ability for the filesystem driver to set some parameter at runtime is also needed for best performance. For example, when the filesystem block size is larger than the disk sector size, to instruct the disk driver to load the whole block at a time.

Nothing new here, this is how the big brother is working, and ELKS has to follow :smile:

ghaerr:
> After a bit more thinking, we could also optionalize the trial-and-error probing, and / or move it to the 'probing boot payload' to help to user to find out the correct CHS values.

I think the idea of providing a "probe" boot/disk for users (possibly provided as a bootable disk image fd-probe.bin as part of the build, easily done with our new tools) that displays not only probe results but also displays what the kernel floppy algorithms would use to determine geometry would go a long way to helping debug our new system as well as user's old hardware variations.

tkchia:

Thanks!  I think I will cancel this pull request for now, and see how I can repurpose the "get drive parameter" call.

(For future reference: [the commit in the pull request](https://github.com/elks-org/elks/commit/9e37ffbeabb6d918109d3f1eb4093454a75505f9).)

Mellvik:
In scanning this thread - and most likely missing important points - I have to ask:

Most physical floppy drives support two or more formats. 3,5" drives support 720k and 1,44M, some even 2.88M, while all 5.25" drives that support 1.2M also read and write 360k floppies. So it has been my impression that the BIOS probing was the easy way to figure out what's actually in the drive.

My concern would be to ensure we keep this flexibility with ELKS. The format determination happens before boot if booting from floppy, and the BIOS will correctly (?) report the actual format. If booting from HD or mounting a floppy in a 2nd drive, the format determination process must be repeated. Is ELKS now taking over this responsibility, or am I just reading the thread too fast?

Mellvik

tkchia:
Hello @Mellvik,

From what I can see, the BIOS "get drive parameter" call (`int $0x13, %ah = 8`) only returns the floppy _drive_ capabilities, not the _disk_'s geometry.  So if a 1.44 MiB drive has a 720 KiB disk inserted in it, the BIOS call will still say 1.44 MiB.

Thank you!

PR #315, opened 2020-02-08
[fat] Add fsdebug() macro for FAT filesystem debugging

ghaerr:
Added varargs-capable fsdebug() macro in FAT code to eliminate the need to comment in and out printk for debugging. Default OFF but set to ON when #define FSDEBUG 1 set (probably in include/linuxmt/debug.h).

Fix kernel compile when #define DEBUG 1 set in debug.h.

The ELKS kernel is too large to link when built with DEBUG=1. In addition, the entire kernel source uses an extremely outdated method of debug(), debug1(), debug2() macros that specify how many extra arguments printk needs, all designed prior to GCC implementing the varargs-capable #define debug(...) mechanism in C.

It would be a bit of typing, but I propose that an improved method of kernel printk debugging be written, which does away with all the cruddy debugX() macros that require an explicit argument count, only to be replaced with a debug macro that uses the current kernel (set by a user program)  "debug level" that allows deeper levels of detail to be produced. A set of macros (same name, different meaning, I know) like debug1(), debug2(), debug3() would printk when the current debug level was set to 1, 2 or 3 etc., respectively. The current debug method is entirely unusable since the kernel won't even build with it turned on. Thoughts?

This PR uses a temporary fsdebug() macro to allow the nasty last known FAT bug to be tracked, without having to rewrite all the kernel debug statements.

On the business of debugging this last FAT problem, I am starting to think that is is not a straight-forward buffer overrun into data or code space, but instead, something to do with kernel lock timing, or corrupted kernel data as a result of not-working or unimplemented required locking in the FAT code. The reason for this is that I have *never* seen the bug when kernel printk is turned on, whose effect would be to massively slow down and change the kernel timing. With printk's off, an occasional hang is seen.

@mfld-fr: I need a mechanism to turn on and off fsdebug from userland. Thanks for your `knl` comments, after looking at its source and your direction, I think we should just remove it, as it doesn't work anyways. We still need a program and a way to communicate to some kernel global variable, or u-space structure that could allow fsdebug to be turned on or off... this will allow us to get closer to the last remaining FAT bug. Please let me know what you think.

### DIFF
--- a/elks/fs/inode.c
+++ b/elks/fs/inode.c
@@ -349,7 +349,7 @@ struct inode *__iget(struct super_block *sb,
 	} while ((inode = inode->i_prev) != NULL);
     } while (n_ino == NULL);
     inode = n_ino;			/* Inode not found, use the new structure */
-    debug1("iget: got one... (%x)!\n", empty);
+    debug("iget: got one...\n");
 
     inode->i_sb = sb;
     inode->i_dev = sb->s_dev;
--- a/elks/fs/minix/bitmap.c
+++ b/elks/fs/minix/bitmap.c
@@ -169,7 +169,7 @@ void minix_free_inode(register struct inode *inode)
     else {
 	map_buffer(bh);
 	if (!clear_bit((unsigned int) ((unsigned int)inode->i_ino & 8191), bh->b_data)) {
-	    debug1("%s: bit %ld already cleared.\n",ino);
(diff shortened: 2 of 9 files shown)


mfld-fr:
About communication between the kernel and the user land : the current mechanism is to perform 'ioctl' on special devices, like /dev/mem. Not as fancy as modern ones in Linux, but it makes the job while keeping the things simple.

PR #316, opened 2020-02-09
[boot] Tweak makefile to honour $(ASFLAGS) and further shrink boot sector code

tkchia:
The size of the boot code in the first sector now stands at `0x1b1` bytes.


mfld-fr:
Should I understand through your PR's that you are making room for FAT BPB 😉 ?

tkchia:
Hello @mfld-fr,

Yes, that is so. :-)  I will say a bit more at the relevant thread (https://github.com/elks-org/elks/issues/296).  Thank you!

PR #318, opened 2020-02-09
[fat] Fix longstanding system crash on FAT directory write

ghaerr:
remove non-working backup FAT table code and increase speed greatly on cluster adds
	heavy cleanup and some refactoring of FAT filesystem code for speed and size
	add fsdebug macro for filesystem printk debugging, set with FSDEBUG=1
	add check for disk too large on FAT16/32 when CONFIG_32BIT_INODES off
	Remove unneeded CONFIG_32BIT_INODES #ifdefs
	create fd2880-fat.bin in package manager for FAT16 testing

No more known bugs in FAT filesystem. Ready for user testing.

### DIFF
--- a/elks/fs/inode.c
+++ b/elks/fs/inode.c
@@ -169,12 +169,8 @@ static void list_inode_status(void)
     register struct inode *inode = inode_llru;
 
     do {
-#ifdef CONFIG_32BIT_INODES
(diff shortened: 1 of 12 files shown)


mfld-fr:
Nice milestone ! Only missing some feedback from the HW target users...

PR #319, opened 2020-02-10
[boot] Attempt to fit a /linux bootloader for FAT12/16 within 512 bytes

tkchia:


mfld-fr:
Okay, let me take it as a first step toward the final design I am expecting. Thanks !

ghaerr:
@tkchia and @mfld-fr:

Congratulations! A new baby is born! And the child is FAT :)

I ran `setboot fd1440-fat.bin elkscmd/bootblocks/fat.bin` and ELKS booted from FAT for the first time!!! Well done @tkchia!

The kernel initializes as expected, then the system fails during what appears to be /bin/init initialization. See screenshot. I will look into all this later today, but haven't been able to find where the string "init panic" comes from yet.

Screen Shot 2020-02-10 at 8 51 27 AM

ghaerr:
I've located the 'init panic' message, the full FAT boot sequence is failing in /bin/init, which uses the same message when it can't find certain /dev entries, can't open /etc/inittab, and various other issues, possibly including the /etc/rc.d/rc.sysinit filename issue mentioned earlier.

The system is running after the login: prompt above, but fails login because it can't find /bin/sh, which should be on the boot disk. So things look good, but there's some issue with the filesystem at boot.

I'm going to enhance the /bin/init error messages and will look more into the entire system startup to fix this later today. I feel once we find the basic problem here, ELKS FAT boot will be functional for testing :)

tkchia:

Thanks too to both of you for your hard work!

tkchia:
Hello @ghaerr,

> I feel once we find the basic problem here, ELKS FAT boot will be functional for testing :)

:-)

PR #321, opened 2020-02-10
[kernel] Fix: `head -1 /dev/fd1' gave /dev/fd0 contents instead

### LINKED ISSUE #313: Reading from /dev/fd1 yields contents of /dev/fd0 ?
I ran ELKS under QEMU, with a 1.44 MiB disk image for ELKS in drive A:, and a 720 KiB FAT12 disk image in drive B:.

Dumping the contents of `/dev/fd0` and `/dev/fd1` in ELKS yields identical results for some reason, even though ELKS correctly figures out that their drive geometries and filesystems are different:

Thank you!

tkchia:
This should fix https://github.com/elks-org/elks/issues/313 .

### DIFF
--- a/elks/fs/block_dev.c
+++ b/elks/fs/block_dev.c
@@ -67,7 +67,7 @@ size_t block_read(struct inode *inode, register struct file *filp,
 	if (inode->i_op->getblk)
 	    bh = inode->i_op->getblk(inode, (block_t)chars, 0);
 	else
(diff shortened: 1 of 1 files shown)


ghaerr:
Nice catch @tkchia! Considering that the longstanding FAT hang bug was also a miscoding on the use of a kdev_t type, I wonder how many other places in the kernel this sort of thing may still be lurking?!

PR #325, opened 2020-02-11
[boot] Fix build for ROMFS or hard disk

### LINKED ISSUE #323: Build broken by latest FAT boot sector
Some build failures while maturing the latest changes:
* when CONFIG_IMG_HD selected from default configuration (reported by Paul on the mailing list)
* using 'config-emu86' as tested configuration

Also what I think a design defect: duplicated FAT BPB in boot sector code and as generated by 'mformat'.

tkchia:
This should fix https://github.com/elks-org/elks/issues/323

### DIFF
--- a/elkscmd/bootblocks/boot_sect_fat.h
+++ b/elkscmd/bootblocks/boot_sect_fat.h
@@ -37,8 +37,42 @@
 	.set FAT_TABLE_SIZE, 9
 	.set FAT_SEC_PER_TRK, 18
 	.set FAT_NUM_HEADS, 2
+#elif defined(CONFIG_IMG_HD)
+	.set FAT_ROOT_ENT_CNT, 512
+	.set FAT_TOT_SEC, 2 * CONFIG_IMG_BLOCKS
+// Default cluster sizes for various partition sizes --- see Microsoft's 2000
+// white paper (reference below), p. 20
+.if FAT_TOT_SEC < 262144
+	.set FAT_SEC_PER_CLUS, 2
+.elseif FAT_TOT_SEC < 524288
+	.set FAT_SEC_PER_CLUS, 4
+.elseif FAT_TOT_SEC < 1048576
+	.set FAT_SEC_PER_CLUS, 8
+.elseif FAT_TOT_SEC < 2097152
+	.set FAT_SEC_PER_CLUS, 16
+.elseif FAT_TOT_SEC < 4194304
+	.set FAT_SEC_PER_CLUS, 32
+.else
+	.set FAT_SEC_PER_CLUS, 64
+.endif
+	.set FAT_MEDIA_BYTE, 0xf8
+// Default file allocation table size --- see 2000 white paper (again), p. 21
+	.set FAT_TMP_VAL_1, \
+	     FAT_TOT_SEC - 1 - (FAT_ROOT_ENT_CNT * 32 + 511) / 512
+	.set FAT_TMP_VAL_2, 256 * FAT_SEC_PER_CLUS + 2
+	.set FAT_TABLE_SIZE, \
+	     (FAT_TMP_VAL_1 + FAT_TMP_VAL_2 - 1) / FAT_TMP_VAL_2
+	.set FAT_SEC_PER_TRK, CONFIG_IMG_SECT
+	.set FAT_NUM_HEADS, CONFIG_IMG_HEAD
 #else
-	.error "Unknown disk medium!"
+	.warning "Unknown disk medium!"
+	.set FAT_SEC_PER_CLUS, 0
+	.set FAT_ROOT_ENT_CNT, 0
+	.set FAT_TOT_SEC, 0
+	.set FAT_MEDIA_BYTE, 0
+	.set FAT_TABLE_SIZE, 0
+	.set FAT_SEC_PER_TRK, 0
+	.set FAT_NUM_HEADS, 0
 #endif
 
 // BIOS Parameter Block (BPB) for FAT filesystems, as laid out in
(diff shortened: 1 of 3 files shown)


elkscmd/bootblocks/boot_sect_fat.h
```diff
@@ -37,8 +37,42 @@
 	.set FAT_TABLE_SIZE, 9
 	.set FAT_SEC_PER_TRK, 18
 	.set FAT_NUM_HEADS, 2
+#elif defined(CONFIG_IMG_HD)
```
mfld-fr, line 40: There is something I still don't understand... why do you maintain a FAT BPB here, while the `mformat` tool already set that BPB in the image, and could be with different values ?

ghaerr: I am guessing that the reason is having the values here allow fat.bin to be used without modification for booting from the floppy disk type specified in .config.

Currently, setboot is only run by the package manager, and the default build does not build a FAT bootable ELKS disk.

In any case, it has been useful for the time being. I hope to finish the setboot mods later today that will include an option to NOT copy a specified section of the FAT BPB from the the passed boot sector to the final volume image. The option will allow a size parameter so that enhanced FAT BPB's can be accommodated in the future.

With regards to 'different values', I presume that the subsequently run setboot -B<sectors>,<heads> command to make a FAT bootable volume will copy sectors and heads to BOTH the FAT BPB as well as the ELKS PB, thus ensuring they are the same, correct?

ghaerr: After the setboot modification described above is completed, these FAT values could be removed and just reserved and set to 0 for simplicity, provided that the fat.bin boot block will never be used (e.g. directly substituted in an existing formatted MSDOS disk) without the mformat tool when creating FAT volumes.

Currently, neither mformat or setboot are strictly required for using fat.bin, provided that it be able to be written to an already-formatted FAT disk that matches the .config build configuration. Since the ELKS FAT bootable image creating process is unlikely to not use mformat or setboot, these values could default to 0 as described above for simplicity, after the setboot enhancement described above.

mfld-fr: Currently there is either FAT BPB or ELKS BPB, but not both, in the boot sector. If the FAT boot sector is first installed in the disk image with `setboot <boot sector>` (without any care of the previous contents), then that image is processed by the `mformat` tool that will patch only the FAT BPB as I understand, why having to fill the FAT BPB in this code (if the `setboot` tool is also invoked in the image build with `.config`, of course) ?

ghaerr: Agreed, the FAT BPB could/should just be reserved and set to 0 for the scenario you just mentioned. However, the mformat tool can NOT currently be used with an intact current fat.bin file, since mformat ALWAYS writes an extended FAT BPB, thus overwriting @tkchia's boot code and the boot fails.

That fix requires that more space be reserve in the boot block BPB, up to offset 61. @tkchia, do we have the space?

Thus, to reduce complexity, I suggest l leaving this in the commit for now, and then having it removed AFTER we get the mformat/setboot problems etc resolved.

mfld-fr: I think it is time to insert the FAT case in the `image/Makefile` to avoid such FAT BPB duplication...

ghaerr: I am confused by the statement that we have either the FAT BPB or the ELKS BPB, but not both. For some reason, my code inspection showed that both were included, I might be wrong.

If it's only one or the other, that's probably even better - except then the setboot program will need to operate differently than described above, and only modify the FAT BPB sectors and heads values. ALSO - the kernel should then be modified to attempt to recognize a FAT BPB (we need more discussion on exactly now) - and then use the sectors/heads from that.

An easier path for the time being would be to include both FAT BPB and EPB, which was what I though was happening, and the reason that the kernel understood the disk parameters (by reading the EPB as it does now).

ghaerr: > I think it is time to insert the FAT case in the `image/Makefile` to avoid such FAT BPB duplication...

Agreed. However, that will break the build for folks that don't have mtools-4.0.23 installed.

ghaerr: I am worried that if we change these values now, it could stop my ability to mount and/or boot from a FAT volume.

How about this - accept this PR, which fixes the CONFIG_IMG_HD build problem, then later today after work I update setboot to work for FAT boot sectors, submit that PR, then after that, remove the FAT BPB initialization from this FAT boot code, and add the mformat and setboot to the image/Makefile main build. That will ensure things stay working, and solve both issues you're concerned with.

tkchia: Hello @ghaerr, yes, I omitted the ELKS parameter block for the FAT bootloader --- this was mainly due to space concerns.  (There was an `#ifndef BOOT_FAT` surrounding the EPB portion.)  Since there is enough space now, I have added a commit to add the EPB for FAT as well.

tkchia: Hello @mfld-fr, hello @ghaerr, I will also look into reserving space for the whole MS-DOS extended BPB, if that becomes necessary.  Thank you!

ghaerr: @tkchia: it is more important that we make room for the full "extended" BPB (reserved offsets from 3-61) than it is that we include an EPB. I am guessing this is the reason that the `mcopy` command is failing to recognize a valid image with your current boot block installed, and most certainly a full BPB will be compatible with FAT32 and all MSDOS versions.

If an EPB can fit after also having a full BPB, I see the advantages would be: 1) a signature-verified method for setting any other parameters ELKS might need other than sector & head values (such as tracks), 2) compatibility with the future UBL (unified boot loader) and 3) kernel recognition of ELKS boot disk and ELKS sectors/head values. The disadvantages I see of having both are that the sector and head values will be duplicates, and could get out of sync (unlikely). 

Regardless of what happens with including an EPB or not, the kernel will need the ability to read in-the-wild FAT disks (not containing an ELKS boot sector), and read the sector/head values from the FAT BPB, or otherwise resort to ELKS probing and ignoring the formatted disk's sector/head values.

mfld-fr: Okay, let me pull that one as a quick fix to unblock CONFIG_IMG_HD and EMU86 testing, but please now follow the roadmap as defined in #327.

PR #326, opened 2020-02-12
[image] Fix "make images-fat" build

ghaerr:
@mfld-fr:

I pulled down the latest upstream master, which only included my two latest [fat] cleanup commits, and your latest branch 'next' to 'master' merge. I noticed that the "cd image; make images-fat" was now failing, after the last executed `mcopy` inside Make.packages at the end of the do/done shell loop. It appears, for reasons I just cannot find or understand, that the exit value from the last execution of mcopy is stopping the build??! I have double-checked everything and the mcopy is not changed, and is copying, created filesystem disk image is not full.

I have looked through your commit, and can't find anything that could cause this, there's also been nothing changed on my end. Perhaps another shell is being used within make? I saw some changes to use /bin/bash and removal of CFG_SHELL in your commit, but these have nothing to do with this, my env SHELL=/bin/bash. I just don't get it.

In any case, inserting a 'true' into the loop fixes this very strange problem! I would like to ask you try running 'make images' on your system to see whether this is only occurring on OSX. If so, then I guess this PR will be needed for me, for reasons I can't explain.

Thank you!!

### DIFF
--- a/image/Make.package
+++ b/image/Make.package
@@ -94,6 +94,7 @@ msdos:
 	do \
 		[ -d $(DESTDIR)/$$f ] && mmd -i $(MSDOS_IMAGE) ::$$f; \
 		[ -f $(DESTDIR)/$$f ] && mcopy -i $(MSDOS_IMAGE) $(MSDOS_COPYOPTS) $(DESTDIR)$$f ::$$f; \
+		true; \
 	done
 	rm Filelist
 	#mdir -i $(MSDOS_IMAGE) -/ -a ::
--- a/image/Make.package
+++ b/image/Make.package
@@ -84,7 +84,7 @@ msdos:
 #	write ELKS boot sector, mformat will only modify BPB
 #	setboot $(MSDOS_IMAGE) $(BPB) $(FD_BOOTSECT)
 	mformat -i $(MSDOS_IMAGE) $(MSDOS_MKFSOPTS)
-	-rm linux; touch linux
+	rm -f linux; touch linux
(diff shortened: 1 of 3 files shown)


ghaerr:
Does the CI system run "make images" yet?

Can you explain what testing the CI system you've just implemented tests? Is it for completion of a default .config build only, or are there other cases? I'm new to CI. Thank you.

mfld-fr:
@ghaerr : please use the mailing list to discuss such topic not directly related to an issue or a PR. It becomes hard these times to follow all the threads scattered over many posts !

image/Make.package
```diff
@@ -94,6 +94,7 @@ msdos:
 	do \
 		[ -d $(DESTDIR)/$$f ] && mmd -i $(MSDOS_IMAGE) ::$$f; \
 		[ -f $(DESTDIR)/$$f ] && mcopy -i $(MSDOS_IMAGE) $(MSDOS_COPYOPTS) $(DESTDIR)$$f ::$$f; \
+		true; \
```
mfld-fr: This is error masking. The Makefile **has** to fail whenever the `mcopy` command returns an unexpected exit code. Please investigate the reason why `mdir` returns a non-zero code and add a check for that code if actually expected.

mfld-fr (changes requested):

ghaerr:
Added enhanced error detection and handling when non-existent files specified in Packages file, and stop the build with a readable error message. Tested on both minix and fat image package builds, "make images" now works properly.

mfld-fr (approved):

PR #328, opened 2020-02-13
[fat] Finish up 360k, 720k, 1440k and 2880k FAT boot images

ghaerr:
Add setboot -F parameter to skip FAT BPB's when writing boot sector
	Expand FAT BPB boot sector to be fully compatible with all FAT filesystems and MSDOS tools
	Update build and image scripts to only use setboot for all boot sector operations

This PR adds the ability to boot from all ELKS FAT images, including 360k, 720k, 1440k and 2880k FAT12 and FAT16 disks. In addition the FAT boot sector is expanded to a full FAT BPB (offsets 3-61), thus fixing the problem of ELKS-created FAT images being incompatible with any MSDOS version or MSDOS filesystem utility. Finally, all ELKS boot sector operations are solely managed by a single execution of `setboot` per image build, thus simplifying the build process for ELKS-specific parameters and boot sector writing to a single tool.

Tested and working with all ELKS supported images, including .config MINIX build and all package manager builds (3 for MINIX and 4 for FAT).

[EDIT: All the .config-based initialization in the FAT BPB boot sector, questioned by @mfld-fr in another thread, can now be removed and set to 0, as the standard FAT filesystem utility `mformat` is used to initialize all FAT BPB contents at image creation time.]

### DIFF
--- a/elks/tools/setboot/setboot.c
+++ b/elks/tools/setboot/setboot.c
@@ -3,7 +3,7 @@
  *
  * 7 Feb 2020 Greg Haerr <greg@censoft.com>
  *
- * Usage: setboot <image> [-B<sectors>,<heads>] [<input_boot_sector>]
+ * Usage: setboot <image> [-F] [-B<sectors>,<heads>] [<input_boot_sector>]
  *
  *	setboot writes image after optionally reading an input boot sector and
  *		optionally modifying boot sector disk parameters passed as parameters.
@@ -13,6 +13,8 @@
  * 			-> set 9 sectors 2 heads on boot sector or bootable disk <image> file
  *		setboot <image> -B18,2 <bootsector>
  *			-> read <bootsector>, set 18 sectors 2 heads, write <image>
+ *		setboot <image> -F <bootsector>
+ *			-> read <bootsector>, copy skipping FAT BPB to output <image>
  *
  *	Currently only writes ELKS sector_max and head_max values.
  */
@@ -27,6 +29,10 @@
 /* See bootblocks/minix.map for the offsets */
 #define ELKS_BPB_SecPerTrk	0x1F9		/* offset of sectors per track (byte)*/
 #define ELKS_BPB_NumHeads	0x1FA		/* offset of number of heads (byte)*/
+
+/* FAT BPB start and end offsets*/
+#define FATBPB_START	3
+#define FATBPB_END		61
 
 static int SecPerTrk, NumHeads;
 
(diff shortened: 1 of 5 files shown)


elks/tools/setboot/setboot.c
```diff
@@ -66,44 +72,78 @@ int main(int argc,char **argv)
...
+			printf("Updating BPB to %d sectors, %d heads\n", SecPerTrk, NumHeads);
+			argv++;
+			argc--;
+		}
+		else if (argv[1][1] == 'F') {
+			opt_skipfatbpb = 1;			/* skip FAT BPB specified*/
+			printf("Skipping FAT BPB\n");
+			argv++;
+			argc--;
+		}
```
mfld-fr, line 101: Do you know the `getopt` library (https://en.wikipedia.org/wiki/Getopt) ?

image/Make.package
```diff
@@ -96,10 +93,17 @@ msdos:
 		[ -f $(DESTDIR)/$$f ] && mcopy -i $(MSDOS_IMAGE) $(MSDOS_COPYOPTS) $(DESTDIR)$$f ::$$f; \
 	done
 	rm Filelist
-	#mdir -i $(MSDOS_IMAGE) -/ -a ::
-	## FIXME setboot needed to copy fat.bin to FAT sector 1 for .config configuration
-	## setboot hasn't been updated yet for -B<secs>,<heads> and not overwrite FAT BPB
-	## thus package manager builds won't create non-.config based FAT boot images yet
+	# Read boot sector, skip FAT BPB, set ELKS PB sectors/heads and write boot
+ifeq ($(SIZE), 360)
+	setboot $(MSDOS_IMAGE) -F -B9,2 $(TOPDIR)/elkscmd/bootblocks/fat.bin
```
mfld-fr: Please factorize as $(FD_BOOTSECT)

elkscmd/bootblocks/boot_sect.S
```diff
@@ -507,14 +503,16 @@ copy_sect:
...
 msg_boot:
-	.asciz "ELKS boot\r\n"
+	.asciz "ELKS"
 
 msg_error:
 	.asciz "!\r\n"
 
 msg_reboot:
-	.asciz "Press key\r\n"
+	.asciz "Presskey\r\n"
```
mfld-fr, line 512: Beuark 😄 !

elks/tools/setboot/setboot.c
ghaerr: Yes, I'm well aware of it and try to stay as far away from it as possible, as its various implementations across platforms have proven to be completely dysfunctional. You will notice that in my development of `mfs`, I ended up having to include the source code to a very small 'getoptx.c' in order to get mfs to work across the Linux and OSX platforms, where unfortunately they are not compatible.

image/Make.package
ghaerr: FD_BOOTSECT won't work in this case because FD_BOOTSECT is set by Make.defs to ".../minix.bin". We need fat.bin here. I can rearrange Make.defs and create two FD_BOOTSECTS with different names, but didn't want to get into that without your input. Using the prefix $(MINIX_BOOT) would more properly set the directory to use, but its name is quite confusing given this is the FAT image build section.

elkscmd/bootblocks/boot_sect.S
tkchia: Oh well... if (and when) this pull request is merged, I will see if I can wring out a few more bytes of space from the FAT boot code... :)

ghaerr: @tkchia: Yes - I view it as a stroke of good luck that, thanks to your tight work on the FAT boot loader, that there was any room at all to expand the FAT BPB for compatibility, *and* have the EPB included. The good news is, that boot messages are rarely seen, except in the case of a bad boot. And when they are (I ran into boot problems heavily while developing this PR), they are still very understandable, and provide all the same information required for developer debugging.

image/Make.package
mfld-fr: I am afraid you are confusing `FD_BOOTSECT` and `FD_BOOTBLOCK`...

ghaerr: Yes, I'm sorry. There is no FD_BOOTSECT in the source tree. 

FD_BOOTBLOCK won't work in this case because FD_BOOTBLOCK is set by Make.defs to ".../minix.bin". We need fat.bin here. I can rearrange Make.defs and create two FD_BOOTBLOCK with different names, but didn't want to get into that without your input. Using the prefix $(MINIX_BOOT) would more properly set the directory to use, but its name is quite confusing given this is the FAT image build section.

mfld-fr: In my first comment, I just asked to put the repeated string:
```
$(TOPDIR)/elkscmd/bootblocks/fat.bin
```
into a local FD_BOOTSECT to ease the reading...

ghaerr: Got it. Just pushed a cleanup. I had some ideas for cleaning the whole image specification mechanism in the package manager, but wanted to keep this PR clean from that sort of stuff.

mfld-fr: Thanks !

PR #330, opened 2020-02-14
[fat] More FAT build tuning

ghaerr:
Adjust setboot FAT BPB start location to Microsoft Spec (11-61)
	Set RHS attributes on /linux to ensure it remains as contiguous sectors

Setboot now excludes OEMNAME from the FAT BPB "protected area", as per MS specifications, and copies this field from the passed boot sector. `mformat` doesn't write the OEMNAME field on format, and this change keeps our 'ELKSFAT1' oem name in the boot sector, rather than 0 as it was.

Add Read-only, Hidden and System attributes to /linux on FAT images, thus assuring that no MSDOS system will overwrite or change the file location on disk and cause boot failure. Note that ELKS currently ignores the System and Hidden file attributes on mounted FAT filesystems, so /linux will still be visible there.

### DIFF
--- a/elks/tools/setboot/setboot.c
+++ b/elks/tools/setboot/setboot.c
@@ -31,8 +31,8 @@
 #define ELKS_BPB_NumHeads	0x1FA		/* offset of number of heads (byte)*/
 
 /* FAT BPB start and end offsets*/
-#define FATBPB_START	3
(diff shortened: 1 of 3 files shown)


mfld-fr:
One more step toward the release... thanks @ghaerr !

PR #331, opened 2020-02-14
[libc] use `_start' not `entry' as program entry point symbol

tkchia:
This tackles an issue that cropped up earlier when trying to write ELKS programs that do not use the standard `crt0.o` (https://github.com/tkchia/gcc-ia16/issues/50#issuecomment-586292623).

The updated linker scripts will still recognize `entry` as an entry point, if `_start` happens to be undefined.

### DIFF
--- a/elks/elks-small.ld
+++ b/elks/elks-small.ld
@@ -11,7 +11,7 @@ SECTIONS {
 		LONG(SIZEOF (.text));
 		LONG(SIZEOF (.data));
 		LONG (SIZEOF (.bss));
-		LONG (entry);    /* entry point */
+		LONG (_start);    /* entry point */
 		LONG (_total_adjusted); /* total memory allocated */
 		LONG (0);        /* symbol table size */
 		}
@@ -24,6 +24,7 @@ SECTIONS {
 		ASSERT (. + 0x100 <= 0x10000,
 		    "Error: too large for a small-model ELKS a.out file.");
 	}
+	PROVIDE (_start = entry); /* `entry' was the old entry point symbol */
 	PROVIDE (_total = 0);
 	_total_adjusted = _total == 0 ? 0
 			  : MIN (0x10000, MAX (. + 0x100, _total));
--- a/libc/crt0.S
+++ b/libc/crt0.S
@@ -13,12 +13,10 @@
 	.global _start
 
 _start:
-	jmp _startup
 
 // C runtime startup
 // Stack is empty and immediately followed by argc, argv and envp
 
-_startup:
 	push %bp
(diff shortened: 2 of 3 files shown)


ghaerr:
@tkchia: Looks good. One other small note I wanted to mention to you is that libc/Makefile.inc has a hardcoded VERSION=0.3.0 in it... should that be changed to pick up the ELKS version number specified elsewhere?

tkchia:
Hello @ghaerr,

Thanks for your feedback!  Yes, it probably should pick up the ELKS version from `elks/Makefile-rules` or thereabouts, but this is tricky to do right, so I think I will leave that for a separate pull request.

Thank you!

elks/elks-small.ld
```diff
@@ -11,7 +11,7 @@ SECTIONS {
 		LONG(SIZEOF (.text));
 		LONG(SIZEOF (.data));
 		LONG (SIZEOF (.bss));
-		LONG (entry);    /* entry point */
```
mfld-fr, line 14: I intentionally renamed the entry point symbol when moving from BCC to GCC, to detect and review all that entries, plus to forbid any try to link from some old BCC stuff... what is the reason to rename it back today ?

libc/crt0.S
```diff
@@ -10,9 +10,9 @@
 
 // This is the program entry point
 
-	.global entry
+	.global _start
 
-entry:
+_start:
 	jmp _startup
```
mfld-fr: Plus the fact that `_start` could be confused with C `_startup`...

elks/elks-small.ld
tkchia: Hello @mfld-fr,

I see.  My initial reasoning was that many other environments, including GCC for Linux, and also `gcc-ia16` for MS-DOS, use `_start` as the entry point.  And --- I think this is quite a good reason --- `entry` is not a reserved identifier in ANSI C, and thus may clash with valid C code.

I think there is a low risk of accidentally linking with stuff meant for `bcc` --- the `gcc-ia16` toolchain does not prepend underscores to C symbols (by default), while `bcc` does.

Thank you!

libc/crt0.S
tkchia: On second thought, I think this jump is not very useful.  I will remove it.

mfld-fr: Yeah, well done, it solves the 'confusion point'.

elks/elks-small.ld
mfld-fr: OK then if no accidental linking. Thanks !

PR #332, opened 2020-02-14
[fs] Buffers and devices cleanup

mfld-fr:
* Cleanup of some warnings
* Cleanup of some commented out code
* Make buffer constants for sizing as private
* More consistent buffer locking
* No more HD partitions than allowed by MBR


ghaerr:
This seems like a pretty big change this late in the release cycle. I see most of it is cleanup, but a few changes with map/unmap (I don't fully understand the depths of internal buffer management though). How have you tested that the revised locking/mapping algorithm works?

I notice that some read-ahead code was deleted. Did it ever work? Our implementation of the FAT filesystem would be sped up with working read-ahead, particularly in the case of searching  directories when creating a new entry (entire directory needs to be searched before create).

What is your rationale for determining the number of NR_MAPBUFS, up from 8 to 12 when CONFIG_FS_FAT is set? I'm interested in what why you think it should be bigger, as well as your thoughts on our kernel data size getting even larger (it seems we still have room to spare though).

Definitely like the slimming down of the block special device count, I noticed it was way too large when working on the package manager and considering the possibility of which ones to add to FAT /dev.

mfld-fr:
* Sure, this is a working branch, not for next release
* I always test my changes, but I can miss somes cases, as anybody 😄 
* Read ahead strategy is at file system level,  neither at block device nor at buffer level, so the cleanup
* Buffer reference counting and buffer locking are today confused (incorrect design)
* Reference counting is needed for more efficient RAM disk (in buffers only)

mfld-fr:
Have to split the changes in several commits however...
Edit: DONE

mfld-fr:
To be rebased on latest HEAD. Closing for now.

mfld-fr:
Branch 'next' can be deleted, now working in my fork.

Issue #303, opened 2020-02-06
[config] Config system needs updating


ghaerr:
The KConfig-based system is broken on OSX, so I can't update or test various CONFIG options. Following are ones needed in order to test new features being added to MSDOS-FS. These need to be added to various config.in scripts, and arch/i86/defconfig, I think.

These options should be accessible when turning on and off CONFIG_MSDOS_FS
CONFIG_FULL_VFS          (currently working)
CONFIG_MSDOS_DEV     (turns on /dev support on FAT volumes)
CONFIG_32BIT_INODES  (missing from some config scripts it seems)

Will the "make config" option work, instead of "make menuconfig", if the above options are fixed?

It also seems that a "make clean" does not reset the system state to start. Does the .config file need to be removed for that to work?

Sorry for the dumb questions, I haven't been able to run the config system and builds are starting to get messed up strangely.

mfld-fr:
As I said, I think it is worth to take a little time to solve the 'menuconfig' on OSX before continuing the rush...

ghaerr:
How do you think its best to proceed on fixing menuconfig/config, without having to jump in too far into this old and needs-updated code? You were mentioning that perhaps it should be updated to a new version, I don't know anything about that? On one of my OSX systems, the make config fails with ncurses library entry points not found, even though I have ncurses, and on another, make menconfig runs but crashes when running through various screens.

Would you like me to post error messages so you or others can direct me?

ghaerr:
We desperately need to get MSDOS_CONFIG_DEV and the CONFIG_32BIT_INODES inserted properly in a config.in an defconfig files. Would you like me to prepare a PR for this, even though I can't run it?

Also, I have been hand-editing include/autoconfig.h, otherwise I couldn't test CONFIG_MSDOS_DEV support, that might be part of my very strange build problems going on. I have to admit, I don't yet understand the build process with .config and include/autoconfig.h being used to generate make steps...

mfld-fr:
@fonin highlighted in his commit bfb0496f369fdcc022c62556a29e692191a5e80c that one needs the 'development package' of 'libncurses' to build `menuconfig`.

fonin:
@ghaerr if in urge, you might configure on another system with working menuconfig, and place .config on the target system. Or even edit .config manually, it is pretty simple.

ghaerr:
[...]

PR #335, opened 2020-02-15
[image] FAT image in .config

mfld-fr:
* Integrate FAT image in standard build
* Remove useless `knl` program
* 32-bit inode number as a visible option (for #303)

Problem: boot hangs with `fd1440.bin` (with FAT image option in `.config`) just after mounting root... while it succeeds with `fd1440-fat.bin`. I should have missed something ?..

Edit: no problem with MINIX image option.

Edit 2: problem solved with the help from @ghaerr.


ghaerr:
First look: did you mean 'ifdef CONFIG_FAT_DEV' instead of CONFIG_IMG_DEV in line 64 of image/Makefile?

ghaerr:
While you're at it, the MINIX build has consistently spat out an error on the following lines in image/Makefile:

```
    find $(DESTDIR) \
        -type d ! -name root -execdir chmod go+rx '{}' ';' -o \
        ! -type d ! -name .profile -execdir chmod go+r '{}' ';'
```

Now that we use the `mfs` tool to build images, can't we just delete that altogether?

ghaerr:
I didn't see where you removed `knl` from .config or the build, good idea. Can you also remove it from image/Packages so I don't have to?

ghaerr (commented):
I'm still new with Config: where are the new menuconfig options for CONFIG_IMG_MINIX and CONFIG_IMG_FAT?

mfld-fr:
> First look: did you mean 'ifdef CONFIG_FAT_DEV' instead of CONFIG_IMG_DEV in line 64 of image/Makefile?

No, no, it is correct : CONFIG_IMG_DEV means "*the image format supports the special devices files*". So YES for MINIX and ROM, NO for FAT.

CONFIG_FS_DEV is to enable the FAT 'fake dev' mechanism. Will mean "*enable devfs*" in the future, so the name.

mfld-fr:
> I'm still new with Config: where are the new menuconfig options for CONFIG_IMG_MINIX and CONFIG_IMG_FAT?

Sorry, did not catch the question ?

ghaerr:
> > First look: did you mean 'ifdef CONFIG_FAT_DEV' instead of CONFIG_IMG_DEV in line 64 of image/Makefile?
> 
> No, no, it is correct : CONFIG_IMG_DEV means "_the image format supports the special devices files_". So YES for MINIX and ROM, NO for FAT.
> 
> CONFIG_FS_DEV is to enable the FAT 'fake dev' mechanism. Will mean "_enable devfs_" in the future, so the name.

Confusing, but I get it now.

I was thinking that the reason the boot may be failing was that the /dev entries were not present on the FAT filesystem. I updated /bin/init to give decent error messages when it can't open various /dev files... what other process does the kernel do after mounting the root filesystem but before running /bin/init? Perhaps also run `mdir` on fd1440.bin and take a quick look at it.

ghaerr:
> > I'm still new with Config: where are the new menuconfig options for CONFIG_IMG_MINIX and CONFIG_IMG_FAT?
> 
> Sorry, did not catch the question ?

I don't understand Config yet: Where are the questions in config.in that `menuconfig` asks for the above new options? That would allow me to see what is being asked and defaulted. Same for my confusion with CONFIG_IMG_DEV and CONFIG_FAT_DEV: I didn't see the code in `menuconfig` for those entries (I realize they are in various config.in files in the appropriate directories).

ghaerr:
Just thought of the likely problem: The directory /dev must be created in fatfs:, and be within the first two directory entries for the 'automatic' CONFIG_FAT_DEV to work. I did this initially to avoid creating a /dev system on in-the-wild disks that happened to have a /dev in them. I can rewrite that check to something different if this proves problematic, but I think running an `mmd -i ... /dev` after the `mcopy -i ... ::/linux` will solve the problem.

mfld-fr:
Yeah, it works with your '/dev' trick :smiley: ! Thank you !

mfld-fr:
> While you're at it, the MINIX build has consistently spat out an error on the following lines in image/Makefile:
> 
> ```
>     find $(DESTDIR) \
>         -type d ! -name root -execdir chmod go+rx '{}' ';' -o \
>         ! -type d ! -name .profile -execdir chmod go+r '{}' ';'
> ```
> 
> Now that we use the `mfs` tool to build images, can't we just delete that altogether?

I remember these lines were added by @tkchia in https://github.com/elks-org/elks/pull/227.

PR #337, opened 2020-02-15
[chmem] Add ELKS chmem command

ghaerr:
Implements ELKS /bin/chmem.

Sort now works with `chmem +48000 /bin/sort`, and bc is running with `chmem +32768 /bin/bc`.

It was a little tricky setting the total data segment size correctly, and I have not yet tested with non-seperate I&D (how would that be specified to `gcc-ia16`?

This solves @Mellvik problem listed in #123 regarding running out of memory.

Added debug messages to `chmem` for the time being (see below).

Screen Shot 2020-02-15 at 3 13 58 PM

### DIFF
--- a/elkscmd/sys_utils/Makefile
+++ b/elkscmd/sys_utils/Makefile
@@ -26,6 +26,7 @@ PRGS = \
 	knl \
 	man \
 	poweroff \
+	chmem \
 	# EOL
 
 # clock disabled because direct I/O port access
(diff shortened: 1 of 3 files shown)


tkchia:
Hello @ghaerr,

"Combined I/D" or "tiny" memory model `a.out` programs are currently not supported by the ELKS kernel (if they ever were...).  But in any case, the way to compile such a program with `gcc-ia16` would be to specify the options `-melks-libc -mcmodel=tiny`.

Thank you!

ghaerr:
Changed chmem max data size allowed to 65520, as kernel exec fails with any larger values.

mfld-fr:
Nice runtime tool. Thanks @ghaerr !

PR #339, opened 2020-02-16
[nanox] Include landmine game rather than nclock for Nano-X demo

ghaerr:
Suggest to use '/bin/landmine' rather than nclock to show off ELKS' graphical capabilities. It's actually a usable game and lots better looking.

Screen Shot 2020-02-15 at 9 41 52 PM


mfld-fr:
Well, why not ? Better for advertising :rose: :tulip:

PR #340, opened 2020-02-16
LD scripts: allow setting non-static data size, via a _chmem variable

tkchia:
This is to support setting the size of the non-static data space from an `ia16-elf-gcc` command line, i.e. while building an ELKS program (https://github.com/elks-org/elks/issues/241).

Newer versions of `ia16-elf-gcc` will have a `-maout-chmem=`... option to set this `_chmem` value.  Alternatively, even without such an option, one can say e.g. `-Wl,--defsym=_chmem=0x4000`.

(_This is complementary to @ghaerr's pull request_ (https://github.com/elks-org/elks/pull/337) _which adds support for changing the data segment allocation of ELKS binaries after they are built._)

### DIFF
--- a/elks/elks-small.ld
+++ b/elks/elks-small.ld
@@ -21,13 +21,13 @@ SECTIONS {
 	.data 0 : AT(_begindata) { *(.nildata*) *(.rodata*) *(.data*) }
 	.bss : {
 		*(.bss) *(COMMON)
-		ASSERT (. + 0x100 <= 0x10000,
+		ASSERT (. + 0x100 <= 0xfff0,
 		    "Error: too large for a small-model ELKS a.out file.");
 	}
 	PROVIDE (_start = entry); /* `entry' was the old entry point symbol */
 	PROVIDE (_chmem = 0);
 	PROVIDE (_total = _chmem == 0 ? 0 : . + _chmem);
 	_total_adjusted = _total == 0 ? 0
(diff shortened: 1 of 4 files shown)


tkchia:
I also added a patch for `elksemu`, to get it to honour an `a.out` program's `a_total` field, like the ELKS kernel does.

ghaerr:
Hello @tkchia,

Looks like good modifications to elksemu, which previously would run applications like `sort` with a 64k data segment, regardless of the value of the chmem/total header field. Have you tested that `sort` now does not run (which would be correct, given it's total field is 0)?  Same for `bc`. I plan up updating both with your new `gcc-ia16` -ma.out-chmem= option after regression tests with appropriate dynamic data values.

What is the elks-tiny.ld linker script used for? Weren't you saying that `gcc-ia16` and/or ELKS doesn't support tiny model? I remain somewhat confused on this issue.

mfld-fr:
> What is the elks-tiny.ld linker script used for? Weren't you saying that `gcc-ia16` and/or ELKS doesn't support tiny model? I remain somewhat confused on this issue.

GCC-IA16 does support both 'tiny' and 'small' memory model. The 'tiny' model is used for `bootsect` and `setup` (`grep` is your friend :smile:). ELKS does not know today how to load 'tiny', but it could at a low cost.

PR #343, opened 2020-02-17
[cmds] Fix ls -F option

ghaerr:
Fixes previously reported `ls -F /` or `ls -lf /`, which hangs ELKS.

Upon further inspection, it was determined that ELKS wasn't hung, but instead the `ls` process busy-looping inside ELKS `malloc` inside `realloc`. Since I couldn't determine whether this was an `ls` bug overwriting memory, or an ELKS libc `malloc/realloc` bug, I rewrote `ls` to use a different method than calling `realloc`. Not sure whether there are regression tests for ELKS libc functions.

### DIFF
--- a/elkscmd/file_utils/ls.c
+++ b/elkscmd/file_utils/ls.c
@@ -248,6 +248,7 @@ static void lsfile(char *name, struct stat *statbuf, int flags)
 #endif
     }
 
+#if 0 // broken - hangs in malloc in realloc()
 /* If a class character exists for the file name, add it on */
     if (class != '\0') {
 	    len = strlen(name);
@@ -263,14 +264,20 @@ static void lsfile(char *name, struct stat *statbuf, int flags)
 	    classp++;
 	    *classp = '\0';
     }
+#endif
 
     {
 	char *cp;
+	char buf[255];
 
 	cp = strrchr(name, '/');
 	if (!cp) cp = name;
 	else cp++;
-	printf(fmt, cp);
+	if (class)		/* handle class without realloc*/
+		sprintf(buf, "%s%c", cp, class);
+	else
+		strcpy(buf, cp);
+	printf(fmt, buf);
     }
 
 #ifdef S_ISLNK


mfld-fr (changes requested):
Sorry, but this PR only workarounds something that should be understood, as it could fail in other places, and the new implementation is inefficient : we cannot afford allocating 255 more bytes on the stack just to append one class character. I know you can do better than this quick & dirty 😃.

ghaerr:
Agreed that this fix doesn't determine whether `ls` is buggy, or instead `realloc`. I don't have the step-debugging tools to determine that. The `malloc` implementation is quite complicated, and didn't want to touch that at this time. Subsituting another implementation could make sense, but it's likely quite dependent on ELKS `sbrk`.

With regards to inefficiency of the proposed fix, I don't understand: allocating 255 bytes of a temporary buffer to copy a filename into vs 14 doesn't cost anything more. I allocated 255 in the case of long filename support, so that we didn't have to rewrite it later. With regards to appending a single class character, that is complicated because the following `printf` in the code uses a variable format string for filename spacing and can't just be appended.

So actually my fix isn't quick and dirty :)

mfld-fr:
I do not agree. Allocating 255 bytes on the stack is nothing on today PC, but has an impact on a data segment with KiB as order of magnitude. Especially just to add one character to the filename.

ghaerr:
Sorry, I'm still not following you: whether its 14 or 255 stack bytes, the code is contained in a block and the stack use deallocated after each name printed, so no aggregation of space used?

mfld-fr:
Mmm... I feel I must take some times to write an 'implementation guidelines' paper and put inside all that @jbruchon and @lithoxs explained before 🤔...

ghaerr:
> Mmm... I feel I must take some times to write an 'implementation guidelines' paper and put inside all that @jbruchon and @lithoxs explained before 🤔...

Perhaps there should be one... unfortunately I was not on the list when this was being discussed, and still don't understand what you're suggesting I should do on this. Would using a 14+2+1 = 17 byte buffer meet the requirements?

ghaerr:
@mfld-fr: Please let me know what you think is a better way than I have coded, to fix `ls -F`. Agreed that we don't need a 255 buffer, 17 will work. If you would like me to make that change, please let me know.

Otherwise, just close this PR, I am tired of trying to fix user land commands. Boy, am I sorry I opened up this ongoing mess with my two `ls` PRs!!!

mfld-fr:
This `name` buffer reallocation to insert the class character, before your first fix, is typically a 'short-sight' patch. Let us have a look from a higher standpoint...

That `name` buffer is allocated through a `strdup` and passed to `pushstack` (except when passing an already allocated buffer in line 519). Just have to reserve one more character for the class, by replacing `strdup` by a `malloc(strlen()+ 1)` and a `strcpy`. Then you have room to insert the class character without any more allocation / reallocation later, thus saving much processing at the cost of 1 byte per name.

ghaerr:
Thank you for pointing out the strdup, I missed that. I will submit a better patch for you.

Sent from my iPhone

PR #344, opened 2020-02-17
[stat] Fix st_ino type to be constant size in stat system call

ghaerr:
Use 32-bit `u_ino_t` size for `st_ino` field in `stat` system call, which allows applications to run without special coding for CONFIG_32BIT_INODE setting.

Since this was required to get `ls -i` operational, I'm submitting this PR. If the `gcc-ia16` toolchain also needs stat.h to be fixed internally, I'll defer that to @tkchia.

### DIFF
--- a/elks/include/arch/stat.h
+++ b/elks/include/arch/stat.h
@@ -5,11 +5,7 @@
 
 struct stat {
     dev_t	st_dev;
-#ifdef CONFIG_32BIT_INODES
-    __u16	st_ino;
-#else
-    ino_t	st_ino;
-#endif
+    u_ino_t	st_ino;
     mode_t	st_mode;
     nlink_t	st_nlink;
     uid_t	st_uid;


elks/include/arch/stat.h
```diff
@@ -5,11 +5,7 @@
 
 struct stat {
     dev_t	st_dev;
-#ifdef CONFIG_32BIT_INODES
-    __u16	st_ino;
-#else
-    ino_t	st_ino;
-#endif
+    u_ino_t	st_ino;
```
mfld-fr, line 8: Not compliant to POSIX standard.

mfld-fr (changes requested):

elks/include/arch/stat.h
ghaerr: Ok. What do you suggest, as the current implementation causes all internal and external applications that access the `st_ino` field to 1) have to test the size of the that field, and handle each case separately, as is the reason `ls` printf fails, since it uses a hard-coded `%ld` when the inode is 16-bit, and 2) every internal and external application will have to be recompiled when CONFIG_32BIT_INODE changes otherwise.

mfld-fr: It looks like first the stat.h is not correct : ino_t is already & correctly defined in linux/types.h, this conditional in arch/stat.h is useless and erroneous.

ghaerr: I'll let you decide what to do about it all... I remain concerned about the fact that the size changes in my comment above, and that affect on applications, that's all. After all, this is the reason why `ls -i` is failing, and nobody knew about it.
[EDIT: I mean this is the reason why `ls -i` is failing.]

mfld-fr: Yeah, you well spotted the defect, thanks for that, but I think we should cleanup the room first to avoid building our house on wrong basements.

ghaerr: > It looks like first the stat.h is not correct : ino_t is already & correctly defined in linux/types.h, this conditional in arch/stat.h is useless and erroneous.

No, the way it is currently coded ensures that `st_ino` is always 16 bits, removing the conditionals would break the ABI. My fix ensures that `st_ino` is always 32 bits.

With no change, `ls -i` can be easily fixed with a `printf` change, but will report incorrect i-node numbers when running large FAT16 or FAT32 disks (in which case the inode number is 'fake' anyways). With this change, the inode number would always be correct, but apparently is not POSIX compliant. Allowing the field member to change size with different kernel compilations would break the ABI and be a larger mistake.

After more thought, perhaps it's easier to just fix `ls -i` using the first method, which is no change to stat.h and a simpler change to `ls`.
[EDIT: sorry, I keep typing `ls -F` when I mean `ls -i`. Corrected above.]

tkchia: Hmm... I will have to find the time to look into these changes more closely. :-|

(Perhaps another possibility is to add a new `stat32` or `stat64` syscall, and then try to figure out how to wrap up all the underlying syscalls inside a `stat (`...`)` library routine at the userland level (argh).)

ghaerr: > (Perhaps another possibility is to add a new `stat32` or `stat64` syscall, and then try to figure out how to wrap up all the underlying syscalls inside a `stat (`...`)` library routine at the userland level (argh).)

That seems like overkill when the only time 32-bit inodes are needed is when reading a FAT32 or large FAT16 disk. Even in those cases, returning a 32-bit inode number has dubious value, as the FAT 'inodes' are generated by the driver, not real inodes.

mfld-fr: Fixed in a clean manner in commits 12f1995c and 32f25812.

mfld-fr:
Closing this one as I just fixed in commits 12f1995 and 32f2581.

PR #345, opened 2020-02-17
[image] Remove unneeded chmod from image generation template

ghaerr:
Remove commit from PR #227. 

This code used to change ownership of target files in the case of a git pull with a restricted umask, which then caused a build failure on the kernel-mounted filesystem when changing directory to /dev. Since ELKS now uses `mfs` and `mtools` to create and populate each disk image with the proper permissions under any UID/GID user, this code is no longer needed (and `chmod` emits an error message during the build anyways).

### DIFF
--- a/image/Makefile
+++ b/image/Makefile
@@ -52,10 +52,6 @@ template:
 ifdef CONFIG_IMG_BOOT
 	install $(ELKS_DIR)/arch/i86/boot/Image $(DESTDIR)/linux
 endif
-	# pull request #227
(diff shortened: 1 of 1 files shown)


mfld-fr:
Waiting for feedback from @tkchia...

tkchia:

Yes, it looks like the image generation does work correctly even without the `chmod`'s.  Please go ahead and remove the command.

Thank you!

mfld-fr:
Thanks @tkchia for the feedback !

PR #346, opened 2020-02-17
[elksemu] Fix gremlins making elksemu fail on openSUSE 13.2 x64

tkchia:
@georgp24 (https://github.com/elks-org/elks/issues/334) had reported that `elksemu` was failing with "waitpid failed" under openSUSE 13.2 for x86-64, which uses Linux kernel 3.16.x.

It turns out that there are some subtle but important differences between Linux kernel version 3.16.x and the newer 4.x kernels in how the `waitpid` and `ptrace` syscalls are implemented.

`elksemu` should now work under either type of kernel.

### DIFF
--- a/elksemu/elks.c
+++ b/elksemu/elks.c
@@ -193,7 +193,7 @@ static int wait_for_child(void)
 {
 	pid_t child = elks_cpu.child, pid;
 	int status;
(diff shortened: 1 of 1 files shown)


mfld-fr:
Wow, so tricky...

PR #357, opened 2020-02-21
Third burst of patches from Marcin (2)

mfld-fr:
Patches posted by @marcin-laszewski on the mailing list, so finishing my maintainer job by pushing them to upstream.


marcin-laszewski (approved):

marcin-laszewski (approved):

marcin-laszewski (approved):

ghaerr:
@mfld-fr: thank you.

@marcin-laszewski: Wow, you've been busy :) I'd like to discuss various aspects of this large change before committing, and then open the discussion of how we can get busyElks best integrated into ELKS. Thus, a first few questions and comments:

1) The biggest issue with accepting this all at once is that makes quite a few changes to libc  (much needed however, thank you), and some discussion is needed as to whether this should go in now, or after this release. Also, `awk` is used which has seen some compatibility problems with OSX so I'd like to test that.

2) Where did the additional libc functions come from, are they from the `dev86` libc tree?

3) Nice job on the split-ups of malloc and printf etc. Was additional code added with each of the split-ups, or are they only that - splitting?

4) Are the alloca and mini_malloc also from `dev86`?

5) Many of the string functions use the `BCC_ASM` ifdef from the old `bcc` compiler, so of course they'll be compiled using the slow C version. These functions will want to be replaced by @tkchia or other's implementations.

6) Another issue with ELKS libc in general that I'd like to have discussed has to do with having separate locations for the libc (mostly asm) functions used by the kernel, versus user land. I lean towards the idea of having them kept in separate places. @tkchia, can you comment on this?

7) Does this add all of (I assume) `dev86` libc that was missing, or does this just add the functions that are needed for busyElks? I had suggested a year ago that all of `dev86` libc be added to have a complete C library.

Thanks for the large movement forward in separating up major functions in libc and adding so many other functions. I will bring up issues of getting busyElks integrated into a separate post.

ghaerr:
On the business of getting busyELKS (or any other large group of new applications) integrated into ELKS, I have the following thoughts:

1) The `busyElks/` directory should be moved into `elkscmd/busyElks/`.

2) For the *short term*, rather than adding another .config/Config entry on whether to build these applications, all busyElks applications would be built in every make, but NOT installed. This ensures that all applications remain compileable while we make ABI and/or other libc changes.

3) The application-set contributor would create a custom 'busyElks.package' file for their contribution, which would use the package manager to build a bootable FAT or MINIX floppy image with their intended application-set. This would very quickly and easily allow the ELKS community users access to images for testing and comment. It is also quite necessary, since we're completely out of space for existing applications, yet alone new ones, in the 1440k image.

4) In the elks/images directory, the application-set contributor would do something like the following to create a packages file once, from the already built application tree:
```
ls -R template > busyElks.package
{edit package file to manually add or remove applications}
```
(Please see the existing `image/Packages` and `Make.package` to see how things currently work).
(This also necessitates any new command sets be placed in directories where they don't conflict with existing command names - only for the short term).

After editing the new `busyElks.package` file created as above from the ELKS bootable image template tree (`/template`), the following package manager command would allow easily building the application-set image, which could then be added to the 'make images' Makefile:
```
make -f Make.package PACKAGE=busyElks.package SIZE=1440 FS=minix
```
This would create a `fdbusyElks-1440-minix.bin` in elks/image, ready for distribution and testing by the community. I am thinking that a 'make all' from top-level ELKS would compile ELKS and create all images.

There are still a few more changes I need to make before the above command will function, but this gives the general overview of solving our current problem of "too many applications, too little disk space". Comments welcome.

[EDIT: The package manager also uses a 'tags' mechanism that allows a single Package file to specify file inclusion in an image for different image sizes or types using a ":tag" mechanism. The 'make -f' above would also handle an additional `TAGS=` parameter for this functionality. Symlinks are also properly currently handled.]

jbruchon:
I'd like to know about merging this relatively soon, if someone wouldn't mind doing a little review and letting me know what they think.

ghaerr:
@jbruchon: I just wrote up a lengthy review with questions for the contributor, are you looking for more than that? There are definite issues rushing this commit, since there are lots of libc changes.

jbruchon:
Sorry, I see that now. Let's wait and see what comes up.

marcin-laszewski:
2020-02-21 16:30 GMT+01:00, Gregory Haerr <notifications@github.com>:
> (...)
>
> 1) (...)
> `awk` is used which has seen some compatibility problems with OSX so I'd
> like to test that.

AWK script had been already used to created one ASM file "syscall1.s".
I have modified it and now, it creates one *.s file per each syscall.

> 2) Where did the additional libc functions come from, are they from the
> `dev86` libc tree?

There were no additional function.

> 3) Nice job on the split-ups of malloc and printf etc. Was additional code
> added with each of the split-ups, or are they only that - splitting?

Splitting only.

> 4) Are they alloca and mini_malloc also from `dev86`?

No. I have splitted already existing *.c files in libc/ directory/ only.

> 5) Many of the string functions use the `BCC_ASM` ifdef from the old `bcc`
> compiler, so of course they'll be compiled using the slow C version. These
> functions will want to be replaced by @tkchia or other's implementations.

I didn't touch them.

> 6) (...)
> 7) (...)

No addressed to me.

-- 
Pozdrawiam,
Marcin

marcin-laszewski:
2020-02-21 17:25 GMT+01:00, Gregory Haerr <notifications@github.com>:
> On the business of getting busyELKS (or any other large group of new
> applications) integrated into ELKS, I have the following thoughts:
>
> 1) The `busyElks/` directory should be moved into `elkscmd/busyElks/`.

I did it yesterday.

> 2) For the *short term*, rather than adding another .config/Config entry on
> whether to build these applications, all busyElks applications would be
> built in every make, but NOT installed. This ensures that all applications
> remain compileable while we make ABI and/or other libc changes.

I have integrated busuelks with ELKS' menuconfig. It works and I test it now.
I will send my patches soon (Monday).

> 3) The application-set contributor would create a custom 'busyElks.package'
> (...)
>
> 4) In the elks/images directory, the application-set contributor would do
> something like the following to create a packages file once, from the
> already built application tree:
> (...)

Good idea.

> There are still a few more changes I need to make before the above command
> will function, but this gives the general overview of solving our current
> problem of "too many applications, too little disk space". Comments
> welcome.

-- 
Pozdrawiam,
Marcin

ghaerr:
@tkchia: Now that @marcin-laszewski has indicated that this large patch is only a split-up of existing `elks/libc/` functions, cleanup and a bit of restructuring of function declarations in the existing libc header includes, I am in favor of committing it, provided that this restructuring doesn't overly complicate the work required with your last idea of bringing over `dev86` includes. I still need clarification on the questions I asked regarding whether you're talking about a new `elks-libc/` from `dev86`, or `elks/libc/` having `dev86` include files.

ghaerr:
> 2) For the *short term*, rather than adding another .config/Config entry on whether to build these applications, all busyElks applications would be built in every make, but NOT installed. This ensures that all applications remain compileable while we make ABI and/or other libc changes.
> I have integrated busyelks with ELKS' menuconfig. It works and I test it now.

The integration into the menuconfig system this early introduces a problem: since the fd1440.bin disk image is almost full, then busyElks may have to default OFF. If the default is OFF, then the busyElks applications won't be built in every build, and a later package manager run would fail unless ELKS is rebuilt with option ON. If busyElks default ON in menuconfig still fits on 1440k, then none of this applies, but then there would be duplicate-named commands in /bin and /usr/bin. These are reasons why I suggested going the package-manager only route in the short term, so that others can comment on usability. A further complicating case is that the FAT package manager images would have to use the original, non-busyElks commands.

> I will send my patches soon (Monday).

Would it be a possibility that you could send future changes via a PR, rather than patches on the mailing list? I for one am not (yet) monitoring the mailing list and a direct PR would be much easier. @mfld-fr has greatly helped but possibly indicated he may not be continuing to package your patches into PRs?

Issue #222, opened 2018-06-13
ELKS uses weak sleep_on()


mfld-fr:
As documented in LDD3 and ULK2 books, there is a possible race condition in the current sleep / wake implementation, as a task could be woken up between checking the sleep condition and putting the task to sleep.

It is unlikely to occur between tasks with a single CPU core if the condition check & sleep put are close to each other, but is more likely to occur when waking up from within an interrupt routine for fast devices, like the Ethernet and the HR timers.

Solution is to replace the current `sleep_on ()` by the more modern `wait_event ()` implementation, but with the condition expression replaced by a test function, to reduce the cost of macro expansion in term of code size.

mfld-fr:
Implemented the `wait_event()` function, staged in NE2K driver.

mfld-fr:
Will continue my work of improving the kernel synchronization in my own fork. Closing this one.

PR #361, opened 2020-02-21
[kernel] New synchronization primitives

mfld-fr:
No impact except NE2K driver, tested through `telnetd` under QEMU.


ghaerr:
Does this fix a bug, or is this a new feature? If only a new feature, is NE2K the only code that uses the previous/current wait primitives? Just asking for stability purposes for upcoming release.

Where is the old, replaced lock code? I don't see that being deleted.

A couple questions on include files:

1. What is the purpose of the include/arch/*.h files, versus include/linuxmt/*.h? That is, since atomic.h is now in linuxmt/, and defines entry points in architecture-specific atomic.S, shouldn't atomic.h be in linuxmt/?

2. I also realize that our include files are a bit messy at this point. Is the new standard to remove whole-file ifdef LX86_LINUXMT_WAIT_H for instance, and always use pragma once? I don't know whether that pragma is a standard or whether we're now marrying our .h files to a gcc-only compiler.

Nice to see unneeded lock.h includes removed from various kernel sources.

Thank you!

tkchia:
Hello @ghaerr,

> I don't know whether that pragma is a standard or whether we're now marrying our .h files to a gcc-only compiler.

It is not standard, but apparently [quite widely supported](https://en.wikipedia.org/wiki/Pragma_once#Portability).  Personally I would still prefer the original include guards though.

Thank you!

tkchia:
Regarding the locking logic itself, I am not familiar with the ins and outs of this particular type of Ethernet card, so I will have to defer to you all.

Thank you!

mfld-fr:
Not a bug fix - rather a preparation step.

The current `sleep_on` is still unsafe (see #222) and it fails when used with NE2K as a fast device (ethernet...). Fixed by introducing a first version of `wait_event` some times ago, only for NE2K. Now introducing some more safe primitives, but had to rework `wait_event` for non-static conditions, so impacting again NE2K.

Other parts of the kernel are unchanged. Still use the unsafe `sleep_on`, the old semaphore, or even shared data **not** protected against concurrent access. New synchronization primitives will help to have a more robust code for reference counting and locking.

About the headers : `atomic.h` and `lock.h` are in `include/linuxmt` because interface of the primitives is not architecture dependent. Atomics implementation is specific to 8086 architecture, so `atomic.S` is under `arch/i86`.

ghaerr:
> Regarding the locking logic itself, I am not familiar with the ins and outs of this particular type of Ethernet card, so I will have to defer to you all.

Neither am I, so I defer to @mfld-fr as well.

@mfld-fr, thank you for the explanation, and this contribution.

PR #363, opened 2020-02-21
[fat] Fix create file bug on deleted FAT directories

ghaerr:
This fixes the first of @Mellvik 's reported FAT bugs in #123 from his supplied disk image, which is that running `touch` on a certain MSDOS floppy caused ELKS to hang.

This took a while. The technical problem is that @Mellvik 's floppy's first available unused directory entry happened to be that of a previously deleted directory entry. Apparently MSDOS does not zero the deleted directory 'first cluster'  entry when deleting directories, and when when our FAT code tried to compute the size of what was a directory entry (when building a 'fake' inode for it), it followed an invalid FAT chain that resulted in an infinite loop in the kernel FAT driver.

### DIFF
--- a/elks/fs/msdos/dir.c
+++ b/elks/fs/msdos/dir.c
@@ -119,7 +119,7 @@ int msdos_readdir(struct inode *inode, register struct file *filp, char *dirent,
 	ino = msdos_get_entry(inode,&filp->f_pos,&bh,&de);
 	while (ino != (ino_t)-1L) {
 		/* Check for long filename entry */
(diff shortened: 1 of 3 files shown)


ghaerr:
Second commit also fixes @Mellvik 's last two directory display problems from uploaded images in #123. This should fix all of the known FAT problems reported as open issues.

@Mellvik: Please test, if your images all work, #123 can be closed, as it's getting rather long!

ghaerr:
Third commit fixes disk too large kernel check for FAT16 and FAT32 volumes requiring CONFIG_32BIT_INODES to be set ON. Thanks @mfld-fr for finding this.

PR #365, opened 2020-02-22
[tools] [elkscmd] compile `bc` & `sort` programs with new gcc-ia16 `-maout-chmem=` option

tkchia:
These commits do a few things to allow the `bc` and `sort` programs to finally work, rather than bail out due to lack of memory.

  * I updated the `gcc-ia16` version to a more recent one (https://github.com/tkchia/gcc-ia16/commit/53a9614cbbc5880a1e11bdd73f8af933202ff24c) which includes a `-maout-chmem=`_x_ option for setting the amount of memory reserved for non-static data (stack, heap, etc.).  The new `gcc-ia16` version also allows a simpler command line syntax for building ELKS programs (`-melks-libc -mcmodel=small` → `-melks`), and includes some other features and fixes.  I have already run the new version through GCC's (plus my own) regression test suite; so it is well-tested.
  * I modified the makefiles for `bc` and `sort`, so that they now use `-maout-chmem=`_x_ when building these programs.

All this should further advance the solving of issue https://github.com/jbruchon/elks/issues/241 .

### DIFF
--- a/elkscmd/bc/Makefile
+++ b/elkscmd/bc/Makefile
@@ -30,6 +30,9 @@ LEX = flex -I -8
 
 LOCALFLAGS=-D_POSIX_SOURCE
 
+# For ELKS, bc needs more data segment space than the kernel-given default.
+LDFLAGS += -maout-chmem=0xc000
+
 OFILES = scan.o util.o main.o number.o storage.o load.o execute.o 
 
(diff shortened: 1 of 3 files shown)


ghaerr:
Looks great, thank you. Two more broken commands now working!

PR #370, opened 2020-02-23
[image] Standardize package mgr args for use with multiple package files

ghaerr:
This commit allows easy creation of bootable busyElks or other application packages for testing by use of the following in `elks/images/Makefile`:
```
make -f Make.packages PACKAGE=busyElks.package NAME=busyelks FS=minix SIZE=360
```
In the above usage, a bootable 360k minix disk image `busyelks-minix.bin` will be created with all files specified in the `busyElks.package` file.

The ELKS PB and minix or fat mkfs options are set automatically from the SIZE= parameter, which can be one of 360, 720, 1440 or 2880. An optional TAGS= parameter allows the use of a single package file for multiple disk sizes, if desired.

### DIFF
--- a/Documentation/text/package-manager.txt
+++ b/Documentation/text/package-manager.txt
@@ -1,0 +1,25 @@
+ELKS Package Manager
+
+The package manager facility lives in the files image/Makefile, image/Make.package and image/Packages.
+
+On the business of getting any group of new applications integrated into ELKS, the following process can be followed.
+
+The application-set contributor would create a custom, say, 'busyElks.package' file for their contribution, which would use the package manager to build a bootable FAT or MINIX floppy image with their intended application-set. This would very quickly and easily allow the ELKS community users access to images for testing and comment. It is also quite necessary, since we're completely out of space for existing applications, yet alone new ones, in the 1440k image.
+
+In the elks/images directory, the application-set contributor would do something like the following to create a packages file once, from the already built application tree:
+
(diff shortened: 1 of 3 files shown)


jbruchon:
Would you mind documenting this capability somewhere before I merge? I'd like it to be findable by newbies to the project. A non-discoverable feature may as well not exist :-)

ghaerr:
@jbruchon, did you see my detailed writeup in PR #357 about how this should get used now?

This is a work in process, and the intent of the package manager for now is to make available multiple disk images outside the Config system, and enable quick user testing of various image sizes. Thus its operation is likely to change as it evolves. There has been considerable debate between myself and @mfld-fr as to whether this feature should remain or not, and exactly how, which is ongoing.

ghaerr:
> @jbruchon, did you see my detailed writeup in PR #357 about how this should get used now?

Not sure if you saw this... would you like the writeup I did in #357 in a .txt file in the Documentation/ directory?

jbruchon:
That would probably be good. I honestly don't have lots of time to devote to this project, so I am reviewing things quickly and don't catch everything all the time. Thank you for putting up with my occasional spaced-out-ness. :-)

ghaerr:
@jbruchon: There you go, documentation in Documentation/text/package-manager.txt.

jbruchon:
And with that, we've run out of pull requests!

PR #373, opened 2020-02-23
Modify exec.c to load and run bcc programs again

georgp24:
I made a small change to `elks/fs/exec.c`. Now programs compiled with bcc load and run again. Programs compiled with gcc-ia16 run as well as before.


georgp24:
Apparently I had made a last minute change which I did not test properly. In the end I just had to modify the sanity check in `exec.c` to get the code to support bcc executables.

As tkchia indicated he prepares additional changes for gcc-ia16 support. Therefore I would prefer to just leave it as it is for now.

ghaerr:
@georgp24: since this PR does not fix the issue you intend properly, and removes a required a.out header check, it should not be committed. Can you close this PR, since changes will be made in bcc instead?

There have been discussions to support non-split I&D, that can be added separately should it be needed.

ghaerr:
> So one might reasonably expect that an executable program with separate instruction and data segments will have `A_EXEC | A_SEP`, i.e. `0x30`.
> 
> However, I just downloaded some actual Minix 1 and 2 images [from David Given's Github page](https://github.com/davidgiven/minix2/releases). It turns out that the programs in the images only have `A_EXEC` in the third byte, i.e. `0x04200301`.
> 
> So I guess I was wrong: even if `0x04203001` is theoretically "erroneous", it was the actual historical practice, and we should probably make concessions to it. :-|

@tkchia: To summarize your findings: the A_SEP (0x10) bit is properly set for `gcc-ia16`, and not set properly by `bcc` and early MINIX programs, even though all programs compiled are actually separate I&D. Correct?

If this is the case, yes, we should consider just checking for the A_EXEC bit, (and always assuming the executables are actually separate I&D). This would eliminate future support for the executing proper non-separate I&D, (which isn't really a big deal). Correct?

tkchia:
@ghaerr: no, it is the `A_EXEC` flag that is set properly by `gcc-ia16`.  That is, `gcc-ia16` sets both `A_EXEC` (executable) and `A_SEP` (separate instruction and data) --- `0x04300301`.

`bcc` --- and whatever compiler that the historical Minix used --- only set `A_SEP` (which gives `0x04200301`).  However, `a.out` files with only `A_SEP` set were also treated as executable.

An executable file with non-separate instruction and data will always have `A_EXEC` set (but not `A_SEP`).  That will be `0x04100301`.

Thank you!

[_edited to add: I made a typo in my above write-up.  Sorry for the error._]

tkchia:
I have submitted an alternative pull request which should finally resolve this problem (https://github.com/jbruchon/elks/pull/375).  Thank you!

georgp24:
I downloaded the files tkchia has modified and elks and elksemu both support bcc executables now.

jbruchon:
Should this be closed?

ghaerr:
Yes. @tkchia 's PR fixed this issue in a different way.

PR #374, opened 2020-02-23
[elkscmd] `clock` now uses `clr_irq ()` & `set_irq ()` in `<arch/irq.h>`

tkchia:
This commit tidies up the code for the `/bin/clock` system utility, so that there is no longer a bunch of `#ifdef`'s just for disabling and enabling interrupts.  The `cli` and `sti` instructions are inlined under `gcc-ia16` --- as before --- but are now out-of-line, and thus a bit slower, if compiled with `bcc`.

I also slightly altered the port I/O code to use "pausing" versions of the port I/O operations, which should hopefully make the operations a bit more reliable.

(Side note: `<arch/irq.h>` should probably be cleaned up, and moved into `libc`, or explicitly shared between kernel and `libc`.)

### DIFF
--- a/elkscmd/sys_utils/clock.c
+++ b/elkscmd/sys_utils/clock.c
@@ -7,6 +7,7 @@
 #include <sys/time.h>
 #include <string.h>
 #ifdef __ia16__
+#include <arch/irq.h>
 #include <arch/io.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


jbruchon:
I'll merge this, but I think it'd be a good idea to consider shuffling the low-level code in this program into the kernel.

tkchia:
@jbruchon: thank you!

Issue #358, opened 2020-02-21
Distinguishing between ino_t and ... ino_t


tkchia:
(Follow-up to https://github.com/jbruchon/elks/issues/354 .)  Currently the type `ino_t` is used in ELKS to represent file serial numbers, both
  * for the kernel's internal processing, e.g. in `struct inode`, and
  * to communicate file serial numbers to userland, e.g. in `struct stat` and `struct dirent`.

(It is also used in documentation --- in `Documentation/text/minix_fs.txt` --- but this is a more minor problem.)

With the default ELKS build settings --- which only does 16-bit inodes --- I can write a program which reads the file's owner id, which will compile to something like this:
```
...
0000003B  8D46E2            lea ax,[bp-0x1e]    ; stat("/dev/null", &sb);
0000003E  50                push ax
0000003F  B80400            mov ax,0x4
00000042  50                push ax
00000043  E87E01            call word 0x1c4
00000046  FF76EA            push word [bp-0x16] ; printf("%#o\n", (unsigned)sb.st_uid);
...
000001C4  B81200            mov ax,0x12
000001C7  E93101            jmp word 0x2fb
...
000002E4  CD80              int 0x80
...
000002F4  89E3              mov bx,sp
000002F6  8B5F02            mov bx,[bx+0x2]
000002F9  EBE9              jmp short 0x2e4
```
The problem is that we should be able to run the same _binary_ program on another ELKS system, perhaps one which supports 32-bit inodes, and the program should still work.

I.e. if the program invokes syscall #18 (`0x12`), **it should reasonably expect that the `.st_uid` value will _still_ be at `[bp-0x16]` on the second ELKS system, not at some other offset into the structure**.

tkchia:
So to make the {user ↔ kernel} interface stable enough, one thing that will definitely be needed is a way to distinguish between the kernel's internal inode number type, and the inode number type(s) that it uses to interface with userland.

One possibility is to follow the approach used in Linux.  Namely, define `ino_t` differently in C headers used by the kernel, and C headers used by userland code.  (In particular: Linux 5.5.5 defines `ino_t` to `__kernel_ino_t` for the kernel.  GLibC does not use `__kernel_ino_t` to define `ino_t`.)

This means that from the kernel's viewpoint, a `struct stat` may have an `.st_ino` field that is _not_ of `ino_t` type, but some other type (e.g. `uint16_t` or `uint32_t`) actually expected by the user process.

There may be other possibilities.

[...]

PR #382, opened 2020-02-25
[abi] First pass at bug fixing and cleaning up dirent and stat in kernel and libc

ghaerr:
Expose kernel dirent struct in dir.h, remove from libc dirent.h, add sys/dir.h
	Fixed width stat and dirent structs using 32-bit u_ino_t, remove ino32_t
	Fix errors in dirent struct
	Fix error in stat.h kernel copy to userspace
	Fix errors in FAT ino_t handling

This first pass fixes internal kernel and FAT bugs and inconsistencies and paves the way for moving to POSIX-compatible `dirent` and `stat` structs. Next step if strict POSIX compliance is desired, a new kernel type would replace `ino_t` internally and then only use `ino_t` on external `dirent` and `stat` structs, replacing `u_ino_t`, which would affect a large number of kernel files.

[EDIT: A simpler next step would be to replace `u_ino_t` with `__u32` on `dirent` and `stat`structs, to eliminate a new type on external interfaces, at the risk of no longer having an identifiable type for tracking and bug fixing purposes.]

Also fixed typo in `deprecated` keyword from ignored compiler warning on committing PR #361.


mfld-fr:
Not standard compliant : `dirent` structure contains `ino_t`, not `u_ino_t`, and header is `dirent.h`, not `dir.h`. See https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/dirent.h.html for standard API.

tkchia:

> `dirent` structure contains `ino_t`, not `u_ino_t`

`u_ino_t` should work, as long as `ino_t` refers to the same type in user space.  So we can do something like
```
typedef __u32 u_ino_t;
#ifndef __KERNEL__
typedef u_ino_t ino_t;
#else 
...
#endif
```

> header is `dirent.h`, not `dir.h`.

Hmm... yes, indeed.  It seems that `<sys/dir.h>` is a pre-POSIX (and pre-C89) thing.  From your citation:

> Information similar to that in the `<dirent.h>` header is contained in a file `<sys/dir.h>` in 4.2 BSD and 4.3 BSD. The equivalent in these implementations of `struct dirent` from this volume of POSIX.1-2017 is `struct direct`.

GNU `libc` provides a `<sys/dir.h>`, but only as a kind of compatibility wrapper around `<dirent.h>` for old code.  I think we can do the same thing for ELKS, and have `<dirent.h>` directly include `__SYSINC__(dir.h)` instead.

Thank you!

mfld-fr:
> GNU libc provides a <sys/dir.h>, but only as a kind of compatibility wrapper around <dirent.h> for old code. I think we can do the same thing for ELKS, and have <dirent.h> directly include `__SYSINC__(dir.h)` instead.

Agree.

Edit : and rename `dir.h` to `dirent.h` in `linuxmt` also.

ghaerr:
Hello @tkchia and @mfld-fr, Thank you for your comments. Lets divide the issues up into two categories, 1) the name of the include files, and 2) the `ino_t` issue in the structs:

1)

 > GNU libc provides a <sys/dir.h>, but only as a kind of compatibility wrapper around <dirent.h> for old code. I think we can do the same thing for ELKS, and have <dirent.h> directly include `__SYSINC__(dir.h)` instead.
> 
> Agree.

This is exactly what the PR does: Userland code includes libc <dirent.h>, and libc <dirent.h> includes libc <sys/dir.h> for backwards compatibility. 


> 
> Edit : and rename `dir.h` to `dirent.h` in `linuxmt` also.

I named the kernel file that defines `dirent` dir.h instead of dirent.h to avoid confusion with the libc dirent.h to avoid confusion. So are you suggesting:

libc <dirent.h> includes libc <sys/dir.h> includes kernel <dirent.h> instead?

ghaerr:
Hello @tkchia and @mfld-fr:

> `u_ino_t` should work, as long as `ino_t` refers to the same type in user space. So we can do something like
> 
> ```
> typedef __u32 u_ino_t;
> #ifndef __KERNEL__
> typedef u_ino_t ino_t;
> #else 
> ...
> #endif
> ```

Would you like me to make the above change?

tkchia:
Hello @ghaerr,

> Would you like me to make the above change?

Yep, please do.

> This is exactly what the PR does: Userland code includes libc <dirent.h>, and libc <dirent.h> includes libc <sys/dir.h> for backwards compatibility.

No, I mean to flip it around: let `<sys/dir.h>` include `<dirent.h>`.

`<dirent.h>` will be the main (POSIX-compliant) header.  `<sys/dir.h>` just needs to be a simple shim to allow old pre-POSIX programs to work.

Here is the entirety of `<sys/dir.h>` from GNU `libc`:
```
/* Copyright (C) 1991-2016 Free Software Foundation, Inc.
   This file is part of the GNU C Library.

   The GNU C Library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2.1 of the License, or (at your option) any later version.

   The GNU C Library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
   Lesser General Public License for more details.

   You should have received a copy of the GNU Lesser General Public
   License along with the GNU C Library; if not, see
   <http://www.gnu.org/licenses/>.  */

#ifndef	_SYS_DIR_H
#define	_SYS_DIR_H	1

#include <features.h>

#include <dirent.h>

#define	direct	dirent

#endif	/* sys/dir.h  */
```
Thank you!

ghaerr:
> > Would you like me to make the above change?
> 
> Yep, please do.
> 
> > This is exactly what the PR does: Userland code includes libc <dirent.h>, and libc <dirent.h> includes libc <sys/dir.h> for backwards compatibility.
> 
> No, I mean to flip it around: let `<sys/dir.h>` include `<dirent.h>`.
> 
> `<dirent.h>` will be the main (POSIX-compliant) header. `<sys/dir.h>` just needs to be a simple shim to allow old pre-POSIX programs to work.

@tkchia, thank you, I understand now about libc <sys/dir.h>. 

Let me make the above two changes in this commit for your and @mfld-fr 's inspection.

ghaerr:
Hello @tkchia and @mfld-fr, please review.

tkchia:
Hello @ghaerr,

I think `<sys/dir.h>` still needs a `#define direct dirent` (as in GNU `libc`), since programs that use this header will expect to use a structure called `struct direct` (with a `c`).  But other than that, if everything runs, then it looks good to me.  Thank you!

tkchia:
Hello @ghaerr,

I think `<sys/dir.h>` still needs a `#define direct dirent` (as in GNU `libc`), since programs that use this header will expect to use a structure called `struct direct` (with a `c`).  But other than that, if everything runs, then it looks good to me.  Thank you!

ghaerr:
@tkchia here you go. Yes, I have tested in both inode size configurations.

mfld-fr:
Please rename `elks/include/linuxmt/dir.h` to `elks/include/linuxmt/dirent.h`, and squash your commits for history readability.

jbruchon:
I'll squash the merge, don't worry about that

ghaerr:
Ok @jbruchon, I've renamed the header as @mfld-fr suggested. Please squash the commits for me, thanks!

ghaerr:
Hello @tkchia,

I was reviewing your analysis #358 and noticed your earlier work on `ino_t` in userspace in the `dirent` and `stat` structs in your own fork. Now that we're using the 'ifndef \__\__KERNEL\__\__' trick you used there and suggested here, shouldn't I finish your excellent design and make that change here? Then we'd have the full POSIX compatibility that has been the goal.

tkchia:
Hello @ghaerr,

> shouldn't I finish your excellent design and make that change here?

I think your pull request already includes what I did, plus more (such as casting `-1` to `ino_t`, something I had not thought about).  So I would prefer to go along with your PR. :)

(Also, I had used `__u_ino_t` because I was worried at first that `u_ino_t` would intrude on the application program's namespace.  But then I checked the [POSIX.1-2017 documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html), and it turns out that _all_ names ending with ...`_t` are considered to be reserved for use by headers.  So it seems that my worry was unfounded, and `u_ino_t` is fine.)

Thank you!

tkchia:
Hello @ghaerr,

If you wish though, feel free to include the comments I wrote in my commit (https://github.com/tkchia/elks/commit/8a471a8ebdb91d64d0adb6d8062d923e13b4ee3e).  Thank you!

ghaerr:
@tkchia,

The issue that I bring up is that in this commit, the `dirent` and `stat` both structs use `u_ino_t` and not `ino_t`. With 'ifndef \__\__KERNEL\__\__' used, those structs can use `ino_t`. 

I was still suggesting using my commit, as it fixes other inode related issues within the kernel.

tkchia:
@ghaerr: I would guess that this is not a huge problem --- they are already the same type anyway in userland.  Thank you!

PR #385, opened 2020-02-25
[elks] remove CONFIG_SHORT_FILES checks in defining fd_set etc.

tkchia:
`Documentation/historic/CHANGELOG` says of ELKS 0.0.79, "Added `CONFIG_SHORT_FILES` which limits file descriptors to 16."  However, this is no longer a configuration option now, so there is no real need to check for it.

Removing this check simplifies the code for processing `fd_set`'s in `select(`...`)` calls, and removes a potential source of ABI instability.

### DIFF
--- a/elks/include/arch/posix_types.h
+++ b/elks/include/arch/posix_types.h
@@ -14,36 +14,27 @@
 
 #undef	__FD_SET
 #ifndef __KERNEL__
-#define __FD_SET(fd,fdsetp) {				\
-		int mask, addr = ((int) fdsetp);	\
-							\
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @tkchia,

I ran into this last night while looking at types.h. I agree with the simplification. 

There are more issues like this, for instance, CONFIG_EXEC_ELKS in elks/fs/exec.c.

jbruchon:
Good catch. These kinds of cleanups are highly desirable right now. There is a lot of junk like this littered around from before I was the maintainer.

PR #389, opened 2020-02-26
[busyelks] bugfix: elkscmd/Makefile

marcin-laszewski:
Fix conditional rules.
Should be: ('y' OR 'b'). Unfortunately, there are single condition only in Makefile (ifdef, ifndef, etc...).
Marc's solution (mfld-fr@f38921c) works ok.

### DIFF
--- a/elkscmd/Makefile
+++ b/elkscmd/Makefile
@@ -60,11 +60,11 @@ ifeq ($(CONFIG_APP_BUSYELKS), y)
 INSTDIRS += busyelks
 endif
 
-ifneq ($(CONFIG_APP_DISK_UTILS), n)
+ifdef CONFIG_APP_DISK_UTILS
 INSTDIRS += disk_utils
 endif
 
-ifneq ($(CONFIG_APP_FILE_UTILS), n)
+ifdef CONFIG_APP_FILE_UTILS
 INSTDIRS += file_utils
 endif
 
@@ -76,7 +76,7 @@ ifeq ($(CONFIG_APP_KTCP), y)
 INSTDIRS += ktcp
 endif
 
-ifneq ($(CONFIG_APP_MINIX1), n)
+ifdef CONFIG_APP_MINIX1
 INSTDIRS += minix1
 endif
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @mfld-fr and @marcin-laszewski,

I'm getting confused as to how the busyelks option is supposed to work. For testing, I manually edited .config to set the following values and ran make:
```
CONFIG_APP_BUSYELKS=y
CONFIG_APP_DISK_UTILS=b
CONFIG_APP_FILE_UTILS=b
CONFIG_APP_MINIX1=b
CONFIG_APP_MINIX3=b
CONFIG_APP_MISC_UTILS=b
```
That is supposed to include the busyELKS (symlinked) versions of each of the applets in each category, right? And NOT include the standalone versions of the applets, since they're the same filename and location (e.g. /bin/ls). 

Another issue is when busyelks is set to build via 'y', then answer 'y' to any of the above applet categories would also break the image build in the same fashion. Thus, when busyELKS is being built, there should only be the question of 'b' or 'n' for the above applet categories.

With this commit, the busyELKS versions *and* the standalone versions are built when option='b', thus overwriting each other in the install. In elkscmd/Makefile:
```
ifdef CONFIG_APP_DISK_UTILS
```
should be changed back to
```
ifeq ($(CONFIG_APP_DISK_UTILS), y)
```
so the standalone versions are only built when 'y' is selected, not 'b'.

Finally, busyelks/Makefile does generate a binary with the included commands, but busyelks.tar.xz does not include the symlinks:
```
Gregs-MBP2:busyelks greg$ ls -l busyelks*
-rw-r--r--   1 greg  staff  28164 Feb 27 20:39 busyelks.tar.xz
-rwxr-xr-x   1 greg  staff  46453 Feb 27 20:36 busyelks
-rw-r--r--   1 greg  staff      0 Feb 27 20:36 busyelks.map
Gregs-MBP2:busyelks greg$ tar Jtf busyelks.tar.xz 
./
./bin/
./bin/busyelks
Gregs-MBP2:busyelks greg$ 
```

PR #390, opened 2020-02-26
[libc] Split exec.c

marcin-laszewski:


ghaerr:
Nice split-up. Your work in getting these function definitions also placed in the standard libc header files is helping clean up our large compiler warning mess. Thank you!

PR #393, opened 2020-02-27
Kernel local heap & generic cache

mfld-fr:
**!!!DO NOT MERGE THIS PR!!!**

This is a RFC on adding a local heap to the kernel, to move from fixed to dynamic pools (for inodes, pipe queues, buffers and tasks), and thus to make room in the kernel data segment.

Please ignore commits on other parts than the kernel.


ghaerr:
While the idea of adding the ability to dynamically adjust the size of various kernel resources (particularly buffers) is very interesting, this code doesn't show any of that, it is just code for a heap manager. Since the kernel text size is already 52k (+55k bss), I hesitate at adding yet more code for managing memory, when there's already `mm_alloc`, which is used to managing user program code and data allocations. Why not use that with far pointers allowing greatly increased access to kernel buffers?

We're tight on code space in the kernel, and I would guess that most all the bss is kernel buffers, is that correct? Thus the biggest real data memory issue ELKS faces is almost all buffer use. By using the same memory manager for both buffers and user programs, we save considerably on code while allowing buffers to be anywhere, providing `gcc-ia16` supports far data pointers. The far pointer would only be used for the buffer address, not the other internal structures, to keep its use minimized?

Another thought is to use a slightly modified `mm_alloc` to work within a constrained data segment, i.e. the kernel data segment, and use that code to manage a varying-sized heap in the kernel just like it does for user program `brk` requests. That accomplishes what this code is doing without permanently adding more code to the kernel.

tkchia:

> By using the same memory manager for both buffers and user programs, we save considerably on code while allowing buffers to be anywhere, providing `gcc-ia16` supports far data pointers.

Yes, `gcc-ia16` does do far pointers --- see "IA-16 Named Address Spaces" in `elks/cross/share/info/gcc.info`.

Thank you!

mfld-fr:
Here are the kernel largest fixed pools:
```
.bss    0x0000000000001ba0  0x359e fs/fs.a(buffer.o)
.bss    0x0000000000005166  0x1d44 fs/fs.a(inode.o)
.bss    0x0000000000006eaa  0x1002 fs/fs.a(pipe.o)
.bss    0x0000000000008126  0xa46 arch/i86/drivers/block/blk_drv.a(doshd.o)
COMMON  0x0000000000008cbe  0x380 fs/fs.a(file_table.o)
COMMON  0x000000000000903e  0x44af kernel/kernel.a(sched.o)
COMMON  0x000000000000d50c  0x1b96 arch/i86/drivers/char/chr_drv.a(ntty.o)
COMMON  0x000000000000f0a2  0x485 arch/i86/drivers/char/chr_drv.a(tcpdev.o)
```
First is not the buffer L1 cache, as asserted above, but the task array, that cannot be moved out of the kernel data segment without much overhead on the task switching. 

Second is the buffer L1 cache, that is confused above with L2. As the task array, L1 cannot be moved out without high impact on all the block drivers and filesystems code. Far pointers only help for L2.

Dynamic allocation is to optimize kernel local data space usage. The very first step for dynamic allocation is a heap manager, with a good balance between simplicity and performance, so that it can be inserted in the remaining code space.

ghaerr:
Hello @mfld-fr,

Thank you for your detailed explanations, it helps a lot to understand your work. Using your fixed pool sizes as a guide, I looked more closely and found the following details of the largest kernel data allocated statically:
```
15 max tasks at approx 1k each = 15k
8-12 L1 buffers at 1k each = 8-12k
8 max pipes at 512 = 4k
96 max inodes at approx 77 bytes = 7k
3 consoles, 4 serial ports, 4 ptys = 11 ttys at 512 input buffer and 64 output = approx 6k
Total of just these items = 40k
```
Wow, 40k! You have opened my eyes! :) 

Are you thinking of using dynamic allocation for all of these largest items, allocated when needed and then freed afterwards? Since the allocation units greatly vary between a 77 byte inode and an L1 buffer, you'll likely have to use a 'fixed-size' pool allocator, that allocates fixed-sized blocks from linked pools for protecting against bad fragmentation, right?

If I'm imagining this the same way that you are, I can now see that this will greatly help allocate kernel data segment resources, based on the way they're dynamically needed, rather than configuration settings and very large static data. Would the overall kernel data segment 'break' be (re)allocated using `mm_alloc` so as to allow a truly dynamic sizing of the kernel data segment, based on the dynamic allocator's need for memory?

mfld-fr:
Kernel data segment has fixed size, exactly as user processes, and in both cases, 'breaking' has no effect. It just checks that the 'break' does not collide with the stack. Setup has to be modified to honor the 'maximum data size' in the kernel executable header, to allow 'chmem' - like setting.

Function 'mm_alloc' is actually a wrapper to the 'segment' allocator, that I rewrote some times ago to relax the constraint on the maximum number of segments, and to prepare dynamic allocation of the L2 buffers. It is light, reduces fragmentation with the 'best fit & coalesce' algorithm, but is inefficient. It has to be improved before supporting L2 with little additional code as possible.

Suggested 'pool' allocators are very efficient, but require more code space. I am also working on a generic cache layer to regroup inodes and blocks caching code, thus making room also in the kernel code segment for allocator improvements and dirent caching.

ghaerr:
Hello @mfld-fr,

Again, thank you for your explanations.

> Kernel data segment has fixed size, exactly as user processes, and in both cases, 'breaking' has no effect. It just checks that the 'break' does not collide with the stack. Setup has to be modified to honor the 'maximum data size' in the kernel executable header, to allow 'chmem' - like setting.

I hadn't realized that both user processes and the kernel data segment reserve the entire max data from the chmem/total data field and can't allocate any more (because the stack is above the heap in the same segment). I had been wondering whether Al Riddoch's possibly unfinished work with putting the stack below the heap using the (unusable with gcc-ia16 a.out header?) CONFIG_EXEC_ELKS option would help this issue. If not, perhaps that option should be removed, it is currently set default on, but our compiler never produces that a.out header.


> Function 'mm_alloc' is actually a wrapper to the 'segment' allocator, that I rewrote some times ago to relax the constraint on the maximum number of segments, and to prepare dynamic allocation of the L2 buffers. It is light, reduces fragmentation with the 'best fit & coalesce' algorithm, but is inefficient. It has to be improved before supporting L2 with little additional code as possible.

I see now, nice work. It looks like your CONFIG_MEM_LIST allocator also saves kernel data space. Do we still need the CONFIG_MEM_TABLE option, now that 286 protected mode has been removed?

mfld-fr:
Well... I am quite uncertain about what to do with that 'alternate' (aka 'ELKS executable format') data segment layout... on one side, we have to reserve space for the heap, on the other side, we have to reserve space for the stack and relocate 'data' and 'bss'... not really a game changer...

I am also unsatisfied by my current 'list' implementation of the segment allocator, because it is hardly compatible with the protected mode, so forbids using it to extend beyond the 1 MB barrier. Even within the 1 MB, protected mode can help to make ELKS more robust to SEGV.  So I am currently thinking about a mixed mode compatible with a size-limited GDT, but my ideas are not mature yet... so I prefer to keep the 'table' implementation as a reference for now.

ghaerr:
> Well... I am quite uncertain about what to do with that 'alternate' (aka 'ELKS executable format') data segment layout... on one side, we have to reserve space for the heap, on the other side, we have to reserve space for the stack and relocate 'data' and 'bss'... not really a game changer...

It seems to me the big benefit of allowing a strictly optional stack-under-the-heap memory layout would be that the memory requirements of variable heap applications would not have to be set in advance, or to a high amount with -maout-chmem= as is required now, and the 'break' could actually allocate more data memory to the process when needed. We are in the process of adding space for various apps, like `bc`, `sort`, and now `fsck`, which users are finding don't run well, which this could fix. IMO, this would also mandate a header field for stack size in addition, to enable tuning the default stack, because it wouldn't be able to be resized once running. In addition, this inverted layout should remain a compile time option, to allow those programs with large stacks or possible medium sized stacks and heaps to work growing towards each other as they do now. Agreed its hard to see how useful this might be, given it's frequency of need and the ability to set -maout-chmem=0xfffc now.

mfld-fr:
Just pushed a first draft of the generic cache in commit https://github.com/jbruchon/elks/pull/393/commits/565520c1d8d5bd78ec8b9d8390e0d5d3d5a03572. I am currently thinking about the best strategy to discard cache entries in low memory conditions.

ghaerr:
Hello @mfld-fr,

I want to make sure I fully understand how the cache system is intended to work, for consideration of your question as to what to discard.

The cache system is built on top of the new kernel data segment heap system. L1 buffers will use the cache system, correct? Doesn't the current buffer system decide on an LRU basis what to discard? So this cache system will replace the buffer system's LRU mechanism and discard buffers when not mapped/locked, correct?

Would other 'cached' items include inodes? I assume you're meaning caching inodes after calls to 'iput', so that the kernel can get them again in 'iget' without having to read the inode table? If so, this could work well on MINIX, but I can't guarantee it will work at all on FAT, since inodes are faked, and I haven't checked to see if there may be disk I/O underneath the inode structure directly to the MSDOS directory structure inside FAT. So it will be complicated.

For other cached entries, it only seems to make much sense to cache them if they are regularly created and destroyed but have persistence requirement. Task blocks, tty structs, pipe buffers, etc don't need caching but the kernel will benefit just using the new heap manager. I think that consideration for what to discard needs to look at the type of data in that structure, and some discussion of the regularity of its creation/destruction is needed. What other items than disk buffers and inodes (those desiring availability but having external persistence) are there? I think it would be good to start getting a real list together of the likely candidate structures for caching vs heap management, to help answer your question.

For strict caching, both an LRU list as well as a cache 'use count' that would be incremented might help set the priority for discard based on previous cached usage, like the root or /bin inode, for instance.

Given that IMO the majority of the improved cache/heap mechanism will end up being disk buffers, and thus will likely immediately get to a 'low memory' (all kernel data segment used up) state, it still seems to me that separate handling of disk buffers versus smaller objects would be in order.

mfld-fr:
Classify between cached and not cached is not enough to manage the memory as a shared resource. When allocating non-cached object from the heap, the kernel could stress the cache to make room, and it is hard to define a limit to that cache pressure to preserve I/O performance.

mfld-fr:
About the fake inodes, I had a look to the current implementation of FAT in Linux mainstream. It is using the inode cache, building the inodes on the fly, and just discard them. I see no problem to cache them ?

ghaerr:
> About the fake inodes, I had a look to the current implementation of FAT in Linux mainstream. It is using the inode cache, building the inodes on the fly, and just discard them. I see no problem to cache them ?

I was just saying that the FAT system currently does some direct directory-contents manipulation outside of the inode structure, and I haven't given thought as to whether that might conflict with inode caching, since the contents could change on disk without the inode cache knowing it. Case in point is `msdos_rmdir` and `msdos_unlink`, for example. We're probably ok.

jbruchon:
FYI: I'm doing a whitespace cleanup. If you'd like to replicate the results to minimize conflicts while you're working on this separate branch, run this command in each subdirectory of the project:

`grep -rl '\s$' * | while read LINE; do sed -i 's/\s\s*$//' "$LINE"; done`

mfld-fr:
After more thinking about the dynamic memory management design, I have to reconsider some initial assumptions... closing this one for now, would push some mature pieces separately.

ghaerr:
Hello @mfld-fr,

> After more thinking about the dynamic memory management design, I have to reconsider some initial assumptions...

Out of curiosity, what issues are you thinking about for reconsideration? I have been thinking more about this as a result of your RFC and remain interested in your thoughts. It seems still a good idea to get rid of some large blocks of static kernel data and allocate when needed, especially for pipes, ttys, possibly tasks, and other large structures (except buffers, which seem more complicated).

Issue #360, opened 2020-02-21
Shared libc code between kernel and user land


jbruchon:
I would really like to see at least the libc code shared among all processes in some way. Perhaps further out in development, but one of the driving factors behind me trying to write BusyELKS was how every single executable had its own copy of libc functions, and some programs were being written with write() purely to avoid the weight of printf().

If there were some way of the kernel providing libc and linking executables to the internal libc when loaded, that would be fantastic. I know that the segment limits become a major problem with such a thing. I suggested elsewhere that perhaps the kernel should take some pages from microkernels like Minix and split off some non-critical parts from the core such that they'll get their own segments. If the kernel was split into such components, libc could easily be one of them, and the kernel would simply link itself to the libc that is packed with it once the early boot code hands over control.

tkchia:
Hello @jbruchon,

I am curious to know if there are any existing systems that try to do such a thing.  How did they do it, and what were the kind of issues that faced?  I know Linux has a `vdso`(7) mechanism for exporting certain kernel functionality as userland code, but I do not know of any OS that goes further.

Thank you!

tkchia:
Hello @jbruchon,

Another issue I am wondering about is how this would interact with different function calling conventions.  I implemented a `regparmcall` calling convention (`-mregparmcall`, `__attribute__ ((regparmcall))`) for `gcc-ia16` last year, where functions will try to pass their arguments in registers `%ax`, `%dx`, and `%cx`, and arrange for the callee to pop any stack arguments.

`regparmcall` support is now complete for the MS-DOS platform.  I am currently considering adding support for a `regparmcall` calling convention to ELKS's `libc` as a way to save code space, and possibly getting the ELKS kernel to also use it internally.

Thank you!

ghaerr:
Hello @tkchia,

Your support for `regparmcall` could be very interesting to reduce kernel and application sizes, would setting a global kernel and application -m option allow that to be tried out?

With regards to some kind of minimal 'shared library' support - this could possibly be implemented through the notion of a 'secondary' text segment at ES: and using trampolines and/or stub functions to get in and out of it. If libc has any static data (I presume yes), then things get quite a bit more complicated. If the call chain allows procedure calls back out of the secondary segment, also a big complication. Then there's all the ES protection wrappers to consider as well.

[...]

PR #398, opened 2020-02-29
[libc] [elkscmd] add support for `stdcall' calling convention, & try building `bc' using this convention

tkchia:
These commits update the ELKS C library, and its `make` scripts, to support using the `stdcall` function calling convention --- where function arguments are passed on the stack, and the callee function (if non-variadic) cleans the stack.  The `gcc-ia16` option to use this convention is `-mrtd`.

The build procedure in `elks/libc/` will now build and install versions of the C library (`libc.a`) and startup module (`crt0.o`) for both the `stdcall` _and_ the default `cdecl` calling conventions.

As a quick test, I also modified the build procedure for the `bc` program to use the `stdcall` convention.  This shrank the size of `bc` from 38,124 bytes to 36,860 bytes --- and the program still works.

(As I said earlier (https://github.com/jbruchon/elks/issues/360#issuecomment-590373973), I hope to also have the C library support the `regparmcall` calling convention, which should lead to even smaller and faster code.)


ghaerr:
Hello @tkchia,

Nice work! I can see from your testing that `stdcall` could ultimately save quite a bit of space in user land, and possibly the kernel.

However, I have a big concern: has anyone looked at all the many compiler warnings that are spit out when compiling `elkscmds/`? With libc forced `stdcall`, a single missing include in an application would generate `cdecl` calling convention on a `stdcall` function, with crash results. If I remember correctly, there are tons of missing includes in the current command compilations. Perhaps this needs to be default off until it is assured that the applications are properly and fulled cleaned up?

A comment: I see in the various 'fast' asm versions of string routines, the following construct is now used:
```
#ifdef __IA16_CALLCVT_CDECL
 	ret
 #elif defined __IA16_CALLCVT_STDCALL
 	ret $6
 #else
 #error "unknown calling convention"
 #endif
```
Given that we likely need to port quite a few more string routines in to assembly for speed, would it be possible to use a macro or define (with pop count argument), for returning from functions without all this messiness?

tkchia:
Hello @ghaerr,

> Perhaps this needs to be default off until it is assured that the applications are properly and full[y] cleaned up?

It is still not the default: with my commits, currently only the `bc` application is compiled with the `stdcall` convention.  The other programs are still compiled with the `cdecl` convention.  I agree that we should go through each application to see if it might break when using `stdcall` (and also spot and iron out any C library bugs that might remain).

> would it be possible to use a macro or define (with pop count argument), for returning from functions without all this messiness?

Yes, definitely.

Thank you!

ghaerr:
Hello @tkchia,
 
> It is still not the default: with my commits, currently only the `bc` application is compiled with the `stdcall` convention. The other programs are still compiled with the `cdecl` convention. I agree that we should go through each application to see if it might break when using `stdcall`.

I am sorry, I misunderstood and thought that your proposal was to compile only `libc` with a specified calling convention, and then have applications link with it, possibly using a different calling convention. 

I now understand that ELKS libc is being compiled as a multilib, and applications can specify their own compilation, as well as the selection of the appropriate multilib, with the CFLBASE += -mrtd option. Thus, the entire application and libc, rather than either portion, would always use the same convention. Correct?

That should work great and allow each application to be cleaned up individually for both include files and calling conventions. When selecting `stdcall`, it remains very important that forward declarations of libc calls be in place, or a crash will result (but the compiler warning will be there).

Thanks!

PR #400, opened 2020-02-29
[libc] bugfix: `stdcall': variadic syscall funcs. should pop no args.

tkchia:


ghaerr:
Hello @tkchia,

I am curious, how did you find this bug? Was it leaving the stack corrupted on all calls to system call variadic functions and causing a crash, or was this just observed during your `stdcall` work?

I am still seeing some strange application bugs here and there, and trying to understand more specifically which system calls this bug might have affected.

Thanks!

tkchia:
Hello @ghaerr,

> Was it leaving the stack corrupted on all calls to system call variadic functions and causing a crash

Not really --- I just happened to remember that I got the pop count wrong.

> I am still seeing some strange application bugs here and there, and trying to understand more specifically which system calls this bug might have affected.

Can you elaborate a bit more?  Thank you!

tkchia:
Hello @ghaerr,

The syscall routines that accept variable numbers of arguments will be `open`, `fcntl`, and `ioctl`.  These are listed with a `!` in the master syscall list in `elks/elks/arch/i86/kernel/syscall.dat`.

It happens that all these syscalls each accept 2 words of arguments plus an optional third word.

Thank you!

ghaerr:
> Can you elaborate a bit more? 

Running `ls -l / | sort` shows memory corruption, for instance. I have seen other cases but can't remember specifics. I haven't looked at sort source code yet. And there was the problem with having to rewrite around `realloc` to fix the `ls -F` bug; it was never determined exactly where the original error was, or whether the original source actually worked for that option.

Knowing that the bug affects `open`, `fcntl` and `ioctl` helps, thank you. That gives me more information should I run into further problems with applications running strangely (but fixed now).

tkchia:
Hello @ghaerr,

Well... actually I will be surprised if this fix somehow works on `ls -l / | sort` --- it only fixes a bug in the code for the `stdcall` calling convention, which is (for now) used only by the `bc` program.

Probably the bug with `ls -l / | sort` lies elsewhere.  Perhaps I will try running `elksemu` to see if I can get any clues.

Thank you!

PR #403, opened 2020-02-29
[elksemu] `stat' now gives 32-bit inode nos., per new kernel ABI

tkchia:


ghaerr:
It looks like `elksemu` has to keep its own copies updated of various ELKS structures. Is the problem that ELKS headers aren't easily exported, or is the problem deeper than that?

tkchia:
Hello @ghaerr, `elksemu` needs to be built for the host system (Linux/x86-32 or Linux/x86-64), while the ELKS headers are for the target system (i.e. ELKS on x86-16).  Thank you!

PR #406, opened 2020-03-01
Hd script

georgp24:


mfld-fr:
This PR contains binary files without the corresponding sources, so cannot be modified or regenerated by another contributor. Moreover, the design is not consistent with the current configuration and image generation workflow, and conflicts with the current HD image generation.

georgp24:
I mentioned in the script that the sources are included in the DEV86 package. However, they cannot be compiled with gcc-ia16 so I did not include them.

You modified this script on the  4th of Mar 2019 indicating that it is broken. I fixed it and now you can generate bootable hard disks again. 
If you manage to provide a better solution there will be no need for this script. Until then, why not use it?

ghaerr:
Hello @georgp24,

We're all interested in getting HD image generation working, and thank you for working on it.

However, I agree with @mfld-fr, we can't accept this PR for two reasons: 1) we need source code for the binaries, so that it can be built, understood and possibly modified. I, for one, don't actually understand what your solution is, or how it works, because I can't read any source. 2) We're no longer using 'root' privileges and 'mount' that use the host (only Linux) OS filesystem to build images. I spent quite a bit of time last year writing tools to build filesystems that will run on any OS. I'm running OSX and can't use your script to build an HD image.

Perhaps you can look in image/Makefile and see how `mfs` is used to create a MINIX image, and `setboot` to write a boot sector, and then explain a bit more about what needs to happen to support HD image creation using those (possibly modified) tools, and I'll help you to get HD images generation working. If you can point out where the sources you're using are in `dev86`, that will help too.

georgp24:
The script uses the boot blocks ELKS used for hard disks before the move to gcc-ia16 and I wanted to avoid that this caused you to need an as86 compiler to compile ELKS - just to generate 2 kb of binaries. Anyway, I will provide the sources in another commit. But I will not rewrite them from Intel to AT&T syntax.
The script only uses linux commands like echo, mount, umount, sfdisk, losetup, mkfs, mkdir, cp and if you can develop for ELKS on OSX  I think you will have these available on your system. "All" the script does is to generate a bootable Minix hard disk image with these bootblocks and copy the files from the fd1440 image to that. I came up with this this script in 2017 and Jody rewrote it a few months later. After the move to gcc-ia16 it did not work anymore.

ghaerr:
> The script uses the boot blocks ELKS used for hard disks before the move to gcc-ia16 and I wanted to avoid that this caused you to need an as86 compiler to compile ELKS - just to generate 2 kb of binaries. Anyway, I will provide the sources in another commit. But I will not rewrite them from Intel to AT&T syntax.

We won't be able to use the as86 sources if not rewritten, so just point out where the sources are in `dev86` and others can do the rewrite.

> The script only uses linux commands like echo, mount, umount, sfdisk, losetup, mkfs, mkdir, cp and if you can develop for ELKS on OSX I think you will have these available on your system. 

No Georg, there is no mount or umount for MINIX on OSX, nor is there losetup or mkfs. Please look at how image/Makefile is doing it for further reference.

> "All" the script does is to generate a bootable Minix hard disk image with these bootblocks and copy the files from the fd1440 image to that. I came up with this this script in 2017 and Jody rewrote it a few months later. After the move to gcc-ia16 it did not work anymore.

I still don't understand, but that's ok. I was looking for a top-level view of how exactly our current process of creating images needs to be modified to support HD (MBR?) images.

georgp24:
The script shall run in the elks/image directory on a Linux system and not in ELKS.

The sources are here: https://github.com/jbruchon/dev86/tree/master/bootblocks
mbr.s will compile to mbr.bin, minix.c is the source for minixhd.bin and minix_elks.c for the boot file. You copy mbr.bin into the boot sector of the HD, minixhd.bin into the first two sectors of the partition and that calls the boot file in the boot directory which loads the linux kernel in the boot directory.

georgp24:
Maybe the Readme file is helpful here: 
https://github.com/lkundrak/dev86/blob/master/bootblocks/README
```
Minixfs boot block

   This boot block has varients for floppy and harddisk and works similarly
   for both. For the hard disk it must be installed in the partition boot
   block with a normal MBR in sector zero of the disk. This boot sector can
   be installed with makeboot or simply by copying the 1k file to the start
   of the partition (or floppy) to be booted.

   The sector looks for a file or directory called 'boot' if it's a
   directory it loads that and does the search again. When it finds a
   file it loads it at location $10000 and executes it, note this
   is limited to a file size of 519k.

   There is also support for a helper boot which mean this is the only
   boot sector able to load an ELKS image (almost) directly.
```
```
Booting Elks kernel

   Only the minix boot sector can directly boot an elks kernel and even that
   needs a helper function because of the complexity. The helper is called
   'minix_elks.bin' and needs to be copied onto the disk as '/boot/boot'
   with the ELKS image copied on as '/boot/linux'. This works, with the
   correct boot block, on either floppy or harddisk.
```
This is the reason I put a partition table on the disk, the mbr.bin file into sector zero of the disk and the boot block called minixhd.bin to the start of the partition. Then generate a boot directory with the files boot and linux in it. Plus all the other directories and files you would put into FD1440.bin.

Mellvik:
It seems to me that the HD boot issue has become more about old vs new code rather than functions. Most of us know this by heart - the PC boot process is very simple: the MBR (disk sector 0) has the partition table and the code to check which partition is active, then load the first sector of that partion and branch off. All OS independent. Which means we can boot a DOS floppy, rund FDISK /MBR to clean sector 0 and (still) be happy. Regardless.

What that secondary boot code does is OS dependent, and we can do whatever we want for ELKS w/o any consequences 'outside'. The normal is reading the root dir of the filesystem on the partition, find some predefined file, or - many unixes did this and I suspect Linux did too - issue a single character prompt, wait 2-5 seconds to give the operator/user a chance to act, then continue by loading the default OS file (vmunix, unix, linux, boot, watever).

This challenge should really be the least of our worries regarding ELKS. Most of you can fix/create the assembly code for the 2nd stage loader in 30 minutes.

--Mellvik

jbruchon:
Are we closing this for now?

ghaerr:
Yes.

ghaerr:
We will get what @georgp24 wants using another method that works on all systems.

ghaerr:
Hello @georgp24,

I have rewritten your shell script to work the way that ELKS images are currently built, so that you can see how its done and also possibly help complete it for full MBR support. Once you have configured the latest source to build an HD image (setting CONFIG_IMG_HD), this script can be used to separately build a 'flat' HD image from scratch, and then use that to add an MBR sector from `dev86` in front of it.

It seems that the 'mbr.bin' boot sector from `dev86` needs to be partitioned using `sfdisk`, which OSX doesn't have. The attached script tries to boot but can't find an active partition, so I'm not able to continue. Perhaps you can play with it and test having it create an active partition by running `sfdisk` as indicated. We can then determine whether we want to add partition code into an MBR sector or not, as it is a bit complicated, because the CHS needs to be known as well as the start and ending filesystem sectors.

@tkchia, after looking more deeply into the many mbr*.s variants in `dev86`, it appears that all of them depend on an external tool to actually partition a disk and then write the partition table. So I'm wondering how much utility there really is in building just a portion of the solution, even using `as86` to assemble it. That's when I thought perhaps this script would be a good idea for @georgp24.

The attached script builds a very minimal system, with only /linux and /bin/sh, for ease of possibly poking around with /linux. Later, a full system image can be easily built by uncommenting the `mfs genfs` line as indicated.
```
# Name of flat HD image to create
HD=testhd.bin
SIZE=32200

# use current ELKS boot sectors and linux
# (configure CONFIG_IMG_HD and make)
BOOTSECT=bootblocks/minix.bin
LINUX=target/linux

# MBR image (remove exit below)
# assemble dev86/bootblocks/mbr.s with 'use512=1' internally:
#   cd dev86/bootblocks; make mbr.bin
MBR=mbr.bin
# output MBR HD image
MBRHD=mbrtest.bin

# minimal system: just /linux and /bin/sh
# mkfs HD image
mfs $HD mkfs -n14 -1 -s$SIZE
# copy linux
cp $LINUX linux.bin
# write ROOT_DEV as 0301 (0103 little endian) at position 0x1FC=508 in linux
#echo -en \\x01 | dd of=linux.bin bs=1 count=1 seek=508 conv=notrunc
mfs $HD cp linux.bin /linux
rm linux.bin
# mkdir /bin and copy sh and ls
mfs $HD mkdir /bin
mfs $HD cp target/bin/sh /bin/sh
mfs $HD cp target/bin/ls /bin

# full system (comment out minimal system above)
#mfs $HD genfs -n14 -1 -s$SIZE genfs target

# mkdir /dev and create devices
mfs $HD mkdir /dev
make -f image/Make.devices "MKDEV=mfs $HD mknod"

# make image bootable
setboot $HD $BOOTSECT

# list image root directory
mfs $HD ls -l /

# $HD is bootable flat image
# run qemu.sh with IMAGE="-hda testhd.bin"
exit

# combine MBR with bootable flat HD image
cat $MBR $HD > $MBRHD

# make partition table
echo ',,80,*' | sfdisk --quiet -u S $MBRHD

# should have MBR bootable system
# run qemu.sh with IMAGE="-hda mbrtest.bin"
exit

# not sure these are needed
#write hard disk boot block to sector 0 in partition 1 of the hard disk image
#dd if=$BOOTSECT of=$MBRHD seek=2048 bs=512 count=2 conv=notrunc
#write mbr into disk sector zero. Length of 0x1B5==437 = 19*23
#dd if=$MBR of=$MBRHD bs=23 count=19 conv=notrunc
```

georgp24:
Hi Greg,
I have packed my bags and will be travelling for a week so I cannot work on that.
Georg

georgp24:
Hi Greg,
I am not familiar with `mfs` and `setboot`. Also I would need the bootsectors from `dev86` . If I had the time I would follow my script manually and try to do the same with `mfs` and `setboot` . Using a hex editor after each step to see if the image generated is the same. You do not need a loopback device if you can write the files, which would usually be written to the 1.44 floppy, to the hard disk image directly. If there is no `sfdisk` usage allowed, one could use `dd` to write a prepared partition table to the disk image. So `menuconfig` would just allow hard disks of 10, 30 and 60 MB in size and the partition table for one of these three disk sizes is then written with `dd` to the disk image.

Currently most of my attention gets the coronavirus problem and the curfew here.

Georg

Issue #404, opened 2020-02-29
umount does not leave MINIX-fs in a 'checked' state


pawosm-arm:
**Description**
I've mounted MINIX-fs in `/mnt` then un-mounted it, then mounted it again. Turned out, MINIX-fs became 'unchecked' after `umount`.
- What is expected ?
As a 'checked' MINIX filesystem was mounted, it should remain 'checked' after `umount`.
- What is observed ?
`umount` does not leave MINIX-fs in a 'checked' state.

**Configuration**
- Version (commit or release):
939241982f3c80ab93178fba8fc35cb23163726a
- Emulator or HW platform:
Amstrad PC 2086

**How to reproduce ?**
- Does the problem always happen ?
Yes
- List the steps to reproduce the problem.
With rootfs mounted from floppy, I'm mounting MINIX-fs installed in XT-IDE CF card adapter device, directly on `/dev/bda` (i.e. there's no partition table).
1. login as `root`
2. `mount /dev/bda /mnt`
3. `ls /mnt` - should list files and directories on CF card
4. `umount /dev/bda` .or. `umount /mnt`
5. `ls /mnt` - should list no files or directories
6. `mount /dev/bda /mnt` - it prints:
```
MINIX-fs: mounting unchecked file system, running fsck is recommended.
```

The filesystem is mounted regardless, e.g. I can see the files and directories doing `ls /mnt`.

After I'd un-mounted it again, I tried to run `fsck /dev/bda` on it. Unfortunately, it refuses to operate as such:
```
sys_brk failed: len 41360 > endseg 33168
fsck: Unable to allocate buffer for zone count
```

This system has 640kB of ram, no EMS or high memory of any kind. `CONFIG_MEM_PAGES=10` is set in the `.config`.

While booting, ELKS prints:
```
640K base RAM.
525 K of memory for user processes.
```

**Raw data**
- Screenshot(s):
- Log file:

**Additional information**
Add any other context information about the problem here.

ghaerr:
@pawosm-arm, thank you for your bug report.

> I've mounted MINIX-fs in `/mnt` then un-mounted it, then mounted it again. Turned out, MINIX-fs became 'unchecked' after `umount`.

It looks like the MINIX superblock VALID_FS flag is being improperly reset in `elks/fs/minix/inode.c` function `minix_write_super`. I don't yet understand why it isn't setting the flag instead, nor why remounting a previously written MINIX filesystem doesn't show this message an a reboot, for instance.

[...]

PR #408, opened 2020-03-01
[minix] Fix buffer problems and filesystem check status in MINIX filesystem

ghaerr:
This fixes the MINIX filesystem check status bug brought up in #404, and corrects and cleans up various other buffer issues in the MINIX filesystem.

### DIFF
--- a/config/lxdialog/Makefile
+++ b/config/lxdialog/Makefile
@@ -27,6 +27,7 @@ CFLAGS = -std=c99 -O2 -Wall -pedantic -DLOCALE $(CURSES_INC)
 
 LDFLAGS = -s
 LDLIBS = $(shell pkg-config ncurses --libs)
+#LDLIBS = /usr/local/opt/ncurses/lib/libncurses.a
 
 #########################################################################
 # What do we compile?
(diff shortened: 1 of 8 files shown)


mfld-fr:
It would be easier to distinguish actual bug fixes and spacing changes if you could stop to mix them within the same commit...

jbruchon:
I'm squashing mutli-commits anyway, so it does not matter.

mfld-fr:
@jbruchon : True after the merge, false when reviewing before the merge.

ghaerr:
Hello @mfld-fr,

> It would be easier to distinguish actual bug fixes and spacing changes if you could stop to mix them within the same commit...

I understand what you mean, and will try to do so in the future. It's just that with the mixed spaces and tabs approach used in a lot of the source, it's hard to understand the whole of the code without reformatting it to be completely understandable. Would you mind various other commits just for reformatted readability in the future?

With regards to the bug fix itself - the problem wasn't the unmap_buffer call at the end of read_super. Instead, a pointer to a mapped buffer was saved in the memory superblock, which was then used to access the filesystem checked flag when remounting or unmounting. I removed that pointer and instead remapped the buffer in both those cases, while also cleaning up the code considerably by putting most of that into a new routine, minix_set_super_state. Another flag was added to the memory super_block to save the super check state to save an additional map buffer in remount. The rest of the commit is cleanup and reformatting.

mfld-fr:
OK, got it, thanks.

PR #411, opened 2020-03-05
[kernel] Fix kernel sbrk system call

ghaerr:
This fixes a major bug in the ELKS kernel, which affected all programs that use `sbrk`, including programs calling `malloc`, `execvp`, and `execve`, and more.

`sbrk` is supposed to return the start address of allocated memory, instead it was returning the address just above the allocated memory. Thus, any program using `malloc` or `realloc` would end up with a pointer to bad memory on a request for more memory. It looks like this bug has been around for almost two years. I ended up finding it by actually using both `sash` and `sh` and noticed the system would crash after a few commands were run.

The C environment was being trashed by shell calls to `exec`, `sort` displayed corrupted data, and I'm sure this was the problem with the `realloc` call in `ls` earlier. All these problems are now fixed.

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -30,7 +30,7 @@
 #include <arch/segment.h>
 #include <asm/seglist.h>
 
-#define MIN_STACK_SIZE 0x1000
+#define MIN_STACK_SIZE 0x1000	/* 4k min stack above heap*/
 
 #ifdef CONFIG_MEM_TABLE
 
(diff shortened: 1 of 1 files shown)


jbruchon:
I knew there was something wrong related to this call a long time ago. Thanks for tracking it down. It could be responsible for tons of inexplicable issues. I'm committing it immediately. If anyone has problems, we can always revert later.

ghaerr:
Thanks, it took a while to find it. I ended up using the `brk` debug message you put in the kernel long ago, which helped!

jbruchon:
Now you know why it got put there! I was so frustrated with that. I was nowhere near as skilled in C back then as I am today, so I couldn't identify the problem after quite a bit of poking around and I gave up. I'm glad you said it was difficult. Now I know that I was justified in dropping my efforts (and I think something else was committed that made me think the issue was fixed.)

mfld-fr:
Nice spot !

PR #415, opened 2020-03-05
[libc/misc] Split ltostr.c, use atol as atoi.

marcin-laszewski:


ghaerr:
@jbruchon: the mismatch of this commit and mine has caused a build failure. I will submit a PR shortly.

jbruchon:
@ghaerr I noticed; thanks for taking care of it.

PR #416, opened 2020-03-05
[cmds] First cleanup of various commands not compiling

ghaerr:
Initial cleanup of elkscmd/. Removes and rearranges the non-working config options into their own set, turned off by default.

@jbruchon: Please commit this after my two previous commits #412 and #413 (sash and RAW fixes), since there is a possibility of merge conflict with elkscmd/config.in. I will fix it here if there is a merge conflict, thanks!


ghaerr:
@jbruchon, I will close this and open another PR right now, due to the changes @marcin-laszewski made after this.

jbruchon:
OK. Sorry for making things more difficult.

Issue #288, opened 2020-01-25
panic: No init or sh found


alejandrocobo:
Hi. I've tryed ELKS 0.2.0 on my 8088 with 640kB of RAM laptot and I am unable to run it. I downloaded 0.2 version because it is precompiled. I tried to install gcc-ia16 to compile 0.3.0 on Debian but something is missing in intkchia repositories.

The PC has no hard drive; just two dd floppy drives. I tried both "comb" and "comb_net" with the same result. The errors I saw where these:

Console: Direct 80x25 emulating vt52 (3 virtual consoles)
PC/XT class machine, NEC V30 CPU
640KB base RAM
[...]
Partition check:
bda:hd: error: AX=0x0
BIOSHD:I/O error
dev 300, sector 0
unable to real particion table
bdb:hd: error: AX=0x0
BIOSHD:I/O error
dev 340, sector 0
unable to real particion table
ELKS version 0.2.0
[...]
Unable to open /dev/tty (error 2)
dup failed
No init - running /bin/sh
hd: error: AX=0x0
BIOSHD:I/O error
dev 380, sector 0
sys_execve("/bin/sash", args, 18) => 2.

panic: No init or sh found
apparent call stack
[...]
SYSTEM LOCKED

Thanks!

mfld-fr:
Hello @alejandrocobo, the release 0.2.0 is quite old and numerous bug fixes & improvements have been made since. Could you please retry with the latest master ? What is missing in @tkchia repositories ?

alejandrocobo:
apt-get sais something like (this is a free translation from Spanish) "the repository "http://ppa.launchpad.net/tkchia/build-ia16/ubuntu" doesn't have a release file". 

And then "The repository data cannot be authentified so it use is highly risky".

Finally "Fail to obtain  http://ppa.launchpad.net/tkchia/build-ia16/ubuntu/dists/focal/main/binary-amd64/Packages 404 Not found".

I'm on Debian 10 and following these steps: https://launchpad.net/~tkchia/+archive/ubuntu/build-ia16

BTW, why there is no precompiled disk images for 0.3.0 like in 0.2.0? That would easy the use of ELKS for those who don't have Linux or don't have the development chain.

Thank you very much!

mfld-fr:
@alejandrocobo : Okay, thanks for the additional info.

@tkchia : Would you be so kind to have a look ? It seems related to your binaries packages you pushed some times ago.

tkchia:
[...]

PR #422, opened 2020-03-06
[boot] [elks] allow bootloaders to load /linux as one blob

tkchia:
Previously, bootloaders have to do the job of splitting up `/linux` into its "setup" portion (which goes to `INITSEG:0` and `SETUPSEG:0`) and its "kernel" portion (`SYSSEG:0`) in memory.

With this change, a bootloader can load `/linux` as one single binary blob at `INITSEG:0`, and inform `/linux`'s "setup" portion about this (via a new `elks_flags` field in the loaded `/linux`).

This helps simplify the logic (and shrink the sizes) of the Minix and FAT filesystems' boot code.

### DIFF
--- a/elkscmd/bootblocks/boot_minix.c
+++ b/elkscmd/bootblocks/boot_minix.c
@@ -9,7 +9,7 @@
 
 // Global constants
 
-#define LOADSEG 0x1000
+#define LOADSEG 0x0100
 
 // Global variables
 
--- a/elks/arch/i86/boot/bootsect.S
+++ b/elks/arch/i86/boot/bootsect.S
@@ -563,8 +563,8 @@ root_flags:
 	.word ROOTFLAGS
 syssize:
 	.word SYSSIZE
-unused:
-	.word 0
+elks_flags:
+	.word ELKSFLAGS
 ram_size:
 	.word RAMDISK
 vid_mode:
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -23,6 +23,9 @@
 !   This code is called after the BIOS-POST and replaces
 !   the BIOS OS/loader
 !
+! changes to support having setup + ELKS kernel as one single blob
+! March 2020 https://github.com/tkchia
+!
 ! The following data is passed to the main kernel (relative to INITSEG)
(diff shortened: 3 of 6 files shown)


ghaerr:
Nice work @tkchia!!

Could we also use another flag bit in elks_flags that would allow, say, the lower byte of elks_flags to hold the BIOS disk number for setup, so that setup could then set ROOT_DEV properly? This would solve the HD boot problem in a couple threads, if I understand correctly.

ghaerr:
Hello @tkchia,

Also, it would be nice to have magic numbers like 0x1f7/0x01F7 placed in a header file for readability and compatibility between the boot sources and setup.S.

Thank you!

elkscmd/bootblocks/boot_minix.c
```diff
@@ -9,7 +9,7 @@
 
 // Global constants
 
-#define LOADSEG 0x1000
+#define LOADSEG 0x0100
```
mfld-fr, line 12: This is a risky change. I already mentionned in https://github.com/jbruchon/elks/issues/288#issuecomment-581034966 that some BIOSes are intolerant with LOADSEG not aligned to 64K. You should better keep loading the 'blob' to 0x1000:0 and then split it just after.

tkchia: Hello @mfld-fr, my suspicion is that at least some of the errors were happening because the reads were crossing a DMA boundary.  I had (earlier) modified `disk_read` to fall back to reading just 1 sector whenever a multi-sector read fails, which should allow such reads to work as long as the buffer is 512-byte aligned.

tkchia: Maybe I will see if I can tweak the code to read to `0:0x1000` (the same physical address) instead of `0x0100:0`.  This is relatively easy to do for the FAT bootloader, but a bit harder to do for the Minix FS bootloader.

tkchia:
Hello @ghaerr,

> Could we also use another flag bit in elks_flags that would allow, say, the lower byte of elks_flags to hold the BIOS disk number for setup, so that setup could then set ROOT_DEV properly?

I will look into that.  Thanks!

PR #424, opened 2020-03-06
[libc] Split termcap.c

marcin-laszewski:


ghaerr:
@marcin-laszewski: Thank you for your tireless work in splitting up libc routines so that extra routines aren't called in needlessly. As I mentioned in another post, adding in undefined libc routines into the appropriate headers is helping to reduce application compilation warnings, to keep that up too!

I notice that you misspelled '-Demacs' in libc/termcap/Makefile. Since the '#ifndef emacs' check was removed when you split out the routines, this -D and $(EMACS) check don't look like they are necessary anymore, and can be removed in a future commit.

PR #426, opened 2020-03-07
Add tool for unreal mode testing

georgp24:
The unreal16 utility enables unreal mode to allow 16 bit applications to access 4 Gbyte of memory. Requires a 386 processor.


ghaerr:
Unreal mode seems potentially interesting, but of course only works on 32-bit processors. I have a few questions:

Can the memory locations of the GDT be overwritten, once the program exits? That is, does the system continue to work after running `unreal16`, when other applications overwrite that space?
In protected modes, this is not the case, but I'm unfamiliar with unreal mode.

The following lines of `unreal16`:
```
# Unreal mode is identical with real mode with one exception: 32-bit
 # addresses greater than 0xFFFF are allowed (they do not cause INT 0x0D,
 # as they do in true real mode).
 #
 # Unreal mode does not work with the CS or SS registers; nor with V86 mode.
```
Do you have sample code that demonstrates the advantage of unreal mode, that is, data segments > 64k? What exactly can be run after running `unreal16` that demonstrates its usefulness? The ELKS kernel can't be easily changed to support it, since it would require far pointers everywhere and then only work on 32 bit systems. Thus, is this limited to a demo application(s)?

The utility should probably be in sys_utils/, but that can easily be changed later.

georgp24:
I run another program that used 32 bit offsets after unreal16 terminated and that worked fine. I assume ELKS will have loaded that in the same memory space that unreal16 used before and did overwrite the GDT. The operation for unreal mode is as follows:

> switch to protected mode temporarily, just to define segment descriptors with 4GB limit (as we would to access the whole memory in protected mode), load the appropriate selectors to all data segment registers, thus changing their descriptor cache, and drop back to real mode by flipping the PE bit back to 0.

> With the PE bit being 0, addresses are formed exactly like in regular 16bit real mode with segment and offset, but now the descriptor caches specify that our segments are 4GB long, allowing us to access the whole 32bit address space with zeroing out all segments and using the whole 32bit values as offset (for example using es:edi with es being 0, and edi holding any 32bit value.

So you set the segment to zero and then use a 32bit offset which allows you to access up to 4 GB.
An example how to use this is included in unreal16, it writes to video memory this way. Setting es to zero and use the edi register to access the display adapter (es:edi). Or use es:0xb8296 to poke a byte into it.

There is no memory protection, the applications have to agree on the areas they use. In DOS the emm386.exe driver enabled unreal mode and XMS managed the memory access.

tkchia:
Hello @ghaerr,

> Can the memory locations of the GDT be overwritten, once the program exits? That is, does the system continue to work after running unreal16, when other applications overwrite that space?

> In protected modes, this is not the case, but I'm unfamiliar with unreal mode.

The system should continue to work, due to the "segment descriptor cache" (there is a [write-up](http://www.rcollins.org/ddj/Aug98/Aug98.html) by Robert Collins on this, but the behaviour is probably mentioned somewhere in Intel's own manuals too).  The CPU only consults the GDT when it is in protected mode.  After a switch to real mode (by writing to `%cr0`), the CPU will simply use whatever segment limits, access rights, etc. that were set previously in protected mode.

Thank you!

tkchia:
Incidentally, it is best not to confuse the "unreal mode" in this pull request, with the actual [`loadall`](http://www.rcollins.org/articles/loadall/tspec_a3_doc.html) instructions on the 286 and 386 processors.

Thank you!

georgp24:
Yes, the title is misleading.

jbruchon:
So...merge?

ghaerr:
Yes, thanks.

PR #433, opened 2020-03-08
[elks] [boot] allow booting from Minix/FAT FS on whole hard disk

tkchia:
ELKS can now automatically start up from a Minix or FAT filesystem installed on an entire (unpartitioned) hard disk.

Partly addresses https://github.com/jbruchon/elks/issues/348 & https://github.com/jbruchon/elks/pull/422#issuecomment-595895145 .

PR #438, opened 2020-03-09
[config] [boot] fix bootblocks build gremlins from config changes

tkchia:
This should address some build issues reported by @ghaerr at https://github.com/jbruchon/elks/pull/433#issuecomment-596240964 .

Changes:
  * `make defconfig` now correctly starts from a clean slate; previously it was reusing some of the settings in the existing `.config`.
  * `make all` now regenerates `include/autoconf.h` to synchronize it with the `.config` settings.
  * `bootblocks/Makefile` now rebuilds the boot blocks whenever `include/autoconf.h` is updated.
  * The generic bootloader code in `bootblocks/boot_sect.S` now correctly sets the track count for 1.2 MiB and (?) 1.68 MiB floppy disks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,9 @@ endif
 
 include $(TOPDIR)/Make.defs
 
-.PHONY: all clean kconfig defconfig config menuconfig
+.PHONY: all clean kconfig defconfig reconfig config menuconfig
 
-all: .config
+all: .config reconfig
 	$(MAKE) -C libc all
 	$(MAKE) -C libc DESTDIR='$(TOPDIR)/cross' install
 	$(MAKE) -C elks all
@@ -30,7 +30,7 @@ clean:
 	    echo ' * `make config` or `make menuconfig` to configure it.' ;\
 	    echo ;\
 	fi
-	
+
 elks/arch/i86/drivers/char/KeyMaps/config.in:
 	$(MAKE) -C elks/arch/i86/drivers/char/KeyMaps config.in
 
@@ -38,6 +38,11 @@ kconfig:
 	$(MAKE) -C config all
 
 defconfig:
+	$(RM) .config
+	@yes '' | ${MAKE} config
+
+# Update include/autoconf.h from .config.
+reconfig: .config
 	@yes '' | ${MAKE} config
 
 config: elks/arch/i86/drivers/char/KeyMaps/config.in kconfig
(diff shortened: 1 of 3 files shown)


ghaerr:
Hello @tkchia,

> * `make defconfig` now correctly starts from a clean slate; previously it was reusing some of the settings in the existing `.config`.

That's ok, I can see it removes .config before running a defconfig.

> * `make all` now regenerates `include/autoconf.h` to synchronize it with the `.config` settings.

This default `make` also recreates .config, which I'm not crazy about, just so that 'include/autoconf.h' is recreated, just so that the boot blocks are rebuilt. Isn't there a better way to do this, that doesn't touch .config nor autoconf.h. These files shouldn't be recreated with a 'make', IMO. How about just making `bootblocks/Makefile` dependent on .config and/or autoconf.h? Sometimes it is convenient to modify either .config or autoconf.h without them getting overwritten again. (For quick insertion of debug defines or image creation changes, for instance).

Another option would be to leave `make all` as is, but have `make` not recreate .config and autoconf.h; this would solve my problem.

> * `bootblocks/Makefile` now rebuilds the boot blocks whenever `include/autoconf.h` is updated.

That's ok, but the real bug I experienced was the lack of having CONFIG_IMG_FD1440 in boot_sect.S, wasn't it? (below). The bug persisted past a `make clean' and 'make config' beforehand.

> * The generic bootloader code in `bootblocks/boot_sect.S` now correctly sets the track count for 1.2 MiB and (?) 1.68 MiB floppy disks.

I have tested this and it works.

I tested this PR with a download patch file, since I'm waiting for @jbruchon to revert and/or force reset head for the whitespace commits so I don't touch files that I'm working on. I would still like to see this above issues possibly changed after that if you're willing, or I could do it.

jbruchon:
They were reverted earlier today.

tkchia:
Hello @ghaerr,

I have submitted another pull request to get `make all` not to regenerate `.config`.

I think it is still a good idea to regenerate `include/autoconf.h` if `.config` is newer --- this means that if you modify `.config` manually, then `include/autoconf.h` will be updated and get the right settings.

In any case, with the PR, `make all` now uses an ordinary target, not a phony target, to update `include/autoconf.h`.  This means that you can do a `touch include/autoconf.h` if you really want to force things.

Thank you!

PR #441, opened 2020-03-10
[minix] Fix Minix umount super block unchecked flag, add new debug mechanism

### LINKED ISSUE #437: 'shutdown' and 'reboot' should un-mount rootfs upon termination
As in the subject. The `shutdown` command does not seem to do anything (the system is still alive after `System halted` message) while rebooting the system with the `reboot` command results in un-checked rootfs filesystem warning message upon the subsequent system boot.

ghaerr:
This fixes #437. The Minix superblock is now force written on remounts, rather than previously trying to keep track of its status (which didn't work after `sync`). The `reboot` command does a sync, then mounts the root filesystem read-only, then sleeps for three seconds and reboots.

The `sleep` system call is not implemented (I recommend removing it entirely), and libc `sleep` is broken, so `sleep` is now (properly) written in reboot.c, for the time being. I will have a PR for libc `sleep`, although making any changes in `libc/system/sleep.c` are ignored (a problem in libc make?).

It took quite a bit to trace this down, and I've added an improved debug mechanism.
This mechanism allows for very interesting tracing, I will comment more on it after the next commit, which expands on it for the broken `signal` processing in ELKS.

@jbruchon, please commit this PR before my next PR (signal fix), as that may help with merge issues on my second PR.

### DIFF
--- a/libc/include/unistd.h
+++ b/libc/include/unistd.h
@@ -54,7 +54,7 @@ int unlink(const char *fname);
 
 pid_t fork(void);
 pid_t getpid(void);
-pid_t setsidvoid();
+pid_t setsid(void);
 
 uid_t getuid (void);
 
--- a/libc/include/unistd.h
+++ b/libc/include/unistd.h
@@ -52,9 +52,9 @@ int isatty (int fd);
 off_t lseek (int fildes, off_t offset, int whence);
 int unlink(const char *fname);
 
-pid_t fork ();
-pid_t getpid ();
-pid_t setsid ();
+pid_t fork(void);
+pid_t getpid(void);
+pid_t setsidvoid();
 
 uid_t getuid (void);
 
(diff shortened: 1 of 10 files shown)


libc/include/unistd.h
```diff
@@ -52,9 +52,9 @@ int isatty (int fd);
 off_t lseek (int fildes, off_t offset, int whence);
 int unlink(const char *fname);
 
-pid_t fork ();
-pid_t getpid ();
-pid_t setsid ();
+pid_t fork(void);
+pid_t getpid(void);
+pid_t setsidvoid();
```
mfld-fr: Not a typo here ?

mfld-fr (changes requested):

libc/include/unistd.h
ghaerr: Sure is, thanks for finding this. Because of the combined branchs (I created sup first, then sig as a sub-branch, but GitHub is combing the second one), I'd like to fix this in another commit, if that's  ok.

PR #442, opened 2020-03-10
[signal] Fix userland signals

### LINKED ISSUE #387: ^C hangs ELKS
Since grep and bc etc. are now working, it's time for some more command testing. I'm collecting those for a summary, but here is a potentially more general issue:
More + ^C hangs ELKS. Terminal driver issue? (more has its own problems, that's for later).

-Mellvik

ghaerr:
This PR fixes #387.

Getting ^C and ^Z handling working involved fixing one big thing and lots of little things: 
First, the entire userland signal numbers were incorrect, due to libc having it own list of signals requiring 32-bit signal masks, and the kernel's version, which uses 16-bit masks. Since the signal numbers themselves were all different, an entirely different signal was being generated on any signal event. Libc now includes the kernel signal.h header to rectify this.

Various important utilities, include getty, login, ash and sash, didn't handle ^Z (SIGTSTP) properly and had to be modified. I went through all applications and found that there are still a couple that won't work properly, as SIGTTIN and SIGUSR2 aren't implemented.

This fix normally works well, but there is still a ELKS startup race problem having to do with /bin/init and `setsid`, which sets the controlling terminal/process group. When this race condition manifests itself, one will see that ^C and ^Z have no effect, since there is no controlling terminal. This seems to occur 20% of the time, and is related to the process startup and termination when running the /etc/rc.s/rc.sys script. More details upon request. I added a TTY field to `ps` to see whether a process has a controlling terminal.

For the files that had to be modified, I've also started cleanups, which can be seen in ntty.c, for example. Also trying to pare down the warnings for compilations of each file touched as well.

This PR also expands the debug mechanism (see linuxmt/debug.h) and allows tracing of signals, which is quite interesting. I will give a complete writeup once I change the FAT filesystem debug mechanism, which is very similar, to the new mechanism, which allows for fine granular control of seeing what's happening in the kernel. One may well be surprised with what is seen!

@jbruchon, please commit this PR after the #441, just in case of merge issues thanks!

### DIFF
--- a/elks/arch/i86/drivers/char/ntty.c
+++ b/elks/arch/i86/drivers/char/ntty.c
@@ -23,7 +23,7 @@
 #include <linuxmt/chqueue.h>
 #include <linuxmt/ntty.h>
 #include <linuxmt/init.h>
-#include <linuxmt/major.h>
+#include <linuxmt/debug.h>
 
 /*
  * XXX plac: setting default to B9600 instead of B1200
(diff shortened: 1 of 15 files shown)


ghaerr:
@jbruchon, for some reason, this PR wants to add the previous [minix] fix as well, even though I created a separate branch for it. It is also showing the combined changes. I can resubmit this after the first PR is committed, unless it is easy to separate out the [minix] commit above.

I'm sorry, I just don't know git well enough, lots more to learn!

ghaerr:
@jbruchon: Alternatively, it looks like committing only this PR, and closing the #441 would also solve the problem. Let me know what you'd like me to do.

jbruchon:
You have contributor access to the repo now, so you're able to merge PRs and manage issues without waiting for me.

ghaerr:
@jbruchon, ok, thanks. I'll see what other comments there are, then fix my own issues, and commit :)

PR #443, opened 2020-03-10
[config] `make all` no longer regenerates .config

tkchia:
`include/autoconf.h` may still be regenerated, but only if it is out of date with respect to `.config`.

This addresses some issues brought up by @ghaerr (https://github.com/jbruchon/elks/pull/438#issuecomment-596903198).


ghaerr:
Thanks @tkchia!

Issue #440, opened 2020-03-09
64K DMA boundary


mfld-fr:
@tkchia : I think I got at least the explanation for the "*do not cross the 64K boundary*".

I was playing with `setup` to move the kernel data segment in memory, when I got the error:
```
hd: error: AX=0x09
BIOSHD:I/O error
dev 380, sector 6
```

In IBM-PC manual, error 09H is "Attempt to DMA across a 64k boundary".

Then I found this page about the DMA controller: https://en.wikipedia.org/wiki/Intel_8237#IBM_PC_use

Do you understand the same thing as me ?

mfld-fr:
More info: in my test, the kernel data segment starts at D63:0h, and `system.map` gives:
```
 .bss           0x0000000000001baa     0x359e fs/fs.a(buffer.o)
 .bss           0x0000000000005148        0x0 fs/fs.a(super.o)
```
So yes, the array of L1 buffers is crossing the 64K boundary within the linear range from F1DAh to 12778h.

I added a trace just before the error occurs, and it fails when filling a 1K buffer at D63:27C7h, so less than 1K before the 64K boundary.

If I force the kernel data segment to starts at 1000:0h, no more error.

ghaerr:
> If I force the kernel data segment to starts at 1000:0h, no more error.

Out of curiosity, could the L1 buffer declarations be taken out of C code and instead be placed by startup code, say, above the kernel stack at a 64k boundary (and KSP reset below it) , so that instead of memory being wasted, that memory goes into the future heap/stack area?

mfld-fr:
@ghaerr : sorry, I did not understand your question ?

tkchia:
Hello @mfld-fr,

> Do you understand the same thing as me ?

Yes, this is what I was referring to.  Thank you!

mfld-fr:
@tkchia : OK, so it means the current memory layout is unsafe, and works by chance because the data segment is relocated as low as possible (actually starts at 80:0h), so it would not cross the first 64K boundary as long as its size is less or equal to F800h (= 62K).

mfld-fr:
Issue was already caught in #107 (`BUFSEG` conditionnal).

Other references:
- https://github.com/jbruchon/elks/pull/422#discussion_r389224706
- https://github.com/jbruchon/elks/pull/429#issue-385144653
- https://github.com/jbruchon/elks/commit/9fab4d52ff285735935838e252b0ee5617c2d86d
[...]

PR #445, opened 2020-03-10
[kernel] Swap back code and data segment

mfld-fr:
Prepare the data segment extent to maximum (64K) and local heap introduction.


ghaerr:
Hello @mfld-fr,

I don't quite yet understand what this is doing, can you explain it a bit more? I referenced this in a comment on #440 and don't know exactly which problem this is solving.

mfld-fr:
@ghaerr : all is already explained in #440 

@tkchia : can you review please ?

tkchia:
Hello @mfld-fr,

Thanks --- it looks OK (so far) to me.

Well, except for one thing: `do_bioshd_request()` does not seem to limit the size of each disk I/O request to the size of the DMA buffer (`DMASEG:0`)?

Perhaps there should be a `DMASEGSZ` constant or some such to give the size of this DMA buffer, and the `DMA_OVR` code should limit each disk I/O request to this number of bytes.

Thank you!

mfld-fr:
@tkchia : Do you mean a more explicit limitation than the current one in lines 706 and 729 ?

tkchia:
Hello @mfld-fr,

> @tkchia : Do you mean a more explicit limitation than the current one in lines 706 and 729 ?

Yes.  Thank you!

ghaerr:
Hello @mfld-fr,

Do you want this committed now, or are you making a change per @tkchia's review on lines 706/729?

mfld-fr:
@ghaerr : I think @tkchia is right when asking a more 'safe' limit to not overflow DMASEG.

mfld-fr:
@tkchia : added the 'safer' limit.

ghaerr:
Hello @mfld-fr,

I think I understand this PR now, but am trying to keep up with you and @tkchia on the DMA boundary issue and this PR: It seems that we don't do BIOS disk reads into L1 or L2 buffers yet, that instead, all doshd.c I/O uses DMA and goes into the DMASEG buffer, which this PR is now relocating and enhancing? Then, that DMA buffer contents are copied using `fmemcpy` to another buffer, is that correct?

If so, then the suggestion of moving the L1 buffers to a 1k boundary is moot, since no I/O occurs directly into them, right?

mfld-fr:
@ghaerr : correct, temporary use of DMASEG (until I/O directly in L2 buffers) relaxes the alignment constraint on the L1 buffers, so they are now free to be allocated anywhere in the kernel data segment with the upcoming heap allocator.

PR #450, opened 2020-03-11
[setboot] Update setboot to set numtracks for HD testing

ghaerr:
This updates `setboot` to allow an optional third parameter in the -B option for setting the number of tracks to write in the ELKS PB. Discussed in #433.

The package manager is updated for testing HD images and now successfully creates a 32MB test image 'hd32-minix.bin' along with the fd*.bin images. This is for testing purposes for others to easily test (currently flat) HD images along with floppy images on their equipment. 

Note: Currently unable to create a FAT HD image for testing, as `mformat` won't allow a size larger than 2.88M. Does anyone have any ideas for other tools for large FAT filesystem creation?

Hello @tkchia, on another note, while looking at bootblocks/boot_sect.S for this, I noticed what I think ultimately caused the bootblocks build error to go unnoticed (until you fixed it). That is, the following code is used as a last resort when the number of tracks, heads or sectors is unknown:
```
#   else
    .warning "Unknown number of disk tracks!"
    .word 0
#   endif
```
I'm thinking that perhaps the code should throw up an error that will stop the build, rather than issuing a warning and continuing. I'm sure that the warning went unnoticed during the build, and inserted 0 tracks for the CONFIG_IMG_FD1440 case that happened to be left out.

### DIFF
--- a/elks/tools/setboot/setboot.c
+++ b/elks/tools/setboot/setboot.c
@@ -3,7 +3,7 @@
  *
  * 7 Feb 2020 Greg Haerr <greg@censoft.com>
  *
- * Usage: setboot <image> [-F] [-B<sectors>,<heads>] [<input_boot_sector>]
+ * Usage: setboot <image> [-F] [-B<sectors>,<heads>[,<tracks>]] [<input_boot_sector>]
  *
  *	setboot writes image after optionally reading an input boot sector and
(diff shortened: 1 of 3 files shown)


tkchia:
Hello @ghaerr,

My PR https://github.com/jbruchon/elks/pull/451 should fix the problem you mentioned.  Thanks!

PR #455, opened 2020-03-11
[libc] [kernel] [elkscmd] start tidying up header files

tkchia:
Changes:

  * `<linuxmt/arpa/inet.h>` has been moved from the ELKS kernel subtree to the `libc` subtree.  It is used only in userland code.
  * `<types.h>` and `<asm/types.h>` have been removed.  Their functionality has been added to `<arch/types.h>`, `<linuxmt/types.h>`, and `<linuxmt/wait.h>`.
  * If `__KERNEL__` is undefined, `<linuxmt/fs.h>` now only defines macros needed by userland programs (e.g. `fsck`).
  * Other files have been updated correspondingly.


ghaerr:
Hello @tkchia,

Looks good, thanks! It's good to see libc and kernel includes get cleaned up.

I have another issue about libc I've been meaning to ask you about: I recently fixed `reboot` which called `sleep(3);` and found that there was no delay. Further inspection showed that `sleep` was set as a system call connected to no_sys (unimplemented). I fixed reboot, but had to include `sleep` in reboot.c, which resulted in the inability to use <unistd.h>, since sleep is defined there.

I had tried fixing sleep in 'libc/system/sleep.c', but found that when I make any changes at all to that file, it doesn't even get compiled! I don't know multilib well enough to fix this. I would like to use the *second* implementation in 'libc/system/sleep.c', remove sleep from 'reboot.c', and remove the `sleep` kernel entry point (which is a no-op anyways). 

What do you think?

tkchia:
Hello @ghaerr,

I took a look, and it seems this is not really due to the multilib mechanism, but the (admittedly somewhat arcane) mechanism for compiling syscall stubs and alternate implementations of system routines.  I will see how I can solve this `sleep` problem.

Thank you!

PR #456, opened 2020-03-11
[minix] Force superblock write on RW mount

ghaerr:
Fixes remaining problem in #437.


ghaerr:
Hello @mfld-fr,

I notice that the PR's that I have been committing for myself and others are producing two commits in the repo, something that you said earlier should be avoided. Instead of using 'Merge Pull Request', should the 'Rebase and Merge' option be used? The 'Squash and Merge' option also produces a separate merge commit.

mfld-fr:
@ghaerr : no, I said earlier we should not mix the commits with effective changes and others with cosmetics like reindentation or respacing. IMO having a merge tree is better for history and merge back to forks than a linear and constantly rebased master.

Issue #410, opened 2020-03-04
ELKS applications reorganization RFC


ghaerr:
I have analyzed the current layout and state of the ELKS applications directory elkscmd/, and have a number of ideas on how to improve speed, disk usage, and the overall usability of ELKS at the command line. This would include solving the problem of multiple sources of commands that don't quite do the same thing, as well as providing two config-selected shells, one oriented towards speed and small kernel buffer usage, and the other for full BASH shell scripting, capable of working together and installed separately or together, depending on user requirements.

I can work on this and will submit a detailed plan and set of PRs, but wanted to seek comments first from those that are interested. Following is a detailed illumination of the current state of elkscmd/, along with a suggestion for moving to core_utils, which solves the source duplication and shell builtin problems, but also allows for very small binaries and keeping busyelks commands identical to standalone and shell versions.

The `elkscmd/' directory definitely needs some reorganization and cleanup. I would like comments and some approval before going forward.
```
Config Option           Directory       Comments
----------------------  --------------  ------------------------------------------------------------
# shells
CONFIG_APP_ASH          ash/            Default shell, installs as /bin/sh. Early BASH compatible.
                                        Std builtins: cd, chdir, '.', source, echo, eval, exec, exit,
                                            export, readonly, getops, hash, jobid, jobs, read,
                                            set, setvar, shift, trap, ':', true, false, umask, unset, wait
                                        Scripting builtins: break, continue, return
                                        No non-std builtins.

CONFIG_APP_SASH         sash/           Not selectable when ash selected. Installs as /bin/[sh,sash].
                                        This is not a BASH-compatible shell and does not perform std scripting.
                                        It has some standard and large number of useful builtins.
                                        Std builtins: alias, unalias, cd, exec, exit, source, setenv,
                                            prompt, umask
                                        Builtins:
                                            chown, chgrp, cp, echo, ln, ls, mkdir, mkdir, mknod, printenv,
                                            pwd, sync, rm, rmdir, sync, mv, mount, umount,
                                            dd, ed, grep, more, kill, tar, touch
                                        Other builtins: help, quit

# New config option
CONFIG_APP_COREUTILS    core_utils/     Core utilities, Installed in /bin. Sources moved from other directories
                                        Already included in sash (but sources to be merged):
                                            chown, chgrp, cp, echo, ln, ls, mkdir, mkdir, mknod, printenv,
                                            pwd, sync, rm, rmdir, sync, mv, mount, umount,
[...]

PR #457, opened 2020-03-11
'grep' from 'minix1' only.

marcin-laszewski:


ghaerr (commented):
I noticed this last night myself. 

However, I am thinking the decision to use minix1/grep.c may have to be reversed, and instead use the file_utils/grep, since part of the elkscmd/ reorganization is to have the exact same user command behavior   with `sash`  internal commands and `sh` external /bin commands (and this should be the same for busyelks too eventually, see comment in #458). And then there's the issue of 'fgrep' also being around. If we're going to use minix1/grep, with its possibly superior dash options (but untested otherwise), then it should be integrated into sash/cmd_grep.c as well. Do you have a specific preference for 'grep' functionality at this time, or is this just a quick effort to get rid of one of them?

Also, the `rm -f grep` is not needed; I'm guessing you added that after removing 'grep' from OUTS, which then affects the `make clean` operation.

mfld-fr:
Mmm... IMO `sash` goal is not to have exactly the same behavior as the standalone commands, but rather to be a MINIMAL shell for 'rescue' or 'init' (see https://en.wikipedia.org/wiki/Stand-alone_shell).

I am using it in very tight memory condition on my SBC with both kernel and ROMFS in a few KB Flash, and I don't agree to have it bloated with unnecessary additions outside that goal.

Moreover, we are on version 1.0 (according to the `README`). I did a quick compare with the current 3.8 version, and it could be upgraded with minimal effort. That upgrade would be more difficult with more changes, so let us keep as it is.

For the code duplication between `busyelks` and standalone commands, I already suggested to share the very similar code with symbolic links on sources, and just change the entry points.

I agree the contributors that tell `busyelks` is too big for slow floppies, but the goal of `busyelks` is NOT 'I/O efficiency'. It is 'storage efficiency'. So thank you @marcin-laszewski for your work, it really helps in my use case with little storage but high bandwidth within memory.

For that particular PR, I agree @ghaerr : you should not impact the standalone commands that are under heavy work, but just link sources from that standalone commands to assemble `busyelks`, and only touch them to insert the conditionals for the entry points as said above.

ghaerr:
Hello @mfld-fr, thank you for your comments.

> Mmm... IMO `sash` goal is not to have exactly the same behavior as the standalone commands, but rather to be a MINIMAL shell for 'rescue' or 'init' (see https://en.wikipedia.org/wiki/Stand-alone_shell).

Yes, agreed. I have recently made a few small enhancements to `sash` that allow its use as a more normal shell for very small systems but in both cases, it should remain small.

Almost all of the standalone commands that are also in `sash` are, in fact, originally copied from sash, and modified slightly to compile standalone. One of the few standalone commands that has been enhanced from that point is `ls`, with possibly a very few others.

So - clarification: When I say 'exactly the same behaviour' - I mean for *just* those commands that have been enhanced (e.g. `ls -F`), those same enhancements would be shared with `sash`, The source code for those commands that are present in both should be shared, since in almost every case, they are quite close.


> I am using it in very tight memory condition on my SBC with both kernel and ROMFS in a few KB Flash, and I don't agree to have it bloated with unnecessary additions outside that goal.

Agreed. The ELKS command reorganization #410's goal is to reorganize the structure within elkscmd/ and tidy things up, particularly some directory structures and trying to share source. It is not to add tons of commands and bloat `sash`. 


> Moreover, we are on version 1.0 (according to the `README`). I did a quick compare with the current 3.8 version, and it could be upgraded with minimal effort. That upgrade would be more difficult with more changes, so let us keep as it is.

Agreed.


> For the code duplication between `busyelks` and standalone commands, I already suggested to share the very similar code with symbolic links on sources, and just change the entry points.

Unfortunately, it's not quite that simple, but the goal is sharing. I didn't want to have to come up with a full-blown 'source-applet' structure, and think we may not need to, but there are considerations like not having `exit` called from anywhere other than 'main', etc.

We are not yet at the point where a symbolic link could be used, more comparison and work needs to be done, again, which is the goal of #410.


> I agree the contributors that tell `busyelks` is too big for slow floppies, but the goal of `busyelks` is NOT 'I/O efficiency'. It is 'storage efficiency'.
 So thank you @marcin-laszewski for your work, it really helps in my use case with little storage but high I/O within memory.

I would like to see busyelks continue to be enhanced as well. Also, @marcin-laszewski has performed a number of required seperations of duplicated routines which is very useful for #410's goals.


> For that particular PR, I agree @ghaerr : you should not impact the standalone commands that are under heavy work, but just link sources from that standalone commands to assemble `busyelks`, and only touch them to insert the conditionals for the entry points as said above.

Agreed for @marcin-laszewski , but that also is not so simple. Busyelks requires a few other things, like removing duplicate routines called between programs, from the source. So both busyelks and elkscmd/ are works under construction. The conversation on how to fix that should probably be under #410.

What I am suggesting is to try to keep the elkscmd/  Makefile changes to a reasonable minimum until a reasonable source integration plan can be developed.

ghaerr:
@marcin-laszewski: Lets place this PR and #458 on hold for a few days while the initial changeover to moving installation out of elkscmd Makefiles get implemented. I am sorry for the confusion, and thank you for trying to get this to work. Those changes should get you out of having to deal with installation Makefiles and able to concentrate on busyelks code instead. I hope to be able to use your 'busyelks.fs' or something very similar for the symlink creations.

marcin-laszewski:
ok

Issue #402, opened 2020-02-29
Can't get busyELKS options to build proper image


ghaerr:
Hello @marcin-laszewski and @mfld-fr,

When running `make config` and selecting busyELKS in two different scenarios, both produce failures in the resulting image. I am starting from scratch by deleting .config.

When all options are set to default, except answering Y to this question:
```
busyelks (CONFIG_APP_BUSYELKS) [N/y/?] (NEW) 
```
an fd1440.bin image is produced, but the symlinked set of busyELKS commands are not included in it. That is, in elkscmd/busyELKS/:

 ```
Gregs-MBP2:busyelks greg$ ls -l busyelks*
 -rw-r--r--   1 greg  staff  28164 Feb 27 20:39 busyelks.tar.xz
 -rwxr-xr-x   1 greg  staff  46453 Feb 27 20:36 busyelks
 -rw-r--r--   1 greg  staff      0 Feb 27 20:36 busyelks.map
 Gregs-MBP2:busyelks greg$ tar Jtf busyelks.tar.xz 
 ./
 ./bin/
 ./bin/busyelks
 Gregs-MBP2:busyelks greg$ 
 ```

On the other hand, when I select busyELKS as above in config, and also answer 'b' to each of the other busyELKS application categories as follows:
```
busyelks (CONFIG_APP_BUSYELKS) [N/y/?] (NEW) y
dbgutils (CONFIG_APP_DBG_UTILS) [Y/n/?] 
diskutils (CONFIG_APP_DISK_UTILS) [Y/b/n/?] b
elvis (CONFIG_APP_ELVIS) [Y/n/?] 
fileutils (CONFIG_APP_FILE_UTILS) [Y/b/n/?] b
levee (CONFIG_APP_LEVEE) [Y/n/?] 
m4 (CONFIG_APP_M4) [Y/n/?] 
minix1 (CONFIG_APP_MINIX1) [Y/b/n/?] b
minix2 (CONFIG_APP_MINIX2) [Y/n/?] 
minix3 (CONFIG_APP_MINIX3) [Y/b/n/?] b
miscutils (CONFIG_APP_MISC_UTILS) [Y/b/n/?] b
```
then the fd1440.bin image produced does not contain commands required to login, as /bin/login, /bin/cat, etc are missing.

Who is testing this? I'm not getting either option to work.

I repeat below my questions brought up in PR #389, which were unanswered.

[...]

PR #458, opened 2020-03-11
[elkscmd] Makefile bugfix

marcin-laszewski:


ghaerr:
Hi @marcin-laszewski,

This is starting to get a little ugly, and it's only going to get worse, as you port more
utilities to busyelks. Is this what is required to fix the OSX and Gentoo problems?

I don't like the multiple nested if statements in every Makefile for every command
category that's going to be used in busyelks, as well as the following concerns:

There has been more discussion as to what the fate of busyelks should be,
have you been following that? I also remain concerned that the sources you've
copied into elkscmd/busyelks/cmd/ are getting out of date, and, as you've seen,
there are already multiple copies of `ls`, `grep`, and others and now your third
copies. This is all heading towards a maintenance nightmare.

How about the an alternate approach: lets try, at least for the time being,
on containing busyelks so as not to interfere as much with the improvements,
Makefiles, and installation directories used by the existing elkscmd/ commands.

That is: remove all checking for 'CONFIG_APP_\*' from all elkscmd/*/Makefiles, but
keep the OUTS= and OUTS_BE +=.  Then, add another 'install' line just below the
regular one, which installs OUTS_BE to $(DESTDIR)/ebin. In the top-level elkscmd/Makefile,
check for CONFIG_APP_BUSYELKS and if NOT, them 'rm -rf $(DESTDIR)/ebin'.

Your elkscmd/busyelks/Makefile can then prepare links into $(DESTDIR/ebin, and 
PATH=/ebin:$(PATH) can be added to /root/.profile. This puts busyelks completely
within /ebin on the target disk, lets you develop at will without having to worry about
the elkscmd/ cleanup that is occurring, and others being able to use and test busyelks
should they be interested. After both busyelks and the elkscmd/ cleanup are done, 
we can worry about a more proper integration.

marcin-laszewski:
> (...) it's only going to get worse, as you port more utilities to busyelks.

Selecting [ ] or [*] in menuconfig will fix this problem.

> Is this what is required to fix the OSX and Gentoo problems?

I don' know - I have no OSX.



> I don't like the multiple nested if statements in every Makefile for every command (...)

It is OK. Look at, for exmaple, makefiles of glibc.

> If an old command work correctly it is ok for me.
> (...) as you've seen, there are already multiple copies of ls, grep, (...)

Good news. We can choose.

> (...) now your third copies.

No. This a is renovation of already existing command 'busyelks'.

> (...) /ebin (...)

We have removed /usr/bin. And now, are you going to create /ebin ?


There is a nice file 'busyelks.fs'. It contains all symlinks to busyelks. Change it as you wish.

mfld-fr:
@ghaerr : please NEVER insert your response in another contributor's post. One cannot easily determine who the sentence is coming from, until realizing the post was edited by another person. I did that earlier to cleanup 'copy of original post' when one replied by mail, but I never dare to alter one's text !

mfld-fr:
@ghaerr : so please restore original post from @marcin-laszewski and post your response after.

elkscmd/disk_utils/Makefile
```diff
@@ -13,9 +13,11 @@ include $(BASEDIR)/Make.rules
 OUTS = ramdisk mkfs fsck partype
 OUTS_BE = fdisk
 
-PROGS = $(OUTS)
-ifeq "$(CONFIG_APP_DISK_UTILS)" "y"
+ifdef CONFIG_APP_DISK_UTILS
+	PROGS = $(OUTS)
+ ifeq "$(CONFIG_APP_DISK_UTILS)" "y"
 	PROGS += $(OUTS_BE)
+ endif
```
mfld-fr, line 20: In a previous discussion with @ghaerr, we agreed to build ALL the code in the repository, whatever is installed in the `target` and put in the `image`, so that we could easily detect a broken build dependency. Could you please move the conditionals to the `install` target rather than the `all` ?

ghaerr:
> @ghaerr : please NEVER insert your response in another contributor's post. One cannot easily determine who the sentence is coming from, until realizing the post was edited by another person. I did that earlier to cleanup 'copy of original post' when one replied by mail, but I never dare to alter one's text !

@mfld-fr : I completely agree. I am unable to determine how this happened, other than by mistake selecting Edit rather than Quote Reply. It was never intended to rewrite anyone's text. I am guessing that 'Collaborator' inserts Edit menu item when it was not available before, I will be more careful!

mfld-fr (changes requested):

ghaerr:
> (...) it's only going to get worse, as you port more utilities to busyelks.
> Selecting [ ] or [*] in menuconfig will fix this problem.

My concern is with the elkscmd/ Makefiles getting complicated as busyelks is expanded
 with compiling \*.c files already present in busyelks/cmd that compete with
 their duplicate versions in elkscmd/\*/\*.c. Selecting them out of building
 doesn't solve the problem.
 
> Is this what is required to fix the OSX and Gentoo problems?
> I don' know - I have no OSX.
  
I am referring to open issue #402, and specifically @pawosm-arm's preparing
 of a script file for you, does this fix that problem? He is running Linux, and it
 doesn't work on his system either.
  
Also, the '-r' option needs to be removed for sure on OSX, since apparently
 the busyelks build doesn't terminate with an error and still doesn't build a symlink.
 I'm unsure as to the status of that fix (also from #402):
  
> I think the first bug is the ‘-r’ option on ln. it would be easier to debug if you used separate commands in shell or make rather than using &&. Then command echoing could be turned on, and the script would stop with errors.
  
> I don't like the multiple nested if statements in every Makefile for every command (...)
> It is OK. Look at, for exmaple, makefiles of glibc.
  
No one has been maintaining elkscmd/ and it is already too messy, so I don't agree.
  
> If an old command work correctly it is ok for me.
> (...) as you've seen, there are already multiple copies of ls, grep, (...)
> Good news. We can choose.
  
Yes, but users are asking for enhancements in commands and there's three
different source versions for many of them.
  
> (...) now your third copies.
> No. This a is renovation of already existing command 'busyelks'.
 
I meant that the original copy of files from elkscmd/ into busyelks (how or
whomever that happened), ended up such that there are three different
versions of almost every command in file_utils. One in file_utils, the other in sash/,
and the third in busyelks/cmd/.
 
> (...) /ebin (...)
> We have removed /usr/bin. And now, are you going to create /ebin ?
  
No, not really! I just quickly suggested that to try to encapsulate busyelks,
since it doesn't yet build on OSX (nor Gentoo), and no one seems to know the
reason.
  
> There is a nice file 'busyelks.fs'. It contains all symlinks to busyelks. Change it as you wish.
 
and the 'ln -r' fix doesn't fully fix the problem, as I tried removing it and still failed to build links.
 This was the reason I was asking whether this PR fixes #402 (in some other way I
 don't realize).
 
 
Ultimately, I don't have a big problem with this PR, it's just that it seems to go the wrong direction with #410, which users have asked for. I suppose it's a question of priorities - should ELKS commands in general be fixed up and reorganized, or should busyelks be integrated? I'm trying to accommodate both, but it seems they are going in opposite directions, especially since busyelks isn't fully working yet.

ghaerr:
@mfld-fr : The thread has been restored. Thank you for catching this.

mfld-fr:
@ghaerr : thanks for the restore, easier to follow the discussion now.

elkscmd/disk_utils/Makefile
ghaerr: @mfld-fr: Agreed. However, even without this PR, the goal of building all commands is not met. When the 'B' option is selected in config for any commands categories, then only those commands not duplicated in buyselks are built, only the ones in elkscmd/. This is part of my problem with the current integration of busyelks and IMO should be handled differently. I would like to see all elkscmd/ commands built, *and* busyelks built at all times, and *only* the 'install' portions of elkscmd/ and busyelks/ operate differently, depending on the config selection.

[EDIT: ... only those commands NOT duplicated in busyelks are built... above]

ghaerr: After looking at this more closely, I think @mfld-fr's suggestion of moving all conditionals to the `install` target will allow building of all commands at all times, and the 'B' option will work as well, thus solving all problems. If the busyelks/ `install` target removes the non-portable '-r' option from `ln`, busyelks install should work on all systems as well.

ghaerr:
Hello @marcin-laszewski,

I have submitted a PR (#466) that addresses this issue by removing all CONFIG_APP_xxx installation config options to just elkscmd/Makefile, and fixes the link creation for OSX in busyelks. I also modified busyelks/Makfile so that it only creates links for those sub-commands that are actually being compiled in. This should allow you to easily add more commands into busyelks and only having to Modify elkscmd/Applications (changing, say a ':fileutil' to ':be-fileutil') for each command. If a new commands category (like sys_utils) needs to be added to busyelks, then a slight adjustment to elkscmd/Makefile to process ':be-sysutil' is all that would be needed.

I would like your comments, and if OK, this issue can be closed and that PR accepted.

marcin-laszewski:
I generally use "make" only and I avoid "crearors", "generators" and additonal config files: to make a program use the make program.

ghaerr:
Hello @marcin-laszewski,

Agree compilation systems to make a program are best built using make, rather than other methods. This problem isn't for compilation, but the issue of determining which files should be installed for an application, based on config settings.

The core of the installation problem is that the current config installation mechanism uses make to copy either all, or none of the files in a given category, say diskutils or sysutils, to  target image. Each category is actually a source code directory of a group of applications, not a meaningful user-selectable category as more application source directories are added. When only a portion of files from a source directory should be copied, the current design requires individual Makefile changes for each package (like busyelks) that wants to install portions of that category.

The redesign in #466, started but not finished, allows for more useful 'virtual' categories to be selected, of which the files selected are not limited to having to be in one source directory. This is exactly the problem for busyelks installation, it needs to not install certain commands from source directories where it instead creates links to busyelks. I don't think it best to have to maintain complicated Makefiles for every  application, like busyelks, that requires finer-grained application installation than only that of "all or none" for source directories. The same problem exists for creating multiple disk image sizes, where the applications desired are more fine-grained than all or none of source directory boundaries.

ghaerr:
Closing, handled as discussed in #466.

PR #461, opened 2020-03-12
[libc] use alarm & signal, not sleep syscall, to realize sleep(.)

tkchia:
This commit addresses an issue mentioned by @ghaerr at https://github.com/jbruchon/elks/pull/455#issuecomment-597746144 .

This disables the `sleep(.)` syscall routine which invokes syscall number 35, which happens to be unimplemented --- and enables the alternate implementation of `sleep(.)` (in `libc/system/sleep.c`) using the `alarm(.)` and `signal(, )` syscalls.

### DIFF
--- a/elks/arch/i86/kernel/syscall.dat
+++ b/elks/arch/i86/kernel/syscall.dat
@@ -53,7 +53,7 @@ chroot		+31	1
 vfork		+32	0
 access		+33	2	 
 nice		34	1
-sleep		35	1	- use alarm & signal instead
+sleep		35	1	- use alarm & signal, or select, instead
 sync		+36	0	 
 kill		+37	2	 
 rename		+38	2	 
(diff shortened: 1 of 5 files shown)


ghaerr:
Hello @tkchia,

Thank you for removing the sleep system call and enabling the compilation of sleep.c. I can now see how that is done.

I am thinking that perhaps the better implementation of `sleep`, now that `select` is implemented (and mentioned in sleep.c) is the second version. This would be turned on by changing the '#if \__ELKS__' to '#if 0' in sleep.c. This is also the tested implementation that was used in `reboot`. 

What do you think of using that routine instead, as it doesn't suffer from potential interruption of other `signal` processing that an application might be using?

ghaerr:
Thank you for your help on this @tkchia!

PR #466, opened 2020-03-14
[cmds] Reorganize ELKS command installation

ghaerr:
First big step forward on reorganizing ELKS command building and installation Makefiles.

All (working) source directories in elkscmd/ are always compiled during 'make'. CONFIG_APP_XXX installation information is no longer present in any build Makefiles. Instead, (almost) all application installation is handled in 'elkscmd/Makefile', where config-specified options are used to install all commands into $(DESTDIR) using the new table-driven specification file 'elkscmd/Applications'. Symlink creation is currently handled in 'busyelks/Makefile', but could be moved out later.

This PR also fixes all busyelks issues and properly creates portable symlinks in /bin to /bin/busyelks for any command sets specified with the busyelks config 'B' options. It is now quite simple to add or subtract commands from busyelks or other directories, without having to modify any Makefiles. Only the 'Applications' specification file is edited, everything else is automatic.

Everything works well, but this is tested only on OSX. There is more to do with possibly allowing more specificity to include/exclude individual programs within command sets using config, as well as automatic selection of smaller command sets when smaller (360k) disk images are created.

This is the first step in allowing the 'package manager' to be moved out of image building, but those changes will come in a subsequent PR, after more discussion about the next steps.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -1,0 +1,155 @@
+# ELKSCMD Application Installation Table
+#	3/13/20 Greg Haerr
+#
+# This file is used to specify when each application in elkscmd/ should
+# be installed to $(DESTDIR).
+#
+# The file format is the application build filename in the first field,
+# followed by a tab and optional '::' installation filename (default is /bin),
+# followed by tabs and any number of tags specifying which configuration
+# sets each application belongs to.
+# A tag is an alphanumeric name preceded by a colon.
+#
+# Multiple tags are combined with the | character in elkscmd/Makefile
+# to specify the set of files to be installed.
+# The applications are installed from their build directory to $(DESTDIR).
+#
+# Following are the current ELKS application packages and description:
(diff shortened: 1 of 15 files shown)


ghaerr:
Hello @mfld-fr, 

Do you know which shell (bash, dash, zsh?) is running on CI? I'm building a linux VM to get this check failure debugged but am worried about shell compatibilities, and I'm no shell scripting expert. I am thinking the build is dying on the line 155 in elkscmd/Makefile but can't see which line for sure:
```
		if [ "$${dir:0:2}" == "::" ]; \
```

mfld-fr:
@ghaerr : all is explained here: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

As you can read, you could also setup another workflow to build under macOS... so that it would ensure build non regression on both Ubuntu and macOS.

PR #467, opened 2020-03-14
[kernel] Local heap introduction

mfld-fr:
Empty `arch_boot` cleanup, more Linux-like atomics and local heap introduction.


ghaerr:
Nice work. 

I'm interested in what kinds of test programs might be used to test dynamic memory management fragmentation and when to free, especially for those items likely to stay across multiple process invocations. I'm presuming there will also likely have to be upper limits for larger allocations like L1 buffers, based on total heap capacity.

mfld-fr:
@ghaerr : thanks, but that is only a small piece of the staging code I am trying to maturate on my side.

I cannot respond to your question about the large objects yet, because I have still many issues in that staging code related to the overall memory management, in order to keep the system working smoothly even in low memory condition. Stay tuned...

Issue #469, opened 2020-03-14
[kernel] Improve the segment allocator


mfld-fr:
Preview of upcoming changes related to segment allocator in commit 420adfc2:
- segment descriptors in local heap
- segment allocator code simplified (no more peek / poke)
- `seglist` code removal

ghaerr:
Hello @mfld-fr,

Thank you for offering a preview. I have studied the code quite a bit and have the following observations and thoughts:

__Summary__
While it is exciting to bring local heap management into ELKS, using local heap to hold descriptors for global memory is not the place for it. It is mentioned in the code that this is done for speed, and also for "when 286 protected mode returns someday". I will explain later in more detail why the former can be made even quicker with a design change, and the return of 286 mode is not a good reason, all the code was thrown out for that last year. The local heap will become needlessly fragmented by separating the header from the storage. The header should be kept contiguously with the allocated data for both local and global allocations. The preview code requires at least four sections of kernel global allocation use to be rewritten, but there is no need for that.

The rewritten portions of mm_alloc, etc getting rid of `pokew` are a good thing, along with defining structures for normal C access. However, likely because thoughts of far pointers being slower, the designed has changed from the elegant (but non-portably coded) earlier implementation of mm_alloc, to the overly complicated proposed version, which requires extra dereferences. Let me explain shortly.

Jumping to my conclusion, I can see the need for a local heap manager, *and* a global heap manager in the kernel. As I think deeper into it, I realize that these could and should be accessible identically to the kernel programmer. With both local and global heap allocators available, solutions become simpler - the mm_alloc rewrite absolutely should use the global heap manager, which will also manage the L2 buffers. Neither of these uses should require space in the local heap (except for buffer headers, something different).

__Details__
The programmatic API for both heap managers should operate as easily as possible for kernel programmers. That is, the main allocation routine should return an immediately usable pointer, *not* a structure pointer that needs another dereference. To wit:
```
/* C local heap*/
char *p = malloc(256);     /* get pointer to memory*/
free(p);       /* free uses allocated ptr and subtracts fixed amount to get private hdr*/
```
[...]

Issue #470, opened 2020-03-15
Latest build not bootable


Mellvik:
Unless I'm doing something stupid, the latest build is not bootable.

```
...
fd: probing disc in /dev/fd0
fd: floppy drive autoprobe failed!
hd: error: AX=0x09
BIOSHD: I/O error
dev: 300, sector: 2
...
```
Tried with several physical floppies. 
Yesterday's build worked fine.

--Mellvik

ghaerr:
Hello @Mellvik,

It looks like ELKS is trying to boot from dev 300, which is HD not floppy. @mfld-fr's modification to setup.S, where the BIOS boot device is passed from the boot loader, may have broken that,  is the only commit made in the last two days.

Mellvik:
Thanks @ghaerr,

does that analysis hold up with the fact that the image is bootable in QEMU?

Anyway, I'm looking forward to a fix - this is getting interesting. (re. test report #471).

—Mellvik

ghaerr:
@Mellvik,

I am mistaken. Yesterday's commit #467 did not touch setup.S. Instead, I was thinking of #445, from four days ago, which corrects a DMA segment boundary issue, and did modify 'setup.S'. You might try backing either of these out and seeing whether your HW floppy boot works or not. @mfld-fr can probably help you with instructions for how exactly to revert either of these on your local repo, I don't know git well enough, I'm sorry.

tkchia:
Hello @Mellvik, hello @ghaerr,

For some reason I could not reproduce the problem, after a `make clean && make` command.  Perhaps --- once again --- something wonky is going on with the dependency checking, which is causing some files not to be recompiled properly?

Thank you!

Mellvik:
Hi @tkchia,
Are you saying you could reproduce the problem before the make clean?

Anyway - I did a `make clean; make` to be sure, and the problem was still there ...
An image from 4 days back boots fine.

--Mellvik

mfld-fr:
`git clean -xdf`

Mellvik:
Thanks, @mfld,

I did all that. Even rebuilt tools and pulled a new copy of the entire tree. Still no boot (and to make sure, this is a different machine from yesterday):

mfld-fr:
Then, `git bisect` please

Mellvik:
OK:

usage: git bisect [help|start|bad|good|new|old|terms|skip|next|reset|visualize|view|replay|log|run]

Please advise.

M

[...]

PR #472, opened 2020-03-15
[kernel] Segment list to local heap

mfld-fr:
See #469


ghaerr:
Lots of big changes to the global memory allocator... considering that the two other previous options are completely removed with this commit, how do we know it works? What/how have you tested this? I am somewhat concerned about what can be done if for some reason new bugs appear in program execution at a later time with no easy options. (Reference sudden unknown issue with #470).

Do you want this merged before or after the 'latest build not bootable' #470 fix is found?

ghaerr:
Hello @mfld-fr,

Do you want this merged now? I didn't hear back as to what regression tests were performed, given that this replaces two previously selectable memory management options.

It would be nice to have some kind of user mode exercise application that allows improvements to be tested and later quantified/displayed as ELKS memory management is updated, exercising both local and global heaps.

PR #480, opened 2020-03-20
Updated more() & more

Mellvik:
more() partly rewritten, plus proposed minor changes to /etc/passwd and rc.sys.

Comments solicited - also on coding style/standards.

### DIFF
--- a/elkscmd/file_utils/more.c
+++ b/elkscmd/file_utils/more.c
@@ -19,17 +19,69 @@
 #include <utime.h>
 #include <errno.h>
 
+int fd;
+char mbuf[BUFSIZ];
+
+int more_wait(int fin, int fout, char *msg)  {
+	char buf[80], ch;
+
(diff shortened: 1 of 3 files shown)


ghaerr:
Looks like some nice enhancements to `more`. (And your first PR!)

Does hitting 'N' work without requiring a  \<return> afterwards? I was thinking setting raw mode would be required. It would also be nice to have \<space> work to get next page.

Adding the `/etc/passwd` user is nice way to start `sash` instead of `sh`. I am thinking perhaps a username like `root2` or perhaps `sash` might be better than `toor`.

Coding style is perfect. Nice work, thanks!

Mellvik:
Thanks @ghaerr,

raw mode for more() is coming up next - along with some more functionality (mentioned separately yesterday I guess).

'toor' - is root backwards, and a common pun - FWIW.

—Mellvik

PR #474, opened 2020-03-17
[kernel] [boot] fix DDPT clobbering in setup.S, & spew more informative error message upon disk read error in kernel

tkchia:

Issue #327, opened 2020-02-12
Roadmap for FAT tuning


mfld-fr:
Things are getting messy with 3 (very) active contributors on the same features and still some confusion or disagreement on the final FAT or boot design. So it is worth to agree on a roadmap to conclude and to keep everybody happy with the project:

- [x] create a staging branch 'fat_tune' (`master` now reserved for bug fixes and release 0.3.1): DONE
- [x] add 32-bit inode in configuration menu: DONE in 9350dd05674d2411c7c0d9aca2cadb7deb1ccc54
- [x] tune `setboot` to preserve existing extended FAT BPB (as described in https://github.com/elks-org/elks/pull/325#discussion_r377739122): DONE in a0007756958fc41d4d5d76484b49d38e07e5dec3
- [x] add FAT image case to `image/Makefile` : DONE in 60092a7af7d3ba54c1fa6030cda674e35bb664ea
- [ ] tune `setboot` to update ELKS BPB (geometry and payload as 2 different options)
- [ ] move FAT payload out of boot sector to relax space constraint
- [ ] always include ELKS BPB in boot sector: almost done in 7ac93835a3fbd437e9973d8b6542ed6258f681f1
- [ ] create a new `boot` file with the existing FAT payload
- [ ] add `boot` installation to package manager
- [ ] check 'HRS' FAT attributes for `boot` (as explained in https://github.com/elks-org/elks/issues/296#issuecomment-584208943)
- [ ] drop the FAT BPB duplication in payload
- [ ] extract FAT location of `boot` with `mtools` (`awk` is your friend, see https://github.com/elks-org/elks/issues/296#issuecomment-583873256)
- [ ] add generic payload location in ELKS BPB
- [ ] option : convert FAT payload in `boot` from ASM to C
- [ ] option: improve `boot` to handle `linux` fragmentation
- [ ] option: honor MBR parameters in boot sector (passed by MBR code)
- [ ] tune the CI to rename `fd1440.bin` to `fd1440-minix.bin`
- [ ] tune the CI to generate `fd1440-fat.bin`
- [ ] option: add non-regression testing for EMU86 in CI
- [ ] merge `fat_tune` into `master` and release (0.3.2 ?)

tkchia:
Hello @mfld-fr, hello @ghaerr,

Thank you.  Am I correct that this is a road map for the longer term?

From what I see, this is the current situation:
[...]

PR #482, opened 2020-03-20
[boot] modify _and_ restore floppy parameter table

tkchia:
This is a follow-up to the earlier quick bugfix in the commit 65da1a7d06567bb138f73abda441b4acd9dea1ef (_edit: which was part of PR_ https://github.com/jbruchon/elks/pull/474).

This new commit modifies the Diskette Drive Parameter Table (DDPT, `int $0x1e`) to give the correct number of sectors per track, and uses the new DDPT to read `/linux`.

It also now _restores_ the old DDPT just before handing over to `/linux`, or (upon an error) before rebooting.

To accommodate the code to restore the old DDPT, I also had to trim down the disk read logic for the FAT bootloader --- it now reads `/linux` more slowly, one sector at a time.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -20,12 +20,15 @@
 #define ELKS_INITSEG (0x0100)
 #define ELKS_SYSSEG  (0x1000)
 
-// Whether to copy and modify the BIOS floppy parameter table to allow
-// whole-track reads in the bootloader
+// Whether to try to do whole-track reads, or read one sector at a time
 //
-// This is currently disabled because it may clash with setup.S which may
-// overwrite the area used by the new table
-#undef FROB_DDPT
+// Whole-track reads are disabled for the FAT bootloader because we are a
+// bit short of code space :-(
+#ifndef BOOT_FAT
+#   define FAST_READ
+#else
+#   undef FAST_READ
+#endif
 
 	.code16
 
@@ -35,12 +38,10 @@
 
 entry:
 
-#ifdef FROB_DDPT
 // Allow the new BIOS floppy parameter table to clobber the first few bytes
 // of the boot sector after we are done with them
 
 floppy_table:
-#endif
 
 #ifdef BOOT_FAT
 	jmp entry1
@@ -68,6 +69,8 @@ entry1:
 	sub $0x1000,%ax
 	and $0xF000,%ax
 	mov %ax,%es
+	mov %ax,%ss   // automatic CLI for next instruction
+	xor %sp,%sp
 
 	xor %di,%di
 	mov %di,%ds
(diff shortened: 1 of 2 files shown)


ghaerr:
Hello @tkchia,

Getting complicated, huh?

I'm trying to fully understand... why not just leave DDPT in high memory, and have setup.S relocating itself to just below it, thus avoiding all this?

Also, why does boot code have to restore DDPT before handing over to /linux (which I assume is setup.s)? Couldn't setup.S find out DDPT has been moved?

Finally, on a hardware reboot, doesn't the BIOS reset the DDPT, thus our boot doesn't have to do it?

tkchia:
Hello @ghaerr,

> I'm trying to fully understand... why not just leave DDPT in high memory, and have setup.S relocating itself to just below it, thus avoiding all this?

Possible, but how do we do this in a well-defined manner?  For now `setup.S` thinks that all the memory after the BIOS data area up to the end of conventional memory is available for its use.  If some conventional RAM needs to be reserved for the DDPT, then we should probably update the protocol between the bootloader and `setup.S` to handle this.

> Finally, on a hardware reboot, doesn't the BIOS reset the DDPT, thus our boot doesn't have to do it?

The bootloader currently does a warm reboot (`int $0x19`) upon an error, and this does not reset any vectors.  Do you think we should change this?  (A `ljmp $0xffff, $0` should reset everything.)

Thank you!

mfld-fr:
Restoring the table before passing control is the safest way, and accomodates any boot protocol.

About the 'trim' you were forced to do with impact on performance, may be it would make you to understand why I am asking to move to a two stage boot loader (see #327) ?

tkchia:
Hello @mfld-fr,

> About the 'trim' you were forced to do with impact on performance, may be it would make you to understand why I am asking to move to a two stage boot loader (see #327) ?

Point taken; but I think even a 2-stage bootloader will still have to have FAT-specific code (rather than filesystem-independent code) in both stages.

Thank you!

ghaerr:
Hello @tkchia,

>  For now setup.S thinks that all the memory after the BIOS data area up to the end of conventional memory is available for its use.

Since apparently it's not a big deal to setup a DDPT, why not always have setup.S prepare a DDPT before doing any of its own I/O, and put it where it wants? Setup is already passed the CHS values required. This would seem to negate the requirement of communicating a boot DDPT to setup, as well as having to have boot reset it, using up valuable space. Boot could also do a hard rather than warm reset on error, why not?

tkchia:
Hello @ghaerr,

> This would seem to negate the requirement of communicating a boot DDPT to setup, as well as having to have boot reset it, using up valuable space.

No, no, no!  To wit:
  * The bootloader modifies the DDPT for its own use, mainly so that it can get the BIOS (i.e. `int $0x13`) to correctly do whole-track reads to read `/linux` (setup + kernel).
  * The kernel, well, also uses the BIOS to do disk I/O, which also involves the DDPT.
  * Setup does not do any disk I/O, but if the DDPT happens to be positioned in conventional RAM, there is a danger that setup (or the kernel) may clobber the DDPT, making disk I/O impossible afterwards.

Thank you!

tkchia:
Hello @ghaerr,

I will tweak the bootloader to do a hard(er) reboot.  Thank you!

ghaerr:
Good morning @tkchia,

Thank you for your explanations. I know understand that changing the DDPT is only for boot loader use, and that kernel disk I/O using BIOS should use the BIOS default DDPT, at least for now until the kernel does more than single-sector reads.

tkchia:
Thank you @ghaerr!  I hope to see how to make the kernel's disk I/O more efficient, at a later time.

PR #485, opened 2020-03-22
[image] Replace package manager with config image build system

ghaerr:
This PR removes the 'package manager' and replaces it with a system usable from `make config` or `make menuconfig`.

A new config option under 'Target Image' has been added 'Build extra binary images'. If selected, the same set of images built previously via the package manager are now automatically built after the normal configured image is built. These images include 360k, 720k, 1440 and 2880k images for Minix and FAT filesystems, along with a 32Mb Minix HD image. Image names have not been modified in this PR. A 32Mb FAT HD image can't yet be built due to `mformat` limitations.

The applications selected for each binary image are based on the new 'Select Applications by Image Size' config option implemented in #479.

Fixed `fd2880-fat.bin` to use a cluster size of 2 so that FAT12 can use the whole image size.

Addresses issues brought up in #290, which can now be closed.


ghaerr:
Hello @tkchia,

I wanted to give you a heads up on how this PR has slightly changed the image build process that you may not be yet aware of. The PR builds both the selected config image, and optionally additional extra images previously built by the package manager, all using the same process. Specifically, the following has changed:

- `setboot` is always run with -B to stamp the EPB on all images, including the config image.
- The 'image/Make.image' Makefile continues to remake the bootblocks directory, but this will have no effect on 'extra' images being built (thus requiring `setboot` run above).

The way the updated build mechanism works is by allowing an alternate 'config' file, in this case generated on the fly, to be used in 'elkscmd' (to install selected binaries), and 'image' (to build extra and normal images) directories. The 'bootblocks' directory always uses $TOPDIR/.config config file, and thus doesn't rebuild bootblocks based on a passed config file, so `setboot` is used.

After thinking about this, it may be desirable to always force-build 'bootblocks' as was done before, but enhance it to use a passed config file. In this way, the extra builds are always using compiled code, rather than stamping the EPB instead. Alternatively, the code and #ifdefs dealing with image size could be removed from bootblocks, and `setboot` always used. I'm thinking it might be wiser to use the former, although the latter is simpler.

What is your preference?

ghaerr:
Hi @tkchia,

I quickly looked at 'bootblocks/Makefile' and 'bootblocks/boot_sect.S' and see that it uses 'autoconf.h', rather than a config file, for when to remake as well as an include file. Thus, the problem discussed above is a bit more complicated. It was originally suggested in #290 that config files be used for extra images, but these need a shell script or tool conversion into an autoconf.h-like file that could be alternatively included in boot_sect.S to force its remake. These methods leave .o files that don't match the regular (non-extra) build, unless a different output directory was specified. Looking for suggestions at this point, unless leaving preset CHS values out of boot files and always using `setboot` to write the EPB is chosen as a good option.

tkchia:
Hello @ghaerr,

> Looking for suggestions at this point, unless leaving preset CHS values out of boot files and always using `setboot` to write the EPB is chosen as a good option.

Thanks for the heads-up!

The `setboot` arrangement already works for now, and is simple enough, so I guess my (admittedly very personal) preference will be to continue with it as far as it is feasible.

Indeed, I am wondering if, in the future, the bootloader code might actually need the kind of flexibility you allude to --- e.g. it might need to assemble different code for a hard disk bootloader and a floppy disk bootloader.  However, I hope that we do not have to go there.

Thank you!

ghaerr:
Hi @tkchia,

> The `setboot` arrangement already works for now, and is simple enough, so I guess my (admittedly very personal) preference will be to continue with it as far as it is feasible.

Ok.

> might need to assemble different code for a hard disk bootloader and a floppy disk bootloader. 

Since `setboot` takes a boot file parameter specified in 'image/Make.defs', should the bootloaders require different assembly options, outputting them as separate files easily solves this problem.

PR #488, opened 2020-03-23
[kernel] Small cleanups of offset macro and kdev_t.h

ghaerr:
Remove 'offsetof' defines in various places, already handled by 'stddef.h'.

Remove unnecessary outside 'include <stdint.h>' and  kdev_t definition for non-\_\_KERNEL__ use in 'kdev_t.h'.

Remove 'elkscmd/lib/\*', all unused routines.


ghaerr:
Second commit removes unnecessary handling of include <linuxmt/types.h> in 'kdev_t.h'.
Original offender was 'config.h' included 'kdev_t.h' unnecessarily.

'config.h' should not include 'major.h' but will leave that for another day.

PR #489, opened 2020-03-23
[libc] Libc header file cleanup

ghaerr:
This commit cleans up most compile-time warnings in libc. 

During cleanup of `_getuid` and `_getgid`, realized that both functions expect kernel to return effective uid or gid in passed 'int *', which is not implemented in kernel. Thus, both `geteuid` and `getegid` are buggy. This commit doesn't fix that.

There doesn't seem to be a way to know whether the 'syscalls.dat' argument count is in fact used by the kernel...


tkchia:
Hello @ghaerr,

Aye, this is a kernel bug which really needs to be fixed. :-|  (And, if I remember correctly, the `getpid` syscall should by right also return the parent process id.)  Thank you!

ghaerr:
@tkchia, I'll fix all three syscalls in a separate PR, thanks!

PR #492, opened 2020-03-26
[kernel] Last compiler warning cleanup for kernel and libc, other cleanup

ghaerr:
Eliminates all kernel and libc compiler warnings, except those that need careful attention and testing to fix.

Lots of minor fixes, including BCC-register-optimizations of using `register char *` then casting with `(int)`. I suspect this generated better BCC code, but is terrible use of C with `gcc`.

While reading compiler warnings for elimination, the following bugs were found and fixed:
Direct console driver ANSI insert line broken (affects `vi`).
`mkdir` on Minix FS buffer release bug on non-empty directory.
BIOS HD partition start sector and numsectors truncated when > 65535.

In order to reduce kernel compiler warning output, the various semaphore operators tagged as `deprecated` are temporarily turned off using 'DEPRECATED' define (@mfld-fr: one line change in 'sched.h' to turn them back on when you're ready to replace them).
 
Other minor changes:
BIOS floppy and hard disk driver always uses `bioshd:` for kernel messages, instead of `doshd:` and `hd:`.
Libc `const` decls added in some string functions (more to do, but stopping at compiler warnings).
Added many missing include files for libc functions in `ash`.
Added `debug_net` debugging macro that allowed tracing down networking application startup race condition over `tcpdev_inuse` kernel variable. The race condition is fixed in the next PR.

With kernel and libc warnings down to a bare minimum, it should be easier to track down bugs in new code, and fix remaining issues using the remaining warnings. 

A huge amount of compiler warnings remain for ELKS applications, far too many to tackle easily.

### DIFF
--- a/elks/arch/i86/drivers/block/doshd.c
+++ b/elks/arch/i86/drivers/block/doshd.c
@@ -165,19 +165,19 @@ static struct gendisk bioshd_gendisk = {
 static unsigned short int bioshd_gethdinfo(void)
 {
     unsigned short int ndrives = 0;
-    register char *drive;
+    int drive;
     register struct drive_infot *drivep = &drive_info[0];
 
-    drive = (char *)0x80;
+    drive = 0x80;
     do {
-	if ((drivep->heads = setupw((int)drive))) {
-	    drivep->sectors = setupw((int)drive + 2);
-	    drivep->cylinders = setupw((int)drive + 4) + 1;
+	if ((drivep->heads = setupw(drive))) {
+	    drivep->sectors = setupw(drive + 2);
+	    drivep->cylinders = setupw(drive + 4) + 1;
 	    drivep->fdtype = -1;
 	    ndrives++;
 	}
 	drivep++;
-    } while ((int)((drive += 6) <= 0x86));
+    } while ((drive += 6) <= 0x86);
     return ndrives;
 }
 
@@ -306,7 +306,7 @@ static void reset_bioshd(int drive)
  */
 #if 0
     if (CARRY_SET)
-	printk("hd: unable to reset.\n");
+	printk("bioshd: unable to reset.\n");
 #endif
 }
 
@@ -317,7 +317,7 @@ int read_sector(int drive, int track, int sector)
  * when new floppy probe is used
  */
 
-    register char *count = (char *)MAX_ERRS;
+    register int count = MAX_ERRS;
 
     do {
 	/* BIOS read sector */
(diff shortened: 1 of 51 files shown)


Mellvik:
@ghaerr,

this is an impressive milestone. And like so often recently, you go right ahead and fix something that's on my (growing) list. Much appreciated. This time it is (or I suspect it is) vi (ANSI fixes), which I wanted to check out right away.

However, after pulling all the updates, the 1440 image does not boot.
Probably a minor thing, maybe even my config, but still. QEMU screenshot enclosed.

—Mellvik

ghaerr:
Hi @Mellvik,

Can you post the screenshot you mentioned? It isn't attached to your last post.

Had you pulled down the latest prior to this PR and verified that your system boots? I want to try to make sure this is the PR that is affecting your boot. Despite the large cleanup, there are only a few places changed that should generate different code.

Mellvik:
Hi @ghaerr,

I noticed the screenshot was missing so I uploaded it, you'll get it via the web interface. (QEMU)

And no, I haven't pulled updates for some days, so I'm not sure about that.
I'll prepare a bisect as soon as I have created a PR for the sash changes this afternoon.

—Mellvik

ghaerr:
@Mellvik, I noticed I had left a 'reboot' at the end of /etc/rc.d/rc.sys, which is removed now. Not sure whether that was your problem, I was using it to iterate boots to catch the telnetd hang problem.

Mellvik:
@ghaerr,
I don't think so, my case stops in 'ELKS Setup'.

—Mellvik

ghaerr:
Still no screenshot.

If the hang is just after displaying 'ELKS Setup', there have been only two recent commits in that area, one 4 and the other 6 days ago, by @tkchia. You'll see them in the commit log. They are associated with modifying the DDPT buffer location and restoring it. Start there, that will allow a possibly quicker bisect.

Mellvik:
OK; thanks, I'll wind 6 back and start from there.
Image uploaded again, have a look.

-M

ghaerr:
Still no image showing here! But if it just shows a boot ending with 'ELKS Boot', no need, as the problem is in a known area.

IIRC, @tkchia fixed your first boot problem with a commit 8 days ago, also having to do with the DDPT area. The commit 6 days ago was supposed to be a permanent fix, and the one two days after a slight cleanup. These are associated with initial boot and setup operations before giving control to the kernel.

ghaerr:
Looking at this a bit further, there is also the possibility that my commit '[image] Replace package manager with config build system' could cause a boot failure. Try testing that one if the DDPT commits work.

Mellvik:
Skjermbilde 2020-03-26 kl  11 16 05

Mellvik:
@ghaerr,
I took a second run at make clean, this time with a menuconfig & save, and we're up and running. sorry for the confusion.

—Mellvik

PR #497, opened 2020-03-27
	add command history to sash(), fix wc()

Mellvik:
-	This commit adds basic csh-like command history to sash()
	!!, !-<num>, !<num>, ^x^y command line edit, HISTORY env variable to set the size of the hsitory list (def=20).
	And a 'history' command to list history.
-	TODO: Command search, save history on exit, change history size while runing, copy last argument ('!$') into current line.

Also in this commit: Fix wc() to work without arguments in a pipe.

### DIFF
--- a/elkscmd/minix1/wc.c
+++ b/elkscmd/minix1/wc.c
@@ -73,8 +73,6 @@ int main(int argc, char **argv)
   char *cp;
   int tflag, files;
 
-  if (argc < 2) goto usage;
-
   /* Get flags. */
   files = argc - 1;
   k = 1;
--- a/elkscmd/sash/Makefile
+++ b/elkscmd/sash/Makefile
@@ -12,7 +12,7 @@ include $(BASEDIR)/Make.rules
 
 ###############################################################################
 
-OBJS = sash.o cmds.o cmd_dd.o cmd_ed.o cmd_grep.o cmd_ls.o cmd_tar.o utils.o
+OBJS = sash.o cmds.o cmd_dd.o cmd_ed.o cmd_grep.o cmd_ls.o cmd_tar.o utils.o cmd_history.o
 #LDFLAGS += -maout-chmem=0xc000
 
 all: sash
--- a/elkscmd/sash/cmd_history.c
+++ b/elkscmd/sash/cmd_history.c
@@ -1,0 +1,249 @@
+/*
+ * HS (@mellvik) for the ELKS project - march 2020 
+ * 
+ * implements csh like command history for sash
+ * 
+ */
+
+#include "sash.h"
+#ifdef CMD_HISTORY
+
+#define HISTMAX 200
+#define HISTMIN 20
+static  int     lastcom = -1;   /* index of most recent command */
+static  int     histind = 0;    /* cmd # for history list */
+static  char    **histbuf;      /* array holding commands */
+int     histcnt = HISTMIN;
+static	void	phex(char *);
+static	void	phlist();
+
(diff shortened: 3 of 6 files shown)


ghaerr:
Nice work @Mellvik! 

I'll play around with it and let you know of any issues.

A couple minor issues for your next round: The compiler will check that no arguments are passed to a function when declared as 'int foo(void)'; not using 'void' turns off that checking. Also, 'static void init_hist()' in sash.c should be 'extern' as it's not in that file. Finally, using 'static' for the functions that aren't called outside cmd_history.c protects name collision and tells the programmer they aren't entry points.

Mellvik:
Thanks @ghaerr,

I'll put that on the todo list for sash! 
BTW - I'm wondering if it really makes sense to have a read only tar() as an internal sash command.
And the internal ls still acts as normal 'ls -a1' (that is 'a and one'). Is that what you wanted?

The system is now becoming quite useable. I'm hoping to (among other thing) get at getty going on com1 @ 19200 and see what happens real soon. Tha will put a few things to the test...

—Mellvik

ghaerr:
> I'm wondering if it really makes sense to have a read only tar() as an internal sash command.
And the internal ls still acts as normal 'ls -a1' (that is 'a and one'). Is that what you wanted?

Everything has been left as-is as part of the original `sash` so far, as bug fixing has been needed. 

The plan in 'core_utils' was to sort out how different the 'external' file_utils are from the built-in 'sash' commands, as all the file_utils external commands came into existence by someone pulling them out of `sash`.  Most of that has been done. A big difference is that `ls` has been greatly enhanced, and it may not be easy to insert it back into `sash`.  The 'columnize' function is engrained in the rest of the external 'ls', and so far have left in the 'a' since `sash` was supposed to be an emergency standalone shell. 

So, not quite sure yet what to do about 'tar', 'ls', and some others. Agree that the 'ls' builtin should be updated though.

ghaerr:
> The system is now becoming quite useable. I'm hoping to (among other thing) get at getty going on com1 @ 19200 and see what happens real soon. 

I'm very interested in your results on using ELKS through the serial port. Make sure to ^C at different times (getty, login, ash and sash) to test the controlling TTY and other fixes put in earlier!

Do you have any way to test networking over serial?

Mellvik:
@ghaerr,
My setup may lend itself to slip testing, I've never considered that. The terminal end is a raspberrypi2, so supposedly anything is possible. 
If anyone on this list has slip working on raspbian, it would be a good start.

I'll keep you posted when regular login is operational.

-- Mellvik

Mellvik:
﻿@ghaerr,

thanks for the clarification - I suggest ditching tar, fixing the issues with ls. ls() is sort of what makes sash great - instantaneous response.

BTW, I have a new tar running, and built-in tar doesn't seem to rest the tar files correctly. Also, while testing tar, i needed rm -rf, which is almost working. Never a dull moment.

-- Mellvik

Mellvik:
Sorry, the spelling helper is killing me, 'rest' should be 'read' in tar files...

--M

ghaerr:
I'll plan on porting 'ls' over to sash (or at least a good portion of it).

> I have a new tar running, and built-in tar doesn't seem to read the tar files correctly.

Test the external tar too, to make sure it has the same result, which I'm almost sure it will.
It would be nice to figure out whether its a structure problem for reading the tar directory or something else. Likely will remove sash tar and replace external tar with your version.

>  i needed rm -rf, which is almost working.

Post that in a PR too, when you have it. That was on my list too!

PR #500, opened 2020-03-28
[kernel] do disk I/O via DMASEG:0 only when necessary

tkchia:
This commit makes the ELKS kernel do BIOS disk reads/writes directly to/from kernel buffers, whenever there is no danger of crossing a 64 KiB DMA boundary, rather than always doing disk I/O through the special `DMASEG:0` buffer.

### DIFF
--- a/elks/arch/i86/drivers/block/doshd.c
+++ b/elks/arch/i86/drivers/block/doshd.c
@@ -44,9 +44,6 @@
 
 #ifdef CONFIG_BLK_DEV_BIOS
 
-#define DMA_OVR  // activate DMASEG patch
-/* #define MULT_SECT_RQ */
-
 /* the following must match with /dev minor numbering scheme*/
 #define NUM_MINOR	32	/* max minor devices per drive*/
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

I moved the `/* #define MULT_SECT_RQ */` line near the top of the file in my last commit, I think this is the only conflicting part, I'm sorry.

On another note, would you like me to go ahead with ignoring the BIOS disk information saved in 'setup.S' at 0x86 etc and doing it in `bioshd_gethdinfo` to complete the transition to 8 BIOS devices? I'll leave the decision as to whether afterwards the existing setup.S code should be removed or not to you.

tkchia:
Hello @ghaerr,

> I moved the `/* #define MULT_SECT_RQ */` line near the top of the file in my last commit, I think this is the only conflicting part, I'm sorry.

It is a small problem.  I have resolved the conflict on my end.

> On another note, would you like me to go ahead with ignoring the BIOS disk information saved in 'setup.S' at 0x86 etc and doing it in `bioshd_gethdinfo` to complete the transition to 8 BIOS devices? 

I am cool with that.

Thank you!

Issue #199, opened 2018-04-23
Move HD image creation back into main build


mfld-fr:
From review and testing of PR #197:

The standalone script `build_hd_image.sh` from @georgp24 generates an HD image by partitioning, installing the MBR-related boot blocks, and expanding the FD1440 image. Generate this HD image directly in the main build process.

mfld-fr:
HD image (**without MBR**) now included in configuration. Bootable as 'big' floppy image, but not as true HD image (see #348). Still some work to add an MBR into that HD image.

ghaerr:
Hello @mfld-fr,

How are you thinking this should be completed, since everyone seems to want this now?

How about the following, which would help for development and testing: When CONFIG_IMG_HD is set in config, the build creates two output files (for the time being), a 'flat or floppy' fd.bin, which is built using the size and CHS configured, using the same method now in image/Makefile, and also an hd.bin image, which would be the result of assembling a dev86 'bootblocks/mbr.s' ported to the new asm format and taking in the config CHS parameters, which would then be prepended to fd.bin, creating an MBR hard disk image. Initially, this would only be done for MINIX bootable images, leaving FAT to worry about later.

This would allow testing of both flat and MBR disk images, since there seems to be some booting issues presently with both. I'm trying to understand what exactly needs to get done to move HD boot and image creation forward.

mfld-fr:
@ghaerr : I like the idea to generate both FD and HD images in the same run. Moreover, if the FD image has always the `fd.bin` name, one would not have to edit the `qemu.sh` script whatever the image format is.

To complete the current HD generation, I think the easiest and most consistent would be to reuse @georgp24's stuff from his standalone script and inject it into the current image Makefile.

georgp24:
@mfld-fr : The menuconfig system currently supports generating a hard disk, but this is not partitioned nor formatted. So if you select that, you get no disk to boot from.

Therefore I suggest to modify menuconfig so you can generate both a floppy image plus a hard disk to be used as DISK2 in qemu.sh. With other words, not just either a floppy image or a hard disk image but floppy image and hard disk image to select and configure separately.

ghaerr:
Hello @mfld-fr,

>  I like the idea to generate both FD and HD images in the same run. Moreover, if the FD image has always the `fd.bin` name, one would not have to edit the `qemu.sh` script whatever the image format is.

Yes, a name change would make things more constistent, and then when building HD images in images/Makefile, both a flat and MBR version created.

 
[...]

PR #506, opened 2020-03-30
[kernel][boot] Add MBR boot from HD partition to ELKS

ghaerr:
This demonstration PR adds the ability to create and boot MBR partitions from ELKS Minix and FAT HD images. When finished, this addresses issues #199 and #406. 

This PR won't be committed as-is. As will be seen, there are issues that need discussion in order to continue.

After the extra images are built, the 'fixpartition' script prepends a single empty MBR sector to Minix and FAT HD flat disk images. Then, a temporary 'writepartion' program writes an active partition entry to the MBR boot sector. A very slightly-modified MBR boot (438 bytes) from `dev86` bootblocks/mbr.s is then written prior to the partition table.

When an MBR boot sector passes control to the ELKS boot sector (VBR), it passes a magic number in AX so the ELKS boot sector can determine whether the MBR boot loaded it. If so, the VBR records the disk image starting sector (16-bit sector offset) from DS:SI at offset 8, and uses that to offset all disk reads, which include both the second sector for Minix and all of the sectors for '/linux'. It then records this sector offset, if any, to a new 'setup' INITSEG location 0x1e4. When the ELKS kernel starts, it uses this sector offset at the time of converting the BIOS number to ROOT_DEV, to set ROOT_DEV to `/dev/bda1` if an offset is present. In this way, floppy images, flat images, and MBR images all work.

Not yet done: A port of an 'fdisk' program will be coming to allow better partition creation. The dev86 mbr.s should be converted to `ia16-elf-as` format.
(The 'fixpartition' script also modifies MSDOS 6.22 HD MBR to boot under QEMU by modifying the ending head and sector numbers).

Hello @tkchia, there are a number of problems I would like your input on:
- boot code mods are temporary and should be made cleaner.
- returning BIOS device in root_dev is getting messy. We need a cleaner way of passing a real root_dev (for ROM or possibly all cases).
- returning the BIOS device to kernel isn't really needed, perhaps have setup convert it to a root_dev (which is does anyways for ROM), and increment it with the partition number.
- I can't figure out a way to find the booted partition number with just the DS:SI pointer. If only the offset is used (as is now), it could be made to work by comparing partition sector offsets in the 'bioshd_conv_bios_drive()', by adding more code (not preferable).
- If the booted partition number were known, it could be just added to root_dev, and returning the BIOS number wouldn't be needed.

MBR ELKS Boot from HD partition 1:
Screen Shot 2020-03-29 at 9 33 26 PM

The dev86 'mbr.s' mods are:
```
diff --git a/bootblocks/mbr.s b/bootblocks/mbr.s
index 25f29fe..3c9c1dc 100644
--- a/bootblocks/mbr.s
+++ b/bootblocks/mbr.s
@@ -11,7 +11,7 @@ ORGADDR=$0600
 BOOTADDR=0x7c00
 
 pbr=0	        ! Make this a partition boot record for ldboot.
-direct=1        ! Direct boot from MBR to any sector.
+direct=0        ! Direct boot from MBR to any sector.
 
 copyright=0	! Add in the copyright message; if room.
 mbrkey=0	! Option to choose the boot record based on keystroke (107)
@@ -20,9 +20,9 @@ message=1	! Display boot message (6+message space)
 markptab=1	! Put an end marker just below the partition table.
 use512=0	! Put the end marker at byte 510..512
 
-linear=1	! Use the LBA BIOS addresses.
+linear=0	! Use the LBA BIOS addresses.
 useCHS=1	! Disable CHS if you need space.
-linCHS=1        ! Calculate CHS from linear mbr values.	 (41 bytes)
+linCHS=0        ! Calculate CHS from linear mbr values.	 (41 bytes)
 
 preboot=0	! Include the pre-boot loader. (40 bytes)
 
@@ -235,6 +235,7 @@ sector_loaded:
 	jnz	try_next_part	! No? Try next partition.
 
 	mov	bp,si		! LILO says some BBs use bp rather than si
+	mov	ax,#$4c65	! 'eL' ELKS magic number
 	jmpi	#BOOTADDR,#0	! Go!
 
 no_partition:
```


ghaerr:
Update:

No further boot modifications needed, only saving VBR sector offset from MBR.
Partition number is calculated during partition scanning from boot sector_offset (elks_offset).
BIOS drive and root_dev can stay as-is.

Next step: partition tools and convert mbr.s source format.

ghaerr:
Add `setboot -P` option to create partition table in MBR boot sector.
Added `hd32mbr-minix.bin` and `hd32mbr-fat.bin` to extra images build.
Fix FAT VBR boot sector, now works.
Removed 'fixpartition' and 'writepartition'.

Todo: convert mbr.s source format.

ghaerr:
Add 'bootblocks/mbr.S' MBR boot sector to ELKS build, and included on hd32mbr\*.bin images.
Revised and updated from original `dev86` mbr.s source.

Will boot ELKS MINIX, FAT or other OS partition. Small monitor allows selecting any partition or floppy to boot, or auto-boots active partition after a few seconds.

This should allow users to create partitioned disks to run ELKS and other OS's from.

Almost done. Updated `fdisk` (needs rename) utility for ELKS and host OS coming. 

First MBR boot

tkchia:
Helo @ghaerr,

Sorry for the late reply, and thanks for this work!

> * returning the BIOS device to kernel isn't really needed, perhaps have setup convert it to a root_dev (which is does anyways for ROM), and increment it with the partition number.
> * I can't figure out a way to find the booted partition number with just the DS:SI pointer. If only the offset is used (as is now), it could be made to work by comparing partition sector offsets in the 'bioshd_conv_bios_drive()', by adding more code (not preferable).

My current half-baked idea is that, since the kernel already scans the hard disks' partition tables anyway, we can probably modify e.g. `add_partition(`...`)` in `genhd.c` to compare partition offsets to figure out the root device _while_ doing the scanning.

I _think_ this will be less work than adding a separate scanning pass to `bioshd_conv_bios_drive(`.`)`, or trying to intuit the partition number from within the bootloader.

I am not sure if doing this work in "setup" will be better.  I guess there will be little problem if the root device will always be one of the 4 primary partitions.  If it is not, then things can get messy.

tkchia:
Hello @ghaerr,

> * boot code mods are temporary and should be made cleaner.

Roger that. ☺  Some questions and feedback:

  * Do you intend to support MBRs other than the `dev86` MBR in the future?  (I do realize that this will be somewhat difficult, since some MBRs such as the [Windows Vista MBR](https://thestarman.pcministry.com/asm/mbr/VistaMBR.htm) only properly set `%ds:%bp` and not `%ds:%si`.)
  * Any particular issues with expanding the `sect_offset` (partition start sector) to 32 bits?  I think `disk_read(`...`)` can be easily extended to support partitions that go beyond the first 65,536 sectors --- as long as the sector number within the partition is below 65,536.  But maybe there is something else I missed.

Thank you!

ghaerr:
Hello @tkciha, thanks for your comments!

> I  can't figure out a way to find the booted partition number with just the DS:SI pointer. 
>> we can probably modify e.g. add_partition(...) in genhd.c to compare partition offsets to figure out the root device *while* doing the scanning.

After asking that question, I realized just what you're suggesting, and the code in genhd.c was changed to compare offsets to determine the partition number, which is then used in `bioshd_conv_bios_drive`.

> Do you intend to support MBRs other than the dev86 MBR in the future? 

I have been reading a lot of boot code lately, including Linux and GRUB. All of the major OSes have their own MBR, with the exception of Linux which uses separate projects to boot it, and GRUB has separate VBRs for Linux, CD, etc. There are lots of reasons for each of them. I would summarize that it's all about tradeoffs.

ELKS and the present code doesn't have separate HD and FD boot sectors (VBR's) for simplicity. The current code has to use a magic number in order to determine whether an MBR boot is in process or not. In order to support other MBRs, as well as get rid of the current magic number, ELKS needs to have separate VBRs for floppy versus hard disks (we already have separate VBRs for Minix vs FAT).

It would seem to be a good idea that ELKS should be bootable from other (standard) MBRs. This requires separate FD versus HD boot blocks, but I think it would be worth the tradeoff. The boot_sect.S source would be ifdef'd like it is now, so maintainability shouldn't be an issue.

> Any particular issues with expanding the sect_offset (partition start sector) to 32 bits? 

No, except that we're out of space in the FAT boot to do that. The 16-bit start sector works up to 32MB drives (CHS 63x16x63) for the extra images produced now. With more space, this is easily changed. I'm not certain that larger hard disk sizes have been tested with ELKS.

I changed the boot_sect.S 32-bit ELKS magic magic number check for setup in the FAT boot to 16 bits in order to add the MBR code (I don't really think 32-bit check is needed, everywhere else magic numbers are 16 bits). Because the HD VBR now has to save the start offset, more bytes than should be needed are used to setup a segment register for that, only to be redone after the copy to high memory.  A slight rewrite of the boot sector startup should be able to save space. I have some thoughts on that after reading all the `dev86` boot blocks.

My plan is to commit the current code now, which will allow others to play with an operational system, then work with you for permanent changes to the boot sector(s) for alternate MBRs and 32-bit sector offsets.

tkchia:
Hello @ghaerr,

> > Any particular issues with expanding the sect_offset (partition start sector) to 32 bits?

> No, except that we're out of space in the FAT boot to do that. The 16-bit start sector works up to 32MB drives (CHS 63x16x63) for the extra images produced now. With more space, this is easily changed.

OK, I understand now.

> My plan is to commit the current code now, which will allow others to play with an operational system, then work with you for permanent changes to the boot sector(s) for alternate MBRs and 32-bit sector offsets.

I am fine with this.  Thank you!

PR #508, opened 2020-03-30
[setup] update comments: setup.S no longer reads hard disk geometries

tkchia:
Follow-up to bd8d28ef6dcfae74b068ef751bc5f6ca7f8cf0ad .


ghaerr:
Hello @tkchia,

This is partly correct.  At least for the time being, due to complexity in the BIOS disk driver determining exactly which floppy format to use and when, setup.S still reads the disk geometries for floppy drives. Only hard drives were moved out of setup.

I think in order to remove the floppy geometry read from setup.S, which could still be a good idea, a very close look at doshd.c is needed, given the config variations for floppy setup.

tkchia:
Hello @ghaerr,

> setup.S still reads the disk geometries for floppy drives. 

Whoops --- thanks for the correction!  I will modify the commit to correct this.

PR #509, opened 2020-03-31
[kernel][cmds] Fix negative PIDs, baud rate on /etc/inittab line

### LINKED ISSUE #507: Wild processes and weird behaviour
I ran into this when attempting to run getty on the serial line (from inittab). The approach was apparently not very smart, and getty is broken (on my list), but that's a different story.
What happens is (probably) that getty starts looping.
The first symptom that something is wrong is an ash() error message for all commands: 
`Cannot fork`
ash() does fork however, and the commands work just fine.
The other odd behaviour is from ps - reporting several zombies, garbage command lines and negative process numbers. Further - init (process 1) has an interesting command line 'umped'.

I'm not going to try to analyze this, but figured we have one or more bugs here. The inittab line is
`2:2345:respawn:/bin/getty /dev/ttysS0 9600`
See attached screen dump.

--Mellvik
Skjermbilde 2020-03-30 kl  12 34 26

ghaerr:
Fixes #507.

Kernel no longer allocates negative process IDs.
/bin/init now properly processes lines containing optional baud rate after tty line.
Heavy cleanup of getty.c to enable console debug messages for debugging.

@Mellvik,
As usual, long story on all your findings. Turns out that `/bin/getty` never even got executed with the optional baud rate added after /dev/ttyS0. `/bin/init` was trying to open "/dev/ttyS0 9600" and failed, and immediately tried again until no process slots. After that was fixed, once getty finally ran, it didn't set the passed baud rate. Let me know whether the baud rate is set correctly, but should you need to fix things, that's all handled in getty.c.

### DIFF
--- a/elks/kernel/fork.c
+++ b/elks/kernel/fork.c
@@ -21,7 +21,7 @@ pid_t get_pid(void)
 	if (p->pid == last_pid || p->pgrp == last_pid ||
 	    p->session == last_pid) {
 	  startgp:
-	    if (++last_pid < 0)
+	    if ((int)(++last_pid) < 0)
 		last_pid = 1;
 	    p = &task[0];
 	}
--- a/elkscmd/rootfs_template/etc/inittab
+++ b/elkscmd/rootfs_template/etc/inittab
@@ -11,5 +11,6 @@ si::sysinit:/etc/rc.d/rc.sys
 #ud::once:/sbin/update
 
 1:2345:respawn:/bin/getty /dev/tty1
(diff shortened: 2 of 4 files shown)


Mellvik:
@ghaerr,
With a smile I'm thinking you're having as much fun fixing these things as I have finding them :-). Impressive work, ditto speed. Let's keep it up, and we're have an (even more) useable system real soon.

BTW I have a bunch of updates pending, coming when I win my current fight with git. I also have some facts on the disk parameter issues we discussed over the weekend. More on that on a different thread.

--Mellvik

PR #520, opened 2020-04-05
[boot] [kernel] allow boot from hd partitions above first 32 MiB

tkchia:
I also disabled the code for modifying the DDPT when the bootloader only does sector-by-sector reads (i.e. for the FAT bootloader).  This frees up some code space.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -32,14 +32,16 @@
 
 // Macro to restore the original (i.e. BIOS's) Diskette Drive Parameter Table
 //
-// Note: this clobbers BX and DS, and resets SP!  Use it only upon an error
+// Note: this may clobber BX and DS, and reset SP!  Use it only upon an error
 // or just before handing over to /linux
 .macro RESTORE_DDPT
+#ifdef FAST_READ
 	mov $-8,%sp
 	pop %bx
 	pop %ds
 	popw (%bx)
 	popw 2(%bx)
+#endif
 .endm
 
 	.code16
@@ -67,12 +69,17 @@ entry1:
 #endif
 
 	// instructions through _next0 not needed for floppy, only MBR boot
+	cld
 	cmp $0x4c65,%ax			// coming from ELKS MBR?
 	jnz _next0
-	mov 8(%si),%bx			// get MBR start sector from partion table at DS:SI
-	xor %ax,%ax				// save in sect_offset
-	mov %ax,%ds
-	mov %bx,BOOTADDR+sect_offset
+	mov 8(%si),%ax			// get MBR start sector from partion table at DS:SI
+	mov 10(%si),%si
+	push %cs
+	pop %es
+	mov $BOOTADDR+sect_offset,%di
+	stosw
+	xchg %ax,%si
+	stosw
 _next0:
 
 	// Get the memory size
@@ -97,10 +104,10 @@ _next0:
 	mov %di,%ds
 	mov $BOOTADDR,%si
 	mov $256,%cx  // 256 words = 512 B
-	cld
 	rep
 	movsw
 
+#ifdef FAST_READ
 	// Copy the current BIOS floppy parameter table
 	// so that we could modify it locally
 	//
@@ -129,6 +136,7 @@ _next0:
 	mov %cx,%ds    // CX = 0
 	popw (%bx)
 	mov %es,2(%bx)
+#endif
 
 	// Rebase CS DS to work in the 64K segment
 
@@ -151,10 +159,12 @@ _next1:
 	mov $msg_boot,%bx
 	call _puts
 
+#ifdef FAST_READ
 	// Set sector count in floppy parameter table
 
 	mov sect_max,%al
 	mov %al,floppy_table + 4
+#endif
 
 #ifndef BOOT_FAT
 	// Load the second sector of the boot block
@@ -274,7 +284,6 @@ _disk_read:
 	mov $5,%bh
 	push %bp
 	mov %sp,%bp
-	add sect_offset,%ax
 	push %ax
 	push %dx
 	push %cx
@@ -295,6 +304,8 @@ dr_loop:
 	mulb head_max
 	xchg %ax,%bx
 	xor %dx,%dx
+	add sect_offset,%ax
+	adc sect_offset+2,%dx
 	div %bx
 	mov %al,%ch      // CH = low 8 bits of physical cylinder number
 	ror %ah
@@ -491,6 +502,7 @@ binfile:
 	mov drive_num,%dl
 	xor %dh,%dh
 	mov sect_offset,%cx
+	mov sect_offset+2,%bx
 	mov $LOADSEG,%ax
 	mov %ax,%ds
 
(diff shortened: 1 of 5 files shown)


ghaerr (commented):
Hello @tkchia, thanks for the fixups!

I had been wondering about the need for the DDPT without multi-sector reads, this clears that up.
And nice moving 'sect_offset' into the FAT BPB for space! I never thought of that.

The 32-bit partition offset code will only be needed when ELKS has been installed on a large hard drive outside of ELKS, as all our builds put ELKS starting at the second cylinder. However - for the 32-bit offset case, the current ELKS VBR requires our MBR to be installed, which can't be assumed. (This is because of the magic number processing first thing). So adding this basically mandates the use of separate VBRs for ELKS MBR images vs flat/floppy images going forward, to get rid of the need for the magic number processing (which I think should be done anyways).

Has this been tested? Do we need to generate a large disk image for it? None of the current extra images are large enough to use more than 16 bits of offset. I can add a larger image to the extra build if needed.

I don't understand the ROOT_DEV | minor test in genhd.c line 74. Is that supposed to be ensuring that we only check partitions on the BIOS drive that's being booted, not a second one?  There is no minor number set in ROOT_DEV at that time, and ROOT_DEV is a BIOS number at that point correct?

I have some other comments about the boot sector in general that I'll put in another comment.

ghaerr:
Hello @tkchia,

It seems that we're using extra code to setup segment registers to save the DS:SI addressable start sector at the start of the boot sector code, only to reset the seg registers for the VBR copy, then set them up again. I am wondering about ways to make the boot process simpler.

If the VBR could be kept in low memory somehow, addressable by zero'd seg registers, things would get lots simpler.

What is the reason for copying the VBR to high memory? Is this because the system blob is being loaded at 0:7C00 in order to guarantee not crossing a 64k boundary for BIOS sector reads? Thus only an issue for the multi-sector case?

ghaerr:
Oops - my standardization commit moved a few lines of code above the partition offset check, causing this conflict.

tkchia:
Hello @ghaerr,

> What is the reason for copying the VBR to high memory? Is this because the system blob is being loaded at 0:7C00 in order to guarantee not crossing a 64k boundary for BIOS sector reads? Thus only an issue for the multi-sector case?

The `/linux` blob is currently loaded at `ELKS_INITSEG:0`, i.e. `0x100:0`.  Also, the bootloader --- especially `boot_minix.c` --- may need quite a fair amount of static memory to maintain its own data structures.  And, I guess we would like the bootloader's static memory not to cross a DMA boundary.

I guess some simplification _might_ be possible, e.g. perhaps the bootloader can move itself to `0:0x600` or some such, and `/linux` can go to `0x1000:0`, then `setup.S` can move itself and the kernel where it should go (`0xc0:0`), etc. ...

Thank you!

tkchia:
Hello @ghaerr,

> So adding this basically mandates the use of separate VBRs for ELKS MBR images vs flat/floppy images going forward, to get rid of the need for the magic number processing (which I think should be done anyways).

I will see if I can tweak the bootloader (VBR) code to do some basic sanity checking of the partition table entry (`%ds:%si` or `%ds:%bp`) in a way that works with other MBRs.

Thank you!

ghaerr:
Hello @tkchia, thanks for your explanation.

> I guess some simplification might be possible, e.g. perhaps the bootloader can move itself to 0:0x600 or some such, and /linux can go to 0x1000:0, then setup.S can move itself and the kernel where it should go (0xc0:0), etc. .

That would be nice, but the MBR code already relocates itself to 0:0x0600, so that area can't be used.

> Also, the bootloader --- especially boot_minix.c --- may need quite a fair amount of static memory to maintain its own data structures.

Thus the boot loader and the payload .bss must operate lower than ELKS_INITSEG, right?

tkchia:
Hello @ghaerr,

> > e.g. perhaps the bootloader can move itself to `0:0x600` or some such

> That would be nice, but the MBR code already relocates itself to 0:0x0600, so that area can't be used.

Actually it can. 😃  The bootloader (VBR) just needs to extract the partition start sector before relocating itself (in fact it already does this).  Once the VBR knows the partition start, it has no more use for the MBR.

> Thus the boot loader and the payload .bss must operate lower than ELKS_INITSEG, right?

Yes.  Thank you!

ghaerr:
> I will see if I can tweak the bootloader (VBR) code to do some basic sanity checking of the partition table entry (%ds:%si or %ds:%bp) in a way that works with other MBRs.

This seems to be quite varied, from what I've seen in other VBRs. Many don't use DS:SI or DS:BP at all, and get the partition offset from other quite complicated means. I'm not sure we have/want those means. For instance, the following ignore the partition table pointer entirely:

- For Linux GUID volumes, the offset is taken from the "PARTID=" (exact name may be wrong) text descriptor attached to each GUID volume.
- Grub uses a known offset and builds a specific VBR with the offset compiled in.
- Windows Vista ignores it and uses an internal method later.

Looking at the Linux boot and partition management code, there is *WAY* too much code written to deal with identifying partitions, so it seems simplest and quite doable to use separate VBRs for MBR vs flat images, use the partition table pointer as-is (without validation) and pass the partition offset up to setup/kernel. It will be the rare case that ELKS is used with other MBRs, and we won't be able to make it work with all MBRs.

ghaerr:
> Actually it can. 😃 The bootloader (VBR) just needs to extract the partition start sector before relocating itself (in fact it already does this). Once the VBR knows the partition start, it has no more use for the MBR.

That's very good news. With the VBR relocating to 0:0x0600 (like the MBR already does), the segment registers can zeroed or set such that all VBR and buffer variables can be accessed, with a lot less fooling around. This could be a bit of a rewrite, but things should be a lot simpler and smaller.

Porting the `dev86` MBR to `gas` brought to light some important differences in the way `gas` vs `as86` boot assembler programs have to be written. The original `as86` MBR was easily written to start with all segment registers zero with `as86`s interpretation of .org, but `gas` makes that harder since an '.org' statement would fill the binary with 0's. In mbr.S, I was able to go halfway and have one segment register set to 0, while the other set to access 'local' variables at 0x0600, and in that way allow easy access to disk buffer contents as well as local variables without having to reload segment registers later. See 'mbr.S' startup for details.

> Once the VBR knows the partition start, it has no more use for the MBR.

Check out the Windows Vista VBR (https://thestarman.pcministry.com/asm/mbr/VistaVBR.htm), it uses the partition table pointer to hold the variables for the sector reads, (they happen to all be in the right place and order, since the partition table was in fact designed to be used this way). Thanks for sending the Vista MBR link, very interesting reading!

ghaerr:
> if (ROOT_DEV == (0x80 | minor >> hd->minor_shift)) {

Ok, I understand now. We should perhaps have a comment indicating that ROOT_DEV at this point contains the boot BIOS drive number, since the value and meaning of ROOT_DEV changes later to be a kdev_t.

ghaerr:
Thanks @tkchia. I am fine with your committing this.

tkchia:
Hello @ghaerr, OK, thanks!

PR #525, opened 2020-04-07
sash history bugfix

Mellvik:
-- Fixed missing string termination in history cmd edit
	-- :p - print command from history w/o execute did not work
	-- Added history command augmentation (!23 more arguments).
	-- Added global startupfile for sash


ghaerr:
Can you change 'GMT' to 'GMT0' in /etc/sashrc? That meets the POSIX standard for TZ and also shows users that the numeric portion needs to be set, as tzset won't work with say 'TZ=PST'.

It might be a better idea to use /root/.sashrc than /etc/sashrc. ELKS doesn't currently set global settings for `sh` in /etc and it might be easier to find and set in the home directory. Using ~/.sashrc at least matches what we do with the other shell. This also allows testing for other users by not having a globally-mandated setting. I had previously gone through both sh and sash source and set defaults such that no startup config is required.

A comment on 'magic' numbers like 80, 82 etc in cmds_history.c: it would be nice to have the base number (80) grabbed from a common include file, so that should the buffer size increase, your code doesn't break. [EDIT: looks like you're only using '80' in that particular function, so ignore my comment!]

Mellvik:
Thanks @ghaerr,

> 7. apr. 2020 kl. 18:01 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> Can you change 'GMT' to 'GMT0' in /etc/sashrc? That meets the POSIX standard for TZ and also shows users that the numeric portion needs to be set, as tzset won't work with say 'TZ=PST'.
> 
> 
Sorry, I forgot that one!
> It might be a better idea to use /root/.sashrc than /etc/sashrc. ELKS doesn't currently set global settings for sh in /etc and it might be easier to find and set in the home directory. Using ~/.sashrc at least matches what we do with the other shell. This also allows testing for other users by not having a globally-mandated setting. I had previously gone through both sh and sash source and set defaults such that no startup config is required.
> 
> 
Sounds good. I had the idea that it would be useful for new users to find the global startupfiles  and thus know about their existence and purpose. The same can be argued for the per user startup scripts though.
> A comment on 'magic' numbers like 80, 82 etc in cmds_history.c: it would be nice to have the base number grabbed from a common include file, so that should the buffer size increase, your code doesn't break.
> 
> 
OK, this is more like quick and dirty, not nice.

I will pull the PR and make changes.


—Mellvik

ghaerr:
> I will pull the PR and make changes.

If you 'git checkout test2', make changes, and commit, they will automatically show up and be merged here.

Mellvik:
Thanks - that saved me quite some time.

-M

ghaerr:
Looks good @Mellvik but you still didn't change the TZ GMT0!

ghaerr:
Also is the buffer overflow still possible on line 117?

Mellvik:
Yes, it is possible, but not likely. cmd[] is the command line passed down from main, length CMDLEN, 512. It's good coding practice to check, and it's more compact the way it is.


What do you think?

-M

ghaerr:
> Yes, it is possible, but not likely. What do you think?

I think we're ok for now, thanks!

ghaerr:
@Mellvik,

Now that `sash` is reading .sashrc, it's also producing prompts when it executes those commands. Can you look in to turning off the prompt during that process? For some reason, I thought you had already handled that. 
Screen Shot 2020-04-07 at 3 42 05 PM

Mellvik:
Indeed I had!! just looked through the diffs, and it's not there. So the got lost in one of my early fights with git.

I'll add them back in the morning and update the PR. Sorry about that …


-Mellvik

Mellvik:
Thanks for the patience, @ghaerr,
a timely reminder to test, test, test. Changes have been pushed.

-M

ghaerr:
> Changes have been pushed.

If you mean 'pushed to GitHub to add to your last commit', that only works when while the PR is open.

Mellvik:
I didn't notice the PR had been merged … apologies!

-M

PR #529, opened 2020-04-08
Sash history fix

Mellvik:
Added 'lost' fixes to sash.c - read startup files correctly.

Cleanup in sash/cmd_history.c, added 3rd caret in history editing, enabling substitutions to have trailing blanks: ^this^that  ^ (shell command line processing removes trailing blanks which makes this necessary).

Temporary fix to more(1) to work on serial lines. /dev/tty apparently  some work.

### DIFF
--- a/elkscmd/file_utils/more.c
+++ b/elkscmd/file_utils/more.c
@@ -62,21 +62,20 @@ int cat_file(int m_in, int m_out) {
 	}
 	return (m_stat);
 } 
- 
 
 int main(int argc, char **argv)
 {
 	int	cin, multi, mw;
-	char	*name;
-	char	ch;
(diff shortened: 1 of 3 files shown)


ghaerr:
Thanks for fixing more also. I'm not sure yet of how to fix /dev/tty, as I need to be able to emulate serial and can't yet.

Glad to see you're back looking at compiler warnings too. I fixed one of these in my PR cleanup late last night, but I'll commit yours first and change mine!

Mellvik:
Thanks @ghaerr.

BTW - the init problem disappeared for now, cannot repeat it, will keep it on my list.

Tar() is backlogged, but builtin sash tar now reads the files, so we're OK on that part.

—Mellvik

Issue #530, opened 2020-04-09
[Enhancement] 1.2M FD image


Mellvik:
Enhancement request:

Need to build 1.2M floppy images for testing. 

--Mellvik

PR #533, opened 2020-04-10
[config] Allow creation of 1.2M (1200K) floppy images

ghaerr:
Create 1.2M floppy images from config, for #530. Creates 'fd1200.bin'.

Since the standard applications are too big for 1.2M, you have to select 'Apps based on Image Size' as well in config in 'Userland'.


tkchia:
Looks good to me... I guess? :-)

Mellvik:
Thanks!!

I'll have a look and close up.

-M

Issue #454, opened 2020-03-11
Serial port access unreliable in ELKS


pawosm-arm:
I've seen this problem on two very different XT machines: Turbo-XT with 8bit ISA card (two bare 8250's on it) and Amstrad PC 2086 (on-board Amstrad 40049 UART, the same as in Amstrad Portable PC models). Communication with other PC running full blown Linux is haunted by lost or malformed characters and using this port for a Microsoft Serial Mouse causes nano-X to hang entire machine.
Contrary to this, Telix running under FreeDOS can go as fast as 115200bps on both XT machines.

ghaerr:
> The system is completely dead and unresponsive after the crash, pressing Esc has no effect, it must be kernel that hung.

I am wondering whether the mouse was sending ^C and ^Z characters and the system thought that the serial port had a controlling TTY, and sent signals... I recently fixed signal handling (yesterday's commit), it would be interesting to know whether this still happens, or whether this is a bug in the input processing of the serial driver.

ghaerr:
A good test to determine whether signal handling is the problem for nano-X or other applications would be for you to comment out the following lines in 'elks/arch/i86/driver/char/ntty.c':
```
    if (key == ttyp->termios.c_cc[VINTR])
        sig = SIGINT;
    if (key == ttyp->termios.c_cc[VSUSP])
        sig = SIGTSTP;
```
A possible problem is that otherwise, all programs would have to either set raw tty mode, or ignore signals with an explicit call to signal(..., SIG_IGN), which most do not. 

I am thinking of removing ^Z/SIGTSTP signal generation in ntty.c in ELKS for this reason.

pawosm-arm:
> I am wondering whether the mouse was sending ^C and ^Z characters and the system thought that the serial port had a controlling TTY, and sent signals... I recently fixed signal handling (yesterday's commit), it would be interesting to know whether this still happens, or whether this is a bug in the input processing of the serial driver.

I've tried current master, nano-X still hangs, yet this time I pressed Ctrl+C, the on-screen blob got bigger (as if there was a message outputed in it, though it can't be read as the blob is filled with stripes). Pressing Ctrl+C again and again, I've filled entire screen with the blob. I've tried to type `reboot` command, yet with no success. The kernel isn't entirely hung then, yet nano-X (landmine game) can't return to shell.

ghaerr:
I see. Sounds like it could be the shell returning after ^C, and displaying (badly) on screen.

[...]

Issue #515, opened 2020-04-03
getty/serial test report


Mellvik:
Summary: Close, but no cigar
Testing on physical HW & virtualbox, the latter turns out to lend itself well to serial testing: Serial line set up to listen to a TCP port, which can be connected to using netcat.

1. There is a problem with the serial driver (or maybe somewhere in the call chain from open() and down to physical) as we've seen before: `sash < /dev/ttyS0 > /dev/ttyS0 &` should work. When it eventually does, we're almost there. The error message is `Cannot create /dev/ttyS0: error 16`. Replacing sash with `getty` or `login root` doesn't make any difference. Neither does changing the shell we're running from. Replacing ttyS0 with tty2 or tty3 (virtual consoles) works fine. Further - `sash < /dev/ttyS0 &` works, takes input from serial, output to the console. And vice versa if using ttyS0 for output instead. Seemingly, when the device can accept several opens, this is going to work. (?)
2. There seem to be a stdio-problem and possibly a file descriptor inheritance problem in getty. The device in arg[1] is never opened by getty. Thus starting getty writes to and reads from the console regardless. More about that below.
3. Starting getty from inittab 'hangs' (loops) the system. All testing is done from the command line.
4. Opening the device explicitly in getty and then replacing stdio 0,1,2 before the execv using dup2() works. There are some minor problems with the terms settings - to be sorted out. I cannot figure out why dup2 is accepted but not the shell redirects.

--Mellvik
<img width="734" alt="Skjermbilde 2020-04-03 kl  11 04 32" src="https://user-images.githubusercontent.com/3629880/78343309-ef64de80-759a-11ea-96f2-38307a352fef.png">

Mellvik:
UPDATE:
Some adjustments to getty, and **serial login is working** via inittab.

Still some minor issues (commands echoed twice), but seemingly - when started from inittab, the device gets opened only once, and we're OK.

Great work, @ghaerr !!

ghaerr:
> Some adjustments to getty, and serial login is working via inittab.

Good news! What kind of changes were made to getty?

A few explanations on your statements above:

> The error message is Cannot create /dev/ttyS0: error 16. 

Errno 16 is device busy, /dev/ttyS0 is protected against being opened twice.
Trying to run a shell with both 0 and 1 redirected isn't really the way to do it.

> There seem to be a stdio-problem and possibly a file descriptor inheritance problem in getty. The device in arg[1] is never opened by getty.

[...]

PR #535, opened 2020-04-10
[kernel] Rewrite serial interrupt service routine for speed

ghaerr:
I'm unable to test this on real hardware right now. Testing with QEMU shows much less character loss, but still having some. @Mellvik, if you could test whether this runs on your hardware when you have time, that would be much appreciated. Should it not work, the "#if 1" can be changed to "#if 0" to get the old routine back.

This work should also help #454.

The revised interrupt routine quickly reads from the UART with interrupts off, into a small buffer. Afterwards with interrupts enabled, the characters are inserted into the tty queue and waiting processes are woken up. The previous routine did the queue insertion and wakeup for each character received, while the chip still may have had the interrupt ID or data ready bits set.

### DIFF
--- a/elks/arch/i86/drivers/char/ntty.c
+++ b/elks/arch/i86/drivers/char/ntty.c
@@ -300,34 +300,33 @@ size_t tty_read(struct inode *inode, struct file *file, char *data, size_t len)
 int tty_ioctl(struct inode *inode, struct file *file, int cmd, char *arg)
 {
     register struct tty *tty = determine_tty(inode->i_rdev);
-    register char *ret;
+    int ret;
 
     switch (cmd) {
     case TCGETS:
-	ret = (char *)verified_memcpy_tofs(arg, &tty->termios, sizeof(struct termios));
+	ret = verified_memcpy_tofs(arg, &tty->termios, sizeof(struct termios));
 	break;
     case TCSETS:
     case TCSETSW:
     case TCSETSF:
-	ret = (char *)verified_memcpy_fromfs(&tty->termios, arg, sizeof(struct termios));
+	ret = verified_memcpy_fromfs(&tty->termios, arg, sizeof(struct termios));
 
 	/* Inform driver that things have changed */
-	if (tty->ops->ioctl != NULL)
-	    tty->ops->ioctl(tty, cmd, arg);
+	if (ret == 0 && tty->ops->ioctl != NULL)
+	    ret = tty->ops->ioctl(tty, cmd, arg);
 	break;
     default:
-	ret = (char *)((tty->ops->ioctl == NULL)
+	ret = ((tty->ops->ioctl == NULL)
 	    ? -EINVAL
 	    : tty->ops->ioctl(tty, cmd, arg));
     }
-    return (int)ret;
+    return ret;
 }
 
-int tty_select(struct inode *inode,	/* how revolting, K&R style defs */
-	       struct file *file, int sel_type)
+int tty_select(struct inode *inode, struct file *file, int sel_type)
 {
     register struct tty *tty = determine_tty(inode->i_rdev);
-    register char *ret = 0;
+    int ret = 0;
 
     switch (sel_type) {
     case SEL_IN:
@@ -336,14 +335,14 @@ int tty_select(struct inode *inode,	/* how revolting, K&R style defs */
 	select_wait(&tty->inq.wait);
 	break;
     case SEL_OUT:
-	ret = (char *)(tty->outq.len != tty->outq.size);
+	ret = (tty->outq.len != tty->outq.size);
 	if (!ret)
 	    select_wait(&tty->outq.wait);
 	    /*@fallthrough@*/
 /*      case SEL_EX: */
 /*  	break; */
     }
-    return (int) ret;
+    return ret;
 }
 
 /*@-type@*/
--- a/elks/arch/i86/drivers/char/serial.c
+++ b/elks/arch/i86/drivers/char/serial.c
@@ -9,8 +9,10 @@
 #include <linuxmt/ntty.h>
 #include <linuxmt/termios.h>
 #include <linuxmt/debug.h>
-
 #include <arch/io.h>
+
+#define NEW		1	/* set =0 for old driver code if needed*/
+#define FIFO		0	/* set =1 to enable new FIFO code for testing*/
 
 #if defined (CONFIG_CHAR_DEV_RS) || defined (CONFIG_CONSOLE_SERIAL)
 
@@ -95,7 +97,7 @@ static void flush_input_fifo(register struct serial_info *sp)
     } while (--i && (inb_p(sp->io + UART_LSR) & UART_LSR_DR));
 }
 
-#if 1
+#if NEW
 static int rs_probe(register struct serial_info *sp)
 {
     int status, type;
@@ -256,7 +258,7 @@ static int rs_write(struct tty *tty)
     return (int)i;
 }
 
-#if 1
+#if NEW
 void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
 {
     struct serial_info *sp;
@@ -270,10 +272,9 @@ void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
     io = sp->io;
 
     status = inb_p(io + UART_LSR);			/* check for data overrun*/
-    /* read from uart into temp buffer*/
+    /* read quickly from uart into temp buffer*/
     //do {
 	//status = inb_p(io + UART_LSR);
-	/* QEMU sometimes has interrupt w/o data available bit set*/
 	//if (status & UART_LSR_DR)			/* Receiver buffer full? */
 	    do {
 		buf[i++] = inb_p(io + UART_RX);		/* Read received data */
@@ -283,10 +284,11 @@ void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
 
     if (status & UART_LSR_OE)
 	printk("serial: data overrun\n");
-    if ((status & UART_LSR_DR) == 0)
-	printk("serial: interrupt w/o data available\n");
+    /* QEMU sometimes has interrupt w/o data available bit set*/
+    //if ((status & UART_LSR_DR) == 0)
+	//printk("serial: interrupt w/o data available\n");
 
-    /* then process received chars*/
+    /* process received chars*/
     q = &sp->tty->inq;
     for (j=0; j < i; j++) {
 	if (!tty_intcheck(sp->tty, buf[j]))
@@ -364,8 +366,10 @@ static int rs_open(struct tty *tty)
 #define UART_FCR_ENABLE_FIFO8	(UART_FCR_ENABLE_FIFO | 0x80)
 #define UART_FCR_ENABLE_FIFO4	(UART_FCR_ENABLE_FIFO | 0x40)
     /* enable FIFO with 8 byte trigger */
-    //if ((port->flags & SERF_TYPE) > ST_16450)
-	//outb_p(UART_FCR_ENABLE_FIFO8, port->io + UART_FCR);
+#if FIFO
+    if ((port->flags & SERF_TYPE) > ST_16450)
+	outb_p(UART_FCR_ENABLE_FIFO8, port->io + UART_FCR);
+#endif
 
     inb_p(port->io + UART_IIR);
     inb_p(port->io + UART_MSR);
--- a/elks/arch/i86/drivers/char/serial.c
+++ b/elks/arch/i86/drivers/char/serial.c
@@ -30,6 +30,8 @@ struct serial_info {
 #define ST_8250		0
 #define ST_16450	1
 #define ST_16550	2
+#define ST_16550A	3
+#define ST_16750	4
 #define ST_UNKNOWN	15
 
 };
@@ -92,6 +94,55 @@ static void flush_input_fifo(register struct serial_info *sp)
 	inb_p(sp->io + UART_RX);
     } while (--i && (inb_p(sp->io + UART_LSR) & UART_LSR_DR));
 }
+
+#if 1
+static int rs_probe(register struct serial_info *sp)
+{
+    int status, type;
+    unsigned char scratch;
+
+    inb(sp->io + UART_IER);
+    outb_p(0, sp->io + UART_IER);
+    scratch = inb_p(sp->io + UART_IER);
+    outb_p(scratch, sp->io + UART_IER);
+    if (scratch)
+	return -1;
+
+    /* try to enable 64 byte FIFO and max trigger*/
+    outb_p(0xE7, sp->io + UART_FCR);
+
+    /* then read FIFO status*/
+    status = inb_p(sp->io + UART_IIR);
+    if (status & 0x40) {
+	if (status & 0x80) {		/* FIFO enabled*/
+	    if (status & 0x20)		/* 64 byte FIFO enabled*/
+		type = ST_16750;
+	    else type = ST_16550A;
+	} else type = ST_16550;
+    } else {
+	/* no FIFO, try writing arbitrary value to scratch reg*/
+	outb_p(0x2A, sp->io + UART_SCR);
+	if (inb_p(sp->io + UART_SCR) == 0x2a)
+	    type = ST_16450;
+	else type = ST_8250;
+    }
+    sp->flags = SERF_EXIST | type;
+
+    /*
+     *      Reset the chip
+     */
+    outb_p(0x00, sp->io + UART_MCR);
+
+    /* FIFO off, clear RX and TX FIFOs */
+    outb_p(UART_FCR_CLEAR_RCVR | UART_FCR_CLEAR_XMIT, sp->io + UART_FCR);
+
+    /* clear RX register */
+    flush_input_fifo(sp);
+
+    return 0;
+}
+
+#else
 
 static int rs_probe(register struct serial_info *sp)
 {
@@ -159,6 +210,7 @@ static int rs_probe(register struct serial_info *sp)
 
     return 0;
 }
+#endif /* old rs_probe function*/
 
 static void update_port(register struct serial_info *port)
 {
@@ -170,6 +222,8 @@ static void update_port(register struct serial_info *port)
     if (cflags & CBAUDEX)
 	cflags = B38400 + (cflags & 03);
     divisor = divisors[cflags];
+
+    //FIXME: update lcr from parity and word termios values
 
     clr_irq();
 
@@ -227,8 +281,8 @@ void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
 	//}
     //} while (!(inb_p(io + UART_IIR) & UART_IIR_NO_INT) && i < 4);
 
-    if (status & UART_LSR_FE)
-	printk("serial: framing error\n");
+    if (status & UART_LSR_OE)
+	printk("serial: data overrun\n");
     if ((status & UART_LSR_DR) == 0)
 	printk("serial: interrupt w/o data available\n");
 
@@ -276,7 +330,7 @@ void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
 #endif
     } while (!(inb_p(sp->io + UART_IIR) & UART_IIR_NO_INT));
 }
-#endif
+#endif /* old rs_irq function*/
 
 static void rs_release(struct tty *tty)
 {
@@ -306,6 +360,9 @@ static int rs_open(struct tty *tty)
     /* Flush input fifo */
     flush_input_fifo(port);
 
+    /* enable FIFO with 1 byte trigger */
+    //outb_p(UART_FCR_ENABLE_FIFO, port->io + UART_FCR);
+
     inb_p(port->io + UART_IIR);
     inb_p(port->io + UART_MSR);
 
@@ -330,12 +387,11 @@ static int set_serial_info(struct serial_info *info,
 			   struct serial_info *new_info)
 {
     register struct tty *tty;
-    register char *errp;
+    int err;
 
     tty = info->tty;
-    errp = (char *) verified_memcpy_fromfs(info, new_info,
-					   sizeof(struct serial_info));
-    if (!errp) {
+    err = verified_memcpy_fromfs(info, new_info, sizeof(struct serial_info));
+    if (!err) {
 	/* shutdown serial port and restart UART with new settings */
 
 	/* witty cheat :) - either we do this (and waste some DS) or duplicate
@@ -343,9 +399,9 @@ static int set_serial_info(struct serial_info *info,
 	 */
 	info->tty = tty;
 	rs_release(tty);
-	errp = (char *) rs_open(tty);
+	err = rs_open(tty);
     }
-    return (int) errp;
+    return err;
 }
 
 static int get_serial_info(struct serial_info *info,
@@ -362,11 +418,12 @@ static int rs_ioctl(struct tty *tty, int cmd, char *arg)
     /* few sanity checks should be here */
     debug2("rs_ioctl: sp = %d, cmd = %d\n", tty->minor - RS_MINOR_OFFSET, cmd);
     switch (cmd) {
-	/* Unlike Linux we use verified_memcpy*fs() which calls verify_area() for us */
     case TCSETS:
     case TCSETSW:
-    case TCSETSF:		/* For information, return value is ignored */
-	update_port(port);
+    case TCSETSF:
+	/* verified_memcpy*fs() already called by ntty.c ioctl handler*/
+	//FIXME: update_port() only sets baud rate from termios, not parity or wordlen*/
+	update_port(port);	/* ignored return value*/
 	break;
     case TIOCSSERIAL:
 	retval = set_serial_info(port, (struct serial_info *)arg);
@@ -391,6 +448,8 @@ int rs_init(void)
 	"n 8250",
 	" 16450",
 	" 16550",
+	" 16550A",
+	" 16750",
 	" UNKNOWN",
     };
 
@@ -398,9 +457,6 @@ int rs_init(void)
 	if (!rs_probe(sp) && !request_irq(sp->irq, rs_irq, NULL)) {
 	    sp->tty = tty;
 	    update_port(sp);
-#if 0
-	    outb_p(? ? ? ?, sp->io + UART_MCR);
-#endif
 	}
 	tty++;
     } while (++sp < &ports[NR_SERIAL]);
--- a/elks/arch/i86/drivers/char/serial.c
+++ b/elks/arch/i86/drivers/char/serial.c
@@ -263,7 +263,7 @@ void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
     struct ch_queue *q;
     int i, j, status;
     char *io;
-    unsigned char buf[4];
+    unsigned char buf[10];
 
     i = 0;
     sp = &ports[(int)irq_port[irq - 2]];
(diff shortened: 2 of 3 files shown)


ghaerr:
I read the UART data sheet and revised the receive character interrupt routine further. This has stopped any dropped characters on my testing.

Although still using an emulator, with this revision I can't get any dropped characters.
If this could be checked on real hardware, then I will commit it.

Hello @pawosm-arm, if you're still out there, it would be nice to see whether this fixes the serial data loss described in your issue #454.

Mellvik:
@ghaerr,
I have tested the new serial.c on real hw @ 9600 ans 19200 bps.
Testing by pasting a 10,2k text file into the terminal window, cat > /tmp/xx at the receiving end, echo on - so there is both rcv and snd interrupts.

There is no significant difference between the old and the new code. @9600bps about 9.6K of 104 gets saved, at 19200 about 6.9k. Further, when using the new code, every other test has a different (and repeatable) pattern: After pasting, there is no echo for 6-10 seconds, then a few 100 bytes and we’re done. So - @ every other test it seems like the system stops or get overloaded for quite some time and then recovers.

Possibly even more interesting, when I turn off echo (stty -echo; cat > /tmp/yy; stty echo), the situation gets worse, not better. Less than 1k of the 10.4k gets into the file.

The appearance of ‘full stop' happens interactively some times too - response just dies for 20 seconds or so, then comes back - echoing all the stuff I did in between - @ the shell.

On the good side, elvis now works on serial.

Let me know if there is something in particular you’d like me to test while I’m at it.

—Mellvik

ghaerr:
@Mellvik, thanks a lot for your testing and quick response!

> There is no significant difference between the old and the new code. @9600bps about 9.6K of 104 gets saved, at 19200 about 6.9k.

You're saying that both old and new code lose a significant amount of received data?

This would confirm that the serial port code could be improved, and that the new code doesn't solve the problem.

Do you happen to know what type of uart is in your system, and whether it has a hw fifo?


>  Further, when using the new code, every other test has a different (and repeatable) pattern: After pasting, there is no echo for 6-10 seconds, then a few 100 bytes and we’re done.

The new code saves received characters into a small buffer during the interrupt routine, then processes them all at once, although still during the interrupt routine. It sounds like the system could be overloaded just receiving the characters. I am thinking that a deeper rewrite may be needed - one that only puts the received characters into a ring buffer at interrupt time, and handles everything else outside it.

>  So - @ every other test it seems like the system stops or get overloaded for quite some time and then recovers.

Very strange with "every other test."

>  Possibly even more interesting, when I turn off echo (stty -echo; cat > /tmp/yy; stty echo), the situation gets worse, not better. Less than 1k of the 10.4k gets into the file. 

This should be tested with the old driver too to see whether it is the new driver or something else responsible for the worsening.

> The appearance of ‘full stop' happens interactively some times too - response just dies for 20 seconds or so, then comes back - echoing all the stuff I did in between - @ the shell. On the good side, elvis now works on serial.

The elvis fix occurred in the previous PR. I need to know that nothing done there might cause this 'full stop' you're talking about.

At this point, the new serial driver fails to improve the system, and possibly hinders it further.
I think the best option is to not commit the driver change, even though it appeared better using QEMU.

> Let me know if there is something in particular you’d like me to test while I’m at it.

Something very useful would be to test and find a lower baud rate that completely works with serial testing. In this way, one can start with something working, rather than possibly testing better handling of error conditions as it appears is happening now, since neither driver passes testing at 9600 baud.

Once a baud rate is found that can be completely handled by ELKS (using echo on or off), it would be interesting to try to the test at a higher baud rate with the new driver to see whether that helps or not.

Thanks a lot!

Mellvik:
@ghaerr,
Yes, there's some really strange behaviour here. And it still bugs me that you managed to loos characters when typing while my testing indicates that the system can easily handle 1K bytes @ 9600 before loosing anything, and many Ks w/o echo (more results below). Is it possible that there is something else influencing the results? FWIW my config has no network, no VGA, no ACPI,  very simple.

> 11. apr. 2020 kl. 15:40 skrev Gregory Haerr <notifications@github.com>:
> 
> ﻿
> @Mellvik, thanks a lot for your testing and quick response!
> 
> There is no significant difference between the old and the new code. @9600bps about 9.6K of 104 gets saved, at 19200 about 6.9k.
> 
> You're saying that both old and new code lose a significant amount of received data?
> 
Yes, and sorry about the typo above. W/echo on, we lose about 6% @9600, 30% @ 19200 - old AND new code. Now, here's some new data. Using the new code, turning off echo makes things worse. Using the old code, turning off echo makes a big difference. @9600 I lose 47bytes out of 10.2K, @19200 I lose 273 bytes. We can almost claim 9600 is OK. I know, there is no such thing as OK when it's not, and I will test 4800 an 38400 too).
> This would confirm that the serial port code could be improved, and that the new code doesn't solve the problem.
> 
> Do you happen to know what type of uart is in your system, and whether it has a hw fifo?
> 
Yes, this an AT compatible and the UART is the 16550 or compatible. So the FIFO is there, is it being used by the driver?
> Further, when using the new code, every other test has a different (and repeatable) pattern: After pasting, there is no echo for 6-10 seconds, then a few 100 bytes and we’re done.
> 
> The new code saves received characters into a small buffer during the interrupt routine, then processes them all at once, although still during the interrupt routine. It sounds like the system could be overloaded just receiving the characters. I am thinking that a deeper rewrite may be needed - one that only puts the received characters into a ring buffer at interrupt time, and handles everything else outside it.
> 
> So - @ every other test it seems like the system stops or get overloaded for quite some time and then recovers.
> 
> Very strange with "every other test."
> 
Indeed, there is an instability somewhere, and from the looks it is interrupt related.
> Possibly even more interesting, when I turn off echo (stty -echo; cat > /tmp/yy; stty echo), the situation gets worse, not better. Less than 1k of the 10.4k gets into the file.
> 
> This should be tested with the old driver too to see whether it is the new driver or something else responsible for the worsening.
> 
> The appearance of ‘full stop' happens interactively some times too - response just dies for 20 seconds or so, then comes back - echoing all the stuff I did in between - @ the shell. On the good side, elvis now works on serial.
> 
> The elvis fix occurred in the previous PR. I need to know that nothing done there might cause this 'full stop' you're talking about.
> 
Sorry about bringing elvis into the picture. Elvis has nothing to do with this. Elvis just works. 
> At this point, the new serial driver fails to improve the system, and possibly hinders it further.
> I think the best option is to not commit the driver change, even though it appeared better using QEMU.
> 
> Let me know if there is something in particular you’d like me to test while I’m at it.
> 
> Something very useful would be to test and find a lower baud rate that completely works with serial testing. In this way, one can start with something working, rather than possibly testing better handling of error conditions as it appears is happening now, since neither driver passes testing at 9600 baud.
> 
I don't think 9600 is achieveable on a multitasking 286 w/o using FIFO in the UART. Elks has very little going on in the background, but disk read and write is really slow and presumably consumes significant 'Interrupt time'.
Question - I planned to look this up in the source, but here we go: does the driver implement HW and/or handshake? If it does, I'd like to put up two machines back to back and have them throw bytes at each other...
> Once a baud rate is found that can be completely handled by ELKS (using echo on or off), it would be interesting to try to the test at a higher baud rate with the new driver to see whether that helps or not.
> 
> Thanks a lot!
> 
> 
> 
-Mellvik

9600:
Hellooooo, you called?

pawosm-arm:
> Hello @pawosm-arm, if you're still out there, it would be nice to see whether this fixes the serial data loss described in your issue #454.

I'm sorry I couldn't engage in ELKS recently and my old Amstrad PC wasn't booted for a while. Yet after that break I'm observing a lot of positive changes in ELKS both in `menuconfig` and in quality of the startup messages. I'm not sure if it was displayed before, there's `ttyS0` related line printed at boot:
```
ttyS0 at 0x3f8, irq 4 is an 8250
```

I've enabled serial console in /etc/inittab at speed 9600bps and indeed it's usable now. Then I reverted it to the usual BIOS console and tried to run nxlandmine (with Microsoft mouse connected to the `/dev/ttyS0`), and unfortunately, it still hangs after a while of moving a mouse.

Also there's some regression in `miniterm`, it can't open serial port device as it prints:
```
/dev/ttyS0: Invalid argument
```

ghaerr:
Hello @9600, someone mentioned a baud rate @9600, sorry about that!

Hello @Mellvik,

Until your testing finds a baud rate where there is no data loss, we can't know whether the driver is inherently buggy, or instead that a certain machine hardware can't keep up with the speed. Currently, I am thinking the former, that there is an inherent bug in the driver, which causes data loss on mice at 1200 baud (#454), or just typing (on QEMU). So lets not worry now about the exact results of a test that shows data loss - instead lower the baud rate and try again. A proper test will allow a large file to be received without any data loss whatsoever.

It is quite complicated how interrupt processing can affect ELKS kernel and programs. There can be an interrupt while doing nothing, while in the kernel processing a request, and an interrupt while processing an interrupt. Each of these has different ways that the scheduler ends up allocating the CPU, and moving the received bytes to the program waiting in `read`.

> So the FIFO is there, is it being used by the driver?

No, it is not. This can be added once the regular received serial port data input is fixed or better understood.

> Sorry about bringing elvis into the picture. Elvis has nothing to do with this. Elvis just works.

I mention this because there is a possibility that my previous commit, which fixed several bugs in the serial driver, but also fixed init and vi, could have caused problems. But I don't think so at this point. I presume everything is running very well on serial from that commit, that the only problems are the testing you're doing now?

> does the driver implement HW and/or handshake?

No, not yet. This is another enhancement that could be added after we get to the bottom of the basics.

ghaerr:
Hello @pawosm-arm, thanks for your response!

> ttyS0 at 0x3f8, irq 4 is an 8250

Yes, that was always there. It's nice to be able to see it :) The startup messages were revamped so that all messages should be visible for all boots, allowing more information for users.

> I've enabled serial console in /etc/inittab at speed 9600bps and indeed it's usable now. 

Good news. @Mellvik has also confirmed that, so we're making progress.

> nxlandmine (with Microsoft mouse connected to the /dev/ttyS0), and unfortunately, it still hangs after a while of moving a mouse.

Thanks, I suspected that would happen. I think this shows a major serial driver problem remains. Do you know if the baud rate for the mouse is 1200 or 9600?

It would be interesting to try using the revised serial driver in this commit to see whether that changes the mouse behavior or not.

> Also there's some regression in miniterm, it can't open serial port device as it prints:
/dev/ttyS0: Invalid argument

Hmmm, I'll look into fixing that.

pawosm-arm:
> I suspected that would happen. I think this shows a major serial driver problem remains. Do you know if the baud rate for the mouse is 1200 or 9600?

I've used `stty` to set baud rate, e.g.:
```
stty 1200 </dev/ttyS0
```
...no difference in nxlandmine's behaviour observed though.

> It would be interesting to try using the revised serial driver in this commit to see whether that changes the mouse behavior or not.

I've tried `mou_ser-new.c` and it definitely wants to print something (a lot of 'it' I would say) on the crash. I've tried to redirect `stdout` and `strerr` to a file (namely, doing `nxlandmine 2>&1 >landmine.txt`), yet this file does not survive subsequent reboot... Redirecting console to a serial port would be helpful, unfortunately, the only serial port Amstrad PC has is already occupied by Microsoft mouse.

ghaerr:
> ...no difference in nxlandmine's behaviour observed though.

Thanks for testing. I think we should use the `mouse` testing app I added some time ago, rather than a graphical app, for testing. We can then see more clearly when there is data loss. It appears that lowering the baud rate doesn't help, not sure yet.

> I've tried mou_ser-new.c 

I've looked at that, and it should not be used, it just introduces more variables to this problem.

I have another version of the serial interrupt routine that prints 'Framing error' (buffer overrun), and 'interrupt w/o receive data available' to the console that will help debug this.

ghaerr:
Hi @Mellvik and @pawosm-arm,

I've added an update to display a console message when the serial driver chip gets a data overrun or interrupt w/o data (happens on QEMU). This should help identify whether the chip is unable to keep up with the data stream, or something else. 

Thanks again for both your testing!

pawosm-arm:
> I think we should use the mouse testing app I added some time ago, rather than a graphical app, for testing. We can then see more clearly when there is data loss. It appears that lowering the baud rate doesn't help, not sure yet.

I've tried this `mouse` testing app, it prints x, y and button states. It lags terribly whenever I move the mouse slightly faster. Eventually it dies to the point power down is required. No diagnostic message is printed though, it just dies :(

> Hmmm, I'll look into fixing that.

As I can see, `miniterm` is fixed on `master`. Can you rebase your PR?

Mellvik:
Will test in the morning. In the meanwhile there is ntty.c (probably) bu that may affect mouse usage. Stty -echo doesn't turn off echo for newlines.

-M

ghaerr:
> As I can see, miniterm is fixed on master. Can you rebase your PR?

I just left traveling, it might be easier to just comment out the two `ioctl` calls from lines 436 to 458 in misc/miniterm.c, thanks. The ioctl's aren't implemented, and return -1, causing miniterm to exit.

Mellvik:
Sorry, my misunderstanding — I hadn't noticed echonl being on. I'm proposing that setting being changed in getty.c.


—Mellvik

Mellvik:
@ghaerr,

I added your latest serial.c for testing this morning.

- There are no printk messages on the console.
- The results are repeatable - all tests are pasting a 10.2K text file into the terminal window (into the serial line):
Line speed is 19200:
with echo on, there is no response (echo) for about 6 seconds, then the first 513 bytes of the file are echoed and the same 513 bytes make it into the file, the rest is lost.
with echo off, I most of the time get exactly the same: 513 first bytes of the file, sometimes I get about 9960 bytes.
Line speed reduced to 2400 baud:
Like before, when echo is on, there is nothing for 6-8 seconds, then the content starts echoing - the 512 first bytes the file, then varying character losses [the file is xxd output so it's easy to see where the deviations are]. 7180 bytes make it into the file.
Echo off, and 8088 bytes make it into the file.

I guess the important observations are:

- no printk messages
- The line speed has minimal effect on character loss (you were right, @ghaerr)
- Something happens after the first 512 bytes of input
- And something happens immediately, since it takes so long for the first input to echo.
- The previous serial.c does not show some of the symptoms (like reported before), and generally is more responsive and has less character loss, but the main symptom is the same: Character loss regardless of line speed.

The observations are consistent with the reported serial mouse problems - char-loss at any speed.

@ghaerr, Let me know if you'd like a recording of the output in real time.


—Mellvik

ghaerr:
Hello @Mellvik and @pawosm-arm , thank you!

As a result of your testing, I am almost sure I know what is going on:

> no printk messages

Good news. The serial driver isn't getting framing errors, nor bad interrupts (unlike QEMU).
This means that the (new) interrupt routine is able to handle incoming data properly.

> The line speed has minimal effect on character loss (you were right, @ghaerr)

This means that CPU/ELKS can't keep up with sustained input data even at 1200 baud.

> Something happens after the first 512 bytes of input

Yep. The TTY drivers all work using a fixed-size input ring buffer of 512 bytes. When this buffer is full, the kernel flat drops additional characters. I am pretty sure that the kernel is spending all its time successfully adding received bytes in the interrupt processing routine, and there is no extra time for the application ("cat") to run before the ring buffer is full, then the kernel drops characters with no error given. All but @pawosm-arm's test results fit this, varying only by the detail of the precise scheduling timing which results in different amounts of total characters received.

I think of three solutions to this:
- Allow program-settable ring queue buffer size (>= to max packet size). This may be a requirement for mouse-driven programs, to allow a much larger buffer for typical mouse movements.
- Software flow control (this hasn't been discussed but is very effective for certain non-binary data).
- Hardware flow control (requires special cables or it won't work).

Because there have been no console messages on anyone's testing, I feel confident that the revised serial driver is working for all it can do, which is to successfully receive characters without data overrun (and works on real hardware and QEMU). However - there are still unexplained "crashes" detailed by @pawosm-arm, but its not certain whether these are kernel-related, or just application hangs due to data loss.

So - IMO the discussion now needs to center around what we want to do with the solutions, given that ELKS seemingly cannot keep up with constant 1200 baud.

pawosm-arm:
> Software flow control (this hasn't been discussed but is very effective for certain non-binary data).

You mean XON/XOFF? Effectively requires support on both sides, which is not always available, it's solving one problem by introducing new one...

> Hardware flow control (requires special cables or it won't work).

I'd definitely opt for that. Also AFAIR it can be detected allowing fallback code to be used with 3wire cables.

> However - there are still unexplained "crashes" detailed by @pawosm-arm, but its not certain whether these are kernel-related, or just application hangs due to data loss.

Not seeing any printable manifestation of the root cause, it's better to eliminate problems that we already aware of (which is serial communication layer).

ghaerr:
> You mean XON/XOFF? Effectively requires support on both sides, which is not always available, it's solving one problem by introducing new one...

Definitely requires both sides, but is sometimes only option when HW flow control not possible.


> Also AFAIR it can be detected allowing fallback code to be used with 3wire cables.

I am not aware of any way to make HW flow control work without DTR/CTS wires being available between systems.


I just checked SLIP in `ktcp` and it uses 256 byte packets - so flow control should not be required.

A question remains - what actual applications need to run, given ELKS can't handle sustained serial input?


> Not seeing any printable manifestation of the root cause, it's better to eliminate problems that we already aware of (which is serial communication layer).

Yes. I am not sure whether we have all root causes for this, especially with the strange mouse and other hangs. I am pretty sure we've just identified the 512-char ring buffer character drop problem.

Mellvik:
@ghaerr,

a couple of comments:
1) the serial driver before the current one was a lot better in terms of performance (i.e. could take more data before dropping and always echoing without the multisecond hangs).  [reported before]
2) I think the optimal sequence of improvement now is
	1) FIFO suport (BTW - my systems have a 16550, ELKS reports 16540 with no FIFO, so chip detection may be an issue too)
	2) dependable Xon/Xoff
	3) hardware handshake (can possibly be pulled out of Minix 2 rs232.c)

—Mellvik

ghaerr:
> 1) the serial driver before the current one was a lot better in terms of performance (i.e. could take more data before dropping and always echoing without the multisecond hangs).

That's because it called 'wake_up' after every character received, which caused the scheduler to run programs a bit earlier, but wasted CPU time doing that for every character, in the interrupt routine. 

I can put the framing error detect code in the old routine if you like. When FIFO's are implemented, we will definitely need the newer version which reads as many characters as possible during the interrupt routine.

> 2) I think the optimal sequence of improvement now is 1) FIFO support 

16 character FIFO support may not solve our problem of ELKS not being able to keep up for 512+ byte sequences. However, it can reduce the interrupt overhead, which could possibly be enough time to stop 512 byte queue overruns, but not likely.

I will turn FIFO on in the new driver and we can test.

> (BTW - my systems have a 16550, ELKS reports 16540 with no FIFO, so chip detection may be an issue too) 

I have found a good article on UART programming which is quite informative (https://en.wikibooks.org/wiki/Serial_Programming/8250_UART_Programming). It has a different algorithm for UART recognition. I will change the UART probe to that, since you're saying the current version misrecognizes your chip.

2) dependable Xon/Xoff 3) hardware handshake 

What would be used to test any implementation?

ghaerr:
Hi @pawosm-arm,

> As I can see, miniterm is fixed on master. Can you rebase your PR?

I'm not a git expert... what do you need me to do for you?

ghaerr:
Hello @Mellvik and @pawosm-arm,

I pushed another serial driver update. This one has a rewritten UART detection routine which hopefully fixes @Mellvik's system incorrectly reporting a 16450 instead of 16550. It also recognizes 16550A and 16750 chips.

While reviewing the driver, I realized the previous version was displaying framing errors on the console (serial words not framed by start/stop bits correctly), rather than overrun errors (new data received by the UART prior to the CPU reading the prior data). It would be interesting to run the tests again as my previous conclusion could be incorrect, I'm sorry.

The FIFO can now be enabled by uncommenting line 364 in serial.c. This might help improve data overrun errors if we are receiving them. I don't think it will fix being able to keep up with fast data streams, but it could help.

Mellvik:
@ghaerr,
thanks - we're getting there.
The new chip detection code is better, if possibly not perfect. My machines turn out to have 16540 chips, not 16550 like I thought, so I added an ISA card with two 16550 chips, now a total of 3 serial lines on the machine.

The old driver reported S0 as 16540 (correct), S1 as 16540 (incorrect), S2 as 16550 (correct). Interesting behaviour.
Your new revision reports 1 16540 and 2 16550A, which is OK. I don't know what the differences are between the 16550 and 16550A are, but I figure we'll just leave it there for now. 

Since here is a new twist:
After adding more serial interfaces, there is no more input. I can send data from ELKS to all ports correctly, and I can see on the line tester that data is coming in on the serial lines, but nothing passes the driver. Not even on ttyS0 which used to work OK.

Also, before getting into the multiline testing, I bombed the serial line S0 from the outside @ 19200 using the modified driver and got tons of overruns from printk. 

Testing continues.

—Mellvik

ghaerr:
Good news on the new chip recognition algorithm.

I'd like to defer any multi-line testing for now, and recommend that you remove the second ISA card during testing, so that we can concentrate on getting the basics working, before looking at even more problems with multiple serial lines.

> I bombed the serial line S0 from the outside @ 19200 using the modified driver and got tons of overruns from printk.

That's to be expected, what about 1200 baud?

ghaerr:
Pushed another optional (commented out) capability to enable the FIFO with an 8-byte trigger. I hadn't realized that the FIFO will timeout and produce an interrupt with less than 8 bytes in the FIFO, after 4 character timings. This could possibly solve the problem of data overrun.

When the FIFO is manually enabled in rs_open and recompiled, there will only be 1 interrupt per 8 characters, which could help ELKS ability to keep up with fast data streams and mouse action.
Unfortunately, leaving the FIFO enabled, even for 16550's and above only, makes QEMU unusable for typing on the serial console. Not sure why that is yet.

I'm thinking for the longer term that setting stty VMIN value > 1 may be appropriate to set the FIFO trigger size, depending on whether having the 8 byte FIFO setting shows an observable delay in users typing on the serial line.

Mellvik:
Any idea why multiple serial lines breaks input? I can't test this until we figure that out since I have no fifo on ttyS0.

-M

ghaerr:
> Any idea why multiple serial lines breaks input? 

Only guessing that the IRQ on them overlap perhaps. COM1 should be 0x3f8 IRQ 4,
COM2 0x2f8 IRQ 3. ELKS COM3 is expecting 0x3E8 IRQ 5. All ports and IRQs need to be separate, see if they are configurable. If not, you can set them to your boarding 'serial_info' struct in serial.c.

ghaerr:
IIRC, IRQ 5 might be used for something else in ELKS.

Transmit doesn't use an IRQ so it looks like the port base addresses are correct. Only receive uses interrupt processing at this point.

Mellvik:
Irqs are fine, and reported as such during boot (i guess they wouldn't if not).

-m

ghaerr:
> Irqs are fine, and reported as such during boot (i guess they wouldn't if not).

No, the 'report' doesn't test the IRQ, it only lists the IRQ expected in its setup table. It doesn't test IRQ. So I think that may still be your problem.

Mellvik:
Ok, if so, does that explain that input is gone from all lines?

-M

ghaerr:
If the IRQs aren't working or are crossed over, etc, then no input will be received, as all input is interrupt driven. OTOH, all output is not interrupt driven.

I am guessing your extra card interrupt lines are goofed up or not working, especially since plugging it in breaks COM1.

ghaerr:
@Mellvik and @pawosm-arm,

I've added 'NEW' (default ON) and 'FIFO' (default OFF) compilation options to the top of 'elks/arch/i86/drivers/char/serial.c' that control whether the new or old serial driver code is compiled, as well as enabling the hardware fifo, if present.

I am working on serial driver enhancements for processing VTIME which is required to get `vi` running using the full terminal height on serial port, and need the `ioctl` and QEMU fixes in the serial driver, so I'm committing this now. Discussion on serial i/o testing and issues can continue on the open issue #515. If for some reason the new serial driver does not work well, the old driver can be enabled by setting '\#define NEW 0" in serial.c. The new code will be required for any FIFO operation anyways, so we might as well keep moving forward.

I remain interested in testing comments with the FIFO option enabled. I think this should help with mouse support as well.

Mellvik:
Good work, @ghaerr, and I agree, we need to keep moving. 

I'm continuing serial testing, there are issues in serial.c and lrqtab.S (of course...), three serial lines are now operating. I'm collecting issues for a summary later this week.


--Mellvik

pawosm-arm:
@ghaerr I had 2 days break, will look at this possibly tomorrow.

pawosm-arm:
@ghaerr I've tested what's on master now with `miniterm` and this time I've tested it at speed 115200. Sending to Linux was OK, yet receiving (especially, longer text copy-pasted at Linux side) resulted very quickly in "serial: buffer overrun" messages. Similarly with `ktcp`. Strangely, `mouse` test app still isn't printing any error message when it hangs.

ghaerr:
> resulted very quickly in "serial: buffer overrun" messages.

This means that the interrupt routine can't process input data before another character arrives. Using the hardware FIFO, if you have a 16650 on your machine, should help this at lower baud rates. We need to find a baud rate and serial driver combo that works, before trying to push the limits with high speeds at first.

I merged #553 into master, which allows transferring larger amounts of received data per read.
You might try enabling the hardware FIFO in serial.c using \#define FIFO 1 at the top of the file, and lowering the baud rate to see when the serial overrun messages stop.

> mouse test app still isn't printing any error message when it hangs.

Thanks. It's running at much lower baud rate, so less surprise no serial overrun messages. It seems we still haven't found the real problem for the hang. You might try setting the VMIN value from 0 to 1 in mouse.c to see if that changes anything. That will stop the constant polling otherwise used.

pawosm-arm:
I've introduced those two changes:
```
diff --git a/elks/arch/i86/drivers/char/serial.c b/elks/arch/i86/drivers/char/serial.c
index abfdc6e9..b71175b9 100644
--- a/elks/arch/i86/drivers/char/serial.c
+++ b/elks/arch/i86/drivers/char/serial.c
@@ -12,7 +12,7 @@
 #include <arch/io.h>
 
 #define NEW            1       /* set =0 for old driver code if needed*/
-#define FIFO           0       /* set =1 to enable new FIFO code for testing*/
+#define FIFO           1       /* set =1 to enable new FIFO code for testing*/
 
 #if defined (CONFIG_CHAR_DEV_RS) || defined (CONFIG_CONSOLE_SERIAL)
 
diff --git a/elkscmd/sys_utils/mouse.c b/elkscmd/sys_utils/mouse.c
index 8ab0a234..11069df7 100644
--- a/elkscmd/sys_utils/mouse.c
+++ b/elkscmd/sys_utils/mouse.c
@@ -116,7 +116,7 @@ open_mouse(void)
        termios.c_iflag &= ~(ICRNL | INPCK | ISTRIP | IXON | BRKINT | IGNBRK);
        termios.c_cflag &= ~(CSIZE | PARENB);
        termios.c_cflag |= CS8;
-       termios.c_cc[VMIN] = 0;
+       termios.c_cc[VMIN] = 1;
        termios.c_cc[VTIME] = 0;
        if(tcsetattr(mouse_fd, TCSAFLUSH, &termios) < 0) {
                printf("Can't set termio on %s, error %d\n", MOUSE_DEVICE, errno);
```
Unfortunately, I'm not observing much on the difference in either of `miniterm`, `ktcp` and `mouse`. Actually, `mouse` hung much earlier than usual, yet still with no error message.

ghaerr:
> Unfortunately, I'm not observing much on the difference in either of miniterm, ktcp and mouse. 

Have you tried running miniterm and ktcp with a much lower baud rate, say 1200?

Will ^C abort the mouse hang?

Sounds like the hardware reliability problem with the mouse action is in a different area than being worked on.

pawosm-arm:
> Have you tried running miniterm and ktcp with a much lower baud rate, say 1200?

Yes, `miniterm` works more or less fine at lower speeds. And even at higher speed it does not hang, even after printing buffer overrun error message. With `ktcp` it's more complicated. It lacks command line parameters that `slattach` has (e.g. baud rate, hw control, protocol: slip, cslip, adaptive), so I don't even know if both sides talk the same.

> Will ^C abort the mouse hang?

It doesn't respond to ^C when it hung, and if you don't press ctrl+alt+del immediately, hardware power down is needed.

PR #536, opened 2020-04-10
Add line editing and command history to ash.

georgp24:
I adapted the "Linenoise" line editing code to ELKS and now you can use ash with line editing and command history.

### DIFF
--- a/Documentation/text/shell.txt
+++ b/Documentation/text/shell.txt
@@ -1,0 +1,52 @@
+ELKS uses the ash or Almquist shell. This is the port for ELKS which was ported 
+from the MINIX version in 1998.
+
+This had commented code to include the GNU Readline library or the editline 
+library (by Simmule Turner and Rich Salz). 
+
+These libraries provide a set of functions that allow users to edit command 
+lines as they are typed in. They also include additional functions to maintain 
+a list of previously-entered command lines, to recall and perhaps reedit those 
+lines. Both libraries were not used with ELKS, probably because of memory space
+requirements.
+
+I now added these functions using the small Linenoise library by Salvatore 
+Sanfilippo from Sicily. It is available on GitHub at 
+https://github.com/antirez/linenoise . This library was reduced in size and
+adapted to ELKS and its ash port.
+
+This library adds the following commands to allow for line editing and command
+history:
+
+Cursor-Up
+    Search backward in the history.
+Cursor-Down
+    Search forward in the history.
+
+Cursor-Left / Control-b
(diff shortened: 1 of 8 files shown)


ghaerr:
Hello @georgp24,

Wow, nicely done! Nice how it was attached with minimal interference to the rest of the shell.
I think 'linenoise_elks.c' should '#include "shell.h"' and then have an '#if LINENOISE' around the entire file so that code is not included if it is turned off in shell.h.

Have you tested this much? How much more code does it add to `sh`?

georgp24:
Seems your major concern is to turn linediting off :-) There were 20 lines of code added to input.c plus the file linenoise_elks.c. If I add your suggestions, there will be no additional lines added if you set LINENOISE to zero.
However, I think line editing should be the default mode for ash.

Yes, I tested it, otherwise I could not make the changes required to linenoise to get it to work with ELKS. But if someone else tests it, he may find problems I did not.

Mellvik:
Hi @georgp24,

This is great news, I'm looking forward to beat it up!
And like @ghaerr, I'm interested in the size of the increased footprint when LINENOISE is included.

--Mellvik

georgp24:
I added Greg's suggestion to `linenoise_elks.c` and compiled `sh` with the following results:

with `LINENOISE 1: 51.382 bytes`
with `LINENOISE 0: 44.292 bytes`

So linenoise_elks adds `7.100 bytes` to sh.
However, I currently try to enable command-line completion and this will add some bytes to that.

ghaerr:
@georgp24, thank you for this contribution and your work on it.

> I currently try to enable command-line completion

That would be a nice addition.

Is it required to have MULTILINE set on? The linenoise documentation states that this is only required for multi-line edits... is that used in shell command-line editing?

ghaerr:
@georgp24,

I just tested the the shell in the normal use (not serial, but /dev/tty1). Typing ^C kills the shell, and login: appears.

The enhancement needs a closer look at what else is executed in 'linenoise_elks.c', in order to remain turned on by default. I have just finished carefully fixing signal processing in the shells, and am currently trying to perfect 'vi' and shell operation when run from the serial line. It seems that linenoise does a bit more than just edit a line of input... we need to get a grasp on that.

georgp24:
Yes, CTRL-C terminates the shell. This is the code:
    `case CTRL_C:     /* ctrl-c */`
`                errno = EAGAIN;`
`                return -1;`

So what should replace this? Clear the editing buffer and pass CTRL-C to ash?

PR #538, opened 2020-04-11
Modify CTRL-C handling. It will no longer terminate the shell.

georgp24:
It  just clears the input buffer now.


ghaerr:
Thanks @georgp24. I want to also take a look at the serial console problem and test this before committing.

georgp24:
Linenoise checks if the connected terminal is an `ansi` type. If the serial terminal does not report that, it will be handled as an unsupported terminal. I did not test how well that works.

Also I use a few printf statements, I do not know if these are redirected to the serial terminal. I cannot test terminals connected to the serial port here. Maybe I have to replace these printf statements with write(fileno(stdout)...).

ghaerr:
Hi @georgp24, this change was added to PR #541, along with fixes for running on a serial port, and fixing whether a terminal type is supported, so I'm closing this PR.

Are you thinking of adding command completion?

PR #540, opened 2020-04-12
Minor getty-fixes, typo in elvis/test.c.

Mellvik:
Getty: /etc/issue messages was written before terminal/line speed was set.
Getty: Changed line initialization to not set `echonl`.

Fixed minor typo in elvis test.c.

### DIFF
--- a/elkscmd/elvis/test.c
+++ b/elkscmd/elvis/test.c
@@ -33,7 +33,7 @@ int fd;
 	termios.c_cc[VINTR] = 3;
 	termios.c_cc[VMIN] = 1;
 	termios.c_cc[VTIME] = 0;
-	tcsetattr(2, TCSANOW, &termios);
+	tcsetattr(fd, TCSANOW, &termios);
 }
 
 setcooked(fd)
--- a/elkscmd/sys_utils/getty.c
+++ b/elkscmd/sys_utils/getty.c
@@ -229,6 +229,26 @@ int main(int argc, char **argv)
 	debug("'%s' %ld\n", argv[1], baud);
     }
 
+    /* setup tty termios state*/
+    baud = convert_baudrate(baud);
+    if (tcgetattr(STDIN_FILENO, &termios) >= 0) {
+        termios.c_lflag |= ISIG | ICANON | ECHO | ECHOE;
+        termios.c_lflag &= ~(IEXTEN | ECHONL | ECHOK | NOFLSH);
+        termios.c_iflag |= BRKINT | ICRNL;
+        termios.c_iflag &= ~(IGNBRK | IGNPAR | PARMRK | INPCK | ISTRIP | INLCR | IGNCR
+		| IXON | IXOFF | IXANY);
+        termios.c_oflag |= OPOST | ONLCR;
+        termios.c_oflag &= ~XTABS;
+        if (baud)
+            termios.c_cflag = baud;
+        termios.c_cflag |= CS8 | CLOCAL | HUPCL;
+        /*termios.c_cflag |= CRTSCTS;*/
+        termios.c_cflag &= ~(PARENB | CREAD);
+        termios.c_cc[VMIN] = 0;
+        termios.c_cc[VTIME] = 0;
+        tcsetattr(STDIN_FILENO, TCSAFLUSH, &termios);
+    }
+
     fd = open(ISSUE, O_RDONLY);
     if (fd >= 0) {
 	put('\n');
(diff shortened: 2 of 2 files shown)


ghaerr:
Why are you proposing turning off ECHONL? IMO, it should be left on, there may not be further line changes by login/sh/etc, so the setting is important.

Is there a reason other than not running `stty -echo -echonl` for testing?

Looks like you prepared this commit from an earlier source tree, I added the new debug "else debug("tcsetattr(0) failed")" which conflicts.

Mellvik:
@ghaerr,

> 12. apr. 2020 kl. 17:06 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> Why are you proposing turning off ECHONL? IMO, it should be left on, there may not be further line changes by login/sh/etc, so the setting is important.
> 
> 
I have never seen this option enabled before (i've been doing serial lines for while). It does nothing good, except causing confusion IMHO. The only case in which it has effect is if echo is otherwise off. I believe it is a reasonable expectation that -echo means that - literally, not leaving something still being echoed, like in this case newlines.
> Is there a reason other than not running stty -echo -echonl for testing?
> 
> 
Yes, I reducing the interrupt load on the driver and the system, testing whether that makes any difference.
> Looks like you prepared this commit from an earlier source tree, I added the new debug "else debug("tcsetattr(0) failed")" which conflicts.
> 
> 
Yes I noticed. Word, I synced my fork with master this morning. I'll cancel and fix.
> —
> 
BTW - I'm continuously testing serial, and it's obvious you're right, something is rotten in there and it is not serial overrun.


—Mellvik

ghaerr:
Thanks @Mellvik, please resubmit your fix for getty's handling of /etc/issue.

At this point, since many other routines turn on and off ECHO and ECHONL together, I would prefer not to change the ECHNONL setting in getty at this point, and instead have you execute `stty -echo -echonl` for testing.

You can quickly prepare shell scripts for use in your own tree, without any affect on the master tree. For instance, put the above stty command in a file 'xx' in 'elkscmd/rootfs_template/root/xx'. That file will appear after "make" in the root's login directory on ELKS boot. Running "./xx" can be run for testing etc with little typing. The file 'xx' isn't checked in, so it stays around between 'git pull upstream master'.

ghaerr:
> At this point, since many other routines turn on and off ECHO and ECHONL together, I would prefer not to change the ECHNONL setting in getty at this point, and instead have you execute stty -echo -echonl for testing.

@Mellvik - I looked closer at the TTY code, and didn't realize that ECHONL only matters when echo is off. Thus, I am reversing my opinion on your submission, go ahead and turn ECHONL off in getty. Sorry for my confusion!

Mellvik:
@ghaerr,

> 12. apr. 2020 kl. 17:46 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> Thanks @Mellvik <https://github.com/Mellvik>, please resubmit your fix for getty's handling of /etc/issue.
> 
> 
I'm still getting 'can't automatically merge' warnings, and I'm fully in sync with origin/master. Hints?
> At this point, since many other routines turn on and off ECHO and ECHONL together, I would prefer not to change the ECHNONL setting in getty at this point, and instead have you execute stty -echo -echonl for testing.
> 
> 
No problem, but can you give examples where echonl is useful and used? If a program here and there turns it off for good measure is good programming, but does any program actually set it? As you can tell, I'm considering it utterly useless, confusing and better left off so as to not create trouble.
> You can quickly prepare shell scripts for use in your own tree, without any affect on the master tree. For instance, put the above stty command in a file 'xx' in 'elkscmd/rootfs_template/root/xx'. That file will appear after "make" in the root's login directory on ELKS boot. Running "./xx" can be run for testing etc with little typing. The file 'xx' isn't checked in, so it stays around between 'git pull upstream master'.
> 
> 
Thanks, yes I really need to do that for inittab.

BTW - on a different tangent: Elvis is setting stderr into RAW or CBREAK, which doesn't make sense to me. Makes it impossible to debug elvis. Your take?

—Mellvik

ghaerr:
> I'm still getting 'can't automatically merge' warnings, and I'm fully in sync with origin/master.

Are you sure? Make sure you've done 'git pull upstream master' followed by 'git push' to your own repository, all after a 'git checkout master'.

> on a different tangent: Elvis is setting stderr into RAW or CBREAK, which doesn't make sense to me. Makes it impossible to debug elvis. Your take?

Elvis is quite complicated, I would not suggest changing anything major. Report bugs and we can try fixing them...

Mellvik:
> 12. apr. 2020 kl. 20:52 skrev Gregory Haerr <notifications@github.com>:
> 
> ﻿
> I'm still getting 'can't automatically merge' warnings, and I'm fully in sync with origin/master.
> 
> Are you sure? Make sure you've done 'git pull upstream master' followed by 'git push' to your own repository, all after a 'git checkout master'.
> 
Thanks, will take another stab at it.
> on a different tangent: Elvis is setting stderr into RAW or CBREAK, which doesn't make sense to me. Makes it impossible to debug elvis. Your take?
> 
> Elvis is quite complicated, I would not suggest changing anything major. Report bugs and we can try fixing them...
> 
Minor change in curses.c, I'll test and propose in a PR.
> —
> 
> 
-Mellvik

PR #542, opened 2020-04-13
Add TAB completion

georgp24:
This pull request adds TAB completion or autocompletion to the shell. See the extended documentation in which situations this can be used.

### DIFF
--- a/Documentation/text/shell.txt
+++ b/Documentation/text/shell.txt
@@ -13,7 +13,7 @@ requirements.
 I now added these functions using the small Linenoise library by Salvatore 
 Sanfilippo from Sicily. It is available on GitHub at 
 https://github.com/antirez/linenoise . This library was reduced in size and
-adapted to ELKS and its ash port.
+adapted to ELKS and its ash shell port.
 
 This library adds the following commands to allow for line editing and command
 history:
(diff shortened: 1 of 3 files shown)


ghaerr:
Hi @georgp24,

This work can be added, however it will have to be modified to use the version of 'linenoise_elks.c' that I put up in PR #541, as considerable work was done getting your original contribution to work on serial lines, as well as other fixes. The changes from 'printf' to 'write' have no effect on that.

I plan on commiting #541 in the morning (for me, 8 hours from now). I would like to keep the subsequent changes to linenoise_elks.c to a minimum going forward, as it was tricky getting everything to work on serial and direct console.

Because the tab processing will likely get enhanced (different search paths, etc), it would probably be best to keep it in its own file, e.g. 'line_completion.c' so as to keep changes to the central `ash` source to a minimum. If you like, I can help you with that, since almost all the new contribution is in the completion routine and documentation. Let me know which route you'd prefer.

georgp24:
I did not see that you planned to commit changes to linenoise_elks.c. I will look at it and add the TAB completion to the latest version. I do not think it requires changes to the modifications you made.

ghaerr:
> I will look at it and add the TAB completion to the latest version.

Thanks, #541 is committed now. If you could put the command completion in a new file, say line_completion.c or autocomplete.c, instead of in 'input.c', that will allow the ongoing changes to command completion to be outside the `ash` sources for better maintainability. Also, no need to cast (char *) on `strdup` and `fopen`, this masks the error, I plan on fixing the compiler warnings for those two functions in libc, where the fixes belong.

The completion callback function registration doesn't need another variable for whether its been set, it can be registered every time (see linenoise_elks.c, its a single line function).

PR #544, opened 2020-04-13
Minor update to getty, elvis

Mellvik:
Getty: /etc/issue content was displayed before terminal initalization. Also changed ECHONL to be off by default.
Elvis: Fixed typo in test.c, changed user I/O to use stdout, other than stderr, simplifying debugging.

### DIFF
--- a/elkscmd/elvis/curses.c
+++ b/elkscmd/elvis/curses.c
@@ -16,6 +16,7 @@
 
 #include "config.h"
 #include "vi.h"
+#include <unistd.h>
 
 #if ANY_UNIX
 # if UNIXV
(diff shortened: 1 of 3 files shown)


ghaerr:
@Mellvik, thanks. What are you trying to debug in vi?

The only remaining problem I've seen running vi on a serial line is the 'O' command, where an insert line is used to open the line. It seems that the reason for the bug is that vi doesn't know the size of the emulator window, and vi is running at the top of the emulation, not the bottom, so the insert line fails to remove the 24th line on the screen. This could possibly be fixed if your emulator implements the ESC [ 6 n "get screen size" ansi function.

Mellvik:
Thank you @ghaerr.

Re. vi() using stderr for I/O:
 I'm not testing anything in vi right now, but came across this issue when trying to figure out the 'screen too small' problem a week back. Later I was diagnosing something that seemed like a raw mode problem and turned out to be netcat behaviour. In both cases, having stderr available for debugging/errors would have been beneficial. Thus the changes.


-M

PR #546, opened 2020-04-13
Add autocompletion to ash

georgp24:
Based on the latested changes by Greg in #541. See details in the shell.txt file.

### DIFF
--- a/Documentation/text/shell.txt
+++ b/Documentation/text/shell.txt
@@ -13,7 +13,7 @@ requirements.
 I now added these functions using the small Linenoise library by Salvatore 
 Sanfilippo from Sicily. It is available on GitHub at 
 https://github.com/antirez/linenoise . This library was reduced in size and
-adapted to ELKS and its ash port.
+adapted to ELKS and its ash shell port.
 
 This library adds the following commands to allow for line editing and command
 history:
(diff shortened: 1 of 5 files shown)


ghaerr:
Thanks for the changes.

The change at 935 (939) needs to return -1, not 0. This is how the shell exits when ^D is typed. Did you test this?

Also, did you test the result of ^C not entering the previous command into the history? I thought I did and the added code didn't do anything, which is why I deleted it.

georgp24:
Well, I entered CTRL-C and the characters I had entered on this line before pressing CTRL-C could be recalled with cursor up. Not so with my version.
I removed the "-1" with CTRL-D because I see no reason to terminate the shell and you complained that CTRL-C did that before. I understood that terminating the shell is always considered an error. If not, set it to -1 again.

ghaerr:
> Well, I entered CTRL-C and the characters I had entered on this line before pressing CTRL-C could be recalled with cursor up. Not so with my version.

Ok. I had also added a write \\r\\n to simulate what other shells do on ^C, but that's ok you removed that, I suppose.

> I removed the "-1" with CTRL-D because I see no reason to terminate the shell and you complained that CTRL-C did that before. I understood that terminating the shell is always considered an error. If not, set it to -1 again.

^D needs to terminate the shell (all shells do this), so please change that back and I'll commit this.

ghaerr:
@georgp24,

Bug report on autocompletion:

Typing ^C prints prompts w/o newlines. The previous write(l.ofd, "\\r\\n", 2) needs to go back in.

Typing 'ls / TAB' or 'ls /bin/ TAB' doesn't do anything... should it?
Typing 'cd /', 'ls TAB' doesn't do anything.. ?
When in /root, typing 'ls TAB' works. 
What else is supposed to expand when TAB is used?

georgp24:
I tested the following (taken from the shell.txt file):
```
Linenoise features autocompletion (or TAB completion) which has been adapted 
for ELKS.

a) if you enter one or more letters on the command line and press TAB, 
the shell will list the programs in the /bin directory that begin with 
these letters on the screen and put the first one of these on the command line.
If only one program matches the letters entered, there will be no list shown 
and just this program will be put on the command line.
You can select the program which is put on the command line by pressing enter
or enter TAB again to get the next program in the list being displayed on the
command line which you can then select with enter.

b) if you enter "./" on the command line and press TAB, the shell will list
the programs in the current directory and show the first of these on the
command line. You can either select that with enter or press TAB to put the
next program on the command line.

c) if you enter e.g. ls, cat, vi etc. plus a blank and press TAB, the shell
will list all the files in the current directory which are not executables on
the screen and put the first one on the command line. You can select that with
enter or press TAB to get the next one on the command line. This way you can
select a file to use with ls, cat or vi using autocompletion.

d) if you enter "cd " on the command line and press TAB, the shell will list
all the directories in the current directory and put the first of them on the 
command line. If you press enter you can move into that directory, or select
"cd .." from the list by pressing TAB to move up one directory. 
If you enter "cd /" and press TAB, the shell will display all the directories 
in the root directory. If you press TAB again you can select one of these.
```
It did not work properly, when I did no `make clean` before entering `make`.

PR #547, opened 2020-04-13
libmem.a, cgatext lib, /dev/cgatext

marcin-laszewski:
- Library of memory primitives (poke*, peek*, etc).
- Direct acces to CGA text mode framebuffer.
- /dev/cgatext driver.
- Demo program.


ghaerr:
Looks good, you got any screenshots?

Any reason you used 10,1 rather than 10,0 for /dev/cgatext?

In elkscmd/Makefile, if you add cgatext into SUBDIRS, it gets built (but not installed) regardless of the config setting, which helps ensure it doesn't break when not set. FYI.

marcin-laszewski:
10,1 - your are right. Minor number should starts from '0'.

Are you able to run 'cgatext' program?

- make menuconfig
- select 'cgatext' in 'Kernel/Drivers' and 'cgatext' in 'User space programs'.
- make clean && make
- Launch ELKS.
- Run 'cgatext' program.

ghaerr:
> Are you able to run 'cgatext' program?

I haven't yet, but will after this is committed. I think CGA is supposed to be emulated by QEMU, we will find out.

marcin-laszewski:
It is called "cgatext" because 80x25 color text mode was introduced with CGA.
It it suppported by EGA, VGA, SVGA, too.

ghaerr:
Thanks. 

Do you want this committed as-is, or did you want to fix /dev/cgatext minor to be 0, or anything else?

marcin-laszewski:
I have corrected this minor number of course.
Theres is a new PR.

2020-04-14 6:31 GMT+02:00, Gregory Haerr <notifications@github.com>:
> (...) did you want to fix /dev/cgatext minor to be 0 ?

PR #548, opened 2020-04-14
cgatext

marcin-laszewski:
- libmem.a: Useful memory primitives libs from kernel lib.
- cgatext.h: CGA framebuffer color texmode lib (header file only).
- Char device: /dev/cgatext
- cgatext demo.


ghaerr:
Not sure if you know this, but if you make a change on your existing local branch to say, 'Make.devices', `git commit -a` and `git push` again, GitHub will pick up that commit and add it to your already open PR. That saves you the trouble of deleting the branch and opening up another PR.

Thank you!

PR #553, opened 2020-04-14
[kernel][cmds] Enhance kernel, vi and sh on serial ports

ghaerr:
Lots of changes to allow `vi` and `sh` to work on variable-sized serial consoles (using a terminal emulator for access to ELKS /dev/ttyS0). Changing window size while `vi` is running is not supported.

- Add ANSI DSR (device status report) cursor location handler to direct console driver.
- Implement tty/serial VTIME timeout processing for received characters.
- Don't update serial divisor latch unless changed for TCSET `ioctl`.
- Enhance `vi` and `sh` (linenoise) to use DSR sequence to determine console size/width.
- Allows running serial console of any size and proper operation of `vi`, providing terminal emulator implements DSR response (ESC [ 6 n).
- Enhance linenoise to use VTIME timeout when reading DSR response.
- Enhance linenoise to only check console width once.
- Linenoise ^C operation identical to ^U (clear input line).
- Match `getty` line settings to default tty values, ECHONL default off.
- Revise QEMU defaults in serial driver. QEMU has serious serial port emulation problems and continues to drop input characters.

Please test operation of `vi` on real serial port. Hopefully running `vi` multiple times works perfectly when reading DSR response to find window size, unlike QEMU which loses character input data regularly with no overrun or other UART error bit.

### DIFF
--- a/elks/arch/i86/drivers/char/dircon.c
+++ b/elks/arch/i86/drivers/char/dircon.c
@@ -197,6 +197,20 @@ static int parm2(register unsigned char *buf)
     return parm1(buf);
 }
 
+static void itoaQueue(int i)
+{
+   unsigned char a[6];
+   unsigned char *b = a + sizeof(a) - 1;
+
+   *b = 0;
+   do {
+      *--b = '0' + (i % 10);
+      i /= 10;
+   } while (i);
+   while (*b)
+	AddQueue(*b++);
+}
+
 static void AnsiCmd(register Console * C, char c)
 {
     register unsigned char *p;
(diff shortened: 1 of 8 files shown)


ghaerr:
Rewrote tty/serial `read` routine to fill buffer with as much data as is available up to requested length, rather than returning immediately after VTIME timeout or VMIN characters. This will improve kernel -> user data throughput greatly for any serial input. 

All four combinations of VMIN and VTIME line settings are now implemented. (See http://unixwiz.net/techtips/termios-vmin-vtime.html for details).

PR #556, opened 2020-04-16
Added tty(1) command to misc_utils.

Mellvik:
The tty() command suddenly becomes important when using several serial lines.


ghaerr:
@Mellvik - Just happens we're both adding new commands in misc_utils. I'm pretty certain that as soon as I commit #554, it will cause a conflict with this, or vice versa. So I'll commit mine first, then fix yours, and push up a new PR for it, so you don't need to do anything. FYI.

ghaerr:
Also FYI, there is no "1200k" disk image size tag for Applications, so I'll move tty to "1440k".

ghaerr:
Resubmitted in #557.

PR #560, opened 2020-04-16
[init] Stop init respawns if getty exits quickly

ghaerr:
Modify `init` to stop continual 'respawn' processing of a command in /etc/inittab if process exits within 3 seconds of starting. No current way of automatically restarting, but that shouldn't be a problem.

Fixes issue mentioned in #515.

Modify /etc/inittab to défault running `getty` onto /dev/tty1 and /dev/ttyS0.
ELKS now boots fine from serial or console and allows access from serial or console with no /etc/inittab modifications.

Tested on QEMU with serial port setup and disabled.

@Mellvik, please test. I was able to fix the problem where `getty` was cycling, but am not able to see `init` hang. Perhaps they are the same thing.

### DIFF
--- a/elkscmd/rootfs_template/etc/inittab
+++ b/elkscmd/rootfs_template/etc/inittab
@@ -11,6 +11,6 @@ si::sysinit:/etc/rc.d/rc.sys
 #ud::once:/sbin/update
 
 1:2345:respawn:/bin/getty /dev/tty1
-#s0:2345:respawn:/bin/getty /dev/ttyS0 9600
+s0:2345:respawn:/bin/getty /dev/ttyS0 9600
 #2:2345:respawn:/bin/getty /dev/tty2
 #3:2345:respawn:/bin/getty /dev/tty3
--- a/elkscmd/sys_utils/init.c
+++ b/elkscmd/sys_utils/init.c
@@ -92,7 +92,9 @@
 */
 struct tabentry {
 	char id[3];
+	char norespawn;			/* don't respawn if fails within 3 seconds*/
 	pid_t pid;
+	time_t starttime;		/* start time of first spawn*/
 };
 
 static char *initname;			/* full path to this program*/
@@ -253,7 +255,8 @@ pid_t respawn(const char **a)
 	    baudrate = strchr(devtty, ' ');
 	    if (baudrate)
 		*baudrate++ = 0;
-	    if ((fd = open(devtty, O_RDWR)) < 0) fatalmsg("Can't open %s\n", devtty);
+	    if ((fd = open(devtty, O_RDWR)) < 0)
+			fatalmsg("Can't open %s (errno %d)\n", devtty, errno);
 
 	    argv[0] = GETTY;
 	    argv[1] = devtty;
@@ -270,7 +273,8 @@ pid_t respawn(const char **a)
 	}
 	else
 	{
-	    if ((fd = open(DEVTTY, O_RDWR)) < 0) fatalmsg("Can't open %s\n", DEVTTY);
+	    if ((fd = open(DEVTTY, O_RDWR)) < 0)
+		fatalmsg("Can't open %s (errno %d)\n", DEVTTY, errno);
 
 	    argv[0] = SHELL;
 	    argv[1] = "-e";
@@ -371,6 +375,7 @@ void enterRunlevel(const char **a)
 
 	if (!a[1][0] || strchr(a[1], runlevel)) {
 		int andWait=0;
+		struct tabentry *t;
 
 		/* if not running, spawn it */
 		if (!matchId(a[0])) {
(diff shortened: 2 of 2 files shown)


Mellvik:
@Ghaerr,

I've been testing this today - it works as expected on real HW, not on VirtualBox.

I think we can leave it at that and close the issue. There are too many more important issues on the plate to worry about VirtualBox right now.

-Mellvik

ghaerr:
> I've been testing this today - it works as expected on real HW, not on VirtualBox.

Thanks for testing. So we know it works on QEMU and real hardware. What does it do on VB, hang, or loop? This can be easily debugged now by setting DEBUG 1 in `init`, the messages have been updated to display making much more sense.

Mellvik:
OK; I'll take a stab at it.

VB does both - in different contexts: If the device is missing but enabled (inittab), init loops,
If the device is there (socket) and no connection at the other end, it blocks.


I'll be back.

—M

ghaerr:
>  If the device is missing but enabled (inittab)

The 'device' /dev/ttyS0 is always present but in the case of serial, it's possible that VB hangs in the middle of that. I added an error message to the init /dev/ttyX open to return errno, which sometimes helps.

>  init loops,

A respawn process is now stopped if it exits within 3 seconds, this can be seen with DEBUG on.
If it's taking longer than 3 seconds to open/exit, then it won't stop, this can also be seen.

> If the device is there (socket) and no connection at the other end, it blocks.

ELKS sockets used in inittab are an entirely different matter - are you still talking about inittab getty's here, or something else?

Mellvik:
I'm talking about VB serial connection to a host socket, sorry about the confusion.
ELKS has 3 serial lines enabled in this setup (inittab), VB has one line (ttyS0) defined.
The first screen is from ELKS waiting for  the other end of the serial line to be opened with netcat:
Skjermbilde 2020-04-17 kl  18 17 14

The next shot os from the looping, which starts immediately after opening the virtual serial line (which I can log on to w/o problems, and the system is useable).
Skjermbilde 2020-04-17 kl  18 19 04

--M

Mellvik:
BTW - in the last picture above, the looping goes like
-- hold ca. 1 sec
-- spawn on both
-- hold again

-M

ghaerr:
> BTW - in the last picture above, the looping goes like
-- hold ca. 1 sec

Try changing the 3 second timeout in 'init' line 412 to something higher, like 10.

ghaerr:
> The first screen is from ELKS waiting for the other end of the serial line to be opened with netcat

The hang issue is strange, seems VB is stopping emulation waiting for the netcat connection to start, possibly by intention. Not sure this can be fixed from ELKS. 

We should be able to fix the looping problem though, I'm thinking the timeout value or 'time(NULL)' itself may be causing the timeout/stop threshold not to be reached.

Mellvik:
@ghaerr,
I'm still working on this one.

> The first screen is from ELKS waiting for the other end of the serial line to be opened with netcat
> 
> The hang issue is strange, seems VB is stopping emulation waiting for the netcat connection to start, possibly by intention. Not sure this can be fixed from ELKS.
> 
That may well be the case. The new init (with levels working) may provide an opportunity to verify this.
> We should be able to fix the looping problem though, I'm thinking the timeout value or 'time(NULL)' itself may be causing the timeout/stop threshold not to be reached.
> 
Possibly a variable initialization issue: it works on a fresh virtual instance, loops after a reboot.
> —
> 
-M

Mellvik:
Update on this one:
- Verified that VB is indeed stopping emulation when waiting for the netcat.
- The reported reboot problem (cannot amount, resource busy) is a bug (or feature?) in the UTMP code. I had enabled USE_UTMP. Maybe we should log that as an issue.
- There is a bug preventing init from reading the last line of inittab. I'm submitting a fix for that.
- The looping - getty on nonetsiting port) is gone.

-M

ghaerr:
@Mellvik - thanks for the followups!

> Verified that VB is indeed stopping emulation when waiting for the netcat.
> The looping - getty on nonetsiting port) is gone.

Good news, sounds like the previous `init` problems are fixed (except for the last line not read).

> The reported reboot problem (cannot amount, resource busy) is a bug (or feature?) in the UTMP code.

That's because of /var/run/utmp, created and used by init and login.

> Maybe we should log that as an issue.

I think not, it's written up as buggy in login and also would cause a filesystem write on any boot, which I don't think is desirable. ELKS doesn't really need to be desktop UNIX/Linux all over again.

> There is a bug preventing init from reading the last line of inittab. I'm submitting a fix for that.

What a nightmare the init code has been, buggy everywhere. I thought I had all that fixed, and was trying to keep the stdio library out of init for size reasons, since it's always in memory. I'll take a look at it and your fix.

Mellvik:
thanks, 
yes, I'm familiar with utmp, I actually wanted to start using it :-)

git blew up on me, I cannot get the PR through, seems like some internal error. Will look at it again, maybe it's temporary. 

M

Mellvik:
Anyone seen this before?


Skjermbilde 2020-04-23 kl  18 25 20

Issue #372, opened 2020-02-23
IRQ setting in config? (Enhancement)


Mellvik:
Somewhere down the line ELKS needs a simplified way to set irq# for peripheral interfaces. 
My encounter with NE2k a couple of years back was interesting, but not something a regular user would be able to understand - 3 changes i 2 files, and a combination of numbers and bits (mask).

jbruchon:
We would probably benefit by having something like sysfs on Linux which makes stuff like this not only easy to change after building and booting, but also scriptable.

pawosm-arm:
I've encountered the same problem with the ne2k card clone I have. I always need to edit ELKS sources to build the kernel for the XT-Turbo machine equipped with that ISA8 card. And it's not only IRQs. I/O ports and I/O mem also need to be changed. Definitely, ability to set it up in `menuconfig` would be beneficial.

@Mellvik was your encounter successful? What numbers and bits did you change?

Mellvik:
@pawosm-arm,
Yes, it was successful, and not all that complicated (I got help from the people on the list), but to a 'regular' (non technical) user, it would presumably be really hard.

In my case, I needed IRQ 11, here it is:

https://github.com/Mellvik/elks/commit/209a0eb0606ecf3340db49d4bfe0422be44d6d17

-Mellvik

pawosm-arm:
> In my case, I needed IRQ 11, here it is:

Ah, so you have 286 AT machine, so it's easier. It is much harder on XT where there's only one 8259 significantly limiting the number of available IRQs. In my case, I was left with only IRQ2 free for the network card. Similar problem with I/O ports. The only way to set IRQ, I/O ports and I/O mem not colliding with the other ISA8 cards (e.g. I could not let it use I/O base 300h as it was already taken by the XT-IDE controller) was to run DOS program that configures settings held in the card's EEPROM.

What I eventually came up with was:
IRQ 2
I/O base 240h

...and my patch was slightly longer:

```
diff --git a/elks/arch/i86/drivers/net/eth-main.c b/elks/arch/i86/drivers/net/eth-main.c
index 99b20971..00a95385 100644
--- a/elks/arch/i86/drivers/net/eth-main.c
+++ b/elks/arch/i86/drivers/net/eth-main.c
@@ -21,7 +21,7 @@

 static byte_t eth_inuse = 0;

-static byte_t mac_addr [6] = {0x52, 0x54, 0x00, 0x12, 0x34, 0x56};  // QEMU default
[...]

PR #561, opened 2020-04-17
[kernel] Add customizable IRQ and I/O ports for all drivers

ghaerr:
Allows programmatic customization of all IRQ \#'s and port addresses for all drivers that use interrupts by editing a single new file, 'elks/include/arch/ports.h'. 

Requested in #372. 

This change should allow customization of all hardware device port numbers and IRQs.
After working through this, there are quite a few available IRQs available for use. See ports.h for details. Here is the top of that file:
```
/*
 * ELKS I/O port and IRQ mappings
 *
 * PC IRQ default mappings (* = possible conflicting uses)
 *
 *  IRQ ELKS Device         Config Option           Status
 *  0   Timer                                       Required
 *  1   Keyboard            CONFIG_CONSOLE_DIRECT   Optional
 *  2*  Cascade -> 9 AT, Unused on XT               Optional on XT, not available on AT
 *  3   Com2 (/dev/ttyS1)   CONFIG_CHAR_DEV_RS      Optional
 *  4   Com1 (/dev/ttyS0)   CONFIG_CHAR_DEV_RS      Optional
 *  5*  Unused
 *  5*  Com3 (/devb/ttyS2)  CONFIG_CHAR_DEV_RS      Optional
 *  5*  HW IDE hard drive   CONFIG_BLK_DEV_HD       Driver doesn't work
 *  6*  Unused
 *  6*  HW floppy drive     CONFIG_BLK_DEV_FD       Driver doesn't compile
 *  7   Unused (LPT)
 *  8   Unused (RTC)
 *  9   NE2K (/dev/eth)     CONFIG_ETH_NE2K         Optional
 * 10   Unused (USB)                                Turned off
 * 11   Unused (Sound)                              Turned off
 * 12   Unused (Mouse)                              Turned off
 * 13   Unused (Math coproc.)                       Turned off
 * 14*  AT HD IDE (/dev/hda) CONFIG_BLK_DEV_HD      Driver doesn't work
 * 15*  AT HD IDE (/dev/hdb) CONFIG_BLK_DEV_HD      Driver doesn't work
 *
 * Edit settings below to change port address or IRQ:
 *   #define CONFIG_NEED_IRQx to map interrupt vector to ELKS
 *   Change I/O port and driver IRQ number to match your hardware
 */
```

Not completed: 
- The NE2K network driver needs more modifications to 'elks/arch/i86/drivers/net/ne2k-mac.s' to access port ranges by adding to a base port, rather than using set constants.
- Customization is through editing ports.h only, not by `menuconfig`.

### DIFF
--- a/elks/arch/i86/drivers/block/directhd.c
+++ b/elks/arch/i86/drivers/block/directhd.c
@@ -58,7 +58,7 @@ static struct file_operations directhd_fops = {
 /* what is this good for ? */
 static int access_count[4] = { 0, };
 
-static int io_ports[2] = { 0x1f0, 0x170 };
+static int io_ports[2] = { HD1_PORT, HD2_PORT };
 
 static int directhd_initialized = 0;
 
--- a/elks/arch/i86/drivers/block/floppy.c
+++ b/elks/arch/i86/drivers/block/floppy.c
@@ -54,11 +54,12 @@
  * removed volatile and fixed function definitions
  */
 
+#ifdef CONFIG_BLK_DEV_FD
+
 #define TRP_TIMER
 #define TRP_SIG
 
 #define REALLY_SLOW_IO
-#define FLOPPY_IRQ 6
 #define FLOPPY_DMA 2
 
 #include <linuxmt/config.h>
@@ -77,15 +78,6 @@
 #include <arch/io.h>
 #include <arch/segment.h>
 
-#ifdef CONFIG_BLK_DEV_FD
-#if 1
-/* Hack until this code works */
-void floppy_init(void)
-{
-/* Do nothing */
-}
-
-#else
 
 #define FLOPPYDISK
 #define MAJOR_NR FLOPPY_MAJOR
@@ -1450,7 +1442,5 @@ void floppy_init(void)
 	reset_floppy();
     }
 }
-
-#endif
 
 #endif
--- a/elks/arch/i86/drivers/char/serial.c
+++ b/elks/arch/i86/drivers/char/serial.c
@@ -10,6 +10,7 @@
 #include <linuxmt/termios.h>
 #include <linuxmt/debug.h>
 #include <arch/io.h>
+#include <arch/ports.h>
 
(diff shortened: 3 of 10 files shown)


Mellvik:
Thanks @ghaerr,
this is a lot of work and a lot of changes in a short time. And a vast improvement to say the least.

My only comment would be about the 'system' interrupts, which may be more clearly identified in the overview in order to avoid confusion for less experienced experimenters.

That is, 0, 1, 8 and 13 are reserved, 0 and 8 are required for system (elks) operation, I guess 1 can be turned off as indicated, so as not to cause spurious interrupts. 13 is some times a jumper on the MB, some times more or less automatic BIOS config, living its own life so to speak.

A major achievement from my point of view! Looking forward to test!

-Mellvik

ghaerr:
> My only comment would be about the 'system' interrupts, which may be more clearly identified in the overview in order to avoid confusion for less experienced experimenters.

The way the ELKS interrupt initialization works is that if an interrupt is used by elks (using a \#define CONFIG_NEED_IRQx), then that interrupt vector is patched to point to ELKS and the old vector discarded. The old vector isn't saved or ever called. That means if the interrupt is required by the BIOS, for instance to do hard disk I/O, then the associated BIOS calls won't work, and an ELKS driver must be present. This is why the interrupt vectors are mostly controlled by device driver CONFIG options.

> That is, 0, 1, 8 and 13 are reserved

As I mentioned in #372, many of these are hard-wired on the motherboard, right? 
This PR doesn't change anything from the way ELKS previously worked, just makes it more configurable. Keyboard INT 1 was not patched if the console driver isn't turned on, and INT 8 and 13 were never patched (used by ELKS). So that doesn't mean they're "available".

I'm open for all suggestions on how to proceed.

Mellvik:
Agree @ghaerr,
my comment was not related to content, only presentation: Ensuring that new users get the right picture right away.


As to IRQ and wiring: Yes, they are indeed hardwired. You can turn them off and on, but their use is locked.

—Mellvik

PR #564, opened 2020-04-18
[shell] Enhance shell autocomplete to operate identical to BASH

ghaerr:
Enhance TAB completion in `sh` to operate very closely to BASH autocomplete.

Thanks again to @georgp24 for bringing the lightweight `linenoise` library to ELKS.

Command line editing and TAB completion make using ELKS `sh` as easy to use as `bash`, and in a lot less space.

### DIFF
--- a/elkscmd/ash/autocomplete.c
+++ b/elkscmd/ash/autocomplete.c
@@ -1,116 +1,116 @@
 /*
- * This implements the autocompletion callback (or TAB completion) for Linenoise
- * see the shell.txt file in the Documentation/text directory for details
+ * This implements the autocompletion callback (or TAB completion) for ash/linenoise.
+ * Autocompletion operates identically to BASH shell. Greg Haerr 17 Apr 2020
  */
 
 #include "shell.h"
 
-#if LINENOISE
-
+#if LINENOISE		/* entire file if LINENOISE=0 in shell.h*/
 #include <dirent.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <stdio.h>
 #include <string.h>
 #include "linenoise.h"
 
-void completion(const char *buf, linenoiseCompletions *lc) {
-/* completion callback for ELKS */
+void completion(const char *buf, linenoiseCompletions *lc)
+{
+	char *arg, *file;
+	DIR *fp;
+	struct dirent *dp;
+	int count;
+	char dir[128];
+	char line[128];
+	char result[128];
+	char lastentry[128];
 
-int files_found=0;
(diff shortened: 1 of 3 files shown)


Mellvik:
@ghaerr, @georgp24

thanks !!!! — a great and very welcome improvement to (a)sh.
It may make sense to move command history over (from sash) as well.

If you think it’s worthwhile I’1ll put it on my list.

—M

ghaerr:
> It may make sense to move command history over (from sash) as well.

That will end up being quite a bit more complicated in `ash` than in `sash`, I think, and
take a different approach, since linenoise already deals with some history. I have been intently studying linenoise, its an interesting small piece of code that does quite a bit.

> If you think it’s worthwhile I’1ll put it on my list.

It if can be contained solely within linenoise and clean, that would be nice. Otherwise we need to worry about `sh` breaking its BASH script execution capability, and so far, my attempts to understand its operation have met with lots and lots of code!

I would still like to get a patch on the FIFO detect that you pointed out is incorrect, regardless of the serial I/O testing results, before we get too far down this rabbit hole :)

Mellvik:
Roger that. Admittedly I haven't even looked at ash yet. And a couple of important features are still missin from sash.

Serial: turns out my 3rd ISA card is 16550A too, so there is no testing of the buggy FIFO chip. I'll check FIFO on physical tomorrow.

Took the weekend off. Recommended :-)

--Mellvik

ghaerr:
> Serial: turns out my 3rd ISA card is 16550A too, so there is no testing of the buggy FIFO chip. 

Does your revised (working) FIFO detect code recognize that properly now?

> Took the weekend off. Recommended :-)

Yes, I'm trying to do that too lol.

georgp24:
Mellvik,
you could use the linenoise version Greg and I have made with sash too. Just replace the read command for the command line in sash with a call to linenoise. This returns a line edited and autocompleted command line input. See the lines 214-234 in `ash/input.c.` This is the code change required for ash.
Georg

Mellvik:
Thanks @georgp24,

I'll take a close look.

It is also important to keep in mind that the two shells have different purposes. Making ash more bash-like is good, keeping sash lightweight is good.

—Mellvik

georgp24:
Linenoise is very lightweight, so you could use that with sash!

I took a look at the sash code. In the readline() function in sash.c there is a fgets() call used and stdin opened. Replace this code following along the lines 214-234 in ash/input.c.

You will need to add to the sash makefile the files ash/autocomplete.c ash/linenoise.c and ash/linenoise.h

PR #570, opened 2020-04-19
Shell config file cleanup

cjsthompson:
* Make sash also source /etc/sashrc

* Move shell config files to /etc so not only root gets their
  environment set.

* Trim down $PATH since ELKS now only uses /bin

Screenshot_2020-04-19_15-34-12


ghaerr:
Hi @cjsthompson, thanks for your contribution!

A couple small points, in order to cut down on the number of additional \#ifdef's and improve readability, `buf` on line 241 could go to 230, and `buf` on 251 could go to 282.

`sourcecfg` should probably append the "/" itself, since it's passed a path, and it's not obvious the passed path has to end in "/".

cjsthompson:
Thanks for your review. I added your suggestions and fixed a problem I introduced.

ghaerr:
Thank you Christoph!

PR #572, opened 2020-04-19
Fix chgrp and add a clear command.

cjsthompson:
While attempting to edit files with vi as a normal user, I noticed chgrp wasn't working and I guessed this must be the chmem problem. So I added the compiler flag. Not sure if this is the right value for the flag. But now it works.

Before:

Screenshot_2020-04-19_18-18-55

After:

Screenshot_2020-04-19_21-08-01

Also while I was at it I added a very simple clear command. Linux uses ncurses to do that, I'm guessing because it works with every terminal since the neolithic. But since ELKS doesn't have a curses I guess this may be good enough until then.

Screenshot_2020-04-19_21-49-22

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -84,6 +84,7 @@ sys_utils/poweroff				:sysutil			:720k
 sys_utils/who					:sysutil				:1440k
 sys_utils/unreal16				:sysutil				:1440k
 sh_utils/basename				:be-shutil			:720k
+sh_utils/clear					:be-shutil		:1440k
 sh_utils/date					:be-shutil		:360k
 sh_utils/dirname				:be-shutil			:720k
 sh_utils/false					:be-shutil				:1440k
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -84,7 +84,7 @@ sys_utils/poweroff				:sysutil			:720k
 sys_utils/who					:sysutil				:1440k
 sys_utils/unreal16				:sysutil				:1440k
 sh_utils/basename				:be-shutil			:720k
-sh_utils/clear					:be-shutil		:1440k
+sh_utils/clear					:shutil		:1440k
 sh_utils/date					:be-shutil		:360k
 sh_utils/dirname				:be-shutil			:720k
 sh_utils/false					:be-shutil				:1440k
--- a/elkscmd/file_utils/Makefile
+++ b/elkscmd/file_utils/Makefile
@@ -19,7 +19,7 @@ cat: cat.o
 	$(LD) $(LDFLAGS) -o cat cat.o $(LDLIBS)
 
 chgrp: chgrp.o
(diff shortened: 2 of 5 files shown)


ghaerr:
Thanks @cjsthompson!

It's nice to have the system actually used by another person! And, of course, for now, that usually means finding another broken command lol :)

The PR looks fine, except one thing: when adding `clear`, you'll want to use just the ":shutil" tag, not ":be-shutil", as the latter is reserved for commands which are also implemented in busyelks, which this is not.

 When figuring an amount of "working memory (data segment total size)" for an `sbrk` failure, you can start with the ELKS `chmem` from within ELKS. It prints it's usage with no arguments, and starting with `chmem +1024 chgrp` (in decimal) or something like that will let you play with the broken command a bit before adding a permanent -ma.out-chmem= (in hex) fix. In this case, allocating 0xc000 (48k) to the data segment is overkill, but it's deallocated immediately thereafter, so no big deal.

You beat me to the `clear` command... after watching your video I figured you might write one!

cjsthompson:
> Thanks @cjsthompson!
> 
> It's nice to have the system actually used by another person! And, of course, for now, that usually means finding another broken command lol :)

I hope to use it again someday on real hardware. It think it works much better now than 0.1.x which I thought was better than Minix 1 at least because it's Linux-like.

> The PR looks fine, except one thing: when adding `clear`, you'll want to use just the ":shutil" tag, not ":be-shutil", as the latter is reserved for commands which are also implemented in busyelks, which this is not.

Change ready to push.
 
> When figuring an amount of "working memory (data segment total size)" for an `sbrk` failure, you can start with the ELKS `chmem` from within ELKS. It prints it's usage with no arguments, and starting with `chmem +1024 chgrp` (in decimal) or something like that will let you play with the broken command a bit before adding a permanent -ma.out-chmem= (in hex) fix. In this case, allocating 0xc000 (48k) to the data segment is overkill, but it's deallocated immediately thereafter, so no big deal.

Thanks for the explanation. I think I got it right.

Screenshot_2020-04-19_22-52-05

17432 is the first value in increments of 1024 where the command succeeds. So it should be -maout-chmem=0x4418 ?
 
> You beat me to the `clear` command... after watching your video I figured you might write one!

I type 'clear' faster than I think so it's nice to have :) Would you have done it differently?

ghaerr:
> 17432 is the first value in increments of 1024 where the command succeeds. So it should be -maout-chmem=0x4418 ?

Go ahead and round it up to 0x6000 for now. I have been looking into this since your PR, and I think there's a problem with memory allocation in libc. There's 16k allocated for heap and stack to each process, and the the sys_brk error occurs when there's less than 4k left... I've found that over 12k is being eaten up!

> I type 'clear' faster than I think so it's nice to have :) Would you have done it differently?

Nope - couldn't have done it better. Thanks!

PR #576, opened 2020-04-23
Bugfixes in init.c, more.c + expanded termcap

Mellvik:
- init() failed to read the last line of inittab, rewrote the scanFile function (kept the old code for review in #if 0)
- more() didn't wait for input, set VMIN to 1 and fixed uninitialized terms struct.
- added insert line(s), delete line(s), delete character(s) to ansi termcap entry, tested in VB and on serial.

### DIFF
--- a/elkscmd/file_utils/more.c
+++ b/elkscmd/file_utils/more.c
@@ -25,8 +25,9 @@ int more_wait(int fout, char *msg)
 
 	if (tcgetattr(1, &termios) >= 0) {
 		struct termios termios2;
+		tcgetattr(1, &termios2);
 		termios2.c_lflag &= ~ICANON;
-		termios2.c_cc[VMIN] = 0;
+		termios2.c_cc[VMIN] = 1;
 		tcsetattr(1, TCSAFLUSH, &termios2);
 	}
 	read(1, buf, sizeof(buf));
@@ -45,7 +46,7 @@ int more_wait(int fout, char *msg)
 		return 1;
 	case 'Q':
 	case 'q':
-		write(1, "        \r", 9);
+		write(1, "          \r", 11);
 		close(fd);
 		return -1;
 	}
--- a/elkscmd/misc_utils/time.c
+++ b/elkscmd/misc_utils/time.c
@@ -13,8 +13,8 @@ static void printt(char * s, long se, long us)
 	long mins, secs;
 
 	if (se == -1 ) {		/* microsecs only */
-		if (us < 1000)		/* round up to 1/1000 second*/
-			us = 1000;
+		if (us < 1000L)		/* round up to 1/1000 second*/
+			us = 1000L;
 
 		mins = us / 60000000L;
 		if (mins)
@@ -24,9 +24,9 @@ static void printt(char * s, long se, long us)
 		if (secs)
 			us -= secs * 1000000L;
 	} else {
-		mins = se / 60;
+		mins = se / 60L;
 		secs = se;
-		if (mins) secs -= se * 60;
+		if (mins) secs -= se * 60L;
 	}
 	us /= 1000L;
 
@@ -69,8 +69,8 @@ int main(int argc, char **argv)
 	after.tv_sec -= before.tv_sec;
 	after.tv_usec -= before.tv_usec;
         if (after.tv_usec < 0)
-                after.tv_sec--, after.tv_usec += 1000000;
-	if (after.tv_usec > 500000) after.tv_sec++;
+                after.tv_sec--, after.tv_usec += 1000000L;
+	if (after.tv_usec > 500000L) after.tv_sec++;
 	fprintf(stderr, "\n");
 	printt("real", after.tv_sec, after.tv_usec);
 	//printt("user", -1, end.tms_cutime - start.tms_cutime);
--- a/elkscmd/misc_utils/time.c
+++ b/elkscmd/misc_utils/time.c
@@ -8,34 +8,45 @@
 #include <sys/times.h>
 #include <sys/wait.h>
 
-static void printt(char *s, long a)
+static void printt(char * s, long se, long us)
 {
 	long mins, secs;
 
-	if (a < 1000)			/* round up to 1/1000 second*/
-		a = 1000;
+	if (se == -1 ) {		/* microsecs only */
+		if (us < 1000)		/* round up to 1/1000 second*/
+			us = 1000;
 
-	mins = a / 60000000L;
-	if (mins)
-		a %= mins * 60000000L;
+		mins = us / 60000000L;
+		if (mins)
+			us -= mins * 60000000L;
 
-	secs = a / 1000000L;
-	if (secs)
-		a %= secs * 1000000L;
+		secs = us / 1000000L;
+		if (secs)
+			us -= secs * 1000000L;
+	} else {
+		mins = se / 60;
+		secs = se;
+		if (mins) secs -= se * 60;
+	}
+	us /= 1000L;
 
-	fprintf(stderr, "%s\t%lum%lu.%03lus\n", s, mins, secs, a / 1000L);
+	fprintf(stderr, "%s\t%lum%lu.%03lus\n", s, mins, secs, us);
+	
 }
+
 
 int main(int argc, char **argv)
 {
 	int status, p;
 	struct tms end, start;
+	struct timeval before, after;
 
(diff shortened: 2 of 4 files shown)


ghaerr:
I'm rerunning the checks... I've seen GitHub get confused before on some PRs.

On the /etc/termcap fix, there are now duplicate entries for cd and ce. Can you fix that? 
When making termcap changes, I prefer to keep all the changes on a separate line, just in case problems arise. Could you fix the dups and put the other additions all on one line perhaps just above the cursor keys or at the end of the entry?

ghaerr:
Geez, looks like I have time.c all messed up...

There shouldn't be any difference between the 'real' and 'sys' time, that indicates that the `times` system call is faulty. Are you displaying both to show the difference?

I thought I had the math correct for the time printing... have you tested your mods while running `time vi` for over a minute? I had everything working until I tried that.

A side note, all constants over 65535 should have an 'L' after them, some compilers may truncate to 'int' without saying anything, and the checks could fail (50000, etc).

Mellvik:
time: I really needed wall clock time, now it works. try it with time `cat /dev/bda > /dev/null`, you get a few secs. I need to get it over to physical to get minutes.
Init: please review and I'll ditch the old code.

ghaerr:
> I need to get it over to physical to get minutes.

I was saying that if you run 'time vi' and sit there for a minute, that will check the minutes code.

> Init: please review and I'll ditch the old code.

Your fix is fine, thank you. Lets run with your new code, and leave the old code in until I can look at it, I spent way too long trying to fix that the first time, and want to take another look at it when I have a moment for it.

ghaerr:
> leave the old init code in until I can look at it, 

I looked at it, way too complicated. And also realized that the stdio lib is already included for the `fatalmsg` routines. So go ahead and delete the old code if you like. And thanks!

Mellvik:
Yes, well, as I mentioned in my complaint about sash the other day, time was reporting .8s when real time was more like 20s. 
I needed clock time andD got that with this mod. times() still does what it's supposed to - reporting the time consumed by the process itself. The numbers seem reasonable. There is no difference between system time and user time, I guess that's an overt simplification.

PR mod is coming.

BTW - would be nice to have cpu time in ps() one of these days.

-M

ghaerr:
> times() still does what it's supposed to - reporting the time consumed by the process itself. 

No, `times` isn't even implemented as a system call - look at libc/system/times.c - it uses `gettimeofday` just like you're doing. It has to stuff the two-element time into a single long however, as that's what `times` expects.

> There is no difference between system time and user time

That should be the case, since libc `times` only measures clock time. The changes made outside of 'printt' to 'time.c' should be put into 'printt' so that extra processing isn't needed outside of `times`, especially for when it is implemented in the kernel.

We could keep them in there for the time being for your comparative testing, but I now realize that `times` isn't buggy, just time.c printt function.

cjsthompson:
I've been rewriting more.c for the last 3 days because I noticed in didn't work as well :)

Mellvik:
@ghaerr, thanks for the clarification.

> 23. apr. 2020 kl. 22:22 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> times() still does what it's supposed to - reporting the time consumed by the process itself.
> 
> No, times isn't even implemented as a system call - look at libc/system/times.c - it uses gettimeofday just like you're doing. It has to stuff the two-element time into a single long however, as that's what times expects.
> 
> There is no difference between system time and user time
> 
> That should be the case, since libc times only measures clock time. The changes made outside of 'printt' to 'time.c' should be put into 'printt' so that extra processing isn't needed outside of times, especially for when it is implemented in the kernel.
> 
> We could keep them in there for the time being for your comparative testing, but I now realize that times isn't buggy, just time.c printt function.
> 
OK, admittedly I didn't check the original prints-logic.
so - what you're saying is that there is no 'time usage recording' in the process structure in ELKS?

Anyway this explains why I'm getting the same millisecond count from times and gettimeofday :-).

Fine, let's keep it like this for testing, I'll remove the one that doesn't work right in a few days.


—Mellvik

Mellvik:
Christoph, 
don't let my fixes dissuade you from making improvements.

That said, IMHO porting more from Minix2 would be a good idea. It has quite reasonable functionality and is reasonably compact.

—Mellvik

cjsthompson:
> Christoph, don't let my fixes dissuade you from making improvements. That said, IMHO porting more from Minix2 would be a good idea. It has quite reasonable functionality and is reasonably compact. —Mellvik

I'm not even remotely aiming for POSIX compliance (even 'more' from util-linux doesn't) :) But I think there's some small things like reverse video for the prompt, inline help, showing % read when reading from files and TIOCGWINSZ that can be done without curses, don't require a lot of extra code and are nice to have. I need maybe another day or two for finishing all this and testing with all kinds of files to make sure it works as expected and I'll make a pull request,

I've been thinking about what other tools would be useful. I thought about writing fold, nl and expand as well. But then I realized that sed can do all that too. So if all that works with the sed included with ELKS, I'm not sure there's a good reason to have those.

Mellvik:
Good reasoning, Christoph. And I agree with you, those are nice to have features. On my list for more is '1G' and 'G' commands - move to the beginning or end of a file.
I don't know anything about POSIX more, but you may get som inspiration from Minix2 yap (yet another pager, aka more) and BSD4.2 more.

on elks, more() currently has a footprint of < 3k, which is really good.

-M

cjsthompson:
> Good reasoning, Christoph. And I agree with you, those are nice to have features. On my list for more is '1G' and 'G' commands - move to the beginning or end of a file.

I can see how to do that easily without curses or without malloc() which is my goal to keep things slim.

>I don't know anything about POSIX more,

I found a document "POSIX commands" from 2009 and there is a description of what more should do in there. It has way more options and commands than the 'more' of util-linux which has almost 2000 lines of code.

>but you may get some inspiration from Minix2 yap (yet another pager, aka more) and BSD4.2 more. on elks, more() currently has a footprint of < 3k, which is really good.

I'll check them out. Thanks.

ghaerr:
> OK, admittedly I didn't check the original prints-logic.

I'm not following you, you made changes to that logic that were needed, right?

> so - what you're saying is that there is no 'time usage recording' in the process structure

Correct.

> Fine, let's keep it like this for testing, I'll remove the one that doesn't work right in a few days.

Ok. I'll commit this as is, but we need to get `time` fixed so it does not use any `gettimeofday` calls outside of `times`. If the original method works with your revised `printt` function, then the other code in `main` can be removed.

Also, please test running > 1 minute. I did this by running `time vi` and then editing a file for a minute.

Mellvik:
@ghaerr 

>    OK, admittedly I didn't check the original printt-logic.

> I'm not following you, you made changes to that logic that were needed, right?
Sorry, I misread you. The fact is, I (almost) fixed yours and screwed mine. Pushing a fix. Minutes are good.

BTW - what's the reason for not using gettimeofday?

-M

ghaerr:
> BTW - what's the reason for not using gettimeofday?

`gettimeofday` is used by `times`. I was saying that the permanent time.c fix should not include gettimeofday outside of the `times` function, since times will be expected to work in other circumstances, not just time.c. That said, time.c's display code should also be self contained, and not require other calculations outside of it.

Mellvik:
> BTW - what's the reason for not using gettimeofday?
> 
> gettimeofday is used by times. I was saying that the permanent time.c fix should not include gettimeofday outside of the times function, since times will be expected to work in other circumstances, not just time.c. That said, time.c's display code should also be self contained, and not require other calculations outside of it.
> 
> 
> 
OK, I didn't really understand that, but take your word for it. It just seemed complicated to me. 
I'll have a fix tomorrow.

--M

ghaerr:
> It just seemed complicated to me.

You're welcome to rewrite all of the code - I was just suggesting to try to contain it, rather than having it spread across time.c etc. If the libc `times` function needs changing also, go for it.

Hopefully we can use an updated `times` routine, rather than separate code using `gettimeofday`, to make time.c simpler, (except for the print code). That would allow its inclusion in the kernel someday.

(A simpler way to code up libc routines is to replace/place them in the application's .c file outside the library until they're working. This saves lots of recompilation time since libc has to be recompiled from scratch sometimes or things don't work, FYI.)

Mellvik:
@ghaerr,
I'm trying to get my head around how things work, and this helps.

PR #577, opened 2020-04-24
[kernel] Dynamically allocate TTY input/output queues

ghaerr:
Allocate TTY input and output character queues dynamically [EDIT: allocated on open].
That allows different-sized queues based on hardware/tty device.

Console TTY and PTY input queues reduced from 512 to 128 bytes. All output queues 64 bytes.

Serial TTY input queue increased from 512 to 1200 bytes (= SLIP MTU + 136 for `ktcp`). This should stop the data loss on SLIP packets, and improve data-loss problems reported in #515.

Add `sercat` test program to help @Mellvik's throughput and data-loss testing in #515.
Usage: `sercat [-v] [serial device] [ > file]`. On serial login port, use `sercat > file`. From console, use `sercat -v /dev/ttyS0 > file`, then send file across serial line. ^D to quit.

After testing with `sercat`, I've realized that the VMIN/VTIME termios settings are not working properly. Subsequent PR will provide a fix, which will improve overall serial performance for high-speed serial input data. The current implementation requires VMIN=1 and will not wait with timeout for a larger packet size (e.g. =1024) to be received before returning.

### DIFF
--- a/elks/arch/i86/drivers/char/dircon.c
+++ b/elks/arch/i86/drivers/char/dircon.c
@@ -559,12 +559,14 @@ static int Console_write(register struct tty *tty)
 
 static void Console_release(struct tty *tty)
 {
-/* Do nothing */
+    tty_freeq(tty);
 }
 
 static int Console_open(register struct tty *tty)
 {
-    return (tty->minor >= NumConsoles) ? -ENODEV : 0;
+    if (tty->minor >= NumConsoles)
+	return -ENODEV;
+    return tty_allocq(tty, INQ_SIZE, OUTQ_SIZE);
 }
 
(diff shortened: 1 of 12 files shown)


ghaerr:
Pushed fix for proper VMIN/VTIME operation.

`sercat` sets VMIN to 255, and VTIME to 1, which should allow up to 255 characters to be read from the serial line at a time (until a 0.1 second timeout). This change combined with a 1200 byte input queue should allow for much greater system throughput from UART to disk.

A clarification on TTY queue allocation, queues are allocated on device open, and deallocated on close. So, for most operations, even thought the serial input queue size has increased, overall kernel memory usage is lowered.

PR #579, opened 2020-04-25
Last past of sash cmd history implemented.

Mellvik:
Added history search (`!ls` etc.), recall last arg from previous command (`!$`) and support for history commands anywhere on the command line.

### DIFF
--- a/elkscmd/sash/cmd_history.c
+++ b/elkscmd/sash/cmd_history.c
@@ -88,7 +88,7 @@ cmd_search(char * pat) {
 	int idx, i;
 	char *found;
 
-	printf("search %s, cht %d\n", pat, histcnt);
+	//printf("search %s, cht %d\n", pat, histcnt);
 	for (i = 0; i < histcnt; i++) {
 		idx = map_ind(-i);
 		if ((found = strstr(histbuf[idx], pat)))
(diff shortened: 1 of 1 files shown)


ghaerr:
I'll have to give this a test, thanks.

No big deal, but functions (cmd_search, fixbuf, etc) in "library" code that are linked to other applications (like sash, in this case), should probably be declared static so that its obvious they aren't called outside the implementation, and won't possibly interfere when reused elsewhere.

Mellvik:
OK, makes sense. I'll fix that along with a fix for a (new) bug I just found in the command edit function (^old^new).

I also want to take a look at the SIGINT problem we discussed before.

--Mellvik

ghaerr:
> I also want to take a look at the SIGINT problem we discussed before.

Are you talking about trying to get the sash prompt displaying properly after an interrupt in all cases? I looked into that and I don't think an easy fix is possible, without using `setjmp`, which is a lot of trouble. I'm interested if you find a simpler solution.

ghaerr:
@Mellvik, I just noticed that the compilation of your cmd_history file is spitting out a number of warnings, all of which should be looked into. This should be standard procedure when compiling new code. This can be made easier amidst the mess of elkscmd/ warnings by compiling within elkscmd/sash initially, rather than with a full make.

PR #580, opened 2020-04-25
LZSS compression lib & tools 

marcin-laszewski:
Tiny code: encoding - 2.2kB,  decoding - 1.7kB.

User-space tools ('lzss' in menuconfig)
.................................................
lzss-c < /bin/ls | lzss-d > /tmp/ls
sum /bin/ls /tmp/ls
ls -l /bin/ls /tmp/ls
.................................................

Kernel: menuconfig -> Kernel & Hardware -> Kernel settings -> Compressed executables
IT DOES NOT WORK. I don't know why.... Help me, please.


marcin-laszewski:
It is not for merge of course.
(ok, not yet).

ghaerr:
> IT DOES NOT WORK. I don't know why.... Help me, please.

I have reviewed your code, and offer the following advice:

First, the implementation in 'fs/exec.c' is far too complicated, with far too many \#ifdefs. I have no idea why it doesn't work, and now neither do you, way too hard to follow. The proposed implementation adds too many special case checks, and instead of using the `read` function pointer at 'filp->f_op->read', introduces lots of static functions and special case handling in an important kernel routine. Even if this worked, it could not be merged.

The function pointers in ELKS are good design. Use them. An elegant way to solve the problem of reading compressed files (and executing them) would be to replace the filp->f_op->read function pointer to an entry point in the LZSS library that reads the possibly compressed file, and always returned an uncompressed stream, regardless of whether the file was compressed or not. In this way, little or no code in fs/exec.c would change. Almost all code would be contained within the LZSS library, thus making debugging considerably easier. It could be bootstrapped by having the LZSS library originally just be a NOP, that is, always returning the original data stream without trying to decompress it.

More detail on how to do this would be to look at `open_filp` in fs/exec.c and fs/file_table.c, and modify that routine to pass the filp pointer to an LZSS init routine. That routine would set the filp->f_op->read pointer to the LZSS read entry point, and save the original read pointer as well. Then, the LZSS library, when requested to read through the f_op read function pointer in fs/exec.c, would read from the saved read function, perform the decompression, and return the requested uncompressed bytes back up the chain. No changes to fs/exec.c.

Other comments: to be considered for merge, Makefiles cannot \#ifdef and include specialized \*.mk make instructions. This includes fs/lzss.mk, elkscmd/Make.install, and elkscmd/Makefile. It is too difficult to add varying make instructions for individual contributions, they should conform to the method that ELKS is already using.

The elks/lib/lzss library should be moved to elks/fs/lzss, since its not really a general library. Instead, its a replacement `read` function with the same interface as other read functions, to be used only in the kernel VFS code.

As a side note, if the design were changed to replace the `read` f_op function whenever configured, this would allow for arbitrary reading of ANY file on the filesystem in compressed format. That is, since the decompression function would be looking at its own magic number in the first few bytes of the file, reading any file would have the effect of returning an uncompressed byte stream. That means that almost every file could be compressed on the filesystem, not just executables, and everything would just work.

marcin-laszewski:
I think this problem is related to #591.

ghaerr:
> I think this problem is related to #591.

I hadn't noticed the large buffer allocation since it was hidden in lzss.h, but allocating a 4k buffer on the stack definitely won't work in the kernel!

Providing kernel file decryption is looking more and more to me like an intermediate 'filesystem' style driver. The pipe implementation in elks/fs/pipe.c shows one way to manage intermediate kernel data storage. Another thought is using the system buffer mechanism, they are tied to disk blocks, but could possibly be used.

PR #582, opened 2020-04-25
[cmds] Fix grep

ghaerr:
I noticed that `grep` from file_utils is totally broken: `ls -l / | grep home` produces nothing.

Replaced with duplicate grep from elkscmd/minix1/.

### DIFF
--- a/elkscmd/APPS
+++ b/elkscmd/APPS
@@ -31,9 +31,9 @@ CONFIG_APP_COREUTILS	core_utils/		Core utilities, Installed in /bin. Sources mov
 											cat, date, env
 										Move from sys_utils: clock, getty, init, login, mount, umount
 										Move from file_utils:
-											cat, chgrp, chmod, chown, cp, dd, grep, ln, mkdir, mknod,
+											cat, chgrp, chmod, chown, cp, dd, ln, mkdir, mknod,
 											more, mv, rm, rmdir, sync, touch
-										Move from minix1:
+										Move from minix1: grep
 										Move from minix2: env
 										Move from minix3: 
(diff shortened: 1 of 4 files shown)


Mellvik:
@ghaerr -
I'm still having trouble with grep, although I suspect the problem is somewhere else.
Have a look at the attached screen shot.
After reboot, group works. Then - after `time cat /bin/* > /dev/null`  grep bombs the system, either rebooting or reporting a segmentation fault.

--Mellvik 

Skjermbilde 2020-04-27 kl  16 09 39

ghaerr:
Hmmm, possible serious problems.

I haven't looked at the new `grep`, I just replaced the file_utils version with the minix3 version. Allocating a large (> 16k bytes) buffer on the stack would cause this, but I doubt that's what it's doing.

Once a stack overflow is detected, the system should be regraded as toast, and rebooted. Thus, the 'cat' problem following the stack overflow could have messed things up worse.

I'll take a look at it. Any more test results would be appreciated, especially if stack overflow occurs after a command besides grep (after a reboot).

Mellvik:
Is it possible that the wildcard expansion is overflowing the shell's cmd line buffer?

-M

ghaerr:
> Is it possible that the wildcard expansion is overflowing the shell's cmd line buffer?

Anything is possible, but the 'cat' command was executed after grep corrupted the stack. The system didn't reboot, the shell died, that's why you got the login prompt. So we can't look at the cat issue as a problem, unless it happens after a reboot and grep never run.

Mellvik:
Test - FWIW: Definitely not the shell wildcard expansion.

PR #591, opened 2020-04-26
Are we able to allocate more then (ca.) 700b bytes on stack?

marcin-laszewski:


ghaerr:
Hello @marcin-laszewski,

The answer is no, definitely not. In linux kernel programming, it is considered bad practice to allocate data buffers on the stack, instead system data structures are used. One must be very careful to use as little stack as possible, or risk hard-to-find system crashes.

For ELKS, there are three possible stacks that may be switched to at any time: The application stack, which is shared with the application heap area, and is set to 16K bytes by default, with the heap allowed to grow until only 4K stack is available. Then there's the kernel stack, used during system calls and most of the time kernel routines are run, which is 988 bytes total and declared in elks/linuxmt/sched.h. And there's the interrupt stack, in elks/arch/i86/kernel/irqtab.S which is 512 bytes total, which is switched to immediately upon any hardware interrupt, including nested ones, and also has to handle any kernel routine called during interrupt processing.

marcin-laszewski:
Thank you very much for your comprehensive explanation.
4K buffer is a must. How to obtain it: heap_alloc, seg_alloc, other?

ghaerr:
heap_alloc will allocate off the local kernel heap, which is kernel DS addressable, but is in extremely short supply. For playing around with your first attempt, that will be easiest, but not acceptable for any final version. seg_alloc will allocate any amount from main memory, but is not directly addressable.  The fmemcpyb functions or similar can be used to access it. I still suggest looking/learning at the filesystem map/unmpp_buffer code and pipe to see how this should be properly handled. 

Perhaps its easiest though getting it running in the arrangement you currently have, then converting to a sustainable design.

marcin-laszewski:
Can't we allocate 4kB in kernel? Linux offers 'kmalloc', 'vmalloc'. I can split my 4kB in pieces so e.g 8 buffers of 512 bytes is not a problem.

ghaerr:
Yes, seg_alloc can allocate 4kB, it will be allocated in global memory. 4kB may be able to be allocated in the near (local) heap. `meminfo` now shows kernel memory allocations. Please re-read my previous comment for details. Kernel programming requires a very structured approach, as the conditions under which the code operates is very different than user programs. IMO, the best way to learn is to read lots of kernel code and try to understand what it is doing.

marcin-laszewski:
Yes, I have started. This is my first lesson. I will deliver my code and we can close this issue now.

Issue #586, opened 2020-04-25
Regression? Boot stopped after "VFS: Mounted root (minix filesystem)"


pawosm-arm:
I wasn't building 'master' for last few days, so it must have been introduced by one of very recent changes.

pawosm-arm:
Right, master updated again today. Let me check if it's any better.

pawosm-arm:
Unfortunately, it's still stuck right after VFS.

ghaerr:
Can you give more details, perhaps a photo/screenshot? Is this on real hardware?

There have been changes to the serial port code, and the /etc/rc.d/rc.sys script is run immediately after VFS mount. It could be dying running that script, which starts "Running /etc/rc.d/rc.sys".

Are the serial ports identified on the boot screen?

`init` is running at that point, may need to turn on init debugging.

ghaerr:
More messages about exactly what is happening can be output by editing 'elkscmd/sys_utils/init.c' and setting DEBUG 1 on line 40, then remaking and rebooting.

pawosm-arm:
I don't get it. I changed the console to 'serial' to get some logs for you, and it booted!

There's one thing I need to check though. In order to connect null-modem cable, I had to disconnect serial mouse. Maybe it should not be connected at boot time?!

ghaerr:
> I had to disconnect serial mouse.

The default /etc/inittab now runs getty on /dev/ttyS0. That means that getty will interact with a mouse if there's one on it. Its possible something strange is happening there. Turn on init DEBUG as described above and we can see exactly what's happening.

It is possible that the mouse interaction is an issue, but init shouldn't be running getty until AFTER the rc.sys script. So we need to see the debug output from init.

You can also set the default init level to 1 in elkscmd/rootfs_template/etc/inittab by changing the first line to 'id:1:initdefault', which will boot into level 1 rather than 3. This will keep the serial ports out of the picture.

pawosm-arm:
Nah, my suspicion was wrong. With nothing connected to serial port it didn't boot either. Also booting with null modem cable connected and minicom started on the other side did not help. Booting stuck after VFS and nothing appeared in minicom (set to 9600bps).
I'll try to build with DEBUG.

pawosm-arm:
This change did not reveal anything useful:
```
diff --git a/elkscmd/rootfs_template/etc/inittab b/elkscmd/rootfs_template/etc/inittab
index a4193446..18476c44 100644
[...]

Issue #539, opened 2020-04-11
ktcp should provide ability to use hardware flow control on SLIP connections


pawosm-arm:
According to https://github.com/jbruchon/elks/blob/master/Documentation/text/networking.txt in order to establish SLIP connection between ELKS and Linux one needs to type:
```
slattach -p slip -L -s 9600 /dev/ttyS1 &
```
The `-L` option means `Enable 3 wire operation. The terminal is moved into CLOCAL mode, carrier watching is disabled.`. I assume this is due to `ktcp`'s inability to specify (and utilize?) hardware flow control. Lack of hardware flow control makes serial connection unusable.

ghaerr:
> I assume this is due to ktcp's inability to specify (and utilize?) hardware flow control. 

Yes. `ktcp` can't utilize hardware flow control because the kernel doesn't support it. At this point, neither hardware flow control nor UART FIFO buffering is implemented. Both could be added, but we need to figure out some basic problems in the serial driver first. I plan on implementing FIFO buffering immediately after understanding exactly what the problem is in the current serial driver.

My latest push on the open serial port PR #535 will display errors (finally) by the UART on the console, including framing error (received data overrun), and interrupt signaled without receive data ready (QEMU seems buggy in this respect). I think we're making progress, but I can't personally test on real hardware, so thank you for your testing.

> Lack of hardware flow control makes serial connection unusable.

I thought some users had been using serial TCP/IP already? Can you test it, or is this just a request that ELKS needs it?

This issue should likely be renamed "Provide serial port hardware flow control [enhancement]", as `ktcp` should default to using hardware flow control unless -L is specified.

Mellvik:
Hi @pawosm-arm, @ghaerr,

> Lack of hardware flow control makes serial connection unusable.
> 
This is not the case - for a number of reasons. HW flow control is useful and increases effective bandwidth for obvious reasons, but SLIP will work fine with a reasonable amount of character loss (and subsequent packet retransmissions). Thus the option to turn off that requirement from the command line.
In fact, as is today, SLIP will probably work fine in ELKS @ 9600 bps, where the character loss is minimal. 

Also remember that modern day serial lines hardly ever have any flow control at all, just TTL level RX, TX, GND - possibly augmented by a 232MAX chip to yank up the voltages to RS232 levels. The assumption seems to be that the hardware is fast enough to keep up with the 115200 bps speeds and higher anyway.
[...]

PR #592, opened 2020-04-27
[cmds] Ktcp enhancements and various small cleanups

ghaerr:
Add '-s baudrate' option to `ktcp`, requested in #539. Modified /etc/rc.d/rc.sys, no need to run `stty` separately beforehand anymore.

Set new VMIN/VTIME serial line parameters for `ktcp` slip testing for @pawosm-arm, from #515.

Fixed lots of compiler warnings in ktcp, many more still there.

Add process ID to kernel `sys_brk` process out of memory errmsg, discussed off-topic in #586.

Temporary change to `telnetd` to exec /bin/login rather than /bin/sh for @pawosm-arm testing.

Quick one-line guess/fix to ash/mkinit.c to try to debug the occasional GitHub CI build errors that are seen just after merges/push to GitHub. This is the 'invalid character in file' that is generated and stops the build (see picture). 

mkinit

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -227,8 +227,8 @@ int sys_brk(__pptr newbrk)
 
     if (currentp->t_begstack > currentp->t_endbrk) {	/* Old format? */
         if (newbrk > currentp->t_endseg - MIN_STACK_SIZE) {	/* Yes */
-			printk("sys_brk failed: brk %x > endseg %x\n",
-				newbrk, currentp->t_endseg - MIN_STACK_SIZE);
+			printk("sys_brk(%d) fail: brk %x over by %d bytes\n",
+				currentp->pid, newbrk, newbrk - (currentp->t_endseg - MIN_STACK_SIZE));
             return -ENOMEM;
 		}
     }
--- a/elkscmd/ash/mkinit.c
+++ b/elkscmd/ash/mkinit.c
@@ -343,7 +343,7 @@ void dodecl(char *line1, FILE *fp)
 		if (! amiddecls)
 			addchar('\n', &decls);
 		q = NULL;
-		for (p = line1 + 6 ; *p != '=' && *p != '/' ; p++);
+		for (p = line1 + 6 ; *p != '=' && *p != '/' && *p != '\0'; p++);
 		if (*p == '=') {		/* eliminate initialization */
 			for (q = p ; *q && *q != ';' ; q++);
 			if (*q == '\0')
--- a/elkscmd/inet/telnetd/telnetd.c
+++ b/elkscmd/inet/telnetd/telnetd.c
@@ -25,7 +25,8 @@ static char buf_out [MAX_BUFFER];
 
 static int tfd, tfs;
 pid_t pid;
-char * nargv[2] = {"/bin/sh", NULL};
+//char * nargv[2] = {"/bin/sh", NULL};
+char * nargv[2] = {"/bin/login", NULL};
 
 void sigchild(int signo)
 {
@@ -37,7 +38,6 @@ int term_init()
 {
 	char pty_name[12];
 	int n = 0;
-	int rc;
 	
 	struct termios slave_orig_term_settings; // Saved terminal settings
 	struct termios new_term_settings; // Current terminal settings
@@ -61,7 +61,7 @@ again:
 	}
 	if (pid>0) {
 		// Save the default parameters of the slave side of the PTY - unused yet
-		rc = tcgetattr(tfs, &slave_orig_term_settings);
+		tcgetattr(tfs, &slave_orig_term_settings);
 		new_term_settings = slave_orig_term_settings;
 		// Set raw mode on the slave side of the PTY - not line oriented
 		//cfmakeraw (&new_term_settings);
@@ -222,6 +222,7 @@ int main(int argc, char *argv[]) {
 	dup2(ret, 2);
 	if (ret > 2)
 		close(ret);
+	setsid();
 
 	while (1) {
 		connectionfd = accept (sockfd, (struct sockaddr *) NULL, NULL);
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -9,7 +9,10 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/select.h>
 #include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
 
 #include <linuxmt/arpa/inet.h>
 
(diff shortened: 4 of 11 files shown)


pawosm-arm:
The new `-s` ktcp option looks ok, I'm not sure about the other changes though (~ECHOE, VMIN, VTIME), a telnet connection from ELKS to custom-made telnet server running on Linux seems extremely slow, no matter what speed I tried (up to 4800, as for 9600 the telnet connection could not even be established).
I couldn't benefit from IDs printed along with `sys_brk` failure message: changing `sh` to `login` solved the problem (so I guess it wasn't `telnetd` to blame). Unfortunately, I couldn't experience Linux to ELKS connection: although `login` process was spawned (as seen by `ps` output on ELKS), nothing appeared in telnet client on Linux. I suspect serial communication issue.

Generally, this (already merged) PR gives some benefit and addresses the matter of this ticket, yet underlying issue of interrupt-driven serial communication remains.

ghaerr:
> I'm not sure about the other changes though (~ECHOE, VMIN, VTIME)

Those changes will help, they don't change the serial port directly but allow larger data counts per read to be passed back to ktcp.

> a telnet connection from ELKS to custom-made telnet server running on Linux seems extremely slow, no matter what speed I tried (up to 4800, as for 9600 the telnet connection could not even be established).

It appears your original issue is not solved. Did this fix allow your system to work faster than 300 baud? I am interested in results of any improvement. The kernel will now display serial overruns, so that is not apparently your problem.

>  changing sh to login solved the problem (so I guess it wasn't telnetd to blame).

Good to know. I will be able to debug sh vs login when I am able to get QEMU setup for networking, I need help on that.

> Unfortunately, I couldn't experience Linux to ELKS connection: although login process was spawned (as seen by ps output on ELKS), nothing appeared in telnet client on Linux. I suspect serial communication issue.

I see. Did this not work at 300 baud either?

> Generally, this (already merged) PR gives some benefit and addresses the matter of this ticket, yet underlying issue of interrupt-driven serial communication remains

Thanks. Agreed.

How do we know that your system UART chip in fact works correctly? Does this repeat on other machines? I'm sorry if you've already explained, but am just wondering about hardware possibilities at this point. We haven't got the full test results from @Mellvik on the improved `sercat` and serial handling yet, which is important.

pawosm-arm:
> Did this fix allow your system to work faster than 300 baud?

Nah, it was possible before to go as fast as 4800bps, but the quality of service was deteriorating with growing speed. It's just that 300bps allows to sustain established telnet communication for longer. Note that many times it was just a matter of luck how long the communication worked before stalled, and this didn't change. The only improvement I see is that deterioration (at faster speeds) less often cause the system to hang and even with connection stuck, it's still keyboard-responsible.

> How do we know that your system UART chip in fact works correctly?

It works on FreeDOS at all speeds available. I'm either redirecting FreeDOS console to serial (using `ctty` command) or use `Telix` which allows to connect at greater speeds (it is possible since Telix makes use of hardware control flow lines, and a full hardware control null-modem cable is the only one I have here). Yet it's easy with DOS: it's a single process system, hence the communication program is never preempted.
Also, I've installed Windows 3.0 on top of FreeDOS (the last Windows version one could install on 8086/8088, the next version, Windows 3.1, required 80286 at least), it has the HyperTerminal program. Although its top speed is limited to 19200, it also offers stable communication with the other side.

ghaerr:
> It was possible before to go as fast as 4800bps, but the quality of service was deteriorating with growing speed. 

So previous to all the serial mods in the last weeks, it worked better? Dang!

> The only improvement I see is that deterioration (at faster speeds) less often cause the system to hang and even with connection stuck, it's still keyboard-responsible.

The recent changes have helped the system in general, I think, and are needed for higher throughput. But none of these changes seem related to the core issue with your machine.

> It works on FreeDOS at all speeds available.

Great news, its a software problem! :) Time to look at the FreeDOS serial driver.

> (it is possible since Telix makes use of hardware control flow lines, and a full hardware control null-modem cable is the only one I have here).

Oh. Maybe it isn't a software only problem - hardware flow control is needed for high speeds on your machine? We'll wait for @Mellvik's driver, since I don't (yet) have hardware to easily test with.

Can you confirm that you get high-speed data INPUT on FreeDOS without using the hardware null-modem cable? We're not using interrupt-driven serial output (yet).

Thanks for your testing!

pawosm-arm:
> Time to look at the FreeDOS serial driver.

I don't think it's the right stuff to examine. DOS serial port is hard to deal with (even its 'mode' command to configure communication parameters doesn't expose all the options both Telix and HyperTerminal have). Most of the DOS programs talk to the hardware directly (bypassing DOS) and have ability to provide ioports and IRQ.

> Can you confirm that you get high-speed data INPUT on FreeDOS without using the hardware null-modem cable? We're not using interrupt-driven serial output (yet).

As I said, I don't have any other null-modem cable. Yet I managed to do some test with a DOS software that definitely can't handle hardware flow control. Up to speed 2400bps it's fine and stable. At 4800bps the inbound traffic (a longer piece of text pasted to `minicom` on the Linux side) started to suffer from minor character loses. That only confirms, not using hardware flow control is a bad idea...

Mellvik:
IMO the serial driver may not be the primary culprit here. During my extensive testing over the past weeks, I have noticed (and some times reported) a system hang situation that severely affect serial reliability. More than half of the time when I dump my 12kb test file onto the serial line, one buffer (512b) or less actually make it to the destination file. The line speed makes a difference, but not much.

With the current serial driver and what I like to call normal system activity, ELKS should easily handle 2400 bps w/o character loss even if there is no flow control, probably 4800 too. [Remember - with careful preparation(!) and no disturbances (whatever that is), we're able to get all of 12k through - w/only 12 characters lost. BTW - that is 12 characters every time, one per KB.] 
So while continuing the work on serial improvements, I think we need to be on the outlook for 'the other culprit'. I have no idea where to look. It is quite possible that The new sercat tool will help narrow it down.
I plan to start working on the flow control issues toward the end of the week - possibly starting with soft flow control (XON/XOFF). In the meanwhile I'll be mostly offline.

Apropos drivers to look at - Minix 2 seems like a good place to start. 

-M

> ﻿
> Time to look at the FreeDOS serial driver.
> 
> I don't think it's the right stuff to examine. DOS serial port is hard to deal with (even its 'mode' command to configure communication parameters doesn't expose all the options both Telix and HyperTerminal have). Most of the programs talk to the hardware directly and have ability to provide ioports and IRQ.
> 
> Can you confirm that you get high-speed data INPUT on FreeDOS without using the hardware null-modem cable? We're not using interrupt-driven serial output (yet).
> 
> As I said, I don't have any other null-modem cable. Yet I managed to do some test with a DOS software that definitely can't handle hardware flow control. Up to speed 2400bps it's fine and stable. At 4800bps the inbound traffic (a longer piece of text pasted to minicom on the Linux side) started to suffer from minor character loses. That only confirms, not using hardware flow control is a bad idea...
> 
> —
>

PR #594, opened 2020-04-27
Fixed time command, cleanups in sash command history.

Mellvik:
time(1) now displays real time (and only real time) correctly. 

sash command history cleanup: compiler warnings, static declarations (and a bug).

### DIFF
--- a/elkscmd/misc_utils/time.c
+++ b/elkscmd/misc_utils/time.c
@@ -8,29 +8,21 @@
 #include <sys/times.h>
 #include <sys/wait.h>
 
-static void printt(char * s, long se, long us)
+static void printt(char * s, long us)
 {
 	long mins, secs;
 
-	if (se == -1 ) {		/* microsecs only */
-		if (us < 1000L)		/* round up to 1/1000 second*/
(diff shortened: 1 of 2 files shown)


ghaerr:
Thanks! Are you going half-crazy yet with the time calculations? I think I spent too much time on this originally and still didn't get it right. Glad you've got it working, and it seems the troubles were all in the print routine.

Mellvik:
I am (going half-crazy that is). And yes, everything was indeed in printt. Incredible how complicated the simplest things are when you cannot get your thinking straight.
I probably spent two hours trying to figure out why I was getting an occasional one second difference between my gettimeofday() data and your times() data. Never mind, now it's correct.

I guess this is certainly not the time (!) to introduce clock tics into the task struct. BTW - It was fun looking up what it looked like in BSD4.2:

myname# !364
time cat /dev/ra0a > /dev/null 
3.4u 0.2s 0:03 100% 10+14k 3976+3io 1pf+0w

Clock time < sys+user time, hmmmm - so I guess we're in good company (may actually be SIMH's fault …)

—Mellvik

PR #596, opened 2020-04-28
Implement POSIX style basename(3) and dirname(3) in libc 

cjsthompson:
This is based on the code of the basename and dirname sh_utils utilities. Update these aforementioned utilities as well as their busyelks counterparts accordingly.


ghaerr:
Nice work Christoph! Your attention to detail is nice to see.

PR #599, opened 2020-04-29
[serial] Read data until no interrupt flag

ghaerr:
Hello @pawosm-arm,

This PR causes serial data to be read in the interrupt handler until the hardware interrupt flag is reset, rather than the previous way which reads data until the read data available bit is reset. I would like you to test with this code compiled in the kernel, and see whether the mouse still locks up your system.

From reading the MINIX 2.0.4 serial driver code (https://github.com/chenshuo/old-minix/blob/master/src/kernel/rs232.c), it is mentioned that some chips lock up if this is not done. (see the serial interrupt routine rs232_handler()).

@Mellvik, this should not affect the FIFO operation, but I am unable to test that. I am assuming that the FIFO hardware continues to assert the interrupt pending flag in the IID while FIFO data is available. If not, then this would have the effect of only reading one character per interrupt.

### DIFF
--- a/elks/arch/i86/drivers/char/serial.c
+++ b/elks/arch/i86/drivers/char/serial.c
@@ -223,9 +223,9 @@ void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
 	//if (status & UART_LSR_DR)			/* Receiver buffer full? */
 	    do {
 		buf[i++] = inb_p(io + UART_RX);		/* Read received data */
-	    } while ((inb_p(io + UART_LSR) & UART_LSR_DR) && i < MAX_RX_BUFFER_SIZE);
+	    //} while ((inb_p(io + UART_LSR) & UART_LSR_DR) && i < MAX_RX_BUFFER_SIZE);
 	//}
-    //} while (!(inb_p(io + UART_IIR) & UART_IIR_NO_INT) && i < MAX_RX_BUFFER_SIZE);
+    } while (!(inb_p(io + UART_IIR) & UART_IIR_NO_INT) && i < MAX_RX_BUFFER_SIZE);
 
     if (status & UART_LSR_OE)
 	printk("serial: data overrun\n");


Mellvik:
@ghaerr,
It may make sense to test this, but I read the comment in the minix source differently (or I missed something).
The 16550 application note says approximately the same thing, check all interrupts, not because there are more bytes to read (which is impossible anyway on a 8250), but because there are many sources. If those sources are not enabled, as in our case (for now), It is indeed a waste of time like the comment says. 
Btw, I found the 16550 app note I referred to before, on line if you're interested...

-Mellvik

ghaerr:
> It may make sense to test this

We have not found the base reason for @pawosm-arm's problems, and we are handling the interrupt differently as far as which HW registers are read, so I'd like to test this.

>I found the 16550 app note I referred to before, on line if you're interested... 

Sure, go ahead and post it.

> check all interrupts, not because there are more bytes to read (which is impossible anyway on a 8250)

The ELKS interrupt routine isn't the same as MINIX, as there is only one interrupt enabled. The difference this code tries is to read bytes while the RDA interrupt is set in the IIR, rather than reading bytes while the RDA bit is set in the LSR. This is the same path the MINIX driver takes.

pawosm-arm:
Hi @ghaerr,

I'm not sure how to summary this. The `ktcp` issues prevail, but I doubt this patch was to solve them (I assume, as long as `ktcp` isn't making use of hardware flow control, requiring `slattach` to be started with `-L` to tell it to work without hardware flow control, this will always choke).
What I observed is that the serial communication problems aren't quiet anymore. It is like this for some time, so I could observe "Buffer overrun" for some time, and such message does not make system hung. The new thing though is that after applying this PR, I'm seeing new message not seen before: "panic: Unable to get timer", and the system hangs totally so even Ctrl+Alt+Del isn't working, despite the on-screen message encouraging to do it. I wonder if this adds any clue to our investigation.

ghaerr:
Hello @pawosm-arm, thank you again for your testing.

Unfortunately my guess at fixing our serial code by following MINIX did not work at all. While it didn't fail on QEMU that's obviously not the case on your hardware. It might be interesting to see whether this crashes the system on @Mellvik's hardware, to help gather more data about your specific hardware problems (or not), but there's no point in considering this patch for committing.

I will have to wait until I get access to serial hardware to continue writing code to improve the serial driver, I'm afraid. 

> after applying this PR, I'm seeing new message not seen before: "panic: Unable to get timer", and the system hangs totally 

Considering that the panic message mentioned is from code that is executed extremely early in the ELKS boot process, I can only gather that the reason for this is that the system was completely crashed and happened to execute that code, out of control. My guess is that a kernel interrupt stack overflow is happening, which could cause this and all your crashes. Given this never happens but on your machine, my guess is it might be related to the hardware-initiated IRQ processing occurring on top of each other, until stack overflow, but just a guess.

> I could observe "Buffer overrun" for some time

Guessing again that this means the new serial interrupt handler isn't accessing/resetting your UART hardware properly, and the chip is reporting data overruns. MINIX source comments that some serial cards behave differently (see @Mellvik's comment above about interpretation of the MINIX source comment). All-in-all, though, rather than being a "waste of processing time", this patch flat doesn't work.

> I assume, as long as ktcp isn't making use of hardware flow control, requiring slattach to be started with -L to tell it to work without hardware flow control, this will always choke).

I had hoped to fix the core serial driver issue(s) to handle high speed input without flow control. I am still interested in the `sercat` testing results, which will confirm or not whether the throughput changes recently made actually help. It looks at this point it might be prudent to stop working on software serial speed enhancements and go directly to hardware flow control implementation.

pawosm-arm:
> Considering that the panic message mentioned is from code that is executed extremely early in the ELKS boot process...

This panic occurred long after the system was fully booted and doing serial communication initiated from userspace. So the kernel went massively out of control. A kernel interrupt stack overflow seems likely cause of such behaviour.

Issue #600, opened 2020-04-30
Enhancement: add new miniterm option '-l = add LF to every CR received'


pawosm-arm:
This option is present in most of serial port terminal programs and I've found it extremely useful (e.g. when sending copy-pasted blocks of text, LF's are never sent to the other side, so the receiving side should add LF before every CR to prevent screen being messy). It would be great to have it in `miniterm` too. The logic is simple. With this option specified: for every 'CR' received, cursor should go to the next line (as if LF was received) before returning to the beginning of the line (which is the result of CR).

ghaerr:
Hello @pawosm-arm,

Would you like to take a pass at implementing this, since you'll also be able to test it?

pawosm-arm:
Sure, I'll look into it.

pawosm-arm:
Putting this aside, I could for a first time observe how miniterm is built:

```
miniterm.c: In function 'main':
miniterm.c:389:27: warning: variable 'enable_dtr' set but not used [-Wunused-but-set-variable]
  bool enable_rts = false, enable_dtr = false;
                           ^~~~~~~~~~
miniterm.c:389:7: warning: variable 'enable_rts' set but not used [-Wunused-but-set-variable]
  bool enable_rts = false, enable_dtr = false;
       ^~~~~~~~~~
```
Looks like miniterm does nothing about RTS and DTR...

ghaerr:
> Looks like miniterm does nothing about RTS and DTR...

Correct, those options are controlled by -d and -R. The plan is to use the -r option to turn on hw flow control by unsetting CLOCAL and setting CRTSCTS. The serial driver will then enable DTR and use the RTS line to control incoming data.

None of the options have ever been operational in ELKS, so -r currently does nothing as well.

pawosm-arm:
merged.

PR #601, opened 2020-04-30
Enhancement: add new miniterm option '-l = add LF to every CR received' (#600)

pawosm-arm:
This patch adds a new command line option as well as new control keys
to control reaction to every received CR key.

This is to address #600.

### DIFF
--- a/elkscmd/misc_utils/miniterm.c
+++ b/elkscmd/misc_utils/miniterm.c
@@ -70,6 +70,8 @@ typedef int bool;
 #define DTR_OFF_CHARACTER 'd'	    /**< DTR off character */
 #define RTS_ON_CHARACTER 'R'	    /**< RTS on character */
 #define RTS_OFF_CHARACTER 'r'	    /**< RTS off character */
+#define CRLF_ON_CHARACTER 'L'	    /**< CR+LF on character */
+#define CRLF_OFF_CHARACTER 'l'	    /**< CR+LF off character */
 #define EXIT_CHARACTER '.'	    /**< Exit character */
 #define EXIT_NO_RST_CHARACTER 'x'   /**< Exit with no reset character */
 
@@ -341,6 +343,7 @@ static void usage(const char *argv0)
 		"  -r          enable RTS/CTS hardware flow control (default: disable)\n"
 		"  -d          enable DTR (default: disable)\n"
 		"  -R          enable RTS when flow control disabled (default: disable)\n"
+		"  -l          add LF to every CR received (default: disable)\n"
 		"  -x          print received data in hex (read-only)\n"
 		"  -S          print received data as SLIP packets (read-only)\n"
 		"  -h          print this message\n"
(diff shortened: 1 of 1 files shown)


ghaerr:
Everything looks quite nice, except for the portion that actually converts CR's to CRLFs:

The original for loop tries to write up-to-4096 bytes that have been just read by writing multiple times, should the write not accept all 'len' bytes. Your code only checks the first byte of a write  buffer to be CR, while there could be many. Thus, it really only works when a single CR is received. In addition, 3 write system calls are used every time a CR is received, which is very inefficient.

May I suggest the following: when add_lf is true, convert the entire buffer before the for loop with something like:
```
char *wbuf;
char buf[1024];   /* 4096 input buffer is too large, could cause stack problems*/
char obuf[2048]; /* CRLF output buffer*/
...
if (add_lf) {
   len = convCR(buf, len, obuf);   /* convert all CR's to CRLF and return new buffer count*/
   wbuf = obuf;
} else wbuf = buf;
```
and then use 'wbuf + wrote' in the subsequent write.

pawosm-arm:
I wasn't sure if we're happy with extra memory use.
I'll change it later today.

pawosm-arm:
> Your code only checks the first byte of a write buffer to be CR, while there could be many.

Nah, I'm going byte by byte, so all CR's should be spotted.

The last param of write() determines how much I go...
As I said, I wanted to do avoid the need for extra memory.

ghaerr:
> The last param of write() determines how much I go...

Got it, I missed that. Single-character output using write is very inefficient, especially to the console. It would be interesting to test whether this has a noticeable processing effect on ELKS, with fast serial input. I realize that ELKS can't currently handle fast serial input. Should that finally get fixed, then this bottleneck risks a possibility of having the TTY queue overflow and discarding characters. 

> As I said, I wanted to do avoid the need for extra memory.

The extra memory is on the stack, so it's already in use and costs nothing. That said, the 4096 buffer is overkill, and I just fixed a stack overflow bug in sash grep that allocated 8192 bytes on the stack - too much, when the ELKS default is 16k for stack and heap, and the program uses heap storage.

Ultimately its probably not a big deal either way, I'm just commenting on it. I'll commit this with whichever way you choose.

pawosm-arm:
Hi @georgp24 

I came up with different solution: using memchr() for looking for CR key and splitting write operation only when needed. Worked well enough on my XT.
Also, to make things easier, LF is added after CR, not before, hence I renamed all LFCR and LF+CR to CRLF and CR+LF respectively. The option name is still 'l' for line-feed that is added.

Updated PR is available now.

georgp24:
Paul,
I assume you meant Greg not me?
Georg

ghaerr:
Paul - nicely done.

It will be impossible for ELKS to ever return more than the TTY queue buffer size (which was just increased from 512 to 1200) in a single read. Do you want to adjust the 4096 buffer down to 1200, thus saving miniterm stack usage, or just leave it as is?

pawosm-arm:
> Paul,
> I assume you meant Greg not me?
> Georg
> 

Hey, auto-completion failed me

pawosm-arm:
> Do you want to adjust the 4096 buffer down to 1200, thus saving miniterm stack usage, or just leave it as is?

It's not a yes/no question. Sadly, buffers of 4096 bytes size are defined in many places in `miniterm.c`. I guess, `#define` constant should be defined for this and all those places refined. It's outside of this ticket's scope...

ghaerr:
> It's outside of this ticket's scope...

No problem, I haven't really looked at miniterm that much other than fixing the baud rate a while ago. Not surprised there's all sorts of things that can be made better. Thanks for your work on this!

PR #602, opened 2020-05-01
Get the screen window manager to compile and run

georgp24:
Modified the screen program so it will compile now and run on ELKS

### DIFF
--- a/Documentation/Readme.md
+++ b/Documentation/Readme.md
@@ -15,6 +15,8 @@ This section contains some HOWTOs to help using ELKS.
 
 * [Shell command line editing and Tab completion](https://raw.githubusercontent.com/jbruchon/elks/master/Documentation/html/user/shell.html)
 
+* [The screen window manager](https://raw.githubusercontent.com/jbruchon/elks/master/Documentation/html/user/screen.html)
+
 * [Writing applications for ELKS in C](https://htmlpreview.github.io/?https://github.com/jbruchon/elks/blob/master/Documentation/html/user/writing_apps_in_C.html)
 
 * [Writing applications for ELKS in assembler](https://htmlpreview.github.io/?https://github.com/jbruchon/elks/blob/master/Documentation/html/user/writing_apps_in_assembler.html)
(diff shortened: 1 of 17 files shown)


ghaerr:
Nice work Georg!!! And complete with documentation, thank you.


A couple questions and comments:

- In screen.c, is there are problem using the FD_SET etc defines that are also defined in time.h, I see you're having to \#undef them. These should be already in a libc or kernel header file. If they are broken, then they should be fixed in those header files, depending on what the issue is.

- Not a big deal, but in ansi.c, lots of places where a 'return 0' has been added, with no other return possibilities. These would best be handled as declaring the function type 'void', rather than returning a value that is never checked. It appears screen was written prior to void functions.

- Since you're updating the dates in various places, do you think the version number might want to be bumped as well? And possibly make them match.

Great job cleaning up this one, looks like it was a lot of work!

georgp24:
Greg,
this screen code is based on screen version 2.0a. Apparently for ELKS the version was extended to 2.0a.1. So I updated this to 2.0a.2 now.

In this old code fd_set was defined as a structure of two integers and therefore I could not use the definitions in the standard time.h header. I now used fd_set as defined in the header file and that works as well. So I could remove the definitions of FD_SET etc. in the screen.c code.

Yes, I also thought `void` would be better than a `return 0` but it were so many functions that I moved on to other issues.

It was a lot of work just to get it to compile and then to get it to run too.
Georg

ghaerr:
Thanks Georg for the FD_SET changes. Did you test that and is it ready to commit now?

georgp24:
Yes, I tested it and you can commit it.

ghaerr:
Georg, I just noticed one other thing: POLL_MAX was increased from 4 to 16, which affects the size of every process task structure. What was the reason for this? Does screen requiring polling that many file descriptors in select?

georgp24:
I had to increase this limit to get screen to run. I thought programs would only use as many as they needed and screen needed more than 4.
Just tested it. The current version is fine with POLL_MAX set to 4. It has been over a week ago that I made this change but it helped me to get it to run then.

ghaerr:
> I thought programs would only use as many as they needed and screen needed more than 4.

Yes, programs use only as many as they need, but the POLL_MAX requires static room in every task entry. I'll commit this now, but if you find out how many file descriptors actually needs, then we can set POLL_MAX to that value, since no other programs use that many, in an effort to keep task memory space down.

PR #603, opened 2020-05-02
Reduce POLL_MAX in limits.h file to 6.

georgp24:
When you open four windows with screen and enter `ps` in the last window you get the message:
`panic: select_wait: no slot left` and you need to reboot. When `POLL_MAX` is set to `6` this does not happen. 
However, should the number of `ptyp`s be extended in ELKS some day, the problem will arise again.

Also modified screen documentation file and added screen to i86/defconfig.


georgp24:
I have a question: I added screen to the `elkscmd/config.in` and the `elks/arch/i86/defconfig` files setting it to `y`. However, when I execute `make menuconfig` there is still no `*` marked in the `screen` selection line in `userland`. What is missing for that?

ghaerr:
Georg, thanks for the testing to determine the actual need for POLL_MAX.

> However, when I execute make menuconfig there is still no * marked in the screen selection line in userland. What is missing for that?

Could be that you didn't remove .config first. "make defconfig" should work, try that. It seems that "make config" and make "menuconfig" operate slightly differently.

georgp24:
When I run `make menuconfig` all the items that have a "y" in the `elkscmd/config.in` and `elks/arch/i86/defconfig` files show a `*` in the menu. Just screen does not. I can select it in menuconfig and it will be build and included in the disk image. But I would expect that it is marked with a star as the default setting when I set it to `y` in the two mentioned files.
I did remove the `.config` file in the elks top directory before running `make menuconfig`.

ghaerr:
I agree that it is seems incorrect, but I'm not a config expert. I find that I have to `rm .config; make defconfig` whenever new configuration items are added, and then make menuconfig or config after that.

PR #604, opened 2020-05-02
Bugfixes in sash, libc/putenv.

Mellvik:
- process sash startupfile(s) (set environment variables etc.) before initializing history and running scripts.
- fixed bug in history command search and added check for buffer overflow.
- setting environment variable HISTORY to zero now REALLY turns off history in sash.
- fixed bug in libc/putenv.c that allowed environment variables w/o value to be added to the environment, multiple times.

### DIFF
--- a/elkscmd/sash/cmd_history.c
+++ b/elkscmd/sash/cmd_history.c
@@ -32,13 +32,16 @@ init_hist() {
                 if (histcnt > HISTMAX)
                         histcnt = HISTMAX;
         }
+	if (!histcnt) return;
         histbuf = malloc(histcnt * sizeof(char*));
         if (!histbuf) {
                 fprintf(stderr, "No history buffer, history turned off\n");
                 histcnt = 0;
-        }
-        while (k < histcnt)     /* initialize history list */
-                histbuf[k++] = NULL;
+        } else {
+        	while (k < histcnt)     /* initialize history list */
+                	histbuf[k++] = NULL;
+	}
+	return;
 }
 
 
@@ -92,7 +95,7 @@ cmd_search(char * pat) {
 	for (i = 0; i < histcnt; i++) {
 		idx = map_ind(-i);
 		if ((found = strstr(histbuf[idx], pat)))
-			return(found);
+			return(histbuf[idx]);
 	}
 	return(NULL);
 }
@@ -202,10 +205,14 @@ fixbuf(char *cmd, char *prev, char *pos, char *cm) {
 
 	char buf[CMDBUF];
 
-	strcpy(buf, cmd);
-	strcat(buf, prev);
-	strcat(buf, pos);
-	strcpy(cm, buf);
+	if ((strlen(cmd) + strlen(prev) + strlen(pos)) > (CMDBUF -1)) 
+		fprintf(stderr, "Buffer overflow in command history \n");
+	else {
+		strcpy(buf, cmd);
+		strcat(buf, prev);
+		strcat(buf, pos);
+		strcpy(cm, buf);
+	}
 	return;
 }
 
--- a/elkscmd/sash/cmds.c
+++ b/elkscmd/sash/cmds.c
@@ -727,7 +727,9 @@ do_setenv(argc, argv)
 		strcat(buf, "=");
 		strcat(buf, argv[2]);
 	}
-	putenv(buf);
+	if (putenv(buf)) 
+		fprintf(stderr, "usage: setenv VAR=value - value is optional, '=' is required.\n");
+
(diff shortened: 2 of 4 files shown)


ghaerr:
Hi Helge, thanks for the fixes.

> fixed bug in libc/putenv.c that allowed environment variables w/o value to be added to the environment, multiple times.

Does this fix disallow sash's previous "setenv SDF" and require a value? How does one set an empty environment variable in sash now?

Mellvik:
.
> Does this fix disallow sash's previous "setenv SDF" and require a value? How does one set an empty environment variable in sash now? 
> 
Well - yes and no. Previously, setenv SDF would delete SDF if it existed already and had a value. Otherwise it would create it - with no value - even if it existed. So I decided to take the csh route (the origin of the setenv etc. commands), and disallow environment variables w/o value. Forgetting that the change may actually affect ash too. I have to look at that.

Right now putenv() (libc) returns errno 22 if there's no value, this is not picked up by the setenv command in sash. It's just silently ignored, which is not good...

I'll fix that in an updated PR. And I have no objections to valueless environment variables in sash btw. @ghaerr, what's your take?

--Mellvik

ghaerr:
> So I decided to take the csh route (the origin of the setenv etc. commands), and disallow environment variables w/o value. 

Sounds good. So sash should not accept "setenv SDF", and instead show usage.



> Forgetting that the change may actually affect ash too. I have to look at that.

Just test ash to make sure it works with the modified putenv. Ash allows "export SDF=" and should insert SDF= into the environment. `getenv` would return a pointer to a 0 length ('\0' terminated) string, not NULL, which I'm sure it already does. Before this modification, "export SDF" will insert SDF= into the environment, just the same as "export SDF=", so it's probably ok.

> Right now putenv() (libc) returns errno 22 if there's no value, 

That's proper behavior. `putenv` should accept strings like `SDF=`, and reject `SDF`.

However, putenv should include errno.h and set errno = EINVAL, don't use constants.

> this is not picked up by the setenv command in sash. It's just silently ignored, which is not good...

"setenv SDF" should print the usage line, IMO.

Mellvik:
Thanks @ghaerr,
I will update the PR. 

While at it, I'll add a '-c' option to sash for compatibility. Currently sash fails when run as a subshell from interactive programs (editors etc.) because all (most) such programs executes $SHELL -c <command>. 


--Mellvik

Mellvik:
sash -c is not included in this PR.

ghaerr:
Helge,

I don't think adding the putenv error message is correct, for two reasons: first, the usage of sash setenv is not 'setenv SDF=value', but 'setenv SDF value'. Run 'setenv' with no arguments or look in /etc/sashrc. Secondly, sash already has its own builtin argument error checking and usage display mechanism. This includes the required and min and max args. This mechanism should be used, rather than adding an unneeded error message from a library routine.

Mellvik:
@ghaerr,

> 
> the usage of sash setenv is not 'setenv SDF=value', but 'setenv SDF value'.
> 
Actually both work. I'll add error checking to setenv, keeping it in line with the csh style.
> sash already has its own builtin argument error checking and usage display mechanism.
> 
I was looking for that, and missed it.
> 
I'd appreciate if you could merge this PR and I'll put the changes into the next PR, with fixed '-c' and some additional fixes.


—Mellvik

ghaerr:
> Actually both work.

That's only because of the way that sash built the environment string with one argument, passed to putenv. Its a bug when the same mechanism is tried with two arguments.

>  I'll add error checking to setenv, keeping it in line with the csh style.

That's not needed, contributes to bloat, and means something different when minargs = maxargs = 2.

> I'd appreciate if you could merge this PR and I'll put the changes into the next PR

Ok. I hope you understand that my approach is to try to get all contributions to be be thoroughly tested, as a common practice. The problem is, that few test a new feature other than the originating programmer!

Mellvik:
@ghaerr,

> Ok. I hope you understand that my approach is to try to get all contributions to be be thoroughly tested, as a common practice. The problem is, that few test a new feature other than the originating programmer!
> 
I appreciate your intentions - and share the goal. Your critical reviews and comments are very useful - and important. 
Indeed there are too few of us, and less testing than desired. Still the rate of progress has been inspiring recently.

> I'll add error checking to setenv, keeping it in line with the csh style.
> 
> That's not needed, contributes to bloat, and means something different when minargs = maxargs = 2.
> 


SASH: I hadn't fully grasped the functionality of the command table. The error message in setenv is now simply a perror().
Awaiting the merge of your sash updates.

—Mellvik

PR #605, opened 2020-05-03
Fix a fair amount of Elvis warnings

cjsthompson:
Mostly "implicit declaration" warnings. Add appropriate headers to fix this. Also reformat some function declarations to ANSI C and some comments to fit in 80 columns.

One "implicit declaration" warning that is still left is for "tmpnam" whose declaration is missing in ELKS' headers. Not sure whether this has to go in libc or gcc headers. In Linux it's declared in stdio.h according to the man page.

Elvis still compiles and runs and edits in ELKS.

### DIFF
--- a/elkscmd/elvis/blk.c
+++ b/elkscmd/elvis/blk.c
@@ -13,6 +13,9 @@
  */
 
 #include "config.h"
+
+#include <unistd.h>
+
 #include "vi.h"
 
 #ifndef NBUFS
--- a/elkscmd/elvis/cmd1.c
+++ b/elkscmd/elvis/cmd1.c
@@ -13,9 +13,17 @@
  */
 
 #include "config.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
 #include <ctype.h>
+
 #include "vi.h"
 #include "regexp.h"
+#include "tmp.h"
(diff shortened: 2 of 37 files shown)


ghaerr:
Nice work Christoph! And elvis had a ton of warnings. There may be a day yet where ELKS application compilation warnings are no longer, and this is a big step forward!

> One "implicit declaration" warning that is still left is for "tmpnam" whose declaration is missing in ELKS' headers. Not sure whether this has to go in libc or gcc headers. In Linux it's declared in stdio.h according to the man page.

Since ELKS libc/misc has tmpnam.c, and that's being used, go ahead and add the tmpnam prototype to libc/include/stdio.h, and that'll be one less to worry about. I'll commit this after you push that additional fix.

A quick search shows tmpnam is used in elvis, sys_utils/passwd.c, and busyelks/cmd/passwd.c, and they include <stdio.h>.

cjsthompson:
Here it is. I've got more easy warning fixes. But I'd thought I'd group them by module instead of doing a massive pull request that is harder to review.

ghaerr:
Ok, I'll commit this now. In the future, we're not using the older "extern ..." in libc header files, instead just the straight prototype. You can fix extern char *tmpnam() in a future commit.

cjsthompson:
I will. Thanks.

PR #607, opened 2020-05-03
Fix many warnings in ktcp

cjsthompson:
Fix many "implicit declaration" and "unused variable" warnings in ktcp.

Comment out unused variables. Add new headers. Reorder the inclusion
of some headers. Declare tcpdevfd in files that use it and remove it
from ip.h. Make some functions that return nothing void.

Compiles. Untested. If it did work, it should probably be tested by
someone who has the setup to do so before merging.

### DIFF
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -20,6 +20,7 @@
 #include "ip.h"
 #include "arp.h"
 
+static int tcpdevfd;
 
 /* ARP operations */
 /* as big endian values */
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -17,10 +17,10 @@
 #include <linuxmt/arpa/inet.h>
 
 #include "deveth.h"
+#include "tcp.h"
+#include "tcpdev.h"
 #include "ip.h"
 #include "arp.h"
-
-static int tcpdevfd;
 
 /* ARP operations */
 /* as big endian values */
--- a/elkscmd/ktcp/arp.h
+++ b/elkscmd/ktcp/arp.h
@@ -33,6 +33,6 @@ void arp_cache_add (ipaddr_t ip_addr, eth_addr_t * eth_addr);
 int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr);
 
 void arp_proc (char * packet, int size);
-
+int arp_request(ipaddr_t ipaddress);
 
 #endif /* !ARP_H */
--- a/elkscmd/ktcp/deveth.c
+++ b/elkscmd/ktcp/deveth.c
@@ -19,7 +19,9 @@
 #include <arch/ioctl.h>
 
 #include "config.h"
+#include "tcp.h"
 #include "deveth.h"
+#include "ip.h"
 #include "tcpdev.h"
 #include "arp.h"
 
--- a/elkscmd/ktcp/icmp.c
+++ b/elkscmd/ktcp/icmp.c
@@ -13,8 +13,8 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include "slip.h"
-#include "icmp.h"
 #include "ip.h"
+#include "icmp.h"
 #include <linuxmt/arpa/inet.h>
 
 int icmp_init(void)
--- a/elkscmd/ktcp/icmp.h
+++ b/elkscmd/ktcp/icmp.h
@@ -3,4 +3,7 @@
 
 #define PROTO_ICMP	1
 
-#endif
+int icmp_init(void);
+void icmp_process(struct iphdr_s *iph, char *packet);
+
+#endif /* !ICMP_H */
--- a/elkscmd/ktcp/ip.c
+++ b/elkscmd/ktcp/ip.c
@@ -13,16 +13,19 @@
  * TODO : IP fragmentation and reassemply of fragmented IP packets
  */
 
+#include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
 
 #include "ip.h"
 #include "icmp.h"
+#include "slip.h"
 #include "tcp.h"
 #include "tcpdev.h"
 #include <linuxmt/arpa/inet.h>
 #include "deveth.h"
+#include "arp.h"
 
 #if 0
 #define IP_VERSION(s)	((s)->version_ihl>>4&0xf)
@@ -123,7 +126,8 @@ void ip_print(struct iphdr_s *head)
 void ip_recvpacket(char *packet,int size)
 {
     struct iphdr_s *iphdr;
-    __u8 *addr, *data;
+    __u8 *data;
+    //__u8 *addr;
 
     iphdr = (struct iphdr_s *)packet;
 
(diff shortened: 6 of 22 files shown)


ghaerr:
Christoph, thanks for this cleanup, it is definitely needed!

`ktcp` definitely does work, and is an important component in ELKS. Both ethernet and slip work, but I have yet to test either. I have also not yet figured out how to configure QEMU for ethernet card, which is required for testing this. We are also actively engaged improving the serial port driver, and @pawosm-arm tests `ktcp`, albeit using 300 baud. I have completed changes that should enable much higher baud rates, but @pawosm-arm's system has particular serial hardware issues still unresolved.

Thus, I am hesitant, as are you, to commit this without testing. That said, I really want to, as it removes a large number of compiler warnings.

With regards to the biggest change, the 'static int tcpdevfd' mess. In looking very closely at the code, there is only one `open` in tcpdev.c, and then a copy is made by the calling module ktcp.c into the same-named static variable. Terrible coding, but I can see how it worked. However, arp.c then uses its own static copy in select, which has to be buggy.

Since there is only one open, and undoubtably these all need to be the same variable, I suggest the following:
- put 'extern int tcpdevfd' in a header file.
- declare 'int tcpdevfd' in ktcp.c
- use a local variable in tcpdev_init() and return the file descriptor
- remove all static declarations of tcpdevfd

This will ensure there is only one copy, and not have that copy written to twice. 

We still should test this before committing. I desperately need help in getting QEMU setup for network emulation. That will also help me or others finish the configurable IRQ stuff for the NE2000 card, which needs its base address to be programmable.

Mellvik:
I can set up a physical config to test this. 
It may even be testable with VB.

@Christoph, make the changes and send me a diff.

-M

cjsthompson:
Changes done. Still compiles and I fixed some more easy warnings I missed at the same time.

> I desperately need help in getting QEMU setup for network emulation. 

I tried to get qemu to work a few days ago for something else, and I gave up. Couldn't get an ne2000 isa network card to work either. The command line options handling is truly a mess.

> make the changes and send me a diff.

Do you need a diff of all the changes to ktcp or only this last commit?

ghaerr:
> I can set up a physical config to test this.
> It may even be testable with VB.

@Mellvik, that would be great! I am interested in both serial and ethernet testing.
If you can figure out how to get VB working, I'll set that up as well, although I normally use QEMU.

I have had some trouble attaching hard drives to VB, but networking testing can be done from a 1440k image.

cjsthompson:
I have an ISA ne2000 RJ 45 network card that should work in an 8 bit slot. Whenever I get the monitor repaired I'll be able to test too. Browsing gopher on my Amstrad, now that would be awesome.

ghaerr:
> else if ((!h->flags) & TF_ACK)

In the last commit, I think this is an error. I realize that it first looked like 
'else if (!h->flags & TF_ACK)', but neither make any sense to me; if true, the code prints NO ACK received in the next statement - it is meant to test for ACK received (e.g. in h->flags).

My analysis is that h->flags is always non-zero entering this routine. Thus !h->flags is zero. Anding 0 with TF_ACK is 0, thus the printf is skipped and the state established, which is why this bad code has always worked.

Suggest changing to 'else if ((h->flags & TF_ACK) == 0)'.

cjsthompson:
It's just me that read too fast and didn't see this was a bitwise operation and not a &&.

ghaerr:
> It's just me that read too fast and didn't see this was a bitwise operation and not a &&.

No problem - its better when changing untested code to leave the semantics intact like you did, until we actually understand it. Thanks for helping clean this up.

You can see the need to test ktcp. We have a number of open issues surrounding it, and no easy way for the ELKS users to test it. We desperately need a real serial application like slip to test serial, which should be easily tested in an emulator. I've got QEMU redirecting serial I/o but only to a terminal.

I don't know enough about the older network cards, I think QEMU emulates an NE1000, but still unsure of its complicated command line option to initialize that to talk with the host network.

cjsthompson:
> We have a number of open issues surrounding it, and no easy way for the ELKS users to test it.

PCem can emulate an NE2000 card for an XT. But I haven't given it a try yet with something like mTCP which I know works for sure.

> I don't know enough about the older network cards, I think QEMU emulates an NE1000, but still unsure of its complicated command line option to initialize that to talk with the host network.

I think NE1000 cards are only available with BNC connectors which makes them not really a great choice with real hardware these days.

Mellvik:
Christoph,
> 
> I have an ISA ne2000 RJ 45 network card that should work in an 8 bit slot. Whenever I get the monitor repaired I'll be able to test too.
> 
Actually - you can configure ELKS to run console on serial and you're in operation. With the added benefit og real life testing serial console.

—M

Mellvik:
> I can set up a physical config to test this.
> It may even be testable with VB.
> 
> @Mellvik <https://github.com/Mellvik>, that would be great! I am interested in both serial and ethernet testing.
> If you can figure out how to get VB working, I'll set that up as well, although I normally use QEMU.
> 
> I have had some trouble attaching hard drives to VB, but networking testing can be done from a 1440k image.
> 
@ghaerr,
I can send you an 'exported appliance' from VB, which can be imported and you're up and running.
However, VB does not support NE2000 or anything else pre PCI. So while VB is good for serial testing and most other stuff (really easy to configure), it's unusable for ELKS network testing.

Eventually, I'll get you a serial login to ELKS on one of my Compaqs via the Internet. :-)

—Mellvik

https://www.dropbox.com/s/5tmbhbta9ul9mzc/Elks%200.3.ova?dl=0 <https://www.dropbox.com/s/5tmbhbta9ul9mzc/Elks%200.3.ova?dl=0>
The image has two disk drives, the first a small DOS disk (can be mounted), the second is a 500MB drive - attempting to mount it gives a 'No such device or address' (and boot says bdb:(0,1023120) bdb1:(…) skipped ), so I'm guessing there is nos pace allocated for a 2nd drive in the config.

cjsthompson:
> Christoph,
> Actually - you can configure ELKS to run console on serial and you're in operation. With the added benefit og real life testing serial console. —M

I can't do that as the Amstrad will not work with any other monitor than the one designed for it, as the monitor includes the power supply that powers both the monitor and the motherboard. Also the connectors for video and power coming out of the monitor are one of a kind.

ghaerr:
Christoph, lets keep this on temp hold, due to problems with @Mellvik's testing which this PR seemingly somehow breaks. I will be submitting a PR to change just the static int tcpdevfd issues, which I have working on my system, without this PR. After we get through that, and the additional regression problems just reported in #612, we'll get back to getting this PR committed.

cjsthompson:
Sure. Just trying to fix the warnings I can see ktcp seems to need a bit of work and I was surprised it did work because of the "work in progress" state of the code. In the meantime I'm fixing more warnings and cleaning up a bit elkscmd inet utilities in another local branch that I'll be able to test when ktcp problems are sorted out.

ghaerr:
Hello @cjsthompson,

Given I had to pull out my machete to try to hack `ktcp` into submission, which has likely changed a lot code making even more work to get this PR merged, I'm thinking of taking this work and merging it with my next ktcp PR, since I can check that ktcp continues to operate, which was the primary reason this PR wasn't commit long ago. Does that sound alright with you? I see that as the quickest path to getting your work merged and ktcp cleanup continued moving forward.

cjsthompson:
Hi,

Yes, this is completely fine with me.

ghaerr:
Christoph - I have added your changes from this PR to #626, thanks! This PR can be closed.

Issue #608, opened 2020-05-03
Misc bugs - ps(), sash(), ash(), tail()


Mellvik:
Elvis (and probably there interactive programs)  cannot spawn subshells when stared from sash (reported before, will be fixed). Just as interesting, after attempting such a spawn (e.g. `:!ls`), graces of the attempt lingers in ps. More attempts adds more entries to ps with the same data. Eventually ps() prints lots of garbage - seemingly an array out of bounds type problem. See screenshots. The situation does not affect system stability.

Not related, but ash also fails when programs are forking subshells. While sash complains about '-c not found', which is easy to explain, ash simply reports 'fork failed'.

Finally, tail has a memory (sbrk) problem, try `hd /bin/vi | tail -100`. 
All testing i qemu, build from d5b0d45f.

--Mellvik

<img width="725" alt="Skjermbilde 2020-05-03 kl  18 02 36" src="https://user-images.githubusercontent.com/3629880/80919515-97bcbd00-8d6a-11ea-9e85-2851c7ae0520.png">

Mellvik:
Update:
- when staring with ash as login shell, the behaviour is slightly different:
- forking from elvis via ash works (output somewhat garbled since the console is still in raw mode, it's on my list of elvis problems).
- but after such fork (repeated forks work, even :!ps), and exit from elvis, ps() prints garbage.

ghaerr:
hello @Mellvik, there are a number of separate issues in the items you bring up. 

> after attempting such a spawn (e.g. :!ls), graces of the attempt lingers in ps. 

Note the processes are in Zombie state. Their task structure entry will only be cleaned up after a 'wait' command. This usually means after the shell as executed the next command. sh and sash have entirely different methods for waiting, which differs in sash depending on whether it was an internal command, had to run a /bin program, or had to run /bin/sh -c for the command line.
If the parent is dead (subshell, etc), then `init` performs the wait. It's a bit complicated.

More attempts adds more entries to ps with the same data.

These are all Zombie states, waiting to be cleaned up.

> Eventually ps() prints lots of garbage - seemingly an array out of bounds type problem. The situation does not affect system stability.

There are only 15 process slots available, was this what you saw? I don't see the garbage output, except for the SIZE obviously incorrect. This is likely due to the task resources having been released or reallocated. `ps` tries to figure all this out by reading /dev/kmem, so there are limits in what it does.

[...]

PR #609, opened 2020-05-04
[cmds] Fix ps, sash, tail, elvis, wait

ghaerr:
Fixes all items mentioned in #608.

Fixes `ps`, problem was DSEG was deallocated. Thus, there is no reliable way to print command arguments, as segment is discarded. Output truncated after CSEG/DSEG displayed as 0.

Fixes `elvis` to properly display ":!cmd" output. Turns out this was related to the `wait` system call returning -EINTR, rather than a process id. Thus, control returned to elvis prior to the command output being displayed, and by that time elvis had reset raw terminal mode.

Fixes libc `wait` to retry on EINTR, to be compatible with Linux. This fixed the elvis problem, and has the side effect of the kernel `wait` waking up `init` to process zombie entries, so `ps` is unlikely to show zombie processes, even though it is also fixed.

Adds -c option to `sash`. Traced raw mode behavior in `ash` and determined not to be the problem. Added `sash -c` to fully test above wait() fix.

Fixes `tail` by allocating 1280 bytes rather than 20k bytes. Seems to work well but needs further testing.

### DIFF
--- a/elkscmd/minix3/tail.c
+++ b/elkscmd/minix3/tail.c
@@ -66,7 +66,7 @@ extern int optind;
  * but we'll specify it here just in case it's been left out.
  */
 #ifndef LINE_MAX
-#define LINE_MAX 2048		/* minimum acceptable lower bound */
+#define LINE_MAX 128		/* minimum acceptable lower bound */
 #endif
 
 /* Magic numbers suggested or required by Posix specification */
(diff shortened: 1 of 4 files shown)


Mellvik:
Thanks @ghaerr,
again - what can I say? The time difference seems to be working greatly to my advantage: Report problems, go to sleep, get up bright and early - to a brand new (!) ELKS. :-) 

Testing now.

—Mellvik

Mellvik:
@ghaerr,

Testing sash, ps, elvis, tail and more - no more comments.

Great work.

—Mellvik

PR #611, opened 2020-05-05
[kernel] Add boot command line options, switchable system console

ghaerr:
This enhancement allows passing boot options to ELKS in the form of command lines in the file /bootopts on the booted root filesystem. Currently, only one option is implemented, 'serial' which will cause ELKS to boot with /dev/ttyS0 as the serial console (with no recompilation). All startup kernel printk's will go to /dev/ttyS0. Later, arbitrary options can be made available, and parsed in the system setup code, setup.S.

With ELKS getting more complex, it is useful to redirect all console printk output to a terminal emulator with a larger screen size or scroll back, for system analysis. Since /dev/console is now also implemented, both `init` and `getty` can be compiled in debug mode and their /dev/console output redirected to serial ttyS0 or tty2, for instance, without recompiling them for the console switch.

The serial console and logins can be easily tested using QEMU by uncommenting one of the CONSOLE= lines in qemu.sh.

In addition, this PR allows runtime switching system consoles to any TTY device, using the new `console` command, e.g. `console /dev/ttyS0` or `console /dev/tty2`. No recompilation needed.

Due to very little boot code space available, boot cmd lines are currently only available for MINIX boots.

Other enhancements, developed during this implementation:
- Implement /dev/console. Writing to this will write to the system console, wherever it is. All kernel printk's go to /dev/console.
- Enhanced `hd` to inspect global PC memory directly. Usage `hd seg:off#size`, e.g. `hd 0:0` to look at low core, or `hd 60:0#512` to look at segment 60:0000 for 512 bytes. Using a ':' in the otherwise filename is interpreted as global memory.
- For direct console, console switching (display of tty1/tty2/tty3 and logins, etc) is done by using F1, F2, F3 respectively, rather than Alt-F1... This was made because there are no working Alt keys on my laptop keyboard. All ttys can be enabled for login by running `init 6`. More fixes coming on multiple consoles.
- Bug fixes on tty queue allocations and added tty open/close use counts to handle them.


Mellvik:
@ghaerr,
again a milestone afaik - I can't wait to get this merged in so we can start using it.

One comment: I'm not happy about the def of ttyS2 going away in inode.c since I'm using it off an on for testing. Also - it doesn't seem to make much sense as long ar there are resource allocations for 4 serial lines in serial.c. Which is possibly overkill and something to be configurable at some point.

Further - I'd like to suggest (like we talked about before) we change the minor numbering for serial lines now that you're at it, to make it easy to add with/without HW flow control distinction later. How about keeping the no-flow control devices at even minor, with hw flow control at odd numbers? /dev/ttyS0 = minor 64, /dev/ttyS0m = minor 65 etc. (m as in modem, may be anything)?

—Mellvik

ghaerr:
> I'm not happy about the def of ttyS2 going away in inode.c since I'm using it off an on for testing. 

I mentioned that we basically out of devfs space in FAT. Here's the list, which should be removed instead of ttyS2? tty2 comes to mind, but then FAT can't use mutiple consoles... ? /dev/zero is an option, as is /dev/mem. /dev/kmem is required for ps, but /dev/mem isn't used.
```
    { "bda",    S_IFBLK | 0644, MKDEV(3, 0) },
    { "bda1",   S_IFBLK | 0644, MKDEV(3, 1) },
    { "fd0",    S_IFBLK | 0644, MKDEV(3, 128)},
    { "fd1",    S_IFBLK | 0644, MKDEV(3, 160)},
    { "mem",    S_IFCHR | 0644, MKDEV(1, 1) },
    { "kmem",   S_IFCHR | 0644, MKDEV(1, 2) },
    { "null",   S_IFCHR | 0644, MKDEV(1, 3) },
    { "zero",   S_IFCHR | 0644, MKDEV(1, 5) },
    { "tcpdev", S_IFCHR | 0644, MKDEV(8, 0) },
    { "eth",    S_IFCHR | 0644, MKDEV(9, 0) },
    { "tty1",   S_IFCHR | 0644, MKDEV(4, 0) },
    { "tty2",   S_IFCHR | 0644, MKDEV(4, 1) },
    { "ttyS0",  S_IFCHR | 0644, MKDEV(4, 64)},
    { "ttyS1",  S_IFCHR | 0644, MKDEV(4, 65)},
    { "ttyS2",  S_IFCHR | 0644, MKDEV(4, 66)},
    { "tty",    S_IFCHR | 0666, MKDEV(4, 255) },
```

> Also - it doesn't seem to make much sense as long ar there are resource allocations for 4 serial lines in serial.c. Which is possibly overkill and something to be configurable at some point.

With the recent updates, this is a nonissue. There are no longer any significant resources used in the tty or serial_info structures, so its not worth a configuration option.

ghaerr:
> we change the minor numbering for serial lines now that you're at it, to make it easy to add with/without HW flow control distinction later. How about keeping the no-flow control devices at even minor, with hw flow control at odd numbers? /dev/ttyS0 = minor 64, /dev/ttyS0m = minor 65 etc. (m as in modem, may be anything)?

Odd/even is not a good idea, the minor numbers are used to index the serial_info array. 

The naming scheme isn't bad, but non-standard.

I'm not worried about creating these when you need them - but given the already present problems in FAT devfs, I suggest we wait until you actually develop a working hw flow control driver, and use CLOCAL vs CRTSCTS in termios to turn it on and off while testing. This will be far easier.

I plan to submit a PR soon with CLOCAL and CRTCTS reviewed and ready to go in all the potential applications that might use them, to ease your testing while developing.

ghaerr:
>  I'm not happy about the def of ttyS2 going away in inode.c since I'm using it off an on for testing. 

Ok, I've swapped out /dev/mem, which isn't used anywhere and put back /dev/ttyS2.

Issue #16, opened 2016-01-20
full3 image crashes on boot


deephooloovoo:
I was trying to compile elks and get networking running with the full3 image, but I was not able to crash because there was an error about a double free which crashed completely. I narrowed down the issue to elks/arch/i86/mm/malloc.c in commit 4786296011e318ca250ad096ce3ccd4ae01b21fe

I am not familiar enough with the internals of malloc to fix the problem myself, as some of the optimizations look like they could be helpful when specific conditions are met.

jbruchon:
Is this happening on real hardware or in an emulator?

deephooloovoo:
It's happening on hardware, both a really old 286 and an also pretty old P3 laptop. I haven't been testing on an emulator.

lithoxs:
It is a bug in malloc.c inadvertently introduced by me 2 commits ago. I already set a pull request that fixes this problem.

jbruchon:
Fixed in commit 7ae683f0aefcbe8cf4fd75d69f6e7a8988e2d128

PR #613, opened 2020-05-07
[kernel] Decrease and tune main memory usage

ghaerr:
Various changes made to decrease application main memory requirements and fine-tune application heap and stack defaults. Saves 48k of main memory on ELKS boot w/networking, from 212k free -> 260k unused main memory.

Decrease default application heap and stack from 16k to 8k. As can be seen from the `ps` listings,  almost all applications have large amounts of free memory. This change reduces 8k of memory footprint per application, except for `ktcp`, which requires 12k. With more fine tuning, and an upcoming ability to separately specify heap and stack sizes, most ELKS apps will be able to run with an even smaller footprint. Should applications run out of heap memory, the kernel will print the amount requested to help tune the application, and a -maout-chmem= option can be added at link time to specify memory required.

Decrease default L2 buffers from 128 to 64 (original setting).

When `sash` shell is run via the -c option (e.g. `vi` and ":!cmd"), don't fork, just exec. (Thanks to @Mellvik for this suggestion).

Add fine-grained display options for `meminfo`: '-a, -f, -t, -b' respectively to show application, free, tty and buffer memory only.

Application memory usage before:
16k default
And after:
8k default


Mellvik:
> 
> Various changes made to decrease application main memory requirements and fine-tune application heap and stack defaults. Saves 48k of main memory on ELKS boot w/networking, from 212k free -> 260k unused main memory.
> 
> 
FWIW: This is a major improvement - not only for usability, bit for testing! We can now have ktcp loaded and still do work. With vi…

-M

Mellvik:
One more thing - now about ps():
The screenshots above (in the thread) reminded me about something I noticed the other day: The tty column is no longer (necessarily) correct. AFAIK it should say '1' for the shell process (#16) and ps (#18), while ktcp (#6) should have a blank field.
The problem seem to be limited to tty1, ttyS0, tty2 etc. seem fine.

Am I missing something?

--Mellvik

ghaerr:
> The tty column is no longer (necessarily) correct. AFAIK it should say '1' for the shell process (#16) and ps (#18), while ktcp (#6) should have a blank field.

Good catch, the shell should be '1', as well as `ps`. This likely has to do with the implementation of /dev/console during serial console development. I'll look in to fixing it.

`ktcp` is correct (for now). For debugging reasons, it is keeping its original /dev/tty file descriptors open. This will be changed once we get networking fully operational.

ghaerr:
> FWIW: This is a major improvement - not only for usability, bit for testing! We can now have ktcp loaded and still do work. With vi…

I am working on system handling of low memory situations. There are still some problems with the system running well in low memory situations. Also, some commands still use too much stack and won't run well as we tune the system to use much less memory, so keep an eye out and try running a bunch of less-used commands. The kernel will display stack overflow and malloc/break overflow will display by how much. You can test arbitrary command memory usage by running using 'command | ps', which will run the command while ps is running.

The next step, now that we can see process heap and stack usage, is to allow separate heap and stack requirements in the program a.out header, and try to move to a 2k default heap and 2k default stack. More on that coming. Its nice to have the system running well enough and with enough information to allow us to start fine tuning!

PR #615, opened 2020-05-08
[kernel] De-duplicate console code, update BIOS console

ghaerr:
Brings BIOS console completely up-to-date with all direct console enhancements by de-duplicating all shared code into a new file console.c. 

BIOS console now operates well with `vi`, `kilo`, `sh` w/linenoise, etc.

Note: Direct and BIOS keyboards operate quite differently, with direct console supporting international keymaps, and is quite a bit faster. Not all function/arrow/special keys are supported with BIOS console. Use of direct console is recommended.

There is no code in ELKS keyboard drivers that operate the keyboard LEDs.

### DIFF
--- a/elks/arch/i86/drivers/char/bioscon.c
+++ b/elks/arch/i86/drivers/char/bioscon.c
@@ -36,16 +36,6 @@
 /* Assumes ASCII values. */
 #define isalpha(c) (((unsigned char)(((c) | 0x20) - 'a')) < 26)
 
-#if 0
-/* public interface of bioscon.c: */
-
-int Console_write(struct tty *tty);
-void Console_release(struct tty *tty);
-int Console_open(struct tty *tty);
-struct tty_ops dircon_ops;
-void init_console(void);
-#endif
-
 /* Function from bioscon-low */
 /* TODO: move these functions to a BIOS library */
 
@@ -79,8 +69,9 @@ typedef struct console Console;
 struct console {
     int cx, cy;			/* cursor position */
     void (*fsm)(register Console *, char);
-    int pageno;			/* video ram page # */
     unsigned char attr;		/* current attribute */
+    unsigned char XN;		/* delayed newline on column 80 */
+    unsigned char color;	/* fg/bg attr */
 #ifdef CONFIG_EMUL_VT52
     unsigned char tmp;		/* ESC Y ch save */
 #endif
@@ -89,6 +80,7 @@ struct console {
     unsigned char *parmptr;	/* ptr to params */
     unsigned char params[MAXPARMS];	/* ANSI params */
 #endif
+    int pageno;			/* video ram page # */
 };
 
 static struct wait_queue glock_wait;
@@ -97,8 +89,6 @@ static Console *glock;		/* Which console owns the graphics hardware */
 static int Width, MaxCol, Height, MaxRow;
 static unsigned short int NumConsoles = MAX_CONSOLES;
 static struct timer_list timer;
-
-extern struct tty ttys[];
 
 /* for keyboard.c */
 int Current_VCminor = 0;
@@ -112,16 +102,9 @@ int kraw = 0;
 #define TERM_TYPE " dumb "
 #endif
 
+extern void AddQueue(unsigned char Key);
 static void std_char(register Console *, char);
 static void Console_set_vc(unsigned int);
-
-static void AddQueue(unsigned char Key)
-{
-    register struct tty *ttyp = &ttys[Current_VCminor];
-
-    if (!tty_intcheck(ttyp, Key))
-	chq_addch(&ttyp->inq, Key);
-}
 
 static void kbd_timer(int __data);
 /*
@@ -161,21 +144,29 @@ static void kbd_timer(int __data)
 	    AddQueue(dav & 0x7F);
 	else {
 	    dav = (dav >> 8) & 0xFF;
-	    if ((dav >= 0x3B) && (dav < 0x45))	/* Function keys */
+	    if (dav >= 0x3B && dav <= 0x3D) {	/* temp console switch on F1-F3*/
+		Console_set_vc(dav - 0x3B);
+		dav = 0;
+	    }
+	    else if (dav >= 0x3B && dav < 0x45)	/* Function keys */
 		dav = dav - 0x3B + 'a';
-	    else if ((dav >= 0x57) && (dav < 0x59))
+	    else if (dav >= 0x57 && dav < 0x59)
 		dav = dav - 0x57 + 'k';
-	    else if ((dav>= 0x48) && (dav <= 0x50) && ((1 << (dav - 0x48) & 0x129))) {
-		dav = (((0x4A - dav) >> 1) & 0x07) + ('A' - 3); /* Arrows */
-		if (dav < 'A')
-		    dav = 'A';
+	    else if ((dav >= 0x48 && dav <= 0x50) && ((1 << (dav - 0x48) & 0x129))) {
+		/* arrow keys*/
+		switch(dav) {
+		case 0x48: dav = 'A'; break;
+		case 0x50: dav = 'B'; break;
+		case 0x4d: dav = 'C'; break;
+		case 0x4b: dav = 'D'; break;
+		default:   dav = 0;
+		}
 	    }
 	    else if ((dav >= 0x68) && (dav < 0x6B)) {	/* Change VC */
 		Console_set_vc((unsigned)(dav - 0x68));
 		dav = 0;
 	    }
-	    else
-		dav = 0;
+	    else dav = 0;
 	    if (dav) {
 		AddQueue(ESC);
 #ifdef CONFIG_EMUL_ANSI
@@ -188,7 +179,7 @@ static void kbd_timer(int __data)
     restart_timer();
 }
 
-static void PositionCursorSet(register Console * C)
+static void PositionCursor(register Console * C)
 {
     int x, y, p;
 
@@ -216,7 +207,7 @@ static void VideoWrite(register Console * C, char c)
     VideoWriteLow (c, a, p);
 }
 
-static void genscro(register Console * C, int n, int x, int y, int xx, int yy)
+static void scroll(register Console * C, int n, int x, int y, int xx, int yy)
 {
     int a;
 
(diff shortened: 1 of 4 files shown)


Mellvik:
When the system boots, somehow a 'b' makes it into the console input buffer and shows up after the login: prompt (not on serial console). Can be deleted.

ghaerr:
> When the system boots, somehow a 'b' makes it into the console input buffer and shows up after the login: prompt (not on serial console). Can be deleted.

Try hitting enter without deleting it, to see whether the 'b' is actually there, or is instead the last output of a program that wrote that to the console but got rescheduled and didn't finish writing until login waited. A screenshot would be nice. I'm pretty sure this problem is that of process scheduling and the character is not in the tty buffer.

Does it happen every time? If so, send a screenshot, thanks!

Mellvik:
> When the system boots, somehow a 'b' makes it into the console input buffer and shows up after the login: prompt (not on serial console). Can be deleted.
> 
> Try hitting enter without deleting it, to see whether the 'b' is actually there, or is instead the last output of a program that wrote that to the console but got rescheduled and didn't finish writing until login waited. A screenshot would be nice. I'm pretty sure this problem is that of process scheduling and the character is not in the tty buffer.
> 
> Does it happen every time? If so, send a screenshot, thanks!
> 
> 
Yes, it happens every time, but not if the console is redirected to serial. And it's a 'd', not  a 'b' like I said. And yes, it's real, getty reads it and login asks for a pw.

I'll attach a photo via the web interface.

-Mellvik

ghaerr:
> Yes, it happens every time, but not if the console is redirected to serial. And it's a 'd', not  a 'b' like I said.

Did this start happening only after this particular commit, or sometime before? I hadn't committed this #615 until about an hour ago. I am not getting this to duplicate on QEMU. Perhaps you can check if this happens on VB or only on real hardware, thanks!

Mellvik:
I've been using your changes from the diff for testing.

And no, while not depending entirely on my memory, I believe this came with these changes. 
Additional info: When using direct console, it's a 'db', not a 'b' at the login prompt - and again it's repeatable.

As to keyboards: I don't think emulator keyboards necessarily represent dependable tests. There's a long way from the actual keyboard to the scan codes and onward. And no, I have never tested this on any emu, I don't have a keyboard with a keypad.

Ignoring the LEDs for now, I've compared the chars generated by the keypad in both direct and BIOS mode, and they are very different. Unsurprising, the numlock key in direct is nonfunctional, so numlock is always off. Capslock works as expected, w/o the LED indicator. 

IMG_6029
IMG_6030

Mellvik:
Looking at the code in xt_key.c it all makes sense. This is what the name says, art-keyboard support. There is no keypad, no leds etc. - like you pointed out before, @ghaerr.

So - since I have the setup to test it, I'm moving over AT keyboard support from the Minix 2 driver. Seems like a reasonable challenge.

--Mellvik

ghaerr:
Hello @Mellvik,

> And no, while not depending entirely on my memory, I believe this came with these changes.
Additional info: When using direct console, it's a 'db', not a 'b' at the login prompt - and again it's repeatable.

I would like to ask that you try repeating this strange issue after editing /etc/rc.d/rc.sys and commenting out `ktcp` startup. I have recently found that ktcp has definite memory corruption problems. Thanks for the screenshots. I would like a screenshot minus ktcp that shows everything from "Elks Setup" to login, to try to debug this.


> As to keyboards: I don't think emulator keyboards necessarily represent dependable tests. There's a long way from the actual keyboard to the scan codes and onward. And no, I have never tested this on any emu, I don't have a keyboard with a keypad.

True. This also means there are a wide variety of possibly scan code variations, all of which have to be handled with keymaps, which is currently done with direct console. International usage requires it, and only direct console supports this.

> Ignoring the LEDs for now, I've compared the chars generated by the keypad in both direct and BIOS mode, and they are very different. Unsurprising, the numlock key in direct is nonfunctional, so numlock is always off. Capslock works as expected, w/o the LED indicator.

The solution is simple, the details complicated: My recent changes never allow a scan-code style return value (i.e. > 127) to be returned to an ELKS application. Instead, ALL special keys return ESC sequences. This is all handled in the console driver(s).

> Looking at the code in xt_key.c it all makes sense. This is what the name says, art-keyboard support. There is no keypad, no leds etc.

We need to separately discuss LEDs, Caps Lock, Num Lock, and Arrow key problems.

Due to tons of other far more pressing issues, I am only interested in solving the arrow key problem immediately. I don't have data from you on what exactly happens when the arrow keys are used in direct console. Specially, what codes are sent different on your system that prevent their being turned into ESC [ A sequences in xt_key.c.


> So - since I have the setup to test it, I'm moving over AT keyboard support from the Minix 2 driver. Seems like a reasonable challenge.

This opens up another big can of worms - I would like to delay its implementation until other problems are solved. Specifically, we cannot accept a PR of a new keyboard driver unless it uses all international keymaps currently used by dircon.c and keymap/\*.h, and is tested in each of those languages/regions.

I don't recommend writing a new keyboard driver at this time. We already have two, and we're running out of space in the kernel. Lets just get arrow keys working for now on direct console - which work on other hardware, its your system that they're not working on.

Mellvik:
> Additional info: When using direct console, it's a 'db', not a 'b' at the login prompt - and again it's repeatable.
> 
> I would like to ask that you try repeating this strange issue after editing /etc/rc.d/rc.sys and coming out ktcp startup. I have recently found that ktcp has definite memory corruption problems. Thanks for the screenshots. I would like a screenshot minus ktcp that shows everything from "Elks Setup" to login, to try to debug this.
> 
Your hunch is correct, @ghaerr. No ktcp, no extra chars. As an extra indication that it's a memory corruption issue, the chars are gone after a 'no ktcp/reboot/with ktcp/reboot'. A powercycle is required to get it back. 
Do you still need screenshots?
> Looking at the code in xt_key.c it all makes sense. This is what the name says, art-keyboard support. There is no keypad, no leds etc.
> 
Again my helpful spelling checker screwed me - but you got the drift. 
> We need to separately discuss LEDs, Caps Lock, Num Lock, and Arrow key problems.
> 
It' hard to separate them completely since we're talking about the same keys (same keypad actually). And I didn't suggest writing a new driver, but updating xt_key.c (possibly changing the name to boot). 

I agree with you with regard to the requirement for international keyboard support. Actually  I'd like to add norwegian while at it. I have a NO kbd to do the testing with.
Since the basics are already in place, this does not seem like a great ordeal. I may have something functional by tomorrow (w/o more locales in the first take).

--Mellvik

ghaerr:
> Your hunch is correct, @ghaerr. No ktcp, no extra chars. 

More networking bugs. Geez. I'm not surprised its not a direct console bug actually, as I've been using that for months.

> Do you still need screenshots?

Nope. This entire open issue is now about enhancement requests, as there are no actual ELKS bugs (other than ktcp, which is filled with them).

> Actually  I'd like to add norwegian while at it. I have a NO kbd to do the testing with.

Add Norwegian as a config option and separate keymaps-no.h file, just like the others.

> Since the basics are already in place, this does not seem like a great ordeal. I may have something functional by tomorrow (w/o more locales in the first take).

That's fine, but I am becoming very concerned about using up lots of kernel code for small features. If you submit something, please consider using as little code as possible, especially considering that BIOS console solves your problem and is now working. We don't need direct console to do everything at this point, and when we run out of kernel space (less than 9k left) we won't be able to add anything at all! So having configurable options like BIOS vs direct console, may be a good idea after all, to keep rarely-needed features and kernel code to a minimum.

Also - we have added lots of code for serial, with even more enhancements planned... and we haven't even tested that code yet, while we're off on this and the networking tangents. I would prefer to get one enhancement done well before enhancing another, unless its a kernel bug. I'm feeling we're starting to go a bit around a big circle...

Mellvik:
> 
> Since the basics are already in place, this does not seem like a great ordeal. I may have something functional by tomorrow (w/o more locales in the first take).
> 
> That's fine, but I am becoming very concerned about using up lots of kernel code for small features. If you submit something, please consider using as little code as possible, especially considering that BIOS console solves your problem and is now working. We don't need direct console to do everything at this point, and when we run out of kernel space (less than 9k left) we won't be able to add anything at all! So having configurable options like BIOS vs direct console, may be a good idea after all, to keep rarely-needed features and kernel code to a minimum.
> 
I get that - and share that goal. And yes, BIOS console is functionally operational making direct console an obvious lower priority. Having spent some time in xt_key.c and (and added most of the code required), I'll make it work and stash it for now. As to size, I don't think adding AT kbd support will add much. xt_key is ugly - with ample room for improvements, also in terms of size.
> Also - we have added lots of code for serial, with even more enhancements planned... and we haven't even tested that code yet, while we're off on this and the networking tangents. I would prefer to get one enhancement done well before enhancing another, unless its a kernel bug. I'm feeling we're starting to go a bit around a big circle...
> 
Not a bad description - but allow me to call it a spiral: Going round but always going forward :-).  


--M

ghaerr:
> Having spent some time in xt_key.c and (and added most of the code required), I'll make it work and stash it for now. 

I would like to know whether you have fixed the arrow keys for direct console, and what the problem was.

If you've got an enhancement (or above bug fix) coded, send it over, lets have a look at it.

Mellvik:
> 
> Having spent some time in xt_key.c and (and added most of the code required), I'll make it work and stash it for now.
> 
> I would like to know whether you have fixed the arrow keys for direct console, and what the problem was.
> 
> If you've got an enhancement (or above bug fix) coded, send it over, lets have a look at it.
> 
> 
> 
@ghaerr,
yes, I added a quick&dirty fix for the arrow keys - adding the explicitly (~276):

#ifdef CONFIG_EMUL_ANSI
        code = mode = 0;
        switch ((unsigned int)keyp_E0) {
        case 0xb7: code = 'H'; break;                   /* home*/
        case 0xb1: code = 'F'; break;                   /* end*/
        case 0xb9: code = '5'; mode = '~'; break;       /* page up*/
        case 0xb3: code = '6'; mode = '~'; break;       /* page dn*/
        case 0xb8: code = 'A'; break;                   /* up*/
        case 0xb2: code = 'B'; break;                   /* down */
        case 0xb6: code = 'C'; break;                   /* right */
        case 0xb4: code = 'D'; break;                   /* left */
        }

The code has some logic related to NUMLOCK, but never checks for numlock.

My quick fix version adds this and LED support. 

And like we discussed before, xt_key.c needs an overhaul - which is partly done.

—Mellvik

ghaerr:
Hi @Mellvik, thank you for getting back to me on the problem on your system, I now understand it.

> yes, I added a quick&dirty fix for the arrow keys - adding the explicitly (~276):

It seems that my QEMU system is returning E0 extended scan codes for the arrow keys (see line 230), while yours does not.

> The code has some logic related to NUMLOCK, but never checks for numlock.
My quick fix version adds this and LED support.

Lets add your fix for the non-E0 function keys as well as fixed numlock and LED support now.
Please submit a PR for your "quick-fix" version.

> And like we discussed before, xt_key.c needs an overhaul - which is partly done.

We need to be very careful with a full overhaul - I'm afraid that it will be too hard to test the vast variety of systems keyboards etc, and likely shouldn't be messed with.

Mellvik:
> 
> It seems that my QEMU system is returning E0 extended scan codes for the arrow keys (see line 230), while yours does not.
> 
> 
That's the difference between the XT and AT keyboards (our rather, one of them).
> The code has some logic related to NUMLOCK, but never checks for numlock.
> My quick fix version adds this and LED support.
> 
> Lets add your fix for the non-E0 function keys as well as fixed numlock and LED support now.
> Please submit a PR for your "quick-fix" version.
> 
Will pull out the experimental stuff and test. PR later this week.
> And like we discussed before, xt_key.c needs an overhaul - which is partly done.
> 
> We need to be very careful with a full overhaul - I'm afraid that it will be too hard to test the vast variety of systems keyboards etc, and likely shouldn't be messed with.
> 
Agreed, on the other hand, few of the improvements in ELKS over the past months mould have been there if we were afraid of breaking things, right? Anyway, this is not top priority at the moment.

—M

ghaerr:
> Will pull out the experimental stuff and test. PR later this week.

Great! Glad to get this understood and solved.

> on the other hand, few of the improvements in ELKS over the past months mould have been there if we were afraid of breaking things, right?

There is a huge difference - testing. A key component of all of us moving things forward is the ability to test. As we've seen, there have been far too many 20-year-old contributions that were never tested... amazing that we're finding these almost daily now!

WRT keyboards - as you said, it is usually a quite complicated path getting keystrokes from hardware to application - and because ELKS pretty much has to support ancient systems - I don't want to break support for them. And we don't need more multiple drivers either - too much maintenance.

That said - at some point we must make decisions on what ancient code or stuff to throw out in the name of maintainability. I'm just not sure something as complicated as keyboard scan codes is the place to do that. Case in point is the E0-prefix versus single-byte scan code difference in BIOS for this very issue.

PR #620, opened 2020-05-09
Add cron job scheduler to ELKS

georgp24:
Ported a version of cron and crontab to ELKS and added documentation for it.


ghaerr:
Hello Georg, you're porting more applications for ELKS complete with documentation, thank you.

I have a couple questions and comments, and then a change or two that I'd like to see before getting this committed.

Have you tested this much? I see the demo example after installation writes to a cron.out file, what directory is that file created in?

With regards to memory requirements, I'm working on trying to get applications, especially daemons, made very much smaller so they can co-exist with all the other processes that are likely to be running at boot time. The -maout-chmem= option is likely required, but I see you've set it to 32k for both cron and crontab. That's going to be a lot of extra memory for just cron, and will likely result in users not using cron as much as they might if the exact requirements were better known. Can you look at cron and crontab's memory usage (both heap and stack) using `ps` to determine how much heap and stack is actually needed? Very soon we will be setting separate heap and stack values for those applications that need higher memory amounts than 4k heap and 4k stack, which is the new default (half of the previous 16k). I hope to get most ELKS applications running in 4k total heap and stack. The -maout-chmem= option should only be used for the applications that require it individually in Makefiles, not for the entire Makefile. A few days ago, the stdio BUFSIZ was lowered from 2048 to 1024 to further reduce heap requirements, considering that no ELKS filesystem read will ever return > 1024 bytes, based on L2 buffer size-limited I/O.

Putting the crontab file in /var/spool/cron/root uses 2k extra disk space for every disk (two extra directories), for every floppy, whether they use cron or not. I'd like to have all cron disk usage moved to just /var. Ideally, that file would be changed to /var/cron-root or something similar, eliminating other directories. I plan on moving /var/run/utmp to /var/utmp as well. The directory and filename should be \#define'd in cron source so they're easily changed should ELKS be further adapted.

elkscmd/cron/crontab (binary file) should be deleted from this PR. Create an elkscmd/cron/.gitignore including cron and crontab.

Thanks!

georgp24:
The directory `/var/spool/cron/` is used by all Linux distros as the location for the crontab file or files. They usually get cron running on startup and the users put their crontab files at this location. As ELKS just has one user this file is called root. I used the standard location for Linux. I could make it `var/cron` for ELKS. I do not want to move away much from this because all the tutorials and documention on the internet for cron are mentioning this location. 
The locations are specified in #define statements in the code.

I will look into the memory requirements. Crontab is just executed temporarily to edit or generate a crontab file in the var/spool/cron directory. This probably does not need the chmem option.

ghaerr:
Thanks Georg, I'll commit this now.

Since you're now checking for /var/cron to exist and returning a possible error, I think an even better approach, given than /var always exists, would be to not have a pre-built rootfs_template /var/cron/root, but instead to programmatically mkdir the /var/cron directory as well as create the root file when checking, instead of returning an error. An error would only be returned if /var/cron or /var/cron/root could not be created and that would be one less thing a user would have to worry about, in the case he wanted to run cron on a non-boot or non-distribution disk. This has the benefit of requiring less distribution disk space until used, and unlikely to error. Also, your cron system would be completely self contained and not require rootfs template files in order to work.

georgp24:
Ok, I will implement that.

ghaerr:
Ok, great! Just push it in your open #625 and I'll commit both afterwards.

PR #621, opened 2020-05-10
Proposed support for "medium memory model" programs in ELKS kernel

tkchia:
This PR adds initial support for some form of "medium memory model" `a.out` programs to the ELKS kernel.

The format for such programs is (sort of) documented in the updated `elks/include/linuxmt/minix.h` file.  It adds optional fields to the `a.out` header to allow a 64 KiB "far text" section, in addition to the existing near text, data, and BSS sections.  Segment relocations are 8 bytes each, per Minix's documented relocation format.

I also bumped up the `gcc-ia16` toolchain version and added a linker script, to allow experimenting with writing programs for this new format.

This is a follow-up on https://github.com/tkchia/build-ia16/issues/14 .

Thank you!

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -232,11 +232,11 @@ int sys_brk(__pptr newbrk)
             return -ENOMEM;
 		}
     }
-#ifdef CONFIG_EXEC_ELKS
+#ifdef CONFIG_EXEC_LOW_STACK
     if (newbrk > currentp->t_endseg) {
         return -ENOMEM;
     }
-#endif /* CONFIG_EXEC_ELKS */
+#endif /* CONFIG_EXEC_LOW_STACK */
     currentp->t_endbrk = newbrk;
 
     return 0;
--- a/elks/elks-medium.ld
+++ b/elks/elks-medium.ld
@@ -1,0 +1,35 @@
+OUTPUT_FORMAT(elks)
+ENTRY(_start)
+
+SECTIONS {
+	.text 0 : {
+		CREATE_OBJECT_SYMBOLS
+		*(.text .text.*
+		  /*
+		   * Place the text segments from about half the input
+		   * modules into the near text segment.  Let the rest go
+		   * into .fartext .
+		   */
+		  .fartext.f.*0$ .fartext.f.*2$ .fartext.f.*4$ .fartext.f.*6$)
+		. = ALIGN(0x10);
+	}
+	.fartext 0 : AT(0x10000) {
+		*(.fartext .fartext$ .fartext.*)
+		. = ALIGN(0x10);
+	}
+	.data 0 : AT(0x20000) {
+		*(.nildata .nildata.*)
+		*(.rodata .rodata.*)
+		*(.data .data.*)
+		CONSTRUCTORS
+		. = ALIGN(0x10);
+	}
+	.bss : {
+		*(.bss .bss.*)
+		*(COMMON)
+		. = ALIGN(0x10);
+		ASSERT (. + 0x100 <= 0xfff0,
+		    "Error: too large for a medium-model ELKS a.out file.");
+	}
+	/DISCARD/ : { *(*) }
+}
--- a/elks/fs/config.in
+++ b/elks/fs/config.in
@@ -40,6 +40,6 @@ mainmenu_option next_comment
 
 	comment 'Executable file formats'
 
-	bool 'Support ELKS file format'        CONFIG_EXEC_ELKS           y
+	bool 'Support medium memory model'     CONFIG_EXEC_MMODEL         y
 
 endmenu
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -48,8 +48,7 @@
 #include <arch/segment.h>
 
 static struct minix_exec_hdr mh;
-#ifdef CONFIG_EXEC_ELKS
-static struct minix_supl_hdr msuph;
+#ifdef CONFIG_EXEC_MMODEL
 static struct elks_supl_hdr esuph;
 #endif
 
@@ -58,7 +57,7 @@ static struct elks_supl_hdr esuph;
 #define INIT_HEAP  0x0     // For future use (inverted heap and stack)
 #define INIT_STACK 0x2000  // 8 K (space for both stack and heap)
 
-#ifdef CONFIG_EXEC_ELKS
+#ifdef CONFIG_EXEC_MMODEL
 // Read relocations for a particular segment and apply them
 // Only IA-16 segment relocations are accepted
 static int relocate(seg_t place_base, lsize_t rsize, segment_s *seg_code,
@@ -176,20 +175,35 @@ int sys_execve(char *filename, char *sptr, size_t slen)
     // as we now master the executable header content
     // with the new GNU build tool chain (custom LD script)
 
+			/* executable with no far text, no relocations */
+#define EXEC_HDR_SIZE_0	sizeof(struct minix_exec_hdr)
+			/* executable with relocations, no far text (?) */
+#define SUPL_HDR_SIZE_1	offsetof(struct elks_supl_hdr, esh_ftseg)
+#define EXEC_HDR_SIZE_1	(EXEC_HDR_SIZE_0 + SUPL_HDR_SIZE_1)
+			/* executable with far text (and maybe relocations) */
+#define SUPL_HDR_SIZE_2	(sizeof(struct elks_supl_hdr) - SUPL_HDR_SIZE_1)
+#define EXEC_HDR_SIZE_2 (EXEC_HDR_SIZE_1 + SUPL_HDR_SIZE_2)
+
     switch ((unsigned) mh.hlen) {
-#ifdef CONFIG_EXEC_ELKS
-    case 0x30:
-    case 0x40:
+    case EXEC_HDR_SIZE_0:
+#ifdef CONFIG_EXEC_MMODEL
+	memset(&esuph, 0, sizeof(esuph));
+#endif
+	break;
+#ifdef CONFIG_EXEC_MMODEL
+    case EXEC_HDR_SIZE_1:
+    case EXEC_HDR_SIZE_2:
 	/* BIG HEADER */
-	retval = filp->f_op->read(inode, filp, (char *) &msuph, sizeof(msuph));
-	if (retval != (int)sizeof(msuph)) {
+	retval = filp->f_op->read(inode, filp, (char *) &esuph,
+				  SUPL_HDR_SIZE_1);
+	if (retval != SUPL_HDR_SIZE_1) {
 	    debug1("EXEC: Bad secondary header, result %u\n", retval);
 	    goto error_exec3;
 	}
-	if (mh.hlen == 0x40) {
-	    retval = filp->f_op->read(inode, filp, (char *) &esuph,
-				      sizeof(esuph));
-	    if (retval != (int)sizeof(esuph)) {
+	if (mh.hlen == EXEC_HDR_SIZE_2) {
+	    retval = filp->f_op->read(inode, filp, (char *) &esuph.esh_ftseg,
+				      SUPL_HDR_SIZE_2);
+	    if (retval != SUPL_HDR_SIZE_2) {
 		debug1("EXEC: Bad secondary header, result %u\n", retval);
 		goto error_exec3;
 	    }
@@ -198,29 +212,29 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 	    if (esuph.esh_ftrsize % sizeof(struct minix_reloc) != 0)
 		goto error_exec3;
 	} else
-	    memset(&esuph, 0, sizeof esuph);
+	    memset(&esuph.esh_ftseg, 0, SUPL_HDR_SIZE_2);
 	debug3("EXEC: text reloc size 0x%lx, data reloc size 0x%lx, "
 	       "text base 0x%lx\n",
-	       msuph.msh_trsize, msuph.msh_drsize, msuph.msh_tbase);
-	if (msuph.msh_trsize % sizeof(struct minix_reloc) != 0 ||
-	    msuph.msh_drsize % sizeof(struct minix_reloc) != 0 ||
-	    msuph.msh_tbase != 0)
+	       esuph.msh_trsize, esuph.msh_drsize, esuph.msh_tbase);
+	if (esuph.msh_tbase != 0)
 	    goto error_exec3;
-	base_data = msuph.msh_dbase;
+	if (esuph.msh_trsize % sizeof(struct minix_reloc) != 0 ||
+	    esuph.msh_drsize % sizeof(struct minix_reloc) != 0)
+	    goto error_exec3;
+#ifdef CONFIG_EXEC_LOW_STACK
+	base_data = esuph.msh_dbase;
 	if (base_data & 0xf)
 	    goto error_exec3;
 	if (base_data != 0) {
 	    debug1("EXEC: New type executable stack = %x\n", base_data);
 	    len = mh.dseg + mh.bseg + base_data + INIT_HEAP;
 	}
-	break;
-#endif
-    case 0x20:
-#ifdef CONFIG_EXEC_ELKS
-	memset(&msuph, 0, sizeof msuph);
-	memset(&esuph, 0, sizeof esuph);
+#else
+	if (base_data != 0)
+	    goto error_exec3;
 #endif
 	break;
+#endif
     default:
 	goto error_exec3;
     }
@@ -237,7 +251,7 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 	segext_t paras;
 	retval = -ENOMEM;
 	paras = (mh.tseg + 15) >> 4;
-#ifdef CONFIG_EXEC_ELKS
+#ifdef CONFIG_EXEC_MMODEL
 	paras += (esuph.esh_ftseg + 15) >> 4;
 #endif
 	debug1("EXEC: Allocating 0x%x paragraphs for text segment(s)\n",
@@ -251,7 +265,7 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 		   retval, (unsigned)mh.tseg);
 	    goto error_exec4;
 	}
-#ifdef CONFIG_EXEC_ELKS
+#ifdef CONFIG_EXEC_MMODEL
 	if (esuph.esh_ftseg) {
 	    currentp->t_regs.ds = seg_code->base + ((mh.tseg + 15) >> 4);
 	    retval = filp->f_op->read(inode, filp, 0, (size_t)esuph.esh_ftseg);
@@ -265,7 +279,7 @@ int sys_execve(char *filename, char *sptr, size_t slen)
     } else {
 	seg_get (seg_code);
 	filp->f_pos += mh.tseg;
-#ifdef CONFIG_EXEC_ELKS
+#ifdef CONFIG_EXEC_MMODEL
 	filp->f_pos += esuph.esh_ftseg;
 	need_reloc_code = 0;
 #endif
@@ -287,10 +301,10 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 	goto error_exec5;
     }
 
-#ifdef CONFIG_EXEC_ELKS
+#ifdef CONFIG_EXEC_MMODEL
     if (need_reloc_code) {
 	/* Read and apply text segment relocations */
-	if (relocate(seg_code->base, msuph.msh_trsize, seg_code, seg_data,
+	if (relocate(seg_code->base, esuph.msh_trsize, seg_code, seg_data,
 		     currentp, inode, filp) != 0)
 	    goto error_exec5;
 	/* Read and apply far text segment relocations */
@@ -299,11 +313,11 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 	    goto error_exec5;
     } else {
 	/* If reusing existing text segments, no need to re-relocate */
-	filp->f_pos += msuph.msh_trsize;
+	filp->f_pos += esuph.msh_trsize;
 	filp->f_pos += esuph.esh_ftrsize;
     }
     /* Read and apply data relocations */
-    if (relocate(seg_data->base, msuph.msh_drsize, seg_code, seg_data,
+    if (relocate(seg_data->base, esuph.msh_drsize, seg_code, seg_data,
 		 currentp, inode, filp) != 0)
 	goto error_exec5;
 #endif
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -117,7 +117,9 @@ int sys_execve(char *filename, char *sptr, size_t slen)
     segment_s * seg_code;
     segment_s * seg_data;
     lsize_t len;
+#ifdef CONFIG_EXEC_MMODEL
     int need_reloc_code = 1;
+#endif
 
     /* Open the image */
     /*debug1("EXEC: opening file: %s\n", filename);*/
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -177,54 +177,36 @@ int sys_execve(char *filename, char *sptr, size_t slen)
     // as we now master the executable header content
     // with the new GNU build tool chain (custom LD script)
 
-			/* executable with no far text, no relocations */
-#define EXEC_HDR_SIZE_0	sizeof(struct minix_exec_hdr)
-			/* executable with relocations, no far text (?) */
-#define SUPL_HDR_SIZE_1	offsetof(struct elks_supl_hdr, esh_ftseg)
-#define EXEC_HDR_SIZE_1	(EXEC_HDR_SIZE_0 + SUPL_HDR_SIZE_1)
-			/* executable with far text (and maybe relocations) */
-#define SUPL_HDR_SIZE_2	(sizeof(struct elks_supl_hdr) - SUPL_HDR_SIZE_1)
-#define EXEC_HDR_SIZE_2 (EXEC_HDR_SIZE_1 + SUPL_HDR_SIZE_2)
+#ifdef CONFIG_EXEC_MMODEL
+    memset(&esuph, 0, sizeof(esuph));
+#endif
 
     switch ((unsigned) mh.hlen) {
-    case EXEC_HDR_SIZE_0:
-#ifdef CONFIG_EXEC_MMODEL
-	memset(&esuph, 0, sizeof(esuph));
-#endif
+    case EXEC_MINIX_HDR_SIZE:
 	break;
 #ifdef CONFIG_EXEC_MMODEL
-    case EXEC_HDR_SIZE_1:
-    case EXEC_HDR_SIZE_2:
+    case EXEC_RELOC_HDR_SIZE:
+    case EXEC_FARTEXT_HDR_SIZE:
 	/* BIG HEADER */
 	retval = filp->f_op->read(inode, filp, (char *) &esuph,
-				  SUPL_HDR_SIZE_1);
-	if (retval != SUPL_HDR_SIZE_1) {
+				  mh.hlen - EXEC_MINIX_HDR_SIZE);
+	if (retval != mh.hlen - EXEC_MINIX_HDR_SIZE) {
 	    debug1("EXEC: Bad secondary header, result %u\n", retval);
 	    goto error_exec3;
 	}
-	if (mh.hlen == EXEC_HDR_SIZE_2) {
-	    retval = filp->f_op->read(inode, filp, (char *) &esuph.esh_ftseg,
-				      SUPL_HDR_SIZE_2);
-	    if (retval != SUPL_HDR_SIZE_2) {
-		debug1("EXEC: Bad secondary header, result %u\n", retval);
-		goto error_exec3;
-	    }
-	    debug2("EXEC: far text size 0x%x, far text reloc size 0x%lx\n",
-		   esuph.esh_ftseg, esuph.esh_ftrsize);
-	    if (esuph.esh_ftrsize % sizeof(struct minix_reloc) != 0)
-		goto error_exec3;
-	} else
-	    memset(&esuph.esh_ftseg, 0, SUPL_HDR_SIZE_2);
-	debug3("EXEC: text reloc size 0x%lx, data reloc size 0x%lx, "
-	       "text base 0x%lx\n",
-	       esuph.msh_trsize, esuph.msh_drsize, esuph.msh_tbase);
+	debug4("EXEC: text reloc size 0x%lx, data reloc size 0x%lx, "
+	       "far text reloc size 0x%x, text base 0x%lx\n",
+	       esuph.msh_trsize, esuph.msh_drsize, esuph.esh_ftrsize,
+	       esuph.msh_tbase);
 	if (esuph.msh_tbase != 0)
 	    goto error_exec3;
 	if (esuph.msh_trsize % sizeof(struct minix_reloc) != 0 ||
-	    esuph.msh_drsize % sizeof(struct minix_reloc) != 0)
+	    esuph.msh_drsize % sizeof(struct minix_reloc) != 0 ||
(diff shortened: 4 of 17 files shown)


ghaerr:
Hello @tkchia,

Thank you for your support in research and development of far code segments for ELKS!
Hopefully this work will allow the eventual possibility of running the kernel in medium model as an option.

I have a few questions and comments.

It would be nice to have an example application, or perhaps a Makefile option to compile an existing application in elkscmd/ medium model, to learn what new compiler and/or linker options are necessary, and test and observe the kernel loading, etc.

With regards to medium model libc support, is that entirely turned off for the time being, or just the .s files... that is, is there partial support yet for linking libc with an ELKS application, or does crt0.S and other .s files need conversion first?

I would like to see the CONFIG_EXEC_ELKS remain separate from this new work. For the time being, I'm thinking of leaving in all CONFIG_EXEC_ELKS code but setting that default OFF. It would be nice to have this new option available under a separate config option, suggest perhaps CONFIG_EXEC_MMODEL. In that way, either option could be configured separately, since we are down to less than 9k code space available with FAT FS, serial and networking compiled in.

Suggest removing \#ifdef CONFIG_EXEC_ELKS protection in minix.h, not needed and improves readability, a comment could suffice instead.

Do we really need an 8-byte minix_reloc? This certainly isn't important immediately, but I feel longer term it very important to keep on-disk (and in-memory reloc) size to a minimum given the systems ELKS runs on. The MZ format successfully uses 4-byte values, can you see a path where this could be shrunk, even though we're also supporting multiple segments rather than flat MZ-style? I haven't studied the BFD backend at all, it seems it has been derived from i386 executable support and has yet to discard 32 bit support in favor of 16-bit only... or is that the intent?

Thank you!

tkchia:
Hello @ghaerr,

> I would like to see the CONFIG_EXEC_ELKS remain separate from this new work. For the time being, I'm thinking of leaving in all CONFIG_EXEC_ELKS code but setting that default OFF. It would be nice to have this new option available under a separate config option, suggest perhaps CONFIG_EXEC_MMODEL. In that way, either option could be configured separately, since we are down to less than 9k code space available with FAT FS, serial and networking compiled in.

Done.  I have subdivided `CONFIG_EXEC_ELKS` further into `CONFIG_EXEC_LOW_STACK` (whether to allow stack-before-data programs) and `CONFIG_EXEC_MMODEL` (medium model).

> With regards to medium model libc support, is that entirely turned off for the time being, or just the .s files... that is, is there partial support yet for linking libc with an ELKS application, or does crt0.S and other .s files need conversion first?

Yes, the `.s` and `.S` files in `libc` will need to be updated to work properly under the medium model.

In fact, I found that the `signal(`_signum_`,` _handler_`);` syscall interface will also need updating --- _i.e. a change in the kernel ↔ userland ABI_. 😑  Namely, on an asynchronous event, the kernel assumes that _handler_ can simply return to a near (16-bit) address in the same segment as _handler_ itself.  But this may not be true if an asynchronous event happens while a process is running far code.

My current proposal is to keep the _handler_ pointer as 16-bit --- and limit it to the near code segment --- but require that the handler routine return to a 32-bit address.  (_handler_ will most likely be `libc`'s `_syscall_signal` anyway, if it is not `SIG_DFL` etc.)

> It would be nice to have an example application, or perhaps a Makefile option to compile an existing application in elkscmd/ medium model, to learn what new compiler and/or linker options are necessary, and test and observe the kernel loading, etc.

Thanks, I am working on this.  (Since there is no `libc` (including `crt0`) support yet, currently programs must be built using `-nostdlib`.  It is also necessary to specify with `-T`... the location of the linker script, as it is not yet installed in the right place.)

Thank you!

ghaerr:
Hello @tkchia!

> I have subdivided CONFIG_EXEC_ELKS further into CONFIG_EXEC_LOW_STACK (whether to allow stack-before-data programs) and CONFIG_EXEC_MMODEL (medium model).

Thank you, this is looking a lot better.

I am wondering, though, what is the purpose of keeping CONFIG_EXEC_ELKS? Shouldn't we just keep the two new options, and eliminate CONFIG_EXEC_ELKS entirely? This is going to get possibly quite a bit more (needlessly) complicated when heap and stack requirements are added to the header. I recommend removing CONFIG_EXEC_ELKS now, for reasons below.

Looking more closely, it seems that MMODEL support is now working with two supplemental headers, "minix_supl_hdr", and "elks_supl_hdr". IMO, we would be better off with using a single supplemental header for each option, which would be a union. There no known tools which actually produce the "minix_supl_hdr" and run with its intended low stack implementation. Now, we are creating yet another supplemental header which requires a previously unused one.

Keeping a single complete supplemental header definition for each available config option, but combining them into a union, simplifies exec.c code (allows a single read of the entire extended header, no separate code for separate options with same information, etc), and moves the design issues into one place - minix.h, without having to modify exec.c heavily when a header changes (as it might, as we improve this and other pending designs).

Other thoughts are to move away from constants in exec.c like 0x30, 0x40 and 0x20, and replace these with well-named defines based on sizeof(minix_exec_hdr)+sizeof(option_supl_hdr). Each supplemental header would be renamed closer to the option which selects it. The union of extended headers would be named elks_supl_hdr.

Because there are likely to be continuing (possibly temporary, for development) changes to a.out header formats, IMO it will be far simpler to keep only one complete supplemental header per option, which will also match (possibly temporary or developmental) linker scripts. This would be my plan for finally implementing configurable process heap + stack size, while also allowing development of medium model support, and changing mostly minix.h, rather than exec.c. When configurable heap+stack gets implemented, it will be a system-wide change that will require adoption of an extended header of some kind (unless we can use the last "unused" a.out header entry, which is/was possibly used for symbol table size). This system-wide change won't be configurable, as there are too many issues with tuning individual applications to their required heap/stack using compiler options, in the effort to get ELKS running well on smaller systems.

Other small issue:

Since CONFIG_EXEC_LOW_STACK will be turned off by default, and careful inspection of fork.c shows no relation to this option, I suggest changing the \#ifdef CONFIG_EXEC_LOW_STACK to \#if 1 for the `vfork` implementation. That is, `do_fork(0)` will always be executed for the time being.

Thank you for your work and consideration on this!

ghaerr:
Hello @tkchia,

> Yes, the .s and .S files in libc will need to be updated to work properly under the medium model.

Is there an easily-available macro/define that would allow picking up function argument offsets without using an include file? 

Looking at the crt0.S and the asm/ string functions, perhaps creating macros for 'call' and function epilogue and using a macro include file would allow for not having to modify .S files for different multilib cases, as well as removing lots of \#ifdefs.

> Namely, on an asynchronous event, the kernel assumes that handler can simply return to a near (16-bit) address in the same segment as handler itself. But this may not be true if an asynchronous event happens while a process is running far code.

Nice catch, I hadn't even thought about that problem!

> My current proposal is to keep the handler pointer as 16-bit --- and limit it to the near code segment --- but require that the handler routine return to a 32-bit address. (handler will most likely be libc's _syscall_signal anyway, if it is not SIG_DFL etc.)

I hope I'm understanding this properly - are you talking about the *handler* argument to libc `signal`, or the always-used `_syscall_signal`? I would think we would need libc `signal` to accept a 32-bit handler argument that is stored in _sigtable[], and far-called after the unchanged kernel code asynchronously near-calls _syscall_signal. The user program developer can't be expected to have a near-call application C function, which would likely always be a far function signal handler.

tkchia:
Hello @ghaerr,

> Looking more closely, it seems that MMODEL support is now working with two supplemental headers, "minix_supl_hdr", and "elks_supl_hdr". IMO, we would be better off with using a single supplemental header for each option, which would be a union. There no known tools which actually produce the "minix_supl_hdr" and run with its intended low stack implementation. Now, we are creating yet another supplemental header which requires a previously unused one.

The medium model support _does_ make use of the `.msh_trsize` and `.msh_drsize` fields of the `minix_supl_hdr` --- to keep track of the sizes of the near text and data relocation records.  The stack-before-data stuff uses only the `.msh_dbase` field.  I guess I will need to think about this a bit more...

Thank you!

tkchia:
Hello @ghaerr,

> Is there an easily-available macro/define that would allow picking up function argument offsets without using an include file?

There are C preprocessor macros (for `.S` files, though not for `.s` files) `__TINY__`, `__SMALL__`, and `__MEDIUM__` (plus `__COMPACT__`, `__LARGE__`, and `__HUGE__`, reserved for the future).  One of these will be defined according to the memory model specified to GCC.  If `crt0.S` (say) includes a header file, the header can test for these macros.

Other than that, though, there are no predefined macros for argument offsets.

> > My current proposal is to keep the handler pointer as 16-bit --- and limit it to the near code segment --- but require that the handler routine return to a 32-bit address. (handler will most likely be libc's `_syscall_signal` anyway, if it is not `SIG_DFL` etc.)

> I hope I'm understanding this properly - are you talking about the handler argument to `libc` signal, or the always-used `_syscall_signal`?

I meant the `_syscall_signal` --- sorry for the confusion.

Another possibility I am thinking of is to have the `signal` syscall actually accept a 32-bit `_syscall_signal` pointer, and modify the kernel's `struct signal_struct` to store only this one 32-bit pointer.  The rest of the `struct signal_struct` then only needs to record one of 3 values per signal type: whether the signal's disposition is `SIG_DFL`, `SIG_IGN`, or the custom handler.

Perhaps I will try to tackle this particular subproblem in a separate pull request.

Thank you!

ghaerr:
> The medium model support does make use of the .msh_trsize and .msh_drsize fields of the minix_supl_hdr --- to keep track of the sizes of the near text and data relocation records. The stack-before-data stuff uses only the .msh_dbase field. I guess I will need to think about this a bit more...

Yes, I noticed that. I'm just suggesting that you add all the minix_supl_hdr fields into your header, call it mmodel_supl_hdr or something like that, and then put both into a union elks_sup_hdr which is used in exec.c. There isn't much reason to keep minix_supl_hdr around separately [EDIT: for your mmodel code], since its not even possible to use LOW_STACK. Frankly, I think the LOW_STACK option, including its header, should be kept around for informational purposes only, and have CONFIG_EXEC_LOW_STACK as well as CONFIG_EXEC_ELKS removed as options in config.in. This then leaves only CONFIG_EXEC_MMODEL as a config option, and allows us to move forward with a much cleaner slate. CONFIG_EXEC_LOW_STACK could remain in exec.c to look at, but there's not much point in keeping it configurable as it can't even be tested.

ghaerr:
> I meant the _syscall_signal --- sorry for the confusion.

Thanks, that makes more sense now.

> Another possibility I am thinking of is to have the signal syscall actually accept a 32-bit _syscall_signal pointer, and modify the kernel's struct signal_struct to store only this one 32-bit pointer. The rest of the struct signal_struct then only needs to record one of 3 values per signal type: whether the signal's disposition is SIG_DFL, SIG_IGN, or the custom handler.

Either option sounds OK to me. Having the kernel handle 32-bit code segment pointers for signal would allow not only `arch_setup_sighandler_stack` to get written and tested, but also perhaps pave the way for `arch_build_stack` and the scheduler `tswitch` code when/if the time comes for a medium-model kernel. I view the biggest benefit of the current medium model R&D would be a medium-model kernel, even as an experimental option, since it would result in the availability of a single ELKS kernel with all options enabled, for maximum usability.

> Perhaps I will try to tackle this particular subproblem in a separate pull request.

Good idea for both libc and signal issues; that will allow this commit to proceed unattached to all its subsequent requirements.

ghaerr:
Hello @tkchia,

>  just suggesting that you add all the minix_supl_hdr fields into your header, call it mmodel_supl_hdr or something like that, and then put both into a union elks_sup_hdr which is used in exec.c

Perhaps, we should consider now, the problem of how configurable heap+stack might get implemented, as thoughts on that were driving my previous comments on header formats.

I was imagining that heap+stack would require a supplemental header, unless the unsigned long "symtab_size" were (re)used, which I'm not sure is a good idea. Configurable a.out heap+stack would be attached to two optional new compiler options -maout-heap and -maout-stack, and if specified, much like medium model support, would use a specific linker script which would generate a specific a.out header. Of course, the standard elks-small.ld could be modified to always output the new header. Given that we've only got the hlen member to differentiate headers easily, I was thinking that exec.c would handle heap+stack headers with a single read of the extended header, shared with other extended read headers in a singlely-coded kernel read using hlen.

In this scenario, "old-style" executables produced from dev86 or bcc would run with an 8k stack+heap, and new-style would run with configured (or default if =0) heap+stack. Now enter MM support. Should it also support configurable heap+stack? Yes, but it might not at the developmental point where it is automatically added. Thus the union of headers. We don't really need a union header, but it comes in handy if for various reasons a particular format does not allow for other header-implemented features. Features (and headers) are effectively consolidated when the first few members of each header are identical. If this ends up being the case for all extended headers at some point in the future, then the union can be easily be removed and exec.c slightly simplified.

The entire idea is wrapped around allowing separate development to take place (in this case MM headers and stack+heap headers, along with possible information-only stack_low headers) to coexist completely separately, until enough experience and testing is gained to merge them when they are finished.

If you would rather, we can commit this PR now, and I can do the followup work. But I am thinking it would be nice to agree on an approach to move forward on with these issues considered, and would also like your input on compiler or other options for heap+stack. (The compiler options aren't strictly needed although useful; a separate utility `chmem` could be used, much like the ELKS `chmem` use to tune at runtime).

tkchia:
Hello @ghaerr,

> Either option sounds OK to me. Having the kernel handle 32-bit code segment pointers for signal would allow not only `arch_setup_sighandler_stack` to get written and tested, but also perhaps pave the way for `arch_build_stack` and the scheduler `tswitch` code when/if the time comes for a medium-model kernel.

Thank you.  I have sent in a PR (https://github.com/jbruchon/elks/pull/629) to change the `_signal` ABI; I think that can go ahead first.

> I was imagining that heap+stack would require a supplemental header, unless the unsigned long "symtab_size" were (re)used, which I'm not sure is a good idea. Configurable a.out heap+stack would be attached to two optional new compiler options -maout-heap and -maout-stack, and if specified, much like medium model support, would use a specific linker script which would generate a specific a.out header.

Perhaps we do not need a new type of header for this: maybe a new syscall to let a program set a maximum `sbrk` will do the trick.

Thank you!

ghaerr:
Hi @tkchia,

> Perhaps we do not need a new type of header for this: maybe a new syscall to let a program set a maximum sbrk will do the trick.

It is important that an application's heap/stack be able to be specified without having to add new code to it. 

While the separate heap size isn't actually required, a minimum stack size is what is missing, as that is now a fixed 4k and checked on sys_break requests. If a minimum stack could be specified, then the "stack+heap" total could remain in the a.out chmem/total field, instead of being specified seperately in the compiler command line.

Your idea of "no new header" gives me an idea that would work: using the upper 16-bits of a.out chmem/total - if the upper 16 bits are 0, (which is always the case now), then use the default min stack size. If non-zero, then use that value as the min stack. If the compiler allows a long value to be specified using -maout-chmem=, this solves the problem with no new header and no tools change! (And also puts to use more unused bits in a header longword).

ghaerr:
Hello @tkchia,

> If the compiler allows a long value to be specified using -maout-chmem=,

It seems any value for the a.out long chmem field is silently truncated in the elks-small.ld linker script to 0xFFF0.

After changing the linker script as follows:
```
        PROVIDE (_start = entry); /* `entry' was the old entry point symbol */
        PROVIDE (_chmem = 0);
        PROVIDE (_total = _chmem == 0 ? 0 : . + _chmem);
-       _total_adjusted = _total == 0 ? 0
-                         : MIN (0xfff0, MAX (. + 0x100, _total));
+       _total_adjusted = _total == 0 ? 0 : MAX (. + 0x100, _total);
```
I have a modified exec.c that correctly works to extract separate min stack and heap values.

Using this method, there seems no easy way to specify a minimum stack AND keep the 16-bit "chmem" field 0 for default allocation. Thus, using this method one needs to specify the heap and stack together.

It is possible with more arithmetic in the linker script, that could be achieved also.

BTW, what is the reason for adding 256 to the total specified? Is it for segment alignment? That should mean 16 would suffice.

If you are OK with this approach, it seems this could be made to work without the need for an additional header.

tkchia:
Hello @ghaerr,

> Frankly, I think the LOW_STACK option, including its header, should be kept around for informational purposes only, and have CONFIG_EXEC_LOW_STACK as well as CONFIG_EXEC_ELKS removed as options in config.in.

Thanks!  I have removed the `CONFIG_EXEC_LOW_STACK` option (though I have left most of the `#ifdef`'s in place).

I have also added C library support for the medium model to this PR.  You should now be able to cross-compile a medium model ELKS program by saying `ia16-elf-gcc -melks -mcmodel=medium` ... .

Thanks!

tkchia:
Hello @ghaerr,

> After changing the linker script as follows:
> ```
>         PROVIDE (_start = entry); /* `entry' was the old entry point symbol */
>         PROVIDE (_chmem = 0);
>         PROVIDE (_total = _chmem == 0 ? 0 : . + _chmem);
> -       _total_adjusted = _total == 0 ? 0
> -                         : MIN (0xfff0, MAX (. + 0x100, _total));
> +       _total_adjusted = _total == 0 ? 0 : MAX (. + 0x100, _total);
> ```
> I have a modified exec.c that correctly works to extract separate min stack and heap values.

I look forward to seeing your changes. :)  As for the linker script, how about something like this?
```
@@ -13,5 +13,6 @@ SECTIONS {
 		LONG (SIZEOF (.bss));
 		LONG (_start);    /* entry point */
-		LONG (_total_adjusted); /* total memory allocated */
+		SHORT (_total_adjusted); /* total memory allocated */
+		SHORT (_stack_adjusted); /* stack size */
 		LONG (0);        /* symbol table size */
 		}
@@ -26,5 +27,10 @@ SECTIONS {
 	}
 	PROVIDE (_start = entry); /* `entry' was the old entry point symbol */
-	PROVIDE (_chmem = 0);
+	PROVIDE (_stack = 0);
+	PROVIDE (_heap = 0);
+	PROVIDE (_stack_adjusted = _stack == 0 ? 0
+				   : MIN (0xfff0 - ., _stack));
+	PROVIDE (_chmem = _stack_adjusted == 0 || _heap == 0 ? 0
+			  : _stack_adjusted + _heap);
 	PROVIDE (_total = _chmem == 0 ? 0 : . + _chmem);
 	_total_adjusted = _total == 0 ? 0
```

> BTW, what is the reason for adding 256 to the total specified? Is it for segment alignment? That should mean 16 would suffice.

The `. + 0x100` adds `0x100` to the linker's location counter (`.`) --- which at that point is at the end of the static data + BSS area.  My original idea was to try to make sure that there are at least `0x100` bytes available for the non-static (stack + heap) area.

Thank you!

ghaerr (commented):
Hello @tkchia,

For the work in a.out header and exec.c: Great job! This is coming along quite nicely :)

For the work in MM libc: Wow, you have been busy! Looks fantastic. I definitely like the use of the FAR_ADJ_, CALL_\*, and RET_ macros in the .S files, much improved and very easy to understand.

The work in codefile headers and exec.c look pretty much ready to go, I suggest the following for more readability and small cleanup:

Use `unsigned long` in the msh_\* first optional header fields.

Do we really need 'unused4' and 'unused5' in the secondary header? Removing them now allows for using hlen in the future if really needed. The 8086 really doesn't have any alignment issues nor need to have structure sizes paragraph aligned, and the header isn't read into executable memory, which would otherwise affect .text alignment.

Suggest renaming (to something like this):
EXEC_HDR_SIZE_0 to EXEC_MINIX_HDR_SIZE
EXEC_HDR_SIZE_1 to EXEC_RELOC_HDR_SIZE
EXEC_HDR_SIZE_2 to EXEC_FARTEXT_HDR_SIZE
and the SUPL_\* the same, and all moved to minix.h.

You've done a nice job keeping support for the EXEC_RELOC_HDR, using two reads to get the information required for MM support. This is nice for backwards compatibility and now, but after this is working, we really need to investigate `dev86` and determine whether any tools actually produce an EXEC_RELOC_HDR. Of course we could always modify the existing toolchain to support such an option. I somehow think neither will be needed. This can be taken up later.

There is quite a lot of error checking on sizeof(minix_reloc), and agree it is needed now. Should this be retained indefinitely? (Thinking of kernel code limits for the future).

For the libc MM work, do we need to dig up some applications for testing `setjmp`, `memcpy`, and all others changed? (In a separate thread, I just found a major error in the C version of `memchr`, apparently two years old from the switch from bcc which caused a C version to be used; it broke all `telnet` negotiations and no one noticed!) 

Are there more libc functions to convert for MM support?

Thank you very much for this excellent work!!

ghaerr:
Hello @tkchia,

>  As for the linker script, how about something like this?

Thank you, that helps a lot. I will use that in my ongoing development.

> I look forward to seeing your changes. :) 

I will wait until MM support (or at least this PR and associated updates) is committed, since it involves work in the same exec.c file. I'm quite deep in the networking code at the moment so good timing.

Thank you!

tkchia:
Hello @ghaerr,

> Use `unsigned long` in the `msh_*` first optional header fields.
> ...
> Suggest renaming (to something like this):
> `EXEC_HDR_SIZE_0` to `EXEC_MINIX_HDR_SIZE`
> `EXEC_HDR_SIZE_1` to `EXEC_RELOC_HDR_SIZE`
> `EXEC_HDR_SIZE_2` to `EXEC_FARTEXT_HDR_SIZE`
> and the `SUPL_*` the same, and all moved to `minix.h`.

Thanks for the suggestions!  I have made the proposed changes, and also somewhat simplified the code for dealing with the different header sizes.

> Do we really need 'unused4' and 'unused5' in the secondary header? Removing them now allows for using hlen in the future if really needed. The 8086 really doesn't have any alignment issues nor need to have structure sizes paragraph aligned, and the header isn't read into executable memory, which would otherwise affect .text alignment.

The `unused4` and `unused5` padding fields are probably not strictly necessary.  But if we are going to use the same file format for the kernel, the padding may make things a little bit easier for `setup.S` --- it currently relies on the kernel's text segment starting on a paragraph boundary.

> There is quite a lot of error checking on sizeof(minix_reloc), and agree it is needed now. Should this be retained indefinitely? (Thinking of kernel code limits for the future).

I think yes, they are useful to keep around indefinitely --- and in fact, I just further added code to check that `unused3`, `unused4`, `unused5` are indeed zero.  I think the more stringent checking will help to ensure forward compatibility with future `a.out` formats which put these fields to some use --- I would like an older kernel to fail gracefully on such newer `a.out`'s, rather than weird out.

(I think at some point ELKS will need to give programs some structured way to dynamically emit or modify their own code, and maybe some sort of `ld.so`-like mechanism to let userland code handle more complex types of `a.out` files.)

Thank you!

ghaerr (commented):
Hello @tkchia,

Moving to only one additional read for the header is a nice improvement.

Since we're down at the finer details now, and always concerned about what the compiler generates, I have a few more observations, only concerned about extreme efficiencies. Overall, this looks quite nice!

In many cases, generating code for longword compares is not needed:

Line 192 & 193: Using (int)mh.hlen would be preferred, since not sure the compiler wouldn't do a longword subtract before converting to int.

When a header field is known to be always be generated as zero (say for the upper bits of a longword size field on 8086's where the size will never be > 0xFFFF), and the upper bits not used, I question whether it is useful, or a good idea, to do longword compares, especially when the ultimate value is truncated to 16 bits anyways. Thus, some consideration could be given to using (int) to cast off the upper bits in lines 201-205. I say this, knowing the implications, but consider that all of the msh_ header members should really be unsigned short now, but are being left as unsigned long, just for the purpose of backwards-compatibility with a tool that likely doesn't exist. Do we burden our kernel with longword compares for these edge cases? I think not. The header fields should stay longwords, but I recommend using (int) casting on 201-205 and all other places possible.

While I agree that checking header values for data that is actually used is probably good (lower 16 bits of lines 203-205 are modulo 8), I think that checking "unused" header values for 0 (line 206) is going the wrong direction: these values are not used, so why check them? Should those values be needed for, perhaps, a userland enhancement, the kernel won't load the executable. I suggest thinking about the idea that those three  members be named "reserved", rather than unused, and not checked. (And not checked for longword compares either :) )

`relocate` doesn't need, and probably shouldn't, pass `currentp`, but just use `current` internally, it will never be used to relocate a non-current task. The reason currentp was declared in `exec` was to find a previously-loaded text segment, but it is then reset on line 97. Its subsequent use as a register pointer to the current task struct stems from the bcc days for efficiency, but I wonder the efficiency (and larger code generated) now with gcc using what is likely not stored in a register instead of the global current pointer. There are a lot of compiler assumptions like this in the kernel which should likely be removed. (The greatest of which you've likely seen - declaration and use of a char \* that is then used as an integer with lots of (int) and (char \*) castings all over).

Thank you!

tkchia:
Hello @ghaerr,

> Since we're down at the finer details now, and always concerned about what the compiler generates, I have a few more observations, only concerned about extreme efficiencies. Overall, this looks quite nice!

Thanks!

> Line 192 & 193: Using (int)mh.hlen would be preferred, since not sure the compiler wouldn't do a longword subtract before converting to int.

😮!!  It seems I missed that `mh.hlen` was actually a 32-bit longword in `elks/include/linuxmt/minix.h`.  In `elks/arch/i86/tools/a.out.h`, the 4 bytes are actually considered to be 3 fields
```
    unsigned char a_hdrlen;	/* length of header */
    unsigned char a_unused;	/* reserved for future use */
    unsigned short a_version;	/* version stamp (not used at present) */
```

I have fixed the treatment of the header length field accordingly, and hope to work on the other 32-bit fields.

Thank you!

tkchia:
Hello @ghaerr,

> Since we're down at the finer details now, and always concerned about what the compiler generates

One quick way I am using to gauge the size of the compiled code is to run `ia16-elf-objdump -D -r -Mi8086` ... on the intermediate `.o` files.  Currently (e.g.) it shows that the `sys_execve` routine takes up `0x588` bytes, and the internal `relocate` routine `0xb7` bytes.  This can probably be further improved.

Thank you!

tkchia:
Hello @ghaerr,

I have made the changes.  Do let me know if you think the patches are good to go.  Thank you!

georgp24:
@tkchia 
You wrote: "There are C preprocessor macros (for .S files, though not for .s files) __TINY__, __SMALL__, and __MEDIUM__"
Does this mean, you can specify `-mcmodel=tiny` when compiling a C program and "just" the ELKS program loader will not load these programs?

ghaerr:
Hello @tkchia,

> One quick way I am using to gauge the size of the compiled code is to run ia16-elf-objdump -D -r -Mi8086 ... on the intermediate .o files.

Thank you, that will be very useful, I was not aware of the disassembly capabilities. I had been resorting to adding -S to Makefiles.

I am now curious to check whether gcc is smart enough to convert 
```
if (esuph.msh_trsize % sizeof(struct minix_reloc) != 0)
```
to  ... `mov esuph+msh_trsize,%ax; and $7,%ax; jnz ...` and skip the longword modulo.

I see you've brought out the light saber for removing cseg/data/bss longword adds and paragraph conversions :) Looks good, I wasn't aware of `__builtin_add_overflow`. That stuff may want to go into a linuxmt/arch header file, but I'm fine with leaving it in exec.c for now.

There is still a bit of extra `base_data` handling even though without CONFIG_LOW_STACK it is guaranteed to be zero. I'm thinking that the code on line 233, and 245-248 be moved to the CONFIG_LOW_STACK section after line 221 (and 232 to 216) to keep it self-contained.

What else needs to be done to test-build MM applications after this is committed?

tkchia:
Hello @ghaerr,

> I'm thinking that the code on line 233, and 245-248 be moved to the CONFIG_LOW_STACK section after line 221 (and 232 to 216) to keep it self-contained.

I have tried to consolidate the special handling of `base_data` and `CONFIG_LOW_STACK` into two places.

> What else needs to be done to test-build MM applications after this is committed?

Other than further testing, the remaining big piece will be to get `elksemu` to handle medium model executables --- I will probably try to do this in a separate pull request.

Thank you!

ghaerr:
Hello @tkchia,

This PR looks very nicely done, please commit.

Thank you!

tkchia:
Thanks @ghaerr !

ghaerr:
Hello @tkchia,

I have thought through the kernel changes for the configurable process heap and stack, and have a couple questions about the linker script you suggested:
```
@@ -13,5 +13,6 @@ SECTIONS {
 		LONG (SIZEOF (.bss));
 		LONG (_start);    /* entry point */
-		LONG (_total_adjusted); /* total memory allocated */
+		SHORT (_total_adjusted); /* total memory allocated */
+		SHORT (_stack_adjusted); /* stack size */
 		LONG (0);        /* symbol table size */
 		}
@@ -26,5 +27,10 @@ SECTIONS {
 	}
 	PROVIDE (_start = entry); /* `entry' was the old entry point symbol */
-	PROVIDE (_chmem = 0);
+	PROVIDE (_stack = 0);
+	PROVIDE (_heap = 0);
+	PROVIDE (_stack_adjusted = _stack == 0 ? 0
+				   : MIN (0xfff0 - ., _stack));
+	PROVIDE (_chmem = _stack_adjusted == 0 || _heap == 0 ? 0
+			  : _stack_adjusted + _heap);
 	PROVIDE (_total = _chmem == 0 ? 0 : . + _chmem);
 	_total_adjusted = _total == 0 ? 0
```
I don't quite understand how the passed `-maout-chmem=` value is passed to the _stack_adjusted value in the new header, given that _stack seems to be always set to 0.

More specifically, I am trying to handle the following:
```
Proposed change will add minstack field in high 16 bits of previous chmem longword.
If either field is zero, default settings would apply.
The previous total (chmem) field would change conservatively from data+bss+heap+stack
to data+bss+heap to be minix-compatible, with the stack allocated seperately as MIN_STACK.

For backwards compatibility, interpret non-zero chmem as heap-only, instead
of heap+stack, and add default MIN_STACK to it when allocating the program data segment.

The kernel changes are minimal, as follows:
A task struct t_minstack is added, which is set to either MIN_STACK (4k) or minstack a.out field.
This is only used in arch/i86/mm/malloc.c during sys_brk checks.

INIT_HEAP   4096    unless chmem != 0, use chmem less data+bss
INIT_STACK  4096    unless minstack != 0, use minstack
This would have the effect of allocating 4k extra for programs compiled with old 
-maout-chmem= values, which in general likely won't happen.

The effective meaning of the a.out header fields would be:
chmem:      total data+bss+heap or 0 for default heap
minstack:   minstack or 0 for default stack
```
I am looking for suggestions on how to write a linker script to handle the following:
```
Proposed is change to interpretation of -maout-chmem= compiler argument,
implemented in the linker script, as follows:

                            NEW                         OLD
gcc -maout-chmem=0          default stack+default heap default stack+heap
gcc -maout-chmem=0x4000     default stack + 16k heap  16k data+bss+heap+stack
gcc -maout-chmem=0x0800     default stack + 2k heap   2k data+bss+heap+stack
gcc -maout-chmem=0x08000800 2k minstack, 2k heap (=4k heap+stack, with heap+data+bss added to chmem field)
gcc -maout-chmem=0x10002000 4k minstack, 8k heap (=12k heap+stack, with heap+data+bss added to chmem field)
gcc -maout=chmem=0x08000000 2k stack, no heap (=2k heap+stack, with heap+data+bss added to chmem field)
gcc -maout=chmem=0x04000000 1k stack, no heap
gcc -maout=chmem=0x04000400 1k stack, 1k heap
gcc -maout=chmem=0x00000001 default stack, no (1 byte) heap

In the proposed change, the linker script would be used to take the specified
separate stack and heap values, set the minstack separately, and  add in the 
heap+data+bss before writing the chmem field.

There is some thought of also adding the stack to the chmem header field, but that
causes other issues when the default stack (=0 minstack) is specified - one cannot
specify "default stack, heap=" in that case.
```

There are still some problems with what to do in the kernel if data+bss+heap+default_stack > 0xFFF0 when not separately specified, but I think a printk followed by reject will be fine. I don't think the linker script should silently truncate a specified heap or stack value, as this could allow a program to run with less than that specified.

Thank you!

tkchia:
Hello @ghaerr,

> I don't quite understand how the passed `-maout-chmem=` value is passed to the _stack_adjusted value in the new header, given that _stack seems to be always set to 0.

Sorry for the confusion.  I have just been [working](https://github.com/tkchia/gcc-ia16/commit/3da3c6820a9016a78797d1a5b124f7229f1e0675) on separate `-maout-stack=` and `-maout-heap=` parameters in the `gcc-ia16` front-end, which will end up defining the `_stack` and/or `_heap` symbols for the linker script.

Thank you!

Issue #622, opened 2020-05-10
Problem with wait()


georgp24:
I did not develop the `cron` program with the latest ELKS version and now found that it would not work with the latest version. The processes were not terminated and the system run out of memory. 

I found that the problem results from the change made to `wait.c` on the 4th of May. This stopped the `wait()` call in the cron program to work as expected. In the man page for `wait()` it says:
> The call wait(&status) is equivalent to: `waitpid(-1, &status, 0);`

Therefore I recommend to translate `wait()` to `waitpid` in wait.c.

I replaced the `wait()` call with `waitpid(-1, &status, 0)` and the `cron` program works with the latest ELKS version.

ghaerr:
Georg,

Thanks for testing. The general `wait` problem is unfortunately more complicated than just translating wait to waitpid. I fixed the libc wait() for another problem where it returned without waiting for a different reason.

It seems I still do not have it right, and will research this further and fix it.

I need to have available the original exact source code statements using `wait` are, so I know when I have it fixed properly. Can you leave the original statements in, commented out? That's the only way I'll be able to determine it working properly.

Thanks!

georgp24:
The cron version that does not work is in the PR #620 now. It is the wait in the runjob.c file at line 66. In my local version here I commented this out and made a waitpid() below that instead. I will upload that tomorrow.

georgp24:
The `cron` program uses `wait()` or `waitpid()` to reap the child process. When I committed my last version of cron, this worked with `waitpid()` but not with `wait().` With the latest version of ELKS it does not work with `waitpid()` too. So the child processes are no longer removed from the process table and after a while ELKS reports that it runs out of slots. This refers to `wait()` or `waitpid()` in `runjob.c.`

Also zombie processes are not removed by init. I cannot see if these are not properly reparented to `init` or what the reason is. They just stay "forever" in the process table.

ghaerr:
Hi @georgp24,

Thank you again for continued testing on this. Geez, it seems I just haven't been able to get `wait` working quite right. I had been seeing zombie processes pile up in `ps`, but tested this and thought `init` was getting them after a while.

> When I committed my last version of cron, this worked with waitpid() but not with wait(). 
> With the latest version of ELKS it does not work with waitpid() too.

I think both `wait` and `waitpid` are working to hold the parent process from running until the child exits in this version, please confirm. The problem is that the child processes are not being reaped, right? That is what I have been occasionally seeing.

I will fix this, thanks.

[...]

PR #625, opened 2020-05-11
Add .gitignore file to cron directory

georgp24:


ghaerr:
Thanks Georg!
 

Also, PR #628 should fix the `wait` problem you had to rewrite to `waitpid` for cron. Would you like to revert that to the original code, or just leave it?

Let me know, and I can leave this PR open until you change it and test, or I can commit now.

When you are satisfied, your issue #622 can be closed.

georgp24:
I would prefer to leave it with `waitpid`. So please commit.

PR #626, opened 2020-05-12
[ktcp] Further cleanup and enhancements to networking

ghaerr:
Further enhancements to `ktcp`, `sh` and pty driver for networking.
Detailed changes include:

Add @cjsthompson's compilation warning cleanup from #607 (thanks Christoph!)
Exec /bin/login instead of sash for `telnetd`
Turn off telnet local echo
Turn on telnet character mode: this allows use of `vi` in telnet sessions
Move `sh` linenoise dynamic edit linelen determination to an option, default off. This allows `sh` in telnet sessions.
Remove CONFIG_INET_STATUS config option
Add keyboard signal processing (^C) to pty driver
Allow for configurable NE2K base address in arch/i86/drivers/net/ne2k-mac.s (@Mellvik, if you can, please test on one of your configurable cards - it would be best try running at the 300h base address first and irq 9, then changing just the base address and recompiling). If this works, I will work to merge this with ports.h. There is a possibility that the 'tx_first_word' buffer addresses also need to be changed, see file for details.


Mellvik:
Thanks @ghaerr, this is moving forward at incredible speeds.

Looking forward to testing later today & tomorrow.

Quick question about coding style/practice:
When you have a situation like send_char() in slip.c where a function 'degenerates' into a single line of code, is there an advantage to keeping the function or might it as well be a macro?


--Mellvik

> Allow for configurable NE2K base address in arch/i86/drivers/net/ne2k-mac.s (@Mellvik, if you can, please test on one of your configurable cards - it would be best try running at the 300h base address first and irq 9, then changing just the base address and recompiling). If this works, I will work to merge this with ports.h. There is a possibility that the 'tx_first_word' buffer addresses also need to be changed, see file for details.
>

PR #631, opened 2020-05-14
[build] fix: beef up checking of gcc-ia16 toolchain versions

tkchia:
`build.sh` was assuming that the toolchain did not need to be rebuilt whenever `cross/.gcc.install` exists, but this fails if we need to upgrade to a newer set of `gcc-ia16` tools.

This fix instead makes `build.sh` hand over to `tools/build.sh` (and `tools/Makefile`), which will do a more complete check.  See https://github.com/jbruchon/elks/pull/629#issuecomment-628105141 .

I also tweaked `tools/Makefile` to handle cases where a tool's `.zip` file was downloaded only partially, and to clean away old versions of `.zip` files.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -26,7 +26,7 @@ clean_exit () {
 
 if [ "$1" != "auto" ]; then
 	mkdir -p "$CROSSDIR"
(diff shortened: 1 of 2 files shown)


ghaerr:
Hello @tkchia,

Thank you for this enhancement!

Issue #67, opened 2017-02-26
[ktcp] ARP sequencing error


mfld-fr:
The current implementation of the ARP module makes the assumption that the next incoming packet after sending an ARP request will be the expected ARP reply. It works in QEMU, but in a real & more loaded network, some other packets might come between the request and the reply. Therefore the implementation of ARP should be kept asynchronous as the IP module.

georgp24:
At the moment I do not expect a loaded network on ELKS. But I do not object to modify this.

mfld-fr:
In my own roadmap, networking in QEMU is nice but a preliminary step before networking on my real SBC, that is connected to my internet box, and thus "bombed" at a high rate with many packets that come at any time. So this improvement is a must have.

georgp24:
Rereading my post I agree that it indicates that I volunteered to do that. But if you need that I would prefer if you have a go at it.

The ARP request is not triggered by the socket interface but from ip.c resulting from a send packet by an application which is currently in progress. So making this asynchronous is not straightforward.

mfld-fr:
Hello Georg, no problem, I started working on it.

jbruchon:
Is this bug still present?

georgp24:
I cannot test it on real hardware and I expected Marc to fix this.

mfld-fr:
Yes, still present.

Mellvik:
I can't help fix this, but I can do the physical hardware testing. Let me know.

—Mellvik

ghaerr:
This three-year-old issue finally fixed and tested with PR #667. Closing issue.

mfld-fr:
Nice, thank you !

PR #632, opened 2020-05-15
[ktcp] First pass at fixing synchronous ARP request problem

ghaerr:
Tries to fix issue #67 for now by time-waiting for an ARP reply, then dropping the IP packet, rather than permanently hanging in an ARP request/reply cycle. This fix is necessary for telnet to run on QEMU without having to have an inbound connection first to fill the ARP table. 

There are several debug statements left on by default to determine asynchronous ARP reply effectiveness, RST processing, and packet refusals. Try running first without additional debug turned on to see any messages, then ping can be debugged by setting DEBUG_ARP and DEBUG_IP to 1 in config.h.

If ARP fails to work using this new method, turn on the old method, \#define ARP_WAIT_KLUGE in config.h.

Other changes:
Fix bad pointer memory corruption when TCP RST received.
Improve htons and htonl macros.
Endian cleanup.
ARP debugging, cleanup, and packet display.
IP packet display.
Add in_ntoa, cleanup IP address display code.
Fix console output with NUL character (received via telnet on login).

### DIFF
--- a/elks/arch/i86/drivers/char/console.c
+++ b/elks/arch/i86/drivers/char/console.c
@@ -317,8 +317,10 @@ static void std_char(register Console * C, char c)
 #ifdef CONFIG_CONSOLE_BIOS
 	PositionCursor(C);
 #endif
-	VideoWrite(C, c);
-	C->cx++;
+	if (c >= ' ') {
+		VideoWrite(C, c);
+		C->cx++;
+	}
       linewrap:
 	if (C->cx > MaxCol) {
 
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -68,7 +68,7 @@ file_utils/mv					:fileutil		:360k
 file_utils/ln					:fileutil			:720k
 file_utils/ls					:fileutil		:360k
 file_utils/rm					:fileutil		:360k
-file_utils/rmdir				:fileutil		:360k
+file_utils/rmdir				:fileutil			:720k
 file_utils/sync					:fileutil		:360k
 file_utils/touch				:fileutil			:720k
 sys_utils/chmem					:sysutil			:720k
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -39,21 +39,21 @@ struct arp_cache_s {
 
 typedef struct arp_cache_s arp_cache_t;
 
-static arp_cache_t _arp_cache [ARP_CACHE_MAX];
+static arp_cache_t arp_cache [ARP_CACHE_MAX];
 
 
 static void arp_cache_init (void)
 {
-	memset (_arp_cache, 0, ARP_CACHE_MAX * sizeof (arp_cache_t));
+	memset (arp_cache, 0, ARP_CACHE_MAX * sizeof (arp_cache_t));
 }
 
 
 int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr)
 {
-	arp_cache_t * entry = _arp_cache;
+	register arp_cache_t * entry = arp_cache;
 
 	/* First pair is the more recent */
-	while (entry < _arp_cache + ARP_CACHE_MAX) {
+	while (entry < arp_cache + ARP_CACHE_MAX) {
 		if (!entry->ip_addr)
 		    break;
 		if (entry->ip_addr == ip_addr) {
@@ -65,8 +65,7 @@ int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr)
 	}
 
 	debug_arp("arp: no cached entry for %s\n", in_ntoa(ip_addr));
-	memset(eth_addr, 0, sizeof(eth_addr_t));
-	return -1;
+	return 1;
 }
 
 
@@ -75,8 +74,8 @@ void arp_cache_add (ipaddr_t ip_addr, eth_addr_t * eth_addr)
 	if (arp_cache_get (ip_addr, eth_addr)) {
 
 		/* Shift the whole cache */
-		arp_cache_t * entry = _arp_cache + ARP_CACHE_MAX - 1;
-		while (entry > _arp_cache) {
+		arp_cache_t * entry = arp_cache + ARP_CACHE_MAX - 1;
+		while (entry > arp_cache) {
 			memcpy (entry, entry - 1, sizeof (arp_cache_t));
 			entry--;
 		}
@@ -90,7 +89,7 @@ void arp_cache_add (ipaddr_t ip_addr, eth_addr_t * eth_addr)
 }
 
 
-static char *mac_ntoa(unsigned char *p)
+char *mac_ntoa(unsigned char *p)
 {
     static char b[18];
 
@@ -98,7 +97,7 @@ static char *mac_ntoa(unsigned char *p)
     return b;
 }
 
-static void arp_print(struct arp *arp)
+static void arp_print(register struct arp *arp)
 {
 #if DEBUG_ARP
     printf("ARP: ");
@@ -121,44 +120,42 @@ int arp_init (void)
 
 void arp_reply(unsigned char *packet,int size)
 {
+    register struct arp *arp = (struct arp *)packet;
     struct arp_addr apair;
-    struct arp *arp_r;
 
-    arp_r = (struct arp *) packet;
-
-    debug_arp("arp: SEND reply to %s\n", in_ntoa(arp_r->ip_src));
+    debug_arp("arp: SEND reply to %s\n", in_ntoa(arp->ip_src));
 
     /* swap ip addresses and mac addresses */
-    apair.daddr = arp_r->ip_src;
-    apair.saddr = arp_r->ip_dest;
-    memcpy(apair.eth_dest, arp_r->eth_src, 6);
+    apair.daddr = arp->ip_src;
+    apair.saddr = arp->ip_dest;
+    memcpy(apair.eth_dest, arp->eth_src, 6);
     memcpy(apair.eth_src, eth_local_addr, 6);
 
     /* build arp reply */
-    arp_r->op = htons(ARP_REPLY);
-    memcpy(arp_r->ll_eth_dest, apair.eth_dest, 6);
-    memcpy(arp_r->ll_eth_src, apair.eth_src, 6);
+    arp->op = htons(ARP_REPLY);
+    memcpy(arp->ll_eth_dest, apair.eth_dest, 6);
+    memcpy(arp->ll_eth_src, apair.eth_src, 6);
 
-    memcpy(arp_r->eth_src, apair.eth_src, 6);
-    arp_r->ip_src=apair.saddr;
-    memcpy(arp_r->eth_dest, apair.eth_dest, 6);
-    arp_r->ip_dest=apair.daddr;
+    memcpy(arp->eth_src, apair.eth_src, 6);
+    arp->ip_src=apair.saddr;
+    memcpy(arp->eth_dest, apair.eth_dest, 6);
+    arp->ip_dest=apair.daddr;
 
-    arp_print(arp_r);
+    arp_print(arp);
     deveth_send(packet, sizeof (struct arp));
 }
 
 int arp_request(ipaddr_t ipaddress)
 {
-    static struct arp arpreq;
+    struct arp arpreq;
 
     debug_arp("arp: SEND request\n");
 
     /* build arp request */
     memset(arpreq.ll_eth_dest, 0xFF, 6);	/* broadcast*/
-	memcpy(arpreq.ll_eth_src, eth_local_addr, 6);
+    memcpy(arpreq.ll_eth_src, eth_local_addr, 6);
     /*specify below in big endian*/ //FIXME
-	arpreq.ll_type_len = ETH_TYPE_ARP;
+    arpreq.ll_type_len = ETH_TYPE_ARP;
     arpreq.hard_type=0x0100;
     arpreq.proto_type = ETH_TYPE_IPV4;
     arpreq.hard_len=6;
@@ -177,7 +174,7 @@ int arp_request(ipaddr_t ipaddress)
     struct timeval timeint;
     fd_set fdset;
     timeint.tv_sec  = 0;
-    timeint.tv_usec = 500000L;	/* 1/2 second*/
+    timeint.tv_usec = 200000L;	/* 200 msec*/
     FD_ZERO(&fdset);
     FD_SET(tcpdevfd, &fdset);
     int i = select(tcpdevfd + 1, &fdset, NULL, NULL, &timeint);
@@ -189,24 +186,22 @@ int arp_request(ipaddr_t ipaddress)
     return 1;		/* error*/
 }
 
-
 /* Process incoming ARP packets */
 void arp_recvpacket(unsigned char *packet, int size)
 {
-	struct arp * arp_r;
+	register struct arp *arp = (struct arp *)packet;
 
-	arp_r = (struct arp *) packet;
-	arp_print(arp_r);
-	switch (ntohs(arp_r->op)) {
+	arp_print(arp);
+	switch (ntohs(arp->op)) {
 	case ARP_REQUEST:
 		debug_arp("arp: incoming REQUEST\n");
-		arp_cache_add (arp_r->ip_src, arp_r->eth_src);
+		arp_cache_add (arp->ip_src, &arp->eth_src);
 		arp_reply (packet, size);
 		break;
 
 	case ARP_REPLY:
 		debug_arp("arp: incoming REPLY\n");
-		arp_cache_add (arp_r->ip_src, arp_r->eth_src);
+		arp_cache_add (arp->ip_src, &arp->eth_src);
 		break;
 	}
 }
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -14,6 +14,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
+#include <time.h>
 
 #include <linuxmt/arpa/inet.h>
 
@@ -25,20 +26,16 @@
 
 
 /* ARP operations */
-/* as big endian values */
-
-#define ARP_REQUEST  0x0100
-#define ARP_REPLY    0x0200
-
+#define ARP_REQUEST  1
+#define ARP_REPLY    2
 
 /* Local ARP cache */
-
 #define ARP_CACHE_MAX 5
 
 struct arp_cache_s {
 	ipaddr_t   ip_addr;   /* IPv4 address */
 	eth_addr_t eth_addr;  /* MAC address */
-	};
+};
 
 typedef struct arp_cache_s arp_cache_t;
 
@@ -53,32 +50,31 @@ static void arp_cache_init (void)
 
 int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr)
 {
-	int err = -1;
+	arp_cache_t * entry = _arp_cache;
 
 	/* First pair is the more recent */
-
-	arp_cache_t * entry = _arp_cache;
 	while (entry < _arp_cache + ARP_CACHE_MAX) {
-		if (!entry->ip_addr) break;
-
+		if (!entry->ip_addr)
+		    break;
 		if (entry->ip_addr == ip_addr) {
 			memcpy (eth_addr, entry->eth_addr, sizeof (eth_addr_t));
-			err = 0;
-			break;
+			debug_arp("arp: using cached entry for %s\n", in_ntoa(ip_addr));
+			return 0;
 		}
-
 		entry++;
 	}
 
-	return err;
+	debug_arp("arp: no cached entry for %s\n", in_ntoa(ip_addr));
+	memset(eth_addr, 0, sizeof(eth_addr_t));
+	return -1;
 }
 
 
 void arp_cache_add (ipaddr_t ip_addr, eth_addr_t * eth_addr)
 {
 	if (arp_cache_get (ip_addr, eth_addr)) {
-		/* Shift the whole cache */
 
+		/* Shift the whole cache */
 		arp_cache_t * entry = _arp_cache + ARP_CACHE_MAX - 1;
 		while (entry > _arp_cache) {
 			memcpy (entry, entry - 1, sizeof (arp_cache_t));
@@ -86,32 +82,35 @@ void arp_cache_add (ipaddr_t ip_addr, eth_addr_t * eth_addr)
 		}
 
 		/* Set the first pair as the more recent */
-
 		entry->ip_addr = ip_addr;
 		memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
-	}
+		debug_arp("arp: adding cache entry for %s\n", in_ntoa(ip_addr));
+	} else
+		debug_arp("arp: already cache entry for %s, ignored\n", in_ntoa(ip_addr)); //FIXME
 }
 
 
-void arp_print(struct arp *arp_r)
+static char *mac_ntoa(unsigned char *p)
 {
-    __u8 *addr;
+    static char b[18];
 
-    addr = arp_r->ll_eth_src;
-    printf("ll_eth_src: %2X.%2X.%2X.%2X.%2X.%2X ",addr[0],addr[1],addr[2],addr[3],addr[4],addr[5]);
-    addr = arp_r->ll_eth_dest;
-    printf("ll_eth_dest: %2X.%2X.%2X.%2X.%2X.%2X \n",addr[0],addr[1],addr[2],addr[3],addr[4],addr[5]);
+    sprintf(b, "%02x.%02x.%02x.%02x.%02x.%02x",p[0],p[1],p[2],p[3],p[4],p[5]);
+    return b;
+}
 
-    printf("Op: %2X \n",arp_r->op);
-
-    addr = (__u8 *)&arp_r->ip_src;
-    printf("ip_src : %d.%d.%d.%d ",addr[0],addr[1],addr[2],addr[3]);
-    addr = (__u8 *)&arp_r->ip_dest;
-    printf("ip_dest: %d.%d.%d.%d \n",addr[0],addr[1],addr[2],addr[3]);
-    addr = arp_r->eth_src;
-    printf("eth_src: %2X.%2X.%2X.%2X.%2X.%2X ",addr[0],addr[1],addr[2],addr[3],addr[4],addr[5]);
-    addr = arp_r->eth_dest;
-    printf("eth_dest: %2X.%2X.%2X.%2X.%2X.%2X \n",addr[0],addr[1],addr[2],addr[3],addr[4],addr[5]);
+static void arp_print(struct arp *arp)
+{
+#if DEBUG_ARP
+    printf("ARP: ");
+    printf("op:%d ", ntohs(arp->op));
+    printf("ll:%s->", mac_ntoa(arp->ll_eth_src));
+    printf("%s ", mac_ntoa(arp->ll_eth_dest));
+    printf("%s->", in_ntoa(arp->ip_src));
+    printf("%s ", in_ntoa(arp->ip_dest));
+    printf("eth:%s->", mac_ntoa(arp->eth_src));
+    printf("%s", mac_ntoa(arp->eth_dest));
+    printf("\n");
+#endif
 }
 
 int arp_init (void)
@@ -120,14 +119,14 @@ int arp_init (void)
 	return 0;
 }
 
-void arp_reply(char *packet,int size)
+void arp_reply(unsigned char *packet,int size)
 {
     struct arp_addr apair;
     struct arp *arp_r;
 
     arp_r = (struct arp *) packet;
 
-    /* arp_print(arp_r); */
+    debug_arp("arp: SEND reply to %s\n", in_ntoa(arp_r->ip_src));
 
     /* swap ip addresses and mac addresses */
     apair.daddr = arp_r->ip_src;
@@ -136,7 +135,7 @@ void arp_reply(char *packet,int size)
     memcpy(apair.eth_src, eth_local_addr, 6);
 
     /* build arp reply */
-    arp_r->op=0x200; /*response - big endian*/
+    arp_r->op = htons(ARP_REPLY);
     memcpy(arp_r->ll_eth_dest, apair.eth_dest, 6);
     memcpy(arp_r->ll_eth_src, apair.eth_src, 6);
 
@@ -145,72 +144,68 @@ void arp_reply(char *packet,int size)
     memcpy(arp_r->eth_dest, apair.eth_dest, 6);
     arp_r->ip_dest=apair.daddr;
 
+    arp_print(arp_r);
     deveth_send(packet, sizeof (struct arp));
 }
 
 int arp_request(ipaddr_t ipaddress)
 {
-    int i;
-    fd_set fdset;
-    struct arp *arp_r;
-    static char packet[sizeof(struct arp)];
-    arp_r = (struct arp *)packet;
+    static struct arp arpreq;
 
-    debug_arp("arp: send request\n");
+    debug_arp("arp: SEND request\n");
+
     /* build arp request */
-    for (i=0;i<6;i++) arp_r->ll_eth_dest[i]=0xFF; /*broadcast*/
-    memcpy(arp_r->ll_eth_src, eth_local_addr, 6);
-    /*specify below in big endian*/
-    arp_r->ll_type_len=0x0608;
-    arp_r->hard_type=0x0100;
-    arp_r->proto_type=0x0008;
-    arp_r->hard_len=6;
-    arp_r->proto_len=4;
-    arp_r->op=0x0100; /*request - big endian*/
-    memcpy(arp_r->eth_src, eth_local_addr, 6);
-    arp_r->ip_src=local_ip;
-    for (i=0;i<6;i++) arp_r->eth_dest[i]=0;
-    arp_r->ip_dest=ipaddress;
+    memset(arpreq.ll_eth_dest, 0xFF, 6);	/* broadcast*/
+	memcpy(arpreq.ll_eth_src, eth_local_addr, 6);
+    /*specify below in big endian*/ //FIXME
+	arpreq.ll_type_len = ETH_TYPE_ARP;
+    arpreq.hard_type=0x0100;
+    arpreq.proto_type = ETH_TYPE_IPV4;
+    arpreq.hard_len=6;
+    arpreq.proto_len=4;
+    arpreq.op = htons(ARP_REQUEST);
+    memcpy(arpreq.eth_src, eth_local_addr, 6);
+    arpreq.ip_src=local_ip;
+    memset(arpreq.eth_dest, 0, 6);
+    arpreq.ip_dest=ipaddress;
 
-    deveth_send(&packet, sizeof (struct arp));
+    arp_print(&arpreq);
+    deveth_send((unsigned char *)&arpreq, sizeof(arpreq));
 
-    /* wait for reply */
-selectagain:
+    /* timeout wait for reply */
     debug_arp("arp: wait reply\n");
-
+    struct timeval timeint;
+    fd_set fdset;
+    timeint.tv_sec  = 0;
+    timeint.tv_usec = 500000L;	/* 1/2 second*/
     FD_ZERO(&fdset);
     FD_SET(tcpdevfd, &fdset);
-    i = select(tcpdevfd + 1, &fdset, NULL, NULL, NULL);
-    if (i < 0) {
-	if (errno == EINTR) {
-		fprintf(stderr, "arp: select EINTR\n");
-		goto selectagain;
-	}
-	perror("select");
-	return -1;
+    int i = select(tcpdevfd + 1, &fdset, NULL, NULL, &timeint);
+    if (i >= 0) {
+	printf("arp: got reply on timed wait\n");
+	deveth_process();
+	return 0;	/* success*/
     }
-    debug_arp("arp: got reply\n");
-    deveth_process(); /*get reply*/
-
-    return 0;
+    return 1;		/* error*/
 }
 
 
 /* Process incoming ARP packets */
-
-void arp_proc (char * packet, int size)
+void arp_recvpacket(unsigned char *packet, int size)
 {
 	struct arp * arp_r;
 
-	debug_arp("arp: incoming packet\n");
 	arp_r = (struct arp *) packet;
-	switch (arp_r->op) {
+	arp_print(arp_r);
+	switch (ntohs(arp_r->op)) {
 	case ARP_REQUEST:
+		debug_arp("arp: incoming REQUEST\n");
 		arp_cache_add (arp_r->ip_src, arp_r->eth_src);
 		arp_reply (packet, size);
 		break;
 
 	case ARP_REPLY:
+		debug_arp("arp: incoming REPLY\n");
 		arp_cache_add (arp_r->ip_src, arp_r->eth_src);
 		break;
 	}
--- a/elkscmd/ktcp/arp.h
+++ b/elkscmd/ktcp/arp.h
@@ -32,8 +32,7 @@ int arp_init (void);
 void arp_cache_add (ipaddr_t ip_addr, eth_addr_t * eth_addr);
 int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr);
 
-void arp_proc (char * packet, int size);
+void arp_recvpacket (unsigned char * packet, int size);
 int arp_request(ipaddr_t ipaddress);
-
 
 #endif /* !ARP_H */
--- a/elkscmd/ktcp/arp.h
+++ b/elkscmd/ktcp/arp.h
@@ -35,4 +35,6 @@ int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr);
 void arp_recvpacket (unsigned char * packet, int size);
 int arp_request(ipaddr_t ipaddress);
 
+char *mac_ntoa(unsigned char *p);
+
 #endif /* !ARP_H */
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -9,6 +9,8 @@
 #define DEBUG_IP	0
 #define DEBUG_ARP	0
 
+#define DEBUG_CB	0		/* dump control blocks*/
+
 #if DEBUG_TCP
 #define debug_tcp	printf
 #else
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -5,10 +5,13 @@
 
 //#define ARP_WAIT_KLUGE	/* force loop on arp reply*/
 
+/* turn these on for ELKS debugging*/
 #define DEBUG_TCP	0
 #define DEBUG_IP	0
 #define DEBUG_ARP	0
 
+/* leave these off for now - too much info*/
+#define DEBUG_MEM	0		/* debug memory allocations*/
 #define DEBUG_CB	0		/* dump control blocks*/
 
 #if DEBUG_TCP
@@ -27,6 +30,12 @@
 #define debug_arp	printf
 #else
 #define debug_arp(...)
+#endif
+
+#if DEBUG_MEM
+#define debug_mem	printf
+#else
+#define debug_mem(...)
 #endif
 
 #endif
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -3,6 +3,8 @@
 
 #include <autoconf.h>
 
+//#define ARP_WAIT_KLUGE	/* force loop on arp reply*/
+
 #define DEBUG_TCP	0
 #define DEBUG_IP	0
 #define DEBUG_ARP	0
--- a/elkscmd/ktcp/deveth.c
+++ b/elkscmd/ktcp/deveth.c
@@ -33,7 +33,7 @@ static int devfd;
 static eth_addr_t broad_addr = {255, 255, 255, 255, 255, 255};
 
 
-void deveth_printhex(char* packet, int len)
+void deveth_printhex(unsigned char *packet, int len)
 {
   unsigned char *p;
   int i;
@@ -112,13 +112,13 @@ void deveth_process ()
 	  break;
 
   case ETH_TYPE_ARP:
-	  arp_proc (sbuf, len);
+	  arp_recvpacket (sbuf, len);
 	  break;
   }
 }
 
 
-void deveth_send(char *packet, int len)
+void deveth_send(unsigned char *packet, int len)
 {
     //printf("deveth_send:\n");
     //deveth_printhex(packet,len);
--- a/elkscmd/ktcp/deveth.c
+++ b/elkscmd/ktcp/deveth.c
@@ -26,6 +26,7 @@
 #include "arp.h"
 
 eth_addr_t eth_local_addr;
+int eth_device;
 
 static unsigned char sbuf [MAX_PACKET_ETH];
 static int devfd;
@@ -33,23 +34,20 @@ static int devfd;
 static eth_addr_t broad_addr = {255, 255, 255, 255, 255, 255};
 
 
+#if DEBUG
 void deveth_printhex(unsigned char *packet, int len)
 {
-  unsigned char *p;
-  int i;
+  unsigned char *p = packet;
+  int i = 0;
   printf("deveth_process():%d bytes\n",len);
   if (len > 128) len = 128;
-  p = packet;
-  i = 1;
   while (len--) {
-	printf("%02X",*p);
-	if ((i % 2) == 0) printf(" "); /*%=modulo*/
-	if ((i % 16) == 0) printf("\n"); /*%=modulo*/
-	p++;
-	i++;
+	printf("%02X ", *p++);
+	if ((i++ & 15) == 15) printf("\n");
   }
   printf("\n");
 }
+#endif
 
 int deveth_init(char *fdev)
 {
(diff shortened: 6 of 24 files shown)


ghaerr:
Additional changes:

Fix bugs in handling received RST packets, connection refused: should allow multiple telnet connections out after previous connection failure.

Don't retransmit ACK packets w/no data.

Increase window size from 256 to 2048 (buffers were already allocated).
  Ultimately window size will be a tradeoff with total memory used vs throughput

Found another nasty corruption bug: somehow a "struct net_device *dev" pointer was introduced into the code via a header file only, it was never initialized nor seen. Then when ethernet or slip was initialized by "dev->type = [1 or 0]", it scribbled into ktcp memory. Turns out the entire struct net_device was thrown in and never used.

Add mac_ntoa for arp and ethernet address display.

More code and warning cleanups.

ghaerr:
Added capability to use telnet/telnetd on local system without passing packets over ethernet or slip. This is a major step forward, as now ktcp can be tested completely on a single ELKS system.

There are still major issues, even though no packet loss, since no packets ever written to ethernet, Connecting to localhost on ktcp sometimes runs for a while, other times crashes quickly. There's a couple big bugs still hanging around. This capability should allow those problems to be found quickly, I hope. When not connecting to localhost, or when running serial debug, strangely enough, the system runs without errors...

Usage: `telnet 10.0.2.15'. (localhost or 127.0.0.1 not yet implemented). Still requires an installed ethernet card or slip, that will also be fixed shortly).

Other changes:
Add checksum to icmp echo reply (untested).
Add icmp destination unreachable message when received.
Cleanup kluge memmove on adding ethernet link layer
Cleanup IP layer send packet
Add debug output for sent TCP packets
Update retransmit debug output
Tested IP and TCP checksums - valid, never was a problem.
Reverted ntohs() macro

Bad memory problems still remain. Best to operate using DEBUG_TCP/IP/ARP in config.h, then run with serial console enabled.

Mellvik:
While the ARP problem - as described - has been fixed, it does have a cousin that I've been struggling with for a while. 

If an ARP request comes in while a ping is active, and timing will have it such that elks is responding to a PING before responding to the ARP (sequence -> incoming ICMP request -> Incoming ARP request -> outgoing PING reply -> outgoing ARP reply). The last deveth_send never returns, possibly memory corruption issue. If the ARP reply follows immediately after the ARP request, there is no problem. The hang is in ktcp, not the system.
Changing ping packet size and/or frequency changes the point at which this happens, but nothing else. 
Ideas solicited.

--Mellvik

ghaerr:
Hello @Mellvik,

>  sequence -> incoming ICMP request -> Incoming ARP request -> outgoing PING reply -> outgoing ARP reply

In the above sequence, from the point of view of the packet driver, ktcp is still working on a response to the ICMP request when the driver receives the ARP request. The driver signals an interrupt, which is to wake up ktcp, but ktcp is already awake and running, so the ARP request packet stays unread in the driver hardware. ktcp then sends the PING reply (with the unread ARP request still in the driver hardware), then reads the ARP request, then sends the ARP reply. 

Is this problem with the unmodified driver, or after the changes you've made for the receive processing? I don't yet understand how the card handles input buffering during processing of write request. Are the input and output buffers seperate? Perhaps the onboard queuing is getting confused when doing output and input still pending.

Mellvik:
Thanks @ghaerr,

I'm using the new driver which - apart from this oddity, now seems very stable. So stable it makes ktcp look very good. Anyway, staying on the problem at hand: I found a case where the exact same sequence works, hand have narrowed it down to timing. Which is why I haven't the the problem before. Removing one debug print after the other reveal new scenarios and (quite frequently) issues, as you've commented before.

IN this case, if the distance (in time) between an ICMP and an ARP request is less than 0.05 seconds (approx), the problem appears. So it seems I have a problem getting the driver to handle back to back packets. The NIC is good at this, the ring buffer stuff inside the NIC is quite sophisticated (I've put some details about that in my notes, coming along with the changes).

It's actually possible that I have an issue with interrupts akin to the ring buffer issues you found in the serial driver. I'll just continue the search (I will be off physical hardware for a few days starting tomorrow, so I figured I'll do QEMU testing with the new driver).

Now that the debug prints are coming out of the driver, I've seen long telnets in and out being impressively stable. Pings w/ large packets, pings w/ short intervals (0.1 secs) etc. working fine - until ARP screws it all up.

More later.

-M

ghaerr:
> IN this case, if the distance (in time) between an ICMP and an ARP request is less than 0.05 seconds (approx), the problem appears. So it seems I have a problem getting the driver to handle back to back packets. 

Yes, that sounds likely.

> It's actually possible that I have an issue with interrupts akin to the ring buffer issues you found in the serial driver.

Well, the serial driver interrupt routine is generated when a byte is ready and also reads the byte, resetting the UART receive available bit, while the NE2K driver interrupt routine only wakes up the ktcp process waiting on a read() and never talks with the network card. The ktcp process then runs, which calls the asm routine that actually reads the packet with `eth_read` (and presumably resets the driver packet available flag/mechanism). Looking at the driver, it seems that if two packets arrive before a process reads them, two interrupts occur, select() stays true and ktcp read is called twice to read each packet properly. This is controlled by the asm function ne2k_rx_stat().

The ability to write a packet is similarly controlled by ne2k_tx_stat(). It could be possible that if a packet is received, and then another received, that ne2k_tx_stat() still doesn't return true, which will hang ktcp in `eth_write`. Perhaps put some debug statements to show when ktcp calls wait_event_interruptable() and returns from it.



> Now that the debug prints are coming out of the driver, I've seen long telnets in and out being impressively stable. Pings w/ large packets, pings w/ short intervals (0.1 secs) etc. working fine

Nice! Agreed that ktcp seems stable (with serial and ethernet, except for your problems) for long periods, although I can duplicate outgoing telnet on QEMU viciously dying with a kernel read() returning more data than it asked for, and there are numerous problems running any ktcp activity after a connection is closed or dropped. I plan on fixing up the session close/RST handling after getting the memory corruption bug(s) eliminated.

Mellvik:
﻿Thanks @ghaerr,
it’s always useful to get another perspective in. So easy to get stuck on a specific angle when banging the head against the wall. While I’m now on friendly terms with just about every single line in the ne2k-driver - and have a reasonable understanding of ktcp, the layers between (select, wait, wakeup …) etc. are unclear to say the least.

Anyway, your input eliminated some of the alternatives I was looking at and sent me back to arp. ARP is always involved when there is a prblem, and ping flooding doesn’t cause any other problems than lost packets, so there is no back-to-back problem in the driver.

Actually good old ARP  may be THE problem, and at least part of it. Basically, the way things are presently set up, when elks sends an ARP request, it assumes the next incoming packet is the ARP reply. When it isn't we get an interesting scenario with nested calls, more arp requests, more nested calls etc.
Like this [not a complete call trace, just the 'important' ones]
-> fresh boot, no traffic yet
-> ip_recvpacket
-> icmp_process
-> ip_sendpacket (preparing echo reply)
-> arp_request (don't have the address)
-> deveth_send (ARP request gets on the wire)
-> select_wait
-> got ‘reply’ - but it’s another ICMP echo request, not an ARP reply 
-> deveth_process
-> ip_recvpacket (reentered)
-> icmp_process
-> ip_sendpacket (reentered)
-> new arp_request
This loop continues for a while - until ELKS finally gets one of the many (probably the first) ARP replies, updates the arp cache and kicks off 15 ICMP replies - all with (the same) erroneous content, overwritten somewhere in the process. [Possibly a separate bug in itself]

Actually very much like the previous ARP problem we fixed by giving ARP a separate send buffer to play with. The out-of-sequence ARP-handling in elks is biting again, and I’m tempted to fix it by disconnecting the driver from the protocol handling, and adding a ring buffer for received packets (and outgoing packets, but that can wait). Then an ARP packet is just another packet - part of the stream, no special arrangements. Otherwise we would seemingly require more special arrangements such as flags indicating that an ARP request is outstanding for this particular address, just ignore input packets till the reply comes - with timeouts, repeats etc.

I’m not saying this is the only problem we’ve left. For example it does not explain the following:
If ELKS is 'semi-flooded' with ARP requests for other addresses (the test is 3 pings in parallel to non existent addresses, generating continuous 3 arp request per second on the wire), ktcp hangs after say 15-20 seconds. This may be related to the timing of the requests, two requests coming very close in time, I haven't looked at the timing of this scenario yet. The key observation being that there is no send-packet involved, reducing the scope of the search to receive only.

Anyway, arp is involved in every case, so one test I may make before I do anything else (when I’m back to the hardware late this week), is to add a static entry to the arp cache for testing. Traffic is still not ARP free, Raspbian is sending ARP renewals quite frequently, but these seem to be handles ok by ELKS..
 
If there are more or less obvious faults in my logic, I'd appreciate feedback. Maybe we'll just redo the arp handling for now, not waiting for an arp-reply but processing it whenever it comes, and having a status in the arp-table indicating arp_pending ...

 --M

ghaerr:
Hello @Mellvik!

> -> fresh boot, no traffic yet
-> ip_recvpacket
-> icmp_process
-> ip_sendpacket (preparing echo reply)
-> arp_request (don't have the address)
-> deveth_send (ARP request gets on the wire)
-> select_wait
-> got ‘reply’ - but it’s another ICMP echo request, not an ARP reply
-> deveth_process
-> ip_recvpacket (reentered)
-> icmp_process
-> ip_sendpacket (reentered)
-> new arp_request
This loop continues for a while - until ELKS finally gets one of the many (probably the first) ARP replies, updates the arp cache and kicks off 15 ICMP replies

Great job on debugging ktcp!!! Just looking at your trace shows definite big problems with the way we're handling ARP, and that my latest "fix" is just as much a band-aid as when the kluge was originally written.


 > ... kicks off 15 ICMP replies all with (the same) erroneous content, overwritten somewhere in the process. [Possibly a separate bug in itself]

Most definitely a separate bug. Ktcp uses two static buffers for TCP and IP packet sending and two buffers (ARP and TCP/IP) for packet receiving. Whenever an IP or TCP output routine is reentered, its buggy as the same output buffer is used for the first packet. ICMP is a little bit different, and uses one of the two buffers we setup for IP(ICMP)/ARP input processing.


> Basically, the way things are presently set up, when ELKS sends an ARP request, it assumes the next incoming packet is the ARP reply. 

Actually, not quite - my attempt at fixing the ARP wait kluge introduced separate buffers for handling an IP packet separately from ARP, and in the case of a received ICMP IP reply packet, things work. 

> When it isn't we get an interesting scenario with nested calls, more arp requests, more nested calls etc.

Yep. I hadn't considered the problem of an IP (or ICMP) packet that generates an TCP or IP response while waiting for the ARP reply, which is happening in your scenario. Just goes to show that quick fixes to kluges is not the answer...



> Actually very much like the previous ARP problem we fixed by giving ARP a separate send buffer to play with. The out-of-sequence ARP-handling in elks is biting again

Yep.

> I’m tempted to fix it by disconnecting the driver from the protocol handling and adding a ring buffer for received packets (and outgoing packets, but that can wait). 

A received packet queue could possibly work. I had considered that a real fix would require buffering outgoing packets. If we buffered input packets, then they could just be queued and ignored while waiting for a received ARP packet (or timeout). It also could be done without rewriting all the code dealing with static buffers for TCP and IP outgoing packets. If, instead, outgoing packets were queued, we introduce the problem of having to check them on TCP timeout retransmissions and other possible syncing issues. The problem is ktcp could run out of memory queuing incoming packets waiting for an ARP reply, along with not processing input when received, which isn't a great idea.

The flip side is implementing an outgoing packet queue - this would allow for immediate processing of any received ARP/IP/TCP packet, all of which would immediately be sent, except for those waiting for an ethernet address, which would be queued waiting for an ARP reply, or timeout and dropped.

I need to think a bit more whether it would be better to just jump in and implement output queueing from the start, or whether input queuing would fully solve our problem or not. I don't think we need both.


> Then an ARP packet is just another packet - part of the stream, no special arrangements. Otherwise we would seemingly require more special arrangements such as flags indicating that an ARP request is outstanding for this particular address, just ignore input packets till the reply comes - with timeouts, repeats etc.

Yes. Perhaps a simple implementation where, since outgoing IP would still not be reentrant, only a single IP datagram would be allowed to wait in `ip_sendpacket`. If an ARP reply comes in, it would get sent to the ethernet driver, otherwise deleted on timeout. Any other received packet would be queued. This isn't fully asynchronous, building a full async solution will require dynamically allocating all packets in and out, which would be nice to do after we've solved the still-remaining memory corruption bugs in ktcp.

Thinking about this a bit more, if outgoing IP is not reentrant, then incoming ICMP would also have to wait before any processing, including from hosts with a known ARP/MAC address. It seems implementing outgoing queuing may be a better idea, as the input queue could get large and all other remote hosts/systems would have to wait, although an ARP reply shouldn't take too long (unless dropped or lost).

> If ELKS is 'semi-flooded' with ARP requests for other addresses (the test is 3 pings in parallel to non existent addresses, generating continuous 3 arp request per second on the wire), ktcp hangs after say 15-20 seconds. 

Not sure what this is... remember there are only five entries max in the ARP table. Can you generate ARP requests only separately without any other IP/ICMP/TCP? That might help locate this problem. It would be nice to understand whether this is yet another big bug, or is instead related to the still-unsolved memory corruption.

> The key observation being that there is no send-packet involved, reducing the scope of the search to receive only.

Agreed.

> Anyway, arp is involved in every case, so one test I may make before I do anything else (when I’m back to the hardware late this week), is to add a static entry to the arp cache for testing. Traffic is still not ARP free, Raspbian is sending ARP renewals quite frequently, but these seem to be handles ok by ELKS..

The ability to have a static ARP entry sounds like a great idea for testing. This could enable us to figure out whether ktcp in fact works well when not having to wait for an ARP response, which might help in determining whether to implement input or output queueing.


>  Maybe we'll just redo the arp handling for now, not waiting for an arp-reply but processing it whenever it comes, and having a status in the arp-table indicating arp_pending ..

Not waiting for an ARP reply means that we'd have to drop sending the packet waiting for ARP unless we queue it, which I don't think will work well. We need to think hard about whether input queuing will work, otherwise go straight towards output packet queuing when no ARP entry present.

I can write the input or output queueing, and can do initial tests on QEMU then hand it over to you for heavier use and testing. I haven't yet got my real hardware setup for ethernet but should be able to do so soon.

Mellvik:
Thanks for the feedback, @ghaerr, 
this does indeed require some careful thinking. Like you pointed out a while back - activities away from screens and keyboards helps - like a week in the mountains...


> If ELKS is 'semi-flooded' with ARP requests for other addresses (the test is 3 pings in parallel to non existent addresses, generating continuous 3 arp request per second on the wire), ktcp hangs after say 15-20 seconds.
> 
> Not sure what this is... remember there are only five entries max in the ARP table. Can you generate ARP requests only separately without any other IP/ICMP/TCP? That might help locate this problem. It would be nice to understand whether this is yet another big bug, or is instead related to the still-unsolved memory corruption.
> 

Yes, that’s the scenario. Nothing but ARPs on the wire, requests for non existent addresses. Turns out (I’ve done some more testing remotely) this is the ‘packets too close in time’ problem we’ve seen before. Having it happen with only ARP requests on the wire, moves us closer to a solution. Seems like another case of reentry, overwriting values etc - quite possibly in the driver, which either causes a tight loop (some status test) or interrupts not being reset and thus remaining blocked. I’ll dig to the bottom of this one when I’m back.

> A received packet queue could possibly work. I had considered that a real fix would require buffering outgoing packets. If we buffered input packets, then they could just be queued and ignored while waiting for a received ARP packet (or timeout). It also could be done without rewriting all the code dealing with static buffers for TCP and IP outgoing packets. If, instead, outgoing packets were queued, we introduce the problem of having to check them on TCP timeout retransmissions and other possible syncing issues. The problem is ktcp could run out of memory queuing incoming packets waiting for an ARP reply, along with not processing input when received, which isn't a great idea.
> 
> The flip side is implementing an outgoing packet queue - this would allow for immediate processing of any received ARP/IP/TCP packet, all of which would immediately be sent, except for those waiting for an ethernet address, which would be queued waiting for an ARP reply, or timeout and dropped.
> 
> I need to think a bit more whether it would be better to just jump in and implement output queueing from the start, or whether input queuing would fully solve our problem or not. I don't think we need both.
> 
> Then an ARP packet is just another packet - part of the stream, no special arrangements. Otherwise we would seemingly require more special arrangements such as flags indicating that an ARP request is outstanding for this particular address, just ignore input packets till the reply comes - with timeouts, repeats etc.
> 
> Yes. Perhaps a simple implementation where, since outgoing IP would still not be reentrant, only a single IP datagram would be allowed to wait in ip_sendpacket. If an ARP reply comes in, it would get sent to the ethernet driver, otherwise deleted on timeout. Any other received packet would be queued. This isn't fully asynchronous, building a full async solution will require dynamically allocating all packets in and out, which would be nice to do after we've solved the still-remaining memory corruption bugs in ktcp.
> 
> Thinking about this a bit more, if outgoing IP is not reentrant, then incoming ICMP would also have to wait before any processing, including from hosts with a known ARP/MAC address. It seems implementing outgoing queuing may be a better idea, as the input queue could get large and all other remote hosts/systems would have to wait, although an ARP reply shouldn't take too long (unless dropped or lost).
> 
Very useful discussion. You’re seeing buffers at a higher level and with more ‘intelligence’ than I have envisioned. My idea  has been just a simple queue between the IP layer and the driver, no timeouts or any intelligence. For outbound traffic such a queue will make sending fully asynchronous and hopefully simplify the sending part IP-level and up.

However, this will not alleviate the 'outgoing packets awaiting ARP resolution’ problem. Then again I’m leaning towards just discarding those packets, assisted by a status field in the arp cache structure: If arp_cache_get returns e.g. ARP_PENDING, then discard - instead of triggering another ARP request. And probably a counter taking care of the case of a lost ARP request.

Further, I think it may be a good idea - at least during the debug phase - to pick up the incoming mac address off of ICMP packets, and add it to the ARP cache (if it’s local). More flexible than my previous proposal to add a static arp entry for testing. Such ARP cache entries may even carry their own cache-flag making them secondary citizens until updated by a proper ARP reply.

It seems to me that discarding outgoing packets destined to addresses with a pending ARP request is a small price to pay given the simplicity it provides. TCP (and eventually UDP) level traffic will hardly be affected at all (unless the other end has very short retransmission timeouts). And it reduces the need to get an output buffering mechanism going right away.

I will be looking more into this on friday.

—Mellvik

ghaerr:
> For outbound traffic such a queue will make sending fully asynchronous and hopefully simplify the sending part IP-level and up.
> However, this will not alleviate the 'outgoing packets awaiting ARP resolution’ problem. Then again I’m leaning towards just discarding those packets

The problem with just discarding outgoing packets that haven't been ARPed (i.e. have an unknown ethernet address) is that this method would always drop the first IP/IPCMP packet to an unknown host, which would always happen for the first ping or connection attempt.

> assisted by a status field in the arp cache structure: If arp_cache_get returns e.g. ARP_PENDING, then discard - instead of triggering another ARP request. And probably a counter taking care of the case of a lost ARP request.

No, I don't think we need have state in the arp_cache, nor keep a counter on "lost" ARP requests. If an ARP response is never received, the outgoing packet being held will simply be dropped.

I think this whole ARP wait kluge/reentrant mess can be solved simply and permanently with the following: 1) move all ARP and ethernet header handling out of the IP layer entirely (that is, out of ip_sendpacket). 2) change the ktcp ethernet handling in deveth.c to handle all ARP requests and replies (it already handles replies). 3) add a link-layer only outgoing packet queue to hold just those ethernet packets that don't have an ethernet address (no ARP reply).

Here's how it would work: neither TCP nor IP would care about any packet queuing, and ip_sendpacket would pass the packet to the appropriate link-layer (SLIP or ethernet) for handling. For ethernet, the arp-cache would be checked for an ethernet address for the IP packet; if so, prepare the ethernet header and send it. If not, copy and queue the IP (only) packet and issue an ARP request, and return. When the ethernet layer receives an ARP response, update the arp-cache as it does now, and check the outgoing ethernet queue for any IP packets waiting for a that address, and remove and send them. 

Note, the upper level TCP or IP thinks the packet is sent for all it knows, and the packet is only queued in the link layer, and will be removed by the link layer after some timeout. This timeout needs to be less than the TCP retransmit timeout. All queuing would be entirely encapsulated in the ethernet link layer, unlike it is now. If an IP/ICMP packet needs queuing, additional memory would be allocated so that any other received ethernet IP packets could "reenter" (call again, not actually reenter) ip_sendpacket without consequence. 

> Further, I think it may be a good idea - at least during the debug phase - to pick up the incoming mac address off of ICMP packets, and add it to the ARP cache (if it’s local). 

You're saying that the ethernet link layer could stuff the arp-cache if empty with the incoming remote ethernet address, and thus skip the ARP request it would otherwise make? That could be a good idea, we might want to look at the ARP RFC to see what they say about that. If you're going to "steal" ICMP ethernet remote addresses, why not steal the remote ethernet address from received IP(TCP) packets as well? 

> More flexible than my previous proposal to add a static arp entry for testing. Such ARP cache entries may even carry their own cache-flag making them secondary citizens until updated by a proper ARP reply.

I don't know about adding cache-flags and secondary status... I would say either use the address or not. BTW, ktcp never updates its ARP cache after an initial entry nor times any entry out, which is also contrary to RFC.

In summary: I think I can write a simple queuing capability to be fully encapsulated in deveth.c, thus removing any ARP waits and ethernet handling at the IP layer, which solves our IP reentry problem. In addition, ktcp would handle any other incoming IP/ICMP/TCP without delay while waiting for an ARP response. Of course, if these additional incoming packets are all from the same unknown host, then the link layer packet queue will grow, but this could be managed with appropriate timeouts (which should be just a bit more than the average ARP reply perhaps from the RFC).

Mellvik:
> For outbound traffic such a queue will make sending fully asynchronous and hopefully simplify the sending part IP-level and up.
> However, this will not alleviate the 'outgoing packets awaiting ARP resolution’ problem. Then again I’m leaning towards just discarding those packets
> 
> The problem with just discarding outgoing packets that haven't been ARPed (i.e. have an unknown ethernet address) is that this method would always drop the first IP/IPCMP packet to an unknown host, which would always happen for the first ping or connection attempt.
> 
Actually, that's not what I'm suggesting. I mean 'awaiting ARP resolution' literally: If there already is a packet waiting for ARP resolution to a particular address, more packets waiting for the same doesn't make much sense and is unlikely to happen as part of regular traffic. It does require state in the arp cache and it is protection against overrun/overload.
> I think this whole ARP wait kluge/reentrant mess can be solved simply and permanently with the following: 1) move all ARP and ethernet header handling out of the IP layer entirely (that is, out of ip_sendpacket). 2) change the ktcp ethernet handling in deveth.c to handle all ARP requests and replies (it already handles replies). 3) add a link-layer only outgoing packet queue to hold just those ethernet packets that don't have an ethernet address (no ARP reply).
> 
> Here's how it would work: neither TCP nor IP would care about any packet queuing, and ip_sendpacket would pass the packet to the appropriate link-layer (SLIP or ethernet) for handling. For ethernet, the arp-cache would be checked for an ethernet address for the IP packet; if so, prepare the ethernet header and send it. If not, copy and queue the IP (only) packet and issue an ARP request, and return. When the ethernet layer receives an ARP response, update the arp-cache as it does now, and check the outgoing ethernet queue for any IP packets waiting for a that address, and remove and send them.
> 
> Note, the upper level TCP or IP thinks the packet is sent for all it knows, and the packet is only queued in the link layer, and will be removed by the link layer after some timeout. This timeout needs to be less than the TCP retransmit timeout. All queuing would be entirely encapsulated in the ethernet link layer, unlike it is now. If an IP/ICMP packet needs queuing, additional memory would be allocated so that any other received ethernet IP packets could "reenter" (call again, not actually reenter) ip_sendpacket without consequence.
> 
Yes, this sounds good! Sounds like an approach that will save the day permanently.
> Further, I think it may be a good idea - at least during the debug phase - to pick up the incoming mac address off of ICMP packets, and add it to the ARP cache (if it’s local).
> 
> You're saying that the ethernet link layer could stuff the arp-cache if empty with the incoming remote ethernet address, and thus skip the ARP request it would otherwise make? That could be a good idea, we might want to look at the ARP RFC to see what they say about that. If you're going to "steal" ICMP ethernet remote addresses, why not steal the remote ethernet address from received IP(TCP) packets as well?
> 
I found nothing about this in the RFC, and I suppose there must be good reasons for this not being the regular case. One that comes to mind is security. Not that ARP is all that secure, but having a bogus ICMP (bogus sender address) pollute the ARP cache may have some interesting consequences :-). Anyway, I'll implement this for debugging while we think about it. My preliminary setup has a status word in the arp cache structure flagging the address as 'temporary', which is reset by a regular ARP reply confirming the entry.
> More flexible than my previous proposal to add a static arp entry for testing. Such ARP cache entries may even carry their own cache-flag making them secondary citizens until updated by a proper ARP reply.
> 
> I don't know about adding cache-flags and secondary status... I would say either use the address or not. BTW, ktcp never updates its ARP cache after an initial entry nor times any entry out, which is also contrary to RFC.
> 
I realize that, and I'm thinking this has to be fixed at some point, time at which the status word will be rechristened to something else and 'temporary status' replaced with a short timeout or just gone.
> In summary: I think I can write a simple queuing capability to be fully encapsulated in deveth.c, thus removing any ARP waits and ethernet handling at the IP layer, which solves our IP reentry problem. In addition, ktcp would handle any other incoming IP/ICMP/TCP without delay while waiting for an ARP response. Of course, if these additional incoming packets are all from the same unknown host, then the link layer packet queue will grow, but this could be managed with appropriate timeouts (which should be just a bit more than the average ARP reply perhaps from the RFC).
> 
> 
I'd say go a head with this significant improvement. It will simplify and add robustness to the implementation. I realize I'm running out of time getting a PR out with the new driver and associated changes ...

Now - here's a case for a similar but much simpler receive buffer: In high load situations (which I'm creating plenty of for testing), we sometimes get NIC buffer overruns, which now get handled via an interrupt. Part of the recovery routine is to relieve the before of packet thereby making more space available. Which means that the recovery routine actually reads one or more packets, but har nowhere to put them. In the version now being prepared for a PR, I'm just stuffing a packet into recv_buf in eth-main.c which affectively drops the packet, possibly overwrites the content if this happens in the middle of something else. Alternatively I could temporary allocate local buffer space on the stack, then discard, but optimally the data should be read and passed on. Opinions?

—Mellvik

ghaerr:
> Actually, that's not what I'm suggesting. I mean 'awaiting ARP resolution' literally: If there already is a packet waiting for ARP resolution to a particular address, more packets waiting for the same doesn't make much sense and is unlikely to happen as part of regular traffic. It does require state in the arp cache and it is protection against overrun/overload.

I see. This idea could help when being blasted by a fast ping while waiting for an ARP reply. The state in the ARP cache would be whether or not there is a single packet being "held" waiting for an ARP response. If you think it worthwhile, I could implement only a single packet hold rather than a list and see how that fares, before implementing a queue. I agree it seems unlikely to have multiple non-ping packets being held in this case for ELKS being essentially a single-user system.

> I found nothing about this in the RFC, and I suppose there must be good reasons for this not being the regular case. One that comes to mind is security. Not that ARP is all that secure, but having a bogus ICMP (bogus sender address) pollute the ARP cache may have some interesting consequences :-). Anyway, I'll implement this for debugging while we think about it. My preliminary setup has a status word in the arp cache structure flagging the address as 'temporary', which is reset by a regular ARP reply confirming the entry.

Your security scenario is a good reason not to implement it. We should save your code nonetheless so it can also be tested along with the link layer (single) queuing option to see how it fares when also added. In general though, it seems this would only help when a very fast sequence of received packets arrives from a new host requiring an ARP request that gets lots somehow.


> I'd say go a head with this significant improvement. It will simplify and add robustness to the implementation. I realize I'm running out of time getting a PR out with the new driver and associated changes ...

No rush on getting your PR out, I suggest you split your work out into a driver change and ktcp PRs. This would allow using your improved driver with the ktcp link layer architecture change should I get that done first. I will still have to test the link layer change using QEMU for the time being.


> Now - here's a case for a similar but much simpler receive buffer: In high load situations (which I'm creating plenty of for testing), we sometimes get NIC buffer overruns, which now get handled via an interrupt. Part of the recovery routine is to relieve the before of packet thereby making more space available. 

It would seem the best thing to do in a buffer overrun interrupt routine would be to clear the overrun error and possibly drop the last (not first) packet in the ring buffer. Explained more below. Its just like the UART ring buffer - if the system cannot keep up with received data, the latest received data is summarily dropped. I know you disagree with this, but, for various longer reasons, the kernel can't force data into a user process that's not ready for it, and has place to put it. That's what the ring buffers are for in the first place - temp holding. In the serial I/O case, there's sometimes a hw FIFO and then a kernel ring buffer; in the ethernet case, there's a hw ring buffer.


> Which means that the recovery routine actually reads one or more packets, but har nowhere to put them. In the version now being prepared for a PR, I'm just stuffing a packet into recv_buf in eth-main.c which affectively drops the packet, 

Is it possible to leave the first packet in the ethernet buffer and drop the latest packet? That would be better?


> possibly overwrites the content if this happens in the middle of something else.

Your idea of stuffing the latest packet into recv_buf would be good except that you're in an interrupt handler, which 1) could be interrupting a user process regular read() from the ethernet card at the same time, so that won't work; and 2) its a bad idea to copy a buffer in an interrupt routine, as this takes a long time and no other interrupts get processed, causing other problems (like dropped serial characters, for instance).

To fix problem 1) above, you'd have to protect both the user read() routine by turning off interrupts, and have the buffer overrun routine run with interrupts off, with potential dire consequences for any other time-sensitive kernel processing duties. We're far better off to just drop the latest received (partial) packet (if possible), since ELKS isn't keeping up with the received data anyway. Any fully-intact packets could be left in the hw queue, but there always needs to be ring buffer space for receiving one more packet without overwriting the first received packet that hasn't been read yet. Lowering MTU or outstanding TCP window size would likely be better options for flow control.


> Alternatively I could temporary allocate local buffer space on the stack, then discard,

You definitely can't allocate much temp data space on the stack in an interrupt routine, and very little in normal kernel processing. If you look at arch/i86/kernel/irqtab.S::_irqit(), you'll see that the ELKS assigns all interrupt routines a max 512-byte stack (for ALL overlapped interrupts). The "normal" kernel stack for non-interrupt use is defined in linuxmt/include/sched.h::KSTACK_BYTES as 988 bytes for the kernel stack and is held at the top of the task structure.

> but optimally the data should be read and passed on. Opinions?

IMO, the first received packets should be kept until the hw ring buffer is full, then subsequent packets dropped. No special handling the hw buffer overflow other than that, and let the normal read() processing work through its interrupt routine and normal process read() handling. There always needs to be space for in the hw queue so as to not overwrite a received packet that has already caused an interrupt.

Mellvik:
OK; we're getting there. Re. buffer overruns, here's some  important background:

The mechanism in the NIC accommodates for reading packets from the (ring) buffer in the sequence they entered. In the case of an overrun, the packet in which the overrun occurred, is discarded - being incomplete anyway, while the rest is intact. Getting the last packet entered before the error is not possible unless we empty the entire buffer. 

Given your very valid discussion about the alternatives (all bad), the quick fix may be to malloc (can we do that here?) a packet buffer for these (odd) situations and set a flag so that this becomes the next read, and consider it a temp fix until a real receive buffer is in place? If there are more overruns in a row (buffer busy) just discard (oddly enough we need a buffer for that too …).

> 
> Actually, that's not what I'm suggesting. I mean 'awaiting ARP resolution' literally: If there already is a packet waiting for ARP resolution to a particular address, more packets waiting for the same doesn't make much sense and is unlikely to happen as part of regular traffic. It does require state in the arp cache and it is protection against overrun/overload.
> 
> I see. This idea could help when being blasted by a fast ping while waiting for an ARP reply. The state in the ARP cache would be whether or not there is a single packet being "held" waiting for an ARP response. If you think it worthwhile, I could implement only a single packet hold rather than a list and see how that fares, before implementing a queue. I agree it seems unlikely to have multiple non-ping packets being held in this case for ELKS being essentially a single-user system.
> 
For all reasonable situations that would do it. The unreasonable situation (banging the system from several sources) should then cause  discards, not even cause an ARP request until the first is OK. Robustness above 'by the book'.
> I found nothing about this in the RFC, and I suppose there must be good reasons for this not being the regular case. One that comes to mind is security. Not that ARP is all that secure, but having a bogus ICMP (bogus sender address) pollute the ARP cache may have some interesting consequences :-). Anyway, I'll implement this for debugging while we think about it. My preliminary setup has a status word in the arp cache structure flagging the address as 'temporary', which is reset by a regular ARP reply confirming the entry.
> 
> Your security scenario is a good reason not to implement it. We should save your code nonetheless so it can also be tested along with the link layer (single) queuing option to see how it fares when also added. In general though, it seems this would only help when a very fast sequence of received packets arrives from a new host requiring an ARP request that gets lots somehow.
> 
Ok, I'll keep it in a ICMP_ARP_KLUDGE ifdef.
> I'd say go a head with this significant improvement. It will simplify and add robustness to the implementation. I realize I'm running out of time getting a PR out with the new driver and associated changes ...
> 
> No rush on getting your PR out, I suggest you split your work out into a driver change and ktcp PRs. This would allow using your improved driver with the ktcp link layer architecture change should I get that done first. I will still have to test the link layer change using QEMU for the time being.
> 
Good point, I'll do that. There are a few dependencies 'upwards', they will be easy to handle.
> …..
> 
> You definitely can't allocate much temp data space on the stack in an interrupt routine, and very little in normal kernel processing. If you look at arch/i86/kernel/irqtab.S::_irqit(), you'll see that the ELKS assigns all interrupt routines a max 512-byte stack (for ALL overlapped interrupts). The "normal" kernel stack for non-interrupt use is defined in linuxmt/include/sched.h::KSTACK_BYTES as 988 bytes for the kernel stack and is held at the top of the task structure.
> 
Thanks, that was useful indeed.
> 


—mellvik

ghaerr:
> The mechanism in the NIC accommodates for reading packets from the (ring) buffer in the sequence they entered.

Ok, FIFO, that's what I was suggesting. We always want to read the packets directly from the driver, using the non-interrupt read() routine that already exists, until no more packets are available. Any overrun packets should just be dropped and not handled specially.


> In the case of an overrun, the packet in which the overrun occurred, is discarded - being incomplete anyway, while the rest is intact. 

Great, so nothing actually has to be done in the case of an overrun packet, really.

> Getting the last packet entered before the error is not possible unless we empty the entire buffer.

That's ok, since we want the first packet received, not the last. Sorry for any confusion. 


> Given your very valid discussion about the alternatives (all bad), the quick fix may be to malloc (can we do that here?) a packet buffer for these (odd) situations and set a flag so that this becomes the next read,

No, its not a good idea to alloc kernel near memory for any kind of receive buffers, for a number of reasons: 1) it could run out, 2) all the other memory manipulation routines would have to be critical section protected with interrupts off, causing other issues previously discussed, and 3) interrupt routines need to be quick.


>  and consider it a temp fix until a real receive buffer is in place? If there are more overruns in a row (buffer busy) just discard (oddly enough we need a buffer for that too …).

Perhaps my last response was confusing - I mean to say that we only really want the *first* packet received, etc, until the hw buffer overflows, in which case we simply drop the packets. If the driver already does this, then no action needed. If the driver doesn't fully handle the buffer overrun with regards to resetting any device bits, then that needs to be added. But in general, no special handling of "later" packets by buffering "previous" packets should be done by the driver - the layers above the link layer handle lost packets using the TCP/IP protocol.

I've been on vacation in beautiful North Idaho but am returning today and I'll try to get on the packet queuing mod ASAP, for your integration into your debugged ethernet driver and possible other ktcp fixes. I'll just implement a single-packet delayed queue (per ARP entry) thanks to your suggestions.

Mellvik:
@ghaerr,
It's actually quite amusing. Part of what I do for a living is teaching tech people about the importance of clarity of communication - agreedupon definitions of words and terms Etc etc, and yet here we are.
So you misunderstood me and I misunderstood you, and we're back to square one. Just as well, the discussion was still fruitful, I'm coming back to that.

> Perhaps my last response was confusing - I mean to say that we only really want the first packet received, etc, until the hw buffer overflows, in which case we simply drop the packets. If the driver already does this, then no action needed. If the driver doesn't fully handle the buffer overrun with regards to resetting any device bits, then that needs to be added. But in general, no special handling of "later" packets by buffering "previous" packets should be done by the driver - the layers above the link layer handle lost packets using the TCP/IP protocol.
> 
The Ne2k (actually the 8390 chip) is quite sophisticated for its age. It can pass along everything on the wire, or filter any or all kinds of errors with high resolution on the filtering (such as accept crc errors, discard runts etc.). Thus many of the sanity checks in the old driver (and still present in the rewritten version) are actually superflous unless someone change the setup/initalization. An extra header (4 bytes) in front of every received packet as transferred to the driver contains a status byte indicating / among other things / whether the address was unicast or multi/broadcast, which errors - if allowed etc. Unused at this stage, but could save some processing if we wanted to.
Now, here's the point: The ring buffer handling is quite sophisticated, and among other things discards packets automatically if there is no space. But - and now we're back to the original discussion -  in order to reset the overflow condition, one or more packets must be removed from the ring in order to make space for more. So a packet transfer is part of the overflow reset procedure.
After our discussion I came to the conclusion that the best way to do this is to merge the overrun handling with a regular read. The low level read can easily check for an overrun condition before anything else and do whatever recovery necessary 'in-band' instead of the 'out-of-band' scenario I proposed initially. I'll make that change and test this weekend.


The ring buffer mechanism used in the NIC is inspirationally simple and smart. The (configurable) 14.5k receive buffer can hold up to 54 packets with no overhead.  I guess that's subject for another day...
>> 
>> I've been on vacation in beautiful North Idaho but am returning today and I'll try to get on the packet queuing mod ASAP, for your integration into your debugged ethernet driver and possible other ktcp fixes. I'll just implement a single-packet delayed queue (per ARP entry) thanks to your suggestions.
>> 
Way to go (vacation). Clears the mind. No rush.

//Mellvik

ghaerr:
@Mellvik, thanks for the educational description on the capabilities of the NE2K card and 8390 chip.

> The Ne2k (actually the 8390 chip) is quite sophisticated for its age. It can pass along everything on the wire, or filter any or all kinds of errors with high resolution on the filtering (such as accept crc errors, discard runts etc.). 

You may have seen that ktcp uses two (slow) memcmp's to filter out all packets other than broadcast and those directed to its MAC address. Ideally, this would be done during driver init, since ktcp isn't doesn't need promiscuous mode.


> Thus many of the sanity checks in the old driver (and still present in the rewritten version) are actually superflous unless someone change the setup/initalization. 

Like what, can you elaborate?

> An extra header (4 bytes) in front of every received packet as transferred to the driver contains a status byte indicating / among other things / whether the address was unicast or multi/broadcast, which errors - if allowed etc. Unused at this stage, but could save some processing if we wanted to.

Is this one of the options that can be turned on? I'm not sure we need this particular info, other than discarding the packet if there's an error. Ktcp doesn't use unicast/multicast, etc.


> After our discussion I came to the conclusion that the best way to do this is to merge the overrun handling with a regular read. The low level read can easily check for an overrun condition before anything else and do whatever recovery necessary 'in-band' instead of the 'out-of-band' scenario I proposed initially. I'll make that change and test this weekend.

That idea occurred to me, but I didn't know whether the interrupts could be merged. Sounds like doing both in the read routine solves the issues with the card ring buffer and is the right thing to do.


> The ring buffer mechanism used in the NIC is inspirationally simple and smart. The (configurable) 14.5k receive buffer can hold up to 54 packets with no overhead.  I guess that's subject for another day...

So, your testing on ring buffer overflow requires > 54 packets (ping?) in order to overflow? Is it set up for max packets now? I know very little about what settings are being used in the current or revised driver.

Thank you!

Mellvik:
@Ghaerr,
Many interesting details, maybe the subject for the upcoming wiki, thanks for the initiative.
> The Ne2k (actually the 8390 chip) is quite sophisticated for its age. It can pass along everything on the wire, or filter any or all kinds of errors with high resolution on the filtering (such as accept crc errors, discard runts etc.).
> 
> You may have seen that ktcp uses two (slow) memcmp's to filter out all packets other than broadcast and those directed to its MAC address. Ideally, this would be done during driver init, since ktcp isn't doesn't need promiscuous mode.
> 
Ruling out the use of promiscuous mode simplifies the driver a lot since we can depend on the nic filtering all the time. Further - when ktcp get the packet from the driver, we already know if there was an error (there wasn't since we've programmed them out) and why the packet was picked up (unicast, broadcast). The memcmp's aren't really required.
> Thus many of the sanity checks in the old driver (and still present in the rewritten version) are actually superflous unless someone change the setup/initalization.
> 
> Like what, can you elaborate?
> 
Like - the driver currently (I'm deleting this later) tests for runt packet and long packets. Will never happen since the NIC filters them.
> An extra header (4 bytes) in front of every received packet as transferred to the driver contains a status byte indicating / among other things / whether the address was unicast or multi/broadcast, which errors - if allowed etc. Unused at this stage, but could save some processing if we wanted to.
> 
> Is this one of the options that can be turned on? I'm not sure we need this particular info, other than discarding the packet if there's an error. Ktcp doesn't use unicast/multicast, etc.
> 
> This is what I was alluding to above. Unicast is the regular case, filtering on the programmed mac address, broadcast is ARP only. And there is not reason to compare, we have the status already, except it's hidden below ktcp.
> 
> After our discussion I came to the conclusion that the best way to do this is to merge the overrun handling with a regular read. The low level read can easily check for an overrun condition before anything else and do whatever recovery necessary 'in-band' instead of the 'out-of-band' scenario I proposed initially. I'll make that change and test this weekend.
> 
> That idea occurred to me, but I didn't know whether the interrupts could be merged.
> 
They're not, but a buffer overrun means there is plenty to read, so it's a manual merge sort of.
> Sounds like doing both in the read routine solves the issues with the card ring buffer and is the right thing to do.
> 
After testing, the obvious (!) came to play: Relieving the buffer of one packet is preparing for the next overrun to happen. So I'm experimenting with shutting down the receiver till the buffer is empty or maybe half empty, which requires some state in the driver that I don't like.
> The ring buffer mechanism used in the NIC is inspirationally simple and smart. The (configurable) 14.5k receive buffer can hold up to 54 packets with no overhead. I guess that's subject for another day...
> 
> So, your testing on ring buffer overflow requires > 54 packets (ping?) in order to overflow? Is it set up for max packets now? I know very little about what settings are being used in the current or revised driver.
> 
Yeah, this is the way it's set up, now and before. The old driver had the same buffer initialization but an overly complicated and seemingly erroneous logic for the buffer wrap-around handling ( and a bunch of other stuff that did no harm other than taking time and being superflous). Now, what it takes to overflow obviously depends on. Pinging large packets, short intervals, many sources, ARPs on the wire etc., and not the least - whether any traffic sends ELKS to the floppy (i.e. dead for 3-5-8 seconds while searching and excecuting...). In the meantime the RX interrupts are piling up, buffer filling up. BTW - I've seen test cases where the NIC buffer fills to 60-70%, then elks slowly catching up - it's inspiring, in particular when passing the wraparound point without a hitch. ... :-)

The early cards have 16k of buffer space, of which 1.5K is set aside (programmatically) for transmit. The 8390 NIC chip can handle 64 k (newer chips even more), and I have on my list to figure out a way to detect the RAM size on the card and take advantage of it. For now all are deemed equal in the face of elks: 16k.

Really, this is all about robustness: the old driver had no provision for handling buffer overruns. The NIC would hang in such cases. My goal is for elks ethernet to remain alive under attack so to speak, while happily discarding packets when necessary. And let the network handle retransmissions - which it is quite good at.

BTW - one of the NS application notes for the 8390 has code samples for a driver. Full of bugs and incomplete, but still useful. In the example, buffer overflow handling is part of read, but it is assumed that a low level read continues until the on board ring buffer is empty, i.e. it transfers to a higher level buffering mechanism. Fully asynchronous. ELKS does one packet at a time. 
Adding such buffering - like we've talked about a few times - is technically easy, but does require the RAM space. When or if we decide to take that route, it may be interesting to look at the NIC ring buffer mechanism for its simplicity and (memory) efficiency. It's using blocks ('pages') of 256 bytes - which among other things has the advantage that we can address 64k of memory in one byte. And for almost all interactive traffic, 256 bytes will hold a full tcp packet - a remarkable low level efficiency. 

It has been an extremely interesting and educating journey.


//M

ghaerr:
> After testing, the obvious (!) came to play: Relieving the buffer of one packet is preparing for the next overrun to happen.

Not sure what you mean? The read() routine copies a packet out of the hw buffer, but also passes it to the waiting process (ktcp). There's one less packet in the buffer, so another overrun is less likely.


> So I'm experimenting with shutting down the receiver till the buffer is empty or maybe half empty, which requires some state in the driver that I don't like.

I am not sure that's a good idea - it would be better to keep the receiver simple and always operational, and let ELKS either keep up, or not. When not, later-received packets will be dropped as overruns, and ELKS will read the next packet when the application is ready and requests it from the FIFO buffer and reset any overrun.

> any traffic sends ELKS to the floppy (i.e. dead for 3-5-8 seconds while searching and excecuting...). In the meantime the RX interrupts are piling up, buffer filling up.

Now that is a serious problem I hadn't considered! Currently, ELKS uses the BIOS for all floppy I/O, which means it gives control to the BIOS and nothing happens in ELKS until it returns. If interrupts are enabled, which they likely are, serial input will still queue into kernel ring buffers, but the ethernet card will have to buffer received packets. We are too tight on kernel data space to buffer ethernet packets, and there'd be nowhere to send them, since no ELKS processes can run anyways.

I think this problem may best be fixed with MTU and window size, if the packets are coming from the same host. There will be a limit on how many remote hosts ELKS can handle simultaneously. Frankly, ktcp barely runs long enough to worry about this too much at this point, and there are tons of other problems that need fixing within ktcp.

I will have to look at it, it could be possible to schedule out the task doing the disk I/O more during larger disk operations, but that should wait until an analysis is done to determine how often larger buffered disk I/O actually takes place. Another option could be to use async BIOS I/O, but early PCs don't support it.

>  i.e. it transfers to a higher level buffering mechanism. Fully asynchronous. ELKS does one packet at a time.

The problem with ELKS is that almost all the buffers are small, including the application's doing the reading. ELKS barely has space for efficient throughput, and all kernel I/o is buffered at 1k blocks, with disk I/O never in larger chunks. For this reason, I've already started tuning SLIP for <= 1024 byte MTU, and we might want to look at that with ethernet too. This can all be better analyzed once we have TCP/IP working reliably. For instance, ktcp doesn't even honor asked-for window sizes currently.

I'm mid-way through implementing the asynchronous ARP fix, I should have something in a day or so. The changes will only be in ktcp, so it should be easy to test with your enhanced driver.

Mellvik:
> ﻿
> After testing, the obvious (!) came to play: Relieving the buffer of one packet is preparing for the next overrun to happen.
> 
> Not sure what you mean? The read() routine copies a packet out of the hw buffer, but also passes it to the waiting process (ktcp). There's one less packet in the buffer, so another overrun is less likely.
> 
That's the thing. It's very likely. When we have overruns (should be rare cases in real life, but this is testing and hardening), the problem is that the elks hw is a lot slower than the network. So the freed up space is filled with another packet before elks has had the time to process it, and we have a new overrun.
> So I'm experimenting with shutting down the receiver till the buffer is empty or maybe half empty, which requires some state in the driver that I don't like.
> 
> I am not sure that's a good idea - it would be better to keep the receiver simple and always operational, and let ELKS either keep up, or not. When not, later-received packets will be dropped as overruns, and ELKS will read the next packet when the application is ready and requests it from the FIFO buffer and reset any overrun.
> 
In many cases, elks will not be able to keep up. That's the situation I'm attempting to address. Even when all the printk's etc are gone, this will happen occasionally.
> any traffic sends ELKS to the floppy (i.e. dead for 3-5-8 seconds while searching and excecuting...). In the meantime the RX interrupts are piling up, buffer filling up.
> 
> Now that is a serious problem I hadn't considered! Currently, ELKS uses the BIOS for all floppy I/O, which means it gives control to the BIOS and nothing happens in ELKS until it returns. If interrupts are enabled, which they likely are, serial input will still queue into kernel ring buffers, but the ethernet card will have to buffer received packets. We are too tight on kernel data space to buffer ethernet packets, and there'd be nowhere to send them, since no ELKS processes can run anyways.
> 
Yes, interrupts are indeed enabled, I have a single char being printed for each interrupt and can watch them piling up - literally. BTW the best roundtrip time I have on elks/8390NIC/386-20MHz and nothing else on the wire is 170ms, which gives us an idea about max performance. 
> I think this problem may best be fixed with MTU and window size, if the packets are coming from the same host. There will be a limit on how many remote hosts ELKS can handle simultaneously. Frankly, ktcp barely runs long enough to worry about this too much at this point, and there are tons of other problems that need fixing within ktcp.
> 
Agreed about ktcp - and at the same time, the level we're at is encouraging. When ARP/ICMP/driver level are stable (real soon now :-) ) we have a working system with stable outbound, somewhat stable inbound telnet. I believe I have data (logs) that will make the remaining incoming telnet issues shallow. The sudden disconnects are repeatable and predictable.

Back to packet loss: Actually we are discussing two different problems. The one at hand is handling the fact that there will be situations in which the performance discrepancy between the system and the network is impossible. We cannot get the packets to ktcp fast enough regardless. Discards are unavoidable and our discussion is really about smart discarding, not avoidance.
The next level is IP and above, which can take advantage of the mechanisms you mention. @ reasonable loads, the NIC buffer will even out transients at the OS level, and ktcp can do the smart things with MTU & windows. In particular, cards with larger buffers (if I can figure out how to detect mem size) will be heaven sent for this situation. I'm switching to a smarter card when things are stable, but don't yet know the buffer size.
> I will have to look at it, it could be possible to schedule out the task doing the disk I/O more during larger disk operations, but that should wait until an analysis is done to determine how often larger buffered disk I/O actually takes place. Another option could be to use async BIOS I/O, but early PCs don't support it.
> 
Agreed, this requires some research. BIOS disk and floppy I/O should be capable of using DMA - maybe they're using it – why would there otherwise be reserved DMA channels for them? Another project for another day.
> i.e. it transfers to a higher level buffering mechanism. Fully asynchronous. ELKS does one packet at a time.
> 
> The problem with ELKS is that almost all the buffers are small, including the application's doing the reading. ELKS barely has space for efficient throughput, and all kernel I/o is buffered at 1k blocks, with disk I/O never in larger chunks. For this reason, I've already started tuning SLIP for <= 1024 byte MTU, and we might want to look at that with ethernet too. This can all be better analyzed once we have TCP/IP working reliably. For instance, ktcp doesn't even honor asked-for window sizes currently.
> 
> I'm mid-way through implementing the asynchronous ARP fix, I should have something in a day or so. The changes will only be in ktcp, so it should be easy to test with your enhanced driver.
> 
Looking forward to it. I keep breaking and fixing the ne2k driver daily while experimenting. It's getting smaller and smaller. :-)
BTW, while we're on ARP and buffers: Our kludge with an extra transmit buffer for outgoing packets to avoid overwrites: Setting aside a max-ether-pkt-size buffer is really wasteful. The ARP packet is less than 64 bytes.

--Mellvik

PR #635, opened 2020-05-16
Add list header function to chmem utility

georgp24:
Added a `-l` parameter to chmem to list the complete header of the executable.

### DIFF
--- a/elkscmd/sys_utils/chmem.c
+++ b/elkscmd/sys_utils/chmem.c
@@ -54,7 +54,8 @@ int main(int argc, char **argv)
 
   p = argv[1];
   if (argc != 3) 
-	fatalmsg("Usage: %s {=+-}<# bytes dynamic data> <executable>\n", argv[0]);
+	fatalmsg("Usage: %s {=+-}<# bytes dynamic data> <executable>\n\
+       or\n       %s -l <executable> to list the header\n", argv[0], argv[0]);
   if (*p != '=' && *p != '+' && *p != '-') usage();
   lsize = atol(p+1);
   if (lsize > MAX) fatalmsg("chmem: %lu too large, max %lu\n", lsize, MAX);
@@ -72,6 +73,12 @@ int main(int argc, char **argv)
 	olddynam = 0;
   else olddynam = header[TOT] - dsegsize;
   if (separate == 0) olddynam -= header[TEXT];
+  
+  if (*(p+1)=='l') {
+        printf("Header of '%s': TEXT %lu DATA %lu BSS %ld TOT %lu DYNMEM %lu [TOT-(DATA-BSS)]\n",
+	    argv[2], header[TEXT], header[DATA], header[BSS], header[TOT], olddynam);
+        exit(0);
+    }
 
   printf("Old %s: DATA %lu BSS %ld TOT %lu DYNMEM %lu\n",
 	argv[2], header[DATA], header[BSS], header[TOT], olddynam);


tkchia:
Actually Minix (where `chmem` came from) also happens to come with a [`size` utility](https://github.com/davidgiven/minix2/blob/master/minix-2.0/fs/usr/src/commands/simple/size.c).  I think another possibility might be to port that over to ELKS too...

Thank you!

Mellvik:
That would be useful, @thchia - 
I was looking for size() a few weeks back, it has been in unixes since V7.

—Mellvik

> 
> Actually Minix (where chmem came from) also happens to come with a size utility <https://github.com/davidgiven/minix2/blob/master/minix-2.0/fs/usr/src/commands/simple/size.c>. I think another possibility might be to port that over to ELKS too...
>

georgp24:
When I look at the minix size utility it prints out:
`exec.a_text, exec.a_data, exec.a_bss, dynam, allmem, name`
the `-l` parameter in chmem prints:
`argv[2], header[TEXT], header[DATA], header[BSS], header[TOT], olddynam`
This is the same information, just the name is at the beginning and dynam after total(allmem).

Should we make a symbolic link called size to `chmem -l`?

ghaerr:
@georgp24,

Is this really needed? I use `chmem +0 ...` to display the header.

I plan on updating `chmem` in the near future with support for separate stack and heap sizes... this PR adds a third unneeded printf instead of reusing code. How about using the +0 option for now and I'll add an option to display the header using the existing printfs after I add the heap/stack support?

georgp24:
Mellivik, tkchia and I did not know that you can use +0 to display header fields. This also does not show the TEXT segment size. I thought if you already use two printf statements I could use one too.

Shall I make a separate program called size?

tkchia:
Hello @georgp24,

> Shall I make a separate program called size?

There _is_ already a `size` program under Minix.  We should be able to use it under ELKS with a few changes.

Thank you!

georgp24:
Hallo tkchia,
I had seen that you mentioned the size program above. When I wrote my remark I thought it would be quicker to base this on the chmem code which I know quite well now.

When following your link I found an old [df program](https://github.com/davidgiven/minix2/blob/826a75d9e62686af19cb67e0261e5a63206146d8/minix-1.7/fs/usr/src/commands/simple/df.c) which ELKS is lacking. However, I guess this is difficult to port to ELKS.

Georg

georgp24:
I just checked, I could simply replace my printf with fatalmsg(). Works just as well.

Mellvik:
FWIW - yes, I do think there should be a size program because that's what so many of us expect (and thus would be looking for). If it's a hard link to something else - like chmem, all the better (reduced size, fewer components).


—Mellvik

ghaerr:
I have looked further into the idea of a `size` program for ELKS, using the suggested minix2 version. While I agree that a separate size utility is the way to go, the useful functionality of displaying the dynamic memory separately from the total, and its calculations, are included in chmem.c but not size.c. In addition, it seems I took our chmem.c from minix1 rather than minix2, before the a.out.h header was present, and it doesn't use the minix_header struct, which should be fixed.

Finally, given that @tkchia is just now introducing medium model support which adds some new headers whose contents may want to be reported, and that I will be very shortly be introducing the seperate "min_stack" enhancement, I think the best route will be for me to update chmem to use the minix.h header file and the min_stack changes, and then create a size.c from there. 

Including this PR now will only complicate the updating of chmem min_stack, as there will be 3 printfs to update, rather than 2 (which I hope to reduce to 1). So lets hold on this a bit longer.

ghaerr:
@georgp24,

I rewrote `chmem` with your comments in mind, this has been committed in #644.
Please take a look at it, it also doubles as a size program, and then close this PR if you're satisfied.

I am aware of a shell or exec() memory problem when doing "cd /bin; chmem \*", to see the sizes of all executables. This will be fixed in a different issue.

Issue #634, opened 2020-05-16
Linenoise translates the delete key to backspace


georgp24:
As the title says, when you press the delete key, a backspace key is generated. The character at the cursor position should be removed and the rest of the line moved one position to the left.

ghaerr:
Hello @georgp24.

> The character at the cursor position should be removed and the rest of the line moved one position to the left.

Ok.

I will raise the priority of my planned fix for this, the full fix is a bit more complicated. 
Looks we need to completely separately handle DEL vs BS. The additional issue is `stty` default settings on serial logins. My macOS laptop doesn't have a separate BS key and they've placed `delete` where Backspace should be, and it sends DEL. For serial logins, Delete doesn't erase characters.

I think the full solution is to separate DEL and BS as suggested, and add a config option (for at least myself, although default OFF) that sends a BS when the Delete key is sent, implemented in the BIOS and Direct console drivers.

georgp24:
I use almost exclusively the delete key. Only when erasing mistyped passwords I use backspace. So the conversion of the delete key to a backspace key in linenoise is a nuisance for me.

ghaerr:
@georgp24, I committed your PR #636. Please close this issue if that handles this problem, thanks.

georgp24:
Great that it works now!

PR #636, opened 2020-05-17
Handle delete key differently from backspace in linenoise

georgp24:
Made a simple change that handles the delete key correctly. Works fine for me. Based on a pull request at the GitHub kilo repo: [RiordanIX - Made DEL work like in Nano](https://github.com/antirez/kilo/pull/46/commits/8ab276f650398c2c17b3f84d5aa9ff7675b968ac)

### DIFF
--- a/elkscmd/ash/linenoise_elks.c
+++ b/elkscmd/ash/linenoise_elks.c
@@ -192,7 +192,7 @@ enum KEY_ACTION{
 	CTRL_D = 4,         /* Ctrl-d */
 	CTRL_E = 5,         /* Ctrl-e */
 	CTRL_F = 6,         /* Ctrl-f */
-	CTRL_H = 8,         /* Ctrl-h */
+	CTRL_H = 8,         /* Ctrl-h - Backspace*/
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello Georg,

I will commit this, but as mentioned in https://github.com/jbruchon/elks/issues/634#issuecomment-629671913 I want to create an option for Delete equals Backspace and handle some other related issues before doing so. If committed now, the only dedicated Delete key on my macOS laptop would stop working for backspace.

georgp24:
I would expect your delete key to work as a delete key and Ctrl-H to work as backspace.

ghaerr:
Hello @georgp24,

Sorry this took so long. I finally looked into this and QEMU on macOS is sending a ^H for the Delete key, so there is probably no need for a separate config option as discussed in #634.

The "Backspace" key above the pipe symbol is sending ^H, even though it is labeled Delete on my Mac Pro laptop, and the keypad Delete key is sending 127 on my iMac. No key is sending the Delete "ESC [ 3 ~" sequence, so I think it's all OK.

PR #642, opened 2020-05-20
[cmds] Add df (disk free) program

ghaerr:
Add `df` utility for elks, ported from Minix 2.

Hello @tkchia,

The BIOS floppy disk driver (/dev/fd) does a probe on driver open(), even though the disk hasn't changed. Do you have any ideas for how to best get rid of the "fd: ..." kernel probing messages (see below)? We probably ought to keep them for boot, and also might want them when the floppy has changed. Or we could just disable them?

Screen Shot 2020-05-20 at 1 01 16 AM

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -59,6 +59,7 @@ file_utils/chmod				:be-fileutil	:360k
 file_utils/chown				:be-fileutil		:720k
 file_utils/cmp					:be-fileutil			:1440k
 file_utils/cp					:be-fileutil	:360k
+file_utils/df					:be-fileutil			:1440k
 file_utils/dd					:be-fileutil			:1440k
 file_utils/mkdir				:fileutil		:360k
 file_utils/mknod				:fileutil				:1440k
--- a/elkscmd/file_utils/.gitignore
+++ b/elkscmd/file_utils/.gitignore
@@ -4,6 +4,7 @@ chmod
 chown
 cmp
 cp
+df
 dd
 grep
 l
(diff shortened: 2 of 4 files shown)


tkchia:
Hello @ghaerr,

> We probably ought to keep them for boot, and also might want them when the floppy has changed. Or we could just disable them?

I think that is a possibility --- use the BIOS (viz. `int 0x13` functions `0x15` and `0x16`) to detect disk changes, for those floppy drives that support this.

Another possibility: presumably `df` only needs to read `/dev/fd0` if it is actually mounted somewhere in the file hierarchy?  If a floppy disk is still somehow "in use", I think it is reasonable to assume that the disk has not been changed.

Thank you!

ghaerr:
Thanks @tkchia.

> I think that is a possibility --- use the BIOS (viz. int 0x13 functions 0x15 and 0x16) to detect disk changes, for those floppy drives that support this.

QEMU reports no disk change recognition capability with function 0x15, so I've left this out for now, and I think it is not easily tested, after looking at both mount and device open code.

> Another possibility: presumably df only needs to read /dev/fd0 if it is actually mounted somewhere in the file hierarchy? If a floppy disk is still somehow "in use", I think it is reasonable to assume that the disk has not been changed.

I also looked into this for some time, and have decided that I can't reliably remove floppy probing using this method, as the floppy is mounted early in the boot sequence, prior to the bios_open routine being called.

For now, I think its OK to leave in the disk probing, and the subsequent kernel printk.
While not pretty, we now have disk free space reporting (for MINIX drives only).

Thank you!

PR #643, opened 2020-05-20
Some fixes for elksemu for dealing with process ids and signals

tkchia:


ghaerr:
Hello @tkchia,

That's quite a bit of logic to properly handle elks vs linux pids... how long has this bug been in there, since the beginning, not just since elks pids were fixed to 15 bits, right?

tkchia:
Hello @ghaerr,

> That's quite a bit of logic to properly handle elks vs linux pids... how long has this bug been in there, since the beginning, not just since elks pids were fixed to 15 bits, right?

I guess it was not actually a bug --- at least, not until I ported `elksemu` to x86-64.  `elksemu` was originally written to run only on Linux/x86-32, where --- according to the `proc`(5) man page --- pids can go only up to 32,767 (just like in ELKS!).  But now `elksemu` also runs under Linux/x86-64, where pids can go all the way up to 2²² - 1.

I hit upon this glitch while trying to test out a patch to update `dev86`'s implementation of `_signal` (https://github.com/jbruchon/dev86/pull/21).  I found that
```
raise (SIGQUIT);
```
under `elksemu` was constantly trying to send a signal to pid 65,535 (`0xffff`).

Thank you!

jbruchon:
I don't care about elksemu getting bigger if needed. Plenty of room. If it solves a problem then I have no objection to the change.

ghaerr (approved):
Hello @tkchia,

I have to admit that I don't understand it; not sure how a previous direct mapping of elks-to-linux pids worked, and why now a complex mapping is required. (I also can't test elksemu since I'm running on macOS).

Thank you!

tkchia:
Hello @ghaerr,

> I have to admit that I don't understand it; not sure how a previous direct mapping of elks-to-linux pids worked, and why now a complex mapping is required.

Yes, my previous code was probably much more complex than needed.  I have modified the pid mapping logic to make it much more straightforward.

The original way (which works for Linux/x86-32) was to simply treat Linux and ELKS pids as interchangeable.  This works because pids under Linux/x86-32 can never exceed 32,767 anyway, like in ELKS.

On Linux/x86-64, pids can exceed 32,767.  So it is possible that `elksemu` is started under a large pid, which would pose problems if the ELKS program tries to do stuff with its own pid (e.g.).

> (I also can't test elksemu since I'm running on macOS).

Do you think that `elksemu` is something you would like to have working on macOS?  Perhaps I will see if I can help somehow...

Thank you!

ghaerr:
Hi @tkchia,

> On Linux/x86-64, pids can exceed 32,767. So it is possible that elksemu is started under a large pid, which would pose problems if the ELKS program tries to do stuff with its own pid (e.g.).

Oh, now I get it!

> Do you think that elksemu is something you would like to have working on macOS? Perhaps I will see if I can help somehow...

Given that macOS is quite different than Linux, that's probably way more trouble than its worth.

It would sometimes be nice to have a debugger that allows for procedure-level tracing, but given all the other current issues with multiple other emulators, I'll stick to my older-style printf/printk debugging for now, unless someone has a nice graphical interface for development :)

The kernel is and has been quite stable with regards to basic stack-switching, memory management and interrupt handling, and most applications are starting to become bug-free, with the huge exception of `ktcp`, which is proving to be quite a mess. The last undebugged portion of ktcp is tracing through the user<->kernel socket /dev/tcp implementation, which still has problems.

PR #644, opened 2020-05-21
[kernel] Add configurable process heap+stack size

ghaerr:
Adds separately configurable heap and stack size for ELKS processes.
Adds per-process minimum stack size to kernel.

These will be used shortly for fine-tuning heap and stack allocations on a per-process basis to help get most ELKS applications to run in a default 2k heap, 2k stack environment.

Completely rewrote `chmem`, allows heap and stack to be specified with -h and -s options.
Can also be run with multiple arguments to display sizes of multiple programs.
New usage is:
`chmem [-h heap] [-s stack] file ...`
When run with no options, it functions as a size program (fixes request in #635).
Heap and stack values can be specified in hex or decimal.

The elks-small.ld linker script has not yet been modified to store the upper 16-bits of the -maout-chmem= value into the new a.out minstack header field.

Screen Shot 2020-05-20 at 9 12 19 PM

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -11,8 +11,6 @@
 
 #include <arch/segment.h>
 
-#define MIN_STACK_SIZE 0x1000	/* 4k min stack above heap*/
-
 // TODO: reduce size
 
 // Segment descriptor
@@ -225,18 +223,17 @@ int sys_brk(__pptr newbrk)
     if (newbrk < currentp->t_enddata)
         return -ENOMEM;
 
-    if (currentp->t_begstack > currentp->t_endbrk) {	/* Old format? */
-        if (newbrk > currentp->t_endseg - MIN_STACK_SIZE) {	/* Yes */
+    if (currentp->t_begstack > currentp->t_endbrk) {				/* stack above heap?*/
+        if (newbrk > currentp->t_endseg - currentp->t_minstack) {
 			printk("sys_brk(%d) fail: brk %x over by %d bytes\n",
-				currentp->pid, newbrk, newbrk - (currentp->t_endseg - MIN_STACK_SIZE));
+				currentp->pid, newbrk, newbrk - (currentp->t_endseg - currentp->t_minstack));
             return -ENOMEM;
 		}
     }
 #ifdef CONFIG_EXEC_LOW_STACK
-    if (newbrk > currentp->t_endseg) {
+    if (newbrk > currentp->t_endseg)
         return -ENOMEM;
-    }
-#endif /* CONFIG_EXEC_LOW_STACK */
+#endif
     currentp->t_endbrk = newbrk;
 
     return 0;
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -52,11 +52,6 @@ static struct minix_exec_hdr mh;
 static struct elks_supl_hdr esuph;
 #endif
 
-// Default data sizes
-
-#define INIT_HEAP  0x0     // For future use (inverted heap and stack)
-#define INIT_STACK 0x2000  // 8 K (space for both stack and heap)
-
 #ifndef __GNUC__
 /* FIXME: evaluates some operands twice */
 #   define add_overflow(a, b, res) \
@@ -75,8 +70,10 @@ static struct elks_supl_hdr esuph;
 #endif
 
 #ifdef CONFIG_EXEC_MMODEL
-// Read relocations for a particular segment and apply them
-// Only IA-16 segment relocations are accepted
+/*
+ * Read relocations for a particular segment and apply them
+ * Only IA-16 segment relocations are accepted
+ */
 static int relocate(seg_t place_base, lsize_t rsize, segment_s *seg_code,
 		    segment_s *seg_data, __ptask currentp, struct inode *inode,
 		    struct file *filp)
@@ -133,7 +130,7 @@ int sys_execve(char *filename, char *sptr, size_t slen)
     seg_t base_data = 0;
     segment_s * seg_code;
     segment_s * seg_data;
-    size_t len, min_len;
+    size_t len, min_len, stack;
 #ifdef CONFIG_EXEC_MMODEL
     int need_reloc_code = 1;
 #endif
@@ -186,10 +183,6 @@ int sys_execve(char *filename, char *sptr, size_t slen)
     if (add_overflow(min_len, mh.bseg, &min_len))
 	goto error_exec3;
 
-    // TODO: revise the ELKS specific executable format
-    // as we now master the executable header content
-    // with the new GNU build tool chain (custom LD script)
-
 #ifdef CONFIG_EXEC_MMODEL
     memset(&esuph, 0, sizeof(esuph));
 #endif
@@ -224,37 +217,41 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 	if (base_data != 0)
 	    debug1("EXEC: New type executable stack = %x\n", base_data);
 
-	if (add_overflow(min_len, base_data, &min_len))
+	if (add_overflow(min_len, base_data, &min_len))	/* adds stack size*/
 	    goto error_exec3;
 #else
 	if (base_data != 0)
 	    goto error_exec3;
 #endif
 	break;
-#endif
+#endif /* CONFIG_EXEC_MMODEL*/
     default:
 	goto error_exec3;
     }
 
     /*
-     * Size for data segment
-     * mh.chmem was used by old ld86
-     * New LD script sets this to zero (default)
+     * chmem is size of data+bss+heap, 0 means use default heap
+     * old ld86 used chmem as size of data+bss+heap+stack
      */
-    len = mh.chmem;
+    len = mh.chmem;					/* = data+bss+heap*/
     if (!len) {
 	len = min_len;
-#ifdef CONFIG_EXEC_LOW_STACK
-	if (base_data) {
-	    if (add_overflow(len, INIT_HEAP, &len))
-		goto error_exec3;
-	} else
-#endif
-	if (add_overflow(len, INIT_HEAP + INIT_STACK, &len))
+	if (add_overflow(len, INIT_HEAP, &len))		/* add default heap*/
 	    goto error_exec3;
-    } else if (len < min_len)
+    } else {
+	debug2("heap %u len %u\n", mh.chmem - min_len, len);
+	if (len < min_len)
+	   goto error_exec3;
+    }
+#ifndef CONFIG_EXEC_LOW_STACK
+    stack = mh.minstack;
+    /* if no explicit stack specified, allow older max heap+stack programs to run*/
+    if (!stack && len < 0xFFF0)
+	stack = INIT_STACK;
+    if (add_overflow(len, stack, &len))			/* add stack*/
 	goto error_exec3;
-
+    if (stack) debug1("stack %u\n", stack);
+#endif
     /* Round data segment length up to a paragraph boundary */
     if (add_overflow(len, 15, &len))
 	goto error_exec3;
@@ -351,6 +348,7 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 	: currentp->t_endseg) - slen;
     currentp->t_regs.sp = (__u16)(currentp->t_begstack);
     fmemcpyb((byte_t *)currentp->t_begstack, seg_data->base, (byte_t *)sptr, ds, (word_t) slen);
+    currentp->t_minstack = stack;
 
     /* From this point, the old code and data segments are not needed anymore */
 
--- a/elks/include/linuxmt/minix.h
+++ b/elks/include/linuxmt/minix.h
@@ -18,6 +18,10 @@
 #define MINIX_S_SPLITID	0x04600301UL
 #define MINIX_DLLID	0x04A00301UL
 
+/* Default data allocations*/
+#define INIT_HEAP  4096		/* Default heap*/
+#define INIT_STACK 4096		/* Default stack*/
+
 struct minix_exec_hdr {
     unsigned long	type;
     unsigned char	hlen;
@@ -27,7 +31,8 @@ struct minix_exec_hdr {
     unsigned short	dseg, reserved3;
     unsigned short	bseg, reserved4;
     unsigned long	entry;
-    unsigned short	chmem, reserved5;
+    unsigned short	chmem;
+    unsigned short	minstack;
     unsigned long	syms;
 };
 
--- a/elks/include/linuxmt/sched.h
+++ b/elks/include/linuxmt/sched.h
@@ -52,6 +52,7 @@ struct task_struct {
     __pptr			t_begstack;
     __pptr			t_endbrk;
     __pptr			t_endseg;
+    int				t_minstack;
 
 /* Kernel info */
     pid_t			pid;
--- a/elkscmd/bc/Makefile
+++ b/elkscmd/bc/Makefile
@@ -34,7 +34,7 @@ LOCALFLAGS=-D_POSIX_SOURCE
 CFLBASE += -mrtd
 
 # For ELKS, bc needs more data segment space than the kernel-given default.
-LDFLAGS += -maout-chmem=0xc000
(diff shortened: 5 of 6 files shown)


tkchia (approved):

tkchia:
Hello @ghaerr,

Thanks for the PR!  (It does look like the vestigial `CONFIG_EXEC_LOW_STACK` code is starting to cause maintenance issues.  I would have no problem if you decide to throw it out.)

> The elks-small.ld linker script has not yet been modified to store the upper 16-bits of the -maout-chmem= value into the new a.out minstack header field.

Thanks.  I am currently modifying the `gcc-ia16` front-end to add `-maout-stack=`... and `-maout-heap=`... options, to (at least) define symbols `_stack` and `_heap` so that `elks-small.ld` can use them.  In the long term I also hope to get `-maout-chmem=` etc. options to work for the _medium_ model --- currently there is no support for them.

Thank you!

ghaerr:
Hello @tkchia,

>  I have just been working on separate -maout-stack= and -maout-heap= parameters in the gcc-ia16 front-end, which will end up defining the _stack and/or _heap symbols for the linker script.

Oh, nice!

Do you plan on keeping -maout-chmem= ? If so, will it be effective there for backwards compatibility and work identically to -maout-heap= ?

Hopefully it is clear from the `chmem` and kernel source how the heap and stack parameters are passed using `chmem` and used in my PR. But just to be perfectly clear: the stack value is *not* added to the `chmem` field, but stored directly in the minstack header field. If a heap value is given, it is added to the data+bss values then stored in the chmem field. The chmem field no longer holds the total data segment allocation; the specified or default stack must also be added, at kernel load time only. Keeping a concept of default heap and stack is valuable, which allows all ELKS programs to be automatically squeezed down in allocation requirements by slowly lowering the INIT_HEAP and INIT_STACK defaults, and only adding -maout-heap/stack to those programs that need special attention.

This business of adding the data segment size and the heap (=total) into chmem is of dubious value, except for backwards compatibility with minix a.out programs. This is because with the various ABI changes made, older MINIX programs likely wouldn't run anyways. Of course, new programs could be developed (and have been using Tiny model), using `dev86`. 

So there is some consideration to be given as to whether ELKS should support MINIX binaries-only and keep the chmem field meaning under MINIX. If `dev86` tools were updated, (which is quite straightforward, only used in `ld86`), then it is a possibility to make the chmem field be heap-only. Or just leave it all alone, which was the path I've taken.

A better permanent move would be to relabel the `chmem` field to be heap and use another bit in the header type field to indicate heap-only, but this just involves even more code, and if MINIX compatibility means anything, the backwards compatible code would have to be handled anyways.

A final issue I've come across is handling a zero heap request from `chmem` (or now `gcc`). If the stack is set, but not heap, then the passed 0 heap ends up interpreted as default heap. The workaround is when 0 heap is desired, to pass 1 byte instead. At the end of the day, for chmem, one has to pass both heap and stack at the same invocation, or the non-passed field will end up being zeroed (which means default).

I hope all this makes sense!

Thank you!

tkchia:
Hello @ghaerr,

> So there is some consideration to be given as to whether ELKS should support MINIX binaries-only and keep the `chmem` field meaning under MINIX. If dev86 tools were updated, (which is quite straightforward, only used in `ld86`), then it is a possibility to make the `chmem` field be heap-only. Or just leave it all alone, which was the path I've taken.

Hmm... actually, if backwards compatibility is a goal, I think it might be simpler to let `mh.chmem` continue to refer to the total data segment size, as before --- at least, when `mh.minstack` == 0.  I just looked at some of the code in `dev86`, and there is quite a fair amount of code that assumes that `__heap_top` --- the name given by `ld86` to the field value --- can be used to initialize `sp`, or be used to gauge the total data segment size.  (See e.g. `bootblocks/boot_win.c` and `libc/bios/bios.c`.)

A possible alternative arrangement: for new `a.out` files, instead of using 0 to denote a "default" heap or stack size, use a different magic value, say `0xffff`.  Then `mh.minstack` = 0 or 1 will indicate an "old-style" `a.out` file where `mh.chmem` has the old meaning --- since a 0-byte or 1-byte stack is definitely bogus.  And we are free to give new meanings to `mh.chmem` for cases where `mh.minstack` = `0xffff`.

Well, it does seem that this data segment allocation problem gets more complicated the more I look at it... :-|

Thank you!

ghaerr:
> Hmm... actually, if backwards compatibility is a goal, I think it might be simpler to let mh.chmem continue to refer to the total data segment size, as before --- at least, when mh.minstack == 0.

Well, that's the way I've got it coded for now, and it works.

> I just looked at some of the code in dev86, and there is quite a fair amount of code that assumes that __heap_top --- the name given by ld86 to the field value --- can be used to initialize sp, or be used to gauge the total data segment size. (See e.g. bootblocks/boot_win.c and libc/bios/bios.c.)

I'll take a look at that. Has `dev86` been kept up, and tested, with regards to ABI changes? The problem is, the compiler is so old, I'm wondering whether there's much use in it all, except for perhaps to produce "tiny"-style binaries. Frankly, it is probably best just left alone, and then make changes to ELKS should we need to for it, like you did the for AHISTORICAL header, etc.

> A possible alternative arrangement: for new a.out files, instead of using 0 to denote a "default" heap or stack size, use a different magic value, say 0xffff. 

I think moving to another magic value introduces many more changes than is needed, and throws backwards compatibility into a tedious chore. Also, 0xffff is too close to 0xfff0, and could be found in executables looking for max memory.

The simpler the better, IMO. The new design is very straightforward for stack usage. Extra heap is only compounded by having the added data+bss, and no stack. And it all work well given backwards compatibility, where the extra stack used to be included. The current approach will allocate extra stack. Should this prove not to run, running `chmem -h 0xffff` will correctly calculate the max data+heap+(default)stack and stick it in the chmem field, showing both before and after.

> Then mh.minstack = 0 or 1 will indicate an "old-style" a.out file where mh.chmem has the old meaning --- since a 0-byte or 1-byte stack is definitely bogus.

Min_stack isn't the problem - the value isn't added to anything and is stored. Keep 0=default matches the old settings for heap. The issue is only with the chmem field, and the difference between a 0 byte and 1 byte heap, if we want to maintain backwards compatibility, isn't a big deal.

> And we are free to give new meanings to mh.chmem for cases where mh.minstack = 0xffff.

I don't like the idea of introducing yet more special cases with fields.

> Well, it does seem that this data segment allocation problem gets more complicated the more I look at it... :-|

Yep. I've spent a lot of time thinking about all of the issues - and came up with the current PR, which adds a minimum of kernel code, and gracefully handles backwards compatibility, at least for the time being.

My thoughts are to continue by getting new compiler options and a linker script working, then decide at a later date whether it makes sense to continue with backwards compatibility, or improving what the meaning of chmem should be. Ultimately, later changes won't change either the compiler -maout settings nor the `chmem` settings for an application.

With a linker script change, everything is working and ready to go to start tuning applications, especially those started at boot time, and reduce the overall footprint of required system memory. 

On another note - I was wondering why the medium model linker script didn't use the "binary" output method, but an internal "elks" one, seemingly. I guess its some complicated under-the-hood stuff that prevents an external linker script to calculate the a.out header values for that? I kind of like the small model approach of handling this stuff outside the compiler.

ghaerr:
@tkchia,

In the discussion above of the possible interpretations of chmem and stack fields, in rereading your input and my reply, we may have been talking about slightly different things.

However - this brings up what I think to be an important point - that I think the kernel interpretation of these fields remain very strict, without (many) special interpretations, and instead have the tools deal with the more complex issues. Case in point, despite my earlier comment about not truncating an input value, I decided it would be good practice to allow `chmem -h 0xffff` to always truncate the passed heap value (with a warning), and instead calculate the maximum heap that would be allowed, using both the known program data+bss, and the currently available INIT_STACK size (even though the stack size would not be saved in either chmem nor minstack). This allows ELKS to continue moving forward with kernel process allocation changes with a minimum of Makefile compiler option changes for special programs.

The linker script could operate similarly. This allows user specification of a max heap, without having to play around trying to find a specific value, like we've had to do before. The same could be done for the stack, but I figured to leave it untouched, since the vast majority of programs don't need max stack, they need either 1) no heap, 2) small heap, 3) known heap, 4) small stack, and sometimes, 5)  larger than default stack. All of these are easily specified, and chmem handles the case where a max heap is desired.

The linker script and chmem should IMO write the same header values, given the same heap and stack inputs. I ended up completely rewriting `chmem` so that understanding it would be very straightforward.

tkchia:
Hello @ghaerr,

> Has `dev86` been kept up, and tested, with regards to ABI changes?

Yes, I have kind of been working on that.

> And it all work well given backwards compatibility, where the extra stack used to be included. The current approach will allocate extra stack.

But what is the huge issue with just using the old interpretation of the `mh.chmem` field, that it is necessary to change it?  And would you still want to keep `gcc-ia16`'s `-maout-total=`... option around?

I still think it is simpler to just let it continue to denote the total data segment size (including the stack).  Let the kernel "carve out" a suitable stack area from the `mh.chmem` bytes of data segment memory.  We can even get rid of the `mh.chmem` ≥ `0xfff0` special case checking this way.

I am not sure I see much value in the current "halfway approach" which is neither completely compatible nor completely incompatible with the old way.

There is only one problem I can think of with using the old interpretation (`mh.chmem` = total data): if only the heap size is specified, then the linker (or the `chmem` program) must come up with its own reasonable value for the stack size, and add that in.  But I am not sure that this is such a huge problem.

Thank you!

ghaerr:
> But what is the huge issue with just using the old interpretation of the mh.chmem field, that it is necessar to change it?
> I still think it is simpler to just let it continue to denote the total data segment size (including the stack).

The reason is exactly your last point:

> There is only one problem I can think of with using the old interpretation (mh.chmem = total data): if only the heap size is specified, then the linker (or the chmem program) must come up with its own reasonable value for the stack size, and add that in.

When only a heap size is specified, the older "compatible" way requires that some stack value be assumed, but can't be known. Thus a new stack field is born. It could make sense to always have the stack field pre-added into the chmem field, but then that means that all headers would have to have the stack field non-zero. Having the compiler produce headers with preset stack values goes against the idea of shrinking down process requirements, as all would have to be re-compiled.

> I still think it is simpler to just let it continue to denote the total data segment size (including the stack). 

Ok - except for:

> Let the kernel "carve out" a suitable stack area from the mh.chmem bytes of data segment memory. 

The kernel can't know what that "carve out" should be - especially when that area is now supposed to be just the extra heap value.

> We can even get rid of the mh.chmem ≥ 0xfff0 special case checking this way.

I'm ok with dropping the >= 0xfff0 check, it was thrown in at the last minute for programs like kilo, which were compiled for max memory.

Ultimately, this discussion hinges on a few important, unresolved issues. First, "compatibility". What exactly are we trying to be compatible with? Minix binaries? Dev86 tools? Neither help us. The whole problem with the "compatible" legacy is that the stack value isn't known. The legacy design cannot help us, and actually hinders us, in moving ELKS towards tuning for precisely-specified execution environments.

With a seperate stack value, why keep it added in? The second big discussion item is what, exactly, is "chmem" supposed to be? It started life as a "total" field, as a quick-and-dirty way to specify more (all) extra memory; then, it became called "chmem" because people started thinking of it as heap memory, but for compatibility, kept adding in other header values, and an assumed stack. (In my opinion, that was a mistake made that we're still living with - the total/chmem field should always have been only extra heap memory, and the kernel just read the dseg and bss values separately).

> I am not sure I see much value in the current "halfway approach" which is neither completely compatible nor completely incompatible with the old way.

I don't see much value in "compatible", frankly, given the above and ABI changes. My proposed design tried to keep some compatibility, and removed the stack value and put it in a separate field. Thinking forward, if we were to drop "compatibility", then chmem should be renamed "heap", and used only for heap.

If you feel strongly that "halfway" isn't the way to go - I just thought of another approach - look at the original Minix a.out here: https://github.com/chenshuo/old-minix/blob/master/include/a.out.h
We could break apart the current longword "type" into the original Minix a.out fields (something I think we should do anyways), and use an unused A_HEAPSTACK flag to indicate: if 0) completely compatible interpretation of chmem (data+bss+heap+stack or 0), or if 1) interpret the min_stack and heap fields as *strictly* the stack and heap values. The kernel code would be larger, and all new executables would have this bit set, but it would pave the way for an easier to understand ELKS, while maintain backwards compatibility completely if the bit wasn't set. I don't really like the idea of writing yet more code in exec.c to handle all of this, but perhaps fully supporting the compatible and new ways separately should be seriously considered.

ghaerr:
Hello @tkchia,

> And would you still want to keep gcc-ia16's -maout-total=... option around?

No. Is there a good reason for it? Other than perhaps for having some kind of option for "compatible" executables... but that would require yet another linker script. 

Thank you!

ghaerr:
Hello @tkchia, Instead of using a flags bit, the "version" field could be used... if you can find that it's now always zero (that was most recently added to zero-check in your last commit), then we could increment version to 1, and treat the min_stack and heap fields without concern for backwards compatibility. Version 0 - full compatibility with bcc/ld86/dev86, Version 1 - on to the future, heap and stack specification completely standalone!

[EDIT: backwards compatibility could become a config option, should it seem that there's lots of space allocated to handling legacy versus current needs. Very few people need the legacy option, but would be available separately should they want it, or have a reference while we're building and testing new code, like CONFIG_EXEC_LOW_STACK].

ghaerr:
Hello @tkchia,

I have diagrammed out what I think each separate option would be:
```
v0:
total =
    header.chmem ?  header.chmem :
        (header.dseg +
         header.bss +
         INIT_HEAP +
         INIT_STACK);

v1:
total =
    header.dseg +
    header.bss +
    (header.heap ? header.heap : INIT_HEAP) +
    (header.minstack ? header.minstack : INIT_STACK);

low_stack:
total =
    header.msh_dbase +      // stack
    header.chmem ?  header.chmem :
        (header.dseg +
         header.bss +
         INIT_HEAP);

What we need is something like this, which loads ax,dx with 0's,
then adds ax and an argument with carry into dx. Then dx is checked
at the end:

    __adcstart();
    __adc(header.dseg);
    __adc(header.bss);
    __adc(INIT_HEAP);
    __adc(INIT_STACK);
    if (_adc_carry())
        goto error;
```
With an assembly macro, treating each case separately would use a minimum of code space and be quite clear.

tkchia:
Hello @ghaerr,

> Version 0 - full compatibility with bcc/ld86/dev86, Version 1 - on to the future, heap and stack specification completely standalone!

Thanks --- yes, I think that will be a good way to go forward. :-)  I am now updating the `gcc-ia16` and `binutils-ia16` code to implement your proposal.  For now I think I will leave the `-maout-total=`... and `-maout-chmem=`... compiler options around, but treat them as deprecated.

Thank you!

tkchia:
Hello @ghaerr,

One thing I am trying out, for new `mh.version` = 1 executables, is to separately account for the sizes of the `argp` and `envp` data passed to the new processes, rather than considering them to be part of the `mh.minstack` bytes of stack.

The program being invoked cannot really control the `argp` and `envp` it receives, so including them in `mh.minstack` does not make much sense to me.

Thank you!

ghaerr:
Hi @tkchia,

> The program being invoked cannot really control the argp and envp it receives, so including them in mh.minstack does not make much sense to me.

I hadn't considered that, but agree. Prior, this is probably the reason that, for instance, `chmem /bin/\*` fails with "chmem: not enough memory." However, I need to chase that down to see if that's the shell printing that from the failed `sbrk` in `execvp` or not.

Given this issue, plus the potential issue of a a program size increasing and perhaps not being able to run with its previously-specified heap value, I'm thinking that, at least for now, the kernel might want to truncate the (specified or default) heap value by the amount needed to run, and issue a printk warning the user. The kernel may possibly have to make more judgements on permitting a process to start, given a large argument list and specified heap. 

Also - I have been working on the implementation of the kernel heap/stack myself, and have almost finished an implementation, I was not aware you had also taken on this task - so I will probably post my implementation of exec.c (only) as well, at least for comparison. 

That said, I've already reviewed your changes to `chmem` and see the header version is only updated to v1 if heap/stack are separately specified - another good idea I hadn't thought of.
I don't see a linker script update with your PR, are you thinking of updating the existing script in the same way as `chmem`, only setting the header version =1 when heap or stack are specified?

PR #645, opened 2020-05-22
[build] [kernel] [elkscmd] new a.out version with new chmem semantics

tkchia:
See https://github.com/jbruchon/elks/pull/644#issuecomment-632415153 .


ghaerr:
Hello @tkchia,

Very nice work. I especially like the fixes to `chmem` so that heap and stack can be specified in separate invocations, and the good idea that a header remain version 0 unless heap or stack have been specified.

Also very nice catch in thinking to add the passed argp/envp data length to the stack allocation for both v0 and v1 headers. The prior method of handing this could be considered a bug, and certainly would have issues with our plans to run with very small (~2k) stack sizes!

It seems to be the case that this benefit of extending the stack allocation for passed argp/envp cannot be achieved for CONFIG_EXEC_LOW_STACK headers because of the fixed stack/data segment start, since the data segment by default doesn't have any means of relocation, even with MMODEL, right?

On lines 269-270, I think it too dangerous for a version 0 header w/non-zero chmem to be given a minstack of 0; there are just too many cases of `malloc` getting very close to or exceeding the process heap break limit, and a process could easily make a single procedure call with a stack allocation that would overwrite the heap. Thus, and also given that v0 processes have been running with an effective t_minstack of 4k-8k bytes for a while, I think line 269 should be changed to 'stack = INIT_STACK' (to change t_minstack later). We can then assume that the kernel runs all processes with a known (or explicitly specified) minstack. 

The only other option I can think of here would be to increase hdr.chmem another INIT_STACK bytes if possible, and run with the extra amount allocated (or INIT_STACK, whichever is greater). In general, running v0 applications with more memory is always better than with less, and if there are memory hog issues, the application can be changed with `chmem`. This 2nd option when a less than INIT_STACK allocation can be granted might want to wait for the "truncation" enhancements below, so that the kernel isn't adjusting items without telling the user.

> I have been working on the implementation of the kernel heap/stack myself, and have almost finished an implementation

If your (this) version has been tested, (mine has not), I suggest we commit this version after the above change and run with it; I hadn't finished my version anyways.

> I'm thinking that, at least for now, the kernel might want to truncate the (specified or default) heap value by the amount needed to run, and issue a printk warning the user. 

Lets run with this version first, an enhanced version with more specialized handling of max heap and/or large argument length can come after tuning several important applications.

> I don't see a linker script update with your PR, are you thinking of updating the existing script in the same way as chmem, only setting the header version =1 when heap or stack are specified?

I'm assuming that after this commit, the existing use of -maout-chmem in various Makefiles will be deprecated, then changed to explicit -maout-heap= and -maout-stack= values, to work with a linker script in another commit that behaves much like `chmem`?

tkchia:
Hello @ghaerr,

> It seems to be the case that this benefit of extending the stack allocation for passed argp/envp cannot be achieved for CONFIG_EXEC_LOW_STACK headers because of the fixed stack/data segment start, since the data segment by default doesn't have any means of relocation, even with MMODEL, right?

Yes, unfortunately that seems to be the case.

> If your (this) version has been tested, (mine has not), I suggest we commit this version after the above change and run with it; I hadn't finished my version anyways.

Thanks.  I would like to have a look at your code too.  Meanwhile I think I might need to keep this PR open a bit longer, since it seems I might need to make a few more changes to the compiler toolchain, and I know that rebuilding the toolchain takes a long time.  (This itself is another problem that probably needs to be solved...)

Thank you!

ghaerr (commented):
Hello @tkchia,

Thank you for your hard work on this. These further changes from my first commit with the compiler and chmem supporting separate heap and stack are much needed!

I am fine with the code changes in this commit; but feel more discussion is needed:

I have gained more perspective looking at this from the perspective of a compiler toolchain maintainer as well, seeing the beneficial change for other projects, of being able to produce both "compatible" version 0 executables, and for ELKS, our new v1 executable.

However, I find myself thinking that we're somewhat back on the fence again, when it comes to what path ELKS is going to take forward: continuing to produce and maintain version 0 executables would be a good idea, except for those with non-zero chmem. The design of the legacy chmem field is such that good interpretation of it going forward in an improved ELKS execution environment seems to be not possible.

I agree with our decision to have ELKS run both legacy v0 executables, or "new" v1 executables, at least for a period of time, or when a legacy config option is set. I can see from a compiler-writer's perspective that it is probably a good idea to continue to support v0 executables in general. I see a problem now, given that the default executable format for ELKS remains v0, that we can't ever remove kernel v0 support - and thus also have to live with its chmem field. The current approach will not allow a config option to turn off legacy support in ELKS. It was originally thought that only bcc/ld86/dev86 would produce these executables, but with the current design, our main compiler toolchain produces them by default. I am not sure what the value is in having to continue v0 support in ELKS forever, and don't think we should force a toolchain revision when its desired to turn that off.

A possible solution to all this would be to always produce v1 executables when the compiler is invoked with the -melks-libc flag (not a great option, is there another?). Such an option would allow for v0 compiler support, but force all new executables for ELKS to be v1.

Summary: I think it important that ELKS be released from "must-have" support for legacy executables, to "can support if desired". This is because of the problems of running v0 executables with nonzero chmem, as well as getting away from kernel cruft.

On another topic: ELKS should have a standard way to specify "max heap" for (v1) executables from the compiler command line. With your changes, `chmem -h 0xffff` will always work, by truncating the heap to fit. Does the compiler support -maout-heap=0xffff to do the same?

Thank you!

ghaerr:
Hello @tkchia,

> A possible solution to all this would be to always produce v1 executables when the compiler is invoked with the -melks-libc flag (not a great option, is there another?). Such an option would allow for v0 compiler support, but force all new executables for ELKS to be v1.

Another idea, providing you agree with it, would be to use the elks-small.ld and medium linker scripts to always write the header as version 1. This decouples the version number from the compiler toolchain, since it now can produce both, and adding any new -melks-\* ELKS-specific compiler option is probably a bad idea. Using the linker script(s) to do this, if possible, would allow ELKS to have a config option for or remove v0 support in the future, when desired.

tkchia:
Hello @ghaerr,

You have a point.  I have modified the compiler toolchain to try to create v1 `a.out` programs, except when a `-maout-total=` or `-maout-chmem=` value is expliictly specified.  I have also updated the `chmem` utility to always write out a v1 header.

Eventually, I think we can get the kernel to only support v0 programs with `mh.chmem` == 0, or to remove support for v0 programs altogether.

Thank you!

ghaerr (approved):
Hello @tkchia,

Thank you, it's a pleasure working with you!

ghaerr:
Hello @tkchia,

When running ./build.sh, or tools/build.sh, the cross tools build is failing on one of my machines because of the inability to ftp or otherwise transfer the various .zip files from their servers. This has now happened 4 times (on my iMac, my laptop build is working). 

Do you have any ideas for how to proceed without having to start from scratch each time? I am playing around with tools/Makefile and commenting out various "rm -rf ..." src dirs in an attempt to get the build to start/succeed.

I'm not sure how much work it would be, but given the problem of FTP servers refusing connections, perhaps the build Makefile should try to get all the required files first, in a way that would allow running the build script multiple times on file transfer failures. Or at least have a Makefile entry point that would allow for getting the files, without removing the previous (same) ones.

Screen Shot 2020-05-25 at 10 13 36 AM

Thank you!

ghaerr:
It seems to be consistently dying in the "contrib/download_prerequisites" Makefile.
I was able to get the build to start (and will likely now succeed) by making the following changes below:
```
diff --git a/tools/Makefile b/tools/Makefile
index 96e57145..628277f5 100644
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -24,15 +24,15 @@ BINUTILS_DIST=binutils-ia16-$(BINUTILS_VER)
 
 $(DISTDIR)/$(BINUTILS_DIST).zip:
        mkdir -p $(DISTDIR)
-       rm -rf $(DISTDIR)/binutils-ia16-*.zip $(DISTDIR)/binutils-ia16-*.zip.tmp
+       #rm -rf $(DISTDIR)/binutils-ia16-*.zip $(DISTDIR)/binutils-ia16-*.zip.tmp
        cd $(DISTDIR) && wget https://github.com/tkchia/binutils-ia16/archive/$(BINUTILS_VER).zip -O $(BINUTILS_DIST).zip.tmp
        cd $(DISTDIR) && mv $(BINUTILS_DIST).zip.tmp $(BINUTILS_DIST).zip
 
 $(BUILDDIR)/.binutils.src: $(DISTDIR)/$(BINUTILS_DIST).zip
        mkdir -p $(BUILDDIR)
-       rm -rf $(BUILDDIR)/$(BINUTILS_DIST)
-       cd $(BUILDDIR) && unzip -q $(DISTDIR)/$(BINUTILS_DIST).zip
-       rm -rf $(BUILDDIR)/binutils-src
+       #rm -rf $(BUILDDIR)/$(BINUTILS_DIST)
+       #cd $(BUILDDIR) && unzip -q $(DISTDIR)/$(BINUTILS_DIST).zip
+       #rm -rf $(BUILDDIR)/binutils-src
        cd $(BUILDDIR) && mv $(BINUTILS_DIST) binutils-src
        touch $(BUILDDIR)/.binutils.src
 
@@ -56,16 +56,16 @@ GCC_DIST=gcc-ia16-$(GCC_VER)
 
 $(DISTDIR)/$(GCC_DIST).zip:
        mkdir -p $(DISTDIR)
-       rm -rf $(DISTDIR)/gcc-ia16-*.zip $(DISTDIR)/gcc-ia16-*.zip.tmp
+       #rm -rf $(DISTDIR)/gcc-ia16-*.zip $(DISTDIR)/gcc-ia16-*.zip.tmp
        cd $(DISTDIR) && wget https://github.com/tkchia/gcc-ia16/archive/$(GCC_VER).zip -O $(GCC_DIST).zip.tmp
        cd $(DISTDIR) && mv $(GCC_DIST).zip.tmp $(GCC_DIST).zip
 
 $(BUILDDIR)/.gcc.src: $(DISTDIR)/$(GCC_DIST).zip
        mkdir -p $(BUILDDIR)
-       rm -rf $(BUILDDIR)/$(GCC_DIST)
-       cd $(BUILDDIR) && unzip -q $(DISTDIR)/$(GCC_DIST).zip
-       rm -rf $(BUILDDIR)/gcc-src
-       cd $(BUILDDIR) && mv $(GCC_DIST) gcc-src
+       #rm -rf $(BUILDDIR)/$(GCC_DIST)
+       #cd $(BUILDDIR) && unzip -q $(DISTDIR)/$(GCC_DIST).zip
+       #rm -rf $(BUILDDIR)/gcc-src
+       #cd $(BUILDDIR) && mv $(GCC_DIST) gcc-src
        cd $(BUILDDIR)/gcc-src && contrib/download_prerequisites
        touch $(BUILDDIR)/.gcc.src
```
then commenting out various sections of received files in the download_prerequisites file, based on what had previously been received.
```
#wget ftp://gcc.gnu.org/pub/gcc/infrastructure/$MPFR.tar.bz2 || exit 1
#tar xjf $MPFR.tar.bz2 || exit 1
#ln -sf $MPFR mpfr || exit 1

#wget ftp://gcc.gnu.org/pub/gcc/infrastructure/$GMP.tar.bz2 || exit 1
#tar xjf $GMP.tar.bz2  || exit 1
#ln -sf $GMP gmp || exit 1

#wget ftp://gcc.gnu.org/pub/gcc/infrastructure/$MPC.tar.gz || exit 1
#tar xzf $MPC.tar.gz || exit 1
#ln -sf $MPC mpc || exit 1
```

The root problem is the connection refused rejections, I'm not sure why that is, as it seems to be happening consistently on my iMac machine now.

ghaerr:
Hello @tkchia,

I got the system compiled with the updated compiler.

I thought to test the new -maout-heap= option, and started with kilo.

Following is directly after the current build, with kilo still being compiled with -maout-chmem=0xfff0. Running `chmem` on it, notice it's a v0 executable (correct) and that the heap is displayed as 60736, and the total memory 65520. This is correct, since the v0 executables don't really account for a known stack size. Then, I run `chmem -h 0xfff0` on it, to check the max heap option. This converts the executable to v1, and now shows a smaller heap of 56640, with a total data size of 61424. This is also correct, since the chmem is subtracting the default stack allocation of 4096, which when adding together with data+bss, heap and stack = 65520.
Screen Shot 2020-05-25 at 11 49 19 AM
So far so good. I then replaced the kilo Makefile's -maout-chmem=0xfff0 option with -maout-heap=0xfff0, which should have the same effect as the `chmem -f 0xfff0` option above. The result is that the heap value is set directly to 65520, and the kernel silently rejects the load. (The shell, given a -ENOEXEC return, tries to run it as a shell script, but that's a problem for another day):
Screen Shot 2020-05-25 at 11 51 23 AM
The bug is that either the linker script or the toolchain must check the data+bss and default (or specified) stack values and possibly truncate the heap, when the -maout-heap= option is specified. The code should use either the default stack value, or the passed stack value, in the calculations. (`chmem` has access to the INIT_STACK value in minix.h when compiled, I think we're ok with the compiler or linker using a constant of 4096, as the initial stack values are likely to get smaller, not larger. The workaround for this would be always specifying a stack when specifying a heap).


I think it is correct that the kernel silently not run the executable, as the new approach relies on tools (either chmem or the compiler & linker scripts), to always produce runnable executables or error during the link, which is why the current behavior is a bug. Note that in the `chmem -h 0xfff0` above, chmem works but issues a warning that the requested heap size has been truncated. I don't think it matters whether the linker emits a warning or not.

I'm sorry, I had asked about this before committing, but for some reason this slipped through the cracks. I will be looking at fine-tuning other programs in the next few days, and changing all other -maout-chmem= to heap and/or stack specifications, so that all executables are built with v1 headers.

Thank you!

ghaerr:
Hello (again) @tkchia,

I'm afraid getting our new heap/stack/v1 changes just right seems more complicated every time I look deeper into it. Consider the following (for simplicity lets take v1 headers only):

> I think it is correct that the kernel silently not run the executable, 

> the new approach relies on tools (either chmem or the compiler & linker scripts), to always produce runnable executables or error during the link, which is why the current behavior is a bug.

Both of the above statements used to be true, but, given the (last minute but possibly still good idea) of adding a variable-sized argp/envp into a near-full data segment, they may both be false.

> The bug is that either the linker script or the toolchain must check the data+bss and default (or specified) stack values and possibly truncate the heap, when the -maout-heap= option is specified.

I still think this to be a bug, but when max heap is specified and truncation performed, the kernel may still not be able to run the executable because any 'slen' > 15 will overflow 0xffff in the add-carry calculations.

Looking hard at the source code shows still another bug: 'slen' is now added into the required data segment size in addition to the stack size, and later code in sys_exec (line 389) then sets t_begstack = t_endseg - slen. That is, the stack starts at the end of the data segment less the argp/envp data. However, the sys_brk stack size checking code in malloc.c (line 227) checks the stack length against t_endseg - t_minstack, which includes slen! This would have to be changed to t_begstack - t_minstack to be correct. (I don't want to change that code yet).

At this point, it is important to maintain stability in the kernel, since there is lots of other development taking place. We also need to get the non-slen v1 heap/stack sorted out now so that important tuning can start (which could possibly illuminate other considerations on the topic).  

SO - this is what I suggest:
- Temporarily comment out (do not remove), the two "add_overflow(len, slen...)" calls in exec.c
- Fix the heap truncation issue/"bug" pointed out in my prior comment using a linker script, NOT in the compiler source. A linker script gives far greater flexibility, IMO needed at this point.

The reason for asking for a linker script rather than additional compiler change is that changes to the kernel required for handling max data segment would most easily be made and tested using chmem (at first), and the linker script (for the build process). Requiring a toolchain rebuild during the process of getting v1 heap/stack/slen etc worked out will be a lot more work.

If you could build/modify a linker script that would perform the heap truncation using the algorithm suggested in the last comment, I will take care of updating it along with the subsequent kernel modifications required during finer tuning.

After getting the new heap/stack working and tested using various values (max heap, no heap, small stack, small heap+small stack), the 'slen' issue can be revisited. The nature of the slen issue is that in general, 'slen' is usually quite small, so it won't matter to be taken from the stack for now. It becomes an issue with max heap/small stack, and of course when slen is large, and I would like to see what the tuning brings before diving back into its changes and ramifications. The kernel will ultimately have to interpret some heap/stack values specially with large slen, I think.

Thank you!

tkchia:
Hello @ghaerr,

Thanks for your feedback!  I am starting to suspect that maybe we are trying to make the kernel, the linker script, and `chmem` a bit too smart here...

To support things like text editors that want as much heap space as possible, maybe we just need to allow a program to explicitly say "I want lots of heap space".  One way would be to reserve certain values of `mh.chmem` --- say, all values from `0xfff0` up --- to mean "maximum heap", and leave it at that.  So if a program has `mh.chmem` == `0xffef`, the kernel can take it to mean that the program really wants `0xffef` heap bytes.  I think this should simplify things a good lot.

Thank you!

ghaerr:
Hi @tkchia,

> to explicitly say "I want lots of heap space". One way would be to reserve certain values of mh.chmem --- say, all values from 0xfff0 up --- to mean "maximum heap", and leave it at that.

I had considered the special value of 0xffff (or >= 0xfff0 which is probably easier at this point), to mean just that, but was hanging on to the idea that chmem and the compiler should attempt to pre-calculate what heap values will work and give an error at compilation or chmem-running time.

For instance, max heap aside, using -maout-heap=0xffef will produce an executable that won't run, since the default stack will overflow the data segment. `chmem -h -0xffef` now produces an executable with a truncated heap that will run. I had designed chmem to be user-friendly - that is it would produce a header that would run by calculating the heap allowed rather than have to have the user try to do hex or heap arithmetic by trial and error. In the same thought process, I figured -maout-heap= should operate the same way.

Are you suggesting just having the compiler stick in any passed heap and stack value into the executable? Would chmem operate that same way, or remain as-is? Either way, the kernel will have to handle heap set to 0xfff0 specially (but not 0xffef)... and we're back to having the kernel changing passed header values in order to run... something my original design was trying to avoid.

I wanted to get to some real-use cases before making the decision of how much kernel "involvement" should be needed, if any. That is why I made the earlier suggestion of removing slen processing for now and using a linker script that would allow for the kernel to run executables without having to specially subtract passed values, which will be required for your idea of max heap, and also, I'm afraid, later for slen. All said - I don't know yet which is the better path.

Thank you!

ghaerr:
Hello @tkchia,

> I am starting to suspect that maybe we are trying to make the kernel, the linker script, and chmem a bit too smart here...

Perhaps so. If you think it better that the compiler/linker script and chmem stay out of the business of figuring everything out beforehand, then the kernel could do it, with a minimum of changes, and possibly added on to your PR #647.

I propose the following loading algorithm for v1 and v0 executables, which I wrote trying to keep as closely matched to what we already have coded in exec.c. This method uses heap values at or over 0xfff0 as specifying max heap, and a passed heap value would not be truncated by the compiler or chmem, other than basic checking. This also clarifies the sys_exec return values, suggested in https://github.com/jbruchon/elks/pull/647#issuecomment-634147248.

These algorithms also more properly handle adding slen in for both v1 and v0. In addition, a stack of INIT_STACK is always required for v0 executables.

```
v1:
total = (header.heap >= 0xfff0)? 0xfff0: 
    header.dseg +
    header.bss +
    (header.minstack ? header.minstack : INIT_STACK) +
    slen +
    (header.heap ? header.heap : INIT_HEAP);

v1 loading algorithm:
min_size = header.dseg
if add_overflow(min_size, header.bss) return -EINVAL // Invalid executable
total = min_size
stack = header.minstack? head.minstack: INIT_STACK
if add_overflow(total, stack) return -ERANGE        // Can't fit required stack
if add_overflow(total, slen) return -E2BIG          // Argument list too long
if header.heap >= 0xfff0                            // Max heap specified
    total = 0xfff0
else 
    heap = header.heap? header.heap: INIT_HEAP
    if add_overflow(total, heap) return -ERANGE     // Can't fit required heap
endif

v0 loading algorithm:
min_size = header.dseg
if add_overflow(min_size, header.bss) return -EINVAL // Invalid executable
stack = INIT_STACK
if header.chmem == 0
    total = min_size
    if add_overflow(total, INIT_HEAP+INIT_STACK) return -ERANGE // Can't fit heap+stack
    if add_overflow(total, slen) return -E2BIG      // Argument list too long
else
    total = header.chmem
    if total < min_len return -ERANGE               // Bad chmem
    if header.chmem < 0xfff0                        // Max heap not requested
        haveextra = total - min_len                 
        if haveextra < INIT_STACK                   // try allocating more for stack
            if add_overflow(total, INIT_STACK-havextra) return -ERANGE
        if add_overflow(total, slen) return -E2BIG  // Argument list too long
    endif
endif
```

If you agree, and would like to write it, go ahead. Otherwise, I can implement it after #647 is committed.

Thank you!

PR #647, opened 2020-05-26
[kernel] make execve syscall return more useful error codes

tkchia:
...such as `-EINVAL` and `-ENOMEM`, when processing a.out headers, not always `-ENOEXEC`.

### DIFF
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -183,7 +183,7 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 
     min_len = mh.dseg;
     if (add_overflow(min_len, mh.bseg, &min_len)) {
-	retval = -ENOMEM;
+	retval = -EINVAL;
 	goto error_exec3;
     }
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

This is a definite improvement, and you're beating me to it with a lot of code lately :)

My only concern is whether the now-multiple meanings of -ENOMEM is proper, or whether another error should be used: sys_exec will also return -ENOMEM when there is no system memory to load the executable, which might be useful for the caller to know. The newly-introduced -ENOMEM may want to be -EINVAL or something else, if the executable can't run because of 64k limitations. Those cases probably shouldn't be -ENOMEM, because the system can't ever run them.

Thank you!

ghaerr:
If slen processing is going to stay in, its overflow should probably return -E2BIG (Argument list too long), distinct from other errors. That would allow the shell to let users know (`sash` already does).

Executables that can't run because of improper data+bss values (other than default heap and default stack), that can't load should return -EINVAL.

Rejections caused by too much heap/stack, etc but not essentially malformed headers, could return -EFBIG (file too large) or perhaps -ERANGE (math result not representable).

ghaerr (approved):
Thank you!

PR #649, opened 2020-05-29
[kernel] Implement max heap for v1 header executables

ghaerr:
Implements max heap specified through new compiler option -maout-heap=0xffff.
See https://github.com/jbruchon/elks/pull/645#issuecomment-634382907 for algorithm.

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -225,8 +225,8 @@ int sys_brk(__pptr newbrk)
 
     if (currentp->t_begstack > currentp->t_endbrk) {				/* stack above heap?*/
         if (newbrk > currentp->t_begstack - currentp->t_minstack) {
-			printk("sys_brk(%d) fail: brk %x over by %d bytes\n",
-				currentp->pid, newbrk, newbrk - (currentp->t_endseg - currentp->t_minstack));
+			printk("sys_brk(%d) fail: brk %x over by %u bytes\n",
+				currentp->pid, newbrk, newbrk - (currentp->t_begstack - currentp->t_minstack));
             return -ENOMEM;
 		}
     }
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -224,7 +224,7 @@ int sys_brk(__pptr newbrk)
         return -ENOMEM;
 
     if (currentp->t_begstack > currentp->t_endbrk) {				/* stack above heap?*/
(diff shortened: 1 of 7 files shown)


ghaerr:
`sys_brk` now calculates stack and args/environment properly; this fixes "echo \*" stack overflow.

Fixes `chmem` close fd bug, "cd /bin; chmem \*" now works. (This works all the time with `sash`, `ash` runs out of memory, on second try, will be fixed with additional heap in tuning step next).

v0 header w/chmem field now allocates more space if required to get environment and full default stack.

ghaerr:
Latest commit corrects `chmem` to include heap+stack in TOTDATA.
Document chmem and output fields in chmem.c
Fix `chmem -h 0xffff` for max heap
Reproportion `sh` 8k heap+stack to 6k heap, 2k stack: fixes "chmem \*" mentioned above.
Add EFBIG and EINVAL shell exec error strings
Reserve space for environment when truncating chmem specified heap

tkchia (commented):
Hello @ghaerr,

Thanks for your work!  Some quick comments...

Lines 262--263 in `exec.c`: it might be good to also check for the case where `len` is just about to overflow, but not yet overflowed.  I think something like this will work:
```
if (heap >= 0xFFF0) {
    if (len < 0xFFF0)
        len = 0xFFF0;
}
```

Also, the logic at lines 281---288 seems a bit convoluted, and I am not sure I understand what is going on there.  Perhaps it can be simplified?

Thank you!

ghaerr:
Hello @tkchia,

> the logic at lines 281---288 seems a bit convoluted, and I am not sure I understand what is going on there.

That section of code handles the rare case of a legacy (v0) header specifying less than max heap, and specifying less than what is required for a full stack and the passed environment. It computes 'haveextra', which is the actual amount of (heap) space available for the executable less stack and environment, and checks to make sure the environment and a (full-sized) INIT_STACK will fit within it (stack and environment are allocated from within the passed heap specification in v0). If not, allocation of more memory than requested is attempted, rather than the previous approach of running with less than an INIT_STACK. The previous version also did not consider the environment size in this case.

Thank you!

ghaerr:
Hello @tkchia,

This code can be made simpler by just rejecting a v0 executable that doesn't specify enough memory for an INIT_STACK and the environment. I felt it a bad idea to allow a process to run with less than the standard stack, and then added code to allocate more memory than specified. It's not really the kernel's business to do that. A more proper approach would be to reject both cases.

Thank you!

tkchia (approved):
Thank you @ghaerr!

PR #651, opened 2020-05-31
[elkscmd] Fine tune application heap & stack usage

ghaerr:
Now that the ELKS kernel supports separately specified heap and stack sizes, fine tuning of ELKS applications can begin in earnest. This PR reduces the size of all programs started at boot, allowing for more memory for normal user programs. These include `init`, `getty`, `httpd`, `telnetd`, `ktcp` and `sh`.  The savings has been in the range of 2k-6k each. Although this doesn't sound like much, this was obtained with no changes to the program sources and adds up.

`bc`, `cron` and `sort` were also updated.

There may be some issues with smaller stack sizes, even after analyzing the source code of each to determine tuning sizes. On occasion, `chmem /bin/*` is producing a stack overflow message; reasons unknown. For some reason, `ps` is reporting different FREE sizes for `sh` on serial and console logins. 

Before tune:
old

After tune:
new

### DIFF
--- a/elks/arch/i86/kernel/process.c
+++ b/elks/arch/i86/kernel/process.c
@@ -43,14 +43,23 @@ void stack_check(void)
     register __ptask currentp = current;
     register char *end = (char *)currentp->t_endbrk;
 
-    if (currentp->t_begstack > currentp->t_enddata) {
-	if (currentp->t_regs.sp > (__u16)end)
-	    return;
-    }
-    else {
+#ifdef CONFIG_EXEC_LOW_STACK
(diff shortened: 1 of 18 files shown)


ghaerr:
Latest commit adds stack over min stack check kernel printk for stack tuning
Fix `sh` stack overflow on "echo /bin/*", needs 5000 byte stack
Lower `sash` internal cmd requirements: cp, dd, tar, needs 3k stack. Cmdlen 128, maxargs 256
Fix sbrk no memory return in execvp
Return E2BIG in execvp and execve on sbrk failure

PR #653, opened 2020-06-01
New version of kilo and updated documentation. Adding support for the insert key to xt_key.c driver and keymaps

georgp24:


georgp24:
I added the following features to the Kilo editor:

A Help page
Toggle Insert/Replace mode
Save file as ...
Open new file without exiting Kilo
List of recently opened files
Find string and replace
Bookmarks support
Remove current line
Copy and paste a block of text
Write copied block to a text file
Insert a file from disk
Move to Top/End of file
Simple Undo function
Add syntax highlighting for python, go and javascript

ghaerr:
Georg,

There are a number of problems with this PR that prohibit it from being committed as-is.


Problems:
- In adding support for the Insert key, you used 0176 (=0x7e) in the keys\*.h files as the low-level value to pass up to the keyboard driver. This is the ~ key, are you aware of that? So your design removes the ~ key and changes it to Insert. A value greater than 0200 (=0x80) must be used. I would advise you build a map (in octal) of the keys\*.h values before proceeding further. This would be good documentation, as you can't just pick any value, and we need to know that the keyboard continues to work for all countries.
- In addition, in the german keys-de.h unsigned char array, you changed both 0267 and 0261 to 267 and 261. This can't possibly work, as those values are greater than 255, and break the previously working keymap. It appears completely untested. I suspect it produced an unheeded compiler warning as well.

In the many kilo enhancements, there is still a lot of cruft that should be removed. For instance "char test[64];" which does nothing, various stray printfs about malloc failure, and your original CheckAvailableMemory function. These should all be removed. If you want to show malloc failures, those printfs should display in the kilo status bar. Frankly, with as much original and new code calling malloc and not checking for NULL, the malloc function should probably be replaced with a ckmalloc() that writes failure to the status bar at least, as the program will surely crash and lose the user's edits the way it is now on malloc failure.

How well have you tested the large number of changes to kilo? I worry a bit about it's stability, especially given its original malloc issues and the large changes all at once. How does one quit the editor if the file is not saved? I see the three ^Q mechanism has been changed.

Ideally, the kernel keyboard driver and kilo changes should be separated into multiple PRs, in the case they have to be backed out.

georgp24:
I will use a different octal number for the insert key then. However, I do not understand the pattern used for the octal numbers in the keymap. Can I just take an unused octal number such as 258 or 260?

georgp24:
I can remove the parts of the code you mentioned, however, kilo fails as in the screenshot below when you load e.g. kilo.c (with 2k stack):
kilo-sys_brk1
So there is no chance to write anything to the status bar.

I did not add any new malloc or realloc calls, these were all present in the first commit. And all malloc and realloc calls did not test for the result! I started to add these tests hoping I could terminate loading a file with a "file too big" message. 

Yes, Kilo requires a lot of memory, but that's not because I added code. Kilo stores each line of a file as the original text, a screen representation expanding the TABs and a line containing the highlighting for each character. So this requires three times the memory as the size of the file on disk.

I tested each feature I added until it worked as expected. Of cause a different user will notice things which I did not observe.

ghaerr:
> I can remove the parts of the code you mentioned, however, kilo fails as in the screenshot below when you load e.g. kilo.c (with 2k stack):

Georg, removing the test code and finding that kilo fails with no memory are two different things.
I recommend removing test code so that we don't have useless cruft in a program source.
With regards to the kilo not loading files, read on.

> Yes, Kilo requires a lot of memory, but that's not because I added code. Kilo stores each line of a file as the original text, a screen representation expanding the TABs and a line containing the highlighting for each character. So this requires three times the memory as the size of the file on disk.

Agreed, Georg. I am not saying that your code lowers the size of files kilo can load. Thanks for explaining how much memory kilo uses per line of text. Its unfortunate it uses 3x memory, but that's the way it is.

> I did not add any new malloc or realloc calls, these were all present in the first commit. And all malloc and realloc calls did not test for the result! 

Yes, I saw that the original code sometimes did not error check. All will have to be checked in order for kilo to be considered a safe editor, so I am in favor of that. I was just suggesting that the printfs be changed to display in the status bar, or perhaps program exit.

> I started to add these tests hoping I could terminate loading a file with a "file too big" message.

Another possible way of quickly learning the amount of memory kilo can use would be to do a "char *heap = sbrk(0);" at startup. This will give you a number which is the start heap address. The amount of memory available is approximately "65520 - (int)heap - stack". The stack is set by default to 4096, but can be changed with a -maout-stack=2048, for instance. If you do that calculation, then divide by 3, that should give you a very approximate value for the size of file that kilo can edit. Try it. You may need to subtract another 256 bytes or so for the environment and argument strings. You could test for that file size or larger and issue a warning or not load the file.

> So there is no chance to write anything to the status bar.

I see, the kernel printk() is disrupting the screen. We may need to devise a way to turn that off so that kilo functions well. You can comment that out for your testing while we think about how to best turn the message off.

Thanks!

ghaerr:
> However, I do not understand the pattern used for the octal numbers in the keymap. Can I just take an unused octal number such as 258 or 260?

Georg, I don't understand the octal number use either. It is a bad idea to program not knowing what is going on. The keymap char[] arrays are a map, that map a scan code from 0-255 to a keyboard return character, or octal value. What is needed is for someone to produce (document) the scan code map, and show the relationship between each scan code and its mapped value, including the octal values. In that way, we can learn what the octal values are being used for now. I would do this myself, but my plate is full. Since you are designing the change, I suggest that you do it.

Once we can see the mapping of scan codes to keyboard (and octal) values, we could then use \#define values instead of octal for new meanings, it would be so much clearer. But it can be done without defines, once we see which octal values are used for which scan codes.

georgp24:
I found the original octal codes in the minix1 tty.c code. There they are listed somewhat sequencially, so it may explain it a bit: [minix1 tty.c](https://github.com/gdevic/minix1/blob/3475e7ed91a3ff3f8862b2543059580a01c1b1a7/kernel/tty.c)
Here is a screenshot of this file:
Minix1-keycodes
I put the octal codes from the ELKS keymap file into the following list of scan codes using red color:
ELKS-keycodes
Here is the source of this list [minix2 list](https://raw.githubusercontent.com/chenshuo/old-minix/de2d2a615df86ae4643effde5474c8b5b275d700/src/kernel/keymaps/us-std.src)

I would use the code octal 272 (0xba) for Insert which would fit into this sequence.

ghaerr:
Hello @georgp24,

Nice work!!!!! :) Thank you!

I was wondering where the ELKS keyboard code came from, thanks for finding that its all from MINIX 1.

I very much like your 2nd image, which shows the mappings of the scan codes to the (now much less mysterious) octal codes.

You should post your updated 2nd image or text file into a Documentation PR, so that the keyboard mappings stay with the ELKS source code for reference.

I will have to analyze it a bit more, but this information can now be used to see whether the xt_key.c driver is returning ANSI values for all the function/arrow/key keys it is receiving.

> I would use the code octal 272 (0xba) for Insert which would fit into this sequence.

Yes, that looks like the next unused code.

Mellvik:
Out of curiosity @ghaerr,
How come you were outright negative when I made exacly the same suggestion of using the minix 2 table a few weeks back?

-M

ghaerr:
@Mellvik,

> How come you were outright negative when I made exacly the same suggestion of using the minix 2 table a few weeks back?

I thought you were talking about rewriting the keyboard driver with a new table, which I felt would be very hard to test. What has been suggested here is to use the table for documentation on the octal codes, not rewriting the driver tables... ?

PR #654, opened 2020-06-02
Add support for insert key using octal 272 in keymap files and add AN…

georgp24:
…SI sequence in xt_key.c


ghaerr:
Thanks Georg!

ghaerr:
Hello @georgp24,

You're running direct console and the arrow keys work for you with this commit, right? @Mellvik has a system where they are not working with direct console. I just want to clarify they are working for you so we can determine a plan for the issue with his system. I'm not yet sure whether the scan codes are different, or whether it has something to do with the E0 prefix.

georgp24:
Yes, I use direct console on qemu with the german keyboard driver and the arrow keys do work.

ghaerr:
Hello @mellvik,

So it seems both Georg and I use direct console with QEMU, but real hardware arrow keys don't work, at least certainly not on your system. I have been studying xt_key.c today and it's kind of a mess... I was thinking of cleaning it up, but don't want to start if you have a fix for your arrow keys already. Do you have a fix for the arrow keys, or were you saying that xt_key.c needs a rewrite in order to handle them? I also noticed the problems you previously mentioned regarding caps lock etc, which are half-way handled. A lot of this can be cleaned up considerably, without having to go through the problem of introducing entirely new scan code tables for 8 countries.

With the octal mappings now understood, I also realize that some (octal) key codes are possibly being ignored. In addition, we now have hex and octal being used for the same key codes, which is quite confusing. I was thinking it might be nice to have a mechanism that would show when a key was pressed that didn't generate a keycode (which I think is your case). I was thinking of taking a pass at a cleanup that would allow the driver to be more easily understood, but would like to learn about your fix beforehand if you have one.

PR #656, opened 2020-06-04
Add documentation for the keymaps header files

georgp24:
Providing documentation for the octal codes in the keymaps header files.

### DIFF
--- a/Documentation/Readme.md
+++ b/Documentation/Readme.md
@@ -79,6 +79,8 @@ This section contains some HOWTOs to help using ELKS.
 
 * [Makefile structure](https://htmlpreview.github.io/?https://github.com/jbruchon/elks/blob/master/Documentation/html/technical/Makefile-structure.html)
 
+* [Keymaps for the keyboard driver](https://htmlpreview.github.io/?https://github.com/jbruchon/elks/blob/master/Documentation/html/technical/keymaps.html)
+
 * [Explanations on the spanish keymap](https://raw.githubusercontent.com/jbruchon/elks/master/Documentation/text/keymap.es.txt)
 
 * [Implementation of module code](https://raw.githubusercontent.com/jbruchon/elks/master/Documentation/text/modules.txt)
--- a/Documentation/html/technical/keymaps.html
+++ b/Documentation/html/technical/keymaps.html
@@ -1,20 +1,22 @@
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">
 <html>
 <head>
-	<meta http-equiv="content-type" content="text/html; charset=windows-1252"/>
+	<meta http-equiv="content-type" content="text/html; charset=utf-8"/>
 	<title></title>
-	<meta name="generator" content="LibreOffice 5.2.7.2 (Windows)"/>
+	<meta name="generator" content="LibreOffice 6.3.3.2.0 (Linux)"/>
 	<meta name="author" content="Georg Potthast"/>
 	<meta name="created" content="2020-06-04T00:00:00.019382087"/>
-	<meta name="changed" content="2020-06-04T21:50:04.077000000"/>
+	<meta name="changed" content="2020-06-06T07:41:40.434116043"/>
 	<style type="text/css">
-		@page { margin: 2cm }
-		p { margin-bottom: 0.21cm }
-		a:link { so-language: zxx }
+		@page { size: 21cm 29.7cm; margin: 2cm }
+		p { margin-bottom: 0.21cm; background: transparent; line-height: 115%; background: transparent }
+		td p { background: transparent; background: transparent }
+		a:link { color: #000080; so-language: zxx; text-decoration: underline }
+		a:visited { color: #800000; so-language: zxx; text-decoration: underline }
 	</style>
 </head>
-<body lang="de-DE" dir="ltr">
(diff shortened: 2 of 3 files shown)


ghaerr:
Hi Georg, thanks for the documentation.

The documentation mentions (lines 73-74) that codes > 0200 have 0200 subtracted from them... 
I'm not sure that's true. It seems that function keys are handled specially, without looking at the per-country table at all, but instead use the never-changed `tb_state[]` array in xt_key.c. Arrow keys and other keys prefixed with E0, the same thing. Only for those keys that are otherwise specified as "simple scan code" keys are the per-country values used.

For those keys, 0200 is never subtracted. Instead, a few special-cased (home/pgup/insert, etc) keys are handled in a case statement to convert them to ANSI strings, and the rest are passed through untouched, as special per-country characters greater than 127.

For this last set of keyboard values, which are likely Latin-1, I'm not sure if we can guarantee that those per-country values are different than our 0261-0272 values we're now translating to ANSI sequences. Note, for instance, that the German keyboard file has various hard-coded 8-bit values coded as single characters in quotes, as well as some 0x80 vs 0200 values. So I'm still a bit worried about our whole approach, when a country-specific character happens to be mapped into an ANSI-mapped value.

georgp24:
What I meant is that 267 octal for the home key is converted to 0xb7 by the driver. So the 8th bit is cleared and thus 200 octal is subtracted. But I agree this is not the case for all keys, so I will rephrase that. I could mention that this is only the case with the HOME, END, INSERT, PageUP and PageDown keys. I would like to mention this because this is not obvious without studying the xt_keys.c code.

ssieb:
> What I meant is that 267 octal for the home key is converted to 0xb7 by the driver. So the 8th bit is cleared and thus 200 octal is subtracted. 

Did you write something wrong there? 0267 is the same as 0xb7.

georgp24:
Samuel,
I never use octal codes so this is confusing for me. Yes, you are right, I will correct this.
Georg

georgp24:
Hi Greg,

I will correct it and not mention that the driver deducts octal 200.

It seems that the keyboard driver does not support ASCII codes above 0x7F or 177 octal. I tried to enter German umlauts at the console prompt but the cursor did not move nor display the character. In the Kilo and vi editors the cursor did move but nothing is displayed.

So apparently the driver can use the codes 200 octal and higher for other purposes because these ASCII codes are not supported.

The German keyboard driver is handy to enter English text therefore I did not have problems with it. All the ASCII codes up to 0x7F are mapped to the correct keys. But you cannot use the German umlauts: üöäÜÖÄ and ß

Georg

ghaerr:
> I will correct it and not mention that the driver deducts octal 200.

Ok

> But you cannot use the German umlauts: üöäÜÖÄ and ß

Geez. And here I thought that ELKS worked for all the international keyboards.

> It seems that the keyboard driver does not support ASCII codes above 0x7F or 177 octal. I tried to enter German umlauts at the console prompt but the cursor did not move nor display the character. In the Kilo and vi editors the cursor did move but nothing is displayed.

I think the driver returns the values, but I haven't yet traced the 8-bit path back through `sh`, line noise, kilo and direct console output. As you know, the BIOS character generator has a fixed character set.  If it is of interest (I thought this already worked), then we first have to decide on a character set that will work for the countries supported (probably PC OEM), then verify the country keymap files and likely come up with a scheme that will allow these extra characters to work with the special values we are reserving for function keys etc, in octal.

Another approach would be to implement UTF-8, but I think that will end up being more trouble than its worth, since none of the applications support it, and we may be stuck with the PC OEM charset. Since we're just now realizing that non-ASCII characters don't work, I wonder about its usefulness, although I think the current keyboard driver should probably be debugged to allow 8-bit passthrough.

As you may have seen in my PR #657, it seems that the `tb_state[]` array is overriding the country keyboard files, and that will also hinder using country codepage table entries from working properly.

ghaerr:
@georgp24,

Here is some discussion of the PC OEM character set (code page 437) https://en.wikipedia.org/wiki/Code_page_437. The "Internationalization" section discusses the characters included and excluded for various european countries. Those international characters are from 0x80 through 0xAF, so this could work with our current use of higher values in the 0xC0 range for arrow and keypad keys. Codepage 437 makes sense for ELKS because it was the original ROM character set for IBM PCs.

A possible design would be to use CP 437 codepage values 0x80-0xAF in the country code tables, and define our own set of arrow/fnkey/special keys in the 0xC0 range. These would be used in the keyboard driver. For direct or BIOS console output, all characters in the 0x20-0xFF range could be used for display. Characters lower than 0x20 (space) would remain used for some control characters.

This could be tested in direct and BIOS console by writing a program that displays the above characters, and also echos characters input by the keyboard driver (except for those characters that result in special handling, including all the ANSI ESC sequence producing keys).

ghaerr:
> But you cannot use the German umlauts: üöäÜÖÄ and ß

I've been looking into this and it appears that one of my earlier PRs has broken output of all the OEM codepage 437 characters > 127. This was done when fixing a problem with outputting '\0', (the NUL character). I'll fix this in my still-open PR #657, which should allow the above to work.

The fix is in line 320 of console.c: `if (c >= ' ')` should be `if (c != 0)`. Since I missed seeing 'c' is declared as a "char" variable, all values of c > 127 will be interpreted as < 0.

PR #657, opened 2020-06-04
[kernel] Fix direct console bugs and cleanup

ghaerr:
Fixes bug in `tb_state[]` array that had arrow keys hard-coded.
Cleanup considerably.
Switch key values coded in octal to match keymap files for understandability.
Temporarily displays a message when a key is pressed which needs an ANSI sequence.

@mellvik: As I was reviewing the code mentioned in https://github.com/jbruchon/elks/pull/654#issuecomment-639106559 I noticed the hardcoded arrow keys problem above, and finally understood (and cleaned up) the code regarding E0 prefix processing. I'm pretty sure the issue with ELKS not recognizing your arrow keys is that they're being sent without an E0 prefix, whereas QEMU always does. This, combined with the arrow keys being hard coded into the state table rather than the country tables lead me to start cleaning this up and fix the issue. It should be ready for your keyboard LED enhancements when you have time.

### DIFF
--- a/elks/arch/i86/drivers/char/console.c
+++ b/elks/arch/i86/drivers/char/console.c
@@ -317,7 +317,7 @@ static void std_char(register Console * C, char c)
 #ifdef CONFIG_CONSOLE_BIOS
 	PositionCursor(C);
 #endif
-	if (c >= ' ') {
+	if (c != 0) {
 		VideoWrite(C, c);
 		C->cx++;
 	}
--- a/elks/arch/i86/drivers/char/xt_key.c
+++ b/elks/arch/i86/drivers/char/xt_key.c
@@ -36,8 +36,8 @@
 extern struct tty ttys[];
 extern void AddQueue(unsigned char Key);
 
-#define ESC 27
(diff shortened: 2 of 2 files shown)


Mellvik:
This is good.

Now, since we seem to have consensus that the Minix2 keyboard driver is a good model for an improved ELKS driver, it makes sense to start that process now instead of continuing to add more hacks to xt_key. Such as numlock, scrollock, expanded international support etc. Minix2 has all that - very well organized, and flexible as to adding more international variants.

During my work on xt_key.c a few weeks back, I found too many ‘structural’ problems originating from the fact that it was created for the ancient XT-keyboard. For example - as we add more states, either the modestate byte must be expanded, or more hacks added. The upper keyrow (numeric) should ignore capslock while act on shift - which doesn’t work. That too can easily be hacked but the hacks are likely to break when more international variants are added. etc.

SO from my point of view, replacing xt_key.c with minix2 keyboard.c & keytables is the sensible way forward. Not the least because of its apparently solid support for international codepages and layouts.

—Mellvik

PR #662, opened 2020-06-12
[ktcp] Fix ARP request and cache handling

ghaerr:
Contributed by @Mellvik. Corrects `ktcp` problem of answering any and all ARP requests. Also updates ARP cache/merge handling to RFC 826 algorithm.

@Mellik,

Thanks for the fix!

I would like to get rid of the merge_flag global in arp.c, but have left this exactly as your diff for now. As I was getting ready to make the merge_flag change to be a passed parameter, I noticed that the code in arp_recvpacket() calls arp_cache_get when merge_flag = 1, but also calls arp_cache_add when not found, and that routine calls arp_cache_get again (with an implied merge_flag = 1). I want to clarify whether both calls to arp_cache_get need merge_flag set, as then arp_cache_add would also have to be have a merge_flag parameter added. That is - when do you see the merge requirement, in the first or second call to arp_cache_get (or both)? It seems it is only needed in the first call to arp_cache_get, but since I can't test, I didn't want to change it.

Thanks!

### DIFF
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -40,7 +40,7 @@ struct arp_cache_s {
 typedef struct arp_cache_s arp_cache_t;
 
 static arp_cache_t arp_cache [ARP_CACHE_MAX];
-
+static int merge_flag = 0;     /* tell arp_cache_get to merge the incoming address */
 
 static void arp_cache_init (void)
 {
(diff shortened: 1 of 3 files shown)


Mellvik:
> I would like to get rid of the merge_flag global in arp.c, but have left this exactly as your diff for now. As I was getting ready to make the merge_flag change to be a passed parameter, I noticed that the code in arp_recvpacket() calls arp_cache_get when merge_flag = 1, but also calls arp_cache_add when not found, and that routine calls arp_cache_get again (with an implied merge_flag = 1). I want to clarify whether both calls to arp_cache_get need merge_flag set, as then arp_cache_add would also have to be have a merge_flag parameter added. That is - when do you see the merge requirement, in the first or second call to arp_cache_get (or both)? It seems it is only needed in the first call to arp_cache_get, but since I can't test, I didn't want to change it.
> 
> 
> 
Good catch, my mistake: The first arp_cache_add is a leftover - should have been deleted. As it stands, it has no effect. But before you proceed, I'll delete it and test.

The (intended) logic is as follows:
	- incoming ARP req (any dest):
		- do we have this IP in the cache?
			- yes: update Ether
	- our IP?
		- update cache (unless done above)
		- reply
There are obviously variants of this that I cannot test, so your walkthrough logically is valuable.

-M

ghaerr:
> The first arp_cache_add is a leftover - should have been deleted.

Actually that was my mistake, your diff shows it deleted.

I think the code does follow your intended logic. I'll update the code with merge_flag passed as a parameter and you can test that.

ghaerr:
@Mellvik, would you mind if I merged this? It should finalize the fix you identified and coded up. It can then be tested when you update your tree.

The Compaq 386 arrived today, network card supposed to be in tomorrow :)

Mellvik:
No, please go ahead.

I haven't tested it yes, because of the ongoing process and all the small debug mods I currently have in the code, but if anything turns up, we'll amend it to a bigger network PR real soon.

I'm working my way backwards through my changes to ne2k-mac.s in order to pinpoint exactly where the ping24 issue was. There are surprises coming...

-M

Issue #638, opened 2020-05-17
kilo editor cannot load kilo.c


georgp24:
Kilo is compiled with `-maout-chmem=0xfff0`. However, when I move the `kilo.c` file to ELKS I cannot load that in kilo, it fails with `sys_brk`. It will only work when I reduce the `kilo.c` file to 149 lines or `5343 bytes`.
I would assume `-maout-chmem=0xfff0` would allow for bigger files than 5k.

ghaerr:
Hello @georgp24,

Thank you for actually trying to use ELKS, its nice to try to get it to the point of being actually useful. That said, using ELKS usually means finding more problems, but I think we're making progress, thanks to your and everyone else's help :)

I was aware of the fact that `kilo` can't load large files, but given its data+heap allocation of 65520 bytes and small data/bss usage, it should at least be able to load files larger as large as 32k, and with some changes, ~50k+ or so.

I think this problem has to do with libc `malloc` internally using `sbrk` to allocate memory. The problem that 8086 `sbrk` has is that when passed an allocation request larger than 32767, that value is interpreted as a negative number, and `sbrk` fails, or worse, succeeds and offers "new" memory below the current heap! I will see if I can get malloc to work directly with `brk`, otherwise at least limit malloc allocations to 32k - 1 to prohibit corruption errors.

In debugging `kilo`, I noticed that `ps` shows negative "free" values when editing a blank file, and  a negative heap value when editing a larger file. This is likely due to heap and free values also being treated as signed numbers,  and there could be more to the problem. I'll take a look at that too.

The new "min_stack" coming shortly will also allow for smaller stack requirements, and a larger total heap, for cases like this where a process needs as much heap as it can get. I will consider kilo when doing the min_stack work as well.

Thank you!

georgp24:
Maybe Kilo could be compiled with the new medium memory model to have access to more memory.

I looked through the pull requests on the Kilo site, debugged and added a few of these to Kilo. This are delete line, copy line and paste line, plus a separate help page.

ghaerr:
Hello @georgp24,

I've looked into the libc memory allocation routines, and they already correctly handle the problem I thought `kilo` might be having. Looking further into kilo, it seems that instead of a single large malloc for the entire file, it malloc's each line separately, along with a number of other allocations. So it seems that kilo uses up tons of memory for small files. I'm not sure what can be done about this.

> Maybe Kilo could be compiled with the new medium memory model to have access to more memory.

[...]

PR #663, opened 2020-06-18
Update and extend the exe_format documentation file

georgp24:
As suggested by @ghaerr in #638 I updated the `exe_format.html` page to provide more information about ELKS memory allocation and executables. 
This page had been renamed quite a while ago, the old name was `bin_formats.txt`.


ghaerr:
Thanks for the documentation, Georg. 

I've been thinking about the idea of creating a Wiki page here on Github for ELKS, and contributing some documentation myself as to the more technical ways ELKS can be used. We could also point to documentation directly from there as well. What do you think of that idea?

georgp24:
It would be great if you could add some documentation. Wiki pages have the advantage that they do not need to be committed using Git and therefore are easier to write and modify.

The existing Documentation directory has information that is available locally. The Readme.md file in this directory provides access to this documentation on GitHub using a browser.

PR #667, opened 2020-07-14
[ktcp] Rewrite ARP handling to be asynchronous

ghaerr:
This fixes issue #67 which has been open for over three years.

Move ethernet routines out of IP layer.
Add `ip_route` function to centralize routing.
Create separate link layer to centralize ethernet sending and ARP processing.
Add `eth_route` function to handle ARP request/reply and single packet queuing.
Remove ARP wait kluge. ARP and IP packets all handled in main loop asynchronously.
Add single ethernet packet queuing per host to handle async ARP reply responses.
Queued packet freed when ARP reply seen, otherwise ARP request sent.

Tested on QEMU. Needs testing on real hardware!

### DIFF
--- a/elkscmd/ktcp/arp.c
+++ b/elkscmd/ktcp/arp.c
@@ -1,6 +1,8 @@
 /*
  * This file is part of the ELKS TCP/IP stack
  *
+ * 13 Jul 20 Greg Haerr - rewrote ARP handling, ARP and IP packets handled in main loop
+ *
  *	This program is free software; you can redistribute it and/or
  *	modify it under the terms of the GNU General Public License
  *	as published by the Free Software Foundation; either version
@@ -13,17 +15,17 @@
 #include <fcntl.h>
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
 #include <errno.h>
 #include <time.h>
 
 #include <linuxmt/arpa/inet.h>
 
-#include "deveth.h"
 #include "tcp.h"
 #include "tcpdev.h"
 #include "ip.h"
+#include "deveth.h"
 #include "arp.h"
-
 
 /* ARP operations */
 #define ARP_REQUEST  1
@@ -32,66 +34,75 @@
 /* Local ARP cache */
 #define ARP_CACHE_MAX 5
 
-struct arp_cache_s {
-	ipaddr_t   ip_addr;   /* IPv4 address */
-	eth_addr_t eth_addr;  /* MAC address */
-};
+static struct arp_cache arp_cache [ARP_CACHE_MAX];
 
-typedef struct arp_cache_s arp_cache_t;
-
-static arp_cache_t arp_cache [ARP_CACHE_MAX];
-
-static void arp_cache_init (void)
+int arp_init (void)
 {
-	memset (arp_cache, 0, ARP_CACHE_MAX * sizeof (arp_cache_t));
+	memset (arp_cache, 0, ARP_CACHE_MAX * sizeof (struct arp_cache));
+	return 0;
 }
 
-
-int arp_cache_get (ipaddr_t ip_addr, eth_addr_t * eth_addr, int merge)
+struct arp_cache *arp_cache_get(ipaddr_t ip_addr, eth_addr_t *eth_addr, int flags)
 {
-	register arp_cache_t * entry = arp_cache;
+	register struct arp_cache *entry = arp_cache;
 
 	/* First pair is the more recent */
 	while (entry < arp_cache + ARP_CACHE_MAX) {
 		if (!entry->ip_addr)
 		    break;
 		if (entry->ip_addr == ip_addr) {
-		    if (merge) {
+			if ((flags & ARP_VALID) && !entry->valid)
+				return NULL;	/* not yet valid - awaiting ARP reply*/
+			if (flags & ARP_UPDATE) {
 				memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
-				debug_arp("arp: updating cached entry for %s\n", in_ntoa(ip_addr));
+				debug_arp("arp: merging cached entry for %s\n", in_ntoa(ip_addr));
 			} else {
 				memcpy (eth_addr, entry->eth_addr, sizeof (eth_addr_t));
 				debug_arp("arp: using cached entry for %s\n", in_ntoa(ip_addr));
 			}
-			return 0;
+			return entry;	/* success*/
 		}
 		entry++;
 	}
 
 	debug_arp("arp: no cached entry for %s\n", in_ntoa(ip_addr));
-	return 1;
+	return NULL;			/* not found*/
 }
 
-
-void arp_cache_add (ipaddr_t ip_addr, eth_addr_t * eth_addr)
+struct arp_cache *arp_cache_update(ipaddr_t ip_addr, eth_addr_t *eth_addr)
 {
-	if (arp_cache_get (ip_addr, eth_addr, 0)) {
+	struct arp_cache *entry;
 
-		/* Shift the whole cache */
-		arp_cache_t * entry = arp_cache + ARP_CACHE_MAX - 1;
-		while (entry > arp_cache) {
-			memcpy (entry, entry - 1, sizeof (arp_cache_t));
-			entry--;
-		}
-
-		/* Set the first pair as the more recent */
-		entry->ip_addr = ip_addr;
+	if ((entry = arp_cache_get(ip_addr, eth_addr, 0))) {
 		memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
-		debug_arp("arp: adding cache entry for %s\n", in_ntoa(ip_addr));
-	} else
-		debug_arp("arp: already cache entry for %s, ignored\n", in_ntoa(ip_addr)); //FIXME
+		entry->valid = 1;
+	} else debug_arp("arp: no cached entry to update for %s\n", in_ntoa(ip_addr));
+
+	return entry;
 }
 
+struct arp_cache *arp_cache_add(ipaddr_t ip_addr, eth_addr_t *eth_addr)
+{
+	struct arp_cache *entry;
+
+	/* Shift the whole cache */
+	entry = arp_cache + ARP_CACHE_MAX - 1;
+	while (entry > arp_cache) {
+		memcpy (entry, entry - 1, sizeof (struct arp_cache));
+		entry--;
+	}
+
+	/* Set the first entry as the more recent */
+	entry->ip_addr = ip_addr;
+	if (eth_addr) {
+		memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
+		entry->valid = 1;
+	} else entry->valid = 0;	/* no MAC address yet, awaiting ARP reply*/
+	entry->qpacket = NULL;
+	debug_arp("arp: adding cache entry for %s, valid=%d\n", in_ntoa(ip_addr), entry->valid);
+
+	return entry;
+}
 
 char *mac_ntoa(unsigned char *p)
 {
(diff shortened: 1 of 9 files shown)


Mellvik:
Thank you @ghaerr, 
I guess your reference to issue #67 says it all: This is something we've been waiting for for a long time. Great work.

I'll start testing this immediately - a fresh clone of the current repository with your PR added, then starting the merge of my driver updates. I can see immediately that it does not make sense to keep my ICMP_ARP_KLUDGE. While it has been useful for debugging, it's just too complicated to make the merge worthwhile. That said, the 'valid' flag in the arp struct lends itself well to this (and possibly other) purpose. Maybe a more general name than 'valid' would be an idea?

While working with ARP and debugging over the past month, I've frequently needed a dump of the arp cache. How about adding a signal handler to ktcp that dumps the arp cache to stderr?

In deveth.c line 77, the tests for unicast/broadcast should be commented out since there is no way to put the nic into promisc mode anyway.
Also in deveth.c, eth_route(): (line 116) - is it possible under high load that the previous arp-request is still in the output buffer? I guess I'll test that on physical. If it happens we'll have another case of reentry...

Again, great contribution!

--Mellvik



14. jul. 2020 kl. 02:46 skrev Gregory Haerr <notifications@github.com>:

ghaerr:
> I can see immediately that it does not make sense to keep my ICMP_ARP_KLUDGE. While it has been useful for debugging, it's just too complicated to make the merge worthwhile.

Can you remind me again what your ICMP_ARP_KLUGE is doing? Perhaps I can get the same functionality added.


> That said, the 'valid' flag in the arp struct lends itself well to this (and possibly other) purpose. Maybe a more general name than 'valid' would be an idea?

The 'valid' flag is 1 iff the ethernet address field is valid. When 0, the arp_cache_add has been called with NULL as the ethernet address, which is used to place an IP address entry into the ARP cache, whose purpose is to hold the queued packet pointer for sending when the ARP reply is received for that IP address. What are you thinking its purpose could be expanded to?

> While working with ARP and debugging over the past month, I've frequently needed a dump of the arp cache. How about adding a signal handler to ktcp that dumps the arp cache to stderr?

The ARP cache could probably be displayed with "netstat -a" or something like that (along with other network stats like packets received/dropped/etc). Would that work?

> In deveth.c line 77, the tests for unicast/broadcast should be commented out since there is no way to put the nic into promisc mode anyway.

Ok I'll comment it out; I was wondering about that but didn't know the details about the NIC, thanks.

> Also in deveth.c, eth_route(): (line 116) - is it possible under high load that the previous arp-request is still in the output buffer? I guess I'll test that on physical. If it happens we'll have another case of reentry...

That arp_request was added at the last minute, and hasn't been tested - but arp_request uses its own stack buffer to build the request in, and calls eth_write directly, so there's no reentry buffer problems. Do you mean the card buffer? If not ready, the card should block the process in the write() call until ready.

The purpose of this additional arp_request is that if the first ARP request (and packet queueing) is lost on transmit and/or never replied to, then all subsequent packets routed to that address will not be queued (since there's one packet queued already), and no other ARP request sent. So the system will never move forward unless an ARP reply is seen for that IP address. It would be nice to test this somehow with a dropped ARP response. There is also no ARP cache timeout currently implemented, so a lost ARP reply will drop the subsequent packet and resend the queued packet in that circumstance. I considered adding timeout but since no ARP cache entries are expired, I left it alone for now.

Thank you!

Mellvik:
> I can see immediately that it does not make sense to keep my ICMP_ARP_KLUDGE. While it has been useful for debugging, it's just too complicated to make the merge worthwhile.
> 
> Can you remind me again what your ICMP_ARP_KLUGE is doing? Perhaps I can get the same functionality added.
> 
The 'kludge' picks up the mac address from an incoming ICMP which would normally trigger an ARP request and then wait. The effect is like a hard wired arp cache entry, except more flexible, and the ICMP can be responded to right away. I had envisioned to just let the arp request proceed normally in parallel, but that was too complicated.
> That said, the 'valid' flag in the arp struct lends itself well to this (and possibly other) purpose. Maybe a more general name than 'valid' would be an idea?
> 
> The 'valid' flag is 1 iff the ethernet address field is valid. When 0, the arp_cache_add has been called with NULL as the ethernet address, which is used to place an IP address entry into the ARP cache, whose purpose is to hold the queued packet pointer for sending when the ARP reply is received for that IP address. What are you thinking its purpose could be expanded to?
> 
This is the ICMP_ARP_KLUDGE again. In my debugging setup I have a similar flag (called 'status') in order to keep track of how the entry got there ... Really useful mostly for debugging.
> While working with ARP and debugging over the past month, I've frequently needed a dump of the arp cache. How about adding a signal handler to ktcp that dumps the arp cache to stderr?
> 
> The ARP cache could probably be displayed with "netstat -a" or something like that (along with other network stats like packets received/dropped/etc). Would that work?
> 
Even better … (I grew up with 'arp -a' which linux still uses, and wouldn't mind that either, along with arp -d to delete and arp -a to add hardwired entries :-) )
> Also in deveth.c, eth_route(): (line 116) - is it possible under high load that the previous arp-request is still in the output buffer? I guess I'll test that on physical. If it happens we'll have another case of reentry...
> 
> That arp_request was added at the last minute, and hasn't been tested - but arp_request uses its own stack buffer to build the request in, and calls eth_write directly, so there's no reentry buffer problems.
> 
I did mean arp_request -> eth_write, sounds like it's OK.

> Do you mean the card buffer? If not ready, the card should block the process in the write() call until ready.
> 
> The purpose of this additional arp_request is that if the first ARP request (and packet queueing) is lost on transmit and/or never replied to, then all subsequent packets routed to that address will not be queued (since there's one packet queued already), and no other ARP request sent. So the system will never move forward unless an ARP reply is seen for that IP address. It would be nice to test this somehow with a dropped ARP response. There is also no ARP cache timeout currently implemented, so a lost ARP reply will drop the subsequent packet and resend the queued packet in that circumstance. I considered adding timeout but since no ARP cache entries are expired, I left it alone for now.
> 
> 
MAkes sense, I'll make a note of that and see if I can force the situation somehow.

Some test observations are forthcoming. Thanks.

—M

ghaerr:
> The 'kludge' picks up the mac address from an incoming ICMP which would normally trigger an ARP request and then wait. The effect is like a hard wired arp cache entry, except more flexible, and the ICMP can be responded to right away. I had envisioned to just let the arp request proceed normally in parallel, but that was too complicated.

Ok, got it. You're welcome to add the code (should be very easy), in `icmp_process` to do this. Take a close look at line 192 in the new arp.c, where `arp_cache_update(ipsrc, ethsrc)` is used to possibly "merge" (now called update) an existing entry. The following `arp_cache_add` then adds the entry if its for ktcp's IP. So 2-3 lines of code should do it. When the ICMP response goes out through `ip_sendpacket`, the ARP cache will find the entry and no ARP request generated. This method does "shortcut" and could cause a problem if an ICMP request comes in after an ARP request is lost (and a packet queued). When the ICMP response is generated in this case, it would send it without knowing a queued packet is waiting, and no additional ARP request would be sent. On the flip side, if a packet were queued awaiting an ARP response, and an ICMP was received beforehand, the current code would drop the ICMP response and it would never be regenerated. The dropped packet alert is currently a printf rather than debug_ so that you one can always see dropped packets while testing. We will need better stats (coming) to fully understand what the network stack is doing.

Using the above method, there's no need to use the 'valid' flag, and all the ktcp code uses a printf or debug_arp do display when the entry is added. The code should be \#ifdef'd, because it's not really needed, and for the security reasons you mentioned. We can add flag bits to the valid flag if it makes sense to track these things.

> Even better … (I grew up with 'arp -a' which linux still uses, and wouldn't mind that either, along with arp -d to delete and arp -a to add hardwired entries :-) )

Its a bit complicated how netstat communicates with ktcp, so it's probably easier to keep all that in a single program. Display, adding and deleting ARP entries wouldn't be a big deal from netstat however.

Mellvik:
@ghaerr,
Initial testing leaves a couple of questions/comments:

1. (this is unrelated to the arp update - I've been working with older code for a while): When ktcp starts, it reports `link 0 `and a bunch of serial related info even if we're using ethernet.
2. Elks is now (some scenarios) sending  duplicate replies when responding to ping, see log below. This test is with the old driver, so there is a possible caveat, but I have a hard time seeing how the driver can cause this. Also, it does not seem to happen with small (standard) ICMP packets, this test is with 400 byte packets. Further - 1200 byte packets do not trigger this problem, so maybe it's lower level stuff anyway.
3. (observation) We need to get the HW ether address back into the config message. It's on my list.

The testing also illustrates more clearly the NIC buffer handling problems of the old driver. Increasing the ICMP packet size so the resulting packet exceeds 1 page (256 bytes) cause ELKS to return echo replies with erroneous data - typically one off (e.g. b3 instead of b4), indicating one byte lost. This happens when using the 8390 NIC, not the newer RTL PnP NIC.

```
link 0 serdev /dev/eth baud 57600 mtu 1024
# icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 1)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 2)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 3)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 4)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 5)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 6)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 7)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 8)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 9)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 10)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 11)
icmp: PING from 10.0.2.2 (len 408 id 5235 seqnum 12)
```
Output from the pinging host:
```
helge@raspberrypi:~ $ ping -s 400 10.0.2.15
PING 10.0.2.15 (10.0.2.15) 400(428) bytes of data.
408 bytes from 10.0.2.15: icmp_seq=1 ttl=64 time=62.4 ms
408 bytes from 10.0.2.15: icmp_seq=1 ttl=64 time=62.7 ms (DUP!)
408 bytes from 10.0.2.15: icmp_seq=2 ttl=64 time=60.3 ms
408 bytes from 10.0.2.15: icmp_seq=3 ttl=64 time=60.3 ms
408 bytes from 10.0.2.15: icmp_seq=3 ttl=64 time=60.7 ms (DUP!)
408 bytes from 10.0.2.15: icmp_seq=4 ttl=64 time=60.2 ms
408 bytes from 10.0.2.15: icmp_seq=5 ttl=64 time=60.2 ms
408 bytes from 10.0.2.15: icmp_seq=6 ttl=64 time=60.3 ms
408 bytes from 10.0.2.15: icmp_seq=6 ttl=64 time=60.7 ms (DUP!)
408 bytes from 10.0.2.15: icmp_seq=7 ttl=64 time=60.2 ms
408 bytes from 10.0.2.15: icmp_seq=8 ttl=64 time=60.2 ms
408 bytes from 10.0.2.15: icmp_seq=8 ttl=64 time=60.6 ms (DUP!)
408 bytes from 10.0.2.15: icmp_seq=9 ttl=64 time=60.2 ms
408 bytes from 10.0.2.15: icmp_seq=9 ttl=64 time=60.7 ms (DUP!)
408 bytes from 10.0.2.15: icmp_seq=10 ttl=64 time=61.3 ms
408 bytes from 10.0.2.15: icmp_seq=11 ttl=64 time=61.3 ms
408 bytes from 10.0.2.15: icmp_seq=12 ttl=64 time=61.4 ms
^C
--- 10.0.2.15 ping statistics ---
14 packets transmitted, 12 received, +5 duplicates, 14% packet loss, time 13027ms
rtt min/avg/max/mdev = 60.218/60.848/62.795/0.774 ms
```

More tests later.

--M

ghaerr:
@Mellvik, thanks for your testing and comments!!

> this is unrelated to the arp update - I've been working with older code for a while): When ktcp starts, it reports link 0 and a bunch of serial related info even if we're using ethernet.

Actually the "link 0" means 0=ethernet. I will update the display so that it makes more sense.

> Elks is now (some scenarios) sending duplicate replies when responding to ping, see log below. This test is with the old driver, so there is a possible caveat, but I have a hard time seeing how the driver can cause this. Also, it does not seem to happen with small (standard) ICMP packets, this test is with 400 byte packets. Further - 1200 byte packets do not trigger this problem, so maybe it's lower level stuff anyway.

Interesting. This could be the new queuing mechanism somehow sending the packet twice. Turn on DEBUG_ARP in config.h and lets see what the output shows. That shows all the queuing and ARP activity.

> We need to get the HW ether address back into the config message. It's on my list.

You mean in the initial "ktcp: ..." message? I'll add that, the link message, and delete the memcmp for promiscuous mode and update this PR.

> The testing also illustrates more clearly the NIC buffer handling problems of the old driver. Increasing the ICMP packet size so the resulting packet exceeds 1 page (256 bytes) cause ELKS to return echo replies with erroneous data - typically one off (e.g. b3 instead of b4), indicating one byte lost. This happens when using the 8390 NIC, not the newer RTL PnP NIC.

Interesting. Since it only happens on the 8390, it definitely sounds like a NIC driver issue, rather than ktcp. I'm looking forward to getting your enhanced driver!

ghaerr:
Enhance ktcp startup display
Fix net start slip/cslip parameter errors
Remove memcmp checking for packets since promiscuous mode not used

Mellvik:
> Ok, got it. You're welcome to add the code (should be very easy), in icmp_process to do this. Take a close look at line 192 in the new arp.c, where arp_cache_update(ipsrc, ethsrc) is used to possibly "merge" (now called update) an existing entry. The following arp_cache_add then adds the entry if its for ktcp's IP. So 2-3 lines of code should do it. When the ICMP response goes out through ip_sendpacket, the ARP cache will find the entry and no ARP request generated. This method does "shortcut" and could cause a problem if an ICMP request comes in after an ARP request is lost (and a packet queued). When the ICMP response is generated in this case, it would send it without knowing a queued packet is waiting, and no additional ARP request would be sent. On the flip side, if a packet were queued awaiting an ARP response, and an ICMP was received beforehand, the current code would drop the ICMP response and it would never be regenerated. The dropped packet alert is currently a printf rather than debug_ so that you one can always see dropped packets while testing. We will need better stats (coming) to fully understand what the network stack is doing.
> 
> Using the above method, there's no need to use the 'valid' flag, and all the ktcp code uses a printf or debug_arp do display when the entry is added. The code should be #ifdef'd, because it's not really needed, and for the security reasons you mentioned. We can add flag bits to the valid flag if it makes sense to track these things.
> 
OK, I'll take a close look at this approach.
> Even better … (I grew up with 'arp -a' which linux still uses, and wouldn't mind that either, along with arp -d to delete and arp -a to add hardwired entries :-) )
> 
> Its a bit complicated how netstat communicates with ktcp, so it's probably easier to keep all that in a single program. Display, adding and deleting ARP entries wouldn't be a big deal from netstat however.
> 
> 
Sounds good - and I agree, less code in ktcp itself is good regardless. Netstat may even have an 'arp' alias to do the traditional arp things the arp() way.

—Mellvik

PR #668, opened 2020-07-15
[kernel] Add keyboard LED support

ghaerr:
Add keyboard LED support to direct console from Minix 2.

@Mellvik - Can you take a look at this for me? Although the LED-setting code is almost entirely lifted from Minix 2, none of the keyboard LEDs turn on or off when any of CapsLock/NumLock/ScrollLock are pressed/released. I have placed a printk in the set_leds() function that shows what I think is the proper values changing, but nothing happens to the LEDs. I am reading conflicting documentation regarding whether port 0x64 or port 0x60 should be used on PC/ATs, I have tried setting the KEYBD define both ways to the same no result. Tested not working on both Compaq Portable and desktop 386. Apparently the XT doesn't support keyboard LEDs.

Note: only the CapsLock key is toggling properly in this code (even though no LEDs change). The NumLock key needs a bit more logic, I think line 190 needs changing for NumLock toggle. I'm using the unchanged ModeState variable, which is used elsewhere in the keyboard driver. I haven't written that code as the LEDs don't do anything yet. It looks like Minix 2 calls set_leds() with a 0 value initially, whose function is to turn off NumLock at boot.

The Compaq Portable boots normally with all LEDs off, while the desktop 386 boots with NumLock on. I would like to boot and force the keyboard in to NumLock off state as well.


ghaerr:
Updated: parameters to outb() are reversed in ELKS vs Minix. After this change, ELKS boots on desktop 386 with NumLock off, but then hangs after NumLock/CapsLock pressed, with no LED change and not processing any other characters. I have printk() in between each step in set_leds(). Displays "12345[20]" and hangs.

Mellvik:
@ghaerr,
I'm not surprised - this stuff is so counter intuitive that - in my experience - any attempt to apply optimization logic is going to break it. Even regular logic. I spent a lot of time getting ModeState to work and my attempt to merge LED states into it failed miserably.

Anyway - I ported some of my changes over to your file and it works - LEDs and num keypad - with the following caveats:

When capslock is on, the upper row is unshifted (as it should) but shift does not work. I.e shifted chars on the upper keyrow are unavailable when capslock is on. Not right.
Capslock breaks the keypad.
SLOCK does not work. I did put it in (LED only), and then removed it because in order to get it right, we have to change the codes used in the tb_state table. In my (local only) version SLOCK is 0x80, SSC is 0 and some logic dependent upon SSC being 0xC0 was changed. It works but doesn't fix the problems above, probably the opposite. So I'm not convinced it's a good idea and left that part out. OTOH we cannot reasonably expand the table to words instead of bytes - for one single not-so-important state. So I figured we'll either come up with something special just for SLOCK or more or less forklift the minix2 version over.


I'm uploading the diff via web.

—Mellvik

> 15. jul. 2020 kl. 05:44 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> Updated: parameters to outb() are reversed in ELKS vs Minix. After this change, ELKS boots on desktop 386 with NumLock off, but then hangs after NumLock/CapsLock pressed, with no LED change and not processing any other characters. I have printk() in between each step in set_leds(). Displays "12345[20]" and hangs.
> 
> —
> 
>

Mellvik:
Here's the code. Capslock work as numlock, numlock has no real effect, led works on SLOCK.
It seems to me the reason for NUMlock not working is that there is no key table supporting that. When that's been added this should be OK.
```
helge@iMac27 char % cat ~/Downloads/xt_key.diff 
39a40,41
> static int  scan_keyboard(void);
> 
42a45
> #define toggle(x) (x = 1-x)
51c54,57
< static unsigned int ModeState = 0;	/* also led state*/
---
> static unsigned int ModeState = 0;
> static int capslock;
> static int numlock;
> static int slock;
68a75
> #define SLOCK	0x80	// non functional, need to redefine SSC in order for this to work
79,81c86,88
<     NUM, SSC, SSC,				/*45->47*/
<     SSC, SSC, SSC, SSC, SSC, SSC, SSC, SSC,	/*48->4F*/
<     SSC, SSC, SSC, SSC, SSC, SSC, SSC, 'k', 'l' /*50->58, F11-F12*/
---
>     NUM, SLOCK, SSC,				/*45->47*/ 
>     0xB7, SSC, SSC, 0xBA, SSC, 0xB9, SSC, SSC,	/*48->4F*/
>     0xB8, SSC, SSC, SSC, SSC, SSC, SSC, 'k', 'l' /*50->58, F11-F12*/
118,119d124
<     set_leds();		/* turn off numlock led*/
< 
120a126
> 
122a129,135
> 
>     capslock = 0;
>     numlock = 0;
>     slock = 0;
> 
>     set_leds();
>     (void)scan_keyboard();      /* clean leftovers */
133d145
<     static int capslocktoggle =0;
139,145c151
<     code = inb_p((void *) KBD_IO);
< 
<     /* Necessary for the XT. */
<     mode = inb_p((void *) KBD_CTL);
<     //printk("\nkey read1:%X,mode:%X,ModeState:%d\n",code,mode,ModeState);    
<     outb_p((unsigned char) (mode | 0x80), (void *) KBD_CTL);
<     outb_p((unsigned char) mode, (void *) KBD_CTL);
---
>     code = scan_keyboard(); // Does the right thing for XT too ...
179d184
<     if (!(mode & 0xC0)) {
180a186,187
>     if (!(mode & 0xC0) || mode == SLOCK) {  // Not a simple scancode 
>  
187,195c194,224
< 	if (keyReleased) {
< 	    if (mode == CAPS)
< 		capslocktoggle = !capslocktoggle;
< 	    if (mode != CAPS || !capslocktoggle)
< 		ModeState &= ~mode;
< 	} else
< 	    ModeState |= mode;
< 	set_leds();
< 	return;
---
> 	if (keyReleased) { // the key was released
> 
>           switch (mode) {
>           case CAPS:
>                 toggle(capslock);
>                 if (capslock) mode = 0;
>                 //printk("C%d",capslock);
>                 set_leds();
>                 break;
>  
>           case NUM:
>                 //printk("N");
>                 toggle(numlock);
>                 if (numlock) mode = 0;
>                 set_leds();
>                 break;
>  
>           case SLOCK:
>                 printk("S");
>                 toggle(slock);
>                 //if (slock) mode = 0; /* for now slock doesn't do anything */
>                 set_leds();
>           }
>           ModeState &= ~mode;   /* clear these and other modes when key released */     
>         } else
>           ModeState |= mode;  /* key down: set mode bit */
> 
>         printk("M%02x", ModeState);
> 
>         /* ModeState updated - now return */
>         return;
306a336,348
> static int
> scan_keyboard(void) {
>     int code, mode;
> 
>     code = inb_p((void *) KBD_IO);
>     mode = inb_p((void *) KBD_CTL);
> 
>     outb_p((unsigned char) (mode | 0x80), (void *) KBD_CTL);
>     outb_p((unsigned char) mode, (void *) KBD_CTL);
> 
>     return(code);
> }
> 
334a377
>     unsigned leds;
337c380,382
< printk("1");
---
>     /* encode LED bits */
>     leds = slock | (numlock << 1) | (capslock << 2);
> 
340d384
< printk("2");
342d385
< printk("3");
345,347c388
< printk("4");
<     outb_p(ModeState, KEYBD);	/* give keyboard LED values */
< printk("5");
---
>     outb_p(leds, KEYBD);	/* give keyboard LED values */
349d389
< printk("[%x]", ModeState);
```

ghaerr:
Thanks @Mellvik!!

I really appreciate your help on this. It looks like what I was missing is the LED bits aren't matched to the returned mode bits like I thought, but just occupy the lower 3 bits. When I set an upper bit 0x20 for CapsLock, it hung the system.

```
>     0xB7, SSC, SSC, 0xBA, SSC, 0xB9, SSC, SSC,	/*48->4F*/
>     0xB8, SSC, SSC, SSC, SSC, SSC, SSC, 'k', 'l' /*50->58, F11-F12*/
```
What is the reason for the 0xB7-0xBA change instead of SSC above? I thought I reviewed this when fixing the arrow keys and they should be removed. The arrow keys are supposed to be picked up in the code tables, I thought.

> When capslock is on, the upper row is unshifted (as it should) but shift does not work. I.e shifted chars on the upper keyrow are unavailable when capslock is on. Not right.
> Capslock breaks the keypad.

Do these two items still need fixing, or does your code handle them?

Thank you!

Mellvik:
> I really appreciate your help on this. It looks like what I was missing is the LED bits aren't matched to the returned mode bits like I thought, but just occupy the lower 3 bits. When I set an upper bit 0x20 for CapsLock, it hung the system.
Happy to contribute, makes the time I spent scratching my head worth it.
> >     0xB7, SSC, SSC, 0xBA, SSC, 0xB9, SSC, SSC,	/*48->4F*/
> >     0xB8, SSC, SSC, SSC, SSC, SSC, SSC, 'k', 'l' /*50->58, F11-F12*/
> What is the reason for the 0xB7-0xBA change instead of SSC above? I thought I reviewed this when fixing the arrow keys and they should be removed. The arrow keys are supposed to be picked up in the code tables, I thought.
> 
Hate to admit, I forget. I have to dive back into it to recall. Will do that.

> When capslock is on, the upper row is unshifted (as it should) but shift does not work. I.e shifted chars on the upper keyrow are unavailable when capslock is on. Not right.
> Capslock breaks the keypad.
> 
> Do these two items still need fixing, or does your code handle them?
> 
There are still several issues. I'm taking a break from the network testing tomorrow and looking into it. The numlock key as of now does not cause selection of the correct keytable. Minor thing, but this is an incredible mess (really, the hardware). I guess that's 1980 for you.

SO - unless you're diving back in right now, I'll add some fixes tomorrow - then you do the PR...

BTW - do you have any thoughts on activating the Scroll Lock key? Like, having it emit alternating ^S ^Q to stop and restart output?

--M
>

ghaerr:
Hello @Mellvik,

I've taken your working code (thank you very much!), fixed a few small problems, added ScrollLock handling and documented the seven step process for better understanding and added it to this PR. All previous code is now bug-free, with only NumLock and CapsLock issues remaining, should we want to fix them.

Details:
Added @Mellvik's working LED and NumLock code.
Added ScrollLock LED processing.
Fixed CapsLock bug when NumLock on.
Complete source code documentation of the process from scan code to keyboard character(s).
Neither NumLock or ScrollLock are coded to do any extra processing at this point. Internal NumLock processing is handled by the controller, so keyboard characters are modified by NumLock, just not by the ELKS driver.
Since NumLock isn't the same as CAPS or SHIFT, further thinking is required to determine what, if anything, special should be done by ELKS.

Bugs:
Shift-CapsLock still doesn't get the correct character on the top (numeric keys) of the keyboard. This can probably be implemented by XORing the SHIFT state in a special line of code, but I didn't want to do that until this is tested on more hardware.
NumLock ON doesn't work on some 10-key keypads. This will likely require another per-country table to work properly, although it is possible a single system-wide table could work. Numlock OFF is working properly, which is the main reason I wanted to contribute this PR in the first place, since one of my 386's boots with Numlock ON (until now!).

Tested on desktop 386 and Compaq Portable.

> do you have any thoughts on activating the Scroll Lock key? Like, having it emit alternating ^S ^Q to stop and restart output?

That's a possibility, but I am now thinking of something more cool: ScrollLock could, when the light is on, turn on certain kernel debug messages, and turn them off by hitting ScrollLock again. This could also be programmed for other library debugging, like `malloc/free` heap dumps. I've been wondering about how this could be quickly done programmatically and now realize that a keyboard key could be exactly what's needed for efficient debugging.

Mellvik:
Great @ghaerr - I’ll test again, and include the compaq slt386 (‘laptop’) in the setup. Different keyboard, new layout (arrow keys etc.)

> 
> Complete source code documentation of the process from scan code to keyboard character(s).
> Neither NumLock or ScrollLock are coded to do any extra processing at this point. Internal NumLock processing is handled by the controller, so keyboard characters are modified by NumLock, just not by the ELKS driver.
> 
Are you sure about this? The hardware (at least what hardware I’ve seen) is unable to do this.
> Since NumLock isn't the same as CAPS or SHIFT, further thinking is required to determine what, if anything, special should be done by ELKS.
> 
NUMLOCK is actually shift-keypad. And that’s what the Minix2 driver is taking advantage of. More below.
> Bugs:
> Shift-CapsLock still doesn't get the correct character on the top (numeric keys) of the keyboard. This can probably be implemented by XORing the SHIFT state in a special line of code, but I didn't want to do that until this is tested on more hardware.
> 
Agreed.
> NumLock ON doesn't work on some 10-key keypads.
> 
Are you saying it works on some keypads? While I haven’t looked closely at your latest changes yet, my take is that all we need is a line with something like this:
if (scancode is in keypad range [HOME <= ch && ch <= INSRT]) && NUMLOCK set), use shifted keytable. I’ll test that tomorrow unless you’ve beaten me to it. :-)
> This will likely require another per-country table to work properly, although it is possible a single system-wide table could work. Numlock OFF is working properly, which is the main reason I wanted to contribute this PR in the first place, since one of my 386's boots with Numlock ON (until now!).
> 
The NumLock on at boot time is a CMOS/config setting on the machine, you can change it using the compaq setup program.
> Tested on desktop 386 and Compaq Portable.
> 
> do you have any thoughts on activating the Scroll Lock key? Like, having it emit alternating ^S ^Q to stop and restart output?
> 
> That's a possibility, but I am now thinking of something more cool: ScrollLock could, when the light is on, turn on certain kernel debug messages, and turn them off by hitting ScrollLock again. This could also be programmed for other library debugging, like malloc/free heap dumps. I've been wondering about how this could be quickly done programmatically and now realize that a keyboard key could be exactly what's needed for efficient debugging.
> 
Agreed, great idea!!! That will be very welcome! Somewhat like Minix2, which is using regular FKeys and only for device stats if I remember correctly. 
BTW - I was hoping we could add SLOCK functionality w/o expanding the table from bytes to words. Only (!) about 60 bytes, but then again …. Opinion?

—Mellvik

ghaerr:
@Mellvik, thanks for your comments, they helped me understand how to best implement the fixes for keypad w/Numlock. The code is a bit tricky, as the keyboard controller ORs in a LSHIFT bit on arrow keys when Numlock is on, it appears. Took me a while to understand that.

The last commit fixes both Numlock and Capslock bugs/issues mentioned above. Everything should be working now.

> I was hoping we could add SLOCK functionality w/o expanding the table from bytes to words. Only (!) about 60 bytes, but then again ….

I looked hard at this, and we need a bit for simple scan code as well as one for SLOCK, and there was only 1 bit left. Its possible I suppose to encode things a bit more, but given the complexity of the code already, I don't think its a big deal to expand the small table from byte to word.

Thanks a lot for your help on this! We should be good to go. I believe I finally fully understand the keyboard driver, and am starting to understand the keyboard controller's strange responses!

Mellvik:
Again thanks @ghaerr - great work. This issue has been hanging there unresolved for too long. Now we're 'home'.
> I was hoping we could add SLOCK functionality w/o expanding the table from bytes to words. Only (!) about 60 bytes, but then again ….
> 
> I looked hard at this, and we need a bit for simple scan code as well as one for SLOCK, and there was only 1 bit left. Its possible I suppose to encode things a bit more, but given the complexity of the code already, I don't think its a big deal to expand the small table from byte to word.
> 
I have an idea. My previous solution was -as you may remember - to change SSC to zero, which works but has some side effects in that a few tests further down become more complicated. The reason it works is that SSC keys are not mode keys so they don't require a mode bit. The same can be argued for SLOCK. Whether it's ZERO or 0x10000 doesn't really matter - there is no mode bit to set or test, just a value. Its state does not affect any other keys. Even when you add functionality to the slock key later, it doesn't need a mode bit - which is required only when the state affects other keys.

I'll be testing this later today. Thanks again.

//Mellvik

ghaerr:
Reduce scan code class table size back from int to byte.
Fixed Alt-F1 through Alt-F3 to switch virtual consoles.
Final cleanup, use \#defines for scan codes.
No logic changed, no re-testing needed.

> The same can be argued for SLOCK. Whether it's ZERO or 0x10000 doesn't really matter - there is no mode bit to set or test, just a value. Its state does not affect any other keys. Even when you add functionality to the slock key later, it doesn't need a mode bit - which is required only when the state affects other keys.

Thanks @Mellvik, I was able to use your idea and set SLOCK to 0, which allowed the table entry size to be set back to byte.

> works perfectly on all, with one small exception: The keypad is still affected by capslock. It shouldn't.

Looking into this, it looks like the fix is to change the per-country CAPS scan code table entries for this problem. They need to be changed from numeric to the (undocumented) function key codes for the keypad in the NORMAL state. I don't want to change any of the scan code tables in this PR so that will wait for later. Further, the testing of this mode is a bit complicated, since both NumLock, CAPS and SHIFT all interact and produce different results, which I think are now correct with the exception of this issue.

Mellvik:
Again thanks @ghaerr,
I agree with you - this is next to perfect, and should be merged.

There are two items on the (future) fix list (yes, I'm nitpicking ...)
-- The inline doc should be updated to reflect the SLOCK change (mode bits)
-- CAPS: I have not been as specific as I should have on this issue. Here's the thing: CAPS should affect letters only. Which means that not only should the upper row be untouched by CAPS status, everything else should too, including the keys around the return key (and keypad etc.). As it is, the -?|[]{} etc are now inconsistent and incorrect. I guess the most important consequence of this observation is that our thinking around CAPS should be reversed: It's not which keys should be unaffected, it's which keys should be affected. Alphabetic characters only. I suspect btw that there may be some inaccuracies in the US keymap (possibly others). To be investigated (I didn't have time for a closer look when I discovered this).

--Mellvik

ghaerr:
> -- The inline doc should be updated to reflect the SLOCK change (mode bits)

Not sure what exactly needs to be updated?

> CAPS should affect letters only. Which means that not only should the upper row be untouched by CAPS status, everything else should too, including the keys around the return key (and keypad etc.). As it is, the -?|[]{} etc are now inconsistent and incorrect.

A quick test of the US table shows that -?[]| work correctly with CapsLock on, and ;:'\ are broken. Each table will have to be gone through individually with that country's keyboard, I'm afraid.

> our thinking around CAPS should be reversed: It's not which keys should be unaffected, it's which keys should be affected


The keyboard driver has four per-country tables: normal, shift, caps and ctrl-alt. The scancode-to-character conversions for these first three states are entirely determined by the tables. It looks like what you're saying is that the caps table entries are likely incorrect for many countries.

Agreed country tables should be modified in a separate PR.

PR #670, opened 2020-07-17
[keyboard] Update US CAPS table for non-alphabetic keys

ghaerr:
Implement @Mellvik's suggestion that CAPS state should only shift alphabetic characters.

Updates US CAPS keyboard table for proper operation of non-alphabetic characters in CAPS state.

This fixes the ;'`\ characters, but another problem remains: when CapsLock is ON, everything is correct, but then using SHIFT does not shift these characters to their labeled shift state (like it does with the top row). The top row is handled specially within the driver, but I don't think it a good idea to special case these scan codes, as that's what the tables are for.

Just substituting the per-country "shift" or "normal" table also won't work when CapsLock and SHIFT are set. May have to live with this small annoyance using current table design.


ghaerr:
@Mellvik - 

Figured a way to make CapsLock capitalize only alphabetic characters as you proposed. The interpretation of the per-country CAPS table is now such that it only contains non-zero values for scancodes that should be capitalized when CapsLock is ON. Zero values will use existing other characters from the Normal and Shift tables automatically.

This modification allowed removing the special-cased handling for the main top-row keys, as they are already specified in the normal and shift tables.

Everything now works as it should, including CapsLock, CapsLock w/Shift, NumLock, and keypad keys.

There are no other modifications made to the per-country tables. I cleaned up all the US table values for easier readability, using ASCII characters rather than octal values for all ASCII (< 0x7f).

If this passes your testing, I suggest having you make the same changes to the Norwegian CAPS keyboard table and submit a PR for it. For those country CAPS tables that haven't been tested or modified, they will continue to work, except the CapsLock with SHIFT won't work.

Mellvik:
Congratulations!
I'm all smiles, and looking forward to more testing in the morning.

More comments privately.

--M

Mellvik:
Just briefly testing this this morning - looks all good with (the usual) minor exception: When in NUMLOCK mode, the keypad key to the right of the '0' (scan code 71) should emit '.', not DEL.

--Mellvik

ghaerr:
> When in NUMLOCK mode, the keypad key to the right of the '0' (scan code 71) should emit '.', not DEL.

Actually, I'm aware of this, and decided to leave it be for the time being: there are some issues with QEMU on macOS, where on Apple keyboards the "grey keys" (which include Delete) sometimes operate as though Numlock is on, and there is no Numlock key. I need to investigate this further on QEMU, and wanted to make sure I had a separate Backspace vs Delete key for `sh` linenoise testing!

PR #672, opened 2020-07-18
[ktcp] Enhance netstat w/packet statistics, add arp command

ghaerr:
First pass at adding networking stats to `ktcp`.

TCP, IP, ICMP, ARP and ETH sent/received/dropped/chksum error, etc reported.

Accessed via `netstat`.

Comments welcome!

Screen Shot 2020-07-17 at 6 47 39 PM

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -150,6 +150,7 @@ ktcp/ktcp						:net
 ktcp/resolv.conf	::etc/resolv.conf	:net
 inet/nettools/netstat			:net
 inet/nettools/nslookup			:net
+inet/nettools/arp				:net
 inet/telnet/telnet				:net
 inet/telnetd/telnetd			:net
 inet/tinyirc/tinyirc			:net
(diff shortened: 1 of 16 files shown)


Mellvik:
Great tool,
thanks @ghaerr.

Some thoughts:
Eventually some options to select which stats are interesting
Arp cache manipulation like we tdiscussed before (maybe using an arp alias for the command)
It may be an idea to add an IOCTL to the ne2k driver to pull the stats from the NIC (collision, runts, too big packets, checksums, buffer overflow etc.)
maybe even the same for serial, chosing to consider the serial port a network device

… of the top of my head.


— Mellvik

ghaerr:
Add `arp` command to display ARP cache.
Fix ARP bug on previous commit.

@Mellvik, here you go, 'arp' and 'netstat' can be used to inspect the ARP cache as well as see deeper into network activity. This PR also fixes an inadvertent bug I introduced in the last commit, which had the effect of zero-ing the ARP cache and writing to a zero-valued ethernet destination. This could have caused some problems in your testing.

Screen Shot 2020-07-18 at 3 03 02 PM

Mellvik:
@ghaerr,
My morning coffee just got even better. Thanks.
And no, I didn't get around to do any testing yesterday, so your transient bug never hit me.

I'm looking forward to share a very populated netstat screenshot.

--Mellvik

PR #678, opened 2020-07-19
[ktcp] Retune timeouts for TCP retransmission timers

ghaerr:
Adjust default TCP timer settings for retransmitting packets.
Set minimum timeout for SLIP/CSLIP to be 1/2 second. This should fix the needless retransmissions seen by @pawosm-arm and mentioned in https://github.com/jbruchon/elks/issues/610#issuecomment-660626265.

### DIFF
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -29,16 +29,6 @@
 /* bytes to subtract from window size and when to force app write*/
 #define PUSH_THRESHOLD	512
 
-/* timeout values in seconds*/
-//#define TIMEOUT_ENTER_WAIT	30	/* length of TIME_WAIT state*/
-#define TIMEOUT_ENTER_WAIT	10	/* length of TIME_WAIT state*/
-//#define TIMEOUT_CLOSE_WAIT	240	/* length of CLOSING/LAST_ACK/FIN_WAIT states*/
-#define TIMEOUT_CLOSE_WAIT	10	/* length of CLOSING/LAST_ACK/FIN_WAIT states*/
-//#define TIMEOUT_INITIAL_RTT	4	/* initial RTT before retransmit*/
-#define TIMEOUT_INITIAL_RTT	1	/* initial RTT before retransmit*/
-
-/* following timeout is in 1/16 ticks, not seconds*/
-#define TIMEOUT_MIN_SLIP	8	/* minimum retrans timeout for slip/cslip (1/2 sec)*/
 
 #define PROTO_TCP	0x06
 
@@ -60,7 +50,7 @@
 
 #define TCP_SETHDRSIZE(c,s)	( (c)->data_off = (s) << 2 )
 
-#define ENTER_TIME_WAIT(cb)	{ (cb)->time_wait_exp = Now + (TIMEOUT_ENTER_WAIT << 4); \
+#define ENTER_TIME_WAIT(cb)	{ (cb)->time_wait_exp = Now + (30 << 4); \
 				  (cb)->state = TS_TIME_WAIT; \
 				  tcp_timeruse++; \
 				  cbs_in_time_wait++; }
@@ -116,7 +106,7 @@ struct tcpcb_s {
 	__u16	remport;
 
 	__u8	state;
-	timeq_t	rtt;				/* in 1/16 secs*/
+	timeq_t	rtt;
 
 	__u32	time_wait_exp;
 	__u8	unaccepted;			/* boolean */
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -29,6 +29,16 @@
 /* bytes to subtract from window size and when to force app write*/
 #define PUSH_THRESHOLD	512
 
+/* timeout values in seconds*/
+//#define TIMEOUT_ENTER_WAIT	30	/* length of TIME_WAIT state*/
+#define TIMEOUT_ENTER_WAIT	10	/* length of TIME_WAIT state*/
+//#define TIMEOUT_CLOSE_WAIT	240	/* length of CLOSING/LAST_ACK/FIN_WAIT states*/
+#define TIMEOUT_CLOSE_WAIT	10	/* length of CLOSING/LAST_ACK/FIN_WAIT states*/
+//#define TIMEOUT_INITIAL_RTT	4	/* initial RTT before retransmit*/
+#define TIMEOUT_INITIAL_RTT	1	/* initial RTT before retransmit*/
+
+/* following timeout is in 1/16 ticks, not seconds*/
+#define TIMEOUT_MIN_SLIP	8	/* minimum retrans timeout for slip/cslip (1/2 sec)*/
 
 #define PROTO_TCP	0x06
 
@@ -50,7 +60,7 @@
 
 #define TCP_SETHDRSIZE(c,s)	( (c)->data_off = (s) << 2 )
 
-#define ENTER_TIME_WAIT(cb)	{ (cb)->time_wait_exp = Now + (30 << 4); \
+#define ENTER_TIME_WAIT(cb)	{ (cb)->time_wait_exp = Now + (TIMEOUT_ENTER_WAIT << 4); \
 				  (cb)->state = TS_TIME_WAIT; \
 				  tcp_timeruse++; \
 				  cbs_in_time_wait++; }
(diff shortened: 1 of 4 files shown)


pawosm-arm:
Sadly, retransmissions still occur, less often and with somewhat bigger `rto` values, e.g.:
```
retrans retry #1 rto 163460 mem 21
```

ghaerr:
> retrans retry #1 rto 163460 mem 21

I'm glad to hear they are less often, so the tuning has helped. Let me explain the RTT/RTO a bit, as I've come to believe we may not be able to fully stop retransmissions, but it depends on what you are doing.

First, the rto value above is huge, and is indicative of an internal error. The value is in 1/16 seconds, so 163460/16 = 10,216 seconds, too long for you to have waited. Is that the correct number? If so, we have memory corruption.

When a TCP connection is first started, ktcp sets the RTT estimate to 1 second (=16 RTT/RTO). Whenever a packet is received, the round trip time (=RTT) is calculated, and 10% of that value is used to update the RTT (90% alpha, i.e., the RTT is slowly moved towards the "seen" RTT time. On slip connections, the initial connect took a while, which upped the RTT. Then, after the shell ran, typing returns etc with small packets brought the RTT down to 1/8 second or so (RTT 2). Then, when "ls" was run, ELKS took a while to load ls and that caused a timeout and retransmission. So I set a minimum RTT for slip of 1/2 second so that the system wouldn't retransmit every time a program was run.

When ktcp sends a packet, it also queues that packet up for retransmit, and sets the timeout to twice the current RTT and sets that as RTO. Then, when that time elapses without an ACK packet, it retransmits that packet and doubles the RTO.

Thus, the system initially sets 1 second RTT, which means a 2 second RTO. This allows for the login/sh to run, and then the RTT/RTO drops down to the 1/2 (minimum range) quickly, as the user hits return and gets shell prompts. If the user then quickly types a command that requires long disk access, the RTO could get exceeded and a retransmit will occur.

Hopefully this helps you/me understand what is happening. I have no idea how the RTO could get set so high to 163460 though!

pawosm-arm:
This value caught my attention exactly because it was huge. Yet keep in mind, I'm pinging ELKS all the time so those retrans messages are interleaved with ping reports that ELKS outputs when pinged (maybe we should get rid of them eventually), so I can't guarantee this retrans message wasn't disturbed.

ghaerr:
Well, this is memory corruption for sure. Geez.

How are you pinging ELKS, I thought you were using SLIP only, not ethernet? It is definitely possible that ping could be corrupting things, @Mellvik solved the last major corruption bug when he found that the ping output buffer was overwriting the previously receive buffer when ip_sendpacket was reentered.

On another topic, have you tried CSLIP? I can't get it to work at all, ktcp shows receive TCP checksum errors whenever I try it. The underlying VJHC protocol doesn't actually pass the checksums across, so the VJHC code is likely in error. It produces some bad compiler warnings and is on my list. Funny thing is that long ago, others have reported CSLIP working, or so i thought.

> ping reports that ELKS outputs when pinged (maybe we should get rid of them eventually)

Agreed, thanks for the reminder. I was thinking of emitting these messages when ktcp -d is used, but have kept some important ones in for now to easily see certain activity and errors.

pawosm-arm:
I've just tried to use CSLIP (set on both sides, in slattach and ktcp). Although TCP connections can be established, TCP checksum errors soon start to pop out, so it's unusable.

I don't understand ping question. I'm opening `sl0` network interface on the Linux side on top of SLIP connection. I can ping it, and those ICMP pings return regularly (which is a progress as it wasn't the case some time ago). I've avoided pinging during my most recent test so I could observe those tco values in retrans error messages. And they are always that big!

It is very suspicious though that only Linux to ELKS telnet connection is disrupted. The other way round, ELKS to Linux is still alive and kicking. Something wrong going on with ELKS's telnetd?

ghaerr:
> I don't understand ping question. I'm opening sl0 network interface on the Linux side on top of SLIP connection. 

Sorry - I got confused thinking of 'arping' and ARP from an earlier comment. I'll re-test and try to duplicate the larger RTO values now that you're duplicating them without ping.

> It is very suspicious though that only Linux to ELKS telnet connection is disrupted. The other way round, ELKS to Linux is still alive and kicking. Something wrong going on with ELKS's telnetd?

Indeed, very suspicious, thank you. I was thinking the reason was that in the Linux->ELKS case only ELKS was sending packets, but both the user and the shell should be sending data with each keypress, unless somehow local echo is being incorrectly set. In any case, the ELKS->Linux connection should produce the same RTO issue, at least for the terminal packets input from the user. I'll look further into ELKS->Linux and try to duplicate RTO issues.

pawosm-arm:
Blah, regarding the disrupted telnet connection issue, I guess it's partially the client issue. I'm using PuTTy for Linux as a telnet client on the Linux side. Yet now, I've just tried the normal command-line telnet client on the Linux side and the connection isn't disrupted for a long time. Even with having ELKS to Linux telnet connection opened simultaneously. I suspect, PuTTy sends some kind of keepalive message when idle, and it makes telnetd go haywire.
This observation does not affect huge rto's issue.

ghaerr:
> I suspect, PuTTy sends some kind of keepalive message when idle, and it makes telnetd go haywire.

Sounds like we need to add the ability to debug/display telnet sequences when they are sent or received?

FYI, there have been terrible problems with telnet processing both on our existing telnet, which is so bad that it needs wholesale replacement, and with telnetd, which had no telnet processing at all (and somehow managed to work years ago, I know not why). I ended up wholesale-adding an outside telnet processing library to telnetd, but it has not been thoroughly tested.

pawosm-arm:
I guess more pressing issue is to find out why those rto values are so huge...

pawosm-arm:
> I ended up wholesale-adding an outside telnet processing library to telnetd, but it has not been thoroughly tested.

Hah, I had the same problem while developing stdiotelnetd. If you used libtelnet for implementing telnetd, it is already known to be treating PuTTy badly. Yet it should not cause entire daemon to die....

ghaerr:
I just booted ELKS with the latest source and am connecting from debian Linux 4.19 to ELKS, as well as ELKS to Linux, both over SLIP at 19200 baud.

This only works if I start telnet from ELKS to Linux first. I'm using 'net start slip 19200' before running telnet.

Can't get any RTO messages despite running for quite a while.

I wonder if this corruption and RTO have something to do with telnet sequences being produced and/or processed by telnet/telnetd, and corrupting memory? Very strange.

Restarted ELKS and ran 'net start slip 4800', restarted slip at 4800 on Linux side.
First time, got 'ktcp: Refusing packet' multiple times, no connect.

Restarted ELKS again, ran net start slip 4800, and got telnet out to connect, then telnet in and connect.

Ran some pings from Linux->ELKS. Received and responded properly.

Not getting any retransmission RTO messages ever.

Strange that 4800 baud didn't work a couple times, then worked.

Booted ELKS a 3rd time, net start slip 4800. Now works with connect from Linux first, then telnet from ELKS.

Looks like we have various spurious networking problems, including possible sync issues getting SLIP working after changing baud rates. Not getting any RTO messages/retries.

pawosm-arm:
> First time, got 'ktcp: Refusing packet' multiple times, no connect.

I saw it only once today. I thought it was a serial connection glitch so I rebooted ELKS and everything was fine with SLIP after that...

> This only works if I start telnet from ELKS to Linux first. I'm using 'net start slip 19200' before running telnet.

I guess you must be doing that on a much faster machine (386+) than I do as it's capable to do 19200 without hardware flow control. I suspect I'm seeing those retransmissions for that reason: the XT machine I'm running ELKS on is just too slow.

ghaerr:
> I saw it only once today. I thought it was a serial connection glitch so I rebooted ELKS and everything was fine

Seeing 'ktcp: Refusing packet' means ktcp received a well-formed packet, but that ktcp doesn't have a connection control block open for that address/port pair. That could happen if the serial line had queued data, or slattach retried sending old data for some reason. I suspect there may be sync issues possibly in getting SLIP working without possible garbage characters or packets in between. For now, we can not worry about it since things work once synced.

> I suspect I'm seeing those retransmissions for that reason: the XT machine I'm running ELKS on is just too slow.

My machine is faster, certainly, but I don't think your system is too slow, the time calculations are based on time() reported from the ELKS clock. Does date seem to run correctly?

Given that your system is in fact running ELKS TCP/IP with some success, I say congratulations! We will get things figured out.

In any case, even with a very slow machine, one should never see a super large RTO like you did, so I suspect the real problem is just continued old-fashioned memory corruption in other areas. We're now running network sessions long enough to start seeing the other bugs.

pawosm-arm:
> Does date seem to run correctly?

Yes, it gives consistent values, monotonic at least.

> Given that your system is in fact running ELKS TCP/IP with some success, I say congratulations!

Indeed if we agree that PuTTy issue was client specific, ELKS networking seems to be robust and fully functional now, those retransmissions don't seem to disturb TCP or ICMP communication.

PR #681, opened 2020-07-27
[ktcp] Fix network ^C handling

ghaerr:
Fixes ktcp and telnetd exiting when ^C generated SIGINT for another process, because they hadn't set a new process group with `setsid`.


Mellvik:
Wow,
Small change in code, big change in functionality!!

Thanks, @ghaerr.

--Mellvik

PR #682, opened 2020-08-05
[kernel] Fix select busyloop, networking semaphore lockups, strace, ...

ghaerr:
Batch of kernel fixes as a result of debugging "telnet 10.0.2.15" (localhost) networking, which now works great.
Fixes two of the five major outstanding network problems.

Fixes occasional but persistent hangs when running telnet & telnetd locally.
Fixes bad socket return value from inet_read/write causing buffer overflow in telnet.
Adds support for multiple network processes in same routines, fixes previous bad semaphore locking with tcpdev driver and ktcp; finally fixes all telnet localhost connect and hang issues by adding global kernel semaphore for inet_read/write.
Removes async read mechanism from tcpdev/ktcp, read with no data returns -EINTR (may change to -EAGAIN). All network apps must call select to block and read returns EINTR if no data. tcpdev_checkread now updates sock->data_avail for select blocking.
Modifies telnet and telnetd to report buffer overflow and EINTR read returns to display when errors occur.
Fixes kernel busy looping in select() as the result of race in inet_select; race condition with wake_up_process sets TASK_RUNNING and do_select loop.

Got CONFIG_STRACE (trace system calls w/strace) working. Modified _strace, _ret_strace, never worked since gcc conversion. Sync strace.h to syscall.dat. Had to compile strace.c -fno-optimize-sibling-calls for no tail calls, as was using interrupt parameter stack as temp space in tail call to printk(). Updated syscall table parameter types, and display of pointer to long and long params.

Got CHECK_SS working. Double checks SS == current->t_regs.ss on entry from user mode syscall or interrupt. All interrupt and syscall entry code gone through and should all be working well, except for two calls to schedule() and do_signal() which still run with interrupts disabled in the return to user mode from hardware interrupts from user mode, which could be excessive.

Remove unneeded wake_up in chq_getch() - should dramatically affect speed on serial I/O, as all characters received were calling wake_up which for looped the entire task list on each character.

Made regular serial driver default.

Fix ttyname(2) to not stat "." entries.

Tested on QEMU, will test on real hw before removing extra debug statements.


Mellvik:
Very inspiring, @ghaerr!! Another mega-PR!

A few questions.


> Adds support for multiple network processes in same routines, fixes previous bad semaphore locking with tcpdev driver and ktcp; finally fixes all telnet localhost connect and hang issues by adding global kernel semaphore for inet_read/write.
> 
Does this mean multiple telnet/telnetd sessions will now work?
> Removes async read mechanism from tcpdev/ktcp, read with no data returns -EINTR (may change to -EAGAIN). All network apps must call select to block and return returns EINTR if no data. tcpdev_checkread now updates sock->data_avail for select blocking.
> 
Seems EAGAIN makes most sense...
> Modifies telnet and telnetd to report buffer overflow and EINTR read returns to display when errors occur.
> Fixes kernel busy looping in select() as the result of race in inet_select; race condition in wake_up_process sets TASK_RUNNING and do_select loop.
> 
This may be totally unrelated, but then again: I’ve noticed that when a telnet (in or out) is active (probably any tcp connection), eth_select gets called twice a second when there is no traffic, more often when there is traffic. After the connection is terminated, these timed calls continues for the lifetime of the tcp process. Is this fix related to that behaviour?
> Got CONFIG_STRACE (trace system calls w/strace) working. Modified _strace, _ret_strace, never worked since gcc conversion. Sync strace.h to syscall.dat. Had to compile strace.c -fno-optimize-sibling-calls for no tail calls, as was using interrupt parameter stack as temp space in tail call to printk(). Updated syscall table parameter types, and display of pointer to long and long params.
> 
Thank you - this is very valuable indeed.
> Remove unneeded wake_up in chq_getch() - should dramatically affect speed on fast serial I/O, as all characters received were calling wake_up which for looped the entire task list on each character.
> 
Wow - looking forward to more serial testing…

Thanks again - testing late next week.


—Mellvik

ghaerr:
> Does this mean multiple telnet/telnetd sessions will now work?

Yes for telnet, no for telnetd. Telnetd isn't written to handle multiple clients, its on my list to enhance shortly.


> This may be totally unrelated, but then again: I’ve noticed that when a telnet (in or out) is active (probably any tcp connection), eth_select gets called twice a second when there is no traffic, more often when there is traffic. After the connection is terminated, these timed calls continues for the lifetime of the tcp process. Is this fix related to that behaviour?

No, not related: the twice-a-second polling by ktcp is not changed (yet). I added that for debugging some time ago; it will be removed once everything is fixed.

On another note, after deeply inspecting all interrupt handling routines in ELKS, I realized we have a potential big problem in the ethernet driver. Specifically, the ethernet driver C interrupt routine `ne2k_int` is programmed to be called whenever the NIC finishes receiving a packet. The first thing ne2k_int does is call `ne2k_int_stat` to get the RCV and XMIT status bits, in order to wake up any sleeping process waiting on RCV or XMIT activity (which would be ktcp, sleeping in `select` after calling `eth_select` above). The problem is, that ne2k_int_stat calls page_select, which re-loads the NIC command registers. If the ELKS kernel happens to be in the middle of the NIC driver, say,  reading or transmitting a packet, and the NIC card finishes receiving or sending a (different) packet, then the interrupt routine will change the NIC page register right in the middle of it, without the driver knowing it.

I haven't tested anything yet, just observed this problem, now that I fully understand exactly how ELKS interrupt processing works. Running the interrupt routine with interrupts off won't help. I would need more information on exactly o the NIC expects interrupts to be handled in order to devise a solution. It is possible this isn't a problem, depending on how long it takes to generate another interrupt once a packet is read or transmitted, but FYI the driver page_select routine could be called as well as io_ne2k_command register reloaded at any time, unless various parts of the original driver were protected with cli/sti pairs where this matters.


> Got CONFIG_STRACE (trace system calls w/strace) working. Modified _strace, _ret_strace, never worked since gcc conversion. Sync strace.h to syscall.dat. Had to compile strace.c -fno-optimize-sibling-calls for no tail calls, as was using interrupt parameter stack as temp space in tail call to printk(). Updated syscall table parameter types, and display of pointer to long and long params.
> Thank you - this is very valuable indeed.

Yes, try it - one can learn a lot about the system by watching all system calls with parameters on the serial debug console. This is how I happened to find the ttyname(2) directory searching issue also fixed.


> Remove unneeded wake_up in chq_getch() - should dramatically affect speed on fast serial I/O, as all characters received were calling wake_up which for looped the entire task list on each character.
> Wow - looking forward to more serial testing… 

Yes - amazingly enough, lots of overhead on every single character. I will be testing this also when I return from vacation.

Mellvik:
Interesting indeed.
> On another note, after deeply inspecting all interrupt handling routines in ELKS, I realized we have a potential big problem in the ethernet driver. Specifically, the ethernet driver C interrupt routine ne2k_int is programmed to be called whenever the NIC finishes receiving a packet. The first thing ne2k_int does is call ne2k_int_stat to get the RCV and XMIT status bits, in order to wake up any sleeping process waiting on RCV or XMIT activity (which would be ktcp, sleeping in select after calling eth_select above). The problem is, that ne2k_int_stat calls page_select, which re-loads the NIC command registers. If the ELKS kernel happens to be in the middle of the NIC driver, say, reading or transmitting a packet, and the NIC card finishes receiving or sending a (different) packet, then the interrupt routine will change the NIC page register right in the middle of it, without the driver knowing it.
> 
> I haven't tested anything yet, just observed this problem, now that I fully understand exactly how ELKS interrupt processing works. Running the interrupt routine with interrupts off won't help. I would need more information on exactly o the NIC expects interrupts to be handled in order to devise a solution. It is possible this isn't a problem, depending on how long it takes to generate another interrupt once a packet is read or transmitted, but FYI the driver page_select routine could be called as well as io_ne2k_command register reloaded at any time, unless various parts of the original driver were protected with cli/sti pairs where this matters.
> 
You’re touching right in the middle of what I’ve been struggling with the past 2 weeks. And you’re absolutely right, disabling interrupts doesn’t help. And this is not a problem unless we have high traffic load (back to back packets almost). There are two mechanisms that I believe will alleviate this. One is using a semaphore in  the low level read and write routines, to avoid reentry, which is what the minix driver is using. The other is very careful selection of which bits to touch in the command register. The NIC is designed to handle changes to the command reg while doing transfers, but the docs are imprecise as to what is actually possible or safe. SO it has to be tested manually, and retested for different NIC chip types. I’m on it - literally.
> Got CONFIG_STRACE (trace system calls w/strace) working. Modified _strace, _ret_strace, never worked since gcc conversion. Sync strace.h to syscall.dat. Had to compile strace.c -fno-optimize-sibling-calls for no tail calls, as was using interrupt parameter stack as temp space in tail call to printk(). Updated syscall table parameter types, and display of pointer to long and long params.
> Thank you - this is very valuable indeed.
> 
> Yes, try it - one can learn a lot about the system by watching all system calls with parameters on the serial debug console. This is how I happened to find the ttyname(2) directory searching issue also fixed.
> 
> 
Will do! Right now I’m trying to get qemu to work with tun/tap so we can get full networking tests w/o hardware. You tried that?

Expect to push a first round of NIC driver changes tomorrow.


—Mellvik

ghaerr:
>  disabling interrupts doesn’t help.

For the interrupt entry point, true. Easily fixed by disabling interrupts in the driver, but a potentially high cost, e.g. serial input. We can estimate times by counting instructions in the main driver. I have done a study, which I am thinking of opening an issue on, where the number of instructions for the fast serial interrupt routine are counted, as well as the ELKS general interrupt processing overhead. I have also analyzed all kernel procedures called by all interrupt routines, in an attempt at quantifying the amount of overhead used by interrupt routines.


> There are two mechanisms that I believe will alleviate this. One is using a semaphore in  the low level read and write routines, to avoid reentry, which is what the minix driver is using. 

That won't work - using any kind off semaphore in the interrupt routine is disallowed, as interrupts can not be rescheduled in ELKS.


> The other is very careful selection of which bits to touch in the command register. The NIC is designed to handle changes to the command reg while doing transfers, but the docs are imprecise as to what is actually possible or safe.

Yes, this sounds good. My first thought is to eliminate the call to page_select in the interrupt entry point - is it really needed? All that is needed is the RX and TX status bits from the io_ne2k_int_stat register. If reading that register requires page 0 selected, then all other driver code must be manually inspected to and cli/sti blocks put around calls to page_select and the next register access. Short sections of cli/sti should not be a big problem.

Another idea would be to turn off TX interrupts entirely. With an outgoing ring buffer, it is unlikely that ELKS would have to wait for ring buffer space in order to transmit a packet. With TX interrupts off, the RX interrupt would not have to read the interrupt status register, thus removing the above problems entirely. However, it appears the interrupt still needs to be acknowledged by writing the interrupt status register. If this can be done without a page 0 select, then this idea will work fine.


>  SO it has to be tested manually, and retested for different NIC chip types.

The data sheet should talk about whether page 0 is required to read the interrupt status register.

> Expect to push a first round of NIC driver changes tomorrow.

If you've got your other fixes ready to go (great!), then perhaps it would be best to get them in before tackling this newly found (at least by me) problem.

ghaerr:
Reimplement blocking read in inet_read for network sockets; required for proper operation of httpd and other network utilities. Implements read wait for sock->avail_data.

Fix httpd leaving zombie procs after each request.
Fix libc fread & gets to properly handle negative (-EAGAIN) return from read() and set stdio MODE_ERR.
Reset TTY to blocking mode before ash prompt to handle killed or crashed telnet sessions.  This fixes continuous "#" prompt display due to tty being in non-blocked mode. Sash will now exit with error with libc fix in this case.

Mellvik:
﻿
> disabling interrupts doesn’t help.
> 
> For the interrupt entry point, true. Easily fixed by disabling interrupts in the driver, but a potentially high cost, e.g. serial input. We can estimate times by counting instructions in the main driver. I have done a study, which I am thinking of opening an issue on, where the number of instructions for the fast serial interrupt routine are counted, as well as the ELKS general interrupt processing overhead. I have also analyzed all kernel procedures called by all interrupt routines, in an attempt at quantifying the amount of overhead used by interrupt routines.
> 
Very interesting indeed. That’s a lot of work.
> There are two mechanisms that I believe will alleviate this. One is using a semaphore in the low level read and write routines, to avoid reentry, which is what the minix driver is using.
> 
> That won't work - using any kind off semaphore in the interrupt routine is disallowed, as interrupts can not be rescheduled in ELKS.
> 
If we were talking about true interrupt service routines, I would agree. In this context I don't.  It may even be argued that these routines (in the NIC driver) aren’t really interrupt routines at all - per my definition, which you may not share. To me a (real) interrupt routine can only be started by an interrupt. This is not what we have here. Actually, ELKS can run ethernet tcp traffic w/o NIC interrupts enabled - as I found out but accident. Eventually it will hang, but if you turn off all NIC interrupts and initiate a telnet connection from elks, it will work. And once started, it will respond to ICMP, ARP etc. quite 'normally'. Until some ‘unfortunate’ coincidence that requires an interrupt. To me, a clean interrupt driver structure is one that isolates the system (OS) from the interrupt driven data IO. Asyncronous. The lack of such asynchronicity in ELKS creates the kind of situations we're discussing. Including the possiblity of an 'interrupt routine' being reentered. Thus the semaphore. Which is perfectly viable, a safety measure. 
This is an observation about the current system, not a critique. The 'elks way' has the distinct benefit that it saves bufferspace. It also complicates things - and quite possibly makes the system appear slower and/or less agile.
The point in this context is that the read and write routines in the ne2k must mot be reentered, and the semaphore mechanism will ensure that. Return EBUSY if the flag is set. To be tested.
> The other is very careful selection of which bits to touch in the command register. The NIC is designed to handle changes to the command reg while doing transfers, but the docs are imprecise as to what is actually possible or safe.
> 
> Yes, this sounds good. My first thought is to eliminate the call to page_select in the interrupt entry point - is it really needed? All that is needed is the RX and TX status bits from the io_ne2k_int_stat register.
> 
> If reading that register requires page 0 selected, then all other driver code must be manually inspected to and cli/sti blocks put around calls to page_select and the next register access. Short sections of cli/sti should not be a big problem.
> 
There are several ways to fix this, and believe me I've been through a few over the past month. Including av variant of your suggeastion: Since register page 1 is used just a few times in well defined places, it is entirely possible to isolate those, enclose them in disabled interrupts and then make the 'reliable' assumption that register page 0 is otherwise always active. I decided against it because the NIC is clearly prepared to handle frequent accesses to the command register without disturbing anything. 
> Another idea would be to turn off TX interrupts entirely. With an outgoing ring buffer, it is unlikely that ELKS would have to wait for ring buffer space in order to transmit a packet. With TX interrupts off, the RX interrupt would not have to read the interrupt status register, thus removing the above problems entirely. However, it appears the interrupt still needs to be acknowledged by writing the interrupt status register. If this can be done without a page 0 select, then this idea will work fine.
> 
This may well Work, but from my point of view, it's curing the symptoms instead of the illness. Incidentally, this is pretty much the way things work now. The TX interrupt is information only, turning it off makes no difference (see above)
> SO it has to be tested manually, and retested for different NIC chip types.
> 
> The data sheet should talk about whether page 0 is required to read the interrupt status register.
> 
It is required.
> Expect to push a first round of NIC driver changes tomorrow.
> 
> If you've got your other fixes ready to go (great!), then perhaps it would be best to get them in before tackling this newly found (at least by me) problem.
> 
Like I said, I've been at this for some time. There is nothing here that affects operation under qemu. Under qemu, the current code works great (btw - there is no such thing as a tx interrupt for the ne2k in QEMU, never happens). There are two situations that are still problematic for the rewritten ne2k driver: high load (incoming packets with < 10ms interval) and long interrupt blocking (the good old floppy prog load problem, 5-10 seconds during which interrupts are being recorded but not responded to - i.e. overruns etc). However extreme, this must be handled in a predictable way.

Bottom line: This is not a problem, it either has been or will be fixed in the new ne2k driver. When I have beaten it up more on real hardware and tested the buffer overflow handling, this will work smoothly. It has been done before, we'll do it again.


//Mellvik

ghaerr:
>> That won't work - using any kind off semaphore in the interrupt routine is disallowed, as interrupts can not be rescheduled in ELKS.

> If we were talking about true interrupt service routines, I would agree. In this context I don't.  It may even be argued that these routines (in the NIC driver) aren’t really interrupt routines at all - per my definition, which you may not share. To me a (real) interrupt routine can only be started by an interrupt. This is not what we have here.

I am not sure we are talking about the same thing. I am talking about "true" interrupt service routines. Let me be very clear what I am trying to say with regards to the problem I have identified with the NE2K interrupt service routine, and where semaphores are not allowed:

Following is the interrupt service routine for the NE2K driver:
```
// Interrupt handler

static void ne2k_int (int irq, struct pt_regs * regs, void * dev_id)
{
    word_t stat;

    stat = ne2k_int_stat ();
    if (stat & NE2K_STAT_RX)
        wake_up ((struct wait_queue *) &rx_flag);

    if (stat & NE2K_STAT_TX)
        wake_up ((struct wait_queue *) &tx_flag);
}
```
This C function is registered to be called whenever hardware interrupt IRQ 9 (or 12, if setup that way) is triggered, by the following code in `eth_drv_init`:

        err = request_irq (NE2K_IRQ, ne2k_int, NULL);

The function `ne2k_int` is called, and the kernel (or wherever the CPU is) interrupted, on hardware interrupt 9. The first statement in the interrupt routine calls `ne2k_int_stat()` which looks like this:

```
   .global ne2k_int_stat

ne2k_int_stat:

    // select page 0

    xor     %al,%al
    call    page_select

    // get interrupt status

    xor     %ah,%ah

    mov     $io_ne2k_int_stat,%dx
    in      %dx,%al
    test    $0x03,%al
    jz      nis_next

    // acknowledge interrupt

    out     %al,%dx

nis_next:

    ret
```

The call to `page_select` is made during an interrupt routine, and scribbles on the NIC registers, setting page 0, regardless of what any other portion of the NE2K driver might have been doing. No other portions of the NE2K driver are interrupt routines.

My point is that the NE2K driver can be interrupted at any time, and the NE2K interrupt routine, above, will call page_select, get the status bits, acknowledge the interrupt, and return. After return, the NE2K driver will resume where it left off, but the NIC chip will be set to page 0, even if it was just set to page 1 and needs to be by the driver.

One cannot use semaphores in the driver code to keep the interrupt routine out, as the interrupt cannot be put to sleep.

With regards to the NE2K driver being "reentered", this will never happen, as the ELKS kernel will never stop processing any kernel code unless either 1) the current process is purposely put to sleep (wait on even) by calling a xxx_sleep_on() function, or 2) a hardware interrupt occurs. In the latter case, the only "reentry" will be routines specifically called by the interrupt routine, which in this case is only `page_select`, so page_select could be technically "re-entered", but that's not the real problem here.

The only protection available without rewriting the driver would be to protect critical regions in the driver with cli/sti pairs. 
I will comment on driver rewriting possibilities in another post.

I hope this clears things up!

ghaerr:
> Actually, ELKS can run ethernet tcp traffic w/o NIC interrupts enabled - as I found out but accident. Eventually it will hang, but if you turn off all NIC interrupts and initiate a telnet connection from elks, it will work. And once started, it will respond to ICMP, ARP etc. quite 'normally'. 

The reason this works is that both the read and write packet entry points in the NE2K driver first check the RX and TX status, respectively. If they are set (meaning that a packet has already been received, or the NIC is ready to accept a packet to transmit), then the process is not put to sleep, and the driver is immediately called to get or send a packet.

This will continue until the NIC isn't ready at the time requested, in which case the process is put to sleep until a NIC interrupt occurs to wake up the process. Since it seems that the ELKS CPU processing in ktcp is considerably slower than sending an ethernet packet, this seems to work OK, since the NIC remains generally ready.

> Until some ‘unfortunate’ coincidence that requires an interrupt. 

Well, not really. The NE2K driver is programmed to sleep the process if the NIC isn't ready, pure and simple. Here's the code for the read process, for instance:

```
static size_t eth_read (struct inode * inode, struct file * filp,
    char * data, size_t len)
{   
    size_t res;
    
    while (1) {
        size_t size;  // actual packet size
        
        if (ne2k_rx_stat () != NE2K_STAT_RX) {          // this code checks NIC is RX ready
            
            if (wait_event_interruptible (&rx_cond)) {  // this code will put the process to sleep if no RX packet 
                res = -EINTR;  // interrupted by signal
                break;
            }
        }
        
        if (ne2k_pack_get (recv_buf)) {                     // otherwise this code reads the packet
            res = -ERESTARTSYS;  // panic
            break;
        }
```
In the interrupt service routine, the following code determines whether the NIC is RX ready, and then wakes up the process waiting in the wait_event_interruptable routine above:
```
static void ne2k_int (int irq, struct pt_regs * regs, void * dev_id)
{
    word_t stat;

    stat = ne2k_int_stat ();
    if (stat & NE2K_STAT_RX)
        wake_up ((struct wait_queue *) &rx_flag);    // wakes up process waiting in wait_event_interruptable if RX ready

    if (stat & NE2K_STAT_TX)
        wake_up ((struct wait_queue *) &tx_flag);
}
```

> The point in this context is that the read and write routines in the ne2k must mot be reentered, and the semaphore mechanism will ensure that. Return EBUSY if the flag is set. To be tested.

The read and write routines will never be reentered, instead, the calling process will either perform the read or write immediately, or will be put to sleep using the standard Linux/ELKS sleep mechanism, only to be "woken up" (stack-switched back) after the NIC sends and interrupt indicating packet received or ready to transmit.

In some sense, these standard sleep/wake routines are in a sense a kind of semaphore, in that they prevent a process from continuing. However, use of any actual ELKS semaphore isn't ever needed, as neither the read nor write entry points will ever be "reentered", except any routines called by the driver interrupt service routine (which is only page_select in this driver).

Note, that a driver entry point can be used by multiple processes "at a time", but never simultaneously, as the ELKS kernel will never switch stacks/tasks in the middle of a kernel routine, EXCEPT when a hardware interrupt occurs, where ELKS calls the registered interrupt service routine and its called functions.

ghaerr:
> There are several ways to fix this, and believe me I've been through a few over the past month. Including av variant of your suggeastion: Since register page 1 is used just a few times in well defined places, it is entirely possible to isolate those, enclose them in disabled interrupts and then make the 'reliable' assumption that register page 0 is otherwise always active. I decided against it because the NIC is clearly prepared to handle frequent accesses to the command register without disturbing anything.

I rather like this idea - always leaving the NIC in page 0 except for cli/sti portions requiring page 1. Can this be done for very short periods of time?

This would completely fix the interrupt routine problem of having to switch to page 0, and/or make its call to page_select non-harmful or unnecessary.

We need some kind of (simple) fix for this interrupt routine problem, as it will definitely corrupt the page value based on a packet received or finishing transmit if ELKS happens to be in the NIC driver ASM code.

> There are two situations that are still problematic for the rewritten ne2k driver: high load (incoming packets with < 10ms interval) and long interrupt blocking 

My close inspection of the ELKS interrupt/syscall processing code revealed that in certain cases, schedule() and do_signal() can be called with interrupts off, which could be the reason we sometimes see an occasional serial character drop at high speeds. This only occurs when an interrupt occurs and the process is in user, not kernel code, such as the clock or keyboard interrupt occurring outside a system call. I have some ideas for fixing this, coming up.

Mellvik:
@ghaerr,
> There are several ways to fix this, and believe me I've been through a few over the past month. Including av variant of your suggeastion: Since register page 1 is used just a few times in well defined places, it is entirely possible to isolate those, enclose them in disabled interrupts and then make the 'reliable' assumption that register page 0 is otherwise always active. I decided against it because the NIC is clearly prepared to handle frequent accesses to the command register without disturbing anything.
> 
> I rather like this idea - always leaving the NIC in page 0 except for cli/sti portions requiring page 1. Can this be done for very short periods of time?
> 
This is indeed easy to implement - and might even reduce the total # of instructions slightly.
> This would completely fix the interrupt routine problem of having to switch to page 0, and/or make its call to page_select non-harmful or unnecessary.
> 
I guess I’ve said it before - this should not be a problem — unless...
> We need some kind of (simple) fix for this interrupt routine problem, as it will definitely corrupt the page value based on a packet received or finishing transmit if ELKS happens to be in the NIC driver ASM code.
> 
… we have a concurrency problem, which I assume is what you’re alluding to. Ideally, there should be no chance of concurrency in the driver, and my assumption that it was ‘clean’ in that regard cost me countless hours while debugging the driver. The skeleton driver in one of the app notes, solves this by disabling the NIC IRQ (just a couple of instructions to the 8259) on driver entry, reenabling on exit. 
I’ve assumed we cannot easily do that in ELKS, so I chose the alternative: Delaying the interrupt acknowledgement until the processing is done, which is the natural thing to do anyway and doing otherwise is considered a bug. When the NIC fields an interrupt, the INT line remains ‘high’ until all unmasked interrupts have been cleared in the intr status reg. By delaying the RX intr ack, the handling of back to back packets improved drastically. There is still a problem, and it seems to be related to outgoing packets ‘colliding’ time wise with back to back incoming packets.
When I get back to the physical hardware on friday, I’ll get to test that by using sti as you suggested.
> There are two situations that are still problematic for the rewritten ne2k driver: high load (incoming packets with < 10ms interval) and long interrupt blocking
> 
> My close inspection of the ELKS interrupt/syscall processing code revealed that in certain cases, schedule() and do_signal() can be called with interrupts off, which could be the reason we sometimes see an occasional serial character drop at high speeds. This only occurs when an interrupt occurs and the process is in user, not kernel code, such as the clock or keyboard interrupt occurring outside a system call. I have some ideas for fixing this, coming up.
> 
It will be interesting to see if some of the shipload of recent improvements from you will fix the 'long delay' problem. Looking forward to it.

—Mellvik

ghaerr:
> … we have a concurrency problem, which I assume is what you’re alluding to. 

No, there should not be any concurrency problem. I have studied the ELKS code on interrupt entry, process scheduling and sleeping, and now (finally) completely understand how it all works. Actually, it is very straightforward, well-designed, and completely controls exactly what happens on any interrupt and task switch. (Unsurprisingly I suppose, but kind of surprised me). The design is the same as Linux 2.0, with the exception of SMP (multiprocessor support).

The Linux 2.0 design has three rules, also present in ELKS (excerpted from Linux Kernel Internals, 2nd Ed, 1998):

- No process running in kernel mode is interrupted by another process in kernel mode, except when it releases control and sleeps. This rule ensures that large areas of the kernel are atomic with respect to other processes and thus simplifies many functions in the Linux kernel.
- Interrupt handling can interrupt a process in kernel mode, but that in the end control is returned back to this same process. A process can block interrupts and thus ensure it will not be interrupted.
- Interrupt handling can only be interrupted by an interrupt of higher priority. In ELKS, all interrupt code runs on a separate stack and never allows a process switch if kernel mode code was interrupted.

The ELKS kernel in this regard is well-coded and provides a very simple mechanism that guarantees execution in certain ways for kernel code and interrupt drivers. I do not believe it is buggy, but instead quite stable. This implementation guarantees that the kernel itself is NOT reentrant on system calls. That is - a system call, such as read (a file, TTY, ethernet packet, etc) will not be interrupted by another process calling read or any other system call. The kernel code will always finish by returning to the user application that called it, OR be put to sleep using an explicity-coded sleep_on/wait_event/etc style wait function.

However, hardware interrupts ARE allowed to interrupt ONLY when three things are present: the 8259 controller must not be processing a higher-priority interrupt, it must have had any previous interrupt acknowledged, AND the CPU must not have interrupts disabled. Obviously, the hardware must have been previously programmed to enable interrupts, and appropriately registered the C routine to call with request_irq().

SO - I am almost certain that the problems you were/are seeing in the NE2K driver have to do with the interrupt service routine 'ne2k_int' running and reloading page 0 at any time, OR possible hardware issues relating to the NIC interrupt line and the 8259.

I'm not an 8259 expert, but the ELKS kernel handles it as follows:

- On software interrupts, or traps, nothing is sent to the 8259.
- On hardware interrupts, after saving registers and executing code to ensure the rules above are followed, CPU interrupts are re-enabled and the registered interrupt service handler is called. On return from the interrupt handler, ELKS disables interrupts, and then outputs an acknowledgement to the chained 8259 for interrupts 8-15, and then an acknowledgement to the master 8259. The saved registers are restored and an IRET executed to return to kernel mode, if the CPU was in kernel code, otherwise a possible task switch is performed to user mode.
- The 8259 will not allow another interrupt of the same priority until acknowledged by the ELKS kernel. 

With regards to NE2K interrupt handling:
- The NE2K interrupt handler, ne2k_int, only checks RX/TX status and calls wake_up, which wakes up any process that tried to read or write when the card wasn't ready to transmit or had no packet. This is the unusual case, as it seems that most kernel read calls don't get put to sleep.
- The NE2K interrupt handler acknowledges the interrupt at the end of the ne2k_int_stat (get status) call.
- The NE2K driver read and write routines run as a "normal" kernel process - NOT protected by 8259 interrupt handling. That is, if an incoming interrupt could derail the read or write process in the driver, then either CPU interrupts need to be disabled, or perhaps better the NIC interrupt turned off for the duration of the read/write routine.

> Ideally, there should be no chance of concurrency in the driver, and my assumption that it was ‘clean’ in that regard cost me countless hours while debugging the driver. 

I hear you on the countless hours of debugging. I've spent way too many hours trying to get networking running, and the last mess debugging the tcpdev interface that implements the kernel<->ktcp interface forced me to study until I fully understood how scheduling and interrupts actually work.

> The skeleton driver in one of the app notes, solves this by disabling the NIC IRQ (just a couple of instructions to the 8259) on driver entry, reenabling on exit.

I hadn't thought of this solution until just now, and it might be far easier than cli/sti pairs or rewriting the NE2K interrupt status to not reset page 0. Disabling the NIC interrupt by writing the NIC registers would be even easier than reprogramming the 8259, in my opinion. Then you can be guaranteed you won't interrupted in the driver code by your own interrupt handler, which I think is what's happening!


> I’ve assumed we cannot easily do that in ELKS, so I chose the alternative: Delaying the interrupt acknowledgement until the processing is done, which is the natural thing to do anyway and doing otherwise is considered a bug. When the NIC fields an interrupt, the INT line remains ‘high’ until all unmasked interrupts have been cleared in the intr status reg. By delaying the RX intr ack, the handling of back to back packets improved drastically. 

I propose not delaying any interrupt acknowledgements, since there shouldn't be any anyways except for in the actual interrupt driver routine 'ne2k_int', which already acks it. Instead, turning off the NIC interrupt immediately on entry to both the read and write routines is better, and there will be no need to deal with any interrupt or acknowledgments.

> There is still a problem, and it seems to be related to outgoing packets ‘colliding’ time wise with back to back incoming packets.

This will likely have to be handled by careful reading of the NIC data sheet.

**Summary**
ELKS isn't buggy with regards to interrupt handling.

I now realize that since the NE2K interrupt routine never reads a packet, but always writes the page select register and queries status, it is the likely culprit in messing things up.

The NE2k read and write routines can be interrupted at any time by the NE2K interrupt routine. Because of this, both the read and write routines should call a routine to disable NIC interrupts on entry, and enable them on exit. This should be done by writing a NIC register, and not by reprogramming the 8259, if possible. If the 8259 needs reprogramming, it needs to be done with interrupts disabled during the reprogramming and then reenabled.

The NE2k read and write routines should remove any code that acknowledges interrupts.

With these mods, the NE2K interrupt handler will never be called in the middle of a read or write, and the NE2K handler will never interrupt its own driver code, which it is certainly doing now. No other cli/sti pairs should be needed, and the NE2K interrupt routine remains only to wake process put to sleep in the mostly-unused case.

Thank you!

Mellvik:
Thank you @ghaerr for yet another detailed rundown of the inner workings of ELKS. Very educating and great material for the wiki. You’ve saved many of us hours of head scratching. Int’s very encouraging to know that we’re building on a sound foundation - makes it easier to spend the time and effort required to lift elks to the next level.

Further, the discussion contribute to our better understanding of the challenges with - and workings of - the NIC HW and driver & I/O. I still don’t understand why loading a file from floppy block everything, but as you have pointed out before, it may be a characteristic of the BIOS driver. On the other hand, while working on the NE2K driver I have noticed that when such activities (floppy I/O) are going on, interrupts are being recognized (ne2k-int gets called) but do not get serviced. SO packets are piling up in the NIC ring buffer which eventually overruns.
I’m getting ahead of myself, this issue is being revisited below - along with some clarifications.

> With regards to NE2K interrupt handling:
> 
> The NE2K interrupt handler, ne2k_int, only checks RX/TX status and calls wake_up, which wakes up any process that tried to read or write when the card wasn't ready to transmit or had no packet. This is the unusual case, as it seems that most kernel read calls don't get put to sleep.
> The NE2K interrupt handler acknowledges the interrupt at the end of the ne2k_int_stat (get status) call.
This is a bug AFAIK; I’ve changed this as described under bugfixes in the PR.
> The NE2K driver read and write routines run as a "normal" kernel process - NOT protected by 8259 interrupt handling. That is, if an incoming interrupt could derail the read or write process in the driver, then either CPU interrupts need to be disabled, or perhaps better the NIC interrupt turned off for the duration of the read/write routine.
A great summary - and this has been one of my key points all along: The ne2k driver is not interrupt-driven in a real sense and some of the routines are subject to reentry. We had a great example of this with the ARP code not long ago, and the ‘out of sequence nature’ of ARP processing may still be problematic. 
A while back I proposed to introduce send and receive buffers to ‘isolate' the driver from the kernel. That would make the driver truly interrupt driven, always cleaner - in my opinion. I still think it’s a good idea - but it does require more buffer memory (and some work).
> Ideally, there should be no chance of concurrency in the driver, and my assumption that it was ‘clean’ in that regard cost me countless hours while debugging the driver.
> 
> I hear you on the countless hours of debugging. I've spent way too many hours trying to get networking running, and the last mess debugging the tcpdev interface that implements the kernel<->ktcp interface forced me to study until I fully understood how scheduling and interrupts actually work.
> 
Very useful indeed! 
> The skeleton driver in one of the app notes, solves this by disabling the NIC IRQ (just a couple of instructions to the 8259) on driver entry, reenabling on exit.
> 
> I hadn't thought of this solution until just now, and it might be far easier than cli/sti pairs or rewriting the NE2K interrupt status to not reset page 0. Disabling the NIC interrupt by writing the NIC registers would be even easier than reprogramming the 8259, in my opinion. Then you can be guaranteed you won't interrupted in the driver code by your own interrupt handler, which I think is what's happening!
> 
Yes, this is what my changes to the int_stat routine does, although (possibly) in a slightly different way than you suggest. 
> I’ve assumed we cannot easily do that in ELKS, so I chose the alternative: Delaying the interrupt acknowledgement until the processing is done, which is the natural thing to do anyway and doing otherwise is considered a bug. When the NIC fields an interrupt, the INT line remains ‘high’ until all unmasked interrupts have been cleared in the intr status reg. By delaying the RX intr ack, the handling of back to back packets improved drastically.
> 
> I propose not delaying any interrupt acknowledgements, since there shouldn't be any anyways except for in the actual interrupt driver routine 'ne2k_int', which already acks it. Instead, turning off the NIC interrupt immediately on entry to both the read and write routines is better, and there will be no need to deal with any interrupt or acknowledgments.
> 
I’m curious: What’s the difference from your perspective? To me it seems natural (and easier) to make marginal changes to what we’re already doing instead of introducing yet another NIC register in the regular operation of the NIC. As mentioned above, the general ACK in int_stat is gone already (and a bug as far as I’m concerned), with positive consequences. 
> There is still a problem, and it seems to be related to outgoing packets ‘colliding’ time wise with back to back incoming packets.
> 
> This will likely have to be handled by careful reading of the NIC data sheet.
> 
Well, I know the datasheets by heart by now and there are - as always - things the datasheets don’t mention. Things we figure out with logic analyzers and/or tedious debugging. :-)
> Summary
> ELKS isn't buggy with regards to interrupt handling.
> 
> I now realize that since the NE2K interrupt routine never reads a packet, but always writes the page select register and queries status, it is the likely culprit in messing things up.
> 
Maybe - this is very easy to verify, and will be the first thing I test when I get back to the HW.
> The NE2k read and write routines can be interrupted at any time by the NE2K interrupt routine. Because of this, both the read and write routines should call a routine to disable NIC interrupts on entry, and enable them on exit. This should be done by writing a NIC register, and not by reprogramming the 8259, if possible. If the 8259 needs reprogramming, it needs to be done with interrupts disabled during the reprogramming and then reenabled.
> 
The current driver does this, by delaying the reset of the actual status bit in the interrupt status register, which in turn prohibits a new interrupt from being generated.
> The NE2k read and write routines should remove any code that acknowledges interrupts.
> 
That will not work and to me it does not make sense. A completion interrupt (like TX completed) can be acked immediately since it’s just a notification, while an action interrupt (like packet received) should not be acked until processed, in particular if the ack will enable another interrupt of the same type before processing is finished. There are other ways to do this (I’m repeating myself, bear with me), but using the INTR status register is (to me) the simplest: nothing to do, just change where something is done.
> With these mods, the NE2K interrupt handler will never be called in the middle of a read or write, and the NE2K handler will never interrupt its own driver code, which it is certainly doing now. No other cli/sti pairs should be needed, and the NE2K interrupt routine remains only to wake process put to sleep in the mostly-unused case.
> 
I may be missing something here, but here’s my take: Since the driver is not interrupt driven and networking actually works with interrupts disabled, dealing with interrupts alone is not going to guarantee anything - AFAIK. Again the ARP code is a great example - it’s calling sendpacket w/o going through the interrupt routine - for obvious reasons. BTW, if I remember correctly, NIC TX interrupts don’t even work in QEMU, and the system still operates fine. 

What we’re after is to ensure the atomicity of the driver calls. This discussion and your work on mapping out the inner workings of the ELKS interrupt handling will deliver that. I’m confident that we'll have a ne2k driver by the end of next weekend that can take any beating. I’m rather looking forward to it.

Thanks for an enlightening discussion, @ghaerr!

—Mellvik

ghaerr:
>  I still don't understand why loading a file from floppy block everything, but as you have pointed out before, it may be a characteristic of the BIOS driver.

The BIOS calls that ELKS uses do disk I/O synchronously. In a regular Linux system, the requesting process schedules disk I/O and is put to sleep waiting on a buffer, to be woken up by the I/O hardware complete interrupt. In ELKS, there is are no "disk I/O finished" interrupts. Instead, the kernel, when ready to start disk I/O, calls a BIOS routine which doesn't return until I/O is complete. Once the ELKS kernel calls the BIOS routine, the ELKS kernel itself is doing absolutely nothing, as it lost the CPU and will not regain the CPU until the BIOS reports back success or failure. During this time, hardware interrupts for other devices occur, but, since the kernel is not interruptible, as stated above, when the ELKS hardware interrupt service routine completes, no task switch will ever be performed. (It cannot, precisely since the BIOS is expecting the CPU and can't be "task-switched" away from the middle of a BIOS call and have it work).

Mellvik:
Thanks @ghaerr -
That makes sense - and (to me) revives the non functional disk/floppy drivers. We really need those,  and I'd love to (slowly) approach the challenge. Floppy first - as that's where the difference will be greatest. 

-Mellvik

PR #683, opened 2020-08-06
bugfix: Add forgotten include file in DEBUG mode.

marcin-laszewski:


ghaerr (changes requested):
Thanks for finding this. Can you remove the \#ifdef DEBUG around the header file and just include the header file, please? This will help reduce unneeded ifdefs.

marcin-laszewski:
Yes, of course! But in my opinion it is ok:

- DEBUG flags indicates that we need 'stat.h' file in "debug mode" only.
- Checking this flag is faster than include and parse 'stat.h' each time.

Your decision.

ghaerr:
> Checking this flag is faster than include and parse 'stat.h' each time.

The issue is not about compilation speed - but more importantly the philosophy of less \#ifdefs (more readability, no conditional compilation and associated issues). There are also no other cases where an ELKS header is included via \#ifdef.

Thank you!

marcin-laszewski:
OK. Done.

PR #684, opened 2020-08-06
Heavily updated ne2k driver

Mellvik:
This is a major rewrite of the ELKS NE2K driver. A short rundown of the changes:

**Fixes**
- A major flaw in the NIC ring buffer handling code caused random (occasional) data loss and driver hangs. The bug affected only 8390-based NICs, not newer PnP NICs, for reasons unknown. The fix replaces the rather convoluted (and slow) ring buffer wraparound code with a simple mechanism suggested by one of the 8390 application notes available on line.
- The old driver cleared the interrupt status register bits at the start of the interrupt service routines rather than at the end. For interrupts that trigger processing rather than indicating completion, this is some times catastrophic. As was the case with packet reads. If the time between arrival of packets was less than 10 milliseconds, a race condition occured with unpredictable results, normally a hung driver and hard hung system. The fix moves the clearing of interrupt status bits to the actual processinc routines rather than the  general interrupt dispatcher.
- NIC buffer overruns are handled, see below.

**Speed improvements**
The rewritten driver has a number of speed/resource improvements:
- Data transfer from to NIC to the system is now in word units rather than bytes, doubling the transfer speed.
- The bug fixes in the ring buffer handling will also improve speed significantly, leaving much more of the mechanics to the NIC rather than the driver.
- For most packets, there is only one data transfer from the NIC to main memory, instead of two, some times 3 previously. The NIC ring buffer consists of pages @ 256bytes ea. The old driver would read the first 4 bytes of a packet to get its size, the field a 2nd read to get the rest, and finally - if the packet traversed the ring buffer wraparound, a 3rd read to get the rest. The rewritten driver reads a full page (256 bytes), which in many cases contains the entire packet. For larger packets there is a 2nd read, and (for reasons mentioned above) there is never a 3rd read.

**Functional improvements**
- Notably, the driver now reads the NIC PROM at initalization and sets the MAC address to the NICs own address rather than a predetermined QEMU address. The MAC address can still be set via an IOCTL.
- The assembly part of the driver now uses the C preprocessor, eliminating the need for manual edit for port addresses other than the default 0x300.
- The rewritten driver handles NIC buffer overruns, which the old driver didn't. Such overruns will trigger console messages. Overruns require a complete reset of the NIC. This operation does not erase stored packets. However, when overruns occur, the reason is either a problem on the host or just too many incoming packets, so the overflow handler has no option but to discard data from the ring buffer. How to do this is determined by parameters to the handler from the interrupt dispatcher in eth-main.c. If the 'skip' parameter is 0, the ring buffer is cleared. Otherwise the 'skip' number of packets are deleted from the NIC ring buffer, currently 3 - while testing.

**TODO**
- An IOCTL to retrieve the stats from the NIC - runt packets, oversize packets, collissions, overruns, …
- Real DMA: Rudimentary code has been added to the driver to support DMA transfer from the NIC to system RAM. Using DMA will simplify the read code significantly and reduce system load - if the ELKS DMA support is actually working. No attempt has been made to test this yet.
- More testing, see below.
- The ‘get entire packet in one read’ is unfinished in that it currently re-reads the first 252 bytes of the packet if the total length is longer than 252.
- The buffer overrun handling needs testing on real hardware.

The driver has been extensively tested with a 8390 based card, a RT8139 (PnP) card and QEMU. It seems 100% stable on QEMU. On real hardware, the driver will hang under heavy load and ‘long’ periods of interrupt blocking (floppy access). An upcoming fix will remove these problems.

There is stil quite a bit of debugging code in the source, supporting the ongoing test/improvement.

**Testing**
Pings using a wide range of packet sizes and speeds (packets per sec) have bombarded the system from several sources - some times concurrently. Telnet in and out likewise. Incoming telnet some times gets a RST while a ling listing (my favourite is hd /bin/vi). This does not happen on QEMU.

--Mellvik

### DIFF
--- a/elks/arch/i86/drivers/net/eth-main.c
+++ b/elks/arch/i86/drivers/net/eth-main.c
@@ -1,5 +1,5 @@
 //-----------------------------------------------------------------------------
-// Ethernet driver
+// NE2K Ethernet driver (really DP8390 driver)
 //-----------------------------------------------------------------------------
 
 #include <linuxmt/errno.h>
@@ -21,9 +21,10 @@
 
 static byte_t eth_inuse = 0;
 
-static byte_t mac_addr [6] = {0x52, 0x54, 0x00, 0x12, 0x34, 0x56};  // QEMU default
+static byte_t def_mac_addr [6] = {0x52, 0x54, 0x00, 0x12, 0x34, 0x56};  // QEMU default
+static byte_t mac_addr [6]; // Current MAC address, from HW or set
 
-static byte_t recv_buf [MAX_PACKET_ETH];
+static byte_t recv_buf [MAX_PACKET_ETH+4];
 static byte_t send_buf [MAX_PACKET_ETH];
 
 
@@ -45,6 +46,7 @@ static size_t eth_read (struct inode * inode, struct file * filp,
 	char * data, size_t len)
 {
 	size_t res;
+	word_t page;	//debug
 
 	while (1) {
 		size_t size;  // actual packet size
@@ -62,6 +64,8 @@ static size_t eth_read (struct inode * inode, struct file * filp,
 			}
 		}
 
+                //page = ne2k_getpage();
+                //printk("/C%02x|B%02x/ ", (page>>8)&0xff, page&0xff);
 		if (ne2k_pack_get (recv_buf)) {
 			res = -ERESTARTSYS;  // panic
 			break;
@@ -71,8 +75,9 @@ static size_t eth_read (struct inode * inode, struct file * filp,
 		// otherwise end of packet will be lost
 
 		size = *((word_t *) (recv_buf + 2));
+                //printk("|RD: l%d s%d b%04x|", len, size, *(unsigned short *)recv_buf);
 		if (len > size) len = size;
-		memcpy_tofs (data, recv_buf + 4, len);
+		memcpy_tofs (data, recv_buf + 4, len);	// discarding statusbyte & counter
 
 		res = len;
 		break;
@@ -141,6 +146,8 @@ int eth_select (struct inode * inode, struct file * filp, int sel_type)
 {
 	int res = 0;
 
+        //printk("S%d|", sel_type); // Enable this to see when and how often
+
 	switch (sel_type) {
 		case SEL_OUT:
 			if (ne2k_tx_stat () != NE2K_STAT_TX)
@@ -175,14 +182,38 @@ int eth_select (struct inode * inode, struct file * filp, int sel_type)
 
 static void ne2k_int (int irq, struct pt_regs * regs, void * dev_id)
 {
-	word_t stat;
+	word_t stat, page;
 
 	stat = ne2k_int_stat ();
-	if (stat & NE2K_STAT_RX)
-		wake_up ((struct wait_queue *) &rx_flag);
+        page = ne2k_getpage();
+        printk("$%02x$B%02x$", (page>>8)&0xff, page&0xff);
 
-	if (stat & NE2K_STAT_TX)
+        if (stat & NE2K_STAT_OF) {
+		printk("Warning: NIC receive buffer overflow\n");
+		//page = ne2k_getpage();
+		//printk("/C%02x|B%02x/ ", (page>>8)&0xff, page&0xff);
+		// FIXME: This procedure is still being debugged
+		wake_up ((struct wait_queue *) &rx_flag);
+		//ne2k_clr_oflow(recv_buf); // The reset procedure needs to read the
+                                          // last complete packet in the buffer.
+		// If recv_buf is busy, this will overwrite the contents.
+		// What do we do with the contents? As is, this is a discard..
+	}
+
+	if (stat & NE2K_STAT_RX) {
+		//printk("|r|");
+		wake_up ((struct wait_queue *) &rx_flag);
+	}
+
+	if (stat & NE2K_STAT_TX) {
+		//printk("|t|");
 		wake_up ((struct wait_queue *) &tx_flag);
+	}
+	if (stat & NE2K_STAT_RDC) {
+		printk("RDC intr.\n");
+		// debug only, but keep this code, the RDC interrupt should be disabled in
+		// the low level driver.
+	}
 }
 
 
@@ -193,6 +224,8 @@ static int eth_ioctl (struct inode * inode, struct file * file,
 
 	{
 	int err = 0;
+
+	// Add ioctl to reset NIC
 
 	switch (cmd)
 		{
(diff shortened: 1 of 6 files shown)


Mellvik:
The minor change to netstat identifies incoming vs. outgoing ARP packets.

Mellvik:
I suggest changing the default IRQ for ne2k cards to 12 and update qemu.sh accordingly. IRQ 9 is a bad choice because many (PnP) cards disallow this - for unknown reasons.

ghaerr:
Wow @Mellvik! Thank you for this great contribution!

Particularly nice is moving from byte to word moves for speed, and getting rid of most of the ring buffer complexities :)

I haven't had time to digest your comments nor your code yet, and plan to tomorrow. I have noticed a few unrelated inclusions (like CONFIG_NEED_IRQ5 for serial, etc) that I will be code-reviewing quickly, before getting into the meat of this super contribution.

Thank you!

ghaerr (changes requested):
In the file ports.h, all lines above 49 should not be included, as they affect serial port configuration, not NE2K. Also, my PR #682 changes some of these, and if I commit that before this, the merge will be blocked.

Mellvik:
Yeah, I forgot.
I've been off the grid for a couple of days. What's the easy way to fix this?

--Mellvik

ghaerr:
The easiest way is to push another commit from the same branch:
```
git checkout ether
[remove changes to ports.h]
git commit -a
git push
[inspect GitHub after refresh]
```

Mellvik:
The original ports.h has been restored. This will prevent the ethernet driver from compiling. I will create a separate PR with the required NE2K changes, including the change to IRQ12 and the accompanying mod to qemu.sh.

ghaerr:
> I will create a separate PR with the required NE2K changes, including the change to IRQ12 

If you still have this branch (ether) on your system, just make the changes to ports.h on that branch, commit and push it, and it will all be included in this PR. Do you need help on that?

ghaerr:
@Mellvik,

Given the possible solution you and I have found in https://github.com/jbruchon/elks/pull/682#issuecomment-671061100, do you want to consider committing this now, or perhaps recoding and commit a final fix? I think the idea of turning off NIC interrupts in the driver is a very good idea.

I have reviewed the code, and other than the (tons of) debugging statements etc, which can be removed later, looks fine. I would like to accept a single commit that compiles and runs though, so that it can be easily reverted if needed.

Mellvik:
@ghaerr,

I updated and pushed ports.h (again) and qemu.sh as suggested. Should be OK now.

I’d go for merge now - it makes it easier (for me) to get current with all your recent updates. Further - since I’m now away from the physical hw, the next round of updates will have to wait til next weekend.

I’ll dive into #688 (comment) tomorrow.

—Mellvik

Mellvik:
The 2nd instance of this update is being readied for PR - and testing with all the recent higher level changes are very encouraging. Not the heavy banging yet, but telnet in, then telnet out, heavy ICMP Echo load. RTT (ping) is now (printk's gone) < 20ms, connects, disconnects, reconnects etc.  
14100DA9-3587-4E0F-9E8E-EA34034CFC17

ghaerr:
@Mellvik,

Very encouraging!!! Between your new driver and the many ktcp and networking bug fixes, it sounds like we are very close to having a reliable network operating system.

Super screen shot BTW. I'd like to use it in the planned updated of the ELKS home page, where we talk a bit more about usability and have a number of cool screenshots...

Mellvik:
Great idea!!

When I've cleaned up my git botches (!), the PR is ready... 

--M

Mellvik:
@ghaerr,

while preparing the PR, I managed to crash tic (finally): several active (in and out) sessions, and eventually I got this (it's repeatable, just takes some time to trigger):

```
                                                                                         retrans retry #1 rto 4 mem 240
retrans retry #1 rto 4 mem 383
retrans retry #1 rto 4 mem 285
retrans retry #1 rto 4 mem 405
retrans retry #1 rto 4 mem 427
retrans retry #1 rto 4 mem 547
retrans retry #1 rto 4 mem 547
retrans retry #2 rto 8 mem 405
retrans retry #1 rto 4 mem 525
retrans retry #2 rto 8 mem 525
retrans retry #1 rto 4 mem 645
retrans retry #1 rto 4 mem 480
retrans retry #2 rto 8 mem 600
retrans retry #1 rto 4 mem 720
retrans retry #1 rto 4 mem 720
retrans retry #1 rto 4 mem 840
retrans retry #1 rto 4 mem 960
retrans retry #1 rto 4 mem 1080
retrans retry #2 rto 8 mem 1200
retrans retry #1 rto 4 mem 1320
retrans retry #1 rto 4 mem 1440
retrans retry #2 rto 8 mem 1560
retrans retry #1 rto 4 mem 1680
retrans retry #1 rto 4 mem 1800
retrans retry #1 rto 4 mem 1920
retrans retry #1 rto 4 mem 2040
tcp: RETRANS limit, timeruse 18
retrans retry #1 rto 4 mem 480
retrans retry #1 rto 4 mem 600
retrans retry #1 rto 4 mem 720
retrans retry #1 rto 4 mem 840
retrans retry #1 rto 4 mem 960
retrans retry #1 rto 4 mem 1080
retrans retry #1 rto 4 mem 1200
retrans retry #1 rto 4 mem 1320
retrans retry #1 rto 4 mem 1440
retrans retry #1 rto 4 mem 1560
retrans retry #1 rto 4 mem 1680
tcp: RST received, removing retrans packets
ktcp: panic in write
```

The active session was an incoming telnet (to elks) listing a long stream of text (hd output). At the point of the RST, the telnet client (on macOS) disconnected: 
```
0019b0: 8b 4e fc e9 74 ff 56 57  55 89 e5 1e 8b 7e 0e e8  .N..t.VWU....~..
0019c0: 92 f7 8b 75 0c f7 c6 80  00 74 07 57 e8 c5 f5 83  ...u.....t.W....
0019d0: c4 0Connection closed by foreign host.
➜  elks git:(master) ✗ 
➜  elks git:(master) ✗ 
```

—Mellvik

ghaerr:
Thanks. There are at least three known problems that you are experiencing above, which we may need to try to break down further:

1. ktcp doesn't shutdown a TCP connection properly on receiving an RST, so after that happens, ktcp will be toast. (Same on sending an RST).

2. Once the "retrans limit" message is displayed, ktcp is usually in trouble. You'll notice that the alloc memory is gradually increasing on each retrans retry. The limit was added months ago to stop ktcp from running out of memory and forced dying. I think the issue you are having is that ktcp lost an ACK packet, but keeps on sending new packets, and each of them gets a retry since their ACK doesn't match. An immediate temp fix for this would be to limit sending while not having  received an ACK on previous packets.

3. There is no code in ktcp for limiting packets on some sending window size.
4. ktcp ignores all sliding window requests.
5. Once ktcp has to drop a packet, it gets very confused and can't recover.

There's still some serious work to be done to debug ktcp problems in all these areas. It is possible the reason this whole scenario happens is a lost ethernet packet, which happened as the result of a collision or other reasons.

PR #693, opened 2020-08-11
[kernel] Implement boot option parsing in /bootopts

ghaerr:
Set a variety of boot options in /bootopts file, without requiring kernel recompilation.

This PR adds Linux 2.0-style command line boot options stored in the root file /bootopts, and parsed at boot time to enable a variety of configurations without having to recompile anything. Very useful for debugging and quick configuration changes.

Allows setting the console to be serial (or any other device, such as virtual Alt2 /dev/tty2), root mount device (FAT and Minix), read-only/read-write, alternative startup process & args, init run level, network start, and set environment variables for the initial startup process, such as a shell or shell script.

Here's an example of the distribution /bootopts file, with everything commented out:
```
##      
#console=ttyS0 root=fd0 net=eth 3 # condensed
#init=/bin/init 3   # multiuser serial
#init=/bin/sh       # singleuser shell
#console=ttyS0      # serial console
#console=tty2       # alt2 console
#root=bda1          # hd partition 1
#ro                 # read-only root
#net=slip
```
For validation, the /bootopts file must start with ## and file size less be than 256 bytes.
\# = comment, options are space separated, multiple lines ok.
Devices can be specified by /dev/ prefix or without.
Non-reserved options without '=' inside them are treated as arguments to /bin/init or init=, and options with '=' are passed as environment variables to the init (or other init=) process.

Here's some more examples, hopefully self-explanatory:
```
init=/bin/init 3 or just 3
init=/bin/sh HOME=/ TERM=ansi
console=ttyS0, console=/dev/tty2, etc
console=tty1
root=fd1, root=/dev/fd1, root=bda[1-4], bdb[1-4], bdc[1-4]
net=eth, net=slip, net=cslip  starts predefined network configurations
ro, rw
```

To use, edit the file elkscmd/rootfs_template/bootopts and run "make" to create another image, or edit /bootopts when running ELKS and reboot.

The entire process is configured by selecting "Boot Options in /bootopts" in the Kernel section of `make menuconfig`, and is by default ON. Kernel code segment usage: 768 bytes, total 1392 bytes.

Other changes in this PR, noticed while getting it running:
- Pass startup parameter to /bin/init to override /etc/inittab initial run level.
- /etc/rc.d/rc.sys uses "net=" environment variable to start any of several predefined network configurations at boot.
- Added /dev/ttyp0 and /dev/pty0 to FAT filesystem for networking.
- Fixed libc `atoi` bug.

Bugs and known problems:
- Mounting root read-only fails, as ELKS fails the /dev/console open for read-write.
- Mounting a root FAT filesystem results in the inability to use /dev/ttyS0 (serial) as all filename characters are converted to lower case. We may have to rename ttyS0 to ttys0 system-wide in order to fix this.


Mellvik:
Again thanks @ghaerr. The first round of bootopts w/serial console changed my round trip time for new kernels dramatically. This round will effectively eliminate many of them, making 'remote debugging' physical hardware much more efficient.
BTW (Unrelated) - I don't think it has been mentioned officially: the urlget command works well on elks, making it possible to download any binary file (kernel, commands or a full floppy for that matter, etc.)

A few questions:
- How are multiple occurences of the same option (or env variable) handled? Will elks try one after the other (say root=) and select the first that works (that would be extremely valuable)? Environment variable: First or last?
- Several concurrent consoles (I believe Linux supports that but possibly not Linux 2.0)?
- I'm particularly interested in the net= option, will several occurences add up, (say) eth to slip to csclip?

And finally, there may be a bug (or oversight) in init: It has been  a while since i tested this – if the run level specified on the command line does not exist (try init 7), init loops.


Can't wait to (ab)use this!

--Mellvik

ghaerr:
> the urlget command works well on elks, making it possible to download any binary file (kernel, commands or a full floppy for that matter, etc.)

That's good news, can you give an example of how what client and server commands are used to transfer files both ways?


> How are multiple occurences of the same option (or env variable) handled? Will elks try one after the other (say root=) and select the first that works (that would be extremely valuable)? 

No, the boot options are all parsed very early in the boot process and stored in variables, with the exception of environment strings. The (last) stored value would then be used, typically overriding the default compiled-in value.

> Environment variable: First or last?

The environment variables are put in the environment array in the order specified with no checking. What happens within an application depends on the application - for instance /bin/sh will use the last value, whereas /bin/sash keeps them all, and will use the first value since that's what libc getenv does. Note, environment variables are only usable by the init (or other init=) process. When /bin/login runs, the environment is reset. 

 > Several concurrent consoles (I believe Linux supports that but possibly not Linux 2.0)?

ELKS only has the notion of a single console, as kernel printk and /dev/console go to the same place. I have had the idea of changing the console at runtime, however, which is still not implemented.

 I'm particularly interested in the net= option, will several occurences add up, (say) eth to slip to csclip? 

The process is extremely simple - a 'net=xxx' value in /bootopts just sets a 'net=xxx' environment variable, which the shell script /etc/rc.d/rc.sys uses to run the preset "net start xxx" command. See /etc/rc.d/rc.sys for details or expansion ideas.
The idea was to quickly allow changes from say, a serial-login configuration using getty's using run level 3 to an network over ethernet and/or network over slip/cslip with run level 1 without recompiling, all by using a single line in /bootopts.

Specifying environment variables multiple times with the same name is basically an error, but as said above it depends on how an application interprets the environment array as to what happens.

Since the boot options parsing code is in the kernel, where we're short of space and only used once, its size is kept as small as possible while maximizing functionality.

>a bug (or oversight) in init: It has been a while since i tested this – if the run level specified on the command line does not exist (try init 7), init loops.

Noted, thanks.

PR #694, opened 2020-08-11
[kernel] Update NE2K driver to use simpler wait routines

ghaerr:
@Mellvik,

Here is the driver updated to use the faster and simpler wait routines, which are also being used in the ring buffer code.
The older wait_event code method ends up testing the condition (e.g. status() != RX_READY) up to three times before waiting, while the improved approach only tests once.

Most of this shouldn't make a material difference other than being cleaner. I did notice that the ring buffer overflow also tried to wake up any reader(s) waiting, which I commented out. We need more discussion on what to do on buffer overflow, because any routine called during that interrupt will cause problems if a kernel process happens to be in the read or write routine, which is likely what's causing problems when received packets come in very fast.

In general, the ONLY thing we should do on RX or TX interrupt is wake up any waiting process. However, since the interrupt handler calls a get status routine which selects page 0 and writes registers, we're pretty much guaranteed to have problems unless the rest of the driver is protected against interrupt entry by delaying the acknowledgement or turning off interrupts.

### DIFF
--- a/elks/arch/i86/drivers/net/eth-main.c
+++ b/elks/arch/i86/drivers/net/eth-main.c
@@ -18,6 +18,8 @@
 
 
 // Static data
+struct wait_queue rxwait;
+struct wait_queue txwait;
 
 static byte_t eth_inuse = 0;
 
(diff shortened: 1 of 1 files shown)


Mellvik:
Thanks @ghaerr,

I’ll include this before I continue the testing this weekend. As to the buffer overflow handling, our discussion yesterday made it clear that it must be handled differently. So I decided to move it back into the read (get_packet) routine. At the end of a packet read a check of the interrupt status reg will determine if there is an overrun, which will be handled right there. That will delay the read return somewhat, but will also ensure there is no attempted reads during the NIC reset.

I’m also inclined to disable tx interrupts entirely as you suggested, since it doesn’t do any good as long as we’re running (almost) 'interrupt-less’.

More can be said about this, but I see there are more messages waiting, so more later.

—Helge

PR #701, opened 2020-08-14
Fix bug: undeclared 'init_command'

marcin-laszewski:
if CONFIG_BOOTOPS is undefined.
Is it correct solution? I'm not sure.


ghaerr:
Thanks @marcin-laszewski!

PR #702, opened 2020-08-14
New 'debugmem' replaces 'debugmem[1-9]'

marcin-laszewski:


ghaerr:
Thank you. 

BTW, do you know why these last two PR's consist of 27 commits each? I'm using squash and merge to make them a single commit, but don't know/understand git well enough to know why so many merges are appearing for these small changes.

ssieb:
It looks like he's doing merges instead of rebasing from the master repo or something like that.  You can see it if you look at the network graph.  His tree should be a branch off the top of this repo, but instead it's a long line with a series of pulls in from this repo.  Open https://github.com/jbruchon/elks/network and compare his tree with all the other branches.  When he makes a pull request, it includes all those merge commits since the last rebase from the origin because those aren't part of the origin tree.

ssieb:
A "git pull" from the upstream repo is all that should be needed to keep your master branch up-to-date.  Make sure it says it's doing a fast-forward update and no mention of merging.  Any sub branches should then use "git rebase master".  Once you end up with a merge instead of a rebase, it's tricky to get back.

ssieb:
I removed the incorrect information about resetting from the previous comment.  I did some testing and here's the real fix. :-)
On your master branch do `git reset --hard d052b846c44e35fd8d1f28ad1537847e50273432`.  That will go back to the latest shared commit.  Then do `git pull upstream master` where `upstream` is whatever you called the remote pointing to the main repo.  Then you'll need a `git push -f` to get this changed tree into the github repo.  If you have any other branches you're working on, you should do a `git rebase master` on them to fix their history and they might need force pushes as well.

ghaerr:
Hello @ssieb,

Thanks for your help on @marcin-laszewski's git repo!

Depending on whether his tree can be easily fixed up, I can still do squash-merges with no effect to this master, correct?

ssieb:
As long as he has all the changes in his tree, a squash and merge from it will still be correct.  But if he follows those instructions, it will be fixed.  I cloned his tree and tested it myself.

PR #703, opened 2020-08-15
[LD scripts] allow small-model a.out's to (maybe) have far text

tkchia:
This change to `elks-small.ld` allows small-model programs to "break out" of the small memory model a bit, and place some of their function code in a far text segment outside the normal text segment (https://github.com/tkchia/gcc-ia16/issues/63).

E.g. one can now do this:
```
$ cat test.c
#include <stdio.h>

static __far __attribute__ ((far_section, noinline)) int
foo (void)
{
  printf ("Hello world!\n");
  return 0;
}

int
main (void)
{
  return foo ();
}
$ ia16-elf-gcc -melks -O3 test.c -o test
```

### DIFF
--- a/elks/elks-small.ld
+++ b/elks/elks-small.ld
@@ -7,6 +7,16 @@ SECTIONS {
 		*(.text .text.*)
 		. = ALIGN(0x10);
 	}
+	.fartext 0 : AT(0x10000) {
+		/*
+		 * Some small-model programs might want to break out of the
+		 * small memory model a bit --- and choose to place some of
+		 * their code in a far segment outside the default text
+		 * segment.  Allow this.
+		 */
+		*(.fartext .fartext$ .fartext.*)
+		. = ALIGN(0x10);
+	}
 	.data 0 : AT(0x20000) {
 		*(.nildata .nildata.*)
 		*(.rodata .rodata.*)


ghaerr:
Hello @tkchia,

Thank you for making this change!

So I completely understand it, does the `far_section` specified in the attributes in the above example default to automatically generate code into the `.fartext` segment, which is then addressed in the linker script?

What about the example posted in https://github.com/tkchia/gcc-ia16/issues/63#issuecomment-674380340 where `section(".fartext")` was used - would that only be necessary if a differently-named far section was desired?

If additional, separately-named far sections are specified, could they be be grouped and handled similarly by a modified linker script for that named section? I am thinking about initialization sections (for the kernel) that would be discarded after execution.

And finally, I assume that the segment relocations necessary for this (for user mode applications only at this point), are handled in the sys_exec function when CONFIG_EXEC_MMODEL is set, right?

tkchia:
Hello @ghaerr,

> So I completely understand it, does the `far_section` specified in the attributes in the above example default to automatically generate code into the `.fartext` segment, which is then addressed in the linker script?

`far_section` (without a `section` attribute) will output the code into _a_ segment with a name starting with `.fartext.` ... .  (The linker script as written will ultimately combine all the `.fartext`, `.fartext$`, and `.fartext` ... sections into a single section though.)

> If additional, separately-named far sections are specified, could they be be grouped and handled similarly by a modified linker script for that named section? I am thinking about initialization sections (for the kernel) that would be discarded after execution.

This is a bit hard to do at the moment...  To keep things simple (for now), and bearing in mind the existing limitations of `a.out`, I had only added support for one single far text section to the ELKS `a.out` format.  Do you think this needs to be changed?

(Even for now though, I guess you can order the input sections in such a way that all the initialization code ends up at the tail end of the output `.fartext` segment --- perhaps using `EXCLUDE_FILE` and other linker script tricks.  The tail end of the segment can then be jettisoned at runtime.)

> And finally, I assume that the segment relocations necessary for this (for user mode applications only at this point), are handled in the sys_exec function when CONFIG_EXEC_MMODEL is set, right?

That is so.

Thank you!

ghaerr:
> `far_section` (without a `section` attribute) will output the code into _a_ segment with a name starting with `.fartext.` ... . (The linker script as written will ultimately combine all the `.fartext`, `.fartext$`, and `.fartext` ... sections into a single section though.)

Got it. So either method will work fine since the far code will all be combined into a single additional segment. When an init section is desired (and implemented), a section attribute with a separate name could be used, then using some of the linker tricks you've mentioned to get that code linked at the tail of addresses generated, then jettisoned after execution.

> 
> > If additional, separately-named far sections are specified, could they be be grouped and handled similarly by a modified linker script for that named section? I am thinking about initialization sections (for the kernel) that would be discarded after execution.
> 
> This is a bit hard to do at the moment... To keep things simple (for now), and bearing in mind the existing limitations of `a.out`, I had only added support for one single far text section to the ELKS `a.out` format. Do you think this needs to be changed?

No, I don't think this needs changing, certainly not now. There's a bit of work to do to get the kernel executing any far text section code, and, except for a possible 'init' section, they're fine all be combined together anyways.

I'll try compiling up a kernel with a single far text function that calls a near function, and look at the code generated; then possibly take a pass at relocating the far segment and calling the function... :)

ghaerr:
Hello @tkchia,

I just added a function `foo` declared below to the `start_kernel` function in elks/init/main.c, after the idle task while loop, so that it won't get called, but nonetheless included in the kernel. I was thinking of seeing whether the current kernel might load while skipping the relocations for an uncalled procedure.

```
void __far __attribute__ ((far_section, noinline)) foo(void);
```

The kernel compile fails with:
```
init/main.c: In function ‘foo’:
init/main.c:117:1: error: cannot create ‘__far’ function or static storage variable
 foo(void)
 ^~~
init/main.c:117:1: note: use ‘-msegment-relocation-stuff’ to allow this
```
Are there other compiler options than the above you might recommend specifying? I thought I'd ask first. The user-mode option uses just -melks, while the kernel uses
```
CROSS_CFLAGS = -ffreestanding -fno-inline -melks -mcmodel=small -mno-segment-relocation-stuff
```
Was the -mno-segment-relocation-stuff added for protection only, or for other reasons required?

tkchia:
Hello @ghaerr,

> Was the -mno-segment-relocation-stuff added for protection only, or for other reasons required?

It is there just for protection.  You can remove it if you would like to try out stuff.

Thank you!

ghaerr:
Hello @tkchia,

FYI - After changing -mno-segment-relocation to -msegment-relocation in elks/Makefile, and adding this PR's change to elks-small.ld, function `foo` appears being added properly to the kernel (verified through 'system.map').

Booting that kernel fails immediately after setup.S gains control with the message "Bad system header!", so setup.S (which I have not yet looked at) apparently checks for relocation size and fails if nonzero. 

I'm wondering if that section of setup.S might better be rewritten/refactored as a C procedure? What are your thoughts on that? With some thought, some of the relocation code in sys_exec could be shared with the setup.S code... ?

And, go ahead and commit this PR, thank you!

tkchia:
Hello @ghaerr,

> I'm wondering if that section of setup.S might better be rewritten/refactored as a C procedure? What are your thoughts on that? With some thought, some of the relocation code in sys_exec could be shared with the setup.S code... ?

I think this might be worth doing. :-)

> And, go ahead and commit this PR, thank you!

Thanks!

PR #713, opened 2020-08-18
[networking] Add standard network headers to libc

ghaerr:
Add standard network libc headers netinet/in.h, arpa/inet.h, netdb.h.
Remove kernel headers from all network application source.
Add in_ntoa, in_aton, in_gethostbyname to libc (non-standard, required by ktcp).

Ktcp can now use /etc/host names/aliases for startup arguments rather than ip address.
urlget, tinyirc and nslookup can now take hostname specified in /etc/host as well as ip address.


ghaerr:
Hello @tkchia,

While reorganizing the libc header files for this commit, and specifically removing libc/include/linxumt/arpa/inet.h, I found a problem in the libc "make clean" which I thought to bring to your attention:

An ELKS "make clean" (or libc "make clean") doesn't remove the copied libc header files in the cross/ia16-elf/lib/elkslibc/include directory. Since those files are not part of the git repo, I had to manually remove .../linuxmt/arpa/inet.h, as several not-yet-modified source files used it and compiled, when they should not have, since I removed the master file in libc/include/linuxmt/arpa/inet.h.

I didn't modify libc/Makefile in this PR because I didn't want to break anything, but perhaps a "make clean" in libc should remove the entire cross/ia16-elf/lib/elkslibc directory?

You'll be able to see this just after I commit this PR. Nothing has to be done, but the removed inet.h file will remain in other developer's cross directory, possibly causing problems later.

Thank you!

tkchia:
Hello @ghaerr,

Thanks for the report.  Hmm... the removal should probably be done with a `make` ,,, `uninstall` from the `libc/Makefile`, since the header file was installed via `make` ... `install` from there.  I will need to check why the makefile is not doing that.

Thank you!

tkchia:
Hello @ghaerr,

OK --- it turns out this is because `make` ... `uninstall` in `libc/Makefile` was trying to only "surgically" remove the header files that it knew that it installed.  I did this to avoid deleting header files that might happen to be installed from elsewhere, i.e. not `elks-libc`.

The header files installed by `libc/Makefile` originally included `<linuxmt/arpa/inet.h>`, but after your change, `libc/include/linxumt/arpa/inet.h` is no more --- and as a result `libc/Makefile` no longer knows it needs to uninstall `<linuxmt/arpa/inet.h>` (from the `cross/` tree).

I have sent in a PR which should fix this problem.

Thank you!

PR #714, opened 2020-08-18
[libc] `make uninstall' now zaps all headers in the target include subdirectories

tkchia:
This allows `make uninstall` to clean away obsolete internal header files such as `<linuxmt/arpa/inet.h>` --- see discussion at https://github.com/jbruchon/elks/pull/713 .

### DIFF
--- a/libc/Makefile
+++ b/libc/Makefile
@@ -166,12 +166,20 @@ uninstall:
 	      $(DESTDIR)/ia16-elf/lib/$(MULTISUBDIR)/elks-small.ld \
 	      $(DESTDIR)/ia16-elf/lib/$(MULTISUBDIR)/elks-tiny.ld \
 	      $(MULTIINCDIR)/autoconf.h \
-	      $(patsubst include/%,$(MULTIINCDIR)/%, \
-		  $(wildcard include/*.h include/asm/*.h include/sys/*.h \
-			     include/arpa/*.h include/netinet/*.h)) \
-	      $(patsubst $(TOPDIR)/elks/include/%,$(MULTIINCDIR)/%, \
-		  $(wildcard $(TOPDIR)/elks/include/arch/*.h \
-			     $(TOPDIR)/elks/include/linuxmt/*.h))
+	      $(patsubst include/%,$(MULTIINCDIR)/%,$(wildcard include/*.h))
+	# Assume that all <asm/*>, <sys/*>, <arpa/*>, <netinet/*>, <arch/*>,
+	# & <linuxmt/*> header files were installed from here (elks-libc) &
+	# not from elsewhere.
+	#
+	# This also allows us to clean away obsolete internal header files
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

It seems that this modification performs pretty much the same as previously - that is, when "make uninstall" is run, the directories specified in libc/Makefile are removed. The previous version worked similarly, using wildcards on specified directories, which removed header files, although not directories. Both of which are fine - however, neither solves the problem that when a "make clean" is done from either the top level or from libc, header files newly removed in libc are not removed in the gcc duplicate storage area. That is, when a PR is committed that changes header files in libc, users will not know to run "make uninstall" in libc, essentially causing header synchronization errors.

Thank you!

tkchia:
Hello @ghaerr,

A `make clean` at the top level _will_ do a `make` ... `uninstall` in the `libc/` subtree, and so should clean away the installed header files.

But you have a point: a simple `make all` at the top level may not clean away the old include files.  I think a good way to fix this would be to get `make install` in `libc/` to also clean away any old header files in the installation (`cross/`) subdirectory, before installing updated header files.  I have pushed a new commit for this.

I am not sure if it is a good idea to have `make clean` in `libc/` touch anything outside the build directories.

Thank you!

ghaerr:
Hello @tkchia,

> A make clean at the top level will do a make ... uninstall in the libc/ subtree, and so should clean away the installed header files.

Ok, I didn't see that.

 >  I think a good way to fix this would be to get make install in libc/ to also clean away any old header files in the installation (cross/) subdirectory, before installing updated header files. I have pushed a new commit for this.

Allright, I will follow that in the future. For developers, doing a "make clean" at the top level will eliminate all issues, which should be standard practice anyways after pulling new code.

Thank you, Ok with me for commit.

tkchia:
@ghaerr: thanks!

Mellvik:
This may not be related, but after pulling in the most recent commits, I can no longer `make` after a `make clean`:
```
ia16-elf-gcc -ffreestanding -fno-inline -melks -mtune=i8086 -mcmodel=small -mno-segment-relocation-stuff -I/Users/helge/src/mellvik/elks/include -I/Users/helge/src/mellvik/elks/libc/include -I/Users/helge/src/mellvik/elks/elks/include -D__LIBC__ -D__LIBC_VER__='"elks-0.3.0"' -D__HAS_NO_FLOATS__ -Wall -Os -melks-libc -c -o in_aton.o /Users/helge/src/mellvik/libc/net/in_aton.c
/Users/helge/src/mellvik/libc/net/in_aton.c:1:23: fatal error: arpa/inet.h: No such file or directory
 #include <arpa/inet.h>
                       ^
compilation terminated.
```

--Mellvik

Mellvik:
[Correction] Please disregard previous message, my directory structure was hosed.

--Mellvik

Issue #705, opened 2020-08-15
Discussion on far procedures in the kernel


ghaerr:
Hello @tkchia,

I opened up a new issue to keep from straying too far from the point of PR #703.  

I have analyzed the code produced by gcc using the `far_section` attribute applied to C procedures. Unfortunately, for far procedures calling near procedures, as you mentioned, the inefficiency is such it cannot likely be used in the kernel. I hadn't thought of the problem of having to rewrite the stack parameters due to their offsets being incorrect.

It seems it might be far faster, both in execution and programmer time, to just compile and port the asm files to medium model, than trying to use a mixed model approach. However, before making that decision, I thought to ask a few more questions after having looked at the code produced.

Here's the far call thunk source (for easy reference):
```
        /* These routines must go into the near text segment, if at all.  */
        .text

        .global __ia16_far_call_thunk.bx.si
        .type   __ia16_far_call_thunk.bx.si, @function
        .global __ia16_far_call_thunk.bx.di
        .type   __ia16_far_call_thunk.bx.di, @function
__ia16_far_call_thunk.bx.si:
        pushw   %bp                     /* set up stack frame */
        movw    %sp,    %bp
        pushw   %bx                     /* preserve %bx and %cx */
        pushw   %cx
        movw    %si,    %cx             /* move stack args. size to %cx */
        jmp     0f
__ia16_far_call_thunk.bx.di:
        pushw   %bp
        movw    %sp,    %bp
        pushw   %bx
        pushw   %cx
        movw    %di,    %cx
0:                                      /* re-push stack arguments */
        jcxz    2f
        leaw    4(%bp), %bx
        addw    %cx,    %bx
        shrw    %cx
1:
        pushw   %ss:(%bx)
        decw    %bx
        decw    %bx
        loop    1b
2:
        movw    -4(%bp), %cx            /* reload %cx for callee */
        callw   *-2(%bp)                /* call function */
        movw    %bp,    %sp             /* tear down stack frame and return */
        popw    %bp
        lretw
```
And the kernel C test routines:
```
void nearproc(void)
{           
    boot_console = 0;           
}  
            
void __far __attribute__ ((far_section, noinline)) farproc(void)
{  
    nearproc();
}
```
And the generated code:
```
00000000 <farproc>:
   0:   57                      push   %di
   1:   bb 00 00                mov    $0x0,%bx
            2: R_386_16 nearproc
   4:   31 ff                   xor    %di,%di
[...]

PR #715, opened 2020-08-18
[tools] [boot] bump gcc-ia16 version & use new gcc-ia16 feature

tkchia:
`gcc-ia16` now uses more efficient code for far → near function calls.  This should make it easier to try out "mixed memory model" programming (https://github.com/jbruchon/elks/issues/705).

While at it, I also slightly updated the Minix filesystem bootloader code to use a recently-introduced `gcc-ia16` built-in function.


ghaerr:
Hello @tkchia,

Does _builtin_ia16_near_data_segment require an updated compiler chain? If so, I would like to hold/delay this commit (or at least that portion of it) for a bit, as I have been expecting a new ethernet driver from @Mellvik very shortly and don't want to introduce the need for a new tools build for such a small change. At the moment, there is lots of PR activity and it would be nice to desync the absolute requirement to update the tool chain, especially if more changes may be forthcoming for some of the discussion about far text segments.

Thank you!

tkchia:
Hello @ghaerr,

OK.  Maybe I will withdraw this PR and concentrate on other changes first.

Thank you!

PR #716, opened 2020-08-20
2nd round of updates to ne2k ethernet driver.

Mellvik:
Includes updates accidentally left out of the previous PR.
Plus:
- Buffer Overrun code completed and well tested.
- All outstanding (known) problems resolved.
- Added IOCTLS to get stats and get/set  the # of packet to discard when overrun happens. Meaningful to adapt to slow peripherals and long I/O waits.
- Replaced and simplified old NIC register page selection code which turned out to cause problems under high traffic loads.
- Tested on two different NIC types and QEMU.

Testing with the most recent updates of kernel and ktcp code is very encouraging.

--Mellvik

### DIFF
--- a/elks/arch/i86/drivers/net/eth-main.c
+++ b/elks/arch/i86/drivers/net/eth-main.c
@@ -28,6 +28,7 @@ static byte_t mac_addr [6]; // Current MAC address, from HW or set
 
 static byte_t recv_buf [MAX_PACKET_ETH+4];
 static byte_t send_buf [MAX_PACKET_ETH];
+extern word_t _ne2k_skip_cnt;
 
 // Get packet
 
@@ -76,7 +77,6 @@ static size_t eth_read (struct inode * inode, struct file * filp,
 	return res;
 }
 
-
 // Put packet
 
 static size_t eth_write (struct inode * inode, struct file * file,
@@ -93,7 +93,7 @@ static size_t eth_write (struct inode * inode, struct file * file,
 				break;
 			}
 			do_wait();
-			if (current->signal) {
+			if(current->signal) {
 				res = -EINTR;
 				break;
 			}
@@ -157,26 +157,32 @@ int eth_select (struct inode * inode, struct file * filp, int sel_type)
 
 static void ne2k_int (int irq, struct pt_regs * regs, void * dev_id)
 {
-	word_t stat;
+	word_t stat, page;
 
 	stat = ne2k_int_stat ();
-        //word_t page = ne2k_getpage();
+        //page = ne2k_getpage();
         //printk("$%02x$B%02x$", (page>>8)&0xff, page&0xff);
+	//printk("/%02X",stat&0xff);
 
         if (stat & NE2K_STAT_OF) {
-		printk("Warning: NIC receive buffer overflow\n");
+		word_t skip_buf[2] = {};	// the low level routine needs this one.
+
+		_ne2k_skip_cnt = 0;	// Zero is the default, discard entire buffer
+					// On a floppy based system, anything else is useless.
+		printk("Warning: NIC receive buffer overflow, skipping ");
+		if (_ne2k_skip_cnt) 
+			printk("%d packets.\n", _ne2k_skip_cnt);
+		else
+			printk("all packets.\n");
 		//page = ne2k_getpage();
 		//printk("/C%02x|B%02x/ ", (page>>8)&0xff, page&0xff);
-		// FIXME: probabluy shouldn't wake up reader if no packet ready,
-		// and the check below will do so if STAT_RX bit is set anyways.
-		// Could handle the error here, except this code could be interrupting
-		// a process already in the read routine, so that will likely mess up
-		// that reader and/or the driver state.
-		//wake_up (&rxwait);
-		//ne2k_clr_oflow(recv_buf); // The reset procedure needs to read the
-                                          // last complete packet in the buffer.
-		// If recv_buf is busy, this will overwrite the contents.
-		// What do we do with the contents? As is, this is a discard..
+
+		// FIXME: This procedure is still being debugged
+		//wake_up ((struct wait_queue *) &rx_flag);
+
+		page = ne2k_clr_oflow(skip_buf); // Reset the nic, discard the given number of packets
+		//printk("/C%02x|B%02x/ ", (page>>8)&0xff, page&0xff);
+		//printk("/N%02x|S%02x/ ", (skip_buf[0]>>8)&0xff, skip_buf[0]&0xff);
 	}
 
 	if (stat & NE2K_STAT_RX) {
@@ -189,9 +195,13 @@ static void ne2k_int (int irq, struct pt_regs * regs, void * dev_id)
 		wake_up(&txwait);
 	}
 	if (stat & NE2K_STAT_RDC) {
-		printk("RDC intr.\n");
-		// debug only, but keep this code, the RDC interrupt should be disabled in
-		// the low level driver.
+		printk("Warning: RDC intr.\n");
+		// The RDC interrupt should be disabled in the low level driver.
+		// When real DMA transfer from NIC til system RAM is enabled, this is where
+		// we handle transfer completion.
+		// NOTICE: If we get here, a remote DMA transfer was aborted, probably a bug
+		// in the driver.
+		ne2k_rdc();
 	}
 }
 
@@ -219,6 +229,29 @@ static int eth_ioctl (struct inode * inode, struct file * file,
 		case IOCTL_ETH_ADDR_SET:
 			memcpy_fromfs (mac_addr, (char *) arg, 6);
 			ne2k_addr_set (mac_addr);
+			break;
+
+		case IOCTL_ETH_HWADDR_GET:
+			// Get the hardware address of the NIC,	which may be different
+			// from the currently programmed address. Be careful with this,
+			// it may interrupt ongoing send/receives.
+			// FIXME: Needs more testing.
+			ne2k_get_hw_addr((word_t *) arg);
+			break;
+
+		//case IOCTL_ETH_GETSTAT:
+			// Get statistics from the NIC hardware (error counts etc.)
+			// NOt implemented
+
+		case IOCTL_ETH_OFWSKIP_SET:
+			// Set the number of packets to skip @ ring buffer overflow
+			_ne2k_skip_cnt = arg;
+			//printk("ne2k: OFW skip-cnt is now %d.\n", _ne2k_skip_cnt);
+			break;
+
+		case IOCTL_ETH_OFWSKIP_GET:
+			// Get the current overflow skip counter
+			arg = _ne2k_skip_cnt;
 			break;
 
 		default:
--- a/elks/arch/i86/drivers/net/ne2k-mac.S
+++ b/elks/arch/i86/drivers/net/ne2k-mac.S
@@ -61,28 +61,10 @@ rx_last            = 0x80
 // --- not the right thing to do ...
 _ne2k_next_pk:
 	.word 0	// being used as byte ...
-	
 
-//-----------------------------------------------------------------------------
-// Select register page
-//-----------------------------------------------------------------------------
-
-// AL : page number (0 or 1)
-
-page_select:
-
-	mov     %al,%ah
-	and     $0x01,%ah
-	shl     $6,%ah
-
-	mov     $io_ne2k_command,%dx
-	//in      %dx,%al
-	//and     $0x3F,%al
-	mov	$0x22,%al
-	or      %ah,%al
-	out     %al,%dx
-
-	ret
+	.global _ne2k_skip_cnt
+_ne2k_skip_cnt:
+	.word 0	// # of packets to skip if buffer overrun, default is all (0)
 
 //-----------------------------------------------------------------------------
 // Set unicast address (aka MAC address)
@@ -100,10 +82,9 @@ ne2k_addr_set:
 
 	mov     4(%bp),%si
 
-	// select page 1
-
-	mov     $1,%al
-	call    page_select
+	mov     $io_ne2k_command,%dx
+	mov	$0x42,%al	// page 1
+	out	%al,%dx
 
 	// load MAC address
 
@@ -117,6 +98,10 @@ ems_loop:
 	out     %al,%dx
 	inc     %dx
 	loop    ems_loop
+
+	mov     $io_ne2k_command,%dx
+	mov	$0x02,%al	// back to pg 0
+	out	%al,%dx
 
 	pop     %si
 	pop     %bp
@@ -134,10 +119,7 @@ dma_init:
 	push    %ax
 	push    %dx
 
-	// select page 0
-
-	xor     %al,%al
-	call    page_select
+	// assume page 0
 
 	// set DMA start address
 
@@ -183,6 +165,7 @@ dma_write:
 	push    %dx
 	push    %si
 
+	cli		//Experimental
 	inc     %cx     // make byte count even
 	and     $0xfffe,%cx
 	call    dma_init
@@ -192,9 +175,9 @@ dma_write:
 	// start DMA write
 
 	mov     $io_ne2k_command,%dx
-	in      %dx,%al
-	and     $0xC7,%al
-	or      $0x10,%al  // 010b : write
+	//mov	$0x0a,%al	// per app note (P 1-17)
+	//out     %al,%dx
+	mov	$0x12,%al	// per app note (P 1-17)
 	out     %al,%dx
 
 	// I/O write loop
@@ -221,6 +204,7 @@ check_dma_w:
 	clc
 
 
+	sti		// Experimental
 	pop     %si
 	pop     %dx
 	pop     %cx
@@ -290,6 +274,7 @@ dma_read:
 	push    %dx
 	push    %di
 
+	cli		//Experimental
 	inc     %cx     // make byte count even
 	and     $0xfffe,%cx
 	call    dma_init
@@ -298,10 +283,10 @@ dma_read:
 	// start DMA read
 
 	mov     $io_ne2k_command,%dx
-	in	%dx,%al
-	and	$0xC7,%al
-	or	$0x08,%al	// 0x8 = read
-	//mov	$0x0a,%al	// 0ah per application note
+	//in	%dx,%al
+	//and	$0x07,%al	// ensure pg 0
+	//or	$0x08,%al	// 0x8 = read
+	mov	$0x0a,%al	// 0ah per application note
 	out     %al,%dx
 
 	// I/O read loop
@@ -336,10 +321,13 @@ check_dma_r:
 	mov     $0x40,%al       // reset ISR (RDC bit only)
 	out     %al,%dx
 
+	sti		//Experimental
+
 	pop     %di
 	pop     %dx
 	pop     %cx
 	pop     %ax
+
 	ret
 
 //
@@ -351,15 +339,17 @@ check_dma_r:
 	.global ne2k_getpage
 
 ne2k_getpage:
-	mov     $1,%al
-	call    page_select
+	mov	$0x42,%al	// page 1
+	mov	$io_ne2k_command,%dx
+	out	%al,%dx
 
 	mov     $io_ne2k_rx_put,%dx     // CURRENT
 	in      %dx,%al
 	mov     %al,%cl
 
-	xor     %al,%al
-	call    page_select
+	mov	$0x02,%al	// page 0
+	mov	$io_ne2k_command,%dx
+	out	%al,%dx
 
 	mov     $io_ne2k_rx_get,%dx     // BOUNDARY
 	in      %dx,%al
@@ -383,18 +373,21 @@ ne2k_rx_stat:
 
 	// get RX put pointer
 
-	mov     $1,%al
-	call    page_select
+	mov	$0x42,%al	// page 1
+	mov	$io_ne2k_command,%dx
+	out	%al,%dx
 
 	mov     $io_ne2k_rx_put,%dx
 	in      %dx,%al
 	mov     %al,%cl
 
+	mov	$0x02,%al	// back to page 0
+	mov	$io_ne2k_command,%dx
+	out	%al,%dx
+
 	// get RX get pointer
 
 	mov	_ne2k_next_pk,%al
-
-nrs_nowrap:
 
 	// check ring is not empty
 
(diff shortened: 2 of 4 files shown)


ghaerr:
Thank you very much @Mellvik for this rewritten driver and continued testing.

I can't (yet) test this, but am very encouraged by your recent testing.

I notice there is still quite a bit of commented out debugging statements. We now have a mechanism to leave debugging statements in the kernel in a clean way, if you want to leave them in long term. I suggest something like this within eth-main.c:
```
#define DEBUG_ETH        0 /* set =1 for debugging*/

#if DEBUG_ETH
#define debug_eth   printk
#else
#define debug_ethk(...)
#endif
```
We don't have a clean way (yet) for debug and other statements in .S code. Let me know if you think we need to invent one.

I'll commit this now and then review the code as a whole with more suggestions, if any.

Thanks!!!

Mellvik:
DEBUG_ETH: Yes, this is a great way of doing it. 

Certainly, not all the debug stuff is required long term. In particular the rather extensive NIC ring buffer tracing is now done. OTOH leaving one such statement in in order to remember how to do it may be OK. Memory (I'm not talking about RAM) is a fragile thing … :-)

Thanks.

—Mellvik

> 20. aug. 2020 kl. 17:32 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> Thank you very much @Mellvik <https://github.com/Mellvik> for this rewritten driver and continued testing.
> 
> I can't (yet) test this, but am very encouraged by your recent testing.
> 
> I notice there is still quite a bit of commented out debugging statements. We now have a mechanism to leave debugging statements in the kernel in a clean way, if you want to leave them in long term. I suggest something like this within eth-main.c:
> 
> #define DEBUG_ETH        0 /* set =1 for debugging*/
> 
> #if DEBUG_ETH
> #define debug_eth   printk
> #else
> #define debug_ethk(...)
> #endif
> We don't have a clean way (yet) for debug and other statements in .S code. Let me know if you think we need to invent one.
> 
> I'll commit this now and then review the code as a whole with more suggestions, if any.
> 
> Thanks!!!
>

Mellvik:
Apropos cleanup, @ghaerr:

eth_open blocks attempts to reopen the device. If a utility is to access the IOCTLs, how is to going to get to them?
Should multiple opens really be allowed and _open changed so that it initialises only on the first open?

—Mellvik

> 20. aug. 2020 kl. 17:48 skrev Helge Skrivervik <helge@mymayday.com>:
> 
> DEBUG_ETH: Yes, this is a great way of doing it. 
> 
> Certainly, not all the debug stuff is required long term. In particular the rather extensive NIC ring buffer tracing is now done. OTOH leaving one such statement in in order to remember how to do it may be OK. Memory (I'm not talking about RAM) is a fragile thing … :-)
> 
> Thanks.
> 
> —Mellvik
> 
>> 20. aug. 2020 kl. 17:32 skrev Gregory Haerr <notifications@github.com <mailto:notifications@github.com>>:
>> 
>> 
>> Thank you very much @Mellvik <https://github.com/Mellvik> for this rewritten driver and continued testing.
>> 
>> I can't (yet) test this, but am very encouraged by your recent testing.
>> 
>> I notice there is still quite a bit of commented out debugging statements. We now have a mechanism to leave debugging statements in the kernel in a clean way, if you want to leave them in long term. I suggest something like this within eth-main.c:
>> 
>> #define DEBUG_ETH        0 /* set =1 for debugging*/
>> 
>> #if DEBUG_ETH
>> #define debug_eth   printk
>> #else
>> #define debug_ethk(...)
>> #endif
>> We don't have a clean way (yet) for debug and other statements in .S code. Let me know if you think we need to invent one.
>> 
>> I'll commit this now and then review the code as a whole with more suggestions, if any.
>> 
>> Thanks!!!
>> 
>

Mellvik:
@ghaerr,
apropos netstt and arp messages:

how about this one (proposal):




—Mellvik

> 21. aug. 2020 kl. 11:09 skrev Helge Skrivervik <helge@mymayday.com>:
> 
> Apropos cleanup, @ghaerr:
> 
> eth_open blocks attempts to reopen the device. If a utility is to access the IOCTLs, how is to going to get to them?
> Should multiple opens really be allowed and _open changed so that it initialises only on the first open?
> 
> —Mellvik
> 
>> 20. aug. 2020 kl. 17:48 skrev Helge Skrivervik <helge@mymayday.com <mailto:helge@mymayday.com>>:
>> 
>> DEBUG_ETH: Yes, this is a great way of doing it. 
>> 
>> Certainly, not all the debug stuff is required long term. In particular the rather extensive NIC ring buffer tracing is now done. OTOH leaving one such statement in in order to remember how to do it may be OK. Memory (I'm not talking about RAM) is a fragile thing … :-)
>> 
>> Thanks.
>> 
>> —Mellvik
>> 
>>> 20. aug. 2020 kl. 17:32 skrev Gregory Haerr <notifications@github.com <mailto:notifications@github.com>>:
>>> 
>>> 
>>> Thank you very much @Mellvik <https://github.com/Mellvik> for this rewritten driver and continued testing.
>>> 
>>> I can't (yet) test this, but am very encouraged by your recent testing.
>>> 
>>> I notice there is still quite a bit of commented out debugging statements. We now have a mechanism to leave debugging statements in the kernel in a clean way, if you want to leave them in long term. I suggest something like this within eth-main.c:
>>> 
>>> #define DEBUG_ETH        0 /* set =1 for debugging*/
>>> 
>>> #if DEBUG_ETH
>>> #define debug_eth   printk
>>> #else
>>> #define debug_ethk(...)
>>> #endif
>>> We don't have a clean way (yet) for debug and other statements in .S code. Let me know if you think we need to invent one.
>>> 
>>> I'll commit this now and then review the code as a whole with more suggestions, if any.
>>> 
>>> Thanks!!!
>>> 
>> 
>

ghaerr:
> Should multiple opens really be allowed and _open changed so that it initialises only on the first open?

No, adding such code opens the door for more unneeded complexity. Since ktcp is the only application that uses /dev/eth, and any reporting would likely be in "netstat" (or something like "arp" which uses the same local socket to ktcp), ktcp can collect the information and pass it to netstat. There's only a single stat being maintained, right?

> apropos netstt and arp messages:
how about this one (proposal):

Nothing shown!

Mellvik:
> 
> Should multiple opens really be allowed and _open changed so that it initialises only on the first open?
> 
> No, adding such code opens the door for more unneeded complexity. Since ktcp is the only application that uses /dev/eth, and any reporting would likely be in "netstat" (or something like "arp" which uses the same local socket to ktcp), ktcp can collect the information and pass it to netstat. There's only a single stat being maintained, right?
> 
> 
I wan't really thinking about stats, but general access to (and the meaningfulness of) ioctls.
> apropos netstt and arp messages:
> how about this one (proposal):
> 
> Nothing shown!
> 
> 
I keep forgetting that gitbub filters pictures.
Will upload via web.

—M

Mellvik:
Skjermbilde 2020-08-21 kl  14 57 28

ghaerr:
(regarding netstat display of ARP activity):

I actually had them displayed that way originally, but changed it so that everything in the left column was "Received" and everything in the right column "Sent". Are you thinking that its far easier to read ARP information grouped vertically rather than horizontally?

Mellvik:
Re: open --
Actually, what was on my mind was a diag utility for ne2k, but I guess it makes sense that it would have to be used when ktcp was not active.

re: ARP display --
Well, I like you original idea then. It seems logical (to me) to pair up requests -> replies horizontally. Then again it's a matter of taste ...

--M

Mellvik:
@ghaerr,

re. cleanup and a comment you made in a different channel:

While cleaning up, I made some changes to the packet read (added optimisation I've been talking about before), so I'm running a new round of tests before the PR.

—Mellvik

ghaerr:
> Well, I like you original idea then. It seems logical (to me) to pair up requests -> replies horizontally. 

Prepare a PR with the modifcations, except please capitalize and spell out Sent and Received for each, to match the existing capitalization.

pawosm-arm:
I've managed to get to my old flat where my old TurboXT is stashed. As I mentioned times before, it's equipped with rare 8-bit ISA network card with RJ-45 connector. It's SMC 8003WC (ex. WD8003). In theory it's ne2k clone so I wanted to give this driver a try (note it's the first time ELKS managed to boot from CF Card on this machine!). This card uses jumperless configuration held in its ROM and it goes like that:

IRQ: 2
I/O ports base: 0x240
I/O memory base: 0xCCE00 (right after XT-IDE ROM!)

Under FreeDOS this card is controlled by a packet driver.

I've made the following changes to current master:
```
diff --git a/elks/arch/i86/drivers/net/ne2k-low.s b/elks/arch/i86/drivers/net/ne2k-low.s
index 1733f709..da52a9fd 100644
--- a/elks/arch/i86/drivers/net/ne2k-low.s
+++ b/elks/arch/i86/drivers/net/ne2k-low.s
@@ -7,7 +7,7 @@
 
 ; On Advantech SNMP-1000 SBC, the ethernet interrupt is INT0 (0Ch)
 
-int_vect    EQU $0C
+int_vect    EQU $0A ; NOT SURE ABOUT THIS ONE. IS IT REALLY USED AT ALL?
 
 
        .TEXT
diff --git a/elks/arch/i86/drivers/net/ne2k-phy.s b/elks/arch/i86/drivers/net/ne2k-phy.s
index acbd673f..3b55b70c 100644
--- a/elks/arch/i86/drivers/net/ne2k-phy.s
+++ b/elks/arch/i86/drivers/net/ne2k-phy.s
@@ -5,10 +5,10 @@
 
 ; TODO: move definitions to ne2k.h
 
-; I/O base @ 300h
+; I/O base @ 240h
 
-io_ne2k_mdio     EQU $314
-io_ne2k_gpio     EQU $317
+io_ne2k_mdio     EQU $254
+io_ne2k_gpio     EQU $257
 
 ; MDIO register
 
diff --git a/elks/include/arch/ports.h b/elks/include/arch/ports.h
index 604a8087..d8a294d8 100644
--- a/elks/include/arch/ports.h
+++ b/elks/include/arch/ports.h
@@ -54,7 +54,7 @@
 
 #ifdef CONFIG_ETH_NE2K
 //#define CONFIG_NEED_IRQ9
-#define CONFIG_NEED_IRQ12
+#define CONFIG_NEED_IRQ2
 #endif
 
 /* unused*/
@@ -99,8 +99,8 @@
 #define COM4_IRQ       2               /* unregistered unless COM4_PORT found*/
 
 /* ne2k, eth-main.c*/
-#define NE2K_IRQ       12
-#define NE2K_PORT      0x300
+#define NE2K_IRQ       2
+#define NE2K_PORT      0x240
 
 /* obsolete - experimental IDE hard drive, directhd.c (broken)*/
 #define HD1_PORT       0x1f0
```

Unfortunately, during boot process I could see that:

```
eth: NE2K at 0x240, irq 2, MAC 01:01:01:01:01:01
```

The MAC address was read wrong, so I guess it's not working.

ghaerr:
Hi @pawosm-arm,

That would be fun if we can get that card working. I'm going to defer to @Mellvik as to what he thinks the compatibility of that card is, he knows more about network hardware than I do.

With regards to your changes, funnily enough I was just talking to him about a couple of the files you've modified this morning. That is, both ne2k-phys.s and ne2k-low.s are not used in the current driver at all! He's going to remove in his next commit, which will be a cleanup of that and some other things.

The changes to port.h look correct, however.

@Mellvik also wrote the MAC address read routine, so I'll let him comment on that. There may be debug statements that can be turned on that will display more of what is happening during probe time or whether the MAC address read succeeded.

I had thought a while back that it could be interesting to write some user-level packet driver code that would allow ktcp to load a packet-driver based network card driver in user mode and run from there... but its probably a lot of work, and I'm not sure how much use such a thing would get.

Mellvik:
Sorry, Pawosm-arm,
I'm afraid you're out of luck here.
The ne2k driver will not work with a ne1k card, which is what you need on a xt.


--mellvik

ghaerr:
> note it's the first time ELKS managed to boot from CF Card on this machine!

That's great! Believe it or not, I hadn't ever got confirmation from anybody that the BIOS drive enhancements, added way earlier this year, actually worked :)

Mellvik:
@pawosm-arm,
Some elaborations on this issue:

The ne2k driver can easily be expanded to cover the n1k. However, the wd8003 is not - as far as I can tell - ne1k compatible, so it would not help in your case.

That said, there are linux drivers for the wd8003 from the early nineties, and having thw hw available, it may be a tempting challenge to get it to work?

--mellvikl

pawosm-arm:
> The ne2k driver can easily be expanded to cover the n1k.

Yup, e.g. QNX 4.25 has one driver for both n1k and n2k.

> However, the wd8003 is not - as far as I can tell - ne1k compatible, so it would not help in your case.

> That said, there are linux drivers for the wd8003 from the early nineties, and having thw hw available, it may be a tempting challenge to get it to work?

I'm thinking on pulling it out of this TurboXT and bring it with me to my new home and put it into Amstrad PC2086, so I could work on it. Is ELKS ne2k driver abstract enough to serve as a staple for other network drivers?

Mellvik:
> 
> However, the wd8003 is not - as far as I can tell - ne1k compatible, so it would not help in your case.
> 
> That said, there are linux drivers for the wd8003 from the early nineties, and having thw hw available, it may be a tempting challenge to get it to work?
> 
> I'm thinking on pulling it out of this TurboXT and bring it with me to my new home and put it into Amstrad PC2086, so I could work on it. Is ELKS ne2k driver abstract enough to serve as a staple for other network drivers?
> 
> 
I wouldn't bet on it - maybe conceptually. I'd recommend looking up an old linux driver for the WD card, or even better, find the specs for it. With either of those I'd be happy to provide an opinions as to what seems like a reasonable starting point.

—mellvik

ghaerr:
> Is ELKS ne2k driver abstract enough to serve as a staple for other network drivers?

Actually I believe it is, or can be slightly rearranged to be. The overall driver structure is in two parts - a .c file which will remain extremely similar for the next network driver, and a .S file, which essentially contains three routines, read, write and init.

The top-level C code ensures that the lower level assembly code is only entered when the NIC is ready to receive or transmit, and will handle the technical details of putting a process to sleep until a read or write is ready to execute. It also handles the integration of read/write/select/ioctl/init with the kernel, which will likely change very little or not at all with a new driver.

The lower level .S assembly code basically handles three entry points - read, write and init. Of course, they can be a bit complicated, such as the read routine having to handle NIC overflow, and both read and write setting up internal DMA. As @Mellvik suggests, if you know the specs for the card, or have looked at an old Linux driver for it, that should help speed up the process.

> I'm thinking on pulling it out of this TurboXT and bring it with me to my new home and put it into Amstrad PC2086, so I could work on it.

Yes, lets do it! I can write a Wiki article on how the upper level of a network driver works if that will help.

pawosm-arm:
> Actually I believe it is, or can be slightly rearranged to be. The overall driver structure is in two parts - a .c file which will remain extremely similar for the next network driver, and a .S file, which essentially contains three routines, read, write and init.

So you say that all of the `NE2K_`, `ne2k_` and `_ne2k_` prefixed stuff needs to be renamed to `ETH_`, `eth_` and `_eth_` respectively in the `eth-main.c` file?

pawosm-arm:
> So you say that all of the NE2K_, ne2k_ and _ne2k_ prefixed stuff needs to be renamed to ETH_, eth_ and _eth_ respectively in the eth-main.c file?

Answering myself, `eth-main.c` isn't abstract enough. Either it should be refactored so it could make calls to callbacks registered for lower level assembly routines (ne2k vs wd8003 or anything else), or every driver should provide its own implementation of eth-main.

ghaerr:
> So you say that all of the NE2K_, ne2k_ and _ne2k_ prefixed stuff needs to be renamed to ETH_, eth_ and _eth_ respectively in the eth-main.c file?

No, most of the names will stay the same in eth-main.c, although eth-main.c is being renamed ne2k.c in @Mellvik's next cleanup commit. 

The interface entry points to the ethernet driver will stay named eth_xxx, although they don't strictly have to be. They should all be static (currently they are not), including the "static struct file_operations eth_fops", which is only exposed by passing to register_chrdev.

But the reason the names might as well stay eth_xxx is that, unlike Linux, ELKS currently can only have a single ethernet driver linked in, as that driver explicitly manages /dev/eth. In Linux and other operating systems, there is no /dev/eth, the interfaces are enabled using ifconfig or similar, and ethernet probing routines are called, with all the drivers linked in.

> eth-main.c isn't abstract enough. Either it should be refactored so it could make calls to callbacks registered for lower level assembly routines (ne2k vs wd8003 or anything else), or every driver should provide its own implementation of eth-main.

The latter part of your statement would be true now for ELKS: each config-selectable ethernet driver would provide its own implementation of eth-main.c (renamed to better associate the NIC, just like we'll be doing with eth-main.c -> ne2k.c). 

I agree with your refactoring comment, for later - as there may still be a need slight modifications to the driver "top end". After a second driver is written, it should be possible to refactor the top end to be shared by both network drivers, and renamed ethernet.c, with yet another structure defining the lower-level entry points, called through a jump table.

pawosm-arm:
> That said, there are linux drivers for the wd8003 from the early nineties, and having thw hw available, it may be a tempting challenge to get it to work?

Although I'm on vacations for next week (with no access to the XT hardware), I've taken a look at the Linux 2.0 sources, and what I'm observing is that both ne2k and wd drivers in Linux make use of 8390 driver (and indeed, WD83C690 is the biggest chip on board this SMC8003WC card); it is commented as such:
```
  This is the chip-specific code for many 8390-based ethernet adaptors.
  This is not a complete driver, it must be combined with board-specific
  code such as ne.c, wd.c, 3c503.c, etc.
```
I wonder if ELKS's ne2k driver could have 8390-specific part extracted?

Mellvik:
Thank you @pawosm-arm,
This is interesting. Since you started this thread I've been looking for more info about the WD8xxx interfaces without much success. Now you seem to have located at least something - that would be useful.

If the WD8003 does indeed use the 8390, it is most likely NE1k compatible. And adding ne1k support to the ne2k driver - or create as modified copy, is not a big deal when you have hardware available for debugging.
> ﻿
>  This is the chip-specific code for many 8390-based ethernet adaptors.
> 
>   This is not a complete driver, it must be combined with board-specific
>   code such as ne.c, wd.c, 3c503.c, etc.
> I wonder if ELKS's ne2k driver could have 8390-specific part extracted?
> 
In fact, almost all of ne2k-asm.S is 8390 specific, and very little of ne2k.c is hw specific, the exception being the ioctls.
If you can point me to the source of this info, I'll have a look and give you an opinion. I have the feeling you may be very close to having something operational if the WD card has a 8390!

--Mellvik

pawosm-arm:
> If you can point me to the source of this info, I'll have a look and give you an opinion.

Back in the old days of Linux 2.0 it was an opening comment in the `8390.c` file. The file was later moved and renamed (to `lib8390.c`, so now it's not a driver, it's a library acting as a basis for the ne2k and wd/smc drivers), but the comment is still there: https://github.com/torvalds/linux/blob/master/drivers/net/ethernet/8390/lib8390.c

> I have the feeling you may be very close to having something operational if the WD card has a 8390!

Before I left, I managed to see 'little something' from this card:

```
        for (i = 0; i < 6; i++)
                printk(" %2.2X", inb(0x240 + 8U + i));
```
...where 0x240 is this card's I/O base as written into its PROM. This code has written the real MAC address of this card during ELKS bootup. The idea was taken from here: https://github.com/torvalds/linux/blob/master/drivers/net/ethernet/8390/wd.c#L193

IMHO, Linux driver does much too much during 'probe', ELKS tends to avoid autodetection and I/O ports, device memory, DMA channels and IRQ lines are either put into headers or into `.config`, so the `probe` function can be much shorter (or do literally nothing, as probe for ne2k). Also I'd limit the support to card I can physically test (8-bit 8003 card) resting the code for 16-bit 8013 card in peace.

Mellvik:
Thanks @pawosm-arm,

interesting reading indeed, and helpful in visualizing what/where the differences between the various cards using the same chip are.
-> Card Prom (with station address and some identification) and how to access it
-> How much and where (address) the ring buffer is located
-> bus width supported

There is a lot of card detection and adaption code in here. Interesting - and fortunately something we don't need: Most of the cards are history and our goal is much simpler: Get something running with THIS particular card, whatever THIS is. 
There is also probably enough guidance to help you to a function 8-bit ELKS wd-driver - based on the ne2k-driver we have now. 

I actually got a couple of ideas from the first link, such as allocating a second TX buffer on the card in order to speed up transmission. Right now I'm seeing a big speed difference in file transfers to and from ELKS. It will be interesting to see if double buffering will help. It's a modest driver change.

@ghaerr, any opinions on this?

I may not have mentioned this before: As long as telnet/telnetd is not touched, tcp (HTTP) file transfers are stable on ELKS both ways. Transferring (in) huge files (urlget) to /dev/null w/o a hitch. And out - importing /dev/bda1 to the local host (via curl), no problem but slow. Very encouraging.

—Mellvik

ghaerr:
Hello @pawosm-arm,

I have also read the Linux 2.0 8390 sources, but hadn't yet seen they were moved to a lib file in later versions.

> IMHO, Linux driver does much too much during 'probe', ELKS tends to avoid autodetection and I/O ports, device memory, DMA channels and IRQ lines are either put into headers or into .config, so the probe function can be much shorter (or do literally nothing, as probe for ne2k). Also I'd limit the support to card I can physically test (8-bit 8003 card) resting the code for 16-bit 8013 card in peace.

I think you're on the right track here. Almost all of the "top level" NIC code in the Linux 2.0 kernel is related to fancy probing and getting around bugs in various cards, and then appropriately connecting to the (lower level, now apparently library) code to actually drive the chip, using variations passed from the lop level.

I would encourage you to take @Mellvik's working NE2K code as a base, don't do any probing, and slowly bring the driver up to working, as you've already started to do reading the PROM. I haven't read the differences (if any) in the chipsets, but those would be worth concentrating on first.

I'll help you however I can. After you get even a basic driver up and running, we can look at how those differences should be integrated into ELKS, either as two separate drivers, or possibly using a core library such as Linux now uses.

ghaerr:
> I actually got a couple of ideas from the first link, such as allocating a second TX buffer on the card in order to speed up transmission. Right now I'm seeing a big speed difference in file transfers to and from ELKS. It will be interesting to see if double buffering will help. It's a modest driver change.

Given the single-threaded nature of ktcp and it being the only process reading and writing the NIC, I doubt this would help much during regular networking operations, although it could speed up special cases somewhat when there are a number of unread packets waiting, as the driver write may not have to wait for a packet to be fully transmitted before starting the next write. The only case I can think of this happening would be multiple fast pings received, but the overall throughput change would depend on the ratio of how long it takes ktcp to process the reads versus the time saved not waiting for a TX buffer, at the cost of one less RX buffer.

>  As long as telnet/telnetd is not touched, tcp (HTTP) file transfers are stable on ELKS both ways. Transferring (in) huge files (urlget) to /dev/null w/o a hitch. And out - importing /dev/bda1 to the local host (via curl), no problem but slow. Very encouraging.


Great news! However, the difference in speed could well be the lack of actual disk I/O in the first case.

Mellvik:
@ghaerr,
> 
>  The only case I can think of this happening would be multiple fast pings received, but the overall throughput change would depend on the ratio of how long it takes ktcp to process the reads versus the time saved not waiting for a TX buffer, at the cost of one less RX buffer.
> 
> 
Agreed, the only gain is the time it takes to actually transmit a packet, e.g. making transmit back-to-back possible. May be too marginal to worry about. There is another consideration that I hadn't thought about: Taking out a full 6 pages (@256bytes) from the receive buffer is actually 3, some times 6 packets since our window size is so small (always 6 for interactive traffic). So it may be a high price to pay. 
BTW - this may be the time to test larger windows sizes?
> As long as telnet/telnetd is not touched, tcp (HTTP) file transfers are stable on ELKS both ways. Transferring (in) huge files (urlget) to /dev/null w/o a hitch. And out - importing /dev/bda1 to the local host (via curl), no problem but slow. Very encouraging.
> 
> Great news! However, the difference in speed could well be the lack of actual disk I/O in the first case.
> 
Yes, that's at least part of the deal. Now that I can mount /dev/bda1 I get do to more disk related testing, and it's obviously different from working floppy only.

BTW2 - there is one more thing that disturbs (breaks) large file transfers (outgoing from elks): running netstat. ktcp panics, netstat hangs, but pings continue to work.

—Mellvik

PR #717, opened 2020-08-23
[elksemu] handle new medium-model programs, relocations & all

tkchia:


ghaerr:
Hello @tkchia,

Nice catchup on a lot of features that that have been added the past several months!
All changes look fine for commit. I'm running macOS, so I can't actually test, unfortunately.

Reading these changes brought up a few questions:

- in elksemu/elks.c lines 146 and 154: there are several places in ELKS itself where argv and envp arrays are created, both in user and kernel. Should we worry there also about ensuring the argc/envp arrays are on even addresses? On a similar note, I've noticed a few kernel asm files where variables may be placed on odd addresses. Something to worry about? The 8088 doesn't care, but all other processors require an extra bus cycle every time this happens. I haven't checked whether gcc enforces this in .c files.
- in elksemu/elks.h the structures are using \_\_attribute__((packed)): are some compilers packing at larger than 32-bit boundaries (e.g. 64 bit systems)? Do we need to worry about any related packing issues with the ia16-elf-gcc compiler?
- (not included in this commit) in elksemu/elks.h original line 23 comment: no more issues with inodes > 65535, just inodes > 2**32, correct?

Thank you!

tkchia:
Hello @ghaerr,
 
Thanks!

>  * in elksemu/elks.c lines 146 and 154: there are several places in ELKS itself where argv and envp arrays are created, both in user and kernel. Should we worry there also about ensuring the argc/envp arrays are on even addresses? On a similar note, I've noticed a few kernel asm files where variables may be placed on odd addresses. Something to worry about? 

Hmm... actually we probably should!  The alignment of the `argc` value and the `argv` and `envp` pointers will --- if I remember correctly --- also affect where `%sp` is when a new program starts up.

I see that ELKS's `execve` works somewhat differently from (say) Linux's: it basically accepts a blob from userland (`sptr`, `slen`) which should contain all the `argc`, `argv`, and `envp` information, and copies them to the new process, fixing up the pointers.

If you ask me, I think it is OK for the kernel to accept an unaligned blob from the old process; but the kernel should probably align the blob once it is in the new process.

> I haven't checked whether gcc enforces this in .c files.

`gcc-ia16` will align static structures at even addresses, but does not really enforce any alignment beyond that.

>  * in elksemu/elks.h the structures are using __attribute__((packed)): are some compilers packing at larger than 32-bit boundaries (e.g. 64 bit systems)? Do we need to worry about any related packing issues with the ia16-elf-gcc compiler?

There are at least some structures with 4-byte fields which straddle 4-byte boundaries --- e.g. `.est_inode` in `struct elks_stat`, and `.esh_ftrsize` in `struct elks_supl_hdr`.  These will be aligned wrong if we use normal alignment rules when compiling `elksemu` on a normal GCC.

>  * (not included in this commit) in elksemu/elks.h original line 23 comment: no more issues with inodes > 65535, just inodes > 2**32, correct?

That is correct --- I forgot to update that comment.

Thank you!

PR #719, opened 2020-08-24
[ne2k][eth] major cleanup, more inline documentation, small driver fi…

Mellvik:
[ne2k]:
- cleanup debugging stuff and commented out code
- some minor fixes - such as changing packet read to avoid re-reading of the first part of medium/large packets.
- more inline explanatory comments
- removed old and unused source files
- renamed the 2 remaining files, eth-main.c -> ne2k.c, ne2k-mac.S -> ne2k-asm.S


ghaerr:
Hi @Mellvik,

Thanks a lot! I can see you are definitely shoulders-deep in assembly code! Nice work with the cleanup.

Do you feel this driver is getting pretty stable, has this version been tested heavily like the last one?

One point, could/will be a problem: the asm routine ne2k_clr_oflow uses DI but does not save it. Since it is called from C, DI needs to be saved. Depending on the calling procedure, this may or not be a problem with the immediate version, but the procedure will need to save DI. Would you like me to commit this now and have you fix that with an update, or instead push a fix for that now?

Thank you for getting the NE2K driver in fine shape!

Mellvik:
> Thanks @ghaerr,
> 

> Do you feel this driver is getting pretty stable, has this version been tested heavily like the last one?
> 
> 
It has. The speed issue we've talked about has not been resolved - we're back to 60ms RTT now that all the printks are out. The search continues. 
> One point, could/will be a problem: the asm routine ne2k_clr_oflow uses DI but does not save it. Since it is called from C, DI needs to be saved. Depending on the calling procedure, this may or not be a problem with the immediate version, but the procedure will need to save DI. Would you like me to commit this now and have you fix that with an update, or instead push a fix for that now?
> 
> 
Thanks for pointing this out. I'll fix that but please commit now, as I have an other 'command related) commit in the pipeline, and I'll push a PR for this tomorrow.
> Thank you for getting the NE2K driver in fine shape!
> 
> 
A very educating experience indeed.

—mellvik

pawosm-arm:
Sadly, `elks/include/arch/ports.h` still refers to `eth-main.c`.

Mellvik:
Thanks for pointing that out, @pawosm-asm,

I'll include the change in the pending PR.


—Mellvik

pawosm-arm:
Also, config/Configure.help seems to be missing CONFIG_ETH_NE2K description pinning CONFIG_ETH to only one ethernet device, NE2K.

PR #720, opened 2020-08-24
[cmds] Fixes, updates, new commands: uuencode/uudecode

Mellvik:
Minor bugfixes & enhancements and a couple of new commands:
- netstat: Changed legend and sequence for arp stats.
- ICMP: commented out debug print.
- hd: hd produces and extra line if the source file size was a multiple of 16, which  - when re-read by hd -r - would add a null byte to the resulting file. So two presumably equivalent files would have different checksums and sizes.
- more: Added a ':1G' command (a la vi) to rewind to the beginning of the file, and some minor visual fixes.
- stty: stty sane would ignore the fact that speed was zero (yes, it happens!). stty will now restore line speed to stty default, which is (now) 9600bps.
- added uuencode/uudecode as aids for transferring binary files in ascii via serial lines.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -72,6 +72,8 @@ file_utils/rm					:fileutil		:360k
 file_utils/rmdir				:fileutil			:720k
 file_utils/sync					:fileutil		:360k
 file_utils/touch				:fileutil			:720k
+file_utils/uuencode				:fileutil			:1440k
+file_utils/uudecode				:fileutil			:1440k
 sys_utils/chmem					:sysutil			:720k
 sys_utils/kill					:sysutil			:720k
 sys_utils/ps					:sysutil		:360k
--- a/elkscmd/file_utils/Makefile
+++ b/elkscmd/file_utils/Makefile
@@ -12,7 +12,7 @@ include $(BASEDIR)/Make.rules
 
 # note: grep doesn't read stdin, replaced with minix1/grep
 PRGS = ln ls mkdir mkfifo mknod more mv rm rmdir sync touch \
-	cat chgrp chmod chown cmp cp dd df grep
+	cat chgrp chmod chown cmp cp dd df grep uuencode uudecode
 
 all: $(PRGS)
 
@@ -78,6 +78,12 @@ sync: sync.o
 
 touch: touch.o
 	$(LD) $(LDFLAGS) -o touch touch.o $(LDLIBS)
+
+uuencode: uuencode.o
+	$(LD) $(LDFLAGS) -o uuencode uuencode.o $(LDLIBS)
+
+uudecode: uudecode.o
+	$(LD) $(LDFLAGS) -o uudecode uudecode.o $(LDLIBS)
 
 install: $(PRGS)
 	$(INSTALL) $(PRGS) $(DESTDIR)/bin
--- a/elkscmd/file_utils/more.c
+++ b/elkscmd/file_utils/more.c
@@ -20,6 +20,7 @@ int more_wait(int fout, char *msg)
 {
 	struct termios termios;
 	char buf[80], ch;
+	int cnt, ret = 0;
 
 	write(fout, msg, strlen(msg));
 
@@ -30,27 +31,38 @@ int more_wait(int fout, char *msg)
 		termios2.c_cc[VMIN] = 1;
 		tcsetattr(1, TCSAFLUSH, &termios2);
 	}
-	read(1, buf, sizeof(buf));
-	tcsetattr(1, TCSAFLUSH, &termios);
 
-	write(fout,"\r",1); /* move cursor to pos 0 */
+	cnt = read(1, buf, sizeof(buf));
+
 	ch = buf[0];
-	if (ch == ':')
+	if (ch == ':') {
+		write(fout, "\r          \r:", 13);
+		if (cnt < 2) 
+	           cnt = read(1, &buf[1], sizeof(buf-1));
 		ch = buf[1];
-
+	}
 	switch (ch) {
 	case 'N':
 	case 'n':
 		close(fd);
 		fd = -1;
-		return 1;
+		ret = 1;
+		break;
 	case 'Q':
 	case 'q':
-		write(1, "          \r", 11);
 		close(fd);
-		return -1;
+		ret = -1;
+		break;
+	case '1':
+                if (cnt < 2) 
+                   cnt = read(1, &buf[2], sizeof(buf-2));
+		if (buf[2] == 'G') 	/* rewind to beginning of file */
+		    (void) lseek(fd, 0, SEEK_SET);
+		/* else just ignore */
 	}
-	return 0;
+	write(fout, "\r          \r", 12);
+	tcsetattr(1, TCSAFLUSH, &termios);
+	return ret;
 }
 
 int cat_file(int m_in, int m_out) {
--- a/elkscmd/file_utils/uudecode.c
+++ b/elkscmd/file_utils/uudecode.c
@@ -1,0 +1,168 @@
+/*
+ * uudecode [input]
+ *
+ * create the specified file, decoding as you go.
+ * used with uuencode.
+ * Based on BSD2.9 version, ELKS version by HS/aug/2020
+ */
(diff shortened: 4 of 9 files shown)


ghaerr:
Hello @Mellvik,

Thanks for the bunch of cool fixes and new commands!

Glad to know we don't need to print the ping ICMP response anymore, because it works! Progress is being made :)
I'll probably convert that to a debug message in a future PR.

Mellvik:
Thanks @ghaerr,
The small problems and their fixes keep piling up which we're busy on the big ones.... - as usual :-). It's good to get (some of) them out of the stash.

> Glad to know we don't need to print the ping ICMP response anymore, because it works! Progress is being made :)
> I'll probably convert that to a debug message in a future PR.
> 
Makes sense. Actually - it would be very useful to be able to turn this one (maybe others) on and off on a running system. How about a soft debug switch accessed via netstat?

-M

Mellvik:
Apropos ICMP - 

I'm removing the line

if (len & 1) printf("ip: chksum on bad length %d\n", len); //FIXME

in ip.c. From what I can tell, odd packet lengths are ok (at least for ICMP), and no other devices have any reactions to the likes of

ping -s 11 <address>

where -s is the payload size for the ICMNP packet (default is 64, very useful for testing, like -s 1500 …)
Even my pfSense router is more than happy to handle odd size packets...

—Mellvik

> 25. aug. 2020 kl. 09:32 skrev Helge Skrivervik <helge@mymayday.com>:
> 
> Thanks @ghaerr,
> The small problems and their fixes keep piling up which we're busy on the big ones.... - as usual :-). It's good to get (some of) them out of the stash.
> 
>> Glad to know we don't need to print the ping ICMP response anymore, because it works! Progress is being made :)
>> I'll probably convert that to a debug message in a future PR.
>> 
> Makes sense. Actually - it would be very useful to be able to turn this one (maybe others) on and off on a running system. How about a soft debug switch accessed via netstat?
> 
> -M

ghaerr:
@Mellvik,

Are you ever getting any IP checksum errors? We may want to just comment that line out for now, as the line following "len >>= 1" discards odd lengths, and seemingly calculates a checksum ignoring the odd length. 

I would like more information on odd length packets before entirely deleting this. Does it display only on odd-length pings? I don't know yet whether ping puts a non-zero value in the ping data area or not.

Mellvik:
> @Mellvik <https://github.com/Mellvik>,
> 
> Are you ever getting any IP checksum errors? We may want to just comment that line out for now, as the line following "len >>= 1" discards odd lengths, and seemingly calculates a checksum ignoring the odd length.
> 
@ghaerr,
There are no checksum errors.
> I would like more information on odd length packets before entirely deleting this. Does it display only on odd-length pings? I don't know yet whether ping puts a non-zero value in the ping data area or not.
> 
> 
It does indeed only show on odd length pings. I don't think and 'real' traffic would have odd lengths. That's an assumption though, I cannot remember having seen an RFC as 'open ended' as the ICMP echo request packet.

Ping always put content into the data, in order do be able to pinpoint errors and at which byte they happen. Very useful when I was redoing the packet read code and missed a page from the ring buffer :-).
> 
Sure, I'll comment it out for now!

—Mellvik

ghaerr:
> It does indeed only show on odd length pings. I don't think and 'real' traffic would have odd lengths. 

Ok, I've figured out the issue. IP headers will never have an odd length, and the data portion of an IP packet (TCP, ICMP, etc) use their own checksum calculation. For ICMP echo packets received, ktcp is not checking any ICMP checksum, and always generates an ICMP echo reply. When generating ICMP echo reply packets, ktcp is using the same IP checksum routine, and passing the same length as received. From looking at the code, that echo reply has an incorrect checksum if the echo was odd length. It appears that other TCP/IP stacks aren't checking the received checksum either, just using it.

I'll fix this all in a PR, lets leave the line as-is for now, instead of commenting it out.

Mellvik:
@ghaerr, thanks - that was fast.

Now - here's some addition info regarding checksums:

>> Are you ever getting any IP checksum errors? We may want to just comment that line out for now, as the line following "len >>= 1" discards odd lengths, and seemingly calculates a checksum ignoring the odd length.
>> 
> @ghaerr,
> There are no checksum errors.

There are indeed checksum errors, they just haven't occurred before. I looked back at previous traces and no errors from tcpdump, while the most recent one goes like this:

14:32:54.445331 IP (tos 0x0, ttl 64, id 18965, offset 0, flags [DF], proto ICMP (1), length 429)
    10.0.2.2 > 10.0.2.15: ICMP echo request, id 4423, seq 8, length 409
14:32:54.855105 IP (tos 0x0, ttl 64, id 18998, offset 0, flags [DF], proto ICMP (1), length 429)
    10.0.2.2 > 10.0.2.15: ICMP echo request, id 4423, seq 9, length 409
14:32:55.265062 IP (tos 0x0, ttl 64, id 19009, offset 0, flags [DF], proto ICMP (1), length 429)
    10.0.2.2 > 10.0.2.15: ICMP echo request, id 4423, seq 10, length 409
14:32:55.404095 IP (tos 0x0, ttl 64, id 343, offset 0, flags [none], proto ICMP (1), length 429)
    10.0.2.15 > 10.0.2.2: ICMP echo reply, id 4423, seq 8, length 409 (wrong icmp cksum 9ae1 (->ae1)!)
14:32:55.493957 IP (tos 0x0, ttl 64, id 344, offset 0, flags [none], proto ICMP (1), length 429)
    10.0.2.15 > 10.0.2.2: ICMP echo reply, id 4423, seq 9, length 409 (wrong icmp cksum d69f (->469f)!)
14:32:55.585014 IP (tos 0x0, ttl 64, id 345, offset 0, flags [none], proto ICMP (1), length 429)
    10.0.2.15 > 10.0.2.2: ICMP echo reply, id 4423, seq 10, length 409 (wrong icmp cksum b69f (->269f)!)

So it appears that we need to add a zero byte on order to avoid checksum errors. Do you agree?


As to the 3 in/3 out nature in the listing - I'm looking at that. Some ktcp activity will cause this 'stuttering' and it will not go away until ktcp is restarted. Is it possible that the odd length is causing memory corruption in one single byte which subsequently causes this? I haven't seen this until last night, which when I got the weird idea of testing odd packet (payload) length pings.



> 
> It does indeed only show on odd length pings. I don't think and 'real' traffic would have odd lengths.
> 
> Ok, I've figured out the issue. IP headers will never have an odd length, and the data portion of an IP packet (TCP, ICMP, etc) use their own checksum calculation. For ICMP echo packets received, ktcp is not checking any ICMP checksum, and always generates an ICMP echo reply. When generating ICMP echo reply packets, ktcp is using the same IP checksum routine, and passing the same length as received. From looking at the code, that echo reply has an incorrect checksum if the echo was odd length. It appears that other TCP/IP stacks aren't checking the received checksum either, just using it.
> 
> I'll fix this all in a PR, lets leave the line as-is for now, instead of commenting it out.
> 
> 
Sure, no problem.


—Mellvik

ghaerr:
> So it appears that we need to add a zero byte on order to avoid checksum errors. Do you agree?

Yes. I'll fix this and produce a PR. One more checksum bug identified and soon to be fixed.

Looking more deeply into this, I find that ip_calc_chksum is called in three places - IP headers, ICMP checksums, and CSLIP packets. This means that odd-length CSLIP checksums are sure to have an error. It looks like the CSLIP compression routine is using ip_calc_checksum to compute checksums on a compressed TCP header. I am hoping this may finally fix our CSLIP TCP checksum problem!!

The fix could be tricky, as it seems that CSLIP is passing portions (deltas) of an internal buffer to the checksum routine, and we can't just add a nul byte to the buffer, but have to virtually add one in the computation.

> As to the 3 in/3 out nature in the listing - I'm looking at that. Some ktcp activity will cause this 'stuttering' and it will not go away until ktcp is restarted. 

Ok, we need more information on that.


> Is it possible that the odd length is causing memory corruption in one single byte which subsequently causes this? I haven't seen this until last night, which when I got the weird idea of testing odd packet (payload) length pings.

Unlikely, although the received ICMP checksum error on your host could slow down the sending ping, depending on how it's implemented.

PR #722, opened 2020-08-25
[ktcp] Fix odd length IP/ICMP checksum calcs

ghaerr:
Fixes ICMP echo reply bad checksum.

@Mellvik, please test that this fixes the wrong icmp checksum reported in https://github.com/jbruchon/elks/pull/720#issuecomment-680071734, thanks!

### DIFF
--- a/elkscmd/ktcp/icmp.c
+++ b/elkscmd/ktcp/icmp.c
@@ -33,11 +33,9 @@ void icmp_process(struct iphdr_s *iph,unsigned char *packet)
     switch (packet[0]){
     case ICMP_TYPE_ECHO_REQ:
 	len = ntohs(iph->tot_len) - 4 * IP_HLEN(iph);
-#if 0
-	/* for debugging */
-	printf("icmp: PING from %s (len %d id %u seqnum %u)\n",
+	debug_ip("icmp: PING from %s (len %d id %u seqnum %u)\n",
 	    in_ntoa(iph->saddr), len, ntohs(ep->id), ntohs(ep->seqnum));
-#endif
+
 	ep->type = ICMP_TYPE_ECHO_REPL;
 	ep->code = 0;
 	/* return received id, seqnum and extra data*/
--- a/elkscmd/ktcp/ip.c
+++ b/elkscmd/ktcp/ip.c
@@ -38,14 +38,15 @@ int ip_init(void)
 __u16 ip_calc_chksum(char *data, int len)
 {
     __u32 sum = 0;
-    register __u16 *p = (__u16 *) data;
+    __u16 *p = (__u16 *) data;
 
-if (len & 1) printf("ip: chksum on bad length %d\n", len); //FIXME
-    len >>= 1;
+//if (len & 1) printf("ip: chksum on bad length %d\n", len); //FIXME
 
-    do {
+    for (; len > 1 ; len -= 2)
 	sum += *p++;
-    } while (--len > 0);
+
+    if (len == 1)
+	sum += (__u16)(*(__u8 *)p);
 
     while (sum >> 16)
 	sum = (sum & 0xffff) + (sum >> 16);


Mellvik:
@ghaerr,

this works fine! And there is something else - that probably has little to do with this particular fix but something elske you did over the past 24-48 hours:

The ping RTT - which has been going up and down during testing recently, between 20ms and 60ms, is now suddenly 2ms. Holy cr..

helge@raspberrypi:~ $ ping  -s 65 10.0.2.15
PING 10.0.2.15 (10.0.2.15) 65(93) bytes of data.
73 bytes from 10.0.2.15: icmp_seq=1 ttl=64 time=2.36 ms
73 bytes from 10.0.2.15: icmp_seq=2 ttl=64 time=2.35 ms
73 bytes from 10.0.2.15: icmp_seq=3 ttl=64 time=2.37 ms

How did that happen? I'm investigating ...

—Mellvik

ghaerr:
Glad to hear the odd-sized ping checksum errors are now gone.

Is it possible that the reason for the slower earlier response was the host system logging the checksum error?

Otherwise, I also fixed the odd stack pointer problem (a one-liner change in #725) which will have sped up system execution, although I don't think by quite that much!

Mellvik:
@ghaerr,
> 
> Glad to hear the odd-sized ping checksum errors are now gone.
> 
> Is it possible that the reason for the slower earlier response was the host system logging the checksum error?
> 
> 
No, the Pi doesn't get affected, it's a fast bugger, even though it's a 2B.
> Otherwise, I also fixed the odd stack pointer problem (a one-liner change in #725 <https://github.com/jbruchon/elks/pull/725>) which will have sped up system execution, although I don't think by quite that much!
> 
> 
Yes, I do think that may have visible effects. Although the major reason turns out to be the removal of the printf in icmp.c. Now that I've reintroduced some printk's, the RTT is getting back to 'normal' …

I screwed up the local (on stack) variable assembly stuff in the overflow routine. Pulling it back together right now.


Anyway, it's really encouraging that we can get that kind of speed (RTT) on the physical box (this is with the oldest of the ne2k cards, the one with jumpers).

—Mellvik

Mellvik:
@ghaerr,

one more thing (and no, I cannot let the ne2k driver go just yet):

When beating it up real hard, I get this:
-----------------------------------------------------

eth: NIC ring buffer overflow, skipping all packets.
                                                                                                                   /C7b|B7a/ ktcp: eth_process error -1, discarding packet

-------------------------------------------------------
which (probably) means eth_process tried to read and the data was gone.

That's possibly OK, but what we can do to eliminate the situation is to move the call to the overflow detection and processing to the end of the packet read routine. That way there will never be a failed read, just discarded packets (and one less than before). 
The overflow condition will no longer generate its own interrupt, just go into the read flow.

What do you think?


—Mellvik


> 26. aug. 2020 kl. 15:42 skrev Helge Skrivervik <helge@mymayday.com>:
> 
> @ghaerr,
>> 
>> Glad to hear the odd-sized ping checksum errors are now gone.
>> 
>> Is it possible that the reason for the slower earlier response was the host system logging the checksum error?
>> 
>> 
> No, the Pi doesn't get affected, it's a fast bugger, even though it's a 2B.
>> Otherwise, I also fixed the odd stack pointer problem (a one-liner change in #725 <https://github.com/jbruchon/elks/pull/725>) which will have sped up system execution, although I don't think by quite that much!
>> 
>> 
> Yes, I do think that may have visible effects. Although the major reason turns out to be the removal of the printf in icmp.c. Now that I've reintroduced some printk's, the RTT is getting back to 'normal' …
> 
> I screwed up the local (on stack) variable assembly stuff in the overflow routine. Pulling it back together right now.
> 
> 
> Anyway, it's really encouraging that we can get that kind of speed (RTT) on the physical box (this is with the oldest of the ne2k cards, the one with jumpers).
> 
> —Mellvik
>

ghaerr:
Looking at the ne2k.c driver, I would say that moving the overflow routine to the end of the read and being able to turn off the overflow interrupt would simplify the driver.

> which (probably) means eth_process tried to read and the data was gone.

Yes. That printf will also be removed at some point.

However, what happens if, for some reason, there is no process waiting on the read, and an overflow happens - with no interrupt, the card would seemingly be hung, since you've said the overflow must be handled quickly. If a later read or select were called, I presume the RX status could still be read from the card, which would allow that process to continue, so it should be OK. This case shouldn't happen, as if ktcp dies, the kernel will close this driver for it, in which case I assume your release routine turns off all interrupts. 

Go ahead and try it, I like getting all the overflow handling out of the interrupt loop, and one less interrupt to handle specially.

Mellvik:
> 
> Looking at the ne2k.c driver, I would say that moving the overflow routine to the end of the read and being able to turn off the overflow interrupt would simplify the driver.
> 
Yes. I was thinking just adding a call the overflow routine (with some logic) at the end of packet_get.
> which (probably) means eth_process tried to read and the data was gone.
> 
> Yes. That printf will also be removed at some point.
> 
Sounds good.
> However, what happens if, for some reason, there is no process waiting on the read, and an overflow happens - with no interrupt, the card would seemingly be hung, since you've said the overflow must be handled quickly.
> 
This is a point. My stressing the need for speedy handling was more concerned about the desire to get back in operation asap than concern for the NIC status. I think the NIC can handle it, but this needs to be tested. I believe that a situation in which overflow happens is a high traffic situation in which there will be reads pending anyway.
> If a later read or select were called, I presume the RX status could still be read from the card, which would allow that process to continue, so it should be OK.
> 
yes.
> This case shouldn't happen, as if ktcp dies, the kernel will close this driver for it, in which case I assume your release routine turns off all interrupts.
> 
yes.
> Go ahead and try it, I like getting all the overflow handling out of the interrupt loop, and one less interrupt to handle specially.
> 
> —
> 
Agreed, and it's a fairly easy test to do.


—Mellvik

Mellvik:
@ghaerr,
I just tried to create a PR for the ne2k overflow stack botch - and it seems the previous PR somehow got stuck
 - could you take a look and see what it takes to get the PR through?

—Mellvik

> 26. aug. 2020 kl. 18:06 skrev Helge Skrivervik <helge@mymayday.com>:
> 
>> 
>> Looking at the ne2k.c driver, I would say that moving the overflow routine to the end of the read and being able to turn off the overflow interrupt would simplify the driver.
>> 
> Yes. I was thinking just adding a call the overflow routine (with some logic) at the end of packet_get.
>> which (probably) means eth_process tried to read and the data was gone.
>> 
>> Yes. That printf will also be removed at some point.
>> 
> Sounds good.
>> However, what happens if, for some reason, there is no process waiting on the read, and an overflow happens - with no interrupt, the card would seemingly be hung, since you've said the overflow must be handled quickly.
>> 
> This is a point. My stressing the need for speedy handling was more concerned about the desire to get back in operation asap than concern for the NIC status. I think the NIC can handle it, but this needs to be tested. I believe that a situation in which overflow happens is a high traffic situation in which there will be reads pending anyway.
>> If a later read or select were called, I presume the RX status could still be read from the card, which would allow that process to continue, so it should be OK.
>> 
> yes.
>> This case shouldn't happen, as if ktcp dies, the kernel will close this driver for it, in which case I assume your release routine turns off all interrupts.
>> 
> yes.
>> Go ahead and try it, I like getting all the overflow handling out of the interrupt loop, and one less interrupt to handle specially.
>> 
>> —
>> 
> Agreed, and it's a fairly easy test to do.
> 
> 
> —Mellvik
>

ghaerr:
> could you take a look and see what it takes to get the PR through?

I'm not sure what problem you are having - all the PRs have been committed. Try creating a new branch that doesn't already exist, and make sure the branch is created off of master.

PR #728, opened 2020-08-27
[ether][ne2k]Fixed local variable stack allocation in overflow routine.

Mellvik:
Fix for stack allocation error in nic buffer overflow routine (ne2k-asm.S).

Moved initialization of overflow packet discard count (_ne2k_skip_cnt) to the driver init routine (ne2k.c).


ghaerr:
Thanks @Mellvik!

PR #732, opened 2020-08-29
[kernel] Allow MBR CHS data to deviate from CMOS data (#730)

### LINKED ISSUE #730: Partition tables [again]
This issue has been discussed before, and recently it became my (!) problem:

The BIOS HD driver gets physical CHS values from the system NVRAM/CMOS. Then - when reading the partition table from the MBR, it decides to skip the partition if the difference is significant. Such as
```Partitions: bda:(0,614880)  bda1:(63,997857) skipped```

While this may seem like an error in the MBR, it's not. Some old clunkers have a fixed table of drive types to choose from, which may not match what we put in them today. In my case (compaq is probably the worst offender in this), I have a 512MB CF card with CHS parameters that match nothing in the Compaq list. 

My first idea was that the CF card, being what it is, must be reprogrammable - so I set out to change the CHS values. No cigar, it turns out to not be possible, at least for the Transcend cards I have.

Fortunately, the BIOS uses the CMOS CHS values only for booting, so if the heads and sectors per track values matches, the C doesn't matter.  So the CMOS tells the system I have a 300MB drive, which DOS happily boots and disk tells me the real story, 512MB. As does disk on ELKS. But mount on ELKS fail because we've decided that the CMOS is more reliable than the MBR.
```# fdisk /dev/bda
Geometry: 610 cylinders, 16 heads, 63 sectors.
Command (? for help): p

                           START              END          SECTOR
Device           #:ID   Cyl Head Sect    Cyl Head Sect  Start   Size

/dev/bda1       *1:06     0    1    1     989   15   63     63 997857
/dev/bda2        2:00     0    0    0      0    0    0      0      0
/dev/bda3        3:00     0    0    0      0    0    0      0      0
/dev/bda4        4:00     0    0    0      0    0    0      0      0
Command (? for help): 
```

There may be more to this, but my suggestion is to either choose MBR over CMOS or put in an additional sanity check that it is only the cylinder value that differs.

Opinions solicited.

--Mellvik

Mellvik:
- Remove 'sanity check' on partition tables to allow real disk capacity to deviate from BIOS CHS data. Allows CF disks to be fully utilized on old systems with a fixed set of BIOS-supported disk drives. This closes #730.

- Also add user 'tty' to /etc/password - which executes /bin/tty and exits.

This fix for #730 brings up a followup issue: Now /dev/bda and (say) /dev/bda1 mismatch, and /dev/bda is no longer a useable access point for the entire disk drive. Would it not make sense to update the date for the 'master' device when detecting new data in the MBR?

### DIFF
--- a/elks/arch/i86/drivers/block/genhd.c
+++ b/elks/arch/i86/drivers/block/genhd.c
@@ -72,12 +72,14 @@ static void add_partition(struct gendisk *hd, unsigned short int minor,
      * A CHS cylinder can have 63 max sectors * 255 heads, so adjust for that.
      */
     sector_t adj_nr_sects = hd0->nr_sects + 63 * 255;
+#if 0	/* partition skipping disabled as virtual cylinder values sometimes needed for CF cards*/
     if (start > adj_nr_sects || start+size > adj_nr_sects) {
 	printk("skipped ");
 	hdp->start_sect = -1;
 	hdp->nr_sects = 0;
 	return;
     }
+#endif
 
     /*
      * Save boot partition # based on start offset.  This is needed if
--- a/elkscmd/rootfs_template/etc/passwd
+++ b/elkscmd/rootfs_template/etc/passwd
@@ -13,6 +13,7 @@ uucp:*:10:14:UUCP Daemon:/var/spool/uucp:
 operator:*:11:0:Operator:/root:/bin/sash
 man:*:13:15:Manual pages:/usr/man:
 ftp:*:14:50:FTP User:/home/ftp:/bin/sash
+tty::15:0:tty command:/home:/bin/tty
 postmaster:*:14:12:Postmaster:/var/spool/mail:/bin/sash
 nobody:*:99:99:Nobody:/tmp:
 user1::501:501:User 1:/home:/bin/sh


ghaerr:
@Mellvik,

I don't see the \#if 0 in genhd.c in this PR, would you like to push that up?

> a followup issue: Now /dev/bda and (say) /dev/bda1 mismatch, and /dev/bda is no longer a useable access point for the entire disk drive. Would it not make sense to update the date for the 'master' device when detecting new data in the MBR?

Yes, possibly. Not sure how to best do this though. There are lots of bogus MBR CHS entries out there, based on the now-removed "skip" partition boot displays I've seen. It might be better to just allow only the partition size modification, like we're doing, and limit possible exposure to that partition. It could also be possible that an MBR has bogus (smaller) size, which would affect /dev/bda in a harsher way.

Mellvik:
> I don't see the #if 0 in genhd.c in this PR, would you like to push that up?
> 
Thanks @ghaerr, my relationship with git is obviously still strained. Your attention to detail is much appreciated. Update coming.
> a followup issue: Now /dev/bda and (say) /dev/bda1 mismatch, and /dev/bda is no longer a useable access point for the entire disk drive. Would it not make sense to update the date for the 'master' device when detecting new data in the MBR?
> 
> Yes, possibly. Not sure how to best do this though. There are lots of bogus MBR CHS entries out there, based on the now-removed "skip" partition boot displays I've seen. It might be better to just allow only the partition size modification, like we're doing, and limit possible exposure to that partition. It could also be possible that an MBR has bogus (smaller) size, which would affect /dev/bda in a harsher way.
> 
This is a useful discussion, and I agree, there is no right answer. My take is that the MBR partition table must be considered the most dependable source of such data simply because it is the partition table that make a partition (and thus a file system) accessible -- unless we're talking about a flat disk w/o the MBR.

That said it may also be argued that the partitions in the MBR may intentionally (or accidentally) not make use of the entire disk, which may lead to the conclusion that the size of the disk as used in the 'master device' (like /dev/bda) should be the larger of the two. 

And finally, since we now have /bootopts (thanks again for that one!), a per device CHS parameter may be added to override whatever the CMOS and MBR says.

In any case, it seems nonsensical (and impractical) that the 'master device' does not span the available (and useable) partitions.
> 
//Mellvik

ghaerr:
Thanks @Mellvik!

Mellvik:
Thanks @ghaerr,

final thought on the disk size sanity issue:

The most reliable way of getting drive data would most likely be to query the drive, which - if architecture is AT or later - will be IDE [CF cards adhere til the IDE standard]. And then stick to CMOS CHS data for older systems.

—Mellvik

PR #737, opened 2020-09-04
[kernel] Use number rather than string for mount(2) system call fs type

ghaerr:
Moves mount -t types from kernel into mount command.
Cleans up kernel filesystem structures, reduces kernel code.

Allows `mount -t fat ...` as synonym for `mount -t msdos ...`.


Mellvik:
Another BIG thanks, @ghaerr!

Small (not really) change, big improvement.

—Mellvik

PR #742, opened 2020-09-05
[hd driver] Added IDE drive identity query to get more reliable CHS data.

Mellvik:
Code pulled from directhd.c and adapted. Tested on qemu (2 drives), physical (1 drive).

Makes the entire physical disk available via the 'master' device (/dev/bda etc).

For review.

### DIFF
--- a/elks/arch/i86/drivers/block/Makefile
+++ b/elks/arch/i86/drivers/block/Makefile
@@ -37,7 +37,7 @@ OBJS  = init.o genhd.o ll_rw_blk.o rd.o
 
 # BIOS floppy and hard drive support
 ifeq ($(CONFIG_BLK_DEV_BIOS), y)
-OBJS += bioshd.o
+OBJS += bioshd.o idequery.o
 endif
 
 ifeq ($(CONFIG_ROMFS_FS), y)
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -175,7 +175,7 @@ static unsigned short int bioshd_gethdinfo(void) {
 	if (arch_cpu > 5) {	/* Do this only if AT or higher */
 	    if (!get_ide_data(drive, drivep)) {	/* get CHS from the drive itself */
 		/* sanity checks already done, accepting data */
-	    	printk("bioshd: bd%c IDEdata CHS %d,%d,%d\n", 'a'+drive, drivep->cylinders,
+		printk("bioshd: bd%c IDE CHS %d,%d,%d\n", 'a'+drive, drivep->cylinders,
 		drivep->heads, drivep->sectors);
 	    }
 	}
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -130,7 +130,6 @@ static int bioshd_ioctl(struct inode *, struct file *, unsigned int, unsigned in
 static int bioshd_open(struct inode *, struct file *);
 static void bioshd_release(struct inode *, struct file *);
 static void bioshd_geninit(void);
-extern int get_ide_data(int, struct drive_infot *); 
 
 static struct gendisk bioshd_gendisk = {
     MAJOR_NR,			/* Major number */
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -42,9 +42,11 @@
 #include <linuxmt/config.h>
 #include <linuxmt/debug.h>
 
+#include <arch/io.h>
 #include <arch/segment.h>
 #include <arch/system.h>
 #include <arch/irq.h>
+#include <linuxmt/heap.h>
 
 /* the following must match with /dev minor numbering scheme*/
 #define NUM_MINOR	32	/* max minor devices per drive*/
@@ -55,6 +57,20 @@
 
 #define MAJOR_NR BIOSHD_MAJOR
 #define BIOSDISK
+#define HD1_PORT	0x1f0	/* Port address, first controller [IDE] */
+#define HD2_PORT	0x170	/* Port address, first controller [IDE] */
+#define IDE_DRIVE_ID	0xec	/* IDE command to get access to drive data */
+#define IDE_STATUS	7	/* get drive status */
+#define IDE_ERROR	1
+
+#define IDE_PROBE_ENABLE	/* enable CHS priobing via the disk IDE interface */
+#define IDE_DEBUG	0	/* IDE probe debugging */
+
+/* For IDE/ATA access */
+#define STATUS(port) inb_p(port + IDE_STATUS)
+#define ERROR(port) inb_p(port + IDE_ERROR)
+
+#define WAITING(port) (STATUS(port) & 0x80) == 0x80
 
 #include "blk.h"
 
@@ -74,6 +90,7 @@ struct elks_boot_sect {
 } __attribute__((packed));
 
 static int bioshd_initialized = 0;
+static int io_ports[2] = { HD1_PORT, HD2_PORT };	/* physical port addresses */
 
 static struct biosparms bdt;
 
@@ -129,6 +146,7 @@ static int bioshd_ioctl(struct inode *, struct file *, unsigned int, unsigned in
 static int bioshd_open(struct inode *, struct file *);
 static void bioshd_release(struct inode *, struct file *);
 static void bioshd_geninit(void);
+static int get_ide_data(word_t port, struct drive_infot *); 
 
 static struct gendisk bioshd_gendisk = {
     MAJOR_NR,			/* Major number */
@@ -172,9 +190,22 @@ static unsigned short int bioshd_gethdinfo(void)
 	    /* NOTE: some BIOS may underreport cylinders by 1*/
 	    drivep->cylinders = (((BD_CX & 0xc0) << 2) | (BD_CX >> 8)) + 1;
 	    drivep->fdtype = -1;
-	    printk("bioshd: gethdinfo CHS %d,%d,%d\n", drivep->cylinders,
+	    printk("bioshd: bd%c BIOS CHS %d,%d,%d\n", 'a'+drive, drivep->cylinders,
 		drivep->heads, drivep->sectors);
 	}
+#ifdef IDE_PROBE_ENABLE
+	if (arch_cpu > 5) {	/* Do this only if AT or higher */
+		struct drive_infot ide_data;
+		if (!get_ide_data(drive, &ide_data)) {	/* get CHS from the drive itself */
+	    		printk("bioshd: bd%c IDEdata CHS %d,%d,%d\n", 'a'+drive, ide_data.cylinders,
+			ide_data.heads, ide_data.sectors);
+			/* sanity checks already done - using IDE data */
+			drivep->cylinders = ide_data.cylinders;
+			drivep->heads = ide_data.heads;
+			drivep->sectors = ide_data.sectors;
+		}
+	}
+#endif
 	drivep++;
     }
     return ndrives;
@@ -882,5 +913,116 @@ kdev_t bioshd_conv_bios_drive(unsigned int biosdrive)
 	partition = boot_partition;	/* saved from add_partition()*/
     } else
 	minor = (biosdrive & 0x03) + DRIVE_FD0;
+
     return MKDEV(BIOSHD_MAJOR, (minor << MINOR_SHIFT) + partition);
 }
+
+#ifdef IDE_PROBE_ENABLE
+/*
+ * Query the IDE/ATA drive for physical data
+ * Added by HS - sep-2020
+ *
+ * ELKS: ported from directhd.c 
+ */
+
+void insw(word_t port, word_t *buffer, int count) {
+    int i;
+
+    for (i = 0; i < count / 2; i++)
+	buffer[i] = inw(port);
+    return;
+}
+
+void out_hd(word_t drive, word_t nsect, word_t sect,
+	    word_t head, word_t cyl, word_t cmd)
+{
+    word_t port;
+
+    port = io_ports[drive / 2];
+
+    outb(0xff, ++port);		/* the supposedly correct value for WPCOM on IDE */
+    outb_p(nsect, ++port);
+    outb_p(sect, ++port);
+    outb_p(cyl, ++port);
+    outb_p(cyl >> 8, ++port);
+    outb_p(0xA0 | ((drive % 2) << 4) | head, ++port);
+    outb_p(cmd, ++port);
+
+    return;
+}
+
+static int get_ide_data(word_t drive, struct drive_infot *drive_info) {
+
+	int port;
+	word_t *ide_buffer = (word_t *)heap_alloc(512, 0);
+
+	/* send drive_ID command to drive */
+	port = io_ports[drive / 2];
+	out_hd(drive, 0, 0, 0, 0, IDE_DRIVE_ID);
+
+	/* wait */
+	while (WAITING(port));
+	//printk("get_ide_data: drive %i at 0x%3X\n", drive, port);
+
+	if ((STATUS(port) & 1) == 1) {
+	    /* error - drive not found or non-ide */
+#if IDE_DEBUG
+	    printk("bd%s: drive at port 0x%x not found\n", 'a'+drive, port);
+#endif
+	   return -1;
+	}
+
+	/* get drive info */
+
+	insw(port, ide_buffer, 512);
+
+	/* Gather useful info	- this obviously works only for
+	 * IDE/ATA drives, don't call this function if arch < 5.
+	 * Could detect ATAPI drives here (extreme head count).
+	 *
+	 * Safety check - head, cyl and sector values must be other than
+	 * 0 and buffer has to contain valid data (first entry in buffer
+	 * must be other than 0)
+	 *
+	 * This is some sort of bugfix, we will use same method as real Linux -
+	 * work with disk geometry set in current translation mode rather than
+	 * using physical drive info. Physical info might correspond to logical
+	 * info for some drives (those which don't support/use LBA mode).
+	 *
+	 * We're using 'current numbers' from the ID data, could have used 'default CHS 
+	 * numbers' -  words 1, 3 and 6 instead of 54, 55 & 56.
+	 * Apparently, on some drives, the CHS values are programmable. In that case
+	 * the latter are the morer dependable ...
+	 * Surprisingly, the programmability does not seem to be the case on CF cards. 
+	 *  HS sep2020
+	 */
+
+	if ((ide_buffer[54] < 34096) && (*ide_buffer != 0)	/* this is the real sanity check */
+	    && (ide_buffer[54] != 0) && (ide_buffer[55] != 0)
+	    && (ide_buffer[56] != 0)) {
+	    /* Physical value word offsets: cyl 2, heads 3, sectors 6 */
+#if IDE_DEBUG
+	    ide_buffer[20] = 0;
+	    printk("IDE default CHS: %d/%d/%d serial %s\n", ide_buffer[1], ide_buffer[3], ide_buffer[6],
+			&ide_buffer[10]);
+#endif
+	    drive_info->cylinders = ide_buffer[54];
+	    drive_info->heads = ide_buffer[55];
+	    drive_info->sectors = ide_buffer[56];
+
+	}
+
+	/* print drive info */
+	/* sanity check */
+	if (drive_info->heads != 0) {
+#if IDE_DEBUG
+	    printk("bda: %d heads, %d cylinders, %d sectors\n",
+		   drive_info->heads, drive_info->cylinders, drive_info->sectors);
+#endif
+	} else 
+		printk("bd%c: Error in IDE device data.\n", 'a'+drive);
+
+	heap_free(ide_buffer);
+	return 0;
+}
+#endif
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -46,7 +46,6 @@
 #include <arch/segment.h>
 #include <arch/system.h>
 #include <arch/irq.h>
-#include <linuxmt/heap.h>
 
 /* the following must match with /dev minor numbering scheme*/
 #define NUM_MINOR	32	/* max minor devices per drive*/
@@ -57,20 +56,7 @@
 
 #define MAJOR_NR BIOSHD_MAJOR
 #define BIOSDISK
-#define HD1_PORT	0x1f0	/* Port address, first controller [IDE] */
-#define HD2_PORT	0x170	/* Port address, first controller [IDE] */
-#define IDE_DRIVE_ID	0xec	/* IDE command to get access to drive data */
-#define IDE_STATUS	7	/* get drive status */
-#define IDE_ERROR	1
-
 #define IDE_PROBE_ENABLE	/* enable CHS priobing via the disk IDE interface */
-#define IDE_DEBUG	0	/* IDE probe debugging */
-
-/* For IDE/ATA access */
-#define STATUS(port) inb_p(port + IDE_STATUS)
-#define ERROR(port) inb_p(port + IDE_ERROR)
-
-#define WAITING(port) (STATUS(port) & 0x80) == 0x80
 
 #include "blk.h"
 
@@ -90,8 +76,6 @@ struct elks_boot_sect {
 } __attribute__((packed));
 
 static int bioshd_initialized = 0;
-static int io_ports[2] = { HD1_PORT, HD2_PORT };	/* physical port addresses */
-
 static struct biosparms bdt;
 
 /* Useful defines for accessing the above structure. */
@@ -146,7 +130,7 @@ static int bioshd_ioctl(struct inode *, struct file *, unsigned int, unsigned in
 static int bioshd_open(struct inode *, struct file *);
 static void bioshd_release(struct inode *, struct file *);
 static void bioshd_geninit(void);
-static int get_ide_data(word_t port, struct drive_infot *); 
+extern int get_ide_data(int, struct drive_infot *); 
 
 static struct gendisk bioshd_gendisk = {
     MAJOR_NR,			/* Major number */
@@ -168,8 +152,7 @@ static struct gendisk bioshd_gendisk = {
 
 #ifdef CONFIG_BLK_DEV_BHD
 
-static unsigned short int bioshd_gethdinfo(void)
-{
+static unsigned short int bioshd_gethdinfo(void) {
     unsigned short int ndrives = 0;
     int drive;
     register struct drive_infot *drivep = &drive_info[0];
@@ -178,6 +161,7 @@ static unsigned short int bioshd_gethdinfo(void)
     BD_DX = 0x80;
     BD_ES = BD_SI = 0;	/* some buggy BIOS's need this acoording to INT13 on Wiki*/
     ndrives = (call_bios(&bdt) ? 0 : BD_DX & 0xff);
+
     if (ndrives > NUM_DRIVES/2)
 	ndrives = NUM_DRIVES/2;
     for (drive = 0; drive < ndrives; drive++) {
@@ -195,14 +179,11 @@ static unsigned short int bioshd_gethdinfo(void)
 	}
 #ifdef IDE_PROBE_ENABLE
 	if (arch_cpu > 5) {	/* Do this only if AT or higher */
-		struct drive_infot ide_data;
-		if (!get_ide_data(drive, &ide_data)) {	/* get CHS from the drive itself */
-	    		printk("bioshd: bd%c IDEdata CHS %d,%d,%d\n", 'a'+drive, ide_data.cylinders,
-			ide_data.heads, ide_data.sectors);
-			/* sanity checks already done - using IDE data */
-			drivep->cylinders = ide_data.cylinders;
-			drivep->heads = ide_data.heads;
-			drivep->sectors = ide_data.sectors;
+
+		if (!get_ide_data(drive, drivep)) {	/* get CHS from the drive itself */
+			/* sanity checks already done, accepting data */
+	    		printk("bioshd: bd%c IDEdata CHS %d,%d,%d\n", 'a'+drive, drivep->cylinders,
+			drivep->heads, drivep->sectors);
 		}
 	}
 #endif
@@ -916,113 +897,3 @@ kdev_t bioshd_conv_bios_drive(unsigned int biosdrive)
 
     return MKDEV(BIOSHD_MAJOR, (minor << MINOR_SHIFT) + partition);
 }
-
-#ifdef IDE_PROBE_ENABLE
-/*
- * Query the IDE/ATA drive for physical data
- * Added by HS - sep-2020
- *
- * ELKS: ported from directhd.c 
- */
-
-void insw(word_t port, word_t *buffer, int count) {
-    int i;
-
-    for (i = 0; i < count / 2; i++)
-	buffer[i] = inw(port);
-    return;
-}
-
-void out_hd(word_t drive, word_t nsect, word_t sect,
-	    word_t head, word_t cyl, word_t cmd)
-{
-    word_t port;
-
-    port = io_ports[drive / 2];
-
-    outb(0xff, ++port);		/* the supposedly correct value for WPCOM on IDE */
-    outb_p(nsect, ++port);
-    outb_p(sect, ++port);
-    outb_p(cyl, ++port);
-    outb_p(cyl >> 8, ++port);
-    outb_p(0xA0 | ((drive % 2) << 4) | head, ++port);
-    outb_p(cmd, ++port);
-
-    return;
-}
-
-static int get_ide_data(word_t drive, struct drive_infot *drive_info) {
-
-	int port;
-	word_t *ide_buffer = (word_t *)heap_alloc(512, 0);
-
-	/* send drive_ID command to drive */
-	port = io_ports[drive / 2];
-	out_hd(drive, 0, 0, 0, 0, IDE_DRIVE_ID);
-
-	/* wait */
-	while (WAITING(port));
-	//printk("get_ide_data: drive %i at 0x%3X\n", drive, port);
-
-	if ((STATUS(port) & 1) == 1) {
-	    /* error - drive not found or non-ide */
-#if IDE_DEBUG
-	    printk("bd%s: drive at port 0x%x not found\n", 'a'+drive, port);
-#endif
-	   return -1;
-	}
-
-	/* get drive info */
-
-	insw(port, ide_buffer, 512);
-
-	/* Gather useful info	- this obviously works only for
-	 * IDE/ATA drives, don't call this function if arch < 5.
-	 * Could detect ATAPI drives here (extreme head count).
-	 *
-	 * Safety check - head, cyl and sector values must be other than
-	 * 0 and buffer has to contain valid data (first entry in buffer
-	 * must be other than 0)
-	 *
-	 * This is some sort of bugfix, we will use same method as real Linux -
-	 * work with disk geometry set in current translation mode rather than
-	 * using physical drive info. Physical info might correspond to logical
-	 * info for some drives (those which don't support/use LBA mode).
-	 *
-	 * We're using 'current numbers' from the ID data, could have used 'default CHS 
-	 * numbers' -  words 1, 3 and 6 instead of 54, 55 & 56.
-	 * Apparently, on some drives, the CHS values are programmable. In that case
-	 * the latter are the morer dependable ...
-	 * Surprisingly, the programmability does not seem to be the case on CF cards. 
-	 *  HS sep2020
-	 */
-
-	if ((ide_buffer[54] < 34096) && (*ide_buffer != 0)	/* this is the real sanity check */
-	    && (ide_buffer[54] != 0) && (ide_buffer[55] != 0)
-	    && (ide_buffer[56] != 0)) {
-	    /* Physical value word offsets: cyl 2, heads 3, sectors 6 */
-#if IDE_DEBUG
-	    ide_buffer[20] = 0;
-	    printk("IDE default CHS: %d/%d/%d serial %s\n", ide_buffer[1], ide_buffer[3], ide_buffer[6],
-			&ide_buffer[10]);
-#endif
-	    drive_info->cylinders = ide_buffer[54];
-	    drive_info->heads = ide_buffer[55];
-	    drive_info->sectors = ide_buffer[56];
-
-	}
-
-	/* print drive info */
-	/* sanity check */
-	if (drive_info->heads != 0) {
-#if IDE_DEBUG
-	    printk("bda: %d heads, %d cylinders, %d sectors\n",
-		   drive_info->heads, drive_info->cylinders, drive_info->sectors);
-#endif
-	} else 
-		printk("bd%c: Error in IDE device data.\n", 'a'+drive);
-
-	heap_free(ide_buffer);
-	return 0;
-}
-#endif
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -90,12 +90,7 @@ static struct biosparms bdt;
 #define BD_ES bdt.es
 #define BD_FL bdt.fl
 
-static struct drive_infot {		/* CHS per drive*/
-    int cylinders;
-    int sectors;
-    int heads;
-    int fdtype;				/* floppy fd_types[] index  or -1 if hd */
-} drive_info[NUM_DRIVES];
+static struct drive_infot drive_info[NUM_DRIVES];
 
 /* This makes probing order more logical and
  * avoids a few senseless seeks in some cases
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -179,12 +179,11 @@ static unsigned short int bioshd_gethdinfo(void) {
 	}
 #ifdef IDE_PROBE_ENABLE
 	if (arch_cpu > 5) {	/* Do this only if AT or higher */
-
-		if (!get_ide_data(drive, drivep)) {	/* get CHS from the drive itself */
-			/* sanity checks already done, accepting data */
-	    		printk("bioshd: bd%c IDEdata CHS %d,%d,%d\n", 'a'+drive, drivep->cylinders,
-			drivep->heads, drivep->sectors);
-		}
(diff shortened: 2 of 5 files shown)


ghaerr:
Hi Helge,

In general, this looks good. However, its quite a bit of code for a small feature, and we are just 7k bytes from the kernel code segment being completely used up, so time to optimize. A few suggestions follow.

- Since this code shouldn't really be in the BIOS driver, lets make sure all defines start with IDE_, like some already do. Rename STATUS, ERROR and WAITING to IDE_GETSTATUS, etc. I'm not sure the IDE defines should be at the top of the bios driver, as its confusing enough already. Perhaps move all IDE code and defines to another C file, ideqeury.c, or at least to the last portion of bioshd.c.
- All port numbers need to be in ports.h, so we don't start all over with ports in .c files. Rename HD1_PORT etc to BIOS_HD1_PORT and move them to ports.h. (The HD1_PORT etc are already in ports.h, but used by directhd.c. These are now in the BIOS driver and should be named so).
- Line 199: get_ide_data should be passed a drivep and directly copy to it, which removes all the code in lines 200-206.
- Line 941: why drive / 2? Isn't drive either 0 or 1? Looks like this may not have been tested on second drive.
- Line 948: why use drive % 2, same reason as above? Just 'drive' should work in both cases.
- Line 961: why pass all those zero parameters that are never used to a function called once? Put the code inline in get_ide_data and save code.
- Lines 1022-1023: include in the debug code, which should be removed once known working.
- Most of the comment, including copied "bug fix" code from directhd.c, should be removed and made much more concise.
- Line 933, 951 extraneous return statements not required.
- Most of insw() can be made smaller and quicker just using:
```
do {
   *buffer++ = insw(port);
} while (--count);
```
and pass the word count 256.

Optimize, optimize :)

Thank you!

Mellvik:
Thanks for a thorough review, @ghaerr!!

> 5. sep. 2020 kl. 18:53 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> Hi Helge,
> 
> In general, this looks good. However, its quite a bit of code for a small feature, and we are just 7k bytes from the kernel code segment being completely used up, so time to optimize.
> 
I agree, the bloat is worrisome. If it wasn’t for the fact that we really need ‘real’ HD parameters, this would be less important. As it turns out, even the QEMU HD-image many of us keep using for testing, has a /dev/bda that is unable to reach the entire drive. So yes, let’s optimize - and learn.
> A few suggestions follow.
> 
> Since this code shouldn't really be in the BIOS driver, lets make sure all defines start with IDE_, like some already do. Rename STATUS, ERROR and WAITING to IDE_GETSTATUS, etc. I'm not sure the IDE defines should be at the top of the bios driver, as its confusing enough already. Perhaps move all IDE code and defines to another C file, ideqeury.c, or at least to the last portion of bioshd.c.
I like the separate file idea, will separate them!
> All port numbers need to be in ports.h, so we don't start all over with ports in .c files. Rename HD1_PORT etc to BIOS_HD1_PORT and move them to ports.h. (The HD1_PORT etc are already in ports.h, but used by directhd.c. These are now in the BIOS driver and should be named so).
I should have looked that up. What I’m using is the ones in ports.h.
> Line 199: get_ide_data should be passed a drivep and directly copy to it, which removes all the code in lines 200-206.
Of course! I moved all the sanity checking to the get_ide_data routine and missed the obvious.
> Line 941: why drive / 2? Isn't drive either 0 or 1? Looks like this may not have been tested on second drive.
Not quite. And yes, this has been tested with 2 drives, works well (qemu). There is support for two controllers, each with two drives, which is the typical (vintage) PC main board. We would save a few bytes by ignoring the second controller, but not much.
> Line 948: why use drive % 2, same reason as above? Just 'drive' should work in both cases.
See above.
> Line 961: why pass all those zero parameters that are never used to a function called once? Put the code inline in get_ide_data and save code.
Not obvious, but all (or some depending on the device) those zeros are required. Not according to the spec, but in practice. (Took me hours to figure that one out when I did an IDE/CF card driver for CP/M in Z80 assembler a year back.) We can still move it inline, but it’s kind of ugly. Or possibly a macro.
> Lines 1022-1023: include in the debug code, which should be removed once known working.
For once I don’t agree - this should not happen, and when it does, I’d like to know about it. On the other hand, there is the printk bloat factor ...
> Most of the comment, including copied "bug fix" code from directhd.c, should be removed and made much more concise.
Agreed.
> Line 933, 951 extraneous return statements not required.
Thanks.
> Most of insw() can be made smaller and quicker just using:
> do {
>    *buffer++ = insw(port);
> } while (--count);
> and pass the word count 256.
> 
> Optimize, optimize :)
> 
Indeed. It will be interesting to see what gcc does about that. Might get close to the assembly routine I never got to work …

Thanks again,
PR update coming, then we’ll take another round!

—Mellvik

Mellvik:
This update to the PR includes the suggestions from @ghaerr, saves several hundred bytes (!) in size and adds more functionality. Tested with up to 4 drives on VirtualBox (see below) - wasted a lot of time on QEMU which does not respond (sanely) to the IDE getID command on the 2nd and 3rd drive. 

There are a couple of issues to be resolved:
1) Now that the IDE code is in its own file, the defs of `struct drive_infot` and the static array `io_ports` should be moved to a header file.
2) Whether using this code should be a menuconfig option or not.

Having spent quite some time in the directhd driver, I'm tempted to take a stab at getting it to work, in order to get a comparison to the bios driver in terms of performance. If we rule out support for pre-IDE drives, the driver may be about the size of the BIOS-variant, maybe smaller: The IDE command set is really simple.

Output from virtualbox - in this test-setup, drive 3 (master drive on the 2nd controller) is a CDROM, which both BIOS-query and IDE-query quietly  ignores, and continues to the 4th drive, which becomes bdc on ELKS. The output has DEBUG enabled.

```
lp: no ports found
eth: NE2K at 0x300, irq 12, MAC 52:54:00:12:34:56
bioshd: bda BIOS CHS 203,16,63
get_ide_data: drive 0 at 0x1F0, buffer @ D5E2
IDE default CHS: 203/16/63 serial BV15f1b497f-09f749 1
bda: 16 heads, 203 cylinders, 63 sectors
bioshd: bda IDEdata CHS 203,16,63
bioshd: bdb BIOS CHS 1015,16,63
get_ide_data: drive 1 at 0x1F0, buffer @ D5E2
IDE default CHS: 1015/16/63 serial BV4e3065a09-e8e441 a
bdb: 16 heads, 1015 cylinders, 63 sectors
bioshd: bdb IDEdata CHS 1015,16,63
bioshd: bdc BIOS CHS 1024,3,50
get_ide_data: drive 2 at 0x170, buffer @ D5E2
IDE default CHS: 203/16/63 serial BV087aa71b2-2cca28 6
bdd: 16 heads, 203 cylinders, 63 sectors
bioshd: bdc IDEdata CHS 203,16,63
bioshd: 1 floppy drive and 3 hard drives
Partitions: bda:(0,204624)  bda1:(63,204737) 
 bdb:(0,1023120)  bdb1:(63,1023057) 
 bdc:(0,204624)  bdc1:(63,204737) 
device_setup: BIOS drive 0x0, root device 0x380
PC/AT class machine, Unknown x86 CPU, 639K base RAM.
ELKS kernel 0.3.0 (58608 text + 8032 data + 46672 bss + 10832 heap)
Kernel text at c0:0000, data at f0f:0000, 514K for user processes.
fd: found valid ELKS disk parameters on /dev/fd0 boot sector
fd: /dev/fd0 probably has 18 sectors, 2 heads, and 80 cylinders
MINIX-fs: mounting unchecked file system, running fsck is recommended.
Running /etc/rc.d/rc.sys script
Starting networking on eth
ktcp -b 10.0.2.15 10.0.2.2 255.255.255.0
ktcp: ethernet 52.54.00.12.34.56 mtu 1024
Starting daemons telnetd httpd 
ELKS built from commit 67dbe4e0
Mon Sep 07 17:52:05 2020


ELKS 0.3.0

login: 

```

ghaerr:
> saves several hundred bytes (!) in size and adds more functionality. 

That's a lot more than I suspected - nice!

This second version is much cleaned up, thank you.

Following are a list of items and answers to your questions:

In idequery.c:

- Lines 37-38: You've reversed the proposed do {} while() to a while {} loop. This is now buggy as well as slower. The while loop will have to be changed to while (count--) which doesn't generate as good of code and also tests an extra time which would never be needed. It now copies only 255 words when passed 256. (Visualize passing count=1 to understand). Suggest keeping do {} while loop.
- Line 59: same problem as above.
- Instead of using the messy \#ifdef IDE_DEBUG all over the place, suggest using something like the way include/debug.h works, but keeping it within the driver. All debug printk() get changed to debug() and the code looks much cleaner and easier to follow.
```
#if IDE_DEBUG
#define debug   printk
#else
#define debug(...)
#endif
```
The following are minor, but could generate lots of code if not checked. I haven't checked the gcc code generated:
- Line 44: instead of "drive / 2" use "drive >> 1", so compiler never emits divide instruction.
- Line 51: I try not to use % unless you have to, since many compilers emit code to a special modulo routine. Instead of "drive % 2" use "drive & 1".

> 1. Now that the IDE code is in its own file, the defs of `struct drive_infot` and the static array `io_ports` should be moved to a header file.

Lets put struct drive_infot into blk.h under the BIOSDISK ifdef for now.
extern `get_ide_data` can also go into blk.h.

There's no need to make io_ports global.


> 2. Whether using this code should be a menuconfig option or not.

Probably should, but I'm ok with leaving the config option out for the time being. It can then be tested by other users and myself when I return from vacation.

>  the driver may be about the size of the BIOS-variant, maybe smaller

Just remember that the BIOS driver will have to remain anyways because of the need for floppy I/O, probing, etc.

Mellvik:
Thanks again, @ghaerr - 
there are still lots to be learned, not the least about gcc...

> 
> In idequery.c:
> 
> Lines 37-38: You've reversed the proposed do {} while() to a while {} loop. This is now buggy as well as slower. The while loop will have to be changed to while (count--) which doesn't generate as good of code and also tests an extra time which would never be needed. It now copies only 255 words when passed 256. (Visualize passing count=1 to understand). Suggest keeping do {} while loop.
An intentional test that I forgot to reverse (I had postdecremented the counter and got one too many in the first round). Now since you pointed it out, the do{} while(—x) does indeed save two bytes.
> Line 59: same problem as above.
> Instead of using the messy #ifdef IDE_DEBUG all over the place, suggest using something like the way include/debug.h works, but keeping it within the driver. All debug printk() get changed to debug() and the code looks much cleaner and easier to follow.
> #if IDE_DEBUG
> #define debug   printk
> #else
> #define debug(...)
> #endif
Thanks.
> The following are minor, but could generate lots of code if not checked. I haven't checked the gcc code generated:
> 
> Line 44: instead of "drive / 2" use "drive >> 1", so compiler never emits divide instruction.
> Line 51: I try not to use % unless you have to, since many compilers emit code to a special modulo routine. Instead of "drive % 2" use "drive & 1».
yes, will do. turns out that gcc demonstrates both smartness and stupidity in this case. It does not recognize the /2 and change it into a shift, but it does (and maybe this is the reason) keep the remainder after the divide for use in the drive%2 case. Anyways this adjustment should save another 6 (!) bytes :-).
> Now that the IDE code is in its own file, the defs of struct drive_infot and the static array io_ports should be moved to a header file.
> Lets put struct drive_infot into blk.h under the BIOSDISK ifdef for now.
> extern get_ide_data can also go into blk.h.
> 
Sounds good.
> There's no need to make io_ports global.
> 
> Whether using this code should be a menuconfig option or not.
> Probably should, but I'm ok with leaving the config option out for the time being. It can then be tested by other users and myself when I return from vacation.
> 
> the driver may be about the size of the BIOS-variant, maybe smaller
> 
> Just remember that the BIOS driver will have to remain anyways because of the need for floppy I/O, probing, etc.
> 
That’s a good point, I’ll keep that in mind. There is something else next on my list: TCP speedup. Different issue.

PR update coming.


—Mellvik

Mellvik:
Probably now ready for merge.
An interesting observation from Virtualbox:
```
bioshd: bda BIOS CHS 203,16,63
bioshd: bda IDEdata CHS 203,16,63
bioshd: bdb BIOS CHS 1015,16,63
bioshd: bdb IDEdata CHS 1015,16,63
bioshd: bdc BIOS CHS 1024,3,50
bioshd: bdc IDEdata CHS 203,16,63
```
... where the BIOS value for bdc is off the wall (too big for a change) while the IDE data are OK.

Turns out that file systems from /dev/bdcX are not mountable, is there a max 2 drives restriction in the kernel somewhere?
1st mount /dev/bda1:
```
FAT: me=f8,csz=4,#f=2,floc=1,fsz=62,rloc=125,#d=512,dloc=157,#s=63441,ts=0
FAT: 31M, fat16 format
```
2nd mount /dev/bdb1:
```
FAT: me=f8,csz=16,#f=2,floc=1,fsz=250,rloc=501,#d=512,dloc=533,#s=1023057,ts=1023057
FAT: 511M, fat16 format
```
3rd mount /dev/bdc1:
```
FAT: me=0,csz=0,#f=0,floc=0,fsz=0,rloc=0,#d=0,dloc=0,#s=0,ts=0
FAT: Unsupported format
```
/dev/bdc is an exact copy of /dev/bda.

ghaerr:
> Probably now ready for merge.

Your modifications look good, thanks. I'll answer your questions, now I have some concerns about the integrity checking when no IDE drives are present and testing that.

 
> ```
> bioshd: bda BIOS CHS 203,16,63
> bioshd: bda IDEdata CHS 203,16,63
> bioshd: bdb BIOS CHS 1015,16,63
> bioshd: bdb IDEdata CHS 1015,16,63
> bioshd: bdc BIOS CHS 1024,3,50
> bioshd: bdc IDEdata CHS 203,16,63
> ```

If you change "IDEdata" to "IDE " that list will be much easier to read.

> 
> ... where the BIOS value for bdc is off the wall (too big for a change) while the IDE data are OK.

I wouldn't say "off the wall", as those are potentially valid CHS values, just the BIOS has them wrong, correct?

> 
> Turns out that file systems from /dev/bdcX are not mountable, is there a max 2 drives restriction in the kernel somewhere?
> ```
> FAT: me=0,csz=0,#f=0,floc=0,fsz=0,rloc=0,#d=0,dloc=0,#s=0,ts=0
> FAT: Unsupported format
> ```
No - the drive count is correct since it's looping through the bioshd: stuff above with 3 drives. Look at the tail end of genhd.c, that's where the partitions info is read. I need to see the "Partitions:"  kernel display, that will show what it thinks the partitions are.

I notice that you've created /dev/bdc1, as there isn't one already. Add /dev/bdc1 to image/Make.device. I'm trying to keep the /dev device list small, so just add that one for now, which will at least allow for MBR thumb drives to be tested with two additional HDs.

It is quite strange that the values for /dev/bdc1 above sent to FAT are all zero - I would suspect a possible VB problem?

Back to your idequery - I am concerned that the sanity checks starting on line 116 are useless. For instance, with no IDE drives present, the insw could return all 1's and the check would pass. We need to test the sanity checking on systems without IDE and see what junk is returned. All the check does now is check C < 34096 and that C, H and S are nonzero.

The CHS values should be checked for being in real range for CHS. That is, min/max cylinders (harder to know), but heads will always be [1..255] and S probably the same. You'd probably know more than I what valid values are.

Mellvik:
> bioshd: bdc BIOS CHS 1024,3,50
> bioshd: bdc IDEdata CHS 203,16,63
> If you change "IDEdata" to "IDE " that list will be much easier to read.
> 
OK.
> ... where the BIOS value for bdc is off the wall (too big for a change) while the IDE data are OK.
> 
> I wouldn't say "off the wall", as those are potentially valid CHS values, just the BIOS has them wrong, correct?
> 
OK, I agree, looking at it again the total is quite normal and my (mental) calculator got a far too high number.
> Turns out that file systems from /dev/bdcX are not mountable, is there a max 2 drives restriction in the kernel somewhere?
> 
> FAT: me=0,csz=0,#f=0,floc=0,fsz=0,rloc=0,#d=0,dloc=0,#s=0,ts=0
> FAT: Unsupported format
> No - the drive count is correct since it's looping through the bioshd: stuff above with 3 drives. Look at the tail end of genhd.c, that's where the partitions info is read. I need to see the "Partitions:" kernel display, that will show what it thinks the partitions are.
> 
Partitions: bda:(0,204624)  bda1:(63,204737) 
 bdb:(0,1023120)  bdb1:(63,1023057) 
 bdc:(0,204624)  bdc1:(63,204737) 
> I notice that you've created /dev/bdc1, as there isn't one already. Add /dev/bdc1 to image/Make.device. I'm trying to keep the /dev device list small, so just add that one for now, which will at least allow for MBR thumb drives to be tested with two additional HDs.
> 
OK
> It is quite strange that the values for /dev/bdc1 above sent to FAT are all zero - I would suspect a possible VB problem?
> 
Cannot be ruled out, but the drive is fine when the elks floppy is removed and the virtual machines boots DOS.
> Back to your idequery - I am concerned that the sanity checks starting on line 116 are useless. For instance, with no IDE drives present, the insw could return all 1's and the check would pass. We need to test the sanity checking on systems without IDE and see what junk is returned. All the check does now is check C < 34096 and that C, H and S are nonzero.
> 
I have not considered that scenario as it has been an assumption that using this is a menuconfig option. It’s easy enough to add more sanity checks, but it also adds more code.
> The CHS values should be checked for being in real range for CHS. That is, min/max cylinders (harder to know), but heads will always be [1..255] and S probably the same. You'd probably know more than I what valid values are.
> 
This is where the real fun starts - we can query the drive’s capability to do LBA and switch it over to that if we like. Ignoring that, the fact is that ’sane’ numbers apply only to really old (or small) drives. Newer drives use all kinds of tricks to be compatible with old hardware (convert LBA back to CHS), such as claiming insane head counts and ditto sectors. Which is probably why Linux only tests for zeros. A reasonable sanity test would be to calculate C*H*S and compare it to the max sector count also given by the IDE data. If the product is larger, it’s wrong, if it’s more than say 20% lower, it’s wrong.
Still, I’m not sure we want all this code ...

—Mellvik

ghaerr:
> Partitions: bda:(0,204624) bda1:(63,204737) bdb:(0,1023120) bdb1:(63,1023057) bdc:(0,204624) bdc1:(63,204737)

Looks like it is recognized as a partition.

I don't know the reason for the failure yet, we will need to jump further into this. Make sure you've created /dev/bdc1 properly, it should be 3,65.

I think you can turn on DEBUG_BLK in debug.h and see the block numbers being requested, that may help.


> I have not considered that scenario as it has been an assumption that using this is a menuconfig option. It’s easy enough to add more sanity checks, but it also adds more code.

A menuconfig option doesn't stop the problem of booting the kernel with IDE drives removed, there should be a better sanity check, IMO.


> This is where the real fun starts - we can query the drive’s capability to do LBA and switch it over to that if we like. Ignoring that, the fact is that ’sane’ numbers apply only to really old (or small) drives. Newer drives use all kinds of tricks to be compatible with old hardware (convert LBA back to CHS), such as claiming insane head counts and ditto sectors. Which is probably why Linux only tests for zeros.

Lets not even get into LBA, ELKS doesn't use it, and you are currently querying the IDE drive for valid CHS values.

I'm a bit confused, your driver will readily accept almost any non-zero CHS value returned, yet now you're talking about LBA and special conversions, none of which will work on ELKS. ELKS needs a sane CHS, period.


> A reasonable sanity test would be to calculate C*H*S and compare it to the max sector count also given by the IDE data. If the product is larger, it’s wrong, if it’s more than say 20% lower, it’s wrong.

No, very simple - CHS needs to be sanity checked for ELKS-valid CHS values, in order to replace the BIOS values with a potential read from a non-existent IDE port.

>  Still, I’m not sure we want all this code ..

The current code has, (other than checking nonzero ide_buffer[0] for some reason) 4 checks for CHS. We expand them to six checks:  0 < C < 34096 (or perhaps 1024), 0 < H < 256, and 0 < S < 64, or something like that.

This article https://en.wikipedia.org/wiki/Cylinder-head-sector discusses what is likely to be seen on older systems.

ghaerr:
> The fact is that ’sane’ numbers apply only to really old (or small) drives. Newer drives use all kinds of tricks to be compatible with old hardware (convert LBA back to CHS), such as claiming insane head counts and ditto sectors. 

This brings up the question of whether this patch should be concerned with fixing a problem you're currently seeing with ELKS and older drives, or that of getting ELKS to work with modern drives, LBA, and all that entails.

For this PR, I suggest limiting its extent to older valid CHS values (as described above) and attempting to make it work without having to have it configured separately, which means not allowing just nonzero values to be accepted as valid IDE data. We know that ELKS doesn't work with LBA and associated large drives, and there are many concerns with large disks given our support for only MINIX v1 filesystems.

Should the desire to mount and run large drives be wanted, that can be done in a later PR, after field testing this one without a config setting. (If users really want to support a large IDE drive, perhaps using the directhd driver would be the way to go anyways, and it wouldn't use this routine jammed into the supposed BIOS-only driver).

> Which is probably why Linux only tests for zeros.

Maybe - but I doubt it. The directhd code you took this from doesn't even work, and these sources are from very early versions of drivers. Best look at Linux 2.0+ and see what they're doing, but that also isn't valid for ELKS since all modern systems got out of CHS support and moved to LBA+ as soon as they could.

Mellvik:
> I don't know the reason for the failure yet, we will need to jump further into this. Make sure you've created /dev/bdc1 properly, it should be 3,65.
> 
> I think you can turn on DEBUG_BLK in debug.h and see the block numbers being requested, that may help.
> 
WIll check that. If it's VB we'll figure out that too.
> I have not considered that scenario as it has been an assumption that using this is a menuconfig option. It’s easy enough to add more sanity checks, but it also adds more code.
> 
> A menuconfig option doesn't stop the problem of booting the kernel with IDE drives removed, there should be a better sanity check, IMO.
> 
I don't understand what problem you're trying to solve here. I've been attacking the 'unreliable CMOS&BIOS drive data problem'. If there is no drive, the code will return an error - which is probably an improvement.
> A reasonable sanity test would be to calculate CHS and compare it to the max sector count also given by the IDE data. If the product is larger, it’s wrong, if it’s more than say 20% lower, it’s wrong.
> 
> No, very simple - CHS needs to be sanity checked for ELKS-valid CHS values, in order to replace the BIOS values with a potential read from a non-existent IDE port.
> 
AFAIK - There is no chance of that in the current code.
> Still, I’m not sure we want all this code ..
> 
> The current code has, (other than checking nonzero ide_buffer[0] for some reason) 4 checks for CHS. We expand them to six checks: 0 < C < 34096 (or perhaps 1024), 0 < H < 256, and 0 < S < 64, or something like that.
> 
AFAIK the only sanity check we need (possibly with the exception of the ATAPI check) is to verify that the device is there. That's what the check @ buffer[0] is for. If I remember correctly, and IN instruction cannot (on a PC) return a value other than 00 if there is a device at that address.
> This article https://en.wikipedia.org/wiki/Cylinder-head-sector discusses what is likely to be seen on older systems.
> 
Thanks, It's not what I set out to do, but it would be a valuable addition. I'll take a look at it in the morning - and have someone check it out on older HW.

Not quite done yet, but close.

--Mellvik

ghaerr:
> AFAIK the only sanity check we need (possibly with the exception of the ATAPI check) is to verify that the device is there. That's what the check @ buffer[0] is for. If I remember correctly, and IN instruction cannot (on a PC) return a value other than 00 if there is a device at that address.

Ok - I am not aware of that... we should verify this to be the case. But before reading the data, if the IDE controller is not present, it seems we have a bigger problem:
```
            out_hd(drive, IDE_DRIVE_ID);
     	    while (WAITING(port));
```
This code writes a port, then busy loops waiting for 0x80 to go to zero. Perhaps that depends on the inb returning 0 on a non-existent port, I don't know. Otherwise, this WAITING code will never hang on a system that doesn't have an IDE controller. The code did come from a driver that already tested, I presume, the presence of a controller during probe time.
Then, the following code:
```
	    if ((STATUS(port) & 1) == 1) {
```
will return a 0 if what you say is true, and then your code will incorrectly assume an IDE controller is present, since the port didn't return 0x01 in the lower bit indicating no controller.

With no controller present, and your assumption that reading I/O ports with nothing there returns 0, the WAITING doesn't loop, and the STATUS returns 0, and then the insw returns all zeros. If the assumption is false, then this code could easily hang ELKS at boot.

If the BIOS reads two hard drives present, but no controller present, then this code will confusingly say "Drive at port not found".

Mellvik:
FWIW - I never suggested doing anything withLBA, just observed that the IDE interfaces have capabilities that can be (ab)used whenever. Including- in our case - testing whether the CHS values are within the block count of the device. 

BTW it is also my opinion - and the motivation for this patch - that the IDE data are reliable, period. So, AFAIK when the presence of an IDE device is ascertained, check for ATAPI (sane head count), and accept the data. While the reliability may not be 100%, it's much better than CMOS.

> ﻿
> AFAIK the only sanity check we need (possibly with the exception of the ATAPI check) is to verify that the device is there. That's what the check @ buffer[0] is for. If I remember correctly, and IN instruction cannot (on a PC) return a value other than 00 if there is a device at that address.
> 
The fist word is an IDE signature, <>0 seems to be the litmus test for a functioning IDE. Also, like you point out below, at this stage it can be assumed that a probe has been done. We know for sure that the BIOS/CMOS says the device is there, otherwise we would not be here.
> Ok - I am not aware of that... we should verify this to be the case. But before reading the data, if the IDE controller is not present, it seems we have a bigger problem:
> 
>             out_hd(drive, IDE_DRIVE_ID);
>      	    while (WAITING(port));
> This code writes a port, then busy loops waiting for 0x80 to go to zero. Perhaps that depends on the inb returning 0 on a non-existent port, I don't know. Otherwise, this code will hang on a system that doesn't have an IDE controller. The code did come from a driver that already tested, I presume, the presence of a controller during probe time.
> Then, the following code:
> 
> 	    if ((STATUS(port) & 1) == 1) {
> will return a 0 if what you say is true, and then your code will incorrectly assume an IDE controller is present, since the port didn't return 0x01 in the lower bit indicating no controller.
> 
I think your analysis is correct on both accounts and that the key observation is that we're not probing, that has been done before. We're gathering data. OTOH, avoiding possible hang situations is valuable, and of the above, only the status loop is critical, the waiting() is handled by 0 so to speak (will verify that too).
> See where I'm going with this?
> 
I (possibly) am,and my point is that while this may not be perfect, it's a big step in the right direction. We cannot possibly guess all scenarios that may occur, that has been the pc's curse since day one. Again, it was never the idea to handle pre ide devices, let those be handled by bios data. Thusly, menuconfig is an option for those who don't want or need the idequery functionality.

--M
> .

ghaerr:
Hi @Mellvik,

I got myself a bit confused in my last post, I have edited it.

In general, I agree with your comments above, but since it seems that it is not easy to determine whether an IDE controller is in fact present, I still think it a good idea to test at least the H and S values for something in the normal range, otherwise it is possible an IDE drive could be recognized when there isn't one, although I suppose the worst that would happen is that the mount would fail due to bad data.

> the key observation is that we're not probing,

What I meant is the probing to determine whether a device is present should be done before writing a register stack and reading 512 from a possibly non-existent device. Is this all the probing that Linux 2.0 does to determine whether an IDE device is present? I think this code is meant to read the IDE status register as to what the drive specs are, not to determine whether an controller is present.

If you feel that a menuconfig option would be better at this time, go ahead, although that limits the normal user ability to use this feature unless they 1) know about it, or 2) it defaults to ON. If it defaults to on, then my issues become more important, and the main use of turning it off would be to stop it from malfunctioning.

Mellvik:
@ghaerr,
It occured to me overnight that most of this discussion is actually moot.
No PC that I've ever been in contact with will pass POST if there is a configured drive missing.
So if elks boots, it is safe to assume that the configured drives are there - and we're back to the initial challenge of determining reliable CHS data. My take is still that IDE IDent data is the most reliable source possible.

--Mellvik

Mellvik:
I think we're at the end of this discussion, so let me summarize:
 - The purpose of the enhancement is to get CHS data from IDE-drives when available. When such data is found, it is preferred over CMOS/BIOS data.
 - PCs from AT and up support IDE drives, they do not support older drives. The code will not run unless arch is AT+.
 - The code is run after CMOS data has been retrieved via the BIOS and assumes the device is there, which is safe since a PC will not boot if there is a configured drive missing.
 - The code supports the two standard PC/AT HD IO addresses only.
 - The code does rudimentary sanity checking of the IDE data. This part can be improved, but when the IDE ID signature word is nonzero, and the CHS values are all non-zero, plus the # of heads is reasonable, it is reasonable to trust the data (insofar as we trusted the CMOS data before).
- Testing with 3 and 4 drives continues, it is unlikely that the mount problems are related to the idequery function.

Tested on physical (1 drive), qemu (2 drives), VB (3 drives).

I suggest this PR get merged and we add fixes when neccessary. The addition can easily be dibbled by removing the IDE_PROBE_ENABLE define in bioshd.c.

--Mellvik

ghaerr:
Ok, I now get your point that this routine only runs when the BIOS has already reported a hard drive present.

Except that the routine does probe drive 3 if drive 2 fails to respond. That's ok I suppose, unless that happens on a four drive system, in which case its buggy.

> The code does rudimentary sanity checking of the IDE data. This part can be improved

That's all I was asking for - a little more checking. I'm fine with everything else.

> but when the IDE ID signature word is nonzero, and the CHS values are all non-zero,

If an in_b returns nonzero for a non-existent port, this would set all words non-zero. Just saying. We never did verify what actually happens on a PC bus for non-existent ports. 

The BIOS could also have been setup with one USB drive, for all I know, no IDE controller, and this routine would still run, and we would be querying a non-existent IDE controller, and overwriting the result on the USB drive info!

> plus the # of heads is reasonable

The only check on the # heads is 1-65535 (i.e. nonzero). I don't think just non-zero is reasonable, is it? We should be checking for 1-255. And sectors 1-63. You'll notice that all your testing has confirmed to have heads and sectors within that range. And keep the cylinders check to be 1-34095, as it is now. This would lower the probability of the USB drive bug described above from happening by quite a bit, I think.

I am guessing, but suspect, that all drives would use up cylinders (likely to 34095) first before moving H and S out of compatibility range, just for extreme compatibility.

You've done a good job on this. I'm not trying to drag it out, just being careful. I feel its better to be a bit on the conservative side, and then open up when, say, super large drive support is required, rather than suddenly say a booting problem that we don't have any idea what the problem is. That's all!

Thank you!

ghaerr:
Also I thought you were going to add /dev/bdc1 to image/Make.devices.

Mellvik:
﻿
> Except that the routine does probe drive 3 if drive 2 fails to respond. That's ok I suppose, unless that happens on a four drive system, in which case its buggy.
> 
No, not quite. This works either way - 3 drives or 4 drives, master, slave.
> The code does rudimentary sanity checking of the IDE data. This part can be improved
> 
> That's all I was asking for - a little more checking. I'm fine with everything else.
> 
> but when the IDE ID signature word is nonzero, and the CHS values are all non-zero,
> 
> If an in_b returns nonzero for a non-existent port, this would set all words non-zero. Just saying. We never did verify what actually happens on a PC bus for non-existent ports.
> 
There is exactly zero chance for this to happen. The system will not boot if this is the case.
> The BIOS could also have been setup with one USB drive, for all I know, no IDE controller, and this routine would still run,
> 
I have no clue about USB drives on old systems. I didn't know ELKS supported USB. If used in an IDE converter (like CF cards), we should be fine. If there is no IDE interface, this code will never run.
> and we would be querying a non-existent IDE controller, and overwriting the result on the USB drive info!
> 
> plus the # of heads is reasonable
> 
> The only check on the # heads is 1-65535 (i.e. nonzero). I don't think just non-zero is reasonable, is it? We should be checking for 1-255. And sectors 1-63.
> 
That depends on to what extent we trust the IDE data. My point is that we don't sanitycheck CMOS data - known to be unreliable, so it's hard to argue for rigorous santiychecking on data known to BE reliable. Technically, testing for IDE max values for heads and sectors is - like you point out - trivial. The balance between 'good enough' and 'perfect'.
> You'll notice that all your testing has confirmed to have heads and sectors within that range.
> 
Yes. This is the max range for BIOS INT 13 calls, so if testing these are definitely the numbers to test against.
> And keep the cylinders check to be 1-34095, as it is now. This would lower the probability of the USB drive bug described above from happening by quite a bit, I think.
> 
I'm not familiar with the 'USB bug'. What is it?
> I am guessing, but suspect, that all drives would use up cylinders (likely to 34095) first before moving H and S out of compatibility range, just for extreme compatibility.
> 
The balance between cyls and heads varies - such as in this case:
...
bioshd: bdc BIOS CHS 1024,255,63
IDE default CHS: 16383/16/63 serial BVf0a0e5824-2253de d
bdd: 16 heads, 16383 cylinders, 63 sectors
bioshd: bdc IDE CHS 16383,16,63
> 
—Mellvik

ghaerr:
> This is the max range for BIOS INT 13 calls, so if testing these are definitely the numbers to test against.
> The balance between 'good enough' and 'perfect'.

I'm not trying to nitpick this, seriously. I've just seen too many cases of "can't happen" become "just happened". It is the kernel's responsibility to not crash, it has to be very careful "taking other devices word for it". I initially thought the same way as you on this while updating the sys_exec routine when evaluating a.out header fields: "why spend so much kernel code checking things, if the user sends us a bogus program, its his own fault, and couldn't happen anyways given the compiler". But the kernel is protecting itself, not the user; better for it to keep running, than stop operating correctly because of a user or strange system/header/device/unexpected situation, etc.

Since this is called from the BIOS driver, we should check the H value as 0 < H < 256 and 0 < S < 64, so that we prevent the bios driver from ever sending garbage to INT 13. Since we're already checking non-zero, a single additional check for H, and for S would do it. After all your other explanations, it seems we really don't need a config file option and can remain always on, except for special systems where even the probe would cause problems, and IDE_ENABLE_PROBE can remove it.

> I have no clue about USB drives on old systems. I didn't know ELKS supported USB. 

USB thumb drives are only supported through the BIOS. We've been able to boot from them since revising the ELKS bios driver /dev structure, which allowed BIOS to pass 82h or 83h (drive C and D) in addition to hard drives at 80h and 81h, for use in the ELKS bios driver.

Mellvik:
> 
> Since this is called from the BIOS driver, we should check the H value as 0 < H < 256 and 0 < S < 64, so that we prevent the bios driver from ever sending garbage to INT 13. Since we're already checking non-zero, a single additional check for H, and for S would do it.
> 
The balance between code size and being careful ...
> After all your other explanations, it seems we really don't need a config file option and can remain always on, except for special systems where even the probe would cause problems, and IDE_ENABLE_PROBE can remove it.
> 
> I have no clue about USB drives on old systems. I didn't know ELKS supported USB.
> 
> USB thumb drives are only supported through the BIOS. We've been able to boot from them since revising the ELKS bios driver /dev structure, which allowed BIOS to pass 82h or 83h (drive C and D) in addition to hard drives at 80h and 81h, for use in the ELKS bios driver.
> 
I don't see how this setup can affect the IDEquery setup. If there is no IDE interface the code will detect it. If there is it should work fine. I guess we'll just have to wait till someone tests it …

I'll add the tests and update the PR. 

BTW —  The 'cannot access /dev/bdc' problem went away after a make clean.
And a new problem popped up — when writing to /dev/bdc (FAT fs).

panic: Process 44 exceeded kernel stack limit! magic 2a29

apparent call stack:
Line: Addr    Parameters
~~~~: ~~~~    ~~~~~~~~~~

etc etc


I'm looking into the repeatability of that now.


—M

ghaerr:
> And a new problem popped up — when writing to /dev/bdc (FAT fs).
> panic: Process 44 exceeded kernel stack limit! magic 2a29

Dang. That's very likely the result of my trying to decrease the kernel stack size (per task) from 988 to 512.
Let me know if you can get it to repeat, as this signals too many stack variables in some of the FAT routines.
It can be fixed by increasing KSTACK_BYTES in include/linuxmt/sched.h.

Mellvik:
> 
> And a new problem popped up — when writing to /dev/bdc (FAT fs).
> panic: Process 44 exceeded kernel stack limit! magic 2a29
> 
> Dang. That's very likely the result of my trying to decrease the kernel stack size (per task) from 988 to 512.
> Let me know if you can get it to repeat, as this signals too many stack variables in some of the FAT routines.
> It can be fixed by increasing KSTACK_BYTES in include/linuxmt/sched.h.
> 
> 
Turns out it is completely repeatable. (block debug on)

—Mellvik

-------------------------------------------------------------------

…
Partitions: bda:(0,204624) BLK 0 read
 bda1:(63,204737) 
 bdb:(0,1023120) BLK 0 read
 bdb1:(63,1023057) 
 bdc:(0,16514064) BLK 0 read
 bdc1:(63,1023057) 
device_setup: BIOS drive 0x0, root device 0x380
…

# ls /mnt
BLK 8 read
BLK 990 read
BLK 10 read
1  2  3  
# mount -t msdos /dev/bda1 /mnt/1
BLK 696 read
BLK 697 read
BLK 1066 read
BLK 0 read
FAT: me=f8,csz=4,#f=2,floc=1,fsz=62,rloc=125,#d=512,dloc=157,#s=63441,ts=0
FAT: 31M, fat16 format
BLK 62 read
# mount -t msdos /dev/bdb1 /mnt/2
BLK 0 read
FAT: me=f8,csz=16,#f=2,floc=1,fsz=250,rloc=501,#d=512,dloc=533,#s=1023057,ts=1023057
FAT: 511M, fat16 format
BLK 250 read
# mount -t msdos /dev/bdc1 /mnt/3
BLK 0 read
FAT: me=f8,csz=16,#f=2,floc=1,fsz=250,rloc=501,#d=512,dloc=533,#s=1023057,ts=1023057
FAT: 511M, fat16 format
BLK 250 read
# ls /mnt/3/bin
BLK 274 read
BLK 275 read
arp       banner    basename  bc        bin       cal       cat       
chgrp     chmem     
# rm /mnt/3/bin/*
BLK 841 read
BLK 842 read
BLK 843 read
BLK 844 read
BLK 845 read
BLK 846 read
# cp /bin/* /mnt/3/bin
BLK 768 read
BLK 769 read
BLK 770 read
BLK 771 read
BLK 772 read
BLK 276 read
BLK 277 read
BLK 278 read
BLK 279 read
BLK 280 read
BLK 281 read
BLK 282 read
BLK 750 read
BLK 290 read
BLK 751 read
BLK 752 read
BLK 753 read
BLK 895 read
BLK 298 read
BLK 896 read
BLK 897 read
BLK 898 read
BLK 899 read
BLK 900 read
BLK 901 read
BLK 4 read
BLK 184 read
BLK 306 read
BLK 280 read
BLK 314 read
BLK 281 read
BLK 282 read
BLK 283 read
BLK 284 read
BLK 285 read
BLK 286 read
BLK 288 read
BLK 287 read
BLK 289 read
BLK 322 read
BLK 290 read
BLK 291 read
BLK 292 read
BLK 293 read
BLK 294 read
BLK 295 read
BLK 296 read
BLK 297 read
BLK 330 read
BLK 298 read
BLK 299 read
BLK 300 read
BLK 301 read
BLK 302 read
BLK 303 read
BLK 304 read
BLK 305 read
BLK 330 write
BLK 329 write
BLK 328 write
BLK 327 write
BLK 326 write
BLK 325 write
BLK 324 write
BLK 323 write
BLK 322 write
BLK 321 write
BLK 320 write
BLK 319 write
BLK 318 write
BLK 317 write
BLK 316 write
BLK 315 write
BLK 314 write
BLK 313 write
BLK 312 write
BLK 311 write
BLK 310 write
BLK 309 write
BLK 308 write
BLK 307 write
BLK 306 write
BLK 305 write
BLK 304 write
BLK 303 write
BLK 302 write
BLK 301 write
BLK 300 write
BLK 299 write
BLK 298 write
BLK 297 write
BLK 296 write
BLK 295 write
BLK 294 write
BLK 293 write
BLK 292 write
BLK 291 write
BLK 290 write
BLK 289 write
BLK 288 write
BLK 287 write
BLK 286 write
BLK 285 write
BLK 284 write
BLK 283 write
BLK 282 write
BLK 250 write
BLK 0 write
BLK 274 write
BLK 338 read
BLK 306 read
BLK 307 read
BLK 308 read
BLK 309 read
BLK 310 read
BLK 311 read
BLK 312 read
BLK 313 read
BLK 346 read
BLK 314 read
BLK 315 read
BLK 316 read
BLK 317 read
BLK 274 read
BLK 16 read
BLK 18 read
BLK 5 read
BLK 990 read
BLK 250 read
BLK 275 read
BLK 276 read
BLK 277 read
BLK 278 read
BLK 279 read
BLK 280 read
BLK 281 read
BLK 282 read
BLK 506 read
BLK 354 read
BLK 507 read
BLK 508 read
BLK 509 read
BLK 510 read
BLK 51 read
BLK 362 read
BLK 52 read
BLK 373 read
BLK 370 read
BLK 374 read
BLK 375 read
BLK 376 read
BLK 377 read
BLK 545 read
BLK 7 read
BLK 990 read
BLK 250 read
BLK 275 read
BLK 276 read
BLK 277 read
BLK 278 read
BLK 279 read
BLK 280 read
BLK 281 read
BLK 282 read
BLK 863 read
BLK 378 read
BLK 864 read
BLK 865 read
BLK 866 read
BLK 867 read
BLK 868 read
BLK 16 read
BLK 18 read
BLK 545 read
BLK 7 read
BLK 990 read
BLK 378 write
BLK 377 write
BLK 376 write
BLK 375 write
BLK 374 write
BLK 373 write
BLK 372 write
BLK 371 write
BLK 370 write
BLK 369 write
BLK 368 write
BLK 367 write
BLK 366 write
BLK 365 write
BLK 364 write
BLK 363 write
BLK 362 write
BLK 361 write
BLK 360 write
BLK 359 write
BLK 358 write
BLK 357 write
BLK 356 write
BLK 355 write
BLK 354 write
BLK 353 write
BLK 352 write
BLK 351 write
BLK 350 write
BLK 349 write
BLK 348 write
BLK 347 write
BLK 274 write
BLK 346 write
BLK 345 write
BLK 344 write
BLK 343 write
BLK 342 write
BLK 341 write
BLK 340 write
BLK 339 write
BLK 0 write
BLK 330 write
BLK 338 write
BLK 337 write
BLK 336 write
BLK 335 write
BLK 334 write
BLK 333 write
BLK 332 write
BLK 331 write
BLK 250 read
BLK 275 read

panic: Process 33 exceeded kernel stack limit! magic 0

apparent call stack:
Line: Addr    Parameters
~~~~: ~~~~    ~~~~~~~~~~
0: 29EE => 1320 3324 1320 3930 1000 0021 0017
1: BD7C => 0032 0093 0082 007E 008D 0028 0024
2: 001B => 6973 657A 3E20 3220 3535 000A 6F72
3: 2072 => 0000 0000 0000 0000 0000 0000 0000
4: 0000 => 632F 6E6F 6F73 656C 7300 5500 616E
5: 7665 => 0000 0000 0000 0000 0000 0000 0000
6: 0000 => 632F 6E6F 6F73 656C 7300 5500 616E
7: 7665 => 0000 0000 0000 0000 0000 0000 0000
8: 0000 => 632F 6E6F 6F73 656C 7300 5500 616E
SYSTEM HALTED - Press CTRL-ALT-DEL to reboot:

ghaerr:
Thanks a lot Helge.

I'll take up the kernel stack overflow and fix in another PR.

PR #743, opened 2020-09-06
[kernel] Allow kernel to load and run with far text segments

ghaerr:
This PR is meant to show progress and invite review of adding far text procedure capabilities to the ELKS kernel, in an effort to reduce .text code segment size for future enhancements. Code reviews welcome.

Technical discussion on #705.

Current status is kernel booting from normal and extended a.out header kernels.
Revised setup.S to read far text section if present.
Added thunking code to kernel library.
Fixed tools/build to allow supplemental header binaries.

Note done yet:
Relocation routines not done, coming.
Will display far text code size separately from near code size.
Any far text procedure added to kernel will be prototyped such that far text will be a configurable option.

Unsure yet how to handle CONFIG_ROMCODE and !REL_SYS configurations. Romcode depends on whether gcc can emit pre-relocated segment relocations for code at fixed addresses or not.

Also have concerns as to whether DEF_SYSSEG at 0x1000 may be overwritten by large kernel during REL_SYS copy. Suggest moving to 0x2000.

### DIFF
--- a/elks/Makefile-rules
+++ b/elks/Makefile-rules
@@ -132,7 +132,7 @@ endif
 # No relocation stuff related to MSDOS 'MZ' executable
 
 CROSS_CC = ia16-elf-gcc
-CROSS_CFLAGS = -ffreestanding -fno-inline -melks -mcmodel=small -mno-segment-relocation-stuff
+CROSS_CFLAGS = -ffreestanding -fno-inline -melks -mcmodel=small -msegment-relocation-stuff
 
 #########################################################################
 # Define architecture-specific flags.
--- a/elks/arch/i86/boot/crt0.S
+++ b/elks/arch/i86/boot/crt0.S
@@ -20,10 +20,12 @@ entry:
 ! Setup.S already initialized DS and ES (but not SS)
 ! In addition, registers contain:
 !   BX, Text size
+!   DI  Far text size
 !   SI, Data size
 !   DX, BSS size
 */
 	mov	%bx,_endtext
+	mov	%di,_endftext
 	mov	%si,_enddata
 	add	%dx,%si
 	mov	%si,_endbss
@@ -82,6 +84,7 @@ ep_end:
 
 	.data
 	.global _endtext
+	.global _endftext
 	.global _enddata
 	.global _endbss
 	.extern	kernel_cs
@@ -89,6 +92,9 @@ ep_end:
 	.extern task
 
 _endtext:
+	.word	0
+
+_endftext:
 	.word	0
 
 _enddata:
--- a/elks/arch/i86/boot/crt0.S
+++ b/elks/arch/i86/boot/crt0.S
@@ -82,6 +82,12 @@ ep_end:
 
 //	Segment beginnings
 
+// Zero for NULL pointers (near and far)
+// Will be linked as first section in data segment
+	.section .nildata
+	.word 	0
+	.word	0
+
 	.data
 	.global _endtext
 	.global _endftext
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -540,7 +540,10 @@ _next1:
 	jne sys_hdr_bad
 	cmpw $0x0430,2   // i8086 - executable with separated I & D
 	jne sys_hdr_bad
-	cmpb $0x20,4     // header size (no relocs)
+	mov 4,%bx        // BX = header size
+	cmp $0x20,%bx   // header size (no relocs)
+	je sys_hdr_good
+	cmp $0x40,%bx   // header size (relocs)
 	jne sys_hdr_bad
 	jmp sys_hdr_good
 
@@ -565,6 +568,7 @@ sys_hdr_good:
 	add $0xF,%dx     // align on next paragraph (issue #209)
 	mov	$4,%cl
 	shr %cl,%dx   // DX = code segment size in paragraph
+	shr %cl,%bx   // BX = header size in paragraphs
 
 	mov 12,%ax    // data size
 	push %ax
@@ -575,7 +579,8 @@ sys_hdr_good:
 
 // Relocate kernel code
 
-	mov $SYSSEG+2,%ax
+	mov $SYSSEG,%ax
+	add %bx,%ax   // skip header
 	mov %ax,%ds
 	mov $REL_SYSSEG,%ax
 	mov %ax,%es
@@ -587,11 +592,49 @@ sys_hdr_good:
 	rep
 	movsb
 
+// Relocate kernel far text
+
+	xor %ax,%ax   // far text size = 0
+	push %ax      // -10(%bp)
+	push %ax      // -12(%bp)
+	push %ax      // -14(%bp)
+	cmp $0x04,%bx // extended header?
+	jne sys_no_far_text
+	mov $SYSSEG,%ax
+	mov %ax,%ds
+	mov 0x30,%ax  // far text size in bytes
+	mov %ax,-12(%bp)
+	mov 0x20,%ax  // text relocation size in bytes
+	mov %ax,-14(%bp)
+	add $15,%ax
+	mov $4,%cl
+	shr %cl,%ax   // far text size in paragraphs
+	mov %ax,-10(%bp)
+
+	mov 0x30,%cx  // far text size in bytes
+	cmp %cx,%cx
+	jz sys_no_far_text
+	mov $SYSSEG,%ax
+	add %bx,%ax   // skip header
+	add %dx,%ax   // skip code
+	mov %ax,%ds
+	mov $REL_SYSSEG,%ax
+	add %dx,%ax
+	mov %ax,%es
+
+	xor	%si,%si
+	mov	%si,%di
+	cld
+	rep
+	movsb
+sys_no_far_text:
+
 // Relocate kernel data (not bss)
 // Kernel resets bss itself
 
-	mov $SYSSEG+2,%ax
-	add %dx,%ax
+	mov $SYSSEG,%ax
+	add %bx,%ax   // skip header
+	add %dx,%ax   // skip code
 	mov %ax,%ds
 	mov $REL_SYSSEG,%ax
 	add %dx,%ax
@@ -599,7 +642,8 @@ sys_hdr_good:
 
 	mov -4(%bp),%cx  // data size
 	xor %si,%si
-	mov %si,%di
+	add -12(%bp),%si // skip far text
+	xor %di,%di
 	cld
 	rep
 	movsb
@@ -610,6 +654,7 @@ sys_hdr_good:
 	mov %ax,%ds
 
 	mov -2(%bp),%bx    // code size
+	add -12(%bp),%bx   // FIXME may overflow, needs farcode size
 	mov -4(%bp),%si    // data size
 	mov -6(%bp),%dx    // bss size
 	mov -8(%bp),%cx    // entry point
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -454,11 +454,12 @@ size_ok:
 */
 #if defined(CONFIG_ROMCODE)
 
-// BX,CX,DX,SI,DS,ES are expected in kernel crt0.S
+// BX,CX,DX,SI,DI,DS,ES are expected in kernel crt0.S
 
 	mov 0x10,%dx  // bss size
 	mov 0x0c,%si  // data size
 	mov 0x08,%bx  // text size
+	xor %di,%di   // far text size
 	mov 0x18,%ax  // entry point
 	mov %es,%cx   // data segment
 	mov %cx,%ds
@@ -558,72 +559,147 @@ sys_hdr_msg:
 	.ascii "Bad system header!"
 	.byte 0
 
+// Add AX to DS:SI and normalize segment
+add_ptr:
+	push %cx
+	add %si,%ax
+	mov	%ax,%si
+	and $15,%si
+	mov $4,%cl
+	shr %cl,%ax
+	mov %ds,%cx
+	add %cx,%ax
+	mov %ax,%ds
+	pop %cx
+	ret
+
+putc:
+	push %bx
+	push %cx
+	push %dx
+	mov   $0x0E,%ah
+	mov   $07,%bx
+	int   $0x10
+	pop %dx
+	pop %cx
+	pop %bx
+	ret
+
+hex1:
+	push %ax
+    and $0x0F,%al
+    add $'0',%al
+    cmp $'9',%al
+    jle 1f
+    add $('A'-'9'-1),%al
+1:  call putc
+	pop %ax
+	ret
+
+hex2:
+	push %ax
+	push %cx
+	push %dx
+    mov %al,%dl
+
+    mov $4,%cl
+    shr %cl,%al
+    call hex1
+    mov %dl,%al
+    call hex1
+	pop %dx
+	pop %cx
+	pop %ax
+	ret
+
+hex4:
+    push %ax
+    push %ax
+    mov %ah,%al
+    call hex2
+    pop %ax
+    call hex2
+	pop %ax
+	ret
+
+
 // System header is good
 
 sys_hdr_good:
 
 	mov %sp,%bp
-	mov 8,%dx     // code size
+	mov 8,%dx     // code size -2(%bp)
 	push %dx
 	add $0xF,%dx     // align on next paragraph (issue #209)
 	mov	$4,%cl
 	shr %cl,%dx   // DX = code segment size in paragraph
 	shr %cl,%bx   // BX = header size in paragraphs
 
-	mov 12,%ax    // data size
+	mov 12,%ax    // data size -(4%bp)
 	push %ax
-	mov 16,%ax    // bss size
+	mov 16,%ax    // bss size -6(%bp)
 	push %ax
-	mov 20,%ax    // entry point
+	mov 20,%ax    // entry point -8(%bp)
 	push %ax
 
 // Relocate kernel code
 
+	mov $'T',%ax
+	call putc
+
 	mov $SYSSEG,%ax
 	add %bx,%ax   // skip header
 	mov %ax,%ds
+	xor	%si,%si
+
 	mov $REL_SYSSEG,%ax
 	mov %ax,%es
+	xor %di,%di
+	call hex4
 
-	mov -2(%bp),%cx  // code size
-	xor	%si,%si
-	mov	%si,%di
+	mov -2(%bp),%cx  // code size in bytes
 	cld
 	rep
 	movsb
 
 // Relocate kernel far text
 
-	xor %ax,%ax   // far text size = 0
-	push %ax      // -10(%bp)
-	push %ax      // -12(%bp)
-	push %ax      // -14(%bp)
+	xor %ax,%ax   // far text size -10(%bp)
+	push %ax
+	push %ax      // text reloc size -12(%bp)
+	push %ax      // far text reloc size -14(%bp)
+	push %ax      // data reloc size -16(%bp)
 	cmp $0x04,%bx // extended header?
 	jne sys_no_far_text
 	mov $SYSSEG,%ax
 	mov %ax,%ds
-	mov 0x30,%ax  // far text size in bytes
+	mov 0x20,%ax  // text reloc size
 	mov %ax,-12(%bp)
-	mov 0x20,%ax  // text relocation size in bytes
+	mov 0x32,%ax  // far text reloc size
 	mov %ax,-14(%bp)
-	add $15,%ax
-	mov $4,%cl
-	shr %cl,%ax   // far text size in paragraphs
-	mov %ax,-10(%bp)
-
-	mov 0x30,%cx  // far text size in bytes
-	cmp %cx,%cx
+	mov 0x24,%ax  // data reloc size
+	mov %ax,-16(%bp)
+	mov 0x30,%cx  // far text size
+	mov %cx,-10(%bp)
+	and %cx,%cx
 	jz sys_no_far_text
-	mov $SYSSEG,%ax
-	add %bx,%ax   // skip header
-	add %dx,%ax   // skip code
-	mov %ax,%ds
-	mov $REL_SYSSEG,%ax
-	add %dx,%ax
-	mov %ax,%es
 
+	mov $'F',%ax
+	call putc
+
+	mov $SYSSEG,%ax
+	add %bx,%ax       // skip header
+	mov %ax,%ds
 	xor	%si,%si
-	mov	%si,%di
+	mov -2(%bp),%ax   // skip code bytes
+	call add_ptr
+
+	mov $REL_SYSSEG,%ax
+	add %dx,%ax       // skip code paragraphs
+	mov %ax,%es
+	xor %di,%di
+	call hex4
+
 	cld
 	rep
 	movsb
@@ -632,32 +708,105 @@ sys_no_far_text:
 // Relocate kernel data (not bss)
 // Kernel resets bss itself
 
+	mov $'f',%ax
+	call putc
+	mov -10(%bp),%ax // far text size
+	mov $4,%cl
+	shr %cl,%ax
+	mov %ax,%cx
+	call hex4
+
+	mov $'D',%ax
+	call putc
+
 	mov $SYSSEG,%ax
-	add %bx,%ax   // skip header
-	add %dx,%ax   // skip code
+	add %bx,%ax      // skip header
 	mov %ax,%ds
+	xor %si,%si
+	mov -2(%bp),%ax  // skip code bytes
+	call add_ptr
+	mov -10(%bp),%ax // far text size
+	call add_ptr
+	push %si
+
 	mov $REL_SYSSEG,%ax
-	add %dx,%ax
+	add %dx,%ax      // skip code paragraphs
+	add %cx,%ax      // skip far text
 	mov %ax,%es
+	call hex4
 
 	mov -4(%bp),%cx  // data size
-	xor %si,%si
-	add -12(%bp),%si // skip far text
 	xor %di,%di
 	cld
 	rep
 	movsb
 
+// Handle code/far text/data segment relocation
+
+	pop %si
+	mov -4(%bp),%ax  // skip data size
+	call add_ptr
+
+	push %es
+	mov -12(%bp),%cx // text reloc size
+
+text_reloc:
+	jcxz 1f
+	mov $'t',%ax
+	call putc
+	mov $REL_SYSSEG,%ax
+	mov %ax,%es      // ES -> text segment
+
+	call relocat
+	sub $8,%cx
+	add $8,%si
+	jmp text_reloc
+1:
+
+	mov -14(%bp),%cx // far text reloc size
+ftext_reloc:
+	jcxz 2f
+	mov $'f',%ax
+	call putc
+	mov $REL_SYSSEG,%ax
+	add %dx,%ax      // skip code
+	mov %ax,%es      // ES -> far text segment
+
+	call relocat
+	sub $8,%cx
+	add $8,%si
+	jmp ftext_reloc
+2:
+
+	mov -16(%bp),%cx // data reloc size
+data_reloc:
+	jcxz 3f
+	mov $'d',%ax
+	call putc
+	pop %es          // ES -> data segment
+	push %es
+
+	call relocat
+	sub $8,%cx
+	add $8,%si
+	jmp data_reloc
+3:
+	mov $'\r',%ax
+	call putc
+	mov $'\n',%ax
+	call putc
+
 // Load registers as kernel expects
 
+	pop %es          // ES -> data segment
 	mov %es,%ax
 	mov %ax,%ds
 
 	mov -2(%bp),%bx    // code size
-	add -12(%bp),%bx   // FIXME may overflow, needs farcode size
 	mov -4(%bp),%si    // data size
 	mov -6(%bp),%dx    // bss size
 	mov -8(%bp),%cx    // entry point
+	mov -10(%bp),%di   // far text size
 
 // Jump to kernel entry point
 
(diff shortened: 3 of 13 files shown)


ghaerr:
Added relocation routines to setup.S for handling text, far text and data relocations from gcc.

Kernel now boots and runswell when calling far functions. However, crashes when calling a near function from a far function (using gcc thunk).

All relocation seems to be performing correctly. Still having problems calling near procedures using thunk routine. Thunk routine was ported from ia16 gcc lib without change. I've traced down the relocations (only two in current kernel), one segment fixup in the near segment to call the farproc from start_kernel (works fine), and the second is the fixup in far text farproc when calling the thunk routine for calling the nearproc. The thunk routine is in the near text section. If the nearproc is changed to a far function, (as is the case in the current source), everything works fine.

@tkchia, do you have any ideas as to what I might be missing here?

ghaerr:
Found and fixed nearproc calling problem by adding .code16 directive to thunk routine. Didn't realize that gcc .S library code required that to work.

Also added fast thunk code when near functions with no arguments called from far functions.

Identified problem in kernel crt0.S and added .section .nildata to prevent the first .rodata string from being placed at address 0, and thus being confused with a NULL pointer. This fixes a long-standing sys_open("/dev/console",...) problem during kernel startup.

It seems that everything is now working - we are able to add far text procedures into the kernel, thus moving them out of the near-full .text segment. More testing will ensue as well as optimization of the setup.S copy and segment relocation routines.

tkchia:
Hello @ghaerr,

Cool!  It looks like kernels with far text are working quite well now. :-)

> Added thunking code to kernel library.

Regarding this: I suspect it might be easier in the long term to simply link against `libgcc` (i.e. `-lgcc`), rather than duplicate them in the kernel tree.

In my latest `gcc-ia16` release (https://github.com/tkchia/gcc-ia16/commit/606351cf50a27db562faa0768fbe760b89bbe5dc) I have added some thunk routines to specially handle far → near function calls with a small fixed argument count, and the GCC-generated code will call these.

> Unsure yet how to handle CONFIG_ROMCODE and !REL_SYS configurations.

As far as I can tell (`elks/include/linuxmt/config.h`), the `! REL_SYS` code seems to be used only when compiling for the Psion SIBO (?).  So it looks like we need to figure out what to do with this platform, or just unify the `REL_SYS` and `! REL_SYS` code to simplify things...

Thank you!

ghaerr:
> Cool! It looks like kernels with far text are working quite well now. :- )

Thanks. It took a bit to get working with no debugger, but I think its solid now. I'll be cleaning up the routines and then asking you for a code review.

I haven't yet been able to test relocating an S_DATA record. Do you know what exactly would cause compiler/linker to emit one?

 
> Regarding this: I suspect it might be easier in the long term to simply link against `libgcc` (i.e. `-lgcc`), rather than duplicate them in the kernel tree.

I think I mentioned that a couple years ago but there was a reason it wasn't done.


> In my latest `gcc-ia16` release ([tkchia/gcc-ia16@606351c](https://github.com/tkchia/gcc-ia16/commit/606351cf50a27db562faa0768fbe760b89bbe5dc)) I have added some thunk routines to specially handle far → near function calls with a small fixed argument count, and the GCC-generated code will call these.

Very much looking forward to seeing those! They will greatly help with execution time and code generated.

I plan on compiling all kernel init code, as a configurable option, into far procs. We will then have a better understanding of the frequency of use of each of the thunk variations, which I think will be useful for optimization of your thunks. Since we're limited to only a single final far code segment, it might be worth putting the filesystem code into the far segment as well, further saving lots of space. Ultimately, the init procedures could be named into a separate section, which the linker could place at the end of the far text segment, allowing release into the general memory pool after execution. This would put a hole in the middle of the kernel memory allocated, unless the far text were placed after the data segment initially, which I'm considering.



> > Unsure yet how to handle CONFIG_ROMCODE and !REL_SYS configurations.
> 
> As far as I can tell (`elks/include/linuxmt/config.h`), the `! REL_SYS` code seems to be used only when compiling for the Psion SIBO (?). So it looks like we need to figure out what to do with this platform

That makes some sense, as the ! REL_SYS code tries running the executable as-is, without any movement of the code or data segments. I don't think this can ever work reliably in the current system, as the data segment immediately follows a paragraph unaligned code segment in the minix image, prohibiting a proper DS pointer if (.texzt size & 15) != 0. With far text included, it seems the far text segment is always extended to paragraph alignment, correcting the data segment alignment problem.

I've been wondering what to do with the CONFIG_SIBO platform: its only use now seems to be knowing where to touch the kernel if porting to a new platform, but most of it is just hacked in, making the code visually messy.

> or just unify the `REL_SYS` and `! REL_SYS` code to simplify things...

Given the problem above and the current compiler, it doesn't seem that removing ! REL_SYS would hurt anything. 
Even if the SIBO were to be brought back to life, using the current compiler would allow for it to be relocated in RAM before startup just like REL_SYS does now, and placed at a different DEF_SYSSEG.

That would leave the system with REL_SYS for PC boots, and CONFIG_ROMCODE otherwise. 

Then, that too could be optimized into only checking CONFIG_ROMCODE set or not. The system is extremely confusing the way it is now, and I can't see the need for three sets of DEF_SYSSEG, REL_SYSSEG and CONFIG_ROM_KERNEL_CODE settings. The whole mess should be consolidated in config.h with a single set of defines used externally from there in any boot or setup.S code.

Thank you!

tkchia:
Hello @ghaerr,

> I haven't yet been able to test relocating an S_DATA record. Do you know what exactly would cause compiler/linker to emit one?

`S_DATA` relocations will be produced if the program makes references from some segment _to_ the data segment.

These relocations do not pop up by themselves very often (in most cases the compiler assumes it can get the data segment value from `%ds` or `%ss`).  But you can kind of force the compiler to ultimately emit an `S_DATA` relocation by saying something like
```
unsigned seg;
__asm volatile("movw $.data@OZSEG16, %0" : "=r" (seg));
```
(`OZSEG16` or `R_386_OZSEG16` is the relocation code used in the intermediate ELF `.o` file.)

ELKS relocations have yet to undergo much testing, so I would be interested to know of any bugs which you may find. :-)

> I think I mentioned that a couple years ago but there was a reason it wasn't done.

OK.  I guess I was not in on the reason, whatever it was.

> Very much looking forward to seeing those! They will greatly help with execution time and code generated.

It is tested and ready to go.  You can bump the GCC version number in `tools/Makefile` to try it out if you wish.  (You will probably also need to copy the extra thunk routines into the kernel tree, of course.)

Thank you!

ghaerr:
Cleanup segment copy and relocation routines.
Cleanup some setup.S routines for better readability and code flow.

PR is almost ready to commit. Works with normal .text-only and .fartext sections in a.out headers.

ghaerr:
Hello @tkchia,

I decided to fixup build.c at the last minute to properly display the far text size, as the last item in this before commit. Doing this required adding knowledge of the new supplemental exec header fields to build.c. However, build.c uses its own a.out.h, which is a problem, but I'll fix that in a separate PR. The bigger problem I realized after a while: build.c is a host program, thus a 32-bit program; I copied the ELKS declaration for it  into its a.out.h. Here's ELKS version:
```
struct minix_exec_hdr {
    unsigned long   type;
    unsigned char   hlen;           // 0x04
    unsigned char   reserved1;
    unsigned short  version;
    unsigned short  tseg, reserved2;    // 0x08
    unsigned short  dseg, reserved3;    // 0x0c
    unsigned short  bseg, reserved4;    // 0x10
    unsigned long   entry;
    unsigned short  chmem;
    unsigned short  minstack;
    unsigned long   syms;
};

struct elks_supl_hdr {
    /* optional fields */
    unsigned long   msh_trsize; /* text relocation size */  // 0x20
    unsigned long   msh_drsize; /* data relocation size */  // 0x24
    unsigned long   msh_tbase;  /* text relocation base */
    unsigned long   msh_dbase;  /* data relocation base */
    /* even more optional fields --- for ELKS medium memory model support */
    unsigned short  esh_ftseg;  /* far text size */     // 0x30
    unsigned long   esh_ftrsize;    /* far text relocation size */  // 0x32
    unsigned short  esh_reserved1;
    unsigned long   esh_reserved2, esh_reserved3;
};
```
The problem is with the esh_ftseg declaration: on a 32-bit machine, the packing of a short causes a compiler insertion of another short before the long esh_ftrsize, which breaks build.c unless special packing is used.

I am thinking that it would be a good idea for exec headers to be portable across 16-bit and 32-bit compilations. This was achieved with the original a.out.h Linux header, which was apparently used by Linus when he originally wrote build.c. That same header appears to have been pulled into dev86, but now we've changed it. Here's the revised header from elks/arch/i86/tools/a.out.h where build.c lives:
```
truct exec {           /* a.out header */
    unsigned char a_magic[2];   /* magic number */
    unsigned char a_flags;  /* flags, see below */
    unsigned char a_cpu;    /* cpu id */
    unsigned char a_hdrlen; /* length of header */
    unsigned char a_unused; /* reserved for future use */
    unsigned short a_version;   /* version stamp (not used at present) */
    int32_t a_text;     /* +8:  size of text section in bytes */
    int32_t a_data;     /* +12: size of data section in bytes */
    int32_t a_bss;      /* +16: size of bss  section in bytes */
    int32_t a_entry;    /* +20: entry point */
    int32_t a_total;    /* +24: total memory allocated */
    int32_t a_syms;     /* +28: size of symbol table */

    /* SHORT FORM ENDS HERE */
    uint32_t a_trsize;      /* text relocation size */
    uint32_t a_drsize;      /* data relocation size */
    uint32_t a_tbase;       /* text relocation base */
    uint32_t a_dbase;       /* data relocation base */
    /* even more optional fields --- for ELKS medium memory model support */
    uint16_t esh_ftseg;     /* far text size */
    uint16_t esh_ftrsize;   /* far text relocation size */ //FIXME should be uint32
    uint32_t esh_reserved1;
    uint32_t esh_reserved2;
    uint32_t esh_reserved3;
};
```
For now, I have shortened the with of `esh_ftrsize` to 16 bits (this only affects build.c display, but I'm suggesting a broader fix). I think ideally, `esh_ftseg` should also be long, and keep `esh_ftrsize` as long. This would require a compiler and setup.S change to be done after this commit. Not doing anything means messing around like this and not being able to use our minix.h header outside ELKS easily. On another small note, I think the structure should be moved out of minix.h and put into an a.out.h so that it can be used by tools (like build). 

What do you think?

I would like to commit this PR as-is, after your approval, and deal with this problem separately, in any case.

Thank you!

tkchia (approved):

tkchia:
Hello @ghaerr,

> I think ideally, `esh_ftseg` should also be long, and keep `esh_ftrsize` as long.

OK.  Anyway, my original thinking was to allot 32 bits to `esh_ftrsize`, because each relocation entry is rather large (8 bytes), and limiting it to 16 bits will mean there can be at most 8,191 relocation entries for the far text segment, which is kind of limiting.

I am not sure if there is any good use for `esh_ftseg` being more than 16 bits.  It seems this will just complicate things for `setup.S` and the kernel's `execve` routine?

In fact I am now wondering if we should instead stash the far text size field just after the near text size field (`tseg`) --- since that space is currently unused anyway --- and then bump up the `version`.  What are your thoughts on this?

> The problem is with the esh_ftseg declaration: on a 32-bit machine, the packing of a short causes a compiler insertion of another short before the long esh_ftrsize, which breaks build.c unless special packing is used.

For specifying unaligned structure fields, as you know, the solution I used for `elksemu` was to specify `__attribute__ ((packed))`, which should work for both host and target.

Thank you!

ghaerr:
Hi @tkchia,

> OK.  Anyway, my original thinking was to allot 32 bits to `esh_ftrsize`, because each relocation entry is rather large (8 bytes), and limiting it to 16 bits will mean there can be at most 8,191 relocation entries for the far text segment, which is kind of limiting.

Completely agree.

> I am not sure if there is any good use for `esh_ftseg` being more than 16 bits.  It seems this will just complicate things for `setup.S` and the kernel's `execve` routine?

Actually, I think it is a possible mistake to have any size fields less than 32 bits, for a variety of reasons. I propose the following, which would be easily handled by setup.S and sys_execv:

Move `tseg`, `dseg`, `bseg`, and `ftseg` back to unsigned longs. All routines, as they already do, will treat them as (intel order) unsigned shorts. Specifically, this is easily done in sys_execv by accessing `tseg` as `(size_t)tseg`, etc. In setup.S, we'll just use the low-word offset for these fields, which is what is being done now. Only one offset changes - esh_ftrsize.

> In fact I am now wondering if we should instead stash the far text size field just after the near text size field (`tseg`) --- since that space is currently unused anyway --- and then bump up the `version`.  What are your thoughts on this?

No, I don't think squishing ftseg next to tseg is a good idea - it moves one of the required far text components out of the supl_header, forces a version bump, and locks sizes into 16 bits. The above suggestion does not, as it returns the a.out header fields to their original size which there has and would be no interpretation change anyways, so no version bump required.

Should you agree, we could put this into the next compiler chain update, some easy source changes in exec.c, and one offset change in setup.S for the increased offset of esh_ftrsize (which would remain a long).

BTW - I spent many hours last night moving quite a few routines into a .fartext.init segment - and found that there are problems. I suspect, but do not know yet, there are tool bugs. We could likely need a compiler update anyways. More on that later. I plan on committing this basic work, then testing further in user mode with mixed-model segments with larger programs.

> For specifying unaligned structure fields, as you know, the solution I used for `elksemu` was to specify `__attribute__ ((packed))`, which should work for both host and target.

Yes, I was trying to avoid that wherever the header was used outside of ELKS kernel.

Thank you!

tkchia:
Hello @ghaerr,

> BTW - I spent many hours last night moving quite a few routines into a .fartext.init segment - and found that there are problems. I suspect, but do not know yet, there are tool bugs. We could likely need a compiler update anyways.

Well, yes, it happens that I am now trying to do some sort of major update of `binutils-ia16`.  I have occasionally getting random crashes or other weirdness in Binutils, and I am having difficulty tracking them down (no thanks to `ld.bfd`'s spaghetti control flow...).  So now I am trying to 

  * rebase my IA-16 patches on top of a more recent version of mainline Binutils,
  * switch from the default `ld.bfd` linker to `ld.gold`, which seems more straightforward to deal with,

among other things.

Unfortunately the changes may cause some disruption on the ELKS end --- e.g. the link process will probably need rewriting --- but I am hoping to reduce the disruption.

Thank you!

ghaerr:
> Unfortunately the changes may cause some disruption on the ELKS end --- e.g. the link process will probably need rewriting --- but I am hoping to reduce the disruption.

Sounds like you have your hands full.

How about if I commit this PR now, and then, prepare a PR with the a.out changes I propose, with the sys_execv changes made and working, for your consideration? Then everything will be fixed on this side (with the exception of the new offset for esh_ftrsize). No compiler changes needed. That will straighten out the problems over here.

I'll continue working on identifying where the tool chain or other problems are with the far text issues, and when/if you have time to make the esh_ftrsize offset change on your side, a compiler chain update and slight change to setup.S can be made and we'll be done.

Thank you!

tkchia:
@ghaerr: That should be fine.  Thanks!

ghaerr:
Hello @tkchia,

I'm starting to test far text segments in the kernel again, this time with the new linker.

> I suspect it might be easier in the long term to simply link against libgcc (i.e. -lgcc), rather than duplicate them in the kernel tree.

I'm thinking, at least at this point, that it would indeed be better to link against the cross-gcc library, rather than import the thunk routines into ELKS. (And nice job with the fast thunks, they will greatly help :)

Adding `-lgcc` won't quite work as recommended because in the last steps of the kernel build, we're invoking ld.gold directly, and it doesn't know about that. So, my kernel build is dying with a bunch of undefined references to the thunk routines.

Can you tell me exactly what should be specified in the link line below, rather than -lgcc, in order to grab the correct versions?
```
# Begin PC image build

boot/system:    $(XINCLUDE) $(AARCHIVES) $(ADRIVERS) boot/crt0.o
    (cd $(BASEDIR) ; $(LD) $(LDFLAGS) -M $(ARCH_LD) -T $(TOPDIR)/elks/elks-small.ld \
        $(ARCH_DIR)/boot/crt0.o \
        init/main.o '-(' $(ARCHIVES) $(DRIVERS) '-)' \
        -o $(ARCH_DIR)/boot/system -lgcc > $(ARCH_DIR)/boot/system.map ; \
        $(POSTLINK) $(ARCH_DIR)/boot/system)
```

[EDIT: I looked around a bit and added '$(TOPDIR)/cross/build/gcc-build/ia16-elf/libgcc/libgcc.a', but get the following error]
```
cd ../.. ; ia16-elf-ld.gold  -s -M  -T /Users/greg/net/elks-gh/elks/elks-small.ld \
		arch/i86/boot/crt0.o \
		init/main.o '-(' fs/fs.a kernel/kernel.a lib/lib.a net/net.a fs/minix/minixfs.a fs/msdos/msdos.a arch/i86/kernel/akernel.a arch/i86/lib/lib86.a arch/i86/mm/mm.a  arch/i86/drivers/char/chr_drv.a arch/i86/drivers/block/blk_drv.a arch/i86/drivers/net/net_drv.a '-)' \
		/Users/greg/net/elks-gh/cross/build/gcc-build/ia16-elf/libgcc/libgcc.a \
		-o arch/i86/boot/system > arch/i86/boot/system.map ; \
		elf2elks arch/i86/boot/system)
elf2elks: error: R_386_SEGRELATIVE relocation has bad in-place addend 0
make[2]: *** [boot/system] Error 1
```
Placing the libgcc.a within the -( ... -) also produces the same resulting error.

[EDIT2: I just re-read the error, and it seems perhaps that the thunks are in fact being linked, but elf2elks is running into a relocation it can't handle... if that is the case, I can produce a smaller test case for a .fartext kernel compile.]

Thank you!

ghaerr:
Hello @tkchia,

Indeed, the `elf2elks` program seems unable to handle a .fartext relocation when building the ELKS kernel. After adding a single .fartext procedure (see diff below), it fails with `elf2elks: error: R_386_SEGRELATIVE relocation has bad in-place addend 0`. This used to work with the old linker.

```
Gregs-MBP2:elks-gh greg$ git diff
diff --git a/elks/arch/i86/Makefile b/elks/arch/i86/Makefile
index a3997c80..77ce10b7 100644
--- a/elks/arch/i86/Makefile
+++ b/elks/arch/i86/Makefile
@@ -113,6 +113,7 @@ boot/system:        $(XINCLUDE) $(AARCHIVES) $(ADRIVERS) boot/crt0.o
        (cd $(BASEDIR) ; $(LD) $(LDFLAGS) -M $(ARCH_LD) -T $(TOPDIR)/elks/elks-small.ld \
                $(ARCH_DIR)/boot/crt0.o \
                init/main.o '-(' $(ARCHIVES) $(DRIVERS) '-)' \
+               $(TOPDIR)/cross/build/gcc-build/ia16-elf/libgcc/libgcc.a \
                -o $(ARCH_DIR)/boot/system > $(ARCH_DIR)/boot/system.map ; \
                $(POSTLINK) $(ARCH_DIR)/boot/system)
 #              sort -k3,4 $(ARCH_DIR)/boot/system.tmp > $(ARCH_DIR)/boot/system.map ; \
diff --git a/elks/init/main.c b/elks/init/main.c
index a9ca3646..ac46d594 100644
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -54,6 +54,11 @@ static void init_task(void);
 extern int run_init_process(char *cmd);
 extern int run_init_process_sptr(char *cmd, char *sptr, int slen);
 
+void FARPROC test(void)
+{
+       printk("hello far kernel\n");
+}
+
 void start_kernel(void)
 {
     seg_t base, end;
@@ -97,6 +102,7 @@ void start_kernel(void)
 #endif
 
     mm_stat(base, end);
+       test();
 
     kfork_proc(init_task);
     wake_up_process(&task[1]);
Gregs-MBP2:elks-gh greg$ 
```
Here's the build failure:
```
(cd ../.. ; ia16-elf-ld.gold  -s -M  -T /Users/greg/net/elks-gh/elks/elks-small.ld \
		arch/i86/boot/crt0.o \
		init/main.o '-(' fs/fs.a kernel/kernel.a lib/lib.a net/net.a fs/minix/minixfs.a fs/msdos/msdos.a arch/i86/kernel/akernel.a arch/i86/lib/lib86.a arch/i86/mm/mm.a  arch/i86/drivers/char/chr_drv.a arch/i86/drivers/block/blk_drv.a arch/i86/drivers/net/net_drv.a '-)' \
		/Users/greg/net/elks-gh/cross/build/gcc-build/ia16-elf/libgcc/libgcc.a \
		-o arch/i86/boot/system > arch/i86/boot/system.map ; \
		elf2elks arch/i86/boot/system)
elf2elks: error: R_386_SEGRELATIVE relocation has bad in-place addend 0
make[2]: *** [boot/system] Error 1
```

The error is occurring in the convert_reloc function, where it is seemingly unable to pair up the relocation with one of the S_TEXT, S_FTEXT or S_DATA section addresses. I don't understand this part of elf2elks well enough to proceed with a fix.

Thank you!

tkchia:
Hello @ghaerr,

Do you still have a copy of the (ELF-format) `elks/arch/i86/boot/system`?  If you upload it I will see if I can figure out what is up.  I do not have any problems adding far routines to the kernel on my end.

> Can you tell me exactly what should be specified in the link line below, rather than -lgcc, in order to grab the correct versions?

Probably the best way is to ask `ia16-elf-gcc` for the correct path.  Pass the option `-print-libgcc-file-name`, along with the options used for compilation, to `ia16-elf-gcc`, then use the output file name:
```
$(LD) ... '-(' ... '-)' "`$(CC) $(CFLAGS) -print-libgcc-file-name`" -o ...
```

Thank you!

ghaerr:
Hello @tkchia,

Here's the system, as well as the patch that produced it.

I'll add the -print-libgcc-file-name in a separate PR, once this gets resolved.

Thanks!
[system.zip](https://github.com/jbruchon/elks/files/5256665/system.zip)
[farproc.patch.zip](https://github.com/jbruchon/elks/files/5256669/farproc.patch.zip)

ghaerr:
Hello @tkchia,

Oops - it seems that binutils-ia16 version might have gotten bumped by you during the time the ld.gold PR was open.

I'm showing I have binutils-ia16-819e9646a9b2f3f32286ff0e11ff13829e3e6eb8, while it seems the tools/build.sh version specifies f19a905d1434da03940951913514a9ef12f0d7dd. I didn't remake the cross compiler after the PR was committed.

This is likely the issue?

Thank you!

ghaerr:
Hello @tkchia,

Sorry for the confusing messages... I have switched to my laptop which does have the proper binutils, and still get the elf2elks error. Please find attached a system produced by it, with the same patch.

Thank you!
[system.zip](https://github.com/jbruchon/elks/files/5256713/system.zip)

tkchia:
Hello @ghaerr,

Thanks!  Interesting...
```
$ ia16-elf-readelf -r system

Relocation section '.rel.dyn' at offset 0xf680 contains 2 entries:
 Offset     Info    Type            Sym.Value  Sym. Name
00010360  00000030 R_386_SEGRELATIVE
0002000a  00000030 R_386_SEGRELATIVE
```
The far routine reads
```
$ ia16-elf-objdump -D -r -Mi8086 system | less
...
00020000 <.fartext>:
   20000:       b8 76 00                mov    $0x76,%ax
   20003:       50                      push   %ax
   20004:       bb f3 2e                mov    $0x2ef3,%bx
   20007:       9a 6f e6 00 00          lcall  $0x0,$0xe66f
   2000c:       83 c4 02                add    $0x2,%sp
   2000f:       cb                      lret   
```
where for some reason the segment value before relocation says `$0x0` when it should be `$0x1000` (`0x1000 * 0x10` is the starting offset of `.text` in the ELF file).  The offset `$0xe66f` is correct and refers to this thunk routine:
```
   1e66f:       55                      push   %bp
   1e670:       89 e5                   mov    %sp,%bp
   1e672:       ff 76 06                pushw  0x6(%bp)
   1e675:       ff d3                   call   *%bx
   1e677:       89 ec                   mov    %bp,%sp
   1e679:       5d                      pop    %bp
   1e67a:       cb                      lret   
```
The other relocation is in the `.text` segment at the place where the initialization code calls out to the far routine:
```
   1035d:       9a 00 00 00 20          lcall  $0x2000,$0x0
```
This part is also correct.

I will need to look at this more closely.

Thank you!

ghaerr:
Hello @tkchia,

> where for some reason the segment value before relocation says $0x0 when it should be $0x1000 (0x1000 * 0x10 is the starting offset of .text in the ELF file)

Could this be for some reason that the FARPROC test() happens to be the very first .fartext routine, and is associated as address 0 even though the linker script later puts it at 0x10000 ? Perhaps adding another FARPROC routine would help, so there isn't just one at possible address 0.

Thank you!

tkchia:
Hello @ghaerr,

Not quite --- the wrong value is where the far routine `test()` calls _back_ to the near text segment.  The near text segment is placed at offset `0x10000` in the ELF file, while the far text segment is at `0x20000`.

tkchia:
Hello @ghaerr,

May I know if you also have the `system.map` from the build?  I am wondering if the segment base symbol for the thunk routine `__ia16_far_call_thunk.bx.2!` was erroneously set to 0 on your end (it should be `0x10000`).

Thank you!

ghaerr:
Here is the system.map, the segment base looks correct.

[system.map.zip](https://github.com/jbruchon/elks/files/5256907/system.map.zip)

tkchia:
Hello @ghaerr ,

Thanks!  `__ia16_far_call_thunk.bx.2` is in the right place, but the symbol table is missing a `__ia16_far_call_thunk.bx.2!` symbol (with the exclamation mark).

> > ```
> > $(LD) ... '-(' ... '-)' "`$(CC) $(CFLAGS) -print-libgcc-file-name`" -o ...
> > ```

> I'll add the -print-libgcc-file-name in a separate PR, once this gets resolved.

I now suspect that this particular change might actually fix your issue.  I just remembered that `cross/build/gcc-build/ia16-elf/libgcc/libgcc.a` is the library for linking with default options, i.e. for an MS-DOS target.  I have not yet brought the MS-DOS target over to the new relocation representation (as I have done with ELKS), mainly due to historical reasons.

I.e. you need to link against a different `libgcc.a`.

Using `$(CC) $(CFLAGS) -print-libgcc-file-name` should lead you to the correct library.  This will be something like ...`/cross/lib/gcc/ia16-elf/6.3.0/elkslibc/libgcc.a`.

Thank you!

ghaerr:
Hello @tkchia,

That indeed fixes the problem, thank you!

I should have waited, rather than guessing about a libgcc.a location!! :)

On another note, I remain a bit worried about the ld.gold linking w/no errors when a symbol isn't present - this could cause a number of future issues, don't you think? Or is this just one of hazards with linking with an external library for the kernel build?

(What is the purpose of all the new xxx! symbols?

I'll submit more test results shortly, now that this is figured out.

Thank you!

ghaerr:
Particularly:
```
Disassembly of section .fartext.f.main.c.53160$:

00000000 <test>:
   0:   b8 72 00                mov    $0x72,%ax
            1: R_386_16 .rodata.str1.1
            1: R_386_SUB16  .rodata.str1.1!
   3:   50                      push   %ax
   4:   bb 00 00                mov    $0x0,%bx
            5: R_386_16 printk
            5: R_386_SUB16  printk!
   7:   9a 00 00 00 00          lcall  $0x0,$0x0
            8: R_386_16 __ia16_far_call_thunk.bx.2
            8: R_386_SUB16  __ia16_far_call_thunk.bx.2!
            a: R_386_SEG16  __ia16_far_call_thunk.bx.2!
   c:   83 c4 02                add    $0x2,%sp
   f:   cb                      lret
```
Any compiler-emitted call to a xxx! address won't necessarily be picked up or tagged by the new linker (if missing).

ghaerr:
Hello @tkchia,

Well, I moved to my much larger kernel patch, which before the linker change compiled and linked, but didn't run.

When I apply that (slightly modified) patch, I get the same `elf2elks: error: R_386_SEGRELATIVE relocation has bad in-place addend 0` error message, using the proper libgcc.a library. This (large and currently ugly) patch tries to convert quite a few kernel init routines to a new "INITSEG" (bad name) section (#define INITSEG     __far __attribute__ ((far_section, noinline, section (".fartext.init")))".

I am not sure why this is happening, other than ld.gold is missing a xxx! symbol, although I now have no idea why.
I am assuming the reason for this is likely that I have perhaps forward-declared a routine INITSEG perhaps, and maybe missed the INITSEG declaration, or perhaps .fartext.init doesn't work. But the problem now is that ld.gold doesn't give a link error, so I have no idea why.

I've attached the patch just to give you an idea of it (ignore cruft due to bad naming of INITSEG for now).

What do you think, can ld.gold be changed to emit a link error? Or do you think this is yet a different problem?

Thank you!
[initseg.patch.zip](https://github.com/jbruchon/elks/files/5257335/initseg.patch.zip)

tkchia:
Hello @ghaerr,

> What is the purpose of all the new xxx! symbols?

An _x_`!` symbol marks the segment base for the symbol _x_.  E.g. a symbol `foo` which might appear in a classical IA-16 linker map file as, say, `0x1234:0x5678`, will be represented under the new `segelf` scheme as

  * `foo = 0x12340 + 0x5678 = 0x179b8`  # `foo`'s complete linear address
  * `foo! = 0x12340`  # `foo`'s segment base

Mr. H. P. Anvin [came up with this scheme](https://git.zytor.com/users/hpa/segelf/abi.git/plain/segelf.txt) as a way to represent _segment_`:`_offset_ symbols under the ELF format, which only really knows about flat addresses.

(In my [earlier representation scheme](https://github.com/tkchia/build-ia16/blob/master/elf16-writeup.md), the value of `foo` would simply be `0x5678`, there is no `foo!`, and the segment component would be represented in a sort of "out-of-band" way.  This scheme is (well) somewhat simpler, but has some limitations.)

> On another note, I remain a bit worried about the ld.gold linking w/no errors when a symbol isn't present - this could cause a number of future issues, don't you think? Or is this just one of hazards with linking with an external library for the kernel build?

Currently the missing _x_`!` symbols are "weakly defined", which means they will get a value of 0 if they are not explicitly defined.

I am starting to wonder though if it is a bad idea to do this --- I might change this behaviour in the future.

Thank you!

ghaerr:
Hello @tkchia,

Thank you for your explanation, that really helps. I started looking at ld.gold source but found it a bit complicated, and never realized that x! symbols are weakly defined.

> I am starting to wonder though if it is a bad idea to do this --- I might change this behaviour in the future.

I think perhaps it might be asking for trouble, especially when linking with external libraries "not in know", as we've already seen. When a linker does not give an error on a symbol not found in a requested relocation emitted by the compiler, the generation of an invalid relocation, found only by a post-processing file-format converter, could make debugging things much harder in larger systems in the future.

Looking at the following code:
```
Disassembly of section .fartext.f.main.c.53160$:

00000000 <test>:
   0:   b8 72 00                mov    $0x72,%ax
            1: R_386_16 .rodata.str1.1
            1: R_386_SUB16  .rodata.str1.1!
   3:   50                      push   %ax
   4:   bb 00 00                mov    $0x0,%bx
            5: R_386_16 printk
            5: R_386_SUB16  printk!
   7:   9a 00 00 00 00          lcall  $0x0,$0x0
            8: R_386_16 __ia16_far_call_thunk.bx.2
            8: R_386_SUB16  __ia16_far_call_thunk.bx.2!
            a: R_386_SEG16  __ia16_far_call_thunk.bx.2!
   c:   83 c4 02                add    $0x2,%sp
   f:   cb                      lret
```
Would it possible to, rather than generating tons of xxx! symbols (which would also greatly simplify the .map file), instead have the compiler make an assumption about just either the .data!, .text!, or .fartext! section the symbol belongs in, at least for now (since there's never multiple data sections and only at most one .fartext section)? The idea would be to use the normal xxx symbol for the R_386_16 and R_386_SUB16 relocations, and one of the .data!/.text!/.fartext! symbols only for the R_386_SEG16 relocation. [EDIT: In this way, there wouldn't ever be a missing xxx! symbol causing problems, since only three weakly defined symbols would be needed for the linker.]

Now, for sources compiled together by gcc and linked, (for instance, the kernel and now libgcc), there should always be an xxx! symbol, so I'm thinking perhaps this isn't as big a problem as my post above described. It seems we've had a number of problems (first the original linker crashing or emitting bad relocations, then the kernel stack being overwritten), that have impeded moving forward quickly with a mixed-mode kernel. I plan on pressing forward, using smaller commits, getting the first version of a configurable INITPROC (.fartext.init section) kernel compiling and running.

That said, I still don't see how an "xxx!" type relocation symbol was apparently missing when building from my large patch. But I'm sure we'll run into it again as we further press on!

Thank you!

ghaerr:
Hello @tkchia,

More specifically about the compiler making an assumption on the R_386_SEG16 section a given symbol is in:
```
p = &foo;
```
Since foo needs to have some sort of forward declaration, the compiler will know whether foo is data or code; if data, then .data! would be used.

If code, then the compiler needs to have a proper forward declaration of foo being either far or not. Otherwise, 1) obviously a call to the procedure could fail if mismatchedl, and 2) the storage size needs to be known for the above pointer initialization. Thus, either .text! or .fartext! would be used accordingly.

If multiple .fartext or .text segments were supported, this wouldn't work. But since there's at most one of either (as well as .data), it seems (guessing since I am not very familiar with gcc internals) that such a thing could work.

Thank you!

Issue #729, opened 2020-08-27
Kernel stated as 'non re-entrant" ?


mfld-fr:
I am a bit surprised to read in PR #695 that the ELKS kernel is stated as 'non re-entrant'. Indeed, this looks true, but only for the system calls.

As far as I can see in the interrupt handling code, an asynchronous call can occur at any moment upon an HW interrupt, but a task switch can only occur when the global interrupt level goes back to 0 at the end of the interrupt handler.

It means that the kernel code is actually re-entrant for HW interrupt handling, but not for system calls.

This looks like to be an unnecessary & strong limitation, because I see no reason in the interrupt handler to forbid a task switch when returning from any HW interrupt to any system call ?

ghaerr:
Hello @mfld-fr,

It appears from after the first submission of this issue, through the subsequent edits, you have figured out that the ELKS kernel is indeed non-reentrant, in that hardware interrupts during kernel code execution will never result in a task switch, by design. This means that a kernel system call will never be interrupted, which has big benefits, explained below.

> As far as I can see in the interrupt handling code, an asynchronous call can occur at any moment upon an HW interrupt, but a task switch can only occur when the global interrupt level goes back to 0 at the end of the interrupt handler.

That is correct. In addition, any interrupt occurring during kernel execution, or during another interrupt (same thing), executes on a special "interrupt stack" instead of the kernel stack in the task structure.

The exact details of how this works are explained in two articles I wrote on the wiki: [System Calls & Interrupts Part I,](https://github.com/jbruchon/elks/wiki/System-calls-and-interrupts) [Part 2](https://github.com/jbruchon/elks/wiki/Syscalls-&-interrupts,-Part-II).

> It means that the kernel code is actually re-entrant for HW interrupt handling, but not for system calls.

Only kernel routines registered through `request_irq` need worry about reentrancy, along with routines called from them, which are quite limited. The vast majority of the kernel is not reentrant, nor needs to be, the major benefit being not having to be protected from reentrancy. Kernel interrupt routines can never call the sleep routines, for instance.

> This looks like to be an unnecessary & strong limitation, 

It is indeed a strong limitation. For single-CPU, non-realtime systems, there aren't any good reasons for the kernel to be reentrant. Should the clock tick during a system call, the process won't get rescheduled, but allowed to run until the next clock tick occurring in "user mode". 

Some of the many benefits of non-reentrancy include:
[...]

Issue #709, opened 2020-08-16
Enhancement: support for noncontiguous memory blocks


pawosm-arm:
I've managed to buy lo-tech's UMB memory 8-bit ISA card for my Amstrad PC 2086. It can fill unused chunks of upper memory addresses with RAM. It works with FreeDOS where `USE!UMBS.SYS` driver must be used. As this driver is unable to autodetect where upper memory RAM blocks are located, it needs to be patched before use (it's easy with the user manual provided). Each block is defined by hardware address (in upper memory area) and size.

To achieve the same with ELKS, ability to define memory map (allowing to have holes between available memory blocks) in `.config` would be required. A page granularity should be ok.

ghaerr:
The ELKS memory manager is elks/arch/i86/mm/malloc.c, which you might want to take a look at.

In the function mm_init, the beginning and end of conventional memory is passed at boot, and a single "free segment" in the MM segment list is created. It would not be hard to add additional segments (paragraph boundary granularity) through perhaps a new mm_add function, which would add additional free segments, being passed an address and length.

However, the MM does try to combine/merge segments on free, which would require some careful modifications to the segment flag bit to prohibit that, since that won't work with your card memory.

I assume there are no hardware access issues with this card, other than knowing the address and length of each memory block allocated, since it doesn't require the A20 line fiddling etc required for EMS.

Adding upper memory capability to ELKS may be worth the effort, since no special code other than enhancing the memory manager as described is needed. However, I don't have access to such a card for development or testing. If you'd like to take a pass at it, study malloc.c and I can help you.

If the card could be set such that the upper memory start address is just above your machine's exiting (640k?) memory, I am thinking some very simple modifications could be made that would allow the memory scan performed in setup.S to just add the UMA to the 640k and pass the entire contiguous memory size to mm_init.

pawosm-arm:
> If the card could be set such that the upper memory start address is just above your machine's exiting (640k?) memory, I am thinking some very simple modifications could be made that would allow the memory scan performed in setup.S to just add the UMA to the 640k and pass the entire contiguous memory size to mm_init.

Unfortunately, VGA memory starts right after 640k boundary and it should not be used for general purposes. The first free block I could find is at D000h, right after area occupied by lo-tech's XT-IDE BIOS.

ghaerr:
> Unfortunately, VGA memory starts right after 640k boundary and it should not be used for general purposes.

[...]

Issue #708, opened 2020-08-16
Enhancement request: EMS memory support


pawosm-arm:
I've managed to buy lo-tech's 2MB EMS memory 8-bit ISA card for my Amstrad PC 2086. It exposes EMS protocol version 4.0 and works lovely under FreeDOS (although I had to replace default command line interpreter with 4DOS which can utilize EMS memory for its internal data freeing up a chunk of so precious main memory). Things that didn't work before now started to work thanks to that!

I guess gaining access to some extra 2MB RAM could boost ELKS dramatically. It is accessed through 64kB memory window which can be configured to start at selected address in the upper memory address space (usually E000h). The DOS driver and its source code (8086 assembly) is available here: https://www.lo-tech.co.uk/downloads/2MB-EMS-Board/LTEMM-r01.zip

ghaerr:
I investigated EMS pretty carefully several months ago when I was adding the configurable external L2 buffers, and came to the conclusion that, because of the amount of code and special handling required to access extended memory, that such a thing didn't really help ELKS that much. This is because, unlike upper memory (UMA - between 640k and 1MB), extended memory (EMS) cannot be used to run ELKS applications in, but instead only contain more L2 external system buffers. The extended system buffering would help speed by not requiring disk I/O in some cases, of course, but no more applications could be run, and ELKS is already set to only 16 processes max for other reasons.

In order to access EMS, the A20 line has to be toggled, and in addition the CPU must be put into protected mode, which is only supported by 80286+ processors. The whole thing is pretty complex, and would require more kernel space (we're down to 8k left currently) that it's worth, IMO. The other issue we have with ELKS is that the kernel can't even access the L2 external buffers directly, all access has to them has to be copied in and out to yet another local set of L1 buffers, since the kernel doesn't use any far pointers. A close examination of a full read/write sequence to disk or the network shows way more copying than is needed for efficient operation. This would want a rewrite also, before implementing EMS.

pawosm-arm:
@ghaerr, your description of the challenges fits to the XMS (exTended memory) which indeed requires 286+, not EMS (exPanded memory) which came to the market much earlier and works with any 8088/8086 (and later) PC. The LTEMM driver for DOS is for all of the PC computers, including old XT's and it definitely does not require protected mode. Even Windows 3.0 (which was the last Windows version to support EMS) could make use of EMS memory only in real mode (started by `win /r`).

ghaerr:
[...]

PR #744, opened 2020-09-07
[kernel] Improve heap allocator

mfld-fr:
Add a list of the free blocks to speed up allocation in the local heap.
Tested under default configuration and QEMU in my fork.
Proposed for review, but still a glitch in the free list order, so please wait for the end of review before merging.

### DIFF
--- a/elks/arch/i86/drivers/char/mem.c
+++ b/elks/arch/i86/drivers/char/mem.c
@@ -168,7 +168,6 @@ int kmem_ioctl(struct inode *inode, struct file *file, int cmd, char *arg)
 {
     unsigned short retword;
     struct mem_usage mu;
-    extern heap_s *_heap_first;
 
     switch (cmd) {
 #ifdef CONFIG_MODULES
@@ -198,7 +197,7 @@ int kmem_ioctl(struct inode *inode, struct file *file, int cmd, char *arg)
 #endif
 	return 0;
     case MEM_GETHEAP:
-	retword = (unsigned short)_heap_first;
+	retword = (unsigned short) &_heap_all;
 	break;
     default:
 	return -EINVAL;
--- a/elks/arch/i86/kernel/system.c
+++ b/elks/arch/i86/kernel/system.c
@@ -47,6 +47,7 @@ void setup_arch(seg_t *start, seg_t *end)
 	unsigned int endbss = (unsigned int)(_endbss + 1) & ~1;
 
 	/* Now insert local heap at end of kernel data segment */
+	heap_init ();
 	heap_add ((void *)endbss, 1 + ~endbss);
 
 	/* Misc */
--- a/elks/include/linuxmt/heap.h
+++ b/elks/include/linuxmt/heap.h
@@ -10,6 +10,8 @@
 
 // Heap block header
 
+// Tag bits 0-6 help for statistics
+
 #define HEAP_TAG_FREE    0x00
 #define HEAP_TAG_USED    0x80
 #define HEAP_TAG_TYPE    0x0F
@@ -17,8 +19,12 @@
 #define HEAP_TAG_STRING  0x02
 #define HEAP_TAG_TTY     0x03
 
+// TODO: move free list node from header to body
+// to reduce overhead for allocated block
+
 struct heap {
-	list_s node;
+	list_s all;
+	list_s free;
 	word_t size;
 	byte_t tag;
 	byte_t unused;
@@ -26,12 +32,17 @@ struct heap {
 
 typedef struct heap heap_s;
 
+// Heap data
+
+extern list_s _heap_all;
+
 // Heap functions
 
 void * heap_alloc (word_t size, byte_t tag);
 void heap_free (void * data);
 
 void heap_add (void * data, word_t size);
+void heap_init ();
 
 #ifdef HEAP_DEBUG
 void heap_iterate (void (* cb) (heap_s * h));
--- a/elks/lib/heap.c
+++ b/elks/lib/heap.c
@@ -5,20 +5,22 @@
 #include <linuxmt/heap.h>
 //#include <linuxmt/lock.h>
 
-// Minimal block size to hold header and free list node
-#define HEAP_MIN_SIZE (sizeof (heap_s) + sizeof (list_s))
+// Minimal block size to hold heap header
+// plus enough space in body to be useful
+// (one paragraph for now but to be tuned
+// to size of smallest allocation)
+
+#define HEAP_MIN_SIZE (sizeof (heap_s) + 16)
 
 // Heap root
-// TODO: regroup in one structure
 
 // locks not needed unless SMP or reentrant kernel
 //static lock_t _heap_lock;
 #define WAIT_LOCK(lockp)
 #define EVENT_UNLOCK(lockp)
 
-heap_s * _heap_first;
-// TODO: free block list
-//static heap_s * _heap_free;
+list_s _heap_all;
+static list_s _heap_free;
 
 
 // Split block if enough large
@@ -32,9 +34,10 @@ static void heap_split (heap_s * h1, word_t size0)
 
 		heap_s * h2 = (heap_s *) ((byte_t *) (h1 + 1) + size0);
 		h2->size = size2 - sizeof (heap_s);
-		h2->tag = 0;  // free
+		h2->tag = HEAP_TAG_FREE;
 
-		list_insert_after (&(h1->node), &(h2->node));
+		list_insert_after (&(h1->all), &(h2->all));
+		list_insert_after (&(h1->free), &(h2->free));
 	}
 }
 
@@ -43,26 +46,24 @@ static void heap_split (heap_s * h1, word_t size0)
 
 static heap_s * free_get (word_t size0, byte_t tag)
 {
+	// First get the smallest suitable free block
+
 	heap_s * best_h  = 0;
 	word_t best_size = 0xFFFF;
+	list_s * n = _heap_free.next;
 
-	if (!_heap_first) return 0;
-	heap_s * h = _heap_first;
+	while (n != &_heap_free) {
 
-	// First get the smallest suitable free block
-	// TODO: improve speed with free list
-
-	while (1) {
+		heap_s * h = structof (n, heap_s, free);
 		word_t size1  = h->size;
 
-		if (!(h->tag & HEAP_TAG_USED) && (size1 >= size0) && (size1 < best_size)) {
+		if ((h->tag == HEAP_TAG_FREE) && (size1 >= size0) && (size1 < best_size)) {
 			best_h  = h;
 			best_size = size1;
 			if (size1 == size0) break;
 		}
 
-		h = structof (h->node.next, heap_s, node);
-		if (h == _heap_first) break;
+		n = h->free.next;
 	}
 
 	// Then allocate that free block
@@ -70,6 +71,7 @@ static heap_s * free_get (word_t size0, byte_t tag)
 	if (best_h) {
 		heap_split (best_h, size0);
 		best_h->tag = HEAP_TAG_USED | tag;
+		list_remove (&(best_h->free));
 	}
 
 	return best_h;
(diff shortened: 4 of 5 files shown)


ghaerr:
Hello @mfld-fr!

Does this change just execute faster, or does it increase any functionality? I haven't fully reviewed, but it seems you're keeping two lists now. Is the free list always maintained fully merged when possible?

The kernel now depends on the heap (and main memory segment) allocators much more so than six months ago. The allocation speed is not that important, but possible enhanced functionality (primarily to the segment allocator) would help with a couple enhancement requests (such as #709 and later #708, UMB support and the ability to initialize the allocator with multiple non-contiguous memory spaces). Any help would be appreciated.

Please test this on a current source tree (somewhere) before merge request. It looks there have been almost 700 commits since your fork's last update. A particular good test is that we are almost out of system resources when ELKS telnet networking is used to connect twice simultaneously to localhost, which stresses the system with four PTYs, 15 task slots, etc. All tty drivers (including PTY and serial) now use the near heap for allocating different-sized chqueues, and there are sometimes problems when the near heap is very full. Since the PTYs are allocated off an almost-full heap, depending on other issues, there may be a reason to discuss LIFO vs best-fit policy for the above reasons. LIFO (when same size) is likely a good policy for these PTY networking allocations to keep down checkerboarding.

`meminfo` now prints a full list of heap allocations to help debugging. Unfortunately it will need updating when the heap structures change. Meminfo has been an invaluable tool for debugging other kernel issues like those discussed above.

Screen Shot 2020-09-07 at 12 59 49 PM

mfld-fr:
Code change in `heap.c / heap.h` is fully tested in my fork, but not the 'in-kernel scan' performed by `meminfo.c`, where I just updated the mainstream code to pass the build.

Edit: tested code is here: https://github.com/mfld-fr/elks/commit/475396d5a544a37aa6bad0cfde41affcaca612eb

I will never merge back that 'scan' into my fork, because I strongly disagree such design where kernel internal structures are accessed from within the user space, and I won't spend time to test things I disagree with.

ghaerr:
> I will never merge back that 'scan' into my fork, because I strongly disagree such design where kernel internal structures are accessed from within the user space, and I won't spend time to test things I disagree with.

That's ok to disagree - but many, many kernel bugs have been fixed since your departure, and previously very little time was spent working on disagreeable things, so as a result the system as a whole remained quite buggy. `ps` uses the same method. I suggest you come up with a better design for debugging tools, especially if you won't test disagreeable things but still want to merge your items that break existing useful debugging tools!

mfld-fr:
I already proposed a design I think far more powerful for debugging : enable task switching on interrupt even in kernel space, so that we could exploit the 'break' and 'step by step' traps to suspend the current task and switch to a monitor task. 

The monitor task is demonstrated in https://github.com/mfld-fr/mon86, and the task switching in kernel space is demonstrated in https://github.com/mfld-fr/sys86.

You rejected the first step of improving the kernel for such task switching in #729.

ghaerr:
> I already proposed a design I think far more powerful for debugging : enable task switching on interrupt even in kernel space, so that we could exploit the 'break' and 'step by step' traps to suspend the current task and switch to a monitor task.

We are talking about four different things here.

Just previously, we were talking about changing kernel structures and the effect that has on utilities written to visually see internal structures while the system is in operation.

Now, you're talking about the ability to support single-instruction-step debugging, which is quite different.

And, you're saying that a full, normal task switch within kernel mode is required to support such a debugger.

And, you're talking about modifying a kernel that was explicitly designed (single CPU) to run system calls without being interrupted to be radically changed so as to support that one feature, without concern for the many possible correctness problems that will be caused, and the extreme number of kernel design and assumption changes required.

It is definitely not clear that the last item needs to be done in order to support a monitor, and I suspect the third item isn't either.


> The monitor task is demonstrated in https://github.com/mfld-fr/mon86, and the task switching in kernel space is demonstrated in https://github.com/mfld-fr/sys86.

I have been following the changes made in sys86. As you're coming to understand how that system must work, requiring an interrupt stack, where registers must get saved, etc, etc, hopefully you will come to see the nuances of the very-well-thought-out and reliable design that the ELKS kernel is built on. The task switching in sys86 hardly covers the problems involved with getting the ELKS kernel rewritten, let alone operational and bug-free, for a useful comparison.

Take the ELKS kernel, fork a new repository elks-reentrant, port the sys86 changes to it, and then run all the applications already written and now working, and see what happens. That's where it will become apparent the work required. For instance, the sys86 serial driver has the same problem the ELKS driver had for years, until high-speed serial I/O was fixed, where it became apparent how the the interrupt entry affected the kernel. These are fixed in ELKS.


> You rejected the first step of improving the kernel for such task switching in #729.

Yes and no. I mentioned a few of the large number of issues and problems, for even such a step. None were replied to.
I have spent hundreds of hours thinking about this, while getting to know the kernel deeply getting networking and other major system bugs solved. 

Yes - I am not interested in changing the very solid design and foundation of ELKS kernel within this repository; such a long-term project would require taking the system from its current state, and making a long series of slow changes, all of which will definitely effect system stability during the entire period. The mechanism currently in use in sys86 for saving user stacks won't work in ELKS and would require the rewrite of all system calls both on the libc and kernel side.

If your interest is just getting a decent low-level monitor built that works with the ELKS kernel, I think there are other ways to achieve this without the massive rewrite you seem intent on. I am happy to give help, but I don't want to fight over it, as it seems our priorities are quite different.

mfld-fr:
Yes, definitively agree on the priorities disagreement. If the project followed your current priorities the years before your arrival, ELKS would have stayed on DEV86 / BCC and never switched to GCC-IA16... I consider ELKS for 'educational & fun', so allowing big changes & related unstability, you looks to consider it as a 'product'... so beyond priorities, we even disagree on the project goals.

Indeed, your current 'far procedure' experiment looks fun & interesting to me... at least you are attacking big things that could change the game again...

Back to the impact of the proposed heap improvement on the utilities : as far as I see, only `mem.c` and `meminfo.c` are impacted by modifying the `heap` structure, and only minor updates were required in a few lines. I already performed these updates in this PR, but I let you run this... horrible... `meminfo`.

I do prefer the kernel to dump the heap itself, directly to the system console, or better in a short form to a virtual file, that a user land utility can expand to a readable form (= a kind of 'procfs').

ghaerr:
> as far as I see, only mem.c and meminfo.c are impacted by modifying the heap structure, and only minor updates were required in a few lines.

Thank you, that helps a lot.

> still a glitch in the free list order

What is the problem in the free list order?

Can you give a brief explanation of what the benefit is of the extra free list? Is it for quicker allocations, as both the free and allocated blocks are stored in the 'all' list? Are all modifications to the free list duplicated in the all list?

Thank you!

mfld-fr:
> Can you give a brief explanation of what the benefit is of the extra free list? Is it for quicker allocations, as both the free and allocated blocks are stored in the 'all' list?

Today, the allocator spends more and more time to traverse the 'all blocks' list to find a free block. Introducing the 'free blocks' list makes the allocator to skip the already allocated blocks, resulting in a dramatic speed improvement after allocating only a few blocks.

Suppose you already allocated 9 blocks. Previous version iterates on that 9 blocks, then returns the 10th. Current version directly jumps to the 10th.

> Are all modifications to the free list duplicated in the all list?

The 'all' list is unchanged, and still tracks all the blocks in memory order, so the minimal impact on the code that traverses that list. The 'free' list only tracks the free blocks, and is updated at the cost a very few more lines of code.

> What is the problem in the free list order?

The glitch is having a free list not in memory order after some heap fragmentation. As far as I see while testing the new allocator, there is actually no problem. I even start to think that pushing the non-merged freed blocks to the head of the free list could contribute to reduce that heap fragmentation in addition to the 'best fit' algorithm, as it tries to reallocate that blocks first.

ghaerr:
Thanks for the detailed explanation.

> The glitch is having a free list not in memory order after some heap fragmentation. 

Line 135 inserts the free block at the head of the free list, so the free list isn't kept in memory order, correct?
But the allocator will always use best fit, rather than LIFO for allocations, correct?

I may have been incorrect in saying that it when the heap is near full using LIFO would likely be better. Now I'm realizing to avoid checkerboarding caused by frequent variable sized PTY queue pairs (for instance), it is probably better to ensure that a previous small (PTY) reallocation doesn't take a large free block first, thus leaving no space for the second larger PTY allocation.

> The 'all' list is unchanged, and still tracks all the blocks in memory order

Does the free list have to be in memory order?  When using best-fit strategy, it would seem that all free blocks would have to be looked at.

>  I even start to think that pushing the non-merged freed blocks to the head of the free list could contribute to reduce that heap fragmentation in addition to the 'best fit' algorithm, as it tries to reallocate that blocks first.

Two points: not sure not merging free blocks is a good idea. And still trying to understand whether unsorted free list matters or not, when using best-fit.

mfld-fr:
The 'free' list does not have to be in memory order, and yes, all the free blocks are presented to the 'best fit' algorithm anyway (_except in the 'exact fit' - see below_). Merging the free blocks is a must have to limit the heap fragmentation.

But the order of presenting has a performance impact when repeatedly allocating / deallocating blocks of the same size, so when having a chance to hit the 'exact fit' case that shorten the 'best fit' loop.

Maybe the outcome of this review is that the 'free' list should be merely LIFO for the best performance, and not trying to keep memory order as in the 'all' list ?

ghaerr:
> Maybe the outcome of this review is that the 'free' list should be merely LIFO for the best performance, and not trying to keep memory order as in the 'all' list ?

No - I think of several things: 

First, I don't actually know whether LIFO versus best-fit will be better for checkerboarding. This will have to be tested running the networking or other scenarios, which I can do, after commit. 

Secondly, I don't frankly think that performance is that big a deal, given the small number of allocations traversed in ELKS; but, in the interest of experimentation and a cost of only two extra bytes per allocation, the added free list offers the opportunity to test allocation techniques and also possibly offer LIFO, which the current implementation cannot.

Third, I bet that LIFO and not using best-fit will hurt ELKS when tight on memory; perhaps best-lifo-fit (which I think is what this code is now doing?) would be better. But all this needs a working system with lots of processes and action for both of us to find out.

So, the best order for the free list IMO has yet to be determined. But I like the having the free list separate from the main chain, as it allows experimentation within the working system (not for performance reasons, but for heap fragmentation control).

With regards to my thoughts on ELKS performance issues in a memory allocator: after tracing down disk read requests, serial port reads and other, ELKS spends far too much time memcpy'ing data, in almost every level of the kernel layers. This time usage (e.g. copy 1024 bytes) far outweighs any benefit of a faster allocator that has perhaps 35 links in a loaded system. That said, I can understand why you would like to build and provide the fastest allocator possible :)

mfld-fr:
Pushed a latest version that tries to benefit from both 'best fit' to minimize heap fragmentation and from 'exact fit' / LIFO for speed (see comments in code). Also increased the minimal block size to avoid any empty splitted block.

ghaerr:
Thanks, it seems this does solve the issues discussed in this thread regarding exact fit and LIFO. Are there other features you are considering adding at this time?

It seems that, in lines 128-129, I found, while determining whether fp would ever be used uninitialized or not, that it would make more sense to discard 'mn', and initialize fp to NULL. Then, just test fp non-NULL later instead of using two variables.

Another small issue is: is declaring variables mid-function a GCC extension, or C standard? If not C standard, all variables ought to be declared at start of function, like they are everywhere else in the kernel.

mfld-fr:
C89 / ANSI allows to declare variables at the beginning of a block, and C99 allows to declare local variables everywhere before they are used. Using a C99 compiler was one driver to move from BCC to GCC-IA6 (see issue #166).

Reworked `heap_free` to push ALL the merged blocks to the free list tail (as no more chance for 'exact hit' after a merge), and the still alone blocks to the free list head (as more chance).

I think the heap optimization by adding a free list is now complete, and could be merged now.

ghaerr:
Hello @mfld-fr,

Thanks for your heap allocator improvements.

After doing some testing, I'm not sure what benefit the extra space in HEAP_MIN_SIZE is good for. Currently set at 16, this uses 120 extra heap bytes in a normal system startup. (Shown is `console=ttyS0 net=eth 3` /bootopts which sets serial console, starts networking and init level 3). 

Setting the extra value to 12 (size of mm_alloc segment from near heap), or to 0, is much better, and either gives the exact same result. Suggest changing to 0 or just eliminating it.

Extra set to 16 (note large allocations for some segment alloc/frees and lower heap largest free space):


Extra set to 0 or 12 (optimal in either case, but could become non-optimal on possible smaller minimal heap allocation than 12):


What do you think?

Thank you!

mfld-fr:
Yes, as told in the code comment, minimum body size should be set to the size of the smallest allocated object, currently the segment descriptor, so 12. It should not be zero, otherwise some useless headers would be inserted that would only slow down the 'best fit' loop.

mfld-fr:
Minimum size adjusted in #748

PR #747, opened 2020-09-13
[kernel] [tools] [elkscmd] [boot] Work with major gcc-ia16 toolchain update

tkchia:
I am kind of rushing out this PR because I hope to move ELKS away from the default `ld.bfd` linker in `binutils-ia16`, which is rather hard to make changes to, and seems to be acting up of late (https://github.com/jbruchon/elks/pull/743#event-3748875074).

On the compiler toolchain end, I had made these changes:

  * The linker used for ELKS is now [`ld.gold`](https://opensource.googleblog.com/2008/04/gold-google-releases-new-and-improved.html) rather than `ld.bfd`.
  * The compiler now internally uses a different method to represent _segment_`:`_offset_ addresses inside ELF `.o files.
    * See [H. P. Anvin's `segelf` proposal](https://git.zytor.com/users/hpa/segelf/abi.git/plain/segelf.txt) and [my write-up](https://github.com/tkchia/build-ia16/blob/master/elf16-writeup.md) for more information.
    * Basically, rather than modelling the IA-16 segments for text, far text, and data as ELF sections with overlapping virtual linear addresses --- but distinct load addresses --- the proposed method models them as having separate virtual addresses in the same linear address space, and uses special symbols to tell the segments apart.

(I have not yet rebased my patches on top of a newer version of mainline Binutils; I hope to do so in the longer term.)

On the ELKS end, the following changes were necessary:

  * The linker scripts were modified to work with the `segelf` scheme.
  * `ld.gold` can only output ELF executable files --- unlike `ld.bfd`.  To correctly build the bootloader and kernel, I added a separate post-link pass to convert ELF "executables" into ELKS `a.out`'s.  I have added a program `elf2elks` into the ELKS tree to do this.  (This means that we longer need the `i386elks` back-end I had added to Binutils.)
  * All program entry points were renamed to `_start`.  (`ld.gold` for some reason does not honour `PROVIDE (_start =` ...`);` in linker scripts.)
  * All `.o` files had to be rebuilt.

This PR also updates the size of the `a.out` header's `esh_ftseg` field to 32 bits, as a follow-up to https://github.com/jbruchon/elks/pull/746 .

Thank you!


tkchia:
I forgot to mention that `elf2elks` currently requires the `libelf` library (from [`elfutils`](https://sourceware.org/elfutils/)) to be installed.  I have not added the `libelf` source code into the ELKS tree, but perhaps I should (?).

Thank you!

ghaerr:
Hello @tkchia,

Wow, lots of changes! A quick review of all of them looks pretty good, I still have a few questions and comments:

- What about user programs? Will they need separate $(POSTLINK) processing if not built using a separate current $(LD) ... link line from source file compilation? There may be some one-step compilations performed, I haven't fully checked the tree yet.
- It looks like the compiler no longer knows about -maout-stack or -maout-heap options, are these being renamed --stack and --heap in elf2elks? If so, then this commit will break user application builds unless those options changed throughout elkscmd/. 
- I'm wondering whether it might (or might not) be a good idea to have elftoelks remain under cross/ and have ia16-unknown-gcc frontend automatically run it instead of separate POSTLINK requirement, as well as converting or passing -maout-\* options, etc. ?
- Since packages aren't necessarily easily available on macOS (which I build on), I suggest we move libelf into source tree, how about at tools/elftoelks/lib which keeps it self-contained, unless elftoelks moved to cross/.
- I don't fully understand the magic of the updated linker scripts (yet) - will they still be able to correctly process, say, far text sections named ".fartext.init" for upcoming initialization sections (with nothing placed after it), or should they be named something different? This option is currently kept in include/linuxmt/config.h. The idea is the .fartext.init section would be sorted last in all the far text code.
- Have you tested the setup.S/minix.h modifications when producing a kernel with a far text segment? Or should I?
- Line 102 of elksemu/elks.h still needs uint32_t for esh_ftseg.
- Do you have more hard information about previous compiler emitting bad code or relocation information for far text programs and whether this fixes it? I had mentioned this in #743 and still have test patch for trying to get kernel booted with .fartext.init sections.
- Finally, have you tested this on entire ELKS system, running various application programs? Or will we need more testing for those programs that use -maout-stack= options as described above? I may have missed something in elftoelks.

Thank you!

tkchia:
Hello @ghaerr,

> What about user programs? Will they need separate $(POSTLINK) processing if not built using a separate current $(LD) ... link line from source file compilation? There may be some one-step compilations performed, I haven't fully checked the tree yet.

If you link programs using the `ia16-elf-gcc` front end, the front end will automatically run the post-link step.  This should be the case for most userland programs.  Even `ia16-elf-gcc -nostdlib` ... will work.

You only need a separate `$(POSTLINK)` step if you invoke the linker (`ia16-elf-ld.gold`) directly, as is done for the bootloader and kernel.

And yes, I have rebuilt the entire ELKS tree and (briefly) tested the resulting disk image. :-)

> It looks like the compiler no longer knows about -maout-stack or -maout-heap options, are these being renamed --stack and --heap in elf2elks?

These should still be supported by `ia16-elf-gcc` --- it should transform these options into `--stack` and `--heap` options when invoking `elf2elks`.

> Since packages aren't necessarily easily available on macOS (which I build on), I suggest we move libelf into source tree, how about at tools/elftoelks/lib which keeps it self-contained, unless elftoelks moved to cross/.

Yes, OK.  In fact I just found that Sourceware's `libelf` is heavily dependent on GNU `libc`, which is problematic as I am also trying to get `libelf` to work on DJGPP.

Apparently FreeBSD has [an alternate (?) implementation](https://wiki.freebsd.org/LibElf) of the `libelf` interface which might be more suited for other OSes, including macOS.  I will see if I can get that to work.

> * ... will they still be able to correctly process, say, far text sections named ".fartext.init" for upcoming initialization sections (with nothing placed after it), or should they be named something different?

That should work.  The additional linker script nonsense is mainly to make sure that the special sections (and symbols) marking the start of each IA-16 segment will end up in the right places.

> * Do you have more hard information about previous compiler emitting bad code or relocation information for far text programs and whether this fixes it?

Unfortunately no, except that the old linker would sometimes crash, or emit relocations at the wrong place, or other randomness.

If you need a quick way to see what is going on inside an `a.out`, for now, running `ia16-elf-objdump -D -r -Mi8086` on the program will still work.  (This is still a useful functionality from my mostly-unneeded `i386elks` back-end, I guess...)

Thank you!

tkchia:
Hello @ghaerr,

> * Have you tested the setup.S/minix.h modifications when producing a kernel with a far text segment? Or should I?

I have not really tested these yet.  If you would like to help test them, please do.  Thanks!

tkchia:
Hello @ghaerr,

> * Since packages aren't necessarily easily available on macOS (which I build on), I suggest we move libelf into source tree, how about at tools/elftoelks/lib which keeps it self-contained, unless elftoelks moved to cross/.

I have added the BSD implementation of `libelf` under the `elf2elks` subtree, where it can be built.  I managed to avoid roping in BSD `make` as a dependency --- `libelf`'s makefile uses BSD `make` syntax --- though the `m4` program is still needed to generate some source files.

Thank you!

ghaerr:
Hello @tkchia,

Thank you for answering all the questions I had about this PR, I am fine with your plan and these changes.

I would like to try confirming the libelf make before committing this, so as not to be potentially down should there be issues. I am traveling but should be able to get that done within the next 24 hours.

I am also looking forward to testing the new linker with the far text kernel after commit, and will report back results.

Also, elks.h still needs a uint32_t type for esh_ftseg.

Thank you!

tkchia:
Hello @ghaerr,

> I am traveling but should be able to get that done within the next 24 hours.

Thanks very much!

> Also, elks.h still needs a uint32_t type for esh_ftseg.

It is correct (kind of).  I did move the `esh_ftrsize` relocation size field two bytes down.

Thank you!

ghaerr:
Hello @tkchia,

I've fetched PR 747 (without commit) and started rebuilding the compiler. It seems to be failing in the 'gold' directory:
```
g++ -DHAVE_CONFIG_H -I. -I../../binutils-src/gold  -I../../binutils-src/gold -I../../binutils-src/gold/../include -I../../binutils-src/gold/../elfcpp -DLOCALEDIR="\"/Users/greg/net/elks/cross/share/locale\"" -DBINDIR="\"/Users/greg/net/elks/cross/bin\"" -DTOOLBINDIR="\"/Users/greg/net/elks/cross/ia16-elf/bin\"" -DTOOLLIBDIR="\"/Users/greg/net/elks/cross/ia16-elf/lib\""   -W -Wall    -Werror -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -frandom-seed=gold-threads.o -I../../binutils-src/gold/../zlib -Qunused-arguments -pthread -g -O2 -MT gold-threads.o -MD -MP -MF .deps/gold-threads.Tpo -c -o gold-threads.o ../../binutils-src/gold/gold-threads.cc
../../binutils-src/gold/gold-threads.cc:288:13: error: expected expression
    : once_(PTHREAD_ONCE_INIT)
            ^
/usr/include/pthread.h:210:27: note: expanded from macro 'PTHREAD_ONCE_INIT'
#define PTHREAD_ONCE_INIT {_PTHREAD_ONCE_SIG_init, {0}}
                          ^
1 error generated.
make[5]: *** [gold-threads.o] Error 1
make[4]: *** [all-recursive] Error 1
make[3]: *** [all] Error 2
make[2]: *** [all-gold] Error 2
make[1]: *** [all] Error 2
make: *** [/Users/greg/net/elks/cross/build/.binutils.build] Error 2
```
I'm on macOS Sierra 10.12.6.

Looking further into the failure, it's dying in this section of gold-threads.cc, which is \#ifdef'd ENABLE_THREADS:
```
#ifdef ENABLE_THREADS

// Class Once_initialize.  This exists to hold a pthread_once_t
// structure for Once.

class Once_initialize
{
 public:
  Once_initialize()
    : once_(PTHREAD_ONCE_INIT)
  { }

  // Return a pointer to the pthread_once_t variable.
  pthread_once_t*
  once_control()
  { return &this->once_; }

 private:
  pthread_once_t once_;
};

#endif // defined(ENABLE_THREADS)
```
Shall I configure without threads? If you could hint where I might do that or with what ./configure option, I will keep trying.

Thank you!

tkchia:
Hello @ghaerr,

Thanks for your report!  That was ... quite unexpected.

I think building with threads is still OK. :-)  The issue was that `<pthread.h>`'s `pthread_once_t` is a structure type on macOS --- and so `PTHREAD_ONCE_INIT` (= `{`...`, {0}}`) is a structure initializer list, but unfortunately you cannot use `{`...`, {0}}` as a function argument expression.

(On Linux/x86-64, and presumably other platforms where `gold` was tested, `pthread_once_t` is a scalar type, so this issue did not arise.)

You should be able to fix this for the time being, by rewriting the constructor as
```
  Once_initialize()
    : once_((pthread_once_t) PTHREAD_ONCE_INIT)
  { }
```
which should yield a valid "compound literal" expression on macOS.

On my end I will try to put out an updated Binutils, and also see how I can speed up the downloading and building of the toolchain in the ELKS tree.

Thank you!

ghaerr:
Hello @tkchia,

Thank you for the pthread_once_t init fix, that allowed the tools compilation to complete. I haven't yet tried the revised sped-up download from your very last push, but instead used the new toolchain to attempt an ELKS tree compile. The kernel was built, but now the make is failing while trying to link elkscmd/ash/ash:
```
ia16-elf-gcc -fno-inline -melks-libc -mcmodel=small -mno-segment-relocation-stuff -mtune=i8086 -Wall -Os -maout-heap=6144 -maout-stack=2048 -o ash builtins.o cd.o dirent.o error.o eval.o exec.o expand.o input.o jobs.o mail.o main.o memalloc.o miscbltin.o mystring.o nodes.o options.o parser.o redir.o show.o signames.o syntax.o trap.o output.o var.o init.o bltin/echo.o bltin/expr.o bltin/regexp.o bltin/operators.o linenoise_elks.o autocomplete.o 
/Users/greg/net/elks/cross/bin/../lib/gcc/ia16-elf/6.3.0/../../../../ia16-elf/lib/elkslibc/crt0.o:(.text+0x11): relocation truncated to fit: R_386_16 against `environ'
/Users/greg/net/elks/cross/bin/../lib/gcc/ia16-elf/6.3.0/../../../../ia16-elf/lib/elkslibc/crt0.o: in function `exit':
(.text+0x26): relocation truncated to fit: R_386_16 against `_cleanup'
cd.o: in function `getcomponent':
(.text+0x1): relocation truncated to fit: R_386_16 against `.bss'
(.text+0x1c): relocation truncated to fit: R_386_16 against `.bss'
(.text+0x28): relocation truncated to fit: R_386_16 against `.bss'
cd.o: in function `updatepwd':
(.text+0x44): relocation truncated to fit: R_386_16 against `.bss'
(.text+0x4d): relocation truncated to fit: R_386_16 against `stacknxt'
(.text+0x50): relocation truncated to fit: R_386_16 against `stacknleft'
(.text+0x53): relocation truncated to fit: R_386_16 against `sstrnleft'
(.text+0x5f): relocation truncated to fit: R_386_16 against `curdir'
(.text+0x72): additional relocation overflows omitted from the output
/Users/greg/net/elks/cross/bin/../lib/gcc/ia16-elf/6.3.0/../../../../ia16-elf/bin/ld: ash: can not represent section `.comment' in ELKS object file format
/Users/greg/net/elks/cross/bin/../lib/gcc/ia16-elf/6.3.0/../../../../ia16-elf/bin/ld: final link failed: nonrepresentable section on output
collect2: error: ld returned 1 exit status
make[2]: *** [ash] Error 1
make[1]: *** [all] Error 1
make: *** [all] Error 2
```

ghaerr:
@tkchia - please temporarily ignore my last comment... I am using a new clone of the ELKS tree and it seems my PATH is possibly corrupted, even after running `. env.sh`:
```
PATH=/Users/greg/net/elks/elks/tools/bin:/Users/greg/net/elks/cross/bin:/Users/greg/net/elks-gh/elks/tools/bin:/Users/greg/net/elks-gh/cross/bin:/opt/local/bin:/opt/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin
```
The "elks-gh" is the OLD tree, "elks" is the NEW tree. Let me look at this for a minute...

tkchia:
Hello @ghaerr,

Yes, it seems the messages are coming from the old default linker (`ld.bfd`).  With my updates, all ELKS code should now be built with the `gold` linker, which reports relocation overflows (when they happen) in a different way.

Thank you!

ghaerr:
Hello @tkchia,

OK, the ELKS tree now builds and runs. However, just for the heck of it, I thought to try networking, and with the new build, it is failing when running `telnet localhost` with a large errno, running as follows:
```
ELKS 0.3.0
login: root
# net start
...
# telnet localhost
Connecting to localhost (127.0.0.1:23)
Connected
write: Unknown error 26174
write: Unknown error 26174
write: Unknown error 26174
write: Unknown error 26174
write: Unknown error 26174
login: Connection closed: Unknown error 26174
```
The above works fine on the old compiler/linker. This is sometimes indicative of a stack overflow. I then looked further, checking to see whether perhaps --stack/--heap was perhaps not working and ran into a problem from my previous #746, where the a.out header changed and chmem.c wasn't updated - so it's buggy. I am going to fix that on both trees and continue looking into why networking isn't working with the new toolchain.


Thank you!

ghaerr:
Hello @tkchia,

I have tracked down the problem, it has to do with the nuances of the v0 and v1 header, and the elf2elks command setting a default v0 header when stack and heap are 0. Take a look at the following screenshots of telnet's data allocation:

OLD telnet data segment: (CORRECT)
old

NEW telnet data segment:
new

The chmem program displays the effective heap with the following code:
```
        if (header.chmem == 0) {                                /* default heap*/
                oldheap = INIT_HEAP;
                displayheap = 0;
        } else if (header.version == 1)
                displayheap = oldheap = header.chmem;
        else
                displayheap = oldheap = header.chmem - dsegsize;        /* v0 displays effective heap, not header value*/
```

I haven't yet determined the exact problem, but it seems that elf2elks is always setting a v0 header unless --stack/--heap are passed. I think it always needs to set a v1 header.

I am testing this with the fixed chmem in #751, and then executing "init 3" after login, then "chmem /bin/\*".

tkchia:
Hello @ghaerr,

> ... but it seems that elf2elks is always setting a v0 header unless --stack/--heap are passed. I think it always needs to set a v1 header.

Whoops --- thanks for highlighting this.  I have fixed this particular problem (and also Git-rebased my changes).

With the fix, `elf2elks` currently still defaults to using a v0 `a.out` header for the tiny model though.  This is because (at least for now) `elks/arch/i86/tools/build.c` does not like v1 headers on the bootloader and `setup.S` code.  But this can probably be changed later.

Thank you!

ghaerr:
Hello @tkchia,

Thanks for your changes. I have verified that the data segment is now allocated properly with these changes. However, `telnet localhost` is still failing (with identical source tree prior to this PR). Notice below that the new version telnet happens to have 48 bytes less text size (8016 bytes) than the working version (8064 bytes, see first screenshot in my post above). I find that a strange number, considering that the code-generating portions should not have changed, right? Note, we are talking about load text size, not relocation info.

Do you have any ideas as to why this might be? I suspect it is related to it not running properly with this PR.

New (not working) telnet:
Screen Shot 2020-09-17 at 9 38 53 AM

Thank you!

tkchia:
Hello @ghaerr ,

This is strange.  Do you still have a copy of the old binary (before this PR)?  I am thinking it might be useful to try to do something like a byte-by-byte comparison to figure out what is up.

Thank you!

ghaerr:
Yes, I am running dual trees right now. Here's a zipped version of the entire elkscmd/inet/telnet directory, which includes the telnet binary as well as .o files for easy inspection.

[telnet.zip](https://github.com/jbruchon/elks/files/5240742/telnet.zip)

ghaerr:
I used objdump to display ttn.o on both trees. There is no difference in code generated, only relocation info.
There is no difference in tt_conf.o.

Here is the diff for ttn.o:
```
Gregs-Home-iMac:telnet greg$ diff /tmp/file /tmp/file2
22a23
> 			19: R_386_SUB16	.bss!
53a55
> 			4a: R_386_SUB16	.bss!
60a63
> 			57: R_386_SUB16	.rodata.str1.1!
101a105
> 			97: R_386_SUB16	.rodata.str1.1!
104a109
> 			9b: R_386_SUB16	.rodata.str1.1!
130a136
> 			d0: R_386_SUB16	DO_echo!
133a140
> 			d7: R_386_SUB16	DO_echo_allowed!
152a160
> 			101: R_386_SUB16	DO_echo!
161a170
> 			118: R_386_SUB16	.bss!
167a177
> 			125: R_386_SUB16	.rodata.str1.1!
178a189
> 			136: R_386_SUB16	DO_suppress_go_ahead!
181a193
> 			13d: R_386_SUB16	DO_suppress_go_ahead_allowed!
184a197
> 			144: R_386_SUB16	DO_suppress_go_ahead!
211a225
> 			184: R_386_SUB16	.bss!
217a232
> 			192: R_386_SUB16	.bss!
225a241
> 			1a2: R_386_SUB16	.bss!
226a243
> 			1a4: R_386_SUB16	.rodata.str1.1!
230a248
> 			1ac: R_386_SUB16	.bss!
232a251
> 			1b0: R_386_SUB16	.bss!
243a263
> 			1cb: R_386_SUB16	.bss!
258a279
> 			1e7: R_386_SUB16	.bss!
265a287
> 			1f5: R_386_SUB16	.rodata.str1.1!
275a298
> 			206: R_386_SUB16	stderr!
278a302
> 			20a: R_386_SUB16	.rodata.str1.1!
337a362
> 			28a: R_386_SUB16	WILL_terminal_type!
340a366
> 			291: R_386_SUB16	WILL_terminal_type_allowed!
346a373
> 			2a2: R_386_SUB16	WILL_terminal_type!
348a376
> 			2a7: R_386_SUB16	.rodata.str1.1!
353a382
> 			2ae: R_386_SUB16	.bss!
358a388
> 			2b9: R_386_SUB16	.bss!
359a390
> 			2bb: R_386_SUB16	.rodata.str1.1!
367a399
> 			2d0: R_386_SUB16	.bss!
373a406
> 			2dd: R_386_SUB16	.rodata.str1.1!
395a429
> 			301: R_386_SUB16	.rodata.str1.1!
398a433
> 			305: R_386_SUB16	.rodata.str1.1!
408a444
> 			318: R_386_SUB16	.bss!
415a452
> 			326: R_386_SUB16	.rodata.str1.1!
445a483
> 			36a: R_386_SUB16	.rodata!
449a488
> 			371: R_386_SUB16	.rodata.str1.1!
452a492
> 			375: R_386_SUB16	.rodata.str1.1!
462a503
> 			389: R_386_SUB16	stderr!
465a507
> 			38d: R_386_SUB16	.rodata.str1.1!
473a516
> 			39d: R_386_SUB16	stderr!
476a520
> 			3a1: R_386_SUB16	.rodata.str1.1!
479a524
> 			3a6: R_386_SUB16	stderr!
482a528
> 			3aa: R_386_SUB16	.rodata.str1.1!
485a532
> 			3af: R_386_SUB16	stderr!
488a536
> 			3b3: R_386_SUB16	.rodata.str1.1!
491a540
> 			3b8: R_386_SUB16	stderr!
494a544
> 			3bc: R_386_SUB16	.rodata.str1.1!
497a548
> 			3c1: R_386_SUB16	stderr!
500a552
> 			3c5: R_386_SUB16	.rodata.str1.1!
503a556
> 			3ca: R_386_SUB16	stderr!
506a560
> 			3ce: R_386_SUB16	.rodata.str1.1!
509a564
> 			3d3: R_386_SUB16	stderr!
512a568
> 			3d7: R_386_SUB16	.rodata.str1.1!
522a579
> 			3ef: R_386_SUB16	stderr!
526a584
> 			3f6: R_386_SUB16	.rodata.str1.1!
566a625
> 			451: R_386_SUB16	.rodata.str1.1!
640a700
> 			4fc: R_386_SUB16	.rodata.str1.1!
643a704
> 			500: R_386_SUB16	stderr!
661a723
> 			520: R_386_SUB16	.bss!
668a731
> 			52e: R_386_SUB16	.rodata.str1.1!
679a743
> 			540: R_386_SUB16	.rodata.str1.1!
685a750
> 			54a: R_386_SUB16	.rodata.str1.1!
738a804
> 			5b3: R_386_SUB16	.rodata.str1.1!
741a808
> 			5b7: R_386_SUB16	.rodata.str1.1!
991a1059
> 			0: R_386_SUB16	.text!
993a1062
> 			2: R_386_SUB16	.text!
994a1064
> 			4: R_386_SUB16	.text!
996a1067
> 			6: R_386_SUB16	.text!
997a1069
> 			8: R_386_SUB16	.text!
999a1072
> 			a: R_386_SUB16	.text!
1000a1074
> 			c: R_386_SUB16	.text!
1002a1077
> 			e: R_386_SUB16	.text!
1004a1080
> 			10: R_386_SUB16	.text!
1006a1083
> 			12: R_386_SUB16	.text!
1008a1086
> 			14: R_386_SUB16	.text!
1009a1088
> 			16: R_386_SUB16	.text!
1011a1091
> 			18: R_386_SUB16	.text!
1013a1094
> 			1a: R_386_SUB16	.text!
1015a1097
> 			1c: R_386_SUB16	.text!
1033a1116
> 			19: R_386_SUB16	.rodata.str1.1!
1036a1120
> 			1d: R_386_SUB16	stderr!
1045a1130
> 			2b: R_386_SUB16	.bss!
1052a1138
> 			35: R_386_SUB16	finish!
1065a1152
> 			49: R_386_SUB16	.bss!
1084a1172
> 			74: R_386_SUB16	.rodata.str1.1!
1122a1211
> 			ca: R_386_SUB16	.rodata.str1.1!
1130a1220
> 			d9: R_386_SUB16	.bss!
1137a1228
> 			e6: R_386_SUB16	.rodata.str1.1!
1142a1234
> 			f2: R_386_SUB16	.rodata.str1.1!
1156a1249
> 			10e: R_386_SUB16	.bss!
1180a1274
> 			139: R_386_SUB16	.rodata.str1.1!
1187a1282
> 			144: R_386_SUB16	.bss!
Gregs-Home-iMac:telnet greg$ 
```

tkchia:
Hello @ghaerr,

`telnet.zip` seems to be an empty `.zip` file.  Perhaps you can repack and re-upload it?

I am kind of occupied at the moment --- I will take a further look once I have more time.

Sorry and thank you!

ghaerr:
Hello @tkchia,

Here's a working zip.

No problem on the timeframe; I would like to solve this problem before committing though. I'll look into this as well.

[telnet.zip](https://github.com/jbruchon/elks/files/5240806/telnet.zip)

ghaerr:
@tkchia,

More news: I thought to try running the older-compiled telnet in the new kernel, and it fails the same way. Very strange. The failure is an incorrect, large errno return I previously fixed when debugging kernel networking a couple months ago. The actual reason for it had to do with a shared kernel buffer getting overwritten, which was solved with a global semaphore. I can't think of how that would be related to this, especially since in all other respects, both telnet and the kernel seem to be operating normally. (Except still unknown reason why text size differs between telnet).

Between my two build trees, otherwise seemingly identical source except for this PR, I find the following differences in kernel sizes:

Old kernel build
```
tools/build boot/bootsect boot/setup boot/system > boot/Image
Root device is (3, 128)
Boot sector 512 bytes.
Setup is 1505 bytes.
System is 114192 B (59520 B code, 0 B fartext, 7952 B data and 46720 B bss)
System is 67504
```

New kernel build
```
tools/build boot/bootsect boot/setup boot/system > boot/Image
Root device is (3, 128)
Boot sector 512 bytes.
Setup is 1505 bytes.
System is 113552 B (58992 B code, 0 B fartext, 7840 B data and 46720 B bss)
System is 66864
```

The new system is quite a bit smaller in text and data, but yet no changes made to the kernel in this PR (except setup). The above text/code/bss display is final size, not including relocation info. There should be no relocation info in the kernel or any program unless a far text segment is used, correct?

The system sizes differ by 640 in both cases, which exactly equals the text (528) and data (112) size differences between the kernels.

Hopefully this helps.

Thank you!

tkchia:
Hello @ghaerr,

> The system sizes differ by 640 in both cases, which exactly equals the text (528) and data (112) size differences between the kernels.

Ah yes --- I just did a comparison of your older `telnetd` and my locally compiled new `telnetd`, then looked at the old commit number for `gcc-ia16` (= https://github.com/tkchia/gcc-ia16/commit/d3440aa96f9768b7268174278f9b80c1a85410bd) and the newly bumped commit number (= https://github.com/tkchia/gcc-ia16/commit/45605f97ac4d313f8783e1ba9924d0aec0b1d6e8).

And I now recall that I _did_ add at least one minor optimization to `gcc-ia16` between the two commits, namely in https://github.com/tkchia/gcc-ia16/commit/819954fe595c392d0144ee35a5c5c66905a21f8c .  Basically it tries to rewrite, say,
```
        add $6, %sp
        mov %bp, %sp
```
into simply
```
        mov %bp, %sp
```
or in some cases
```
        add $6, %sp
```

I will be quite surprised though if this change is causing code correctness problems, since I did run this change through the GCC's C regression tests.

Thank you!

ghaerr:
Hello @tkchia,

> comparison of your older telnetd and my locally compiled new telnetd

I assume you mean `telnet`, which is the program having problems, right?

> And I now recall that I did add at least one minor optimization to gcc-ia16 between the two commits

Ok, this may account for the code size difference in telnet. It doesn't answer why the kernel data size should be different, however.

> I will be quite surprised though if this change is causing code correctness problems, since I did run this change through the GCC's C regression tests.

Well, the most complex test I could think of to test this PR was to run `telnet localhost`, which heavily exercises the system, and the system doesn't work with these changes for a reason unknown to us. I can't guess whether the issue might be with the new linker, or the compiler changes. I could try building a third source tree and use your updated compiler commit, but without the new linker, if you think that would allow us to bifurcate this problem? Let me know what change I would need to make (prior to this PR) to tools/Makefile for that test.

If testing the updated compiler but using the old linker is easier on your end, the system failure is easily seen by running the following commands after an ELKS boot:
```
net start
telnet localhost
```

Thank you!

tkchia:
Hello @ghaerr,

Perhaps you can try to temporary "revert" the https://github.com/tkchia/gcc-ia16/commit/819954fe595c392d0144ee35a5c5c66905a21f8c and https://github.com/tkchia/gcc-ia16/commit/f1ca2046c11232e4a7ce82b28c111fda99e0396a commits under the new `gcc-ia16` tree, to see if they are indeed causing the problem?

This patch should be able to do the reversion: [`20200918.unopt.diff.txt`](https://github.com/jbruchon/elks/files/5246957/20200918.unopt.diff.txt).

If this `gcc-ia16` optimization is indeed causing issues then I will probably disable it first.

(I am having trouble getting `net start` to work in my Qemu emulation, so I cannot yet reproduce your issue.  My apologies.)

Thank you!

tkchia:
Hello @ghaerr,

> I could try building a third source tree and use your updated compiler commit, but without the new linker, if you think that would allow us to bifurcate this problem?

One relatively straightforward way to do _that_, would be to bump up the `gcc-ia16` compiler in the _old_ tree to just before the `ld.gold` and `segelf` changes (but including the `add $6, %sp; movw %bp, %sp` optimization I mentioned).  That commit is https://github.com/tkchia/gcc-ia16/commit/606351cf50a27db562faa0768fbe760b89bbe5dc .

Thank you!

tkchia:
Hello @ghaerr,

Incidentally, may I know what options you are passing to QEMU to enable networking support?  Thanks!

ghaerr:
Hello @tkchia,

Thanks for the patch. I'll apply that to the cross/ directory in my current test tree with the new linker, then run tools/build.sh again, right?

I've attached a "qs" shell script that I use for qemu, it also will allows serial console for logins and long text displays after an "init 3". Perhaps that will work. If not, please describe the qemu output and I'll try to help. It could be the vast variations between QEMU versions, though.

Just rename it and run "./qs" in the elks directory.

Thank you!
[qs.txt](https://github.com/jbruchon/elks/files/5247009/qs.txt)

ghaerr:
> bump up the gcc-ia16 compiler in the old tree to just before the ld.gold and segelf changes

Sorry for the dumb question, but would that mean just changing GCC_VER=606351cf50a27db562faa0768fbe760b89bbe5dc in tools/Makefile and running tools/build.sh?

tkchia:
Hello @ghaerr,

> Sorry for the dumb question, but would that mean just changing GCC_VER=606351cf50a27db562faa0768fbe760b89bbe5dc in tools/Makefile and running tools/build.sh?

Yes.  Also, this should be in the _old_ ELKS tree (with old linker scripts).  Thanks!

ghaerr:
Hello @tkchia,

I initially tried applying your uploaded patch, but can't find the directory to apply it to underneath elks/cross. I had saved a copy of the elks/cross (new linker) tree beforehand.

Then tryed to rebuild with GCC_VER=606351cf50a27db562faa0768fbe760b89bbe5dc in tools/Makefile. Unfortunately, I am running into both a "Directory not empty" build error (our problem) and a previously-discussed "Connection refused" (my system problem for some unknown reason), and can't get the compiles to start... (see below, FYI. No need to debug at this point, but would be nice for the build script to not rm -rf everything every time, to give the "Connection refused" another chance each time). I have another computer that doesn't get the Connection refused messages, if absolutely required.

I still have the new linker cross directory with PR747 as as branch, as well as the old compiler tree. I saved a copy of the old compiler tree so we still have that for testing.

Were you able to get "net start" working?

```
Gregs-Home-iMac:elks greg$ tools/build.sh
TOPDIR set to /Users/greg/net/elks
CROSSDIR set to /Users/greg/net/elks/cross
PATH set to /Users/greg/net/elks/elks/tools/bin:/Users/greg/net/elks/cross/bin:/opt/local/bin:/opt/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin
MAKEFLAGS set to
mkdir -p /Users/greg/net/elks/cross/build
rm -rf /Users/greg/net/elks/cross/build/gcc-ia16-606351cf50a27db562faa0768fbe760b89bbe5dc
cd /Users/greg/net/elks/cross/build && unzip -q /Users/greg/net/elks/cross/dist/gcc-ia16-606351cf50a27db562faa0768fbe760b89bbe5dc.zip
rm -rf /Users/greg/net/elks/cross/build/gcc-src
cd /Users/greg/net/elks/cross/build && mv gcc-ia16-606351cf50a27db562faa0768fbe760b89bbe5dc gcc-src
cd /Users/greg/net/elks/cross/build/gcc-src && contrib/download_prerequisites
--2020-09-18 11:32:34--  ftp://gcc.gnu.org/pub/gcc/infrastructure/mpfr-2.4.2.tar.bz2
           => ‘mpfr-2.4.2.tar.bz2’
Resolving gcc.gnu.org... 2620:52:3:1::246e:9693:128c, 8.43.85.97
Connecting to gcc.gnu.org|2620:52:3:1::246e:9693:128c|:21... connected.
Logging in as anonymous ... Logged in!
==> SYST ... done.    ==> PWD ... done.
==> TYPE I ... done.  ==> CWD (1) /pub/gcc/infrastructure ... done.
==> SIZE mpfr-2.4.2.tar.bz2 ... 1077886
==> EPSV ... done.    ==> RETR mpfr-2.4.2.tar.bz2 ... done.
Length: 1077886 (1.0M) (unauthoritative)

mpfr-2.4.2.tar.bz2            100%[=================================================>]   1.03M  1.98MB/s    in 0.5s    

2020-09-18 11:32:36 (1.98 MB/s) - ‘mpfr-2.4.2.tar.bz2’ saved [1077886]

--2020-09-18 11:32:36--  ftp://gcc.gnu.org/pub/gcc/infrastructure/gmp-4.3.2.tar.bz2
           => ‘gmp-4.3.2.tar.bz2’
Resolving gcc.gnu.org... 2620:52:3:1::246e:9693:128c, 8.43.85.97
Connecting to gcc.gnu.org|2620:52:3:1::246e:9693:128c|:21... connected.
Logging in as anonymous ... Logged in!
==> SYST ... done.    ==> PWD ... done.
==> TYPE I ... done.  ==> CWD (1) /pub/gcc/infrastructure ... done.
==> SIZE gmp-4.3.2.tar.bz2 ... 1897483
==> EPSV ... done.    ==> RETR gmp-4.3.2.tar.bz2 ... done.
Length: 1897483 (1.8M) (unauthoritative)

gmp-4.3.2.tar.bz2             100%[=================================================>]   1.81M  2.65MB/s    in 0.7s    

2020-09-18 11:32:38 (2.65 MB/s) - ‘gmp-4.3.2.tar.bz2’ saved [1897483]

--2020-09-18 11:32:39--  ftp://gcc.gnu.org/pub/gcc/infrastructure/mpc-0.8.1.tar.gz
           => ‘mpc-0.8.1.tar.gz’
Resolving gcc.gnu.org... 2620:52:3:1::246e:9693:128c, 8.43.85.97
Connecting to gcc.gnu.org|2620:52:3:1::246e:9693:128c|:21... connected.
Logging in as anonymous ... Logged in!
==> SYST ... done.    ==> PWD ... done.
==> TYPE I ... done.  ==> CWD (1) /pub/gcc/infrastructure ... done.
==> SIZE mpc-0.8.1.tar.gz ... 544950
==> EPSV ... couldn't connect to 2620:52:3:1::246e:9693:128c port 61480: Connection refused
make: *** [/Users/greg/net/elks/cross/build/.gcc.src] Error 1
Gregs-Home-iMac:elks greg$ tools/build.sh
TOPDIR set to /Users/greg/net/elks
CROSSDIR set to /Users/greg/net/elks/cross
PATH set to /Users/greg/net/elks/elks/tools/bin:/Users/greg/net/elks/cross/bin:/opt/local/bin:/opt/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin
MAKEFLAGS set to
mkdir -p /Users/greg/net/elks/cross/build
rm -rf /Users/greg/net/elks/cross/build/gcc-ia16-606351cf50a27db562faa0768fbe760b89bbe5dc
cd /Users/greg/net/elks/cross/build && unzip -q /Users/greg/net/elks/cross/dist/gcc-ia16-606351cf50a27db562faa0768fbe760b89bbe5dc.zip
rm -rf /Users/greg/net/elks/cross/build/gcc-src
rm: /Users/greg/net/elks/cross/build/gcc-src: Directory not empty
make: *** [/Users/greg/net/elks/cross/build/.gcc.src] Error 1
Gregs-Home-iMac:elks greg$ 
```

Thank you!

tkchia:
Hello @ghaerr,

> Were you able to get "net start" working?

I did, sort of --- `net start slip` ended up starting the `telnet` daemon, and I was able to reproduce your problem.

I am now trying to recompile `gcc-ia16` with the optimization patch reverted.  I will see how it goes.

Thank you!

ghaerr:
Hello @tkchia,

I ended up using another system to pull down your earlier tool chain just before the ld.gold and segelf changes (but including the `add $6, %sp; movw %bp, %sp` optimization) and compiling up the system. Following are the old and new kernel sizes:

Original: (note - this doesn't match up with my first post above because a debug statement was previously included there)
```
Setup is 1505 bytes.
System is 114144 B (59504 B code, 0 B fartext, 7920 B data and 46720 B bss)
System is 67456
```
Enhanced compiler (but not ld.gold):
```
Setup is 1505 bytes.
System is 113632 B (58992 B code, 0 B fartext, 7920 B data and 46720 B bss)
System is 66944
```
The system compiled with this enhanced compiler (*not* using the ld.gold linker) works. And, the code size is reduced 512 bytes, nice work!

Notice, however that the kernel data size is not changed, which is not what is happening with the ld.gold linker. That's not to say the new linker might not be removing some unused data regions somehow, but I fear something with the ld.gold linker is happening which prohibits correct system operation.

It seems that the problem is not in previously made compiler enhancements, but possibly in the ld.gold linker and/or differences in .o files produced.

Here is the output using the ld.gold compiler/linker:
```
Setup is 1505 bytes.
System is 113552 B (58992 B code, 0 B fartext, 7840 B data and 46720 B bss)
System is 66944
```
The code size is identical but the data section is 80 bytes shorter.

Let me know if you can think of further test cases that might help track down what is happening. Perhaps we should take a look at system.map to try to see where the missing .data might be. Or perhaps the issue is one of data-reordering.

I can occasionally reproduce the telnet crash issue on an otherwise working kernel by removing the down(&rwlock)/up(&rwlock) semaphores protecting a kernel critical region. The global semaphore is declared "static short rwlock" and all are in elks/net/ipv4/af_inet.c. I haven't yet tried chasing down whether or not rwlock is perhaps misplaced or being overwritten with the ld.gold linker.

It is possible, although unlikely, that this crash is just exposing another kernel bug as a result of radically reordered text or data offsets... although I would guess that it might be prudent to understand why the kernel data segment isn't the same size when ld.gold is used.

Thank you!

ghaerr:
Hello again @tkchia,

I have been looking a the system.map file from the ld.gold linker. I see there are two places, that while I can't see how they cause our telnet failure, could be a problem, and likely is the reason for the smaller data size produced. These are the "merge strings" and "merge constants" section, which can be seen just following .nildata. These are produced as .rodata, but I can't guarantee that the kernel in fact treats all strings as read-only, so merging could be a problem. Is there a linker option too turn off both these merges?

```
.nildata       0x00030000        0x4 arch/i86/boot/crt0.o
 ** merge strings
                0x00030004     0x125b
 .rodata        0x00031260        0x8 fs/fs.a(super.o)
 .rodata        0x00031268        0xa fs/fs.a(fcntl.o)
 .rodata        0x00031272       0x1a fs/fs.a(inode.o)
 .rodata        0x0003128c        0x5 fs/fs.a(namei.o)
 .rodata        0x00031292       0x1e net/net.a(socket.o)
 .rodata        0x000312b0       0x18 fs/minix/minixfs.a(inode.o)
 .rodata        0x000312c8       0x20 fs/msdos/msdos.a(misc.o)
 ** merge constants
                0x000312e8        0x2
 .rodata        0x000312ea       0x3a arch/i86/drivers/char/chr_drv.a(serial.o)
 .rodata        0x00031324       0x12 arch/i86/drivers/char/chr_drv.a(dircon.o)
 .rodata        0x00031336       0x69 arch/i86/drivers/char/chr_drv.a(xt_key.o)
 .rodata        0x0003139f        0xd arch/i86/drivers/block/blk_drv.a(bioshd.o)
 .rodata        0x000313ac        0x4 arch/i86/drivers/block/blk_drv.a(idequery.o)
 .rodata        0x000313b0       0x1a arch/i86/drivers/char/chr_drv.a(mem.o)
 .rodata        0x000313cc        0xc arch/i86/drivers/net/net_drv.a(ne2k.o)
 .data          0x000313d8        0x8 arch/i86/boot/crt0.o
```

I'm kind of grasping at straws here, but this is an ld.gold enhancement not found in the old linker. It seems the new linker could in fact be more sophisticated, the merge option could be useful in the future, but not likely without more information as to when a merge takes place (e.g. const keyword used, etc) and close inspection of ELKS kernel code. I am guessing the compiler and linker treat C string literals as constants, for instance.

Thank you!

ghaerr:
Hello @tkchia,

> Is there a linker option too turn off both these merges?

I've added -fno-merge-constants to elks/Makefile-rules. We now have exactly the same code and data size, but telnet still crashes :(

I plan on continuing to look at system.map differences, so far can't find anything else. Very strange. Perhaps this is just exposing a latent kernel bug due to different data or code sizes or data offsets. 

I may try turning off the `add $6, %sp; movw %bp, %sp` optimization but still use the ld.gold linker, just to give that scenario a try.

[EDIT: here is some more information on string-table-related problems that seem to come from binutils, not gcc, and reside in ELF files]

https://stackoverflow.com/questions/51551068/how-to-stop-gcc-from-combining-string-literals-that-share-the-same-suffix


Thank you!

ghaerr:
More strange test results (using ld.gold compiler, this PR and -fno-merge-constants):

I thought to play around with compiler optimization, just to see what happens.

Editing elks/Makefile-rules:
- Remove -Os entirely: kernel compilation fails, bad asm-offsets.h produced.
- Use -O1: kernel compilation succeeds, code barely fits at 65120 bytes, "telnet localhost" same exact error, telnet outside of ELKS hangs. Rest of system seemingly operates normally. Kernel data size is 8160, rather than 7920. Not sure why data segment is 240 bytes larger.
- Use -O2 or -O3: kernel compilation fails, ".fartext moved backwards..." (I think kernel too large).

All tests were made changing kernel only, using working, unmodified applications built from optimizing compiler before ld.gold that work with non-ld.gold built kernel (including ktcp application). Only kernel changed!

The fact that all other aspects of the system seem to work well, and that the only seen problems appear to be in kernel networking is very strange. I'm begging to think networking is broken in other ways and new linker is just illuminating unknown networking problems.

- Were any other gcc library routines changed in between stack optimization compiler version and ld.gold compiler version?

I don't have a debugger that can allow me to step through the execution of af_inet.c code to trace bad write syscall in telnet.

tkchia:
Hello @ghaerr,

Not sure if the merging of strings is causing the problem (though I will not rule out any possibility).  But I am starting to get some hints of when at run time the lossage is occurring.  If I add debugging statements to `elkscmd/ktcp/tcpdev.c` like so:
```
static void tcpdev_write(void)
{
    ...
    cb->data = db->data;
printf("!! %p %#x\n", sock, db->size);

    tcp_output(cb);

printf("! %p %#x\n", sock, db->size);
    retval_to_sock(sock, db->size);
}
```
then the console output suggests that `db->size` was corrupted with a data pointer some time during the `tcp_output(.)`:
```
!! 009810 3
! 009810 9630
write: Unknown error 27088
```
The next question I suppose is what exactly is causing this corruption.

Thank you!

tkchia:
Hello @ghaerr ,

Another weird thing I found: the output `system.map` does not seem to mention any `COMMON` area (though it does have a `.bss`).  The `tdout_buf` happens to go into the `COMMON` area, according to `tcpdev.o`.  I will need to check and see if that might be related to the current problem.

(**Update**: OK, `system.map` _does_ mention a common area, but `ld.gold` lays it out differently.)

Thank you!

ghaerr:
Hello @tkchia,

> The next question I suppose is what exactly is causing this corruption.

Thanks to your fantastic debugging, I have identified and understand the cause of corruption, which is a previously unknown user-mode buffer overwrite in ktcp. I have a fix which allows the networking subsystem to work properly. I am still chasing down exactly why this occurs only with the ld.gold-build kernel. More details coming.

Thank you!!

ghaerr:
Hello @tkchia,

Thanks to your help, I have definitively determined the reasons behind the `telnet localhost` failure on the ld-gold-built kernel. It is now certain that the new compiler and linker are not the actual cause of the problem, and I feel with the addition of adding `-fno-merge-constants` to elks/Makefile-rules, this PR will be ready for commit.

The good news is that this PR has uncovered various pre-existing problems with the kernel which I intend to address later. I am preparing a PR now for the networking fix, which I would like to commit first, then after you adding the compiler option above to this PR, if you are ok with it, this PR should be committed.

For your information, here is exactly how the ld.gold kernel exposed the longstanding networking bug. In this scenario, we are dealing with `telnet localhost`, which after connecting with `telnetd`, tries writing three telnet negotiation sequences of 3 characters each using the network socket to telnetd. Since most of networking is implemented in the user-mode `ktcp` application, the tcpdev kernel interface intercepts the socket read/writes to pass them to ktcp (and back).

1. The tcpdev_write routine in ktcp has telnet's write request of 3 characters. The routine uses a global buffer for all communication between the kernel and ktcp. Ktcp gets the write data and size from the buffer, sends it to tcp_output, then returns the size written to the socket:
```
struct tdb_write *db = (struct tdb_write *)sbuf;
sock = db->sock;
cb->datalen = db->size;
cb->data = db->data;
tcp_output(cb);
retval_to_sock(sock, db->size);
```
2. Tcp_output forms an TCP packet, sends it via ip_sendpacket, which ends up when routed to localhost, instead of though the link layer, back up to ip_recvpacket and then tcp_process. Tcp_process then processes the "received" packet. As part of the that processing, tcp_process must tell the kernel there is received data available for the socket. That check involves using the tcpdev interface, so tcpdev_checkread is called, which uses the same global buffer for kernel/ktcp transfer:
```
struct tdb_return_data *ret_data = (struct tdb_return_data *)sbuf;
ret_data->type = TDT_AVAIL_DATA;
ret_data->ret_value = cb->bytes_to_push;
ret_data->sock = sock;
write(tcpdevfd, sbuf, sizeof(struct tdb_return_data));
```
Thus, the previous db->size data is overwritten, since both use the same sbuf buffer for kernel communication.

The two structures used are tdb_write and tdb_return_data:
```
struct tdb_write {
    unsigned char cmd;
    struct socket *sock;
    int size;
    int nonblock;
#define TDB_WRITE_MAX   100
    unsigned char data[TDB_WRITE_MAX];
};

struct tdb_return_data {
    char type;
    int ret_value;
    struct socket *sock;
    int size;
    unsigned char data;
};
```
Note the size field overlaps the socket field in the structures. The size field is overwritten with a socket \*.

3. Tcpdev_write passes the size return value (now a socket \*) to the kernel. The kernel inet_write routine checks the socket return value as follows:
```
        ret = ((struct tdb_return_data *)tdin_buf)->ret_value;
        ...
        if (ret < 0) {
                ...
                return ret;
```
Thus, when the return value is negative, telnet gets back the address of a socket, rather than the errno. If the return value is positive, then no error, and the routine eventually returns the original correct size passed originally to it.

In the kernel, a socket is initially allocated through a long sequence that ends up allocating an incore inode structure from a `static struct inode node_block[NR_INODE]` in inode.o. Thus, the socket address returned incorrectly above depends particularly on where the inode block is allocated in .bss by the linker.

In the old kernel, the address was:
```
 .bss           0x0000000000005982     0x1e06 fs/fs.a(inode.o)
 .bss           0x0000000000007788        0x0 fs/fs.a(file_table.o)
```
In the new kernel:
```
 .bss           0x000390aa     0x1e06 fs/fs.a(inode.o)
```

That is, in the old kernel, a socket \* was always linked at an address less than 0x8000, while in the new kernel the address is 0x90aa or above. Thus, the tcpdev_write routine would interpret the overwritten return value (now socket \*) as < 0, and return an error, which caused telnet to display the write error messages. The older kernel never returned the wrong write count, as the inet_write routine only checks for negative ret_values, and ignores positive (incorrect) return values, returning instead the originally-passed write size.

Thank you, and I look forward to using the new linker!!

tkchia:
Hello @ghaerr,

> Thank you, and I look forward to using the new linker!!

Cool, and thanks!

It happens that I had also been looking at the code around `sbuf[]` (though you figured out the bug first ☺☺).

Some code cleanups that I would like to propose:
  * It seems to me that in many cases we do not really _need_ to have a `struct tdb_return_data *` specifically pointing to the start of `sbuf[]` --- and a simple automatic variable would do the job just fine.  E.g. instead of the above we can have
    ```
    struct tdb_return_data ret_data;
    ret_data.type = TDT_AVAIL_DATA;
    ret_data.ret_value = cb->bytes_to_push;
    ret_data.sock = sock;
    write(tcpdevfd, &ret_data, sizeof(struct tdb_return_data));
    ```
    This should work for most of the `elkscmd/ktcp/tcpdev.c` routines, except for `tcpdev_read ()`.
  * And, since the final `data` field in `struct tdb_return_data` is really the start of a variable-length array, so perhaps it is better expressed as a C99 "flexible array member", like so:
    ```
    struct tdb_return_data {
        ...
        unsigned char data[];
    };
    ```
    If anything, this will give GCC a better indication of what is going on, and also (slightly) simplify the arithmetic in the kernel and `ktcp` for counting the bytes to pass around through `/dev/tcpdev`.

But all this can probably wait for a separate PR. ☺

Thank you!

tkchia:
Hello @ghaerr,

> I feel with the addition of adding `-fno-merge-constants` to elks/Makefile-rules, this PR will be ready for commit.

I think `-fmerge-constants` is a rather useful optimizations to have.  Do you happen to see (or foresee) any particular problems of merging constants within the ELKS kernel?  Thanks!

ghaerr:
Hello @tkchia,

> a simple automatic variable would do the job just fine
> struct tdb_return_data ret_data;

Agreed. I started noticing that as well after looking at all the sbuf[] stuff.

> This should work for most of the elkscmd/ktcp/tcpdev.c routines, except for tcpdev_read ().

Agreed. I'll prepare another PR where I think we can lose the sbuf[] buffer entirely. I am a bit concerned about ktcp stack usage, and will make the determination after more inspection.

> And, since the final data field in struct tdb_return_data is really the start of a variable-length array, so perhaps it is better expressed as a C99 "flexible array member",

Agreed. I will look into changing that also.

> and also (slightly) simplify the arithmetic in the kernel and ktcp for counting the bytes to pass around through /dev/tcpdev.

I hadn't considered that, thank you.

> I think -fmerge-constants is a rather useful optimizations to have. Do you happen to see (or foresee) any particular problems of merging constants within the ELKS kernel? 

I agree it is a useful optimization. Given the problems we just went through, I thought it best to have a look at the kernel first, before turning that option on. I would like to learn more about exactly what is put into the .rodata section, then peruse the kernel, then turn the option on, as a way to keep unforseen bugs to a minimum. We are at a stage where I am more interested in stability than quick changes, given the amount of activity and testing going on. 

[EDIT: My understanding is that the old linker won't merge constants, and it is a new feature in the ld.gold linker to do so, unless turned off. Thus, the default compiler option of merging constants was effectively ignored by the old linker, but would need to be turned off in the new linker, in order to not change the .data segment, correct?]

I would like to suggest that it be turned off with this commit, so as to not change to much at once, then turned on again very shortly in another commit, after understanding .rodata a bit better?

On another note, may I ask how you were able to find the initial db->size overwrite bug so quickly? Was it through elksemu? I am not able to run that on macOS, and am relegated to printk debugging for the large part.

Thank you!

ghaerr:
Hello @tkchia,

> Do you happen to see (or foresee) any particular problems of merging constants within the ELKS kernel?

Here are some quick possible issues. I haven't yet looked much at the objdump .o output on these from main.o:

```
static char bininit[] = "/bin/init";
static char *init_command = bininit;

static char *argv_init[80] = { NULL, bininit, NULL }; <--- is this .rodata? No, but is init_command? How about bininit?
       /* run /bin/init or init= command, normally no return*/
        run_init_process_sptr(init_command, (char *)argv_init, argv_slen); <--- arg_init will be overwritten by parse code

num = sys_open(s="/dev/console", O_RDWR, 0);  <---- variable assigned to .rodata, could get changed, still is .rodata?

static char *root_dev_name(int dev)
{
        int i;
#define NAMEOFF 13
        static char name[18] = "ROOTDEV=/dev/";  <----- overwritten at [13]

        for (i=0; i<5; i++) {
                if (devices[i].num == (dev & 0xfff0)) {
                        strcpy(&name[NAMEOFF], devices[i].name);
                        if (i < 3) {
                                if (dev & 0x03) {
                                        name[NAMEOFF+3] = '0' + (dev & 3);
                                        name[NAMEOFF+4] = 0;
                                }
                        }
                        return name;
                }
        }
        return NULL;
}

/*
 * Convert a /dev/ name to device number.
 */
static int parse_dev(char * line)
{
        int base = 0;
        struct dev_name_struct *dev = devices;

        if (strncmp(line,"/dev/",5) == 0)     <----- will this get merged with static buffer above? No problem yet as [13] is offset of overwrite, but could be
                line += 5;
        ...
```

A quick look at .o output shows that `const` is not needed for strings to be put in .rodata. I'm not yet sure about whether non-const pointers to strings move the string to .data from .rodata.

Also, don't understand what/when a (non-string) constant is, and when that is moved to .rodata.

Thank you!

ghaerr:
Hello @tkchia,

I have looked further into how gcc differentiates between .rodata and .data sections, and have answered most of my questions asked in the post above. In addition, with the ld.gold linker string/constant merge default ON, only 80 bytes get merged, so the algorithm is obviously quite conservative. I am not worried about string merging in the kernel now that I understand it better.

So - I'm fine not having -fno-merge-constants added to this PR. I have also tested it with the fixed networking, and its fine.

Go ahead and commit this PR as-is. Nice work, and thank you for your help in tracking down the networking problem exposed by it. I'll commit my cleanup PR #754 after this is committed.

Thank you!

tkchia:
Hello @ghaerr,

> On another note, may I ask how you were able to find the initial db->size overwrite bug so quickly? Was it through elksemu?

I guess I kind of lucked out. :-)  I decided to focus examining the code in the kernel and userland that might be dealing with return byte counts.  I put some `printk`'s inside `elks/arch/i86/drivers/char/tcpdev.c`, but did not find anything useful there, so next I looked at the program interacting with `tcpdev`, i.e. `ktcp`.

I suspect `elksemu` would not have been very useful for finding this bug --- since it does not emulate the `/dev/tcpdev` device file which `ktcp` depends on.

Thank you!

ghaerr:
Hello @tkchia,

> I suspect elksemu would not have been very useful for finding this bug --- since it does not emulate the /dev/tcpdev device file which ktcp depends on.

I see. For some reason I thought that elksemu worked by executing all of ELKS as a Linux debug process - including the ELKS kernel interpretation of its own /dev entries. In the /dev/tcpdev case, the ELKS kernel just moves data back and forth between itself and the ELKS task associated with it. I figured elksemu doesn't interpret any ELKS /dev entries, but instead emulates BIOS INT calls for disk I/O and possibly console output. I will have to look at it further on a Linux system.

I have some further interest in emulating the ELKS environment using other tools and on other not-quite-PC 8086 environments, so knowing exactly where ELKS touches the hardware, versus uses BIOS INT calls, or still in some cases BIOS memory pokes, is very important.

In any case, nice work! I had previously fixed just this problem within the TCPDEV/af_inet code, and was convinced the issue might still be there, and wasn't ready to consider that ktcp itself could be corrupting input to the kernel.

Thank you!

ghaerr:
Hello @tkchia,

> For some reason I thought that elksemu worked by executing all of ELKS as a Linux debug process 

I looked through elksemu - I understand now it just runs ELKS applications. I thought it ran the ELKS kernel itself. I get it now.

Thank you!

PR #749, opened 2020-09-13
[kernel] Improve segment allocator

mfld-fr:
Same algorithm as for heap allocator.

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -11,7 +11,10 @@
 
 #include <arch/segment.h>
 
-// TODO: reduce size
+// Minimal segment size to be useful
+// (= size of the smallest allocation)
+
+#define SEG_MIN_SIZE 1
 
 // Segment descriptor
 
@@ -20,6 +23,7 @@
 // whenever that mode comes back one day
 
 static list_s _seg_all;
+static list_s _seg_free;
 
 
 // Split segment if enough large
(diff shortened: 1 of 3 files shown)


ghaerr:
I noticed more changes to heap.c free list, as well as extra size moved back from 12 to 16. What is the purpose of each?
Is the near heap allocator still under construction?

mfld-fr:
- Segment descriptor now contains the free list node, so its size is now 12 + 4 = 16 bytes.
- No design change in heap, just coding style for better lisibility.

ghaerr:
Thanks, had just noticed reason for + 16. Not sure its a great idea to have to hand-tune with constant, as will surely be missed if structure size changes in future.

I don't know if removing SEG_FLAG_USED bit in malloc.c line 70 change is a good idea: this 1) forces non-zero tag to be used with all calls to seg_alloc (not yet checked or enforced), and 2) is inconsistent with ORing SEG_FLAG_USED usage in line 86, right? Perhaps line 70 should not be changed?

We need to be careful with the business of using tag == 0 possibly meaning free, for both heap and segment allocator. For instance, a temp allocation of near heap buffer is currently passing a 0 tag (since there was no otherwise applicable tag, before your changes), in #742 (idquery.c, line 79).

So far, changes just increase code and allocation sizes with no benefit (except possible benefit through changing allocation scheme using new free list, of course). Would be nice to discuss or have a test scenario where varying main memory allocation schemes could be used to actually determine possible system benefit.

If previous large reduction to kernel stack size (988 -> 512) in #696 can't be kept, due to excessive kernel stack used with FAT filesystem found in https://github.com/jbruchon/elks/pull/742#issuecomment-690913748, free space of near heap will become very important again. Total kernel stack usage per system call and per application can now be calculated at runtime using STRACE option implemented in #686.

Issue #697, opened 2020-08-12
[INFORMATION] How to transfer files to/from elks


Mellvik:
### Intro
This is an informational post, not a bug report. 

ELKS is rapidly getting closer to being self-sufficient. Being able to transfer files into and out of the system is an important milestone towards that goal. Recent improvements in networking support and serial port reliability & speed have made that possible.

**Serial**
Transferring limited size ascii files can be done with reasonable reliability via serial lines and terminal emulation. `cat > filename`and then paste text into the terminal is reasonably efficient. If this fails marginally, turn off echo in order to reduce the interrupt load on the system:
```
stty raw; cat > filename; stty -raw
```
or use the `sercat` command:
```
sercat > filename
```
Terminate such transfers with ^D (EOF).

To check whether the transfer was error free, use the `sum` command at both ends, after ensuring that the byte count is exactly the same (`ls -l`). Note that it takes only an extra newline or CR at the end of the file to make the `sum` command useless.

Even binary files may be transferred this way by first converting them to ascii. Use the `hexdump` command on Linux:
```
hexdump -e '"%06.6_ax:" 8/1 " %02x" " " 8/1 " %02x" "  " ' -e '16/1 "%_p" "\n"' "$@"
```
to create the ascii file, then `hd -r` on ELKS to convert it back.
A far more efficient (and traditional!) way is using `uuencode` and `uudecode`. Instead of turning 16 bytes into 69 which `hd`does, `uuencode`converts 2 bytes into 3 ascii characters. The commands are part of the uucp package. The (simplified) syntax is
```
uuencode newname < inputfile > outputfile
```
where `newname` is the filename to be used when the file is decoded and recreated. Even the mode (protection codes) of the file is stored in the encoded file and will be recreated during decoding if possible. To decode, simply give the command
```
uudecode < inputfile
```
and the old file (and filename + modes) will be recreated. There are many options to override the defaults, check the man-pages for details.

Taking ascii files from ELKS via serial is usually easy (and fully reliable) since most systems will have no problems keeping up with the speed of the ELKS system.

Mechanisms to make serial transfers binary and fully reliable - Kermit, X/Y/Z-modem etc. - are forthcoming. 

### ELKS networking
[...]

PR #750, opened 2020-09-14
[cmd] urlget: Bugfixes and enhancements.

Mellvik:
`urlget` turns out to the universal network file import tool for ELKS, supporting 3 different URLs and 3 different 'incarnations': `urlget`, `httpget` and `ftpget`. This PR contains a number og fixes and enhancements - and is accompanied by a total rewrite of the ELKS file transfer document (#697).

The ELKS build system does not currently support the creation of hard links for programs with multiple personalities, to you'll have to create those locally:
```
ln /bin/urlget /bin/ftpget
ln /bin/urlget /bin/httpget
```

Options to add outgoing file transfer - via ftp and tcp/nc - is in the works.

### DIFF
--- a/elkscmd/inet/urlget/urlget.c
+++ b/elkscmd/inet/urlget/urlget.c
@@ -3,6 +3,16 @@
  * 04/05/2000 Michael Temari <Michael@TemWare.Com>
  * 09/29/2001 Ported to ELKS(and linux) (Harry Kalogiroy <harkal@rainbow.cs.unipi.gr>)
  *
+ * Implements HTTP POST & GET, FTP GET and directory listings, TCP GET (typical via 
+ * netcat at the other end.
+ * 
+ * Assumes hard links to the approproate names - ftpget, urlget, tcpget.
+ * Optons (http only):
+ * 	-h -- include header in output stream
+ * 	-d -- discard data
+ * 	-p -- post instead of get, data to post (ascii/UTF) is appended to the URL (after a '?').
+ *	-v -- for ftp, verbose file listing
+ *
  */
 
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Thanks @Mellvik!

I will find some time to support links in Applications ASAP.

Do you know that the Wiki uses the same format (Markup) as PRs and Issues? I'm thinking perhaps you should copy #697 into the Wiki. This can easily be done by creating a new page, rather than trying to edit existing pages. Try it on the Wiki page by hitting "New Page" and then setting the title to "File Transfers". You can then edit and update it there afterwards, and the "File Transfers" will show in the sidebar.

Mellvik:
Thanks, @ghaerr -
I agree it’s time to move the file transfer doc. I pasted it in and will make the necessary adapting To the new context later today.

AS indicated in the PR, I believe further work should focus on FTP as the more ubiquitous of the protocols, do you agree?

—M
[ iPhone ]

ghaerr:
>  I believe further work should focus on FTP as the more ubiquitous of the protocols, do you agree?

It would certainly be nice to have a small FTP server (and standard FTP client). A standard FTP client isn't great for automation like ftpget is though. What file transfer features are on your list? Are you thinking about updating an ELKS distribution while the system is running, or other problems?

HTTP PUT would also be nice, since HTTP is even more widely supported these days. I looked at busybox's HTTP put but that was implemented through CGI and shell scripts. I haven't been able to find a small general-purpose HTTP put implementation.

Mellvik:
Thanks @ghaerr,

My plan is (and somewhat adjusted by you feedback):
- add ftpput (single file)
- add httput (pun intended)
both of these are simple add ons to the current code
- add a simple ftp client (interactive), supporting just a few commands: get, put, dir, ldir, type, cd, then globbing and put/mget, md.
This one is a slightly bigger project, but I've looked at various clients available in source, and found them utterly complicated - for two reasons: They support everything (in the FTP protocol) and the support every platform (ifdef galore).

A tiny ftp server would be great, I agree, I haven't looked at that yet.

—Mellvik

PR #752, opened 2020-09-18
[cmds] Add link capability to Applications installation

ghaerr:
Creates links from urlget to ftpget and httpget as requested in #750 

Use ":::" to specify linked file in Applications table.
Uses symlinks for now.
FAT filesystems will copy file separately.

@Mellvik, please test the build on Linux if possible, tested only on macOS. The mechanism creates a symlink rather than hard link on the disk image for now, there are some issues rewriting `mfs` to support hard links that will be done later. Things seem to work well, although I am noticing a possible bug in the usage display output when urlget is run from the shell as "/bin/ftpget" (and /bin/httpget).

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -157,6 +157,8 @@ inet/telnet/telnet				:net
 inet/telnetd/telnetd			:net
 inet/tinyirc/tinyirc			:net
 inet/urlget/urlget				:net
+inet/urlget/urlget :::ftpget	:net
+inet/urlget/urlget :::httpget	:net
 #mtools/mcopy					:mtools					:1400k
 #mtools/mdel					:mtools					:1440k
 #mtools/mdir					:mtools					:1400k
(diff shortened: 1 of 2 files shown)


Mellvik:
Thanks @ghaerr!!

I'll do the Linux testing over the weekend. 
My Linux (raspberrypi2b) repo is out of sync and needs a full build, including tools. The tools alone take something like 6 hrs,  it's a speed that makes it possible to actually read the messages as they tickle by…


—Mellvik

ghaerr:
> The tools alone take something like 6 hrs

Well, we're likely to get an additional tool update soon, #747 is pending and may need another update, not sure. So, no rush on the Linux testing. I'll commit this now and you can play with it on macOS and/or QEMU.

Mellvik:
Well, this all goes on in the background.

A day later - and all is good, the (symbolic) links
are there.

Thanks.

-M

Issue #745, opened 2020-09-08
[ktcp] TCP performance


Mellvik:
First, networking status summary per september 8th, 2020: [@ghaerr - please correct if my details are off]
 - Serial IP (SLIP/CSLIP): Working @ moderate baud rates, needs serial HW flow control to get to the next level.
 - Ethernet Link level NE2k: Stable
 - ICMP, ARP, TCP: Stable
 - Telnet/PTYs: Not stable
 - The ELKS httpserver can deliver any file and file size reliably.
 - `urlget` is reliable for incoming data.
 - `urlget` is also `ftpget`, which currently does not work but 'in the works'.

Stable TCP-level communication enables file transfers (#697) to/from ELKS, and it's time to get transfer speeds from 'working' to 'useable'. Currently, transfer speed from ELKS (IDE harddisk) via HTTP is 25.5KB/s, packet size 140 bytes (Compaq Portable 386/20, ISA16 NE2k card). For incoming file transfer (to /dev/null) the advertised window size is 255, effective packet size is 295, speed 36KB/s. We may be able to improve both by an order of magnitude, but half of that is enough to make for a very useable system.

@ghaerr - I'm assuming there are some low threshold improvements that can be made immediately, any hints as to where to start?

--Mellvik

ghaerr:
Here is my take on the status of TCP/IP. I'm not sure I'd use "Stable", but instead "Buggy":

 * Serial IP (SLIP/CSLIP): Working @ moderate baud rates, needs serial HW flow control to get to the next level.
SLIP - Not buggy; CSLIP - Not working at all. Both dependent on serial driver, not TCP/IP for higher baud rates. I can run at 38400 on all systems using fast serial driver, and 115200 in my faster box using regular driver. Agreed serial HW flow control likely to help, but that's really quite separate from TCPIP and unclear as to how much throughput will actually increase, given max serial speed on old boxes even with flow control.

 * Ethernet Link level NE2k: Stable - AGREE, needs more testing since you're the only one running it.

 * ICMP, ARP, TCP: Stable - ICMP, ARP: Not buggy. TCP - Stable, but extremely buggy on lost packets and closing sockets. No sliding window implemented, and unable to accept large window (or packets) since currently the only real flow control TCP has is to drop a packet when the upstream isn't ready, or doesn't have buffer space to accept data.
[...]

Issue #521, opened 2020-04-05
Disk/floppy performance


Mellvik:
Finally some disk performance metrics on physical HW.

HW: 286 12MHz, FD1,2M, HD type 17 42MB Conner, booting from Minix floppy
- `ls -l /bin` [7s]
- `cp /bin/vi /bin/xx` [28.5s]
- `ps  [2s]`
- `cat /etc/rc.d/rc.sys` [4s]
- `cp 245kfile xx`  [HD to HD, FATfs] [24s] (DOS 3s, venix 13s on same disk, Unix V7 filesystem)

--Mellvik

Mellvik:
Live performance:
https://drive.google.com/file/d/1t49cysn2xT2QoqvYD-UL1WIamPNnYEDk/view?usp=drivesdk

ghaerr:
That is quite slow... although seeing the screen does remind me of the speed of yesterday's computers.

After we get your latest round of basic bugs fixed, we can look seriously into ELKS I/O being quite slow compared to MSDOS And Venix. I suspect a lot has to do with ELKS only doing single sector I/O on all disks.

Mellvik:
> ﻿
> That is quite slow... although seeing the screen does remind me of the speed of yesterday's computers.
> 
> After we get your latest round of basic bugs fixed, we can look seriously into ELKS I/O being quite slow compared to MSDOS And Venix. I suspect a lot has to do with ELKS only doing single sector I/O on all disks.
> 
> —
> 
Agree, and yes, what you see is the real thing of yesteryear. My 386/20 is more pleasant, but admittedly, coming to DOS and starting Elvis in one second is ....
Then again, notice the dir listing slowing down. I believe this is unique to the FAT filesystem, but need to be more empirical on that.
A reasonable ambition for elks would be to beat Venix. 

-M

Mellvik:
As one problem after the other are being fixed (fast),  and new features in ELKS seem to come by the day, I'd like to keep this issue close to the top of the list. 
From my perspective - on physical systems, fd/hd performance is now the primary barrier between a 'really useable system' (1.0) and 'early development' (0.3).

ghaerr:
> From my perspective - on physical systems, fd/hd performance is now the primary barrier between a 'really useable system' (1.0) and 'early development' (0.3).

[...]

PR #760, opened 2020-09-22
eth: WD8003 driver (based on ne2k driver code)

pawosm-arm:
This driver was tested with my SMC 8003WC 8-bit ISA card.

Note that some of the corners were cut, e.g. less popular
ioctls are stubbed with -ENOSYS, overflow interrupt is handled
by brutal reset of entire thing and so on.

As both ne2k and wd8003 cards and their clones are based on
8390 chip, it would be beneficial to merge those drivers into one
or extract common part of them e.g. into lib8390 code.

The card was configured using DOS ezsetup utility as such:
IRQ: 2
I/O base: 0x240
I/O mem: 0xce000 (right after XT-IDE I/O mem area)
DOS packet driver interrupt: 0x60

### DIFF
--- a/elks/arch/i86/drivers/net/Makefile
+++ b/elks/arch/i86/drivers/net/Makefile
@@ -31,6 +31,9 @@ OBJS = net_drv.o
 ifeq ($(CONFIG_ETH_NE2K), y)
 OBJS += ne2k-asm.o ne2k.o
 endif
+ifeq ($(CONFIG_ETH_WD), y)
+OBJS += wd.o
+endif
 
 all: net_drv.a
 
--- a/elks/arch/i86/drivers/net/config.in
+++ b/elks/arch/i86/drivers/net/config.in
@@ -7,7 +7,9 @@ mainmenu_option next_comment
 	if [ "$CONFIG_ARCH_SIBO" != "y" ]; then
 		bool 'Ethernet device support' CONFIG_ETH n
 		if [ "$CONFIG_ETH" = "y" ]; then
-			bool 'NE2K device driver' CONFIG_ETH_NE2K n
+			choice 'Select Ethernet device' \
+				"ne2k   CONFIG_ETH_NE2K \
+				 WD/SMC CONFIG_ETH_WD" ne2k
 		fi
 	fi
 endmenu
--- a/elks/arch/i86/drivers/net/net_drv.c
+++ b/elks/arch/i86/drivers/net/net_drv.c
@@ -8,4 +8,7 @@ void eth_init(void)
 #ifdef CONFIG_ETH_NE2K
 	eth_drv_init();
 #endif
+#ifdef CONFIG_ETH_WD
+	eth_drv_init();
+#endif
 }
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -1,0 +1,626 @@
+/*
+ *
+ * WD Ethernet driver - supports NICs using the WD8003 and compatible
+ * chip sets. Tested with SMC 8003WC ISA 8-bit card.
+ *
+ */
+
+#include <arch/io.h>
+#include <arch/ports.h>
+#include <arch/limits.h>
+#include <arch/segment.h>
+#include <linuxmt/memory.h>
+#include <linuxmt/errno.h>
+#include <linuxmt/major.h>
+#include <linuxmt/ioctl.h>
+#include <linuxmt/fcntl.h>
+#include <linuxmt/fs.h>
+#include <linuxmt/sched.h>
+#include <linuxmt/limits.h>
+#include <linuxmt/mm.h>
+
+#define DEBUG_ETH	0	/* set =1 for debugging */
+
+#if DEBUG_ETH
+#define debug_eth	printk
+#else
+#define debug_eth(...)
+#endif
+
+/* I/O delay settings */
+#define INB	inb	/* use inb_p for 1us delay */
+#define OUTB	outb	/* use outb_p for 1us delay */
+
+#define WD_SHMEMSEG	0xce00U
+
+#define WD_STAT_RX	0x0001U	/* packet received */
+#define WD_STAT_TX	0x0002U	/* packet sent */
+#define WD_STAT_OF	0x0010U	/* RX ring overflow */
+
+#define WD_START_PG	0x00U	/* First page of TX buffer */
+#define WD_STOP_PG	0x20U	/* Last page + 1 of RX ring */
+
+#define TX_2X_PAGES	12U
+#define TX_1X_PAGES	6U
+#define TX_PAGES	TX_1X_PAGES
+
+#define WD_FIRST_TX_PG	WD_START_PG
+#define WD_FIRST_RX_PG	(WD_FIRST_TX_PG + TX_PAGES)
+
+#define RX_PAGES	(WD_STOP_PG - WD_FIRST_RX_PG)
+
+#define ETH_PAGE_SIZE	256U
+
+#define WD_RESET	0x80U	/* Board reset */
+#define WD_MEMENB	0x40U	/* Enable the shared memory */
+#define WD_IO_EXTENT	32U
+#define WD_8390_OFFSET	16U
+#define WD_8390_PORT	(WD_PORT + WD_8390_OFFSET)
+
+#define E8390_RXCONFIG	0x04U	/* EN0_RXCR: broadcasts, no multicast,errors */
+#define E8390_RXOFF	0x20U	/* EN0_RXCR: Accept no packets */
+#define E8390_TXCONFIG	0x00U	/* EN0_TXCR: Normal transmit mode */
+#define E8390_TXOFF	0x02U	/* EN0_TXCR: Transmitter off */
+#define E8390_STOP	0x01U	/* Stop and reset the chip */
+#define E8390_START	0x02U	/* Start the chip, clear reset */
+#define E8390_TRANS	0x04U	/* Transmit a frame */
+#define E8390_RREAD	0x08U	/* Remote read */
+#define E8390_RWRITE	0x10U	/* Remote write  */
+#define E8390_NODMA	0x20U	/* Remote DMA */
+#define E8390_PAGE0	0x00U	/* Select page chip registers */
+#define E8390_PAGE1	0x40U	/* using the two high-order bits */
+#define E8390_PAGE2	0x80U	/* Page 3 is invalid */
+#define E8390_CMD	0x00U	/* The command register (for all pages) */
+
+/* For register EN0_ISR. */
+#define E8390_TX_IRQ_MASK	0x0aU
+#define E8390_RX_IRQ_MASK	0x05U
+
+/* Page 0 register offsets. */
+#define EN0_CLDALO	0x01U	/* Low byte of current local dma addr  RD */
+#define EN0_STARTPG	0x01U	/* Starting page of ring bfr WR */
+#define EN0_CLDAHI	0x02U	/* High byte of current local dma addr  RD */
+#define EN0_STOPPG	0x02U	/* Ending page +1 of ring bfr WR */
+#define EN0_BOUNDARY	0x03U	/* Boundary page of ring bfr RD WR */
+#define EN0_TSR		0x04U	/* Transmit status reg RD */
+#define EN0_TPSR	0x04U	/* Transmit starting page WR */
+#define EN0_NCR		0x05U	/* Number of collision reg RD */
+#define EN0_TCNTLO	0x05U	/* Low  byte of tx byte count WR */
+#define EN0_FIFO	0x06U	/* FIFO RD */
+#define EN0_TCNTHI	0x06U	/* High byte of tx byte count WR */
+#define EN0_ISR		0x07U	/* Interrupt status reg RD WR */
+#define EN0_CRDALO	0x08U	/* low byte of current remote dma address RD */
+#define EN0_RSARLO	0x08U	/* Remote start address reg 0 */
+#define EN0_CRDAHI	0x09U	/* high byte, current remote dma address RD */
+#define EN0_RSARHI	0x09U	/* Remote start address reg 1 */
+#define EN0_RCNTLO	0x0aU	/* Remote byte count reg WR */
+#define EN0_RCNTHI	0x0bU	/* Remote byte count reg WR */
+#define EN0_RSR		0x0cU	/* rx status reg RD */
+#define EN0_RXCR	0x0cU	/* RX configuration reg WR */
+#define EN0_TXCR	0x0dU	/* TX configuration reg WR */
+#define EN0_COUNTER0	0x0dU	/* Rcv alignment error counter RD */
+#define EN0_DCFG	0x0eU	/* Data configuration reg WR */
+#define EN0_COUNTER1	0x0eU	/* Rcv CRC error counter RD */
+#define EN0_IMR		0x0fU	/* Interrupt mask reg WR */
+#define EN0_COUNTER2	0x0fU	/* Rcv missed frame error counter RD */
+
+/* Page 1 register offsets. */
+#define EN1_PHYS	0x01U	/* This board's physical enet addr RD WR */
+#define EN1_CURPAG	0x07U	/* Current memory page RD WR */
+#define EN1_MULT	0x08U	/* Multicast filter mask array (8 bytes) RDWR */
+
+/* Bits in received packet status byte and EN0_RSR. */
+#define ENRSR_RXOK	0x01U	/* Received a good packet */
+#define ENRSR_CRC	0x02U	/* CRC error */
+#define ENRSR_FAE	0x04U	/* frame alignment error */
+#define ENRSR_FO	0x08U	/* FIFO overrun */
+#define ENRSR_MPA	0x10U	/* missed pkt */
+#define ENRSR_PHY	0x20U	/* physical/multicase address */
+#define ENRSR_DIS	0x40U	/* receiver disable. set in monitor mode */
+#define ENRSR_DEF	0x80U	/* deferring */
+
+/* Bits in EN0_ISR - Interrupt status register. */
+#define ENISR_RX	0x01U	/* Receiver, no error */
+#define ENISR_TX	0x02U	/* Transmitter, no error */
+#define ENISR_RX_ERR	0x04U	/* Receiver, with error */
+#define ENISR_TX_ERR	0x08U	/* Transmitter, with error */
+#define ENISR_OVER	0x10U	/* Receiver overwrote the ring */
+#define ENISR_COUNTERS	0x20U	/* Counters need emptying */
+#define ENISR_RDC	0x40U	/* remote dma complete */
+#define ENISR_RESET	0x80U	/* Reset completed */
+#define ENISR_ALL	0x3fU	/* Interrupts we will enable */
+
+typedef struct {
+	unsigned char status;	/* status */
+	unsigned char next;	/* pointer to next packet */
+	unsigned short count;	/* header + packet length in bytes */
+} __attribute__((packed)) e8390_pkt_hdr;
+
+static struct wait_queue rxwait;
+static struct wait_queue txwait;
+
+static byte_t wd_inuse = 0U;
+static byte_t mac_addr[6U];
+
+static byte_t recv_buf[MAX_PACKET_ETH + sizeof(e8390_pkt_hdr)];
+static byte_t send_buf[MAX_PACKET_ETH];
+
+static unsigned char current_rx_page = WD_FIRST_RX_PG;
+
+static word_t wd_rx_stat(void);
+static word_t wd_tx_stat(void);
+
+/*
+ * Get MAC
+ */
+
+static void wd_get_hw_addr(word_t * data)
+{
+	unsigned u;
+
+	for (u = 0U; u < 6U; u++)
+		data[u] = INB(WD_PORT + 8U + u);
+}
+
+/*
+ * Reset
+ */
+
+static void wd_reset(void)
+{
+	OUTB(WD_RESET, WD_PORT);
+	OUTB(((WD_SHMEMSEG >> 9U) & 0x3fU) | WD_MEMENB, WD_PORT);
+}
+
+/*
+ * Init
+ */
+
+static void wd_init_8390(void)
+{
+	unsigned u;
+
+	OUTB(E8390_NODMA | E8390_PAGE0 | E8390_STOP,
+		WD_8390_PORT + E8390_CMD);
+	OUTB(0x48U, WD_8390_PORT + EN0_DCFG); /* 0x48 vs 0x49: 8 vs 16 bit */
+	/* Clear the remote byte count registers. */
+	OUTB(0x00U, WD_8390_PORT + EN0_RCNTLO);
+	OUTB(0x00U, WD_8390_PORT + EN0_RCNTHI);
+	/* Set to monitor and loopback mode. */
+	OUTB(E8390_RXOFF, WD_8390_PORT + EN0_RXCR);
+	OUTB(E8390_TXOFF, WD_8390_PORT + EN0_TXCR);
+	/* Set the transmit page and receive ring. */
+	OUTB(WD_FIRST_TX_PG, WD_8390_PORT + EN0_TPSR);
+	OUTB(WD_FIRST_RX_PG, WD_8390_PORT + EN0_STARTPG);
+	OUTB(WD_STOP_PG - 1U, WD_8390_PORT + EN0_BOUNDARY);
+	OUTB(WD_STOP_PG, WD_8390_PORT + EN0_STOPPG);
+	/* Clear the pending interrupts and mask. */
+	OUTB(0xffU, WD_8390_PORT + EN0_ISR);
+	OUTB(0x00U, WD_8390_PORT + EN0_IMR);
+	/* Copy the station address into the DS8390 registers. */
+	clr_irq();
+	OUTB(E8390_NODMA | E8390_PAGE1 | E8390_STOP,
+		WD_8390_PORT + E8390_CMD);
+	for(u = 0U; u < 6U; u++)
+		OUTB(mac_addr[u], WD_8390_PORT + EN1_PHYS + u);
+	OUTB(WD_FIRST_RX_PG, WD_8390_PORT + EN1_CURPAG);
+	OUTB(E8390_NODMA | E8390_PAGE0 | E8390_STOP,
+		WD_8390_PORT + E8390_CMD);
+	set_irq();
+}
+
+static void wd_init(void)
+{
+	wd_init_8390();
+	OUTB(0xffU, WD_8390_PORT + EN0_ISR);
+	OUTB(ENISR_ALL, WD_8390_PORT + EN0_IMR);
+	OUTB(E8390_NODMA | E8390_PAGE0 | E8390_START,
+		WD_8390_PORT + E8390_CMD);
+	OUTB(E8390_TXCONFIG, WD_8390_PORT + EN0_TXCR); /* xmit on */
+	/* Only after the NIC is started: */
+	OUTB(E8390_RXCONFIG, WD_8390_PORT + EN0_RXCR); /* rx on */
+}
+
+/*
+ * Start
+ */
+
+static void wd_start(void)
+{
+	OUTB(((WD_SHMEMSEG >> 9U) & 0x3fU) | WD_MEMENB, WD_PORT);
+}
+
+/*
+ * Stop
+ */
+
+static void wd_stop(void)
+{
+	OUTB(((WD_SHMEMSEG >> 9U) & 0x3fU) & ~WD_MEMENB, WD_PORT);
+}
+
+/*
+ * Termination
+ */
+
+static void wd_term(void)
+{
+	wd_init_8390(); /* SIC! */
+}
+
+/*
+ * Get packet
+ */
+
+static int wd_pack_get(byte_t * pack)
+{
+	const e8390_pkt_hdr *rxhdr = (e8390_pkt_hdr *)pack;
+	byte_t *hdr_start;
+	unsigned char this_frame;
+	int res = 0;
+
+	clr_irq();
+	do {
+		/* Remove one frame from the ring. */
+		/* Boundary is always a page behind. */
+		this_frame = INB(WD_8390_PORT + EN0_BOUNDARY) + 1U;
+		if (this_frame >= WD_STOP_PG)
+			this_frame = WD_FIRST_RX_PG;
+		if (this_frame != current_rx_page)
+			debug_eth("Mismatched read page pointers %2x vs %2x.\n",
+				this_frame, current_rx_page);
+		hdr_start = (byte_t *)((this_frame - WD_START_PG) << 8U);
+		fmemcpyb(pack, kernel_ds, hdr_start, WD_SHMEMSEG,
+			ETH_PAGE_SIZE);
+		if ((rxhdr->count < 64U) ||
+		    (rxhdr->count > (MAX_PACKET_ETH + sizeof(e8390_pkt_hdr)))) {
+			debug_eth("eth: bogus packet size: %d, "
+				"status = %#2x nxpg = %#2x.\n",
+				rxhdr->count, rxhdr->status, rxhdr->next);
+			res = -EIO;
+			break;
+		}
+		current_rx_page = rxhdr->next;
+		if ((rxhdr->status & 0x0fU) != ENRSR_RXOK) {
+			debug_eth("eth: bogus packet: "
+				"status = %#2x nxpg = %#2x size = %d\n",
+				rxhdr->status, rxhdr->next, rxhdr->count);
+			res = -EIO;
+		} else if (rxhdr->count > ETH_PAGE_SIZE) {
+			fmemcpyb(pack + ETH_PAGE_SIZE, kernel_ds,
+				hdr_start + ETH_PAGE_SIZE, WD_SHMEMSEG,
+				rxhdr->count - ETH_PAGE_SIZE);
+		}
+		OUTB(current_rx_page - 1U, WD_8390_PORT + EN0_BOUNDARY);
+	} while (0);
+	set_irq();
+	return res;
+}
+
+static size_t wd_read(struct inode * inode, struct file * filp,
+	char * data, size_t len)
+{
+	size_t size;
+	size_t res = 0U;
+
+	do {
+		prepare_to_wait_interruptible(&rxwait);
+		if (wd_rx_stat() != WD_STAT_RX) {
+			if (filp->f_flags & O_NONBLOCK) {
+				res = -EAGAIN;
+				break;
+			}
+			do_wait();
+			if (current->signal) {
+				res = -EINTR;
+				break;
+			}
+		}
+		if (wd_pack_get(recv_buf)) {
+			res = -EIO;
+			break;
+		}
+		size = ((e8390_pkt_hdr *)(recv_buf))->count;
+		if (len > size) len = size;
+		memcpy_tofs(data, recv_buf + sizeof(e8390_pkt_hdr), len);
+		res = len;
+	} while (0);
+	finish_wait(&rxwait);
+	return res;
+}
+
+/*
+ * Pass packet to driver for send
+ */
+
+static int wd_pack_put(byte_t * pack, word_t len)
+{
+	int res = 0;
+
+	clr_irq();
+	do {
+		fmemcpyb((byte_t *)((WD_FIRST_TX_PG - WD_START_PG) << 8U),
+			WD_SHMEMSEG, pack, kernel_ds, len);
+		OUTB(E8390_NODMA | E8390_PAGE0, WD_8390_PORT + E8390_CMD);
+		if (INB(WD_8390_PORT + E8390_CMD) & E8390_TRANS) {
+			printk("eth: attempted send with the tr busy.\n");
+			res = -EIO;
+			break;
+		}
+		OUTB(len & 0xffU, WD_8390_PORT + EN0_TCNTLO);
+		OUTB(len >> 8U, WD_8390_PORT + EN0_TCNTHI);
+		OUTB(WD_FIRST_TX_PG, WD_8390_PORT + EN0_TPSR);
+		OUTB(E8390_NODMA | E8390_TRANS | E8390_START,
+			WD_8390_PORT + E8390_CMD);
+	} while (0);
+	set_irq();
+	return res;
+}
+
+static size_t wd_write(struct inode * inode, struct file * file,
+	char * data, size_t len)
+{
+	size_t res = 0U;
+
+	do {
+		prepare_to_wait_interruptible(&txwait);
+		if (wd_tx_stat() != WD_STAT_TX) {
+			if (file->f_flags & O_NONBLOCK) {
+				res = -EAGAIN;
+				break;
+			}
+			do_wait();
+			if(current->signal) {
+				res = -EINTR;
+				break;
+			}
+		}
+		if (len > MAX_PACKET_ETH)
+			len = MAX_PACKET_ETH;
+		memcpy_fromfs(send_buf, data, len);
+		res = len;
+		if (len < 64U) len = 64U;  /* issue #133 */
+		if (wd_pack_put(send_buf, len)) {
+			res = -EIO;
+			break;
+		}
+	} while (0);
+	finish_wait(&txwait);
+	return res;
+}
+
+/*
+ * Clear overflow
+ */
+
+static void wd_clr_oflow(void)
+{
+	printk("eth: overflow.\n");
+	wd_stop();
+	wd_term();
+	current_rx_page = WD_FIRST_RX_PG;
+	wd_reset();
+	wd_init();
+	if (INB(WD_PORT + 14U) & 0x20U) /* enable IRQ on softcfg card */
+		OUTB(INB(WD_PORT + 4U) | 0x80U, WD_PORT + 4U);
+	wd_start();
+}
+
+/*
+ * Test for readiness
+ */
+
+static word_t wd_rx_stat(void)
+{
+	unsigned char rxing_page;
+
+	clr_irq();
+	/* Get the rx page (incoming packet pointer). */
+	OUTB(E8390_NODMA | E8390_PAGE1, WD_8390_PORT + E8390_CMD);
+	rxing_page = INB(WD_8390_PORT + EN1_CURPAG);
+	OUTB(E8390_NODMA | E8390_PAGE0, WD_8390_PORT + E8390_CMD);
+	set_irq();
+	return (current_rx_page == rxing_page) ? 0U : WD_STAT_RX;
+}
+
+static word_t wd_tx_stat(void)
+{
+	return (INB(WD_8390_PORT + E8390_CMD) & E8390_TRANS) ? 0U :
+		WD_STAT_TX;
+}
+
+static int wd_select(struct inode * inode, struct file * filp, int sel_type)
+{
+	int res = 0;
+
+	switch (sel_type) {
+		case SEL_OUT:
+			if (wd_tx_stat() != WD_STAT_TX) {
+				select_wait(&txwait);
+				break;
+			}
+			res = 1;
+			break;
+		case SEL_IN:
+			if (wd_rx_stat() != WD_STAT_RX) {
+				select_wait(&rxwait);
+				break;
+			}
+			res = 1;
+			break;
+		default:
+			res = -EINVAL;
+	}
+	return res;
+}
+
+/*
+ * I/O control
+ */
+
+static int wd_ioctl(struct inode * inode, struct file * file,
+	unsigned int cmd, unsigned int arg)
+{
+	int err = 0;
+
+	switch (cmd) {
+	case IOCTL_ETH_ADDR_GET:
+		memcpy_tofs((char *)arg, mac_addr, 6U);
+		break;
+	case IOCTL_ETH_ADDR_SET:
+		err = -ENOSYS;
+		break;
+	case IOCTL_ETH_HWADDR_GET:
+		/* Get the hardware address of the NIC,	which may be different
+		 * from the currently programmed address. Be careful with this,
+		 * it may interrupt ongoing send/receives.
+		 * arg must be a 32 bytes array.
+		 */
+		wd_get_hw_addr((word_t *)arg);
+		break;
+	case IOCTL_ETH_GETSTAT:
+		/* Get statistics from the NIC hardware (error counts etc.)
+		 * arg is byte[3].
+		 */
+		err = -ENOSYS;
+		break;
+	case IOCTL_ETH_OFWSKIP_SET:
+		/* Set the number of packets to skip @ ring buffer overflow. */
+		err = -ENOSYS;
+		break;
+	case IOCTL_ETH_OFWSKIP_GET:
+		/* Get the current overflow skip counter. */
+		err = -ENOSYS;
+		break;
+	default:
+		err = -EINVAL;
+	}
+	return err;
+}
+
+/*
+ * Device open
+ */
+
+static int wd_open(struct inode * inode, struct file * file)
+{
+	int err = 0;
+
+	do {
+		if (wd_inuse) {
+			err = -EBUSY;
+			break;
+		}
+		wd_reset();
+		wd_init();
+		if (INB(WD_PORT + 14U) & 0x20U) /* enable IRQ on softcfg card */
+			OUTB(INB(WD_PORT + 4U) | 0x80U, WD_PORT + 4U);
+		wd_start();
+		wd_inuse = 1U;
+	} while (0);
+	return err;
+}
+
+/*
+ * Release (close) device
+ */
+
+static void wd_release(struct inode * inode, struct file * file)
+{
+	wd_stop();
+	wd_term();
+	wd_inuse = 0U;
+}
+
+/*
+ * Ethernet operations
+ */
+
+static struct file_operations wd_fops =
+{
+	NULL,         /* lseek */
+	wd_read,
+	wd_write,
+	NULL,         /* readdir */
+	wd_select,
+	wd_ioctl,
+	wd_open,
+	wd_release,
+#ifdef BLOAT_FS
+	NULL,         /* fsync */
+	NULL,         /* check_media_type */
+	NULL,         /* revalidate */
+#endif
+};
+
+/*
+ * Interrupt handler
+ */
+
+static word_t wd_int_stat(void)
+{
+	byte_t interrupts;
+	word_t retval = 0U;
+
+	/* Change to page 0 and read the intr status reg. */
+	OUTB(E8390_NODMA | E8390_PAGE0, WD_8390_PORT + E8390_CMD);
+	interrupts = INB(WD_8390_PORT + EN0_ISR);
+	if (interrupts & ENISR_RX)
+		retval |= WD_STAT_RX;
+	if (interrupts & ENISR_TX)
+		retval |= WD_STAT_TX;
+	if (interrupts & ENISR_OVER)
+		retval |= WD_STAT_OF;
+	else if (interrupts)
+		OUTB(0xffU, WD_8390_PORT + EN0_ISR); /* ack ALL */
+	OUTB(E8390_NODMA | E8390_PAGE0 | E8390_START,
+		WD_8390_PORT + E8390_CMD);
+	return retval;
+}
+
+static void wd_int(int irq, struct pt_regs * regs, void * dev_id)
+{
+	word_t stat;
+
+	stat = wd_int_stat();
+	debug_eth("/%02X", stat & 0xffU);
+	if (stat & WD_STAT_OF)
+		wd_clr_oflow();
+	if (stat & WD_STAT_RX)
+		wake_up(&rxwait);
+	if (stat & WD_STAT_TX)
+		wake_up(&txwait);
+}
+
+/*
+ * Ethernet main initialization (during boot)
+ */
+
+void eth_drv_init()
+{
+	int err;
+	unsigned u;
+	word_t hw_addr[16U];
+
+	do {
+		err = request_irq(WD_IRQ, wd_int, NULL);
+		if (err) {
+			printk("eth: WD IRQ %d request error: %i\n",
+				WD_IRQ, err);
+			break;
+		}
+		err = register_chrdev(ETH_MAJOR, "eth", &wd_fops);
+		if (err) {
+			printk("eth: register error: %i\n", err);
+			break;
+		}
+		wd_get_hw_addr(hw_addr);
+		for (u = 0U; u < 6U; u++)
+			mac_addr[u] = (hw_addr[u] & 0xffU);
+		printk ("eth: SMC/WD8003 at 0x%x, irq %d, MAC %02X",
+			WD_PORT, WD_IRQ, mac_addr[0]);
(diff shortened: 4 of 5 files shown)


ghaerr:
Hello @pawosm-arm,

I have done a quick review and found a few items that might want changing. Let me know if you'd rather consider adding them after this commit (or not at all), otherwise feel free to push changes to the same branch on this PR.

In wd.c:
- Line 567: Do you really mean "else if"? I think "if" is what is wanted.
- Line 140: Change "4" to "sizeof(e8390_pkt_hdr)".
- Line 553: With regards to the wd_int_stat interrupt handler routine, are you totally OK with its register writes happening at absolutely any time, including when the kernel may be executing within the wd_pack_put or wd_pack_get routines, possibly also in the middle of reading or writing another chip register? If not, then you will need to protect critical sections in the packet routines with clr_irq() and set_irq() around the chip register manipulation. This can be tested by using @Mellvik-style heavy incoming bombardment of packets to the driver while running a network connection to outside.
- Line 596: The 32-byte stack variable is OK here, since at init time we're running on the task 0 stack (512 bytes), and nothing else is running. FYI, this would be too large a stack variable if in other kernel routines.
- I see you're using mostly outb_p, but sometimes outb. Do you mean for all of them to be outb_p? The outb_p version adds another "in $0x80,%ax" instruction to slow I/O. I'm interested in thoughts on this, I think the pause was added in Linux 32-bit systems, because the machines had very fast processors (compared to an 8086). I'm not sure whether ELKS machines need this. For the serial driver, I've changed all calls to a macro "OUTB" which can be defined either way, with the default being outb (no pause). I'm not sure what the point of inb_p is, given that the bus has (presumably) already settled by the time the "in" instruction has executed... what would the point of an additional pause be on input?
- Line 273: Using `fmemcpyb` is a great idea when lots of data needs to be transferred, but I'm thinking that for just inspecting a 4-byte header, a far pointer to the e8390_pkt_hdr structure will be much quicker and less code than copying:
```
struct e8390_pkt_hdr _far *hdr = (struct e8390_pkt_hdr _far *)_MK_FP(WD_SHMEMSEG, hdr_start);
if (hdr->count < 64U) ...
if (hdr->status & 0x0fU) ...
   ...
```

All-in-all, fantastic work.

Thank you!

pawosm-arm:
Hi @ghaerr 

It's after midnight here, so I'll probably fix those things tomorrow.

> Line 567: Do you really mean "else if"? I think "if" is what is wanted.

I meant exactly what is there: I'm not unmasking overflow, as further on, the device will be reset anyway

> Line 140: Change "4" to "sizeof(e8390_pkt_hdr)".

Good point.

> Line 553: With regards to the wd_int_stat interrupt handler routine, are you totally OK with its register writes happening at absolutely any time, including when the kernel may be executing within the wd_pack_put or wd_pack_get routines, possibly also in the middle of reading or writing another chip register? If not, then you will need to protect critical sections in the packet routines with clr_irq() and set_irq() around the chip register manipulation.

It may be needed. Original ne2k code does it in DMA routines, likewise, this driver should do that too.

> Line 596: The 32-byte stack variable is OK here, since at init time we're running on the task 0 stack (512 bytes), and nothing else is running. FYI, this would be too large a stack variable if in other kernel routines.

It's no secret I based my code on ELKS ne2k driver. Similar routine in ne2k.c creates even bigger stack variables. Other bulkies are defined as global variables.

> I see you're using mostly outb_p, but sometimes outb. Do you mean for all of them to be outb_p?

This is exactly how Linux driver has it. There was some explanation why they didn't use outb_p everywhere, can't find it right now.

> Line 273: Using fmemcpyb is a great idea when lots of data needs to be transferred, but I'm thinking that for just inspecting a 4-byte header, a far pointer to the e8390_pkt_hdr structure will be much quicker and less code than copying

I've used `fmemcpyb` following your recommendation on copying data to/from upper addresses area. I'll test your proposed change tomorrow.

Cheers

ghaerr:
Hello @pawosm-arm,

>  I meant exactly what is there: I'm not unmasking overflow, as further on, the device will be reset anyway

Ok, I get it now.

Regarding disabling and reenabling interrupts: using save_flags/restore_flags is only required in very special cases within the kernel, where interrupts need to be disabled, and then restored to their previous state. Since this will never be an issue in a driver, (since the 8259 interrupt controller handles interrupt hierarchy and all ELKS interrupt handlers execute with interrupts enabled anyways), just use set_irq() instead to reenable interrupts.

Mellvik:
Congratulations are in order @pawosm-arm, this is really encouraging. It seems to me you've done a great job here.

Like @ghaerr, I have a few questions/comments, not the least about testing and timing. 
@ghaerr's point about outb_p/inb_p is very valid. The extra read/write was introduced for 386 and faster machines to give the old interfaces time to react. In your case - with an 8-bit bus and a slow XT class, you would benefit significantly from removing the _p variant.

I'm curious as to the round trip times reported when you're pinging the system - plain ping (minimal packet, 1 per sec) and more heavy testing. My favourite is 
ping -i 0.5 -s 1400 10.0.2.15
and even shorter intervals.

How does telnet and urlget work out for you?

As to another comment from @ghaerr, about using fmemcpyb for 4 bytes, I suggest you read all 256 bytes instead, like the ne2k driver does. This will speed up processing significantly since most packets are less that 256 bytes, and you avoid a 2nd read (skipping the fmemcpyb in 327).

- As I think I have mentioned before, the packet validity tests (275-288) have no value since you have programmed the NIC not to accept such packets.

- I've learned the hard way that touching the CURRENT register should be avoided when possible, not the least because it requires a switch to register page 2. The test in 270 is not required, since the presence of data is tested before pkt_get is called. Thus you can eliminate the access to the CURRENT reg in 260.

- The test in 291 is buggy and my recommendations is to eliminate. While it's nice to test and accommodate for clones with bugs, the likelihood that they still exist AND should end up in a ELKS machine is minute :-). The bug is in 294, where current_rx_page = WD_FIRST_RX_PG. 
IN the next line, the BOUNDARY reg is set to current_rx_page-1, so BOUNDARY is now outside its range.

- IN wd_reset() you may need to wait for the NIC to become ready even though this is a slow machine. Polling the high bit of the Interrupt Status Reg is a safe way to do this.

- The clear overflow routine resets the NIC completely and starts from scratch. This is fine if such overflow is something that rarely happens. I practice, and again since the machine is so slow (my 386 is slow too, btw), you will have overflows all that time, and it's important that the overflow routine is 'nice' - both in terms of timing and to the data. You can have a look at the ne2k code if you're interested.

- Not that it makes much difference, but in 596, hw_addr is allocated 16 words, but only uses 6.

Again, great work. I'm very much looking forward to your test results.


—Mellvik

pawosm-arm:
> As I think I have mentioned before, the packet validity tests (275-288) have no value since you have programmed the NIC not to accept such packets.

Argh, this test for "bogus packet size" was my first line of defense when shared memory address was messed up...

Mellvik:
> 23. sep. 2020 kl. 11:32 skrev pawosm-arm <notifications@github.com>:
> 
> 
> As I think I have mentioned before, the packet validity tests (275-288) have no value since you have programmed the NIC not to accept such packets.
> 
> Argh, this test for "bogus packet size" was my first line of defense when shared memory address was messed up...
> 
> 
It's apparent that you've been around the block a few times @pawosm-arm - :-)  this is possibly known as something like 'the random printk blessing'.

Again - great piece of work. It's inspiring to see results like this.

—M

pawosm-arm:
> - IN wd_reset() you may need to wait for the NIC to become ready even though this is a slow machine. Polling the high bit of the Interrupt Status Reg is a safe way to do this.

As I mentioned it before, Linux driver does not do that and doing so made `ktcp` hang on my other XT machine (8088 Turbo-XT). Unfortunately, I'm not having this machine here, I'd be able to test it on it when I'll be visiting the continent again this December, so I guess for now this piece of code stays as it is...

pawosm-arm:
The second take is waiting.

Ping measurements after those changes:

```
$ ping 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 56(84) bytes of data.
64 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=8.33 ms
64 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=8.35 ms
64 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=8.09 ms
64 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=8.09 ms
64 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=8.35 ms
^C
--- 192.168.1.44 ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 4005ms
rtt min/avg/max/mdev = 8.093/8.241/8.348/0.121 ms

$ ping -i 0.5 -s 1400 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=53.9 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=54.0 ms
1408 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=54.0 ms
1408 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=54.0 ms
1408 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=54.0 ms
1408 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=53.9 ms
^C
--- 192.168.1.44 ping statistics ---
8 packets transmitted, 6 received, 25% packet loss, time 3505ms
rtt min/avg/max/mdev = 53.859/53.946/53.995/0.041 ms

$ arping 192.168.1.44
ARPING 192.168.1.44 from 192.168.1.3 enp2s0
Unicast reply from 192.168.1.44 [00:00:C0:16:8F:43]  7.024ms
Unicast reply from 192.168.1.44 [00:00:C0:16:8F:43]  7.023ms
Unicast reply from 192.168.1.44 [00:00:C0:16:8F:43]  7.007ms
Unicast reply from 192.168.1.44 [00:00:C0:16:8F:43]  6.973ms
Unicast reply from 192.168.1.44 [00:00:C0:16:8F:43]  6.975ms
^CSent 5 probes (1 broadcast(s))
Received 5 response(s)
$
```

Mellvik:
Very encouraging indeed, @pawosm-arm!

Are these number with or without extra debug messages on the console?

Also, I recommend running each of them for a couple of minutes in order to push the system a bit. And do the while doing some disk i/o and other i/o on the system, such as logging in (lots of i/o), 
hd /bin/vi  (takes a while)
dd if=/dev/fd1 of=/dev/null
etc.
all while the big ping is running, possibly shortened to -i 0.3 

The purpose is to trigger overruns, which the dd above should fix.

—Mellvik

pawosm-arm:
> Are these number with or without extra debug messages on the console?

These are without debug messages. I've turned them off for the time of upstreaming.

pawosm-arm:
The `hd` test started in a telnet session itself did not cause overruns, the pings were returning at times between 140 and 300ms. Yet when I started `hd` in a telnet session and in the ELKS console simultaneously, eth overruns on the kernel side with retrans retries on the ktcp side started to appear on the ELKS console screen. Yet pings were still returning, at times between 300ms and 800ms.

The `dd` test started in telnet froze the telnet session for couple of seconds and pings were not returning at the same time. Everything returned to normal though, and during copying from floppy, pings were returning at times between 1000ms and 2000ms. After test was over, pings were returning as usual at 54ms.

I'm observing some strange issue though. While being on telnet session (Linux to ELKS), I'm repeatedly trying to run `urlget` to get a web page from external host. After nth repetition, `urlget` process got stuck (pings are still returning from ELKS at usual times) blocking entire telnet session. There's a message on ELKS console because of that: `SYN sent, wrong ACK`. Does not occur when I'm repeatedly running `urlget` from ELKS console... Not sure if it is the driver or the ktcp daemon issue.

ghaerr:
> There's a message on ELKS console because of that: SYN sent, wrong ACK.

Is this the "SYN sent, wrong ACK (listen port not expired)"?

If so, that's a known close bug in ktcp. What is happening is that one side of the TCP connection closed the connection, and ktcp is waiting in a "TCP CLOSEWAIT state" waiting for the other side to close, or for a multi-second timeout. The bug prohibits other connections to the same address while the timeout (1-4 seconds) is occurring. I need to raise the priority for the fix on this one, it seems. It has nothing to do with the network driver. This can also be duplicated by running telnet localhost, logging in, then ^D exiting and trying to connect again, which will fail with the same error.

ghaerr:
> As to another comment from @ghaerr, about using fmemcpyb for 4 bytes, I suggest you read all 256 bytes instead, like the ne2k driver does. This will speed up processing significantly since most packets are less that 256 bytes, and you avoid a 2nd read (skipping the fmemcpyb in 327).

Reading @Mellvik's comment, I tend to agree, why not just read a portion of the packet, which will always include the header? This negates the need for the struct _far \* pointer as well as a separate fmemcpyb. However, I suggest the idea of only reading 64+4 bytes, (the minimum), rather than 256 (which wastes time on small packets), and then read the rest of the packet once the size is known?

pawosm-arm:
> Is this the "SYN sent, wrong ACK (listen port not expired)"?

It was exactly that.

Mellvik:
@ghaerr & @pawosm-arm,

the idea with reading the full page (252 bytes of packet data) is that you get most ELKS packets with one single read. For the time being. The only exception is incoming ftp/http traffic which will require a 2nd page to be transferred (gross packet size 295 bytes). Outgoing file transfers currently used between 140 and 200 bytes per packet. So IMHO reading 64+4 rather than 4 doesn't really get us much.

For now, the only situation in which large packets are transferred, is the ping test with the -s option.

—Mellvik

Mellvik:
@pawosm-arm,

another piece of experience you may find useful for driver testing:

Because of the bugs that still haunt the TCP stack, stress testing the driver is most effectively done avoiding TCP, using ping (ICMP) with varying frequencies and packet sizes along with floppy activity via the console. Console output is also good (like the 'hd' you already tried - hs /dev/mem is another of my favourites).

Long running TCP connections may work if nothing else is going on on the system, such as exporting the fill contents of the hard disk via httpd (outgoing), or using urlget to import large files (to /dev/null).

my 2cents.

—Mellvik

pawosm-arm:
> the idea with reading the full page (252 bytes of packet data)

Is it 252 or 256? All of the 8 bit shifts in wd.c suggest it's 256 bytes.

Mellvik:
The page size is 256, the packet data in the first page is 256-4 (the header).

-M

pawosm-arm:
@Mellvik I guess it's what current code does.

Mellvik:
If you mean the ne2k driver, your right.

your driver reads 4 bytes, then whatever the length is in a 2nd operation, which is fine (and the normal you'll find in most drivers), but not optimal. For old clunkers, 'optimal' has a special significance. :-)

—M

ghaerr:
> the idea with reading the full page (252 bytes of packet data) is that you get most ELKS packets with one single read. For the time being. The only exception is incoming ftp/http traffic which will require a 2nd page to be transferred (gross packet size 295 bytes). 

I'll leave this to both of you, but I don't understand yet: the ethernet driver only returns one packet at a time (to the kernel). So does your driver buffer that data? I thought it didn't. So exactly how does reading more than one packet at a time from the NIC work?

> Outgoing file transfers currently used between 140 and 200 bytes per packet. So IMHO reading 64+4 rather than 4 doesn't really get us much.

I was talking about the driver read handler, the write handler doesn't read any headers, correct?

pawosm-arm:
> @Mellvik: your driver reads 4 bytes, then whatever the length is in a 2nd operation

Nah, I've changed that on the last force push (see history of this pull request). Now it always reads 256 bytes which includes the header, and if the header says it's more than 256 bytes waiting, then I call fmemcpyb again, with known size this time.

> @ghaerr: I was talking about the driver read handler, the write handler doesn't read any headers, correct?

Yup, the header is only in rx packets, hence rx buffer is allocated 4 bytes larger than tx buffer.

Mellvik:
I have a few misunderstandings to clear out - sorry for the confusion!!

> 23. sep. 2020 kl. 19:04 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> the idea with reading the full page (252 bytes of packet data) is that you get most ELKS packets with one single read. For the time being. The only exception is incoming ftp/http traffic which will require a 2nd page to be transferred (gross packet size 295 bytes).
> 
> I'll leave this to both of you, but I don't understand yet: the ethernet driver only returns one packet at a time. So does your driver buffer that data? I thought it didn't. So exactly how does reading more than one packet at a time from the NIC work?
> 
The driver does definitely not buffer anything, this is the details of the transfer from the ring buffer to whatever buffer is provided by the packet_get call. The transfer from the NIC ring buffer is normally two operations, the first to transfer 4 bytes in order to find out how much data to read (get the length of the packet), the 2nd to read that data. The ring buffer is divided into 256 byte pages, and a packet always starts on a page boundary. So making the first read a full page saves a full read operation off the packet is less than 253 bytes ...
> Outgoing file transfers currently used between 140 and 200 bytes per packet. So IMHO reading 64+4 rather than 4 doesn't really get us much.
> 
> I was talking about the driver read handler, the write handler doesn't read any headers, correct?
> 
> 
You're right, this has nothing to do with the read operation, again sorry for creating confusion. 

—M

Mellvik:
> 
> your driver reads 4 bytes, then whatever the length is in a 2nd operation
> 
> Nah, I've changed that on the last force push (see history of this pull request). Now it always read 256 bytes which includes the header, and if the header says it's more than 256 bytes waiting, then I call fmemcpyb again, with known size this time.
> 
> 
Sorry; i didn't see that, should be good then!!!

—M

ghaerr:
@pawosm-arm,

I think I understand now about the packet buffer in the NIC. However, in your most recent code:
```
		} else if (rxhdr->count > ETH_PAGE_SIZE) {
 			fmemcpyb(pack, kernel_ds, hdr_start, WD_SHMEMSEG,
 				rxhdr->count);
 		}
```
Instead of copying the first 256 (ETH_PAGE_SIZE) again, wouldn't you want something like:

```
		} else if (rxhdr->count > ETH_PAGE_SIZE) {
 			fmemcpyb(pack + ETH_PAGE_SIZE, kernel_ds, hdr_start + ETH_PAGE_SIZE, WD_SHMEMSEG,
 				rxhdr->count - ETH_PAGE_SIZE);
 		}
```

pawosm-arm:
> Instead of copying the first 256 (ETH_PAGE_SIZE) again, wouldn't you want something like:Instead of copying the first 256 (ETH_PAGE_SIZE) again, wouldn't you want something like:

Ahh, yes it was meant to be like that. Will change it soon.

pawosm-arm:
@ghaerr proposed changes were force-pushed into this pull request. It seems beneficial, `ping -i 0.5 -s 1400` dropped from 54ms to 53ms.

ghaerr:
> It seems beneficial, ping -i 0.5 -s 1400 dropped from 54ms to 53ms.

Well, if not copying 256 bytes saved 1ms, it would be interesting to see what not copying 192 bytes would be, since pings only result in 64 byte packets (right? this should be tested). If ETH_PAGE_SIZE were changed to something like "ETH_READ_AHEAD" and set to 64, what is the result? That's what I was trying to get to in my previous post/question. In my limited network testing, I found lots of (TCP) packets less than 128 bytes, and I think I computed the average once on an interactive session to be somewhere in the 78-100 byte region. No sense copying more data than necessary, as the ELKS kernel copies received data again, and again, and again already, unfortunately.

Other thoughts are that it would be nice to keep track of the min and max packet sizes with various operations, that could be displayed in some way. Rather than writing all sorts of special user-to-kernel programs and ioctls, I have been thinking of implementing a mechanism such that a registered kernel routine would be called when the SCROLL_LOCK key is toggled, for instance. That could allow for either turning on and off debug, or momentary one-time debug messages, that could be displayed right while you're in the middle of testing.

Mellvik:
> ﻿
> It seems beneficial, ping -i 0.5 -s 1400 dropped from 54ms to 53ms.
> 
If this is a single measure, it is below any reasonable statistical noise limit, isn't it? 
> Well, if not copying 256 bytes saved 1ms, it would be interesting to see what not copying 192 bytes would be, since pings only result in 64 byte packets (right? this should be tested). If ETH_PAGE_SIZE were changed to something like "ETH_READ_AHEAD" and set to 64, what is the result?
> 
We may be heading somewhere with this, but we're gearing for real use, not ping, right?
> That's what I was trying to get to in my previous post/question. In my limited network testing, I found lots of (TCP) packets less than 128 bytes, and I think I computed the average once on an interactive session to be somewhere in the 78-100 byte region. No sense copying more data than necessary, as the ELKS kernel copies received data again, and again, and again already, unfortunately.
> 
My idea with retrieving a page instead of 4 bytes, was to eliminate the setup time for "remote dma transfer", about (guessing) 40 instructions. If the farmemcopy routine has no significant setup penalty, this does not make much sense.
> Other thoughts are that it would be nice to keep track of the min and max packet sizes with various operations, that could be displayed in some way. Rather than writing all sorts of special user-to-kernel programs and ioctls, I have been thinking of implementing a mechanism such that a registered kernel routine would be called when the SCROLL_LOCK key is toggled, for instance. That could allow for either turning on and off debug, or momentary one-time debug messages, that could be displayed right while you're in the middle of testing.
> 
Interesting idea, let's check it out!
> 
--Mellvik

pawosm-arm:
> Rather than writing all sorts of special user-to-kernel programs and ioctls, I have been thinking of implementing a mechanism such that a registered kernel routine would be called when the SCROLL_LOCK key is toggled, for instance. That could allow for either turning on and off debug, or momentary one-time debug messages, that could be displayed right while you're in the middle of testing.
> Interesting idea, let's check it out!

Guys, I'm sorry to say that but although I can test your solutions on my XTs, my coding involvement cannot be as extensive as I'd wish it to be. I enjoy those hacking days episodes though.

ghaerr:
> Guys, I'm sorry to say that but although I can test your solutions on my XTs, my coding involvement cannot be as extensive as I'd wish it to be. I enjoy those hacking days episodes though.

That's ok - let me know when you're done and would like this committed, and thank you!!

ghaerr:
@Mellvik,

> My idea with retrieving a page instead of 4 bytes, was to eliminate the setup time for "remote dma transfer", about (guessing) 40 instructions. If the farmemcopy routine has no significant setup penalty, this does not make much sense.

I will submit a way for this to be objectively measured via instruction counting, if we're talking about the same thing.
With your NIC, you only have DMA setup overhead, but with the WD, there's no DMA, plus it takes time for the CPU to copy each byte within fmemcpy. I figured DMA is faster than manually with the CPU. Nonetheless, my measurement proposal would be to measure the number of instructions executed to setup fmemcpy, including its entry overhead, then measure the amount of time taken to move 256 bytes. Compare that with the same entry entry overhead twice, and say, the time for 80-100 byte copies. As I think more about it, a single "word_t far \*" pointer could quickly grab the length byte and then copy the required exact packet length in a single fmemcpyb, which would likely be way faster and less code than either method.

In the meantime, I have some ideas for implementing the keystroke-based debug on/off method, which I think you'll find useful when debugging either ktcp TCP retransmits and getting NIC card statistics.

pawosm-arm:
@ghaerr the driver as it is now seems to work well with `ktcp`, although comparing to ne2k driver it lacks some things. Overflow should be handled less brutal way and consequently, `IOCTL_ETH_OFWSKIP_SET` ioctl should be implemented (and how to test it? does `ktcp` invoke it at all?). The `IOCTL_ETH_ADDR_SET` could be probably easily implemented by `outb()`'s yet I don't even know how to test it with ELKS (does it have something `ifconfig`-alike?). Those missing features could be implemented by means of later updates of this pull request or as separate PR's later on. Or it could be a part of a larger job of having lib8390 introduced into ELKS.

ghaerr:
> the driver as it is now seems to work well with ktcp, 

I would say, lets commit this now then, especially if you may have problems getting further coding time in the near future.

> although comparing to ne2k driver it lacks some things. Overflow should be handled less brutal way 

That's Ok - we don't need to have a perfect driver in the first commit! It can always be improved later, when you get time.


> and consequently, IOCTL_ETH_OFWSKIP_SET ioctl should be implemented (and how to test it? does ktcp invoke it at all?). The IOCTL_ETH_ADDR_SET could be probably easily implemented by inb()'s yet I don't even know how to test it with ELKS (does it have something ifconfig-alike?). 

None of the IOCTLs other than IOCTL_ETH_ADDR_GET or IOCTL_ETH_HWADDR_GET are implemented in ktcp, nor will they likely be. They may be implemented in additional user commands or put into netstat or arp. I have actually recommended that all unused IOCTLs be commented out in both drivers using "#if later" until the time that they are actually implemented, because there isn't room in ELKS for non-implemented cruft.


> Those missing features could be implemented by means of later updates of this pull request or as separate PR's later on. Or it could be a part of a larger job of having lib8390 introduced into ELKS.

Yes. I suggest separate PRs or a library, or both, and committing this now. We can always prepare additional updates.

pawosm-arm:
@ghaerr ok, feel free to proceed.

Mellvik:
Thanks for the ideas, @ghaerr.
It seems to me I've stirred up a lot more dust here than the issue deserves. When recommending @pawosm-arm use my mechanism for optimizing these transfers, I did not consider the difference in data transfer methods.
As this discussion has proceeded I've come to the conclusion that the original method must be better (or at worst) equal to mine for the shared memory transfer.
For the ne2k case it is relatively easy to determine the dma setup cost by pinging with a packet size that just fits into one page and then add one byte. I'll do that when I get rid of all the printks currently rolling over my screen in order to find the retransmission bug (ARP is free to go for now btw, and the issue is very much timing related and there seem to be more than one bug active).
Let me also point out that the so called DMA on the ne2k NIC is not (currently) NIC-to-PC DMA. All data transfers are programmed (stosw/stosb). This may not be all that different from what fmemcpy does timingwise. The 'dma setup' is loading the NIC registers with counts and (internal) addresses plus status check looping at the end (monitoring the Remote DMA Complete interrupt bit).
So - if I read you correctly @ghaerr - your conclusion is that for the wd driver, the original setup (read 4, then read the rest whatever it is), is optimal, and I fully agree. Actually - @pawosm-arm - if your current version still has the modified setup which reads a full page in the first read, you kan make the test I suggested above, just for the heck of it. Make sure the packet size includes headers etc. (use tcpdump -v to check).

As to the scrollLock issue, it's a great idea, I am very much looking forward to it. Maybe it would make sense to have repeated keypushes scroll through a list of debug flags ...

Thanks.

--Mellvik

pawosm-arm:
@ghaerr I've tried following change and the driver started to fail badly:

```
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -49,8 +49,6 @@
 
 #define RX_PAGES       (WD_STOP_PG - WD_FIRST_RX_PG)
 
-#define ETH_PAGE_SIZE  256U
-
 #define WD_RESET       0x80U   /* Board reset */
 #define WD_MEMENB      0x40U   /* Enable the shared memory */
 #define WD_IO_EXTENT   32U
@@ -254,7 +252,7 @@ static void wd_term(void)
 
 static int wd_pack_get(byte_t * pack)
 {
-       const e8390_pkt_hdr *rxhdr = (e8390_pkt_hdr *)pack;
+       e8390_pkt_hdr *rxhdr = (e8390_pkt_hdr *)pack;
        byte_t *hdr_start;
        unsigned char this_frame;
        int res = 0;
@@ -270,8 +268,8 @@ static int wd_pack_get(byte_t * pack)
                        debug_eth("Mismatched read page pointers %2x vs %2x.\n",
                                this_frame, current_rx_page);
                hdr_start = (byte_t *)((this_frame - WD_START_PG) << 8U);
-               fmemcpyb(pack, kernel_ds, hdr_start, WD_SHMEMSEG,
-                       ETH_PAGE_SIZE);
+               *rxhdr = *((e8390_pkt_hdr __far *)
+                       (_MK_FP((this_frame - WD_START_PG) << 8U, WD_SHMEMSEG)));
                if ((rxhdr->count < 64U) ||
                    (rxhdr->count > (MAX_PACKET_ETH + sizeof(e8390_pkt_hdr)))) {
                        debug_eth("eth: bogus packet size: %d, "
@@ -286,11 +284,10 @@ static int wd_pack_get(byte_t * pack)
                                "status = %#2x nxpg = %#2x size = %d\n",
                                rxhdr->status, rxhdr->next, rxhdr->count);
                        res = -EIO;
-               } else if (rxhdr->count > ETH_PAGE_SIZE) {
-                       fmemcpyb(pack + ETH_PAGE_SIZE, kernel_ds,
-                               hdr_start + ETH_PAGE_SIZE, WD_SHMEMSEG,
-                               rxhdr->count - ETH_PAGE_SIZE);
-               }
+               } else
+                       fmemcpyb(pack + sizeof(e8390_pkt_hdr), kernel_ds,
+                               hdr_start + sizeof(e8390_pkt_hdr), WD_SHMEMSEG,
+                               rxhdr->count - sizeof(e8390_pkt_hdr));
                OUTB(current_rx_page - 1U, WD_8390_PORT + EN0_BOUNDARY);
        } while (0);
        set_irq();
```

ghaerr:
Hello @Mellvik,

> your conclusion is that for the wd driver, the original setup (read 4, then read the rest whatever it is), is optimal, and I fully agree. 

Yes, we are now on the same "page" (pun intended).

> I've tried following change and the driver started to fail badly:

By inspection, this looks correct, but I suspect the problem is something to do with the line:
```
               *rxhdr = *((e8390_pkt_hdr __far *)
                       (_MK_FP((this_frame - WD_START_PG) << 8U, WD_SHMEMSEG)));
```
This tries to do a "structure copy", which I don't recommend, as the compiler may call yet another library routine for that, especially since its a far copy. I recommend using a seperate __far \* pointer and accessing the ethernet header elements  directly, via the far pointer. gcc may not support a __far structure copy, you should use the "objdump ..." alias for looking at the code produced to find out for sure.

HOWEVER - looking hard at this code has revealed a major speed issue with both the NE2K and WD drivers - that of TWO fmemcpyb's back to back, needlessly slowing down the system. I did mention that the kernel is full of memory copies, right? Well, here's another case right under our noses:

I'll use the wd.c code as an example, but ne2k.c does the same thing:

1. The wd_read procedure is called with a kernel supplied buffer pointer "data" and size "len". (this happens to be the actual application buffer in another address space).
2. wd_pack_get is called with a local buffer "recv_buf".
3. wd_pack_get reads the ethernet header as well as the data into recv_buf.
4. wd_read uses "size = ((e8390_pkt_hdr *)(recv_buf))->count" to access the actual data size after wd_pack_get.
5. wd_read uses memcpy_to_fs (which is essentially another fmemcpyb) to copy the data *again* to its destination.

Fixing this requires knowing a bit more about how the kernel works, but essentially the following should be done (this is the simple version:)
1. wd_pack_get should only access the packet length and status *inside* itself, and not copy that data into recv_buf.
2. wd_pack_get should return the packet length as a its function return value.
3. Use memcpy_tofs as is to copy recv_buf which doesn't contain ethernet header (yes, double copy for now).

After getting that working, the following should be done:
1. The wd_read "data" and "len" should be passed to wd_pack_get and wd_pack_get should write directly into the passed (far) buffer. (details on how to get that pointer are in memcpy_tofs).
2. Slight complications arise if the passed length to wd_read is smaller than the packet. (this will never happen, as ktcp always passes the largest size possible, but should still be handled for good kernel programming).
3. This approach gets ride of the entire local recv_buf buffer, saving the kernel another 1k bytes of data, as well as greatly speeds up the driver.

In addition, memcpy_tofs, in a protected mode system, does a length-check to assure that the receive buffer (which is actually the far pointer to the application buffer itself, is valid. The 8086 doesn't do that, but we should have a check anyways). More info in that later, if anyone is interested.

Thank you!

Mellvik:
> 
> your conclusion is that for the wd driver, the original setup (read 4, then read the rest whatever it is), is optimal, and I fully agree.
> 
> Yes, we are now on the same "page" (pun intended).
> 
@ghaerr, hopefully we learned something along the way :-). BTW I have another (surprising?) 'same page' message for you - separately.
> HOWEVER - looking hard at this code has revealed a major speed issue with both the NE2K and WD drivers - that of TWO fmemcpyb's back to back, needlessly slowing down the system. I did mention that the kernel is full of memory copies, right? Well, here's another case right under our noses:
> 
> I'll use the wd.c code as an example, but ne2k.c does the same thing:
> 
Thank you for pointing this out @ghaerr, I seem to remember we briefly touched this many months ago, now it the time to fix (and measure the improvement).
> The wd_read procedure is called with a kernel supplied buffer pointer "data" and size "len". (this happens to be the actual application buffer in another address space).
> wd_pack_get is called with a local buffer "recv_buf".
> wd_pack_get reads the ethernet header as well as the data into recv_buf.
> wd_read uses "size = ((e8390_pkt_hdr *)(recv_buf))->count" to access the actual data size after wd_pack_get.
> wd_read uses memcpy_to_fs (which is essentially another fmemcpyb) to copy the data again to its destination.
> Fixing this requires knowing a bit more about how the kernel works, but essentially the following should be done (this is the simple version:)
> 
> wd_pack_get should only access the packet length and status inside itself, and not copy that data into recv_buf.
> wd_pack_get should return the packet length as a its function return value.
> Use memcpy_tofs as is to copy recv_buf which doesn't contain ethernet header (yes, double copy for now).
> After getting that working, the following should be done:
> 
> The wd_read "data" and "len" should be passed to wd_pack_get and wd_pack_get should write directly into the passed (far) buffer. (details on how to get that pointer are in memcpy_tofs).
> Slight complications arise if the passed length to wd_read is smaller than the packet. (this will never happen, as ktcp always passes the largest size possible, but should still be handled for good kernel programming).
> This approach gets ride of the entire local recv_buf buffer, saving the kernel another 1k bytes of data, as well as greatly speeds up the driver.
> In addition, memcpy_tofs, in a protected mode system, does a length-check to assure that the receive buffer (which is actually the far pointer to the application buffer itself, is valid. The 8086 doesn't do that, but we should have a check anyways). More info in that later, if anyone is interested.
> 
> 
I like this change, thank you for the detailed procedure. It will kill my dma read optimisation entirely, but presumably the benefit will balance the disadvantage. I'll record some benchmarks before making the change.

It's easy to see the advantages both in terms of speed and space/footprint. I'll add this to the ne2k driver before submitting all the other changes I've been talking about. In doing so I might need some input as to setting up the far copying in asm code.

Thanks!!!

—Mellvik

ghaerr:
Looking at the implementation of `memcpy_tofs`, I find that
```
memcpy_tofs(data, recv_buf, len)
```
is equivalent to
```
fmemcpy(data, current->t_regs.ds, recv_buf, kernel_ds, len);
```
The 2nd parameter current->t_regs.ds uses the current task's (that is, the application calling the driver) DS register, which is it's data segment.

Also, memcpy_tofs doesn't do any application buffer length checks. The kernel routine verified_memcpy_tofs does that, but that looks to have been skipped by the original driver writer. It really isn't technically needed, (see routine for details), but checks to see against bad values passed to the system `read` call in the first place.

Another possible big system speedup would be to slightly rewrite memcpy_tofs to use word moves only, followed by a single byte move if the length is odd. I have added that to my list.

ghaerr:
> t will kill my dma read optimisation entirely, but presumably the benefit will balance the disadvantage. I'll record some benchmarks before making the change.

I wasn't aware that NE2K didn't actually do hardware DMA. 

I have to say, there are big advantages to have a driver written in C. Prior to Paul writing the WD driver, I didn't really understand the overview of exactly what is happening in what now seems like both drivers.

Perhaps you should seriously consider adapting the WD driver C code for the NE2K NIC (or at least the wd_pack_get portion)?

> It's easy to see the advantages both in terms of speed and space/footprint. I'll add this to the ne2k driver before submitting all the other changes I've been talking about.

It is pretty easy to call a C procedure from ASM if converting the NE2K driver is out of the question for now. I still don't quite get the DMA setup vs directly memory-mapped method of reading the NIC data packet.

>  In doing so I might need some input as to setting up the far copying in asm code.

If you want to keep it all in ASM, just look at the memcpy_tofs ASM source, its in elks/arch/i86/mm/segment.S:
```
/* void memcpy_tofs(void *daddr, void *saddr, size_t len);*/
    .global memcpy_tofs

memcpy_tofs:
    mov %si,%ax
    mov %di,%dx
    mov %es,%bx
    mov current,%si
    mov TASK_USER_DS(%si),%es
    mov %sp,%si
    mov 2(%si),%di
    mov 6(%si),%cx
    mov 4(%si),%si
    cld
    rep
    movsb
    mov %bx,%es
    mov %dx,%di
    mov %ax,%si
    ret
``` 
Straightforward, with the exception of using a non-standard stack frame to quickly get at the arguments passed to it and later quickly exit. The "current" and "TASK_USER_DS" essentially setup the ES register for the destination segment, which is "current->t_regs.ds".

Mellvik:
@ghaerr,
the wording is confusing - easy to handle on the inside, meaningless on the outside. There is no DMA - on the outside (yet, may do something about that later). There is programmet IO.
The DMA (name) usage comes from the docs and the internal architecture of the interface - there is DMA between the NIC chip and the memory chips and DMA between the memory chips and the programmed IO interface. SO there are DMA registers, DMA operations etc, which makes the use of the acronym reasonable.

> 24. sep. 2020 kl. 18:57 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> t will kill my dma read optimisation entirely, but presumably the benefit will balance the disadvantage. I'll record some benchmarks before making the change.
> 
> I wasn't aware that NE2K didn't actually do hardware DMA.
> 
This is about reading the full (first) ring buffer page vs the first two bytes and then the rest. Having only the kernel buffer available, there will obviously always be 2 operations. In NIC parlance, 2 dma operations (!)
> I have to say, there are big advantages to have a driver written in C. Prior to Paul writing the WD driver, I didn't really understand the overview of exactly what is happening in what now seems like both drivers.
> 
> Perhaps you should seriously consider adapting the WD driver C code for the NE2K NIC (or at least the wd_pack_get portion)?
> 
I have no problems with the advantages of using C, and I can go on for a while with the advantages of asm. We'll keep it for now, I like the speed and being close to the HW, in particular when there are tricky problems to solve. Rewriting the asm code in C would probably take less than a day, but for now what we have is my preference. Gave me an excuse to learn 86 assembly, btw.
> It's easy to see the advantages both in terms of speed and space/footprint. I'll add this to the ne2k driver before submitting all the other changes I've been talking about.
> 
> It is pretty easy to call a C procedure from ASM if converting the NE2K driver is out of the question for now. I still don't quite get the DMA setup vs directly memory-mapped method of reading the NIC data packet.
> 
I hope the explanation above did the trick … it's just programmed I/O with a different name.
> In doing so I might need some input as to setting up the far copying in asm code.
> 
> If you want to keep it all in ASM, just look at the memcpy_tofs ASM source, its in elks/arch/i86/mm/segment.S:
> 
> /* void memcpy_tofs(void *daddr, void *saddr, size_t len);*/
>     .global memcpy_tofs
> 
> memcpy_tofs:
>     mov %si,%ax
>     mov %di,%dx
>     mov %es,%bx
>     mov current,%si
>     mov TASK_USER_DS(%si),%es
>     mov %sp,%si
>     mov 2(%si),%di
>     mov 6(%si),%cx
>     mov 4(%si),%si
>     cld
>     rep
>     movsb
>     mov %bx,%es
>     mov %dx,%di
>     mov %ax,%si
>     ret
> Straightforward, with the exception of using a non-standard stack frame to quickly get at the arguments passed to it and later quickly exit. The "current" and "TASK_USER_DS" essential setup the ES register for the destination segment, which is "current->t_regs.ds".
> 
> 
Thanks - I'll do phase one of the change now, phase 2 when the tcp stuff is done.

—Mellvik

pawosm-arm:
> This tries to do a "structure copy", which I don't recommend, as the compiler may call yet another library routine for that, especially since its a far copy.

FYI, this didn't work either:

```
@@ -270,8 +269,11 @@ static int wd_pack_get(byte_t * pack)
                        debug_eth("Mismatched read page pointers %2x vs %2x.\n",
                                this_frame, current_rx_page);
                hdr_start = (byte_t *)((this_frame - WD_START_PG) << 8U);
-               fmemcpyb(pack, kernel_ds, hdr_start, WD_SHMEMSEG,
-                       ETH_PAGE_SIZE);
+               farxhdr =
+                       (_MK_FP((this_frame - WD_START_PG) << 8U, WD_SHMEMSEG));
+               rxhdr->status = farxhdr->status;
+               rxhdr->next = farxhdr->next;
+               rxhdr->count = farxhdr->count;
                if ((rxhdr->count < 64U) ||
                    (rxhdr->count > (MAX_PACKET_ETH + sizeof(e8390_pkt_hdr)))) {
                        debug_eth("eth: bogus packet size: %d, "
```

ghaerr:
@pawosm-arm,

Can you do a `ia16-elf-objdump -D -r -Mi8086 wd.o` and find that portion of wd_pack_get that loads farxhdr? That is likely the problem. 

(I alias odd='ia16-elf-objdump -D -r -Mi8086' since I use it a lot).

[EDIT: the problem is that the segment and offset are reversed in _MK_FP, sorry!!!]

pawosm-arm:
@ghaerr you were right.

With this change, I did `ping -i 0.5 -s 1400` test, and the results are still close to 53ms:

```
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=53.2 ms
1408 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=53.0 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=53.0 ms
1408 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=53.1 ms
1408 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=53.0 ms
1408 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=53.0 ms
1408 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=53.0 ms
1408 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=53.1 ms
1408 bytes from 192.168.1.44: icmp_seq=10 ttl=64 time=53.3 ms
1408 bytes from 192.168.1.44: icmp_seq=11 ttl=64 time=53.3 ms
1408 bytes from 192.168.1.44: icmp_seq=12 ttl=64 time=53.3 ms
1408 bytes from 192.168.1.44: icmp_seq=14 ttl=64 time=53.3 ms
1408 bytes from 192.168.1.44: icmp_seq=15 ttl=64 time=53.3 ms
1408 bytes from 192.168.1.44: icmp_seq=16 ttl=64 time=53.3 ms
1408 bytes from 192.168.1.44: icmp_seq=17 ttl=64 time=53.2 ms
1408 bytes from 192.168.1.44: icmp_seq=18 ttl=64 time=53.3 ms
1408 bytes from 192.168.1.44: icmp_seq=19 ttl=64 time=53.1 ms
1408 bytes from 192.168.1.44: icmp_seq=20 ttl=64 time=53.0 ms
1408 bytes from 192.168.1.44: icmp_seq=21 ttl=64 time=52.9 ms
1408 bytes from 192.168.1.44: icmp_seq=23 ttl=64 time=53.0 ms
1408 bytes from 192.168.1.44: icmp_seq=24 ttl=64 time=53.0 ms
```

ghaerr:
> With this change, I did ping -i 0.5 -s 1400 test, and the results are still close to 53ms:

Yes, not much difference in 256 byte double-copy. But we're now closer (in design) to implementing getting rid of the 3rd copy (of the entire packet), which, over time or with long file transfers, will certainly add up.

I'm really glad you chose to write the WD driver in C. This has helped illuminate other issues, which will help with issues #521 and #745.

Mellvik:
Med vennlig hilsen/best regards,
Helge Skrivervik
myMAYDAY.com - 9201 7146

Mellvik:
> If you want to keep it all in ASM, just look at the memcpy_tofs ASM source, its in elks/arch/i86/mm/segment.S:
> 
> /* void memcpy_tofs(void *daddr, void *saddr, size_t len);*/
>     .global memcpy_tofs
> 
> memcpy_tofs:
>     mov %si,%ax
>     mov %di,%dx
>     mov %es,%bx
>     mov current,%si
>     mov TASK_USER_DS(%si),%es
>     mov %sp,%si
>     mov 2(%si),%di
>     mov 6(%si),%cx
>     mov 4(%si),%si
>     cld
>     rep
>     movsb
>     mov %bx,%es
>     mov %dx,%di
>     mov %ax,%si
>     ret
> Straightforward, with the exception of using a non-standard stack frame to quickly get at the arguments passed to it and later quickly exit. The "current" and "TASK_USER_DS" essential setup the ES register for the destination segment, which is "current->t_regs.ds".
> 
> —
> 
> 
Quick question @ghaerr,
What will be the setup here in a programmed i/o setting?
inw %dx,%ax; stosw; loop
How to get the es/di (?) pointers right?

- Mellvik [iPad]

ghaerr:
> What will be the setup here in a programmed i/o setting?
inw %dx,%ax; stosw; loop
How to get the es/di (?) pointers right?

This needs to be done in two steps, as suggested in my post above, The first step doesn't require any segment register changes, those changes will be harder to understand because your routine is in assembly.

For ne2k.c step 1, your driver needs to change the code, just as suggested in wd.c, to read the ethernet header values separately into registers, rather than the recv_buffer, and return the size instead:
```
        
                if (ne2k_pack_get(recv_buf)) {    <-- return size and check for 0 (error) here
                        res = -EIO;
                        break;
                }
        
                size = *((word_t *) (recv_buf + 2)); <- eliminate this line
                if (len > size) len = size;
                memcpy_tofs(data, recv_buf + 4, len);   /* +4: Discard NIC header. */ <- eliminate + 4 here
```
It is important to use two steps so that you can make the change without losing control over the stability of your driver.

The second step will be passing "data, len" directly to ne2k_pack_get, which will then have to check for "if (len > size) len = size" internally, and use the same segment registers as before, except that it will use the application's data segment in ES, rather than the kernel data segment already in DS (and copied to ES).

The memcpy_tofs code above shows how to get ES loaded with the application data segment. It uses a temporary SI as a pointer into the task structure, which is always pointed to by the kernel global "current":
```
     mov current,%si
     mov TASK_USER_DS(%si),%es
```
I wouldn't advise doing the second step until you fully understand the memcpy_tofs code that is in your current driver. Taking time to fully understand what we are doing here will help you understand the kernel.

Mellvik:
Thanks, @ghaerr, 
What I really was after, was whether what we're trying to do here is indeed doable inside a stosw-loop. Making the effort worthwhile... Seems like it is :-)

-m

ghaerr:
For step 2, in your code, instead of:
```

        // I/O read loop

        mov     %ds,%ax
        mov     %ax,%es
```
You might use (if BX available):
```
#include <arch/asm-offsets.h>
       ...
      .data
      .extern current
      ...
      mov current,%bx
      mov TASK_USER_DS(%bx),%es
```
As suggested, look at arch/i16/mm/segment.S for the memcpy_tofs code. Look at linuxmt/sched.h and arch/asm-offsets.h to see what TASK_USER_DS is (it is the offset of the saved DS register when the application made the system call to read,which is saved in task_struct's t_regs structure which is defined in elks/include/arch/types.h). Complicated, but worth tracing through to understand the kernel data structures when a driver is called. The point is that the application's data segment is always in DS, which gets saved onto the kernel stack, which is always pointed to by current, which is a global pointer to the process that was running's task_struct at the time of hw interrupt or syscall. That's where you will be copying the data, directly from the NIC, using the ES register as the segment and the passed "data" value as the offset. The C code for the same segment value is "current->t_regs.ds", and "data" parameter. Pretty cool, really.

Get step 1 running while you study step 2.

ghaerr:
> whether what we're trying to do here is indeed doable inside a stosw-loop.

The real question is: do all bytes have to be read from the NIC's memory-mapped (or as you call DMA) buffer for a given packet on each read call?

Since the "if (len > size) len = size" would now have to be handled inside the driver packet read routine, that means that an application that asks for less than the available packet would not copy from the NIC (or DMA, as you call it) all of the packet data in that case. Does that matter?

Note this is completely rhetorical at this point since ktcp always passes a full ethernet sized buffer to read. But just wondering. If the NIC needs to have all the bytes read in order to advance an internal buffer counter, then the loop will have to be split into two parts, one with inb/stosw, and just inb with no stosw past the requested length.

For the WD C version, this could all be handled in a small change to a new memcpy_tofs routine that is passed two length parameters, instead of all sorts of ASM complications. If it doesn't matter, then fmemcpyb could be used directly, and passed the requested len = MIN(size, len) parameter with the destination segment being current->t_regs.ds.

Mellvik:
Thanks again @ghaerr - 
This turns out to be another great learning experience,like you alluded to.

Appreciated. 
More later, 

-M

Mellvik:
> 
> whether what we're trying to do here is indeed doable inside a stosw-loop.
> 
> The real question is: do all bytes have to be read from the NIC's memory-mapped (or as you call DMA) buffer for a given packet on each read call?
> 
No, we read as much as we like all the time.
> Since the "if (len > size) len = size" would now have to be handled inside the driver packet read routine, that means that an application that asks for less than the available packet would not copy from the NIC (or DMA, as you call it) all of the packet data in that case. Does that matter?
> 
Well, yes it does. The driver does not support coming back and read more from the same packet, that would require local buffering, or at least some local state not currently supported.
> Note this is completely rhetorical at this point since ktcp always passes a full ethernet sized buffer to read. But just wondering. If the NIC needs to have all the bytes read in order to advance an internal buffer counter, then the loop will have to be split into two parts, one with inb/stosw, and just inb with no stosw past the requested length.
> 
No problem, except as noted above.
> For the WD C version, this could all be handled in a small change to a new memcpy_tofs routine that is passed two length parameters, instead of all sorts of ASM complications. If it doesn't matter, then fmemcpyb could be used directly, and passed the requested len = MIN(size, len) parameter with the destination segment being current->t_regs.ds.
> 
Agreed, and afaik - what can be done in C, ...
> —
> 
> 
-M

ghaerr:
>>  Since the "if (len > size) len = size" would now have to be handled inside the driver packet read routine, that means that an application that asks for less than the available packet would not copy from the NIC (or DMA, as you call it) all of the packet data in that case. Does that matter?

> Well, yes it does. The driver does not support coming back and read more from the same packet, that would require local buffering, or at least some local state not currently supported.

No, that's not what I'm asking about. What happens if the driver doesn't read (via programmed I/O, memory mapped addresses, DMA, or whatever you want to call it) all of the packet data from the NIC? Anything? Does the card care how much is read (via programmed I/O, etc) after a read interrupt?

I can submit an updated wd.c with all the changes for highest performance (but not tested), if anyone is interested, to see what I'm talking about.

Mellvik:
Sorry @ghaerr,
I did understand the question and figured it was covered by the first part of the answer:
> The real question is: do all bytes have to be read from the NIC's memory-mapped (or as you call DMA) buffer for a given packet on each read call?
> 
    No, we read as much as we like all the time.

The additional comment was an observation about the unread data.

So if the len and actual packet size are different, we use the smaller. This has been working in the driver since a few days back. Then I ran into a memory corruption issue that needs to be figured out before I can continue.

Thanks for the intro to the setup of far 'memory transfers', it turned out to be easy after that. No performance data yet, till the other issue is sorted out. 
Is this (i e shortcutting a buffer level) something we can (reasonably) do for transmit as well? Performance enhancement is very welcome...

National Semi's use of DMA in this context is really confusing, but it's all  over the documentation and in most drivers, so it sticks. One of these days it may be an interesting experiment to see if real DMA works and is beneficial. Is there any ELKS driver that is actually using DMA - so that we have a working example?

--Mellvik

pawosm-arm:
> I can submit an updated wd.c with all the changes for highest performance (but not tested), if anyone is interested, to see what I'm talking about.

@ghaerr as I stated before, I can make those tests for you

ghaerr:
> Then I ran into a memory corruption issue that needs to be figured out before I can continue.

Is that a new corruption issue in the NE2K driver you're talking about?

> Is this (i e shortcutting a buffer level) something we can (reasonably) do for transmit as well? 

Yes, it looks so. The application buffer is completely copied to send_buf before calling ne2k_pack_put. Just like with reading, the only real difference is coding a far address in the actual NIC packet routine and using the passed length parameter. 
However - this will only work should the NIC routine wait until the write buffer is completely copied to the NIC before transmission, which I believe to be the case in both the WD and NE2K drivers.

> it may be an interesting experiment to see if real DMA works and is beneficial.

There may be some benefit, but not much: the read routine (and process) has to wait until the packet data is fully copied to the application buffer, via programmed I/O or hw DMA, as the NIC read wakes up the sleeping task at the time of NIC packet available. On the write side, hw DMA could be used to start the copy from the application buffer to the NIC, but then the kernel would have to keep the data in a secondary buffer like send_buf (copied from the application), as it can't allow the application to continue running until the application buffer has been fully copied.

> Is there any ELKS driver that is actually using DMA

Not that I'm aware of - almost all ELKS I/O is synchronous; all disk I/O uses synchronous BIOS calls, and the buffer I/O subsystem has never been operational using the interrupt-driven method DMA would require. I think ELKS would need non-BIOS floppy and hard disk drivers in order to take advantage of DMA. And then there's the fact that ELKS 8086 systems just aren't that fast, and most of the time there's only a single application running which is waiting for the disk to spin to the correct sector anyways...

I'll submit a revision of the WD driver with both the read and write routines optimized to demonstrate what I've been talking about, as Paul can test it.

Mellvik:
> 
> ﻿
> Then I ran into a memory corruption issue that needs to be figured out before I can continue.
> 
> Is that a new corruption issue in the NE2K driver you're talking about?
> 
Yes, this is me creating work for myself. It's the cost of learning afaik.
> Is this (i e shortcutting a buffer level) something we can (reasonably) do for transmit as well?
> 
> Yes, it looks so. The application buffer is completely copied to send_buf before calling ne2k_pack_put. Just like with reading, the only real difference is coding a far address in the actual NIC packet routine and using the passed length parameter.
> However - this will only work should the NIC routine wait until the write buffer is completely copied to the NIC before transmission, which I believe to be the case in both the WD and NE2K drivers.
> 
Yes, this is the way theNIC works. Transferring data to the ring buffer and then to the wire are atomic processes. 
> it may be an interesting experiment to see if real DMA works and is beneficial.
> 
> There may be some benefit, but not much: the read routine (and process) has to wait until the packet data is fully copied to the application buffer, via programmed I/O or hw DMA, as the NIC read wakes up the sleeping task at the time of NIC packet available. On the write side, hw DMA could be used to start the copy from the application buffer to the NIC, but then the kernel would have to keep the data in a secondary buffer like send_buf (copied from the application), as it can't allow the application to continue running until the application buffer has been fully copied.
> 
Ok, sounds like simply 'not worth the effort' for now.
> Is there any ELKS driver that is actually using DMA
> 
> Not that I'm aware of - almost all ELKS I/O is synchronous; all disk I/O uses synchronous BIOS calls, and the buffer I/O subsystem has never been operational using the interrupt-driven method DMA would require. I think ELKS would need non-BIOS floppy and hard disk drivers in order to take advantage of DMA. And then there's the fact that ELKS 8086 systems just aren't that fast, and most of the time there's only a single application running which is waiting for the disk to spin to the correct sector anyways...
> 
Agreed. Then there is appearance. ELKS is - I'm probably repeating myself on this one, but it's sort of my metric for ELKS since it's on the same physical machine - slow. So I'm always interested in the dma/interrupt potential. Including hd/fd drivers squeezing the last millisecond out of the hw.
> I'll submit a revision of the WD driver with both the read and write routines optimized to demonstrate what I've been talking about, as Paul can test it.
> 
Great, I'm really looking forward to the ping numbers!! 

-- M

Mellvik:
Apropos the C vs asm issue:

I have a partly working 3C509 driver that I have decided  to rewrite in C when I get around to it. Found the interface in my stash 3 weeks ago and gave it a few hours starting with the ne2k-base. Interrupts are responding, no data transfer yet. might be a useful addition to the growing ELKS stables.

 -M

PR #761, opened 2020-09-23
[kernel] Implement operating mixed-model kernel

ghaerr:
Demonstrates first fully working, configurable mixed-model ELKS kernel using a far text section.

The ELKS kernel code segment 64k limitation can now be, in theory, doubled to 128k, since a 64k .text section and an additional 64k .fartext code section are now supported. This PR reduces the kernel .text size over 4k bytes from 58896 bytes to 54736 by moving kernel init routines outside the almost-full .text segment.

Tested only on REL_SYS (default) system generation. ROM code not yet supported.
Enabled by 'Far text kernel' (CONFIG_FARTEXT_KERNEL) in menuconfig. Default ON for now. 
You'll notice a few new debug displays early on the boot process: these are the text, fartext, and data segments and relocations.

Moves almost all kernel initialization routines into a .fartext.init section, with the idea that memory used for kernel initialization can be freed and reused for user processes in the future. Future plans include moving MINIX and/or FAT filesystem code into .fartext segment to greatly reduce near text kernel code size.

There is a pretty small overhead when executing .fartext code that calls .fartext code, given the space saving benefit. However, there is much more overhead when calling near routines from .fartext functions. More work and analysis is ongoing in this area. For instance, while the execution speed of initialization code matters little, the init code calling near functions increased the .fartext section size by around 640 bytes. Thus, the system overall is a bit larger (until reclamation, at least).

A big thanks to our resident master compiler enhancer and maintainer @tkchia. His work work adding medium model support, a much updated compiler and linker (in #747), and help in debugging the kernel has allowed us to achieve this milestone. We now have a way to add lots more code to the ELKS kernel in support of enhancements as well as the ability to produce a single kernel that supports all drivers and filesystems linked in, should we want to.


Mellvik:
Congratulations @tkchia & @ghaerr, this is indeed a milestone.
The way I understand it, this opens up room for new functionality that was previously impossible - or required something to go in order to get in.

Two questions:
 - Now that the FARTEXT option is enabled by default, what should we be looking for that may be different, if anything - particularly on physical hw? Performance, kernel messages, ... in order to contribute to debugging/ehnancements/...
 - If the FARTEXT config option is off, is there any change at all?

Thank you!
--Mellvik

ghaerr:
Hello @Mellvik,

>  - Now that the FARTEXT option is enabled by default, what should we be looking for that may be different, if anything - particularly on physical hw? Performance, kernel messages, ... in order to contribute to debugging/ehnancements/...

So far, there should be no difference in kernel execution, after boot and subsequent system initialization.

Currently, the process of converting various sections of the kernel to far text (meaning medium model) is somewhat slow and tedious, as each and every C function needs to be tagged separately with an attribute specifier. Moreover, it is imperative that a forward-declaration exist and its header file included in each compilation unit, as otherwise the possibility exists that the function be compiled far but called near, which will cause a  system crash when called, yet not tagged by the compiler/linker. (This is correct, @tkchia? Or could name-mangling be used to prevent such from occurring?).

So far, only initialization routines have marked far and moved. That means once the kernel successfully boots, the rest of the kernel should operate identically as before, which is why the default has been set ON for now. 

Code running in medium model runs at the same speed as small model, except for the overhead of function calls, where CS and IP are changed on a function call/return, versus just IP in small model. In addition, there is more overhead when a near function is called from a far function; this is called thunk overhead.

It has yet to be decided which other routines may want to be converted to medium model, but my thoughts have been that moving the filesystem code may make sense, as 1) they are large and removing them opens lots of additional .text space, and 2) they're relatively slow, since disk I/O is greatly slower than any medium model conversions would be.

To directly answer your first question: at this time, separating driver code that is used once (and only once) into a separate procedure that could be marked INITPROC (initialization far procedure) allows that code to reside in a separate segment, thus allowing for more complex initialization or probing routines that don't contribute to kernel .text size, should be considered. Take a quick look at bioshd_init in bioshd.c and idquery.c to see more details.

I haven't yet converted any of the character special device driver init routines; these include console, m TTY, PTY, serial, network, etc, so the networking drivers are completely unaffected, since they're currently under construction.

The data segment is still fixed at 64k, so keeping kernel messages short and to the point is important. The data portion of driver initialization routines cannot be discarded, although it might be possible to move fixed strings (.rodata) into the far text segment; I will have to consult with @tkchia on that one.

>  - If the FARTEXT config option is off, is there any change at all?

No, none at all. Thus it can be easily determined whether the far text option is a producing a bug.

Thank you!

PR #762, opened 2020-09-26
[kernel] Toggle debug display in kernel and applications with keystroke

ghaerr:
This implements a general purpose event system for kernel and user mode, which by default uses Ctrl-P from the console or serial login to toggle debug display in ktcp and selected debug classes (from debug.h) in the kernel.

Communication with user mode processes is done using the SIGURG signal, which works well because the default processing for that signal is to ignore it. When Ctrl-P is typed, the kernel toggles its displays of debug_xxx calls, which are routed to dprintk, and a SIGURG signal is sent to all user processes, for them to control their own debug modes.

To test:
Boot normally, run "net start", then telnet. Use ^P to turn on and off kernel and user mode networking. By default, ktcp is compiled with DEBUG_TCP on, but all can be configured in ktcp/config.h.

Enabled by \#define DEBUG_EVENT in debug.h (default ON).
Allows kernel or user mode startup with debug on or off using DEBUG_STARTDEF.
Implements debug callbacks and dprintk in the kernel.
Fixes signal \#16, assigns it to SIGURG.
Fixes no wakeup on PTY or serial TTY unless non-signal characters received, which prevents EINTR returns.
Improves PTY execution speed by not calling wake_up on each received character.
Fixes getty/login EINTR return checks to stop exiting.
Fixes ktcp exiting on interrupted select call.
Fixes ash shell to allow ISIG processing in linenoise, allows ^C/^Z signals.
Fixes telnet to not turn off OPOST or ISIG processing for ^P, disallows ^C/^Z signals.
Moves DEBUG_ETH out of ne2k.c driver into debug.h (not yet done for wd.c)

Adds temporary demonstration code in ne2k.c to show both toggle display of debug_eth statements as well as debug event display of Skip Count when ^P pressed.

### DIFF
--- a/elks/arch/i86/drivers/char/ntty.c
+++ b/elks/arch/i86/drivers/char/ntty.c
@@ -400,6 +400,7 @@ int tty_ioctl(struct inode *inode, struct file *file, int cmd, char *arg)
     case TCSETSW:
     case TCSETSF:
 	ret = verified_memcpy_fromfs(&tty->termios, arg, sizeof(struct termios));
+
 	/* Inform subdriver of new settings*/
 	if (ret == 0 && tty->ops->ioctl != NULL)
 	    ret = tty->ops->ioctl(tty, cmd, arg);
--- a/elks/arch/i86/drivers/char/ntty.c
+++ b/elks/arch/i86/drivers/char/ntty.c
@@ -81,6 +81,12 @@ int tty_intcheck(register struct tty *ttyp, unsigned char key)
 	    sig = SIGINT;
 	if (key == ttyp->termios.c_cc[VSUSP])
 	    sig = SIGTSTP;
+#if DEBUG_EVENT
+	if (key == ('P' & 0x1f)) {	/* ctrl-P*/
+	    debug_event();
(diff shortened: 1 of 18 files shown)


ghaerr:
@Mellvik,

This PR adds some temporary debug code to ne2k.c, used to demonstrate how the toggle debug system can be used to easily display kernel variables, such as _ne2k_skip_cnt. If you've been working on this portion of the NE2K driver, feel free to just overwrite this version with the version you've been working on.

Thanks!

Mellvik:
@ghaerr,
Yes I do have a number of changes 'in the works', but 
I'd also like to keep your updates, what's the smart way to keep both? Your debug toggle is something I've really looked forward to take advantage of.

-M

ghaerr:
> I'd also like to keep your updates, what's the smart way to keep both?

Just keep a copy of your ne2k.c before doing a pull upstream master, then copy my ne2k.c out of the way, and replace ne2k.c with your original. There are only a few lines change in ne2k.c, and only to show you how to use debug_setcallback(). Take a quick look at the diff.

PR #764, opened 2020-09-26
[tools] Update options to pass to binutils-ia16 ./configure, and bump up binutils-ia16 version

tkchia:
On the `binutils-ia16` end, I had changed the Binutils build option `--enable-hpa-segelf` to `--enable-x86-hpa-segelf`, to follow Binutils's conventions on option names.  (This option enables the [new IA-16 segment relocation representation](https://git.zytor.com/users/hpa/segelf/abi.git/plain/segelf.txt) now used in ELKS.)

With this commit, `tools/Makefile` now passes both the old and new option names to Binutils `./configure`.  This should correctly enable the option if we bump up the Binutils version in the future.


ghaerr:
Hello @tkchia,

Just after I pulled upstream and pushed this commit to my fork, I got the following "Run failed for master" error... I'm not sure why it would be happening on my fork and possibly not this one. Do you have any thoughts on this? I did commit #762 just prior, it also builds OK on my local systems.

I haven't run tools/build.sh on my local systems, the entire ELKS build proceeds without error on them. Unfortunately, as discussed prior, there isn't any way to determine the symbol that might be causing this error.

Screen Shot 2020-09-26 at 10 59 41 AM

Thank you!

tkchia:
Hello @ghaerr,

Argh!  It seems I broke the handling of the _x_`!` segment base symbols in a slightly earlier `binutils-ia16` commit (https://github.com/tkchia/binutils-ia16/commit/702b69b9bb5337f105b0d8942c5956f77b048c5d).  I will submit a PR to bump up the `binutils-ia16` version again.  Apologies and thanks!

ghaerr:
> I will submit a PR to bump up the binutils-ia16 version again. Apologies and thanks!

At this point, I am trying to revert the binutils version update in some other code I uploaded in #765.

It's probably better at this point for me to commit #765, and then leave the binutils version change alone, until ELKS really needs it. I think I erred in suggesting that we should do a quick update of the version number. It would be better to just leave it alone for now, for the other developers here, I think.

Thank you!

tkchia:
@ghaerr: OK.  Thanks!

tkchia:
@ghaerr: I guess I will need to see how I can beef up the regression testing for building ELKS binaries on my end (this is still a weak point, it seems).  Thanks!

ghaerr:
> I guess I will need to see how I can beef up the regression testing for building ELKS binaries on my end (this is still a weak point, it seems). 

Yes, and I think we will both need to be more careful in casually updating the GCC/binutils version numbers. Ideally, we might think of only updating the toolchain once per month or so, or when it absolutely has to be.

I also should have waited for the automatic test run to complete before merging the PR!

Thanks!

Issue #755, opened 2020-09-21
TCP retransmit issue


Mellvik:
While debugging some odd network send issues that may be ne2k driver related, I came across this one. And as much as I'd like to just fix it and go on, I figure it's useful with some advice.

Here's the scenario:
The outside server is pulling data (lots) from the httpd on ELKS. This was reasonably stable a while back, now I'm seeing regular problems, and this is one of them:
 - transfer is running smoothly, one packet in, one out.
 - then ELKS gets an ARP, and replies in stride
 - then two packets later trouble starts and the sequence goes like this (numbers just to illustrate the sequence):
```
...
ELKS                        server
data 101 ------------->
        <----------------   ACK 101
        <----------------   ARP req
data 102 ------------->
        <----------------    ACK 102
ARP reply ------------>
data 103 ------------->
       <-----------------   ACK 103
data 104 -------------> 
       <-----------------   ACK 104
data 103 ------------->                           retransmit
       <-----------------   ACK 104
data 105 ------------->
       <-----------------   ACK 105
data 104 ------------->                          retransmit
      <-----------------    ACK 105
data 106 ------------->
     <-----------------     ACK 106
data 105 ------------->                         retransmit
     <-----------------     ACK 106
...
```
... and so on a little longer until `ktcp: panic in write` terminates transmission.
For every retransmit there is a `retrans retry #1 rto 4 mem 360` message with increasing numbers, starting @ 369, ending @ 1988.

Any hints very welcome.

--M

ghaerr:
Hi @Mellvik,

> This was reasonably stable a while back, now I'm seeing regular problems, and this is one of them

I'm pretty sure this is the same nagging problem ktcp has had all along - for some reason, whenever ktcp gets tripped up and has to start doing retransmits, everything goes to heck. I have mentioned earlier that, for instance, when ktcp is forced to drop a packet (when either its ring buffer is full or an application hasn't run long enough to empty that buffer), ktcp drops the packet and the same thing happens.

> For every retransmit there is a retrans retry #1 rto 4 mem 360 message with increasing numbers

I hadn't noticed that the retransmits end up allocating yet more memory - they should just retransmit from the stored TCP packet, that's it. Although what could be happening is ALL the subsequent packets end up going through a alloc/retrans process.

[...]

PR #767, opened 2020-09-26
[kernel] Remove send_buf double buffer in wd.c network driver

ghaerr:
Hello @pawosm-arm,

Here's the second improvement to the wd.c driver. You can either apply the patch to your tree to test, or wait until the new toolchain build is done and apply this PR.

Thanks!

[wd2.patch.zip](https://github.com/jbruchon/elks/files/5287249/wd2.patch.zip)

### DIFF
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -18,14 +18,7 @@
 #include <linuxmt/sched.h>
 #include <linuxmt/limits.h>
 #include <linuxmt/mm.h>
-
-#define DEBUG_ETH	0	/* set =1 for debugging */
-
-#if DEBUG_ETH
-#define debug_eth	printk
-#else
-#define debug_eth(...)
-#endif
+#include <linuxmt/debug.h>
 
 /* I/O delay settings */
 #define INB	inb	/* use inb_p for 1us delay */
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -439,6 +439,7 @@ static int wd_ioctl(struct inode * inode, struct file * file,
 	case IOCTL_ETH_ADDR_GET:
 		memcpy_tofs((char *)arg, mac_addr, 6U);
 		break;
+#if 0 /* unused*/
 	case IOCTL_ETH_ADDR_SET:
 		err = -ENOSYS;
 		break;
@@ -446,7 +447,7 @@ static int wd_ioctl(struct inode * inode, struct file * file,
 		/* Get the hardware address of the NIC,	which may be different
 		 * from the currently programmed address. Be careful with this,
 		 * it may interrupt ongoing send/receives.
-		 * arg must be a 32 bytes array.
+		 * arg must be a 6 word array.
 		 */
 		wd_get_hw_addr((word_t *)arg);
 		break;
@@ -464,6 +465,7 @@ static int wd_ioctl(struct inode * inode, struct file * file,
 		/* Get the current overflow skip counter. */
 		err = -ENOSYS;
 		break;
+#endif
 	default:
 		err = -EINVAL;
 	}
@@ -572,7 +574,7 @@ void wd_drv_init(void)
 {
 	int err;
 	unsigned u;
-	word_t hw_addr[16U];
+	word_t hw_addr[6U];
 
 	do {
 		err = request_irq(WD_IRQ, wd_int, NULL);
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -140,8 +140,6 @@ static struct wait_queue txwait;
 static byte_t wd_inuse = 0U;
 static byte_t mac_addr[6U];
 
-static byte_t send_buf[MAX_PACKET_ETH];
-
 static unsigned char current_rx_page = WD_FIRST_RX_PG;
 
 static word_t wd_rx_stat(void);
@@ -321,18 +319,19 @@ static size_t wd_read(struct inode * inode, struct file * filp,
  * Pass packet to driver for send
  */
 
-static int wd_pack_put(byte_t * pack, word_t len)
+static size_t wd_pack_put(char *data, size_t len)
 {
-	int res = 0;
-
 	clr_irq();
 	do {
+		if (len > MAX_PACKET_ETH)
+			len = MAX_PACKET_ETH;
+		if (len < 64U) len = 64U;  /* issue #133 */
 		fmemcpyb((byte_t *)((WD_FIRST_TX_PG - WD_START_PG) << 8U),
-			WD_SHMEMSEG, pack, kernel_ds, len);
+			WD_SHMEMSEG, (byte_t *)data, current->t_regs.ds, len);
 		OUTB(E8390_NODMA | E8390_PAGE0, WD_8390_PORT + E8390_CMD);
 		if (INB(WD_8390_PORT + E8390_CMD) & E8390_TRANS) {
 			printk("eth: attempted send with the tr busy.\n");
-			res = -EIO;
+			len = -EIO;
 			break;
 		}
 		OUTB(len & 0xffU, WD_8390_PORT + EN0_TCNTLO);
@@ -342,13 +341,13 @@ static int wd_pack_put(byte_t * pack, word_t len)
 			WD_8390_PORT + E8390_CMD);
 	} while (0);
 	set_irq();
-	return res;
+	return len;
 }
 
 static size_t wd_write(struct inode * inode, struct file * file,
 	char * data, size_t len)
 {
-	size_t res = 0U;
+	int res;
 
 	do {
 		prepare_to_wait_interruptible(&txwait);
@@ -363,15 +362,7 @@ static size_t wd_write(struct inode * inode, struct file * file,
 				break;
 			}
 		}
-		if (len > MAX_PACKET_ETH)
-			len = MAX_PACKET_ETH;
-		memcpy_fromfs(send_buf, data, len);
-		res = len;
-		if (len < 64U) len = 64U;  /* issue #133 */
-		if (wd_pack_put(send_buf, len)) {
-			res = -EIO;
-			break;
-		}
+		res = wd_pack_put(data, len);
 	} while (0);
 	finish_wait(&txwait);
 	return res;


pawosm-arm:
Still works,

```
$ ping -i 0.5 -s 1400 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=57.6 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=57.7 ms
1408 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=57.6 ms
1408 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=57.6 ms
1408 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=57.6 ms
1408 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=57.6 ms
1408 bytes from 192.168.1.44: icmp_seq=9 ttl=64 time=57.6 ms
1408 bytes from 192.168.1.44: icmp_seq=10 ttl=64 time=57.6 ms
^C
--- 192.168.1.44 ping statistics ---
10 packets transmitted, 8 received, 20% packet loss, time 4511ms
rtt min/avg/max/mdev = 57.579/57.614/57.651/0.023 ms

$ telnet 192.168.1.44
Trying 192.168.1.44...
Connected to 192.168.1.44.
Escape character is '^]'.
login: root
# meminfo
  HEAP   TYPE  SIZE    SEG   TYPE    SIZE  CNT
  cdce   SEG     16   1edb   BUF    65536    1
  cdea   SEG     16   2edb   BUF    65536    1
  ce06   SEG     16   3edb   CSEG    6832    1
  ce22   SEG     16   4086   DSEG    8816    1
  ce3e   SEG     16   42ad   free   18544    0
  ce5a   TTY    128
  cee6   TTY     64
  cf32   SEG     16   5f9d   free  108144    0
  cf4e   SEG     16   7a04   DSEG   57120    1
  cf6a   SEG     16   4b13   CSEG   46016    2
  cf86   SEG     16   564f   DSEG   18544    1
  cfa2   SEG     16   87f6   DSEG    6192    1
  cfbe   SEG     16   5ad6   CSEG   19568    1
  cfda   SEG     16   8c7d   CSEG    6144    2
  cff6   SEG     16   8dfd   free    4400    0
  d012   SEG     16   8f10   DSEG   18544    1
  d02e   SEG     16   8afc   CSEG    4544    1
  d04a   SEG     16   8979   DSEG    6192    1
  d066   SEG     16   4734   DSEG   12112    1
  d082   TTY     64
  d0ce   TTY    256
  d1da   SEG     16   8c18   free    1616    0
  d1f6   SEG     16   4a29   free    3744    0
  d212   SEG     16   9397   free   50832    0
  d22e   free 11730
  Total heap  12862     Total mem  528976
  Memory usage  517KB total,  334KB used,  183KB free
#
```

pawosm-arm:
It was faster though... Could the performance loss be caused by other changes on master which I've rebuilt meanwhile?

pawosm-arm:
I've tried to revert your patch. After revert it's even worse:

```
$ ping -i 0.5 -s 1400 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=61.0 ms
1408 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=61.0 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=61.0 ms
1408 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=60.9 ms
1408 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=61.0 ms
1408 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=61.0 ms
1408 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=61.0 ms
1408 bytes from 192.168.1.44: icmp_seq=9 ttl=64 time=61.0 ms
1408 bytes from 192.168.1.44: icmp_seq=11 ttl=64 time=61.1 ms
^C
--- 192.168.1.44 ping statistics ---
11 packets transmitted, 9 received, 18.1818% packet loss, time 5005ms
rtt min/avg/max/mdev = 60.946/61.003/61.149/0.059 ms
```

Something on the master affected it.

ghaerr:
> It was faster though... Could the performance loss be caused by other changes on master which I've rebuilt meanwhile?

Hmmmm... a quick way would be to take just these wd2.patch changes out and retest. Or - ktcp is now being built with debug changes ON (use ^P at the console to turn on and off debug messages). So, even if the debug messages are toggled off (which is the default on boot), ktcp will be slower. To turn off the debug messages completely, edit "config.h" in the ktcp directory and turn off DEBUG_TCP and DEBUG_EVENT.

pawosm-arm:
You were right about DEBUG messages. Without them performance gain is observed.

Before your patch:
```
$ ping -i 0.5 -s 1400 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=49.8 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=49.5 ms
1408 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=49.6 ms
1408 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=49.6 ms
1408 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=49.6 ms
1408 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=49.6 ms
1408 bytes from 192.168.1.44: icmp_seq=9 ttl=64 time=49.6 ms
1408 bytes from 192.168.1.44: icmp_seq=10 ttl=64 time=49.8 ms
^C
--- 192.168.1.44 ping statistics ---
10 packets transmitted, 8 received, 20% packet loss, time 4510ms
rtt min/avg/max/mdev = 49.538/49.613/49.793/0.098 ms
```

With your patch applied:
```
$ ping -i 0.5 -s 1400 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=46.2 ms
^C
--- 192.168.1.44 ping statistics ---
8 packets transmitted, 7 received, 12.5% packet loss, time 3503ms
rtt min/avg/max/mdev = 46.173/46.200/46.225/0.017 ms
```

ghaerr:
> performance gain is observed.

Well, this is all good news! We are saving quite a bit of time getting rid of the multiple buffer copies in your WD driver. Thank you very much for testing. I'm going to push one other small change to this commit (the DEBUG_ETH setting has been moved to the system include file debug.h) and then commit this PR.

> You were right about DEBUG messages.

Great. I would like to leave the DEBUG messages on but toggled off, at least for a bit, so that we can get to the bottom of the TCP retransmit #755 and some other issues quickly. So, now we know that your driver can achieve ~46ms ping times, which is pretty fantastic! The new "baseline" for these will be ~11ms higher, with debugging toggle enabled.

Thanks for your quick testing!

ghaerr:
Oh @pawosm-arm, one other BIG thing:

I see that you've completely surrounded both\ wd_pack_get and wd_pack_out with interrupts off. That's really not good, considering that a very long (relatively) fmemcpyb is performed with interrupts off. This will be very bad should the serial port be receiving high speed data, for instance. We need to set interrupts off for a much shorter period.

We need to discuss where, in fact, those clr_irq/set_irq pairs need to go. They are not needed in memory copies, I'm not sure exactly where they might be needed with regards to protecting the interrupt handler from munging things... can you explain a bit what you're thinking needs protection exactly?

We will handle this interrupt issue in another PR.

Thanks!

pawosm-arm:
> We need to discuss where, in fact, those clr_irq/set_irq pairs need to go. They are not needed in memory copies, I'm not sure exactly where they might be needed with regards to protecting the interrupt handler from munging things... can you explain a bit what you're thinking needs protection exactly?

I'd rather ask @Mellvik as he seems to know better which operations on 8390 need protection.

ghaerr:
I pushed one more commit that reduces kernel stack size on the wd_get_hw_addr call in the init routine and ifdef'd out the almost completely unused portions of ioctl.

pawosm-arm:
> I pushed one more commit that reduces kernel stack size on the wd_get_hw_addr call in the init routine and ifdef'd out the almost completely unused portions of ioctl.

The driver still works. No impact on pings (as expected though).

Mellvik:
@pawosm-arm, @ghaerr,
my morning coffe just got a lot better. Very encouraging indeed.

@pawosm-arm, could you send some timings for ping without options, no debug? 

As to interrupts, my recommendation is to keep them on all the time and see what happens. Then put them back in and experiment with stepwise reduced scope. 
 In the ne2k driver, the only part that is protected is the so called remote dma, really the register setup and the PIO transfer loop. I'm going to experiment with these too when everything else stabilizes, as my understanding of the ELKS interrupt structure improves  (thanks @ghaerr).

--Mellvik

pawosm-arm:
> @pawosm-arm, could you send some timings for ping without options, no debug?

```
$ ping 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 56(84) bytes of data.
64 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=7.15 ms
64 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=7.14 ms
64 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=7.40 ms
64 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=7.40 ms
64 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=7.34 ms
64 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=7.34 ms
64 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=7.33 ms
64 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=7.10 ms
64 bytes from 192.168.1.44: icmp_seq=9 ttl=64 time=7.14 ms
64 bytes from 192.168.1.44: icmp_seq=10 ttl=64 time=7.09 ms
^C
--- 192.168.1.44 ping statistics ---
10 packets transmitted, 10 received, 0% packet loss, time 9011ms
rtt min/avg/max/mdev = 7.088/7.241/7.403/0.122 ms

$ ping -i 0.5 -s 1400 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=46.0 ms
1408 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=46.0 ms
1408 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=46.3 ms
1408 bytes from 192.168.1.44: icmp_seq=9 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=10 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=11 ttl=64 time=46.2 ms
^C
--- 192.168.1.44 ping statistics ---
11 packets transmitted, 9 received, 18.1818% packet loss, time 5007ms
rtt min/avg/max/mdev = 45.981/46.170/46.294/0.094 ms
```
> As to interrupts, my recommendation is to keep them on all the time and see what happens.

Not much as I can observe:

```
$ ping 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 56(84) bytes of data.
64 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=7.11 ms
64 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=7.11 ms
64 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=7.43 ms
64 bytes from 192.168.1.44: icmp_seq=4 ttl=64 time=7.43 ms
64 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=7.44 ms
64 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=7.41 ms
64 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=7.13 ms
64 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=7.13 ms
64 bytes from 192.168.1.44: icmp_seq=9 ttl=64 time=7.15 ms
64 bytes from 192.168.1.44: icmp_seq=10 ttl=64 time=7.35 ms
^C
--- 192.168.1.44 ping statistics ---
10 packets transmitted, 10 received, 0% packet loss, time 9011ms
rtt min/avg/max/mdev = 7.105/7.269/7.443/0.145 ms

$ ping -i 0.5 -s 1400 192.168.1.44
PING 192.168.1.44 (192.168.1.44) 1400(1428) bytes of data.
1408 bytes from 192.168.1.44: icmp_seq=1 ttl=64 time=46.4 ms
1408 bytes from 192.168.1.44: icmp_seq=2 ttl=64 time=46.4 ms
1408 bytes from 192.168.1.44: icmp_seq=3 ttl=64 time=46.4 ms
1408 bytes from 192.168.1.44: icmp_seq=5 ttl=64 time=46.3 ms
1408 bytes from 192.168.1.44: icmp_seq=6 ttl=64 time=46.3 ms
1408 bytes from 192.168.1.44: icmp_seq=7 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=8 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=9 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=10 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=11 ttl=64 time=46.2 ms
1408 bytes from 192.168.1.44: icmp_seq=12 ttl=64 time=46.4 ms
^C
--- 192.168.1.44 ping statistics ---
12 packets transmitted, 11 received, 8.33333% packet loss, time 5509ms
rtt min/avg/max/mdev = 46.183/46.289/46.426/0.095 ms

$ telnet 192.168.1.44
Trying 192.168.1.44...
Connected to 192.168.1.44.
Escape character is '^]'.
login: root

# urlget http://some.external.ip.address
Host: some.external.ip.address
Port: 0
User: 
Pass: 
Path: /USER=root
Type: i
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL was not found on this server.</p>
</body></html>
# ps
  PID   GRP  TTY USER STAT CSEG DSEG  HEAP   FREE   SIZE COMMAND
    1     0      root    S 3ed9 4084  3072   2014  15648 /bin/init 
   12    12    1 root    S 4b11 564d  1166   6746  64560 -/bin/sh 
   19    18      root    R 8c35 8931     0   1976  12336 telnetd 
   15    15      root    S 5ad4 79df  9376  26402  75456 ktcp -b -d -m 1500 -p eth -l /dev/eth 192.168.1.44 192.168.1.254 255.255.255.0 
   18    18      root    S 8c35 87ae     0   2006  12336 telnetd 
   20    20   p0 root    S 4b11 8ec8  1166   6818  64560 -/bin/sh 
   23    20   p0 root    R 8ab4 4732  1024   6428  18592 ps 
# meminfo
  HEAP   TYPE  SIZE    SEG   TYPE    SIZE  CNT
  cdbe   SEG     16   1ed9   BUF    65536    1
  cdda   SEG     16   2ed9   BUF    65536    1
  cdf6   SEG     16   3ed9   CSEG    6832    1
  ce12   SEG     16   4084   DSEG    8816    1
  ce2e   SEG     16   42ab   free   18544    0
  ce4a   TTY    128
  ced6   TTY     64
  cf22   SEG     16   5f71   free  108256    0
  cf3e   SEG     16   79df   DSEG   56560    1
  cf5a   SEG     16   4b11   CSEG   46016    2
  cf76   SEG     16   564d   DSEG   18544    1
  cf92   SEG     16   87ae   DSEG    6192    1
  cfae   SEG     16   5ad4   CSEG   18896    1
  cfca   SEG     16   8c35   CSEG    6144    2
  cfe6   SEG     16   8db5   free    4400    0
  d002   SEG     16   8ec8   DSEG   18544    1
  d01e   SEG     16   8ab4   CSEG    4544    1
  d03a   SEG     16   8931   DSEG    6192    1
  d056   SEG     16   4a27   free    3744    0
  d072   TTY     64
  d0be   TTY    256
  d1ca   SEG     16   9ef1   free    4336    0
  d1e6   SEG     16   4732   DSEG   12112    1
  d202   SEG     16   934f   free   47648    0
  d21e   SEG     16   8bd0   free    1616    0
  d23a   free 11718
  Total heap  12878     Total mem  529008
  Memory usage  516KB total,  332KB used,  184KB free
# ps
  PID   GRP  TTY USER STAT CSEG DSEG  HEAP   FREE   SIZE COMMAND
    1     0      root    S 3ed9 4084  3072   2014  15648 /bin/init 
   12    12    1 root    S 4b11 564d  1166   6746  64560 -/bin/sh 
   19    18      root    R 8c35 8931     0   1976  12336 telnetd 
   15    15      root    S 5ad4 79df  9376  26402  75456 ktcp -b -d -m 1500 -p eth -l /dev/eth 192.168.1.44 192.168.1.254 255.255.255.0 
   18    18      root    S 8c35 87ae     0   2006  12336 telnetd 
   20    20   p0 root    S 4b11 8ec8  1166   6818  64560 -/bin/sh 
   26    26   S0 root    S 4b11 934f  1166   6746  64560 -/bin/sh 
   28    20   p0 root    R 8ab4 42ab  1024   6428  18592 ps 
# dd if=/dev/fd0 of=/dev/null
1440+0 records in
1440+0 records out
737280 bytes (720 KiB) copied
#
```

While doing `dd` I was pinging ELKS with `ping -i 0.5 -s 1400`, as a result, I've observed following on the ELKS console:

```
eth: overflow
eth: overflow
retrans retry #1 rto 4 mem 23
retrans retry #2 rto 8 mem 23
eth: overflow
```

At the same time, the serial console (enabled at `init 3` step) remained responsible. I didn't observe any problems after removing all occurences of the calls to `clr_irq()`/`set_irq()` from `wd.c` code.

pawosm-arm:
Ah, I didn't paste `init 3` step above...

```
# init 3
# ps
  PID   GRP  TTY USER STAT CSEG DSEG  HEAP   FREE   SIZE COMMAND
    1     0      root    S 3ed9 4084  3072   2014  15648 /bin/init 
   12    12    1 root    S 4b11 564d  1166   6746  64560 -/bin/sh 
   19    18      root    R 8c35 8931     0   1976  12336 telnetd 
   15    15      root    S 5ad4 79df  9376  26402  75456 ktcp -b -d -m 1500 -p eth -l /dev/eth 192.168.1.44 192.168.1.254 255.255.255.0 
   18    18      root    S 8c35 87ae     0   2006  12336 telnetd 
   20    20   p0 root    S 4b11 8ec8  1166   6818  64560 -/bin/sh 
   26    26   S0 root    S 4b11 934f  1166   6746  64560 -/bin/sh 
   28    20   p0 root    R 8ab4 42ab  1024   6428  18592 ps
```

Mellvik:
Thanks @pawosm-arm,
this is good news. It seems - as @ghaerr alluded to - that there is no need to protect any part of the wd driver from interrupts.

I'm hoping the same turns out to be the case for the ne2k driver!!

6 ms ping is OK. we're noe down to 2.someting on the ne2k/16bit/386 20MHz, and that's probably as fast as it gets. Maybe just below 2ms when the revised buffer handling is ready.

We're getting there!

—Mellvik

ghaerr:
> As to interrupts, my recommendation is to keep them on all the time and see what happens. Then put them back in and experiment with stepwise reduced scope.

> that there is no need to protect any part of the wd driver from interrupts.

It is important to understand exactly what context kernel drivers run in, or we'll never get this right. We need to identify all specific code that needs to be protected, then test against that. Throwing in cli/sti pairs needlessly usually harms the system.

The kernel is particularly well designed in the way it allows, or does not allow, interrupted execution of code. Concentrating only on kernel code for now, a kernel routine (including all driver code) may be interrupted, but the interrupted code will ALWAYS be resumed, kernel code will never be "switched-out" and another process/task executed. The ONLY code that can interrupt a kernel routine is a kernel interrupt handler. When that handler completes, we are guaranteed to immediately return to our original driver code. The interrupt routine always executes with interrupts enabled. We can ignore nested interrupts for this discussion.

Thus, we need to start by looking very hard at the WD interrupt handler. I have finally fully reviewed it. Since its written in C, it is easy to understand. I particularly like the use of "OR-ed" commands sent to the NIC command register, which allows for easy inspection of intention.

The WD interrupt handler comprises the two routines wd_int and wd_int_stat. They touch the NIC chip in a four places: select page 0, read ISR, ack ISR, and select page 0 | start. 

It seems to me the only real stateful portion of the NIC chip is the page select, which uses the upper two bits of each command sent to the NIC. Thus, any driver code that requires reading page 1 needs to be protected from interruption by the driver's own interrupt handler. Looking for any code that uses E8390_PAGE1, I see they are all properly protected with cli/sti pairs. All other cli/sti pairs are unneeded. Go ahead, take a look and search for PAGE throughout the driver.

I will submit a PR with the changes I think should be made. Hopefully they are correct; with cli/sti use, you have to have a theory; proof is thorough testing. The test process should concentrate on when the page is set to 1, which appears to be only in init (which can be basically ignored), and wd_rx_stat, which is only called when a process (ktcp) is either in select waiting on a packet to arrive, or in read, where it is double-checked again. I just noticed that the WD wd_read routine is coded differently than the NE2K routine in the case of a packet not being available after select and returns 0 in that rare case, which is not correct (small issue for now).

Thank you!

ghaerr:
> I just noticed that the WD wd_read routine is coded differently than the NE2K routine in the case of a packet not being available after select and returns 0 in that rare case, which is not correct (small issue for now).

Sorry - my error. It is coded correctly.

Mellvik:
> 27. sep. 2020 kl. 15:57 skrev Gregory Haerr <notifications@github.com>:
> 
> 
> As to interrupts, my recommendation is to keep them on all the time and see what happens. Then put them back in and experiment with stepwise reduced scope.
> 
> that there is no need to protect any part of the wd driver from interrupts.
> 
> It is important to understand exactly what context kernel drivers run in, or we'll never get this right. We need to identify all specific code that needs to be protected, then test against that. Throwing in cli/sti pairs needlessly usually harms the system.
> 
@ghaerr - this is what I'm saying, keep interrupts on alle the time, don't block them.


> The WD interrupt handler comprises the two routines wd_int and wd_int_stat. They touch the NIC chip in a four places: select page 0, read ISR, ack ISR, and select page 0 | start.
> 
> It seems to me the only real stateful portion of the NIC chip is the page select, which uses the upper two bits of each command sent to the NIC. Thus, any driver code that requires reading page 1 needs to be protected from interruption by the driver's own interrupt handler. Looking for any code that uses E8390_PAGE1, I see they are all properly protected with cli/sti pairs. All other cli/sti pairs are unneeded. Go ahead, take a look and search for PAGE throughout the driver.
> 
> 
I have already made such changes to the ne2k driver, which is the reason for my observations earlier today, that we probably don't need any 'interrupt protection' in any of these drivers.
Like I also said, this has not been fully tested on the case of the ne2k driver, but I'm optimistic about the prospect of just leaving interrupts on all the time. My test case is to bombard the system with serial interrupts while the net is heavily loaded.

—Mellvik

pawosm-arm:
> My test case is to bombard the system with serial interrupts while the net is heavily loaded.

Somehow I can't build a mental link between serial and network interrupts, these are different IRQ lines...

Mellvik:
@pawosm-arm,

That's quite OK, there is no such link. We know per @ghaerr's previous explanations that the read/writes in the driver are atomic, and there is really no chance that it can step on itself. I was for a while under the impression that there may be concurrent reads & writes active, but even that seems to have been ruled out (right, @ghaerr?). 

The point is to stress test the driver, now that the cli/sti pairs no longer protect against disturbances. You've done this before using ping and floppy access. With interrupt protection gone, the wd driver may be interrupted anywhere and the testing will reveal possible time & other sensitivities. AS before, the floppy read/writes are the toughest because they block for long periods,  while the serial interrupts represent a different type of load.


—Mellvik

ghaerr:
> We know per @ghaerr's previous explanations that the read/writes in the driver are atomic,

When you say that the read/writes are atomic, do you mean "will the kernel interrupt the wd_read routine to go and execute the wd_write routine for any reason"? No - it will never do that. Will the kernel interrupt the wd_read routine to execute an ethernet device interrupt handler routine? Absolutely, whatever code is being executed will be immediately interrupted.

>  there is really no chance that it can step on itself

Completely incorrect, as explained in https://github.com/jbruchon/elks/pull/769#issuecomment-699683662.
The driver is the only code that will step on itself, and it will always be the interrupt handler routine doing the stepping.

> there may be concurrent reads & writes active, but even that seems to have been ruled out

Correct. The kernel is atomic from the perspective of any system call: a process (always ktcp in this case) will never be rescheduled during a system call, only after the system call. Don't confuse this with the kernel being able to be interrupted, and an interrupt handler executed, that happens all the time.

> With interrupt protection gone, the wd driver may be interrupted anywhere and the testing will reveal possible time & other sensitivities.

The critical sections always need protection. Any driver may be interrupted anywhere, anytime, except for in its protected critical sections.

Perhaps what may be confusing here is that the ELKS network drivers use their interrupt handler to read a page 1 status register, and then wakeup either read- or write- waiting tasks from select, which later run and execute the read or write routine, where the data is finally transferred by those routines, outside of any interrupt handler. With the serial driver, received data is always read into a TTY queue during the interrupt by the interrupt handler; it then wakes up the waiting process which later reads from the TTY queue. (Prior to a couple months ago, the TTY queue was not protected by cli/sti pairs - and that meant that if a process happened to be reading from the TTY queue at the time it was interrupted by the serial interrupt handler which then also inserted characters into the same queue - the queue became corrupted and serial data was lost. It took a while to find that one, until I realized that the serial routine's interrupt handler steps on its own read routine's data structures.)

Here is an excerpt from "Understanding the Linux Kernel, 1st Edition" by O'Reilly. (Worth reading, available as download on the net). ELKS behaves identically to early Linux in this respect:

11.2.1 Nonpreemptability of Processes in Kernel Mode

As already pointed out, the Linux kernel is not preemptive, that is, a running process cannot be preempted (replaced by a higher-priority process) while it remains in Kernel Mode. In particular, the following assertions always hold in Linux:

• No process running in Kernel Mode may be replaced by another process, except when the former voluntarily relinquishes control of the CPU.[1]

[1] Of course, all context switches are performed in Kernel Mode. However, a context switch may occur only when the current process is going to return in User Mode.

• Interrupt or exception handling can interrupt a process running in Kernel Mode; however, when the interrupt handler terminates, the kernel control path of the process is resumed.

• A kernel control path performing interrupt or exception handling can be interrupted only by another control path performing interrupt or exception handling.

Mellvik:
@ghaerr,
I don't know what triggered this reaction, and you're right, we've been down this road before, there is no reason to go there again. We're essentially saying the same thing -- short version, long version, different words and from different angles. Taking something out of context and proclaiming it 'completely incorrect' is not contributing to a fruitful discussion.

I believe my advice on how to determine the critical sections of the driver is useful, it certainly has been to me. 

--Mellvik

Mellvik:
> It seems to me the only real stateful portion of the NIC chip is the page select, which uses the upper two bits of each command sent to the NIC. Thus, any driver code that requires reading page 1 needs to be protected from interruption by the driver's own interrupt handler. Looking for any code that uses E8390_PAGE1, I see they are all properly protected with cli/sti pairs. All other cli/sti pairs are unneeded. Go ahead, take a look and search for PAGE throughout the driver.
> 
> 
> 
This observation is correct, and I've proposed a complete elimination of the rx_stat call before. I've been using it on the ne2k driver for some time, it's solid.
It may be what's needed to remove the remaining interrupt blocking from the net drivers.

Not having the somewhat sensitive I/O loops should make eliminating cli/sti pairs easier in the wd/driver than in the ne2k.

--Mellvik

PR #768, opened 2020-09-26
[libc] Update library routines & crt0 to really work with segelf relocations in new toolchain

tkchia:


tkchia:
I am in the process of ironing out any issues in using `elks-libc` to build ELKS applications that use the medium memory model.

ghaerr:
Hello @tkchia,

These look fine, although I have a few questions, so that I can keep up with you:

- I presume the call-cvt.h change is to emit the second required (R_386_SEG16) relocation required by the SEGELF ABI. Does the assembler emit the first required relocation (R_386_SUB16) automatically on the lcall?
- What does the `as` --32-segelf option do? Just turn on the __IA16_ABI_SEGELF define?
- It looks like elf2elks tries to display an error message showing at least the section the bad addend is in. I'm guessing there's no easy way for ld.gold to catch that prior?
- I'm not following the BSS change in elf2elks. Will that affect the kernel build? I just found out from @pawosm-arm that the far text kernel doesn't actually boot on real hardware. I only tested on QEMU, so I'm going to see what happens on my real hardware systems.

Thank you!

tkchia:
Hello @ghaerr,

> * I presume the call-cvt.h change is to emit the second required (R_386_SEG16) relocation required by the SEGELF ABI. Does the assembler emit the first required relocation (R_386_SUB16) automatically on the lcall?

Yes.  I have not yet found out how to get the GNU assembler directly accept both the segment and offset for a non-absolute immediate `lcall`.  I need to look more closely at the GNU `as` code to see how to do this.

> * What does the as --32-segelf option do? Just turn on the __IA16_ABI_SEGELF define?

It also makes the assembler automagically generate the _x_`!` segment base symbols and also spew out the `R_386_SUB16` relocations to subtract them from operands.

> * It looks like elf2elks tries to display an error message showing at least the section the bad addend is in. I'm guessing there's no easy way for ld.gold to catch that prior?

That is less straightforward, and will probably mean that I need to inform `ld.gold` that there can only be three IA-16 segments --- text, far text, and data + BSS --- in our particular use case.

> * I'm not following the BSS change in elf2elks. Will that affect the kernel build? I just found out from @pawosm-arm that the far text kernel doesn't actually boot on real hardware. I only tested on QEMU, so I'm going to see what happens on my real hardware systems.

I found that, in some user programs, the BSS section for some reason did not immediately follow the data section, i.e. there was a small gap between the two.  Previously `elf2elks` would error out if it spotted this.

I would like to know what exactly was happening during @pawosm-arm's boot attempt.

Thank you!

pawosm-arm:
> I would like to know what exactly was happening during @pawosm-arm's boot attempt.

It just prints this:

```
t00C0 f0DE0 d0F02 t
```

tkchia:
Hello @pawosm-arm,

> It just prints this:
> ```
> t00C0 f0DE0 d0F02 t
> ```
Erm... can you upload a copy of the `elks/arch/i86/boot/system` file so we can take a look?  Thank you!

tkchia:
Hello @pawosm-arm,

Also, perhaps a copy of the disk image (`image/fd*.bin`) will be useful too.  I am thinking of dumping the image into [PCjs](https://www.pcjs.org/) to see if I can spot anything useful.

Thank you!

pawosm-arm:
@tkchia although since ELKS can be booted from CF card I'm not building floppy images that often, I've built one for you. It also fails to boot and it prints:

```
t00C0 f0DDE d0F00 t
```
Attached are gzipped binary images that you requested:
[system.gz](https://github.com/jbruchon/elks/files/5288170/system.gz)
[fd720.bin.gz](https://github.com/jbruchon/elks/files/5288171/fd720.bin.gz)
Aaargh, now I realised, 720k floppies were not known at the time PCJr was released. I'll try to build 360kB image for you.

pawosm-arm:
Here it is:
[fd360.bin.gz](https://github.com/jbruchon/elks/files/5288179/fd360.bin.gz)
The `system.gz` file has not change after rebuild.

tkchia:
Hello @pawosm-arm,

Thanks!  Let me see what is going on...

ghaerr:
> t00C0 f0DE0 d0F02 t

This means text was copied to segment 00C0, fartext to 0DE0, and data at 0F02. It did one text relocation and appears to have crashed.

On my system (including the latest commit PR #767), running under QEMU, the .fartext and .data get relocated to different addresses, which could be the result of a different .config file. My addresses are t00C0 f0E27 d0F53. 

Two possibilities are: 1) the particular config that is being built by @pawosm-arm is fartexzt buggy under QEMU and real hardware, but has never been tested; and 2) possibly post-8086 instructions were coded into the relocation routine. Is there a way to tell `ia16-elf-as` to assemble only 8086 instructions, no 80186 etc?

@pawosm-arm, can you upload your config file?

ghaerr (approved):

pawosm-arm:
@pawosm-arm, can you upload your config file?

Sure, it's here:
[config.gz](https://github.com/jbruchon/elks/files/5288341/config.gz)

tkchia:
Hello @ghaerr,

> 2) possibly post-8086 instructions were coded into the relocation routine. 

Yes --- and PCjs comes to the rescue again. :)  It turns out that some conditional jumps in `setup.S` were assembled not as short jumps, but as near jumps, which only exist on the 286 and above.  On the instruction sequence
```
        cmp $80,%ax        // R_SEGWORD
        jnz sys_hdr_bad
```
PCjs showed (and ran) the second instruction as a `pop %cs`.

> Is there a way to tell ia16-elf-as to assemble only 8086 instructions, no 80186 etc?

You can use `.arch i8086` or `.arch i8086, nojumps`.  This pseudo-op is also [part of mainline Binutils](https://sourceware.org/binutils/docs/as/i386_002dArch.html).

Thank you!

tkchia:
My proposed patch:
```
diff --git a/elks/arch/i86/boot/setup.S b/elks/arch/i86/boot/setup.S
index cbd09cb5..8c42be9b 100644
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -118,6 +118,7 @@
   SETUPSEG = CONFIG_ROM_SETUP_CODE
 #endif
 
+.arch i8086, nojumps
 .code16
 
 .text
@@ -724,7 +725,7 @@ relocat:
 	mov (%si),%di      // get r_vaddr
 	mov 6(%si),%ax     // get r_type
 	cmp $80,%ax        // R_SEGWORD
-	jnz sys_hdr_bad
+	jnz 9f
 	mov 4(%si),%ax     // get r_symndx
 	cmp $-2,%ax        // S_TEXT
 	jnz 1f
@@ -739,7 +740,7 @@ relocat:
 	jmp 3f
 2:
 	cmp $-3,%ax        // S_DATA
-	jnz sys_hdr_bad
+	jnz 9f
 
 	mov -22(%bp),%ax   // kernel .data segment
 3:
@@ -764,6 +765,8 @@ relocat:
 #endif
 	mov %ax,%es:(%di)
 	ret
+9:
+	jmp sys_hdr_bad
 
 // Add AX to DS:SI and normalize segment
 add_ptr:
```

ghaerr:
Hello @tkchia,

Thanks for debugging @pawosm-arm's crash and finding my relocation code problem!

Please submit a PR for your proposed patch. Your discovery also answers a question I had when I was looking at some what I thought strange code in setup, where jumps around jumps were taken. Now I understand that 8086s only support short jumps, good to know.

> .arch i8086, nojumps

This should probably also be placed in other important .S/.s files, or their headers. Do we need to make a list?

Thank you!

tkchia:
Hello @ghaerr,

> Please submit a PR for your proposed patch.

OK, thanks!

> Now I understand that 8086s only support short jumps, good to know.

Well, short conditional jumps, to be precise. :-)

> This should probably also be placed in other important .S/.s files, or their headers. Do we need to make a list?

Unless I am mistaken I do not think that there are that many `.S` or `.s` files in the source tree.  But a checklist might be useful. :-) :-)

Thank you!

tkchia:
> > which only exist on the 286 and above.

(In fact, if I remember correctly now, the near conditional jumps were introduced only with the 80**3**86.  In any case, we should avoid them.)

Thank you!

PR #769, opened 2020-09-27
[kernel] Suggestion for cli/sti for WD network driver

ghaerr:
Hello @pawosm-arm and @Mellvik,

Here is my proposal for interrupt protection in the WD driver, as discussed in https://github.com/jbruchon/elks/pull/767#issuecomment-699638919.

### DIFF
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -247,7 +247,6 @@ static int wd_pack_get(char *data, size_t len)
 	unsigned char this_frame;
 	int res = -EIO;
 
-	clr_irq();
 	do {
 		/* Remove one frame from the ring. */
 		/* Boundary is always a page behind. */
@@ -280,7 +279,6 @@ static int wd_pack_get(char *data, size_t len)
 		}
 		OUTB(current_rx_page - 1U, WD_8390_PORT + EN0_BOUNDARY);
 	} while (0);
-	set_irq();
 	return res;
 }
 
@@ -314,7 +312,6 @@ static size_t wd_read(struct inode * inode, struct file * filp,
 
 static size_t wd_pack_put(char *data, size_t len)
 {
-	clr_irq();
 	do {
 		if (len > MAX_PACKET_ETH)
 			len = MAX_PACKET_ETH;
@@ -333,7 +330,6 @@ static size_t wd_pack_put(char *data, size_t len)
 		OUTB(E8390_NODMA | E8390_TRANS | E8390_START,
 			WD_8390_PORT + E8390_CMD);
 	} while (0);
-	set_irq();
 	return len;
 }
 


pawosm-arm:
> Here is my proposal for interrupt protection in the WD driver

I wouldn't call it 'protection' anymore as it effectively removes it :)

Anyway, @Mellvik do you propose a testing strategy to try?

ghaerr:
> I wouldn't call it 'protection' anymore as it effectively removes it :)

LOL, yes, true for these two pairs proposed being removed. The remaining pairs seem to do a good job of protection though.

Did you remove just these two yesterday, or were all the cli/sti pairs removed in your last testing?

pawosm-arm:
I've removed them all

Mellvik:
I guess my previous email clarifies my opinions on this, including a test strategy.

It seems to me your testing @pawosm-arm confirms, or at leat indicates what we all have been saying in different ways, that interrupt blocking is not required.

This is all good news, although the proofing (heavy testing) remains.


—Mellvik

ghaerr:
> It seems to me your testing @pawosm-arm confirms, or at leat indicates what we all have been saying in different ways, that interrupt blocking is not required.

No - I disagree!

It seems to me we are going around on a big circle on this. I just gave a full explanation. In the WD driver, the critical section (example below) obviously must be protected (and is protected) from the WD interrupt handler:
```
static word_t wd_rx_stat(void)
{
    unsigned char rxing_page;

    clr_irq();
    /* Get the rx page (incoming packet pointer). */
    OUTB(E8390_NODMA | E8390_PAGE1, WD_8390_PORT + E8390_CMD);    <----- window starts here (A)
    rxing_page = INB(WD_8390_PORT + EN1_CURPAG);
    OUTB(E8390_NODMA | E8390_PAGE0, WD_8390_PORT + E8390_CMD);   <---- window ends here (B)
    set_irq();
    return (current_rx_page == rxing_page) ? 0U : WD_STAT_RX;
}
```
When the wd_read routine executes, there is a VERY SMALL but nonetheless POSSIBLE way that the `rxing_page` variable above could be set incorrectly. The "window" for badness is if the WD interrupt occurs immediately after the first OUTB at (A). Look what the WD interrupt handler does:
```
static void wd_int(int irq, struct pt_regs * regs, void * dev_id)
{
    word_t stat;

    stat = wd_int_stat();  <---- first thing the handler does is call wd_int_stat
 ...
static word_t wd_int_stat(void)
{
    byte_t interrupts;
    word_t retval = 0U;

    /* Change to page 0 and read the intr status reg. */
    OUTB(E8390_NODMA | E8390_PAGE0, WD_8390_PORT + E8390_CMD);  <-- second thing the handler does 
                                                                                      is reset PAGE 0 (C)
...
```
If the WD interrupt handler interrupts directly after (A) (setting PAGE 1), but before (B) (reading the result), the CPU will execute (C), which resets the command register to PAGE 0 again, before reading the value properly in (B). Please let me know if you don't follow this.

Thus, a cli/sti pair must be placed around the "critical section", as they are now correctly. They are also placed correctly in the init routines, which could also be interrupted in the same way.

This PR removes the cli/sti pairs around the other code, which otherwise only serves to lock out other interrupts, which is a bad thing. That is why I am removing them. The reason they are a problem is that disabling interrupts for a long period while doing a 900 byte memcpy could result in a serial interrupt handler not being able to read a UART character before the next one shifts into the UART receiver register, resulting in a serial overrun.

> Somehow I can't build a mental link between serial and network interrupts, these are different IRQ lines...

Correct, as far as testing the network driver critical sections, which is what I was talking about doing. To test the network driver critical section, one must somehow get an interrupt generated between A and B, which will be extremely infrequent. I can only think of running continuous inbound packets for hours and see whether anything happens.

With regards to testing the system as a whole, testing serial I/O against the WD (or NE2K) driver makes good sense.
With the network driver being pushed hard, running serial IN (not OUT) would test whether the WD code locks out interrupts for too long of a period, such that the serial receive register is overwritten. This should be tested with FIFO OFF. With the two improper cli/sti pairs NOT removed from the non-critical sections of code, it could be the case that during long fmemcpy's the system has interrupts disabled and the UART loses data at very high baud rates.

I haven't analyzed the NE2K driver fully, it is harder to understand since it is written in assembly. The same principles apply: identify critical sections that could cause a problem by the drivers' own interrupt handler, protect them, unprotect non-critical sections, then test.

I hope this long explanation helps.

pawosm-arm:
@ghaerr first of all, I wasn't sure the implementation of wd_rx_stat() (a polling function used by `select()`) is correct, somehow I couldn't find similar code in the Linux 8390 driver. Maybe there's a cheaper way to do that? E.g. some atomic variable set by interrupt handler whenever new data arrived?

pawosm-arm:
> @ghaerr first of all, I wasn't sure the implementation of wd_rx_stat() (a polling function used by select()) is correct, somehow I couldn't find similar code in the Linux 8390 driver. Maybe there's a cheaper way to do that? E.g. some atomic variable set by interrupt handler whenever new data arrived?

Answering myself. Linux driver reads received data in a loop, so they don't really need to poll the device for the data packets available: once they are received, they're moved to memory buffers. ELKS driver is simpler, yet after reading one header and amount of data declared in this header, it does not check for possibility of the next header being available, this can be only figured out by polling which in turn needs cli/sti protection.

Mellvik:
@pawosm-arm,

the rx_stat routine is correct, and - like I've pointed out before, it's easily replaced by a simple status flag. I did that in the ne2k driver a while back and have been testing it for about a month. It works fine - and will be part of an upcoming PR.

—Mellvik

ghaerr:
Hello @pawosm-arm,

> I wasn't sure the implementation of wd_rx_stat() (a polling function used by select()) is correct

I missed your comment yesterday. I wanted to add the following: Yes, it's used as a polling function when called by select(), but it's also as a normal function call in wd_read(), which is required to force the process to wait for a packet, in the case that a read system call is made without using select. This will never happen with the way ktcp is written, but still required.

> this can be only figured out by polling which in turn needs cli/sti protection.

So, it's not really the "polling" by select or read, but the fact that the chip command register needs to be set to page 0 at interrupt time that forces the need for cli/sti protection. Thus, since wd_rx_stat requires setting page 1, it needs to be protected from interruption by wd_int_stat, which can execute at any time.

It's not a big deal at all having a few instructions protected with cli/sti pairs, and not necessarily worth rewriting lots of code to eliminate them, if the critical section is quite small. My concern in this PR was eliminating the fmemcpyb's from critical sections, as the time for their execution is too large for interrupts to be disabled.

@Mellvik, I notice that the current NE2K driver also has interrupts disabled during the entire programmed I/O (dma_read/dma_write) data copy, including a CPU spin-loop waiting for I/O to complete, labelled experimental. These cli/sti pairs should be removed in the final version, as interrupts are disabled for long periods, and IMO not needed.


> ELKS driver is simpler, yet after reading one header and amount of data declared in this header, it does not check for possibility of the next header being available

Yes. ELKS just doesn't have the storage space to read all packets available, so instead of reading all the packets at interrupt time, like Linux does, ELKS reads one packet at a time only when a packet is known available and a process has executed the associated read system call. So, this would be effectively "polling" for packets.


Thank you!

Mellvik:
> 
> @Mellvik <https://github.com/Mellvik>, I notice that the current NE2K driver also has interrupts disabled during the entire programmed I/O (dma_read/dma_write) data copy, including a CPU spin-loop waiting for I/O to complete, labelled experimental. These cli/sti pairs should be removed in the final version, as interrupts are disabled for long periods, and IMO not needed.
> 
> 
Indeed @ghaerr, this is what I've mentioned a couple of times over the past days. I've been testing w/o interrupt blocking at all for quite some time, It's very encouraging — but there is some heavy beating remaining, along with the far buffer testing that I hope to get to today.

—Mellvik

Mellvik:
@ghaerr,
related question - now that the far buffer usage in the network drivers is working, is there a point in keeping the 'old way' as an ifdef in the code?

Better to just clean out and move on?

—Mellvik


> 28. sep. 2020 kl. 18:29 skrev Helge Skrivervik <helge@mymayday.com>:
> 
>> 
>> @Mellvik <https://github.com/Mellvik>, I notice that the current NE2K driver also has interrupts disabled during the entire programmed I/O (dma_read/dma_write) data copy, including a CPU spin-loop waiting for I/O to complete, labelled experimental. These cli/sti pairs should be removed in the final version, as interrupts are disabled for long periods, and IMO not needed.
>> 
>> 
> Indeed @ghaerr, this is what I've mentioned a couple of times over the past days. I've been testing w/o interrupt blocking at all for quite some time, It's very encouraging — but there is some heavy beating remaining, along with the far buffer testing that I hope to get to today.
> 
> —Mellvik

ghaerr:
> Better to just clean out and move on?

Yep!

Same goes for other cruft and commented out lines; one can always look at GitHub history.

PR #771, opened 2020-09-28
[kernel][fat] Move FAT filesystem code to far text section

ghaerr:
This PR moves the FAT filesystem code out of the kernel .text section, into a far section named .fartext.fatfs.
System tested with various FAT filesystem mounts and seems to work well. This is not yet as efficient as possible, but the kernel is now at 48k code, with 11k far text. This is now a considerable savings and provides much more breathing room for kernel expansion.

It is intended to commit this PR, but I desire to show and get comments on various code efficiency issues for a short period.

The FAT (and other) filesystem function entry points continue to be near text. More .text space could be saved by moving to far function pointers in the various kernel file_ops structures in the future, but this has too many ramifications for now.

The FAT and INIT far text code continues to be selectable using Far Text Kernel config option. The FAT section can be turned off individually in the header file linuxmt/msdos_fs.h, if desired.

### DIFF
--- a/elks/fs/msdos/dir.c
+++ b/elks/fs/msdos/dir.c
@@ -68,7 +68,7 @@ struct inode_operations msdos_dir_inode_operations = {
 
 /*@+type@*/
 
-static int
+static int FATPROC
 unicode_to_ascii(char *ascii, unsigned char *uni)
 {
 	char *op = ascii;
@@ -86,7 +86,7 @@ unicode_to_ascii(char *ascii, unsigned char *uni)
  *	returns 1 on short filename
  *	returns 2 on long filename
  */
-int msdos_get_entry_long(
+int FATPROC msdos_get_entry_long(
 	struct inode *dir,		/* scan this directory*/
 	off_t *pos,				/* starting at this offset*/
 	struct buffer_head **bh,/* using this buffer*/
--- a/elks/fs/msdos/fat.c
+++ b/elks/fs/msdos/fat.c
@@ -16,7 +16,7 @@ static struct fat_cache *fat_cache,cache[FAT_CACHE];
 /* Returns the this'th FAT entry, -1 if it is an end-of-file entry.
    If new_value is != -1, that FAT entry is replaced by it. */
 
-long fat_access(register struct super_block *sb,long this,long new_value)
+long FATPROC fat_access(register struct super_block *sb,long this,long new_value)
 {
 	struct buffer_head *bh,*bh2;
 	unsigned char *p_first,*p_last;
@@ -133,7 +133,7 @@ long fat_access(register struct super_block *sb,long this,long new_value)
 }
 
 
-void cache_init(void)
+void FATPROC cache_init(void)
 {
 	static int initialized = 0;
 	int count;
@@ -148,7 +148,7 @@ void cache_init(void)
 }
 
 
-void cache_lookup(struct inode *inode,long cluster,long *f_clu,long *d_clu)
+void FATPROC cache_lookup(struct inode *inode,long cluster,long *f_clu,long *d_clu)
 {
 	register struct fat_cache *walk;
 
@@ -168,7 +168,7 @@ printk("cache hit: %ld (%ld)\r\n",walk->file_cluster,*d_clu);
 
 
 #ifdef DEBUG
-static void list_cache(void)
+static void FATPROC list_cache(void)
 {
 	struct fat_cache *walk;
 
@@ -182,7 +182,7 @@ static void list_cache(void)
 #endif
 
 
-void cache_add(struct inode *inode,long f_clu,long d_clu)
+void FATPROC cache_add(struct inode *inode,long f_clu,long d_clu)
 {
 	register struct fat_cache *walk,*last;
 
@@ -223,7 +223,7 @@ list_cache();
 /* Cache invalidation occurs rarely, thus the LRU chain is not updated. It
    fixes itself after a while. */
 
-void cache_inval_inode(struct inode *inode)
+void FATPROC cache_inval_inode(struct inode *inode)
 {
 	register struct fat_cache *walk;
 
@@ -233,7 +233,7 @@ void cache_inval_inode(struct inode *inode)
 }
 
 
-void cache_inval_dev(int device)
+void FATPROC cache_inval_dev(int device)
 {
 	register struct fat_cache *walk;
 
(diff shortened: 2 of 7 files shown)


ghaerr:
Hello @tkchia,

I wanted to bring to your attention two issues I ran into while examining the compiler output for this PR. The good news is that everything is working quite well; this discussion would just be for future efficiency improvements, if possible.

In one of the FATPROC (far text) functions, `msdos_create_entry` in elks/fs/msdos/namei.c, I am looking at the code generated for the following line of code, which basically accesses a couple global longwords and does a divide and some arithmetic on the result before calling a far function:
```
struct timeval xtime;    (unsigned long variables)
jiff_t xtime_jiffies;    (unsigned long variable)
#define CURRENT_TIME (xtime.tv_sec + (jiffies - xtime_jiffies)/HZ)
...
date_unix2dos(CURRENT_TIME,&de->time,&de->date);
```
Code produced:
```
  75:   c7 47 1a 00 00          movw   $0x0,0x1a(%bx)
  7a:   a1 00 00                mov    0x0,%ax
            7b: R_386_16    jiffies
            7b: R_386_SUB16 jiffies!
  7d:   8b 16 02 00             mov    0x2,%dx
            7f: R_386_16    jiffies
            7f: R_386_SUB16 jiffies!
  81:   8d 4f 18                lea    0x18(%bx),%cx
  84:   51                      push   %cx
  85:   83 c3 16                add    $0x16,%bx
  88:   53                      push   %bx
  89:   8b 1e 00 00             mov    0x0,%bx
            8b: R_386_16    xtime_jiffies
            8b: R_386_SUB16 xtime_jiffies!
  8d:   8b 0e 02 00             mov    0x2,%cx
            8f: R_386_16    xtime_jiffies
            8f: R_386_SUB16 xtime_jiffies!
  91:   29 d8                   sub    %bx,%ax
  93:   19 ca                   sbb    %cx,%dx
  95:   ff 76 f6                pushw  -0xa(%bp)
  98:   b9 64 00                mov    $0x64,%cx
  9b:   51                      push   %cx
  9c:   52                      push   %dx
  9d:   50                      push   %ax
  9e:   bb 00 00                mov    $0x0,%bx
            9f: R_386_16    __udivsi3
            9f: R_386_SUB16 __udivsi3!
  a1:   9a 00 00 00 00          lcall  $0x0,$0x0
            a2: R_386_16    __ia16_far_call_thunk.bx.8
            a2: R_386_SUB16 __ia16_far_call_thunk.bx.8!
            a4: R_386_SEG16 __ia16_far_call_thunk.bx.8!
  a6:   83 c4 08                add    $0x8,%sp
  a9:   93                      xchg   %ax,%bx
  aa:   89 d1                   mov    %dx,%cx
  ac:   bf 00 00                mov    $0x0,%di
            ad: R_386_16    xtime
            ad: R_386_SUB16 xtime!
  af:   8b 15                   mov    (%di),%dx
  b1:   8b 45 02                mov    0x2(%di),%ax
  b4:   01 da                   add    %bx,%dx
  b6:   11 c1                   adc    %ax,%cx
  b8:   51                      push   %cx
  b9:   52                      push   %dx
  ba:   9a 00 00 00 00          lcall  $0x0,$0x0
            bb: R_386_16    date_unix2dos
            bb: R_386_SUB16 date_unix2dos!
            bd: R_386_SEG16 date_unix2dos!
```
First, a relatively small issue: there are two 8-byte relocations produced per shortword data access. While the code itself ends up being the same size, I wonder why .data relocations need to be produced at all, since the data segment value isn't being accessed. This only contributes to a larger a.out size, with final code size remaining the same. Or can the relocations be optimized out later by the linker? [EDIT: I recompiled near model and find that the same relocations are emitted. I think I understand now that new compiler always emits both relocations and the x! symbol is used to determine the segment; no R_386_SEG16 relocation is emitted into the final a.out, so this issue can be ignored.]

The second issue is inefficiency: it seems that a near __udivsi3 needs to be called to do the longword division, and it has to be thunked, rather than (somehow) perhaps having a version of it available in the far text segment. I'm wondering if there might be a tricky way we could perhaps forward-declare a far version of this (internal) function (perhaps name mangled to __udivs3i3_far if section not .text) and have it pulled from the kernel's library, as __udivsi3 was before. Another thought would be to always have far versions of these helper functions forward declared, available in the kernel library, and always called as far functions, even from near code.

What do you think?

Thank you!

tkchia:
Hello @ghaerr,

> I'm wondering if there might be a tricky way we could perhaps forward-declare a far version of this (internal) function (perhaps name mangled to __udivs3i3_far if section not .text) and have it pulled from the kernel's library, as __udivsi3 was before. Another thought would be to always have far versions of these helper functions forward declared, available in the kernel library, and always called as far functions, even from near code.

Well, either of your proposed schemes might be possible, although I expect the implementation will be quite messy.

GCC apparently has different internal mechanisms for dealing with normal function calls and dealing with such implicit library calls ("libcalls").  In particular libcalls do not have function prototypes, so it seems there is no neat way to tell GCC that e.g. `__udivsi3` is `__far` and has a non-default prototype, and calls to `__udivsi3` should behave accordingly.  There are probably some non-neat ways to do this, though. :-)

Thank you!

ghaerr:
Hi @tkchia,

Thank you for your response. I dug into this a week ago or so and saw that gcc is quite complicated in how various platforms register various libcall functions. In general, since things seem to be working well, this is a pretty low priority. I've tried to confine the far text sections of the kernel to non-time-sensitive routines like initialization and filesystem code for the time being, where the extra thunking time probably doesn't matter much.
```
 9e:   bb 00 00                mov    $0x0,%bx
            9f: R_386_16    __udivsi3
            9f: R_386_SUB16 __udivsi3!
  a1:   9a 00 00 00 00          lcall  $0x0,$0x0
            a2: R_386_16    __ia16_far_call_thunk.bx.8
            a2: R_386_SUB16 __ia16_far_call_thunk.bx.8!
            a4: R_386_SEG16 __ia16_far_call_thunk.bx.8!
  a6:   83 c4 08                add    $0x8,%sp
```
Looking at the code emitted again, I thought this morning of another possibility: doing all the magic in the thunking code that you have written. It seems that the thunking code is invoked by gcc, knowing that the __udivsi3 function is near, and that thunking is needed. Perhaps if the thunking code, using a list of predefined function names and a special switch to turn this on, (or just passing the list of special functions as parameters to gcc) would in the case of a match, instead of loading the address of __udivsi3 into BX and emitting the thunk code, just name-mangle it to __udivsi3_far and lcall it, instead of emitting the normal thunk. The thunking code would essentially handle the entire issue, specially.

I'm sure this is all way more complicated than I'm seeing. Is the thunk code all handled in a lisp-like language internal to gcc?

Thank you!

PR #772, opened 2020-10-05
[ktcp] TCP retransmit cleanup and fixes, socket write throttling

ghaerr:
Cleanup and fixes to ktcp retransmit routines. Should fix most issues in #755.

Fix major memory corruption bug in calling tcpcb_remove with bad parms due to lax function prototyping.
Limit retransmit waits to 1/4 second min, 4 seconds max, and 3 retransmits max.
Add experimental code to ktcp's tcpdev write handler to delay/throttle writes after SEND_WINDOW_MAX (1024) bytes unacked data outstanding.
Cleanup tcpdev handling for writing closed socket an retrans limit exceeded (returns -EPIPE and -ENOMEM).
Reduce httpd's buffer to 1024, fix file size bug, checks write return value.
Fine tuned ^P debug display.
Add comments in ports.h for IRQ selection.
Removed fartext relocation debug display at boot.

Tested on NE2K driver on real hardware.

@Mellvik - give this a try to test issues discussed in #755. I'm still seeing infrequent long loops when writes are delayed at times, with display "limiting write to max send window".
Also, the NE2K driver doesn't always work on first try, sometimes gets "NIC ring buffer overflow, skip all packets", and "RDC intr" quite a bit. Not sure why RDC intr is enabled, once it happens the driver is toast. I think the NE2K driver may still be buggy (at least on my faster 386).
I'm also having problems with telnet from Linux into ELKS - seems to work first time only with ^P debugging ON; timing related.
The send window size throttling can be changed or turned off entirely in tcpdev.c: line 346.

### DIFF
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -658,8 +658,8 @@ sys_no_far_text:
 	mov -12(%bp),%cx // text reloc size
 text_reloc:
 	jcxz 1f
-	mov $'t',%ax
-	call putc
+//	mov $'t',%ax
+//	call putc
 	mov -18(%bp),%ax // kernel .text segment
 	mov %ax,%es
 
@@ -672,8 +672,8 @@ text_reloc:
 	mov -14(%bp),%cx // far text reloc size
 ftext_reloc:
 	jcxz 2f
-	mov $'f',%ax
-	call putc
+//	mov $'f',%ax
+//	call putc
 	mov -20(%bp),%ax // kernel .fartext segment
 	mov %ax,%es
 
@@ -686,8 +686,8 @@ ftext_reloc:
 	mov -16(%bp),%cx // data reloc size
 data_reloc:
 	jcxz 3f
-	mov $'d',%ax
-	call putc
+//	mov $'d',%ax
+//	call putc
 	mov -22(%bp),%ax   // kernel .data segment
 	mov %ax,%es
 
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -8,6 +8,7 @@
 //-----------------------------------------------------------------------------
 
 #include "arch/ports.h"
+#include <arch/asm-offsets.h>
 
 	.code16
 
@@ -34,6 +35,7 @@ io_ne2k_dma_len1   = base+0x0A  // page 0 - write
 io_ne2k_dma_len2   = base+0x0B  // page 0 - write
 
 io_ne2k_rx_stat    = base+0x0C  // page 0 - read
+io_ne2k_tx_stat    = base+0x04  // page 0 - read
 
 io_ne2k_rx_conf    = base+0x0C  // page 0 - write
 io_ne2k_tx_conf    = base+0x0D  // page 0 - write
@@ -61,6 +63,7 @@ rx_last            = 0x80
 
 //-----------------------------------------------------------------------------
 	.data
+	.extern current
 
 _ne2k_next_pk:
 	.word 0	// being used as byte ...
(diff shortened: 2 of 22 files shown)


Mellvik:
Thanks @ghaerr, this is all good news.

> Cleanup and fixes to ktcp retransmit routines. Should fix most issues in #755 <https://github.com/jbruchon/elks/issues/755>.
> 
> Fix major memory corruption bug in calling tcpcb_remove with bad parms due to lax function prototyping.
> Limit retransmit waits to 1/4 second min, 4 seconds max, and 3 retransmits max.
> Add experimental code to ktcp's tcpdev write handler to delay/throttle writes after SEND_WINDOW_MAX (1024) bytes unacked data outstanding.
> Cleanup tcpdev handling for writing closed socket an retrans limit exceeded (returns -EPIPE and -ENOMEM).
> Reduce httpd's buffer to 1024, fix file size bug, checks write return value.
> Fine tuned ^P debug display.
> Add comments in ports.h for IRQ selection.
> Removed fartext relocation debug display at boot.
> 
> Tested on NE2K driver on real hardware.
> 
> @Mellvik <https://github.com/Mellvik> - give this a try to test issues discussed in #755 <https://github.com/jbruchon/elks/issues/755>.
> 
I’ll take another round of TCP testing today.
> I'm still seeing infrequent long loops when writes are delayed at times, with display "limiting write to max send window".
> Also, the NE2K driver doesn't always work on first try,
> 
Could you elaborate on this? What does ‘first try’ mean?
> sometimes gets "NIC ring buffer overflow, skip all packets", and "RDC intr" quite a bit. Not sure why RDC intr is enabled, once it happens the driver is toast. I think the NE2K driver may still be buggy (at least on my faster 386).
> 
Yes, there are a number of bugfixes in the upcoming patch. RDC is not enabled. What you’re seeing is that the RDC bit is set when an interrupt is being processed. This should not happen, that’s why the printk is there. The bug that cause this is that the Interrupt handler continues after processing an overflow interrupt. It should exit, since the overflow cause the NIC to be reset, there is nothing else to process. This is handled correctly by the updated driver.
> I'm also having problems with telnet from Linux into ELKS - seems to work first time only with ^P debugging ON; timing related.
> The send window size throttling can be changed or turned off entirely in tcpdev.c: line 346.
> 
> 
So, the first item on the list is a PR for the ne2k updates, later today.

—Mellvik

ghaerr:
@Mellvik,

I've pushed a fix that seems to work well, but it only works after the "first try". By first try, I mean immediately after booting, the NE2K driver will give a receive buffer overrun, skipping all packets, then an RDC error if a telnet connection from linux to ELKS is attempted. However, if ^P is selected prior (at login or shell prompt), which slows ktcp down with debug messages, then telnet connect, then ^P to turn off messages, the NE2K driver doesn't get an overrun and my latest push (which delays writes 10ms when the send window is full) works all the time.

I don't really know what the problem is connecting right after boot - whether its timing related (seems so), or the NE2K driver.

Mellvik:
@ghaerr,
This is interesting. Do you have a tcpdump trace of this?

It sounds hw related. Not ruling out driver  issues, but then again the driver has been seriously beaten up using this and one other card. Never seen a symptom like this. Sounds like there is still a interrupt problem. I would recommend testing on a compaq for reference.

The new version of the ne2k driver is ready, pending my fight with git fast-forward. 

--Mellvik

ghaerr:
I don't have a trace yet, still fooling around with timing related issues. If a single debug statement is left on, everything works. It is possible that the read buffer is in fact being overwritten due to ktcp not running... I'll try this with your new driver first.

When you have time, try this code with your new driver, and lets see what happens.

Mellvik:
@ghaerr,
if you enable just the printk @ line 168 of ne2k.c you’ll see exactly what’s going on.

—M

ghaerr:
Added @Mellvik's PR #773 to this commit to fix a merge conflict.

After this commit, the retransmit issues brought up in #755 should be tested. There are few retransmits anymore, and only because of actual floppy I/O speed, and I think the retransmit system is working well. I haven't tested the httpd issue directly, but httpd will be delayed 10ms when the send window exceeds 1024 bytes, which hopefully will solve the data transfer problem.

The RST issues in #755 have not been addressed yet.

PR #773, opened 2020-10-05
[ne2k] Enhancements and bugfixes

Mellvik:
Ne2k driver update

Bugfixes and enhancements:
Changing to a newer (and faster) physical interface and beating it up real hard with large back-to-back packets, turned up some new bugs - and opportunities for  enhancements:
- Sending packets back-to-back would sometimes interrupt the ongoing send and cause erroneous packets on the net. Added a check for transmit complete in the pkt_put routine.
- Occasional unexplainable interrupt status codes under heavy receive load were traced back to get_rx_stat frequently pulling the NIC for status info. Created a status flag for the state of the NIC ring buffer which keeps track of buffer empty/not empty, which in addition to eliminating the spurious problems, reduced code size and increased speed.
- Changed the overflow handling to never discard all packets. Instead - if the skip_packets variable is 0, the oldest packet is left in the buffer (if it's 1, the newest packet is retained). This eliminates the occasional problem i deveth_read when getting no data. Moving the overflow handler to the end of the pkt_get routine turned out to not work well. Thus the overflow handler continues as a true interrupt routine.
- Changed the interrupt handler in ne2k.c to return after processing an overflow. Since the overflow handler soft-resets the interface, there is nothing more to process, and the interrupt status bits are actually bogus.
- Changed the probe code to actually look for the presence of an interface. The probe tests for presence, not for signature, so many cards will be recognized. The actual litmus test for a ne2k card is in the get_HW_address routine, which will fail if it's not an ne2k. A 3COM 309 card will not be detected (which is good).
- Removed the code that set a default address if a hardware address is not found. This setup works fine with qemu. 
- Many optimizations - in particular removal of unneeded push/pop pairs.
- Removed local buffering, data is copied directly to/from the caller via far pointers.

Pending: 
- More testimng is to eliminate the sti/cli pairs in order to see if they're actually necessary.
- changes to take advantage of the debug on/off toggle.
- cleanups

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -8,6 +8,7 @@
 //-----------------------------------------------------------------------------
 
 #include "arch/ports.h"
+#include <arch/asm-offsets.h>
 
 	.code16
 
@@ -34,6 +35,7 @@ io_ne2k_dma_len1   = base+0x0A  // page 0 - write
 io_ne2k_dma_len2   = base+0x0B  // page 0 - write
 
 io_ne2k_rx_stat    = base+0x0C  // page 0 - read
+io_ne2k_tx_stat    = base+0x04  // page 0 - read
 
 io_ne2k_rx_conf    = base+0x0C  // page 0 - write
 io_ne2k_tx_conf    = base+0x0D  // page 0 - write
@@ -61,6 +63,7 @@ rx_last            = 0x80
 
 //-----------------------------------------------------------------------------
 	.data
+	.extern current
 
 _ne2k_next_pk:
 	.word 0	// being used as byte ...
@@ -68,6 +71,10 @@ _ne2k_next_pk:
 	.global _ne2k_skip_cnt
 _ne2k_skip_cnt:
 	.word 0	// # of packets to skip if buffer overrun, default is all (0)
+
+	.global _ne2k_has_data
+_ne2k_has_data:
+	.word 0 
 
 	.text
 
@@ -115,14 +122,11 @@ ems_loop:
 //-----------------------------------------------------------------------------
 // DMA initialization - Prepare for internal NIC DMA transfer
 //-----------------------------------------------------------------------------
-
+// Uses: DX, AX
 // BX : chip memory address (4000h...8000h)
 // CX : byte count
 
 dma_init:
-
-	push    %ax
-	push    %dx
 
 	// set DMA start address
 
@@ -144,8 +148,6 @@ dma_init:
 	mov     %ch,%al
 	out     %al,%dx
 
-	pop     %dx
-	pop     %ax
 	ret
 
 //-----------------------------------------------------------------------------
@@ -159,9 +161,9 @@ dma_init:
 
 dma_write:
 
-	push    %ax
 	push    %cx
-	push    %dx
+	push	%bx	// TODO check if this is required (2)
+	push	%ds
 	push    %si
 
 	cli		// Experimental
@@ -179,6 +181,8 @@ dma_write:
 
 	// I/O write loop
 
+	mov	current,%bx		// setup for far memory xfer
+	mov	TASK_USER_DS(%bx),%ds
 	mov     $io_ne2k_data_io,%dx
 	cld
 
@@ -198,14 +202,13 @@ check_dma_w:
 
 	mov     $0x40,%al       //clear DMA intr bit in ISR
 	out     %al,%dx
-	clc
 
 
 	sti		// Experimental
 	pop     %si
-	pop     %dx
+	pop	%ds
+	pop	%bx
 	pop     %cx
-	pop     %ax
 	ret
 
 #if 0
@@ -254,29 +257,36 @@ rlp_ret:
 	
 //-----------------------------------------------------------------------------
 // Read block from chip with internal DMA
-//
-// FIXME: The first read operation should get a full page (256 bytes)
-// instead of just the first 4 bytes. That way a single DMA read operation will cover
-// most incoming packets in interactive sessions.
 //-----------------------------------------------------------------------------
-
+//
 // BX    : chip memory to read from
 // CX    : byte count
 // ES:DI : host memory to write to
+// AL:	 : 0: buffer is local, <>0: buffer is far
 
 dma_read:
 
-	push    %ax
-	push    %cx
-	push    %dx
 	push    %di
-	push    %es  // compiler scratch
+	push    %es
+	push	%bx
+	push	%ax
 
 	cli		//Experimental
 	inc     %cx     // make byte count even
 	and     $0xfffe,%cx
 	call    dma_init
 	shr     %cx     // half -> word size transf
+
+	mov     %ds,%bx
+	mov     %bx,%es
+	pop	%ax
+	cmp	$0,%al		// Use local buffer if zero
+	jz	buf_local
+	mov	current,%bx	// Normal: read directly into the (far) buffer
+	mov	TASK_USER_DS(%bx),%es
+buf_local:
+	pop	%bx
+	push	%bx
 
 	// start DMA read
 
@@ -286,12 +296,8 @@ dma_read:
 
 	// I/O read loop
 
-	mov     %ds,%ax
-	mov     %ax,%es
-
 	mov     $io_ne2k_data_io,%dx
 	cld
-
 emr_loop:
 	in      %dx,%ax
 	stosw
@@ -305,15 +311,15 @@ check_dma_r:
 	test    $0x40,%al       // dma done?
 	jz      check_dma_r     // loop if not
 
+	//mov	$0xabcd,%ax	// TEST
+	//mov	%ax,10(%di)
 	mov     $0x40,%al       // reset ISR (RDC bit only)
 	out     %al,%dx
 	sti		//Experimental
 
+	pop	%bx
 	pop	%es
 	pop     %di
-	pop     %dx
-	pop     %cx
-	pop     %ax
 
 	ret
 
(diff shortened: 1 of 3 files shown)


ghaerr:
@Mellvik,

Since this PR has merge conflicts, I've pulled it down and manually merged it. Rather than fixing this PR, I will merge it with #772, along with the other TCP retransmit fixes. I found the retransmit problem mentioned in #772, and on my fast 386, it was because the ne2k_pack_put routine was not properly sending a second transmit packet when sending back-to-back packets. I solved the problem by adding a wait for TX completion in the ne2k_pack_put routine before returning, and now, combined with the retransmit and max send window size fixes in #772, the system works great.

There are no more startup issues, RDC errors, or receive buffer overflows. The system is now quite fast, and was tested with telnet's going in both directions displaying simultaneously with success.

After committing #772, I will close this PR, and you can pull a full source tree down for testing the issues you've brought up in #755.

Thank you!

Mellvik:
Thank you, @ghaerr - this enhancement was really too long in the making.

> 
>  I found the retransmit problem mentioned in #772 <https://github.com/jbruchon/elks/pull/772>, and on my fast 386, it was because the ne2k_pack_put routine was not properly sending a second transmit packet when sending back-to-back packets. I solved the problem by adding a wait for TX completion in the ne2k_pack_put routine before returning, and now, combined with the retransmit and max send window size fixes in #772 <https://github.com/jbruchon/elks/pull/772>, the system works great.
> 
Great. And interesting because I found exactly the same when I switched interfaces, and thought I had it fixed (as mentioned in the PR comment). We’re getting there.
> There are no more startup issues, RDC errors, or receive buffer overflows. The system is now quite fast, and was tested with telnet's going in both directions displaying simultaneously with success.
> 
> 
Indeed. While doing the final testing before submitting the PR, I sometimes got < 2ms on regular pings. That’s a 5-8% improvement over the best I’ve seen before, so the far buffer and the other enhancements ARE making a difference.
> After committing #772 <https://github.com/jbruchon/elks/pull/772>, I will close this PR, and you can pull a full source tree down for testing the issues you've brought up in #755 <https://github.com/jbruchon/elks/issues/755>.
> 
> 
I’m looking forward to a new round of TCP testing … thanks for the merge.

—Mellvik

Mellvik:
@ghaerr,

I've looked at the change you made to the send-packet function, and it got me thinking. From your description it sounds like the problem you're trying to solve is exactly the same as I thought I had solved with the exact same loop a couple of instructions above. What's in between is the actual sending of the packet on the wire.

What I'm seeing is that the loops cause delays which cure the symptoms, but not the problem. In my case I put in a faster interface and met problems. In your case you have a faster machine, and get problems. The problem being that we're seeing send errors.

The purpose of the first TX status test loop is to eliminate the chance that a packet gets on the wire before the previous has finished. Fair enough, it does that, but I don't think that could ever happen. The 2nd loop simply blocks the driver while data is going onto the wire.

Here's why the first loop propably has no effect other than a small delay: The sequence of events when sending is  
1. data gets transferred to the send buffer.
2. when complete, the put-data-on-the-wire command is given.
3. when put-data-on-wire is complete, the NIC interrupts.

So, when the interrupt comes, we know there is no tx in progress. Further, in select and write, we check that no transfer is in progress before any data is sent into the NIC tx buffer. Important since we have only one send buffer, and writing to it while sending will disturb the put-data-on-the-wire process.

Is this important? Depends on. We can leave it as it is now, under the motto 'if it ain't broken etc.', but that depends on the definition of 'broken'. Your fix turns the NIC effectively into a half duplex device: The driver loops while it could be reading incoming data. And - if we had 2 or more send buffers - it could have witten one buffer while the other was being transferred to the wire.

So the question is: Where's the bug? I suspect it's our reliance of the TXP bit in the Command register, and that we should solely depend on the TX status reg instead. I'm going to remove both loops in the packet_put routine, change all the tx status checks and see what happens. I'm also doing another run-through of potential concurrency-problems. The half-dupex effect of your fix will undoubtedly eliminate any concurrency issues, such as the ones I mentioned - reading data while sending on the wire.

Hopefully, we have another PR in the works.

-M

ghaerr:
I'm aware that the fix I hastily put in blocks the system while it could be doing other things. I was kind of desperate because I had finally figured out by looking at tcpdump that the system was losing the second send packet in the four-packet TCP sequence between systems when a character is typed in telnet. I also didn't know whether my other fixes worked, and then when the system started working, I realized that almost all the problems I was having were a result of the driver dropping a send packet. The good news is that I think the retransmit code is now working well.

I tried putting the check in the beginning of the ne2k_pack_put routine, thinking that perhaps a check should be made before copying data to the NIC buffer with dma_write; that had no effect (but could still be an issue?). Then I thought of putting it afterwards, just for safety, and it worked. So - it seems that the ne2k_pack_out routine itself may not be the whole problem.

If you get something that works on your system and you'd like me to try it on mine, just send me a patch file uploaded here, and I'll give it a shot.

In the meantime, I am very glad we have a (at least temporarily) working system!! Despite this issue, your new driver works much better than before, it seems :)

I need you to test the urlget/httpd part of the overall fix, even with the temporary driver. That's the only way I know the code changes I made actually work. 

Thank you!

Mellvik:
Thanks, @ghaerr - 
> I'm aware that the fix I hastily put in blocks the system while it could be doing other things. I was kind of desperate because I had finally figured out by looking at tcpdump that the system was losing the second send packet in the four-packet TCP sequence between systems when a character is typed in telnet. I also didn't know whether my other fixes worked, and then when the system started working, I realized that almost all the problems I was having were a result of the driver dropping a send packet. The good news is that I think the retransmit code is now working well.
> 
I agree with the priorities! ‘Works’ always beats ‘perfect’ when ‘works’ means ’now’. 

> I tried putting the check in the beginning of the ne2k_pack_put routine, thinking that perhaps a check should be made before copying data to the NIC buffer with dma_write; that had no effect (but could still be an issue?). Then I thought of putting it afterwards, just for safety, and it worked. So - it seems that the ne2k_pack_out routine itself may not be the whole problem.
> 
It’s not. And it sounds familiar –– I seem to have been through all these in-and-out tests more times than I remember :-), eventually finding a problem somewhere else and forgetting to get back and proceed with the original ’thread’...
> If you get something that works on your system and you'd like me to try it on mine, just send me a patch file uploaded here, and I'll give it a shot.
> 
Will do.  
> In the meantime, I am very glad we have a (at least temporarily) working system!! Despite this issue, your new driver works much better than before, it seems :)
> 
I really hope so - the long turnaround time between each compile when on physical drivers me crazy some times. Having urlget operational makes a hell of a difference though.
> I need you to test the urlget/httpd part of the overall fix, even with the temporary driver. That's the only way I know the code changes I made actually work.
> 
> 
I’ve probably mentioned this before, but here it is again - my favorite transmit test:

- create a link to a raw disk device in the http home directory on elks
	ln -s /dev/bda /var/www/xx
- curl from the linux host
	curl http://10.0.2.2/xx <http://10.0.2.2/xx> > bda.xx (or just /dev/null)
Keep watching via tcpdump
Then do logins, floppy access, disk access, pings and other stuff from other windows. Not having a physical floppy is a drag :-), but ping -i 0.2 -s 1300 does the trick too…

—M

ghaerr:
> I’ve probably mentioned this before, but here it is again - my favorite transmit test:

> - create a link to a raw disk device in the http home directory on elks
	ln -s /dev/bda /var/www/xx
> - curl from the linux host
	curl http://10.0.2.2/xx <http://10.0.2.2/xx> > bda.xx (or just /dev/null)
Keep watching via tcpdump
Then do logins, floppy access, disk access, pings and other stuff from other windows. Not having a physical floppy is a drag :-), but ping -i 0.2 -s 1300 does the trick too…

I have some good news @Mellvik:

Using the updated NE2K driver and the recently committed other networking code changes, running the above curl test, along with ping, logins, and other things results in file transfers continually working.

We are seeing "retrans retry" when telnetting in from Linux to ELKS during a large curl transfer, and even an occasional "retrans memory over limit", as well as an occasional "eth: NIC receive oflow", but both the curl transfer as well as the telnet session do no display errors, and continue working.

I had changed the retrans code to either return -ENOMEM to the calling process in the worst case (which will now return an error to httpd (not seen yet)), or to just free the retrans memory, but not reset the connection. It appears that these changes are working, and also have not caused the telnet or curl connection to stop working.

I haven't been able to test using inserted ARP packets directly. I'm also unclear as to exactly why a retrans retry is occurring every time telnet is used simultaneously with curl; they are happening at "rto 8", which means 1/2 second delay. This could just be delays as a result of normal interruption of curl transfer with disk I/O to run "ls -lR /" within telnet.

Also, typing ^C to curl seems to work well now. The previous commit included fixes for the memory corruption that occurred whenever a socket was closed, including for RST connections, and the "panic: write" error was changed to a "write to unknown socket" without affecting ktcp for other processes. It appears this is working well.

Looking forward to your testing!

Mellvik:
> Using the updated NE2K driver and the recently committed other networking code changes, running the above curl test, along with ping, logins, and other things results in file transfers continually working.
> 
Very encouraging indeed, @ghaerr. This is a new level of stability.
> We are seeing "retrans retry" when telnetting in from Linux to ELKS during a large curl transfer, and even an occasional "retrans memory over limit", as well as an occasional "eth: NIC receive oflow", but both the curl transfer as well as the telnet session do no display errors, and continue working.
> 
This must mean that there is reasonable error recovery and that it is actually working when pushed (at the tcp level) in addition to the low level being ok). Congratulations!!
> I haven't been able to test using inserted ARP packets directly. I'm also unclear as to exactly why a retrans retry is occurring every time telnet is used simultaneously with curl; they are happening at "rto 8", which means 1/2 second delay. This could just be delays as a result of normal interruption of curl transfer with disk I/O to run "ls -lR /" within telnet.
> 
We need to study this one carefully with tcpdump tracings, and it may be easier on a slower system (and it may go away on a slower system). The thing with ARP on my raspberry (maybe on linux in general) is that the ARP cache times out even when a connection is active, at least it does when the traffic is ICMP, so there will be arp traffic at least every minute. That is very useful for testing. I'll keep this one in mind when I get back to testing. That said, inserting ICMP echos should create almost the same type of disturbance once the traffic is flowing. Further, adding a 2nd and 3rd host on the segment to push the arp cache is useful. I have another raspi and a dos machine on the same segment.
> Also, typing ^C to curl seems to work well now. The previous commit included fixes for the memory corruption that occurred whenever a socket was closed, including for RST connections, and the "panic: write" error was changed to a "write to unknown socket" without affecting ktcp for other processes. It appears this is working well.
> 
Heck, this is a lot in one week. Now that I have compilation working again, it will be my pleasure to start a new round of tests...
> 
--Mellvik

PR #775, opened 2020-10-08
[kernel]][ktcp] Fine tune network buffer sizes for speed

ghaerr:
Fine tune the kernel and ktcp for faster networking speeds.
Achieves a maximum data size of 512 bytes/packet on telnet and httpd outbound transfers.
Fixes all issues in #755, with an emphasis for speed.

Testing on with @Mellvik's curl test, and 23.5M file shows the following results, almost halving the time to send the file:
```
Time    Speed    Packet Size
1:45    230k       128   (prior to this PR)
0:54    455k       512   (with this PR)
0:56    427k       512   (with this PR and ping -i 0.2 -s 1300 running)
```
For now, 512 bytes is the likely packet limit as PTY queue, tcpdev queues, kernel, ktcp and telnetd buffers all need to increase at the same time.

Tested on 386 desktop system. No system failures with simultaneous pings, inbound telnet and httpd transfer running.
Occasional retransmits and a few NIC receive buffer overruns seen, but all recovered with no data loss.
After testing, #755 can be closed if successful, unless RST issues seen.

Change PTY_OUTQSIZE = TDB_WRITE_MAX = telnetd buffer size.
Change PTY output queue size from 256 to 512: telnetd reads from pty out queue.
Change TDB max from 100 to 512: telnetd writes to ktcp and network.
Increase max retrans memory from 2048 to 4096: stops retrans limits exceeded on large httpd transfers.
Set receive window size < 255 when little space available.
Call wake_up once in pty_read for speed.

### DIFF
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -1178,7 +1178,7 @@ no_sig_mess:	.ascii	"No ELKS setup signature found ..."
 		.byte	0x00
 
 hello_mess:
-	.ascii "\r\nELKS Setup\r\n"
+	.ascii "\r\nELKS Setup "
 	.byte 0
 
 // variables in ROM are not very usefull
--- a/elks/arch/i86/drivers/char/pty.c
+++ b/elks/arch/i86/drivers/char/pty.c
@@ -93,9 +93,10 @@ size_t pty_read (struct inode *inode, struct file *file, char *data, size_t len)
 
 		put_user_char (tty_outproc (tty), (void *)(data++));
 		count++;
-		wake_up (&tty->outq.wait);  /* because tty_outproc does not */
 	}
 
+	if (count > 0)
+		 wake_up(&tty->outq.wait);  /* because ttyoutproc does not*/
 	return count;
 }
 
--- a/elks/arch/i86/drivers/char/tcpdev.c
+++ b/elks/arch/i86/drivers/char/tcpdev.c
@@ -24,8 +24,8 @@
 
 #ifdef CONFIG_INET
 
-unsigned char tdin_buf[TCPDEV_INBUFFERSIZE];
-unsigned char tdout_buf[TCPDEV_OUTBUFFERSIZE];
+unsigned char tdin_buf[TCPDEV_INBUFFERSIZE];	/* for reading tcpdev*/
+unsigned char tdout_buf[TCPDEV_OUTBUFFERSIZE];	/* for writing tcpdev*/
 
 short bufin_sem, bufout_sem;	/* Buffer semaphores */
 
--- a/elks/include/linuxmt/ntty.h
+++ b/elks/include/linuxmt/ntty.h
@@ -10,7 +10,7 @@
 #define OUTQ_SIZE	64	/* tty output queue size*/
 
 #define PTYINQ_SIZE	64	/* pty input queue size*/
-#define PTYOUTQ_SIZE	256	/* pty output queue size*/
+#define PTYOUTQ_SIZE	512	/* pty output queue size (=TDB_WRITE_MAX and telnetd buffer)*/
 
 #define RSINQ_SIZE	1024	/* serial input queue SLIP_MTU+128+8*/
 #define RSOUTQ_SIZE	64	/* serial output queue size*/
--- a/elks/include/linuxmt/tcpdev.h
+++ b/elks/include/linuxmt/tcpdev.h
@@ -10,8 +10,11 @@
 
 #define TCP_DEVICE_NAME	"tcpdev"
 
+/* should be equal to PTYOUTQ_SIZE and telnetd buffer size*/
+#define	TDB_WRITE_MAX		512	/* max data in tdb_write packet to ktcp*/
+
 #define TCPDEV_INBUFFERSIZE	1024
-#define TCPDEV_OUTBUFFERSIZE	128
+#define TCPDEV_OUTBUFFERSIZE	(TDB_WRITE_MAX + sizeof(struct tdb_write))
 
 #define TCPDEV_MAXREAD TCPDEV_INBUFFERSIZE - sizeof(struct tdb_return_data)
 
@@ -65,7 +68,6 @@ struct tdb_write {
     struct socket *sock;
     int size;
     int nonblock;
-#define	TDB_WRITE_MAX	100
     unsigned char data[TDB_WRITE_MAX];
 };
 
--- a/elkscmd/inet/telnetd/telnetd.c
+++ b/elkscmd/inet/telnetd/telnetd.c
@@ -20,7 +20,7 @@
 
 //#define RAWTELNET	/* set in telnet and telnetd for raw telnet without IAC*/
 
-#define MAX_BUFFER 100
+#define MAX_BUFFER 512		/* should be equal to TDB_WRITE_MAX and PTYOUTQ_SIZE*/
 static char buf_in  [MAX_BUFFER];
 static char buf_out [MAX_BUFFER];
 
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -15,7 +15,7 @@
 
 /* leave these off for now - too much info*/
 #define DEBUG_TCPDEV	0		/* tcpdev_ routines*/
-#define DEBUG_MEM0		/* debug memory allocations*/
+#define DEBUG_MEM	0		/* debug memory allocations*/
 #define DEBUG_CB	0		/* dump control blocks*/
 
 #if USE_DEBUG_EVENT
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -23,8 +23,10 @@
 #define IP_BUFSIZ	(TCP_BUFSIZ + sizeof(iphdr_t) + sizeof(struct ip_ll))
 
 /* control block input buffer size - max window size*/
-//#define CB_IN_BUF_SIZE	1024	/* must be power of 2*/
 #define CB_IN_BUF_SIZE	4096	/* must be power of 2*/
+
+/* max outstanding send window size*/
+#define TCP_SEND_WINDOW_MAX	1024	/* should be less than TCP_RETRANS_MAXMEM*/
 
 /* bytes to subtract from window size and when to force app write*/
 #define PUSH_THRESHOLD	512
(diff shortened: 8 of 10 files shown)


Mellvik:
Very nice numbers @ghaerr!!
And more importantly - impressive improvement. I’m assuming your machine is pretty high end compared to what I’m using these days - guessing Pentium-something at 1GHz? Of course I’m envious.

At the last test I got (as reported before) 22k outgoing, max 28k, expecting to see the double now. Not impressive, but we’re getting there. And I’m inclined to test double buffering on output just to see how much of a difference it makes. Also, since the driver is now effectively running half duplex, we should have an improvement coming when that has been fixed too.

Numbers are coming - again delayed, which ended up triggering an idea.

I pulled the last ELKS version from the repository (by now no longer the last, so I’ll get to measure the improvements from the last PR), compiled, transferred the new kernel to the elks machine (ftpget), but forgot to transfer ktcp, telnet and https. So - when rebooting, the boot process got stuck after loading telnetd. Stuck. And I’m 300 mi away, what do I do?

Here’s the idea that would prevent such problem from happening: If inittab is processed before running the rc.sys script, the system would boot even when the network is broken. It will look somewhat messy on the console startup screen, but then again, that is quite messy already. Most importantly, it would benefit debugging. [BTW there is a typo in line 47 in rc.sys: it says  'net start clsip’ . ]

After this lapse of consciousness ( :-) ) - I figured the machine would be unavailable till I get home. But before shutting down completely, I decided to reboot (I have a phone controlled power switch on the machine) and hit ^C continuously from the point where the VFS mount messages appear. That turned out to work. I managed to interrupt the system startup script before it got to the network startup and got a getty @ the console. And on ttyS1.
A huge sigh -  I’m back in business.

A couple of questions about the code:
During the TCP testing that eventually lead to the test reports last week, I repeatedly came across line 234 in tcp.c - and never got around to analyze it. It’s an if statement that is always true:

        if (1 || (h->flags & TF_PSH) || CB_BUF_SPACE(cb) <= PUSH_THRESHOLD) {
How does this makes sense?

Also, in tcp_output.c I noticed that the global retransmit memory counter tcp_retrans_memory counts payload only, not the memory footprint of the entire packet. Again I did not get around to check whether this is OK, it may well be - but I had to ask.

More coming.

—Mellvik

ghaerr:
> If inittab is processed before running the rc.sys script, the system would boot even when the network is broken

I'll look into having /bin/init process /etc/inittab in a different order. That file specifies both the init level and the sysinit script to run.

>
        if (1 || (h->flags & TF_PSH) || CB_BUF_SPACE(cb) <= PUSH_THRESHOLD) {
> How does this makes sense?

The "if (1 ..." is the remains of earlier debugging code I added months ago to force certain actions to simplify TCP processing. In this case, this forces "pushing" data from the network to the ELKS application on every packet. I plan on slowly removing certain conditions like this after we have truly stable networking. A cleanup PR is already in the works, since we recently have achieved a new stability level, subject to your testing.

> Also, in tcp_output.c I noticed that the global retransmit memory counter tcp_retrans_memory counts payload only, not the memory footprint of the entire packet.

Its OK - it only stores the TCP packet data, and calls the IP layer to rebuild the IP (and lower level) portions of the packet on each retransmit.

Mellvik:
Thanks @ghaerr,
I’ll make that adjustment and test. 

I’m still tasing the the TCP code, looking very promising. I’m seeing a ~100% speed increase in send, just like you. From ~24 to ~51kBps, which is very significant.

The robustness is now impressive. I’m getting the retrans retry messages you mentioned, lots of them when there is heavy load and overload, along with NIC receive buffer overruns, but I’ve not been able to crash ktcp or an individual connection, which must mean that the recovery when packets are lost, is working. 

I’m also confirming the ‘write to unknown socket’ when aborting curl.

Running off of floppies I’ve not been able to do any real speed tests (the files are too small and it doesn’t really make sense to measure I/O off of floppies, just too slow) - I’ll get back to that when I get the symbolic link to FAT fs working.

Down an entirely different road - Is it me, or is there a problem with bootopts these days? I ran into problems - first getting boot to recognize anything in the file except ’serial=ttyS0 net=eth 6’ - anything else and the file just gets ignored. Tested on QEMU and VB, and I found not get them to recognize bootopts at all. May be something screwed in my setup. Or is there a size/# of lines issue?

More later.

—Mellvik

ghaerr:
> I’m seeing a ~100% speed increase in send, just like you. From ~24 to ~51kBps, which is very significant.

Wow! Super improvement even on a slow machine. I've looked up some research on TCP packet size, and found the following graph showing theoretical system speed based on TCP packet size. Note that above  512, the improvement drops off rapidly, which is great, considering that ELKS has previously discussed issues with increasing multiple buffers to the full ethernet packet size.

TCP Payload

The next (small) enhancement is to change `fmemcpyb` and family in the kernel to use word copies internally, which should help on most systems with a 16-bit bus. More on that coming. I also plan on improving MTU handling, and possibly using that ktcp -m MTU start option for more than advertising receive MTU size in the initial SYN IP packet.


> is there a problem with bootopts these days? I ran into problems - first getting boot to recognize anything in the file except ’serial=ttyS0 net=eth 6’ - anything else and the file just gets ignored. 

I think the problem is that when the /bootopts file is larger than 255 bytes, the system ignores it for correctness reasons. Long story short, there is only room for a 256 byte /bootopts file. And of course I loaded it up with lots of commented out options, so in many cases, the edited file (with commented out options) exceeds the 256 byte limit. The kernel should print a "Ignoring /bootopts: header not ## or size > 255" message at the very start of boot when this happens.

So - remove all the other commented out options and see if that helps.

Mellvik:
Indeed, @ghaerr - this is making a hell of a difference. And I fully agree with you with regard to the balance between packet size and beneficial speed improvement. BTW - It's important to remember - and very beneficial - that we're pushing the networking implementation from almost opposite sides. While your machine is 10 maybe 50 times faster than the 10MHz network, mine is a lot slower than the network. In your case ELKS is pushing the network, in my case the network is pushing ELKS. Between us - given persistence - we have a perfect lineup to push out whatever is there in terms of bugs and weaknesses.  

As for the other adjustments, a few percent here and there adds up. Which reminds me - among all the improvements IP and TCP level, ping (64bytes) (ICMP) is 10% slower than before. Any clues? I need to look closer at that, but the first test indicate the it's not the driver.

Encouraged by the recent huge step forward in terms of both speed and robustness (thanks again), I'll check the effect of the 'half-duplex' change this weekend. What is missing is a reasonable way to measure inbound transfers. My plan is to add that to urlget very soon (-t option), something I started 3 weeks back. Right now ping delivers the only inbound metric available.

The /bootopts challeng - I'll check that and I have a suggestion: To mee it seems important that the file is selfdocumenting. I just cannot keep all the options in my head for very long. So how about the file being any length while boot always reads just 255 bytes? And that fact being documented clearly in the file?? My 2cents...

--Mellvik

ghaerr:
>  among all the improvements IP and TCP level, ping (64bytes) (ICMP) is 10% slower than before. 

My take is that the wait loop at the end of ne2k_packet_put could cause this. The reason is that my change likely doesn't affect continuous outbound traffic, like your curl test, very much, because the wait for a packet to complete doesn't affect much when the next driver call is likely to send another packet. The driver would have to wait anyways. However, with ping, the response could be being sent, and waiting for complete, instead of processing the next inbound packet. Easily tested by temporarily removing the wait.

Unfortunately, on my fast system, not having the wait caused the driver to fail when writing two consecutive packets. So the driver definitely has a problem.

>  I'll check the effect of the 'half-duplex' change this weekend. 

If your system is slower than the network as you say, then unfortunately you won't find out much. If you end up rewriting the TX bit to test, so that the driver properly handles fast packet writes back-to-back, just send me a patch and I'll test on my fast system, since it won't reproduce on yours.

That said, I don't anticipate the wait hack I added to affect much - perhaps 10% on pings... if that's even the problem.
Far better to have absolute reliability than worrying about performance. That's why I waited until now to add performance enhancements.

> So how about the [/bootopts] file being any length while boot always reads just 255 bytes? 

I considered that heavily during the design of the enhancement, but decided against it because there would be no error displayed and an option would be ignored if at the end of the file. I think the bigger issue is that the printk stating /bootopts had been ignored was not seen by you. Perhaps it should be in CAPS, or it scrolled off the top of the screen. I'm still working on keeping the kernel boot log display well under 24 lines.

Mellvik:
@ghaerr,
catching up on a few things ..
> among all the improvements IP and TCP level, ping (64bytes) (ICMP) is 10% slower than before.
> 
> My take is that the wait loop at the end of ne2k_packet_put could cause this. The reason is that my change likely doesn't affect continuous outbound traffic, like your curl test, very much, because the wait for a packet to complete doesn't affect much when the next driver call is likely to send another packet. The driver would have to wait anyways. However, with ping, the response could be being sent, and waiting for complete, instead of processing the next inbound packet. Easily tested by temporarily removing the wait
> 
This one remains a mystery since the wait loop mentioned was commented out when the test was run, as was my wait loop just above it. Anyway, the RTT is back to its best after all the other changes, now recording down to 1.90ms on the 386/20, Eagle interface.
> 
> I'll check the effect of the 'half-duplex' change this weekend.
> 
> If your system is slower than the network as you say, then unfortunately you won't find out much. If you end up rewriting the TX bit to test, so that the driver properly handles fast packet writes back-to-back, just send me a patch and I'll test on my fast system, since it won't reproduce on yours.
> 
It does not, and - with all the other changes introduced over the pat few weeks, the change I put in doesn't matter either. Hundreds of thousands of packets and there is not a trace. The testing is in fact becoming boring (!!) as there are no errors to report. Details in #745.
> That said, I don't anticipate the wait hack I added to affect much - perhaps 10% on pings... if that's even the problem.
> Far better to have absolute reliability than worrying about performance. That's why I waited until now to add performance enhancements.
> 
> So how about the [/bootopts] file being any length while boot always reads just 255 bytes?
> 
> I considered that heavily during the design of the enhancement, but decided against it because there would be no error displayed and an option would be ignored if at the end of the file. I think the bigger issue is that the printk stating /bootopts had been ignored was not seen by you.
> 
Turns out this is not the issue. The file is significantly less than the 255 limit, and I'm going to dive back in to this one later this week.
> Perhaps it should be in CAPS, or it scrolled off the top of the screen. I'm still working on keeping the kernel boot log display well under 24 lines.
> 
I don't necessarily think this is meaningful. The # of lines at boot is dependent upon many variables such as the # of partitions and disks and floppies and ttys. And to me it's never a problem since I hardly ever use the physical console as the ELKS console. The important improvement as fas as console messages are concerned is already done, the '\r' on serial…


—Mellvik
> —
> 
>

Issue #776, opened 2020-10-08
Kernel Panic when opening symbolic link to FAT file


Mellvik:
For all I know this may be a known problem, but here we go:

When https opens a file (for send) and the file is a symbolic link to a file on a mounted FAT filesystem, the kernel crashes like this. The problem is repeatable.
```
login: root
# mount -t fat /dev/bda1 /mnt
FAT: me=f8,csz=16,#f=2,floc=1,fsz=244,rloc=489,#d=512,dloc=521,#s=997857,ts=997857
  FAT: 498M, fat16 format
                         # ls -l /var/www
lrwxrwxrwx  1 root     root           15  Oct 08 15:47 image -> /mnt/elks/image
-rwxr-xr-x  1 root     root          275  Sep 30 12:34 index.html
lrwxrwxrwx  1 root     root            8  Sep 30 18:48 xx -> /dev/bda
# cat /var/www/image > dev/null
#
  panic: Process 26 exceeded kernel stack limit! magic 8

                                                        apparent call stack:
                                                                            Line: Addr    Parameters
                    ~~~~: ~~~~    ~~~~~~~~~~
                                               0: 2A3B => 1400 1C08 1400 1C40 0400 001A 0011
               1: 3694 => 0000 0000 0000 0000 0000 0000 0000
                                                               2: 0000 => 642F 7665 632F 6E6F 6F73 656C 5500
                               3: 0000 => 642F 7665 632F 6E6F 6F73 656C 5500
                                                                               4: 0000 => 642F 7665 632F 6E6F 6F73 656C 5500
                                               5: 0000 => 642F 7665 632F 6E6F 6F73 656C 5500
               6: 0000 => 642F 7665 632F 6E6F 6F73 656C 5500
                                                               7: 0000 => 642F 7665 632F 6E6F 6F73 656C 5500
                               8: 0000 => 642F 7665 632F 6E6F 6F73 656C 5500
                                                                            SYSTEM HALTED - Press CTRL-ALT-DEL to reboot:
```

Mellvik:
Update: 
The `cat /var/www/image > /dev/null` works once. The next attempt hangs the system hard, no messages.

ghaerr:
Hi @Mellvik,

This crash is the same as the previous kernel stack limit exceeded problem you reported several weeks ago.

The problem is easily fixed by increasing the kernel stack size in sched.h:
```
#define KSTACK_BYTES 600    /* Size of kernel stacks */
```
[...]

PR #778, opened 2020-10-08
[kernel] Reduce kernel stack usage on symlinks to FAT filesystem

ghaerr:
This should fix issue #776 and the kernel stack overflow mentioned in passing regarding cp /bin/\* in #742.

Instead of increasing kernel stack size from 512 to 600, several large stack buffers were found in various FAT functions, which could be converted to static to conserve stack space, as the routines are not required to be reentrant.

Looking further into the reason for #776, I find that the MINIX symlink follow code is recursive, which uses too much stack space. This is already documented as a FIXME in minix/symlink.c. That will have to be fixed at a later time.

Testing with CONFIG_STRACE shows that the current high kernel stack usage marker is 426 bytes with this fix, instead of the 524 earlier which caused stack overflow.

### DIFF
--- a/elks/fs/msdos/dir.c
+++ b/elks/fs/msdos/dir.c
@@ -99,7 +99,8 @@ int FATPROC msdos_get_entry_long(
 	off_t oldpos = *pos;	/* location of next directory entry start*/
 	int is_long;
 	unsigned char alias_checksum = 0;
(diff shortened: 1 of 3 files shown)


Mellvik:
Thanks @ghaerr,

You made that one very easy on me. Will test tomorrow.


--M

PR #779, opened 2020-10-08
[kernel] Add CR before NL on serial printk display

ghaerr:
Fixes kernel printk on serial windows on Linux.

Also fixes problem of network logins asking for Password: after 8 characters typed.

### DIFF
--- a/elks/kernel/printk.c
+++ b/elks/kernel/printk.c
@@ -37,9 +37,9 @@
 
 dev_t dev_console;
 
+extern void rs_conout(dev_t, char);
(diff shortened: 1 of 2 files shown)


Mellvik:
@ghaerr, now that this change has been in effect for a few days, it turns out to be one of this small things that make a big difference. Those of us using serial console only for months have gotten so used to the messed up kernel messages that we (!) need to check a couple of times extra to verify that this is indeed ELKS booting. MacOS loves it. Thanks.

Issue #73, opened 2017-03-02
[inet] Hashed stream in case of full TCP window


mfld-fr:
While testing the telnet client, I observe that, when the server sends much data, the client reads and displays only a part of this data, then blocks. One has to press a key to make the client to display next data. In Wireshark, these blocking times looks like to happen when the server reaches the TCP window limit advertised by ELKS, that is currently 2 * slip_mtu = 2 * 256 = 512 bytes.

georgp24:
I observed the same with urlget, however, after a short while transfer continues, no need to press a key.

I had already mentioned that using slip_mtu should be modified for ethernet. Slip_mtu is used in tcp_output.c. One could test if the interface is ethernet, as being done e.g. in ip.c, and then use a different mtu.

mfld-fr:
Having small TCP MSS (not MTU) & window size causes more (and smaller) packets to transfer data, but is not the root cause of the problem. These parameters are initially negotiated between the client and the server at TCP connection phase, then the window size is adjusted continuously during the streaming phase. We should keep small values for these parameters, because ELKS has no much room for additional socket buffers.

mfld-fr:
To help understanding how it works and thus how it should be implemented: http://www.tcpipguide.com/free/t_TCPSlidingWindowAcknowledgmentSystemForDataTranspo.htm

ghaerr:
> Having small TCP MSS (not MTU) & window size causes more (and smaller) packets to transfer data, but is not the root cause of the problem.

Actually, the root cause of this problem is the implementation of the PTY driver in ELKS.

Should a network process present more data (that is, queue more data into the chqueue queues) than free queue space available, the PTY driver will block the process.

Should more network packet data arrive into ktcp than are buffers available, ktcp will drop the packet. This currently causes an unrecoverable retransmission cycle, but I believe that can be fixed to recover from a dropped receive packet.

The former is the problem.

> While testing the telnet client, I observe that, when the server sends much data, the client reads and displays only a part of this data, then blocks. One has to press a key to make the client to display next data. 

After the PTY process is blocked, which is actually telnetd, nothing happens as telnetd is put to sleep. When a character is typed, the TTYP process ends up doing a wake_up which allows telnetd to resume processing normally.

PR #707 fixes this problem by increasing the size of the output queue from 128 to 256 (it was 64 in 2018). Since the current TCP/IP stack has no notion of input nor output window handling, it is currently advertises a window size such that the output queue won't be likely filled.

[...]

PR #781, opened 2020-10-09
[kernel] Fix longstanding PTY bug requiring keypress to continue

ghaerr:
Finally tracked down the bug which caused certain telnet sessions to stop displaying, which would start again on any keypress.

Initially seen years ago in #73, and temporarily fixed in #707.

PTY queues no longer have to be larger than the input queue, as stated in https://github.com/jbruchon/elks/issues/73#issuecomment-680274260, as the real problem was the PTY `ttyp_write` subdriver was blocking the process needlessly when the queue was full. The queue full check and required wait is already performed by the `ch_wait_wr` call in the master TTY `tty_write` routine.

This was tracked down using two nested `telnet localhost` connections and then a long "ls -lR /", which failed using the most recent networking code.

### DIFF
--- a/elks/arch/i86/drivers/char/pty.c
+++ b/elks/arch/i86/drivers/char/pty.c
@@ -145,8 +145,6 @@ static void ttyp_release(struct tty *tty)
 
 static int ttyp_write(register struct tty *tty)
 {
-    if (tty->outq.len == tty->outq.size)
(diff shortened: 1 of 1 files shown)


Mellvik:
Congratulations @ghaerr - that was a tricky one indeed.
Will commence testing on sunday.

—M

PR #786, opened 2020-10-11
[elksemu] Add -t option: checks if Linux host supports elksemu

tkchia:
`elksemu` currently relies on using the `modify_ldt`(2) system call to create 16-bit protected mode segments to run ELKS programs in.  Apparently there are some Linux hosts --- such as the one used by [GitLab](https://gitlab.com/) --- where the kernel does not allow using `modify_ldt` to create 16-bit segments.

To check for this, I added a `-t` option flag to `elksemu`, which will do a quick test to see if the Linux host can support `elksemu`.  This makes it easier for automated scripts to distinguish between cases where

  * `elksemu` is simply not supported on a test system, or
  * `elksemu` is failing for some other reason, e.g. a bug in `elks-libc`.

Thank you!

### DIFF
--- a/elksemu/elks.c
+++ b/elksemu/elks.c
@@ -76,17 +76,26 @@ static void elks_init()
 			  = ds = ds_idx * 8 + 7;
 	dbprintf(("LDT descriptor for text is %#x\n", cs));
 	dbprintf(("LDT descriptor for data is %#x\n", ds));
-	/* Try to place dummy descriptors in the LDT */
+	/* Try to place dummy descriptors in the LDT.  For each dummy
+	 * descriptor, make sure that either the base or the limit is
+	 * non-zero, so that Linux does not think that we are trying to clear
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

Looks good. Are you using `elksemu` in an automated environment for testing ELKS executables?

Thank you!

tkchia:
Hello @ghaerr,

> Looks good. Are you using `elksemu` in an automated environment for testing ELKS executables?

Yes --- in fact I am --- as part of the current [`build-ia16`](https://github.com/tkchia/build-ia16) scripts for building and testing the `gcc-ia16` toolchain.  I hope to get some assurance that any changes to the toolchain or ELKS will still allow them to produce working ELKS executables.

(As a side note, I am also hoping to port `elksemu` to FreeBSD some time, with a view to eventually getting it to run on macOS.)

Thank you!

PR #801, opened 2020-10-18
[tools] Update EMU86 version in tools/Makefile

ghaerr:
This updates the EMU86 8086 emulator to run with the current version of ELKS.

Hello @tkchia,

Is this the proper way to do this? I haven't actually tested running build/tools.sh, since I'm still working. The idea would be that running "tools/build.sh" would only update EMU86 but not rebuild the gcc tools. [EDIT: I have tested this and it works fine.]

Thank you!


tkchia:
Hello @ghaerr,

Yes, the `emu86` build does not depend on the `gcc-ia16` build --- in fact from what I remember `emu86` is actually built after `gcc-ia16`.  Thanks!

Issue #774, opened 2020-10-06
Tools not compiling


Mellvik:
`tools/build.sh`  no longer builds on ELKS.

The final output is
```
checking how to run the C preprocessor... gcc -E
checking build system compiler gcc... rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory
no
configure: error: Specified CC_FOR_BUILD doesn't seem to work
make[2]: *** [configure-gmp] Error 1
make[1]: *** [all] Error 2
make: *** [/Users/helge/src/mellvik/elks/cross/build/.gcc.build] Error 2
```
... but there are a number of similar situations before this `rm <some directory>.dSYM: is a directory`

Clues anyone? This is on a mac, updated Xcode tools.

--Mellvik

ghaerr:
Hello @Mellvik,

I'm on macOS Mohave 10.14.6 and the tools build works for me.

I'm running gcc (Apple LLVM) 9.0.0 (clang-900.0.38).
The following versions are in my tools/Makefile:
BINUTILS_VER=f19a905d1434da03940951913514a9ef12f0d7dd
GCC_VER=45605f97ac4d313f8783e1ba9924d0aec0b1d6e8

We'll see what @tkchia has to say.

ghaerr:
Looking at elks/cross/build/gcc-src/gmp/configure, I can see several places where "rm -f conftest\*" or similar are used, which could be an issue for you if your system failed some of the configure tests, and the rm was executed. You may want to try editing that file, search for "conftest" and replace the number of "rm -f ..." with "rm -rf ..." for the conftest\* and conftest.\* used since .dSYMs are directories on macOS.

Then restart the build. I'm not quite sure how to restart the build using make instead of tools/build.sh, but I think something like "make -C tools all" will do it.

tkchia:
Hello @Mellvik, hello @ghaerr,

Very interesting --- and weird --- error.

It seems that this happens not when building GCC proper, but when building the GNU MP library (which GCC depends on).

I am not sure the `rm` was causing the build to fail, since you mentioned that there were several similar such `rm` messages just before the failure.

If the build still does not work, consider checking `cross/build/gcc-build/gmp/config.log`, which should give detailed logs from the `configure` run.  This should show more clearly what was leading to the "`Specified CC_FOR_BUILD doesn't seem to work`" message.

Thank you!

Mellvik:
Thanks,
[...]

PR #804, opened 2020-10-18
[cross] bugfix: GCC was not rebuilt when GMP, MPFR, or MPC updated

tkchia:
Follow-up to 5ddd725257d6080e893940b5012ee6fd841c5729 .

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -71,29 +71,28 @@ MPC_DIST=mpc-$(MPC_VER)
 
 $(DISTDIR)/$(GMP_DIST).tar.bz2:
 	mkdir -p $(DISTDIR)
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

Looks good, do you we need to find a test case for this? It is of course fine to commit now and wait until the next GCC or associated libraries update otherwise.

Thank you!

tkchia:
Hello @ghaerr ,

Well, the rebuilding is certainly triggered on my setup, so I guess it is working (and not breaking anything).  I am not sure though if this will finally solve @Mellvik's problem (https://github.com/jbruchon/elks/issues/774), which this is supposed to fix.  Perhaps I will commit this first and see.

Thank you!

PR #811, opened 2020-10-22
[kernel] Debug and rewrite ramdisk

ghaerr:
Reviewed, debugged and rewrote ELKS ramdisk driver.
It now works great and is smaller and efficient.

For those unfamiliar with ELKS ramdisk, it can be used as an alternative to CONFIG_FS_EXTERNAL_BUFFER, which are general purpose buffers, which change as the system reads or writes data. For those that want specific programs or data to be fast-accessed, a ramdisk works well.

A ramdisk is created using the `ramdisk` program, followed by `mkfs` to create a MINIX filesystem. To create a 100k MINIX ramdisk use the following:
```
ramdisk /dev/rd0 make 100
mkfs /dev/rd0 100
fsck -lvf /dev/rd0 (optional)
mount /dev/rd0 /mnt
```


Mellvik:
Thank you, @ghaerr - you're right, there must be many of us not familiar with elks ramdisk.
For sure an issue for the wiki soon. 

So, unless I'm totally off, the ramdisk eats off of the regular ram. What about upper ram, expanded or extended? Are we on a road to take advantage of these goodies anytime soon?

 Actually, this is quite exciting - in particular for someone who's been living off of floppies for 12 months. Think about it, /bin in ram?

-M

ghaerr:
> So, unless I'm totally off, the ramdisk eats off of the regular ram. What about upper ram, expanded or extended? Are we on a road to take advantage of these goodies anytime soon?

Actually, adding the ability for the `ramdisk` application to specify a ramdisk to be created using RAM outside of ELKS, but under 1M, would be extremely easy. We would just pass the segment address of the memory and its length (max 64k) to the program, which uses ioctl to tell the driver. This may be the easiest use of upper/expanded memory yet!

Multiple ramdisks can be created using /dev/ram0 (=/dev/ramdisk) and /dev/ram1, which would possibly be useful for two separate upper/expanded memory segments.

If this is of interest, we need an ability to test it.


>Think about it, /bin in ram?

Yes, I figured you'd like it. However, remember that ramdisks are pretty small, compared to the size of some of our binaries. We won't be able to get anywhere's near all of /bin in ram.

ghaerr:
> What about upper ram, expanded or extended? Are we on a road to take advantage of these goodies anytime soon?

I went ahead and added support for upper memory to be used within ram disks in this PR. This could be useful with enhancement requests #709 and #708.

Two new configuration options are now visible and set in .config when "RAM disk support" is enabled:
```
Preload RAM disk segment address   CONFIG_RAMDISK_SEGMENT 0 (e.g. set to "top-0x1000" in boot screen or 0xD000)
Preload RAM disk sectors           CONFIG_RAMDISK_SECTORS 128 (=64k in 512 byte sectors. Use 256 for 128k)
```

If CONFIG_RAMDISK_SEGMENT is set to 0xD000, then 128 sectors (64k) of upper memory will be used in the /dev/rd0 ramdisk, automatically configured at boot.

To setup a preloaded ramdisk using normal system memory, first boot the system without the ramdisk configured. A new display will show the segment address of the top of memory (here segment 9fc0):
```
Kernel text at c0:0000, ftext cbd:0000, data fa0:0000, top 9fc0, 512K free
```
If a 64k ramdisk is desired, subtract 0x1000 (=64k in paragraphs) from the "top" value of 9fc0 to get the following in .config:
```
CONFIG_RAMDISK_SEGMENT=0x8fc0
```
When ELKS boots with a pre-loaded ramdisk, it will show the address, which should match the new top of memory:
```
rd: 64K ramdisk at 8fc0:0000
...
Kernel text at c0:0000, ftext cbd:0000, data fa0:0000, top 8fc0, 448K free
```
As long as the "top" value matches the ramdisk start segment, and the number of sectors is set match the amount of top memory subtracted, things will all work out.

I also changed the default ramdisk device names to /dev/rd0 and /dev/rd1, which look and act similarly to /dev/fd0.
An additional ramdisk can be setup along with a preloaded ramdisk by using the `ramdisk` utility, which will use normal system memory for it:
```
ramdisk /dev/rd1 make 100
mkfs /dev/rd1 100
mount /dev/rd1 /mnt 
...
```
@pawosm-arm, you may want to try testing the upper memory ramdisk with one of your memory cards. I've tested this with QEMU, which emulates memory at 0xD000.

pawosm-arm:
@ghaerr I've tried the following settings:

```
CONFIG_BLK_DEV_RAM=y
CONFIG_RAMDISK_SEGMENT=0xd000
CONFIG_RAMDISK_SECTORS=128
```

...and I have immediate question:

Why two ramdisk devices are created?

```
# ls /dev
console  eth      fd0      fd1      hda      hda1     hda2     hda3
hda4     hda5     hda6     hda7     hda8     hdb      hdb1     hdb2
hdb3     hdb4     hdc      hdc1     hdd      kmem     lp       mem
null     ptyp0    ptyp1    ptyp2    ptyp3    rd0      rd1      tcpdev
tty      tty1     tty2     tty3     ttyS0    ttyS1    ttyS2    ttyS3
# mkfs /dev/rd0 64
21 inodes
64 blocks
Firstdatazone=5 (5)
Zonesize=1024
Maxsize=65536

# mkfs /dev/rd1 64
21 inodes
64 blocks
Firstdatazone=5 (5)
Zonesize=1024
Maxsize=65536

# mkdir /mnt/rd0
# mount /dev/rd0 /mnt/rd0
# mkdir /mnt/rd1
# mount /dev/rd1 /mnt/rd1
# df /dev/rd0
/dev/rd0             64       58        6  10%    10%
# df /dev/rd1
df: Can't read super block of /dev/rd1
```

The last line shows that it isn't right.

Offtop regarding #708: I had to pull out the EMS memory card to make place for SMC networking card. All of the Amstrad PC's have limited number of ISA slots available, and I don't have technical capabilities for doing cards piggybacking (and wasting two ISA slots just for having UMB and EMS memory cards is just screaming for piggybacking those two). A tough choice between UMB and EMS was resolved under DOS: the biggest beneficiary of EMS was 4DOS (memory effective COMMAND.COM replacement) which is able to organize swap for its own purposes in a regular DOS file (when EMS is unavailable), it doesn't hurt that much when CF card is in use, could be worse on any old MFM HDD. Having more main memory freed by moving all of the drivers to UMB outweighs all other benefits of EMS, it's clearly visible when working with Windows 3.0.

ghaerr:
> Something is wrong with that option. It causes make menuconfig to hang with no way to exit.

Thanks. It appears the menuconfig code to ask for a hex value isn't working. I'll fix it.

> Why two ramdisk devices are created?

Two ram disks are not created. It appears that your system is working properly using /dev/rd0 and the UMB memory at 0xD000. Good news!

The problem is that ramdisk driver isn't checking for a valid /dev/rd1 when you try to use it (with mkfs). It should be returning an error. The second /dev/rd1 device is always present in /dev, it just should not be active. I'll fix it.

pawosm-arm:
Two ram disks are not created. It appears that your system is working properly using /dev/rd0 and the UMB memory at 0xD000. Good news!

Yeah, boot messages confirm that:

ghaerr:
> Yeah, boot messages confirm that

Pretty cool, thanks for the screenshot!

Mellvik:
Hi @pawosmarm,
interesting numbers you have on the hda4 partition ???

—Mellvik

pawosm-arm:
> interesting numbers you have on the hda4 partition ???

I have no partitions table on hda. It's a known issue with ELKS, it tries to find partitions even if there's none. Those 'interesting numbers' come from the desperate attempt to interpret as a partition table something that isn't.

Mellvik:
Ahh,
that explains it. I'm sure it's on @ghaerr's list already …

Thanks.

—H

ghaerr:
> It's a known issue with ELKS, it tries to find partitions even if there's none. Those 'interesting numbers' come from the desperate attempt to interpret as a partition table something that isn't.

Hmmm... not trying to sound desperate, but I removed the code that eliminates an unlikely partition table entry from consideration at @Mellvik's request!

 IIRC this was for CF card "virtual cylinders".  Both "flat" boot sectors as well as MBR boot sectors are have 0xAA55 at offset 510. I'm not aware of any way to distinguish between the two... I researched this pretty well when I got all this code working, so perhaps I am getting desperate! :)

>  I'm sure it's on @ghaerr's list already …

We could undo your request change (at line 70 of genhd.c), or change it when the numbers see completely ridiculous ? I don't know any way to tell an MBR from a flat boot sector.

Mellvik:
@ghaerr,
yes I vaguely remember this, although not the discussion behind the decision (or opinion). I guess we simply have another instance of the workings of the law of unintended consequences.

That said, I do think your suggestion that we add a sanitycheck is the right one.

One that I've seen (and used) in the past is quite simple and reliable: Get the first byte of each partition entry, if one is 0x80 and the rest is 00, we have a valid partition table, otherwise we don't.
Of course we may have 2nd and 3rd drives that don't need an 'activated' partition, so that all 4 entries are zero, but that's also an illegal partition table…

My 2Cents - again.

—Mellvik

pawosm-arm:
I don't remember if this was already proposed, but, why can't we make partition table support optional? E.g. on Linux one can select among various partitioning schemes (DOS, MacOS, BSD disk labels, etc.) which leaves possibility of not selecting any partitioning scheme at all, resulting in a system which can work only with the drives entirely filled with a single partition...

ghaerr:
> One that I've seen (and used) in the past is quite simple and reliable: Get the first byte of each partition entry, if one is 0x80 and the rest is 00, we have a valid partition table, otherwise we don't.

ELKS currently checks each partition table separately, and if offset 0 (the active/inactive byte) is either 0x00 or 0x80, ELKS assumes that is a valid partition. What ELKS does NOT do is check all table entries at once for 0x00 or 0x80. Thus, I would assume that @pawosm-arm's boot sector happens to have an 0x00 or 0x80 in the fourth partition table active/inactive byte.

How about we add code to check all four table entries at once for valid (either 0x00 or 0x80), instead of each individually? That should solve this problem. We can still leave the other partition size checking code disabled.

> Of course we may have 2nd and 3rd drives that don't need an 'activated' partition, so that all 4 entries are zero, but that's also an illegal partition table…

I'm not sure it will be needed, but we could add that check also, for the case where the boot sector is all zeros.

pawosm-arm:
> Thus, I would assume that @pawosm-arm's boot sector happens to have an 0x00 or 0x80 in the fourth partition table active/inactive byte.

AFAIR that's the feature of MINIX fs...

ghaerr:
PR #814 fixes the misidentification of partition \#4 as a valid partition on @pawosm-arm's HD. It turns out this was occurring on all ELKS-bootable MINIX hard disk images. These disk images had all partition entries set to 0 (as it was empty space set to 0 after the boot code), except for the ELKS disk parameter block which overlapped the partition 4 sector count.

PR #812, opened 2020-10-23
[cmds] Bugfixes and enhancements in dd & sum.

Mellvik:
- dd: now accepts input from stdin and output to stdout.
- sum: Bugfix in block counter, moved counter out of inner (byte) loop.

### DIFF
--- a/elkscmd/file_utils/Makefile
+++ b/elkscmd/file_utils/Makefile
@@ -38,7 +38,7 @@ df: df.o
 	$(LD) $(LDFLAGS) -o df df.o $(LDLIBS)
 
 dd: dd.o
-	$(LD) $(LDFLAGS) -o dd dd.o $(LDLIBS)
+	$(LD) $(LDFLAGS) -maout-heap=12288 -o dd dd.o $(LDLIBS)
 
 grep: grep.o
 	$(LD) $(LDFLAGS) -o grep grep.o $(LDLIBS)
--- a/elkscmd/file_utils/dd.c
+++ b/elkscmd/file_utils/dd.c
@@ -22,6 +22,7 @@
 #define	PAR_COUNT	4
 #define	PAR_SEEK	5
 #define	PAR_SKIP	6
+#define DEFBUF		10 * BUFSIZ
 
 
 struct param {
@@ -41,7 +42,7 @@ static struct param params[] = {
 
 
 /* Fixed buffer */
-static char localbuf[BUFSIZ];		/* use disk block size for efficiency*/
+static char localbuf[DEFBUF];		/* use disk block size for efficiency*/
 
 
 /*
@@ -115,7 +116,7 @@ int main(int argc, char **argv)
 		str = *++argv;
 		cp = strchr(str, '=');
 		if (cp == NULL) {
-			write(STDERR_FILENO, "Missing or invalid argument(s)\n", 31);
+			fprintf(stderr, "Missing or invalid argument(s)\n");
 			goto usage;
 		}
 		*cp++ = '\0';
@@ -128,7 +129,7 @@ int main(int argc, char **argv)
 		switch (par->value) {
 			case PAR_IF:
 				if (infile) {
-					write(STDERR_FILENO, "Multiple input files illegal\n", 29);
+					fprintf(stderr, "Multiple input files illegal\n");
 					goto usage;
 				}
 	
@@ -137,7 +138,7 @@ int main(int argc, char **argv)
 
 			case PAR_OF:
 				if (outfile) {
-					write(STDERR_FILENO, "Multiple output files illegal\n", 30);
+					fprintf(stderr, "Multiple output files illegal\n");
 					goto usage;
 				}
 
@@ -147,7 +148,7 @@ int main(int argc, char **argv)
 			case PAR_BS:
 				blocksize = getnum(cp);
 				if (blocksize <= 0) {
-					write(STDERR_FILENO, "Bad block size value\n", 21);
+					fprintf(stderr, "Bad block size value\n");
 					goto usage;
 				}
 				break;
@@ -155,7 +156,7 @@ int main(int argc, char **argv)
 			case PAR_COUNT:
 				count = getnum(cp);
 				if (count < 0) {
-					write(STDERR_FILENO, "Bad count value\n", 16);
+					fprintf(stderr, "Bad count value\n");
 					goto usage;
 				}
 				break;
@@ -163,7 +164,7 @@ int main(int argc, char **argv)
 			case PAR_SEEK:
 				seekval = getnum(cp);
 				if (seekval < 0) {
-					write(STDERR_FILENO, "Bad seek value\n", 15);
+					fprintf(stderr, "Bad seek value\n");
 					goto usage;
 				}
 				break;
@@ -171,26 +172,22 @@ int main(int argc, char **argv)
 			case PAR_SKIP:
 				skipval = getnum(cp);
 				if (skipval < 0) {
-					write(STDERR_FILENO, "Bad skip value\n", 15);
+					fprintf(stderr, "Bad skip value\n");
 					goto usage;
 				}
 				break;
 
 			default:
-				write(STDERR_FILENO, "Unknown dd parameter\n", 21);
+				fprintf(stderr, "Unknown dd parameter\n");
 				goto usage;
 		}
 	}
 
 	if (infile == NULL) {
-		//write(STDERR_FILENO, "No input file specified\n", 24);
-		//goto usage;
 		infile = "-";
 	}
 
 	if (outfile == NULL) {
-		//write(STDERR_FILENO, "No output file specified\n", 25);
-		//goto usage;
 		outfile = "-";
 	}
 
@@ -198,8 +195,8 @@ int main(int argc, char **argv)
 	if (blocksize > sizeof(localbuf)) {
 		buf = malloc(blocksize);
 		if (buf == NULL) {
-			write(STDERR_FILENO, "Cannot allocate buffer\n", 23);
-			goto usage;
+			fprintf(stderr, "Cannot allocate buffer\n");
+			return retval;
 		}
 	}
 
@@ -236,7 +233,7 @@ int main(int argc, char **argv)
 				}
 
 				if (incc == 0) {
-					write(STDERR_FILENO, "Skipped beyond end of file\n", 27);
+					fprintf(stderr, "Skipped beyond end of file\n");
 					goto cleanup;
 				}
 			}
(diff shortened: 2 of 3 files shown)


Mellvik:
More dd-fixes:
- replaced all write(fileno(STANDARD_ERROR)... ) with fprintfs
- send all messages to stderr 
- allocate 10k static buffer (was 1k)
- still needs expanded stack space in order for the bs= parameter to work beyond the static buffer size.

ghaerr:
> allocate 10k static buffer (was 1k)

I'm curious, since ELKS never does I/O in blocks greater then 1k, why the buffer size was increased to 10k. Have you tested this and measured to see if there is indeed any increased benefit, or was this just added for good measure? The increased buffer size otherwise just makes dd require more memory in order to run.

For the sum fix, please remove all "//" commented-out lines and \# 0's. Since put has been replaced by printf, there is no reason keep the old code around. Failure to remove this now usually means that cruft will stay forever.

> still needs expanded stack space in order for the bs= parameter to work beyond the static buffer size.

You mean for the malloc? That would be heap space. That can be added using --maout-heap=10240. In general, I believe there is no benefit in large buffers on ELKS, go ahead and make the heap change now but test, thanks. If you increase both the static buffer as well as heap size by 10k each, then the memory requirement for dd goes up by 20k. It may be better to keep the static buffer 1k bytes and just increase the heap. I mention all this because this dd command is actually originally from sash, not busybox nor busyelks. Thus, some of these changes may want to be ported back to sash, but definitely not the increased buffer size.

Thank you!

Mellvik:
> 
> allocate 10k static buffer (was 1k)
> 
> I'm curious, since ELKS never does I/O in blocks greater then 1k, why the buffer size was increased to 10k. Have you tested this and measured to see if there is indeed any increased benefit, or was this just added for good measure? The increased buffer size otherwise just makes dd require more memory in order to run.
> 
The intention is speed, the reason is that malloc failed, even for 1blk. The ability to define block size has always (unix v6) been a major feature of dd - for a good reason. It was created to copy data between devices. Which is still one of its primary feats. using a block size of (say) 80b when copying floppy to harddisk is a significant improvement on any system with limited lower level buffering.

If malloc works, there is no good reason for static buffer allocation at all.
> For the sum fix, please remove all "//" commented-out lines and # 0's. Since put has been replaced by printf, there is no reason keep the old code around. Failure to remove this now usually means that cruft will stay forever.
> 
> 
Point taken.
> still needs expanded stack space in order for the bs= parameter to work beyond the static buffer size.
> 
> 
> You mean for the malloc? That would be heap space. That can be added using --maout-heap=10240.
> 
Sorry, I meant heap space. And OK; I'll add that. The usefulness of dd increases with the size of the available buffers.

—Mellvik

ghaerr:
> using a block size of (say) 80b when copying floppy to harddisk is a significant improvement on any system with limited lower level buffering.

Are you just saying that for ELKS, since that is true for all other systems, or have you tested and measure that with "time"? I am saying that because ELKS *never* reads or writes more than 1k bytes at a time there may not be any speed increase. I would like to know based on your slower floppy system whether there is a difference or not, and how much. This will help for our open floppy speed issues within the kernel.

> If malloc works, there is no good reason for static buffer allocation at all.

Ok, lets go with adding the --maout-heap= option and keeping the original 1k buffer. This will keep the memory requirements of dd down, while still allowing you to allocate large buffers (and I'm still interested in whether they actually help).

Mellvik:
> 
> using a block size of (say) 80b when copying floppy to harddisk is a significant improvement on any system with limited lower level buffering.
> 
> Are you just saying that for ELKS, since that is true for all other systems, or have you tested and measure that with "time"? I am saying that because ELKS never reads or writes more than 1k bytes at a time there may not be any speed increase. I would like to know based on your slower floppy system whether there is a difference or not, and how much. This will help for our open floppy speed issues within the kernel.
> 
Well, yes and no. No measurements - a mix of perception and expectation: Reading a track at a time (or half a track if that's what the buffer allows) from a slow floppy before writing to whatever seems like an obvious optimization regardless. I'm always using bs=80b on Linux when writing floppies, but haven't measured the effect of various block sizes for ages.
> If malloc works, there is no good reason for static buffer allocation at all.
> 
> Ok, lets go with adding the --maout-heap= option and keeping the original 1k buffer. This will keep the memory requirements of dd down, while still allowing you to allocate large buffers (and I'm still interested in whether they actually help).
> 
Will measure the difference with time() when I get the chance.


Pushing updates to the PR first.

—Mellvik

Mellvik:
OK; setting the heap to 10240 allows 19 blocks,
upping it to 10752 allows 20block, and seems to work even though malloc complains:

# mount -t fat /dev/hda1 /mnt
FAT: me=f8,csz=16,#f=2,floc=1,fsz=244,rloc=489,#d=512,dloc=521,#s=997857,ts=997857
FAT: 498M, fat16 format
# dd if=/dev/fd0 of=/mnt/xx bs=20b
sys_brk(25) fail: brk 4050 over by 506 bytes
sys_brk(25) fail: brk 4054 over by 510 bytes
sys_brk(25) fail: brk 4056 over by 512 bytes
sys_brk(25) fail: brk 4058 over by 514 bytes
sys_brk(25) fail: brk 405a over by 516 bytes
sys_brk(25) fail: brk 3efa over by 164 bytes
144+0 records in
144+0 records out
1474560 bytes (1440 KiB) copied
# 


Is this OK?

—Mellvik

> 23. okt. 2020 kl. 17:13 skrev Helge Skrivervik <helge@mymayday.com>:
> 
>> 
>> using a block size of (say) 80b when copying floppy to harddisk is a significant improvement on any system with limited lower level buffering.
>> 
>> Are you just saying that for ELKS, since that is true for all other systems, or have you tested and measure that with "time"? I am saying that because ELKS never reads or writes more than 1k bytes at a time there may not be any speed increase. I would like to know based on your slower floppy system whether there is a difference or not, and how much. This will help for our open floppy speed issues within the kernel.
>> 
> Well, yes and no. No measurements - a mix of perception and expectation: Reading a track at a time (or half a track if that's what the buffer allows) from a slow floppy before writing to whatever seems like an obvious optimization regardless. I'm always using bs=80b on Linux when writing floppies, but haven't measured the effect of various block sizes for ages.
>> If malloc works, there is no good reason for static buffer allocation at all.
>> 
>> Ok, lets go with adding the --maout-heap= option and keeping the original 1k buffer. This will keep the memory requirements of dd down, while still allowing you to allocate large buffers (and I'm still interested in whether they actually help).
>> 
> Will measure the difference with time() when I get the chance.
> 
> 
> Pushing updates to the PR first.
> 
> —Mellvik

ghaerr:
Let’s set it to 12k, I forgot about other stdio overhead - 12288.

Mellvik:
Actually, there was nothing important about the number 20. For my assumed purpose, 18 is more relevant, one floppy track. So - 12k is fine.

Perhaps unsurprisingly, the practical testing reveals the suggested unimportance of block size for ELKS. What I had completely forgotten was that ELKS does not have raw devices, who's purpose is - as the name implies - to provide raw access, bypassing the kernel buffering system for block devices. Right now I wish we had such access, but this is what we've got for now. It is a fact that the purpose and benefit of raw devices has dwindled as the devices themselves have added layers of intelligence and buffering which is generally outside our control, but for old stuff, the case may still be made for having 'character access' to block devices.

Reading a 1.44 floppy takes 6 minutes plus a few seconds regardless, which is extremely slow. Along the way, `time` delivers some really strange results. In the case where `time` reports 17 minutes, the real time is about like the others, 6+. The recorded speed is 4KB/s, which also applies to writing. Taking the disk out of the loop (using /dev/null) takes off about 20s from the total time. 

```
# time dd if=/dev/fd0 of=/mnt/elks/yy
2880+0 records in
2880+0 records out
1474560 bytes (1440 KiB) copied

Real    6m10.800s
# time dd if=/dev/fd0 of=/mnt/elks/yy bs=18b
160+0 records in
160+0 records out
1474560 bytes (1440 KiB) copied

Real    6m4.660s
# time dd if=/dev/fd0 of=/mnt/elks/yy bs=9b
320+0 records in
320+0 records out
1474560 bytes (1440 KiB) copied

Real    17m38.787s
# 
# time dd if=/dev/fd0 of=/mnt/elks/yy bs=2b
1440+0 records in
1440+0 records out
1474560 bytes (1440 KiB) copied

Real    6m4.810s
# 
```
Just out of curiosity, I tested the effect of `dd bs=` on linux, transferring a floppy image from disk to a USB floppy drive. No raw device, just /dev/sda:
```
helge@raspberrypi:~ $ sudo bash
[sudo] password for helge: 
root@raspberrypi:/home/helge# time dd if=fd1440.bin of=/dev/sda
2880+0 records in
2880+0 records out
1474560 bytes (1.5 MB, 1.4 MiB) copied, 177.236 s, 8.3 kB/s

real    2m57.249s
user    0m0.009s
sys     0m0.121s
root@raspberrypi:/home/helge# time dd if=fd1440.bin of=/dev/sda bs=72b
40+0 records in
40+0 records out
1474560 bytes (1.5 MB, 1.4 MiB) copied, 54.1165 s, 27.2 kB/s

real    0m54.129s
user    0m0.009s
sys     0m0.023s
root@raspberrypi:/home/helge# time dd if=fd1440.bin of=/dev/sda bs=36b
80+0 records in
80+0 records out
1474560 bytes (1.5 MB, 1.4 MiB) copied, 77.2023 s, 19.1 kB/s

real    1m17.398s
user    0m0.000s
sys     0m0.037s
root@raspberrypi:/home/helge# 
```
So - even without raw access, dd's own buffer management has significant effect - when there is enough memory to throw around.

--M

ghaerr:
> Perhaps unsurprisingly, the practical testing reveals the suggested unimportance of block size for ELKS. 

After some of your last statements about reading all the sectors into a buffer first, then writing, I was actually thinking perhaps it would make a difference, but your testing has shown it does not.

> What I had completely forgotten was that ELKS does not have raw devices, who's purpose is - as the name implies - to provide raw access, bypassing the kernel buffering system for block devices.

I think the speed issue is not related to having a raw device or not, but instead completely related to being able to read many sectors in a single read operation. Consider this - the most important speed issue is whether or not a rotating sector is coming up towards the disk head when a read operation is in place. Should the sector pass the head unread, another full revolution is needed. It doesn't matter whether kernel buffers are used, or a raw device is used, as long as 1) the system tries to read multiple sectors (which ELKS reads max 2, and only sometimes), and 2) whether the system buffering is fast enough such that the next rotating sector can still be read.

Issue 2 in the last sentence is solved by doing interrupt I/O - which ELKS does not use. That is, on Linux, a read for 18 sectors is requested, and the disk driver hardware issues an interrupt, or uses DMA, to cause the requested sector to be read into a buffer when the disk sector is under the head. The dd or other application is blocked through all of this.

In ELKS, with only 1-2 sector reads, unless the disk sector numbers are formatted specially given the amount of time it takes to issue the BIOS I/O calls, we're missing the sectors every time. That's why its so slow. If you were to figure out the sector numbering with special formatting such that it matched ELKS' 2-sector BIOS call timing, speed would improve dramatically.

> So - even without raw access, dd's own buffer management has significant effect - when there is enough memory to throw around.

No - when you specify a large block size to dd, dd executes a single read for that block size, which causes the kernel to read all sectors at once, that is why it is fast. It is not because of dd's buffer management. Dd has to have a larger buffer because it issued a large read system call.

My $0.02!

On another note, the time command is buggy when it states 17m, that's that old time calculation code showing it doesn't work properly. I took a pass at it twice, you rewrote it once, and we both still don't have it totally right.

ghaerr:
Thanks for the bug fixes @Mellvik !

Mellvik:
> I think the speed issue is not related to having a raw device or not, but instead completely related to being able to read many sectors in a single read operation. Consider this - the most important speed issue is whether or not a rotating sector is coming up towards the disk head when a read operation is in place. Should the sector pass the head unread, another full revolution is needed. It doesn't matter whether kernel buffers are used, or a raw device is used, as long as 1) the system tries to read multiple sectors (which ELKS reads max 2, and only sometimes), and 2) whether the system buffering is fast enough such that the next rotating sector can still be read.
> 
> Issue 2 in the last sentence is solved by doing interrupt I/O - which ELKS does not use. That is, on Linux, a read for 18 sectors is requested, and the disk driver hardware issues an interrupt, or uses DMA, to cause the requested sector to be read into a buffer when the disk sector is under the head. The dd or other application is blocked through all of this.
> 
> In ELKS, with only 1-2 sector reads, unless the disk sector numbers are formatted specially given the amount of time it takes to issue the BIOS I/O calls, we're missing the sectors every time. That's why its so slow. If you were to figure out the sector numbering with special formatting such that it matched ELKS' 2-sector BIOS call timing, speed would improve dramatically.
> 
Ok @ghaerr, I read you and partly agree. This is a road I'm very familiar with - including special formatting to accomplish (de)skewing in hardware so to speak. (UCSD Pascal dod some great tricks to that effect - on 8" floppies)
The purpose of the raw device is to get around the buffering in the kernel, which in addidion to shorten the way from physical sector to application, avoids caching which simplifies measurements.
In our case shortcutting the kernel buffer mechanisms must be a big win regardless of the number of sectors read per operation (which will be an interesting project down the line btw). We cannot hope to process a sector before the next one passes, but we can reduce the number of rotations between each read. I haven't done the math, but 4 or 6 minutes  to get 2880 sectors @ 360 rpm and 18x2 available per rotation is a hell of a lot of missed rotations. I honestly think we should  be able to read (or write) a 1.44 floppy in less than 2 minutes.
That said, spending much time optimizing floppy io is probably not worth it. OTOH, if it contributes to general disk io speed, it is. 
> So - even without raw access, dd's own buffer management has significant effect - when there is enough memory to throw around.
> 
> No - when you specify a large block size to dd, dd executes a single read for that block size, which causes the kernel to read all sectors at once, that is why it is fast. It is not because of dd's buffer management. Dd has to have a larger buffer because it issued a large read system call.
> 
Forgive my pun, calling it buffer management was a joke. That said, the numbers underscore my point above. That said, 'normal' unix dd does indeed do buffer managenent through the ibs= and obs= options. That's a different story.

We'll get back to this issue at a later time. Thanks for the discussion.

-Mellvik

PR #814, opened 2020-10-24
[kernel] Better identify MBR partitions

ghaerr:
Fixes the problem @pawosm-arm has in booting flag hard disk images with no MBR in https://github.com/jbruchon/elks/pull/811#issuecomment-715423636. Partition \#4 was being misidentified as a valid MBR partition.

Added code to check all partitions for a valid partition status byte of 0x00 or 0x80. If all partitions inactive (=0x00), further check for ELKS Extended Parameter Block boot disk magic number, which just happens to use the last two fields in the fourth partition entry.

Thus, this corrects misidentification of all ELKS bootable flat hard disk images, as well as still allow real MBR boot blocks for images used as 2nd and 4rd hard drives with no active partition.


pawosm-arm:
I guess it looks better now:

PR #816, opened 2020-10-25
[cmds] Add fdtest floppy speed testing program

ghaerr:
@Mellvik,

Here's a floppy testing program that can be used to test floppy disk reads on your systems, since you're the only guy that actually has a floppy drive! 

I've got it setup for testing single-sector, double-sector (1k blocks), and full-track reads, with one second granularity for reporting times. I'm very interested in the results. It can easily be modified for other disk sizes by modifying the MAX parameter and recompiling. This program works completely outside the kernel by calling the INT 13h BIOS disk I/O routine directly, using a kernel library. It also includes the memory copies from simulated kernel to user for speed accuracy.

I'm hoping this will show us which way to proceed for issue #521.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -104,6 +104,7 @@ sh_utils/whoami					:shutil					:1440k
 sh_utils/xargs					:shutil					:1440k
 sh_utils/yes					:shutil				:720k
 misc_utils/miniterm				:miscutil			:720k
+misc_utils/fdtest				:miscutil				:1440k
 misc_utils/tar					:miscutil				:1440k
 misc_utils/od					:miscutil				:1440k
 misc_utils/hd					:miscutil		:360k
--- a/elkscmd/misc_utils/.gitignore
+++ b/elkscmd/misc_utils/.gitignore
@@ -2,6 +2,7 @@ miniterm
 compress
 compress.host
 ed
+fdtest
 tar
 od
 hd
--- a/elkscmd/misc_utils/Makefile
+++ b/elkscmd/misc_utils/Makefile
@@ -10,11 +10,12 @@ include $(BASEDIR)/Make.rules
 
 HOSTCC = gcc
 HOST_CFLAGS = -O2
+KERNEL_LIBS = $(TOPDIR)/elks/arch/i86/lib/lib86.a
 
 ###############################################################################
 
 #TODO: fix compress uncompress zcat lpfilter disabled
-PRGS = tar miniterm ed od hd time kilo tty uuencode uudecode
+PRGS = tar miniterm ed fdtest od hd time kilo tty uuencode uudecode
 
 #PRGS_HOST=compress.host
 
@@ -22,6 +23,9 @@ all: $(PRGS) $(PRGS_HOST)
 
 ed: ed.o
 	$(LD) $(LDFLAGS) -o ed ed.o $(LDLIBS)
+
+fdtest: fdtest.o
+	$(LD) $(LDFLAGS) -o fdtest fdtest.o $(KERNEL_LIBS) $(LDLIBS)
 
 tar: tar.o
 	$(LD) $(LDFLAGS) -o tar tar.o $(LDLIBS)
--- a/elkscmd/misc_utils/fdtest.c
+++ b/elkscmd/misc_utils/fdtest.c
@@ -49,6 +49,8 @@ void read_disk(unsigned short drive, unsigned short cylinder, unsigned short hea
 int main(int ac, char **av)
 {
 	unsigned short drive, cylinder, head, sector;
+	unsigned short start_dma_page, end_dma_page;
+	unsigned long normalized_seg;
 	int i;
 	time_t t;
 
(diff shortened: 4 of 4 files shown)


Mellvik:
@ghaerr,

I've been experimenting with fdtest and a lot of different parameters, and found that if interrupted by SIGINTR, the program will hang, some times crash the system when run a 2nd time.
I may have introduced bugs which making changes, but it still seems odd that an INTR should affect the next run?

—M

ghaerr:
> if interrupted by SIGINTR, the program will hang, some times crash the system when run a 2nd time.

Hard to say. Whenever anything hangs, there's likely system memory corruption because of programs run in unprotected memory. So I'm not surprised about the second run not working.

I'll have to look more into the crashing - could be that since we're resetting the applications DS (for the period of the fmemcpy), SIGINT ends up interrupting and running with an invalid DS. You might try commenting out the fmemcpy and see if it repeats, since the fmemcpy isn't strictly needed.

Mellvik:
Thanks,

It turns out any program runs after the interrupt will fail or hang the system

Not important unless it has consequences outside of fdtest.

—M

ghaerr:
> Not important unless it has consequences outside of fdtest.

Looking at the signal code, I think you've found another bug: if an application resets DS and receives a signal, things get screwed up. 

Can you comment out the fmemcpy in retest.c to verify this? Thanks!

Mellvik:
> 
> 
> Looking at the signal code, I think you've found another bug: if an application resets DS and receives a signal, things get screwed up.
> 
> Can you comment out the fmemcpy in retest.c to verify this? Thanks!
> 

Thanks, will do - but I'm not optimistic. I did that at one point w/o effect, but will retest to make sure.

M

Mellvik:
@ghaerr,

I did some more testing on this one, including removing the fmemcpyw call - which didn't make any difference:

The experience is that any system activity causing a context switch while fdtest is running, will hang the system hard.

—mellvik

> 26. okt. 2020 kl. 19:45 skrev Helge Skrivervik <helge@mymayday.com>:
> 
>> 
>> 
>> Looking at the signal code, I think you've found another bug: if an application resets DS and receives a signal, things get screwed up.
>> 
>> Can you comment out the fmemcpy in retest.c to verify this? Thanks!
>> 
> 
> Thanks, will do - but I'm not optimistic. I did that at one point w/o effect, but will retest to make sure.
> 
> M

Mellvik:
One more thing - 
I'm totally unfamiliar with the FP_SEG/FP_OFF etc macros, but it seems odd to me that the start_dma_page, end_dma_page and normalized_seg variables aren't used anywhere ...


—Mellvik


> 27. okt. 2020 kl. 10:49 skrev Helge Skrivervik <helge@mymayday.com>:
> 
> @ghaerr,
> 
> I did some more testing on this one, including removing the fmemcpyw call - which didn't make any difference:
> 
> The experience is that any system activity causing a context switch while fdtest is running, will hang the system hard.
> 
> —mellvik
> 
>> 26. okt. 2020 kl. 19:45 skrev Helge Skrivervik <helge@mymayday.com <mailto:helge@mymayday.com>>:
>> 
>>> 
>>> 
>>> Looking at the signal code, I think you've found another bug: if an application resets DS and receives a signal, things get screwed up.
>>> 
>>> Can you comment out the fmemcpy in retest.c to verify this? Thanks!
>>> 
>> 
>> Thanks, will do - but I'm not optimistic. I did that at one point w/o effect, but will retest to make sure.
>> 
>> M
>

ghaerr:
> I did some more testing on this one, including removing the fmemcpyw call - which didn't make any difference

I now finally understand it: fdtest is in the middle of a BIOS call (nothing to do with fmemcpy), and you hit ^C. This registers a need for SIGINT, which is to terminate the program. Since fdtest is not in a system all, nothing happens until the next clock interrupt, which interrupts fdtest and executes the timer interrupt routine. At the end of this routine, the kernel checks that a signal is waiting and terminates fdtest (which means it deallocates its memory and does not return control to fdtest). Except that fdtest was in the middle of a BIOS call!!! So that's the problem. The fix is to ignore signals in fdtest, I'll patch it.

> I'm totally unfamiliar with the FP_SEG/FP_OFF etc macros, but it seems odd to me that the start_dma_page, end_dma_page and normalized_seg variables aren't used anywhere ...

That is some very technical stuff that calculates whether or not the I/O would overlap a 64k physical memory boundary, which is required for the old 8237 DMA controller; that chip is used by the BIOS. The calculations are only for display purposes in fdtest, the real use is in the ELKS BIOS driver. These were added to ensure that fdtest didn't need additional code.

Thanks!

Mellvik:
> 
> I did some more testing on this one, including removing the fmemcpyw call - which didn't make any difference
> 
> I now finally understand it: fdtest is in the middle of a BIOS call (nothing to do with fmemcpy), and you hit ^C. This registers a need for SIGINT, which is to terminate the program. Since fdtest is not in a system all, nothing happens until the next clock interrupt, which interrupts fdtest and executes the timer interrupt routine. At the end of this routine, the kernel checks that a signal is waiting and terminates fdtest (which means it deallocates its memory and does not return control to fdtest). Except that fdtest was in the middle of a BIOS call!!! So that's the problem. The fix is to ignore signals in fdtest, I'll patch it.
> 
OK; makes sense to me, thanks for the explanation. I guess something similar happens if the kernel preempts the fdtest process and goes to do something else. So we may need more than signal blocking in fdtest to fix this, right?
> I'm totally unfamiliar with the FP_SEG/FP_OFF etc macros, but it seems odd to me that the start_dma_page, end_dma_page and normalized_seg variables aren't used anywhere ...
> 
> That is some very technical stuff that calculates whether or not the I/O would overlap a 64k physical memory boundary, which is required for the old 8237 DMA controller; that chip is used by the BIOS. The calculations are only for display purposes in fdtest, the real use is in the ELKS BIOS driver. These were added to ensure that fdtest didn't need additional code.
> 
Ah, very educating indeed.

—Mellvik

ghaerr:
> I guess something similar happens if the kernel preempts the fdtest process and goes to do something else. So we may need more than signal blocking in fdtest to fix this, right?

You are correct, I didn't even think of that. The fdtest program should be run without any other tasks running, i.e. ktcp should not be running. There isn't any way currently to tell ELKS to stop multitasking, and the timer interrupt will cause a reschedule if any other tasks are runnable or became runnable through timeouts.

To ignore SIGINT, add signal(SIGINT, SIG_IGN) to main(), which will at least stop ^Cs from exiting fdtest.

When you've completed various test scenarios, submit a PR for the various enhancements/changes you've made to fdtest.

Issue #815, opened 2020-10-24
Mouse unusable in nx GUI on slower XT machines


pawosm-arm:
Mouse arrow movements are lagging massively making it unusable in nx GUI. I guess the way mouse is handled isn't very optimal and it's hitting hard the performance on XT machines. The projects like FreeGEM demonstrate that it is possible to achieve smooth mouse pointer movements on those machines.

ghaerr:
Lets see if the "fast" serial driver helps solve this problem (assuming you're using a serial mouse on com1: 

Try editing ports.h, uncomment "CONFIG_FAST_IRQ4" and comment out "CONFIG_NEED_IRQ4", then make clean; make.

Making sure the serial port is running as quickly as possible is the first step. Secondly, the mouse threshold and acceleration can be adjusted in elkscmd/nano-X/drivers/mou_ser.c:
```
#define SCALE       3   /* default scaling factor for acceleration */
#define THRESH      5   /* default threshhold for acceleration */
```

pawosm-arm:
> Try editing ports.h, uncomment "CONFIG_FAST_IRQ4" and comment out "CONFIG_NEED_IRQ4", then make clean; make.

I've tried this and it does not seem to be sufficient, there's still annoying lag.

> Secondly, the mouse threshold and acceleration can be adjusted in elkscmd/nano-X/drivers/mou_ser.cSecondly, the mouse threshold and acceleration can be adjusted in elkscmd/nano-X/drivers/mou_ser.c

What are meanings and accepted range for those values?

Mellvik:
@pawosm-arm,
just in case - did you check the active tty params on the port after the port has been activated (i.e. mouse is active)?

stty -a < /dev/ttyS0

… or whatever the name of the port is. Maybe there is a clue —

—Mellvik

pawosm-arm:
@Mellvik before running `nxdemo`:

```
speed 9600 baud; cs8 -parenb -parodd -hupcl -cstopb -cread -clocal
-ignbrk brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl -ixon
-ixoff -ixany
opost onlcr -xtabs
isig icanon -iexten echo echoe -echok -echonl -noflsh
eof = ^D eol = <undef> erase = ^H intr = ^C kill = ^U quit = ^\ susp = ^Z
start = ^Q stop = ^S min = 1 time = 0
```

after running `nxdemo`:

```
[...]

PR #817, opened 2020-10-26
[nano-X] Improve mouse throughput on slow systems

ghaerr:
Fixes issue #815.

SLOW_CPU is always set, as most systems aren't very vast running Nano-X. Works for both PC and MS mice.

@pawosm-arm, this is the same fix you tested, except enhanced to work for both mouse types.
BTW, if you want to see what Nano-X / Microwindows is capable of these days, check out the screenshots here: https://github.com/ghaerr/microwindows. Unfortunately, it needs 32-bit systems for that kind of graphics, the last version of Nano-X I was able to port to ELKS was done back in 1999!

### DIFF
--- a/elkscmd/nano-X/Makefile
+++ b/elkscmd/nano-X/Makefile
@@ -19,6 +19,8 @@ NANOXDEMOS += bin/nxdemo
 #NANOXDEMOS += bin/nxvgatest		# low-level vga routines test
 
 # mouse and kbd drivers
+# SLOW_CPU discards unprocessed mouse input for slow systems
+LOCALFLAGS += -DSLOW_CPU=1
 DRIVERS += drivers/mou_ser.o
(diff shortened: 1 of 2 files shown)


pawosm-arm:
Ouch, so if it diverged so much, maybe the choice of graphics environment needs to be reconsidered. I'm still thinking that graphics environment dedicated to low-end 8- and 16-bit systems of the late 80's may behave better performance-wise. John Elliot's fork of GEM from 2000's (FreeGEM) works smoothly on both of my XT's (8088-based low-end Turbo XT and 8086-based more featured Amstrad PC 2086), the graphics libraries behind it (GSX and VDI extensions) were designed with limitations of the contemporary machines in mind. It's true they are very much DOS oriented, yet some level of abstraction can still be achieved.

PR #823, opened 2020-10-27
[kernel] Implement disk drive track buffering for speed

ghaerr:
Here you go, @Mellvik! I'm anxious to hear how this works with your floppies. This is implemented based on your testing results from `fdtest` from #816. 

Implements full-track read buffering for BIOS devices. Should speed up floppy operations considerably.
Reads from requested sector to end of track into cache, then reuses data until cache miss, then repeats.
Best speed improvement should be seen when loading programs from floppies, or reading files without other disk activity.
No write buffering.
Possibly solves speed issues brought up in #521.

Uses low memory segment for BIOS I/O shared with DMASEG for compatibility with 8237 DMA controller and 64K boundaries.
Currently cache set to 18 sectors (9Kib).

BIOS Disk Drive Parameter Table (DDPT) not yet implemented - may not be compatible with all BIOSes.
Tested only with QEMU.

Adds new "debug" keyword in /bootopts to turn on debugging output at boot.
Turn on debug with ^P or bootopts "debug" to observe which sectors and being buffered, for analysis.

After testing, a config option will be added.


Mellvik:
Thank you @ghaerr!!!!!
Another incredibly speedy process - exciting.

I'm on my way to the testing, but before I get there, a couple of questions.
> ﻿
> Here you go, @Mellvik! I'm anxious to hear how this works with your floppies. This is implemented based on your testing results from fdtest from #816.
> 
> Implements full-track read buffering for BIOS devices. Should speed up floppy operations considerably.
> Reads from requested sector to end of track into cache, then reuses data until cache miss, then repeats.
> Best speed improvement should be seen when loading programs from floppies, or reading files without other disk activity.
> No write buffering.
> Possibly solves speed issues brought up in #521.
> 
Why not read 18 sectors regardless of track boundaries (unless hitting the end of the device of course)? The BIOS handles it just fine - and it's fast.
> Uses low memory segment for BIOS I/O shared with DMASEG for compatibility with 8237 DMA controller and 64K boundaries.
> Currently cache set to 18 sectors (9Kib).
> 
> BIOS Disk Drive Parameter Table (DDPT) not yet implemented - may not be compatible with all BIOSes.
> Tested only with QEMU.
> 
I'm not familiar with the DDPT, and I support that it's just being ingored till we have a care to test against. Over the years I've done a lot of 'manual' BIOS disk/floppy I/O via DOS DEBUG and QBASIC (!) for hw test/debugging w/o any problems, so either DOS is very good at using and updating the DDPT, or it's an uncommon problem. It is wise - as suggested - to take a close look at what FreeDOS does about this.


Test report is coming.

--M

Mellvik:
@ghaerr,

First tests - printks active, no timings:

Booting and the doing a dozen fil system oriented commands (rm, mv, ftpget, chmod, cd etc.) left me with a total of 364 'bioshd:' messages of which 267 (73%) are cache hit messages. That's more than pretty good AFAIK.

Taking out the messages (thanks ^P), ls -l /bin is substantially faster than before, with a noticeable stop when reading more data. This lag obviously was there before too, now it's higher, but just barely so. This issue - increased lag during reads, needs more practical use to evaluate.

The dd-test with bs=18b is - as could be expected - just great: 

# time dd if=/dev/fd0 of=/dev/null bs=18b
160+0 records in
160+0 records out
1474560 bytes (1440 KiB) copied

Real 1m6.820s
# 

This is an almost 6 times improvement. Using other (lower) block sizes (including one) with dd deliver the same timing (within a reasonable margin of error).

An interesting side effect of the buffering (may be considered a bug, but seems to work just fine) is that it applies to the hard disk too, at the mount point, when the HD is mounted on a floppy root:

…...
bioshd: buffer sectors 63 to 80
bioshd: cache hit sector 63
bioshd: cache hit sector 65
bioshd: cache hit sector 67
bioshd: buffer sectors 13671 to 13688
bioshd: cache hit sector 13671
bioshd: cache hit sector 13687
# pwd
… … … 

Out of curiosity I modified fdtest to read full tracks from start to end of device, to establish a baseline, first the 1.2M drive, then the 1.4M drive:

# time fdtest -i 15 -d 1 -D -m 15
Testing entire drive 1, max 15 incr 15
READ ERROR at CHS 80,0,1

Read 160 blocks in 43 seconds

Real 0m42.890s
# time fdtest -i 18 -D -m 18
Testing entire drive 0, max 18 incr 18
READ ERROR at CHS 80,0,1

Read 160 blocks in 51 seconds

Real 0m50.620s
# 

I guess this is what we can call 'physical speed' in this setting.


Thanks again for a great contribution to a more speedy ELKS. Maybe some of it may be used on hard drives as well?


—Mellvik

ghaerr:
@Mellvik, thanks for your tests! I am glad this is working! I am traveling today so I might not be able to answer all your questions, but will get to it. As I understand further how track-increment may work using BIOS, we can more further explore possible multi-track buffering (or not, depending on older BIOSes, bugs, and @tkchia's input).

Here is a full patch that implements a DDPT, and installs a new one should the max_track value in the DDPT not be the same as the ELKS drive geometry. You're welcome to add and test this also. It can be disabled by commenting out the set_ddpt and reset_ddpt calls:
```
diff --git a/elks/arch/i86/drivers/block/bioshd.c b/elks/arch/i86/drivers/block/bioshd.c
index e9dbde65..249748b5 100644
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -749,11 +749,36 @@ static int do_bios_readwrite(struct drive_infot *drivep, int cmd,
 static sector_t cache_startsector;
 static sector_t cache_endsector;
 
+static unsigned long __far *vec1E = _MK_FP(0, 0x1E << 2);
+static unsigned long oldvec;
+
+int set_ddpt(unsigned int max_sector)
+{
+       oldvec = *vec1E;
+       unsigned char __far *org_ddpt = (void __far *)oldvec;
+       static unsigned char ddpt[12];
+
+       if (org_ddpt[4] == (unsigned char)max_sector)
+               return 0;
+
+       dprintk("DDPT %d\n", org_ddpt[4]);
+       fmemcpyw(ddpt, _FP_SEG(ddpt), (void *)(unsigned)oldvec, _FP_SEG(oldvec), 6);
+       ddpt[4] = (unsigned char)max_sector;
+       *vec1E = (unsigned long)(void __far *)ddpt;
+       return 1;
+}
+
+void reset_ddpt(void)
+{
+       *vec1E = oldvec;
+}
 /* read from start sector to end of track into DMASEG track buffer*/
 static void bios_readtrack(struct drive_infot *drivep, sector_t start)
 {
        unsigned int cylinder, head, sector, num_sectors;
        int drive = drivep - drive_info;
+       int err, set;
 
        drive = hd_drive_map[drive];
        get_chst(drivep, start, &cylinder, &head, &sector, &num_sectors);
@@ -765,8 +790,10 @@ static void bios_readtrack(struct drive_infot *drivep, sector_t start)
        BD_DX = (head << 8) | drive;
        BD_ES = DMASEG;
        BD_BX = 0;
-       //FIXME add DDPT
-       if (call_bios(&bdt)) {
+       set = set_ddpt(drivep->sectors);
+       err = call_bios(&bdt);
+       if (set) reset_ddpt();
+       if (err) {
                printk("bioshd: track read error sector %d num %d\n", sector, num_sectors);
                set_cache_invalid();
                return;
@@ -850,12 +877,12 @@ static void do_bioshd_request(void)
        if (req->rq_cmd == READ) {
            if (cache_valid(drivep, req, start)) {      /* try cache first*/
                end_request(1);
-               return;
+               continue;
            }
            bios_readtrack(drivep, start);              /* read whole track*/
            if (cache_valid(drivep, req, start)) {      /* try cache again*/
                end_request(1);
-               return;
+               continue;
            }
        }
        set_cache_invalid();
```

ghaerr:
> Maybe some of it may be used on hard drives as well?

At this point, the track caching is enabled for both floppies and hard disks. I am not sure whether or not HD needs track caching, and copying from floppy to HD would invalidate the cache since they use the buffer, so it should probably be restricted to floppies only. The DDPT is also for floppies only, but this code doesn't check that. So there's ultimately more to be done than a days work :)

Mellvik:
Thanks @ghaerr,

I'll include this one right away. Now, mail expands all tabs and git apply barfs - maybe you can send it as an attachment, or is there a simple trick to get the tabs back?


—Mellvik

ghaerr:
> maybe you can send it as an attachment, or is there a simple trick to get the tabs back?

The website version can be copied/pasted usually without error. I don't think there's a need yet for you to add this. Lets see what @tkchia is coding up and has to say; the DDPT patch is only seemingly required when the max_sector value differs from the disk geometry that's in the drive. I don't yet understand how/when the BIOS updates its own ROM DDPT when different disks are inserted in the drive. We may need to add this to `fdtest` for further analysis.

Mellvik:
> 
> maybe you can send it as an attachment, or is there a simple trick to get the tabs back?
> 
> The website version can be copied/pasted usually without error. I don't think there's a need yet for you to add this.
> 
OK (source availability).
Adding: I'm curious enough to apply it and verify that it works on the Compaq - which is known to be an oddball BIOS wise. 
> Lets see what @tkchia <https://github.com/tkchia> is coding up and has to say; the DDPT patch is only seemingly required when the max_sector value differs from the disk geometry that's in the drive. I don't yet understand how/when the BIOS updates its own ROM DDPT when different disks are inserted in the drive. We may need to add this to fdtest for further analysis.
> 
> 
Indeed. I added a bunch of options and the ability to time the reading of the entire drive to fdtest - let me know if I should upload it (after including your signal fix).

—Mellvik

ghaerr:
Hello @Mellvik,

I've added the DDPT handling to this PR, go ahead and give it some testing when you have time.

There is still a caching issue with 360k and 720k floppies - due to an implementation issue on my part, the last (odd-numbered) sector on every other track, and the subsequent first sector on the next track, are split across a 1k block (upper level) boundary, and thus aren't cached. This shows up on any disk with an odd number of sectors-per-track.

For now, I'm primarily interested in how well it works on 1.44M floppies, and whether the internal DDPT handler causes any issues. If this works well, it will be committed, and the above issue handled separately.

Hello @tkchia,

Thanks for your help on this in #521. I got a copy of the Award BIOS source code, and found that it, and apparently some other BIOSes, have extended the DDPT from 11 to 13 bytes internally. Then I found [this](http://www.delorie.com/djgpp/doc/rbinter/it/64/12.html) which shows the IBM SurePath BIOS as having 14 bytes. The 12th and 13th offsets match those in the Award BIOS. FYI!

For the DDPT implementation, I reasoned that some BIOSes may attempt to reset the DDPT 1Eh vector to another ROM location when floppies are changed or for other reasons. Thus, rather than putting the DDPT copy in the probe routine, I settled on copying 14 bytes once in the driver init routine, then setting the proper Sectors-Per-Track in our DDPT and reseting the vector before each read/write/probe call as a precaution of the BIOS repointing the INT 1Eh vector back to ROM.

The DDPT handler is also now used in your original BIOS read/write routine, because it reads two sectors at a time and could possibly be subject to problems on track wrap, even though we've never seen them.

The next step is going to be eliminating the code and issues around the 64KiB wraparound problem of the 8237 DMA controller. This will be done by allocating the 1024-byte L1 buffers on a 1KiB memory boundary, rather than having them be arbitrarily placed by the linker in the .bss segment; the L2 buffers will also be 1KiB-aligned in main memory.

Thank you!

Mellvik:
Thanks @ghaerr,
> I've added the DDPT handling to this PR, go ahead and give it some testing when you have time.
> 
> 
Will do asap! Thanks.
> There is still a caching issue with 360k and 720k floppies - due to an implementation issue on my part, the last (odd-numbered) sector on every other track, and the subsequent first sector on the next track, are split across a 1k block (upper level) boundary, and thus aren't cached. This shows up on any disk with an odd number of sectors-per-track.
> 
> 
I'm in the middle of this, and just managed to format a 720k floppy on a physical machine (DOS) and make it DOS bootable.
The next step is to see if ELKS can boot the image, which needs adjustments in the ELKS configuration: It is not currently possible to build 720k boot images. Can you have a look at that, @ghaerr?
> For now, I'm primarily interested in how well it works on 1.44M floppies, and whether the internal DDPT handler causes any issues. If this works well, it will be committed, and the above issue handled separately.
> 
> 
Agreed. Then again, I've been using 1.2M floppies all along with ELKS. They are 15 SPT, so odd sectored devices (other than 720k) have been working OK all along. 
Testing this drive with the new buffering seems to work fine. I copied /bin from fd0 to fd1 (which takes a loooong time still, writing is indeed slow), and then from fd1 to hd1. There are a number of double NO-CACHE messages and there is no doubt they affect performance.
Also, as the snippet below shows, there are some interesting cases where buffering does not seem to work:
—
bioshd: buffer sectors 36 to 44
bioshd: cache hit sector 36
bioshd: buffer sectors 990 to 1004
bioshd: cache hit sector 990
bioshd: NO-CACHE read sector 14 len 1
bioshd: NO-CACHE read sector 15 len 1
bioshd: buffer sectors 32 to 35
bioshd: cache hit sector 32
bioshd: buffer sectors 551 to 566	<------
bioshd: cache hit sector 551
bioshd: buffer sectors 553 to 566	<------
bioshd: cache hit sector 553
bioshd: buffer sectors 28615 to 28632
bioshd: cache hit sector 28615
bioshd: cache hit sector 28631
bioshd: buffer sectors 28633 to 28650
—

The 2nd marked line indicates a re-read of already buffered content.

Finally, 
While doing this I was missing the device # from the 'bioshd: debug' messages BTW. Than would be a useful addition.

More about 720k floppies when I get the config working.


—Mellvik

Mellvik:
And one more thing, @ghaerr.

The compaq is adamant about keeping INT13 requests within the track. Requesting 2 sectors starting @ 18 (or 15 for the 1.2 drive) cause an error return to fdtest. Likewise for requesting 20 starting at 1.

So you're definitely on the right track here!


—M

ghaerr:
> It is not currently possible to build 720k boot images.

You mean with menuconfig? I'll look at that. The workaround is to build all the images: after "make" do a "cd image; make images". This will build a fd720-minix.bin.

> there are some interesting cases where buffering does not seem to work:

Yes, this is because the track cache is invalidated on any floppy write. It does this because we don't keep a seperate cache for each BIOS device. Getting write caching working will be considerably more complicated, and considered later.

> While doing this I was missing the device # from the 'bioshd: debug' messages 

Ok, will add.

> The compaq is adamant about keeping INT13 requests within the track. Requesting 2 sectors starting @ 18 (or 15 for the 1.2 drive) cause an error return to fdtest. Likewise for requesting 20 starting at 1.

Good to know, as this is different from some statements possibly made about INT 13h in the past. 

To be clear, then: some BIOSes support "multitrack" where the CHS is properly incremented when reading past track end, but the Compaq does not. So ELKS cannot use INT 13h multitrack, ever.

Thanks for your testing!

When ELKS requests reading a 1k block that just happens to start on the last sector of a floppy with an odd sectors-per-track, the current code reads that one sector, then also only the first sector of the next track. I am thinking about how to update the code to read that first sector, but then cache the whole next track. This means there will still be the case of reading a single sector, but not two single sectors before the next track cache. I think this will be the best we can hope for: reading whole tracks at a time, starting at the first sector requested.

ghaerr:
> It is not currently possible to build 720k boot images.

Yes, it works. Remember, by default you have all the applications selected, so that there's not enough space.
Under "Userland", the "Select Applications by Image Size" has to be selected, then 720k also.

Unless you're doing it a lot, I think it is easier to just make all the images using "make images".

Mellvik:
> 
> The compaq is adamant about keeping INT13 requests within the track. Requesting 2 sectors starting @ 18 (or 15 for the 1.2 drive) cause an error return to fdtest. Likewise for requesting 20 starting at 1.
> 
> Good to know, as this is different from some statements possibly made about INT 13h in the past.
> 
> To be clear, then: some BIOSes support "multitrack" where the CHS is properly incremented when reading past track end, but the Compaq does not. So ELKS cannot use INT 13h multitrack, ever.
> 
> 
Update on this one: Yes, this is indeed different from past experience and from what I've said recently. Unsurprisingly, Qemu is very flexible about this.

And Correction: MY first statement about the compaq was incorrect: Switching heads inside a read is fine, changing cylinder is not.


—M

Mellvik:
I stand corrected - tank you!

Mellvik:
No, I don't think this is going to happen all that often :-)

Anyway - FWIW - booting elks from a 720K image works fine on the compaq. 
I could test booting off of 1.2M and 360K images too, let me know if it's useful.


—M

ghaerr:
Hello @tkchia and @Mellvik,

I have been reading the Award 6.00 BIOS source code, which also just happens to be installed on my 386 desktop machine. Some interesting findings on DDPT and the BIOS Parameter Block (BPB) below.

The Award BIOS installs a DDPT at INT 1Eh according to the configured boot floppy drive. The BIOS never changes the DDPT, so our ELKS enhancement of using a custom DDPT has no effect. However, the Award BIOS doesn't check the passed sector number in INT 13h Function 02 (read sectors) against a max SPT, so the ELKS track reads will always read to the specified last sector on whatever disk is in the drive.

Thus, the Award BIOS should be able to work with the ELKS enhancement and in testing, it seems to do just fine.

I did find that the Award BIOS will reset certain values in the BPB whenever the boot sector (CHS 0/0/1) is read or written. So there is some hidden communication between the BIOS and BPB, which is undoubtably used by MSDOS (and unneeded by ELKS). In reading various emulator BIOS source, they also do the same thing, with some updating their internal DDPT. I haven't yet found one that updates a custom DDPT. This probably means the current code that forces the INT 1Eh vector to our custom DDPT on each read is unneeded, and possibly bad.

I added a boot display of the initial BIOS DDPT sectors-per-track (SPT) value, should we want to get into debugging what the BIOS does when disks are swapped. This may only be a problem on certain BIOSes where the system is booted from HD, and then ELKS is asked to read a floppy which is different than the BIOS boot configuration for that drive. In that case, ELKS will force its idea of the drive SPT by patching the DDPT. This may only be a problem on a BIOS which checks max SPT on reads, such as Compaq. If the BIOS doesn't recognize our custom DDPT SPT, then we may have to find an additional way to tell the BIOS what the max SPT is. 

Frankly, I'm beginning to wonder whether a custom DDPT is actually necessary at all, in both the BIOS driver as well as the boot code. It would be nice to find a BIOS that requires it, for verification.


@Mellvik - The bioshd: messages now show the BIOS drive number, to help your testing. Note that this is the BIOS drive number internal to the driver, so HD will show as 0, and fd0 is 4, fd1 is 5. 

Thank you!

Mellvik:
> When ELKS requests reading a 1k block that just happens to start on the last sector of a floppy with an odd sectors-per-track, the current code reads that one sector, then also only the first sector of the next track. I am thinking about how to update the code to read that first sector, but then cache the whole next track. This means there will still be the case of reading a single sector, but not two single sectors before the next track cache. I think this will be the best we can hope for: reading whole tracks at a time, starting at the first sector requested.
> 
> 
@ghaerr, I fully agree. 
I think this is a good idea, and the best we can do given what we know at this time.

—M

Mellvik:
> 
> Hello @tkchia <https://github.com/tkchia> and @Mellvik <https://github.com/Mellvik>,
> 
> I have been reading the Award 6.00 BIOS source code, which also just happens to be installed on my 386 desktop machine. Some interesting findings on DDPT and the BIOS Parameter Block (BPB) below.
> 
> The Award BIOS installs a DDPT at INT 1Eh according to the configured boot floppy drive. The BIOS never changes the DDPT, so our ELKS enhancement of using a custom DDPT has no effect. However, the Award BIOS doesn't check the passed sector number in INT 13h Function 02 (read sectors) against a max SPT, so the ELKS track reads will always read to the specified last sector on whatever disk is in the drive.
> 
> Thus, the Award BIOS should be able to work with the ELKS enhancement and in testing, it seems to do just fine.
> 
> I did find that the Award BIOS will reset certain values in the BPB whenever the boot sector (CHS 0/0/1) is read or written. So there is some hidden communication between the BIOS and BPB, which is undoubtably used by MSDOS (and unneeded by ELKS). In reading various emulator BIOS source, they also do the same thing, with some updating their internal DDPT. I haven't yet found one that updates a custom DDPT. This probably means the current code that forces the INT 1Eh vector to our custom DDPT on each read is unneeded, and possibly bad.
> 
> I added a boot display of the initial BIOS DDPT sectors-per-track (SPT) value, should we want to get into debugging what the BIOS does when disks are swapped. This may only be a problem on certain BIOSes where the system is booted from HD, and then ELKS is asked to read a floppy which is different than the BIOS boot configuration for that drive. In that case, ELKS will force its idea of the drive SPT by patching the DDPT. This may only be a problem on a BIOS which checks max SPT on reads, such as Compaq. If the BIOS doesn't recognize our custom DDPT SPT, then we may have to find an additional way to tell the BIOS what the max SPT is.
> 
> Frankly, I'm beginning to wonder whether a custom DDPT is actually necessary at all, in both the BIOS driver as well as the boot code. It would be nice to find a BIOS that requires it, for verification.
> 
> @Mellvik <https://github.com/Mellvik> - The bioshd: messages now show the BIOS drive number, to help your testing. Note that this is the BIOS drive number internal to the driver, so HD will show as 0, and fd0 is 4, fd1 is 5.
> 
> Thank you!
> 
> 
Thanks again, @ghaerr. 
Again, I agree. It makes sense to wait for the problem to appear, then attack it.

And to do the same with the 720k Odd Bios thing. Ensure general 720k support (which we have) and then attack a problem if/when it pops its head.

—Mellvik

tkchia:
Hello @ghaerr, hello @Mellvik,

> However, the Award BIOS doesn't check the passed sector number in INT 13h Function 02 (read sectors) against a max SPT, so the ELKS track reads will always read to the specified last sector on whatever disk is in the drive.

But I am guessing it does pass the "sectors per track" value to the floppy disk controller?    The ["Floppy Disk Controller"](https://wiki.osdev.org/Floppy_Disk_Controller) entry on the OSDev Wiki suggests that "multitrack" mode can be turned on at the floppy disk controller level.  It might be useful to see how different BIOSes decide on this setting.

Thank you!

Mellvik:
> 
> But I am guessing it does pass the "sectors per track" value to the floppy disk controller? The "Floppy Disk Controller" entry on the OSDev Wiki suggests that "multitrack" mode can be turned on at the floppy disk controller level. It might be useful to see how different BIOSes decide on this setting.
> 
> Thank you!
> 
I belive this is the normal. For the nec765 , multitrack means both sides, which is what the compaq (maybe most systems) do. Multicylinder not supported.

- Mellvik

ghaerr:
@Mellvik,

I revised the track cache to handle odd-sectored floppies, so this enhancement should now run best-case for all floppies.

The old driver read code is still in, as a fallback, should reads through end-of-track fail or retries are needed.

I've also added an option (USE_DDPT) for the DDPT custom handler, which is now default OFF, along with commenting out the hack fix for the 720k floppy four years ago which inhibits performance. At this point, it's probably better to see what is actually needed before adding more code and complexity.

There are no multi-track sector reads ever requested, so I think we're OK without a need for the custom DDPT, since the original driver didn't use one either.

If your testing goes well, I'll add a BIOS Floppy/HD Track Cache config option and commit this, which should allow us to close #521 for now. As stated previously, floppy writes are not easily sped up with a cache, since the data must be written immediately, and we only get data two sectors at a time (one Linux block) from the upper level.

Mellvik:
Thanks @ghaerr,
also for the improved bioshd messages.

Testing completed, this just works. Tested on 1.2M, 1.44M floppies.
General perception of responsiveness on a floppy based system is significantly improved. Login feels 3 times faster, more like 3 seconds now.

dd timings are just great.

```
# time dd if=/dev/fd0 of=/dev/null bs=18b
160+0 records in
160+0 records out
1474560 bytes (1440 KiB) copied

Real 1m6.160s
# time dd if=/dev/fd0 of=/dev/null bs=9b
320+0 records in
320+0 records out
1474560 bytes (1440 KiB) copied

Real 1m4.430s
# time dd if=/dev/fd0 of=/dev/null bs=2b
1440+0 records in
1440+0 records out
1474560 bytes (1440 KiB) copied

Real 1m4.440s
# 
```

So - in sum a dramatic improvement in average, real and perceived response on floppy based systems. This rocks.

Thank you!

Final question - this time on writing: My memory may be unreliable, but I seem to remember seeing some of the same block handling mechanisms in ELKS as are used in unixes and other linuxes - a queueing system with sorting etc for asynchronous interrupt driven write. If this is the case and elks write is indeed synchronous, there must be a significant potential for improvement?

--M

tkchia:
Hello @ghaerr,

> There are no multi-track sector reads ever requested, so I think we're OK without a need for the custom DDPT, since the original driver didn't use one either.

I am more concerned about the possibility of ELKS _accidentally_ doing a multitrack read when it only really wants to do a single-track and single-side read.

Though it is possible that I have been misunderstanding how the FDC's "multitrack" facility works, and such accidents will never happen (and that will be a good thing).  Perhaps I will look into augmenting `fdtest`'s functionality to really stress-test the BIOS and FDC's response to changes in the "sectors per track" DDPT field.

Thank you!

ghaerr:
@Mellvik,

> So - in sum a dramatic improvement in average, real and perceived response on floppy based systems. This rocks.

Great news! Your particular testing was done without a custom DDPT handler, so we know your BIOS works without it. However, from what we're discussing in #824, I'll need to add back in the DDPT handler for compatibility with the IBM XT and AT BIOSes. This won't have any affect on the disk throughput speed however, so retesting won't be needed.

We may need to re-test on 720k floppies, as the #39/#44 is removed in this PR, but shouldn't be needed with a patched DDPT handler added back in. Ultimately, we need to find an IBM with an XT BIOS to test on to prove both the DDPT and 720k fixes still work with the track caching code.

> I seem to remember seeing some of the same block handling mechanisms in ELKS as are used in unixes and other linuxes - a queueing system with sorting etc for asynchronous interrupt driven write. 

Yes, that is all present, although completely untested, since there are no disk I/O interrupts when using ELKS' only working FD/HD driver - the BIOS driver.

> If this is the case and elks write is indeed synchronous, there must be a significant potential for improvement?

The ELKS BIOS read and write code are certainly synchronous, and can't be interrupted, as your `fdtest` crashes have shown. 

But even writing an interrupt driven floppy driver wouldn't immediately speed up any write operations, as the sorted I/O queue would have to be aggregated into single multi-sector track-cache writes with more special code and buffer to cache copying. 

We could write that special code today without interrupt handling, except for one big problem: the track-caching code only works with consecutive physical sectors. Any disk file written would have to use previously untouched disk blocks from a newly-mkfs'd disk, or the sectors wouldn't be consecutive. So floppies would quickly degrade with disk fragmentation. The only reason we're getting big improvements on the read track caching is that the filesystem created by ELKS creates all files using the untouched MINIX free list, which allocates sectors sequentially.

Mellvik:
> We may need to re-test on 720k floppies, as the #39 <https://github.com/jbruchon/elks/issues/39>/#44 <https://github.com/jbruchon/elks/pull/44> is removed in this PR, but shouldn't be needed with a patched DDPT handler added back in. Ultimately, we need to find an IBM with an XT BIOS to test on to prove both the DDPT and 720k fixes still work with the track caching code.
> 
> 
I realize you didn't refer to the Compaq, but I'll check that it still boots off of 72k floppies anyway.
> I seem to remember seeing some of the same block handling mechanisms in ELKS as are used in unixes and other linuxes - a queueing system with sorting etc for asynchronous interrupt driven write.
> 
> Yes, that is all present, although completely untested, since there are no disk I/O interrupts when using ELKS' only working FD/HD driver - the BIOS driver.
> 
> If this is the case and elks write is indeed synchronous, there must be a significant potential for improvement?
> 
> The ELKS BIOS read and write code are certainly synchronous, and can't be interrupted, as your fdtest crashes have shown.
> 
> But even writing an interrupt driven floppy driver wouldn't immediately speed up any write operations, as the sorted I/O queue would have to be aggregated into single multi-sector track-cache writes with more special code and buffer to cache copying.
> 
Sorry for my lack of clarity: I meant the exact opposite - getting rid of it since it isn't useful. I gather from your answer though that it's not even used, so it's a non issue. 
> We could write that special code today without interrupt handling, except for one big problem: the track-caching code only works with consecutive physical sectors. Any disk file written would have to use previously untouched disk blocks from a newly-mkfs'd disk, or the sectors wouldn't be consecutive. So floppies would quickly degrade with disk fragmentation. The only reason we're getting big improvements on the read track caching is that the filesystem created by ELKS creates all files using the untouched MINIX free list, which allocates sectors sequentially.
> 
I do realize that, and I keep thinking we're fine with what we now got. What we're gaining here is a much more responsive floppy-root based system. Really - we have a multitasking, multiuser, networked floppy based system with decent performance on (quite) slow HW. How'bout that? 
Even though we (maybe me in particular) have been running elks off of physical floppies a lot over the part two years, the floppy-base must be considered temporary, a step towards HD installation and a more decent environment.
I think it is time to move our attention to other areas in need of improvement.

Now, since you mentioned it the other day, I also think it's time to yank up the ELKS version #. We need to fix the FAT large file issue, which I'm returning to now that the floppy issues are done. 
After that - and given the level ELKS now is at, I'd say what we're ready for ELKS 1.0.

—Mellvik

ghaerr:
Final version ready for commit. In summary:
- BIOS drive track caching is a config option (CONFIG_TRACK_CACHE), default ON.
- Adds patched DDPT for all BIOS read/write calls, fixes BIOS driver for early BIOSes.
- Removes 'dma wait' critical section, as BIOS I/O will never sleep a process.
- Removes separate patch for #39/#44, as patched DDPT releases the need for it.
- Adds "debug" option in /bootopts to toggle startup debug setting.
- Adds DEBUG_BIOS in debug.h for debug_bios debugging macros.

PR #825, opened 2020-10-31
[cmd] Functional enhancements to fdtest

Mellvik:
Adding functionality to @ghaerr's `fdtest` program - expanding speed tests to cover BIOS and drive testing.

`fdtest` now has two modes: 
- Standard mode: Runs (and reports elapsed time) 3 tests on the selected drive (single block, double block, entire track), with selectable starting points.
- Direct mode: Reads (and reports time consumption) the entire drive using default or command line parameters

`fdtest` reports errors and continues until the requested test has completed. Thus is is possible to stress-test the BIOS in one or a few runs.

The command line parameters are:
```
-d <device> Read from device 0 or 1, default 0
-D Enter direct mode, read entire drive, ignore parameters adjusting the starting point
-b <blocksize> Read <blocksize> sectors per operation, max 25, default 1, intentionally not checked against SPT
-h <head> Starting head, 0 or 1, default 0, ignored in direct mode
-u Print usage and exit
-t <spt> Set sectors per track, default is 18
-s <sector> Starting sector (normal mode)
-v Verbose output
-c <cyl> Starting cylinder (normal mode)
-o Send read data to stdout - remember to redirect
```
Interrupts (^C) are disabled - a forced exit will hang or seriously screw up there system. Similarly, `fdtest` must be run on a totally quiet system. If preempted by a context switch, the system will most likely hang hard.

### DIFF
--- a/elkscmd/misc_utils/fdtest.c
+++ b/elkscmd/misc_utils/fdtest.c
@@ -27,7 +27,6 @@ struct biosparms bdt;
 #define MAX		30		/* max # of sectors per read */
 #define SECSIZE		512		/* sector size*/
 
-unsigned char user_buf[MAX * SECSIZE];
 unsigned char kernel_buf[MAX * SECSIZE];
 int verbose = 0;
 int out = 0;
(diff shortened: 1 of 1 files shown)


ghaerr:
Wow @Mellvik, you went to town on this! Nice enhancements!

Would you like to remove the code still commented out, for cleanup? "user_buf" can also be deleted.

Thank you!

ghaerr:
Thanks for the nice enhancements @Mellvik!

PR #826, opened 2020-10-31
[boot] copy 14 bytes, not just 11, of floppy parameter data to RAM

tkchia:
The Diskette Drive Parameter Table (DDPT) at `int $0x1e` is normally 11 bytes, but Ralf Brown's Interrupt List says that the IBM SurePath BIOS uses a 14-byte table.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -127,7 +127,10 @@ _next0:
 	mov $floppy_table,%di
 .endif
 	push %di
-	mov $6,%cl     // 12 bytes (actually 11 in the table) and CH = 0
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

The source code to the original IBM BIOSes shows that the DDPT was increased to 13 bytes as early as IBM XT BIOS v2 (1986), so it looks like the IBM SurePath version must have increased it one more to 14 bytes.

Thank you!

tkchia:
Hello @ghaerr,

It does look that way.  RBIL describes the extra fields as
```
---IBM SurePath BIOS---
 0Bh    BYTE    maximum track number
 0Ch    BYTE    data transfer rate
 0Dh    BYTE    drive type in CMOS
```
Thank you!

Issue #824, opened 2020-10-30
BIOS disk interface and other data structures


ghaerr:
Hello @tkchia,

I thought to move the discussion of how ELKS should work with the BIOS to get/set disk drive parameters here from #823 and #521, since it was getting increasingly off topic.

I have been reading various BIOS source code, as well as looking hard at the ELKS BIOS driver and boot code. The BIOS sources include Award 6.00 source, used in many machines including my own desktop 386, the SeaBios source, used in the QEMU emulator, EMU86 which emulates certain BIOS calls, and the tiny8086 emulator, which provides its BIOS source. All are quite illuminating, although sometimes complicated.

The purpose of this discussion is to better understand how ELKS should interface with a BIOS, with a short term plan to enhance and cleanup the ELKS BIOS driver and the MINIX and FAT boot code, which are all quite messy. With some thought, I think all these can be cleaned up considerably while maintaining known compatibility with older PCs. 

A longer term goal is to understand exactly what external interfaces (BIOS calls, BIOS data areas, special I/O ports, etc) ELKS is reliant on, so that these can be consolidated, to help with a port of ELKS to more modern SBCs, or even another CPU.

The first topic on this subject is "How is the floppy sectors per track (SPT) communicated with the BIOS?"

Increasingly, it is looking like the Diskette Drive Parameter Block (DDPT), sometimes called the Disk Base Table (DBT) may be used to publish BIOS SPT for non-BIOS use, but not to set BIOS SPT. In all BIOS source checked, BIOS INT 13h function 02/03 (read/write sectors) never looks at a "custom DDPT" SPT field to get max sector information. None of the BIOS code ever checks for a rewritten INT 1E, and most never update the DDPT after boot. In several BIOSes, the only time the internal DDPT is accessed is to return it for INT 13h function 08 (get disk parameters).

Thus, the current ELKS "custom DDPT" approach of resetting INT 1Eh and replacing the DDT SPT field should be considered for removal.

It is complicated how each of the BIOSes come up with the SPT to use in the INT 13h read/write calls. Two of them use use the BPB present in a FAT boot sector. For certain drive types, the BIOS will check for validity, then copy the BPB SPT after reading or writing the boot sector (CHS 0/0/1). Thus, it seems that a possible good place to store and communicate SPT is in the ELKS boot sector.

[...]

PR #827, opened 2020-11-01
[kernel] Fix BIOS disk driver bad floppy count with early BIOSes

ghaerr:
Fixes incorrect floppy count on machines with early BIOSes (pre-1986, IBM XT v1 BIOS and earlier) where INT 13h function 8 (get drive parameters) is not implemented for floppy drives.

Bug found and tested by @toncho11 in https://github.com/jbruchon/elks/issues/824#issuecomment-720127139.

Although INT 13h function 8 is not implemented by the BIOS in these early machines, the BIOS scans for adapter ROMs, and the hard drive ROM does implement INT 13h functions 00-14 as discussed in [IBM PC BIOS version history](https://sites.google.com/site/pcdosretro/ibmbioshist). So, no special handling is required for hard drives.

When the Get Drive Parameters function is not implemented, this fix uses the BIOS data area to get the floppy drive count.

Also more cleanup up bioshd.c source code around this fix.


ghaerr:
Remove ASM floppy query code in setup.S.
Properly interpret BIOS data area floppy disk count (same as INT 11h).
Only call INT 13h function 8 (Get disk parameters) on AT systems or greater, otherwise default to BIOS data equipment flags.
Remove CONFIG_HW_FLOPPY, was used only in setup.S.
Remove CONFIG_HW_USE_INT13_FOR_FLOPPY, now handled without need for option in BIOS driver.

PR #828, opened 2020-11-02
[kernel] Remove need for DMASEG 64K wrap handling in BIOS disk driver

ghaerr:
Removes special handling for BIOS I/O requests copying data to/from DMASEG.
Replaces early_printk with early_putchar to enable kernel `printk` to be used before console initialization. (Used to debug this PR).

This performance enhancement and code reduction is made possible by adding a new flag, SEG_ALIGN1K to the system global memory manager `seg_alloc`, and allocating all L2 buffers on an even 1KiB physical memory boundary. Since L2 buffers are allocated up to 64k at a time, memory penalties are minor.

In addition, a previous PR (#819) removed requiring `map_buffer` calls before any I/O, so that L1 buffer alignment (in the kernel data segment) is no longer an issue here.

BIOS I/O can now be started directly into an L2 buffer, with possible BIOS DMA by the 8237 hardware wraparound at 64K boundaries never being an issue, since the L2 buffers are always sized 1024 bytes and allocated on a 1024-byte boundary. Kernel memory copies before or after the I/O are no longer necessary, and that code and the testing for wraparound can be removed. This was a lot of extra code to handle only 1 out of every 64 buffers!

The DMASEG is now only used when BIOS track caching is enabled (default ON), as well as for disk probing. 

Tested and confirmed working on QEMU and real hardware. The L2 buffer allocation segments can be observed using `meminfo`.

### DIFF
--- a/elks/arch/i86/boot/crt0.S
+++ b/elks/arch/i86/boot/crt0.S
@@ -13,7 +13,7 @@
 
 	.global _start
 	.extern	start_kernel
-	.global early_printk
+	.global early_putchar
 
 _start:
 
@@ -58,25 +58,12 @@ _start:
 int3:	int	$3		// C breakpoint for emu86
 	ret
 
-early_printk:
-	push  %bp
-	mov   %sp,%bp
-	push  %si
-	mov   4(%bp),%si
-	cld
-
-ep_loop:
-	lodsb
-	or    %al,%al
-	jz    ep_end
+early_putchar:
+	mov   %sp,%bx
+	mov   2(%bx),%al
 	mov   $0x0E,%ah
 	mov   $0x0007,%bx
 	int   $0x10
-	jmp   ep_loop
-
-ep_end:
-	pop   %si
-	pop   %bp
 	ret
 
 //	Segment beginnings
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -694,51 +694,21 @@ static int do_bios_readwrite(struct drive_infot *drivep, sector_t start, char *b
 
 	errs = MAX_ERRS;	/* BIOS disk reads should be retried at least three times */
 	do {
-		unsigned rq_start_dma_page, rq_end_dma_page;
-		int need_dma_seg;
-
-		/* Try to gauge if we can do the read/write directly on the actual source/
-		 * target buffer without crossing a 64 KiB DMA boundary
-		 * If not, then do the read/write by way of DMASEG:0
-		 */
-		rq_start_dma_page = (seg + ((unsigned) buf >> 4)) >> 12;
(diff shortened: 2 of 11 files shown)


tkchia:
Hello @ghaerr,

Thank you.  If the disk I/O works, then I am mostly fine with the patch.

Is there any particular reason why the cursor position is now shifted down one row (`C->cy++` etc.) when initializing the console drivers?  What happens if the cursor is already on the bottom row?

ghaerr:
Hello @tkchia,

> If the disk I/O works, then I am mostly fine with the patch.

I'm interested in your concerns. This PR is part of a longer series of revising the entire kernel I/O subsystem for increased performance, which should eliminate all cases of multiple fmemcpy's. The removal of the requirement for buffers to be mapped to the kernel data L1 area (#819) was the first big step. This PR was checked carefully for correctness internally, and on older real hardware. It makes the assumption that the 8237 DMA controller will work with physical addresses above 64K, but fails only when its internal 16-bit address overflows and wraps. With all L2 1KiB buffers aligned on a 1KiB physical boundary, the 8237 address register won't wrap.

> Is there any particular reason why the cursor position is now shifted down one row (C->cy++ etc.) when initializing the console drivers? 

Yes, adding the ability to printk at any time slightly changed exactly when the kputchar function moves from calling `early_putchar` to the linked-in console drivers conout routine. Basically, in console_init, the console drivers all printk their info line (which uses early_printk), after their X,Y position initialization, but before the call to set_console. This causes the Y value to be off-by-one.

> What happens if the cursor is already on the bottom row?

Good catch. This will never happen on a PC boot, nor under QEMU, but could happen on another emulator. I'll have to look into a fix for that.

Thank you!

ghaerr:
Hello @tkchia,

I reordered the calling sequence of console initialization to remove the need for adjusting the cursor position. With early printk, this can now be done without issue for all consoles.

Thank you!

tkchia (approved):

tkchia:
Hello @ghaerr,

> With all L2 1KiB buffers aligned on a 1KiB physical boundary, the 8237 address register won't wrap.

Thanks!  I guess that should indeed work for now, as long as only 2 sectors are read directly into the L2 buffers each time (thanks to `do_bioshd_request()`).  Please proceed to merge.

(Other than that, I will probably also submit a small patch later to improve the handling of the DDPT.)

Thank you!

ghaerr:
Hello @tkchia,

>  I guess that should indeed work for now, as long as only 2 sectors are read directly into the L2 buffers each time (thanks to do_bioshd_request()).

Yes, exactly.

Despite my attempts to remove it, it looks like DMASEG will be here to stay with track caching, which greatly improves floppy throughput. I can't yet see any other way to improve performance.

> (Other than that, I will probably also submit a small patch later to improve the handling of the DDPT.)

Please do. There's been a lot of big changes lately, I also plan on some possible fine tuning of other portions of the BIOS driver.

Thank you!

PR #833, opened 2020-11-03
[kernel] Add experimental display showing disk access

ghaerr:
This PR adds a "spinning wheel" in the upper right corner of the display whenever disk I/O is occurring, kind of like the macOS "beachball" cursor when forced to wait.

I thought this might be useful, or fun, to visually see when the kernel is waiting for I/O to complete, especially now that the track cache is default on.

In general, QEMU I/O is too fast to see, so this is mostly for real hardware.

It might be possible to show TCP/IP activity as well, but need to think a bit more about it.


Mellvik:
wow @ghaerr,

I'm obviously not the only one going to town these days :-) .

Can't wait to test it!

—M

Issue #583, opened 2020-04-25
mv() incomplete


Mellvik:
This may be one someone's list already, it certainly is on mine. But for the record:

mv(1) is incomplete in that it (possibly among other things) cannot move directories. Probably an issue in rename() or link(), and hopefully an easy fix.

ghaerr:
I looked at the Minix filesystem source. In elks/fs/namei.c::sys_rename(), `rename` calls `link` then `unlink`, which works for files. The `link` system call does not allow a directory to be hard-linked, and that's why this isn't working (permitted).

Unfortunately, ELKS doesn't have any filesystem entry points for `rename`, which means that all filesystems' file_operations function arrays would have to be modified, and then a function could be written, separately for each filesystem. That'll take a fair amount of work, and all the filesystems would have to be tested, as the file_operations structs are all over the place, but I'll look into it. This probably won't happen quickly.

toncho11:
Is it possible to just do it another way? Create a folder, copy the files, delete the old one.

Mellvik:
@toncho11 - 
that is indeed possible, but the 'raison d'être'  for `mv` is to change names and avoid copying - which will happen only when moving files across file systems. `cp` should do that what you suggest, but `cp` (like `mv` and many other utilities) is still functionally incomplete in ELKS, so there is work to be done for both.

--Mellvik

ghaerr:
Hello @toncho11,

> Is it possible to just do it another way? Create a folder, copy the files, delete the old one.

Nice idea, thanks. The files could be linked, rather than copied, which would make this operation fairly fast. It is a bit of a kluge to implement this in the kernel though. The "proper" way is to implement a new "rename" filesystem entry point, but that's complicated in its own right. I'll take a look at it, now that we have more code space available in the kernel.

Hello @Mellvik,

> cp should do that what you suggest, but cp (like mv and many other utilities) is still functionally incomplete in ELKS, 

What are the unimplemented features in cp and mv? We would need a recursive version of cp that uses links instead of copying, to do what you suggest. That would better be handled by ln, I suppose. However, handling this in userspace rather than the kernel is probably a good idea, until a proper fix to the kernel can be made.

Thank you!

Mellvik:
> 
> cp should do that what you suggest, but cp (like mv and many other utilities) is still functionally incomplete in ELKS,
> 
[...]

PR #834, opened 2020-11-03
[cmds] Enhance mv to overwrite target and move directories

ghaerr:
Provides workaround for broken kernel rename problem identified in issue #583.

`mv` can now be used to rename a directory. This takes @toncho11's idea from https://github.com/jbruchon/elks/issues/583#issuecomment-720961426 and uses links instead of copying.

Also enhances `mv` to overwrite the target file, if it's not a directory.

This version will fail midway should the source directory contain subdirectories or symlinks (like /bin, but renames /bin without httpget and ftpget). Will probably add some error checking code in a further commit.

### DIFF
--- a/elkscmd/file_utils/mv.c
+++ b/elkscmd/file_utils/mv.c
@@ -20,6 +20,7 @@
 #include <grp.h>
 #include <utime.h>
 #include <errno.h>
+#include <dirent.h>
 
 #define BUF_SIZE 1024 
 
@@ -37,6 +38,60 @@ int isadir(name)
 		return 0;
 
 	return S_ISDIR(statbuf.st_mode);
+}
+
+/*
+ * Build a path name from the specified directory name and file name.
+ * If the directory name is NULL, then the original filename is returned.
+ * The built path is in a static area, and is overwritten for each call.
+ */
+char *buildname(char *dirname, char *filename)
+{
+	char		*cp;
+	static	char	buf[PATHLEN];
+
+	if ((dirname == NULL) || (*dirname == '\0'))
+		return filename;
+
+	cp = strrchr(filename, '/');
+	if (cp)
+		filename = cp + 1;
+
+	strcpy(buf, dirname);
+	strcat(buf, "/");
+	strcat(buf, filename);
+
+	return buf;
+}
+
+/* link all files in srcname directory to destname directory*/
+int linkfiles(char *srcname, char *destname)
+{
+	DIR *dirp;
+	struct dirent *dp;
+	char *newsrc;
+	char newdst[PATHLEN];
+
+	dirp = opendir(srcname);
+	if (!dirp)
+		return 1;
+	while ((dp = readdir(dirp)) != NULL) {
+		if (!strcmp(dp->d_name, ".") || !strcmp(dp->d_name, ".."))
+			continue;
+		strcpy(newdst, buildname(destname, dp->d_name));
+		newsrc = buildname(srcname, dp->d_name);
+		/* link will fail if directory or symlink*/
+		if (link(newsrc, newdst) < 0) {
+			perror(newsrc);
+			return 1;
+		}
+		if (unlink(newsrc) < 0) {
+			perror(newsrc);
+			return 1;
+		}
+	}
+	closedir(dirp);
+	return 0;
 }
 
 /*
@@ -137,30 +192,6 @@ error_exit:
 	return 0;
 }
 
-/*
- * Build a path name from the specified directory name and file name.
- * If the directory name is NULL, then the original filename is returned.
- * The built path is in a static area, and is overwritten for each call.
- */
-char *buildname(char *dirname, char *filename)
-{
-	char		*cp;
-	static	char	buf[PATHLEN];
-
-	if ((dirname == NULL) || (*dirname == '\0'))
-		return filename;
-
-	cp = strrchr(filename, '/');
-	if (cp)
-		filename = cp + 1;
-
-	strcpy(buf, dirname);
-	strcat(buf, "/");
-	strcat(buf, filename);
-
-	return buf;
-}
-
 
 int main(int argc, char **argv)
 {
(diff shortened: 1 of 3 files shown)


ghaerr:
Added code for directory rename to pre-check source directory for subdirectories or symlinks.

@Mellvik, give this a whirl, it should work great, and fix #583.

Mellvik:
Thank you @ghaerr, one of those small things that makes a big difference.

Incidentally, I found a couple of issues while testing:
```
login: root
# mkdir a b c
# mv c a
# touch 1 2 3 4 5 6 7 8 
# ln -s 3 link
# ls -l
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 1
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 2
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 3
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 4
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 5
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 6
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 7
-rw-rw-rw-  1 root     root            0  Nov 04 11:32 8
drwxrwxrwx  3 root     root           48  Nov 04 11:31 a
drwxrwxrwx  2 root     root           32  Nov 04 11:31 b
lrwxrwxrwx  1 root     root            1  Nov 04 11:32 link -> 3
# mv ? b
Can't move directory or symlink: a/c
# ls
a     b     link  
# ls b
1  2  3  4  5  6  7  8  a  
# 
```
`mv ? b` attempts to move the `a` directory and creates an entry for it before realizing that it contains another directory.
I may have misunderstood, but this was supposed to work, right? 
Also, it would make sense to have an error message when trying to move a directory into itself. BSD mv would complain `mv: b: rename: File exists`
BTW, is there a reason for treating symlinks differently?
Finally, there is another case where `mv` doesn't work and I cannot figure out why ...
```
# touch x/c/file
# mv x/c a
x/c/file: Not a directory 
# ls x/c
file
#  
```

ghaerr:
@Mellvik,

The improved mv cannot at this point "rename" a directory that contains subdirectories or symlinks. While this could be made possible, it would take a lot more tricky coding to ensure correctness, since mv is having to link/unlink files rather than just execute a kernel rename, since that isn't implemented.

The reason subdirectories are not supported is that the entire linkfiles routine would have to be made recursive, which I thought too hard to test, since the mkdir is handled outside of it. Symlinks require special handling because merely accessing one causes its link to be followed, so the special system call `readlink` must be used, and handled/copied specially. My first try was just a quick attempt at adding basic directory rename functionality.

> Also, it would make sense to have an error message when trying to move a directory into itself.

I'll look into that.

> \# touch x/c/file
\# mv x/c a
x/c/file: Not a directory 
\# ls x/c
file
\#  

I can't get the first line to repeat - touch doesn't support making directories on ELKS. If I start with an empty directory and type "mkdir x x/c" instead of touch, it works fine.

Mellvik:
> # touch x/c/file
> # mv x/c a
> x/c/file: Not a directory
> # ls x/c
> file
> #
> 
> I can't get the first line to repeat - touch doesn't support making directories on ELKS. If I start with an empty directory and type "mkdir x x/c" instead of touch, it works fine.
> 
Sorry @ghaerr,
I didn't include the entire command history (it was in the example above this one). Please add
# mkdir x c
# mv c x
… ahead of the snippet.


> The improved mv cannot at this point "rename" a directory that contains subdirectories or symlinks. While this could be made possible, it would take a lot more tricky coding to ensure correctness, since mv is having to link/unlink files rather than just execute a kernel rename, since that isn't implemented.
> 

OK; I get that. There was still a bug in there though? (the 'a' directory being duplicated)


—Mellvik

ghaerr:
> There was still a bug in there though? (the 'a' directory being duplicated)

No? It seems to do the same as on Linux or macOS: the x/c directory is renamed to a. That's what is intended, right?

Mellvik:
> 
> 
> There was still a bug in there though? (the 'a' directory being duplicated)
> 
> No? It seems to do the same as on Linux or macOS: the x/c directory is renamed to a. That's what is intended, right?
> 
> —
> 
Not exactly. Look at my first snippet, the long one starting with login: root

Towards the end, look at the two ls commands. They list two different directories, both containing the a directory. So the mv failed midway and left the destination there anyway.

—Mellvik

ghaerr:
> So the mv failed midway and left the destination there anyway.

Got it, thanks. Fixed in #836.

Mellvik:
@ghaerr, 
I'm still getting the same errors - did I miss something? 

```
# mkdir a b c
# mv c a
# touch 1 2 3 4 5 6 7 8
# ln -s 3 link
# mv ? b
Can't move directory or symlink: a/c                   <---- this should work, also it should complain about not being able to move b into itself
# ls b
1  2  3  4  5  6  7  8  a                 <--- since it failed, a should not be here
# ls a

# mkdir a/y
# touch a/y/file
# mkdir d
# mv a/y d
a/y/file: Not a directory     <--- this should work (?), and the error message does not make sense (to me at least). 
# 
```

ghaerr:
@Mellvik,

This is getting quite tricky due to the static buffer in buildnameI(). I've patched a fix in #838, try it out.

There are a number of preexisting problems with mv, and its not perfect. Renaming symlinks is known buggy and won't be fixed at this time. In general, this fix is meant to allow directory renames only because of the kernel rename problem, not arbitrary larger number of arguments to mv, just regular files and directories.

This fixes the last example above, which was definitely buggy. The first example is correct, as when "mv ? b" ends up executing "mv a b", it should fail since a contains the subdirectory c.

Lets use more simple examples with only two arguments to mv for the time being, so as to not introduce other mv bugs or enhancements needed at the same time.

Thanks!

Mellvik:
Thanks @ghaerr,
I realize this gets tricky because of lacking functionality 'under the hood', so opting out some nice to have (or 'what we're used to') functions is perfectly fine.

> The first example is correct, as when "mv ? b" ends up executing "mv a b", it should fail since a contains the subdirectory c.
The problem with the first example is that it still leaves the 'a' directory in to places.


I'll test the fix asap.

—Mellvik

PR #837, opened 2020-11-04
[kernel] improve procedure for modifying floppy parameter table

tkchia:
With this patch, the ELKS kernel now
  * resets the floppy drives whenever the DDPT is changed, as recommended by Ralf Brown's Interrupt List; and
  * only updates the DDPT when really necessary.

`read_sector(...)` no longer updates the DDPT, since it only reads a single sector, and so should work correctly, whatever the DDPT "sectors per track" might say.

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -302,10 +302,39 @@ static void bioshd_release(struct inode *inode, struct file *filp)
     }
 }
 
-/* set our DDPT sectors per track value*/
+static void reset_bioshd(int drive);
+
+/* set our DDPT sectors per track value and reset the floppy drives if
+ * necessary */
 static void set_ddpt(int max_sectors)
 {
+    /* We want to prevent the BIOS from accidentally doing a "multitrack"
+     * floppy read --- and wrapping around from one side to the next ---
+     * when ELKS only wants to read from a single track.
+     *
+     * (E.g. if DDPT SPT = 9, and a disk has 18 sectors per track, and we
+     * want to read sectors 9--10 from track 0, side 0, then the BIOS may
+     * read sector 9 from track 0, side 0, followed by sector 1 from track
+     * 0, side 1, which will be wrong.)
+     *
+     * To prevent this, we set the DDPT SPT field to at least the actual
+     * sector count per track in the detected disk geometry.  The DDPT SPT
+     * should never be smaller than the actual SPT, but it can be larger.
+     *
+     * Also, Ralf Brown's Interrupt List recommends that, if the DDPT is
+     * changed, "INT 13/AH=00h should be called to ensure that the floppy
+     * disk controller is appropriately reprogrammed".  So do a reset if the
+     * DDPT is changed.  (I am not sure how necessary this is in practice,
+     * but I think a reset at this point should not degrade floppy I/O
+     * performance too much.) -- tkchia 20201104
+     */
+    if (DDPT[SPT] < (unsigned char) max_sectors) {
+	int drive;
 	DDPT[SPT] = (unsigned char) max_sectors;
+	for (drive = 0; drive < _fd_count; ++drive)
+	    reset_bioshd(drive);
+	debug_bios("bioshd: DDPT SPT changed to %d\n", max_sectors);
+    }
 }
 
 /* get the diskette drive parameter table from INT 1E and point to our RAM copy of it*/
@@ -323,7 +352,6 @@ static void copy_ddpt(void)
 	fmemcpyw(DDPT, _FP_SEG(DDPT), (void *)(unsigned)oldvec, _FP_SEG(oldvec),
 		sizeof(DDPT)/2);
 	debug_bios("bioshd: DDPT vector %x:%x SPT %d\n", _FP_SEG(oldvec), (unsigned)oldvec, DDPT[SPT]);
-	set_ddpt(DDPT[SPT]);
 	*vec1E = (unsigned long)(void __far *)DDPT;
 }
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

Thank you for this PR, as it brings up some possible issues for (re)consideration. Unfortunately, however, I am not in agreement that the large changes this PR suggests should be implemented. I will cover each item separately.

- BIOS reset

Both `do_bios_readwrite` and `bios_readtrack` are called for all BIOS I/O, floppy *and* hard disk. This code will cause a floppy disk reset when doing HD I/O. IMO, floppy resets should not be performed during HD I/O.

Should a user perform a file copy from one floppy to another with different SPT, this code will cause both floppy drives to be reset during the disk copy operation. In addition, the SPT will be technically incorrect for one of the drives during the copy operation, because the SPT is only changed whenever the new value is less than the previous. (If this really mattered, why not just set the SPT to a large value (e.g. 36) and never change it.) The current code tries to be accurate in maintaining at least a technically proper SPT at all times.

The current BIOS driver *never* performs a reset operation, even at open time, unless a BIOS I/O call fails. Adding multiple disk resets at read/write times, but not even at initialization, is a big change. For instance, the XT v1 BIOS source shows that a floppy reset forces full drive recalibration, and may turn off the motor (untested) if on. This definitely doesn't want to happen during a read/write routine that is meant for speed.

The current code was very carefully crafted after a long discussion in #824. It has been shown to work on a 1983 IBM XT w/v1 BIOS (which previously did not work), a 1986 IBM XT, @Mellvik's PCs, a Compaq 386 and a 386 Desktop. Adding reset code like this definitely requires another full test round, even if it was agreed that adding resets during read/write calls was a good idea.

- if (DDPT[SPT] < (unsigned char) max_sectors)

This code attempts to minimize calls to BIOS reset, but also leaves the DDPT in a technically incorrect state, as the SPT value won't match the disk undergoing I/O, if the I/O SPT are less than the last call to set_ddpt(). Since set_ddpt is shared between hard and floppy disks, the DDPT SPT value is sometimes not matching the current floppy disk, which I don't think is a good idea.

All-in-all, the current code is well tested and shown working across a wide range of real hardware. Given that the current set_ddpt code is very fast, I can't see the benefit in not keeping the SPT value correct at all times.

- Ralf Brown's Interrupt List recommends that, if the DDPT is changed, "INT 13/AH=00h should be called 

Ralf's [description here](http://www.delorie.com/djgpp/doc/rbinter/id/55/24.html) talks about doing a BIOS disk reset when the 1E vector is changed, *not* when changing the SPT value(s) of a custom DDPT. Given that the current code has never called BIOS disk reset, I really don't think it necessary, and definitely don't think it should be done in-between BIOS I/O.

If you feel that a BIOS disk reset should be called, I think this should be done in one place, and only once - in the BIOS open routine. I still have mixed feelings about whether it should ever be called, since it never has (except on disk errors), and things have always worked well. Adding more BIOS calls than necessary complicates the driver, IMO. ELKS does *not* use the BIOS's idea of what the disk drive parameters are; instead, ELKS probes during open and decides for itself what the parameters are. With this line of thinking, I don't think it necessary to "tell" the BIOS about DDPT changes, since ELKS isn't dependent on the strict BIOS disk parameters to be correct. 

Talking about the probing routine for a minute, I have been considering further possible changes that might make ELKS more portable, and quicker, instead of our older-fashioned probing. These include a possible move to recognizing DOS BPB-formatted boot blocks (instead of, or possibly in addition to the ELKS EPB) to get the disk CHS parameters. This would allow for more disks to be immediately recognized without going through a lengthy probe (or disk reset, for that matter). DOS tries recognizing the diskette from BPB instead of probing. IMO, it would be good to make ELKS floppies more compatible with DOS floppies at large, rather than having our own EPB format. (Yes, there are other boot issues to consider, off-topic for this thread).

Getting back to RBIL's disk reset, if we really think we need it, it should be in the probe routine, and the copy_ddpt done at that time, with the SPT still always updated before each BIOS I/O call. As said before, I'm not in favor of it since it introduces a possible large variety of unknown BIOS behaviors, and during our recent heavy testing across a variety of PCs, no system has shown a need for it.

In summary, I think this change is overkill with disk resets, but also introduces more potential problems that have not been ever tested, and keeps the SPT less accurately than currently. The current BIOS driver never performs any multi-track I/O; after the discussions in #824, it was learned that very likely the only purpose of DDPT's SPT is for multi-track. Thus, just settlng SPT to a high value like 36 (as Riley Williams suggested in 2002) and leaving it alone, with no disk resets or changes ever, might be a better idea. Ultimately, I didn't implement it that way from the start because SPT could *possibly* be interpreted when not doing multi-track, which lead to changing only SPT before each I/O.

Thank you!

tkchia:
Hello @tkchia 

> Unfortunately, however, I am not in agreement that the large changes this PR suggests should be implemented.

OK.  I guess I can defer to the existing implementation, which currently works.

Anyway, if it helps: if a diskette track has actual sectors per track smaller than the DDPT SPT (e.g. 9 &lt; 12), you can think of it as a sort of "deficient" track with some of the larger-numbered sectors missing (or bad) from the track.

From what I gather, normal read and write operations do not need the DDPT SPT to be the same as the exact geometry, though it does need to be exactly correct when formatting a track.

Thank you!

Mellvik:
@ghaerr,

Talking about the probing routine for a minute, I have been considering further possible changes that might make ELKS more portable, and quicker, instead of our older-fashioned probing. These include a possible move to recognizing DOS BPB-formatted boot blocks (instead of, or possibly in addition to the ELKS EPB) to get the disk CHS parameters. This would allow for more disks to be immediately recognized without going through a lengthy probe (or disk reset, for that matter). DOS tries recognizing the diskette from BPB instead of probing. IMO, it would be good to make ELKS floppies more compatible with DOS floppies at large, rather than having our own EPB format. (Yes, there are other boot issues to consider, off-topic for this thread).


I missed this one in my first read through - so FWIW, I think this is an interesting idea. The current probing has never failed in my experience, but it is slow.

—Mellvik

PR #838, opened 2020-11-05
[cmds] More bugfixes to mv

ghaerr:
Bugfixes pointed out in https://github.com/jbruchon/elks/pull/834#issuecomment-722316250.

Certain debug printf's left in for now.

### DIFF
--- a/elkscmd/file_utils/mv.c
+++ b/elkscmd/file_utils/mv.c
@@ -106,10 +106,12 @@ int linkfiles(char *srcdir, char *destdir)
 		newsrc = buildname(srcdir, dp->d_name);
 
 		/* link will fail if directory or symlink*/
+//printf("link %s to %s\n", newsrc, newdest);
 		if (link(newsrc, newdest) < 0) {
 			perror(newsrc);
 			return 0;
 		}
+//printf("unlink %s\n", newsrc);
 		if (unlink(newsrc) < 0) {
 			perror(newsrc);
 			return 0;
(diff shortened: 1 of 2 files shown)


Mellvik:
Tested - OK.

-M

Mellvik:
Thanks again @ghaerr,

There are improvements - and new issues, some of which may be just the state of things.

On minix fs:

# pwd
/root
# mkdir a b
# touch 3
# ln -s 3 link
# rm 3					<— 'link' is now a dead symbolic link, but still a link
# 
# mv link y
link: No such file or directory 	<— should just change the name of 'link'
# ls
a     b     link  y     		<— 'y' has been created (empty file)
# mv link b
link: No such file or directory 	<— should be able to move 'link' to directory 'b'
# ls -l
drwxrwxrwx  3 root     root           48  Nov 06 09:59 a
drwxrwxrwx  2 root     root          176  Nov 06 10:00 b
lrwxrwxrwx  1 root     root            1  Nov 06 10:00 link -> 3
-rw-r--r--  1 root     root            0  Nov 06 10:09 y
# touch 3				<- 'link' now has a valid target
# mv link x				<- Now mv apparently works … but
# ls
3  a  b  x  y  				<— it does not rename 'link' to 'x', but makes a copy of the target file instead.
# ls -l
-rw-r--r--  2 root     root            0  Nov 06 10:09 3
drwxrwxrwx  3 root     root           48  Nov 06 09:59 a
drwxrwxrwx  2 root     root          176  Nov 06 10:00 b
-rw-r--r--  2 root     root            0  Nov 06 10:09 x
-rw-r--r--  1 root     root            0  Nov 06 10:09 y
# 

If this is kind of chaotic, let me know… and I'll make a list of points instead.

On FAT fs, mv does not seem to work at all:

# cd /mnt/elks
# ls -l
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:02 big
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:03 big2
-rwxr-xr-x  1 root     root     226366500  Nov 06 10:09 bigg
drwxr-xr-x  3 root     root         8192  Aug 28 13:15 bin
-rwxr-xr-x  1 root     root      1451520  Oct 08 15:47 image
-rwxr-xr-x  1 root     root      6094030  Oct 14 09:11 image.hex
-rwxr-xr-x  1 root     root     37727750  Nov 06 10:02 xx
drwxrwxrwx  3 root     root         8192  Nov 06 11:36 yy
# ls yy

# mv xx yy				<— should be able to move file xx into dir yy
yy/xx: No such file or directory 
# mv xx x				<— should be able to change name of x
x: No such file or directory 
# 

— Mellvik

ghaerr:
@Mellvik,

`mv` has never handled symlinks properly, and I have noted this and others on my (huge) list. I don't think it will get fixed for this version; there are a number of tricky cases to handle.

Regarding FAT, mv uses links and FAT doesn't support links, so mv won't ever work. `cp` should be used in all cases. I would remove mv from the FAT distribution but there's currently no easy way to do that.

At this point, I'd prefer to close #583 now, and open new issues after the new version, when issues are found that affect user operation. As you know, there's really no end of bugs and enhancements in the ELKS command set!

Thanks for your testing on this, it helped a lot.

Issue #796, opened 2020-10-16
[kernel][fat] Fatal FAT fs bug?


Mellvik:
Writing BIG file to FAT fs crashes entire disk:
```
# mount -t fat /dev/bda1 /mnt
# cd /mnt
# dd if=/dev/bda of=10m-file bs=2b count=10000
# echo no problem
# dd if=/dev/bda of=100m-file bs=2b count=100000
# ls
garbage listing
```
The MBR and the partition boot block are both gone. No error messages. The final `ls` seems to mess up the kernel too, a following `ps` does not list all processes.

--M

ghaerr:
This is reading and writing from the same filesystem, so it seems the free space on the FAT filesystem will always be used up.

Thus, it is hard to tell if this is a write issue with large files, or more likely a problem with running out of FAT filesystem space.

Further testing could include showing the FAT display after the mount, to know the filesystem specifics, as well as turning on DEBUG_FAT (and ^P before testing to turn on debug messages) to display what is happening in more detail.

Mellvik:
> This is reading and writing from the same filesystem, so it seems the free space on the FAT filesystem will always be used up.
> 
> Thus, it is hard to tell if this is a write issue with large files, or more likely a problem with running out of FAT filesystem space.
> 
Not really. It's reading from the raw device, not the file system, and it's repeatable on qemu (fs size 512mb). My guess (will check) is that using /dev/zero as the input file will give the same result.
> Further testing could include showing the FAT display after the mount, to know the filesystem specifics, as well as turning on DEBUG_FAT (and ^P before testing to turn on debug messages) to display what is happening in more detail.
> 
Good idea, will include that in the next post.

--M
> —

Mellvik:
Probably unrelated, but while working on this (now with a destroyed FAT file system so a good time to do something else with that disk) I repartitioned and created a /dev/bda2 size 65535. Then copied /bin from floppy to the new file system, and df() reports odd numbers:
```
# mkfs /dev/bda2 65535
63424 inodes
65535 blocks
Firstdatazone=2000 (2000)
Zonesize=1024
Maxsize=67107840

# mount /dev/bda2 /mnt
# mkdir /mnt/bin
# cp /bin/* /mnt/bin
# df 
Filesystem    1k-Blocks     free     used    %  FUsed%
/dev/fd0           1440      377     1063  74%    50%
# df /dev/bda2
Filesystem    1k-Blocks     free     used    %  FUsed%
[...]

PR #839, opened 2020-11-06
[kernel] Expand buffer block numbers to 32 bits to support 500MB FAT filesystems

ghaerr:
Fixes FAT filesystem corruption error detailed in https://github.com/jbruchon/elks/issues/796#issuecomment-721218376.

Implements 32-bit block numbers in ELKS buffer system.

All FAT filesystems larger than 65MB suffered from this problem, causing incorrect data reads, and writing over incorrect portions of the disk when the sector number was greater than 65535.

To keep things small, since MINIX FS never uses block numbers greater than 16 bits, two new routines were introduced, called only by the FAT filesystem driver: `getblk32` and `bread32`, which take (unsigned long) block32_t types instead of their 16-bit counterparts, `getblk` and `bread` using block_t.

The FAT filesystem sector numbers were being silently truncated to 16-bit `block_t` values, which caused all block numbers to be effectively calculated mod 65536. When files were written with sector numbers just over 65535, the boot sector and FAT table was being overwritten. 

@Mellvik, this should work fine, please test. I have not yet fixed the FAT "multiple mount" problem, so test only from a single using a single FAT mount. I'm also not yet sure whether writing files (large or small) and running out of space is buggy or not. I suspect it could be, more testing needed.

### DIFF
--- a/elks/arch/i86/drivers/block/ll_rw_blk.c
+++ b/elks/arch/i86/drivers/block/ll_rw_blk.c
@@ -248,14 +248,13 @@ static void add_request(struct blk_dev_struct *dev,
     set_irq();
 }
 
-static void make_request(unsigned short int major, int rw,
-			 register struct buffer_head *bh)
+static void make_request(unsigned short major, int rw, struct buffer_head *bh)
 {
-    register struct request *req;
+    struct request *req;
     sector_t sector, count;
     int max_req;
 
-    debug_blk("BLK %d %s\n", bh->b_blocknr, rw==READ? "read": "write");
+    debug_blk("BLK(%x) %lu %s\n", bh->b_dev, bh->b_blocknr, rw==READ? "read": "write");
     count = (sector_t) (BLOCK_SIZE >> 9);
     sector = bh->b_blocknr * count;
 
--- a/elks/fs/buffer.c
+++ b/elks/fs/buffer.c
@@ -288,7 +288,7 @@ struct buffer_head *readbuf(register struct buffer_head *bh)
     return bh;
 }
 
-static struct buffer_head *find_buffer(kdev_t dev, block_t block)
+static struct buffer_head *find_buffer(kdev_t dev, block32_t block)
 {
     register struct buffer_head *bh = bh_llru;
 
@@ -302,7 +302,7 @@ struct buffer_head *get_hash_table(kdev_t dev, block_t block)
 {
     register struct buffer_head *bh;
 
-    if ((bh = find_buffer(dev, block)) != NULL) {
+    if ((bh = find_buffer(dev, (block32_t)block)) != NULL) {
 	INR_COUNT(bh);
 	wait_on_buffer(bh);
     }
@@ -320,7 +320,14 @@ struct buffer_head *get_hash_table(kdev_t dev, block_t block)
  * when the filesystem starts to get full of dirty blocks (I hope).
  */
 
+/* 16 bit block numbers for super blocks and MINIX filesystem driver*/
 struct buffer_head *getblk(kdev_t dev, block_t block)
+{
+	return getblk32(dev, (block32_t)block);
+}
+
+/* 32 bit block numbers for FAT filesystem driver*/
+struct buffer_head *getblk32(kdev_t dev, block32_t block)
 {
     register struct buffer_head *bh;
     register struct buffer_head *n_bh;
@@ -367,9 +374,16 @@ struct buffer_head *getblk(kdev_t dev, block_t block)
  * it. It returns NULL if the block was unreadable.
  */
 
+/* 16 bit block numbers for super blocks and MINIX filesystem driver*/
 struct buffer_head *bread(kdev_t dev, block_t block)
 {
     return readbuf(getblk(dev, block));
+}
+
+/* 32 bit block numbers for FAT filesystem driver*/
+struct buffer_head *bread32(kdev_t dev, block32_t block)
+{
+    return readbuf(getblk32(dev, block));
 }
 
 #if 0
--- a/elks/fs/msdos/fat.c
+++ b/elks/fs/msdos/fat.c
@@ -40,7 +40,7 @@ long FATPROC fat_access(register struct super_block *sb,long this,long new_value
 	}
 #endif
 	if (!(bh = msdos_sread(sb->s_dev,
-			(long)(MSDOS_SB(sb)->fat_start+(first >> SECTOR_BITS)),&data))) {
+			(sector_t)(MSDOS_SB(sb)->fat_start+(first >> SECTOR_BITS)),&data))) {
 		printk("FAT: bread fat failed\n");
 		return 0;
 	}
@@ -49,7 +49,7 @@ long FATPROC fat_access(register struct super_block *sb,long this,long new_value
 		data2 = data;
 	} else {	/* FAT entry crosses sector boundary, only possible with fat12*/
 		if (!(bh2 = msdos_sread(sb->s_dev,
-				(long)(MSDOS_SB(sb)->fat_start+(last >> SECTOR_BITS)),&data2))) {
+				(sector_t)(MSDOS_SB(sb)->fat_start+(last >> SECTOR_BITS)),&data2))) {
 			unmap_brelse(bh);
 			printk("FAT: bread fat failed\n");
 			return 0;
@@ -91,11 +91,11 @@ long FATPROC fat_access(register struct super_block *sb,long this,long new_value
 				*p_first = new_value & 0xff;
 				*p_last = (*p_last & 0xf0) | (new_value >> 8);
 			}
-			if (bh != bh2) debug_fat("fat_access block bh2 dirty %d\n", bh2->b_blocknr);
+			if (bh != bh2) debug_fat("fat_access block bh2 dirty %lu\n", bh2->b_blocknr);
 			bh2->b_dirty = 1;
 		}
 #endif
-		debug_fat("fat_access block bh dirty %d\n", bh->b_blocknr);
+		debug_fat("fat_access block bh dirty %lu\n", bh->b_blocknr);
 		bh->b_dirty = 1;
 #if 0000 //FIXME
 		/* update extra FAT table copies*/
@@ -106,21 +106,21 @@ long FATPROC fat_access(register struct super_block *sb,long this,long new_value
 		//FIXME or perhaps just make a complete FAT table copy on unmount!
 		for (copy = 1; copy < MSDOS_SB(sb)->fats; copy++) {
 			if (!(c_bh = msdos_sread(sb->s_dev,
-						(long)(MSDOS_SB(sb)->fat_start+(first >> SECTOR_BITS) +
+						(sector_t)(MSDOS_SB(sb)->fat_start+(first >> SECTOR_BITS) +
 						MSDOS_SB(sb)->fat_length*copy),&c_data))) break;
 			memcpy(c_data,data,SECTOR_SIZE);
-		debug_fat("fat_access block cbh write %d\n", bh->b_blocknr);
+		debug_fat("fat_access block cbh write %lu\n", bh->b_blocknr);
 			c_bh->b_dirty = 1;
 			if (data != data2 || bh != bh2) {
 				if (!(c_bh2 = msdos_sread(sb->s_dev,
-						(long)(MSDOS_SB(sb)->fat_start+(first >> SECTOR_BITS) +
+						(sector_t)(MSDOS_SB(sb)->fat_start+(first >> SECTOR_BITS) +
 						MSDOS_SB(sb)->fat_length*copy + 1),&c_data2))) {
 					unmap_brelse(c_bh);
 					break;
 				}
 				memcpy(c_data2,data2,SECTOR_SIZE);
 				c_bh2->b_dirty = 1;		//FIXME looks like bug without this
-		debug_fat("fat_access block cbh2 write %d\n", c_bh2->b_blocknr);
+		debug_fat("fat_access block cbh2 write %lu\n", c_bh2->b_blocknr);
 				unmap_brelse(c_bh2);
 			}
 			unmap_brelse(c_bh);
@@ -255,7 +255,7 @@ long FATPROC get_cluster(register struct inode *inode,long cluster)
 }
 
 
-long FATPROC msdos_smap(struct inode *inode,long sector)
+sector_t FATPROC msdos_smap(struct inode *inode, sector_t sector)
 {
 	register struct msdos_sb_info *sb;
 	long cluster;
--- a/elks/fs/msdos/fat.c
+++ b/elks/fs/msdos/fat.c
@@ -16,12 +16,12 @@ static struct fat_cache *fat_cache,cache[FAT_CACHE];
 /* Returns the this'th FAT entry, -1 if it is an end-of-file entry.
    If new_value is != -1, that FAT entry is replaced by it. */
 
-long FATPROC fat_access(register struct super_block *sb,long this,long new_value)
+cluster_t FATPROC fat_access(struct super_block *sb,cluster_t this,cluster_t new_value)
 {
 	struct buffer_head *bh,*bh2;
 	unsigned char *p_first,*p_last;
 	void *data,*data2;
-	long first,last,next;
+	cluster_t first,last,next;
 	int fatsz = MSDOS_SB(sb)->fat_bits;
 	//long copy;
 	//void *c_data, *c_data2;
@@ -148,7 +148,8 @@ void FATPROC cache_init(void)
 }
 
 
-void FATPROC cache_lookup(struct inode *inode,long cluster,long *f_clu,long *d_clu)
+void FATPROC cache_lookup(struct inode *inode,cluster_t cluster,
+	cluster_t *f_clu, cluster_t *d_clu)
 {
 	register struct fat_cache *walk;
 
@@ -180,7 +181,7 @@ static void FATPROC list_cache(void)
 #endif
 
 
-void FATPROC cache_add(struct inode *inode,long f_clu,long d_clu)
+void FATPROC cache_add(struct inode *inode, cluster_t f_clu, cluster_t d_clu)
 {
 	register struct fat_cache *walk,*last;
 
@@ -230,7 +231,7 @@ void FATPROC cache_inval_inode(struct inode *inode)
 }
 
 
-void FATPROC cache_inval_dev(int device)
+void FATPROC cache_inval_dev(kdev_t device)
 {
 	register struct fat_cache *walk;
 
@@ -239,9 +240,9 @@ void FATPROC cache_inval_dev(int device)
 }
 
 
-long FATPROC get_cluster(register struct inode *inode,long cluster)
+cluster_t FATPROC get_cluster(register struct inode *inode, cluster_t cluster)
 {
-	long this,count;
+	cluster_t this, count;
 
 	if (!(this = inode->u.msdos_i.i_start)) return 0;
 	if (!cluster) return this;
@@ -258,7 +259,7 @@ long FATPROC get_cluster(register struct inode *inode,long cluster)
 sector_t FATPROC msdos_smap(struct inode *inode, sector_t sector)
 {
 	register struct msdos_sb_info *sb;
-	long cluster;
+	cluster_t cluster;
 	int offset;
 
 	sb = MSDOS_SB(inode->i_sb);
@@ -280,9 +281,9 @@ sector_t FATPROC msdos_smap(struct inode *inode, sector_t sector)
 /* Free all clusters after the skip'th cluster. Doesn't use the cache,
    because this way we get an additional sanity check. */
 
-int FATPROC fat_free(register struct inode *inode,long skip)
+int FATPROC fat_free(register struct inode *inode, cluster_t skip)
 {
-	long this,last;
+	cluster_t this, last;
 
 	debug_fat("fat_free\n");
 	if (!(this = inode->u.msdos_i.i_start)) return 0;
--- a/elks/fs/msdos/file.c
+++ b/elks/fs/msdos/file.c
@@ -66,7 +66,7 @@ static size_t msdos_file_read(register struct inode *inode,register struct file 
 {
 	char *start;
 	size_t left,offset,size;
-	long sector;
+	sector_t sector;
 	struct buffer_head *bh;
 	void *data;
 
@@ -99,7 +99,7 @@ static size_t msdos_file_read(register struct inode *inode,register struct file 
 static size_t msdos_file_write(register struct inode *inode,register struct file *filp,char *buf,
     size_t count)
 {
-	long sector;
+	sector_t sector;
 	int offset,size,written;
 	int error;
 	char *start;
@@ -125,7 +125,7 @@ static size_t msdos_file_write(register struct inode *inode,register struct file
 	for (start = buf; count; count -= size) {
 		while (!(sector = msdos_smap(inode,filp->f_pos >> SECTOR_BITS)))
 			if ((error = msdos_add_cluster(inode)) < 0) break;
-		if (error) break;
+		if (error) break;	//FIXME check error return
 		offset = (int)filp->f_pos & (SECTOR_SIZE-1);
 		size = MIN(SECTOR_SIZE-offset,count);
 		if (!(bh = msdos_sread(inode->i_dev,sector,&data))) {
@@ -140,7 +140,7 @@ static size_t msdos_file_write(register struct inode *inode,register struct file
 			inode->i_size = filp->f_pos;
 			inode->i_dirt = 1;
 		}
-		debug_fat("file block write %d\n", bh->b_blocknr);
+		debug_fat("file block write %lu\n", bh->b_blocknr);
 		bh->b_dirty = 1;
 		unmap_brelse(bh);
 	}
--- a/elks/fs/msdos/file.c
+++ b/elks/fs/msdos/file.c
@@ -153,10 +153,10 @@ static size_t msdos_file_write(register struct inode *inode,register struct file
 
 void msdos_truncate(register struct inode *inode)
 {
-	int cluster;
+	cluster_t cluster;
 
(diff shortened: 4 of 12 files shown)


Mellvik:
Thank you @ghaerr,

this is really good news. I've been beating the fat filesystem backwards and forwards this morning, with files up to 226MB in size, and not seen any problems at all. Exceeding the available space is handled correctly, subsequent deletions are fine and the FS amounts and remounts just fine.

Per your suggestion I didn't test multiple mounts.

There are mv problems on FAT, that's a different issue.
Also, I discovered that if an incoming telnet session has been active on the FAT filesystem (changed files) and has a FAT FD current directory when exiting, there is no way to umount the filesystem after the telnet exit. It may make sense to create new issue on this one.

Great work!

—Mellvik

ghaerr:
> Per your suggestion I didn't test multiple mounts.

I'll submit a PR for fixing the multi-mount FAT issue and if no issues are found, #796 can be closed, or are there other issues there?

> I discovered that if an incoming telnet session has been active on the FAT filesystem (changed files) and has a FAT FD current directory when exiting, there is no way to umount the filesystem after the telnet exit. 

Is this unique to FAT? I'm wondering whether the system doesn't clear the current directory on process exit, or whether this is a FAT-only issue. Can you test this on a MINIX FD also?

Glad this works, it was fun jumping back into the FAT filesystem code. Its in better shape that I had originally thought, and several other potential problems with disks with large cluster sizes were found and corrected.

Thanks!

Mellvik:
> 
> I discovered that if an incoming telnet session has been active on the FAT filesystem (changed files) and has a FAT FD current directory when exiting, there is no way to umount the filesystem after the telnet exit.
> 
> Is this unique to FAT? I'm wondering whether the system doesn't clear the current directory on process exit, or whether this is a FAT-only issue. Can you test this on a MINIX FD also?
> 
@ghaerr, I just tested this, and it works fine with minix. Also, it seems if I do a sync before terminating the telnet connection, the umount is fine. That may help diagnose the issue. (with linux, no such sync is required).
BTW - doing a sync from another login session does not help: When umount is stuck, it stays that way till reboot.

BTW2 - the FAT multiple mount issue which apparently is the reason for the multiple FAT DIR listings in the debug output: This may be  the reason for FAT directory listings not only being slow, but slowing down while listing. Something we discussed (an filmed) more than a year ago. Looking forward to testing this one whenever it's there.
> Glad this works, it was fun jumping back into the FAT filesystem code. Its in better shape that I had originally thought, and several other potential problems with disks with large cluster sizes were found and corrected.
> 
> 
Yes, the beating it got today shows surprising (encouraging) robustness.

Thank you.

—Mellvik


PS: I just noticed, there is another FAT related PR - I'm on it.

Mellvik:
One more thing - I just noticed, the 'cannot unmount' issue has nothing to do with telnet,
it happens when you cd away from the fat volume without a sync.

I need to look more at this over the weekend.

—Mellvik


> 6. nov. 2020 kl. 16:47 skrev Helge Skrivervik <helge@mymayday.com>:
> 
>> 
>> I discovered that if an incoming telnet session has been active on the FAT filesystem (changed files) and has a FAT FD current directory when exiting, there is no way to umount the filesystem after the telnet exit.
>> 
>> Is this unique to FAT? I'm wondering whether the system doesn't clear the current directory on process exit, or whether this is a FAT-only issue. Can you test this on a MINIX FD also?
>> 
> @ghaerr, I just tested this, and it works fine with minix. Also, it seems if I do a sync before terminating the telnet connection, the umount is fine. That may help diagnose the issue. (with linux, no such sync is required).
> BTW - doing a sync from another login session does not help: When umount is stuck, it stays that way till reboot.
> 
> BTW2 - the FAT multiple mount issue which apparently is the reason for the multiple FAT DIR listings in the debug output: This may be  the reason for FAT directory listings not only being slow, but slowing down while listing. Something we discussed (an filmed) more than a year ago. Looking forward to testing this one whenever it's there.
>> Glad this works, it was fun jumping back into the FAT filesystem code. Its in better shape that I had originally thought, and several other potential problems with disks with large cluster sizes were found and corrected.
>> 
>> 
> Yes, the beating it got today shows surprising (encouraging) robustness.
> 
> Thank you.
> 
> —Mellvik
> 
> 
> PS: I just noticed, there is another FAT related PR - I'm on it.

ghaerr:
> I just noticed, the 'cannot unmount' issue has nothing to do with telnet,
it happens when you cd away from the fat volume without a sync.

Perfect, I was hoping for a simpler test case. I'll add this to the list.

> the FAT multiple mount issue which apparently is the reason for the multiple FAT DIR listings in the debug output

No, the multiple mount issue is entirely seperate from this multiple FAT DIR listings in the debug output.


> This may be  the reason for FAT directory listings not only being slow, but slowing down while listing.

There is nothing that can be done about long FAT directory listings - the entire directory always has to be searched on FAT, because of the way VFAT (long filenames) is implemented. So it can be safely ignored.

ghaerr:
>  I just noticed, the 'cannot unmount' issue has nothing to do with telnet,
it happens when you cd away from the fat volume without a sync.

I can't get this to repeat:
```
mount -t msdos /dev/hda1 /mnt
cd /mnt
cd /
mount /dev/hda1
```
Works fine.

Mellvik:
@ghaerr,
you have to create some activity on the mounted FAT drive, it must be dirty. Create a couple of files & directories.

Like I said, if you sync before cd'ing away from the volume, it's ok.


—M

ghaerr:
> you have to create some activity on the mounted FAT drive, it must be dirty. Create a couple of files & directories.
Like I said, if you sync before cd'ing away from the volume, it's ok.

Dang, can't get it to repeat, even after creating files:
```
mount -t msdos /dev/hda1 /mnt
cd /mnt
cp /bootopts 1
mkdir 2
cd /
umount /dev/hda1
```
Works fine.

Mellvik:
@ghaerr,
it turns out the umount problem is a secondary effect of the double mount issue. I should have seen that right away, sorry about that.

As you suggested, mounting 2 FAT file systems has issues.  As evidenced by the following (after copying some files from FAT volume /dev/hda1 [/mnt/a]  to /dev/hdb1 [/mnt/b]):
```
# cd tmp
# ls
bigg  x     
# ls -l
-rwxr-xr-x  1 root     root     808460346  Nov 06 16:09 bigg
-r--r--r--  1 root     root     540358176  Nov 07 12:58 x
# pwd
/mnt/b/tmp
# cd /
# umount /mnt/a
# umount /mnt/b
/mnt/b: Device or resource busy 
#
```
The sizes of the files are totally off the wall, and the file x doesn't exist per DOS dir:
Skjermbilde 2020-11-07 kl  14 28 54

Intrestingly, rebooting and mounting **only** the 'bad' FAT file system, still shows the erroneous `bigg` file.

The following sequence of tests may contribute to narrow down the problem:
- Viped the 2nd FAT volume clean (with DOS format)
- Restarted elks, mounted both volumes
- `cd` to the 2nd volume, created files, copied files, deleted and checked regularly for umount failure: No deal.
- Copied large files from the 1st volume (75MB, 226MB) to the 2nd, no problem.
- Duplicated the 226MB file locally on the 2nd vol in order for the available volume space to be exceeded.
- And there we have the error situation, which not only cause `umount` to fail, but `touch x` apparently creates a humongous file:
```
# cd /mnt/b/y
# ls -l 
-rwxr-xr-x  1 root     root     226366500  Nov 07 14:11 bigg
-rwxr-xr-x  1 root     root     75455500  Nov 07 14:16 xx
# cp bigg bigger
bigger: No space left on device 
Failed to copy bigg -> bigger
# ls -l
-rwxr-xr-x  1 root     root     226366500  Nov 07 14:11 bigg
-rwxr-xr-x  1 root     root     133038080  Nov 07 14:17 bigger
-rwxr-xr-x  1 root     root     75455500  Nov 07 14:16 xx
# rm bigger
# ls
bigg  xx    
# cd /
# umount /mnt/b
/mnt/b: Device or resource busy 
# cd /mnt/b/y
# ls -l
-rwxr-xr-x  1 root     root     226366500  Nov 07 14:11 bigg
-rwxr-xr-x  1 root     root     75455500  Nov 07 14:16 xx
# sync
# cd /
# umount /mnt/b
/mnt/b: Device or resource busy 
# cd /mnt/b
# touch x
# mkdir xx
# ls -l
-rwxr-xr-x  1 root     root       297184  Nov 07 14:07 b
-rwxr-xr-x  1 root     root            0  Nov 07 14:20 x
drwxrwxrwx  3 root     root         8192  Nov 07 14:20 xx
drwxr-xr-x  3 root     root         8192  Nov 07 14:00 y
# cd xx
# touch yy
# ls -l 
-rwxr-xr-x  1 root     root     908081712  Nov 07 14:22 yy
# 
 
```
Just to make sure, I rebooted, mounted only the first FAT volume, copied the same big file locally in order to run out of space, cd'd to / and unmounted the fat volume. No problem:

```
login: root
# mount -t fat /dev/hda1 /mnt/a
# cd /mnt/a/elks
# ls -l
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:02 big
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:03 big2
-rwxr-xr-x  1 root     root     226366500  Nov 06 10:09 bigg
drwxr-xr-x  3 root     root         8192  Aug 28 13:15 bin
-rwxr-xr-x  1 root     root      1451520  Oct 08 15:47 image
-rwxr-xr-x  1 root     root      6094030  Oct 14 09:11 image.hex
drwxr-xr-x  3 root     root         8192  Nov 06 11:36 yy
# ls yy

# rmdir yy         <---- Just creating some file system activity
# ls -l 
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:02 big
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:03 big2
-rwxr-xr-x  1 root     root     226366500  Nov 06 10:09 bigg
drwxr-xr-x  3 root     root         8192  Aug 28 13:15 bin
-rwxr-xr-x  1 root     root      1451520  Oct 08 15:47 image
-rwxr-xr-x  1 root     root      6094030  Oct 14 09:11 image.hex
# cp bigg bigger
bigger: No space left on device 
Failed to copy bigg -> bigger
# ls -l 
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:02 big
-rwxr-xr-x  1 root     root     75455500  Nov 06 10:03 big2
-rwxr-xr-x  1 root     root     226366500  Nov 06 10:09 bigg
-rwxr-xr-x  1 root     root     98017280  Nov 07 14:34 bigger
drwxr-xr-x  3 root     root         8192  Aug 28 13:15 bin
-rwxr-xr-x  1 root     root      1451520  Oct 08 15:47 image
-rwxr-xr-x  1 root     root      6094030  Oct 14 09:11 image.hex
# rm bigger
# cd /
# umount /mnt/a
# 
```

So it seems that using two mounted FAT volumes concurrently works fine except for a bug that appears when the 2nd volume runs out of space.

That said, for both FAT filesystems, DOS `scandisk` will report an incorrect file allocation table after the disk space was exceeded. 
Skjermbilde 2020-11-07 kl  15 43 53

Also - FWIW, seemingly without any errors reported on ELKS, DOS `scandisk` will regularly report the superblock backup to be bad on the volume.
I can investigate both these problems more thoroughly if that's useful. I'm using qemu for most of these tests.

-M

ghaerr:
Hi @Mellvik,

It looks complicated, and there's definitely some bugs there.

I have a couple quick questions so that I'm clear on what you're finding:

Are you testing with the latest "Fix hack in FAT..." (#840) included? 

Are all known FAT problems related to having two FAT filesystems mounted? That is, there are no bugs in single-mounted FAT?

Are there any issues at all with running out of space with single-mounted FAT filesystems, and scandisk?

> So it seems that using two mounted FAT volumes concurrently works fine except for a bug that appears when the 2nd volume runs out of space.

That doesn't answer the problem of "cd /" away from a FAT filesystem when two are mounted. Is that an additional bug?

> That said, for both FAT filesystems, DOS scandisk will report an incorrect file allocation table after the disk space was exceeded.

This is a single-mount FAT bug also? It would be nice to duplicate an out-of-space problem on single mount FAT if possible.

>  DOS scandisk will regularly report the superblock backup to be bad on the volume.

Yes, the ELKS code does not keep up a second FAT table, it was removed for speed reasons when I initially got the FAT filesystem working early this year. For the time being, lets not consider this a bug. I am guessing that scandisk might be the only program that reads that table.

Nice testing. I would like to separate any single-mount FAT bugs separately first, before getting into multi-mount issues. I don't yet have any idea (other than the hack FAT fix), of why multi-mount issues exist.

Thank you!

Mellvik:
> Hi @Mellvik <https://github.com/Mellvik>,
> 
> It looks complicated, and there's definitely some bugs there.
> 
It may be me complicating it while attempting to cover all the possible angles :-)
> I have a couple quick questions so that I'm clear on what you're finding:
> 
> Are you testing with the latest "Fix hack in FAT..." (#840 <https://github.com/jbruchon/elks/pull/840>) included?
> 
> 
Yes.
> Are all known FAT problems related to having two FAT filesystems mounted? That is, there are no bugs in single-mounted FAT?
> 
> 
That's correct. 
> Are there any issues at all with running out of space with single-mounted FAT filesystems, and scandisk?
> 
This is a new finding. I haven't done any testing before using scandisk for verification, so the answer is 'i don't know yet'. I need to do some more testing on that one.
> So it seems that using two mounted FAT volumes concurrently works fine except for a bug that appears when the 2nd volume runs out of space.
> 
> That doesn't answer the problem of "cd /" away from a FAT filesystem when two are mounted. Is that an additional bug?
> 
It does. Look at the logs I submitted, you'll see that the cd /; umount happens after the out of space condition. Not apparently from the log is that it never happens before.
> That said, for both FAT filesystems, DOS scandisk will report an incorrect file allocation table after the disk space was exceeded.
> 
> This is a single-mount FAT bug also? It would be nice to duplicate an out-of-space problem on single mount FAT if possible.
> 
The (in)famous yes and no. There are no apparent problems on elks when beating up a single mounted fat volume real hard, including out of space. The 'incorrect file allocation table' is a new finding, so like I said above, I have to investigate that one further.
> DOS scandisk will regularly report the superblock backup to be bad on the volume.
> 
> Yes, the ELKS code does not keep up a second FAT table, it was removed for speed reasons when I initially got the FAT filesystem working early this year. For the time being, lets not consider this a bug. I am guessing that scandisk might be the only program that reads that table.
> 
> 
Agreed, added mostly for the record.


Additional testing may have to wail till tomorrow.

—M

Mellvik:
The 'incorrect file allocation table' scandisk error may be less server than it appeared:
Skjermbilde 2020-11-07 kl  16 46 57

How du you read this @ghaerr?

M

ghaerr:
> How du you read this?

I think we're fine running without the 2nd copy of the FAT table. The good news is that scandisk can update it, if someone cares.

Mellvik:
So the conclusion is that as far as we're concerned at this point, one FAT volume mounted is rock solid, two mounted will fail if the 2nd runs out of space. (which does not affect the first).

—M

ghaerr:
> That said, for both FAT filesystems, DOS scandisk will report an incorrect file allocation table after the disk space was exceeded.

Also: IMO, scandisk will always report the BACKUP copy of the FAT table is incorrect, since ELKS doesn't keep it up.

This is not the same as a FAT table problem because of exceeding disk space. So that needs to be verified separately. There may not be any problems with the FAT table, or running out of space (just the bugs you mention when creating files with dual mounted FAT filesystems).

Mellvik:
@ghaerr,
I guess I just managed to confuse the issue bringing in scandisk to the equation. So let's forget that.

The issue I believe is documented in the debug logs above is that the 2nd FAT volume works fine until ut runs out of space. After that (and space made available immediately after), the file system is screwed. To the extent that umount is not possible. And remounting after reboot still shows a screwed FS.

So something (bad) happens when the end of medium is reached that does not happen on the first volume.
Scandisk does not see or fix this problem.

—Mellvik

ghaerr:
> I guess I just managed to confuse the issue bringing in scandisk to the equation. So let's forget that.

Agreed. However it would be nice to confirm using scandisk any corruption results other than "backup copy of FAT incorrect".

> The issue I believe is documented in the debug logs above is that the 2nd FAT volume works fine until ut runs out of space. After that (and space made available immediately after), the file system is screwed. To the extent that umount is not possible. And remounting after reboot still shows a screwed FS.

Got it.

> So something (bad) happens when the end of medium is reached that does not happen on the first volume.
> Scandisk does not see or fix this problem.

So the bad event is triggered by running out of space, and the filesystem contents are not what is expected, however the filesystem is reported completely OK by scandisk (other than bad duplicate FAT)? And this only happens when two FAT filesystems are mounted? Strange.

Looking hard at the FAT source code, there are two cases that are handled a bit specially, although these should not be related to multiple-mount FAT. These two issues are mkdir and file create. It would be nice to see whether you can get this problem to reproduce WITHOUT using mkdir/rmdir, or creating a new file. The special code handles some locking when a new directory entry needs to be created.  You can previously create a new file (directory entry), just don't rename or delete it. You could copy a file on top of it, or edit it with vi, instead. If it only happens with large files, then just create a file with the same name prior to the file copy.

The other item that is strange is that scandisk doesn't report any file access table corruption when the disk has run out of space. So the issue may not be strictly related to that.

Thanks!

ghaerr:
One other thing @Mellvik: we haven't tested much with multiple filesystems mounted. It is possible this issue isn't in FAT, but in the VFS, meaning the same problem could occur on dual-mounted MINIX filesystems. The ELKS VFS layer handles all the upper level file I/O, including current directory, etc., then calls the MINIX, FAT or ROMFS filesystem to change the filesystem data, all on top of the buffered block I/O system.

Mellvik:
> 
> So the bad event is triggered by running out of space, and the filesystem contents are not what is expected, however the filesystem is reported completely OK by scandisk (other than bad duplicate FAT)? And this only happens when two FAT filesystems are mounted? Strange.
> 
No, not exactly. The space overrun condition does not (afaik) cause any change or inconsistency, but something gets overwritten somewhere that cause subsequent operations to be out of whack. Such as umounts not being possible and a simple touch command apparently creating a huuuge file.
> Looking hard at the FAT source code, there are two cases that are handled a bit specially, although these should not be related to multiple-mount FAT. These two issues are mkdir and file create. It would be nice to see whether you can get this problem to reproduce WITHOUT using mkdir/rmdir, or creating a new file. The special code handles some locking when a new directory entry needs to be created. You can previously create a new file (directory entry), just don't rename or delete it. You could copy a file on top of it, or edit it with vi, instead. If it only happens with large files, then just create a file with the same name prior to the file copy.
> 
It seems to me the trigger here is the space overrun, not whether the file causing it is big or small (its inconvenient to create the condition using small files).
> The other item that is strange is that scandisk doesn't report any file access table corruption when the disk has run out of space. So the issue may not be strictly related to that.
> 
Agreed, this is indeed strange. 
I'll do some more testing tomorrow to ascertain whether this is indeed related to fat or is more general.

-Mellvik
>

Mellvik:
@ghaerr,

The new test setup to beat up mounted minix filesystems revealed a number of interesting things. Here's a summary:

- I immediately hit the hard limit of 4 mounts. Mounting the 5th FS (including root), leaves the system dead - no messages, just hang.
	I suggest we increase that to 6 or 8.

- (enhancement request) we really need a mount command that when given no arguments, lists mounted filesystems.

- fsck may have a heap problem (e g sys_brk(32) fail: brk f864 over by 2274 bytes) - reported as 
	fsck: Unable to allocate buffer for zone count

- I don't have a mountable minix hs image at hand (actually I didn't look, because that wasn't the target of the test), so I don't know whether any of these findings also applies to the fist minix fs mount.

- There seem to be severe problems with mounted minix file systems - and possibly  the df command:
-------------------------------------------------------------
# mkfs /dev/hdb3 65535
21845 inodes
65535 blocks
Firstdatazone=696 (696)
Zonesize=1024
Maxsize=67107840

# mount /dev/hdb3 /mnt/m2
# du /mnt/m2
1	/mnt/m2
# df /mnt/m2
/dev/hdb3         65535    44266    21269  33%     1%	<—— this is an empty - freshly created - file system
# umount /mnt/m2
# fsck -s /dev/hdb3				<—— this instance of fsck didn't fail
fsck 1.0.1 - 09/01/20
21845 inodes
-1 blocks					<—— I guess this is just the signed version of 65535
Pre-zone blocks 2+3+8+683 696
Firstdatazone=696 (696)
Zonesize=1024
Maxsize=67107840 bytes
Filesystem state=1
namelen=14

# 
———————————————————————————————------------------------------------

- At this point things look OK, except for 'df' reporting 33% used capacity.

- Recreating the same file system on the same device and tuning df again, shows different numbers (44% used):

------------------------------------------------------------------------
# df .
/dev/hdb3         65535    37103    28432  44%     1%
# 
-----------------------------------------------------------------------

- Creating smaller files - up to a few megs - on the newly created file system works fine.

- Creating a larger file (say 20M) cause 'no space left on device' error and the message 'mnb: still zero bit!8192' on the console.
The command sequence looks like this (this is now the only mounted file system in addition to root):

-----------------------------------------------------------------------
# cd /mnt/m2
# df .
/dev/hdb3         65535    37103    28432  44%     1%
# dd if=/dev/zero bs=1k count=20000 of=20Mfile
20Mfile: No space left on device 
8174+0 records in
8173+0 records out
8369152 bytes (8173 KiB) copied
# df .
/dev/hdb3         65535    28697    36838  57%     1%
# ls -l
-rw-r--r--  1 root     root      8369152  Nov 08 11:51 20Mfile
#
-----------------------------------------------------------------------

So, an 8M file has been created, there is still apparently space, and an odd error.

I've run through a number of related tests before and after this, and the errors are consistent. Adding another mounted minix fs does now change anything. The third file system shows the same behaviour - unsurprisingly.

- There is never a problem unmounting the minix file systems.


As confusing as it seems (at least to me it does), it may seem your hunch about something wrong above fs level indeed has merit.


—Mellvik


> 7. nov. 2020 kl. 21:19 skrev Helge Skrivervik <helge@mymayday.com>:
> 
>> 
>> So the bad event is triggered by running out of space, and the filesystem contents are not what is expected, however the filesystem is reported completely OK by scandisk (other than bad duplicate FAT)? And this only happens when two FAT filesystems are mounted? Strange.
>> 
> No, not exactly. The space overrun condition does not (afaik) cause any change or inconsistency, but something gets overwritten somewhere that cause subsequent operations to be out of whack. Such as umounts not being possible and a simple touch command apparently creating a huuuge file.
>> Looking hard at the FAT source code, there are two cases that are handled a bit specially, although these should not be related to multiple-mount FAT. These two issues are mkdir and file create. It would be nice to see whether you can get this problem to reproduce WITHOUT using mkdir/rmdir, or creating a new file. The special code handles some locking when a new directory entry needs to be created. You can previously create a new file (directory entry), just don't rename or delete it. You could copy a file on top of it, or edit it with vi, instead. If it only happens with large files, then just create a file with the same name prior to the file copy.
>> 
> It seems to me the trigger here is the space overrun, not whether the file causing it is big or small (its inconvenient to create the condition using small files).
>> The other item that is strange is that scandisk doesn't report any file access table corruption when the disk has run out of space. So the issue may not be strictly related to that.
>> 
> Agreed, this is indeed strange. 
> I'll do some more testing tomorrow to ascertain whether this is indeed related to fat or is more general.
> 
> -Mellvik
>>

ghaerr:
Hello @Mellvik,

Yes, I am guessing there are loads of issues right around the corner on with heavier usage of creating filesystems from within ELKS, as well as multi-mounted filesystems. I'll comment on your findings, but at this point want to stay on-thread, which is whether or not FAT has any bugs when single-mounted.

> I immediately hit the hard limit of 4 mounts. Mounting the 5th FS (including root), leaves the system dead - no messages, just hang.	I suggest we increase that to 6 or 8.

I wasn't even aware there is a limit. I'll make a note of it.

> (enhancement request) we really need a mount command that when given no arguments, lists mounted filesystems.

Agreed, next version.


> fsck may have a heap problem (e g sys_brk(32) fail: brk f864 over by 2274 bytes) - reported as
	fsck: Unable to allocate buffer for zone count

I believe fsck has a number of problems with filesystems larger than 1.4M. There just isn't enough space to check without running out of heap space in 64k. This also will have to wait until the next version.

> mkfs /dev/hdb3 65535
> There seem to be severe problems with mounted minix file systems - and possibly  the df command:

I'm not yet sure mkfs works with a max-size filesystem, it will need further testing on single-mounted fs first.
Noted about the df command, will need further testing.

> fsck -s /dev/hdb3				<—— this instance of fsck didn't fail

fsck will have to be tested using a known good filesystem maker (using mfs), before relying on internal mkfs.
All the of internal filesytem utilities will have to be painstakingly tested on single-mounted filesystems first, using outside utilities to understand the limitations. I know there are problems with fsck with large filesystems, for instance.

I suggest all of this gets tabled to the next version, as there is literally no end of problems with ELKS commands.
It seems that single-mounted FAT filesystem work very well now, where it didn't before, thus this issue should be closed, and another opened, after some single-mount testing on MINIX, after the 0.4 release.

Thanks!

Mellvik:
Agreed, @ghaerr.
The issue of the thread is fixed. FAT single mount seems rock solid - even with large files.

Before going back to tcp data reliability testing,
I'm going to give minix fs hd single mount a run for the money and open a new issue if I find anything.

-Mellvik

ghaerr:
> I'm going to give minix fs hd single mount a run for the money and open a new issue if I find anything.

Sounds good. Hopefully everything is fine, and we'll push all the multi-mount issues to the next version.

> Before going back to tcp data reliability testing...

Since I've been completely swamped, perhaps you'd like to add the TCP ack/seqno displays to your liking, discussed in https://github.com/jbruchon/elks/issues/841#issuecomment-723448403. Then, since the ^P stuff won't be changing, #841 could be closed as well.

Mellvik:
> Since I've been completely swamped, perhaps you'd like to add the TCP ack/seqno displays to your liking, discussed in #841 (comment). Then, since the ^P stuff won't be changing, #841 could be closed as well.
> 

I was going to suggest that, will do.
Adding a suggestion to the ^P issue before closing.

Thank you.

--M
>

Issue #835, opened 2020-11-03
Sys tool


toncho11:
Hi,

I am suggesting creating a tool similar to DOS sys tool. DOS won also because it was simple. You do sys c: or a: and you are done. 

This new tool called "sys" should:
- copy the bootloader code from the current media (or the one given as parameter)
- create folders and copy the kernel and the most important folders (or everything from the source media) to the destination media

ELKS is bootable on fat32. So one can create easily a FAT 32 partition on a hdd using windows or fdisk. Next from a bootable fat32 ELKS you do:
`sys /dev/fd0 /dev/hda1`
and you are done. You can boot from your hdd and add more programs or compile your own and add them easily.

Or maybe not a program but a script? First dd the bootloader, then mount and copy?

ghaerr:
Hello @toncho11,

This is a pretty good idea, which really makes ELKS easier to use. Implementing it will be a bit complicated since ELKS supports two different types of boot sectors, MINIX and FAT, as well as other issues:

Supporting only either MINIX or FAT may not be a big deal in a first version, which would only "sys" a new volume with the same type of boot sector the system was booted from. That means if you booted from FAT, you could only "sys" another volume with FAT boot. Same with MINIX. The suggested first  parameter isn't needed, sys would use the boot drive to get the system.

A second issue is that ELKS doesn't currently support changing floppy media. This means that, in order to "sys" when booted from floppy on /dev/fd0, one would only be able to sys a hard disk partition or /dev/fd1, *not* /dev/fd0. So sys would not work well to create floppy disks on single-floppy systems. One could, however, sys a hard disk, boot, then sys a floppy, if one had a single floppy, single hard disk system.

The third issue is a bit more technical: ELKS needs to match the media/drive CHS (number of cylinders/heads/sectors), and the CHS is stamped into the boot sector. For dual-floppy systems, sys would assume the same CHS when creating a system on a second drive, which may not be what is wanted. When sys'ing from a floppy boot to a hard drive, the CHS can be read from the HD partition table, so that would work. However, sys'ing from a hard disk boot to a floppy, sys might not have a way of knowing the floppy CHS, so that might not work.

What are your (and other's) thoughts on these possible limitations? There are possible ways around some of them, we can discuss further if interested.

Thank you!

toncho11:
I think sys'ing fd0->hdd and hdd->fd0 is enough for now. As first version even fd0->hdd is enough.
[...]

PR #846, opened 2020-11-08
[cmds] Create sys tool for making boot disks from ELKS

ghaerr:
Creates `sys` script and helper tool `makeboot` requested in #835 to create ELKS disk images from ELKS.

Usage: sys /dev/{target drive}

Tested only on QEMU for /dev/fd1 and /dev/hda1.
Temporarily only creates bootable disk with /bin/sh and a few other utilities to test. More files can be automatically copied once the basic ability to sys various targets is tested.

Needs testing on both MINIX and FAT floppies and hard drives on real systems.
Requires that boot filesystem type is the same type as target drive (MINIX or FAT).
Full debug output is currently on.

Added more debug code to bioshd driver, as some funny things are happening in QEMU when trying to sys a 360k floppy from a 1.44M floppy, which isn't working.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -560,6 +560,7 @@ msg_reboot:
 	.asciz "Press key\r\n"
 
 #ifndef BOOT_FAT
+	.global sect_offset
 sect_offset:
 	.long 0
 #endif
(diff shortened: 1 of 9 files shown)


ghaerr:
Committing for now, further patch coming based on discussions in #835.

PR #848, opened 2020-11-09
[boot] Stamp start sector offset into boot sector for MBR boots

ghaerr:
Enhances the way ELKS uses MBR and boot sectors to boot from hard drives.

Use fixed (but currently different) locations in MINIX and FAT boot sectors to store starting sector offset for partition boots.
Updates `setboot` to stamp partition start offset into FAT and MINIX boot sectors.
Updates `makeboot` to use new fixed location for MINIX boot sectors.

Allows booting ELKS from any MBR, provided that the partition boot sector has the EPB, BPB and start sector offsets updated properly. The new sys/makeboot programs ensure that, and ELKS MBR images created using "make images" are updated properly with set boot.

Removes previous kluge code for determining when booting from an ELKS MBR; no longer required. This also frees up valuable space in the boot sector, which will be very useful should we decide to add FAT32 support to ELKS boot.

Tested on QEMU for all FD, HD MBR and HD flat images.

Currently, setboot.c and makeboot.c contain their own duplicate versions of boot sector offset information; this will be cleaned up in an upcoming boot cleanup PR, which will allow for EPB or MINIX-only offsets to be more easily changed if required.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -548,19 +548,19 @@ msg_reboot:
 end_of_code:
 //------------------------------------------------------------------------------
 
-#ifndef BOOT_FAT
-	.org 0x1F3              // Fixed address of sect_offset for MINIX boot sector
-	.global sect_offset
-sect_offset:
-	.long 0
-#endif
-
 // ELKS disk parameter structure
 // For future expansion, fields should be added at the _front_ of structure
 
-	.org 0x1F7
-
 elks_parms_start:
+
+#ifdef BOOT_FAT
+	.org 0x1F7              // FAT boot sectors use older EPB for now
+#else
+	.org 0x1F3
+	.global sect_offset
+sect_offset:
+	.long 0                 // Partition start sector of MINIX boot block
+#endif
 
 // Disk geometry (CHS)
 
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -551,12 +551,14 @@ end_of_code:
 // ELKS disk parameter structure
 // For future expansion, fields should be added at the _front_ of structure
 
-elks_parms_start:
-
 #ifdef BOOT_FAT
 	.org 0x1F7              // FAT boot sectors use older EPB for now
+elks_parms_start:
+
 #else
 	.org 0x1F3
+elks_parms_start:
+
 	.global sect_offset
 sect_offset:
 	.long 0                 // Partition start sector of MINIX boot block
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -68,21 +68,6 @@ floppy_table:
 
 entry1:
 #endif
-
-	// instructions through _next0 not needed for floppy, only MBR boot
-	cld
-	cmp $0x4c65,%ax			// coming from ELKS MBR?
-	jnz _next0
-	mov 8(%si),%ax			// get MBR start sector from partion table at DS:SI
-	mov 10(%si),%si
-	push %cs
-	pop %es
-	mov $BOOTADDR+sect_offset,%di
-	stosw
-	xchg %ax,%si
-	stosw
-_next0:
-
 	// Get the memory size
 	// TODO: optional BIOS INT 12h
 
@@ -559,15 +544,16 @@ msg_error:
 msg_reboot:
 	.asciz "Press key\r\n"
 
+	.global end_of_code
+end_of_code:
+//------------------------------------------------------------------------------
+
 #ifndef BOOT_FAT
+	.org 0x1F3              // Fixed address of sect_offset for MINIX boot sector
 	.global sect_offset
 sect_offset:
 	.long 0
 #endif
-
-	.global last_code
-last_code:
-//------------------------------------------------------------------------------
 
 // ELKS disk parameter structure
 // For future expansion, fields should be added at the _front_ of structure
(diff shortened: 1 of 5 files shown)


tkchia:
Hello @ghaerr,

  * You probably want to also move `elks_parms_start` from offset `0x1F7` to `0x1F3` --- if `BOOT_FAT` is undefined --- so that the sector offset is included in the ELKS parameter block's length calculation (at offset `0x1FD`).  (This means Minix partitions will "officially" have a longer EPB than FAT partitions.  But that is OK --- `bioshd.c` will gracefully handle the extra field.)
  * The patch means you can no longer simply write a prebuilt disk partition image to an actual partition, and have it "just work" --- since the partition needs to know its own location.  I recall in a previous discussion that you would like to have this functionality?

Other than those, the patch looks good to me.

Thank you!

ghaerr:
Hello @tkchia,

I wasn't sure whether having different EPB versions for FAT and MINIX was Ok, but its backwards ordering solves that problem. I'll move the `elks_param_start` to 0x1F3, which anchors the sect_offset location in the case the EPB is subsequently expanded.

 > The patch means you can no longer simply write a prebuilt disk partition image to an actual partition, and have it "just work" --- since the partition needs to know its own location. I recall in a previous discussion that you would like to have this functionality?

Now that we have a functional `sys` utility, it can ensure that a system can be created that "just works", and it already has to rewrite portions of the target boot sector(s).

After quite a bit of research, there seems to be no rock-solid single way of an MBR communicating the partition being booted. Given that `setboot` still needs to update an ELKS boot sector, and now sys/makeboot also required to do the same anyways, moving the start sector offset into the boot sector allows any MBR to be used, which I thought to be a greater benefit than the previous requirement of using our own MBR and klugy magic numbers. `setboot` already has to modify our own MBR on installation, writing the start sector offset into the active partition.

Although two 32MB FAT and MINIX MBR hard disk images are created by "make images", most users will likely have their own MBR already installed. I am thinking of adding mbr.bin into sys/makeboot to handle the case where a user wants to have ELKS partition their hard disk from scratch. Since we only have mkfs and no "mkfat", this would only work for MINIX filesystems though.

Thank you!

tkchia (changes requested):
I think the `elks_parms_start:` label definition should come _after_ the `.org` directives?  I.e.
```
#ifdef BOOT_FAT
	.org 0x1F7
elks_parms_start:
#else
	.org 0x1F3
elks_parms_start:
	...
#endif
```
Otherwise a lot of extra bytes might be considered to be part of the EPB, when they should not be.

Thank you!

tkchia:
Hello @ghaerr ,

> which I thought to be a greater benefit than the previous requirement of using our own MBR and klugy magic numbers. 

Fair enough. :-)  I guess I should also see if the freed-up space in the boot sector can be put to good use, once this PR is in.

> I am thinking of adding mbr.bin into sys/makeboot to handle the case where a user wants to have ELKS partition their hard disk from scratch. Since we only have mkfs and no "mkfat", this would only work for MINIX filesystems though.

Well... time to port `mkfs.fat` to ELKS, perhaps? :-)

About the `mkfs` command name, the current practice on Linux/x86-64 is to have `mkfs` as just a filesystem-independent front-end which invokes `mkfs.minix`, or `mke2fs`, or `mkfs.fat`, etc.  I think ELKS can probably also use such a scheme.

Thank you!

tkchia (approved):

ghaerr:
Hello @tkchia,

> I think the elks_parms_start: label definition should come after the .org directives?

Oops, thanks.

> I guess I should also see if the freed-up space in the boot sector can be put to good use

If you have time to look at possible support for a FAT32 boot, and/or allowing /linux to be within the first directory sector, rather than the first entry, that would be great:

- Relaxing the requirement for /linux to be the first entry would allow for an ATTR_VOLUME volume label or a "System Volume" special directory, both of which seem to be created in that order by Windows 10 format. The current requirement for /dev to be in the first directory sector, instead of in the first two entries can be easily changed in fs/msdos/inodes.c. All this would allow `sys` to function directly after a Windows 10 format and boot.

- A quick look at the FAT boot shows that only a few changes probably need to be made for FAT32 boot - the root cluster location requiring some decisions as to BPB layout. In [BPB Descriptions](https://en.wikipedia.org/wiki/BIOS_parameter_block) the FAT32 root cluster is offset 0x2C of a DOS 7.1 BPB. We currently support a DOS 4.0 BPB. Rather than extend our 4.0 BPB to 7.1, which requires more reserved bytes, notice that the root cluster offset in 7.1 is overlapped with the DOS 4.0 volume label. 

I propose just adding a bpb_root_cluster label within our existing 4.0 BPB. Then, all ELKS utilities will still protect the FAT BPB locations 11 through 61, but when a FAT32 boot block or drive is encountered, the root clusters longword will be valid, and our boot block will be applied outside that area, and reference it for booting. This approach should guarantee compatibility with DOS disks through MSDOS 6.22. I don't think moving to 7.1 compatibility, with its much larger BPB, is necessary for ELKS at this time.

A FAT32 volume is presumably defined by having the older 16-bit "total logical sectors" at offset 0x13 zero (and the "total large logical sectors" at offset 0x20 non-zero, but that shouldn't need to be checked).

The other change for FAT32 could be to use the newer "hidden sectors" field at 0x1C, instead of "reserved logical sectors" at 0x0E. Offset 0x1C now has the valid partition sector offset and the root directory cluster can then be used to get the base root directory sector. I am not yet certain how/whether the "reserved logical sectors" at 0x0E is valid for FAT32 or not.

Thank you!

Issue #841, opened 2020-11-06
Minor debug ^P issue ++


Mellvik:
A couple of things (really enhancement requests) re. the new dprintk mechanism and debug messages.

- When debug is enabled in /bootopts, enabled debug types, like BLK and FAT, will roll over the screen as expected. TCP debugs will not. ^P will toggle both, turns off those enabled by the debug statement, turns on those in networking (possibly only TCP). I may be wrong, but this cannot be intentional?

- Talking about TCP debug, it would be very helpful if
  - acks and sequence numbers were in decimal
  - acks and sequence numbers were printed as deltas instead of absolute numbers (actual byte count instead of byte count + starting number) (like tcpdump does by default).

--Mellvik

ghaerr:
> When debug is enabled in /bootopts, enabled debug types, like BLK and FAT, will roll over the screen as expected. TCP debugs will not. ^P will toggle both, turns off those enabled by the debug statement, turns on those in networking (possibly only TCP). I may be wrong, but this cannot be intentional?

This is a known issue, not intentional. Not sure how to fix it. The problem is that there's a lot under the hood with ^P: the printk option can be enabled via the /bootopts debug statement, but ktcp isn't yet running to send a special signal to, to toggle its debugging. So when the first (or any subsequent) ^P is typed, the kernel printk toggle occurs, and the signal is sent, which ktcp toggles separately. The signaling method doesn't allow a parameter or kernel communication to determine what the debug state is, ktcp just toggles the state with every signal. The SIGURG signal is used, because that is the only signal that won't terminate a process if received without a signal handler.

This issue normally occurs only when trying to debug kernel startup with printk. Otherwise, it is a better idea to just remove the debug keyword in /bootopts, and hit ^P after boot. Then, both ktcp and the kernel debug will be in sync.

> acks and sequence numbers were in decimal

Ok - I thought it would be easier in hex. I'll just change to your delta idea below, or do you want both?

> acks and sequence numbers were printed as deltas instead of absolute numbers (actual byte count instead of byte count + starting number) (like tcpdump does by default).

Which should be subtracted from the other? I guess we would always like to show a positive number, which would be the difference between the current sequence number and the last ack received?

Mellvik:
Having spent weeks and months with tcpdump listings, I've come to appreciate the speed at which it is possible to scroll through and check acks against sequence numbers when they are in decimal and (relatively small). No thinking, just a snap. 

[...]

PR #849, opened 2020-11-09
[ktcp] updated debug output, mostly for acks and seq numbers.

Mellvik:
Changed ktcp debug output per #841 and #839 (comment):

- removed ICMP debug output from debug_tcp
- changed presentation of sequence numbers and acks to be easier to read (always starting @ 0)
- now displays checksums from the packets in stead of calculating them
- hex output nor preceded with 0x in order to distinguish from decimal
- some seemingly extraneous debug messages commented out for now

Tested on qemu & physical.

### DIFF
--- a/elkscmd/ktcp/icmp.c
+++ b/elkscmd/ktcp/icmp.c
@@ -46,7 +46,7 @@ void icmp_process(struct iphdr_s *iph,unsigned char *packet)
 	apair.daddr = iph->saddr;
 	apair.saddr = iph->daddr;
 	apair.protocol = PROTO_ICMP;
-	ip_sendpacket(packet, len, &apair);
+	ip_sendpacket(packet, len, &apair, NULL);
 	netstats.icmpsndcnt++;
 	break;
    case ICMP_TYPE_DST_UNRCH:
--- a/elkscmd/ktcp/ip.c
+++ b/elkscmd/ktcp/ip.c
@@ -94,7 +94,7 @@ static void ip_print(struct iphdr_s *head, int size)
     debug_ip("tos:%d len:%u ", head->tos, ntohs(head->tot_len));
     debug_ip("id:%u ", ntohs(head->id));
     debug_ip("offs:%d ", ntohs(head->frag_off));	//FIXME add FM_ flags
-    debug_ip("chk:0x%x ", head->check);
+    debug_ip("chk:0x%x ", ip_calc_chksum((char *)head, 4 * IP_HLEN(head)));
     debug_ip("ttl:%d ", head->ttl);
     debug_ip("prot:%d\n", head->protocol);
 #if DEBUG_IP > 1
--- a/elkscmd/ktcp/ip.c
+++ b/elkscmd/ktcp/ip.c
@@ -93,7 +93,8 @@ static void ip_print(struct iphdr_s *head, int size)
     debug_ip("v%d hl:%d ", IP_VERSION(head), IP_HLEN(head));
     debug_ip("tos:%d len:%u ", head->tos, ntohs(head->tot_len));
     debug_ip("id:%u ", ntohs(head->id));
-    debug_ip("offs:%d ", ntohs(head->frag_off));	//FIXME add FM_ flags
+    debug_ip("fl:0x%x ", ntohs(head->id)>>13);
+    debug_ip("fo:%d ", ntohs(head->frag_off)&0x1fff);
     debug_ip("chk:0x%x ", ip_calc_chksum((char *)head, 4 * IP_HLEN(head)));
     debug_ip("ttl:%d ", head->ttl);
     debug_ip("prot:%d\n", head->protocol);
--- a/elkscmd/ktcp/ip.c
+++ b/elkscmd/ktcp/ip.c
@@ -93,8 +93,8 @@ static void ip_print(struct iphdr_s *head, int size)
     debug_ip("v%d hl:%d ", IP_VERSION(head), IP_HLEN(head));
     debug_ip("tos:%d len:%u ", head->tos, ntohs(head->tot_len));
     debug_ip("id:%u ", ntohs(head->id));
-    debug_ip("fo:%x ", ntohs(head->frag_off));	//FIXME add FM_ flags
-    debug_ip("chk:%x ", ip_calc_chksum((char *)head, 4 * IP_HLEN(head)));
+    debug_ip("offs:%d ", ntohs(head->frag_off));	//FIXME add FM_ flags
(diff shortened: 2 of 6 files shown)


ghaerr:
@Mellvik,

Nice job, it took more changes than I thought, mostly I guess since you needed the initial seq/ack values for subtraction.

What is the use of showing the actual checksum values, shouldn't it be important to know what the actual packet checksum is? 

I see also a change is made in TCP such that if the TCP packet has an invalid checksum, it will be noted "Packet dropped, bad checksum" as usual, but there will be no display of that packet for possible analysis. Is that intended?

Mellvik:
@ghaerr,
Thanks for taking a look!

> What is the use of showing the actual checksum values, shouldn't it be important to know what the actual packet checksum is?
> 
I don't really understand the question. To me 'actual checksum' is what's in the packet. For outgoing packets that's what we just calculated, so it didn't make sense to do the same calculation again. For incoming packets a difference n checksums will be caught and flagged anyway, so recalculating it for every packet seemed like a slowdown for old machines. That said, I wasn't particularly interested in checksums at this point, but none of them worked (they were always zero), and I didn't immediately see what was wrong. IOW I found the change useful.
> I see also a change is made in TCP such that if the TCP packet has an invalid checksum, it will be noted "Packet dropped, bad checksum" as usual, but there will be no display of that packet for possible analysis. Is that intended?
> 
Certainly not intentional, I'll take a look at that one. Thanks.

-- M
>

ghaerr:
> For incoming packets a difference n checksums will be caught and flagged anyway, so recalculating it for every packet seemed like a slowdown for old machines. That said, I wasn't particularly interested in checksums at this point, but none of them worked (they were always zero)

Back when I was actually debugging the packet checksums, it was useful to see that the computer checksum was 0, rather than the actual value, which always differed. In this original way, the displayed packet value didn't change with every packet, which otherwise would have made it harder to see other information in a complex trace. It is also nice to have the display fit within 80 columns.

With regards to slowdown for older machines, this is debug code, and only gets calculated when debug is on, which won't be the case with the final release. If you don't really care that much, I'd rather see the computed checksum, which of course should always be 0. I had always found tcpdump to be very confusing in this regard, since it continually displays checksum values which are non-zero (and mentions bad checksum), while ELKS and Linux don't complain about the checksum values (because they were valid).

I think the key here is ease of readability - which was the reason for changing the seq/ack numbers to be relative. For me, keeping the computed checksum value is also easier to read.

Thank you!

Mellvik:
> 
> Back when I was actually debugging the packet checksums, it was useful to see that the computer checksum was 0, rather than the actual value, which always differed. In this original way, the displayed packet value didn't change with every packet, which otherwise would have made it harder to see other information in a complex trace. It is also nice to have the display fit within 80 columns.
> 
Good point, I can appreciate that.
> With regards to slowdown for older machines, this is debug code, and only gets calculated when debug is on, which won't be the case with the final release.
> 
I'm not worried about production code, but the debugging code's effect on the behaviour. Dealing with time sensitive stuff like this, I'm always looking for optimization potential.
> If you don't really care that much, I'd rather see the computed checksum, which of course should always be 0. I had always found tcpdump to be very confusing in this regard, since it continually displays checksum values which are non-zero (and mentions bad checksum), while ELKS and Linux don't complain about the checksum values (because they were valid).
> 
Again I can appreciate that. My debugging needs seq/ack focus, not checksums. I'll revert the checksums to the previous immediate calculations.
-M

ghaerr:
I'm ok with this, but - what happens if we want to debug ICMP for some reason?

It looks like Lines 35-41 in tcp.c should stay, and then just use a line like the following in ip.c, so that ICMP could be turned on again when needed:
```
if (iph->protocol == PROTO_TCP /*|| iph->protocol == PROTO_ICMP*/ ) {
```

I can see its a little hard to fix debug display dropped or bad checksum packets, I can fix that later.

Mellvik:
@ghaerr,

> 
> I'm ok with this, but - what happens if we want to debug ICMP for some reason?
> 
> 
I removed ICMP because it is covered by DEBUG_IP in config. The output becomes really messy when both are active (and tcp_print is not good for ICMP logging). This gives us the option to have one without the other if desired.
> It looks like Lines 35-41 in tcp.c should stay, and then just use a line like the following in ip.c, so that ICMP could be turned on again when needed:
> 
> if (iph->protocol == PROTO_TCP /*|| iph->protocol == PROTO_ICMP*/ ) {
> I can see its a little hard to fix debug display dropped or bad checksum packets, I can fix that later.
> 
> 
I will change tcp_print to skip acks and sequence numbers if the tcpcb pointer is NULL. Then the call from tcp_process can be moved back up to before the checksum test.

Thank you!

—Mellvik

ghaerr:
> I removed ICMP because it is covered by DEBUG_IP in config. The output becomes really messy when both are active (and tcp_print is not good for ICMP logging). This gives us the option to have one without the other if desired.

Ok.

> I will change tcp_print to skip acks and sequence numbers if the tcpcb pointer is NULL. 

But then tcpcb will always be NULL, right? Since the following call that gets the cb is later:
```
cbnode = tcpcb_find(iph->saddr, ntohs(tcph->dport), ntohs(tcph->sport));
```

Mellvik:
> I will change tcp_print to skip acks and sequence numbers if the tcpcb pointer is NULL.
> 
> But then tcpcb will always be NULL, right? Since the following call that gets the cb is later:
> 
> cbnode = tcpcb_find(iph->saddr, ntohs(tcph->dport), ntohs(tcph->sport));
I will move the cbnode search up and add another test to make it right.

We can also do the refused packet test first, but I guess the order doesn't make much difference here..


    /* *cb initialized to NULL */
    cbnode = tcpcb_find(iph->saddr, ntohs(tcph->dport), ntohs(tcph->sport));
    if (cbnode) cb = &cbnode->tcpcb;
    tcp_print(&iptcp, 1, cb);

    if (tcp_chksum(&iptcp) != 0) {
        printf("tcp: BAD CHECKSUM (%x) len %d\n", tcp_chksum(&iptcp), iptcp.tcplen);
        netstats.tcpbadchksum++;
        return;
    }

    if (!cbnode) {
        printf("tcp: Refusing packet %s:%d->%d\n", in_ntoa(iph->saddr),
                ntohs(tcph->sport), ntohs(tcph->dport));

        /* TODO : send RST and stuff */
        netstats.tcpdropcnt++;
        return;
    }

— Mellvik

ghaerr:
> I will move the cbnode search up and add another test to make it right.

Ok.

ghaerr:
Well done @Mellvik, thanks!

PR #850, opened 2020-11-10
[fat] Expand FAT /dev from 16 to 28 entries, fix ttyS0 open

ghaerr:
Various small fixes and enhancements for FAT filesystem.

Expands /dev with 8 more entries, including /dev/hda2 and /dev/hdb, /dev/hdb1.

Allows activation of "fake" /dev if /dev is in first four entries in root directory, rather than two. This, along with a possible future relaxation of the requirement of /linux to be the first entry in the root directory, will help booting off filesystems created using Windows 10 format.

Fixes inability to open /dev/ttyS0 when booted from FAT because of MSDOS upper -> lower case filename conversions. The new code treats /dev entries specially.

### DIFF
--- a/elks/fs/msdos/inode.c
+++ b/elks/fs/msdos/inode.c
@@ -15,26 +15,32 @@
 #include <linuxmt/debug.h>
 
 #ifdef CONFIG_FS_DEV
-/* max 16 entries in FAT device table*/
+/* FAT device table*/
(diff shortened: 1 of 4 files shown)


toncho11:
Get pull request:
` git fetch origin pull/850/head:fat5`

Switch to branch:
`git checkout fat5`

PR #851, opened 2020-11-10
[kernel] Limit size of /dev/hda to max filesize (2^31)-1

ghaerr:
Fixes issue in https://github.com/jbruchon/elks/issues/835#issuecomment-724827503.
Also adds sync to mkfs.

@toncho11, try this, it should fix your fdisk problem on /dev/hda with a HD size > 2GB.

[fd360-fat.bin.zip](https://github.com/jbruchon/elks/files/5519374/fd360-fat.bin.zip)
[fd360-minix.bin.zip](https://github.com/jbruchon/elks/files/5519375/fd360-minix.bin.zip)

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -496,7 +496,10 @@ static int bioshd_open(struct inode *inode, struct file *filp)
 	probe_floppy(target, hdp);
 #endif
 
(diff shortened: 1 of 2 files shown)


toncho11:
IMG_20201110_193102

It does.

ghaerr:
Great. Let me know what other problems you have with sys on FAT, so that we can move to the next step and enhance sys with other options/files.

PR #852, opened 2020-11-11
[cmds] Add -M option to sys/makeboot to write MBR

ghaerr:
Enhances sys and makeboot to take -M option to write MBR, which allows creating ELKS hard drive (MBR) volumes completely from scratch from a running ELKS system.

Solves all known problems with `sys` found and discussed in #835.

For example, `sys -M /dev/hda1` will now copy the boot drive's boot sector, /linux and other system files to hard disk partition 1, and write an MBR.

The process with which to copy/create ELKS systems from ELKS is now:
- Use `fdisk` to create a new partition.
- For MINIX, use `mkfs <partition> <size>` to create MINIX filesystem.
- For FAT, use an external mkfs.fat to create FAT filesystem on partition.
- Run `sys <partition>` to copy boot block and /linux. Run `sys -M <partition>` if fdisk was used on never-used hard disk.

\<partition> is the hard disk and partition number, e.g. /dev/hda1, /dev/hda2, /dev/hdb1, etc.
\<size> is in 1024-byte blocks.

Restrictions: The target device must use the same filesystem as the booted ELKS device.

Example: to create/copy ELKS to new hard drive partition 1 of 65MB, using MINIX:
```
fdisk (create partition 1 of size 65MB, make active)
mkfs /dev/hda1 65535
sys /dev/hda1
```

This PR also cleans up makeboot considerably, and adds error checking to sys parameters.
The MBR is created using `xxd -i` to convert the real mbr.bin MBR into C source, then included into makeboot.c!

@toncho11, thanks for your marathon testing. Please continue to exercise this, and be careful with your partition start/end sizes :) I've also included new binaries for you.

[fd360-fat.bin.zip](https://github.com/jbruchon/elks/files/5521285/fd360-fat.bin.zip)
[fd360-minix.bin.zip](https://github.com/jbruchon/elks/files/5521287/fd360-minix.bin.zip)

Everyone: we need lots more testing on MINIX and FAT, but I think this may be working pretty well now.

### DIFF
--- a/bootblocks/.gitignore
+++ b/bootblocks/.gitignore
@@ -1,2 +1,3 @@
 *.bin
 minix_first.s
+mbr_autogen.c
--- a/bootblocks/Makefile
+++ b/bootblocks/Makefile
@@ -42,6 +42,7 @@ mbr.bin: mbr.S
 	$(CC) $(INCLUDES) -E -o mbr.tmp mbr.S
 	$(AS) $(ASFLAGS) -o mbr.o mbr.tmp
 	$(LD) $(LDFLAGS) -M -o mbr.bin mbr.o > mbr.map
+	xxd -i mbr.bin > mbr_autogen.c
 	rm -f mbr.tmp
 
 boot_sect_fat.o: boot_sect.S boot_sect_fat.h $(TOPDIR)/include/autoconf.h
@@ -50,7 +51,7 @@ boot_sect_fat.o: boot_sect.S boot_sect_fat.h $(TOPDIR)/include/autoconf.h
 	rm -f boot_sect_fat.tmp
 
 clean:
-	rm -f *.o *.bin *.map
+	rm -f *.o *.bin *.map mbr_autogen.c
 
 # Boot blocks are not part of the target filesystem
 # but embedded in the target disk image
--- a/bootblocks/mbr.S
+++ b/bootblocks/mbr.S
@@ -279,4 +279,6 @@ end_of_code:
 
 	.org	0x1BE	// MBR partition table (left out)
 
+	.org	0x1FE	// Boot signature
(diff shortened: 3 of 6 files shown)


toncho11:
If you are writing the MBR for the fat version, then this means that you set fat16 or fat32 in the MBR, right?
I can not see this in the code.
So how do you switch between fat16 and fat32?
On needs to change the partition type in the MBR?

ghaerr:
> If you are writing the MBR for the fat version, then this means that you set fat16 or fat32 in the MBR, right?

No, the MBR doesn't need to know the filesystem type, since it only reads a single sector, which is the second boot sector. The sector it reads is at the "start offset" position specified in the partition table.

> So how do you switch between fat16 and fat32?

A FAT filesystem is set by mkfs.fat to either FAT12, FAT16, or FAT32 based on the size of the partition and the cluster size; one cannot specify the type. If a partition is less than 4085 clusters, (~2MB), it will be FAT12. If less than 63504 clusters (~32MB), it will be FAT16, otherwise FAT32. This is assuming clustersize=sectorsize.

The mkfs.fat program sets the filesystem type in the second boot sector (Volume boot record or VBR). The sys program doesn't touch that data, but only sets the CHS values in that boot sector.

The ELKS boot sector currently supports booting from FAT12 and FAT16 filesystems, but not yet FAT32. I think we will be able to add FAT32 support, but it isn't started yet.


> On needs to change the partition type in the MBR?

No, the partition type doesn't matter for ELKS, it isn't actually used.

toncho11:
Thanks. 
"Usage" in the sys tool is missing the "[options]" and the M explication.

ghaerr:
Thanks @toncho11. I also updated the messages a bit with some ideas from our conversation in #835.

PR #854, opened 2020-11-11
[boot] FAT12/16 boot: search entire root directory for /linux

tkchia:
With this patch, `/linux` can now be anywhere in the root directory of a FAT12 or FAT16 filesystem.  However, `/linux`'s contents must still be laid out in contiguous sectors.

The bootloader now also checks that
  * it is not started from a FAT32 partition;
  * the root directory is not too large for it to process;
  * and that the `/linux` entry is not actually a volume label or part of a long file name.

If the bootloader detects either of the first two conditions, it will now say `5!`.

### DIFF
--- a/bootblocks/boot_err.h
+++ b/bootblocks/boot_err.h
@@ -4,4 +4,4 @@
 #define ERR_PAYLOAD     2
 #define ERR_NO_SYSTEM   3
 #define ERR_BAD_SYSTEM  4
-
+#define ERR_FS_ODDITY   5
--- a/bootblocks/boot_sect_fat.h
+++ b/bootblocks/boot_sect_fat.h
@@ -143,45 +143,65 @@ bpb_fil_sys_type:			// Filesystem type (8 bytes)
 .macro FAT_LOAD_AND_RUN_KERNEL
 	.set buf, entry + 0x200
 
-	// Load the first sector of the root directory
+	// Load the root directory
 	movb bpb_num_fats,%al
 	cbtw
 	mulw bpb_fat_sz_16
 	addw bpb_rsvd_sec_cnt,%ax
-	push %ax
-	inc %dx				// Assume DX was 0 (from mulw); if
-					// DX != 0 then we are in trouble
-					// anyway
+	mov bpb_root_ent_cnt,%bp	// Check that the root directory is
+	dec %bp				// not too large for us to process
+					// (this is unlikely to happen: MS
+					// says, "For maximum compatibility,
+					// FAT16 volumes should use the [count]
+					// value 512"); while at it also check
+					// that this is not actually a FAT32
+					// filesystem (root entries = 0)
+	cmp $(0xf000-(buf-entry))/0x20-1,%bp
+	ja fs_oddity
+	inc %bp
+	mov %bp,%dx
+	add $0xf,%dx
+	mov $4,%cl
+	shr %cl,%dx
+	push %dx			// Remember the number of root
+					// directory sectors...
+	push %ax			// ...& the root directory start
 	mov $buf,%cx
-	mov %cx,%si
 	push %ss
 	call disk_read
 
-	// See if the first directory entry is /linux; bail out if not
-	mov $kernel_name,%di		// SI = buf
+	// See if /linux appears in the root directory; bail out if not
+	mov $buf-0x20,%si
+find_system:
+	add $0x20,%si
+	decw %bp
+	js no_system
+	mov $kernel_name,%di
 	mov $0xb,%cx
+	push %si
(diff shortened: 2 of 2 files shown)


ghaerr:
Hello @tkchia,

Looks great, thanks for doing this  :) 

Have you tested it much, for instance with the Windows 10 formatter and its "System Volume" directory entry?

Thank you!

tkchia:
Hello @ghaerr,

> Have you tested it much, for instance with the Windows 10 formatter and its "System Volume" directory entry?

Not yet.  This probably needs a lot more stress-testing than I can do right now. :-|

Thank you!

ghaerr:
Hello @tkchia,

A potential problem with allowing the entire root directory to be searched, rather than just permitting, say the first 4 entries, is that it permits a large (likely fragmented) disk to be created with /linux not contiguous, and then booted. It is very unlikely that /linux would be contiguous after a few root directory entries are written. Thus, the boot will NOT display an error, and the system will crash while loading a non-contiguous /linux.

What do you think about this? Wouldn't it be better to just limit /linux to be within the first 4 directory entries?

Thank you!

tkchia:
Hello @ghaerr ,

You have a point.  Let me rework the patch and resubmit it.  Thanks!

ghaerr:
Hello @tkchia,

> You have a point. Let me rework the patch and resubmit it.

Thank you. I am hoping that, with less code and checking required, it will also still leave enough extra space for a FAT32 boot to be considered!

tkchia:
Hello @ghaerr,

> What do you think about this? Wouldn't it be better to just limit /linux to be within the first 4 directory entries?

I have submitted a PR which searches slightly more directory entries --- currently 8, but this can be adjusted in the source code.

One thing that came to my mind was that a VFAT long file name will take up more than one directory entry slot.  The good old `/System Volume Information` name easily takes up 3 slots.

> I am hoping that, with less code and checking required, it will also still leave enough extra space for a FAT32 boot to be considered!

I suspect having one unified bootloader binary for FAT12, FAT16, and FAT32 would be tricky.  A separate bootloader for FAT32 is probably more feasible --- plus, FAT32 expressly supports multiple reserved sectors. ☺

Thank you!

ghaerr:
Hello @tkchia,

> One thing that came to my mind was that a VFAT long file name will take up more than one directory entry slot. The good old /System Volume Information name easily takes up 3 slots.

You're right - I didn't even think of that.

> FAT32 expressly supports multiple reserved sectors.

That's just the "reserved logical sectors" word at BPB 0x0E (not to be confused with the "count of hidden sectors before the volume" longword at 0x1C), right? 

FAT16 usually as 1 reserved logical sectors, and it seems FAT32 has 32 (at least). We're already adding that word in, what is the concern here?

> I suspect having one unified bootloader binary for FAT12, FAT16, and FAT32 would be tricky.

I'll take a pass at it, unless you'd like to :) 

I think it really depends on how much space we have left - with what we have now, there are only two big changes left:
- The FAT32 root directory is at a different cluster, at 0x2C (and almost always =2).
- Determining that the volume is FAT32. This is easily done by checking "total logical sectors" at 0x13 == 0.

I think that's it.

Thank you!

tkchia:
Hello @ghaerr,

> I'll take a pass at it, unless you'd like to :)

OK, please do. ☻

> FAT16 usually as 1 reserved logical sectors, and it seems FAT32 has 32 (at least). We're already adding that word in, what is the concern here?

Less of a concern but more of a potential opportunity.  It means that we can use more than 1 sector for the FAT32 bootloader code, i.e. we can afford to make it more powerful (though there are probably some subtle details to sort out if we do this).

Thank you!

ghaerr:
Hello @tkchia,

> Determining that the volume is FAT32. This is easily done by checking "total logical sectors" at 0x13 == 0.

I misspoke: FAT32 is determined by checking whether the word "logical sectors per FAT" (0x16) == 0, in which case the volume is deemed FAT32 and the FAT32 EBPB longword at sector offset 0x24 is used for logical sectors/FAT. This longword sits right on top of the current ELKS bpb_drv_num, bpb_reserved, bpb_bootsig and bpb_volid first byte, which ELKS doesn't use.

The FAT32 "root directory cluster number" at 0x2C also sits on top of the BPB 4.0 volume label (bpb_volid) which ELKS also don't use.

According to [Wikipedia](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system#FAT32_Extended_BIOS_Parameter_Block), it can get tricky to recognize every FAT16 or FAT32 volume created by various OS's, but I don't think ELKS needs to be concerned about that at this point. 

Thank you!

PR #857, opened 2020-11-12
[boot] FAT12/16 boot: search 1st 8 root dir. entries for /linux

tkchia:
(Reworked from https://github.com/jbruchon/elks/pull/854 .)

With this patch, the bootloader will be able to find `/linux` as long as it is in the first 8 root directory entries of a FAT12 or FAT16 filesystem (including hidden entries for long file names etc.).  However, `/linux`'s contents must still be laid out in contiguous sectors.

The bootloader now also checks that the `/linux` directory entry it finds is for a normal file.

This patch probably needs more stress-testing.

### DIFF
--- a/bootblocks/boot_sect_fat.h
+++ b/bootblocks/boot_sect_fat.h
@@ -153,20 +153,35 @@ bpb_fil_sys_type:			// Filesystem type (8 bytes)
 					// DX != 0 then we are in trouble
 					// anyway
 	mov $buf,%cx
-	mov %cx,%si
 	push %ss
 	call disk_read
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

Looks good, thanks for the change. Lets go ahead and commit, to enable further testing.
I think we should tighten down the ROOT_ENTS_TO_CHECK after learning a bit more about exactly how many directory entries various Windows and other formatters add to the root directory; these may also vary based on FAT type. I suppose the sys command could be made to check for correctness of /linux at the time it writes it, as well.

Thank you!

tkchia:
@ghaerr: OK, thanks!

PR #858, opened 2020-11-13
[boot] Enhance boot loader to boot ELKS on FAT32 filesystems

ghaerr:
The ELKS FAT boot sector is enhanced to boot FAT12, FAT16 or FAT32 filesystems.

Works with `mkfat` and other FAT32 filesystem formatters and should allow booting from them after running `sys`.

Along with @tkchia's enhancement #857 relaxing the restriction of /linux having to be the first root directory entry, this change should allow booting from all freshly created FAT32 filesystems, including those made with ELKS mkfat and Windows 10.

The FAT boot sector code assumes that the root cluster will be at 2, which is always the first data cluster available. This will always be the case unless bad sectors are present, since the first data cluster (numbered 2) will always be allocated directly after the FAT tables. Checking for a root directory cluster other than 2 could be added after testing, but there are no known utilities that put the FAT32 root directory at another cluster location, so testing that is difficult, and likely unnecessary, given the other assumptions already made in the FAT boot code.

Also adds "sys -F" option to allow sys'ing flat (non-MBR) hard disks. This was required for my testing.
The msdos_fs.h header is updated with BPB offsets to help debugging.

Tested on QEMU with FAT16 and FAT32 filesystems of different sizes.

@toncho11, it would be interesting to try booting a FAT32 partition on your hardware, made using `mkfat` or Windows 10 format. I've attached updated 360k images with all the past few days sys fixes and the FAT32 boot sector.

[fd360-fat.bin.zip](https://github.com/jbruchon/elks/files/5534707/fd360-fat.bin.zip)
[fd360-minix.bin.zip](https://github.com/jbruchon/elks/files/5534708/fd360-minix.bin.zip)

### DIFF
--- a/bootblocks/boot_sect_fat.h
+++ b/bootblocks/boot_sect_fat.h
@@ -139,6 +139,7 @@ bpb_vol_label:				// Volume label (11 bytes)
 bpb_fil_sys_type:			// Filesystem type (8 bytes)
 	.ascii "FAT12   "
 .endm
+	.set bpb_fat_sz_32,0x24		// FAT32 fat size, 32-bit
 
 .macro FAT_LOAD_AND_RUN_KERNEL
 	.set buf, entry + 0x200
@@ -146,7 +147,11 @@ bpb_fil_sys_type:			// Filesystem type (8 bytes)
 	// Load the first sector of the root directory
 	movb bpb_num_fats,%al
 	cbtw
-	mulw bpb_fat_sz_16
+	mov bpb_fat_sz_16,%bx		// check FAT16 fat size
+	and %bx,%bx
+	jnz 1f				// nonzero means FAT16 filesystem
+	mov bpb_fat_sz_32,%bx		// get loword FAT32 fat size instead
+1:	mulw %bx
 	addw bpb_rsvd_sec_cnt,%ax
 	push %ax
 	inc %dx				// Assume DX was 0 (from mulw); if
--- a/elks/include/linuxmt/msdos_fs.h
+++ b/elks/include/linuxmt/msdos_fs.h
@@ -63,16 +63,16 @@ struct msdos_boot_sector {
 	unsigned short fat_length;  /* sectors/FAT 22*/
 	unsigned short secs_track;  /* sectors per track (unused) 24*/
 	unsigned short heads;	    /* number of heads (unused) 26*/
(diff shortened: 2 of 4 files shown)


toncho11:
So it worked first boot on fat32 formatted with mkfat.
It booted also after when "Storage Service" on Windows added the "System Volume". 

It also booted after installing on fat32 formatted with Windows 10 (with "System volume" folder). But here there was an error:
```
VFS: mounted root 0x0301 (msdos filesystem).
Unable to open /dev/console (error -2).
```
If a boot from floppy the fs is mountable.

tkchia (approved):

tkchia:
Hello @ghaerr,

Thanks for the changes.

I guess my main worry is that there are more ways for things to fail, now that `/linux` may be written onto a FAT filesystem with existing files, and now that we want to also support booting from FAT32.

It is probably time to look into how to get `sys` to check that `/linux` is written in the right place(s), or how to make the bootloader more flexible.

Thank you!

ghaerr:
Hello @toncho11,

> So it worked first boot on fat32 formatted with mkfat.
> It booted also after when "Storage Service" on Windows added the "System Volume".

Great, good news!

> It also booted after installing on fat32 formatted with Windows 10 (with "System volume" folder). But here there was an error:
Unable to open /dev/console (error -2).

I suspect the reason for this is that the /dev directory entry in the root directory was not within the first four entries. I will increase that to 8, same as the boot sector requirement for /linux.

Can you send me an image of a new Windows 10 FAT32 formatted volume (with System Volume) but before running sys?

Thank you!

ghaerr:
Hello @tkchia,

> It is probably time to look into how to get sys to check that /linux is written in the right place(s), or how to make the bootloader more flexible.

Given the amount of free space in the boot sector, its probably better to have sys check and complain, which is the same as what MSDOS does. I've added that to my list.

Thank you!

toncho11:
@ghaerr here it is the empty fat32 image **without** "System volume"
[elks_hdd_image_fat32_empty.zip](https://github.com/jbruchon/elks/files/5537775/elks_hdd_image_fat32_empty.zip)

And here it is **with** "System volume":
[elks_hdd_image_fat32_empty_with_volume.zip](https://github.com/jbruchon/elks/files/5537809/elks_hdd_image_fat32_empty_with_volume.zip)

PR #861, opened 2020-11-13
[cmds] makeboot: check that FAT32 root dir. starts at cluster 2

tkchia:
`makeboot` (and `sys`) now checks that the root directory of a target FAT32 partition starts right after the allocation tables --- as needed by the current FAT bootloader --- before modifying the partition.

### DIFF
--- a/elkscmd/sys_utils/makeboot.c
+++ b/elkscmd/sys_utils/makeboot.c
@@ -55,9 +55,12 @@
 #define FATBPB_END		61				/* through end of file system type*/
 
 /* FAT-only offsets */
+#define FAT_BPB_RootEntCnt	0x11		/* offset of FAT12/16 root directory entry count (word) */
+#define FAT_BPB_FATSz16		0x16		/* offset of FAT12/16 allocation table sector count (word) */
 #define FAT_BPB_SecPerTrk	0x18		/* offset of sectors per track (word) */
 #define FAT_BPB_NumHeads	0x1A		/* offset of number of heads (word) */
 #define FAT_BPB_SectOffset	0x1C		/* offset of partition start sector (long) */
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

This is fine, except that both the FAT boot loader as well as the FAT filesystem driver use offset 0x16 (number of sectors for FAT for FAT16), rather than offset 0x11 (number of FAT16 root directory entries), for determining whether a filesystem is FAT32 or not. Should not we keep everything the same, rather than introducing another way of recognizing FAT32?

Here's the relevant code in the FAT driver:
```
   if(!b->fat_length && b->fat32_length){
        fat32 = 1;
        ...
```
The FAT boot code uses the following to determine FAT32:
```
        mov bpb_fat_sz_16,%bx           // check FAT16 fat size
        and %bx,%bx
        jnz 1f                          // nonzero means FAT16 filesystem
        mov bpb_fat_sz_32,%bx           // get loword FAT32 fat size instead
```

Also, '\#define FAT_BPB_BkBootSec	0x32' is unused.

Thank you!

tkchia:
Hello @ghaerr,

> Should not we keep everything the same, rather than introducing another way of recognizing FAT32?

OK.  (It seems the official Microsoft way to recognize FAT32 is to actually calculate the cluster count, but that seems overkill.)

Thank you!

ghaerr:
Hello @tkchia,

> It seems the official Microsoft way to recognize FAT32 is to actually calculate the cluster count, but that seems overkill.

Yes, agreed. I have been pretty deep into all the FAT documentation and our source, and this isn't the only area where ELKS, as well as other filesystem code, diverges! I thought about changing things but figured we're probably best leaving things the way they are, at least until we find other issues.

Thank you!

PR #867, opened 2020-11-16
[cmds] Add cp -R recursive copy command option

ghaerr:
Usage: cp [-R][-v] source [ ... ] target_file_or_directory

Adds recursive copy option, with special consideration for copying entire disks. For instance, the following will work:
```
cp -R / /mnt      or
cp -R / /mnt/save
```

Copies symlinks and special device files (char/block devices) properly.
FAT filesystems will give errors on symlinks and special files, but keep copying.
Currently, -v (verbose) is always set, for testing and observation.

This is the smallest recursive copy by far. Compare with BSD, Android, BusyBox or Coreutils, they are huge!
There may be a few more enhancements to make to do with FAT copying and possibly removing error messages for FAT, but I wanted to get the majority of this available before 0.4.0.

We're running out of space on 1.44M floppies, removed nxterm.

@Mellvik: `rm -rf` is broken for large recursive removes. I tried debugging and saw that it used a large buffer in a recursive routine, which immediately caused stack overflow. I changed that in this PR, but it's still broken. This can all stay until after 0.4.0, I'd rather get the network PTY 512 character drop fixed instead, if you have time.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -177,6 +177,6 @@ prems/pres/pres					:other
 nano-X/bin/nxclock				:nanox					:1440k
 nano-X/bin/nxdemo				:nanox					:1440k
 nano-X/bin/nxlandmine			:nanox				:720k
-nano-X/bin/nxterm				:nanox					:1440k
+nano-X/bin/nxterm				:nanox
 nano-X/bin/nxworld				:nanox					:1440k
 nano-X/bin/nxworld.map	::lib/nxworld.map	:nanox		:1440k
--- a/elkscmd/file_utils/Makefile
+++ b/elkscmd/file_utils/Makefile
@@ -32,7 +32,7 @@ cmp: cmp.o
 	$(LD) $(LDFLAGS) -o cmp cmp.o $(LDLIBS)
 
 cp: cp.o
-	$(LD) $(LDFLAGS) -o cp cp.o $(LDLIBS)
+	$(LD) $(LDFLAGS) -maout-heap=0xffff -o cp cp.o $(LDLIBS)
 
 df: df.o
 	$(LD) $(LDFLAGS) -o df df.o $(LDLIBS)
@@ -68,7 +68,7 @@ mv: mv.o
 	$(LD) $(LDFLAGS) -o mv mv.o $(LDLIBS)
 
 rm: rm.o
-	$(LD) $(LDFLAGS) -o rm rm.o $(LDLIBS)
+	$(LD) $(LDFLAGS) -maout-stack=12228 -o rm rm.o $(LDLIBS)
 
 rmdir: rmdir.o
 	$(LD) $(LDFLAGS) -o rmdir rmdir.o $(LDLIBS)
--- a/elkscmd/file_utils/cp.c
+++ b/elkscmd/file_utils/cp.c
@@ -43,9 +43,6 @@ int copyfile(char *srcname, char *destname, int setmodes);
 #define MAJOR_SHIFT	8
 #endif
 
-typedef unsigned short u16_t;
-typedef unsigned long u32_t;
-
 static char buf[BUF_SIZE];
 
 struct list_node_s {
@@ -56,7 +53,7 @@ typedef struct list_node_s list_node_t;
 
 struct list_root_s {
 	list_node_t	node;
-	u16_t		count;
+	unsigned short count;
 };
 typedef struct list_root_s list_root_t;
 
@@ -66,9 +63,8 @@ struct inode_build_s {
 	list_node_t	node;
 	list_root_t	entries;/* list of entries for directory */
 	char       *path;
-	u16_t		index;	/* inode #*/
-	int			dev;	/* major & minor for device */
-	u16_t		flags;
+	dev_t		dev;	/* major & minor for device */
+	unsigned short flags;
 	unsigned long blocks;	/* disk blocks required*/
 };
 typedef struct inode_build_s inode_build_t;
@@ -149,7 +145,7 @@ static inode_build_t * inode_alloc(inode_build_t * parent, char *path)
 	list_init(&inode->entries);
 	inode->path = strdup(path);
 	list_add_tail(&inodes, &inode->node);
-	inode->index = inodes.count;	/* 0 = no inode */
+	//inode->index = inodes.count;	/* 0 = no inode */
 	return inode;
 }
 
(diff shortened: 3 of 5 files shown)


Mellvik:
Thanks @ghaerr, this is good news!

I have a few comments - cp:
- In unix and linux, the option for recursive in cp is lower r, IMHO it is user friendly to keep it for consistency.
- Further, in the name of the same consistency (!), since neither Linux or Unix allow copying of devices, I’d recommend a special option for that, such as the cap R you are introducing (too bad I haven't completed the new tar yet, which would have covered that).
- Finally, copying symlinks would also benefit from its own option, which would determine whether to copy the link or follow it.
I believe these changes would be possible without significant ‘fattening’ of ls. Also, adding -p to keep the time stamps would be very beneficial...

rm:
Agreed.

PTY: This is a tricky beast, I’ve spent what seems like enormous amounts of time determining what isn’t the problem(!). I’m traveling this week but still on it. BTW - possibly unrelated, but I’ve repeatable indications that running netstat while heavy (telnet) transfers are ongoing, will affect transfers, sometimes cause dropped packets (IP/TCP level). Also, there are indications that the lost data in telnet sessions are always full packets, size varying between 2 bytes and 512 bytes.

More about that separately, I suggest we create an issue for the PTY(?) problem to keep track of the process.

—Mellvik

ghaerr:
@Mellvik,

Regarding -r vs -R: -R was chosen precisely because the primary purpose of getting a recursive `cp` was to copy disks, which require the /dev directory to be copied.

Regarding -r, macOS has this to say:
```
Historic versions of the cp utility had a -r option.  This implementation
     supports that option; however, its use is strongly discouraged, as it does
     not correctly copy special files, symbolic links, or fifo's.
```

Regarding options for symlink and other desired options: these can be added later, once this large change gets better tested. Currently, `cp` always keeps the time stamps identical in both normal and recursive modes.


Thanks for your continued concern regarding one of our last known TCP/IP data drop failures. It sounds like there's more than one issue there. I would say the 2-512 byte drop issues (that I am guessing is PTY-related) is the most important. When you have time, or learn more about it, a seperate issue can be opened for it. Given the speed with which we are moving on it, this may or may not make it into 0.4.0, which otherwise seems quite stable.

Mellvik:
> Regarding options for symlink and other desired options: these can be added later, once this large change gets better tested.
> 
Agreed.
> Currently, cp always keeps the time stamps identical in both normal and recursive modes.
> 
OK, that’s good to know since it’s different from other *x systems. Also - I guess ownership is not very important at this point, everything being root, but it may make sense to have cp without and with -p  behave like other systems somewhere down the line.
> Thanks for your continued concern regarding one of our last known TCP/IP data drop failures. It sounds like there's more than one issue there. I would say the 2-512 byte drop issues (that I am guessing is PTY-related) is the most important. When you have time, or learn more about it, a seperate issue can be opened for it. Given the speed with which we are moving on it, this may or may not make it into 0.4.0, which otherwise seems quite stable.
> 
Yes, this is indeed slow coming. I’ve just verified that data is disappearing in whole packets, small or large, some times several packets in a row (missing data may be 2 bytes, 5k bytes or somewhere in between). Any hints as to where to look appreciated.

There is another piece of testing pending - on my list - the large file creation/deletion on minix FS, including creating errors by overflowing the FS.
I was hoping to nail the PTY issue first.


—mellvik

ghaerr:
> I guess ownership is not very important at this point, everything being root,

Sorry - I mean to say that both time stamps and ownership have always been copied by ELKS cp.

> I’ve just verified that data is disappearing in whole packets, small or large, some times several packets in a row (missing data may be 2 bytes, 5k bytes or somewhere in between). Any hints as to where to look appreciated.

Well, there are no 5K size packets being sent, so that might be something else.

> Any hints as to where to look appreciated.

As mentioned previously, we can fairly easily debug the PTY code, if that's the issue. Change the PTY_OUTQSIZE in include/linuxmt/ntty.h to something other than 512, like perhaps 400 or 800, and see whether that immediately changes the size of lost data to something near the new number.

Mellvik:
> 
> I’ve just verified that data is disappearing in whole packets, small or large, some times several packets in a row (missing data may be 2 bytes, 5k bytes or somewhere in between). Any hints as to where to look appreciated.
> 
> Well, there are no 5K size packets being sent, so that might be something else.
> 
No, it's not. Obvoiusly several packets in a row will easily add up to 5k.

> As mentioned previously, we can fairly easily debug the PTY code, if that's the issue. Change the PTY_OUTQSIZE in include/linuxmt/ntty.h to something other than 512, like perhaps 400 or 800, and see whether that immediately changes the size of lost data to something near the new number.
> 
This does not make sense, the hint being 'always full packets lost'. 512 is not an interesting number here.

-M

>

ghaerr:
> 512 is not an interesting number here.

Yes, it is - since we are only seeing this on telnetd, and the PTY driver is the only character queue that is 512 bytes long. My theory is, that the kernel wakeup doesn't happen at the right time, and the PTY driver character queue overflows, and then writes a few more characters into the now-wrapped buffer. This is why we sometimes see other numbers above or around the "modulo 512" number. My idea is to test using the above procedure, which will either show PTY as being in or out with regards to being part of the problem.

Mellvik:
> ﻿
> 512 is not an interesting number here.
> 
> Yes, it is - since we are only seeing this on telnetd, and the PTY driver is the only character queue that is 512 bytes long. My theory is, that the kernel wakeup doesn't happen at the right time, and the PTY driver character queue overflows, and then writes a few more characters into the now-wrapped buffer. This is why we sometimes see other numbers above or around the "modulo 512" number. My idea is to test using the above procedure, which will either show PTY as being in or out with regards to being part of the problem.
> 

Ok, somehow we ended up on a wrong tangent  here. Let's forget ANY number mentioned in the past,. There is no 5k, no 512 bytes, just lost packets. 1 byte, 2 - 3 - 5 - 6 - 248 - 249 bytes, some times several packets in a row missing between entering ktcp and leaving ktcp. So wherever the loss happens, the unit is packet(s). 

At one point recently I was tempted to throw in the towel on this one, since the only scenario in which it happens is telnet inot elks, telnet out. Esoteric. Further, the issue is very timing sensitive. Enabling tcp or ip debugging on elks removes all symptoms. Also, there is no absolute repeatability, repeated tests show losses at different byte counts and lengths. Also, external clients with different speeds emit different patterns.
Anyway, I'm continuing - and there is no reason to make the issue a showstopper for lifting elks to the next level!

-M

ghaerr:
@Mellvik,

> rm -rf is broken for large recursive removes. I tried debugging and saw that it used a large buffer in a recursive routine, which immediately caused stack overflow. I fixed that in this PR, but it's still broken.

I rewrote `rm` such that it doesn't use a large stack buffer on directory recursion. Long story short, it still didn't fix the problem, which is that it hangs, just like your original version does, when removing lots of nested files. After a long debug session, I've finally determined that both versions hang in the kernel, waiting for a free buffer. This only occurs on the MINIX filesystem. We currently allocate 12 1024-byte "L1" buffers; the kernel "maps" a buffer for various operations and the combination of reading directories, truncating files, and removing directories overloads ELKS, which finally puts the process to sleep waiting for another buffer to free up, which won't ever happen. I'm going to leave both this kernel problem and the rm stack overflow to after v0.4.0, since we are already using 48K kernel data (mostly all buffers of various kinds), and only have 10k free heap. FYI, I didn't want to you go off and try fixing rm!!

Mellvik:
Thanks for the heads up, @ghaerr - appreciated.
And you're right, I might have done just that. :-)

Would it be reasonable to limit the # of recursions for now and emit an error message if exceeded, in order to make rm -r usable w/o risk of hanging the system?

--M

ghaerr:
> Would it be reasonable to limit the # of recursions for now and emit an error message if exceeded, in order to make rm -r usable w/o risk of hanging the system?

No, I thought of that, but don't know yet exactly how the directory structure in fact turns into increased buffer needs. Also, it would possibly leave the directories it it supposed to be erasing in a half-done state. For now, I think we're ok; no one has ever run into this, I just found it while testing cp -R. It'll get fixed after 0.4.0.

PR #878, opened 2020-11-21
[emu86] Update support for EMU86

ghaerr:
Updates ELKS tools/Makefile to use latest EMU86 version, for synchronization with v0.4.0.

For those that haven't looked at or tried EMU86 from @mfld-fr, it's pretty cool. EMU86 is an 8086 emulator, with specific target support for emulating ELKS' PC BIOS requirements, or other platforms, such as Advantach's SNMP-1000 SBC. This latest version runs the ROM code version of ELKS v0.4.0 right out of the box. You can single step through the startup code, with either AT&T or Intel format instruction display, or just let it run full speed, and use ^C to interrupt and display the current CS:IP being executed. Its a great way to get a feel for what is happening under the hood, although admittedly rather technical.

To use:
```
cp config-emu86 .config
make clean; make
./emu86.sh
[ELKS with a ROM filesystem will boot and run directly on your terminal screen]
```
Support for running non-ROM disk images of ELKS is being discussed, as well as BIOS INT 13h emulation, which will allow single-stepping through ELKS floppy boot code all the way to the login: prompt. I am looking forward to working on it.


Mellvik:
@ghaerr, 
does this mean it now works on macos?

-M

ghaerr:
> does this mean it now works on macos?

Yes, EMU86 does. That's not to be confused with elksemu, which runs ELKS binaries on Linux systems.

PR #881, opened 2020-11-25
[kernel][net][wd8003] Fixed receive ring-buffer wrap-around problem.

Mellvik:
Added code to handle receive ring buffer wrap-around.

### DIFF
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -5,6 +5,7 @@
  *
  */
 
+#include <stddef.h>
 #include <arch/io.h>
 #include <arch/ports.h>
 #include <arch/segment.h>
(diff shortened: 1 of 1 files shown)


ghaerr:
Thanks @Mellvik for all your help in getting the NIC driver fixed!!

Regarding the \#include <stddef.h>: why was that added, memory.h and others already include it. Was there a reason it needed to be added?

Mellvik:
Thanks @ghaerr,

And yes, there is a reason. Surprisingly, wd.c will not compile without this addition. Gcc complains about size_t not being defined, but used in memory.h.
It took me some time to get it right, and admittedly I cured the symptoms, not the root cause...

--Mellvik

ghaerr:
Hello @Mellvik,

As part of my review of the WD/SMC 8003 items remaining to be completed, I removed \#include <stddef.h> from wd.c and it compiled fine. Can you do that on your system and report the results? If it does not work, please report the gcc version using `ia16-elf-gcc --version`.

Thank you!

Mellvik:
@ghaerr,
this time it works fine at this end too.
I'm still at loss as to why it didn't work a week ago - anyway, I'll submit a PR with it removed.


Thank you.

—Mellvik

PR #883, opened 2020-11-27
[tools] [kernel] [setup] Allow far text kernel support to work with ROM kernel

tkchia:
As suggested at https://github.com/tkchia/gcc-ia16/issues/66#issuecomment-732300514 .

I added
  * new options `--aout-seg` and `--data-seg` to `elks/tools/elf2elks/elf2elks.c`, to allow `elf2elks` to produce a kernel which uses a far text but can still be burnt into a ROM;
  * changes to `elks/arch/i86/Makefile` to get it
    * to use the new `elf2elks` options when building a ROMable kernel, and
    * to properly calculate a checksum for an output ROM image; and
 * changes to `elks/arch/i86/boot/setup.S` to allow the initial data contents of the kernel data to lie outside the `a.out`'s first 64 KiB (this may be useful if the ROM kernel becomes bigger in the future).

With appropriate configuration settings, the resulting ROM image can now be run as an option ROM under QEMU:

One minor problem is that, if the kernel is built as a ROM kernel with far text, then `setup.S` no longer passes the correct far text size to the kernel.  `elf2elks` effectively combines the near text and far text segments into one single blob, in order to output an `a.out` with a short header.

### DIFF
--- a/elks/arch/i86/Makefile
+++ b/elks/arch/i86/Makefile
@@ -40,6 +40,21 @@ SIBOFLAGS	=
 endif
 
 #########################################################################
+# ROM specific options.
+
+ifeq ($(CONFIG_ROMCODE), y)
+
+# Make sure that the a.out header & data segment values are prefixed with `0x'
+ROMPOSTLINKFLAGS = --aout-seg 0x$(CONFIG_ROM_KERNEL_CODE:0x%=%) \
+		   --data-seg 0x$(CONFIG_ROM_KERNEL_DATA:0x%=%)
+
+else
+
+ROMPOSTLINKFLAGS =
+
+endif
+
+#########################################################################
 # Objects to be compiled.
 
 AARCHIVES = kernel/akernel.a lib/lib86.a mm/mm.a
@@ -103,7 +118,7 @@ boot/system: $(XINCLUDE) $(AARCHIVES) $(ADRIVERS) sibo/crt1.o sibo/crt0.o
 		$(ARCH_DIR)/sibo/crt0.o $(ARCH_DIR)/sibo/crt1.o \
 		init/main.o $(ARCHIVES) $(DRIVERS) $(LIBGCC) \
 		-o $(ARCH_DIR)/boot/system > $(ARCH_DIR)/boot/system.tmp ; \
-		$(POSTLINK) $(ARCH_DIR)/boot/system ; \
+		$(POSTLINK) $(ROMPOSTLINKFLAGS) $(ARCH_DIR)/boot/system ; \
 		sort -k3,4 $(ARCH_DIR)/boot/system.tmp > $(ARCH_DIR)/boot/system.map ; \
 		rm -f $(ARCH_DIR)/boot/system.tmp )
 
@@ -121,7 +136,7 @@ boot/system:	$(XINCLUDE) $(AARCHIVES) $(ADRIVERS) boot/crt0.o
 		$(ARCH_DIR)/boot/crt0.o \
 		init/main.o '-(' $(ARCHIVES) $(DRIVERS) '-)' $(LIBGCC) \
 		-o $(ARCH_DIR)/boot/system > $(ARCH_DIR)/boot/system.map ; \
-		$(POSTLINK) $(ARCH_DIR)/boot/system)
+		$(POSTLINK) $(ROMPOSTLINKFLAGS) $(ARCH_DIR)/boot/system)
 #		sort -k3,4 $(ARCH_DIR)/boot/system.tmp > $(ARCH_DIR)/boot/system.map ; \
 #		rm -f $(ARCH_DIR)/boot/system.tmp )
 
@@ -139,12 +154,8 @@ ifeq ($(CONFIG_ROM_BOOTABLE_BY_RESET), y)
     RVECT = -r $(CONFIG_ROM_RESET_ADDRESS)   
 endif
 
-# Disable temporary the checksum to not overwrite the image
-# as it is actually not needed for execution in EMU86
-
 Image:	toolkit boot/setup boot/system  
-#	tools/mkbootloader $(RVECT) -c $(CONFIG_ROM_SETUP_CODE) $(CONFIG_ROM_CHECKSUM_SIZE) boot/Image $(CONFIG_ROM_BASE) -a boot/setup $(CONFIG_ROM_SETUP_CODE) -s boot/system $(CONFIG_ROM_KERNEL_CODE) $(CONFIG_ROM_BIOS_MODULE) $(CONFIG_ROM_BIOS_MODULE_ADDR)
-	tools/mkbootloader $(RVECT) boot/Image $(CONFIG_ROM_BASE) -a boot/setup $(CONFIG_ROM_SETUP_CODE) -s boot/system $(CONFIG_ROM_KERNEL_CODE) $(CONFIG_ROM_BIOS_MODULE) $(CONFIG_ROM_BIOS_MODULE_ADDR)
+	tools/mkbootloader $(RVECT) -c $(CONFIG_ROM_SETUP_CODE) $(CONFIG_ROM_CHECKSUM_SIZE) boot/Image $(CONFIG_ROM_BASE) -a boot/setup $(CONFIG_ROM_SETUP_CODE) -s boot/system $(CONFIG_ROM_KERNEL_CODE) $(CONFIG_ROM_BIOS_MODULE) $(CONFIG_ROM_BIOS_MODULE_ADDR)
 	$(CONFIG_ROM_SIMULATOR_PROGRAM)
 
 # End PC image build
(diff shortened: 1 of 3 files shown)


ghaerr:
Hello @tkchia,

Thanks for doing this!

Have you tried running the kernel under EMU86? Prior to v0.4.0, I made some enhancements that allow ELKS ROM kernels to be fully tested with it. This can be done using the following lines (also found in emu86.sh):
```
cp emu86-rom.config .config
make clean; make
emu86 -w 0xe0000 -f elks/arch/i86/boot/Image -w 0x80000 -f image/romfs.bin -x 0xe000:0x14 ${1+"$@"}
```
Using your proposed commit, this executes fine (and will boot to a standalone shell prompt, unlike QEMU).

However, the "standard" romcode build (emu86-rom.config) does not set CONFIG_FARTEXT_KERNEL=y, so actually booting a ROM fartext kernel is not tested. When I set CONFIG_FARTEXT_KERNEL=y, recompile everything and run EMU86, the ROM boot fails on emu86 as follows:
```
info: load address E0000h
info: load file elks/arch/i86/boot/Image
info: file size=87D0h
success: file loaded
info: load address 80000h
info: load file image/romfs.bin
info: file size=2C0D9h
success: file loaded
info: execute address E000h:0014h

ELKS Setup 
ERROR: kernel size > 64k
```
(type ^C to get to the EMU86 prompt, then q to quit).

Thus, is seems that something is still amiss with regards to setup.S being able to handle the a.out header produced when a .fartext section is present.

Also, can you explain a bit about what the ROM checksum and ROM byte 3 (space for lengthcode (promsize / 512) is supposed to do? I've never fully understood it. Is that for a certain type of ROM boot loader?


Thank you!

toncho11:
Hi,

What is kernel ROM image anyway?
Can it be put into a real ROM chip of IBM 5160 for example? If could replace ROM Basic with ELKS that would be useful.

tkchia:
Hello @toncho11,

> What is kernel ROM image anyway?
> Can it be put into a real ROM chip of IBM 5160 for example?

Well, hopefully it will be possible to do that soon...

tkchia:
Hello @ghaerr,

> Also, can you explain a bit about what the ROM checksum and ROM byte 3 (space for lengthcode (promsize / 512) is supposed to do? I've never fully understood it. Is that for a certain type of ROM boot loader?

My understanding is that the ROM-ified ELKS is meant to be burnt onto an option ROM, to be physically installed alongside a main ROM BIOS --- one which is at least partly compatible with the IBM PC BIOS.

The [_Technical Reference_ for the IBM PC AT 5170](https://archive.org/details/IBMPCATIBM5170TechnicalReference1502243MAR84/) has this to say (pp. 5-11 and 5-12):

> The ROM BIOS provides a way to integrate adapters with on-board ROM code into the system. ... The absolute addresses hex C8000 through E0000 are scanned in 2K blocks in search of a valid adapter ROM.  A valid ROM is defined as follows:
> 
> | | |
> | --- | --- |
> | **Byte 0** | Hex 55 |
> | **Byte 1** | Hex AA
> | **Byte 2** | A length indicator representing the number of 512-byte blocks in the ROM.
> | **Byte 3** | Entry via a CALL FAR
>
> A checksum is also done to test the integrity of the ROM module.  Each byte in the defined ROM module is summed modulo hex 100.  This sum must be 0 for the module to be valid.
>
> When the POST [power-on self-test] identifies a valid ROM, it does a far call to byte 3 of the ROM, ...
>
> The POST [also] provides a way to integrate additional ROM modules' code into the system. ... The absolute addresses hex E0000 through EFFFF are scanned in a 64K block in search of a valid checksum.  Valid ROM is defined as follows:
>
> | | |
> | --- | --- |
> | **Byte 0** | Hex 55 |
> | **Byte 1** | Hex AA
> | **Byte 2** | Not used
> | **Byte 3** | Entry via a CALL FAR

tkchia:
Hello @ghaerr,

> However, the "standard" romcode build (emu86-rom.config) does not set CONFIG_FARTEXT_KERNEL=y, so actually booting a ROM fartext kernel is not tested. When I set CONFIG_FARTEXT_KERNEL=y, recompile everything and run EMU86, the ROM boot fails on emu86 as follows:

Did you also include my [second commit](https://github.com/jbruchon/elks/pull/883/commits/b468886c3fe31a5450d4b8fbf1c359dbcd2d54fc) from this PR?  Without this patch, the `elks/arch/i86/Makefile` will not supply the correct options to `elf2elks` to fashion a ROMable kernel with a short `a.out` header.

(The patch also happens to pad the resulting ROM image to 64 KiB (`0x10000` bytes) to add a checksum byte.  `emu86` does not need the checksum to be correct, but QEMU's SeaBIOS does.)

Thank you!

ghaerr:
Hello @tkchia,

> Without this patch, the elks/arch/i86/Makefile will not supply the correct options to elf2elks to fashion a ROMable kernel with a short a.out header.

Sorry, my error on quickly testing. Everything works very well :)

With this PR, ROM fartext kernels are now supported! However, should the overall fartext kernel be > 64K, then the setup.S code checking the hi word of the text size will fail the boot. What do you think about that? Should that check be removed, and allow a larger than 64K text blob to run? (Does elf2elks support that?) The original reason for the fartext support was that we are running out of space in a single code segment, so this would be useful. I haven't tried it yet, but the regular ELKS fartext kernel should be able to be built with CONFIG_ROMCODE and run in emu86, now that BIOS INT 13h support has been added to emu86.

When compiling elf2elks.c, the following warning is generated; can that be silenced?
```
elf2elks.c:829:6: warning: variable 'value' is used uninitialized whenever switch default
      is taken [-Wsometimes-uninitialized]
            default:
            ^~~~~~~
elf2elks.c:834:35: note: uninitialized use occurs here
          buf[offset_in_scn] = (uint8_t) value;
                                         ^~~~~
elf2elks.c:816:18: note: initialize the variable 'value' to silence this warning
          uint16_t value, offset_in_scn;
                        ^
                         = 0
```

Thanks for the explanations on the IBM option ROM checksum and zero-fill, I didn't realize ELKS supported (and now correctly again supports) option ROMs.

Thank you!

tkchia:
Hello @ghaerr,

> With this PR, ROM fartext kernels are now supported! However, should the overall fartext kernel be > 64K, then the setup.S code checking the hi word of the text size will fail the boot. What do you think about that? Should that check be removed, and allow a larger than 64K text blob to run? (Does elf2elks support that?)

Yes, `elf2elks` supports having [near text] + [far text] > 64 KiB, and it should be possible to tweak `setup.S` to support this too.  At the moment though, I am not too sure about the best way to handle such a situation.  In particular, besides the `setup.S` check, the `emu86-rom.config` also currently uses 64 KiB as the ROM size for computing the checksum.  There are probably a few extra things that need to be changed to truly support large kernels, if at all.

@mfld-fr : may I know if you think a ROM ELKS kernel larger than 64 KiB would make sense?  What might be a reasonable way to lay out such a thing in memory, on an actual embedded system?

> When compiling elf2elks.c, the following warning is generated; can that be silenced?

I will fix this.  Thanks!

ghaerr:
Hello @tkchia,

> elf2elks supports having [near text] + [far text] > 64 KiB, and it should be possible to tweak setup.S to support this too.

Since `elf2elks` supports > 64KiB blobs now, and there may be some issues regarding exactly how IBM option ROMs should work, it would be fine by me to commit this PR now, and deal separately with those issues.

>  At the moment though, I am not too sure about the best way to handle such a situation. In particular, besides the setup.S check, the emu86-rom.config also currently uses 64 KiB as the ROM size for computing the checksum. 

Also, reading the IBM PC AT technical manual, it seems the E0000-EFFFFh ROM checksum is placed in the last byte of the ROM image, which could be problematic for > 64KiB ROMs. I'm not sure where the ROM checksum for C8000-DFFFFh ROM images is placed, is that what the Byte 2 'length indicator' is used for?

I think the "large" ELKS ROM support would probably be most useful for embedded systems, rather than IBM PC ROM use.

> There are probably a few extra things that need to be changed to truly support large kernels, if at all.

Using this PR as-is, I've got the full "disk-boot" ELKS fartext kernel compiled up and running as a ROM image on emu86. The full system is currently at 59KiB, so we've still got room before worrying about > 64KiB a.out header issues. The full system (for testing) includes the Minix filesystem running on BIOS INT 13h as well as the Rom filesystem at 8000:0000h. There are some issues to sort out, including where in memory various parts of the system should reside, which filesystem to boot from, etc. I'll submit a separate PR for that, with a couple scenarios for "large" and "small" ROM systems.

Thank you for this enhancement!

ghaerr (approved):

mfld-fr:
@tkchia :
> @mfld-fr : may I know if you think a ROM ELKS kernel larger than 64 KiB would make sense? What might be a reasonable way to lay out such a thing in memory, on an actual embedded system?

My opinion is biased because I am working with only one embedded system with its specific design, so I am afraid I could not answer that question for any IA16 embeded system in general.

mfld-fr:
Next version of EMU86 will detect any BIOS extension as specified in @tkchia post above, to avoid the `-x 0xe000:*` argument on the command line in `emu86.sh`.

Issue #763, opened 2020-09-26
GCC 10 fixes in upstream binutils-ia16


xenoxaos:
There's issues with binutils-ia16. I've already created a PR with the fixes in it:
https://github.com/tkchia/binutils-ia16/pull/8

If I remember, I'll create a PR with an updated tools/Makefile to point to the commit, but I might forget once it's merged. So, creating an issue for one of us to check it out

tkchia:
Hello @ghaerr,

In sum, it seems there were some errors when building `binutils-ia16` with GCC 10.  (I did not encounter them as I was building with GCC 9.)  Do you think you would like to bump up the `binutils-ia16` version to use, to address this?  Thanks!

ghaerr:
Hello @tkchia,

Considering that Mike required it for his build, you might as well bump the version number as well in your fix for it in #764, along with a message that a rebuild is not required and the fix only applies if using GCC 10. For our users that have not yet rebuilt their toolchain for the new ld.gold linker support, they'll get the all the fixes.

Thank you!

tkchia:
Hello @ghaerr: OK, I have updated the PR.  Thanks!

ghaerr:
Hello @tkchia,

Is this issue finalized?

I'd like to close it unless there's a reason to keep it open across v0.4.0.

Thank you!

tkchia:
Hello @ghaerr,

> Is this issue finalized?

Not quite --- if I remember correctly, the `binutils-ia16` version used in ELKS has not yet been bumped up to a newer version.  For now though, this only affects people trying to build ELKS and the toolchain under newer versions of host GCC.

Thank you!

tkchia:
With the PR https://github.com/jbruchon/elks/pull/884 , this problem should be fixed now.  Thanks!

PR #884, opened 2020-11-28
[tools] Bump up binutils-ia16 & gcc-ia16 versions

tkchia:
This should fix some compilation errors that happened when building `binutils-ia16` with host GCC 10 (https://github.com/jbruchon/elks/issues/763), &amp; also fix some inefficiencies &amp; corner case bugs in `gcc-ia16`.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -23,7 +23,7 @@ BUILDDIR=$(CROSSDIR)/build
 
 # BINUTILS for IA16
 
-BINUTILS_VER=f19a905d1434da03940951913514a9ef12f0d7dd
+BINUTILS_VER=164dc55d3d9488a487d39c2e7f3f8cadf6dc12f5
 BINUTILS_DIST=binutils-ia16-$(BINUTILS_VER)
 
 $(DISTDIR)/$(BINUTILS_DIST).tar.gz:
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

Should we all update our toolchain now with this, or is this more for maintenance, keeping your changes updated in ELKS (apart from GCC 10 issues)?

Thank you!

tkchia:
Hello @ghaerr,

This is more for maintenance --- there should be no urgent need to upgrade the toolchain for now.

(The corner case bugs involve

  * far functions which are defined weak or defined with `extern inline`, and
  * tests for the `__FAR` or `__SEG_SS` macros in C++ code.

These are not currently used in ELKS.)

Thank you!

ghaerr:
Hello @tkchia,

Can you look at the check failed on my last PR, #887? The build has been failing since I committed your (this) PR #884. For some reason, GitHub didn't say anything about it until I pushed another PR.

Thank you!

tkchia:
Hello @ghaerr,

Thanks for the report!

It seems that the issue only affects the building of the kernel --- specifically, some of the special [`segelf`](https://git.zytor.com/users/hpa/segelf/abi.git/plain/segelf.txt) symbols turn out to be undefined in the assembly language modules in the kernel.  This is most probably because the assembly modules were not built with `segelf` enabled.

I am not sure why this problem is only surfaced now, but it is definitely a problem, and an (earlier) oversight on my part.

I am now trying to see if modifying
```
ASFLAGS = $(CPU_AS) $(ARCH_AS)
```
to
```
ASFLAGS = $(CPU_AS) $(ARCH_AS) --32-segelf
```
in `elks/Makefile-rules` will solve the problem.

Thank you!

PR #887, opened 2020-11-30
[cmds] Fix cp command

### LINKED ISSUE #886: Failed install with sys
The sys command does not complete the installation: no files are copied to /bin and /etc folders. 
The cp command does not give a reason for the failure or provides an error code - nothing. 
It happens at the end of the copy process, because it takes time (proportional to the file I think) before it fails. 
The system can boot after that even without the files in /bin with some errors.
Both fat16 and fat32 are affected.
There are no other error messages except for cp.
Partition was created with fdisk and formatted with mkfat.
Problem encountered several times.

ghaerr:
Fixes #886. 

So much for a perfect v0.4.0. Looks like my `cp -R` enhancement broke cp without -R.

### DIFF
--- a/elkscmd/file_utils/cp.c
+++ b/elkscmd/file_utils/cp.c
@@ -589,10 +589,7 @@ int copyfile(char *srcname, char *destname, int setmodes)
 	}
 
 	close(rfd);
-	if (close(wfd) < 0) {
-		perror(destname);
-		return 1;
-	}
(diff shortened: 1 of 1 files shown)


tkchia:
Hello @ghaerr,

Perhaps you can try rebasing this on top of my PR https://github.com/jbruchon/elks/pull/888 ?  Thank you!

PR #889, opened 2020-12-01
[cmds] Enhance mv to support renaming symlinks

### LINKED ISSUE #882: Mv doesn't support renaming symlinks
I believe this is a recently introduced 'anomaly' to `ls`: 

If listing the current directory, symbolic links show up as their target, if in a different directory (i.e. there is a path), the display is fine.

```
# cd /var/www
# ln -s /dev/hda hdimage
# ls -l 
brw-r--r--  2 root     root       3,   0  Nov 26 09:57 hdimage
-rwxr-xr-x  1 root     root          275  Nov 26 09:57 index.html
# mkdir x
# ln -s /dev/fd1 x/fd
# ls -l x
lrwxrwxrwx  1 root     root            8  Nov 26 14:00 fd -> /dev/fd1
# mv x/fd .
# ls -l 
brw-r--r--  2 root     root       3, 160  Nov 26 09:57 fd
brw-r--r--  2 root     root       3,   0  Nov 26 09:57 hdimage
-rwxr-xr-x  1 root     root          275  Nov 26 09:57 index.html
drwxrwxrwx  2 root     root           48  Nov 26 14:02 x
# 
```

ghaerr:
Fixes #882.

Also updates `access` to use F_OK as suggested by @toncho11.

### DIFF
--- a/elkscmd/file_utils/mv.c
+++ b/elkscmd/file_utils/mv.c
@@ -192,10 +192,7 @@ int copyfile(srcname, destname, setmodes)
 	}
 
 	close(rfd);
-	if (close(wfd) < 0) {
-		perror(destname);
-		return 0;
-	}
+	close(wfd);
 
 	if (setmodes) {
 		(void) chmod(destname, statbuf1.st_mode);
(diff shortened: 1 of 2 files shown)


toncho11:
Is this code in copyfile() in mv.c

```
if (close(wfd) < 0) {
		perror(destname);
		return 0;
	}
```
problematic? Same code was corrected in cp.c in #887 
In general it is a good idea to check if fclose < 0 , but then this seems to be a problem in cp.c.
And it is return 0, instead of return 1.

ghaerr:
Hello @toncho11,

Thanks for the code review.

> In general it is a good idea to check if fclose < 0 , but then this seems to be a problem in cp.c.

Yes, with `fclose` there may be buffered contents so a final write error could be returned at close.
`close` on the other hand will never return a write error, only an error on bad file descriptor.

> And it is return 0, instead of return 1.

Yes, this is confusing, as the routines (although documented) returns opposite error codes. I've changed it.

Thank you!

PR #890, opened 2020-12-02
[kernel] Allow legacy v0 a.out programs to run with minimal stack

ghaerr:
Fixes kernel to be able to run legacy ELKS (v0 header) programs that allocated less memory than INIT_STACK bytes for stack.
Runs programs produced by [Alfonso Marton's](http://www.alfonsomartone.itb.it/fhlvnr.html) `exe2elks` Turbo-C DOS to ELKS conversion program.

Incompatibility found by @toncho11 in https://github.com/jbruchon/elks/issues/871#issuecomment-729857520.

### DIFF
--- a/elks/arch/i86/kernel/process.c
+++ b/elks/arch/i86/kernel/process.c
@@ -66,6 +66,7 @@ void stack_check(void)
     {
 	/* optional: check stack over min stack*/
 	if (currentp->t_regs.sp < currentp->t_begstack - currentp->t_minstack) {
+	  if (currentp->t_minstack)	/* display if protected stack*/
 	    printk("(%d)STACK OVER MINSTACK by %u BYTES\n", currentp->pid,
 		currentp->t_begstack - currentp->t_minstack - currentp->t_regs.sp);
 	}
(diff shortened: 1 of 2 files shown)


tkchia:
Hello @ghaerr,

We should probably check that `heap` > `slen` (or even `heap` ≥ `slen` + [some minimal value]) before calculating `stack = heap - slen;` ?  Thanks!

ghaerr:
Hello @tkchia,

Thank you for your comment. I've realized that, for v0 header programs with nonzero chmem, calculating any arbitrary "stack" amount in exec.c is incorrect. In v1 program execution, the a.out header stack value (called minstack) specifies a protected stack size (or default of INIT_STACK) for the process. This value is saved and SP checked at each system call entry by stack_check, whose purpose is to inform the user should the stack overflow its allowed size (possibly into the allocated heap). In v0 program execution, and the ELKS kernel prior to v1 headers, the stack and heap grow towards each other and there is no stack overflow checking. For the kernel to claim an arbitrary protected stack area would be to decrease possibly required heap space set by the original program author, when chmem is specified (nonzero).

I think the best solution to this is to treat v0 header programs with nonzero chmem as having zero protected stack area, and let the program run with its requested total data space. This will allow v0 programs to use its requested heap or stack, although with no checking (heap overflow is still checked). This will require a small change in stack_check. I will update this PR after some testing.

Thank you!

tkchia (approved):

tkchia:
Hello @ghaerr,

Thank you!  Please merge once you are ready.

toncho11:
Yes, it does work with Alfonso's `hello` program.

PR #891, opened 2020-12-03
[net] Removed extraneous include in wd.c

Mellvik:
Extra #include <stddefs.h> removed.


ghaerr:
As part of this fix, I was going to update the description of the WD/SMC card to "WD/SMC 8003" in config.in as you suggested. To avoid two PR's why don't you add that also as part of this PR, thanks.

Mellvik:
Good Idea, @ghaerr,

Will do that.

—M

PR #895, opened 2020-12-07
[kernel] Rewrite wait system call, fixes cron

### LINKED ISSUE #892: cron
Is cron working?

I editted /var/cron/root with this line with kilo:
`*/1 * * * * echo "fsdfsdfsdsd" >> /root/cron.log`

but I do not see a message every minute.

ghaerr:
Fixes #892.

There was a race condition between reaping process exit status and re-using unused task structures, which caused `cron` to fail. `sys_wait4` was rewritten to eliminate this problem. Handling of orphans is now performed at process exit time, rather than in wait, which fixes the race problem. Also previously, orphaned zombies were reparented to init, for reaping exit status. Now, the task structure entry is marked unused, for better use of the limited task structure during fast task creation periods.

Fixed a number of kernel problems where the `for_each_task` define was used to iterate through all tasks; it is an error to check task structure fields without examining whether the task state is TASK_UNUSED or not, as values are never cleared. 

Removed unused task "family" fields: p_child, p_nextsib, p_prevsib.

Finally, testing on QEMU has proved to be a real problem: for some reason, it seems QEMU limits process usage and greatly slows executing instructions, including the timer interrupt, after several minutes, or until console I/O. This causes the `sleep` wait in cron to linger well past its requested real time and thus cron jobs don't execute when they should. I've added a "TEST" define in cron's config.h that shows this happening with debug output of jobs. Thus, cron's timing itself of job execution has not been tested. I plan on further investigation of QEMU's instruction throttling when I have time.

### DIFF
--- a/Documentation/html/user/cron.html
+++ b/Documentation/html/user/cron.html
@@ -47,8 +47,8 @@
 			run ELKS on qemu, the hardware time will be UTC or Greenich mean
 			time.<BR>Therefore the jobs will not be executed at the local time
 			if you do not set<BR>the ELKS time to your local time. If you
-			schedule a lot of jobs you may<BR>run out of slots. Increase
-			POLL_MAX in the limits.h file then.<BR><BR>To display the contents
+			schedule a lot of jobs you may<BR>run out of task slots.
+			<BR><BR>To display the contents
 			of the crontab file call the crontab program<BR>with the &quot;-l&quot;
 			parameter:<BR><BR><FONT FACE="Courier New, monospace">$ crontab
 			-l</FONT><BR><BR>To edit the cron jobs, do:<BR><BR><FONT FACE="Courier New, monospace">$
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -486,7 +486,9 @@ int sys_execve(char *filename, char *sptr, size_t slen)
      */
     arch_setup_user_stack(currentp, (word_t) mh.entry);
 
+#if 0	/* used only for vfork()*/
     wake_up(&currentp->p_parent->child_wait);
+#endif
 
     /* Done */
 
--- a/elks/include/linuxmt/sched.h
+++ b/elks/include/linuxmt/sched.h
@@ -71,9 +71,11 @@ struct task_struct {
     pid_t			pgrp;
     struct tty			*tty;
     struct task_struct		*p_parent;
+#if BLOAT
     struct task_struct		*p_prevsib;
     struct task_struct		*p_nextsib;
     struct task_struct		*p_child;
+#endif
     struct wait_queue		child_wait;
     int				exit_status;	/* process exit status*/
     struct inode		*t_inode;
--- a/elks/kernel/exit.c
+++ b/elks/kernel/exit.c
@@ -12,34 +12,43 @@
 extern int task_slots_unused;
 extern struct task_struct *next_task_slot;
 
-int sys_wait4(pid_t pid, int *status, int options)
+static void reparent_children(void)
 {
 	register struct task_struct *p;
-	struct task_struct *q;
-	int waitagain, orphans;
-
-	debug_wait("WAIT(%d) for %d %s\n", current->pid, pid, (options & WNOHANG)? "nohang": "");
-
- do {
-	waitagain = 0;
-
-	/* reparent orphan zombies to init*/
-	orphans = 0;
-	for_each_task(p) {
-		if (p->state == TASK_ZOMBIE && p->p_parent) {
-			debug_wait("Zombie pid %d ppid %d\n", p->pid, p->p_parent->pid);
-			if (p->p_parent->state == TASK_UNUSED) {
-				debug_wait("WAIT(%d) reparenting %d to 1\n", current->pid, p->pid);
-				p->p_parent = &task[1];
-				orphans++;
-			}
-		}
-	}
-	if (orphans)
-		wake_up(&task[1].child_wait);
 
 	for_each_task(p) {
 		if (p->p_parent == current) {
+			/* remove orphaned zombies, no need to reparent them to init*/
+			if (p->state == TASK_ZOMBIE) {
+				debug_wait("Zombie orphan pid %d ppid %d removed\n", p->pid, p->ppid);
+				p->state = TASK_UNUSED;		/* unassign task entry*/
+				next_task_slot = p;
+				task_slots_unused++;
+			}
+
+			/* reparent orphans to init*/
+			if (p->state != TASK_UNUSED) {
+				debug_wait("Reparenting orphan pid %d ppid %d to init\n",
+					p->pid, p->p_parent->pid);
+				p->p_parent = &task[1];
+				p->ppid = task[1].pid;
+			}
+		}
+	}
+}
+
+int sys_wait4(pid_t pid, int *status, int options)
+{
+	register struct task_struct *p;
+	int waitagain;
+
+	debug_wait("WAIT(%d) for %d %s\n", current->pid, pid, (options & WNOHANG)? "nohang": "");
+
+ for (;;) {
+	waitagain = 0;
+
+	for_each_task(p) {
+		if (p->p_parent == current && p->state != TASK_UNUSED) {
 		  if (p->state == TASK_ZOMBIE || p->state == TASK_STOPPED) {
 			if (pid == -1 || p->pid == pid || (!pid && p->pgrp == current->pgrp)) {
 				if (status) {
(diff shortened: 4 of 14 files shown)


Mellvik:
This is good news, @ghaerr. Thank you!

I just wanted to comment not the QEMU hitting molasses issue: 

I've seen this (or something similar) quite frequently recently. While not investigating it, I've noticed that it happens predictably when using serial console (attaching to the windows which runs the qemu process) and still keeping a graphic 'terminal' (PC screen) open.
It does not happen immediately at boot, but just a few commands (on the serial console) and it's there. Also - if I remember correctly - the response seems to be OK in the graphic window.

Let me know if this is useful and I'll see if I can make a repeatable case.

—Mellvik

ghaerr:
>  the response seems to be OK in the graphic window.
> Let me know if this is useful and I'll see if I can make a repeatable case.

Yes, I very much would like to learn more about this. Interesting that it may work properly in the graphical window, I hadn't tried that.

If you set TEST=1 in cron's config.h, this can be very easily identified.

Thank you!

Mellvik:
> Yes, I very much would like to learn more about this. Interesting that it may work properly in the graphical window, I hadn't tried that.
> 
I'll have a closer look in the morning.
> If you set TEST=1 in cron's config.h, this can be very easily identified.
> 
Is that a #define ? I could not find 'TEST' used anywhere in the cron directory?

-M

ghaerr:
> Is that a #define ? I could not find 'TEST' used anywhere in the cron directory?

Have you fetched this PR? It is on line 8 of config.g in the cron source dir.

ghaerr:
@Mellvik,

> I've seen this (or something similar) quite frequently recently. While not investigating it, I've noticed that it happens predictably when using serial console (attaching to the windows which runs the qemu process) and still keeping a graphic 'terminal' (PC screen) open.
It does not happen immediately at boot, but just a few commands (on the serial console) and it's there. Also - if I remember correctly - the response seems to be OK in the graphic window.

I have confirmed your observations about QEMU: when running from the graphics console only, things operate just fine. When operating from a logged in serial port (terminal), QEMU is throttled back to the point of one timer interrupt per second or so (instead of 100/second), after a minute or so of operation, every time. Since the timer interrupt occurs 100 times more slowly than it should, this obviously screws up cron scheduling.

I'm not yet sure if this is a "feature" of QEMU, macOS throttling back the system when too many system calls are made, possibly exceeding a system ulimit (for accessing the redirected serial console), or something to do with QEMU's emulation of the serial port hardware.

I have also just confirmed that cron is working correctly, at least for the once/minute and once every 10 minutes default crontab operations, when run from graphic console. (This is with TEST=0. When TEST=1, a cron job is executed every two seconds, which I used for my testing. This also allows one to easily see QEMU stop executing quickly after a minute or so).

Thank you!

Mellvik:
Thank you @ghaerr,
As far as cron is concerned - it does indeed works as expected.

The QEMU issue is more of a mystery. We're seeing the same, and it does not seem to be related to MacOS.
While keeping cron (TEST=1) active on the serial console, I did a telnet from the graphics console to the mac and started a ls -LR, which runs for like 10 minutes (don't ask …). 
While the telnet ls is running full tilt, the cron messages suddenly started coming slower (like 1 per 10 seconds), no change in the ls listing speed. Further - a few minutes later, the cron messages are back to the 2 second interval schedule. If I continue, the slow tick comes and goes (few minute intervals) while the telnet ls keeps running.

Beats me, but it may seem we have a problem in elks that is somehow affecting clock ticks. The fact that it seems to appear only when we have a serial console, may be an indication as to where to look, although I have no suggestions.

Talking about serial console - I've noticed that when all lines in /bootopts are commented out, there are no boot messages on the graphics console. I was assuming that would be the default?

—Mellvik

Mellvik:
@ghaerr,

yes, cron works as far as I can tell - right now I became more interested in the erratic clock behaviour.
I've kept it running in QEMU for hours, and notice that the frequency of messages from cron are indeed erratic. Every other second for a short while, then silence, not necessarily in a pattern.

This is what I've been noticing before: Some times in the serial console window I can hit enter or write a few chars, and there is no response until  maybe 4 or 10 seconds later.

My guess is that this is a QEMU issue - I'm moving the testing to real hardware to verify.
BTW - I just noticed - cron is a quite large process(38k), and I am getting sh: out of memory messages for periods.
Also, FWIW cron reports varying values in 'sleep return', usually 2 or 0, some times 1 occasionally -1.

—Mellvik

toncho11:
That is why I am testing on VirtualBox. It looks like there is no problem on VB if you do not put a cap on the CPU.
So for now it looks like a QEMU issue because it works on both VB and my Amstrad 1640. 
Still further tests are required.

ghaerr:
> Talking about serial console - I've noticed that when all lines in /bootopts are commented out, there are no boot messages on the graphics console. I was assuming that would be the default?

By default, all lines are commented out in /bootopts. Are you getting no kernel startup messages on the graphics console? This would happen if CONFIG_SERIAL_CONSOLE is set.

> Also, FWIW cron reports varying values in 'sleep return', usually 2 or 0, some times 1 occasionally -1.

Sleep returns the number of remaining seconds. Sleep can return early, if interrupted by a signal, in which case -1 would be returned rather than 0, if no time was left. Cron uses signals to catch and reap the child processes forked for each line in crontab.

toncho11:
@ghaerr Please make the temporary file created by `crontab -e` shorter (8.3). Otherwise it does not work on FAT. Maybe you can just remove the prefix ".cron.".

ghaerr:
@toncho11,

> Please make the temporary file created by crontab -e shorter (8.3). Otherwise it does not work on FAT. 

I pushed a fix for this, it should work now on FAT, please let me know, thanks.

> (cross-post reponse from https://github.com/jbruchon/elks/issues/892#issuecomment-742003776):
> So I need to use crontab -e because it takes care of reloading the crontab file.

Yes, I think using crontab -e is preferable since it handles dealing with cron's internal issues.

> If I use kilo and edit it manually I will need to restart cron myself. 

If you set EDITOR=/bin/kilo in your environment, crontab -e should run kilo instead of vi.


> Also manual editing might lead to cron multiplying processes?
Is that correct?

Not necessarily, except that cron could be in the middle of reading the file when you're editing it. That's what crontab -e is supposed to be for.

toncho11:
Hi,

I am getting this error after boot and executing `crontab -e`: 
ELKS_out_of_memory
on both QEMU and VB. 
The error appears after exiting vi.

ghaerr:
@toncho11,

Sadly, it appears that cron may be requiring more memory that already specified in it's Makefile, using the -maout-heap=0x3800, which is 14k. Looking at the code, it seems that cron allocates memory for the job environment, as well as each line in the /var/cron/root cron file. I also can't find where it frees that memory, so replacing the cron file with crontab -e could result in cron requiring double the memory otherwise required.

Depending on what is in your cron file, cron may need more memory. If you can build cron, change the -maout-heap= value to higher (by at least 4096 bytes, and use decimal rather than hex for ease of understanding).

After your test results, we can determine what we should set it at. Keep in mind that when cron forks, it copies all its data segment, and it forks twice per crontab line (explained in detail https://github.com/jbruchon/elks/issues/892#issuecomment-741424813).

This entire exercise we're going through is a good example of how porting a "regular" Linux command like cron doesn't work well on ELKS - one must completely understand how the program works, and the resources needed, when picking which source to use to port each program. ELKS is very resource constrained, and this version of cron doesn't appear to be too concerned about that.

Thanks!

toncho11:
@ghaerr,

I am missing something.  `cron` was not running when I issue `crontab -e`. I do not see where `crontab` is starting/forking `cron`. For me `cron` is not working at this point, because I did not start it yet. For me the memory error is rather coming from `crontab`, not `cron`?

ghaerr:
> For me the memory error is rather coming from crontab, not cron?

In that case, add the -maout-heap=8192 to crontab's Makefile line.

toncho11:
It did not help. 
I did try even with  -maout-heap=32000
Same problem if editor is kilo. But 32000 is 32 kb, so it should have been enough ?

ghaerr:
> It did not help.
I did try even with -maout-heap=32000

Strange, as I set the following line and it works (using 8192 did not, it proved too small a heap):
```
 crontab: crontab.o $(common)
       $(CC) $(CFLAGS) -maout-heap=14336 -o $@ crontab.o $(common)
```

I've pushed a commit with this fix for you.

ghaerr:
> I do not see where crontab is starting/forking cron.

Agree, it seems it does not. Thus, IMO, crontab -e and -r are somewhat useless: neither interfaces with cron. It seems all that crontab does is mess around with various permissions for the /var/cron/root (and other users) cron files, which is pretty much not needed on ELKS. We might be better off just removing crontab, except that it allows the location of the crontab file  to be unknown to the user.

Mellvik:
FWIW - I support the idea of removing crontab.
BTW, crontab was never intended to start (or check on) cron, just an easy way for non-root users to run cron jobs (ie create crontab entries). As such it does fine, but its usefulness on elks is indeed questionable.

-m

toncho11:
Or crontab can be reduced to only creating the directories and example entries (commented). And this simplified version should not crash as the current on exit.

ghaerr:
> BTW, crontab was never intended to start (or check on) cron, just an easy way for non-root users to run cron jobs (ie create crontab entries). As such it does fine, but its usefulness on elks is indeed questionable.

Thanks, I didn't realize that, but that now makes sense. I would say there isn't much need for a seperate program on ELKS to handle cron permissions, and confuse users, when we're already very tight on disk space on the 1.44M floppy,.

> Or crontab can be reduced to only creating the directories and example entries (commented). 

No need to add that to crontab, as cron already does that on startup. The example entries can get commented out once this PR is ready for commit.

Since having both cron and crontab have been confusing for users on this thread (including myself), crontab could be moved out of the 1.44M floppy distribution, and cron moved into the 'misc utils' distribution so as to not have to have a seperate config entry to get one program.

> And this simplified version should not crash as the current on exit.

I fixed crontab with the last push to this PR, please test, as you had said your version with the heap change didn't work.

toncho11:
Now it works. I saw in the code why. 

Cron is very useful and should be in every distribution by default. 

If we use crontab it can create problems if not simplified. 
If we do not use it, I do not know if the user will be able to create properly the root file in /var/cron. It must be tested and documented before removing crontab.

ghaerr:
> If we do not use it, I do not know if the user will be able to create properly the root file in /var/cron. It must be tested and documented before removing crontab.

Well, crontab doesn't do any checking on what is entered with crontab -e, nor crontab \<file>. I think @Mellvik is right, its only real purpose is to mess with permissions for different users of cron. Personally, I found having the two seperate programs quite confusing, as did you when first trying things out. It's another UNIX-style program where one has to know everything in order to make it work, including its special file formats and usage patterns.

toncho11:
Yes. I did create the cron folder and file with vi. It works. We just need to document it. In a wiki page? Write that the user should check cron.log for any errors, etc.

Mellvik:
@toncho11, @ghaerr - I cannot resist the temptation to provide another small piece of history - which may turn out to be useful - or not.

Early unix systems were extremely simple and efficient - as such, great models for elks. Cron was created to take care of regular maintenance tasks, as specified in /etc/crontab, which is now /var/cron/root in elks. Only root could edit the crontab, while cron could (and can) set the userid for the jobs when starting them.
When regular users needed to run batch jobs, at() and atrun() were created. atrun is run by cron every (say) 30 seconds, and will run jobs created by users using the at command:
at <time> <command>
The commands were placed in /usr/spool/at/<file> and - for simplicity, <file> was the time at which the command should be run. And deleted as soon as it has been run.

As I mentioned before, the crontab command was added much later and has an unfortunate name because it created confusion between the origininal crontab (file) and the command. Crontab (the command) took care of the need for regular users to run commands regularly - which had previously been done by at-commands scheduling a new run of themselves.

at/atrun are very simple - or they used to be before todays bloated variants. If it is determined that such functionality is needed in elks, at/atrun may be a better choice than crontab (the command).

--Mellvik

ghaerr:
@toncho11 and @Mellvik,

Thanks for all your testing and input on this. I'm going to commit this now, as wait and cron now work, and cleanup cron in a separate commit.

Thank you!

PR #899, opened 2021-01-16
Handle AltGr key mode with FR keyboard

hmlth:
Without this the AltGr don't work as expected with the french
azerty keyboard, for example the AltGr+" key generates a graphical
character and not #.

Tested on a IBM XT/286 and a french Model M AT keyboard.


ghaerr:
Hello @hmlth,

Thank you for using ELKS and providing this fix!

PR #901, opened 2021-03-29
[chmem] Also include far text segment size in output

tkchia:


tkchia:
This should address a flaw in `chmem` highlighted at https://github.com/tkchia/gcc-ia16/issues/73 .

ghaerr (approved):
Looks great @tkchia, thanks!

PR #904, opened 2021-03-30
[dircon] Add ANSI sequences necesssary for nano and mined editors

ghaerr:
Adds the following ANSI sequence recognitions to Direct and BIOS consoles:
```
\E[d    vpa
\E[G    hpa
\E[J    clear screen to bottom
\E[K    clear line to EOL
\E[1K   clear line to BOL
\E[2K   clear line
\E[p1;p2;p3m    varargs ansi color/mode
```

### DIFF
--- a/elks/arch/i86/drivers/char/console.c
+++ b/elks/arch/i86/drivers/char/console.c
@@ -105,8 +105,18 @@ static void AnsiCmd(register Console * C, char c)
 	if (C->cy > MaxRow)
 	    C->cy = MaxRow;
 	break;
+    case 'd':			/* Vertical position absolute */
+	C->cy = parm1(C->params) - 1;
+	if (C->cy > MaxRow)
+	    C->cy = MaxRow;
+	break;
     case 'C':			/* Move right n characters */
 	C->cx += parm1(C->params);
+	if (C->cx > MaxCol)
+	    C->cx = MaxCol;
+	break;
+    case 'G':			/* Horizontal position absolute */
+	C->cx = parm1(C->params) - 1;
 	if (C->cx > MaxCol)
 	    C->cx = MaxCol;
 	break;
@@ -118,12 +128,23 @@ static void AnsiCmd(register Console * C, char c)
     case 'H':			/* cursor move */
 	Console_gotoxy(C, parm2(C->params) - 1, parm1(C->params) - 1);
 	break;
-    case 'J':			/* erase */
-	if (parm1(C->params) == 2)
+    case 'J':			/* clear screen */
+	n = atoi((char *)C->params);
+	if (n == 0) { 		/* to bottom */
+	    ClearRange(C, C->cx, C->cy, MaxCol, C->cy);
+	    if (C->cy < MaxRow)
+		ClearRange(C, 0, C->cy, MaxCol, MaxRow);
+	} else if (n == 2)	/* all*/
 	    ClearRange(C, 0, 0, MaxCol, MaxRow);
 	break;
-    case 'K':			/* Clear to EOL */
-	ClearRange(C, C->cx, C->cy, MaxCol, C->cy);
+    case 'K':			/* clear line */
+	n = atoi((char *)C->params);
+	if (n == 0)		/* to EOL */
+	    ClearRange(C, C->cx, C->cy, MaxCol, C->cy);
+	else if (n == 1)	/* to BOL */
+	    ClearRange(C, 0, C->cy, C->cx, C->cy);
+	else if (n == 2)	/* all */
+	    ClearRange(C, 0, C->cy, MaxCol, C->cy);
 	break;
     case 'L':			/* insert line */
 	ScrollDown(C, C->cy);
@@ -132,18 +153,21 @@ static void AnsiCmd(register Console * C, char c)
 	ScrollUp(C, C->cy);
 	break;
     case 'm':			/* ansi color */
-	n = atoi((char *)C->params);
-	if (n >= 30 && n <= 37) {
+      {
+	char *p = (char *)C->params;
+	do {
+	  n = atoi(p);
+	  if (n >= 30 && n <= 37) {
 	    C->attr &= 0xf8;
 	    C->attr |= (n-30) & 0x07;
 	    C->color = C->attr;
-	}
-	else if (n >= 40 && n <= 47) {
+	  }
+	  else if (n >= 40 && n <= 47) {
 	    C->attr &= 0x8f;
 	    C->attr |= ((n-40) << 4) & 0x70;
 	    C->color = C->attr;
-	}
-	else switch (n) {
+	  }
+	  else switch (n) {
 	    case 0:
 		C->attr = C->color;
 		break;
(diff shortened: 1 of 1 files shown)


Mellvik:
@ghaerr, 
thanks for a veritable queue of important  improvements - I'm trying to catch up with testing without running out of breath ...

While I haven't tested the augmented nano, I noted that 'colourful ls' (as in most Linuxes) doesn't like the new ANSI codes. At least I'm assuming that to be the explanation - see pictures below. I'm telnetting to a Linux box (RaspberryPi), and doing a `ls -l` (using the fd image you supplied a few days ago, thanks BTW!). Before this change, there was no reverse video or any other effects when running color-ls via telnet.


Is there an easy fix for this?



--Mellvik

IMG_6661
Skjermbilde 2021-04-07 kl  13 11 18

ghaerr:
Hello @Mellvik,

As usual, thanks a lot for your testing!

>  I noted that 'colourful ls' (as in most Linuxes) doesn't like the new ANSI codes. At least I'm assuming that to be the explanation - see pictures below. I'm telnetting to a Linux box (RaspberryPi), and doing a ls -l (using the fd image you supplied a few days ago, thanks BTW!). Before this change, there was no reverse video or any other effects when running color-ls via telnet.
> Is there an easy fix for this?

The new ANSI sequences shouldn't affect this, but the color and reverse video mode sequence `ESC [ ... m` was rewritten, that's likely where the problem is.

The first thing to check is whether the RPi TERM variable is set to ANSI, rather than xterm-256color. But still, it seems something is broken.

Can you capture an "ls -l > file" and post the capture? That would help, since my system doesn't have color ls.

Is the Compaq screenshot showing the ls -l color output using `miniterm`?

> using the fd image you supplied a few days ago, thanks BTW!

If you're running that image on the Compaq, then you're running the new compressed binaries! Can you notice any increase in speed from floppy reads, or decrease in speed from decompressing (should be fast)?

Thank you!

ghaerr:
> Can you capture an "ls -l > file" and post the capture? That would help, since my system doesn't have color ls.

Oops - redirecting color ls output causes ls to not emit any color sequences. I've got macOS `ls -G` displaying color sequences, but can't reproduce your problem yet on macOS.

Looking at `man ls`, a couple environment variables control `ls` color output: LSCOLORS and CLICOLOR. I don't see these exported in the macOS environment, can you check your RPi system to see if they might be exported by `env`? The man page also says setting "CLICOLOR_FORCE" in the environment will force keeping the escape sequences emitted when redirecting to a file: "CLICOLOR_FORCE= ls -G > file" should do the trick.

A closer look at your screenshots seem to show that the bug occurs at the end of a line where reverse video is being shown, and ELKS console is carrying over the reverse video onto the next line. There may be an additional problem with reverse video on line 25 as well. 

Another thought would be to unset LSCOLORS and CLICOLOR in the environment, to see how that affects things.

Mellvik:
Thanks @ghaerr,
And yes, ls will not do that, but a screen() session output file (screenlog.0 etc.) does.
So I got a full dump, esc sequences and the shebang. Will upload and analyze in the morning.

-M

Mellvik:
OK; we got some more data on the issue. 

> Is the Compaq screenshot showing the ls -l color output using miniterm?

No, that screenshot was not from the compaq at all (it was from the mac ssh'd to the raspi), only to show the `ls` output from a 'regular' terminal window.

And yes, the `TERM` is set to ansi.

It turns out from inspecting the `ls` output (see below), that

- the `set green color` escape sequence ( ^[[01;32m) for some reason now triggers reverse video
- the `reset color` sequence (^[[0m) does not reset it
- ... but `set blue color` does the reset (^[[01;34m)
- I haven't been able to figure out why the reverse video sometimes extents to the end of the line (screen), and some times does not.

```
drwxr-xr-x  8 helge helge      4096 Oct 28  2018 ^[[01;34mvenix^[[0m^M
-rwxr-x---  1 helge helge      1574 May 14  2017 ^[[01;32mvenix-getty.mod^[[0m^M
drwx------  3 helge helge      4096 Feb 10  2016 ^[[01;34mvenix-tenox-images^[[0m^M
drwxr-xr-x 13 helge helge      4096 Oct 28  2018 ^[[01;34mvenix-umansky^[[0m^M
-rw-r--r--  1 helge root    1474560 Nov 11  2016 venix-xfer.dd^M
drwxr-xr-x  2 helge helge      4096 Apr 30  2018 ^[[01;34mvenix.backup^[[0m^M
-rw-r--r--  1 root  root   42606080 Oct 12 12:53 venix.raw.disk.oct2020.img^M
drwx------  7 helge helge      4096 May  3  2016 ^[[01;34mvenix_good_160502^[[0m^M
-rwxr-xr-x  1 helge helge      7483 Jul 23  2020 ^[[01;32mvtemu.com^[[0m^M
-rw-r--r--  1 helge helge     22712 Nov 13 13:05 x^M
drwxr-xr-x  2 helge helge      4096 Dec  3  2018 ^[[01;34mzmodem.debug^[[0m^M
^[[01;32mhelge@raspberrypi^[[00m:^[[01;34m~ $^[[00m 
```

@ghaerr, hope you can make some sense of this. Let me know if you need more data. 

--Mellvik

ghaerr:
Hello @Mellvik,

Thanks, I should be able to fix the issue from this data. Any chance you could upload the data file itself? I could then double check by cat-ing it out, rather than having to create a file with those contents myself.

Mellvik:
[ls-out-raspi.zip](https://github.com/jbruchon/elks/files/6279929/ls-out-raspi.zip)

The complete `ls` output included, containing the lines from the listing above towards the end.
Notice that the list is not identical to the ones in the previous screenshots, for some reason the sorting is different, possibly a locale related issue.

M

ghaerr:
Hello @Mellvik,

I think I've got this figured out. With the enhancement of parsing ESC [ p1;p2 m sequences with multiple parameters, the color attributes started working. The bug is that ESC [ 0 m doesn't turn off the color, just the bright attribute.

> the set green color escape sequence ( ^[[01;32m) for some reason now triggers reverse video

I'm thinking that's an issue on the Compaq Portable III mono screen: some "colors" are interpreted as reverse video, some not.

> I haven't been able to figure out why the reverse video sometimes extents to the end of the line (screen), and some times does not.

Since the ESC [ 0 m sequence doesn't turn off color, the color ends up eventually being shown in reverse video on the bottom line. When that happens, and a LF (scroll up) occurs, the right-hand side of the line is scrolled up with reverse video as well, by the BIOS.

I'll post a PR and bootable image for you.

Thank you!

ghaerr:
@Mellvik, try this bootable image. This should fix the color issue. There are two files in the root directory, cat-ing "file2" should look like this:
Screen Shot 2021-04-08 at 10 16 37 AM

[EDIT: Note in the image above, there are a couple lines offset by the "high intensity" colors. This is because of the ELKS-unimplemented ESC ( B sequence. When that is removed, the test displays perfectly.]

[fd1440.bin.zip](https://github.com/jbruchon/elks/files/6280008/fd1440.bin.zip)

Issue #909, opened 2021-04-03
Possible continuing toolchain issues?


ghaerr:
Hello @tkchia,

Using the most recent source and toolchain, I have run into a *very* strange problem when compiling the kernel and booting ELKS. The problem seemingly manifests only with the new toolchain, and removing both of our recent group of  commits (back to Jan 16) makes no difference: adding the following single line causes a startup failure using the new toolchain, and not with the old:
```
diff --git a/elks/fs/exec.c b/elks/fs/exec.c
index b168ce49..80233e76 100644
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -337,6 +337,7 @@ int sys_execve(char *filename, char *sptr, size_t slen)
 		   retval, (size_t)mh.tseg);
 	    goto error_exec4;
 	}
+printk("compr tseg xx bytesx1\n");
 #ifdef CONFIG_EXEC_MMODEL
 	if ((size_t)esuph.esh_ftseg) {
 	    currentp->t_regs.ds = seg_code->base + bytes_to_paras((size_t)mh.tseg);
```
With the above printk added, things fail. If a single byte is removed, the trailing "1" in the printk, things work.

The failure itself is quite strange: the boot appears to hang after the kernel relocation is applied:
<img width="832" alt="Screen Shot 2021-04-03 at 2 05 31 PM" src="https://user-images.githubusercontent.com/11985637/113493599-e0f71e80-949d-11eb-83d6-eaebbba92b9a.png">
However, the system will in fact boot and run seemingly normally if two characters are typed on the console!

The map files show no difference in text sections (when removing the single "1" character). I have printk-debugged the problem all the way into 'elks/arch/i86/drivers/char/xt_key.c::kbd_init()':
```
void kbd_init(void)
{
    /* Set off the initial keyboard interrupt handler */

    if (request_irq(KBD_IRQ, keyboard_irq, NULL))
    panic("Unable to get keyboard");

    set_leds();
    kb_read();      /* discard any unread keyboard input*/
}
```
The boot "hangs" just during calling set_leds(), and what appears to be happening is continuous interrupts (calls to) 'keyboard_irq'. I haven't been able to get much deeper as I'm only able to printk-debug and can't run elks-emu. Typing not one, but two characters at the console and the system continues the boot normally.

Does this happen on your system? 

[...]

PR #910, opened 2021-04-04
[kernel] bugfix: buf. overflow in mapping kbd. scancode to action

tkchia:
This bug may make ELKS "hang" at startup until the user presses some keys (https://github.com/jbruchon/elks/issues/909).

`set_leds()` causes the 8042 keyboard controller to return a `0xFA` (= KB_ACK) on the 8042 data port, and also issue an IRQ 1 for the `KB_ACK` (!).  Without this patch, the IRQ 1 handler may misinterpret the `0xFA` as a Caps Lock, Num Lock, or Scroll Lock key --- then it may erroneously invoke `set_leds()` again, leading to a spurious IRQ 1, leading to another spurious `set_leds()` call, etc.

### DIFF
--- a/elks/arch/i86/drivers/char/xt_key.c
+++ b/elks/arch/i86/drivers/char/xt_key.c
@@ -89,6 +89,8 @@ static unsigned char tb_state[] = {
     SSC, SSC, SSC, SSC, SSC, SSC, SSC, 'k', 'l' /*50->58, F11-F12*/
 };
 
+#define TB_STATE_MAX	(sizeof(tb_state) / sizeof(tb_state[0]) - 1)
+
 /*
  * Map CAPS|ALT|CTL|SHIFT into NORMAL,SHIFT,CAPS,CTL-ALT,
  * which are used to index into scan_tabs[].
@@ -192,7 +194,7 @@ static void keyboard_irq(int irq, struct pt_regs *regs, void *dev_id)
      *         10xx xxxxB, 0x80 Extended scan code
      *         11xx xxxxB, 0xC0 Simple Scan Code
      */
-    if (code >= 0x1C)
+    if (code >= 0x1C && code - 0x1C <= TB_STATE_MAX)
 	mode = tb_state[code - 0x1C]; /*is key a modifier key?*/
     else
         mode = SSC;


tkchia:
With this patch, the `0xFA` (`KB_ACK`) will be harmlessly interpreted as a release of a (likely non-existent) key with scan code `0x7A`.

ghaerr:
Hello @tkchia,

Amazing debugging work, thank you!!

Wow. So this kernel bug has been around since set_leds() was implemented, and depends on the contents of the overrun tb_state[] array, indexed by 0xFA (a key-up of non-existent scan code 7A and the keyboard ACK response). And it just so happens that with the new toolchain building the fartext kernel that the byte at that index equals the value of CAPS, NUM or SLOCK, causing another call to set_led() at interrupt time!!

> Under a buggy kernel, the bug is also triggered if you press Caps Lock, Num Lock, or Scroll Lock at some point after the system starts --- the terminal will weird out for a while.

I have tested the lock keys on real hardware. I suppose this is then the result of the hardware race between exactly where the kernel is inside kb_wait() and the subsequent forced calls to set_led() causing the loop. The interrupt timing must be different on hardware vs QEMU.

Separately, I'm wondering whether the whole approach taken by set_leds() is incorrect, with kbd_ack() busy looping, waiting for an 0xFA KBD_ACK from the keyboard controller; this will always cause a keyboard interrupt and possible race condition for reading the contents of the keyboard port at 0x64?

Thank you very much for this bug fix!!

ghaerr (approved):

tkchia:
Hello @ghaerr,

> Separately, I'm wondering whether the whole approach taken by set_leds() is incorrect, with kbd_ack() busy looping, waiting for an 0xFA KBD_ACK from the keyboard controller; this will always cause a keyboard interrupt and possible race condition for reading the contents of the keyboard port at 0x64?

Indeed --- it seems the code is mixing interrupt-driven I/O with polled I/O on the same ports.

Perhaps, if each write to port `0x60` to set the LED, i.e.
```
outb_p(LED_CODE, KEYBD);
```
and
```
outb_p(leds, KEYBD);
```
will result in an IRQ 1, maybe we should rely on the IRQ 1 invocations to tell when each command byte has been received.  This would mean turning `set_leds()` into some sort of interrupt-driven state machine.  [**Edit:** perhaps we should also need some sort of queue structure to hold the list of things we want to send to the 8042...]

Thank you!

PR #911, opened 2021-04-04
[kernel] use interrupt-driven I/O when setting keyboard LEDs

tkchia:
Follow-up to the PR https://github.com/jbruchon/elks/pull/910 .

### DIFF
--- a/elks/arch/i86/drivers/char/xt_key.c
+++ b/elks/arch/i86/drivers/char/xt_key.c
@@ -19,6 +19,7 @@
  *
  * 14 Jul 20 Helge Skrivervik Added LED and NUMLOCK/CAPSLOCK processing.
  * 17 Jul 20 Greg Haerr Added SCRLOCK processing, documentation, rewrite for clarity.
+ * 15 Apr 21 TK Chia Using interrupt-driven I/O to update keyboard LEDs.
  */
 
 #include <linuxmt/config.h>
@@ -28,6 +29,7 @@
 #include <linuxmt/errno.h>
 #include <linuxmt/fs.h>
 #include <linuxmt/fcntl.h>
+#include <linuxmt/timer.h>
 #include <linuxmt/chqueue.h>
 #include <linuxmt/signal.h>
 #include <linuxmt/ntty.h>
@@ -71,6 +73,17 @@ static unsigned int ModeState = 0;
 static int capslock;
 static int numlock;
 static int scrlock;
+/*
+ * Whether we are currently trying to send a command to the keyboard
+ * controller to update the LEDs, and at what stage are we in sending the
+ * command.
+ */
+static enum {
+    KS_FREE,
+    KS_SETTING_LED_1,
+    KS_SETTING_LED_2
+} kb_cmd_state;
+static struct timer_list kb_cmd_timer;
 
 /*
  * Table for mapping scancodes >= 0x1C into scan code class.
@@ -132,9 +145,28 @@ void kbd_init(void)
     if (request_irq(KBD_IRQ, keyboard_irq, NULL))
 	panic("Unable to get keyboard");
 
-    set_leds();
+    clr_irq();
     kb_read();      /* discard any unread keyboard input*/
+    set_irq();
+
+    set_leds();
 }
+
+/* Definitions from MINIX 2*/
+
+/* Standard and AT keyboard.  (PS/2 MCA implies AT throughout.) */
+#define KEYBD		0x60	/* I/O port for keyboard data */
+
+/* AT keyboard. */
+#define KB_COMMAND	0x64	/* I/O port for commands on AT */
+#define KB_STATUS	0x64	/* I/O port for status on AT */
+#define KB_ACK		0xFA	/* keyboard ack response */
+#define KB_OUT_FULL	0x01	/* status bit set when keypress char pending */
+#define KB_IN_FULL	0x02	/* status bit set when not ready to receive */
+#define LED_CODE	0xED	/* command to keyboard to set LEDs */
+#define KBIT		0x80	/* bit used to ack characters to keyboard */
+
+static void kbd_send_cmd(int);
 
 /*
  *	XT style keyboard I/O is almost civilised compared
@@ -362,22 +394,6 @@ static void keyboard_irq(int irq, struct pt_regs *regs, void *dev_id)
     }
 }
 
-/* LED routines from MINIX 2*/
-
-/* Standard and AT keyboard.  (PS/2 MCA implies AT throughout.) */
-#define KEYBD		0x60	/* I/O port for keyboard data */
-
-/* AT keyboard. */
-#define KB_COMMAND	0x64	/* I/O port for commands on AT */
-#define KB_STATUS	0x64	/* I/O port for status on AT */
-#define KB_ACK		0xFA	/* keyboard ack response */
-#define KB_OUT_FULL	0x01	/* status bit set when keypress char pending */
-#define KB_IN_FULL	0x02	/* status bit set when not ready to receive */
-#define LED_CODE	0xED	/* command to keyboard to set LEDs */
-#define MAX_KB_ACK_RETRIES 0x1000	/* max #times to wait for kb ack */
-#define MAX_KB_BUSY_RETRIES 0x1000	/* max #times to loop while kb busy */
-#define KBIT		0x80	/* bit used to ack characters to keyboard */
-
 /* Read keyboard and acknowledge controller */
 static int kb_read(void)
 {
@@ -392,44 +408,62 @@ static int kb_read(void)
     return(code);
 }
 
-/* Wait until the controller is ready; return zero if this times out. */
-static int kb_wait(void)
-{
-    int retries;
-    unsigned char status;
+static void restart_timer(void);
 
-    retries = MAX_KB_BUSY_RETRIES + 1;	/* wait until not busy */
-    while (--retries != 0 && (status = inb_p(KB_STATUS)) & (KB_IN_FULL|KB_OUT_FULL)) {
-	if (status & KB_OUT_FULL)
-	    inb_p(KEYBD);		/* discard */
-    }
-    return(retries);		/* nonzero if ready */
-}
-
-/* Wait until kbd acknowledges last command; return zero if this times out. */
-static int kb_ack(void)
-{
-    int retries;
-
-    retries = MAX_KB_ACK_RETRIES + 1;
-    while (--retries != 0 && inb_p(KEYBD) != KB_ACK)
-	;			/* wait for ack */
-    return(retries);		/* nonzero if ack received */
-}
-
-/* Set the LEDs on the caps, num, and scroll lock keys */
-static void set_leds(void)
+/* Called by the scheduler or keyboard_irq() to send a command to the
+   keyboard controller.  IRQs are assumed to be enabled. */
+static void kbd_send_cmd(int data)
 {
     unsigned char leds;
 
+    for (;;) {
+	switch (kb_cmd_state) {
+	case KS_SETTING_LED_1:
+					/* poll for buffer empty */
+	    if (inb_p(KB_STATUS) & KB_IN_FULL) {
+		restart_timer();	/* if not, schedule a later poll */
+		return;
+	    }
+	    outb_p(LED_CODE, KEYBD);	/* prepare keyboard to accept LED
+					   values */
+	    kb_cmd_state = KS_SETTING_LED_2;
+	    break;
+	case KS_SETTING_LED_2:
+	    if (inb_p(KB_STATUS) & KB_IN_FULL) {
+		restart_timer();
+		return;
+	    }
+					/* encode LED bits */
+	    leds = scrlock | (numlock << 1) | (capslock << 2);
+	    outb_p(leds, KEYBD);	/* give keyboard LED values */
+	    kb_cmd_state = KS_FREE;
+	default:
+	    return;
+	}
+    }
+}
+
+/* Arrange to call kbd_send_cmd() after a short period of time. */
+static void restart_timer(void)
+{
+    init_timer(&kb_cmd_timer);
+    kb_cmd_timer.tl_expires = jiffies + 2;	/* every 2/100 second*/
+    kb_cmd_timer.tl_function = kbd_send_cmd;
+    add_timer(&kb_cmd_timer);
+}
+
+/* Set the LEDs on the caps, num, and scroll lock keys.  IRQs are assumed to
+   be enabled. */
+static void set_leds(void)
+{
     if (!(sys_caps & CAP_KBD_LEDS)) return;	/* PC/XT doesn't have LEDs */
 
-    kb_wait();			/* wait for buffer empty  */
-    outb_p(LED_CODE, KEYBD);	/* prepare keyboard to accept LED values */
-    kb_ack();			/* wait for ack response  */
-
-    kb_wait();			/* wait for buffer empty  */
-    leds = scrlock | (numlock << 1) | (capslock << 2);	/* encode led bits*/
-    outb_p(leds, KEYBD);	/* give keyboard LED values */
-    kb_ack();			/* wait for ack response  */
+    clr_irq();
+    if (kb_cmd_state == KS_FREE) {
+	/* if already in the middle of setting LEDs, then nothing to do;
+	   otherwise, schedule a timer event to to set LEDs */
+	kb_cmd_state = KS_SETTING_LED_1;
+	restart_timer();
+    }
+    set_irq();
 }


ghaerr:
Hello @tkchia,

Thank you for this fix.

I've been knee-deep implementing executable compression and haven't had the time to fully review this yet.

I do have a couple comments based a first look:
- Has this been tested on real hardware yet?
- I am a bit concerned about the interrupt "handler" calling set_led_2(), which then calls kb_wait(), which could busy-loop and hold off other interrupts for some time... I realize the original code did just this with the calls to set_led() on CAPS/NUM/SLOCK key releases, but can you think of any other way to accomplish this without busy looping at interrupt time? (Without using a scheduler timer, that is?)
- I would like to see the handler integrity check and possible call to panic() removed on 439-440. I've spent a lot of time getting rid of non-essential panic calls in the kernel, for a variety for reasons. In this case, there isn't really a need after the routine is debugged.

Thank you!

tkchia:
Hello @ghaerr,

> I've been knee-deep implementing executable compression and haven't had the time to fully review this yet.

Thanks, understood.  This PR can wait.

> * Has this been tested on real hardware yet?

No unfortunately --- I do not really have a way to test this on bare metal with my setups at hand.

> * I am a bit concerned about the interrupt "handler" calling set_led_2(), which then calls kb_wait(), which could busy-loop and hold off other interrupts for some time... I realize the original code did just this with the calls to set_led() on CAPS/NUM/SLOCK key releases, but can you think of any other way to accomplish this without busy looping at interrupt time? (Without using a scheduler timer, that is?)

Not yet. 😐  I will probably need to see how other OSes deal with this problem.  RBIL does not seem to give any clues on how I might do that (or maybe I am looking in the wrong places...).

> * I would like to see the handler integrity check and possible call to panic() removed on 439-440. I've spent a lot of time getting rid of non-essential panic calls in the kernel, for a variety for reasons. In this case, there isn't really a need after the routine is debugged.

OK.  I have force-pushed a new commit to remove the `panic( )` call.

Thank you!

ghaerr:
Hello @tkchia,

> No unfortunately --- I do not really have a way to test this on bare metal with my setups at hand.

I will test this on real hardware once it's complete.

> (Without using a scheduler timer, that is?)

I think the problem can be artfully solved using a scheduler timer. Since setting the LEDs isn't really time-dependent, the system can take a few milliseconds before changing the LED status.

Are you familiar with how ELKS scheduler timers work? Kernel code can request that a procedure callback is run some number of msecs in the future. In this case, the approach would be that instead of using the handler callback mechanism you've coded, whenever the LED status should be changed, a scheduler timer would be set. The called back procedure would check once whether the kbd port is ready to accept input; if so, then the LED status change would be written, otherwise, the timer would be set again for a few msecs ahead. Finally, the keyboard interrupt handler would always just eat the ACK.

The ELKS scheduler runs on hardware clock ticks, and will schedule a callback in approximately the number of jiffies ahead of current time specified. The callback procedure then runs on the interrupt stack of the interrupted kernel process at hw timer interrupt time.

A good example of using a scheduler timer is in the BIOS "headless" console, which needs to poll the keyboard, since it doesn't use the keyboard interrupt (elks/arch/i86/drivers/char/headless-bios.c):
```
/* called by scheduler to poll BIOS keyboard*/
static void kbd_timer(int data)
{   
   int dav;

    if ((dav = bios_kbd_poll())) {
      ...
       Console_conin(dav & 0x7F);
    }
    restart_timer();
}

/* initialize timer for scheduler*/
static void restart_timer(void)
{
    static struct timer_list timer;

    init_timer(&timer);
    timer.tl_expires = jiffies + 8;     /* every 8/100 second*/
    timer.tl_function = kbd_timer;
    add_timer(&timer);
}
```

I think this method should solve the problem gracefully, finally.

Thank you!

tkchia:
Hello @ghaerr,

Sorry for the late reply!  Yes, I guess using the scheduler timer might be a better idea.  I will see if I can implement something along these lines the next few days.

Thank you!

ghaerr:
Hello @tkchia,

>  I will see if I can implement something along these lines the next few days.

Great! As a slight modification to the implementation description I described above, I am thinking that instead of using seperate handler functions for each state, using a single scheduler-timer handler function, which operates a state machine based on its own internal state, will be easiest to understand. The INT 1 keyboard handler, upon seeing an ACK, will merely update the state machine, with the next operation handled by the single handler function.

Thank you!

tkchia:
Hello @ghaerr,

Thanks!  I have updated the implementation as advised.  The PR now has a single function `kbd_send_cmd(.)` responsible for updating the state machine.  It may either be called from the timer scheduler, or from the IRQ 1 handler --- if the IRQ 1 handler happens to see an ACK before the timer event occurs, it will try to call `kbd_send_cmd(.)` early to advance the state machine.

Again, this probably needs more testing on a real machine.

---

(Incidentally, I am getting some spurious Git repository changes in `elkscmd/nano/nano-2.0.6/src/.deps/*.Po` when I rebuild ELKS.  Perhaps you have accidentally placed some non-source files under version control?)

Thank you!

Mellvik:
Thank you @tkchia,

I can test this on physical HW - 

what should I be looking for?


—mellvik

tkchia:
Thanks @Mellvik !  Mainly I hope to know whether the code to update the keyboard LEDs is good:

  * whether ELKS starts up normally without "hanging" (https://github.com/jbruchon/elks/issues/909) shortly after boot-up;
  * whether keyboard input works properly without weirding out, when you press Caps Lock, Num Lock, or Scroll Lock; and
  * whether the LEDs for Caps Lock, Num Lock, and Scroll Lock are updated right.

Thank you!

Mellvik:
Hi @tkchia,

I've tested your changes briefly on a Compaq, and everything seems fine. 

Thank you!

—Mellvik

tkchia:
Hello @Mellvik, hello @ghaerr,

I just spotted a corner case bug in my code --- `kbd_send_cmd(.)` might be called (from the timer event scheduler) with IRQs enabled, so a keyboard interrupt (IRQ 1) to service an ACK might happen while the processor is running _inside_ `kbd_send_cmd(.)`, which would cause a race condition and mess things up.

The new commit (https://github.com/jbruchon/elks/commit/d2b7d40bfaded9e4334af0d81a37142b84186f7f) makes `kbd_send_cmd(.)` disable IRQs temporarily, which should hopefully fix this --- and hopefully still work.

Apologies and thanks!

ghaerr:
Hello @tkchia,

I may need a bit to fully think through, and look at a non-diff listing of your solution. First impression is that it is much superior to the previous two implementations, and should work well.

> kbd_send_cmd(.) might be called (from the timer event scheduler) with IRQs enabled, so a keyboard interrupt (IRQ 1) to service an ACK might happen while the processor is running inside kbd_send_cmd(.), which would cause a race condition and mess things up.

I was thinking of just that issue the other day, and the need to protect *some* critical section, since the scheduler timer callback can occur in about three slightly different instances. Here are some comments pertaining to that:

- I think the `save_flags/clr_irq/restore_flags` approach is correct, vs using `clr_irq/set_irq`, but want to think through that a bit to make sure.  If `kbd_send_cmd` is never called via a scheduler timer callback, then it is not necessary.
- Regarding exactly when a kernel timer callback occurs: the kernel of course is interrupted whenever a hardware interrupt occurs and interrupts are enabled. ~~But kernel timer callbacks will never interrupt previously executing kernel code, as the timers are only called when `schedule()` is executed.~~ `schedule` may be called arbitrarily during kernel interrupt execution, to sleep a task, or after the timer interrupt interrupts an executing process in user mode for a task switch before switching stacks. If the kernel is "double interrupted", rescheduling is not allowed, and the previous kernel code will always be allowed to resume executing, *except for the case of timer callbacks, which are run by the IRQ 0 timer interrupt handler with interrupts enabled*. ~~In this latter case,~~ The timer interrupt and timer callback procedure may execute on the seperate kernel "interrupt" stack. ~~and the timer callback would occur either on the next clock tick interrupting user mode, or just before the next task switch.~~ [EDIT: I am sorry, I incorrectly thought the timers were called by `schedule`. They are called by the timer interrupt handler. ELKS kernel timer callbacks may interrupt previous executing kernel code, as they are called only during the timer IRQ 0 interrupt by `do_timer -> run_timer_list`. However, all kernel timer callbacks, (as well as all ELKS kernel interrupt handling routines in C), are executed with interrupts enabled.]
- The critical section is large again and kbd_send_cmd further attempts some busy-looping waiting for data with interrupts off. I'm not sure this is a good idea. Why not just remove the extra code polling (attempting to short-circuit a timer wait) and always wait for a timer callback? This would decrease the critical section greatly, eliminate any kernel polling, decrease kernel code, and eliminate the seemingly unneeded kbd_short_wait routine.
- Regarding your comment on line 431 of whether its a good idea to read and drop data - it may not be a good idea. Why not just let the IRQ 1 handler always read the kbd data, and have it handle the ACK specially?
- Line 482: We can probably decrease the jiffies wait to +1 (or +2 to be sure the jiffies has not expired) and decrease the time wait. I don't know how the long the keyboard controller takes to process commands.

Thank you!

ghaerr:
Hello @tkchia,

> (Incidentally, I am getting some spurious Git repository changes in elkscmd/nano/nano-2.0.6/src/.deps/*.Po when I rebuild ELKS. Perhaps you have accidentally placed some non-source files under version control?)


Geez. This is a by-product of the ridiculously complicated automake mechanism that I tried to get away from without actually rewriting the nano Makefiles from scratch. I don't see the problem on my end, but suspect that's because the \*.Po files have absolute paths (my paths anchored with /Users/greg/...) in them. Thank you for telling me, I suppose I'll need to rewrite the nano Makefiles in order to fix this.

Thank you!

ghaerr:
Hello @tkchia,

> kbd_send_cmd further attempts some busy-looping waiting for data with interrupts off. 

After further review, I now see what you're doing. Of course, you'll have poll once to see whether the keyboard can accept data. Thus, I'm thinking that no busy looping should occur - just a single `inb_p(KB_STATUS) & KB_IN_FULL)`, no procedure call necessary, for both KS_SETTING_LED{1,2}. Don't busy loop if not ready, instead, restart the timer as you are doing. In addition, don't read any kbd input data. That resolves most of the issues brought up above, and keeps the critical section both time and space wise very small.

Thank you!

tkchia:
Hello @ghaerr,

Thanks!  I have force-pushed a new commit to incorporate your suggestions.

> [... However, all kernel timer callbacks, (as well as all ELKS kernel interrupt handling routines in C), are executed with interrupts enabled.]

Hmm... indeed, after double-checking, I found that my understanding of how the `add_timer(.)` event handlers and the IRQ 1 handler `keyboard_irq(`...`)` was wrong:
  * Both kinds of event handlers are called with interrupts enabled.
  * `keyboard_irq(`...`)` is called before an EOI is issued, so it will most likely not re-enter itself, but it might still be pre-empted by a higher-priority (?) IRQ.

With the new understanding, I have updated the `set_leds()` and `kbd_send_cmd(.)` routines to use `clr_irq()` and `set_irq()`, hopefully in the right ways.

Thank you!

ghaerr:
Hello @tkchia,

Thank you for hanging in there through this long discussion, the implementation is now quite small and elegant, and the design is growing more interesting and smaller simultaneously! :)

> keyboard_irq(...) is called before an EOI is issued, so it will most likely not re-enter itself, but it might still be pre-empted by a higher-priority (?) IRQ.

Yes. The 8259 interrupt controller will prevent an IRQ 1 keyboard interrupt from interrupting itself, but the keyboard interrupt handler can certainly be interrupted by the only remaining higher priority interrupt, IRQ 0, the timer. And that means that kernel timer callback handlers can interrupt any other interrupt handling routines, including IRQ 1.

So this begs the question of exactly where is the critical section? More on that, but first a couple minor issues:
- The kernel is compiled with `-fno-inline` (see elks/Makefile-rules) so the routine `static inline int kb_in_ready(void)` should either have 'inline' removed or be inserted into the calling routine. Given that `inb_p` is an asm macro, it is probably less code with it inserted twice.
- The `flags_t flags` (line 428) can be removed., 

Now, the final unresolved issue, where is the critical section (as only that section needs clr_irq/set_irq protection)?

I am thinking that `kbd_send_cmd` isn't in fact the critical section: it doesn't actually need to run with interrupts disabled. This is because, in fact, it only runs as a timer callback procedure, at IRQ 0 time. Well, almost: there remains a problem area, lines 204-205 when an ACK is received in the IRQ 1 interrupt handler, and kbd_send_cmd is called. Calling kbd_send_cmd from IRQ 1 complicates things by requiring it to be protected. Consider the following:
```
    if (code == KB_ACK && kb_cmd_state != KS_FREE)
 	restart_timer();
```
This would allow clr_irq/set_irq to be removed entirely from kbd_send_cmd, since it would only then run from a timer callback, at IRQ 0 time. It could not be interrupted.

There is in fact another race condition (which can occur anytime a state machine variable is checked outside the state machine) in the above code, but it doesn't matter: Suppose that just after loading the variable "kbd_cmd_state" above in the IRQ 1 handler, (but before comparing it to KS_FREE), the IRQ 0 hardware interrupt occurs, and a scheduled kernel timer callback to send_kbd_cmd is executed. This has the potential to change the kb_cmd_state variable before returning, thus then having the value of kb_cmd_state incorrect in the IRQ 1 handler. However, in this case, we're only comparing to KS_FREE and worst case would result in another timer callback scheduled. Ultimately, testing any volatile state outside the state machine means having other parts of the kernel to be knowledgable about the state machine. Thus, perhaps 
```
    if (code == KB_ACK)
 	restart_timer();
```
would be superior, but it involves an extra timer scheduled every time.


So the critical section is only where a state machine transition is done *outside* of kbd_send_cmd, which is already protected and done in `set_leds` (lines 477-484).

What do you think?

Thank you!

tkchia:
Hello @ghaerr,

> ```
>     if (code == KB_ACK)
>  	restart_timer();
> ```

Well... actually my original intention was to allow `keyboard_irq` to speed up the state machine, by processing any `KB_ACK`'s right away.  And this is only useful if we do not need to wait for another timer event after seeing a `KB_ACK`.

Perhaps for now, I think I will `#if 0` away the code for handling `KB_ACK`, and leave everything to the timer scheduler.

Thank you!

ghaerr:
Hello @tkchia,

> Well... actually my original intention was to allow keyboard_irq to speed up the state machine, by processing any KB_ACK's right away. And this is only useful if we do not need to wait for another timer event after seeing a KB_ACK.

I see, good idea. However, I thought the kbd controller required that we wait for an ACK, rather than just waiting for the kbd input ready bit, before sending the next byte to it.


> Perhaps for now, I think I will #if 0 away the code for handling KB_ACK, and leave everything to the timer scheduler.

I'm not that familiar with exactly how the IBM PC 8042 will take input - if we don't need to wait for ACK, but just the input ready, that would work. The original code specifically waited for an ACK, before again waiting for keyboard input ready, however.

Thank you!

tkchia:
Hello @ghaerr,

> I'm not that familiar with exactly how the IBM PC 8042 will take input - if we don't need to wait for ACK, but just the input ready, that would work. The original code specifically waited for an ACK, before again waiting for keyboard input ready, however.

This probably needs more testing, again. :-|

(Anyway, I looked up `osdev.org` on the topic of keyboard controller I/O, and interestingly it [says](https://wiki.osdev.org/%228042%22_PS/2_Controller):

> If you send a command to the PS/2 controller that involves a response, the PS/2 controller may generate IRQ1, IRQ12, or no IRQ (depending on the firmware) when it puts the "response byte" into the buffer.

I am not sure if that means that there are some keyboard controllers which do not trigger an IRQ 1 when they send out an ACK.  This would make the ACKs kind of hard to handle in an interrupt-driven manner.)

Thank you!

ghaerr:
Hello @tkchia,

I got myself a bit confused on exactly how the state machine worked - I had thought the ACK processing in keyboard_irq was to restart the state machine (from waiting to send the next command), but now realize it was to speed up processing.

I don't know which is correct - I had assumed from reading the original code that one must wait for an ACK to proceed, after writing a command byte - which the current state machine doesn't do. That would require two more ACK waiting states. However, with a long enough jiffies timer, the 8042 keyboard controller ACK and wait could be effectively ignored, as your proposed \#if 0 does, without having explicit wait states.

I've never seen the IBM PC 8042 firmware source, is there information on what it actually requires somewhere, (like JBIL)?

Thank you!

tkchia:
> I've never seen the IBM PC 8042 firmware source, is there information on what it actually requires somewhere, (like JBIL)?

No idea.  If you manage to find any concrete information, I would like to know too.

Thank you!

ghaerr:
Hello @tkchia,

> I am not sure if that means that there are some keyboard controllers which do not trigger an IRQ 1 when they send out an ACK. This would make the ACKs kind of hard to handle in an interrupt-driven manner.)

Very interesting. Perhaps you're on the right track by completely ignoring any ACK in IRQ 1 now: if the timer callback jiffies were increased to a reasonably large value, then the command bytes would always be sent slowly without regard to ACKs received. This will likely be able to be confirmed through testing on an old PC.

Thank you!

ghaerr:
Hello @tkchia,

I will test the updated code on real hardware, I have a "faster" old desktop 386 and a slow Compaq Portable. If I see issues, I will increase the jiffies offset higher and try again. Upon confirmation of it working, I suggest we remove the "\#if 0" code entirely and commit the PR.

Thank you!

ghaerr (approved):
Hello @tkchia,

I have extensively tested this on my two real hardware machines, including typing whilst simultaneously pressing Caps, Num and Scroll Lock keys, and everything works perfectly.

Given what you've found regarding some keyboard controllers never returning an ACK, I would suggest permanently removing the \#if 0 code and commiting this. It seems the current jiffies timeout factor is large enough to handle the 8042 response wait and no ACK check need be performed.

Great job, thank you for the contribution!

tkchia:
Thanks @ghaerr !  I have updated the commit and merged it.

PR #903, opened 2021-03-30
[LD scripts] bugfix: handle -ffunction-sections & "unique" sections

tkchia:
Addresses https://github.com/tkchia/gcc-ia16/issues/73#issuecomment-810344556 .

PR #915, opened 2021-04-08
[cmds] Port nano-2.0.6, libncurses-5.2 and libregex to ELKS

ghaerr:
This PR adds the popular Linux text editor `nano`, along with the `ncurses` and `regex` libraries, to ELKS.

The build uses the recently-added compressed executables from #912 and the seperate function section compilation from #903.

This port is a proof-of-concept showing the now-reliable implementation of large medium-model binaries as well as effective compressed executables. Uncompressed, `nano` is 142,392 bytes, compressed size is 92,142 (36% reduction while retaining all capabilities). The text segments are nearly full at a combined 112k. That said, nano is somewhat ridiculously large for what it does, but I guess that's what you get when you port an in-the-wild program designed for desktops to a small system these days. It does show that our tools and the kernel are now able to handle much larger and more complex programs though.

Thank you @tkchia for help with debugging the `nano` port, and getting compiled function sections working with the linker to allow the regex library functions to be more evenly distributed between text and .fartext sections, which otherwise prohibited the link because of its large, single-file 34k code segment compiled normally.

The "Makefiles" of this particular port are unusual for ELKS - instead of being small and clean, they consist of the horrendous "automake/autoconf" automatically-generated Makefiles, which change on almost every individual system built on. After getting nano completely ported to ELKS on my laptop, I think an extra three hours was spent messing around with getting the build to run on the GitHub workflow image, and about ten extra commits. This can also be considered a lesson in porting to ELKS: for long-term use, it's probably better to rewrite the Makefiles entirely, but if you look at nano-2.0.6 and ncurses-5.2, you'll see why I didn't!

468 of the 470 files changed for this port live under the elkscmd/nano directory. Two more nano-X programs were temporarily removed from the distribution to make space for the nano binary.

Screen Shot 2021-04-07 at 5 20 55 PM


ghaerr:
Based on @Mellvik's comment about nano v2.0.6 running out of memory on almost all operations in https://github.com/jbruchon/elks/pull/912#issuecomment-815652107, I am going to hold this PR until more is known on v2.0.6's usefulness. Otherwise, I will update this PR to use nano v1.0.6, which doesn't have these memory issues.

ghaerr:
Further investigation shows that both nano v2.0.6 and v1.1.6 suffer from bad out-of-memory conditions. By slimming down on almost all configurable options, nano v1.1.6 was able to edit its own .nanorc file (4k) (!). Unfortunately, it too runs out of memory merely editing a file twice that size, 8k bytes. Thus, the nano editors really aren't practical for any real use on ELKS, and definitely won't default to taking up valuable space on the 1.44M floppy distribution.

I am wondering whether to shelve this PR entirely, or nonetheless add one of the versions, for the original reason of proof of concept for large medium model programs, as well as ANSI console compliance. The PR brings in tons of files, but all live only under the single elkscmd/nano directory. The port was tremendous help in debugging some of the last linker issues for medium model, as well as providing a large test case for executable decompression.

I suppose it comes down to whether anyone cares about the size of the ELKS git repository, vs having large testing programs available. Thoughts?

Mellvik:
@ghaerr,

Good and insightful summary. I vote for keeping it - and keeping the largest version. Since neither version is useful for what  it does, the largest must be better because it pushes the tools and the system harder.

BTW - will the regex and termcap libraries be useful outside the realms of nano?


Thank you.

—M

ghaerr:
Hello @Mellvik,

> BTW - will the regex and termcap libraries be useful outside the realms of nano?

Yes, although I wouldn't jump to use them unless you have to. regex is 34k code, and the ncurses library is large, for what it does on ELKS. It doesn't actually search a terminfo database, nor /etc/termcap: it is using a 'fallback' feature and only supports ANSI. There isn't an option in building ncurses to just use /etc/termcap. Nonetheless, it could be useful for porting programs heavily built around curses.

Thank you!

toncho11:
It ncurses ported? 

Would this example below work?

```
#include<ncurses.h>
main() {
WINDOW *w;
char list[5][7] = { "One", "Two", "Three", "Four", "Five" };
char item[7];
int ch, i = 0, width = 7;
initscr(); // initialize Ncurses
w = newwin( 10, 12, 1, 1 ); // create a new window
box( w, 0, 0 ); // sets default borders for the window
// now print all the menu items and highlight the first one
for( i=0; i<5; i++ ) {
if( i == 0 ) 
wattron( w, A_STANDOUT ); // highlights the first item.
else
wattroff( w, A_STANDOUT );
sprintf(item, "%-7s",  list[i]);
mvwprintw( w, i+1, 2, "%s", item );
}
wrefresh( w ); // update the terminal screen
i = 0;
noecho(); // disable echoing of characters on the screen
keypad( w, TRUE ); // enable keyboard input for the window.
curs_set( 0 ); // hide the default screen cursor.
// get the input
while(( ch = wgetch(w)) != 'q'){ 
// right pad with spaces to make the items appear with even width.
sprintf(item, "%-7s",  list[i]); 
mvwprintw( w, i+1, 2, "%s", item ); 
// use a variable to increment or decrement the value based on the input.
switch( ch ) {
case KEY_UP:
i--;
i = ( i<0 ) ? 4 : i;
break;
case KEY_DOWN:
i++;
i = ( i>4 ) ? 0 : i;
break;
}
// now highlight the next item in the list.
wattron( w, A_STANDOUT );
sprintf(item, "%-7s",  list[i]);
mvwprintw( w, i+1, 2, "%s", item);
wattroff( w, A_STANDOUT );
}
delwin( w );
endwin();
}
```

ghaerr:
Hello @toncho11,

> It ncurses ported?

Yes, with caveats. It was ported as part of bringing the `nano` editor to ELKS. Both nano and ncurses, while supposedly 'small' are beasts, which require the entire GNU './configure' mechanism to build, which are ill-suited to the ELKS build mechanisms. And they're memory hogs. `nano` and the ncurses library will build on most, but not all systems. As such, it was removed from the elks/elkscmd/Makefile standard ELKS applications build because some systems failed. You can test it by adding 'nano' back into the SUBDIRS= line in that Makefile and both ncurses and nano will hopefully make on your system.

> Would this example below work?

Yes, but you'll be a bit on your own grabbing the header file and library.

The ncurses library is built in elks/elkscmd/nano/ncurses-5.2/lib, and the header file is in elks/elkscmnd/nano/ncurses-5.2. 

Getting nano built was a lot of work, and its dependencies on ncurses and the build environment were never fully unravelled; doing so would have meant complete incompatibility with any future nano or ncurses library, so they were mostly left as-is.

If the nano build works on your system, you should be able to create a special Makefile for building most other ncurses applications, including your example. Note: ncurses is currently hard-coded to use the 'ansi' terminal type, which will work on the ELKS console and serial lines with terminal emulators emulating ANSI.

Thank you!

PR #917, opened 2021-04-08
[console] Reset color on ESC [ 0 m sequence

ghaerr:
Fixes color problem found my @Mellvik in #904.


Mellvik:
Thanks again @ghaerr,

I tested using your latest fd image this morning - it works as expected. 

Great work!

—M

PR #931, opened 2021-04-30
[emu86] Use BIOS extension detection

mfld-fr:
Simplify `emu86.sh` by using the detection mechanism implemented in mfld-fr/emu86#25.


ghaerr:
Thanks @mfld-fr!

PR #933, opened 2021-05-08
[console] Add configurable scancode vs BIOS keyboard driver

ghaerr:
Adds new option to configure the Direct Console and use a BIOS (non-interrupt, polling INT 16h) keyboard driver, rather than the more complex scancode driver (previously xt_kbd.c, now kbd-scancode.c).

This is being done not for PC machines, but instead for use in testing @mfld-fr's EMU86 8086 emulator, where emulation of EGA adaptor RAM in text mode is being added. In the future, this will also allow better portability to non-PC based systems that still wish to have the benefits of a direct-to-ram based console display.

Selecting the BIOS Console and Scancode Keyboard is not (yet) supported.


mfld-fr:
Useful option to ease video emulation development in EMU86, and better source organization (video adapter and keyboard are two different devices, even if linked together in a console).

PR #935, opened 2021-05-09
[config] Update EMU86 configuration files

ghaerr:
Updates `emu86-disk.config` for ELKS Direct Console with INT 16h (polling) keyboard for testing an almost-unmodified standard ELKS kernel now that EMU86 has been updated with EGA video text mode emulation. This configuration can be emulated using `emu86 -I image/fd1440.bin`.

Updates `config.mk` for ELKS configuration when building `emu86`, requested by @mfld-fr in https://github.com/mfld-fr/emu86/pull/54#issuecomment-835844302.

Will wait for a few outstanding EMU86 PRs to be approved and then update EMU86_VER in tools/Makefile before committing.


ghaerr:
Will remove commented-out `cp config-elks.mk config.mk` after updating EMU86_VER in seperate PR, after EMU86 stabilizes.

PR #936, opened 2021-05-12
Move 8259 PIC and 8254 PIT handlers to seperate source files

ghaerr:
First of a few cleanup PRs to seperate chip-specific kernel files into their own source file for ease of porting.
Adds `irq-8259.c` and `timer-8254.c` source files for IBM PC architecture. Chip-specific PIC code in `irqtab.S` still remains.

This will be useful should @cocus or @mfld-fr submit PRs for 8086 SBCs with non-IBM PC programmable interrupt controllers (PICs) or programmable interval timers (PITs).

Updated porting-guide.txt to only functions required in ROM BIOS for ELKS.
The SIBO port is slowly being removed with the cleanup PRs.

### DIFF
--- a/Documentation/text/porting-guide.txt
+++ b/Documentation/text/porting-guide.txt
@@ -1,6 +1,6 @@
 ELKS Porting Guide
 ==================
-8 May 2021 - Rev 1 ghaerr
+11 May 2021 - Rev 2 ghaerr
 
 Platforms (target)
 -------------------------------------------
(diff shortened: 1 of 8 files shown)


mfld-fr:
Good first step, but I wonder how you will move the EOI out from the interrupt routine to make it generic ? Because EOI is implicit on PC, while it is explicit on some other platforms, so only the platform driver or device driver  knows what kind of EOI is needed.

ghaerr:
>  I wonder how you will move the EOI out from the interrupt routine to make it generic ? Because EOI is implicit on PC, while it is explicit on some other platforms, so only the platform driver or device driver knows what kind of EOI is needed.

Yes, the constant EOI as well as the (two) port numbers are hard-coded in irqtab.S. I figure to wait until someone submits a working change for their PIC/PIT and then craft a design that remains fast. but portable between the two platforms.

In the meantime, will continue to separate chip-specific code in kernel to new files.

PR #937, opened 2021-05-13
[kernel] Allow setup.S INITSEG values to be specified in config.h

ghaerr:
Further work on easing porting of ELKS to embedded or non-PC systems.

This PR allows system configuration normally determined by setup.S at kernel pre-boot time to be specified at compile-time in config.h to lessen dependencies on IBM PC BIOS architecture. This also centralizes more porting dependencies in one file, config.h.

Adds defines for 8259 PIC command and data ports.
Additional cleanup work in architecture-specific setup routine `setup_arch()`.
More SIBO code removed.

At some point in the future, a lot of the initial setup.S code dealing with system startup can be moved into a seperate area, allowing easier removal for porting to other systems.

### DIFF
--- a/elks/arch/i86/drivers/block/genhd.c
+++ b/elks/arch/i86/drivers/block/genhd.c
@@ -87,7 +87,7 @@ static void INITPROC add_partition(struct gendisk *hd, unsigned short int minor,
      * a device number, then we do not really need boot_partition.
      */
     if (ROOT_DEV == (0x80 | minor >> hd->minor_shift)) {
-	sector_t boot_start = setupw(0x1e2) | (sector_t) setupw(0x1e4) << 16;
+	sector_t boot_start = SETUP_PART_OFFSETLO | (sector_t) SETUP_PART_OFFSETHI << 16;
 	if (start == boot_start)
 	    boot_partition = minor;
     }
--- a/elks/arch/i86/drivers/block/init.c
+++ b/elks/arch/i86/drivers/block/init.c
@@ -49,7 +49,7 @@ void INITPROC device_init(void)
      * drive number.  If so, convert it into a proper <major, minor> block
      * device number.  -- tkchia 20200308
      */
-    if (!boot_rootdev && (setupw(0x1f6) & EF_BIOS_DEV_NUM) != 0) {
+    if (!boot_rootdev && (SETUP_ELKS_FLAGS & EF_BIOS_DEV_NUM) != 0) {
 	extern kdev_t INITPROC bioshd_conv_bios_drive(unsigned int biosdrive);
 
 	kdev_t rootdev = bioshd_conv_bios_drive((unsigned)ROOT_DEV);
--- a/elks/arch/i86/drivers/char/bioscon.c
+++ b/elks/arch/i86/drivers/char/bioscon.c
@@ -168,10 +168,10 @@ void console_init(void)
     register Console *C;
     register int i;
 
-    MaxCol = (Width = setupb(7)) - 1;
+    MaxCol = (Width = SETUP_VID_COLS) - 1;
 
     /* Trust this. Cga does not support peeking at 0x40:0x84. */
-    MaxRow = (Height = setupb(14)) - 1;
+    MaxRow = (Height = SETUP_VID_LINES) - 1;
 
     if (peekb(0x49, 0x40) == 7)  /* BIOS data segment */
 	NumConsoles = 1;
--- a/elks/arch/i86/kernel/irq-8259.c
+++ b/elks/arch/i86/kernel/irq-8259.c
@@ -12,6 +12,12 @@
 #include <arch/io.h>
 #include <arch/irq.h>
 
+/* 8259 Command/Data ports */
+#define PIC1_CMD   0x20   /* master */
+#define PIC1_DATA  0x21
+#define PIC2_CMD   0xA0   /* slave */
+#define PIC2_DATA  0xA1
+
 static unsigned char cache_21 = 0xff, cache_A1 = 0xff;
 
 /*
@@ -21,7 +27,7 @@ static unsigned char cache_21 = 0xff, cache_A1 = 0xff;
 void init_irq(void)
 {
 #ifdef CONFIG_HW_259_USE_ORIGINAL_MASK	/* for example Debugger :-) */
-    cache_21 = inb_p(0x21);
+    cache_21 = inb_p(PIC1_DATA);
 #endif
 }
 
@@ -32,10 +38,10 @@ void enable_irq(unsigned int irq)
     mask = ~(1 << (irq & 7));
     if (irq < 8) {
 	cache_21 &= mask;
-	outb(cache_21,((void *) 0x21)); //FIXME
+	outb(cache_21, PIC1_DATA);
     } else {
 	cache_A1 &= mask;
-	outb(cache_A1,((void *) 0xA1));
+	outb(cache_A1, PIC2_DATA);
     }
 }
 
@@ -58,10 +64,10 @@ void disable_irq(unsigned int irq)
     clr_irq();
     if (irq < 8) {
 	cache_21 |= mask;
-	outb(cache_21,((void *) 0x21));
+	outb(cache_21, PIC1_DATA);
     } else {
 	cache_A1 |= mask;
-	outb(cache_A1,((void *) 0xA1));
+	outb(cache_A1, PIC2_DATA);
     }
     restore_flags(flags);
 }
--- a/elks/arch/i86/kernel/irq.c
+++ b/elks/arch/i86/kernel/irq.c
@@ -78,15 +78,12 @@ int request_irq(int irq, void (*handler)(int,struct pt_regs *,void *), void *dev
     flag_t flags;
 
     irq = remap_irq(irq);
-    if (irq < 0)
+    if (irq < 0 || !handler)
 	return -EINVAL;
 
     action = irq_action + irq;
     if (action->handler != default_handler)
 	return -EBUSY;
-
-    if (!handler)
-	return -EINVAL;
 
     save_flags(flags);
     clr_irq();
@@ -94,7 +91,7 @@ int request_irq(int irq, void (*handler)(int,struct pt_regs *,void *), void *dev
     action->handler = handler;
     action->dev_id = dev_id;
 
-    enable_irq((unsigned int) irq);
+    enable_irq(irq);
 
     restore_flags(flags);
 
@@ -122,8 +119,6 @@ void free_irq(unsigned int irq)
 
     action->handler = default_handler;
     action->dev_id = NULL;
-/*    action->flags = 0;
-    action->name = NULL;*/
 
     restore_flags(flags);
 }
--- a/elks/arch/i86/kernel/system.c
+++ b/elks/arch/i86/kernel/system.c
@@ -12,34 +12,28 @@
 
 byte_t sys_caps;		/* system capabilities bits */
 
-#ifdef CONFIG_ARCH_SIBO
-extern long int basmem;
-#endif
-
 void INITPROC setup_arch(seg_t *start, seg_t *end)
 {
 #ifdef CONFIG_COMPAQ_FAST
-/*
- *	Switch COMPAQ Deskpro to high speed
- */
-    outb_p(1,0xcf);
+	outb_p(1,0xcf);	/* Switch COMPAQ Deskpro to high speed */
 #endif
 
-/*
- *	Fill in the MM numbers - really ought to be in mm not kernel ?
- */
-
-	/* Extend kernel data segment to maximum of 64K */
-	/* to make room for local heap */
+	/*
+	 * Set start to beginning of available main memory, which
+	 * is directly after end of the kernel data segment.
+	 *
+	 * Extend kernel data segment to maximum of 64K to make room
+	 * for local heap.
+	 *
+	 * Set end to end of available main memory.
+	 *
+	 * If ramdisk configured, subtract space for it from end of memory.
+	 */
 
 	/* *start = kernel_ds + (((unsigned int) (_endbss+15)) >> 4); */
 	*start = kernel_ds + 0x1000;
 
-#ifdef CONFIG_ARCH_SIBO
-	*end = basmem << 6;
-#else
-	*end = (seg_t)setupw(0x2a) << 6;
-#endif
+	*end = (seg_t)SETUP_MEM_KBYTES << 6;
 
 #if defined(CONFIG_RAMDISK_SEGMENT) && (CONFIG_RAMDISK_SEGMENT > 0)
 	if (CONFIG_RAMDISK_SEGMENT <= *end) {
@@ -56,12 +50,12 @@ void INITPROC setup_arch(seg_t *start, seg_t *end)
 	heap_add ((void *)endbss, 1 + ~endbss);
 
 	/* Misc */
-	ROOT_DEV = setupw(0x1fc);
+	ROOT_DEV = SETUP_ROOT_DEV;
 
 #ifdef SYS_CAPS
 	sys_caps = SYS_CAPS;	/* custom system capabilities */
 #else
-	byte_t arch_cpu = setupb(0x20);
+	byte_t arch_cpu = SETUP_CPU_TYPE;
 	if (arch_cpu > 5)		/* IBM PC/AT capabilities */
 		sys_caps = CAP_ALL;
 #endif
--- a/elks/arch/i86/kernel/timer-8254.c
+++ b/elks/arch/i86/kernel/timer-8254.c
@@ -38,14 +38,14 @@
 void enable_timer_tick(void)
 {
     /* set the clock frequency */
-    outb (TIMER_MODE2, (void*)TIMER_CMDS_PORT);
-    outb (TIMER_LO_BYTE, (void*)TIMER_DATA_PORT);	/* LSB */
-    outb (TIMER_HI_BYTE, (void*)TIMER_DATA_PORT);	/* MSB */
+    outb (TIMER_MODE2, TIMER_CMDS_PORT);
+    outb (TIMER_LO_BYTE, TIMER_DATA_PORT);	/* LSB */
+    outb (TIMER_HI_BYTE, TIMER_DATA_PORT);	/* MSB */
 }
 
 void disable_timer_tick(void)
 {
-    outb (TIMER_MODE0, (void*)TIMER_CMDS_PORT);
-    outb (0, (void*)TIMER_DATA_PORT);
-    outb (0, (void*)TIMER_DATA_PORT);
+    outb (TIMER_MODE0, TIMER_CMDS_PORT);
+    outb (0, TIMER_DATA_PORT);
+    outb (0, TIMER_DATA_PORT);
 }
--- a/elks/arch/i86/mm/init.c
+++ b/elks/arch/i86/mm/init.c
@@ -53,7 +53,7 @@ void INITPROC mm_stat(seg_t start, seg_t end)
 	*cp++ = setupb(i++);
 	if (i == 0x40) {
 	    printk("PC/%cT class machine, %s CPU, %uK base RAM",
-		    (sys_caps & CAP_PC_AT) ? 'A' : 'X', proc_name, setupw(0x2a));
+		    (sys_caps & CAP_PC_AT) ? 'A' : 'X', proc_name, SETUP_MEM_KBYTES);
 	    cp = proc_name;
 	    i = 0x50;
 	}
--- a/elks/include/linuxmt/config.h
+++ b/elks/include/linuxmt/config.h
@@ -5,6 +5,20 @@
 #include <linuxmt/major.h>
 
 /*
+ * Setup data - normally queried by startup setup.S code, but can
+ * be overridden for embedded systems with less overhead.
+ * See setup.S for more details.
+ */
+#define SETUP_VID_COLS		setupb(7)		/* BIOS video # columns */
+#define SETUP_VID_LINES		setupb(14)		/* BIOS video # lines */
+#define SETUP_CPU_TYPE		setupb(0x20)	/* processor type */
+#define SETUP_MEM_KBYTES	setupw(0x2a)	/* base memory in 1K bytes */
+#define SETUP_ROOT_DEV		setupw(0x1fc)	/* root device, kdev_t or BIOS dev */
+#define SETUP_ELKS_FLAGS	setupw(0x1f6)	/* flags for root device type */
+#define SETUP_PART_OFFSETLO	setupw(0x1e2)	/* partition offset low word */
+#define SETUP_PART_OFFSETHI	setupw(0x1e4)	/* partition offset high word */
+
+/*
  * System capabilities - configurable for ROM or custom installations.
  * Normally, all capabilities will be set if arch_cpu > 5 (PC/AT),
  * except when SYS_CAPS is defined for custom installations or emulations.


ghaerr:
Hello @cocus, hopefully these changes are not getting in the way of your already-coded work building a port for your hardware. Please let me know if you'd like me to hold off in certain areas to minimize rebasing, thanks!

cocus:
> Hello @cocus, hopefully these changes are not getting in the way of your already-coded work building a port for your hardware. Please let me know if you'd like me to hold off in certain areas to minimize rebasing, thanks!

That's fine! I've added support for the IRQ and PIT (wip), and now I will discuss some questions I had about the headless console. These changes don't affect my work at all. The previous one was a little bot tricky to rebase from, but that's okay

ghaerr:
> The previous one was a little bot tricky to rebase from, but that's okay

Sorry about that, but most of the big structural changes are done, except for possibly some cleanup in system.c. I'll wait for your changes before changing that.

> I will discuss some questions I had about the headless console.

Ok. I suggest starting with copying headless-bios.c for the first port. As you have probably seen, the PC version relies on two INT 16h calls. Feel free to rewrite or replace entirely for your "console I/O" as headless-xxx.c, as we don't need the headless console concept to remain tied to a BIOS function.

cocus:
Yes, I think there's something not correct in using the headless-bios.c file for my device, but I think the serial consolr should be used. I wanted to know if it's possible to discard the headless console all together and direct everything thru the serial one. If that's not possible, then I will make the headless console to take over the serial port, even if it doesn't sound correct.

ghaerr:
The "serial console" in ELKS technically means something a little bit different than what you might need.

Regardless of what we call it, if the existing serial.c (which should be eventually be renamed serial-8251.c for the PC UART) won't work on your device, we have two options: either add a serial driver for your own device (serial-xxx.c), *and* setup serial console in ELKS. This has the potential problem of requiring /etc/inittab and /bin/init to be slightly updated to be your system's login devices, etc in the longer term. This has the benefit of being able to be used as an ELKS Serial device by other programs though. Another problem with this route is we will still need a "console", which in this case would be effectively a "null" console, headless-none.c, yet to be written.

OR, and possibly simpler for first the first round, would be just coding headless-xxx.c from scratch to do console I/O on your system's serial device, which will be /dev/console (and /dev/tty1).

ghaerr:
> I wanted to know if it's possible to discard the headless console all together and direct everything thru the serial one. 

So - yes. By recoding a "headless console" .c file to use your serial port.

> If that's not possible, then I will make the headless console to take over the serial port, even if it doesn't sound correct.

These two are pretty much the same options.

ghaerr:
> if it's possible to discard the headless console all together and direct everything thru the serial one.

A little more clarification: in ELKS, consoles, devices, and TTY drivers are different. The TTY driver allows a hardware device or console to be attached as a Linux-style TTY with termios control, etc. Thus, the TTY driver could be attached to a Direct Console, a BIOS Console, or a Headless Console. A TTY driver can also be attached to a PTY (pseudo tty) or a serial device driver.

ELKS needs a Console. When the config "serial console" option is selected, /dev/console is assigned such that a TTY driver is attached to a serial device, and then /dev/console is the serial port, but there is still an ELKS Console, which can be Direct, BIOS, or Headless, which is used for /dev/tty1. 

If you select Headless Console, and then write a serial driver within a headless console .c file, then the Console will be a serial port, and is automatically assigned to /dev/console. As mentioned above, this would not allow the serial device to be used to be attached to another TTY device.

I hope this makes sense, it's a little complicated!

cocus:
> OR, and possibly simpler for first the first round, would be just coding headless-xxx.c from scratch to do console I/O on your system's serial device, which will be /dev/console (and /dev/tty1).

I thought this would be the easiest route. I know that a serial tty doesn't mean a "serial console", and I should launch the login there as well, but I thought this could be changed somehow.


> If you select Headless Console, and then write a serial driver within a headless console .c file, then the Console will be a serial port, and is automatically assigned to /dev/console. As mentioned above, this would not allow the serial device to be used to be attached to another TTY device.

I completely understand, I was under the impression that I could disable the BIOS TTY, and redirect the main OS console to something like /dev/ttyS1 (which would be the driver I would write for this platform's serial port).

I think I will hack the serial port into the headless-console.c; probably creating a PC version and my own one for this platform.

ghaerr:
> I was under the impression that I could disable the BIOS TTY, and redirect the main OS console to something like /dev/ttyS1 (which would be the driver I would write for this platform's serial port).

You can; you would have to essentially create a headless-none.c that never returns any input and discards all output, and then write a serial-xxx.c driver for your platform, then set CONFIG_SERIAL_CONSOLE as well as adjust various settings in /etc/inittab, etc, which would essentially make /dev/console be /dev/ttyS1. (And no console).

> I think I will hack the serial port into the headless-console.c; probably creating a PC version and my own one for this platform.

Ok. I don't think its a great idea to implement a PC version for commit containing an 8251 UART driver, as then we'll have two serial driver files to maintain, and its already complicated with hardware FIFO support, 16450 support, etc. But creating one for your own use "headless-8251.c" in getting it working on a PC first before recoding for your own platform could help porting time. There would also be the confusing multiple options for PC headless console/serial console, etc.

For the headless console for your platform, name it headless-xxx.c (I forget your platform or chip name), and it should then be automatically selected based on your overall platform selection in menuconfig.

ghaerr:
After creating a serial headless console that works in your own system using headless-\<platform>.c, I can help going the first route by making some changes so that a headless-none.c console is a config option, and having the system auto-configure other files, and it should be pretty straightforward converting your code to a serial-xxx.c driver instead. 

That will allow future ELKS ports to embedded hardware to just write a serial driver and everything else works. We're not quite there yet!

cocus:
Hi @ghaerr , can you have a look at https://github.com/cocus/elks/tree/cocus/80c188-port? I've added the headless console for my board. I'm still working on trying to add the reception of the characters from the serial port, but I've spotted an issue on my irq file (I was trying to enable the IRQ of the serial port, and I forgot to map it as an irq number there).

ghaerr:
>  I've added the headless console for my board.

Looks good on the code, I haven't reviewed the config.in changes yet.

> I'm still working on trying to add the reception of the characters from the serial port

It might be simplest at first to use polling rather than interrupt drive for the serial reception. Take a look at kbd-bios.c, that shows how a kernel timer callback can be setup to poll your UART, prior to building an interrupt-driven version.

cocus:
> It might be simplest at first to use polling rather than interrupt drive for the serial reception. Take a look at kbd-bios.c, that shows how a kernel timer callback can be setup to poll your UART, prior to building an interrupt-driven version.

Indeed, but I would like to make this "boot" under emu86 and the settings I have for my board. In particular, booting without a bios.
That's the next part of my porting, which I don't know how to address. I will have to research how PC bios extensions work, so I can call/jump into them from a bootloader (which doesn't do that much, except initializing the chip selects, serial and such before jumping into ELKS code).

ghaerr:
The ICU and PTU code looks great!

How are you thinking of approaching the PIC EOI differences in irqtab.S:
```
//
//  Send EOI to interrupt controller
//
    cli         // Disable interrupts to avoid reentering ISR
    cmp $16,%ax
    jge was_trap    // Traps need no reset
    or  %ax,%ax     // Is int #0?
    jnz a4
//
//  IRQ 0 (timer) has to go on to the bios for some systems
//
    decw    _bios_call_cnt_l // Will call bios int?
    jne a4
    movw    $5,_bios_call_cnt_l
    pushf
    lcall   *_stashed_irq0_l
    jmp was_trap    // EOI already sent by bios int
a4:
    cmp $8,%ax
    mov $0x20,%al   // EOI
    jb  a6      // IRQ on low chip
/*
!
!   Reset secondary 8259 if we have taken an AT rather
!   than XT irq. We also have to prod the primay
!   controller EOI..
!
*/
    out %al,$0xA0
    jmp a5
a5: jmp a6
a6: out %al,$0x20       // Ack on primary controller
```

> In particular, booting without a bios.

Yes - you can just take the kernel timer callback code from kbd-bios.c and add that to your headless console, and skip any BIOS related stuff. Use the kernel timer callback to poll the UART.

ghaerr:
> That's the next part of my porting, which I don't know how to address. I will have to research how PC bios extensions work, so I can call/jump into them from a bootloader 

I think anything related to PC BIOS, including extensions, can be tossed. There is still a "setup" stub in front of the kernel, which is added using the kernel make. The source for this routine is in elks/arch/i86/boot/setup.S. You will gain control in the CONFIG_ROMCODE section. I would suggest taking a look at that, then hacking out all the BIOS code etc it calls, and special "INIT" data sections it sets up. That was the purpose of my adding SETUP_XXX in config.h in the last commit - you will define those to hard coded values in config.h and remove them from setup.S.

After your setup.S hack, we can look at ways of separating it into multiple files for a commit with no BIOS dependencies.

ghaerr:
> so I can call/jump into them from a bootloader (which doesn't do that much, except initializing the chip selects, serial and such before jumping into ELKS code).

The particular system-initialization code should probably go into another file alongside setup.S. This new code would gain control at power on, initialize chips, etc, and then jump to a label in setup.S that skips any BIOS-related initialization. We can then reorder some setup.S stuff so that any BIOS dependencies can be easily \#ifdef'd out, or moved into a seperate setup-bios.S file.

cocus:
I neglected the EOI completely. I'll work on that today. Since the startup code starts at the very end of the flash, and considering that the ELKS binary image doesn't consider that, I think my init code will be on a different project altogether for now, and jump into an appropriate offset into the ROM code.

cocus:
I neglected the EOI completely. I'll work on that today. Since the startup code starts at the very end of the flash, and considering that the ELKS binary image doesn't consider that, I think my init code will be on a different project altogether for now, and jump into an appropriate offset into the ROM code.

mfld-fr:
> I neglected the EOI completely.

This is precisely the point I am watching out :smirk: 

> my init code... jump into appropriate offset into the ROM code.

If reusing the existing ELKS kernel ROM config, you just have to load it @ E0000h (64 K), and make your init code to first call E000:0003h, then INT 19h. See first `exec` line in `emu86.sh` for Flash layout, and [rom-elks.c](https://github.com/mfld-fr/emu86/blob/master/rom-elks.c) for details about the BIOS extensions (extension header is just magic 0xAA55, and the checksum over 64K should be zero).

ghaerr:
Although the initial plan is to take PRs for new chip PIC/PIT support, etc. to get a particular platform to work, it looks like ELKS will probably need the ability to be separately configured and built, not just for particular chips, but also for users that want particular options like @mfld-fr's Advantech, who I think plans on using a (minimal) BIOS, and those like @cocus' 80c188, who seems to want to run pretty much on bare metal with no BIOS. This is just as well, as it will allow further splitting up the startup code in setup.S, which is still a bit of a mish-mash of stuff all crammed into a single file.

As discussion or PRs progress, we can discuss more exactly how the split-up should occur. I think it will be nice to have the ELKS dependencies (chip, port, BDA, setup, BIOS, boot) all sorted out. :)

mfld-fr:
I am sharing the same goal as @cocus : no more BIOS, even minimal. Just a bootstrap near FFFF:0h, a minimum setup of the pinmuxing / GPIO / chip select, and a jump to ELKS kernel entry point. And it also looks like we share the same kind of chip.

ghaerr:
> Just a bootstrap near FFFF:0h, a minimum setup of the pinmuxing / GPIO / chip select, and a jump to ELKS kernel entry point.

> Since the startup code starts at the very end of the flash, and considering that the ELKS binary image doesn't consider that, I think my init code will be on a different project altogether for now, and jump into an appropriate offset into the ROM code.

Can both of your issues be solved by creating a single flash image containing both ELKS and startup code, or will the power-on init code have to be flashed separately? I suppose the ROM base address could be adjusted and some filler space created by a directive for the linker could seperate the two, in some way, allowing a single image to be used.

mfld-fr:
As @cocus said, the startup code from FFFF:0h to ELKS entry point is very small and specific to each SBC. No benefit to share it, only burden.

ghaerr:
@cocus : 

> can you have a look at https://github.com/cocus/elks/tree/cocus/80c188-port?

On further review I suggest the following:
- On all inb/inb_p/outb/outw, etc: remove "(void *)" casting of 2nd parm, it is both unnecessary and undesirable. I removed these from ELKS code after you copied the files, I think.
- Remove "disable_irq" function entirely, since unused in ELKS
- Add your name to the ICU and PTU source files :)
- Change printk("Headless Console") to "801C88/R8810 Console", or something like that.
- Delete the following code:
```
#ifdef CONFIG_HW_259_USE_ORIGINAL_MASK	/* for example Debugger :-) */
     cache_08 = inw_p(CONFIG_80C18X_PCB_OFFSET + 0x08);
#endif
```

I'm not sure whether we should add a CONFIG_SYSCAP_CUSTOM in config yet, as there are likely quite a few other changes required for these ports. For now, I suggest keeping a section of config.h delineated by CONFIG_ARCH_80C18X that sets SYS_CAPS, along with the set of SETUP_XXX etc that the port needs. The existing SETUP_XXX can be delineated by CONFIG_ARCH_IBMPC. 

After we see all the changes required, we can decide on how we might configure it all in menuconfig. This will keep things quite a bit simpler for now, and not interfere with the rest of the working ELKS on IBM PC. 

Looking at config.h, it needs some reworking already, so lets start there with the configuration changes instead of config.in. In general, ELKS contains too many configuration options already, many rarely used, or committed long ago and not used. I want to think clearly about which exactly which options we add, so that we don't end up with tons of \#ifdef code. A lot of work has been done cleaning things up, and we're just now getting to the point of being able to corral all the ELKS dependencies (this project is started over 20 years ago)!

ghaerr:
@cocus,

Also, currently, because there are so many, most IN/OUT port assignments are being kept and configured in elks/include/arch/ports.h. I'm seeing that the existing IBM PC PIT TIMER_CMD port is in there, but the PIC command ports are not (I just created defines for the previous hard-coded values and they're still in irq-8259.c). The current naming scheme does not consider any platforms or chips not found on the IBM PC. 

So, we should probably continue and keep/move the base port (CONFIG_80C18X_PCB_OFFSET) into ports.h as well, and probably come up with a consistent naming convention across all platforms and devices. I'll hold off on moving any IBM PC PIC/PIT and other defines until your first PR, so as to not require any rebasing now.

After all this is done, we can consider whether to move the whole batch of embedded configuration options into config.in, or somewhere else.

So, to recap, we'll use config.h and ports.h for the time being for both embedded and IBM PC platforms as we sort out all the changes that need to be made, and configurable. This will keep all platforms operational through the process of getting ELKS running on its first embedded system.

Thank you'!

ghaerr:
Hello @mfld-fr,

You had mentioned in a post that there may be a problem with the 8810 that there may be an issue when sending an EOI from the general ELKS service routine, because the interrupt controller would need to know the specific interrupt to be acknowledged by the device having sent the interrupt. The following page from the AMD 80C188 (with integrated interrupt controller, likely compatible but I'm not completely sure) shows that there is a "non-specific" EOI that will acknowledge end of interrupt for whatever is being processed by the active service routine. It seems using this non-specific EOI command (rather than the alternative specific, which indicates the particular IS bit to acknowledge) should work in the ELKS _irqit routine.

Screen Shot 2021-05-14 at 7 16 08 PM

cocus:
Hi @ghaerr , thanks for all your comments. It's starting to become a quite large task to port ELKS! Thanks for your help.
In any case, right now I'll move remove the SYSCAP thing and set a SETUP_80C18X and SETUP_IBMPC if either the 80c18x or ibmp-pc are selected (respectively) to denote which SYSCAPs are being used.
Now, I'll address all the changes you've specified on your comment:
> * On all inb/inb_p/outb/outw, etc: remove "(void *)" casting of 2nd parm, it is both unnecessary and undesirable. I removed these from ELKS code after you copied the files, I think.
> * Remove "disable_irq" function entirely, since unused in ELKS
> * Add your name to the ICU and PTU source files :)
> * Change printk("Headless Console") to "801C88/R8810 Console", or something like that.
> * Delete the following code:
These seem trivial.
About the EOI, I'll have a deeper look at it, but if we can use the generic EOI bit, then we should be all set.

And regarding the Advantech board and its processor, and since @mfld-fr has discovered that the processor contains the same bitmapping of the registers, although on a different offset, we should define these on ports.h, probably relying on something like the previously discussed SETUP_XXXX. The code itself should be pretty much the same. There might be some cases where it's needed to branch off.

Finally, I agree with @mfld-fr that in our scenario, we just want a single image to flash on the eeprom/flash and call it a day. It would be really cool if the ELKS build could provide that by default. I know it's possible, but for this early stage of testing, a simple init code can be compiled outside, and placed at the very end of the flash image; allowing ELKS to boot. I'm not sure about the int 19 you've mentioned there. Without a BIOS, that's going to be difficult to achieve. Maybe jumping to a different address? Or setting up a dummy int 19 handler just so the code at 0xe000 is happy. (In the future we can even get modify this behavior for our platforms, so there's no overhead of a "hacked" bios extension loader, but rather just a baremetal ELKS loader).

ghaerr:
> It's starting to become a quite large task to port ELKS! 

If you want to just stick to writing the chip drivers, I can help through most of the rest.

I agree with your comments, but I think you've got my "SETUP_XXX" confused: in config.h, there are things like SETUP_VID_COLS, SETUP_VID_LINES, etc. These things are macros to a call to setupb(...) which read a setup.S data area. For your port, we don't want any of that, and these macros will just be set to the configured values we need for your port. I can/will suggest settings for each of them so you don't have to guess. If in doubt, just copy the existing macro for IBM PC, and we will update later. They are all set in setup.S, which at some point you'll have to look at. We plan on removing them from setup.S and just hard-coding them in config.h for your port, since there will be no BIOS.

So you might need to do a little re-arranging in config.h, but not much. If you want, I can post a PR for you that takes a first pass at this, or go for it if you don't mind the learning curve :)


> In any case, right now I'll move remove the SYSCAP thing and set a SETUP_80C18X and SETUP_IBMPC if either the 80c18x or ibmp-pc are selected (respectively) to denote which SYSCAPs are being used.

What I am saying here is not to use SETUP_IBMPC or SETUP_80C81X. Instead, we will use one of two global configuration settings (which you have already added to config.in) to include an entire set of options in config.h. These global configuration settings will be CONFIG_ARCH_IBMPC and the one you've just added CONFIG_ARCH_80C18X, that's it. So the config.h would end up looking something like:
```
#ifdef CONFIG_ARCH_IBMPC
/*
 * Setup data - normally queried by startup setup.S code, but can
 * be overridden for embedded systems with less overhead.
 * See setup.S for more details.
 */
#define SETUP_VID_COLS      setupb(7)       /* BIOS video # columns */
#define SETUP_VID_LINES     setupb(14)      /* BIOS video # lines */
#define SETUP_CPU_TYPE      setupb(0x20)    /* processor type */
#define SETUP_MEM_KBYTES    setupw(0x2a)    /* base memory in 1K bytes */
#define SETUP_ROOT_DEV      setupw(0x1fc)   /* root device, kdev_t or BIOS dev */
#define SETUP_ELKS_FLAGS    setupw(0x1f6)   /* flags for root device type */
#define SETUP_PART_OFFSETLO setupw(0x1e2)   /* partition offset low word */
#define SETUP_PART_OFFSETHI setupw(0x1e4)   /* partition offset high word */
#ifdef CONFIG_ROMCODE
#define SYS_CAPS    (CAP_PC_AT|CAP_DRIVE_PARMS)
#endif 
#endif   

#ifdef CONFIG_ARCH_801C8X
#define SETUP_VID_COLS 80
#define SETUP_VID_LINES 25
#define SETUP_MEM_KBYTES 256 (or whatever, etc)
...
#define SYSCAPS ...
#endif
```
 
> And regarding the Advantech board and its processor, and since @mfld-fr has discovered that the processor contains the same bitmapping of the registers, although on a different offset, we should define these on ports.h

Yes, just move CONFIG_80C18X_PCB_OFFSET to ports.h, along with any other I/O port values that are otherwise hard-coded for your port. We will move them to config.in later. This will keep the config.in changes to a minimum for now.

Other configuration settings, like CONFIG_80C18X_FCPU, or 80C18X variant configuration settings, can be put in config.h for now. This will also allow us/me to sort out what really needs configuration settings after the port, in which case config.in will be modified.


> Finally, I agree with @mfld-fr that in our scenario, we just want a single image to flash on the eeprom/flash and call it a day. It would be really cool if the ELKS build could provide that by default. I know it's possible, but for this early stage of testing, a simple init code can be compiled outside, and placed at the very end of the flash image

Actually, I think what @mfld-fr was saying is that we should NOT try to combine the startup code with ELKS, but instead keep the separately, and just jump to the ELKS entry point.

>  I'm not sure about the int 19 you've mentioned there. Without a BIOS, that's going to be difficult to achieve.

You don't need INT 19h or any of this with no BIOS. @mfld-fr was talking about how the BIOS option ROMs work, but it looks like neither of you want to go that route.

Hang in there, by reducing the load of having to set all the configuration options up front in config.in, and just placing that stuff in ports.h and config.h for now, you can concentrate on just the required chip functionality. You're doing great!
 
Thank you!

cocus:
Oh I wasn't aware about these setupX() functions. 
In any case I forgot to mention something about the meaning of CONFIG_80C18X_PCB_OFFSET. This offset is actually not a "fix" for the differences we have between the 80C188XL and EB; but rather the actual offset to the PCB. By default this is set to 0xff00, and it can be moved by writing to a fixed IO word.
Then the problem arises when accessing the peripherals registers inside the PCB, for instance, the Timer0 control register is located at PCB_OFFSET + 0x56 for the -XL variants, but on PCB_OFFSET + 0x36 on the -EB variants. I think this is the "main" differences between the 80C18XEB/XL and @mfld-fr's 8810.
The fact that I've created a config for that is mostly for educational purposes. There's NO real need to relocate the PCB out of its original offset, but it's good to have that option. The assembly startup should move the PCB to the same offset as the one configured on that config. In any case, do you want to keep this config, or shall we assume that the system is going to hardcode this to 0xff00?

mfld-fr:
> I'm not sure about the int 19 you've mentioned there.

No need to implement INT 19h. ELKS pre-init code that you first call @ E000:3h sets up the vector, so you just have to do INT 19h after that call. The ELKS handler will then branch into the actual ELKS init.

```
start_elks:
    CALLF  E000:3h
    INT 19h  ; never returns
```

mfld-fr:
> In any case, do you want to keep this config, or shall we assume that the system is going to hardcode this to 0xff00?

Let us keep the things simple : default PCB mapping is I/O FF00h, and there is no need to move it.

ghaerr:
Hello @cocus,

> In any case I forgot to mention something about the meaning of CONFIG_80C18X_PCB_OFFSET. This offset is actually not a "fix" for the differences we have between the 80C188XL and EB; but rather the actual offset to the PCB. By default this is set to 0xff00, and it can be moved by writing to a fixed IO word.

Thank you for the info. I agree with @mfld-fr, lets keep it simple and use default mapping of FF00h.


> Then the problem arises when accessing the peripherals registers inside the PCB, for instance, the Timer0 control register is located at PCB_OFFSET + 0x56 for the -XL variants, but on PCB_OFFSET + 0x36 on the -EB variants. I think this is the "main" differences between the 80C18XEB/XL and @mfld-fr's 8810.

I see. Geez! There are a couple ways of solving this, I suggest just getting your system working first, then we or @mfld-fr can update this for his system or the differences. One way to do that later would be defines for each register in a new ports-8018x.h file, another would be an ifdef in the timer-8018x.c file for the variant based on a as-yet-unnamed config option.

Thank you!

cocus:
Perfect! I've been busy with my life, so I couldn't update my port, but I'll gladly accept any help from you guys. In any case, I'll keep pushing changes without trying to be "port-friendly" for now, so we can just get it working. Thanks again!

cocus:
@mfld-fr please have a look at this skeleton "loader" that I've created for my board: https://github.com/cocus/8018x-elks-ldr
I would love to add a small printout that says "LDR" or something like that, and then make it more generic so it can be easily used with your board as well. If you have spare time, I would gladly accept changes for it.
Thanks

mfld-fr:
Yes, sure, but let us continue the discussion about the portage in the related issue #932, because this PR is now closed.

PR #938, opened 2021-05-15
[config] Prepare config.h for 8018X port

ghaerr:
@cocus : 

These are the changes you need in config.h for the conversation in https://github.com/jbruchon/elks/pull/937#issuecomment-841586363.

Let me know if you'd like me to commit this, or if you would rather just add this to your branch.

Thank you!

### DIFF
--- a/elks/include/linuxmt/config.h
+++ b/elks/include/linuxmt/config.h
@@ -4,6 +4,7 @@
 #include <autoconf.h>
 #include <linuxmt/major.h>
 
+#ifdef CONFIG_ARCH_IBMPC
 /*
  * Setup data - normally queried by startup setup.S code, but can
  * be overridden for embedded systems with less overhead.
@@ -17,6 +18,22 @@
 #define SETUP_ELKS_FLAGS	setupw(0x1f6)	/* flags for root device type */
 #define SETUP_PART_OFFSETLO	setupw(0x1e2)	/* partition offset low word */
 #define SETUP_PART_OFFSETHI	setupw(0x1e4)	/* partition offset high word */
+#ifdef CONFIG_ROMCODE
+#define SYS_CAPS	(CAP_PC_AT|CAP_DRIVE_PARMS)
+#endif
+#endif
+
+#ifdef CONFIG_ARCH_801C8X
+#define SETUP_VID_COLS		80	/* BIOS video # columns */
+#define SETUP_VID_LINES		25	/* BIOS video # lines */
+#define SETUP_CPU_TYPE		5	/* processor type 80186 */
+#define SETUP_MEM_KBYTES	256	/* base memory in 1K bytes */
+#define SETUP_ROOT_DEV		0x0600	/* root device ROMFS */
+#define SETUP_ELKS_FLAGS	0	/* flags for root device type */
+#define SETUP_PART_OFFSETLO	0	/* partition offset low word */
+#define SETUP_PART_OFFSETHI	0	/* partition offset high word */
+#define SYS_CAPS	0
+#endif
 
 /*
  * System capabilities - configurable for ROM or custom installations.
@@ -38,7 +55,6 @@
 #define DEF_SYSSIZE	0x2F00
 
 #ifdef CONFIG_ROMCODE
-#define SYS_CAPS	(CAP_PC_AT|CAP_DRIVE_PARMS)
 #define SETUP_DATA	CONFIG_ROM_SETUP_DATA
 
 #ifdef CONFIG_BLK_DEV_BIOS    /* BIOS disk driver*/
--- a/elks/include/linuxmt/config.h
+++ b/elks/include/linuxmt/config.h
@@ -32,7 +32,10 @@
 #define SETUP_ELKS_FLAGS	0	/* flags for root device type */
 #define SETUP_PART_OFFSETLO	0	/* partition offset low word */
 #define SETUP_PART_OFFSETHI	0	/* partition offset high word */
-#define SYS_CAPS	0
+#define SYS_CAPS		0	/* no XT/AT capabilities */
+
+#define CONFIG_80C18X_FCPU	20
+#define CONFIG_80C18_EB
 #endif
 
 /*
--- a/elks/include/linuxmt/config.h
+++ b/elks/include/linuxmt/config.h
@@ -23,7 +23,7 @@
 #endif
 #endif
 
-#ifdef CONFIG_ARCH_801C8X
+#ifdef CONFIG_ARCH_8018X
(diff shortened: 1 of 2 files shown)


cocus:
In this case, you're setting the 80C18X values, in a harcoded fashion. In this case, I should only leave the config.in with the 80C18X arch config, right? FCPU and the model variant are set here (for now).

elks/include/linuxmt/config.h
```diff
@@ -4,6 +4,7 @@
 #include <autoconf.h>
 #include <linuxmt/major.h>
 
+#ifdef CONFIG_ARCH_IBMPC
```
mfld-fr, line 7: ARCH is already used for CPU level architecture. Suggest to use PLATFORM for chipset / SoC level architecture, and TARGET for target level architecture.

mfld-fr: In @cocus case : ARCH=8018X, PLATFORM=8018XEB, TARGET=OR556. In my case, ARCH=8018X, PLATFORM=R8810, TARGET=ADVTECH. For IBM PC : ARCH=808X, PLATFORM=8XXX, TARGET=PC. User should just have to select the target in the configuration, no more.

elks/include/linuxmt/config.h
```diff
@@ -17,6 +18,25 @@
 #define SETUP_ELKS_FLAGS	setupw(0x1f6)	/* flags for root device type */
 #define SETUP_PART_OFFSETLO	setupw(0x1e2)	/* partition offset low word */
 #define SETUP_PART_OFFSETHI	setupw(0x1e4)	/* partition offset high word */
+#ifdef CONFIG_ROMCODE
+#define SYS_CAPS	(CAP_PC_AT|CAP_DRIVE_PARMS)
+#endif
+#endif
+
+#ifdef CONFIG_ARCH_801C8X
```
mfld-fr: Avoid using the 'C' middle letter. The fact that we are using the CMOS version of the MCU has no matter in the port, until some advanced power management is implemented.

mfld-fr (commented):
In a perfect world, all the SETUP_* stuff should not exist. As soon as the init code places the kernel segments at the right place in memory, all the hardware & FW probing should reside in the drivers, and no setup() should be used at all. If changing legacy code would be painful, let us take the opportunity of the portage to make the things right in the new code.

ghaerr:
> In a perfect world, all the SETUP_* stuff should not exist. As soon as the init code places the kernel segments at the right place in memory, all the hardware & FW probing should reside in the drivers, and no setup() should be used at all. 

Agreed - I have been working towards the goal of removing most of setup.S for the past year. Cleaning up the mess that has been setup.S has been complicated by the fact that a single file has been used to try to maintain ROM, BIOS, BDA, config, relocation, boot data passalong and other code. Add to that, most of the drivers have been written with arbitrary peeking/poking, grabbing setup data, etc without a cohesive framework. Much of this is completed, but we need to take things incrementally to ensure our path forward to bare-metal embedded keeps running on hard-to-test legacy systems. I have a plan for further cleanup, and will take another pass at this.

In the first (temporary) increment, I would like to try to get @cocus and @mfld-fr get a kernel booting without having to make major changes to ELKS source files (except for new chip drivers). We will then have a better grasp of the overall changes that have been required, and can make wise decisions on how to reorganize the source.

elks/include/linuxmt/config.h
ghaerr: Thanks for the naming ideas. Good start, but needs more discussion.

> ARCH is already used for CPU level architecture.

[EDIT: I misunderstood your statement - ARCH alone is and should be used for CPU architecture.]

CONFIG_ARCH_IBMPC is now used for platform in the config.in configuration system to select the highest-level options for the IBMPC vs SIBO platforms. Given one or the other, a multitude of other options are displayed or not. CONFIG_ARCH_IBMPC has not been used in the code (and is only temporarily being suggested for use in config.h, to keep things simple and delay changes in the configuration system until a solid design and naming scheme is agreed on.


> Suggest to use PLATFORM for chipset / SoC level architecture, and TARGET for target level architecture.
In @cocus case : ARCH=8018X, PLATFORM=8018XEB, TARGET=OR556. In my case, ARCH=8018X, PLATFORM=R8810, TARGET=ADVTECH. For IBM PC : ARCH=808X, PLATFORM=8XXX, TARGET=PC. 

Good ideas, but I'm not sure the need for three separate hierarchies/selections. Let me look through the source and think about this a bit more.

> User should just have to select the target in the configuration, no more.

Agreed, although it is not fully known yet which sub-items should be configurable within a .in vs .h file.

elks/include/linuxmt/config.h
ghaerr: Agreed, these are in use in @cocus' tree. I'll make the change to CONFIG_ARCH_8018X (at least for now) and suggest @cocus change new driver files to timer-8018x.c and irq-8018x.c.

ghaerr:
As @mfld-fr pointed out above, CONFIG_ARCH_IBMPC is currently being used as a platform, but is labeled as an architecture. Instead of reorganizing that now, let us delay to minimize config.in changes for the current porting effort. Likely permanent name will be CONFIG_TARGET_IBMPC.

Changed config option to CONFIG_ARG_8018X for CPU architecture for current porting effort.

Next step is to introduce some changes in irqtab.S for separating between IBMPC and 8018X to allow @cocus easier submission of changes for port.

ghaerr:
Hello @cocus,

> In this case, you're setting the 80C18X values, in a harcoded fashion. In this case, I should only leave the config.in with the 80C18X arch config, right? FCPU and the model variant are set here (for now).

Yes. I think will be easier for your port to just hardcode values in config.h, and only make single addition to config.in of CONFIG_ARCH_8018X for now, and no other config.in changes (for now). Sorry for that news, but probably better to allow you to proceed with porting effort while we all come to decisions on how to name/reorganize ARCHs, platforms, targets, etc.

I plan on committing this PR soon so you can continue making progress on your port.

Thank you!

ghaerr:
Committing now for incremental improvement for 8018X ease of porting. No change to any code produced for IBM PC.

CONFIG_ARCH_IBMPC being (temporarily) used to seperate unmodified PC-specific code from required changes for 8018X port in irqtab.S and config.h.

Noted @mfld-fr's comments on the need for ARCH, PLATFORM and/or TARGET, but being delayed so as to keep progress on @cocus 8018X port moving forward.

elks/include/linuxmt/config.h
cocus: No worries. I was using that name because that's the name of the processor I have :) but @mfld-fr is completely right.

PR #940, opened 2021-05-15
[kernel] Add dynamic interrupt handling

mfld-fr:
Following the [discussion about ELKS weakness when taking over any ongoing & fast interrupt](https://github.com/mfld-fr/emu86/issues/41#issuecomment-832074696).

First step : allocate the interrupt handler on the heap and check it works, before removing all the static handlers from `irqtab.S`. Kernel code increased from 50928 to 51024 in `emu86-disk.config`. Will decrease after the static handlers removal in the next step.


elks/arch/i86/kernel/irq.c
```diff
@@ -91,10 +104,24 @@ int request_irq(int irq, void (*handler)(int,struct pt_regs *,void *), void *dev
...
+	// that redirects to generic handler
+
+	int_handler_s * h = (int_handler_s *) heap_alloc (sizeof (int_handler_s), HEAP_TAG_INTHAND);
+	if (!h) return -ENOMEM;
+
+	h->call = 0xE8;    // CALL
+	h->proc = _irqit;  // generic interrupt handler
+	h->irq  = irq;
+
+	int_vector_set (irq + ((irq > 8) ? 0x70 : 0x08), h);
```
mfld-fr: Small bug here : replace 0x70 by 0x68.

ghaerr:
Well, you beat me to it on this one :)

How tested is this?

This will affect a lot of the complexity in ports.h previously as compile-time options, now become run-time options, which is nice enhancement. I presume all the CONFIG_NEED_IRQx options in ports.h will be removed, so I need some time to look through each of the drivers to ensure this plan will work without hiccup.

```
extern void int_vector_set (int vect, void * proc);
extern void _irqit (void);
```
Please move these forward declarations to elks/include/arch/irq.h, with 'extern' removed. I also prefer to copy the forward C declaration into the ASM routine as a comment, for reference.

Also, there is a bit of tricky critical section that is changing with this commit: previously, interrupts were off through enable_irq until the last restore_flags call (effectively the entire routine). Now, interrupts are enabled in int_vector_set, prior to the call to enable_irq. I don't think this will be an issue on ELKS with real hardware, but it could be an issue with EMU86 since the 8259 PIC interrupt mask bits are not emulated, depending on exactly how enable_irq is implemented. Perhaps int_vector_set should not change interrupts, or at least reset them in the same manner as restore_flags. To guarantee compatibility, interrupts would be left off until the final call to restore_flags, which would need to be moved to after enable_irq.

ghaerr:
Also, did you really want to make the `../emu86` changes in emu86.sh? I use a temporary `./es` copy of emu86.sh for testing, so as to not have to modify emu86.sh all the time.

ghaerr:
@mfld-fr :  

There is another little snag in moving to dynamic IRQ handling (in irqtab.S):
```
#if defined(CONFIG_NEED_IRQ3) || defined(CONFIG_FAST_IRQ3)
_irq3:                  // COM2
#ifdef CONFIG_FAST_IRQ3
        .global _irq_com2       // fastest possible interrupt routine, fixed com2 port
        jmp     _irq_com2
        .byte   3               // ignored
#endif
#ifdef CONFIG_NEED_IRQ3
        call    _irqit          // original ELKS overhead interrupt routine
        .byte   3
#endif
#endif
#if defined(CONFIG_NEED_IRQ4) || defined(CONFIG_FAST_IRQ4)
_irq4:                  // COM1
#ifdef CONFIG_FAST_IRQ4
        .global _irq_com1       // fastest possible interrupt routine, fixed com1 port
        jmp     _irq_com1
        .byte   4               // ignored
#endif
#ifdef CONFIG_NEED_IRQ4
        call    _irqit          // original ELKS overhead interrupt routine
        .byte   4
#endif
#endif
```
A while back, for a variety of reasons, a "fast" serial I/O handler was implemented to achieve the fastest serial input possible. This required an interrupt handler routine that bypassed the _irqit mechanism. In order to keep that working, it looks like we may need a parameter passed to request_irq that indicates whether to use the default `_irqit` service routine, or, say `_irq_com1`. I suggest perhaps using 0 to mean use _irqit, otherwise the passed handler routine.

Looking further, it looks like all cases of request_irq use NULL as the third parameter, perhaps that should be used as the hw irq entry point.

mfld-fr:
Before processing the special case of the fast interrupt handlers, I have to solve a problem that makes this first step proposal to fail : the dynamic handler is allocated on the heap, so in the data segment, while the `int_vector_set()` uses CS... it worked only by chance because there was another bug that made the timer vector unchanged and still pointing to the static handler.. :disappointed:

ghaerr:
Hello @mfld-fr, 

Another subtle point to think about for moving to dynamic IRQ: in the original code, the configuration specifies all interrupts that may possibly be handled by the kernel, or are installed in the configured platform. Early in startup, and while interrupts are disabled, all these vectors are initialized, and end up go through the default handler, which prints "Unexpected interrupt".  With this PR as-is, there will be a period where interrupts will be enabled, but various device drivers, such as NE2K, will not have been initialized. What happens when a hw interrupt occurs between these points? Do we need to have the low-core hardware interrupt vectors initialize to a handler anyways (prior to a call to request_irq)? If not, system crash could result. If so, then we're going to have to leave all the CONFIG_NEED_IRQx tables in place. If all the tables remain in place, it is questionable what the benefit of dynamic IRQ handler is.

[EDIT: Also, dynamic IRQ handling is a big change and needs to be considered from the point of view of BIOS-based systems, where certain vectors may need to be specified in a configuration to be handled, as well as the new embedded systems, whose operation I am not that familiar with. Perhaps this entire change should hold until after first port of embedded system @cocus is working on, since it is not required for embedded port. ? ]

Thank you!

mfld-fr:
Problem solved by adding some segments in the play. Also included some of your points. Now let me remove the static handlers to show you how much it changes the game at source level...

Kernel code size before the cleanup : 51056.

ghaerr:
> it worked only by chance because there was another bug that made the timer vector unchanged and still pointing to the static handler.

Exactly what bug is this? Existing bug in ELKS, or bug in first commit to this PR?

mfld-fr:
Looks like `meminfo` should be also updated to display the dynamic handlers allocated on the heap.

Edit: now kernel code is 50992, so +74 bytes for a big source simplification.

elks/arch/i86/kernel/irq.c
```diff
@@ -72,6 +73,42 @@ static void default_handler(int i, void *regs, void *dev)
...
+static int int_handler_add (int irq, int vect)
+	{
+	int_handler_s * h = (int_handler_s *) heap_alloc (sizeof (int_handler_s), HEAP_TAG_INTHAND);
+	if (!h) return -ENOMEM;
+
+	h->call = 0x9A;    // CALLF opcode
+	h->proc = _irqit;  // generic interrupt handler
+	h->irq  = irq;
+
+	h->seg = int_vector_set (vect, (int_proc) h, seg_data ());
```
ghaerr: Don't use another function wrapper `seg_data()`. Use existing `kernel_ds`, as used all over elsewhere in kernel.
Consider why even pass argument at all, since almost all other kernel ASM uses existing DS for accessing kernel data.

elks/arch/i86/kernel/irq.c
```diff
@@ -91,6 +128,9 @@ int request_irq(int irq, void (*handler)(int,struct pt_regs *,void *), void *dev
     action->handler = handler;
     action->dev_id = dev_id;
 
+	if (int_handler_add (irq, irq + ((irq > 8) ? 0x68 : 0x08)))
```
ghaerr: Passing hard-coded low-core interrupt table addresses re-introduces architecture-specific into this routine, and we just removed all that. (for this call and system call 0x80 below). IMO, `int_handler_add` should only be passed a single parameter - `irq`, and `int_vector_set` should handle architecture-specific low-memory interrupt address from passed irq.

elks/arch/i86/kernel/irq.c
mfld-fr: 1. Global variable `kernel_ds` is today questionnable, as it resides in a code segment (moreover, the code segment is not the same for all the functions as you switched to somewhat a medium model). I already complained a lot about this point, and asked many times to move that variable in the well known INIT segment. So I prefer to wrap it to not be impacted by the move of this variable (one day ?), at the cost of a very few bytes.

2. You need this segment argument for installing your 'fast handlers', as such handlers are not on the heap so in the data segment, but in the code segment. But the question is : 'what code segment ?', again because of the medium memory model...

elks/arch/i86/kernel/irq.c
```diff
@@ -72,6 +73,42 @@ static void default_handler(int i, void *regs, void *dev)
...
+
+// Add a dynamically allocated handler
+// that redirects to the generic handler
+
+static int int_handler_add (int irq, int vect)
+	{
+	int_handler_s * h = (int_handler_s *) heap_alloc (sizeof (int_handler_s), HEAP_TAG_INTHAND);
+	if (!h) return -ENOMEM;
+
+	h->call = 0x9A;    // CALLF opcode
```
ghaerr: 0x9A adds architecture-specific code back into this routine. Consider calling `int_vector_set(irq, h)` and having it handle all architecture-specific code in irqtab.S, which is not cleaned up. If rather stay some in C, then this routine needs to move to irq-8259.c (except then it needs duplicating in irq-8018x.c, so we're not solving our cleanup.)

Also, now every interrupt costs extra far call...

elks/arch/i86/kernel/irq.c
mfld-fr: Agree, the `irq-8859.c` should provide an IRQ to vector mapping, as it already provides the IRQ remapping.

elks/arch/i86/kernel/irq.c
mfld-fr: You are in the folder `arch/i86`, mister :laughing:

mfld-fr: Yes, one more segment push & pop on each interrupt... wow... what a big performance impact :smirk: !

elks/arch/i86/kernel/irq.c
ghaerr: > * Global variable `kernel_ds` is today questionnable, as it resides in a code segment

No - you are confusing `kernel_ds`, which is used everywhere, and `ds_kernel`, which is used in irqtab.S.
Even if they were the same, I don't agree with changing only in one place and leaving everything else. That's how the kernel got so messy in the first place over 20 years.

mfld-fr: Oup's... yes, sorry, got confused... will replace `seg_data()` by `kernel_ds`. But still expecting `ds_kernel` to move to the INIT segment one day...

elks/arch/i86/kernel/irq.c
ghaerr: Hey, it all adds up. I am still considering what is the true benefit of using dynamic approach, as well as configuration issue brought up in https://github.com/jbruchon/elks/pull/940#issuecomment-841700174. What do you think? This approach, which I am for, but only cautiously, is already bigger than anticipated by both of us, and certainly will have to go through hardware and device testing before commit.

elks/arch/i86/kernel/irq.c
ghaerr: Hey I'm not fighting you, just trying to keep everything straight and working over here! 😏 There's still tons to do to clean ELKS up, I agree with you there.

ghaerr: >  But still expecting `ds_kernel` to move to the INIT segment one day...

You mean the "setup" INITSEG? I thought, and agree with, your comments from earlier today in https://github.com/jbruchon/elks/pull/938#pullrequestreview-660324248 that all INIT probing code (and data) should be removed eventually. That means almost no need for INITSEG, except for boot-parameter passing. I would like to see INITSEG go away, and my last PR #941 takes another step in that direction.

elks/arch/i86/kernel/irqtab.S
```diff
@@ -433,7 +217,8 @@ save_regs:
 	pop	(%si)		// DI
 	pop	2(%si)		// SI
 	pop	8(%si)		// DS
-	pop	%di		// Pointer to interrupt number
+	pop	%di		// Return offset is actually a pointer to the IRQ
+	pop	%ds		// Return segment of the dynamic handler = kernel DS
```
ghaerr, line 194: I don't understand how the above `pop %ds` works. It isn't necessarily guaranteed that _irqit is only called by int_vector_set, which is passed a seperate 'kernel_ds' value anyways. Should pop %ds not be kernel DS, lines 223-226 will badly fail. Is this only being done to get DS for 244? Line 168 already sets DS = kernel_ds anyways. 

If Line 168 setting kernel DS is incorrect because we're executing in heap-allocated data segment, then Line 172 comparing _gint_count will fail also. So I don't get it.

mfld-fr:
To be clear on the approach, and in a few words:

* If an interrupt was set up and is handled by the BIOS (or by any preceding FW), and ELKS has no driver to handle it, it should left it as is, to let the initial things continue to work (remember: the 'welcome music'). No more need to enable it again, like the keyboard one already used by the BIOS to fill the keyboard buffer. In a next step, I will remove the initial masking of all the interrupts (DONE in 1497894).

* If an interrupt was not set up before ELKS, be sure that the BIOS has already disabled its source, and masked it, to avoid any crash before ELKS. Moreover, many devices have interrupt disabled by default.

*  In both cases, whenever ELKS has a driver to take over, the transition will be fully controlled by the ELKS driver : mask the interrupt, reset the device (so disable the source), replace the vector, setup the device (so enabling the source), unmask the interrupt. In the big brother of ELKS, this is exactly how things are going on, especially with the unloadable modules, that also leave the interrupt masked, the source disabled, and the device in a clean state.

* The default handler has no meaning now, so I will also remove it (DONE in d4218d7).

elks/arch/i86/kernel/irqtab.S
mfld-fr: * Nothing else than a dynamic handler would call `_irqit`.
* Overwriting the same value in DS is a trick to avoid using another register, and for a minimal impact on a critical routine.

ghaerr: So the question really is: what benefit really is there to dynamic IRQ handling, since we're now at 3 extra instructions per interrupt. If problem is to reduce source clutter, perhaps just removing all CONFIG_IRQ_x options and always maintaining non-ifdefed table of `_irqtable` entry points in existing kernel code segment solves that problem. Then your routine can dynamically set handler within CS:_irqtable and point low-core vector to it . At this point, easier design, faster, and less messy. That method still has its problems though, we lose configurability as to which low-memory entry points are patched, which needs discussion. If we keep _irqtable as CS entry point and patch low-memory at runtime, perhaps that's the best answer.

elks/arch/i86/kernel/irq.c
mfld-fr: The fact is that the INITSEG is the only fixed anchor in the memory layout, apart the BDA and the vector table, and able to write to... plus the fact that the init code installs (and may decompresses) the kernel segments in memory, so it knowns very well where is the kernel code segment. But any data that is not related to this memory installation is still to be wiped out...

elks/arch/i86/drivers/char/serial.c
```diff
@@ -432,15 +436,19 @@ static void rs_init(void)
...
 	    default:
-#if defined(CONFIG_NEED_IRQ4) || defined(CONFIG_NEED_IRQ3)
+#if !defined(CONFIG_FAST_IRQ3) && !defined(CONFIG_FAST_IRQ43)
 		request_irq(sp->irq, rs_irq, NULL);
 #endif
 		break;
 #ifdef CONFIG_FAST_IRQ4
 	    case 4:
+		// FIXME: how to find code segment of the fast handler ?
+		int_vector_set (0x0C, _irq_com1, /*seg_code()*/);
```
ghaerr: This is getting far too messy. We don't need or want another, second way of specifying an interrupt handler for an IRQ number. Especially one where one needs to know the interrupt vector number separately from the irq number.

Proper way to handle all of this would be to use `request_irq(sp->irq, ... )` and pass a __far pointer to the interrupt handler routine; everything else is handled under the hood.

But we don't need to pass far pointer because I now realize that better method is to leave _irqit alone and keep a full copy (non-ifdef'd) of the `_irqtab` table, and have under-the-hood write that table as well as low-memory vector. This achieves any benefit of dynamic IRQs without any risk of meddling in well-tested codepaths.

ghaerr:
> To be clear on the approach, and in a few words:

In general, I agree with your first three points. However, testing done by the community over the last year on all sorts of old hardware has shown that many assumptions we have made regarding how BIOS's "work" have not been true, and that making seemingly simple changes has stopped legacy hardware from running ELKS. For instance, it is not necessary certain that a BIOS has enabled the IRQ 1 keyboard interrupt. Of course, it should have. So I hesitate to start removing code or changing things in the name of making things better, without being able to show prior working systems still run.

> In both cases, whenever ELKS has a driver to take over, the transition will be fully controlled by the ELKS driver : mask the interrupt, reset the device (so disable the source), replace the vector, setup the device (so enabling the source), unmask the interrupt. In the big brother of ELKS, this is exactly how things are going on, especially with the unloadable modules, that also leave the interrupt masked, the source disabled, and the device in a clean state.

In contrast to my above statement, the starting and stopping of ELKS device drivers is something we can completely control. So I think this would be a good first goal to consider, that of inspecting and improving the entire set of ELKS drivers, with a simple API, that accomplishes your points. This may want to occur before the transition to dynamic IRQ handling. Although ELKS doesn't yet support it, the idea of cleanly starting up and stopping a driver is a good one.

elks/arch/i86/kernel/irq.c
mfld-fr: Done in cbe3e07

elks/arch/i86/kernel/irq.c
mfld-fr: Done in bfdabb7

elks/arch/i86/kernel/irqtab.S
mfld-fr: > since CS now points to heap data

:interrobang:

ghaerr: Look at the beginning of the routine!!! CS points to heap data:
```
_irqit:
 //
//	Make room
//
	push	%ds
	push	%si
	push	%di
//
//	Recover data segment
//
//	seg	cs
	mov	%cs:ds_kernel,%ds     <----- CS points to heap data, so this loads DS with garbage.
//
//	Determine which stack to use
//
	cmpw	$1,_gint_count   <---- this now compares a bogus word somewhere in memory
                              // kernel will now make incorrect decision on which stack to run on
                              // later incw _gintr_count at saveregs trashes memory from bad DS, etc
```
If DS was set correctly at the beginning of the routine, why would you need your additional `pop %ds`?
⁉️⁉️

ghaerr: I get it now, sorry I missed that `int_vector_set` returns CS which is used to set the far call segment in the allocated trampoline. And the pop ds is required since far call was used to get to _irqit. Tricky.

Still uses 3 extra instructions per interrupt, forces code execution in data segment, and requires heap allocation which may fail. Using static _irqtable in code segment solves all those problems in better way, except that of compiled-in _irq_com1/2 fast serial driver.

Thus, to truly provide benefit, I suggest not using klugy `int_vector_set` secondary API call from serial driver routine, but instead use request_irq for fast serial drivers as well, but enhanced to take handler routine as far pointer with a flag to indicate whether or not to trampoline through _irqit or not. The third parameter to request_irq is currently unused and should be used as flag instead.

In the current kernel, there are only four calls to request_irq, so this PR is a lot of work and change without yet much benefit. If request_irq was required to be used from all drivers (not int_vector_set, etc), then option could be added that would allow both fast and slow serial drivers to be compiled in, for instance, and either one to be dynamically selected at runtime, even on ROM system. Using static _irqtable would not work in this case, resulting in a real benefit to this work. This would also provide your own stated goal of ensuring that drivers go through a standard process for gaining access to interrupts: use request_irq only. `enable_irq` could be removed in routines already calling request_irq.

elks/arch/i86/kernel/irq.c
```diff
@@ -58,19 +43,43 @@ static struct irqaction irq_action[] = {
 void do_IRQ(int i,void *regs)
 {
     register struct irqaction *irq = irq_action + i;
-
-    irq->handler(i, regs, irq->dev_id);
+    if (irq) irq->handler(i, regs, irq->dev_id);
```
ghaerr: Add back in print "Unexpected interrupt: %d" in the case where irq == NULL. This is still useful for debugging device drivers.

Remove irq->dev_id, pass NULL instead. Always unused.

elks/arch/i86/kernel/irq.c
```diff
@@ -32,24 +34,7 @@ struct irqaction {
 	void *dev_id;
```
ghaerr: Delete dev_id, always unused and delete struct irqaction, as what we use now is just array of handler pointers.

elks/arch/i86/drivers/char/serial.c
```diff
@@ -432,15 +436,19 @@ static void rs_init(void)
...
 	    default:
-#if defined(CONFIG_NEED_IRQ4) || defined(CONFIG_NEED_IRQ3)
+#if !defined(CONFIG_FAST_IRQ3) && !defined(CONFIG_FAST_IRQ43)
 		request_irq(sp->irq, rs_irq, NULL);
 #endif
 		break;
 #ifdef CONFIG_FAST_IRQ4
 	    case 4:
+		// FIXME: how to find correct code segment of the fast handler ?
+		int_vector_set (0x0C, _irq_com1, kernel_cs);
```
ghaerr: Replace int_vector_set with request_irq(sp->irq, _irq_com1, 1) using changed third parameter as flag indicating no trampoline used. 

If you don't want to pass far pointer to handler now, that's OK, we can simplify things and limit interrupt routines to be in the near kernel code segment for now, which is same limitation as all other handler routines, including fast _irq_com1.

Remove enable_irq on line 453.

Regarding runtime switching between fast and slow serial driver: this first version won't allow that since we don't have way to de- or re-allocate vectors. But no problem, still could have benefit of compiling both in, then having runtime option of selecting one or other at system start, using /bootopts or another method. This will result in big benefit for serial network testing, which is reason for fast driver anyways.

elks/arch/i86/kernel/irq-8259.c
```diff
@@ -54,6 +54,16 @@ int remap_irq(int irq)
...
 }
 
+// Get interrupt vector from IRQ
+
+int irq_vector (int irq)
+	{
+	// IRQ 0-7  are mapped to vectors INT 08h-0Fh
+	// IRQ 8-15 are mapped to vectors INT 70h-77h
+
+	return irq + ((irq > 8) ? 0x68 : 0x08);
```
ghaerr: Shouldn't this be ... (irq >= 8)... ?

mfld-fr: Fixed in 990a917

elks/arch/i86/kernel/irq.c
mfld-fr: Done in 990a917

elks/arch/i86/kernel/irq.c
mfld-fr: Done in 990a917

elks/arch/i86/kernel/irqtab.S
mfld-fr: Here is another benefit : `ds_kernel` could now be dropped, as DS is already pushed by the CALLF in the dynamic handler. See commit 990a917. Still have to complete the work on the fast handlers as you told.

mfld-fr:
> Also, did you really want to make the `../emu86` changes in emu86.sh? I use a temporary `./es` copy of emu86.sh for testing, so as to not have to modify emu86.sh all the time.

Reverted.

mfld-fr:
> Please move these forward declarations to elks/include/arch/irq.h, with 'extern' removed.

Moved.

elks/arch/i86/kernel/irq.c
mfld-fr: Discussion about `ds_kernel` has no more matter, as that variable could be now merely dropped.

elks/arch/i86/kernel/irq.c
mfld-fr: Approach exposed in the main thread.

elks/arch/i86/drivers/char/serial.c
mfld-fr: Done in 1db6bc5. Moved the fast handlers in the same module as the generic handler to ensure CS is consistent with `int_vector_set` even in medium memory model.

elks/arch/i86/drivers/char/serial.c
mfld-fr: Only one way to set an IRQ handler after 1db6bc5.

elks/arch/i86/kernel/irqtab.S
mfld-fr: Reworked the fast handlers in 1db6bc5.

mfld-fr:
Latest small benefit : now able to see the active interrupts with `meminfo`.

elks/arch/i86/drivers/char/serial.c
```diff
@@ -265,14 +265,16 @@ void fast_com2_irq(void)
 }
 #endif
 
-#if defined(CONFIG_NEED_IRQ4) || defined(CONFIG_NEED_IRQ3)
+
+#if !defined(CONFIG_FAST_IRQ3) || !defined(CONFIG_FAST_IRQ43)
```
mfld-fr: Typo here.

elks/arch/i86/drivers/char/serial.c
```diff
@@ -298,7 +300,9 @@ void rs_irq(int irq, struct pt_regs *regs, void *dev_id)
     if (q->len)		/* don't wakeup unless chars else EINTR result*/
 	wake_up(&q->wait);
 }
-#endif
+
+#endif  // !defined(CONFIG_FAST_IRQ3) && !defined(CONFIG_FAST_IRQ43)
```
mfld-fr: Inconsistent comment

elks/arch/i86/kernel/irq.c
```diff
@@ -21,75 +21,85 @@
...
+	int_proc proc;
+	word_t seg;
+	byte_t irq;
+} __attribute__ ((packed));
+
+typedef struct int_handler int_handler_s;
+
+
+// Add a dynamically allocated handler
+// that redirects to the generic handler
```
mfld-fr: Oudated comment

elks/arch/i86/kernel/irqtab.S
```diff
@@ -137,200 +35,61 @@ irqtab_init:
...
+	cli
+	mov %cx,0(%bx)
+	mov %dx,2(%bx)
+	popf
+
+	pop %ds
+
+	// As ELKS kernel is now in medium memory model,
+	// return the code segment of the current module
+	// that contains the generic interrupt handler
```
mfld-fr: ...and the specific ones also.

elks/arch/i86/kernel/irq-8259.c
```diff
@@ -18,17 +18,13 @@
...
 
-static unsigned char cache_21 = 0xff, cache_A1 = 0xff;
 
 /*
  *	Low level interrupt handling for the X86 PC/XT and PC/AT platform
  */
 
 void init_irq(void)
 {
-#ifdef CONFIG_HW_259_USE_ORIGINAL_MASK	/* for example Debugger :-) */
```
mfld-fr, line 29: Cleanup that configuration option.

ghaerr:
Thank you for all the changes. Most looks very good.

I was hoping to keep this PR from getting so complicated that it require lots of testing, and we are possibly beyond that point. When problems occur, which is frequent with old hardware, I would have to revert entire PR, which I don't want to do.

The remaining areas of concern are:
- removal of enable_irq calls in BIOS disk driver and other routines
- most recent change to _irqit routine using SS to find DS and possible further slowdown. I would very much prefer that big rewrites to _irqit be placed in *separate* PR, so as to not require large PR rollback on unconsidered system crashes. I do not want even more rewrites to _irqit to be in this PR, as you are considering.
- Moving away from cached PIC values
- fast serial port driver moved into irqtab.S
- increasing use of // comments in kernel files that only used /* */ prior
I will address most of these in seperate comment.

Benefits are increasing, which is great. Summary of benefits below, perhaps you can comment on any missed:
- Possible handling of any INT trap, rather than IRQs (what exactly can be done here?)
- Enhanced request_irq allows fast and slow interrupt serial interrupt handling (specific and generic)
- Ability to use fast or slow compiled in serial drivers specified at boot (not done yet, todo in seperate PR)
- Possible upgrade to meminfo to show active interrupts in use
- Ability to dynamically assign interrupt handler to interrupt (e.g. fast serial vs slow serial)
- Removal of requirement to specify interrupts required at compile-time
- Massive source cleanup

Disadvantages:
- low core interrupt table not replaced all at once, but only at driver initialization time (also benefit, but could cause issues with timing so probably need to test drivers)
- Requires code execution in data segment (not a big deal in real mode, and security isn't real issue)
- Slower interrupt response (not a big deal as I thought as most handlers written in C anyways)

ghaerr:
With regards to moving the fast serial driver, that needs to be reverted. I used to have the serial driver in this directory, but it belongs over with the char drivers and was moved a while back.

To fully explain how kernel "medium model" was implemented: it isn't actually a medium model similar to Microsoft, where any function may have a seperate CS code segment. Because of gcc limitations, and also for reasons of exact layout in a.out header, in actuality there are only two code segments, ever, near and far text. All far functions are grouped together by the linker and placed in SINGLE far text segment, with guaranteed only one CS value for all. The limit on near and far text sections is 64k each.

All kernel code will always be placed into the near segment, unless the following attribute is specified on a per-function basis (here is one for INIT routines. I have seperate one for FAT filesystem):
```
#define INITPROC __far __attribute__ ((far_section, noinline, section (".fartext.init")))
```
Thus, in kernel, whether or not a specific routine is near or far is completely controlled by programmer. Currently, I have only two areas of kernel moved to far text: kernel initialization routines and FAT filesystem. 

The next step in the kernel initialization is to discard the INITPROC far text after execution, and place the kernel data segment directly following. That has not been done yet.

So, as far as fast serial driver is concerned, it is guaranteed to be in near text section, you have nothing to worry about. So it needs to move back to char device driver directory.

elks/arch/i86/drivers/block/bioshd.c
```diff
@@ -544,17 +544,9 @@ int INITPROC bioshd_init(void)
...
 #endif
 #ifdef CONFIG_BLK_DEV_BHD
     _hd_count = bioshd_gethdinfo();
-    if (sys_caps & CAP_PC_AT) {	/* PC/AT or greater */
-	enable_irq(HD_IRQ);	/* AT ST506 */
-	enable_irq(15);		/* AHA1542 */
-    }
-    else {
-	enable_irq(5);		/* XT ST506 */
-    }
```
ghaerr, line 557: I don't think it is a wise idea to remove these in this PR. Removal now requires lots of testing which I don't have time to do, and we can't guarantee that certain old BIOS's don't require explicit IRQ enabling. I understand you think that because this PR removes caching these can go too, but that is not good enough reason without lots of testing.

Move these to seperate PR please (or keep for now and mark TODO, since we don't have time for lots of testing on legacy hardware).

elks/arch/i86/drivers/char/kbd-bios.c
```diff
@@ -81,6 +81,5 @@ static void restart_timer(void)
 
 void kbd_init(void)
 {
-    enable_irq(1);		/* enable BIOS Keyboard interrupts */
```
ghaerr, line 84: I suppose it is things like this that caused you to write this PR, the idea being that BIOS had previous functionality and ELKS should leave it alone, right?

Since this is in polling BIOS driver, I suppose leave change, since easy workaround with configuration. This could be used as test on real hardware to determine how accurate our thoughts are on BIOS interrupts. (Do you have any other outside information on BIOS's using interrupts??)

elks/arch/i86/kernel/irq-8259.c
ghaerr: I am still a little worried about not using cached values here. In previous kernel, ELKS was the master and completely controlled system. In new approach, this leaves open for startup BIOS to have setup interrupts in certain way.

For embedded systems, this is probably a good idea, since the startup code is outside ELKS. 

For legacy IBM PC systems with pre v1 BIOS, etc. I don't know whether this is good idea.

elks/arch/i86/kernel/irq.c
```diff
@@ -21,75 +21,85 @@
...
-	printk("Unexpected trap: %u\n", i-16);
-    else
-#endif
-	printk("Unexpected interrupt: %u\n", i);
-}
 
-int request_irq(int irq, void (*handler)(int,struct pt_regs *,void *), void *dev_id)
+// TODO: simplify the whole by replacing IRQ by INT number
+// Would also allow to handle any of the 0..255 interrupts
+// including the 0..7 processor exceptions & traps
```
ghaerr, line 50: What is plan here? Will that be part of this PR?

I haven't checked to see whether upper limit on IRQ/INT number is being checked, should probably be to avoid array out of bounds.

elks/arch/i86/kernel/irq.c
```diff
@@ -132,24 +140,23 @@ void INITPROC irq_init(void)
 {
     flag_t flags;
 
-    init_irq();
+    init_irq();  // PIC initialization
+
+	// TODO: no more need to disable the timer
+	// BEFORE the interrupt initialization
+	// as timer vector is kept as is
```
ghaerr, line 149: Probably true. Perhaps this should be held to separate PR is that rollback risk decreased, based on legacy PC BIOS. Very early BIOS could not properly handle this case, I really don't know. This is same issue as enable_irq in BIOS driver. Both should wait until seperate PR.

elks/arch/i86/kernel/irq.c
```diff
@@ -132,24 +140,23 @@ void INITPROC irq_init(void)
...
 
+	// TODO: move that timer stuff in `timer.c`
+	// to clearly separate the PIC and PIT driving
     disable_timer_tick();
```
ghaerr, line 153: I suggest waiting to next PR on this. Possible no need at all for disable timer function in ELKS, given new way of handling interrupt vectors.

elks/arch/i86/kernel/irq.c
```diff
@@ -132,24 +140,23 @@ void INITPROC irq_init(void)
...
-    /* Re-start the timer only after irq is set */
+    /* Re-start the timer */
     enable_timer_tick();
-
-    if (sys_caps & CAP_IRQ2MAP9) {	/* PC/AT or greater */
-	save_flags(flags);
-	clr_irq();
-	enable_irq(2);		/* Cascade slave PIC */
-	restore_flags(flags);
-    }
```
ghaerr, line 154: For reasons previously stated, I would like to revert this change for this PR. Changing this has possible bad effect base on unknown legacy BIOS behavior.

If we stick all the PC BIOS legacy problem behavior in a seperate PR, it will be easy to rollback for systems that exhibit problems when finally tested, which won't be before this PR committed.

elks/arch/i86/drivers/block/bioshd.c
mfld-fr: I cannot figure out how the BIOS services can even work without the BIOS to setup properly the related interrupts, so I think the risk here is very low, and there is no need to modify that interrupts when using BIOS services in ELKS.

elks/arch/i86/drivers/char/kbd-bios.c
mfld-fr: You suppose right. IRQ1 is already activated, otherwise we could not use INT 16h functions.

elks/arch/i86/kernel/irqtab.S
```diff
@@ -378,10 +139,12 @@ _irqit:
...
 //
-//	Recover data segment
+//	Recover kernel data segment
+//	Was pushed by the CALLF of the dynamic handler
+//	TODO: BP is better for stack work
 //
-//	seg	cs
-	mov	%cs:ds_kernel,%ds
+	mov	%sp,%si
+	mov	%ss:8(%si),%ds
```
ghaerr, line 141: I understand this change, makes routine more dependent on DS trampoline, just as pop %ds does below.
Change is elegant fix, but I hope you understand I have to maintain this code, not you. I really would prefer ds_kernel change to be in seperate PR. Has this been tested on ROM?

elks/arch/i86/kernel/irq.c
mfld-fr: No plan here, just some ideas... I realized through this PR that INT management and IRQ management are not exactly the same things. For example, installing the INT 80h for system calls, or the CPU traps. So I am still thinking about it, and placing reminders in the code, without changing further.

elks/arch/i86/kernel/irq.c
mfld-fr: Again only an idea, no change for now here.

elks/arch/i86/kernel/irq.c
mfld-fr: Again only an idea, not changing this for now.

elks/arch/i86/drivers/char/kbd-bios.c
ghaerr: Such supposition is not proven, there are old BIOS that have bugs. IRQ 1 is not required by hardware for reading keyboard, and ELKS only uses INT 16h polling functions. But agree we can test on this small case easily on real hardware, at least on newer systems.

elks/arch/i86/kernel/irq.c
mfld-fr: Again, I cannot figure out how AT BIOS services (HDD...) could work without enabling the 8259 cascade...

elks/arch/i86/kernel/irqtab.S
mfld-fr: Yes, it works well in ROM, and EMU86 helps a lot to check the new code.

elks/arch/i86/kernel/irq.c
ghaerr: How about we reach a compromise here, and wrap the code for the enable_irq's with \#if 0 for now?

That would allow me to fix things in the event of a problem quickly, rather than muddling through huge changes in this PR.

After testing, the code could be removed. I don't have time for complete regression testing now, but other users could quickly test on their hardware, and easily change if 0 to if 1 etc for remote debugging help.

mfld-fr:
Thanks for your explanations about the kernel memory model. I realize now that the grouping of the handlers is not needed to ensure CS consistency, so I will revert that grouping.

elks/arch/i86/kernel/irq-8259.c
mfld-fr: Well... maybe we could use EMU86 to monitor the PIC access from MS-DOS, and check it does not try to modify the interrupt masking ? If that monitoring shows that MS-DOS, the most widely deployed OS on the IBM PC, does not do that, maybe you could be convinced that ELKS has no reason to do it ?

elks/arch/i86/kernel/irqtab.S
ghaerr: On a side note, sometimes I think it might be good idea to have mechanism in EMU86 where register dump is displayed on same lines (or directly above or below) for "deep inspection" of processor registers when running in 'c' continuous trace mode. This would help greatly instead of having to single-step through possible offending code.

elks/arch/i86/kernel/irq-8259.c
ghaerr: Interesting idea, but not certain it's really relevant to the time I have to spend maintaining ELKS on super old PCs that anybody tries to run ELKS on. 

I would like to know answer, can breakpoints be set on I/O ports in EMU86?

I am willing to live with caching removed in this PR, but leaving in enable_irq's on BIOS driver and others (except keyboard IRQ 1) with if 0 wrappers. Then at least it won't take too much of my time if you are wrong!

mfld-fr: > I would like to know answer, can breakpoints be set on I/O ports in EMU86?

No breakpoint as the existing ones on the command line, but it is easy to add a test in `io-elks.c` when accessing PIC I/O ports and break at that moment. Let me check MSDOS that way...

I will also surround all the removed pieces of code with 0 wrappers.

mfld-fr: Result from MSDOS monitoring : it installs some handlers for some traps (INT1, INT3 and INT4), sends one spurious EOI at boot, but **never** masks / unmasks any interrupts...

ghaerr: Well, that's possibly good news. MSDOS has a way of having its drivers spewn all over the place, in loadable TSRs, himem.sys, etc, so hard to tell for all devices. But I'm OK with leaving enable_irq's out but 0 wrapper, and we'll see. It is interesting to think that perhaps all the enable_irq previous code was all there just to keep up with ELKS own unneeded cached value of the PIC mask, in other words, bad design. And cached value, if not needed, isn't critical path since only read and written on request_irq or disable.

elks/arch/i86/kernel/irq.c
```diff
@@ -21,75 +21,85 @@
...
+// that redirects to the generic handler
+
+static int int_handler_add (int irq, int vect, int_proc proc)
+	{
+	int_handler_s * h = (int_handler_s *) heap_alloc (sizeof (int_handler_s), HEAP_TAG_INTHAND);
+	if (!h) return -ENOMEM;
+
+	h->call = 0x9A;  // CALLF opcode
+	h->proc = proc;
+	h->seg  = int_vector_set (vect, (int_proc) h, kernel_ds);
```
ghaerr: Now that I finally fully understand the new _irqit code, I think a change should be made here to clean up the hack in int_vector_set of returning CS. This hard-to-find result actually delayed my understanding of your new approach.

But the real reason is that now that you understand how kernel medium model works, returning CS has no business in int_vector_set. Change that function to be a void proc and set h->seg = kernel_cs here. In this way, system won't be dependent on ASM routine in the case we want to move _irqit to different segment.

elks/arch/i86/kernel/irq-8259.c
mfld-fr: I tried to use HIMEM.SYS and CD.SYS, but these modules require 80286 / 386 instructions and protected mode. At least it should bring confidence about the timer, keyboard, serial, parallel & floppy IRQs, as these devices are used though the BIOS during MSDOS boot.

ghaerr: I am gaining confidence, but we will see.

At least with 0 wrapper approach, they could be easily changed to \#ifdef CONFIG_BIOS_NEEDS_IRQ_ENABLE or something should this all be a problem. If not, then lots more excess cruft removed!

elks/arch/i86/kernel/irq.c
mfld-fr: Agree, returning CS is useless. Will remove that.

elks/arch/i86/drivers/char/serial.c
mfld-fr: Fixed in 5adc931

elks/arch/i86/drivers/char/serial.c
mfld-fr: Fixed in 5adc931

elks/arch/i86/kernel/irqtab.S
mfld-fr: Removed in 5adc931

elks/arch/i86/kernel/irq-8259.c
mfld-fr: Previous code wrapped in 5adc931

elks/arch/i86/kernel/irq.c
mfld-fr: Removed in 5adc931

elks/arch/i86/drivers/char/serfast.S
```diff
@@ -27,7 +25,11 @@ _irq_com1:
 	push	%dx
 	push	%ds
 
-	mov	%cs:ds_kernel,%ds	// recover kernel data segment
+	// Recover kernel data segment
+	// Was pushed by the CALLF of the dynamic handler
+	mov	%sp,%bx
+	mov	%ss:12(%bx),%ds
+
```
ghaerr, line 32: Thanks for the update here. So, the fast interrupt routines will still use a dynamic handler trampoline? For some reason I was thinking they were going to be routed directly from low-core vector. 

But old code still went through _irqtable stub, and we will likely always require a way to get kernel DS, and this removes ds_kernel kluge.

Did you by chance test this to make sure the 12 offset is correct?

To make sure I fully understand: since SS can be absolutely anything at the time of fast serial interrupt, we are using SS here to access the far call segment pushed by ~~hardware interrupt to get to the~~ trampoline? And that trampoline segment will always be the kernel DS?

~~If so, that means your comment "Was pushed by the CALLF of the dynamic handler" isn't correct: the data segment was pushed by the hardware interrupt prior to execution of the CALLF in the dynamic handler...
Should say "Was pushed by hardware interrupt to dynamic handler in kernel data segment"~~

It seems there are some nice benefits to using the trampoline in kernel data after all. Since old method used trampoline in code segment, it was not easy to get at kernel data segment without kluge. This way, we get kernel data easily, but have no idea where kernel code segment is, and don't care. Correct?

elks/arch/i86/kernel/irqtab.S
```diff
@@ -392,7 +149,7 @@ ktask:				// Already using interrupt stack
 //
 //	Already using interrupt stack, keep using it
 //
-	mov	%sp,%si
+	//mov	%sp,%si
```
ghaerr, line 152: Might as well delete this since its' commented out and loaded above.

If you want to also delete the various other commented out segment prefixes from the BCC conversion years ago:
// seg cs (or es etc), you might as well do that too. (Lines 32 and 35)

elks/arch/i86/drivers/char/serfast.S
mfld-fr: Yes, the fast static handlers are also called by the 'trampoline', because now the kernel DS is pushed by that trampoline and not stored anymore in `ds_kernel`. Before the changes, there were a JMP and a CS prefixed MOV from memory. 

Now there are a CALL FAR (pushing two words), a MOV from register to register (very fast as CPU internal), and a SS prefixed MOV from memory (memory offset has no impact). Therefore the path is a bit slower because of two more pushes, but I think it won't change the game.

In slow or fast interrupt, whatever the initial SS:SP, the HW first pushes 6 bytes (FL, CS, IP) and JMP FAR to the trampoline at kernel DS:xxx (in the heap). Then the trampoline CALL FAR actually pushes 4 bytes (DS, offset of `irq`) and JMP FAR to the static handler. On static handler routine entry, the stack is:
```
pushed by the trampoline:
    irq offset  <-- SS:SP
    kernel DS
pushed by the HW:
    next IP after interrupt
    interrupted CS
    interrupted FL
```

That makes both the kernel DS and the pointer to IRQ number available on the stack for the static handler. Also the comment is correct, kernel DS was pushed by the trampoline.

elks/arch/i86/kernel/irqtab.S
mfld-fr: I kept that MOV instruction commented to hightlight the fact it is performed on routine entry just above and is no more necessary at that place.

elks/arch/i86/drivers/char/serfast.S
ghaerr: Thanks for the comments. I get it now, and comment is correct. 

New method, found by having to add CALLF in trampoline, turns out to have nice benefits for ELKS interrupt handling. I agree now that overall solution is superior, since stack frame contains all information needed for interrupt handler, whereas before it did not and required cs:ds_kernel kluge. Nice work!

elks/arch/i86/kernel/irqtab.S
ghaerr: OK - but far more than that is required for others to understand this routine. At this point, a cleaner diff resulting from commit will be easier to read without it, IMO.

You could mention in comment above that SI has current stack pointer instead.

mfld-fr: Okay, that change without any impact would be performed in a next PR.

ghaerr:
All looks very good, with very minor exceptions on some comment cleanup. Overall, very nice addition to mess of previous IRQ configuration hassles, ability to handle fast and slow serial handlers without recompilation, and improved way of interrupt handling and writing interrupt handlers without requiring previous cs:ds_kernel kluge.

I would like to test this on a real PC I have before commit. That make take a day or so.

Thank you!

mfld-fr:
Thank you for your time to understand the new design ! Please wait before merging, I would like to have another fresh review after a few days, I would like to walk again and again the new code with EMU86 to be sure I missed nothing, and last but not least, I still have to squash the many [TEMP] commits.

mfld-fr:
Maybe @Mellvik, our super tester, could try on his oldies too ?

ghaerr:
> Please wait before merging, I would like to have another fresh review after a few days, I would like to walk again and again the new code with EMU86 to be sure I missed nothing, 


Good idea. Since the internal handling of _irqit and stack switching didn't change, only very entrance, I feel pretty good about those changes never causing system crashes.

We do have some possible concern over ELKS now running with no changes to low-memory interrupt vectors at boot (until irq_request). If for some reason one requires a handler that isn't compiled-in, this could cause a problem, such as with network card. 

It would be nice to test fast serial driver. However, I can do that when adding ability to select fast/slow version from /bootopts, in another commit.

> Maybe @Mellvik, our super tester, could try on his oldies too ?

If @Mellvik has time, this PR could be fetched and tested without change to .config file. Things should just work with no modification. And obviously continue to work while being banged hard with network I/O and serial :)

> and last but not least, I still have to squash the many [TEMP] commits.

Yes. Could I ask you show me the terminal commands you'll use to do that. I don't want to screw up my manual squash attempt the next time!

mfld-fr:
How to squash a branch:
- get a clear picture of the tree: `git log --oneline --decorate --graph`
- check the branch is really a leaf in that tree (no merge from that branch to another one)
- check the number of commits since the base of the branch (not included)
- rebase the branch on itself: `git rebase -i HEAD~nnn` where `nnn` is the number of commits
- Git opens an editor with the list of the commits
- interactively tell Git what to do with the commits (see inline help) : usually 'r' to reword, and 'f' to merge with previous
- save and exit
- you are done!

ghaerr:
Hello @Mellvik and others,

Not sure who's been following the intense set of conversations on this PR, but now that its finished, the upshot for kernel users is that now, the ports.h configuration file no longer has to be edited to specify one of the mess of CONFIG_IRQx interrupt numbers, only the interrupt number itself needs to be correct, like NE2K_IRQ.

Now that I think of it, this change allows a future simple change of the driver interrupt number to be specified at runtime, without recompile. This is a huge benefit. I will add code after commit to parse some new options in /bootopts, that will allow setting at least the network driver and serial drivers to use a runtime-specified IRQ.

For the time being, any testing of serial, network and hard disk drivers would be great. There has been a small change, but significant, in that ELKS now leaves the BIOS to enable or disable its own interrupts for devices like the keyboard and disk I/O. (The BIOS disk I/O doesn't change with this PR, only the non-working, non-BIOS HD/FD drivers changed).

Thank you!

Mellvik:
Thank you @ghaerr,
I've been following the conversations and development with great interest, (honestly) occationally holding my breath for then next move.

I'll be enthusiastically participating in the testing (the ability to set IRQs at runtime is an incredible step forward) as soon as I've figured out why the tools are acting up on me again. 

we have a new macos and history repeats itself. I'm getting exactly the same as #774. Apparently the fixes introduced the last time were not sufficient.

I'll be back when I've figured it out.

@mellvik

ghaerr:
>  I've been following the conversations and development with great interest, (honestly) occationally holding my breath for then next move.

Well, you should head over to @mfld-fr's EMU86 project sometime, we've both been working hard there with lots of improvements. You can now emulate ELKS from boot (PC or ROM version) with a QEMU-style display, and see a continuous instruction trace of 8086 instructions from the very first instruction, or, say, from an application program. We hope to soon have the ability to trace instructions showing the actual ELKS kernel source symbol names, and that's a heck of a nice tool to learn what is going on inside ELKS.

>  as soon as I've figured out why the tools acting up on me again. we have a new macos and history repeats itself. I'm getting exactly the same as #774. 

Can you post the display of issues you're having in a separate issue? Perhaps @tkchia can more quickly determine what needs to be done to do to get you back operational.

Thank you!

ghaerr:
@mfld-fr : 

Thinking about more about the methods of IRQ and PIC initialization changed by this PR: in the old way, upon boot, ELKS immediately replaced all compiled-in hw interrupt vectors to _irqit stubs. Shortly thereafter, but not immediately, various default_handlers were replaced by request_irq, and then the PIC mask finally updated to allow interrupts. Interestingly, the default mask of 0xFF was not sent to the PIC until the first request_irq. Anyways, since ELKS grabbed the hw int vectors early, it pretty much had to use an 0xFF mask or there'd be a period of possible "Unexpected interrupt" display until the driver was initialized. Now, we do it differently, and don't grab any hw vector until request_irq time. 

So far, so good. Prior, ELKS assumed nothing about the BIOS initialitizing the PIC, except possibly for devices the BIOS knows about. But even then, ELKS re-enabled the PIC mask bits, both because it had to because of its design, but possibly as a precautionary measure. With the new method, I don't think this poses much of a problem for the devices the BIOS knows about, or that ELKS has drivers for.

However, the PC BIOS may not know about all devices plugged in. Some may not have an option ROM for initialization. For devices compiled into ELKS, request_irq will handle everything properly, but what about the slave PIC, for instance? Who should set that up, the BIOS, or ELKS?

It seems to me that we should not trust the BIOS to setup the 8259 slave PIC properly, because the BIOS may not know about devices with interrupts 8-15. In addition, and more importantly, ELKS is absolutely responsible to ensure that request_irq work properly *for the PC target*. In addition, ELKS has a (newer) SYS_CAPS mechanism that basically instructs ELKS on whether the system has or has not IRQ2 -> IRQ9 cascade, called CAP_IRQ2MAP9.

Thus, I think it hurts nothing, and helps our PC users to ensure that, when the target system capabilities show that IRQ 2 should be mapped to IRQ 9, that the slave PIC be initialized as well by ELKS at init time. 

So the following code, should be left in, not 0 wrapped, and moved to the PC target-specific irq-8259.c irq_init routine:
```
    if (sys_caps & CAP_IRQ2MAP9) {	/* PC/AT or greater */
 	save_flags(flags);
 	clr_irq();
 	enable_irq(2);		/* Cascade slave PIC */
 	restore_flags(flags);
```

This solves any issue of faulty BIOS's not initializing the slave PIC, and lowers our risk of a user's network card not working when installed in the IRQ 8-15 region, since this code is essentially left unchanged.

[EDIT: I have not found any specification that says BIOS has to initialize the slave PIC. There could be an XT-like machine out there with an XT BIOS but slave PIC installed. For non-IBMPC compatible systems, but those still running on the CONFIG_ARCH_IBMPC "architecture", we cannot actually assume that there is a compatible BIOS present. ELKS source is still dependent on BIOS for PC architecture, but it would be nice to remove that dependency, and have ELKS run on IBM PC with no BIOS. In all cases, the SYS_CAPS would control whether there's a slave PIC, and that would be used to initialize it in the target-specific irq-8295.c file.]

I don't see any other issues at the moment; I think the other enable_irq's we left commented out will be OK, since the BIOS would need to know about those devices, *and* ELKS doesn't have interrupt handlers for those IRQs anyways.

Thank you!

mfld-fr:
According to the IBM BIOS technical reference, the BIOS **has** to enable the PIC cascade starting from the AT model, otherwise some services of INT 1Ah (AH=06h RTC alarm) and INT 15h (AH=83h / 86h) **won't work**, as they use INT 70h = IRQ 8 (RTC interrupt).

For other devices the BIOS would not know, all the silicon makers disable the interrupt on reset pin activated. I have never seen any chip with some interrupt enabled after a reset since I started to read some datasheets in 1984...

Mellvik:
@ghaerr,

> > I've been following the conversations and development with great interest, (honestly) occationally holding my breath for then next move.
> 
> Well, you should head over to @mfld-fr's EMU86 project sometime, we've both been working hard there with lots of improvements. You can now emulate ELKS from boot (PC or ROM version) with a QEMU-style display, and see a continuous instruction trace of 8086 instructions from the very first instruction, or, say, from an application program. We hope to soon have the ability to trace instructions showing the actual ELKS kernel source symbol names, and that's a heck of a nice tool to learn what is going on inside ELKS.

Exciting indeed, I'll do just that, thank you.
> 
> > as soon as I've figured out why the tools acting up on me again. we have a new macos and history repeats itself. I'm getting exactly the same as #774.
> 
> Can you post the display of issues you're having in a separate issue? Perhaps @tkchia can more quickly determine what needs to be done to do to get you back operational.
> 

I'll reopen #774  since the error is exactly the same, messages and all. Weird - since we did test the fix thoroughly last fall.

@Mellvik

ghaerr:
> According to the IBM BIOS technical reference, the BIOS has to enable the PIC cascade starting from the AT model, otherwise some services of INT 1Ah (AH=06h RTC alarm) and INT 15h (AH=83h / 86h) won't work, as they use INT 70h = IRQ 8 (RTC interrupt).

Ok, good to know the requirement to initialize is in the IBM BIOS tech reference. That hasn't stopped various other vendors from non-compliance, including IBM themselves for other things in very early BIOSes, but we'll see. This code should still be moved, 0 wrapped, out of the mainline, into the init routine in irq-8259.c.

mfld-fr:
Agree, I will move that commented code to the 8259 part.

ghaerr:
Tested on 1990's 386 system with Award BIOS 6.00, as well as Compaq Portable 386.

System boots and runs fine. Tested normal (not fast) serial port, works fine. Both these systems have AT BIOS with devices in IRQ 8-15 range, but neither were able to be tested for HD access. 

Unable to test networking since having trouble getting proper IRQ sorted out on 386 motherboard/card (not ELKS problem yet).

Compiled with CONFIG_FAST_IRQ4, compile fails. Following patch required:
```
diff --git a/elks/arch/i86/drivers/char/serial.c b/elks/arch/i86/drivers/char/serial.c
index 22cbdfce..b8c2df41 100644
--- a/elks/arch/i86/drivers/char/serial.c
+++ b/elks/arch/i86/drivers/char/serial.c
@@ -230,6 +230,7 @@ void rs_pump(void)
  * emits code using SP or BP addressing, as SS is not set and not guaranteed to equal DS.
  * Use 'ia16-elfk-objdump -D -r -Mi8086 serial.o' to look at code generated.
  */
+extern void _irq_com1(int irq, struct pt_regs *regs);
 void fast_com1_irq(void)
 {
     struct serial_info *sp = &ports[0];
@@ -248,6 +249,7 @@ void fast_com1_irq(void)
 #endif
 
 #ifdef CONFIG_FAST_IRQ3
+extern void _irq_com2(int irq, struct pt_regs *regs);
 void fast_com2_irq(void)
 {
     struct serial_info *sp = &ports[1];
```

After working compile, system runs and boots. However, ELKS crashes (becomes unresponsive) when running `miniterm` with fast serial driver to login to remote system. System crashes only after sending single serial character, which means very likely either _irq_com1 is not being called properly by new request_irq, OR problem is in "12" offset in `mov %ss:12(%bx),%ds` in serfast.S. Can you calculate a quick-double check on that offset, while I do the same and continue to try to debug fast serial driver? Also please look through request_irq code path for INT_SPECIFIC case.

Thank you!

ghaerr:
Found the problem. Add the following patch, as well as please move commented code discussed earlier to irq-8259.c.
```
diff --git a/elks/arch/i86/drivers/char/serfast.S b/elks/arch/i86/drivers/char/serfast.S
index ab4a23a1..2e1c29c3 100644
--- a/elks/arch/i86/drivers/char/serfast.S
+++ b/elks/arch/i86/drivers/char/serfast.S
@@ -41,6 +41,7 @@ _irq_com1:
        pop     %cx
        pop     %bx
        pop     %ax
+       add $4,%sp
        iret
 #endif
 
@@ -73,5 +74,6 @@ _irq_com2:
        pop     %cx
        pop     %bx
        pop     %ax
+       add $4,%sp
        iret
 #endif
```

New INT_SPECIFIC interrupt handlers still have to dump the two word trampoline DS/IP before returning from interrupt.

Fast serial routine now works great!

BTW, you may not know about a great feature added to `hd` I added a while back: you can inspect memory at runtime using stuff like `hd 0:0`, `hd 0:0#512`, `hd 02c0:e29a`, etc. I used this to determine that the vectors were all correct, until it finally dawned on me the stack was not.

ghaerr:
Also, FYI: while looking through source, realized that you're still using 4 character tabs while most of ELKS kernel is using 8 character tabs, EXCEPT where it is using 4 spaces for first indent AND 8 character tabs. So irq.c now doesn't display properly using either 4 nor 8 character tabs. FYI.

This can be cleaned up after commit, but wanted to bring your attention to it, so it doesn't get worse. This problem throughout the kernel, but I've been resisting reformatting lots of code. I have vi set to use 'k' and 'l' to switch between 4- and 8-character tabs, because I have to in order to read ELKS!

mfld-fr:
I did not miss the additional POP in `_irqit`, but I missed the needed two in the fast handlers :worried:  

Thanks for the fix !

ghaerr:
Since adding one more byte to dynamic trampoline allocation won't change heap usage because paragraph aligned, perhaps trampoline should be changed from "springboard" to true trampoline by adding "IRET" instruction after CALLF, so that specific interrupt routine writers don't have to remember to adjust stack before returning. Also would be useful for forced calls to interrupt routines that can't be rewritten (e.g. are in ROM). This would be a better solution, no?

mfld-fr:
Interesting idea... but I would prefer first to stabilize after that burst of changes, and take time to think quietly about the impact of replacing IRET by RETF in the static handlers...

ghaerr:
> but I would prefer first to stabilize after that burst of changes, and take time to think quietly about the impact of replacing IRET by RETF in the static handlers...

Ok, fine with me since you now have code which has been tested. 

Also, just realized that my proposed solution doesn't work for redirecting to ROM interrupt routines always, since they're stuck with using IRET.

We will stay with using `add $4,%sp` for now, thanks.

mfld-fr:
Ho ho... looks like I also have a problem while compiling Nano... :open_mouth:

ghaerr:
Good gosh... I guess nano's 11,305 line configure script has reasons to be required...

I suppose I'll just comment out nano completely in elkscmd/Makefile, unless you have any better ideas? A full-blown GNU "product" doesn't seem to be easily maintainable without running ./configure, which isn't easily possible in the ELKS application building mechanism without a lot of work. Nano is basically too big to be useful on ELKS anyways.

mfld-fr:
RIP Nano :skull_and_crossbones:

ghaerr:
Are your new squashed commits ready for a final review pass, or are you thinking of waiting a little while longer on this?

mfld-fr:
Well... it looks mature enough now... I think we could go ahead.

Mellvik:
@ghaerr,
Right now I'm unable to compile the latest commits.
If you send me a zipped fd1440 image,  I'll be happy to run it in the morning.

Thank you.

--M

ghaerr:
@mfld-fr : Thank you for this very useful enhancement and cleanup to ELKS! I enjoyed working with you on it.

I plan on adding boot-time configurable network card IRQ shortly, to help with @Mellvik's testing as well as to add a long-requested feature that requires this PR.

@Mellvik : Will attach zip image here. Will also try to have configurable network IRQ as well.

ghaerr:
@Mellvik, 

Here's kernel compiled with latest source with this PR, as well as PR #950 (Runtime NIC IRQ selection) included for your testing.

[fd1440.bin.zip](https://github.com/jbruchon/elks/files/6504800/fd1440.bin.zip)


Thank you!

Mellvik:
Thank you @ghaerr,

the image boots fine, no problem with interrupts - serial, console, clock, whatever, except networking, which works only after commenting out the netirq= statement in bootopts. Minor glitch I suppose.

Will continue to use the image through the day.

—Mellvik

mfld-fr:
Thank you @Mellvik for testing !

ghaerr:
> except networking, which works only after commenting out the netirq= statement in bootopts. Minor glitch I suppose.

Your card is set for IRQ 12, right? What exactly is the `netirq=` line being used in /bootopts, and is it combined with anything else?

Mellvik:
@ghaerr,
you got me there (again). No wonder it didn't work with =9. IRQ 9 is my other NE2K card. Apologies.

Slight consolation :-)  the test sent me down a path to test some oddity I noticed in /bootopts a while back. 
Moving that to its own issue.

Thank you!!

--Mellvik

Mellvik:
Never mind (bootopts) - it was the byte limit again.

Would it be an idea to read the first 256 bytes and ignore the rest? Then we could keep all kinds of comments and examples at the end of the file?

Anyway - I've been beating the system really hard - serial, disk, flp and ether in particular. It's stable.

--M

ghaerr:
> Would it be an idea to read the first 256 bytes and ignore the rest? 

That's what it does now, except instead of ignoring the rest, the message `Ignoring /bootopts: header not ## or size > 255` is displayed early at boot. Is that occurring? The message is easily missed, since its way up top of the boot messages. The current implementation is supposed to always display an error message and never ignore /bootopts contents.

If the message is not being displayed, lets check the /bootopts file length to be sure.

Mellvik:
Thank you @ghaerr,

there are technical and practical issues here, plus the fact that your adding netirq= opens a whole new world of ideas (suggestions)(great enhancement).

So I'm moving this to its own enhancement request issue.

—Mellvik

PR #941, opened 2021-05-15
[setup] Allow IBM PC specific code to be removed from setup.S for new ports

ghaerr:
When CONFIG_ARCH_IBMPC not set, removes all BIOS INT calls in setup.S for probing system configuration.
These values are currently set for non-IBM PC ports in config.h.

Setup data segment still exists for the time being.

@cocus, you should be good to go for your CONFIG_ROMCODE and CONFIG_ARCH_8018X port, as far as setup.S is concerned, as well as no BIOS dependencies in startup code. The entry point for you after system init will be `start_os` in setup.S.

### DIFF
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -230,6 +230,16 @@ start_os:
 	stosw
 #endif
 
+#ifdef CONFIG_ARCH_IBMPC
+//
+// Set keyboard repeat rate using INT 16h
+// Determine display type using INT 10h AH=12 and INT 10h AH=1A
+// Get display page and mode using INT 10h AH=0F
+// Determine CPU type using somewhat outdated methods
+// Clear BIOS data area for fd/hd (24 bytes)
+// Determine size of main memory using INT 12h
+// FIXME move to separate setup-bios.S file or individual drivers
+
 #ifdef CONFIG_HW_KEYBOARD_BIOS
 	mov	$0x0305,%ax	// set the keyboard repeat rate to the max
 	xor	%bx,%bx
@@ -262,7 +272,7 @@ start_os:
 #else
         mov   $25,%al		// height of display in rows
 #endif
-#endif
+#endif /* CONFIG_HW_VGA*/
 
 novga:	mov	%al,14		// CGA 25 rows
 
@@ -290,6 +300,7 @@ novga:	mov	%al,14		// CGA 25 rows
 	mov	%ax,%ds
 	int	$0x12		// determine the size of the basememory
 	mov	%ax,0x2a
+#endif /* CONFIG_ARCH_IBMPC*/
 
 // End of shared setup code
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
@@ -735,8 +746,9 @@ add_ptr:
 
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
+#ifdef CONFIG_ARCH_IBMPC
 /*
-! TODO: move probing to boot tools
+! TODO move to setup-cpuid.S or remove entirely (inaccurate/outdated)
 ! Probe for the CPU
 ! These information is taken from "PC intern 3.0", Data Becker Verlag, 1992
 ! and from the Linux-Kernel, arch/i386/kernel/head.S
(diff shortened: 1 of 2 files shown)


ghaerr:
@cocus, I am committing this now to allow you to continue to proceed with getting the 8018X port operational. Please let me know what else you need.

I will be holding the increasingly-complicated changes and testing required by #940 until after we get a first port of your or @mfld-fr's platform, so as to maintain a stable environment as well as ease the already-complicated source changes required for a port. With what I think are all of the BIOS and setup dependencies removed, at least on the CONFIG_ROMCODE and non-CONFIG_ARCH_IBMPC configuration, we shouldn't be that far away from getting a kernel to boot. 

Thank you!

cocus:
@ghaerr I've rebased my changes on top of your latest merges on the same branch as before (https://github.com/cocus/elks/tree/cocus/80c188-port), and addressed all of your (and @mfld-fr's) comments about the changes. I'm only missing the EOI stuff and the polling of the serial port for the headless console by the looks of it. Maybe I've missed something else. All feedback is welcome!

ghaerr:
Hi @cocus,

Thanks for your continued work on this. It all looks very good, thank you very much!

Actually, even the define_bool CONFIG_CPU_80186 should probably be deleted from config.in, since there is no place within ELKS where that option is even looked at. (Yes, there are a bunch of unused options that need further cleanup!)

If you would like to submit a PR of all these 4 commits in your work branch now, I will commit then to master immediately, so that you won't have to continue to rebase whilst ELKS receives further PRs and activity. We can continue to accept further changes from you afterwards, and hopefully the system will be at the point where a boot attempt can be tried soon.

> I'm only missing the EOI stuff and the polling of the serial port for the headless console by the looks of it. 

Go ahead and add both of those after sending a PR for what you have now. You'll see where to add the EOI stuff in irqtab.S, and the serial polling can be added to headless-8018x.c directly. So you're (so far) the only one working on these files. At some point, @mfld-fr may jump in with some organizational ideas since his target is almost, but not quite, the same.

On another note, @mfld-fr found that the startup code in setup.S still uses an INT 10h AH=0Eh BIOS call to output info and error messages to the console during boot. Since you're going to be running a serial headless console that won't have been initialized yet, this call to ASM procedure `putc` will likely have to be commented out, unless you have another idea or want very early console output.

Thank you!

cocus:
Regarding the int10h, I set up the serial port early on my "loader" for the ELKS that I've just pushed to a new repo (https://github.com/cocus/8018x-elks-ldr), so a pretty straightforward implementation of "wait until tx buffer is empty" can be added for this platform instead of relying on the PC bios services.
On the other hand, I'll go ahead and remove the `define_bool CONFIG_CPU_80186` on config.in and create a PR from my branch.

mfld-fr:
Please keep the CONFIG_CPU_80186, as it would allow us to optimize code for our targets.

ghaerr:
> Please keep the CONFIG_CPU_80186, as it would allow us to optimize code for our targets.

The entire ELKS project uses none of the variety of CONFIG_CPU_xxx in any of the source, with the exception of the buggy `getcpu` routine which is supposed to return the CPU type. That routine and the defines around it are fraught with problems.

All ELKS assembly code is now free from any processor-specific instructions. The last mess we had to clean up was when it was found the PCjr wouldn't execute certain jump instructions, so the assembler directive `.arch i8086, nojumps` was added for enforcement.

We don't need to re-introduce the set of CONFIG_CPU_xxx defines in the project. In this case, @mfld-fr, if you feel the need for extreme optimization, you already have CONFIG_ARCH_8018X which implies this anyways. In addition, we've already agreed that all driver code for architecture-specific chips and targets will reside in seperate source files, where you can apply your optimizations without the define.

Thank you!

mfld-fr:
OK, got it.

ghaerr:
> Regarding the int10h, I set up the serial port early on my "loader" for the ELKS that I've just pushed to a new repo (https://github.com/cocus/8018x-elks-ldr), so a pretty straightforward implementation of "wait until tx buffer is empty" can be added for this platform instead of relying on the PC bios services.

Ok, sounds good. Just wanted to remind you that the INT 10h is still in setup.S. I'm going to post a separate PR that removes it for non-CONFIG_ARCH_IBMPC systems, and then you can replace it with your own communications mechanism for boot messages, if desired. This can be further sorted out after your PR is submitted and we get closer to booting an 8018X system.

PR #945, opened 2021-05-17
[setup] Rearrange setup.S to arch-specific sections

ghaerr:
Removes last BIOS dependency in setup.S for non IBMPC targets. 
With this added, and CONFIG_ARCH_IBMPC not defined, EMU86 should show no BIOS dependencies. (not tested yet).
I'll leave this PR open until I can verify.

Referenced in https://github.com/jbruchon/elks/issues/932#issuecomment-841820046.

@mfld-fr : We probably need to develop an alternative small API for non-BIOS targets for ELKS boot (and possibly more). This PR removes any output during boot for embedded targets, which isn't ideal. In addition, the BIOS INT 12h get memsize function is currently replaced with a hard-coded SETUP_MEMKBYTES in config.h, which should also be enhanced.

An alternative to a small INT API (which forces implementation of some kind of INT xxh on target before porting), is moving the some of the set of setup.S requirements out of hard-coded config.h defines into a set of ASM routines that would then be called from setup.S and filled in with whatever code the port implementer desires. These would be kept in a seperate setup-\<arch>.S file on linked in on a per-target basis.

In other words, should we use a linked-in function call approach, or an INT approach for these items? I'm thinking that since setup.S needs to be split into separate files anyways, the more general approach of letting the implementer do whatever he wants might be better. Thoughts?

### DIFF
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -968,9 +968,12 @@ puts:	lodsb
 putc:	push %bx
 	push %cx
 	push %dx
+#ifdef CONFIG_ARCH_IBMPC
+	// TODO: add non-BIOS output char for other targets
 	mov $0x0E,%ah
 	mov $7,%bx      // page 0
 	int $0x10
+#endif
 	pop %dx
 	pop %cx
 	pop %bx
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -111,7 +111,7 @@ _start:
 
 #ifdef CONFIG_ROMCODE
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-// Entry point for non-EMU86 ROM systems
+// Entry point for IBM PC compatible option ROMs
 
 	.byte 0x55,0xaa		// sign for ROM-Extension
 	.byte 2*CONFIG_ROM_CHECKSUM_SIZE // space for lengthcode (promsize/512)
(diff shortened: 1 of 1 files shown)


mfld-fr:
No thought... just waiting for @cocus to open his PR and try to play his port in EMU86 with TARGET=advtech...

ghaerr:
@cocus : IBM PC and 8081X specific code is now consolidated to its particular source section at end of file labeled CONFIG_ARCH_IBMPC and CONFIG_ARCH_8018X.
Nothing needs to be done for your first port. 

Boot messages will be discarded for 8018X until a call to your own firmware to output AL is written in the `putc` routine, but that's OK.

Later, system memory size should be calculated in this section and config.h SETUP_MEM_KBYTES changed in config.h.
Architecture-specific sections likely to move to their own source file. We can do all this once you get a kernel booting.

ghaerr:
Committing now. System-specific code for particular for the 8018X ports can be temporarily placed in CONFIG_ARCH_8081X section.

Next step is to introduce PLATFORM / FIRMWARE configuration selections that will allow target-specific code to compiled in as separate files for a particular target. Design is not complete, but thinking something like:

CONFIG_PLATFORM_IBMPC=y (selects required platform drivers irq-8259.c, timer-8254.c, etc)
CONFIG_PLATFORM_8018X=y (selects required platform drivers irq-8081x.c, timer-8018x.c, etc)
CONFIG_FIRMWARE=bios (selects filename extension setup-bios.S used in setup.S, console/screen, etc)
CONFIG_FIRMWARE=advtech (selects setup-advtech.S initial kernel setup, kbd-advtech.c, console-advtech.c, etc)

A CONFIG_TARGET option could then be added provides some combinations of the PLATFORM/FIRMWARE, e.g. IBM PC w/PC BIOS, IBM PC on bare metal (no BIOS), 8081X on r8810, 8081X on advtech, etc.

PR #946, opened 2021-05-17
8018x port of PIT, PIC + headless console

cocus:
This PR adds support for the 8018x embedded platforms, supporting on this stage the integrated peripherals that equate to the PIC (interrupt controller), PIT (timer/counters) and a headless console (via the serial port).

### DIFF
--- a/elks/arch/i86/drivers/char/Makefile
+++ b/elks/arch/i86/drivers/char/Makefile
@@ -39,6 +39,11 @@ ifeq ($(CONFIG_ARCH_SIBO), y)
 OBJS  = init.o sibo/sibo_con.o sibo/sibo_conasm.o sibo/sibo_font.o \
 	sibo/sibo_key.o sibo/sibo_keyasm.o mem.o \
 	meta.o ntty.o sibo/sibo_serasm.o pty.o
+else ifeq ($(CONFIG_ARCH_8018X), y)
+	OBJS  = init.o mem.o \
+		ntty.o meta.o tcpdev.o pty.o bell.o
+
+	OBJS += headless-8018x.o
 else
 OBJS  = serial.o serfast.o lp.o init.o mem.o \
 	ntty.o meta.o tcpdev.o pty.o bell.o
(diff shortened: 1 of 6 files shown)


ghaerr:
Looks good @cocus, thank you!!

@mfld-fr : do you want to run this through EMU86 before my commit, or shall I commit now and await your feedback for further changes?

mfld-fr:
I am afraid I cannot test until a few days now, so...

ghaerr:
@cocus, I'm committing this now, please prepare additional PRs as you develop remaining parts. @mfld-fr will likely join in with changes too for 8018X port, while I concentrate on continuing to cleanup and reorganize other parts of ELKS for port requirements.

Thank you!

cocus:
Pretty good!
So next item on the list would be the EOI handler, then the int 10h on the setup.S.
Afterwards I should be able to test if I can boot it somehow on emu86, but I should add a lot of PIT and ICU changes to do so.

ghaerr:
> So next item on the list would be the EOI handler, then the int 10h on the setup.S.

The spot for your code, wrapped by CONFIG_ARCH_8018X should be directly after existing CONFIG_ARCH_IBMPC EOI code.

I handled the INT 10h issue in setup.S in #945, which I will commit soon after considering a few other possible cleanups in setup.S, so no need to do anything in setup.S before your first boot. That will mean there would be no "ELKS Setup..." boot message for now, but that's ok.


> Afterwards I should be able to test if I can boot it somehow on emu86, but I should add a lot of PIT and ICU changes to do so.

Yes, or just change following options from `emu86-rom.config` to your target values, then run PC version in EMU86 with changed ROM values, then set CONFIG_ARCH_8018X instead and flash your target and try boot!
```
CONFIG_ROM_SETUP_CODE=e000 <-- change to your target
CONFIG_ROM_KERNEL_CODE=e060 <-- change to above + 0x60
CONFIG_ROM_SETUP_DATA=0060
CONFIG_ROM_KERNEL_DATA=0080
CONFIG_ROM_BASE=e000 <-- change to your target
CONFIG_ROM_CHECKSUM_SIZE=64 <-- probably not needed, leave same (for option ROM, ignored by you)
CONFIG_ROMFS_FS=y
CONFIG_ROMFS_BASE=8000 <-- need to figure where you'll stick the ROM filesystem (romfs.bin flashed separately)
```
The kernel should boot and display normal start messages out serial port even without ROMFS flashed, so ROMFS  doesn't have to be flashed for first boot attempt.

PR #947, opened 2021-05-17
Char driver: /dev/port

marcin-laszewski:


ghaerr:
Hello @marcin-laszewski,

Committing now. May I ask what program you are considering that require use of /dev/ports, since IN/OUT instructions are not privileged in real mode and thus /dev/ports not technically required? Or is this to learn about writing kernel drivers?

Thank you!

marcin-laszewski:
Hello Gegory,

I wanted to enable hardware ports for userland programs.
The idea was taken from: https://tldp.org/HOWTO/IO-Port-Programming-2.html (2.2)
I have used LPT port for years as "General purpose IO" - I was useful
and easy to maintain.
There are good exmpales of usage:  https://tldp.org/HOWTO/Coffee-3.html

I've checked this piece of software with LED connected to LPT.
Look at the attached photo - it was working under ELKS in QEMU (but
the cat had changed the console).

I will send my  program as PR to browse.

Thank you for commiting.

---
ml.

PR #948, opened 2021-05-18
/dev/port example program.

marcin-laszewski:
Demo just for browsing, not for commiting.


ghaerr:
Thank you, I see how you are using it.

toncho11:
But to use the above example with "/dev/port" the kernel must be compiled with "port" option enabled? So this option should be enabled by default? 

I have a page here using the memory address directly: https://github.com/jbruchon/elks/wiki/Using-Parallel-Port-in-ELKS

marcin-laszewski:
2021-05-21 11:48 GMT+01:00, toncho11 ***@***.***>:
> But to use the above example with "/dev/port" the kernel must be compiled
> with "port" option enabled? So this option should be enabled by default?

Yes, of course.

> I have a page here using the memory address directly:
> https://github.com/jbruchon/elks/wiki/Using-Parallel-Port-in-ELKS
It is possible to use hardware port directly but, according to "Unix
Spirit": users should 'device as files'.

Unfortunatelly it is not so fast as direct access but it is "right
way" and we, for exampel, can control access to the device setting
'chmod/chown/chgrp' of this file.

And it is compatible with Linux.

ghaerr:
Hello @marcin-laszewski,

> Demo just for browsing, not for commiting.

Since this user program doesn't do anything but test your kernel enhancement and comments seem to have ended, we should close this issue. If you'd like to keep the test code in the tree, please change its location to the other test application test modules, under elkscmd/test/port.

Thank you!

marcin-laszewski:
Your are right, it is demo/test only, and it should be removed and closed.
I am writing a more useful program which uses /dev/port now.

PR #950, opened 2021-05-18
[boot] Runtime selection of network card IRQ

ghaerr:
Adds the ability to specify the IRQ for NE2K and WD network interface cards at boot, in /bootopts, rather than having to recompile the kernel.

Use `netirq=9` etc to specify in /bootopts when making distribution, or can also be done from ELKS by editing /bootopts, saving, sync, reboot.

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k.c
+++ b/elks/arch/i86/drivers/net/ne2k.c
@@ -20,6 +20,8 @@
 
 #include "ne2k.h"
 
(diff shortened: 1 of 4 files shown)


toncho11:
Hi,

Is it possible to do the same for the: IO base address(PORT) ?

Mellvik:
Hi @toncho11,
I'm working on that (for network cards, but easily extendable). It's in 'test-mode' and ended up being downgraded on my priority list a couple of weeks back. Thanks for kicking it back to the top.
Stay tuned.

--Mellvik

PR #952, opened 2021-05-21
[kernel] Reorganize console source code for 8018X port

ghaerr:
Reorganizes source code for Direct, BIOS and Headless consoles as discussed in https://github.com/jbruchon/elks/issues/932#issuecomment-843745315.

Adds "conio" API for easy porting to new platforms using headless console.

Moves PC-specific bell speaker code into bell-8254.c.

Adds conio-8018x.c for 8081X port. @cocus : this just needs the conio_poll() routine implemented and headless console should be complete and working for your port.

Updates 8018x.config to use just 1 external system buffer, required for 128K 8018X (ROMFS doesn't need external buffers).


cocus:
@ghaerr I've built the latest emu86 and run it with this code. It still outputs text properly. I've added the following on my conio_poll() routine:
```
    /* S0STS bit 0x40 RI (receive interrupt) */
    if (inw(0xff00 + 0x66) & 0x40) {
        return inb(0xff00 + 0x68); /* R0BUF */
    }
    return 0;
```
Now, I've also modified emu86 to set the "RI" bit (0x40) whenever there's something ready on the receive buffer. The interesting part of this is that I don't see the ELKS polling the S0STS register after it stopped writing stuff to the console.
I don't know if the system halted, or if something else is going on.
EDIT: Maybe the interrupts are not working at all, and the EOI is needed? I might have started tackling the wrong task first!

ghaerr:
Hello @cocus,

> I've added the following on my conio_poll() routine

That was quick!

> The interesting part of this is that I don't see the ELKS polling the S0STS register after it stopped writing stuff to the console.

The con_poll() routine is called by the kernel timer callback mechanism every 8/100 seconds  or so. If the timer isn't firing, then the polling won't happen.


> Maybe the interrupts are not working at all, and the EOI is needed?

That's almost surely it. Since there was no EOI received by the interrupt controller after the very first timer interrupt, the controller won't let another occur... and 8 timer interrupts are needed to accumulate the 8/100 seconds to kick off the call to con_poll.

cocus:
> That's almost surely it. Since there was no EOI received by the interrupt controller after the very first timer interrupt, the controller won't let another occur... and several timer interrupts are needed to accumulate the 8 msecs to kick off the call to con_poll.

Ahg! okay. Well, I've spotted an issue on emu86 as well, regarding timer2. it's set to use a register with offset 10, but in reality, it's the 11th. After fixing that, I get consistent TIMER0 ints (in our case timer2 is used as the count source of timer0).
I'll get to understand how to do these EOIs properly.

ghaerr:
> After fixing that, I get consistent TIMER0 ints (in our case timer2 is used as the count source of timer0).

EMU86 should be firing IRQ 0 timer interrupts regularly. There isn't a way to set an option to show when INT 8h is being called, but you should be able to see it in a continuous trace.

> I'll get to understand how to do these EOIs properly.

Ideally, you'll only need to send a single non-specific EOI from irqtab.S to signal to the controller that the current interrupt is acknowledged, and ELKS will handle everything automatically for nested interrupts. If we don't use the non-specific EOI, then things could get more complicated to handle within ELKS.

cocus:
> Ideally, you'll only need to send a single non-specific EOI from irqtab.S to signal to the controller that the current interrupt is acknowledged, and ELKS will handle everything automatically for nested interrupts. If we don't use the non-specific EOI, then things could get more complicated to handle within ELKS.

From the datasheet, looks like I can send a non-specific EOI:
```
An interrupt handler typically ends with an EOI command. There are two types of EOI commands: nonspecific and specific. A nonspecific EOI simply clears the In-Service bit of the highest priority interrupt.
To issue a nonspecific EOI command, set the NSPEC bit. (Write 8000H to the EOI register.)
A specific EOI clears a particular In-Service bit. To issue a specific EOI command, clear the NSPEC bit and write the VT4:0 bits with the interrupt type of the interrupt whose In-Service bit you wish to clear. For example, to clear the In-Service bit for INT2, write 000EH to the EOI register. The timer interrupts share an In-Service bit. To clear the In-Service bit for any timer interrupt with a specific EOI, write 0008H (interrupt type 8) to the EOI register.
```

So, on irqtab.S, should I add something like this for the 8018X case?
```
//
//	Determine if trap or interrupt
//
	cmp	$16,%ax
	jge	was_trap	// Traps need no reset
	or	%ax,%ax		// Is int #0?
	jz	dont_do_eoi // note that this logic is inverted to the IBM counterpart

        mov $0x8000, %ax // set the NSPEC bit on the
        mov $0xff02, %dx // EOI register
        out %ax, %dx
dont_do_eoi:
```

ghaerr:
> So, on irqtab.S, should I add something like this for the 8018X case?

Yes, but we don't need special handling on IRQ 0. It also needs an EOI, unless it is specially handled by the 8018X.

The reason IRQ 0 was handled specially in the IBM PC is because that code calls the original BIOS vector for IRQ 0, and the BIOS sends its own EOI. Your system doesn't have a BIOS so it needs an EOI for IRQ 0 also.

cocus:
Oh I see.
I've updated that, and seems like I got the timer running! (I added a conio_out on the timer procedure just to validate this)
Now seems like I forgot to handle reads of bytes on port 0xff68 (R0BUF) on emu86, d'oh! Maybe I'll hack this to read an entire word for now.
EDIT:
```
info: load address E0000h
info: load file /home/cocus/mystuff/elks/elks/arch/i86/boot/Image
info: file size=10000h
success: file loaded
info: load address 80000h
info: load file /home/cocus/mystuff/elks/image/romfs.bin
info: file size=BF54h
success: file loaded
Headless console
8018X machine, 256K base RAM.
ELKS kernel 0.4.0 (25616 text, 0 ftext, 3232 data, 35424 bss, 26880 heap)
Kernel text at e062:0000, data 80:0000, top 4000, 190K free
VFS: Mounted root 0x0600 (romfs filesystem).
No init - running /bin/sh
# ls  
.
..
bin
bootopts
dev
etc
home
lib
mnt
root
tmp
var
# nice, it works!
```

ghaerr:
> I've updated that, and seems like I got the timer running! 

Very nice, Santiago!!!

Remember to also set SETUP_MEM_KBYTES to 128 in config.h for your system, until we write a routine or API to get the memory size at boot.

ghaerr:
Since you've actually got ELKS up and running on 8081X on EMU86, I'll commit this now so that you can post your changes as a PR after my changes. That can then be used by @mfld-fr to test the EMU86 PRs you've opened.

Nice work!!

cocus:
Indeed, I'll stash whatever I have and create a PR after this gets merged. This is a pretty good beginning for this port!

ghaerr:
> This is a pretty good beginning for this port!

It sure is! Of course we have to see it run on real hardware too!

I need to make a few more changes after your PR so that additional utilities are added to the 8018X / ROMFS filesystem so that we can look further into the system operation, like `ps` and `meminfo`. That will allow us to see more of memory available, etc.

When you make the SETUP_MEM_KBYTES change to 128, I think you'll find there's only 11K free, so I'm not sure yet you'll be able to execute much of anything lol. But we can also reduce the kernel heap with a config.h option, and that should allow more programs to run until we do some finer tuning.

Thank you!

cocus:
As long as the kernel boots, I can solder an additional 128k or even 512k chips for the RAM. Since the 80C188EB that I'm using has (at least) one additional general purpose chip select that I'm not using, I could map the additional chip very easily. My "bitbanged SD" interface uses some of these and I'll have to double check if I can still use it afterwards. Worst case scenario, I end up removing the original 128k chip and replacing it with 512k.

PR #953, opened 2021-05-21
8018x: conio fixes and eoi impl

cocus:
conio-8018x.c: add poll logic

Adds the logic on the poll function to read received characters
on the integrated serial port present on the 8018x.

irqtab.S: add EOI logic for 8018x

Sets the NSPEC bit on the ICU's EOI register which clears the
In-Service bit of the highest priority interrupt.

### DIFF
--- a/elks/arch/i86/drivers/char/conio-8018x.c
+++ b/elks/arch/i86/drivers/char/conio-8018x.c
@@ -24,7 +24,10 @@ void conio_init(void)
  */
 int conio_poll(void)
 {
-	// add code to poll UART here and return character if available
+    /* S0STS bit 0x40 RI (receive interrupt) */
+    if (inw(0xff00 + 0x66) & 0x40) {
+        return inw(0xff00 + 0x68); /* R0BUF */
+    }
     return 0;
 }
 
--- a/elks/arch/i86/kernel/irqtab.S
+++ b/elks/arch/i86/kernel/irqtab.S
@@ -327,6 +327,17 @@ do_eoi:
 	jmp	a5
 a5:	jmp	a6
 a6:	out	%al,$0x20		// Ack on primary controller
+
+#elif defined(CONFIG_ARCH_8018X)
+//
+//	Determine if trap or interrupt
+//
+	cmp	$16,%ax
+	jge	was_trap	// Traps need no reset
+
+	mov $0x8000, %ax // set the NSPEC bit on the
+	mov $0xff02, %dx // EOI register so the ICU
+	out %ax, %dx // acks the highest priority interrupt
 #endif
 
 //


ghaerr:
Thank you! Would you like this committed now, or are you going to push more stuff?

cocus:
Nope, nothing else to commit. These were the only changes needed :)

ghaerr:
Ok. If you run this on real hardware, you'll want the SETUP_MEM_KBYTES change for 128K. You and @mfld-fr can sort out how to calculate the total system memory in setup.S, or come up with an ABI to your INIT ROM to get mem size, and we'll make a change to SETUP_MEM_KBYTES at that time.

cocus:
Yes, we can come up with some kind of ABI to pass this information, or reuse the settings of the /LCS (assuming this is the only RAM available in the system)

ghaerr:
Hello @cocus,

Now that you've got the EOI, PIC and PIT working, it would be nice, when you have time, to clean up things a bit and replace hard I/O port values like 0xFF00 + 0x08 etc with something like \#define PCB_IRQ_MASK 0xFF08 (or named better since you know the chip), and place all these new constants in a new architecture-specific file elks/include/arch/8018x.h. If you like, the entire set of PCB register offsets/defines for all internal devices could be laid out in that file.

Should there be differences in offsets between chip variations, that can be handled in that header file, rather than by using \#ifdefs in the .c files. We will think up a configure option for the case variations, when we need them.


This should be done for all 8018X changed files, including irqtab.S, irq-8018x.c, timer-8018x.c and conio-8018x.c.

Also, did you see @mfld-fr's recent rewrite of ELKS' interrupt vector handling in #940? In particular, the concept of caching the mask for the interrupt controller was removed, and instead, the controller mask is read, masked, then written by the enable_irq routine. This has the new distinct advantage of ELKS not having to make assumptions  about BIOS (or in your case, INIT ROM) interrupt handling that may or may not be setup prior to booting ELKS.

So, at some point, consider rewriting the following code to not use a static mask cache in irq-8018x.c:
```
/* TMR and SER interrupts are enabled by default */
unsigned int cache_08 = 0xfa;
```




Thank you!

cocus:
> Now that you've got the EOI, PIC and PIT working, it would be nice, when you have time, to clean up things a bit and replace hard I/O port values like 0xFF00 + 0x08 etc with something like #define PCB_IRQ_MASK 0xFF08 (or named better since you know the chip), and place all these new constants in a new architecture-specific file elks/include/arch/8018x.h. If you like, the entire set of PCB register offsets/defines for all internal devices could be laid out in that file.
> 
> Should there be differences in offsets between chip variations, that can be handled in that header file, rather than by using #ifdefs in the .c files. We will think up a configure option for the case variations, when we need them.

I would like to add this. I want to check the registers the CPU @mfld-fr has to make this as friendly as possible; where we can go and make this portable across these two CPUs on a single header file (maybe adding a config to break up the offsets on some registers), or I can use a naming scheme that would fit both (considering a new .h file is to be added and a config to select which one to use).
At the same time, it would be nice to architect this header file to be added on assembly sources, like irqtab.S.




> Also, did you see @mfld-fr's recent rewrite of ELKS' interrupt vector handling in #940? In particular, the concept of caching the mask for the interrupt controller was removed, and instead, the controller mask is read, masked, then written by the enable_irq routine. This has the new distinct advantage of ELKS not having to make assumptions about BIOS (or in your case, INIT ROM) interrupt handling that may or may not be setup prior to booting ELKS.
> 
> So, at some point, consider rewriting the following code to not use a static mask cache in irq-8018x.c:
> 
> ```
> /* TMR and SER interrupts are enabled by default */
> unsigned int cache_08 = 0xfa;
> ```

I haven't checked these changes, but I'll bring them to the irq-8018x.c so we don't cache this value anymore.
I would like to clarify, but in this scenario, are we using other IRQs besides the timer?

ghaerr:
> where we can go and make this portable across these two CPUs on a single header file (maybe adding a config to break up the offsets on some registers),

Yes

>  or I can use a naming scheme that would fit both (considering a new .h file is to be added and a config to select which one to use).

> I want to check the registers the CPU @mfld-fr has to make this as friendly as possible;

[EDIT: Agreed, it would be best to understand the differences before the design, although it was mentioned that the purposes are the same, just the offsets differ.]

No - that would involve ifdefing each .c source file. Better to keep the variations of the address in the .h file, using the same name, if possible.


> At the same time, it would be nice to architect this header file to be added on assembly sources, like irqtab.S.

As long as you use /* */ comments instead of // in the .h file, it can be included in .S files with no problems, provided it consists of just defines, which is what we want. The ELKS make mechanism automatically adds a C preprocessor pass prior to the assembly if the filename ends in .S vs .s.

> I would like to clarify, but in this scenario, are we using other IRQs besides the timer?

We may or may not be using other IRQs, the new mechanism allows the INIT BIOS to setup serial IRQ if wanted, or not,, for instance. Or ELKS can use serial IRQs by calling request_irq, which ends up calling enable_irq to unmask the interrupt controller mask at driver init time.

The new interrupt handling mechanism would OR-in the mask bit when needed at enable_irq time. So it will work with other IRQs, or not, provided that the IRQ mask can be read.

Right at the moment, the 8018X ELKS port seems to only be using the timer IRQ, but that could change on the addition of an additional driver.

cocus:
> > where we can go and make this portable across these two CPUs on a single header file (maybe adding a config to break up the offsets on some registers),
> 
> Yes

So, for instance, I understand that the map from the R8810 and the 80188eb is identical on the first section, but there's a gap after some registers. It should be easy to add something like this:
`#define PCB_THIS_IS_A_REGISTER_WITH_OFFSET_ON_R8810 ((CONFIG_ARCH_R8810 ? <constant offset for r8810> : 0) + offset_for_this_register)`
But it might be better to just take a constant offset for these affected registers, where it has the appropriate offset for R8810 and 0 for 80188eb. Now, to achieve this, an #ifdef is required, or a config with the appropriate offset set in the config.in.
What are your thoughts?

> [EDIT: Agreed, it would be best to understand the differences before the design, although it was mentioned that the purposes are the same, just the offsets differ.]

This is correct, but there are some differences, not many. For instance, in the 80188eb I can mask or unmask an interrupt source by just going into the MASK register, where on the R8810 you have to unmask it on the appropriate interrupt control for that particular interrupt. On the 80188eb you can also do the same, and these bits on the MASK register and each int. control register are mirrored; but for compatibility reasons, it is better to write into the int. control registers instead of the MASK register just to achieve compatibility.




> > At the same time, it would be nice to architect this header file to be added on assembly sources, like irqtab.S.
> 
> As long as you use /* */ comments instead of // in the .h file, it can be included in .S files with no problems, provided it consists of just defines, which is what we want. The ELKS make mechanism automatically adds a C preprocessor pass prior to the assembly if the filename ends in .S vs .s.

Got it.



> > I would like to clarify, but in this scenario, are we using other IRQs besides the timer?
> 
> We may or may not be using other IRQs, the new mechanism allows the INIT BIOS to setup serial IRQ if wanted, or not,, for instance. Or ELKS can use serial IRQs by calling request_irq, which ends up calling enable_irq to unmask the interrupt controller mask at driver init time.
> 
> The new interrupt handling mechanism would OR-in the mask bit when needed at enable_irq time. So it will work with other IRQs, or not, provided that the IRQ mask can be read.
> 
> Right at the moment, the 8018X ELKS port seems to only be using the timer IRQ, but that could change on the addition of an additional driver.

Okay, so this is important to know. Thanks

ghaerr:
> So, for instance, I understand that the map from the R8810 and the 80188eb is identical on the first section, but there's a gap after some registers. It should be easy to add something like this:
#define PCB_THIS_IS_A_REGISTER_WITH_OFFSET_ON_R8810 ((CONFIG_ARCH_R8810 ? <constant offset for r8810> : 0) + offset_for_this_register)

Too complicated and hard to understand, and also won't work for inclusion in .S files.

> But it might be better to just take a constant offset for these affected registers, where it has the appropriate offset for R8810 and 0 for 80188eb. 

Lets keep it simple and readable. Suggest just using 0xFF02 say for one variant, and 0xFF28 for another.

> Now, to achieve this, an #ifdef is required, 

Yes, but only in the .h file. Suggest starting file with all names that are (currently known to be) identical at the top.
Then, a seperate section, ifdefed with a new config option (use if 0 or 1 for now) that contain all differences between variations. If this proves to be too complicated, we can sort that out later. The seperate differences section doesn't need to be added yet, until @mfld-fr does his port.

> or a config with the appropriate offset set in the config.in.

A potential issue with doing arithmetic in the .h file can come up, but is usually optimized out. I suggest keeping things simple at this point, until we see how many different options may be required.

In other words, lets start simple, just get some defines going for your port in a new header file. As @mfld-fr and you progress, this can be made more complex as required.


> For instance, in the 80188eb I can mask or unmask an interrupt source by just going into the MASK register, where on the R8810 you have to unmask it on the appropriate interrupt control for that particular interrupt. On the 80188eb you can also do the same, and these bits on the MASK register and each int. control register are mirrored; but for compatibility reasons, it is better to write into the int. control registers instead of the MASK register just to achieve compatibility.

Sounds like things could get more complicated. Suggest just cleaning up your port first using read/or/write in irq-8018x.c and defines for registers and ports. Then you and @mfld-fr can work out differences as they come up.

PR #954, opened 2021-05-21
[docs] Update porting guide and memory layout docs

ghaerr:


ghaerr:
@cocus : Can you give me your system/board name/manufacturer and 8018X CPU variant? I would like to add that to this documentation. 

@mfld-fr : Your Advantech SNMP-1000B is built with the RDC 8810 CPU, correct? Is there an 8018X-style variant suffix on that? And 512K RAM?

Does each manufacturer add slightly different serial, timer, interrupt controllers, other PIO etc to their designs, or is it just the PCB offsets that differ? I'm trying to increase my understanding of the implementations that are out there.

mfld-fr:
R8810, as 8018XEB, are rather MCU than CPU. After comparing bits, some integrated devices are quite the same, for example the timers, some others quite different, for example the serial port... this is why I am decoupling these devices in EMU86, avoiding the single `8018x.h` file approach, to be able to combine them into many 8018X variants.

ghaerr:
> this is why I am decoupling these devices in EMU86, avoiding the single 8018x.h file approach, to be able to combine them into many 8018X variants.

I have asked @cocus to prepare a single 8018x.h file for now, to move from hard-coded constants in the associated interrupt controller, timer and conio serial drivers. When you do your port, more reconsideration and changes are welcome for best accommodating 8018X variants, in source and configuration files. Hopefully this can be done with a minimum of source files and source near-duplication, but I suppose that remains to be seen.

Documentation/text/memlayout.txt
```diff
@@ -34,3 +27,31 @@ ELKS typical memory layout of 1M physical memory
...
+ | ...
+ | application memory   2080:0000 20800       (starts at ~130K with 64K sys buffers)
+ | ...
+ | ...
+ | top of 256k ram      4000:0000 40000 40000 (256k)
+ | ...
+ | top of 512k ram      8000:0000 80000 80000 (512k)
+ | start of ROM disk    8000:0000 80000 30000 (~180K, CONFIG_ROMFS_BASE=0x8000)
+ | end of ROM disk      B000:0000 B0000
+ | ...
```
mfld-fr, line 55: You forgot the kernel @ E000:0h

ghaerr: Oops! Thanks.

cocus:
> @cocus : Can you give me your system/board name/manufacturer and 8018X CPU variant? I would like to add that to this documentation.

This board was used for some HDSL to E1 converters, the manufacturer name is "Orckit" and this particular board has a model "OR566", but I also have two additional modules from the same manufacturer that I assume have the same memory layout and can be easily hacked as well.
I've removed the E1 framer chips since I don't need them for anything. And I've disabled the hardware watchdog since I don't want to keep constantly kicking it when I was just tinkering with this board. There was an I2C memory connected on this board, which was bit-banged from the processor, but I took that one out so I could connect an SD card.

The overall specs of this board are:
* Intel 80188EB-20 (clocked at 16MHz with a 32MHz crystal)
* 128k SRAM connected on the LCS pin
* 512k FLASH connected to the UCS pin
* serial port exposed on a female db9 port
* 8 status LED outputs mapped on address 0 of the GCS2

I have some pictures of this board, but they don't like too nice since they have all the "sd mod" wires attached.
IMG_20210411_165819
In the meantime, you can see the board on this ebay publication (https://www.ebay.ie/itm/ORCKIT-OR566-E2-Module-/122858618441). Note that I've taken that "ALTERA" cpld board off, since I originally got this board to take that CPLD for a different project :)

PR #956, opened 2021-05-22
[kernel] Add option to specify size of kernel local heap

ghaerr:
Also adds `make kclean` that allows kernel-only (no tools) clean, very useful for switching between config files and recompiling the kernel.

With some smaller memory (128K RAM) systems being ported to, this option will allow a very small (~1024 byte) kernel near heap, up to the maximum size which by default is the amount remaining when the kernel data+bss is extended to 64K. 

Given that the current 8018X port on a 128K Orckit OR566 defaults to a 26K near heap, with only 11K free for applications once /bin/sash is loaded, this option will allow many programs to run which otherwise could not.

Once committed, set `SETUP_HEAPSIZE` in config.h to configure a reduced kernel heap size. Setting too large a value will always be truncated to max heap that will fit within the remaining kernel data segment. This seems to vary between 10K and 26K based on .config file options.

NOTE: Current PR cannot yet be committed, as it seems a bug has been found in the kernel local heap manager. SETUP_HEAPSIZE is set to 704 (bytes) for testing, in elks/arch/i86/kernel/system.c. Kernel boots and meminfo shows 100 bytes free in near heap. Changing the value any lower results in `meminfo` continuously looping. This PR has temporary debug code added to meminfo to stop after first loop, and free/used memory is incorrect.

@mfld-fr : Something very strange is going on when kernel near heap gets near full. What appears to be happening when heap is specified to be small (but large enough for kernel operation) is that the last list item in the heap linked list loops back to the starting point, rather than pointing to `&_heap_all`. Thus, `meminfo` loops continuously if heap size is less than 704, say 700. 

In order to test that it wasn't meminfo's problem, HEAP_DEBUG can be turned on in heap.h, and then your unmodified kernel heap testing code will loop continuously, which shows heap list corruption. This code will run triggered by meminfo running:
```
// Dump heap
#ifdef HEAP_DEBUG
void heap_iterate (void (* cb) (heap_s *))
{
    list_s * n = _heap_all.next;

    while (n != &_heap_all) {
        heap_s * h = structof (n, heap_s, all);
        (*cb) (h);
        n = h->all.next;
    }
}
#endif /* HEAP_DEBUG */
```

I haven't gotten too far into the heap manager yet, but there seems to be data corruption in the heap manager when manipulating the end list item (which is only seen when a small heap is configured). I suspect we have never seen this error before since ELKS has always had a large enough heap such that the last link item has always been free.

This problem is easily reproducible when compiling a standard kernel, no special options, then changing SETUP_MEMSIZE to a number less than 704. It seems ELKS needs around 624 byte heap minimum.

Any thoughts on how to best debug this? Any help would be appreciated. I have thoroughly tested all the other code in this PR to be correct.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ all: .config include/autoconf.h
 	$(MAKE) -C image all
 	$(MAKE) -C elksemu PREFIX='$(TOPDIR)/cross' elksemu
 
+kclean:
+	$(MAKE) -C elks kclean
+
 clean:
 	$(MAKE) -C libc clean
 	$(MAKE) -C libc DESTDIR='$(TOPDIR)/cross' uninstall
--- a/elks/Makefile
+++ b/elks/Makefile
@@ -209,6 +209,8 @@ tar:	dist
 #########################################################################
 # Configuration stuff
 
+kclean: dokclean
+
 clean:	nodep doclean
 
 dep:	$(TOPDIR)/include/autoconf.h
--- a/elks/Makefile-rules
+++ b/elks/Makefile-rules
@@ -252,6 +252,14 @@ elks:
 #########################################################################
 # Standard commands.
 
+dokclean: $(CLEANDEP)
+	rm -fv *~ *.a *.bak *.lint *.map *.o *.tmp core $(CLEANME)
+	@for DIR in */ ; do \
+		if [ -f "$$DIR/Makefile" -a "$$DIR" != "tools" ]; then \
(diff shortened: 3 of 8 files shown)


ghaerr:
Preparing for commit to allow continued development for small-memory 8018X systems.
Heap size for 8018X changed to 1024 in config.h, no observed kernel heap issues seen at that heap size.

Tracking possible kernel heap bug separately in #957.

Issue #962, opened 2021-05-25
Cross build broken by latest EMU86 config for ELKS


mfld-fr:
The automatic build of the cross tools is broken by the latest EMU86 configuration for ELKS. That configuration requires SDL2. Here is the patch to fix:
```
diff --git a/.github/workflows/cross.yml b/.github/workflows/cross.yml
index 78ceb8b1..8df8c92b 100644
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -12,8 +12,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: setup
-        run: 'sudo apt install bison flex texinfo'
+      - name: setup-1
+        run: 'sudo apt-get update'
+
+      - name: setup-2
+        run: 'sudo apt-get install libsdl2-dev'
 
       - name: checkout
         uses: actions/checkout@v2
```
Note 1: bison, flex and texinfo are reported to be already installed.
Note 2: there is still missing the `cp config-elks.mk config.mk` in the cross tools Makefile.

ghaerr:
> Note 2: there is still missing the cp config-elks.mk config.mk in the cross tools Makefile.

I had that commented out until stable version of EMU86. Shall I set it now along with this patch to the latest version?

mfld-fr:
Yes please, as I got the cross build error while trying to update to the latest EMU86.

ghaerr:
@mfld-fr : Looks like the build is failing from this PR #964 (at least on my repo), can you debug? Unfortunately I don't have time tonight to look into this, thanks.
```
2021-05-26T02:43:03.1744315Z rm -rf /home/runner/work/elks/elks/cross/build/emu86-f99937262b2566d3d20339e5b200e8483cdec3a2
2021-05-26T02:43:03.1752360Z cd /home/runner/work/elks/elks/cross/build && tar -xzf /home/runner/work/elks/elks/cross/dist/emu86-f99937262b2566d3d20339e5b200e8483cdec3a2.tar.gz
2021-05-26T02:43:03.1942546Z rm -rf /home/runner/work/elks/elks/cross/build/emu86
2021-05-26T02:43:03.1951073Z cd /home/runner/work/elks/elks/cross/build && mv emu86-f99937262b2566d3d20339e5b200e8483cdec3a2 emu86
2021-05-26T02:43:03.1968045Z touch /home/runner/work/elks/elks/cross/build/.emu86.src
[...]

PR #964, opened 2021-05-25
[cross] Update cross build to latest EMU86

ghaerr:
Adds request made in #962.


mfld-fr (changes requested):
The following line is still commented out:
```
#	cp $(BUILDDIR)/emu86/config-elks.mk $(BUILDDIR)/emu86/config.mk
 ```

EDIT: fixed by commit a982ec9

PR #967, opened 2021-06-14
[kernel] Add working SSD block device driver framework

ghaerr:
Rewritten SSD block device driver supporting 32-bit sector numbers and internal testing using main memory allocation.
Selecting 'SSD disk support' in Block Device configuration (CONFIG_BLK_DEV_SSD) enables this option, and /dev/ssd (2,0) is created for mount/umount of the block device.

Three procedures to be written (init, read, write) for actual SSD device when ported to real hardware. Discussed by @cocus in  https://github.com/jbruchon/elks/issues/932#issuecomment-846467305 as possible project.

Using 32-bit sector numbers allows for FAT32 volume support. This driver could also be used instead of ROMFS, in the case that a non-main memory mapped ROM might be desired to be used for a ROM filesystem, with a much larger size than the typical 128K max ROMFS limit due to main memory address bit shortage.

Defining TEST=1 in the driver source allows for testing the driver using allocated main memory, while developing for real hardware. The main memory is allocated by the `ramdisk` utility similar to ramdisk installation, and allows for `mkfs`, `fsck` and `mount/umount` to be used for testing.


cocus:
Pretty good! That'll be really handy for SD cards! As a side note, I have to apologize for being absent, it's just that I'm having a hard time finding a free spot on my life. Thanks!

ghaerr:
> Pretty good! That'll be really handy for SD cards! 

Yes, I designed it so that you should not have too much trouble hacking your bit-banged SD card into usefulness, hopefully. Do you happen to know the hardware interface used by that device, is it SPI?

> As a side note, I have to apologize for being absent, it's just that I'm having a hard time finding a free spot on my life.

No problems at all - let me know if you need any further help, thank you!

cocus:
> Yes, I designed it so that you should not have too much trouble hacking your bit-banged SD card into usefulness, hopefully. Do you happen to know the hardware interface used by that device, is it SPI?
> 
> >

So, SD cards can operate using SPI, or SDIO. I don't know if the latest SD cards (i.e. the ones that are pci-express) still support SPI, but at least the latest ones before PCIe do support it.
For more performance, an SDIO interface can be used, but that would require at least multiple IO lines (not just one bit), and it might be too much for these systems. SPI should be good enough!


> > As a side note, I have to apologize for being absent, it's just that I'm having a hard time finding a free spot on my life.
> 
> No problems at all - let me know if you need any further help, thank you!

Actually, thank you for all the help.

Thanks

PR #970, opened 2021-08-30
Add Dockerfile for local cross compiler and kernel compilation

chabala:
In order to get my network card working in ELKS, I needed to recompile the kernel. When I compile or install software I'm not familiar with, I tend to sandbox it in a Docker image so I can understand what the true requirements are.

Motivations:
 * a Docker image is a sharable, well-defined build environment
 * prevents polluting the host system with dependencies of the build
 * prevents having unknown system libraries affect the build
 * may be helpful for issues like #774 and #962

As an example, I started out with Ubuntu 21.04 as the base image, but discovered that code would not compile due to deprecation warnings failing the build. Then I switched to Ubuntu 20.04 based on what the last GitHub workflow used, and the compile succeeded. For Docker, this is a one line change, but downgrading my system to compile ELKS would have been a non-starter.

This is not a replacement for the GitHub workflows, but it's much easier to run locally.

I hope others find this useful, and may even want to iterate on it.

Some ideas for improvements:
 * If the final bin artifacts were placed in an empty directory, that would be a good place for a docker volume mount, and tarballing files and copying the tarball out of the container would not be necessary. Since `image` already contains files, mounting a volume there becomes much more complicated.
 * Just like how the cross directory can be built separately in the GitHub workflow and then shared instead of rebuilding, the layers in the Dockerfile can be structured to build cross first and then only rebuild it if the files needed for that layer have changed on the host. Unfortunately, I'm not an expert at Makefiles, and determining exactly which files were needed for building cross vs everything else was beyond me.
 * The flow of `source code -> docker build -> bin disk images` is somewhat broken by having to run `make menuconfig` interactively. If a `.config` is supplied, in theory the interactive steps could be skipped, but I was not able to get that working.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git
+.github
+Documentation
+Screenshots
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,44 @@
+# suggested usage: docker build --rm -t elks-dist:latest . && docker run -it --name elks-dist elks-dist:latest
+FROM ubuntu:20.04
+ENTRYPOINT ["bash"]
+# install required tools, make non-root user and switch to it
+ENV USER=builder \
+    UID=1000 \
+    GID=1000 \
+    DEBIAN_FRONTEND=noninteractive \
+    DEBCONF_NONINTERACTIVE_SEEN=true
+WORKDIR /elks
+RUN apt-get update -qq \
+ && apt-get install -y --no-install-recommends \
+  flex bison texinfo libncurses5-dev \
+  bash make g++ git libelf-dev \
+  xxd ca-certificates wget mtools \
+ && rm -r /var/lib/apt/lists /var/cache/apt/archives \
+ && addgroup \
+    --gid $GID \
+    "$USER" \
+ && adduser \
+    --disabled-password \
+    --gecos "" \
+    --home "/elks" \
+    --ingroup "$USER" \
+    --no-create-home \
+    --uid "$UID" \
+    "$USER" \
+ && chown $UID:$GID /elks
+USER $USER
+# copy in code and build cross tooling
+COPY --chown=$USER:$USER . /elks
+RUN mkdir -p "cross" \
+ && tools/build.sh
+
+# run the rest of the build interactively from step 3: https://github.com/jbruchon/elks/blob/master/BUILD.md
+# . ./env.sh
+# make menuconfig
+# make all
+
+# tarball the results and copy them out of the container
+# cd image && tar -cvzf binfiles.tar.gz *.bin
+
+# outside the container, on the host
+# docker cp elks-dist:/elks/image/binfiles.tar.gz .


ghaerr:
Hi @chabala,

Thanks for your Docker workflow contribution and interesting comments. I'm out on vacation in the Idaho wilderness for a few more days and will look at this in more detail upon my return.

Thank you!

ghaerr:
Hello @chabala,

Thanks again for your interesting contribution. This could help with our various build issues that come up from time to time, especially for testing with different or upgraded compilers. I'm not that familiar with Docker, but your script looks very straightforward and will be committed after some discussion. Here's my thoughts on some of your ideas for improvement:

> The flow of source code -> docker build -> bin disk images is somewhat broken by having to run make menuconfig interactively. If a .config is supplied, in theory the interactive steps could be skipped, but I was not able to get that working.

Running `make config` will generate a default .config file from the command line. This still requires hitting return for each configurable default, so running `yes $'\n' | make config` works for me to automatically generate a .config file in a script. Would you like to make that change and test that it works for you?

> Just like how the cross directory can be built separately in the GitHub workflow and then shared instead of rebuilding, the layers in the Dockerfile can be structured to build cross first and then only rebuild it if the files needed for that layer have changed on the host. Unfortunately, I'm not an expert at Makefiles, and determining exactly which files were needed for building cross vs everything else was beyond me.

Yes, that would be a good idea preventing the very time-consuming task of Docker rebuilding the cross/ directory if it doesn't need to. The tools/Makefile contains various commit version numbers for the various components. This Makefile is a bit complicated and was contributed by @tkchia. Perhaps he can comment on how much work it might take to get this working in a Dockerfile script. Since running `tools/build.sh` already won't rebuild components in the cross/ directory when not necessary, would keeping these existing files be kept in subsequent Docker image runs solve that problem? Where/how would they be kept?

> If the final bin artifacts were placed in an empty directory, that would be a good place for a docker volume mount, and tarballing files and copying the tarball out of the container would not be necessary. Since image already contains files, mounting a volume there becomes much more complicated.

I'm not quite following you on this. Are you talking about the elks/target/bin directory here, which contains the ELKS standard application set, or instead the image/fd1440.bin kernel+application boot image stored in image/? It would be relatively easy to move the final fd\*.bin images out of the image/ subdirectory, perhaps in a new upper-level target-image directory. I assume this would be for using the images after a full Docker build.

Thank you!

chabala:
> Running make config will generate a default .config file from the command line. This still requires hitting return for each configurable default, so running yes $'\n' | make config works for me to automatically generate a .config file in a script. Would you like to make that change and test that it works for you?

My thought here is that I needed to run menuconfig interactively first to pick my options, since I wanted to use SMC networking. So hardcoding the Dockerfile to use a default config would seem to be of limited use.

I saved my .config and tried including it in subsequent runs, and skipping `make menuconfig` , but then the build failed with some sort of KeyMaps error, as I recall, so I think that `make menuconfig` step has unexpected side effects, at least, unexpected by me. I punted on trying to solve this on the first pass, as it might need restructuring of the Makefiles. The Dockerfile still has value as a repeatable process even if the last few steps have to be done interactively.

> Since running tools/build.sh already won't rebuild components in the cross/ directory when not necessary, would keeping these existing files be kept in subsequent Docker image runs solve that problem? Where/how would they be kept?

Essentially, any time you make a change and need to rebuild a Docker image, Docker will start from scratch and walk through the steps of the Dockerfile, but it will check each line of the file for:

1. Do I have this layer of the image in my cache?
2. Did any of the files needed to build this layer change?

And if nothing changed, that layer will not be rebuilt, and the cached layer will be used. The issue is in this line https://github.com/jbruchon/elks/pull/970/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R31

Because I've copied `.` into the image, any changes in the source directory will invalidate the cache at that layer, which means all of cross would be rebuilt. Any caching logic in tools/build.sh won't matter, because inside the Docker image, everything is starting from scratch. The solution is to break up the COPY command, so that the first COPY only copies in exactly what's needed for building cross, and nothing more. Then, if those files should change, Docker would be correct in needing to rebuild cross. A later COPY command would bring in the rest of the source tree for the proper elks build, which would be the layer most often in need of being rebuilt.

I know for sure `tools/build.sh` is needed, but it sources `/env.sh`, and then calls `tools/Makefile`, which includes `/Make.defs`, ... which became a bit much for me to follow. If the building of cross were more self-contained, it would be easier to reason about. Again, this limits the usefulness somewhat because the Dockerfile hits 20 mins of building cross much more often than it needs to, but fixing it may require restructuring the Make system, which was out of scope for me.

> Are you talking about the elks/target/bin directory here, which contains the ELKS standard application set, or instead the image/fd1440.bin kernel+application boot image stored in image/?

Yes, the latter. For me at least, the boot images are the final artifacts of interest to be produced by the build, and represent the things I care about retrieving from it. As you suggest, a new top level directory could work, but even just a known empty subdirectory of `image` would be good as well. If I knew the boot images would be placed in `image/bootfs` or something, I could mount that directory as a volume on the docker host, and the files would magically appear there after the container finished running.

In summary, the things I left in 'ideas for improvements' would be nice, but will likely require some level of project restructuring, which is unfair of me to ask as a random contributor. I just wanted to get a clean build running locally.

ghaerr:
> I saved my .config and tried including it in subsequent runs, and skipping make menuconfig , but then the build failed with some sort of KeyMaps error, as I recall, so I think that make menuconfig step has unexpected side effects, at least, unexpected by me.

The make menuconfig and make config mechanism also build a config.h if the .config has changed; that's probably the problem. A way to make the system work correctly would be to `cp your_config .config` or `touch .config` and the system will automatically run make config and update the config.h file.

> a new top level directory could work, but even just a known empty subdirectory of image would be good as well. If I knew the boot images would be placed in image/bootfs or something, I could mount that directory as a volume on the docker host, and the files would magically appear there after the container finished running.

Perhaps a simple fix for this would be to use an environment variable override in the make, such as `mkdir /elks/bootfs; make IMG_DIR=/elks/bootfs` that would override the default setting of IMG_DIR=$(TOPDIR)/image in image/Make.defs (line 9).

chabala:
> The make menuconfig and make config mechanism also build a config.h if the .config has changed; that's probably the problem. A way to make the system work correctly would be to cp your_config .config or touch .config and the system will automatically run make config and update the config.h file.

I can experiment with this some more. I was including my configuration file as `.config`, but perhaps using `make config` instead of `make menuconfig` would do the trick. At that point, I may need to consider if there should be two Dockerfiles, a fully autonomous build as well as one for interactive use, or if one Dockerfile could be configured for both use cases.

> Perhaps a simple fix for this would be to use an environment variable override in the make ...

Perhaps. It does look like `IMG_DIR` is only used as a destination: https://github.com/jbruchon/elks/blob/1865cc9d9f6284f45a07c874ca6f8f97eff4b256/image/Make.defs#L9

I'm not a Makefile expert, would that line need to be wrapped in `ifndef IMG_DIR` if we propose to redefine it dynamically?

ghaerr:
> I was including my configuration file as .config, but perhaps using make config instead of make menuconfig would do the trick. 

Perhaps keep configuration files as config.\*, then pass the desired configuration file as a Docker parameter, e.g. `cp config.wd .config` for a WD network build, and have the (usually known working) default generation of .config by `make config` with no Docker config parameter. Another option (which I'm not crazy about) would be using `sed` to directly change a .config parameter to a desired value. We don't really have a good solution for automating various configuration builds (yet), but such a thing would be very useful for testing.

> would that line need to be wrapped in ifndef IMG_DIR if we propose to redefine it dynamically?

No. When a parameter is passed to make in the form of VAR=value, make forces any VAR variable assignments inside Makefiles to use the command line value. I just tried "make IMG_VAR=/tmp" from the top level make and fd1440.bin is created in /tmp.

ghaerr:
Hello @chabala,

Are you still thinking of enhancing this PR with the some of the enhancements discussed, or should I commit this as a work in progress to be enhanced later?

Thank you!

chabala:
Sorry for the silence, I have been a little distracted with other tasks.

I did make some progress with the improvements we discussed, starting with using `IMG_DIR` to relocate the output filesystem images into a volume mounted from the docker host, but it suffers from a strange issue where the build fails after making the minix images. However, on a second run, both the minix and fat images are created. My hunch is that the `IMG_DIR` variable is not being used everywhere in the build and it fails to find something it expected to exist, and/or that some artifact from the first run is helping the second run to complete.

Without fully understanding the issue, this inconsistency seems like a negative. I'd be in favor of merging this as is for now, and I could then commit my in-progress enhancement to the branch to get more eyes on it.

ghaerr:
Hello @chabala,

> I'd be in favor of merging this as is for now, and I could then commit my in-progress enhancement to the branch to get more eyes on it.

Ok, I'll merge this now. Please post a followup PR with the in-progress enhancement(s) and we can discuss ongoing issues there.

Thank you!

PR #971, opened 2021-09-26
8018x: timer 1 and 2 + irqs rework

cocus:
Adds proper handling of the Timer 1 and Timer 2 on the 8018x platforms,
allowing the system to have timers. This is particularly useful so
polling drivers can schedule some cpu time, for instance the kbd-poll
driver which is currently used on this platform.

### DIFF
--- a/elks/arch/i86/kernel/irq-8018x.c
+++ b/elks/arch/i86/kernel/irq-8018x.c
@@ -11,11 +11,9 @@
 #include <linuxmt/sched.h>
 #include <linuxmt/types.h>
 
+#include <arch/ports.h>
 #include <arch/io.h>
 #include <arch/irq.h>
-
-/* TMR and SER interrupts are enabled by default */
-unsigned int cache_08 = 0xfa;
 
 /*
  *	Low level interrupt handling for the X86 8018X
@@ -24,18 +22,16 @@ unsigned int cache_08 = 0xfa;
 
 void init_irq(void)
 {
+    /* Mask all interrupt sources */
+    outw(0xfd, 0xff00 + 0x08); /* IMASK */
 }
 
 void enable_irq(unsigned int irq)
 {
-    unsigned int mask;
-
-    mask = ~(1 << irq);
-
-    cache_08 &= mask;
-
-    /* TODO: Create a list of PCB registers instead of using hardcoded offsets */
-    outw(cache_08, 0xff00 + 0x08); /* IMASK */
+    if (irq == TIMER_IRQ) {
+        /* 8018x: Timer Interrupt unmasked, priority 7 */
+        outw(7, 0xff00 + 0x12); /* TCUCON */
+    }
 }
 
 int remap_irq(int irq)
@@ -46,8 +42,10 @@ int remap_irq(int irq)
 // Get interrupt vector from IRQ
 
 int irq_vector (int irq)
-	{
-	// IRQ 0-7  are mapped to vectors INT 08h-0Fh
-
-	return irq + 0x08;
-	}
+{
+    if (irq == TIMER_IRQ) {
+        /* 8018x: Timer 1 Interrupt => Interrupt type (vector) 18 */
+        return 18;
+    }
+    return -EINVAL;
+}
--- a/elks/arch/i86/kernel/timer-8018x.c
+++ b/elks/arch/i86/kernel/timer-8018x.c
@@ -27,12 +27,12 @@
 #define TIMER2_INTERVAL (CONFIG_8018X_FCPU * 1000000L / (4 * 1000))
 
 /*
- * With a reference of 1ms (1kHz) coming into Timer0, calculate how
+ * With a reference of 1ms (1kHz) coming into Timer1, calculate how
  * many counts are needed to achieve the "HZ" rate (usually 100Hz).
  *
- * Timer0Interval = 1kHz / HZ
+ * Timer1Interval = 1kHz / HZ
  */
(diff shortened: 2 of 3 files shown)


ghaerr:
Hello @cocus,

Thanks for the working update!

There's still quite a bit of cleanup needed, as discussed in https://github.com/jbruchon/elks/issues/960#issuecomment-914650174. However, since this is the first time a timer interrupt has been made to work on the 8018X platform, we might want to commit first and cleanup afterwards, so that we have a working base version. I can provide you with other details we'll need to see in a more evolved implementation in the discussion in #960. Which approach would you like to take?

> allowing the system to have timers. This is particularly useful so
polling drivers can schedule some cpu time

Actually, ELKS requires the timer interrupt in order to implement multi-tasking, so it's not just required for the polling keyboard driver. Thus, this PR is essential for any 8018X port.

Thank you!

cocus:
Yeah, I know that I should remove the LUT, but at the same time, it might be improved upon on the next cleanup. It'd be needed for other things in the OS that do have IRQs anyway (like the proper IRQ-driven serial port, I think).

Probably we should get the proper serial console working instead of the polled one we have enabled right now.
What do you think?

> Actually, ELKS requires the timer interrupt in order to implement multi-tasking, so it's not just required for the polling keyboard driver. Thus, this PR is essential for any 8018X port.

Yeah, I understand, I wanted to say that this timer is essential for the entire OS, and it's not something a 8018x needs to have set up.

ghaerr:
> Yeah, I know that I should remove the LUT, but at the same time, it might be improved upon on the next cleanup. 

There's no need for a table-driven lookup, especially when the contents need to be changed whenever TIMER_IRQ is redefined, which is in a seperate file. As it stands now, we have the IRQ number in ports.h, and the vector number hard-coded, which works poorly since the two go together. Also, both routines are supposed to work with a dynamically-passed IRQ number passed as a parameter, and the table lookup is already being special cased with an if statement.

I suggest keeping things extremely simple for now, and only add complexity when required. See https://github.com/jbruchon/elks/issues/960#issuecomment-914650174.

The immediate question is, would you like this committed now, then cleanup, or vice versa?

> It'd be needed for other things in the OS that do have IRQs anyway (like the proper IRQ-driven serial port, I think).

Yes. The "generic" irq-8018x.c file is supposed to be independent of the particular devices such as timer and serial port. But the 8018X interrupt mapping ties them together more tightly, since those devices are implemented on-chip. I'm thinking of combining the both 8081X specific files into a single irqtimer-8018x.c, which will reduce having things split across multiple files for such a small amount of code.


> Probably we should get the proper serial console working instead of the polled one we have enabled right now.
> What do you think?

We should probably clean up the interrupt/timer management first. I can do that, if you'd like to start working on an interrupt-driven serial port. Either is OK, I don't want to have committed code waiting for cleanups too long though.

Thank you!

cocus:
I've updated it, removing the LUT, since as you said, each case is being handled with an if(). I think this is okay now.

ghaerr:
Thanks, that looks better. 

I do have a question about `init_irq()`: the comment says mask all interrupts, but an 0xFD is being sent, which leaves one interrupt unmasked. Will this code work if an 0xFF is sent? (I don't have the 8018X interrupt controller manual page open right now). 

I'm not sure what is being preset by your boot code, but in general the kernel should not make assumptions about that.

Also, what is the purpose of the CONFIG_8018X_FCPU define in config.h? It (along with CONFIG_8018X_EB) isn't used anywhere.

cocus:
> I do have a question about `init_irq()`: the comment says mask all interrupts, but an 0xFD is being sent, which leaves one interrupt unmasked. Will this code work if an 0xFF is sent? (I don't have the 8018X interrupt controller manual page open right now).

That is because the IMASK register has an unused bit at that location. The datasheet states you should write a zero to these bits to ensure compatibility.


> I'm not sure what is being preset by your boot code, but in general the kernel should not make assumptions about that.

Just the serial port, and the chip select ranges for the FLASH and RAM. Nothing related to the timers nor interrupts.




> Also, what is the purpose of the CONFIG_8018X_FCPU define in config.h? It (along with CONFIG_8018X_EB) isn't used anywhere.

It is used to calculate a prescaler value for the Timer2. In any case:
```
/*
 * Main goal is to have Timer2 generating a 1ms period signal, so
 * calculate the appropriate value of the prescaler counter.
 * Note that 1kHz corresponds to 1ms period. Timer2 is clocked
 * by 1/4 of FCPU.
 *
 * Timer2Interval = (FCPU / 4) / 1kHz
 */
#define TIMER2_INTERVAL (CONFIG_8018X_FCPU * 1000000L / (4 * 1000))
```

I just figured out I forgot to rename the usage of Timer1 instead of Timer0 on these comments. I'll push again changing that comment to "timer 1" instead of timer 0.

ghaerr:
Thanks, I'll commit this now.

Did you ever figure out why timer 0 never worked?

cocus:
> Did you ever figure out why timer 0 never worked?

I'm really puzzled on why it doesn't work. Maybe some testing on other device might show us what was going on, maybe my CPU is faulty, who knows! @mfld-fr did you encounter any issues when using Timer 0 on your R188 processor?

mfld-fr:
Hello ! No, timer 0 looks fine on my side. Maybe you found the reason why your SBC was trashed 😃 ...

cocus:
It wasn't trashed, just phased out. It was part of an HDSL modem converter to an E1 line, something that died at least 15 years ago and got scrapped but I got it along some other equipment :)

PR #974, opened 2021-10-12
[net] Runtime selection of network card I/O-port in /bootopts (netport=)

Mellvik:
Runtime selection of network card I/O port is now supported. 

The /bootopt syntax is similar to @ghaerr's `netirq=`option: `netport=`. Tested on qemu and physical hardware.

Summary:
- Minimal changes to `init/main.c`, mainly a simple ascii-to-hex conversion routine.
- Many cleanups in `ne2k-asm.S`, removing leftovers from previous debugging sessions.
- The default MAC address is back in `ne2k.c`.
- The `netport=` command in `/bootopts` will accept '0x' prefix and no prefix, both will be considered hex.
- The new /bootopts file includes a sample config line with both `netirq=` and `netport=`, but is pushing its size limit, weighing in at 255 bytes!
- The ne2k driver will correctly report a missing NIC if the I/O port is wrong, but will not detect and erroneous IRQ.

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -15,7 +15,7 @@
 
 // register array - offset from base I/O address
 
-io_ne2k_command    = 0x00  // command
+io_ne2k_command    = 0x00  // command register at base address
 io_ne2k_rx_first   = 0x01  // page 0
 io_ne2k_rx_last    = 0x02  // page 0
 io_ne2k_rx_get     = 0x03  // page 0
@@ -92,9 +92,7 @@ ne2k_addr_set:
 
 	mov     4(%bp),%si
 
-	mov	net_port,%dx
-	// Since $io_ne2k_command is ZERO, the 'add' is a 'noop'
-	//add	$io_ne2k_command,%dx
+	mov	net_port,%dx	// command-register
 	mov	$0x42,%al	// page 1
 	out	%al,%dx
 
@@ -112,8 +110,7 @@ ems_loop:
 	inc     %dx
 	loop    ems_loop
 
-	mov	net_port,%dx
-	//add	$io_ne2k_command,%dx
+	mov	net_port,%dx	// command register
 	mov	$0x02,%al	// back to pg 0
 	out	%al,%dx
 
@@ -178,15 +175,14 @@ dma_write:
 
 	// start DMA write
 
-	mov	net_port,%dx
-	//add	$io_ne2k_command,%dx
+	mov	net_port,%dx	// command register
 	mov	$0x12,%al	
 	out     %al,%dx
 
 	// I/O write loop
 
 	mov	net_port,%dx	// Do this before changing the data segment
-	push	%dx
+	push	%dx		// Command register
 	add	$io_ne2k_data_io,%dx
 	mov	current,%bx		// setup for far memory xfer
 	mov	TASK_USER_DS(%bx),%ds
@@ -241,8 +237,7 @@ dma_r:	// Use the send data command to read exactly one backet,
 	mov	$0x0f,%al	// prep for using the 'send packet' cmd
 	out	%al,%dx
 
-	mov	net_port,%dx
-	//add	$io_ne2k_command,%dx
+	mov	net_port,%dx	// command register
 	mov	$0x18,%al	// send packet
 	out	%al,%dx
 	// now the dma does the rest, and an RDC interrupt is fielded when complete
@@ -301,8 +296,7 @@ buf_local:
 
 	// start DMA read
 
-	mov	net_port,%dx
-	//add	$io_ne2k_command,%dx
+	mov	net_port,%dx	// command register
 	mov	$0x0a,%al	// RD0 & STA
 	out     %al,%dx
 
@@ -351,8 +345,7 @@ check_dma_r:
 
 ne2k_getpage:
 	mov	$0x42,%al		// page 1
-	mov	net_port,%dx
-	//add	$io_ne2k_command,%dx
+	mov	net_port,%dx	// command register
 	out	%al,%dx
 
 	mov	net_port,%dx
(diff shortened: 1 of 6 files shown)


Mellvik:
@ghaerr - I know we've been down this road before, but still: Is it time to increase the size of `/bootopts`? Its value as an example cannot be overstated, but it's also very easy to screw up by hitting the ceiling so to speak- 

BTW - Apologies for taking so long to complete this. I spent inordinate amounts of time on a minor snag: The change of data segment in the driver (for the far buffers) - which missed by attention and in a couple of places caused the kernel variable `net_port` to contain garbage... so obvious yet so elusive. 

--Mellvik

ghaerr:
Hello @Mellvik,

Thank you for this very nice enhancement and good to see you back here!

Overall, looks great and well done. A few questions:
- What is the change regarding "adding back" the default MAC address?
- Is there any chance ne2k_command could be anything other than 0?
- There's a problem in `hex2i`: line 430 checks ahead of the string 2 bytes for an 'x'; should the string be just `netport=` (nul terminated) and 'x' happen to be in RAM, this would cause the kernel to crash on startup.  Also, 434-436 has duplicated code which could probably be more efficient. `hex2i` should be declared INITPROC, like `atol` since they're only used once at init time and will eventually be discarded. I think we can probably find a much smaller hex to ascii routine, let me look around a bit.
- Line 444: Error checking the netport address range doesn't belong in a hex conversion routine, but somewhere else. I suggest just dropping it entirely since we're not checking input anyways.

>  Is it time to increase the size of /bootopts? Its value as an example cannot be overstated, but it's also very easy to screw up by hitting the ceiling so to speak-

There really is no good way to do this that I can see: the limit is caused because we're "stealing" a max 256-byte low memory area at segment 0x50, and the kernel is loaded directly above that. We can probably cut back some of the sample comments so that the user has room to use /bootopts without inadvertently busting the limit.

Thank you!

Mellvik:
Good to see you too, @ghaerr!

> 
> * What is the change regarding "adding back" the default MAC address?

It has been commented out for a while (by me). Then I needed it when doing some testing in VirtualBox (unsuccessful btw), so I put it back in a way that didn't use any RAM. 6 bytes here, 6 bytes there :-)

> * Is there any chance ne2k_command could be anything other than 0?

No. It's always and per definition at the IO port base address.

> * There's a problem in `hex2i`: line 430 checks ahead of the string 2 bytes for an 'x'; should the string be just `netport=` (nul terminated) and 'x' happen to be in RAM, this would cause the kernel to crash on startup.  Also, 434-436 has duplicated code which could probably be more efficient. `hex2i` should be declared INITPROC, like `atol` since they're only used once at init time and will eventually be discarded. I think we can probably find a much smaller hex to ascii routine, let me look around a bit.

I'll await your input on that one.

> * Line 444: Error checking the netport address range doesn't belong in a hex conversion routine, but somewhere else. I suggest just dropping it entirely since we're not checking input anyways.

OK

> 
> > Is it time to increase the size of /bootopts? Its value as an example cannot be overstated, but it's also very easy to screw up by hitting the ceiling so to speak-
> 
> There really is no good way to do this that I can see: the limit is caused because we're "stealing" a max 256-byte low memory area at segment 0x50, and the kernel is loaded directly above that. We can probably cut back some of the sample comments so that the user has room to use /bootopts without inadvertently busting the limit.

OK, maybe it's a better idea to add a second file, /bootopts.sample to assume the role of education and reference?

--Mellvik

ghaerr:
> No. It's always and per definition at the IO port base address.

Ok. We should probably just continue cleanup of ne2k.S by removing the commented-out add instructions, and just indicate in a comment in the `mov netport,%dx // get base address` line or similar. The less commented-out code the better and easier to read, in asm files, IMO.

> maybe it's a better idea to add a second file, /bootopts.sample to assume the role of education and reference?

That's a great idea, which can be added in a seperate PR.

ghaerr:
Hello @Mellvik,

> I'll await your input on that one.

I've got a simple `strtol` that we can use that will replace/augment both your `hex2i` as well as the current `atol` and `atoi` which will allow any /bootopts parameter to be specified in base 8, 10 or 16.

Since it involves changing out various other routines, I'll wait for you to push the couple changes described above, then commit this PR with `hex2i` as-is, and I'll submit a separate PR for the `strtol` enhancement.

Thank you!

Mellvik:
Sounds good, @ghaerr - 

PR update coming. A sketchy /bootopts.sample included as a starting point.


—Mellvik

Mellvik:
@ghaerr -- 
I decided to leave out the new `bootopts.sample` file from the PR update. Turns out that 'bootopts.sample' is one char too long for the minix FS, so there is a decision to be made: Keep 'bootopts.sampl' or change it to something else, like 'bootopt.sample', 'bootopts.info' or something else. My favorite is the first - 'bootopts.sampl'.

When the sample file has been added, we should immediately remove some of the sample lines from the real bootopts file. Needless to say, I'm hitting the ceiling all the time... :-)

--Mellvik

FWIW - the first cut sample file looks like this:
```
#
# bootopts quick reference
# Refer to the ELKS Wiki for more details about boot options.
#
# NOTE: The /bootopts file MUST start with a double hash (##) and
#	be max 255 bytes in size, otherwise it will be ignored.
#	Study the boot messages if things don't work as expected.
#
# Sample /bootopts file, '#' starts a comment:

##
console=ttyS0 debug net=eth 3 # condensed
#init=/bin/init 3 n	# multiuser serial no sysinit
#init=/bin/sh		# singleuser shell
#console=ttyS0,19200	# serial console
#root=hda1 ro	# hd partition 1, read-only root
#netirq=9 netport=0x300	# set NIC parms
```

ghaerr:
Hello @Mellvik,

Thanks for the cleanup, I will commit this now, followed by a replacement `strtol` for atoi/atol/hex2i in a seperate PR.

> I decided to leave out the new bootopts.sample file from the PR update. Turns out that 'bootopts.sample' is one char too long for the minix FS, so there is a decision to be made

Ok, I will think further on this, as we also have the FAT filesystem 8.3 issue for some filesystem operations.

Thank you!

PR #975, opened 2021-10-14
[kernel] Add simple_strtol, replaces atoi, atol, hex2i

ghaerr:
Adds ability to decode ascii strings as decimal, octal and hex values. Programmatic second parameter to `simple_strtol` specifies base, use 0 to allow any base, depending on prefix (0x or 0X for hex, 0 for octal, otherwise decimal).

Replaces `int atoi()`, `long atol()`, and `int hex2i()`.

Total overall 'cost': only 112 byte kernel size increase, because three prior routines were able to be completely removed.

Currently, only /bootops `netport=` parameter allows specification in other than decimal.

@Mellvik: I also added a couple of parens to stop compiler warnings in ne2k.c and also always output the network port and IRQ number, even when the NIC isn't responding. This allows easier catching of improper settings.

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k.c
+++ b/elks/arch/i86/drivers/net/ne2k.c
@@ -366,16 +366,17 @@ void ne2k_drv_init(void)
 		ne2k_get_hw_addr(prom);
 
 		/* If there is no prom (i.e. emulator), use default */
-		if ((prom[0]&0xff == 0xff) && (prom[1]&0xff == 0xff)) {
+		if (((prom[0]&0xff) == 0xff) && ((prom[1]&0xff) == 0xff)) {
 			err = -1;
 		} else {
 			while (i < 6) hw_addr[i] = prom[i]&0xff,i++;
 			err = 0;
 		}
 
+		printk ("eth: NE2K at 0x%x, irq %d, ", net_port, net_irq);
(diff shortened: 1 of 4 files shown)


Mellvik:
Thank you @ghaerr,
Also for the added cleanups in ne2k.c. I must be using a different toolchain since I have not been getting any warnings at all. Or maybe we use different settings?
The IRQ/PORT reporting if NIC unresponsive was already there, but the (admittedly smart) changes reduces the overall footprint. 

The new strtol is indeed a compact piece of code, where did you find it? Or did you write it?

Thank you

--Mellvik

ghaerr:
>  I must be using a different toolchain since I have not been getting any warnings at all. Or maybe we use different settings? 

That is strange... not sure the reason for that.
```
$ ia16-elf-gcc --version
ia16-elf-gcc (GCC) 6.3.0
```


> The IRQ/PORT reporting if NIC unresponsive was already there, but the (admittedly smart) changes reduces the overall footprint. 

The reason for that change was to report the port/irq setting in the case where the NIC was unresponsive.

> The new strtol is indeed a compact piece of code, where did you find it? Or did you write it? 

First looked at a couple smaller libc versions to get a general approach, then wrote it using code from the kernel `atoi` and stripping down a libc version, along with an idea from your `hex2i` :)

PR #977, opened 2021-10-18
[cmd] urlget: Outgoing ftp file transfer (ftpput), fixes

Mellvik:
- New functionality: `ftpput` enables outgoing file transfer via the FTP protocol
- Fixes: improved error messages and error handling
- If the incoming data cannot be saved, `urlget` (in ftp mode) gobbles up the rest of the data stream to avoid congestion.
- See the wiki for details about `ftpput` usage.

### DIFF
--- a/elkscmd/inet/urlget/urlget.c
+++ b/elkscmd/inet/urlget/urlget.c
@@ -2,16 +2,17 @@
  *
  * 04/05/2000 Michael Temari <Michael@TemWare.Com>
  * 09/29/2001 Ported to ELKS(and linux) (Harry Kalogiroy <harkal@rainbow.cs.unipi.gr>)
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @Mellvik,

Looks great, thanks for the urlget enhancements!

Also glad to hear that purging fixes the lingering connection problem discussed in https://github.com/jbruchon/elks/issues/976#issuecomment-945753623. Does this handle both cleanup/close cases in #976, or only the 2nd one?

> In order to create a viable test scenario, we need a special version of urlget - with just the gobbling statement commented out. How to create that on QEMU is a different story -- ideas solicited.

How about a '-d' (debug) option that disables the purging for testing?

Would you like this committed now, or do you want to push additional changes or enhancements?

Thank you!

Mellvik:
@ghaerr,
Just adding a -d is great - let me put it in before you commit.

Thank you.

-- Mellvik

PR #979, opened 2021-10-22
[kernel][cmds] Properly cleanup TCP connection on close/exit

ghaerr:
Proposed fix for #976, allows proper FIN and CLOSE wait timeouts on socket close and/or application exit.

Enhances `netstat` to show timer values for better understanding of TCP wait states and possible fine tuning.
Improved debug display for both kernel and `ktcp`.

Also adds (commented out) sending RST on receipt of refused packets.

Internal structure changes, `make clean` after pull before compiling.

@Mellvik: although this looks like lots of changes, almost all is reformatting of kernel debug messages to display process IDs rather than internal buffer addresses for ease of tracing. Also added DEBUG_CLOSE (default on) option to `ktcp` to trace close activity on ^P (DEBUG_TCP also on).

I also added lines 431-442 in elkscmd/ktcp/tcp.c (default commented out) to send an RST just after printing "Refusing packet". Unfortunately, I am not able to test this, so I'd like you to test `urlget` as previously discussed both with and without the RST sending turned on. Since there is no TCP control block when we need to send the RST, it has to be built from scratch, and there's no way of seeing it unless you've got an external packet tracer. I could have possibly got the network/host byte order incorrect or an internal control block field improper which might prevent the single RST from going on the wire.

If the RST works and is left in, we should perform further normal testing of say, streaming `ls -l` telnet while ^D typed, or other fast application exits while data may be still outgoing after the application has exited. We can always leave the RST code out for the time being.

### DIFF
--- a/elks/arch/i86/drivers/char/tcpdev.c
+++ b/elks/arch/i86/drivers/char/tcpdev.c
@@ -44,7 +44,7 @@ char *get_tdout_buf(void)
 static size_t tcpdev_read(struct inode *inode, struct file *filp, char *data,
 		       unsigned int len)
 {
-    debug("TCPDEV: read( %p, %p, %p, %u )\n",inode,filp,data,len);
+    debug("TCPDEV(%d) read %u\n", current->pid, len);
     while (tdout_tail == 0) {
 	if (filp->f_flags & O_NONBLOCK)
 	    return -EAGAIN;
@@ -57,47 +57,46 @@ static size_t tcpdev_read(struct inode *inode, struct file *filp, char *data,
      *  If the userspace tcpip stack requests less data than what is in the
      *  buffer it will lose data, so the tcpip stack should read BIG.
      */
-    len = len < tdout_tail ? len : tdout_tail;
-    debug("TCPDEV: read() mark 1 - len = %u\n",len);
+    if (len < tdout_tail)
+	debug_net("TCPDEV(%d) read len too small %u\n", len);
+    else len = tdout_tail;
     memcpy_tofs(data, tdout_buf, len);
     tdout_tail = 0;
     up(&bufout_sem);
-    debug("TCPDEV: read() returning with bufout_sem = %d\n",bufout_sem);
-    if (bufout_sem > 0)
-	panic("bufout_sem tragedy");
+    debug("TCPDEV(%d) read retval %u bufout_sem %d\n", current->pid, len, bufout_sem);
+    //if (bufout_sem > 0)
+	//panic("bufout_sem tragedy");
     return len;
 }
 
-int tcpdev_inetwrite(char *data, unsigned int len)
+int tcpdev_inetwrite(void *data, unsigned int len)
 {
-    debug("TCPDEV: inetwrite( %p, %u )\n",data,len);
+    debug("TCPDEV(%d) inetwrite %u\n", current->pid, len);
     if (len > TCPDEV_OUTBUFFERSIZE) {
-	debug_net("tcpdev_inetwrite: len too large\n");
+	debug_net("TCPDEV(%d) inetwrite len too large %u\n", current->pid, len);
 	return -EINVAL;
     }
-    debug("TCPDEV: inetwrite() mark 1.\n");
 
     /* Data already in tdout_buf buffer */
     tdout_tail = len;
     wake_up(&tcpdevq);
-    debug("TCPDEV: inetwrite() returning\n");
     return 0;
 }
 
 void tcpdev_clear_data_avail(void)
 {
     up(&bufin_sem);
-    debug("TCPDEV: clear_data_avail()\n");
-    if (bufin_sem > 0)
-	panic("bufin_sem tragedy");
+    debug("TCPDEV(%d) clear data_avail\n", current->pid);
+    //if (bufin_sem > 0)
+	//panic("bufin_sem tragedy");
 }
 
 static size_t tcpdev_write(struct inode *inode, struct file *filp,
 			char *data, size_t len)
 {
-    debug("TCPDEV: write( %p, %p, %p, %u )\n",inode,filp,data,len);
+    debug("TCPDEV(%d) write %u\n", current->pid, len);
     if (len > TCPDEV_INBUFFERSIZE) {
-	debug_net("tcpdev_write len too large\n");
+	debug_net("TCPDEV(%d) write len too large %u\n", len);
 	return -EINVAL;
     }
     if (len > 0) {
@@ -109,7 +108,7 @@ static size_t tcpdev_write(struct inode *inode, struct file *filp,
 	/* Call the af_inet code to handle the data */
 	inet_process_tcpdev((char *)tdin_buf, len);
     }
-    debug("TCPDEV: write() returning %u\n", len);
+    debug("TCPDEV(%d) write retval %u\n", current->pid, len);
     return len;
 }
 
@@ -117,41 +116,39 @@ static int tcpdev_select(struct inode *inode, struct file *filp, int sel_type)
 {
     int ret = 0;
 
-    debug("TCPDEV: select( %p, %p, %d )\n",inode,filp,sel_type);
     switch (sel_type) {
     case SEL_OUT:
-	debug("TCPDEV: select() chose SEL_OUT\n");
+	debug("TCPDEV(%d) select SEL_OUT\n", current->pid);
 	if (bufin_sem == 0)
 	    ret = 1;
 	else
 	    select_wait(&tcpdevq);
 	break;
     case SEL_IN:
-	debug("TCPDEV: select() chose SEL_IN\n");
+	debug("TCPDEV(%d) select SEL_IN\n", current->pid);
 	if (tdout_tail != 0)
 	    ret = 1;
 	else
 	    select_wait(&tcpdevq);
 	break;
     default:
-	debug("TCPDEV: select() chose unknown option %d.\n",sel_type);
+	debug("TCPDEV(%d) select bad type %d\n", current->pid, sel_type);
     }
-    debug("TCPDEV: select() returning %d\n",ret);
+    debug("TCPDEV(%d) select retval %d\n", current->pid, ret);
     return ret;
 }
 
 static int tcpdev_open(struct inode *inode, struct file *file)
 {
-    debug_net("TCPDEV: open(%p, %p) tcpdev_inuse %d\n",inode,file, tcpdev_inuse);
+    debug_net("TCPDEV(%d) open inuse %d\n", current->pid, tcpdev_inuse);
     if (!suser()) {
-	debug_net("TCPDEV: open() returning -EPERM\n");
+	debug_net("TCPDEV open retval -EPERM\n");
     	return -EPERM;
     }
     if (tcpdev_inuse) {
-	debug_net("TCPDEV: open() returning -EBUSY\n");
+	debug_net("TCPDEV open retval -EBUSY\n");
 	return -EBUSY;
     }
-    debug_net("TCP open OK\n");
     tdin_tail = tdout_tail = 0;
     tcpdev_inuse = 1;
     return 0;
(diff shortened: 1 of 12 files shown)


Mellvik:
Thank you @ghaerr, 
this is great work. I wouldn't stay the results are in yet, but some of them are, so here's the first round:

- Basically there's little change in the ELKS behaviour at the time of ^C and onwards
- It seems (I'll have to measure that) that transfer speed has improved (!)
- The interval between the one byte packet exchanges that seem to go on forever after the INTR, is much shorter.
- There are about 10 full size packets between the INTR and the start of the one byte packets, presumably whatever was in the server's buffers.
- I'm sure you can read a lot of interesting stuff from the netstat outputs, I'll post them and some of the TCPdump separately to get the formatting right.

I will enable the RST in the code next to see what happens.

—Mellvik

ghaerr:
Thanks for the report @Mellvik. 

> I just looked at a packet trace between two raspberrypis and a scp transfer.
When interrupted, the client immediately sends an RST. The server probably drops the connection right away, but seems to do the opposite: continues to send data, presumably because they're already in the buffer.

> The client never ACKs any data after the RST, but continues to respond to every single new data packet with the same (length zero) RST packet.

There's a question as to whether my completely untested code actually sends a RST properly as coded, but once you determine that it gets on the wire correctly, we can choose whether to send a RST "immediately", or only after any data packet is "refused". It could be that the RPI system is actually just appearing to send the RST immediately, because the very next event after the socket close is the kernel reading previously buffered data, which it immediately responds with RST.

Sounds like we'll want to send RST on refused packets as well, which is exactly what the PR code is supposed to do.

Thank you!

ghaerr:
Hello @Mellvik,

I pushed more cleanups and enhancements. Be sure to `make clean` as structure field sizes have changed.

I went ahead with adding an optional "SEND_RST_ON_CLOSE" temporary testing option, which defaults to OFF, in elkscmd/ktcp/tcpdev.c. This option allows you to experiment with what should become a `setsockopt` SO_LINGER option that would allow an application to specify whether or not to send an RST on ^C to immediately shutdown the server. You can test this both with and without the previous "\#if 0" option in tcp.c to send an RST on Refused packets, to determine how each option works best with remote servers and `urlget`.

During implementation of the SEND_RST_ON_CLOSE option using QEMU and connecting to localhost, I discovered a big bug in `ktcp`: it previously would hang the process whenever *any* RST was received from any connection! That is now fixed. I also cleaned up even more socket code.

We need to test first, the option of sending an RST on Refused packets (change \#if 0 to \#if 1 in tcp.c). That should help with your FTP server. After learning from that, we should turn on the optional SEND_RST_ON_CLOSE, which forces a RST to be sent on socket close or application exit, rather than a FIN, which is never sent with that option. I don't believe this option should be permanently set, as it could cause data loss. The plan will be to add it to certain programs for quicker cleanup of ^C handling on forced exits.

Thank you!

ghaerr:
@Mellvik,

As per my comment in https://github.com/jbruchon/elks/pull/980#issuecomment-950195603, I'm going to push another quick cleanup commit, then commit this PR and start working on the SO_LINGER option to allow the TCP stack to handle the FTP ABOR(T) and possibly other problems you've just brought up in https://github.com/jbruchon/elks/issues/976#issuecomment-950195573. I will code up a `urlget` with the linger option for your testing, after which we can determine whether #980 is necessary, or we can get back to a simple `urlget` that works under error conditions without special TCP error handling in the application itself.

Thank you!

ghaerr:
Final push moves SEND_RST_ON_CLOSE and SEND_RST_ON_REFUSED_PKT to elkscmd/ktcp/config.h. 

@Mellvik, if you have time, please test these two options with `urlget` and other ELKS clients or servers, thank you!

Next step is implementing SO_LINGER option to programmatically effect SEND_RST_ON_CLOSE in `urlget`.

Mellvik:
Thank you @ghaerr - we're getting there.

Short status on this one - and good news: Enabling SEND_RST_ON_CLOSE stops the data stream (almost) instantly. One single data packet comes in after the RESET, presumably too far along in the send chain to be stopped.

The SEND_RST_ON_REFUSET_PKT also works in the sense that the RSTs get sent, but it has no effect. It seems - from the trace - that the server keeps sending the one byte packet every 120 seconds regardless for whatever time it takes to timeout. Ignoring the RST. Further, this seems to be the FTP server, not the stack, trying to revive the connection.

IOW - it may be the best idea to just ignore these packets and eventually refer the messages to some debug level, like you suggested a few days back.

# net start                                                                     
Starting networking on eth                                                      
ktcp -b 10.0.2.15 10.0.2.2 255.255.255.0                                        
ktcp: ethernet 00.80.29.ef.d9.51 mtu 1024                                       
Starting daemons telnetd httpd                                                  
# urlget ***@***.***/pcplus.img > /dev/null                  
^C                                                                          
# tcp: refusing packet 10.0.2.2:55035->1025  	<- this is the one packet that makes it after the RST,
						It triggers and RST, but would go away after this packet anyway                                
tcp: refusing packet 10.0.2.2:52953->1027                                       
                                                                                
# tcp: refusing packet 10.0.2.2:52953->1027                                     
tcp: refusing packet 10.0.2.2:52953->1027                                       
tcp: refusing packet 10.0.2.2:52953->1027                                       
tcp: refusing packet 10.0.2.2:52953->1027 
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                


—Mellvik

ghaerr:
Hello @Mellvik,

> Short status on this one - and good news: Enabling SEND_RST_ON_CLOSE stops the data stream (almost) instantly. One single data packet comes in after the RESET, presumably too far along in the send chain to be stopped.

Good news. I will be removing the SEND_RST_ON_CLOSE compile option in my next push to #981, since we now have setsockopt to turn that on or off from the application (urlget only, for now).


> The SEND_RST_ON_REFUSED_PKT also works in the sense that the RSTs get sent, but it has no effect. It seems - from the trace - that the server keeps sending the one byte packet every 120 seconds regardless for whatever time it takes to timeout. Ignoring the RST. Further, this seems to be the FTP server, not the stack, trying to revive the connection.

There is a possibility that the new code for sending the RST in this case is incorrect - we should check this more carefully on a packet trace to see that the RST packet is properly formed with the correct source and dest addresses and ports.


> IOW - it may be the best idea to just ignore these packets and eventually refer the messages to some debug level, like you suggested a few days back.

Yes, the plan is to only show the "refusing packet" message when ^P debug is on. I would like to have networking "debug levels" but haven't figured out a way to implement that yet.

Mellvik:
@ghaerr,
I didn't quite get to finish the testing before dinner time, but here's the status (not including your latest push, the SEND_RST_ON_XXX both turned off:
- really no change from the previous report: Aborting a running transfer works perfectly fine (!!!!!). NOrmal completion does not (actually, the data transfer is fine, but the messages indicate otherwise, see previous report).
- Local write error (and no purge in urlget): Exactly as before, goes on forever with single byte packets. Urlget hangs.

So the removal of SEND_RST_ON_CLOSE=1 didn't change anything.

There is another anomaly that just came to my attention, and may be totally inconsequential (I'll look closer at this one in the morning): In a 'normal' transfer - no INTR, no error, normal completion, after ELKS receives the first FIN, the ACKed sequence number is off by one, always one ahead. Doesn't seem to do any harm at this (late) stage in the process, and the other side doesn't not seem to notice, but strange. Packet trace later.

Thank you!
--Mellvik

ghaerr:
@Mellvik, thanks for your daily testing, it helps greatly!

> -really no change from the previous report: Aborting a running transfer works perfectly fine (!!!!!).

Yeah, one problem solved!


>  Normal completion does not (actually, the data transfer is fine, but the messages indicate otherwise, see previous report).
> Local write error (and no purge in urlget): Exactly as before, goes on forever with single byte packets. Urlget hangs.

I finally figured this out as a result or your last message: urlget does an `fdopen` on the net_connect file descriptor, then `fclose` on them before the `net_close`. This has the effect of closing the underlying socket BEFORE net_close turns off send RST; thus the RST is always sent on FTP closes.

I've fixed this and will commit the PR so that we start fresh. I think this will fix it.

> So the removal of SEND_RST_ON_CLOSE=1 didn't change anything.

I think it allowed us to find this last problem in `urlget`.

> In a 'normal' transfer - no INTR, no error, normal completion, after ELKS receives the first FIN, the ACKed sequence number is off by one, always one ahead. Doesn't seem to do any harm at this (late) stage in the process, and the other side doesn't not seem to notice, but strange. Packet trace later.

Hmmm, are you sure it should be the same sequence number? All the TCP code in ktcp/tcp.c increments the number by one on an ACK received, so that the reported ACK is 1 greater than the previously received length. This is done using `cb->rcv_nxt++`.

Thank you!

Mellvik:
> In a 'normal' transfer - no INTR, no error, normal completion, after ELKS receives the first FIN, the ACKed sequence number is off by one, always one ahead. Doesn't seem to do any harm at this (late) stage in the process, and the other side doesn't not seem to notice, but strange. Packet trace later.
> 
> Hmmm, are you sure it should be the same sequence number? All the TCP code in ktcp/tcp.c increments the number by one on an ACK received, so that the reported ACK is 1 greater than the previously received length. This is done using cb->rcv_nxt++.
> 
Yes, I'm sure. And yes, this is intriguing - while possibly totally without merit or consequence.
All ACKs before the FIN are 'in line', after the FIN, one ahead. Presumably, any (?) IP implementation would be programmed to just ignore 'future' acks.
Stay tuned, details tomorrow.

--Mellvik

ghaerr:
> Hmmm, are you sure it should be the same sequence number? 

> Yes, I'm sure.

> All ACKs before the FIN are 'in line', after the FIN, one ahead

This article [What should next sequence number be after FIN](https://serverfault.com/questions/975113/what-should-be-the-next-sequence-number-after-a-packet-with-the-fin-containing-p) says that RFC 793 specifies that FIN should occupy one sequence number, which is contradicting your statement. ELKS TCP follows RFC 793, it seems. Further [research](http://www.surendranathcollege.org/new/upload/RUMA_MISTRYTCP,%20TCP%20frame%20format%20and%20Connections2020-04-08TCP.pdf) shows that a FIN+ACK packet should consume one sequence number, if it doesn't carry any data, while lone ACKs without data do not. In any case, if this proves to be a problem, lets discuss it in another issue, so we can close out the several we are working on as reported urlget bugs.

Mellvik:
Thanks you @ghaerr, this is great research (and I'm blushing for not doing it right away :-) ).

> This article What should next sequence number be after FIN <https://serverfault.com/questions/975113/what-should-be-the-next-sequence-number-after-a-packet-with-the-fin-containing-p> says that RFC 793 specifies that FIN should occupy one sequence number, which is contradicting your statement.
NO, this is actually saying that what I'm seeing is normal, not erroneous.
And I don't think this is a problem regardless, it's so close to termination of the connection that ACKs don't matter.
What's interesting though is that the server does not increment the sequence number, but the client (ELKS) is incrementing the ACK, som there is now a discrepancy.
Take a look at the trace in #981, which demonstrates this. This behaviour is entirely consistent, I've been through an handful of traces and found the exact same thing.

> ELKS TCP follows RFC 793, it seems. Further research <http://www.surendranathcollege.org/new/upload/RUMA_MISTRYTCP,%20TCP%20frame%20format%20and%20Connections2020-04-08TCP.pdf> shows that a FIN+ACK packet should consume one sequence number, if it doesn't carry any data, while lone ACKs without data do not. In any case, if this proves to be a problem, lets discuss it in another issue, so we can close out the several we are working on as reported urlget bugs.



Since the issue does not seem to have any practical consequences, I suggest we just treat it as a curiosity, not a problem. Also, it seems the server (Linux in this case), not the client may be the culprit.

—Mellvik

PR #980, opened 2021-10-23
[urlget] bugfix: Don't wait for server reply when aborting a ftp tran…

Mellvik:
`urlget` would wait for server replies to all FTP commands and hang if they don't show. 
This is guaranteed to happen when using the -d option and a local write error occurs. This change adds a ´wait' parameter to the `ftpcmd` function which will force it to skip the wait.

A more elegant solution would be to use a timeout. Maybe later.


ghaerr:
Is this solution required because of the backed-up packets on the data channel and other problems associated with #976, or will it still be required when that is solved?

Should we wait for commit until after RST testing on #979?

Thank you!

Mellvik:
@ghaerr,
Not doing this defeats the purpose of the debugging switch IMHO. So this is necessary - until we put a socket option receive timeout into the equation, which probably is the right thing to do. I didn't take the time to check if this is available in ELKS.

On my list (but admittedly, when the ftp client/server pair is ready, it may stay on the list for a while).

-- Mellvik

ghaerr:
> Not doing this defeats the purpose of the debugging switch IMHO. So this is necessary 

I am starting to fully understand the problem we are having with FTP `urglget`: aside from two major TCP bug fixes uncovered as a result of your testing and packet traces, the original problem remains: lots of data packets continue to be received from the FTP server, either after a programmatic `close` in `urlget`, or forced close via ^C. These data packets may be buffered by the network, and/or continued to be sent by the FTP server, but in both cases, `urlget` is now in the proper FIN_WAIT_2 state, having transitioned there from its socket close, and the FTP server can continue transmitting and/or its data packets continue to be received by the ELKS TCP stack. After reading the specification, it is implementation-dependent what happens when one side (client in our case) closes its half of the connection, and the server continues to write. Those packets could be ACK'd, as ELKS is doing, or could be dropped, which could result in retransmissions, but the server has not yet closed its socket, so no FIN has been received by ELKS TCP. The server does not know the client is terminated, and/or the ELKS TCP stack is ACKing data rather than sending an RST to end it all.

> until we put a socket option receive timeout into the equation, which probably is the right thing to do. 

We don't need a receive timeout option, we need an option that says "this close also means stop sending me data". That option is `setsockopt(..., SO_LINGER, [timeval=0]);` which will send an RST instead of FIN on the close or process exit. It is an abrupt TCP termination that will only be used on ^C interrupt, or error closes. If a transfer completes normally, the linger option will be turned off before the programmatic `close` in `urlget`.

The idea of having to 'purge' data in the `urlget` client is improper - we can't rewrite all our ELKS clients to consume data from a remote server because they won't quit sending data; we need to tell the lower level TCP stack to drop that data instead. This will be done automatically using the SO_LINGER close, since that will send an immediate RST, then deallocate the TCP control block. Any further data received will be immediately RST'd as well and packet dropped. The application (`urlget`) won't ever receive any further data, so no need to ever purge.

So, the idea that we should continue adding code to complete our 'debugging option' in `urlget` is a bad idea, IMO. It complicates code and now we're working around additional issues, with an option that shouldn't exist. `urlget` should just work, with or without a debug option.

For this reason, I suggest that we hold this PR, and that the TCP bug fixes, cleanups, and commented-out RST code in #979 be committed. That will allow you to test RST options, and me to implement the SO_LINGER code, which is a little tricky. I will then submit that enhancement along with proper `urlget` SO_LINGER example for you to test, after which the entire 'purge' code and option can also be deleted, and this wait option likely not needed. `urlget` will work properly in all cases without additional options.

> I didn't take the time to check if this is available in ELKS.

It's not yet! We need to add a new `setsockopt` system call, and also pass the linger timeout value to `ktcp`, which are straightforward, but I'd like to commit the previous TCP fixes/cleanup first.

Mellvik:
@ghaerr,
I'll get back to you in detail in the morning, I think we're on the right track.
Just two things 
1) like I've already pointed out, the ftp server is not a part in this, it's a stack only issue.
2) the PR is a bug fix, not an add-on or whatever. Whatever we do now or later doesn't make my error right...

--M

Mellvik:
@ghaerr, we've already moved past this point, so just a couple of observations.
> 
> The idea of having to 'purge' data in the urlget client is improper - we can't rewrite all our ELKS clients to consume data from a remote server because they won't quit sending data; we need to tell the lower level TCP stack to drop that data instead.
> 
The purge was a temporary band aid all along, I'm not aware of any suggestion to use the mechanism anywhere else. It is also harmless in the sense that it will still work when the underlying problem has been fixed.
> This will be done automatically using the SO_LINGER close, since that will send an immediate RST, then deallocate the TCP control block. Any further data received will be immediately RST'd as well and packet dropped. The application (urlget) won't ever receive any further data, so no need to ever purge.
> 
> So, the idea that we should continue adding code to complete our 'debugging option' in urlget is a bad idea, IMO. It complicates code and now we're working around additional issues, with an option that shouldn't exist. urlget should just work, with or without a debug option.
> 
Again there was no such idea from anywhere that I'm aware of. My fix PR was and is a fix to an oversight on my part, which does not qualify as 'continue adding code' to anything. So let's keep things straight.
> For this reason, I suggest that we hold this PR, and that the TCP bug fixes, cleanups, and commented-out RST code in #979 <https://github.com/jbruchon/elks/pull/979> be committed. That will allow you to test RST options, and me to implement the SO_LINGER code, which is a little tricky. I will then submit that enhancement along with proper urlget SO_LINGER example for you to test, after which the entire 'purge' code and option can also be deleted, and this wait option likely not needed. urlget will work properly in all cases without additional options.
> 
Yes, it can be deleted along with the ftp part of the -d option. 
> I didn't take the time to check if this is available in ELKS.
> 
> It's not yet! We need to add a new setsockopt system call, and also pass the linger timeout value to ktcp, which are straightforward, but I'd like to commit the previous TCP fixes/cleanup first.
> 
Sounds good. The problem that the recent fix to urlget takes care of doesn't go away until such timeout is available. If the server (or connection) at any point goes away, programs like urlget and any ftp client will hang - presumably until a timeout at a lower level will close the connection.


—mellvik

ghaerr:
> The problem that the recent fix to urlget takes care of doesn't go away until such timeout is available. If the server (or connection) at any point goes away, programs like urlget and any ftp client will hang - presumably until a timeout at a lower level will close the connection.

I see what you're saying. However I think we're getting confused on the "timeout" I'm describing. There is no socket option to take down a connection after a certain timeout. That is, a TCP connection will remain open indefinitely, until the application closes it. I wasn't proposing such a timeout, instead, am describing the functionality of the "linger" timeout field - which is used for a different purpose: when linger timeout is 0, an RST is sent on close, which is now implemented. When linger timeout is set to non-zero, the TCP stack is instructed to keep the connection alive until the last buffered packet has been sent. This is always done by ktcp as the writes are synchronous, so implementing that "timeout" would have no effect.

Thus, it seems an actual application timeout (alarm signal callback) would be the proper fix for the urlget issue you're discussing. This would then clean things up rather than hang. I'm not sure whether the ability to not wait for an FTP command result is proper in that case, but I'll leave that to you.

Mellvik:
@ghaerr,

good point. We are indeed talking about two very different timeouts, but from the dialogue it seems we're on the same page all the time. 

> 24. okt. 2021 kl. 16:48 skrev Gregory Haerr ***@***.***>:
> 
> 
> The problem that the recent fix to urlget takes care of doesn't go away until such timeout is available. If the server (or connection) at any point goes away, programs like urlget and any ftp client will hang - presumably until a timeout at a lower level will close the connection.
> 
> I see what you're saying. However I think we're getting confused on the "timeout" I'm describing. There is no socket option to take down a connection after a certain timeout. That is, a TCP connection will remain open indefinitely, until the application closes it. I wasn't proposing such a timeout, instead, am describing the functionality of the "linger" timeout field - which is used for a different purpose: when linger timeout is 0, an RST is sent on close, which is now implemented. When linger timeout is set to non-zero, the TCP stack is instructed to keep the connection alive until the last buffered packet has been sent. This is always done by ktcp as the writes are synchronous, so implementing that "timeout" would have no effect.
> 
> Thus, it seems an actual application timeout (alarm signal callback) would be the proper fix for the urlget issue you're discussing. This would then clean things up rather than hang. I'm not sure whether the ability to not wait for an FTP command result is proper in that case, but I'll leave that to you.
> 
From the past day's experience, this would be a good thing. OTOH - when interrupts work and connections can be terminated forcibly if necessary, the need for such timeouts diminish - even in the case of man interactive FTP client. There is litte a ^C cannot fix via properly coded signal handlers.
Which will not be part of my first cut 200 line FTP client, maybe later :-)

—Mellvik

Mellvik:
@ghaerr, `urlget` is ready for merge. 
The `ftpput` function needs more testing (in error handling), but we don't get to do that till the `ktcp` throttling problems have been resolved.
I'll keep that on my list.

-M

ghaerr:
Things look good, except for the last "minor fix". What exactly is that fixing, I don't understand it?

Mellvik:
The previous fix to clean the path from the prog variable was - well - ineffective,
this one does it.

—M

ghaerr:
> The previous fix to clean the path from the prog variable was - well - ineffective

```
   if ((prog = strrchr(*argv, '/')))
 		prog++;
    else prog = *argv;
```
For some reason, I can't figure out why your previous fix above doesn't work!
```
   prog = strrchr(*argv, '/');
   if (prog == (char *)NULL)
        prog = *argv;
```
I can see why the original code doesn't.

Mellvik:
That makes it two of us.

And instead of investigating, I took the easy way out :-).


—M

ghaerr:
> That makes it two of us.

Good. The code seems proper.


> And instead of investigating,

What exactly were you seeing that didn't work? That is, what was the passed value of *argv that failed?

>  I took the easy way out :-).

Well, I've learned the hard way that uncaught "bugs" always get you next time, so I'd like to figure this out!

ghaerr:
Out of curiosity, I tested your previous fix:
```
   if ((prog = strrchr(*argv, '/')))
    prog++;
   else
    prog = *argv;
printf("ARG '%s'\n", prog);
```
and got these results, indicating that it works perfectly:
```
# cd /
# urlget
ARG 'urlget'
Usage: urlget [-h] [-p] url
e.g. urlget http://216.58.209.67/index.html
# bin/urlget
ARG 'urlget'
Usage: urlget [-h] [-p] url
e.g. urlget http://216.58.209.67/index.html
# /bin/urlget
ARG 'urlget'
Usage: urlget [-h] [-p] url
e.g. urlget http://216.58.209.67/index.html
# 
```
I suggest reverting to that code, since it is so much nicer than the hack `if (prog == (char *)1)`. If it does not work on your system, perhaps we need to look into the possible toolchain issue on your system, which was noted during a discussion of having received no warnings on the same code compiled on our respective systems. Can you please retest with the original fix?

Mellvik:
Thank you  @ghaerr - you beat me to it (again). WIll repeat the test - seems I may have rushed this one.

Now that we have a solution to the ktcp outbound window problem, I get to test ftpput under stressed conditions too, so there may be other adjustments to urlget before it's ready for merginh

--Mellvik

ghaerr:
Thanks for all the testing and enhancements to `urlget` @Mellvik!

PR #981, opened 2021-10-23
[kernel] Add getsockopt system call and SO_LINGER for RST on close

ghaerr:
Implements ability to send TCP RST instead of TCP FIN on socket close, particularly on forced application exit, as discussed in #976 and #980.

The SO_LINGER option only implements l_linger == 0 when l_on_off == 1, in which case a RST is sent instead of FIN on socket close. The option can be used on any socket, and changed anytime before `close` or application exit.

Adds capability to `urlget`, but untested. Needs testing with `urlget -d`. The changes add two library calls (in net.c), `net_close_error`, which when called will send RST on close (which is also the default when a ^C signal kills the process), as well as `net_close_noerror`, which will allow for a normal close and associated TCP FIN cleanup when no file transfer errors have occurred. The `net_connect` call will default to sending RST on close, to handle the case of ^C interrupts while transfer is in progress.

Adds tested capability to `telnet`, tested but OFF by default, as unneeded.

@Mellvik, if this works, then the purge option can be deleted, as the TCP stack will send RST on close to shutdown the remote TCP packet sending. We may want to also turn on SEND_RST_ON_REFUSED_PKT as well, depending on what the packet trace shows. This may not be required, depending on the remote TCP stack. We will have to see. In either case, no data will be passed upwards to the application, and we should comment out the purge code and call close first with RST, to test that the low-level TCP handles the excess data instead (by dropping it). This will allow the user to operate without concern after ^C and no delays.

### DIFF
--- a/elks/arch/i86/kernel/strace.h
+++ b/elks/arch/i86/kernel/strace.h
@@ -248,6 +248,7 @@ struct syscall_info elks_table[] = {
     { "listen",		packinfo(2, P_SSHORT, P_SSHORT,  P_NONE   ) },
     { "accept",		packinfo(3, P_SSHORT, P_DATA,    P_PSSHORT) },
     { "connect",	packinfo(3, P_SSHORT, P_PDATA,   P_SSHORT ) },
+    { "setsockopt",	packinfo(5, P_SSHORT, P_SSHORT,  P_SSHORT, P_PDATA,   P_SSHORT ) },
 };
 
 #endif
--- a/elks/arch/i86/kernel/syscall.dat
+++ b/elks/arch/i86/kernel/syscall.dat
@@ -99,6 +99,7 @@ bind		+200	3	= CONFIG_SOCKET
 listen		+201	2	= CONFIG_SOCKET
 accept		+202	3	= CONFIG_SOCKET
 connect		+203	3	= CONFIG_SOCKET
+setsockopt	+204	5	= CONFIG_SOCKET
 #
 # Name			No	Args	Flag&comment
 #
--- a/elks/include/linuxmt/net.h
+++ b/elks/include/linuxmt/net.h
@@ -71,6 +71,7 @@ struct proto_ops {
 #define SO_ACCEPTCON	(1 << 1)
 #define SO_WAITDATA	(1 << 2)
 #define SO_NOSPACE	(1 << 3)
+#define SO_RST_ON_CLOSE	(1 << 4)
 
 struct net_proto {
     char *name;			/* Protocol name */
--- a/elks/include/linuxmt/socket.h
+++ b/elks/include/linuxmt/socket.h
@@ -12,6 +12,17 @@ struct sockaddr {
     char sa_data [MAX_SOCK_ADDR];
 };
 
+/* for setsockopt(2) */
+#define SOL_SOCKET	1
+
+/* careful: option name scheme interferes with internal SO_ options in net.h*/
+#define SO_LINGER	13		/* only implemented for l_linger = 0*/
+
+struct linger {
+        int             l_onoff;        /* Linger active                */
+        int             l_linger;       /* How long to linger for       */
+};
+
 struct msghdr {
     void *		msg_name;
     int 		msg_namelen;
--- a/elks/include/linuxmt/tcpdev.h
+++ b/elks/include/linuxmt/tcpdev.h
@@ -29,6 +29,7 @@
 struct tdb_release {
     unsigned char cmd;
     struct socket *sock;
+	int reset;
 };
 
 struct tdb_accept {
--- a/elks/net/ipv4/af_inet.c
+++ b/elks/net/ipv4/af_inet.c
@@ -95,11 +95,12 @@ static int inet_release(struct socket *sock, struct socket *peer)
     int ret;
 
     debug_net("INET(%d) release sock %x\n", current->pid, sock);
-	if (!tcpdev_inuse)
-		return -EINVAL;
+    if (!tcpdev_inuse)
+	return -EINVAL;
     cmd = (struct tdb_release *)get_tdout_buf();
     cmd->cmd = TDC_RELEASE;
     cmd->sock = sock;
+    cmd->reset = sock->flags & SO_RST_ON_CLOSE;
     ret = tcpdev_inetwrite(cmd, sizeof(struct tdb_release));
     return (ret >= 0 ? 0 : ret);
 }
--- a/elks/net/socket.c
+++ b/elks/net/socket.c
@@ -588,7 +588,6 @@ int sys_setsockopt(int fd, int level, int option_name, void *option_value,
     if (l.l_onoff != 0 && l.l_linger == 0)
 	sock->flags |= SO_RST_ON_CLOSE;
     else sock->flags &= ~SO_RST_ON_CLOSE;
-    printk("setsockopt reset %x\n", sock->flags);
 
     return 0;
 }
--- a/elks/net/socket.c
+++ b/elks/net/socket.c
@@ -567,4 +567,29 @@ int sys_socket(int family, int type, int protocol)
     return fd;
 }
 
+int sys_setsockopt(int fd, int level, int option_name, void *option_value,
+	unsigned int option_len)
+{
+    register struct socket *sock;
+    int err;
+    struct linger l;
+
+    if (!(sock = sockfd_lookup(fd, NULL)))
+	return -ENOTSOCK;
+
+    if (option_name != SO_LINGER || option_len != sizeof(struct linger))
+	return -EINVAL;
+
+    err = check_addr_to_kernel(option_value, sizeof(struct linger));
+    if (err < 0)
+	return err;
+    memcpy_fromfs(&l, option_value, sizeof(struct linger));
+
+    if (l.l_onoff != 0 && l.l_linger == 0)
+	sock->flags |= SO_RST_ON_CLOSE;
+    else sock->flags &= ~SO_RST_ON_CLOSE;
+    printk("setsockopt reset %x\n", sock->flags);
+
+    return 0;
+}
 #endif /* CONFIG_SOCKET */
--- a/elkscmd/inet/telnet/ttn.c
+++ b/elkscmd/inet/telnet/ttn.c
@@ -156,9 +156,8 @@ int main(int argc, char *argv[])
 	struct linger l;
 
 	l.l_onoff = 1;
-	l.l_linger = 0;
-	ret = setsockopt(tcp_fd, SOL_SOCKET, SO_LINGER, &l, sizeof(l));
-	printf("setsockopt returns %d\n", ret);
+	l.l_linger = 0;		/* send RST on close */
+	setsockopt(tcp_fd, SOL_SOCKET, SO_LINGER, &l, sizeof(l));
 }
 #endif
 	ret = bind(tcp_fd, (struct sockaddr *)&locadr, sizeof(struct sockaddr_in));
--- a/elkscmd/inet/telnet/ttn.c
+++ b/elkscmd/inet/telnet/ttn.c
@@ -151,6 +151,16 @@ int main(int argc, char *argv[])
 	locadr.sin_port = PORT_ANY;
 	locadr.sin_addr.s_addr = INADDR_ANY;
 
+#if 0	/* test code to force RST on telnet exit*/
+{
+	struct linger l;
+
+	l.l_onoff = 1;
+	l.l_linger = 0;
+	ret = setsockopt(tcp_fd, SOL_SOCKET, SO_LINGER, &l, sizeof(l));
+	printf("setsockopt returns %d\n", ret);
+}
+#endif
 	ret = bind(tcp_fd, (struct sockaddr *)&locadr, sizeof(struct sockaddr_in));
 	if (ret < 0){
 		perror("Bind failed");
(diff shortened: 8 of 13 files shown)


Mellvik:
Thank you @ghaerr - great work!

I'm looking forward to see this in action - testing this as we speak. And while in the process - a couple of questions:

- Is it the idea that the net_close_xxx routines are permanent, or temporary till we have some setsockopt call ?
- Is the 'noerror' variant really necessary? Wouldn't a normal close be sufficient?

Test report coming separately.

—Mellvik

Mellvik:
The results are in. Indeed we're getting there!
- The `net_close_error` seems to work precisely as intended.
- The `net_close_noerror` does not, causing a `socket operation on non-socket` error
- The `SEND_RST_ON_REFUSED_PKT` does not seem to have any effect.
- With `urlget`, the `-d` makes no difference.
- When a transfer id run to normal completion, the server sends `FIN`, ELKS responds with `RST`, see packet trace at the bottom.

--Mellvik

```
# urlget ftp://helge:xxxxxxxx@10.0.2.2/pcplus.img > /dev/null                  
setsockopt reset 10                                                             
setsockopt reset 10                                                             
^C                                                                             
# tcp: refusing packet 10.0.2.2:47505->1025                  <---- final packet from server                   
                                                                                
# urlget ftp://helge:xxxxxxxx@10.0.2.2/sh > sh                  <---- this one ran to completion
setsockopt reset 10                                                             
setsockopt reset 10                                                             
setsockopt reset 1                                                              
Setsockopt send FIN on close failed: tcp: refusing packet 10.0.2.2:21->1026     <---- noise from previous aborted connection
tcp: refusing packet 10.0.2.2:21->1026                                          
tcp: refusing packet 10.0.2.2:21->1026                                          
Socket operation on non-socket                                                  
# urlget ftp://helge:xxxxxxxx@10.0.2.2/pcplus.img > /dev/null                  
setsockopt reset 10                                                             
setsockopt reset 10                                                             
^C                                                                 <---- repeat of the first test, check repeatability
# tcp: refusing packet 10.0.2.2:42131->1029                                     <---- More noise
# urlget ftp://helge:xxxxxxxxx@10.0.2.2/sh > /dev/null                      <---- runs to completion   
setsockopt reset 10                                                             
setsockopt reset 10                                                             
setsockopt reset 1                                                              
Setsockopt send FIN on close failed: tcp: refusing packet 10.0.2.2:21->1030     
tcp: refusing packet 10.0.2.2:21->1030                                          
tcp: refusing packet 10.0.2.2:21->1030                                          
Socket operation on non-socket                                                  
# 
```
Packet trace from 'normal' termination:
```
12:55:49.016417 IP (tos 0x8, ttl 64, id 47195, offset 0, flags [DF], proto TCP (6), length 295)
    10.0.2.2.55531 > 10.0.2.15.1033: Flags [P.], cksum 0x192a (incorrect -> 0x73ea), seq 50236:50491, ack 1, win 29200, length 255
12:55:49.023183 IP (tos 0x0, ttl 64, id 2752, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.1033 > 10.0.2.2.55531: Flags [.], cksum 0xe828 (correct), ack 50491, win 255, length 0
12:55:49.023311 IP (tos 0x8, ttl 64, id 47196, offset 0, flags [DF], proto TCP (6), length 295)
    10.0.2.2.55531 > 10.0.2.15.1033: Flags [P.], cksum 0x192a (incorrect -> 0x15e6), seq 50491:50746, ack 1, win 29200, length 255
12:55:49.030039 IP (tos 0x0, ttl 64, id 2753, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.1033 > 10.0.2.2.55531: Flags [.], cksum 0xe729 (correct), ack 50746, win 255, length 0
12:55:49.030116 IP (tos 0x8, ttl 64, id 47197, offset 0, flags [DF], proto TCP (6), length 255)
    10.0.2.2.55531 > 10.0.2.15.1033: Flags [FP.], cksum 0x1902 (incorrect -> 0xabcc), seq 50746:50961, ack 1, win 29200, length 215
12:55:49.036842 IP (tos 0x0, ttl 64, id 2754, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.1033 > 10.0.2.2.55531: Flags [.], cksum 0xe651 (correct), ack 50962, win 255, length 0
12:55:49.065865 IP (tos 0x0, ttl 64, id 2755, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.1033 > 10.0.2.2.55531: Flags [R], cksum 0xe65d (correct), seq 391540709, win 255, length 0
12:55:49.067922 IP (tos 0x0, ttl 64, id 2756, offset 0, flags [none], proto TCP (6), length 46)
    10.0.2.15.1032 > 10.0.2.2.ftp: Flags [P.], cksum 0x18b4 (correct), seq 52:58, ack 614, win 255, length 6: FTP, length: 6
	QUIT
12:55:49.068197 IP (tos 0x10, ttl 64, id 40793, offset 0, flags [DF], proto TCP (6), length 54)
    10.0.2.2.ftp > 10.0.2.15.1032: Flags [P.], cksum 0x1839 (incorrect -> 0x6070), seq 614:628, ack 58, win 29200, length 14: FTP, length: 14
	221 Goodbye.
12:55:49.070708 IP (tos 0x0, ttl 64, id 2757, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.1032 > 10.0.2.2.ftp: Flags [.], cksum 0xc061 (correct), ack 628, win 255, length 0
12:55:49.096256 IP (tos 0x10, ttl 64, id 40794, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.ftp > 10.0.2.15.1032: Flags [F.], cksum 0x182b (incorrect -> 0x4f4f), seq 628, ack 58, win 29200, length 0
12:55:49.113165 IP (tos 0x0, ttl 64, id 2758, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.1032 > 10.0.2.2.ftp: Flags [.], cksum 0xc060 (correct), ack 629, win 255, length 0
12:55:49.119454 IP (tos 0x0, ttl 64, id 2759, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.1032 > 10.0.2.2.ftp: Flags [R], cksum 0xc06c (correct), seq 391540757, win 255, length 0
```

ghaerr:
Hello @Mellvik,

Thanks for your testing. It seems that `urlget` is always sending an RST. Do you have SEND_RST_ON_CLOSE set to 1 in config.h? That needs to be 0 (OFF) for this test, as that overrides `setsocketopt` in `net_close_noerror`.
The "Socket operation on non-socket" is the result of a system call returning -ENOTSOCK, which I think means that the connection was closed (possibly as a result of a previous RST sent on another channel?) prior to the call.

I will answer the other questions but wanted to make sure the SEND_RST_ON_CLOSE is 0 for this testing.

Thank you!

ghaerr:
> Is it the idea that the net_close_xxx routines are permanent, or temporary till we have some setsockopt call ?

The net_close routines were added to match the net_connect call which urlget always uses. I'm going to push a simpler fix which uses a second parameter to indicate error, rather than a seperate call. This PR implements the required setsockopt call, which sets or unsets the linger timeout, which is used to turn on or off sending a RST on close. A close call will always need to take place.

The idea was that any file descriptor returned from net_connect should call net_close_xxx.

> Is the 'noerror' variant really necessary? Wouldn't a normal close be sufficient?

Actually, its the reverse: since net_connect turns on "send RST on close", the noerror version needs to turn that off before calling close.

> The net_close_error seems to work precisely as intended.

This just calls close without resetting "send RST on close", so a RST is sent.

> The net_close_noerror does not, causing a socket operation on non-socket error

I'm pretty sure that SEND_RST_ON_CLOSE is compiled in to ktcp, so the channel is being shutdown prior to the setsockopt call, which returns -ENONSOCK, and the code reports an error. We may want to remove that error display for the case where a shutdown has previously occurred.


> The SEND_RST_ON_REFUSED_PKT does not seem to have any effect.

I have never actually tested this - can you look at the packet trace to see whether the src and dest inet addr and ports are correct? If the connection is previously shutdown, this wouldn't have an effect here, but we should leave it in anyways. 


> With urlget, the -d makes no difference.

That's good news - the purge is no longer needed.

> When a transfer id run to normal completion, the server sends FIN, ELKS responds with RST, see packet trace at the bottom.

I think this is the result of SEND_RST_ON_CLOSE being compiled in.

Mellvik:
Thank you @ghaerr -
I missed that one. SEND_RST_ON_CLOSE was set during this testing, so I'll have to redo that.

Stay tuned.

-M

Mellvik:
@ghaerr,
I've just completed some testing based on the last committed tree. What I get is the following:
- Interrupted transfer to `/dev/null`: Perfect
- Completed transfer to `/dev/null`: Fine, except for the message `Setsockopt send FIN on close failed: Socket operation on non-socket` at completion
- Transfer to local file: This is odd - no network activity, but the following (at least 40 of them, all different values):
```
# urlget ftp://helge:xxxxxxxxxx@10.0.2.2/sh > sh
mfb (0380:60191): trying to free block not in datazone                          
mfb (0380:22233): trying to free block not in datazone                          
mfb (0380:21847): trying to free block not in datazone                          
mfb (0380:58761): trying to free block not in datazone                          
mfb (0380:60547): trying to free block not in datazone                          
mfb (0380:35590): trying to free block not in datazone 
...  
```
... and the terminal hangs. Doing a `ps` from a different terminal shows the `urlget` process gone. Ktcp still works. Trying the same command again gives no messages, just a hung terminal : `urlget ftp://helge:xxxxxxxxx@10.0.2.2/sh > sh` 

Rebooting and repeating w/o the first `urlget` to `dev/null` gives a slightly different result - `urlget` still hangs with no network activity, but there are no mfb-messages.

Did I miss something here?

--M

ghaerr:
> * Interrupted transfer to `/dev/null`: Perfect

Fixed!

> * Completed transfer to `/dev/null`: Fine, except for the message `Setsockopt send FIN on close failed: Socket operation on non-socket` at completion

The remnant message is likely the result of the close of the 2nd file descriptor for FTP transfers... the "socket operation on non-socket" means the file descriptor is already closed. I'll remove the message.



> * Transfer to local file: This is odd - no network activity, but the following (at least 40 of them, all different values):
> 
> ```
> # urlget ftp://helge:xxxxxxxxxx@10.0.2.2/sh > sh
> mfb (0380:60191): trying to free block not in datazone  

This is a corrupted MINIX filesystem. Not sure exactly why yet, but I would guess something to do with possibly application forced or unforced termination and kernel filesystem left in unclean state. For now, we should test only with newly mkfs'd floppies, if possible. We can debug MINIX fs issues separately.


> Rebooting and repeating w/o the first `urlget` to `dev/null` gives a slightly different result - `urlget` still hangs with no network activity, but there are no mfb-messages.

Guessing that filesystem is corrupted.

We still need a test on urlget error when the MINIX floppy is almost full, as that triggers separate error code in urlget.

Thanks for your testing!

ghaerr:
> except for the message Setsockopt send FIN on close failed: Socket operation on non-socket at completion

If you are getting that message, it means you're also using old code - the latest repo message is different "setsockopt FIN". The latest repo also has the change to urlget to properly close the FTP file descriptors in the correct order.

Mellvik:
Wow @ghaerr,
I need to backtrace the process and find out what happened. 
Stay tuned.

-M

Mellvik:
OK, mea culpa - wrong floppy :-) Sorry about that, @ghaerr.

Updated test results (good news all over):

- Interrupted file transfer: OK - the RSTs kill the connection immediately, no leftovers.
- Successful transfers: All good.
- File system full (file write error): `urlget` hangs in `ftpreply`, waiting for reply to the ABORt command. Commenting out the actual `ftpreply`-call fixes the problem and error termination works as expected. That said, since we now forcibly close the data channel on error, the ABORt command has no meaning. The channel is gone before the server process can do anything about it. So I deleted both the ABORt cmd and the repy-wait in `urlget`. Further, and not very surprising, it turns out that when the Ftp data channel has been forcibly terminated by an RST, the command channel can be closed in a normal manner (FIN), which is cleaner. This is application level stuff though. I will PR these minor changes to `urlget` along with a couple of other fixes soon.

IOW - `ktcp` is now well behaving in all these situation. Well done (again) @ghaerr, thank you! I think it's time to let the `refusing packet` messages go.
BTW - and while we're at it, any chance we can up the receive packet size to 512, like the send packets? That would likely double the transfer speed...

--Mellvik

ghaerr:
Ok, good news @Mellvik.

> So I deleted both the ABORt cmd and the repy-wait in urlget. Further, and not very surprising, it turns out that when the Ftp data channel has been forcibly terminated by an RST, the command channel can be closed in a normal manner (FIN), which is cleaner. This is application level stuff though. I will PR these minor changes to urlget along with a couple of other fixes soon.

Ok, go ahead and close #980 and submit a PR with the ABORT ftpsend removed, as well as the purge and '-d' option removed. It appears that urlget will be much closer to where it started (i.e. net_close fixes only), now that we have TCP working properly.

> I think it's time to let the refusing packet messages go.

I need to keep them to test sending RST on refused packets in #985.

> any chance we can up the receive packet size to 512, like the send packets? That would likely double the transfer speed...

I'll look into that, I seem to remember it was a bit complicated but I'll try!

PR #982, opened 2021-10-23
ssd: add generic SD card ssd driver.

cocus:
8018x: add bit-banged low level SD card driver for ssd.

Adds support for generic SD cards as ssd block devices.
Adds a low-level SPI bit-banged driver that allows the usage
of the generic SD card driver on the 8018x platform. GPIO
lines are configurable on the source code to some degree.
Confirmed to work with a 2GB SanDisk SD card, formatted
with fat16.


ghaerr:
Hello @cocus,

Looks great, thanks for doing such a great job on this :)

Please push two fixes and I'll commit this:
- In ll_rw_blk.c, use "\#if defined(CONFIG_BLK_DEV_SSD) || defined(CONFIG_BLK_DEV_TEST)"
- In image/Make.devices, lines with a \# are actually comments. You'll probably have to use "`ifneq ($(CONFIG_DEV_BLK_NONE), y)`, followed by `endif`

There are some other areas where we might want to use segment and offset rather than a __far pointer, to match the way almost all other 8086 addresses are used in the kernel, but we'll deal with those another day.

Thank you!

cocus:
Done deal!

> * In ll_rw_blk.c, use "#if defined(CONFIG_BLK_DEV_SSD) || defined(CONFIG_BLK_DEV_TEST)"

I ended up doing a `#if defined(CONFIG_BLK_DEV_TEST) || defined(CONFIG_BLK_DEV_SSD_SD8018X)` since `CONFIG_BLK_DEV_SSD` doesn't exist anymore.

toncho11:
Hi @cocus

What is your hardware indeed? What motherboard or embedded platform are you using? Can you please show a photo of your hardware?

cocus:
@toncho11
Hi! I'm using an old HDSL to E1 (or T1, no idea) rack-mounted modem from Orckit. It's not an SBC nor a development kit, but... here we are :)
I don't have many pictures of the original board, but I've posted one on my blog at https://cocus.rf.gd/2022/01/10/an-80x86-adventure-part-1-of-x/.
There's another picture I've shared on this repo on a "issue" before, but I don't recall in which one.
In any case, this device didn't have SD (nor ISA) native support, I added it by hacking some wires onto specific GPIO pins of the 80C188EB CPU that nothing else is using.
In fact, it originally had a 128kB RAM which I promptly upgraded to 512kB since ELKS wasn't happy about the small RAM size :).
Needless to say that everything I've done is a "hack" on this poor board.

Issue #986, opened 2021-10-28
KTCP ignoring peer's closing window


Mellvik:
Summary:
When sending data, `ktcp` seems to (at least partly) ignore a peer's closing window.

Here's what seems to be going on:
- Elks is happily sending data to the recipient (ftpput), 512b payload per packet, receiving ACKs for every 2 or 3 packet.
- The recipient is a raspberrypi, very fast compared to the ELKS machine (386/20), with a floppy disk (USB) for the occasion, to slow down the transfer and cause write errors.
- When the recipient suddenly needs ELKS to slow down, it reduces its window to 512 (from 2048 or larger in the preceding packets).
- Elks seemingly reponds to this by reducing the next packet payload to 360.
- At this point - until the recipient has told ELKS otherwise, the actual window is 512-360=152b.
- But ELKS doesn't wait. Immediately after the 360 size packet, comes another series of 512 byte packets, which the recipient is now unable to process.
- At this point, after 4 512-byte packets, the recipient desperately sends an empty ACK back to ELKS advertizing a ZERO window and repeating the previous ACK, not acknowledging any of the last 4 packets.

There are other observations to be made from the following packets, but it's probably not interesting if the problem is that `ktcp` fails to recognize that the recipient window has closed.

Any suggestions, @ghaerr?

--Mellvik

ghaerr:
I seem to remember that `ktcp` may not stop sending packets based on the received window size at all, which sounds like what is happening. I'll look at the code and respond with what `ktcp` actually is coded for and respond. 

In summary, ELKS TCP is now receiving quite well, and with some minor tweaks can improve that speed with some possibly dynamically specified broadcast window size, but now you've found that sending data from ELKS is problematic, as described above, correct?

In this non-stop sending, do the systems ever return to sync, or does the outgoing transfer finally fail?

Mellvik:
@ghaerr,
Now that you mention it - I do recall we've touched that one before.
Maybe this is a good time to revisit, now that the receving is under control.

> In this non-stop sending, do the systems ever return to sync, or does the outgoing transfer finally fail?

Eventually ktcp realizes the missing ACKs and starts repeating the last packet. And eventually the recipient reopens the window, without ACKing any of the un-ack'ed packets. KTCP keeps resending the last packet, the recipient closes the window again and so on. Finally ktcp just goes quiet and hangs hard. See the netstat from the previous post. 
Ktcp is still operational though, telnet in and out works.

More details on this tomorrow.

--M

[...]

PR #987, opened 2021-10-29
[ktcp] Fine tune TCP send and receive window sizes

ghaerr:
Allows for larger (current max 512) receive window to increase file transfer speed, tested as fixed in https://github.com/jbruchon/elks/issues/976#issuecomment-953244002.

Should fix server receive window ignored when sending packets (#986).

@Mellvik, as you can see, the send packet limiting code now compares against the receive window size, or 1024 (TCP_SEND_WINDOW_MAX), whichever is smaller. A message "tcp write: limiting write at max window ..." will temporarily be displayed when/if ktcp stops sending for 10ms to give time for the server to process the previous packets. If you uchange line 341 in ktcp/tcpdev.c to `printf` it will display the outstanding window and unacked send size should more information be needed to trigger the limiting.

If this message appears too frequently, the ktcp write packet rescheduling timer can be increased from 10ms by adjusting the following code in elks/net/ipv4/af_inet.c in inet_write():
```
        
        if (ret < 0) {
            if (ret == -ERESTARTSYS) {
                /* delay process 10ms*/
                current->state = TASK_INTERRUPTIBLE;
                current->timeout = jiffies + 10;  // <-- change from 10 to a higher number to wait longer
                schedule();
            } else
```
I can only test on QEMU at the moment, and this fix doesn't change any telnet/localhost behavior.

### DIFF
--- a/elks/net/ipv4/af_inet.c
+++ b/elks/net/ipv4/af_inet.c
@@ -196,13 +196,8 @@ static int inet_listen(register struct socket *sock, int backlog)
     tcpdev_inetwrite(cmd, sizeof(struct tdb_listen));
 
     /* Sleep until tcpdev has news */
-    while (bufin_sem == 0) {
+    while (bufin_sem == 0)
         interruptible_sleep_on(sock->wait);
-        if (current->signal) {
-printk("inet_listen: RESTARTSYS\n");
-            return -ERESTARTSYS;
-        }
-    }
 
     ret = ((struct tdb_return_data *)tdin_buf)->ret_value;
     tcpdev_clear_data_avail();
@@ -231,7 +226,7 @@ static int inet_accept(register struct socket *sock,
         interruptible_sleep_on(sock->wait);
         //sock->flags &= ~SO_WAITDATA;
         if (current->signal) {
-printk("INET(%d) accept RESTARTSYS\n", current->pid);
+	    debug_net("INET(%d) accept RESTARTSYS\n", current->pid);
             return -ERESTARTSYS;
 	}
     }
--- a/elks/net/ipv4/af_inet.c
+++ b/elks/net/ipv4/af_inet.c
@@ -351,7 +351,7 @@ static int inet_write(register struct socket *sock, char *ubuf, int size,
 
 	if (ret < 0) {
             if (ret == -ERESTARTSYS) {
-		/* delay process 10ms*/
+		/* delay process 100ms*/
 		current->state = TASK_INTERRUPTIBLE;
 		current->timeout = jiffies + 10;
                 schedule();
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -8,6 +8,7 @@
 #define USE_DEBUG_EVENT 1	/* use CTRLP to toggle debug output*/
 #define DEBUG_STARTDEF	0	/* default startup debug display*/
 #define DEBUG_TCP	1
+#define DEBUG_RETRANS	1
 #define DEBUG_CLOSE	1
 #define DEBUG_IP	0
 #define DEBUG_ARP	0
@@ -31,6 +32,12 @@ void dprintf(const char *, ...);
 #define debug_tcp	DPRINTF
 #else
 #define debug_tcp(...)
+#endif
+
+#if DEBUG_RETRANS
+#define debug_retrans	DPRINTF
+#else
+#define debug_retrans(...)
 #endif
 
 #if DEBUG_CLOSE
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -2,7 +2,7 @@
 #define CONFIG_H
 
 /* compile time options*/
-#define CSLIP			1	/* compile in CSLIP support*/
+#define CSLIP		0	/* compile in CSLIP support*/
 
 /* turn these on for ELKS debugging*/
 #define USE_DEBUG_EVENT 1	/* use CTRLP to toggle debug output*/
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -7,9 +7,10 @@
 /* turn these on for ELKS debugging*/
 #define USE_DEBUG_EVENT 1	/* use CTRLP to toggle debug output*/
 #define DEBUG_STARTDEF	0	/* default startup debug display*/
-#define DEBUG_TCP	1
-#define DEBUG_RETRANS	1
-#define DEBUG_CLOSE	1
+#define DEBUG_TCP	1	/* TCP ops*/
+#define DEBUG_TCPDATA	0	/* TCP data packets*/
+#define DEBUG_RETRANS	1	/* TCP retransmissions*/
+#define DEBUG_CLOSE	1	/* TCP close ops*/
 #define DEBUG_IP	0
 #define DEBUG_ARP	0
 #define DEBUG_ETH	0
@@ -32,6 +33,12 @@ void dprintf(const char *, ...);
 #define debug_tcp	DPRINTF
 #else
 #define debug_tcp(...)
+#endif
+
+#if DEBUG_TCPDATA
+#define debug_tcpdata	DPRINTF
+#else
+#define debug_tcpdata(...)
 #endif
 
 #if DEBUG_RETRANS
--- a/elkscmd/ktcp/tcp.c
+++ b/elkscmd/ktcp/tcp.c
@@ -29,25 +29,43 @@ timeq_t Now;
 extern int cbs_in_time_wait;		// FIXME remove extern
 extern int cbs_in_user_timeout;
 
+#if DEBUG_TCPDATA
+static char *tcp_flags(int flags)
+{
+    static char buf[7];
+    char *p = buf;
+
+    if (flags & TF_FIN) *p++ = 'F';
+    if (flags & TF_SYN) *p++ = 'S';
+    if (flags & TF_RST) *p++ = 'R';
+    if (flags & TF_PSH) *p++ = 'P';
+    if (flags & TF_URG) *p++ = 'U';
+    if (flags & TF_ACK) *p++ = 'A';
+    *p = 0;
+    return buf;
+}
+#endif
+
 void tcp_print(struct iptcp_s *head, int recv, struct tcpcb_s *cb)
 {
-#if DEBUG_TCP
-    debug_tcp("TCP: %s ", recv? "recv": "send");
-    debug_tcp("src:%u dst:%u ", ntohs(head->tcph->sport), ntohs(head->tcph->dport));
-    debug_tcp("flags:0x%x ",head->tcph->flags);
+#if DEBUG_TCPDATA
+    debug_tcpdata("tcp: %s ", recv? "recv": "send");
+    debug_tcpdata("%u->%u ", ntohs(head->tcph->sport), ntohs(head->tcph->dport));
+    debug_tcpdata("[%s] ", tcp_flags(head->tcph->flags));
     if (cb) {
       if (recv) {
-    	if (head->tcplen > 20) debug_tcp("seq:%ld-%ld ", ntohl(head->tcph->seqnum) - cb->irs, 
+	if (head->tcplen > 20) debug_tcp("seq %lu-%lu ", ntohl(head->tcph->seqnum) - cb->irs,
 				ntohl(head->tcph->seqnum) - cb->irs+head->tcplen-20);
-	debug_tcp("ack:%ld ", ntohl(head->tcph->acknum) - cb->iss);
+	debug_tcpdata("ack %lu ", ntohl(head->tcph->acknum) - cb->iss);
       } else {
-    	if (head->tcplen > 20) debug_tcp("seq:%ld-%ld ", ntohl(head->tcph->seqnum) - cb->iss, 
+	if (head->tcplen > 20) debug_tcpdata("seq %lu-%lu ",
+	ntohl(head->tcph->seqnum) - cb->iss,
 	ntohl(head->tcph->seqnum) - cb->iss+head->tcplen-20);
- 	debug_tcp("ack:%ld ", ntohl(head->tcph->acknum) - cb->irs);
+	debug_tcpdata("ack %lu ", ntohl(head->tcph->acknum) - cb->irs);
       }
     }
-    debug_tcp("win:%u urg:%d ", ntohs(head->tcph->window), head->tcph->urgpnt);
-    debug_tcp("chk:0x%x len:%u\n", tcp_chksum(head), head->tcplen);
+    debug_tcpdata("win %u urg %d ", ntohs(head->tcph->window), head->tcph->urgpnt);
+    debug_tcpdata("chk %x len %u\n", tcp_chksum(head), head->tcplen);
 #endif
 }
 
@@ -141,6 +159,7 @@ printf("SYN sent, wrong ACK (listen port not expired)\n");
 	cb->send_una++;
 	cb->state = TS_ESTABLISHED;
 	cb->flags = TF_ACK;
+	debug_tcp("TS_ESTABLISHED\n");
 
 	cb->datalen = 0;
 	tcp_output(cb);
@@ -247,7 +266,7 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 
     if (h->flags & TF_RST) {
 	/* TODO: Check seqnum for security */
-	printf("tcp: RST from %s:%u->%d\n", in_ntoa(cb->remaddr), ntohs(h->sport), ntohs(h->dport));
+	printf("tcp: RST from %s:%u->%u\n", in_ntoa(cb->remaddr), ntohs(h->sport), ntohs(h->dport));
 	rmv_all_retrans_cb(cb);
 	if (cb->state == TS_CLOSE_WAIT) {
 	    ENTER_TIME_WAIT(cb);
@@ -285,6 +304,7 @@ static void tcp_synrecv(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 	debug_tcp("tcp: NO ACK IN SYNRECV\n");
     else {
 	cb->state = TS_ESTABLISHED;
+	debug_tcp("TS_ESTABLISHED\n");
 	tcpdev_checkaccept(cb);
 	tcp_established(iptcp, cb);
     }
@@ -390,7 +410,7 @@ static void tcp_last_ack(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 void tcp_update(void)
 {
     if (cbs_in_time_wait > 0 || cbs_in_user_timeout > 0) {
-debug_tcp("ktcp: update %x,%x\n", cbs_in_time_wait, cbs_in_user_timeout);
+	debug_tcp("tcp: time_wait %d user_timeout %d\n", cbs_in_time_wait, cbs_in_user_timeout);
 	tcpcb_expire_timeouts();
     }
 
@@ -484,7 +504,6 @@ void tcp_process(struct iphdr_s *iph)
 	    break;
 
 	case TS_ESTABLISHED:
-	    debug_tcp("TS_ESTABLISHED\n");
 	    tcp_established(&iptcp, cb);
 	    break;
 
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -45,7 +45,7 @@
 /* retransmit settings*/
 #define TCP_RTT_ALPHA			90
 #define TCP_RETRANS_MAXMEM		4096	/* max retransmit total memory*/
-#define TCP_RETRANS_MAXTRIES		3	/* max # retransmits*/
+#define TCP_RETRANS_MAXTRIES		6	/* max # retransmits (~12 secs total)*/
 /* timeout values in 1/16 seconds*/
 #define TCP_RETRANS_MAXWAIT		64	/* max retransmit wait (4 secs)*/
 #define TCP_RETRANS_MINWAIT_SLIP	8	/* minimum retrans timeout for slip/cslip (1/2 sec)*/
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -28,6 +28,9 @@
 /* max outstanding send window size*/
 #define TCP_SEND_WINDOW_MAX	1024	/* should be less than TCP_RETRANS_MAXMEM*/
 
+/* max advertised receive window size*/
+#define THROTTLE_MAX_WINDOW	512	/* FIXME CB_IN_BUF_SIZE when PTY fixed*/
+
 /* bytes to subtract from window size and when to force app write*/
 #define PUSH_THRESHOLD	512
 
--- a/elkscmd/ktcp/tcp_cb.c
+++ b/elkscmd/ktcp/tcp_cb.c
@@ -70,7 +70,7 @@ struct tcpcb_list_s *tcpcb_new(void)
 
     n = (struct tcpcb_list_s *) malloc(sizeof(struct tcpcb_list_s));
     if (n == NULL) {
-	debug_tcp("ktcp: Out of memory 2\n");
+	debug_tcp("ktcp: Out of memory for CB\n");
 	return NULL;
     }
     debug_mem("Alloc CB %d bytes\n", sizeof(struct tcpcb_list_s));
@@ -117,21 +117,19 @@ void tcpcb_remove(struct tcpcb_list_s *n)
     struct tcpcb_list_s *next = n->next;
 
     debug_tcp("tcp: REMOVING control block %x\n", n);
+    debug_mem("Free CB\n");
     tcpcb_num--;	/* for netstat*/
 
     if (n->prev)
 	n->prev->next = next;
     else {
-
 	/* Head update */
 	n = next;
 	n->prev = NULL;
 
 	rmv_all_retrans(tcpcbs);
-	debug_mem("Free CB\n");
 	free(tcpcbs);
 	tcpcbs = n;
-
 	return;
     }
 
@@ -140,7 +138,6 @@ void tcpcb_remove(struct tcpcb_list_s *n)
 
     rmv_all_retrans(n);
     free(n);
-    debug_mem("Free CB\n");
 }
 
 struct tcpcb_list_s *tcpcb_check_port(__u16 lport)
--- a/elkscmd/ktcp/tcp_cb.c
+++ b/elkscmd/ktcp/tcp_cb.c
@@ -229,7 +229,7 @@ void tcpcb_expire_timeouts(void)
 	    int secs = (unsigned)(n->tcpcb.time_wait_exp - Now);
 	    unsigned int tenthsecs = ((secs + 8) & 15) >> 1;
 	    secs >>= 4;
-	    debug_close("ktcp: check expire %s (%d.%d)\n",
+	    debug_close("tcp: expire check %s (%d.%d)\n",
 		tcp_states[n->tcpcb.state], secs, tenthsecs);
 	}
 #endif
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -176,14 +176,14 @@ rmv_from_retrans(struct tcp_retrans_list_s *n)
 
     tcp_timeruse--;
     tcp_retrans_memory -= n->len;
+    debug_mem("retrans free: (cnt %d mem %u)\n", tcp_timeruse, tcp_retrans_memory);
+
     if (n->prev)
 	n->prev->next = next;
     else {
-
 	/* Head update */
 	n = next;
 	n->prev = NULL;
-	debug_mem("retrans free: (cnt %d mem %u)\n", tcp_timeruse, tcp_retrans_memory);
 	free(retrans_list);
 	retrans_list = n;
 
@@ -192,8 +192,6 @@ rmv_from_retrans(struct tcp_retrans_list_s *n)
 
     if (next)
 	next->prev = n->prev;
-
-    debug_mem("retrans free: (cnt %d mem %u)\n", tcp_timeruse, tcp_retrans_memory);
     free(n);
 
     return next;
@@ -241,7 +239,7 @@ void add_for_retrans(struct tcpcb_s *cb, struct tcphdr_s *th, __u16 len,
 
     n = (struct tcp_retrans_list_s *)malloc(sizeof(struct tcp_retrans_list_s) + len);
     if (n == NULL) {
-	printf("ktcp: Out of memory\n");
+	printf("ktcp: Out of memory for retrans\n");
 	return;
     }
 
@@ -269,7 +267,7 @@ void add_for_retrans(struct tcpcb_s *cb, struct tcphdr_s *th, __u16 len,
     n->retrans_num = 0;
     n->first_trans = Now;
 
-    n->rto = cb->rtt << 1;
+    n->rto = cb->rtt << 1;			/* set retrans timeout to twice RTT*/
     if (linkprotocol == LINK_ETHER) {
 	if (n->rto < TCP_RETRANS_MINWAIT_ETH)
 	    n->rto = TCP_RETRANS_MINWAIT_ETH;	/* 1/4 sec min retrans timeout on ethernet*/
@@ -282,15 +280,20 @@ void add_for_retrans(struct tcpcb_s *cb, struct tcphdr_s *th, __u16 len,
 
 void tcp_reoutput(struct tcp_retrans_list_s *n)
 {
-    n->retrans_num++;
-    n->rto *= 2;
+    unsigned int datalen = n->len - TCP_DATAOFF(&n->tcphdr[0]);
+
+    if (datalen < n->cb->rcv_wnd)		/* don't record retry if not in recv window*/
+	n->retrans_num++;
+    n->rto <<= 1;				/* double retrans timeout*/
     if (n->rto > TCP_RETRANS_MAXWAIT)		/* limit retransmit timeouts to 4 seconds*/
 	n->rto = TCP_RETRANS_MAXWAIT;
     n->next_retrans = Now + n->rto;
-printf("tcp retrans: seq %lu size %d rcvwnd %u unack %lu rto %ld (retry %d cnt %d mem %u)\n",
-ntohl(n->tcphdr[0].seqnum), n->len - TCP_DATAOFF(&n->tcphdr[0]),
-n->cb->rcv_wnd, n->cb->send_nxt - n->cb->send_una,
-n->rto, n->retrans_num, tcp_timeruse, tcp_retrans_memory);
+
+    printf("tcp retrans: seq %lu size %d rcvwnd %u unack %lu rto %ld rtt %ld (retry %d cnt %d mem %u)\n",
+	ntohl(n->tcphdr[0].seqnum), n->len - TCP_DATAOFF(&n->tcphdr[0]),
+	n->cb->rcv_wnd, n->cb->send_nxt - n->cb->send_una,
+	n->rto, n->cb->rtt, n->retrans_num, tcp_timeruse, tcp_retrans_memory);
+
     ip_sendpacket((unsigned char *)n->tcphdr, n->len, &n->apair, n->cb);
     netstats.tcpretranscnt++;
 }
@@ -315,27 +318,31 @@ void tcp_retrans(void)
     while (n != NULL) {
 	datalen = n->len - TCP_DATAOFF(&n->tcphdr[0]);
 
-	//debug_mem("retrans seqno: %lx %lx", ntohl(n->tcphdr[0].seqnum) + datalen,n->cb->send_una);
+	/* calc RTT and remove if seqno was acked*/
 	if (SEQ_LEQ(ntohl(n->tcphdr[0].seqnum) + datalen, n->cb->send_una)) {
 	    if (n->retrans_num == 0) {
 		rtt = Now - n->first_trans;
 		if (rtt > 0)
 		    n->cb->rtt = (TCP_RTT_ALPHA * n->cb->rtt + (100 - TCP_RTT_ALPHA) * rtt) / 100;
-		debug_tcp("rtt %d RTT %ld RTO %ld\n", rtt, n->cb->rtt, n->rto);
+		debug_tcp("tcp: rtt %d RTT %ld RTO %ld\n", rtt, n->cb->rtt, n->rto);
 	    }
-	    debug_mem("retrans removed: %d seqno >= unacked\n", n->retrans_num);
+	    debug_retrans("retrans remove: seq %ld unack %ld\n",
+		ntohl(n->tcphdr[0].seqnum), n->cb->send_una);
 	    n = rmv_from_retrans(n);
 	    continue;
-	}
+	} else
+	    debug_retrans("retrans check: seq %ld unack %ld time %ld\n",
+		ntohl(n->tcphdr[0].seqnum) + datalen, n->cb->send_una, n->next_retrans - Now);
 
-	//debug_mem(" time check %lx %lx\n", Now, n->next_retrans);
-	/* check for retrans time up and receive window big enough*/
-	if (TIME_GEQ(Now, n->next_retrans) && datalen < n->cb->rcv_wnd) {
+	/* check for retrans time up*/
+	if (TIME_GEQ(Now, n->next_retrans)) {
 	    tcp_reoutput(n);
 	    if (n->retrans_num >= TCP_RETRANS_MAXTRIES) {
-		 debug_mem("retrans removed: %d max tries exceeded\n", n->retrans_num);
-		 n = rmv_from_retrans(n);
-		 continue;
+		printf("retrans max retries: seq %ld unack %ld time %ld\n",
+		    ntohl(n->tcphdr[0].seqnum), n->cb->send_una, n->next_retrans - Now);
+		tcp_send_reset(n->cb);		/* CB deallocated on received RST*/
+		n = rmv_from_retrans(n);
+		continue;
 	    }
 	}
 	n = n->next;
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -287,7 +287,10 @@ void tcp_reoutput(struct tcp_retrans_list_s *n)
     if (n->rto > TCP_RETRANS_MAXWAIT)		/* limit retransmit timeouts to 4 seconds*/
 	n->rto = TCP_RETRANS_MAXWAIT;
     n->next_retrans = Now + n->rto;
-printf("retrans retry: #%d rto %ld (cnt %d, mem %u)\n", n->retrans_num, n->rto, tcp_timeruse, tcp_retrans_memory);
+printf("tcp retrans: seq %lu size %d rcvwnd %u unack %lu rto %ld (retry %d cnt %d mem %u)\n",
+ntohl(n->tcphdr[0].seqnum), n->len - TCP_DATAOFF(&n->tcphdr[0]),
+n->cb->rcv_wnd, n->cb->send_nxt - n->cb->send_una,
+n->rto, n->retrans_num, tcp_timeruse, tcp_retrans_memory);
     ip_sendpacket((unsigned char *)n->tcphdr, n->len, &n->apair, n->cb);
     netstats.tcpretranscnt++;
 }
@@ -296,7 +299,8 @@ printf("retrans retry: #%d rto %ld (cnt %d, mem %u)\n", n->retrans_num, n->rto, 
 void tcp_retrans(void)
 {
     struct tcp_retrans_list_s *n;
-    int datalen, rtt;
+    int rtt;
+    unsigned int datalen;
 
     /* avoid running out of memory with excessive retransmits*/
     if (tcp_retrans_memory > TCP_RETRANS_MAXMEM) {
@@ -325,7 +329,8 @@ void tcp_retrans(void)
 	}
 
 	//debug_mem(" time check %lx %lx\n", Now, n->next_retrans);
-	if (TIME_GEQ(Now, n->next_retrans)) {
+	/* check for retrans time up and receive window big enough*/
+	if (TIME_GEQ(Now, n->next_retrans) && datalen < n->cb->rcv_wnd) {
 	    tcp_reoutput(n);
 	    if (n->retrans_num >= TCP_RETRANS_MAXTRIES) {
 		 debug_mem("retrans removed: %d max tries exceeded\n", n->retrans_num);
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -352,8 +352,9 @@ void tcp_output(struct tcpcb_s *cb)
 
     len = CB_BUF_SPACE(cb) - PUSH_THRESHOLD;
     if (len <= 0)
-	len = 1;		/* Never advertise zero window size */
-    else len = 255;	//FIXME testing only
+	len = 1;			/* Never advertise zero window size */
+    if (len > THROTTLE_MAX_WINDOW)	/* throttle down to allow ELKS apps to keep up */
+	len = THROTTLE_MAX_WINDOW;
     th->window = htons(len);
     th->urgpnt = 0;
     th->flags = cb->flags;
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -289,8 +289,8 @@ void tcp_reoutput(struct tcp_retrans_list_s *n)
 	n->rto = TCP_RETRANS_MAXWAIT;
     n->next_retrans = Now + n->rto;
 
-    printf("tcp retrans: seq %lu size %d rcvwnd %u unack %lu rto %ld rtt %ld (retry %d cnt %d mem %u)\n",
-	ntohl(n->tcphdr[0].seqnum), n->len - TCP_DATAOFF(&n->tcphdr[0]),
+    printf("tcp retrans: seq %lu size %d rcvwnd %u unack %lu rto %ld rtt %ld (RETRY %d cnt %d mem %u)\n",
+	ntohl(n->tcphdr[0].seqnum) - n->cb->iss, n->len - TCP_DATAOFF(&n->tcphdr[0]),
 	n->cb->rcv_wnd, n->cb->send_nxt - n->cb->send_una,
 	n->rto, n->cb->rtt, n->retrans_num, tcp_timeruse, tcp_retrans_memory);
 
@@ -326,20 +326,22 @@ void tcp_retrans(void)
 		    n->cb->rtt = (TCP_RTT_ALPHA * n->cb->rtt + (100 - TCP_RTT_ALPHA) * rtt) / 100;
 		debug_tcp("tcp: rtt %d RTT %ld RTO %ld\n", rtt, n->cb->rtt, n->rto);
 	    }
-	    debug_retrans("retrans remove: seq %ld unack %ld\n",
-		ntohl(n->tcphdr[0].seqnum), n->cb->send_una);
+	    debug_retrans("tcp retrans: remove seq %lu unack %lu\n",
+		ntohl(n->tcphdr[0].seqnum) - n->cb->iss, n->cb->send_nxt - n->cb->send_una);
 	    n = rmv_from_retrans(n);
 	    continue;
 	} else
-	    debug_retrans("retrans check: seq %ld unack %ld time %ld\n",
-		ntohl(n->tcphdr[0].seqnum) + datalen, n->cb->send_una, n->next_retrans - Now);
+	    debug_retrans("tcp retrans: check seq %lu unack %lu time %ld\n",
+		ntohl(n->tcphdr[0].seqnum) - n->cb->iss + datalen,
+		n->cb->send_nxt - n->cb->send_una, n->next_retrans - Now);
 
 	/* check for retrans time up*/
 	if (TIME_GEQ(Now, n->next_retrans)) {
 	    tcp_reoutput(n);
 	    if (n->retrans_num >= TCP_RETRANS_MAXTRIES) {
-		printf("retrans max retries: seq %ld unack %ld time %ld\n",
-		    ntohl(n->tcphdr[0].seqnum), n->cb->send_una, n->next_retrans - Now);
+		printf("tcp retrans: max retries exceeded seq %lu unack %lu time %ld\n",
+		    ntohl(n->tcphdr[0].seqnum) - n->cb->iss, n->cb->send_nxt - n->cb->send_una,
+		    n->next_retrans - Now);
 		tcp_send_reset(n->cb);		/* CB deallocated on received RST*/
 		n = rmv_from_retrans(n);
 		continue;
--- a/elkscmd/ktcp/tcpdev.c
+++ b/elkscmd/ktcp/tcpdev.c
@@ -298,6 +298,7 @@ static void tcpdev_write(void)
     struct tcpcb_s *cb;
     void *  sock = db->sock;
     int size;
+    unsigned int maxwindow;
 
     sock = db->sock;
     /*
@@ -337,13 +338,16 @@ static void tcpdev_write(void)
 	return;
     }
 
-    debug_tcpdev("tcpdev write: window %ld retrans cnt %d\n",
-	cb->send_nxt - cb->send_una, tcp_timeruse);
+    debug_tcpdev("tcpdev write: window %u unack %ld retrans cnt %d\n",
+	cb->rcv_wnd, cb->send_nxt - cb->send_una, tcp_timeruse);
 
     /* delay sending if outstanding send window too large*/
-    if (cb->send_nxt - cb->send_una > TCP_SEND_WINDOW_MAX) {
-	debug_tcp("tcp write: limiting write %d at max send window %ld\n",
-	    size, cb->send_nxt - cb->send_una);
+    maxwindow = cb->rcv_wnd;
+    if (maxwindow > TCP_SEND_WINDOW_MAX)
+	maxwindow = TCP_SEND_WINDOW_MAX;
+    if (cb->send_nxt - cb->send_una > maxwindow) {
+	printf("tcp write: limiting write %d at max window %u unack %ld\n",
+	    size, maxwindow, cb->send_nxt - cb->send_una);
 	retval_to_sock(sock, -ERESTARTSYS);	/* kernel will retry 10ms later*/
 	return;
     }
--- a/elkscmd/ktcp/tcpdev.c
+++ b/elkscmd/ktcp/tcpdev.c
@@ -337,7 +337,7 @@ static void tcpdev_write(void)
 	return;
     }
 
-    /* delay sending if outstanding send window too large*/
+    /* Delay sending if outstanding send window too large. FIXME could hang if no ACKs rcvd*/
     maxwindow = cb->rcv_wnd;
     if (maxwindow > TCP_SEND_WINDOW_MAX)	/* limit retrans memory usage*/
 	maxwindow = TCP_SEND_WINDOW_MAX;
@@ -348,7 +348,7 @@ static void tcpdev_write(void)
 	return;
     }
 
-    printf("tcp write: seq %lu size %d rcvwnd %u unack %lu retrans cnt %d\n",
+    debug_tcp("tcp write: seq %lu size %d rcvwnd %u unack %lu retrans cnt %d\n",
 	cb->send_nxt, size, cb->rcv_wnd, cb->send_nxt - cb->send_una, tcp_timeruse);
 
     cb->flags = TF_PSH|TF_ACK;
--- a/elkscmd/ktcp/tcpdev.c
+++ b/elkscmd/ktcp/tcpdev.c
@@ -343,13 +343,14 @@ static void tcpdev_write(void)
 	maxwindow = TCP_SEND_WINDOW_MAX;
     if (cb->send_nxt - cb->send_una + size > maxwindow) {
 	printf("tcp limit: seq %lu size %d maxwnd %u unack %lu rcvwnd %u\n",
-	    cb->send_nxt, size, maxwindow, cb->send_nxt - cb->send_una, cb->rcv_wnd);
+	    cb->send_nxt - cb->iss, size, maxwindow, cb->send_nxt - cb->send_una, cb->rcv_wnd);
 	retval_to_sock(sock, -ERESTARTSYS);	/* kernel will retry 100ms later*/
 	return;
     }
 
-    debug_tcp("tcp write: seq %lu size %d rcvwnd %u unack %lu retrans cnt %d\n",
-	cb->send_nxt, size, cb->rcv_wnd, cb->send_nxt - cb->send_una, tcp_timeruse);
+    debug_tcp("tcp write: seq %lu size %d rcvwnd %u unack %lu (cnt %d, mem %u)\n",
+	cb->send_nxt - cb->iss, size, cb->rcv_wnd, cb->send_nxt - cb->send_una,
+	tcp_timeruse, tcp_retrans_memory);
 
     cb->flags = TF_PSH|TF_ACK;
     cb->datalen = size;
--- a/elkscmd/ktcp/tcpdev.c
+++ b/elkscmd/ktcp/tcpdev.c
@@ -308,14 +308,14 @@ static void tcpdev_write(void)
 
     /* This is a bit ugly but I'm to lazy right now */
     if (tcp_retrans_memory > TCP_RETRANS_MAXMEM) {
-	printf("ktcp: RETRANS limit exceeded\n");
+	printf("ktcp: RETRANS memory limit exceeded\n");
 	retval_to_sock(sock, -ENOMEM);
 	return;
     }
 
     n = tcpcb_find_by_sock(sock);
     if (!n || n->tcpcb.state == TS_CLOSED) {
-	printf("ktcp: write to unknown socket\n");
+	debug_tcp("tcp: write to unknown socket\n");
 	retval_to_sock(sock, -EPIPE);
 	return;
     }
@@ -342,7 +342,7 @@ static void tcpdev_write(void)
     if (maxwindow > TCP_SEND_WINDOW_MAX)	/* limit retrans memory usage*/
 	maxwindow = TCP_SEND_WINDOW_MAX;
     if (cb->send_nxt - cb->send_una + size > maxwindow) {
-	printf("tcp limit: seq %lu size %d maxwnd %u unack %lu rcvwnd %u\n",
+	debug_tcp("tcp limit: seq %lu size %d maxwnd %u unack %lu rcvwnd %u\n",
 	    cb->send_nxt - cb->iss, size, maxwindow, cb->send_nxt - cb->send_una, cb->rcv_wnd);
 	retval_to_sock(sock, -ERESTARTSYS);	/* kernel will retry 100ms later*/
 	return;
--- a/elkscmd/ktcp/tcpdev.c
+++ b/elkscmd/ktcp/tcpdev.c
@@ -297,8 +297,7 @@ static void tcpdev_write(void)
     struct tcpcb_list_s *n;
     struct tcpcb_s *cb;
     void *  sock = db->sock;
-    int size;
-    unsigned int maxwindow;
+    unsigned int size, maxwindow;
 
     sock = db->sock;
     /*
@@ -338,19 +337,19 @@ static void tcpdev_write(void)
 	return;
     }
 
-    debug_tcpdev("tcpdev write: window %u unack %ld retrans cnt %d\n",
-	cb->rcv_wnd, cb->send_nxt - cb->send_una, tcp_timeruse);
-
     /* delay sending if outstanding send window too large*/
     maxwindow = cb->rcv_wnd;
-    if (maxwindow > TCP_SEND_WINDOW_MAX)
+    if (maxwindow > TCP_SEND_WINDOW_MAX)	/* limit retrans memory usage*/
 	maxwindow = TCP_SEND_WINDOW_MAX;
-    if (cb->send_nxt - cb->send_una > maxwindow) {
-	printf("tcp write: limiting write %d at max window %u unack %ld\n",
-	    size, maxwindow, cb->send_nxt - cb->send_una);
-	retval_to_sock(sock, -ERESTARTSYS);	/* kernel will retry 10ms later*/
+    if (cb->send_nxt - cb->send_una + size > maxwindow) {
+	printf("tcp limit: seq %lu size %d maxwnd %u unack %lu rcvwnd %u\n",
+	    cb->send_nxt, size, maxwindow, cb->send_nxt - cb->send_una, cb->rcv_wnd);
+	retval_to_sock(sock, -ERESTARTSYS);	/* kernel will retry 100ms later*/
 	return;
     }
+
+    printf("tcp write: seq %lu size %d rcvwnd %u unack %lu retrans cnt %d\n",
+	cb->send_nxt, size, cb->rcv_wnd, cb->send_nxt - cb->send_una, tcp_timeruse);
 
     cb->flags = TF_PSH|TF_ACK;
     cb->datalen = size;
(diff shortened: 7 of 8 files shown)


Mellvik:
Thank you @ghaerr,
and yes, I'm getting a lot og the 'limiting write…' messages, eventually we end up in an endless loop printing these messages

tcp write: limiting write 512 at max window 1024 unack 1344

with no network activity at all.
That may not be very interesting at this point though. What I'm seeing is that the recipient closes the window and ktcp continues to bang on the window every 100ms. Which is ok, except that ktcp keep sending the wring packet.

The way I read the packet trace, the transfer would have resumed as soon as the recipient reopens the window if ktcp had used the oldest un-ack-ed packet to knock on the window, instead of the last packet sent.

I suggest that be the next change for continued testing.

I'll upload part of the trace via the web IF.

—Mellvik

ghaerr:
> The way I read the packet trace, the transfer would have resumed as soon as the recipient reopens the window if ktcp had used the oldest un-ack-ed packet to knock on the window, instead of the last packet sent.

I see. `ktcp` keeps a "retransmit list" of every unacked packet... and its sending only the last one, not the entire bunch? Or at least not the earliest one? I think the retransmit list is resent in "timeout" order, which one would think would send the earliest first. I'll look into that.

ghaerr:
> Which is ok, except that ktcp keep sending the wring packet.

Is it possible to send the console "retrans retry" messages? That will show what ktcp is thinking it's doing. Ktcp tries to calculate a complicated RTT (round trip time) for previous packets and uses that to determine when to resend a packet. This all occurs within the `tcp_retrans()` function. I may have to add more debugging statements in there to show what is going on, including the packets skipped for retransmit. Using ^P will show the RTT times, but DEBUG_MEM may need to be turned on also for max info.

I have debugged this before, when fixing the RTO/RTT timings. I am guessing that ktcp may try to resend all the packets in a short period of time, it is possible that, depending on the order sent and the receive window, the server may be dropping the retrans packets as well.

Mellvik:
> 
> Is it possible to send the console "retrans retry" messages? That will show what ktcp is thinking it's doing. Ktcp tries to calculate a complicated RTT (round trip time) for previous packets and uses that to determine when to resend a packet. This all occurs within the tcp_retrans() function. I may have to add more debugging statements in there to show what is going on, including the packets skipped for retransmit. Using ^P will show the RTT times, but DEBUG_MEM may need to be turned on also for max info.
> 
> I have debugged this before, when fixing the RTO/RTT timings. I am guessing that ktcp may try to resend all the packets in a short period of time, it is possible that, depending on the order sent and the receive window, the server may be dropping the retrans packets as well.
> 
Yes, I'm preparing a console trace and packet trace as we speak.

From what I gather from the packet trace, ktcp keeps sending the same packet, which makes sense - except it's the wrong one. It should send only the oldest packet as a 'window knock' until the window opens again. Then the receiver will ack that packet and we're back in business.

—M

Mellvik:
Console output from testing - starting just before the recipient closes the window:
```
tcp write: limiting write 512 at max window 1024 unack 1536                     
tcp write: limiting write 512 at max window 1024 unack 1536                     
tcp write: limiting write 512 at max window 512 unack 1024                      
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
retrans retry: #1 rto 8 (cnt 1, mem 532)                                        
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
retrans retry: #2 rto 16 (cnt 1, mem 532)                                       
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
retrans retry: #3 rto 32 (cnt 1, mem 532)                                       
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
retrans retry: #1 rto 8 (cnt 2, mem 872)                                        
retrans retry: #1 rto 8 (cnt 2, mem 872)                                        
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
retrans retry: #2 rto 16 (cnt 2, mem 872)                                       
retrans retry: #2 rto 16 (cnt 2, mem 872)                                       
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344                     
retrans retry: #3 rto 32 (cnt 2, mem 872)                                       
retrans retry: #3 rto 32 (cnt 1, mem 532)                                       
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344   
```

Packet trace from the same segment, with some comments embedded:
```
13:32:53.096820 IP (tos 0x8, ttl 64, id 19411, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4f7d), ack 652097, win 2048, length 0
13:32:53.183531 IP (tos 0x0, ttl 64, id 1321, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x2669 (correct), seq 652097:652609, ack 1, win 512, length 512
13:32:53.188567 IP (tos 0x0, ttl 64, id 1322, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x0f2d (correct), seq 652609:653121, ack 1, win 512, length 512
13:32:53.193672 IP (tos 0x0, ttl 64, id 1323, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x2dbf (correct), seq 653121:653633, ack 1, win 512, length 512
13:32:53.226822 IP (tos 0x8, ttl 64, id 19412, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4f7d), ack 653633, win 512, length 0   <--- this is where the recipient reduces the window
13:32:53.313522 IP (tos 0x0, ttl 64, id 1324, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xba14 (correct), seq 653633:654145, ack 1, win 512, length 512
13:32:53.318578 IP (tos 0x0, ttl 64, id 1325, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x8464 (correct), seq 654145:654657, ack 1, win 512, length 512 <--- ktcp should not send this one, it's exceeding the window, but then again, it may have sent it before noticing.
13:32:53.356824 IP (tos 0x8, ttl 64, id 19413, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0   <---- windows now closed
13:32:53.433524 IP (tos 0x0, ttl 64, id 1326, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <--- ktcp keeps sending new packets, it should have noticed the window by now.
13:32:53.433599 IP (tos 0x8, ttl 64, id 19414, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0
13:32:53.799608 IP (tos 0x0, ttl 64, id 1327, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512
13:32:53.799733 IP (tos 0x8, ttl 64, id 19415, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0
13:32:54.349959 IP (tos 0x0, ttl 64, id 1328, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512
13:32:54.350087 IP (tos 0x8, ttl 64, id 19416, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0
13:32:55.339950 IP (tos 0x0, ttl 64, id 1329, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512
13:32:55.340042 IP (tos 0x8, ttl 64, id 19417, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0
13:32:57.130211 IP (tos 0x8, ttl 64, id 19418, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:57.183609 IP (tos 0x0, ttl 64, id 1330, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512
13:32:57.183741 IP (tos 0x8, ttl 64, id 19419, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:57.188218 IP (tos 0x0, ttl 64, id 1331, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320
13:32:57.188312 IP (tos 0x8, ttl 64, id 19420, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:57.460174 IP (tos 0x0, ttl 64, id 1332, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320
13:32:57.460264 IP (tos 0x8, ttl 64, id 19421, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:57.474202 IP (tos 0x0, ttl 64, id 1333, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512
13:32:57.474280 IP (tos 0x8, ttl 64, id 19422, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:58.010569 IP (tos 0x0, ttl 64, id 1334, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320
13:32:58.010723 IP (tos 0x8, ttl 64, id 19423, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:58.024825 IP (tos 0x0, ttl 64, id 1335, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512
13:32:58.024902 IP (tos 0x8, ttl 64, id 19424, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:59.000469 IP (tos 0x0, ttl 64, id 1336, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320
13:32:59.000563 IP (tos 0x8, ttl 64, id 19425, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:59.014880 IP (tos 0x0, ttl 64, id 1337, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512
13:32:59.014959 IP (tos 0x8, ttl 64, id 19426, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:33:02.997171 IP (tos 0x8, ttl 64, id 19427, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x997c), ack 654657, win 46080, length 0
13:48:08.250262 IP (tos 0x8, ttl 64, id 19428, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [F.], cksum 0x182b (incorrect -> 0x4f7b), seq 1, ack 654657, win 65024, length 0
13:48:08.253186 IP (tos 0x10, ttl 64, id 52345, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.ftp > 10.0.2.15.1024: Flags [F.], cksum 0x182b (incorrect -> 0x18b7), seq 577, ack 53, win 29200, length 0
``

**Comment by ghaerr** [comment]
@Mellvik, thanks for all your testing on this.

I can't tell from the ktcp console traces exactly which packet is being resent, so I think I'll push more code that displays precisely the packet being resent.

Unfortunately, I can't reproduce this issue on QEMU alone (yet), and I'm traveling so can't access my normal external 386 PC test rig. Did you ever figure out how to setup QEMU to transfer packets to a hosted FTP server? I could load that on macOS.

From your testing comments, it seems what's happening here is that ktcp doesn't realize early enough to stop sending, and sends a few more packets. That timing issue could be somewhat normal though, given async activities between sides. However, ktcp keeps all unacked packets in a retransmit buffer. I'm guessing that the original "group" of unacked packets were sent quite quickly, possibly so quickly that the time stamp on them is such that when the retransmit timeout occurs, more than one packet is resent, and the server receive window may be so tight that some of them are also dropped. It's not clear that ktcp sends the retransmit packets in time-stamped order. I'm not quite sure the fix yet, but we need to be able to see exactly when each packet is sent, rather than just the "retrans \#" which is displayed now.

BTW, another reason ktcp's max send window is always limited to 1024 is that ktcp must malloc memory for each of the unacked packets at the time of sending, for every open connection... this caused ktcp to run out of memory quickly and was one of the first things fixed when we first started fixing TCP. I have seen this behavior before, where ktcp runs into retransmit troubles, and never recovers. I suspect we will need to rewrite the retransmit routine, thinking carefully about exactly when each packet should be retransmitted. We can't just assume that the receiver always gets or loses packets in sequential order. Another question is whether the server will ACK packets previously received - as not doing so would definitely cause ktcp to hang up like we're seeing.

**Comment by ghaerr** [comment]
@Mellvik, please take a look at the attached trace, I have added more comments to it. There is something going on I don't fully understand - the server starts with a small window of 2048, which then goes to 0, while ktcp sends packets. Ultimately, the window size is 0, and it seems irregardless of whether a packet is dropped by the server with zero window size, it ACKs packet 5, after which ktcp sends and resends packet 6 and 7, but the server is hung up on ACKing 5, which it already ACKd. Of course, ktcp will remove packet 5 from the retransmit buffer as soon as it's ACKd, so things fail.

Can you check my comments to see that I am correct in this assessment, and what do you think about it? Far easier to read using textedit on [trace.txt](https://github.com/jbruchon/elks/files/7444192/trace.txt).
```
13:32:53.096820 IP (tos 0x8, ttl 64, id 19411, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4f7d), ack 652097, win 2048, length 0
13:32:53.183531 IP (tos 0x0, ttl 64, id 1321, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x2669 (correct), seq 652097:652609, ack 1, win 512, length 512 <- ktcp packet #1
13:32:53.188567 IP (tos 0x0, ttl 64, id 1322, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x0f2d (correct), seq 652609:653121, ack 1, win 512, length 512 <- ktcp packet #2
13:32:53.193672 IP (tos 0x0, ttl 64, id 1323, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x2dbf (correct), seq 653121:653633, ack 1, win 512, length 512 <- ktcp packet #3
13:32:53.226822 IP (tos 0x8, ttl 64, id 19412, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4f7d), ack 653633, win 512, length 0   <--- this is where the recipient reduces the window, and acks #1-3
13:32:53.313522 IP (tos 0x0, ttl 64, id 1324, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xba14 (correct), seq 653633:654145, ack 1, win 512, length 512 <- ktcp packet #4
13:32:53.318578 IP (tos 0x0, ttl 64, id 1325, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x8464 (correct), seq 654145:654657, ack 1, win 512, length 512 <--- ktcp should not send this one, it's exceeding the window. packet #5
13:32:53.356824 IP (tos 0x8, ttl 64, id 19413, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0   <---- windows now closed, ack #4-5 (this releases #4-5 from ktcp retrans buffer!!!)
13:32:53.433524 IP (tos 0x0, ttl 64, id 1326, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <--- ktcp keeps sending new packets, packet #6
13:32:53.433599 IP (tos 0x8, ttl 64, id 19414, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0 <- server acks #5 again (ERROR:? already acked #5 above)
13:32:53.799608 IP (tos 0x0, ttl 64, id 1327, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <- ktcp retry of packet #6 (server already ack’d #5, so correct)
13:32:53.799733 IP (tos 0x8, ttl 64, id 19415, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0 <- server acks #5 for 3rd time 
13:32:54.349959 IP (tos 0x0, ttl 64, id 1328, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <- ktcp retry #2 of packet #6
13:32:54.350087 IP (tos 0x8, ttl 64, id 19416, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0 <- server acks #5 for 4th time
13:32:55.339950 IP (tos 0x0, ttl 64, id 1329, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <- ktcp retry #3 of packet #6
13:32:55.340042 IP (tos 0x8, ttl 64, id 19417, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0 <- server acks #5 for 5th time
13:32:57.130211 IP (tos 0x8, ttl 64, id 19418, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0 <- server acks #5 for 5th time, increases window greatly
13:32:57.183609 IP (tos 0x0, ttl 64, id 1330, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512 <- ktcp packet #7
13:32:57.183741 IP (tos 0x8, ttl 64, id 19419, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0 <- server acks #5 for 6th time 
13:32:57.188218 IP (tos 0x0, ttl 64, id 1331, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320 <- ktcp packet #8
13:32:57.188312 IP (tos 0x8, ttl 64, id 19420, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0 <- server continually acks #5 from here on out…
13:32:57.460174 IP (tos 0x0, ttl 64, id 1332, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320 <- ktcp retry #1 packet #8
13:32:57.460264 IP (tos 0x8, ttl 64, id 19421, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:57.474202 IP (tos 0x0, ttl 64, id 1333, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512 <- ktcp retry #1 packet #7
13:32:57.474280 IP (tos 0x8, ttl 64, id 19422, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:58.010569 IP (tos 0x0, ttl 64, id 1334, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320 <- ktcp retry #2 packet #8
13:32:58.010723 IP (tos 0x8, ttl 64, id 19423, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:58.024825 IP (tos 0x0, ttl 64, id 1335, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512 <- ktcp retry #2 packet #7
13:32:58.024902 IP (tos 0x8, ttl 64, id 19424, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:59.000469 IP (tos 0x0, ttl 64, id 1336, offset 0, flags [none], proto TCP (6), length 360)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xe3ae (correct), seq 655681:656001, ack 1, win 512, length 320 <- ktcp retry #3 packet #8
13:32:59.000563 IP (tos 0x8, ttl 64, id 19425, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0
13:32:59.014880 IP (tos 0x0, ttl 64, id 1337, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0xbdc3 (correct), seq 655169:655681, ack 1, win 512, length 512 <- ktcp retry #3 packet #7
13:32:59.014959 IP (tos 0x8, ttl 64, id 19426, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0xf57c), ack 654657, win 22528, length 0 <- server sends FIN, why? (was ^C typed?)
13:33:02.997171 IP (tos 0x8, ttl 64, id 19427, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x997c), ack 654657, win 46080, length 0 < server sends FIN again, why?
13:48:08.250262 IP (tos 0x8, ttl 64, id 19428, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [F.], cksum 0x182b (incorrect -> 0x4f7b), seq 1, ack 654657, win 65024, length 0 <- server sends FIN 3rd time
13:48:08.253186 IP (tos 0x10, ttl 64, id 52345, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.ftp > 10.0.2.15.1024: Flags [F.], cksum 0x182b (incorrect -> 0x18b7), seq 577, ack 53, <- server sends FIN on cmd channel
```
> we end up in an endless loop printing these messages

> tcp write: limiting write 512 at max window 1024 unack 1344
>  with no network activity at all..

Yes - I get it now. `ktcp` retries a max of 3 times per packet, which it did for packets \#6, 7 and 8, during which the server ACKed packet \#5, which it had previously ACKed. Finally, after all the retries, ktcp ends up in a loop outputting the "write limit" message with no network writes because the receiver and sender are unacked for more than 1024 bytes after all retries. That particular issue can be fixed later, but it appears we have a more basic out-of-sync issue between ktcp and the remote server.

**Comment by ghaerr** [comment]
Ok @Mellvik, I think I have got this figured out, finally:

What's happening is that the server (rPI) starts sending zero-window ACKs (ACK bit with window size 0). Reading up, this means "SHUT UP, PACKET WILL BE DROPPED" and is part of a slow-start congestion algorithm, I think. Meanwhile, ktcp sends packet \#6, which is zero-window ACK'd \#5, and then proceeds to send the three retries for packet \#6, all of which are zero-window ACK'd \#5. After that, we're screwed because ktcp has MAX_RETRIES of 3 and proceeds to send packet \#7 and further when the window size increases again. However, the server is still ACKing \#5 and there's o way to send \#6 again, since ktcp removed it after 3 retries. The remaining packets that were sent ahead of \#6 eventually get retried all while getting ACK \#5, and finally the process loops/hangs on too large a send window.

A couple fixes are necessary: the first will be to not send an original packet nor a retransmit packet if the receive window is 0. That should fix this problem. The second fix will have to deal with max retries on a packet - we can't remove it if it's unACK'd just because 3 retries were made, nor do we want to continue forever. Perhaps resetting the connection in this case, while also going to 5 retries or so...  ?

I'll push another commit you can test with.

Thanks for your excellent TCP logs and testing!

**Comment by Mellvik** [comment]
@ghaerr,

I think your analysis is pretty much correct, and corresponding to  mine. However, I think we're overcomplicating things.
My take (I'm being boring now, I know): (and oversimplified, possibly) If ktcp can resend the first (oldest) instead of the last, we're good. Everything else seems ok to me. It may take more than 3 retries though, to get the window to open up. Think floppy speed, like a second or two. Then again, repeated 0 window size is a 'special case' flag that can be used to keep the packet longer than MAX_RETRIES.
As far as I am concerned, ktcp should always resend the oldest (lowest seq #) packet when things stop like this, not assuming that unacknowledged packets have made it through.

Also, it's imporamt to remember that the sender must keep knocking, just not too frequently. If the knovk is the first (oldest) unacked packet, things start flowing easily...

More testing tomorrow.

-M

**Comment by ghaerr** [comment]
Last push fixes tcp send window limit calculation by considering current packet size in addition to unack'd data against receive window size. Also won't send retransmit packets if receive window too small.

Both these fixes should solve the problem of TCP sending data when the receive window is too small; rPI in particular uses zero-window ACKs to inform the sender to stop sending immediately, which ktcp wasn't doing. Retransmit packets are also held until the receive window is large enough, as this was the primary reason the two systems got out of sync.

Since write and/or retrans packets may be held when larger than the receive window size and unack'd data, this approach is dependent on receiving a future ACK from the server with a larger receive window size. If this doesn't happen, the system could hang. The rPI system supposedly sends a zero-window ACK when it is processing data, and according to [this article](https://networklessons.com/cisco/ccie-routing-switching-written/tcp-window-size-scaling) it will always send an update ACK with a larger window size. Should that ACK be lost or not retransmitted, the system could hang. I'm not quite sure yet how to handle that case, other than perhaps always sending the retrans packets regardless of receive window size, but perhaps not incrementing the retrans retry count if the receive window is too small. Thoughts?

There are lots of printf statements left in for temp debugging info in this push.

Following is the analysis of the last trace session, confirmed with @Mellvik's packet trace:
```
tcp write: limiting write 512 at max window 1024 unack 1536 | wait packet #4                 
tcp write: limiting write 512 at max window 1024 unack 1536     
              						    | ACK 1-3 received, win 512
							    | send #4, win 512 unack 0
							    | send #5, win 512 unack 512
tcp write: limiting write 512 at max window 512 unack 1024  | wait packet #6
							    | ACK 5 received, win 0
							    | send #6, win 0, unack 0                
tcp write: limiting write 512 at max window 0 unack 512     | wait packet #7                    
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
retrans retry: #1 rto 8 (cnt 1, mem 532)                    | retry 1 packet #6                    
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
retrans retry: #2 rto 16 (cnt 1, mem 532)                                       
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
retrans retry: #3 rto 32 (cnt 1, mem 532)                   | last retry on packet #6
							    | packet 6 free'd, after this
							    | no possibility of resync                     
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 0 unack 512                         
tcp write: limiting write 512 at max window 1024 unack 1344  | server recv window increased
							     | allows more packets to be sent but never synced as #6 was freed from retrans buffer                  
tcp write: limiting write 512 at max window 1024 unack 1344                     
tcp write: limiting write 512 at max window 1024 unack 1344         
```
@Mellvik, please test; I hope this solves the outbound problem, but is completely untested.

One more fix will be needed, but hopefully not seen in your testing: when retrans max retries are exceeded, there's no point in continuing the TCP connection, as ktcp deallocates the saved packet, so it can never be resent. Thus, it would be better to RST the connection, rather than get into the infinite loop seen when the server ACK is always less than the deallocated max retries packet. I will add this once we see a recovery from the zero-window ACKs sent in the rPI testing.

**Comment by Mellvik** [comment]
Thank you @ghaerr, this is great work.
We're coming from totally opposite sides - you looking at the ELKS logs, I look at the tcpdumps - and we end up with the same analysis. That must be good.
Your discussion brought my attention to a couple of things though: I hadn't noticed that ktcp no longer gets several packets ahead while the window is closed. I think your previous PR fixed that.
> Last push fixes tcp send window limit calculation by considering current packet size in addition to unack'd data against receive window size. Also won't send retransmit packets if receive window too small.
> 
> Both these fixes should solve the problem of TCP sending data when the receive window is too small; rPI in particular uses zero-window ACKs to inform the sender to stop sending immediately, which ktcp wasn't doing. Retransmit packets are also held until the receive window is large enough, as this was the primary reason the two systems got out of sync.
> 
> Since write and/or retrans packets may be held when larger than the receive window size and unack'd data, this approach is dependent on receiving a future ACK from the server with a larger receive window size. If this doesn't happen, the system could hang. The rPI system supposedly sends a zero-window ACK when it is processing data, and according to this article it will always send an update ACK with a larger window size. Should that ACK be lost or not retransmitted, the system could hang. I'm not quite sure yet how to handle that case, other than perhaps always sending the retrans packets regardless of receive window size, but perhaps not incrementing the retrans retry count if the receive window is too small. Thoughts?
> 
Thanks for the link, I was unaware that the recipient actually does this. And - like you - I have some concern that some implementations may not (say, will ktcp do that - if we have a back-to-back elks installation?).
Like I've mentioned before, I don't think it's bad behaviour for the client to keep knocking on the window as long as the frequency of knocks is 'benign' - e.g. starting at 50ms, increasing to 1s or something like that.
BTW - I believe the behaviour we're seeing from the RPi (which is debian Linux) is the same on any Linux system, pretty much all of them use the same IP stack AFAIK. Which is good.
> @Mellvik, please test; I hope this solves the outbound problem, but is completely untested.
> 
WIll test later today!
> One more fix will be needed, but hopefully not seen in your testing: when retrans max retries are exceeded, there's no point in continuing the TCP connection, as ktcp deallocates the saved packet, so it can never be resent. Thus, it would be better to RST the connection, rather than get into the infinite loop seen when the server ACK is always less than the deallocated max retries packet. I will add this once we see a recovery from the zero-window ACKs sent in the rPI testing.
> 
Agreed. Then again I think the MAX_RETRIES should be significantly increased. My testcase involves a floppy and some operations may take seconds. Also, we need to think the mentioned ELKS-to-ELKS scenario, which I hope to be testing some time next month.

Thanks again - we're getting there!

--Mellvik

**Comment by ghaerr** [comment]
>  ktcp no longer gets several packets ahead while the window is closed. I think your previous PR fixed that.

Yesterday's push prior to this comment fixed that: now, the packet size is added to the outstanding unacked packets and compared to the receive window size, and the packet delayed for 100ms before checking again if the last-received receive window is too small. As mentioned previously, there is the possibility of total hang if the network were completely disconnected and the receive window never updated, but we'll worry about that issue later.

> Thanks for the link, I was unaware that the recipient actually does this.

It appears that raspian starts with a tiny 2048 byte window (possibly related to our advertised max 512 byte window at SYN time) and then lowers it rapidly as ELKS fires off the first four 512-byte packets non-stop. It ACKs the first 3 with a 512-byte window, and then apparently advertises a zero-size window for some time, as judged by the ACK on packets 4-5.

I'm not sure the reason for the zero-window, but the last push will stop further packets that would overrun the advertised window.

> I don't think it's bad behaviour for the client to keep knocking on the window as long as the frequency of knocks is 'benign' -

Good point - the last push stopped retransmit packets from sending if they might overrun the window; but that sets up the possibility of a deadlock wait, both sides waiting for an ACK packet (with a new receive window) that never comes, if retransmits also wait. I will push another fix that reverts the retransmit behavior to almost as it was: they will always retransmit at the computed time to do so, EXCEPT that their retransmit count will not be incremented if the advertised receive window is too small. This will keep the link active and retransmits will never timeout against a low receive window, and allow for more retransmits in the case they are being dropped by the server when the window size is too small or zero.

>  e.g. starting at 50ms, increasing to 1s or something like that.

This brings up a final point: in our heavily analyzed trace above:
```
13:32:53.433524 IP (tos 0x0, ttl 64, id 1326, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <--- ktcp keeps sending new packets, packet #6
13:32:53.433599 IP (tos 0x8, ttl 64, id 19414, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0 <- server acks #5 again (ERROR:? already acked #5 above)
13:32:53.799608 IP (tos 0x0, ttl 64, id 1327, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <- ktcp retry of packet #6 (server already ack’d #5, so correct)
13:32:53.799733 IP (tos 0x8, ttl 64, id 19415, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0 <- server acks #5 for 3rd time 
13:32:54.349959 IP (tos 0x0, ttl 64, id 1328, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <- ktcp retry #2 of packet #6
13:32:54.350087 IP (tos 0x8, ttl 64, id 19416, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.47529 > 10.0.2.15.1025: Flags [.], cksum 0x182b (incorrect -> 0x4d7d), ack 654657, win 0, length 0 <- server acks #5 for 4th time
13:32:55.339950 IP (tos 0x0, ttl 64, id 1329, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1025 > 10.0.2.2.47529: Flags [P.], cksum 0x29a9 (correct), seq 654657:655169, ack 1, win 512, length 512 <- ktcp retry #3 of packet #6
13:32:55.340042 IP (tos 0x8, ttl 64, id 19417, offset 0, flags [DF], proto TCP (6), length 40)
```
Packet \#6, the problem packet, is sent out initially, along with three retries (then deallocated), all within two seconds. 

Ktcp uses a RTT (round trip time) calculation per connection, whose RTT is updated on every packet ACKd, applied 10% to the previous value. For every original packet sent, a retransmit is allocated, and an RTO (retransmit time out) calculated which starts with 2 times the RTT, and doubles every time. On ethernet, the min allowed RTO is 1/4 sec and the max is 4 secs. 

After packet \#6 was sent out, retry 1 was in approx 3/8 sec, retry 2, 1/2 sec, and retry 3, 1 second (total 1.8 secs). 

> I think the MAX_RETRIES should be significantly increased. My testcase involves a floppy and some operations may take seconds

Agreed. For now, I've increased it to 6, which will allow 3 more. Retry 4 would be 2 secs, with retry 5 and 6 at 4 seconds each (total about 12 secs).Debug output for RTT and RTO calculations can be turned on using printf in `debug_tcp("tcp: rtt %d RTT %ld RTO %ld\n", rtt, n->cb->rtt, n->rto);`, which are displayed in increments of 1/16 seconds (i.e. 16 = 1 second, 4 = 1/4 second, etc). In your next test round, we should be able to see exactly what happens when retries occur during zero-windows, and the associated RTT/RTOs when this is done.

Finally, on max retries (6), an RST will be sent to attempt force closing the connection, since now we know why no recovery is possible when a retransmit packet is deallocated. This should not happen in your testing, but allows future bad connections or bugs to terminate the connection earlier.

**Comment by ghaerr** [comment]
Latest push adds "net stop" support discussed in https://github.com/jbruchon/elks/issues/976#issuecomment-954575632 to terminate `ktcp`, `telnetd`, and `httpd` daemons. The network can be started up again using "net start".

**Comment by Mellvik** [comment]
Great summary @ghaerr - I agree on all accounts.

And I finally got around to look at yeaterday's traces which confirm what I suspected: After I increased the max_retries limit, it works as expected. Recovers and continues! Congratulations!!

There are some more details to be learned from the trace, I'll get back to that later. One thing right away though - just to make sure I'm not creating any misunderstadnings:
> It appears that raspian starts with a tiny 2048 byte window (possibly related to our advertised max 512 byte window at SYN time) and then lowers it rapidly as ELKS fires off the first four 512-byte packets non-stop. It ACKs the first 3 with a 512-byte window, and then apparently advertises a zero-size window for some time, as judged by the ACK on packets 4-5.
> 
Remember that the trace is just a clip starting just before the problem appears. There are thousands of packets before that. The connection starts with a recipient window of approx 30k, slowly increasing to 64k (65535 to be exact), where it stays until the recipient gets busy with other stuff, presumably writing to the floppy. 
Then the window slowly decreases to the level shown in the trace excerpt. Not that it matters, just interesting - the decrease is not proportional to the amount of data received, but seemingly linear, some 800 bytes reduction per 1024 bytes received.
At one point - and this is positively interesting - the window shrinks to 320 and ELKS goes silent! Exactly as it should per this fix.
For 7 seconds(!!). Until the receiver suddenly opens up again, with a 48k window. This is good - but it wouldn't harm if ELKS kept knocking, it would probably speed things up. I'll look into that when testing the latest fixes.
Have a look at this trace, which starts at the point where the receiver has reduced the windo to 832 bytes:

13:05:12.999533 IP (tos 0x8, ttl 64, id 47785, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.56425 > 10.0.2.15.1029: Flags [.], cksum 0x182b (incorrect -> 0xca7b), ack 684097, win 832, length 0
13:05:13.023915 IP (tos 0x0, ttl 64, id 4233, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1029 > 10.0.2.2.56425: Flags [P.], cksum 0x2ea6 (correct), seq 684097:684609, ack 1, win 512, length 512
13:05:13.066841 IP (tos 0x8, ttl 64, id 47786, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.56425 > 10.0.2.15.1029: Flags [.], cksum 0x182b (incorrect -> 0xca7b), ack 684609, win 320, length 0
13:05:20.612708 IP (tos 0x8, ttl 64, id 47787, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.56425 > 10.0.2.15.1029: Flags [.], cksum 0x182b (incorrect -> 0x0bbb), ack 684609, win 49152, length 0
13:05:20.645168 IP (tos 0x0, ttl 64, id 4234, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1029 > 10.0.2.2.56425: Flags [P.], cksum 0x276d (correct), seq 684609:685121, ack 1, win 512, length 512
13:05:20.645340 IP (tos 0x8, ttl 64, id 47788, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.56425 > 10.0.2.15.1029: Flags [.], cksum 0x182b (incorrect -> 0x0bbb), ack 685121, win 48640, length 0
13:05:20.670046 IP (tos 0x0, ttl 64, id 4235, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1029 > 10.0.2.2.56425: Flags [P.], cksum 0xe480 (correct), seq 685121:685633, ack 1, win 512, length 512
13:05:20.695124 IP (tos 0x0, ttl 64, id 4236, offset 0, flags [none], proto TCP (6), length 552)
    10.0.2.15.1029 > 10.0.2.2.56425: Flags [P.], cksum 0xfe44 (correct), seq 685633:686145, ack 1, win 512, length 512
> 
More later - there are a lot of traces to digest.


--Mellvik

**Comment by Mellvik** [comment]
More good news, @ghaerr - 

I've tested the latest updates to this PR - and it's all good news:

- Transfer recovers smoothly from delays.
- The server (recipient) resets the connection on error, it's all handled well.
- There are NO retransmits whatsoever, which is good in the sense that it shows everything working as intended, and something to fix per our previous discussions (window knocking).
- net stop works

As usual; I've run all tracing on TCPdump - and it's very encouraging.

Here's some console output from ELKS:

This is from the middle of the transfer:

tcp limit: seq 901121 size 512 maxwnd 1024 unack 1024 rcvwnd 4992               
tcp limit: seq 902145 size 512 maxwnd 1024 unack 1024 rcvwnd 3968               
tcp limit: seq 903169 size 512 maxwnd 1024 unack 1024 rcvwnd 2944               
tcp limit: seq 904001 size 512 maxwnd 1024 unack 832 rcvwnd 1920                
tcp limit: seq 905025 size 512 maxwnd 1024 unack 1024 rcvwnd 1088               
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 906561 size 512 maxwnd 1024 unack 1024 rcvwnd 23040              
tcp limit: seq 907585 size 512 maxwnd 1024 unack 1024 rcvwnd 22016              
tcp limit: seq 908609 size 512 maxwnd 1024 unack 1024 rcvwnd 20992              
tcp limit: seq 909633 size 512 maxwnd 1024 unack 1024 rcvwnd 19968              
tcp limit: seq 910657 size 512 maxwnd 1024 unack 1024 rcvwnd 18944              
tcp limit: seq 911681 size 320 maxwnd 1024 unack 1024 rcvwnd 17920              
tcp limit: seq 912513 size 512 maxwnd 1024 unack 832 rcvwnd 16896               
tcp limit: seq 913537 size 512 maxwnd 1024 unack 1024 rcvwnd 16064              
tcp limit: seq 914561 size 512 maxwnd 1024 unack 1024 rcvwnd 15040              
tcp limit: seq 915585 size 512 maxwnd 1024 unack 1024 rcvwnd 14016              
tcp limit: seq 916609 size 512 maxwnd 1024 unack 1024 rcvwnd 12992              
tcp limit: seq 917633 size 512 maxwnd 1024 unack 1024 rcvwnd 11968              
tcp limit: seq 918657 size 512 maxwnd 1024 unack 1024 rcvwnd 10944              
tcp limit: seq 919681 size 320 maxwnd 1024 unack 1024 rcvwnd 9920               
tcp limit: seq 920513 size 512 maxwnd 1024 unack 832 rcvwnd 8896                
tcp limit: seq 921537 size 512 maxwnd 1024 unack 1024 rcvwnd 8064               
tcp limit: seq 922561 size 512 maxwnd 1024 unack 1024 rcvwnd 7040               
tcp limit: seq 923585 size 512 maxwnd 1024 unack 1024 rcvwnd 6016               
tcp limit: seq 924609 size 512 maxwnd 1024 unack 1024 rcvwnd 4992               
tcp limit: seq 925633 size 512 maxwnd 1024 unack 1024 rcvwnd 3968               
tcp limit: seq 926657 size 512 maxwnd 1024 unack 1024 rcvwnd 2944               
tcp limit: seq 927681 size 320 maxwnd 1024 unack 1024 rcvwnd 1920               
tcp limit: seq 928513 size 512 maxwnd 896 unack 832 rcvwnd 896                  
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 928513 size 512 maxwnd 64 unack 0 rcvwnd 64                      
tcp limit: seq 930049 size 512 maxwnd 1024 unack 1024 rcvwnd 23040              
tcp limit: seq 931073 size 512 maxwnd 1024 unack 1024 rcvwnd 22528              


This is from the termination - the recipient runs out of space:

tcp limit: seq 1566145 size 512 maxwnd 1024 unack 1024 rcvwnd 61952             
tcp limit: seq 1567169 size 512 maxwnd 1024 unack 1024 rcvwnd 60928             
tcp limit: seq 1573121 size 512 maxwnd 1024 unack 1024 rcvwnd 65024             
tcp limit: seq 1574145 size 512 maxwnd 1024 unack 1024 rcvwnd 64512             
tcp limit: seq 1584001 size 512 maxwnd 1024 unack 832 rcvwnd 65024              
tcp: RST from 10.0.2.2:37659->1027                                              
ktcp: write to unknown socket                                                   
ftpput: Network write error, transfer aborted.                                  
#

Removing the 'tcp limit' messages and we're all set (maybe keep on the list to find out why QEMU needs those windows size reductions at all …).

Again, congratulations - this has been another interesting journey.

—Mellvik

**Comment by ghaerr** [comment]
Hello @Mellvik, I'm glad to hear everything is working!

I'll change the "tcp limit" messages to be displayed only under DEBUG_TCP, commit this, and close #986.
Additional errors can be discussed in a new issue.

> it wouldn't harm if ELKS kept knocking, it would probably speed things up. 

We have "knocking" in the retransmit code, which we haven't seen or tested with the latest fixes, probably because there haven't been any actual packet losses, since the sending code now respects the receive window. The retransmit code also provides a timeout-driven mechanism for "knocking" when an ACK or other packet is lost.

We do have a timer set for retrying writes on a receive window limit, but that mechanism doesn't (yet) send a lone ACK on a timer should the receive window never be updated. We will probably have to wait for that until we can think of a way to test it, as it appears the Linux TCP stacks do send timeout ACK updates when their window size changes. Of course, one way to test would be to unplug the (SLIP) connection in the middle of a transfer, then replug. I suggest we work on getting "normal" transfers completely working before worrying about total robustness in the ELKS stack, as there's still a number of things areas it doesn't conform to the specification.

Finally, I don't think the lack of a knocking timer while waiting for a window size increase will speed things up, as we're technically not allowed to send over the receive window, and the Linux TCP always sends a window update, which allows ELKS to start sending again. So things are working in the normal case!

Thanks for all your help on this, another major TCP bug fixed!

**Comment by Mellvik** [comment]
Thank you @ghaerr - 
I agree - we're there, and what we're now discussing is akin to the finer print of the deal ...

> 
> it wouldn't harm if ELKS kept knocking, it would probably speed things up.
> 
> We have "knocking" in the retransmit code, which we haven't seen or tested with the latest fixes, probably because there haven't been any actual packet losses, since the sending code now respects the receive window. The retransmit code also provides a timeout-driven mechanism for "knocking" when an ACK or other packet is lost.
> 
In one of the recent traces, I commented that the server waited 7 seconds before reopening the window. If there was a timeout for knocking from the elks side, I would have expected it to kick in by then.

This is also  the reason for my comment about speed. Having a 'knock' from ELKS would reveal whether the server side reopens immediately or after a timeout, in the latter case the knock would speed up things. 

Admittedly this is more like esoteric stuff, and I'm fine with letting it go for now. We have something that works, seems solid and is within standards. That's pretty good by any metric. A major improvement.


Thanks again.

—M

**Comment by ghaerr** [comment]
> This is also  the reason for my comment about speed. Having a 'knock' from ELKS would reveal whether the server side reopens immediately or after a timeout, in the latter case the knock would speed up things.


There isn't such a thing as a general "knock" or ping for TCP. In the case of a sending TCP bumping up against a smaller receive window, the spec says to stop sending - not to start pinging. That is, it is normal to have an outstanding send window where the transmitter stops, and awaits ACKs before continuing. I'm pretty sure this is what's happening. For the sender to send more "pings" (whatever packet that might actually be) during a send window wait may not be correct, and would contribute to network congestion.

In the case of an advertised zero window, the spec treats that somewhat different, and there are some RFCs that discuss ZWP (zero window probing). We aren't seeing that anymore, likely because our flow transmission is working. Ktcp does not currently implement ZWP.

My made-up idea of a "knock" would be sending an ACK packet. However, in the case of a sending system, the receiver is supposed to be ACKing, not the sender. An ACK from the sending side implies something about the traffic in the reverse direction. So there's a possibility of having an incorrect sequence or ack number in the "knock" packet that could cause more problems. Thus, I think we need a better understanding of exactly how TCP flow control is supposed to work with sliding windows; since there's no such thing as a general "knock" packet, this means a better understanding of packet resending on the client and ACK resending on the server. I'll read up on it.

> I commented that the server waited 7 seconds before reopening the window. If there was a timeout for knocking from the elks side, I would have expected it to kick in by then.

Was this when you were writing a floppy on raspian? If so, it's probably just normal behavior, given your comments on floppy time delays.

I would summarize that we're likely seeing correct flow behavior, and shouldn't probably be guessing about adding extra packet sends to speed things up. One of the things I originally liked about `ktcp` is it's a well-written state machine (at its core), even though there's been lots of bugs and missing features. The state machine seems to follow the specification.

Thanks for your help and insights.

**Comment by Mellvik** [comment]
Thank you, @ghaerr.
> This is also the reason for my comment about speed. Having a 'knock' from ELKS would reveal whether the server side reopens immediately or after a timeout, in the latter case the knock would speed up things.
> 
> There isn't such a thing as a general "knock" or ping for TCP.
> 
Yeah, I know, I've been teaching this stuff since the 80s.
> In the case of a sending TCP bumping up against a smaller receive window, the spec says to stop sending - not to start pinging. That is, it is normal to have an outstanding send window where the transmitter stops, and awaits ACKs before continuing.
> 
Yes, this is exactly what's happening. And while it is very minor in the overall scheme of things, it's an interesting discussion.
Which must start with a basic premise of standards in general and tcp/ip in particlular: not everything is in the standard, and some times standards  are contradictory.

> I'm pretty sure this is what's happening. For the sender to send more "pings" (whatever packet that might actually be) during a send window wait may not be correct, and would contribute to network congestion.
> 
No. An occasional 'knock' can never contribute to congestion.
> In the case of an advertised zero window, the spec treats that somewhat different, and there are some RFCs that discuss ZWP (zero window probing). We aren't seeing that anymore, likely because our flow transmission is working. Ktcp does not currently implement ZWP.
> 
You made the point before that if the ACK reopening the window is lost, we may have a hung situation. Not likely because the lost ACK would eventually be resent. But - and here's my point - the extra knock from the sender would avoid that. And there is nothing in the standard making it illegal, or even bad behaviour. Rather - and again, my take - the typical extra precaution that keeps the network ticking. 
> My made-up idea of a "knock" would be sending an ACK packet.
> 
No, it would be repeating the oldest (in our case the only) unACKed acket.
> However, in the case of a sending system, the receiver is supposed to be ACKing, not the sender. An ACK from the sending side implies something about the traffic in the reverse direction.
> 
No, I don't think do. What might that be? Repeating an unacknowledged packet is just about the most ordinary thing on the net.
> So there's a possibility of having an incorrect sequence or ack number in the "knock" packet that could cause more problems.
> 
I just don't see that. To me this is simply good (polite) behaviour.
> Thus, I think we need a better understanding of exactly how TCP flow control is supposed to work with sliding windows; since there's no such thing as a general "knock" packet, this means a better understanding of packet resending on the client and ACK resending on the server. I'll read up on it.
> 
I think we both have a perfectly good understanding of the mechanics. The sliding windows now work fine in ktcp - with retransmissions and the whole shebang. You fixed that...
> I commented that the server waited 7 seconds before reopening the window. If there was a timeout for knocking from the elks side, I would have expected it to kick in by then.
> 
> Was this when you were writing a floppy on raspian? If so, it's probably just normal behavior, given your comments on floppy time delays.
> 
Yes, it's normal. But if there was a timeout for a retransmission (knock), which was what I responded to, it should have kicked in long before 7 seconds.
> I would summarize that we're likely seeing correct flow behavior, and shouldn't probably be guessing about adding extra packet sends to speed things up.
> 
We are not. And again, you brought this up - and we're simply talking about a sound precaution against a hung situation.
> One of the things I originally liked about ktcp is it's a well-written state machine (at its core), even though there's been lots of bugs and missing features. The state machine seems to follow the specification.
> 
I agree, we're seeing correct behaviour. What we're discussing is robustness. Let's leave it at that. 
I think the state we're at is worth a celebration. It has been a long journey from 'hello world' to the first shell prompt - figuratively speaking,
> 

-- Mellvik

**Comment by ghaerr** [comment]
@Mellvik, I'm not in full agreement with your comments, or at least am having problems moving from generalizations to specific things that might be changed with code. I think the reason is that I'm in the middle of a detailed implementation, while you are talking about higher level concepts. In the interests of clarification, and to learn what I might be missing, I am continuing to respond to some of your statements, thank you!

> You made the point before that if the ACK reopening the window is lost, we may have a hung situation. 

Yes. I will get to that specific situation later.

> Not likely because the lost ACK would eventually be resent.

No, that's not how it works. When a sender doesn't receive an ACK, the receiver never sends another ACK. The receiver wouldn't know when to send one anyways. Instead, the sender retransmit timer solves the problem - never by sending an ACK, but by resending any packets for which no ACK was received. Thus, this general "knock" you're talking about in this case is a normal packet resend (not to be confused with an ACK response, as there never is one).

> But - and here's my point - the extra knock from the sender would avoid that. 

So there's no "extra knock", just retransmits after timeouts, period.

Since in the last testing round we saw no retransmits, then this particular "extra knock" cannot be done otherwise.

Now lets get into the actual remaining problem in ktcp: what to do when a "tcp limit" situation occurs. Now, ktcp tells the kernel to continue waiting the application `write` system call and reschedules the kernel to wake it up 100ms later, and retry the system call. This will continue forever, with no other actions taken.

This only occurs when the receive window is smaller than the combined packet size and outstanding send window; there are two cases to consider: that of when there is (1) an outstanding unacked send window, and the case (2) where there is no outstanding unacked packets, but the receive window is less than the packet size, or zero.

Here's a portion of the last trace showing both conditions:
```
tcp limit: seq 904001 size 512 maxwnd 1024 unack 832 rcvwnd 1920 <- unack 832, outstanding sent unacked packets
tcp limit: seq 905025 size 512 maxwnd 1024 unack 1024 rcvwnd 1088 <- same
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64 <- no outstanding packets, receive window < packet size
tcp limit: seq 905025 size 512 maxwnd 64 unack 0 rcvwnd 64 <- same
```
In the first case above, since there is outstanding unacked packets, then hanging in a busy loop waiting should never be a problem, since the receiver *necessarily* has to ACK those previously sent packets. Should those ACKs be lost, the packet retransmit code will resend the packets. I presume here your "knock" means retransmit (not ACKing), ktcp is coded to do that.

The second case above is the problem situation. ELKS TCP will hang forever if the receive window is less than the packet size. Since ktcp always transmits the same size packets as sent by the application, it will never rebuild a 512-byte packet and send 64 bytes, as could be done above. (This is one good thing about our smaller packet size, I suppose).  In addition, and only in this case where there are no unacked packets, we have a problem. This could happen either when the receiver arbitrarily lowers the receive window to less than 512, or to 0 (zero window, a separate RFC distinction), or if even in the initial connection the receive window is less than 512 (probably won't run into this). 

I am not aware of what to do, other than wait, when the receive window size is less than the intended packet size (or zero), and there are no unacked outstanding packets. The sender can't ACK anything, since it didn't send anything (yet). And a receiver doesn't ACK an ACK, except in SYN or FIN processing. Is your "extra knock" a suggestion to probe on a non-zero window? 

For the zero window case, the standard talks about ZWP (zero window probing). What is your knowledge of how this is handled? I haven't found exactly what the sender sends to probe. We might be able to use a similar mechanism to ZWP, but this could be quite tricky, as the standard allows zero window to be used to, say, stop print jobs from going to a printer that's out of paper for two days, while still requiring the connection to be open (and not probed every 100ms, that's what I meant by network congestion).

> But if there was a timeout for a retransmission (knock), which was what I responded to, it should have kicked in long before 7 seconds.

Since we never hung, that means either the receiver advertised a lower window and then sent an update ACK for last packet received, in which case we responded within 100ms, OR a retransmission occurred as a result of RTT/RTO timeout. Since we didn't see any retransmit activity, I didn't think this was any bug in ktcp. However, the RTT/RTO calculations (max 4 seconds and ~12 seconds max over 6 retries) still apply, and they could be buggy and not retransmit when they should have. We might want to debug this, but probably need RTT/RTO and receive window to be printed out every packet send. I can push this if you'd like.

> You made the point before that if the ACK reopening the window is lost, we may have a hung situation. 

Yes, that and any packet waiting on send window limit with no unacked data.

> Not likely because the lost ACK would eventually be resent. 

No, lost ACKs are never resent. Only packet retransmits. And those would not be resent if they aren't sent in the first place.

> But - and here's my point - the extra knock from the sender would avoid that. And there is nothing in the standard making it illegal, or even bad behaviour. Rather - and again, my take - the typical extra precaution that keeps the network ticking.

What exactly is the extra knock packet you're talking about, and what flags are set in it?

> I think the state we're at is worth a celebration.

I suppose we shouldn't be too worried about a theoretical hang until we actually have a problem, given all the problems we've already solved!

Thank you!

ghaerr:
> What exactly is the extra knock packet you're talking about, and what flags are set in it?

In reading up more on zero window probing (hard to find much info), it seems that a ZWP packet is a non-ACK packet sent somewhat differently depending on the OS. In Linux, a ZWP is almost entirely the same as a retransmit packet, except 0-length data. That is, no ACK, just 0 data, with the last sequence number that would have been used in a retransmit packet. Windows apparently sends one more byte of the data (which some kernels reject). We can't do that, as we have to send either the whole application packet or nothing.

Thus, one solution would be to effectively queue a zero-length "retransmit" packet, but not send the actual data packet (it would remain in the kernel write routine). This would be tricky, as the kernel task is rewoken every 100ms and we don't want to requeue a retransmit packet each time. Also, this doesn't solve the problem of non-zero receive window waiting, when the specs says we should do nothing, but would solve the (unseen) problem of when the receive window is always less than 512 bytes.

[EDIT: Also, normally a TCP stack will never hold data on a non-zero receive window; it will instead decrease the packet size to the receive window and send that. Ktcp will never do that, so, for ELKS, we should consider any receive window less than the current packet size, with no outstanding ACKs, to be equivalent to a zero window. We could then use ZWP processing for that case as well. This seems to be why none of the literature talks about what to do if the receive window is less than the packet size - the sender will send a smaller packet size normally until the receive window is zero.]

Very [interesting article (section 4.3)](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.24.7007&rep=rep1&type=pdf) on zero window probing, using a retransmitted data segment as a ZWP.

I think perhaps we should wait on all this, and continue to test "normal" operation, trying to see if we can get further failures. Once that's done, diving into robustness by purposely interrupting the connection might be worth trying.

I got caught up in trying to understand "resending the ACK", when "resending the packet" or a zero-length one, does solve the problem. This is probably the "extra knock" you've been talking about!

Mellvik:
I agree, @ghaerr - 
the discussion got out of hand on technicalities the had different meanings to us. The 'knock' I have been referring to, was always a retransmit of a the oldest (usually the only) unacked data packet.

So let's get back to where we started which is just as you summarize:
To resend the unacked data packet even if the window is closed. With a low(er) frequency and - like you suggested - not counting against the max_retry_limit.

Thank you.

—Mellvik

ghaerr:
> To resend the unacked data packet even if the window is closed. With a low(er) frequency and - like you suggested - not counting against the max_retry_limit.

Yes, so that should completely handle the case where the receiver shuts the sender down, even for extended periods with zero window. The remaining case , that of a sender waiting for a packet to be sent, with no outstanding unacked data already sent, may only fail if the receiver's ACK is lost. That won't be handled (yet), as it is unlikely to lose packets on a LAN or SLIP connection.

So - at this time we think that we have finally fixed all the issues that ELKS TCP had with losing sync between systems where there is no packet loss, or when the user force terminates the session. Nice! We should continue pushing this scenario for the time being to ensure this is the case. I'll try to get QEMU setup to test FTP transfers as well and report any problems.

Thank you!

Mellvik:
Indeed @ghaerr - 

and testing will continue as the new FTP server comes to life, then the FTP client.
Ample opportunity to create havoc - if possible. Then there is the ELKS back-to-back scenario coming up…

BTW - on my list is an attempt to recreate the jittery response you're getting in telnet when the window is 'lett loose' - can you send me the qemu script separately?

Thank you!

—Mellvik

ghaerr:
> as the new FTP server comes to life

A tiny FTP server modeled after tiny `httpd` would be nice. Are you writing that from scratch?

Mellvik:
Hi @ghaerr,
That depends on the definition of 'scratch'.

I'm combining httpd with elements from urlget (net_connect, net_close etc.) and a simple ftpclient/server pair from github (the client is a seperate project).

A few busy days now, but I expect to ne testing it on friday this week.

-- Mellvik

PR #988, opened 2021-11-01
[net] Use GCC builtin for ntohs/ntohl macros

ghaerr:
Use gcc builtins for better code size/speed in network applications.
Also minor cleanups in `ktcp`.

No need for additional testing.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -163,6 +163,7 @@ inet/telnetd/telnetd			:net
 inet/tinyirc/tinyirc			:net
 inet/urlget/urlget				:net
 inet/urlget/urlget :::ftpget	:net
+inet/urlget/urlget :::ftpput	:net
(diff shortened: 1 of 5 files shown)


ghaerr:
Also add `ftpput` hard link to `urlget` in image build.

@Mellvik, I'm now able to test using `ftpget`, `ftpput`, and `urlget` in both directions from ELKS to a macOS ftpd. Things work well, I can see a few small cleanups, like not resetting the connection when a remote file is not found (i.e. `netclose(..., 0)` in that case), that could be made. The biggest issue I'm having so far is that ktcp runs out of internal memory for TCP control blocks if ftpget (or any other network program) is run back-to-back too quickly, as the CLOSING and TIME-WAIT counters continue to consume memory after program exit. I don't want to increase ktcp memory just yet, am still thinking of possible ways to fix that without having to wait 10 seconds. More testing coming.

PR #989, opened 2021-11-02
[cmds] Fine tune compressed executables

ghaerr:
In thinking about the usability of ELKS on floppy, (and in preparation for a v0.5.0), I noticed while testing ftpget transfers that there is only 69K free on a 1.44Mb floppy, not even enough to transfer a kernel for temp storage. We are also almost completely out of space on our 360K boot floppy. More space is needed if one is actually trying to use ELKS for something real.

Quite a bit of work was put in to ELKS compressed executables in v0.4.0, but a distribution floppy of compressed executables was never produced. In addition, there were some comments made about the usability and speed of some compressed programs on older systems, based on the tradeoff of program disk size vs decompression speed. Looking through the large variety of (mostly unused) programs on the 1.44Mb floppy distribution, it was found that if all executables were compressed, there would be a whopping 327K made available! This would greatly improve the actual usability of running ELKS on non-hard drive systems.

This PR produces the standard disk image with all executables compressed, if CONFIG_EXEC_COMPRESS is set in config, which is default ON. Changes were made to `elks-compress` so that a wildcard could be used to compress all applications at once, for simplicity. It will work properly with symlinks as well as some edge cases that were previously rejected.

The issue now may be, which executables should not be compressed? 

@Mellvik, would you be willing to test a bit to see how this feels? Given the ultimate increased usability, it seems we're better off, even on slower systems, sacrificing some speed for utility in much increased free floppy space. Certainly, most programs should be compressed, and possibly all programs that are used only once, like /bin/init. /bin/sh is a bit trickier, in that we could compress just the text section, which might allow for the data section to be left uncompressed for slightly faster shell scripts on floppies (this would be tested using `elks-compress -d target/bin/sh`). Another option is to rewrite the kernel decompression routine in ASM, which would probably help the speed.

Of course, a user option to support compressed executables (CONFIG_EXEC_COMPRESS), but not write any (not ON by default, but selectable) could be a good option. That's not yet added in this PR. Another option would be a per-application selection of which NOT to compress. This PR just changes one line in images/Make.images to do the compression of all executables for the time being.

### DIFF
--- a/elks/arch/i86/defconfig
+++ b/elks/arch/i86/defconfig
@@ -243,6 +243,7 @@ CONFIG_IMG_FD1440=y
 # CONFIG_IMG_HD is not set
 CONFIG_IMG_DEV=y
 CONFIG_IMG_BOOT=y
+# CONFIG_APPS_COMPRESS is not set
 
 #
 # Binary Images
--- a/elks/tools/elks-compress/elks-compress.c
+++ b/elks/tools/elks-compress/elks-compress.c
@@ -12,9 +12,6 @@
 #include <sys/stat.h>
 #include "minix.h"
 
-int do_text = 0;
-int do_ftext = 0;
-int do_data = 0;
 int keep_infile = 0;
 int verbose = 0;
 char exomizer_binary[] = "exomizer";
@@ -58,7 +55,7 @@ static void copyrest(int ifd, int ofd, char *filename)
 	int size;
 	char buf[1024];
 
-	while ((size = read(ifd, buf, 1024)) > 0)
+	while ((size = read(ifd, buf, sizeof(buf))) > 0)
 	{
 		if (write(ofd, buf, size) != size)
 		{
@@ -72,7 +69,7 @@ error:
 	if (size != 0) goto error;
 }
 
-static int compress(char *infile, char *outfile)
+static int compress(char *infile, char *outfile, int do_text, int do_ftext, int do_data)
 {
 	int ifd, ofd, efd, n;
 	long orig_size, compr_size;
@@ -103,7 +100,7 @@ static int compress(char *infile, char *outfile)
 			(elks_size_t)mh.tseg == 0)
 	{
 		printf("Exec header not supported: %s\n", infile);
-		return 1;
+		return 2;
 	}
 
 	if (mh.hlen == EXEC_FARTEXT_HDR_SIZE)
@@ -255,9 +252,10 @@ static int compress(char *infile, char *outfile)
 		}
 		if (sbuf.st_size > szdata)
 		{
-			printf("Rejecting conversion of %s: compressed data larger than data\n", infile);
+			printf("Rejecting data conversion of %s: compressed data larger than data\n", infile);
 			unlink(exomizer_outfile);
-			return 2;
+			do_data = 0;
+			goto next;
 		}
 		if (sbuf.st_size >= 65520)
 		{
@@ -281,6 +279,7 @@ static int compress(char *infile, char *outfile)
 		if (verbose) printf("compressed data from %d to %d\n", szdata, compr_szdata);
 	}
 
+next:
 	if ((ofd = creat(outfile, 0755)) < 0) {
 		printf("Can't create %s\n", outfile);
 		close(ifd);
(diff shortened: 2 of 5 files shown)


ghaerr:
Add 'Compressed executables' config option (default OFF) to menuconfig.
`chmem` displays compressed text, ftext and data sizes if executable compressed.

Mellvik:
Thank you @ghaerr, very pertinent and timely issue.

First of all - the binary compression support was an impressive feat which AFAIK turned out to have a disappointingly high penalty. Now that it's there and is working, looking into how to improve it is probably worthwhile. And I will enthusiastically participate in physical testing - on 286 ans 386 based systems. What not to compress? I'd say sh and a few very frequently used commands (to be determined - I'm sure we all have our personal favourites, my #2 would be vi…).

That said, it may be a good idea to look at targeted 'distributions'. Say, remove all the nx* programs (which I do all the time) opens up 120K right there. By Targeted I mean 'this package for developers, this for passers by, if we can define 2 or 3 such profiles.

In any case, having an 'ALL_FILES_COMPRESSED (except sh)' is a good start.

I might be interested in asm-coding the decompressor, or maybe you'd like to go there yourself? :-)

—Mellvik

Mellvik:
@ghaerr,
while we're at it - 

urlget is now 4 programs. If the floppy generating procedure creates hard links instead of symlinks, we save one disk block per program (except the 'first'), and improve startup speed a tiny little bit. I haven't put in the creation of ftpput yet btw. All the little things: A block here, a block there ...

The other issue is not related to distribution/boot floppies and space, but a practical thing:
With the ftp-daemon on the way we get yet another daemon to consume RAM - if we choose to start it. This may be a good time to consider a BSD-style inetd, which kicks in telnetd, httpd, ftpd, whatever is in the /etc/services file, as required. It's a very simple program and would - in the general case - save RAM and keep the network daemons available.  

Any thoughts?

—Mellvik

ghaerr:
> the binary compression support was an impressive feat which AFAIK turned out to have a disappointingly high penalty. 

IIRC, this was tested with CONFIG_TRACK_CACHE turned on, which has it's own possible degradation issues, that of reading an entire track when much less is needed. OTOH, reading large programs like `vi` are probably very much helped by that option.

I know you detailed much of that in a previous discussion, we need to think of some ways of quantifying the disk speed tradeoffs with track caching combined with decompressing savings and costs, the interaction is kind of complicated.


> Now that it's there and is working, looking into how to improve it is probably worthwhile. And I will enthusiastically participate in physical testing - on 286 ans 386 based systems. 

> What not to compress? I'd say sh and a few very frequently used commands (to be determined - I'm sure we all have our personal favourites, my #2 would be vi…). 

It's not clear that `sh` should not be compressed - as I tried to clarify above, once the text segment is in memory, the kernel shared memory support means the text (code) portion of the disk file does not need to be re-read. That means that we should be a able to possibly compress the text portion, but not the data portion, depending on the relative size. 


> That said, it may be a good idea to look at targeted 'distributions'. Say, remove all the nx* programs (which I do all the time) opens up 120K right there. By Targeted I mean 'this package for developers, this for passers by, if we can define 2 or 3 such profiles.

Agreed, but I think we're not there yet in the number of ELKS users to go to the trouble of individual application selection (other than editing the elkscmd/Applications file).

That said, selecting exactly which programs are put on the target, and selecting exactly which programs are compressed, is pretty much the same work, once we figure how it should be done. I don't have much feedback on which users prefer the "disk image size" selection versus the "application category" selection is used. Perhaps they need to be combined: i.e. the target image size just selects the target disk size, and application category is always used to select groups of programs to be written to the image. We could invent finer grained selection for the application categories than we have now. (See menuconfig "Userland" option for details).


>  In any case, having an 'ALL_FILES_COMPRESSED (except sh)' is a good start.

Ok. I think I'll commit this now, since the default is no compression, and the discussion can be continued as we contemplate how to select which files are compressed to the image. I'm thinking that perhaps using finer-grained application categories as discussed above, and duplicated with the ability to select compressed for each category or not, might fit the bill.

ghaerr:
>  I might be interested in asm-coding the decompressor, or maybe you'd like to go there yourself? :-) 

The decompression is a bit complicated, as there are lots of compress options, we're using the `exomizer raw -q -C -b` option(s) to compress the executable sections. I built a version that lives outside of the kernel that made initial work and testing a lot simpler, which is in 'elks/tools/elks-compress/exomizer/rawdecrs/exodecr-elks.c::exo_decrunch()'. There is already a contributed 8086 ASM decoder 'elks/tools/elks-compress/exomizer/rawdecrs/8086/deexo.asm.txt' that works with the `exomizer raw -P47 -b` decompressor, but I can't get the P47 compression option to work with the C decompressor.  IIRC, the ASM decoders take advantage of a much smaller lookup table, instead of the compressed-size table being used now, which allows for a much faster decode to be implemented, at the cost of less compression.

At one point, @tkchia researched exomizer's algorithms a bit more theoretically, perhaps he has some input on this? I would definitely be interested in having a fast 8086 ASM version of our decompressor to help the utility of compressed binaries in ELKS. All the testing routines in the rawdecrs subdirectory could be converted to ELKS applications for testing on ELKS, but outside the kernel, as a start.

ghaerr:
> urlget is now 4 programs. If the floppy generating procedure creates hard links instead of symlinks, we save one disk block per program (except the 'first'), and improve startup speed a tiny little bit. 

Yes, I (re)noticed that as I was writing this PR. I'll add that as a lower priority to the TODO list, its a bit complicated to implement in `mfs` having to do with ensuring the original file is specified before the hard linked files and some other considerations. 


> I haven't put in the creation of ftpput yet btw. 

I already snuck that fix into one of the previous network fix PRs.

> With the ftp-daemon on the way we get yet another daemon to consume RAM - if we choose to start it. This may be a good time to consider a BSD-style inetd, which kicks in telnetd, httpd, ftpd, whatever is in the /etc/services file, as required. It's a very simple program and would - in the general case - save RAM and keep the network daemons available.

Adding more files, like /etc/services, and /etc management daemons probably just makes ELKS bigger needlessly, at this point, since most folks won't run out of RAM running the network stack and a single network application. When it comes to multi-user use, we have exactly one process slot left after running two telnet's. So I'd say probably not needed at this time. A super small `inetd` could be nice if you'd like to write it, but it might be easier to have "net start" options specified in /bootopts for simplicity if this really matters.

Mellvik:
> I haven't put in the creation of ftpput yet btw.
> 
> I already snuck that fix into one of the previous network fix PRs.
> 
> 
Thanks! I hadn't noticed.
> With the ftp-daemon on the way we get yet another daemon to consume RAM - if we choose to start it. This may be a good time to consider a BSD-style inetd, which kicks in telnetd, httpd, ftpd, whatever is in the /etc/services file, as required. It's a very simple program and would - in the general case - save RAM and keep the network daemons available.
> 
> Adding more files, like /etc/services, and /etc management daemons probably just makes ELKS bigger needlessly, at this point, since most folks won't run out of RAM running the network stack and a single network application. When it comes to multi-user use, we have exactly one process slot left after running two telnet's. So I'd say probably not needed at this time. A super small inetd could be nice if you'd like to write it, but it might be easier to have "net start" options specified in /bootopts for simplicity if this really matters.
> 
Good points, @ghaerr. I'll put the idea on the backburner for now. The idea is to free up process slots - in addition to RAM. And for a (really) tiny inetd, the config might as well be in a header file. OTOH a 'net start ftpd', 'net stop ftpd' is actually a very attractive option, /bootopts options too, which brings back the inssue of a bootopts.sample file. How about (and ignoring FAT for now) /bootopts.smple - it should be obvious enough?

Thank you.

—Mellvik

PR #990, opened 2021-11-05
80386 Unreal mode research

ghaerr:
This PR aims to use the `unreal16` program to allow testing the ability to run the ELKS kernel and various applications after having turned on the 80386 processor "unreal" mode, which is an operating mode in 80386 processors and later that allow segment register limits to be expanded from 64K to access all of machine memory in a 32-bit linear fashion, while retaining 16-bit real mode compatibility.

The previously contributed `unreal16` application is cleaned up to be a proper ELKS application (no BIOS calls), and GDT documentation added. 

For testing, run `unreal16`. This should display some black-on-green text indicating that unreal mode is operating properly. Further testing is required by continuing to run ELKS on real hardware using other applications to ensure that the kernel actually functions properly with unreal mode permanently on, as it is left on on unreal16 application exit.

Next, various methods to turn on the A20 memory line will be added (and possibly left on), which should then immediately allow very fast access to any machine memory above 1Mb (high memory). Unreal16 will then be used for reading and writing high memory for testing on different real PCs. After that, a kernel modification to allow external L2 cache buffers above 1Mb to be used, which should be pretty straightforward. This would greatly speed up ELKS operation on 80386 or later processors with extra memory.

From my current research, there are different mechanisms used to turn on/off A20 across PC systems, and the BIOS is the only portable method. Ultimately, the testing learned from running unreal16 as an application will allow moving its code into the ELKS kernel, and turning on unreal mode if the system is configured for high memory buffers and the processor is 80386 or later. It is not clear whether the A20 line can be left on, which would stop auto-wrap of 8086 segments,  which could possibly cause misbehaved programs to operate differently.

The 80286 processor could be supported using the undocumented LOADALL instruction, but that may be more trouble than its worth, instead of the straightforward approach usable with the 80386. That support could come afterwards if desired.

### DIFF
--- a/elkscmd/sys_utils/unreal16.S
+++ b/elkscmd/sys_utils/unreal16.S
@@ -29,7 +29,9 @@
 # 12 Sep 2020:
 # Use `_start' rather than `entry' as program entry point. TK Chia
 #
-# 4 Nov 2021: bug fixes and cleanup ghaerr
+# 4 Nov 2021: ghaerr
+# bug fixes and cleanup
+# add A20 enable and verify (uses A20.ASM code by Chris Giese)
 ##############################################################################
 
 sys_exit = 1
@@ -44,8 +46,7 @@ videoline = 0xb8000+80*2*22
 _start:
 
 # check for 32-bit CPU
-        mov     $needs_386_msg,%dx
-        mov     $needs_386_len,%cx
+        mov     $needs_386_msg,%si
 
         pushf
                 pushf
@@ -63,8 +64,7 @@ _start:
         je msg_and_exit
 
 # check if (32-bit) CPU is in V86 mode
-        mov     $v86_msg_len,%dx
-        mov     $v86_msg,%cx
+        mov     $v86_msg,%si
 
         smsww %bx               # 'SMSW' is a '286+ instruction
         andb $1,%bl
@@ -117,8 +117,8 @@ _start:
 # Action of "rep movsb" in (un)real mode:
 # With a32 prefix byte (0x67):  byte [ES:EDI++] <- [DS:ESI++], ECX times
 # Without prefix byte:          byte [ES:DI++] <- [DS:SI++], CX times
-		cld
-        .byte 0x67 #a32
+        cld
+        .byte 0x67		# addr32
         rep movsb
 
 # or poke byte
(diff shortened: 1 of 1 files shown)


Mellvik:
Very interesting indeed, @ghaerr.

And - I think - something that will be very useful in addition to being a really inserting technical challenge.

Of course I'd like to see the 286 covered, but admittedly that's quite a hurdle to pass. OTOH - getting control of the A20 line would benefit the 286 as well, and make highmem available.

One step at a time. I look forward to play with and get acquainted with unreal16.

Thank you!

—Mellvik

ghaerr:
@Mellvik: I've added A20 gate enable and verification. I'd like you to test this on various 386+ machines you might have. 

After enabling unreal mode, `unreal16` checks the current status of the A20 gate, and if not already enabled, displays the message "A20 gate disabled" and tries to enable it, then displays the final enable/disable status. QEMU seems to start with A20 already enabled, but the code should work for you.

Here's a sample screen shot of proper output on QEMU. Your PC may show "A20 gate disabled" before the final message, which is OK
Screen Shot 2021-11-05 at 2 21 54 PM

If this works, and ELKS runs fine in unreal mode with A20 enabled, we're ready to add the high memory buffers!

Mellvik:
Thank you @ghaerr,
This is exciting. I look forward to dive into this as soon as the ftp server (first cut) is good to go (being tested right now, it ended up as a total rewrite).

Will test on 286 too, might be interesting ...

- Mellvik

ghaerr:
> Will test on 286 too, might be interesting ...

"Unreal" mode requires a 32-bit CPU, which `unreal16` checks for. It should say "Sorry, 80386+ CPU required" on a 80286 or 8086 processor and exit.

While the 80286 introduced a segmented protected mode and the associated "selectors" used instead of physical segment addresses in the segment registers, the method used by unreal mode and that which ELKS will use for high memory buffers requires a 32-bit register index off a segment register (DS or ES) addressing physical (high) memory. While an 80286 "selector" could be used to address high memory, this would involve a complete switch to and from protected mode each time access to high memory is needed. Because the entire ELKS kernel is written for "real" mode and computes segment register values directly using address arithmetic, we can't run the kernel in protected mode, even a 16-bit protected mode, but should be able to run permanently in unreal mode. [EDIT: It is possible the undocumented "LOADALL" 80286 instruction may be able to be used to access high memory using a different technique; this could also be experimented by rewriting portions of unreal16.S by people with access to a real 80286 processor.]

Note that this capability is separate from the A20 line gate, which apparently was added by IBM to "fix" the 80286 CPU's normal behavior of *not* wrapping back to 0 at the end of a segment. Apparently some programs required this and thus all PC hardware defaulted to A20 gate off on startup. On DOS,  HIMEM.SYS turned the A20 gate off after every program termination. In ELKS, we don't have that problem, and plan to leave both the A20 gate ON and the system in "unreal" mode after boot, and use the 0x67 "address 32 override" instruction prefix which allows a 32-bit address register to be used during a 16-bit instruction for high memory access.

A close inspection of the unreal16 ASM code shows exactly how both these functions are implemented. The A20 line gate is using the "classic" method of programming bit 1 of the 8042 keyboard controller, which should work on "most" systems. We will see!

Mellvik:
> ﻿Will test on 286 too, might be interesting ...
> "Unreal" mode requires a 32-bit CPU, which unreal16 checks for. It should say "Sorry, 80386+ CPU required" on a 80286 or 8086 processor and exit.
> 
Yes, that was my point - ensure that architecture detection actually works. It hasn't always been simple. Part of the fun in testing...
> Note that this capability is separate from the A20 line gate, which apparently was added by IBM to "fix" the 80286 CPU's normal behavior of not wrapping back to 0 at the end of a segment. Apparently some programs required this and thus all PC hardware defaulted to A20 gate off on startup. On DOS, HIMEM.SYS turned the A20 gate off after every program termination. In ELKS, we don't have that problem, and plan to leave both the A20 gate ON and the system in "unreal" mode after boot, and use the 0x67 "address 32 override" instruction prefix which allows a 32-bit address register to be used during a 16-bit instruction for high memory access.
> 
Fond (?) DOS memories - and the reason for my A20+286 interest. 
> A close inspection of the unreal16 ASM code shows exactly how both these functions are implemented. The A20 line gate is using the "classic" method of programming bit 1 of the 8042 keyboard controller, which should work on "most" systems. We will see!
> 
Than would make the 64k HMA available on 286 systems - by default, so to speak. For buffers, possibly code. Maybe the unreal16 code could be expanded to look for HMA memory if it determines the processor is not a 386 - or even better, that it is a 286.

One of these days we'll take a close look at the HIMEM source to see how it's using loadall to get access to all 16MB on the 286. Like you pointed out before - a few megs in a RAM drive would make ELKS scream, even on the 286. Next year maybe :-)

Thank you @ghaerr!

--Mellvik 
>

ghaerr:
> Than would make the 64k HMA available on 286 systems - by default, so to speak. For buffers, possibly code. Maybe the unreal16 code could be expanded to look for HMA memory if it determines the processor is not a 386 - or even better, that it is a 286.

Yes - it could do all that. The changes required for supporting HMA (apart from the A20 gate, which unreal16 now handles in the classic 8042 case), are very similar to those required for supporting EMS (expanded memory) discussed in #708. The ELKS memory manager needs to be expanded to support "holes" for both and more than 20 bits for HMA.

My plan is to get XMS (extended) memory access working first for much larger external buffer support on 386 systems, and go from there.

Mellvik:
This is great, @ghaerr -

Don’t know if I can contribute with coding, but I’ll certainly keep up the physical testing!!

Thank you.

—Mellvik

PR #994, opened 2021-11-09
[kernel] Add CONFIG_FS_XMS_BUFFER option to enable kernel unreal mode/A20 gate

ghaerr:
Adds "XMS buffer cache" config option (default ON) which sets CONFIG_FS_XMS_BUFFER. For now, when set, the kernel attempts to enable unreal mode and the A20 gate line during startup. If successful on both, it will display "XMS: unreal mode and A20 enabled". If not, either "XMS: Can't enable unreal mode" or "XMS: Can't enable A20 gate" will be displayed. This should allow easy within the kernel.

Turning off the option will disable any XMS/unreal/A20 support; this is the default for 8018X and ROM based configs.

An additional option, CONFIG_FS_NR_XMS_BUFFERS is also added, which will set the amount of XMS memory to use. It is being ignored until the XMS memory management under construction is operational.

Next step is rewriting the low-level block drivers and buffer routines to use a new `xms_fmemcpyw` routine that will allow copying to and from main/far memory and XMS memory, based on a new `ramdesc_t` physical ram descriptor data type. Since I/O can't be directly performed into XMS buffers, an extra (1K) memory copy will be required from the block device driver's buffer, which will either be DMASEG for the BIOS FD/HD driver, or a local buffer for the SSD driver. The ramdisk driver will need further rewriting in order to work with XMS memory, but could allow much larger ram disks, based on XMS memory size.


Mellvik:
Hi @ghaerr,

This is indeed exciting (and fastmoving). I will not pretend to understand all the nitti-gritty stuff right off the bat, but I'll get there eventually.

The road to understanding will as usual be littered with questions, so here's a faint start:

- You chose to use the XMS 'name' (instead of, say, HMA) because we're really talking about access to all available memory above 1M, not only the HMA (which is the limit on 286 based hw), right?

- XMS memory is useful for buffers and storage, but ELKS cannot run code in this area, even on 32bit hw, right?

- Would it be (technically) possible to DMA to/from XMS memory, thus freeing the kernel to do other stuff while (big) memory transfers are going on? 

- While direct I/O to/from the XMS area is generally not possible, is the HMA area within reach for I/O?

- E.g. The ne2k driver currently uses far memory buffers for IO, can it take advantage of the 64k HMA area in order to expand packet buffering?

More questions as we proceed...

I just ran the most recent commit (24ec88c0) on physical HW (386), but I don't get any XMS messages during boot and unreal16 just hangs (interruptible). I'm guessing something is not right in my config - any ideas?


CONFIG_MINIX_FS=y
# CONFIG_ROMFS_FS is not set
CONFIG_FS_FAT=y
CONFIG_FS_DEV=y
# CONFIG_ROOT_READONLY is not set
# CONFIG_FS_RO is not set
CONFIG_FULL_VFS=y
CONFIG_32BIT_INODES=y
CONFIG_FS_EXTERNAL_BUFFER=y
CONFIG_FS_NR_EXT_BUFFERS=64
CONFIG_FS_XMS_BUFFER=y
CONFIG_FS_NR_XMS_BUFFERS=128
CONFIG_PIPE=y
# CONFIG_EXEC_COMPRESS is not set
CONFIG_EXEC_MMODEL=y

#


Thank you!

—Mellvik

ghaerr:
> This is indeed exciting (and fastmoving). I will not pretend to understand all the nitti-gritty stuff right off the bat, but I'll get there eventually. 

Let's just say I'm learning quickly too... between understanding UMB, EMS, HMA, XMS, unreal mode, A20 gates and processor differences, there's lots of room for error. Unfortunately, there's some hardware variation as well, especially in the A20 gate management, which we might run in to on physical PCs.

> You chose to use the XMS 'name' (instead of, say, HMA) because we're really talking about access to all available memory above 1M, not only the HMA (which is the limit on 286 based hw), right?

Yes, right. The HMA is the 65520 (64K minus 0x10) bytes of memory just over the 1M mark. HMA can be accessed in real mode on any processor with more than 20 bits of address bus by using the segment:offset FFFF:0010 through FFFF:FFFF. The A20 "gate" has to be enabled, which allows the address arithmetic to *not* wrap around back to low core, but instead up into just less than 64K bytes of the "extended" memory. Since the 8086 only has 20 address lines, it won't work, but the 286 and above will. IBM put the A20 gate hardware into the PC AT (286) because some DOS real mode programs *depended* on the wrap back to low memory, and it's been a pain dealing with this ever since! The A20 line defaults to OFF on PC hardware at boot, ON in QEMU.

ELKS could run code in HMA, but that will require an enhancement to the memory allocator, which would also allow UMB blocks to be used for storage or execution as well, as requested in #709 (noncontiguous memory blocks) and possibly #708 (EMS memory support).

 > XMS memory is useful for buffers and storage, but ELKS cannot run code in this area, even on 32bit hw, right? 
 
Correct, outside the HMA. We use the 386 "unreal" mode which trickily uses a 32-bit data pointer indexed off of DS:ESI or ES:EDI to directly access the memory. This only works for the DS or ES segments, not for SS or CS, so code won't work. We could only run code in that area if protected mode were turned on. There is such a thing as "16-bit" protected mode that would allow running up there, but that's pretty much outside the scope of the ELKS kernel.


> Would it be (technically) possible to DMA to/from XMS memory, thus freeing the kernel to do other stuff while (big) memory transfers are going on?

That would depend on the number of address lines on the DMA chip: if there are more than 20, then this could work. ELKS doesn't use any hardware DMA currently, although there is a dma.c source file that might be a place to start. In general, given the single-user activity of ELKS, having lots of asynchronous I/O doesn't help that much, as the application the user is using would be scheduled out, waiting for I/O anyways.

Because BIOS I/O can't go directly into XMS, we're having to do an "extra" memory copy after disk I/O from a low-core single-purpose buffer up to XMS, which is the price to pay for buffering XMS instead of buffering into main memory. Since it's only associated with the actual I/O, which is a lot slower, this should still result in a large speedup for most activities. However, that single-purpose buffer is also the track cache buffer, which works without the extra copy if I/O is within the track cache.

> While direct I/O to/from the XMS area is generally not possible, is the HMA area within reach for I/O? - E.g. The ne2k driver currently uses far memory buffers for IO, can it take advantage of the 64k HMA area in order to expand packet buffering?

Yes, we should be able to use HMA for I/O, provided the processor is 286 or better and the A20 gate is working. This complicates the driver a bit, since it would have to handle both the 286+/A20 enabled and the regular 8086 case, allocating the buffer differently. I plan on looking into kernel HMA support after getting XMS working.


> I just ran the most recent commit on physical HW (386), but I don't get any XMS messages during boot and unreal16 just hangs (interruptible).

Geez. The XMS message is the very first one after the "ELKS Setup t02C0 f0FB6 d12C6" message, it could have scrolled off the screen, it displays something for both success and failure cases. I am guessing that unreal16 is hanging in the A20 gate enabling routine; I'll add more messages to unreal16 for better debugging for you. The A20 gate hardware is using an "extra" pin on the PC keyboard controller (!) and is programmed by sending a control byte to the 8042 keyboard controller and then busy-looping for a response. Adding to the mess is that the direct console uses keyboard interrupts, and the keyboard controller issues IRQ 1 interrupts, although I think not during this operation (interrupts are turned off in unreal16 as a possibility).


>  I'm guessing something is not right in my config - any ideas? 

It looks correct, but try running the kernel without enabling XMS with the following, and then running unreal16 to see what it says, since it's already having problems. Use the following:
```
CONFIG_FS_EXTERNAL_BUFFER=y 
CONFIG_FS_NR_EXT_BUFFERS=64 
# CONFIG_FS_XMS_BUFFER is not set
```
Thanks for your testing! I'll push another commit with more error messages.

Issue #996, opened 2021-11-10
Network socket close bug


Mellvik:
When an application closes a network socket just after having written to it, the socket remains open 'under the hood' for up to 10 seconds. 

This is evidenced by new attempts to open a connection using the same local port number, which gives a `EADDRINUSE` error until some timeout causes a real close.

The test case is the (upcoming) ELKS `ftpd` server and the `MGET` command, which sends a series of files back-to-back back to the client. When programmed to reuse the same port and wait until the `EADDRINUSE` error goes away, it takes between 5 and 10 seconds for that to happen.

If the socket is being read from instead of written to by ELKS, as is the case with the `FTP` `MPUT` command, there is no problem. The socket is being fully closed immediately and can be reopened back-to-back using the same local port number.

The new `ftpd` which turns out to be very helpful provoking the issue, will be posted shortly.

--Mellvik

ghaerr:
> This is evidenced by new attempts to open a connection using the same local port number, which gives a EADDRINUSE error until some timeout causes a real close.

I assume you're testing using an ftp client from outside ELKS to your new FTPD server on ELKS.

Is it the external ftp client that is getting the EADDRINUSE error? Is that on the data connection (not command channel)?

Or is this seen on the ELKS ftp daemon, when opening the data connection out to the controlling ftp client? I am trying to understand exactly which connection is failing.

Mellvik:
Additional comments:
- The timeout before a socket again becomes available is always exactly 10 seconds. The variation mentioned before came from an experiment using 2 different port numbers every other time.
- From the TCP trace, the connection is always properly closed. A FIN which is immediately ACKed.

ghaerr:
> From the TCP trace, the connection is always properly closed. A FIN which is immediately ACKed.

There needs to be two FINs, one sent from each side, each ACKed, in order to fully shut the TCP connection down without going into TIME WAIT state. Are you seeing the second ACK? Not sure yet which direction we're talking about.

ghaerr:
> When programmed to reuse the same port and wait until the EADDRINUSE error goes away

Can you remind me exactly what is supposed to happen on an FTP transfer (from ftpd to ftp)? Does the client specify a port number on its side that the daemon is supposed to connect to? Is that a reserved port, or is there communication as to a new port number every time? I am trying to understand the binding (and reuse) of the port numbers and don't have the FTP protocol in my head, to debug this problem.

Mellvik:
[...]

PR #997, opened 2021-11-10
[net] ftp server for ELKS, supporting basic file transfer functionality.

Mellvik:
`ftpd` implements basic `FTP` file transfer service for ELKS. The server has been rewritten from scratch because everything else found on the net was either too big and complicated or too simple and special purpose oriented.

This is a first cut, which has been tested on physical hardware and generally works. There's a lot of debugging stuff in the code, and a `-d` option has been added to activate ample debug output. The amount of debug output will be reduced later.

In this version, the program does not disconnect from the controlling terminal. Use `&` to background it if desired. This will be fixed later.

Other issues:
- Passive mode is not yet supported.
- Globbing is not support yet. `MGET <directory>` works, `MGET x*` does not.
- Commands supported: `get mget put mput dir user pass type cwd pwd bye` and some other commands not visible to the user.
- `mget` will currently not filter out directories from the list of files to get.
- Login  is supported (required) but nonfunctional in the sens it does not check anything. Access is entirely open.  

The codes has a number of FIXMEs, to be addressed later.


ghaerr:
Wow, you have been busy! Nice work!

Let's see whether the hack suggested in https://github.com/jbruchon/elks/issues/996#issuecomment-965547081 to ktcp for reusing the local port works.

If so, I can add the SO_REUSEADDR option and you can push a fix to this, or would you rather this get committed now?

Can you add elkscmd/inet/ftp/Makefile and other changes so this builds with the make system?

Mellvik:
> 
> Let's see whether the hack suggested in #996 (comment) to ktcp for reusing the local port works.
> 
First thing in the morning, thank for a very efficient an productive session.
> If so, I can add the SO_REUSEADDR option and you can push a fix to this, or would you rather this get committed now?
> 
> Can you add elkscmd/inet/ftp/Makefile and other changes so this builds with the make system?
> 
Will update the PR tomorrow, thank you.

M

Mellvik:
Added Makefile, also added a minor fix to `ls`: When outputting to a non-tty, `ls` must deliver clean file names, not space padding. This fix removed the padding.

The socket address reuse problem is still pending.

Mellvik:
Comment about the `ls` fix:
`ls` is frequently used in scripts (via pipes) and other contexts to provide lists of files, one per line. In such contexts it is always assumed that the file list is clean. i.e. no beginning or trailing spaces, no extraneous characters except newlines.

ELKS `ls` added extra blanks to file names in such contexts, thus making its output unusable in scripts (and in contexts such as `ftpd` where I decided to ditch it in favor of a much faster `opendir()/readdir()` ).

This minimal change fixes the problem.

ghaerr:
@Mellvik, are you OK with me committing this now?

> ELKS ls added extra blanks to file names in such contexts, thus making its output unusable in scripts (and in contexts such as ftpd 

Got it, sounds good.

I would like to test some of the issues mentioned in #999, as well as commit #1000, to help solve the remaining out of memory problem(s).

PR #995, opened 2021-11-09
[kernel] Add XMS memory support for kernel buffers

ghaerr:
This is a first-draft set of commits implementing the ability to use memory above 1MB for kernel system buffers, freeing up space from main memory for application programs. Requires 386 or later and compatible A20 gate handling.

Tested and working on QEMU, with up to 256K in XMS buffers allocated.

More work is in process, before commit: SSD and RAM disk drivers need to be made compatible with changes, they won't work with XMS in this PR. Also will add debug output to `unreal16` for real hardware testing.

Unfortunately, we're running into a few problems supporting a large (e.g. > 256K) amount of kernel buffers in XMS memory. The reason is that the "buffer headers", a data structure header that points to the actual XMS buffer, must reside in the kernel data segment. Each buffer header is 26 bytes, which means that 26K of data would be used just to track 1MB of XMS buffers; we only have 64K total and there's only 8K to spare with 128K of XMS buffers. Not sure how to work this out, but still saves 64-128K+ of main memory on 386 PCs.

PR #998, opened 2021-11-10
[kernel] Add additional A20 gate handler and debug options to unreal16

ghaerr:
Adds some additional options for the problem discussed in #995.

After seeing the output from this version of `unreal16`, if the A20 gate is still not enabled, another option was added, which uses the (non-OEM) HIMEM.SYS method for enabling A20. 

This is done in elks/arch/i186/lib/unreal.S by uncommenting the 2nd option:
```
# configurable options for A20 gate
#   USE_A20ASM      - use Chris Giese A20.ASM method (send D0, read, OR 2, D1, write)
#   USE_HIMEM_AT    - use himem.sys PC AT method (send D1,DF,FF)
#define USE_A20ASM
//#define USE_HIMEM_AT
```

Also, XMS buffering is turned off by default, to allow kernel boot normally unless specifically testing XMS (set by make menuconfig).


Mellvik:
@ghaerr,

I tested this one just now,
it says "xms: can't enable A20 gate".

Then boots fine.

The unrea16 program still loops (interruptible).

As to COMPAQ 386 peculiarities, there are some - although not likely to influence this particular situation:

Compaq remaps the physical memory from 640k-1M to the top of the 16MB memory area, where the top 128K are reserved for system use (actually, it copies the BIOS to this area and runs it from there, thus making it available in protected mode), the rest is available as extended memory (15.6M-15.9M).
There are no software or hardware configurations that mentions A20

—Mellvik

ghaerr:
>  I tested this one just now, it says "xms: can't enable A20 gate". Then boots fine. 

This means that unreal mode has been enabled, but that the memory test after trying to have enabled A20 failed.
I added another A20 gate routine, the one used by non-OEM HIMEM.SYS. Can you try it by setting the following lines in elks/arch/i86/lib/unreal.S:
```
# configurable options for A20 gate
#   USE_A20ASM      - use Chris Giese A20.ASM method (send D0, read, OR 2, D1, write)
#   USE_HIMEM_AT    - use himem.sys PC AT method (send D1,DF,FF)
//#define USE_A20ASM
#define USE_HIMEM_AT <-- we want to have this uncommented to enable the HIMEM.SYS code for PC AT
```

> The unrea16 program still loops (interruptible). 

The program writes directly to the direct console, as well as to the current tty. Are you running on serial? Can you check to see exactly what is displayed on the direct console also? 

Is there anything displayed on the serial line after running unreal16? Or is it hanging directly after running it, and then you ^C out.

Sorry for the trouble, PC incompatibilities in this area abound, unfortunately.


> As to COMPAQ 386 peculiarities, there are some - although not likely to influence this particular situation: Compaq remaps the physical memory from 640k-1M to the top of the 16MB memory area, where the top 128K are reserved for system use (actually, it copies the BIOS to this area and runs it from there, thus making it available in protected mode), the rest is available as extended memory (15.6M-15.9M). There are no software or hardware configurations that mentions A20 —

Are you saying that the Compaq BIOS copies itself at boot up to high memory? Geez, that could mean that the BIOS itself is involved turning on and off A20, or that it is already turned on for special extended memory BIOS access.

I'll read up more on the Compaq 386, but I'm guessing that it's special memory handling has a lot to do with the issues here, since the ELKS kernel itself now handles "linear" memory pointers pretty well - it just needs A20 access to use it!

It would be interesting to test this on other machines. I will have access to my 386 again this weekend and can start physical testing there as well.

ghaerr:
> Can you try it by setting the following lines in elks/arch/i86/lib/unreal.S:

I just realized there is another A20 gate option for "newer" machines, that is, using the BIOS itself to enable it. I'll push another commit with that code so you can test both options, which might be easier for you. Thanks for your testing!

Mellvik:
@ghaerr - 
again I thing we're iterating slowly towards something that actually woks :-).
> 
> I tested this one just now, it says "xms: can't enable A20 gate". Then boots fine.
> 
> This means that unreal mode has been enabled, but that the memory test after trying to have enabled A20 failed.
> I added another A20 gate routine, the one used by non-OEM HIMEM.SYS. Can you try it by setting the following lines in elks/arch/i86/lib/unreal.S:
> 
> # configurable options for A20 gate
> #   USE_A20ASM      - use Chris Giese A20.ASM method (send D0, read, OR 2, D1, write)
> #   USE_HIMEM_AT    - use himem.sys PC AT method (send D1,DF,FF)
> //#define USE_A20ASM
> #define USE_HIMEM_AT <-- we want to have this uncommented to enable the HIMEM.SYS code for PC AT
This is the one I tested. The previous (ASM) was the one that hung the system, right?
> The unrea16 program still loops (interruptible).
> 
> The program writes directly to the direct console, as well as to the current tty. Are you running on serial? Can you check to see exactly what is displayed on the direct console also?
> 
When testing this stuff, I use the console.
> Is there anything displayed on the serial line after running unreal16? Or is it hanging directly after running it, and then you ^C out.
> 
well, yes - and I guess I didn't mention that. It says:
Trying to enable unreal mode...
> Sorry for the trouble, PC incompatibilities in this area abound, unfortunately.
> 
> As to COMPAQ 386 peculiarities, there are some - although not likely to influence this particular situation: Compaq remaps the physical memory from 640k-1M to the top of the 16MB memory area, where the top 128K are reserved for system use (actually, it copies the BIOS to this area and runs it from there, thus making it available in protected mode), the rest is available as extended memory (15.6M-15.9M). There are no software or hardware configurations that mentions A20 —
> 
> Are you saying that the Compaq BIOS copies itself at boot up to high memory? Geez, that could mean that the BIOS itself is involved turning on and off A20, or that it is already turned on for special extended memory BIOS access.
> 
Yes, that may very well the the case. Actually it's hard to imagine this setup being compatible with old software that depends on the  (in)famous wraparound. OTOH - maybe it just isn't compatible …

FWIW the setup seems compatible with whatever (newer) DOS versions  do - but it might be a point to test that, see what the DOS mem command and the like says about the HW.
> I'll read up more on the Compaq 386, but I'm guessing that it's special memory handling has a lot to do with the issues here, since the ELKS kernel itself now handles "linear" memory pointers pretty well - it just needs A20 access to use it!
> 
> It would be interesting to test this on other machines. I will have access to my 386 again this weekend and can start physical testing there as well.
> 
> 
Again - we're getting there :-)

Thank you!

-M

PR #999, opened 2021-11-10
[kernel] Add SO_REUSEADDR setsockopt option for servers

ghaerr:
Adds the ability to reuse local port numbers in `bind`. No other checking is performed when requested, so this option should be used carefully, and only with servers.

Cleaned up some messy naming confusing between public SO_xxx symbols and internal flags. The internal flags were renamed SF_xxx.

@Mellvik, this should fix the problem being discussed in #996 and #997.
The following code should be added to ftpd before calling bind:
```
int enable = 1;
setsockopt(fd, SO_REUSEADDR, &enable, sizeof(int));
```

### DIFF
--- a/elks/include/linuxmt/net.h
+++ b/elks/include/linuxmt/net.h
@@ -67,11 +67,13 @@ struct proto_ops {
     int (*fcntl) ();
 };
 
-#define SO_CLOSING	(1 << 0)
-#define SO_ACCEPTCON	(1 << 1)
-#define SO_WAITDATA	(1 << 2)
-#define SO_NOSPACE	(1 << 3)
-#define SO_RST_ON_CLOSE	(1 << 4)
+/* careful: option names are close to public SO_ options in socket.h */
+#define SF_CLOSING	(1 << 0)
+#define SF_ACCEPTCON	(1 << 1)
+#define SF_WAITDATA	(1 << 2)
+#define SF_NOSPACE	(1 << 3)
+#define SF_RST_ON_CLOSE	(1 << 4)
+#define SF_REUSE_ADDR	(1 << 5)
 
 struct net_proto {
     char *name;			/* Protocol name */
--- a/elks/include/linuxmt/socket.h
+++ b/elks/include/linuxmt/socket.h
@@ -15,7 +15,8 @@ struct sockaddr {
 /* for setsockopt(2) */
 #define SOL_SOCKET	1
 
-/* careful: option name scheme interferes with internal SO_ options in net.h*/
+/* careful: option names are close to internal SF_ options in net.h*/
+#define SO_REUSEADDR	2
 #define SO_LINGER	13		/* only implemented for l_linger = 0*/
 
 struct linger {
--- a/elks/include/linuxmt/tcpdev.h
+++ b/elks/include/linuxmt/tcpdev.h
@@ -29,7 +29,7 @@
 struct tdb_release {
     unsigned char cmd;
     struct socket *sock;
-	int reset;
+    int reset;
 };
 
 struct tdb_accept {
@@ -48,6 +48,7 @@ struct tdb_listen {
 struct tdb_bind {
     unsigned char cmd;
     struct socket *sock;
+    int reuseaddr;
     struct sockaddr_in addr;
 };
 
--- a/elks/net/ipv4/af_inet.c
+++ b/elks/net/ipv4/af_inet.c
@@ -49,7 +49,7 @@ int inet_process_tcpdev(register char *buf, int len)
         tcpdev_clear_data_avail();
 	debug_net("INET(%d) chg_state sock %x %d\n", current->pid, sock, sock->state);
 	if (sock->state == SS_DISCONNECTING) {
-	    sock->flags |= SO_CLOSING;
+	    sock->flags |= SF_CLOSING;
 	    wake_up(sock->wait);
 	}
         break;
@@ -100,7 +100,7 @@ static int inet_release(struct socket *sock, struct socket *peer)
     cmd = (struct tdb_release *)get_tdout_buf();
     cmd->cmd = TDC_RELEASE;
     cmd->sock = sock;
-    cmd->reset = sock->flags & SO_RST_ON_CLOSE;
+    cmd->reset = sock->flags & SF_RST_ON_CLOSE;
     ret = tcpdev_inetwrite(cmd, sizeof(struct tdb_release));
     return (ret >= 0 ? 0 : ret);
 }
@@ -121,6 +121,7 @@ static int inet_bind(register struct socket *sock, struct sockaddr *addr,
     cmd = (struct tdb_bind *)get_tdout_buf();
     cmd->cmd = TDC_BIND;
     cmd->sock = sock;
+	cmd->reuseaddr = sock->flags & SF_REUSE_ADDR;
     memcpy_fromfs(&cmd->addr, addr, sockaddr_len);
 
     tcpdev_inetwrite(cmd, sizeof(struct tdb_bind));
@@ -222,9 +223,9 @@ static int inet_accept(register struct socket *sock,
 
     /* Sleep until tcpdev has news */
     while (bufin_sem == 0) {
-        //sock->flags |= SO_WAITDATA;
+        //sock->flags |= SF_WAITDATA;
         interruptible_sleep_on(sock->wait);
-        //sock->flags &= ~SO_WAITDATA;
+        //sock->flags &= ~SF_WAITDATA;
         if (current->signal) {
 	    debug_net("INET(%d) accept RESTARTSYS\n", current->pid);
             return -ERESTARTSYS;
@@ -257,7 +258,7 @@ static int inet_read(register struct socket *sock, char *ubuf, int size,
     /* ensure read blocks until data - wait for ktcp to report data available*/
     while (sock->avail_data == 0) {
 	/* return EOF on socket remote closed*/
-	if (sock->flags & SO_CLOSING)
+	if (sock->flags & SF_CLOSING)
 	    return 0;
 	debug_net("INET(%d) read waiting on sock->avail_data sock %x\n", current->pid, sock);
 	interruptible_sleep_on(sock->wait);
--- a/elks/net/ipv4/af_inet.c
+++ b/elks/net/ipv4/af_inet.c
@@ -117,11 +117,12 @@ static int inet_bind(register struct socket *sock, struct sockaddr *addr,
 
     /* TODO : Check if the user has permision to bind the port */
 
-	down(&rwlock);
+    down(&rwlock);
     cmd = (struct tdb_bind *)get_tdout_buf();
     cmd->cmd = TDC_BIND;
     cmd->sock = sock;
-	cmd->reuseaddr = sock->flags & SF_REUSE_ADDR;
+
+    cmd->reuseaddr = sock->flags & SF_REUSE_ADDR;
     memcpy_fromfs(&cmd->addr, addr, sockaddr_len);
 
     tcpdev_inetwrite(cmd, sizeof(struct tdb_bind));
@@ -132,7 +133,7 @@ static int inet_bind(register struct socket *sock, struct sockaddr *addr,
 
     ret = ((struct tdb_return_data *)tdin_buf)->ret_value;
     tcpdev_clear_data_avail();
-	up(&rwlock);
+    up(&rwlock);
 
     return (ret >= 0 ? 0 : ret);
 }
--- a/elks/net/nano/af_nano.c
+++ b/elks/net/nano/af_nano.c
@@ -247,9 +247,9 @@ static int nano_accept(register struct socket *sock,
 	if (flags & O_NONBLOCK)
 	    return -EAGAIN;
 
-	sock->flags |= SO_WAITDATA;
+	sock->flags |= SF_WAITDATA;
 	interruptible_sleep_on(sock->wait);
-	sock->flags &= ~SO_WAITDATA;
+	sock->flags &= ~SF_WAITDATA;
 
 	if (current->signal /* & ~current->blocked */ )
 	    return -ERESTARTSYS;
@@ -316,9 +316,9 @@ static int nano_read(register struct socket *sock,
 	if (nonblock)
 	    return -EAGAIN;
 
-	sock->flags |= SO_WAITDATA;
+	sock->flags |= SF_WAITDATA;
 	interruptible_sleep_on(sock->wait);
-	sock->flags &= ~SO_WAITDATA;
+	sock->flags &= ~SF_WAITDATA;
 
 	if (current->signal /* & ~current->blocked */ )
 	    return -ERESTARTSYS;
@@ -386,11 +386,11 @@ static int nano_write(register struct socket *sock,
     pupd = NA_DATA(sock)->npd_peerupd;	/* safer than sock->conn */
 
     while (!(space = NA_BUF_SPACE(pupd))) {
-	sock->flags |= SO_NOSPACE;
+	sock->flags |= SF_NOSPACE;
 	if (nonblock)
 	    return -EAGAIN;
 
-	sock->flags &= ~SO_NOSPACE;
+	sock->flags &= ~SF_NOSPACE;
 	interruptible_sleep_on(sock->wait);
 
 	if (current->signal /* & ~current->blocked */ )
@@ -465,7 +465,7 @@ static int nano_select(register struct socket *sock,
      *      Handle server sockets specially.
      */
 
-    if (sock->flags & SO_ACCEPTCON) {
+    if (sock->flags & SF_ACCEPTCON) {
 	if (sel_type == SEL_IN) {
 	    if (sock->iconn)
 		return 1;
--- a/elks/net/socket.c
+++ b/elks/net/socket.c
@@ -142,7 +142,7 @@ static size_t sock_read(struct inode *inode, struct file *file,
 	return -EBADF;
     }
 
-    if (sock->flags & SO_ACCEPTCON)
+    if (sock->flags & SF_ACCEPTCON)
 	return -EINVAL;
 
     if (size == 0)
@@ -165,7 +165,7 @@ static size_t sock_write(struct inode *inode, struct file *file,
 	return -EBADF;
     }
 
-    if (sock->flags & SO_ACCEPTCON)
+    if (sock->flags & SF_ACCEPTCON)
 	return -EINVAL;
 
     if (size == 0)
@@ -200,7 +200,7 @@ int sock_awaitconn(register struct socket *mysock,
     /*
      *      We must be listening
      */
-    if (!(servsock->flags & SO_ACCEPTCON))
+    if (!(servsock->flags & SF_ACCEPTCON))
 	return -EINVAL;
 
     /*
@@ -408,7 +408,7 @@ int sys_listen(int fd, int backlog)
     if (ops && ops->listen)
 	ops->listen(sock, backlog);
 
-    sock->flags |= SO_ACCEPTCON;
+    sock->flags |= SF_ACCEPTCON;
     return 0;
 
 }
@@ -426,7 +426,7 @@ int sys_accept(int fd, struct sockaddr *upeer_sockaddr, int *upeer_addrlen)
     if (sock->state != SS_UNCONNECTED)
 	return -EINVAL;
 
-    if (!(sock->flags & SO_ACCEPTCON))
+    if (!(sock->flags & SF_ACCEPTCON))
 	return -EINVAL;
 
     if (!(newsock = sock_alloc())) {
@@ -571,23 +571,41 @@ int sys_setsockopt(int fd, int level, int option_name, void *option_value,
 	unsigned int option_len)
 {
     register struct socket *sock;
-    int err;
+    int flags;
+    int setoption = 0;
     struct linger l;
 
     if (!(sock = sockfd_lookup(fd, NULL)))
 	return -ENOTSOCK;
 
-    if (option_name != SO_LINGER || option_len != sizeof(struct linger))
+    flags = check_addr_to_kernel(option_value, option_len);
+    if (flags < 0)
+	return flags;
+
+    switch (option_name) {
+    case SO_LINGER:
+	if (option_len != sizeof(struct linger))
+	    return -EINVAL;
+	memcpy_fromfs(&l, option_value, sizeof(struct linger));
+	flags = SF_RST_ON_CLOSE;
+	if (l.l_onoff != 0 && l.l_linger == 0)
+	    setoption = 1;
+	break;
+
+    case SO_REUSEADDR:
+	if (option_len != sizeof(int))
+	    return -EINVAL;
+	memcpy_fromfs(&setoption, option_value, sizeof(int));
+	flags = SF_REUSE_ADDR;
+	break;
+
+    default:
 	return -EINVAL;
+    }
 
-    err = check_addr_to_kernel(option_value, sizeof(struct linger));
-    if (err < 0)
-	return err;
-    memcpy_fromfs(&l, option_value, sizeof(struct linger));
-
-    if (l.l_onoff != 0 && l.l_linger == 0)
-	sock->flags |= SO_RST_ON_CLOSE;
-    else sock->flags &= ~SO_RST_ON_CLOSE;
+    if (setoption)
+	sock->flags |= setoption;
+    else sock->flags &= ~setoption;
 
     return 0;
 }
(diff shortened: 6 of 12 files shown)


ghaerr:
@Mellvik, 

I forgot to mention that although I think this will work, that is, allow the `bind` to proceed, ktcp will still likely have the local previously-bound closed TCP control block in a TIME_WAIT state for 10 seconds, using up valuable heap space.

If this commit allows ftpd to transfer a few files quickly, but then ktcp runs out of heap space, perhaps the SO_REUSEADDR can check for if the previous socket is in the TIME_WAIT state and remove it. This may not be completely per spec, but would be useful for small-memory systems like ELKS, and only decreases security a small amount. (The TIME_WAIT is meant to guarantee that no other "lost" packets slowly come in after a connection is closed down and interfere with a newly-created connection on the same port).

I'll push another commit (commented out) that does just that, so you can try it out should the rest work.

Thank you!

ghaerr:
@Mellvik, I've pushed a commit that should just work; I ended up getting the issue to repeat using `telnetd` and was able to see what's happening.

Also added SO_REUSEADDR to telnetd and httpd. For now, ktcp will automatically remove a previous TCP control block that is in the TIME_WAIT state, if SO_REUSEADDR is being set. This hopefully will completely solve the ktcp excess heap problem with ftpd as well.

Test after adding setsockopt w/SO_REUSEADDR to ftpd.

Mellvik:
@ghaerr,
I tested this using the last committed code (and not #1000) - with some improvement, but there is still a memory allocation issue:
```
cmd: 'TYPE I' 13
cmd: 'PORT 10,0,2,2,132,209' 6
client_ip: 10.0.2.2 client_port: 34001
tcp: port 20 reused, other socket NOT in time_wait!

cmd: 'RETR ftpd' 2
RETR: <ftpd>
cmd: 'PORT 10,0,2,2,171,179' 6
client_ip: 10.0.2.2 client_port: 43955
tcp: port 20 reused, other socket NOT in time_wait!

cmd: 'RETR x1' 2
RETR: <x1>
cmd: 'PORT 10,0,2,2,185,23' 6
client_ip: 10.0.2.2 client_port: 47383
sys_brk(11) fail: brk da12 over by 1556 bytes
sys_brk(11) fail: brk da16 over by 1560 bytes
sys_brk(11) fail: brk d516 over by 280 bytes
sys_brk(11) fail: brk d518 over by 282 bytes
sys_brk(11) fail: brk d51a over by 284 bytes
Bind: Could not connect on port 20
bind error: Out of memory 
PORT command failed.
cmd: 'PORT 10,0,2,2,153,33' 6
client_ip: 10.0.2.2 client_port: 39201
sys_brk(11) fail: brk da12 over by 1556 bytes
sys_brk(11) fail: brk da16 over by 1560 bytes
sys_brk(11) fail: brk d516 over by 280 bytes
sys_brk(11) fail: brk d518 over by 282 bytes
sys_brk(11) fail: brk d51a over by 284 bytes
Bind: Could not connect on port 20
bind error: Out of memory 
PORT command failed.
cmd: 'PORT 10,0,2,2,175,185' 6
client_ip: 10.0.2.2 client_port: 44985
sys_brk(11) fail: brk da12 over by 1556 bytes
sys_brk(11) fail: brk da16 over by 1560 bytes
sys_brk(11) fail: brk d516 over by 280 bytes
sys_brk(11) fail: brk d518 over by 282 bytes
sys_brk(11) fail: brk d51a over by 284 bytes
Bind: Could not connect on port 20
bind error: Out of memory 
PORT command failed.
cmd: 'PORT 10,0,2,2,176,223' 6
client_ip: 10.0.2.2 client_port: 45279
sys_brk(11) fail: brk da12 over by 1556 bytes
sys_brk(11) fail: brk da16 over by 1560 bytes
sys_brk(11) fail: brk d516 over by 280 bytes
sys_brk(11) fail: brk d518 over by 282 bytes
sys_brk(11) fail: brk d51a over by 284 bytes
Bind: Could not connect on port 20
bind error: Out of memory 
PORT command failed.
```

The code btw, now looks like this:
```
        if ( (fd = socket(AF_INET, SOCK_STREAM, 0)) < 0){
                perror("socket error");
                return -1;
        }
        if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &on, sizeof (on)) < 0) {
                /* This should not happen */
                printf("Data channel socket error: REUSEADDR, closing channel.\n");
                close(fd);
                return -1;
        }

```
Awaiting your input ...

--Mellvik

ghaerr:
Hello @Mellvik! Thank you for your testing. Believe it or not, we're getting there...

> but there is still a memory allocation issue

Did you try this with ktcp recompiled to use max heap using `-maout=0xffff` ? That should help, but I think it'll still eventually run out again; ftpd uses lots of sockets!

I think the new code (for both reusing sockets and your ftpd server) are written properly and not buggy. I am thinking that ftpd just happens to be doing two things which end up using lots of memory: it is opening up lots of sockets, and it's doing it quickly. See below for what I think is happening:


```
 cmd: 'TYPE I' 13
 cmd: 'PORT 10,0,2,2,132,209' 6
 client_ip: 10.0.2.2 client_port: 34001
 tcp: port 20 reused, other socket NOT in time_wait!
```
It appears that ftpd is using a new socket not just for data transfer, but for getting transfer management information, etc, from the other side. Each socket of those sockets will use 4K memory (even with the reduced ktcp memory PR #1000).

The last line above indicates that SO_REUSEADDR code is working, that is, the local port is being reused, but that the "previous" socket data can't be deallocated, since it's not in the fully closed state (TIME_WAIT). I think what's happening is that ftpd is running so quickly that, even though it executed a previous `close`, the TCP data FIN/ACKs have not been fully processed under the hood. Because of that, it can't deallocate them at this moment of using another data socket.

Possible fixes could be to 1) delay slightly between files (only files, not every data channel), 2) increase ktcp max heap, 3) add another socket option (SO_RCVBUF) to set a smaller receive buffer size on certain sockets where there won't be much data transfer.

I think I'll commit ftpd so that I can play with it to develop these options.
```
 'PORT 10,0,2,2,153,33' 6
 client_ip: 10.0.2.2 client_port: 39201
 sys_brk(11) fail: brk da12 over by 1556 bytes
 ```

This is ktcp running out of heap.

**Comment by Mellvik** [comment]
Indeed @ghaerr, we're getting there. 

As I understand more of the ktcp intricacies and you become familiar with the ftp peculiarities, the sum of knowledge will eventually lead to something that not only works, but works right.

Now, here's the thing with FTP: It never ever uses more than 2 sockets.
The command channel stays open for the duration of the session, and as we've seen, the data connections come and go, some times rapidly.

And while we've made some progress, I've got this feeling we're missing something, barking up the wrong tree, or - at least - not seeing part of the problem.

So let me try this analysis on you:
Back-to-back high speed transfers work fine when ELKS is the recipient, not when ELKS is the sender.
The data channels are opened by the same code in exactly the same way in both cases.
When an incoming data transfer is complete, the external client closes the connection by sending a FIN.
ELKS ACKs this FIN properly and immediately, but does not send its own FIN with the ACK (which I figured it should, but it doesn't seem to matter).
Having received and ACKed the FIN, everything is clean and clear, ktcp closes and deallocates resources and is immediately ready for a new connection. Using the same socket or a different one - the point being that the previous connection is gone and nothing is lingering.
When an outgoing data transfer is complete, it is ftpd who initiated the close. ktcp sends the FIN, gets a FIN & ACK back immediately - but in this case ktcp decides to handle the close differently, delaying the deallocation of resources, and I cannot figure out why. I.e. why one case works and the other does not. Having sent and received FINs mean there is nothing more to say and no reason for the 10 second (or any) wait – which IIRC is there to handle lingering data should the FINs not be there.

This logic - if correct - indicates that we're really looking for a bug preventing immediate shutdown when the close is actually clean. Thus I don't think - at this time - curing symptoms via ftpd delays is the way to go.

I'm all ears.

Thank you.

—Mellvik

**Comment by ghaerr** [comment]
> Back-to-back high speed transfers work fine when ELKS is the recipient, not when ELKS is the sender.

They actually work fine the other direction too, we're just running out of memory, correct?

>  but in this case ktcp decides to handle the close differently, delaying the deallocation of resources, and I cannot figure out why. I.e. why one case works and the other does not.

There is one part of the TCP state machine that differs between the sending and receiving cases - that of what happens after the socket is closed. This has to do with the fact that after each FIN is received, a final ACK is sent. That ACK can be lost in one of the cases, so the state machine demands a TIME_WAIT state. That TIME_WAIT state keeps the control block allocated for 10 seconds, which is causing the memory shortage.

Here's the [state diagram](https://www.fishercom.xyz/digital-transmission/tcp-state-transition-diagram.html):
8664_481_279-penjelasan-tcp-state-transition-diagram

Notice that, on the client close, versus the server close, a TIME_WAIT state is entered, whereas not in the other direction.

> This logic - if correct - indicates that we're really looking for a bug preventing immediate shutdown when the close is actually clean. 

So, a "clean" close still requires a TIME_WAIT in one direction. PR #1000 is going to help save memory, but the full 4K buffer is still allocated during the wait state. Another solution may be to seperate the buffer from the control block, and reduce it's size greatly when in TIME_WAIT, thus freeing substantial memory. I'll look into that. Another thought is to use much smaller buffers for the TCP sockets ftpd uses that are not actually transferring large data file contents, but only commands or directory listings, for instance. So we're likely going to have to get meticulous on memory usage and enhance our TCP API to allocate less when possible.

> ELKS ACKs this FIN properly and immediately, but does not send its own FIN with the ACK (which I figured it should, but it doesn't seem to matter).

Yes, ktcp is pretty simple, and sends multiple packets instead of combining them, since the packets are created directly from its internal state machine, rather than buffering possible output and combining flags in this case.

**Comment by Mellvik** [comment]
@ghaerr, I'm still digesting this, but two quick point off the top of ny head:

> 
> Back-to-back high speed transfers work fine when ELKS is the recipient, not when ELKS is the sender.
> 
> They actually work fine the other direction too, we're just running out of memory, correct?
> 
> 
No, the out of memory condition is a consequence of the delayed release of resources, not the actual problem.
> but in this case ktcp decides to handle the close differently, delaying the deallocation of resources, and I cannot figure out why. I.e. why one case works and the other does not.
> 
> There is one part of the TCP state machine that differs between the sending and receiving cases - that of what happens after the socket is closed. This has to do with the fact that after each FIN is received, a final ACK is sent. That ACK can be lost in one of the cases, so the state machine demands a TIME_WAIT state. That TIME_WAIT state keeps the control block allocated for 10 seconds, which is causing the memory shortage.
> 
> 
Yes, and the way I read it (still digesting) – the final ACK has been received so there is nothing to wait for.

I'll pass on some possibly enlightening output for the sometimes incredibly useful  ^P debugger 

—M

**Comment by ghaerr** [comment]
Another idea for saving memory could be a socket option to reduce the TIME_WAIT value significantly, say to 1/2 second. This would probably help quite a bit.

To understand the ramifications of that, consider (again) that the reason the TIME_WAIT state is necessary, is that, if the final ACK sent (after the double FIN handshake) is lost, there is no way for the sender (in this particular case) to know, since the sender is shutting down (and going into TIME_WAIT state).  Should the receiver resend a FIN, because it never got the final ACK, and the TIME_WAIT state not be happening, that could interfere with a new connection to the same port number, since the local port is the same.

Admittedly, this is a small worry for most ELKS connections, especially on LANs and not on the internet.

In addition to a per-socket TIME_WAIT setting, I'm also thinking about adding the ability to change the global setting. We just want to make sure we have TCP mostly working before changing timeouts and adding more fuel to a fire!

**Comment by Mellvik** [comment]
Some additional info, @ghaerr ghaerr - 
using the magic `kctp debugger` you built a while back, fist from a transfer INTO ELKS:
```
cmd: 'PORT 10,0,2,2,148,71' 6
client_ip: 10.0.2.2 client_port: 37959

tcp: rtt 0 RTT 16 RTO 32
tcp retrans: remove seq 0 unack 0
TS_SYN_SENT
TS_ESTABLISHED
tcp write: seq 2181 size 30 rcvwnd 29200 unack 0 (cnt 0, mem 0)
tcp retrans: check seq 2211 unack 30 time 20
tcp: rtt 1 RTT 9 RTO 20
tcp retrans: remove seq 2181 unack 0
cmd: 'STOR x3' 3
tcp write: seq 2211 size 46 rcvwnd 29200 unack 0 (cnt 0, mem 0)
tcp retrans: check seq 2257 unack 46 time 18
tcp retrans: check seq 2257 unack 46 time 17
tcp retrans: check seq 2257 unack 46 time 17
tcp retrans: check seq 2257 unack 46 time 17
tcp: rtt 2 RTT 8 RTO 18
tcp retrans: remove seq 2211 unack 0
tcp write: seq 2257 size 24 rcvwnd 29200 unack 0 (cnt 0, mem 0)
tcp retrans: check seq 2281 unack 24 time 16
tcpdev release: close socket aae2, tcb 952a
tcp: rtt 0 RTT 16 RTO 32
tcp retrans: remove seq 1 unack 1
tcp: rtt 1 RTT 7 RTO 16
tcp retrans: remove seq 2257 unack 0
tcp: time_wait 0 user_timeout 1
tcp: expire check LAST_ACK (-1.3)
TS_LAST_ACK
tcp: REMOVING control block 952a
```
And here's similar output from an OUTGOING - from ELKS - transfer
```
tcp: rtt 1 RTT 1 RTO 4
tcp retrans: remove seq 10753 unack 512
tcp write: seq 11777 size 260 rcvwnd 52736 unack 0 (cnt 1, mem 532)
tcp retrans: check seq 12037 unack 260 time 4
tcp: rtt 2 RTT 1 RTO 4
tcp retrans: remove seq 11265 unack 260
tcp write: seq 2157 size 24 rcvwnd 29200 unack 0 (cnt 1, mem 280)
tcp retrans: check seq 2181 unack 24 time 24
tcp: rtt 1 RTT 1 RTO 4
tcp retrans: remove seq 11777 unack 0
tcpdev release: close socket aa96, tcb 952a
tcp: rtt 0 RTT 1 RTO 4
tcp retrans: remove seq 12037 unack 1
tcp: rtt 1 RTT 10 RTO 24
tcp retrans: remove seq 2157 unack 0
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_1 (-1.3)
TS_FIN_WAIT_1
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-1.4)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-3.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-4.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-5.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-6.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-7.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-8.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-9.2)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-10.2)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-11.2)
tcp: REMOVING control block 952a
```
It seems to me we're going into FINWAIT even though the FIN has been sent, ACKed and received.

What do you see?

-M

ghaerr:
> It seems to me we're going into FINWAIT even though the FIN has been sent, ACKed and received.

Notice the difference between the two final states: one is LAST_ACK and the other is FIN_WAIT_2. Look at the state diagram, the FIN_WAIT_2 sends a final ACK, then goes into TIME_WAIT for the reason I gave above. That final ACK could be lost. TCP requires the TIME_WAIT, and ktcp is operating per spec.  The wait is not for another ACK, the wait is because we sent a final ACK that could be lost.

The two different cases are happen because of the difference between which side does the close first. I don't think this is buggy, but we can lower the TIME_WAIT to help, given that we're unlikely to have lost packets.

ghaerr:
Reading the state diagram, we only have to implement TIME_WAIT for 2 \* MSL (twice the max segment life).

For the internet, this value is supposed to be 2 minutes!! I lowered it to 10 seconds a long time ago. Perhaps an setup/bootopts/net start option to specify that LAN use, which would allow lowering MSL to msecs. This would prevent ktcp from keeping them around for very long at all.

We need to adhere to the spec, but probably need to consider memory saving options special for ELKS, since we are very limited on memory and can't afford to keep buffers as long as we are. This could be the default, with an option to increase them for WAN operations, for instance.

Mellvik:
@ghaerr,
I may seem I like nagging about things. Well, I don't :-). Then again it makes for useful discussions to see things differently.

I do see the state differences between the two cases, but I'm still unconvinced about the correctness of the long wait. Bear with mo for another round.

> 11. nov. 2021 kl. 17:52 skrev Gregory Haerr ***@***.***>:
> 
> 
> It seems to me we're going into FINWAIT even though the FIN has been sent, ACKed and received.
> 
> Notice the difference between the two final states: one is LAST_ACK and the other is FIN_WAIT_2. Look at the state diagram, the FIN_WAIT_2 sends a final ACK, then goes into TIME_WAIT for the reason I gave above. That final ACK could be lost.
> 
That's the thing: As far as I can see, the final ack has been received. If we follow the state diagram, we should AFAIK never enter FIN_WAIT2. While in FIN_WAIT1, we're getting a combined FIN & ACK from the other side, aren't we? So we go directly to TIME_WAIT - which may not much difference if it's 10 secs, but that's not what I'm seeing from the ^P debugger. It looks like 1 sec. 
What I see is a loop involving FIN_WAIT2 waiting for an ACK that has already been received. Again, I don't think we should be in FIN_WAIT2 in the first place.

Am I making sense?

—M

ghaerr:
@Mellvik, I'll post a PR with some options for TIME_WAIT reduction. The value is currently specified in seconds, I think we need to test at 1/16 second intervals and see what effect this has, before making any final decisions. 

This should work fine on a LAN. If TIME_WAIT should not be reduced (possibly for WAN, or otherwise), then it might make sense to slow down the back-to-back file transfers on outgoing files, since that's the case where the TIME_WAIT state is required on socket close.

ghaerr:
> Am I making sense?

Yes, but you're not following the state machine spec. ALL states from a client-closed socket MUST go to TIME_WAIT. That includes FIN_WAIT_1, FIN_WAIT_2, and CLOSING. We can't hack the spec without suffering consequences, and it sounds like you're arguing that we shouldn't go into TIME_WAIT, when the spec demands it. The server-closed sockets go through an entirely different set of states - CLOSE_WAIT and LAST_ACK, which never require TIME_WAIT. We are seeing both these happening properly according to spec.

Talking about that we've received an ACK - yes, both CLOSING and FIN_WAIT_1 receive a final ACK and go into TIME_WAIT. As I mentioned before, it's not about receiving the final ACK, its about the other side getting our final ACK, which could be lost. In the case where we don't go into TIME_WAIT, the remote side IS going into TIME_WAIT for the same socket.

Here are some articles talking more about the issues with and understanding the TIME_WAIT state:
http://www.serverframework.com/asynchronousevents/2011/01/time-wait-and-its-design-implications-for-protocols-and-scalable-servers.html

https://knowledgebase.progress.com/articles/Article/Can-the-time-a-socket-spends-in-TIMED-WAIT-state-be-reduced

Bottom line: I don't dare redesign the TCP state machine because I'm no expert, but I think we can reliably reduce the time wait considerably, into the 200-500 msec range, for local ELKS connections, and only then because we have to, and knowing the risk doing so described in the first (very well written) article above.

> If we follow the state diagram, we should AFAIK never enter FIN_WAIT2. 

Which side we end up on is entirely dependent on which side closes the socket first. We are seeing the proper state transitions. What happens on the wire is a consequence of which side closed the socket first, since TCP does allow half-duplex connections afterwards (ktcp does not, but that doesn't matter).

Mellvik:
> 
> Am I making sense?
> 
> Yes, but you're not following the state machine spec. ALL states from a client-closed socket MUST go to TIME_WAIT. That includes FIN_WAIT_1, FIN_WAIT_2, and CLOSING. We can't hack the spec without suffering consequences, and it sounds like you're arguing that we shouldn't go into TIME_WAIT, when the spec demands it. The server-closed sockets go through an entirely different set of states - CLOSE_WAIT and LAST_ACK, which never require TIME_WAIT. We are seeing both these happening properly according to spec.
> 
NO, I'm not. to the contrary - I'm pointing out that we're in FIN_WAIT2, which is not according to spec. And that - from the ^P output, that's the problem, not the TIME_WAIT. As to the lengt of TIME_WAIT that's a different discussion.
> Talking about that we've received an ACK - yes, both CLOSING and FIN_WAIT_1 receive a final ACK and go into TIME_WAIT. As I mentioned before, it's not about receiving the final ACK, its about the other side getting our final ACK, which could be lost. In the case where we don't go into TIME_WAIT, the remote side IS going into TIME_WAIT for the same socket.
> 
We must be reading the outputs differently. I don't see us coming to TIME_WAIT at all, lingering in FIN_WAIT2 for all of 10 seconds.
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-1.4)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-3.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-4.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-5.3)
tcp: time_wait 0 user_timeout 1
tcp: expire check FIN_WAIT_2 (-6.3)
tcp: time_wait 0 user_timeout 1
etc etc

Thanks for the links, that will be my evening reading. Helluva lot better than all the bad news from around the world.


> If we follow the state diagram, we should AFAIK never enter FIN_WAIT2.
> 
> Which side we end up on is entirely dependent on which side closes the socket first. We are seeing the proper state transitions.
> 
Well I don't - how can it possibly be correct given the packet exchanges to end up in FIN_WAIT2?


Checking out for the evening. I think a # of great optimizations are coming out of this. Bending the specs? No, wouldn't cross my mind. It works for a reason.

—M

ghaerr:
@Mellvik,

> While in FIN_WAIT1, we're getting a combined FIN & ACK from the other side, aren't we?

I finally see your point and a potential problem: looking at source code, it seems that ktcp doesn't handle the case of a combined FIN and ACK, and always transitions to FIN_WAIT_2 regardless, where it should transition directly to TIME_WAIT in this case. Did you see the FIN and ACK on the wire?

I'll submit a PR with a fix for this, and also commit ftpd and the other TCP buffer reduction PR beforehand.


>  So we go directly to TIME_WAIT - which may not much difference if it's 10 secs, but that's not what I'm seeing from the ^P debugger. It looks like 1 sec.

Turns out there are separate timers for "CLOSE" waiting, which includes the CLOSING and both FIN_WAIT timers, and a different one for TIME_WAIT. Both are set for 10 secs. I'll add change to allow both to be specified in sub-second intervals, and default TIME_WAIT lower for testing. I think the CLOSE waits will have to stay higher, but thanks to your keen observation, we shouldn't stuck in FIN_WAIT_2 state anyways.

Thank you!

Mellvik:
Weird - (I got som extra time)
and I was just looking at the same code. The trace is unequivocal:

16:32:09.699078 IP (tos 0x0, ttl 64, id 591, offset 0, flags [none], proto TCP (6), length 64)
    10.0.2.15.ftp > 10.0.2.2.53540: Flags [P.], cksum 0x8304 (correct), seq 2157:2181, ack 654, win 512, length 24: FTP, length: 24
	226 Transfer Complete.
16:32:09.699163 IP (tos 0x10, ttl 64, id 40736, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.53540 > 10.0.2.15.ftp: Flags [.], cksum 0x182b (incorrect -> 0xe1d1), ack 2181, win 29200, length 0
16:32:09.756749 IP (tos 0x0, ttl 64, id 592, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.45271: Flags [F.], cksum 0x8034 (correct), seq 12037, ack 1, win 512, length 0
16:32:09.757269 IP (tos 0x8, ttl 64, id 5732, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.45271 > 10.0.2.15.ftp-data: Flags [F.], cksum 0x182b (incorrect -> 0xb032), seq 1, ack 12038, win 53760, length 0
16:32:09.828860 IP (tos 0x0, ttl 64, id 593, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.45271: Flags [.], cksum 0x8033 (correct), ack 2, win 512, length 0

Something is missing from the state machine (in tcp_fin_wait_1()).

More later.

-M

PR #1001, opened 2021-11-11
[net] Various fixes to ftpd, ktcp, and TIME_WAIT

ghaerr:
Various fixes discussed in #999 and elsewhere.

Summary:

- ftpd: add code to become a daemon, add to build, distribution and net start/stop.
- ktcp: fix FIN_WAIT2 bug when FIN and ACK received in FIN_WAIT_1; now goes directly to TIME_WAIT state.
- All timeouts now have 1/16 second granularity.
- TIME_WAIT default to half second.

I'm currently unable to test ELKS ftpd as I probably have the wrong settings for ftp'ing into QEMU.

### DIFF
--- a/elks/arch/i86/lib/unreal.S
+++ b/elks/arch/i86/lib/unreal.S
@@ -28,6 +28,7 @@
 #define USE_A20ASM
 //#define USE_HIMEM_AT
 
+	.arch	i386,nojumps
 	.code16
 	.text
 
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -151,7 +151,6 @@ disk_utils/partype				:diskutil				:1440k
 disk_utils/ramdisk				:diskutil				:1440k
 disk_utils/fdisk				:be-diskutil	:360k
 busyelks/busyelks				:busyelks
-inet/httpd/httpd				:net
 inet/httpd/sample_index.html	::var/www/index.html	:net
 ktcp/ktcp						:net
 ktcp/resolv.conf	::etc/resolv.conf	:net
@@ -160,6 +159,8 @@ inet/nettools/nslookup			:net
 inet/nettools/arp				:net
 inet/telnet/telnet				:net
 inet/telnetd/telnetd			:net
+inet/httpd/httpd				:net
+inet/ftp/ftpd					:net
 inet/tinyirc/tinyirc			:net
 inet/urlget/urlget				:net
 inet/urlget/urlget :::ftpget	:net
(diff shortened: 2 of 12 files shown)


Mellvik:
Thanks @ghaerr, that was fast indeed!

And thanks again for the TIME_WAIT links - very enlightening.

Looking forward to test in the morning.

--Mellvik

ghaerr:
@Mellvik,

Pushed more fixes:

- Fixed in_gethostbyname, was always return last entry when host not found in /etc/hosts
- Better error message in net_connect in ftpget for when hostname not found
- Changed ftpget/ftpput to return actual program name in error messages instead of always 'ftp:' for clarity
- Reset TIME_WAIT to 4 seconds (instead of 1/2 second), still down from 10 seconds

Now that I realize what is happening the last ^P debug trace now makes sense: when the following line was being seen
```
tcp: port 20 reused, other socket NOT in time_wait!
```
I didn't at first understand it. The reason is, the bug @Mellvik found where the closed socket was going into FIN_WAIT_2 state, rather than TIME_WAIT, and so previous socket alloc was not being released when port 20 was being reused.

Now that this PR should fix that problem, I don't think there's a need to radically reduce TIME_WAIT, since the ftp daemon will cause that socket control block to be freed when it reuses the port. This will in turn free up the memory that is causing ktcp to run out of heap. The message we should be seeing if working properly is:
```
"tcp: port 20 reused, freeing previous socket in time_wait
```
This PR does reduce the TIME_WAIT to 4 seconds, which should help a bit more to release memory, while not interfering in possible close problems (until we finally figure them all out!!).

Since I can't (yet) access ftpd outside of ELKS (in QEMU), I started trying to use ftpget to access ftpd. It is always failing with 'ftpget: Error listing directory: 502`. I haven't jumped into this, but was wondering if you could fix that? That would be very helpful!

Thank you!

Mellvik:
This is great, @ghaerr. 

I'll be back shortly with test results. I created a collection of about 50 small files to transfer - that will be the litmus test for the new socket resource release regime. 
Now - from one of the TIME_WAIT links you sent yesterday, it is likely to fail eventually - which is OK. It's more like a feature, not a bug. And it's our privilgege to adjust the timeout to make sense - in our setting. Ideally, there would be an under the hood, automatic, per connection round trip time recording which would set the timeout. I don't think that exists anyware.

Will look into the urlget/ftpd problem right away. Are you using urlget on the mac?

--Mellvik

PR #1002, opened 2021-11-12
[net] Cleanup ftpget, set reuse port 21 in ftp daemon

ghaerr:
Cleans up urlget/ftpget source, standardizes error messages. "Usage" lines still need work.

Use setsockopt to reuse port 21 in ftp daemon.

Sets buffer to power of two in urlget/ftpget (8000 -> 4096). Not sure proper size due to track caching; all programs will wait in kernel doing 512 byte writes to network, and BIOS I/O only works in 1K chunks, so disk reads are only optimized through track caching.

No other functional changes.

### DIFF
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -425,6 +425,11 @@ int main(int argc, char **argv){
 		exit(1);
 	}
 
+	/* set local port resuse, allows server to be restarted in less than 10 secs */
+	ret = 1;
+	if (setsockopt(listenfd, SOL_SOCKET, SO_REUSEADDR, &ret, sizeof(ret)) < 0)
+		perror("setsockopt SO_REUSEADDR");
+
 	bzero(&servaddr, sizeof(servaddr));
 	servaddr.sin_family      = AF_INET;
 	servaddr.sin_addr.s_addr = htonl(INADDR_ANY);
--- a/elkscmd/inet/urlget/net.c
+++ b/elkscmd/inet/urlget/net.c
@@ -18,9 +18,9 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 
-int net_connect(host, port)
-char *host;
-int port;
+void errmsg(const char *, ...);
+
+int net_connect(char *host, int port)
 {
 	int netfd;
 	struct sockaddr_in in_adr;
@@ -28,15 +28,20 @@ int port;
 	int ret;
 	
 	netfd = socket(AF_INET, SOCK_STREAM, 0);
+	if (netfd < 0) {
+		errmsg("Can't open socket (check if ktcp is running)");
+		return -1;
+	}
 	
 	in_adr.sin_family = AF_INET;
 	in_adr.sin_port = PORT_ANY;
 	in_adr.sin_addr.s_addr = INADDR_ANY;
 
     ret = bind(netfd, (struct sockaddr *)&in_adr, sizeof(struct sockaddr_in));
-	if (ret < 0){
+	if (ret < 0) {
 		perror("Bind failed");
-		exit(1);
+		close(netfd);
+		return -1;
 	}
 	
 	l.l_onoff = 1;	/* turn on linger option: will send RST on close*/
@@ -49,14 +54,13 @@ int port;
 	in_adr.sin_port = htons(port);
 	in_adr.sin_addr.s_addr = in_gethostbyname(host);
 	if (in_adr.sin_addr.s_addr == 0) {
-		fprintf(stderr, "Can't find host '%s' in /etc/hosts\n", host);
-		close (netfd);
+		errmsg("Can't find host '%s' in /etc/hosts", host);
+		close(netfd);
 		return -1;
 	}
 
 	ret = connect(netfd, (struct sockaddr *)&in_adr, sizeof(struct sockaddr_in));
 	if (ret < 0){
-		perror("Connect failed");
 		close(netfd);
 		return -1;
 	}
@@ -65,9 +69,7 @@ int port;
 }
 
 /* if errflag set, send TCP RST on close, else send FIN */
-void net_close(fd, errflag)
-int fd;
-int errflag;
+void net_close(int fd, int errflag)
 {
 	int ret;
 	struct linger l;
--- a/elkscmd/inet/urlget/urlget.c
+++ b/elkscmd/inet/urlget/urlget.c
@@ -7,7 +7,7 @@
  * Implements HTTP POST & GET, FTP GET,PUT and directory listings, TCP GET (typical via 
  * netcat at the other end.
  * 
- * Assumes hard links to the approproate names - ftpget, urlget, tcpget, tcpput.
+ * Assumes hard links to the appropriate names - ftpget, ftpput, urlget, tcpget, tcpput.
  * Optons (http only):
  * 	-h -- include header in output stream (httpget)
  * 	-d -- discard data (httpget)
@@ -43,7 +43,7 @@ _PROTOTYPE(int ftpcmd, (FILE *fpw, FILE *fpr, char *cmd, char *arg));
 _PROTOTYPE(int ftpio, (char *host, int port, char *user, char *pass, char *path, int type, int verbose));
 _PROTOTYPE(int tcpget, (char *host, int port, char *user, char *pass, char *path));
 _PROTOTYPE(int main, (int argc, char *argv[]));
-_PROTOTYPE(void usage, (char * prg));
+_PROTOTYPE(void usage, (void));
 
 char ftpphost[15+1];
 unsigned int ftppport;
@@ -57,7 +57,7 @@ char *progname;
 
 #define TRANS_DEBUG	0	/* for debug dumps */
 
-char buffer[8000];
+char buffer[4096];
 
 #if 0
 _PROTOTYPE(int strncasecmp, (const char *s1, const char *s2, size_t len));
@@ -195,6 +195,18 @@ int skipit(char *buf, int len, int *skip) {
    return(0);
 }
 
+/* Display an error message*/
+void errmsg(const char *msg, ...)
+{
+    va_list args;
+
+    va_start(args, msg);
+    fprintf(stderr, "%s: ", progname);
+    vfprintf(stderr, msg, args);
+    fprintf(stderr, "\n");
+    va_end(args);
+}
+
 int httpget(char *host, int port, char *user, char *pass, char *path, 
 	int headers, int discard, int post) {
 
@@ -207,7 +219,7 @@ int httpget(char *host, int port, char *user, char *pass, char *path,
 
    fd = net_connect(host, port);
    if (fd < 0) {
-	fprintf(stderr, "httpget: Could not connect to %s:%d\n", host, port);
+	errmsg("Could not connect to %s:%d", host, port);
    	return(-1);
    }
 
@@ -351,7 +363,7 @@ int ftpio(char *host, int port, char *user, char *pass, char *path, int type, in
    }
    fd = net_connect(host, port);
    if (fd < 0) {
-	fprintf(stderr, "%s: Could not connect to %s:%d\n", progname, host, port);
+	errmsg("Could not connect to %s:%d", host, port);
    	return(-1);
    }
    fpr = fdopen(fd, "r");
@@ -359,7 +371,7 @@ int ftpio(char *host, int port, char *user, char *pass, char *path, int type, in
 
    s = ftpreply(fpr);	/* Get server ID message, should be 220 status. */
    if (s / 100 != 2) {
-	fprintf(stderr, "%s: Connect error: %d\n", progname, s);
+	errmsg("Connect error: %d", s);
 	goto error;
    }
    s = ftpcmd(fpw, fpr, "USER", *user ? user : "ftp");
@@ -367,7 +379,7 @@ int ftpio(char *host, int port, char *user, char *pass, char *path, int type, in
    	s = ftpcmd(fpw, fpr, "PASS", *pass ? pass : "urlget@x.com");
 
    if (s / 100 != 2) {
-	fprintf(stderr, "%s: Authentication failed: %d\n", progname, s);
+	errmsg("Authentication failed: %d\n", s);
 	goto error;
    }
 
@@ -380,14 +392,14 @@ int ftpio(char *host, int port, char *user, char *pass, char *path, int type, in
 		if (infile >= 0) {		/* ftpput: Try to create the directory */
    			s = ftpcmd(fpw, fpr, "MKD", unesc(p));
 			if ((s/100) != 2) {	/* 257 = success */
-				fprintf(stderr, "%s: Create directory %s failed, error %d\n", progname, p, s);
+				errmsg("Create directory %s failed, error %d", p, s);
 				goto error;
 			} 
-			if (verbose) fprintf(stderr, "%s: Destination directory created: %s\n", progname, p);
+			if (verbose) errmsg("Destination directory created: %s", p);
    			s = ftpcmd(fpw, fpr, "CWD", unesc(p)); /* assume success */
 
 		} else {
-			fprintf(stderr, "%s: Remote change directory failed: %s -- status %d\n", progname, p, s);
+			errmsg("Remote change directory failed: %s -- status %d", p, s);
 			goto error;	/* assume this is fatal */
 		}
 	}
@@ -397,7 +409,7 @@ int ftpio(char *host, int port, char *user, char *pass, char *path, int type, in
    sprintf(typec, "%c", type == 'd' ? 'A' : type);
    s = ftpcmd(fpw, fpr, "TYPE", typec);
    if (s / 100 != 2) {
-	fprintf(stderr, "%s: Type error: %d\n", progname, s);
+	errmsg("Type error: %d", s);
 	goto error;
    }
    if (strlen(p) == 0) type = 'd'; 	/* last char is '/', it's a directory, list files */
@@ -406,12 +418,12 @@ int ftpio(char *host, int port, char *user, char *pass, char *path, int type, in
    s = ftpcmd(fpw, fpr, "PASV", "");
  
    if (ftppport==0) {			/* set in ftpcmd */
-	fprintf(stderr, "%s: Error listing directory: %d\n", progname, s);
+	errmsg("Error listing directory: %d", s);
 	goto error; 
    }
    fd2 = net_connect(ftpphost, ftppport);
    if (fd2 < 0) {	
-	fprintf(stderr, "%s: Network connect error.\n", progname);
+	errmsg("Network connect error");
 	goto error;
    }
    if (verbose) 
@@ -421,7 +433,7 @@ int ftpio(char *host, int port, char *user, char *pass, char *path, int type, in
    if (infile < 0) { 		/* ftpget */
 	s = ftpcmd(fpw, fpr, type == 'd' ? list : "RETR", unesc(p));
    	if ((s/100) != 1) {
-		fprintf(stderr, "%s: Cannot open remote file: %d\n", progname, s);
+		errmsg("Cannot open remote file: %d", s);
 		goto error;
    	}
 	while ((s = read(fd2, buffer, sizeof(buffer))) > 0) {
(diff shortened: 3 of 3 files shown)


Mellvik:
Thank you for the much needed cleanup in urlget, @ghaerr!

I'll keep the usage() update on my list. The File transfer wiki is up to date (missing ftpd though, to be added when the ftp client is ready).

Reuseaddr for port 21 in ftpd is really a NOOP, it doesn't do any good or harm. The port stays for an entire session (in my case for hours btw), closes only when the process exits. I haven't seen REUSEADDR used on the control connection in any of the dozen or so ftp servers I've looked at over the past week. Then again, it doesn't do any harm.

I'm going to add passive mode to ftpd shortly. Client applications like curl assumes passive mode (although many implementations has acive mode as an option), it makes life easier for users if it's there, and it's a nobrainer now that we're this far.

Next is the ftp client - and testing ktcp back-to-back, which I hope to do this weekend.

Thanks again, this rocks. 
BTW - what's the plan for 0.5 - anything that needs to take priority?

--Mellvik

Mellvik:
Test update with the new socket close code - this is looking REALLY GOOD:

- When attempting to run a 2nd `ftpd`, which tries to open port 21 (which is busy), the error message is `tcp: port 21 reused, can't free previous socket in state `. This may be intentional, but to me, EBUSY would make more sense.
- Transferring many small files (MGET) from ELKS **works great**. The test case is 55 files @ 200 bytes each, and there is minimal noticeable delay - like 0.5 seconds or so every (ca.) 10 files. And a number of the expected messages `tcp: port 20 reused, freeing previous socket in time_wait`.

Looks like you nailed it, @ghaerr, congrats!!
I came across a few snags in `ftpd`- no big surprises. PR is coming when passive mode has been tested.

I'll say, `ktcp` has made another significant step towards 'industrial quality'.

--M

Mellvik:
One more thing, @ghaerr - seems we're not quite home yet:

I just noticed - there is a window problem with incoming transfers. It's really weird: 
When opening the connection, elks advertises a 128 byte (!) receive window. Then, after receiving the first (128 bytes) packet, the window goes to 1. Next, after getting 1 byte, the window expands to 127 bytes, then back to 1 - and this continues for the duration of the transfer. Alternating 127 and 1. Obviously very slow.

Then it becomes even more intriguing since 1) outgoing transfers are 'normal', having the 512 byte window they've had since we made changes to windowing a couple of weeks back. 2) transfers outside `ftpd` are normal (`ftpput/ftpget`), and 3) the window size was normal yesterday, I just looked back at the traces.

So between yesterday and today a change has had the unintended consequence of disturbing the receive window. And I haven't found anything in `ftpd` itself that would explain it. Any clues?

--M

ghaerr:
> Transferring many small files (MGET) from ELKS works great. The test case is 55 files @ 200 bytes each, and there is minimal noticeable delay - like 0.5 seconds or so every (ca.) 10 files.

Fantastic!! We did it, fast back-to-back file transfers and no excess memory usage!

> Looks like you nailed it, @ghaerr, congrats!

Congrats to you, @Mellvik, you did half the work. And I'm especially glad you noticed a third state transition out of FIN_WAIT_1, matched that to your packet trace, and forced me to finally see it!

> When attempting to run a 2nd ftpd, which tries to open port 21 (which is busy), the error message is tcp: port 21 reused, can't free previous socket in state . This may be intentional, but to me, EBUSY would make more sense.

Well, this only happens during bind when the SO_REUSEADDR is in effect. The problem with removing it is that, although it would work for the case of a 2nd daemon being started, it could cause problems in the case of multiple transfers FTP outbound transfers where the previous socket isn't fully closed yet (into TIME_WAIT). And we need to use the SO_REUSEADDR for the FTP data channel.

> One more thing, @ghaerr - seems we're not quite home yet

Geez, I just received this while still writing the congratulations response!! :)

> So between yesterday and today a change has had the unintended consequence of disturbing the receive window. And I haven't found anything in ftpd itself that would explain it. Any clues?

Yep - part of the new code uses a small, 128 byte TCP control block buffer on "server binds" that intend to call listen, then accept. This was part of the way we've reduced memory usage. The idea is that the bind and listen socket uses a 128 byte buffer (that's why you're seeing the 127 byte window, a SYN was received), and then when an incoming connection is granted, that new socket is created with a full 4096 window.

I just figured it out - your FTPD code uses a "server-style bind" with the port set locally at bind time, THEN calls connect, rather than listen, which uses the same small buffer. I'll have to fix this using a different mechanism.

In the meantime, if you could print a printf in the tcp_output routine that shows the CB_BUF_SPACE value, that would be nice. I'm thinking that it's getting 128 byte data packets and the receive window is 0, then being reset to 1.

ghaerr:
> In the meantime, if you could print a printf in the tcp_output routine that shows the CB_BUF_SPACE value, that would be nice. I'm thinking that it's getting 128 byte data packets and the receive window is 0, then being reset to 1.

No need for extra recompilation and testing. What we're seeing is definitely the case of doing I/O on a 128 byte buffer, and in this case, the tcp_output code advertises the full output window, receives a single packet of the same size, which reduces the available space to 0, which sets 1, then repeats, flipping back and forth from max to min. I'll fix that too - ktcp should never advertise the actual max size of its input buffer, a threshold should always be subtracted. 

For the memory fix, we can still use small TCP control blocks for servers waiting for SYN packets. The problem really is that in the Berkeley socket design, one doesn't know what the application intends to do at strictly bind time, it normally calls listen when a non-zero port is specified, but not always, as in the ftpd non-passive file transfer case.

I'm going to add the SO_RCVBUF socket option, which is already standard. And add that to each of the servers to reduce their listen memory usage. All other sockets will use the standard larger size, and this problem be solved.

Mellvik:
Sounds good, @ghaerr - 

I already had the CB_BUF_SPACE printout - and it showed exactly what you expected.

I'll be awaiting the fix and test again.

Thank you!

—M

ghaerr:
@Mellvik,

Also, I'm thinking of fine-tuning the actual buffer size, and advertised buffer size, so that we have maximum throughput, in this fix for this latest problem. We currently use 4096 bytes, which isn't fine tuned. Here are my calcs:
```
Ethernet MTU = 1500
SLIP MTU = 1064, but was lowered to 1024 since our serial ring buffers are 1024 bytes
Using the higher ethernet MTU, subtract 40 byte IP header = 1460
Allow three packets in sliding window = 3 * 1460 = 4380
Ktcp subtracts a "PUSH_THRESHOLD" (=512) from the advertised window size, which is supposed to be
  the amount of data we wait for before sending to an application. This can be lowered to 64, to save
  memory, and will fit more nicely with our "listen" buffer size of 128.
Thus our buffer size should be 4380+64 = 4444 bytes.
```
Currently, ktcp passes any data received up to the application, ignoring the push threshold, which was commented out a long time ago when things worked very poorly. I think we can re-enable that code, with the threshold set for 64 bytes, and set the new (slightly larger) buffer size for max throughput.

What do you think?

Mellvik:
@ghaerr,
I thing this makes a lot of sense. A couple of observations:
Elks ethernet does not ever advertize a 1460 mtu as would be expected (and which linux does), it always advertizes a 984 mtu. So either we fix that or change the calculation, right?

The other thing is an idea from way back, thusly very foggy: What if we had a general packet buffer pool instead of a per connection pool (of 3 in this case). Assuming we (ktcp) have space for say 16 active connections, that is 48 buffered packets, of which a connection could have say, max 6, dynamically reduced to 3, when the load is high, we'd get more bang for the buck unless the increased buffer admin ends up a killer ...

Like I said, I forget the details, origin and fate of this idea, but it was there at some point... and it seems worthwhile airing - and possibly just killing.

As to the push threshold, yes, I think it should be reintroduced. We need to see how it works in this environment. IIRC many (most?) packets from ELKS have the push flag set, we may have some interesting consequences from the change. Such as affecting the 'push any key to continue' issue ...

IOW  - go for it!

Thank you - this is interesting and encouraging.

--Mellvik

ghaerr:
> Elks ethernet does not ever advertize a 1460 mtu as would be expected (and which linux does), it always advertizes a 984 mtu. So either we fix that or change the calculation, right?

Yes, this should be changed also to 1500 - 40. The 984 is 1024 - 40 bytes for the IP header. Ktcp currently always uses MTU_SLIP, which is 1024. I'll change this in the upcoming PR to advertise the appropriate MTU.

> As to the push threshold, yes, I think it should be reintroduced. We need to see how it works in this environment. 

I have been reviewing the code for this, and it is tricky; more on this below. I plan enabling this in two PRs, the first to fix the current ftpd 128-byte socket buffer problem and prepare the buffer sizes and calculations for a larger MTU and receive window, and then enable that in a 2nd PR. I don't want to break what we've just done.

> IIRC many (most?) packets from ELKS have the push flag set, we may have some interesting consequences from the change. 

Yes, I think ELKS always sets it, but that won't necessarily be the case with external incoming packets. And there is still the problem of how fast ELKS can "eat" data from external sources without causing problems. Which brings us to...

> Such as affecting the 'push any key to continue' issue ...

After pondering the problem of exactly what to set the receive window to when we may have variable buffer sizes (like 4444 and 128), I did a thought experiment and think I've discovered what the 'push any key to continue' bug really is.

You may remember that even though we've allocated all these large (4096) buffers, in both the TCP control blocks per connection as well as in a couple different static buffers within ktcp, all to allow receipt of a full-sized ethernet packet or large SLIP frame, we hack our receive window back to THROTTLE_MAXWINDOW (=512) regardless. This was done, because when the throttle-back is increased beyond, say 760 or so, we get the 'push any key to continue' behavior. When that bug occurs, the data is delayed until about a second (becomes jerky), or is immediate when a key is pressed.

I'm almost sure the real reason why isn't anything to do with a PTY, but instead is this: when the receive window is large, the sender sends more data in less packets, and stops sending quicker with no more packet I/O because the receive window is full quicker. Meanwhile, completely separately, ktcp and the kernel tcpdev handler are being scheduled in and out amongst other ELKS applications, and ktcp ends up trying to write more data to tcpdev (which then gets eventually to the application) than it can send in one write. (Tcpdev buffer is limited to 1024 for kernel space reasons). All the while, ktcp's main loop is to sit in a select statement waiting for an event (like a packet coming in or other I/O). The select statement waits forever unless there's a timer outstanding or data that needs to be pushed, in both cases it sets a 1 second timeout.

What's happening I think is that in the large buffer case, ktcp can't fulfill writing all the data to tcpdev and so goes into a select with a one-second timeout, after which the next "block" of incoming data is processed. And since the bug is noticed during a telnet or telnetd operation, there's a socket connected to a shell, and that socket has I/O on it when a character is pressed, which causes the select to return immediately. There you have it! A temp fix might be to lower the select timeout to a 20ms value or so to allow ktcp to run another cycle and push more data, instead of waiting.

It's all quite complicated, I'll try to solve it with a few PRs after getting things repeatable.

> Assuming we (ktcp) have space for say 16 active connections, that is 48 buffered packets, of which a connection could have say, max 6, dynamically reduced to 3, when the load is high, we'd get more bang for the buck

I'm going to suggest we put that idea off until we can actually use the buffer sizes we have, and get ktcp working with large receive windows and incoming packet sizes before considering this additional complexity.

On another note, I found another great [article](https://vincent.bernat.ch/en/blog/2014-tcp-time-wait-state-linux#purpose) talking about time_wait states on Linux, and how they shouldn't be shortened (Linux uses a 1 minute time_wait). I had been wondering how Linux might manage the large number of time_wait states lingering for a large web server.  The answer is: memory, lots of it. It mentions you might use ~2.5MB of extra memory, just for that!

Mellvik:
@ghaerr, there went the entire Saturday morning :-)

You know - when you're reading the news in the morning and catch yourself looking for sequence numbers and acks, you know you're in serious trouble...

Joking aside - what interesting (and encouraging btw) reading. Thank you for a thorough and insightful analysis. This is indeed the type of thinking that solves real problems instead of curing symptoms. ﻿
> Elks ethernet does not ever advertize a 1460 mtu as would be expected (and which linux does), it always advertizes a 984 mtu. So either we fix that or change the calculation, right?
> 
> Yes, this should be changed also to 1500 - 40. The 984 is 1024 - 40 bytes for the IP header. Ktcp currently always uses MTU_SLIP, which is 1024. I'll change this in the upcoming PR to advertise the appropriate MTU.
> 
You know - with this change and the window size fix, we'll increase file transfer speed by another 40%  give or take a few!
> As to the push threshold, yes, I think it should be reintroduced. We need to see how it works in this environment.
> 
> I have been reviewing the code for this, and it is tricky; more on this below. I plan enabling this in two PRs, the first to fix the current ftpd 128-byte socket buffer problem and prepare the buffer sizes and calculations for a larger MTU and receive window, and then enable that in a 2nd PR. I don't want to break what we've just done.
> 
Great plan.  
> IIRC many (most?) packets from ELKS have the push flag set, we may have some interesting consequences from the change.
> 
> Yes, I think ELKS always sets it, but that won't necessarily be the case with external incoming packets. And there is still the problem of how fast ELKS can "eat" data from external sources without causing problems.
> 
Actually - if I remember the intention of PSH correctly, it would make more sense for ELKS to never set it, as it encourages the other end to send faster (and smaller). For file transfers that does not make sense (it doesn't happen when the other end is Linux, it might with other clients, such as ELKS :-) ). Fot telnet, it's a different story.
> Which brings us to...
> 
> Such as affecting the 'push any key to continue' issue ...
> 
> After pondering the problem of exactly what to set the receive window to when we may have variable buffer sizes (like 4444 and 128), I did a thought experiment and think I've discovered what the 'push any key to continue' bug really is.
> 
> You may remember that even though we've allocated all these large (4096) buffers, in both the TCP control blocks per connection as well as in a couple different static buffers within ktcp, all to allow receipt of a full-sized ethernet packet or large SLIP frame, we hack our receive window back to THROTTLE_MAXWINDOW (=512) regardless. This was done, because when the throttle-back is increased beyond, say 760 or so, we get the 'push any key to continue' behavior. When that bug occurs, the data is delayed until about a second (becomes jerky), or is immediate when a key is pressed.
> 
> I'm almost sure the real reason why isn't anything to do with a PTY, but instead is this: when the receive window is large, the sender sends more data in less packets, and stops sending quicker with no more packet I/O because the receive window is full quicker. Meanwhile, completely separately, ktcp and the kernel tcpdev handler are being scheduled in and out amongst other ELKS applications, and ktcp ends up trying to write more data to tcpdev (which then gets eventually to the application) than it can send in one write. (Tcpdev buffer is limited to 1024 for kernel space reasons). All the while, ktcp's main loop is to sit in a select statement waiting for an event (like a packet coming in or other I/O). The select statement waits forever unless there's a timer outstanding or data that needs to be pushed, in both cases it sets a 1 second timeout.
> 
> What's happening I think is that in the large buffer case, ktcp can't fulfill writing all the data to tcpdev and so goes into a select with a one-second timeout, after which the next "block" of incoming data is processed. And since the bug is noticed during a telnet or telnetd operation, there's a socket connected to a shell, and that socket has I/O on it when a character is pressed, which causes the select to return immediately. There you have it! A temp fix might be to lower the select timeout to a 20ms value or so to allow ktcp to run another cycle and push more data, instead of waiting.
> 
I think you're spot on. This just makes sense. Even the fact that I'm not seeing the same symptoms on real HW makes sense - it's timing-related. Even though I haven't found a logical explanation (yet), this may also be the problem causing telnet to loose charaters in some circumstances. You may remember me chasing that problem a year back. I never found it, and again, the pty driver became the primary suspect. 
As to the solution, it seems we need a signaling mechanism so tcpdev can tell ktcp that it's ready for more? Now, if the temp solution you suggest doesn't kille the iron performancewise, let's check it out!
> It's all quite complicated, I'll try to solve it with a few PRs after getting things repeatable.
> 
> Assuming we (ktcp) have space for say 16 active connections, that is 48 buffered packets, of which a connection could have say, max 6, dynamically reduced to 3, when the load is high, we'd get more bang for the buck
> 
> I'm going to suggest we put that idea off until we can actually use the buffer sizes we have, and get ktcp working with large receive windows and incoming packet sizes before considering this additional complexity.
> 
Agreed, let's shelve this one for good. After thinking about it it seems too complicated. And I finally remembered where it came from - a code review of a terminal server (you remember those) in the late 80s. Special purpose, special 'fixes' (and - like our case, very strained resources - 80186 etc.).
> On another note, I found another great article talking about time_wait states on Linux, and how they shouldn't be shortened (Linux uses a 1 minute time_wait). I had been wondering how Linux might manage the large number of time_wait states lingering for a large web server. The answer is: memory, lots of it. It mentions you might use ~2.5MB of extra memory, just for that!
> 
Now this is where the rest of my morning went. GREAT article. 

And at the end of the day it seems the path we ended up taking, is the right one for ELKS. There are more tricks to play without breaking the standard, but it seems what we have is good enough - at least for now. The article even recommends expanding the port range if the situation becomes tight. One of the FTP servers I looked at did that. If a new data connection from port 20 could not be opened and the error was EADDRINUSE, it decremented the port number and tried again. 
Which works. I tested that using port numbers in the 800 range and then in the 50000 range, no problem - except we run out of socket buffers. So it's a bad idea unless we restrict the number to say 5, and circulate among those. Which also works - and in our case reduced the delay significantly. 
Now that you've tuned the timeouts, it's unneccessary. The little delays we notice when pulling lots of files are minimal. And the normal case would have a mix of larger and smaller files (very few as small as the 200 byte test case), which would vipe out the visible delays entirely.
So, from my perspective, we're home as far as the TIME_WAIT situation is concerned. Ktcp handles the state changes  correctly and - like we talked about before, the litmus test will be ktcp back-to-back. Which is coming.

Thanks again!

--Mellvik

ghaerr:
> The article even recommends expanding the port range if the situation becomes tight. One of the FTP servers I looked at did that. If a new data connection from port 20 could not be opened and the error was EADDRINUSE, it decremented the port number and tried again.
Which works. I tested that using port numbers in the 800 range and then in the 50000 range, no problem - except we run out of socket buffers. 

Interesting, I didn't know you tried that. Ultimately, for ELKS, I think we're better off leaving it at port 20, like you've done, since ELKS is so limited in socket buffer (and other) memory. Ktcp doesn't implement reserved ports and all ports up to 65535 are open.

> Now that you've tuned the timeouts, it's unneccessary. 

Yes - however: the reason the ftp file transfers are working well is not because of the timeout tuning; it is because we're freeing the previous port 20 socket that's in the time_wait state. That was possible because we added SO_REUSEADDR and you identified the problem of closed sockets remaining in one of the close wait states, and not going to time_wait.

The only timeout tuning that was done is reduce time_wait to 4 seconds; I'm not sure what that's actually helping, other than to release the closed socket control blocks earlier (against the recommendation of both articles, but still OK for ELKS for most uses, I think).

> So it's a bad idea unless we restrict the number to say 5, and circulate among those. Which also works - and in our case reduced the delay significantly.
> The little delays we notice when pulling lots of files are minimal. And the normal case would have a mix of larger and smaller files (very few as small as the 200 byte test case), which would vipe out the visible delays entirely.

What delays are you talking about here? I wasn't aware we're seeing any delays.

Thank you!

Mellvik:
@ghaerr, yes - we're getting there.
> Now that you've tuned the timeouts, it's unneccessary.
> 
> Yes - however: the reason the ftp file transfers are working well is not because of the timeout tuning; it is because we're freeing the previous port 20 socket that's in the time_wait state. That was possible because we added SO_REUSEADDR and you identified the problem of closed sockets remaining in one of the close wait states, and not going to time_wait.
> 
> The only timeout tuning that was done is reduce time_wait to 4 seconds; I'm not sure what that's actually helping, other than to release the closed socket control blocks earlier (against the recommendation of both articles, but still OK for ELKS for most uses, I think).
> 
Agreed!!
> So it's a bad idea unless we restrict the number to say 5, and circulate among those. Which also works - and in our case reduced the delay significantly.
> The little delays we notice when pulling lots of files are minimal. And the normal case would have a mix of larger and smaller files (very few as small as the 200 byte test case), which would vipe out the visible delays entirely.
> 
> What delays are you talking about here? I wasn't aware we're seeing any delays.
> 
As reported earlier in this thread (https://github.com/jbruchon/elks/pull/1002#issuecomment-967144827), there is an occasional slight delay when back-to-back transferring small files. It's practically insignificant, technically interesting. I need to narrow down the environment and eliminate disturbing elements in order to clarify the issue. I hadn't planned to do that,  but will do so anyway. Get the floppy out of the way (transfer from HD) etc. It seems there is a slight delay every 4th or 5th file transferred. May be just scheduling for all I know. And, like I said, when the file sizes are more normal, this will be incredibly hard to notice.

-- Mellvik

PR #1003, opened 2021-11-12
[kernel] Try BIOS INT 15h AX=2401 to enable A20 first for XMS

ghaerr:
On system boot with XMS buffers enabled, but also in `unreal16`, tries to enable A20 using the BIOS first, before using one of two methods using the keyboard controller. Hopefully this will work on the Compaq 386.

The routines now save the state of the interrupt flag and restore it properly, disabling interrupts only when necessary.

@Mellvik, if you could test `unreal16` again with no modifications and report the console and tty output, that would be helpful. If it still doesn't work, perhaps try the USE_HIMEM_AT option in elks/arch/i86/unreal.S.

### DIFF
--- a/elks/arch/i86/lib/unreal.S
+++ b/elks/arch/i86/lib/unreal.S
@@ -23,6 +23,7 @@
 #
 
 # configurable options for A20 gate
+#   USE_BIOS		- use BIOS AX=2401h INT 15h method (always tried first)
 #   USE_A20ASM		- use Chris Giese A20.ASM method (send D0, read, OR 2, D1, write)
 #   USE_HIMEM_AT	- use himem.sys PC AT method (send D1,DF,FF)
 #define USE_A20ASM
@@ -71,6 +72,7 @@ enable_unreal_mode:
 	addl $gdt, %eax
 	movl %eax, gdt_ptr+2
 
+	pushf			# save interrupt status
 	cli                     # interrupts off
 	pushl %ds
 	pushl %es
@@ -96,7 +98,7 @@ enable_unreal_mode:
 	popl %fs
 	popl %es
 	popl %ds
-	//sti
+	popf			# restore interrupt status
 
 	mov	$1,%ax		# system is in unreal mode, return 1
 	ret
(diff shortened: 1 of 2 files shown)


Mellvik:
Hi @ghaerr,
we have a verdict:


USE_A20ASM:
	As before: 
	unreal16 loops @ 'trying to enable unreal mode …'
	kernel boot reports 'xms: can't enable A20 gate

USE_HIMEM_AT:
	kernel boot reports: 'xms: unreal mode and A20 enabled'
Unfortunately, booting stops at the same place as before :

ttyS0 at 0x3f8, irq 4 is a 16450
ttyS1 at 0x2f8, irq 3 is a 16550A
ttyS2 at 0x3e8, irq 5 is a 16550A
eth: NE2K at 0x300, irq 12, MAC 00:80:29:ef:d9:51
bioshd: hda BIOS CHS 610,16,63
bioshd: 2 floppy drives and 1 hard drive
Partitions: hda:(0,1000944) 

The fact that the rest of the Partitions: message is missing may be misleading (system hung before the output buffer got flushed),
so my guess is that the stop is in the next phase - 'device_setup'.

Any other combinations that are meaningful to test?

—Mellvik

Mellvik:
Update, @ghaerr - 

please put the previous report on hold until further notice.

I now have a situation where the system doesn't boot regardless of the setting in unreal.S.

Until I've found the reason for this, the previous report is unreliable.

—Mellvik

ghaerr:
Hi @Mellvik,

Thanks for the testing on this. I really don't know what the reasons are, but I suspect that something strange is happening with regards to Compaq's BIOS handling of its extended memory or the A20 line. If the A20 line is not working, but tests as though it is, then ELKS will definitely crash as soon as it does BIOS disk I/O and that data is copied into an XMS buffer, which I think is what you're seeing when reading the partition table (that code uses kernel buffers to read the table). With A20 zero, for instance, that copy would result in trashing low memory instead of writing XMS memory in extended memory, and the kernel wouldn't know it. BTW, the XMS buffers are set to start at a fixed 1MB physical address, for now.

It would probably be best to compile a regular (CONFIG_FS_XMS_BUFFERS not set) kernel, boot it, and run `unreal16`, and report exactly what is displayed on the console and on the tty line. That will tell us more information hopefully.

Do you have any other 386 machines (other than Compaq 386's) that you could test on with a non-XMS-enabled kernel to run unreal16 on?

I'll test this on real hardware in the next few days, after I return home from SoCal, and that should help tell us whether this is a problem on all 386's, or just Compaqs.

PR #1004, opened 2021-11-13
[net] Add SO_RCVBUF socket option to enable smaller TCP buffers for listening

ghaerr:
@Mellvik,

This should fix the 128-byte buffer problem you saw on the packet trace discussed in #1002, please test.

`telnetd`, `ftpd` and `httpd` use the new SO_RCVBUF setsockopt option, which for now, only sets the buffer size for listen buffers. This reduces the ktcp heap requirement of all three servers running by around 12K. Ftpd transfers should work well in both directions.

This PR sets the MTU for ethernet properly at 1500, but that should have no effect, since the receive window calculations have not changed yet. That, the PUSH_THRESHOLD, allowing a larger receive window, and tuning the buffer sizes will all come in the next PR.

A couple of the ktcp "port reused" message have been set to debug only messages.

On a separate note, testing `ftpget` for directory listings results in a 502 error message. I presume that's because ftpd doesn't support passive transfers yet?

### DIFF
--- a/elks/include/linuxmt/net.h
+++ b/elks/include/linuxmt/net.h
@@ -22,6 +22,7 @@ struct socket {
     short type;
     unsigned char state;
     unsigned char flags;
+    unsigned int rcv_bufsiz;
     struct proto_ops *ops;
     void *data;
 
@@ -38,7 +39,6 @@ struct socket {
 
     struct wait_queue *wait;
     struct inode *inode;
-    struct fasync_struct *fasync_list;
     struct file *file;
 };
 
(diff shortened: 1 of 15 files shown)


Mellvik:
Thank you @ghaerr - 

I just tested this PR - you can pop the cork. Really!

The buffer size problem is gone, the MSS reporting is 1460 and multi-file transfer both ways work great.

The urlget problem - I'll look into that. Also - I'll do some testing with larger windows just for the heck of it.

Thank you! We're home on this one.!!!


—M

PR #1005, opened 2021-11-14
[net] Fix networking to receive 1500 MTU ethernet packets

ghaerr:
This PR fixes the longstanding 'space to continue' problem discussed in https://github.com/jbruchon/elks/pull/1002#issuecomment-967713316 and elsewhere for some time. The symptoms were that when the ktcp receive window was set larger than around 700 bytes, allowing large packets to be received, incoming streaming data coming into ELKS telnet from a remote server became jerky, in one second intervals. Pressing space bar caused the streaming to continue for a bit longer, then repeat.

The problem was not the PTY driver, and it was only partially the ktcp "select" wait of 1 second described earlier, which did need a much smaller timeout when data needed pushing. Ultimately, the problem was that when the receive window allowed large packets to be received, and the advertised window stopped the remote system from transmitting packets, everything would halt until the remote system resent the last packet, which caused everything to startup again.

The solution involves sending an ACK when the receive window becomes larger again, i.e. when an application reads the data from the TCP control block buffer. This, combined with the select not waiting when data needed to be pushed seems to solve the problem. [EDIT: Note, the application must perform a read of at least 1500 bytes in order to maximize throughput. The maximum copy now allowed from ktcp to the kernel to the application is 1500 bytes. The other direction remains 512 bytes.]

This PR enables 1500 MTU by default, space for 3 outstanding ethernet packets within each TCP buffer, and also attempts to tune telnet, telnetd, httpd and ftpd to use MTU size buffers for receive. Note - ktcp is still limited to sending 512-byte outbound packets.

@Mellvik: please test, I'm hoping you find big speed improvements. I tried cleaning up ftpd carefully, and with the larger buffers, I tried to minimize the 3K+ stack size for two 1500 MTU buffers by decreasing the size of the command channel buffer, and increasing the size of the data channel buffer.

In order to see throughput, I've added a new "debug_tune" option for both the kernel and ktcp. ^P will turn on the ktcp debug_tune, but the kernel will need to be specially recompiled with DEBUG_TUNE set in debug.h. This will then display each application read request, along with the amount of tcp-buffered TCP data available, and the associated window size. I used it for tuning telnet and things are quite speedy now.

The receive window "throttling" is now completely removed.

I'll commit this after I receive initial testing results. It would also be nice to know how ftpget/ftpput work, I can't test those on my system easily (yet). (Also looking forward to PASV mode! :)


Mellvik:
Almost there @ghaerr...

ELKS now advertizes a 4380 opening window, shrinking and expanding during transfer as it should.
I've changed the test case to 100 files @ 2000 bytes each, using the harddisk (actually flash card) on ELKS instead of floppy. The transfer is smooth (both ways), occasional small delays attributable to ELKS disk activity.
ELKS still limits outgoing packets to 512 bytes.
There is noticeable speed increase, numbers coming when we've nailed down the final snags.
There is a data loss problem when transferring into ELKS, which I'm looking into right now. Only 40% of the files contain 2000 bytes on arrival, the rest has 1492 bytes - 508 bytes missing. And there is no particular pattern in the OK/not OK sequence. 

I'll continue testing later today.


—Mellvik

ghaerr:
Hello @Mellvik!

Glad to hear things are almost working! Thank you for testing.

> There is a data loss problem when transferring into ELKS, which I'm looking into right now. Only 40% of the files contain 2000 bytes on arrival, the rest has 1492 bytes - 508 bytes missing. 

I think I know what this is, as I had been looking for code to handle it while researching the last PR, and couldn't find any: when a FIN packet comes in, the application gets notified immediately of the remote close (e.g. gets a 0 return in the application), and the remaining TCP buffer contents are never read.

In the new fast code, the application will get a "full" packet read of 1492 bytes, which is handled by transferring the incoming 1492-sized packet completely to the application in one write. (Why its always 1492 instead of 1500 I'm not sure, that that's ok). In your test case, another packet has been received in the buffer, which is the final file content of 508 bytes. This packet may have the FIN bit set, (or not), but, depending on timing, when the final FIN packet is received, ktcp notifies the kernel that the remote side is closed, which apparently taking precedence over the remaining data contents.

This is all good, because it seems this corresponds to the data loss problem you talked about yesterday.

Thank you!

ghaerr:
@Mellvik,

I think I've fixed the dropped data problem with the last push, ktcp was clearly shutting down the socket before the application read the final TCP packet contents. I've left debugging code in that shows the amount of unread buffer data at the time of receiving a FIN, as well as the last read data count when the socket is finally shutdown, these numbers should match. I'll remove them after your testing.

With this commit, my ftpget/ftpput test works great, time after time, with 4 back-to-back transfers. In the midst of it, another strange bug was found: everything works perfectly (after boot), running the above test any number of times, *except* after netstat is run, then the next file transfer hangs. Funny thing is running netstat again (from another console) allows the transfer to continue, but networking hangs on-and-off from that point on. I haven't tracked this fully down yet, but I think netstat ends up leaving a kernel semaphore in the wrong state. Fix coming, so don't test after netstat is run, for the time being!

ghaerr:
> another strange bug was found: everything works perfectly (after boot), running the above test any number of times, except after netstat is run, then the next file transfer hangs.

In chasing down this last problem with netstat screwing things up, I ran into a kernel overwrite bug at boot time, which showed up when I enabled DEBUG_NET because the system image was too large, and the system wouldn't boot. It took well over four hours, but I've finally identified and fixed the kernel overwrite. I'm going to commit this now so I can post commit that fix also, then continue working on the netstat problem with lots of kernel debug messages. I will fix any issues you have in testing in a seperate PR.

Mellvik:
This is looking really good, @ghaerr - actually excellent.

Big ( in terms of # of files) file transfers in and out (FTP) run smoothly. looking at the traces, all connects/disconnects work exactly as they should. 
As you mentioned, there is a console message about the FIN and the data remaining:
tcp: application read too small after FIN, data_avail 508
one per file (incoming). Nice.

The one thing remaining is the outbound packet size, which - as mentioned yesterday - is still restricted to 512 bytes and is unaffected by window sizes. With that restriction lifted we can do another round of speed measurements.

The recent testing has revealed a few small snags in ftpd, all being fixed along with the PASV mode addition. Again running off of a floppy turns out to be great for finding the esoteric bugs, such as running out of inodes, which is different from running out of file space: It turns out that sending an RST on the data connection when no data has been transferred, hangs the client - and is entirely unnecessary. 

I'm going to rerun the telnet testing that caused lost data last year, to see if that problem is gone, too.

Thank you. I guess I keep repeating myself, but we have another milestone in terms of kernel/ktcp stability & robustness.

—Mellvik

Mellvik:
@ghaerr,
the usual one more thing. You might be interested in this on - which is a new scenario:

For each incoming 1460 packet, `ktcp` responds with 2 ACKs, the first with a very small window (60-160), the second with a larger window (1500 and up). I haven't seen this before, it's extra packets from `ktcp` but functionally it makes no difference.
```
12:26:06.634933 IP (tos 0x8, ttl 64, id 52964, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.47179 > 10.0.2.15.ftp-data: Flags [.], cksum 0x1ddf (incorrect -> 0x0185), seq 5841:7301, ack 1, win 29200, length 1460
12:26:06.675461 IP (tos 0x0, ttl 64, id 3102, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xbbdc (correct), ack 7301, win 64, length 0
12:26:06.699264 IP (tos 0x0, ttl 64, id 3103, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xb608 (correct), ack 7301, win 1556, length 0
12:26:06.699385 IP (tos 0x8, ttl 64, id 52965, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.47179 > 10.0.2.15.ftp-data: Flags [.], cksum 0x1ddf (incorrect -> 0xb541), seq 7301:8761, ack 1, win 29200, length 1460
12:26:06.743537 IP (tos 0x0, ttl 64, id 3104, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xb608 (correct), ack 8761, win 96, length 0
12:26:06.767283 IP (tos 0x0, ttl 64, id 3105, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xb034 (correct), ack 8761, win 1588, length 0
12:26:06.767363 IP (tos 0x8, ttl 64, id 52966, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.47179 > 10.0.2.15.ftp-data: Flags [.], cksum 0x1ddf (incorrect -> 0xd8e2), seq 8761:10221, ack 1, win 29200, length 1460
12:26:06.807883 IP (tos 0x0, ttl 64, id 3106, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xb034 (correct), ack 10221, win 128, length 0
12:26:06.831764 IP (tos 0x0, ttl 64, id 3107, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xaa60 (correct), ack 10221, win 1620, length 0
12:26:06.831865 IP (tos 0x8, ttl 64, id 52967, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.47179 > 10.0.2.15.ftp-data: Flags [P.], cksum 0x1ddf (incorrect -> 0x49de), seq 10221:11681, ack 1, win 29200, length 1460
12:26:06.879854 IP (tos 0x0, ttl 64, id 3108, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xaa60 (correct), ack 11681, win 160, length 0
12:26:06.903599 IP (tos 0x0, ttl 64, id 3109, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xa48c (correct), ack 11681, win 1652, length 0
12:26:06.903668 IP (tos 0x8, ttl 64, id 52968, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.47179 > 10.0.2.15.ftp-data: Flags [.], cksum 0x1ddf (incorrect -> 0x93db), seq 11681:13141, ack 1, win 29200, length 1460
12:26:06.944832 IP (tos 0x0, ttl 64, id 3110, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0xa48c (correct), ack 13141, win 192, length 0
12:26:06.968571 IP (tos 0x0, ttl 64, id 3111, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp-data > 10.0.2.2.47179: Flags [.], cksum 0x9eb8 (correct), ack 13141, win 1684, length 0
```

--M

ghaerr:
>  there is a console message about the FIN and the data remaining: tcp: application read too small after FIN, data_avail 508 one per file (incoming). 

Actually, this is the one message I left in, which is not showing on my system. It indicates that there was more than one application read required after the FIN was received by the network. The message doesn't indicate a problem, just what's happening. (Full report available with ^P on). I suspect the reason is that your system is slower than my laptop and takes longer for the application to actually read the data.



> The one thing remaining is the outbound packet size, which - as mentioned yesterday - is still restricted to 512 bytes and is unaffected by window sizes. 

I'll look into this. The original reason it was left small is that receiving large packets didn't work, and that the large static kernel buffers use up lots of very limited kernel data segment. 

> The recent testing has revealed a few small snags in ftpd, all being fixed along with the PASV mode addition. 

What kinds of things? I definitely would like to see PASV mode, but am interested in the problems you're seeing with a full data connection.


> Again running off of a floppy turns out to be great for finding the esoteric bugs, such as running out of inodes, which is different from running out of file space

The MINIX floppies that are created by ELKS `mfs` for the distribution have had the root directory created with  256 i-inodes. This can be changed using the '-i' parameter set in images/Make.defs for experimentation. We need to keep the number small for the standard distribution because of lack of space on the floppies. (Another reason we should use compressed executables for most of the applications).

> It turns out that sending an RST on the data connection when no data has been transferred, hangs the client - and is entirely unnecessary.

Which client hangs, and exactly why?

Agreed that urlget/ftpget/ftpput send RST when any error occurs, and should not. The changes that were made early in our testing should be revisited, and net_close should be called with error ONLY after data transfers that failed, NOT with other failures, like file not found. A careful re-inspection should be made of the code and another PR submitted. I was going to do that, but you might as well since you're in the middle of it testing ftpd.

>  I'm going to rerun the telnet testing that caused lost data last year, to see if that problem is gone, too.

I remember that now. I'm sure that problem is fixed, since ktcp was usually losing the last packet of data received before a FIN, and always when the data arrived with the FIN.

> You might be interested in this on - which is a new scenario:
> For each incoming 1460 packet, ktcp responds with 2 ACKs, the first with a very small window (60-160), the second with a larger window (1500 and up). I haven't seen this before

Yes - this is the solution to the "space to continue" bug: when the sender stopped sending because of a small window, there appeared to be a timing mismatch/overlap, where, even though all packets were originally ACKed, a second ACK was required to restart the sender, and that's what you're seeing. The "jerkiness" that was seen was actually the sender's timeout expiring and a retransmit sent, which started everything moving again. This is all eliminated with the second ACK being sent. (This is almost the same scenario as a ZWP (zero window probe), where a sender "second knock" is required at times to find the receive window status when the receiver's eventual ACK to restart is lost - but in reverse).

At this point, ktcp sends an ACK after every application read, which fixes the problem, and sometimes results in an extra ACK sent that would otherwise not be strictly required, but should never pose a problem.

>  I guess I keep repeating myself, but we have another milestone in terms of kernel/ktcp stability & robustness. 

I think we have, thank you!

Mellvik:
Thank you @ghaerr!
> 
> The recent testing has revealed a few small snags in ftpd, all being fixed along with the PASV mode addition.
> 
> What kinds of things? I definitely would like to see PASV mode, but am interested in the problems you're seeing with a full data connection.
> 
These are practical (not protocol related) things - such as changing to the logged in user's home directory, having the DIR command buffering the output instead og sending many small packets, skipping the 'deamonization' if in debug mode and a few others.
> Again running off of a floppy turns out to be great for finding the esoteric bugs, such as running out of inodes, which is different from running out of file space
> 
> The MINIX floppies that are created by ELKS mfs for the distribution have had the root directory created with 256 i-inodes. This can be changed using the '-i' parameter set in images/Make.defs for experimentation. We need to keep the number small for the standard distribution because of lack of space on the floppies. (Another reason we should use compressed executables for most of the applications).
> 
Actually I meant this as a benefit :-) - having a small number of inodes makes it easy to create that type of error for the testing.
> It turns out that sending an RST on the data connection when no data has been transferred, hangs the client - and is entirely unnecessary.
> 
> Which client hangs, and exactly why?
> 
This is the Linux client. What happens (btw this has been fixed) is that after opening the data connection, ftpd aborts (with an RST) when the open/O_CREAT fails. Which leaves the client in a confused (interruptible) state.
> Agreed that urlget/ftpget/ftpput send RST when any error occurs, and should not. The changes that were made early in our testing should be revisited, and net_close should be called with error ONLY after data transfers that failed, NOT with other failures, like file not found. A careful re-inspection should be made of the code and another PR submitted. I was going to do that, but you might as well since you're in the middle of it testing ftpd.
> 
Yes, I'll do that right after the next ftpd PR.

It's inspiring getting things to work!!
Thank you.


— M

ghaerr:
> The one thing remaining is the outbound packet size, which - as mentioned yesterday - is still restricted to 512 bytes and is unaffected by window sizes.
> The original reason it was left small is that receiving large packets didn't work, and that the large static kernel buffers use up lots of very limited kernel data segment.

I finally remember the two real reasons we don't use large send buffers for ELKS TCP: 1) ktcp has to keep a retransmit malloc for every packet sent in the sliding window, in case a retransmit is needed, and 2) ktcp can't repackage an application write into a smaller packet size. The first limitation is a problem because ktcp is already scarce on memory, and and memory usage would triple per connection, for instance, to any host with 3 outstanding packets in a send sliding window. The remote host would likely allow more in their receive window, and we've already written special code to throttle ELKS back to stop retransmits that we had already discarded (that was the ZWP discussion). The second limitation is even worse, because if a remote system were not able to accept packets in our chosen "MTU" (which would actually be the hard-coded application write size), then no communication would be possible at all, since the TCP packet is never fragmented from the application.

For these reasons, for now, I think we should leave things as they are. We lose a possible slight speed increase, but we gain reliability and the ability to work well with all systems we've tested against (so far). Instead, getting our ftp, telnet (and http) client and servers running well without errors would be a better next step.

> It's inspiring getting things to work!!

Yes, I am somewhat surprised myself how much we've accomplished in really getting the ELKS TCP stack not only working, but fine-tuned :)

Thank you!

Mellvik:
Thank you @ghaerr,
> ﻿
> 
> The one thing remaining is the outbound packet size, which - as mentioned yesterday - is still restricted to 512 bytes and is unaffected by window sizes.
> The original reason it was left small is that receiving large packets didn't work, and that the large static kernel buffers use up lots of very limited kernel data segment.
> 
> I finally remember the two real reasons we don't use large send buffers for ELKS TCP: 1) ktcp has to keep a retransmit malloc for every packet sent in the sliding window, in case a retransmit is needed, and 2) ktcp can't repackage an application write into a smaller packet size. The first limitation is a problem because ktcp is already scarce on memory, and and memory usage would triple per connection, for instance, to any host with 3 outstanding packets in a send sliding window. The remote host would likely allow more in their receive window, and we've already written special code to throttle ELKS back to stop retransmits that we had already discarded (that was the ZWP discussion). The second limitation is even worse, because if a remote system were not able to accept packets in our chosen "MTU" (which would actually be the hard-coded application write size), then no communication would be possible at all, since the TCP packet is never fragmented from the application.
> 
> For these reasons, for now, I think we should leave things as they are. We lose a possible slight speed increase, but we gain reliability and the ability to work well with all systems we've tested against (so far). Instead, getting our ftp, telnet (and http) client and servers running well without errors would be a better next step.
> 
Now that you mention it, I do remember we've been down this road before. And I agree with your analysis. For now, this is fine - and I'm assuming dynamic allocation of packet buffer space is just too complicated. 

At some point we may want to revisit the fragmentation issue. Given that fragmentation hardly ever happens even on the Internet anymore, and definitely never happens in a LAN environment (and MTU lower than 1460 is rare), it may make sense to make forced smaller packets an option instead of a rule. When and if the first reason above has been handled.

That said, I was looking for an easy way to temporarily increase (actually change, not necessarily increase) the send packet size for testing. I revisited the 'telnet losing characters' issue yesterday, and found it not only present, but even more pronounced than before, and with a characteristic 512 byte pattern. If changing the max outbound packet size changes that, it may be a clue as to what's going on. The first thing through is to see if I can get it to repeat on QEMU. 

Which I did as I was writing this, and sure enough - if you telnet to elks from the mac (or whatever client), then telnet back to the same machine (or a different one, which becomes somewhat more complicated with the QEMU config), not even a 'ls -l' will come out as it should.
So it seems we're getting closer to the telnet problem thanks to the recent networking updates  - which still may well be in the pty driver … The trace below shows the test. Notice the 'helge$' at the end of the last line - which is part of the shell prompt after ls completed. The second output is what the 'ls' should have looked like.

—Mellvik

-------------------------------------------------------------------------------------------

www9:~ helge$ telnet localhost 2323
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
login: root
# telnet 10.0.2.2
Connecting to 10.0.2.2 (10.0.2.2:23)
Connected


Darwin/BSD (www9.local) (ttys006)

login: helge
Password:
Last login: Tue Nov 16 09:59:58 on ttys000
www9:~ helge$ ls -l 
total 17743832
drwx------    4 helge  staff         128 Nov 18  2015 Applications
drwxr-xr-x    4 helge  staff         128 Dec 23  2011 Beth Hart
drwxr-xr-x    4 helge  staff         128 Jul 30 22:19 Chris Botti
drwxr-xr-x    8 helge  staff         256 Oct  7 11:53 Consensys
drwx------@ 277 helge  staff        8864 Oct 22 14:20 Desktop
drwx------@  91 helge  staff        2912 Nov 16 09:40 Documents
-rw-------    1 helge  staff    25000448 Oct 17  2019 DosTest [6.22].ova
drwx------+   6 helge  staff         192 Jul 31 12:45 Downloads
drwx------@  71 helge  staff        2272 Apr  8  2021 Library
drwx------+  14 helge  staff         448 Jan 11  2020 Movies
drwx------+   9 helge  staff         288 Oct  7 11:53 Music
drwx------+   3 helge  staff          96 Nov 16  2014 Pictures
drwxr-xr-x+   5 helge  staff         160 Nov 16  2014 Publhelge$ 

——————————————————————————————————————————————

www9:~ helge$ ls -l 
total 17743832
drwx------    4 helge  staff         128 Nov 18  2015 Applications
drwxr-xr-x    4 helge  staff         128 Dec 23  2011 Beth Hart
drwxr-xr-x    4 helge  staff         128 Jul 30 22:19 Chris Botti
drwxr-xr-x    8 helge  staff         256 Oct  7 11:53 Consensys
drwx------@ 277 helge  staff        8864 Oct 22 14:20 Desktop
drwx------@  91 helge  staff        2912 Nov 16 09:40 Documents
-rw-------    1 helge  staff    25000448 Oct 17  2019 DosTest [6.22].ova
drwx------+   6 helge  staff         192 Jul 31 12:45 Downloads
drwx------@  71 helge  staff        2272 Apr  8  2021 Library
drwx------+  14 helge  staff         448 Jan 11  2020 Movies
drwx------+   9 helge  staff         288 Oct  7 11:53 Music
drwx------+   3 helge  staff          96 Nov 16  2014 Pictures
drwxr-xr-x+   5 helge  staff         160 Nov 16  2014 Public
-rw-r--r--@   1 helge  staff       50953 May  7  2020 Skattemelding2019tore.pdf
-rw-r--r--@   1 helge  staff       92452 May  7  2020 Skatteoppgjørkari2019.pdf
drwxr-xr-x    5 helge  staff         160 Oct 17  2019 VirtualBox VMs
-rw-r--r--    1 helge  staff  8060338437 Jul 20  2019 helge-hdsound.tjz
-rw-r--r--    1 helge  wheel      174188 Nov 20  2014 itunes.bad
-rw-r--r--    1 helge  staff       24945 Aug  5  2020 ne2k-mac.S
-rw-r--r--    1 helge  staff   910174999 Sep  3  2019 raspi-radio-full.tjz
drwxr-xr-x    9 helge  staff         288 Nov 10 14:42 src
drwxr-xr-x    6 helge  staff         192 Nov 11 10:20 tmp
-rw-r--r--    1 helge  staff    21411840 Feb  6  2016 venix-full.img
-rw-r--r--    1 helge  staff    21411840 Feb 18  2016 venix-hd.img
drwx------   13 helge  staff         416 Mar  5  2016 venix21
-rwxr-xr-x    1 helge  staff    15974400 May 16  2017 venix86hd.chd
-rwxr-xr-x    1 helge  staff    11161600 Feb  5  2016 venixhd.img
www9:~ helge$

ghaerr:
@Mellvik, I've opened a seperate issue to discuss the telnet data loss in #1012.

> That said, I was looking for an easy way to temporarily increase (actually change, not necessarily increase) the send packet size for testing.

Changing TDB_WRITE_MAX in elks/include/linuxmt/tcpdev.h from 512 to another number should allow you to play with various send packet sizes. (You'll need to `make kclean` to rebuild the kernel entirely afterwards). I'm not sure everything is tagged to that define, but the kernel will break up application writes to that size, as that's the maximum amount of data that can be sent to the /etc/tpcdev device which then transfers the data to ktcp. 

Ktcp gets the data in elkscmd/ktcp/tcpdev.c in the function tcpdev_write. It compares the current receive window there to TCP_SEND_WINDOW_MAX which is 1024, so that would have to increase if you set TDB_WRITE_MAX higher than 1024. You'll want to turn on DEBUG_TCP and ^P to see when/if ktcp limits sending in a busy loop as per the discussion on ZWP some time ago.

I'd be interested to hear the results of playing around with this.

PR #1008, opened 2021-11-16
[kernel] Fix unreal mode and XMS buffers on real 386 hardware

ghaerr:
Finally got a chance to test booting an XMS-enabled kernel on a 386 desktop machine. It died in the partition code just like @Mellvik's testing found.

Found two possible problems, the A20 verification code improperly returned the wrong value (indicating always enabled), which likely skipped the following actual A20 enable method. Kernel now boots and operates on 386 desktop, saving 64K in main memory kernel buffers, and uses 128K in XMS memory. Runs great!

@Mellvik, hopefully this will work better on your Compaq. If not, it should at least fail properly and not enable XMS if A20 not functional.

### DIFF
--- a/elks/arch/i86/lib/unreal.S
+++ b/elks/arch/i86/lib/unreal.S
@@ -145,14 +145,14 @@ verify_a20:
 
 	popw	0
 
-	popf			# restore interrupt status
-	pop	%es
-	pop	%ds
 	jz	1f		# if ZF=1, the A20 gate is NOT enabled
 	mov	$1,%ax		# return 1 if enabled
+9:	popf			# restore interrupt status
+	pop	%es
+	pop	%ds
 	ret
 1:	mov	$0,%ax		# return 0 if disabled
-	ret
+	jmp	9b
 
 #
 # enable/disable A20 gate using keyboard port/controller, entry AH=0 disable
--- a/elks/arch/i86/mm/xms.c
+++ b/elks/arch/i86/mm/xms.c
@@ -27,7 +27,7 @@ static long_t xms_alloc_ptr = 0x00100000L;	/* 1M */
 /* try to enable unreal mode and A20 gate. Return 1 if successful */
 int xms_init(void)
 {
-	if (enable_unreal_mode()) {
+	if (enable_unreal_mode() > 0) {
 		if (/*verify_a20() ||*/ enable_a20_gate()) {
 			xms_enabled = 1;	/* enables xms_fmemcpyw()*/
 
--- a/elkscmd/sys_utils/unreal16.S
+++ b/elkscmd/sys_utils/unreal16.S
@@ -40,6 +40,7 @@ sys_write = 4
 syscall = 0x80
 videoline = 0xb8000+80*2*22
 
+    .arch	i386,nojumps
     .code16
     .text
 	.extern	enable_unreal_mode


Mellvik:
@ghaerr,
I finally got around to test this - and it seems I'm stuck.

I'm still getting the 'hang during boot' situation when using the A20_ASM define (the console message being positive, 'unreal mode and A20 enabled'), while the HIMEM_AT variant gives the 'can't enable A20 gate' message and boots normally.

In the latter case, I can run ureal16, which loops forever but is interruptible.

This doesn't make sense. I've run MSDOS 6.22 MEM and MEMMAKER commands, and they access all memory just fine - indicating that standard A20 mechanisms should indeed work.

Anything else I can check out here?

—Mellvik

ghaerr:
> I'm still getting the 'hang during boot' situation when using the A20_ASM define (the console message being positive, 'unreal mode and A20 enabled'), while the HIMEM_AT variant gives the 'can't enable A20 gate' message and boots normally.

This sounds like A20_ASM is working as far as enabling the A20 gate, and HIMEM_AT is saying it cannot enable A20, and thus doesn't work, but boots, since A20 is turned off.

> In the latter case, I can run ureal16, which loops forever but is interruptible.

Probably a case where HIMEM_AT is hanging in the A20 enable code, but you can ^C out.



> This doesn't make sense. I've run MSDOS 6.22 MEM and MEMMAKER commands, and they access all memory just fine - indicating that standard A20 mechanisms should indeed work.

I am guessing what is happening is that A20_ASM is working, and that A20 is enabled. The XMS kernel starts writing to address 1M with 128 or 256 buffers immediately upon any I/O, such as the partition table read, where I think you said it's dying. 

You had said that Compaq did some strange stuff like copy the BIOS to high memory for speed, etc. If that BIOS is at 1M, then we'd be overwriting it and/or the Compaq BIOS may be using that memory for its own uses, for some reason. I think we need to learn more about exactly what the Compaq BIOS does with using high memory for itself, and then have the ELKS kernel start the XMS buffers at a different address. This explains what is going on, it seems. I previously tried to research about the Compaq BIOS, but found nothing.

> Anything else I can check out here?

Yes - do you have any other 386 systems this can be tested on? I'll test on my Compaq 386/III Portable, see what happens, and report back.

ghaerr:
@

ghaerr:
@Mellvik, my Compaq Portable 386 does NOT boot with XMS. It also says 'unreal mode and A20 enabled' then crashes just before the VFS mount shown below:
```
fd: /dev/fd0 probably has 18 sectors, 2 heads, and 80 cylinders
```
Other than trying on a few other PCs you may have, I will go ahead and try to add some more debug display code to see whether A20 is already enabled at boot (which I think it is on Compaq possibly), and also perhaps add some code to test XMS memory before using it, selectable from /bootopts, so we can better pin down what is happening (short of getting Compaq documentation on BIOS).

My system that does boot ELKS XMS is 1990's vintage desktop 386 with an ASUS BIOS.

ghaerr:
@Mellvik,

I finally found something about the Compaq's support for extra memory modules. They had a patented CEMM (Compaq Expanded Memory Manager) that used V86 (virtual 8086) mode in the 386 processor to implement EMS (expanded) memory in DOS at the time. IIRC EMS was more of a special bank-switched style memory area around 64K in size, that could be programmatically changed to allow lots of memory in the sub 1M address space. The info on CEMM mentions that the early versions didn't support UMBs, only EMS (likely from the complicated EMS specification). Thus, the memory didn't automatically map into UMB memory locations that DOS might easily use, but only interfaced through the more complex EMS spec.

Here's a couple articles about CEMM:
https://www.os2museum.com/wp/original-cemm-unearthed/
https://en.wikipedia.org/wiki/CEMM
https://en.wikipedia.org/wiki/EMM386

While ELKS's XMS mode won't work with V86, and is supposed to check for V86 mode and fail, that isn't probably the whole problem of why this isn't working. I suspect there's more hardware trickiness on this early system, built especially for EMS and DOS, rather than for the straight linear addresses desirable when Win32 was in widespread use. I don't think this system every ran Windows 386, but instead only the early 3.x versions, but I could be wrong. The EMM386 article above gets into some of this.

Until we learn more about the internals of CEMM or the Compaq memory hardware, getting ELKS XMS to work on this system is probably going to have to be put on hold. It might be as simple as that the "linear" hardware memory that CEMM uses starts at a different address than 1M, or it may have to be switched in and out with special hardware registers and a driver. I'm beginning to think it requires a driver that maps the memory to special locations, that would answer your statement about the BIOS being copied into RAM (at real memory segment F000) for speed, as well as how EMS worked by mapping into UMB blocks.

Mellvik:
> You had said that Compaq did some strange stuff like copy the BIOS to high memory for speed, etc. If that BIOS is at 1M, then we'd be overwriting it and/or the Compaq BIOS may be using that memory for its own uses, for some reason. I think we need to learn more about exactly what the Compaq BIOS does with using high memory for itself, and then have the ELKS kernel start the XMS buffers at a different address. This explains what is going on, it seems. I previously tried to research about the Compaq BIOS, but found nothing.
> 
I guess it's more unusual than strange. The machine can have max 10M physmem, so the area up to 16M is useless. So what the bios upstart does is to remap the physical memory area from 640 to 1M to the top of the addressable 16M. Then it copies itself to the top of this area, making itself available in protected mode, and makes the rest of the 386k available for caching, buffers etc. - from DOS via their own drivers. Kinda neat, and it shouldn't affect what we're trying to do since it doesn't affect dos..
The thing is, though: the machine still offers upper memory to DOS, so part of the remapping is either transparent or reversable.
> Anything else I can check out here?
> 
> Yes - do you have any other 386 systems this can be tested on? I'll test on my Compaq 386/III Portable, see what happens, and report back.
> 
No, this is what I got, a few compaq portable 386es an a couple of the 286 model, some of them under (very slow) repair...

Thank you.
-Mellvik

ghaerr:
> The machine can have max 10M physmem, so the area up to 16M is useless,

What do you mean, I don't follow? What does the 16M boundary have to do with anything?

> Then it copies itself to the top of this area, making itself available in protected mode, and makes the rest of the 386k available for caching, 

I'm not following you here either. Where is the "extra" 386K exactly?

I can see that it would need to use protected mode to access itself up in extended memory, but there'd still need to be protected/real mode switching handled in low memory where the BIOS used to be, right?

> So what the bios upstart does is to remap the physical memory area from 640 to 1M to the top of the addressable 16M.

Ok, I think I get this: upper memory is remapped to extended memory, and protected mode (e.g. EMS) is used to access it directly from DOS using V86 mode.

Do you have any information as to where the "protected mode" memory might start? Perhaps it starts at 16M and comes downward?

If we know the physical memory address of a block of memory, then we can use the UNREAL mode to access it. It sounds like that start address might be different based on how much memory the machine has.

> The thing is, though: the machine still offers upper memory to DOS, so part of the remapping is either transparent or reversable.

The articles mention that V86 mode is used to do this. It all got quite tricky when having to interface with other protected mode programs or drivers. But ELKS doesn't need any of that. All we need is to know where some real memory is (and A20 on, which I think is already the case).

I'm still a bit unclear on this, but we might be able to make this work yet, if we can get a hardware physical address where there is memory. (not 1M!)

Thank you!

Mellvik:
Apologies @ghaerr, I forgot to paint the picture before discussing its contents.


> 18. nov. 2021 kl. 22:11 skrev Gregory Haerr ***@***.***>:
> 
> 
> The machine can have max 10M physmem, so the area up to 16M is useless,
> 
> What do you mean, I don't follow? What does the 16M boundary have to do with anything?
> 
The compaq - like most early 386 machines - had address lines to physically address 16M RAM (24 lines IIRC). That's the ceiling of addressability so to speak. Since the machine (the portable 386) can hold 10M physical memory, the upper 6M are 'useless'.
> Then it copies itself to the top of this area, making itself available in protected mode, and makes the rest of the 386k available for caching,
> 
> I'm not following you here either. Where is the "extra" 386K exactly?
> 
This is what I've been describing before (the number is 384k BTW - the upper memory, from 640k to 1M). The compaq remaps the upper memory RAM to the top of the addressable range, just below the 16M limit. The TOP 128K being a copy of the BIOS itself. The remaining 384-128=256 becomes available via the COMPAQ utilities you describe in another post, for RAMdisk, caching, etc. 
> I can see that it would need to use protected mode to access itself up in extended memory, but there'd still need to be protected/real mode switching handled in low memory where the BIOS used to be, right?
> 
yes, and having looked more at it, I don't think this mechanism is more than a curiosity in our context, something to be aware of.
> So what the bios upstart does is to remap the physical memory area from 640 to 1M to the top of the addressable 16M.
> 
> Ok, I think I get this: upper memory is remapped to extended memory, and protected mode (e.g. EMS) is used to access it directly from DOS using V86 mode.
> 
Yes.
> Do you have any information as to where the "protected mode" memory might start? Perhaps it starts at 16M and comes downward?
> 
> If we know the physical memory address of a block of memory, then we can use the UNREAL mode to access it. It sounds like that start address might be different based on how much memory the machine has.
> 
No, it's always the same. Another curiosity is that the compaq can be configured to have less than 640k 'main memory', either 512 or 256, which will expand the RAM available in the top memory area. Useless, but interesting.
> The thing is, though: the machine still offers upper memory to DOS, so part of the remapping is either transparent or reversable.
> 
> The articles mention that V86 mode is used to do this. It all got quite tricky when having to interface with other protected mode programs or drivers. But ELKS doesn't need any of that. All we need is to know where some real memory is (and A20 on, which I think is already the case).
> 
> I'm still a bit unclear on this, but we might be able to make this work yet, if we can get a hardware physical address where there is memory. (not 1M!)
> 
I'll post some drawings from a booklet I got on eBay a while back (Compaq Portable 386 Service Training Workbook), describing among other things the memory arrangements.

More comments in a separate post.

Thank you.

—Mellvik

Mellvik:
@ghearr,

this is very interesting indeed. Fills in big holes and refreshes ancient experiences. I guess we've all used various EMM386 products over the years - I didn't realize that compaq literally defined the 'standard' so to speak.

Thank you for the links - very enlightening!

> 
> I finally found something about the Compaq's support for extra memory modules. They had a patented CEMM (Compaq Expanded Memory Manager) that used V86 (virtual 8086) mode in the 386 processor to implement EMS (expanded) memory in DOS at the time. IIRC EMS was more of a special bank-switched style memory area around 64K in size, that could be programmatically changed to allow lots of memory in the sub 1M address space.
> 
Exactly - this is the LIM (Lotus/Intel/Microsoft) stuff, which works on the 286 too.
> The info on CEMM mentions that the early versions didn't support UMBs, only EMS (likely from the complicated EMS specification). Thus, the memory didn't automatically map into UMB memory locations that DOS might easily use, but only interfaced through the more complex EMS spec.
> 
Sounds right - it was a big deal when we actually could start using the UMBs for drivers and (maybe ) buffers.
> Here's a couple articles about CEMM:
> https://www.os2museum.com/wp/original-cemm-unearthed/ <https://www.os2museum.com/wp/original-cemm-unearthed/>
> https://en.wikipedia.org/wiki/CEMM <https://en.wikipedia.org/wiki/CEMM>
> While ELKS's XMS mode won't work with V86, and is supposed to check for V86 mode and fail, that isn't probably the whole problem of why this isn't working. I suspect there's more hardware trickiness on this early system, built especially for EMS and DOS, rather than for the straight linear addresses desirable when Win32 was in widespread use. I don't think this system every ran Windows 386, but instead only the early 286 versions, but I could be wrong.
> 
I'm unsure about that, but that can be tested.  Like booting Solaris 386 (early versions) or some other Unix386 assuming linear memory in virtual mode. What that will tell is whether the mechanisms we're talking about are generic or compaq special. The way DOS 6.22 works indicate that what we're trying to do should be possible.
> Until we learn more about the internals of CEMM or the Compaq memory hardware, getting ELKS XMS to work on this system is probably going to have to be put on hold. It might be as simple as that the "linear" hardware memory that CEMM uses starts at a different address than 1M, or it may have to be switched in and out with special hardware registers and a driver. I'm beginning to think it requires a driver that maps the memory to special locations, that would answer your statement about the BIOS being copied into RAM (at real memory segment F000) for speed, as well as how EMS worked by mapping into UMB blocks.
> 
There may certainly be some trickery here, but again the fact that DOS MEM and MEMMAKER works and finds both UBMs and HIMEM makes me hopeful.

Pictures coming up in a separate post.

—Mellvik

Mellvik:
[Compaq memory config.pdf](https://github.com/jbruchon/elks/files/7569649/Compaq.memory.config.pdf)

ghaerr:
Hello @Mellvik,

Fantastic information in the Compaq Service Training Workbook!! They mention "how to use" the memory on Pages 30-31.
Is there any hardware access information on those pages?

I performed more testing by adding a new function that uses unreal mode to read 1 word of RAM on every megabyte boundary from 0-15, just to see what might be there. This happens in the ELKS boot process just after enabling unreal mode and XMS. Funny thing is that the function hangs immediately at the 1MB boundary. It seems there might be some additional hardware port/mechanism required to turn on either A20 properly, or perhaps to enable the extended memory at all. This could be a BIOS function call or hardware I/O port, gate, something like that. Can you see if the book has any information about that kind of thing?

My machine has 2M memory. That means there should be 1M available at the 1M boundary (this is hanging when read directly using unreal mode with my function above), and there should be an additional 256K available at the 15.6M boundary, which is 0xFA00000L. I will try reading that and report results.

There is some LIM/bank switching hardware I/O ports or whatever that we need more information on. It seems the hardware has the ability to remap memory, and that might not be setup at boot, since normal DOS users will load the CEMM utility to handle it.

We're making progress!

Mellvik:
[Mer compaq.pdf](https://github.com/jbruchon/elks/files/7573151/Mer.compaq.pdf)

Mellvik:
I don't think the last couple of pages were particularly useful, @ghaerr , but yes, we're making progress.

Mellvik:
BTW @ghaerr - 
Re. A20 handler:

Have you seen this HIMEM.ASM source?
https://github.com/MikeyG/himem/blob/master/oemsrc/himem.asm
With a binary of this one we could test whether the A20 mechanism used, works.

--M

ghaerr:
> Have you seen this HIMEM.ASM source?

Yes, the ELKS USE_HIMEM_AT is from that file. It doesn't work on the Compaq. I've managed to verify that the USE_A20_ASM does work.

After reading the information you uploaded in the "Compaq Portable 386 Service Training Workbook" (an incredible find, BTW), I searched heavily online and am not able to find any reference of it, nor of the very important "Compaq Personal Computer Technical Reference Guide (PN 107920-001)". The latter is referenced in a manual [Compaq Portable 386 Personal Computer Maintenance and Service Guide](https://www.manualslib.com/manual/273076/Compaq-Portable-386.html). That manual describes all the hardware memory switches, and corresponds with some of the information from your manual.

The big find came when I found [this page](https://www.pcjs.org/machines/pcx86/compaq/deskpro386/) which contains the complete Technical Reference Manuals for the Compaq Deskpro 386, which although not for the Portable 386, are from the same vintage (1986) and list an unbelievable amount of information, including 386 processor hardware bugs directly affecting the ELKS unreal implementation (and also included in HIMEME.SYS source) and the complete list of Compaq BIOS entry points.

With that information, I was able to realize that the Compaq BIOS (for all their early 386 computers) are extremely sophisticated and make use of real-to-protected mode switching all the time for access to extended memory and other things. I was able to debug exactly where the system is hanging in the ELKS XMS implementation, which very strangely enough is immediately upon using one of the 32-bit linear descriptors setup for unreal mode access to any (not just extended) memory. 

You may recall that the way unreal mode works is that the processor is put in protected mode, and a new GDT is loaded into the 386 GDT register, which is then used to specify 4GB range for a special "linear" selector. Protected mode is then turned off, but the 386 doesn't change the cached limit for that selector, and a 32-bit address override can be used with DS:ESI or ES:EDI to access any memory in the machine.

I am not able to understand (yet) why the processor seems to halt/crash upon use in real mode of this mechanism on the Compaq, even with interrupts off. I suspect that the system can't handle a "user" re-loaded GDT, or something like that.

Rather than give up, or figure out exactly why this is happening, I think that it is probably too dangerous for ELKS to depend on unreal mode always working, when the BIOS is known to implement protected mode itself, which requires reloading of the GDT register. That would likely cause failure as soon as any BIOS routine did it, and ELKS is still dependent on BIOS for disk I/O.

I noticed in the tech ref manual that the Compaq has an "extended memory block move" BIOS entry point, which allows copying XMS memory to/from main memory. Since both of us have Compaq Portables, I decided to move forward with an additional redesign of the ELKS XMS implementation that uses this BIOS entry point (for Compaq systems only) instead of the kernel setting unreal mode and A20 itself. I hope to have that working very shortly.

The only hangup I see with the BIOS block move approach is that the manual states that interrupts are turned off for the entire move (of 512 words). This will likely wreak havoc with our serial interrupt handling. A possibly workaround will be to break up the XMS moves into smaller sections, which although slower, would still be lots faster than disk I/O.

Mellvik:
@ghaerr,
This is great research!
> I noticed in the tech ref manual that the Compaq has an "extended memory block move" BIOS entry point, which allows copying XMS memory to/from main memory. Since both of us have Compaq Portables, I decided to move forward with an additional redesign of the ELKS XMS implementation that uses this BIOS entry point (for Compaq systems only) instead of the kernel setting unreal mode and A20 itself. I hope to have that working very shortly.
> 
I like this approach. As soon as SOMETHING is working, we can start poking around at various addresses, like you alluded to before.
> The only hangup I see with the BIOS block move approach is that the manual states that interrupts are turned off for the entire move (of 512 words). This will likely wreak havoc with our serial interrupt handling. A possibly workaround will be to break up the XMS moves into smaller sections, which although slower, would still be lots faster than disk I/O.
> 
It seems to me that 2k bytes memory to memory on a 386 @ 20 MHz should be fast enough not to affect serial @ (max) 38400. I didn't do the math though... and it depends on the quality of the bios code...

Looking forward to it ...

-Mellvik

PR #1015, opened 2021-11-17
8018x: IRQ-driven serial console + header for platform

cocus:
Adds a platform-dependant .h file with all the "PCB" (peripheral control block) port addresses, so they can be used via their names rather than hardcoding their offsets.
Modifies the irq and timer platform sources to make use of these new PCB defines.
Adds a new IRQ-driven serial console (defaults to 9600bps) with a new console type "8018x".


ghaerr:
@cocus,

This looks fantastic, thank you for your great work on this!

Can you do one last change: update the 8018x.config file so that it matches the one you're testing with? Or at least change the CONFIG_CONSOLE_HEADLESS to CONFIG_CONSOLE_8018X, etc. I don't think the current checked-in version is created from make menuconfig, but I would like to be able to compile your version using 8018x.config for regression testing.

Thank you!

ghaerr:
I also just happened to notice another small bug, the ioctl entry point returns -EINVAL on the valid TCGETX functions:
```
static int sercon_ioctl(struct tty *tty, int cmd, char *arg)
 {
     struct serial_info *port = &ports[0]; /* TODO: add support for Serial 1 */

     switch (cmd) {
     case TCSETS:
     case TCSETSW:
     case TCSETSF:
         update_port(port);
     break;

     default:
         return -EINVAL;
     }

     return -EINVAL;
```
The last line should be changed to "return 0;"

cocus:
> Can you do one last change: update the 8018x.config file so that it matches the one you're testing with? Or at least change the CONFIG_CONSOLE_HEADLESS to CONFIG_CONSOLE_8018X, etc. I don't think the current checked-in version is created from make menuconfig, but I would like to be able to compile your version using 8018x.config for regression testing.

Will do! I forgot about that. Do you think it would make sense to edit the current file, or copy my current .config file on it?



> The last line should be changed to "return 0;"

Good catch! Didn't notice it!

ghaerr:
I suggest the edit the file and add differences.

PR #1018, opened 2021-11-21
[kernel] Add XMS support for Compaq Portable 386

ghaerr:
Adds Compaq Portable 386 to the hardware supported for the recently added extended memory buffers.

The system must be built with an additional 'Use BIOS INT 15H instead of unreal mode' set, in addition to 'XMS buffer cache' being set in config. The default number of (1K) buffers is 256, which really speeds things up. BIOS INT 15 is a built-in BIOS function that copies physical memory to/from any address. It seems to be present in almost all 386 systems, and also works in QEMU. I suspect it is slower than the ELKS unreal mode, which doesn't require interrupts to be disabled during XMS transfers. However, the Compaq BIOSes don't seem to tolerate unreal mode, so this option is necessary.

A later option may try to enable XMS and/or INT 15h automatically, but for now each must be explicitly configured into the kernel.

@Mellvik, this should work fine on your Compaq portables and is well tested on mine. If you want to really see some speed, try running 'net start' followed by 'net stop', and then repeat (since after that everything is buffered). The system can then start or stop the network in under a second.

Note: since this is new code, and interrupts are disabled during the XMS transfers, please don't use a kernel built with this config option when testing the open networking issues. However, it would be interesting to know whether ethernet networking suffers much with during XMS I/O separately.


Mellvik:
@ghaerr,
Now isn't that a piece of great sunday morning news! I can't wait to see how this affects the overall performance.
And yes, I'll keep it out of the loop while debugging other issues. 

A couple of questions:
Where did you put the buffers? Just above 1M?
Did you add code to poke the first byte of every M like you mentioned?
Now that this works, what does the unreal16 program do?
Btw - good to have you back on real hardware :-) 

-M

Mellvik:
@ghaerr,

Your reference to INT 15 sent me on a google chase - which is always dangerous, in particular when the pile of other things to do keeps growing.

Anyway, it seems (this was news to me btw) INT 15h is a rather broad utility covering cassette tapes, mouse drivers, extended memory and much more. What caught my attention in particular was the call to map all available memory, which I suspect you’re already aware of:

http://www.uruk.org/orig-grub/mem64mb.html <http://www.uruk.org/orig-grub/mem64mb.html>

Now that we’re playing with ‘big’ (in elks terms) memory, this would be really useful AFAIK.
Also, it will be rather interesting to see what the Compaq 286 BIOS does about this. I have 6MB (Compaq original) memory on one of the 826 systems.

Thank you.


—Mellvik

Mellvik:
This is very encouraging, @ghaerr!!

Worked right out our the box as you described, net stop/net start - and a lot more works really fast when buffered. An entirely new experience - in particular since the starting point is a floppy based system.

One and two concurrent telnets also work well, I can't see any adverse effects so far.

Congrats ! This is great!

—M

ghaerr:
> Where did you put the buffers? Just above 1M?

Yes. I've also tested it at 15.6M (in the Compaq additional memory) and it works there too. However, that address is not as compatible as using 1M. The INT 15 block move BIOS entry point is available on almost all 386 systems, so using it turns out to be pretty compatible, although slower than using unreal mode. At some point, we may need a config option as to which memory to use. 

> Did you add code to poke the first byte of every M like you mentioned?

No, this version doesn't do any memory testing, and without using unreal mode, there would be quite a bit of code written using the block move function just to test memory. I have an unreal mode version of peek/poke though. There isn't any real need to test memory yet, as almost every 386 has 256K at 1M.

> Now that this works, what does the unreal16 program do?

On Compaq Portables, it crashes immediately, just like before. As mentioned prior, for some reason touching any memory in unreal mode, even with A20 enabled crashes. I have A20 enabling figured out, and may continue to try to debug unreal mode on the Compaq, but with its sophisticated BIOS, interrupts would have to be off anyways. That's why I went the INT 15 route.

> Btw - good to have you back on real hardware :-)

LOL, yes, lets just say this didn't work the first time!

> it seems (this was news to me btw) INT 15h is a rather broad utility covering cassette tapes, mouse drivers, extended memory and much more.

Yes, INT 15 is used for "system" BIOS enhancements.

> What caught my attention in particular was the call to map all available memory, which I suspect you’re already aware of

Yes, that BIOS entry point is not available on the early Compaq BIOS. After finding and reading the Compaq DeskPro Technical Manual circa 1986, the INT 15 we're using is one of only two extended memory functions that are implemented on the Compaq. The other returns the amount of contiguous memory above 1M.

In general, though, I'm trying to stay away from BIOS, and run the kernel without being dependent on it. That's why unreal mode was developed first - the INT 15 driver was an addition just to get the Compaq running, since its BIOS isn't compatible with unreal mode (later BIOSes seem to be).

>. Now that we’re playing with ‘big’ (in elks terms) memory, this would be really useful AFAIK.

We're about out of memory in the kernel data segment, just to track the 256K of extended memory. But with the addition of the dynamic pipe buffers and possible other dynamic kernel enhancements, we might be able to support more.

There are all sorts of potentially fun things to do, but you end up fighting PC incompatibility, especially on older systems. A fun, but time consuming project would be to run all of ELKS in Virtual 8086 mode on a 386, for instance. But why?! :) ELKS could also be rewritten to use 16-bit selectors rather than real mode segments, and run in 16-bit protected mode. I ask them same question: why? The kernel needs to maintain compatibility with real mode 8086 processors, and fit in tiny spaces for all non-386 systems.

> Also, it will be rather interesting to see what the Compaq 286 BIOS does about this. 

I have no idea whether the BIOS implements INT 15 block move, and there's no way to tell other than booting it. The INT 15 config option does not make a processor check (yet). If you could find a technical manual for the system that explained it's BIOS entry points, that would be great.

> Your reference to INT 15 sent me on a google chase - which is always dangerous, in particular when the pile of other things to do keeps growing.

Yes, it is dangerous, I want to make sure we finalize our outstanding networking issues before going elsewhere :) Thank you for testing this PR though, I'm glad it works!

Mellvik:
@ghaerr,
 somewhere in between the previous post in this thread (nov 21st) and today, XMS buffering was broken (on real hardware). Enabling XMS i menuconfig with all the latest changes pulled in, cause `elks` to hang during boot.

I haven't made any attempt to 'bisect' into this to fint the culprit, I might do that unless you have some immediate hunch as to what/where it is.
The picture shows where it stops, the next expected output would be the partition data for hda1.

IMG_7388

ghaerr:
Hi @Mellvik,

> Enabling XMS i menuconfig with all the latest changes pulled in, cause elks to hang during boot.

On the Compaq series machines, "unreal mode" (as see in the xms: boot line) won't work. The CONFIG_FS_XMS_INT15=y (BIOS XMS transfer) option must be set in .config. I see that I didn't include that as an option in config.in, I'll submit a PR for that.

The Compaq BIOS already uses protected mode internally and interferes with unreal mode.

Thank you!

ghaerr:
@Mellvik,

Oops, there is already an option for Compaq and other machines requiring INT15: "Use BIOS INT 15H instead of unreal mode" in config.

Mellvik:
Thank you, @ghaerr -
I Should have remembered - obvious memory leak!

-M

Mellvik:
Followup on this one, @ghaerr:

Would it be hard to delay the console message about unreal mode and XMS buffers so that it goes to the serial console if enabled?


—M

ghaerr:
> Would it be hard to delay the console message about unreal mode and XMS buffers so that it goes to the serial console if enabled?

I'll look into this - the buffer initialization is one of the first kernel init's, but if it could be delayed until after /bootopts, then xms/ext and buffer counts could become a boot option also, which would be nice.

PR #1019, opened 2021-11-21
[kernel] Add getsockname, getpeername sys calls, fix accept

ghaerr:


ghaerr:
Adds `getsockname` and `getpeername` system calls.
Fixes `accept` to return IP address and port number connected to.

`getsockname` can be called after `bind` in get local host IP and bound port number. This gets rid of the need for the slower mechanism of using `in_gethostbyname` with a magic hostname.

Examples commented out in elkscmd/inet/telnetd/telnetd.c.

Considerable cleanup in kernel socket structure. Managed to add features without increasing static inode table size.
Removes CONFIG_SOCK_CLIENTONLY.

Issue #1016, opened 2021-11-18
[net][ftpd] Challenges with FTP passive mode


Mellvik:
While working on **passive mode** for `ftpd` - which is taking a lot longer than anticipated, I've come across a number of issues that need to be discussed.

First of all - some background.
FIle transfers with `ftp` come in two flavours. FTP operates via two channels or connections: The command channel which is open for the duration of the session and passes commands and status responses back and forth, usually between an interactive client and a server. 
When the client wats to send or receive data, a data channel is opened for that particular transfer, normally a file or a directory listing. In the old days, when the Internet was 'open', the client would tell the server
'I have opened a socket on address/port a,b,c,d,x,y' where a,b,c,d is the IP address in 4 decimal numbers and 'x,y' is the port number, x*256+y. All in ASCII, easy to see, send, decode etc. This is know as **active mode**: The server connects back to the client for a transfer. When the transfer is complete (or has an error), the data channel is closed. If we're transferring 100 or 10.000 files, a new connection is opened per file. Simple, if not very efficient. 

Active mode works fine in a LAN environment, but not on the Internet: In todays NAT-oriented internet, the client is normally hidden behind a NAT router, possibly a firewall, and connecting to it from the outside is - or should be - impossible. Thus **passive mode**, which is just like active mode, but the opposite. The sever provides the client with a point to connect to for the data transfer:
'I've opened a socket on address/port a,b,c,d,x,y, ready for you to use' and the client initiates the connection. No problems with NAT or firewalls - if the port number is acceptable, which it normally is, but that's a different story.

When opening sockets on Unix/Linux systems, probably others too, a program/server normally leaves to the OS to add local addresses, some times also port numbers for the connection. This is convenient for a number of reasons, relieving the application from keeping track of interfaces and their addresses. On the other hand, in some cases the program needs to find out what address (and port #) the OS assigned to a connection. Such as with  our Passive Mode, where we're supposed to tell the client in clear text where to connect - address and port.

On most  systems, this is easy - such as calling `getsockname` on the socket.  But ELKS does not have this function. How do we implement PASV (technical name for Passive Mode)? Either we implement `getsockname`, which I discussed with @ghaerr and turns out to be a major effort in both code and time, use hard-coding (not really acceptable) or use the `/etc/hosts` file and the `in_gethostbyname` library routine added by @ghaerr last year.

[...]

PR #1024, opened 2021-11-22
[net] Debugging code for CLOSE_WAIT bug

ghaerr:
Adds debugging printf code to see what's happening in https://github.com/jbruchon/elks/issues/1016#issuecomment-975753093.

@Mellvik, this should automatically display exactly what is happening with the undesired CLOSE_WAIT state after closing a socket with data remaining. The closing state transitions taken are all displayed. 

Please capture the results and post. No ^P required.

### DIFF
--- a/elkscmd/inet/nettools/netstat.c
+++ b/elkscmd/inet/nettools/netstat.c
@@ -28,7 +28,7 @@
 #include <ktcp/tcp.h>
 #include <ktcp/netconf.h>
 
-char tcp_states[11][13] = {
+char *tcp_states[11] = {
 	"CLOSED",
 	"LISTEN",
 	"SYN_SENT",
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -55,7 +55,8 @@ void dprintf(const char *, ...);
 #endif
 
 #if DEBUG_CLOSE
-#define debug_close	DPRINTF
+//#define debug_close	DPRINTF
+#define debug_close	printf
 #else
 #define debug_close(...)
 #endif
--- a/elkscmd/ktcp/tcp.c
+++ b/elkscmd/ktcp/tcp.c
@@ -276,6 +276,7 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 
     if (h->flags & TF_FIN) {
 	cb->rcv_nxt++;
+	debug_close("tcp[%p] packet in established, fin: 1, data: %d, setting state to CLOSE_WAIT\n", cb->sock, datasize);
 	cb->state = TS_CLOSE_WAIT;
 	cb->time_wait_exp = Now;	/* used for debug output only*/
 	debug_tune("tcp: got FIN with data %d buffer %d\n", datasize, cb->buf_used);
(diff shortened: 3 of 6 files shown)


Mellvik:
Thank  you @ghaerr,

here's the first round - just one single test, see print below. There is a mix of ftpd debug prints, I added some comments - hopefully contributing to making the output more understandable. `ftpd` is running in passive mode.

```
getsockname: adr 10.0.2.15, port 49821
Accept                        <--- data connection established
tcpdev release: close socket 00a27e, state is LISTEN          <---- listen) socket released
cmd: 'RETR ls' 13
RETR: <ls> fd 6                  <--- FTP retrieve commend, asking ELKS to send the file 'ls'
RETR error fd 6 len 1500
Data write error: Broken pipe           <----- send fail immediately, this is the s.c. accept-problem, we'll leave that for later
tcpdev release: close socket 00a2ce, state is 
tcp[006f00] NO state change from  on release   <--- at this point, ftpd closes the data socket, not data has made it to the wire
cmd: '????ABOR' 33               <--- the client responds to ^C and sends ABOR (I haven't looked into what the extra chars in front of the ABOR' are yet.)
tcp[000000] packet in established, fin: 1, data: 0, setting state to CLOSE_WAIT
cmd: 'QUIT' 6            <--- the client logs out and disconnects
Quitting...
Child process exiting...
tcpdev release: close socket 00a1de, state is ESTABLISHED   <--- this is the command socket
tcp[00a1de] setting state to FIN_WAIT_1
tcp[00a1de] packet in fin_wait_1, fin: 0
tcp[00a1de] set state CLOSED
tcpdev release: close socket 00a40e, state is ESTABLISHED  <---- this is netstat
```

It is  quite possible that this is somehow related tothe sc. accept() problem, which I've spent more time on today without getting anywhere. The thing is - it is not dependably repeatable. The above command sequence may fail 5 or 10 times in a row (reboots in between), then succeed. And when it succeeds, the same command repeated will succeed forever until a change is made. Such as changing a directory (CWD) or do a directory list instead of a file transfer etc. But that's the accept-problem ...

ghaerr:
Thanks for the testing, it is definitely a corruption problem:
```
tcp[006f00] NO state change from  on release   <--- at this point, ftpd closes the data socket, not data has made it to the wire
cmd: '????ABOR' 33               <--- the client responds to ^C and sends ABOR (I haven't looked into what the extra chars in front of the ABOR' are yet.)
tcp[000000] packet in established, fin: 1, data: 0, setting state to CLOSE_WAIT
```
The values in the square brackets are the socket values in hex. Notice the null pointer in the last statement, and 6f00 isn't valid either. (The valid values look like 'a1de', higher in the heap). The null pointer is causing ktcp to dereference improperly, and that's the reason it's going into the CLOSE_WAIT state. In other words, nothing wrong with ktcp in that area, its being passed a back socket value for some reason. The bad code is likely in your application , but could also be in some other kernel or ktcp code.

I'm betting that the problem is that something is dereferencing a null pointer, which does undetected on 8086. I would suggest that you post (even temporarily) your ftpd code in a PR so we can have a look at it. We need to go through every line looking for corruption, and backtrace this. This is undoubtably the reason for the ???? in the ABORT message as well.

I'm pretty sure the corruption happens before the "broken pipe" reporting as well. Debug code could be added there to show the socket value to confirm that.

Mellvik:
Sounds good, @ghaerr,

will post the code later today. It's ugly as hell at the moment - full of all kinds of debug stuff, far off a mergeable state anyway.

—M

ghaerr:
Thanks @Mellvik. I'm committing more code that will display more information on the -EPIPE read error, should that pop up again. From the conversation in https://github.com/jbruchon/elks/pull/1027#issuecomment-976931484, it seems like both the close_wait and accept problems in ftpd are due to ftpd data corruption.

Nonetheless, please run this entire test suite tomorrow with the new code, which will include some changes made in the SWS PR, but SWS will be turned off. We're going to delay any further work on SWS until ftpd is working in passive mode, to reduce variables.

After your commit of an ftpd update which works, (including 1 second delay and SO_REUSEADDR unused), we can work on both SO_REUSEADDR and SWS separately.

Thank you!

Issue #1022, opened 2021-11-22
[net] Receive window size timing issues


ghaerr:
Opening new issue for @Mellvik, from https://github.com/jbruchon/elks/issues/1016#issuecomment-975756845.

In short what happens is that
- connection is opened (accept()),
- the client sends 3 full size 1460 packets,
- ELKS ACKs them one by one, shrinking the window along the way,
- to 1 in the 3rd ACK.
Now, this one may be interesting (I need to look more at this later). The window size advertized by the 2nd ACK is 1460 exactly, which is also the size of the next packet payload.
So the next window should arithmetically be 0, but cannot be zero so it's one by definition. If this is accommodated for in the code, no problem, but what happens next is that the client sends one byte, ELKS ACKs one byte, and this continues for a long time - single byte packets -
until ktcpd finally says 'tcp: dropping packet, no buffer space: 1 > 0' 9 times and hangs.

I've seen this behaviour a number of times over the past few days, and I don't know for sure that this is at all related to the ACK/timing issue. My full analysis will compare a transfer that works with one as described. What sent me down the ACK/timing road was the fact that in between the packets on the data channel, there is a packet on the command channel that may rock the boat.

This is a preliminary description - i'm working on a more detailed rundown for you to look at.

Mellvik:
Packet trace for the scenario above:
```
10:54:02.732527 IP (tos 0x10, ttl 64, id 33839, offset 0, flags [DF], proto TCP (6), length 51)
    10.0.2.2.42822 > 10.0.2.15.ftp: Flags [P.], cksum 0x1836 (incorrect -> 0x0ac2), seq 92:103, ack 475, win 29200, length 11: FTP, length: 11
	STOR ftpd
10:54:02.759715 IP (tos 0x0, ttl 64, id 9154, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp > 10.0.2.2.42822: Flags [.], cksum 0x1154 (correct), ack 103, win 4369, length 0
10:54:02.780656 IP (tos 0x0, ttl 64, id 9155, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.ftp > 10.0.2.2.42822: Flags [.], cksum 0x1149 (correct), ack 103, win 4380, length 0
10:54:03.065811 IP (tos 0x0, ttl 64, id 9156, offset 0, flags [none], proto TCP (6), length 88)
[...]

PR #1025, opened 2021-11-22
[net] Implement SWS handling on advertised receive window

ghaerr:
Implements receive-side silly window syndrome handling in ktcp.

@Mellvik, I plan on committing this and you can then test #1022. Basically, ktcp must either have half the buffer size available, or MTU bytes available, whichever is smaller, or the receive window is set to 0. This stops the sending of small packets and is the implementation of Dave Clark's algorithm.

The good news about this change is we no longer have to have an additional unused 512 bytes allocated in the TCP receive buffer, helping conserve more memory in ktcp. The receive buffer is now 3\*1460 = 4380 bytes per socket.

You can use ^P to show the advertised window size, as well as the calculated sms minimum window size and buffer remaining.

Tested on QEMU internally and to remote telnet server with high-speed input. The receive window gets set to 0 which stops the sender for a bit and things seem to operate smoothly.

### DIFF
--- a/elkscmd/ktcp/tcp.c
+++ b/elkscmd/ktcp/tcp.c
@@ -246,7 +246,7 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 	tcpcb_buf_write(cb, data, datasize);
 
 	/* always push data for now*/
-	if (1 || (h->flags & TF_PSH) || CB_BUF_SPACE(cb) <= PUSH_THRESHOLD) {
+	if (1 /*|| (h->flags & TF_PSH) || CB_BUF_SPACE(cb) <= PUSH_THRESHOLD*/) {
 	    if (cb->bytes_to_push <= 0)
 		tcpcb_need_push++;
 	    cb->bytes_to_push = cb->buf_used;
--- a/elkscmd/ktcp/tcp.c
+++ b/elkscmd/ktcp/tcp.c
@@ -238,7 +238,8 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 
 	/* check if buffer space for received packet*/
 	if (datasize > CB_BUF_SPACE(cb)) {
-	    printf("tcp: dropping packet, no buffer space: %u > %d\n", datasize, CB_BUF_SPACE(cb));
+	    printf("tcp: dropping packet, no buffer space: %u > %d\n",
+		datasize, CB_BUF_SPACE(cb));
 	    netstats.tcpdropcnt++;
 	    return;
 	}
@@ -279,7 +280,7 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
 	debug_close("tcp[%p] packet in established, fin: 1, data: %d, setting state to CLOSE_WAIT\n", cb->sock, datasize);
 	cb->state = TS_CLOSE_WAIT;
 	cb->time_wait_exp = Now;	/* used for debug output only*/
-	debug_tune("tcp: got FIN with data %d buffer %d\n", datasize, cb->buf_used);
+	debug_tcp("tcp: got FIN with data %d buffer %d\n", datasize, cb->buf_used);
 	if (cb->bytes_to_push <= 0)
 	    tcpdev_sock_state(cb, SS_DISCONNECTING);
     }
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -26,13 +26,14 @@
  * control block input buffer size - max window size, doesn't have to be power of two
  * default will be (ETH_MTU - IP_HDRSIZ) * 3 = (1500-40) * 3 = 4380
  */
+//#define CB_NORMAL_BUFSIZ	4892	/* bufsize for older receive window algorithm*/
 #define CB_NORMAL_BUFSIZ	4380	/* normal input buffer size*/
 
 /* max outstanding send window size*/
 #define TCP_SEND_WINDOW_MAX	1024	/* should be less than TCP_RETRANS_MAXMEM*/
 
 /* threshold to wait before pushing data to application (turned off for now) */
-//#define PUSH_THRESHOLD	512
+#define PUSH_THRESHOLD	512
 
 /* timeout values in 1/16 seconds, or (seconds << 4). Half second = 8*/
 #define TIMEOUT_ENTER_WAIT	(4<<4)	/* TIME_WAIT state (was 30, then 10)*/
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -24,15 +24,15 @@
 
 /*
  * control block input buffer size - max window size, doesn't have to be power of two
- * default will be (ETH_MTU - IP_HDRSIZ) * 3 + PUSH_THRESHOLD = (1500-40) * 3 + 512 = 4892
+ * default will be (ETH_MTU - IP_HDRSIZ) * 3 = (1500-40) * 3 = 4380
  */
-#define CB_NORMAL_BUFSIZ	4892	/* normal input buffer size*/
+#define CB_NORMAL_BUFSIZ	4380	/* normal input buffer size*/
 
 /* max outstanding send window size*/
 #define TCP_SEND_WINDOW_MAX	1024	/* should be less than TCP_RETRANS_MAXMEM*/
 
-/* bytes to subtract from window size (was remaining buf space to force push)*/
-#define PUSH_THRESHOLD	512
+/* threshold to wait before pushing data to application (turned off for now) */
+//#define PUSH_THRESHOLD	512
 
 /* timeout values in 1/16 seconds, or (seconds << 4). Half second = 8*/
 #define TIMEOUT_ENTER_WAIT	(4<<4)	/* TIME_WAIT state (was 30, then 10)*/
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -26,8 +26,8 @@
  * control block input buffer size - max window size, doesn't have to be power of two
  * default will be (ETH_MTU - IP_HDRSIZ) * 3 = (1500-40) * 3 = 4380
  */
-//#define CB_NORMAL_BUFSIZ	4892	/* bufsize for older receive window algorithm*/
-#define CB_NORMAL_BUFSIZ	4380	/* normal input buffer size*/
+#define CB_NORMAL_BUFSIZ	4892	/* bufsize for older receive window algorithm*/
+//#define CB_NORMAL_BUFSIZ	4380	/* normal input buffer size*/
 
 /* max outstanding send window size*/
 #define TCP_SEND_WINDOW_MAX	1024	/* should be less than TCP_RETRANS_MAXMEM*/
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -390,8 +390,8 @@ void tcp_output(struct tcpcb_s *cb)
 	len = 1;			/* Never advertise zero window size */
 #endif
 
-    debug_tune("tcp output: space %u min sws %u window %u\n",
-	CB_BUF_SPACE(cb), minwindow, len);
+    debug_tune("tcp output: min sws %u space %u window %u\n",
+	minwindow, CB_BUF_SPACE(cb), len);
     th->window = htons(len);
     th->urgpnt = 0;
     th->flags = cb->flags;
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -366,7 +366,7 @@ void tcp_output(struct tcpcb_s *cb)
 {
     struct tcphdr_s *th = (struct tcphdr_s *)tcpbuf;
     struct addr_pair apair;
-    int header_len, len, minwindow;
+    int header_len, len;
 
     th->sport = htons(cb->localport);
     th->dport = htons(cb->remport);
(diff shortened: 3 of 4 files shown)


ghaerr:
@Mellvik,

This seems to work OK, but after more testing, I think it's not tuned. I plan on committing this initial version, since it solves the problem of never sending a zero window, which is definitely required to shutdown a sender, which is the problem in #1022. 

What I'm not sure of is whether this algorithm (see tcp_output.c to compare old and new) should be used all the time, or only on recovering from zero window. In essence, the calculation of the receive window is critical to keeping things flowing smoothly, this version always uses Clark's algorithm. RFC 1122 and other documents specify more detail which I haven't had time to read.

While I think this version works better than the previous, especially since it will stop a sender and should not halt, I'm not sure it uses all the buffer space it could. It is unclear how fast this is, which will depend on whether a full 3 packets can be outstanding very often. This will have to be tested.

Finally, I have noticed that in certain medium `ls -l` listings, the display in telnet will hang for a second then continue. Long listings work fine. This is not the PTY problem, I'm not sure what it is. I never see it in file transfers. So we're going to have to continue to test.

This sliding window change can be tested simultaneously with the close_wait problem. To turn on debugging for sliding window, use ^P. The close_wait printf's are always on in this version, and will be taken out after your next test run.

Thank you!

Mellvik:
Thank you @ghaerr - 

I will test this later today.
> This seems to work OK, but after more testing, I think it's not tuned. I plan on committing this initial version, since it solves the problem of never sending a zero window, which is definitely required to shutdown a sender, which is the problem in #1022.
> 
> What I'm not sure of is whether this algorithm (see tcp_output.c to compare old and new) should be used all the time, or only on recovering from zero window. In essence, the calculation of the receive window is critical to keeping things flowing smoothly, and always uses Clark's algorithm. RFC 1122 and other documents specify more detail which I haven't had time to read.
> 
> While I think this version works better than the previous, especially since it will stop a sender and should not halt, I'm not sure it uses all the buffer space it could. This will have to be tested.
> 
> Finally, I have noticed that in certain medium ls -l listings, the display in telnet will hang for a second then continue. Long listings work fine. This is not the PTY problem, I'm not sure what it is. I never see it in file transfers. So we're going to have to continue to test.
> 
> This sliding window change can be tested simultaneously with the close_wait problem. To turn on debugging for sliding window, use ^P. The close_wait printf's are always on in this version, and will be taken out after your next test run.
> 
Off the top of my head the thinking is that
- the situation we're trying to resolve is driven primarily by the speed difference between the sender and the recipient (elks) AND max size packets (file transfer). So by the time ktcp reponds to the first packet, the entire window has been filled.
- since other other uses - even a telnet ls -l - sends smaller packets, it may be wise to use the new algorithm when the situation occurs rather than all the time.

I guess further testing will provide a reasonable anseer to that question.

--Mellvik

Mellvik:
@ghaerr - I just tested this one.

The window closes completely as expected, but never opens up again. 
`ktcp` actually skips the second window reduction (the one advertising the 1460 window), and goes direct from 2920 to 0.

```
16:55:16.721668 IP (tos 0x0, ttl 64, id 135, offset 0, flags [none], proto TCP (6), length 92)
    10.0.2.15.ftp > 10.0.2.2.43028: Flags [P.], cksum 0xf0db (correct), seq 171:223, ack 59, win 4380, length 52: FTP, length: 52
	150 Opening BINARY data connection for 'uuencode'.
16:55:16.722061 IP (tos 0x8, ttl 64, id 31376, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.59316 > 10.0.2.15.49821: Flags [.], cksum 0x1ddf (incorrect -> 0xb0b5), seq 1:1461, ack 1, win 29200, length 1460
16:55:16.722193 IP (tos 0x8, ttl 64, id 31377, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.59316 > 10.0.2.15.49821: Flags [P.], cksum 0x1ddf (incorrect -> 0xd537), seq 1461:2921, ack 1, win 29200, length 1460
16:55:16.722273 IP (tos 0x8, ttl 64, id 31378, offset 0, flags [DF], proto TCP (6), length 1500)
    10.0.2.2.59316 > 10.0.2.15.49821: Flags [.], cksum 0x1ddf (incorrect -> 0x4847), seq 2921:4381, ack 1, win 29200, length 1460
16:55:16.735102 IP (tos 0x0, ttl 64, id 136, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.49821 > 10.0.2.2.59316: Flags [.], cksum 0x6723 (correct), ack 1461, win 2920, length 0
16:55:16.746861 IP (tos 0x0, ttl 64, id 137, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.49821 > 10.0.2.2.59316: Flags [.], cksum 0x6cd7 (correct), ack 2921, win 0, length 0
16:55:16.758533 IP (tos 0x0, ttl 64, id 138, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.49821 > 10.0.2.2.59316: Flags [.], cksum 0x6723 (correct), ack 4381, win 0, length 0
16:55:16.762518 IP (tos 0x10, ttl 64, id 31912, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.43028 > 10.0.2.15.ftp: Flags [.], cksum 0x182b (incorrect -> 0xde98), ack 223, win 29200, length 0
16:55:16.982532 IP (tos 0x8, ttl 64, id 31379, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.59316 > 10.0.2.15.49821: Flags [.], cksum 0x182b (incorrect -> 0xf513), ack 1, win 29200, length 0
16:55:16.984437 IP (tos 0x0, ttl 64, id 139, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.49821 > 10.0.2.2.59316: Flags [.], cksum 0x6723 (correct), ack 4381, win 0, length 0
16:55:17.432557 IP (tos 0x8, ttl 64, id 31380, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.59316 > 10.0.2.15.49821: Flags [.], cksum 0x182b (incorrect -> 0xf513), ack 1, win 29200, length 0
16:55:17.434482 IP (tos 0x0, ttl 64, id 140, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.49821 > 10.0.2.2.59316: Flags [.], cksum 0x6723 (correct), ack 4381, win 0, length 0
16:55:18.342556 IP (tos 0x8, ttl 64, id 31381, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.59316 > 10.0.2.15.49821: Flags [.], cksum 0x182b (incorrect -> 0xf513), ack 1, win 29200, length 0
16:55:18.344454 IP (tos 0x0, ttl 64, id 141, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.49821 > 10.0.2.2.59316: Flags [.], cksum 0x6723 (correct), ack 4381, win 0, length 0
16:55:20.182563 IP (tos 0x8, ttl 64, id 31382, offset 0, flags [DF], proto TCP (6), length 40)
    10.0.2.2.59316 > 10.0.2.15.49821: Flags [.], cksum 0x182b (incorrect -> 0xf513), ack 1, win 29200, length 0
16:55:20.184450 IP (tos 0x0, ttl 64, id 142, offset 0, flags [none], proto TCP (6), length 40)
    10.0.2.15.49821 > 10.0.2.2.59316: Flags [.], cksum 0x6723 (correct), ack 4381, win 0, length 0
```

ghaerr:
> The window closes completely as expected, but never opens up again.
ktcp actually skips the second window reduction (the one advertising the 1460 window), and goes direct from 2920 to 0.

Thank you! Can you tell me what exactly your test is, and also send the ^P trace? I need that to understand the window calculations and to be able to duplicate.

Mellvik:
Thank you @ghaerr,

The test case is ftpd in passive mode, the client sending a random file about 6k size - immediately after FTP login.

IN the output, the actual file transfer starts after the 'STOR uuencode' command.

—Mellvik

# ./ftpd -d

ktcp: debug on
tcp output: min sws 1500 space 4380 window 4380
New Client Detected...
Child running - cmd chan is 5.
tcp output: min sws 1500 space 4380 window 4380
tcp: recv data len 12 avail 4380
tcp output: min sws 1500 space 4368 window 4368
tcp output: min sws 1500 space 4380 window 4380
tcp output: min sws 1500 space 4380 window 4380
tcp: recv data len 12 avail 4380
tcp output: min sws 1500 space 4368 window 4368
tcp output: min sws 1500 space 4380 window 4380
tcp output: min sws 1500 space 4380 window 4380
tcp: recv data len 6 avail 4380
tcp output: min sws 1500 space 4374 window 4374
tcp output: min sws 1500 space 4380 window 4380
cmd: 'SYST' 29
tcp output: min sws 1500 space 4380 window 4380
tcp: recv data len 8 avail 4380
tcp output: min sws 1500 space 4372 window 4372
tcp output: min sws 1500 space 4380 window 4380
cmd: 'TYPE I' 10
tcp output: min sws 1500 space 4380 window 4380
tcp: recv data len 6 avail 4380
tcp output: min sws 1500 space 4374 window 4374
tcp output: min sws 1500 space 4380 window 4380
cmd: 'PASV' 9
tcp: port 49821 already bound, rejecting (use SO_REUSEADDR?)
PASV bind failure: host 0.0.0.0 port 49821
: Address already in use 
getsockname: adr 10.0.2.15, port 49822
tcp output: min sws 1500 space 4380 window 4380
tcp output: min sws 1500 space 4380 window 4380
Accepted connection from 10.0.2.2/38738 on fd 6
tcp: recv data len 15 avail 4380
tcp output: min sws 1500 space 4365 window 4365
tcpdev release: close socket 009bee, state is LISTEN
tcp output: min sws 1500 space 4380 window 4380
cmd: 'STOR uuencode' 14
tcp output: min sws 1500 space 4380 window 4380
tcp: recv data len 1460 avail 4380
tcp output: min sws 1500 space 2920 window 2920
tcp: recv data len 1460 avail 2920
tcp output: min sws 1500 space 1460 window 0
tcp: recv data len 1460 avail 1460
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0

ghaerr:
@Mellvik, thank you for the test and output data so quickly!

With this result, it is apparent the simple receive window advertise change has big ramifications. Thus, I don't want to change things in the middle of our other big problems we're trying to debug.

The problem is, this PR has a couple other small fixes that I want for the close_wait (and likely ftpd corruption) problem.

So - I am going to commit this code, but with the old (non-SWS) receive window behavior, since I'm tracking too many things right now. The new code will be there but commented out. I will then revisit this issue after the other problems. This change, which I thought small, just introduces too many variables to our otherwise well-working system (except for the outstanding bugs :)

Another push will be forthcoming, that will also include a couple other printfs for the close_wait issue for your continued testing tomorrow after making the suggested ftpd changes.

Sound good?

Mellvik:
Good, thank you!
I support that. 

Maybe we'll try your original idea, just increase the buffer size a little so that it's not an even number of max size packets.

Now we're REALLY getting there!

—M

ghaerr:
> Maybe we'll try your original idea, just increase the buffer size a little so that it's not an even number of max size packets.

You can try that, but it won't fully solve the problem. As this [article](http://www.tcpipguide.com/free/t_TCPSillyWindowSyndromeandChangesTotheSlidingWindow.htm) notes, "using unrestricted window size adjustments *ensures* that under conditions of heavy load, the window size will become small...". This is because the sender just takes advantage of a smaller, but still non-zero receive window size, until its all used up.

The proper solution has already been figured out - we must implement SWS. This first incarnation does not fully work, but will be committed turned OFF, so that we can move forward with other fixes included in it, and test after more research on SWS.

ghaerr:
I can see an SWS calculation error now in the following section of your trace:
```
tcp output: min sws 1500 space 4380 window 4380
tcp: recv data len 1460 avail 4380
tcp output: min sws 1500 space 2920 window 2920
tcp: recv data len 1460 avail 2920
tcp output: min sws 1500 space 1460 window 0 <-- here space available is less than MTU so zero window advertised
tcp: recv data len 1460 avail 1460
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
```
What is happening is that the minimum SWS is set to be 1500, while the actual max packet size is 1460 (because of the IP header being subtracted from the max TCP data length). Thus, our SWS implementation advertises 0, thinking there's not enough room in to accept a full packet (this is Clark's algorithm). I'll fix this. My system never sends full 1460 packets so I never saw it.

However, I can think of another case where this results in a problem: when the sender's effective MSS becomes lower than what ktcp thinks the MTU is, for whatever reason. Should this happen, the above 0 window case will continue to happen.

I am thinking that the SWS calculation will have to take into account the outstanding receive window size (i.e., the amount of data that received but still unACKed by ktcp). If there is unACKed data, this calculation should stay, but if not, then the max window size should be made available.

I remain concerned that the calculation still doesn't allow a full three packets to be received. You might look at my calculation in tcp_output.c as well as test, to see if this is the case.

I'll post another PR with this change included, and SWS turned on (uncommitted) for your testing.

Thank you!

ghaerr:
Studying SWS is still giving me fits...
```
tcp: recv data len 1460 avail 2920
tcp output: min sws 1500 space 1460 window 0 <-- Same comment, we should be using 1460 as min SWS
tcp: recv data len 1460 avail 1460 <-- I missed this, server sent another (final) 1460 packet
tcp output: min sws 1500 space 0 window 0 <-- now ktcp buffer space is actually zero, correct advertisement
tcp output: min sws 1500 space 0 window 0 <-- strange continued packet sends by ktcp, with no input packet processing??
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
tcp output: min sws 1500 space 0 window 0
```
Perhaps we will need an additional trace with DEBUG_TCP and ^P on, so that we can see why ktcp continues to send data (ACK only, or data also in packet??) and doesn't end up doing any receive processing. I'll add that info to the printf as well.

When our window size goes to zero, we still process inbound packets, and ACK every one of them. That ACK will contain a new window size, which should become non-zero once the receive buffer has enough unused data again. So I don't see yet why we're stuck in a loop.

Mellvik:
@ghaerr,
This is an intriguing problem.
And as to your last observation about why ktcp sends another packet, I can start by going back to the trace.
And yes, we possibly need another debug trace. I'll look at that in the morning.

A couple of observations about your discussion in the previous message. It seems to me that the reasoning in the article is somewhat aged. The example assumes the server buffer is equal to or smaller than the mss. So, when he's talking about min SWS, it's either equal to mss, or half mss. If this is the case (and it makes sense), we could test the system with a 720 min SWS.

Also, we need to keep in mind that in our case, the speed difference between ktcp and most other systems is such that the window will be filled immediately always, there will be no effective window reduction downwards, always max-then-min-then-rising-then back to min. I guess this is what you're talking about too, with the idea of taking the outstandng, unacked into account.

Then finally we have the case of  back-to-back elks, which is going to be very interesting.

More about this tomorrow.

-M

PR #1027, opened 2021-11-23
[ftpd] intermediate commit - for review only

Mellvik:
For code review only !!

- added passive mode (unstable)
- more error checking
- debug option prevents disconnect from controlling terminal
-

### DIFF
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -1,4 +1,3 @@
-
 /*
  * Minimal FTP server for ELKS
  * November 2021 by Helge Skrivervik - helge@skrivervik.com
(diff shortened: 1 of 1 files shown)


Mellvik:
@ghaerr, agreed - the symptoms point to a null reference or something similar. And it's more likely in my code that in `ktcp` - maybe another set of eyes can locate it ...

ghaerr:
Couple quick things to try:
- Set CMDBUFSIZ back to 1500
- In lines 669-679 and elsewhere, a valid file descriptor must include 0. That is, change tests to if (datafd >= 0) wherever tested, and set datafd = -1 when invalid.
- Possibly use -maout-stack=8192 to increase stack in Makefile, since your stack usage is so high (default is 4K, you'll use 3K for sure with increased CMDBUFSIZ)

Mellvik:
Well I'll be damned!
That did it - I'm getting back-to-back file transfers in passive mode (until the SWS hits).

I'll check (tomorrow) whether the problem was the CMDBUF size or the heap.

Thank you @ghaerr!!

—Mellvik

ghaerr:
> Well I'll be damned!
> check (tomorrow) whether the problem was the CMDBUF size or the heap.

That would be interesting to know. I'm pretty sure its CMDBUF; I'm the one who changed it, and knew it was a bit dangerous, since the command buffer was used to directly read from a socket. This is particularly dangerous when you might have to do multiple reads (or not) in order to get back the desired data from the command channel.

IMO, the design where buffers are allocated by one function, and then passed around a lot to other functions, which no longer know their size, is not great design. This is only a problem when the second-called function actually writes into the passed buffer, which I recommend against, for this reason.

On the other hand, you could have just run out of stack space - which is part of the reason why I don't like 1500 byte buffers allocated on the stack for ELKS applications. We (the programmers) have to be aware that ELKS is quite small and should try to build designs that don't require a known buffer size, and definitely don't want an increased buffer to blow the stack inadvertently just because the size changed.

That said - a better design would likely be using the iobuf ONLY for inner-loop buffering of lots of data; the iobuf should never be passed to another function, like it is now on occasion. Most other ELKS applications uses a dedicated static buffer for this purpose, not stack allocated. The size of iobuf should be able to be tuned to the OS, like we've done moving from 512 to 1500 bytes for MTU. Limit it's use to tight read/write loops only. Similarly, the cmdbuf should NOT depend on being a certain size. I recommend that the entire reading and parsing of any command channel return only be done in a single function, and that function should allow for any size cmdbuf.

> That did it - I'm getting back-to-back file transfers in passive mode 

Well, good to hear that the entire close_wait problem is likely non-existent. We'll still leave in the tons of close_wait debug code until you confirm this was the entire problem. The DEBUG_CLOSE can then be turned off (until the next bug!)

> (until the SWS hits).

As mentioned elsewhere, I'm going to turn off SWS completely for now, so that you can test ftpd with our known-working TCP stack, with the exception of the single issue causing 1-byte packets in #1022. That will be fixed later.

So tomorrow, the committed code will include all fixes but SWS will be inoperational for your finalized testing of #1024.

Thank you!

ghaerr:
@Mellvik, thanks for posting this. After your testing tomorrow, please comment on the close_wait PR #1024 whether ftpd is working or not. You can post what you found out about cmdbuf vs stack here, but if there are new or continuing issues not discussed elsewhere, please open another issue. 

When you're done optimizing ftpd as discussed, you're welcome to open another temp PR for a code review. I'm closing this now so as to not confuse issues, since we've been in a flurry of them! :)

Thank you!

PR #1028, opened 2021-11-24
[net] Non-SWS enhancement to receive window advertisement

ghaerr:
Possible solution to initial receive window hang discussed in #1022. Does NOT use silly window solution.

Includes silly window solution, but turned off for now. Implementing a full-scale receive window calculation is turning out to be very touchy. After reading the Linux 2.0 sources, they strongly recommend against Clark's algorithm, saying BSD systems  get quite confused by it. They implement a slightly more complex solution that will require quite a bit more testing. At this time, what is important to ELKS is that we get reliable networking, not super-efficient networking. We have not achieved the former yet.

@Mellvik, I have come to the conclusion that now is definitely not the time to be changing this algorithm, as even though it seems simple, it becomes very complicated when taking into account the remote end's interpretation of our changing window size and extra ACK sent after ever application read. The current code, including this PR, has worked quite well, except for the single hanging problem discovered in #1022. In addition, a prior PR fixed an additional problem by sending an "extra knock" ACK which ktcp currently still requires. When adding SWS support, this "extra" ACK is causing more complications than I have time to debug and analyze at this time, especially with all the outstanding other problems. While the contained SWS solution works, it is not causing the remote end to send larger, more efficient packets, and also causes a strange jerkiness again in telnet on occasion.

This PR keeps the original, very basic algorithm with one change: originally when the window was previously calculated as negative or 0, it was set to 1. This PR sets it to 0. That is, exactly the same except that we won't advertise a byte of space if we have none. I am interested to see whether this changes the behavior and stops the hang in #1022, or otherwise affects all our previously working programs.

That said, there is lots more to be learned on packet traces, especially on fine tuning for speed, after we get all the other networking bugs fixed, ftpd fully working, and have tested ELKS TCP to our liking. Only after that should we embark on this road, and we may not ever need to anyways.

Thus, this PR will be committed and can be tested with file transfers and telnet with no changes. The previous SWS code is left in but should not be turned on or tested at this time. If this PR fails to work better, then we will be reverting to never advertising a zero window, and advertising 1 byte as it used to, and #1022 will stay open until our other problems are solved. 

There is still quite a bit of extra close_wait and other code that needs to be turned off, as soon as the ftpd server starts working reliably in passive (and active) mode. I look forward to hearing how/if this last change works.

Thank you!

### DIFF
--- a/elkscmd/ktcp/tcp.c
+++ b/elkscmd/ktcp/tcp.c
@@ -288,6 +288,7 @@ static void tcp_established(struct iptcp_s *iptcp, struct tcpcb_s *cb)
     if (datasize == 0 && ((h->flags & TF_ALL) == TF_ACK))
 	return; /* ACK with no data received - so don't answer*/
 
+    debug_tune("ACK %d\n", datasize);
     cb->rcv_nxt += datasize;
     tcp_send_ack(cb);
 }
--- a/elkscmd/ktcp/tcp.h
+++ b/elkscmd/ktcp/tcp.h
@@ -26,8 +26,13 @@
  * control block input buffer size - max window size, doesn't have to be power of two
  * default will be (ETH_MTU - IP_HDRSIZ) * 3 = (1500-40) * 3 = 4380
  */
+#define USE_SWS		0		/* =1 to use silly window algorithm */
+
+#if USE_SWS
+#define CB_NORMAL_BUFSIZ	4380	/* normal input buffer size*/
+#else
 #define CB_NORMAL_BUFSIZ	4892	/* bufsize for older receive window algorithm*/
-//#define CB_NORMAL_BUFSIZ	4380	/* normal input buffer size*/
+#endif
 
 /* max outstanding send window size*/
 #define TCP_SEND_WINDOW_MAX	1024	/* should be less than TCP_RETRANS_MAXMEM*/
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -362,6 +362,35 @@ void tcp_retrans_retransmit(void)
     }
 }
 
+static int tcp_calc_rcv_window(struct tcpcb_s *cb)
+{
+    int len;
+
+#if USE_SWS
+    /* must have half the buffer or MTU bytes available for nonzero window */
+    int minwindow = cb->buf_size >> 1;
+    if (minwindow > MTU-40)
+	minwindow = MTU-40;
+    len = CB_BUF_SPACE(cb);
+    if (len < minwindow)
+	len = 0;			/* advertise zero window, stops sender */
+    debug_tune("tcp output: datalen %d min sws %u space %u window %u\n",
+	cb->datalen, minwindow, CB_BUF_SPACE(cb), len);
+#else
+    len = CB_BUF_SPACE(cb);
+    if (cb->buf_size > PUSH_THRESHOLD)	/* FIXME temp hack for small listen buffers */
+	len -= PUSH_THRESHOLD;
+    if (len < 0)
+	len = 0;			/* zero window will stop sender */
+    //if (len <= 0)
+	//len = 1;			/* Never advertise zero window size */
+    debug_tune("tcp output: datalen %d space %u window %u\n",
+	cb->datalen, CB_BUF_SPACE(cb), len);
+#endif
+
+    return len;
+}
+
 void tcp_output(struct tcpcb_s *cb)
 {
     struct tcphdr_s *th = (struct tcphdr_s *)tcpbuf;
@@ -375,25 +404,7 @@ void tcp_output(struct tcpcb_s *cb)
 
     cb->send_nxt += cb->datalen;
 
-#if 0
-    /* must have half the buffer or MTU bytes available for nonzero window */
-    int minwindow = cb->buf_size >> 1;
-    if (minwindow > MTU)
-	minwindow = MTU;
-    len = CB_BUF_SPACE(cb);
-    if (len < minwindow)
-	len = 0;			/* advertise zero window, stops sender */
-    debug_tune("tcp output: min sws %u space %u window %u\n",
-	minwindow, CB_BUF_SPACE(cb), len);
-#else
-    len = CB_BUF_SPACE(cb);
-    if (cb->buf_size > PUSH_THRESHOLD)	/* FIXME temp hack for small listen buffers */
-	len -= PUSH_THRESHOLD;
-    if (len <= 0)
-	len = 1;			/* Never advertise zero window size */
-    debug_tune("tcp output: space %u window %u\n", CB_BUF_SPACE(cb), len);
-#endif
-
+    len = tcp_calc_rcv_window(cb);
     th->window = htons(len);
     th->urgpnt = 0;
     th->flags = cb->flags;
--- a/elkscmd/ktcp/tcpdev.c
+++ b/elkscmd/ktcp/tcpdev.c
@@ -258,6 +258,10 @@ static void tcpdev_read(void)
 	return;
     }
 
+    if (data_avail > cb->buf_used)	// FIXME testing only
+	printf("tcpdev_read: data_avail > used, avail %d used %d\n",
+	    data_avail, cb->buf_used);
+
     data_avail = db->size < data_avail ? db->size : data_avail;
     cb->bytes_to_push -= data_avail;
     if (cb->bytes_to_push <= 0)
@@ -287,8 +291,10 @@ static void tcpdev_read(void)
 
     /* send ACK to restart server should window have been full (unless it's netstat)*/
     if (cb->remport != NETCONF_PORT || cb->remaddr != 0)
-	if (cb->remport != local_ip)	/* no ack to localhost either*/
-		tcp_send_ack(cb);
+	if (cb->remport != local_ip) {	/* no ack to localhost either*/
+	    debug_tune("addtl ACK, app read %d bytes\n", data_avail);
+	    tcp_send_ack(cb);
+	}
 }
 
 /* inform kernel of socket data bytes available*/


Mellvik:
Thank you @ghaerr - not the least for even more insight into a formerly grey, or rather quite black, area of knowledge on my part.

That's part of what makes it intriguing, in addition to being able to test the various alternatives in real time on physical hw, with (kinda) extreme conditions in terms of speed difference.

I strongly support simplicity, and the approach you suggest seems to be exactly that. So let's get at it:

The first tests are very encouraging indeed. And - incidentally - with this solution the 'extra knock' seems a blessing.  I've transferred more than a thousand files into ELKS via ftpd (active mode still, passive mode needs some more work), here are som observations:

— Overall - no problems at all, no delays, no silly windows, in fact hardly any packets below 1460 payload. (!!)
- KTCP reduces the window as expected, then climbs back. 
- If the climb back is fast, the client doesn't pick the first open window, but waits for something decent (!). 'Fast' in this context means less than 100ms, maybe less than 50ms.
- This is where the extra knock comes in handy for performance: ktcp acks the latest received packet with a window @ say, 560 or even 1322, then 30ms later the knock announces window size 1580. The client picks the latter and kicks in with a full size packet. Performance!
- OTOH - if the smaller window stays for more than ca. 0.3s, the client uses the available window. The smallest packet I've seen (except the final data packet of a file transfer) is 512b.

This is in fact excellent. And I'm growing more and more curious about what elks-to-elks will look like…

While my eyesight is recovering from zillions of tracelines, more file transfers are rolling.

More later.

—M

ghaerr:
@Mellvik,

Thanks for the testing, and I'm glad the small change (advertising 0 sometimes instead of 1) doesn't cause more problems.

Did you see any zero window advertisement, with subsequent recovery in this last testing? If so, this means likely that this fixes #1022. If not, we should continue testing to actually see a zero window stop the sender, rather than the previous result of 1-byte packets and then hang.

> And I'm growing more and more curious about what elks-to-elks will look like…

I'm testing elks-to/from-remote using your enhanced ftpget/ftpput:
```
remote=192.168.179.103

ftpput -v $remote /linux test test
ftpput -v $remote yy test test

ftpget -v $remote / test test
ftpget -v $remote /yy test test

urlget ftp://test:test@$remote/yy
``` 
This runs well every time with no errors, including running it again quickly after exit (it's a shell script). After your next changes to ftpd, this should work elks-to-elks, correct? (i.e. ftpget uses passive mode, right?)

> with this solution the 'extra knock' seems a blessing.

In the Comer TCP/IP Vol 2 book (p267), this is called a "gratuitous acknowledgement" whose purpose is to restart a zero window quickly and advertise the new window size simultaneously.

ghaerr:
@Mellvik,

I have been researching more about SWS, and wanted to give you an update. I think this explains why my last two attempts at implementing SWS (though commented out and not used) don't work.

In RFC 1122, the original TCP specification, the receive window is NOT the number of bytes of buffer space available, but is actually defined as the range of sequence numbers that are acceptable for the receiver, with any outside that range to be dropped and not processed. Since the receive window (16 bits) isn't actually a (32-bit) sequence number, the receive window value defines the range (window)  from the last ACK'd sequence number up to ACK+rcv_window. This is a subtle but important difference.

The spec goes on to say TCP should never "shrink the window". That is, it should never advertise an available range, then advertise a smaller range. This is because the sender may have already sent multiple packets upon seeing the first range. The tricky part is they're not talking about the "receive window" advertisement shrinking; the advertised receive window naturally gets smaller as buffer space is used (with their associated packets unACKed). The "receive window" is conceptually the size of the available buffer PLUS the amount of data already received. This window size is supposed to remain constant, and become the value of the round trip delay times the number of outstanding bytes, in order for the sender to send without becoming jerky.

So, long story short, what I now believe is happening, is that ELKS's simplistic approach of advertising just the remaining buffer size actually works, because the remaining buffer size is always equal to the total buffer size less the amount received and unprocessed! This simplistic approach only works well when the remote TCP is smart - and the spec has "robustness" mechanisms to deal with less sophisticated TCPs.

The initial SWS algorithm only tracked the buffer size, it did NOT take into account the last ACK'd sequence number. Thus, when the simple SWS algorithm "shut down" the sender when the remaining buffer size was less than an MSS packet and advertising zero, what was actually happening is "shrinking the window". That is, the previously advertised available space was shrunk rapidly, even though that much data had not been received. This causes the remote TCP to act badly, and subsequent packets get dropped. This is the reason it caused jerkiness, which I saw.

I finally figured this all out through more testing, and then found yet another section of code that actually drops packets that arrive out of sequence. This was never displayed so we never knew about it. Ktcp never reassembles packets, so ANY packet arriving out-of-order is summarily dropped. Additionally, there was a bug in the htons calculation. This didn't affect much, but still buggy.

I will be preparing one more PR which will not change the existing algorithm, and will remove the "SWS" implementation, since it doesn't take into account the previously received (but unACKed) packets since the last window advertisement. It will also display the additional case of when ktcp drops packets, which should be illuminating, as bad receive window advertisement results in unseen packet drops, even on networks that aren't losing data. The sender is sending outside the range of advertised window.

Our simpler approach of just advertising the available buffer space seems to work (for now), as explained above. This may result in occasional smaller packets with heavy activity, but should recover. I've also found through testing that we don't need to subtract a "PUSH_THRESHOLD" and can run with a 4380 byte (3\*MTU) buffer directly, also saving another 512 bytes per connection.

I hope this makes sense, it's the latest take on the situation. Hopefully this fixes #1022 hanging, although we may still small packets on occasion.

Mellvik:
> Thanks for the testing, and I'm glad the small change (advertising 0 sometimes instead of 1) doesn't cause more problems.
> 
> Did you see any zero window advertisement, with subsequent recovery in this last testing? If so, this means likely that this fixes #1022. If not, we should continue testing to actually see a zero window stop the sender, rather than the previous result of 1-byte packets and then hang.
> 
I guess I didn't mention that explicitly. But yes, the window goes to zero all the time.
> And I'm growing more and more curious about what elks-to-elks will look like…
> 
> I'm testing elks-to/from-remote using your enhanced ftpget/ftpput:
> 
> remote=192.168.179.103
> 
> ftpput -v $remote /linux test test
> ftpput -v $remote yy test test
> 
> ftpget -v $remote / test test
> ftpget -v $remote /yy test test
> 
> urlget ftp://test:test@$remote/yy
> This runs well every time with no errors, including running it again quickly after exit (it's a shell script). After your next changes to ftpd, this should work elks-to-elks, correct? (i.e. ftpget uses passive mode, right?)
> 
Yes, it uses passive mode, and right, that should work well. Hopefully, i'll get time do the final fixing tomorrow.
> with this solution the 'extra knock' seems a blessing.
> 
> In the Comer TCP/IP Vol 2 book (p267), this is called a "gratuitous acknowledgement" whose purpose is to restart a zero window quickly and advertise the new window size simultaneously.
> 
I now understand why. Btw I'll post some tcptraces tomorrow. As  crazy as it sounds, it's kinda enjoyable to follow the window changes.

-M

Mellvik:
Thank you for another thorough rundown, @ghaerr.

While this all makes more sense now, I really need to re-digest this in the morning!

We're getting there!

-M

Mellvik:
Again thank you for doing the research, @ghaerr.

And again - the refresh works wonders for bringing old and forgotten stuff back to the top og the 'stack'.

> 
> 
> @Mellvik <https://github.com/Mellvik>,
> 
> I have been researching more about SWS, and wanted to give you an update. I think this explains why my last two attempts at implementing SWS (though commented out and not used) don't work.
> 
> In RFC 1122, the original TCP specification, the receive window is NOT the number of bytes of buffer space available, but is actually defined as the range of sequence numbers that are acceptable for the receiver, with any outside that range to be dropped and not processed. Since the receive window (16 bits) isn't actually a (32-bit) sequence number, the receive window value defines the range (window) from the last ACK'd sequence number up to ACK+rcv_window. This is a subtle but important difference.
> 
> The spec goes on to say TCP should never "shrink the window". That is, it should never advertise an available range, then advertise a smaller range. This is because the sender may have already sent multiple packets upon seeing the first range. The tricky part is they're not talking about the "receive window" advertisement shrinking; the advertised receive window naturally gets smaller as buffer space is used (with their associated packets unACKed). The "receive window" is conceptually the size of the available buffer PLUS the amount of data already received. This window size is supposed to remain constant, and become the value of the round trip delay times the number of outstanding bytes, in order for the sender to send without becoming jerky.
> 
Yes, you're right - this is exactly what it was like before we started fiddling with dynamic windows. The simple picture (in my mind, and admittedly I really needed a refresh on the theory about this, which you again initiated) is that the receiver advertises the available buffer size (not the space) at connect time, and never change it. When thinking about it, it's pure simplicity. The sliding window has a constant size, the start- and endpoints sliding along driven by the ACKs from the receiver. The sender knows exactly how much space is available at all times - by (avail = windowsize - (last_sender_seq# - last-receiver_ack). So a working connection doesn't ever need the window to be changed – unless there is a resource related reason for it. Such as congestion, long round trip times and high bandwidth (there famous bandwidth*delay product).
So - assuming that TCP implementations adhere to the standard (they mostly do), and also assuming very predictable environments, ripping out all the dynamism related to window sizing would be a good thing. However, when communicating outside the local LAN, and possibly even within, given the extreme differences in speed between nodes, this will not work well.
> So, long story short, what I now believe is happening, is that ELKS's simplistic approach of advertising just the remaining buffer size actually works, because the remaining buffer size is always equal to the total buffer size less the amount received and unprocessed! This simplistic approach only works well when the remote TCP is smart - and the spec has "robustness" mechanisms to deal with less sophisticated TCPs.
> 
Agreed. Still, it would be an interesting experiment to see what happens if we just kept the window size constant. The other (linux) end should handle that just fine, and the result should be about the same. That said, I'm seeing quite a bit of dynamism in the window advertisements from Linux too, usually in the 16-30k area, some times up to 60+k bytes.
> The initial SWS algorithm only tracked the buffer size, it did NOT take into account the last ACK'd sequence number. Thus, when the simple SWS algorithm "shut down" the sender when the remaining buffer size was less than an MSS packet and advertising zero, what was actually happening is "shrinking the window". That is, the previously advertised available space was shrunk rapidly, even though that much data had not been received. This causes the remote TCP to act badly, and subsequent packets get dropped. This is the reason it caused jerkiness, which I saw.
> 
Yes, this makes sense now.
> I finally figured this all out through more testing, and then found yet another section of code that actually drops packets that arrive out of sequence. This was never displayed so we never knew about it. Ktcp never reassembles packets, so ANY packet arriving out-of-order is summarily dropped. Additionally, there was a bug in the htons calculation. This didn't affect much, but still buggy.
> 
Interesting. Then I guess it's not very surprising now that it's on there table. The implementation is (relatively) simple, there has to be shortcuts. If the choices are wise, we'll never see then, as you note.
> I will be preparing one more PR which will not change the existing algorithm, and will remove the "SWS" implementation, since it doesn't take into account the previously received (but unACKed) packets since the last window advertisement. It will also display the additional case of when ktcp drops packets, which should be illuminating, as bad receive window advertisement results in unseen packet drops, even on networks that aren't losing data. The sender is sending outside the range of advertised window.
> 
Very interesting. Although I don't think we've had much packet drop at the tcp level.
> Our simpler approach of just advertising the available buffer space seems to work (for now), as explained above. This may result in occasional smaller packets with heavy activity, but should recover. I've also found through testing that we don't need to subtract a "PUSH_THRESHOLD" and can run with a 4380 byte (3*MTU) buffer directly, also saving another 512 bytes per connection.
> 
> I hope this makes sense, it's the latest take on the situation. Hopefully this fixes #1022 <https://github.com/jbruchon/elks/issues/1022> hanging, although we may still small packets on occasion.
> 
Looking forward to test it, and yes, it does make sense. More understanding equals better choices. Saving memory and reducing bugs seems to be what we've been at for some time now. With great achievements. Happy thanksgiving.

—M

ghaerr:
> Yes, you're right - this is exactly what it was like before we started fiddling with dynamic windows. The simple picture (in my mind, and admittedly I really needed a refresh on the theory about this, which you again initiated) is that the receiver advertises the available buffer size (not the space) at connect time, and never change it. When thinking about it, it's pure simplicity. The sliding window has a constant size, the start- and endpoints sliding along driven by the ACKs from the receiver. The sender knows exactly how much space is available at all times - by (avail = windowsize - (last_sender_seq# - last-receiver_ack). So a working connection doesn't ever need the window to be changed – unless there is a resource related reason for it. Such as congestion, long round trip times and high bandwidth (there famous bandwidth*delay product).
So - assuming that TCP implementations adhere to the standard (they mostly do), and also assuming very predictable environments, ripping out all the dynamism related to window sizing would be a good thing. However, when communicating outside the local LAN, and possibly even within, given the extreme differences in speed between nodes, this will not work well.

You got it, on all points! And you can also see why any "improved" version of our simple window advertisement calculation would have to keep track of the last ACK sequence numbers at the time of transmit, etc. When looking at truly robust TCP implementations, it gets quite a bit more complicated. But the good news is, those implementations are specifically built to handle simpler implementations like ours. I think ELKS, given its limit on size and speed, will always lean hard towards simplicity.

I plan on removing the temp/nonworking SWS option as you're hinting at, but am leaving it In for now, as it is proving useful for a way to quickly generate packet drops with fancier (but incorrect) advertised window sizes.

> Still, it would be an interesting experiment to see what happens if we just kept the window size constant. The other (linux) end should handle that just fine, and the result should be about the same. 

When you say "keeping the window size constant", do you mean the conceptual receive window, or the advertised window?

As we both noted above, the conceptual receive window should remain constant. If we make the advertised window constant, you'll get continuous packet drops when the sender continuously overruns your buffer space. Nonetheless, it is interesting to try this yourself and look at it. I did just that to finally understand how my initial attempt at SWS was (and still is) incorrect!

> That said, I'm seeing quite a bit of dynamism in the window advertisements from Linux too, usually in the 16-30k area, some times up to 60+k bytes.

Yes, this is the result of more complex internal calculations. The "extra ACK" we're sending definitely adds to the complexity as well, you'll see.

Mellvik:
@ghaerr,
> Still, it would be an interesting experiment to see what happens if we just kept the window size constant. The other (linux) end should handle that just fine, and the result should be about the same.
> 
> When you say "keeping the window size constant", do you mean the conceptual receive window, or the advertised window?
> 
I mean both - the allocated buffer space and the advertized window size. I get your reservation and you may be right. But then again, this is what the original spec says and even though later amendments have opened for more flexible uses of the window advertisement, I would be surprised if the original doesn't work. Still - just a curiosity...
> As we both noted above, the conceptual receive window should remain constant. If we make the advertised window constant, you'll get continuous packet drops when the sender continuously overruns your buffer space.
> 
Well, I'm not sure it will. Because the sender always knows what has been acked and what's outstanding, and should be doing the math w/o the receiver's help.

> Nonetheless, it is interesting to try this yourself and look at it. I did just that to finally understand how my initial attempt at SWS was (and still is) incorrect!
> 
> That said, I'm seeing quite a bit of dynamism in the window advertisements from Linux too, usually in the 16-30k area, some times up to 60+k bytes.
> 
> Yes, this is the result of more complex internal calculations. The "extra ACK" we're sending definitely adds to the complexity as well, you'll see.
> 
> 

Thank you!

-Mellvik

ghaerr:
> Well, I'm not sure it will. Because the sender always knows what has been acked and what's outstanding, and should be doing the math w/o the receiver's help.

Definitely try it, right now! You'll learn how quickly the system changes when the advertised receive window is not calculated correctly, or is confused with the conceptual receive window. And I want to hear the results. 

If the sender knew all you're talking about (which it can't because packets are lost, dropped by other TCPs, reordered, etc), then the receive window and all its associated code in Linux would not be necessary. This stuff is only working well for ELKS now because the remote TCP is much smarter than us.

Mellvik:
Yeah, you may be right.

I'll give it a round for the heck of it in the morning.


—M

PR #1029, opened 2021-11-25
[net] Cleanup non-SWS receive window implementation

ghaerr:
Cleaned up version of working receive window implementation, discussed in #1028.

Removes 512-byte PUSH_THRESHOLD subtraction from buffer space available but otherwise same algorithm as was last tested by @Mellvik.

Debug close code not displayed unless ^P.
Added additional dropped packet printf, also displays ACK and "extra" ACK sequence numbers in ^P mode for debugging.
SWS window management code still in, but commented out. Left in to show bad behavior in debug traces for now.

Should work well. If so, #1022 can be closed.

### DIFF
--- a/elkscmd/ktcp/config.h
+++ b/elkscmd/ktcp/config.h
@@ -55,8 +55,7 @@ void dprintf(const char *, ...);
 #endif
 
 #if DEBUG_CLOSE
-//#define debug_close	DPRINTF
-#define debug_close	printf
+#define debug_close	DPRINTF
 #else
 #define debug_close(...)
 #endif
--- a/elkscmd/ktcp/tcp.c
+++ b/elkscmd/ktcp/tcp.c
@@ -51,8 +51,9 @@ void tcp_print(struct iptcp_s *head, int recv, struct tcpcb_s *cb)
     debug_tcpdata("[%s] ", tcp_flags(head->tcph->flags));
     if (cb) {
       if (recv) {
-	if (head->tcplen > 20) debug_tcp("seq %lu-%lu ", ntohl(head->tcph->seqnum) - cb->irs,
-				ntohl(head->tcph->seqnum) - cb->irs+head->tcplen-20);
+	if (head->tcplen > 20)
+	    debug_tcpdata("seq %lu-%lu ", ntohl(head->tcph->seqnum) - cb->irs,
+		ntohl(head->tcph->seqnum) - cb->irs+head->tcplen-20);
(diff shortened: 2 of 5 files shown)


Mellvik:
@ghaerr - first impression is that this one works well.
The pattern in the packet traces are about the same as yesterday, even fewer packet below maxlen.

I'm seeing some of these:
```
tcp: application read too small after FIN, data_avail 508
tcp: dropping packet, bad seqno: need 49641 got 48181 size 1460
tcp: dropping packet, bad seqno: need 49641 got 48181 size 1460
tcp: dropping packet, bad seqno: need 49641 got 48181 size 1460
tcp: dropping packet, bad seqno: need 49641 got 48181 size 1460
tcp: dropping packet, bad seqno: need 49641 got 48181 size 1460
tcp: application read too small after FIN, data_avail 1464
```
(not the `application read too small` - they are coming all the time, but the `dropping packet`. It has always been legitimate so far. `ktcp` is ditching retransmits from the client, who gets impatient when ELKS is writing to the floppy.

-M

ghaerr:
> first impression is that this one works well.

Glad to hear that! It seems we've fixed the problem of window size advertisements. This is essentially our original algorithm, but advertising 0 at times instead of 1, which stops the sender. We've also stopped subtracting a "push threshold" of 512 byte every time, which allows our buffers to become 512 bytes smaller with the same results.

> It has always been legitimate so far. ktcp is ditching retransmits from the client, who gets impatient when ELKS is writing to the floppy.

Yes - this is the code I mentioned previously that's been dropping packets all the time, without our ever having seen it. I'll leave it on a while longer, as it's interesting to see when it happens. I started seeing them with the buggy SWS code, for instance. When we get packet drops with no packet loss (and no hardware timing issues like slow floppies), we know there's still a protocol error. Of course, it's normal for a retransmit if the floppy takes longer than 1-2 secs.

> the application read too small - they are coming all the time

I'll turn this off in the next commit. It was put in to show the problem we fixed earlier, which is that the application hasn't had time to read all the buffered TCP data before a FIN was sent by the remote. It was originally put in to suggest that an application buffer is too small to read the data received, but now it's showing that there are more than one unread packets when the FIN arrives. (Everything is working fine).

Please close #1022 when you feel this has solved the receive window timing issues.

Mellvik:
Agreed @ghaerr. 

And since you opened the issue on my behalf, you'll get the pleasure of closing it :-)

Thank you!


—M

PR #1030, opened 2021-11-25
Ftpd: Another round of code review

Mellvik:
@ghaerr,
The fixes after your comments the other day seemed to work.  Then they didn't. I'm still getting the same symptoms as before when using PASV mode, and need another pair of eyes to glance over it. 

You've seen the consequences before - like socket [000000] with random consequences. `Ktcp` is the latest merge including #1029 . Here is one of them, although a rare specimen:
```
tcp: extra ACK seq 76, app read 9 bytes
cmd: 'STOR ls' 14
tcp output: len 0 space 4380 window 4380
tcp output: len 46 space 4380 window 4380
tcp[000000] packet in established, fin: 1, data: 0, setting state to CLOSE_WAIT     <--------------
tcp: ACK seq 2 len 0
tcp output: len 0 space 4380 window 4380
tcp: dropping packet, bad seqno: need 4381 got 4380 size 0
tcp[00a95e] packet in established, fin: 1, data: 0, setting state to CLOSE_WAIT
tcp: ACK seq 77 len 0
tcp output: len 0 space 4380 window 4380
tcp output: len 0 space 4380 window 4380
```
Here's another one - this one ir more common and results in a (soft) hung `ftpd`:
```
tcp output: len 0 space 4371 window 4371
tcpdev release: close socket 00a4fe, state is LISTEN
tcp: extra ACK seq 64, app read 9 bytes
cmd: 'STOR ls' 14
tcp output: len 0 space 4380 window 4380
tcp output: len 46 space 4380 window 4380
tcp[000000] packet in established, fin: 1, data: 0, setting state to CLOSE_WAIT
tcp: ACK seq 2 len 0
tcp output: len 0 space 4380 window 4380
```

Thank you.

-M

### DIFF
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -1,4 +1,3 @@
-
 /*
  * Minimal FTP server for ELKS
  * November 2021 by Helge Skrivervik - helge@skrivervik.com
@@ -29,31 +28,54 @@
 #define		TRUE		1
 #define		FALSE		0
 #define		FTP_PORT	21
+#define		PASV_PORT	49821;	/* 'random' port for passive connections */
+
+#define		UC(b)		(((int) b) & 0xff)
 
 /* Commands */
-#define		CMD_LIST	1
-#define		CMD_RETR	2
-#define		CMD_STOR	3
-#define		CMD_SYST	4
-#define		CMD_PASV	5
-#define		CMD_PORT	6
-#define		CMD_PASS	7
-#define		CMD_USER	8
-#define		CMD_CLOS	9
-#define		CMD_ABOR	10
-#define		CMD_QUIT	11
-#define		CMD_SKIP	12
-#define		CMD_TYPE	13
-#define		CMD_PWD		14
-#define		CMD_DELE	15
-#define		CMD_CWD		16
-#define		CMD_MKD		17
-#define		CMD_NLST	18
-#define		CMD_UNKNOWN	30
+enum {
+	CMD_USER,
+	CMD_PASS,
+	CMD_ACCT,
+	CMD_CWD,
+	CMD_CDUP,
+	CMD_SMNT,
+	CMD_QUIT,
+	CMD_REIN,
+	CMD_PORT,
+	CMD_PASV,
+	CMD_TYPE,
+	CMD_STRU,
+	CMD_MODE,
+	CMD_RETR,
+	CMD_STOR,
+	CMD_STOU,
+	CMD_APPE,
+	CMD_ALLO,
+	CMD_REST,
+	CMD_RNFR,
+	CMD_RNTO,
+	CMD_ABOR,
+	CMD_DELE,
+	CMD_RMD,
+	CMD_MKD,
+	CMD_PWD,
+	CMD_LIST,
+	CMD_NLST,
+	CMD_SITE,
+	CMD_SYST,
+	CMD_STAT,
+	CMD_HELP,
+	CMD_NOOP,
+	CMD_UNKNOWN,
+	CMD_EMPTY,
+	CMD_CLOSE
+};
 
+//static long myaddr;
 static int debug = 0;
 
-//function trims leading and trailing whitespaces
+/* Trim leading and trailing whitespaces */
 void trim(char *str) {
 	int i;
 	int begin = 0;
@@ -130,12 +152,14 @@ int setup_data_connection(char *client_ip, unsigned int client_port, int server_
     		perror("socket error");
     		return -1;
 	}
+#if 0
 	if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &on, sizeof (on)) < 0) {
 		/* This should not happen */
 		printf("Data channel socket error: REUSEADDR, closing channel.\n");
 		close(fd);
 		return -1;
 	}
+#endif
 
 	/* bind port for data connection to be server port - 1 using a temporary struct sockaddr_in */
 	bzero(&tempaddr, sizeof(tempaddr));
@@ -162,7 +186,7 @@ int setup_data_connection(char *client_ip, unsigned int client_port, int server_
 
 	cliaddr.sin_family = AF_INET;
 	cliaddr.sin_port = htons(client_port);
-	cliaddr.sin_addr.s_addr = in_gethostbyname(client_ip);
+	cliaddr.sin_addr.s_addr = in_aton(client_ip);
 
 	if (connect(fd, (struct sockaddr *) &cliaddr, sizeof(cliaddr)) < 0) {
     		perror("connect error");
@@ -196,7 +220,6 @@ int get_command(char *command){
 	if (strncmp(command, "LIST", 4) == 0)	  {value = CMD_LIST;}
 	else if (strncmp(command, "RETR", 4) == 0) {value = CMD_RETR;}
 	else if (strncmp(command, "STOR", 4) == 0) {value = CMD_STOR;}
-	else if (strncmp(command, "SKIP", 4) == 0) {value = CMD_SKIP;}
 	else if (strncmp(command, "ABOR", 4) == 0) {value = CMD_ABOR;}
 	else if (strncmp(command, "SYST", 4) == 0) {value = CMD_SYST;}
 	else if (strncmp(command, "QUIT", 4) == 0) {value = CMD_QUIT;}
@@ -208,7 +231,7 @@ int get_command(char *command){
 	else if (strncmp(command, "PWD", 3) == 0) {value = CMD_PWD;}
 	else if (strncmp(command, "CWD", 3) == 0) {value = CMD_CWD;}
 	else if (strncmp(command, "MKD", 3) == 0) {value = CMD_MKD;}
-	/* FIXME: Add unknown command as default */
+	else value = CMD_UNKNOWN;
 	return value;
 }
 
@@ -220,19 +243,18 @@ int get_command(char *command){
 int do_login(int controlfd, char *buf) {
 	char *str = "331 User OK.\r\n";
 
-	//bzero(buf, CMDBUFSIZ);
 	if(read(controlfd, buf, CMDBUFSIZ) < 0)
 		return -1;
 	if (strncmp(buf, "USER", 4) < 0) 
 		return -1;
 	write(controlfd, str, strlen(str));
 
-	//bzero(buf, CMDBUFSIZ);
 	if(read(controlfd, buf, CMDBUFSIZ) < 0)
 		return -1;
 	if (strncmp(buf, "PASS", 4) < 0) 
 		return -1;
 	str = "230 Password OK.\n";
+	chdir("/root");		/* brute force ... */
 	write(controlfd, str, strlen(str));
 	return 0; /* always OK */
 }
@@ -261,7 +283,7 @@ int do_list(int controlfd, int datafd, char *input){
 		strcat(cmd_buf, "/bin/ls -l");
 	}
 
-	if (debug) printf("LIST/NDIR: '%s'\n", cmd_buf);
+	if (debug) printf("LIST/NDIR: '%s' (%d)\n", cmd_buf, datafd);
     	dir = opendir(iobuf); 	/* test for existence */
     	if (!dir) {
     		str= "550 No Such File or Directory.\r\n";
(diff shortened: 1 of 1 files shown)


ghaerr:
> The fixes after your comments the other day seemed to work. Then they didn't. 

You still have a "good" copy you're testing with though, right? Or did you lose that and we now have to find what happened?

I'll take a look at this version and comment. Unfortunately, this PR doesn't show the diffs from the last working version, which is why we might want to consider lesser commits of working code so we can keep a working baseline easier.

> tcp[000000] packet in established, fin: 1, data: 0, setting state to CLOSE_WAIT     <--------------

Yes, it appears that, for whatever reason, a NULL socket ends up dereferencing to a value that ends up in the CLOSE_WAIT switch case every time!

Mellvik:
@ghaerr,

> 
> The fixes after your comments the other day seemed to work. Then they didn't.
> 
> You still have a "good" copy you're testing with though, right? Or did you lose that and we now have to find what happened?
> 
Unfortunately, the good copy turned out to be not so good after the nest reboot.
> I'll take a look at this version and comment. Unfortunately, this PR doesn't show the diffs from the last working version, which is why we might want to consider lesser commits of working code so we can keep a working baseline easier.
> 
> tcp[000000] packet in established, fin: 1, data: 0, setting state to CLOSE_WAIT <--------------
> 
> Yes, it appears that, for whatever reason, a NULL socket ends up dereferencing to a value that ends up in the CLOSE_WAIT switch case every time!
> 
Yes it does, with the oddest of consequences - some times reading or writing a non existing socket, other times tricking ktcp into believing the sequence numbers are off by one, or leaving ftpd hanging in a read.

Thank you!

—M

ghaerr:
> Unfortunately, the good copy turned out to be not so good after the nest reboot.

That is very confusing. Now I don't know whether we have a close_wait problem still, or not.  Remember, in one direction, that is, when the socket is closed by the remote side, it is correct to enter the close_wait state.

Are the bugs only in passive mode? Does active still work great, every time?

Does passive mode fail the first time, or only sometimes? With a buggy program, it may be that we can't do multiple file transfers, as the first one corrupts something, not yet seen. Eliminate variables!

One thing for sure - immediately after finding a NULL dereference, there's absolutely no point in debugging further, as data is corrupted. This was the mistake we made the first time with accept(), by continuing to debug close_wait when we knew accept failed, which was the result of corruption. 

The 8086 never faults on that or data overruns, so we have to be very careful. You may have to add tons of printf statements in ftpd, or possibly in ktcp, just to find out exactly when the data corruption is occurring, since we're pretty certain its in ftpd.

Another option is to take a step back and redesign ftpd so that it doesn't pass buffer pointer around so much. All of the ideas I mentioned in the first code review should probably be done, since we're back into bug city again.

ghaerr:
> #define 	CMDBUFSIZ 	256
> \+ #define 	CMDBUFSIZ 	512

What the heck, did you lose the version you were working on? CMDBUFSIZ has to be equal to IOBUFSIZ, which is 1500. This was the whole original problem!

Mellvik:
Sorry about that @ghaerr!

> 
> Unfortunately, the good copy turned out to be not so good after the nest reboot.
> 
> That is very confusing. Now I don't know whether we have a close_wait problem still, or not. Remember, in one direction, that is, when the socket is closed by the remote side, it is correct to enter the close_wait state.
> 
Mea culpa - I was too fast in casino in the chips on this one. That said I do believe the close_wait problem is fixed.
> Are the bugs only in passive mode? Does active still work great, every time?
> 
Active mode is rock solid. Thousands of transfers, never a hitch.
> Does passive mode fail the first time, or only sometimes? With a buggy program, it may be that we can't do multiple file transfers, as the first one corrupts something, not yet seen. Eliminate variables!
> 
Trust me, I've been through the hoops. There is (yet) no repeatable pattern. I can do numerous file transfers and dir listings at times, then suddenly it bombs. What I'm down to is a reboot between every test which is extremely time consuming.
> One thing for sure - immediately after finding a NULL dereference, there's absolutely no point in debugging further, as data is corrupted. This was the mistake we made the first time with accept(), by continuing to debug close_wait when we knew accept failed, which was the result of corruption.
> 
Agreed!
> The 8086 never faults on that or data overruns, so we have to be very careful. You may have to add tons of printf statements in ftpd, or possibly in ktcp, just to find out exactly when the data corruption is occurring, since we're pretty certain its in ftpd.
> 
> Another option is to take a step back and redesign ftpd so that it doesn't pass buffer pointer around so much. All of the ideas I mentioned in the first code review should probably be done, since we're back into bug city again.
> 
I've been over this one today, eliminated some, checked others carefully. The one thing still on my list is to go back to in_gethostbyname(), just to have been there. PASV was bombing before making that change too, but then again, leave no stone unturned.

Thank you!

—M

Mellvik:
> 
> #define CMDBUFSIZ 256
> + #define CMDBUFSIZ 512
> 
> What the heck, did you lose the version you were working on? CMDBUFSIZ has to be equal to IOBUFSIZ, which is 1500. This was the whole original problem!
> 
We thought it was. It wasn't. So I've experimented with various sizes (larger, smaller) on various arrays looking for patterns and changed symptoms. And other adjustments.  No cigar yet. So the search for a pattern continues.

-M

ghaerr:
My $0.02 on debugging:

> We thought it was. It wasn't. 

When I suggested changing CMDBUFSIZ to IOBUFSIZ = 512, as well as -maout-stack=8192, and one other thing which I've forgotten, your response was that everything started working.

Now, you're back to wondering what the problem is.

First, whenever anything complex starts working suddenly, immediately and always snapshot the source, period. That gives you a restart point should you have to backtrack, like we're doing now.

Second, all three items, including CMDBUFSIZ, must be set back to the values that caused the system to work, even for a little while. This is because, at one point, they caused a non-working system to start working, so we don't want to change them again. Those were the original values you were thinking of when you designed the system. The likely issue is that there is yet an additional problem, but we cannot just assume this isn't it and assign a lesser value to the buffer/stack size, when we know that fixed a problem, for at least a little while.

Change back to 1500, set stack to 8192, add back the other item.

Notice line 616 has a hard-coded command buffer size of 256. What the heck? This needs to be CMDBUFSIZ, and that needs to be 1500. I wouldn't doubt this is the whole problem, again.

We can probably salvage a copy of the "working" version from the PR you opened and was then closed. Then diff that file with what you've got. I would strongly suggest going back to that version and then diff in the new changes, rather than the other way around, until you figure out what is going on. You have added too many new variables to effectively debug this hard-to-find issue by using this vastly updated version, considering you're saying now the first version didn't work either.

Enums start with 0, set CMD_USER = 1 as the first item.

More coming. Still too many functions that return values via arguments. Too many function passed buffers of unknown size.

Mellvik:
@ghaerr,
I appreciate you taking a stab at this. I'd also appreciate a less aggressive tone. I may be rusty, but I've been around the block more than a few times.

And I repeat - what you're seeing is a snapshot. One of very many. You'll find many odd things. Like I said before, it's part of the debugging. My metodology may be different from yours, and quite possibly less effective, but the doesn't make it bad.

Btw the unix/linux (and elks) networking code is full of functions returning values via arguments. Which doesn't make it good coding, but it doesn't make it useless either. 

-M

ghaerr:
> I appreciate you taking a stab at this. I'd also appreciate a less aggressive tone. I may be rusty, but I've been around the block more than a few times.

Sorry, not intentional. I think I've been debugging too much code lately.

> what you're seeing is a snapshot. One of very many.

I was hoping you'd have the first snapshot, directly after it seemed to start working again, since that's what I had previously reviewed. It is easier for me to help when there are less variables, that's all.

> unix/linux (and elks) networking code is full of functions returning values via arguments. 

Removing the output arguments was only suggested because we're likely getting memory corruption via pointers.

Mellvik:
@ghaerr,

> I appreciate you taking a stab at this. I'd also appreciate a less aggressive tone. I may be rusty, but I've been around the block more than a few times.
> 
> Sorry, not intentional. I think I've been debugging too much code lately.
> 
It has been some intense months. Maybe time to take a break for the holidays.
> what you're seeing is a snapshot. One of very many.
> 
> I was hoping you'd have the first snapshot, directly after it seemed to start working again, since that's what I had previously reviewed. It is easier for me to help when there are less variables, that's all.
> 
I did have that. After a reboot, it didn't work anymore - I was back to where I was before. That's the basis for the ftpd we have today, which - as it turns out - work just fine. As is. See below.


I decided it was time for a fresh approach, so I changed a few lines of code in ftpd and updated the qemu.sh. Got a decent FTP client for the Mac - and surprise. Passive mode works just fine. Close_wait works just fine now. There are challenges with fast reuse of sockets of course. But it feels like we've moved a big step forward. Not that the problems revealed by ftpd passive mode are solved, but I guess we know were to look. And we know it's timing related. 

As underscored by yet another setting: When routing traffic from my Mac to ELKS thru 2 intermediate routers, passive mode works almost fine even on the physical hardware. More details later.


—Mellvik

Mellvik:
@ghaerr,

Possibly an issue on its own - or a non-issue altogether:

I'm getting this warning from gcc:

ftpd.c: In function ‘do_pasv’:
ftpd.c:448:58: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘long int’ [-Wformat=]
   sprintf(str, "227 Entering Passive Mode (127,0,0,1,31,%d)\r\n", 105 + (port-PASV_PORT));

The problem is that 'port' is declared 'int'. And if gcc some times treats it as int, some times long int, it's a real problem.

Any ideas?

--Mellvik

> 26. nov. 2021 kl. 16:54 skrev Helge Skrivervik ***@***.***>:
> 
> ***@***.***,
> 
>> I appreciate you taking a stab at this. I'd also appreciate a less aggressive tone. I may be rusty, but I've been around the block more than a few times.
>> 
>> Sorry, not intentional. I think I've been debugging too much code lately.
>> 
> It has been some intense months. Maybe time to take a break for the holidays.
>> what you're seeing is a snapshot. One of very many.
>> 
>> I was hoping you'd have the first snapshot, directly after it seemed to start working again, since that's what I had previously reviewed. It is easier for me to help when there are less variables, that's all.
>> 
> I did have that. After a reboot, it didn't work anymore - I was back to where I was before. That's the basis for the ftpd we have today, which - as it turns out - work just fine. As is. See below.
> 
> 
> I decided it was time for a fresh approach, so I changed a few lines of code in ftpd and updated the qemu.sh. Got a decent FTP client for the Mac - and surprise. Passive mode works just fine. Close_wait works just fine now. There are challenges with fast reuse of sockets of course. But it feels like we've moved a big step forward. Not that the problems revealed by ftpd passive mode are solved, but I guess we know were to look. And we know it's timing related. 
> 
> As underscored by yet another setting: When routing traffic from my Mac to ELKS thru 2 intermediate routers, passive mode works almost fine even on the physical hardware. More details later.
> 
> 
> —Mellvik
> 
> 
> 
> 
>

ghaerr:
> ftpd.c:448:58: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘long int’ [-Wformat=]
   sprintf(str, "227 Entering Passive Mode (127,0,0,1,31,%d)\r\n", 105 + (port-PASV_PORT));

> The problem is that 'port' is declared 'int'. And if gcc some times treats it as int, some times long int, it's a real problem.

The reason is that PASV_PORT is 48921, which is outside the range of 'int', being larger than 32767. Thus, it is treated as a compiler as 'long'. When long and int are mixed in a C expression, the entire expression is converted to long. It would have to be defined as 48921U in order to work as you have coded.

Lesser issues are that PASV_PORT is defined with a semicolon, 'port' should be an unsigned int, in the case it is ever used in a comparison, due to C's unsigned rules precedence, and the '%d' designator should be '%u' for unsigned, since port values are unsigned (values from 1-65535 allowed).

BTW, what was the fix you made that made ftpd mostly work again?

Mellvik:
@ghaerr - 
> 
> ftpd.c:448:58: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘long int’ [-Wformat=]
> sprintf(str, "227 Entering Passive Mode (127,0,0,1,31,%d)\r\n", 105 + (port-PASV_PORT));
> 
> The problem is that 'port' is declared 'int'. And if gcc some times treats it as int, some times long int, it's a real problem.
> 
> The reason is that PASV_PORT is 48921, which is outside the range of 'int', being larger than 32767. Thus, it is treated as a compiler as 'long'. When long and int are mixed in a C expression, the entire expression is converted to long. It would have to be defined as 48921U in order to work as you have coded.
> 
oops, thank you - I should have caught that one!
> Lesser issues are that PASV_PORT is defined with a semicolon, 'port' should be an unsigned int, in the case it is ever used in a comparison, due to C's unsigned rules precedence, and the '%d' designator should be '%u' for unsigned, since port values are unsigned (values from 1-65535 allowed).
> 
> BTW, what was the fix you made that made ftpd mostly work again?
> 
That's the interesting thing: the only changes was to accommodate for Qemu - and a couple of minor things like the semicolon you mentioned. So the same code works perfectly well on Qemu and flaky on Compaq when connected directly to the Raspberry. And almost stable between the mac and the Compaq with a couple of routers in between, causing (I guess) enough delay to make a difference.

Which doesn't necessarily exonerate ftpd, but it does seem like there may be a timing problem somewhere. How that could create the type of problems we're seeing, is a mystery though.

BTW I've been having some fun with ELKS back-to-back this weekend. A 286 and a 386. Very interesting, not entirely smooth sailing, but telnet between them is flawless. More about that later.


Thank you!

-M

ghaerr:
> So the same code works perfectly well on Qemu and flaky on Compaq when connected directly to the Raspberry. 



In the last round, we found a null pointer (socket) which was displayed in the ktcp debug statements. I can possibly add code in the kernel to check for a proper socket pointer, possibly also in ktcp, although it would likely be a lot of work.

The funny thing is that ftpd doesn't allocate any memory, so I don't think the null pointer is related to the typical no memory scenario. There are lots of copies to buffer pointers though, so we should set all buffers to the same size for the time being.

However, there are a number of `bzero` function calls in ftpd, which could be overwriting data with zeros, please take a close look at them.

Did you change the single cmd buf back from 256 to 1500, as well as the others? I'd like you to make all the suggestions I made in https://github.com/jbruchon/elks/pull/1030#issuecomment-979439941 prior to posting a PR with the continuing issue documented as to its behavior and the source code associated with it. I can then try to track it down. Frankly, I still suspect ftpd as the problem, but of course, it could be anything. We haven't had a buffer overrun issue in ktcp for quite some time.

> And almost stable between the mac and the Compaq with a couple of routers in between, causing (I guess) enough delay to make a difference.

I'm guessing you're saying that the problem resulting less frequently with a different timing issue results the same null pointer problem?

> but it does seem like there may be a timing problem somewhere. How that could create the type of problems we're seeing, is a mystery though.

The socket pointer is passed as another part of a tcpdev buffer along with received or sending data. I have no idea yet how it could be overwritten, other than perhaps as part of a sophisticated kernel locking problem between competing reads and writes. However, with only one application (ftpd) running on ELKS, this shouldn't be an issue. 

It is also remotely possible that the larger buffer sizes we're now allowing could be overrunning a kernel buffer due to a kernel coding error. The way we'd check this would be to change ALL buffer sizes in ftpd back to, say, 512, and see what happens. We definitely need to have all buffer sizes be the same size at this point to eliminate the variable of them writing over each other. Thinking about this, it seems that the kernel is limiting writes to 1492 and has to loop to get the data sent. So we should definitely try a smaller buffer size, perhaps 1492 and/or 512 to continue to see what happens on your side.

Thank you!

ghaerr:
@Mellvik,

My last response considering the possibility of a kernel "locking" problem, combined with your previous statements that changing the buffer sizes did not fix the issue, got me thinking a bit more about how a socket pointer could get zeroed.

Looking at the kernel include/linuxmh/tcpdev.h structures, which are used to communicate data to and from ktcp, show that there are a couple cases where the socket pointer and the return value are not in the same order. In particular, the cases for read, write, return codes and accept differ:
```
struct tdb_read {
    unsigned char cmd;
    struct socket *sock;
    int size;
    int nonblock;
};

struct tdb_write {
    unsigned char cmd;
    struct socket *sock;
    int size;
    int nonblock;
    unsigned char data[TDB_WRITE_MAX];
};
struct tdb_return_data {
    char type;
    int ret_value;
    struct socket *sock;
    int size;
    unsigned char data[];
};

struct tdb_accept_ret {
    char type;
    int ret_value;
    struct socket *sock;
    __u32 addr_ip;
    __u16 addr_port;
};
```
While not in itself buggy, if, for instance, there were a locking bug where a read or write was "interrupted" by a different process which just performed an accept, you can see that the accept return value could overwrite a read/write socket pointer, possibly with zero. The kernel uses the same buffer for all input from tcpdev, so locking is required. It would be great to allocate the buffers dynamically, but I'm afraid we might run out of memory too easily (using the kernel local heap).

The kernel has to "lock" other processes out of interfering with various operations, but the asynchronous non-normal nature of ktcp has been increasing lately, especially when we're talking about async accept's occurring during another read or write. I would like to hear from you exactly what your test is that causes the error, as we need to understand exactly what processes are running on ELKS, other than ftpd. I am assuming that ftpd itself doesn't fork or run any network transfers asynchronously.

I don't really think this is probably the issue, but the interactions between ktcp and the /etc/tcpdev kernel driver are very tricky. I could add debug code that verifies a returned socket pointer every time (which should probably have been done anyways), to show this. I would like to do this after you've had a chance to change all buffer sizes from 1500 to, perhaps 1492 or 512, to see whether that makes any difference. If not, then this path is probably worth pursuing.

Changing the structures won't fix the problem, but would only hide the problem further, while possibly not overwriting the socket, it could overwrite the socket with a different process' socket pointer, for instance. Again, this is only a guess, given what you've said about my suggestions definitely not solving the problem you're seeing.

Keep me posted, thank you.

Mellvik:
> So the same code works perfectly well on Qemu and flaky on Compaq when connected directly to the Raspberry.
> 
> In the last round, we found a null pointer (socket) which was displayed in the ktcp debug statements. I can possibly add code in the kernel to check for a proper socket pointer, possibly also in ktcp, although it would likely be a lot of work.
> 
I agree, we're not there yet.
> The funny thing is that ftpd doesn't allocate any memory, so I don't think the null pointer is related to the typical no memory scenario. There are lots of copies to buffer pointers though, so we should set all buffers to the same size for the time being.
> 
Actually, there is only one buffer - iobuf, always local to the routine using it. And always the same size. There are string arrays - and I've been through each and every one serveral times without much succes thus far.
> However, there are a number of bzero function calls in ftpd, which could be overwriting data with zeros, please take a close look at them.
> 
Hmmm, I was considering memset() instead, but decided I need to look at the source for bzero first, to make sure it's not a waste of time.
> Did you change the single cmd buf back from 256 to 1500, as well as the others? I'd like you to make all the suggestions I made in #1030 (comment) prior to posting a PR with the continuing issue documented as to its behavior and the source code associated with it. I can then try to track it down. Frankly, I still suspect ftpd as the problem, but of course, it could be anything. We haven't had a buffer overrun issue in ktcp for quite some time.
> 
Like I said, all buffers and arrays have been through the hoops  several times, ridicolously low and high.
> And almost stable between the mac and the Compaq with a couple of routers in between, causing (I guess) enough delay to make a difference.
> 
> I'm guessing you're saying that the problem resulting less frequently with a different timing issue results the same null pointer problem?
> 
Yes, the instabilty (unpredictabilty) is a killer. Like this:
- the first time after boot, a certain sequence of commands will always fail.
- the next run, same daemon, new fork, that sequence of commands will work fine, so will other sequences - until not
- and the behaviour over the next tries is completely unpredictable
Restarting ktcp will not bring us back to the boot state.
Indicating that some state has changed at the kernel level and only a reboot will reset it.
> but it does seem like there may be a timing problem somewhere. How that could create the type of problems we're seeing, is a mystery though.
> 
> The socket pointer is passed as another part of a tcpdev buffer along with received or sending data. I have no idea yet how it could be overwritten, other than perhaps as part of a sophisticated kernel locking problem between competing reads and writes. However, with only one application (ftpd) running on ELKS, this shouldn't be an issue.
> 
> It is also remotely possible that the larger buffer sizes we're now allowing could be overrunning a kernel buffer due to a kernel coding error. The way we'd check this would be to change ALL buffer sizes in ftpd back to, say, 512, and see what happens. We definitely need to have all buffer sizes be the same size at this point to eliminate the variable of them writing over each other. Thinking about this, it seems that the kernel is limiting writes to 1492 and has to loop to get the data sent. So we should definitely try a smaller buffer size, perhaps 1492 and/or 512 to continue to see what happens on your side.
> 
This is a new idea. I'll try that.
ANyway - I find it really hard to explain that it all works fine on qemu. Som comparing packet traces is on the agenda too.

Thank you.

--Mellvik
>

ghaerr:
@Mellvik, I've posted a PR #1035 that will show more data to hopefully illuminate when the socket is getting set to null, as well as a potential race condition fix for accept. It seems there may be a problem when there are lots of accepts back-to-back, we will see.

If the socket is going to zero, this should help chase down the problem. This really depends on whether the socket is always null in the error conditions you describe above. If this doesn't help, I will push more debug statements to chase down exactly when the socket is going to zero. 

The other potential problem could be when a socket is closed on the remote end, and ftpd issues another accept before all the data is received. This could occur if ftpd did not close the network file descriptor but issued another accept. This could also be possible if the command channel file descriptor is closed by ftpd, but the remote sends a reply that is unread by the time a subsequent accept is executed by ftpd. I haven't added debug for that case yet, as I can't see where it might use a zero socket. So at this point we need to chase down exactly when the socket get zeroed by either ktcp or the kernel. It appears that only ktcp is seeing a null socket, not the kernel. You might turn on DEBUG_NET in debug.h to see where the socket may be null in the kernel in the case #1035 does not fix the problem.

Thank you!

PR #1036, opened 2021-11-30
Ftpd: Passive mode and much more

Mellvik:
`ftpd` now supports passive mode. Other enhancements - in addition to some bugfixes - include:
- Commands: Delete file, make directory, remove directory
- New debug stuff, activated by the `-d` option
- The `-d` option prevents disconnect from the controlling terminal
- A new -q option enables `ftpd` to be run in QEMU i passive mode, using the external port numbers from 49821-49825.

Tested under QEMU and on physical hardware.
Pending enhancements include faster multifile transfers when the SO_REUSEADDR socket option has been tested.

### DIFF
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -1,4 +1,3 @@
-
 /*
  * Minimal FTP server for ELKS
  * November 2021 by Helge Skrivervik - helge@skrivervik.com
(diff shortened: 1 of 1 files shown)


ghaerr:
@Mellvik,

What a fantastic enhancement for ELKS, thank you very much!! All the work you've put in writing and testing ftpd has really helped ELKS networking become very reliable, as well as added an important capability for getting data into and out of ELKS. Super job!!

I'll commit this now along with #1035 to provide a known working baseline. I'm looking forward to learning how well this works with ftpget and ftpput, which I'll test.

A quick question - what's the "qemu hack" `sleep(1)`? Is that required for operating on QEMU? Are there any special instructions for running/testing on QEMU? I haven't used your port forwarding script yet but will after commit.

Thank you!

Mellvik:
@ghaerr,
IIRC the sleep(1) in the qemu hack was necessary to slow down the reuse of ports, otherwise we'll just run through the available 5 and run out of memory immediately - because the emulator is so fast.

If you run ftpd -q under QEMU and use the changes I send for qemu.sh, you can ftp into elks by
ftp 127.0.0.1 8021
and then do whatever as long as you use passive mode. the example shows what happens if you don't. In ftp, use help or ? to get the commands.

Thank you!


—M

www9:elks helge$ ftp 127.0.0.1 8021
Connected to 127.0.0.1.
220 Welcome - ELKS minimal FTP server speaking.
Name (127.0.0.1:helge): <—— just enter, or type anything 
331 User OK.
Password: <— just enter or type anything 
230 Password OK.
ftp> dir /
425 Can't open data connection.
503 Unknown command 
503 Bad sequence of commands.
ftp> passi
Passive mode on.
ftp> dir /
227 Entering Passive Mode (127,0,0,1,31,105)
150 Opening ASCII mode data connection for /bin/ls.
drwxr-xr-x  2 root     root         1984  Nov 30 10:02 bin
-rw-r--r--  1 root     root          214  Nov 12 12:30 bootopts
-rw-r--r--  1 root     root          581  Oct 13 08:49 bootopts.sampl
drwxr-xr-x  2 root     root          784  Nov 30 10:02 dev
drwxr-xr-x  3 root     root          208  Nov 30 10:02 etc
drwxr-xr-x  2 root     root           32  Nov 30 10:02 home
drwxr-xr-x  2 root     root           48  Nov 30 10:02 lib
-rwxr-xr-x  1 root     root        75896  Nov 30 10:02 linux
drwxr-xr-x  2 root     root           32  Nov 30 10:02 mnt
drwxr-xr-x  2 root     root           32  Nov 30 10:02 root
drwxr-xr-x  2 root     root           32  Nov 30 10:02 tmp
drwxr-xr-x  4 root     root           64  Nov 30 10:02 var
226 Transfer Complete.
ftp> 

...

Mellvik:
Thank you @ghaerr,
It seems every new project opens a new can of worms. :-)

 It wasn't intended to be that way, but it ended up being a very joint effort, and a very educating one (again). Remote debugging has challenges, but as you point out, what came out of it makes it all worthwhile.

Armed with the satisfaction of nailing this one, I guess we're ready for another one? I have a feeling the back-to-back exercise  has some  surprises up its sleeve. 
We'll know soon enough!

Thaks again.

-M

ghaerr:
> It seems every new project opens a new can of worms. :-)

Yes. Sadly though, this has been the realistic state of where TCP on ELKS has been for years. It seems it's a long haul getting a TCP stack throughly debugged, but I am learning a lot :)

> Armed with the satisfaction of nailing this one, I guess we're ready for another one? I have a feeling the back-to-back exercise  has some  surprises up its sleeve.

I hope not too ugly, but we'll see.

I just got the remote ftp talking to ftpd with your QEMU port forwarding magic, thank you. I will play with that a while and get more familiar with what you've created. I've already ran into the problem of ftpd not working with QEMU without -q. We definitely need to get that fixed, which I'll look into. Also need a way to start ftpd with debugging on at boot.

How do you switch between passive and active modes on remote ftp? Is that just the "pasv" command?

Thank you!

Mellvik:
@ghaerr,
> It seems every new project opens a new can of worms. :-)
> 
> Yes. Sadly though, this has been the realistic state of where TCP on ELKS has been for years. It seems it's a long haul getting a TCP stack throughly debugged, but I am learning a lot :)
> 
Likewise. And I forgot to mention yesterday: Yes, ftpd and the bugfixes are important contirbutions, but the enhancements are just as important - such as adding getsockname, fixing return values in accept() etc.
> Armed with the satisfaction of nailing this one, I guess we're ready for another one? I have a feeling the back-to-back exercise has some surprises up its sleeve.
> 
> I hope not too ugly, but we'll see.
> 
I think - from what I've seen this far (and from reasonable expectations), that what we'll find will be timing related - overruns, retransmits, lost packets etc. I'm rather looking forward to it, and ugly, no I don't think so :-)
> How do you switch between passive and active modes on remote ftp? Is that just the "pasv" command?
> 
It's like the 'prompt' command in ftp, you toggle it by repeating the command. And the command is 'passive' - all commands can be shortened as long as they're still unique. Thus the real DIR command is 'directory'. I cannot promise such luxury in the ELKS ftp vlient, but then, some of the functions are really low threshold. 

--Mellvik

PR #1039, opened 2021-12-01
[net] First pass fixing ftpd on QEMU

ghaerr:
Enhancements to allow easier operation of ftpd on QEMU, for testing.

@Mellvik, tell me what you think about this. I'm already finding a number of potential issues between ftpd and QEMU.
The following changes are made in this PR:
- SO_REUSEADDR turned on for passive mode file transfers (both QEMU and real hardware). This should be safe, but still should be tested on real hardware at some point. SO_REUSEADDR should not yet be enabled elsewhere (pending more work on it).
- Eliminate 1 second sleep in ftpd for socket timeout, uses SO_REUSEADDR and displays appropriate messages.
- Added ability to set ftpd -q or ftpd -q in /bootopts: use "ftpd=-q" when running QEMU for testing. Use "ftpd=-d" for debug output at boot without having to kill and restart.
- ftpd -q and -d options don't disable becoming a daemon; run ftpd -d -d for old behavior.
- Added ftp port forwarding lines in qemu.sh. It would be nice to figure out how to do this without having to kluge ftpd to prepare a special passive mode string!
- Forward 9 ports rather than 5 for more realistic emulation. (May be off-by-one error in ftpd on wrap).

In quick testing, I found the following issues (could be my mistakes):
- managed to send files to ELKS, using "mput foo*", but can't get "mget foo*" to work to send from ELKS afterwards.
- QEMU seems to stop working after a bit... the INT 0 timer stops, and ftpd stops working, only when using serial console. Very strange, but have seen this behavior before. I think its a QEMU problem. This can be duplicated by uncommenting the first two entries in /bootopts and running a few file transfers.
- Sometimes ftpd just stops responding. Not sure if that's the above problem or not.
- Any ideas on how active mode might be tested using QEMU?

Overall, this first pass makes things a lot easier for testing. @Mellvik, let me know if you're OK with commit.

### DIFF
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -29,6 +29,7 @@
 #define		FALSE		0
 #define		FTP_PORT	21
 #define		PASV_PORT	49821U	/* 'random' port for passive connections */
+#define		QEMU_PORT	8041U	/* outside port for QEMU */
 #define		BLOAT
 
 #ifndef MAXPATHLEN
@@ -397,7 +398,6 @@ int do_pasv(int controlfd, int *datafd) {
 	if (setsockopt(fd, SOL_SOCKET, SO_RCVBUF, &i, sizeof(int)) < 0)
                 perror("SO_RCVBUF");
 
-	//bzero(&pasv, sizeof(pasv));
 	pasv.sin_family = AF_INET;
 	pasv.sin_addr.s_addr = htonl(INADDR_ANY);
 	pasv.sin_port = htons(port);
@@ -414,9 +414,7 @@ int do_pasv(int controlfd, int *datafd) {
 		} 
 		port++;
 		if (qemu) {
-			//sleep(1);
-			//usleep(800);	/* Experimental */
-			if (port >= (PASV_PORT + 8)) port = PASV_PORT;
+			if (port >= (PASV_PORT + 9)) port = PASV_PORT;
 		}
     		pasv.sin_port = htons(port);
 	}
@@ -438,8 +436,11 @@ int do_pasv(int controlfd, int *datafd) {
 	sprintf(str, "227 Entering Passive Mode (%d,%d,%d,%d,%d,%d)\r\n", UC(a[0]),
 		UC(a[1]), UC(a[2]), UC(a[3]), UC(p[0]), UC(p[1]));
 
-	if (qemu)
-		sprintf(str, "227 Entering Passive Mode (127,0,0,1,31,%u)\r\n", 105 + (port-PASV_PORT));
+	if (qemu) {
+		unsigned int qport = QEMU_PORT + port - PASV_PORT;
+		sprintf(str, "227 Entering Passive Mode (127,0,0,1,%u,%u)\r\n",
+			htons(qport)&0xff, htons(qport)>>8);
+	}
     	write(controlfd, str, strlen(str));
 	if (debug) printf("%s", str);
 	i = sizeof(pasv);
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -391,8 +391,8 @@ int do_pasv(int controlfd, int *datafd) {
 		return -1;
 	}
 
-	//if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &i, sizeof(i)) < 0) 
-		//perror("SO_REUSEADDR");
+	if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &i, sizeof(i)) < 0)
+		perror("SO_REUSEADDR");
 	i = SO_LISTEN_BUFSIZ;
 	if (setsockopt(fd, SOL_SOCKET, SO_RCVBUF, &i, sizeof(int)) < 0)
                 perror("SO_RCVBUF");
@@ -414,9 +414,9 @@ int do_pasv(int controlfd, int *datafd) {
 		} 
 		port++;
 		if (qemu) {
-			sleep(1);
+			//sleep(1);
 			//usleep(800);	/* Experimental */
-			if (port >= (PASV_PORT + 4)) port = PASV_PORT;
+			if (port >= (PASV_PORT + 8)) port = PASV_PORT;
 		}
     		pasv.sin_port = htons(port);
 	}
@@ -598,7 +598,7 @@ int main(int argc, char **argv){
 		perror("Error in listen");
 		exit(3);
 	}
-	if (!debug) {
+	if (debug < 2) {
 		/* become daemon, debug output on 1 and 2*/
 		if ((ret = fork()) == -1) {
 			fprintf(stderr, "ftpd: Can't fork to become daemon\n");
--- a/elkscmd/ktcp/tcp.c
+++ b/elkscmd/ktcp/tcp.c
@@ -460,7 +460,7 @@ void tcp_process(struct iphdr_s *iph)
     }
 
     if (!cbnode) {
-	printf("tcp: refusing packet %s:%u->%d\n", in_ntoa(iph->saddr),
+	printf("tcp: refusing packet from %s:%u to :%u\n", in_ntoa(iph->saddr),
 		ntohs(tcph->sport), ntohs(tcph->dport));
 
 	if (tcph->flags & TF_RST)
--- a/elkscmd/rootfs_template/bin/net
+++ b/elkscmd/rootfs_template/bin/net
@@ -60,7 +60,7 @@ start_network()
 	echo $ktcp
 	if $ktcp; then
 		echo -n "Starting daemons "
-		for daemon in telnetd httpd ftpd
+		for daemon in telnetd httpd
 		do
 			if test -f /bin/$daemon
 			then
@@ -68,6 +68,11 @@ start_network()
 				$daemon || true
 			fi
 		done
+		if test -f /bin/ftpd
+		then
+			echo -n "ftpd $ftpd"
+			ftpd $ftpd || true
+		fi
 		echo ""
 	else
 		echo "Network start failed"
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -1,8 +1,8 @@
 ##
 #console=ttyS0 debug net=eth 3
-#ftpd=-q
+#ftpd=-q net=eth
 #init=/bin/init 3 n	# multiuser serial no sysinit
 #init=/bin/sh		# singleuser shell
 #console=ttyS0,19200	# serial console
-#root=hda1 ro	# hd partition 1, read-only root
-#netirq=9 netport=0x300	# NIC parms
+#root=hda1 ro		# hd part 1, read-only root
+#netirq=9 netport=0x300
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -1,7 +1,8 @@
 ##
-#console=ttyS0 debug net=eth 3 # condensed
+#console=ttyS0 debug net=eth 3
+#ftpd=-q
 #init=/bin/init 3 n	# multiuser serial no sysinit
 #init=/bin/sh		# singleuser shell
 #console=ttyS0,19200	# serial console
 #root=hda1 ro	# hd partition 1, read-only root
-#netirq=9 netport=0x300	# set NIC parms
+#netirq=9 netport=0x300	# NIC parms
--- a/qemu.sh
+++ b/qemu.sh
@@ -55,7 +55,7 @@ KEYBOARD=
 #SERIAL="-chardev msmouse,id=chardev1 -device isa-serial,chardev=chardev1,id=serial1"
 
 # Uncomment this to route ELKS /dev/ttyS0 to host terminal
-#CONSOLE="-serial stdio"
+CONSOLE="-serial stdio"
 # Hides qemu window also
 #CONSOLE="-serial stdio -nographic"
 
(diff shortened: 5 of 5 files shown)


Mellvik:
@ghaerr,
Thank you for diving right into this!

I concur with most of this - and have a few comments:
> Enhancements to allow easier operation of ftpd on QEMU, for testing.
> 
> @Mellvik, tell me what you think about this. I'm already finding a number of potential issues between ftpd and QEMU.
> The following changes are made in this PR:
> 
> SO_REUSEADDR turned on for passive mode file transfers (both QEMU and real hardware). This should be safe, but still should be tested on real hardware at some point. SO_REUSEADDR should not yet be enabled elsewhere (pending more work on it).
> Eliminate 1 second sleep in ftpd for socket timeout, uses SO_REUSEADDR and displays appropriate messages.
> Added ability to set ftpd -q or ftpd -q in /bootopts: use "ftpd=-q" when running QEMU for testing. Use "ftpd=-d" for debug output at boot without having to kill and restart.
> ftpd -q and -d options don't disable becoming a daemon; run ftpd -d -d for old behavior.
I guess this is a matter of taste. I'd prefer -D instead of -d -d - but I'm fine with this.
> Added ftp port forwarding lines in qemu.sh. It would be nice to figure out how to do this without having to kluge ftpd to prepare a special passive mode string!
I don't see how you can do this. Unless someone dives into QEMU and figures out how to configure it to use TUN/TAP and thus have its own interface on the system. The problem is - obviously - that the addresse the server sees are different from what the client sees. Always.
> Forward 9 ports rather than 5 for more realistic emulation. (May be off-by-one error in ftpd on wrap).
This hack (4+1) was to avoid ktcp crashing because too many sockets were lingering. If SO_REUSEADDR works, this should be OK.
> In quick testing, I found the following issues (could be my mistakes):
> 
> managed to send files to ELKS, using "mput foo*", but can't get "mget foo?" to work to send from ELKS afterwards.
Ftpd does not implement globbing (yet). So what you'd have to do is 'mget .' (or some other directory). This is on the todo list.
> QEMU seems to stop working after a bit... the INT 0 timer stops, and ftpd stops working, only when using serial console. Very strange, but have seen this behavior before. I think its a QEMU problem. This can be duplicated by uncommenting the first two entries in /bootopts and running a few file transfers.
Interesting. I haven't seen this, but then again I haven't used qemu much ...
> Sometimes ftpd just stops responding. Not sure if that's the above problem or not.
Hopefully we find a way to recreate this one. I haven't seen it, so it's likely the speed difference between systems that triggers something.
> Any ideas on how active mode might be tested using QEMU?
I honestly don't think this is possible without a QEMU hacked ftp client which makes the port # predictable.
> Overall, this first pass makes things a lot easier for testing. @Mellvik, let me know if you're OK with commit.
> 

Go ahead, @ghaerr - this is great!

--Mellvik
>

ghaerr:
> I guess this is a matter of taste. I'd prefer -D instead of -d -d - but I'm fine with this.

Go ahead and change it however you like. I went the easy route and didn't rewrite the option parsing, instead just use "debug < 2" as to when to fork(). My primary purpose was to try to get a working QEMU test platform for ftpd so I can more easily exercise it. I would like to move towards using `net start/stop` or other options to start/stop/debug with daemons so we don't have to debug with special "harnesses".

> This hack (4+1) was to avoid ktcp crashing because too many sockets were lingering. If SO_REUSEADDR works, this should be OK.

SO_REUSEADDR seems to be working well, for this use case. I increased the number of usable ports so that the QEMU platform is more like real hardware and might fail in the same way. Also just added a QEMU_PORT to specify the outside port number for the time being, until we figure out how to do away with special qemu hacks.

I have noticed a couple of cases where netstat shows CLOSE_WAIT for some sockets. I don't think this is buggy in ktcp, as CLOSE_WAIT means FIN was received, but local socket not closed. This may be because of an error situation not yet handled properly by ftpd. 

Now that QEMU can be easily used to test ftpd, we can more quickly track down possible errors in ftpd.

> So what you'd have to do is 'mget .' (or some other directory)

I've now tested transfers both ways with QEMU, and things seems to be working well, at least for the case of ELKS is running single-user. I have looked into the ELKS multiuser case a bit more, and QEMU has known problems where the hardware timer stops ticking. This seems to manifest itself on ELKS whenever the serial port is emulated. So I have modified the /bootopts test line for QEMU to run in single user mode, and I haven't seen any ftpd "stopping" issues anymore.

Here's some discussion on the QEMU buggy hardware timer emulation issues:
https://github.com/zephyrproject-rtos/zephyr/issues/14173
https://github.com/zephyrproject-rtos/zephyr/issues/12553

Sometimes I think we might want to use a different emulator. When the bug hits, the ELKS "date" command shows time isn't incrementing. Being forced to use console-only mode to test ftpd means that lots of debug information has already scrolled off the screen.

> I honestly don't think this is possible without a QEMU hacked ftp client which makes the port # predictable.

Kind of amazing, is this the case with all emulators? Say even for MSDOS or Windows emulated, one can't ftp in to a known port without MSDOS or Windows server being specially modified?

It would be really nice if we could keep special QEMU-cased hacks out of ELKS applications (and bootopts)!

Mellvik:
@ghaerr,
> I went the easy route and didn't rewrite the option parsing, instead just use "debug < 2" as to when to fork(). My primary purpose was to try to get a working QEMU test platform for ftpd so I can more easily exercise it. I would like to move towards using net start/stop or other options to start/stop/debug with daemons so we don't have to debug with special "harnesses".
> 
I'm fine with this. And option parsing is far down on my list too.
> This hack (4+1) was to avoid ktcp crashing because too many sockets were lingering. If SO_REUSEADDR works, this should be OK.
> 
> SO_REUSEADDR seems to be working well, for this use case. I increased the number of usable ports so that the QEMU platform is more like real hardware and might fail in the same way. Also just added a QEMU_PORT to specify the outside port number for the time being, until we figure out how to do away with special qemu hacks.
> 
This is good. I'll reenable SO_REUSEADDR if you haven't done so already.
> I have noticed a couple of cases where netstat shows CLOSE_WAIT for some sockets. I don't think this is buggy in ktcp, as CLOSE_WAIT means FIN was received, but local socket not closed. This may be because of an error situation not yet handled properly by ftpd.
> 
I'll keep an eye on this as testing contines.
> Now that QEMU can be easily used to test ftpd, we can more quickly track down possible errors in ftpd.
> 
> So what you'd have to do is 'mget .' (or some other directory)
> 
> I've now tested transfers both ways with QEMU, and things seems to be working well, at least for the case of ELKS is running single-user. I have looked into the ELKS multiuser case a bit more, and QEMU has known problems where the hardware timer stops ticking. This seems to manifest itself on ELKS whenever the serial port is emulated. So I have modified the /bootopts test line for QEMU to run in single user mode, and I haven't seen any ftpd "stopping" issues anymore.
> 
> Here's some discussion on the QEMU buggy hardware timer emulation issues:
> zephyrproject-rtos/zephyr#14173
> zephyrproject-rtos/zephyr#12553
> 
> Sometimes I think we might want to use a different emulator. When the bug hits, the ELKS "date" command shows time isn't incrementing. Being forced to use console-only mode to test ftpd means that lots of debug information has already scrolled off the screen.
> 
Well, this is an interesting line of thinking. What are the alternatives btw? Is the qemu problem general (any serial port) or only serial console. Maybe just too complicated, but now that ramdisks work, a dmesg like file in ram may possibly alleviate the problem...
Btw, I've run a few multiuser tests w/o any particular issues, other than more retransmits. Will do more of that now that we're basically stable.
> I honestly don't think this is possible without a QEMU hacked ftp client which makes the port # predictable.
> 
> Kind of amazing, is this the case with all emulators? Say even for MSDOS or Windows emulated, one can't ftp in to a known port without MSDOS or Windows server being specially modified?
> 
This is why passive mode is dominant. Active mode has seen very little use since NATs became the rule. Which was why passive mode was introduced in the first place. That said, and like I've mentioned before, the problem isn't the fact that we're emulating, it's how we emulate the network i/f. If qemu can use tun/tap and we can figure out how to use it, we're ok.
> It would be really nice if we could keep special QEMU-cased hacks out of ELKS applications (and bootopts)!
> 
Absolutely!!

Thank you.

-M

ghaerr:
> This is good. I'll reenable SO_REUSEADDR if you haven't done so already.

It's already done, but only for the passive file transfer mode. It is required for QEMU because of the speed issues you ran into when getting ftpd to work with it. In almost all cases, due to the TCP state machine, (in this use case only), the previous socket used for passive mode is in the TIME_WAIT state, so it can be reused without harm.

The other SO_REUSEADDRs are still commented out, until the long-term fix in ktcp is made. That fix requires that all sockets in the active list be inspected, not just the first socket found as is currently done. It's kind of tricky to get right, which is why I'm delaying dealing with it until we get most of the other networking stabilized (again).

> What are the alternatives btw? Is the qemu problem general (any serial port) or only serial console.

I don't know! It occurs randomly when using the serial port/console with multiuser enabled. It's deep down having to do with edge vs level PIC triggering in the emulator. I've known about it for a long time. Alternatives are we use another emulator possibly. We don't see it when using the serial console usually because typing on the serial console fixes it. It only happens when accessing ELKS without using the serial port, like from a remote ftp. Not to worry, just mentioning the problem, since this PR is for use with QEMU, in case you run into weird stuff.

ghaerr:
> Btw, I've run a few multiuser tests w/o any particular issues, other than more retransmits. Will do more of that now that we're basically stable.

I would say we should be pretty stable. I'm interested in your testing with ftpget/put, since we now may be able to use those and ftpd to download a kernel automatically as you desired a while back. And it would be interesting to test that over QEMU as well.

I'm thinking about the idea of some preset file transfer regression tests for TCP. They would use a preloaded directory of the git repo, and could be run some shell scripts. Even if only over QEMU, or having shell scripts that worked from externally as well, these would go a long way to proving reliability with ELKS TCP moving forward. Ideas welcome!

Mellvik:
@ghaerr - this is indeed a new level for elks networking!

Apropos emulators, we may take a shot at virtualbox. 
IIRC there is a writeup on that in the Wiki already...
> 
> Btw, I've run a few multiuser tests w/o any particular issues, other than more retransmits. Will do more of that now that we're basically stable.
> 
> I would say we should be pretty stable. I'm interested in your testing with ftpget/put, since we now may be able to use those and ftpd to download a kernel automatically as you desired a while back. And it would be interesting to test that over QEMU as well.
> 
Yes, that's an interesting thought. The manual variant has been invaluable for the past year (or has it been two since we got the ethernet driver and basic ktcp stability going?), now automating it is indeed possible. BTW - the availability of ftpd has already helped: I can now ftp new binaries directly to elks instead of the previous 2step process - first scp to the raspi, then ftp or ftpget. Which reminds me - the File transfer wiki needs to be updated. I'm on it.
And - totally off topic - is a paragraph on using minix fsck for the Wiki on your (endless) list?
> I'm thinking about the idea of some preset file transfer regression tests for TCP. They would use a preloaded directory of the git repo, and could be run some shell scripts. Even if only over QEMU, or having shell scripts that worked from externally as well, these would go a long way to proving reliability with ELKS TCP moving forward. Ideas welcome!
> 
This is a good idea. FWIW - my test setup has been the ELKS binaries + a 200k file (created using dd) and 100 2k files created by splitting the 200k file using split(). And finally a couple of zero size files. Easy to see anomalies with the bare eye, easy to verify to the last bit (the created files are all zeroes - dd if=/dev/zero count=200 bs=1k ...etc.)
> 
Thank you!

-M

Mellvik:
@ghaerr,
a quick heads up: While testing the new fpt client. I ran into an interesting situation:
FTP transfers into ELKS in QEMU using ftpd are not 100% reliable. Bytes get added and I haven't had time to look at it yet.  Keep an eye on transferred file sizes.
Not repeatable on real hardware. More on this as soon as I get a chance.

Teaser: The ELKS ftp client is transferring files and directory listings  in passive mode, the first cut is just around the corner - with some fun stuff for experimentation : select() multiplexing.

-M

ghaerr:
> FTP transfers into ELKS in QEMU using ftpd are not 100% reliable. Bytes get added and I haven't had time to look at it yet. Keep an eye on transferred file sizes.

Should this be a real problem, open a new issue. Please include exact details concerning the transfer: is this elks-to-elks, or remote-to-elks; only using the new ftp client, or repeatable with macOS client, and which way the file transfer is going send or receive from ftpd, etc. In this way I can understand whether this is related to the FIN received with unprocessed data that was recently a problem (and now has debug printf turned off), or another problem.

If the problem only occurs on QEMU, we will also need to see whether this has anything to do with SO_REUSEADDR sockets or not. Most of that debug printf is still turned on.

> the first cut is just around the corner - with some fun stuff for experimentation : select() multiplexing.

Both telnet and ktcp use select multiplexing for their network I/O, in case you haven't already noticed.

Mellvik:
> FTP transfers into ELKS in QEMU using ftpd are not 100% reliable. Bytes get added and I haven't had time to look at it yet. Keep an eye on transferred file sizes.
> 
> Should this be a real problem, open a new issue. Please include exact details concerning the transfer: is this elks-to-elks, or remote-to-elks; only using the new ftp client, or repeatable with macOS client, and which way the file transfer is going send or receive from ftpd, etc. In this way I can understand whether this is related to the FIN received with unprocessed data that was recently a problem (and now has debug printf turned off), or another problem.
> 
> If the problem only occurs on QEMU, we will also need to see whether this has anything to do with SO_REUSEADDR sockets or not. Most of that debug printf is still turned on.
> 
Will do - and like I said, this i QEMU only.
> the first cut is just around the corner - with some fun stuff for experimentation : select() multiplexing.
> 
> Both telnet and ktcp use select multiplexing for their network I/O, in case you haven't already noticed.
> 
> 
No, I wasn't aware of that. I was - for some reason - under the impression that select was not fully tested in networking. Have no idea where I got that. Great to know.

thank you!

—M

PR #1049, opened 2021-12-13
[kernel] Add support for Japanese NEC PC98 computer

ghaerr:
First changes for supporting PC-98, details are being discussed in #1047.

This first commit adds CONFIG_ARCH_PC98 and it's associated test `config.pc98`, which can be copied to .config and/or modified using `make menuconfig`, and used by everyone for test builds, using the 

A number of antiquated and unused CONFIG_PC_xxx and CONFIG_CPU_xxx defines were removed.

@tyama501, I plan on continuing to push more commits to this PR before committing. You're welcome to test by pulling a local copy, and commenting here whether something needs to change, for instance, config.pc98.


ghaerr:
Hello @tyama501,

Let me know if you have any questions, and also whether you'd like me to commit this first part to allow you more time to integrate, and if I should add the 1K sector and filesystem changes in different PR.

Thank you!

tyama501:
Hello @ghaerr ,

I merge your PR to my branch merge_pc98.
I am still using my config file but I could built it.

I got some small issue.
You can see "CR" code and some abnormal characters.
I think "CR" comes from some setup.S code change. 
I'm not sure why cpu name is abnormal.
ELKS98_merge_pc98

One more small thing in setup. May be 8018X is typo.
//+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
// Architecture specific routines for 8018X

Thanks.

ghaerr:
> I merge your PR to my branch merge_pc98.

Great, thank you!

> I am still using my config file but I could built it.

Fantastic to hear that it boots!

> You can see "CR" code and some abnormal characters.
I think "CR" comes from some setup.S code change.
I'm not sure why cpu name is abnormal.

The reason is that `call getcpu` is commented out for PC98. This is because the CPU recognition code is buried in the middle of CONFIG_ARCH_IBMPC stuff and I haven't reorganized it yet. So, the `*cp++ = setupb(i++)` is reading uninitialized memory. We might as well remove that string from your special code in mm/init.c for the time being. Afterwards when the CPU identification code is moved into a proper subroutine, you can add it back in if you like.

> One more small thing in setup. May be 8018X is typo.

Yes it is, I'll fix that.

I think I will commit this PR and open next one with disk management, since you have this merged and working.

Thank you!

ghaerr:
PC98 port contributed by @tyama501, committed by @ghaerr. Thank you @tyama501!

tyama501:
@ghaerr 

Thank you. I understood about CPU code.

For CR, I think because the following part is reverted. I forgot why I should do this before.
I will think how to solve later.

hello_mess:
#ifdef CONFIG_ARCH_PC98
	.ascii "ELKS Setup "
#else
	.ascii "\r\nELKS Setup "
#endif

ghaerr:
> hello_mess:
#ifdef CONFIG_ARCH_PC98
.ascii "ELKS Setup "
#else
.ascii "\r\nELKS Setup "
#endif

I looked at that, and thought there wasn't a need for "ELKS Setup" to be on a new line. Looking at the screenshot you provided, it seems it should be OK if ELKS Setup was on very top line. 

I think the reason "C/R" is displayed is that your 'put' code compares AX rather than just AL for output value. I will fix that in my next PR:
```
putc:
   push %ds
   push %bx
   push %di
   push %es
   push %dx
   push %cs
   pop %ds
   cmp $'\r',%ax <-- should be %al
   jz putesc_r
   cmp $'\n',%ax <-- same
```

tyama501:
Hello @ghaerr ,

Thank you for checking it.
I corrected that in early_puchar before but I didn't do that in setup.S.
And thank you for your additional PR.
I will look it later.

PR #1050, opened 2021-12-15
[kernel] Add disk support for PC98 computer

ghaerr:
Adds floppy disk and 1K sector support for PC98. Boot still not ported.

Contributed by @tyama501, thank you!

Hello @tyama501, this is the majority of the floppy disk routines with exception of boot code. I will push another commit allowing to define CONFIG_IMG_FD1232 to test the 1K sector code.

You should be able to get headless console working, after which you can submit your own PRs for that work.


tyama501:
Hello @ghaerr ,

I merged this PR to my merge_pc98 branch too.
I could build it but unfortunately I could not mount ROOT device.
ELKS98_merge_pc98_2

I will look again to find out what is happening when I have time.

ghaerr:
> I could build it but unfortunately I could not mount ROOT device.

I notice the boot message "bioshd: 0 floppy drives". This means no drives were recognized by BIOS bioshd.c.

I think perhaps this is because you may now require CONFIG_BLK_DEV_BFD_HARD=y to be set in your .config file (as well as your special autoconf.h). I didn't use exactly your code in bioshd.c, the new code depends on the above being set, as well as CONFIG_BLK_DEV_BDF).

ghaerr:
Also, @tyama501, now would be a good time to eliminate the hard-coded autoconf.h file, and use `cp pc98.config .config` for your config file. Save your old one of course, and perhaps you can compare them to see what is different, and submit that so I can see it.

Thank you!

tyama501:
Hello @ghaerr ,

Thank you for the comment.
Now I'm trying to use pc98.config.
I copied it to .config and saw from menuconfig.
I cannot find FD1232. Is that true for now?
Also the CONFIG_BLK_DEV_BFD_HARD is not set in pc98.config.

ghaerr:
> Now I'm trying to use pc98.config.
I copied it to .config and saw from menuconfig.
I cannot find FD1232. Is that true for now?

Oh, you will have to add CONFIG_IMG_FD1232 into pc98.config and .config for now, sorry! It should probably build the CONFIG_IMG_1440 version however. I will add a menuconfig option for CONFIG_IMG_FD1232 now.

> Also the CONFIG_BLK_DEV_BFD_HARD is not set in pc98.config.

Oops! That is also an error. Go ahead and add it into .config, I think that's the reason your system is not booting. I will also make that change in pc98.config now.

Thank you!

tyama501:
@ghaerr 

It booted after I manually set these.
Thank you!
ELKS98_merge_pc98_FD1232

tyama501:
Also nice to see 1232K message!

ghaerr:
> It booted after I manually set these.

Nice! Thank you for your quick testing. I will push the pc98.config changes and commit this.

> Also nice to see 1232K message!

Yes, I kind of snuck that in :)

I also notice that the system isn't reading /bootopts properly. Are you running the standard /bootopts, or something different?

tyama501:
>Are you running the standard /bootopts, or something different?

I don't think I am using something different.
Actually this is the first time to enable CONFIG_BOOTOPS.

It seems default setting makes DISK FULL for FAT1232.
Maybe we need to drop some user application.

ghaerr:
> Actually this is the first time to enable CONFIG_BOOTOPS.
It seems default setting makes DISK FULL for FAT1232.

Ok, that's the reason.

> Maybe we need to drop some user application.

I will fix that in the next PR!

PR #1054, opened 2021-12-17
[new] New ftp client, ftpd fixes

Mellvik:
### FTP client for ELKS

A (quite) compact interactive `ftp` client for  `elks` supporting a reasonable selection (!) of standard FTP commands and modes. 

`ftp` works pretty much like most ftp clients and thus does not need much introduction for 'old' users. For new users, any Linux/Unix ftp(1) man page will do, with a few exceptions, most of them temporary - such as:
- Unless anonymous login is acceptable, user name and password must be provided on the command line (-u / -p options)
- The server address must be present on the command line, the port # is optional.

A typical invocation will look like this
```
ftp -u username -p password server [port]
```
... and deliver output like this, varying somewhat from one server to the next:
```
# ./ftp -u name -p password 10.0.2.2    
220 raspi2b FTP server (Version 6.4/OpenBSD/Linux-ftpd-0.17) ready.             
230-                                                                            
230- The programs included with the Debian GNU/Linux system are free software;  
230- the exact distribution terms for each program are described in the         
230- individual files in /usr/share/doc/*/copyright.                            
230-                                                                            
230- Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent          
230- permitted by applicable law.                                               
230 User helge logged in.                                                       
                                                                                
Login OK.                                                                       
ftp>
```
`ftp` currently supports the following interactive commands - fully or partly (listing from the `help`command):

```
ftp> ?                                                                          
directory:      List remote directory                                           
ls:     List remote directory                                                   
get:    Fetch remote file                                                       
mget:   Fetch multiple remote files                                             
put:    Send file to remote                                                     
mput:   Send multiple files to remote                                           
cd:     Change (remote) working directory                                       
pwd:    Show (remote) working directory                                         
lcd:    Change local working directory                                          
bin:    Set file transfer type to BINARY                                        
ascii:  Set file transer type to ASCII                                          
quit:   Quit program                                                            
bye:    Leave program                                                           
help:   Print this.                                                             
?:      Alias for help.                                                         
!:      Invoke shell to execute local command.                                  
close:  Close connection.                                                       
user:   Open new connection, log in.                                            
open:   Alias for 'user'.                                                       
mkdir:  Create remote directory.                                                
delete: Delete remote file.                                                     
debug:  Set debug level 1-4 or off.                                             
verbose:        Toggle debug off or 1, default 1.                               
status: Show status, modes etc.                                                 
prompt: Toggle interactive multifile transfers    (prompt per file)                             
passive:        Toggle passive/active mode                                      
ftp> 
```
Many commands take parameters, such as `dir subdir` will list the contents of `subdir`.
Commands must be in lower case.

A few additional commands are planned for a subsequent version, notably TYPE, STATUS, OPEN, USER, CLOSE, possibly GLOB.

### Practical

Since the `type` command is missing, `ftp` starts in ascii- (text-)mode, use the `bin` command to set `image` (binary) mode. 
- `debug` can be set to any number, including 0 or `off`. Higher means more output. Currently 3 levels are uses: 0, 1 and >1. Use DEBUG OFF or DEBUG 0 to turn off debugging output. DEBUG 1  will by many be considered a 'normal' level of verbosity.
- `verbose` toggles the debug value between 0 and 1.
- Multiple arguments to file transfer commands is not yet supported. E.g. `get file1 file2` which would fetch file1 from the server and store it as file2 on a fully featured client, does not (yet) work (the second argument is silently ignored). `get file1` works as expected - and uses the same name at both ends.
- Globbing (wildcards in multifile transfers) is not supported, only `mget .` and `mput .` at this time. Use the `prompt` command to toggle between interactive (ask per feil) multifile transfers and batch mode (put/get everything).
- The default transfer mode is passive, use the PASSIVE command to toggle between  modes.
- Most commands can be shortened to 3 characters.
- There is no interrupt handling, ^C will abort the program.
- The stack setting in the `Makefile` is temporary, to be adjusted later.

`ftp` has been tested extensively against the standard Linux ftp server, more testing against other servers will be useful. It has also been tested against the ELKS ftp server (via loopback), with limited success - reasons to be determined.

Other changes in ### this PR:
- `ftpd.c`: 
    - Socket option SO_REUSEADDR has been reenabled
    - When overwriting files on receive, O_TRUNC was missing, with obvious consequences. This was the reason for the previously reported (suspected) inconsistency problem during receive.

### DIFF
--- a/elkscmd/inet/ftp/Makefile
+++ b/elkscmd/inet/ftp/Makefile
@@ -12,17 +12,21 @@ include $(BASEDIR)/Make.rules
 
 ###############################################################################
 
-SRCS = ftpd.c
+SRCS = ftpd.c ftp.c
+BINS = ftp ftpd
 OBJS = $(SRCS:.c=.o)
 
-all:	ftpd
+all:	$(BINS)
 
-ftpd:	$(OBJS)
-	$(LD) $(LDFLAGS) -o ftpd $(OBJS) $(LDLIBS)
+ftp:	ftp.o
+	$(LD) $(LDFLAGS) -maout-stack=8192 -o ftp ftp.o $(LDLIBS)
 
-install: ftpd
-	$(INSTALL) ftpd $(DESTDIR)/bin
+ftpd:	ftpd.o
+	$(LD) $(LDFLAGS) -maout-stack=8192 -o ftpd ftpd.o $(LDLIBS)
+
+install: ftpd ftp
+	$(INSTALL) $(BINS)  $(DESTDIR)/bin
 
 clean:
-	rm -f ftpd *.o
+	rm -f $(BINS) $(OBJS)
 
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -1,0 +1,1132 @@
+/*
+ *
+ * Minimal FTP client for ELKS 
+ * by Helge Skrivervik - helge@skrivervik.com - December 2021
+ *
+ *	TODO:
+ *	- move control channel setup to its own function
+ *	- split open/login into its own function, move them to the main command loop
+ *	  (remove the requirement to have user name and pw on the command line)
+ *	- cmd line server address argument must lookup in /etc/hosts
+ *	- add support for multi-argument commands (e.g. <put sourcefile destfile> ...)
+ *	- add missing commands (status, rmdir, TYPE, ...), 
+ *	- ... create common function for commands that are alomost alike
+ *	- clean up messags and debug levels
+ *	- fix trucation of file names in FAT, minix (warnings, name collisions)
+ *	- lcd with no params should print current local directory
+ *	- evaluate whether using select() is useful or just a complication
+ *	... add ABORT support in PUT
+ *	- handle server timeout (no activity): Unsolicited server input (code 421)
+ *	- add '15632 bytes sent in 0.00 secs (27.5561 MB/s)' type message after transfer
+ *	- more testing of error conditions & error messages
+ */
+
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <string.h>
+#include <arpa/inet.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <ctype.h>
+#include <stdlib.h>
+#include <netinet/in.h>
+#include <stdbool.h>
+#include <fcntl.h>
+#include <sys/select.h>
+#include <sys/time.h>
+#include <dirent.h>
+#include <errno.h>
+
+
+#define	IOBUFLEN 1500
+#define BUFSIZE	512
+#define ADDRBUF	50
+#define NLSTBUF 1500
+
+#define	TRUE	1
+#define	FALSE	0
+#define BIN	'I'
+#define ASCII	'A'
+#define PASV	0
+#define PORT	1
+
+#define UC(b) (((int) b) & 0xff)
+#define MIN(a,b) (((a)<(b))?(a):(b))
+#define MAX(a,b) (((a)>(b))?(a):(b))
+
+#ifndef MAXPATHLEN
+#define MAXPATHLEN 256
+#endif
+
+extern int errno;
+void settype(int controlfd, char t);
+int do_passive(int cmdfd);
+int do_active(int cmdfd);
+
+static FILE *fcmd;
+
+enum {
+	CMD_USER,
+	CMD_PASS,
+	CMD_CWD,
+	CMD_QUIT,
+	CMD_PORT,
+	CMD_PASV,
+	CMD_TYPE,
+	CMD_GET,
+	CMD_PUT,
+	CMD_BIN,
+	CMD_DELE,
+	CMD_RMD,
+	CMD_MKD,
+	CMD_PWD,
+	CMD_DIR,
+	CMD_NULL,
+	CMD_SYST,
+	CMD_STAT,
+	CMD_HELP,
+	CMD_ASCII,
+	CMD_PROMPT,
+	CMD_MGET,
+	CMD_LCWD,
+	CMD_MPUT,
+	CMD_DEBUG,
+	CMD_STATUS,
+	CMD_OPEN,
+	CMD_CLOSE,
+	CMD_VERB,
+	CMD_SHELL
+};
+
+
+struct cmd_tab {
+	char *cmd;
+	int value;
+	char *hlp;
+};
+struct cmd_tab cmdtab[] = {
+	{"directory", CMD_DIR, "List remote directory"},
+	{"ls", CMD_DIR, "List remote directory"},
+	{"get", CMD_GET, "Fetch remote file"},
+	{"mget", CMD_MGET, "Fetch multiple remote files"},
+	{"put", CMD_PUT, "Send file to remote"},
+	{"mput", CMD_MPUT, "Send multiple files to remote"},
+	{"cd", CMD_CWD, "Change (remote) working directory"},
+	{"pwd", CMD_PWD, "Show (remote) working directory"},
+	{"lcd", CMD_LCWD, "Change local working directory"},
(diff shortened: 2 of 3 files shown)


ghaerr:
Hello @Mellvik,

Wow, you have been busy, this is a fantastic contribution, thank you!! Thanks to you, we now have a working ftp client and server for ELKS, this is really great.

I'm wondering whether you might want to copy your extensive "documentation" above into a separate wiki page so that the instructions and features are readily accessible.

I'll find some time to play with it, and also please open an issue to describe problems between ELKS ftp client and server, if you find repeatable problems. I know we still need to fully solve the SO_REUSEADDR issue, but I'm thinking that's pretty safely contained within the ftp client and server here.

With regards to wildcard/globbing support, the `sash` source has a relatively small set of routines that implement that and is tested. Search for "ifdef WILDCARDS". That routine expands a pattern into a passed argc/argv array, would that output format work for what you're looking for? I could pull that out and make it a C library routine perhaps for easy usability.

This PR doesn't add ftp to Applications, would you like me to do that? I know we're running low on space on 1440K drives, but I'm sure I can find less important programs to omit should the multi-image build run out of space.

Thank you!!

Mellvik:
> Wow, you have been busy, this is a fantastic contribution, thank you!! Thanks to you, we now have a working ftp client and server for ELKS, this is really great.
> 
> I'm wondering whether you might want to copy your extensive "documentation" above into a separate wiki page so that the instructions and features are readily accessible.
> 
Yes, I'll do that - actually revise the file transfer wiki.
> I'll find some time to play with it, and also please open an issue to describe problems between ELKS ftp client and server, if you find repeatable problems.
> 
Will do. There are definitely timing problems that also show up ELKS-TO-ELKS. That will be a really interesting exercise.
> I know we still need to fully solve the SO_REUSEADDR issue, but I'm thinking that's pretty safely contained within the ftp client and server here.
> 
Yes, it's looking good.
> With regards to wildcard/globbing support, the sash source has a relatively small set of routines that implement that and is tested. Search for "ifdef WILDCARDS". That routine expands a pattern into a passed argc/argv array, would that output format work for what you're looking for? I could pull that out and make it a C library routine perhaps for easy usability.
> 
That's a great idea. Let me take a look - if we decide to use it, your suggestion to create a C library routine would be perfect.
> This PR doesn't add ftp to Applications, would you like me to do that?
> 
Please!
> I know we're running low on space on 1440K drives, but I'm sure I can find less important programs to omit should the multi-image build run out of space.
> 
:-) yes, I do have a couple of suggestions...
> 
There are a few issues hanging, such as the 'too many open files' problem. I'll poke into those now that the PR is out.

More fun to come.

Thank you!!!

--Mellvik

Mellvik:
@ghaerr, like you suggested I've taken a deep dive into the wildcard support routines in `sash`. 

Very interesting indeed - uncomplicated and smart with regards to memory handling. I support your idea of adding this to a library which will make it easier to use and include in other programs. I think it's important that the supporting routines (the memory chunking and the `match()` routine also become available by themselves. The sorting is not required.

And while we're at it, I suggest adding `getpass()` to a library too. While deprecated in general Linux, I suspect it will last for quite some time in ELKS (and I'm using it in `ftp`.

Wildcard expansion is up and running in `ftp` and works flawlessly. I do run out of memory now and then during testing - the size of directories (the number of files) is limited, but reasonably sized directories work fine.

I will be adding wildcard expansion to `ftpd` as well shortly. If you decide to go ahead with the library idea, I'll wait until its done.

BTW - the reason I'm holding off the `ftp` documentation update is that I'm rewriting the UI to be more complete and more like Linux and other `ftp`-clients. I'm also adding `BLOATED` defines to enable a small and limited version for tight environments.

--M

ghaerr:
> Wildcard expansion is up and running in ftp and works flawlessly. 

Can you post the (hopefully single) zipped .c file used to implement wildcard expansion? I'd like to see it for comparison with the original and can hopefully use it for the basis of a C library entry. Were there any/many modifications required for use in ftp?

> I suggest adding getpass() to a library too. 

Please post that as well, I assume you got it from sys_utils?

> the supporting routines (the memory chunking and the match() routine also become available by themselves.

What would these routines be useful for seperately? We have be careful with name collisions in libc.

> I do run out of memory now and then during testing

Increase the link to max heap (-maout-heap=0xffff) and see if that fixes things.

> The reason I'm holding off the ftp documentation update is that I'm rewriting the UI to be more complete and more like Linux and other ftp-clients

No problem, thank you!

Mellvik:
@ghaerr,

> 
> Wildcard expansion is up and running in ftp and works flawlessly.
> 
> Can you post the (hopefully single) zipped .c file used to implement wildcard expansion? I'd like to see it for comparison with the original and can hopefully use it for the basis of a C library entry. Were there any/many modifications required for use in ftp?
> 
Will do that directly in the web UI. There are almost no changes at all - less than 5 now that I've taken out the the debugging stuff I added to understand it.
> I suggest adding getpass() to a library too.
> 
> Please post that as well, I assume you got it from sys_utils?
> 
I'm loading the .o file from sys_utils.
> the supporting routines (the memory chunking and the match() routine also become available by themselves.
> 
> What would these routines be useful for seperately? We have be careful with name collisions in libc.
> 
Take a look at the match function — maybe you come to the same conclusion. As to the ememoty choking, I'm tempted to use it in the NLST routine in ftp in oder to avoid allocating a large stack buffer and then running out if space.
> I do run out of memory now and then during testing
> 
> Increase the link to max heap (-maout-heap=0xffff) and see if that fixes things.
> 
Will check that. We may be good anyway, directories with more than 200 long files aren't a good idea in ELKS.

—M

Mellvik:
[wildcards.zip](https://github.com/jbruchon/elks/files/7756321/wildcards.zip)

ghaerr:
> Will check that. We may be good anyway, directories with more than 200 long files aren't a good idea in ELKS.

There's no problem using max heap in ftp, since there's unlikely to be another running process that would use the memory anyways. Max heap is only an issue for programs running without user interaction. IMO, the client should not fail, even when it runs into large directories.

> As to the ememoty choking, I'm tempted to use it in the NLST routine in ftp in oder to avoid allocating a large stack buffer and then running out if space.

You don't really need chunking for that. I suggest using max heap and allocating from there, not the stack anyways. The analysis I use is to set the linker stack size to the largest requirement of the constant-coded buffers that are stack-allocated, and set the heap for the max dynamically-required (not constant) buffers that might be required from program execution. In ftp's case, setting max heap is probably the best idea given your directory size issue.

Mellvik:
>  I suggest using max heap and allocating from there, not the stack anyways. The analysis I use is to set the linker stack size to the largest requirement of the constant-coded buffers that are stack-allocated, and set the heap for the max dynamically-required (not constant) buffers that might be required from program execution. In ftp's case, setting max heap is probably the best idea given your directory size issue.
> 

Thank you @ghaerr, this is very useful advice. 

I believe there is potential for reduced stack allocations in both ftp and ftpd.

It's on my list.

—M

Mellvik:
One more thing on the heap/stack issue, @ghaerr:

I added -maout-heap=0xffff to ftp load, and got some interesting results:

The system() call (used inside ftp via the '!' command) just silently returns, and trying to run vi from a different (serial) terminal when ftp is running (109k memory available), fails ('no more memory'). BTW, -maout-stack is set to 8192, which may be too much. 

Any comments on this?

—M
 

> 22. des. 2021 kl. 12:04 skrev Helge Skrivervik ***@***.***>:
> 
>>  I suggest using max heap and allocating from there, not the stack anyways. The analysis I use is to set the linker stack size to the largest requirement of the constant-coded buffers that are stack-allocated, and set the heap for the max dynamically-required (not constant) buffers that might be required from program execution. In ftp's case, setting max heap is probably the best idea given your directory size issue.
>> 
> 
> Thank you @ghaerr, this is very useful advice. 
> 
> I believe there is potential for reduced stack allocations in both ftp and ftpd.
> 
> It's on my list.
> 
> —M
>

PR #1055, opened 2021-12-17
[boot] Add PC98 boot sector code

ghaerr:
This PR adds boot sector contributions by @tyama501 for the NEC PC-98 desktop computer.

Hello @tyama501!

I have updated the modifications trying to keep the boot sector code as readable as possible, while supporting both IBMPC and PC98 architectures. The biggest change is the requirement for 1K boot sector on PC98 when CONFIG_IMG_FD1232 is set. In this case, the ELKS BIOS Parameter Block (EBPB) is moved up 512 bytes to upper (last 512 bytes) portion of the boot sector (i.e., from 0x1F7 to 0x3F7). 

There are a number of boot parameters passed in the EBPB, such as boot hard drive partition start offset, and cylinder, head and track information for the floppy/hard drive when the EBPB signature is found. @tyama501's working boot set these in the first 512 bytes, rather than the last, so they weren't recognized during a floppy disk probe. The BIOS floppy driver also had to be slightly modified to properly read the changed EBPB location when using 1K sectors. I have not been able to test this, but if working, the message "fd: found valid ELKS disk parameters on /dev/fd0" should be displayed at boot.

In addition, the `setboot` command requires modification to indicate 1K sectors are being used, so the -K option was implemented. This should allow the EBPB to be set, and some previous code removed. @tyama501, can you test this with the couple below commented-out options in elks/images/Make.image?:
```
ifdef CONFIG_IMG_BOOT
ifdef CONFIG_IMG_FD1232
    #setboot $(TARGET_FILE) -K $(BPB) $(FD_FAT_BOOT)
    #setboot $(TARGET_FILE) -K -F $(BPB) $(FD_FAT_BOOT)
    setboot $(TARGET_FILE) $(FD_FAT_BOOT)
else
    setboot $(TARGET_FILE) -F $(BPB) $(FD_FAT_BOOT)
endif
```

On power-on boot, the IBM PC BIOS passes the boot drive number to the boot sector in DL. Does this occur on the PC-98? I may have messed this up; if this code doesn't boot, the SAVE_DRVNUM guards may be the issue.


tyama501:
Hello @ghaerr ,

Thank you for the updates!
I will try this.

> On power-on boot, the IBM PC BIOS passes the boot drive number to the boot sector in DL. Does this occur on the PC-98?

No. I don't think it does.
Instead you can get the boot drive number from 0000:0584.
The following part in boot_sect.S.

	mov $0x584,%bx
	mov %es:(%bx),%al    // Physical Device Address
...
	mov %al,drive_num

ghaerr:
> Instead you can get the boot drive number from 0000:0584.

It looks like it might be OK then. If that value is saved in drive_num, some code at the end of the boot, before passing control to the newly-loaded kernel, re-reads drive_num and saves this value in 'root_dev`. That, in turn, later allows the kernel to know to mount, say /dev/fd1 (ROOT_DEV = 0x381), rather than /dev/fd0, if the PC is booted from the second floppy.

tyama501:
I tried setboot $(TARGET_FILE) -K $(BPB) $(FD_FAT_BOOT)
This one cannot boot. I will try another one.

image

ghaerr:
> I tried setboot $(TARGET_FILE) -K $(BPB) $(FD_FAT_BOOT)
This one cannot boot.

Thanks. Also please try without changing any `setboot`; that is, leave both new statements commented out.

tyama501:
setboot /mnt/hdisk/git/elks_pc98/elks/image/fd1232.bin -K -F -B8,2,77 /mnt/hdisk/git/elks_pc98/elks/bootblocks/fat.bin
Usage: setboot <image> [-F] [-K] [-S{m,f}] [-{B,P}<sectors>,<heads>[,<tracks>]] [<input_boot_image>]

The other one seems cannot make image.

> Also please try without changing any setboot; that is, leave both new statements commented out.

Ok I will.

tyama501:
> Also please try without changing any setboot; that is, leave both new statements commented out.

This reseult was the same as the first one.

May I try FD1440 ?

tyama501:
FD1440 booted!

image

ghaerr:
> FD1440 booted!

Great! I think I see the big problem I created for FD1232. I got the 1K sector size stuff mixed up with the INITSEG offsets. I will push a fix.

> setboot /mnt/hdisk/git/elks_pc98/elks/image/fd1232.bin -K -F -B8,2,77 /mnt/hdisk/git/elks_pc98/elks/bootblocks/fat.bin
Usage: setboot  [-F] [-K] [-S{m,f}] [-{B,P},[,]] [<input_boot_image>]

Hmmm. Seems I didn't get something correct with the new -K option. I will fix that also.

I notice that /bootopts is still complaining in the boot screen. Is the disk still full? Can you "cat /bootopts" or "ls -l /bootopts" and see what it looks like?

tyama501:
ls -l /bootopts
image

I couldn't boot from second drive.
image

I have booted real PC from drive 1, 2, 3 before so there is a way :)

tyama501:
I will go sleep today. Thank you.

ghaerr:
I have pushed a fix for FD1232, hopefully it will boot with no changes to `setboot`, then you can also try the other two commented out lines instead.

> I couldn't boot from second drive.

I have also pushed a fix for that as well.

Thank you!

ghaerr:
Also, I've disabled /bootopts for PC98. I finally remembered that /bootopts is not (yet) supported for FAT boots, due to lack of space in the boot sector. So the error message was the result of uninitialized memory.

tyama501:
Hello @ghaerr ,

I merged the additional commits to my merge_pc98 branch but the result seems the same.
Cannot jmp to setup from the bootsector with Error 4.

image

Also FD1440 cannot boot from second drive.

Did I forget something to merge?

tyama501:
By the way, I have never tried MINIX file system before so my original boot code for MINIX has not been debugged.
MINIX file system was not supported.

My disk_read code for PC-98 only read 1 sector  so it maybe get trouble for FAST_READ.

ifndef BOOT_FAT
   define FAST_READ
else
   undef FAST_READ
endif

ghaerr:
> By the way, I have never tried MINIX file system before so my original boot code for MINIX has not been debugged.
MINIX file system was not supported.

Agreed. The build is only building a FAT disk, correct? CONFIG_IMG_FAT should be set.

> My disk_read code for PC-98 only read 1 sector so it maybe get trouble for FAST_READ.

I was wondering about that. FAST_READ is off for FAT though.

You are only testing FAT right now, correct? I definitely want to test FAT only now and MINIX can be debugged later.

1440K floppy still boots, correct? I am thinking we are only debugging issue with 1K sectors,...

ghaerr:
I have looked through every line of this ported boot code and can't see any difference that would affect the boot.

Can you make sure you are using the pc98.config file as .config? Ensure a full "make clean" (or make kclean for better speed), and that a FAT filesystem is produced. If CONFIG_IMG_FD1440 still boots (must change .config), then perhaps we could try using a working boot sector from your original port, and copy that into elkscmd/boot/fat.bin, then let the new setboot write it as the boot block. It is possible that setboot isn't working for the 1K boot sector case.

Otherwise I will have to see how to get Neko emulator running, I suppose.

tyama501:
@ghaerr 

Thank you for the reply.
I closed PC for today but yes I am using pc98.config that you have committed.

〉we could try using a working boot sector from your original port, and copy that into elkscmd/boot/fat.bin

I will try this later with new setboot.

Thank you.

ghaerr:
> we could try using a working boot sector from your original port, and copy that into elkscmd/boot/fat.bin

I think I have figured the problem: the fat.bin boot sector is 1K bytes long, and setboot is overwriting the first 512 bytes of the system image (which includes the "setup.S INITSEG") with it's last 512 bytes. That's where the "ELKS" signature is, which is complaining is not found. So I will look into fixing that. [EDIT: on 2nd thought, I am confused here: setboot writes the floppy image, not the system image, so this is likely not the problem.]

What I haven't figured out yet is how your original version worked, unless setboot was only writing 512 bytes, or your original fat.bin was not 1024 bytes long. 

Can you check it's size on your working system, and perhaps post the working version of fat.bin? Thank you!

ghaerr:
Hello @tyama501,

Ok, I have everything working. It seems the boot problem was related to perhaps a missing commit. In order to debug it, I ended up downloading [Dos-Box-X](https://dosbox-x.com), which can also emulate the PC-98 and BIOS. I ended up having to use your headless console code in order to see what was happening, so I've included that in the last commit, also having moved it to the CONIO API, which allows console-headless.c to be minimally modified.

On Dos-Box-X, both FD1232 and FD1440 boot and are working. 
Here's FD1232 booting using `dosbox-x-sdl -set machine=pc98`, after renaming image/fd1232.bin to pc98.fdd:
Screen Shot 2021-12-18 at 3 41 56 PM
Screen Shot 2021-12-18 at 2 15 00 PM
 
Take a look at the CONIO API in elks/arch/i86/drivers/char/conio-pc98-asm.S, that shows how the unmodified headless console sits on top of it. We may want to rename some of the "early_xxx" entry points to better conio_xxx names, I have left your original routines unmodified. It should be pretty straightforward to port the direct console driver to PC98, and that will allow you to remove any ESC processing in CONIO. We can talk about that more in #1047.

I have checked `setboot` and removed the extra versions from image/Makefile, no need to test. The new version only needs the additional '-K' option which specifies a 1K sector size. The bioshd.c disk probing now properly finds the ELKS BPB and uses that instead of probing, as can be seen from the screenshot above.

Please ensure you have pulled over all the commits into your PC-98 merge branch. If this boots, then I will commit this PR and we can start accepting your PRs as previously talked about, although I may create a few other cleanup PRs as well.

It is exciting actually seeing ELKS running on PC-98!

Thank you!

tyama501:
Hello @ghaerr ,

Thank you very much for these works!
Great! Wow, I didn't know DOSBox-X support PC-98 emulation mode.
Does it have second drive?

I could boot with the FD1232 image from the first drive.
It still have a trouble from booting from the second drive.

From the first drive
ELKS_1232boot_FD1

From the second drive
ELKS_1232boot_FD2

I will look the difference.
Thank you.

ghaerr:
Hello @tyama501,

Thank you for testing this so quickly!

Dos-Box-X does support multiple drives, but I can't yet figure out how to boot from drive B:. However, I was able to figure out why booting from 2nd drive didn't work. It should work now.

I am going to commit this now. Thank you for your help. This should be the last of my major commits for PC98, and I look forward to your future work and PRs, which we can discuss i #1047.

Thank you!

PR #1061, opened 2021-12-20
Sash

Mellvik:
Fixes missing ^C handling in `sash`. The fix makes no attempt to do reasonable recovery if ^C happens during processing of a command file.

Also fixes a couple of minor issues in command history.

### DIFF
--- a/elkscmd/sash/cmd_history.c
+++ b/elkscmd/sash/cmd_history.c
@@ -1,5 +1,5 @@
 /*
- * HS (@mellvik) for the ELKS project - march 2020 
+ * Helge Skrivervik (@mellvik) for the ELKS project - march 2020 
  * 
  * implements csh like command history for sash
  * 
@@ -8,6 +8,7 @@
 #include "sash.h"
 #ifdef CMD_HISTORY
 
+//#define DEBUG
 #define HISTMAX 200
 #define HISTMIN 20
 #define CMDBUF  100
@@ -16,10 +17,10 @@ static  int     lastcom = -1;   /* index of most recent command */
 static  int     histind = 0;    /* cmd # for history list */
 static  char    **histbuf;      /* array holding commands */
 int     histcnt = HISTMIN;
-void	phex(char *);
-void	phlist();
-
-/*#define DEBUG*/
+#ifdef DEBUG
+static	void	phex(char *);
+static	void	phlist(void);
+#endif
 
 void
 init_hist() {
@@ -89,12 +90,12 @@ cmd_get(int idx) { /* return the selected command from the history buffer */
 static char *
 cmd_search(char * pat) {
 	int idx, i;
-	char *found;
 
-	//printf("search %s, histcnt %d\n", pat, histcnt);
+	//printf("search '%s', histcnt %d\n", pat, histcnt);
 	for (i = 0; i < histcnt; i++) {
 		idx = map_ind(-i);
-		if ((found = strstr(histbuf[idx], pat)))
+		//if (strstr(histbuf[idx], pat) != NULL)	/* match 'pat' anywhere in histbuf entry */
+		if (!strncmp(histbuf[idx], pat, strlen(pat)))	/* match 'pat' from start of histbuf entry */
 			return(histbuf[idx]);
 	}
 	return(NULL);
@@ -166,8 +167,8 @@ add_to_history(char *cm) {
         }
         histbuf[lastcom] = p;
         histind++;
-        /*fprintf(stderr, " CMD: %s --- adr: %04x %04x %04x\n", cm, p, histbuf[lastcom]);
-        phlist();*/
+        //fprintf(stderr, " CMD: %s --- adr: %04x %04x %04x\n", cm, p, histbuf[lastcom]);
+        //phlist();
         return(0);
 }
 #ifdef DEBUG
@@ -236,7 +237,7 @@ history(char *cmd) {
         h = echo = 0;
         cm = cmd;       /* save pointer for reuse */
 
-        while (1) {	/* for an easy break */
+        while (1) {
 
         if (*cmd == '^') {
                 /* do substitution on the previous command */
--- a/elkscmd/sash/sash.c
+++ b/elkscmd/sash/sash.c
@@ -11,6 +11,7 @@
 
 #include <signal.h>
 #include <errno.h>
+#include <setjmp.h>
 
 typedef struct {
 	char	*name;
@@ -241,6 +242,7 @@ static	void	showprompt();
 static	BOOL	trybuiltin();
 
 BOOL	intflag;
+jmp_buf jb;
 
 int main(int argc, char **argv)
 {
@@ -348,6 +350,9 @@ readfile(name)
 #endif
 
 	while (TRUE) {
+		if (setjmp(jb)) {
+			printf("Got intr. \n");
+		}
 		fflush(stdout);
 #ifdef CMD_SOURCE
 		if (ttyflag)
@@ -894,6 +899,8 @@ catchint()
 
 	if (intcrlf)
 		write(STDOUT, "\n", 1);
+	printf("gotcha!");
+	longjmp(jb, 1);
 }
 
 


ghaerr:
Hello @Mellvik,

Thanks for the cleanup.

Regarding the SIGINT management, exactly what are you trying to do? I agree with your comment that sash "does nothing" while sitting at a prompt (other than display a newline). 

However, as neat as your proposed setjmp fix seems to be, it's not a great idea in the current form, as it introduces various memory leaks, which will surely kill the shell at some point. That is, for instance, any sub-command (such as 'ls'), that allocates memory will not free that memory if interrupted with ^C. We need more discussion before going this route, as all memory management will have to be rearranged using this approach.

Thank you!

Mellvik:
@ghaerr,

the sash ^C issue has been on my list seemingly forever. Hitting ^C at the prompt (or somewhere on the command line) leaves the shell apparently hanging, eventually returning to the prompt given either enough time or sufficient number of extra newlines.

The simple longjmp fixes that, and like the PR says, does not consider the effects on command files - which are very rare for sash anyway.

What effects it might have on subprocesses and memory management, if any, I don't know. The tests I've run indicate it's fine - if the numbers in the last line of meminfo can be used as a metric.

I'm open to suggestion as to how to make it better. To me, the small change makes sash useable again (there are more minor issues with history, but I suspect I'm the only one using it (history) at this time).

—Mellvik

ghaerr:
> Hitting ^C at the prompt (or somewhere on the command line) leaves the shell apparently hanging, eventually returning to the prompt given either enough time or sufficient number of extra newlines.

The operation of sash with ^C is extremely simple: it displays a newline during the SIGINT signal, then returns to whatever it was going. In the case of just sitting at a shell prompt, it displays a newline, not another prompt. Typing Enter just once will always display another prompt.

To fix this small issue, the solution is (almost) simple: just display a newline, then a prompt. Then things will look normal. Of course, if the shell is not at a prompt, then this will look funny, but probably no less funny than displaying a newline, realistically.

> The simple longjmp fixes that, and like the PR says, does not consider the effects on command files - which are very rare for sash anyway.

Sash doesn't support command files, other than through the 'source' command or a -c argument. When an executable "script" is identified (that is, 'x' bit set but no a.out header), sash hands off processing to whole invocation of the large /bin/sh. This doesn't occur when running stand alone (when sash is /bin/sh - then it gives an error message).

I'm not talking about command files though: if sash is running a long (internal) ls, or any other command that calls `malloc` (internally of course), then the proposed setjmp fix introduces an unacceptable memory leak. This occurs at least during history processing, dd, ls, ed, wildcard expansion and possibly other internal commands. All of these commands could be left in an improper state through introduction of the longjmp non-local goto.

> What effects it might have on subprocesses and memory management, if any, I don't know. The tests I've run indicate it's fine - if the numbers in the last line of meminfo can be used as a metric.

Sash executes all its internal commands without subprocesses. The numbers on the last line of meminfo are not meaningful here - we're talking about sash's internal heap. When it runs out, you'll get the "sys_brk fail" kernel error messages and the user will be left with an inoperational shell. The 'ps' command shows internal heap sizes.

> I'm open to suggestion as to how to make it better. To me, the small change makes sash useable again 

If all you're trying to fix is to show a prompt when at a prompt when ^C is typed, just change the ^C SIGINT routine catchint() to the following, which just displays another prompt. This solves this minor annoyance without adding memory leaks to the system shell:
```
   if (intcrlf) {
       write(STDOUT, "\n", 1);
       showprompt();
    }
```

Mellvik:
@ghaerr,

> Hitting ^C at the prompt (or somewhere on the command line) leaves the shell apparently hanging, eventually returning to the prompt given either enough time or sufficient number of extra newlines.
> 
> The operation of sash with ^C is extremely simple: it displays a newline during the SIGINT signal, then returns to whatever it was going. In the case of just sitting at a shell prompt, it displays a newline, not another prompt. Typing Enter just once will always display another prompt.
> 
No, if it was that simple, it would not be a problem. In QEMU it looks like that, on a real system it looks broken because it take forever to get a prompt back - for obvious reasons when you know what's going on, but very irritating still: You think you interrupt whatever you wrote on the command line, but still sash is trying to execute that line - being it blank or not.
> To fix this small issue, the solution is (almost) simple: just display a newline, then a prompt.
> 
Well - you have to do a little more than that, like flush the input buffer, but like you say, it's uncomplicated.
> Then things will look normal. Of course, if the shell is not at a prompt, then this will look funny, but probably no less funny than displaying a newline, realistically.
> 
Well, again realistically - this is a very different story on real hardware than in QEMU.
> 
> I'm not talking about command files though: if sash is running a long (internal) ls, or any other command that calls malloc (internally of course), then the proposed setjmp fix introduces an unacceptable memory leak. This occurs at least during history processing, dd, ls, ed, wildcard expansion and possibly other internal commands. All of these commands could be left in an improper state through introduction of the longjmp non-local goto.
> 
I misread this one. Your 'sub-command' became 'sub-process' to me, thus the response. I see the complications.

> I'm open to suggestion as to how to make it better. To me, the small change makes sash useable again
> 
> If all you're trying to fix is to show a prompt when at a prompt when ^C is typed, just change the ^C SIGINT routine catchint() to the following, which just displays another prompt. This solves this minor annoyance without adding memory leaks to the system shell:
> 
>    if (intcrlf) {
>        write(STDOUT, "\n", 1);
>        showprompt();
>     }

I'll do some experimentation with this, thank you.

—M

ghaerr:
@Mellvik,

Here's a couple articles on the possible severe problems of calling longjmp from a signal handler. I want you to understand exactly how signals work on ELKS (same as on Linux), and the repercussions of say, being in the middle of 'malloc' when the signal arrives, then the stack reset with longjmp, then the kernel return to the interrupted code after the signal handler returns. The asynchronous behavior of longjmp as well as signal handlers is ripe for corrupting data, with hard-to-reproduce consequences.

https://stackoverflow.com/questions/7334595/longjmp-out-of-signal-handler
https://stackoverflow.com/questions/17862982/stops-working-when-trying-to-longjmp-from-a-signal-handler
https://stackoverflow.com/questions/1715413/longjmp-from-signal-handler
https://pubs.opengroup.org/onlinepubs/9699919799/functions/longjmp.html
https://man7.org/linux/man-pages/man2/sigreturn.2.html

> In QEMU it looks like that, on a real system it looks broken because it take forever to get a prompt back - for obvious reasons when you know what's going on, but very irritating still: You think you interrupt whatever you wrote on the command line, but still sash is trying to execute that line - being it blank or not.

I don't understand what the reason is for sash talking "forever" to get a prompt back. What exactly is the case that isn't working well on real hardware? Are you talking about the fact that sash doesn't clear the command line when ^C is typed, and then when, in the middle of a command line, you type ^C, then Enter? That would then try to execute the previous command line, which takes a lot of time searching the floppy? I would like to get to the bottom of this real hardware problem.

With regards to my suggested fix: if you're used to ^C "clearing" the previously-input command buffer, then that would also have to be added, but we still need to be careful about SIGINT async processing as described in the above articles. If sash is sitting in 'fgets', I think it returns -EINTR, but it's possible it returns the number of characters previously received. That will all have to be dealt with, depending on exactly what you're trying to solve.

Note that none of this actually solves the bigger issue (in my mind): that of say, stopping a long 'ls -l' (internal command) listing. 

Thank you!

ghaerr:
> Well, again realistically - this is a very different story on real hardware than in QEMU.

On another way to emulate real hardware - the [DosBox-x](https://dosbox-x.com) project I found to emulate PC-98 recently runs **much** more realistically compared to real hardware, including the ability to see the "spinning wheel" on disk I/O. This could be very beneficial to use instead of QEMU to reproduce this problem you're talking about. Disk I/O is considerably slower - all by design. If you're interested, download the macOS version and this is the command line to directly boot an ELKS image:

`./dosbox-x-sdl2 -c fd1440.img"` (note .bin files need to be renamed to .img). I put this into a shell script for fast execution.

If you give me more details on the problem you're talking about, I'll try it on DosBox-x myself.

ghaerr:
Looking further into handling ^C and command buffers, it seems that the interrupt handler is already setting a global variable that could be used to fix this problem. That is, let the interrupt occur, and output nothing during the interrupt handler. Then in the main loop, after `fgets`, add the following:
```
        if (fgets(buf, CMDLEN - 1, fp) == NULL) {
            if (ferror(fp) && (errno == EINTR)) {
                clearerr(fp);
                continue;
            }
            break;
        }
        // new code, intcrlf can be removed from sash.c
        if (intflag) {
            buf[0] = '\n'; /* dummy up empty command*/
            buf[1] = '\0';
            write(1, buf, 1);
        }
```
This will clear any previously-entered command line, and output a newline before a new prompt is shown after testing strlen in the following statements.

Mellvik:
Thank you, @ghaerr - I very much appreciate the energy you're putting into this.
> Here's a couple articles on the possible severe problems of calling longjmp from a signal handler. I want you to understand exactly how signals work on ELKS (same as on Linux), and the repercussions of say, being in the middle of 'malloc' when the signal arrives, then the stack reset with longjmp, then the kernel return to the interrupted code after the signal handler returns. The asynchronous behavior of longjmp as well as signal handlers is ripe for corrupting data, with hard-to-reproduce consequences.
> 
> https://stackoverflow.com/questions/7334595/longjmp-out-of-signal-handler
> https://stackoverflow.com/questions/17862982/stops-working-when-trying-to-longjmp-from-a-signal-handler
> https://stackoverflow.com/questions/1715413/longjmp-from-signal-handler
> 
Thank you for the links. I'm well versed in using longjmp & co in programs, but really needed a refresh. 
> In QEMU it looks like that, on a real system it looks broken because it take forever to get a prompt back - for obvious reasons when you know what's going on, but very irritating still: You think you interrupt whatever you wrote on the command line, but still sash is trying to execute that line - being it blank or not.
> 
> I don't understand what the reason is for sash talking "forever" to get a prompt back. What exactly is the case that isn't working well on real hardware? Are you talking about the fact that sash doesn't clear the command line when ^C is typed, and then when, in the middle of a command line, you type ^C, then Enter? That would then try to execute the previous command line, which takes a lot of time searching the floppy? I would like to get to the bottom of this real hardware problem.
> 
@ghaerr, this is not a hardware problem. It's software not acting as expected. And it's perception. When hitting ^C you expect something. Not in a while, but now. As is, sash delivers nothing. A newline and nothing else is as close as you get to nothing. Hit ^C on the commandline in any shell you know about and you get a clean prompt, not a useless attempt to do something with whatever you tried to discard.
This as simple as you suggested when we started: a matter of getting the prompt back fast - which every other shell does fine. And no, it's not about visually clearing the command line. And yes, it't the floppy that takes forever. Even with the limited path in sash, it may be 10 seconds before the floppy search completes.
> With regards to my suggested fix: if you're used to ^C "clearing" the previously-input command buffer, then that would also have to be added, but we still need to be careful about SIGINT async processing as described in the above articles. If sash is sitting in 'fgets', I think it returns -EINTR, but it's possible it returns the number of characters previously received. That will all have to be dealt with, depending on exactly what you're trying to solve.
> 
Yes, getting the expected response when just hitting ^C at the prompt is important enough, but killing whaterver you wrote and start fresh is the real deal as long as we're just talking command line.
> Note that none of this actually solves the bigger issue (in my mind): that of say, stopping a long 'ls -l' (internal command) listing.
> 
Agreed, that is the bigger issue, and the REAL real deal.


-- M

Mellvik:
> Well, again realistically - this is a very different story on real hardware than in QEMU.
> 
> On another way to emulate real hardware - the DosBox-x project I found to emulate PC-98 recently runs much more realistically compared to real hardware, including the ability to see the "spinning wheel" on disk I/O. This could be very beneficial to use instead of QEMU to reproduce this problem you're talking about. Disk I/O is considerably slower - all by design. If you're interested, download the macOS version and this is the command line to directly boot an ELKS image:
> 
> ./dosbox-x-sdl2 -c fd1440.img" (note .bin files need to be renamed to .img). I put this into a shell script for fast execution.
> 
> If you give me more details on the problem you're talking about, I'll try it on DosBox-x myself.
> 
Admittedly, I don't use QEMU much these days. My new setup (and elks networking becoming very stable) makes the turnaround time for new binaries on real hardware about as fast as for QEMU. I just passed process id 500 and uptime > 48 hrs in heavy testing on a physical system. That said, dosbox is really interesting. I don't think it makes much difference for the sash issue, but it sounds like a valuable tool!

--M

PS: Maybe we should rename the image files to .img for compatibility. Would be good for Virtualbox too.
PS2: How about  a time counter in the kernel to work with the uptime command??

Mellvik:
> Looking further into handling ^C and command buffers, it seems that the interrupt handler is already setting a global variable that could be used to fix this problem. That is, let the interrupt occur, and output nothing during the interrupt handler. Then in the main loop, after fgets, add the following:
> 
>         if (fgets(buf, CMDLEN - 1, fp) == NULL) {
>             if (ferror(fp) && (errno == EINTR)) {
>                 clearerr(fp);
>                 continue;
>             }
>             break;
>         }
>         // new code
>         if (intflag && intcrlf) {
>             buf[0] = '\n'; /* dummy up empty command*/
>             buf[1] = '\0';
>             write(1, "\n", 1);
>         }
> This will clear any previously-entered command line, and output a newline before a new prompt is shown after testing strlen in the following statements.
> 
> 
> 
OK, I'll test this in the morning (away from my desk now). Off the top of my head I'm not sure why fgets would return NULL in this case, if there is data on the command line (or return at all since there is no newline). Will find out.

Thank you.

--M

ghaerr:
> Agreed, that is the bigger issue, and the REAL real deal.

I've looked at the sash source for ls, and this case is already handled correctly - by checking 'intflag' (set by the signal handler) and and discontinuing output.

> Off the top of my head I'm not sure why fgets would return NULL in this case, if there is data on the command line (or return at all since there is no newline).

It would return NULL because the read that fgets/fgetc/getchar is hanging on is interrupted by the signal, and   returns -EINTR. If there were no characters in the buffer after the read interruption, fgets will return NULL. If there were characters previously received, then read may return that number of characters (that is, fgets is normal, except a buffer before newline is entered). My last suggested fix (edited) should handle all these cases, and the signal handler only sets a global variable.

Mellvik:
@ghaerr,

fgets does not get interrupted by INTR, so the proposed change doesn't work.

Looking through the mallocs in sash, it seems a reasonable task to handle the longjmp - in ED, DD and the wildcard expander.

What do you think?

—M

ghaerr:
> Looking through the mallocs in sash, it seems a reasonable task to handle the longjmp - in ED, DD and the wildcard expander.
What do you think?

After reading the articles on using longjmp in a signal handler, definitely not. Also, rewriting all subcommands memory allocations just to fix a shell prompt issue on interrupt seems to be a bad idea.

> fgets does not get interrupted by INTR, so the proposed change doesn't work.

Hmmm, I'll look into this. The place to start is looking at the stdio code, so we can understand what is actually happening.

Mellvik:
> 
> Looking through the mallocs in sash, it seems a reasonable task to handle the longjmp - in ED, DD and the wildcard expander.
> What do you think?
> 
> After reading the articles on using longjmp in a signal handler, definitely not. Also, rewriting all subcommands memory allocations just to fix a shell prompt issue on interrupt seems to be a bad idea.
> 
Well, that's not the issue here, albeit one of them. Not having INTR in a shell limits its usability, in particular a shell with so many builtin functions.
BTW - ash is handling this just fine, if not simple.
> fgets does not get interrupted by INTR, so the proposed change doesn't work.
> 
> Hmmm, I'll look into this. The place to start is looking at the stdio code, so we can understand what is actually happening.
> 
> 
Thank you!

—M

ghaerr:
> Not having INTR in a shell limits its usability, in particular a shell with so many builtin functions.
> Well, that's not the issue here, albeit one of them.

What do you mean? Sash handles the ^C internally, all the commands use 'intflag' and stop what they're doing when an interrupt is received. The only issue is the shell prompt display on interrupt. 

What am I missing, what exactly are we trying to fix here?

ghaerr:
@Mellvik:

I think I've finally got to the bottom of this problem (that is, the problem of making sash operate like ash does at the shell prompt when ^C is typed):

`fgets` is ultimately based on `read`, and reading a tty doesn't always return an -EINTR when a signal is received. That is, normally, when an application is waiting on a read (file, socket, etc), it will return with the requested data, or, -EINTR if an interrupt was received which caused the read to return prematurely.

When reading a tty in ELKS, the read will return -EINTR *only* if there are no other characters buffered in the (cooked mode) tty driver (ash operates in raw mode which allows line editing, etc). When there have been other characters received (but the tty driver is still waiting on an Enter to be typed in cooked mode), the driver just returns those characters, without waiting for Enter, and returning the number of characters received, instead of -EINTR. Thus, the calling process never knows an interrupt was received, it just gets more characters (although without an Enter at the end of the buffer).

The stdio library routines thus also don't know an interrupt was received, and specifically, `fgets` keeps waiting in a loop, waiting for an Enter to be received before returning. I think this is an error in ELKS, since `sash` is coded to handle an interrupted return by checking `ferror` and `errno`:
```
        if (fgets(buf, CMDLEN - 1, fp) == NULL) {
            if (ferror(fp) && (errno == EINTR)) {
                clearerr(fp);
                continue;
            }
            break;
```
This code never gets executed in ELKS.

I'm not sure whether this problem should be fixed in the tty driver, or in the stdio library. I'm looking deeper into this to see what might be done.

ghaerr:
@Mellvik: I have a fix, which involves an updated `fread` routine that will go into `sash`.

After painstakingly figuring all that out, I then ran into yet another very strange problem with `sash`: if you boot ELKS and login, then type ^C, just to test ash (which works), *then* run `sash` - guess what? Sash doesn't even accept any input at all!! Complete hang. Try it!

Since this is turning into a far greater problem/mess than I had ever thought, why don't you go ahead and remove the setjmp/longjump changes to sash.c (that is, just checkout -f elkscmd/sash/sash.c), push, and then I'll commit the other changes to this PR. I have so far no idea what the latest problem is. I will then post a fix for sash that includes the original requested fix, as well as the fix to this latest find.

Thank you!

Mellvik:
@ghaerr,

I think it's useful to remember where this started. Santiago has been using sash quite a bit, and claimed ^C isn't working. I've been using sash quite a bit too, but left it behind when line editing came to ash, not because the line editing is better than history, but because INTR wasn't working. We are the users and we have the same experience. Whether 'not working' is technically correct or not, isn't important.

This change is attempting to fix that. And honestly I see no other way to do that than using signals and longjmp, which to me is the normal procedure for handling INTR signals. Of course it needs a lot more work than the simple jump back to the prompt, but it seems to me like a task of moderate complexity. Some of the FTP clients I've looked at recently have more elaborate signal/longjmp handling than (I think) is required here.

The problem with the intflag is that it works leisurely at best, which creates the perception of not working. Most users - myself included - expect something to happen right away when ^C is hit. In sash it does not.

The links we discussed were interesting (and refreshing regarding the considerations to be made when dealing with longjmps) but only partly relevant to ELKS. No threads for example.

So whether INTR is working or not depends on the metric. My metric says it doesn't and I'm inclined to do something about it. Your help in doing so would be appreciated.


—Mellvik

ghaerr:
> So whether INTR is working or not depends on the metric. My metric says it doesn't and I'm inclined to do something about it. Your help in doing so would be appreciated.

I am trying to fix it, and it is solved properly in #1072. From your comments, it seems you may not completely understand exactly what the problem is. I will try to explain.

> Santiago has been using sash quite a bit, and claimed ^C isn't working.

That's a different problem entirely - his serial console driver is not generating a SIGINT for some reason.

> And honestly I see no other way to do that than using signals and longjmp, which to me is the normal procedure for handling INTR signals.

It is not correct to fix this problem with a non-local goto, which generates all sorts of other problems, previously explained in above, but have to do with the problems, say, of getting an interrupt in the middle of a memory allocation, then longjmp-ing out of the routine, and leaving the process memory allocation in a corrupted state. That's one of the reasons why longjmp is highly discouraged from signal handling routines. In order to properly protect against this, interrupts would have to be disabled around all critical sections of library code.

> The problem with the intflag is that it works leisurely at best, which creates the perception of not working. 

I keep thinking you are confusing 'intflag' with the way the shell is supposed to act when a ^C is hit. These are two different things. The intflag is just a global variable that is always set properly when an SIGINT is received, and is already properly handled by sash for its other internal programs. Only the shell handling at the prompt is handled incorrectly.

> Most users - myself included - expect something to happen right away when ^C is hit. In sash it does not.

Of course. The ultimate problem is a bug in the ELKS stdio library. With that changed, as #1072 does, sash works as you expect, with only another single line change, to reset intflag.

> Whether 'not working' is technically correct or not, isn't important.

Disagree here.

> The links we discussed were interesting (and refreshing regarding the considerations to be made when dealing with longjmps) but only partly relevant to ELKS. No threads for example.

The real point is, that a signal is handled by the kernel by creating a new stack frame entirely, which specifically manages to cause an "async interrupt" directly into the running process. The return from the signal handler is very similar to an IRET (interrupt return) back to the previously-interrupted code. Thus the C application has the same problems as a reentrant kernel (also equivalent to reentrant thread issues), and the programmer must be aware of how this works. Since the normal action from a signal handler is to return from it, then none of this applies; however, using a non-local goto means that the procedure return, the IRET, etc never happens, and thus the application and library code is truly interrupted, with the consequences previously explained. I hope this helps!

Mellvik:
> @Mellvik: I have a fix, which involves an updated fread routine that will go into sash.
> 
> After painstakingly figuring all that out, I then ran into yet another very strange problem with sash: if you boot ELKS and login, then type ^C, just to test ash (which works), then run sash - guess what? Sash doesn't even accept any input at all!! Complete hang. Try it!
> 
How interesting! I'll do that!
> Since this is turning into a far greater problem/mess than I had ever thought, why don't you go ahead and remove the setjmp/longjump changes to sash.c (that is, just checkout -f elkscmd/sash/sash.c), push, and then I'll commit the other changes to this PR. I have so far no idea what the latest problem is. I will then post a fix for sash that includes the original requested fix, as well as the fix to this latest find.
> 
Thank you - first thing in the morning! Getting there ...

--M

PR #1063, opened 2021-12-21
[network] add 8 bit access mode for network devices.

cocus:
Useful for the ne1k-comaptible cards, or ISA 8 bit boards.


cocus:
Actually, this doesn't work. I don't get the CONFIG_ETH_BYTE_ACCESS, or the .S source file is not picking it up!
I need to add `#include <linuxmt/config.h>` on the top as well! woops

ghaerr:
> I need to add #include <linuxmt/config.h> on the top as well! woops

Ok, saw that. Does this work now?

cocus:
> Ok, saw that. Does this work now?

Yep!

PR #1067, opened 2021-12-21
[libc] Add wildcard expansion and getpass

ghaerr:
Requested in #1054.

Adds getpass (use #include <pwd.h>)
Adds expandwildcards/freewildcards (use <regex.h>)


Mellvik:
Thanks @ghaerr, that was indeed fast.

What's your thoughts on qsort and resources?

It doesn't do any good in the ftp case, but if it's not eating significant resources it's fine …

—Mellvik

ghaerr:
> What's your thoughts on qsort and resources?

Well, qsort doesn't actually use any extra memory at all, other than it's code - the sort occurs by swapping pointers within the already-allocated arrays.

And wildcard "globbing" is always sorted, so a C lib routine should definitely sort.

> It doesn't do any good in the ftp case, but if it's not eating significant resources it's fine …

I think we're ok. I'll commit this, and you can add wildcards in ftpd. We may need to be a bit careful with memory usage in ftpd, as this will increase it's heap requirements. We may need to think about how much extra memory should be allocated, as this space will always be reserved. What we really need is a way to specify heap size for a fork-exec'd program, so that the original ftpd has little heap, but allows for larger for the temp-running child.

 For ftp, max heap is no big issue, since it's the "top" application program running and all will be released on exit.

ghaerr:
> I'll commit this, and you can add wildcards in ftpd. We may need to be a bit careful with memory usage in ftpd, as this will increase it's heap requirements. We may need to think about how much extra memory should be allocated, as this space will always be reserved. 

@Mellvik: I'm thinking that perhaps, in order to keep memory usage down, and also to avoid crashing ftpd when expanding wildcards and running out of memory, we could specify a max memory to pass to expandwildcards(). If 0, there would be no limit, otherwise, pass a number like 2K or 4K (which would match the amount we'd permanently increase ftpd's linker heap and memory usage size). expandwildcards() would then either return success or fail, depending on whether it could expand the names within that amount of memory.

> What we really need is a way to specify heap size for a fork-exec'd program

I figured a way we could pass a new heap size to a newly exec'd program, but we can't do this when only forking, which is what the network daemons do.

What do you think, is this worth it? I remain concerned about permanent daemon memory usage. Unfortunately, there isn't an easy way to pass the heap size as arguments to a program, but `net start` could possibly do it, although that would require a shell extension and lots more work, and use the fork/exec heap method above.

A useful compromise would be to support globbing, but only for smaller expansions, say in the 2-4K range. That basically fits within the auto-allocated heap for existing ELKS applications.

Mellvik:
@ghaerr,
Please go ahead and commit this, I think we're fine.

Ftpd is not expected to have unlimited resources, and the ftp protocol has error messages to signal out of resources conditions. I'll make sure to include that when I make the changes. What we need is for the wildcarding code to be able to signal the out of resource condition back to the caller. I didn't look closely at the code you committed, but the code I've been using has a simple error printf about this. A minor mod to this and we're fine...

-Mellvik

Ps: sorry about the autocorrection screwup before. My spelling checker cannot fathom that anyone would use anything but norwegian to communicate.

PR #1069, opened 2021-12-21
[libc] Limit wildcard expansion to default 2048 bytes

ghaerr:
Requested in #1067.

> What we need is for the wildcarding code to be able to signal the out of resource condition back to the caller. 

@Mellvik,

`expandwildcards` will now limit the amount of memory allocated for filename expansion to 2K bytes. This should fit within the standard heap allocations of our current net daemons. Should this need to be expanded, wildcard.c can be edited, or we can develop an API change to set it. Of course, make sure to call `freewildcards` either after using the expanded filenames or before the next call to expandwildcards to free the allocated memory. This resets the filename expansion to 2K bytes again.

`expandwildcards` returns -1 on any error, including no matches, and now when the memory limit is exceeded (or there is no more heap). The current version also printf's error messages to stderr, which we might want to change to -ERRNO returns values other than -1, and have the application display them instead. Let me know what you think.

### DIFF
--- a/libc/misc/wildcard.c
+++ b/libc/misc/wildcard.c
@@ -12,6 +12,7 @@
 #include <unistd.h>
 #include <stdlib.h>
 #include <string.h>
+#include <errno.h>
 
 #define FALSE	0
 #define TRUE	1
@@ -165,8 +166,9 @@ namesort(char **p1, char **p2)
  * The expanded names are stored in memory chunks which can later all
  * be freed at once.  Returns zero if the name is not a wildcard, or
  * returns the count of matched files if the name is a wildcard and
- * there was at least one match, or returns -1 if either no filenames
- * matched or too many filenames matched (with an error output).
+ * there was at least one match, or returns < 0 if either no filenames
+ * matched (-ENOENT), too many filenames matched (-NOBUFS), no memory
+ * available (-ENOMEM), or bad regular expression (-EINVAL).
  */
 int
 expandwildcards(char *name, int maxargc, char **retargv)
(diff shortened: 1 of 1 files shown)


ghaerr:
@Mellvik,

I went ahead and removed the printfs from the `expandwildcards` library routine. It now returns standard (negative) errno values when expansion doesn't occur, according to the following. This will allow your clients/servers to do whatever they like directly.
```
/*
 * Expand the wildcards in a filename, if any.
 * Returns an argument list with matching filenames in sorted order.
 * The expanded names are stored in memory chunks which can later all
 * be freed at once.  Returns zero if the name is not a wildcard, or
 * returns the count of matched files if the name is a wildcard and
 * there was at least one match, or returns < 0 if either no filenames
 * matched (-ENOENT), too many filenames matched (-NOBUFS), no memory
 * available (-ENOMEM), or bad regular expression (-EINVAL).
 */
```

Mellvik:
Thank you very much, @ghaerr,
This is exactly what was needed (IMHO), I'll beat it up heavily over the next few days.

Do you think it makes sense to change sash to use the library version too? I certainly do.

--Mellvik

ghaerr:
> Do you think it makes sense to change sash to use the library version too?

No, because that will involve writing even more code in sash to re-display the error messages it already has written into its version.

> This is exactly what was needed (IMHO), I'll beat it up heavily over the next few days.

Ok. Is this code called in your new ftp client update you just posted?

Mellvik:
> This is exactly what was needed (IMHO), I'll beat it up heavily over the next few days.
> 
> Ok. Is this code called in your new ftp client update you just posted?
> 
> 
Yes it is. And it's being worked into the ftp server now, but I will not get that completed until early january I suspect.

Holiday season :-)

—M

PR #1070, opened 2021-12-22
[net] Buggy NE2K probe?

ghaerr:
@Mellvik,

I was trying to get DosBox-X networking working, which I finally did (internal only) but beforehand started looking at the NE2K probe code. I found this error, comparing AL after an INB instruction to the contents of location 0, rather than comparing it to the value of 0. Apparently this error has always worked.

However, I also notice a couple of lines above where you're setting page 0 and then reading it, you're sending 0x20 instead of 0x02, which is in the comments for "page 0" elsewhere in the driver.
```
ne2k_probe:

    // Poke then peek at the base address of the interface.
    // If something is there, return 0.
    // No attempt is made to get details about the i/f.

    mov net_port,%dx    // command register
    mov $0x20,%al   // set page 0 <--- should this be "mov $2,%al"? Setting page 0 uses "2" elsewhere.
    out %al,%dx
    in  %dx,%al
    cmp $0xff,%al   // cannot be FF
    jz  np_err
    cmp $0,%al      // cannot be 0  <-- fixed in this PR
```
If one looks through the driver for "page 0", there are places where 0x02 is used, and others where 0x20 is used. Sorry for my confusion, but just wanted to bring this up!

Thank you!

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -930,7 +930,7 @@ ne2k_probe:
 	in	%dx,%al
 	cmp	$0xff,%al	// cannot be FF
 	jz	np_err
-	cmp	0,%al		// cannot be 0
+	cmp	$0,%al		// cannot be 0
 	jz	np_err
 	
 	xor     %ax,%ax


Mellvik:
Hi @ghaerr -
Very interesting indeed. I need to refresh my memory on this one. It may very well be bugs - I forget if I even made any changed to this code, but it's likely that I did. 
Since this is the probe routine, there may be some trickery here not reflected by the comments.

I'll be back.

--M

Mellvik:
@ghaerr,
after some tracing, I'm definitely guilty of writing this. The original asm-driver had some commented out code in the probe routine and always returned zero. Presumably the code didn't work in QEMU.

Anyway, it's easy to get confused by the page changes in the ASM routines because we some times do more than just change the register page. `0x2` activates the NIC (START) in addition to selecting (or ensuring we're already on) page 0. `0x20` aborts whatever else may be going on at the time - in addition to setting page 0. Just to make sure the NIC is in a known state.
BTW - In the current version of the asm driver, there are no register page changes during normal operation, only in initialization and cleanup after errors.

The register page # is in the two most significant bits of the command byte.

The `cmp 0` is definitely a bug, thanks for fixing!

--M

ghaerr:
Hello @Mellvik,

Thanks for the information. 

> Anyway, it's easy to get confused by the page changes in the ASM routines because we some times do more than just change the register page.

In general, using hard-coded constants in assembly for communication with hardware devices is a bad idea. It would be nice to have the command bytes laid out as defines or equ's so that they are understandable and less prone to error (or memory!). Using 'start' or 'abort' etc would go a long way to understandability in the driver. I got immediately stuck, even after I found an error.

Thank you!

Mellvik:
@ghaerr,

> 
> Anyway, it's easy to get confused by the page changes in the ASM routines because we some times do more than just change the register page.
> 
> In general, using hard-coded constants in assembly for communication with hardware devices is a bad idea. It would be nice to have the command bytes laid out as defines or equ's so that they are understandable and less prone to error (or memory!). Using 'start' or 'abort' etc would go a long way to understandability in the driver. I got immediately stuck, even after I found an error.
> 
I couldn't agree more. Unsurprisingly, that was on the list when doing the work. And as unsurprisingly - it's still there. 

Fortunately I kept the notes, docs, so there's still hope. Actually - there is a wish-list for improvements to the driver that may take precedence over the 3Com driver in the (slow) works.

Thank you!

-M

PR #1071, opened 2021-12-22
[net, ftp] major update to the ftp client.

Mellvik:
A major update to the ELKS `ftp`-client - mostly about the user interface and supported commands. The login process is now resembling what most other `ftp`-clients do, with prompting for ip-address/name, username, password, and `open` + `close` commands enabling several sessions without leaving the program. 

Entering just `ftp` at the command line will bring up the prompt instead of complaining about a missing IP address/name. The command line can still be used as before, and the number of options has been expanded, again akin to other `ftp`-clients. More details about the user interface to be included in an updated ELKS FIle Transfer Wiki.

`ftp` now supports globbing (wildcards) - a major improvement (see #1067). Globbing can be removed at compile time, reducing the size of the binary file about 1.8k and reducing the memory footprint even more.

A compact version of `ftp` with limited functionality may be created by commenting out the BLOATED define in the source. This will bring the binary size down from 19k to 14.5k at the cost of active mode transfers (not much used these days), globbing and a handful of commands including `help` and `status`, remote delete, remote mkdir and a few others. The minimal client requires the server ip/name on the command line.

Be aware the globbing is implemented differently from one server to the next, the results from `mget file-with-wildcards` will vary.

Again, the 'test partner' has been Linux `ftpd`. Testing with other servers is needed.

There is still debug-stuff in the code, and more enhancements are planned, but those will be minor compared to this one.

### DIFF
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -4,12 +4,12 @@
  * by Helge Skrivervik - helge@skrivervik.com - December 2021
  *
  *	TODO:
- *	- move control channel setup to its own function
  *	- split open/login into its own function, move them to the main command loop
  *	  (remove the requirement to have user name and pw on the command line)
  *	- cmd line server address argument must lookup in /etc/hosts
  *	- add support for multi-argument commands (e.g. <put sourcefile destfile> ...)
- *	- add missing commands (status, rmdir, TYPE, ...), 
+ *	- add missing commands (rmdir, mls, nlist, ...), 
+ *	- Add port # to the open command
  *	- ... create common function for commands that are alomost alike
  *	- clean up messags and debug levels
  *	- fix trucation of file names in FAT, minix (warnings, name collisions)
@@ -20,6 +20,11 @@
  *	- add '15632 bytes sent in 0.00 secs (27.5561 MB/s)' type message after transfer
  *	- more testing of error conditions & error messages
  */
+
+#define BLOATED		/* fully featured if defined */
+#ifdef  BLOATED
+#define GLOB		/* include wildcard processing */
+#endif
 
 #include <sys/types.h>
 #include <sys/socket.h>
@@ -37,12 +42,17 @@
 #include <sys/time.h>
 #include <dirent.h>
 #include <errno.h>
+#ifdef GLOB
+#include <pwd.h>
+#include <regex.h>
+#endif
 
 
 #define	IOBUFLEN 1500
-#define BUFSIZE	512
+#define BUF_SIZE 512
 #define ADDRBUF	50
 #define NLSTBUF 1500
+#define MAXARGS	256	/* max names in wildcard expansion */
 
 #define	TRUE	1
 #define	FALSE	0
@@ -60,51 +70,56 @@
 #endif
 
 extern int errno;
-void settype(int controlfd, char t);
-int do_passive(int cmdfd);
-int do_active(int cmdfd);
+extern char *getpass(char *prompt);
+static void settype(int controlfd, char t);
+static int do_passive(int cmdfd);
+static int do_active(int cmdfd);
 
 static FILE *fcmd;
 
-enum {
+enum {	// commands in disconnected mode
+	CMD_OPEN,	// Must be first!!
 	CMD_USER,
-	CMD_PASS,
+	CMD_NOCONNECT,	//flag
+	// commands in connected mode
 	CMD_CWD,
-	CMD_QUIT,
-	CMD_PORT,
-	CMD_PASV,
-	CMD_TYPE,
 	CMD_GET,
 	CMD_PUT,
-	CMD_BIN,
-	CMD_DELE,
 	CMD_RMD,
 	CMD_MKD,
 	CMD_PWD,
 	CMD_DIR,
+	CMD_MPUT,
+	CMD_DELE,
+	CMD_MGET,
+	CMD_CLOSE,
+	CMD_CONNECT,	// flag
+	// commands in any mode
+	CMD_QUIT,
+	CMD_PORT,
+	CMD_PASV,
+	CMD_TYPE,
+	CMD_BIN,
 	CMD_NULL,
 	CMD_SYST,
 	CMD_STAT,
 	CMD_HELP,
 	CMD_ASCII,
 	CMD_PROMPT,
-	CMD_MGET,
 	CMD_LCWD,
-	CMD_MPUT,
 	CMD_DEBUG,
 	CMD_STATUS,
-	CMD_OPEN,
-	CMD_CLOSE,
 	CMD_VERB,
+	CMD_GLOB,
 	CMD_SHELL
 };
-
 
 struct cmd_tab {
 	char *cmd;
 	int value;
 	char *hlp;
 };
+
 struct cmd_tab cmdtab[] = {
 	{"directory", CMD_DIR, "List remote directory"},
 	{"ls", CMD_DIR, "List remote directory"},
@@ -114,29 +129,33 @@ struct cmd_tab cmdtab[] = {
 	{"mput", CMD_MPUT, "Send multiple files to remote"},
 	{"cd", CMD_CWD, "Change (remote) working directory"},
 	{"pwd", CMD_PWD, "Show (remote) working directory"},
-	{"lcd", CMD_LCWD, "Change local working directory"},
 	{"bin", CMD_BIN, "Set file transfer type to BINARY"},
 	{"ascii", CMD_ASCII, "Set file transer type to ASCII"},
 	{"quit", CMD_QUIT, "Quit program"},
+	{"type", CMD_TYPE, "Show file transfer type."},
+	{"prompt", CMD_PROMPT, "Toggle interactive multifile transfers"},
+	{"debug", CMD_DEBUG, "Set debug level 1-4 or off."},
 	{"bye", CMD_QUIT, "Leave program"},
-	{"help", CMD_HELP, "Print this."},
-	{"?", CMD_HELP, "Alias for help."},
+#ifdef BLOATED
 	{"!", CMD_SHELL, "Invoke shell to execute local command."},
+	{"?", CMD_HELP, "Alias for help."},
+	{"help", CMD_HELP, "Print this."},
 	{"close", CMD_CLOSE, "Close connection."},
-	{"user", CMD_USER, "Open new connection, log in."},
-	{"open", CMD_OPEN, "Alias for 'user'."},
+	{"open", CMD_OPEN, "Open new connection, log in."},
 	{"mkdir", CMD_MKD, "Create remote directory."},
 	{"delete", CMD_DELE, "Delete remote file."},
-	{"debug", CMD_DEBUG, "Set debug level 1-4 or off."},
 	{"verbose", CMD_VERB, "Toggle debug off or 1, default 1."},
 	{"status", CMD_STATUS, "Show status, modes etc."},
-	{"prompt", CMD_PROMPT, "Toggle interactive multifile transfers"},
+	{"lcd", CMD_LCWD, "Change local working directory"},
+	{"glob", CMD_GLOB, "Toggle globbing"},
 	{"passive", CMD_PASV, "Toggle passive/active mode"},
+#endif
+	//{"user", CMD_USER, "Open new connection, log in."},
 	{"", CMD_NULL, ""}
 };
 static int debug = 0;
 static char type = ASCII, atype = ASCII;
-static int prompt = TRUE;
+static int prompt = TRUE, glob = TRUE;
 
 /* Trim leading and trailing whitespaces */
 void trim(char *str) {
@@ -155,6 +174,7 @@ void trim(char *str) {
 	str[i - begin] = '\0'; // Null terminate string.
 }
 
+#ifdef BLOATED
 void help() {
 	struct cmd_tab *c = cmdtab;	
 
@@ -164,6 +184,7 @@ void help() {
 	}
 	return;
 }
+#endif
 
 /* Interactive mget/mput */
 
@@ -245,7 +266,6 @@ int get_port_string(char *str, char *ip, int p0, int p1) {
 	for (i = 0; i < strlen(ip); i++) {
 		if (itmp[i] == '.') itmp[i] = ',';
 	}
-
 	sprintf(str, "PORT %s,%d,%d\r\n", itmp, p0, p1);
 	return 1;
 }
@@ -283,6 +303,7 @@ int get_ip_port(int fd, char *ip, int *port) {
 	return 1;
 }
 
+//FIXME: Add multi-parameter support, return in a **argv type array
 int get_parm(char *input, char *fileptr) {
 	char cpy[MAXPATHLEN];
 	char *parm;
@@ -323,13 +344,15 @@ int dataconn(int fd) {		/* wait for the actual data connection in active mode */
 
 int do_nlst(int controlfd, char *buf, int len, char *dir, int mode) {
 	int n, i, datafd;
-	char lbuf[BUFSIZE];
+	char lbuf[BUF_SIZE];
 
 	if (atype != ASCII) settype(controlfd, ASCII);
-	if (mode == PASV)
-		datafd = do_passive(controlfd);
-	else
+#ifdef BLOATED
+	if (mode != PASV)
 		datafd = do_active(controlfd);
+	else
+#endif
+		datafd = do_passive(controlfd);
 	if (datafd <= 0) {
 		printf("NLST: Failed to open data connection.\n");
 		return -1;
@@ -337,8 +360,6 @@ int do_nlst(int controlfd, char *buf, int len, char *dir, int mode) {
 
 	sprintf(lbuf, "NLST %s\r\n", dir);
 	send_cmd(controlfd, lbuf);
-	//write(controlfd, lbuf, strlen(lbuf));
-	//if (debug) printf("---> %s", lbuf);
 
 	if (mode == PORT) {	/* active mode, wait for connection */
 		if ((n = dataconn(datafd)) < 0) {
@@ -439,7 +460,7 @@ int do_get(int controlfd, char *file1, char *file2, int mode) {
 	int maxfdp1, data_finished = FALSE, control_finished = FALSE;
 	fd_set rdset;
 
-	bzero(iobuf, sizeof(iobuf));
+	//bzero(iobuf, sizeof(iobuf));
 	if (*file2 == '\0') file2 = file1;
 
 	if ((fd = open(file2, O_WRONLY|O_TRUNC|O_CREAT, 0664)) < 0) {
@@ -447,10 +468,12 @@ int do_get(int controlfd, char *file1, char *file2, int mode) {
 		return -1;
 	}
 	if (atype != type) settype(controlfd, type);
-	if (mode == PASV)
-		datafd = do_passive(controlfd);
-	else
+#ifdef BLOATED
+	if (mode != PASV)
 		datafd = do_active(controlfd);
+	else
+#endif
+		datafd = do_passive(controlfd);
 	if (datafd < 0) return -1;
 
 	sprintf(iobuf, "RETR %s\r\n", file1);
@@ -476,11 +499,11 @@ int do_get(int controlfd, char *file1, char *file2, int mode) {
 	FD_ZERO(&rdset);
 
 	maxfdp1 = MAX(controlfd, datafd) + 1;
-	printf("Remote: %s, local: %s\n", file1, file2);
+	printf("remote: %s, local: %s\n", file1, file2);
 	/*
 	 * NOTICE: When fetching very small or NULL size files, the '150 Opening'
 	 * message and the 226 Transfer Complete message arrive almost at the same time.
-	 * So when get_reply fetches the former, the latter ends up in the input buffer,
+	 * So when get_reply() fetches the former, the latter ends up in the input buffer,
 	 * and the loop below will not receive anything on the control channel.
 	 */
 	struct timeval tv; int select_return;
@@ -492,9 +515,9 @@ int do_get(int controlfd, char *file1, char *file2, int mode) {
 		select_return = select(maxfdp1, &rdset, NULL, NULL, &tv);
 /// Experimental
 		if (!select_return) { 
-			printf("get: select timeout.\n");
+			if (debug > 2) printf("get: select timeout.\n");
 			if (icount++ > 2 || data_finished == TRUE) {	
-				//Experimental: Got timeout, need reply - it's probalby sitting in the
+				//Experimental: Got timeout, need reply - which is probalby sitting in the
 				// fgets() input buffer.
 				// If it isn't, we're stuck here.
 				get_reply(controlfd, iobuf, IOBUFLEN, 1);
@@ -502,7 +525,6 @@ int do_get(int controlfd, char *file1, char *file2, int mode) {
 			}
 		}
 
-		//bzero(iobuf, (int)sizeof(iobuf));
 		if (FD_ISSET(controlfd, &rdset)) {
 			get_reply(controlfd, iobuf, IOBUFLEN, 1);
 			//printf("get select cntrl\n");
@@ -545,10 +567,12 @@ int do_put(int controlfd, char *file1, char *file2, int mode){
 
 	if (atype != type) settype(controlfd, type);
 	if (*file2 == '\0') file2 = file1;
-	if (mode == PASV)
-		datafd = do_passive(controlfd);
-	else
+#ifdef BLOATED
+	if (mode != PASV)
 		datafd = do_active(controlfd);
+	else
+#endif
+		datafd = do_passive(controlfd);
 	if (datafd < 0) return -1;
 
 	if ((fd = open(file1, O_RDONLY)) < 0) {	/* FIXME: we've checked this already ... */
@@ -558,7 +582,7 @@ int do_put(int controlfd, char *file1, char *file2, int mode){
 
 	sprintf(iobuf, "STOR %s\r\n", file2);
 	send_cmd(controlfd, iobuf);
-	printf("Local: %s, remote: %s\n", file1, file2);
+	printf("local: %s, remote: %s\n", file1, file2);
 
 	get_reply(controlfd, iobuf, IOBUFLEN, 1);
 	
@@ -639,6 +663,7 @@ put_out:
 	return status;
 }
 
+#ifdef BLOATED
 /* 
  * Active mode - PORT command: 
  * Open a port for the server to connect to 
@@ -647,7 +672,7 @@ int do_active(int cmdfd) {
 	struct sockaddr_in myaddr;
 	int fd, i = 0;
 	unsigned int myport;
-	char str[BUFSIZE], *p, *myip;
+	char str[BUF_SIZE], *p, *myip;
 
 	if ((fd = socket(AF_INET, SOCK_STREAM, 0)) < 0) {
 		perror("socket error");
@@ -695,17 +720,16 @@ int do_active(int cmdfd) {
 	}
 
 	get_port_string(str, myip, (int) UC(p[0]), (int) UC(p[1]));
-	//write(cmdfd, str, strlen(str));
-	//if (debug) printf("---> %s", str);
 	send_cmd(cmdfd, str);
 
-	if (get_reply(cmdfd, str, BUFSIZE, 1) < 0) {	// 200 Port command successful ...
+	if (get_reply(cmdfd, str, BUF_SIZE, 1) < 0) {	// 200 Port command successful ...
 		/* command channel error, probably server timeout */
 		close(fd);
 		fd = -1;
 	}
 	return fd;
 }
+#endif /*BLOATED*/
 
 /* PASSIVE MODE:
  * Connect to service point announced by server */
@@ -714,12 +738,11 @@ int do_passive(int cmdfd) {
 	struct sockaddr_in srvaddr;
 	int fd = 0;
 	unsigned int port;
-	char command[BUFSIZE], ip_addr[ADDRBUF];
+	char command[BUF_SIZE], ip_addr[ADDRBUF];
 	//FILE *cmd;
 
 	send_cmd(cmdfd, "PASV\r\n");
-	//write(cmdfd, "PASV\r\n", 6);
-	get_reply(cmdfd, command, BUFSIZE, 1);
+	get_reply(cmdfd, command, BUF_SIZE, 1);
 	if (strncmp(command, "227", 3)) {
 		printf("Server error, cannot connect.\n");
 		return -1;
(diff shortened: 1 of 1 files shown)


Mellvik:
Examples:
```
# ./ftp                                                                         
ftp> open 10.0.2.2                                                              
Connected to 10.0.2.2.                                                          
220 raspi2b FTP server (Version 6.4/OpenBSD/Linux-ftpd-0.17) ready.             
Name (10.0.2.2:ftp): helge                                                      
Password:                                                                       
230-                                                                            
230- The programs included with the Debian GNU/Linux system are free software;  
230- the exact distribution terms for each program are described in the         
230- individual files in /usr/share/doc/*/copyright.                            
230-                                                                            
230- Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent          
230- permitted by applicable law.                                               
230 User helge logged in.                                                       
Remote system is Unix/Linux, using binary mode                                  
ftp> ?                                                                          
directory:      List remote directory                                           
ls:     List remote directory                                                   
get:    Fetch remote file                                                       
mget:   Fetch multiple remote files                                             
put:    Send file to remote                                                     
mput:   Send multiple files to remote                                           
cd:     Change (remote) working directory                                       
pwd:    Show (remote) working directory                                         
bin:    Set file transfer type to BINARY                                        
ascii:  Set file transer type to ASCII                                          
quit:   Quit program                                                            
type:   Show file transfer type.                                                
prompt: Toggle interactive multifile transfers                                  
debug:  Set debug level 1-4 or off.                                             
bye:    Leave program                                                           
!:      Invoke shell to execute local command.                                  
?:      Alias for help.                                                         
help:   Print this.                                                             
close:  Close connection.                                                       
open:   Open new connection, log in.                                            
mkdir:  Create remote directory.                                                
delete: Delete remote file.                                                     
verbose:        Toggle debug off or 1, default 1.                               
status: Show status, modes etc.                                                 
lcd:    Change local working directory                                          
glob:   Toggle globbing                                                         
passive:        Toggle passive/active mode                                      
ftp> 
```
Using the command line:
```
# ./ftp -u helge 10.0.2.2                                                       
Connected to 10.0.2.2.                                                          
220 raspi2b FTP server (Version 6.4/OpenBSD/Linux-ftpd-0.17) ready.             
Name (10.0.2.2:helge):                                                          
Password:                                                                       
230-                                                                            
230- The programs included with the Debian GNU/Linux system are free software;  
230- the exact distribution terms for each program are described in the         
230- individual files in /usr/share/doc/*/copyright.                            
230-                                                                            
230- Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent          
230- permitted by applicable law.                                               
230 User helge logged in.                                                       
Remote system is Unix/Linux, using binary mode                                  
ftp> 
```

ghaerr:
Hello @Mellvik,

Wow, you have been extremely busy, again!!

Thank you for these enhancements and the suggestions around expandwildcards and quick integration. I'll try to find time to test this all out soon :)

ghaerr:
Hello @Mellvik,

Using the latest code, I have a question about how to use FTP within ELKS. After booting ELKS with networking running, running "ftp localhost 8021" works (from outside ELKS), I get connected and asked for Username/Password, then execute "dir" command, etc.

When connecting from within ELKS using the new ftp client, I type "ftp localhost", and, instead of connecting, it returns with "ftp>". Upon entering "open localhost" the ftp client hangs. Are only IP numbers supported? It also seems to operate differently if I start with "ftp 10.0.2.15".

If I type "open 10.0.2.15" at the prompt, it connects. It seems that /etc/hosts (or the automatic conversion of localhost) is not supported, although ftp seems to operate a bit differently with a hostname on the command line - it is ignored.

I'm doing all this on QEMU. This is with the /boottops "ftpd=-q" option turned OFF. Unfortunately, this means that ftpd won't work for external connections, only internal without that startup option. We need to rethink that, as now one needs to contemplate how ftpd will be used before starting it in QEMU. Perhaps ELKS ftp needs to also work with the ftpd -q kluge... Thoughts?

Thank you!

Mellvik:
Congratulations, @ghaerr - you found the first bug :-)

Actually an oversight on my part - I forgot to put the gethostbyname call back in, and used IP numbers all the way. 127.0.0.1 works fine too btw.


> 24. des. 2021 kl. 15:30 skrev Gregory Haerr ***@***.***>:
> 
> 
> 
> Using the latest code, I have a question about how to use FTP within ELKS. After booting ELKS with networking running, running "ftp localhost 8021" works, I get connected and asked for Username/Password, then execute "dir" command, etc.
> 
> When connecting from within ELKS using the new ftp client, I type "ftp localhost", and, instead of connecting, it returns with "ftp>". Upon entering "open localhost" the ftp client hangs. Are only IP numbers supported? It also seems to operate differently if I start with "ftp 10.0.2.15".
> 
> If I type "open 10.0.2.15" at the prompt, it connects. It seems that /etc/hosts (or the automatic conversion of localhost) is not supported, although ftp seems to operate a bit differently with a hostname on the command line - it is ignored.
> 
> I'm doing all this on QEMU. This is with the /boottops "ftpd=-q" option turned OFF. Unfortunately, this means that ftpd won't work for external connections, only internal without that startup option. We need to rethink that, as now one needs to contemplate how ftpd will be used before starting it in QEMU. Thoughts?
> 
> 
Good point - I thing we have to do this explicitly, test the command line before options processing.  I have no better idea right now. Which means that unless the localhost address is specified on the command line it won't work.

Thanks for the ftpd update. Admittedly I had that one running along with some other adjustments, but I'm a lot slower — as usual.

Thank you!

-M

ghaerr:
> I thing we have to do this explicitly, test the command line before options processing.  I have no better idea right now. Which means that unless the localhost address is specified on the command line it won't work.

I'm not following you here, sorry. Are you saying we would support a -q option in ftp, that works for connecting locally to a ftpd running -q?

> I forgot to put the gethostbyname call back in, and used IP numbers all the way.

Yes, "ftp localhost", "ftp elks", or "open..." the same should all work. I'll leave that to you, but I'll go ahead and commit the ftpd zombie fix.

On another note, I've got some ideas on how to fix the problem of the infinite hang when connecting (from either telnet or ftp) to a non-existent address. We'll leave that for later for now, but as you have probably seen, there's no way to abort out of it. I'm thinking of putting a timer in ELKS `connect` that returns failure after a few seconds.

Mellvik:
@ghaerr,
> 
> I thing we have to do this explicitly, test the command line before options processing. I have no better idea right now. Which means that unless the localhost address is specified on the command line it won't work.
> 
> I'm not following you here, sorry. Are you saying we would support a -q option in ftp, that works for connecting locally to a ftpd running -q?
> 
I was contemplating a single line change in the server, checking the address of the client and disabling the QEMU hack if the address is localhost. But I wasn't thinking clearly. Use active mode, and no change is necessary, it works out of the box even if -q is active.
> I forgot to put the gethostbyname call back in, and used IP numbers all the way.
> 
> Yes, "ftp localhost", "ftp elks", or "open..." the same should all work. I'll leave that to you, but go ahead and commit the ftpd zombie fix.
> 
> 
Great!
> On another note, I've got some ideas on how to fix the problem of the infinite hang when connecting (from either telnet or ftp) to a non-existent address. We'll leave that for later for now, but as you have probably seen, there's no way to abort out of it. I'm thinking of putting a timer in ELKS connect that returns failure after a few seconds.
> 
That's a great idea!! It would be very helpful.


Thank you.

-M

ghaerr:
> I was contemplating a single line change in the server, checking the address of the client and disabling the QEMU hack if the address is localhost.

Good potential idea, but ktcp changes localhost to the actual IP address so that wouldn't work. But keep up any ideas for getting rid of -q!! Another idea would be to somehow have ftpd automatically know its running on QEMU. Not sure how to do that, even if we wanted to.

> But I wasn't thinking clearly. Use active mode, and no change is necessary, it works out of the box even if -q is active.

Good to know, I didn't think about active mode. So, ftpd should always be run -q in QEMU, and the user will always use active mode from within ELKS, and always use passive mode from outside.

>  "ftp localhost", "ftp elks", or "open..." the same should all work. I'll leave that to you

When ELKS ftp operates more closely to say, macOS ftp, that will help, as one won't have to operate differently whether inside or outside ELKS.

Mellvik:
> I was contemplating a single line change in the server, checking the address of the client and disabling the QEMU hack if the address is localhost.
> 
> Good potential idea, but ktcp changes localhost to the actual IP address so that wouldn't work.
> 
Good point, and it was a bad idea to begin with. BTW - why does ktcp do that? If it didn't, passive mode would work fine locally even if -q is used.
> But keep up any ideas for getting rid of -q!! Another idea would be to somehow have ftpd automatically know its running on QEMU. Not sure how to do that, even if we wanted to.
> 
That would indeed be great. Maybe a timer can do the trick: If the CPU is insanely fast, it's QEMU (or some other emu)...
> But I wasn't thinking clearly. Use active mode, and no change is necessary, it works out of the box even if -q is active.
> 
> Good to know, I didn't think about active mode. So, ftpd should always be run -q in QEMU, and the user will always use active mode from within ELKS, and always use passive mode from outside.
> 
Yes, that's right. 
> "ftp localhost", "ftp elks", or "open..." the same should all work. I'll leave that to you
> 
> When ELKS ftp operates more closely to say, macOS ftp, that will help, as one won't have to operate differently whether inside or outside ELKS.
> 
That's the intention. We're close. PR coming for the gethostbyname fix.

--M

Mellvik:
>> But keep up any ideas for getting rid of -q!! Another idea would be to somehow have ftpd automatically know its running on QEMU. Not sure how to do that, even if we wanted to.
>> 
> That would indeed be great. Maybe a timer can do the trick: If the CPU is insanely fast, it's QEMU (or some other emu)…

@ghaerr,
Come to think of it, trying to get rid of -q is most likely a bad idea. It’s a hack - as is the QEMU network solution we’re using. Most emulators, including QEMU IIRC - can do full interface emulation if the host allows it, and the issue is moot.

—M

ghaerr:
> It’s a hack - as is the QEMU network solution we’re using. Most emulators, including QEMU IIRC - can do full interface emulation if the host allows it, and the issue is moot.

My understanding was that QEMU requires certain ports to be translated, so as to not interfere with the host system. These include 8080->80, 2323->23, and 8020->20 and 8021->21. Is it that we have to pick some range for the passive mode port (e.g. 8041->49821) for the same reason, that is, to get it to pass through to QEMU?

Obviously, if we don't have to do this, it would be great to figure out how, now!

I have been playing with DosBox-X and have internal networking running, but haven't gotten networking working between it (ELKS) and the host yet. Ideas welcome for other emulators.

> Come to think of it, trying to get rid of -q is most likely a bad idea. 

The entire purpose of '-q' is to translate a range of port address (passive mode only) as described above, correct?
This is definitely a kluge. I would like to at least eliminate the -q command line option, and use a new option (QEMU=1), set in /bootopts, which cause ftpd to do the same thing. This eliminates lots of special "net start" code looking for the kluge "ftpd=-q" environment variable, which only works at boot, and replaces it with the QEMU=1 environment variable that lasts the entire uptime, and also works after subsequent "net stop; net start" which currently doesn't. This mechanism (also now used by TZ=) also allows the boot option (environment variable) to be specified in .profile, which allows this to work on FAT-booted volumes, where /bootopts isn't implemented (such as the current PC-98 port).

So I'll probably go ahead and remove -q and use an internal ftpd mechanism such as "getenv("QEMU")" to set the internal flag until we figure out how to get rid of special casing entirely. In the meantime, users will have to remember to use active mode within ELKS and passive mode from without...

> BTW - why does ktcp do that? 

ktcp translates the address early on in order for loopback to occur internally. This also considerably simplifies things, otherwise the entire stack would have to compare IP multiple addresses all the time, rather than just using the actual IP address.

> If it didn't, passive mode would work fine locally even if -q is used.

I don't understand how passive mode could work just by not translating localhost? I thought the entire active/passive mode kluge problem had to do with whether or not a set of ports had to be translated, not an IP address.

Perhaps ftp should also be modified to translate ports if getenv("QEMU") is set? This would at least make ftp and ftpd work identically on QEMU.

ghaerr:
@Mellvik,

Another thought I just came up with for this QEMU port translation hassle would be for ktcp to do it, silently without any application kluges. In this scenario, we could, say, set a certain range of addresses (e.g. 8041-8049, but better perhaps in the 4000-4095 range), that when a non-localhost binds or connects to that address, it is internally translated appropriately instead before going on the wire. Local packets would not undergo translation.

This approach could possibly "just work" internally to ktcp and stop us from writing kluges to applications that would otherwise require it. It's almost like a kind of NAT translation within ktcp, but only for known QEMU-required port translations. 

What do you think of that?

Mellvik:
> ﻿
@ghaerr,
> It’s a hack - as is the QEMU network solution we’re using. Most emulators, including QEMU IIRC - can do full interface emulation if the host allows it, and the issue is moot.
> 
> My understanding was that QEMU requires certain ports to be translated, so as to not interfere with the host system. These include 8080->80, 2323->23, and 8020->20 and 8021->21. Is it that we have to pick some range for the passive mode port (e.g. 8041->49821) for the same reason, that is, to get it to pass through to QEMU?
> 
That's right. And the reason is that QEMU is using the host's loopback interface. The data channel ports don't really need to be translated, we could use the same numbers (49xxx) on bith sides, but they would still have to be listed in the qemu startupscript.
> Obviously, if we don't have to do this, it would be great to figure out how, now!
> 
LIke I've mentioned a few times before, if qemu had its own (logical) interface, all this would go away. And memory tells me (haven't verified this) that this is possible albeit complicated with qemu, if the host supports tun/tap, which is a mechanism to create logical (tunnel) interfaces. IIRC VirtualBox does this and works well with ELKS. Which reminds me over another issue I mentioned a few daysback: Can we rename the images from .bin to .img? That would make us more compatible with the world and reduce hassle when using other emulators.
> I have been playing with DosBox-X and have internal networking running, but haven't gotten networking working between it (ELKS) and the host yet. Ideas welcome for other emulators.
> 
I haven't played with DosBox yet, I need to do that. I would be surprised if tun/tap is supported though, it's complicated.
> Come to think of it, trying to get rid of -q is most likely a bad idea.
> 
> The entire purpose of '-q' is to translate a range of port address (passive mode only) as described above, correct?
> 
No, well, depends on perpective. In passive mode, the server tells the client - on demand - which ip address and port number to open for the data channel. This cannot possibly work for qemu on a shared loopback interface, since all ports need to be specified in advance. So what -q does is to disable normal passive mode operation and fake the ip address and port numbers to use (in particular, make the port numbers limited and predetermined). So instead of telling the client to use the real thing -10.0.2.15/49041, we tell it to use 127.0.0.1/8041, which qemu maps to the former - because we told it so in the config. This cannot be done anywhere else than in ftpd BTW. 
Now, since the forced ip address is 127.0.0.1, this faking would be invisible to local loopback traffic if ktcp did't internally change the loopback to the real address. Thus my question about that.
> This is definitely a kluge. I would like to at least eliminate the -q command line option, and use a new option (QEMU=1), set in /bootopts, which cause ftpd to do the same thing. This eliminates lots of special "net start" code looking for the kluge "ftpd=-q" environment variable, which only works at boot, and replaces it with the QEMU=1 environment variable that lasts the entire uptime, and also works after subsequent "net stop; net start" which currently doesn't. This mechanism (also now used by TZ=) also allows the boot option (environment variable) to be specified in .profile, which allows this to work on FAT-booted volumes, where /bootopts isn't implemented (such as the current PC-98 port).
> 
Sounds good to me. Whatever we can to to make things simpler. And easier to maintain.
> So I'll probably go ahead and remove -q and use an internal ftpd mechanism such as "getenv("QEMU")" to set the internal flag until we figure out how to get rid of special casing entirely. In the meantime, users will have to remember to use active mode within ELKS and passive mode from without...
> 
SUpported, but can you hold on for a day or two so I can get some ftpd fixes pushed?
> BTW - why does ktcp do that?
> 
> ktcp translates the address early on in order for loopback to occur internally. This also considerably simplifies things, otherwise the entire stack would have to compare IP multiple addresses all the time, rather than just using the actual IP address.
> 
OK, it's still possible to hack this in ftpd if we like, by testing for the request coming from its own address. One more line of code I guess.
> If it didn't, passive mode would work fine locally even if -q is used.
> 
> I don't understand how passive mode could work just by not translating localhost? I thought the entire active/passive mode kluge problem had to do with whether or not a set of ports had to be translated, not an IP address.
> 
I think this one is explained above.

So the bottom line - I think - is that the qemu support we have is about as simple as it gets. We just have to use active mode internally for loopback connections. BTW - there is something odd (buggy) going on when using the new ftp/ftpd via loopback on qemu, I'm looking into that right now.

-M

Mellvik:
> @Mellvik,
> 
> Another thought I just came up with for this QEMU port translation hassle would be for ktcp to do it, silently without any application kluges. In this scenario, we could, say, set a certain range of addresses (e.g. 8041-8049, but better perhaps in the 4000-4095 range), that when a non-localhost binds or connects to that address, it is internally translated appropriately instead before going on the wire. Local packets would not undergo translation.
> 
> This approach could possibly "just work" internally to ktcp and stop us from writing kluges to applications that would otherwise require it. It's almost like a kind of NAT translation within ktcp, but only for known QEMU-required port translations.
> 
> What do you think of that?
> 
An interesting idea, @ghaerr - but I don't see how it would help since we would still need to keep the -q hack in ftpd. And the list in the QEMU config file, because every port we're going use for incoming traffic needs to be in the config file. 
I honestly think what we have is a simple as it gets until we find a way to have qemu use its own interface - or change emulator.

--Mellvik

Mellvik:
@ghaerr -

Apropos QEMU and tun/tap -
I have't tested this yet, but here's the setup I'd like to check out asap:

THis one is not entirely up to date but might be OK:
https://www.dzombak.com/files/qemu-bridging-mavericks.pdf
Installing tun/tap on Macos:
https://macappstore.org/tuntap/

There may be something in this one too - I'm going to check this out tomorrow.
https://superuser.com/questions/1662371/set-up-tap-networking-for-qemu-on-macos-host

-M

ghaerr:
I agree with most of what you're saying, except there's a few things that aren't quite the way you think they are. The biggest is that although we specify 127.0.0.1 in QEMU's startup script, that's only necessary for the macOS side. By the time ktcp gets the packet, it has already been translated to 10.0.2.15, so your comments about localhost translation won't work. That is, QEMU translates the localhost port numbers to the guest OS with a changed IP and changed port. I figure it uses a prior translation from 127.0.0.1 -> 10.0.2.15 to do so. Either that or QEMU is running as 10.0.2.2, so it always translates anything for 10.0.2.2 to 10.0.2.15, since I think that's also QEMU's default "host" IP.

> The data channel ports don't really need to be translated, we could use the same numbers (49xxx) on bith sides, but they would still have to be listed in the qemu startupscript.

If we can do this, why do we have the 8041->49821 translations in QEMU now? I'm confused by that. I realize that ftpd is hacked to send 8041 instead of 49821. But why do that if you can just send a range of high port numbers in and out untranslated?

> Can we rename the images from .bin to .img? 

Yes, I ran into the same problem with DosBox-X. I was wondering whether other users would care, but I think it would help compatibility. There's a bunch of documentation that say .bin though.

> can you hold on for a day or two so I can get some ftpd fixes pushed?

Yes, I'm testing NAT now, it will work. The only change would be that ftpd would always specify 8041-8049 for the port, rather than the high 49821 value. Ktcp would translate the port in and out automatically. This would only be required while the QEMU port hack in the startup script remains, then NAT could be turned off in ktcp.

The idea in adding NAT to ktcp is to keep applications from knowing about emulator-specific networking issues, close the to reasons NAT is used in routers, to keep the computer on it from having to know about all the mess of additional translations necessary.

> So instead of telling the client to use the real thing -10.0.2.15/49041, we tell it to use 127.0.0.1/8041, which qemu maps to the former - because we told it so in the config. This cannot be done anywhere else than in ftpd BTW.

Yes - it can also be done in ktcp, since ktcp never sees 127.0.0.1 from outside QEMU. QEMU translates the IP address prior as discussed above. And ktcp immediately translates 127.0.0.1 as well. We can determine local traffic easily in ktcp by comparing source address == destination address.

> Now, since the forced ip address is 127.0.0.1, this faking would be invisible to local loopback traffic if ktcp did't internally change the loopback to the real address. Thus my question about that.

Actually its the reverse: we just compare source address == destination address (== 10.0.2.15 normally) and that's when loopback occurs in ktcp. In ktcp, any non-local IP is always 10.0.2.2 (== gateway), and ktcp keeps that address as QEMU then rewrites that on the way out of its host interface.

> Apropos QEMU and tun/tap

Let me know if you get that working. I found early versions of tun/tap had kernel problems in my much earlier versions of macOS. BTW, DosBox-X does not support tun/tap, only slirp, which QEMU uses and I could not get working, and pcap.
I still think we're going to need possible NAT in ktcp in order to work with all emulators, and not force hard setups on various emulators, which we can barely get to work well now.

> there is something odd (buggy) going on when using the new ftp/ftpd via loopback on qemu, I'm looking into that right now.

Thanks - yes, there are some definite bugs with ftp/ftpd now, they occur immediately and a connect followed by dir doesn't always work well. I'll stay away from any ftp/ftpd source changes for now, but am looking hard at ktcp NAT translation as a possible additional fix to keep this crud from infesting our networking applications, which could be part of the problem we're now seeing.

ghaerr:
Hello @Mellvik,

> So instead of telling the client to use the real thing -10.0.2.15/49041, we tell it to use 127.0.0.1/8041, which qemu maps to the former - because we told it so in the config. This cannot be done anywhere else than in ftpd BTW.

My apologies - you are correct. I didn't realize that the passive mode actually sends the IP address over as well, which of course is required to be 127.0.0.1, to allow the packet to be translated by QEMU's startup script. Uggh. This means my NAT idea probably won't work, unless there's a way to cause a macOS 10.0.2.15 packet to be routed to QEMU.

Thank you for your previous explanations!

ghaerr:
@Mellvik,

I think I am coming to understand what you mean when saying that QEMU runs using the loopback interface. I had thought all macOS packets were seen on the loopback interface, which would allow for fancier routing of non-localhost packets to QEMU and its guest OS. What is happening is that the loopback interface only sees localhost/127.0.0.1 packets, is that correct? So the only packets that can be routed/intercepted by QEMU are those addressed to IP 127.0.0.1.

What about the possibility of a seperate gateway/interface on macOS, one that accepts all/any packets from/to 10.0.2.2? Is there a way to do that? We're using that now with the `slattach.sh` script that was written for testing ELKS PPP by creating a serial line interface and associating it with IP 10.0.2.2 (taken from slattach.sh, Linux only):
```
/sbin/ifconfig sl0 up $gateway pointtopoint $elks netmask 255.255.255.0 mtu 1024
```

Mellvik:
> I agree with most of what you're saying, except there's a few things that aren't quite the way you think they are. The biggest is that although we specify 127.0.0.1 in QEMU's startup script, that's only necessary for the macOS side. By the time ktcp gets the packet, it has already been translated to 10.0.2.15, so your comments about localhost translation won't work. That is, QEMU translates the localhost port numbers to the guest OS with a changed IP and changed port.
Yes, that's exactly what I'm saying. Please read again.
> 
> The data channel ports don't really need to be translated, we could use the same numbers (49xxx) on bith sides, but they would still have to be listed in the qemu startupscript.
> 
> If we can do this, why do we have the 8041->49821 translations in QEMU now? I'm confused by that. I realize that ftpd is hacked to send 8041 instead of 49821. But why do that if you can just send a range of high port numbers in and out untranslated?
> 
If the same number makes you feel better, go ahead. Like I pointed out, it makes no difference. It was a choice of the moment. When I said it doesn't need to be translated, what I meant is they may be the same, it doesn't make any difference technically. They're always translated - if the number is the same on both sides, fine. If it's translated to the same number, no problem - except it's  easy to forget that it's translated and make the wrong decisions based on that.
> can you hold on for a day or two so I can get some ftpd fixes pushed?
> 
> Yes, I'm testing NAT now, it will work. The only change would be that ftpd would always specify 8041-8049 for the port, rather than the high 49821 value. Ktcp would translate the port in and out automatically. This would only be required while the QEMU port hack in the startup script remains, then NAT could be turned off in ktcp.
> 
> The idea in adding NAT to ktcp is to keep applications from knowing about emulator-specific networking issues, close the to reasons NAT is used in routers, to keep the computer on it from having to know about all the mess of additional translations necessary.
> 
 I haven't understood what the benefit is. BTW real NAT is complicated, withg big tables, timeouts and more. For sure it cannot help the qemu situation. In what setting would this be useful?
> So instead of telling the client to use the real thing -10.0.2.15/49041, we tell it to use 127.0.0.1/8041, which qemu maps to the former - because we told it so in the config. This cannot be done anywhere else than in ftpd BTW.
> 
> Yes - it can also be done in ktcp, since ktcp never sees 127.0.0.1 from outside QEMU. QEMU translates the IP address prior as discussed above.
> 
Why would we do this?
Sorry, it seems like total miscommuncation. We're talking about entirely different things. ANY level of NAT will not help the fptd hack.
> And ktcp immediately translates 127.0.0.1 as well. We can determine local traffic easily in ktcp by comparing source address == destination address.
> 
> Now, since the forced ip address is 127.0.0.1, this faking would be invisible to local loopback traffic if ktcp did't internally change the loopback to the real address. Thus my question about that.
> 
> Actually its the reverse: we just compare source address == destination address (== 10.0.2.15 normally) and that's when loopback occurs in ktcp. In ktcp, any non-local IP is always 10.0.2.2 (== gateway), and ktcp keeps that address as QEMU then rewrites that on the way out of its host interface.
> 
Hmmm, my explanation must have been totally off. I'll try again tomorrow. This just makes sense to me. Like I said above, we seem to be talking about entirely different things.

Further, what about forgetting about the whole thing. Keep the -q option  till we have a tun/tap setup that works and then delete?

-M

Mellvik:
> 
> @Mellvik,
> 
> I think I am coming to understand what you mean when saying that QEMU runs using the loopback interface. I had thought all macOS packets were seen on the loopback interface, which would allow for fancier routing of non-localhost packets to QEMU and its guest OS. What is happening is that the loopback interface only sees localhost/127.0.0.1 packets, is that correct? So the only packets that can be routed/intercepted by QEMU are those addressed to IP 127.0.0.1.
> 
> What about the possibility of a seperate gateway/interface on macOS, one that accepts all/any packets from/to 10.0.2.2? Is there a way to do that? We're using that now with the slattach.sh script that was written for testing ELKS PPP by creating a serial line interface and associating it with IP 10.0.2.2 (taken from slattach.sh, Linux only):
> 
> /sbin/ifconfig sl0 up $gateway pointtopoint $elks netmask 255.255.255.0 mtu 1024

Well, this is getting very close to tun/tap - which would be good to get 'under control' - works on windows too, although with different setup. Will look more at that tomorrow.

@ghaerr, have you tried Virtualbox?


-M

ghaerr:
> Like I said above, we seem to be talking about entirely different things.

Yes, I'm sorry. The last bit of conversation I have both misunderstood how FTP passive mode actually works as well as try to explain an internal ktcp issue, and I have been both incorrect about NAT as well as muddled the conversation. I now understand what you are talking about.

> Further, what about forgetting about the whole thing. Keep the -q option  till we have a tun/tap setup that works and then delete?

Ok, no problem. Now that I actually understand FTP and what you have been saying, it's handled well the way you've done it. I would like to use QEMU=1 environment variable instead of ftpd=-q in /bootopts, so that FAT boots work with ftpd on QEMU.  I can see you've put in tons of work on both these and would rather go with your direction, since you're way ahead of me on this! I'll stay out of FTP modifications and just try to integrate its usage as best we can with /bootopts, etc. 

If you could add a change to ftpd.c that defaults the "qemu" variable to the value of environment variable QEMU=1 (or 0) that would be great. I can then change /boottopts and /bin/net so that we don't have to have the -q hack in there, and it will be self-contained and automatic in ftpd, at least. That should allow both internal and external ftp to work correctly in passive mode, correct?

> have you tried Virtualbox?

Not in a while. I'll look into it and see how it works.

Thank you!

Mellvik:
>  I would like to use QEMU=1 environment variable instead of ftpd=-q in /bootopts, so that FAT boots work with ftpd on QEMU.
> 
I support that - it's simpler and more flexible.
> If you could add a change to ftpd.c that defaults the "qemu" variable to the value of environment variable QEMU=1 (or 0) that would be great.
> 
Certainly - I'll include it with some other fixes pending.
> I can then change /boottopts and /bin/net so that we don't have to have the -q hack in there, and it will be self-contained and automatic in ftpd, at least. That should allow both internal and external ftp to work correctly in passive mode to it, correct?
> 
Well, internally (loopback mode), only active mode will work if the qemu hack is active. I would think that's good enough - ftp via loopback is a very special case anyway. It's also a setting I haven't tested much - yet.
> have you tried Virtualbox?
> 
> Not in a while. I'll look into it and see how it works.
> 
> 
Likewise - I'll pull it out from the digital drawer and see if we can get 'real' networking going.

Thank you.

-M

Mellvik:
Quick update on emulators, @ghaerr -

- VB has no NE2k support
- Parallels may have it, but only the PCI variant. I don't know if that's workable.
- DosBox has no Ethernet except possibly via DOS packet drivers.


Looks like we're stuck with QEMU - so let's see if we can get tap/tun to work.


—M

> 27. des. 2021 kl. 10:58 skrev Helge Skrivervik ***@***.***>:
> 
>>  I would like to use QEMU=1 environment variable instead of ftpd=-q in /bootopts, so that FAT boots work with ftpd on QEMU.
>> 
> I support that - it's simpler and more flexible.
>> If you could add a change to ftpd.c that defaults the "qemu" variable to the value of environment variable QEMU=1 (or 0) that would be great.
>> 
> Certainly - I'll include it with some other fixes pending.
>> I can then change /boottopts and /bin/net so that we don't have to have the -q hack in there, and it will be self-contained and automatic in ftpd, at least. That should allow both internal and external ftp to work correctly in passive mode to it, correct?
>> 
> Well, internally (loopback mode), only active mode will work if the qemu hack is active. I would think that's good enough - ftp via loopback is a very special case anyway. It's also a setting I haven't tested much - yet.
>> have you tried Virtualbox?
>> 
>> Not in a while. I'll look into it and see how it works.
>> 
>> 
> Likewise - I'll pull it out from the digital drawer and see if we can get 'real' networking going.
> 
> Thank you.
> 
> -M
>

Mellvik:
Another update on emulators, @ghaerr:

It turns out we're out of luck on macOS and QEMU advanced networking. Apple does not support a TAP interface which is required to give QEMU its own network interface. And Apple's rather strong security policy makes it virtually impossible to use those implementations that have been available for more than 10 years. And developers stopped maintaining them for that reason.

On Linux/BSD and possibly Windows, it seems possible, I have not looked into that as it wouldn't help me … 

So it seems we're stuck with the current situation for now, unless the javascript emulator turns out to be able to do some magic.


—M

> 27. des. 2021 kl. 11:34 skrev Helge Skrivervik ***@***.***>:
> 
> Quick update on emulators, @ghaerr -
> 
> - VB has no NE2k support
> - Parallels may have it, but only the PCI variant. I don't know if that's workable.
> - DosBox has no Ethernet except possibly via DOS packet drivers.
> 
> 
> Looks like we're stuck with QEMU - so let's see if we can get tap/tun to work.
> 
> 
> —M
> 
>> 27. des. 2021 kl. 10:58 skrev Helge Skrivervik ***@***.*** ***@***.***>>:
>> 
>>>  I would like to use QEMU=1 environment variable instead of ftpd=-q in /bootopts, so that FAT boots work with ftpd on QEMU.
>>> 
>> I support that - it's simpler and more flexible.
>>> If you could add a change to ftpd.c that defaults the "qemu" variable to the value of environment variable QEMU=1 (or 0) that would be great.
>>> 
>> Certainly - I'll include it with some other fixes pending.
>>> I can then change /boottopts and /bin/net so that we don't have to have the -q hack in there, and it will be self-contained and automatic in ftpd, at least. That should allow both internal and external ftp to work correctly in passive mode to it, correct?
>>> 
>> Well, internally (loopback mode), only active mode will work if the qemu hack is active. I would think that's good enough - ftp via loopback is a very special case anyway. It's also a setting I haven't tested much - yet.
>>> have you tried Virtualbox?
>>> 
>>> Not in a while. I'll look into it and see how it works.
>>> 
>>> 
>> Likewise - I'll pull it out from the digital drawer and see if we can get 'real' networking going.
>> 
>> Thank you.
>> 
>> -M
>> 
>

ghaerr:
Hello @Mellvik,

Thanks for your research on emulators. It looked as if the macOS bridge article had some potential but I'm guessing it's not supported given Apple's later security fixes. BTW, I am using "DosBox-X" for PC-98 testing, and it's an enhancement from DosBox, more up-to-date, and supports networking, but only slirp (likely same as QEMU) and pcap.

I figured we might want or be forced to continue using QEMU, but lets look at the bright side - we're miles forward from where we were just a few months ago. Between the network fixes and your ftp and ftpd contributions, ELKS is more usable than ever, on real hardware (and more platforms) as well as within QEMU. So keeping QEMU as well as our (hopefully mostly internal) "kluge-fixes" in order to easily test is fine with me.

I think the key is to work out of the box as simply as possible. My macOS ftp automatically seems to select passive mode, and thus it connects to ELKS ftpd without any options. Does ELKS ftp start in active mode? If so, then, with just the proposed QEMU=1 addition in /bootopts or /root/.profile, ftp networking will also work on real hardware and QEMU without special setup. No one has to know about dash options (-q), it just works.

When you add the QEMU=1 option, go ahead and replace the outdated ftpd=-q option with it. I'll cleanup /etc/rc.d/rc.sys in a later commit (which kluge handles ftpd=-q). As you may recall, any identifier string in /bootopts with an equals sign is automatically placed in the environment for rc.sys, and any capitalized string is also automatically continued into the login environment. Thus, adding QEMU=1 will automatically cause everything to work. You might also add it to /root/.profile (commented out).

Thank you for research and contributions, we've come a long way!

Mellvik:
> I figured we might want or be forced to continue using QEMU, but lets look at the bright side - we're miles forward from where we were just a few months ago. Between the network fixes and your ftp and ftpd contributions, ELKS is more usable than ever, on real hardware (and more platforms) as well as within QEMU. So keeping QEMU as well as our (hopefully mostly internal) "kluge-fixes" in order to easily test is fine with me.
> 
Agreed! I believe it's time to think 0.5 real soon now .........
> I think the key is to work out of the box as simply as possible. My macOS ftp automatically seems to select passive mode, and thus it connects to ELKS ftpd without any options. Does ELKS ftp start in active mode? If so, then, with just the proposed QEMU=1 addition in /bootopts or /root/.profile, ftp networking will also work on real hardware and QEMU without special setup. No one has to know about dash options (-q), it just works.
> 
Actually - it's all in the client. The server doesn't  have modes, it just responds to commands. Every transfer (including directory transfers) is preceded by a command that establishes the data-connection, either passive or active. Most clients (including ours) start in passive mode these days (the simpler ones, like urlget, only support passive mode), so I think your scenario is pretty safe.
> When you add the QEMU=1 option, go ahead and replace the outdated ftpd=-q option with it. I'll cleanup /etc/rc.d/rc.sys in a later commit (which kluge handles ftpd=-q). As you may recall, any identifier string in /bootopts with an equals sign is automatically placed in the environment for rc.sys, and any capitalized string is also automatically continued into the login environment. Thus, adding QEMU=1 will automatically cause everything to work. You might also add it to /root/.profile (commented out).
> 
Coming - along with some other fixes, hopefully tomorrow. I want to get the wildcard support into the server, plus one or two commands that turn out to be more common than I thought.
> Thank you for research and contributions, we've come a long way!
> 
Likewise. It seems every new project opens a new can of worms, and what we get at the other end is much more than the project itself, it's a much better, more stable and robust, system.

--Mellvik

ghaerr:
@Mellvik:

> Which reminds me over another issue I mentioned a few daysback: Can we rename the images from .bin to .img? That would make us more compatible with the world and reduce hassle when using other emulators.

Should we rename .bin to .img before v0.5.0? 

The only real issue is whether this might be confusing for some users. However, if we're going to to it, we should probably do it before the next release.

Mellvik:
@ghaerr - 

yes, this is indeed a good time. And it may cause a few questions, but as long as the qemu.sh script is updated, we should be fine. 


Thank you!

—M

PR #1074, opened 2021-12-23
80188EB board code updates

cocus:
Bumps the total ram to 512kB; this is exactly how much I have on my board, but it's a nice number to set as a default. Enables the dynamic heap instead of limiting it (which was useful on the bringup, but not anymore).
Also defines the TIMER_IRQ so it can be used as a "logical" IRQ number, since the irq-8018x.c will translate this number to the actual integrated timer interrupt vector.

### DIFF
--- a/elks/include/arch/ports.h
+++ b/elks/include/arch/ports.h
@@ -45,6 +45,10 @@
 #define PIC2_DATA  0xA1
 #endif
 
+#ifdef CONFIG_ARCH_8018X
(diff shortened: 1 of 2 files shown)


ghaerr:
Thanks for the update @cocus.

I may add back in the comments and commented-out ability to specify the kernel heap size, since that could likely prove useful for small embedded systems. But I'll commit this now to keep our repos in sync.

PR #1075, opened 2021-12-24
[net] Add tcp hex packet display

ghaerr:
Adds ability to see incoming and outgoing TCP packet data for better debugging. Set using DEBUG_TCPDATA in ktcp config.h.

Each packet is displayed individually along with the direction (incoming or outgoing).

Hopefully this will help debug some of our off-by-one packet problems seen recently, by allowing a closer look of the actual data being sent and received by ktcp.

Example output logging into remote macOS telnet server from ELKS:
Screen Shot 2021-12-23 at 7 25 49 PM


Mellvik:
Another great christmas gift, @ghaerr.
And again, immediately useful.


Thank you.

--M

PR #1076, opened 2021-12-24
[net] Add ability to start specific daemons from bootopts

ghaerr:
Adds NETSTART= to /bootopts to specify daemons to start at boot or with `net start`, to allow for smaller memory use when desired.

Adds ability to specify quoted strings in /bootopts.

For example, use `NETSTART="telnetd ftpd"` to start just those daemons at boot or with `net start`.


Mellvik:
Good idea, @ghaerr.

Immediately helpful!

Thank you.

--M

PR #1078, opened 2021-12-25
Add console-direct-pc98

tyama501:
Please optimize Makefile if necessary.
Thank you.


ghaerr:
Hello @tyama501,

Looks super, very nice now PC-98 supports direct console and it's associated ANSI terminal benefits.

Does the PC-98 support inverse video or color on the display, or is the 2nd tvram byte only used for Japanese DBCS character selection? Also, do you know the encoding of bit 7 of the 1st byte?

Thank you!

PR #1080, opened 2021-12-28
Add color attribute to console-direct-pc98

tyama501:
This console writes the attribute segment A200 of PC-98 for color text.

### DIFF
--- a/elks/arch/i86/drivers/char/console-direct-pc98.c
+++ b/elks/arch/i86/drivers/char/console-direct-pc98.c
@@ -72,6 +72,7 @@ static unsigned short int NumConsoles = MAX_CONSOLES;
 int Current_VCminor = 0;
 int kraw = 0;
 unsigned VideoSeg = 0xA000;
+unsigned AttributeSeg = 0xA200;
 
 #ifdef CONFIG_EMUL_ANSI
 #define TERM_TYPE " emulating ANSI "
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tyama501,

I can see from https://github.com/jbruchon/elks/issues/1047#issuecomment-1002203926 that you have basic color working, but this modification is just a portion of all that needs to be done in order to implement color and/or reverse video in direct console.

For instance, all the console lower level functions, including `ScrollUp`, `ScrollDown` and `ClearRange` need to be modified, so that, when the screen scrolls or a range is cleared, the attribute bytes are also modified. Thus, I don't think we should accept this PR yet, until those modifications are done and you have been able to adequately test them.

You can test scrolling in kilo without arrow keys by just typing in characters, the insert mode will push other characters lower, scrolling them downwards.

To test reverse video, you can run `vi` with no arguments, and the last line should say "NO FILE" in reverse video.

In my opinion, a better way to do this is to write a single function which translates the IBM PC attribute byte into the PC-98 byte/word, and to call that function in only a single extra pokle function call at the end of VideoWrite, something like:
```
pokew(addr, (seg_t) AttributeSeg, conv_pcattr(C->attr));
```
In the ScrollUp/ScrollDown functions, add a seperate memcpy that also moves the PC-98 attribute words the same amount as the text words, but using a different segment.

In the ClearRange function, we need a constant defined value which indicates normal text, equivalent to conv_pcattr(A_NORMAL) but without the function call, to set the attribute word alongside the text pokew that combines the attribute and character for IBM PC.

The conv_attr function needs to be made much faster than you have it, perhaps using a fast table lookup to translate each 4 bits of the 8 bit IBM PC value, or the display speed of the PC-98 may suffer, in my opinion.

Thank you!

tyama501:
Inverting and Clear are working.

It is much better using tables but I need to speed up more.

attribute_after_modification_vi

"kilo" can scroll by return key but it seems it is not using the scroll functions.
I could scroll before adding the functions. Just using the write function.

attribute_after_modification_kilo

"more" command after "kilo".
Attributes are cleared and we can see with white color.

attribute_after_modification_more

Thank you.

ghaerr:
> Inverting and Clear are working.

Thanks for the screenshots, you're making great progress! It seems PC-98 direct console is now working well.


> It is much better using tables but I need to speed up more.

It seems that since the 16-byte translation table is duplicated in 2nd half, that could be changed to use half the space. Also, I notice that "C->vseg" could be removed from struct and replaced entirely with VideoSeg, as PC-98 has only one text mode video address, unlike IBM PC. Overall, it seems your revised conv_pcattr function is very good and not likely to slow things down very much.

> "kilo" can scroll by return key but it seems it is not using the scroll functions.

Oh yes, I forgot: I remember now it rewrites the entire screen with every character, and doesn't use fancy escape sequences.


Your code looks good, I will commit it now. If you wish to make further enhancements or changes, please go ahead and submit them, thank you!

PR #1081, opened 2021-12-28
[net, ftp] ftpd updated to support globbing, improved QEMU support

Mellvik:
Update to the ELKS `ftpd` FTP server:
- Pick up QEMU mode from the environment (`QEMU=1`) to simplify boot procedures (see #1071 (discussion))
- Added wildcard support (#1067)
- Added another (minimal) QEMU hack which enable full and transparent support for active mode file transfers when the QEMU option is enabled.
- Some cleanups and optimizations

For a change, this update has been primarily tested on QEMU. 
More cleanups and support for a few additional commands (EPRT, EPSV) planned.

### DIFF
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -21,6 +21,7 @@
 #include	<time.h>
 #include	<sys/stat.h>
 #include	<sys/types.h>
+#include 	<sys/wait.h>
 #include 	<dirent.h>
 
 #define		BLOAT
@@ -502,15 +503,11 @@ int do_retr(int controlfd, int datafd, char *input){
 				}
 			close(fd);
 		} else {
-			//str = "550 No Such File or Directory.\r\n";
-    			//write(controlfd, str, strlen(str));
 			send_reply(controlfd, 550, "No such file or directory");
     			return -1;
 		}
 	} else {  /* FIXME: this part may not be required */
 		if (debug) printf("File not found: %s\n", cmd_buf);
-		//str = "450 Requested file action not taken.\r\nFilename Not Detected.\r\n";
-    		//write(controlfd, str, strlen(str));
 		send_reply(controlfd, 450, "Requested file action not taken.\r\nFilename Not Detected");
 		return -1;
 	}
@@ -528,8 +525,6 @@ int do_stor(int controlfd, int datafd, char *input) {
 
 	if (get_filename(input, cmd_buf) < 0) {
 		if (debug) printf("No file specified.\n");
-		//str = "450 Requested action not taken - no file.\r\n";
-    		//write(controlfd, str, strlen(str));
 		send_reply(controlfd, 450, "Requested action not taken - no file");
 		return -1;
 	}
@@ -538,8 +533,6 @@ int do_stor(int controlfd, int datafd, char *input) {
 	//trim(cmd_buf);
 	if ((fp = open(cmd_buf, O_CREAT|O_RDWR|O_TRUNC, 0644)) < 1) {
 		perror("File create failure");
-		//str = "552 File create error, transfer aborted.\r\n";
-		//write(controlfd, str, strlen(str));
 		send_reply(controlfd, 552, "File create error, transfer aborted");
 		return -1;
 	}
@@ -553,8 +546,6 @@ int do_stor(int controlfd, int datafd, char *input) {
 			else
 				printf("File write error: %s\n", cmd_buf);
 			close(fp);
-			//str = "552 Storage space exceeded, transfer aborted.\r\n";
-			//write(controlfd, str, strlen(str));
 			send_reply(controlfd, 552, "Storage space exceeded, transfer aborted");
 			return -2;
 		}
@@ -628,7 +619,7 @@ int main(int argc, char **argv) {
 		perror("Error in listen");
 		exit(3);
 	}
-	if (debug) printf("ftpd running - debug on.\n");
+	//if (debug) printf("ftpd running - debug on.\n");
 	if (debug < 2) {
 		/* become daemon, debug output on 1 and 2*/
 		if ((ret = fork()) == -1) {
@@ -652,8 +643,9 @@ int main(int argc, char **argv) {
 			perror("Accept error:");
 			break;
 		}
-		strncpy(real_ip, in_ntoa(client.sin_addr.s_addr), 18); // Save for QEMU hack
-		if (debug) printf("New connection from %s\n", real_ip);
+		waitpid(-1, NULL, WNOHANG);		/* reap previous accepts*/
+		strncpy(real_ip, in_ntoa(client.sin_addr.s_addr), 20); // Save for QEMU hack
+		if (debug) printf("New connection from %s (0x%lx)\n", real_ip, client.sin_addr.s_addr);
 		/* child process */
 		if((pid = fork()) == 0) {
 			close(listenfd);
@@ -694,14 +686,9 @@ int main(int argc, char **argv) {
     					if ((datafd = setup_data_connection(client_ip, client_port, port)) < 0) {
 						if (debug) printf("PORT command failed.\n");
 						datafd = -1;
-						//str = "425 Can't open data connection.\r\n";
-						//write(connfd, str, strlen(str));
 						send_reply(connfd, 425, "Can't open data connection");
-					} else {
-						//str = "200 PORT command successful.\r\n";
-						//write(connfd, str, strlen(str));
+					} else 
 						send_reply(connfd, 200, "PORT command successful");
-					}
 					break;
 
 				case CMD_PASV: /* Enter Passive mode */
@@ -710,8 +697,6 @@ int main(int argc, char **argv) {
 						datafd = -1;
 					}
 					if (do_pasv(connfd, &datafd) < 0) {
-						//str = "502 PASV: Cannot open server socket.\r\n";
-						//write(connfd, str, strlen(str));
 						send_reply(connfd, 502, "PASV: Cannot open server socket");
 						close(datafd);
 						datafd = -1;
@@ -725,22 +710,15 @@ int main(int argc, char **argv) {
 		    					write(connfd, complete, strlen(complete));
 						else {
 							/* NLST - different response codes */
-							//str = "250 List completed successfully.\r\n";
-		    					//write(connfd, str, strlen(str));
-							send_reply(connfd, 250, "List completed successfully");
+							send_reply(connfd, 250, "List completed");
 						}
-					} else {
-						//str = "503 Bad sequence of commands.\r\n";
-						//write(connfd, str, strlen(str));
+					} else 
 						send_reply(connfd, 503, "Bad sequence of commands");
-					}
 					datafd = -1;
 					break;
 
     				case CMD_RETR: /* Retrieve files */
 					if (datafd < 0) { /* no data connection, don't even try ... */
-						//str = "426 Connection closed, transfer aborted.\r\n";
-		    				//write(connfd, str, strlen(str));
 						send_reply(connfd, 426, "Connection closed, transfer aborted");
 						break;
 					}
@@ -755,8 +733,6 @@ int main(int argc, char **argv) {
 
     				case CMD_STOR: /* Store files */
 					if (datafd < 0) { /* no data connection, don't even try ... */
-						//str = "426 Connection closed, transfer aborted.\r\n";
-		    				//write(connfd, str, strlen(str));
 						send_reply(connfd, 426, "Connection closed, transfer aborted");
 						break;
 					}
@@ -785,8 +761,6 @@ int main(int argc, char **argv) {
 					//str = strtok(command, " ");
 					//str = strtok(NULL, " ");
 					//type = *str;
-					//str = "200 Command OK.\r\n";
-		    			//write(connfd, str, strlen(str));
 					send_reply(connfd, 200, "Command OK");
 					break;
 #ifdef BLOAT
--- a/elkscmd/inet/ftp/ftpd.c
+++ b/elkscmd/inet/ftp/ftpd.c
@@ -21,7 +21,7 @@
 #include	<time.h>
 #include	<sys/stat.h>
 #include	<sys/types.h>
-#include 	<sys/wait.h>
+#include	<sys/wait.h>
 #include 	<dirent.h>
 
 #define		BLOAT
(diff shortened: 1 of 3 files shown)


Mellvik:
@ghaerr - I forgot that I was one behind on ftpd.c, what's the easy way to resolve?

BTW - the QEMU fix for active mode was more fun than really required, but its simplicity made it mandatory IMHO.

A fix for `ftp` (adding a QEMU mode for that too) which will enable file transfers from within QEMU, is coming. Possibly just picking up QEMU=1 from the environment like `ftpd`.

-M

ghaerr:
> BTW - the QEMU fix for active mode was more fun than really required, but its simplicity made it mandatory IMHO.

Thanks, I had figured it might come down to fixing ftp as well. I agree it's better to have things work with internal kluges though than make the user suffer with things not working. I'm interested in an explanation of what you had to do so I don't have to glean it from the code though, to keep me up on active/passive mode issues and QEMU :)


> A fix for ftp (adding a QEMU mode for that too) which will enable file transfers from within QEMU, is coming. Possibly just picking up QEMU=1 from the environment like ftpd.

Definitely use QEMU=1 from the environment if possible. And go ahead and replace the "#ftpd=-q" line in /bootopts with "#QEMU=1", and add the same in /root/.profile as well, to this PR (or it's replacement) if possible.

> I forgot that I was one behind on ftpd.c, what's the easy way to resolve?

Geez, it always seems that the same place in a file gets modified, huh? I'm terrible at git, you might try adding the single line "wait" change and pushing it up to see if that changes anything. If not, then easiest way I know is to copy the entire first textbox on this PR, close it, open another PR and paste your text comment so you don't have to retype it.

Regarding this PR (or its replacement), I'd like to see more of the following commented-out lines deleted, if possible:
```
			//str = "550 No Such File or Directory.\r\n";
     			//write(controlfd, str, strlen(str));
 			send_reply(controlfd, 550, "No such file or directory");
```
(And yes, `send_reply` is a nice addition).

Thank you!

Mellvik:
Thank you, @ghaerr.
> BTW - the QEMU fix for active mode was more fun than really required, but its simplicity made it mandatory IMHO.
> 
> Thanks, I had figured it might come down to fixing ftp as well. I agree it's better to have things work with internal kluges though than make the user suffer with things not working. I'm interested in an explanation of what you had to do so I don't have to glean it from the code though, to keep me up on active/passive mode issues and QEMU :)
> 
This one just 'landed' as I was making the qemu-change to the ftp client program. Then it became embarassingly obvious. 
In active mode, the server connects out, which is the easy direction with QEMU - no special mapping. The client sends the PORT command ( 'PORT x,y' where x is the ip-address, y is the port#), and the server makes the connection. The problem is that the given IP address is wrong. The client on the outside of QEMU is talking to 127.0.0.1, while the elks ftpd server is talking to the 'real address', in our case 10.0.2.2. Which can be picked up from the active command channel.
So what this hack does is almost like the passive mode hack, except we don't have to touch the port #: When qemu mode is active and the PORT command arrives, the server ignores the ip address in the PORT command and uses the 'real address' instead.  That's it. Just a few lines to ensure the address is easily available, very clean. So clean it was tempting to ditch the passive mode hack, but we cannot do that - too many cliens depend on it (don't support active mode).
The next question becomes whether ftpd can do this always instead of only when QEMU is set. The answer is yes. It would be non-standard, but it would work in 99.999% of all cases - because the endpoints of the command channel and the data channel are the same. Then it would inexplicably break when this is not the case, so I think we'll keep it like this.
> A fix for ftp (adding a QEMU mode for that too) which will enable file transfers from within QEMU, is coming. Possibly just picking up QEMU=1 from the environment like ftpd.
> 
> Definitely use QEMU=1 from the environment if possible. And go ahead and replace the "#ftpd=-q" line in /bootopts with "#QEMU=1", and add the same in /root/.profile as well, to this PR (or it's replacement) if possible.
> 
Ok, will do.
> I forgot that I was one behind on ftpd.c, what's the easy way to resolve?
> 
> Geez, it always seems that the same place in a file gets modified, huh? I'm terrible at git, you might try adding the single line "wait" change and pushing it up to see if that changes anything.
> 
Yes, time to learn some more git. Everything failed when I was trying to recover from the sash change recently. New attempt pending. BTW, since you have write access, I think it would be easy for you to use the 'resolve conflict' link and edit in the web UI, we can explore that if other attempts fail.
> If not, then easiest way I know is to copy the entire first textbox on this PR, close it, open another PR and paste your text comment so you don't have to retype it.
> 
BTW - there is a problem with ftpd serving internal clients on elks (loopback) when ftpd is in qemu mode. I haven't figured that one out yet. Try it! Then restart ftpd w/o QEMU=1 and it's fine.

--Mellvik

Mellvik:
@ghaerr,
my update didn't resolve anything -- maybe try the 'Resolve conflicts' button?

-M

ghaerr:
> maybe try the 'Resolve conflicts' button?

I just tried hitting the button, it doesn't let me edit anything!? The first "conflict" shows the exact same code. It appears the 2nd conflict with wait was resolved, now it's complaining about a secondary change to a following printf, it seems.

> The next question becomes whether ftpd can do this always instead of only when QEMU is set. The answer is yes. It would be non-standard, but it would work in 99.999% of all cases - because the endpoints of the command channel and the data channel are the same. Then it would inexplicably break when this is not the case, so I think we'll keep it like this.

It sounds to me that for our purposes, your fix would always work. So why not have it be the default, with a seperate option for that .01% case where the IP address has to be specified differently. Whenever I test your ftp/ftpd code, I most often run ftp to ftpd within ELKS for initial testing, then run in->out and out->in ('out' being macOS) also using QEMU (with QEMU=1 set). I would expect those three cases to run without errors. Would those work in this "fix" discussion?

> there is a problem with ftpd serving internal clients on elks (loopback) when ftpd is in qemu mode. 

I have to say, I'm back to getting confused again - we have four uses cases that need to work out of the box, correct? 
- external FTP into ELKS ftpd
- ELKS FTP out to external ftpd
- ELKS FTP to ELKS ftpd
- The above when running ELKS on QEMU (three cases)
Whenever ftp or ftpd is to be run on ELKS with QEMU, one must set the QEMU=1 flag, is that correct for the above three?

My question is: which of these six cases have been tested and work, which do not? For me to help, I have to run ELKS on QEMU, as even in my home office, I still don't have a working NE2K card on the desktop 386 (although do have SLIP).

ghaerr:
> I just tried hitting the button, it doesn't let me edit anything!? 

It appears I don't have the necessary permissions? Have you tried using that button on your copy of the repository, rather than this one? Perhaps that will work? 

What did you do in trying to fix this git mess, add back in the wait code, or take it out? It appears to be fussing about the added '#include <sys/wait>', did you just add that in the last commit, attempting to fix it?

If you are trying to recover your commits, so that at worst case we can delete this and push another PR, you can get those commits in patch format by finding the first of several commits in `git log`, then doing a `git show <commit>` and redirecting the output to a .patch file. Then, `git checkout master`, then `git pull upstream master`, `git push`, then checkout a new branch (new name), patch back your fixes, then submit as usual. You may end up having to delete the two problem areas from your patch file and add them manually, which is what we're trying to do now. I always run `patch -p1 --dry-run` so that patch doesn't mess my source up.

Another thought is to save your changes to your existing branch 'ftp' to a patch file as discussed above, then fixing things by pulling upstream to master etc as discussed above, then finally fixing things on a branch with the same name, and force-pushing that branch, which should update this same PR. That may work, although I've sometimes ended up with globs of merged-globs instead.

 I'm sorry but doing anything fancier for me usually results in a huge mess. I really don't understand git, obviously.

Mellvik:
> 
> I just tried hitting the button, it doesn't let me edit anything!?
> 
> It appears I don't have the necessary permissions? Have you tried using that button on your copy of the repository, rather than this one? Perhaps that will work?
> 
Hmmmm, I don't even get that button on my copy of the repository.
> What did you do in trying to fix this git mess, add back in the wait code, or take it out? It appears to be fussing about the added '#include <sys/wait>', did you just add that in the last commit, attempting to fix it?
> 
well, this one is really weird. In the first version of the PR, I didn't have this include - for obvious reasons, and it wasn't flagged by GIT either. Then, when I added it, suddenly it became a problem. NO clue ...
> If you are trying to recover your commits, so that at worst case we can delete this and push another PR, you can get those commits in patch format by finding the first of several commits in git log, then doing a git show <commit> and redirecting the output to a .patch file. Then, git checkout master, then git pull upstream master, git push, then checkout a new branch (new name), patch back your fixes, then submit as usual. You may end up having to delete the two problem areas from your patch file and add them manually, which is what we're trying to do now. I always run patch -p1 --dry-run so that patch doesn't mess my source up.
> 
> Another thought is to save your changes to your existing branch 'ftp' to a patch file as discussed above, then fixing things by pulling upstream to master etc as discussed above, then finally fixing things on a branch with the same name, and force-pushing that branch, which should update this same PR. That may work, although I've sometimes ended up with globs of merged-globs instead.
> 
I'll attack this one tomorrow. There is always something to learn, and if everything fails, I'll delete the PR and create a new one.
> I'm sorry but doing anything fancier for me usually results in a huge mess. I really don't understand git, obviously.
> 
> 
Actually I feel better now. We're obviously having very similar problems for - most likely - exactly the same reasons: We don't understand what's going on. That said, I think we're better at it than  a year or two ago …


Thank you!


—M

Mellvik:
> The next question becomes whether ftpd can do this always instead of only when QEMU is set. The answer is yes. It would be non-standard, but it would work in 99.999% of all cases - because the endpoints of the command channel and the data channel are the same. Then it would inexplicably break when this is not the case, so I think we'll keep it like this.
> 
> It sounds to me that for our purposes, your fix would always work. So why not have it be the default, with a seperate option for that .01% case where the IP address has to be specified differently. Whenever I test your ftp/ftpd code, I most often run ftp to ftpd within ELKS for initial testing, then run in->out and out->in ('out' being macOS) also using QEMU (with QEMU=1 set). I would expect those three cases to run without errors. Would those work in this "fix" discussion?
> 
@ghaerr,
there are a couple of special cases I need to check in order to commit to making that 'hack' permanent, one being the internal loopback situation.
> there is a problem with ftpd serving internal clients on elks (loopback) when ftpd is in qemu mode.
> 
> 
This is a bug of the weirdest kind. There is no obvious reason why the experienced behaviour should change between the two variants, so I need to look more at that one.
> I have to say, I'm back to getting confused again - we have four uses cases that need to work out of the box, correct?
> 
> external FTP into ELKS ftpd 
check
> ELKS FTP out to external ftpd
not in QEMU yet (the not-committed version I'm running is OK)
> ELKS FTP to ELKS ftpd
Fine, except when QEMU is on
> The above when running ELKS on QEMU (three cases)
> Whenever ftp or ftpd is to be run on ELKS with QEMU, one must set the QEMU=1 flag, is that correct for the above three?
> My question is: which of these six cases have been tested and work, which do not? For me to help, I have to run ELKS on QEMU, as even in my home office, I still don't have a working NE2K card on the desktop 386 (although do have SLIP).
> 
> 

so four out of six are ok, the final two coming when I finish testing.


Thank you.


-M

ghaerr:
@Mellvik,

Thanks for the report.

> so four out of six are ok, the final two coming when I finish testing.

Allright. Well, no problem committing this code and then fixing those cases later! I will wait until you say all things are working before doing my next test round. And thanks for cleaning up the commented out code.

After tomorrow's battle with git, depending on who wins, let me know when I should commit your changes, and we'll keep moving forward.

Thank you!

Mellvik:
@ghaerr -
the 'weird' problem with `ftp` internally on `elks` turns out to be caused by the recently introduced `waitpid` call in `ftpd`.

The problem is only visible on QEMU, so it apparently has to do with timing. What happens is that `accept()` responds twice to the same request. The second time, the remote address is screwed and the system hangs - see screenshot below. Commenting out `waitpid` makes the problem go away.

What's your take on this?

--Mellvik
Screenshot 2021-12-30 at 12 49 23

ghaerr:
> the 'weird' problem with ftp internally on elks turns out to be caused by the recently introduced waitpid call in ftpd.

Wow, how did you manage to figure that out?

Executing accept twice is very strange, can't seem to see how that could happen. The first byte in the returned structure is zeroed, probably because that structure portion was overwritten with another /dev/tcpdev call to ktcp, not sure.

It's almost as if ftpd didn't die when you executed the kill 31 above, and two ftpd's were responding! Or like the waitpid "resurrected" the killed ftpd.

> What happens is that accept() responds twice to the same request. 

That is absolutely strange, I have to say. Off the top of my head, I have no idea why accept would be executed twice, but I'll look into it.

> Commenting out waitpid makes the problem go away.

Very strange. It is in the same location as `telnetd` which doesn't exhibit any problems. Go ahead and delete it, so we have a working ftpd, and we'll re-figure the zombie solution afterwards. I'm wondering if it has something to do with the accept call returning the connection status in accept arguments, which telnetd does not do.

Would you like me to commit this now, now that you have fixed the merge conflict, or hold until the wait line is removed or otherwise debugged?

Thank you!

ghaerr:
@Mellvik,

I have partly debugged the accept problem, I think it has to do with your killing ftpd and restarting it. Does the bug occur if, directly after boot, you don't kill and restart an ftpd, but just use the booted version (or the single version started from net start)?

A potential issue arises when kill is used: the process doesn't actually get removed immediately, instead it is awakened from its block (in accept, in this case) in the kernel, and then, if there is no SIGTERM handler, it is forced to call exit internally. It could be that this is not happening and the previous ftpd is also allowed to continue in accept, for some reason. The waiting ftpd will always be in accept at the time of kill.

Nonetheless, I would prefer that we remove waitpid now, and I will debug this and the original zombie issue after you have these changes committed.

Mellvik:
@ghaerr,
> I have partly debugged the accept problem, I think it has to do with your killing ftpd and restarting it. Does the bug occur if, directly after boot, you don't kill and restart an ftpd, but just use the booted version (or the single version started from net start)?
> 
Yes, it happens always - regardless. The screenshot was not accurate in that sense - the previously running ftpd had the waitpid commented out for testing, so I killed it to get the bad one for the picture...
> A potential issue arises when kill is used: the process doesn't actually get removed immediately, instead it is awakened from its block (in accept, in this case) in the kernel, and then, if there is no SIGTERM handler, it is forced to call exit internally. It could be that this is not happening and the previous ftpd is also allowed to continue in accept, for some reason. The waiting ftpd will always be in accept at the time of kill.
> 
> Nonetheless, I would prefer that we remove waitpid now, and I will debug this and the original zombie issue after you have these changes committed.
> 
> 
OK, I'll keep it commented out for now - and I think it's best if you commit the PR and I submit a new one, so we don't run into more trouble.

Thank you.

—M

ghaerr:
> Yes, it happens always - regardless.

Very strange!! I'll commit this now and then look further into this new issue.

>  I'll keep it commented out for now 
>  I think it's best if you commit the PR and I submit a new one

Ok. Don't forget to pull before your next changes :)

Lets not submit a PR with the waitpid out until we get it debugged, or you have additional changes. I'll cleanup the remaining ftpd=-q kluge in rc.sys.

Thanks for all your hard work on ftp and ftpd!!

PR #1082, opened 2021-12-30
[net] Fix kernel/ftpd accept race bug

ghaerr:
@Mellvik,

Here's a very temporary fix for the accept problem identified in https://github.com/jbruchon/elks/pull/1081#issuecomment-1002996031.

It seems the problem is timing related, but unrelated to `waitpid`. This current PR adds debug support as well as two printf statements in ftpd that show when accept is called and returned from in the main loop. These printfs are required in order to affect the timing and keep accept working, so that you can continue to fix ftp and/or ftpd moving forward.

I noticed a further error in ftpd where the parent process wasn't differentiated properly after the fork(), this is fixed. (Such are problems with including very long switch statements after fork and in main - perhaps that should all be moved to a separate function).

I've also cleaned up the ftpd=-q kluge.

`ftp localhost` or `ftp elks` still not working. Also, `ftp 10.0.2.15` not working with QEMU=1. Current PR is tested working with QEMU=0.

Please go ahead with further fixes or enhancements to ftp or ftpd while I try to understand what is going on with accept. It seems to be related to calling accept too quickly. This is very strange, as telnetd does the exact same thing, seemingly. 

Are there any cases where ftpd would end up calling another accept, for instance for passive mode, immediately after the initial accept causing the forked child process? I'm trying to track down why ftpd is acting differently than telnetd. It is clear that the main loop in ftpd ends up passing through the 2nd accept, causing another child process. I'm wondering if this might be due to a seperate call to accept somewhere in ftpd that is handled in error by the kernel: that is, the parent ftpd is getting accept rather than an inner accept for the child ftpd process passive mode... ? In other words, perhaps two accepts occurring closely is the bug, even though they're in seperate processes.

### DIFF
--- a/elks/include/linuxmt/debug.h
+++ b/elks/include/linuxmt/debug.h
@@ -40,7 +40,7 @@
 #define DEBUG_SIG	0		/* signals*/
 #define DEBUG_SUP	0		/* superblock, mount, umount*/
 #define DEBUG_TTY	0		/* tty driver*/
-#define DEBUG_TUNE	1		/* tunable debug statements*/
+#define DEBUG_TUNE	0		/* tunable debug statements*/
 #define DEBUG_WAIT	0		/* wait, exit*/
 
 #if DEBUG_EVENT
--- a/elks/include/linuxmt/debug.h
+++ b/elks/include/linuxmt/debug.h
@@ -40,7 +40,7 @@
 #define DEBUG_SIG	0		/* signals*/
 #define DEBUG_SUP	0		/* superblock, mount, umount*/
 #define DEBUG_TTY	0		/* tty driver*/
-#define DEBUG_TUNE	0		/* tunable debug statements*/
+#define DEBUG_TUNE	1		/* tunable debug statements*/
 #define DEBUG_WAIT	0		/* wait, exit*/
 
 #if DEBUG_EVENT
--- a/elks/net/ipv4/af_inet.c
+++ b/elks/net/ipv4/af_inet.c
@@ -230,15 +230,16 @@ static int inet_accept(register struct socket *sock, struct socket *newsock, int
     tcpdev_inetwrite(cmd, sizeof(struct tdb_accept));
 
     /* Sleep until tcpdev has news */
-    while (bufin_sem == 0) {
-        //sock->flags |= SF_WAITDATA;
+    do {	/* always sleep once to prevent accept race condition #1082 */
+
         interruptible_sleep_on(sock->wait);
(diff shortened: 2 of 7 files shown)


ghaerr:
@Mellvik,

Found the accept problem, so this PR is no longer temporary. Long story short there was a kernel race condition in accept and a second accept (usually the ftpd main loop accept) was not being blocked.

I've tested on QEMU, can't get a failure. The following problems still exist, as noted above:
`ftp localhost` or `ftp elks` still not working. Also, `ftp 10.0.2.15` not working with QEMU=1. Current PR is tested working with QEMU=0 and QEMU=1 and no accept failures.

Hopefully this fix allows you to continue working getting ftp working with ftpd in either QEMU= mode, or at least refusing to connect instead of hanging. I did find that one can ^C out of ftp, but as you note in your source, ftpd will still be hung in an accept in its child handler.

Mellvik:
Wow, what can I say @ghaerr - 

This is great news - and certainy unsurprising that it was timing related, since physical hw worked fine - before and after the waitpid call. 
So you nailed another esoteric bug - just in time for new  years...
> Found the accept problem, so this PR is no longer temporary. Long story short there was a kernel race condition in accept and a second accept (usually the ftpd main loop accept) was not being blocked.
> 
> I've tested on QEMU, can't get a failure. The following problems still exist, as noted above:
> ftp localhost or ftp elks still not working. Also, ftp 10.0.2.15 not working with QEMU=1. Current PR is tested working with QEMU=0 and QEMU=1 and no accept failures.
> 
Yes, I need to get that ftp PR out today! I'm pulling your changes now and rerunning all the tests.
> Hopefully this fix allows you to continue working getting ftp working with ftpd in either QEMU= mode, or at least refusing to connect instead of hanging. I did find that one can ^C out of ftp, but as you note in your source, ftpd will still be hung in an accept in its child handler.
> 
Yes, I need to take a closer look at how the 'big' servers handle that, and scale it down to our needs.

Thank you!

-M
>

ghaerr:
Hello @Mellvik,

> Yes, I need to get that ftp PR out today! I'm pulling your changes now and rerunning all the tests.

When `ftp` is run without networking running, it gives a generic socket error. We should probably have it respond like telnet, which says "Can't open socket (check if ktcp is running)". It is possible we could display something when ftp requires QEMU=1 as well, depending on whether such errors can be determined prior to hanging, etc.

Happy New Year!

Mellvik:
Yes, that's a good idea @ghaerr - I'll add that.

And Happy New year to you too! I took a break, hope you did too!


—M

PR #1083, opened 2022-01-02
Use default attribute value for PC-98 in console-direct-pc98

tyama501:
To make display speed of the console faster.


ghaerr:
Thank you @tyama501. Did this help increase the speed noticeably, or not so much?

tyama501:

I think it got better for vi editor using the default attribute.

I haven't changed fmemcpyb since most of it uses word access.

PR #1084, opened 2022-01-02
Add kbd-poll-pc98

tyama501:
Add kbd-poll-pc98 and modified conio-pc98 for Arrow keys and Page Up, Page Down.

Roll Down key of PC-98 is assigned to Page Up.
Roll Up key of PC-98 is assigned to Page Down.

There is no keys for Home and End for PC-98, so it is not assigned for now.
We can assign combinational key like Roll Down + Up arrow for Home,
and Roll Up + Down arrow for End if necessary.

It seems the insert and delete keys are not handled in the original kbd-poll.
Please tell me if this is my misunderstanding.

Thank you.

### DIFF
--- a/elks/arch/i86/drivers/char/Makefile
+++ b/elks/arch/i86/drivers/char/Makefile
@@ -70,7 +70,7 @@ ifdef CONFIG_ARCH_PC98
 ifdef CONFIG_CONSOLE_HEADLESS
 OBJS += console-headless.o
 endif
-OBJS += conio-pc98.o conio-pc98-asm.o kbd-poll.o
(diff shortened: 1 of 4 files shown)


ghaerr:
Looks good @tyama501. It is interesting that PC-98 has separate BIOS functions to retrieve arrow keys versus roll keys. What happens when they are both pressed, does the BIOS keep a separate queue for each?

tyama501:
Thank you @ghaerr, 

> What happens when they are both pressed, does the BIOS keep a separate queue for each?

The returned value of int18 AH=04 is assigned to key bit by bit and it is supported multi press.
With AL = 06, the returned value include bits for Roll keys.
With AL = 07, the returned value include bits for Allow keys.
It is assumed to used in gaming.

Issue #1048, opened 2021-12-13
[net] ftp/telnet not working well together.


Mellvik:
The preamble to this issue can be found in #1046 .

**Summary:**
When telnetting into ELKS, then FTPing out of ELKS and doing a DIR command (long directory listing), the output from the listing hangs until a character is entered in `telnet`. This only happens when the listing is large, requiring more than one, probably many ( 3-4 ) packets. Small listings work as expected.

Looking closely at the packet trace, this is what seems to happen:
- When the `FTP DIR` (`LIST`) command is sent to the server, it immediately responds with as many packets as ELKS can eat.
- This happens too fast for the `telnetd` process to take away anything from the input buffer, so the receiving window on the `ftp` data channel shrinks and eventually ends up at zero.
- When the data receive window reaches 0, everything stops. The server keeps retransmitting its last ack packet and everything is at a standstill until something is entered into `telnet`, which apparently wakes the `telnetd` process so it starts eating off of the input buffer and regular operation resumes.

So - viewed from the outside, it looks like the FTP data receive window reaching zero prevents a wakeup of the process waiting to read the data.

@ghaerr - I suspect this may ring a bell already, let me know if there are supplementary tests the may be useful. I'm not sure this can be recreated using QEMU given that timing is a factor. I might repeat the test using ELKS as the `ftp server`, which would probably work just fine - but the other system was unavailable at the point of testing.

Trace below.

-M

```
10:50:24.019227 IP (tos 0x0, ttl 64, id 711, offset 0, flags [none], proto TCP (6), length 46)
    10.0.2.15.1027 > 10.0.2.2.ftp: Flags [P.], cksum 0x1523 (correct), seq 46:52, ack 534, win 4380, length 6: FTP, length: 6
	LIST
10:50:24.021080 IP (tos 0x10, ttl 64, id 24803, offset 0, flags [DF], proto TCP (6), length 95)
    10.0.2.2.ftp > 10.0.2.15.1027: Flags [P.], cksum 0x1862 (incorrect -> 0xb8c7), seq 534:589, ack 52, win 29200, length 55: FTP, length: 55
	150 Opening ASCII mode data connection for '/bin/ls'.
10:50:24.022817 IP (tos 0x0, ttl 64, id 712, offset 0, flags [none], proto TCP (6), length 87)
    10.0.2.15.telnet > 10.0.2.2.36264: Flags [P.], cksum 0x7aaf (correct), seq 25321:25368, ack 190, win 4380, length 47
[...]

PR #1095, opened 2022-01-07
[net, ftp] Another major update to the ftp client.

Mellvik:
### FTP - 3rd round

Another major update to `ftp` with bug fixes, cleanups and the following major improvements:

- Includes globbing via the wildcard routine introduced with `ftpd` recently
- Recognizes the QEMU=1 environment variable which activates qemu mode and enables outgoing file transfers in passive mode. Attempts to use active mode will emit a warning. Active mode support was implemented and tested, but deemed too complicated and not really useful.
- `ftp` will detect if a connection is local (loopback) and automatically disable qemu mode if active. This means that loopback is now fully supported in QEMU (requires a forthcoming update to `ftpd`).
- Improved detection of remote server timeouts.
- Now uses /etc/hosts (`in_gethostbyname`) to look up host names, supports both host and port # in all relevant situations (standard `ftp` syntax).
- Default debug level is now 1, which provides a reasonable level of verbosity and is the default among most ftp clients.
- Possibly most importantly, support for multiple command line arguments has been included. It may seem like a small thing, but it has wide-ranging effects on the usability of `ftp`, some of which are:
    - `get` and `put` can rename files on the fly: `get sourcefile destfile` and `put sourcefile destfile`.
    - `mget` and `mput` will accept a list of files to transfer: `mget file1 file2 file3 file4` etc. Max # of arguments is 30.
    - When globbing is on (default), every argument to `mget/mput` may contain wildcards.
    - The `dir` command can save output to a file: `dir remotedir savefile` (where remotedir usually is '.'). Obviously, `remotedir` cannot be omitted.

Some debug printfs remains in the code pending a final update that will include optimisations and cleanup. All debug output is disabled when debug=0 (default).

A File Transfer Wiki update is pending.

Issues
- Some times (in QEMU), `ftp` reports sys_brk errors. This issue is under investigation.
- If an illegal IP address is specified, `ktcp` will still hang hard - for about 60 seconds.
- See also #1048

### DIFF
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -4,20 +4,16 @@
  * by Helge Skrivervik - helge@skrivervik.com - December 2021
  *
  *	TODO:
- *	- split open/login into its own function, move them to the main command loop
- *	  (remove the requirement to have user name and pw on the command line)
- *	- cmd line server address argument must lookup in /etc/hosts
  *	- add support for multi-argument commands (e.g. <put sourcefile destfile> ...)
  *	- add missing commands (rmdir, mls, nlist, ...), 
- *	- Add port # to the open command
  *	- ... create common function for commands that are alomost alike
- *	- clean up messags and debug levels
  *	- fix trucation of file names in FAT, minix (warnings, name collisions)
  *	- lcd with no params should print current local directory
  *	- evaluate whether using select() is useful or just a complication
  *	... add ABORT support in PUT
  *	- handle server timeout (no activity): Unsolicited server input (code 421)
  *	- add '15632 bytes sent in 0.00 secs (27.5561 MB/s)' type message after transfer
+ *	- validity check on server host name/address to avoid system hang
  *	- more testing of error conditions & error messages
  */
 
@@ -25,6 +21,7 @@
 #ifdef  BLOATED
 #define GLOB		/* include wildcard processing */
 #endif
+#define	QEMUHACK	/* Include code for QEMU special treatment */
 
 #include <sys/types.h>
 #include <sys/socket.h>
@@ -42,6 +39,7 @@
 #include <sys/time.h>
 #include <dirent.h>
 #include <errno.h>
+#include <signal.h>
 #ifdef GLOB
 #include <pwd.h>
 #include <regex.h>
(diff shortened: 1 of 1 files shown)


ghaerr:
Wow @Mellvik, this is a fantastic update!!

I will review, comment and commit after going skiing today :)

Mellvik:
Thank you @ghaerr, hopefully I'll get the `ftpd` out tomorrow.

BTW, please notice that the comment about memory resources has been deleted. It was old and no longer correct.

-M

ghaerr:
> Some times (in QEMU), ftp reports sys_brk errors. This issue is under investigation.

The `expandwildcards` libc function may use up to 3K heap. The default is 4K, so any checkerboarding or other mallocs could easily cause heap overflow. Suggest increasing heap to perhaps 8K on link.

> If an illegal IP address is specified, ktcp will still hang hard - for about 60 seconds.

I didn't realize an application ever returned from a bad connect address, so that's good news. There is no internal timer on this, don't know how that is happening. Nonetheless, I plan on a ktcp fix on this so user programs don't have to deal with alarm() around the connect or other kluges.

Mellvik:
> Some times (in QEMU), ftp reports sys_brk errors. This issue is under investigation.
> 
> The expandwildcards libc function may use up to 3K heap. The default is 4K, so any checkerboarding or other mallocs could easily cause heap overflow. Suggest increasing heap to perhaps 8K on link.
> 
Will test that. Neither ftpd nor ftp call malloc, but both are using (some) buffered IO, and there may be other library components using malloc.
> If an illegal IP address is specified, ktcp will still hang hard - for about 60 seconds.
> 
> I didn't realize an application ever returned from a bad connect address, so that's good news. There is no internal timer on this, don't know how that is happening. Nonetheless, I plan on a ktcp fix on this so user programs don't have to deal with alarm() around the connect or other kluges.
> 
This was an accidental discovery, observed several times but not investigated. I need to create a repeatable scenario for this one.
The (currently disabled) alarm usage in ftp is a shortcut around using select in the main command loop. Most ftp servers eventually time out on non-activity and send a message on the command channel to that effect. In order to catch this (which we don't right now, and it's a convenience more than a necessity), we can either use select or have an alarm go off every once in a while. 
This one is on the backburner for now.

--M

PR #1097, opened 2022-01-09
[net, ftpd] Functional updates, bugfixes to ftpd

Mellvik:
This update to `ftpd` includes many optimizations, some fixes and the following functional enhancements:

- `ftpd` will now recognize a loopback connection when in QEMU mode - and turn it off, which is necessary for loopback to work in QEMU. 
- QEMU support is more robust - active and passive mode.
- More robust handling of single files vs. wildcards in `mget/mput`.

### DIFF
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -253,7 +253,7 @@ int send_cmd(int cmdfd, char *cmd) {
 /* Read and process user commands */
 
 int get_cmd(char *buffer, int len, char **argv) {
-	struct cmd_tab *c;
+	struct cmd_tab *c = cmdtab;
 	int cmdlen;
 	char *cmd;
 
@@ -270,7 +270,6 @@ int get_cmd(char *buffer, int len, char **argv) {
 	parse_cmd(buffer, argv);
 	cmd = *argv;
 	cmdlen = strlen(cmd);
-	c = cmdtab;
 
 	while (*c->cmd != '\0') {
 		//FIXME: Need to check command uniqueness @ 3 chars
@@ -295,7 +294,7 @@ int get_port_string(char *str, char *ip, int p0, int p1) {
 	return 1;
 }
 
-/* Decipher the PASV info from the server - oly called from do_passive() */
+/* Decipher the PASV info from the server - only called from do_passive() */
 
 int get_server_ip_port(char *str, char *server_ip, unsigned int *server_port) {
 	char *n[6];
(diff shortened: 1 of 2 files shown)


Mellvik:
Note: When `ftp` is run internally on ELKS (loopback), `ktcp` will run out of memory pretty quickly (after 3-4 files) when doing a `mget` in passive mode. In active (PORT) mode, there is no problem. This appears on QEMU and physical hardware.

ghaerr:
This looks great @Mellvik!!

> When ftp is run internally on ELKS (loopback), ktcp will run out of memory pretty quickly (after 3-4 files) when doing a mget in passive mode. In active (PORT) mode, there is no problem.

I'm guessing this is because the port isn't able to be reused quickly with SO_REUSEADDR? I'd like to think of a way around that, but haven't yet. We could possibly implement an programmable option to lower a socket's TIME_WAIT state to a lower value. This isn't recommended by the RFC's, but we're already very low and it would probably help this quite a bit. Another possibility would be somehow allocating tcp's socket memory outside the near data segment somehow.

Thank you!

ghaerr:
> In active (PORT) mode, there is no problem.

Another thought would be to issue a warning to the user in QEMU passive mode that active mode is recommended, passive mode may run out of memory, or something to that effect.

Mellvik:
> This looks great @Mellvik <https://github.com/Mellvik>!!
> 
> When ftp is run internally on ELKS (loopback), ktcp will run out of memory pretty quickly (after 3-4 files) when doing a mget in passive mode. In active (PORT) mode, there is no problem.
> 
> I'm guessing this is because the port isn't able to be reused quickly with SO_REUSEADDR?
> 
Yes, I believe this is the case. I want to do some more testing to see if I can provoke the same situation without loopback (e.g simultaneous multifile activity inbound and outbound to/from ELKS.
> I'd like to think of a way around that, but haven't yet. We could possibly implement an programmable option to lower a socket's TIME_WAIT state to a lower value. This isn't recommended by the RFC's, but we're already very low and it would probably help this quite a bit. Another possibility would be somehow allocating tcp's socket memory outside the near data segment somehow.
> 
Actually I was surprised to see the issue on real hardware, which is a couple of orders of magnitude slower than QEMU. Then again, I don't think this should be a high priority issue now, unless I find something interesting via the other tests.

I just found out (by using FileZilla) that the server is missing a couple of commands for moving files. Not high priority but coming.
Also, I was curious as to what FileZilla would do if I attempt to copy a device. So I visually 'lifted' /dev/fd0 to the mac, and sure enough (and very slow), it copied the entire floppy as it should. (!!)


Thank you!

—Mellvik

Mellvik:
Thank you @ghaerr,
yes, that's probably a good idea. 

Then again, I keep thinking that no one in the right mind (!) would do loopback ftp on ELKS anyway. It's really useful for stress testing and debugging only.
That said, the fact that it works as well as it does qualifies cork popping on behalf of ktcp. It has come a looong way in 12 months. Well done indeed!

It's time for ELKS BACK-TO-BACK. Coming up.

—M

PR #1100, opened 2022-01-11
[kernel] Enable interrupts during block I/O

ghaerr:
While reading through the ELKS low-level block I/O code, it was found that interrupts were left disabled needlessly for almost all of the time prior to calling the actual block request handlers (such as bioshd, ssd, ramdisk, etc)!!

ELKS uses BIOS for all disk I/O, but most BIOSes re-enable interrupts immediately upon entry to INT 13h. It is not clear how interrupts were reenabled after other block I/O (such as SSD or ramdisk), other than perhaps through the timer interrupt. The longer period of having interrupts disabled beforehand could have had an effect on serial or network activity, which this should help.

This is tested only on QEMU.

### DIFF
--- a/elks/arch/i86/drivers/block/ll_rw_blk.c
+++ b/elks/arch/i86/drivers/block/ll_rw_blk.c
@@ -152,6 +152,7 @@ static void add_request(struct blk_dev_struct *dev, struct request *req)
     mark_buffer_clean(req->rq_bh);
     if (!(tmp = dev->current_request)) {
 	dev->current_request = req;
+	set_irq();
 	(dev->request_fn) ();
     }
     else {
@@ -162,8 +163,8 @@ static void add_request(struct blk_dev_struct *dev, struct request *req)
 	}
 	req->rq_next = tmp->rq_next;
 	tmp->rq_next = req;
+	set_irq();
     }
-    set_irq();
 }
 
 static void make_request(unsigned short major, int rw, struct buffer_head *bh)


Mellvik:
This is very interesting indeed, @ghaerr.

THe first impression on real hardware is encouragning: the df() command is MUCH faster - almost instant instead of several seconds. Directory listings seem faster, more numbers later this week.

Thank you.


--M

Mellvik:
@ghaerr,

I did some more testing on this one, and I cannot see any detectable difference in any context that I've tried.

Regardless, good discovery and important fix.

Thank you.

-M

ghaerr:
> THe first impression on real hardware is encouragning: the df() command is MUCH faster - almost instant instead of several seconds. 

Are you running XMS buffers? That might be the reason for a very large speed increase.

> I did some more testing on this one, and I cannot see any detectable difference in any context that I've tried.

Thanks for the testing. This change should probably only show improvements like not dropping serial characters at very high speeds, or perhaps not dropping network packets at very high throughput.

Mellvik:
@ghaerr,
> THe first impression on real hardware is encouragning: the df() command is MUCH faster - almost instant instead of several seconds.
> 
> Are you running XMS buffers? That might be the reason for a very large speed increase.
> 
No, I'm not - but I should! thanks for the reminder.
I'm finding that df is now fast with and without this change, and I've been frustrated by the slowness of this command before (like as late as last month), so something seems too have changed :-) to the better.
> I did some more testing on this one, and I cannot see any detectable difference in any context that I've tried.
> 
> Thanks for the testing. This change should probably only show improvements like not dropping serial characters at very high speeds, or perhaps not dropping network packets at very high throughput.
> 
> 
Agreed. And we're circling back to the issue we've been discussing any number of times over the past few years - the lack of control over interrupts when doing BIOS I/O.
I was just dd'ing a floppy image to a physical floppy on ELKS (from the HD), and the system is almost at a complete standstill for the duration of the transfer. Like a ps in a different terminal takes 5+ seconds per line while the transfer is going on.

But this is not the time and place to revisit that road.

I'm working on #1048 right now, trying to narrow down that hang situation.

—M

ghaerr:
> But this is not the time and place to revisit that road.

No, that's a big problem and would require our own floppy driver and will certainly wait post v0.5.0. I have a few comments nonetheless for later, separate from the problem of the kernel not being able to context switch during BIOS I/O:

> dd'ing a floppy image to a physical floppy on ELKS (from the HD), and the system is almost at a complete standstill for the duration of the transfer.

Two additional problems are that ELKS will always write 1K blocks to disk with no attempts to take advantage of writing more than two sectors at a time; which could mean worst case sector-missing on certain systems.

Also, for disk reads, the current track-caching uses the same cache for hard disk as well as floppies. That is, the floppy (read) cache will be invalidated whenever the hard disk is read, and vice-versa. That doesn't affect your disk writing speed above, but ELKS has only a single (required) low-memory track cache buffers for any device read (and no track write caching).

Mellvik:
Thank you @ghaerr,

I fully agree, this is something to attack down the road.
Thanks for the run-through! I guess we at some point - before attacking them, need to prioritize these issues. If performance is the primary goal, what would be your priorities:
- native (non-BIOS) floppy driver
- variable multi-sector I/O
- per device buffers
- anything to gain by changing the low level write (buffers, DMA, ...)
- ...

Again - issues for a later date!

--M

ghaerr:
Thanks for the list, @Mellvik.

Not to belabor the points, but I'm seriously wondering how a non-BIOS floppy driver can be made to work in ELKS without a huge amount of work. I think perhaps better the other issues might be used instead to increase performance, and leave ELKS permanently linked with BIOS, at least for disk I/O. 

Here's some issues associated with the kernel with a non-BIOS floppy driver:
- Any change in floppy hardware would result in changes required in the floppy driver across each "PC" system.
- Mixing interrupt-driven floppy with BIOS hard disk may result in inability to process floppy interrupts, as hard disk BIOS I/O may be in progress. This means a hard disk driver also needs to be written in order to use the floppy driver.
- Any changes in hard disk hardware will result in the need for an additional hard disk driver for ELKS.
- The entire ELKS kernel must be relieved of using BIOS calls for any functions, or the ability to timeshare across processes or interrupts could be lost, or worse, prohibited from being allowed, thus losing any gains.
- The entire lower level ELKS block I/O subsystem needs to be rewritten to be asynchronous. While this would be a fun project, it requires all drivers to be non-BIOS, and cannot be used if any other drivers are non-asynchronous.
- In other words, the entire system and all drivers would need to be converted to asynchronous, interrupt-driven I/O. Any older or other systems (such as PC-98) would need a fallback path (meaning also retaining the current approach) to be compiled in instead. Thus ELKS would need two fully separate low level I/O subsystems to retain compatibility without having to write drivers to even try to use ELKS.

A long list!!

Thanks for your input, I hadn't realized the kernel issues involved until you brought up the 'ps' timesharing issue earlier.

Mellvik:
Thank you again for your usual thoroughness @ghaerr, much appreciated. An quite educational.

While I don't necessarily agree with all your points, I do agree with the conclusion - that using the BIOS for storage type I/O is beneficial for ELKS' purpose and not on the optimization table. I'm convinced there is more than ample potential for optimization before we get to that level.

Apropos - and just for the heck of it, I wanted to see how ELKS responds when there is no disk/floppy IO.
SO I started an outgoing file transfer (http via curl, which continuously reports the actual speed), which on an idle elks system gives about 97k per sec (pretty good for now, @ 512 bytes payload per packet). No disk IO, so the data source on ELKS is /dev/mem (which incidentally has no end). One physical interrupt per 512 bytes of data.
(BTW — my MAC's ftp reports 150k per sec when transferring files into ELKS these days, @ 1460 payload size per packet. I need to verify that, but it looks good).

Then I started a 'hd /dev/mem' from a serial line @ 57600 bps (with FIFO) which reduces the network speed by 90% - 97k to 9.7k. Surprisingly heavy for the serial load, which should interrupt about every 13 or 14th byte with the FIFO active (unless I misunderstood the FIFO,  and it was working only inwards). 

Then - and yes, I'm barking up an entirely different tree, but the opportunity was there - in order to see how much of a factor serial speed is, I reduced the rate to 38400, and the network speed took another small dive - 9.7 to 7.3. 
Surprising. I would have expected lower speed and thus lower traffic on the serial line to have the opposite effect, that the network speed would increase. It did not. So I continued.

[BTW This is when I found that stty <speed> doesn't work anymore - stty seemingly silently ignores the speed parameter, I have to look into that.]

Next level downwards in speed for the serial (16550A, w/FIFO) is 19200, which (again surprisingly) took the network speed down to 5.1k.

Growing curiosity sent me to the other serial line on the machine, which doesn't have a FIFO, and thus should represent a slightly higher load. It does not. At 19200, it loads the system slightly less than the FIFO port - reducing network speed to 5.5k instead of 5.1k. Marginal, but surprising. 
At 38400, the non-FIFO port has the same load factor as the FIFO-line at 57600, leading to the conclusion that there is minimal difference between the two.

Just to make sure, a new test was conducted without serial FIFO support in the kernel. As expected, the two serial lines now deliver equal numbers. And surprising, the load @ 57600 bps is lower than when FIFO was active, the network speed goes to 12-13k/s instead of ca. 9.5k.
As before, the load on the system increases with lower serial speed - at about the same rate as before.

I know I'm totally off topic - apologies for that. But at the end of the day, there are two interesting issues:

- Why does the load increase with lower serial speed?
- Was doesn't FIFO support help on the load?

Admittedly, when it comes to improving ELKS overall, spending time on storage in stead of serial is more valuable.

I'll be back with a PR for stty - probably next week.



Thank you!

—Mellvik

ghaerr:
> Was doesn't FIFO support help on the load?

You're talking about serial output, not input, right?

For serial output in ELKS, the FIFO isn't used at all; instead, the kernel busy-loops waiting for "transmitter buffer empty" for the next character to output. This is because serial output is not interrupt driven. I'll have to check whether a hardware FIFO is available for output characters, but that would lessen only the interrupt overhead load in the interrupt driven output case. Of course, interrupt-driven output would greatly decrease the kernel's time doing serial output.

>  Why does the load increase with lower serial speed?

I'm thinking because the kernel waits for even longer in between characters in a busy-loop!

> I'll be back with a PR for stty - probably next week.

How did you change the serial port speed?

Mellvik:
Thank you @ghaerr -
I realise we've been up and down this road a few times already - then again, every time we do, there is an improvement of some sort.
> Was doesn't FIFO support help on the load?
> 
> You're talking about serial output, not input, right?
> 
> For serial output in ELKS, the FIFO isn't used at all; instead, the kernel busy-loops waiting for "transmitter buffer empty" for the next character to output. This is because serial output is not interrupt driven. I'll have to check whether a hardware FIFO is available for output characters, but that would lessen only the interrupt overhead load in the interrupt driven output case. Of course, interrupt-driven output would greatly decrease the kernel's time doing serial output.
> 
Yes, we're talking output only, and I do recall we've been discussing the benefits of changing to interrupt driven output before. I was also unsure about whether the FIFO worked for output, and looked it up. It seems it does. 
And I keep thinking that it may be useful even without interrupts as a first step. If the busy loop waits for FIFO status instead of a single transmit register status, there may be something to gain. I'll read up on this and see what I can find.
> Why does the load increase with lower serial speed?
> 
> I'm thinking because the kernel waits for even longer in between characters in a busy-loop!
> 
OK; that makes sense - and makes it a prime target for optimization I guess. BTW - I haven tested this before, and 57600 baud is fine on my 386. 115200 is not.
> I'll be back with a PR for stty - probably next week.
> 
> How did you change the serial port speed?
> 
/etc/inittab … I did a zillion logins today. Wouldn't have done so without the XMS buffers making it REAL fast. Wonder why I haven't used that for a while…

Thank you!

—M

ghaerr:
> I was also unsure about whether the FIFO worked for output, and looked it up. It seems it does.

Good to know.

> And I keep thinking that it may be useful even without interrupts as a first step. If the busy loop waits for FIFO status instead of a single transmit register status, there may be something to gain.

This would only help for bursts of characters less than the output FIFO length. For any long stream of characters, like your constant /dev/mem stream, the first FIFO-length of characters would be queued extremely quickly, and all the remaining would have to wait for the FIFO status to be empty, which, on the average, would be just as slow as you're now seeing.

Adding interrupt driven serial output to v0.6.0 is probably worthing considering.

Mellvik:
> 14. jan. 2022 kl. 17:57 skrev Gregory Haerr ***@***.***>:
> 
> 
> I was also unsure about whether the FIFO worked for output, and looked it up. It seems it does.
> 
> Good to know.
> 
> And I keep thinking that it may be useful even without interrupts as a first step. If the busy loop waits for FIFO status instead of a single transmit register status, there may be something to gain.
> 
> This would only help for bursts of characters less than the output FIFO length. For any long stream of characters, like your constant /dev/mem stream, the first FIFO-length of characters would be queued extremely quickly, and all the remaining would have to wait for the FIFO status to be empty, which, on the average, would be just as slow as you're now seeing.
> 
> Adding interrupt driven serial output to v0.6.0 is probably worthing considering.
> 
> 
Sounds like a good plan,

Thank you!

-M

Mellvik:
Apropos serial, @ghaerr - 

any opinions on replacing slip with ppp in ktcp?

PPP seems to me like a more current way to do serial IP, meaning it's easier to find a peer that works …


—M

https://medium.com/swlh/connecting-a-286-dos-pc-to-the-internet-through-a-serial-connection-in-2019-b93a422ff094

> 14. jan. 2022 kl. 18:00 skrev Helge Skrivervik ***@***.***>:
> 
> 
> 
>> 14. jan. 2022 kl. 17:57 skrev Gregory Haerr ***@***.*** ***@***.***>>:
>> 
>> 
>> I was also unsure about whether the FIFO worked for output, and looked it up. It seems it does.
>> 
>> Good to know.
>> 
>> And I keep thinking that it may be useful even without interrupts as a first step. If the busy loop waits for FIFO status instead of a single transmit register status, there may be something to gain.
>> 
>> This would only help for bursts of characters less than the output FIFO length. For any long stream of characters, like your constant /dev/mem stream, the first FIFO-length of characters would be queued extremely quickly, and all the remaining would have to wait for the FIFO status to be empty, which, on the average, would be just as slow as you're now seeing.
>> 
>> Adding interrupt driven serial output to v0.6.0 is probably worthing considering.
>> 
>> 
> Sounds like a good plan,
> 
> Thank you!
> 
> -M

Mellvik:
Just for the record - there is a really tiny (and working)  `ppp` implementation for IPv6 on `github` - 
https://github.com/njh/tinyppp6

which will would probably lend itself well to IPv4-porting with reasonable effort.
It's known to work with the Linux ppp implementation.

ghaerr:
> any opinions on replacing slip with ppp in ktcp?

I'll take a look at these... do you think you could do the slip replacement?

Mellvik:
> any opinions on replacing slip with ppp in ktcp?
> 
> I'll take a look at these... do you think you could do the slip replacement?
> 
Yes, I'd like to take on this one. It seems more immediately useful than the 3c509 driver I have on the backburner.

I'd like to attack the serial output interrupt issue too, but I believe you are in a much better 'position' to do that. Sounds good?

--M

ghaerr:
> I'll take a look at these... do you think you could do the slip replacement?
> Yes, I'd like to take on this one.

Great! My thoughts for getting this going with minimal trouble are:
- Ensure SLIP works great between two systems. Study slip.c, as most of the initialization code can stay as-is. That leaves just two major procedures, slip_process for receiving packets and slip_send.
- Hopefully, a PPP replacement can be "inserted" into the existing slip.c source file (removing SLIP, adding PPP), without making any other changes to ktcp in the first pass. After this is working, the necessary extensions to allow ktcp to support SLIP and PPP can be added. IMO, this is an easier path than going about adding PPP protocol first separately.
- Finally, your slip.c can be renamed ppp.c, and the above extensions added for multiple protocol support.

Let me know any help you need.

> I'd like to attack the serial output interrupt issue too, but I believe you are in a much better 'position' to do that.

Sure. Adding interrupt driven serial output isn't necessary for the PPP support, so the projects won't necessarily be interlinked.

Thank you!

Mellvik:
@ghaerr -
> Ensure SLIP works great between two systems. Study slip.c, as most of the initialization code can stay as-is. That leaves just two major procedures, slip_process for receiving packets and slip_send.
> Hopefully, a PPP replacement can be "inserted" into the existing slip.c source file (removing SLIP, adding PPP), without making any other changes to ktcp in the first pass. After this is working, the necessary extensions to allow ktcp to support SLIP and PPP can be added. IMO, this is an easier path than going about adding PPP protocol first separately.
> Finally, your slip.c can be renamed ppp.c, and the above extensions added for multiple protocol support.
Yes, this is a good strategy.
> Let me know any help you need.
> 
Thank you.
> I'd like to attack the serial output interrupt issue too, but I believe you are in a much better 'position' to do that.
> 
> Sure. Adding interrupt driven serial output isn't necessary for the PPP support, so the projects won't necessarily be interlinked.
> 
Agreed. OTOH, as we experienced recently, there are significant performance implications, and having the ability to 'dump' (say) a full packet (MTU) into a buffer and let the interrupts do the rest, would be great - both ways. Now, that's an excessive buffer size for other uses, maybe an ioctl to set (or ask for) a specific buffer size (for all I know it may be there already for incoming).

Thinking loud again.

—M

ghaerr:
@Mellvik,

> there are significant performance implications, and having the ability to 'dump' (say) a full packet (MTU) into a buffer and let the interrupts do the rest, would be great - both ways. Now, that's an excessive buffer size for other uses, maybe an ioctl to set (or ask for) a specific buffer size (for all I know it may be there already for incoming).

Actually, the way `write` will work to an output interrupt-driven serial port will be the same as it does now: the application's write buffer is the only buffer; characters are not transferred to yet another output buffer. What happens now is that the kernel (serial driver) waits (busy loops) for the UART transmit buffer empty, then reads a character from the application's write buffer and transfers it directly to the UART. The application stays in the kernel, not returning from the write, until the entire buffer is transferred. 

When interrupt-driven serial output is implemented, the only difference will be that the kernel (serial driver) will put the application to sleep waiting for TBE, rather than busy loop. That will allow for other applications to run. If there are no other applications to run (which is the typical case for single users), then there won't be any significant performance increase either.

In either case, there is no need for setting output buffer sizes via ioctl.

Thank you!

Mellvik:
Thank you @ghaerr.

I'm very familiar with interrupt drive io in general and serial in particular, having written a few of those.

The idea that might have positive implications for both system performance and (less) data loss at higher serial speed is to use a packet (MTU) sized buffer, almost like the ethernet interface. Which would - as you point out - allow the process to sleep while the packet is transmitted, and - if the other end has limited receive capability, like a typical ELKS system - introduce a 'wake-up delay' (just the normal rescheduling delay) between packets which may initially seem like a drawback but may actually increase the usable speed. 
OTOH - since most systems ELKS will be talking to have almost unlimited receive speed capability compared to ELKS, this is not a big deal, and the primary  benefit of such a buffering scheme would be ELKS system performance.
Thinking along the same lines on input, which is interrupt-driven already, waking the system per packet would be very efficient and would only take an ioctl to set the biffer size (if needed) and the buffer level at which to wake up the recipient process.

Really some random ramblings about making the system more responsive and serial packet IO possibly faster and/or more reliable. We did something like this - in a very different setting - way back.


—M

Issue #1090, opened 2022-01-05
Misc issues


Mellvik:
Issues encountered during the last leg of `ftp`/`ftpd` testing/development. 
They may need their own individual problem reports for discussions, (or may be known issues, even features), but let's start here anyway. All verifiable in QEMU.

- `ktcp`: When an application repeatedly fails to close opened sockets, `ktcp` will eventually run out of memory. When it happens, that condition is not handled gracefully: A number of `sys_brk` errors are reported on the console, apparently not affecting the application to begin with. Eventually, the system hangs, see screenshot. It would be very beneficial if the out-of-resource condition was communicated back to the application, avoiding the hang.
- `alarm`: The `alarm()` library routine exists, but does not seem to work. This may be a known issue.
- [minor annoyance] A (presumably) debug output from `ktcp` to the console is emitting 10 or so blank lines every so often depending on the amount of traffic.
- If a program is running - and recently started so some of its blocks are in the buffer cache, and the program file is overwritten, and attempt to run the (now replaced) program will fail until the original copy has exited. Example: `ftp` is running on `ttyS0`. A new copy of `ftp` is retrieved via the net on `ttyS1`, overwriting the disk copy of the running instance. An attempt to run the new `ftp` will fail - either just exit or hang, until the running copy has terminated. Is this a feature or a bug?
```
# sum ftp            <---- this is the running copy (in a different window)
57517     42
# sum ftp           <----- New copy transferred via ftp, overwrote the original
27059     42
# ./ftp               <----- just returns to prompt
# sync
# ./ftp              <---- just returns to prompt
# ./ftp              <---- now the other copy has terminated
ftp> pwd
Not connected.
ftp> open 10.0.2.2
Connected to 10.0.2.2.
...
```

- @ghaerr, I've tried to recreate the too many open files problem I reported about a month ago, no luck. There is no console message, but the applications is responding with the correct `perror()` message.

<img width="734" alt="Skjermbilde 2022-01-04 kl  10 58 16" src="https://user-images.githubusercontent.com/3629880/148218651-b4561b57-1da2-4d89-8600-5e98164bff65.png">

ghaerr:
Hello @Mellvik! Here's comments on each of your findings:

[...]

PR #1104, opened 2022-01-12
[net, ftp] bugfixes in ftp

Mellvik:
Fixes the following issues:

- incoming (`get`) zero length files would some times cause ftp to hang because server closed connection to early. See #1048 (discussion).
- when open connection failed, the terminal input remained open and `ftp` would run out of memory.
- Improved server timeout detection.

### DIFF
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -221,6 +221,7 @@ int ask(char *source, char *name) {
 /* Read (and echo) reply from server */
 /* NOTE: The fd argument is not used (replaced by global fcmd for buffered IO). */
 /* fd is kept because it makes the code easier to read. */
+
 int get_reply(int fd, char *buf, int size, int dbg) {
 
 	if (fgets(buf, size, fcmd) == NULL) {
@@ -232,11 +233,18 @@ int get_reply(int fd, char *buf, int size, int dbg) {
 	return 1;
 }
 
-int is_connected(int cmdfd) {
+int is_connected(int cmdfd) {	// return 1 if connected, otherwise 0
 	char str[IOBUFLEN];
 
-	if (send_cmd(cmdfd, "NOOP\r\n") < 0) return errno;	/* verify connection first, detect timeout */
-	return (get_reply(cmdfd, str, IOBUFLEN, 1));
+	if (cmdfd < 0) 
+		return 0;	/* no cmd channel has been opened */
+	if (send_cmd(cmdfd, "NOOP\r\n") < 0) {
+		fclose(fcmd);
+		close(cmdfd);
+		cmdfd = -1;
+		return 0;
+	}
+	return ((get_reply(cmdfd, str, IOBUFLEN, 1) > 0)? 1 : 0 );
 }
 
 /* Send command to server, echo if required */
@@ -488,7 +496,7 @@ int do_ls(int controlfd, int datafd, char **cmdline, int mode) {
 
 int do_get(int controlfd, char *src, char *dst, int mode) {
 	char iobuf[IOBUFLEN+1];
-	int status = 1, bcnt = 0, fd, n, datafd = -1, icount = 0;
+	int status = 1, bcnt = 0, fd, n, datafd = -1;
 	int maxfdp1, data_finished = FALSE, control_finished = FALSE;
 	fd_set rdset;
 
@@ -537,7 +545,7 @@ int do_get(int controlfd, char *src, char *dst, int mode) {
 	 * So when get_reply() fetches the former, the latter ends up in the input buffer,
 	 * and the loop below will not receive anything on the control channel.
 	 */
-	struct timeval tv; int select_return;
+	struct timeval tv; int select_return, icount = 0;
 	while (1) {
 		if (control_finished == FALSE) FD_SET(controlfd, &rdset);
 		if (data_finished == FALSE) FD_SET(datafd, &rdset);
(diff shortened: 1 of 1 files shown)


ghaerr:
@Mellvik,

Thanks for identifying and fixing the zero-length file problem.

Is this ftp out-of-memory fix the only one, or is there still a ktcp memory problem?

Thank you!

Mellvik:
Thank you @ghaerr,
> Thanks for identifying and fixing the zero-length file problem.
> 
Yes, the problem is fixed as far as ftp is concerned, and it may be too much of a special case to go after the real problem in ktcp.
> Is this ftp out-of-memory fix the only one, or is there still a ktcp memory problem?
> 
Ktcp is in the clear. It runs out of memory when it should, so to speak - too many sockets, but the crunch is handled properly. I've been running two ftp clients and one active server concurrently on elks and get lots of sbrk errors (and failed file transfers), but no crashes. Really good.

I found that - when all this is going on concurrently, netstat some times report the same connection twice. Again this is probably too esoteric to worry about.

I'm still experimenting with heap numbers on ftp and ftpd - expandwildcards some times seem to eat more that reasonable.

--M

Mellvik:
An additional comment about `ktcp`, @ghaerr:

Under heavy load, `ktcp` will eventually hang.
An incoming ftp connection doing many back-to-back outbound (`mget *` from the client) transfers in PORT mode (ELKS connection back to the client for the data channel), will generate a lot of
```
  tcp: port 20 REUSED, freeing previous socket in time_wait                     
  tcp: port 20 NOT reused, previous socket in state 4     
```
and an occasional 
```
Bind: Could not connect on port 20                                              
bind error:   Address already in use 
```
The 'strained' conditions seem to be handled gracefully though, there are no transfer errors.
When an outbound `ftp` is running concurrently with the one above, movie back-to-back files into ELKS (`mget *`), there will be occasional retransmits:
```
tcp retrans: seq 10501 size 512 rcvwnd 65535 unack 988 rto 16 rtt 4 (RETRY 1 cnt 2 mem 1028)                                                        
tcp retrans: seq 10025 size 476 rcvwnd 65535 unack 988 rto 20 rtt 4 (RETRY 1 cnt 2 mem 1028)
```
 It seems to me that either the combination or the latter is what eventually causes `ktcp` to hang. At this point, `ktcp` may be killed and the connections will eventually time out.

ghaerr:
> Under heavy load, ktcp will eventually hang.

Thanks for the detailed report, Helge. 

You are really putting ELKS through some heavy work, and overall it is performing quite well. I think we may be reaching the higher side of network reliability for ELKS, although I'm sure there's more bugs to be found. 

For instance, I have a working `alarm` (requested in #1090), that I also added code to work with networking to allow a `connect` to return in, say 10 seconds, if not connected. However, there are problems with the architecture of ktcp being outside the kernel: the client/server approach between them does not allow for unexpected events to be communicated easily. If the alarm fires, the kernel is interrupted and returns -ETIMEDOUT, but ktcp doesn't know about any of this, and there's no easy way to tell it. It's still got an open TCP CB, waiting for a SYN response to its SYN sent. That special handling code needs to be written to allow alarm to be used with arbitrary ELKS networking programs. I don't see any easy way to do it that will always work.

Of course, the ktcp lower level could be modified to somehow accept a "cancel" request, but this would also have to be done for "accept", and possibly all other network operations, and each have their own race conditions that might pop up later. For instance, if the alarm went off, but just before the cancel request, the ktcp accepted a connection, this could leave ktcp in a state where further connection requests wouldn't work, etc. The asynchronous nature of the kernel and ktcp is making things quite complicated at the level of reliability we're at.

I'm sure all of this, including reliability issues, could be solved of course, but I feel it may be getting exponentially harder to handle all the cases.

Given all your testing and simultaneous applications and transfers running, I'd say ELKS is doing quite well.

Mellvik:
Thank you, @ghaerr - 
I agree - ELKS is doing really well on most fronts these days. And yes, I'm pushing reasonable limits when testing, and while doing so it's easy to forget that we're not creating the next VAX - for 100 serial and 50 network users. But rather a single user multi process system which can do wonders with limited resources. This is exactly what ELKS does.
Which brings up a network related issue we've touched before: Creating a container for packet drivers to make them work in ELKS. It may well be more efficient use of time than writing (say) new network drivers.

Back to the case at hand, I fully appreciate the complications you describe and understand their reason and consequences. Now, with alarm() or setitimer() available, how about delegating the issue to ktcp without involving the kernel? Could ktcp have its own little timeout loop kicking off every second, checking on (say) SYNs sent and not responded to? Assuming ktcp has the capability to tear down the incomplete connection in the same way it would if a FIN arrived.

Thanks again for a thorough rundown.

--Mellvik

ghaerr:
Hello @Mellvik,

> how about delegating the issue to ktcp without involving the kernel? Could ktcp have its own little timeout loop kicking off every second, checking on (say) SYNs sent and not responded to?

Yes, that's probably a good way to solve the connect issue. Instead of using an application-based alarm, ktcp could just handle connect timeouts automatically, with a much lower timeout - say 10 seconds. This could be changed with a new SO_TIMEOUT option.

For accept timeouts, we could also use SO_TIMEOUT or something similar. The kernel would not be involved in any timeout, instead, ktcp would notify the kernel that the accept (or connect, as in above) had expired.

I'll add connect timeouts to my list. Do we really need an accept timeout capability?

Regarding alarm - I'll just keep the current implementation on hold. It could be useful in general, but if used for network timeouts, it will introduce unreliability into the networking system. Thoughts?

Thank you!

Mellvik:
@ghaerr,

> 
> how about delegating the issue to ktcp without involving the kernel? Could ktcp have its own little timeout loop kicking off every second, checking on (say) SYNs sent and not responded to?
> 
> Yes, that's probably a good way to solve the connect issue. Instead of using an application-based alarm, ktcp could just handle connect timeouts automatically, with a much lower timeout - say 10 seconds. This could be changed with a new SO_TIMEOUT option.
> 
Yes this is good!
> For accept timeouts, we could also use SO_TIMEOUT or something similar. The kernel would not be involved in any timeout, instead, ktcp would notify the kernel that the accept (or connect, as in above) had expired.
> 
> I'll add connect timeouts to my list. Do we really need an accept timeout capability?
> 
IIRC accept (in most systems) can be declared nonblocking  - I don't know if it's implemented in ELKS. If it isn't I'd think it's better to put that on the list.
> Regarding alarm - I'll just keep the current implementation on hold. It could be useful in general, but if used for network timeouts, it will introduce unreliability into the networking system. Thoughts?
> 
You're seeing sides of this that I don't. My take is that it would be very useful for (say) daemons to be able to close down idle connections. Such as ftpd - closing down after 900 seconds, which seems to be the norm. But there is noe rush on this one.

Thank you - we're getting there.

-M

PR #1105, opened 2022-01-14
[kernel] Move buffer init after console init, add bufs= to /bootopts

ghaerr:
Requested by https://github.com/jbruchon/elks/pull/1018#issuecomment-1013040317.

Allows xms: startup to display on serial console.

Use new 'bufs=xxx' /bootopts option to specify either xms (if compiled in and available) or ext (external main memory) buffer count.

### DIFF
--- a/elks/arch/i86/mm/xms.c
+++ b/elks/arch/i86/mm/xms.c
@@ -49,7 +49,7 @@ int xms_init(void)
 	enabled = verify_a20();
 	printk(" now %s, ", enabled? "on" : "off");
 #ifdef CONFIG_FS_XMS_INT15
-	printk("using int 15,");
+	printk("using int 15, ");
 #else
 	if (!enabled) {
 		printk("xms disabled, A20 error,");
@@ -59,7 +59,7 @@ int xms_init(void)
 		printk("xms disabled, requires 386,");
 		return 0;
 	}
-	printk("using unreal mode,");
+	printk("using unreal mode, ");
 #endif
 	xms_enabled = 1;	/* enables xms_fmemcpyw()*/
 	return xms_enabled;
--- a/elks/fs/buffer.c
+++ b/elks/fs/buffer.c
@@ -26,6 +26,8 @@
 #define NR_MAPBUFS  8
 #endif
 
+int boot_bufs;		/* /bootopts # buffers override */
+
 /* Buffer heads: local heap allocated */
 static struct buffer_head *buffer_heads;
 
@@ -124,17 +126,20 @@ int INITPROC buffer_init(void)
     /* XMS buffers override EXT buffers override internal buffers*/
 #if defined(CONFIG_FS_EXTERNAL_BUFFER) || defined(CONFIG_FS_XMS_BUFFER)
     int bufs_to_alloc = CONFIG_FS_NR_EXT_BUFFERS;
+    int xms_enabled = 0;
 
 #ifdef CONFIG_FS_XMS_BUFFER
-    int xms_enabled = xms_init();	/* try to enable unreal mode and A20 gate*/
+    xms_enabled = xms_init();	/* try to enable unreal mode and A20 gate*/
     if (xms_enabled)
 	bufs_to_alloc = CONFIG_FS_NR_XMS_BUFFERS;
-    printk(" %d %s buffers\n", bufs_to_alloc, xms_enabled? "xms": "ext");
 #endif
-
+    if (boot_bufs)
+	bufs_to_alloc = boot_bufs;
+    printk("%d %s buffers\n", bufs_to_alloc, xms_enabled? "xms": "ext");
 #else
     int bufs_to_alloc = NR_MAPBUFS;
 #endif
+
     buffer_heads = heap_alloc(bufs_to_alloc * sizeof(struct buffer_head),
 	HEAP_TAG_BUFHEAD|HEAP_TAG_CLEAR);
     if (!buffer_heads) return 1;
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -46,6 +46,7 @@ static char *envp_init[MAX_INIT_ENVS+1];
 static unsigned char options[OPTSEGSZ];
 
 extern int boot_rootdev;
+extern int boot_bufs;
 extern int dprintk_on;
 static char * INITPROC root_dev_name(int dev);
 static int INITPROC parse_options(void);
@@ -86,9 +87,6 @@ void INITPROC kernel_init(void)
     sched_init();
     setup_arch(&base, &end);
     mm_init(base, end);
-    if (buffer_init())	/* also enables xms and unreal mode if configured and possible*/
-	panic("No buf mem");
-    inode_init();
     irq_init();
     tty_init();
 
@@ -106,6 +104,10 @@ void INITPROC kernel_init(void)
 #ifdef CONFIG_CHAR_DEV_RS
     serial_init();
 #endif
+
+    inode_init();
+    if (buffer_init())	/* also enables xms and unreal mode if configured and possible*/
+	panic("No buf mem");
 
     device_init();
 
@@ -316,6 +318,10 @@ static int INITPROC parse_options(void)
 		}
 		if (!strncmp(line,"netport=",8)) {
 			net_port = (int)simple_strtol(line+8, 16);
+			continue;
+		}
+		if (!strncmp(line,"bufs=",5)) {
+			boot_bufs = (int)simple_strtol(line+5, 10);
 			continue;
 		}
 		
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -4,9 +4,10 @@
 #QEMU=1	net=eth		# ftpd on QEMU
 #HOSTNAME=elks1	 	# set network IP from /etc/hosts
 #GATEWAY=10.0.2.2
-#NETMASK=255.255.255.0
+#NETMASK=
 #NETSTART="telnetd ftpd"
 #netirq=9 netport=0x300 # NIC
+#bufs=128
 #init=/bin/init 3 n	# multiuser serial no /etc/rc.d/rc.sys
 #init=/bin/sh		# singleuser shell
 #console=ttyS0,19200	# serial console


Mellvik:
Thank you @ghaerr,

Works great,
Are you saying that bufs= will always allocate XMS buffers if XMS is compiled in and otherwise allocate 'regular' external buffers?

-M

ghaerr:
> Are you saying that bufs= will always allocate XMS buffers if XMS is compiled in and otherwise allocate 'regular' external buffers?

Yes. Except if XMS is compiled in and INT 15 or unreal mode is not available, in which case ELKS will use external buffers rather than failing. The config allows for different number of xms vs ext buffers, but /bootopts just allows for one - the idea being that one likely knows whether XMS is available or not.

Mellvik:
> 
> Are you saying that bufs= will always allocate XMS buffers if XMS is compiled in and otherwise allocate 'regular' external buffers?
> 
> Yes. Except if XMS is compiled in and INT 15 or unreal mode is not available, in which case ELKS will use external buffers rather than failing. The config allows for different number of xms vs ext buffers, but /bootopts just allows for one - the idea being that one likely knows whether XMS is available or not.
> 
Thank you @ghaerr, I believe this is an important improvement. Of course it may be me, but I kept on w/o using the xms buffers - forgetting that the final hurdles with the Compaq and unreal mode had been passed. Having the bufs= entry in bootopts is (believe it or not) an important reminder.

Which brings up a different, but somewhat related issue:
I'm inclined to have ftpd look for another env variable , FTPD, To set default options from /etc/profile. That could include a -q, but it makes sense to keep the QEMU variable, there may be other apps in addition to ftp and ftpd that need or benefit from knowing about qemu.
Thoughts?

-M

ghaerr:
Hello @Mellvik,

> I'm inclined to have ftpd look for another env variable , FTPD, To set default options from /etc/profile. That could include a -q, but it makes sense to keep the QEMU variable, there may be other apps in addition to ftp and ftpd that need or benefit from knowing about qemu. Thoughts?

Overall, I'm not a big fan of environment variable (or command line) options to servers unless we have to. IMO, adding options that users need to learn about, rather than the software just working, can end up being like the difference between trying to run programs on Windows versus iOS.

I realize we've added a QEMU= variable, but would love to figure out a way to remove it; we're already having to deal with multiple methods of file transfers depending on that variable. If it is at all possible for the application to figure things out instead of the user, I'm all for it.

That said, if you're just looking for a more temporary way of helping debug software you are developing, then perhaps such a variable should be added, but not planned for long term. Some thoughts on this are:

Using FTPD= in /etc/profile (or /bootopts) for multiple options introduces lots more code in the application, since the multiple options aren't passed into argc/argv format by a shell, but instead remain in a space-separated format, requiring special and possibly duplicate option parsing that isn't compatible with getopts. This is because we removed the older /etc/rc.d/rc.sys method of invoking a server with special environment options (a good thing).

What options are you considering?

> there may be other apps in addition to ftp and ftpd that need or benefit from knowing about qemu.

Like what? The less the better - we've already opened pandora's box, and allowing separate code execution paths for emulator and real hardware sounds like a way to let more bugs in, I think.

Thank you!

Mellvik:
Thank you @ghaerr,
> I'm inclined to have ftpd look for another env variable , FTPD, To set default options from /etc/profile. That could include a -q, but it makes sense to keep the QEMU variable, there may be other apps in addition to ftp and ftpd that need or benefit from knowing about qemu. Thoughts?
> 
> Overall, I'm not a big fan of environment variable (or command line) options to servers unless we have to. IMO, adding options that users need to learn about, rather than the software just working, can end up being like the difference between trying to run programs on Windows versus iOS.
> 
I share your bias, and  sometimes options are necessary - or possibly just convenient. In particular for debugging scenarios which rarely involves 'regular' users. Of course ELKS has an advantage compared to a more general linux case, we don't have to take into account every possible architecture and scenario. What I'm looking for is a way to avoid changing (say) /bin/net all the time.
> I realize we've added a QEMU= variable, but would love to figure out a way to remove it; we're already having to deal with multiple methods of file transfers depending on that variable. If it is at all possible for the application to figure things out instead of the user, I'm all for it.
> 
It's really an interesting oxymoron- qemu's raison d'etre it to be invisible, yet we're searching for ways to 'see' it. I honestly think we've done well in hiding the networking trickery of qemu in ftp and the QEMU env variable. After all qemu is a debugging environment.
> That said, if you're just looking for a more temporary way of helping debug software you are developing, then perhaps such a variable should be added, but not planned for long term. Some thoughts on this are:
> 
> Using FTPD= in /etc/profile (or /bootopts) for multiple options introduces lots more code in the application, since the multiple options aren't passed into argc/argv format by a shell, but instead remain in a space-separated format, requiring special and possibly duplicate option parsing that isn't compatible with getopts. This is because we removed the older /etc/rc.d/rc.sys method of invoking a server with special environment options (a good thing).
> 
Hey, we're discussing a couple of options, not a shipload. Btw, getopts isn't in ftpd and I doubt it is in  any of the other daemons (I hate getopts, but that's a different story). Parsing a few options from two sources is nothing really.
> What options are you considering?
> 
> there may be other apps in addition to ftp and ftpd that need or benefit from knowing about qemu.
> 
> Like what? The less the better
> 
Agreed, and adding more (than the current two) needs careful consideration: given that qemu is a development environment, how much adaption to this artificial environment do we want to create (ease of use motivation)? OTOH, we want more developers, lets make it easy for them, and make sure urlget/put, httpd etc work smoothly in qemu. No, I'm not suggesting that, just thinking loud.

> - we've already opened pandora's box, and allowing separate code execution paths for emulator and real hardware sounds like a way to let more bugs in, I think.
> 
We're really on the same page here, and the bottom line is: We don't have a /etc/inet.conf or the newer variants, but we need a way to 'configure' the daemons. We can do it in /bin/net, but it doesn't taste good, does it?
> 
Thank you!

-M

ghaerr:
> What I'm looking for is a way to avoid changing (say) /bin/net all the time.
> We can do it in /bin/net, but it doesn't taste good, does it?

What sorts of things are you having to modify /bin/net for? I'm trying to get a feel for how ELKS networking should ultimately be configured and be able to be configured. I agree that /bin/net should not be edited.

> We don't have a /etc/inet.conf or the newer variants, but we need a way to 'configure' the daemons. 

I suppose this is the same as the first question - what current and future options are you seeing the need for?

>  I honestly think we've done well in hiding the networking trickery of qemu in ftp and the QEMU env variable. After all qemu is a debugging environment.

Yes, you have done well with your QEMU hacks hidden within ftp and ftpd. Myself, I'm mostly interested in being able to quickly run ELKS on QEMU (and real hardware) and have most things just work (or when they don't, allow me to reproduce quickly). I also wish to expand the ELKS user base by making it easy to use out of the box (which we have). That said, I'm not trying to run a production environment like you are. 

Perhaps a seperate /etc/net.conf file makes sense for /bin/net options. I like /bin/net because its a shell script and easily updated. A single /etc/net.conf file may be advantageous to more and more /bootopts and/or /root/.profile options. If a /etc/net.conf file were actually a shell script itself (setting environment variables or other custom stuff), then /bin/net could source that file directly and things would be very easy. We might not have to write more code to parse files, just execute it as a shell script to set options within /bin/net and environment variables only visible during the /etc/rc.d/rc.sys execution.

ghaerr:
@Mellvik, perhaps something like this:
```
#
# /etc/net.cnf - ELKS Networking Configuration File (Shell Script)
#
# sourced by /bin/net to start ktcp and daemons

# ktcp options
localip=10.0.2.15
gateway=10.0.2.2
netmask=255.255.255.0
device=/dev/ttyS0
link=eth
baud=38400

# /etc/hosts override
HOSTNAME=elks

# network daemons to start
netstart="ftpd telnetd httpd"
QEMU=0

# specific daemon options
ftpd=
telnetd=
httpd=

# custom startup code
if $net != "" then      # network start specified in /bootopts
fi
```

Mellvik:
Thank you @ghaerr,
I've seen the proposed cnf file - I'll comment on that separately. This looks good.


> What I'm looking for is a way to avoid changing (say) /bin/net all the time.
> We can do it in /bin/net, but it doesn't taste good, does it?
> 
> What sorts of things are you having to modify /bin/net for? I'm trying to get a feel for how ELKS networking should ultimately be configured and be able to be configured. I agree that /bin/net should not be edited.
> 
The as-is situation mandates changing /bin/net in order to set or change options for the deamons. And like pointed out before, it's typically debug or admin stituations that require options. Like, right now QEMU=1 sets the ftpd debug option. Useful now but sometimes a real pain. So, setting/resetting a -d (or verbose) option for a daemon would be a typical  situation. Another one possibly coming would be logging, where do we log messages (instead of the console - some times I'd like to send output to tty2 instead of tty1). There are many such situations and more coming up as we start with ppp. 
> We don't have a /etc/inet.conf or the newer variants, but we need a way to 'configure' the daemons.
> 
> I suppose this is the same as the first question - what current and future options are you seeing the need for?
> 
> I honestly think we've done well in hiding the networking trickery of qemu in ftp and the QEMU env variable. After all qemu is a debugging environment.
> 
> Yes, you have done well with your QEMU hacks hidden within ftp and ftpd. Myself, I'm mostly interested in being able to quickly run ELKS on QEMU (and real hardware) and have most things just work (or when they don't, allow me to reproduce quickly). I also wish to expand the ELKS user base by making it easy to use out of the box (which we have). That said, I'm not trying to run a production environment like you are.
> 
This is a interesting observation. And yes, while I wouldn't use the term 'production environment' myself (you're welcome to), there is no question that we have different motivations - which is good. What I do think is the common denominator though, is the goal to make ELKS useful. When I got hooked on ELKS I was looking for a tool to help diagnose, test and fix a 286 klunker (another Compaq, but that's a different story). All the Tom's RTBTs and other single floppy alternatives were too big and required too much resources to be useful. ELKS on the other hand, had obvious potential: A working system on a floppy. And wow have we moved forward since. There's a long list og tools and functionality I'd like to see, when the platform is stable.
Maybe - in the process of turning the 0.5 corner, or possibly 0.6, we should create a (broad) target (or goal) for what ELKS is going to be when it grows up ( :-) )?
> Perhaps a seperate /etc/net.conf file makes sense for /bin/net options. I like /bin/net because its a shell script and easily updated. A single /etc/net.conf file may be advantageous to more and more /bootopts and/or /root/.profile options. If a /etc/net.conf file were actually a shell script itself (setting environment variables or other custom stuff), then /bin/net could source that file directly and things would be very easy. We might not have to write more code to parse files, just execute it as a shell script to set options within /bin/net and environment variables only visible during the /etc/rc.d/rc.sys execution.
> 
I definitely think this is the way to go. Also to offload bootopts. The ability to set addresses and stuff in bootopts has been very convenient, but it makes more sense to have the network related stuff in its own place. And I suspect that as we move on, there will be other, more boot-related options to make room for in bootopts. More comments about net.conf

Thank you.

--Mellvik

Mellvik:
Thank you @ghaerr,
yes, something like this - absolutely.

I do foresee the usefulness of being able to specify a different set of addresses for different link types - in order to be able to easily switch back and forth between ether and ppp or slip.

My host (the one connected to the ELKS machine) has an interface 'named' 10.0.2.2 already, so the slip interface sl0 would be (say) 10.0.3.3.
Maybe you have an idea about how to do that easily (instead of switching config-files, which is certainly doable).

This is great!

—M

ghaerr:
Hello @Mellvik,

> The as-is situation mandates changing /bin/net in order to set or change options for the deamons.
> So, setting/resetting a -d (or verbose) option for a daemon would be a typical  situation. 
> Another one possibly coming would be logging
> instead of the console - some times I'd like to send output to tty2 instead of tty1

I now understand what you're talking about - and the need to have much more configurable networking in ELKS.

> Also to offload bootopts. The ability to set addresses and stuff in bootopts has been very convenient, but it makes more sense to have the network related stuff in its own place. And I suspect that as we move on, there will be other, more boot-related options to make room for in bootopts.

Completely agree. We should move networking setup out of /bootopts and into its own configuration file. Also, /bootopts isn't implemented for FAT boots, so those users can't setup networking with the current design. /bootopts provided a starting point for passing environment variables, but these variables should not necessarily be in the long-term user environment for system networking services, as is the case for capitalized variables. Also, /bootopts is not a shell script, so we can't take advantage of custom-script programming to do special networking stuff.

> My host (the one connected to the ELKS machine) has an interface 'named' 10.0.2.2 already, so the slip interface sl0 would be (say) 10.0.3.3. Maybe you have an idea about how to do that easily (instead of switching config-files, which is certainly doable).

Yes - shell script functions. If the network configuration file is actually a shell script (which normally just sets temporary environment variables for /bin/net), then functionality can easily be expanded using BASH/ash shell script functions when necessary - without doing any C programming.

You have supplied very convincing arguments for enhancing our networking configuration. I have some ideas, very similar to the sample /etc/net.cnf above, and will post a PR for your thoughts shortly. This mechanism will be very similar and mostly compatible with what we use now, with some thoughts around expansion and flexibility. (Note, the name will probably need to end with a 3-letter extension, as the FAT filesystem still doesn't support creating long filenames, just opening them). Thus, /etc/net.cfg is possible a good name for it).

Thank you!

Mellvik:
Thank you @ghaerr,

this sounds really good. The name too (did I mention I hate FAT?).

Looking forward to it!

—M

PR #1110, opened 2022-01-17
[kernel] Fix pty buffer overread bug on signals

ghaerr:
Fixes problem reported in https://github.com/jbruchon/elks/commit/aac5f84e0000105416096445a93cea152b2d844e#commitcomment-63969722.

@Mellvik, it took fooooorrrever to find this, but finally knocked it out. Turns out the PTY driver was reading an extra character from the telnetd user buffer when a character that resulted in a signal (like ^C) was processed. That character was random, as it was off the end of telnetd's internal buffer, which could have anything in it.

### DIFF
--- a/elks/arch/i86/drivers/char/pty.c
+++ b/elks/arch/i86/drivers/char/pty.c
@@ -122,10 +122,9 @@ size_t pty_write (struct inode *inode, struct file *file, char *data, size_t len
 		}
 
 		ret = get_user_char ((void *)(data++));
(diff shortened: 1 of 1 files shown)


Mellvik:
Congrats - and thank you, @ghaerr.

That was indeed a tricky one!
When I woke up this morning, I was convinced that the 'random' character actually was the PID of the process being interrupted. Not a good sign - it's good I was wrong :-)

--M

Issue #1112, opened 2022-01-17
[sash] execl problem


Mellvik:
I'm preparing some minor fixes for `sash` for 0.5.0, and ran into a more serious issue:

In line 674 of `sash.c`,  `execl` is invoked to run `/bin/sh -c` with a (presumed) script as argument (after `execvp` failed, indicating that the target is most likely a script). 
Unfortunately, when called this way, `/bin/sh` has no environment and thus no PATH, so it does not find the target script unless a full path is specified. 
This is confusing and obviously erroneous - and probably an easy fix, but instead of a time consuming experimentation into what `exec`-call does what, I'm asking for advice.

Any suggestions, @ghaerr?

```
# net
net: No such file or directory 
# /bin/net
Usage: net [start|stop] [eth|slip|cslip] [baud] [device]
# 
```

Mellvik:
Correction, the line in question is 670:

```	
        execvp(argv[0], argv);

	/*
	 * If file has access but didn't execute, guess it's a shell script
	 * and run 'sh -c cmd'.
	 */
	if (errno == ENOEXEC)
		execl("/bin/sh", "sh", "-c", cmd, (char*)0);

	perror(argv[0]);	/* Usually 'No such file or directory'*/
	exit(1);
```

ghaerr:
Hello @Mellvik,

> Correction, the line in question is 670:

Strange. The execvp call is supposed to be using the PATH= environment var to find the passed program. It is returning ENOENT, instead of ENOEXEC, so /bin/sh isn't being used to execute `net`. This is handled in the C library rather than the kernel, i'll look into it.

Thank you!

PR #1113, opened 2022-01-17
[libc] Fix improper execvp handling of ENOEXEC and relative paths

ghaerr:
Fixes `sash` problem reported in #1112.

### DIFF
--- a/libc/system/execvp.c
+++ b/libc/system/execvp.c
@@ -126,7 +126,7 @@ execvp(char *fname, char **argv)
    int flen, plen;
    char * bp = sbrk(0);
 
-   if( *fname != '/' && (path = getenv("PATH")) != 0 )
+   if( !strchr(fname, '/') && (path = getenv("PATH")) != 0 )
    {
       flen = strlen(fname)+2;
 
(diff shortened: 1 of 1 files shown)


Mellvik:
@ghaerr,
I believe we're close to a world record. Most os/lib/sys related bugs fixed in < 24hrs. Elks shines.

Apropos stability - I'm regularly seeing elks process ids > 500, sometimes > 1000 these days. Wouldn't happen 3 months ago.

Let's continue to have fun. More active users are solicited!

--M

ghaerr:
> I believe we're close to a world record. Most os/lib/sys related bugs fixed in < 24hrs. Elks shines.

Thanks @Mellvik!

> Apropos stability - I'm regularly seeing elks process ids > 500, sometimes > 1000 these days. Wouldn't happen 3 months ago.

Wow, that's pretty cool. I take it you keep your Compaq Portable 386's running all the time?

Mellvik:
> I believe we're close to a world record. Most os/lib/sys related bugs fixed in < 24hrs. Elks shines.
> 
> Thanks @Mellvik!
> 
> Apropos stability - I'm regularly seeing elks process ids > 500, sometimes > 1000 these days. Wouldn't happen 3 months ago.
> 
> Wow, that's pretty cool. I take it you keep your Compaq Portable 386's running all the time?
> 
Actually, the one I use most of the time has been turned on more than 99% of the time for the past 3 years. It's a good thing it's in the garage, I don't see (or hear) it much at all. Only when changing floppies which happens only when there are big updates/merges. Serial (always 2 ports), now at 57600 baud, is my devoted friend.

-M
>

PR #1115, opened 2022-01-18
Add SCSI hard drive support for PC-98

tyama501:
- Add SCSI ID mapping and CHS parameters function for PC-98 in bioshd.c
- Modified int1B functions for SCSI hard drives

IBM style mbr (not PC-98 style) is needed for mounting partitions for now.
IDE hard drives are not supported.

Thank you!

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -72,6 +72,7 @@
 
 #ifdef CONFIG_ARCH_PC98
 #define MAXDRIVES	4	/* max floppy drives*/
+static unsigned char scsi_drive_map[7];
 #endif
 
 /* comment out following line for single-line drive info summary*/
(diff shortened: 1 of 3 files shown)


ghaerr:
Hello @tyama501,

Another excellent contribution, this is really great!

Does the PC-98 INT 1B AH=84h (BIOS_DRIVE_PARMS) function work for IDE drives, or only for SCSI drives? Just curious as to whether this is a special enumeration function or whether it could be used for IDE drives at some point.

Also, is there documentation on the (upcoming support for) PC-98 MBRs? I assume you're build a "fake" IBM PC MBR just to make the current code work, and that those MBRs won't boot on an actual PC-98, correct?

Thank you!

tyama501:
Hello @ghaerr ,

Thank you for merging it.

> Does the PC-98 INT 1B AH=84h (BIOS_DRIVE_PARMS) function work for IDE drives, or only for SCSI drives? 

I think INT 1B can be used for IDE drives on PC-9821 but device addresses need to change.
Maybe we need to find out the PC has SCSI drives or IDE drives (or both) prior to the function.

> Also, is there documentation on the (upcoming support for) PC-98 MBRs? I assume you're build a "fake" IBM PC MBR just to make the current code work, and that those MBRs won't boot on an actual PC-98, correct?  

Correct.
It won't boot. I have a short documentation about hard disk of PC-98.
It seems the whole cylinder 0 is special and it has an initial program loader, etc.

There is open source IPL and boot selector based on FreeBSD/pc98, I may reference it.
Even it cannot boot, to make dos partition of PC-98 style is worth to do for sharing files with dos system.

I also have some other things to try so let's discuss those on #1047.

Thank you.

PR #1116, opened 2022-01-18
[net] Add net.cfg network configuration file

ghaerr:
Adds new /etc/net.cfg network configuration file discussed in https://github.com/jbruchon/elks/pull/1105#issuecomment-1014894231.

Moves network settings out of /bootopts and /etc/profile (except for possibly QEMU= and net= for automatic network startup at boot).

All network configuration, including custom startup and shutdown code is in /etc/net.cfg. /bin/net should not have to be edited for machine configurations.

The new format allows each daemon's command line to be individually specified, for maximum flexibility and keeping extra environment variables to a minimum. In addition, specific daemons can be easily included or excluded at net startup time.

### DIFF
--- a/elkscmd/rootfs_template/bin/net
+++ b/elkscmd/rootfs_template/bin/net
@@ -12,18 +12,8 @@
 # See qemu.sh NET= line for host forwarding into ELKS using QEMU
 #
 
-# Default IP address, gateway and network mask set by env variables in
-# /bootopts or /etc/profile. They can be IP addresses or names in /etc/hosts.
-localip=$HOSTNAME
-gateway=$GATEWAY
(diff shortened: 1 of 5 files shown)


Mellvik:
Thank you, @ghaerr, great work.

I'll beat this up tomorrow or thursday. With this change I'll also remove the automatic debug activation in ftpd when qemu is set.

I suggest the comment on the QEMU= line in /bootopts be something like 'to use ftp/ftpd in qemu'.

--M

ghaerr:
> I suggest the comment on the QEMU= line in /bootopts be something like 'to use ftp/ftpd in qemu'.

Changed.

PR #1118, opened 2022-01-19
[boot] Add mount.cfg configuration file to mount filesystems at boot

ghaerr:
Edit /etc/mount.cfg instead of /etc/rc.d/rc.sys to control mounting and checking of filesystems at boot time.

/bootopts ignored message changed to indicate not functional on FAT boots.

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -456,7 +456,7 @@ static void probe_floppy(int target, struct hd_struct *hdp)
 	static char sector_probe[5] = { 8, 9, 15, 18, 36 };
 	static char track_probe[2] = { 40, 80 };
 #endif
-	int count;
+	int count, found_EPB = 0;
 
 	target &= MAXDRIVES - 1;
 
@@ -476,8 +476,9 @@ static void probe_floppy(int target, struct hd_struct *hdp)
 
 		if (drivep->cylinders != 0 && drivep->sectors != 0
 		    && drivep->heads != 0) {
-		    printk("fd: found valid ELKS disk parameters on /dev/fd%d "
-			   "boot sector\n", target);
+		    found_EPB = 1;
+		    /*printk("fd: found valid ELKS disk parameters on /dev/fd%d "
+			   "boot sector\n", target);*/
 		    goto got_geom;
 		}
 	    }
@@ -522,12 +523,12 @@ static void probe_floppy(int target, struct hd_struct *hdp)
 	if (drivep->cylinders == 0 || drivep->sectors == 0) {
 	    *drivep = fd_types[drivep->fdtype];
 	    printk("fd: Floppy drive autoprobe failed!\n");
-	}
-	else
-	    printk("fd: /dev/fd%d probably has %d sectors, %d heads, and "
-		   "%d cylinders\n",
-		   target, drivep->sectors, drivep->heads, drivep->cylinders);
+	} else {
+	    printk("fd: /dev/fd%d %s has %d cylinders, %d heads, and %d sectors\n",
+		   target, found_EPB? "found ELKS parm block,": "probably",
+		   drivep->cylinders, drivep->heads, drivep->sectors);
 
+	}
 	hdp->start_sect = 0;
 	hdp->nr_sects = ((sector_t)(drivep->sectors * drivep->heads))
 				* ((sector_t)drivep->cylinders);
--- a/elks/fs/minix/inode.c
+++ b/elks/fs/minix/inode.c
@@ -107,8 +107,9 @@ static struct super_operations minix_sops = {
 static void minix_mount_warning(register struct super_block *sb, char *prefix)
 {
 	if ((sb->u.minix_sb.s_mount_state & (MINIX_VALID_FS|MINIX_ERROR_FS)) != MINIX_VALID_FS)
-		printk("MINIX-fs: %smounting %s, running fsck is recommended.\n", prefix,
-	     !(sb->u.minix_sb.s_mount_state & MINIX_VALID_FS) ? "unchecked file system" : "file system with errors");
+		printk("MINIX-fs: %smounting %s 0x%x, running fsck is recommended.\n", prefix,
+	     !(sb->u.minix_sb.s_mount_state & MINIX_VALID_FS) ?
+		 "unchecked file system" : "file system with errors", sb->s_dev);
 }
 
 
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -120,7 +120,7 @@ void INITPROC kernel_init(void)
 #ifdef CONFIG_BOOTOPTS
     if (opts)
 	finalize_options();
-    else printk("/bootopts IGNORED: header not ## or size > %d\n", OPTSEGSZ-1);
+    else printk("/bootopts ignored: header not ##, size > %d or FAT boot\n", OPTSEGSZ-1);
 #endif
 
     mm_stat(base, end);
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -181,7 +181,7 @@ bc/bc							:other					:1440k
 #prems/pres/pres				:other
 #nano/nano-2.0.6/src/nano		:other					:1440k
 #nano-X/bin/nxclock				:nanox					:1440k
-nano-X/bin/nxdemo				:nanox					:1440k
+#nano-X/bin/nxdemo				:nanox					:1440k
 nano-X/bin/nxlandmine			:nanox					:1440k
 nano-X/bin/nxterm				:nanox
 nano-X/bin/nxworld				:nanox					:1440k
--- a/elkscmd/disk_utils/fsck.c
+++ b/elkscmd/disk_utils/fsck.c
@@ -84,7 +84,7 @@
 #define BITS_PER_BLOCK (BLOCK_SIZE<<3)
 
 static char * program_name = "fsck.minix";
-static char * program_version = "1.0.1 - 09/01/20";
+/*static char * program_version = "1.0.1 - 09/01/20";*/
 static char * device_name = NULL;
 static int IN;
 static int repair=0, automatic=0, verbose=0, list=0, show=0, warn_mode=0,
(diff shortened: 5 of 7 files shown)


Mellvik:
Thank you, @ghaerr.
This is great. More logical, more flexible I think.

I've wanted to comment on the use of $ROOTDEV before, so this is a good time: I find the name confusing since the function is supposed to check whatever fs we pass to it, not (only) a rootdevice. What about a different (more generic) name?

Related question: is it possible to mount the rootfs, say /dev/fd0, readonly, check and fix it, then remount rw?

And finally (this must be on many people's whishlist): there must be some open source FATfs checkers out there, that would be a great addition to elks. I'm finding myself rebooting to dos to run scandisk about twice a week.

-M

ghaerr:
> I've wanted to comment on the use of $ROOTDEV before, so this is a good time: I find the name confusing since the function is supposed to check whatever fs we pass to it, not (only) a rootdevice. What about a different (more generic) name?

I'll change it so the function takes an argument, but the check_filesystem function doesn't need to unmount and remount the filesystem if it's not a root device, thus comparison with $ROOTDEV (and we don't know whether a device is mounted or not). I'll look into seeing how this can be made more generic for both cases.

> Related question: is it possible to mount the rootfs, say /dev/fd0, readonly, check and fix it, then remount rw?

It should - set 'ro' in /bootopts, then `umount /; fsck /dev/fd0; mount -o remount,rw` just like is done in the `check_filesystem` shell script function. In other words, after booting a readonly filesystem, just running 'check_filesystem' should work.

> And finally (this must be on many people's whishlist): there must be some open source FATfs checkers out there, that would be a great addition to elks.

I'll see what I can find out there.

ghaerr:
> I've wanted to comment on the use of $ROOTDEV before, so this is a good time: I find the name confusing since the function is supposed to check whatever fs we pass to it, not (only) a rootdevice.

Fixed.

> Related question: is it possible to mount the rootfs, say /dev/fd0, readonly, check and fix it, then remount rw?

Yes, it now all works automatically, if you uncomment the check_mounted_filesystem and mount lines in mount.cfg.

For now, the filesystem checks are commented out and must be manually edited in. This is for two reasons: first, we don't yet have a way to determine what the filesystem type is on the drive, and running fsck on FAT would make a mess. Secondly, we don't yet have a way to fsck FAT disks. I've added both to my todo list.

The last commit additionally changed the ELKS floppy probe printk to a single line, rather than two lines, and cleaned up the console display for easier understanding of what's happening when filesystem checks are being performed.

Mellvik:
Thank you @ghaerr,

I admit I've pretty much stayed away from fsck this far and built new floppies when the minix fs became unpredictable. That's about to change - now that the boot file system may be checked - as I understand it.

This one is a little late, but I need to ask: Did you consider (venerable) /etc/fstab as an alternative?

--M

ghaerr:
> That's about to change - now that the boot file system may be checked - as I understand it.

The "mount -o remount,rw" feature that allows the boot/root filesystem to be remounted read/write or read-only has been around for some time. This PR basically just moved mount and check-related shell scripting out of /etc/rc.d/rc.sys into a new file, which is sourced by rc.sys, very similar to the way /bin/net sources net.cfg. The idea was to move commonly-changed configuration data into its own file and out of rc.sys, to clean things up a bit before v0.5.

> Did you consider (venerable) /etc/fstab as an alternative?

No, /etc/fstab isn't a shell script and would require large modifications to mount and umount. At this point, most of ELKS post-boot configuration are shell scripts, which allow things to stay very simple, yet provide for individual complexity without writing C code. Of course, all this could be changed, but I don't really see the point in it.

Speaking of which, the extra directory in /etc/rc.d/rc.sys is also overkill, introduces another needless redirection, and slows down boot. Should things not stay simple and just have it be /etc/rc.sys?

Mellvik:
> 
> That's about to change - now that the boot file system may be checked - as I understand it.
> 
> The "mount -o remount,rw" feature that allows the boot/root filesystem to be remounted read/write or read-only has been around for some time. This PR basically just moved mount and check-related shell scripting out of /etc/rc.d/rc.sys into a new file, which is sourced by rc.sys, very similar to the way /bin/net sources net.cfg. The idea was to move commonly-changed configuration data into its own file and out of rc.sys, to clean things up a bit before v0.5.
> 
> 
Great - my using it is indeed overdue.
> Did you consider (venerable) /etc/fstab as an alternative?
> 
> No, /etc/fstab isn't a shell script and would require large modifications to mount and umount. At this point, most of ELKS post-boot configuration are shell scripts, which allow things to stay very simple, yet provide for individual complexity without writing C code. Of course, all this could be changed, but I don't really see the point in it.
> 
OK, I'm perfectly happy with the proposed solution. However I do think fstab is simpler for most users, just a simple table to edit and no scripts to understand. And yes, some additional code in fsck and mount. That said, and ELKS system will not have 10 file systems to worry about, which make 'fsck -a' and 'mount -a' such a breeze aeons ago. Anyway - bygones. 
> Speaking of which, the extra directory in /etc/rc.d/rc.sys is also overkill, introduces another needless redirection, and slows down boot. Should things not stay simple and just have it be /etc/rc.sys?
> 
> 
Absolutely agree. To me the extra rc.d brings up painful memories about Unix System III :-)


Thank you!

-M

ghaerr:
> which make 'fsck -a' and 'mount -a' such a breeze aeons ago

Something like `mount -a` could come in very handy, except we don't know the mount points. We could implement a scheme in mount.cfg and a minor mod in mount to make that happen though. If that configuration info were kept in mount.cfg, but not mounted at boot, the user could simply type "mount -a" when needed, without having, say, to ensure both floppies are inserted at boot time.

In playing around with all this yesterday, I ran into a final problem which prompted me to stop work on writing a program to test whether a block device has a MINIX or FAT filesystem on it: when a block device (e.g. /dev/fd1) is merely opened by a user program, but no floppy inserted, we get no fewer than about 8 kernel probe, sector retry, and finally hard I/O error messages on the console. I realized then we've got lots to do to more effectively handle all this.

> To me the extra rc.d brings up painful memories about Unix System III :-)

I'll remove it.

ghaerr:
> And finally (this must be on many people's whishlist): there must be some open source FATfs checkers out there, that would be a great addition to elks.

The [dosfstools](https://github.com/dosfstools/dosfstools) project is the source of the current Linux `mkfs.fat` and `mkfs.vfat` filesystem checkers. This seems to be the simplest but most comprehensive out there, and might be able to be trimmed down to port to ELKS. Unfortunately, `mtools`, which is used for FAT mkfs in ELKS, and of which an early version is already included in ELKS applications, doesn't have a checker.

Mellvik:
> which make 'fsck -a' and 'mount -a' such a breeze aeons ago
> 
> Something like mount -a could come in very handy, except we don't know the mount points. We could implement a scheme in mount.cfg and a minor mod in mount to make that happen though. If that configuration info were kept in mount.cfg, but not mounted at boot, the user could simply type "mount -a" when needed, without having, say, to ensure both floppies are inserted at boot time.
> 
Agreed. Whether the mechanism is one or the other, we're very much one the same page as to the goal - simplicity.
> In playing around with all this yesterday, I ran into a final problem which prompted me to stop work on writing a program to test whether a block device has a MINIX or FAT filesystem on it: when a block device (e.g. /dev/fd1) is merely opened by a user program, but no floppy inserted, we get no fewer than about 8 kernel probe, sector retry, and finally hard I/O error messages on the console. I realized then we've got lots to do to more effectively handle all this.
> 
Shucks. Indeed. That's useful. Linux used to be reasonably smart about that, but then again, it may be that we just visit the log files rarely, there is a lot of verboseness - which is very useful if it doesn't hit our face all the time.
> 
--M

toncho11:
There should be a message that says that mount.cfg is ignored on fat. 
For example I tried it and I wondered why it did not work.

ghaerr:
Hello @toncho11,

> There should be a message that says that mount.cfg is ignored on fat.

It does work on FAT, I tried it over here (on QEMU) and it is working. I uncommented this line to mount /dev/fd1 as FAT at boot from a FAT boot on /dev/fd0:
```
mount -t msdos /dev/fd1 /mnt || true
```

> For example I tried it and I wondered why it did not work.

Can you be more descriptive, what exactly did you do and what was result displayed?

Thank you!

toncho11:
Sorry, I was missing the " || true".
Now it works.
So both "fat" and "msdos" are possible.

ghaerr:
> Sorry, I was missing the " || true".

The " || true" is only required to allowing running the mount script without having a floppy inserted in drive B. The mount command should work without that, I just tested over here and it works without it.

> So both "fat" and "msdos" are possible.

Yes, the mount -t allowed options are fat, msdos, minix and romfs. Unfortunately, one must still know the type of filesystem before attempting to mount it, which means /etc/mount.cfg must be edited when changing filesystem types. I would like to fix that in v0.6.

toncho11:
I insist that this made a difference for me. 

`mount -t fat /dev/hda2 /mnt || true`
It is fat 32.

ghaerr:
> mount -t fat /dev/hda2 /mnt

If the above doesn't work, that means that mount is returning a non-zero exit value. I looked at mount.c and if the mount fails, it should say "mount failed" and return 1. Does it say that? Perhaps try 
```
mount -t fat /dev/hda2 /mnt || echo $?
```
to show the mount exit value. It is possible I suppose that the FAT32 mount could "fail" with a bad return code, yet still work, since mount.c doesn't do anything else other than return a 1 or 0 exit code after mount fail/succeed.

toncho11:
I am not using the mount.cfg, but testing from the console. 

I did:
`mount -t fat /dev/hda2 /mnt || echo $?`

but I do not see the return code? The mount does not fail and it does not say "mount failed".

ghaerr:
> but I do not see the return code? The mount does not fail and it does not say "mount failed".

When using the || shell command, the latter command will not execute unless the former fails. Thus, in your case, the reason why you don't see either "mount failed" nor the exit code is that the mount command worked, *without* the "||". The "||" is required in shell scripts in order to keep executing on failure, which is why it has been added in mount.cfg - to allow the rc.sys script to continue on mount failure (usually disk not present).

toncho11:
Actually you are right. 
I do not know ... maybe I copied the "true" command to /bin or something like that at the same time I changed the file ...

Thanks!

PR #1120, opened 2022-01-20
Add conio-pc98-asm.h

tyama501:
Hello @ghaerr ,

I added conio-pc98-asm.h as you mentioned before.

early_putchar is duplicated with the definition in kernel.h.

Is it ok?

### DIFF
--- a/elks/arch/i86/drivers/char/conio-pc98-asm.h
+++ b/elks/arch/i86/drivers/char/conio-pc98-asm.h
@@ -1,13 +1,34 @@
 /* conio functions for PC-98 */
 
+/* put character on text vram */
 void early_putchar(int c);
-int bios_getchar();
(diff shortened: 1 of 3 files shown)


ghaerr:
> early_putchar is duplicated with the definition in kernel.h.

Thanks for letting me know. It's ok for now, I will remove the kernel.h definition when early_putchar is (finally) moved out of crt0.S.

The function definitions in conio-pc98-asm.h that take no parameters - can you use (void) for them, to avoid ambiguity with early C compilers where the arguments aren't specified? Also, can you add a single comment to the right (or above for multi-line, as in conio.h) that describes the procedure's functionality? That will help others understand what they do.

Thank you!

tyama501:
Hello @ghaerr ,

Added void and comments.

Thank you!

Issue #1114, opened 2022-01-18
[net, slip] zero window stops transmission


Mellvik:
OK; so in preparation for the coming `ppp` implementation, some `slip`-testing. 

Admittedly, I haven't used `slip` for ages, and it took some time to get it operational against Linux (raspbian) at the other end. I'll add the setup to the Wiki. Seems like an MTU of ~ 600 is about right - at 9600bps.

Anyway, it seems we're hitting some of the problems we've seen (and fixed) with `ether` before. When doing a `ftp` file transfer into ELKS (ELKS being the originating client, issuing a `get` command, but I don't think that matters), transmission hangs when the `ktcp` advertised window becomes 0, as the trace below shows. At this point, `ktcp` debug says
`tcp: dropping packet, bad seqno: need 4381 got 4380 size 0` (repeated for every retransmit from the other end).
At this point, `ktcp` has already acked 4381 and there are no outstanding packets. The other end continues to ack #1, which is correct. 

@ghaerr - unless you have an immediate clue on this one, I'm going to dive into it later this week.

```
    10.0.3.3.ftp > 10.0.3.15.1025: Flags [P.], cksum 0xebf7 (correct), seq 734:799, ack 86, win 11200, length 65: FTP, length: 65
	150 Opening BINARY mode data connection for 'vi' (69008 bytes).
16:58:45.323117 IP (tos 0x8, ttl 64, id 17808, offset 0, flags [DF], proto TCP (6), length 600)
    10.0.3.3.35131 > 10.0.3.15.1027: Flags [.], cksum 0x1978 (correct), seq 1:561, ack 1, win 11200, length 560
16:58:45.432649 IP (tos 0x8, ttl 64, id 17809, offset 0, flags [DF], proto TCP (6), length 600)
    10.0.3.3.35131 > 10.0.3.15.1027: Flags [.], cksum 0x5d64 (correct), seq 561:1121, ack 1, win 11200, length 560
16:58:45.483641 IP (tos 0x0, ttl 64, id 283, offset 0, flags [none], proto TCP (6), length 40)
    10.0.3.15.1025 > 10.0.3.3.ftp: Flags [.], cksum 0x8170 (correct), ack 799, win 4315, length 0
16:58:45.534385 IP (tos 0x0, ttl 64, id 284, offset 0, flags [none], proto TCP (6), length 40)
[...]

PR #1122, opened 2022-01-22
[net, ftp] Fixes to handle slip problem + bugfixes

Mellvik:
One big, many small fixes to `ftp` - and optimized stack/heap values for the Makefile (for `ftp`and `ftpd`).
 
The fix to `ftp` to handle zero length files (#1104) turned out to break most file transfers when using `slip` instead of ethernet (#1114). This update eliminates that problem by changing the way the `ftp`-client handles replies from the server (no more buffered IO). 

Also in this update:
- The call to `system()`  now handles error returns properly (reports errors)
- Closes the command connection properly when login fails
- Option processing was ignoring every other option ....

There are still many extra debug printfs and commented out code, a cleanup will come later when these changes have been properly tested on QEMU (my testing has been HW only).

### DIFF
--- a/elkscmd/inet/ftp/Makefile
+++ b/elkscmd/inet/ftp/Makefile
@@ -19,10 +19,10 @@ OBJS = $(SRCS:.c=.o)
 all:	$(BINS)
 
 ftp:	ftp.o
-	$(LD) $(LDFLAGS) -maout-stack=8192 -o ftp ftp.o $(LDLIBS)
+	$(LD) $(LDFLAGS) -maout-stack=6144 -maout-heap=4096 -o ftp ftp.o $(LDLIBS)
 
 ftpd:	ftpd.o
-	$(LD) $(LDFLAGS) -maout-stack=8192 -o ftpd ftpd.o $(LDLIBS)
+	$(LD) $(LDFLAGS) -maout-stack=4096 -maout-heap=4096 -o ftpd ftpd.o $(LDLIBS)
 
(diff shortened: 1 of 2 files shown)


ghaerr:
Thanks for the further testing and bug fixes Helge.

I have some concerns about line 632 in ftp.c where the select timeout is doubled each loop. This will quickly overflow and turn to 0, which will turn to straight polling with no wait. Is that the intended behavior?

Mellvik:
Thank you, @ghaerr,
No, that's not the intention - good catch. I didn't see that one. The doubling of the timeout at every interation was/is experimental pending some qemu testing. Some times (when running slip @ 19200), the loop runs 8 to 10 times, which is no big deal, but the doubling reduces it to 3. Running ethernet there is never more than one. Also pending is testing on even slower hw (read: 286). 
Like the comment a few lines further down says, the loop needs a bailout point.

-M

ghaerr:
>  Some times (when running slip @ 19200), the loop runs 8 to 10 times, which is no big deal, but the doubling reduces it to 3. Running ethernet there is never more than one. Also pending is testing on even slower hw (read: 286).

Rather than getting fancy, I suggest just setting a minimum timer, (perhaps 50ms as you have it), and just letting ftp timeout and loop without changing the timeout. Trying anything fancier just risks errors (as is the case now), and a few extra loops in select code won't ever be noticed, and will always act properly. (This is exactly what telnetd does now).

PR #1125, opened 2022-01-26
Supported multiple bytes read for TRACK_CACHE in bios1B for PC-98

tyama501:
Hello @ghaerr ,

I could boot FD1232 and FD1440 with TRACK_CACHE and No TRACK_CACHE on the emulator.
The single bios1B code without ifdef TRACK_CACHE worked.

I also tested FD1232 disk with PC-9801RX and I could feel the effect of TRACK_CACHE using "more" command.
Very nice to see the disk access decrease.

Thank you for your support for this work.

### DIFF
--- a/elks/arch/i86/lib/bios1B.S
+++ b/elks/arch/i86/lib/bios1B.S
@@ -59,7 +59,6 @@ call_bios:
 	xchg %ch,%dl     // sector number for PC_98
 	xchg %bx,%bp
 
-
 	push %ax
 	mov %ch,%al
 	and $0x80,%al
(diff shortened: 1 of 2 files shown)


ghaerr:
Glad to see this is working!

It seems that this version will only work with a single drive type (both drives the same), FD1440 or FD1232, correct? If so, a later version, combined with perhaps a C wrapper can look at drivep->sector_size and then there will not have to be an ifdef in the bios1B.S code.

Thank you!

tyama501:

Correct.

And I think we need some mechanism to select device address when probing floppy disk after it is inserted to support the both drive types.
It is different from harddisk since it can remove.

Let me discuss that again when we try it.

Thank you!

Issue #1127, opened 2022-01-31
Failed to build elks on arm device


hex-4d5a9000:
I was trying to build elks on my arm64 Android phone and I was following the build instruction. I installed Ubuntu using Linux Deploy. `GCC-IA16` has been built successfully(with a little modification to the build script) and `menuconfig` also works. However I got into trouble at the 5th step, when building the elks itself, where gcc and make printed the following error message: 
```
In file included from _libelf_config.h:167:0,
                 from _libelf.h:34,
                 from elf.c:29:
/home/android/elks/elks/tools/elf2elks/lib/elftoolchain-0.7.1/common/native-elf-format.h:3:20: error: 'unknown' undeclared here (not in a function)
 #define ELFTC_ARCH unknown
                    ^
_libelf_config.h:170:22: note: in expansion of macro 'ELFTC_ARCH'
 #define LIBELF_ARCH  ELFTC_ARCH
                      ^~~~~~~~~~
elf.c:34:18: note: in expansion of macro 'LIBELF_ARCH'
  .libelf_arch  = LIBELF_ARCH,
                  ^~~~~~~~~~~
make[3]: *** [lib/elftoolchain-0.7.1/libelf/libelf.a] Error 1
make[2]: *** [all] Error 2
make[1]: *** [tools] Error 2
make: *** [all] Error 2
```
This seems to be a problem with elf2elks. A script names `native-elf-format` is used to generate `format.h` but there's no arm-related declaration. How do I resolve this issue?

Btw: the `README` file says that `elksemu` only runs on linux-i386 or similar. Would it be helpful if I cross-compile elksemu to i386 then use qemu-user-i386 to run it?

ghaerr:
Hello @GNUfan42,

> the README file says that elksemu only runs on linux-i386 or similar. Would it be helpful if I cross-compile elksemu to i386 then use qemu-user-i386 to run it?

Yes. `elksemu` requires Linux 386, but ELKS itself will only run on an x86-compatible processor. So you'd need to run qemu-user-i386 to run and boot ELKS, after compiling it using the cross-compiled ELKS toolchain. While ELKS is quite a bit more portable than it has been over the past two years, many aspects of it are very dependent on a segmented 16-bit architecture.

> I was trying to build elks on my arm64 Android phone and I was following the build instruction.
> This seems to be a problem with elf2elks. 

[...]

PR #1128, opened 2022-01-31
[tools] elf2elks: patch elftoolchain to allow AArch64 host system

tkchia:
See https://github.com/jbruchon/elks/issues/1127 .


ghaerr:
Hello @tkchia, thank you!

PR #1133, opened 2022-02-02
Add CONFIG_TRACK_CACHE=y and CONFIG_BLK_DEV_BHD=y in pc98.config

tyama501:
Hello @ghaerr ,

I modified pc98.config to enable track cache and hard disk for the release of 0.5.0.

Thank you.


ghaerr:
Looks good!

We were talking about possibly extending CONFIG_BLK_DEV_BFD0_xxx to add CONFIG_BLK_DEV_BFD0_1232. I'll commit this now, does the current PC-98 code ignore all of these BFD0_xxx values, and just assume all disks are either FD1232 or FD1440, according to the CONFIG_IMG_xxx setting? If so, this might want to be changed to use these BFD0_ values. This can also wait until after 0.5.0 if desired.

Thank you!

tyama501:
Thank you for the merging @ghaerr ,

Current PC-98 ignores these CONFIG_BLK_DEV_BFD0_xxx values.
Also I couldn't find any of these value in bioshd.c.
Where are they used for IBM?

I think 0.5.0 is AS IS and
I will think about probing FD drives after 0.5.0.

ghaerr:
> Also I couldn't find any of these value in bioshd.c.
> Where are they used for IBM?

It appears they are not used... I was thinking perhaps they would be useful for PC-98 with no disk probing.

> I will think about probing FD drives after 0.5.0.

Ok. In that case, I will likely remove the BFD0_xxx config options, as they are not useful for IBM PC, almost all IBM PC except extremely early BIOS support INT 13 BIOSHD_DRIVE_PARMS (AH=08h) function.

> I think 0.5.0 is AS IS 

No problem, thank you!

PR #1140, opened 2022-02-06
[tools] elf2elks: fix: open ELF input in binary mode on MS-DOS

tkchia:
Back in March 2021, I disabled support for the ELKS target, in the DJGPP/MS-DOS hosted version of the `gcc-ia16` toolchain &mdash; partly because `elf2elks` was not working properly under DJGPP.  I am glad to say that I found out why it was not working. :slightly_smiling_face:

### DIFF
--- a/elks/tools/elf2elks/elf2elks.c
+++ b/elks/tools/elf2elks/elf2elks.c
@@ -30,6 +30,8 @@
 #include <sys/stat.h>
 #ifndef __DJGPP__
 # include <libgen.h>
+#else
+# include <io.h>
 #endif
 #include "libelf.h"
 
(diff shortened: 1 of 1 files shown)


toncho11:
So one can build ELKS on Windows?

ghaerr:
Thank you @tkchia!

tkchia:
Thanks @ghaerr !

Hello @toncho11,

> So one can build ELKS on Windows?

Not exactly.  A while back I added &mdash; and later removed &mdash; support for building ELKS _applications_ using an MS-DOS-hosted `gcc-ia16` compiler toolchain.  This was only for building applications, not the whole system.

If there is demand for such a thing, I might add this support back.

Thank you!

PR #1143, opened 2022-02-08
Increment version to v0.5.0

ghaerr:
ELKS version 0.5.0 at last!

Draft release doc and assets at https://github.com/jbruchon/elks/releases/edit/untagged-cec6249943228280b1e6.


jbruchon:
Excellent work, everyone! I'm so glad to see that things have come so far. You're all amazing. Thanks to you all!

Mellvik:
Thank you very much, @ghaerr - this is in fact incredible. The release notes says it all. 

It's so easy to forget all the mountains climbed as we jump to new challenges all the time. 

As a heavy ELKS-on-physical user, it's a pleasure to acknowledge that the system is now in great shape and useful/Useable for practical work. As it turns out, the addition of XMS support makes a helluva difference, not the least since I'm running off of floppies all the time. Thanks for making it all possible. 


I'll update the Wikipedia entry later this week unless someone beats me to it.

-M

ghaerr:
Thanks @Mellvik, and thank you for all you've done for ELKS.

I'm glad you're enjoying XMS - I have a funny story about "large" address spaces - at first, I had envisioned we'd keep a full 1.44M floppy in system buffers (seemingly easily achievable with 2M extra RAM in XMS), which would allow for "super speedy" ELKS execution. However, each system memory buffer still needs approx 26 bytes of header information, that must reside in the kernel data segment. In order to buffer 1440K of disk, one would need 26 \* 1440 = 37,440 bytes of memory just for the headers, over half of the kernel data segment address space! With large address spaces, things like this don't mean anything, but on 16-bits, everything matters.

Mellvik:
A fascinating  thought indeed, @ghaerr.

It's interesting - and inspiring - how limited resources forces us to think differently, be smarter etc.

Back to all this memory, maybe we could adjust the ramdisk driver to use all this memory? It will take some time to transfer the entire floppy but now that ELKS is stable enough to run for days and weeks, that could be really useful. 

Actually - and apropos ramdisk, I have so many times beaten up myself for noe installing a ramdisk just til get /lib/liberror off of the floppy. Every time there is perror(), there is also a 3 second or more delay to get to the file etc etc.
So, yes I like the idea of taking advantage of more ram …

—M

PR #1146, opened 2022-02-11
[kernel] Add wrappers around struct buffer_head member access

ghaerr:
First pass at cleaning up filesystem code to use \#defined wrappers around access to `struct buffer_head` members.

This will be used in next step of design to allow up to 2048 buffer headers and place them in main memory, rather than the kernel data segment, which would take up too much space. A far pointer will be used to access less-used members of the buffer header, keeping near pointers for access to data for filesystem speed and not having to rewrite the filesystem drivers.

This in turn will allow ELKS to access up to 2M of XMS memory, allowing very fast operation with the ability to keep system buffers for more than an entire 1.44M boot floppy.

No change to operation in this commit.


Mellvik:
Well, how 'bout that?
I just can't wait. This is going to eliminate - for most practical purposes - the difference between running physical and qemu for testing
Very much looking forward to the next episode, @ghaerr.

-m

ghaerr:
> This is going to eliminate - for most practical purposes - the difference between running physical and qemu for testing

Yes - I started to think about your request for speeding up `perror` by putting /lib/liberror.txt on a ramdisk, then started thinking this would probably be a better solution. BTW, I'm thinking of moving liberror.txt out of /lib over to /etc to eliminate another 1K disk read, as /etc/ is more likely to be in memory, and there's no great reason for having a separate /lib dir at this time.

Mellvik:
>  BTW, I'm thinking of moving liberror.txt out of /lib over to /etc to eliminate another 1K disk read, as /etc/ is more likely to be in memory, and there's no great reason for having a separate /lib dir at this time.

Supported. I'm fighting the urge to keep thing in their 'right places' :-) - a cause that was 'lost' to common sense a long time ago.

Thank you.
-M

PR #1147, opened 2022-02-12
Increase XMS system buffer memory up to 2.5M

ghaerr:
This PR (along with #1146) adds the capability of having a very large number of ELKS system buffers, stored in XMS (high, above 1M) memory, up to 2.5MiB total. 

These buffers act just like regular system buffers and are copied back and forth into main memory, and should greatly increase the speed of ELKS running on floppy systems, as now almost two full 1.4MM floppies can be completely held in RAM system buffers. Yes - after a bit, floppy based ELKS can now run like QEMU on real hardware!

The change is fully operational, but only tested on QEMU. It should run well on any system capable of 'unreal' mode (386+) or INT 15h/1Fh systems (Compaq Portable 386 and PC-98) with XMS BIOS extensions. To set the number of buffers, use the existing XMS buffer setting, up to and including 2500 (each buffer is 1K):
```
CONFIG_FS_NR_XMS_BUFFERS=2500
```
At this point, one must be sure there is in fact enough XMS memory present, as ELKS doesn't (yet) query for the amount of XMS aviailable.

The system works by splitting the "buffer head" structure into two parts, and keeps the majority of it in (far) main memory, which saves even more room in the kernel data segment for other important things. The memory used for the buffer heads will show up as "BUF" memory using `meminfo` (see below). Total XMS memory used is not (yet) shown by meminfo. For 2500 buffer heads, the total cost of main memory is 64992 bytes (and 2.5Mib of XMS memory, of course).

Screenshot showing boot with 2500 XMS buffers:
xms 2500 1

Running meminfo on system with 2500 XMS buffers and 2.5Mib XMS RAM, with lots of room to spare in main memory. (5000 bytes of kernel heap, and 64992 bytes of main memory required for implementation):
xms 2500 2


Mellvik:
Incredible, @ghaerr!
Can't Wait to test! Should work on my 286 too!

Are you saying I can set buf=2500 in /bootopts?
And is 2500 now a hard limit?

-M

Mellvik:
@ghaerr,

Fabulous! I keep repeating myself - and I do realize that I'm the primary (short term) 'beneficiary' of this improvement, running on hardware than can take advantage of it.

Anyway, I did manage to sneak in a few minutes of testing. This is looking very promising. I've only ran the new XMS stuff on the 286 (!) so far, and only for a few minutes, so there are more experiences to come.

I did run into one 'incident':
After having edited a few (config) files (using vi), I ran a sync - an old habit before a reboot, although it's not necessary. And I got the message below and a hard hang:

 Wrote "/etc/net.cfg"  46 lines, 972 characters 
# sync
ll_rw_blk: Trying to read nonexistent block-device C86A (3362652160)

I have not been able to reproduce it.
COme to think of it - when the above problem occurred, ELKS was booted with buf=2499 in /bootopts, menuconfig XMS set to 2500. Maybe that has something to do with it?
Don't ask why the odd number. I do weird things when testing. It has turned out very useful.

—M

ghaerr:
Hello @Mellvik,

> And is 2500 now a hard limit?

Yes. That's about how many extended buffer heads (size 26) as will fit in a 64K segment.

> This is looking very promising. I've only ran the new XMS stuff on the 286 (!) so far, and only for a few minutes, so there are more experiences to come.

Thanks for the testing!

> I ran a sync - an old habit before a reboot, although it's not necessary. And I got the message below and a hard hang:
> ll_rw_blk: Trying to read nonexistent block-device C86A (3362652160)

Hmmm, that's bad news - what happens once will undoubtably repeat again. It looks to me like the buffer head (which contains the device number (which should have been 0380 in this case) was overrun with other data. Since the buffer heads are now in main memory, we could have a kernel memory allocation problem, an application could be writing outside its boundaries, or the buffer system could be buggy..

With regards to trying to reproduce this bug - if you can remember or retry which commands you ran in which order, this might help, if an application is writing outside its boundaries unbeknownst to us, although the buffer headers are directly after the kernel data segment in main memory, and below any other applications.  Given how many buffers there are, I suppose we ought to write a regression tester to really put the system through its paces

It just occurred to me this enhancement brings another risk - that of not ever writing anything to disk until a sync occurs... and that could be quite a bit of data, (or data lost), so we might want to effect some automatic write-through syncing or have the idle task do it at preset intervals to prevent disaster.

This entire feature is able to be currently turned off by by configuring XMS off. I think it may be safe enough to commit it, and continue testing. We may want an additional option to turn on/off "far" buffer headers from make menuconfig, I'll add that after commit so that if buggy, the enhancement can be disabled if needed.

Thank you!

Mellvik:
> I ran a sync - an old habit before a reboot, although it's not necessary. And I got the message below and a hard hang:
> ll_rw_blk: Trying to read nonexistent block-device C86A (3362652160)
> 
> Hmmm, that's bad news - what happens once will undoubtably repeat again. It looks to me like the buffer head (which contains the device number (which should have been 0380 in this case) was overrun with other data. Since the buffer heads are now in main memory, we could have a kernel memory allocation problem, an application could be writing outside its boundaries, or the buffer system could be buggy..
> 
> With regards to trying to reproduce this bug - if you can remember or retry which commands you ran in which order, this might help, if an application is writing outside its boundaries unbeknownst to us, although the buffer headers are directly after the kernel data segment in main memory, and below any other applications. Given how many buffers there are, I suppose we ought to write a regression tester to really put the system through its paces
> 
I think I can do that - at least, it's worth a try. I have it all in the minicom scroll buffer, will take a look. IIRC just 3 vi-edits, then the sync (and the odd # of buffers of course). I'll be back on that one.
> It just occurred to me this enhancement brings another risk - that of not ever writing anything to disk until a sync occurs... and that could be quite a bit of data, (or data lost), so we might want to effect some automatic write-through syncing or have the idle task do it at preset intervals to prevent disaster.
> 
Yes, I've been thinking about that too. Good old update() and cron() will do it, but that will of course kill the performance while the flush is happening (to floppy). Then again, a flush every (say) 30 seconds is probably better than flushing single dirty blocks every few seconds.

More testing as soon as I get time (had some fun with buffering, doing a 'cat /bin/* > /dev/null' to get everything buffered, wow …).

 
Thanks!!


—M

ghaerr:
> had some fun with buffering, doing a 'cat /bin/* > /dev/null'

You know I had to try that!

And guess what? On QEMU, after doing that directly after boot (which only took 2 seconds), various commands, including vi, don't even run! So we have a repeatable problem.

I'm going to hold off commit while I look further into this.

Thank you!

Mellvik:
Wow - now I feel like a party killer :-)

FWIW - I did not run into problems with it.

Also, now that we're into 'utilities', maybe memento could have an option to display buffer usage, not necessarily status, but the current usage - used, free, dirty, 
Oooops, not only party killer, to-do-list killer too.

Thanks!


-M

ghaerr:
> Oooops, not only party killer, to-do-list killer too.

Well, when it comes to ELKS destroying data with a new feature, I'd rather the party be killed early :)

> could have an option to display buffer usage, not necessarily status, but the current usage - used, free, dirty,

I'm ahead of you on that one - originally thinking of a "buffers" command but could be in meminfo also. I'm already needing this information debugging the problem you've just found...

ghaerr:
> cat /bin/\* > /dev/null
> FWIW - I did not run into problems with it.

Just figured out this problem - I was running QEMU with only 2M emulated memory, so the XMS buffers were in non-existent memory. After changing qemu.sh to "-m 4M" I don't see any problems.

I think I'll work on some shell scripts to really exercise the buffer system.

ghaerr:
Hello @Mellvik,

> I have not been able to reproduce it.

Using shell scripts and running "cat /bin/\* > file1" multiple times and then running `cmp`, I found a problem, which could definitely be the one that caused your hang. The bug is that the buffer routine wasn't allocating enough memory for exactly the last far buffer head, which means it could be overwritten (by an application the kernel placed just after the buffers, for instance). Since the hang was definitely caused by buffer head being overwritten with improper data, this is likely what happened.

It's now fixed. In the meantime, I increased the buffer size in `cmp` from 512 to 2048 which greatly increases its speed.

Thank you for your testing!

Mellvik:
> In the meantime, I increased the buffer size in cmp from 512 to 2048 which greatly increases its speed.
> 
Ripple effect! I suggest the same cure for sum.c (which I'm using a lot).  Maybe even ls.c which is currently @ 1k ...
> 
Thank you!


--M

Mellvik:
Law of unintended consequences @ghaerr.

Here's an interesting one: With all this buffer space, what happens when copying a floppy image to a physical floppy?

- at first, the write seems to just hang.
- then, after hitting ^C, it seems to be working :-) - writing stuff to the floppy, possibly a lot of data depending on how long it was left alone before the ^C
- Then, after nightfall, and it's darker, I notice that the disk LED is blinking, sh.. - something is actually going on, it's not hung anyway...
- Which is about the time what's really going on is dawning on me - ELKS is now buffering the entire image before writing anything to floppy. I must be getting slow.

Anyway, this is not necessarily a good thing, and in traditional Linux/Unix systems, this is what we have raw devices for. Bypassing the buffer system. Should we reintroduce raw storage devices? 

BTW, reading a 1.2M image from disk to XMS buffers, takes quite some time. I haven't measured, seemed like forever, maybe a minute.

--M

ghaerr:
> With all this buffer space, what happens when copying a floppy image to a physical floppy?

What is the actual command you are executing? That matters, as different parts of the system are used depending on the program and device used.

> ELKS is now buffering the entire image before writing anything to floppy. 

Well, ELKS has a complex buffering scheme now, which XMS didn't change. There are internal L1 (cached) buffers, L2 (external or XMS) buffers, then at the lower level we have a DMASEG buffer, as well as a large disk track read (cache) buffer, all different.

> Should we reintroduce raw storage devices?

The system does implement raw block devices, which use the kernel buffer system. That's probably what you're using in this example. Adding character devices won't change much, as there's no place to store the data in the tiny 64k kernel we have. Remember, ALL ELKS I/O is performed in 1K chunks, period. This can't be changed without rewriting the BIOS floppy driver, and has lots of issues with older drives that don't/won't perform multi-track reads. Old hardware is a bitch, and ELKS has to take the lowest common denominator in these cases.

I think this issue is improved by turning on auto-sync, or perhaps having certain applications (dd?) perform sync more regularly.

> BTW, reading a 1.2M image from disk to XMS buffers, takes quite some time. I haven't measured, seemed like forever, maybe a minute.

The read time should not be affect much by XMS vs EXT buffers. 

The intent of any kernel system buffers is supposed to speed up LRU (least recently used) access, not speed up entire disk read/writes. For that, `dd` would usually allow larger buffers for speed, but without application-level memory mapping and direct-to-page I/O, that won't help much on ELKS.

In conclusion, I don't think the XMS enhancement changes much, other than the need for more regular syncing during large transfers.

Mellvik:
> What is the actual command you are executing? That matters, as different parts of the system are used depending on the program and device used.
> 
cat file /dev/fd1
and
dd if=file of=/dev/fd1
the latter some times with bs=4b
> ELKS is now buffering the entire image before writing anything to floppy.
> 
> Well, ELKS has a complex buffering scheme now, which XMS didn't change. There are internal L1 (cached) buffers, L2 (external or XMS) buffers, then at the lower level we have a DMASEG buffer, as well as a large disk track read (cache) buffer, all different.
> 
> Should we reintroduce raw storage devices?
> 
> The system does implement raw block devices, which use the kernel buffer system.
> 
OK, we need to clarify the terms here. In the (traditional) Unix (and I assume Linux) terminology, a device is either raw or block. For storage, a raw device bypasses the buffering system entirely, accessing the devices as directly as technically possible. Thus my question. 
> That's probably what you're using in this example. Adding character devices won't change much, as there's no place to store the data in the tiny 64k kernel we have. Remember, ALL ELKS I/O is performed in 1K chunks, period.
> 
A pure raw storage devices would read or write the smallest block the device supports, but using 1k would be fine for ELKS I suppose.
> This can't be changed without rewriting the BIOS floppy driver, and has lots of issues with older drives that don't/won't perform multi-track reads. Old hardware is a bitch, and ELKS has to take the lowest common denominator in these cases.
> 
I have a hard time seeing that any rewrite would be required anywhere. The raw device is just an access point to a lower level than the block device, using exactly the same code for low level IO.
> I think this issue is improved by turning on auto-sync, or perhaps having certain applications (dd?) perform sync more regularly.
> 
Since you mention it, I have a question about that. Actually the same old story :-): Now that I'm using it (SYNC=30) - and btw, this is a very valuable addition - the 'hangs' caused by sync are very noticeable, some times annoying (until I get used to it) because the system appears dead (i.e. no echo) for up to 5-6 seconds if I have a lot going on. So the question is whether the sync could be easily spread. Like, take a max number per operation in order to make the activity less disturbing - and spread the say, every second. (When I say same old story, I meant the ono-interrupt driven IO).
An entirely different approach would be to introduce the UNIX way of (immediate) character echoing, which - if the block IO lets interrupts get serviced at all - would keep the appearance of being alive up while sync is happening. This is not a complaint, simply attempted creativity.
> BTW, reading a 1.2M image from disk to XMS buffers, takes quite some time. I haven't measured, seemed like forever, maybe a minute.
> 
> The read time should not be affect much by XMS vs EXT buffers.
> 
OK, I didn't go back to the numbers we collected a while back - but I admit to being surprised by a read speed (albeit on FAT) of like 20k/s from hard disk.
> The intent of any kernel system buffers is supposed to speed up LRU (least recently used) access, not speed up entire disk read/writes. For that, dd would usually allow larger buffers for speed, but without application-level memory mapping and direct-to-page I/O, that won't help much on ELKS.
> 
Right. The XMS buffering makes a world of difference in practical usage. My point was to bring up the need for raw device access, not another IO speed discussion. We do have a lot of potential there, but what we have is good for now.


thank you.

-M

ghaerr:
> In the (traditional) Unix (and I assume Linux) terminology, a device is either raw or block.

Yep. And "raw" means a character device, vs block.

> A pure raw storage devices would read or write the smallest block the device supports, but using 1k would be fine for ELKS I suppose.

Yes. That's exactly what the current "block" device does - it always uses 1K blocks. As I mentioned above, we don't have the extra memory to allocate a larger memory-mapped buffer within dd, which would happen on 32-bit paged systems. That is, the character device would map the memory directly into the application user space, thereby eliminating any buffer copies entirely. This isn't possible in ELKS.

> The raw device is just an access point to a lower level than the block device, using exactly the same code for low level IO

Yes. We could do this, but there would be no speed gain. The block device is already doing almost the same thing - ALL I/O is already copied into a buffer, this would never be the application buffer (since IBM PC DMA architecture requires 1K block aligned buffers period), so the block (or character) device does I/O into a system buffer, which is then copied to the application.

> My point was to bring up the need for raw device access, not another IO speed discussion. 

My point is that ELKS is already doing the best it can in the existing special code written for when a block device is opened directly. Writing a seperate character (raw) version would have to do almost exactly the same thing, except it would have to use yet another buffer (internal to kernel?), certainly not the application), from which to do I/O in/out of.

Thank you!

ghaerr:
> So the question is whether the sync could be easily spread. Like, take a max number per operation in order to make the activity less disturbing - and spread the say, every second.

You mean change to a 1 second sync? Perhaps what we need, instead of a timer, is a count of buffers that need to be synced, and (also) start the sync when above that amount? (You can play with this in elks/init/main.c, where the idle task is). 

I can add code to have it count buffers... how many buffers do you think it should wait for, that is, how long is too long when it comes to (slow) floppy I/O? :)

> (When I say same old story, I meant the ono-interrupt driven IO).
An entirely different approach would be to introduce the UNIX way of (immediate) character echoing, which - if the block IO lets interrupts get serviced at all - would keep the appearance of being alive up while sync is happening.

The keyboard is interrupt driven, as is the serial input, and even in UNIX, it uses full-duplex meaning the shell has to "echo" the character (if in raw mode like ash, or do something else, e.g. arrow keys). If in sash, the console echo is ON, which means the character is "echoed" by placing the character in the output queue and output if no other I/O waiting).

I think the problem here is that the kernel itself doesn't run to even receive the interrupt keyboard character, until the BIOS disk read function returns. You talked about this earlier - there is nothing that can be done about this until a non-BIOS floppy driver is written.

Mellvik:
> The raw device is just an access point to a lower level than the block device, using exactly the same code for low level IO
> 
> Yes. We could do this, but there would be no speed gain.
> 
The unix/linux raw devices were never there for speed. The point is bypassing the buffer system in the kernel. Which is my point too. 
> The block device is already doing almost the same thing -
> 
No, it's not and that's my point. Block IO is via the buffer system which the raw device is bypassing.
> My point is that ELKS is already doing the best it can in the existing special code written for when a block device is opened directly.
> 
I thought so. What's missing is access to it via a device. No writing anything new. this is what a raw device is. A hook into the IO level just above the device. Not fast, not fancy, just raw. The goal in my particular example isn't to speed things up, but to bypass the buffering system. If it ends up speeding up something, that would be good although unexpected. That said, what we have now is 60 seconds to fill a buffer, then writing. things being optional the raw read/write may be a memory copy shorter.
> 
--M

ghaerr:
I get your point now about the raw device bypassing the buffer system. What exactly is the benefit of that, if not for speed?
In the ELKS case, I would say we'd want the floppy contents to be available in XMS, to reduce the chance of another very slow disk read.

> No writing anything new. this is what a raw device is. A hook into the IO level just above the device.

ELKS doesn't have any lower level I/O than the block device driver. All I/O is within that (bioshd.c). So we would have to duplicate (or split off) all the disk I/O that is currently only in bioshd.c, just to implement a raw device. The disk interface in ELKS is only from the block device driver. Perhaps that's why there are no character (raw) disk devices in ELKS...

> things being optional the raw read/write may be a memory copy shorter.

As I mentioned above, any raw read/write cannot be directly to the application due to DMA segment wrap issues, thus there will always be one memory copy - exactly what I rewrote the block device system to do about a year ago. All ELKS block I/O is now to a system buffer, which then gets copied exactly once to an application buffer. The ethernet drivers (a character device) were rewritten to directly write into the calling application's buffer, thus bypassing system buffers, but we can't use BIOS disk I/O to do the same, even with a character (raw) device, because of the PC DMA problem on some BIOSes.

> I can add code to have it count buffers... how many buffers do you think it should wait for, that is, how long is too long when it comes to (slow) floppy I/O? :)

I looked further into this, and unfortunately, it isn't going to work well for the idle task to count buffers and then sync... so the remaining options are to add sync calls to dd (which isn't a great idea and also won't work with cat), or decrease the sync= time in order to help solve your issue here.

Regarding sync called from the kernel idle task, while checking for this issue I realized that the kernel idle task isn't a user program, and is not allowed to sleep. I coded that enhancement without regard for the non-sleep issue. There are certain cases where a buffer could be locked, (very unlikely, given our non-interrupt driven IO), where the sync code chain might call sleep, which would then cause a system crash. I am trying to think of an easy answer here, other than moving the sync to an update process (now I know why update is used on UNIX), which would use a full-time process slot, which is a huge waste of memory for this feature. Another thought is to put the sync into the shell or /bin/init.

Mellvik:
> So the question is whether the sync could be easily spread. Like, take a max number per operation in order to make the activity less disturbing - and spread the say, every second.
> 
> You mean change to a 1 second sync? Perhaps what we need, instead of a timer, is a count of buffers that need to be synced, and (also) start the sync when above that amount? (You can play with this in elks/init/main.c, where the idle task is).
> 
No, I think the 30 second strategy is a good one. It's predictable and has served well for ages. What I failed to convey was the attempt to spread it somewhat if there is a lot to sync. Like this: 
- 30-sec-kick-in
- oh, I have X buffers to sync, that's more than my chunk limit Y, I'll flush Y of them now, wait a sec, then do another Y buffers, till I'm done.
- If this takes more than 30 sec, we'll obviously have to increase the size of the chunk.
> 
> An entirely different approach would be to introduce the UNIX way of (immediate) character echoing, which - if the block IO lets interrupts get serviced at all - would keep the appearance of being alive up while sync is happening.
> 
> The keyboard is interrupt driven, as is the serial input, and even in UNIX, it uses full-duplex meaning the shell has to "echo" the character (if in raw mode like ash, or do something else, e.g. arrow keys).
> 
Well, that's a valid point, the linenoise stuff changes the equation quite a bit. Let's ignore the raw mode situation for a now, my point is that linux and unix echoes as typed instead of as read, which has a big influence on the appearance of system response. 
> I think the problem here is that the kernel itself doesn't run to even receive the interrupt keyboard character, until the BIOS disk read function returns. You talked about this earlier - there is nothing that can be done about this until a non-BIOS floppy driver is written.
> 
If that's the case, we're hosed. OTOH, it seems to me the serial response we're getting (SLIP etc.) may indicate otherwise. Might be an interesting experiment for a different time and day.
What might be tempting to test sooner, is immediate echoing, Unix style, knowing it has no effect in raw mode. IIRC a minor change in the tty driver.

--M
>

ghaerr:
> test sooner, is immediate echoing, Unix style, knowing it has no effect in raw mode. IIRC a minor change in the tty driver.

Go ahead and test with sash, which runs with stty echo. There is no change required in the tty driver, the characters ARE echoed immediately (when the kernel runs, that is). 

If you don't see anything, that means the kernel isn't running, since the system is likely stuck in a non-interruptible BIOS disk call. Even in that case though, the kernel should run after every two sectors written or so, which would service the keyboard interrupt and echo the character in non-raw mode.

OTOH, if what you're suggesting is that the kernel "echo" the character during the keyboard interrupt itself (providing BIOS disk hasn't actually disabled them), we can't quite do that - as the console TTY code and other normal functions can't be called from a hardware interrupt - since our ELKS kernel is non-reentrant (!).

ghaerr:
Thinking more about your request for an immediate echo - to let the user know the system isn't crashed during long pauses of no response - if you're on the console (not serial), we do display a rotating icon in the upper right corner when disk I/O is occurring. We're able to do that since that icon is displayed directly into EGA memory, bypassing the TTY and console drivers. It is written directly during the timer interrupt. That should tell you that the system is up. I'm not sure what we could on on a serial line, but we could possibly enhance what is displayed on the console?

Mellvik:
> ﻿
> I get your point now about the raw device bypassing the buffer system. What exactly is the benefit of that, if not for speed?
> In the ELKS case, I would say we'd want the floppy contents to be available in XMS, to reduce the chance of another very slow disk read.
> 
That's the thing. The buffer is caching. If caching is undesirable or useless, that's what the raw device is for. NOt the general case, but the odd cases. Like mine. I don't what this image file to end up in the buffers, I want it on the device. Directly. 
> No writing anything new. this is what a raw device is. A hook into the IO level just above the device.
> 
> ELKS doesn't have any lower level I/O than the block device driver. All I/O is within that (bioshd.c). So we would have to duplicate (or split off) all the disk I/O that is currently only in bioshd.c, just to implement a raw device. The disk interface in ELKS is only from the block device driver. Perhaps that's why there are no character (raw) devices in ELKS...
> 
Admittedly I haven't looked at the ELKS disk/floppy driver for a long time, but my point is that there cannot possibly be block IO without raw IO underneath. Of course if the hook isn't there, it must be created, but the functionality is there.
> things being optional the raw read/write may be a memory copy shorter.
> 
> As I mentioned above, any raw read/write cannot be directly to the application due to DMA segment wrap issues, thus there will always be one memory copy - exactly what I rewrote the block device system to do about a year ago. All ELKS block I/O is now to a system buffer, which then gets copied exactly once to an application buffer. The ethernet drivers (a character device) were rewritten to directly write into the calling application's buffer, thus bypassing system buffers, but we can't use BIOS disk I/O to do the same, even with a character device, because of the PC DMA problem on some BIOSes.
> 
There may be something I don't understand here. How can getting a single block of data (whether 512 or 1024 bytes) in or out be that difficult? The amount of data is so small that getting around DMA/boundary challenges cannot be that big?
> 
Interesting discussion indeed, @ghaerr.
Thank you.

--M

ghaerr:
> my point is that there cannot possibly be block IO without raw IO underneath. Of course if the hook isn't there, it must be created

The "raw" IO or "hook" is the direct call to BIOS INT 13h (read/write disk sector). There is no lower layer, the block device driver handles it all. This is probably because we don't actually have our own floppy or hard disk drivers.

> How can getting a single block of data (whether 512 or 1024 bytes) in or out be that difficult? The amount of data is so small that getting around DMA/boundary challenges cannot be that big?

Yes - you are now starting to get into it. The amount of code that is written, both in our boot sector, as well as setup, to deal with this is unbelievable. (It is now documented in Documentation/text/boot.txt, take a look at it, there are 7 steps, very complicated). ELKS does not have the luxury of requiring a modern machine that doesn't have these problems to run on (like Windows or later versions of Linux require). ELKS has to run on ANY PC.

The overriding issue is ELKS is absolutely dependent on BIOS disk I/O. If we (at boot, setup or runtime) issue a BIOS call that fails, the kernel crashes, hard. We have to deal with every BIOS, back to IBM PC BIOS, version 1. Very long story short, when ES:BX is passed to the BIOS for disk I/O, ES cannot be a value that might cause a segment wrap, due to limitations of the early DMA chip as well as buggy BIOS implementations that didn't rewrite (normalize) ES:BX for the DMA controller. Here is one example of an early issue, finally solved for one user: https://github.com/jbruchon/elks/issues/288 (in particular, see this [comment](https://github.com/jbruchon/elks/issues/288#issuecomment-581034966). )

If it makes you feel any better, it took me about a year to absorb all this, from the previous contributors and maintainers.

The solution I took in rewriting the block I/O subsystem was to enhance our memory manager so that buffers always start on a 1K boundary. This is not a big power of two, but I think the real problem is just segment offset wrapping around 64K. With 1K system buffers, we will never wrap if the buffer is 1K aligned. The EXT buffers and DMASEG buffers are 1K aligned (XMS buffers are too, but we can't do I/O directly into them anyways since they are not addressable by the CPU). We can't guarantee that an application buffer will be aligned on any boundary, which is the reason why a character device driver would have to copy disk contents to another buffer anyways.

I hope this all makes sense!

Mellvik:
Thanks again, @ghaerr - very educational. Appreciated.

I still maintain that we need a way to do IO without polluting the buffer cache (ie raw devices). So why don't we just let it rest right here and revive the issue when I have a concrete proposal to discuss?

That applies to the immediate serial echo issue too.

—M

PR #1149, opened 2022-02-12
[cmds] Add mount options: -a automount, -q query fstype

ghaerr:
One can now mount a floppy or hard drive automatically at boot without having to know what filesystem type it is first.

Use `mount -a /dev/fd1 /mnt` instead of previous `mount -t msdos /dev/fd1 /mnt`, for instance.
Use `mount -q {device} {mount_point}` to return filesystem type as exit code. See shell function in /etc/mount.cfg for suggested usage.

### DIFF
--- a/elks/fs/super.c
+++ b/elks/fs/super.c
@@ -326,7 +326,7 @@ int do_mount(kdev_t dev, char *dir, int type, int flags, char *data)
     dirp = dir_i;
     if ((dirp->i_count != 1 || dirp->i_mount) || (!fs_may_mount(dev)))
  	goto BUSY;
-    sb = read_super(dev, type, flags, data, 0);
+    sb = read_super(dev, type, flags, data, flags & MS_AUTOMOUNT);
     if (!sb) {
 	error = -EINVAL;
 	goto ERROUT1;
--- a/elks/include/linuxmt/fs.h
+++ b/elks/include/linuxmt/fs.h
@@ -89,6 +89,7 @@
 #define MS_NOEXEC	    8	/* disallow program execution */
 #define MS_SYNCHRONOUS	   16	/* writes are synced at once */
 #define MS_REMOUNT	   32	/* alter flags of a mounted FS */
+#define MS_AUTOMOUNT	   64	/* auto mount based on superblock */
 
 #define S_APPEND	  256	/* append-only file */
 #define S_IMMUTABLE	  512	/* immutable file */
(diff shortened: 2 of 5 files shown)


Mellvik:
Now that I've tested `mount -a` quite a few times:

It's very slow, which doesn't immediately make sense. Mounting with at specific `-t` option is fast (reasonably so).

```
login: root
# time mount -a /dev/hda1 /mnt
FAT: me=f8,csz=16,#f=2,floc=1,fsz=244,rloc=489,#d=512,dloc=521,#s=997857,ts=9978
57
FAT: total 498M, free 28M, fat16 format

Real    0m20.810s
# 
```

It takes about 3 seconds before the first `FAT:` message, then 15-18 seconds for the second message. Any way we can speed up this one?

-M

ghaerr:
Thanks for your testing. As usual, it's bringing out more problems with some designs... :)

There are two things going on here: the first is that `mount -a` passes the mount to the MINIX FS, which reads the superblock, and fails MINIX (since it's FAT), and then goes to FAT FS, which reads the block again. This should be buffered by the system buffers, however. I will check, the buffer may be discarded for some reason.

The second, and new big problem you're seeing, is a result of having added the FAT free space display recently. This code necessarily reads all the FAT table entries in order to compute the free space. I plan on adding the free space computation to a new system call `statfs`, which will allow `df` to work on FAT volumes (currently it works only on MINIX). That would delay the computation until the user requests it.

So the question is: should we delete the free space computation and display now, and wait for the `statfs`, or leave it in? We don't really need to know the FAT free space at mount time, but I did find it useful. I didn't think it would take nearly that long to compute free space, especially on a hard drive! But then again, that's why FAT32 has precomputed free space stored in special sectors... gotta love FAT.

I think you'll find that doing `mount -t msdos` will be just as slow, if you do it immediately after boot; if you do it after the mount -a, the system buffers for the FAT entries will be buffered, thus executing quickly.

ghaerr:
> which reads the superblock, and fails MINIX (since it's FAT), and then goes to FAT FS, which reads the block again.

I just checked this, and "mount -a" will end up reading an extra 1K (two) disk sectors, thus being slower than "mount -t". The reason for this is that MINIX superblock is at block 1, while FAT superblock is block 0.

Thus, "mount -a" should only be used for automated mounts when it is not known what the filesystem type is, for speed. I don't see a way around this for automated mounting.

PR #1150, opened 2022-02-13
[cmds] Miscellaneous enhancements and cleanups

ghaerr:
Update build.md instructions for how to create full suite of MINIX/FAT images for all floppy and hard disk formats.

Moves /lib/liberror.txt to /etc/perror (error strings file for perror). Note libc/include/paths.h now contains paths and filenames in one place for easier maintenance and inspection.

Update /etc/passwd to allow 'sync' login.

Update `sum` program to use 2K internal buffer for higher speed.


Mellvik:
Thanks again, @ghaerr.


I have to repeat that the recent buffer changes (in particular) have turned using ELKS on real HW upside down.

It is - like we talked about - almost like running in qemu. With ktcp now even more stable, files and apps are 'flying' back and forth (ftp airlines) and process IDs easily run in the 500-1000 range.

An occasional sync - no problem. Extremely encouraging.

—M

ghaerr:
Thanks @Mellvik, it's been fun figuring out how to add these features recently.

> It is - like we talked about - almost like running in qemu

I'm traveling, but can't wait to try it out on real hardware, my Compaq 386 Portable!

> An occasional sync - no problem. Extremely encouraging.

I'm thinking of adding a (configurable) 30-second sync in the kernel idle loop...

Thank you!

Mellvik:
> 
> An occasional sync - no problem. Extremely encouraging.
> 
> I'm thinking of adding a (configurable) 30-second sync in the kernel idle loop...
> 
> 
That's  probably the smart way to go - in particular since cron isn't needed for anything else at the time. 
And having it configurable from /bootopts would be perfect. Maybe (I haven't really thought this through yet) a max # of blocks to sync every time to limit the time the systems is (more or less) dead.

Since we initiated this discussion yesterday, I tried to look up the venerable update() program, only to discover that it has been gone for so long it doesn't easily show up on the net. But my PDP11/70 replica running BSD2.11 came to the rescue:

     Update is a program that executes the sync(2) primitive
     every 30 seconds.  This insures that the file system is
     fairly up to date in case of a crash.  This command should
     not be executed directly, but should be executed out of the
     initialization shell command file.

The kernel idle loop is even better, and certainly simpler.

Thank you!

—M

PR #1151, opened 2022-02-13
[fat] Calculate and display FAT volume free space on mount

ghaerr:
Calculates and displays free space remaining on FAT/MSDOS volumes at mount time.

With some work, this could be added to `df` command by implementing `statfs` system call.

Add more options to qemu.sh.
Other cleanups and compilation warning suppressions.


Mellvik:
Thank you @ghaerr - 
this is great! You've had a busy weekend. much appreciated.

In particular the addition of the 1200k images to qemu.sh - I've had that on my list for a while.
Still on the list is to populate the 1200 image reasonably - right now I think it has the same content as the 720k image. Needless to say, I'm using this image quite a bit (and ship a 1200k ELKS floppy with every old klunker I refurbish and sell).

—M

PR #1152, opened 2022-02-13
[images] Build reasonable 1200K floppy distribution

ghaerr:
> Still on the list is to populate the 1200 image reasonably

Here you go. Its got almost everything that the 1440K floppy contains except a few nano-X programs.

> and ship a 1200k ELKS floppy with every old klunker I refurbish and sell.

Pretty cool!


Mellvik:
Just made my day!  :-)


-M

PR #1153, opened 2022-02-13
[kernel] Add auto-sync devices to kernel using sync= in /bootopts

ghaerr:
Will sync all devices if sync=[# seconds] set in /bootopts.

Default off for now, pending more usage testing and usability.

### DIFF
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -29,6 +29,7 @@ int root_mountflags = 0;
 #endif
 int net_irq, net_port;
 static int boot_console;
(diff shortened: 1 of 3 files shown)


Mellvik:
Wow @ghaerr, that was fast!

Quick question: In main.c, I don't see  sync_interval be initialized anywhere?

-M

ghaerr:
> In main.c, I don't see  sync_interval be initialized anywhere?

It's there... and tested this before submission :)

Issue #870, opened 2020-11-16
Tinyirc doesn't work (nslookup fixed)


toncho11:
Are you able to use tinyirc?

There is no DNS in ELKS, right?

So I use `tinyirc 185.30.166.38 6667` to connect to irc.freenode.org, but it never connects.

Also `nslookup` is not working.

ghaerr:
There is no DNS in ELKS.

> Are you able to use tinyirc?

I've never used it. I would doubt anyone has run it in the last 20 years. It looks to have been created between 1991-1996.

> Also nslookup is not working.

Have you looked at or changed /etc/resolv.conf? Nslookup uses that to get its DNS data; it likely needs updating.
Due to additional long-name issues with FAT filesystems, this should be tested on a MINIX filesystem. The ".conf" extension can be opened, but not searched with `ls` on FAT, IIRC. Also, netstat has a compiled-in nameserver address when /etc/resolv.conf is missing, which may want to be updated, and perhaps /etc/resolv.conf removed from the distribution.

Volunteers needed for fixing these older utilities. If they can be proven non-operational, perhaps they should be removed from the distribution until fixed?

toncho11:
There are not may things one can do with functional network ... so it will be better to fix tinyirc. Maybe I can work on this in the future. Tinyirc can connect only to plain-text connections, not TLS.

ghaerr:
> Maybe I can work on this in the future. 

You might take a stab at nslookup, its a lot simpler. I'm thinking if it could be made working, we could rename /etc/resolv.conf to /etc/resolv.cnf also, that would solve the FAT issues above. If you start on it, I will help you if needed.

> Tinyirc can connect only to plain-text connections, not TLS.

Oh, so tinyirc works otherwise?

ELKS won't likely ever have TLS support, the crypto libraries are way too large.

toncho11:
I tried two small IRC clients to compile, but:

- C++ is not supported in ELKS?
- some headers were missing in the compiler when trying a C IRC client

ghaerr:
> some headers were missing in the compiler when trying a C IRC client

Which ones? Unable to tell whether they are standard headers or library headers.

ghaerr:
Hello @toncho11 and @tkchia,

> C++ is not supported in ELKS?

I've never used C++ on ELKS, so I don't know - @tkchia, can you comment on that?

Thank you!

tkchia:
Hello @ghaerr,

[...]

PR #1156, opened 2022-02-14
[cmds] Rewrite nslookup - now works

ghaerr:
Rewrote nslookup.c from scratch. Now works!

Usage: nslookup \<domain> [server]

This is the beginning of having real DNS capabilities on ELKS, as nslookup was rewritten so the DNS resolver can be moved to the C library, and used to greatly enhance the current `in_gethostbyname` call to resolve hostnames. This is a *very* simple implementation.

However, there are problems which may have to be "resolved" in order for this to work well:
- We have to use a TCP connection to name server since ktcp doesn't have UDP, which could hang
- There is not (yet) a way to timeout on `connect` to DNS (or any other host), which could cause DNS lookups to hang

I am working on a solution for connect timeouts, but it involves a heavy change to the kernel/ktcp interface to be completely event-driven, rather than waiting for an individual reply (or not), as it does now. This change will also allow the previous `accept` problem of the timing between application accept calls and ktcp SYN received to be fixed as well. More on that later.

Nonetheless, I'd like to get more comments on nslookup now, and plan to add it to the C library so that ELKS network applications can access the net more widely, even before the above problems are fixed.

Here's a screenshot. Most addresses match my desktop's nslookup, except GitHub.com, which is returning a 192.30.255.113 address, which doesn't match (and my browser also refuses to connect to that address because of an invalid certificate). I'm not sure that the reasons are behind this yet. For those that are familiar with DNS records, they can be inspected more deeply by turning on DEBUG in nslookup.c.
Screen Shot 2022-02-14 at 8 22 44 AM

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -156,7 +156,6 @@ inet/httpd/sample_index.html	::var/www/index.html	:net
 ktcp/ktcp						:net
 inet/nettools/netstat			:net
 inet/nettools/nslookup			:net
-#inet/nettools/resolv.cfg	::etc/resolv.cfg	:net
 inet/nettools/arp				:net
 inet/telnet/telnet				:net
 inet/telnetd/telnetd			:net
--- a/elkscmd/inet/nettools/Makefile
+++ b/elkscmd/inet/nettools/Makefile
@@ -18,7 +18,7 @@ all: $(PRGS)
 
 install: $(PRGS)
 	$(INSTALL) $(PRGS) $(DESTDIR)/bin
-	$(INSTALL) resolv.cfg $(DESTDIR)/etc
+#	$(INSTALL) resolv.cfg $(DESTDIR)/etc
 
 netstat: netstat.o
 	$(LD) $(LDFLAGS) netstat.o -o netstat $(LDLIBS)
--- a/elkscmd/inet/nettools/nslookup.c
+++ b/elkscmd/inet/nettools/nslookup.c
@@ -1,3 +1,8 @@
+/*
+ * DNS Name Lookup
+ *
+ * Entirely rewritten by Greg Haerr Feb 2022
+ */
 #include <stdio.h>
 #include <unistd.h>
 #include <stdlib.h>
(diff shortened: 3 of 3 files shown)


Mellvik:
@ghaerr,
this is an important improvement!! I mean generalizing the gethostbyname call. 
I'm very busy with other stuff for the next week and a half, but I'll do some testing of this in between.

Also thanks for diving into the ktcp timeout issue. The final (known) notorious hang situation in ELKS, and thus the single reason for probably 90% of my reboots these days. Misspelling an address and I'm dead …
Is the alarm() call part of this improvement btw?

Thank you !!

—M

ghaerr:
> The final (known) notorious hang situation in ELKS, and thus the single reason for probably 90% of my reboots these days. Misspelling an address and I'm dead …

Yes, we badly need to be able to interrupt or timeout a connect call. I was originally going to use alarm(), but the way ktcp and the kernel communicate, interrupting an otherwise "synchronous" call would hang ktcp for sure. Then I thought of adding a special timeout in ktcp using an SO_xxx setsockopt hack, but while rewriting nslookup realized that we really need to be able to interrupt any networking call (e.g. `read` after having connected to DNS, but no response). That's when I realized the only way to do this is to do it right - and rewrite ktcp and the kernel so that all calls are asynchronous.

My plan is to start with connect in a small way, and post a series of PRs that work on QEMU that you can beat up more heavily. (We need an automated file transfer test suite!! A shell script or two in /root would be fine for now).

> Is the alarm() call part of this improvement btw?

Yes! I've already written it, but we can't use it until ktcp/kernel are rewritten, or any alarm firing would hang the system anyways. In the new design, alarm() would return the process from the kernel, but not leave ktcp in a (hanging) synchronous state, since it would only operate asynchronously anyways. That is, ktcp would "call back" the kernel on a connect (or accept) instead of now, where the kernel issues a connect/accept to ktcp, and ktcp has to block, handle the request and return the result, before doing anything else. The new design should also ktcp socket state transitions as well, setting flags that are then reported to the kernel with callbacks. This would solve the accept/fin/close problems we've worked around.

ghaerr:
Fixes nslookup problem first reported in #870.

Mellvik:
Thank you @ghaerr,

> 
> 
> The final (known) notorious hang situation in ELKS, and thus the single reason for probably 90% of my reboots these days. Misspelling an address and I'm dead …
> 
> Yes, we badly need to be able to interrupt or timeout a connect call. I was originally going to use alarm(), but the way ktcp and the kernel communicate, interrupting an otherwise "synchronous" call would hang ktcp for sure. Then I thought of adding a special timeout in ktcp using an SO_xxx setsockopt hack, but while rewriting nslookup realized that we really need to be able to interrupt any networking call (e.g. read after having connected to DNS, but no response). That's when I realized the only way to do this is to do it right - and rewrite ktcp and the kernel so that all calls are asynchronous.
> 
This does sound like a significant endeavour, and I fully agree - very important, not only to do but to do right. From your description it sound like the result (of making the kernel/ktcp relation fully asynchronous) may even be a more responsive system (less wait loops).
> My plan is to start with connect in a small way, and post a series of PRs that work on QEMU that you can beat up more heavily. (We need an automated file transfer test suite!! A shell script or two in /root would be fine for now).
> 
I can generalize what I have - using ftpget/put and a set of test files from zero bytes to 200k...
> Is the alarm() call part of this improvement btw?
> 
> Yes! I've already written it, but we can't use it until ktcp/kernel are rewritten, or any alarm firing would hang the system anyways. In the new design, alarm() would return the process from the kernel, but not leave ktcp in a (hanging) synchronous state, since it would only operate asynchronously anyways. That is, ktcp would "call back" the kernel on a connect (or accept) instead of now, where the kernel issues a connect/accept to ktcp, and ktcp has to block, handle the request and return the result, before doing anything else. The new design should also ktcp socket state transitions as well, setting flags that are then reported to the kernel with callbacks. This would solve the accept/fin/close problems we've worked around.
> 
> 
Well wouldn't that be something. Looking forward to it.

—M
>

Issue #1155, opened 2022-02-14
[system] Practical experiences


Mellvik:
With recent vast improvements in stability, performance and usability, some of us use ELKS differently - and more. In my case that means 2 physical systems with ethernet and (primarily) serial access. Thus some practical experiences and suggestions.

When starting off with a fresh copy from the repository (no local adaptions), the procedure after first boot is as follows:
- Edit `/bootopts` to set console, run level, console speed, networking, buffers and now sync value. Ethernet defaults are fine.
- Edit `/etc/inittab` to get the serial speed right - for ttyS0 and ttyS1.
- Edit `/etc/net.cfg` to set the right IP address, possibly to eliminate `httpd` from the default startup and (possibly) set or remove the debug option from/to `ftpd`.
- Edit `/etc/mount.cfg` to
 - ... add the a-option to `fsck` (don't want the boot process to stop/wait regardless ...)
 - ... Enable checking of the root floppy (and other minix filesystems, I don't have that yet though)
 - ... Enable mounting of the DOS partition of the HD

All good, these are all settings to adapt to the local requirements.  And yes, I could have my own set of config files ready when building the images. However, recent development has changed those files about as frequently as other code ...

IOW, there are a few things that would have made life a lot easier:
- A LOCALIP (or HOSTNAME) variable in /bootopts would get me going w/o editing more than one file (other adjustments are less critical).
- Having `init` somehow pick up the console speed setting in /bootopts would be very helpfull.
- Having a FSCK variable in /bootopts that 1) tell mount.cfg to fsck root and 2) contain the fsck options for that particular run.

My 2 cents ...

--M

ghaerr:
Thanks @Mellvik, I'll think about this a bit more. 

It sounds like having localip= and a baud= serial port speed in /bootopts would solve most of your problems other than mount checking. Can a single baud rate suffice for all serial ports speed defaults? This would allow the /bin/getty speed parameter to be made optional in /etc/inittab and getty would just use the default speed.

A possible problem with using localip= (rather than LOCALIP=) is that `net start` would only work at boot, not later, unless we kept the environment variable around for the entire ELKS uptime (i.e. use LOCALIP=). I've been trying hard not to require lots of special long-term environment variables, which is why the previous HOSTNAME= was removed.

> add the a-option to fsck (don't want the boot process to stop/wait regardless ...)

The fsck source says that -a isn't implemented, so we actually need to test that more before getting ahead of ourselves. Have you played with it much? 

[...]

PR #1157, opened 2022-02-15
[boot] Add LOCALIP= and getty= to /bootopts

ghaerr:
Adds two items requested in #1155.

LOCALIP= specifies ktcp local IP address, and works whenever `net start` is run.

getty=19200 (or other baud rate) overrides any /bin/getty line in /etc/inittab that also specifies a baud rate. This allows /etc/inittab to remain unchanged, yet be overidden at boot when desired.
/etc/net.cfg defaults to not running httpd for now.

Also fixes ":once:" and non-/bin/getty commands, they were broken, now run correctly using "sh -c" to run from init.
Various other compilation warnings addressed.

### DIFF
--- a/elkscmd/rootfs_template/bin/net
+++ b/elkscmd/rootfs_template/bin/net
@@ -1,12 +1,13 @@
 # Start/stop ELKS networking
 #
-# Usage: net [start|stop] [eth|slip|cslip] [baud] [device]
+# Usage: net [start|stop|show] [eth|slip|cslip] [baud] [device]
 #
 # Examples:
 #	net start eth			start ethernet networking
 #	net start slip			start slip networking
 #	net start slip 19200	start slip at 19200 baud
 #	net start cslip 4800 /dev/ttyS1
+#	net show				shows settings for 'net start'
 #
 # See slattach.sh for Linux slip setup
 # See qemu.sh NET= line for host forwarding into ELKS using QEMU
@@ -17,7 +18,7 @@ source /etc/net.cfg
 
 usage()
 {
-	echo "Usage: net [start|stop] [eth|slip|cslip] [baud] [device]"
+	echo "Usage: net [start|stop|show] [eth|slip|cslip] [baud] [device]"
 	exit 1
 }
 
@@ -82,6 +83,9 @@ start)
 	if test "$4" != ""; then device=$4; fi
 	start_network ;;
 stop) stop_network ;;
+show)
+	echo -n ip $localip gateway $gateway mask $netmask $link
+	if test "$link" != "eth"; then echo "" $baud $device; else echo; fi ;;
 *) usage ;;
 esac
 
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -1,5 +1,5 @@
 ## boot options max size 511 bytes
-#console=ttyS0 debug net=eth 3 # serial console, multiuser, networking
+#console=ttyS0,57600 debug net=eth 3 # serial console, multiuser, networking
 #QEMU=1			# to use ftp/ftpd in qemu
 #TZ=MDT7
 #LOCALIP=10.0.2.16
@@ -8,6 +8,5 @@
 #sync=30		# autosync secs
 #init=/bin/init 3 n	# multiuser serial no /etc/rc.sys
 #init=/bin/sh		# singleuser shell
-#getty=300		# getty baud override
 #console=ttyS0,19200	# serial console
 #root=hda1 ro		# root hd partition 1, read-only
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -1,11 +1,13 @@
 ## boot options max size 511 bytes
 #console=ttyS0 debug net=eth 3 # serial console, multiuser, networking
 #QEMU=1			# to use ftp/ftpd in qemu
-#TZ=MDT7		# timezone
+#TZ=MDT7
+#LOCALIP=10.0.2.16
 #netirq=9 netport=0x300 # NIC
-#bufs=2500		# XMS max, EXT max is 256
-#sync=30		# sync seconds
+#bufs=2500		# system buffers
+#sync=30		# autosync secs
 #init=/bin/init 3 n	# multiuser serial no /etc/rc.sys
 #init=/bin/sh		# singleuser shell
+#getty=300		# getty baud override
 #console=ttyS0,19200	# serial console
 #root=hda1 ro		# root hd partition 1, read-only
--- a/elkscmd/rootfs_template/etc/inittab
+++ b/elkscmd/rootfs_template/etc/inittab
@@ -3,7 +3,7 @@ id:1:initdefault:
 si::sysinit:/etc/rc.sys
 
 #l1:1:wait:/etc/rc 1
-#ud::once:/bin/update
+#ud::once:/bin/printenv
 
 # getty runlevels
 # 1 single user tty1 only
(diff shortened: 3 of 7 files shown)


Mellvik:
Thanks again, @ghaerr - 

this is fast turnaround indeed.

The LOCALIP= is exactly what's needed.

As to the serial settings I'm a more difficult customer. We still need to make two mods in order for this to work:

## boot options max size 511 bytes
console=ttyS0,57600 debug net=eth 3 # serial console, multiuser, networking
#QEMU=1                 # to use ftp/ftpd in qemu
#TZ=MDT7
#LOCALIP=10.0.2.16
#netirq=9 netport=0x300 # NIC
bufs=2500               # system buffers
sync=30         # autosync secs
#init=/bin/init 3 n     # multiuser serial no /etc/rc.sys
#init=/bin/sh           # singleuser shell
getty=57600

It's an improvement since only one file needs to be edited, but to me the addition of another variable seems like a complication that can be avoided. 
Maybe I was ambiguous in my description of the problem. My thinking was that init would pick up the speed setting from the console= line, thus connect the serial port and the speed setting, and use that to override that particular getty. From my perspective it's advantageous if this mod only affects the console, not the other serial lines.

And BTW thanks for the net.cfg mod - appreciated.

—M

ghaerr:
> Maybe I was ambiguous in my description of the problem.

Yes, I didn't understand until your last post and more precise description. 

>  From my perspective it's advantageous if this mod only affects the console, not the other serial lines.

I thought you were saying you needed something that would work for two serial ports. But I think I understand now what you need.

> t's an improvement since only one file needs to be edited, but to me the addition of another variable seems like a complication that can be avoided.

Yes, agree that a 'getty=' line is kind of a hack, now that I think more about it.

> My thinking was that init would pick up the speed setting from the console= line, thus connect the serial port and the speed setting, and use that to override that particular getty. 

We can't use the console= line to pass any data, since that's not an environment variable and that information can't be seen at /bin/init time. Only the env=value lines do that. However, one possible solution would be that the serial port remembers it's last value (or the console= value) on subsequent opens. It may already be doing that, but certainly the /etc/inittab getty lines will override any previous setting. That could maybe be fixed be removing the baud rate from the /etc/inittab line(s).

Overall, still complicated. How about something much simpler? Remove the getty= change, and just ship /bootopts with a console=ttyS0,57600 already set (commented out), and also change the /etc/inittab getty lines to 57600 default, for now?

I have no problem changing to 57600 default in both places, to aid your ability to continue to deploy and test! We can change them to other defaults later, if desired, but 57600 is a more modern serial speed anyways. Also, funnily enough, this issue resulted in finding (yet another) bug in /bin/init - all non-/bin/getty lines failed to function!

Mellvik:
> 
> Overall, still complicated. How about something much simpler? Remove the getty= change, and just ship /bootopts with a console=ttyS0,57600 already set (commented out), and also change the /etc/inittab getty lines to 57600 default, for now?
> 
> I have no problem changing to 57600 default in both places, to aid your ability to continue to deploy and test! We can change them to other defaults later, if desired, but 57600 is a more modern serial speed anyways. Also, funnily enough, this issue resulted in finding (yet another) bug in /bin/init - all non-/bin/getty lines failed to function!
> 
> 
That would certainly do it for now, @ghearr - no problem, appreciated.

While we're at it, and since you mention it: If the tty remembers its speed setting from the console= (likely since it's outputting at that rate during the boot sequence), and since there is an overall system default speed of 9600, it might be possible to have init check if the speed is different (than 9600) on the console port, and override inittab with that speed when starting getty?

Wow, that was a long sentence. Was that intelligible at all?

Thank you!

—M

ghaerr:
> That would certainly do it for now, @ghaerr - no problem, appreciated.

I have this done, except for one item: do you want only ttyS0 set to 57600, or also ttyS1 to 57600 in /etc/inittab? I have changed console=ttyS0,57600 in /bootopts and /bin/getty ttyS0 57600 in /etc/inittab.

>  it might be possible to have init check if the speed is different (than 9600) on the console port, and override inittab with that speed when starting getty?

I understand your request. That means there would be some cross-checking between ports in the serial drive and silently changing a baud rate. I feel we're better off keeping things simple, and keeping the ability that only the serial ports specified (anywhere) have their values changed. This will certainly help when there's a baud rate connection issue, which we know does happen.

I have also added a `net show` option that displays the settings that would/do take place when a `net start` is executed. This will help you quickly see everything, except for the case when specific parameters are passed to net start. You can play with that and see what you think.

ghaerr:
> it might be possible to have init check if the speed is different (than 9600) on the console port, and override inittab with that speed when starting getty?

I might have interpreted this wrong. I think if we're going to do something tricky, it should be done only when there is NO baud rate specified in the /bin/getty line(s) in /etc/inittab. In that way, /bin/init (nor getty) should not change the baud rate at all, but use the previous setting... that is if ELKS currently doesn't reset the baud rate between open/closes. Perhaps you could check that and we'll talk about it after this issue set of changes is committed.

ghaerr:
I still don't know exactly what you're trying to do. Do you want ttyS1 set to a baud rate always specified explicitly in /etc/inittab, but have the ability to change ttyS0 in /bootopts, and also have that setting work for logins on ttyS0? Or something else?

ghaerr:
I just checked the source for getty.c as well as the serial driver. It seems the serial driver will keep a previous baud rate, which is initially either set to the system default (9600), or by /bootopts. If a baud rate is NOT specified in /etc/inittab, then getty does NOT reset the baud rate. So I think your problem is solved by leaving the ttyS0 line with no baud rate in /etc/inittab, and specify it in /bootopts. The ttyS1 baud rate will be set to 9600 at boot, and then reset to whatever value is in /etc/inittab. 

Also, if a /etc/inittab getty spawns a login and the user changes the baud rate, logging out and then in again will not change the baud rate. I think this is what we want, for now.

I just need to know  what to set the /etc/inittab lines for ttyS0 and ttyS1 to. I presume none for ttyS0 and need to know what ttyS1 should be set to by default.

Mellvik:
Agreed, @ghaerr,

let's forget about ttyS1 altogether (IOW keep it as is). It will be a very odd case when console is serial and not ttyS0.

My recent mentioning of ttyS1 was in the context of the general getty= in bootopts, which would set the ports to the same speed. That would not be desirable because they have different uses (the 2nd line more often used for file transfers).

so yes, lets stick to ttyS0, and the problem to solve is simply to ensure that the boot messages and the getty login prompt end up with the same baud rate.

And yes, I'll check whether ELKS actually resets speed between opens/closes. It's easily down with stty.


—M

Mellvik:
COnfirmed: The speed sticks between opens. 

So if getty leaves the speed alone if missing in inittab, we should be fine. Actually OK regardless since then the speed will be 9600 if nothing is changed in /booptopts.

Thank you.

—M

ghaerr:
> so yes, lets stick to ttyS0, and the problem to solve is simply to ensure that the boot messages and the getty login prompt end up with the same baud rate.

> So if getty leaves the speed alone if missing in inittab, we should be fine. Actually OK regardless since then the speed will be 9600 if nothing is changed in /booptopts.

Ok, I'll post this PR for your inspection. It will remove the baud rate from /etc/inittab for ttyS0, and should work. If it doesn't we'll fix it so that it does. This means that /bootopts will control the startup speed ttyS0, using the console= line.

ghaerr:
> COnfirmed: The speed sticks between opens.

Great. I will commit this now.  At this time, I am not thinking of making further changes to mount.cfg discussed in #1155.

I suggest you test this, and then add some specific script code in mount.cfg that seems to work well (for fsck etc), and post that in #1155 for further discussion on what to do, if anything. I can then parameterize it etc for final distribution.

PR #1158, opened 2022-02-15
[cmds] Add -t (sort by reverse modified time) to ls

ghaerr:
Finally, we have `ls -lt`!


Mellvik:
Thank you @ghaerr - 'finally' is absolutely correct. -t has been missed.

Also thanks for increasing the heap - which has been a regular problem recently with the network / file transfer testing and LOTS of files in a directory.

All the small steps.

--M

PR #1159, opened 2022-02-16
[net] First pass at rewriting ktcp/kernel to use event callbacks

ghaerr:
Rewrites ktcp/kernel to use callback for TCP connect events.

With this change, one should be able to ^C out of a connect system call, in the case a bad address was typed. Currently only tested on telnet in QEMU. 

Next step is adding `alarm` system call, which should now to allow connect and read system calls to be interrupted. It will not working with accept, which requires more changes.

### DIFF
--- a/elks/arch/i86/kernel/syscall.dat
+++ b/elks/arch/i86/kernel/syscall.dat
@@ -45,7 +45,7 @@ setuid		+23	1
 getuid		+24	1	* this gets both uid and euid
 stime		25	2	- this must not exist - even as a libc.
 ptrace		26	4	@ adb/sdb/dbx need this.
-alarm		27	2
+alarm		+27	2
 fstat		+28	2	 
 pause		29	0	 
 utime		+30	2	 
--- a/elks/include/linuxmt/errno.h
+++ b/elks/include/linuxmt/errno.h
@@ -141,6 +141,11 @@
 #define	EDQUOT		122	/* Quota exceeded */
 #define	ENOMEDIUM	123	/* No medium found */
 #define	EMEDIUMTYPE	124	/* Wrong medium type */
+#define ENONAMESERVER	125	/* Nameserver not found */
+#define ENONAME		126	/* Name not found */
+#define EBADQUERY	127	/* Bad query format */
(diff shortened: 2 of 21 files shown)


ghaerr:
Adds `alarm` system call. Should be used only for `connect` and `read` for now.

Add `in_connect` to C library. Identical to `connect` except takes fourth parameter which is timeout seconds. Uses alarm and returns -ETIMEDOUT on timeout, appropriate for `perror`. 

Updates `nslookup` for 2 second connect/read timeout, and `telnet` for 10 second connect timeout.

Mellvik:
This is just great, @ghaerr - 
And I have real problems keeping up with you these days. By the time I've generated a new image, there is another merge to pull :-)

Test reports coming.

--M

Mellvik:
Looking really good, @ghaerr,

timeouts and INTR signals in connect! an incredible improvement right there.

So really, I suppose the ftpo client should be modified to use in_connect before I submit the latest changes - or do you consider in_connect a temporary thing?

Comment (minor nuisance) in nslookup:

when it doesn't find (timeout) the name server, IMHO it shouldn't report the requested name as 'not found'.

# nslookup mymayday.com 3.4.5.6
Can't connect to 3.4.5.6
mymayday.com: Name not found
# nslookup mymayday.com 
mymayday.com is 172.67.188.110
# 


Thanks again.

—M

ghaerr:
Hello @Mellvik,

Thank you for the quick testing!

> I suppose the ftpo client should be modified to use in_connect before I submit the latest changes - or do you consider in_connect a temporary thing?

Yes, that would be great if you could update ftp/ftpd with in_connect. The in_xxx series in ELKS libc are useful routines that are almost, but not quite compatible with the BSD sockets standard. They either take a modified parameter (in this case), or return something slightly differently (e.g. in_gethostbyname).

> when it doesn't find (timeout) the name server, IMHO it shouldn't report the requested name as 'not found'.

Good idea - I'll change that. Internally, I'll rewrite the resolver routine to return an appropriate 'errno', which then can be used by calling programs (like nslookup) to print the most appropriate error message.

Thank you!

PR #1160, opened 2022-02-16
[net] Add DNS name resolution to in_gethostbyname and network apps

ghaerr:
Adds automatic DNS name resolution to network applications using `in_gethostbyname`.

Moves `in_resolv` (dns name resolver) to libc.
Adds `in_resolv` to `in_gethostbyname`.

Revises method of reporting errors in libc network routines, in most cases, a single `perror` will result in the correct error message. It is highly suggested to use perror as a result of in_gethostbyname or in_connect failure, to get the correct error message, as now the name translation and connection routines are more complex (e.g., Name not found, Nameserver not found, Connection timed out, etc).

Replaces previous "Socket error: (check if ktcp is running)" with simpler and automatic perror: "Network is down" message in all network applications (except ftp/ftpd).

Rewrite `net_connect` in urlget to use updated method of perror, rather than having perror within network connection routine.

Updates telnet, telnetd, netstat, arp, and urlget perror/error messages to use enhanced system.
Fix telnet to check for error return by in_gethostbyname and not try connect to 0.0.0.0.
Standardizes exit return codes on network programs (0=success, 1=fail).


Mellvik:
Thanks you very much, @ghaerr.
Another great enhancement!

I'll take care of ftp/ftpd.

—M

PR #1162, opened 2022-02-16
[cmds] Add man pages to runtime

ghaerr:
Adds already-written man pages to /lib/man? directories on ELKS. Must be configured with new option CONFIG_APP_MAN_PAGES=y (or use `make menuconfig`), default OFF. Default on when building disk images by size for 1200K, 1232K, 1440K and 2880K or larger.

Usage: `man <command> [section]`. By default, `more` is used to paginate. For more information, `man man` :)

It would be nice to add man pages for a number of our oft-used commands, instead of having to lookup "how to" on the wiki.
If interesting, a template for such could be produced to make it easy for those willing to document a bit.

There's still some work to do, not all man pages troff/nroff commands are recognized (which produce visual error messages), and none of the pages have been inspected or updated with possible directory name, etc changes. That's coming, and contributions are welcome.

Removed a number of outdated or unused commands, both for man pages, as well as for cleanup and duplications.
The full list is ed, logname, mesg, whoami, fgrep (use grep), proto, cksum, env (use printenv), lp, remsync, synctree, and tinyirc (not working). Any of these can easily be added back in.

To see which commands have manual pages, use `ls -R /lib`.

Screenshots:
Screen Shot 2022-02-16 at 4 35 10 PM
Screen Shot 2022-02-16 at 4 35 25 PM


ghaerr:
Updated `more` to know about reverse video and bold to work better with man pages. Also allows Enter to advance by only a single line (with "space" continuing to advance a whole page).
Screen Shot 2022-02-16 at 4 57 11 PM

Mellvik:
Much appreciated @ghaerr!
Manpages have been very much missed, contributions to pages coming.

I suspect formatting time for man pages may become an issue on old klunkers. If so, I have a port of catman() to Venix that should work out of the box. Not useful for floppies (space) but excellent on HD installations.

Thank you!

--M

Mellvik:
Hi @ghaerr - thanks again for reviving, enabling and updating the ELKS manpages! Admittedly I have completely ignored their existence this far.

After having acquainted myself with the setup, I realize that under the hood this is not like Unix/Linux manpages, so my comments above about using catman were based on wrong presumptions.

From what I understand, ELKS man pages have a special markup - very simplified nroff/troff commands that get rendered directly by the `man` command without the need to pipe them through nroff, which would be a performance disaster on ELKS on physical hw. The current setup is fast and relatively simple.

There is however, one disadvantage: It means that importing 'standard' (nroff -man) manpages and adapting them for ELKS is not possible. It is a significant job to adapt them to the simplified markup.

For discussion: There is a way to get around this, however - somewhat similar to what `catman` does on Unix and Linux.  In short, `catman` automatically caches formatted man pages - in .../cat1, .../cat2 etc while the unformatted pages remain in .../man1, .../man2 etc. The `man` command looks for the cached page first, if not found, formats and caches the page. If the original manpage is newer than the formatted, it reformats etc.

This could easily work on ELKS where the `nroff` formatting could be run on the cross host and only the preformatted pages passed on to the ELKS image. Very similar to the current setup, but with the advantage of manpage portability and full `nroff -man` macro support. It would require a version of `nroff` to be part of the cross toolset, which I don't think is much of a challenge.

Any thoughts ??

-M

ghaerr:
Hello @Mellvik,

Thank you for looking under the hood at the man page PR.

> From what I understand, ELKS man pages have a special markup - very simplified nroff/troff commands that get rendered directly by the man command without the need to pipe them through nroff, which would be a performance disaster on ELKS on physical hw

Yes. When I started looking at reviving one of the two `man` programs currently already included in the ELKS application source tree, I selected the one written by Robert de Bath (who also contributed most all of the C library back in 1995) precisely because it did it's own interpretation of the nroff commands. The other version, from the minix distribution, requires nroff, which needless to say, won't be ported to ELKS runtime anytime soon.

This approach has several advantages, speed of course, as well as the ability to control how the output appears (bold, reverse video, colors, etc). I rewrote some code that adds bold and reverse video, but would like to consider colors in the future.

>  ELKS man pages have a special markup - very simplified nroff/troff commands that get rendered directly by the man command 

I wasn't aware that these are using a special "simplified" markup - I figured we'd be better off with simplicity than having complex command requirements, especially if we have to write man pages ourselves. Can you detail a bit more about how this differs from modern man page sources?

Also, it turned out that selecting Robert's man command worked well with the existing few man pages (likely written by Robert himself, and several man pages (e.g. cron) don't display well at all.

> There is however, one disadvantage: It means that importing 'standard' (nroff -man) manpages and adapting them for ELKS is not possible.

I suppose you're probably correct here. And I am thinking about where we might get a set of man pages to start with. I checked "dev86" which Robert contributed heavily to, but didn't find anything.  I was thinking we should try to keep the man pages simple (as we don't really have much space to keep them outside of hard disk?!). Thus, determining which set of nroff/troff macros we might want to use is a big issue. Could we use UNIX v6 commands, for instance?

>  It is a significant job to adapt them to the simplified markup.

I can imagine that doing this by hand would be a huge job, depending on where we select a set of already-written pages from.

> somewhat similar to what catman does on Unix and Linux. In short, catman automatically caches formatted man pages - in .../cat1, .../cat2 etc while the unformatted pages remain in .../man1, .../man2 etc. 

Yes, the other man formatter does just this. I think it should be pretty easy to add this capability to the current man command.

> nroff formatting could be run on the cross host and only the preformatted pages passed on to the ELKS image. Very similar to the current setup, but with the advantage of manpage portability and full nroff -man macro support.

This is certainly an option, and could be the only option, depending on how we select a set of existing man pages. I don't think we should include modern man pages - modern Linux applications have way too many options which aren't implemented in ELKS, and editing these pages down could be as large a task as manually writing pages, it seems.

>  It would require a version of nroff to be part of the cross toolset, which I don't think is much of a challenge.

We will have to see how it hard it might be to get nroff etc running on macOS as well as Linux... if homebrew has something, that will help. And we could keep already-formatted pages checked in to the distribution so that building ELKS doesn't require nroff for systems that don't have it.

> Any thoughts ??

In summary, it seems to me the path forward depends on where we select a set of man pages to start with from. I've been wondering whether there are any very early sets of ELKS applications somewhere else, perhaps with man pages, but haven't found any. I think using a very simple set like UNIX v6, and enhancing from there, would possibly be best, as they'd be small, use less complex nroff commands (although possibly not compatible with our `man`), and fit on a small ELKS distribution. But I'm not sure of that.

Thank you!

Mellvik:
Thank you @ghaerr.

I realize this may seem like a small issue in the overall picture. OTOH - minimizing the job of 'porting' man pages will lower the threshold to get them across and adapting them.


> This approach has several advantages, speed of course, as well as the ability to control how the output appears (bold, reverse video, colors, etc). I rewrote some code that adds bold and reverse video, but would like to consider colors in the future.
> 
Agreed, and I haven't tested whether nroff outputs exactly the same codes so that the mods to more will prevail, I'll look at that later.
> ELKS man pages have a special markup - very simplified nroff/troff commands that get rendered directly by the man command
> 
> I wasn't aware that these are using a special "simplified" markup - I figured we'd be better off with simplicity than having complex command requirements, especially if we have to write man pages ourselves. Can you detail a bit more about how this differs from modern man page sources?
> 
From my brief look at the current man page sources, it seems what they have may be nroff compatible, but limited to markup commands that start in column 0. Very simple but a pain to read, write and edit.
Of course I'm colored by having written and edited lots of manpages way back.
> Also, it turned out that selecting Robert's man command worked well with the existing few man pages (likely written by Robert himself, and several man pages (e.g. cron) don't display well at all.
> 
> There is however, one disadvantage: It means that importing 'standard' (nroff -man) manpages and adapting them for ELKS is not possible.
> 
> I suppose you're probably correct here. And I am thinking about where we might get a set of man pages to start with. I checked "dev86" which Robert contributed heavily to, but didn't find anything. I was thinking we should try to keep the man pages simple (as we don't really have much space to keep them outside of hard disk?!). Thus, determining which set of nroff/troff macros we might want to use is a big issue. Could we use UNIX v6 commands, for instance?
> 
The macro selection is simple. The -man macros were made specifically for manpages, and that what thousands of manpages out there assume. That said there are many versions of the man macros, and my suggestion is to start off with V7 - which isn't all that different from V6, just slightly more 'populated' - the man pages I mean - with options.
> It is a significant job to adapt them to the simplified markup.
> 
> I can imagine that doing this by hand would be a huge job, depending on where we select a set of already-written pages from.
> 
> somewhat similar to what catman does on Unix and Linux. In short, catman automatically caches formatted man pages - in .../cat1, .../cat2 etc while the unformatted pages remain in .../man1, .../man2 etc.
> 
> Yes, the other man formatter does just this. I think it should be pretty easy to add this capability to the current man command.
> 
> nroff formatting could be run on the cross host and only the preformatted pages passed on to the ELKS image. Very similar to the current setup, but with the advantage of manpage portability and full nroff -man macro support.
> 
> This is certainly an option, and could be the only option, depending on how we select a set of existing man pages. I don't think we should include modern man pages - modern Linux applications have way too many options which aren't implemented in ELKS, and editing these pages down could be as large a task as manually writing pages, it seems.
> 
Agreed, using V7 (or V6) is a much more reasonable starting point. I moved tons of man pages to Venix (v7) a while back. The tool compatibility across systems is really good.
> It would require a version of nroff to be part of the cross toolset, which I don't think is much of a challenge.
> 
> We will have to see how it hard it might be to get nroff etc running on macOS as well as Linux…
> 
Actually, this is probably a problem only on Windows, which doesn't run the tools anyway. All the other platforms have nroff (groff). All we have to do is to make sure the macro def file is the correct one (we'd just edit up our own I guess, not that big of a deal).
>  And we could keep already-formatted pages checked in to the distribution so that building ELKS doesn't require nroff for systems that don't have it.
> 
> 
Good point.
> Any thoughts ??
> 
I'll take a deeper dive into this late next week when I'm out of the woods on other tasks.
BTW - If my first impression is right, the current ('special markup') man pages should work just fine with nroff.

—M

ghaerr:
> minimizing the job of 'porting' man pages will lower the threshold to get them across and adapting them.

Completely agree.

> Of course I'm colored by having written and edited lots of manpages way back.

That's great - as I'm not that familiar.

> it seems what they have may be nroff compatible, but limited to markup commands that start in column 0. Very simple but a pain to read, write and edit.

I wasn't aware that newer markup commands can start within text copy. That would be great, as I find the current pages very hard to read and edit. I think I can convert our `man` to accept commands directly within text copy. Does it just need to match the particular two-character macro name, preceded by '.', and just leave it alone if not matched?

> Agreed, using V7 (or V6) is a much more reasonable starting point. I moved tons of man pages to Venix (v7) a while back. 

Great - I'll leave it to you for the decision as to starting with V7 vs V6 man pages and macros.

> All we have to do is to make sure the macro def file is the correct one (we'd just edit up our own I guess, not that big of a deal).

After some research, please post the macro file you think we'd be best using, and I can work on getting `man` to format correctly. I suggest the simpler the better, since we're formatting to text-only (console) output, there aren't that many options other than color and bold/reverse (e.g. no font change) that make a difference.

Go ahead and open a new issue when you've found time to look more into this, and we'll take it from there. Thank you!

ghaerr:
@Mellvik,

Some more info: Thinking about the problem of having a bunch of uncompressed man pages on the system, which we really need to work on floppy systems which don't have much space (I freed up about 100K removing relatively useless binaries last round), I looked into streaming compress/decompression programs (we can't use the executable decompressor as it's non-streaming).

The good news is that we have `decomp16` (recently removed from distro) which will decompress files created using host `compress` (widely available). The bad news is that decomp16 requires a 256K table to decompress anything. The Minix OS version, which we're using, breaks this down into either 4 fork() calls or 4 separate runs of decomp16, each of which require 64K memory (which wouldn't be able to run if networking is running, for instance), and uses 4 process slots, which sometimes aren't available.

I was able to work around this by running decomp16 4 separate times using a temp file each pass, but that'd be so slow on floppies as to be unusable. So we don't have viable decompression application.

All more things to consider, depending on the ultimate number of man pages that might get written...

Mellvik:
Valid point, @ghaerr.

My first thought is that
- there will be relatively few manpages (commands and file formats only, library and systems calls don't make any sense on the platform itself since development is elsewhere)
- The formatted pages will be quite small for most commands (notable exceptions like vi, ash, sash, …)
- reasonable text compression can probably be achieved with venerable 'compress' which IIRC is very lightweight.
- I suggest we start off w/o compression, and determine which compression tool to choose when we see what the requirement is.

My 2C…

Thank you.

—M

ghaerr:
Sounds good!

> reasonable text compression can probably be achieved with venerable 'compress' which IIRC is very lightweight.

No, that's the compressor and format used by `decomp16` which I described above, and is still too heavyweight to decompress, unfortunately.

Mellvik:
> 
> reasonable text compression can probably be achieved with venerable 'compress' which IIRC is very lightweight.
> 
> No, that's the compressor and format used by decomp16 which I described above, and is still too heavyweight to decompress, unfortunately.
> 
OK, let's keep looking for something suitable.

Software tools' 'compact' is better than nothing in that it eliminates ('compresses') repeated characters. Very simple, possibly a starting point.

-M
> —

ghaerr:
@Mellvik,

I was looking through some MINIX 3 code and realized it already has a great set of manual pages, and ELKS already includes a number of MINIX user applications in our current distribution, and many others are very close. Thus, this could be a great place to start, rather than V6 or V7. The man pages are very small, which is also nice.

Here's a [link](https://github.com/0xffea/MINIX3/tree/master/man) to the MINIX 3 man pages on GitHub, tell me what you think about the macros. I see they're using a direct method of changing fonts (\fB for bold, \fR for regular, etc), which also work with our formatter. (They also sometimes use a variant with two backslashes instead of one above, which doesn't work though, but could be easily added I think). Our small number of existing man pages use a macro instead of that method to bold text.

Testing the minix man page entries using `man ./kill.1` (for instance) on macOS displays properly, which is nice for checking without having to run ELKS.

I pulled some of the minix man pages over to ELKS and many mostly display correctly already, with the exception of a few macros not yet identified and the double-backslash font specifiers.

With an existing full set of commands very closely matching ELKS, we could add a few enhancements to our man (such as your suggestion of allowing macros in any column) to more easily allow editing, but the job would probably be a lot smaller to begin with.

Thank you!

PR #1177, opened 2022-02-21
Allow override of toolchain

ConiKost:
This change allows to set own version packages for toolchain,
in order to override the default ones. This can be done via Make.local.

If no settings are set, default is being used.

Signed-off-by: Conrad Kostecki <conikost@gentoo.org>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@
 # Normal output and testing dirs
 #
 
+Make.local
 cross
 target
 
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -5,6 +5,7 @@ $(error TOPDIR is not defined)
 endif
 
 include $(TOPDIR)/Make.defs
+-include Make.local
 
 .PHONY: all prune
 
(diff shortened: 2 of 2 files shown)


ghaerr:
Hello @ConiKost, hello @tkchia,

I understand what you are proposing here, however, I am not convinced we need this feature, which possibly complicates ELKS stability by providing for an easy way for developers to use a non-preferred toolchain, whose version would then not be known by maintainers. We have spent quite a bit of time increasing the stability of ELKS in the last few years, and quite a bit of time has been spent trying to eliminate, rather than increase, variables related to debugging. Thus, I don't think I'm in favor of it, especially since a local saved copy of tools/Makefile would work equally well, until we decide to update the toolchain for the developers.

That said, our toolchain maintainer is responsible for changes to tools/Makefile. @tkchia, do you have an opinion on this?

Thank you!

ConiKost:
> I understand what you are proposing here, however, I am not convinced we need this feature, which possibly complicates ELKS stability by providing for an easy way for developers to use a non-preferred toolchain, whose version would then not be known by maintainers. 

Maybe I am wrong, but you already can just put a different version in the ```cross``` folder and no-one would know about it, when using a non default toolchain.

> Thus, I don't think I'm in favor of it, especially since a local saved copy of tools/Makefile would work equally well, until we decide to update the toolchain for the developers.

The only downsize is, I have to keep it locally ignored. Even, when toolchain will updated in some future, you never know, if we will run again in some sort of GCC problems.

PR #1178, opened 2022-02-24
Add man pages from MINIX 3, revise man page installation

ghaerr:
I found that MINIX has lots of man pages that are (almost) compatible with our `man` formatter.

This PR adds a large initial batch of them, for subsequent editing and correcting for ELKS. The current ELKS man pages, (around 10) are also added to the new man page directory elkscmd/man/*. The man pages can be viewed from the host system using "man ./ps.1", without having to start ELKS.

The MINIX man1 pages defined two unused .de (define) macros, these were removed, as `man` does not support defining macros. 

I feel pretty confident that it will be fairly easy to update these pages for ELKS. I plan on another PR with the first major edits.

The man pages need more than 1440K disk space to install. I am testing using 2880K on QEMU.


Mellvik:
Thank you @ghaerr,

sorry about the delay, I'm just getting my head above the water so to speak.

This is great work, and I agree with you, it's the perfect fast path to ELKS manpages.

I've refreshed my memory on man pages, macros, catman (preformatting) and more - very useful, occasionally surprising :). 
As you have already noted, man macros are generally column one 'dot-commands' (the high level language). The \f and similar that some times appear are troff/nroff commands and akin to assembler (which the macros translate into).

Since I brought the subject up before: Looking at the pages produced by catman (reformatted) [I pulled out a running BSD2.11 and a Venix system] they are generally larger than the raw pages, mostly because they have lots of spaces and blank lines. Consequently they are extremely fast as there is no processing at all - like the name says, just 'cat'. So in order to use preformatted pages on ELKS, we'd need some compression – such as the compact I referred to earlier.

For now though, I think we're best off following the path you started, and using the 'embedded formatter' in the ELKS man command. [As a curiosity, there is a Linux man page how-to on the net (http://www.schweikhardt.net/man_page_howto.html <http://www.schweikhardt.net/man_page_howto.html>) (1995). Another curiosity - MacOS currently has 3300 pages, Debian has 4600 - of which 150 are git related … .]

A distinct disadvantage to not having a macro definitions file is that the ability to set headers and footers in one place goes away. Typically used to identify the host system and version. Instead of putting that in every man-page and potentially having to sed or awk the whole shebang, the man command could have a header file in which such info was defined. Thus removing it from the pages and making the smaller.

It seems the ELKS man command understands a quite complete set of the BSD man macros and the most important inline commands (\fX font variant changes in particular) much better than I initially thought. So I think we should stick with that for compatibility reasons. If we encounter unsupported commands/macros (deemed important), it should be easy to add them I suppose.

Thanks again, I'll contribute to the edit and creation of man pages as we proceed. Starting with fixing the cron.8 page :-)

—M

ghaerr:
> I've refreshed my memory on man pages, macros, catman (preformatting)
> man macros are generally column one 'dot-commands' (the high level language). The \f and similar that some times appear are troff/nroff commands and akin to assembler 

There's going to be no need for pre-formatted pages, as interpreting the small number of dot commands and converting them into formatted pages for the ELKS (very basic 80x25) console is extremely fast, and the `man` page previously written for ELKS works very well already, now that I've read and understand all of it. It supports \f font specifiers, as well as *(xx translation characters, although no translations are currently being performed. Of course, the translations will be quite limited and subject to the IBM PC 437 codepage BIOS font, but still useful. 

My previous comments about \\f etc turned out to be incorrect - those were double escaped parameters within macro definitions, all of which turned out to be unused.

We could still use some compression, but need a small LZW-style compressor that works within 64K - I haven't found one yet. Compressing sequences of identical characters won't help much. The entire man page set is around 400K, I think.

> A distinct disadvantage to not having a macro definitions file is that the ability to set headers and footers in one place goes away. Typically used to identify the host system and version.
 
That problem will be very easily solved by having `man` display the host system and version, without the need for even a separate macro file. It's already on my list.

> I'll contribute to the edit and creation of man pages as we proceed. Starting with fixing the cron.8 page

Thanks. This batch of MINIX pages already includes entries for crontab.1, crontab.5 and cron.8. In the case where ELKS has/(had) a duplicate, which is the case for about 10 commands, I renamed the previous ELKS version with a '2' suffix, e.g. cron2.8. I plan on looking at these shortly.

Mellvik:
> I'll contribute to the edit and creation of man pages as we proceed. Starting with fixing the cron.8 page
> 
> Thanks. This batch of MINIX pages already includes entries for crontab.1, crontab.5 and cron.8. In the case where ELKS has/(had) a duplicate, which is the case for about 10 commands, I renamed the previous ELKS version with a '2' suffix, e.g. cron2.8. I plan on looking at these shortly.
> 
Sounds good, and no, I didn't see there was a cron.8 in your batch. I pushed an edit of the previous one with some other fixes a few minutes ago, which failed with some odd azure error. 

I'll have to look at that. Maybe a temporary system failure.


—M

PR #1179, opened 2022-02-24
[cmds] Bugfixes in sash, fix cron man page

Mellvik:
Misc:
- finally a long promised history fix for sash
- added the new -a option to `mount` inside sash
- changed macros in cron.8 man page

### DIFF
--- a/elks/fs/msdos/inode.c
+++ b/elks/fs/msdos/inode.c
@@ -100,7 +100,7 @@ static struct super_block *msdos_read_super(struct super_block *s, char *data,
 	struct msdos_boot_sector *b;
 	struct buffer_head *bh;
 	long total_sectors, data_sectors;
-	cluster_t cluster;
+	cluster_t cluster, max_clusters;
 	long total_displayed, free_displayed = 0;
 	int fat32;
 
@@ -162,7 +162,7 @@ static struct super_block *msdos_read_super(struct super_block *s, char *data,
 		*((unsigned short *) b->sectors) : b->total_sect;
 	data_sectors = total_sectors - sb->data_start;
 	sb->clusters = sb->cluster_size?  data_sectors/sb->cluster_size : 0;
-	sb->fat_bits = fat32 ? 32 : sb->clusters > MSDOS_FAT12 ? 16 : 12;
+	sb->fat_bits = fat32 ? 32 : sb->clusters > MSDOS_FAT12_MAX_CLUSTERS ? 16 : 12;
 	sb->previous_cluster = 0;
 	unmap_brelse(bh);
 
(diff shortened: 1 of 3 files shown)


ghaerr:
> I'll have to look at that. Maybe a temporary system failure.

Yes, probably.

However, the cron.8 edit is going to cause a conflict with my outstanding MINIX man page PR #1178, unfortunately. I am worried that if I commit yours first, it may screw up that PR, which has 192 files, and I'm no git expert. Can I ask you to remove the cron.8 fix from this PR, and add it after #1178. It renames the file from elkscmd/cron/cron.8 to elkscmd/man/man8/cron2.8.

Thank you!

ghaerr:
Also may I ask what the ESC check is for in sash? Is this because one might mistakenly type ESC before a command then it can't find it? It only checks the first character, not the entire command string, and also doesn't check for other possibly illegal characters. Perhaps it should say "ESC in command". (Also, buf[0] would be less confusing than \*buf since buf is an array, not a pointer).

Mellvik:
Thank you for the code review @ghaerr.


And yes, it's to catch arrow key pushes - inadvertent or just forgetting that there is no arrow key command line editing.

In the general case such pushes doesn't matter, it's just an illegal command. This code was added because when using command history, every such illegal command (blank, just control chars) wastes a position in the history buffer, and there are too few of them already.

I don't think mentioning ESC is a positive change. To most users that will be confusing, as in 'what the he.. is ESC?'.
The test included more characters at first ('['), but it doesn't make sense. If the first char is ESC, it's an illegal command anyway. And testing for values below ' ' instead may be useful, but then again, TABs are allowed…

—M

ghaerr:
> it's to catch arrow key pushes - inadvertent or just forgetting that there is no arrow key command line editing.

I understand now - it wasn't for the ESC key, but rather when an arrow key is pressed, it's ANSI sequence always starts with ESC. Ok.

> I don't think mentioning ESC is a positive change. 

Ok, got it.

Mellvik:
@ghaerr,
I removed the cron.8 file from this PR and git is still not happy. Admittedly I have no clue as to what the next step may be. Do you have the greyed out ResolveConflicts button too?

ghaerr:
> Do you have the greyed out ResolveConflicts button too?

Yes.

> Admittedly I have no clue as to what the next step may be. 

How did you try to remove cron.8? Did you undo the changes, and recommit? Perhaps try directly removing the file from the commit, using something like:
```
git checkout misc
git checkout elkscmd/cron/cron.8. <-- resets cron.8 to original
git commit -a
git push
```
If that doesn't work, unfortunately I would suggest closing this PR and submitting a new one. Sorry.

Mellvik:
Actually this is what I did.

The cron.8 file var reverted to its original and re-committed. Unfortunately (at least this is what I think must have happened) your PR was merged in between, and the reverted cron.8 is different still.

Yeah, I may have to kill it again. I'll see what I can find out.

—m

ghaerr:
> your PR was merged in between, and the reverted cron.8 is different still.

I see. My PR moved cron.8 to elkscmd/man/man8/cron2.8, so that could very well be the problem.

Mellvik:
OK; I finally found a way to do this, and I'm including the procedure for reference:
- Make sure the PR is open
- `git checkout misc` this is the actual branch
- `git branch temp` to keep a copy, I actually made copies of the involved files too
- `git reset --hard HEAD~2` 2 because the PR had two commits
- `git log` just to see that the new HEAD point looks OK
- `git push -f` which pushes an empty PR, `git` discovers that and merges/closes automatically.

PR #1180, opened 2022-02-24
[kernel] Move autosync from kernel to /bin/init

ghaerr:
Moves kernel buffer auto-sync code out of idle task into /bin/init. Takes same sync= /bootopts parameter, so no change to the user. If it is decided that auto-sync should be turned on without /bootopts, this is also easily possible by setting a define in init.c.

It was realized that the idle task cannot be actually call sync, the way a user program does, as there is a possibility that the kernel sync code may sleep the task, which is not allowed for the idle task. So, sync() is now called from /bin/init, which uses an alarm() signal for its timer callback.

@Mellvik, the auto-sync is now in application code, so it can be played with more easily, if desired. Unfortunately, there is no easy way to determine the number of dirty (need-to-be-written) system buffers, so a fancy algorithm isn't (yet) possible. We couldn't easily do this in the kernel either, without an expensive count through all the buffer headers every second or so either, so an improvement is yet unsolved. Perhaps the timer should be set equal to the max amount of time considered acceptable for a user to wait for writing disk sectors at full speed. That will certainly be less than 30 seconds, and the sectors need to be written eventually anyways and doesn't slow down an end-to-end copy. Setting to a value like 10 seconds may be help, and also allow for more interactivity (echo) in between intensive writing.

Also fixes a few small other issues:
- C library actually dereferenced NULL pointer when calling printf("%s", 0).
- Added kernel printk in the case that multiple signals are accumulated, waiting for application callback; this is not currently supported, but no indication was given to user. (May not ever happen).
- /bin/init uses new paths.h header for common system file locations.
- Minor cleanup in FAT code.

### DIFF
--- a/elks/arch/i86/kernel/signal.c
+++ b/elks/arch/i86/kernel/signal.c
@@ -70,6 +70,9 @@ int do_signal(void)
 	    //debug_sig("Stack at %x\n", currentp->t_regs.sp);
 	    *sd = SIGDISP_DFL;
 	    debug_sig("SIGNAL reset pending signals\n");
+	    if (currentp->signal)
+		printk("SIGNAL(%d) ignoring signal (mask=%s)\n",
+		    currentp->pid, currentp->signal);
 	    currentp->signal = 0;
 
 	    return 1;
--- a/elks/fs/msdos/inode.c
+++ b/elks/fs/msdos/inode.c
@@ -172,7 +172,7 @@ printk("FAT: me=%x,csz=%d,#f=%d,floc=%d,fsz=%d,rloc=%d,#d=%d,dloc=%d,#s=%ld,ts=%
 	sb->data_start, total_sectors, b->total_sect);
 
 	/* calculate max clusters based on FAT table size */
-	max_clusters = (unsigned long)sb->fat_length *
+	max_clusters = (cluster_t)sb->fat_length *
 		(SECTOR_SIZE_SB(s) * 8 / sb->fat_bits) - 2;
 	/*
 	 * Allow disks created with too small a FAT to be mounted, but limit free space.
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -29,7 +29,6 @@ int root_mountflags = 0;
 #endif
 int net_irq, net_port;
 static int boot_console;
-static int sync_interval;
 static char bininit[] = "/bin/init";
 static char *init_command = bininit;
 
@@ -71,17 +70,11 @@ void start_kernel(void)
     /*
      * We are now the idle task. We won't run unless no other process can run.
      */
-	jiff_t wake = jiffies;
     while (1) {
         schedule();
 #ifdef CONFIG_IDLE_HALT
         idle_halt ();
 #endif
-		/* sync all devices if sync= set in /bootopts */
-		if (sync_interval && jiffies >= wake) {
-			fsync_dev(0);
-			wake = jiffies + sync_interval * HZ;
-		}
     }
 }
 
@@ -328,10 +321,6 @@ static int INITPROC parse_options(void)
 		}
 		if (!strncmp(line,"bufs=",5)) {
 			boot_bufs = (int)simple_strtol(line+5, 10);
-			continue;
-		}
-		if (!strncmp(line,"sync=",5)) {
-			sync_interval = (int)simple_strtol(line+5, 10);
 			continue;
 		}
 		
--- a/elkscmd/sys_utils/init.c
+++ b/elkscmd/sys_utils/init.c
@@ -34,6 +34,7 @@
 #include <memory.h>
 #include <errno.h>
 #include <time.h>
+#include <paths.h>
 
 #define USE_UTMP	0	/* =1 to use /var/run/utmp*/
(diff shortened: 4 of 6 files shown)


Mellvik:
Thank you @ghaerr.

> 
> @Mellvik <https://github.com/Mellvik>, the auto-sync is now in application code, so it can be played with more easily, if desired. Unfortunately, there is no easy way to determine the number of dirty (need-to-be-written) system buffers, so a fancy algorithm isn't (yet) possible. We couldn't easily do this in the kernel either, without an expensive count through all the buffer headers every second or so either, so an improvement is yet unsolved.
> 
Maybe just take the easy route, a la the mechanism I suggested yesterday, and just set a max threshold for the syncing - then repeat every second until there is nothing to sync. 
- the 30 sec (or whatever) sync timer kicks in
- sync syncs max X blocks, returns a status to indicate whether there is more
- a second later sync another X blocks etc. 
- until done.

We can experiment with the value of X in order to find something that makes sense and appears acceptable. If we want to get fancy, change SYNC=30 to SYNC=30,10 - or something, where the latter value may be optional.

I'll do some testing on this later today.

BTW - I realize this issue may seem very minor. Who cares about a small delay very once in a while - the LED blinking or the floppy making noises? It's obvious what's going on. Until it isn't. In my case I'm hardly ever in the same room as the machine. That's a different experience entirely. The system appears to hang. If typing at the shell prompt or editing a file, 2 seconds 'dead' time is very noticeable, 10 secs is when you start wandering what happened ...


-M

ghaerr:
> The system appears to hang. 
> 10 secs is when you start wandering what happened ...

Agree we should do something about this if possible. I'm wondering however, given the fact that we're (currently) subject to losing control of the kernel echo and interactive responsiveness because of our reliance on BIOS disk I/O, whether any 'sync' changes will actually fix this...:

During what process (exactly) are you losing responsiveness? Copying any large file to disk? Sorry for asking possibly again, but is the problem perhaps that your system is going to become non-responsive during *any* longer floppy disk writes?

Is that the issue, then, rather than being related to XMS buffering, which just possibly delays the disk writes, is the issue that the system will always be unresponsive during multi-sector writes of some length? If so, your idea of writing only so many blocks at a time could work, but I think that might delay a full disk copy write for a very long time. Thus my question of when exactly does this seem to be occurring?

Finally, another thought would be some kind of option that would say, send dots (.) to the serial console when disk I/O is being performed. Or, the dots could be packaged inside some special escape sequences that would put them up in the right corner of the screen, like the rotating console icon, and then change the cursor back to its original location...

Mellvik:
Sorry about the late response to this one, @ghaerr -

> Agree we should do something about this if possible. I'm wondering however, given the fact that we're (currently) subject to losing control of the kernel echo and interactive responsiveness because of our reliance on BIOS disk I/O, whether any 'sync' changes will actually fix this…:
> 
This sounds very serious - much more so than I intended. Let's keep the perspective:
- The huge cache is a BIG performance PLUS on real HW that supports it.
- The added sync is important.
- The lagging is a natural consequence.
- The lagging has always been there, but the huge cache is making it much more noticeable.
- IMHO it's better to accept it as it is than to initiate complicated measures to alleviate it. My 'idea' about character echo doesn't originate here, it has been there all along - on my wish list, because I always liked the Unix/Linux way of doing this.
> During what process (exactly) are you losing responsiveness? Copying any large file to disk? Sorry for asking possibly again, but is the problem perhaps that your system is going to become non-responsive during any longer floppy disk writes?
> 
That's right. As pointed out above, it's the huge cache that's exacerbating something that has always been there. As to what I do - regular use. Like ftping in a couple of files, then running som file transfers - testing other stuff. Sooner or later the data will be flushed by the sync, and it will be noticeable - proportional to the # of dirty buffers.
> Is that the issue, then, rather than being related to XMS buffering, which just possibly delays the disk writes, is the issue that the system will always be unresponsive during multi-sector writes of some length? If so, your idea of writing only so many blocks at a time could work, but I think that might delay a full disk copy write for a very long time.
> 
Spot on, thus (again - and I keep repeating myself) we need to get raw device access to keep the buffers out of the loop.
> Finally, another thought would be some kind of option that would say, send dots (.) to the serial console when disk I/O is being performed. Or, the dots could be packaged inside some special escape sequences that would put them up in the right corner of the screen, like the rotating console icon, and then change the cursor back to its original location…
> 
Thank you and appreciated, but no thank you. This is not a serious enough problem to go to such lengths. And - BTW - to me, knowing the reason why the system lags, it's an annoyance and I'll get used to it. And one of these days I'll get off of using the floppies as root. I'm holding back on that one in order to keep the delays, they are useful for testing networking (retransmits etc.). Kind of weird priorities, but there you go. That said, using HD only doesn't eliminate the lagging, but reduces it significantly.

Straying off topic, but somewhat related: I think we're coming back to non-bios IO at some point. Now that BIOS IO is very stable and available to the hardware that needs it, it may be worthwhile to take another serious look at going native - IDE and 765 support, which covers most PCs AT/286 and newer. Both are reasonably simple, IDE in particular. That's for another time.

—M

Mellvik:
Updated the comment above, BUG should (obviously?) read BIG :-)

PR #1182, opened 2022-02-25
[kernel] Add ustatfs syscall, allows mount to display free space and mount points

ghaerr:
Adds new `ustatfs` system call, which allows programs to:
- Query each mounted filesystem
- Return type of mounted filesystem (MINIX, FAT, ROMFS)
- Return total and free blocks of filesystem
- Return total and free number of files nodes (inodes) on filesystem
- Return mount point of filesystem

This then allowed `mount` to be enhanced to display all mounted filesystems and their mount points.

For mount speed, the kernel no longer calculates FAT filesystem free space at mount time. Instead use `mount` (or soon, an update to `df` for FAT filesystems).

Here's a screenshot showing a 2880K MINIX floppy ELKS boot, then auto-mounting a 1440K FAT floppy, and running `mount`:
Screen Shot 2022-02-24 at 10 56 32 PM

The displayed mount information will be improved to show the filesystem type, along with `df` enhancements for FAT filesystems in a subsequent commit.

### DIFF
--- a/elks/arch/i86/kernel/syscall.dat
+++ b/elks/arch/i86/kernel/syscall.dat
@@ -89,7 +89,8 @@ insmod		65	1	- Removed support for modules
 fchown		+66	3
 dlload		+67	2	- Removed support for dynamic libraries
 setsid		+68	0
-sbrk        +69 2	* Legacy number from Linux
+sbrk		+69	2	* Legacy number from Linux
+ustatfs		+70	2
 knlvsn		+74	1
 #
 # From /usr/include/asm-generic/unistd.h
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -28,7 +28,6 @@
  *			for details.
  */
 
-#include <linuxmt/vfs.h>
 #include <linuxmt/types.h>
 #include <linuxmt/errno.h>
 #include <linuxmt/fcntl.h>
--- a/elks/fs/minix/bitmap.c
+++ b/elks/fs/minix/bitmap.c
@@ -17,8 +17,6 @@
 
 #include <arch/bitops.h>
 
-#ifdef BLOAT_FS
-
 static char nibblemap[] = { 0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4 };
 
 static unsigned short count_used(struct buffer_head *map[],
@@ -41,8 +39,7 @@ static unsigned short count_used(struct buffer_head *map[],
 	    numbits = 0;
 	}
 	for (j = 0; j < end; j++)
-	    sum += nibblemap[bh->b_data[j] & 0xf]
-		+ nibblemap[(bh->b_data[j] >> 4) & 0xf];
+	    sum += nibblemap[bh->b_data[j] & 0xf] + nibblemap[(bh->b_data[j] >> 4) & 0xf];
 	unmap_buffer(bh);
     }
     return sum;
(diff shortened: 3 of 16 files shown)


Mellvik:
Thank you @ghaerr,

I've commented on turnaround time before, will not do it again - other than a big nod. can't wait to test!

Off topic, but somewhat related (and my memory tells me we've touched this before): Floppy probing. When a 1.2M/360k drive sees a 360k floppy, there are 5 or maybe 6 probes in order to check the format - full seeks etc. This takes a long time - compared to DOS which does only a single probe. Is this somehow hidden in the BIOS or something that the driver controls?

—M

ghaerr:
> Floppy probing. When a 1.2M/360k drive sees a 360k floppy, there are 5 or maybe 6 probes in order to check the format - full seeks etc.
> Is this somehow hidden in the BIOS or something that the driver controls?

Our BIOS driver controls this completely, and its next on my list to improve. There are a number of possible improvements, but the simplest is it seems the current driver tries reading cylinders 40 and 80, and then sectors 8, 9, 15, 18 and 36, trying to find the first failure. It then sets the disk type to the value *before* the failure. This is quite longwinded, but the big problem I see is that it retries FIVE times for each of these!

Can you try changing the following code, which will reset the retries to 0, in elks/arch/x86/drivers/block/bioshd.c:
```
static int read_sector(int drive, int track, int sector)
{

/* i took this code from bioshd_open() where it replicates code used
 * when new floppy probe is used
 */

    register int count = 1; <---- change this line from MAX_ERRS (5) to 1

    set_cache_invalid();
    do {
        /* BIOS read sector */

        BD_AX = (unsigned short int) (BIOSHD_READ | 1); /* Read 1 sector  */
```
The driver also does a BIOS disk reset in between retries, we'll see whether that needs to be removed also. It may need to stay, based on BIOSes, but could also be quite slow.

ghaerr:
> When a 1.2M/360k drive sees a 360k floppy, there are 5 or maybe 6 probes

Also, and this is important - is the floppy MINIX or FAT? And if FAT, is it an ELKS FAT disk, meaning bootable? I ask because the BIOS driver, before reading the sectors I mentioned above, first reads the boot sector and tries to get the ELKS BPB (special stamp that ELKS puts on the boot sector that specifies disk CHS). If that is not found, then it goes into the other probe function.

I am also thinking that ELKS should just recognize regular (non-ELKS) FAT floppies, those without an ELKS BPB, by reading the boot sector and interpreting the MSDOS BPB, instead of slow probing. This would speed things up greatly.

In summary, it is important to know what type floppy you're typically reading that is slow in probing...

Thank you!

Mellvik:
@ghaerr,
I realize this is kind of moot by now, but I'll answer anyway.
The 360k floppies are mostly fat, occasionally blank. I haven't noticed anything that made me suspect that the probing actually tried to decipher the content.
I assumed (not thinking that the bios in between might make that impossible) that the probing would simply look for an error free read, and nothing else.

Anyway, looking forward to testing the changes.

Thank you!
-m

ghaerr:
Thanks @Mellvik.

> The 360k floppies are mostly fat, occasionally blank. I haven't noticed anything that made me suspect that the probing actually tried to decipher the content.

The way the probing used to work was to first check for an ELKS "signature" containing CHS values in the last few bytes of the first sector. If that was found, those CHS values were used. Otherwise, a lengthy probing process was started. The ELKS signature would only be found on ELKS boot disks, of course (FAT or MINIX).

In the new method from #1184, the same first check for an ELKS signature is checked as before. If not found, then it checks for an MSDOS-formatted disk from the same already-read buffer, which would contain the head and sector values, and the MSDOS "media byte" is used to guess the cylinders. Finally, (for blank disks), an improved probing process is started. This should be much faster.

> I assumed (not thinking that the bios in between might make that impossible) 

The BIOS itself may do any number of things to keep up its internal tables, but these are usually delayed until the first sector of a floppy is read, and then it update its tables for its own use with the FDC controller. 

> the probing would simply look for an error free read, and nothing else.

The problem with the previous probing method is that it tried 5 times for an "error free read", which was pretty bad, time-wise. It now tries only once, and skips a check that wasn't needed. So it should be much faster.

The probing messages are updated to make it easier to understand what ELKS is thinking. I'm anxious to see what you find on real hardware!

Thank you!

PR #1184, opened 2022-02-25
[kernel] Improve floppy probing performance

ghaerr:
Greatly improves ELKS floppy probing performance by:
- Recognizing DOS 3.31+ floppies not previously seen by ELKS, skipping probe
- Not probing first track/sector as previously done (unneeded)
- Not retrying 5 times as previously done (no retries)

Removes previous double probing message, and improves message showing DOS or ELKS format when recognized.

There is some leeway in which exact format DOS floppies we might recognize. Testing with many different floppy types and filesystems is encouraged.

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -413,7 +413,7 @@ static int read_sector(int drive, int track, int sector)
  * when new floppy probe is used
  */
 
-    register int count = MAX_ERRS;
+    int count = 1;		/* no retries on probing*/
 
     set_cache_invalid();
     do {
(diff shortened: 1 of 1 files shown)


Mellvik:
Thank you @ghaerr, this is an important improvement.

IMHO the next step should be to remove all FS related testing from the low level probing. Leave the FS level to the FS-level so to speak, which borders into the raw device discussion again.

The probing should verify the prescence of the media and - if relevant - the density.
One of the great (and convenient) usages of ELKS is to manipulate raw floppies - regardless of contents. I'm copying or duplicating Venix floppies, SCO floppies, CP/M floppies, Solaris floppies - raw image files. So raw device probing to verify media is necessary, any FS-level probing is superflous and possibly detrimental: Old floppies are often hard to read. Repeated reads some times succeeds - which is aided by not having anything in between to interpret errors, just reporting them. The simpler and more direct, the better - and not touching the caching/buffering system. 
In some cases it would be useful to change the sector size in order to read/write floppies from odd systems, another usage that would be aided by direct low level access.
Right now - in my case - it's simpler to use Venix for some of these tasks because it has raw devices (and admittedly doesn't use the BIOS at all). And it uses different minor device numbers to tell the driver which format to use, sort of manual probing. Unfortunately :-) that machine only has a 3.5in floppy.

My 2 cents - actually 4 or 5 I guess.

--M

PR #1188, opened 2022-02-26
Add PC-98 msdos partition

tyama501:
This modification allows to mount/read/write pc-98 msdos partition under 1023 cylinders.

To read/write partition which has cylinders over 1023, bioshd.c needs to be modified that is the next step.

Thank you.


ghaerr:
Thank you @tyama501, glad you were able to get this all working!

PR #1191, opened 2022-02-27
[cmds] More man page updates

ghaerr:
Most man pages in section 1 and 8 are now formatting correctly. Still not touching cron entries, will do so after @Mellvik's PR.


ghaerr:
@Mellvik, I am thinking of making another pass through the mess of man pages. Would you like me to dig up the change you made to cron.8 from your closed PR and add it to the next round, saving you the trouble?

ghaerr:
> Would you like me to dig up the change you made to cron.8 from your closed PR and add it to the next round, saving you the trouble?

Actually @Mellvik , I just looked at your modifications and they use lots of macros that are not supported by our man. I will take a pass at updating it, looking at your previous changes, such that it works. In this way, many man pages that I have imported will be working, leaving only missing man pages to create.

I have crafted a way that will work for all ELKS man pages using very few macros, all standard. This will be documented shortly. The simplified approach works well when displayed on console as well as serial line to macOS Terminal.

Mellvik:
yes, please do that @ghaerr - 

I'm in this git rabbit hole right now, in addition to checking out a compress/decompress program (turns out I ported it to Venix aeons ago, should compile w/o problems on ELKS. It compresses binaries about 15%, text some 35%, too slow for man pages, but we've been missing a compression alternative, and this may just be it. Useful when tar eventually works (/bin/tar doesn't work at all, the tar builtin in sash works for reading).


—M

ghaerr:
Ok, I'll continue updating the man pages, I'll let you know when done.

I'm chasing down a couple compress ideas as well. If you have a compress.c source that compiles on 16-bit, I'd like to see it. ELKS has one as well, but it's not compiling, will be looking into that.

> /bin/tar doesn't work at all, the tar builtin in sash works for reading

Geez, I've made a note of that. If you'd like some help chasing down the problems with either tar (so we can select on or the other to go forward with), open a new issue with more details.

Mellvik:
> Ok, I'll continue updating the man pages, I'll let you know when done.
> 
> I'm chasing down a couple compress ideas as well. If you have a compress.c source that compiles on 16-bit, I'd like to see it. ELKS has one as well, but it's not compiling, will be looking into that.
> 
> 
This is BSD 'compact/uncompact' - not LZ but 'adaptive Huffman'. Minor changes IIRC - it's coming from the PDP11 (16bit).
> /bin/tar doesn't work at all, the tar builtin in sash works for reading
> 
> Geez, I've made a note of that. If you'd like some help chasing down the problems with either tar (so we can select on or the other to go forward with), open a new issue with more details.
> 
> 
Don't worry about tar, I've had it on my list for ages, will 'up' it on my list. I have a half finished BSD variant (extracts files for now, ran into some directory troubles on writing - but that was in 2017, I've learned a thing or two about elks since). It's convenient to know that sash tar works for extracting (BTW even the standalone variant is extraction only).

thank you @ghaerr,


—M

PR #1193, opened 2022-02-27
[ftp, ftpd] add in_connect for timeouts

Mellvik:
@ghaerr,
I suggest this one is left uncommitted until we figure out the timeout problem. Had to push it now to make sure my exit from the rabbit hole was clean.

-M

### DIFF
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -4,17 +4,13 @@
  * by Helge Skrivervik - helge@skrivervik.com - December 2021
  *
  *	TODO:
- *	- add support for multi-argument commands (e.g. <put sourcefile destfile> ...)
  *	- add missing commands (rmdir, mls, nlist, ...), 
  *	- ... create common function for commands that are alomost alike
  *	- fix trucation of file names in FAT, minix (warnings, name collisions)
  *	- lcd with no params should print current local directory
- *	- evaluate whether using select() is useful or just a complication
- *	... add ABORT support in PUT
+ *	- add ABORT support in PUT
  *	- handle server timeout (no activity): Unsolicited server input (code 421)
  *	- add '15632 bytes sent in 0.00 secs (27.5561 MB/s)' type message after transfer
- *	- validity check on server host name/address to avoid system hang
- *	- more testing of error conditions & error messages
  */
 
 #define BLOATED		/* fully featured if defined */
@@ -282,16 +278,15 @@ int get_reply(int fd, char *buf, int size, int dbg) {
 		while (*cp != '\n' && cp >= &buf[lb]) --cp;
 		cp++;
 		lb += l;
-		if (*(cp+3) != '-') {	/* end of continuation if blank */
+		if (*(cp+3) != '-') 	/* end of continuation if blank */
 			break;
-		}
 	}
 	buf[lb] = '\0';
 	lb = 0;
 
 	// cp points to the start of the last record,
-	// compare starus codes to see if we have an extra
-	// reply to set aside for the next call.
+	// compare status codes to see if we have an extra
+	// reply to save for the next call.
 
 	if (cp != buf) {
 		if (strncmp(buf, cp, 3)) {
@@ -895,7 +890,7 @@ int do_passive(int cmdfd) {
 
 	if (debug > 1) printf("Connecting to %s @ %u\n", ip, port);
 
-	if (connect(fd, (struct sockaddr *) &srvaddr, sizeof(srvaddr)) < 0) {
+	if (in_connect(fd, (struct sockaddr *) &srvaddr, sizeof(srvaddr), 10) < 0) {
 			perror("connect error");
 			return -1;
 	}
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -44,7 +44,7 @@
 
 #define	IOBUFLEN 1500
 #define BUF_SIZE 512
-#define CMD_BUF 80
+#define CMDBUF 80
 #define ADDRBUF	40
 #define NLSTBUF 1500
 #define MAXARGS	256	/* max names in wildcard expansion */
@@ -244,10 +244,11 @@ int get_reply(int fd, char *buf, int size, int dbg) {
  * Multiline control responses (typically at login), will some times arrive in several packets and must be
  * assembled properly.
  * This is not elegant, but it is very memory efficiant (abusing the *buf parameter for all the work).
+ * DO NOT call get_reply with a buf size of less than BUF_SIZ (512b).
  */
 
 int get_reply(int fd, char *buf, int size, int dbg) {
-	static char lbuf[CMD_BUF];
+	static char lbuf[CMDBUF];
 	static int lb = 0;
 	
 	char *cp;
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @Mellvik,

This looks fine, we can go ahead and commit this now, if you like, and debug in_connect separately.

I don't yet have any ideas as to why `in_connect` isn't working on ftp. A possible way to debug this would be to copy the elks/libc/net/in_connect.c code directly into you ftp.c, and add a printf() into the "alarm_cb" function, which is the SIGALRM callback function. That will at least show whether the alarm is firing or not (which interrupts connect normally).

If we commit this, I could try it on QEMU, and try to help debug it from there.

This could also be done directly in the C library, but there are some compilation problems which require recompiling the entire source tree from scratch every time a C library function is modified - so the above is much faster.

Thank you!

Mellvik:
Thank you @ghaerr.

It turns out in_connect is fine, I discovered a stupid mistake on my part and fixed it. The reason for the delay is something else.

Yet another round of ftp testing revealed a bug in the part that fetches multiple files from the remote host (mget, nlst). While testing a fix, I ran into a FAT file system problem which has taken my attention since.

It started with very slow directory listings (45 seconds to list a directory containing 250 short named files) and ended with a situation in which it takes 10 seconds to write and close a (very) small file. The latter is (temporarily) fixed by a reboot, the former is not, and I want to find the parameters affecting the slowness in order to provide a clue to the reason. And how to get the slow as molasses file write to repeat.
At first I figured I had a damaged file system, after all I've had a zillion hard hangs/power cycles recently. It wasn't. The FAT FS was OK, only the spare superblock (MSDOS scandisk fixed it) was out of date which is always the case anyway.

I'm opening an issue for the FAT problem for the record, which can be amended as more testing provide more info. The slow directory listing may be a 'feature' of the current design (I'm adding and deleting thousands of files to figure this out, it takes time), while I figure the deteriorating access time must be a bug.
I suspect FAT hasn't been beaten up this badly before.

I'm amending the ftp PR real soon (!), including a ftp manual page, ftpd page pending (it's much shorter so its a simple one).

—Mellvik

ghaerr:
> The FAT FS was OK, only the spare superblock (MSDOS scandisk fixed it) was out of date which is always the case anyway.

Out of curiosity, did scandisk complain about the 2nd FAT table being out of date? I have that on my list to fix, except that would only further slow down FAT write access. The FAT32 spare superblock is supposed to be kept in sector 6, I'll add that to my list of FAT enhancements.

> including a ftp manual page, ftpd page pending (it's much shorter so its a simple one).

Great. FYI, on macOS doing "man ./ftp.1" etc works well, I suggest using that for preview, then checking using ELKS' `man` command. I mention this as there are still a few formatting enhancements `man` should get, as it's output isn't exactly the same as macOS man yet, although close. In addition, you'll note that the `man` output for italicized text (used normally for command arguments) is implemented using underline on serial console, but not implemented on direct console (since underline is missing on EGA text).

Mellvik:
> 
> The FAT FS was OK, only the spare superblock (MSDOS scandisk fixed it) was out of date which is always the case anyway.
> 
> Out of curiosity, did scandisk complain about the 2nd FAT table being out of date? I have that on my list to fix, except that would only further slow down FAT write access. The FAT32 spare superblock is supposed to be kept in sector 6, I'll add that to my list of FAT enhancements.
> 
Yes, that's the complaint I get every time. Other than that - recently there have been  no FS errors on FAT, presumably thanks to the sync.
Apropos: If SYNC is commented out in bootopts, does that mean 'turned off' or is SYNC=0 the way to turn it off if desired?
BTW - maybe updating the 2nd FAT table should be part of sync instead of of every write ...
> including a ftp manual page, ftpd page pending (it's much shorter so its a simple one).
> 
> Great. FYI, on macOS doing "man ./ftp.1" etc works well, I suggest using that for preview, then checking using ELKS' man command.
> 
YEs, that's what I've been using. I don't have a setup in which I can easily use ELKS man yet, since I'm always building 1200 or 1440 floppies.
> I mention this as there are still a few formatting enhancements man should get, as it's output isn't exactly the same as macOS man yet, although close. In addition, you'll note that the man output for italicized text (used normally for command arguments) is implemented using underline on serial console, but not implemented on direct console (since underline is missing on EGA text).
> 
> 
Thanks - and honestly, I think this is good - and close enough to MacOS. The Compaq (and I suspect, many other old clunkers) can be configured as to what the preference for 'highlighted text' is. In Compaq's case  the options are underline, reverse video, half intensity, even a different charset. Being on serial all the time, I'm more than happy with what you've delivered.

Thank you.

-M

ghaerr:
> Apropos: If SYNC is commented out in bootopts, does that mean 'turned off' or is SYNC=0 the way to turn it off if desired?

Either will work. Prefer just leaving sync commented out.

> BTW - maybe updating the 2nd FAT table should be part of sync instead of of every write ...

Good idea; not sure yet whether we should/want to do this yet. The current code, commented out, actually goes through the trouble of recalculating the entire FAT chain a second time, obviously too slow. My idea would be to essentially copy the FAT disk block entirely (in RAM, and mark it for write with a sync), but that's also slow. Probably better idea first to just implement the 2nd superblock on FAT32, if anything at this time.

> The Compaq (and I suspect, many other old clunkers) can be configured as to what the preference for 'highlighted text' is.

At this point, `man` uses the ANSI bold sequence for bold, and ANSI underline for italic. I added underline (ESC [ 4m) to direct console, but it is coded as normal text, so you won't be able to change it on your Compaq. If Compaq allows for remapping something other than Bold, then let me know. There isn't (yet) an option for bold and italic to format the same, as I thought it didn't look that great on the console (and we need differentiation anyways to solve the problem we're discussing).

Mellvik:
> 
> Apropos: If SYNC is commented out in bootopts, does that mean 'turned off' or is SYNC=0 the way to turn it off if desired?
> 
> Either will work. Prefer just leaving sync commented out.
> 
> 
OK; thanks!


>  If Compaq allows for remapping something other than Bold, then let me know. There isn't (yet) an option for bold and italic to format the same, as I thought it didn't look that great on the console (and we need differentiation anyways to solve the problem we're discussing).
> 
No, it's like EGA - there is only one 'HIGHLIGHT'. The compaq (and some others) provide the capability to select how to display it. And the plasma screen is very different from a CRT in terms of visibility, so it makes sense.

-M

ghaerr:
> [[ftp] Bugfixes in ftp, added man-page.]

The new man page looks great, thanks! BTW, the convention we're following (and you are doing partially in the man page), is for the command names to be in bold, not italic. These then show well on the console. (We are also showing options in bold, which you are also doing). However, at the beginning of the man page, the various .I and .IR commands should be .B and .BR. This will then match with the \fB you're using for 'mget', etc. FYI.

Mellvik:
OK, admittedly I didn't even look, but used the BSD convention assuming that to be the mother of them all :-) (the page is a heavily edited version of the BSD 2.11 page).
I'll fix that when I add the ftpd page.

ghaerr:
Hello @Mellvik,

Are you ready to commit this? It looks OK, except the man pages are still using .I for the command name in certain places, whereas the convention is to use .B. This is the case not only in the full command line under DESCRIPTION, which looks correct, but also throughout the document; all other ELKS man pages currently follow this convention. 

Keep in mind that on direct console, which many will use to find help, italic displays nothing, although on serial console, it displays underline. This is why command names and dash arguments are bold, and any non-dash command arguments are italic - they impart the greatest visibility on the console for quick understanding on the console. Without it, there's just lots of non-highlighted and non-underlined text.

Thank you!

Mellvik:
Thank you @ghaerr,

please go ahead and commit this now.

I'll take another round on the man pages, as you can tell, I'm still on 'BSD-time' so to speak as far as man page standards go.

There will be an update to ftpd in the coming weeks anyway, now that the server timeout can be implemented.

-M

Issue #1203, opened 2022-03-05
mount issues


Mellvik:
Two issues popped up today while testing a newly partitioned HD (tested on VirtualBox only).

Using the same 500MB disk image that was shared the other day, now repartitioned into 4 partitions.

The first issue is minor, but possibly the trigger for the second. The new `mount` command lists mounted file systems and available space, something we've been missing forever. Now here's the thing (see the attached screenshot): `/dev/hda4` is always listed as `/dev/hda`, regardless of the sequence of mounts.

The second issue is more severe: When mounting filesystem #5, the system hangs hard. No warning, no message, just full stop. Maybe 4 mounts is max but it does need to be caught and reported somehow.

<img width="720" alt="Skjermbilde 2022-03-05 kl  15 06 36" src="https://user-images.githubusercontent.com/3629880/156887825-0a6bc35f-cce4-41ef-8c26-815727eb827a.png">

An additional issue, unrelated to `mount`, but instead related to `fsck`: Just like `mount` above, `fsck` doesn't like ´/dev/hda4`, but seem to work (or at least accept) the other file systems. See screenshot.

<img width="717" alt="Skjermbilde 2022-03-05 kl  15 37 48" src="https://user-images.githubusercontent.com/3629880/156888081-15bd27db-84e6-4888-a1d3-6e5fff4108ac.png">

ghaerr:
Hello @Mellvik,

Thanks for the testing!

> /dev/hda4 is always listed as /dev/hda, regardless of the sequence of mounts.

That's a bug, I'll fix it.

> Maybe 4 mounts is max but it does need to be caught and reported somehow.

The variable NR_SUPER in include/linuxmt/fs.h is currently set to 4, should we increase it? It does use kernel data segment space, but perhaps a change to 5 would help with small cost. I'll fix the bug in the kernel not rejecting the mount when over the max allowed. Good catch!

> Just like mount above, fsck doesn't like ´/dev/hda4`, but seem to work (or at least accept) the other file systems

This is a different problem than mount, but nonetheless should work. Could you package up this 4-partitioned hard drive as an image and post it, so I could look more into this? Not sure why this is failing at this point. (And/or give the commands used to partition and mkfs it, or both :)

Thank you!

Mellvik:
Thank you @ghaerr,
> Maybe 4 mounts is max but it does need to be caught and reported somehow.
> 
[...]

PR #1204, opened 2022-03-05
[kernel] Fix mount kernel and command bugs, increase mounts avail to 5

ghaerr:
This should fix all but the `fsck` issue in #1203.

Mount max is now 5, will error if over. Unfortunately, that error is -EINVAL, and is not easy to change, but shouldn't be a big deal (at least no hang).

### DIFF
--- a/elks/fs/super.c
+++ b/elks/fs/super.c
@@ -182,11 +182,13 @@ static struct super_block *read_super(kdev_t dev, int t, int flags,
     type = file_systems[0];
 #endif
 
-    for (s = super_blocks; s->s_dev; s++) {
-	if (s >= super_blocks + NR_SUPER) return NULL;
-    }
+    for (s = super_blocks; s->s_dev; s++)
+	continue;
+    if (s >= super_blocks + NR_SUPER)
+	return NULL;
+
     s->s_dev = dev;
-    s->s_flags = (unsigned short int) flags;
+    s->s_flags = flags;
 
     if (!type->read_super(s, data, silent)) {
 	s->s_dev = 0;
--- a/elks/include/linuxmt/fs.h
+++ b/elks/include/linuxmt/fs.h
@@ -49,7 +49,7 @@
 
 #define NR_INODE	96	/* this should be bigger than NR_FILE */
 #define NR_FILE 	64	/* this can well be larger on a larger system */
-#define NR_SUPER	4
+#define NR_SUPER	5
 
 #define BLOCK_SIZE	1024
 #define BLOCK_SIZE_BITS 10
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -172,6 +172,7 @@ static struct dev_name_struct {
 	{ "hda",     0x0300 },
 	{ "hdb",     0x0320 },
 	{ "hdc",     0x0340 },
+	{ "hdd",     0x0360 },
 	{ "fd0",     0x0380 },
 	{ "fd1",     0x03a0 },
 	{ "ttyS",    0x0440 },
@@ -194,9 +195,9 @@ static char * INITPROC root_dev_name(int dev)
 	for (i=0; i<5; i++) {
 		if (devices[i].num == (dev & 0xfff0)) {
 			strcpy(&name[NAMEOFF], devices[i].name);
-			if (i < 3) {
-				if (dev & 0x03) {
-					name[NAMEOFF+3] = '0' + (dev & 3);
+			if (i < 4) {
+				if (dev & 0x07) {
+					name[NAMEOFF+3] = '0' + (dev & 7);
 					name[NAMEOFF+4] = 0;
 				}
 			}
--- a/elkscmd/disk_utils/fsck.c
+++ b/elkscmd/disk_utils/fsck.c
@@ -542,6 +542,13 @@ void read_tables(void)
 /*	inode_buffer = malloc(BLOCK_SIZE); RUBOUT
 	if (!inode_buffer)
 		die("Unable to allocate buffer for inodes"); RUBOUT */
+
+	/* check for filesystem too large since libc malloc fails improperly FIXME */
+	if (ZONES > 32767) {
+		fprintf(stderr, "Filesystem too large to check (%u zones, max 32767)\n", ZONES);
+		exit(8);
+	}
+
 	inode_count = malloc(INODES);
 	if (!inode_count)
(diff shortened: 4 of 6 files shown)


ghaerr:
Hello @Mellvik,

I've fixed the `fsck` problem too, but not great news: fsck was just happening to fail on the slightly smaller filesystem that you had created as /dev/hda4. Long story short, libc `malloc` is buggy in that it allowed a malloc(65535) on the 65Mb filesystems /dev/hda2 and 3, but caused kernel sys_brk messages on hda4. However, the total data segment size available to fsck is only 64K, so in effect, fsck is completely unable to check filesystems greater than probably 32Mb without completely running out of memory, which doesn't error properly. In the case of /dev/hda2 and 3, fsck only appears to work, while actually scribbling all over internal memory - completely buggy and we didn't even know it!

So, the only quick fix is to limit fsck so that it's calls to malloc are meaningful and will work. I've added all this to the list, but the end result is we don't have an fsck that can work for larger filesystems. Ugh.

At this point, if the other fixes work for you, lets close #1203 and we can open an issue on fsck when we decide we need to up its priority.

Thank you!

Mellvik:
Sounds good (or bad depending on which part)  - @ghaerr.

IMHO - fsck is important so it's good the dirt comes out so it can be cleaned.

Being a true FAT-hater I'm now in the process of transferring the split hard disk image back to the compaq so I can leave FAT behind. Except for specific tests.
That will allow for beating up the minix fs on real HW, which TMK hasn't been done much ....

Also I think we need a discussion on the 'umount /' issue, but not in this thread.


—M

ghaerr:
> Also I think we need a discussion on the 'umount /' issue, but not in this thread.

Oops - the bugs have come in so quickly I forgot about that. I'll test and respond.

ghaerr:
@Mellvik,

My testing of `umount /` shows that no error is given, and no unmount occurs. Thus, I can't duplicate. Can you give more details, or remind me of the thread where you were discussing that problem?

Thank you!

Mellvik:
Yes, I'm getting the same now. 

Don't know what happened - it was repeatable after reboots.

Anyway, I believe it would be good to emit an error message, same as with any other attempt to umount a file system with open files.


—M

ghaerr:
> My testing of umount / shows that no error is given, and no unmount occurs.

> Yes, I'm getting the same now.
> Don't know what happened - it was repeatable after reboots.

It is possible this is a remnant of the kernel problem relating to not checking the max mount correctly. You might continue playing with this on and off...

> I believe it would be good to emit an error message, same as with any other attempt to umount a file system with open files.

Noted. Currently, umount isn't returning any error. I'll look into it.

ghaerr:
I just checked into this - "umount /" silently remounts root read-only, but does the filesystem for any open for write files and will fail if so.

It seems this (silent) operation was added prior to adding mount's "remount" option: "mount -o remount,ro / /", which will properly remount the root filesystem.

However, changing this does complicate the current method of checking a filesystem, including root, in /etc/mount.cfg:
```
check_mounted_filesystem()
{
    echo -n "$fsck $1: "
    umount $1
    $fsck $1
    mount -o remount,rw $1 $2
}
```
I'll have to think about how this shell function should work given any block device and mount point as arguments. (It is highly convenient to allow `umount` to work on any filesystem, including the root filesystem, in this case. Another idea is to allow "umount /" but display a message "Root filesystem mounted read-only".

Mellvik:
It's great to get this sorted out @ghaerr.

Now that ELKS is 'growing up', it's important to always make sure the root fs is clean. 

IIRC the 'normal' procedure is something like
1) check the superblock for clean umount, if yes, mount and go
2) otherwise mount root RO, do the checking, and reboot if changes were made, otherwise remount RW

It looks like we're heading in that direction!

Thank you.


-M

ghaerr:
> IIRC the 'normal' procedure is something like
1) check the superblock for clean umount, if yes, mount and go
2) otherwise mount root RO, do the checking, and reboot if changes were made, otherwise remount RW

Yes, pretty much. The /etc/mount.cfg just runs `fsck -r`, which should return success or error. An error will abort the shell script. Not sure if a reboot is necessary, but the script could certainly call it.

The current problems then are:
1) `fsck -r` - not sure if this works interactively or automatically - you were going to check into this. `fsck -a` is supposed to work automatically (non-interactively). We probably want -m (warn mode) and -f (force) also. There are also -l (list), -v (verbose) and -s (show) options which could be useful if/when errors are found, to keep an eye on fsck.

2) fsck doesn't work on filesystems larger than 32MB, due to running out of application memory. I'm working on a fix for this now, which would require a new system call to allocate far memory outside the fsck data segment.

If fsck doesn't actually work to repair a filesystem automatically (or for that matter, actually repair the filesystem at all), then that needs to be fixed before going to the trouble of adding special code to allow larger filesystems.

We should probably open a new issue on this, if you have time to jump a bit into fsck -r/-a.

Mellvik:
Thank you @ghaerr,

> 9. mar. 2022 kl. 18:32 skrev Gregory Haerr ***@***.***>:
> 
> 
> IIRC the 'normal' procedure is something like
> 
> check the superblock for clean umount, if yes, mount and go
> otherwise mount root RO, do the checking, and reboot if changes were made, otherwise remount RW
> Yes, pretty much. The /etc/mount.cfg just runs fsck -r, which should return success or error. An error will abort the shell script. Not sure if a reboot is necessary, but the script could certainly call it.
> 
> The current problems then are:
> 
> fsck -r - not sure if this works interactively or automatically - you were going to check into this. fsck -a is supposed to work automatically (non-interactively). We probably want -m (warn mode) and -f (force) also. There are also -l (list), -v (verbose) and -s (show) options which could be useful if/when errors are found, to keep an eye on fsck.
> 
> 
Yes, it on my list - got interrupted, back to the top now, ASA my alarm() testing is done.
> 
> fsck doesn't work on filesystems larger than 32MB, due to running out of application memory. I'm working on a fix for this now, which would require a new system call to allocate far memory outside the fsck data segment.
> 
> If fsck doesn't actually work to repair a filesystem automatically (or for that matter, actually repair the filesystem at all), then that needs to be fixed before going to the trouble of adding special code to allow larger filesystems.
> 
> We should probably open a new issue on this, if you have time to jump a bit into fsck -r/-a.
> 
> 
Agreed on both accounts. Let's continue in its own thread.

-M

PR #1206, opened 2022-03-05
[build] Fix CONFIG_ROMFS compilation errors

ghaerr:
Thanks for finding this, @Mellvik !


ghaerr:
Also @Mellvik, I couldn't quite duplicate the CONFIG_FAT_DEV config issue disabling XMS, but I found that particular config line was in error. So this probably fixes however you triggered that problem as well.

Thank you!

Issue #1196, opened 2022-03-01
[net] Attempting TCP access outside LAN directly after boot hangs system


ghaerr:
Opening new issue from discussion in [Add DNS nameserver #1160](https://github.com/jbruchon/elks/commit/f429953537cc603bfde43a05735b841e405ac3a6#commitcomment-67564843).

Issues are:
- Attempting TCP connect outside of LAN directly after boot hangs system on real hardware. However, works if a LAN host is connected to first, and on QEMU.
- On connect outside of LAN, router blocks traffic and replies with ICMP packet "Destination not reachable". Problem is that TCP level connect doesn't know about it, and waits 10 seconds before returning.

ghaerr:
Hello @Mellvik,

> Turns out that telnet to a random (presumably nonexistent) IP address also cause a hard hang.
> So - like you alluded to yesterday - this has nothing to do with DNS

I think the problem is likely not that of connecting to a random or nonexistent IP address, but rather, what's happening under the hood when ktcp tries to send a packet outside the local area network directly after boot, which is hanging the system. After a connection to a host on the LAN, then there seem to be no problems connecting outside the LAN (which is what nslookup does talking to a DNS server). Since this is all handled under-the-hood with ARP requests and replies, we need to see what's happening there, using 'arp' and a packet analyzer, if possible.

> telnet 1.1.1.1
hangs hard and is obviously an external address,
telnet 192.168.10.48
works fine - even though it is external to the ELKS LAN (10.0.2.x) … (it's my LAN; the other side of the Raspi ).

When you say "hangs hard", I assume that you've tried accessing the real hardware console and can't login, the system is crashed, correct? You can't ^C from the serial port, nor does login continue running on the console. This is distinct from a hard ktcp hang, which would still allow console logins and processes to run.

> but (I am guessing) in the code that replaces the destination address with the gateway address.

There is no code that replaces the destination address with the gateway address; the packet addresses are never rewritten. Instead, I'm thinking the problem may be in the link layer code - ktcp uses ARP to route the original packet to the gateway, I presume. Thus, if the ARP table or mechanism is initialized incorrectly, this could be the cause. We are going to need to see debug information on ARP and Ethernet packets.

[...]

PR #1207, opened 2022-03-05
[net] Return from connect on ICMP dest unreachable packet

ghaerr:
Possible fix for ending `connect` socket call in #1196 when ICMP destination unreachable message received.

@Mellvik, I have written this without any testing at all, but included various debug messages so that we can see what is happening on the ICMP message. Can you please test?

The intent is for ktcp to return from connect with a "No route to host" (EHOSTUNREACH) errno.

I can't move forward with the other problem reported in #1196, "system hang when connecting outside LAN directly after boot" without DEBUG_ARP and DEBUG_ETH (or other link level) traces.

Thank you!

### DIFF
--- a/elks/include/linuxmt/errno.h
+++ b/elks/include/linuxmt/errno.h
@@ -129,7 +129,7 @@
 #define	ETIMEDOUT	110	/* Connection timed out */
 #define	ECONNREFUSED	111	/* Connection refused */
 #define	EHOSTDOWN	112	/* Host is down */
-#define	EHOSTUNREACH	113	/* No route to host */
+#define	EHOSTUNREACH	113	/* Host not reachable */
 #define	EALREADY	114	/* Operation already in progress */
 #define	EINPROGRESS	115	/* Operation now in progress */
 #define	ESTALE		116	/* Stale NFS file handle */
--- a/elkscmd/ktcp/icmp.c
+++ b/elkscmd/ktcp/icmp.c
@@ -62,19 +62,18 @@ void icmp_process(struct iphdr_s *iph, unsigned char *packet)
    case ICMP_TYPE_DST_UNRCH:
 	dp = (struct icmp_dest_unreachable_s *)packet;
 	dpip = (struct iphdr_s *)dp->iphdr;
-	len = ntohs(dpip->tot_len) - 4 * IP_HLEN(dpip);
-	dptcp = (struct tcphdr_s *)dp->iphdr + len;
-	printf("icmp: len %d\n", len);
+	len = 4 * IP_HLEN(dpip);
+	dptcp = (struct tcphdr_s *)(dp->iphdr + len);
 	printf("icmp: destination unreachable code %d from %s\n",
 		dp->code, in_ntoa(iph->saddr));
-	printf("icmp: src %s:%u ", in_ntoa(dpip->saddr), ntohs(dptcp->sport));
-	printf("dst %s:%u\n", in_ntoa(dpip->daddr), ntohs(dptcp->dport));
+	debug_ip("icmp: src %s:%u ", in_ntoa(dpip->saddr), ntohs(dptcp->sport));
+	debug_ip("dst %s:%u\n", in_ntoa(dpip->daddr), ntohs(dptcp->dport));
 	cbnode = tcpcb_find(dpip->daddr, ntohs(dptcp->sport), ntohs(dptcp->dport));
 	if (cbnode) {
 	    struct tcpcb_s *cb = &cbnode->tcpcb;
 	    notify_sock(cb->sock, TDT_CONNECT, -EHOSTUNREACH);
 	    tcpcb_remove_cb(cb);	/* deallocate */
-	} else printf("icmp: Connection not found\n");
+	} else debug_ip("icmp: Connection not found\n");
 	break;
     default:
 	printf("icmp: unrecognized ICMP type %d\n", packet[0]);
--- a/elkscmd/ktcp/icmp.c
+++ b/elkscmd/ktcp/icmp.c
@@ -11,13 +11,16 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <fcntl.h>
 #include <stdio.h>
+#include <fcntl.h>
+#include <errno.h>
 #include <arpa/inet.h>
 #include "config.h"
 #include "ip.h"
 #include "tcp.h"
 #include "icmp.h"
+#include "tcp_cb.h"
+#include "tcpdev.h"
 #include "netconf.h"
 
 int icmp_init(void)
@@ -25,14 +28,21 @@ int icmp_init(void)
     return 0;
 }
 
-void icmp_process(struct iphdr_s *iph,unsigned char *packet)
+void icmp_process(struct iphdr_s *iph, unsigned char *packet)
 {
-    struct icmp_echo_s *ep = (struct icmp_echo_s *)packet;
+    struct icmp_echo_s *ep;
     struct addr_pair apair;
+
+    struct icmp_dest_unreachable_s *dp;
+    struct iphdr_s *dpip;
+    struct tcphdr_s *dptcp;
+    struct tcpcb_list_s *cbnode;
+
     int len;
 
     switch (packet[0]){
     case ICMP_TYPE_ECHO_REQ:
+	ep = (struct icmp_echo_s *)packet;
 	len = ntohs(iph->tot_len) - 4 * IP_HLEN(iph);
 	debug_ip("icmp: PING from %s (len %d id %u seqnum %u)\n",
 	    in_ntoa(iph->saddr), len, ntohs(ep->id), ntohs(ep->seqnum));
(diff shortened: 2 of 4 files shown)


Mellvik:
Thank you @ghaerr - 

I'll take a stab at this one tomorrow.

-M

Mellvik:
This is looking really good @ghaerr,

A couple of minor adjustments and your code works exactly as intended.

Thank you!

    case ICMP_TYPE_DST_UNRCH:
        dp = (struct icmp_dest_unreachable_s *)packet;
        dpip = (struct iphdr_s *)dp->iphdr;
-       len = ntohs(dpip->tot_len) - 4 * IP_HLEN(dpip);
-       dptcp = (struct tcphdr_s *)dp->iphdr + len;
+       len = 4 * IP_HLEN(dpip);
+       dptcp = (struct tcphdr_s *)(dp->iphdr + len);
        printf("icmp: len %d\n", len);
        printf("icmp: destination unreachable code %d from %s\n",

Output:

# telnet ams.mymayday.com
Connecting to ams.mymayday.com (178.62.252.72:23)
icmp: len 20
icmp: destination unreachable code 10 from 178.62.252.72
icmp: src 10.0.2.15:1025 dst 178.62.252.72:23
Connection failed: No route to host 
# 


The 'no route to host' may not be the most informative error message since we definitely have a route to the host. I think 'connection refused' is the most accurate - maybe not for all code 10 errors, but for the most common ones.

> I can't move forward with the other problem reported in #1196 <https://github.com/jbruchon/elks/issues/1196>, "system hang when connecting outside LAN directly after boot" without DEBUG_ARP and DEBUG_ETH (or other link level) traces.

I'll get back to you on this one.


—M

Mellvik:
> I can't move forward with the other problem reported in #1196 <#1196>, "system hang when connecting outside LAN directly after boot" without DEBUG_ARP and DEBUG_ETH (or other link level) traces.

@ghaerr,
I finally narrowed down this one. More specifically - I found the reason, not the solution.

The most unlikely of possible culprits: sync.

Commenting out SYNC in `/bootopts` - or setting SYNC=0  eliminates the problem.
Needless to say - it surely beats me. Unless somehow `net_port` gets overwritten in `init` which would surely result in a hard hang. Then again, why would that happen only if some network actions happen in a certain sequence.

It seems more likely that changing `sync_interval` in `init` somehow triggers a bug in the ne2k driver, which I haven't found yet. I'm still looking. The placement of the variables in `init` makes me suspicious.

Still looking for that one.

--M

ghaerr:
Hello @Mellvik,

Thank you for testing (and fixing) my code on this!

> The 'no route to host' may not be the most informative error message since we definitely have a route to the host. 

The error returned is EHOSTUNREACH, and the error string associated with it in /etc/perror is "No route to host". We don't want to use "Connection refused", since that's associated with ECONNREFUSED, which is returned when a RST is received. We will likely want to see the connect error difference between these two cases.

Perhaps the error string should be changed to "Host not reachable" for this error?

Thank you!

Mellvik:
Thank you @ghaerr.

> 
> The 'no route to host' may not be the most informative error message since we definitely have a route to the host.
> 
> The error returned is EHOSTUNREACH, and the error string associated with it in /etc/perror is "No route to host". We don't want to use "Connection refused", since that's associated with ECONNREFUSED, which is returned when a RST is received. We will likely want to see the connect error difference between these two cases.
> 
I see the point - and I don't agree. I think 'informative to the user' should take presedence over technicalities. Also, when looking at other systems, the jury is still out on the issue. While Linux choose the unreachable variant, MacOS uses conn refused.
> Perhaps the error string should be changed to "Host not reachable" for this error?
> 
> 
Yes, makes sense. As opposed to 'network unreachable', it may be argued that 'host unreachable' is true for this particular service.


-M

Mellvik:
A side issue that popped up during debugging:

Since the `arp` command always returns the entire arp table, not only the valid entries, it makes sense to zero the MAC address field of pending entries, otherwise the table may look like this:
```
# arp
10.0.2.21       b8.27.eb.9a.77.bc
10.0.2.20       b8.27.eb.9a.77.bc
10.0.2.2        b8.27.eb.9a.77.bc
0.0.0.0         00.00.00.00.00.00
0.0.0.0         00.00.00.00.00.00
# 

```

Here's my fix:
```
@@ -94,7 +94,10 @@ struct arp_cache *arp_cache_add(ipaddr_t ip_addr, eth_addr_t eth_addr)
        if (eth_addr) {
                memcpy (entry->eth_addr, eth_addr, sizeof (eth_addr_t));
                entry->valid = 1;
-       } else entry->valid = 0;        /* no MAC address yet, awaiting ARP reply*/
+       } else {
+               entry->valid = 0;       /* no MAC address yet, awaiting ARP reply*/
+               bzero(entry->eth_addr, sizeof(eth_addr_t));
+       }
        entry->qpacket = NULL;
``

**Comment by ghaerr** [comment]
> I think 'informative to the user' should take presedence over technicalities. 
> As opposed to 'network unreachable', it may be argued that 'host unreachable' is true for this particular service.

This [ICMP destination unreachable error code table](http://www.networksorcery.com/enp/protocol/icmp/msg3.htm) shows 16 different error codes in the ICMP Destination unreachable message. The particular system in this issue is returning a code 10, which is actually administratively prohibited. This is likely why macOS is returning Connection refused.

Since administratively prohibited is a far cry from destination or host unreachable, it makes sense to possibly separate the errno returned based on the ICMP error. There aren't ELKS errno codes for all 16 return values, and we probably don't need to write tons of code to handle this, but I'll look into it a bit more. 

I'm thinking three categories "Host unreachable", "Destination unreachable" and "Connection Refused" make the most sense, for errors 1, (0, 2-8), and (9-15) respectively.

**Comment by Mellvik** [comment]
Agreed @ghaerr,

and I see you have already done something about it.

Great - thanks - appreciated.


Slightly related question: 
I've added alarm/timeout to the ftpd, but it does not kick in.

It may be because I'm on a different (GIT) branch right now, but let me ask anyway: Is there a chance I'm loading the wrong library?


—M

**Comment by ghaerr** [comment]
> I've added alarm/timeout to the ftpd, but it does not kick in.

I assume you're talking about in_connect? Using `alarm` across other network calls isn't yet supported.

> Is there a chance I'm loading the wrong library?

in_connect is only in one file in the C library, so it should work. There is some trickiness to do with the C library when making other changes though - the entire tree must be recompiled from scratch, or the any C library routine modifications aren't included.

There's been quite a few modifications made lately, it is hard to say what might be your issue at this point, given when your branch was created.

**Comment by Mellvik** [comment]
No,
ftpd needs a timeout to terminate connections from idle clients. So now that we have an alarm() working,
I added that (for the standard 15 minutes) to ftpd.

It may be OK though - I just switched git branch and started again.

Will let you know in a few.

Thank you.

-M

**Comment by ghaerr** [comment]
> ftpd needs a timeout to terminate connections from idle clients. So now that we have an alarm() working,
I added that (for the standard 15 minutes) to ftpd.

One must be very careful with exactly what ftpd might be doing, if an alarm is coded outside a library routine. If ftpd is in a read(), we're probably OK. If ftpd happens to be in accept(), the network stack and/or kernel will likely experience data corruption, producing (more) unknown bugs, even if it seems to work.

Thus - I do not advise using alarm directly in network programs unless it can be proven that the application is not arbitrarily running, but instead waiting in either connect or read, nothing else. Alarm should be set just prior to the (approved) system call, and alarm(0) immediately afterwards. (Each system call must be specially coded to handle a signal interrupt and/or keep track of a dynamic state change. I plan on handling accept at some point, but its a bigger project to ensure no data corruption).

**Comment by Mellvik** [comment]
Yes,
and this one should be safe - just before and after the command read (btw, exactly like the BSD ftpd).

I'm still not getting alarms though. kill -14 to the process works. 

I'll do a cleanup, pull in all the PR merges and make a new system.

-M

**Comment by ghaerr** [comment]
> and this one should be safe - just before and after the command read (btw, exactly like the BSD ftpd).

Ok, that should work. read should return -1, errno=EINTR (not ETIMEDOUT like in connect) on an alarm file.

**Comment by Mellvik** [comment]
> 
> ﻿
> and this one should be safe - just before and after the command read (btw, exactly like the BSD ftpd).
> 
> Ok, that should work. read should return -1, errno=EINTR (not ETIMEDOUT like in connect) on an alarm file.
> 
Are you saying that the function registered in a preceding call to signal() will not be called on SIGALRM?
That's what it looks like - until I dive back into it tomorrow to verify.

-M

**Comment by ghaerr** [comment]
> Are you saying that the function registered in a preceding call to signal() will not be called on SIGALRM?

No. You must register your own empty callback function for alarm() to work, as the timer expiration sends a SIGALRM signal to your process. That callback function looks like this:
```
static void alarm_cb(int sig)
{
    /* no action */
}
```
The process then registers the callback in main() once using:
```
signal(SIGALRM, alarm_cb);
```

The `alarm_cb` signal handler remains installed, regardless of how many times a SIGALRM may be sent.

What I was saying is that the *interrupted* system call (in your case `read`) will return -1, errno=EINTR if the alarm expires.

Mellvik:
> 
Thank you, @ghaerr,
The reason I ask is that the callback never happens. So it seems the socket read is blocking (or rather, ignoring) the signal somehow.
Tests without the read work fine.

-M

ghaerr:
> The reason I ask is that the callback never happens. 

Post a diff so I can see what you're doing so I can help. The signals and alarm definitely work, the previous problem was related to having multiple alarms outstanding.

> So it seems the socket read is blocking (or rather, ignoring) the signal somehow.

No, the socket read will get interrupted as I have discussed when the signal arrives.

ghaerr:
> Tests without the read work fine.

Oops - missed this. I assume that the read is hanging, waiting for data on the command channel? That is, there is no current I/O, and ftpd is just sitting on a read? 

If ftpd executed a fork, you may have to re-execute the signal(SIGLARM, callback); again. Perhaps put that just before the alarmI(15*60) call.

Mellvik:
Thanks @ghaerr,
that put me on the right path.

It's now working fine.

PR with some updates coming.


—M

PR #1210, opened 2022-03-06
[cmds] More reduction in commands size, cleanup libc paths

ghaerr:
More reduction in typical disk size per command (around 1K - 2K smaller each).


Mellvik:
wow, @ghaerr - 
that's quite significant - is that file size, mem footprint or both?

—M

ghaerr:
> is that file size, mem footprint or both?

Just disk file size. Many of the smaller commands were bringing in stdio by using printf, which increases the disk size around 1500 bytes, it seems. Without compression, this added up to quite a few Kb for our basic distribution, of which we've been slowly running out of space and removing less-used commands.

Mellvik:
Thank you @ghaerr - 
Great work -  the total savings must approach 200k, very significant indeed.

--M

PR #1211, opened 2022-03-07
[cmds] Fix external tar command

ghaerr:
Here you go, @Mellvik.

The `tar` command now works. The problem was stack overflow from a large auto buffer.

Note: this external tar command is essentially the same source as the sash internal tar; thus it only provides list and extract functions, no create.


Mellvik:
@thank you @ghaerr - 

Tar (the other 'side') just moved up on my list :-)

-m

PR #1215, opened 2022-03-08
[kernel] Rewrite kernel timer routines from scratch

ghaerr:
Should fix system hang issue in #1196, which is the result of two things: the `alarm` system call was only implemented for one alarm system-wide which reused a single kernel timer, and secondly, it was found that the system hang occurred because the kernel's system timer routines were were buggy, and had to be rewritten. The kernel was busy-looping when a certain pattern occurred in the doubly-linked kernel timer list.

New limit of five alarms system-wide (and a limit of one of those per-process).
The kernel timer routines were rewritten and now use a simpler single-pointer linked list data structure.

The new "sync=" automatic sync interval uses an alarm in /bin/init, and each telnet or ftp connect will use an alarm. Tested with sync=3 and now telnet returns with "Connection timed out" as it should in 10 seconds.

Tested on QEMU, but could use more system testing, as the keyboard LED lights use a kernel timer (yes, they do!), as well as the polling keyboards (not scan code keyboards). A good test would be to set sync=3, telnet to a non-existent address, and hit the CapsLock key repeatedly. (No joke).

### DIFF
--- a/elks/arch/i86/drivers/char/kbd-poll-pc98.c
+++ b/elks/arch/i86/drivers/char/kbd-poll-pc98.c
@@ -67,7 +67,6 @@ static void restart_timer(void)
 {
     static struct timer_list timer;
 
-    init_timer(&timer);
     timer.tl_expires = jiffies + (8 * HZ/100);	/* every 8/100 second*/
     timer.tl_function = kbd_timer;
     add_timer(&timer);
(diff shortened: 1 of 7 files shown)


Mellvik:
@ghaerr,
there seem to be an overflow problem with timing in the `alarm()` routine.

When setting `alarm(900)`, the actual timeout is consistently 244.640 seconds.
`alarm(450)`  works as expected, as does `alarm(600)`.

-M

PR #1221, opened 2022-03-10
[Net] ftp/ftpd fixes and man page updates

Mellvik:
- `ftpd` now has a working idle timeout, defaults to 900 seconds
- `ftpd` and `ftp` support the SITE IDLE command to set or view the timeout value
- various cleanups and optimizations
- Updated man pages

The SITE command support is most useful for debugging and is surrounded by an `ifdef BLOAT`. 

Tested on physical only.

### DIFF
--- a/elkscmd/inet/ftp/ftp.c
+++ b/elkscmd/inet/ftp/ftp.c
@@ -96,6 +96,7 @@ enum {	// commands in disconnected mode
 	CMD_CLOSE,
 	CMD_PORT,
 	CMD_PASV,
+	CMD_SITE,
 	CMD_CONNECT,	// flag
 	// commands in any mode
 	CMD_QUIT,
@@ -150,6 +151,7 @@ struct cmd_tab cmdtab[] = {
 	{"lcd", CMD_LCWD, "Change local working directory"},
 	{"glob", CMD_GLOB, "Toggle globbing"},
 	{"passive", CMD_PASV, "Toggle passive/active mode"},
+	{"site", CMD_SITE, "Send command to server - supports 'idle'"},
 #endif
 	//{"user", CMD_USER, "Open new connection, log in."},
 	{"", CMD_NULL, ""}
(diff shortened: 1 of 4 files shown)


ghaerr:
Looks great @Mellvik! And thanks for all the documentation also :)

Are you ready for commit now?

Mellvik:
Thank you @ghaerr,

yes, it's ready for commit. It could take some more QEMU testing, but then again there haven't been any changes to the qemu specific stuff.

I'm moving on with the elks-to-elks network testing, which stranded on networking issues last time. FTP turns out to be a great asset for that. Also it lets us beat up the file systems with thousands of create/delete/update.

Which reminds me: I ran into an issue with mv on FAT: It simply doesn't work. Is that a known issue?

# mv full.tar full.xxx
full.xxx: No such file or directory 
# mv full.tar tmp
tmp/full.tar: No such file or directory 
# 

—M

ghaerr:
> I ran into an issue with mv on FAT: It simply doesn't work. Is that a known issue?

I'll have to look into that, there are no hard links on FAT, and the ELKS rename system call may not be implemented on FAT.

Mellvik:
Yes, it rings a bell. we've probably been here before.

I guess the easy way out is to have mv revert to cp when the FS is FAT.

One of these days it may be worthwhile to see how freeDOS implements he MOVE command.

-M

ghaerr:
> it rings a bell. we've probably been here before.

Issue #583 was opened by you almost two years ago, which ended up pointing out that our kernel doesn't support renaming directories on MINIX filesystems. We ended up writing some special code in `mv` to work around that.

Looking further into this issue, I realize (again) that FAT doesn't support any rename (of directories or regular files), so this is the problem here.

> I guess the easy way out is to have mv revert to cp when the FS is FAT.

`mv` already has copy code in it, I've written a fix that gets it to do just that on FAT filesystems. PR coming.

> One of these days it may be worthwhile to see how freeDOS implements he MOVE command.

It's not that big of a deal, the FAT directory entry itself just gets the name changed within it. Complications arise with long filenames. However, as is discussed in #583, the ELKS kernel doesn't support a proper rename: it calls sys_link to link to the new name, then unlinks the old name. FAT doesn't support links, so the operation fails.

PR #1223, opened 2022-03-12
Add BOOTCS and removed BOOTSEG for pc-98 hard disk boot

tyama501:
BOOTCS is added and BOOTSEG is removed since the load segment of PC-98 hard disk boot depends on IPL.
Also some modifications are added to enable booting pc-98 HD for particular configuration like SCSI ID=0.
The fat.bin built with CONFIG_IMG_HD is needed to be written in pc-98 partition.

We need to discuss more in #1047 to enhance usability.

Thank you.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -18,11 +18,10 @@
 #ifdef CONFIG_ARCH_PC98
 // Japanese NEC PC98
 
+#define BOOTCS
 #if defined(CONFIG_IMG_FD1232)
-#define BOOTSEG		0x1FC0
 #define SECTOR_SIZE	1024
 #else
-#define BOOTSEG		0x1FE0
 #define SECTOR_SIZE	512
 #endif
 #define MEMTOPSEG	0xA000		// 640K memory
@@ -107,8 +106,8 @@ entry1:
 	xor %sp,%sp
 
 	xor %di,%di
-#ifdef BOOTSEG
-	mov $BOOTSEG,%ax	// DS:SI = BOOTSEG:0
+#ifdef BOOTCS
+	mov %cs,%ax
 	mov %ax,%ds
 	xor %si,%si
 #else
@@ -346,6 +345,9 @@ dr_loop:
 	div %bx
 #ifdef CONFIG_ARCH_PC98
 	mov %al,%cl      // cylinder number for PC_98
+#if defined(CONFIG_IMG_HD)
+	mov %ah,%ch      // higher bits of cylinder number for PC_98
+#endif
 #else
 	mov %al,%ch      // CH = low 8 bits of physical cylinder number
 	ror %ah
@@ -363,7 +365,9 @@ dr_loop:
 #endif
 	div %bx          // DL = physical sector number - 1
 #ifdef CONFIG_ARCH_PC98
+#if defined(CONFIG_IMG_FD1232) || defined(CONFIG_IMG_FD1440)
 	inc %dl          // sector number for PC_98
+#endif
 #else
 	or %dl,%cl       // stash {physical sector number - 1} in CL
 	inc %cx          // base 1 for sector number
@@ -423,6 +427,9 @@ dr_try:
 	mov %es:(%bx),%al // Physical Device Address
 	push %ax
 	and $0x0F,%al
+#if defined(CONFIG_IMG_HD)
+	or $0x80,%al
+#endif
 	mov %al,drive_num
 	pop %ax
 	pop %es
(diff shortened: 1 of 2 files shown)


ghaerr:
Hello @tyama501, a couple questions:

In:
```
#ifdef CONFIG_ARCH_PC98
 	mov %al,%cl      // cylinder number for PC_98
 #if defined(CONFIG_IMG_HD)
 	mov %ah,%ch      // higher bits of cylinder number for PC_98
 #endif
```
are two separate instructions really needed, along with an #ifdef? It seems to me that the entire contents of %ax could just be moved to %cx with `mov %ax,%cx` for all systems?


In lines 368-370:
```
#if defined(CONFIG_IMG_FD1232) || defined(CONFIG_IMG_FD1440)
 	inc %dl          // sector number for PC_98
#endif
```
Why was this added? I thought your previous boot for FD1232 and FD1440 worked? Is the increment sector number needed for floppy, but not HD?

In lines 430-432:
```
#if defined(CONFIG_IMG_HD)
 	or $0x80,%al
#endif
```
Is this just to notify the kernel that the "drive number" is actually a SCSI ID, as previously discussed? Perhaps we should be passing the "physical device address" at 0:584 up to ELKS, which would not require special modification between floppy and hard drive. Then, bioshd_conv_bios_drive() could use that directly, and determine whether floppy or HD, as well as the SCSI ID.

What do you think of that idea?

ghaerr:
>  Perhaps we should be passing the "physical device address" at 0:584 up to ELKS

I have reviewed how this works in the boot sector and setup.S, and I think we can reduce the size of the PC-98 boot sector by not trying to pass the "boot drive" up (through drive_num in boot_sect.S and then into root_dev), to be then checked by bioshd_conv_bios_drive() in device_init(). This has to be done for IBM PC, because the BIOS does not save the boot drive in the BDA (BIOS data area).

For PC-98, all the following code could be removed from boot_sect.S:
```
        push %es
        xor %ax,%ax
        mov %ax,%es
        mov $0x584,%bx
        mov %es:(%bx),%al // Physical Device Address
        push %ax
        and $0x0F,%al
#if defined (CONFIG_IMG_HD)
        or $0x80,%al
#endif
        mov %al,drive_num
        pop %ax
        pop %es
```
whose sole purpose is to dummy up a "BIOS drive number" that ultimately gets passed as an argument to bioshd_conv_bios_drive(). Instead, for PC-98, the bioshd_conv_bios_drive function in bioshd.c can ignore its argument, and inspect 0:548 directly (using `peekw(584, 0)`) and inspect the physical device address directly. That can then be used, as described in https://github.com/jbruchon/elks/issues/1047#issuecomment-1065896418 to know the boot drive, query the SCSI drive table, and construct a rootdev return value, which will work without ifdefs in the boot sector for floppy and hard disk.

Does that make sense to you?

tyama501:
> It seems to me that the entire contents of %ax could just be moved to %cx with mov %ax,%cx for all systems?

Oh, probably your right. I will check that later.

> Why was this added? I thought your previous boot for FD1232 and FD1440 worked? Is the increment sector number needed for floppy, but not HD?

As I explained before, int1B bios, the first sector starts from number "1" for floppy, and it starts from number "0" for harddisk.

>Is this just to notify the kernel that the "drive number" is actually a SCSI ID, as previously discussed? Perhaps we should be passing the "physical device address" at 0:584 up to ELKS

If we passed  "physical device address"  to ELKS for now, I think it will not boot. Notice that "physical device address" is 9* for 1232 fd and 3* for 1440 floppy. I will modifiy after bioshd_conv_bios_drive() worked. 

>whose sole purpose is to dummy up a "BIOS drive number" that ultimately gets passed as an argument to bioshd_conv_bios_drive(). 

No, we need "physical device address" at 0:584 for disk read int1B input parameter.

ghaerr:
> As I explained before, int1B bios, the first sector starts from number "1" for floppy, and it starts from number "0" for harddisk.

Yes, I remember. I had thought that the floppy boot for PC-98 worked, and thus do not understand why this increment is being added now? If previous floppy worked, this increment would not be needed, or did something else change I am not aware of?

> If we passed "physical device address" to ELKS for now, I think it will not boot. 

Correct, other code definitely needs to be added.

> Notice that "physical device address" is 9* for 1232 fd and 3* for 1440 floppy. 

Ok, all that would be used in bioshd_conv_bios_drive. What is the physical device address for for HD SCSI?

> I will modifiy after bioshd_conv_bios_drive() worked.

Ok - I will commit this now, but am suggesting better way to do it that will also solve your original question about HD boot in #1047, now that I am fully understanding PC-98 better.

> No, we need "physical device address" at 0:584 for disk read int1B input parameter.

Ok, got it. In that case, after the next revision as discussed in https://github.com/jbruchon/elks/issues/1047#issuecomment-1065896418 and above, the following code could be deleted:
```
#if defined (CONFIG_IMG_HD)
        or $0x80,%al
#endif
        mov %al,drive_num
```
Thank you!

tyama501:
〉thus do not understand why this increment is being added now? 

I haven't added increments.
I only added ifdefs to exclude hard disk case.

〉the following code could be deleted:

OK, thank you.

ghaerr:
> I haven't added increments.
I only added ifdefs to exclude hard disk case.

Thank you, I finally see that now. Sorry it has taken me so long!

PR #1224, opened 2022-03-12
[tools] Add eman to display ELKS man pages from host system

ghaerr:
Adds `eman` (which is actually the ELKS man pager compiled on the host system) to allow display of ELKS man pages for developers, without having to run ELKS.

Usage is exactly like `man`, e.g. `eman accept`, or `eman ftp`.

This enhancement is quite useful for quickly checking new man pages while writing them, as well as a reference for writing programs for ELKS, for developers.

@Mellvik, I was looking at the ftp and ftpd man pages and saw some errors with respect to bold text; I thought this tool might help make it easier for you to see the pages yourself, without having to create a QEMU or physical hardware version each time.

### DIFF
--- a/elks/tools/.gitignore
+++ b/elks/tools/.gitignore
@@ -6,4 +6,5 @@ bin/objdump86
 bin/elf2elks
 bin/elks-compress
 bin/exomizer
+bin/eman
 elftoolchain-0.7.1
--- a/elks/tools/Makefile
+++ b/elks/tools/Makefile
@@ -34,6 +34,7 @@ all:
 	$(MAKE) -C objdump86 all
 	$(MAKE) -C elf2elks all
 	$(MAKE) -C elks-compress all
+	$(MAKE) -C eman all
 
 
 #########################################################################
--- a/elks/tools/eman/Makefile
+++ b/elks/tools/eman/Makefile
@@ -1,0 +1,51 @@
+# eman - display ELKS manual pages on host
+#
+#########################################################################
+#
+# Note! Dependencies are done automagically by 'make dep', which also
+# removes any old dependencies. DON'T put your own dependencies here
+# unless it's something special (ie not a .c file).
+#
+#########################################################################
+# Relative path to base directory.
+
+BASEDIR 	= ../..
+
+#########################################################################
+# Define the variables required by the standard rules - see the standard
+# rules file (below) for details of these variables.
+
+USEBCC 		= N
+
+CLEANDEP	=
+
+CLEANME 	= ../bin/eman
+
+DEPEND  	=
+
+DISTFILES	=
+
+NOINDENT	=
+
+#########################################################################
+# Include standard commands.
+
+include $(BASEDIR)/Makefile-rules
+
+#########################################################################
+# Objects to be compiled.
+
+SRCS=$(TOPDIR)/elkscmd/sys_utils/man.c
+
+#OBJS=$(SRCS:.c=.o)
+
+#########################################################################
+# Commands.
+
+all:	../bin/eman
+
+../bin/eman: $(SRCS)
+	$(CC) -D_PATH_MANPAGES=\"$(TOPDIR)/elkscmd/man\" -o ../bin/eman $(CFLAGS) $(SRCS)
+
+#########################################################################
+### Dependencies:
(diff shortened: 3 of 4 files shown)


Mellvik:
Indeed, @ghaerr - 

that's very useful. I certainly didn't realize that MacOS man was so smart about guessing what I really meant.

And you're right, I haven't reached the point where I actually view man pages on ELKS itself yet, physical or QEMU.

eman will speed up the process a lot!

Thank you.

-M

Mellvik:
BTW - I'll submit fixes to the ftp/ftpd man pages.

BTW2: There is a problem in both man and man with \fP (it's ignored) when the next char is a single quote (') or a right quote (´), maybe other characters.

BTW3: It would be convenient if man/eman filtered <CR>s which may have snuck into the file. 

Thank you.

-M

ghaerr:
Hello @Mellvik,

> There is a problem in both man and man with \fP (it's ignored) when the next char is a single quote (') or a right quote (´), maybe other characters.

Noted, along with a number of other small problems. The lowest level of `man` reads input differently than nroff, and is unlikely to be able to be easily fixed. There are also issues relating to (auto) font switching between double-quoted strings of text. 

In addition, as previously noted, \\fP barely works for other reasons. My recommendation is to just use \\fR (usually) and proceed to edit the text so that it displays well, until we have time to investigate the small issues in man. That said, I've fixed a number of problems already, but the list is still quite long.

> It would be convenient if man/eman filtered \<CR>s which may have snuck into the file.

Noted. You mean like a DOS-formatted document, or just one somewhere? Does Linux 'man' ignore all input CRs, or just some? 

I feel we're pretty lucky to have found a man formatter that runs as well as this one from Robert de Bath and allows ELKS to have man pages without nroff. Unless truly broken items (or sets of man pages that require a new feature) are found, I suggest we note issues like you're doing above, but nonetheless edit the page for displayability.

Thank you!

Mellvik:
> 
> 
> 
> There is a problem in both man and man with \fP (it's ignored) when the next char is a single quote (') or a right quote (´), maybe other characters.
> 
> Noted, along with a number of other small problems. The lowest level of man reads input differently than nroff, and is unlikely to be able to be easily fixed. There are also issues relating to (auto) font switching between double-quoted strings of text.
> 
> In addition, as previously noted, \fP barely works for other reasons. My recommendation is to just use \fR (usually) and proceed to edit the text so that it displays well, until we have time to investigate the small issues in man. That said, I've fixed a number of problems already, but the list is still quite long.
> 
No surprises there, and it's actually in pretty good shape. 
The thing with \fR is that it's pretty useless without \fP, so it's a pretty big thing. Fortunately it seems to work in many cases, but not when immediately followed by a newline or a quote or (presumably) some other 'special' characters.

> It would be convenient if man/eman filtered <CR>s which may have snuck into the file.
> 
> Noted. You mean like a DOS-formatted document, or just one somewhere? Does Linux 'man' ignore all input CRs, or just some?
> 
I don't know, it's just a convenience. Some times VI accidentally enter ff=dos and writes CRLF, some times I forget to set FTP top binary and the CRs get added.
> I feel we're pretty lucky to have found a man formatter that runs as well as this one from Robert de Bath and allows ELKS to have man pages without nroff. Unless truly broken items (or sets of man pages that require a new feature) are found, I suggest we note issues like you're doing above, but nonetheless edit the page for displayability.

Agreed!

Thanks.

-M

ghaerr:
> The thing with \fR is that it's pretty useless without \fP, so it's a pretty big thing.

Oh - I may be misunderstanding \fP (return to previous font). Isn't that what it does? \\fR (roman/regular) is quite usable when paired with \\fB (bold), which work great. \\fP has some incompatibility because I think nroff keeps a stack of previous, whereas our formatter only keeps the last font, which it returns to.

Mellvik:
No It's actually me.

A typo and an empty memory spot. 
I said \fR, meant \fB and all but forgot that we can use \fR in most of these cases and be fine with that.

Mea culpa - sorry about that.
I'll make those changes and we'll be fine for now.

-M

ghaerr:
> There is a problem in both man and man with \fP (it's ignored) when the next char is a single quote (') or a right quote (´)

I have found that sometimes adding a \\ in front of the single quote solves these kinds of problems with our formatter; try using \\fB\\' to see if that fixes it easily.

PR #1244, opened 2022-04-04
[8018x] interrupt pin mapping to logical IRQs

cocus:
Adds appropriate configurations for the 8018x processors to support
the INTx pins, mapping their interrupts to logical IRQs. This is
quite different than x86 PCs since there's no standard for the
devices using an 8018x.

### DIFF
--- a/8018x.config
+++ b/8018x.config
@@ -9,11 +9,18 @@
 
 
 #
-# Hardware
+# System
 #
 
 # CONFIG_ARCH_IBMPC is not set
 CONFIG_ARCH_8018X=y
+# CONFIG_ARCH_PC98 is not set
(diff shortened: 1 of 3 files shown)


ghaerr:
Looks great @cocus, thanks!

I see that the confusing "CONFIG_CPU_8086=y" is still in ELKS' configuration system, even for the 8018x. I've made a note to fix that.

cocus:
> I see that the confusing "CONFIG_CPU_8086=y" is still in ELKS' configuration system, even for the 8018x. I've made a note to fix that.

Yeah, that's an interesting side effect of this code relying on the 8086 stuff. 

Thanks!

PR #1246, opened 2022-04-06
[libc] Add floating point support to C library

ghaerr:
Add %e,%f,%g floating point output support to printf, sprintf etc.
Add ecvt, fcvt, strtod, atof libc functions.

Temporarily add elkscmd/misc_util/float.c example program showing use of floats. Currently, __STDIO_PRINT_FLOATS is required in main program to enable linking in extended float format support in printf, in order to keep size down. This may change to instead use weak linker symbols. 

Libc float support is being added as a precursor to ongoing work porting a Sinclair/Arduino BASIC interpreter to ELKS, which performs all math in floating point.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -105,6 +105,7 @@ sh_utils/which					:shutil					:1200k	:1440k
 #sh_utils/whoami				:shutil					:1200k	:1440k
 sh_utils/xargs					:shutil					:1200k	:1440k
 sh_utils/yes					:shutil				:720k
+misc_utils/float				:miscutil						:1440k
 misc_utils/compress				:miscutil						:1440k
 misc_utils/miniterm				:miscutil			:720k
 misc_utils/fdtest				:miscutil				:1200k	:1440k
--- a/elkscmd/misc_utils/.gitignore
+++ b/elkscmd/misc_utils/.gitignore
@@ -3,6 +3,7 @@ compress
 compress.host
 ed
 fdtest
+float
 tar
 od
 hd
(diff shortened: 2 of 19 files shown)


Mellvik:
Thank you, @ghaerr - 
Another important enhancement - which brings up the question of floating point instructions/coprocessor support.

Any thoughts on how big a challenge this would be?

--M

ghaerr:
Hello @Mellvik,

> which brings up the question of floating point instructions/coprocessor support.
Any thoughts on how big a challenge this would be?

I don't actually know whether hardware floating point support was ever added to @tkchia's tree. Andrew Jenner said back in 2017 he was thinking about adding it. I have some other questions about float vs double overhead in ia16-elf-gcc and will ask @tkchia about it on his repo.

I haven't (yet) measured the performance (or lack thereof) of software floating point in our new BASIC interpreter, but have been generally impressed with the code generated when only doubles (and no floats) are used. I have also located a couple viable math libraries and plan to port one of them to ELKS libc shortly. 

Do you have a need for very fast floating point? This also brings up the issue of whether a special compile/link would be needed for FP applications to select hardware vs software FP, versus having both contained within gcc's library at runtime.

Thank you!

Mellvik:
Thank you @ghaerr,
> Do you have a need for very fast floating point? This also brings up the issue of whether a special compile/link would be needed for FP applications to select hardware vs software FP, versus having both contained within gcc's library at runtime.
> 

Need - not really. More like the usual urge to see things working and learning about the improvements possible. Having a coprocessor in some of the machines helps.

Anyway - this is curiosity. I'm deep into ktcp these days - with the ambition to eliminate the need for delays on ELKS-to-ELKS. Here's the thing: There is no need for delays between other hosts and ELKS - not mget, not mput, not passive, not active, not when ELKS is the client, not when ELKS is the server. Since they are faster, the need for delays in at least some of these situations should objectively be greater. So - back to the packet traces to track down the differences. (it may turn out to be as simple as packet size, we'll see).

BTW _ while we're into enhancements, I think -a should be the default on mount so it can be omitted on the command line. -t is available if we need to make a mount fs-specific. Automatic is the default on Linux IIRC.

-M;

PR #1248, opened 2022-04-07
[basic] add load/save/dir support on the filesystem and some IO on 8018x

cocus:


ghaerr:
Hi @cocus,

This looks great, thank you. I will commit this now.

I think we should reserve .bas extensions for BASIC source code, and use another extension for tokenized form, perhaps .tok or .bac (do you have a preference)? This will provide for the future, and avoid overwriting source files. In the old days, Sinclair BASIC itself was the editor, so it didn't make much sense to "edit" basic source outside of it. We may have the opposite situation today. I also think the SAVE/LOAD commands should default to adding the tokenized extension automatically, so that one doesn't have to remember extensions, etc. I'll probably take a pass at add these changes.

How would you like to handle "autoexec" of either tokenized or plaintext BASIC? In the Arduino version, the tokenized file contained contained a byte specifying auto-run. We also need to consider what happens when the interpreter is started with a command line argument (of either .tok or .bas or neither).

For now, perhaps leave auto-run out, and just load programs from the command line or using LOAD. If we switch to using .bas (plaintext) files, the solution is simple - just put a "run" command at the end of the file with no line number.

Thank you!

cocus:
> For now, perhaps leave auto-run out, and just load programs from the command line or using LOAD. If we switch to using .bas (plaintext) files, the solution is simple - just put a "run" command at the end of the file with no line number.

That'll be a good thing. Yes, I prefer .bac for the tokenized ones.

Also, there's something really annoying. After a command finishes from the interactive console, I don't get an "OK" thus I don't know where each command (or the execution of the entire program) finishes. I think my old Sinclair machine did this, and all other contemporary machines also did it.

ghaerr:
> I don't get an "OK" thus I don't know where each command (or the execution of the entire program) finishes.

Are saying we should add "OK" after every immediate-mode command, as well as display that when a program finishes execution? (For abnormal termination using a future STOP/BREAK etc we may show something different).

I'll add all these changes shortly, I have a new fix for using float rather than double in the interpreter, which will use less space in the memfile, and we're only running 7 digits of precision anyways for numeric display.

I didn't know you used to have an actual Sinclair machine! :)

ghaerr:
@cocus,

Just to let you know your PR had quite a few compilation warnings, which would result in a system crash should the error code get executed:
```
ia16-elf-gcc -fno-inline -melks-libc -mcmodel=small -mno-segment-relocation-stuff -mtune=i8086 -Wall -Os -Wno-implicit-int  -I/Users/greg/net/elks-gh/include -I/Users/greg/net/elks-gh/libc/include -I/Users/greg/net/elks-gh/elks/include -D__ELKS__ -DELKS_VERSION=\"0.5.0-prei86\" -Wno-maybe-uninitialized -c -o host.o host.c
host.c: In function ‘host_loadProgramFromFile’:
host.c:249:17: warning: passing argument 1 of ‘sprintf’ from incompatible pointer type [-Wincompatible-pointer-types]
         sprintf(stderr, "Error opening file '%s'\n", fileName);
                 ^~~~~~
In file included from host.c:5:0:
/Users/greg/net/elks-gh/libc/include/string.h:69:5: note: expected ‘char *’ but argument is of type ‘FILE * {aka struct __stdio_file *}’
 int sprintf (char * str, const char * format, ...);
     ^~~~~~~
host.c:256:17: warning: passing argument 1 of ‘sprintf’ from incompatible pointer type [-Wincompatible-pointer-types]
         sprintf(stderr, "File too large to fit in memory\n");
                 ^~~~~~
In file included from host.c:5:0:
/Users/greg/net/elks-gh/libc/include/string.h:69:5: note: expected ‘char *’ but argument is of type ‘FILE * {aka struct __stdio_file *}’
 int sprintf (char * str, const char * format, ...);
     ^~~~~~~
host.c:268:13: warning: passing argument 1 of ‘sprintf’ from incompatible pointer type [-Wincompatible-pointer-types]
     sprintf(stderr, "Error reading file '%s'\n", fileName);
             ^~~~~~
```
I didn't catch these in my quick source review. You should definitely not submit code that compiles with warnings, for these very reasons.

`sprintf` takes a `char *` buffer argument, not a `FILE *`. I think what you meant is to call `printf` instead (but not stdout, not stderr, since all other BASIC output is to stdout).

I'll fix these with my next PR, but wanted to bring this to your attention.

Thank you!

cocus:
Woops! That was a slip up! I'll pay more attention in the future. Completely missed this, quite ashamed!

ghaerr:
No problem @cocus, it was late at night, and I'm sure you were coding very quickly :) Thanks for your efforts!

PR #1250, opened 2022-04-08
[libc] Port math library to ELKS libc

ghaerr:
Adds common math functions and math.h to ELKS C library, in both `double` and `float` precisions.

These functions were ported from the [OpenLibm](https://openlibm.org) math library, which itself derives from the [FreeBSD msun](http://svnweb.freebsd.org/base/head/lib/msun/) and [OpenBSD libm](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libm/src/) implementations, which in turn derive from [FDLIBM 5.3](http://www.netlib.org/fdlibm/).

The following functions are now available, and will be automatically linked in when coded into user programs without the use of `-lm`:
```
double floor(double x);
float floorf(float x);

double sqrt(double x);
float sqrtf(float x);

double fabs(double x);
float fabsf(float x);

double exp(double x);
float expf(float x);

double exp2(double x);
float exp2f(float x);

double pow(double x, double y);
float powf(float x, float y);

double log(double x);
float logf(float x);

double log2(double x);
float log2f(float x);

double log10(double x);
float log10f(float x);

double sin(double x);
float sinf(float x);

double cos(double x);
float cosf(float x);

double tan(double x);
float tanf(float x);

double acos(double x);
float acosf(float x);

double asin(double x);
float asinf(float x);

double atan(double x);
float atanf(float x);
```

The float precision version of these functions will be used in the Sinclair BASIC interpreter to implement POW, EXP, LN, SIN, COS, TAN, ASN, ACS and ATN.

The program elkscmd/misc_util/float.c is being used for testing, and I hope to provide a regression test suite as well in the future.

### DIFF
--- a/elkscmd/misc_utils/Makefile
+++ b/elkscmd/misc_utils/Makefile
@@ -23,7 +23,7 @@ PRGS += float
 all: $(PRGS) $(PRGS_HOST)
 
 float: float.o
-	$(LD) $(LDFLAGS) -o float float.o $(ECVT) $(LDLIBS) -lc
+	$(LD) $(LDFLAGS) -o float float.o $(LDLIBS) -lc
 
 ed: ed.o
 	$(LD) $(LDFLAGS) -o ed ed.o $(LDLIBS)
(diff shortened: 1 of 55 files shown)


Mellvik:
This is great, @ghaerr - 

you've been really busy!

Thank you.

—M;

PR #1251, opened 2022-04-08
Add codes for bios_disk_rw, Simplified int1B, Modified bootcode for PC-98

tyama501:
Hello @ghaerr ,

This PR enables the following features.

- Can use 16bits cylinder
- Simplified int1B

- Common boot code for FD1440 and SCSI Hard drive (FD1232 is still different)
- Booting from SCSI ID !=0 
- Booting from second partition

I incorporated your suggestions but please check if it needs to modify more.

Thank you!


ghaerr:
Hello @tyama501,

This looks very well done and super clean, thank you! This greatly simplifies the PC-98 BIOS disk functions and adds proper cylinder handling, along with fixing ELKS to boot from a secondary partition on a secondary drive.

I am thinking your new code in the genhd.c for identifying the boot partition can be used for the IBM PC also; I will test that after commit.

Glad to see this all done, thank you very much for your contribution!

PR #1252, opened 2022-04-08
[cmds] Add math functions to BASIC, allow host compilation

ghaerr:
Adds COS, SIN, TAN, ACS, ASN, ATN, EXP, LN and POW functions to BASIC.
For now, use POW(x,y) instead of the original Sinclair BASIC x^y.
Adds PI constant.
Uses new ELKS math library.

A host version of BASIC (`hostbasic`) for easier testing and development can be compiled by "make hostbasic" in elkscmd/basic.

BASIC now supports single or double precision calculations (about 6 vs 14 digit precision), selectable at compile time.
Or, math functions can be removed, resulting in a smaller binary (~50k vs ~30k).

We still have some floating point formatting/display issues, especially with regards to values extremely close to an integer. Some trig functions display -0 or 0.00001 instead of 0. I'm still working on this. There are also compatibility issues when running on ELKS vs a host platform with regards to very accurate float point text-to-translation and display.

### DIFF
--- a/elkscmd/basic/.gitignore
+++ b/elkscmd/basic/.gitignore
@@ -1,1 +1,2 @@
 basic
+hostbasic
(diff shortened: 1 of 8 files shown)


cocus:
Excellent! Nice improvements, specially on the math library!

Issue #1253, opened 2022-04-08
Obtaining CPU usage stats


cocus:
Hi, this is something I was thinking about from long tine ago, but now that I can telnet into my board, I'm sure I'd like to see how much cpu is being currently used.
Not sure if there's already a way to do so, but I have to ask it.
Also, I was thinking about improvements on the romfs version, like letting sash automatically execute an initial script, but I will asl that in another issue, if needed.

Thanks!

ghaerr:
> I'd like to see how much cpu is being currently used.

There isn't currently any implementation of CPU (or user/kernel) times at present. It could be a lot of work to add this similarly to a Linux kernel, the question would be to what real benefit on a single user system.

However, a simpler approach where the kernel timer tick (100 HZ) just increments a longword in the task structure of the currently running process would be pretty straightforward. Then `ps` could convert those timer ticks to MM:SS or something like that. That would show what processes are consuming time by running `ps` multiple times, but it wouldn't actually calculate CPU usage.

> letting sash automatically execute an initial script

`sash` will run a shell script (of simple internal/external commands only) when passed a file as an argument. Are you thinking you'd like to burn this into ROM for system startup? If that's the case, just put the commands in /etc/sashrc and they'll run when it starts.

Mellvik:
This brings up a much simpler command that I've been missing: uptime.
It should be as simple as recording boot time in a long (and ignoring time changes via the date command)?

-M;

cocus:
Well, I know the idea of this is to not run that many things at the same time, but for instance, I have to run ktcp, telnetd and now basic. I think it would make sense to know which thing is using cpu time.
So, incrementing a 32bit unsigned variable on the process control block seems quite tempting. Then you could expose these somehow, and let the user (in this case it would be a top-like app) do the calculations and obtain the cpu usage %.
I don't think this would be enough to calculate the uptime of the system, except if we consider the time of this variable on the pid 1.

ghaerr:
> Then you could expose these somehow

This doesn't have to get complicated - `ps` already works by opening /dev/kmem and reading each task's task structure to display its results. Adding a longword to each task structure that gets incremented by the HW timer would work to accumulate which task the kernel was running each tick, and an option could be added to `ps` to show the count, or some arithmetic on the count. This would end up combining 'user' vs 'system' CPU times, but it would simple and small.

> except if we consider the time of this variable on the pid 1.

[...]

PR #1265, opened 2022-04-27
[kernel] Add process CPU percentage utilization and uptime calculations

ghaerr:
Adds CPU column to `ps`.
Adds `uptime` command (linked to `ps`).
Adds `ps -l` option to display CSEG and DSEG.
Adds `ps -u` option to display uptime.

CPU percentage utilization is calculated every two seconds using fixed point arithmetic and an exponential decay to average utilization into the task average, using 1/e as the decay function, which results in averaging in approximately 63% of the last-calculated CPU % into the running average. Thus, it is not instantaneous, and `ps` will not likely show the exact CPU % for processes less than 2 seconds old for these and other sampling issue reasons.

Uptime will display properly for about 3/4 of a year before wrapping, using 32-bit calculations.

For now, these features are enabled by setting CONFIG_CPU_USAGE=y, which defaults ON. Only 5 bytes extra kernel stack was used for all calculations, and the averaging arithmetic uses a 10 bit integer, 11 bit fraction in 32 bits for calculations. This may be useful for CPU load or other calculations in the future. It took awhile to completely understand fixed point and exponential decay and testing, so more comments were added to elkscmd/misc_util/float.c and include/linuxmt/fixedpt.h.

Here's a screenshot of `ps` and `uptime` after booting the system and running an infinite loop in BASIC:

Screen Shot 2022-04-27 at 10 37 14 AM

Most of the time, `ps` will show 0 CPU for processes, as it takes quite a bit to consume all of the excess CPU time, given that most programs will be scheduled out on any select, delay or read system call. For instance, the following BASIC code uses very little CPU:
```
100 REM delay 1 msec
200 PAUSE 1
300 GOTO 200
```
Because of this kind of thing, and the tradeoff between instantaneous and decay-based utilization calculations, I am not sure of the true utility in this CPU percentage utilization calculation. Comments and testing feedback welcome!


Mellvik:
Thank you @ghaerr and others working on this.
A very welcome addition indeed - uptime for curiosity, cpu usage for debugging & tuning.
During system development we frequently have hung processes. With cpu usage stats it is easy to see if the problem is application level or system level. Also useful for performance tuning.

Appreciated.

-M

ghaerr:
Adds enhancement(s) requested in #1253, and started by @cocus.

PR #1266, opened 2022-04-27
Add MODE and PLOT for PC-98 to basic

tyama501:
Adding MODE and PLOT using LIO to basic as discussed at #1238 .

Also a sample program test98.bas is included.

Thank you!


ghaerr:
Hello @tyama501,

> I also added test98.bas to using MODE/CLS/PLOT/READ/DATA.

Wow, that looks really great!

> I added host_gcls() to clear graphic plane with CLS command.

Perhaps use only host_cls(), no host_gcls() and move it out of hosts.c into hosts-pc98.c and hosts-stubs.c. Then, for PC-98, have host_cls either emit the ANSI escape codes if not in graphics mode, otherwise clear graphics screen. This keeps extra host_ functions to a minimum, with only one per interpreted BASIC command.

Thank you!

tyama501:
Hello @ghaerr ,

It is ok to move host_cls to host-pc98 but even in MODE 1 the text with PRINT is output to the text plane so we need to have a way to clear it.
(There is a GPUT2 in LIO that put JIS character in graphic plane but ...)

In original N88-BASIC it was like this.
CLS 1 : clear text plane
CLS 2 : clear graphic plane
CLS 3 : clear both plane

IBM PC always put character on graphic plane in graphic mode?

ghaerr:
> It is ok to move host_cls to host-pc98 but even in MODE 1 the text with PRINT is output to the text plane so we need to have a way to clear it.

Well, there is no problem if you want to combine graphics and text based clear in the single call from the interpreter to host_cls().
That is, always emit the ANSI escape sequence and clear the graphics plane(s) if gmode set.

> CLS 1 : clear text plane
CLS 2 : clear graphic plane
CLS 3 : clear both plane

I don't think we need something this complicated at this time; a single CLS will clear the screen regardless of mode, using host_cls().

> IBM PC always put character on graphic plane in graphic mode?

To my knowledge, the IBM PC BIOS does not support text output in graphics modes - the application has to perform that itself.

ghaerr:
@tyama501, thanks for all your work on this!

tyama501:
Hello @ghaerr ,

I moved host_cls() to host-pc98.c and host-stubs.c.
I also needed to set *outfile as extern.

I haven't changed host-8018x.c which also need.

Thank you.

ghaerr:
> I haven't changed host-8018x.c which also need.

That's ok, you can include that in the next PR with your planned DRAW or other enhancements. When @cocus finds time, he can likely easily make changes for the 8018X build, if required.

Thank you!

PR #1269, opened 2022-04-28
[cmds] Add updated tar command

ghaerr:
Replaces `tar` command with enhanced version from Research UNIX V7, which supports create, list and extract.

@Mellvik, let me know what you think of this, it seems to work well, except that when given a directory of files to add, it requires using shell wildcard expansion in order to add files in that directory. I haven't had time to look into that yet. This is the only decently small but fully functional tar command I've been able to find so far.

This one is about twice the size of the last tar, and the old version should probably be removed as an internal `sash` command if replaced.

The compilation still spits out lots of warnings; I will fix then once we know whether this port of tar is suitable for ELKS.

### DIFF
--- a/elkscmd/file_utils/ls.c
+++ b/elkscmd/file_utils/ls.c
@@ -205,7 +205,7 @@ static void lsfile(char *name, struct stat *statbuf, int flags)
     cp = buf;
     *cp = '\0';
 
-    if (flags * (LSF_INODE|LSF_LONG|LSF_CLASS) && !statbuf) {
+    if (flags & (LSF_INODE|LSF_LONG|LSF_CLASS) && !statbuf) {
 	if (LSTAT(name, &sbuf) < 0) {
 	    perror(name);
 	    return;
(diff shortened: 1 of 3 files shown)


ghaerr:
@Mellvik, I've fixed the add directory issue above, using some later code from BSD's tar.c. Symlinks are now also supported.

ghaerr:
Also fixes duplicate display issue in `ls -lR`, apparently from several commits ago.

Mellvik:
Thank you @ghaerr,
This is great. I was going to ask you about symlinks - which was the reason I chose BSD tar over V7 for my (never completed) port some years back.

Now that I'm (almost) out of the rabbit hole I fell into when testing the 8bit ethernet interface 2 months ago, this is first on my list :-)

--M

ghaerr:
> I was going to ask you about symlinks - which was the reason I chose BSD tar over V7 for my (never completed) port some years back.

I haven't tested symlinks, and there's also an 'h' option to follow symlinks. Hard links should also be supported.
I started with UNIX V7's tar.c, and put too much time into it before realizing it read directories by opening the current directory and reading it directly - something that doesn't work since V7. I then found the BSD 4.11 version, which I probably should have used from the start. Luckily, it was pretty obviously ported from UNIX V7 originally. I pasted in most of the upgraded code dealing with reading directories and symlinks, and got everything working. Most of the additions in BSD are related to larger buffer sizes and things that ELKS won't be able to take advantage of, so I think we're ok, especially since we're trying to keep executables small, and tar is now 17K bytes, up from about 6K. It should pretty much do everything we want. Let me know!

Thank you!

PR #1271, opened 2022-04-29
[net] Add setup command to quickly configure hostname and IP address

ghaerr:
Adds `setup` command to allow quickly configuring the last digits of the system hostname and IP address.

Usage: `setup 16` will edit the /bootopts and /etc/hosts files to arrange for the system to have IP address 10.0.2.16 and hostname elks16 after the next boot. After execution, or with no argument, it displays the current settings of the /bootopts and /etc/hosts files.

@Mellvik, I this should ease your configuration duties after any ELKS update by allowing each of your systems to be network configured by only executing setup, followed by reboot. The setup command can be re-executed at any time, but since it uses /bootopts, it is necessary to reboot the system, rather than using net start/stop.

Let me know what you think, and what other changes should be made. You may notice that when using numbers 15, 16 or 17, there is as duplicate line in /etc/hosts, but this was necessary to allow selecting any other node number (3 ... 254) without making /etc/hosts too large or setup too complex.
Screen Shot 2022-04-29 at 3 53 51 PM

### DIFF
--- a/elkscmd/rootfs_template/bin/setup
+++ b/elkscmd/rootfs_template/bin/setup
@@ -1,8 +1,9 @@
 #
-# ELKS image network name/address setup script
+# Quick configure /bootopts and /etc/hosts files
 #
-# Usage: setup node_number
-#   node_number will be used as last digits in IP address and elks hostname
+# Usage: setup node_number [bootopts_enable_string]
+#	sets last digits of LOCALIP and HOSTNAME to node_number
+#	uncomment matching /bootopts lines
 #
 
 # defaults
@@ -10,25 +11,28 @@ network=10.0.2
 bootopts=/bootopts
 hosts=/etc/hosts
 
-# check for argument, use as id
+# 1st arg is node number
 if test "$#" -lt 1; then
-  echo "Usage: setup node_number"
+  echo "Usage: setup node_number [bootops_enable_string]"
   echo
   cat $bootopts
   cat $hosts
   exit 0
 fi
-
-# node_number used as last digits in IP address and hostname
 id=$1
 
-# template code follows
+# 2nd arg matchstr
+match_str=SEDNOMATCH
+if test "$#" -ge 2; then match_str=$2; fi
+
+# template strings
 hostname=elks$id
 localip=$network.$id
 gateway=$network.2
 
 # rewrite /bootopts
-sed -e "s/.*LOCALIP=.*/LOCALIP=$localip/" -e "s/.*HOSTNAME=.*/HOSTNAME=$hostname/" < $bootopts > /tmp/bootopts
+sed -e "s/.*LOCALIP=.*/LOCALIP=$localip/" -e "s/.*HOSTNAME=.*/HOSTNAME=$hostname/" \
+	-e "s/#\(.*$match_str.*\)/\1/" < $bootopts > /tmp/bootopts
 mv /tmp/bootopts $bootopts
 
 # rewrite /etc/hosts
--- a/elkscmd/rootfs_template/bin/setup
+++ b/elkscmd/rootfs_template/bin/setup
@@ -1,0 +1,47 @@
+#
+# ELKS image network name/address setup script
+#
+# Usage: setup node_number
+#   node_number will be used as last digits in IP address and elks hostname
+#
+
+# defaults
+network=10.0.2
+bootopts=/bootopts
+hosts=/etc/hosts
+
+# check for argument, use as id
+if test "$#" -lt 1; then
+  echo "Usage: setup node_number"
+  echo
+  cat $bootopts
+  cat $hosts
+  exit 0
+fi
+
+# node_number used as last digits in IP address and hostname
+id=$1
+
+# template code follows
+hostname=elks$id
+localip=$network.$id
+gateway=$network.2
+
+# rewrite /bootopts
+sed -e "s/.*LOCALIP=.*/LOCALIP=$localip/" -e "s/.*HOSTNAME=.*/HOSTNAME=$hostname/" < $bootopts > /tmp/bootopts
+mv /tmp/bootopts $bootopts
+
+# rewrite /etc/hosts
+cat << EOF > $hosts
+#127.0.0.1	localhost
+$network.$id	elks$id
+$network.15	elks15
+$network.16	elks16
+$network.17	elks17
+$gateway	gateway
+EOF
+
+# display results
+cat $bootopts
+cat $hosts
+sync
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -3,6 +3,7 @@
 #QEMU=1			# to use ftp/ftpd in qemu
 #TZ=MDT7
 #LOCALIP=10.0.2.16
+#HOSTNAME=elks16
 #netirq=9 netport=0x300 # NIC
 #bufs=2500		# system buffers
 #sync=30		# autosync secs
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -1,5 +1,5 @@
 ## boot options max size 511 bytes
-#console=ttyS0,57600 debug net=eth 3 # serial console, multiuser, networking
+#console=ttyS0,57600 debug net=eth 3 # sercons, multiuser, networking
 #QEMU=1			# to use ftp/ftpd in qemu
 #TZ=MDT7
 #LOCALIP=10.0.2.16
--- a/elkscmd/rootfs_template/etc/hosts
+++ b/elkscmd/rootfs_template/etc/hosts
@@ -1,5 +1,5 @@
 #127.0.0.1	localhost
-10.0.2.15	elks1 elks
-10.0.2.16	elks2
-10.0.2.17	elks3
+10.0.2.15	elks15
+10.0.2.16	elks16
+10.0.2.17	elks17
 10.0.2.2	gateway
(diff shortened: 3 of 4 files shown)


Mellvik:
Thank you @ghaerr,
This is great - and actually more helpful than it may initially appear. Not having to load /bin/vi on every boot (new image) saves a lot of time (floppy load) - in particular because vi flushes stdin just before displaying the first screen, so there is no typeahead :-) The latter has been on my list for a long time, will eventually be fixed.

I'd like to add two things:
1) a 'c' or 'console' command line option to setup, which cause the removal of the hash on the first occurence of '#console' in /bootopts
2) add export PS1="${HOSTNAME}# " to /etc/profile

--M

ghaerr:
Hello @Mellvik,

I'm traveling and will add your two requests when I am able. After you start using it, you can then maintain the script with other needed enhancements and submit PRs for them, which is nice since its a shell script and should be easily updatable. This could be useful for other users if documented.

Are you using the serial line or direct console when vi drops input characters? The kernel doesn't actually implement the TCSETF (TCSAFLUSH) ioctl and should never drop input. I am wondering if this has to do with the "automatic window sizing" which allows vi to learn how big the serial terminal window is, by issuing a cursor position followed by DSR (device status report, read cursor position) instead. In that case, typing ahead could possibly interfere with that operation, which would be indicated by possible visually incorrect output in the first few characters displayed by vi. The DSR could be turned off with a command line option, but that would also result in vi thinking the serial console window is always 80x25.

Thank you!

Mellvik:
Thank you @ghaerr, and no problem.
Actually - if you commit as is, I can add these 2 items myself and take it from there.

>  This could be useful for other users if documented.
> 
> 
Absolutely. I'll put it on my doc list - where the ne2k page now resides on top, an important (!) update coming up.
> Are you using the serial line or direct console when vi drops input characters?
> 
I'm using serial most of the time - yes, but this annoyance is universal. Telnet, serial, console.
> The kernel doesn't actually implement the TCSETF (TCSAFLUSH) ioctl and should never drop input. I am wondering if this has to do with the "automatic window sizing" which allows vi to learn how big the serial terminal window is, by issuing a cursor position followed by DSR (device status report, read cursor position) instead. In that case, typing ahead could possibly interfere with that operation, which would be indicated by possible visually incorrect output in the first few characters displayed by vi. The DSR could be turned off with a command line option, but that would also result in vi thinking the serial console window is always 80x25.
> 
Interesting. This might just be it. I'll take a note of this and check it out. I Believe it should be possible to keep the real keystrokes in the queue regardless, since that's what other *ux systems do.

—M

ghaerr:
Hello @Mellvik,

I've added your two latest requests, and enhanced one of them. If HOSTNAME= is found in the environment (from /bootopts), /etc/profile will automatically add the string to the shell prompt. This will work when logging in as root or as an ordinary user (e.g. user1).

Also, if a second parameter is given to the `setup` command, it will search all of /bootopts for a match, and enable those lines which match the argument string. I changed the first console line comment to 'sercons', so that, for you, typing
```
setup 16 sercons
```
will set LOCALIP=10.0.2.16 and HOSTNAME=elks16, along with enabling the console= line that matches 'sercons', which is:
```
console=ttyS0,57600 debug net=eth 3 # sercons, multiuser, networking
```
This matching mechanism allows one to add various tags to /bootopts that can be turned on quickly using `setup`. Note, once enabled, they will not be disabled by an invocation of `setup`. However, `setup` can be run multiple times, with a different second argument to enable an additional /bootopts (matching string) option (e.g. `setup 16 sync` will enable the `sync=30` line).

Try playing around with this a bit, and let me know how it works for you. As you mentioned, we can continue to improve this as the need arises.

Thank you!

ghaerr:
> which allows vi to learn how big the serial terminal window is, by issuing a cursor position followed by DSR (device status report, read cursor position)

>  I Believe it should be possible to keep the real keystrokes in the queue regardless, since that's what other *ux systems do.

Not sure how this can be done - after vi sends the DSR sequence to the terminal, it reads from stdin to get the cursor position, which eats any queued keystrokes.

Mellvik:
This is great @ghaerr - 

yet to be tested, but the 'tag' mechanism is very smart.

Thank you!

—M

Mellvik:
Thank you @ghaerr,

> 
> I Believe it should be possible to keep the real keystrokes in the queue regardless, since that's what other *ux systems do.
> 
> Not sure how this can be done - after vi sends the DSR sequence to the terminal, it reads from stdin to get the cursor position, which eats any queued keystrokes.
> 
> 
without looking at the code (yet), I'm assuming the DSR query is happening just before the screen is 'painted', so the question really is what happens to the input buffer before the query is sent.

Anyway, we can leave this one for now, it's on the list anyway. It's one of those things that pops up every once in a while, so it will not be forgotten.


-M

ghaerr:
> without looking at the code (yet), I'm assuming the DSR query is happening just before the screen is 'painted'

I added the functions getWindowSize and getCursorPosition to elkscmd/elvis/curses.c, the latter which sends DSR and reads stdin. The getCursorPosition function writes DSR and reads stdin, twice - first to save the original cursor position, then again to find out the bottom right margin. These were added so that vi functions properly in a serial terminal environment.

The getWindowSize function is called from initscr(), which is called very early in mainI(). So nothing reads stdin prior. The getCursorPosition function sets up termio struct to read raw characters with 200ms timeout, and calls tcsetattr(...,TCSAFLUSH,...). The kernel ignores TCSAFLUSH and doesn't drop characters, but getCursorPosition tries reading up to 32 characters (w/timeout) looking for the DSR 'R' terminator, then parses the input. If the input doesn't start with ESC [, the function fails. It would seem that if you typed ahead that this would interfere with the function and that vi would not adjust properly to the terminal window size - does it act like 80x25 (top of screen), or is it full screen on serial?

We could keep type-ahead keystrokes by having a switch or env variable to turn off auto-window size; it is only needed on the serial line, as the console is always fixed at 80x25. Another fairly straightforward enhancement would be to turn off auto-window size when not on /dev/ttyS[0-2].

Mellvik:
hmmmm, this is getting more interesting.
> without looking at the code (yet), I'm assuming the DSR query is happening just before the screen is 'painted'
> 
> I added the functions getWindowSize and getCursorPosition to elkscmd/elvis/curses.c, the latter which sends DSR and reads stdin. The getCursorPosition function writes DSR and reads stdin, twice - first to save the original cursor position, then again to find out the bottom right margin. These were added so that vi functions properly in a serial terminal environment.
> 
> The getWindowSize function is called from initscr(), which is called very early in mainI(). So nothing reads stdin prior. The getCursorPosition function sets up termio struct to read raw characters with 200ms timeout, and calls tcsetattr(...,TCSAFLUSH,...). The kernel ignores TCSAFLUSH and doesn't drop characters, but getCursorPosition tries reading up to 32 characters (w/timeout) looking for the DSR 'R' terminator, then parses the input. If the input doesn't start with ESC [, the function fails. It would seem that if you typed ahead that this would interfere with the function and that vi would not adjust properly to the terminal window size - does it act like 80x25 (top of screen), or is it full screen on serial?
> 
OK; this may be the reason I rarely see the window sizing work properly, having the habit of always typing ahead while waiting for vi to load. It's probably smart to dig into this with some debug prints in order to figure out the details. This happens equally on serial and console BTW.
> We could keep type-ahead keystrokes by having a switch or env variable to turn off auto-window size; it is only needed on the serial line, as the console is always fixed at 80x25. Another fairly straightforward enhancement would be to turn off auto-window size when not on /dev/ttyS[0-2].
> 
what we really need is something like ungetc(), which of course isn't possible at this level. Let's have a look and see what vim is doing. I'll be back.



thank you.

—M

ghaerr:
> what we really need is something like ungetc(), which of course isn't possible at this level. 

`vi` could certainly read and save keyboard input up until the point the DSR is sent and response received, and then that saved input used as keyboard input and processed. However, that still doesn't solve the problem of someone typing during the DSR send/response query.

> Let's have a look and see what vim is doing. 

I think what you'll find is that all the screen editors use the TIOCGWINSZ ioctl or SIGWINCH signal to query or receive the terminal window's size. This method doesn't require any of the DSR send/response query and doesn't eat any keyboard input. The problem is, these methods only work on operating systems that are also running the terminal emulator window. ELKS of course doesn't do any windows and an external terminal emulator is being used to access programs via a serial line, which isn't the case for `vim` on almost all the other systems it runs on.

The other approach, which we used for a while, is for the shell to produce updated LINES=/COLS= environment variables so that a screen editor can use use those directly, rather than querying the terminal using DSR. The problem with that method is that the shell must perform the DSR query itself before every command, which we used to do using linenoise library. I removed that a while back after noticing that the cursor was getting screwed up when typing quickly at the shell prompt. The shell now assumes an 80 character wide line when doing any line editing, and is buggy if a user tries editing past 80 columns or on a terminal with less columns. When I removed the DSR query from linenoise and the shell, I added it to vi so that screen editing still worked on serial consoles.

Mellvik:
> what we really need is something like ungetc(), which of course isn't possible at this level.
> 
> vi could certainly read and save keyboard input up until the point the DSR is sent and response received, and then that saved input used as keyboard input and processed. However, that still doesn't solve the problem of someone typing during the DSR send/response query.
> 
That's probably the right way to approach this. And I think the time span of the DSR sequence would be so short the chance of it getting interrupted would be limited.
> Let's have a look and see what vim is doing.
> 
> I think what you'll find is that all the screen editors use the TIOCGWINSZ ioctl or SIGWINCH signal to query or receive the terminal window's size. This method doesn't require any of the DSR send/response query and doesn't eat any keyboard input. The problem is, these methods only work on operating systems that are also running the terminal emulator window.
> 
That may be the case, but it still works remotely. In my daily ELKS work, the iMac has 5+ terminal windows sshd into a mac mini, 5+ sshd into a raspi, all of them window-scaling perfectly fine - and dynamically as the window size changes. IIRC even telnet has commands to send updates about window changes dynamically these days.
> 
To be investigated, not top priority.

Thank you.

-M

PR #1272, opened 2022-04-29
[basic] add stubs for 8018x

cocus:


ghaerr:
Thanks Santiago!

cocus:
Can we add the 8018x and PC-98 builds as part of the CI/CD? Maybe even the version with the XMS enabled and another one with the XMS disabled? would be cool

ghaerr:
> Can we add the 8018x and PC-98 builds as part of the CI/CD? 

Yes, let me look into how to best do that. It would be nice to perform a rebuild without having to start from scratch for each architecture. The kernel can be remade using a "make kclean" which could save quite a bit of time, and then only ps and basic depend on .config settings.

> Maybe even the version with the XMS enabled and another one with the XMS disabled?

That particular option doesn't change much in compilation, but all default binaries should have that option turned off, along with any other options that won't run on the earliest machines.

tyama501:
Thank you for adding stubs @cocus !

@ghaerr , please also consider rebuild clock.c for PC-98. Thank you!

Issue #1273, opened 2022-05-01
RTC support


cocus:
Hi,
I've now assembled [my own 80C188EB SBC](https://oshwlab.com/cocus/80c188eb-sbc) that I've made following the reverse-engineering I did on the actual HDSL modem board I was running all my tests before.
I need to test the ISA interface, but the SD works fine!
On this SBC I've added some new "toys" that I didn't have on the modem. To being with, I've added a bq3285 RTC. I see there's no support for hardware RTC on ELKS, and that'll be really nice to have. This RTC won't behave as a PC's RTC since you have full access to the actual RTC's registers, rather than accessing them thru a buffered port.

Not sure if this is on the list of things to do, but it'll be cool.

There's also an ADC that I took out of the modem board, but I don't think it'll make sense to add it as an ELKS device.
Finally, the secondary UART can be used with my board since I've added the secondary UART IRQ to one of the INT pins, so some work related to the secondary UART will come shortly as well.
And I think I'm having an issue with the timers, since I can't get the "date" to increment, and one of my test apps that does a sleep(1) doesn't return of that. I'll have to dig deeper to understand what's the matter.

cocus:
> And I think I'm having an issue with the timers, since I can't get the "date" to increment, and one of my test apps that does a sleep(1) doesn't return of that. I'll have to dig deeper to understand what's the matter.

I've figured this out. The 80188EB user's manual stated this but it wasn't too clear about it:
```
The timer input pins affect timer counting in three ways (see Table 9-2). The programming of the External (EXT) and Retrigger (RTG) bits in the Timer Control register determines how the input signals are used. When the timers are clocked internally, the RTG bit determines whether the input pin enables timer counting or retriggers the current timing cycle.
```
In this case I wanted to have a continuous (i.e. keeping the timer enabled) timer 1 (timer 2 is not affected by this), so it means that the T1IN pin needs to be high in order for it to work. So this must have been why Timer0 didn't work back when we started the port, its T0IN pin must have been held low, thus not enabling it. I've pulled it high and I have the timer working back again!

Mellvik:
Thank you, Santiago -
To those of us so inclined your detailed reports are very interesting. It's almost like we can smell the smoke from the hot soldering iron and see the signals on the logic analyzer.

Looking forward to the next episodes. :-)

-M

ghaerr:
[...]

PR #1274, opened 2022-05-03
[kernel] Add compile-time TZ= timezone string, FAT use localtime

ghaerr:
Allows for .config to contain TZ= timezone string to be compiled into kernel, eliminating requirement for editing /etc/profile or /bootopts to set local timezone, pursuant to discussion in https://github.com/jbruchon/elks/issues/1273#issuecomment-1115081566.

NOTE: `make menuconfig` will show new CONFIG_TIMEZONE TZ= option, but configure 'string' command is buggy and attempting to edit via `make menuconfig` will hang the configure, requiring a ^C exit. One must edit .config directly and add CONFIG_TIMEZONE="JST-9" etc directly. I have added this to pc98.config.

TZ=MDT6, TZ=JST-9 etc string will be automatically exported by /etc/profile into shell environment for proper libc time conversions. This is done in a bit tricky way by modifying /etc/profile with compiled-in TZ=CONFIG_TIMEZONE for during the build process.

/bootopts will override compiled-in TZ= string if set there.

FAT filesystem timestamps now saved in localtime.

@tyama501, I have tested this, but would definitely appreciate your taking a look at it to see how it works in Japan. I added the appropriate CONFIG_TIMEZONE to pc98.config which will need to be copied to your .config. The `clock -s -u` was not modified in /etc/rc.sys, the -u may need to be removed when a compiled-in timezone is set. We can talk about how we should automatically handle that in rc.sys should that be required. FAT files should now be saved in localtime.


ghaerr:
Hello @jbruchon,

As I posted above, for some reason the 'string' command in the ELKS kernel configure mechanism seems to be completely broken. (That is, using `string 'Enter TZ string value' CONFIG_TIMEZONE ""` will crash menuconfig if entering a string is attempted, see elks/config.in in this PR). There are some other unused "string" commands elsewhere in ELKS but they're never executed or commented out of the config.in scripts, so this seems to be the first time we've tried to use it in years.

I was reading the make menuconfig documentation in elks/config and saw that you were the maintainer... that is back about 25 years ago!!! :) I was wondering if you had any idea how this could be fixed? I started looking at the Configure and Menuconfig scripts and it about blew my mind, and don't know where to start. Any help or tips would be appreciated.

Thank you!

jbruchon:
I don't actually know that much about the inner workings of menuconfig. It was adopted from a very old Linux kernel and I always thought it'd be a good idea to completely redo that by stealing the much better developed menuconfig framework from a much newer Linux kernel. I wasn't the maintainer 25 years ago (I wasn't even in high school yet, IIRC) but I did "take over" the project in 2006 because I didn't want to watch it die. I think it was harkal [on SourceForge](https://sourceforge.net/projects/elks/) that responded to my request to get admin access by basically giving me full control of the project, and I've just sort of kicked it along ever since. It was a shambled-together mess from the start and menuconfig is the biggest legacy mess that has stuck around.

I'd say it's time to rip out the existing config system and pull a better one. I'd actually suggest using the config system used by [BusyBox.](https://www.busybox.net/) The Linux kernel's build system has added a lot of annoying dependencies while BusyBox remains fairly pure, plus BusyBox puts the config data at the top of the C files they manipulate rather than in some giant master config database. It would also be a great opportunity to NOT carry over all the useless and unused stuff such as unused config options.

tyama501:
Hello @ghaerr ,

Thank you for this modification.

> I have tested this, but would definitely appreciate your taking a look at it to see how it works in Japan. 

I tested with PC-9801RX and now I can see the same time stamp for dos and elks with JST-9.

dos partition file from dos (05050115.txt, 22-05-05 1:16)
RTC_IMG_20220505_011638

dos partition file from elks (05050115.txt, May 5 01:16)
RTC_IMG_20220505_012003

elks disk file from elks (05050121.txt, May 5 01:22)
RTC_IMG_20220505_012320

elks disk file from dos (05050121.txt, 22-05-05, 1:22)
RTC_IMG_20220505_012647

I have noticed a issue for "date" command.
I will write about this in #1273.

ghaerr:
Hello @tyama501,

Thank you for your testing! Glad that FAT file dates are now correct.

At some point, I am curious as to whether moving to compiled-in CONFIG_TIMEZONE="JST-9" support will work with clock command still running -u parameter (UTC) in /etc/rc.sys. Did you run the above tests with CONFIG_TIMEZONE="JST-9" compiled in? That should allow you to switch to your (preferred) RTC localtime, but not have to set TZ= every time you build an image.

What was not included in this PR is a way to exclude -u option from clock if TZ is set. Suggesting something like in /etc/rc.sys:
```
utc=
if test "$TZ" = ""; then utc=-u; fi
clock -s $utc
```

Thank you!

tyama501:
Hello @ghaerr ,

Yes, I tested with CONFIG_TIMEZONE="JST-9" compiled in but I removed -u for clock -s in /etc/rc.sys by myself.

I think /etc/rc.sys should exclude -u automatically too.

Thank you!

ghaerr:
> I think /etc/rc.sys should exclude -u automatically too.

Ok, I will add that to next PR.

Are you running with RTC set to localtime with these changes (instead of UTC?)

tyama501:
Yes, I am always using RTC=local time.

Oh, this is not my case but how about the case RTC=UTC and the local time is not UTC?
(I don't think old PCs before world wide web using UTC but...)

In this case TZ will be set  but -u is needed...

ghaerr:
> Oh, this is not my case but how about the case RTC=UTC and the local time is not UTC?
> In this case TZ will be set but -u is needed...

Very good point! Also QEMU scripts are typically being run with -rtc base=utc, which is just that case.
It sounds like additional thought is needed before making change removing clock -u (when CONFIG_TIMEZONE compiled in).

@Mellvik, are you using UTC or localtime on your real PCs? Any thoughts on this?

We could add yet another config option, CONFIG_RTC_IS_UTC, but that would have to translate to an environment variable to be checked in /etc/rc.sys. I don't want to commit to something without fully thinking this through or it could get messy.

ghaerr:
Reading up [a little more](https://forums.linuxmint.com/viewtopic.php?t=265008) on this, it seems that Windows wants RTC in localtime, but Linux needs RTC in UTC. 

Thus, since ELKS is supposed to be Linux, it seems perhaps we should keep clock -u flag, keeping RTC in UTC, and make things work on ELKS using TZ= for localtime offset.

Comments?

tyama501:
@ghaerr ,

Thank you for the information.
For PC-98, Linux is not major so I think most PC-98 has RTC with localtime.
Is it possible to have different rc.sys for CONFIG_ARC_PC98?

ghaerr:
> For PC-98, Linux is not major so I think most PC-98 has RTC with localtime.
Is it possible to have different rc.sys for CONFIG_ARC_PC98?

Yes - but since clock.c is specially modified for CONFIG_ARCH_PC98, it could display message "PC-98: ignoring -u flag, assuming RTC is in localtime", and treat RTC as localtime without need for special modification to /etc/rc.sys.

What I don't understand yet is whether the testing that you are doing will work if RTC is set in UTC. It seems all your testing requires RTC in localtime, correct?

Doesn't running clock -u for you read the RTC in UTC (ignoring the TZ= string from /etc/profile)?

tyama501:
I mean I am also using dos.
I think dos need localtime...

ghaerr:
Hello @tyama501,

Since date command is working with mods, perhaps only thing that needs to be done is to (reversing my previous thought) add config option CONFIG_RTC_LOCALTIME. When this option set, /etc/rc.sys will (somehow) not use clock -u option. The PC-98 build would default to COFNIG_RTC_LOCALTIME, and /etc/rc.sys would just run clock -s.

This would allow all other users using RTC=UTC to operate without modification, unless they also wanted to set RTC in localtime.

Going this route also has the benefit that QEMU scripts can continue to operating using RTC set to UTC.

PR #1278, opened 2022-05-04
[build] Fix build with new CONFIG_TIME_ options

ghaerr:
Allows hardware real time clock to be in localtime by removing `clock -u` option in /etc/rc.sys during image build, discussed in https://github.com/jbruchon/elks/pull/1274#issuecomment-1117598077.

Fixes package manager image builds when timezone compiled in (now using CONFIG_TIME_TZ).

Fixes `date -s` to enter date/time in localtime, discussed in https://github.com/jbruchon/elks/issues/1273#issuecomment-1117594100.

Renames CONFIG_TIMEZONE to CONFIG_TIME_TZ.
Adds CONFIG_TIME_RTC_LOCALTIME=y option to remove clock -u in /etc/rc.sys (default OFF).

`pc98.config` defaults to CONFIG_TIME_RTC_LOCALTIME=y and CONFIG_TIME_TZ="JST-9".


ghaerr:
Hello @cocus,

After deep diving into all the recent time/date/clock issues, I've realized that clock.c is the ONLY code that gets RTC time/date information into/out of the hardware RTC and ELKS system time. There is no kernel interface to RTC, and after looking hard at this, there probably shouldn't be at this point. 

That means that your planned 8018x RTC implementation should be ifdef'd into clock.c source by simply writing read/write routines using IN/OUT instructions. Other parts of clock.c will correctly translate to/from UTC and/or localtime, based on the -u option, which this PR automatically manages.

It will be interesting to see whether the 8018x RTC is in fact the same as the IBM PC, in which case there may not need to be any changes made to clock.c. After getting this working, some discussion can be had about moving clock.c RTC internals over to the kernel, but I'm not sure its worth it at this point, as nowhere else is the RTC accessed than in clock.c presently.

Thank you!

cocus:
> It will be interesting to see whether the 8018x RTC is in fact the same as the IBM PC, in which case there may not need to be any changes made to clock.c. After getting this working, some discussion can be had about moving clock.c RTC internals over to the kernel, but I'm not sure its worth it at this point, as nowhere else is the RTC accessed than in clock.c presently.

Well, it is, but the way to access it is different. That's it. So yeah, most likely I'll end up adding some #ifdefs on the clock.c source to accomplish this.

ghaerr:
> Well, it is, but the way to access it is different. That's it.

Super. I'll be committing this shortly, and should be able to stay out of your way on any clock.c enhancements for 8018x provided this PR tests well!

PR #1279, opened 2022-05-04
[ash] Add internal test, [ commands to shell

ghaerr:
Adds `test` (and `[`) command as `ash` shell builtin. This should improve boot and shell script speed on floppy systems (running /etc/rc.sys, etc).

Turns out `test` was already compiled in, just never activated. Was able to actually shrink shell size removing unused builtins and bcc workarounds from 1998.

Removes /bin/true, false and test, as these external commands now not needed.
Adds `df` to 360k floppy package manager image build.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -51,9 +51,6 @@ sys_utils/mount			:boot	:sysutil
 sys_utils/umount		:boot	:sysutil
 sys_utils/clock			:boot	:sysutil
 sys_utils/shutdown		:boot	:sysutil
-sh_utils/test			:boot	:shutil
-sh_utils/true			:boot	:be-shutil
-sh_utils/echo			:boot	:be-shutil
 sh_utils/uname			:boot	:be-shutil
 file_utils/cat			:boot	:be-fileutil
 file_utils/chgrp				:be-fileutil			:1200k	:1440k
@@ -61,7 +58,7 @@ file_utils/chmod				:be-fileutil			:1200k	:1440k
 file_utils/chown				:be-fileutil			:1200k	:1440k
 file_utils/cmp					:be-fileutil			:1200k	:1440k
 file_utils/cp					:be-fileutil	:360k
-file_utils/df					:be-fileutil			:1200k	:1440k
+file_utils/df					:be-fileutil    :360k
 file_utils/dd					:be-fileutil			:1200k	:1440k
 file_utils/mkdir				:fileutil		:360k
 file_utils/mknod				:fileutil		:360k
@@ -95,7 +92,10 @@ sh_utils/basename				:be-shutil			:720k
 sh_utils/clear					:shutil					:1200k	:1440k
 sh_utils/date					:be-shutil		:360k
 sh_utils/dirname				:be-shutil			:720k
-sh_utils/false					:be-shutil				:1200k	:1440k
+sh_utils/echo			        :be-shutil              :1200k  :1440k
+#sh_utils/test			        :shutil                 :1200k  :1440k
+#sh_utils/false					:be-shutil				:1200k	:1440k
+#sh_utils/true					:be-shutil				:1200k	:1440k
 #sh_utils/logname				:shutil					:1200k	:1440k
 #sh_utils/mesg					:shutil					:1200k	:1440k
 sh_utils/stty					:shutil					:1200k	:1440k
--- a/elkscmd/Make.install
+++ b/elkscmd/Make.install
@@ -140,7 +140,7 @@ ifdef CONFIG_APPS_1232K
 endif
 
 ifdef CONFIG_APPS_1440K
-	TAGS = :boot|:360k|:net|:720k|:1440k|:man|
+	TAGS = :boot|:360k|:net|:720k|:1200k|:1440k|:man|
 endif
 
 ifdef CONFIG_APPS_2880K
--- a/elkscmd/ash/Makefile
+++ b/elkscmd/ash/Makefile
@@ -27,10 +27,13 @@ OBJS=	builtins.o cd.o dirent.o error.o eval.o exec.o expand.o input.o \
 	jobs.o mail.o main.o memalloc.o miscbltin.o mystring.o nodes.o \
 	options.o parser.o redir.o show.o signames.o syntax.o trap.o \
 	output.o var.o init.o \
-	bltin/echo.o bltin/expr.o bltin/regexp.o bltin/operators.o \
 	linenoise_elks.o autocomplete.o
 
-LDFLAGS += -maout-heap=8192 -maout-stack=2048
+# builtins must be manually added
+OBJS += bltin/echo.o
+OBJS += bltin/expr.o bltin/regexp.o bltin/operators.o
+
+LDFLAGS += -maout-heap=8192 -maout-stack=2304
 
 #
 # Set READLINE in shell.h and add -ledit to LIBS if you want to use the
--- a/elkscmd/ash/README.elks
+++ b/elkscmd/ash/README.elks
@@ -91,3 +91,10 @@ TODO:
 (*) Added #define _SMALL_ in shell.h, which when defined removes some
     features that are no use to elks, such as checking for new mail.
 --------------------------------------------------------------
+20220504 Greg Haerr <greg@censoft.com>
+
+Removed Claudio's workarounds for bcc problems
+
+Activated test/expr builtin in build (was already being compiled in unused)
+
+--------------------------------------------------------------
--- a/elkscmd/ash/bltin/expr.c
+++ b/elkscmd/ash/bltin/expr.c
@@ -172,21 +172,8 @@ overflow:		error("Expression too complex");
 	    while (opsp < &opstack[STACKSIZE] && opsp->pri >= pri) {
 		  binary = opsp->op;
 		  for (;;) {
-			/*
-			 * 19980218 Claudio Matsuoka <claudio@conectiva.com>
-			 * Oh my, this is a nasty one! Once again, bcc
-			 * complains about invalid indirect to indirect
-			 * (mov byte ptr -9[bp],#_op_argflag[bx]).
-			 */
-#if 0
 			valsp--;
 			c = op_argflag[opsp->op];
-#else
-			char *kludge;
-			kludge = op_argflag;
-			kludge += opsp->op;
-			c = *kludge;
-#endif
 			if (c == OP_INT) {
 			      if (valsp->type == STRING)
 				    valsp->u.num = atol(valsp->u.string);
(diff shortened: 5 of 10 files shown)


cocus:
Does this means the shell used for ROMFS would also benefit from this?

ghaerr:
> Does this means the shell used for ROMFS would also benefit from this?

No, not if you're using `sash`, which I think you are. This is because sash doesn't have any scripting capabilities at all. You would need to install ash as /bin/sh and sash as /bin/sash in order to take advantage of this, which is what happens on larger disk PC systems.

Removing true, false and test also don't affect sash, as it lacks the capability to interpret the result of those commands.

cocus:
Right. I mixed the shell names!. At this point in time, how big is ash? I can't tell now how much is it for sash, but I would like to switch to something more "complex" if the space on the ROMFS allows to.

ghaerr:
> At this point in time, how big is ash? 

51040 bytes. Huge or tiny, based on where you're coming from! (For ROMFS, kinda big, as we're quite limited by having to have ROMFS mapped into the 640K address space, unless you can think of a way to place it elsewhere?)

There are ways to disable more builtins in ash, but each has to be crafted somewhat manually at this point. Disabling jobs comes to mind, for instance. See elkscmd/ash/builtins.table for the list.

> I can't tell now how much is it for sash

35664 bytes. Not exactly tiny I suppose compared to ash. Sash can also be stripped down heavily also, fairly easily.

ghaerr:
> we're quite limited by having to have ROMFS mapped into the 640K address space

@cocus - how big is your current ROMFS and BIOS and where do they start? I suppose you have a potential 512K if you start ROMFS at 512K and have a small BIOS just below 1M. There may be issues with having a large ROMFS, I can't remember if I enhanced it to allow being larger than 64K but I think I did. IIRC, there is a limit of 64K for each file within the ROMFS (as it uses a single segment pointer to access each file), but we should be able to have a much larger ROMFS if you have a place to put it on 8018x architecture.

cocus:
Yeah, I don't have the numbers here, but indeed my flash is 512kB. Loader takes around 4k at the very end of it. Then I have the romfs at the specific address required by elks, and in the other remaining place the kernel. Have a look at my loader repo here https://github.com/cocus/8018x-elks-ldr/blob/main/Makefile:
```
ROM_ELKS_IMAGE_AT = 917504 # 0xe0000
ROM_ELKS_ROMFS_AT = 524288 # 0x80000
ROM_LOADER_AT = 1044480 # 0xff000
```

My makefile then subtracts the offset of the flash (which is at exactly 512kB from the beginning of the address map) for the `dd` commands.

cocus:
There are 393216 bytes available for the ROMFS to use (i.e. the gap between 0x80000 and 0xe0000, where the kernel is). Of course the kernel can't use more than 126976 bytes, which is kind of amazing that it has less space than the ROMFS. In any case, the ROMFS is actually at the beginning of the flash, followed by the kernel, and finally the loader. I'm a 100% able to move these around, but if that's needed, I should change some stuff on the kernel for it to work on the new place for it and the ROMFS.

ghaerr:
> ROM_ELKS_IMAGE_AT = 917504 # 0xe0000
ROM_ELKS_ROMFS_AT = 524288 # 0x80000
ROM_LOADER_AT = 1044480 # 0xff000

Ok - pretty cool! It looks like you should be able to support a 384K ROMFS at 0x80000, going up to ELKS start. I've just checked the ROMFS sources and it will support that, although it doesn't (yet) support file sizes over 64K. The only executable over 64K is `vi`, and everything else is under 51K. We can add support for larger file ROMFS file sizes if you need vi with not too much work :)




Screen Shot 2022-05-04 at 6 05 59 PM

Do you have all the networking in ROMFS now? How big is it now?

cocus:
> Do you have all the networking in ROMFS now? How big is it now?

I do, and this is the output of my ldr:
```
524288 bytes (524 kB, 512 KiB) copied, 0.00633173 s, 82.8 MB/s
 Adding elksldr.bin at 520192
4092+0 records in
4092+0 records out
4092 bytes (4.1 kB, 4.0 KiB) copied, 0.00856533 s, 478 kB/s
 Adding Image at 393216
65536+0 records in
65536+0 records out
65536 bytes (66 kB, 64 KiB) copied, 0.134337 s, 488 kB/s
 Adding romfs.bin at 0
293737+0 records in
293737+0 records out
293737 bytes (294 kB, 287 KiB) copied, 0.556237 s, 528 kB/s
```
This means, romfs=293737, kernel=65536, loader=4092 (all in bytes).




> Ok - pretty cool! It looks like you should be able to support a 384K ROMFS at 0x80000, going up to ELKS start. I've just checked the ROMFS sources and it will support that, although it doesn't (yet) support file sizes over 64K. The only executable over 64K is `vi`, and everything else is under 51K. We can add support for larger file ROMFS file sizes if you need vi with not too much work :)

I hate `vi`, but it would be nice to have eventually. I think having `ash` is a better thing than having `vi` for the time being. By the looks of it, it should be doable right now, isn't it?

ghaerr:
> This means, romfs=293737, kernel=65536, loader=4092 (all in bytes).

Super, you've definitely got room in ROMFS for ash (renamed /bin/sh). Also, your kernel is relatively tiny so you could move that up to 0xE8000 (or 0xE7000 for some spare room), which would allow an even bigger ROMFS).

> By the looks of it, it should be doable right now, isn't it?

Yep - just change out sash for ash as /bin/sh and you should have full scripting and a much nicer shell, complete with history/editing (using arrow keys).

I just noticed some other unintended bloat in ash and think I can make it even smaller.

cocus:
> Yep - just change out sash for ash as /bin/sh and you should have full scripting and a much nicer shell, complete with history/editing (using arrow keys).

Only on the menuconfig? Or something else I should take care?

ghaerr:
> Only on the menuconfig? Or something else I should take care?

Yes, I think if your config shows CONFIG_APP_ASH=y, that should do it.

We don't quite yet have automatic support for this yet, but if you added /bin/init to ROMFS (somehow), as well as /bin/getty, /bin/login, /etc/inittab and /etc/rc.sys, then the system would automatically ru  /etc/rc.sys with the ash shell, and you could also use `net start` / `net stop` which interprets /etc/net.cfg as well. That would be kind of cool, but we should calculate the sizes of all that stuff before trying it. I suppose there ought to be a config option to add /bin/init stuff (or not).

Another way to do this would be to add ':romfs' into elkscmd/Applications, which would let you finely tune each executable you want added to ROMFS (or perhaps use another replacement Applications table, named RomFSApplications). See elkscmd/Make.install for details on how it (package manager) works.

cocus:
> Another way to do this would be to add ':romfs' into elkscmd/Applications, which would let you finely tune each executable you want added to ROMFS (or perhaps use another replacement Applications table, named RomFSApplications). See elkscmd/Make.install for details on how it (package manager) works.

Yes, this would make sense.



> > Only on the menuconfig? Or something else I should take care?
> 
> Yes, I think if your config shows CONFIG_APP_ASH=y, that should do it.

I'll check, because I think I've tried this in the past, but I didn't get it on the romfs.

cocus:
Ok, I think that actually adds it, but then I don't have any utilities, like `ls`, `meminfo`, `ps`, etc. Should I also enable busy_elks? Or is this something related to some apps not being bundled because of the PM?

ghaerr:
> Ok, I think that actually adds it, but then I don't have any utilities, like ls, meminfo, ps, etc.

I see. And ls is built in to sash, so with ash only you can't even list files. 

> Or is this something related to some apps not being bundled because of the PM?

The problem with the original (non-PM) mechanism using CONFIG_APP_FILE_UTILS, CONFIG_APP_SYS_UTILS etc is that the granularity is way too large - for both 360k/720k floppies as well as ROMFS. That's why the package manager (using elkscmd/Applications) was written. It sounds like its time for a version of that for ROMS. 

I think by adding a similar but seperate elkscmd/Romfs file, which specifies each application desired as well as each file under rootfs_template to include, you could precisely control which files were added, especially since ROMFS space is limited. It might also be a good idea to have a couple of separate tags used in this file, like ":64k", ":128k", "256k" so that different sized ROMs could be accommodated easily. I can look into making this work, but we should start with creating a Romfs file (copied from Applications with the floppy size tags replaced with rom size tags), to understand the priority of what applications to include with what ROM sizes.

Can you take a look at elkscmd/Applications and give me an idea of what programs you'd like to have? This can be done by adding an appropriate romsize tag after each program desired, or nothing to exclude the file. The slightly bigger problem is we probably also want to have finer granularity over what portions of elkscmd/rootfs_template are included, like which files in rootfs_template/etc also.


> Should I also enable busy_elks?

No, unfortunately busy_elks is unmaintained, and suffers from being completely out of date with regards to updates to all the utilities it included. This is because the command .c files were duplicated, and they haven't been kept up for probably two years.

cocus:
> Can you take a look at elkscmd/Applications and give me an idea of what programs you'd like to have? This can be done by adding an appropriate romsize tag after each program desired, or nothing to exclude the file. The slightly bigger problem is we probably also want to have finer granularity over what portions of elkscmd/rootfs_template are included, like which files in rootfs_template/etc also.
> 

So this is a good question. Sadly I don't understand how this file works, and what gets picked up, and how apps are grouped together. Nevertheless, I think the 720k group option should fit my needs. Not all apps make sense in this scenario, like `makeboot`. If you ask me, I would add all of them :) but no space.
Can you explain how the apps are picked up? How the grouping scheme works?
Thanks!

ghaerr:
> Sadly I don't understand how this file works, and what gets picked up, and how apps are grouped together. 
> Can you explain how the apps are picked up? How the grouping scheme works?

Take a look at elkscmd/Make.install. Basically, a TAGS= variable is set, which is concatenated into a longer string of tags, based on increasing disk size, then awk is used to essentially grep the file for these tags, parsing out the tag and producing a file list. Don't worry about all that, other than that we want to keep the applications on the left side, and then a (possible group) of tags to the right of each application, indicating the minimum size ROM it would be included on.

There's a couple of tricky :: (=rename file) and ::: (=sym link to file) that are used as special case tags, all handled in the complicated do loop at the bottom of Make.install. But don't worry about that either, none of that will be rewritten. We'll just use a different set of TAGS= and Romfs instead of Applications in Make.install.

> Nevertheless, I think the 720k group option should fit my needs.

No, that'll be quite a bit over your maximum of 384k!

> Not all apps make sense in this scenario, like makeboot.

Exactly. That's why you just go through each file, and if you don't want it, don't add any tag. You can start with :boot and some of the :sysutil, along with some of the :360k and :net stuff. Take a look and it should make sense. Don't worry about how to parse it, I'll take care of that part. If you want to start basic, just add :romfs after the apps you need, and we can later divide them into more important 64k - 384k divisions.

cocus:
I see, so the :360k and such floppy disk sizes are another kind of group. So each tag is either an "image size" group OR "apps groups" themselves.
So, I agree that 720k exceeds our target, maybe 360k is a good choice, but it's true that we'll need `ps`, `rmdir`, etc. They're only marked for 720k and up.
I think that in the end we'll have to go ahead with the 360k selection and maybe hand pick some apps (like `ps`) from the config dir. That being said, this would mean we don't need to create a different Romfs file. Just piggyback the 360k option of the Applications file. Maybe we can add a mirror tag named "romfs360k", not sure.

ghaerr:
Yeah, that's pretty much how it works.

The :360k tag isn't going to work since you want networking, and that doesn't begin to fit on 360k with the other 360k stuff.

I'll take a pass at this. Can you post an `ls -lR /` running from your current rom image? That's pretty much what is needed to start, and getting your current list will help with my sizing, rather than starting from scratch. Go ahead tell me what else you think you want and I'll add those too (like ps, meminfo, etc that are not on the current 360k). We may be able to use the existing Applications file, not sure yet.

PR #1285, opened 2022-05-06
[build] Fix .S assembly with new rule and --32-segelf flag

ghaerr:
Fixes .S file assembly problem discussed in https://github.com/jbruchon/elks/issues/1238#issuecomment-1118971422.

@tyama501, after merging this commit, you can remove the temporary .S.o rules in basic/Makefile. Thank you!


tyama501:
Thank you!

PR #1287, opened 2022-05-08
[net,ne2k] Major driver update adds full support for 8bit interfaces

Mellvik:
**Summary**
This update enhances the NE2K driver to fully support 8bit interfaces, in particular RTL8019 based interfaces, and 16 bit interfaces in 8bit mode (for 8bit ISAbus). It also adds autoconfiguration and (optionally) bootopts configuration of interface options. The buffer overflow handling has been rewritten and optimized. Error handling has been dramatically improved, catching and recovering from all known and (reasonably) conceivable error conditions. The 8bit support has 2 modes - safe and slow, fast and experimental - configurable via bootopts.
The update also includes a number of general enhancements and bugfixes.
As far as hardware configurations are concerned, there are 4 possible scenarios to accomodate:
1. 16 bit interface in 16 bit ISA bus
2. 16 bit interface in 8 bit ISA bus
3. 8 bit interface in 16 bit ISA bus
4. 8 bit interface in 8 bit ISA bus

The former version of the driver really accomodated only 1), others would possibly work under 'benign' conditions with configurable 8bit mods. This version accomodates 1, 2 and 3. 4 needs testing on real 8bit hardware. 2 has been tested by forcing 16 bit interfaces to run 8 bit mode on 16bit hardware. Also, there are variations between socalled 'compatible' chips, so mileage may vary. And finally, the QEMU NE2K emulator does not work if put in 8 bit mode. It simply hangs.

**Usage**
- The default ELKS config will accommodate both 16bit and 8bit interfaces, not changes required except changing the MTU for 8bit (see below)
- Adding CONFIG_ETH_BYTE_ACCESS will force 8bit mode, but is not required
- Adding '1' to the `netport` address will also force 8bit mode
- Adding '2' to the `netport` address will force 16k buffer usage even if the interface is 8bit
- When using the default configuration, change MTU til 720 for the 8bit interface

Almost needless to say, what started as a bughunt & expeced (fast) fix turned into a project - which eventually delivered some very unexpected results. What follows is an edit of my 'lab notes', sorry about the verboseness.

**The details**
While reorganizing some hardware recently, it was a good time to check out the 8bit NE2K interface acquired a while back. The RTL8019AS based interface from Weird Electronics (via Tindie) is as simple as they come - at a reasonable price. Also, it has been tested with ELKS before and found working after minimal driver changes introduced by @cocus.

No such luck in my case, possibly thanks to a faster machine (386/20). The interface was functional but unstable with telnet, file transfers were not possible (hung), and there was no apparent predictability in the errors. 

Literally an open invitation to take the dive, understand and fix. Luckily I had no clue as to what I was getting myself into: A long (and admittedly very frustrating) chase - I fell into the infamous rabbit hole which eventually turned out to be a cave :-).
The symptoms were lots of oversized packets in read, subsequently a hung NIC. Eventually, the problem was tracked down to an elusive memory corruption issue in the NIC. The type of bug that makes no sense because the problem keeps moving as you close in. Apparently oversized packets, actually corrupted buffer headers sending the NIC's internal dma transfers astray.

Almost accidentally, the book **Networking and Internetworking with Microcontrollers** (Fred Eady) popped up via Google (used copies available on eBay). The book devotes quite a bit of time to the RLT8019 controller and - among lots of other interesting things - reveals something I've not found in any spec:

8-bit mode, both the CS8900A-CQ and the RTL8019AS use b on-chip RAM area

4k buffer max. So simple yet so elusive. Reducing the NIC RAM buffer size to 4K and the card ran like a champ. Well, not quite (see below), but a lot better. 

4K of buffer space is VERY limiting. 16 buffer blocks @ 256 bytes each, 6 of which must be dedicated to the transmit buffer. So we have 10 blocks for receiving packets, and one full size packet takes 6 blocks. Needless to say, incoming transfers create lots of overruns - with grave consequences for transfer speed while not generally a problem for telnet sessions and outgoing transfers. 

When receiving files, almost every other packet initially got 'lost' - discarded because of buffer overflow - and had to be retransmitted. The good news (yes, it's a stretch) - we now have REAL (and extensive) packet loss, which is something we haven't been able to expose ktcp to before. Fits right into the ongoing ktcp/elks-to-elks testing.

The work on handling the frequent buffer overflows exposed another elusive problem: Occasional mysterious NIC hangs during incoming file transfers, unrelated to the initial problem. Never on small files, always on bigger (>10-15k) files that required storage access (i.e. significant delay) while receiving. Another endless chase revealed that NIC buffer overflows sometimes occur while other NIC interrupts are being processed. Not surprising in itself, but it was assumed to be handled by the current setup. 

It wasn't. We had lost receive overflow interrupts, which in turn hung the NIC. Losing other interrupts is a recoverable problem, losing overflows is serious because the NIC has to run through a specific procedure to recover - every time. 
Simple fix when nailed down - just a while loop in the interrupt handler, (re-)checking pending interrupts at the top of the loop. There may be more elegant ways of doing this.

While researching the problem and testing solutions, I found that even receive and transmit interrupts occasionally occur while the interrupt service routine is running - and get lost. This may in some cases cause delays because ktcp is not aware that a) more data is available or b) a transmit has completed and the NIC is ready for the next. Pure speculation at this point, but this may be part of the occasional problem we've seen with 'hit a key to get telnet to continue'. To be investigated/tested.

The third big challenge was to get decent performance with the ultrasmall buffer. Other systems handle this via internal (kernel) buffering, ELKS avoids such buffering for many reasons. Hundreds of tests led to the preliminary (and unsurprising) conclusion that two parameters determine the file transfer speed (incoming): The MTU and the ktcp retransmit timeouts. Finding a suitable MTU reduces the number of NIC overflows, while getting the ktcp retransmits to 'match' the delay introduced by overflows, reduce the 'recovery' delay after each overflow.

For the record, here are some of the options considered and tested in the hunt for decent  performance:

- Adding a single packet receive buffer inside the driver may make a significant difference, this has not been tested.
- Stealing away 2x256b from the send buffer, thus reducing it to 1024 bytes and allowing two full size incoming packets to be buffered. Result: Some improvement - and  some unpleasant surprises, such as unpredictable results when pinging with large (>1k) packets (large outgoing packets will overwrite the read buffer). This can be fixed, but the benefits don't seem to outweigh the complications.
- Manipulate the MTU to take the strained buffer situation into account (add mtu variable to `net.cfg`): Testing indicates that MTU 700 (actually any number that keeps the received packet below 768) improves file transfer speed significantly. Further reductions (all the way down to one buffer page, MTU approx 210) make little difference. Mileage will probably vary. An option to set a lower MTU in net.cfg is part of this PR.
- Pushing the limits: Carefully go beyond the 4k limit on the NIC and see what happens. It's not that the rest of the buffer is unavailable in 8bit mode. The problem is that using it cause unpredictable behaviour as described above. Using 8K instead of 4k was tested, it did not go well. 5k looked more promising, but failed under heavy load. 
- Frequent overflows means that the overflow 'strategy' can make a big difference. The overflow handler has been completely rewritten as part of the testing, and the following strategy has been found suitable:
    - Always keep one packet (the oldest, which is always safe in the sense that it is undamaged). This is the default for 8bit.
    - When using the entire 16k buffer,  keep 3 or 4 packets (this selection may be automated). The default for 16bit is 3. 
    - Purging the entire buffer is almost as good as keeping 1 packet in the 8bit case (set `_ne2k_skip_cnt` to 0). Whether it makes a difference for ktcp, has not been tested.
    - It may be worthwhile to tune the skip_count value per machine in order to get optimal results with 8bit interfaces. Lower MTU -> higher skip_count.

Having the challenges above under control, the natural next (and presumably final) step was to ensure robustness. The driver should keep the NIC going (no crash, no hang) even in the most extreme situations (flood ping large packets from several powerful machines). Continuous long lasting overflows, no packets coming through until the flood retreats - recovery must work. 

After another endless series of fixing and testing, this goal appears to have been achieved. `ktcp` sometimes hangs, the driver does not.

The robustness enhancements required adding more error handling to the driver, to detect and handle individual error types. The previous driver left such error handling to the NIC, by telling it to ignore small, big and bad packets. This worked fine in normal situations, but not under heavy load with overruns, overflows, general overload.

At this point, the 'project' was finished, goal achieved:  8bit interfaces fully, reliably supported plus a host of enhancements. I was preparing the code (and notes) for PR.

But there still was an itch. One more test, what if there is a way to beat the 4k limit after all? Get decent performance from an 8bit interface? The performance problems caused by the limited 4k buffer was a strong incentive to take an extra round. And weeks of testing had revealed at least some patterns in the NIC behaviour when running 8bit/16k and failing.

The effort was not successful, the 'close but no cigar' variant: Improvements, but still the elusive hangs. Then - during final cleanup for the PR, a small optimization: Abort NIC reads immediately if a buffer header is found to be suspicious. It wasn't expected to have any effect on the buffer issue, just a cleanup. And the effect was discovered by accident: I had forgotten to turn off the forced 8bit/16k mode on a 16bit NIC. The error messages during a file transfer didn't make sense. What did I screw up this time? Also, it seemed slow. Until I discovered the mistake. A real 'wow'-moment.

Another round of tests and adjustments, a move to the real 8bit interface, more flood pings, actually tens of thousands of packets, and no hang. Incredible.

Eventually the final test: Remove the reduced MTU on the 8bit NIC machine. Then repeated incoming big file transfers - typically a 1.47M file to `/dev/null`. Still no hangs. Read errors, oversized packets, checksum errors, buffer resets yes, but few compared to what I'd seen before. And the numbers were incredible. File transfers clocked in at just a few percent slower than the 16bit 'real thing'.

Wow. Too good to be true. Another PR delay, but then again, no one was waiting. Now we have two 'working' setups for 8bit interfaces:  One safe and slow, one fast and (still) experimental. That's a lot better than expected just a few weeks back.

**Practical**
- For 16bit interfaces, there are no practical differences, just improvements: stability, possibly marginal speed improvements.
- The 8bit interface (4k buffers) will occasionally report transmit errors and frequently report receive errors under heavy load. These are reported, counted and ignored. In the read case, there is nothing to do. In the transmit case, the packet could be retransmitted, but this may as well be handled by `ktcp`. Adding retransmission logic at the driver level does not make sense. 
- When receiving files (8bit/4k mode), there are frequent buffer overflow errors and many other errors. They can be ignored, recovery works.
- When using the full 16k buffer in 8bit mode, the driver will report occasional oversized packets on read and `ktcp` will complain about checksum errors and read errors. These are recoverable and add minimal delay.
- The new error counters in the driver are globals, available for inclusion in `netstat` eventually.
- Given the number of error messages from the driver when using the 8bit interface, it may be necessary to encapsulate (hide) the messages in debug_eth and activate when required with ^P.

**The good news 1:** The 8bit interface in 4k mode works fine with several concurrent telnet sessions and outgoing file transfers. It is not fast, but stable and reliable for all kinds of file transfers. `ftp` a 1.47M file from Linux (Raspbian) to Elks 8bit, MTU 700 (to `/dev/null`) takes 145 seconds (10k/s). A similar transfer using the 16bit interface, 1500 MTU and no overruns clocks in at 55secs (26k/s). Interestingly, repeating the same transfer to the 16bit ELKS client while running a `ping -i 0.1 -s 800` concurrently, hardly affects the transfer speed at all, indicating that the speed (file transfer) is CPU (ktcp) bound, not hardware/driver bound.

**The good news 2:** Spending all this time in the NE2k driver (again) was opportunitiy for some optimizations and cleanups. Here's the list:
- Using menuconfig to set 8-bit mode is no longer required. All the conditional compile statements for 8bit cards are gone - except one. The driver will autoconfigure predictably on 16 bit machines, needs testing on machines with 8bit ISA only (and emulators). E.g. QEMU ne2k emulates the NIC PROM, but behaves differently than physical so special precautions had to be taken.
- 8bit mode may be forced by either setting the least significant bit in the IO address in `/bootopts` to 1, e.g. `netport=0x301`, or by using the traditional config option. The ability to force 8bit mode is useful for debugging, running a 16 bit interface in 8 bit mode even if the bus is 16 bit. And to force 8 bit mode when using a 16 bit interface in a 8bit-ISA machine.
- If the next bit of the IO address is set in `/bootopts` (e.g. `netport=0x303`), the NIC is forced to use the full 16k buffer space even if it's 8bit. See notes above.
- The driver will report 8 or 16 bit modes during boot.
- Optimizations and improvements
    - Receive error interrupts are now enabled, necessitated by the frequency of errors from the 8 bit interface.
    - Tally counter (error counter overflow) interrupts are processed in order to get rid of the extra (and confusing) status bit.
    - The buffer overflow handling really got a beating and ended up being almost completely rewritten. As mentioned above, various recovery strategies were tested. I ended up with something functionally close to the previous version, but leaner and fully tested with the ability to adapt to the requirements of different interface variants.
    - The code executed at open/release/init(Boot) has been reorganized and may still take some improvement. There are still occasional NIC interrupts before the device has been opened, which should not happen. To be fixed.
    - Some duplicated code has been removed and lots of comments added - reminders and experiences.
    - Counters for receive, transmit, overflow errors have been added for future inclusion into `netstat`
The changes have been extensively tested - using 3 different interfaces (2 16bit, 1 8bit) on physical machines (compaq 386/20). Limited QEMU testing revealed 2 QEMU ne2k bugs: One related to the reading of the HW address PROM and one related to interrupts. Both are handled in the code (and commented). There may be more.

In retrospect it may seem like sticking with a manual 8 vs 16 bit setting (via the `netport` address) would be good enough - and simpler than the automated discovery. I'm open to opinions about that. My guess is that auto-detection adds less than 50 bytes of code total.

Finally - in terms of pure speed outside the context of the overrun situations, as measured by 1 second pings w/ packet size 1200, the 8 bit interface is about 20% slower than 16bit, 10.6ms vs. 8.2ms. File transfers are a different story. Telnet sessions are just great. 8bit interfaces should always run with MTU 720 unless 16k buffer is enabled. When in 16k mode, the 8bit interface clocks in 5% behind its 16bit sibling.

The 8bit/16k 'mode' is still experimental and 'under observation'. Thus a number of debug statements have been commented out instead of removed. Given the nature of this 'hack' - going outside hardware specs - it may not work in all environments. More testing is needed to verify this.

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -1,14 +1,32 @@
 //-----------------------------------------------------------------------------
 // NE2K driver - low part - MAC routines
 //
-// Updated by Santiago Hormazabal on Dec 2021:
-//  . 8 bit access if CONFIG_ETH_BYTE_ACCESS is set, using a ne1k patch from
-//    NCommander.
-// Updated by Helge Skrivervik July 2020:
+// Updated by Helge Skrivervik (HS) July 2020:
 //	. pick up MAC address from prom
 //	. fixed read ring buffer wrap around errors
 //	. added ring buffer overflow handling
+//	. use word I/O
 // oct-2021: Pick up I/O port # from init/bootopts (HS)
+// Updated by Santiago Hormazabal on Dec 2021:
+//  . 8 bit access if CONFIG_ETH_BYTE_ACCESS is set, using a ne1k patch from
+//    NCommander.
+// apr-2022 (HS) : Support auto detection of 8bit mode, set 4k buffer size when in 8bit mode
+//	     assuming RTL8019 buffer restrictions.
+//	     Rewrote overflow handler, added direct access to many registers from C code
+//	     Cleaned up initialization code, optimized 8bit I/O
+//
+//-----------------------------------------------------------------------------
+// Terminology
+// The ring buffer pointer terminology used in the DP8390C/NS3249C document is confusing.
+// 'CURRENT' points to the next block to be filled by an incoming packet
+// 'BOUNDARY' points to the next block to be read from the NIC
+// The two must not be equal. If the ring buffer is empty, BOUNDARY = CURRENT -1
+// So the next block to read is not BOUNDARY but BOUNDARY + 1. Easy - except at
+// the wrap-around point: BOUNDARY may be 80 or 50, while the next block to read is 46.
+// For simplicity (to avoid the ring wraparound logic on every read)
+// the driver uses a variable to hold the next block to read - _ne2k_next_pk.
+// For debugging, this variable is also available to the C part of the driver.
+//
 //-----------------------------------------------------------------------------
 
 #include <linuxmt/config.h>
@@ -60,30 +78,35 @@ io_ne2k_reset      = 0x1F	// Really a port, not a register, force HW reset of th
 
 tx_first           = 0x40
 rx_first           = 0x46
-rx_last            = 0x80
+rx_last_16	   = 0x80
+rx_last_8	   = 0x50	// Use only 4k in 8 bit mode 
 
 //-----------------------------------------------------------------------------
 	.data
 	.extern current
 	.extern	net_port	// io-port base
 
+	.global _ne2k_next_pk
 _ne2k_next_pk:
 	.word 0	// being used as byte ...
-
-	.global _ne2k_skip_cnt
-_ne2k_skip_cnt:
-	.word 0	// # of packets to skip if buffer overrun, default is all (0)
 
 	.global _ne2k_has_data
 _ne2k_has_data:
 	.word 0 
+
+	.global _ne2k_is_8bit	// set if this is an 8bit card
+_ne2k_is_8bit:
+	.word 0
+
+_ne2k_rx_last:
+	.byte rx_last_16	// default to the 16 bit value
+
 
 	.text
 
 //-----------------------------------------------------------------------------
 // Set unicast address (aka MAC address)
 //-----------------------------------------------------------------------------
-
 // arg1 : pointer to unicast address (6 bytes)
 
 	.global ne2k_addr_set
@@ -92,7 +115,7 @@ ne2k_addr_set:
 
 	push    %bp
 	mov     %sp,%bp
-	push    %si  // used by compiler
+	push    %si  // used by compiler (???)
 
 	mov     4(%bp),%si
 
@@ -174,8 +197,6 @@ dma_write:
 	inc     %cx     // make byte count even
 	and     $0xfffe,%cx
 	call    dma_init
-	shr     %cx     // half -> word size transf
-
 
 	// start DMA write
 
@@ -185,41 +206,45 @@ dma_write:
 
 	// I/O write loop
 
-	mov	net_port,%dx	// Do this before changing the data segment
-	push	%dx		// Command register
+	//mov	net_port,%dx	// Do this before changing the data segment
 	add	$io_ne2k_data_io,%dx
+	mov	_ne2k_is_8bit,%ax
 	mov	current,%bx		// setup for far memory xfer
 	mov	TASK_USER_DS(%bx),%ds
 	cld
+	test	%ax,%ax		// checking _ne2k_is_8bit
+	jz	1f
 
-emw_loop:
+	// Byte loop
+4:	lodsb
+	outb     %al,%dx
+	loop    4b
+	jmp	2f
 
-#ifdef CONFIG_ETH_BYTE_ACCESS
-    lodsb
-    outb     %al,%dx
-    lodsb
-    outb     %al,%dx
-#else
-	lodsw
+	// word loop
+1:	shr	%cx
+3:	lodsw
 	out     %ax,%dx
-#endif
-
-	loop    emw_loop
-
+	loop	3b
+2:
 	// wait for DMA completed
 
-	pop	%dx	// instead of: mov	net_port,%dx
+	mov	net_port,%dx
 	add	$io_ne2k_int_stat,%dx
+
 check_dma_w:
 	in      %dx,%al
 	test    $0x40,%al       // dma done?
-	jz      check_dma_w     // loop if not
+	jnz	end_dma_w
 
-	mov     $0x40,%al       //clear DMA intr bit in ISR
+end_dma_w:
+	mov     $0x40,%al       // clear DMA intr bit in ISR
 	out     %al,%dx
+	sti			// Experimental
 
 
-	sti		// Experimental
+	mov	%cl,%al		// Error (debug) return
+	xor	%ah,%ah		// if AL is 0xff, we had a timeout, the dma never completed
 	pop     %si
 	pop	%ds
 	pop	%bx
@@ -228,10 +253,9 @@ check_dma_w:
 
 #if 0
 //-------------------------------------------------------------------------
-// This is an (untested) skeleton routine for DMA-assiste paket transfer
+// This is an (untested) skeleton routine for DMA-assisted packet transfer
 // from the NIC to host memory.
-// TODO: Add DMA channel setup and teardown. Makes sens to do that outiside of this
-// routine.
+// TODO: Add DMA channel setup and teardown. 
 //
 dma_r:	// Use the send data command to read exactly one backet, 
 	// the nic does everything on its own, needs only ES:DI
@@ -279,7 +303,7 @@ rlp_ret:
 // BX    : chip memory to read from
 // CX    : byte count
 // ES:DI : host memory to write to
-// AL:	 : 0: buffer is local, <>0: buffer is far
+// AL:	 : 0: buffer is local (kernel), <>0: buffer is far (process)
 
 dma_read:
 
@@ -288,11 +312,11 @@ dma_read:
 	push	%bx
 	push	%ax
 
-	cli		// Experimental - disable INTR
 	inc     %cx     // make byte count even
 	and     $0xfffe,%cx
+
+	cli		// Experimental - disable INTR
 	call    dma_init
-	shr     %cx     // half -> word size transf
 
 	mov     %ds,%bx
 	mov     %bx,%es
@@ -302,33 +326,35 @@ dma_read:
 	mov	current,%bx	// Normal: read directly into the (far) buffer
 	mov	TASK_USER_DS(%bx),%es
 
-buf_local:
 	pop	%bx
 	push	%bx
 
-	// start DMA read
-
+buf_local:
 	mov	net_port,%dx	// command register
-	mov	$0x0a,%al	// RD0 & STA
-	out     %al,%dx
-
-	// I/O read loop
+	mov	$0x0a,%al	// set RD0 & STA
+	out     %al,%dx		// start DMA read
 
 	mov	net_port,%dx
 	add	$io_ne2k_data_io,%dx
-	cld
-emr_loop:
-#ifdef CONFIG_ETH_BYTE_ACCESS
-    inb      %dx,%al
-    stosb
-    inb      %dx,%al
-    stosb
-#else
+	cld			// clear direction flag
+	cmpw	$0,_ne2k_is_8bit
+	jz	1f
+
+// Byte transfer
+byte_loop:
+	inb      %dx,%al
+	stosb
+	loop    byte_loop
+	jmp	3f
+
+// Word transfer
+1:
+	shr     %cx     // half -> word size transf
+word_loop:
 	in      %dx,%ax
 	stosw
-#endif
-	loop    emr_loop
-
+	loop	word_loop
+3:
 	// wait for DMA to complete
 
 	mov	net_port,%dx
@@ -358,12 +384,13 @@ check_dma_r:
 // 	read point is in the variable _NE2K_NEXT_PK. This trick is necessary
 //	because the internal logic in the NIC will trigger an overrun interrupt
 //	if the BOUNDARY pointer matches or exceeds the CURRENT pointer.
+//---------------
 // Used internally, exposed externally for debugging purposes.
 //
 	.global ne2k_getpage
 
 ne2k_getpage:
-	mov	$0x42,%al		// page 1
+	mov	$0x42,%al	// page 1
 	mov	net_port,%dx	// command register
 	out	%al,%dx
 
@@ -372,11 +399,11 @@ ne2k_getpage:
 	in      %dx,%al
 	mov     %al,%ah
 
-	mov	$0x02,%al		// page 0
+	mov	$0x02,%al	// page 0
 	mov	net_port,%dx	// command register
 	out	%al,%dx
 
-	mov	net_port,%dx
+	//mov	net_port,%dx
 	add	$io_ne2k_rx_get,%dx     // BOUNDARY
 	in      %dx,%al
 
@@ -427,15 +454,15 @@ nrs_exit:
 #else
 	// sep2020: keep ring buffer status in a variable
 	// instead of accessing the NIC registers continuously.
-	movw	_ne2k_has_data,%ax	// TEST !!!!!
+	call	ne2k_clr_int_reg // EXPERIMENTAL, should clear just the read.
+	movw	_ne2k_has_data,%ax
 #endif
 	ret
 
 //-----------------------------------------------------------------------------
 // Get received packet
 //-----------------------------------------------------------------------------
-// arg1 : packet buffer to receive the data. The buffer must be 4 bytes to hold 
-//	the NIC header plus the mac Ethernet frame size.
+// arg1: buffer to receive the data
 // arg2: int - requested read size (max buffer)
 // arg3: int array [2] (return) containing the NIC packet header.
 //
@@ -453,21 +480,22 @@ ne2k_pack_get:
 
 	//sub 	$4,%sp		// temp space
 	//mov	%sp,%di
-	mov	8(%bp),%di	// get the hdr into arg3
 
-	// get RX_put pointer
-
+	// get the 4 byte header first -> arg3
+	mov	8(%bp),%di	
 	mov	_ne2k_next_pk,%bh
 	xor	%bl,%bl		// Next pkt to read in BX
 
-	mov	$4,%cx
+	mov	$4,%cx		// Bytes to read
 	//mov     %ds,%ax
 	//mov     %ax,%es		// local address space
-	xor	%al,%al		// local address space
+	xor	%al,%al		// indicate local address space
 	call	dma_read
 
 	mov	0(%di),%ax	// AH : next record, AL : status
 	mov	2(%di),%cx	// packet size (without CRC)
+
+	// get the actual data
 
 	//add	$4,%sp
 	mov	4(%bp),%di	// Buffer address to receive data.
@@ -478,9 +506,12 @@ ne2k_pack_get:
 
 #if 0
 	// -------------------------------------------------------------
-	// Check packet size - not required since the NIC will not
+	// Packet size check not required since the NIC will not
 	// accept such packets per our initialization. Note, in order to handle
-	// erroneous packets, rx_get (BOUNDARY) pointer must be updated.
+	// erroneous packets, rx_get (BOUNDARY) pointer must be updated
+	// to point to the next packet.
+	// If oversized packets still occur, it's a driver problem (most likely
+	// reading the wrong buffer page).
 	// -------------------------------------------------------------
 	or      %cx,%cx		// zero length
 	jz      npg_err2
@@ -488,6 +519,17 @@ ne2k_pack_get:
 	cmp     $1528,%cx	// max - head - crc
 	jnc     npg_err
 #endif
+	// -------------------------------------------------------------
+	// This section did the smart thing when reading the NIC packet header:
+	// Got the entire block (256b) instead of the 4 first bytes. Which was great 
+	// for small packets (telnet, command packets etc.): One read instead of 2.
+	//
+	// Removed when changing to read directly into the process was introduced,
+	// there is no longer anywhere to put the first 4 bytes.
+	// It may be an idea to reintroduce a 256b buffer to take advantage 
+	// of this 'optimization' and at the same time help strangled 8bit 
+	// interfaces.
+
 	//sub	$252,%cx	// Got entire packet?
 	//jle	npg_cont
 				// If not, get rest.
@@ -500,7 +542,7 @@ npg_cont0:
 				// (keep the 4 byte NIC header)
 	push	%cx		// save length
 	push	%ax
-	add	$4,%bx
+	add	$4,%bx		// Skip the 4 bytes already read
 
 	mov	$1,%al		// use far transfer
 	call    dma_read
@@ -515,7 +557,8 @@ npg_cont:
 	dec	%al
 	cmp	$rx_first,%al
 	jnb	npg_next	// if the decrement sent us outside the ring..
-	mov	$rx_last-1,%al	// make it right ...
+	mov	_ne2k_rx_last,%al
+	dec	%al
 
 npg_next:
 
@@ -523,52 +566,21 @@ npg_next:
 	add	$io_ne2k_rx_get,%dx	// update RX_get (BOUNDARY)
 	out     %al,%dx
 
-#if 0	/* error processing - not used */
-	xor     %ax,%ax
-	jmp     npg_exit
-npg_err:
-	mov     $-1,%ax		// Packet too big
-	jmp	npg_exit
-
-npg_err2:
-	mov	$-2,%ax		// zero length packet
-#endif  /* ---------------------------- */
-
 npg_exit:
-#if 0	/* Handle ring buffer overflow - not used */
-        /* OFLOW handled by its interrupt handler */
-
-	mov	net_port,%dx  
-	add	$io_ne2k_int_stat,%dx  
-	in	%dx,%al			// get the status bits
-	test	$0x10,%al
-	jz	npg_no_oflw
-	push	%bx
-	call	ne2k_clr_oflow
-	pop	%bx
-
-npg_no_oflw:
-#endif
-	// The is effectively the replacement for the rx_stat routine,
+	// This is effectively the replacement for the rx_stat routine,
 	// clear the has_data flag if ring buffer is empty.
-	cli	// Interrupts off (experimental)
-	//mov	_ne2k_has_data,%ax
-	//dec	%ax	// TESTING
-	//cmp	$0,%ax
-	//jz	npg_zero	// don't set to zero unless sure (TESTING)
-	//mov	%ax,_ne2k_has_data
-npg_zero:
+	cli			// Interrupts off (experimental)
 	call	ne2k_getpage
 	cmp	%ah,%bl		// ring buffer empty?
 	jnz	npg_more_data
 	movw	$0,_ne2k_has_data
-npg_more_data:
 
+npg_more_data:
 	sti		// Enable interrupts
 	pop	%ax	// return byte count (from %cx)
 	//pop	%es
 	pop     %di
-	mov	%bp,%sp	// restore stack pointer
+	//mov	%bp,%sp	// restore stack pointer
 	pop     %bp
 	ret
 
@@ -594,24 +606,19 @@ ne2k_tx_stat:
 	jmp     nts_exit
 
 nts_ready:
-
 	mov     $2,%ax
 
 nts_exit:
-
 	ret
 
 //-----------------------------------------------------------------------------
 // Send packet: First transfer packet data to NIC memory, then kick off
 // the actual transmit and return.
 //-----------------------------------------------------------------------------
-
-// arg1 : packet buffer to read from
+// arg1 : packet buffer to transfer
 // arg2 : size in bytes
-
 // returns:
-
-// AX : error code
+//	AX : error code
 
 	.global ne2k_pack_put
 
@@ -632,11 +639,12 @@ ne2k_pack_put:
 	// set TX pointer and length
 
 	mov	net_port,%dx
-	add	$io_ne2k_tx_start,%dx	// FIXME: This may not be required, done
+	add	$io_ne2k_tx_start,%dx	// FIXME: This is probably superfluous, done
 					// at initialization time, never changes.
-	mov     $tx_first,%al
-	out     %al,%dx
+	//mov     $tx_first,%al
+	//out     %al,%dx
 
+	//add	$io_ne2k_tx_len1,%dx	// use inc instead
 	inc     %dx		// io_ne2k_tx_len1
 	mov     %cl,%al
 	out     %al,%dx
@@ -649,15 +657,11 @@ ne2k_pack_put:
 tx_rdy_wait:
 	mov	net_port,%dx	// command register
 	in	%dx,%al
-	test	$0x4,%al	// Check that previous transmit competed.
+	test	$0x4,%al	// Check that previous transmit completed.
 	jnz	tx_rdy_wait
-	mov	$6,%al		// Set TX bit, starts transfer...
+	and	$0x18,%al
+	or	$6,%al		// set TX + STA; keep the others
 	out	%al,%dx
-
-	//mov	net_port,%dx
-	//add	$io_ne2k_int_stat,%dx	// reset tx intr bit
-	//mov	$2,%al		// Test, should not make any difference
-	//out	%al,%dx
 
 1:	mov	net_port,%dx	// command register
 	in      %dx,%al
@@ -674,7 +678,7 @@ tx_rdy_wait:
 // Get NE2K interrupt status
 //-----------------------------------------------------------------------------
 
-// returns:
+// returns interrupt status reg unmodified
 
 // AX : status
 //   01h = packet received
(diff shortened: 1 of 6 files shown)


Mellvik:
@cocus, it would be great if you could take this one for a test drive on 8bit hw!

--M

ghaerr:
Hello @Mellvik!!

Wow. What an incredible report! Absolutely amazing work!!!

Needless to say, I haven't had time for all the things you've learned to sink in, but the finding about the NIC running with a 4K buffer in 8-bit mode is a real gem. I can see how having stats will really help understand what the NIC might be doing, as part of the complex chain of getting continuous network data in and out of ELKS with no hiccups.

I'll have to study this quite a bit, but I'm ready to commit when you are. I have a couple comments - 

- The auto-card detection is great. Is there any benefit (other than testing?) for keeping CONFIG_ETH_BYTE_ACCESS? If not really, then why not remove it - one less problem for less knowledgeable users.
- The `#mtu=720` option in /etc/net.cfg: it appears we need a -m option to set MTU? Should this be `#mtu=-m 720`? Also, no need for previous line `mtu=""`, it can be safely deleted, as non-existent shell variables expand to nothing unless set -u is set. Can you see a way where the MTU could be reduced automatically somewhere?
- Did I see the NIC driver can now determine whether its running on QEMU? If so, perhaps this could also be used to (somehow) set the QEMU= variable, passing upwards to other applications. Not sure that's a great idea yet, but it appears the NIC now needs to automatically know...
- I see a FIXME in ne2k.c, is that something you want left in for now, to be fixed later?

We need to document the special `netport=` and `mtu=` stuff somewhere, so that there's a well-known process to installing a NIC. Most worrisome is mtu=, since its not automatic (yet).

With regards to getting the new NIC status information passed over to `netstat`: the first step in that process will be to put all the variables you want access to in a (preferably single) struct. This might be a bit of a pain if you access those variables from ASM, but you could use defined macros using `offsetof` that build an ASM-usable address of the variable. We can then fairly easily write code to transfer the entire structure over and display it. The struct should live somewhere in elks/include/linuxmt, perhaps a new file netstat.h.

Thank you for your incredible contribution, very well done!!!

Mellvik:
Thank you @ghaerr,

> I'll have to study this quite a bit, but I'm ready to commit when you are. I have a couple comments -
> The auto-card detection is great. Is there any benefit (other than testing?) for keeping CONFIG_ETH_BYTE_ACCESS? If not really, then why not remove it - one less problem for less knowledgeable users.
It was kept for compatibility. I'm more than happy to delete it.
> The #mtu=720 option in /etc/net.cfg: it appears we need a -m option to set MTU? Should this be #mtu=-m 720? Also, no need for previous line mtu="", it can be safely deleted, as non-existent shell variables expand to nothing unless set -u is set. Can you see a way where the MTU could be reduced automatically somewhere?
Sorry about that one, obviously I had the -m in the test version. Will fix.
As to setting the MTU automagically - since the arguments to init() and open() are unused, maybe we can sneak in a return value in one of them? The other alternative, and always safe, is to add an IOCTL to do it.
> Did I see the NIC driver can now determine whether its running on QEMU? If so, perhaps this could also be used to (somehow) set the QEMU= variable, passing upwards to other applications. Not sure that's a great idea yet, but it appears the NIC now needs to automatically know…
Hmmm, yes, the driver is detecting a certain behaviour and concluding it's QEMU. I don't know how reliable the mechanism is though. OTOH - why not make the assumption for now, and fix it when/if it breaks. There are other (more complicated though) ways to do this detection should we need it. We can have an IOCTL for that too, or (less reliable) use the HW address which is always the same in QEMU.
> I see a FIXME in ne2k.c, is that something you want left in for now, to be fixed later?
Thanks - I need to remove that hack.
> We need to document the special netport= and mtu= stuff somewhere, so that there's a well-known process to installing a NIC. Most worrisome is mtu=, since its not automatic (yet).
> 
Indeed!! I guess the priority goes to the wiki. I'm on that. And like I possibly mentioned before, there is a man page for the driver coming up, which must have all these things including the IOCTLs.
I'll update the PR with suggestions for the IOCTLs mentioned here.
> With regards to getting the new NIC status information passed over to netstat: the first step in that process will be to put all the variables you want access to in a (preferably single) struct. This might be a bit of a pain if you access those variables from ASM, but you could use defined macros using offsetof that build an ASM-usable address of the variable. We can then fairly easily write code to transfer the entire structure over and display it. The struct should live somewhere in elks/include/linuxmt, perhaps a new file netstat.h.
> 
It's all in the C code, so we're fine. I'll add that to the updated PR.


We really need more people beating this stuff up on real hardware. 

BTW - I'm really excited about the prospect of being using the full 16k buffer in the 8bit NIC. I'm going to continue to beat that one up with heavy loads 24/7. But even if it works, we don't know if it works on different hardware. OTOH, it's very easy to identify this particular interface and adapt to it, driver level. The PROM ident-string is very specific. FWIW - it says 'ISA8ETH BB', don't know what the BB is.

—M

ghaerr:
> It was kept for compatibility. I'm more than happy to delete it.

Since the new driver is automatic, go ahead.

> [mtu] The other alternative, and always safe, is to add an IOCTL to do it.

Agreed. However, before coding the IOCTLs, lets take one step at a time and discuss exactly how and when ktcp and other programs might access it. This should probably be the case for QEMU detection as well. For now, coding the in-use MTU and QEMU status into global C variables in the driver will be good enough, the next step being putting those and the other status variables into a single struct, so that a single IOCTL will be able to grab them all, for outside uses.

> [qemu] We can have an IOCTL for that too, or (less reliable) use the HW address which is always the same in QEMU.

Does your driver require knowledge about running on QEMU and act differently? It does, since the QEMU 8-bit NIC is buggy, correct? Thus the QEMU global is a NIC mandatory override, just like MTU, right?

> I'll update the PR with suggestions for the IOCTLs mentioned here.

I'm thinking we should add a single IOCTL, after you've moved all the variables programs or you may want to see into the single struct, and use IOCTL_ETH_GETSTAT to read it. Lets add that code in a separate PR.

> It's all in the C code, so we're fine. I'll add that to the updated PR.

I suggest adding this PR almost as-is with few changes, and another one with more changes, for better stability, since you've spent tons of time testing this one.

> We really need more people beating this stuff up on real hardware.

Agreed. I need to get a NIC running on my desktop 386/20, or perhaps find another NIC for it that the motherboard will accept.

> I'm really excited about the prospect of being using the full 16k buffer in the 8bit NIC. I'm going to continue to beat that one up with heavy loads 24/7.

I thought your report says that the NIC hardware only uses 4k in 8-bit mode?  I'll leave it to you, and reread your report!

Mellvik:
Thank you, @ghaerr
>  However, before coding the IOCTLs, lets take one step at a time and discuss exactly how and when ktcp and other programs might access it. This should probably be the case for QEMU detection as well. For now, coding the in-use MTU and QEMU status into global C variables in the driver will be good enough, the next step being putting those and the other status variables into a single struct, so that a single IOCTL will be able to grab them all, for outside uses.
> 
Agreed.
> [qemu] We can have an IOCTL for that too, or (less reliable) use the HW address which is always the same in QEMU.
> 
> Does your driver require knowledge about running on QEMU and act differently? It does, since the QEMU 8-bit NIC is buggy, correct? Thus the QEMU global is a NIC mandatory override, just like MTU, right?
> 
Well...  what the driver does is to detect a certain behavior - which QEMU has. And which other systems/interfaces may have. For example, the rtl8019 emits that same behaviour if the prom is read in byte mode. We're not doing that so we're fine. The thing is - if some system has that same behaviour, it's non-critical, it will work.
Come to think of it, it's probably smart to test for the QEMU default MAC address before concluding that it's QEMU. 
> I'll update the PR with suggestions for the IOCTLs mentioned here.
> 
> I'm thinking we should add a single IOCTL, after you've moved all the variables programs or you may want to see into the single struct, and use IOCTL_ETH_GETSTAT to read it. Lets add that code in a separate PR.
> 
Good idea!
> It's all in the C code, so we're fine. I'll add that to the updated PR.
> 
> I suggest adding this PR almost as-is with few changes, and another one with more changes, for better stability, since you've spent tons of time testing this one.
> 
Ok, great. I'll fix the minor things in the morning and submit. We'll take it from there.
> I'm really excited about the prospect of being using the full 16k buffer in the 8bit NIC. I'm going to continue to beat that one up with heavy loads 24/7.
> 
> I thought your report says that the NIC hardware only uses 4k in 8-bit mode? I'll leave it to you, and reread your report!
> 
As I was climbing out of the rabbit CAVE, I discovered a possible silver lining ...
Yes, it's in the document, towards the bottom.

-M
>

cocus:
> @cocus, it would be great if you could take this one for a test drive on 8bit hw!

Oh yeah, I'll definitely try this tomorrow!

Mellvik:
@ghaerr, I figure we'll phase in the use of more of the `if_status` flags as we go.

M

cocus:
> @cocus, it would be great if you could take this one for a test drive on 8bit hw!
> 
> --M

Hey I've tried it, and seems to work.
```
$ telnet 192.168.137.69
Trying 192.168.137.69...
Connected to 192.168.137.69.
Escape character is '^]'.
# uname -a
ELKS elks 0.5.0 commit 78307bae 09 May 2022 22:50:22 -0300 8018x
#
```

I did set the 8 bit eth access flag on the config, didn't try the 8/16 bit auto config.

EDIT: There's still one thing that doesn't work. If I let the ethernet cable plugged in BEFORE running ktcp, then telnetd freezes. I need to keep my ethernet cable disconnected until I run ktcp.

ghaerr:
> There's still one thing that doesn't work. If I keep the ethernet cable plugged in BEFORE running ktcp, then telnetd freezes. I need to keep my ethernet cable disconnected until I run ktcp.

@cocus, did that used to work before this driver rewrite?

Mellvik:
@cocus - about the cable unplugged problem, could you post the ethernet boot messages? If they're different between the two cases, can you post both?

--M

Mellvik:
@ghaerr - the latest update to this PR includes  'safe' QEMU detection  and moves the MAC address field to the `netif_stat` struct. I think this is a good time to merge, so we can get on with the `ioctl`. And attack the cable problem reported by @cocus. (BTW - I just noticed there is a compilation warning, will fix that in the next round).

I found that the `memcmp` function while defined in strings.h is not in the library, so I added that to `ne2k.c` for now. I'm assuming it's a piece of cake to add that to libc.

I have encouraging news about the 16k hack for the 8bit controller. With the current recovery scheme, I'm unable to get it to hang or otherwise fail. I've poured 100s of megabytes onto it from Linux, while running pings and other activities concurrently, and it just keeps going. I can get ktcp to hang if the other end is also ELKS, but that's a different ballgame which I'm now getting back into. Interestingly though, even when `ktcp` is 'hung', it continues to respond to ICMPs, which is a good sign (and shows the driver is still running).
Transfer speed into the 8bit interface using ftp and the full 16k buffer is around 24k/s, depending on how much ICMP there is on the side. Pretty good.

When using this setup, we get quite a few TCP level checksum errors (see below). `ktcp` recovers nicely from those, and I think a little more tweaking in the driver can stop those errors from occurring. It's on my list.

The way this looks right now, it makes sense to either make the assumption that this hack will work for all 8bit interfaces and make it the default, or to recognize this particular 8bit interface and enable 16k specifically. I'm vouching for the former - because of the simplicity and because when running my 16bit interfaces in forced 8bit mode w 16k buffer, I get the same behaviour as the 8bit interface, just fewer errors (mostly checksum errors). 

What follows is typical (and partial) console output when running big file transfers into the 8bit interface in 16k mode (the number of messages is about the same when using the 4k buffer, but they are different, mostly overruns and receive errors. And the speed falls to about 40% of what we have here.

```
elks16# eth: oversized packet (27b8 -25877), clearing buffer
ktcp: eth_process error -1 (errno 5), discarding packet
tcp: BAD CHECKSUM (0x6762) len 1480
tcp retrans: seq 269+53 size 53 rcvwnd 29200 unack 269 rto 48 rtt 12 (RETRY 1 cnt 1 mem 73)
eth: oversized packet (27b8 -25877), clearing buffer
ktcp: eth_process error -1 (errno 5), discarding packet
tcp: BAD CHECKSUM (0x5a6) len 1480
tcp: BAD CHECKSUM (0x3196) len 1480
eth: oversized packet (27b8 -25877), clearing buffer
ktcp: eth_process error -1 (errno 5), discarding packet
eth: oversized packet (27b8 -25877), clearing buffer
ktcp: eth_process error -1 (errno 5), discarding packet
tcp: BAD CHECKSUM (0x447f) len 1480
tcp: BAD CHECKSUM (0x933) len 1480
eth: oversized packet (27b8 -25877), clearing buffer
ktcp: eth_process error -1 (errno 5), discarding packet
tcp: BAD CHECKSUM (0xf256) len 1480
tcp: BAD CHECKSUM (0x933) len 1480
tcp: BAD CHECKSUM (0x933) len 1480
tcp: BAD CHECKSUM (0x933) len 1480
tcp: BAD CHECKSUM (0x5f30) len 1480
tcp: BAD CHECKSUM (0x693) len 1480
tcp: BAD CHECKSUM (0x3c71) len 1480
tcp: BAD CHECKSUM (0x6899) len 1480
tcp: BAD CHECKSUM (0x20b1) len 1480
tcp: BAD CHECKSUM (0x1d8e) len 1480
tcp: BAD CHECKSUM (0x5ba1) len 1480
eth: RX-error, status 0x02
eth: RX-error, status 0x02
eth: RX-error, status 0x02
tcp: BAD CHECKSUM (0x93b3) len 1480
tcp: BAD CHECKSUM (0xe570) len 1480
eth: oversized packet (27b8 -25877), clearing buffer
ktcp: eth_process error -1 (errno 5), discarding packet
tcp: BAD CHECKSUM (0x958e) len 1480
tcp: BAD CHECKSUM (0x3f95) len 1480
eth: oversized packet (27b8 -25877), clearing buffer
ktcp: eth_process error -1 (errno 5), discarding packet
```

ghaerr:
Hello @Mellvik,

> I found that the memcmp function while defined in strings.h is not in the library, so I added that to ne2k.c for now. I'm assuming it's a piece of cake to add that to libc.

Note that the kernel doesn't use any libc functions (and doesn't have a strings.h header). The kernel has its own library functions which live in elks/lib and elks/arch/i86/lib. I'll add a version of memcmp to the kernel library after this is committed.

I'm still a bit confused about the 8-bit mode(s) you've found in testing, can you clarify them so I can keep up with what you're learning? I have the following questions:
- In the normal (non-16k hack mode) 8-bit mode, does the driver or ktcp get any checksum or driver errors?
- Are you saying that using 16k hack, 8-bit mode runs 40% faster and gets checksum/driver errors but this is worth it because ktcp recovers?
- Is the 16k hack 8-bit mode turned on only for when using `netport=0xXXX3` set in /bootopts? Or does it also require a certain NIC signature somehow?
- Does the driver automatically set MTU based on NIC requirements? What about ktcp, does it require the mtu= change in /etc/net.cfg to be set manually?

With regards to automatic hack mode, I would recommend that if 8-bit normal mode works with no errors, that becomes the default, so that users don't see error displays. If they read the wiki and care about superior performance, they can turn on hack mode using the netport=. (Only on systems that support /bootopts, but we'll leave that as a prerequisite for now). Is this how the driver works now?

So, to summarize what I think you've done: the new driver automatically determines whether the NIC is 8 or 16 bit, and runs without errors on all NICs out of the box. The MTU may (for the time being) have to be manually adjusted in /etc/net.cfg for best performance. For 8-bit NICs, there are two additional options, both set in netport= : =xxx1 forces 8-bit mode, and =xxx3 forces 16k buffer hack.

Thank you!

Mellvik:
HI @ghaerr - yes, I do realize this may be confusing.

> 10. mai 2022 kl. 17:15 skrev Gregory Haerr ***@***.***>:
> 
> 
> 
> I found that the memcmp function while defined in strings.h is not in the library, so I added that to ne2k.c for now. I'm assuming it's a piece of cake to add that to libc.
> 
> Note that the kernel doesn't use any libc functions (and doesn't have a strings.h header). The kernel has its own library functions which live in elks/lib and elks/arch/i86/lib. I'll add a version of memcmp to the kernel library after this is committed.
> 
Thank you. And yes, I forgot about libc not being used by the kernel. strings.h was a misspelling - string.h is correct, and does declare memcmp().
> I'm still a bit confused about the 8-bit mode(s) you've found in testing, can you clarify them so I can keep up with what you're learning? I have the following questions:
> 
> In the normal (non-16k hack mode) 8-bit mode, does the driver or ktcp get any checksum or driver errors?
Yes, it gets both, but different ones. No checksum errors, lots of buffer overflow errors (obviously depending on traffic) and ktcp errors because read sometimes fail (returns -1 from the driver because the packet supposed to be delivered was discarded). There are also read errors and some transmit errors from the driver under heavy load. With MTU=720 the performance is about 40%, some times less, of the 16bit interface. With MTU=1500, file transfers will hang in ktcp because too many packets get lost.
> Are you saying that using 16k hack, 8-bit mode runs 40% faster and gets checksum/driver errors but this is worth it because ktcp recovers?
No, the 40% is the case above. With the 16k hack, the 8bit interface delivers a performance that is just 5% less than the 16bit interface!! And with the latest variant of the recovery code, there are fewer error messages with the hack than without. Not really a good metric though, error messages that are deemed uninteresting (to a 'normal user', that's probably all) can be turned off. And yes, ktcp recovers in both cases, except ELKS-to-ELKS, but that's a different story.
> Is the 16k hack 8-bit mode turned on only for when using netport=0xXXX3 set in /bootopts? Or does it also require a certain NIC signature somehow?
Currently the 16k hack is turned on via bit 1 (counting from 0) only. The bit isn't checked unless the interface is 8bit, automatic or manual configuration. Thus when a 16bit interface is run in 8bit mode, it can be tested with both variants.
> Does the driver automatically set MTU based on NIC requirements? What about ktcp, does it require the mtu= change in /etc/net.cfg to be set manually?
As soon as we enable the IOCTL to get the interface status, ktcp can set the MTU automatically. Now it's manual via net.cfg.
> With regards to automatic hack mode, I would recommend that if 8-bit normal mode works with no errors, that becomes the default, so that users don't see error displays. If they read the wiki and care about superior performance, they can turn on hack mode using the netport=. (Only on systems that support /bootopts, but we'll leave that as a prerequisite for now). Is this how the driver works now?
> 
Yes, but net.cfg still has to be changed manually - and there are lots of error messages regardless - if doing incoming file transfers or - say - an outgoing telnet from ELKS is doing something that generates lots of output.
> So, to summarize what I think you've done: the new driver automatically determines whether the NIC is 8 or 16 bit, and runs without errors on all NICs out of the box. The MTU may (for the time being) have to be manually adjusted in /etc/net.cfg for best performance. For 8-bit NICs, there are two additional options, both set in netport= : =xxx1 forces 8-bit mode, and =xxx3 forces 16k buffer hack.
> 
Except for the error messages this is correct.
And as I mentioned in the 'lab notes', it may be desirable to encapsulate the relevant error messages in the ^P debug scheme to get rid of them. It has been a blessing to have several serial lines active on all 3 machines (and a network terminal server) when working on this.

Bottom line - what seemed like a complete impossibility just a couple of weeks back, now seems like 'the real thing' so to speak. I think 5% off 16bit performance is pretty incredible.


—M

cocus:
> @cocus, did that used to work before this driver rewrite?

I think it did, I'll have to check.


> @cocus - about the cable unplugged problem, could you post the ethernet boot messages? If they're different between the two cases, can you post both?
> 
> --M

They're actually the same. ktcp always works, but the only noticeable difference is telnetd hanging.



And finally, I'm a little bit lost on how the 8/16 bit recognition works. For my particular use case (i.e. the 80188), I'm stuck with an 8 bit ISA port. Should it work as-is or do I have to manually do something to change its mode?

ghaerr:
Thanks for the explanations, it is complicated. Thank you for putting in the time to figure all this out!!

It sounds like the bottom line is that:
- 16 bit mode works great
- our ELKS 8-bit support previously didn't work well at all, and currently works, but will always produce errors, different ones based on the 4k vs 16k buffer size
- there's a good possibility that 8-bit support can run very much faster using a 16k buffer (needs testing on various cards), though possibly with some error messages, and that you're still working on that

Next steps:

I will add memcpy to the kernel libs, and when that's done you can remove your version in a future NIC update PR.

I will put together the basics of getting NIC data to `netstat` via an ioctl that you implement in ne2k.c. The data will be transferred into the same struct in netstat.c. You will then add code to print the information as you desire.

We can discuss the hiding of ktcp errors, etc separately.

Is this accurate? Are you ready for commit now?

Thank you!

ghaerr:
> I'm a little bit lost on how the 8/16 bit recognition works. For my particular use case (i.e. the 80188), I'm stuck with an 8 bit ISA port. Should it work as-is or do I have to manually do something to change its mode?

FWIW, I recommend it be automatic to a working state on 8- or 16- bit NICs, with performance enhancements done separately, if not possible to be completely automatic. The less a user needs to know to get configured and working (which is not the same as highest performance), the better. 

If this is not the way this current PR works, that can be changed in @Mellvik's next PR, since we're about to commit this now.

Mellvik:
Spot on, great summary and good plan, @Ghaerr.

Please go ahead with the commit.

Thank you!

-M

Mellvik:
> I'm a little bit lost on how the 8/16 bit recognition works. For my particular use case (i.e. the 80188), I'm stuck with an 8 bit ISA port. Should it work as-is or do I have to manually do something to change its mode?
> 
> FWIW, I recommend it be automatic to a working state on 8- or 16- bit NICs, with performance enhancements done separately, if not possible to be completely automatic. The less a user needs to know to get configured and working (which is not the same as highest performance), the better.
> 
> If this is not the way this current PR works, that can be changed in @Mellvik's next PR, since we're about to commit this now.
> 
> 
> 
Keep an eye on the boot messages, @cocus, they will tell you 8 or 16 bit. Since it seems to be working, you should be ok. Are you getting the same MAC address as before?
Until the next update, add 3 to the netport address in /bootopts to get best performance. The boot message should then  say 'forced 16k mode'.

-M

ghaerr:
> Until the next update, add 3 to the netport address in /bootopts to get best performance. The boot message should then  say 'forced 16k mode'.

@cocus is running CONFIG_ROMCODE so /bootops is inoperational.

I wonder if it would be worth implementing it for ROMFS, or instead have to modify source code for these kinds of features?
I don't think it a big deal to implement it, and cost would be slightly larger kernel and 512 bytes RAM. [EDIT: actually it could be a big deal to implement, unless @cocus can load segment 0x40 with /bootopts file and then set CONFIG_BOOTOPTS. The problem is in reading the file at boot time from ROMFS, but if it's already in RAM at boot, then easy.]

ghaerr:
Another possibly for @cocus would be to copy /bootopts file at image-making time to somewhere in BIOS memory, and then set OPT_SEG to that segment, then set CONFIG_BOOTOPS. That could work well. This only means much as a simpler way of setting/testing options without recompiling the kernel. The flip side is that BIOS would need to be reflashed as well to change any options, or they would remain set with new kernel.

cocus:
> Another possibly for @cocus would be to copy /bootopts file at image-making time to somewhere in BIOS memory, and then set OPT_SEG to that segment, then set CONFIG_BOOTOPS. That could work well. This only means much as a simpler way of setting/testing options without recompiling the kernel. The flip side is that BIOS would need to be reflashed as well to change any options, or they would remain set with new kernel.

Yes, in theory I could set up that bootops segment with the contents of bootopts. Another way could be using an EEPROM to store it, which would NOT require to reflash the bios in order to do it. For the time being, I can live with the "slower, but quite stable" 8k mode for the 8bit card.

Mellvik:
Ok, the simplest would of course be to have the defaults work out of the box. 
That should be the case now, except for the hack, which I can make the default in the next PR.

-M

ghaerr:
> except for the hack

I'm still bit confused on this - is the "hack" a true hack, meaning not good, or is it correct code that works well on any NE2000 (RealTek?) NIC, and is not a hack? 

Thank you for your mention of the book by Fred Eady, that sounds very interesting. I'm digging up a copy as we speak to read about the RealTek controller in an attempt to follow you. The part about your hack I don't understand is, you had said that Eady mentioned the NIC chip only supports 4k RAM in 8-bit mode: so how can it be made to work? Is that "the hack"? Why don't all software drivers just use your method, does the documented hardware only support the 4k buffer method? Thanks for helping me to understand :)

> which I can make the default in the next PR.

Whichever way you choose, it is important that it must work for all controllers out of the box. Performance is secondary to actually working, although I suppose it comes in a close second if there's a large difference in throughput.

Mellvik:
> ﻿
> except for the hack
> 
> I'm still bit confused on this - is the "hack" a true hack, meaning not good, or is it correct code that works well on any NE2000 (RealTek?) NIC, and is not a hack?
> 
@ghaerr, our definition of hack differs. To me a hack - at the technical level - may well be a positive thing. A smart or creative fix may well qualify as a hack, that eventually turns into a solution.
> Thank you for your mention of the book by Fred Eady, that sounds very interesting. I'm digging up a copy as we speak to read about the RealTek controller in an attempt to follow you. The part about your hack I don't understand is, you had said that Eady mentioned the NIC chip only supports 4k RAM in 8-bit mode: so how can it be made to work? Is that "the hack"? Why don't all software drivers just use your method, does the documented hardware only support the 4k buffer method? Thanks for helping me to understand :)
> 
This was only partly covered in my 'lab notes' and the full explanation is lengthy. The short take is this: The chip supports 4k buffer in 8 bit mode, but it still has 16k - all accessible, most of it unreliable. What the hack does is to detect and handle the unreliablility, which in our case (because 4k works so badly with ELKS) turns out to be a lot better than the alternative.
[BTW I ordered the book on eBay 3 weeks ago, should be arriving soon.]
> which I can make the default in the next PR.
> 
> Whichever way you choose, it is important that it must work for all controllers out of the box. Performance is secondary to actually working, although I suppose it comes in a close second if there's a large difference in throughput.
> 
Agreed. It works out of the box now. There may well be interfaces that don't fit into the autodetection scheme, thus the override - and when they appear, we'll fix them.

--M
>

Mellvik:
Thank you @ cocus!!
Any chance you can do a file transfer into ELKS and record the console messages during the transfer?
And repeat after changing your netport= variable in /bootopts to end with a '3' and reboot?

BTW - curiosity: Your machine has a 8bit ISA bus, right?
And finally, what times do you get if you just ping into ELKS from your linux system?
 
Much appreciated!

--Mellvik

PR #1289, opened 2022-05-10
[basic] Add inp() and out() for both word and byte IO access

cocus:
Adds the INPB, INPW, OUTB and OUTW keywords for input and output
of bytes and words (8 and 16 bits).

### DIFF
--- a/elkscmd/basic/README.md
+++ b/elkscmd/basic/README.md
@@ -63,6 +63,10 @@ COLOR fg,bg (PC-98 only for now)
 PLOT x,y (PC-98 only for now)
 DRAW x,y (PC-98 only for now)
 CIRCLE x,y,r (PC-98 only for now)
+INPB(port) (IO read byte from `port`)
+INPW(port) (IO read word from `port`)
+OUTB port, value (IO write byte `value` from `port`)
+OUTW port, value (IO write word `value` from `port`)
 
 Architecture-specific
 PIN pinNum, value (0 = low, non-zero = high)
@@ -70,7 +74,6 @@ PINMODE pinNum, mode - not implemented
 
 Not yet implemented
 POKE offset,segment,value
-OUT port,value
 RANDOMIZE [nmber]
 ```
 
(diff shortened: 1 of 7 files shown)


cocus:
@ghaerr please check if the additions to the basic-stubs.c are okay, or if I should remove them.

cocus:
Next on the list would be a HEX$() which would be highly useful for the stuff I'm working on.

ghaerr:
This looks great, thanks!

Can push an update to basic/README.md to update the documentation of the two new commands and two new functions? Both IN() and OUT are documented in the Not Yet Implemented sections and need to be separated into the byte and word functions you've added. This will allow us to keep track of our nicely expanding language!

ghaerr:
> check if the additions to the basic-stubs.c are okay, or if I should remove them.

Looks good as you've done it.

cocus:
> This looks great, thanks!
> 
> Can push an update to basic/README.md to update the documentation of the two new commands and two new functions? Both IN() and OUT are documented in the Not Yet Implemented sections and need to be separated into the byte and word functions you've added. This will allow us to keep track of our nicely expanding language!

Right, I forgot about this!

ghaerr:
Also, please check whether the TKN_FMT_POST should be set in the following:
```
    {"INPB", 1|TKN_FMT_POST},
    {"INPW", 1|TKN_FMT_POST},
```
Leaving that in will cause a space to be output after the keyword, which I don't think you want in a function. Type the function into a statement then "LIST" to see.

Thanks!

cocus:
> Leaving that in will cause a space to be output after the keyword, which I don't think you want in a function. Type the function into a statement then "LIST" to see.

Yes, that's true. Should I also remove them from the OUTB/OUTW?

ghaerr:
> Should I also remove them from the OUTB/OUTW?

No, because those aren't functions, they are commands. Also remove the old IN() function from README.md. We're almost there!

cocus:
> > Should I also remove them from the OUTB/OUTW?
> 
> No, because those aren't functions, they are commands. Also remove the old IN() function from README.md. We're almost there!

I didn't see any declaration of the IN() function, only OUT() which I've already modified.

ghaerr:
> I didn't see any declaration of the IN() function

It's the 4th line from the bottom. Don't worry, we can catch that next round.

cocus:
> > I didn't see any declaration of the IN() function
> 
> It's the 4th line from the bottom. Don't worry, we can catch that next round.

woops

PR #1293, opened 2022-05-16
[net, ne2k] Driver adjustments aqnd cleanups

Mellvik:
Update/adjustments to the ne2k driver
- IOCTL to retrieve the status structure from the NIC is now enabled (untested).
- The status struct now includes the rest of the PROM contents (10 bytes), which may be useful for NIC vendor/type identification.
- 16K NIC buffer is now the default for both 8bit and 16bit interfaces, the meaning of the 'force NIC buffer size' config bit in /bootopts has been reversed.
- The 8bit/16k buffer corruption recovery has been slightly adjusted.
- As has been mentioned before, the MTU plays a critical role when using 8bit interfaces and the recommended 4k buffer. Further - also discussed before - the timeouts and other characteristics of `ktcp` have major influence on the resulting transfer speed. Recent adjustments and more testing with varying MTUs reveal that on this particular machine/in this environment, MTU=1000 is ideal. It wasn't a month ago, but a lot has changed since then. With MTU=1000 I get transfer speed (incoming, ftp) close the 16k buffer level, i.e. 24kB/s and only 7 overruns during a 1.4M transfer (MTU=720 -> ca. 10k/s). This is surprising end very encouraging. `/etc/net.cfg` has been updated accordingly. 
- Larger MTUs will 'hang' the connection somehow, while `ktcp` continues to work fine. I put 'hang' in quotes for a reason: The transfer stops, then continues after 120 seconds, exchanges a few packets, stops for another 120 seconds, etc. There may be external factors contributing to this. To be investigated.
- Misc cleanups and comment fixes.

@ghaerr - I guess the next step is to pull the data from the driver into ktcp, but there is no rush in this. It would be convenient to have the ktcp updates I have stashed, committed before continuing.

-M

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -10,8 +10,8 @@
 // Updated by Santiago Hormazabal on Dec 2021:
 //  . 8 bit access if CONFIG_ETH_BYTE_ACCESS is set, using a ne1k patch from
 //    NCommander.
-// apr-2022 (HS) : Support auto detection of 8bit mode, set 4k buffer size when in 8bit mode
-//	     assuming RTL8019 buffer restrictions.
+// apr-2022 (HS) : Support auto detection of 8bit mode, set 4k or 16k buffer size when in 8bit mode
+//	     depending on configuration, 16k is the default, 4k is 'per spec'.
 //	     Rewrote overflow handler, added direct access to many registers from C code
 //	     Cleaned up initialization code, optimized 8bit I/O
 //
@@ -212,7 +212,7 @@ dma_write:
 	mov	current,%bx		// setup for far memory xfer
 	mov	TASK_USER_DS(%bx),%ds
 	cld
-	test	%ax,%ax		// checking _ne2k_is_8bit
+	test	$1,%ax		// checking _ne2k_is_8bit
 	jz	1f
 
 	// Byte loop
@@ -320,16 +320,18 @@ dma_read:
 
 	mov     %ds,%bx
 	mov     %bx,%es
+	movw	_ne2k_is_8bit,%dx
 	pop	%ax
 	cmp	$0,%al		// Use local buffer if zero
 	jz	buf_local
 	mov	current,%bx	// Normal: read directly into the (far) buffer
 	mov	TASK_USER_DS(%bx),%es
 
-	pop	%bx
-	push	%bx
+	//pop	%bx
+	//push	%bx
 
 buf_local:
+	mov	%dx,%bx		// _ne2k_is_8bit
 	mov	net_port,%dx	// command register
 	mov	$0x0a,%al	// set RD0 & STA
 	out     %al,%dx		// start DMA read
@@ -337,7 +339,8 @@ buf_local:
 	mov	net_port,%dx
 	add	$io_ne2k_data_io,%dx
 	cld			// clear direction flag
-	cmpw	$0,_ne2k_is_8bit
+	//cmpw	$0,_ne2k_is_8bit
+	test	$1,%bx		// is 8bit?
 	jz	1f
 
 // Byte transfer
@@ -524,11 +527,8 @@ ne2k_pack_get:
 	// Got the entire block (256b) instead of the 4 first bytes. Which was great 
 	// for small packets (telnet, command packets etc.): One read instead of 2.
 	//
-	// Removed when changing to read directly into the process was introduced,
+	// Removed when changing to read directly into process address space was introduced,
 	// there is no longer anywhere to put the first 4 bytes.
-	// It may be an idea to reintroduce a 256b buffer to take advantage 
-	// of this 'optimization' and at the same time help strangled 8bit 
-	// interfaces.
 
 	//sub	$252,%cx	// Got entire packet?
 	//jle	npg_cont
@@ -765,9 +765,10 @@ ne2k_init:
 	mov     $2,%al  // 2 for loopback
 	out     %al,%dx
 
-	// set RX ring limits - all 16KB on-chip memory
-	// except one TX frame at beginning (6 x 256B)
-	// (unless it's an 8 bit card, then only 4KB for send&receive)
+	// set RX ring limits - 16KB on-chip memory
+	// except one TX frame at beginning (6 x 256B).
+	// If bit 1 in _ne2k_is_8bit is set, use only 4k
+	// buffer.
 
 	mov	net_port,%dx
 	add	$io_ne2k_rx_first,%dx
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @Mellvik,

Thanks for the cleanups. 

A couple comments: the single `memcpy` from cprom to mac_addr for 16 bytes is confusing, and took me a while to understand what is happening, considering that sizeof(mac_addr) is 6. This depends on if_id[10] array being adjacent in the netif_stat struct, and nothing indicates these two struct members must be packed, and adjacent. I suggest using two memcpy statements, or a single mac_addr[16], since the C/ASM code doesn't care about sizeof(mac_addr).

I had fine-tuned /etc/net.cfg's size to be <= 1024 bytes, so as to not require a second sector read on network startup, which takes time on floppies. The extra comments look like they will exceed that limit, but I can't tell.

The above changes can wait, and I can commit this now. However - it seems that, for some reason, this entire PR's code is duplicated in your next PR #1294. I'm wondering if I should commit this first, then #1294, or try committing just #1294. I don't know why git is interpreting this way, but I suspect trying to commit both PR's will end up with conflicts. Let me know which way you'd prefer to go.

Thank you!

Mellvik:
> 
Thank you @ghaerr -
> A couple comments: the single memcpy from cprom to mac_addr for 16 bytes is confusing, and took me a while to understand what is happening, considering that sizeof(mac_addr) is 6. This depends on if_id[10] array being adjacent in the netif_stat struct, and nothing indicates these two struct members must be packed, and adjacent. I suggest using two memcpy statements, or a single mac_addr[16], since the C/ASM code doesn't care about sizeof(mac_addr).
> 
Yes, I can see that this is confusing - it was a late minute addition. I'm having second thoughts about the actual usefulness of including the last 10 bytes, so I'm inclined to remove it.
> I had fine-tuned /etc/net.cfg's size to be <= 1024 bytes, so as to not require a second sector read on network startup, which takes time on floppies. The extra comments look like they will exceed that limit, but I can't tell.
> 
Interesting, and good point, I'll check and make adjustments if required.
> The above changes can wait, and I can commit this now. However - it seems that, for some reason, this entire PR's code is duplicated in your next PR #1294 <https://github.com/jbruchon/elks/pull/1294>.
> 
Yeah, this is weird. I pushed the network stuff first, then changed to a different branch and pushed mount. On there PR page they look entirely separate, then in i different view they look like on top of each other.
> I'm wondering if I should commit this first, then #1294 <https://github.com/jbruchon/elks/pull/1294>, or try committing just #1294 <https://github.com/jbruchon/elks/pull/1294>. I don't know why git is interpreting this way, but I suspect trying to commit both PR's will end up with conflicts. Let me know which way you'd prefer to go.
> 
> 
I suggest merging the net-PR first, the see what happens when you merge the 2nd. If git barfs, we'll just delete it and create it again.

Thanks,
—M

ghaerr:
> I pushed the network stuff first, then changed to a different branch and pushed mount.

I'm wondering whether your second branch was from master? Or perhaps created from your ftp branch, which then caused #1294 to be dependent on it? Good news all is committed.

> I'm having second thoughts about the actual usefulness of including the last 10 bytes, so I'm inclined to remove it.

I would agree. In general, I have been working to reduce clutter in ELKS, and that sometimes means removing stuff was thought to be useful, but never actually used. That also goes towards the ETH_IOCTL entries which aren't actually being used, as we still have considerations about kernel code size (although running out of kernel data has now taken the lead in that issue). The addition of (even temporary) '#if 0' shows the code isn't actually used anywheres, and can be removed at the time the other code that uses them is committed (IOCTL_ETH_OFWSKIP_SET and IOCTL_ETH_ADDR_SET would be a good example of this).

> I guess the next step is to pull the data from the driver into ktcp

Perhaps, I was thinking you wanted the netif_stat information read into `netstat` first. We still need to discuss exactly how ktcp should use its calculated MTU vs the -m MTU passed to it, and when each might override the other.


> but there is no rush in this. It would be convenient to have the ktcp updates I have stashed, committed before continuing.

Ok. You're welcome to code up the netstat display of collected information from the driver as well. Keep in mind the ioctl will fail on non-NE2 NIC drivers (WD 8000).

Mellvik:
> 
> I pushed the network stuff first, then changed to a different branch and pushed mount.
> 
> I'm wondering whether your second branch was from master? Or perhaps created from your ftp branch, which then caused #1294 to be dependent on it? Good news all is committed.
> 
That's got to be it, thanks! It makes sense now! Learning (and unfortunately forgetting) every day.
> I'm having second thoughts about the actual usefulness of including the last 10 bytes, so I'm inclined to remove it.
> 
> I would agree. In general, I have been working to reduce clutter in ELKS, and that sometimes means removing stuff was thought to be useful, but never actually used. That also goes towards the ETH_IOCTL entries which aren't actually being used, as we still have considerations about kernel code size (although running out of kernel data has now taken the lead in that issue). The addition of (even temporary) '#if 0' shows the code isn't actually used anywheres, and can be removed at the time the other code that uses them is committed (IOCTL_ETH_OFWSKIP_SET and IOCTL_ETH_ADDR_SET would be a good example of this).
> 
Good points. I'll take another round of cleaning. Of course there are cases where the #if 0 encapsulates useful experiences to be preserved - or more or less serious intentions for improvement, which almost never happens. That said I'd never expected to dive back into the ne2k driver like this. I'm glad I did. If the 3C509 driver ever gets completed, this (sometimes) excruciating Experience may become useful again.
> I guess the next step is to pull the data from the driver into ktcp
> 
> Perhaps, I was thinking you wanted the netif_stat information read into netstat first. We still need to discuss exactly how ktcp should use its calculated MTU vs the -m MTU passed to it, and when each might override the other.
> 
Admittedly I haven't looked at netstat yet. I mentioned ktcp because I've been assuming we have to get the data through it: we have to open the device to get to the ioctl and ktcp has opened it already. Or we'd have to allow for multiple opens. Did I miss something?
> but there is no rush in this. It would be convenient to have the ktcp updates I have stashed, committed before continuing.
> 
> Ok. You're welcome to code up the netstat display of collected information from the driver as well. Keep in mind the ioctl will fail on non-NE2 NIC drivers (WD 8000).
> 
Good point. We should add an ioctl to it returning just an error code.

Thank you! 

-M

ghaerr:
> Of course there are cases where the #if 0 encapsulates useful experiences to be preserved - or more or less serious intentions for improvement, which almost never happens.

LOL, exactly. Thus, leaving code in the source file but turned off with #if 0 is the way to go in some complicated source files, at least during heavy development. Then someone removes it years later!

> I mentioned ktcp because I've been assuming we have to get the data through it: we have to open the device to get to the ioctl and ktcp has opened it already. Or we'd have to allow for multiple opens

Oh geez, I hadn't thought of the problem of multiple opens. Direct `netstat` would definitely be the method of choice, since `ktcp` is running as a background daemon and is already big. We probably don't want to package the netif_stat struct through a network connection to `netstat`, but may have to. Perhaps best to leave all the ioctl ifdefed out until we figure out the actual method of data transfer. I sometimes use "#if LATER" to indicate that.

> We should add an ioctl to it returning just an error code.

A missing driver ioctl entry point should (already) be properly handled in the kernel and return EINVAL; that way we reduce the "growth bloat" of having to include more code in every other network driver for something that isn't actually even implemented yet!

PR #1294, opened 2022-05-16
Mount: Fix display of mounted filesystems after umount

Mellvik:
`mount` w/o arguments will display mounted filesystems. It does so by running through an array of mounts, and stops when an empty entry is encountered. So if a filesystem is unmounted (and it's not the most recently mounted), thus creating an empty entry in the table, the mounted fs display will stop right there, and not display mounts further down the table.

This simple fix simply runs through the entire table (currently 5 entries), displaying all valid entries.

@ghaerr - I'm hitting the ceiling with only 5 mounts available, it would be useful to add one more (4 partitions + 2 floppies).

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -10,8 +10,8 @@
 // Updated by Santiago Hormazabal on Dec 2021:
 //  . 8 bit access if CONFIG_ETH_BYTE_ACCESS is set, using a ne1k patch from
 //    NCommander.
-// apr-2022 (HS) : Support auto detection of 8bit mode, set 4k buffer size when in 8bit mode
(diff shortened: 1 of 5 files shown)


ghaerr:
> It does so by running through an array of mounts, and stops when an empty entry is encountered.

It seems you've encountered a bug in my recent implementation of the `ustatfs` system call, which was supposed to have reliably allowed for counting the number of mounted filesystems. But it looks like your solution will be OK, I'll look more into it after this commit.

>  I'm hitting the ceiling with only 5 mounts available, it would be useful to add one more (4 partitions + 2 floppies).

OK, I will increase to 6 after checking ustatfs kernel code in seperate PR.

NOTE: as discussed in #1293, the NIC updates are somehow showing up in this PR. Not sure which PR to commit first.

PR #1296, opened 2022-05-23
[net, ne2k] More cleanups, some bugfixes

Mellvik:
More cleanups and some bugfixes in the `ne2k` driver:
- The `ioctl` - which is already used by `ktcp` to get the MAC address - is ready to deliver status info back to `ktcp`or whatever tool needs it, when we decide on a mechanism to bypass the single open challenge.
- A recent addition to the `netif` struct was removed.
- Not implemented, but it may be an idea to use the driver status word to suppress error messages for 8bit interfaces. That would require a new (and very) simple `ioctl` entry.
- Activating receive error interrupts is now 'dynamic' in order to get around a QEMU bug: RXE intr is only enabled if the interface is 8bit, thus avoiding QEMU automatically. It's not useful (hardly ever happens) with 16 bit interfaces.
- A couple of important bugs were found and fixed:
    - A typo in the overflow handler caused some of the reset code to malfunction - which in very rare cases caused weird behaviour (although recoverable) from the overflow. 
    - A bug in the packet transmission code (packet write) made the transmit complete status check nonfunctional. Fixing it revealed another bug in the same part of the code. Presumably (no proof) the bug may have caused occasional (and silent) transmit packet failures.
- RDC (`Remote DMA Complete`) interrupts were somewhat haphazardly cleared which in odd cases could cause a delay in the sending of packets. Minor adjustments eliminate this risk and also qualify the use of CLI/STI pairs in the `dma_read` and `dma_write` routines.
- Lots of comment updates.

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -79,7 +79,7 @@ io_ne2k_reset      = 0x1F	// Really a port, not a register, force HW reset of th
 tx_first           = 0x40
 rx_first           = 0x46
 rx_last_16	   = 0x80
-rx_last_8	   = 0x50	// Use only 4k in 8 bit mode 
+rx_last_8	   = 0x50	// For 4k in 8 bit mode - per spec. 
 
 //-----------------------------------------------------------------------------
 	.data
@@ -88,7 +88,7 @@ rx_last_8	   = 0x50	// Use only 4k in 8 bit mode
 
 	.global _ne2k_next_pk
 _ne2k_next_pk:
-	.word 0	// being used as byte ...
+	.word 0			// being used as byte ...
 
 	.global _ne2k_has_data
 _ne2k_has_data:
@@ -100,7 +100,6 @@ _ne2k_is_8bit:
 
 _ne2k_rx_last:
 	.byte rx_last_16	// default to the 16 bit value
-
 
 	.text
 
@@ -115,7 +114,7 @@ ne2k_addr_set:
 
 	push    %bp
 	mov     %sp,%bp
-	push    %si  // used by compiler (???)
+	push    %si 
 
 	mov     4(%bp),%si
 
@@ -178,10 +177,10 @@ dma_init:
 	ret
 
 //-----------------------------------------------------------------------------
-// Write block to chip with internal DMA
+// Write data block to NIC with internal DMA
 //-----------------------------------------------------------------------------
 //
-// BX    : chip memory address (to write to)
+// BX    : NIC memory address (to write to)
 // CX    : byte count
 // DS:SI : host memory address (to read from)
 //-------------------------------------
@@ -191,11 +190,10 @@ dma_write:
 	push    %cx
 	push	%bx	// TODO check if this is required (2)
 	push	%ds
-	push    %si
 
-	cli		// Experimental
 	inc     %cx     // make byte count even
 	and     $0xfffe,%cx
+	cli		// Mandatory
 	call    dma_init
 
 	// start DMA write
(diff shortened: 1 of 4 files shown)


toncho11:
So this new stuff improves everything? I mean not only ne2k driver? 
And for the smc driver one still needs to recompile the kernel to put the parameters?

Mellvik:
Hi @toncho11,

this is all about the ne2k driver.

And yes, the WD driver needs to be 'manually' configured like before. No doubt the the WD driver would benefit greatly from many of the same improvements. Unfortunately I don't have a WD interface, so I'll have to leave that to someone else …


—M

ghaerr:
Hello @Mellvik,

Thanks for the continued enhancements and explanations!

All looks fine, although I did (finally) just now notice that the driver is not using the `verified_memcpy_{to, from}fs()` routines, instead justing the (unverified, direct) memory copy `memcpy_{to, from}fs()` functions to copy data to/from the user application and the kernel in the ioctl entry point(s). The (required) verify_xxx routines perform the additional kernel check that the user program in fact passed (in or out) a large enough buffer for the data transfer, or a kernel -EFAULT errno is returned instead.

I'm going to commit this now. You can see how the "verified" calls work by looking at the serial driver ioctl entry points (in serial-8250.c) and then submit a change with this checking performed in the future.

In responding to your comment about getting a v0.6.0 out soon, to fix the v0.5.0 XMS boot problems, I agree. Are you thinking the NE2K driver will continue to be modified for the next several weeks? Perhaps we should wait until after you're satisfied with the operation, and then produce another version. I would also like to wait until @tyama501 finishes up his latest round of PC-98 updates for enhanced HDD support.

Thank you!

ghaerr:
> And for the smc driver one still needs to recompile the kernel to put the parameters?

I can easily add code to update the WD driver to accept `netport=` and `netirq=` from /bootopts, but we need someone with a WD card to test.

Mellvik:
IIRC the wd driver also needs a new boot parameter  - the shared memory address.
I'm on the lookout for a card to test.

-M

toncho11:
@Mellvik I can test these 3 parameters. I have a WD/SMC card. Please add them for 0.6 release and I will test them when I have the time. If you can provide me a 1440 SMC enabled elks image with the 3 parameters, I will be able to test faster.

Mellvik:
Great @toncho11!

@ghaerr,
I can take this one, but I suspect you will have a faster turnaround time.

What say you? :-)


—M

ghaerr:
> What say you? :-)

Ok, no problem.

@toncho11, what are the required settings for your WD card? Also, have you successfully ran ELKS networking with the WD card before?

toncho11:
Yes. I have used it before with my SMC card.
IRQ = 3, IO base memory address = 280, RAM based address = 0xd000
So you are going to put these in /bootopts ?
I need 1440 image.

ghaerr:
> IRQ = 3, IO base memory address = 280, RAM based address = 0xd000
So you are going to put these in /bootopts ?

Yes, they are set in /bootopts in PR #1300, please test when you have time.

cocus:
Hi @Mellvik,
I've flashed my board with the latest sources, and it's detecting my 8 bit isa card as a 16 bit one, and not working at all:
```
eth: NE2K (16 bit) at 0x300, irq 12, MAC a0:a0:a0:a0:a0:a0
```
Is there a way to force this to 8 bits?

cocus:
Should I modify the port to be 0x301 for instance? I don't have /bootopts, but I can (in theory) do this from ports.h. Only problem is that this file is shared and we haven't made a platform-specific one.

cocus:
Ok, no, I take it back. Seems like... somehow, something broke on my design and I need to have my "backup" way of addressing the network card because of this. Not sure why... I've tried to set the address to 0x301 even with my bootloader not setting this, and it's still failing. However, with the fallback mechanism, it works... I think I'll have to debug my design. In any case, your changes are sound, and they do work perfectly. Sorry :)

Mellvik:
Thank you @ghaerr,

> All looks fine, although I did (finally) just now notice that the driver is not using the verified_memcpy_{to, from}fs() routines, instead justing the (unverified, direct) memory copy memcpy_{to, from}fs() functions to copy data to/from the user application and the kernel in the ioctl entry point(s). The (required) verify_xxx routines perform the additional kernel check that the user program in fact passed (in or out) a large enough buffer for the data transfer, or a kernel -EFAULT errno is returned instead.
> 
OK; I wasn't aware of this. I guess it has been safe enough so far since the only 'client' has been and still is ktcp. When we figure out a way to open up the IOCTLs to other programs that's admittedly a new scenario. I'll make that change. Also, we talked about getting a kernel memcmp in order to get rid of the one the driver. I'd like to get that fixed too.
> I'm going to commit this now. You can see how the "verified" calls work by looking at the serial driver ioctl entry points (in serial-8250.c) and then submit a change with this checking performed in the future.
> 
Great! Thanks.
> In responding to your comment about getting a v0.6.0 out soon, to fix the v0.5.0 XMS boot problems, I agree. Are you thinking the NE2K driver will continue to be modified for the next several weeks? Perhaps we should wait until after you're satisfied with the operation, and then produce another version. I would also like to wait until @tyama501 <https://github.com/tyama501> finishes up his latest round of PC-98 updates for enhanced HDD support.
> 
I stumbled back into the rabbit hole quite unintentionally this time. Unsatiable curiosity is a dangerous addiction. Now I've put a more solid lid on the hole - supposed to hold for a while. I'd like to get the memxxx stuff fixed while at it though.

Time to get back to the ktcp ELKS-to-ELKS stuff and the ftp updates. Both have been heavily beaten up over these past months. Encouraging indeed.

—M

PR #1298, opened 2022-05-24
[libc] update how makefile parses `$(CC) -print-multi-lib` output

tkchia:
Commit https://github.com/tkchia/gcc-ia16/tree/20220524 of `gcc-ia16` maps the `-melks-libc` option internally to `-mr=elks`.  Although `-melks-libc` is still supported as an option, the output of `ia16-elf-gcc -print-multi-lib` now says `-mr=elks`, and this will confuse the ELKS libc makefile.


ghaerr:
Thank you!

PR #1300, opened 2022-05-24
[net] Add netram= to /bootopts, configure WD8003 from /bootopts

ghaerr:
Requested in https://github.com/jbruchon/elks/pull/1296#issuecomment-1136204442.

@toncho11,

I've tested the /bootopts parameters, but not actually tested on a WD card. I've attached an 1440k image with your settings in /bootopts (which you can also edit with vi, then sync to change), but these should work for you at boot.
[fd1440.img.zip](https://github.com/jbruchon/elks/files/8766582/fd1440.img.zip)

### DIFF
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -19,11 +19,14 @@
 #include <linuxmt/mm.h>
 #include <linuxmt/debug.h>
 
+/* runtime configuration set in /bootopts or defaults in ports.h */
+int net_irq = WD_IRQ;	        /* default IRQ, set via netirq= */
+int net_port = WD_PORT;         /* default IO port, set via netport= */
+unsigned int net_ram = WD_RAM;  /* default shared memory address, set via netram= */
+
 /* I/O delay settings */
 #define INB	inb	/* use inb_p for 1us delay */
 #define OUTB	outb	/* use outb_p for 1us delay */
-
-#define WD_SHMEMSEG	0xce00U
 
 #define WD_STAT_RX	0x0001U	/* packet received */
 #define WD_STAT_TX	0x0002U	/* packet sent */
@@ -45,7 +48,7 @@
 #define WD_MEMENB	0x40U	/* Enable the shared memory */
 #define WD_IO_EXTENT	32U
 #define WD_8390_OFFSET	16U
-#define WD_8390_PORT	(WD_PORT + WD_8390_OFFSET)
+#define WD_8390_PORT	(net_port + WD_8390_OFFSET)
 
 #define E8390_RXCONFIG	0x04U	/* EN0_RXCR: broadcasts, no multicast,errors */
 #define E8390_RXOFF	0x20U	/* EN0_RXCR: Accept no packets */
@@ -126,8 +129,6 @@ typedef struct {
 	unsigned short count;	/* header + packet length in bytes */
 } __attribute__((packed)) e8390_pkt_hdr;
 
-int net_irq = WD_IRQ;		/* default IRQ, changed by netirq= in /bootopts */
-
 static struct wait_queue rxwait;
 static struct wait_queue txwait;
 
@@ -148,7 +149,7 @@ static void wd_get_hw_addr(word_t * data)
 	unsigned u;
 
 	for (u = 0U; u < 6U; u++)
-		data[u] = INB(WD_PORT + 8U + u);
+		data[u] = INB(net_port + 8U + u);
 }
 
 /*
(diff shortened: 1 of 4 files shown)


toncho11:
So this was a terrible experience ... 
There was an IRQ conflict (IRQ 3) with my floppy controller which is also a serial card. 
Changing the jumpers of the serial card did not help ... I kept getting error -16 from the network card.
I tried changing the jumpers of the card, but then it is impossible to update the /bootopts with kilo. I mean what kind of idiot will write a code that re-writes the entire screen every time you move the cursor. It takes 10 minutes to edit a file. 
After that I forgot that I have to sync the filesystem on a floppy ... so I had do it again ... this is madness !!!
I was testing on 8088 which is very slow, but this is what I had available.
IMG_20220529_195233

I changed the card to irq 5, 300, 0xCA000 and updated the bootopts. It did not work.

ghaerr:
Hello @toncho11,

Wow, I can see you are very frustrated!!

> There was an IRQ conflict (IRQ 3) with my floppy controller which is also a serial card.
Changing the jumpers of the serial card did not help ... I kept getting error -16 from the network card.
I changed the card to irq 5, 300, 0xCA000 and updated the bootopts. It did not work.

Unfortunately, setting up a NIC on an IBM PC can be a nightmare... under any operating system. I still don't actually have a working NIC on my desktop 386 for ELKS, but I think that is because of hardware incompatibilities, not ELKS.

> I changed the card to irq 5, 300, 0xCA000 and updated the bootopts. It did not work.

I thought you said you'd had networking previously working on ELKS? That is why I prebuilt a /bootopts so that you wouldn't have to go through all this, I am sorry.

The -16 return on the above screenshot is -EBUSY, which is likely because IRQ 3 was already in use for the serial port.

When you changed the settings to 5,300,0XCA000, what did the boot screen say next to "eth:...". (NOTE: the 0xCA000 above is incorrect - it must be set to 0xCA00, as 0xCA000 is outside the 1M address space. Not sure if this was a typo).


>  it is impossible to update the /bootopts with kilo. I mean what kind of idiot will write a code that re-writes the entire screen every time you move the cursor. It takes 10 minutes to edit a file.

I saw that when I ported kilo to ELKS originally... I think the kilo author saw it as an exercise in how one might write a full-screen editor in as few lines as possible... which trades off usability on old machines completely for size of executable, unfortunately. I tried porting `nano`, a known supposedly small screen editor from Linux/UNIX to ELKS, and it turned out so big we couldn't even edit a file larger than 1K bytes or so. So we don't really have a good WYSIWYG editor on ELKS, it seems.

> After that I forgot that I have to sync the filesystem on a floppy ... so I had do it again ... this is madness !!!

Yeah, that's Linux/UNIX for you. We added the capability to auto-sync every 30 seconds, but that's off by default. We could add auto-sync into every editor after writing, but that still doesn't solve one of UNIX's big filesystem usability problems, compared to MSDOS.

> I was testing on 8088 which is very slow, but this is what I had available.

I'm aware - I think your system is one of the oldest 8088-based systems we've ever had ELKS tested on, with a very early ROM BIOS. Thank you for your testing. Unfortunately, I don't know why networking didn't work, only a few constants were changed to variables in the driver, but it really needs a test from a known previously-working WD8000 machine in order to determine whether this is a new driver issue or a hardware setup issue.

Thank you!

toncho11:
I usually test on my Amstrad 1640, but I thought that it will also work on the 8088. My Amstrad needs soldering two cables on the power supply. On the card it is written "0xD0000" and "0xCA000", but I suppose one needs to remove the last "0" in both cases. 

For the Amstrad 8086 (8Mhz) (where the network card works OK) I will need the same image, but for 360KB. I could test later. Same parameters as before: IRQ 3, 280, 0xd000

Please make the image with FAT, so that I can edit it on Windows if needed. 

Thank you!

ghaerr:
> I will need the same image, but for 360KB. I could test later. Same parameters as before: IRQ 3, 280, 0xd000

Unfortunately, I am not easily able to make a 360k floppy with networking, as the base commands + networking are larger than 360k! It looks like this will need to be manually accomplished. I will look in to what might be done to enable creating a super-small 360k network-capable boot floppy.

toncho11:
I managed to put together a 386. Now it kind of works on IRQ 5. No error messages and reports the MAC address.

Still when I use ktcp with some IP, gateway and mask I am unable to make a network connection to ELKS after that.
When I do ping to ELKS (the 386) it says "host unreachable". When I type arp in ELKS, it says it knows my laptop's IP and MAC address.

ghaerr:
> I managed to put together a 386. Now it kind of works on IRQ 5. No error messages and reports the MAC address.

Oh, thanks! So this should mean that this PR (adding netirq=, netbase= and netram= entries to /bootopts for the WD8000 driver) is working, as you are not using the kernel-compiled values but instead values in /bootopts, correct?

> Still when I use ktcp with some IP, gateway and mask I am unable to make a network connection to ELKS after that.
When I do ping to ELKS (the 386) it says "host unreachable". When I type arp in ELKS, it says it knows my laptop's IP and MAC address.

We may want to transfer this problem to a seperate issue, where the exact specifics of the IP/gateway/mask used are given and we can get some more eyes looking at it. These problems may or may not be related to the WD8000 driver, but instead within ktcp. I'd also like to know whether this occurs only after /etc/net.cfg or ktcp parameters are modified between `net stop/net start`, or immediately after boot.

Thank you!

PR #1304, opened 2022-06-01
basic: add HEX$

cocus:
HEX$() converts a number to its hexadecimal equivalent, using C's "%X" format on sprintf().


ghaerr:
Thanks for the enhancement @cocus!

PR #1305, opened 2022-06-02
[fat] Fix bad directory information on large disks

ghaerr:
On large FAT volumes, the directory contents, including file size information, might be incorrect if the directory entry was not in the first 2MB of the FAT volume. There was also a possibility of filesystem corruption if a directory entry was written which was not at the first 2MB of the volume.

This problem was found by @tyama501 in https://github.com/jbruchon/elks/issues/1238#issuecomment-1138847261.

Tested on HD image provided by @tyama501.
Screen Shot 2022-06-02 at 3 30 24 PM


tyama501:
Great!  Thank you!

PR #1309, opened 2022-06-04
[cmds] Add ^C handling to BASIC

ghaerr:
During BASIC program execution, typing ^C will stop execution and display the line number at which the program was stopped, e.g. "100 Break pressed".

INKEY$ now working.

### DIFF
--- a/elkscmd/basic/TODO
+++ b/elkscmd/basic/TODO
@@ -6,19 +6,12 @@ DIM [comma seperated variable/array list]
 THEN <linenumber>
 END
 
-MODE number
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @tyama501,

With these changes, your snake.bas program should run well with uncommenting out the INKEY$ code. ^C can also be used to stop BASIC program execution, which also works during an INPUT statement.

Thank you!

tyama501:
Wow! Thank you  for the quick modification!
I will look it.

PR #1310, opened 2022-06-05
[basic] Use INKEY$ for snake.bas and modified array

tyama501:
Hello @ghaerr ,

I could use INKEY$ !
It is fun to play :)

I also debugged some problems.
I didn't know the array index of ELKS Basic starts from "1".

Thank you!


ghaerr:
> I didn't know the array index of ELKS Basic starts from "1".

There seems to be little consensus between BASICs about a lot of things, making automatic program porting next to impossible. ELKS Basic is compatible with Sinclair Basic with array indices starting with 1.

PR #1313, opened 2022-06-08
[net, el3] New driver for EtherLink III / 3C509 ethernet NICs

Mellvik:
This is the first cut of an EtherLink III (3C509) driver for ELKS.

The driver is work on progress, works OK, needs tuning and has some quirks under heavy load that need to be sorted out. Not yet ping-safe for large packets and flooding. Currently fails if `ktcp` does retransmits. As is, the the driver/NIC combo delivers roughly the same speed as the ne2k when receiving files and in ping responses. 

Tested with 3C509B only, may take minor mods to work with other EtherLink3 adapters. The 3C509B is designed to even work in 8bit ISA slots, such usage will require some driver modifications, akin to what was done with the ne2k driver at one point. 

The interface may be configured (IRQ & Port#) in menuconfig, bootopts or rely on card config. In this version the driver ignores (but reports) its internal settings for IRQ and IOPort and uses the values from bootopts/menuconfig instead. Only TP is supported (many cards have 2 or 3 connection options).

A manual page is pending.


ghaerr:
Hello @Mellvik,

Thanks for the new 3C509 driver, this is really great! I may have a 3C509 card or two, I'll have to take a look around. This should help expand the usability of ELKS quite a bit.

I have a few questions and comments, which follow. I also just noticed that we can't commit as there's a conflict with ports.h.

> The interface may be configured (IRQ & Port#) in menuconfig, bootopts or rely on card config. In this version the driver ignores (but reports) its internal settings for IRQ and IOPort and uses the values from bootopts/menuconfig instead.

I don't see any code that allows IRQ and port to be set from menuconfig. Instead, it seems this driver reads port and irq from ports.h or /bootopts, which is what all the other NIC drivers do, and the way we want, right?

In menuconfig, the NIC is selectable as 3C509, but almost everywhere else 3C509 is not mentioned, but instead EL3 used. I find this a bit confusing as I hadn't actually run across the EL3 acronym before. Perhaps use "3C509/EL3" in ports.h or other areas where programmers or users may not know they're the same thing.

For the functions in el3-asm.S documented as:
```
 //-----------------------------------------------------------------------------
 // Read data from port
 //-----------------------------------------------------------------------------
 // arg1: Port address (int)
 // arg2: Data pointer (char *)
 // arg3: Words to read (int)
 //
 // returns:
 // Nothing

 	.global el3_insw
```
The rest of the kernel ASM libraries use the C declaration directly in order to inform the programmer of the API, which I recommend changing to:
```
//-----------------------------------------------------------------------------
// Read data from port
//-----------------------------------------------------------------------------
// void el3_insw(int port, char *data, int count)
```
Also, in that function, as well as the others that use `cli/sti` to disable/enable interrupts, I recommend putting the CLI as close the start of loop as possible, for the benefit of other drivers that need interrupts enabled for as long as possible:
```
el3_sendpk:
        cli <--- move this to before wr_loop
        push    %bp
        mov     %sp,%bp
        push    %si
	push	%ds

        mov     8(%bp),%cx      // word count
        mov     6(%bp),%si      // buffer addr
 	mov	4(%bp),%dx	// Port

 	mov	%cx,%ax
 	outw	%ax,%dx		// send length
 	xor	%ax,%ax
 	outw	%ax,%dx		// dummy NULL word
 	inc	%cx
 	shr	%cx		// make word count
 	inc	%cx
 	and	$0xfe,%cx	// make dword aligned
 	mov     current,%bx	// setup for far memory xfer
 	mov     TASK_USER_DS(%bx),%ds

        cli <-- add here
 wr_loop:
 	lodsw
 	out     %ax,%dx
 	loop	wr_loop
        sti <-- add here

	pop     %ds
 	pop	%si
 	pop	%bp

 	sti <-- move to end of wr_loop
 	ret
```
If you're looking for absolute speed, there are some other things that can be done instead of building a full BP frame on the stack, not push-saving most registers used, and not having to use (BP) to access arguments. The argument numbers can also be #defined for ease in porting to medium model. See https://github.com/jbruchon/elks/blob/master/elks/arch/i86/lib/fmemory.S for some good ideas.

In el3.c:
Line 129: I don't recommend redefining the C lib `htons` here, this will cause problems later. Instead include <netinet/in.h>.

Line 348-356: The statement "ELKS does not service network interrupts as they arrive. They're used as reminders
 	to service whenever possible" isn't really correct; by convention, the other ELKS drivers don't service network interrupts as they arrive, but instead when the application calls the driver; the interrupt status is checked and used to possibly sleep the application when not ready, which is not really the same as whenever possible. BTW, if you think that the 3C509 driver should be written differently and should immediately service the interrupt by transferring to a buffer, then go ahead and declare a static buffer and do that. I'm not sure that will be any faster at all though, since on older machines, the extra copy for every network packet will be noticeable for those network applications that aren't ready to read the packet, and, any application that is sleeping will then have to be separately woken and the packet copied again, all slower than the current ELKS convention.

I haven't reviewed the rest of the hardcore driver code, but it looks like you're very deep into it!

Thank you!

Mellvik:
Thank you for the feedback, @ghaerr,
Please look for such card, it would be great to have a second pair of hands/eyes on this!

> I have a few questions and comments, which follow. I also just noticed that we can't commit as there's a conflict with ports.h.
> 
I noticed, and it seems that with write/merge access you should be able to resolve that. I'm getting a page with a greyed out 'conflict resolved' (or something like that) button in Git.
> The interface may be configured (IRQ & Port#) in menuconfig, bootopts or rely on card config. In this version the driver ignores (but reports) its internal settings for IRQ and IOPort and uses the values from bootopts/menuconfig instead.
> 
> I don't see any code that allows IRQ and port to be set from menuconfig. Instead, it seems this driver reads port and irq from ports.h or /bootopts, which is what all the other NIC drivers do, and the way we want, right?
> 
Right - and my error. 'Set in menuconfig' is definitely wrong, 'set in ports.h' is right.
> In menuconfig, the NIC is selectable as 3C509, but almost everywhere else 3C509 is not mentioned, but instead EL3 used. I find this a bit confusing as I hadn't actually run across the EL3 acronym before. Perhaps use "3C509/EL3" in ports.h or other areas where programmers or users may not know they're the same thing.
> 
Sure, and I see the point. It may be smart to delay more changes to ports.h until this has been merged though. Or what do you think?
> For the functions in el3-asm.S documented as:
> 
>  //-----------------------------------------------------------------------------
>  // Read data from port
>  //-----------------------------------------------------------------------------
>  // arg1: Port address (int)
>  // arg2: Data pointer (char *)
>  // arg3: Words to read (int)
>  //
>  // returns:
>  // Nothing
> 
>  	.global el3_insw
> The rest of the kernel ASM libraries use the C declaration directly in order to inform the programmer of the API, which I recommend changing to:
> 
> //-----------------------------------------------------------------------------
> // Read data from port
> //-----------------------------------------------------------------------------
> // void el3_insw(int port, char *data, int count)
Good point. I'll change that.
> Also, in that function, as well as the others that use cli/sti to disable/enable interrupts, I recommend putting the CLI as close the start of loop as possible, for the benefit of other drivers that need interrupts enabled for as long as possible:
> 
> el3_sendpk:
>         cli <--- move this to before wr_loop
>         push    %bp
>         mov     %sp,%bp
>         push    %si
> 	push	%ds
> 
>         mov     8(%bp),%cx      // word count
>         mov     6(%bp),%si      // buffer addr
>  	mov	4(%bp),%dx	// Port
> 
>  	mov	%cx,%ax
>  	outw	%ax,%dx		// send length
>  	xor	%ax,%ax
>  	outw	%ax,%dx		// dummy NULL word
>  	inc	%cx
>  	shr	%cx		// make word count
>  	inc	%cx
>  	and	$0xfe,%cx	// make dword aligned
>  	mov     current,%bx	// setup for far memory xfer
>  	mov     TASK_USER_DS(%bx),%ds
> 
>         cli <-- add here
>  wr_loop:
>  	lodsw
>  	out     %ax,%dx
>  	loop	wr_loop
>         sti <-- add here
> 
> 	pop     %ds
>  	pop	%si
>  	pop	%bp
> 
>  	sti <-- move to end of wr_loop
>  	ret
I had a different objective in mind when placing the cli/STI pair like this. The NIC docs recommend turning off all interrupts for the duration of the read/write, not only the actual PIO section. That cannot easily be done from C and may not be necessary. So let's try it and see if it works. For the sendpk-routine, the CLI would have to come before the first outw instruction, which starts the time critical PIO cycle.
> If you're looking for absolute speed, there are some other things that can be done instead of building a full BP frame on the stack, not push-saving most registers used, and not having to use (BP) to access arguments. The argument numbers can also be #defined for ease in porting to medium model. See https://github.com/jbruchon/elks/blob/master/elks/arch/i86/lib/fmemory.S for some good ideas.
> 
Thanks, very interesting - I'll take close look at this. I was actually hoping to find insw/outsw macros (or code) in the kernel library (like linux). Not finding them, this was the next best alternative. By the way, there should be a delay function in ELKS already that does something like the mdelay() asm routine, right?
> In el3.c:
> Line 129: I don't recommend redefining the C lib htons here, this will cause problems later. Instead include <netinet/in.h>.
> 
> Line 348-356: The statement "ELKS does not service network interrupts as they arrive. They're used as reminders
> to service whenever possible" isn't really correct; by convention, the other ELKS drivers don't service network interrupts as they arrive, but instead when the application calls the driver; the interrupt status is checked and used to possibly sleep the application when not ready, which is not really the same as whenever possible.
> 
Agreed, I'll correct that.
> BTW, if you think that the 3C509 driver should be written differently and should immediately service the interrupt by transferring to a buffer, then go ahead and declare a static buffer and do that. I'm not sure that will be any faster at all though, since on older machines, the extra copy for every network packet will be noticeable for those network applications that aren't ready to read the packet, and, any application that is sleeping will then have to be separately woken and the packet copied again, all slower than the current ELKS convention.
> 
I agree. Having a single buffer in between is no improvement (I actually had that in the first cut, until I figured out a way to deal with the peculiarities of the RxComplete interrupt in the NIC). This is as good as it gets given the ELKS 'architecture'. 
> I haven't reviewed the rest of the hardcore driver code, but it looks like you're very deep into it!
> 
Please do when/If you have time. Starting with a working example (such as a Linux driver in this case) always carries the hope that it will be easy. It never is -and porting becomes development instead, and the underlying hardware needs to be understood. It took a while - and there is a lot of potential improvement.

--M

ghaerr:
@Mellvik: I've fixed the merge conflict. Can you push the other fixes above before I commit the whole shebang? I'd rather go that route so that we get everything done at once, and I think I finally understand how to use the "Fix conflict" button. Thank you!

Mellvik:
Thanks @ghaerr, that's great!!!

And agreed, let's keep this sequence of things. I may not be able to push anything till Saturday, but there's no rush!

—M

Mellvik:
@ghaerr,
I ran into more trouble with ports.h when trying to update this PR.
Can you merge this so I get a clean slate?

-M

ghaerr:
> Can you merge this so I get a clean slate?

Yes - but please make the changes discussed above, thank you!

> I ran into more trouble with ports.h when trying to update this PR.

It appears you may not have an actual clean master branch. Make sure you're on master and pull upstream master, then push to your Github fork. I'll merge this and the fsck PR now so you can do that.

Mellvik:
Coming right up - and just to make sure, ports.h is coming later.

—M

PR #1318, opened 2022-06-12
[net] Updates ports.h for 3C509

Mellvik:
Comment updates only. 

For ELKS, Etherlink III is almost singularly the 3C509B card, which is what this driver supports. That said, supporting the EISA 3C529 would need only minimal alterations. There are other 3C5x9 cards - MCA, PCMCIA, ... all unimportant in this context.

So, as @ghaerr pointed out, it might have been appropriate to use the '3C509' term only (or 3C5x9) for this driver. I like the terseness of EL3. 

Other 3Com NICs need entirely different drivers. Like the 3C503, which would need just minor mods of the wd8003 driver.
If such a card pops up, we could fix that quite easily. The AMD LANCE/PCNET is a different story, an entirely different animal. PCNET support would open up VirtualBox to Elks and networking, so it is an attractive target.

### DIFF
--- a/elks/include/arch/ports.h
+++ b/elks/include/arch/ports.h
@@ -16,7 +16,7 @@
  *  6*  HW floppy drive     CONFIG_BLK_DEV_FD       Driver doesn't compile
  *  7   Unused (LPT)
  *  8   Unused (RTC)
- *  9   EL3 (/dev/eth)      CONFIG_ETH_EL3          Optional
+ *  9   3C509/EL3 (/dev/eth) CONFIG_ETH_EL3         Optional
  * 10   Unused (USB)                                Turned off
  * 11   Unused (Sound)                              Turned off
  * 12   NE2K (/dev/eth)     CONFIG_ETH_NE2K         Optional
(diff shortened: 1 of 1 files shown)


ghaerr:
> it might have been appropriate to use the '3C509' term only (or 3C5x9) for this driver. I like the terseness of EL3.

> Other 3Com NICs need entirely different drivers. Like the 3C503, which would need just minor mods of the wd8003 driver.

Given the two statements above, I definitely think calling this 3C509 is a good idea. I'm still not sure what Etherlink III means, is it an umbrella term for 3C509 and 3C503, or something else? In any case, the less confusion for the non-knowledgable (which seems like me at the moment with regards to network card names) the better.

Mellvik:
> ﻿
> it might have been appropriate to use the '3C509' term only (or 3C5x9) for this driver. I like the terseness of EL3.
> 
> Other 3Com NICs need entirely different drivers. Like the 3C503, which would need just minor mods of the wd8003 driver.
> 
> Given the two statements above, I definitely think calling this 3C509 is a good idea. I'm still not sure what Etherlink III means, is it an umbrella term for 3C509 and 3C503, or something else? In any case, the less confusion for the non-knowledgable (which seems like me at the moment with regards to network card names) the better.
> 
Actually, EtherLink III is the name of the chip that runs all of the x9 NICs. Just like the 8390 is the name of the chip that runs the NE2k variants (including the 3C503 BTW). So you'll see a lot of 8390.c files and references in drivers out there. For the EL3, 3C5x9 (and 3C5x9.c files) is more common, while 'Etherlink III' is used a lot in driver text and docs.

I'll 'turn around' the usage in future PRs, so that 509 becomes the dominant 'id'.

-M

PR #1319, opened 2022-06-12
[misc] cleanups in ne2k driver and mknod command.

Mellvik:
Added verified_memcpy to `ne2k` driver, cleanup in the asm-code and `usage()` fix in `mknod`.

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -248,51 +248,6 @@ check_dma_w:
 	pop     %cx
 	ret
 
-#if 0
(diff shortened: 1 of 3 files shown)


ghaerr:
You're almost there on the verified_memcpy_tofs... however, the verification process may fail, in which case that error (which is always -EFAULT) should be returned to the application. Thus, the forms for each should be something like:
```
if ((err = verified_memcpy_tofs(...)) != 0)
        break;
```
(At a higher level, the purpose of the function wrapper is to check that there the pointer passed from the application to the kernel for the return data is valid. This is done by checking that its offset is within the heap+stack area of the process).

Can you please push this change, then we can commit, thanks!

PR #1320, opened 2022-06-14
[net, el3] Bugfixes, cleanups

Mellvik:
Two important bugfixes brings the EL3/3C509 driver to production quality:

1) A bug that prevented outgoing packet sizes above 512 bytes
2) A race condition in select that under heavy load and with large packets would leave receive interrupts off, thus stopping all comms

The driver is now 'flood ping safe' up to and including 1400 byte packets, there are no known bugs and the effective speed is almost on par with the ne2k. There will be occasional lost packets on receive under heavy load or if the host is busy with floppy accesses or big syncs(), they recover nicely.

No tuning has been done, there may be potential for significant improvement by using the NIC's  'early interrupt' features in both send and receive. Also, there is no collection of statistics at this time. TBD.

### DIFF
--- a/elks/arch/i86/drivers/net/el3-asm.S
+++ b/elks/arch/i86/drivers/net/el3-asm.S
@@ -75,7 +75,7 @@ el3_sendpk:
 	inc	%cx
 	shr	%cx		// make word count
 	inc	%cx
-	and	$0xfe,%cx	// make dword aligned
+	and	$0xfffe,%cx	// make dword aligned
 	mov     current,%bx	// setup for far memory xfer
 	mov     TASK_USER_DS(%bx),%ds
 
--- a/elks/arch/i86/drivers/net/el3.c
+++ b/elks/arch/i86/drivers/net/el3.c
@@ -1,6 +1,6 @@
-/* 3c509.c: A 3c509 EtherLink3 ethernet driver for linux. */
 /*
-	For ELKS by Helge Skrivervik (@mellvik) may/june 2022
+	Driver for 3Com Etherlink III/3C5x9 network interface.
+	For ELKS and 3C509 by Helge Skrivervik (@mellvik) may/june 2022
 
 	Adapted from Linux driver by Donald Becker et al.
 
@@ -35,9 +35,6 @@ static int el3_debug = 0;
 #endif
 
 
-/* To minimize the size of the driver source I only define operating
-   constants if they are used several times.  You'll need the manual
-   anyway if you want to understand driver details. */
 /* Offsets from base I/O address. */
 #define EL3_DATA 0x00
 #define EL3_CMD 0x0eU
@@ -65,6 +62,7 @@ enum c509status {
 	IntReq = 0x0040U, StatsFull = 0x0080U, CmdBusy = 0x1000U, };
 
 static int active_imask = IntLatch|RxComplete|TxAvailable|TxComplete|StatsFull;
+
 /* The SetRxFilter command accepts the following classes: */
 enum RxFilter {
 	RxStation = 1, RxMulticast = 2, RxBroadcast = 4, RxProm = 8 };
@@ -101,7 +99,6 @@ enum RxFilter {
 static int el3_isa_probe();
 //static word_t read_eeprom(int, int);
 static word_t id_read_eeprom(int);
-void el3_drv_init();
 static size_t el3_write(struct inode *, struct file *, char *, size_t);
 static void el3_int(int, struct pt_regs *);
 static size_t el3_read(struct inode *, struct file *, char *, size_t);
@@ -110,13 +107,12 @@ static int el3_open(struct inode *, struct file *);
 static int el3_ioctl(struct inode *, struct file *, unsigned int, unsigned int);
 static int el3_select(struct inode *, struct file *, int);
 static void el3_down();
+static void update_stats();
+void el3_drv_init();
 void el3_sendpk(int, char *, int);
 void el3_insw(int, char *, int);
-extern void el3_mdelay(int);
-static void update_stats();
 
-//static int current_tag;
-// static struct net_device *el3_devs[EL3_MAX_CARDS];
+extern void el3_mdelay(int);
 
 /* Maximum events (Rx packets, etc.) to handle at each interrupt. */
 static int max_interrupt_work = 5;
@@ -128,7 +124,6 @@ int net_port = EL3_PORT; /* default IO PORT, changed by netport= in /bootopts */
 
 struct netif_stat netif_stat =
 	{ 0, 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
-static char *mac_addr = (char *)&netif_stat.mac_addr;
 static int ioaddr;	// FIXME  remove later
 
 // Static data
@@ -198,7 +193,9 @@ static int el3_isa_probe( void )
 {
 	short lrs_state = 0xff;
 	int i;
-	word_t *mac = (word_t *)mac_addr;
+	//word_t *mac = (word_t *)mac_addr;
+	word_t *mac = (word_t *)&netif_stat.mac_addr;
+	byte_t *mac_addr = (byte_t *)mac;
 
 	/* ISA boards are detected by sending the ID sequence to the
 	   ID_PORT.  We find cards past the first by setting the 'current_tag'
@@ -283,13 +280,11 @@ static size_t el3_write(struct inode *inode, struct file *file, char *data, size
 	int res;
 
 	while (1) {
-		/*
-		 *	FIXME: May need to block other interrupts for the duration
-		 */
+
 		prepare_to_wait_interruptible(&txwait);
+
 		if (len > MAX_PACKET_ETH) len = MAX_PACKET_ETH;
 		if (len < 64) len = 64;
-		//printk("T%d^", len);
 
 		if (inw(ioaddr + TX_FREE) < (len + 4)) {
 			// NO space in FIFO, wait
@@ -303,7 +298,9 @@ static size_t el3_write(struct inode *inode, struct file *file, char *data, size
 				break;
 			}
 		}
-		//EXPERIMENTAL
+		//printk("T");
+		//printk("T%d^", len);
+		
 		outw(SetIntrEnb | 0x0, ioaddr + EL3_CMD);	// Block interrupts
 		
 		el3_sendpk(ioaddr + TX_FIFO, data, len);
@@ -311,12 +308,9 @@ static size_t el3_write(struct inode *inode, struct file *file, char *data, size
 		/* Interrupt us when the FIFO has room for max-sized packet. */
 		// FIXME: Unsure about this one ... Tune later
 		outw(SetTxThreshold + 1536, ioaddr + EL3_CMD);
-		//EXPERIMENTAL
-		outw(SetIntrEnb | active_imask, ioaddr + EL3_CMD);	// Reenable interrupts
 
 		outb(0x00, ioaddr + TX_STATUS); /* Pop the status stack. */
 						/* may not be the right place to do this */
-		// END EXPERIMENTAL
 
 		/* Clear the Tx status stack. */
 #if 0		// FIXME - check if this is OK to remove.
@@ -332,12 +326,12 @@ static size_t el3_write(struct inode *inode, struct file *file, char *data, size
 			}
 		}
 #endif
-		// FIXME - need a better way to do this. Maybe trust interrupts ...
-		while (inb(ioaddr + EL3_STATUS) & CmdBusy) 
-			el3_mdelay(1);		// Wait - for now, to avoid collision
+		// FIXME - may not be needed. Maybe trust interrupts ...
+		//while (inb(ioaddr + EL3_STATUS) & CmdBusy) 
+			//el3_mdelay(1);		// Wait - for now, to avoid collision
 
 		res = len;
-		//printk("%02x@", inb(ioaddr + TX_STATUS));
+		outw(SetIntrEnb | active_imask, ioaddr + EL3_CMD);	// Reenable interrupts
 		break;
 	}
 	return res;
@@ -390,10 +384,11 @@ static void el3_int(int irq, struct pt_regs *regs)
 			wake_up(&txwait);
 		}
 		if (status & (AdapterFailure | StatsFull | TxComplete)) {
-			//printk("F");
+			printk("F");
 			/* Handle all uncommon interrupts. */
 			if (status & StatsFull)			/* Empty statistics. */
 				update_stats();
+
 #if 0			// Not currently used
 			if (status & RxEarly) {
 				//el3_rx(dev);
@@ -406,7 +401,7 @@ static void el3_int(int irq, struct pt_regs *regs)
 				int i = 4;
 
 				printk("el3: Transmit error, status %02x\n", inb(ioaddr + TX_STATUS));
-				while (--i>0 && (tx_status = inb(ioaddr + TX_STATUS)) > 0) {
+				while (--i > 0 && (tx_status = inb(ioaddr + TX_STATUS)) > 0) {
 					//if (tx_status & 0x38) dev->stats.tx_aborted_errors++;
 					if (tx_status & 0x30) outw(TxReset, ioaddr + EL3_CMD);
 					if (tx_status & 0x3C) outw(TxEnable, ioaddr + EL3_CMD);
@@ -427,7 +422,7 @@ static void el3_int(int irq, struct pt_regs *regs)
 		}
 
 		if (--i < 0) {	// This should not happen
-			printk("eth: EL3 Infinite loop in interrupt, status %4.4x.\n",
+			printk("eth: Infinite loop in interrupt, status %4.4x.\n",
 				   status);
 
 			/* Clear all interrupts. */
@@ -440,10 +435,6 @@ static void el3_int(int irq, struct pt_regs *regs)
 
 	outw(SetIntrEnb | active_imask, ioaddr + EL3_CMD);
 	//printk("i%04x!", inw(ioaddr + EL3_STATUS));
-	if (el3_debug > 4) {
-		pr_debug("EL3: exiting interrupt, status %4.4x.\n",
-			   inw(ioaddr + EL3_STATUS));
-	}
 
 	return;
 }
@@ -539,11 +530,12 @@ static size_t el3_read(struct inode *inode, struct file *filp, char *data, size_
 
 	while(1) {
 		
-		//printk("R");
+		rx_status = inw(ioaddr + RX_STATUS);
+		//printk("R%04x", rx_status);		// DEBUG
 		prepare_to_wait_interruptible(&rxwait);
 
 		//if (!(rx_status & 0x7ff)) {
-		if ((rx_status = inw(ioaddr + RX_STATUS)) & 0x8000) {
+		if (rx_status & 0x8000) {
 			if (filp->f_flags & O_NONBLOCK) {
 				res = -EAGAIN;
 				break;
@@ -554,10 +546,8 @@ static size_t el3_read(struct inode *inode, struct file *filp, char *data, size_
 				break;
 			}
 		}
-		//printk("%d", rx_status&0x7ff);		// DEBUG
-		// Early read is not implemented, thus only one read operation.
-		// Assuming that we actually have data.
-		res = -EIO;
+
+		outw(SetIntrEnb | 0x0, ioaddr + EL3_CMD);	// Block interrupts
 		if (rx_status & 0x4000) { /* Error, update stats. */
 			//short error = rx_status & 0x3800;
 
@@ -576,14 +566,11 @@ static size_t el3_read(struct inode *inode, struct file *filp, char *data, size_
 #endif
 			inw(ioaddr + EL3_STATUS); 				/* Delay. */
 			while (inw(ioaddr + EL3_STATUS) & 0x1000)
-				pr_debug("	Waiting for 3c509 to discard packet, status %x.\n",
+				pr_debug("eth: discard error packet, status %x.\n",
 					  inw(ioaddr + EL3_STATUS) );
 			res = -EIO;
 		} else {
 			short pkt_len = rx_status & 0x7ff;
-			if (el3_debug > 4)
-				pr_debug("Receiving packet size %d status %4.4x.\n",
-					  pkt_len, rx_status);
 			el3_insw(ioaddr + RX_FIFO, data, (pkt_len + 1) >> 1); //Word size
 
 			outw(RxDiscard, ioaddr + EL3_CMD); /* Pop top Rx packet. */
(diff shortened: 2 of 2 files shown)


Mellvik:
@ghaerr,
there is a minor update just around the corner for this PR. Turns out there was a bug preventing the `bootopts`-specified IRQ from being set in the NIC at initialization time.

-M

ghaerr:
Wow, fast work on getting the driver fully operational!! Nice!

> A race condition in select that under heavy load and with large packets would leave receive interrupts off, thus stopping all comms.

Yes, this is tricky work getting right. Was this solved by reading the rx status prior to the prepare_to_wait_interruptible call?

I notice that you're calling wake_up from the read routine. This may not be proper because that would mean that there must be a process calling read in order to wakeup any waiting task, which wouldn't be possible if there was only one application with the NIC driver open and it was put to sleep. Normally the HW interrupt routine should be the only routine calling wakeup. You've got quite a few changes so I haven't had the chance to fully review, which I'll do after the commit.

Thank you!

Mellvik:
> A race condition in select that under heavy load and with large packets would leave receive interrupts off, thus stopping all comms.
> 
> Yes, this is tricky work getting right. Was this solved by reading the rx status prior to the prepare_to_wait_interruptible call?
> 
Some head scratching, but the problem was finally located in select, which reads the RX status, gets interrupted by a RxComplete, but still acts on the RX status just read. The solution was to turn off NIC interrupts during select.
> I notice that you're calling wake_up from the read routine. This may not be proper because that would mean that there must be a process calling read in order to wakeup any waiting task, which wouldn't be possible if there was only one application with the NIC driver open and it was put to sleep. Normally the HW interrupt routine should be the only routine calling wakeup. You've got quite a few changes so I haven't had the chance to fully review, which I'll do after the commit.
> 
I was going to ask for an opinion on this as it didn't seem or feel right. Here's the thing:

When an RX interrupt is received, the RxComplete interrupt is turned off because we cannot service it right away (otherwise it will be ringing).
The idea with the wakeup in read was to signal back if there is more to read without turning the rx interrupt back on because that's a waste of time if we already know there is more data available.
It occurred to me though, after figuring out the race condition in select above, that read is not the place to do this, while select may be. That is, when a read select finds no more data to read, it can reenable the RxComplete interrupt before returning. 
Or I just cave in an re-enable interrupts in read and take the extra interrupts. I've tested that and it does work. I haven't tested it under heavy load though.

I intend to check out both. Opinions solicited. 

Thank you.


—M

ghaerr:
@Mellvik,

The business of getting things right with regards to writing correct kernel drivers is complicated. The best advice I can give is 1) understand how the kernel scheduling really works; 2) keep things simple (don't hack kluge cases to get things to work); and 3) write the driver entry points using a top-down approach, using procedures to manage complexity.

> The idea with the wakeup in read was to signal back if there is more to read

Wakeup doesn't "signal back" anything, it just executes a for loop through the task structures and possibly sets the status flag of a task structure to TASK_RUNNING; this never causes any immediate rescheduling or procedure call, the single kernel thread remains running uninterrupted with the exception of HW interrupts, just as it always does.

That said, perhaps it is now time to truly learn 1) above. Looking at the read and select code in the NIC, it is very important to see how race conditions are properly prevented in any kernel code <-> hw interrupt handler. There is only one mechanism that works, and the basic idea is to have a sleep-check routine, and a mechanism to cause a kernel thread to start sleeping. All wakeups are always called from the HW interrupt. The basic code to understand is in kernel/sleepwake.c, and it is quite subtle:
```
/*
 * Simple, race-safe versions of sleep/wait functions. Greg Haerr Aug 2020.
 * Only required when wait queue used or condition check changed by hw interrupt.
 *
 * Old:
 *  if (!ready())
 *      interruptible_sleep_on(&waitq);
 * New:
 *  prepare_to_wait_interruptible(&waitq);
 *  if (!ready())
 *      do_wait();
 *  finish_wait(&waitq);
 */

void prepare_to_wait_interruptible(struct wait_queue *p)
{   
    current->state = TASK_INTERRUPTIBLE;
    wait_set(p);
}

void do_wait(void)
{   
    schedule();
}

void finish_wait(struct wait_queue *p)
{   
    current->state = TASK_RUNNING;
    wait_clear(p);
}
```
Notice in the first code comment that *any* check outside of prepare_to_wait/finish_wait is unsafe. Try to find out why (hint: hw interrupt changes task run state after it's already been changed but before schedule, only happens in race condition).

I don't understand the EL3 chip nor your driver enough to respond about exactly what to do yet, but the basic upper-level structure of a kernel driver needs to remain the same - read/write and select calls need to test a status, then possibly sleep the task; the hw interrupt will reawaken them, where they probably don't need to retest, but just execute the read/write packet, and return.

As I now look hard at the NE2K driver, I see that it is possible we have a race condition with select(), which could be what you're seeing in the 3C509: the ready check isn't protected by a prepare_to_wait_interruptible/finish_wait block, which could allow for the problem described in the first code comment above to occur. I will have to look at the select_wait code to see if this is true, as it's a bit complicated in how it works.

ghaerr:
Continuing to look at the kernel select code regarding my comment about the ready check not being protected - I noticed this problem several years ago and fixed it, the internal select routine will loop, and continue to call the ready check routine, such that the task should not be put to sleep and not get woken up:
```
static int do_select(int n, fd_set * in, fd_set * out, fd_set * ex,
                     fd_set * res_in, fd_set * res_out, fd_set * res_ex)
{
...
  repeat:
    /* Note: Race condition here where wake_up_process sets TASK_RUNNING state
     * but check()/fops->select returns 0. This then causes schedule() to
     * reschedule current task since current->state == TASK_RUNNING by wake_up.
     */
    current->state = TASK_INTERRUPTIBLE; <-- this resets task state to sleep until actually ready by NIC select call
    memset (current->poll, 0, sizeof (struct wait_queue *) * POLL_MAX);
    filp = current->files.fd;
    for (i = 0; i < n; i++, filp++) {
        if (*filp) {
            if (FD_ISSET(i, in) && check(SEL_IN, *filp)) { <--- these routines call the NIC select functions repeatedly until ready
                FD_SET(i, res_in);
                count++;
            }
            if (FD_ISSET(i, out) && check(SEL_OUT, *filp)) { <--- same
                FD_SET(i, res_out);
                count++;
            }
            if (FD_ISSET(i, ex) && check(SEL_EX, *filp)) {
                FD_SET(i, res_ex);
                count++;
            }
        }
    }
    if (!count && current->timeout && !(current->signal /* & ~currentp->blocked */ )) {
        debug_sched("select(%d): timeout %lx\n", current->pid, current->timeout); <--- uncomment this to see what is happening
        schedule();
        goto repeat; <-- this goto fixes potential race condition of sleeping forever by retrying ready check
    }
```
Thus, I think we're ok on the select not being guarded by prepare_to_wait_interruptible.

ghaerr:
> When an RX interrupt is received, the RxComplete interrupt is turned off because we cannot service it right away (otherwise it will be ringing).

This could possibly be enhanced by saving RX interrupt received state to static variable, which is then tested (OR'd) into the actual status and cleared afterwards. In this way if you have to turn off interrupts (not recommended), at least the next status calls (from read/write/select) will "remember" that an RX receive occurred.

> The idea with the wakeup in read was to signal back if there is more to read without turning the rx interrupt back on because that's a waste of time if we already know there is more data available.

I very highly recommend removing any wakeup from read, it can't work with multiple applications (even though "currently" we have only one app ever reading a packet). The "if there is more to read" can be solved by checking a static copy of the status as described above.

> It occurred to me though, after figuring out the race condition in select above, that read is not the place to do this, while select may be. That is, when a read select finds no more data to read, it can reenable the RxComplete interrupt before returning.

It sounds to me like the place to do all of this is in a single read-status procedure. That procedure can keep track of receiver/transmitter state, possibly save a static copy of it for calls to itself without reenabling interrupts, and also reenable interrupts just before return RX or TX ready, for later continuance.

> Or I just cave in an re-enable interrupts in read and take the extra interrupts.

I'm not a fan of continually enabling/disabling interrupts unless you have to. The HW interrupt handler should be smart enough to know what to do, especially if some previous state is held in a tightly-bound static variable. The simpler, the better, and use a procedure wrapper around all this to manage complexity. For 3C509, this sounds like all could be put in the hw interrupt and/or ready-check procedures and leave all the other codes vastly simplified.

Mellvik:
Thank you very much for there run-through, @ghaerr - very useful indeed. Appreciated.


> 
> I don't understand the EL3 chip nor your driver enough to respond about exactly what to do yet, but the basic upper-level structure of a kernel driver needs to remain the same - read/write and select calls need to test a status, then possibly sleep the task; the hw interrupt will reawaken them, where they probably don't need to retest, but just execute the read/write packet, and return.
> 
As we've discussed before, the challenge is to get around the NIC's assumption that servicing a read_completion interrupt means relieving it of all pending data. Obviously - from your run-through above - the ambition to actually empty the NIC receive queue before reenabling RxComplete interrupts is futile. No big deal. As it turns out, received packets don't stack up that often, in 95% of the cases there is only one packet in the FIFO when we service it (and - under heavy load, one on its way in). So going back to 'plan C', just reenable receive completion interrupts at the end of the read routine is fine. Which means processing each received packets via the interrupt routine. This woks flawlessly.

What I'm seeing - and this is regardless of the fruitless experimentation with wakeups, is that select some times seems to signal data available when there isn't - under heavy load. Never a problem, just an occasional 
ktcp: eth_process error -1 (errno 5), discarding packet

I'm investigating this to see if there is a way to avoid it. It may be as simple as the NICs way of signalling overrun errors. And that the error detection mechanics in the driver aren't good enough. For sure, they haven't been much tested yet. 
> As I now look hard at the NE2K driver, I see that it is possible we have a race condition with select(), which could be what you're seeing in the 3C509: the ready check isn't protected by a prepare_to_wait_interruptible/finish_wait block, which could allow for the problem described in the first code comment above to occur. I will have to look at the select_wait code to see if this is true, as it's a bit complicated in how it works.
> 
Interesting. Of course it's tempting to just assume the 'if it ain't broken don't fix it' attitude, given that the n2ek driver works really well. OTOH there are obviously more stuff to be understood here. Interrupt protection/isolation in the ne2k driver is much more complicated than in the el3 driver, allowing for (potentially) better performance. For example, at no point except in the interrupt service routine itself, are NIC interrupts completely turned off, and not explicitly, but by the way things work. It's also more risky which has beaten me many times. And rewarding when it works.
Back to the ne2k driver: The read select check for availability is extremely simple, just testing a variable (ne2k_has_data), so there isn't much room for a race condition, but it's not impossible. For added protection I could add interrupt isolation like in the el3-driver, if you think it makes sense.
BTW - given the nature of such race conditions, it would be useful to expand the testing to machines other than the 20MHz 386 compaqs I'm using. I'll pull out the 286/12 tomorrow, but it would be even more useful to test with faster iron. Maybe I'll send you one of my ne2k-s   :-).

While writing this, a quarter of a million 1400 byte pings have bounced off the modified el3 driver (with the wakeups outside the interrupt routine removed).

—M

Mellvik:
metrics:

```
helge@raspi2b:~ $ sudo ping -f -s 1400 10.0.2.15
[sudo] password for helge: 
PING 10.0.2.15 (10.0.2.15) 1400(1428) bytes of data.
..................................................................................................^C
--- 10.0.2.15 ping statistics ---
273750 packets transmitted, 273652 received, 0% packet loss, time 2670666ms
rtt min/avg/max/mdev = 9.338/9.584/517.388/2.352 ms, pipe 26, ipg/ewma 9.755/9.521 ms
helge@raspi2b:~ $ 
```

ghaerr:
> What I'm seeing - and this is regardless of the fruitless experimentation with wakeups, is that select some times seems to signal data available when there isn't - under heavy load.

This is entirely the result of select returning 1 when it should return 0. See your code in el3:
```
    switch (sel_type) {
        case SEL_OUT:
            // FIXME - may need a more accurate tx status test
            if (inw(ioaddr + TX_FREE) < MAX_PACKET_ETH) {
                select_wait(&txwait);
                break;
            }
            res = 1;
            break;

        case SEL_IN:

            // Don't use RxComplete for this test, it has been masked out!
            if (inw(ioaddr+RX_STATUS) & 0x8000) {
                //printk("s%x", inw(ioaddr+RX_STATUS));
                select_wait(&rxwait);
                break;
            }
            res = 1;
            break;
```
After select_wait returns, it always returns 1. If you're fiddling with all sorts of special cases elsewhere, perhaps the status should be rechecked. Keep in mind this is very tricky: the el3_select routine is actually polled by the kernel every timer interrupt, that's how select works. It is important to learn exactly how the kernel sleep and select routines work, if you deviate from a working template (such as ne2k.c).

> The read select check for availability is extremely simple, just testing a variable (ne2k_has_data),

Testing a variable alone is not good enough for race conditions, the test must be protected by prepare_to_wait_interruptible (except select, since it loops) - see my posts above. 

> so there isn't much room for a race condition, but it's not impossible. 

"Isn't much room" doesn't work for race elimination. There must be absolutely no room. See sleepwake.c

>  given the nature of such race conditions, it would be useful to expand the testing to machines other than the 20MHz 386 compaqs I'm using. I'll pull out the 286/12 tomorrow, but it would be even more useful to test with faster iron.

Nothing wrong with testing on more machines, that's for sure. I think the kernel has properly written routines for sleep/wake and select race conditions - its up to the programmer to use them correctly, and understand how kernel scheduling works. Spend the time to study kernel/sleepwake.c and fs/select.c.

> While writing this, a quarter of a million 1400 byte pings have bounced off the modified el3 driver (with the wakeups outside the interrupt routine removed).

Pretty cool, for sure!

> Maybe I'll send you one of my ne2k-s   :-).

I thought I had one - I'm having major issues with my 386 motherboard and getting the NIC up, just like @toncho11 ran into... not sure what to do. And I don't have DOS on that system to test an original-era setup, only Windows XP. (No, I don't want to get into XP again!)

Mellvik:
@ghaerr,
I'm not following you on this one.
> This is entirely the result of select returning 1 when it should return 0. See your code in el3:
> 
>     switch (sel_type) {
>         case SEL_OUT:
>             // FIXME - may need a more accurate tx status test
>             if (inw(ioaddr + TX_FREE) < MAX_PACKET_ETH) {
>                 select_wait(&txwait);
>                 break;
>             }
>             res = 1;
>             break;
> 
>         case SEL_IN:
> 
>             // Don't use RxComplete for this test, it has been masked out!
>             if (inw(ioaddr+RX_STATUS) & 0x8000) {
>                 //printk("s%x", inw(ioaddr+RX_STATUS));
>                 select_wait(&rxwait);
>                 break;
>             }
>             res = 1;
>             break;
> After select_wait returns, it always returns 1.
> 
it returns 0 as it should. res is initialized to 0.

-M
>

ghaerr:
> it returns 0 as it should. res is initialized to 0.

Good catch - you are correct, and I also spoke incorrectly about how select actually works, I have had to reread it to come back to fully remembering how it works: the select routine does NOT get called every timer interrupt, like I said above.

> Keep in mind this is very tricky: the el3_select routine is actually polled by the kernel every timer interrupt, that's how select works.

My statement above is incorrect.

Select works very trickily by calling the driver el3_select routine ONLY once; this code performs the standard read or write status ready, and either returns 1 immediately, or calls select_wait(). It does not matter what is returned after the call to select_wait(), as the process is put to sleep. The process remains sleeping until a wakeup call is executed, which is then compared to the list of "wait addresses" that were stored in slots for each file descriptor in select calling select_wait(). If one of the addresses matches, it means that the wakeup can apply to the process, since one of the addresses is in a currently executing sys_select file descriptor.

Skipping the complicated stuff - the solution is that read, write and select ALL MUST use the same read (or write) status check, or things won't work correctly. That is (and look at ne2k.c to clarify), all read or select IN routines call the same status routine check "ne2k_rx_stat() != NE2K_STAT_RX" and all write or select OUT routines use "ne2k_tx_stat() != NE2K_STAT_TX".
That is, the programmer must use the same criterion for sleeping a read as for sleeping a select IN. Since el3.c uses a seperate criterion for sleeping, you're getting the strange select result:
```
el3_write:
 if (inw(ioaddr + TX_FREE) < (len + 4)) {
            // NO space in FIFO, wait

el3_select:
  switch (sel_type) {
        case SEL_OUT:
            // FIXME - may need a more accurate tx status test
            if (inw(ioaddr + TX_FREE) < MAX_PACKET_ETH) {
                select_wait(&txwait);
```
I hope this makes sense. Please refer and study kernel/sleepwake.c and fs/select.c, it is worth it!

Mellvik:
Thank you @ghaerr - 
> Keep in mind this is very tricky: the el3_select routine is actually polled by the kernel every timer interrupt, that's how select works.
> 
> My statement above is incorrect.
> 
FWIW -this is how select USED to work before you fixed it. I vaguely remember our discussion about that back in (say) 2018 or -19...
> Select works very trickily by calling the driver el3_select routine ONLY once; this code performs the standard read or write status ready, and either returns 1 immediately, or calls select_wait(). It does not matter what is returned after the call to select_wait(), as the process is put to sleep. The process remains sleeping until a wakeup call is executed, which is then compared to the list of "wait addresses" that were stored in slots for each file descriptor in select calling select_wait(). If one of the addresses matches, it means that the wakeup can apply to the process, since one of the addresses is in a currently executing sys_select file descriptor.
> 
That's an educating narrative of how this works. I'm assuming that if there is no wait address found in the list, the wakeup call is just a no-op?
> Skipping the complicated stuff - the solution is that read, write and select ALL MUST use the same read (or write) status check, or things won't work correctly. That is (and look at ne2k.c to clarify), all read or select IN routines call the same status routine check "ne2k_rx_stat() != NE2K_STAT_RX" and all write or select OUT routines use "ne2k_tx_stat() != NE2K_STAT_TX".
> That is, the programmer must use the same criterion for sleeping a read as for sleeping a select IN. Since el3.c uses a seperate criterion for sleeping, you're getting the strange select result:
> 
> el3_write:
>  if (inw(ioaddr + TX_FREE) < (len + 4)) {
>             // NO space in FIFO, wait
> 
> el3_select:
>   switch (sel_type) {
>         case SEL_OUT:
>             // FIXME - may need a more accurate tx status test
>             if (inw(ioaddr + TX_FREE) < MAX_PACKET_ETH) {
>                 select_wait(&txwait);
> I hope this makes sense. Please refer and study kernel/sleepwake.c and fs/select.c, it is worth it!
> 
I get your point - and agree in principle. The difference between the two in this case is intentional and not possibly harmful.
Here's the thing: I don't know if write ever gets called w/o a preceding select. If it is, testing available space against the amount of data to be written is an optimization.
In select, we obviously don't know how much data the client may have so we test for the max, a full packet. Always overbooking, and that's the nature of things.
If a write is always preceded by select, we're fine. If not, we're fine. And we're not seeing any strange results anywhere.

Like I said before, the only cases where I'm seeing errors (and intended to investigate whether select was part of it) is in read under heavy load. This is most likely overrun errors that should be caught earlier (in the interrupt handler) and has nothing to do with select. This is the only issue left in the el3 driver at this point (the setting of the IRQ has been fixed but not pushed).

—M
>

ghaerr:
> I'm assuming that if there is no wait address found in the list, the wakeup call is just a no-op?

Yes. Since every wakeup call passes a "wait queue" address, this is matched against either each task's waitpt pointer, and in the case of select wait, also against each entry in the select wait array. This is how arbitrary wakeup calls get matched to tasks.

> I get your point - and agree in principle. The difference between the two in this case is intentional and not possibly harmful.

I'm just trying to let you know how the kernel works, and that having seperate tests for read vs select IN (or write vs select OUT) is not a good idea.
When select or read are forced to wait for a packet, why would one want to code a different gate-check? The meaning of select IN returning 1 is "read-ready", while then the read itself may sleep, due to a different check. That'll screw up certain programs that depend on read not blocking after select. Etc, etc.

>  don't know if write ever gets called w/o a preceding select. 

And you shouldn't make any assumptions, either.

> If it is, testing available space against the amount of data to be written is an optimization.

In the NIC driver, we can't ever not write the full passed packet, since the kernel doesn't support write retries for network writes. Thus, both select and write MUST assume a possible max-sized packet, or what might happen is a write blocks. While this won't kill ktcp, it introduces hard-to-diagnose behaviors because ktcp is depending on a write to succeed, after select returns 1. That causes ktcp to block in its main loop, throwing off its internal timing, which may introduce other undesirable behavior, or worse, dropped or overrun read packets due to waiting for a write that was previously OK'd. The packet size isn't passed to select so its bad practice to assume an optimization if it might cause a task sleep after having been OK'd by a previous select.

> the only cases where I'm seeing errors (and intended to investigate whether select was part of it) is in read under heavy load. This is most likely overrun errors that should be caught earlier (in the interrupt handler) and has nothing to do with select.

As discussed above, this could definitely be the ktcp main loop blocking on a write instead of reading, and then having another received packet overrun the NIC buffer - all because of using different code gates for write and select OUT. So it could have to do with select. It's best to get things working first, then optimize later.

Thank you for putting in all the hard work of getting a new NIC driver working! I'm coming from the perspective of how to ensure a correct kernel, while perhaps you're trying to optimize to a certain arrangement of the way ktcp interacts with a NIC driver. While ELKS definitely needs optimizations to help networking throughput, some non-standard, the danger here is that when/if ktcp or other higher level network applications change, the assumptions made for optimization in NIC could cause failure.

Mellvik:
Thank you @ghaerr.
> I get your point - and agree in principle. The difference between the two in this case is intentional and not possibly harmful.
> 
> I'm just trying to let you know how the kernel works, and that having seperate tests for read vs select IN (or write vs select OUT) is not a good idea.
> 
I agree, and AFAIK the tests are not separate, they are the same - one simply using knowledge not available to the other (a number).
> When select or read are forced to wait for a packet, why would one want to code a different gate-check? The meaning of select IN returning 1 is "read-ready", while then the read itself may sleep, due to a different check. That'll screw up certain programs that depend on read not blocking after select. Etc, etc.
> 
This isn't select_in, it's select_out. While select_in and read have no clue about the amount of datas coming except with the max limit, that's not the case with select_out and write. In write we know exactly how many bytes to write. So testing for the availability of that many bytes cannot be harmful.
> don't know if write ever gets called w/o a preceding select.
> 
> And you shouldn't make any assumptions, either.
> 
I don't, the tests are safe regardless.
> 
> the only cases where I'm seeing errors (and intended to investigate whether select was part of it) is in read under heavy load. This is most likely overrun errors that should be caught earlier (in the interrupt handler) and has nothing to do with select.
> 
This part has now been sorted out and fixed in the driver. A leftover from the 'model' linux driver, the error handling was in read, we need to have it in the interrupt routine so it can be handled immediately. 

This is (for now :-) ) the final fix. there are occasional overrun errors under heavy load, as is unavoidable, but they now longer affect ktcp. Also the error counters have been added. As have the open counter. 

PR some time tomorrow I hope.

> Thank you for putting in all the hard work of getting a new NIC driver working! I'm coming from the perspective of how to ensure a correct kernel, while perhaps you're trying to optimize to a certain arrangement of the way ktcp interacts with a NIC driver. While ELKS definitely needs optimizations to help networking throughput, some non-standard, the danger here is that when/if ktcp or other higher level network applications change, the assumptions made for optimization in NIC could cause failure.
> 
> 
I appreciate that. And I think it's good coming from different perspectives. 

—M

PR #1326, opened 2022-06-17
[net, el3] More cleanups and fixes for the el3 driver

Mellvik:
Another round of updates for the 3C509B driver. Notably
- a bunch of important small fixes (setting the IRQ from bootopts now works)
- reworked the error handling, extensively tested (I had erroneously assumed that the AdapterFailure interrupt covered rx overruns, it does not).
- added error stats, may be retrieved via IOCTL
- added multiple opens to allow access to the IOCTL
- removed extraneous debug code

Like other interfaces, the 3C509B will be overrun by packets under heavy load. The NIC's total buffer capacity (rx & tx) is 8Kbytes. The 3C509 (without the 'B') has only 4Kbytes, and some other differences. It should still work out of the box with this driver, but will have performance problems because of the limited buffer space.

When overflows happen, the NICs needs to be reset – which is just a simple command, not a complex procedure like in 8390 based NICs. Still, when reset, all buffered data will be discarded, which may be one or more packets. The error message has thus been changed from 'packet discarded' to 'buffer cleared'. 

The ultimate 'beating' of any NIC is a flood ping (very large or very small packets depending on what we want to test) from a fast machine like a recent PC or a Mac (my RaspberryPi 2B maxes out at about 120 1200byte packets per second which is a good beating but slow compared to the other 'beasts') while running the mount command with a (big) FAT file system mounted. In my case this will render the system literally dead for 10-15 seconds.

Such a test will deliver tens - more likely hundreds - of overflow errors, depending on the size of the mounted file system. The test is how well the NIC recovers - repeatedly. I.e. no additional errors, no hang, full functionality when the 'service blockade' from the disk IO is over.

The EL3 driver and NIC now passes this test with flying colors, as does the NE2k 16bit driver. The NE2k 8 also passes, but no 'flying colors'.

The EL3 FIFO architecture provides for 'early read' and 'early write' - the ability to start reading a packet into memory before reception is complete and similarly, transfer data to the NIC FIFO when there is enough space instead of waiting for the previous packet to complete sending. It is likely that taking advantage of these abilities would increase performance in the ELKS environment. For example, since there is always a significant delay in ELKS between the arrival of a RX interrupt and the actual reading of data, using early interrupt (say interrupt @ 60 or 100 bytes) could make a significant difference for incoming file transfers (large packets). That's an experiment for a later day and time.

### DIFF
--- a/elks/arch/i86/drivers/net/el3.c
+++ b/elks/arch/i86/drivers/net/el3.c
@@ -28,13 +28,6 @@
 #include <linuxmt/netstat.h>
 #include <netinet/in.h>
 
-#ifdef EL3_DEBUG
-static int el3_debug = EL3_DEBUG;
-#else
-static int el3_debug = 0;
-#endif
-
-
 /* Offsets from base I/O address. */
 #define EL3_DATA 0x00
 #define EL3_CMD 0x0eU
@@ -61,7 +54,7 @@ enum c509status {
 	TxAvailable = 0x0008U, RxComplete = 0x0010U, RxEarly = 0x0020U,
 	IntReq = 0x0040U, StatsFull = 0x0080U, CmdBusy = 0x1000U, };
 
-static int active_imask = IntLatch|RxComplete|TxAvailable|TxComplete|StatsFull;
+static int active_imask = IntLatch|RxComplete|TxAvailable|TxComplete|AdapterFailure|StatsFull;
 
 /* The SetRxFilter command accepts the following classes: */
 enum RxFilter {
@@ -117,13 +110,11 @@ extern void el3_mdelay(int);
 /* Maximum events (Rx packets, etc.) to handle at each interrupt. */
 static int max_interrupt_work = 5;
 
-#define pr_debug printk
-
 int net_irq = EL3_IRQ;  /* default IRQ, changed by netirq= in /bootopts */
 int net_port = EL3_PORT; /* default IO PORT, changed by netport= in /bootopts */
 
 struct netif_stat netif_stat =
-	{ 0, 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
+	{ 0, 0, 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
 static int ioaddr;	// FIXME  remove later
 
 // Static data
@@ -214,8 +205,8 @@ static int el3_isa_probe( void )
 	}
 	outb(0xd0, el3_id_port);	// Set tag
 
-	outb(0xd0, el3_id_port);			// select tag (0)
-	outb(0xe0 |(ioaddr >> 4), el3_id_port );	// Activate
+	outb(0xd0, el3_id_port);		// select tag (0)
+	outb(0xe0 |(ioaddr >> 4), el3_id_port );// Set IOBASE address, activate
 	printk("eth: 3C509 @ IRQ %d, port %02x", net_irq, ioaddr);
 
 	if ((i = id_read_eeprom(EEPROM_MFG_ID)) != 0x6d50) {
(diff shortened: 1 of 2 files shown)


ghaerr:
Nice improvements, and thank you for the detailed explanation of how things work!

I'm not sure 'usecount' belongs in the external netif_stat structure - that variable is really internal to the driver, and could be declared static unsigned char there. Having it or other non-reporting values external also means that all external programs (netstat, ktcp, etc) have to be recompiled with driver changes.

This is a super enhancement for ELKS, being that 3C509 is such a common NIC. I hope we can get others (including myself) to test and play with this.

Thank you!

Mellvik:
> 
> I'm not sure 'usecount' belongs in the external netif_stat structure - that variable is really internal to the driver, and could be declared static unsigned char there. Having it or other non-reporting values external also means that all external programs (netstat, ktcp, etc) have to be recompiled with driver changes.
> 
Good point - I'll put that on the list for next fix. And make sure the ne2k gets it right before I push the multiple open fix.

Slightly related 1: I got the wd8003 today, will check it out asap. Reading some driver code it seems there is minimal difference beween the 8003 and the 8013 other than the latter being 16 bit, an possibly with more buffer space. We may have to change the netram= parm  to include the end address too, in order to accomodate different buffer sizes unless I find a reliable way of autodetection.

Slightly related 2: I took another dive into VirtualBox to see if it's worthwhile to create a driver for AMD Lance PCnet II for ELKS in order to make the combination attractive. This is the only OLD NIC supported by VirtualBox.  It turns out that while VirtualBox does emulate the PCnet II ISA card, it does so only in Bus Mastering DMA mode, not shared mem. So it's a no-go. I don't feel inclined to play with ELKS and DMA right now :-) .
In addition, these cards (typically the NE2100) are really hard to come by and expensive, so there is definitely no 'market' waiting for Lance support in ELKS. 

While - as you point out - there are plenty of 3C509s for almost no money available, so we should be fine.

--M

ghaerr:
> We may have to change the netram= parm  to include the end address too, in order to accomodate different buffer sizes unless I find a reliable way of autodetection.

I'm not sure if we should get the user into trying to determine the buffer size of the NIC. Would it be possible to use another bit (say, in netport=), to essentially specify the NIC type, and have that internally adjust the shared RAM or other NIC-specific details? In a sense what we're seeing is the possible need for more detailed information about the NIC. We might be better off to have the card "name" and/or variation specified in config/bootopts and have the driver determine what to do from that. IMO that will keep it easier for users to get a NIC running.

Mellvik:
> We may have to change the netram= parm to include the end address too, in order to accomodate different buffer sizes unless I find a reliable way of autodetection.
> 
> I'm not sure if we should get the user into trying to determine the buffer size of the NIC. Would it be possible to use another bit (say, in netport=), to essentially specify the NIC type, and have that internally adjust the shared RAM or other NIC-specific details? In a sense what we're seeing is the possible need for more detailed information about the NIC. We might be better off to have the card "name" and/or variation specified in config/bootopts and have the driver determine what to do from that. IMO that will keep it easier for users to get a NIC running.
> 
I may be possible to do that. We may need to restrict the types supported, but let's take that as it comes. 
From reading the Linux sources, there are many variants, and type may not be enough to determine the buffer size. FWIW - Linux uses start and end address settings in configuration, but then again, supports variants to the moon and back.

We'll get back to this when I have the cards operational.

--M.


>

ghaerr:
@Mellvik,

The last EL3 PR has broken NE2K - this is for the reason discussed above, which is struct netif_stat changed, and ne2k.c wasn't changed with it. Just shows the dangers of externalizing data, without looking at all other source files that use the struct.

So we'll want an updated PR with usecount removed ASAP:
```
ne2k.c:31:2: warning: braces around scalar initializer
  { 0, 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
  ^
ne2k.c:31:2: note: (near initialization for ‘netif_stat.usecount’)
```
The compiler is complaining about the braces around the MAC address, but what this really means is that the initialization is in error, since the struct was changed without updating ne2k.c.

Mellvik:
Thank you @ghaerr,

I just noticed - I'm pushing a fix now which removes this variable from the netstat struct.

--M

PR #1331, opened 2022-06-18
[net] Remove usecount from netstat struct, make local to the drivers.

Mellvik:
Fix re. https://github.com/jbruchon/elks/pull/1326#issuecomment-1159078240

Remove the new `usecount` counter (to allow multiple opens of `/dev/eth`) from shared netstat struct to driver static.

### DIFF
--- a/elks/arch/i86/drivers/net/el3.c
+++ b/elks/arch/i86/drivers/net/el3.c
@@ -114,10 +114,10 @@ int net_irq = EL3_IRQ;  /* default IRQ, changed by netirq= in /bootopts */
 int net_port = EL3_PORT; /* default IO PORT, changed by netport= in /bootopts */
 
 struct netif_stat netif_stat =
-	{ 0, 0, 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
+	{ 0, 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
 static int ioaddr;	// FIXME  remove later
+static int usecount = 0;
 
-// Static data
 struct wait_queue rxwait;
 struct wait_queue txwait;
 
@@ -516,7 +516,7 @@ static void update_stats( void )
 
 static void el3_release(struct inode *inode, struct file *file)
 {
-	if (--netif_stat.usecount == 0) {
+	if (--usecount == 0) {
 		el3_down();
 
 		/* Switching to window 0 disables the IRQ. */
@@ -613,7 +613,7 @@ static int el3_open(struct inode *inode, struct file *file)
 
 	if (!(netif_stat.if_status & NETIF_FOUND)) 
 		return(-EINVAL);	// Does not exist 
-	if (netif_stat.usecount++)
+	if (usecount++)
 		return(0);		// Already open
 
 	EL3WINDOW(0);		// TESTING ONLY
--- a/elks/arch/i86/drivers/net/ne2k.c
+++ b/elks/arch/i86/drivers/net/ne2k.c
@@ -28,9 +28,9 @@
 int net_irq = NE2K_IRQ;	/* default IRQ, changed by netirq= in /bootopts */
 int net_port = NE2K_PORT; /* default IO PORT, changed by netport= in /bootopts */
 struct netif_stat netif_stat = 
-	{ 0, 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
+	{ 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
 
-// Static data
+static int usecount = 0;
 struct wait_queue rxwait;
 struct wait_queue txwait;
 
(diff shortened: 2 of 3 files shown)


ghaerr:
Thanks.

I recommend, in your next PR, changing "static int usecount = 0" to "static unsigned char usecount". The reason is, for the kernel, using unsigned char saves kernel data segment space, and when initialized to 0, increases /linux file size, whereas uninitialized puts the variable in .bss, which is guaranteed to be initialized to 0, but does not increase the file size.

Mellvik:
> I recommend, in your next PR, changing "static int usecount = 0" to "static unsigned char usecount". The reason is, for the kernel, using unsigned char saves kernel data segment space, and when initialized to 0, increases /linux file size, whereas uninitialized puts the variable in .bss, which is guaranteed to be initialized to 0, but does not increase the file size.
> 
> 
Thank you @ghaerr - I didn't know that.

There are probably several status ints in the drivers that just as well could be char. I'll keep that in mind.

—M

ghaerr:
@Mellvik:

I just compiled ELKS using this PR. The NE2K driver fails compilation with the following error (6 times):
```
ne2k.c:31:2: warning: braces around scalar initializer
  { 0, 0, 0, 0, 0, {0x52, 0x54, 0x00, 0x12, 0x34, 0x57}};  /* QEMU default  + 1 */
  ^
ne2k.c:31:2: note: (near initialization for ‘netif_stat.oflow_keep’)
```
It appears the PR was submitted without testing.

The reason for the failure is that when the usecount variable was previously added to the netif_stat struct, the extra element in el3.c was added, but not in ne2k.c, which broke NE2K. Now, with this PR, you've deleted an entry in ne2k.c, when you hadn't actually added one, thus breaking it again.

This whole thing is a good exercise in showing the problems caused with using external structures, with internal data structures depending on them. In the future, when any external struct changes, the programmer must grep the entire source tree for changes, and then test each of those changes with configuration builds; otherwise, this kind of thing can go unnoticed.

Please submit another PR ASAP.

Thank you!

ghaerr:
> There are probably several status ints in the drivers that just as well could be char. 

Yes - and thought as to whether they should be signed or unsigned is important. In general, status flags should always be unsigned. In the usecount case, (borrowed from the TTY driver in ntty.c), careful thought went in to allowing the count to be unsigned char, thus allowing 2x the open count for the same bit-width. (Not that ELKS will ever have 128 opens, but... ).

Another gotcha is when using a byte-sized value, but using "char" instead of "unsigned char". When/if passed to functions, the value will always be sign-extended if char, which may not be what is wanted.

Mellvik:
Sh.. - turns out I was testing with menuconfig set to wd8003. Mea culpa.

fix Coming right up.

—M

ghaerr:
> turns out I was testing with menuconfig set to wd8003.

What I do to avoid these issues and ensure correctness is to have various copies of .config files, named config.wd, config.el3, config.xms, etc. I then "cp config.wd .config; make kclean" and then run with a prebuilt version of qemu.sh I call qs, which allows for easy editing of many startup variations. The make kclean performs a kernel-only clean, which allows for much quicker test turnarounds. I do this before (most) every PR.

For quickly recursive grepping, which I use constantly, I have `alias gr='grep -r -I --exclude-dir cross'` in my .bashrc, and then use `gr netif_stat .` to quickly search from the main directory. There's no way I could keep up without having the `gr` alias.

In addition, I have `alias vc='vi ~greg/.bashrc; source ~greg/.bashrc'` which allows me to quickly edit .bashrc and add new aliases, for which I create new ones for every quick directory I need to get to. Here's a small list of the almost 100 aliases I use:
```
alias ee='cd ~greg/net/elks-gh; . env.sh'
alias ea='cd ~greg/net/elks-gh/elks/arch/i86'
alias eabb='cd ~greg/net/elks-gh/elks/arch/i86/boot'
alias eak='cd ~greg/net/elks-gh/elks/arch/i86/kernel'
alias eal='cd ~greg/net/elks-gh/elks/arch/i86/lib'
alias eam='cd ~greg/net/elks-gh/elks/arch/i86/mm'
alias eab='cd ~greg/net/elks-gh/elks/arch/i86/drivers/block'
alias eac='cd ~greg/net/elks-gh/elks/arch/i86/drivers/char'
alias ean='cd ~greg/net/elks-gh/elks/arch/i86/drivers/net'
alias e='cd ~greg/net/elks-gh/elks'
alias ek='cd ~greg/net/elks-gh/elks/kernel'
alias ei='cd ~greg/net/elks-gh/elks/init'
alias el='cd ~greg/net/elks-gh/elks/lib'
```
I do this for every GitHub project, and it makes getting around so much easier.

Mellvik:
Thank you @ghaerr, much appreciated. 
That's a load of inspiration right there, the gr alias is particularly golden (and got an offspring already, 'gh' which searches history - I use history a lot ).

Your inspiration sent me down a road I've been planing to take for a while: Taking advantage of 'oh my zsh' - zsh add-ons with (among zillions of other things) some really helpful git aliases/functions and not the least - getting the current git branch into the shell prompt - which would have saved me yesterday. 

Needless to say, I caved in to Apple's push towards zsh - and haven't regretted, even though the history handling has important differences which took some time to get used to.

And a piece of good news: While doing the above, I decided to attempt an update and build of ELKS on my desktop mac running Monterey, with a recently updated copy of Apple Xcode tools. Voila, it actually compiles flawlessly. So development can move back from the Mojave machine I've been using for more than a year pending the toolchain to be 'compileable'.
QEMU barfs at the port mappings, requiring root privilege, I'll have to look at that. 

Thanks again.

-M

PR #1332, opened 2022-06-18
[net, ne2k] Fixed botch in netstat struct initialization.

Mellvik:
Quick fix. Being in a rush is not good.


ghaerr:
Please add the "int usecount = 0" in this driver as well to your list as well, thanks.

Mellvik:
Ok, now I'm confused - and admittedly on a tightrope between social activities. Didn't you just recommend using unsigned char and auto initialized bss?

-M

ghaerr:
> Please add the "int usecount = 0" in this driver as well to your list as well
> Ok, now I'm confused

Oops - sorry. I meant "Please change the "static int usecount = 0" to "static unsigned char usecount" in both drivers. My bad!

Issue #1325, opened 2022-06-15
NEC PC-98 enhancements to ELKS


tyama501:
Continued conversation from https://github.com/jbruchon/elks/issues/1238

tyama501:
Hello @ghaerr ,

I have done the first entry in my to-do list on https://github.com/jbruchon/elks/issues/1312#issuecomment-1152870664
and now I have been working the second one.

I modified bioshd as follows, finally I quit using CONFIG_BLK_DEV_BFD_HARD.

https://github.com/tyama501/elks/compare/53171a5...f246f55

Detecting 2 FD drives.
![detect_2fd](https://user-images.githubusercontent.com/61556504/173896632-a4e3b81c-0533-463e-be4f-ddb367d386d2.png)

Mounting 1440/1232 at same time still have some trouble.
I added swich_drive98() and successfully probe floppy and switch device address and sector_size but FAT says 616k and cannot see files. How should I fix this?

Using FD1440 image and mount 1440/1232. 
![mount_1440_and_1232](https://user-images.githubusercontent.com/61556504/173896842-7e914304-b036-48e6-8ad8-b4d29f10a74b.png)

Thank you.

ghaerr:
Hello @tyama501,

> I modified bioshd as follows, finally I quit using CONFIG_BLK_DEV_BFD_HARD.

Nice. I will take a look at this later today when I have some time.

> I added swich_drive98() and successfully probe floppy and switch device address and sector_size but FAT says 616k and cannot see files.

Hmmm... even though you are saying that you have changed the sector_size in the bioshd.c table, it seems that it is still using 512 byte sector size, as 616K is shown, rather than 1232K. Of course, that will be the reason you can't see files, as the block size somehow is still wrong.

I would put in some printk's in various areas to ensure sector_size is in fact 1024, rather than what seems to be 512.

Also, perhaps mount 1232K drive first to see if that changes the results.

Finally, grep for CONFIG_IMG_FD1232 throughout the source and make sure that you can dynamically handle the ifdefs, which will likely have to all be removed (except for boot code), to handle dynamic switching between FD types.

Thank you!

tyama501:
It seems CONFIG_VAR_SECTOR_SIZE is only defined for CONFIG_IMG_FD1232.

elks/include/linuxmt/config.h
#ifdef CONFIG_IMG_FD1232                /* FD1232 uses 1K byte sectors */
[...]

PR #1333, opened 2022-06-18
Pc98 bioshd2

tyama501:
Hello @ghaerr 

As I wrote in https://github.com/jbruchon/elks/issues/1325
this enhancement includes the following.

- Detect numbers of FD Drives.
- Probe 1440/1232 and support both

I installed the kernel to PC-9801BX hard disk and
booted from it and confirmed I could mnt/read/write both 1440/1232 disk.

Thank you!


ghaerr:
Hello @tyama501,

Thanks for the continued enhancements. I'm glad you were able to figure out the kernel problem which resulted in the need for CONFIG_VAR_SECTOR_SIZE being always for PC-98, I hadn't thought of that.

I notice reading through bioshd.c that its starting to get complicated with the IBM vs PC-98 arch differences using ifdefs; at some point we'll probably want to implement function wrappers around a lot of that, to simplify understandability and maintenance. @cocus is asking for multiple partition support on his /dev/ssd flash block device driver and we'll probably need to pull those portions out of bioshd.c and into a shared file at some point also; I haven't really started looking at that.

That said, this is about the end of your wishlist which require changes in the BIOS block device driver, correct, or is there more?

Thank you!

tyama501:
Hello @ghaerr ,

〉starting to get complicated with the IBM vs PC-98 arch differences using ifdefs;

I agree, I also think so.

〉this is about the end of your wishlist which require changes in the BIOS block device driver, correct, or is there more?

For  now, yes it is the end of the list.
I was asked to support 2DD 640KB which is the half of the 1232 on the twitter but I think it is beyond the scope of the 0.7.0.
Thank you.

ghaerr:
> I was asked to support 2DD 640KB which is the half of the 1232 

FD1232 is CHS 77,2,8 with 1024 byte sectors, right? What is FD640, 77,2,8 w/512 byte sectors? That would be very easy to add, just another section in image/Make.defs, I think.

tyama501:
Hello @ghaerr ,

>What is FD640, 77,2,8 w/512 byte sectors?

It seems 80,2,8 w/512 byte sector.
The device address is different (again) so there is some work to do.

PR #1341, opened 2022-06-21
[kernel] Finalize multi-NIC network support

ghaerr:
Completes work on allowing multiple NIC drivers to be compiled into ELKS kernel.

By default, ne2k, wd8003 and 3c509 drivers are now compiled into the kernel. `make menuconfig` allows each NIC to be compiled in individually as desired.
Select NIC to use with `net start nic` or `net=nic` in bootopts (using a nic name from above).

Fiddled around with the init code a bit to have probe routine return NIC found or not for subsequent open. For NE2K driver, no change because of possible QEMU funniness, but for WD8003 and 3C509, probe will judge NIC not found if MAC address is all 0xFF.

Changed the boot printk to display a bit more uniformly.

@Mellvik, this should be complete. Please go ahead with testing and enhancements, thanks!


ghaerr:
Boot screen showing three NICs:
Screen Shot 2022-06-20 at 7 02 31 PM

Mellvik:
Thank you @ghaerr - I'll dive right into it, looks great.
And I have a new challenge for you - came to me overnight :-)

It may be an idea to generalize the 'emuneration' of the NICs into (e.g.) eth0, eth1 etc. Here's the thing:
IN a few days I will have two wd interfaces in the machine sharing the same driver, but using it differently (just like I ran the 8bit and 16bit NE2ks alongside each other a month back).
Not a normal situation, but then again what we're trying to accommodate is development settings. And down the line, if we move ahead with multihoming, 2 interfaces from the same family would be a normal occurrence.
What do you think?

> 21. jun. 2022 kl. 03:11 skrev Gregory Haerr ***@***.***>:
> 
> Completes work on allowing multiple NIC drivers to be compiled into ELKS kernel.
> 
> By default, ne2k, wd8003 and 3c509 drivers are now compiled into the kernel. make menuconfig allows each NIC to be compiled in individually as desired.
> Select NIC to use with net start nic or net=nic in bootopts (using a nic name from above).
> 
> Fiddled around with the init code a bit to have probe routine return NIC found or not for subsequent open. For NE2K driver, no change because of possible QEMU funniness, but for WD8003 and 3C509, probe will judge NIC not found if MAC address is all 0xFF.
> 
This is interesting, I need to look at that. Both the ne2k and the el3 drivers had reliable and tested probing before, so I must have missed something.


Practical experiences coming up. Unfortunately the compaq expansion box has only two slots …

-M

ghaerr:
> It may be an idea to generalize the 'emuneration' of the NICs into (e.g.) eth0, eth1 

No, that won't work with the current design. The intent for 0.7.0 was only to enable the compiled inclusion of all NIC drivers into the kernel so that networking could work by just setting NIC parameters in /bootopts, rather than recompiling. Hopefully this will increase the usage of ELKS by providing working binaries for new users. I don't think we'll be getting to multi-homing anytime soon.

Mellvik:
The point wasn't multihoming, but development.
The current setup doesn't allow for two interfaces of the same type which my next driver project would benefit from (wd8003 and wd8013). 

Minor thing though - by now I have plenty of other interfaces to choose from.

—M

ghaerr:
Ah, I see now. Well, ideas welcomed for how two NICs of the same type could be made to work, as you get familiar with how this new version works. 

The current design also needs clarification on CONFIG_BYTE_ETH_ACCESS (does it apply to all NICs or should we drop it), as well as whether we want to keep the two lower net_port bits to describe the card (8/16 bit & 4k buffer, or another setting for explicit buffer size) or use another way of setting in /booptopts. It seems we might be entering an era of "parameter hell" in order to get networking running with a given NIC variation.

My biggest issue is how to make it easy for the user to setup a NIC - although I'm not sure how easy that can be. Both myself and @toncho11 have yet to get a NIC working. What do you think about the idea of an application program that interrogates the NIC, or the system itself, that might allow for an easier setup or system identification?

Mellvik:
> 
> Ah, I see now. Well, ideas welcomed for how two NICs of the same type could be made to work, as you get familiar with how this new version works.
> 
I'll chew on that one while testing the new eth regime.
> The current design also needs clarification on CONFIG_BYTE_ETH_ACCESS (does it apply to all NICs or should we drop it), as well as whether we want to keep the two lower net_port bits to describe the card (8/16 bit & 4k buffer, or another setting for explicit buffer size) or use another way of setting in /booptopts. It seems we might be entering an era of "parameter hell" in order to get networking running with a given NIC variation.
> 
Yes, I was thinking along the same lines when running menuconfig this morning. I'm still waiting for the wd8013 to arrive, when it shows up I'll make a point of finding a way to auto configure between the two.

The challenge not attacked yet is detecting if a 16 bit NIC is in a 8bit slot. It should be easy with the memory mapped interfaces, less so with the PIO (NE2k) variants. Will test out some ideas in that regard asap.

Regardless, I think we should
1) remove the 8Bit entry in menuconfig
2) add more about ethernet and 8vs16 bit in the docs (I already did some but it's probably not enough)
3) accept that new users run into trouble and ask. They are few end far between, and eventually we'll fix it.
> My biggest issue is how to make it easy for the user to setup a NIC - although I'm not sure how easy that can be. Both myself and @toncho11 <https://github.com/toncho11> have yet to get a NIC working. What do you think about the idea of an application program that interrogates the NIC, or the system itself, that might allow for an easier setup or system identification?
> 
> 
Several times over the past 3 years I've taken a dive into the Linux ne2k-diag.c program, which does all kinds of tricks and tests and settings with many 8309 based NICs. It's complicated and even if we could get something like that running (we could), it would not help users with other interfaces.

It's easy to forget how far we've actually come in terms of ease with the current bootopts regime. New users (like old ones) will have to have some clue about IRQ and port addresses (and settings) anyway. What I think is the major obstacle for many is figuring out what the current card config is and how to change it. BTW - I wasn't aware of this until I got the WE8003 card this weekend: The shared memory address on these cards (at least this generation) is fixed. 0xca00, that's it. Newer cards may be more flexible.

I think it would be helpful if we had a IRQ detection mechanism, a kernel entry for the driver to call just before initiating an interrupt from the NIC, the a 2nd call after the presumed interrupt to check what or if something happened. IIRC this is what Linux is doing (or was doing in the ISA days).
 
Then next thing may be to support PnP (The pre-PnP cards have jumpers or switches so the config can be read and set physically). maybe a PnP dump & config utility is what we should be looking for - like isapnptools for Linux of around 2000. I'm not familiar with it, but it may be worth a shot.

—M

ghaerr:
> I'll make a point of finding a way to auto configure between the two.

Good idea, the more automated the better with self-configuration.

> remove the 8Bit entry in menuconfig

Ok. I also plan on removing CONFIG_CPU_USAGE (leave always on), and CONFIG_INODE_32BITS (FAT basically requires it), and am also considering removing the CPU ID detection code in setup.S, which is both ugly and incorrect. 

> add more about ethernet and 8vs16 bit in the docs

I updated the Wiki on the latest design change, so that's done. Perhaps a "Detailed Networking Setup" section or something for the 8/16 bit stuff and buffer sizes. Isn't there a way to autodetect both of those?

> What I think is the major obstacle for many is figuring out what the current card config is and how to change it

With the default kernel now shipping with all NIC drivers present, just looking at the boot display may help...

> if we had a IRQ detection mechanism, a kernel entry for the driver to call just before initiating an interrupt from the NIC, the a 2nd call after the presumed interrupt to check what or if something happened.

Not following you on this. Is this for development purposes, or for a user to determine whether an IRQ is available/working?

Mellvik:
Early testing looks excellent - just one minor snag so far: /dev/3c509 has ended up as /dev/ec509.

(output from two different systems)

Thanks, @ghaerr. This is easing net driver development tremendously in addition to presumably making life easier for users.


```
ttyS0 at 0x3f8, irq 4 is a 16450
xms: A20 was off now on, using int 15/1F, 1024 xms buffers, 1048576 ram
eth: ne2k (16 bit) at 0x300, irq 12: MAC 00:80:29:ef:d9:51
eth: wd8003 at 0x340, irq 9, ram 0xce00: MAC 00:00:C0:4D:F6:13
eth: 3c509 at 0x330, irq 9: not found
bioshd: hda BIOS CHS 610,16,63
/dev/hda: 993 cylinders, 16 heads, 63 sectors = 488.7 Mb
/dev/fd0: 80 cylinders, 2 heads, 18 sectors = 1440.0 kb
/dev/fd1: 80 cylinders, 2 heads, 15 sectors = 1200.0 kb
Partitions: hda:(0,1000944)  hda1:(63,608769)  hda2:(608832,132048)  hda3:(74088
0,132048)  hda4:(872928,128016) 
PC/AT class machine, Unknown x86 CPU, 640K base RAM.
ELKS kernel 0.6.0 (57936 text, 12112 ftext, 8352 data, 42464 bss, 14718 heap)
Kernel text at d0:0000, ftext ef5:0000, data 11ea:0000, top a000:0, 504K free
VFS: Mounted root 0x0304 (minix filesystem).
Running /etc/rc.sys script
fsck -ra /dev/hda4: /dev/hda4 is clean, no check.
FAT: me=f8,csz=16,#f=2,floc=1,fsz=149,rloc=299,#d=512,dloc=331,#s=608769,ts=6087
69
FAT: total 304M, fat16 format
fsck -ra /dev/hda2: 
Filesystem on /dev/hda2 is dirty, needs checking.
fsck -ra /dev/hda4: 
Filesystem on /dev/hda4 is dirty, needs checking.
mount failed: Device or resource busy 
fsck -ra /dev/fd0: 
fd: /dev/fd0 ELKS bootable, has 80 cylinders, 2 heads, and 18 sectors
/dev/fd0 is clean, no check.
Starting networking on ne2k
ktcp -b -p ne2k 10.0.2.15 10.0.2.2 255.255.255.0
ktcp: /dev/ne2k mac 00.80.29.ef.d9.51 mtu 1500
Starting daemons 'telnetd' 'ftpd -d' 
Tue Jun 21 14:11:45 2022

login: root
elks15# telnet 10.0.2.2
Connecting to 10.0.2.2 (10.0.2.2:23)
Connected
Raspbian GNU/Linux 9
raspi2b login: 

Login incorrect
raspi2b login: 
Login timed out after 60 seconds.

Connection closed
elks15# net stop
Stopping network
elks15# net start wd8003
Starting networking on wd8003
ktcp -b -p wd8003 10.0.2.15 10.0.2.2 255.255.255.0
ktcp: /dev/wd8003 mac 00.00.c0.4d.f6.13 mtu 1500
Starting daemons 'telnetd' 'ftpd -d' 
elks15# telnet 10.0.2.2
Connecting to 10.0.2.2 (10.0.2.2:23)
Connected
Raspbian GNU/Linux 9
raspi2b login: 
Login timed out after 60 seconds.

Connection closed
elks15# 
```

Second sydd:
```
ttyS0 at 0x3f8, irq 4 is a 16450
xms: A20 was off now on, using int 15/1F, 1024 xms buffers, 1048576 ram
eth: ne2k (8 bit) at 0x300, irq 9: MAC 00:1f:11:02:60:2d
eth: wd8003 at 0x240, irq 2, ram 0xce00: not found
eth: 3c509 at 0x330, irq 12: MAC 00:a0:24:67:7f:b1 (HW conf: irq 9 port 330)
bioshd: hda BIOS CHS 523,4,40
/dev/hda: 523 cylinders, 4 heads, 40 sectors = 40.9 Mb
/dev/fd0: 80 cylinders, 2 heads, 18 sectors = 1440.0 kb
Partitions: hda:(0,83680)  hda1:(40,83640) 
device_setup: BIOS drive 0x0, root device 0x380
PC/AT class machine, Unknown x86 CPU, 640K base RAM.
ELKS kernel 0.6.0 (57936 text, 12112 ftext, 8352 data, 42464 bss, 14718 heap)
Kernel text at d0:0000, ftext ef5:0000, data 11ea:0000, top a000:0, 504K free
fd: /dev/fd0 ELKS bootable, has 80 cylinders, 2 heads, and 18 sectors
MINIX-fs: mounting unchecked file system 0x380, running fsck is recommended.
Running /etc/rc.sys script
Starting networking on ne2k
ktcp -b -p ne2k 10.0.2.16 10.0.2.2 255.255.255.0
ktcp: ip 10.0.2.16, gateway 10.0.2.2, netmask 255.255.255.0
ktcp: /dev/ne2k mac 00.1f.11.02.60.2d mtu 1500
Starting daemons 'telnetd' 'ftpd -d' 
Sun Jun 12 22:41:56 2022


ELKS 0.6.0

login: root
elks16# telnet 10.0.2.2
Connecting to 10.0.2.2 (10.0.2.2:23)
Connected
Raspbian GNU/Linux 9
raspi2b login: ^C
Connection closed
elks16# net stop
Stopping network
elks16# net start 3c509
Starting networking on 3c509
ktcp -b -p 3c509 10.0.2.16 10.0.2.2 255.255.255.0
ktcp: ip 10.0.2.16, gateway 10.0.2.2, netmask 255.255.255.0
ktcp: /dev/3c509 mac 00.a0.24.67.7f.b1 mtu 1500
Starting daemons 'telnetd' 'ftpd -d' 
elks16# telnet 10.0.2.2
Connecting to 10.0.2.2 (10.0.2.2:23)
Connected
Raspbian GNU/Linux 9
raspi2b login: ^C
Connection closed
elks16# 

```

ghaerr:
> just one minor snag so far: /dev/3c509 has ended up as /dev/ec509.

Oops, lol! Admittedly I have been typing quite quickly the last few days... and wasn't able to test 3c509 on QEMU. I'll fix it.

> This is easing net driver development tremendously in addition to presumably making life easier for users

Nice, thanks for the testing! Pretty cool isn't it!

Keep an eye out for the ability of the boot init_probe routines to do their job, and print "not found" when the NIC isn't present. I modified both WD8003 and 3C509 for that, but didn't change NE2K, since you have some special stuff in there for QEMU. (One of) the current tests is if the MAC address is reported as all 0xFF, the probe fails, which prevents the net start from succeeding. I'm sure you'll have your own opinions as you test more with various NICs.

> eth: 3c509 at 0x330, irq 12: MAC 00:a0:24:67:7f:b1 (HW conf: irq 9 port 330)

I standardized some of the probe output messages, but see I missed this one - what does "HW conf" mean, IRQ 9 vs operating on 12? Also, port 330 is reported twice, should that be removed? What I'm looking for is the ability to do a quick scan at boot time to see the NIC status, having the init message look very similar and omitting information that isn't needed.

Thanks for sending over the TTY logs... ELKS is really becoming easier to use!

Mellvik:
> 
> What I think is the major obstacle for many is figuring out what the current card config is and how to change it
> 
> With the default kernel now shipping with all NIC drivers present, just looking at the boot display may help…
> 
Well it doesn't. It just tells what the ELKS config is, not what the card config is.
> if we had a IRQ detection mechanism, a kernel entry for the driver to call just before initiating an interrupt from the NIC, the a 2nd call after the presumed interrupt to check what or if something happened.
> 
> Not following you on this. Is this for development purposes, or for a user to determine whether an IRQ is available/working?
> 
This for a driver to detect which interrupt a card is configured to use by creating an interrupt and asking the system where it came. As mentioned before, to the user - knowing how the card is configured (jumperless, some times even if it is jumper-configured) is really hard. IRQ detection helps tremendously, possibly also in detecting  irq-collissions.

—M

Mellvik:
> 
Thank you @ghaerr,
> Keep an eye out for the ability of the boot init_probe routines to do their job, and print "not found" when the NIC isn't present. I modified both WD8003 and 3C509 for that, but didn't change NE2K, since you have some special stuff in there for QEMU.
> 
Yes. I didn't notice any changes in the 3c509 probe (except how it's displayed) and it's fine. The ne2k probe is more like brute force and may report false positives, but only if a different interface responds exactly like the ne2k to a certain command at a certain address. The QEMU detection is separate from card detection and uses address reads only.
> (One of) the current tests is if the MAC address is reported as all 0xFF, the probe fails, which prevents the net start from succeeding. I'm sure you'll have your own opinions as you test more with various NICs.
> 
I do, and the ne2k doesn't use address detection for probing. If it doesn't find a valid address, it uses the default - copied from QEMU. For the wd driver, it seems the address test is OK. I'll take a look and see what Linux does. Theoretically there may be cases where the address prom is missing but the card is still OK. If so, a default address like in the ne2k driver may be appropriate. Not a very likely situation though.
> eth: 3c509 at 0x330, irq 12: MAC 00:a0:24:67:7f:b1 (HW conf: irq 9 port 330)
> 
> I standardized some of the probe output messages, but see I missed this one - what does "HW conf" mean, IRQ 9 vs operating on 12? Also, port 330 is reported twice, should that be removed?
> 
I added the HW conf output to show how the card is configured. Since the EL3 is completely soft-configured and uses whatever is in the bootopts file, the actual HW config is unknown. Whether it's useful is a different story. To me it was - during development. And I gather if someone is experimenting to get ELKS going, it may be interesting too.
> What I'm looking for is the ability to do a quick scan at boot time to see the NIC status, having the init message look very similar and omitting information that isn't needed.
> 
Agreed, and if we keep the HW conf output, it should be rearranged.

Apropos configuration - I think it is a good idea to keep reading the port and address config values from bootopts as hex even if the 0x prefix is omitted - like we had before.


This sprint towards 0.7.0 has been incredible. The distance (in time) between 0.5 and 0.6 was a new record for ELKS. I think we're heading towards another one :-)


—M

ghaerr:
> it is a good idea to keep reading the port and address config values from bootopts as hex even if the 0x prefix is omitted - like we had before.

I'll add that (back) in the next PR coming shortly.

Mellvik:
@ghaerr - as practical experiences tick in, a couple of observations/ideas about NIC configuration in /bootopts.

You asked before about moving options away from the port address encoding scheme we currently use (really intended for debugging, not general use): I think we should go a head and do this, again the Linux way (IIRC). Add an optional  (last) parameter (a hex 'flag' field) to the NIC config line which sets flags for that particular driver. Such as 8bit mode, small buffer, preferred physical interface, overflow strategy  ...

Also, as it turns out, we need a shared_mem_end setting for the WD interface. Their buffer size vary between 8k, 16k and 32k bytes. The former two typically distinguished by whether the interface is 8 or 16 bits, the latter included on some 16bit interfaces and hard or impossible to autodetect. Also, we may have a 16 bit interface in an 8 bit slot and need to manually force the NIC to use 16k buffer.

Something like `wd8003=9,0x340,0xce00,0xd000,3` or `wd8003=9,0x340,0xce00,3` where the shared_mem_end parameter could be recognized by its length.

Eventually, if we decide to go down that road, the config line could possibly read
```
eth0=wd,9,0x340,0xce00,0xd000,5
eth1=ne2k,12,0x300,3
eth2=ne2k,10,0x330,1
```
I'm using 'wd' instead of 'wd8003' to raise a different question. The `wd` driver will soon support both the wd8003 and the wd8013, so how to avoid confusion, or rather - how to make it clear that both are supported? Use `wd` for simplicity or `wd80x3` for generality?

-M

Mellvik:
Additional comments to the discussion above: 

It is quite possible, maybe advantageous,  to use the `flag` to set the size of the shared memory. Such as 'bit 4: force 32k', 'bit 3: force 16k' etc. FreeBSD uses that (and has kept the traditional BSD config style (ours is more compact and suitable)):

`we2 at isa? port 0x300 iomem 0xe0000 irq 15 flags 4`

-M

ghaerr:
Hello @Mellvik,

> It is quite possible, maybe advantageous, to use the flag to set the size of the shared memory.

I like that; less numbers on the options line.

Trying to keep the kernel bootopts code small, the ne2k/wd8003/3c509= options are all parsed the same. That is, the new method would be ne2k=9,0x300,,1 to specify flags=1, and skip the ram address. If an option is not specified, it will default to 0. The WD driver would be wd8003=12,0x240,0ce00,4 to set flags=4.

How does that look? I could add that very quickly, which would result in a new "net_flags" value to be interpreted by the driver. (This value would actually reside as another structure member inside the netif_parms[] array).

While we're in the design stage, can you list the proposed values for the flag word for each driver? We should reserve bits now that make sense into the future, for variations you plan on supporting. Flags = 0 should be reserved to allow any detection to occur automatically, which will be the default. Once a flag word is non-zero, that would force each bit in the flag word. What that means, is perhaps "force 8 bit" and "force 16 bit" should have their own bit in the flag word. (Otherwise, forcing a ram size for instance might also force an 16 bit mode, if 16-bit mode was set with a "0" in its bit position).

> The wd driver will soon support both the wd8003 and the wd8013, so how to avoid confusion, or rather - how to make it clear that both are supported?

If changed, I believe it would be simpler to just specify wd=. Lets leave all of that alone for now, until we see what happens with 3c509 and other NICs so we don't change names multiple times. Given the complexity of getting a NIC up and running, frankly I think the name itself is the least of the users problems...

> Eventually, if we decide to go down that road, the config line could possibly read

```
eth0=wd,...
eth1=ne2k,...
...
```
This really doesn't work for the current design, but its a nice idea for when we get to that point.

Thank you!

Mellvik:
@ghaerr,
> It is quite possible, maybe advantageous, to use the flag to set the size of the shared memory.
> 
> I like that; less numbers on the options line.
> 
> Trying to keep the kernel bootopts code small, the ne2k/wd8003/3c509= options are all parsed the same. That is, the new method would be ne2k=9,0x300,,1 to specify flags=1, and skip the ram address. If an option is not specified, it will default to 0. The WD driver would be wd8003=12,0x240,0ce00,4 to set flags=4.
> 
> How does that look? I could add that very quickly, which would result in a new "net_flags" value to be interpreted by the driver. (This value would actually reside as another structure member inside the netif_parms[] array).
> 
Looks great- all the way.
> While we're in the design stage, can you list the proposed values for the flag word for each driver? We should reserve bits now that make sense into the future, for variations you plan on supporting. Flags = 0 should be reserved to allow any detection to occur automatically, which will be the default. Once a flag word is non-zero, that would force each bit in the flag word. What that means, is perhaps "force 8 bit" and "force 16 bit" should have their own bit in the flag word. (Otherwise, forcing a ram size for instance might also force an 16 bit mode, if 16-bit mode was set with a "0" in its bit position).
> 
Yes, I do have a list, and I agree as to how to treat the flags. I'll submit the list tomorrow.
> The wd driver will soon support both the wd8003 and the wd8013, so how to avoid confusion, or rather - how to make it clear that both are supported?
> 
> If changed, I believe it would be simpler to just specify wd=. Lets leave all of that alone for now, until we see what happens with 3c509 and other NICs so we don't change names multiple times. Given the complexity of getting a NIC up and running, frankly I think the name itself is the least of the users problems...
> 
Agreed on both accounts. Now it has been aired so we'll discuss it when appropriate.

A new version of wd.c is in the works, so if you move forward before I PR it, please leave it out for now.

Thank you. 
-M

Mellvik:
Config flags for networks interfaces

NETIF_8BIT_BUS Force 8 bit
NETIF_16BIT_BUS Force 16 bit
NETIF_4K_BUF Force 4K NIC (ne2k type NIC)
NETIF_8K_BUF Force 8K NIC (default on WD type memory mapped)
NETIF_16K_BUF
NETIF_32K_BUF (wd type only)
NETIF_NOERRMSG skip verbose console errors, useful on some 8 bit NICs 

We may need to make this word size, there may be more to come.

—M



> 23. jun. 2022 kl. 21:56 skrev Helge Skrivervik ***@***.***>:
> 
> @ghaerr,
>> It is quite possible, maybe advantageous, to use the flag to set the size of the shared memory.
>> 
>> I like that; less numbers on the options line.
>> 
>> Trying to keep the kernel bootopts code small, the ne2k/wd8003/3c509= options are all parsed the same. That is, the new method would be ne2k=9,0x300,,1 to specify flags=1, and skip the ram address. If an option is not specified, it will default to 0. The WD driver would be wd8003=12,0x240,0ce00,4 to set flags=4.
>> 
>> How does that look? I could add that very quickly, which would result in a new "net_flags" value to be interpreted by the driver. (This value would actually reside as another structure member inside the netif_parms[] array).
>> 
> Looks great- all the way.
>> While we're in the design stage, can you list the proposed values for the flag word for each driver? We should reserve bits now that make sense into the future, for variations you plan on supporting. Flags = 0 should be reserved to allow any detection to occur automatically, which will be the default. Once a flag word is non-zero, that would force each bit in the flag word. What that means, is perhaps "force 8 bit" and "force 16 bit" should have their own bit in the flag word. (Otherwise, forcing a ram size for instance might also force an 16 bit mode, if 16-bit mode was set with a "0" in its bit position).
>> 
> Yes, I do have a list, and I agree as to how to treat the flags. I'll submit the list tomorrow.
>> The wd driver will soon support both the wd8003 and the wd8013, so how to avoid confusion, or rather - how to make it clear that both are supported?
>> 
>> If changed, I believe it would be simpler to just specify wd=. Lets leave all of that alone for now, until we see what happens with 3c509 and other NICs so we don't change names multiple times. Given the complexity of getting a NIC up and running, frankly I think the name itself is the least of the users problems...
>> 
> Agreed on both accounts. Now it has been aired so we'll discuss it when appropriate.
> 
> A new version of wd.c is in the works, so if you move forward before I PR it, please leave it out for now.
> 
> Thank you. 
> -M

ghaerr:
> Config flags for networks interfaces

Thanks @Mellvik, looks good. I might suggest rather than NOERRMSG, use VERBOSE; that is, don't emit extra error or packet loss console messages which require the user to know how to turn them off; instead use VERBOSE to turn them on for debugging or those that are interested in such things. (Networking is supposed to just work, right?). When I was heavy working on ktcp last year, we went through quite a bit of printk output changes, eventually reducing to almost none (unless turned via debug_xxx) as the code started working well.

> We may need to make this word size, there may be more to come.

Yep - it's already 16 bits.

> a thought: Now that we're expanding what goes into netstat.h, maybe a better name would be appropriate

Good idea. I'm going to clean up netstat.h, which should just be for interfacing with `netstat`. Will create a new ethernet.h header file in a PR now, and you can then add these defines to it. Lets then rename the above "ETHF_xxx" (ethernet flags) instead of NETIF_. 

In the drivers themselves, after the ethernet.h include change, you're welcome to also rename net_flags to eth_flags (or not), if you think that would look better in the source: "if (eth_flags & ETHF_8_BIT)..." etc.

Mellvik:
Thank you @ghaerr -
> 
> Config flags for networks interfaces
> 
> Thanks @Mellvik <https://github.com/Mellvik>, looks good. I might suggest rather than NOERRMSG, use VERBOSE; that is, don't emit extra error or packet loss console messages which require the user to know how to turn them off; instead use VERBOSE to turn them on for debugging or those that are interested in such things. (Networking is supposed to just work, right?). When I was heavy working on ktcp last year, we went through quite a bit of printk output changes, eventually reducing to almost none (unless turned via debug_xxx) as the code started working well.
> 
Agree, good idea.
> We may need to make this word size, there may be more to come.
> 
> Yep - it's already 16 bits.
> 
> a thought: Now that we're expanding what goes into netstat.h, maybe a better name would be appropriate
> 
> Good idea. I'm going to clean up netstat.h, which should just be for interfacing with netstat. Will create a new ethernet.h header file in a PR now, and you can then add these defines to it. Lets then rename the above "ETHF_xxx" (ethernet flags) instead of NETIF_.
> 
> In the drivers themselves, after the ethernet.h include change, you're welcome to also rename net_flags to eth_flags (or not), if you think that would look better in the source: "if (eth_flags & ETHF_8_BIT)..." etc.
> 
> 
Yes, I like that. I'll attack it when the new header file is in place.

Thank you!

-M

ghaerr:
@Mellvik,

I just looked at netstat.h, and didn't realize that a number of NETIF_ flags are already in there, which won't be part of the flag bits passed in /bootopts above (e.g. NETIF_IS_OPEN, NETIF_FOUND, NETIF_IS_QEMU).

Because of that, instead of rearranging everything now and causing more delays/havoc, I think a better idea would be to let you work on cleaning up the driver set using the new working architecture first, then possible rearrange more header files, if really needed. Splitting this out with the existing NETIF_ flags will just be more confusing to both of us.

What I suggest doing is the following:
- Add ETHF_xxx flags into netstat.h and transition your drivers to the new mechanism, where just the ETHF_xxx bits are passed by /bootopts and used to control the driver
- Remove NETIF_ flags like NETIF_FORCE_4K that we'll be using ETHF_ flags for
- Move NETIF_FOUND and NETIF_IS_OPEN into static variables, or remove altogether. (For instance, NETIF_IS_OPEN is redundant when we're also using "static unsigned char usecount". In general, trying to fit all this stuff into bits into a non-existent transfer mechanism to netstat is causing code bloat and confusion at this time. Delete unused NETIF_ bits, and lets keep things simple as possible.
- Generally cleanup and test the new drivers with the new architecture and newly implemented /bootopts ETHF_ bits.
- Add the net flags (or eth_flags) define into the wd.c driver in your next PR, since I left that out.

Thank you!

Mellvik:
Thank you @ghaerr - 

yes, I realize I should have mentioned that.

> I just looked at netstat.h, and didn't realize that a number of NETIF_ flags are already in there, which won't be part of the flag bits passed in /bootopts above (e.g. NETIF_IS_OPEN, NETIF_FOUND, NETIF_IS_QEMU).

From my perspective these are entirely different species and it makes sense to keep them separate. At least that was my first thought. The new flags are settings for the driver, the 'old' ones are statuses possibly used by the driver and available for reporting back to netstat.

I'm open to discussing this though. Here's the thing: If there are no flags specified for a particular interface, we'd still want to know via IOCTL what we have here, the MAC, 8bit or not, 4k or not (the automatic config values so to speak), Some of which come during boot but the boot messages are long gone by the time you need them.

It is true some of the original NETIF flags are already redundant, so no problem there. And yes - the result must be clean and easy to understand. Now here's my other motivation for having two separate headerfiles:
netstat.h for the interface to netstat (the program). 
ethernet.h for the devices, eventually containing the dev-struct that will be needed to pass around to the drivers when one driver serve several physical interfaces.

Seems we're on the same page, just need to decide what's the cleanest way forward. You have a lot of experience in this.

-M

ghaerr:
> From my perspective these are entirely different species and it makes sense to keep them separate. At least that was my first thought. The new flags are settings for the driver, the 'old' ones are statuses possibly used by the driver and available for reporting back to netstat.

Agreed. Some of the old NETIF_ flags will stay, and you'll use the new ETHF_ flags for the /bootopts configuration. I'm just saying that some of the NETIF_ flags are redundant and should be removed.

> Here's the thing: If there are no flags specified for a particular interface, we'd still want to know via IOCTL what we have here, the MAC, 8bit or not, 4k or not (the automatic config values so to speak), Some of which come during boot but the boot messages are long gone by the time you need them.

Thats great - but we don't even have any of that working yet. (ELKS is filled with unfinished stuff like that, I've been cleaning that up for the last two years now). You're welcome to add any status you like, I'm just trying to synchronize what we actually have working and not have a bunch of stuff added that's not implemented. Just trying to keep things simple - for instance, NETIF_IS_OPEN - well of course that will always be set on an netstat ioctl return since the driver has to be opened in order to do an ioctl. We'd be better off returning usecount. My thoughts are that the actual netstat display (not even designed yet) will actually drive the bits needed, and added at that time.

> Now here's my other motivation for having two separate headerfiles:
> netstat.h for the interface to netstat (the program).
> ethernet.h for the devices

Agreed, we can do that after getting the above done.

> just need to decide what's the cleanest way forward. You have a lot of experience in this.

We also have tons of header files in ELKS. At the moment, almost all of netstat.h isn't even implemented. I don't want to create another one until we're through this. (BTW, now that all the NIC drivers are compiled into the kernel, my concerns about code bloat and excessive printk's in the drivers is growing - we only have ~8k left of kernel code space, unless we move to FARPROCs, which brings its own set of slowdowns and uses even more code, albeit in a seperate segment. Functions called directly from a "file_operations" struct (which is most entry points to the NIC driver) aren't eligible for FARPROC).

For ELKS - "Size matters!!"

Mellvik:
Thank you @ghaerr - 

> 
> From my perspective these are entirely different species and it makes sense to keep them separate. At least that was my first thought. The new flags are settings for the driver, the 'old' ones are statuses possibly used by the driver and available for reporting back to netstat.
> 
> Agreed. Some of the old NETIF_ flags will stay, and you'll use the new ETHF_ flags for the /bootopts configuration. I'm just saying that some of the NETIF_ flags are redundant and should be removed.
> 
Of course. They were made redundant just a few days ago, we'll get there.
> Here's the thing: If there are no flags specified for a particular interface, we'd still want to know via IOCTL what we have here, the MAC, 8bit or not, 4k or not (the automatic config values so to speak), Some of which come during boot but the boot messages are long gone by the time you need them.
> 
> Thats great - but we don't even have any of that working yet.
> 
Again, this is work in progress. We cannot have it working till we have the pieces in place...

> (BTW, now that all the NIC drivers are compiled into the kernel, my concerns about code bloat and excessive printk's in the drivers is growing - we only have ~8k left of kernel code space, unless we move to FARPROCs, which brings its own set of slowdowns and uses even more code, albeit in a seperate segment. Functions called directly from a "file_operations" struct (which is most entry points to the NIC driver) aren't eligible for FARPROC).
> 
> For ELKS - "Size matters!!"
> 
Indeed. Let's keep in mind that the motivation for this enhancements was to ease development. Keeping all driver in the kernel is a convenience derived from that ability. It may not be a desirable default. That said, there are things we can to ease the memory footprint, such as collecting all states message strings into an enumerated list in eth.c (or somewhere else) and call (e.g.) eth_msg(IntRxErr, arg1, …), eth_msg(IntOflowErr, arg1) etc. from all drivers. After all, most of the messages are or could be made the same. The drivers could also easily share the select(), write() and ioctl() functions if we wanted it. 

Let's start off with what we have though, as you suggest, get an enhanced netstat up and running and take it from there. 

—M

ghaerr:
> The drivers could also easily share the select(), write() and ioctl() functions if we wanted it.

Neat idea, I now remember the NIC "subdriver template" thoughts we had some time back. This may be restrictive at first during development, but having the upper level of the NIC drivers (and strings with your idea below) the same would be nice for size and other reasons.

> collecting all states message strings into an enumerated list
> most of the messages are or could be made the same

Also good ideas. The linker may be able to be told to combine strings automatically with an option, or with some gcc attribute tags.

Keep these ideas in mind as you lead the way... and we'll see how we might combine things as the drivers continue to be enhanced with the planned netstat and bootopts features.

ghaerr:
> in eth.c (or somewhere else)

Yes - since eth.c is in the drivers/char directory, perhaps just adding a new eth-shared.c file in drivers/net with any shared function calls, string declarations or shared global variables would be simplest, and include that file in drivers/net/Makefile when CONFIG_ETH is defined. This could also work for any upper level driver functions including ioctl. Placing ioctl in there would help force some degree of similarity between the drivers, and save a lot of future code space.

Mellvik:
>>collecting all states message strings into an enumerated list
>>most of the messages are or could be made the same

>Also good ideas. The linker may be able to be told to combine strings automatically with an option, or with some gcc attribute tags.

@ghaerr,
can you elaborate on the above? How do you see this work?
I'm getting close to submitting an ethernet update and I may have missed something here.

--M

ghaerr:
> The linker may be able to be told to combine strings automatically 
> can you elaborate on the above? How do you see this work?

If any literal strings in your driver(s) are *exactly* the same, the linker will combine the strings automatically into the same space in the output file (kernel); this is called "string merging" and is a linker feature. You don't have to do anything.

Mellvik:
Thanks @ghaerr - 

that's very useful. Actually, keeping the printks in place and standardizing the strings is more readable than a shared eth_pmsg routine with standardized texts.
I will take advantage of that.

-M

PR #1349, opened 2022-06-23
[net] Add net_flags as fourth optional parm to ne2k=/wd=/3c509= NIC opts

ghaerr:
As discussed in https://github.com/jbruchon/elks/pull/1341#issuecomment-1164672837.

Adds net_flags to ne2k.c and el3.c. As discussed, net_flags (and net_ram) will be 0 if not specified.

@Mellvik, please add the following to your work-in-process wd.c, just after the `net_ram` define:
```
#define net_ram     (netif_parms[1].ram)
#define net_flags   (netif_parms[1].flags)
```

This PR also fixes the buggy `strchr` I mistakenly broke in my last PR (#1348). That has only been merged for a few hours so hopefully that didn't affect anyone.


Mellvik:
Thank you @ghaerr,

a thought: Now that we're expanding what goes into netstat.h, maybe a better name would be appropriate,
such as netdev.h ??

—M

Issue #2, opened 2014-09-06
Elkscmd build fail


hellocyf:
I have build elks and make boot image.
I want to make image with root system,so goto Elkscmd 
Make images.zip,this makes error .

:~/elks/elkscmd$ sudo make images.zip
cp /home/xxx/elks/elks/arch/i86/boot/Image boot
make[1]: Entering directory `/home/xxx/elks/elkscmd'
/bin/sh: 1: [[: not found
/bin/sh: 1: -e: not found
-e

 **\* Do builds in dev86 and dev86/bootblocks first ***

make[1]: **\* [boot_blocks] Error 1
make[1]: Leaving directory `/home/xxx/elks/elkscmd'
make: **\* [comb] Error 2

hellocyf:
rebuild dev86

make[3]: `bcc' is up to date.
make[3]:`ncc' is up to date.
make[3]: `bcc-cc1' is up to date.
make[3]: Leaving directory`~/elks/dev86/bcc'
cp -p bcc/bcc bin/Bcc
cp -p bcc/ncc bin/ncc
cp -p bcc/bcc-cc1 lib/bcc-cc1
make -C cpp CC='cc' CFLAGS=' -O2 -fno-strict-aliasing' LDFLAGS='' PREFIX=/usr LIBDIR='/usr/lib/bcc' BINDIR='/usr/bin' ANSI='' bcc-cpp
make[3]: Entering directory `~/elks/dev86/cpp'
gperf -aptTc -N is_ctok -H hash1 token1.tok > tmp.h
token1.tok:10: declarations are not allowed in the keywords section.
To declare a keyword starting with %, enclose it in double-quotes.
make[3]: *** [token1.h] Error 1
make[3]: Leaving directory`~/elks/dev86/cpp'
make[2]: **\* [cpp] Error 2
make[2]: Leaving directory `~/elks/dev86'
make[1]: *** [all] Error 2
make[1]: Leaving directory`~/elks/dev86'
make: **\* [all] Error 2

token1.tok
struct token_trans { char \* name; int token; };
%%
...,    TK_ELLIPSIS

> > =,    TK_RIGHT_ASSIGN
> > <<=,    TK_LEFT_ASSIGN
> > +=, TK_ADD_ASSIGN
> > -=, TK_SUB_ASSIGN
> > *=, TK_MUL_ASSIGN
> > /=, TK_DIV_ASSIGN
> > %=, TK_MOD_ASSIGN
[...]

PR #1350, opened 2022-06-24
[8018x] add code to disable IRQs

cocus:
Bit #4 of each IRQ source's config register selects if the interrupt
is masked or not. Provide a mechanism to disable IRQs.

### DIFF
--- a/elks/arch/i86/kernel/irq-8018x.c
+++ b/elks/arch/i86/kernel/irq-8018x.c
@@ -78,7 +78,8 @@ void enable_irq(unsigned int irq)
     struct irq_logical_map* map;
     map = get_from_logical_irq(irq);
     if (map) {
-        outw(map->config_word, map->pcb_register);
+        // set the priority mask and clear the MSK bit (bit 4)
+        outw(map->config_word, map->pcb_register & 0x7);
     }
 }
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Thanks Santiago! The 8018x should now work with the new networking (multi-NIC) architecture. (This change only affects closing then reopening the new /dev/{ne2k,wd8003,3c509} devices replacing /dev/eth, which enable the NIC IRQ on open and disable it again on close).

cocus:
Sorry @ghaerr , I had a bug on this. I'll push some changes. I've mixed the arguments of outw in both cases. My bad!
However, there's something that could potentially break in the future that I just checked.
WD eth card IRQ maps to the same logical IRQ used by my UART TX IRQ (both #2). What do you think we can do to deal with these IRQ "issues"? Because if we disable the IRQ 2, we'd lose the TX IRQ (not used for the UART0, but this is just an example!)

ghaerr:
> WD eth card IRQ maps to the same logical IRQ used by my UART TX IRQ (both https://github.com/jbruchon/elks/issues/2). What do you think we can do to deal with these IRQ "issues"? Because if we disable the IRQ 2, we'd lose the TX IRQ (not used for the UART0, but this is just an example!)

Well, I don't think we should mix up typical IBM PC architecture IRQs with 8018x architecture, as far as "default" numbering goes. However, our current ports.h doesn't take architecture into account.

Are you saying that the default ports.h (for possible NIC as well as serial) are basically incorrect for 8018x? In that case, I suppose we need to have seperate "ports.h" defaults for separate architectures, and in separate files too.

cocus:
I'm pretty sure we should split these up for this architecture, in particular because I can't think of any easy solution to map IBM PC IRQs into bare 8018x IRQs. For now it should be "fine" but this might bite us back in the future

PR #1351, opened 2022-06-24
[8018x] fix logic from last irq-8018x commit, wrong parameter order on outw()

cocus:


cocus:
This one is tested and works fine. Sorry @ghaerr!

ghaerr:
> I had a bug on this. I'll push some changes. I've mixed the arguments of outw in both cases.

Thanks for testing, I'm glad you tested your changes, even after the PR!! :)

PR #1355, opened 2022-06-30
[net, eth] Major wd driver update, support boot flags for all eth dri…

Mellvik:
The first of several ethernet updates with a reworked wd-driver and several steps toward the optimizations discussed in https://github.com/jbruchon/elks/pull/1341#issuecomment-1166318332. A number of choices need to be discussed before moving forward.

With this update, the `el3/3C509` and `wd` ethernet drivers are on par or faster than the `ne2k` driver. This is remarkable, in particular for the `wd8003` which is an 8 bit card - the explanation being its use of memory mapping instead of PIO.

This update has 2 parts:
1. WD driver mega patch
2. General update of the ethernet subsystem with flags, a common message pool and preps for more optimizations.

**The WD driver**
- Eliminated two race conditions which caused occasional hangs under heavy load
- Reorganized initialization/open/release code to accommodate the new multidriver regime and ensure a clean slate on every open. Also accommodates multiple concurrent opens for IOCTL access. Simplified the driver by eliminating routines called only once and redundant code.
- Added a robust overflow handling, including the ability to have one or several packets survive the recovery procedure. This is important in ELKS because by the time an overflow happens, a read wakeup() has already been called. Purging the entire NIC buffer will cause subsequent read errors in Ktcp, which are non-fatal but noisy and inefficient. The wd driver is now ‘flood ping safe’ @ any packet size that Ktcp can handle.
- Replaced calls to clr_irq/set_irq with 'local' interrupt control: Blocking and reenabling the NIC's own interrupts.
- A number of minor optimizations and fixes, including homogenous error messages between the drivers and boot flag processing. The ETHF_VERBOSE flag now needs to be set to enable all but a few error messages from the drivers.
- New /bootopts flags enable forced 8/16bit bus width and buffer size (8-32k). Incidentally, the implemented error recovery scheme will keep the NIC going even if the given buffer size is 'overstated' in bootconfig (not a recommended setting). Actually a user can run forever with an erroneous buffer size without noticing (if error messages are turned off). There will be a performance penalty, but not more than 20% at most.
- A presumably reliable probe routine has been pulled from the Linux wd-driver and simplified. At boot time the driver identifies itself as `wd8003` or `wd8013` depending on what the probe finds. The driver has been partly updated to accommodate the 16bit `wd8013`, but has not been tested with such hardware yet (my 8013 was supposed to come from France, but disappeared in transit and has been refunded by eBay. A new order is in, this time to Australia).

**The general update and discussion items**
- A first round of optimizations included a shared error print routine and shared message strings in a suggested `eth-shared.c`. This was eventually deemed too complicated and replaced with the current `eth_msgs.h` header file which has the shared error message strings. Other strings have been ‘homogenized’ in place for automatic ‘linker sharing’. Yes, it’s yet another header file, comments solicited.
- All drivers now have a static model_name string which replaces the old ‘eth:’ in most error messages. When working with several interfaces and switching back and forth between them, this became a necessity. 
- Added 'flags' to the NIC identification message at boot. A new default setting for flags should be added to the distribution /bootoopts to enable verbose mode - for now (flags=0x80) (not part of this PR). The flags are defined in `netstat.h`.
- The old default MAC address has been removed from all drivers. More sophisticated probing in all drivers make the backup address superfluous. This means the entire `netif_stat` structure is now autoinitialized instead of initialized manually.
- The new config-flags regime (in `/bootopts`) has not been fully implemented in all drivers yet. The flags are defined and rudimentary described in netstat.h. 
- The `eth struct` in `eth.c` has been augmented by a pointer to the per driver `netif_stat struct`, in preparation for removing the ioctl routine from each driver and have the wrapper handle it. `Select()` and `write()` will follow, `read()` is too complicated and too different from one driver to the next.

**Other notes**
The new network driver regime allows for quick switching between active interfaces. As expected this has turned out to be very efficient in a development environment. IOW it works well. However, there is a small snag which we can fix at moderate cost: When switching interfaces we also switch MAC address. Other hosts on the network will continue to use the old MAC address for a while, until they either give up and send a new ARP request or ELKS initiates a connection and thus makes the other end aware of the change. 

The solution is Gratuitous ARP which is an ARP reply broadcasted without a corresponding ARP request. ktcp could always send that on startup (net start) and we'll get rid of the delay caused by the changing MAC address. See https://www.practicalnetworking.net/series/arp/gratuitous-arp/

FWIW - There are some remaining (and minor) driver problems with ne2k 16bit interfaces when forced to run in 8bit mode, under heavy load. Nothing fatal but something that should be assessed.

Finally - a curiosity: Zillions of tests create zillions of numbers. Among a lot of other things, I noticed the new eth.c wrapper  which enables multi driver support, seems to add approximately 120us to the packet processing time (roundtrip). This increase has been mostly recovered by present optimizations, the rest coming when select() and possibly write() move to eth.c.

Extensive testing with physical (386/20MHz), minimal QEMU testing so far.


ghaerr:
Hello @Mellvik,

Wow, nice work! A couple of quick comments, I haven't yet reviewed the driver changes themselves, but for the kernel changes:

Good solution replacing the do loop with a direct lookup for the file_operations, that's definitely quicker. Can you remove the "#if 1/#else" lines so that just your new code remains? Also, the "if (minor <= MAX_ETHS)" on line eth.c line 31 is off-by-one, should just compare "minor < MAX_ETHS".

With regards to passing the ETH_xxx to each driver init (e.g. el3_drv_init(ETH_EL3) - I'm not sure that's a good idea; the reason is that even though you now use that variable to index the eths[] array to point to netif_stats for the driver, in the same .c file we're using hard-coded indexes for the net_port define(s). That is, the driver file should probably not take a parameter to xxx_drv_init, but just use hard-coded ETH_xxx values throughout, since it basically has to with the current design.

For the error messages, there is current a format error that needs fixing for all that use it: "%#x" is not what you want - please change to "#$x" or "0x%x".

Remove "eth_pmsg() declaration in ne2k.h. (BTW, in new compilers, the "extern" in front of a function forward decl is considered obsolete, its only useful now for data declarations, which require some linker magic. I'm in the process of removing "extern" from forward function declarations throughout the rest of the kernel).

I think the "model_name" should match the name in /bootopts: that is, "WD80x3" should probably be changed to "wd8003", to avoid user confusion.

More comments coming on your PR questions.

Thank you!

ghaerr:
Nice work on the wd.c driver enhancements and testing! Have you tested this PR with all three drivers (except the wd8013)?

> Incidentally, the implemented error recovery scheme will keep the NIC going even if the given buffer size is 'overstated' in bootconfig (not a recommended setting). Actually a user can run forever with an erroneous buffer size without noticing (if error messages are turned off). 

Nice enhancement!

> The old default MAC address has been removed from all drivers.

Does this affect QEMU use in any way? I recall there was a time when we thought we needed to use a specific MAC address for QEMU.

> A new default setting for flags should be added to the distribution /bootoopts to enable verbose mode 

Ok - are you talking about the fourth parameter to each of the three NIC= options in /bootopts? If so, go ahead and add them in another PR. Currently, the default NIC parameters are not specified in /bootopts, instead, likely alternate values are specified. Perhaps this should change to have the default values specified, so that flags is easily changed? I'll leave both to your decision.

> The solution is Gratuitous ARP which is an ARP reply broadcasted without a corresponding ARP request. ktcp could always send that on startup (net start) and we'll get rid of the delay caused by the changing MAC address.

Sounds potentially interesting. I assume this would require ktcp to query the driver to find its MAC address at open time, since it can't otherwise send out an ARP without having received one? OTOH, this makes writing a new driver harder, as the driver needs to support the ioctl before ktcp would use it. Are there other situations where we might not want to always send a G-ARP? It's only needed on switching interfaces, correct?

> I noticed the new eth.c wrapper which enables multi driver support, seems to add approximately 120us to the packet processing time (roundtrip).

That's fixed with your rewriting the do loop in eth.c, correct?

Thank you!

Mellvik:
> 
> Good solution replacing the do loop with a direct lookup for the file_operations, that's definitely quicker. Can you remove the "#if 1/#else" lines so that just your new code remains? Also, the "if (minor <= MAX_ETHS)" on line eth.c line 31 is off-by-one, should just compare "minor < MAX_ETHS".

oopps, thanks! will fix both.

> 
> With regards to passing the ETH_xxx to each driver init (e.g. el3_drv_init(ETH_EL3) - I'm not sure that's a good idea; the reason is that even though you now use that variable to index the eths[] array to point to netif_stats for the driver, in the same .c file we're using hard-coded indexes for the net_port define(s). That is, the driver file should probably not take a parameter to xxx_drv_init, but just use hard-coded ETH_xxx values throughout, since it basically has to with the current design.

I can do that - and as is, it doesn't make much sense. There is, however a plan behind the change: To eliminate the hard coded indexes in order to create the flexibility we've talked about. This may not be the way to get there, but it's an idea - and it seems very simple. My goal at the other end is to be able to boot ELKS with two interfaces using the same driver (e.g. ne2k 16 bit and ne2k 8 bit).

> 
> For the error messages, there is current a format error that needs fixing for all that use it: "%#x" is not what you want - please change to "#$x" or "0x%x".

I've been seeing the '%#x' syntax all over Linux drivers recently, and until yesterday I didn't know what it meant. Looking it up, it seemed like a good way to reduce string sizes by eliminating the '0x' all over. I now realize it doesn't work (isn't implemented?) in ELKS `printk`. What does '#$' do? I couldn't find it in K&R.

> 
> Remove "eth_pmsg() declaration in ne2k.h. (BTW, in new compilers, the "extern" in front of a function forward decl is considered obsolete, its only useful now for data declarations, which require some linker magic. I'm in the process of removing "extern" from forward function declarations throughout the rest of the kernel).

Thanks, I didn't know that. Obviously `eth_pmsg()`  is a leftover from my first stab at shared messages.

> 
> I think the "model_name" should match the name in /bootopts: that is, "WD80x3" should probably be changed to "wd8003", to avoid user confusion.

The name is fixed automatically in `probe` after it figures out what type of card it is.

> Nice work on the wd.c driver enhancements and testing! Have you tested this PR with all three drivers (except the wd8013)?

I have. Since my machines only have 2 ISA slots, testing is always in pairs. Two machines - one with the `wd8003` and a `ne2k`, the 2nd with the `el3` and the 8bit `ne2k`.  A new setup which will accommodate 3 interfaces concurrently is in the works. Obviously I'm having a ball.
> 
> > Incidentally, the implemented error recovery scheme will keep the NIC going even if the given buffer size is 'overstated' in bootconfig (not a recommended setting). Actually a user can run forever with an erroneous buffer size without noticing (if error messages are turned off).
> 
> Nice enhancement!

Thanks - it's more like an accident :-), and took me by surprise. What happens is that when the driver tries to read a packet from non existent buffer memory, it resets the NIC and starts from the beginning. Almost like a continuous overrun situation. What's surprising is that it's so fast you may not notice it. We may have mention in the docs that this may indeed happen. :-)

> 
> > The old default MAC address has been removed from all drivers.
> 
> Does this affect QEMU use in any way? I recall there was a time when we thought we needed to use a specific MAC address for QEMU.

QEMU may not have supported reading the MAC address off of a PROM at some point, that's probably where this came from. That was before my (QEMU) time. Still you have a point: This may affect QEMU autodetect. I'll check and fix that. And - come to think of it - make sure we don't get a duplicate MAC when (if) ELKS initializes 2 `ne2k` interfaces.

> 
> > A new default setting for flags should be added to the distribution /bootoopts to enable verbose mode
> 
> Ok - are you talking about the fourth parameter to each of the three NIC= options in /bootopts? If so, go ahead and add them in another PR. Currently, the default NIC parameters are not specified in /bootopts, instead, likely alternate values are specified. Perhaps this should change to have the default values specified, so that flags is easily changed? I'll leave both to your decision.

Another good point. We should do that. Add a default `flags` value of 0x80 and update `/bootopts` to match the defaults for all interfaces.
> 
> > The solution is Gratuitous ARP which is an ARP reply broadcasted without a corresponding ARP request. ktcp could always send that on startup (net start) and we'll get rid of the delay caused by the changing MAC address.
> 
> Sounds potentially interesting. I assume this would require ktcp to query the driver to find its MAC address at open time, since it can't otherwise send out an ARP without having received one? OTOH, this makes writing a new driver harder, as the driver needs to support the ioctl before ktcp would use it. 

`ktcp` does this already, so it's a minimal change.

> Are there other situations where we might not want to always send a G-ARP? It's only needed on switching interfaces, correct?

Yes, but not only in the `net stop; net start`  situation. It's as useful when physically switching interfaces. Of course all of this is typically in a development scenario, but that's where we are, making life easy on ourselves, right? And no, there are no negative effects of G-ARP, except for a few codelines. 

> 
> > I noticed the new eth.c wrapper which enables multi driver support, seems to add approximately 120us to the packet processing time (roundtrip).
> 
> That's fixed with your rewriting the do loop in eth.c, correct?

That helped. However, I'm inclined to take a run at ne2k.c, replacing some of the status checks with `inb` statements. It's on my immediate list now that I'm at it.

Updates to the PR coming. There is no rush with this.

-M

ghaerr:
> I've been seeing the '%#x' syntax all over Linux drivers 

The '#' is an optional format modifier - placed before the format letter ('x' in this case), which means "alternate format". The alternate format of displaying hex numbers is to prefix "0x" (or "0X" for 'X' format) before the output (unless value is 0). This isn't implemented in ELKS printk (yet). I'll take a look to see how many strings might be better handled like this. As it is, few error messages will be misinterpreted without the 0x prefix. I haven't ever seen "%$" or "%#$".

> WD80x3" should probably be changed to "wd8003", to avoid user confusion.

I saw that after reading the driver source. We should probably change WD to lower case though to be consistent.

> make sure we don't get a duplicate MAC when (if) ELKS initializes 2 ne2k interfaces.

We can't yet init two interfaces of the same driver, so we don't have to worry about that yet.

> My goal at the other end is to be able to boot ELKS with two interfaces using the same driver

After all this is working, we can look into what it will take to have multiple NIC cards of the same type working. As a forewarning, all static data (including the rxwait and txwait scheduler wait queues) will have to be stored in a per-instance structure. Since you use .global data variables in your ASM driver code, this will be more than a little work, and increase the size of the drivers considerably, as every data access will need a pointer to a struct member.

Mellvik:
@ghaerr,
almost all of your comments have been worked in. The Gratuitous ARP has been added, it works great! I've checked the QEMU related question (default address), It's fine. I've also worked in the flags defaults. 

There is a small mystery I'm still working on: The speed varies slightly between `net start; net stop` cycles. It's time consuming thing, so I need some time to investigate that. And it's minor - less than 1ms variation, but I'm the curious type.

> We can't yet init two interfaces of the same driver, so we don't have to worry about that yet.

>> My goal at the other end is to be able to boot ELKS with two interfaces using the same driver

> After all this is working, we can look into what it will take to have multiple NIC cards of the same type working. As a forewarning, all static data (including the rxwait and txwait scheduler wait queues) will have to be stored in a per-instance structure. Since you use .global data variables in your ASM driver code, this will be more than a little work, and increase the size of the drivers considerably, as every data access will need a pointer to a struct member.

I do realize that, and it may not be worth it. I'm really happy with the state of this now, and need to move on to the `ktcp` challenges I started on in February. 

-M

ghaerr:
>  I'm really happy with the state of this now

Ok, would you like this committed now, then? I would still like to see a bit of the cleanup on xxx_drv_init(dev) in eth.c, and a few other minor things. If you like, I can clean that up after commit in a seperate PR, none of which will affect driver operation.

Mellvik:
Thank you @ghaerr,

I'll push the remaining cleanups tomorrow! It just occured to me there is one thing I haven't tested yet: Having two (or more) interfaces on the same IRQ and see if switching between them works.  It should - right?

--M

ghaerr:
> I'll push the remaining cleanups tomorrow!

Thanks. Also, in deveth.c, double semicolon line 54, and please use (void) in parameter decl in arp.c line 135.

> Having two (or more) interfaces on the same IRQ and see if switching between them works. 

Yes - that was the point of moving the request_irq code out of init and into open. This will also be the first actual case of using the dynamic design within the kernel with external hardware - pretty cool. I can imagine the same modification being used in the serial drivers, so that they can be dynamically used with shared IRQs with NIC cards - would you like to take a pass at that?

Mellvik:
> Having two (or more) interfaces on the same IRQ and see if switching between them works.
> 
> Yes - that was the point of moving the request_irq code out of init and into open. This will also be the first actual case of using the dynamic design within the kernel with external hardware - pretty cool. I can imagine the same modification being used in the serial drivers, so that they can be dynamically used with shared IRQs with NIC cards - would you like to take a pass at that?
> 
Absolutely! That will be interesting. My 'rigging' should be just right for such a test. Apropos shared interrupts - is shared (DOS style) serial IRQ on your list? Or even desirable from your point of view?

> 
-M

ghaerr:
> Absolutely! That will be interesting. My 'rigging' should be just right for such a test. 

Agreed. Take a look at rs_init() in serial-8250.c, and lets talk when you're ready to proceed. I just looked at it and it might be a bit more complicated than I thought for the serial console case, because the open routine doesn't get called by the kernel. That may be a non-issue since I think the kernel uses only serial output until /bin/sh or /bin/init is run, and those programs open /dev/console.

> Apropos shared interrupts - is shared (DOS style) serial IRQ on your list? Or even desirable from your point of view?

That would require a rewrite of the serial driver to determine at interrupt time which card is interrupting, and service that UART. I'm not sure this is of great value, since that would further slow down the interrupt processing delay, and we're already on the slow side and get receiver overruns at high baud rates.

ghaerr:
Looks good @Mellvik, are you ready for commit?

We're almost there on the small stuff - can you remove the #if 0 still in eth.c, and also the forward decl for arp_gratuitous() should go in arp.h, rather than in deveth.c (all forward decls are in headers now, never used directly in .c files to avoid duplicate incorrect fwd decls).

Mellvik:
Final cleanups and a couple of fixes/enhancements.

- The EL3 driver needed some work to get close/reopen to work reliably. It turns out shutting off the TP interface on close makes the next open unreliable. And since this 'shutdown' is more of a traditional courtesy than a requirement, it has been turned off.
- IRQ sharing has been extensively tested. The conclusion is that unless all cards on the bus are PnP AND the driver uses PnP capabilities to turn IRQs on and off (i.e. tristate the IRQline on close), such sharing cannot work predictably. Older, pre-PnP cards will typically grab the jumper-configured IRQ line at power and keep it. Newer interfaces will typically tristate the line until the card is activated, regardless of whether PnP is enabled or not. For example, the 3c509 and the new 8bit ne2k card (RTL8019) work well together sharing the same IRQ, none of them using PnP, both having the ability to accept soft configuration - and the 3c509 driver using it. OTOH - that's the only combination that works.
- Having 5 different interfaces to play with, the stats are pretty grim -  reliable. Some combinations will not work at all when configured to the same IRQ line.
- Given this experience I don't think it makes sense to start a shared IRQ 'project' for serial interfaces.

ghaerr:
> IRQ sharing has been extensively tested.

Thank you for the detailed explanations!

> Older, pre-PnP cards will typically grab the jumper-configured IRQ line at power and keep it.

Wow, really? Does that mean that the IRQ itself can't be changed reliably after power-up? We don't support changing the IRQ after boot, so this still should not be a problem, right?

When we speak of "sharing" IRQs, if a NIC has not been "opened", that is, the NIC has not enabled any interrupts, could that same IRQ be shared with a serial port? We're not talking about runtime active sharing of interrupts, just whether or not a 8086 low-core IRQ vector could point to a serial versus NIC driver, and only have one of the chips have interrupts enabled at a given time.

What I'm still a bit unclear on is exactly what a NIC might do if it is NOT enabled, (and thus disallowed from generating an interrupt) - can that IRQ be used by the CPU or other device separately?

> Given this experience I don't think it makes sense to start a shared IRQ 'project' for serial interfaces.

If you mean runtime sharing of the IRQ between active serial interfaces, agreed. But could we share, say, IRQ 3 for serial or NIC, and only have one active at a time?

ghaerr:
> Older, pre-PnP cards will typically grab the jumper-configured IRQ line at power and keep it.

I see - the NIC grabs the hardware IRQ line, so that it can't be used by any other hardware. Geez.

Mellvik:
> Older, pre-PnP cards will typically grab the jumper-configured IRQ line at power and keep it.
> 
> I see - the NIC grabs the hardware IRQ line, so that it can't be used by any other hardware. Geez.
> 
Yes, that pretty much sums it up. I haven't looked at schematics or specs and I'm a bit rusty on low level hardware but this is how I remember it, and how it actually seems to act. The older cards have no clue about tri-state so they own the line you assign by the jumper. Whether they use it or not. In order for sharing to work, we need the other (inactive) card to tristate the line when not in use. Most cards we 'meet' won't do that.

-M

ghaerr:
Looks great @Mellvik, thank you for this large effort. Let me know when you're ready for commit.

Mellvik:
Thank you, @ghaerr.

I'm ready. Still experimenting with two of the NICs in order to satisfy my own curiosity (and learning more about ISA along the way), acknowledging that the best we can do as far as irq sharing is concerned, is 'maybe'. The thinking is that if the ability to program (soft-set) the irq is there, switching to a different irq will have the same effect as tri-stating the shared one. Some irq lines are not physically present on the isa bus ... I'm sure you get the drift...

Worth some investigating. And low expectations. Not worth holding up the merge.

-M

ghaerr:
With regards to ELKS kernel sharing of IRQs, I've added changing the serial port IRQ from init to open time to my list. As mentioned before, there may be some issues regarding serial console, which I'll look into.

Once this is done, IRQ 4 and 3 will be "available" to experiment with, just like when those serial cards are not installed, to be usable by other cards, even with the serial card installed. (Currently, the serial init code probes for serial cards and doesn't request the IRQ vector from the kernel if the UART is not found). Thus, once this mod is made, one should be able to use a NIC set to IRQ 3 provided that 1) the serial port using that IRQ is not opened, and 2) both the serial card and NIC behave nicely by not grabbing the hardware IRQ 3.

Mellvik:
> 
> With regards to ELKS kernel sharing of IRQs, I've added changing the serial port IRQ from init to open time to my list. As mentioned before, there may be some issues regarding serial console, which I'll look into.
> 
> Once this is done, IRQ 4 and 3 will be "available" to experiment with, just like when those serial cards are not installed, to be usable by other cards, even with the serial card installed. (Currently, the serial init code probes for serial cards and doesn't request the IRQ vector from the kernel if the UART is not found). Thus, once this mod is made, one should be able to use a NIC set to IRQ 3 provided that 1) the serial port using that IRQ is not opened, and 2) both the serial card and NIC behave nicely by not grabbing the hardware IRQ 3.
> 
Sounds good @ghaerr.
Honestly I don't think this will work, but I'll be happy to prove myself wrong. 

The thing is, most (non-pnp) serial cards (and other cards including NICs as we've seen), expect to own the line at power on, and don't need any open or init to do so. When two cards do that, the system will hang. If one is well behaved and the other not, it will work for a while depending on the sequence of things.
Again, let's prove me wrong. :-)

-M

PR #1358, opened 2022-07-02
[basic] Update Readme of basic

tyama501:
Thank you for merging the previous commit @ghaerr ,

Sorry,  I forgot to update the Readme.

I moved PEEK/POKE as implemented.


ghaerr:
Can you note in the README that PEEK and POKE return/use byte values, since there is no indication in the name as there is with OUTW/OUTB, etc?

PR #1359, opened 2022-07-06
[cmds] Bugfixes in sys and cp

Mellvik:
The `/bin` directory is now so big that `/bin/*`  will cause `sark` problems for several apps, including `cp`. So `sys` failed. Changed the script to use `cp -R .` instead, and fixed `usage()`.

Also changed default verbose mode to off in `cp`.

### DIFF
--- a/elks/arch/i86/drivers/net/el3.c
+++ b/elks/arch/i86/drivers/net/el3.c
@@ -390,7 +390,7 @@ static void el3_int(int irq, struct pt_regs *regs)
 			if (status & StatsFull)			/* Empty statistics. */
 				update_stats();
 
-#if 0			// Not currently used
+#if LATER		/* Candidate for optimization  */
 			if (status & RxEarly) {
 				//el3_rx(dev);
 				outw(AckIntr | RxEarly, ioaddr + EL3_CMD);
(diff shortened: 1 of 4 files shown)


ghaerr:
Will the script still run by using "cp -R /bin $MNT/bin"? That would be preferable to having to change the directory within the script (not changing the current directory may be desirable for other operations).

Mellvik:
Yes, it does work that way too. I probably botched something when testing. Absolutely preferable.

BTW - sys -M works well on any partition, needs fdisk to set active partition. Maybe a message to that effect at the end of sys would be good. 

-M

ghaerr:
> BTW - sys -M works well on any partition, needs fdisk to set active partition. 

Thanks for that testing. It is possible we could have `makeboot` (which is called by sys -M to write the partition boot block) activate the partition in the MBR, but that may not be what the user desires.

> Yes, it does work that way too. Absolutely preferable.

Great. Please change that and we'll commit. We can decide whether to change sys -M behavior in a subsequent commit.

Thank you!

Mellvik:
> 
> BTW - sys -M works well on any partition, needs fdisk to set active partition.
> 
> Thanks for that testing. It is possible we could have makeboot (which is called by sys -M to write the partition boot block) activate the partition in the MBR, but that may not be what the user desires.
> 
Agreed, so it should either be optional or just a message at exit time to that effect.
> Yes, it does work that way too. Absolutely preferable.
> 
> Great. Please change that and we'll commit. We can decide whether to change sys -M behavior in a subsequent commit.
> 
Coming up.

> Thanks,
> 
> M

PR #1361, opened 2022-07-08
[kernel] Cleanup C code using gcc's enhanced warnings

ghaerr:
During cleanup found a couple of "shadow variable" bugs (where a local variable shadows another of the same name out of scope).

@Mellvik, the additional GCC warnings highlighted a couple places of bad code warnings in the network drivers, which you should look at and fix next time you compile the kernel. I decided to leave them alone since I can't test them.

Next step will be adding `const` declarations more deeply into the kernel source.


Mellvik:
Thank you @ghaerr,
this is very useful indeed. I’ll include fixes with the ‘sys’ update later today. Both were harmless btw, but nevertheless important cleanups.

—M

[ iPhone ]

PR #1364, opened 2022-07-11
[tools] elf2elks: fix: forgot to write data seg. reloc. count in a.out header

### LINKED ISSUE #117: Medium model data relocations not emitted for ELKS?
In the process of porting the D-Flat TUI library to ELKS, which requires medium model due to code size, it seems that the .data section R_SEGWORD segment relocations are not being emitted at final link time, at least for the version of the ld.gold linker (or elf2elks perhaps) ELKS is running (ld.gold is a bit old, but elf2elks should be current). I'm wondering whether this might have changed in the last year or so (after the medium model work was first completed by both of us on ELKS).

The compilation options being used are:
`-fno-inline -melks-libc -mcmodel=medium -msegment-relocation-stuff -mtune=i8086 -Wall -Os -ffunction-sections`

In particular, the a.out supplemental header msh_drsize longword is always 0, despite some .data segment relocations being emitted into the .o files: 
```
normal.o:     file format elf32-i386

Disassembly of section .data:

00000000 <classdefs>:
   0:   ff                      (bad)
   1:   ff 00                   incw   (%bx,%si)
            2: R_386_16 NormalProc
            2: R_386_SUB16  NormalProc!
    ...
            4: R_386_SEG16  NormalProc!
            a: R_386_16 ApplicationProc
            a: R_386_SUB16  ApplicationProc!
   b:   00 00                   add    %al,(%bx,%si)
            c: R_386_SEG16  ApplicationProc!
   d:   00 c0                   add    %al,%al
```
It took some time to debug this, but the `classdefs` array above stores a 4-byte function pointer at byte offsets 2-5 as can be seen above (and then again at offsets 0x0A-0x0D etc). What is happening is the segment value of the function pointer stored in the .data section is always 0x1000, rather than the CS-relocated value. I was able to get this to work for the time being by storing a properly-runtime-relocated .text/.fartext-based function pointer into the array at runtime startup, after which all works well (here's a cool screenshot):

ELKS dflat

The TUI library is rather large, and I thought to ask what you think about this before going to the trouble of producing a smaller sample and/or download a later ia16 tree to reproduce the problem. Dumping the a.out file header always shows the `msh_drsize` longword as 0. Checking the ELKS loader at elks/fs/exec.c, I see we do support relocating .data relocations, but IIRC this has never been tested, as we'd never produced an executable that required .data segment relocations.

Thank you!

tkchia:
Should fix https://github.com/tkchia/gcc-ia16/issues/117 .

### DIFF
--- a/elks/tools/elf2elks/elf2elks.c
+++ b/elks/tools/elf2elks/elf2elks.c
@@ -654,6 +654,7 @@ output_header (void)
     {
       mh.hlen = sizeof mh + sizeof esuph1 + sizeof esuph2;
       esuph1.trsize = (uint32_t) text_n_rels * sizeof (struct minix_reloc);
+      esuph1.drsize = (uint32_t) data_n_rels * sizeof (struct minix_reloc);
       esuph2.ftseg = ftext_sh->sh_size;
       esuph2.ftrsize = (uint32_t) ftext_n_rels * sizeof (struct minix_reloc);
       output (&mh, sizeof mh);
@@ -664,6 +665,7 @@ output_header (void)
     {
       mh.hlen = sizeof mh + sizeof esuph1;
       esuph1.trsize = (uint32_t) text_n_rels * sizeof (struct minix_reloc);
+      esuph1.drsize = (uint32_t) data_n_rels * sizeof (struct minix_reloc);
       output (&mh, sizeof mh);
       output (&esuph1, sizeof esuph1);
     }


ghaerr:
Hello @tkchia,

Thank you, I very much appreciate this fix! I have tested this, and it does solve the problem.

toncho11:
Hi @ghaerr ,

What applications are available on this TUI? How is the port going?

PR #1365, opened 2022-07-11
[serial driver] Removed hardcoding of IRQs for S2, S3

Mellvik:
Replaces the former hardcoding of serial IRQs which effectively ignored the IRQ settings in `ports.h` with something that works with any IRQs.

@ghaerr, there are two versions in the code - to be discussed. The commented-out version uses a function (search) to map the irq into the serial struct.  Code size 39 bytes. 
The active version uses a 16 byte array instead, which is both faster and more compact. The array could be shortened by 2 or 3, but I'm not convinced it's worth it. 
Comments solicited.

Needless to say, this one hit me real hard this last week - until I finally remembered. :-)

-M

### DIFF
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -64,7 +64,7 @@ static struct serial_info ports[NR_SERIAL] = {
     {(char *)COM4_PORT, COM4_IRQ, 0, DEFAULT_LCR, DEFAULT_MCR, 0, NULL, 3,0,0},
 };
 
-static char ttytab[16];
+static char irq_to_port[16];
 static unsigned int divisors[] = {
     0,				/*  0 = B0      */
     2304,			/*  1 = B50     */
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @Mellvik,

This is a very nice fix, and gets rid of the longstanding serial driver IRQ kluge! Nice!!! An ingenious solution, it never occurred to me to use a dynamic lookup table built at init time.

I definitely agree with your decision to use the ttytab[] array for very quick indexing, rather than use the slower for-loop solution, given that we've already got serial receive speed issues. So lets keep that and go ahead and delete the commented out solution.

Can you add a comment to the ttytab[] declaration on line 67 "/* convert IRQ to tty index */" or something like that (or perhaps rename ttytab to irq_to_port[] for clarity?

Looking at line 454 `ttytab[sp->irq] = sp->index;`, we don't really need to use a hard-coded index that has to be manually set and use a structure member: rewriting that as `ttytab[sp->irq] = sp - ports;` will accomplish the same thing and not require a ports[] structure member.

Well done, thank you!

Mellvik:
Oops, forgot to update the struct (remove the index), coming up.
M

[ iPhone ]

ghaerr:
Thank you again for the nice technical enhancement!

PR #1369, opened 2022-07-14
[kernel][libc] Fix memcmp

Mellvik:
Some small fixes:

- `memcmp` was broken in both libc and kernel.
- the `cmp` command (which led to the fixes above) reported differences one byte off.
- added `string.h` to `eth.c` to get rid of compiler warnings.

### DIFF
--- a/elks/arch/i86/drivers/char/eth.c
+++ b/elks/arch/i86/drivers/char/eth.c
@@ -8,6 +8,7 @@
 #include <linuxmt/errno.h>
 #include <linuxmt/fs.h>
 #include <linuxmt/netstat.h>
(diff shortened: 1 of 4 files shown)


ghaerr:
Wow, @Mellvik, nice find!

Kind of amazing that my last memcmp change was buggy and I didn't notice. I had replaced your memcmp in ne2k with a smaller version, which I got out of the kernel strings library, and then replaced the C library with that one as well, assuming it worked and only looking at code size and speed. And amazing that the old kernel-supplied memcmp was buggy. Goes to show that every submission needs testing, no matter how small!

Thank you!

Issue #1372, opened 2022-07-22
Compress executable missing on Debian/Ubuntu builds


sparky4:
cant build on Debian 11
- What is expected ?
it is supposed to build
- What is observed ?
compress program not found

**Configuration**
- Version (commit or release):
current

**How to reproduce ?**
- Does the problem always happen ?
yes
- List the steps to reproduce the problem.
run ./build.sh

**Additional information**
```
ifdef CONFIG_APP_MAN_PAGES
        mkdir -p $(DESTDIR)/lib/man1
#       cp -rp man/man1 $(DESTDIR)/lib; compress -b 12 $(DESTDIR)/lib/man1/*
        mkdir -p $(DESTDIR)/lib/man2
#       cp -rp man/man2 $(DESTDIR)/lib; compress -b 12 $(DESTDIR)/lib/man2/*
        mkdir -p $(DESTDIR)/lib/man5
#       cp -rp man/man5 $(DESTDIR)/lib; compress -b 12 $(DESTDIR)/lib/man5/*
        mkdir -p $(DESTDIR)/lib/man7
#       cp -rp man/man7 $(DESTDIR)/lib; compress -b 12 $(DESTDIR)/lib/man7/*
        mkdir -p $(DESTDIR)/lib/man8
#       cp -rp man/man8 $(DESTDIR)/lib; compress -b 12 $(DESTDIR)/lib/man8/*
endif
```
i commented out the man pages stuff and it compiled! <3

toncho11:
Yes, I had the same problem. I had to install "ncompress" on Ubuntu.

ghaerr:
> Yes, I had the same problem. I had to install "ncompress" on Ubuntu.

Super @toncho11, that's definitely the problem and fix here, thanks!

@sparky4, we have a couple options here: since the actual problem is that the `compress` executable apparently isn't installed by default on Ubuntu or Debian 11, we can either 1) issue a helpful error message and stop the build, or 2) not compress the man pages, which will result in a much larger disk image (and the user won't know).

Any thoughts about the best way to proceed?

toncho11:
I think you can update: https://github.com/jbruchon/elks/blob/master/BUILD.md , section "Prerequisites" - you can add that compress is also needed and explain that on Ubuntu and Debian (not sure) the package name is "ncompress".

toncho11:
Actually on ubuntu you also need to do:
`sudo apt install build-essential` to install the C compiler, but everyone is supposed to know that.

[...]

Issue #1368, opened 2022-07-12
Possible ktcp/SLIP bug


Mellvik:
@ghaerr - I'm hoping you had an idea about this one:

I've been running `slip` a lot lately (not `cslip` which has issues, but that's a different story, to be revisited) and keep running into a problem that can be reproduced as follows:

- Boot system with `slip` as the default network setting
- ELKS comes up with `slip` active and working (in my case on `ttyS1`, console is on `ttyS0`, there is a `getty` running on `ttyS2`)
- Run `net stop`, `ps` confirms everything network related is gone
- Run `net start slip`
- The net starts like the first time and works, but all processes connected to `ttyS0` and `ttyS2` are gone and no new `getty` starts.
- Run (from either a telnet session or the physical console) `init 3`, and `getty`s restart on S0 and S2.

So it seems the second (and later) `net start` somehow affect `init` which needs a kick to get back amongst the living.

Ideas welcome.

```
Direct console, scan kbd 80x25 emulating ANSI (3 virtual consoles)
ttyS0 at 0x3f8, irq 4 is a 16550A
ttyS1 at 0x2f8, irq 3 is a 16550A
ttyS2 at 0x3e8, irq 10 is a 16550A
ttyS3 at 0x2e8, irq 7 is a 16550A
64 ext buffers, 65536 ram
eth: ne2k at 0x380, irq 9 (8bit), MAC 00:1f:11:02:60:2d, flags 0x80
eth: wd80x3 at 0x240, irq 2, ram 0xce00 not found
eth: 3c509 at 0x330, irq 11 not found
bioshd: hda BIOS CHS 992,16,63
bioshd: hda  IDE CHS 993,16,63
bioshd: hdb BIOS CHS 728,2,1
/dev/hda: 993 cylinders, 16 heads, 63 sectors = 488.7 Mb
/dev/hdb: 728 cylinders, 2 heads, 1 sectors = 728.0 kb
/dev/fd0: 80 cylinders, 2 heads, 18 sectors = 1440.0 kb
Partitions: hda:(0,1000944)  hda1:(63,696465)  hda2:(696528,132048)  hda3:(828576,132048)  hda4:(960624,41328) 
 hdb:(0,1456)  hdb1:(1,1455) 
device_setup: BIOS drive 0x80, root device 0x303
PC/AT class machine, syscaps 0xff, 608K base ram.
[...]

Issue #1371, opened 2022-07-20
Not reading bootopts


toncho11:
The kernel does not read the bootopts file. I copied the file manually from Windows. The sys command did not put this file. Actually the sys command installation failed with not finding the "rmdir" command using a 360 KB floppy. But it actually boots OK, so I thought the installation went OK.

Is this a file that should be stored on a specific place on the filesystem? How should I put it there?
Or it could be the extra "0x80" for the WD8000? I have not used a fourth parameter before.

```
## boot options max 511 bytes
#console=ttyS0,57600 debug net=ne2k 3 # sercons, multiuser, networking
#QEMU=1			# to use ftp/ftpd in qemu
#TZ=MDT7
LOCALIP=192.168.1.67
#HOSTNAME=elks16
#ne2k=12,0x300,,0x80
wd8003=3,0x280,0xD000,0x80
#3c509=11,0x330,,0x80
#bufs=2500		# system buffers
#sync=30		# autosync secs
#init=/bin/init 3 n	# multiuser serial no /etc/rc.sys
#init=/bin/sh		# singleuser shell
#root=hda1 ro		# root hd partition 1, read-only
#console=ttyS0,19200 3	# serial console
```

toncho11:
At boot time I see the default values for WD, not the ones in the bootopts.

Mellvik:
Hi @toncho11,

can you post the boot messages?

The requirements for /bootopts is that it resides in the root directory, that the file size does not exceed 511 bytes and the two first chars must be ##. It looks good but check the file size anyway. 

If it has been edited in windows, it may have been saved with <CR><LF> line termination instead of just <LF>. I don't know if it makes a difference, but check.

—M

ghaerr:
Hello @toncho11,

I think the problem is that the /bootopts directory entry is not within the first 8 directory entries on the volume. That is, if the /bootopts file were created by copying after the volume already has more than 7 files in the root directory, it won't work. I have noted this, and code in setup.S will have to be changed to read more root directory sectors in order to work.

A temporary workaround for this would be to 1) rename /bootopts to /bootopts.tmp, perform an unsorted directory listing (`ls -U` on ELKS, or I think just `dir` on Windows), and delete any one of the first 8 files shown, then copy /bootopts.tmp to /bootopts; or 2) create a blank diskette and copy /linux and /bootopts to it, before running `sys` on it.

[...]

PR #1373, opened 2022-07-23
[kernel][libc] Fix ANSI colors, libc strtol/strtoul, net script

ghaerr:
Fixes and small enhancements to various issues.

Corrects console color sequences to ANSI standard - ELKS was using CGA/EGA colors instead of ANSI in ESC [ fg;bg m sequence.

Adds ESC 7 and ESC 8 save/restore cursor position for ANSI console.

Enhances `more` to behave like `more -c`, which displays "(END)" even when file total lines are less than 25, and also clears screen on start. This is to work with upcoming `fm` ELKS file manager.

Fixes incorrect result returned in libc `strtol` and `strtoul` when numeric string didn't end in NUL.

Adds <sys/param> libc header file.

Removes `init 1` from `net start slip` script, changing init run levels turned off, so stopping network doesn't turn off any multiuser mode. (#1368).

In testing this PR for colors, I noticed that `kilo`s syntax color scheme is mostly broken... will look further into that.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -8,6 +8,7 @@ To build ELKS, you need a GNU development environment, including:
 - texinfo
 - libncurses5-dev
 - mtools-4.0.23 (for MSDOS/FAT images)
+- compress (for compressed man pages; use `sudo apt-get install ncompress`)
 
 ## Quickstart
 
--- a/elks/arch/i86/drivers/char/console.c
+++ b/elks/arch/i86/drivers/char/console.c
@@ -163,13 +163,11 @@ static void AnsiCmd(register Console * C, char c)
 	  n = atoi(p);
 	  if (n >= 30 && n <= 37) {
 	    C->attr &= 0xf8;
-	    //C->attr |= (n-30) & 0x07;
 	    C->attr |= ega_color[n-30] & 0x0F;  /* some colors bright by default */
 	    C->color = C->attr;
 	  }
 	  else if (n >= 40 && n <= 47) {
 	    C->attr &= 0x8f;
-	    //C->attr |= ((n-40) << 4) & 0x70;
 	    C->attr |= (ega_color[n-40] << 4) & 0x70;
 	    C->color = C->attr;
 	  }
--- a/elks/arch/i86/drivers/char/console.c
+++ b/elks/arch/i86/drivers/char/console.c
@@ -74,6 +74,10 @@ static void itoaQueue(int i)
 	Console_conin(*b++);
 }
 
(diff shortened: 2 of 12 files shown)


ghaerr:
Also fixes #1372, won't compress man pages if compress utility not found on host; built continues normally.

Adds compress prerequisite to BUILD.md.

Changes "Exec format error" error message to "Exec format error (install ash to run shell scripts)", discussed in #1371.

Other small cleanups.

ghaerr:
Hello @tyama501,

This PR does correct some of the colors that are used in the ESC [ fg; bg m ANSI color sequence. Instead of incorrectly passing the fg/bg colors directly (after subtracting 30 or 40 respectively) to the screen attribute location, the new code maps the ANSI-specified color to the EGA/CGA color. This should all work fine, except that in a couple cases, a high-bright foreground color is used by default for better clarity, which adds in +8 to the fg color. I don't know whether the PC-98 is compatible or not with the IBM PC EGA/CGA; if not, then we may need to ifdef the code in console.c. The affected lines are 166 and 78 in console.c.

Thank you!

tyama501:
Hello @ghaerr 

I think there is converting attribute conv_pcattr function in console-direct-pc98.c and masking color to 3bits. So I think there is no effect but I will check later.

pc-98 also has 4bits, 16colors with the most significant bit is the intensity but it seems the bit has no effect by default.

tyama501:
Hello @ghaerr ,

Before I apply this patch, it seems color with kilo already got strange.

kilo

tyama501:
Here is the picture after I applied this patch.

Although I don't know it is the right color, it is much better.

kilo2

tyama501:
I downloaded this IBM image and tried the same thing with QEMU.

The color is same.

So, I think it is OK for PC-98.

kilo3_qemu

ghaerr:
Hello @tyama501,

Thank you for your testing, I can see that the changes are correct, and much better on both the PC-98 and IBM PC. The very strange thing I can't figure out, is that kilo isn't working for me (with or without this patch): the color syntax mechanism doesn't appear to be working. For instance, here's a .c source file, and both reserved words and variable names are the same color:
Screen Shot 2022-07-23 at 8 46 58 AM

The color system itself seems to be working, as ^A (show help) displays properly:
Screen Shot 2022-07-23 at 8 58 33 AM

I can't figure out why its not working on my system!

ghaerr:
> I can't figure out why its not working on my system!

Ok - finally figured it out. I hadn't pulled down @Mellvik's `memcmp` fix from #1369. Once that was included, `kilo` is now working fine.

PR #1376, opened 2022-07-26
[kernel] Fix duplicate character received in serial driver on QEMU

ghaerr:
This fixes a long-standing problem with QEMU and the serial driver: on occasion, QEMU somehow triggers a second serial interrupt and the last received character is "received" again. The fix is to look at the UART data ready register and return if no character is ready to be received.

During the new `fm` file manager port, which uses an ANSI sequence to tell the remote terminal to send streaming mouse data immediately when the mouse is moved, large streams of input data were being received quickly, and it was finally found that this is the problem that caused some mouse and arrow key keypresses to be ignored. The reason is that ELKS received a random portion of the ANSI sequence twice, thus throwing off the application (fm, vi, dflat, etc) that was doing the ANSI sequence recognition. This could also be the reason that `vi` isn't always able to determine the remote terminal window size correctly (using ANSI DSR) every time, but I haven't checked that yet.

@Mellvik, this hasn't been tested on real hardware, but shouldn't be an issue, as some older code was uncommented, which previously recognized this problem. It does add about 10 more instructions per interrupt to our serial IRQ processing, but has to be done in order for the terminal sequences to work properly. This will also fix any serial hardware that might have the same issue. I suspect the QEMU issue is related to their emulation of the 8259 IRQ line being held high for a bit too long, causing another "fake" interrupt, but not sure. Hopefully adding his extra code won't materially affect our serial driver max input baud rate, although the interrupt routine does already operate with interrupts allowed, so it shouldn't, unless the UART gets a receiver overrun.

### DIFF
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -287,14 +287,15 @@ void rs_irq(int irq, struct pt_regs *regs)
     char *io = sp->io;
     struct ch_queue *q = &sp->tty->inq;
 
-#if 0	// turn on for serial stats
     int status = INB(io + UART_LSR);			/* check for data overrun*/
+    if ((status & UART_LSR_DR) == 0)			/* QEMU may interrupt w/no data*/
+	return;
+
+#if 0	// turn on for serial stats
     if (status & UART_LSR_OE)
 	printk("serial: data overrun\n");
     if (status & (UART_LSR_FE|UART_LSR_PE))
 	printk("serial: frame/parity error\n");
-    if ((status & UART_LSR_DR) == 0)			/* QEMU may interrupt w/no data*/
-	printk("serial: interrupt w/o data available\n");
 #endif
 
     /* read uart/fifo until empty*/


Mellvik:
This is good news @ghaerr - 

I'm not sure if I have seen this problem on real hardware - I may have since on my list is an occasional checksum error appearing when running SLIP, which I've done a lot lately. If this is something that happens only when receiving back to back characters (which seems likely, otherwise we'd have seen it in commands and edits over serial) it may also have contributed to the occasional problems in VI, misinterpreting ANSI sequences from arrow keys. Which I've always assumed is a timing problem.

I don't think the extra instructions are a big deal. We'll know when we beat it up for real. I suggest you commit as we take it from there.

Thank you.

-M

toncho11:
yes, vi on ELKS has some problems.

ghaerr:
> If this is something that happens only when receiving back to back characters (which seems likely, otherwise we'd have seen it in commands and edits over serial)

Yes, the problem is specifically that back-to-back characters were repeated, which was found with a new ttyinfo program that displays all keyboard and mouse input sequences and has been quite useful for testing (borrowed from the Cosmopolitan project). After seeing errors in the displayed results, I finally realized the errors were always a duplicate of the same character, and that QEMU's emulation of the hardware UART is buggy. This fixes that.

> it may also have contributed to the occasional problems in VI, misinterpreting ANSI sequences from arrow keys. Which I've always assumed is a timing problem.

Keep an eye out for any additional "occasional" problems. However, the new library code running in `fm` and dflat is based on a new algorithm - one which I also borrowed from Cosmopolitan. It deterministically reads input based on internal knowledge of properly-formed ANSI sequences and knows when they should terminate. Our old algorithm, still being used in vi and sh, use the VMIN and VTIME termios elements and timeout after 100 ms of receiving an ESC or any other character. This means that a timeout or reschedule at a bad time may not properly recognize a proper ANSI sequence. This is necessary for vi since it also uses a lone ESC as a command character. `sh` can probably be updated with the new algorithm. Ultimately, I hope to have all this in library format and combine the two algorithms.

Mellvik:
> 
> Our old algorithm, still being used in vi and sh, use the VMIN and VTIME termios elements and timeout after 100 ms of receiving an ESC or any other character. This means that a timeout or reschedule at a bad time may not properly recognize a proper ANSI sequence. This is necessary for vi since it also uses a lone ESC as a command character. sh can probably be updated with the new algorithm. Ultimately, I hope to have all this in library format and combine the two algorithms.
Yes, it's easy to see that vi is a tough one in this context. And - come to think of it, sh/linenoise has been as much if not more of a pain-in-the-..., so this is a very welcome fix indeed!

Thank you!

-M

ghaerr:
> sh/linenoise has been as much if not more of a pain-in-the-

What problems are you having with sh/linenoise? I previously shut off the line-width determination via ANSI DSR, as that didn't work very well; I am now considering rewriting it and/or turning it back on. I'm interested in exactly what problems you have seen - are they just not recognizing arrow keys on QEMU sometimes, or issues on real hardware as well?

Mellvik:
> 
> sh/linenoise has been as much if not more of a pain-in-the-
> 
> What problems are you having with sh/linenoise? I previously shut off the line-width determination via ANSI DSR, as that didn't work very well; I am now considering rewriting it and/or turning it back on. I'm interested in exactly what problems you have seen - are they just not recognizing arrow keys on QEMU sometimes, or issues on real hardware as well?
> 
> —
> 
Actually it's exactly the same problem as in vi, just more frequent (likely because I use sh much more than vi): instead of the expected cursor movement, i get the char after the [ in the ansi sequence,  'A', B , C , D. In vi it's more fun because A enters insert mode etc... 
'undo' is great...

-M

ghaerr:
> Actually it's exactly the same problem as in vi, just more frequent (likely because I use sh much more than vi): instead of the expected cursor movement, i get the char after the [ in the ansi sequence,  'A', B , C , D. In vi it's more fun because A enters insert mode etc...

Got it. The `fm` program used to use 'M' to run more, but the mouse ending sequence is in M, which would screw things up in the same fashion! It all works now though with this patch.

Does this vi/sh behavior occur on the ELKS console, or just in Terminal?

Mellvik:
> Actually it's exactly the same problem as in vi, just more frequent (likely because I use sh much more than vi): instead of the expected cursor movement, i get the char after the [ in the ansi sequence, 'A', B , C , D. In vi it's more fun because A enters insert mode etc...
> 
> Got it. The fm program used to use 'M' to run more, but the mouse ending sequence is in M, which would screw things up in the same fashion! It all works now though with this patch.
> 
> Does this vi/sh behavior occur on the ELKS console, or just in Terminal?
> 
I cannot recall ever having seen this on the physical console. Then again, I'm on serial 99.5% of the time, probably too little to be relevant in this context. 
Now that you bring it up, I'll keep an eye out for symptoms.

Thank you.

-M

PR #1377, opened 2022-07-26
[images] Fix sash overwriting ash on 2880K and HD builds

ghaerr:
Likely since v0.6.0, all full-sized image builds incorrectly overwrite /bin/sh with `sash`, due to TAGS= in elkscmd/Make.install. Fixes by reversing order of copy in elkscmd/Applications.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -47,8 +47,8 @@
 sys_utils/init			:boot	:sysutil    :128k
 sys_utils/getty			:boot	:sysutil    :128k
 sys_utils/login			:boot	:sysutil    :128k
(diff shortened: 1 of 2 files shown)


toncho11:
I think that was exactly my problem! I was taking ash from an HD image and each time I was taking sash instead of ash!
Well done!

ghaerr:
Well, @toncho11, I was thinking of you when fixing this lol :)

I found this searching for answers on @tyama501's question about /dev/ptyp2 not opening, and he's on FAT disks. When I went to duplicate on 2880K FAT, the shell acted quite strangely, until I figured out it was sash! Took me a while to figure out what was happening.

PR #1385, opened 2022-07-30
[kernel] Serial port driver: request IRQ on open and release on close

ghaerr:
Enhances the IBM PC serial driver to grab the associated hardware IRQ vector only at open time, and release it at close, for other driver use. No changes to the probing function, nor serial port operation.

Discussed in https://github.com/jbruchon/elks/issues/1312#issuecomment-1160558400, this allows for potentially sharing any of the compiled-in serial driver IRQs (COM1-4) with any other driver. By "sharing", it means a driver can use the IRQ only when another driver that might use it (shares the same IRQ number) is closed. If a driver tries to use an already-used IRQ, the system will return -EBUSY ("Device or resource busy").

Also, when looking through the TTY driver source, it was found that O_EXCL is already implemented for all TTYs (and thus serial ports). Thus, any application can already use O_EXCL to open a serial port exclusively. This could be changed for ktcp's SLIP opens, if desired.

During the testing for this PR, it was noticed that the shell only gave an error number, rather than the appropriate "device or address not found" string, when trying to open a non-existent serial port (e.g. `cat > /dev/ttyS1` without COM2 compiled in). This is fixed with the -ENXIO fix below.

This PR also updates the serial driver to return the proper -ENXIO error, rather than -ENODEV, when a device doesn't exist. It looks like a lot of the other character device drivers also do the same. This will be fixed for all them in a seperate PR.

### DIFF
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -319,6 +319,7 @@ static void rs_release(struct tty *tty)
     debug_tty("SERIAL close %d\n", current->pid);
     if (--tty->usecount == 0) {
 	OUTB(0, port->io + UART_IER);	/* Disable all interrupts */
+	free_irq(port->irq);
 	tty_freeq(tty);
     }
 }
@@ -331,14 +332,32 @@ static int rs_open(struct tty *tty)
     debug_tty("SERIAL open %d\n", current->pid);
 
     if (!(port->flags & SERF_EXIST))
-	return -ENODEV;
+	return -ENXIO;
 
     /* increment use count, don't init if already open*/
     if (tty->usecount++)
 	return 0;
 
+    switch(port->irq) {
+#ifdef CONFIG_FAST_IRQ4
+    case 4:
+	err = request_irq(port->irq, (irq_handler) _irq_com1, INT_SPECIFIC);
+	break;
+#endif
+#ifdef CONFIG_FAST_IRQ3
+    case 3:
+	err = request_irq(port->irq, (irq_handler) _irq_com2, INT_SPECIFIC);
+	break;
+#endif
+    default:
+	err = request_irq(port->irq, rs_irq, INT_GENERIC);
+	break;
+    }
+    if (err) goto errout;
+
     err = tty_allocq(tty, RSINQ_SIZE, RSOUTQ_SIZE);
     if (err) {
+errout:
 	--tty->usecount;
 	return err;
     }
@@ -443,21 +462,6 @@ static void rs_init(void)
     do {
 	irq_to_port[sp->irq] = sp - ports;	/* Map irq to tty # */
 	if (!rs_probe(sp)) {
-	    switch(sp->irq) {
-#ifdef CONFIG_FAST_IRQ4
-	    case 4:
-		request_irq(sp->irq, (irq_handler) _irq_com1, INT_SPECIFIC);
-		break;
-#endif
-#ifdef CONFIG_FAST_IRQ3
-	    case 3:
-		request_irq(sp->irq, (irq_handler) _irq_com2, INT_SPECIFIC);
-		break;
-#endif
-	    default:
-		request_irq(sp->irq, rs_irq, INT_GENERIC);
-		break;
-	    }
 	    sp->tty = tty;
 	    update_port(sp);
 	}


ghaerr:
Update: `ktcp` already uses the O_EXCL flag on open, and already properly errors when SLIP is specified on an in-use serial port. See screenshot:
Screen Shot 2022-07-30 at 12 03 18 PM

Mellvik:
Thank you @ghaerr,
That's good news about the O_EXCL, so we're really all set re. the net start problem.

As to the irq sharing, I'm trying to understand the goal: I should now be able to configure 4 ports with the same irq, right? And then use them one at a time timesharing that irq?
Like I said before, I don't think this is physically possible but I'll be happy to test it.

Are you committing this now or should I test the PR?
-M

ghaerr:
> As to the irq sharing, I'm trying to understand the goal

There's a couple goals. The first is that ELKS has had dynamic IRQ handling for over a year, and all drivers should use it, as a natural consequence of the way the ELKS kernel has been moving: dynamic allocation and use of resources. The serial driver now matches the NIC drivers in that way (i.e., the NIC IRQ vector is never actually grabbed until open, and it is released at close).

All ELKS drivers should have been written this way in the beginning, but it wasn't possible then since the IRQ trap handlers were manually coded in irqtab.S. 

The other goal is to be able to reuse a hardware IRQ if possible on a particular system. (If not, no harm done anyways).

More and more of the kernel is now written to allocate dynamically: pipe buffers, for instance, are only allocated when pipes are created (saving 2K of kernel data). We are considering the same for task structures (there are 16 fixed entries now).

> I should now be able to configure 4 ports with the same irq, right? 
> And then use them one at a time timesharing that irq?

Yes. Is that useful?

> Like I said before, I don't think this is physically possible but I'll be happy to test it.
Are you committing this now or should I test the PR?

I've committed it, since its tested on QEMU and is actually a small change. Of course I'm interested in hearing whether this allows better interoperation with various serial/NIC or other hardware. But there's no pressing need to do that; this change just moves the entire kernel approach one more step forward.

Mellvik:
BTW, for this testing and maybe in general it would be very helpful to be able to set the serial IRQs in bootopts, like
`SER_IRQ=,,10,7`
to modify IRQs for COM3 and 4.

-M

Mellvik:
The test report on interrupt sharing in https://github.com/jbruchon/elks/pull/1407#issuecomment-1218220421 was partly incorrect. The git branch in use did not have the latest version of `serial-8250.c` installed. 
- Correction: the NIC gets its requested IRQ even if a serial line has been configured with the same IRQ, but is currently closed.
- Update: Com ports sharing the same IRQ apparently prevent both from working correctly on input (i.e. no input)
  - port allocation/release in the driver works fine 
  - it doesn't matter whether the 2nd port is HW configured to the shared irq or not
 
More test planned, but the power supply caved in. 
To be continued.

-M

ghaerr:
> The git branch in use did not have the latest version of serial-8250.c installed.

I was thinking that, as that would explain everything that was reported.

> the NIC gets its requested IRQ even if a serial line has been configured with the same IRQ, but is currently closed.

I'm not sure I'm following you fully, but the "dynamic" allocation of IRQs in the current repo is first-come, first-served. That is, while the NIC and serial drivers will "report" their IRQ, the actual allocation of the IRQ to the driver occurs on a first-come, first served basis.

> Com ports sharing the same IRQ apparently prevent both from working correctly on input (i.e. no input)

I see: so you're saying that when two serial ports are specified to use the same IRQ, even though the IRQ (not port) allocation/release works fine on open/close, neither works? Not sure whether this is a HW issue, or something else. PR #1385 only changed the time that request_irq is called, no other serial driver changes except free_irq. (However the free_irq call could possibly not be working correctly). If that is the case, then only a single test serial port open should be attempted after boot, with reboots in-between to free the IRQ.

> More test planned, but the power supply caved in.
To be continued.

Geez, just when things were getting close lol!

Mellvik:
> the NIC gets its requested IRQ even if a serial line has been configured with the same IRQ, but is currently closed.
> 
> I'm not sure I'm following you fully, but the "dynamic" allocation of IRQs in the current repo is first-come, first-served. That is, while the NIC and serial drivers will "report" their IRQ, the actual allocation of the IRQ to the driver occurs on a first-come, first served basis.
> 
I'm just stating the obvious - in a somewhat convoluted way. IOW It works as expected.
> Com ports sharing the same IRQ apparently prevent both from working correctly on input (i.e. no input)
> 
> I see: so you're saying that when two serial ports are specified to use the same IRQ, even though the IRQ (not port) allocation/release works fine on open/close, neither works? Not sure whether this is a HW issue, or something else. PR #1385 only changed the time that request_irq is called, no other serial driver changes except free_irq. (However the free_irq call could possibly not be working correctly). If that is the case, then only a single test serial port open should be attempted after boot, with reboots in-between to free the IRQ.
> 
Yes, that's right. I'm pretty sure this is not a hardware issue though, since the two ports have different hw level irqs assigned. IOW there is more to be understood  before concluding. 
> More test planned, but the power supply caved in.
> To be continued.
> 
> Geez, just when things were getting close lol!
> 
They say 'life is - at best - unpredictable'.  So true. 
I'll find something in the stash... :-)

--M

Mellvik:
﻿It must be the irq-to-port mapping table.
Fix coming tomorrow. 

-M

[ iPhone ]

> 18. aug. 2022 kl. 22:12 skrev Helge Skrivervik ***@***.***>:
> ﻿
>> 
>> the NIC gets its requested IRQ even if a serial line has been configured with the same IRQ, but is currently closed.
>> 
>> I'm not sure I'm following you fully, but the "dynamic" allocation of IRQs in the current repo is first-come, first-served. That is, while the NIC and serial drivers will "report" their IRQ, the actual allocation of the IRQ to the driver occurs on a first-come, first served basis.
>> 
> I'm just stating the obvious - in a somewhat convoluted way. IOW It works as expected.
>> Com ports sharing the same IRQ apparently prevent both from working correctly on input (i.e. no input)
>> 
>> I see: so you're saying that when two serial ports are specified to use the same IRQ, even though the IRQ (not port) allocation/release works fine on open/close, neither works? Not sure whether this is a HW issue, or something else. PR #1385 only changed the time that request_irq is called, no other serial driver changes except free_irq. (However the free_irq call could possibly not be working correctly). If that is the case, then only a single test serial port open should be attempted after boot, with reboots in-between to free the IRQ.
>> 
> Yes, that's right. I'm pretty sure this is not a hardware issue though, since the two ports have different hw level irqs assigned. IOW there is more to be understood  before concluding. 
>> More test planned, but the power supply caved in.
>> To be continued.
>> 
>> Geez, just when things were getting close lol!
>> 
> They say 'life is - at best - unpredictable'.  So true. 
> I'll find something in the stash... :-)
> 
> --M

Mellvik:
It must be the irq-to-port mapping table.
Fix coming tomorrow. 

-M

[ iPhone ]

> 18. aug. 2022 kl. 22:12 skrev Helge Skrivervik ***@***.***>:
> 
> ﻿
>> 
>> the NIC gets its requested IRQ even if a serial line has been configured with the same IRQ, but is currently closed.
>> 
>> I'm not sure I'm following you fully, but the "dynamic" allocation of IRQs in the current repo is first-come, first-served. That is, while the NIC and serial drivers will "report" their IRQ, the actual allocation of the IRQ to the driver occurs on a first-come, first served basis.
>> 
> I'm just stating the obvious - in a somewhat convoluted way. IOW It works as expected.
>> Com ports sharing the same IRQ apparently prevent both from working correctly on input (i.e. no input)
>> 
>> I see: so you're saying that when two serial ports are specified to use the same IRQ, even though the IRQ (not port) allocation/release works fine on open/close, neither works? Not sure whether this is a HW issue, or something else. PR #1385 only changed the time that request_irq is called, no other serial driver changes except free_irq. (However the free_irq call could possibly not be working correctly). If that is the case, then only a single test serial port open should be attempted after boot, with reboots in-between to free the IRQ.
>> 
> Yes, that's right. I'm pretty sure this is not a hardware issue though, since the two ports have different hw level irqs assigned. IOW there is more to be understood  before concluding. 
>> More test planned, but the power supply caved in.
>> To be continued.
>> 
>> Geez, just when things were getting close lol!
>> 
> They say 'life is - at best - unpredictable'.  So true. 
> I'll find something in the stash... :-)
> 
> --M

Mellvik:
Here's the verdict after the `irq_port_map` table fix (populate the table in `open` instead of `init`):

When two COM ports share the same irq, only the first line opened after reboot will work. Further - on the SX (SBC) system, only the 'lower' com-port will ever work and only if it is the first to be opened after boot. IOW - say com2 and com3 has IRQ10:
- If com2 is the first to be opened after boot, it works fine.
- Closing com2 and opening com3 reallocates the IRQ in software, but com3 will never interrupt.
- Then closing com3 and reopening com2, will do the proper irq-reallocation in software, but it will never work.
- Opening com3 as the first of the two after boot, will not work (on the SX, but will on the compaq, see below).

I suspect the results may vary from one system to the next, and I'm sort of surprised it doesn't work on this particular system because the IRQs are assigned in the BIOS setup. OTOH - the BIOS setup only programs the multipurpose chips which are supposed to emulate the jumper configured originals.
Repeating the test on the compaq, with ISA cards and jumpers, the results were almost the same. The only difference that the second  line will work right after reboot.

--M

ghaerr:
Hello @Mellvik,

Thanks for the testing and comments, they help quite a bit!

> Here's the verdict after the irq_port_map table fix (populate the table in open instead of init):

Nice catch. Will you be pushing that into #1407?

WRT the IRQ reuse, did you try opening com2, closing it, then opening com2 again? I'm trying to understand if the problem is in free_irq or not. `free_irq` is new code, and could certainly be the problem.

It seems that the SX has hardware issues that may prohibit its reuse of any IRQ. However, the Compaq can use either com2 or com3 immediately after boot, but can't use the other com port afterwards - no interrupts, correct?

Depending on the question of whether com2 can be reused after closing or not, it seems that free_irq is not working. I should be able to duplicate this somehow on QEMU.

Thank you!

Mellvik:
Thank you @ghaerr.

> 
> Here's the verdict after the irq_port_map table fix (populate the table in open instead of init):
> 
> Nice catch. Will you be pushing that into #1407 <https://github.com/jbruchon/elks/pull/1407>?
> 
> 
Yes, this evening or tomorrow.
> WRT the IRQ reuse, did you try opening com2, closing it, then opening com2 again? I'm trying to understand if the problem is in free_irq or not. free_irq is new code, and could certainly be the problem.
> 
It seems to me free_irq is working. net stop/net start uses it and has been working for some time now. And restarting getty on com2 (or whatever com? really) works. My debug output from serial-850.c shows the irq being requested, released and requested again as expected any number of times.
> It seems that the SX has hardware issues that may prohibit its reuse of any IRQ. However, the Compaq can use either com2 or com3 immediately after boot, but can't use the other com port afterwards - no interrupts, correct?
> 
I wouldn't say that. Remember, this is outside any ISA specification so whether a certain behaviour is correct or not depends on the point of view. It's not surprising that there are slight differences in behaviour when the hardware is as different as in these two cases. Also, IIRC (haven't retested this now) releasing and reusing an IRQ between some NICs works OK on both platforms.
> Depending on the question of whether com2 can be reused after closing or not, it seems that free_irq is not working. I should be able to duplicate this somehow on QEMU.
> 
> 
I think we're at the end of the road on this one. For the record, I just tested irq sharing between a NIC and a serial line on the compaq. It boots with ne2k @ irq2 (9), then 'net stop', then starting getty on a com-port using that irq: Works fine.
Next step is killing the getty (debug messages confirming all releases and requests), then running 'net start' again, which seems to work, but it doesn't. No interrupts. So request_irq() works but the hardware doesn't play ball.

It seems com-ports have a more 'hard wired' relationship with their assigned IRQ lines than some NICs, as 
expected (or rather, suspected). 


-M

ghaerr:
> It seems com-ports have a more 'hard wired' relationship with their assigned IRQ lines than some NICs, as
expected (or rather, suspected).

Agreed.

> It seems to me free_irq is working. net stop/net start uses it and has been working for some time now. And restarting getty on com2 (or whatever com? really) works. My debug output from serial-850.c shows the irq being requested, released and requested again as expected any number of times.

Good news. However, a close code review of request_irq, int_handler_add, and free_irq show a couple problems, so this effort was not in vain - it appears free_irq has a memory leak: it doesn't free the previous int_handler struct used internally to point the vector at _irqit; this means that too many open/closes and the kernel would eventually run out of memory. I'm also seeing that should a request fail by running out of memory, interrupts are left off. PR coming to fix both :)

Mellvik:
> 
> Good news. However, a close code review of request_irq, int_handler_add, and free_irq show a couple problems, so this effort was not in vain - it appears free_irq has a memory leak: it doesn't free the previous int_handler struct used internally to point the vector at _irqit; this means that too many open/closes and the kernel would eventually run out of memory. I'm also seeing that should a request fail by running out of memory, interrupts are left off. PR coming to fix both :)
> 
I'll be darned. Great catch indeed. This is serious - and good. The law of unexpected consequences I guess. Including my power supply breakdown which revived a power supply  (I'm not joking) date-marked 12/82. ;-)
Happy to have contributed. We're on a good roll towards 0.7!

--M.

Mellvik:
> More and more of the kernel is now written to allocate dynamically: pipe buffers, for instance, are only allocated when pipes are created (saving 2K of kernel data). We are considering the same for task structures (there are 16 fixed entries now).
> 
Yes, this is very useful - in particular if it would lead to more process slots being available...
> I should now be able to configure 4 ports with the same irq, right?
> And then use them one at a time timesharing that irq?
> 
> Yes. Is that useful?
> 
No, it's more like a test case - proof of concept maybe. It's useful only if the cards are PnP (very unusual for ISA serial cards) and we find a way to support PnP without implementing the whole shebang (like on the 3c509 card).
As a test it's interesting though: 1) If 4 (say) ports are pulling the same IRQ line, can one of them still work? We know it works with 2 ports sharing the IRQ DOS style. 2) If it works, is it actually possible to alternate (timeshare) between them? The drivers may need some extra work to properly shut down in the close routine.
> Like I said before, I don't think this is physically possible but I'll be happy to test it.
> Are you committing this now or should I test the PR?
> 
> I've committed it, since its tested on QEMU and is actually a small change. Of course I'm interested in hearing whether this allows better interoperation with various serial/NIC or other hardware. But there's no pressing need to do that; this change just moves the entire kernel approach one more step forward.
> 
Yes, testing this is on my list, just after checking out the memopad - cool stuff! When I get out of the buffer bug rabbit hole.


Thank you!

—M

Issue #1366, opened 2022-07-12
New Cool Stuff for ELKS


ghaerr:
I'm opening this issue to talk about various projects I'm working on, most of which are running, but not fully integrated into the ELKS build yet. I'm interested to know how interested people might be in each. Here's the first three projects:
- ELKS remote console display
- D-Flat Text UI windowing toolkit
- Debug stack traces for ELKS executables

I'll write up each of these separately.

ghaerr:
ELKS Remote Console

This little program allows one to login through a serial or telnet connection to ELKS (using a macOS or Linux terminal program, for instance), and display what is currently displayed on the ELKS console. This could be useful to see remote kernel messages, or just see exactly what is displayed in full color, using the ROM BIOS 437 character set.

For example, here's a login from macOS Terminal over emulated serial on QEMU, just after boot:
<img width="752" alt="Screen Shot 2022-07-11 at 9 08 23 PM" src="https://user-images.githubusercontent.com/11985637/178402012-10757198-e90a-49f2-bfc5-d3cb203b7c01.png">

The display is updated every second, and the program is terminated with ^C.

The `cons` program reads the PC text RAM contents directly and converts the characters into UTF-8 encoded Unicode values, along with interpreting the PC character attributes (color, reverse video, etc) and translating those into ANSI escape sequences.

The current implementation just displays the console screen, it doesn't try to insert keystrokes, etc.

ghaerr:
D-Flat Text User Interface Library

D-Flat is an old (1995 era) text user interface that allowed overlapping window-based text user interfaces to be written on DOS. I've ported it to UNIX/Linux, and now to ELKS. It is very well written, but unfortunately kind of large, and requires medium model (large code) for compilation. The underlying API is message-based, and very similar to Win32, with SendMessage and window procedures. 

Here's a screenshot of the D-Flat `memopad` program, which allows for editing files:
<img width="832" alt="Screen Shot 2022-07-11 at 9 12 35 PM" src="https://user-images.githubusercontent.com/11985637/178402901-abff0747-653d-4519-8de4-13568e5efbe1.png">

I'm primarily working on it as a possible TUI for the Cosmopolitan project, but thought it would be interesting to get it running on ELKS.

Now to show off the `cons` program more, here's the cons program running on serial macOS Terminal, while `memopad` happens to be running on the console:
[...]

PR #1389, opened 2022-08-03
[kernel] Fix MINIX block allocation routine for out of space issues

ghaerr:
The MINIX filesystem new block allocation routine was incorrectly handling the MINIX ZMAP free blocklist bitmap, which resulted in incorrect identification of disk out of space, as well as improper "mnb: still zero bit!8192" messages. Both of these are fixed.

This was first identified by @Mellvik in https://github.com/jbruchon/elks/issues/1367#issuecomment-1203781403.

Rather than rewriting the routine from scratch, a revised version of the Linux v2.0 MINIX block allocation routine was used for replacement, then rewritten with map_buffer and unmap_buffer added for ELKS. In addition, the set_bit/clear_bit code was cleaned up.

Also, it was noticed that `cat` did not report filesystem out of space when writing. This is also fixed.

It was also noticed that in the prior block allocation routine, when the disk was "out of space", the mapped L1 buffers were not released. This would result in an L1 buffer being permanently allocated, which would have the result of possibly running out of L1 buffers at a later point in time. This has also been fixed.

Tested on QEMU.

@Mellvik, I recommend you update to this latest PR, then restart your testing for #1367.

### DIFF
--- a/elks/arch/i86/lib/bitops.c
+++ b/elks/arch/i86/lib/bitops.c
@@ -12,13 +12,13 @@ unsigned char clear_bit(unsigned int bit, register void *addr)
     flag_t flags;
     unsigned int mask;
 
-    addr = (void *)(((unsigned char *)addr) + (bit >> 3));
+    addr = (unsigned char *)addr + (bit >> 3);
     bit &= 0x07;
-    mask = (1 << bit);
+    mask = 1 << bit;
     save_flags(flags);
     clr_irq();
-    mask &= *((unsigned char *)addr);
-    *((unsigned char *)addr) &= ~mask;
+    mask &= *(unsigned char *)addr;
+    *(unsigned char *)addr &= ~mask;
     restore_flags(flags);
     return mask >> bit;
 }
@@ -28,21 +28,20 @@ unsigned char set_bit(unsigned int bit, register void *addr)
     flag_t flags;
     unsigned int mask, r;
 
-    addr = (void *)(((unsigned char *)addr) + (bit >> 3));
+    addr = (unsigned char *)addr + (bit >> 3);
     bit &= 0x07;
-    mask = (1 << bit);
+    mask = 1 << bit;
     save_flags(flags);
     clr_irq();
-    r = *((unsigned char *)addr) & mask;
-    *((unsigned char *)addr) |= mask;
+    r = *(unsigned char *)addr & mask;
+    *(unsigned char *)addr |= mask;
     restore_flags(flags);
     return r >> bit;
 }
 
 unsigned char test_bit(unsigned int bit,void *addr)
 {
-	return ( ((1 << (bit & 0x07))
-			  & *((unsigned char *)addr + (bit >> 3))) != 0);
+    return (((1 << (bit & 0x07)) & *((unsigned char *)addr + (bit >> 3))) != 0);
 }
 
 /* Ack... nobody even seemed to try to write to a file before 0.0.49a was
@@ -53,9 +52,9 @@ unsigned char test_bit(unsigned int bit,void *addr)
 
 #ifdef BLOAT_FS
 /* Use the old faithful version */
-unsigned int find_first_non_zero_bit(register int *addr, unsigned int len)
+unsigned int find_first_non_zero_bit(int *addr, unsigned int len)
 {
-    register char *bit = 0;
+    unsigned int bit = 0;
     unsigned int mask;
 
     do {
@@ -68,17 +67,17 @@ unsigned int find_first_non_zero_bit(register int *addr, unsigned int len)
 	    break;
 	}
 	addr++;
-    } while ((unsigned int)(bit += 16) < len);
-    if ((unsigned int)bit > len)
-	bit = (char *)len;
-    return (unsigned int)bit;
+    } while ((bit += 16) < len);
+    if (bit > len)
+	bit = len;
+    return bit;
 }
 #endif
 
 /* Use the old faithful version */
-unsigned int find_first_zero_bit(register int *addr, unsigned int len)
+unsigned int find_first_zero_bit(int *addr, unsigned int len)
 {
-    register char *bit = 0;
+    unsigned int bit = 0;
     unsigned int mask;
 
     do {
@@ -91,8 +90,8 @@ unsigned int find_first_zero_bit(register int *addr, unsigned int len)
 	    break;
 	}
 	addr++;
-    } while ((unsigned int)(bit += 16) < len);
-    if ((unsigned int)bit > len)
-	bit = (char *)len;
-    return (unsigned int)bit;
+    } while ((bit += 16) < len);
+    if (bit > len)
+	bit = len;
+    return bit;
 }
--- a/elks/fs/minix/bitmap.c
+++ b/elks/fs/minix/bitmap.c
@@ -62,22 +62,19 @@ unsigned short minix_count_free_inodes(register struct super_block *sb)
 void minix_free_block(register struct super_block *sb, unsigned short block)
 {
     register struct buffer_head *bh;
-    const char *s;
+    const char *s = NULL;
     unsigned int zone;
 
-    s = NULL;
-    if (!sb)
-	s = "bad dev";
-    else if (block < sb->u.minix_sb.s_firstdatazone ||
-	    block >= sb->u.minix_sb.s_nzones)
-	s = "trying to free block not in datazone";
+    if (!sb) return;
+    if (block < sb->u.minix_sb.s_firstdatazone || block >= sb->u.minix_sb.s_nzones)
+	s = "not in datazone";
     else {
 	bh = get_hash_table(sb->s_dev, (block_t) block);
 	if (bh) mark_buffer_clean(bh);
 	brelse(bh);
 	zone = block - sb->u.minix_sb.s_firstdatazone + 1;
 	bh = sb->u.minix_sb.s_zmap[zone >> 13];
-	if (!bh) s = "bad bitbuf";
+	if (!bh) s = "null zmap";
 	else {
 	    map_buffer(bh);
 	    if (!clear_bit(zone & 8191, bh->b_data))
@@ -86,8 +83,8 @@ void minix_free_block(register struct super_block *sb, unsigned short block)
 	    unmap_buffer(bh);
 	}
     }
-    if (s != NULL)
-	printk("free_block (%s:%u): %s\n", kdevname(sb->s_dev), block, s);
+    if (s)
+	printk("free_block: block %u %s\n", block, s);
 }
 
 block_t minix_new_block(struct super_block *sb)
--- a/elks/fs/minix/bitmap.c
+++ b/elks/fs/minix/bitmap.c
@@ -87,46 +87,50 @@ void minix_free_block(register struct super_block *sb, unsigned short block)
 	}
     }
     if (s != NULL)
-	printk("mfb (%s:%u): %s\n", kdevname(sb->s_dev), block, s);
+	printk("free_block (%s:%u): %s\n", kdevname(sb->s_dev), block, s);
 }
 
-block_t minix_new_block(register struct super_block *sb)
+block_t minix_new_block(struct super_block *sb)
 {
-    register struct buffer_head *bh;
-    block_t i, j, k;
-
+    struct buffer_head *bh = NULL;
+    block_t i, j;
     if (!sb) return 0;
 
-    i = 0;
-    do {
-	bh = sb->u.minix_sb.s_zmap[i];
-	map_buffer(bh);
-	j = (block_t) find_first_zero_bit((void *)(bh->b_data), 8192);
-	k = j + i * 8192 + sb->u.minix_sb.s_firstdatazone - 1;
-	if (k < sb->u.minix_sb.s_nzones) {
-	    if (set_bit(j, bh->b_data)) {
-			printk("mnb: already set %d %d\n", j, bh->b_data);
-			return 0;
-	    }
-	    if (j == (block_t)find_first_zero_bit((void *)(bh->b_data), 8192)) {
-			printk("mnb: still zero bit!%d\n", j);
-			return 0;
-		}
-	    mark_buffer_dirty(bh);
-	    unmap_buffer(bh);
-	    if (!k)
-		break;
-	    bh = getblk(sb->s_dev, k);
-	    map_buffer(bh);
-	    memset(bh->b_data, 0, BLOCK_SIZE);
-	    mark_buffer_uptodate(bh, 1);
-	    mark_buffer_dirty(bh);
-	    unmap_brelse(bh);
-	    return k;
-	}
-	unmap_buffer(bh);
-    } while (++i < sb->u.minix_sb.s_zmap_blocks);
-    return 0;
+repeat:
+    j = 8192;
+    for (i = 0; i < sb->u.minix_sb.s_zmap_blocks; i++) {
+        if (i > 0)
+            unmap_buffer(bh);
+        if ((bh = sb->u.minix_sb.s_zmap[i]) != NULL) {
+            map_buffer(bh);
+            if ((j = find_first_zero_bit((void *)bh->b_data, 8192)) < 8192)
+                break;
+        }
+    }
+    if (i >= sb->u.minix_sb.s_zmap_blocks || !bh || j >= 8192) {
+        if (bh) unmap_buffer(bh);
+        return 0;
+    }
+    if (set_bit(j, bh->b_data)) {
+        printk("new_block: bit already set %d", j);
+        unmap_buffer(bh);
+        goto repeat;
+    }
+    mark_buffer_dirty(bh);
+    unmap_buffer(bh);
+    j += i*8192 + sb->u.minix_sb.s_firstdatazone - 1;
+    if (j < sb->u.minix_sb.s_firstdatazone || j >= sb->u.minix_sb.s_nzones)
+        return 0;
+    if (!(bh = getblk(sb->s_dev, j))) {
+        printk("new_block: cannot get block %u", j);
+        return 0;
+    }
+    map_buffer(bh);
+    memset(bh->b_data, 0, BLOCK_SIZE);
+    mark_buffer_uptodate(bh, 1);
+    mark_buffer_dirty(bh);
+    unmap_brelse(bh);
+    return j;
 }
 
 void minix_free_inode(register struct inode *inode)
--- a/elkscmd/file_utils/cat.c
+++ b/elkscmd/file_utils/cat.c
@@ -14,8 +14,10 @@ static int copyfd(int fd)
 {
 	int n;
 
-	while ((n = read(fd, readbuf, sizeof(readbuf))) > 0)
-		write(STDOUT_FILENO, readbuf, n);
+	while ((n = read(fd, readbuf, sizeof(readbuf))) > 0) {
+		if (write(STDOUT_FILENO, readbuf, n) != n)
+			return 1;
+    }
 	return n < 0? 1: 0;
 }
 


ghaerr:
@Mellvik,

After spending considerable time trying to understand exactly why the old MINIX block allocate function (minix_new_block) didn't work, I have now come to fully understand the reason, and its bad: this bug is probably the worst kernel data corruption failure we've seen in a while. 

What was happening is that, whenever any *new* block was being allocated for any file, for any reason, on a MINIX filesystem (temp files, remove a file, recreate, etc etc), the routine attempts to find a new disk block, by inspecting each of up to eight 8192-bit (=1024 bytes) bitmaps, where a 0 bit indicates a free block. These 8192-bit maps are called "Zmaps" internally, are loaded from disk when the MINIX filesystem is mounted, and kept in kernel buffers. (More evidence of much heavier kernel buffer use than expected, BTW - each mounted MINIX filesystem uses a 1K kernel buffer for each 8M of disk size, plus a buffer for the super block).

The minix_new_block routine calls `find_first_zero_bit(..., 8192)` repeatedly eight times, looking for the first free block (up to max 64k bits, for max 64M filesystem). If the bits are all 1's, meaning no free blocks, the routine returns 8192. (This would be normal in a disk that's 1/4 or half full to have all 1 bits in the first Zmaps). The bug is that the kernel goes ahead and calls `set_bit` on that return value, regardless. Recall that 8192 bits (0-8191) will fit in one disk block (=one kernel buffer). When `set_bit` is called with 8192, it doesn't do range checking, and will set the bit *following* the last buffer data byte - that is, the byte in the NEXT L1 buffer! (The L1 buffer data is all continuous in an array of 12k). Total data corruption, but one may never know, depending on whether that next buffer is in use, number of buffers, XMS vs EXT, and exactly which buffers are currently mapped into L1.

Here's the old code in fs/minix/bitmap.c, to help you follow:
```
    i = 0;
    do {
      bh = sb->u.minix_sb.s_zmap[i];
      map_buffer(bh);
       j = (block_t) find_first_zero_bit((void *)(bh->b_data), 8192);  // <-- if all 1's, returns 8192
       k = j + i * 8192 + sb->u.minix_sb.s_firstdatazone - 1;
       if (k < sb->u.minix_sb.s_nzones) {
           if (set_bit(j, bh->b_data)) { // <-- tries to set bit regardless of 8192 error return value
                       printk("mnb: already set %d %d\n", j, bh->b_data); // <- if previous bit was 1, this error given
                       return 0;
           }
           if (j == (block_t)find_first_zero_bit((void *)(bh->b_data), 8192)) { // <-- this error given if previous bit was 0
                       printk("mnb: still zero bit!%d\n", j);
                       return 0;
               }
```
I thought you might appreciate this! Given this finding, we will definitely have to have you use updated code for your continued testing in #1367, as this could definitely be part of the problem.

ghaerr:
Hello @Mellvik,

Guess what? After committing this PR, I started looking further into fs/minix/bitmaps.c, and happened across a routine very similar to the `minix_new_block` routine, called `minix_new_inode`. This routine allocates new inodes (such as when executing `touch x` in your original buffer/fs issue #1367). The routine operates almost identically to the just-fixed Zmap block allocator, but uses the Imap (inode map) instead, and has EXACTLY the same bug!!!! That is, when an Imap bitmap is all ones, the `find_first_zero_bit` returns 8192, and `set_bit` is called, which corrupts data in the next buffer!! 

Thus, I'm almost certain we've now found at least one big problem behind #1367. I don't yet tonight have time for a full rewrite of this additional routine, but suggest you pull the latest commit, and then make the following change. This temp fix just handles the error case of all 1's 8192 return, doesn't call set_bit, and keeps looking for a 0 bit in the next Imap.

I'm not fully sure this is all of #1367, but we've now found two cases of buffer corruption, which could happen whenever a block is added to a file, or, any file (inode) is created. Both corrupt L1 data, which could explain why XMS and EXT buffers operate differently.

In file fs/minix/bitmap.c, make the following changes:
```
struct inode *minix_new_inode(struct inode *dir, __u16 mode)
{
    register struct inode *inode;
    register struct buffer_head *bh;
    /* Adding an sb here does not make the code smaller */
    block_t i, j, k;
    const char *s;

    if (!dir || !(inode = new_inode(dir, mode))) return NULL;
    minix_set_ops(inode);

    s = "No new inodes found!\n";
    i = 0;
    do {
        bh = inode->i_sb->u.minix_sb.s_imap[i];
        map_buffer(bh);
        j = (block_t) find_first_zero_bit((void *)(bh->b_data), 8192);
        if (j >= 8192) continue; // <----- ADD THIS LINE
        k = j + i * 8192;
        if (k < inode->i_sb->u.minix_sb.s_ninodes) {
            if (set_bit(j, bh->b_data)) {       /* shouldn't happen */
                unmap_buffer(bh);
                s = "mni: already set\n";
                break;
            }
```
I plan on writing and testing a replacement routine in the morning.

In order to fully test minix_new_inode, we need a script which creates many files, as this should trigger the bug. Another test would be to put a printk in the line before "continue", confirming the 8192 buggy return. 

Thank you!

ghaerr:
I did some more testing, and unfortunately, even though the minix_new_inode routine is buggy, the bug only shows up when the inode count is > 8192. Since that's almost never the case, there's not much point in adding the above line described above. I am still interested in seeing whether the first minix_new_block fix changes anything, though. We're getting closer!

Mellvik:
Wow @ghaerr - that's a lot of good stuff and good news.  Great work indeed.

I would have liked to do som more testing, but here's what I've got so far:
- As to the buffer/fs bug (that's where more testing is necessary) there is  some change in behaviour which I'd like to look closer at.
- At to the actual bug at hand it seems to be fixed. I cannot reproduce the errors from yesterday and 3M/5M files can be copied around w/o problems.
- Which brings me to the interesting part: Deleting my 10 level directory tree containing some of these 'big' files, hangs the system. Which must be the L1 bug you've been talking about (although from the limited debug prints from truncate it looks exactly like the buffer/fs problem). The `trE` is the entry to truncate(), `td0` is the zone loop in V1_trunc_direct().
- Also and incidentally I ran across this (see screen clip) issue with `find`.
- More testing coming up.

Skjermbilde 2022-08-04 kl  17 39 55

-M

Mellvik:
Additional comment:
- the hang is completely predictable (the number of directory levels is 12)
- `cd`ing halfway down the tree, and `rm -r` works fine.

ghaerr:
@Mellvik,

Thanks for the testing. Glad to hear the disk out of space bug is fixed, and I'm hoping that this helps us track down the fs/buffer bug. If the stack continues to be corrupted, then I'd suggest removing schedule() and adding a printk to wait_on, as discussed in https://github.com/jbruchon/elks/issues/1367#issuecomment-1204275021. There's no need to add lines as I discussed above, however.

Yes - the `rm -rf` fails on deep directory trees, I'm almost positive that's for the L1 issue. Lets keep that completely separate for now. I have another rm -rf bug that appears after using up all the inodes, which we can talk about later. Remember, we can verify an "L1 hang" because the system is operational (by logging in elsewhere), only the rm process is hung.

> the hang is completely predictable (the number of directory levels is 12)
cding halfway down the tree, and rm -r works fine.

Yes, that's the L1 hang symptom, and your td0/tRE loop shows it looping waiting for a deadlocked buffer situation. The reason for the hang is that each directory requires a locked, mapped buffer, it seems, and there's only 12 of them.

I've noted the `find` bug, probably needs more stack/heap size.

Thank you!

ghaerr:
> I've noted the find bug, probably needs more stack/heap size.

I can't reproduce the `find` stack overflow bug... let me know if you can reproduce on QEMU, thanks!

Mellvik:
Hi @ghaerr,

yes I can, it's actually reproducible cross platform.

Here's the trick (on HD even on QEMU, probably the same on floppy image, I haven't checked that).


cd /mnt (that's the HD) which contains a full root file system
cd tmp
mkdir -p a/b/c/d/e/f/g/h/i/j/k
cd ..
find . -name a -print


—M

ghaerr:
> Here's the trick (on HD even on QEMU, probably the same on floppy image, I haven't checked that).

Hi @Mellvik, I can't get that to repeat. Perhaps my copy of /mnt isn't exactly the same as yours, but tried on both FD and HD.

Could there be symlinks in /mnt that my cp -R didn't copy? Not sure...

Mellvik:
Hi @ghaerr, it turns out my example above is slightly (one level, taken from memory) off: you need 13 directory levels to trigger the error. Which admittedly makes the bug kind of esoteric, but still ...

Here's today's QEMU run:
Skjermbilde 2022-08-06 kl  11 51 38

And FWIW - it's repeatable on a floppy image with no HD mounted.

--M

Mellvik:
Hi @ghaerr,
I pulled the latest merge from upstream in order to test the new bitmap fixes - and also look for possible influences on the buffer/fs problem, and got some interesting results.

First of all,  I can now delete files. This may be temporary - I've had situations in the past in which I've reorganized (say) printk's and it worked seemingly OK for a while, until I rocked the boat sufficiently to get wet again. I'll keep rocking.

But here's the thing:
I'm copying a small file (say `bootopts`), then deleting the copy, and getting the output below (ignore the truncate() debug output). This is repeatable on the SX, not on QEMU.

```
login: root
elks17# cd ..
elks17# cp bootopts pp
elks17# rm pp
trEtd0td0free_block: block 12288 already cleared
td0free_block: block 163 not in datazone
td0td0free_block: block 256 not in datazone
td0td0trDtrItrXfree_inode: count 0
elks17#
elks17# cp bootopts pp
elks17# rm pp
trEtd0td0free_block: block 30720 already cleared
td0free_block: block 164 not in datazone
td0td0free_block: block 256 not in datazone
td0td0trDtrItrXfree_inode: count 0
elks17#
```

It may be possible - given that delete files now work - however temporarily, that the effects of the stack corruption have moved to the bitmap  - or is this otherwise explainable?  
I haven't looked at the code, and I'll continue to beat up this build to see if I can trigger the unlink hang again. If not I'll introduce the latest merge into a different test branch with a lot more debug output.

To be continued.

-M

ghaerr:
Hello @Mellvik,

Thanks for the continued testing. I have to say, very strange. I don't quite know whether to be worried or not: changing the MINIX free block (and inode) code is a big deal, and if incorrect, could open up a bunch of instability on ELKS if so.

However, it was proven that the prior code was buggy (in the long discussion prior in this thread). In addition, both routines were found rewritten in Linux 2.0, and now removing a file works (which I finally reproduced the failure on QEMU). So I'm thinking that the new code is working properly, especially since your latest testing doesn't show any errors on QEMU, only SX. Of course, the original problem only showed on the SX as well.

> It may be possible - given that delete files now work - however temporarily, that the effects of the stack corruption have moved to the bitmap - or is this otherwise explainable?

I'm beginning to wonder whether the SX may have a memory problem, perhaps a stuck bit? Admittedly, very strange that the SX was the only system that shows any of these errors, but OTOH this enabled you to find a major MINIX filesystem bug.

> free_block: block 163 not in datazone
> free_block: block 256 not in datazone

These are big-time errors: the meaning is that `free_block` is being called with a block number BELOW the start of the data area of the filesystem (i.e., in the inode area). Should never happen. This implies that the inode being deleted (freed) had these bad block numbers in the DIRECT BLOCK area of the inode. I think the best explanation of this is that the buffer containing the inode data is trashed, which then passes bad data to the free_block routine.

So at the end of the day, we're still experiencing data corruption, very likely inside kernel buffers (which was thought to be the original problem). This fits with the probleming going away when using XMS buffers. If the memory in the EXT buffer region is bad, then this would explain all of what we're seeing.

Can you run a DOS memory tester on the SX system?

Another thought is to either 1) write a bunch of buffer debug code (which is a lot of work if its just the SX system that has a hardware issue), or 2) perhaps write some kind of CRC-checking code for buffer contents, to see when their data is getting trashed. So running your test branch with more debug output is probably a good start.

Thank you!

Mellvik:
Thank you @ghaerr - very interesting indeed, maybe (optimistically)  a path to (finally) figure out what’s going on.
> 
> However, it was proven that the prior code was buggy (in the long discussion prior in this thread). In addition, both routines were found rewritten in Linux 2.0, and now removing a file works (which I finally reproduced the failure on QEMU). So I'm thinking that the new code is working properly, especially since your latest testing doesn't show any errors on QEMU, only SX. Of course, the original problem only showed on the SX as well.
> 
> It may be possible - given that delete files now work - however temporarily, that the effects of the stack corruption have moved to the bitmap - or is this otherwise explainable?
> 
> I'm beginning to wonder whether the SX may have a memory problem, perhaps a stuck bit?
> 
This cannot be ruled out. I’ve run memory tests in DOS in addition to the ‘internal’ (BIOS) mem test, and successfully run W95 on the machine, but still ...
> Admittedly, very strange that the SX was the only system that shows any of these errors, but OTOH this enabled you to find a major MINIX filesystem bug.
> 
> free_block: block 163 not in datazone
> free_block: block 256 not in datazone
> 
> These are big-time errors: the meaning is that free_block is being called with a block number BELOW the start of the data area of the filesystem (i.e., in the inode area). Should never happen. This implies that the inode being deleted (freed) had these bad block numbers in the DIRECT BLOCK area of the inode. I think the best explanation of this is that the buffer containing the inode data is trashed, which then passes bad data to the free_block routine.
> 
> So at the end of the day, we're still experiencing data corruption, very likely inside kernel buffers (which was thought to be the original problem). This fits with the probleming going away when using XMS buffers. If the memory in the EXT buffer region is bad, then this would explain all of what we're seeing.
> 
FWIW I’ve tested with EXT buffers from 30 to 150 w/o any effect.
And BTW - I have been playing with this other (somewhat far fetched) theory - that we have a memory alignment problem that triggers some odd behavior only on the SX, and goes away when XMS buffers are active because the references are 32 bit instead of 16.
> Another thought is to either 1) write a bunch of buffer debug code (which is a lot of work if its just the SX system that has a hardware issue), or 2) perhaps write some kind of CRC-checking code for buffer contents, to see when their data is getting trashed. So running your test branch with more debug output is probably a good start.
> 
I'm working on a setting in which I can pinpoint exactly a memory (stack) location (address) that gets corrupted. The making small mods from that in order to hopefully trace back to the source. It’s slow and tedious, but that’s the nature of such problems. And I’ll run another round of memory tests. Is there such a thing as a symbol table for the linked kernel? That would be helpful.

Getting there …

-M

ghaerr:
> Is there such a thing as a symbol table for the linked kernel? 

Oh yes - its at elks/arch/i86/boot/system.map. As mentioned in https://github.com/jbruchon/elks/issues/1366#issuecomment-1181285906, I've developed some nice stack backtrace and debug routines which depend on a map file for an application, but I haven't implemented a way for that to extend to the kernel itself. (The mechanism appends the map file to the application executable, then the application opens it and stores it in RAM on startup. The kernel would need to do the same, but the I/O is different).

I have a some scripts that sort the map file, and also produce symbols for the static functions and data, which are not shown in the .map file. I plan on sharing those also soon. They work by running `nm` on the ELF binary before the `elf2elks` conversion. @tkchia has added an option to delay the execution of `elf2elks` to ia16-elf-gcc, we haven't updated to that version of the source yet, but will, as soon as these other enhancements are published. Actually, I just realized that we do produce a complete map file for the kernel now automatically, it's in the same directory, named `system-nm.map`.

> I'm working on a setting in which I can pinpoint exactly a memory (stack) location (address) that gets corrupted. The making small mods from that in order to hopefully trace back to the source. It’s slow and tedious, but that’s the nature of such problems. 

I think this is a good track, given the results you're seeing. Keep us posted!

Mellvik:
Thank you @ghaerr, 


> I just realized that we do produce a complete map file for the kernel now automatically, it's in the same directory, named system-nm.map.

This is useful!! Apropos - what do the suffixes ! and & in the symbol table mean?

-M

tkchia:
Hello @Mellvik,

> This is useful!! Apropos - what do the suffixes ! and & in the symbol table mean?

Basically the `!` symbols are a way to kind of represent segment:offset pairs in the context of ELF, which only knows about linear addresses.  See Anvin's `segelf` scheme (https://git.zytor.com/users/hpa/segelf/abi.git/plain/segelf.txt).

For each symbol _foo_ there is a _foo_`!` symbol &mdash; the `!` is part of the symbol name &mdash; which gives the start address of the IA-16 segment containing _foo_.

So you may have e.g. `printk` = `0x12ee2` and `printk!` = `0x10000` which will mean that `printk` is at offset `0x2ee2` into its IA-16 segment.

(In the current linker script implementation, the `.text`, `.fartext`, and `.data` + `.bss` segments are simply placed at linear addresses `0x10000`, `0x20000`, and `0x30000` at link time, for simplicity.  Of course, at run time, the `.text`, `.fartext`, and `.data` + `.bss` segments will most probably be relocated to some other linear addresses.)

(The _foo_`&` symbols mark _ends_ of segments, and are not currently used.)

Thank you!

Mellvik:
Thank you, @tkchia,

that makes sense.

I looked through Anvin's document - very thorough indeed. But no mention of the '&' variant. Maybe you can enlighten me on that one too?


Thank you!


-M

ghaerr:
> Of course, at run time, the .text, .fartext, and .data + .bss segments will most probably be relocated to some other linear addresses.)

The segments used for .text, .fartext and .data can be known by referencing one of the ELKS startup messages, for instance:
```
Kernel text at 2d0:0000, ftext 11f7:0000, data 14f8:0000, top 9fc0:0, 491K free
```
The offsets will remain the same as shown in the .map file. Since the kernel is by default being built with CONFIG_FARTEXT_KERNEL=y, any comparison of the IP (instruction pointer) offset will need a check of the CS (code segment) value as well to determine the executing function.

tkchia:
Hello @Mellvik,

> I looked through Anvin's document - very thorough indeed. But no mention of the '&' variant. Maybe you can enlighten me on that one too?

The _foo_`&` symbol simply marks the end of the segment containing _foo_.  This is not really part of Anvin's proposal, and the _foo_`&` symbols are not actually used at the moment anyway.  You can safely ignore them for now. :slightly_smiling_face: 

(Anvin did mention elsewhere about the possibility of marking ends of segments.  This will mainly be useful e.g. for sizing up segments for 80286 protected mode operation.)

Thank you!

ghaerr:
> But no mention of the '&' variant.

As @tkchia mentioned, that variant describes the end of the corresponding segment, and is not used.

I usually run the .map file through `sed` to delete the "!" and "&" symbols, since they're not really needed:
```
cat system-nm.map | sed -e '/&/d; /!/d' | sort | more
```
This produces a much more readable map file.

ghaerr:
Hello @tkchia and @Mellvik,

I am currently working on an enhancement to `elf2elks` which will allow for an application or kernel map file to be produced, and added in to the MINIX a.out executable file (at the end of the file, just past the relocation information), using the a.out header syms longword to denote its offset.

This, combined with a debug and instrumentation library I've developed that allows an application to read its own symbol table, will allow stack traces to be displayed using the C or ASM symbol names, rather than their addresses. This currently works for the .text section. (.fartext and .data symbols coming later).

As a result of this conversation, I think the symbols could be added to the kernel a.out in exactly the same way, and then use code in setup.S to read the symbol table into main memory (outside the kernel data segment). After that, we should be able to build a "debug kernel" that will allow for display kernel stack traces with C/ASM symbols. When hooked to the keyboard interrupt, this would also allow for real-time stack traces to see what the kernel might be doing at any point in time.

@tkchia, I plan to shortly post the enhancement, for your critique and review.

Thank you!

Mellvik:
> 
> Hello @tkchia <https://github.com/tkchia> and @Mellvik <https://github.com/Mellvik>,
> 
> I am currently working on an enhancement to elf2elks which will allow for an application or kernel map file to be produced, and added in to the MINIX a.out executable file (at the end of the file, just past the relocation information), using the a.out header syms longword to denote its offset.
> 
> This, combined with a debug and instrumentation library I've developed that allows an application to read its own symbol table, will allow stack traces to be displayed using the C or ASM symbol names, rather than their addresses. This currently works for .text and .fartext sections. (.data symbols coming later).
> 
> As a result of this conversation, I think the symbols could be added to the kernel a.out in exactly the same way, and then use code in setup.S to read the symbol table into main memory (outside the kernel data segment). After that, we should be able to build a "debug kernel" that will allow for display kernel stack traces with C/ASM symbols. When hooked to the keyboard interrupt, this would also allow for real-time stack traces to see what the kernel might be doing at any point in time.
> 
> 
Thank you @ghaerr - this will be a very powerful addition - and obviously, I'm particularly interested in the kernel part right now.

Space permitting, it's always useful to have the kernel symbol table available. Possibly allowing the addition of something akin to the WCHAN (BSD) column in ps (-l) which displays the location at which a running process is currently waiting. For debugging it shows where a waiting process is stuck. 

I'm sure commands like nm and strip are also on your list. This is 'professionalizing' ELKS!

Very much looking forward to it!

—M

ghaerr:
> Possibly allowing the addition of something akin to the WCHAN (BSD) column in ps (-l) which displays the location at which a running process is currently waiting. For debugging it shows where a waiting process is stuck.

It would be pretty straightforward to add the wait channel now to `ps`. Of course, with a kernel symbol table, we might be able to display the function the kernel has suspended. This is a bit more complicated than it seems, since the kernel actually waits on a data address, and `ps` would have to attempt a stack backtrace on a changing kernel stack, looking for the return address from the sleep call. But it could probably be figured out.

With regards to having the kernel map file around, it looks like at this point, it'll only take up around ~20-22k of space, not much really. And it doesn't necessarily have to reside in RAM, unless the kernel itself is going to use it. For instance, `ps` could open and read /linux to get the kernel symbol table, even when its not present in RAM. Lots of exciting possibilities.

> I'm sure commands like nm and strip are also on your list.

Actually, both of these are already available as `ia16-elf-nm` and `ia16-elf-strip`. However, they only work on ELF format, which is limited to .o files, as the `elf2elks` program rewrites the ELF executable into the older MINIX a.out format. Also, at this point, there's nothing to strip.

Mellvik:
Thank you @ghaerr.
> 
> Possibly allowing the addition of something akin to the WCHAN (BSD) column in ps (-l) which displays the location at which a running process is currently waiting. For debugging it shows where a waiting process is stuck.
> 
> It would be pretty straightforward to add the wait channel now to ps. Of course, with a kernel symbol table, we might be able to display the function the kernel has suspended. This is a bit more complicated than it seems, since the kernel actually waits on a data address, and ps would have to attempt a stack backtrace on a changing kernel stack, looking for the return address from the sleep call. But it could probably be figured out.
> 
Yes, I see the challenge in this. Out of curiosity, I'll take a look and see how BSD is doing this.
> With regards to having the kernel map file around, it looks like at this point, it'll only take up around ~20-22k of space, not much really.
> 
Sounds reasonable. On BSD2.11 it's about 35k, or ~30% up, 122k stripped, 157k w/symbols.
> And it doesn't necessarily have to reside in RAM, unless the kernel itself is going to use it. For instance, ps could open and read /linux to get the kernel symbol table, even when its not present in RAM. Lots of exciting possibilities.
> 
Indeed - it seems to me loading the symbol table into ram would be the exception. BSD ps reads symbols from the kernel file.

Then again, for now stack tracing is the thing! Exciting.


—M

Mellvik:
Hi @ghaerr - apologies for slow responses, I've been short on 'ELKS-time' these past few days.

This is interesting indeed. I'll look into whether the file system size or content may have something to do with it - or not. ASAP

-M

PR #1394, opened 2022-08-07
[elksemu] fix: buffer overflow in ELKS libc's isatty(.) due to Linux `struct termios` being larger than ELKS `struct termios`

tkchia:
I am a bit surprised that this issue has not cropped up before.  It was causing `elksemu` to crash, when the [`build-ia16/build.sh` script](https://github.com/tkchia/build-ia16/blob/62c1442e70277b4c6f2c4a427a8f9ab4907c6bd3/build.sh#L516) tried to run a relatively simple "hello world" program under `elksemu`.

### DIFF
--- a/elksemu/elks_sys.c
+++ b/elksemu/elks_sys.c
@@ -15,6 +15,7 @@
 #include <time.h>
 #include <signal.h>
 #include <errno.h>
+#include <stddef.h>
 #include <sys/types.h>
 #include <sys/resource.h>
 #include <sys/wait.h>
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

> I am a bit surprised that this issue has not cropped up before.

This is probably my fault - a couple of weeks ago in #1375, I cleaned up struct termios and removed the two unused elements (which had no definitions for indexing `c_cc[]` at all), and then reduced NCCS accordingly. (Also, the position for VEOL2, unused in ELKS, was replaced by VERASE2, which allows for two backspace characters, solving a long-standing use issue between the console and serial terminal emulators).

Nonetheless, with ELKS not having c_ispeed and c_ospeed, which I hadn't noticed, your change is necessary.

Please go ahead and commit, and thank you for the debugging!

tkchia:
Hello @ghaerr,

Ah, that explains it.  Thanks!  :slightly_smiling_face:

PR #1395, opened 2022-08-09
[tools][debug] Add symbol table to ELKS a.out executable files

ghaerr:
Adds symbol table information to ELKS a.out executable files, discussed in https://github.com/jbruchon/elks/pull/1389#issuecomment-1207433253.

The ELKS symbol table file format is newly designed, and the length of the symbol table is kept in the `syms` entry of the minix_exec_hdr struct. The symbol table itself directly follows the relocation entries, and is thus at the end of the file. Symbol table entries aren't automatically loaded into memory by the ELKS exec loader, but some new library code allows an application that wants to access its own symbol table to automatically load the symbol table from itself.

A new host tool `nm86` is created which will list the symbol table of the executable file passed as an argument. The tools and library code is currently in the elkscmd/debug folder, and running "make" in that directory will create an ELKS `testsym` executable which uses the library code to list some of its own symbols.

`elf2elks` was enhanced to take a `--symtab` option, which, when run after a linker run with the new `-mno-post-link` option, will add the .text and .data symbols (both normal and static) to the executable output file. There is not (yet) an option to do this directly from `ia16-elf-gcc`.

Using the new `-mno-post-link` option requires an upgrade to the ia16-elf-gcc compiler: I recommend renaming the elks/cross directory for backup, then running `tools/build.sh` to build the new compiler.

The current implementation in `elf2elks` runs a piped shell command using `popen` and `ia16-elf-nm`, rather than using the elftoolchain library directly to generate the symbol table.

The library code is a first pass and works well, but still calls some heavyweight functions, like sprintf, which are planned to be removed in the future, allowing for a lightweight debug library without dragging in stdio.

The next step will be adding stack backtrace routines to the elkscmd/debug library, which will allow for producing realtime backtraces with symbol names rather than addresses, and after that, adding instrumentation functions that will allow a realtime function trace of the program while running.


elks/tools/elf2elks/elf2elks.c
```diff
@@ -527,6 +531,60 @@ input_for_header (void)
...
+create_symtab (void)
+{
+    FILE *infp;
+    FILE *outfp;
+    char command[256];
+    char buf[128];
+
+    if (!symtab) return;
+
+    snprintf(symtab_filename, sizeof(symtab_filename), "%s-symXXXXXX", file_name);
```
tkchia: Since you are going to plug in the temporary file name into a command line for `popen`, and `file_name` might potentially contain unsafe characters, perhaps it might be better to exclude `file_name` from the file name template?  I.e. something like `"symXXXXXX"` rather than `"%s-symXXXXXX"`.

elks/tools/elf2elks/elf2elks.c
```diff
@@ -527,6 +531,60 @@ input_for_header (void)
...
+    if (!infp)
+        error("Can't popen '%s'\n", command);
+    int lastwastext = 0;
+    while(fgets(buf, sizeof(buf), infp) != NULL) {
+        int addr;
+        unsigned char type;
+        char name[128];
+
+        /* 0x10000+ is text, 0x20000+ is far text 0x30000+ is data */
+        if (sscanf(buf, "%x %c %s", &addr, &type, name) == 3) {
```
tkchia, line 584: The hard-coded linear addresses are kind of an implementation detail in the linker script.  I think it would be better to use the addresses that are actually provided in the intermediate ELF file &mdash; namely, `text_sh->sh_addr`, `ftext_sh->sh_addr`, and `data_sh->sh_addr` &mdash; similar to what `elf2elks` already does.

The `->sh_addr` values should already be available at this point in time, since `input_for_header ()` has already been called.

tkchia (approved):

tkchia:
Hello @ghaerr,

Thank you!  I have submitted some comments on your `elf2elks` changes, for your consideration.  But you might want to proceed to merge your changes first if you prefer. :slightly_smiling_face:

elks/tools/elf2elks/elf2elks.c
ghaerr: Ok. It was a little tricky, given that there may be cases where data_sh and/or bss_sh could be NULL. Does the arithmetic look correct for calculating the data_start and data_end values?

ghaerr:
Hello @tkchia,

Thank you for your suggestions, I have incorporated them. As mentioned above, it was a bit tricky to determine the data start and end locations given that either .data or .bss could be missing. I haven't yet tested with those variations, but will during more testing. I've also added .fartext processing, and `elf2elks --symtab` now properly produces a symbol table for the kernel `system` file, which is correct. (Note, .fartext symbols are stored as type 'F' and 'f' in the symbol table to differentiate them from .text, since the addresses are all 16-bit).

Thank you!

elks/tools/elf2elks/elf2elks.c
tkchia: Hello @ghaerr,

Thanks &mdash; looks OK for now.  On second thought, probably a better way to gauge the segment that a symbol _foo_ belongs to, would be to consult its corresponding _foo_`!` symbol.  But this will probably require maintaining more complex data structures (or perhaps an `awk` script).  Perhaps this part is best left as an exercise for the future. :no_mouth: 

Thank you!

tkchia:
@ghaerr : Cool, thanks!

PR #1398, opened 2022-08-11
[tools] Add --symfile option to elf2elks, produce system.sym symbol table

ghaerr:
Adds `--symfile file` option to `elf2elks`, which is then used when building the `system` kernel to produce a `system.sym` (~15k) file that will later be used to supply symbols for stack traces or disassembly. This will likely be copied onto distribution media as /lib/system.sym in the future, for use with the debug library.

Renames `system.map` to `system-full.map` (~180k).
Renames `system-nm.map` to `system.map`, with `foo!` and `foo&` symbols removed, for easy reading (~25k).

### DIFF
--- a/elks/tools/elf2elks/elf2elks.c
+++ b/elks/tools/elf2elks/elf2elks.c
@@ -85,8 +85,11 @@ struct minix_reloc
 static const char *me;
 static bool verbose = false, tiny = false, romable = false;
 static bool symtab = false;
+static bool symfile = false;
 static const char *file_name = NULL;
 static char *tmp_file_name = NULL;
+static char symtab_filename[256];
+
 static uint16_t total_data = 0, chmem = 0, stack = 0, heap = 0, entry = 0,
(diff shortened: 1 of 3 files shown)


elks/tools/elf2elks/elf2elks.c
```diff
@@ -243,6 +248,12 @@ parse_args (int argc, char **argv)
 		tiny = true;
 	      else if (strcmp (arg + 2, "symtab") == 0)
 		symtab = true;
+	      else if (strcmp (arg + 2, "symfile") == 0)
+		{
+		    symfile = true;
+		    ++i;
+		    strcpy(symtab_filename, argv[i]);
```
tkchia, line 257: Minor nit: You might want to check that `i < argc` still, i.e. there is actually a file name after `--symfile`.

tkchia (approved):

elks/tools/elf2elks/elf2elks.c
ghaerr: Thank you, fixed with error message.

PR #1399, opened 2022-08-11
[cmds] Add disasm 8086 disassembler tool for ELKS runtime

ghaerr:
Adds `disasm`, a standalone disassembler, to ELKS.

`disasm` takes a command line argument just like `hd` for inspecting memory, and can disassemble any location in memory, including itself. For this initial test, when the CS value equals its own code segment, it uses its own internal symbol table for display of call addresses as symbols. Currently only .text symbols are converted (not .data/.bss yet).

Usage:
```
disasm seg:off[#size]
```
The optional size field requests how many lines to be disassembled (default 22, or until RET instruction if size specified as 0).
 
Currently, `disasm` displays its own CS value on startup, so that one can test the mnemonic symbol display when disassembling its own location while running (here's a display of disassembling itself from _start):
```
# ./disasm 56ae:0
CS = 56ae
Disassembly of _start:
56ae:0000  55                push	%bp
56ae:0001  89 e5             mov	%sp,%bp
56ae:0003  8d 5e 04          lea	0x04(%bp),%bx
56ae:0006  8b 4e 02          mov	0x02(%bp),%cx
56ae:0009  89 c8             mov	%cx,%ax
56ae:000b  40                inc	%ax
56ae:000c  d1 e0             shl	$1,%ax
56ae:000e  01 d8             add	%bx,%ax
56ae:0010  a3 f4 07          movw	%ax,(0x07f4)
56ae:0013  50                push	%ax
56ae:0014  53                push	%bx
56ae:0015  8b 1f             mov	(%bx),%bx
56ae:0017  89 1e f2 07       movw	%bx,(0x07f2)
56ae:001b  51                push	%cx
56ae:001c  e8 ab 00          call	main (00ca)
56ae:001f  83 c4 06          add	$0x6,%sp
56ae:0022  50                push	%ax
56ae:0023  e8 01 00          call	exit (0027)
56ae:0026  cc                int	$3
56ae:0027  55                push	%bp
56ae:0028  89 e5             mov	%sp,%bp
56ae:002a  8b 1e 4c 08       movw	(0x084c),%bx
```
Shortly, the ability to load the kernel symbol file `symbol.sym` will be added, which will allow disassembly of the operating kernel with symbols.

`hd` was enhanced to check for overflow in its segment:offset arguments.

@Mellvik, this tool should help you debug the BIOS problem in #1367. As mentioned previously, the toolchain needs to be upgraded using `tools/build.sh`, and then run `make` in elkscmd/debug. `disasm` will be copied to the ELKS root user directory for the time being.

### DIFF
--- a/elkscmd/debug/.gitignore
+++ b/elkscmd/debug/.gitignore
@@ -1,3 +1,4 @@
 *.o
 nm86
 testsym
+disasm
--- a/elkscmd/debug/Makefile
+++ b/elkscmd/debug/Makefile
@@ -13,9 +13,8 @@ HOSTCFLAGS = -O3
 
 ###############################################################################
 
-PRGS = testsym
-OBJS = testsym.o syms.o stacktrace.o printreg.o
-OBJS += disasm.o
+PRGS = testsym disasm
+LIBOBJS += syms.o stacktrace.o printreg.o disasm.o
 #OBJS += ulltostr.o
 CFLAGS += -fno-optimize-sibling-calls
 CFLAGS += -fno-omit-frame-pointer
@@ -27,10 +26,17 @@ all: nm86 $(PRGS)
 nm86: nm86.c syms.c
 	$(HOSTCC) $(HOSTCFLAGS) -o $@ $^
 
-testsym: $(OBJS)
-	$(LD) $(LDFLAGS) -mno-post-link -o $@ $(OBJS) $(LDLIBS)
+TESTSYMOBJS = testsym.o $(LIBOBJS)
+testsym: $(TESTSYMOBJS)
+	$(LD) $(LDFLAGS) -mno-post-link -o $@ $(TESTSYMOBJS) $(LDLIBS)
 	elf2elks --symtab --heap 12000 $@
 	./nm86 $@ > $@.map
+	cp $@ $(TOPDIR)/elkscmd/rootfs_template/root
+
+DISASMOBJS = dis.o $(LIBOBJS)
+disasm: $(DISASMOBJS)
+	$(LD) $(LDFLAGS) -mno-post-link -o $@ $(DISASMOBJS) $(LDLIBS)
+	elf2elks --symtab --heap 12000 $@
 	cp $@ $(TOPDIR)/elkscmd/rootfs_template/root
 
 disasm.o: disasm.c
--- a/elkscmd/debug/dis.c
+++ b/elkscmd/debug/dis.c
@@ -1,0 +1,58 @@
+/*
+ * ELKS 8086 Disassembler
+ *
+ * Aug 2022 Greg Haerr
+ */
+#include <stdio.h>
+//#include <string.h>
+//#include <stdlib.h>
+//#include <unistd.h>
+#include "syms.h"
+#include "disasm.h"
+
+char * noinstrument getsymbol(int seg, int offset)
+{
+    static char buf[32];
+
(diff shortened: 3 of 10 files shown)


Mellvik:
Incredible @ghaerr,

Much appreciated!!

-M

[ iPhone ]

Mellvik:
@ghaerr, I'm having trouble building the updated tools to get `disasm` going.

Seemingly, the file `libquadmath.info` is missing, and I get quite a few errors about that towards the end of the build:
like this 
```
true  DO=all multi-do # /Library/Developer/CommandLineTools/usr/bin/make
true  DO=install multi-do # /Library/Developer/CommandLineTools/usr/bin/make
 ../../../../gcc-src/libquadmath/../install-sh -c -d '/Users/helge/src/mellvik/elks/cross/share/info'
 install-info --info-dir='/Users/helge/src/mellvik/elks/cross/share/info' '/Users/helge/src/mellvik/elks/cross/share/info/libquadmath.info'
install-info: No such file or directory for /Users/helge/src/mellvik/elks/cross/share/info/libquadmath.info
true  DO=all multi-do # /Library/Developer/CommandLineTools/usr/bin/make
true  DO=install multi-do # /Library/Developer/CommandLineTools/usr/bin/make
 ../../../../gcc-src/libquadmath/../install-sh -c -d '/Users/helge/src/mellvik/elks/cross/share/info'
 install-info --info-dir='/Users/helge/src/mellvik/elks/cross/share/info' '/Users/helge/src/mellvik/elks/cross/share/info/libquadmath.info'
install-info: No such file or directory for /Users/helge/src/mellvik/elks/cross/share/info/libquadmath.info
```
... which keeps looping for a while, apparently trying 'higher' directories.

This might not be fatal, but the linker script doesn't seem to be created:
```
ia16-elf-gcc -fno-inline -melks-libc -mcmodel=small -mno-segment-relocation-stuff -mtune=i8086 -Wall -Os -mno-post-link -o testsym testsym.o syms.o stacktrace.o printreg.o disasm.o 
ia16-elf-gcc: error: unable to locate default linker script ‘elks-small.ld’ in the library search paths
/Users/helge/src/mellvik/elks/cross/lib/gcc/ia16-elf/6.3.0/../../../../ia16-elf/bin/ld.gold: error: cannot open : No such file or directory
/Users/helge/src/mellvik/elks/cross/lib/gcc/ia16-elf/6.3.0/../../../../ia16-elf/bin/ld.gold: error: cannot find crt0.o
/Users/helge/src/mellvik/elks/cross/lib/gcc/ia16-elf/6.3.0/../../../../ia16-elf/bin/ld.gold: error: cannot find -lc
testsym.o:function getsymbol: error: undefined reference to 'sprintf'
testsym.o:function disassemble: error: undefined reference to 'printf'
testsym.o:function z: error: undefined reference to 'puts'
...
...

```

Advice needed on how to proceed ...

--M

ghaerr:
Hello @Mellvik,

After the tools rebuild, did you `make clean` and rebuild all of ELKS, or just run `make` in the elksmd/debug directory? You'll need to do the former. If that doesn't work, perhaps @tkchia has a comment about the tools build error messages.

Thank you!

Mellvik:
Aaaah,
I certainly did the latter - being in a rush.

A rerun per your recipe fixed it, thanks!
And we're still good on MacOS Monterey.


-M

PR #1400, opened 2022-08-11
[boot] don't assume bootloader starts with cs = 0, even if targeting IBM PC-compatible platform

tkchia:
Might address https://github.com/jbruchon/elks/issues/1367 .

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -65,7 +65,8 @@
 
 	.text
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Thank you @tkchia for this and your comments and work debugging #1367. Since this fixes the case where CS != 0 for any system, go ahead and commit now, and we'll stay tuned with @Mellvik's continued testing.

PR #1402, opened 2022-08-13
[boot] Disable whole-track reads for PC-98, for now

tkchia:
We probably need to deal with the NEC PC-98's F2DD_POINTER (http://www.webtech.co.jp/company/doc/undocumented_mem/memsys.txt) to properly configure whole-track reads on that platform.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -40,7 +40,13 @@
 //
 // Whole-track reads are disabled for the FAT bootloader because we are a
 // bit short of code space :-(
-#ifndef BOOT_FAT
+//
+// Whole-track reads are also disabled for the NEC PC98 platform for now;
+// it seems that, instead of a DDPT, we need to frob the PC98's F2DD_POINTER
+// (http://www.webtech.co.jp/company/doc/undocumented_mem/memsys.txt)
+// mechanism for configuring the "end of track" value; we will need to
+// somehow figure out this mechanism and test it :-( :-(
+#if !defined BOOT_FAT && !defined CONFIG_ARCH_PC98
 #   define FAST_READ
 #else
 #   undef FAST_READ


tyama501:
Thank you @tkchia ,

FAST_READ is not supported for now.
(I haven't touched any codes under ifdef FAST_READ ...)

tkchia:
Hello @tyama501,

Side question: would you like to try implementing whole-track reads in the ELKS bootloader, for the PC-98 platform?  Such a feature might allow the bootloader to load up the ELKS kernel more quickly &mdash; at least for non-FAT filesystems &mdash; on your machine.  :slightly_smiling_face: 

Thank you!

tyama501:
Hello @tkchia ,

I will study codes for FAST READ :)
I haven't even read boot codes under MINIX filesystem. PC-98 only support FAT for booting.

Thank you!

ghaerr:
Hello @tkchia, Hello @tyama501,

The FAST_READ boot code, along with the bioshd.c CONFIG_TRACK_CACHE code, enable multi-sector reads by passing a sector count greater than 1 to the BIOS disk read function, but rely on the apparently IBM PC-only DDPT array "max sectors" field in order to prohibit the BIOS from skipping to the next track, by the boot code and the bioshd.c driver supplying a larger-than-default max sectors field into a replacement DDPT vector, pointed to by INT 1Eh.

But the PC-98 is configured for CONFIG_TRACK_CACHE, and hasn't any problems. Thus, I don't quite yet understand how this might be working (if the PC-98 BIOS does not read the DDPT array, but instead a F2DD_POINTER).

> I haven't even read boot codes under MINIX filesystem. PC-98 only support FAT for booting.

Considering that no one has yet tried creating a MINIX boot disk in PC-98 format, could there be the possibility that it might boot with FAST_READ? Or am I misunderstanding the problem? (I suppose that it is only chance that is allowing the current multi-sector reads to work in bioshd.c, and that it needs the NEC F2DD_POINTER modification too?)

Thank you!

ghaerr:
I might add that the majority of my PC-98 floppy testing has been on the FD1232-format floppy (CHS 77/2/8), instead of the 1440K floppy (CHS 80/2/18), although both appear to run fine. Both `pc98-1232.config` and `pc98-1440.config` have CONFIG_TRACK_CACHE=y set.

tyama501:
Hello @ghaerr ,

Yes, the track cache is working with PC-98.
I don't understand the relation of the bioshd.c and the boot code but the num sector is translated to the number of bytes to read/write for the parameter of int1B disk BIOS in bioshd.c.

tyama501:
At least the translation for number of bytes to read for the multi sector is needed in boot code too.

tkchia:
Hello @ghaerr, hello @tyama501,

Hmm.  On IBM PC-compatible platforms at least, the problem is that
  * if the "end of track" setting in the drive parameter table is, say, 9,
  * and we try to read sectors 1&mdash;12 at one go from track 0, side 0,
  * then the BIOS will (wrongly) end up doing a "multi-track" read: it will read sectors 1&mdash;9, then move to reading sectors 1&mdash;3 from side 1 (but we really want sectors 10&mdash;12 from side 0).

Thank you!

ghaerr:
> On IBM PC-compatible platforms at least, the problem is that...

Yes. I am guessing, but it seems that there are two things going on here. @tkchia has pointed out that the DDPT is apparently replaced by an undocumented F2DD_POINTER mechanism, and that the DDPT may be unused by PC-98. And, that CONFIG_TRACK_CACHE, which depends on multi-sector reads working until end-of-track, seem to be working properly on PC-98 (despite the DDPT code in both the boot sector and bioshd.c driver).

If so, then it would seem that 1) all DDPT code could possibly be removed from the boot sector for PC-98 (possibly freeing up space for a multi-sector read for FAT boot), 2) the PC-98 BIOS, unlike the IBM PC BIOS, may be automatically handling multi-sector reads for floppies without requiring an updated DDPT (or even possibly an application F2DD_POINTER mechanism). This also could mean that a PC-98 FAST_READ could already work with FAST_READ set (and possibly DDPT code removed).

tkchia:
Hello @ghaerr, hello @tyama501,

  * Do things still work correctly when you insert a different type of floppy into the system, after ELKS has started?  I am thinking of oddball cases like, say, booting up with a FD1232 disk, then later trying to read a 1.44MiB disk.
  * Any idea what exactly is present at `int 0x1e` on a PC-98 system?  I have not found any documentation on this, and (unfortunately) I do not have a PC-98 system to test stuff with.

It will be good, I guess, if we can "simply" remove the DDPT code, but I think it will be prudent to check that things will (still) work correctly.

Thank you!

ghaerr:
Hello @tkchia,

> I am thinking of oddball cases like, say, booting up with a FD1232 disk, then later trying to read a 1.44MiB disk.

Yes, that could certainly trigger the kinds of problems the DDPT is required for. However, note that swapping disks like this even with ELKS and the working DDPT will still fail, as we don't request nor get notified by the BIOS of any disk changes, except when the block device is opened.

> I have not found any documentation on this, and (unfortunately) I do not have a PC-98 system to test stuff with.

The [DosBox-X](https://dosbox-x.com) project works well for me to test and run the ELKS PC-98 kernel and applications. I use the SDL2 version, and the following shell script wrapper to boot and ELKS FAT disks (built using `pc-1232.config` or `pc-1440.config`):
```
$ cat dosbox
exec ./dosbox-x-sdl2 -set machine=pc98 -set "dosbox quit warning=false" -fastlaunch -c "boot image/fd1232-pc98.img"
$ cat dosbox-x.conf
[dosbox]
debug=true

[sdl]
output=bitmap
doublescan=true

[cpu]
cputype=286

```
Running `hd 0:78` should then accurately display what INT 1Eh vector has been set to. DosBox-X seems to do a terrific job of emulating almost all of the PC-98 system, including graphics. `disasm` should also work with seg:off arguments to disassemble various pieces of the running system.

Thank you!

tyama501:
Hello @ghaerr ,

〉This also could mean that a PC-98 FAST_READ could already work with FAST_READ set (and possibly DDPT code removed).

As I said above this is no.
int1B requires number of bytes to read NOT number of sectors. So we need to calculate it in boot codes.

For bioshd.c I supported it in C codes.

I think I also supported disk swapping.

tyama501:
〉then the BIOS will (wrongly) end up doing a "multi-track" read: it will read sectors 1—9, then move to reading sectors 1—3 from side 1 (but we really want sectors 10—12 from side 0).

Oh, I didn't know we really want sectors 10-12 from side 0.
I don't know PC-98 can do this or not.

tyama501:
Where is the int1E used?
I couldn't find in boot_sect.S nor Setup.S.

tkchia:
Hello @tyama501,

> Where is the int1E used?

The [`int 0x1e`](https://www.delorie.com/djgpp/doc/rbinter/id/55/24.html) vector is at the address `0:(4 * 0x1e)` = `0:0x78`, so both `boot_sect.S` and `setup.S` read it from there.  In particular in `boot_sect.S`, I say
```
#ifdef FAST_READ
        ...
        mov $0x78,%bx  // 0:78h (INT vector 1Eh)
        lds (%bx),%si  // ds:si = BIOS original table
        push %ds
        push %si
        push %cx       // push address 0:78h (cx = 0)
        push %bx
        ...
#endif
```
The table is copied to RAM (possibly from ROM) so that we can later do
```
#ifdef FAST_READ
        ...
        mov sect_max,%al
        mov %al,floppy_table + 4
#endif
```
Just before `boot_sect.S` transfers control to the kernel, it restores the BIOS's drive parameter table pointer:
```
.macro RESTORE_DDPT
#ifdef FAST_READ
        mov $-8,%sp
        pop %bx
        pop %ds
        popw (%bx)
        popw 2(%bx)
#endif
.endm
```
Thank you!

tyama501:
Thank you @tkchia ,

I see.
There is no int1E in the code but the vector is used.

@gharr ,
Now I am using FAT and not using MINIX boot nor  
RESTORE DDPT but is this mean we need DDPT for working properly even on FAT system?

I know in bioshd.c
bios_readtrack for TRACK_CACHE limits num_sectors for buffer size, isn't this enough?

if (num_sectors > (DMASEGSZ / drivep->sector_size))
		num_sectors = DMASEGSZ / drivep->sector_size; 

Thank you!

tyama501:
I was told on twitter that int1E of PC-98 is used for N88 BASIC call.

Thank you.

ghaerr:
Hello @tyama501,

I will give a bit more explanation of how the boot sector, as well as the bioshd.c code, use multi sector reads, and how the DDPT 1Eh vector is repointed, to protect the BIOS from automatically incrementing the track when doing so, which is what the IBM PC BIOS does, which we don't want to have happen. Thus, the DDPT is used by ELKS essentially as a protection mechanism from the BIOS incrementing a track on a multi-sector read, when we do not desire such. It seems, from what we have been discussing, that the INT 1Eh vector, and as such, the DDPT, is not used in PC-98 and thus should be removed. That would be finalized in a future PR from you as a result of testing as described below, leaving a bit more code space in the MINIX boot sector for PC-98.

As @tkchia explained, the low memory address of the INT 1Eh vector, i.e. address 0:0078, stores the location of the IBM PC BIOS DDPT. The BIOS uses offset 4 in the DDPT structure to store the FD "max sector" value, so it can automatically increment the track when a multi-sector READ/WRITE is requested, instead of stopping the I/O short of the requested number of sectors. ELKS always calculates its own ending sector value, and never wants to have the BIOS increment a track number, for the reason that ELKS wants very well-defined operations when using multi-sector I/O, without the possible variations seen across differing system BIOS and possible DDPT structures.

In the boot code, when FAST_READ is set, also as @tkchia explains, the INT 1Eh vector is rewritten to a new DDPT, because FAST_READ attempts to greatly increase the ELKS boot speed by reading until end of track using multi-sector reads, rather than always reading single sectors. FAST_READ is not implemented for FAT boots because we have run out of space in the FAT boot sector. Thus, since the current PC-98 port has never booted a MINIX disk, this has never been tested.

What we are proposing testing would be essentially defining FAST_READ, but eliminating all the code that sets up 0:0078 to point to a new DDPT, as well as restoring to the old vector value. Of course, this will only be functional for MINIX boot, as there is no space on FAT. Thus, you will first have to add code to the MINIX boot that calculates the correct byte count as well as the PDA (physical device address) for the INT 1Bh PC-98 BIOS read sector call.

With regards to bioshd.c, we should probably have you comment out the code that sets up and/or copies a DDPT (i.e. all of `set_ddpt()` and `copy_ddpt()`), and confirm that PC-98 works without the DDPT. (It should).

> I know in bioshd.c
bios_readtrack for TRACK_CACHE limits num_sectors for buffer size, isn't this enough?

Yes. As discussed above, it seems that PC-98 BIOS is functioning properly (at least in the test cases when a disk is not swapped), that the INT 1Bh BIOS read function works fine for multi-sector reads, and the bioshd.c code will always calculate the requested sector count such that an end of track auto-increment is never performed. This is likely why CONFIG_TRACK_CACHE is working on PC-98. (On the IBM PC BIOS, only the DDPT is used to determine the floppy disk max sector number, it is not kept internally except for within that table, which is usually in ROM).

Given that the bioshd.c code is the same for MINIX or FAT boot, you might try removing the DDPT code there first, since it can be done quickly, without having to rewrite any boot code for MINIX boot on PC-98.

> I was told on twitter that int1E of PC-98 is used for N88 BASIC call.

If that is the case, then we probably should not be changing the INT 1Eh vector for PC-98, as it would not be serving its purpose, and could potentially cause bad behavior.

I hope all this makes sense :) Please ask any other questions you may have.

Thank you!

tyama501:
Hello @ghaerr ,

Thank you for the precise explanation.
I think I got the summary of the problem.

I will start checking with removing DDPT first when I got time.
Thank you!

tkchia:
Hello @ghaerr,

Meanwhile, are you OK with me merging this particular PR first?  This is expressly meant as a stopgap measure until we can be sure that multi-sector reads are working properly on the PC-98.  Thank you!

ghaerr:
Hello @tkchia,

> are you OK with me merging this particular PR first? This is expressly meant as a stopgap measure until we can be sure that multi-sector reads are working properly on the PC-98.

Yes, go ahead. @tyama501 can modify this accordingly when he works on getting the MINIX boot working with INT 1Bh.

Thank you!

PR #1403, opened 2022-08-13
[nano-X] Add screen and ramfont driver for PC-98

tyama501:
This is an initial commit for PC-98 nano-X screen and romfont drivers.

As we mentioned in https://github.com/jbruchon/elks/issues/1325#issuecomment-1212224170
there are still things to figure out.

Thank you.


toncho11:
So this will allow landmine, nxworld and nxclock to work on PC-98?

tyama501:
Hello @toncho11 ,

The landmine can be displayed but it seems it also needs mouse driver to play.

ghaerr:
Hello @tyama501,

I just noticed this, and forgot it was here. Would you like me to commit now?

Thank you, and sorry for the delay!

tyama501:
Hello @ghaerr ,

No, I would like to fix the font issue.

I will try your font file first.

I also got the comment that I/O read for the rom should wait a lot.

Thank you!

tyama501:
Hello @ghaerr ,

As I wote in https://github.com/jbruchon/elks/issues/1325#issuecomment-1221112987
I added X6x13.c and renamed romfont_pc98.c to ramfont_pc98.

I also added ramfont.h to declare MWIMAGEBITS and MWCFONT for ELKS
and included in X6x13.c and ramfont_pc98.c, but tell me if this is not right place to declare them.
I modified unsignd short to unsigned int and int32_t to int.

Please point out if further modification is needed for merging.

Thank you!

ghaerr:
Hello @tyama501,

> I could see characters on PC-9801BX using the compiled in fonts X6x13!

That's great, thank you for porting compiled-in fonts to Nano-X! This will allow for likely easier internationalization in the future, as well as better customization of fonts. I am a bit concerned about the added .data segment size, but we will see what happens.

> The bits drop issue on nxterm still exists, but I will add this to the PR.
nxlandmie seems ok so it maybe related to the mouse pointer.

This is very strange - I don't think it is related to the mouse pointer. Instead, I am guessing that the problem may be related to drawing black characters on white background... perhaps try changing nxterm display to draw white on black, and see what happens? For instance, this could be issue with using ORBYTE_FP vs ANDBYTE_FP in driver.

Also, looking closely at the pixel drops on nxterm, there seems to be some regularity in the bit drops. I am wondering whether perhaps some other text output on non-graphic screen could be causing that. Perhaps turn debug output off and see what happens, or clear old text screen first in nxterm.c before switching to graphics mode (ESC [H ESC 2 J ). It could be the text underneath the nxterm window that is causing this? (I think it may very well be!)

> Please point out if further modification is needed for merging.

Overall, things look very good for merging. I think we should take one more step of code cleanup, especially in the xxx_pc98.c files, where sometimes ifdef LINUX or MSDOS is used... that unused code can all be deleted to clean things up.

Specifically, consider cleaning up (deleting) the following:
- ramfont_pc98.c: Lines 32, 34, 40, 42, 44-47
- scr_pc98.c: Lines 45, 47, 92, 184-187, 270-272, 289, 293-297, 299-303
- vgaplan4_pc98.c: Lines 25-27, 32, 41-43, 52-55, 66-70, 72-74, 77, 272-281
Is line 46 correct, with SCREENBASE2 before SCREENBASE1?

Also, feel free to update comments at beginnings of each file for correctness, particularly if comments describe a feature or system that is not PC-98. I don't care that the comment headers have to show the file came from my original versions as much as that they are informative and correct.

After commit, consider define NDEBUG to turn off assert() checking, which will help speed up drawing.


Thank you!

tyama501:
Hello @ghaerr ,

I will clean the codes.

I noticed there are no hardware related codes in ramfont_pc98.c.
May change the name to ramfont.c so that IBM PC can also use it?

 >Is line 46 correct, with SCREENBASE2 before SCREENBASE1?

Yes, this is correct.
Since PC-98 rgb plane order is different from IBM PC, I handled it with here.

BTW, what is the psd->addr in line 75 of vgaplane4_pc98.c for?
I used SCREENBASE0 but do I need to consider SCREENBASE1-3?

	//psd->addr = SCREENBASE;		/* long ptr -> short on 16bit sys*
	psd->addr = SCREENBASE0;		/* long ptr -> short on 16bit sys*/

Thank you.

ghaerr:
Hello @tyama501,

> May change the name to ramfont.c so that IBM PC can also use it?

Yes, good idea!

> BTW, what is the psd->addr in line 75 of vgaplane4_pc98.c for?

In general, psd->addr is used for those screen drivers that malloc memory, or, in some cases, to find the base address of the frame buffer. For PC98 and VGA, it doesn't really matter, since the drivers use their own hard-coded address. So you can leave this as is, w/SCREENBASE0.

Thank you!

tyama501:
Hello @ghaerr ,

I cleaned codes and removed/added comments.
I also renamed VGA_ routines to PC98_.

And you are right!
I cleared the text screen and nxterm does not drop bits.

font_20220823_3

I will think how to clear/stop text screen in this driver after this PR is merged.

Thank you!

tyama501:
Oh, I forgot to change the file name I will do it.

ghaerr:
Hello @tyama501,

Thank you for the cleanup, looks great! :)

> And you are right!
I cleared the text screen and nxterm does not drop bits.

Wow, kind of amazing! Considering that your driver does a strict AND or OR operation on the memory planes, it seems perhaps the PC-98 hardware doesn't allow the text plane to be modified when in graphics mode? It seems that way. One memory plane is shared between text and graphics, right?

Glad everything is now working well. It's been kind of a long journey, but you've done a fantastic job of getting the PC-98 to run graphical programs by writing the drivers basically from scratch, nice job!!

> I will think how to clear/stop text screen in this driver after this PR is merged.

Yes, one idea would be to just clear the text screen in the driver before switching or drawing in graphics mode. I have been thinking about the idea of adding the two (commented out) ioctl's for notifying the kernel that the console is going in and out of graphics mode. This could be useful since the kernel will still allow the text console to be switched (using Alt 1-3) when in graphics mode, and the ioctl could be used to stop that, and/or perform other PC-98 required code. However, I'm not sure we really need all that in the kernel, since it would be useless without the Nano-X screen drivers anyways.

Thank you!

tyama501:

I also noticed that the data in graphic VRAM are remaining and show up momentarily when next graphic program is executed, so I maybe add codes to clear the VRAM when closing.

Thank you.

ghaerr:
> so I maybe add codes to clear the VRAM when closing

It might be better to add the clear VRAM code on open, before switching to graphics mode. This is because sometimes it is desirable to keep the graphics content when exiting.

PR #1405, opened 2022-08-14
[debug] Correct disasm disassembly errors

ghaerr:
As described by @tkchia in https://github.com/jbruchon/elks/issues/1367#issuecomment-1214190526, this PR fixes the following errors in `disasm`:
```
0141  8c 0e 4e 00       mov	%cs,(0x004e)    // show %cs instead of %cl
```
```
0145  ea 00 00 70 00    ljmpw   $0070:$0000    // use GNU as syntax for far jumps
```
```
00b1  2e                seg	%cs            // segment override on separate line
00b2  c7 06 48 01 c0 07 movw	$0x7c0,(0x0148)
```
For the time being, a segment override is shown on a separate line, rather than directly preceding the correct instruction operand. This is because displaying the latter is proving a bit complicated. Thus, the segment overrides are displayed the same way that repeat prefixes are - on the preceding line (s). @tkchia, are you OK with this?

Instructions with six bytes are now displayed properly, without having the hex instruction byte display overflow into the mnemonic instruction display column.

Also, the `shl` instruction was sometimes capitalized, which is fixed.

### DIFF
--- a/elkscmd/debug/disasm.c
+++ b/elkscmd/debug/disasm.c
@@ -45,7 +45,7 @@ int disasm(int cs, int ip, int (*nextbyte)(int, int))
 {
 	startCS = cs;
 	startIP = (int)ip;
-    fetchbyte = nextbyte;
+	fetchbyte = nextbyte;
 	cols = 0;
 	//if (!f_asmout) printf("%04hx:%04hx  ", cs, ip);
 	decode();
@@ -276,7 +276,7 @@ nextopcode:
 #else
                 segOver = operation - 4;
                 prefix = true;
-				goto nextopcode;
+                goto nextopcode;
 #endif
             case 0x27: case 0x2f:  // DA
 				outs(opcode == 0x27? "daa": "das", 0);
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tkchia,

Also, I just noticed I did not fix the indirect far jump and call opcodes you mentioned (0xff 0x2e ...) and (0xff 0x1e ...). I will fix those separately.

[EDIT: indirect far jump and call opcodes fixed]:
```
0083  2e                seg	%cs
0084  ff 2e 07 00       ljmpw	*0x0007
```

Thank you!

tkchia:
Hello @ghaerr,

Thanks!  We just need to tweak the output further to follow the GNU assembler syntax more closely. :slightly_smiling_face:

For far jumps: separate the segment and offset with `','` rather than `':'`.  Also, add `0x` prefixes to the segment and offset values, for consistency?
```
0145  ea 00 00 70 00    ljmpw   $0x0070,$0x0000
```

For segment overrides, simply say e.g. `cs`.
```
0083  2e                cs
0084  ff 2e 07 00       ljmpw	*0x0007
```
Or, if you prefer...:
```
0083  2e ff 2e 07 00    cs ljmpw *0x0007
```

(Yes, this segment override syntax is indeed accepted by GNU `as`.  Apparently every assembler has its own peculiar way of specifying these pesky little things.)

Thank you!

tkchia (changes requested):
Please see my feedback at https://github.com/jbruchon/elks/pull/1405#issuecomment-1214850950 .  Thanks!

ghaerr:
Thanks a lot for your review @tkchia , I've implemented all your feedback!

PR #1407, opened 2022-08-17
[bootopts] Add serial IRQ setting at boot time

Mellvik:
Add `bootopts` option to set serial line IRQs at boot time, supplementing the fixed IRQs in `ports.h`.

Syntax:
```
comirq=,,10,5
comirq=5,7,9,
comirq=,4
```

There is no validation of input values, separators must be `,`, otherwise input handling is robust.  
Tested on physical machines only - with up to 4 serial lines.

@ghaerr - I'm adding a change to `ports.h` to this PR. The default IRQ2 for COM4 is not good as it sometimes collides with networks cards @ IRQ9. IRQ7 is a better choice - the chance that a system has 4 COM lines and a parallel port in use should be minimal.

-M

### DIFF
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -50,7 +50,6 @@ struct serial_info {
  * afaik 8250 works fine up to 19200
  */
 
-#define DEFAULT_BAUD_RATE	9600	/* FIXME: also set in ntty.c */
 #define DEFAULT_LCR		UART_LCR_WLEN8
 
 #define DEFAULT_MCR		\
@@ -93,7 +92,8 @@ extern struct tty ttys[];
 #define	RS_IALLMOSTEMPTY	(    INQ_SIZE / 4)
 
 /* allow init to easily update the port irq from bootopts */
-void set_serial_irq(int tty, int irq) {
+void set_serial_irq(int tty, int irq)
+{
 	ports[tty].irq = irq;
 }
 	
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -50,7 +50,7 @@ struct serial_info {
  * afaik 8250 works fine up to 19200
  */
 
-#define DEFAULT_BAUD_RATE	9600
+#define DEFAULT_BAUD_RATE	9600	/* FIXME: also set in ntty.c */
 #define DEFAULT_LCR		UART_LCR_WLEN8
 
 #define DEFAULT_MCR		\
@@ -92,6 +92,11 @@ extern struct tty ttys[];
 #define	RS_IALLMOSTFULL 	(3 * INQ_SIZE / 4)
 #define	RS_IALLMOSTEMPTY	(    INQ_SIZE / 4)
 
+/* allow init to easily update the port irq from bootopts */
+void set_serial_irq(int tty, int irq) {
+	ports[tty].irq = irq;
+}
+	
 /*
  * Flush input by reading RX register.
  * Not used when FIFO enabled, as HW fifo cleared when enabled.
--- a/elks/include/arch/ports.h
+++ b/elks/include/arch/ports.h
@@ -14,7 +14,7 @@
  *  5*  HW IDE hard drive   CONFIG_BLK_DEV_HD       Driver doesn't work
  *  6*  Unused
  *  6*  HW floppy drive     CONFIG_BLK_DEV_FD       Driver doesn't compile
- *  7   Unused (LPT)
+ *  7   Unused (LPT, Com4)
  *  8   Unused (RTC)
  *  9   3C509/EL3 (/dev/eth) CONFIG_ETH_EL3         Optional
  * 10   Unused (USB)                                Turned off
@@ -88,7 +88,7 @@
 #define COM3_IRQ	5		/* unregistered unless COM3_PORT found*/
 
 #define COM4_PORT	0x2e8
-#define COM4_IRQ	2		/* unregistered unless COM4_PORT found*/
+#define COM4_IRQ	7		/* unregistered unless COM4_PORT found*/
 
 /* Ethernet card settings may be overridden in /bootopts using netirq= and netport= */ 
 /* ne2k, ne2k.c */ 
--- a/elks/include/linuxmt/ntty.h
+++ b/elks/include/linuxmt/ntty.h
@@ -106,6 +106,7 @@ extern void ttystd_release(struct tty *tty);
 extern int tty_allocq(struct tty *tty, int insize, int outsize);
 extern void tty_freeq(struct tty *tty);
 		/* Allocate and free character queues*/
+extern void set_serial_irq(int tty, int irq);
 
 extern void set_console(dev_t dev);
 
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -292,16 +292,18 @@ static int INITPROC parse_dev(char * line)
 
 static void comirq(char *line) {
 	int i;
-	char *l, *m;
+	char *l, *m, c;
 
 	l = line;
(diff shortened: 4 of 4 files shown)


Mellvik:
@ghaerr - re. the 'shared irq project' (https://github.com/jbruchon/elks/pull/1385#issue-1323262534) you may be interested in this:

If a COM port is set to IRQ2 (or IRQ9) in ELKS, a NIC configured to use IRQ 9 will fail even though the COM port has not been opened and is actually set to (jumper config) to a different IRQ.
```
ttyS1 at 0x2f8, irq 3 is a 16550A
ttyS2 at 0x3e8, irq 10 is a 16550A
ttyS3 at 0x2e8, irq 2 is a 16550A             <-----------------------------------------
xms: A20 was off now on, using int 15/1F, 500 xms buffers, 512000 ram
eth: ne2k at 0x380, irq 9 (8bit), MAC 00:1f:11:02:60:2d, flags 0x80 <---------------
eth: wd80x3 at 0x240, irq 2, ram 0xce00 not found
eth: 3c509 at 0x330, irq 11 not found
bioshd: hda BIOS CHS 992,16,63
bioshd: hdb BIOS CHS 995,16,1
/dev/hda: 993 cylinders, 16 heads, 63 sectors = 488.7 Mb
/dev/hdb: 995 cylinders, 16 heads, 1 sectors = 7960.0 kb
/dev/fd0: 80 cylinders, 2 heads, 18 sectors = 1440.0 kb
Partitions: hda:(0,1000944)  hda1:(63,696465)  hda2:(696528,132048)  hda3:(828576,132048)  hda4:(960624,41328) 
 hdb:(0,15920)  hdb1:(1,15919) 
device_setup: BIOS drive 0x80, root device 0x304
PC/AT class machine, syscaps 0xff, 609K base ram.
ELKS kernel 0.6.0 (59136 text, 12080 ftext, 8112 data, 42528 bss, 14894 heap)
Kernel text at d0:0000, ftext f40:0000, data 1233:0000, top 9840:0, 472K free
VFS: Mounted root 0x0304 (minix filesystem).
Running /etc/rc.sys script
fsck -r /dev/hda4: /dev/hda4 is clean, no check.
Starting networking on ne2k
ktcp -b -p ne2k 10.0.2.17 10.0.2.2 255.255.255.0
ne2k: Unable to use IRQ 9 (errno -16)    <-----------------------------------------------------
ktcp: failed to open eth device /dev/ne2k
Network start failed
Thu Mar  4 13:45:15 1999


ELKS 0.6.0

login: 
```

Also, setting two COM ports to the same IRQ prevents both from working (interrupting) even though only one is open at a time.

--M

ghaerr:
Hello @Mellvik,

I would like to separate my comments about the PR from the bug reports also mentioned.

WRT the PR, looks good with a few minor nits: 1) let's use the same coding style in the kernel so the code appears uniform, which is that on function definitions, the opening brace is on its own line. This affects set_serial_line and comirq. 2) I don't really understand the FIXME on the DEFAULT_BAUD_RATE in serial-8250.c. In fact, it seems to be used nowhere, so suggest deleting the line entirely. 3) let's move the extern set_serial_irq extern definition to the end of ntty.h: for a variety of reasons, its better to have as many externs in .h files as possible, rather than embedded in .c files.

Thank you!

ghaerr:
> If a COM port is set to IRQ2 (or IRQ9) in ELKS, a NIC configured to use IRQ 9 will fail even though the COM port has not been opened and is actually set to (jumper config) to a different IRQ.

> Also, setting two COM ports to the same IRQ prevents both from working (interrupting) even though only one is open at a time.

I think there are two things going on here: on the first issue, ELKS will remap any IRQ 2 to IRQ 9, as the hardware IRQ 2 line is used for the cascade between the two 8259 PIC controllers. See the code in elks/arch/i86/irq.c and irq-8259.c:
```
int remap_irq(int irq)
{
    if ((unsigned int)irq > 15 || (irq > 7 && !(sys_caps & CAP_IRQ8TO15)))
        return -EINVAL;
    if (irq == 2 && (sys_caps & CAP_IRQ2MAP9))
        irq = 9;                        /* Map IRQ 9/2 over */
    return irq;
}
```
The CAP_IRQ2MAP9 is a sys_cap (system capability) which is always set for IBM AT systems and later. Is this the case with your system(s)? Perhaps a discussion on hardware IRQ 2/9 hardware handling should be had, to see if this is correct.

The second issue, mapping two IRQs to the same number, appears to be a bug, although it could be related to the IRQ 2 -> IRQ 9 above. I suggest adding a printk inside irq.c::request_irq() to determine if/when the kernel is getting the request_irq calls. If the kernel has mapped an IRQ via request_irq, it will return -EBUSY (=-16), which seems to be what ktcp is getting.

I checked the serial driver code, and the request_irq is delayed until open, so this shouldn't be happening - but we need the printk to know for sure.

Mellvik:
Thank you @ghaerr.

> 17. aug. 2022 kl. 17:07 skrev Gregory Haerr ***@***.***>:
> 
> 
> If a COM port is set to IRQ2 (or IRQ9) in ELKS, a NIC configured to use IRQ 9 will fail even though the COM port has not been opened and is actually set to (jumper config) to a different IRQ.
> 
> Also, setting two COM ports to the same IRQ prevents both from working (interrupting) even though only one is open at a time.
> 
> I think there are two things going on here: on the first issue, ELKS will remap any IRQ 2 to IRQ 9, as the hardware IRQ 2 line is used for the cascade between the two 8259 PIC controllers. See the code in elks/arch/i86/irq.c and irq-8259.c:
> 
> int remap_irq(int irq)
> {
>     if ((unsigned int)irq > 15 || (irq > 7 && !(sys_caps & CAP_IRQ8TO15)))
>         return -EINVAL;
>     if (irq == 2 && (sys_caps & CAP_IRQ2MAP9))
>         irq = 9;                        /* Map IRQ 9/2 over */
>     return irq;
> }
> The CAP_IRQ2MAP9 is a sys_cap (system capability) which is always set for IBM AT systems and later. Is this the case with your system(s)? Perhaps a discussion on hardware IRQ 2/9 hardware handling should be had, to see if this is correct.
> 
I realize my message was kind of ambiguous. From my perspective, IRQ 2 and 9 are the same (AT and up), using 2 is just a precaution to make sure something works on XT and lower/older. So the issue wasn't 2 vs 9 but the fact that an IRQ seems taken when it shouldn't be. 

Which is the same as the 2nd issue.

I'll take a closer look and report back - on #1385.


—M

Mellvik:
Thanks @ghaerr,

> 
> WRT the PR, looks good with a few minor nits: 
> 1) let's use the same coding style in the kernel so the code appears uniform, which is that on function definitions, the opening brace is on its own line. This affects set_serial_line and comirq. 
> 2) I don't really understand the FIXME on the DEFAULT_BAUD_RATE in serial-8250.c. In fact, it seems to be used nowhere, so suggest deleting the line entirely. 

It's used OK. If the default speed in `ntty.c` is change to something else, this one changes it back to 9600 after the first line during boot (`Direct console, scan kbd` etc.). It will be my pleasure to remove it.
> 3) let's move the extern set_serial_irq extern definition to the end of ntty.h: for a variety of reasons, its better to have as many externs in .h files as possible, rather than embedded in .c files.

OK

--M

ghaerr:
@Mellvik: looks fine thank you, let me know when ready to commit.

Mellvik:
Thank you,
I just noticed some comments that should be deleted (just ahead of the now deleted 9600 define). I'll get rid of those, then it's ready.

--M

ghaerr:
> I just noticed some comments that should be deleted (just ahead of the now deleted 9600 define)

Actually, I don't see those. Perhaps we should commit now.

Mellvik:
@ghaerr,

please go ahead and commit this one. I blew up the branch, so the irq fix will come in a separate PR.

-M

PR #1414, opened 2022-08-21
[misc] minor fixes and adjustments: cp, fdisk, truncate.c

Mellvik:
`truncate.c` - adding cast in `V1_trunc_direct` per discussion in https://github.com/jbruchon/elks/issues/1367#issuecomment-1202165859

`fdisk`: Use default drive when no drive is specified with `-l` option,  just like `fdisk` with no options.

`cp`:  Require verbose mode (`-v`) to show the `Copying ...` message when using recursive copy (`-R`). Avoids messing up output from `makeboot`.

### DIFF
--- a/elks/fs/minix/truncate.c
+++ b/elks/fs/minix/truncate.c
@@ -41,18 +41,14 @@ static int V1_trunc_direct(register struct inode *inode)
     int retry = 0;
 
   repeat:
-    for (i = (int)DIRECT_BLOCK; i < 7; i++) {
+    for (i = DIRECT_BLOCK; i < 7; i++) {
 	p = &inode->u.minix_i.i_zone[i];
 	if (!(tmp = *p)) continue;
 	bh = get_hash_table(inode->i_dev, (block_t) tmp);
-
-	if (i < (int)DIRECT_BLOCK) {	/* In case the file shrunk while
-					 * we were truncating. Unlikely
-					 * if at all possible. */
+	if (i < DIRECT_BLOCK) {
 	    brelse(bh);
 	    goto repeat;
 	}
-
 	if ((bh && buffer_count(bh) != 1) || tmp != *p) {
 	    retry = 1;
 	    brelse(bh);
--- a/elks/fs/minix/truncate.c
+++ b/elks/fs/minix/truncate.c
@@ -41,14 +41,18 @@ static int V1_trunc_direct(register struct inode *inode)
     int retry = 0;
 
   repeat:
-    for (i = DIRECT_BLOCK; i < 7; i++) {
+    for (i = (int)DIRECT_BLOCK; i < 7; i++) {
 	p = &inode->u.minix_i.i_zone[i];
 	if (!(tmp = *p)) continue;
 	bh = get_hash_table(inode->i_dev, (block_t) tmp);
-	if (i < DIRECT_BLOCK) {
+
+	if (i < (int)DIRECT_BLOCK) {	/* In case the file shrunk while
+					 * we were truncating. Unlikely
+					 * if at all possible. */
 	    brelse(bh);
 	    goto repeat;
 	}
+
 	if ((bh && buffer_count(bh) != 1) || tmp != *p) {
 	    retry = 1;
 	    brelse(bh);
--- a/elkscmd/disk_utils/fdisk.c
+++ b/elkscmd/disk_utils/fdisk.c
@@ -418,10 +418,10 @@ int main(int argc, char **argv)
 
 
     if (mode == MODE_LIST) {
-	if (*dev != 0) {
-	    list_partition(dev);
-	    exit(0);
-	} else goto usage;
+	if (*dev == 0)
+	    strcpy(dev,DEFAULT_DEV);
+	list_partition(dev);
+	exit(0);
     }
 
     if (mode == MODE_EDIT) {
--- a/elkscmd/file_utils/cp.c
+++ b/elkscmd/file_utils/cp.c
@@ -476,7 +476,7 @@ int copy_directory(char *source_dir, char *dest_dir)
 		return 1;
 	}
 
-	printf("Copying %d files, %lu blocks from %s to %s\n", inodes.count, numblocks,
+	if (opt_verbose) printf("Copying %d files, %lu blocks from %s to %s\n", inodes.count, numblocks,
 		source_dir, destination_dir);
 
 	err = do_copies();


ghaerr:
Hello @Mellvik,

I'm ok with all this, but for truncate.c, why not just put the (int) cast in the DIRECT_BLOCK define, to match the other xxx_BLOCK definitions, thus leaving out the (int) cast in the regular code, making it easier to read/understand:
```
#define DIRECT_BLOCK            ((int)((inode->i_size + 1023) >> 10))
```
Thank you!

Mellvik:
>  why not just put the (int) cast in the DIRECT_BLOCK define, to match the other xxx_BLOCK definitions, thus leaving out the (int) cast in the regular code, making it easier to read/understand:
> #define DIRECT_BLOCK            ((int)((inode->i_size + 1023) >> 10))
No problem, I guess it's a matter of taste. That will also make the casts in the other 2 macros superfluous.

-M

ghaerr:
>  I guess it's a matter of taste. That will also make the casts in the other 2 macros superfluous.

I see, good point. For some reason, I'm having trouble trying to understand the truncate code in general here. What exactly is the purpose of your changed (int) cast in the first place? It seems that it will only change the code for the one line:
```
if (i < DIRECT_BLOCK) {
```
Where now `i` will have to be sign-extended for the unsigned long compare, but with the changed code, it will not be sign extended and will do a signed compare.

I'm not quite understanding how the DIRECT_BLOCK macro gets the first 7 blocks of the file by performing a logical right shift of 10 bits (=1K) after averaging up (+1023): that gets the last blocks of the file, not the first, right?

Has anyone tested the truncate routine with a large (~63M) file?

ghaerr:
Hello @Mellvik,

> For some reason, I'm having trouble trying to understand the truncate code in general here.
> I'm not quite understanding how the DIRECT_BLOCK macro ... that gets the last blocks of the file, not the first, right?

Arrghh! I finally realized (and you've said it before) that the inode->i_size field is set to the desired *truncated* value, *not* the previous file size!! It all makes sense now.

> That will also make the casts in the other 2 macros superfluous.

I see, didn't see that. Well, if the `(int)` cast goes into the DIRECT_BLOCK macro, then by all rights the `(int)` cast should be removed from the two following.

I remain concerned about the use of signed `int i`, when used on 64M files, since `i` will be negative at that point. I'm going to test truncate on a 64MB file.

ghaerr:
Hello @Mellvik,

As a result of the larger number of bugs found on testing 64MB disks in #1415, can I ask you to revert the changes in truncate.c? As mentioned in #1415, the `(int)` casts (in both the existing code and your fix) are not correct, and result in incorrect behavior when the MINIX block number is greater than 32767. Thus, I'd rather leave the existing code untouched until we get to the bottom of the problem(s).

I have to say, you have quite the knack for finding ELKS bugs with stuff you poke into!!! :)

Thank you!

Mellvik:
Thank you @ghaerr,
I'll do that.

And happy to have triggered the hunt for another elusive bug. I'm sort of happy to hear you too had trouble wrapping your head around the stuff in truncate. Obviously I never did the block math, but I spent quite some time trying to understand what was going on while on the hunt that turned out to be a virus.

As indicated by the comment in the now deleted `truncate.c`, the `if ( i < j)` part in all 3 truncate routines caused a lot of head scratching. It assumes the `DIRECT_BLOCK` macro may change while the routine is looping, which  is conceivable in a multitasking system. OTOH, if the inode is locked which I think it is or should be, there is no chance for such a change and the code is useless.

-M

ghaerr:
Agreed @Mellvik, the code in truncate is pretty hard to understand. The whole idea of comparing against a value that might have increased, but not worry about the opposite of it decreasing, still makes me think twice. And the comparison seems opposite to what it should be. I guess should the value (file size) increase, the truncate function figures it needs to keep truncating (despite not locking the inode and allowing other processes to write the file)! The inode is locked with a semaphore, but it appears that is only for prohibiting reentrancy of the truncate code itself; reads/writes to the inode are still allowed.

(Amazingly enough, in looking to confirm the read/write possibility above, the kernel explicitly calls `schedule()` after each read and write system call! I suppose we'll leave the reason for that for another conversation.)

>  It assumes the DIRECT_BLOCK macro may change while the routine is looping, which is conceivable in a multitasking system. OTOH, if the inode is locked which I think it is or should be, there is no chance for such a change and the code is useless.

Yes. After reading your comment I pondered whether we might just remove that code, but I finally realized that its not about whether the kernel is multitasking (i.e. reentrant), which ELKS is not. However, the kernel may (in the future) sleep a task on I/O, and as such, the truncating task could be put to sleep while buffers are flushed, possibly allowing another task to write to the same file. (With the ELKS BIOS synchronously doing all I/O, this won't happen, but we've talked about changing that someday).

My final thoughts about truncate are that Linus probably thought a lot more about this than we have, and that once a truncate request has started, it must be finished to completion (i.e. completely truncated every time), regardless of other applications writing to the file losing their data. One reason is that otherwise, the inode could end up in a messed-up state, with portions of the direct, indirect, or double-indirect tables having holes in them (which wasn't allowed until much later).

Mellvik:
Hi @ghaerr - 

> Yes. After reading your comment I pondered whether we might just remove that code, but I finally realized that its not about whether the kernel is multitasking (i.e. reentrant), which ELKS is not. However, the kernel may (in the future) sleep a task on I/O, and as such, the truncating task could be put to sleep while buffers are flushed, possibly allowing another task to write to the same file. (With the ELKS BIOS synchronously doing all I/O, this won't happen, but we've talked about changing that someday).

It's an interesting scenario and I agree - taking the time to understand it completely is indeed valuable, thank you! 
My simple logic (and understanding of inode based file systems) is that the inode must be locked when it is about to be changed (size increase, decrease, other metadata changes). So it should be safe to assume (we can actually test this, maybe we should) that the inode is locked when entering the truncate code, which in this case means that we own the inode and the i_size is under our control.
> My final thoughts about truncate are that Linus probably thought a lot more about this than we have, and that once a truncate request has started, it must be finished to completion (i.e. completely truncated every time), regardless of other applications writing to the file losing their data.
> 
Agreed. That said, this code has been around for a while and presumably gone through a lot of changes over the years. There is a chance this particular piece of code is a leftover from when the functionality was different. One which lost its purpose.
> One reason is that otherwise, the inode could end up in a messed-up state, with portions of the direct, indirect, or double-indirect tables having holes in them (which wasn't allowed until much later).
> 
True, but we would have to assume that an inode lock is always observed, right?

This is truly a small fish in a big bowl and one most likely without consequences. I suspect that in 99+% of the cases in which truncate is called, i_size is zero (rm). So the question ends up being whether it's worth it to create a test case and see what happens.

Thank you !!

—M

PR #1415, opened 2022-08-21
[kernel] Problems with large files and large MINIX disks

ghaerr:
Discussion and preliminary duplication of problem found in #1414.

After my comments about signed/unsigned comparison of block values in https://github.com/jbruchon/elks/pull/1414#issuecomment-1221600471, I decided to go ahead with testing ELKS operation on hard drive images larger than 32MB (64MB is MINIX v1 maximum). 

This PR contains some debug code in fs/minix/truncate.c, as well as a enhancement to produce a 64MB MINIX disk image (built using `cd image; make hd64-minix`). The 64MB image requires the ELKS kernel and applications to use all 16 bits in an integer for the block numbers, which end up causing a wrap to a negative number, producing a number of associated errors.

Right out of the gate, there are the following big problems with ELKS and 64MB MINIX disks:
- The `mfs` program doesn't seem to allocate large files correctly, as they don't take up any of the otherwise free space on the MINIX volume. This can be duplicated by adding a 32MB+ file in elkscmd/rootfs_tempate/root, and observing the booted free space - it does not include the large file space taken out.
- When 64MB disks are mounted (or booted from), the `df` and `mount` commands immediately show different results for the free space.
- The MINIX fs function truncate function fails when INDIRECT_BLOCKS go from negative to positive (see or run included debug code). This is a result of using an `(int)` cast and comparison with an `int i` count, when `unsigned` should be used.

Long story short - MINIX fs is totally broken on 64MB disks!

More commits coming, but posted this to allow others to test using 64MB disks. I'm sure there are more issues.

Here is a sample of `df` and `mount` showing different results, on an otherwise normal 64MB disk, with no large files:
```
xms: A20 was on now on, using unreal mode, 2500 xms buffers, 2560000 ram
eth: ne2k at 0x300, irq 12 (16bit), MAC 52:54:00:12:34:56, flags 0x80
eth: wd80x3 at 0x240, irq 2, ram 0xce00 not found
eth: 3c509 at 0x330, irq 11 not found
rd: 64K ramdisk at d000:0000
ssd: 96K disk
bioshd: hda BIOS CHS 126,16,63
bioshd: hda  IDE CHS 127,16,63    <--- 64MB disk created with 127 cylinders, resulting in 64008 blocks total ~64MB
/dev/hda: 127 cylinders, 16 heads, 63 sectors = 62.5 Mb
Partitions: hda:(0,128016)  no mbr, none.
device_setup: BIOS drive 0x80, root device 0x300
PC/AT class machine, syscaps 0xff, 639K base ram.
ELKS kernel 0.6.0 (62832 text, 12320 ftext, 8944 data, 42672 bss, 13918 heap)
Kernel text at 2d0:0000, ftext 1227:0000, data 1529:0000, top 9fc0:0, 490K free
VFS: Mounted root 0x0300 (minix filesystem).
Running /etc/rc.sys script
Sun Aug 21 16:35:58 2022


ELKS 0.6.0

login: root
# 
# df
/dev/hda          64008    42056    21952  35%    19%    <--- free count incorrect, 2880K image on 64MB disk
# mount
/dev/hda  (minix) blocks  64008 free  62088 mount /   <--- likely correct free space
```

### DIFF
--- a/elks/fs/minix/truncate.c
+++ b/elks/fs/minix/truncate.c
@@ -41,9 +41,11 @@ static int V1_trunc_direct(register struct inode *inode)
     int retry = 0;
 
   repeat:
+  printk("DIRECT_BLOCK %ld\n", DIRECT_BLOCK);
     for (i = DIRECT_BLOCK; i < 7; i++) {
 	p = &inode->u.minix_i.i_zone[i];
 	if (!(tmp = *p)) continue;
+        printk("i=%d\n", i);
 	bh = get_hash_table(inode->i_dev, (block_t) tmp);
 	if (i < DIRECT_BLOCK) {
 	    brelse(bh);
@@ -88,6 +90,7 @@ static int V1_trunc_indirect(register struct inode *inode,
     }
     map_buffer(ind_bh);
   repeat:
+  printk("INDIRECT_BLOCK %d\n", INDIRECT_BLOCK(offset));
     for (i = INDIRECT_BLOCK(offset); i < 512; i++) {
 	if (i < 0) i = 0;
 	else if (i < INDIRECT_BLOCK(offset)) goto repeat;
@@ -145,6 +148,7 @@ static int V1_trunc_dindirect(register struct inode *inode,
     }
     map_buffer(dind_bh);
   repeat:
+  printk("DINDIRECT_BLOCK %d\n", DINDIRECT_BLOCK(offset));
     for (i = DINDIRECT_BLOCK(offset); i < 512; i++) {
 	if (i < 0) i = 0;
 	if (i < DINDIRECT_BLOCK(offset)) goto repeat;
--- a/elks/fs/minix/truncate.c
+++ b/elks/fs/minix/truncate.c
@@ -13,8 +13,8 @@
 #include <linuxmt/fcntl.h>
 
 #define DIRECT_BLOCK		((inode->i_size + 1023) >> 10)
-#define INDIRECT_BLOCK(offset)	((int)(DIRECT_BLOCK)-offset)
-#define DINDIRECT_BLOCK(offset) (((int)(DIRECT_BLOCK)-offset)>>9)
+#define INDIRECT_BLOCK(offset)	((long)(DIRECT_BLOCK)-offset)
+#define DINDIRECT_BLOCK(offset) (((long)(DIRECT_BLOCK)-offset)>>9)
 
 /*
  * Truncate has the most races in the whole filesystem: coding it is
@@ -37,7 +37,7 @@ static int V1_trunc_direct(register struct inode *inode)
     unsigned short *p;
     register struct buffer_head *bh;
     unsigned short tmp;
-    int i;
+    unsigned int i;
     int retry = 0;
 
   repeat:
@@ -45,9 +45,8 @@ static int V1_trunc_direct(register struct inode *inode)
     for (i = DIRECT_BLOCK; i < 7; i++) {
 	p = &inode->u.minix_i.i_zone[i];
 	if (!(tmp = *p)) continue;
-        printk("i=%d\n", i);
 	bh = get_hash_table(inode->i_dev, (block_t) tmp);
-	if (i < DIRECT_BLOCK) {
+	if (i < (unsigned int)DIRECT_BLOCK) {
 	    brelse(bh);
 	    goto repeat;
 	}
@@ -68,10 +67,11 @@ static int V1_trunc_direct(register struct inode *inode)
 }
 
 static int V1_trunc_indirect(register struct inode *inode,
-			     int offset, unsigned short *p)
+			     unsigned int offset, unsigned short *p)
 {
     struct buffer_head *bh;
     int i;
+    long j;
     unsigned short tmp;
     register struct buffer_head *ind_bh;
     unsigned short *ind;
@@ -90,15 +90,15 @@ static int V1_trunc_indirect(register struct inode *inode,
     }
     map_buffer(ind_bh);
   repeat:
-  printk("INDIRECT_BLOCK %d\n", INDIRECT_BLOCK(offset));
-    for (i = INDIRECT_BLOCK(offset); i < 512; i++) {
-	if (i < 0) i = 0;
-	else if (i < INDIRECT_BLOCK(offset)) goto repeat;
-	ind = i + (unsigned short *) ind_bh->b_data;
+  printk("INDIRECT_BLOCK %ld\n", INDIRECT_BLOCK(offset));
+    for (j = INDIRECT_BLOCK(offset); j < 512; j++) {
+	if (j < 0) j = 0;
+	else if (j < INDIRECT_BLOCK(offset)) goto repeat;
+	ind = (int)j + (unsigned short *) ind_bh->b_data;
 	tmp = *ind;
 	if (!tmp) continue;
 	bh = get_hash_table(inode->i_dev, (block_t) tmp);
-	if (i < INDIRECT_BLOCK(offset)) {
+	if (j < INDIRECT_BLOCK(offset)) {
 	    brelse(bh);
 	    goto repeat;
 	}
@@ -128,7 +128,7 @@ static int V1_trunc_indirect(register struct inode *inode,
 }
 
 static int V1_trunc_dindirect(register struct inode *inode,
-			      int offset, unsigned short *p)
+			      unsigned int offset, unsigned short *p)
 {
     int i;
     unsigned short tmp;
@@ -148,7 +148,7 @@ static int V1_trunc_dindirect(register struct inode *inode,
     }
     map_buffer(dind_bh);
   repeat:
-  printk("DINDIRECT_BLOCK %d\n", DINDIRECT_BLOCK(offset));
+  printk("DINDIRECT_BLOCK %ld\n", DINDIRECT_BLOCK(offset));
     for (i = DINDIRECT_BLOCK(offset); i < 512; i++) {
 	if (i < 0) i = 0;
 	if (i < DINDIRECT_BLOCK(offset)) goto repeat;
--- a/elks/fs/minix/truncate.c
+++ b/elks/fs/minix/truncate.c
@@ -11,6 +11,7 @@
 #include <linuxmt/minix_fs.h>
 #include <linuxmt/stat.h>
 #include <linuxmt/fcntl.h>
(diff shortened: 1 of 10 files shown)


ghaerr:
Pushed commit that fixes `df` for disks larger than 32MB. `df` now matches `mount` output.

ghaerr:
Pushed commit that fixes MINIX truncate on files larger than 32MB.
Fix `fsck -lsv` output for correct % zones used, and allow for up to 64K zones/files/dirs, etc.
Debug output still in truncate routine.

toncho11:
Hmm, but fat is working OK for "larger" images, right? So there is an advantage in using fat after all.

ghaerr:
> The mfs program doesn't seem to allocate large files correctly, as they don't take up any of the otherwise free space on the MINIX volume. 

Wow, just figured out what is going on: contrary to my statement in https://github.com/jbruchon/elks/pull/1414#issuecomment-1222523982 that holes are not allowed, both the ELKS kernel and `mfs` definitely allow "holes" in files. If a block in a file is not defined in the MINIX ZMAP (that is, a cleared block bit but within the file size), ELKS will just return a buffer with all zeros. And `mfs`, when reading a file to write to MINIX fs, checks blocks for all zeros and 1) doesn't write the block, and 2) clears the ZMAP bit. This only happens with `mfs`, not when writing files from within ELKS itself while running.

That is, this means that any files written to the filesystem (say the root user directory /root) to be used on a disk image using `dd if=/dev/zero ...` in fact, are *not* written to the filesystem, but *do* work properly, returning all zeros when read. And that's why the `df` program didn't change any space for a 63MB file of all zeros when deleted - it was never allocated! This also means that testing on ELKS using files created this way don't actually test the filesystem read routines, as the blocks aren't actually read!!

Within the ELKS kernel write routine, it doesn't check whether a block is all zeros, possibly because that could decrease I/O speed on old systems substantially - so writes of zero blocks using `dd` or any other method are always written by the ELKS kernel.

@Mellvik, what do you think of all this!!?

I will test deleting non-zero files of large size, and then plan on committing this PR if everything works.

ghaerr:
> fat is working OK for "larger" images, right? So there is an advantage in using fat after all.

Yes. For disk sizes larger than 64MB, FAT is the only option, a huge advantage. (I'm at FAT supporter :)

I just tested `mfs` on MINIX images > 32MB, it doesn't work. More fixes coming.

Mellvik:
Minor correction :-)

> 
> fat is working OK for "larger" images, right? So there is an advantage in using fat after all.
> 
> Yes. For disk sizes larger than 64MB, FAT is the only option, a huge advantage. (I'm at FAT supporter :)
> 
I know you meant partition sizes, @ghaerr — and not being a FAT supporter myself, I admit to reverting to it when the need arises. Maybe minixfs2 should be on the 0.8 wish list?

Thank you!

-M

Mellvik:
> 22. aug. 2022 kl. 18:38 skrev Gregory Haerr ***@***.***>:
> 
> 
> The mfs program doesn't seem to allocate large files correctly, as they don't take up any of the otherwise free space on the MINIX volume.
> 
> Wow, just figured out what is going on: contrary to my statement in #1414 (comment) <https://github.com/jbruchon/elks/pull/1414#issuecomment-1222523982> that holes are not allowed, both the ELKS kernel and mfs definitely allow "holes" in files. If a block in a file is not defined in the MINIX ZMAP (that is, a cleared block bit but within the file size), ELKS will just return a buffer with all zeros. And mfs, when reading a file to write to MINIX fs, checks blocks for all zeros and 1) doesn't write the block, and 2) clears the ZMAP bit. This only happens with mfs, not when writing files from within ELKS itself while running.
> 
> That is, this means that any files written to the filesystem (say the root user directory /root) to be used on a disk image using dd if=/dev/zero ... in fact, are not written to the filesystem, but do work properly, returning all zeros when read. And that's why the df program didn't change any space for a 63MB file of all zeros when deleted - it was never allocated! This also means that testing on ELKS using files created this way don't actually test the filesystem read routines, as the blocks aren't actually read!!
> 
> Within the ELKS kernel write routine, it doesn't check whether a block is all zeros, possibly because that could decrease I/O speed on old systems substantially - so writes of zero blocks using dd or any other method are always written by the ELKS kernel.
> 
> @Mellvik <https://github.com/Mellvik>, what do you think of all this!!?
> 
> I will test deleting non-zero files of large size, and then plan on committing this PR if everything works.
> 

Wow @ghaerr - this sounds BAD. I have to think about this, but off the top of my head I don't think this behaviour is acceptable.


Thank you.

-M

ghaerr:
> Maybe minixfs2 should be on the 0.8 wish list?

That would be a lot of code duplication without much benefit, until people start using larger partitions... but could be done. We're talking about a relatively large kernel code size increase, unless the MINIX v2 filesystem took over as the default. I'd say we can wait awhile, since it seems few people have even used 64MB MINIX filesystems (as evidenced here: they don't work)!

> this sounds BAD. I have to think about this, but off the top of my head I don't think this behaviour is acceptable.

If you're referring to MINIX holes being omitted from disk storage - actually, it's a pretty cool feature. There's no need to store 1024 zero bytes (aligned on a 1K block boundary), and applications won't know the difference. In particular, some older coding styles of unnecessarily initializing .data segment arrays to zero like the following won't produce any storage on disk, for those portions that are block aligned:
```
char buf[4096] = { 0 }; // <--- generates 4K zeros in application .data segment
```
The above could save up to 4 disk blocks using the MINIX zero-holes feature. This also explains why we sometimes see the `df`/`du` values not matching with the free space on volume, which previously I couldn't figure out.

Latest update on `mfs`: turns out any file larger than 519K bytes (which is the first requirement to use double indirect storage blocks =7 + 512, is not stored properly on the disk image, and produces `fsck` errors. Any data stored past 519K is not stored (and likely returned as zero bytes, as a result of the zero-holes feature above).

Mellvik:
Thank you @ghaerr,

> Maybe minixfs2 should be on the 0.8 wish list?
> That would be a lot of code duplication without much benefit, until people start using larger disks... but could be done. We're talking about a relatively large kernel code size increase, unless the MINIX v2 filesystem took over as the default. I'd say we can wait awhile, since it seems few people have even used 64MB MINIX filesystems (as evidenced here: they don't work)!
> 
Agreed (as to V2). V1 is fine for now - it's not like the 64M partitions (and regularly 3 or 4 of them on the drive) fill up all the time.
V1: It's great that we get more bugs out. That said, V1 @ <32M has been reasonable stable (and very usable for small and moderate size files) for quite some time, and got a big lift when you fixed fsck a while back. I've never used anything else except for performance testing and experimentation with big files as mentioned yesterday. Actually, from my perspective it's the increased practical use of the Minix V1 fs that reveal the newfound and admittedly very serious bugs. And it beats FAT hands down performance-wise. It seems we're getting close...
> this sounds BAD. I have to think about this, but off the top of my head I don't think this behaviour is acceptable.
> 
> If you're referring to MINIX holes being omitted from disk storage - actually, it's a pretty cool feature. There's no need to store 1024 zero bytes (aligned on a 1K block boundary), and applications won't know the difference. In particular, some older coding styles of unnecessarily initializing .data segment arrays to zero like the following won't produce any storage on disk, for those portions that are block aligned:
> 
> char buf[4096] = { 0 }; // <--- generates 4K zeros in application .data segment
> The above could save up to 4 disk blocks using the MINIX zero-holes feature. This also explains why we sometimes see the df/du values not matching with the free space on volume, which previously I couldn't figure out.
> 
OK; let me rephrase this one: If what we have is support for sparse files, it's indeed great! Their behaviour is pretty much standardized so their use is portable: Reading non existing parts of the file returns zeroes, writing zeroes result in physical blocks being filled with those zeroes. ls shows the file size including the holes, df shows actual usage.
I think this is what we should be aiming for. If it's there already, fantastic.


-M
>

ghaerr:
Last push fixes three major bugs in `mfs`:
- Invalid image created when any block number > 32767 (any filesystem larger than 32MB)
- Invalid zones created on any file requiring double indirect blocks (any file > 519K)
- Invalid image created when any files > 775K were included in filesystem

This last bug took *way* too long to figure out, but ended up being a single line change, where an indirect block buffer array was defined using host `int` instead of `unsigned short`, which caused critical failure on any double indirect block number > 255 (=775K filesize = 7 direct + 512 indirect + 256 dbl indirect), ending up creating filesystems 4GB in size through seeking to negative numbers. I think I worked on this straight for two days before finding out what was happening. The good news is I understand MINIX indirect and double indirect blocks real well now :)

Removed printk's from truncate, everything seems to work well now. Debug statements left in and available in `mfs` or kernel through define DEBUG.

Also, `make images` now creates a 64MB MINIX image for testing called `hd64-minix.img`. This will remain as part of the standard build, since we obviously need to test filesystem edge cases much more thoroughly.

Tested on QEMU with 775K, 1MB and up to 59MB file sizes, no more errors found. I'm finally feeling pretty good about large MINIX filesystems again. Very glad to be done with this one!

Mellvik:
>  
> Last push fixes three major bugs in mfs:
> 
> Invalid image created when any block number > 32767 (any filesystem larger than 32MB)
> Invalid zones created on any file requiring double indirect blocks (any file > 519K)
> Invalid image created when any files > 775K were included in filesystem
> This last bug took way too long to figure out, but ended up being a single line change, where an indirect block buffer array was defined using host int instead of unsigned short, which caused critical failure on any double indirect block number > 255 (=775K filesize = 7 direct + 512 indirect + 256 dbl indirect), ending up creating filesystems 4GB in size through seeking to negative numbers. I think I worked on this straight for two days before finding out what was happening. The good news is I understand MINIX indirect and double indirect blocks real well now :)
> 
> Removed printk's from truncate, everything seems to work well now. Debug statements left in and available in mfs or kernel through define DEBUG.
> 
> Also, make images now creates a 64MB MINIX image for testing called hd64-minix.img. This will remain as part of the standard build, since we obviously need to test filesystem edge cases much more thoroughly.
> 
> Tested on QEMU with 775K, 1MB and up to 59MB file sizes, no more errors found. I'm finally feeling pretty good about large MINIX filesystems again. Very glad to be done with this one!
> 
> 
> 
Great news - and good work, @ghaerr!

Will test - but when it works on QEMU there is no reason it would not work on physical. No mfs testing yet.

I have a different killer test coming up - maxing out the use of inodes with thousands of small files - with increasing sizes.

Should work fine.


Thank you.

—M

Issue #1390, opened 2022-08-03
Suggestion: Package Manager.


Sasytoremember:
As I see most people who have ELKS on their PC have Ethernet cards, and there being a tiny amount of programs bundled with elks, a package manager would be cool to see.

ghaerr:
Hello @Sasytoremember,

Thank you for the suggestion! Although this is a good idea, I fear we it may be a while before ELKS is ready for than outside package manageris, mostly because of 1) compilation outside of ELKS isn't currently supported well (although I'm working on that), and 2) one must setup and compile a special toolchain, and the C library for that toolchain and ELKS live inside the ELKS source tree. (Moving  libc out of the ELKS tree has its own issues, so that may not happen as quickly).

That said, it would be nice to be able to install newer versions of programs already built from the ELKS tree, without having to replace the existing running filesystem with the default built disk image. I think @Mellvik is working on something like that, which uses the network for transport and installation. (Perhaps this is what you were requesting in the first place? :)

Thank you!

Mellvik:
Thank you @ghaerr, hi @linuxbash, hi @sasytoremember.

Yes, that would be something! As ELKS  grows it does indeed get harder to keep track of things, in particular for visitors dropping by every now and then, but also for those of us living with ELKS  daily. That said, ELKS is still small (and smaller than many, maybe most package managers) and the requirements are different from larger systems. The entire system is really a package. IOW a package manager is probably overkill.

Your question did trigger some thinking though. And like @ghaerr points out, I've been experimenting with various approaches to simplify updating running systems with new binaries. Pre-built HD images are fine for emulators, and the floppy images are fine for physical systems, but a running system needs a way to be kept up to date without overwriting local configurations. So far, ftp and tar files have been sufficient, but a more automated regime would be desirable. BTW - by 'automated' I don't mean automatic updates of a running system, but more or less automatic building of the components required to make such updates simple.

Towards that, here are some ideas covering what I've been missing:

- An automatically generated tar file of a 'complete' system, built along with the HD images.
	A 'complete' system should possibly be configurable - or include all the stuff that goes onto hd images (which I've never used so I don't really know :-) ) - binaries that don't make it on floppy images, man-pages, /lib files like bootblock.bin, etc.

- A script on ELKS that fetches this file via ftp/urlget and installs updated files (I'm not sure the ELKS tar command has a 'dont-overwrite-newer' option, we should add that).

[...]

PR #1419, opened 2022-08-25
[cmds] New and updated commands, man-pages

Mellvik:
Added `split` command (source from BSD4.3reno), very useful for (among other things) file system testing: Generate 10.000 files in a minute...
Includes `man page`.

Updated `df` command to list all mounted file systems when entered without arguments (borrowing some code from `mount`). Previously listed only the root file system. The new behaviour incidentally matches the `man page` - which has been adjusted. Also -
- Removed the old (and non-standard) 'file' mode which was essentially the same as the inode mode.
- Added support for FAT file systems in the default listing, other listings (POSIX, inode) are not meaningful for FAT

@ghaerr - I'm sure there is a fast way to figure out the type of a mounted file system without calling `ustat` and waiting forever. Hints? That would make the usability of df when there are mount FAT filesystems much better.
Actually a Unix/Linux style `mtab` file maintained by `mount` which always has the mounted filesystems and their parameters, would be convenient, but possibly overkill.

Also, when creating lots of files i a directory, `ls` and other utilities run out of heap pretty fast. It would be tempting to use a far segment for increased buffering. How do we `read()`  into a far segment?

Output from `df`: 
```
elks17# ./df
Filesystem    1k-Blocks     free     used    %  FUsed%  Mounted on
/dev/hda3         65535    61443     4092   7%     6%   /
/dev/fd0           1440      380     1060  74%    73%   /fd
/dev/hda1        348226   345512     2714   1%          /mnt (FAT)
/dev/hdb1           726      554      172  24%          /mnt5 (FAT)
/dev/hda2         65535    59812     5723   9%     2%   /mnt3
/dev/hda4         41328    37129     4199  11%     3%   /mnt4
elks17# ./df -i
Filesystem       Inodes     IUsed    IFree    %IUsed   Mounted on
/dev/hda3         21845      1187    20658        6%   /
/dev/fd0            256       185       71       73%   /fd
/dev/hda1     -- Not a MINIX filesystem
/dev/hdb1     -- Not a MINIX filesystem
/dev/hda2         21845       228    21617        2%   /mnt3
/dev/hda4         13776       313    13463        3%   /mnt4
elks17# ./df -P
Filesystem     512-blocks    Used  Available  Capacity   Mounted on
/dev/hda3        131070      8184   122886        7%     /
/dev/fd0           2880      2120      760       74%     /fd
/dev/hda1     -- Not a MINIX filesystem
/dev/hdb1     -- Not a MINIX filesystem
/dev/hda2        131070     11446   119624        9%     /mnt3
/dev/hda4         82656      8398    74258       11%     /mnt4
elks17# 

```

### DIFF
--- a/elkscmd/file_utils/Makefile
+++ b/elkscmd/file_utils/Makefile
@@ -13,7 +13,7 @@ include $(BASEDIR)/Make.rules
 # note: grep doesn't read stdin, replaced with minix1/grep
 # removed: l
 PRGS = ln ls mkdir mkfifo mknod more mv rm rmdir sync touch \
-	cat chgrp chmod chown cmp cp dd df grep
+	cat chgrp chmod chown cmp cp dd df grep split
 
 all: $(PRGS)
 
@@ -73,6 +73,9 @@ rm: rm.o $(TINYPRINTF)
 
 rmdir: rmdir.o
 	$(LD) $(LDFLAGS) -o rmdir rmdir.o $(LDLIBS)
+
+split: split.o
+	$(LD) $(LDFLAGS) -o split split.o $(LDLIBS)
 
 sync: sync.o
 	$(LD) $(LDFLAGS) -o sync sync.o $(LDLIBS)
--- a/elkscmd/file_utils/df.c
+++ b/elkscmd/file_utils/df.c
@@ -22,20 +22,33 @@
 #include <stdlib.h>
 #include <string.h>
 #include <dirent.h>
+#include <sys/mount.h>
 #include <linuxmt/minix_fs.h>
+#include <linuxmt/fs.h>
 
 #define BLOCK_SIZE	1024
 
-int df(char *device);
+int df(char *device, char *mpnt);
 
 typedef unsigned int bit_t;
 bit_t bit_count(unsigned blocks, bit_t bits, int fd);
-char *devname(char *dirname);
+struct dnames {
+	char *name;
+	char *mpoint;
+};
+struct dnames *devname(char *);
+static char *dev_name(dev_t);
 
 int iflag= 0;	/* Focus on inodes instead of blocks. */
 int Pflag= 0;	/* Posix standard output. */
 int kflag= 0;	/* Output in kilobytes instead of 512 byte units for -P. */
 int istty;	/* isatty(1) */
+
+/* (num / tot) in percentages rounded up. */
+#define percent(num, tot)  ((int) ((100L * (num) + ((tot) - 1)) / (tot)))
+
+/* One must be careful printing all these _t types. */
+#define L(n)	((long) (n))
 
 void usage(void)
 {
@@ -47,6 +60,8 @@ int main(int argc, char *argv[])
 {
   char *device = "/";
   char *blockdev;
+  struct dnames *dname;
+  struct statfs statfs;
 
   while (argc > 1 && argv[1][0] == '-') {
   	char *opt= argv[1]+1;
@@ -56,8 +71,7 @@ int main(int argc, char *argv[])
   		case 'i':	iflag= 1;	break;
   		case 'k':	kflag= 1;	break;
   		case 'P':	Pflag= 1;	break;
-		default:
-			usage();
+		default:	usage();
 		}
 	}
 	argc--;
@@ -69,38 +83,49 @@ int main(int argc, char *argv[])
   if (argc > 1)
 	device = argv[1];
 
-  if (!(blockdev = devname(device))) {
+  dname = devname(device);
+  if (!(blockdev = dname->name)) {
 	fprintf(stderr, "Can't find /dev/ device for %s\n", device);
 	exit(1);
   }
 
-  if (Pflag) {
-	printf(!iflag ? "\
-Filesystem    %4d-blocks    Used  Available  Capacity\n" : "\
-Filesystem       Inodes     IUsed    IFree    %%IUsed\n",
-		kflag ? 1024 : 512);
+  if (iflag) {
+	printf("Filesystem       Inodes     IUsed    IFree    %%IUsed   Mounted on\n");
   } else {
-	printf("%s\n", !iflag ? "\
-Filesystem    1k-Blocks     free     used    %  FUsed%" : "\
-Filesystem        Files     free     used    %  BUsed%"
-	);
+	if (!Pflag) 
+		printf("Filesystem    1k-Blocks     free     used    %%  FUsed%%  Mounted on\n");
+	else 
+		printf("Filesystem    %4d-blocks    Used  Available  Capacity   Mounted on\n",
+			kflag ? 1024 : 512);
   }
 
-
-  return df(blockdev);
+  if (argc == 1) {	/* loop through all mounted devices */
+	int i;
+  	for (i = 0; i < NR_SUPER; i++) {
+		if (ustatfs(i, &statfs) >= 0) {
+			char *nm = dev_name(statfs.f_dev);
+			if (statfs.f_type > 2 || (statfs.f_type == 2 && (Pflag||iflag))) 
+				printf("%s     -- Not a MINIX filesystem\n", nm);
+			else if (statfs.f_type == 2 && !Pflag)
+				printf("%-15s %7ld  %7ld  %7ld %3d%%          %s (FAT)\n", nm,\
+				    statfs.f_blocks, statfs.f_bfree,\
+				    statfs.f_blocks-statfs.f_bfree,\
+				    percent(statfs.f_blocks-statfs.f_bfree, statfs.f_blocks),\
+				    statfs.f_mntonname);
+			else
+				df(nm, statfs.f_mntonname);
+		}
+	}
+	return 0;
+  } else
+  	return df(blockdev, dname->mpoint);
 }
 
-/* (num / tot) in percentages rounded up. */
-#define percent(num, tot)  ((int) ((100L * (num) + ((tot) - 1)) / (tot)))
-
-/* One must be careful printing all these _t types. */
-#define L(n)	((long) (n))
-
-int df(char *device)
+int df(char *device, char *mpnt)
 {
   int fd;
   bit_t i_count, z_count;
-  block_t totblocks, busyblocks;
+  long totblocks, busyblocks;
   int n;
   struct minix_super_block super, *sp;
 
@@ -153,26 +178,28 @@ int df(char *device)
 
   /* Print results. */
   printf("%s", device);
-  n= strlen(device);
+  n = strlen(device);
   if (n > 15 && istty) { printf("\n"); n= 0; }
   while (n < 15) { printf(" "); n++; }
 
+  if (iflag) {
+	printf(" %7ld   %7ld  %7ld     %4d%%   %s\n",
+		L(sp->s_ninodes),			/* Inodes */
+		L(i_count),				/* IUsed */
+		L(sp->s_ninodes - i_count),		/* IAvail */
+		percent(i_count, sp->s_ninodes),	/* Capacity */
+		mpnt					/* Mount pnt */
+
+	);
+  }
   if (!Pflag && !iflag) {
-	printf(" %7ld  %7ld  %7ld %3d%%   %3d%%\n",
+	printf(" %7ld  %7ld  %7ld %3d%%   %3d%%   %s\n",
 		L(totblocks),				/* Blocks */
 		L(totblocks - busyblocks),		/* free */
 		L(busyblocks),				/* used */
 		percent(busyblocks, totblocks),		/* % */
-		percent(i_count, sp->s_ninodes)	/* FUsed% */
-	);
-  }
-  if (!Pflag && iflag) {
-	printf(" %7ld  %7ld  %7ld %3d%%   %3d%%\n",
-		L(sp->s_ninodes),			/* Files */
-		L(sp->s_ninodes - i_count),		/* free */
-		L(i_count),				/* used */
-		percent(i_count, sp->s_ninodes),	/* % */
-		percent(busyblocks, totblocks)		/* BUsed% */
+		percent(i_count, sp->s_ninodes),	/* FUsed% */
+		mpnt					/* Mount point */
 	);
   }
   if (Pflag && !iflag) {
@@ -181,19 +208,12 @@ int df(char *device)
   		totblocks *= 2;
   		busyblocks *= 2;
 	}
-	printf(" %7ld   %7ld  %7d     %4d%%\n",
+	printf(" %7ld   %7ld  %7ld     %4d%%     %s\n",
 		L(totblocks),				/* Blocks */
 		L(busyblocks),				/* Used */
 		totblocks - busyblocks,			/* Available */
-		percent(busyblocks, totblocks)		/* Capacity */
-	);
-  }
-  if (Pflag && iflag) {
-	printf(" %7ld   %7ld  %7ld     %4d%%\n",
-		L(sp->s_ninodes),			/* Inodes */
-		L(i_count),				/* IUsed */
-		L(sp->s_ninodes - i_count),		/* IAvail */
-		percent(i_count, sp->s_ninodes)	/* Capacity */
+		percent(busyblocks, totblocks),		/* Capacity */
+		mpnt					/* Mount point */
 	);
   }
   close(fd);
@@ -245,17 +265,22 @@ bit_t bit_count(unsigned blocks, bit_t bits, int fd)
  * $PchId: df.c,v 1.7 1998/07/27 18:42:17 philip Exp $
  */
 
-/* return /dev/ device from dirname*/
-char *devname(char *dirname)
+/* return /dev/ device from dirname */
+struct dnames *devname(char *dirname)
 {
    static char dev[] = "/dev";
    struct stat st, dst;
    DIR  *fp;
    struct dirent *d;
+   static struct statfs statfs;
+   static struct dnames dn;
    static char name[16]; /* should be MAXNAMLEN but that's overkill */
 
-   if (!strncmp(dirname, "/dev/", 5))
-	return dirname;
+   if (!strncmp(dirname, "/dev/", 5)) {
+	dn.name = dirname;
+	dn.mpoint = dirname;
+	return &dn;
+   }
 
    if (stat(dirname, &st) < 0)
       return 0;
@@ -271,12 +296,60 @@ char *devname(char *dirname)
          continue;
       strcpy(name + sizeof(dev), d->d_name);
       if (stat(name, &dst) == 0) {
-		if (st.st_dev == dst.st_rdev) {
-			closedir(fp);
-			return name;
+	 if (st.st_dev == dst.st_rdev) {
+	     if (ustatfs(st.st_dev, &statfs) < 0) {
+		dn.mpoint = NULL;
+	     } else {
+		dn.mpoint = statfs.f_mntonname;
+	     }
+	     closedir(fp);
+	     dn.name = name;
+	     return &dn;
         }
       }
    }
    closedir(fp);
-   return 0;
+   return NULL;
 }
+
+/*
+ * Convert a block device number to name.
+ * From mount.c
+ */
+static struct dev_name_struct {
+        char *name;
+        dev_t num;
+} devices[] = {
+        /* root_dev_name needs first 5 in order*/
+        { "hda",     0x0300 },
+        { "hdb",     0x0320 },
+        { "hdc",     0x0340 },
+        { "hdd",     0x0360 },
+        { "fd0",     0x0380 },
+        { "fd1",     0x03a0 },
+        { "ssd",     0x0200 },
+        { "rd",      0x0100 },
+        { NULL,           0 }
+};
+
+static char *dev_name(dev_t dev)
+{
+	int i;
+#define NAMEOFF 5
+	static char name[10] = "/dev/";
+
+	for (i=0; i<sizeof(devices)/sizeof(devices[0])-1; i++) {
+		if (devices[i].num == (dev & 0xfff0)) {
+			strcpy(&name[NAMEOFF], devices[i].name);
+			if (i < 4) {
+				if (dev & 0x07) {
+					name[NAMEOFF+3] = '0' + (dev & 7);
+					name[NAMEOFF+4] = 0;
+				}
+			}
+			return name;
+		}
+	}
+	return NULL;
+}
+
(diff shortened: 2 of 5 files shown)


ghaerr:
Hello @Mellvik,

Thanks for the additions and enhancements! I take it the main enhancement to `df` is to report on all mounted filesystems, unless an argument is given? Nice.

I have the following comments:
- When the filesystem is FAT, can't we get the free space and blocks used from `statfs`, and use that rather than saying "Not a MINIX filesystem? (I realize I'm the guy who stuck the current hack into `df`)! `df` does its own space calculations, but now that `statfs` is (back) in the kernel, perhaps that should be used instead of copying the bit-lookup code in `df`.
- What is the problem with `ustatfs`? Is it taking forever because of the FAT and/or MINIX free space calculations? If so, perhaps we can send an argument over to the system call to not ask for the free space to be calculated (and call it again if wanted). The original purpose of `ustatfs` was to enumerate filesystems. `statfs` was then added, which also does free space calculations, so perhaps `ustatfs` shouldn't. I need to look back into exactly how both work.
- The 1440k floppy image is completely full, we need to drop some programs. I'll figure that out, so don't change elkscmd/Applications yet, I just wanted you to know.
- At some point, we might want a library routine to translate from device name to dev_t. This is fine for now, but now both `mount` and `df` will use hard-coded translate functions.

> Actually a Unix/Linux style mtab file maintained by mount which always has the mounted filesystems and their parameters, would be convenient, but possibly overkill.

I think that's overkill, but more specifically I would like to try to keep the root filesystem from being written to by various commands (unlike Linux), so that it doesn't get needlessly dirty and is faster on floppy.

The `df` options look great (`-i` and `-P`). While we're on the subject of 1K vs 512-byte "blocks", I was thinking that it would be nice if ELKS *always* reported 1K blocks and only reported 512 byte blocks when used with an option. Do you agree, and does `du` operate the same way (meaning reports 1K blocks)? I find it very hard to switch between 512 and 1K blocks, and its FAR easier to see 1K blocks, IMO. I know that older `du` programs used to report in 512-byte blocks, thereby making it very hard to actually see the amount of disk space used.

I'd like to study this a bit and would like to hear your responses to the questions, but a quick review shows this PR should be ready to commit without further changes. Thank you!

ghaerr:
@Mellvik,

I read both your and the kernel code a little more, and perhaps can answer some of my own questions: I'm thinking that the original MINIX filesystem free and used space calculation in `df` is there to allow displaying an unmounted filesystem, so it should probably stay in.

For your new code, `ustatfs` will do an internal `statfs` on the filesystem, which works for both MINIX and FAT. Thus, you should be able to display the used and free blocks accurately for FAT, rather than displaying "Not a MINIX filesystem". However, we may still want a flag to to `ustatfs` (and possibly `df`) that would stop calculation of free space for FAT filesystems, since it takes so long. In that case, `df` could just display `unavailable, use -f` or something.

What Microsoft did for MSDOS/Windows FAT filesystems is actually record the free space for large hard drives in a special sector on the drive (FAT32 only). We could do that also, but that doesn't solve the problem of perhaps taking too long to calculate FAT12 and FAT16 file sizes. Let me know what think we should do, for both the kernel and `df`.

Thank you!

Mellvik:
> 
> Thanks for the additions and enhancements! I take it the main enhancement to df is to report on all mounted filesystems, unless an argument is given? Nice.
> 
Yes - and a bit more: to make df as useful a tool as on other (unix/linux) systems.
> I have the following comments:
> 
> When the filesystem is FAT, can't we get the free space and blocks used from statfs, and use that rather than saying "Not a MINIX filesystem? (I realize I'm the guy who stuck the current hack into df)! df does its own space calculations, but now that statfs is (back) in the kernel, perhaps that should be used instead of copying the bit-lookup code in df.
That's what it does. The 'not minix' message is displayed when there is nothing relevant to display. Like inodes. The alternative us to leave out non-minix filesystems from the list.
> What is the problem with ustatfs? Is it taking forever because of the FAT and/or MINIX free space calculations? If so, perhaps we can send an argument over to the system call to not ask for the free space to be calculated (and call it again if wanted). The original purpose of ustatfs was to enumerate filesystems. statfs was then added, which also does free space calculations, so perhaps ustatfs shouldn't. I need to look back into exactly how both work.
Yes, that would do it. Avoiding the calculation is the thing. I havent looked at what mount does when mounting a fat fs, but it reports the type and size quite fast.
> The 1440k floppy image is completely full, we need to drop some programs. I'll figure that out, so don't change elkscmd/Applications yet, I just wanted you to know.
I realize that, which brings back the package question: a convenient way to add 'the rest' - or populate a partition via the network.
> At some point, we might want a library routine to translate from device name to dev_t. This is fine for now, but now both mount and df will use hard-coded translate functions.
Agreed, this is kinda clumsy and inflexible.
> Actually a Unix/Linux style mtab file maintained by mount which always has the mounted filesystems and their parameters, would be convenient, but possibly overkill.
> 
> I think that's overkill, but more specifically I would like to try to keep the root filesystem from being written to by various commands (unlike Linux), so that it doesn't get needlessly dirty and is faster on floppy.
> 
I agree. There are ways around those deficiencies, but that would be for a slow day.
> The df options look great (-i and -P). While we're on the subject of 1K vs 512-byte "blocks", I was thinking that it would be nice if ELKS always reported 1K blocks and only reported 512 byte blocks when used with an option.
> 
That's the way df has been working and still works. The 512b display is for posix compliance (-P).
> Do you agree, and does du operate the same way (meaning reports 1K blocks)? I find it very hard to switch between 512 and 1K blocks, and its FAR easier to see 1K blocks, IMO. I know that older du programs used to report in 512-byte blocks, thereby making it very hard to actually see the amount of disk space used.
> 
Yes, du is defaulting to 1k too.
> I'd like to study this a bit and would like to hear your responses to the questions, but a quick review shows this PR should be ready to commit without further changes. Thank you!
> 
Thank you. 

-M

Mellvik:
> 
> I read both your and the kernel code a little more, and perhaps can answer some of my own questions: I'm thinking that the original MINIX filesystem free and used space calculation in df is there to allow displaying an unmounted filesystem, so it should probably stay in.
> 
You must be seeing something I don't. On other systems df -i means display inode stats, the filesystem free does that, but in an unusual way. The same data are presented in the inode stat output. Not that I don't like change, but The goal here is to make life easy, meaning have df -i act as 'expected'.
> For your new code, ustatfs will do an internal statfs on the filesystem, which works for both MINIX and FAT. Thus, you should be able to display the used and free blocks accurately for FAT, rather than displaying "Not a MINIX filesystem".
> 
As pointed out in the previous message, df is already doing that.
> However, we may still want a flag to to ustatfs (and possibly df) that would stop calculation of free space for FAT filesystems, since it takes so long. In that case, df could just display unavailable, use -f or something.
> 
It may be argued that if you're asking for inode stats you know what you're doing so omitting fat filesystems silently would be ok. Those omitted would show in the regular display anyway, and are marked with 'FAT'. Big systems (e.g Linux) have a -t (type) option to restrict the display to a specific fs type... which may be a way to go.
> What Microsoft did for MSDOS/Windows FAT filesystems is actually record the free space for large hard drives in a special sector on the drive (FAT32 only). We could do that also, but that doesn't solve the problem of perhaps taking too long to calculate FAT12 and FAT16 file sizes. Let me know what think we should do, for both the kernel and df.
> 
Ah, so that's why every dir command in dos can end with a display of available space apparently w/o significant penalty. Well, no, it cannot be if it's fat32 only. Anyway, yes I do think this is a sufficiently significant problem to spend some time on. Unfortunately I'm blank on fat...

Thanks.


-M

ghaerr:
Thanks for your clarifications, @Mellvik. You've done a great job on this.

> You must be seeing something I don't. 

I was just talking about the `bit_count` functions (and `df()`) in `df` which calculate their own MINIX filesystem used and free blocks, instead of using `statfs` or `ustatfs`. That gives the ability to display non-mounted filesystem data (which it still does when given an argument). That also implies that when given an argument that is a mounted FAT filesystem, it won't work, since the included code only works for MINIX. Not sure what we should do about this case, except for perhaps trying a `statfs` and then using internal `df()` function if `statfs` fails.

> Also, when creating lots of files i a directory, ls and other utilities run out of heap pretty fast. It would be tempting to use a far segment for increased buffering. How do we read() into a far segment?

We can't directly read int a far segment, but `ls` is being allowed only 20k heap. I suggest having you add `-maout-heap=0xffff` to `ls` file_utils/Makefile and any other utilities that need more heap, and going from there.

I suggest we commit this now, then we can talk more details on how to make a ustatfs or statfs mod that doesn't do free space calculations... that OK with you?

Thank you!

ghaerr:
A quick code review (which you can fix in a subsequent PR):

df.c:
- Lines 107, 109: instead of comparing for "2", compare against FST_MSDOS, which is defined in sys/mount.h (already included).
- Lines 110-113: probably ought to remove the trailing backslashes at the end of the lines; these aren't needed, and cause trouble when/if the code is commented out using //, for instance.

With regards to `ustatfs` enhancements, if the 2nd argument (i.e. the statfs buffer) is NULL, the index is just checked for a valid mount index, and returns 0 (or a -ERRNO). I'm thinking we could return FST_MINIX, FST_MSDOS or FST_ROMFS instead of 0, which you could use the know whether its FAT or not, in a previous loop.

Looking a bit deeper into `statfs`, it turns out ELKS doesn't actually have that system call (yet). What happened is that I used the equivalent half-programmed `statfs` kernel entry points per filesystem, but just hooked that up to `ustatfs`. So at this point, it'd be just as well to not introduce a new system call, but instead probably pass a third argument to `ustatfs` which would be a flag word, indicating whether free space is desired or not (which would then be wired to a `df` argument, I suppose). This could be automated so that MINIX filesystems always show the free space, if `ustatfs(index, NULL, 0)` were used to determine filesystem type beforehand.

What do you think? The change to `ustatfs` is pretty minor if you want to go that route.

ghaerr:
> Avoiding the calculation is the thing. I havent looked at what mount does when mounting a fat fs, but it reports the type and size quite fast.

The FAT mount code used to calculate free space on mount, but that was removed due to complaints about how long it took on HD images. That code was added to `ustatfs` which currently always calculates free space. So, after this PR is committed, both `df` and `mount` will calculate free space on FAT (and MINIX), as they are the only programs using `ustatfs`. 

If we make the suggested mod to `ustatfs` and add a flags argument, then we will need to decide what `mount` should do vs `df`, especially since they both report the statfs used and free blocks at this time. (Do we need options for both to not calc free space, or just on `df`)?

Mellvik:
> > You must be seeing something I don't.
> 
> I was just talking about the `bit_count` functions (and `df()`) in `df` which calculate their own MINIX filesystem used and free blocks, instead of using `statfs` or `ustatfs`. That gives the ability to display non-mounted filesystem data (which it still does when given an argument). That also implies that when given an argument that is a mounted FAT filesystem, it won't work, since the included code only works for MINIX. Not sure what we should do about this case, except for perhaps trying a `statfs` and then using internal `df()` function if `statfs` fails.

I see, I didn't consider the not-mounted fs usage:
```
elks17# df /dev/hda1
Filesystem    1k-Blocks     free     used    %  FUsed%  Mounted on
df: /dev/hda1: Not a MINIX file system
elks17# 
```
Admittedly it does not look good, and I think the sane thing to do is to report the error with something like 'FAT file systems must be mounted to use df'.
 
> 
> > Also, when creating lots of files i a directory, ls and other utilities run out of heap pretty fast. It would be tempting to use a far segment for increased buffering. How do we read() into a far segment?
> 
> We can't directly read int a far segment, but `ls` is being allowed only 20k heap. I suggest having you add `-maout-heap=0xffff` to `ls` file_utils/Makefile and any other utilities that need more heap, and going from there.

OK, I'll experiment with this one. `ls` (or shell wildcard expansion) with 1000 or 3000 files is a challenge. 

> 
> I suggest we commit this now, then we can talk more details on how to make a ustatfs or statfs mod that doesn't do free space calculations... that OK with you?
> 
Yes, that's good.
Thank you.

-M

Mellvik:
> A quick code review (which you can fix in a subsequent PR):
> 
> df.c:
> 
> * Lines 107, 109: instead of comparing for "2", compare against FST_MSDOS, which is defined in sys/mount.h (already included).
> * Lines 110-113: probably ought to remove the trailing backslashes at the end of the lines; these aren't needed, and cause trouble when/if the code is commented out using //, for instance.

Thank you - good points. Actually, the way this is coded right now, introducing a new file system type (minix or unix like), it would have to be numbered below FST_MSDOS.

> 
> With regards to `ustatfs` enhancements, if the 2nd argument (i.e. the statfs buffer) is NULL, the index is just checked for a valid mount index, and returns 0 (or a -ERRNO). I'm thinking we could return FST_MINIX, FST_MSDOS or FST_ROMFS instead of 0, which you could use the know whether its FAT or not, in a previous loop.
> 
> Looking a bit deeper into `statfs`, it turns out ELKS doesn't actually have that system call (yet). What happened is that I used the equivalent half-programmed `statfs` kernel entry points per filesystem, but just hooked that up to `ustatfs`. So at this point, it'd be just as well to not introduce a new system call, but instead probably pass a third argument to `ustatfs` which would be a flag word, indicating whether free space is desired or not (which would then be wired to a `df` argument, I suppose). This could be automated so that MINIX filesystems always show the free space, if `ustatfs(index, NULL, 0)` were used to determine filesystem type beforehand.
> 
> What do you think? The change to `ustatfs` is pretty minor if you want to go that route.

Yes, this does indeed look like the optimal route at this time! 

--M

Mellvik:
> > Avoiding the calculation is the thing. I havent looked at what mount does when mounting a fat fs, but it reports the type and size quite fast.
> 
> The FAT mount code used to calculate free space on mount, but that was removed due to complaints about how long it took on HD images. That code was added to `ustatfs` which currently always calculates free space. So, after this PR is committed, both `df` and `mount` will calculate free space on FAT (and MINIX), as they are the only programs using `ustatfs`.
> 
> If we make the suggested mod to `ustatfs` and add a flags argument, then we will need to decide what `mount` should do vs `df`, especially since they both report the statfs used and free blocks at this time. (Do we need options for both to not calc free space, or just on `df`)?

With `df` now doing the full stats listing(s), I think it would be good to take that away from `mount`, and have it work like  on other `*x` systems: Simply report mounted file systems (and types) when called with no arguments. If active options can be included, that would be a welcome enhancement.

Thank you.

-M

Mellvik:
>> The 1440k floppy image is completely full, we need to drop some programs. I'll figure that out, so don't change elkscmd/Applications yet, I just wanted you to know.

> I realize that, which brings back the package question: a convenient way to add 'the rest' - or populate a partition via the network.

Actually, I must disagree with myself on this one :-). My 1440k images have plenty of space, almost 400k to be exact. It's of course a matter of what each user chooses in menuconfig. So smart grouping of the utilities will still take us a long way. Leaving out the nano-stuff which I always do, helps a lot. Maybe a 'developer tools' group which includes `disasm`, `split`, `sercat`, `fdtest` and others.

Thank you.

-- M

Skjermbilde 2022-08-26 kl  12 13 19

ghaerr:
I'll work on an option to `ustatfs` to skip time-expensive calculations of free space on FAT volumes. When requested, the returned free space value will be -1L (to differentiate from 0). This could then be handled specially by displaying blanks, rather than -1.

I'm in agreement with your idea that `df` report calculated free space status, while `mount` would pass the skip-flag to `ustatfs`. This would allow for fast free space calculations using `mount`. 

`df` should probably be enhanced to try `ustatfs` on a passed argument, and if that fails, the call the `df()` internal routine. This change would entail moving your new display code into a function, much like `df()`. We probably have the best of both worlds at that point, unless we wanted to also move the FAT filesystem used/free code into `df` also, which probably isn't worth it.

ghaerr:
> ls (or shell wildcard expansion) with 1000 or 3000 files is a challenge.

WRT to "unlimited arguments" for wildcard expansion and/or internal heap space - we can't win that battle with ELKS. All arguments must fit within the stack area prior to main's argc/argv (above the environment), and that further reduces the command's heap space. So for `ls` it'd be a double-whammy. Same goes for memory internal to commands - ELKS only has 64k data per application, and the argv vector can't be a far pointer. I guess what I'm saying is we're likely to see lots of problems when there are 3000 files in a directory, and those uses may only be good for unit testing the filesystem, rather than trying to enhance our commands to work in that highly unusual environment.

Another item to consider when adding, say `-maout-heap=0xffff` to a command like `ls`: when the system available memory is low, `ls` currently can run in only 30k (since it only currently requests 20k heap). Requesting 64k means that when system memory is low, `ls` execution will be rejected if there is not 64k + code available. So there's a cost to adding 64K requirement to each command, especially when used in a pipeline. This is the reason why we've chosen to NOT ask for max memory for each command automatically - so that ELKS functions well in a low memory environment.

Because of this, I'm ambivalent about whether `ls` should require 64k to run, just for possibly working better in an unusual testing environment, for such an important command. With networking running, a user running `vi`, and a serial terminal logged in, we're frequently down to less than 100K available.

> So smart grouping of the utilities will still take us a long way. 

Agreed. I'm beginning to think our application selection mechanism is becoming broken, as the two completely different methods (select by image size vs select by application class) don't work well together, and the application classes are very antiquated (with the recently added CONFIG_TUI making things worse, which I'm thinking of reverting).

A reason the two selection methods don't work well is that the "make images" method of creating each of our 12+ or so MINIX or FAT images never uses the application class but only the image size when building. So those images are never the same as the "make" image.

I'll take a pass at adding `split` and possibly removing nano-X applications. I don't like removing the nano-X stuff because a lot of people like games and graphical things, especially on first evaluative use of ELKS. I've also personally fallen out of touch with the 1440k image, because it won't fit the man pages and other things we're currently working on, so my default build is now 2880k (and QEMU testing, since few systems have 2880k floppies). The whole problem of application selection probably needs to move to its own issue.

Mellvik:
Thank you @ghaerr.

> I'll work on an option to `ustatfs` to skip time-expensive calculations of free space on FAT volumes. When requested, the returned free space value will be -1L (to differentiate from 0). This could then be handled specially by displaying blanks, rather than -1.

Great, this is good.

> 
> I'm in agreement with your idea that `df` report calculated free space status, while `mount` would pass the skip-flag to `ustatfs`. This would allow for fast free space calculations using `mount`.

This is OK, my idea was slightly different (remove all space/stats from `mount` to make it more like other *x systems). However, this may be even better, providing an always available fast lane. 
Still, `df` should go the fast way (no calculations) when the -i or -P options are used, which leave out FAT file systems

> 
> `df` should probably be enhanced to try `ustatfs` on a passed argument, and if that fails, the call the `df()` internal routine. This change would entail moving your new display code into a function, much like `df()`. We probably have the best of both worlds at that point, unless we wanted to also move the FAT filesystem used/free code into `df` also, which probably isn't worth it.

Agreed. Actually, it would have been extremely convenient to have a macro or library routine that returned the FS type. 
- In some cases all we want is to know the type, no sizes or stats. 
- In other cases we want the whole shebang. 
- And finally, there are the cases where we want to know the type only to choose whether to accept the FAT calculation delay or not.

Getting there.

--M

ghaerr:
I'll add the following enhancement to `ustatfs`:

> In some cases all we want is to know the type, no sizes or stats.
```
int type = ustatfs(index, NULL, 0);
```
> In other cases we want the whole shebang.
```
ustatfs(index, &statfsbuf, 0);
```
> And finally, there are the cases where we want to know the type only to choose whether to accept the FAT calculation delay or not.
```
if (type == FST_MSDOS)
    ustatfs(index, &statfsbuf, UF_NOFREESPACE);
```

Mellvik:
Great, thank you @ghaerr.

I'll test and submit an updated ´df´ - and ´mount´ unless you have other plans. (and man pages)

—M

Mellvik:
> > ls (or shell wildcard expansion) with 1000 or 3000 files is a challenge.
> 
...
> Because of this, I'm ambivalent about whether `ls` should require 64k to run, just for possibly working better in an unusual testing environment, for such an important command. With networking running, a user running `vi`, and a serial terminal logged in, we're frequently down to less than 100K available.

I agree on all accounts. Keeping `ls` lean is the priority. What I can do to continue my testing is 
- make a special version of `ls` for this particular purpose (no buffering, no sorting, always assuming max name length, always listing files as they appear in the directory. And just keep that as a debugging tool. OR
- Add an undocumented option to `ls` that does that (which requires minimal coding). OR
- Minor update to `sash` `ls` which works with any size directory already, but has limited options.

> 
> > So smart grouping of the utilities will still take us a long way.
> 
> Agreed. I'm beginning to think our application selection mechanism is becoming broken, as the two completely different methods (select by image size vs select by application class) don't work well together, and the application classes are very antiquated (with the recently added CONFIG_TUI making things worse, which I'm thinking of reverting).
> 
> A reason the two selection methods don't work well is that the "make images" method of creating each of our 12+ or so MINIX or FAT images never uses the application class but only the image size when building. So those images are never the same as the "make" image.
> 
> I'll take a pass at adding `split` and possibly removing nano-X applications. I don't like removing the nano-X stuff because a lot of people like games and graphical things, especially on first evaluative use of ELKS. I've also personally fallen out of touch with the 1440k image, because it won't fit the man pages and other things we're currently working on, so my default build is now 2880k (and QEMU testing, since few systems have 2880k floppies). The whole problem of application selection probably needs to move to its own issue.

Agreed, maybe continue on #1390 ?

--M

ghaerr:
> Add an undocumented option to ls that does that (which requires minimal coding).

`ls` is a huge memory hog and allocs way too much memory, so I don't recommend this option.

> Minor update to sash ls which works with any size directory already, but has limited options.

Possibly good idea, except that sash also calls realloc for each directory entry (likely for qsort). 
We also have problems with `realloc`: current version never reallocs in-place, always reallocs total all over again plus extra, then copies. This caused `fm` to run out of memory. I stuck in a v7malloc.c temp fix, but can't recommend using `realloc` for the time being. This is on my list to fix soon systemwide.

> make a special version of ls for this particular purpose (no buffering, no sorting, always assuming max name length, always listing files as they appear in the directory. And just keep that as a debugging tool. 

Looking at the `sash` ls, and just copying that as a separate debug tool, or perhaps a sash option to not sort and not call `realloc` or `qsort` could be a good option. That would just make `sash` bigger, but could possibly be done with a few well-placed if statements. You can be the judge of that. If we keep it in sash, that gives the ability to list super-long listings, without having to have yet another listing program.

> Agreed, maybe continue on https://github.com/jbruchon/elks/issues/1390 ?

Ok. Meanwhile I'm going to post some fixups with `split` included, etc for the images.

> I'll test and submit an updated ´df´ - and ´mount´

Sounds good. I'll post an updated `ustatfs` later today.

PR #1428, opened 2022-08-29
[cmds] Update df and mount to use the enhanced ustatfs

Mellvik:
Updated `mount` and `df` to take advantage of the new `UF_NOFREESPACE`option to `ustatfs`:
- `mount` without arguments no longer waits for the calculation of available free space on FAT file systems
- `df` without arguments lists all mounted filesystems including available and used space - like before
- `df` with options `[ikP]` reports no statistics for FAT filesystems 
- `df` man page updated.

### DIFF
--- a/elkscmd/file_utils/df.c
+++ b/elkscmd/file_utils/df.c
@@ -38,6 +38,7 @@ struct dnames {
 };
 struct dnames *devname(char *);
 static char *dev_name(dev_t);
+int df_fat(char *, int);
 
 int iflag= 0;	/* Focus on inodes instead of blocks. */
 int Pflag= 0;	/* Posix standard output. */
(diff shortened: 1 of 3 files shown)


ghaerr:
Thanks for the `df` enhancements and documentation!

toncho11:
df takes 105 seconds to calculate the free space on 60000 hdd (XT-IDE) on a 8 MHz 8086 with 640 KB RAM. On DOS it takes 30 seconds.

Mellvik:
It's the nature of things.
Use (the new) mount without arguments or df -Pk
if you need to list mounted file systems.

If you really need to see free space, df is your only choice.


BTW, it would be interesting to know how much time a regular dir command under DOS on your HD takes, including the listing of free space at the end. What size is the FAT partition we're talking about?

—M

ghaerr:
> df takes 105 seconds to calculate the free space on 60000 hdd (XT-IDE) on a 8 MHz 8086 with 640 KB RAM. On DOS it takes 30 seconds.

Wow! Unfortunately it seems that the ELKS FAT filesystem driver is pretty inefficient, but good to know good ole DOS is terribly slow also.

If the filesystem is FAT32, then the free space may be pre-computed and stored in a special sector. There is no standard for FAT12 of FAT16 filesystems, even with MSDOS.

What is the setting of your EXTBUFS? 64? Increasing that may help, but it depends on how big the FAT table is. The FAT table is buffered in system buffers, but it is required to read the *entire* FAT table from beginning to end to determine the free space. When ELKS mounts a FAT filesystem, the number of sectors of the FAT table is shown by the `fsz=...` display.

> Use (the new) mount without arguments or df -Pk.

As @Mellvik suggests, using `mount` will skip the FAT free space check. Another idea could be to add an option to `df` to not compute free space, but, hey, that's the point of running `df`, especially if you've only got FAT filesystems mounted.

PR #1431, opened 2022-08-31
[cmds] Updated functionality in makeboot

Mellvik:
Updates to `makeboot` to enable singular operations: MBR only, bootblock only. Fixes loading bootblock from file by padding the `minix.bin` file with NULLS to 1024. Also the new bootblock is signature checked (aa55) as an extra guard against botches.

I noticed after pushing the commit that Linux doesn't like the `dd` options I've used (padding) (which works on MacOS), a small update to fix this is forthcoming.

`makeboot` without any of the new options works like before (which is not optimal, but compatible with `sys`, comments below).

New options:
- `-m` Update  MBR only, no bootblock, no system copy
- `-s` Skip system copy, useful to update bootblock only

Opinion: With the new functionality, it would be beneficial to change `makeboot` options to be more logical, for example:
- no options: Copy bootblock
- `-f` Copy bootblock from file
- `-s` Copy system (in addition to bootblock)
- `-M` Write new MBR to the drive, nothing else unless `-s` is also specified, implies bootblock transfer
- `-F` Write bootblock to drive (flat drive), nothing else unless `-s` is also specified

Such change would require only a very minor change to `sys`.

BTW - it may be a good idea to add an updated version of `sys` to this PR in any case, to fix `/dev/ne2k` and add `/dev/3c509` and `/dev/wd8003`.

### DIFF
--- a/bootblocks/Makefile
+++ b/bootblocks/Makefile
@@ -26,7 +26,7 @@ all: $(TARGETS)
 
 minix.bin: boot_sect.o boot_minix.o
 	$(LD) $(LDFLAGS) -M -o minix.bin boot_sect.o boot_minix.o > minix.map
-	dd if=minix.bin obs=1k conv=osync of=minix.tmp; mv minix.tmp minix.bin
+	dd if=minix.bin ibs=1k conv=sync of=minix.tmp; mv minix.tmp minix.bin
 
 probe.bin: boot_sect.o boot_probe.o
 	$(LD) $(LDFLAGS) -M -o probe.bin boot_sect.o boot_probe.o > probe.map
--- a/bootblocks/Makefile
+++ b/bootblocks/Makefile
@@ -26,6 +26,7 @@ all: $(TARGETS)
 
 minix.bin: boot_sect.o boot_minix.o
 	$(LD) $(LDFLAGS) -M -o minix.bin boot_sect.o boot_minix.o > minix.map
+	dd if=minix.bin obs=1k conv=osync of=minix.tmp; mv minix.tmp minix.bin
 
 probe.bin: boot_sect.o boot_probe.o
 	$(LD) $(LDFLAGS) -M -o probe.bin boot_sect.o boot_probe.o > probe.map
--- a/elkscmd/man/man8/makeboot.8
+++ b/elkscmd/man/man8/makeboot.8
@@ -1,0 +1,104 @@
+.TH MAKEBOOT 8
+.SH NAME
+makeboot \- prepare device as a system boot device
+.SH SYNOPSIS
+\fBmakeboot\fP [\fB\-M|F\fP] [\fB\-s\fP] [\fB\-f file\fR] \fIdevice\fR
+.br
+.SH OPTIONS
+.TP 5
+.B \-M
+Write the ELKS MBR (Master Boot Record) to the first sector of the specified device.
+.TP 5
+.B \-F
+Create a 'flat' (unpartitioned) device. 
+.TP 5
+.B \-f file
+Take the bootblock from the specified instead of the current root device.
+.TP 5
+.B \-s
+Copy system files from the current root to the target device.
+.SH EXAMPLES
+.TP 5
+Copy the bootblock from the current root partition to \fB/dev/hda4\fP:
+.sp
+.nf
+# makeboot /dev/hda4
+System on /dev/hda1: Minix (CHS 7818/16/63 at offset 63)
+Target on /dev/hda4: Minix (CHS 7818/16/63 at offset 7560000)
+Bootblock written
+.fi
+.TP 5
+Write the ELKS MBR til the specified device:
+.sp
+.nf
+# makeboot \-M /dev/hda4
+Writing MBR on /dev/hda
+.fi
+.sp
+Notice that while a partition (/dev/hda4) is specified on the command line, 
+.B makeboot
+changes that into a device (/dev/hda) since that's where the MBR goes.
+.TP 5
+Replace the bootblock on \fB/dev/hda4\fP with the one in \fB./minix.bin\fP and copy system files to the target partition.
+.sp
+.nf
+# makeboot \-s \-f ./minix.bin /dev/hda4
+Target on /dev/hda4: Minix (CHS 7818/16/63 at offset 7560000)
+Bootblock written
+Copying /linux to /tmp/mnt/linux
+Copying /bootopts to /tmp/mnt/bootopts
+System copied
+.fi
+.SH DESCRIPTION
+\fBmakeboot\fR prepares a device or partition for booting by installing a 
+bootloader and \- if the \fB-s\fP option is present \- some system files.
+The partition or device must already contain a file system, which may be either FAT (MSDOS) 
+or Minix.
+.PP
+Without options, a bootblock is copied from the current root device to the target. 
+A different bootblock may be specifying via the \fB-f\fP option, see examples above.
+Such transfer requires that the running system and the target has the same file system types.
+.B makeboot
+will complain and exit if this is not the case.
+.PP
+If the 
+.B -s
+option is present, the ELKS kernel (\fI/linux\fR) and the kernel configuration file 
+.I /bootopts
+will be copied to the target.
+.PP
+The 
+.B -M
+option may be used to write a Master Boot Record (MBR) to the target device. The 
+MBR will preserve the partition information currently present in the first sector of the target device.
+This operation is similar to the MSDOS 
+'fdisk /MBR' command.
+Refer to the
+.BR MBR (8)
+manpage for details about the ELKS MBR.
+.PP
+.B makeboot
+also supports 'flat devices' - aka unpartitioned devices. While lloppies are always unpartitioned,
+hard sisks usually have partitions and a Master Boot Record. By writing the 
+.I bootblock 
+to the first sector(s) of the disk, the entire disk becomes a single partition. The
+.I -F
+option does this.
+.PP
+.B makeboot
+is used by the 
+.B sys
+command to build a fully pouplated and bootable ELKS system from floppy or from another disk.
+.PP
+\fBfloppy\fP:
+on the specified devices.
+.SH BUGS
+.B makeboot
+cannot replace the MBR on the current root device.
+.SH "SEE ALSO"
+.BR sys (8),
+.BR mkfs (8),
+.BR minix (8),
+.BR FAT (8),
+.BR MBR (8),
+.BR boot (8).
--- a/elkscmd/man/man8/setup.8
+++ b/elkscmd/man/man8/setup.8
@@ -1,0 +1,43 @@
+.TH SETUP 8
+.SH NAME
+setup \- set up some basic parameters for ELKS runtime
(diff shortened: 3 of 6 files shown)


ghaerr:
Hello @Mellvik,

Thanks for the makeboot enhancements!

Note with regards to binary sizes: I had previously worked hard to keep `sys` at or under 3K bytes (3072), as otherwise it takes another disk block (=2 disk sectors). It's at 3056 now, so that may or may not be possible. Feel free to remove comments if that would help. The `makeboot` changes probably increase its size as well, the only bigger issue is that both of these have to fit on a 360K floppy, where we recently had to add `rmdir`, so things are getting pretty tight. I make it a habit to look at the binary size produce for any new system-required executables and see how close we are to a 1024-byte boundary, as sometimes adjusting a couple lines of code will save a disk block. (I know, sounds a bit ridiculous but I think everything helps when we're talking about small ancient floppy-based systems).

> Opinion: With the new functionality, it would be beneficial to change makeboot options to be more logical

Go ahead with your proposed more logical changes, since makeboot is almost always used within `sys`.

> Such change would require only a very minor change to sys.
> it may be a good idea to add an updated version of sys to this PR in any case, to fix /dev/ne2k and add /dev/3c509 and /dev/wd8003.

Sure, go ahead and make all these changes too.

A couple code review comments to consider on your next push, and we'll commit after the above changes and consideration of the following:
- Lines 75 (and 74): I'd prefer that global vars have their own line, especially when a comment for one variable is meaningless for the new variable (line 75).
- Line 193 and 311: One cannot assume sizeof(int) == 2, that's unportable code and will break when the compiler changes, or this code is used on a host. Use `unsigned short` instead. (Using `int` can also cause sign extension issues on comparison, although harder to have that happen with literal 0xXXXX int compares, but is possible).
- Line 207: Suggest using wording more similar to other new error message on Line 343. I.e. say something like "Warning: no boot signature in MBR"
- Line 337-339: Suggest formatting braces to match style of rest of program, or drop braces and use indentation for style consistency.
- Line 340: "Size not %d on bootblock %s", n - Instead of saying just "size error", give size to inform user of error details.
- Line 425: Use of binary AND "&" incorrect - use &&.

I assume you've tested everything, since I won't likely be. We can commit based on your testing and go from there, thanks!

Mellvik:
Thank you @ghaerr,
Updates coming tomorrow. The option changes mandated a full round of testing, which for the first time (!) included flat drives.
> Note with regards to binary sizes: I had previously worked hard to keep sys at or under 3K bytes (3072), as otherwise it takes another disk block (=2 disk sectors). It's at 3056 now, so that may or may not be possible. Feel free to remove comments if that would help.
> 
Good point, I hadn't thought of that. A test for error return from makeboot (neither minix nor fat found) pushed it above the limit, requiring a slight adjustment of the usage strings. We're ok.
> The makeboot changes probably increase its size as well, the only bigger issue is that both of these have to fit on a 360K floppy, where we recently had to add rmdir, so things are getting pretty tight. I make it a habit to look at the binary size produce for any new system-required executables and see how close we are to a 1024-byte boundary, as sometimes adjusting a couple lines of code will save a disk block. (I know, sounds a bit ridiculous but I think everything helps when we're talking about small ancient floppy-based systems).
> 
Agreed, and again a good point. Makeboot is staying below 8192.

> Line 425: Use of binary AND "&" incorrect - use &&.
Great catch!


-M

ghaerr:
Thanks for the makeboot man page, nice!

Things look good, except the `short * bsect` is still not portable in two places: on machines (all modern hosts) where sizeof(int) != sizeof(short), using `short` instead of `unsigned short` will cause sign extension of the sector contents and thus not fail the compare with the 0xAA55 (which isn't sign extended). For portability when a structure size is fixed, one must always use `unsigned char/short/long` because of C's default sign extension rules.

Did you make all the other fixes you wanted to `sys`?

Mellvik:
Thank you @ghaerr.

> 
> Thanks for the makeboot man page, nice!
> 
> Things look good, except the short * bsect is still not portable in two places: on machines (all modern hosts) where sizeof(int) != sizeof(short), using short instead of unsigned short will cause sign extension of the sector contents and thus not fail the compare with the 0xAA55 (which isn't sign extended). For portability when a structure size is fixed, one must always use unsigned char/short/long because of C's default sign extension rules.
> 
I will fix this, but I don't see that the elements you list are relevant to the case as there is no ints, only shorts involved. So there is no point at which a sign extension can possibly happen on one side and not the other. If the constant (0xaa55) got converted and sign extended so would the (short) variable and they would still be comparable as far as I can see.
In principle though, I agree with you.
> Did you make all the other fixes you wanted to sys?
> 
Not really, but this is what I could fit in (under 3072) without a major 'operation' - such as saving the cwd, cd'ing to the target directories and thus shorten the individual commands. 
sys needs to do more argument checking so as to not pass illegal arguments to makeboot, which will complain and exit fine, but with confusing messages to the user. Low priority.

I have short man pages for sys and setup to be added too. Coming shortly.

—M

ghaerr:
> but I don't see that the elements you list are relevant to the case as there is no ints, only shorts involved

There is no difference between sign extension for `int` vs `short`: both may be sign extended, depending on the compare. Only `unsigned int` and `unsigned short` are not sign extended.

> If the constant (0xaa55) got converted and sign extended so would the (short) variable and they would still be comparable as far as I can see.

This is a different issue: in C, a literal constant is not sign extended, but short is always sign-extended to int. One has to be extremely careful comparing literals, ints and shorts. These are some of the reasons for the 'u', 'l' and other literal integer constant suffixes. Note also that if one component of any compare is unsigned, the other component is converted to unsigned by the compiler.

> In principle though, I agree with you.

"in principle" -> do it now :)

Yeah, I guess I am a bit sensitive about this: remember the big MINIX file system problem you found and we just fixed in #1415? The entire problem, which took me almost two days to debug and uncover, was the result of an array (of indirect block numbers) declared as `int` instead of `unsigned short`. What happened is that when the indirect disk block happened to get to a number that had the high bit set, it got sign extended in a compare when it should not have. I'm guessing that this particular bug may have worked on early Linux where sizeof(int) != sizeof(short), but then that code was taken and used in ELKS, where it remained for 20 years until we found it.

Thanks for understanding, I now demand that all code be portable, even on ELKS, since at some point that code may be reused on a non-compatible system.

ghaerr:
Thanks for the changes @Mellvik. Are you ready for commit now?

Mellvik:
> 2. sep. 2022 kl. 16:18 skrev Gregory Haerr ***@***.***>:
> 
> 
> but I don't see that the elements you list are relevant to the case as there is no ints, only shorts involved
> 
> There is no difference between sign extension for int vs short: both may be sign extended, depending on the compare. Only unsigned int and unsigned short are not sign extended.
> 
> If the constant (0xaa55) got converted and sign extended so would the (short) variable and they would still be comparable as far as I can see.
> 
> This is a different issue: in C, a literal constant is not sign extended, but short is always sign-extended to int. One has to be extremely careful comparing literals, ints and shorts. These are some of the reasons for the 'u', 'l' and other literal integer constant suffixes. Note also that if one component of any compare is unsigned, the other component is converted to unsigned by the compiler.
> 
> In principle though, I agree with you.
> 
> "in principle" -> do it now :)
> 
> Yeah, I guess I am a bit sensitive about this: remember the big MINIX file system problem you found and we just fixed in #1415 <https://github.com/jbruchon/elks/pull/1415>? The entire problem, which took me almost two days to debug and uncover, was the result of an array (of indirect block numbers) declared as int instead of unsigned short. What happened is that when the indirect disk block happened to get to a number that had the high bit set, it got sign extended in a compare when it should not have. I'm guessing that this particular bug may have worked on early Linux where sizeof(int) != sizeof(short), but then that code was taken and used in ELKS, where it remained for 20 years until we found it.
> 
> Thanks for understanding, I now demand that all code be portable, even on ELKS, since at some point that code may be reused on a non-compatible system.
> 
> 

Point taken @ghaerr, thanks for the run-through.

I do suspect we'll find more of the same still. Like: fsck is looping on me on a file system (inode seemingly) error and it may be another one of these. I need to get it reproducible in a controlled way before reporting. 

Apropos fsck: Unsurprisingly but not obvious at first, fsck will fail (fmemalloc) on full size file systems if XMS buffers are maxed out. I didn't dive into this one, but assume it depends on what else is running on the system too.

-M

PS; Yes pls go ahead with the commit.

ghaerr:
> fsck is looping on me on a file system (inode seemingly) error and it may be another one of these. 

You mean busy-looping and not exiting? I had thought we'd got most of fsck working, but I'm not surprised, I suppose. I am starting to depend on it more, to let us know when a filesystem fix is working (like #1415).

> Unsurprisingly but not obvious at first, fsck will fail (fmemalloc) on full size file systems if XMS buffers are maxed out

IIRC, a full-sized filesystem (65M) would need 64K of memory. The memory (using fmemalloc) never uses XMS buffers, so it must be that there isn't a free contiguous 64K block of main memory. You might use `meminfo` to take a look when this next happens.

(On another note, keep in mind that each 8MB of MINIX filesystem also uses a system buffer at mount time (for each ZMAP bitmap), and remains locked the entire time. So if you've got 3-4 large MINIX filesystems mounted, you're definitely using lots of system buffers. These buffers can be XMS buffers though, but in that case they have to be copied into an L1 buffer every time a file is created or deleted if not already present.)

Mellvik:
Thank you @ghaerr,

> 2. sep. 2022 kl. 18:26 skrev Gregory Haerr ***@***.***>:
> 
> 
> fsck is looping on me on a file system (inode seemingly) error and it may be another one of these.
> 
> You mean busy-looping and not exiting? I had thought we'd got most of fsck working, but I'm not surprised, I suppose. I am starting to depend on it more, to let us know when a filesystem fix is working (like #1415 <https://github.com/jbruchon/elks/pull/1415>).
> 
It's looping through the sam repair sequences over and over - zillions of lines and not meaningful, to debug until I have recreated it in a more controlled setting. There may of course be something totally out of whac in the FS, so I'll keep this one low until more details. I mentioned it just in case you have something fsck related on your active list.
> Unsurprisingly but not obvious at first, fsck will fail (fmemalloc) on full size file systems if XMS buffers are maxed out
> 
> IIRC, a full-sized filesystem (65M) would need 64K of memory. The memory (using fmemalloc) never uses XMS buffers, so it must be that there isn't a free contiguous 64K block of main memory. You might use meminfo to take a look when this next happens.
> 
This one is recreateable all the time, I'm assuming it's the buffer headers for the XMS buffers that are eating the rest of the memory. So I just disable the XMS buffers and I'm fine. It's just an observation really (the system have 'only' 610K memory, the rest taken by the DiskOnChip stuff, completely out of sight. A different project.)
> (On another note, keep in mind that each 8MB of MINIX filesystem also uses a system buffer at mount time (for each ZMAP bitmap), and remains locked the entire time. So if you've got 3-4 large MINIX filesystems mounted, you're definitely using lots of system buffers. These buffers can be XMS buffers though, but in that case they have to be copied into an L1 buffer every time a file is created or deleted if not already present.)
> 
> 
Good point. I'll be back on this.



-M

PR #1434, opened 2022-09-02
[cmds] Add sl program (train when mistyping ls)

ghaerr:
Just had to add this fun program which further tests tiny curses library in ELKS TUI. The `sl` program, originally from https://github.com/mtoyoda/sl and using libncurses, only required a single line change to port to ELKS :)

Unfortunately, `sl` is too big for the 1440K disks, so its only included on 2880K+ images.

Movie of `sl`:

https://user-images.githubusercontent.com/11985637/188228003-3b75a9d1-953a-400d-8e19-9e1bce767958.mov


Mellvik:
Fascinating, @ghaerr.
And too tempting an opportunity to pass:
A long time ago sl was an actual unix program (@ ucb of course). Like its twin, it listed the directory, and like its name, it was backwards. Every single filename was backwards, otherwise like ls. 

A true eye opener at first acquaintance : 'Hell, how smart is this system really?' It wasn't. But the sense of humor - and creativity - among the developers was inspiring.

- M

ghaerr:
Interesting @Mellvik, I didn't know that about the original sl. I did run into the modern version when I mistyped ls on macOS, didn't know it, and then saw a train coming across the terminal lol... took a while to figure what was happening!

> But the sense of humor - and creativity - among the developers was inspiring.

Agreed! And I figured ELKS perhaps needed a little more TUI graphics for showing off, just for fun :)

Mellvik:
> Agreed! And I figured ELKS perhaps needed a little more TUI graphics for showing off, just for fun :)
> 

Spot on!
-M

tyama501:
I also saw a lots of SLs when I started using LINUX long time ago :)

PR #1440, opened 2022-10-14
Recreate the logo as vector

eloydegen:
Replace the old PNG. Original font that was
used seems to be Charter Bd BT Bold Italic.


eloydegen:
This pull request was not created as a fix for an issue, I just noticed that the current logo was a bit pixelated. If this PR is not appreciated, feel free to close it without merging.

ghaerr:
Hello @eloydegen,

Thanks for the artwork, that's great!

PR #1444, opened 2022-10-31
[rs] Add serial port driver for PC-98

tyama501:
Hello @ghaerr ,

I tested the serial port driver for PC-98 with miniterm for the following baud rate using PC-9801BX( 5MHz system ).
These worked so I created this PR.

B300 
B600 
B1200 
B2400
B4800
B9600
B19200
B38400 

Additionaly,
B9600, B19200 are tested with serial console using bootopt and worked well.

All of my PC-98s are 5MHz system so I cannot test 8MHz system.

Thank you!


ghaerr:
Hello @tyama501,

Thank you for this contribution, wonderful to have an 8251 driver. And very well done, as usual!

PR #1449, opened 2022-11-18
Elksemu light proofread

ccoffing:
If you are serious in coding-style.txt about converting anything you touch to the Linux coding style, I would be happy to (gently) do so, and rework my proofread.  ELKS seems similar to BSD in the sense that it "owns" its userspace, so that makes sense to me.

### DIFF
--- a/elksemu/Security
+++ b/elksemu/Security
@@ -1,4 +1,4 @@
-Is is possible to install /usr/bin/elksemu as a suid-root executable.
+It is possible to install /usr/bin/elksemu as a suid-root executable.
 This gives two additional facilities when running elks executables.
 
 1) It is now possible to run programs that are execute only, without
--- a/elksemu/binfmt_elks.c
+++ b/elksemu/binfmt_elks.c
@@ -4,7 +4,7 @@
  *  Copyright (C) 1996  Martin von Löwis
  *  original #!-checking implemented by tytso.
  *  ELKS hack added by Chad Page...
- *  Cleaned up some more by Robert de Bath 
+ *  Cleaned up some more by Robert de Bath
  */
 
 #include <linux/module.h>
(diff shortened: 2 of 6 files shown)


ccoffing:
My workflow is to aggressively rebase to make a clean history, and I don't have experience with rebasing github PRs, so if you want you can consider my coding style question instead for possible future work.

ghaerr:
Hello @ccoffing,

Thanks for the code review and misc fixes and cleanups. I'm not able to actually run `elksemu` as it requires Linux and I only have macOS, so I'd like to ask @tkchia, the elksemu maintainer, to approve this.

> If you are serious in coding-style.txt about converting anything you touch to the Linux coding style, I would be happy to (gently) do so, and rework my proofread.

There a a number of files that have been in this project for likely decades, including coding-style.txt, from Linus's original commit. In general, we appreciate code reviews, cleanups involving unnecessary variable elimination and so forth (and while you're at it, removing extra spaces when found in a modified file). As far as changing formatting or extra spaces just on its own, we've mostly stayed away from such, for a number of reasons; since portions of the ELKS project source are decades old, it simplifies kernel diffs performed to maintain extreme stability and regression bug tracking. And many of the ELKS user space command line tools are from very old versions of otherwise later-updated source, also providing easier diffs when tracking bugs or comparing versions. In many cases, ELKS can't use a later version of the same command line tool, since after 20+ years, its code or data segment(s) have grown too large for our 16-bit-only support.

Thank you!

ccoffing:
Thanks for the explanation, @ghaerr .  Considering the long history of some of these files, that makes sense.

tkchia:
Hello @ccoffing,

Thanks for your PR!  The proposed changes look good to me.  `elksemu` 's code formatting was indeed rather inconsistent in places, and could use some work.  If I get around to it I will also try to see if there are any parts of `elksemu` that are deprecated or otherwise need updating.

Thank you!

PR #1451, opened 2022-11-20
[kernel] Add KIOCSOUND ioctl for console terminal

tkchia:
This ioctl allows a program to play sounds of arbitrary pitches on the PC speaker.  E.g.:

```
int
main (void)
{
  ioctl(1, KIOCSOUND, (void *) ((1193181 + 440 / 2) / 440));
  sleep(1);
  ioctl(1, KIOCSOUND, (void *) 0);
  return 0;
}
```

This `ioctl` can be used to implement userland functions akin to the `sound` and `nosound` functions in Turbo C and Watcom C (https://github.com/jbruchon/elks/issues/871).

(To test the sound output under `qemu-system-i386`, one way is to supply an option `-soundhw pcspk`.)

### DIFF
--- a/elks/arch/i86/drivers/char/bell-8254.c
+++ b/elks/arch/i86/drivers/char/bell-8254.c
@@ -1,37 +1,41 @@
 /*
- *	elks/arch/i86/drivers/char/bell.c
+ *	elks/arch/i86/drivers/char/bell-8254.c
  *
  *	Copyright 1999 Greg Haerr <greg@censoft.com>
+ *	Copyright 2022 TK Chia <@tkchia@mastodon.social>
  *
- *	This file rings the PC speaker at a specified frequency.
+ *	This file rings the PC speaker.
  */
 
 #include <linuxmt/config.h>
 #include <arch/ports.h>
 #include <arch/io.h>
+#include <arch/irq.h>
 
 #define BELL_FREQUENCY 800
 #define BELL_PERIOD (1193181/BELL_FREQUENCY)
-#define BELL_PERIOD_L (unsigned char)(BELL_PERIOD & 0xFF)
-#define BELL_PERIOD_H (unsigned char)(BELL_PERIOD / 256)
 
 /*
- * Turn PC speaker on at specified frequency.
+ * Turn PC speaker on at specified frequency (more precisely, wave period).
  */
-static void sound(void)
+void soundp(unsigned period)
 {
+    clr_irq();
     outb(inb(SPEAKER_PORT) | 0x03, SPEAKER_PORT);
     outb(0xB6, TIMER_CMDS_PORT);
-    outb(BELL_PERIOD_L, TIMER2_PORT);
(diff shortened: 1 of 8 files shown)


tkchia:
@ghaerr : do I need to manually disable interrupts inside the `bell-8254.c` routines?  I am not very familiar with the rules for IRQ handling inside ELKS.  Thank you!

ghaerr:
Hello @tkchia,

> I am not very familiar with the rules for IRQ handling inside ELKS.

An applicable rule here is that any port I/O operation that may be interrupted needs to be protected to completion by disabling and reenabling interrupts, to keep the external hardware in a known state. The 8254 timer command port is only written by `enable_timer`, `disable_timer`, and `bell/soundp/nosound`. Since the ELKS kernel is single threaded and not reentrant (with the exception of interrupt handlers and their called routines), the timer command port code won't be interrupted and subsequently reentered, unless the kernel itself calls `bell()` in an interrupt handler. Currently, `bell` is only called by the console driver on receipt of ASCII 7, and not queued for later processing in a (future) UART TX-empty interrupt handler. 

So we should be OK with your changes as-is.

Thank you!

ghaerr (approved):

tkchia:
Hello @ghaerr,

Thanks for the clarification.  Besides the timer ports, the `soundp` and `nosound` routines also touch the keyboard controller port `0x61` (which also happens to control speaker output, because IBM).  This may be a problem if a keyboard IRQ comes along.

It seems I should probably ensure that interrupts are disabled during `soundp` or `nosound`.

Thank you!

ghaerr:
Hello @tkchia,

> This may be a problem if a keyboard IRQ comes along.
> It seems I should probably ensure that interrupts are disabled during soundp or nosound.

Good catch - I hadn't considered an internal BIOS interrupt handler touching the keyboard controller, thank you!

tkchia:
Hello @ghaerr,

Thanks!  I have tweaked the patch to disable interrupts around the port I/O.  If it is fine with you, I will proceed to merge the change once the tests pass.

Thank you!

ccoffing:
@ghaerr I have a follow-up question about disabling the interrupts.  Since the kernel is single threaded and not reentrant, and `nosound` is a single `out`, is the cli/sti actually required in `nosound()`?  There is no possible way (that I can see) of interleaving hardware state, even without the `cli`/`sti` in this particular case.

Minor point, with minor savings (2 bytes, 4 cycles), and readability arguable either way, so I'm mostly sanity checking my understanding.

Thanks.

tkchia:
Hello @ccoffing, hello @ghaerr,

> Since the kernel is single threaded and not reentrant, and `nosound` is a single `out`, is the cli/sti actually required in `nosound()`?

It is actually an `in` followed by an `out`.  The `in` is to read the existing state of the keyboard controller port B (port `0x61`).

Thank you!

ccoffing:
Wow, I completely misread that!  Thanks for clarifying.

PR #1452, opened 2022-11-22
Delete duplicate leap year function

ccoffing:
Keep the one that short-circuits to avoid idiv


ghaerr:
Thank you again for your keen code review :)

PR #1455, opened 2022-11-24
[libc] Working on tests

ccoffing:
- add some structure to libc tests
- add a few more libc tests
- fix warnings in other tests

Mostly just adding some structure to the libc tests, with plans to fill this out more in the future.  The tests themselves are arbitrary, but they have already turned up a few failures.  I'll examine the failures in a later pass.


ccoffing:
This also gives me building blocks to benchmark some libc changes on actual hardware.

ghaerr:
Hello @ccoffing,

This is great, thank you for writing and submitting this. Having an organized set of test functions will go a long way to both test existing libc (and other) functions, as well as provide an easy way for contributors to test new code. I've been wanting to do something like this for awhile, an am very grateful for your work on it!

Definitely like the macros in test/libc/test.h and revised main.c. I'm thinking that perhaps we should consider moving the `TEST` and `TEST2` defines out of misc.c and consider renaming them `EXPECT_EQ` etc or something like that. Newer test macros can be defined similarly, such as `EXPECT_LT` (less than) or `EXPECT_NE` etc for max readability and usefulness.

An example of a great set of test macros is in [Cosmopolitan's test library header](https://github.com/jart/cosmopolitan/blob/master/libc/testlib/testlib.h), check it out for more good ideas.

Thank you!

ccoffing:
Thanks for the link to Cosmopolitan's test header.  I will certainly check that out.  (Oh, _that_ Cosmopolitan.. I saw blinkenlights a while back and that somehow reminded me that ELKS isn't dead.  lol...)

I know the "build vs buy" balance, and briefly looked at some other libc test suites (musl, for one), but they are typically so large, and the ELKS libc is so small and specialized that it didn't seem to make sense.  That said, learning from mature test suite macros, and even picking up test data sets where they are applicable (like the strtol tests), sounds like a good path forward to me.  Plus, writing tests one-by-one almost requires reading the source code, and that usually turns up other nuggets.

ghaerr:
> I saw blinkenlights a while back and that somehow reminded me that ELKS isn't dead.

Blinkenlights is pretty cool. Recently, it's been spun off into a [standalone project called Blink](https://github.com/jart/blink) that doesn't require the more complicated Cosmopolitan headers/system to build. I've been playing with it to boot ELKS, and I've got a PR ready that fixes some real-mode emulation issues and adds a few BIOS calls; it now almost completes an ELKS floppy image kernel boot. I've been thinking of the idea of emulating ELKS system calls (rather than the existing Linux system calls), which would then allow us at ELKS to have a visual debugger!

toncho11:
Should we add the tests to the Github actions and perform automatic testing?

ccoffing:
I have a few known failures to fix, then once tests are passing reliably, I think adding it to the Github actions would be a great idea.

PR #1456, opened 2022-11-25
[cmds] Updates and fixes to split(1) and fsck(8)

Mellvik:
While working on #1367, `split` and `fsck` became instrumental tools, and were pushed in ways they haven't seem before. Thus these updates.

- `split` - added `dd`-style scaling to the byte count argument (k|b|w) and changed the line-count option to linux-style. Updated man-page accordingly.
- `fsck` - primarily fixes some integer->unsigned formatting (printf) problems, improve a few error messages
- `fsck` - added a (hidden) 'A' (for 'all') reply to the interactive mode yes/no question: After having answered 'yes' to 20 or 200 error messages, switching to 'automatic' is a lot faster than killing `fsck` and starting again. Will add this to the ma-page.


ghaerr:
Thanks for these improvements @Mellvik!

PR #1458, opened 2022-11-25
[cmd/sh_util] Fix potential crashes in 'tr'

ccoffing:
set1 and set2 are assumed null terminated but code did not ensure sufficient space nor actually terminate.

Also improved range checks so large ranges like tr -t ' -~' 1-9 do not trash the heap.

### DIFF
--- a/elkscmd/sh_utils/tr.c
+++ b/elkscmd/sh_utils/tr.c
@@ -56,15 +56,16 @@ void usage(char ** argv)
  *
  */
 
-int do_args(int argc, register char ** argv)
+int do_args(int argc, char ** argv)
 {
-	int i = 1, j;
+	int i = 1;
+	size_t j;
 
 	while (i < argc && argv[i][0] == '-') {
 		for (j = 1; j < strlen(argv[i]); j++) {
 			switch (argv[i][j]) {
 				case 'c':
-					complement1 = 1;
+					complement1 = 1; /* TODO implement */
 					break;
 				case 's':
 					squeeze = 1;
@@ -98,6 +99,20 @@ int do_args(int argc, register char ** argv)
 	return i;
 }
 
+void out_of_mem()
+{
+	perror("malloc");
+	exit(1);
+}
+
+void *xrealloc(void *ptr, size_t size)
+{
+	ptr = realloc(ptr, size);
+	if (ptr == NULL)
+		out_of_mem();
+	return ptr;
+}
+
 /*
  *
  * build_string()
@@ -110,14 +125,13 @@ int do_args(int argc, register char ** argv)
  */
 
 #define BSIZE 64
-#define out_of_mem() { perror("malloc"); exit(1); }
 
 char * build_string(char * set_descr, int num)
 {
-	register char * buf = malloc(BSIZE);
+	char * buf = calloc(1, BSIZE);
 	int size = BSIZE;
 	int tail = 0;
-	int i;
+	size_t i;
 	char ch, n1, n2;
 
 	if (buf == NULL) {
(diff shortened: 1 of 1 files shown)


ccoffing:
Just noticed that "tr -s" acts like "tr -d", but it did that before my changes too.  Will look in to that.

ghaerr:
Hello @ccoffing,

Nice work! And you know I now have to ask: how are you finding all these? Do you have some neat tools, or are you just inspecting the heck out of everything :)

> Just noticed that "tr -s" acts like "tr -d", but it did that before my changes too. Will look in to that.

Want me to hold for that, or commit this now?

Thank you!

ccoffing:
You can merge this one, since I feel good about it as a unit.  I'll want to think a bit (and test) the misbehavior and I might not get to that right away.

I first found this by paying attention to gcc warnings.  tr.c had a "maybe uninitialized" warning, so I read it carefully and things seemed odd, and I started seeing contradiction in the code.  So I built natively and ran it through valgrind repeatedly.

ccoffing:
I wish valgrind had existed when I started programming.  One of the best tools in the toolkit.

ghaerr:
> I first found this by paying attention to gcc warnings.

Nice work. I spent a lot of time a few years ago eliminating a huge number of the easy to fix warnings out of the kernel and cmds trees, but left some requiring more work and testing. Thanks for fixing this. The kernel should have absolutely no warnings (I just saw a new one from a commit a few days ago in bell-8254.c which I'll fix).

ghaerr:
> So I built natively and ran it through valgrind repeatedly.

Another item on my long list is introducing a debug malloc which can check allocated and free'd pointers at runtime, and be turned on or off without recompilation. I've got the code finished in v7malloc.c, but not the dynamic activation capability. There are also some problems with the existing (standard) libc realloc - there are cases where it'll grow the heap unnecessary, causing allocation failures due to lack of new space for the size requested. I've got temp fixes into some programs where v7malloc.c is linked in explicitly, which doesn't have this problem. 

It would be nice to have a debug malloc, and other capabilities somewhat like valgrind (and ASAN) into the ELKS system, but real mode can't catch errors like protected mode can.

ccoffing:
I love the zero warnings mindset.  And improving tools and environment to allow introspection and automate bug finding.  Yet balancing that with still operating efficiently in a restricted environment.

I have done some ad-hoc building of ELKS code natively on "real" Linux.  But it would be interesting to support building (purely as a testing option) libc and some of the cmds as 32 bit and run against the Linux kernel, and then automate some tests (e.g., valgrind) against that.  It would require some preprocessor annotations in libc's malloc/free.  (Valgrind doesn't cope with elksemu's manipulation of the LDT; this would be a way to sidestep that.)  It's something I'd like to play with to see if it's feasible.

ghaerr:
Hello @ccoffing,

> it would be interesting to support building (purely as a testing option) libc and some of the cmds as 32 bit

I would be very interested in that as well, not just for testing but also to promote reusable code. I'm big on that - two of my external projects [D-Flat Text UI](https://github.com/ghaerr/dflat) and [FM file manager](https://github.com/ghaerr/fm) support compilation of the same source, and execution on macOS, Linux and ELKS. (Currently, they use separate Makefiles but for ELKS libc and cmds we might be able to get this done using ifdef'd rules in Makefile-rules and other files).

> It would require some preprocessor annotations in libc's malloc/free. 

Lets talk more precisely about what you think might be needed, perhaps in a new issue for tracking. Using external valgrind and perhaps ASAN would go a long way towards modernized testing. There are other libc functions that could use the same style redirection(s) for testing or function substitution.

> Valgrind doesn't cope with elksemu's manipulation of the LDT

There's a number of other emulators out there that might be worth looking into and/or enhancing: on one hand, we have 8086 emulators that also provide BIOS and EGA/MDA (and other hardware) support, and on the other emulators that will emulate 8086 ELKS applications. Examples of the former include [EMU86](https://github.com/mfld-fr/emu86) which will boot and run the ELKS kernel, including disk I/O. Examples of the latter include [86sim](https://github.com/ghaerr/86sim), which will execute MSDOS and some ELKS system calls, and [blink](https://github.com/jart/blink) which provides a great TUI but doesn't (yet) emulate ELKS system calls, and of course `emu86`, which emulates ELKS system calls on Linux (only).

Thank you!

PR #1459, opened 2022-11-26
[libc] Implement the `regparmcall` parameter passing convention for `elks-libc`

tkchia:
I finally got to tweaking routines in ELKS's `libc` so that they can work with the `regparmcall` calling convention.  This calling convention can be enabled by specifying an `-mregparmcall` option to `ia16-elf-gcc`.

Under this calling convention, the first few non-variadic parameters to a function are passed in the registers `%ax`, `%dx`, and/or `%cx`, rather than on the stack, if possible.  Using `regparmcall` may yield smaller and faster programs.

I have done some tests on the new routine implementations, but further tests should be useful.


ghaerr:
Hello @tkchia,

Wow, you've been busy, and this looks great! This could be very exciting should it prove that function call performance can be potentially increased in some application use scenarios.

Thanks for the what looks like tedious work of converting the libc ASM files. As you saw, for speed most of them aren't using a standard function prologue/epilogue. (We would have the same problem with some kernel ASM files).

Are you thinking that it might be worth it to code up an option for building libc and all cmds using the `regparmcall` convention? I'd be happy to do so - perhaps using an approach of a setting and ifdef'd build rules in elkscmd/Make.defs or libc/?, where `regparmcall` would be set, and then tested by running the subsequently produced ELKS distribution. (At this point, probably no need to worry about converting or testing the kernel until it is known how much smaller or faster the yield may be).

Have you tested these changes much? That is, has the system been run and found to be usable when *not* using `regparmcall`, i.e. as it was? I'm fine with committing this now, but the changes to the `exec*` routines are a bit hard to follow only visually, so It'd be nice to know the system works well in the original calling convention.

What is the `0x3c` in the following code in execlx.S, a special arithmetic operation on AX and immediate mode?
```
# if defined L_execle || L_execlpe
 	.byte 0x3c
 .find_envp:
 	lodsw
```
I'm not quite (yet) following the execlx.S code dealing with envp and the associated lodsw/stosw, nor the need for the expanded ASM in exec handling in general.

Also, does the code in execlx.S and execlx.inc have to be nearly duplicated?

Thank you!

tkchia:
Hello @ghaerr,

> Also, does the code in `execlx.S` and `execlx.inc` have to be nearly duplicated?

`execlx.S` is not actually needed &mdash; I forgot to remove it.  Sorry for the confusion.

> Have you tested these changes much? That is, has the system been run and found to be usable when _not_ using `regparmcall`, i.e. as it was?

Yes, I have briefly tested the existing system, and it does work.

> I'm not quite (yet) following the execlx.S code dealing with envp and the associated lodsw/stosw, nor the need for the expanded ASM in exec handling in general.

The assembly code is not strictly needed, for the `regparmcall` calling convention, but it allows us to avoid making another copy of the command line arguments (`argv[]` = `{ arg0, arg1, ...}`) in memory.  For `regparmcall`, `arg0` is passed in a register, while `arg1`, `arg2`, ... are pushed on to the stack, just next to the return address.

I had added some code to my `build-ia16` script to test the `regparmcall` convention, specifically the `regparmcall` implementation of the `execle` function.  The `execl`... code was in fact based on the `_spawnl`... code I wrote earlier for my `libi86` library, which is pretty well-tested.  (In fact I found that a glitch that was affecting `execle` (and `execve`) turned out to be in the `regparmcall` implementation of `memcpy` (in `libc/asm/`).  So at the moment I am actually a bit more worried about the correctness of the routines in `libc/asm/`.)

> What is the `0x3c` in the following code in execlx.S, a special arithmetic operation on AX and immediate mode?

It "eats up" the `lodsw` opcode &mdash; `0xad` (`lodsw`) becomes `0x3c 0xad` which is `cmp $0xad,%al`.  The intent is to handle the (rare?) case where `argv[]` = `{ arg0 }` = `{ NULL }`.  This case must be handled specially since `arg0` is passed in a register.

Thank you!

tkchia:
Hello @ghaerr,

> Are you thinking that it might be worth it to code up an option for building libc and all cmds using the `regparmcall` convention? I'd be happy to do so - perhaps using an approach of a setting and ifdef'd build rules in elkscmd/Make.defs or libc/?

This sounds like an idea. :slightly_smiling_face:  But perhaps we can look into adding some sort of regression testing for the various `libc` routines first, to give us more confidence in `libc` 's correctness.

(I have added a few short tests to my [`build-ia16`](https://github.com/tkchia/build-ia16) scripts, but one problem is that they rely on `elksemu` which only works on Linux/i386 and Linux/x86-64.)

Thank you!

ghaerr:
Hello @tkchia,

Thank you for your explanations, I know understand the exec ASM code intent, and agree it definitely worthwhile not to copy the command line arguments, as some ELKS applications are already getting tight on memory when globbed with \*.

> It "eats up" the lodsw opcode — 0xad (lodsw) becomes 0x3c 0xad which is cmp $0xad,%al.

Nice trick. Perhaps this should be explained, even with just a one-line comment beside the 0x3c for better future code maintenance and understanding.

> So at the moment I am actually a bit more worried about the correctness of the routines in libc/asm/.

Agreed. I am working with a new contributor @ccoffing who has been adding libc routines into an automatable test suite at elkscmd/test/libc. I am hoping we can go more in the direction of the Cosmopolitan project that promotes correctness through a very well designed testing architecture. BTW, the memcmp and strcmp routines are not currently used in libc/asm: I initially wrote naive versions using 8086 string load/compare/move instructions only to find that they are larger than the code produce by gcc for the .c versions (and also possibly slower on 386+ CPUs). The required register saves are the size culprit, I wonder whether an optional inline version of critical routines might be a good addition. I intend putting more work into  obtaining more precise information on performance and size for various speed critical libc routines, based on instrumentation results I've had on the back burner.

Thank you!

ghaerr (approved):

ghaerr:
Hello @tkchia,

```
-include *.d
```
I just noticed this in libc/Makefile.inc: does this by chance solve the longstanding issue of libc not being properly rebuilt when a source file is changed under elks/libc? (The problem has always been that in order for the elkscmd/ applications to use the modified libc, a `cd libc; make clean; make` must be executed or libc won't be properly copied into the special internal gcc library area.)

[EDIT: Also, one last question - I still don't understand why the exec copy of the command line arguments in the __IA16_CALLCVT_CDECL case - that wasn't required before, why is that newly required now? The older exec code for the CDECL calling conversion didn't copy the arguments again, just used the values copied into the `sbrk` area from the C code, right?]

Thank you!

tkchia:
Hello @ghaerr,

> Nice trick. Perhaps this should be explained, even with just a one-line comment beside the 0x3c for better future code maintenance and understanding.

Done. :slightly_smiling_face: 

> I still don't understand why the exec copy of the command line arguments in the __IA16_CALLCVT_CDECL case - that wasn't required before, why is that newly required now?

It is still not required actually.  The assembly code in `execlx.inc` looks for `arg0` in the stack parameters and points `argv` at it, much like the original C code.  (However, since this is the `cdecl` calling convention, we need to push the `argv` pointer itself onto the stack.)

Thank you!

ghaerr:
Hello @tkchia,

I notice since this PR was committed that the ELKS build proceeds a bit differently than prior, possibly including:
- When .config is changed at all, the entire C lib is rebuilt
- The C library is attempted to be reinstalled into gcc internal even when only a kernel file is modified

Many times, it is convenient to have a fast path build that doesn't rebuild or reinstall libc when doing kernel development. I added `make kclean` a while back, for instance, to allow quick builds for faster development iterations, which cleans only the kernel files and no libc or elkscmd/. Given that the C lib build is now twice as long it used to be, might there be a way this older behavior could be put back? (I'm not sure if this has to do with the removal of `include *.d` in libc/Makefile.inc, as questioned in https://github.com/jbruchon/elks/pull/1459#issuecomment-1328296763)?

Thank you!

ghaerr:
Hello @tkchia,

I was thinking of testing the new regparmcall calling sequence by compiling the ELKS applications and seeing what happens.
Can you confirm that the following change in elkscmd/Make.defs would work to select this (after the initial definition):
```
CFLBASE += -mregparmcall -march=any_186
```
I would do this only after the kernel is compiled, after a `make clean` in elkscmd/.

[EDIT: confirmed - the above does work for compiling and linking the majority of programs. `bc` is failing with something to do with yacc, will look into that.]

Are there other options for the `-march=any_186` setting, or will that work for 8086 only as well? (Possibly confusing with `-mtune=i8086`.

Also, this command line option sets ia16-elf-gcc to pick up the properly-compiled routines for libc as well as gcc libraries, correct? That is, only those applications that have their own .s/.S files would need to be modified, the rest should just work.

[EDIT: Running an ELKS image with applications completely built from regparmcall calling convention fails with /bin/init and /bin/sh crashing the system. However, /bin/sash works, along with /bin/ls and others. I am guessing the problem is possibly related to system calls, and the order of parameters for > 3 parms? Can you suggest your thoughts on ways this might be more easily debugged?]

[EDIT2: /bin/init fixed, problem was system calls with no arguments failed. See #1491].

Thank you!

tkchia:
Hello @ghaerr,

> Are there other options for the `-march=any_186` setting, or will that work for 8086 only as well? (Possibly confusing with `-mtune=i8086`.)

If you would like ELKS to run on a good old 8086 or 8088, you probably do _not_ want `-march=any_186` &mdash; this will output commands that only exist on a 80186 or above (such as `leave` and `pushw $`_immed_).

But it is OK to say e.g. `-mtune=any_186`, which will gear the optimization towards 80186 class machines, while outputting only 8086 instructions.  `-march=` gives the minimum required architecture, while `-mtune=` only controls fine-tuning.

Thank you!

PR #1461, opened 2022-11-27
Misc small libc fixes

ccoffing:


ghaerr:
Hello @ccoffing,

Thanks for the continued fixes, looks great!

In ash/jobs.c, how about moving `int pgrp` from line 526 to inside the if test on line 596? That would eliminate another \#if JOBS above for better readability.

Thank you!

ccoffing:
> In ash/jobs.c, how about moving int pgrp from line 526 to inside the if test on line 596? That would eliminate another #if JOBS above for better readability.

Good idea, I like it.  Limiting scope helps readability.

Related aside:  C99 allows further reduction in variable scope, such as declaring a loop counter as part of the `for`.  My predilection is to migrate towards C99 (the most common changes seem to be:  eliminate `register` keyword, use modern function declaration syntax, and make variable declarations more local.)  But in some cases, making variables more local might increase code size, due to extra stack allocation instructions, but a smart compiler with `-Os` should be able to avoid this -- this is something I will watch out for.

Toggling the JOBS flag causes other build failures.  Perhaps these flags have never been toggled for ELKS.  Please hold...

ghaerr:
> C99 allows further reduction in variable scope, such as declaring a loop counter as part of the for.
> eliminate register keyword, use modern function declaration syntax, and make variable declarations more local.

We haven't decided per se that ELKS should or should not use C99 constructs, but we're certainly using `ia16-elf-gcc` special constructs. There are issues with our compiler getting large upstream enhancements merged, due to the amount of work that's already gone into the 8086-specific code that isn't upstream. Thus, I'd say we need to be somewhat careful adopting a particular C standard.

That said, I've been eliminating the register keyword and changing to modern function declaration syntax for some time. My understanding is the former is completely ignored. The latter is particularly important for automatic parameter conversions, and we require forward function decls for almost all of the code. With regards to local variable scope, I'm halfway between Linus' original requirement that all variables should be declared at the top (which I assume allows for much easier inspection of kernel stack requirements, which is a big deal for us), and more local scope declarations (including what I recommended here for jobs.c), and the nifty for-loop variable declaration capability. I would say that since the majority of the kernel is written with all variables declared at function start, it'd be nice to keep it that way, unless that introduces more cruft, specifically with regards to ifdefs. I try real hard to eliminate ifdefs entirely, if possible and feasible (we're a long way from that!).

> in some cases, making variables more local might increase code size, due to extra stack allocation instructions

For every case I've looked at, it seems ia16-elf-gcc aggregates all required stack space into a single subtract SP (or sometimes junk push instruction) in the function prologue, so that shouldn't be an issue.

> Toggling the JOBS flag causes other build failures. Perhaps these flags have never been toggled for ELKS.

The JOBS flag has been default off to keep the memory requirements of /bin/sh small. As such, it would not surprise me that we might see compilation errors and/or execution bugs when turned on. Note also that kernel support for SIGSTOP/SIGTSTP, some of which are required for jobs support are also incomplete (exactly how much I don't remember offhand).

ccoffing:
After toggling various flags in an attempt to test, I think ash needs a dedicated testing/cleanup pass.  Outside the scope of this PR.  Pushed onto my queue unless someone beats me to it.

PR #1463, opened 2022-11-28
[build] Use Ubuntu 20.04 for CI runner

ghaerr:
Attempts at fixing CI build error on @ghaerr's repo, but seemingly not occurring here.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - name: setup-1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - name: setup
@@ -138,3 +138,8 @@ jobs:
           name: hd64-minix.img
           path: image/hd64-minix.img
 
+      - name: upload17
+        uses: actions/upload-artifact@v1
+        with:
+          name: fd1232.img
+          path: image/fd1232.img
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,8 +18,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: setup
-        run: 'sudo apt install bison flex texinfo libncurses5-dev libelf-dev mtools ncompress'
+      - name: setup1
+        run: 'sudo apt-cache policy mtools'
+
+      - name: setup2
+        run: 'sudo apt install bison flex texinfo libncurses5-dev libelf-dev mtools=4.0.24 ncompress'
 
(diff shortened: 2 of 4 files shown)


ghaerr:
Hello @tkchia,

I wonder if you might have any input/help for me on this strange GitHub CI build problem? Unfortunately, I have little experience with the CI system and apparently don't fully understand it.

As discussed in https://github.com/jbruchon/elks/issues/1325#issuecomment-1329492539, strangely enough, the version of `mtools` (specifically `mformat`, used to format a floppy for the PC-98 1232K image build) is using a different version of `mformat` on @ghaerr than it is here on the master repo. Specifically, the GitHub CI workflow in .github/workflow/main.yml which performs `sudo apt install mtools` seems to install mtools v4.0.24 here (which works), but is installing mtools v4.0.32 on @ghaerr. Given that the main.yml is the same, I'm wondering if this might have something to do with CI caching. I've also tried to specify the version number using `sudo apt install mtools=4.0.23` which fails with version not found.

Working ELKS repo (using mformat v4.0.24):
mtools v4 0 24 install
mtools v4 0 24

Failing @ghaerr repo (using mformat v4.0.32):
mtoolls v4 0 32 GH

Of course, another option would be to specify a different image format for the PC-98 build, as it seems that `mformat` got broken at some point. But it'd be nice to figure out the reason why the installation is different on the Github repos with the same contents.

Thank you!

tkchia:
Hello @ghaerr,

One possibility is that the `apt` package information on some of the GitHub CI virtual machines is out of date, and it needs to be refreshed from the Ubuntu mirror server(s).  Maybe try doing a `sudo apt-get update` before doing the installation &mdash; like what you are already doing in `.github/workflows/cross.yml`.

Thank you!

ghaerr:
Hello @tkchia,

Thank you, that helped: I was able to determine the problem is that, for some reason, the @ghaerr build is using Ubuntu Jammy, whereas the ELKS master repo is seemingly assigned to an Ubuntu Focal system (which is older). Jammy only has `mtools` v4.0.32, which fails, whereas Focal's latest `mtools` is v4.0.24, which works.

I tried adding the focal/main Ubuntu repo using `apt-add-repository`, but couldn't get that to work on Jammy. For the time being, I've turned off the PC-98 CI build, until we can figure a way to perhaps assign my CI to a Focal VM.

[EDIT: [here's](https://launchpad.net/ubuntu/+source/mtools) the mtools packages supported for each Ubuntu version. I suppose it may be possible to build a custom PPA for mtools-4.0.24, but haven't looked into how to do that (especially from macOS!)]

Thank you!

ghaerr:
Hello @tkchia,

FYI - the problem was eventually solved by finding [this article](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/) from earlier this month, where it is explained that *some* CI VMs specifying the latest Ubuntu release would be using Ubuntu 22.04 (while others Ubuntu-20.04). Changing the `runs-on: ubuntu-latest` directive in the .yml files to `runs-on: ubuntu-20.04` solves the problem, as the older image uses the non-broken `mformat` v4.0.24.

Thank you!

PR #1464, opened 2022-11-29
[console] Save BP across INT 10h BIOS calls, fix disc probing

ghaerr:
First fix for INT 10h BIOS bug on early IBM PC and other systems, discussed in https://github.com/jbruchon/elks/issues/1327#issuecomment-1330826978.

### DIFF
--- a/elks/arch/i86/boot/crt0.S
+++ b/elks/arch/i86/boot/crt0.S
@@ -60,7 +60,9 @@ early_putchar:
 	mov   2(%bx),%al
 	mov   $0x0E,%ah
 	mov   $0x0007,%bx
+	push  %bp               // some BIOS may destroy BP
 	int   $0x10
+	pop   %bp
 	ret
 #endif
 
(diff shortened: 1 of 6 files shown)


Vutshi:
Thank you @ghaerr, I'll try this version soon.

ghaerr:
Added DEBUG_PROBE (default on) and FORCE_PROBE (default off) to the BIOS disk driver. This will produce much more information during the kernel boot when trying to identify a floppy, based on the discussion in https://github.com/jbruchon/elks/issues/1327#issuecomment-1331325986.

ELKS 360k FAT boot with EBP (and thus no probing):
Screen Shot 2022-11-29 at 1 35 19 PM

ELKS 360k FAT boot with FORCE_PROBE on:
Screen Shot 2022-11-29 at 1 32 49 PM

ghaerr:
Hello @Vutshi, 

I've looked pretty deep into this and am fairly sure that your problem described in https://github.com/jbruchon/elks/issues/1327#issuecomment-1331348693 is that both the boot sector read and subsequent probe reads are failing (likely due to no retries and possibly need for BIOS disk reset).

I've rewritten the routine to attempt a retry, and, upon failure to use the BIOS disk parms rather than blindly assuming 8 sectors. I think this should fix the problem completely in either case.

Thank you!

PR #1466, opened 2022-11-30
Fix build error in ash due to nonstandard index/rindex defines

ccoffing:


ghaerr:
Thanks Chuck, nice use of weak aliases for these cases :)

PR #1468, opened 2022-12-02
Fix tr so that squeeze does not act like delete.

ccoffing:
When there is only one set, the test len1 > len2 will usually be true, so set2 would get changed from NULL to a padded copy of set1, but since set2 was empty anyway it becomes completely '\0' filled.  Since there is now a non-NULL set2, the "translate" clause of the main loop runs, which would translate the read character to '\0', which is also used to mean "do not print".

### DIFF
--- a/elkscmd/sh_utils/tr.c
+++ b/elkscmd/sh_utils/tr.c
@@ -234,7 +234,7 @@ int main(int argc, char ** argv)
 		len2 = 0;
 	}
 	/* printf("{%d,%d}",len1,len2); */
-	if (len1 > len2) {
+	if (set2 && len1 > len2) {
 		if (truncate1) {
 			set1[len2] = '\0';
 		} else {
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @ccoffing,

Nice job identifying this bug. I have to ask - how did you find this?

Can you give an example of a `tr` command line for which this previously failed?

Thank you!

ccoffing:
I noticed this while testing PR #1458 , which in turn was noticed due to "uninitialized" warnings from the compiler.  There are more such warnings in the codebase (which I hope to investigate more) although most are in hand-coded parsers which might just be complex enough to confuse the compiler.

I compared behavior against GNU tr to determine correctness.

A sample bad behavior was:
```
echo teeesting | tr -s e
tsting

```
GNU tr (and this fixed code) instead would say:
```
echo teeesting | tr -s e
testing

```

ghaerr:
Thanks for the explanation @ccoffing. It's kind of fun learning about how various commands work by chasing down compiler warnings. Several years ago I cleaned up the kernel and learned a lot. ELKS commands are the last set of compilations that still produce warnings... perhaps I'll jump in and we can eliminate them all - it would be pretty cool to have the entire system compiled without any warnings!

It's been somewhat tricky to determine what/how to fix certain programs. For instance, I've held back on `ash` and `vi/elvis` since I have found later versions of the same, that in some cases I thought it made sense to maintain better source compatibility with. I have dug up a number of slightly later versions of `ash` which have a few more internal commands implemented, but compiling those from scratch end up producing new bugs, so I've slowly been back porting features that are suitable for ELKS, and its nice to have uncomplicated diffs. So sometimes its a tradeoff. In general though, I'm also a believer that source in general should produce absolutely no warnings, if for the basic reason that it allows new contribution errors to be more easily noticed and handled at the time of initial submission.

I'll take another pass at looking at the elkscmd/ compiler output and try to report on the categories of warnings it might be worth attacking further. And then of course there's the business of actually testing to determine whether the command actually works as intended. I appreciate your work on this!

PR #1469, opened 2022-12-04
[libc] Implement & use GCC-style static constructors/destructors

tkchia:
Older versions of `elks-libc` use `bcc`'s "autorun" facility &mdash; see https://github.com/jbruchon/dev86/blob/master/libc/Pre_main &mdash; when available, to register initialization & termination functions, a.k.a. constructors and destructors, to be invoked before `main()` and after `exit( )` respectively.

These are used to properly initialize `stdio` data structures on program startup, and to properly flush `stdio` streams on exit.

This patch implements GCC-style constructors and destructors via special `.ctors`[`.*`] and `.dtors`[`.*`] sections, as described in https://gcc.gnu.org/onlinedocs/gccint/Initialization.html .

While at it, I also streamlined the implementation of `atexit( )`.


ghaerr:
Hello @tkchia,

Thank you for this contribution. It looks good to me for commit, provided you've double-checked by running various existing ELKS commands that use `atexit`. I have a few minor comments following:

With regards to the new .S files in this PR (and possibly for a couple of older ones recently committed written by you): Do you think we should insert a `.arch i8086,nojumps` before the `.code16` for each of these? I notice several use `jmp` instructions and I think it was you that debugged the problem of ELKS apps running on the PCjr failing when this directive was absent. I figure we should include it within all .S files, for good measure.

I see the do-global-ctor\*.S code possibly being included into all executables now, even if there are no constructors. Is there an easy way to only bring that code in if actually needed (possibly similar to the way `weaken` works on Cosmopolitan, or at least the way `_cleanup` works)? We have been finding that even small bits of code added into every ELKS executable ends up using 3-4 extra disk blocks on 360k floppies. On the other hand, since stdio seems to require a constructor, maybe this isn't a big deal, since many applications use stdio. (There are quite a few tiny ones that have been (re)written to only use `write` for output and error messages, however).

Should we consider using `FAR_ADJ_` in other ASM code to allow operation in both small and medium models, or is this symbol only relevant for internal GCC files?

On another matter, FYI it has been on my list to eliminate the `__AS386_16__` and `__AS386_32__` defines... these were only used by the bcc devkit and not used by your gcc, correct?

Thank you!

tkchia:
Hello @ghaerr,

> I see the `do-global-ctor*.S` code possibly being included into all executables now, even if there are no constructors. Is there an easy way to only bring that code in if actually needed (possibly similar to the way weaken works on Cosmopolitan, or at least the way `_cleanup` works)?

I do not think there is an easy and clean way to do that using the currently existing linker magic.  One problem is that modules just need to dump a pointer into `.ctors` (or `.ctors.*`) to register a constructor, and there is no way (as far as I know) to tell the `gold` linker something like "rope in the symbol `__CTOR_LIST__` (and/or `__do_global_ctors`) if and only if the `.ctors`[`.*`] sections are not empty".

* One hacky work-around I have considered is to modify the GCC back-end so that, whenever it compiles a module containing an `__attribute__ ((constructor))` function, it automatically creates an external reference to `__CTOR_LIST__`.  Since only programs that use constructors will ever actually need `__CTOR_LIST__`, this solution will work.  _But_ it will also mean I am subtly changing the ABI at the `.o` level, and I am not sure this is such a good idea.
* For a clean(er) solution, I am trying to see if it is feasible to do something involving GCC's [`collect2`](https://gcc.gnu.org/onlinedocs/gccint/Collect2.html) linker wrapper.

Thank you!

tkchia:
Hello @ghaerr,

> Should we consider using FAR_ADJ_ in other ASM code to allow operation in both small and medium models, or is this symbol only relevant for internal GCC files?

`FAR_ADJ_` is actually only defined in `elks-libc` 's `<libc-private/call-cvt.h>`.

I have now modified the `#if FAR_ADJ_` directives to instead look for the macro `__IA16_CMODEL_IS_FAR_TEXT`.  This should be a better way, since `__IA16_CMODEL_IS_FAR_TEXT` is set (or unset) by `gcc-ia16`.

Thank you!

ghaerr:
Hello @tkchia,

> whenever it compiles a module containing an __attribute__ ((constructor)) function

Since the programmer has to write `__attribute__ ((constructor))` anyways, how about just requiring a manual static yoink or something to that effect, that manually pulls in the constructor handling code and `atexit`. I am just looking for a method where the constructor code isn't automatically pulled in, but don't see it as a bad thing, at least for now, that the compiler doesn't automatically do this, especially given the extra code being included.

> I do not think there is an easy and clean way to do that using the currently existing linker magic.

Are you saying that we can't use somewhat the same mechanism as Cosmopolitan to implement a static yoink? Such a construct would be very helpful for our floating point routines in printf as well.

Thank you!

ghaerr:
Hello @tkchia,

I have found some time to more thoroughly review both this PR as well as the existing state of affairs within ELKS libc regarding constructors, destructors, `atexit`, `on_exit`, etc from both the bcc as well as gcc perspective. I must say I hadn't been exactly familiar with the existing means of operation, nor the revised enhanced approach using gcc, but I think I'm finally getting the picture. In summary, I would say the existing mechanism(s) for handling this is a bit messed up, and that your proposed gcc method will be a lot nicer, except that it needs some time tuning in order to best work for ELKS.

I have the following thoughts/points as a result of my review, in no particular order, which would be great to discuss, if needed:
- Having a mechanism where a simple attribute declaration added to a function to designate that function to be a constructor (called before `main`), or a destructor (called from `exit`), will be a nice enhancement to ELKS. However, if added, this feature should be considered standard, and not optional; that is, the library may depend on it, and other (e.g. ANSI C or BCC) mechanisms to perform the same should be removed to avoid confusion and cruft accumulation.
- The libc source already contains too many ifdefs. The existing "\_\_AS386_16__", "\_\_AS386_32__", (and "\_\_GNUC__") ifdefs, along with their enclosed code, should be removed, if associated with constructors/destructors.
- `Inline_init` should be removed from _stdio.h, fread.c and fputc.c, as well as the hacked call to `__io_init_vars` in basic/host.c and sash/sash.c. (The latter two are required to work around `fgets` not returning `ferror/errno` properly on SIGINT). Also, stdio/stdio.c should be deleted.
- It appears that the only existing module that actually requires a constructor or destructor is stdio. Because there are a number of ELKS applications that have been specially rewritten to remove any stdio dependencies (for size), we need to solve the problem of not including `atexit` nor the constructor code into executables that don't need them. For this reason, crt0.S should not directly call constructor/destructor code nor `atexit` without some mechanism to avoid pulling that in (more on that below).
- In addition, there is some special code written around the extern `_cleanup` variable used to call `__do_exit` without being linked in, which handles the `atexit` destructors. Perhaps we could use something like this for the constructors as well? 
- Should destructors only be associated with an auto-call of `atexit(__do_global_dtors)`? Or does it make any sense to eliminate destructors and require a manual call to `atexit` in the constructor? This could save space, unless a more general purpose for destructors is needed. If the linker script could set a global with the size of the destructor list, this could be checked along with _cleanup in `exit`.
- Remove include <stdbool.h> and the associated `bool` typedef. The include nor typedef is not used anywhere else within ELKS, and needn't be introduced here, IMHO. (Is bool int or char, by the way?)
- For the sake of small code, consider deleting `on_exit` entirely - it's nonstandard and only serves to add more code to ELKS. We don't need both. My take is that since `L_on_exit` has never been defined, remove it and all associated code. This makes atexit.c quite a bit smaller, and smaller data. The `int rv` parameter to `__do_exit` should be removed.
- Remove the pseudo "ATEXIT.H" extern decls at the top of atexit.c and just use the declared allocated versions from later in the same source.

Here's a screenshot of `ls -lS /bin` showing some of the smaller executables, of course none of which use stdio:
Of course, each 1K boundary ends up using another whole block in MINIX, although not on FAT filesystems.
Screen Shot 2022-12-06 at 4 28 51 PM

Given that the only existing code that uses constructors is stdio, perhaps using the same mechanism as `_cleanup` in libc/system/cleanup.c (used in stdio and atexit) could allow crt0.S to query a `_stdio` function pointer global that would end up calling `__do_ctors` or something like that. Would there be a way for the linker script to set a global variable to the size of the constructor list, to set or test that variable in crt0.S? If that's not easily possible, perhaps having a global within stdio like `vfuncp _stdio = __do_ctors;` versus an uninitialized `vfuncp _stdio;` without stdio. Using this as a temp workaround would allow this PR to be committed without having to figure out a better approach, like using weak/strong symbols, while allowing for the existing stdio init and atexit code to be considerably cleaned up and modernized.

Let me know what you think, thank you!

tkchia:
Hello @ghaerr,

At the moment, I am not terribly worried about the issue of code size.  I did a quick test by compiling a simple
```
int main() {
  printf("hello world\n");
  return 0;
}
```
program for ELKS for the small model, and the size of the code turned out to increase from 1,280 bytes to ... 1,280 bytes.  It turned out that my hacks to streamline the `atexit( )` code &mdash; mainly to remove the unneeded `on_exit( )` provisions &mdash; were balancing out the code size increase from `__do_global_ctors()` and `__do_global_dtors()`.

And any further improvements from here on, I suppose, can only make things better. :slightly_smiling_face:

Anyway, after looking at GCC's `collect2` code a bit, I now think it is feasible, and probably also worthwhile, to take some time to write a linker wrapper that can optimize the calls to these things.  This will probably require bumping up the version of the `gcc-ia16` toolchain used by ELKS, among other things.  Perhaps in the meantime I will also submit a separate PR to do that.

(**Edit:** https://github.com/jbruchon/elks/pull/1472 .)

Thank you!

ghaerr:
Hello @tkchia,

> I am not terribly worried about the issue of code size. I did a quick test by compiling a simple
> program for ELKS for the small model, and the size of the code turned out to increase from 1,280 bytes to ... 1,280 bytes.

That's great. However, my concern about code size is not so much programs that include stdio, but rather those that do not; these are listed in the screenshot above. Specifically, the concern is code size for programs pulling in constructor/destructor code from crt0.S that do not use stdio (or constructor/destructors).

The capability of a link-time tunable libc for ELKS, especially in the stdio area, is important. For instance, the current overheard of stdio is ~1280 bytes, which rounds up to 2K disk used for each stdio program on a MINIX floppy. This caused image overflow for most floppy sizes and required the need to delete desirable programs from the distribution. For any programs that stay running (e.g. /bin/init, ftpd, all other network servers, sash shell), it gets worse, as the stdio buffers permanently use main memory not available to run other programs. Things are quite tight with networking running. To help solve/tune this issue, a special elkscmd/lib/tiny_vprintf.c was created. It uses much less code and data and is linked into these oft-resident programs.

The code size problem has gotten worse with the recent implementation of `dtostr`, `ecvt`, `fcvt` and printf floating point support. For the time being, a `__STDIO_PRINT_FLOATS` static declaration is being used as a hack to drag in floating point support in `basic` without bloating other binaries. I've looked into Cosmopolitan's implementation of `_weaken`, and it seems we should be able to use the `.weak` asm directive and weak symbol support in the linker to effect this in a more stylish way, instead of the current method, which uses a manually declared function pointer similar to libc's `_cleanup`variable. I now see this isn't enough to support the linker magic required for auto constructor/destructors, but wanted to mention some of the current needs for ELKS libc and the smallest code size possible for most programs.

Thank you!

ghaerr:
Hello @tkchia,

> after looking at GCC's collect2 code a bit, I now think it is feasible, and probably also worthwhile, to take some time to write a linker wrapper that can optimize the calls to these things.

Wow, I didn't even know about collect2, thank you for the reference. In [reading about](https://gcc.gnu.org/onlinedocs/gccint/Collect2.html) how collect2 works, where it "is installed as `ld` and ... creates a temporary .c/.o file of the constructors and then passes that to `ld`", it seems there may be a way, if you are thinking of essentially creating an source or object format list of constructors (only when there are any), to allow crt0.S to drag in `__do_global_ctors` (and etc) only when necessary, using the weak symbol mechanism:

The idea would be that the .o file (temporarily) created by your new revised collect2 process could contain, along with the .ctors and .dtors address list(s), a global definition of a `.weak` variable queried in crt0.S. That is, a `.weak` variable declared in crt0.S would be compared to 0, and if non-zero, would be the address of a function that would essentially be `__do_global_ctors`, but only if the collect2 process actually found any constructors and included the `.global` definition of the symbol.

The `__do_global_ctors` code included from the collect2 process would also reference `atexit` etc and would cause that code to be linked in as well. The same approach could be used for destructors (and possibly revising `_cleanup`) in similar fashion.

Thank you!

ghaerr:
Hello @tkchia,

> I do not think there is an easy and clean way to do that using the currently existing linker magic. One problem is that modules just need to dump a pointer into .ctors (or .ctors.*) to register a constructor,

Ah, I see now. I seem to have missed this comment of yours a few days back, and posted a few more times prior to reading it.

> One hacky work-around I have considered is to modify the GCC back-end so that, whenever it compiles a module containing an __attribute__ ((constructor)) function, it automatically creates an external reference to \_\_CTOR_LIST__.

This is an interesting idea - perhaps just emit a reference to a `.global` symbol other than \_\_CTOR_LIST__, so that the the ABI doesn't effectively change, and only one new global symbol is seen as a result (to resolve the .weak equivalent in crt0.S). As mentioned above, this symbol would be the function to be called to call the constructor list of functions.

> For a clean(er) solution, I am trying to see if it is feasible to do something involving GCC's [collect2](https://gcc.gnu.org/onlinedocs/gccint/Collect2.html) linker wrapper.

Also very interesting, but possibly a lot more work; the former idea may be a lot simpler and produce the same results, given how infrequently constructor function attributes seem to be used (i.e. never, currently in ELKS).

Thank you!

ghaerr:
Hello @tkchia,

This is looking really good - I like your great idea of using `.section`s in order to pull in the generic constructor and destructor code right into the `.inithead` and `.finihead` sections in crt0.S, and not pull in `atexit` or anything extra. Its a much better idea than what I suggested using weak symbols and is even smaller since there's no call required to the constructor code or other special symbols. Nice work!

I assume you're working on a revised toolchain that spits out data for the sorted `__CTOR_LIST` and `__DTOR_LIST` arrays. I have a couple questions:
- Will the toolchain always spit out `__CTOR_LIST` and/or `__DTOR_LIST` data, or only when an `__attribute__((constructor()))` is seen? I am wondering how this might affect the ELKS kernel crt0.S, where we probably don't want to have any of the constructor code included in the `.init*` or `.fini*` sections included, right?
- Less important, but the same question for applications: will the `.inithead` etc sections always be included, along with the (now quite small) function pointer calling code, or will that be left completely out for binaries that have no constructors/destructors?
- What does the `.reloc .,R_386_NONE,__CTOR_LIST__` assembler directive do? I couldn't find it in the manual. Is this the directive that ends up pulling in the `do_global_ctor.S` code? If so, then without this directive in the kernel crt0.S, nothing would get pulled in to the `.init*` sections, right?
- Will the stack frame still contain the equivalent of `push %bp; mov %sp,%bp` at `_start` or before the call to `main`? The enhanced stack tracer and instrumentation functions I've been working on in elkscmd/debug kind of like having BP = 0 in the stack chain to know when to stop backtracing, or at least allow user verification of an uncorrupted stack. (The kernel sets BP=0 before application execution). This can get a little more complicated when the compiler doesn't emit a standard function prologue for tiny functions, but `main` usually gets one emitted, and the back tracer will now go back to `_start`. For total correctness checking, a `-no-omit-frame-pointer` can be used. This debug code is also used to find a function start address from a function pointer + offset when no symbol table is present; that also depends on standard function prologues being used, along with possible `push %si; push %di` following, which may not always be a good idea, but it currently works to produce a nice complete backtrace.

Thank you!

tkchia:
Hello @ghaerr,

> However, my concern about code size is not so much programs that include stdio, but rather those that do not; these are listed in the screenshot above.

Regarding code size, after a few minor hacks, I got a non-`stdio` program to work with GCC-style constructors and destructors, with the same program size as it had before.  With the unpatched `libc`:
```
$ cat test-printf.c
#include <stdio.h>

int main()
{
  printf("hello world\n");
  return 0;
}
$ cat test-write.c
#include <unistd.h>

int main()
{
  write(1, "hello world\n", sizeof("hello world\n") - 1);
  return 0;
}
$ ia16-elf-gcc -melks -Os test-printf.c -o test-printf-old
$ ia16-elf-gcc -melks -Os test-write.c -o test-write-old
$ ls -l test-*-old
-rwx------ 1 user user 1280 Dec 17 07:06 test-printf-old
-rwx------ 1 user user  368 Dec 17 07:06 test-write-old
```
With the patched `libc`, the size of the `stdio` program is now shrunk to 1,152 bytes, while the program that only uses `write` takes up just 368 bytes, as before:
```
$ ia16-elf-gcc -melks -Os test-printf.c -o test-printf-new
$ ia16-elf-gcc -melks -Os test-write.c -o test-write-new
$ ls -l test-*-new
-rwx------ 1 user user 1152 Dec 17 07:04 test-printf-new
-rwx------ 1 user user  368 Dec 17 07:04 test-write-new
```
Right now I have still not yet introduced any additional linker script hacks, or linker wrapper code, to eliminate the constructor or destructor call logic when unneeded.  Again, any further improvements on this front can only make the code smaller.

Thank you!

tkchia:
Hello @ghaerr,

> * Will the toolchain always spit out `__CTOR_LIST` and/or `__DTOR_LIST` data, or only when an `__attribute__((constructor()))` is seen? I am wondering how this might affect the ELKS kernel crt0.S, where we probably don't want to have any of the constructor code included in the `.init*` or `.fini*` sections included, right?

Currently, `__attribute__((constructor( )))` functions will only deposit function pointers inside `.ctors.*` sections.  They will neither define a `__CTOR_LIST__` symbol or create a reference to a `__CTOR_LIST__` symbol.

> * Will the stack frame still contain the equivalent of `push %bp; mov %sp,%bp` at _start or before the call to main? The enhanced stack tracer and instrumentation functions I've been working on in elkscmd/debug kind of like having BP = 0 in the stack chain to know when to stop backtracing, or at least allow user verification of an uncorrupted stack.

Hmm &mdash; I did not know this was a sort of wanted feature. :eyes:   A quick fix might be to do a `xor %bp,%bp` just before `_start` calls `main`, e.g.
```
#ifndef __IA16_CALLCVT_REGPARMCALL
	push %di
	push %si
	push %bp
#else
	xchg %ax,%bp	// ax := argc
	mov %si,%dx	// dx := argv
	mov %di,%cx	// cx := envp
#endif
	xor %bp,%bp
	CALL_(main)
```
This means though that stack traces will stop at `main` rather than `_start`.  If you would like stack traces to go all the way up to `_start`, I probably need to try something different.

Thank you!

ghaerr:
Hello @tkchia,

> __attribute__((constructor( ))) functions will only deposit function pointers inside .ctors.* sections. 

I see. So, without the `.reloc .,R_386_NONE,__CTOR_LIST__` in the kernel crt0.S, any defined constructor functions would not actually be executed as such, just their addresses placed in a `.ctors.*` section. That should work fine for the kernel, which has no constructor functions.

> A quick fix might be to do a xor %bp,%bp just before _start calls main, e.g.

The stack trace code doesn't need a full prologue with SI and DI saved or BP=0; just a 0 pushed as the "prior BP" indicating end of stack chain. I am wondering about the changed code in crt0.S, lines 22-30 replaced by lines  29-50 - was that done as a code optimization, or required for the constructor execution? If that startup code were left unchanged, there would not be an issue with stack tracing.

(Note: the stack backtrace code was developed prior to your REGPARMCALL enhancement, so it might not yet correctly work when that calling sequence is used; I will fix that later).

> I got a non-stdio program to work with GCC-style constructors and destructors, with the same program size as it had before.

Thank you for reducing the executable sizes along with this enhancement, great job!

Aside from the BP stack chain issue above, is this ready for commit? Are you thinking of making any other required small changes (in particular `Inline_init`) discussed in https://github.com/jbruchon/elks/pull/1469#issuecomment-1340233240 as part of this PR, or would you like me to do that afterwards?

> Right now I have still not yet introduced any additional linker script hacks, or linker wrapper code, to eliminate the constructor or destructor call logic when unneeded. 

Does this mean there does not need to be a toolchain upgrade?

If we need a new toolchain with this commit, it would be nice to delay just a bit until we finish testing some longstanding work on https://github.com/jbruchon/elks/issues/1327 that we have almost debugged.

Thank you!

tkchia:
Hello @ghaerr,

> I am wondering about the changed code in crt0.S, lines 22-30 replaced by lines 29-50 - was that done as a code optimization, or required for the constructor execution?

I tweaked the code to compute `argc`, `argv`, and `envp`, mainly to ensure that `environ` and `_program_filename` are initialized correctly early on, in case any constructor routines want to make use of these.

Of course, besides setting `environ` and `_program_filename`, we _also_ need to later pass `argc`, `argv`, and `envp` to main.  So I arranged to put them in call-saved registers, so that constructor routines (called from `libc/gcc/do-global-ctors.S`) wll not clobber.  And there are not that many call-saved registers. :neutral_face: 

If getting proper stack traces is a concern, I guess we will need some other method to do both of the above...

> Are you thinking of making any other required small changes (in particular `Inline_init`) discussed in https://github.com/jbruchon/elks/pull/1469#issuecomment-1340233240 as part of this PR, or would you like me to do that afterwards?

I should be able to work on that.

> > Right now I have still not yet introduced any additional linker script hacks, or linker wrapper code, to eliminate the constructor or destructor call logic when unneeded.

> Does this mean there does not need to be a toolchain upgrade?

Yes, no toolchain upgrade should be needed.

Now that I have moved the compiler front-end's spec strings into this ELKS project (https://github.com/jbruchon/elks/pull/1472), hopefully we will now only need to tweak the ELKS end, to alter the link process.

Thank you!

tkchia:
Hello @ghaerr,

> * It appears that the only existing module that actually requires a constructor or destructor is stdio.

Well, my interest in constructors and destructors for ELKS, is partly because I am hoping to port my `libi86` to ELKS (requested by toncho11, https://github.com/tkchia/libi86/issues/21), and some things are a bit easier there if there are constructors and destructors.

E.g. for the MS-DOS target, the `<conio.h>` code uses constructors to make sure there are file descriptors open to the console device, and a destructor to reset it from raw mode to cooked mode if necessary.

Thank you!

ghaerr:
Hello @tkchia,

Thanks for your explanations, I now see why the `_start` code was rewritten. 

> If getting proper stack traces is a concern, I guess we will need some other method to do both of the above...

I think I see a way that we can have `environ` and `_program_filename` initialized before the constructor calls, push a zero BP stack frame in `_start` for stack tracing, pass the arguments appropriately to `main` *and* save some more code as well:

The idea is to use the old startup code (which doesn't use SI or DI, but pushes a zero BP in a standard prologue), and then pre-push AX, BX and CX for the call to `main` later. The constructor routine is then executed (for which its push/pop SI can be eliminated, since SI isn't used anymore), and then, for REGPARMCALL calling convention only, the registers are popped appropriately for the `main` arguments:
```
_start:
// Stack is empty and immediately followed by argc, argv and envp
// original code here
        push %bp        // BP is zero, push stack trace terminator
        mov %sp,%bp     // standard function prologue for debugging
        lea 4(%bp),%bx  // argv [0]
        mov 2(%bp),%cx  // argc
        mov %cx,%ax
        inc %ax
        shl $1,%ax
        add %bx,%ax     // envp [0]
        mov %ax,environ

// new code here
        push %ax        // pre-push arguments to main for cdecl calling convention
        push %bx        // (or saved for regparmcall convention)
        mov (%bx),%bx
        mov %bx,_program_filename
        push %cx

// ...Code fragments from .init & .init.* sections will go here...
// SI and DI are scratch and can be used by do_global_ctors without saving

 	.section .inittail,"ax",@progbits

// pop saved argc, argv and envp for regparmcall calling convention
#ifdef __IA16_CALLCVT_REGPARMCALL
        pop %ax         // ax := argc
        pop %dx         // dx := argv
        pop %cx         // cx := envp
#endif
        CALL_(main)
```
With this code, SI doesn't need to be saved in either do-global-ctors or do-global-dtors, saving even more space.

> my interest in constructors and destructors for ELKS, is partly because I am hoping to port my libi86 to ELKS (requested by toncho11, https://github.com/tkchia/libi86/issues/21)

Nice, does this mean that DOS source should be easy(er) to port? 

> and some things are a bit easier there if there are constructors and destructors.

Overall, the new constructor/destructor mechanism you have designed is quite nice. It should allow for a lot of nice features in applications programs without having to hack special "onetime" functions in the future.

BTW, what program does the sorting for the constructor/destructor priorities? Or is that unimplemented for the time being?

Thank you!

tkchia:
Hello @ghaerr,

> Nice, does this mean that DOS source should be easy(er) to port?

I am working on that (very slowly) on the `libi86` end.

> BTW, what program does the sorting for the constructor/destructor priorities? Or is that unimplemented for the time being?

Constructor and destructor priorities are encoded in the `.ctors`, `.ctors.*`, `.dtors`, and `.dtors.*` section names.  E.g. an `__attribute__((constructor(100)))` function will have its pointer deposited in a `.ctors.65435` section (where 65,435 = 65,535 - 100).  An `__attribute__((constructor))` function simply has its pointer deposited in `.ctors`.

The linker script arranges to sort the input `.ctors`, `.ctors.*`, `.dtors`, and `.dtors.*` sections before laying them out in the output, so that they appear in the correct order.

Thank you!

ghaerr:
Hello @tkchia,

I have looked further into your last change at `_start`:
```
-	push %bp
- 	mov %sp,%bp
- 	lea 4(%bp),%bx  // argv [0]
- 	mov 2(%bp),%cx  // argc
+ 	pop %cx		// arg
+ 	mov %sp,%bx	// argv [0]
+ 	push %bp	// BP is zero, push stack trace terminator
+ 	mov %sp,%bp	// standard function prologue for debugging
``` 
Have you tested this with `ps`? I don't think it will work: `ps` uses the program's original stack pointer (`task->t_begstack`, which points to `argc`, setup by `exec`) to get the argument count and then traverse the `argv[]` array to display the process name and arguments. Thus, argc can't be popped off the stack into CX for this to work. Although your code is smaller, we should probably stay with leaving the initial argc/argv/envp stack untouched, which means probably using the original code at `_start`. I'm sorry, this slipped my mind, I should have caught it earlier and hadn't realized even your original rewrite popped argc into BP, and would have had the same issue.

Thank you!

tkchia:
Hello @ghaerr,

OK &mdash; I have tweaked the code to push back `%cx` (= `argc`):
```
00000000 <_start>:
   0:	59                   	pop    %cx
   1:	89 e3                	mov    %sp,%bx
   3:	51                   	push   %cx
   4:	55                   	push   %bp
   5:	89 e5                	mov    %sp,%bp
...
```
This is slightly shorter than the original code, i.e.
```
00000000 <_start>:
   0:	55                   	push   %bp
   1:	89 e5                	mov    %sp,%bp
   3:	8d 5e 04             	lea    0x4(%bp),%bx
   6:	8b 4e 02             	mov    0x2(%bp),%cx
...
```
(7 bytes vs. 9 bytes) and should correctly preserve the `argc` value on the stack.

Thank you!

ghaerr:
Hello @tkchia,

> This is slightly shorter than the original code
> (7 bytes vs. 9 bytes) and should correctly preserve the argc value on the stack.

Well done @tkchia! I hadn't thought of your quick fix and now the startup code is even smaller than before.

Also, since we are now doing a good job documenting `_start` for the future generations lol, could you please also document lines 51-53 as to the argc/argv/envp variables we're popping for the regparmcall convention (in your next commit):
```
        pop %ax         // ax := argc
        pop %dx         // dx := argv
        pop %cx         // cx := envp
```

Things are looking pretty good, I'm fine with the code thus proposed. After anything else suggested in https://github.com/jbruchon/elks/pull/1469#issuecomment-1340233240, we do need a quick couple of tests for the stdio constructor code (especially when `Inline_init` is removed from `sash` and `basic`): these can be tested by typing ^C at the `sash` or `basic` prompt, which should just result in a newline displayed for the former and no action for the latter, except that input is discarded, as well as testing `ps`. If you'd like to wait on cleaning up \_\_GNUC\_\_ and \_\_AS386_16\_\_ etc, that's fine, I can do that in a later commit, your choice.

Thank you!

tkchia:
Hello @ghaerr,

> If you'd like to wait on cleaning up `__GNUC__` and `__AS386_16__` etc, that's fine, I can do that in a later commit, your choice.

Thank you.  On a second look, it seems that the changes I made to `stdio/__io_init.c` and `stdio/_stdio.h` are not (yet) enough to bring in the `__io_init_vars()` constructor when needed.  I will need to try to fix that.

More precisely, I probably need to somehow couple it with (say) `stdio/fputc.c` and/or `stdio/fread.c`, so that the module containing the `__io_init_vars()` constructor routine is "yoinked" in whenever one uses `stdio` facilities.

I should also take a look at the other uses of the `__AS386_`...`__` macros.

Thank you!

ghaerr:
Hello @tkchia,

> More precisely, I probably need to somehow couple it with (say) stdio/fputc.c and/or stdio/fread.c, so that the module containing the __io_init_vars() constructor routine is "yoinked" in whenever one uses stdio facilities.

I see, so we're back to somewhat the original problem of how to bring in the constructor, but only when needed. It appears the infamous `Inline_init` lives on, but needs to be renamed and converted to a STATIC_YOINK mechanism in the same files, that we don't yet have.

I was thinking that something like the following could be used to solve this, as well as establish a `_weaken` mechanism that would be useful for other code to determine whether the same symbol was linked in or not (ungracefully lifted from Cosmopolitan and hacked):
```
#define _weakaddr(symbolstr)                                             \
  ({                                                                     \
    intptr_t waddr;                                                      \
    asm(".weak\t" symbolstr "\n\t" mov $symbolstr,%ax\n"); \
    waddr;                                                               \
  })
#define _weaken(symbol)       ((const typeof(&(symbol)))_weakaddr(#symbol))
#define STATIC_YOINK(sym)     asm(".section yoink\n\tmov $sym,%ax\n.previous\n")

// in fread.c and fputc.c, as well as sash and basic:
STATIC_YOINK("__io_init_vars");    // only for example, perhaps use STATIC_YOINK("stdio_init")

// for floating point print in stdio/vprintf.c
if (_weaken(dotostr)) _weaken(dotostr)(va_arg(ap, double), *fmt, preci, ptmp);

// in application desiring floating point printing (basic), instead of using __STDIO_PRINT_FLOATS:
STATIC_YOINK("dotostr);
...
```
In Cosmopolitan, the `yoink` section is discarded by the linker script, but still seems to force linking in the symbol if found in the library.

I realize the above `asm` macros need more general treatment and should not just clobber AX without telling the compiler, I'm not expert on the voodoo required for that.

Thank you!

tkchia:
Hello @ghaerr,

I have (hopefully) finally fixed the `Inline_init` issue.

For the linking in of floating-point support, I have a separate patch for that (https://github.com/tkchia/elks/commit/eb18d3778e86a576276d767685b9c76b5847a827) in the works.  It requires a bug fix on the `binutils-ia16` end though (https://github.com/tkchia/binutils-ia16/commit/cc5f1a9d7437051c3aae6ec16fca27c3b2d31f6c), and in any case it is probably good to have it pushed separately.

Thank you!

tkchia:
Hello @ghaerr,

> I realize the above `asm` macros need more general treatment and should not just clobber AX without telling the compiler, I'm not expert on the voodoo required for that.

The `mov` ... `, %ax` will go into the discarded section and will not appear in the final output.  In fact it can be any instruction or pseudo-op whatsoever, as long as it results in a reference to the symbol in question.

Thank you!

tkchia:
Hello @ghaerr,

> we do need a quick couple of tests for the stdio constructor code (especially when `Inline_init` is removed from `sash` and `basic`): these can be tested by typing ^C at the `sash` or `basic` prompt, which should just result in a newline displayed for the former and no action for the latter, except that input is discarded, as well as testing `ps`.

Done, thanks!  If it is fine with you I will merge this after the tests pass.

ghaerr:
Hello @tkchia,

> a quick couple of tests for the stdio constructor code (especially when Inline_init is removed from sash and basic)

Given that `sash` and `basic` call `__io_init_vars()` themselves, and they also use stdio, doesn't this seem like that will result in the execution of `__io_init_vars()` twice, once by the constructor call in `_start`, and then in the replaced `fread`? 

[EDIT: actually `__io_init_vars()` will get executed once per call to `fread`, since the `if(once)` code was removed, which is unacceptable. Let's change this now to `__YOINK` instead outside the function in both sash and basic. Thanks!]

It seems that the explicit kluge calls to `__io_init_vars()` should be replaced with a `__YOINK`. I realize though that that macro isn't yet available outside of private `_stdio.h`, and the execution of the constructor twice won't hurt anything (for now). Would you like to delay fixing this until we figure an appropriate place to put `yoink.h`?

[EDIT: I notice that asm/yoink.h is included in stdio.h in your weak-symbol PR, so perhaps asm/yoink.h should be included in stdio.h now and basic/sash's use of `__io_init_vars` replaced now as a matter of correctness for stdio operation, especially if the upcoming PR gets delayed.]

Thank you!

ghaerr:
Hello @tkchia,

> The mov ... , %ax will go into the discarded section and will not appear in the final output. In fact it can be any instruction or pseudo-op whatsoever, as long as it results in a reference to the symbol in question.

Yes, I realize that's the case for `__YOINK` but was referring more to `_weaken()`, which returns the address of the passed argument (as the same type as the argument), or zero if not linked in. A C compound literal is used to make that work, and I was assuming the need to force GCC to move the value to AX as its return value without actually coding %ax.


> For the linking in of floating-point support, I have a separate patch for that (https://github.com/tkchia/elks/commit/eb18d3778e86a576276d767685b9c76b5847a827) in the works.

Can't we use a macro like `_weaken` instead of having to declare the possibly-not-linked function as weak? The method proposed requires that, whereas I'd like to have the ability to write library or application code that does not require the function itself to be declared weak (see below for `dtostr`). Then any function could be optionally included by using `_weaken` around the function name, with `__YOINK` used to require its linking (otherwise it would be left out).

I'd prefer for the time being we keep `dtostr` as such without leading underscores. It is used explicitly for testing in some programs (e.g. misc_utils/float.c) for additional testing. When porting that function, I found there are several variations of it in use in embedded systems (although not a C standard), and wanted to have a user-usable function without double underscores. (Yes, we could keep it as `__dtostr` and use a weak symbol alias to `dtostr` or vice versa, if you prefer, much like what was done with `rindex` in the C library libc/string/strrchr.c:
```
char *rindex(const char *s, int c) __attribute__ ((weak, alias ("strrchr")));
```

> It requires a bug fix on the binutils-ia16 end though (https://github.com/tkchia/binutils-ia16/commit/cc5f1a9d7437051c3aae6ec16fca27c3b2d31f6c), and in any case it is probably good to have it pushed separately.

That's fine to push separately, this PR has become complicated enough :)

Pending the handling of basic and sash's `__io_init_vars()` discussed above in another PR soon after, I'm OK with committing this PR now.

Out of curiosity, in what circumstances is your toolchain fix necessary? I thought I had tested weak symbols separately (using data, not code, i.e: `int foo __attribute((weak)) = 0;`, then replaced by another `int foo = 3;` in another file and it seemed to work and linked in the second definition.

Also, if we're going to require a toolchain upgrade, are there other features that would be nice to be included, so we don't require developers to recompile the toolchain twice? I have (not a great, but working) patch for instrumentation using the `-finstrument-functions` used in elkscmd/debug that is used for runtime stack backtracing similar to Cosmopolitan's `program --ftrace` which would be nice to be included. I have had to remove that directory from the standard build process pending an `ia16-elf-gcc` upstream patch. This was initially discussed in https://github.com/tkchia/gcc-ia16/issues/111 back in June).

Thank you!

ghaerr:
Hello @tkchia,

I compared the contents of the applications on a 1440k floppy distribution before and after this PR. The result is that, with this merged, there is a net savings of 10k bytes on the floppy!!! Thought you would like to know :)

In general, the size of application binaries that do not use `stdio` increased on average 16 bytes, while those that use `stdio` decreased 144 bytes. The increase seems to be the net overhead of the entire constructor/destructor facility including `_startup` code changes. The decrease is probably the fact that `atexit` isn't included in `stdio` programs unless explicitly called.

I'm preparing a PR that fixes the multiple calls to `__io_init_vars()` described above by basic and sash, with a few other compilation warning cleanups, to be posted shortly.

Thank you!

ghaerr:
Hello @tkchia,

> For the linking in of floating-point support, I have a separate patch for that (https://github.com/tkchia/elks/commit/eb18d3778e86a576276d767685b9c76b5847a827) in the works. It requires a bug fix on the binutils-ia16 end though (https://github.com/tkchia/binutils-ia16/commit/cc5f1a9d7437051c3aae6ec16fca27c3b2d31f6c), and in any case it is probably good to have it pushed separately.

I pulled down the latest binutils and gcc, and can't seem to get weak/strong symbols to work, even with your linker patch. It seems that once a `.weak sym` is seen by `as`, the symbol is never pulled in. I notice that the patch uses a case for R_386_SEG16, but the relocatable information generated by the compiler (using a `__YOINK` macro as follows), always generates R_386_32 and R_386_SUB32 relocations, never R_386_SEG16. Am I doing something wrong?

`__YOINK`:
```
# define __YOINK(sym)   __asm(".pushsection .discard; " \
                  ".long " __YOINK_STR(#sym) "; " \
                  ".popsection\n")
# define __YOINK_STR(sym) #sym
```
Generated output of `__YOINK(atoi)` (no R_386_SEG16):
```
00000000 <.discard>:
   0:	00 00                	add    %al,(%bx,%si)
			0: R_386_32	atoi
			0: R_386_SUB32	atoi!
	...
```
I'm using the following code:
```
#define _ezlea(symbol) "mov\t$" symbol ",%k"
#define _weakaddr(symbolstr)                                             \
  ({                                                                     \
    intptr_t waddr;                                                      \
    asm(".weak\t" symbolstr "\n\t" _ezlea(symbolstr) "0" : "=r"(waddr)); \
    waddr;                                                               \
  })
#define _strongaddr(symbolstr)                \
  ({                                          \
    intptr_t waddr;                           \
    asm(_ezlea(symbolstr) "0" : "=r"(waddr)); \
    waddr;                                    \
  })
#define _weaken(symbol)       ((const typeof(&(symbol)))_weakaddr(#symbol))
#define _strong(symbol)       ((const typeof(&(symbol)))_strongaddr(#symbol))

extern int atoi(char *) __attribute__((weak));

int main(int argc, char **argv)
{
    if (_strong(atoi)) printf("foo\n");
    if (_weaken(atoi)) printf("atoi yes\n");
        else printf("atoi no\n");
    exit(1);
}
```
Generated code (looks correct):
```
00000000 <main>:
   0:	55                   	push   %bp
   1:	89 e5                	mov    %sp,%bp
   3:	b8 00 00             	mov    $0x0,%ax
			4: R_386_16	atoi
			4: R_386_SUB16	atoi!
   6:	85 c0                	test   %ax,%ax
   8:	74 0a                	je     14 <main+0x14>
   a:	b8 00 00             	mov    $0x0,%ax
			b: R_386_16	.rodata.str1.1
			b: R_386_SUB16	.rodata.str1.1!
   d:	50                   	push   %ax
   e:	e8 fe ff             	call   f <main+0xf>
			f: R_386_PC16	puts
  11:	83 c4 02             	add    $0x2,%sp
  14:	b8 00 00             	mov    $0x0,%ax
			15: R_386_16	atoi
			15: R_386_SUB16	atoi!
  17:	50                   	push   %ax
  18:	b8 04 00             	mov    $0x4,%ax
			19: R_386_16	.rodata.str1.1
			19: R_386_SUB16	.rodata.str1.1!
  1b:	50                   	push   %ax
  1c:	e8 fe ff             	call   1d <main+0x1d>
			1d: R_386_PC16	printf
  1f:	b8 01 00             	mov    $0x1,%ax
  22:	50                   	push   %ax
  23:	e8 fe ff             	call   24 <main+0x24>
			24: R_386_PC16	exit
```
If I remove the `__attribute__((weak))` *and* the `if (weaken(atoi)) ...` then `atoi` will get pulled in, otherwise not.

Is there something else I need to do in order to have this method work? It seems equivalent to the `__dtostr` example you wrote (even though ultimately I'd like the method to work without having to (re)declare the weak/strong function as `__attribute__((weak))` but I can't even get that working).

Thank you!

ghaerr:
Hello @tkchia,

I have made more progress on understanding the issue of weak symbol linking and the desired use of \_\_STDIO_PRINT_FLOATS to be replaced with \_\_YOINK as described above:

I can get weak symbols to work the way we want in the C library, only if the source file that uses the `_weaken` macro is *not* in the same file as that using the `__YOINK` of the same symbol. The symbol does not need to be extern with the `__attribute__((weak))` specification; the use of `_weaken`, which emits a `.weak symbol_name`, is enough. Moreover, this mechanism does not require your updated version of the `ld.gold` linker.

It seems, when mixing of weak / normal symbols are attempted in the same source file, the resulting .o is processed by the linker as though the symbol is always weak, even if a `__YOINK` was included. Perhaps this is because `__YOINK` itself only references a symbol in `.discard` section, but does not set any bits to indicate that symbol should be normal/strong, overriding a `_weaken/.weak` use. I tried adding `.extern` to the `__YOINK` definition but that didn't change the results. I don't know of another ASM directive to try, as `.global` can't be used in this case.

So, it appears that toolchain development is not required to proceed, and for my intended C library usage`__YOINK` won't be required in the same file as the `_weaken` test. I perhaps unwittingly was just testing it that way. Would you mind if I put together a demonstration of this as a PR for the ELKS C library to replace `__STDIO_PRINT_FLOATS`, or are you working on a different approach you'd rather see used?

Thank you!

PR #1471, opened 2022-12-05
Fix bug in msdos_readdir: only call filldir() when success

ccoffing:
Fixes this clang --analyze error:
/dir.c:250:3: warning: 3rd function call argument is an uninitialized value [core.CallAndMessage]
                filldir(dirbuf, name, namelen, dirpos, ino);

Probably not much of a change in practice: error path would have only copied length-limited garbage data into userspace before returning the error.


ghaerr:
Thanks Chuck!

Issue #1270, opened 2022-04-28
Build issue under ubuntu 22.04


cocus:
Hi,
I've set up a docker container with ubuntu 22.04 on it, and I can't build the crosscompiler toolchain (when running `./build.sh`).
This is most likely due to the gcc being a really recent one `gcc version 11.2.0 (Ubuntu 11.2.0-19ubuntu1)`.
The error I'm getting is:
```
/bin/bash ./libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I../../binutils-src/bfd  -DBINDIR='"/workspace/cross/bin"' -I. -I../../binutils-src/bfd -I../../binutils-src/bfd/../include  -DHAVE_i386_elf32_vec -DHAVE_i386_elks_vec -DHAVE_i386_msdos_vec -DHAVE_i386_aout_vec -DHAVE_elf32_le_vec -DHAVE_elf32_be_vec -DHAVE_plugin_vec   -W -Wall -Wstrict-prototypes -Wmissing-prototypes -Wshadow -Wstack-usage=262144 -Werror -I../../binutils-src/bfd/../zlib -g -O2 -MT plugin.lo -MD -MP -MF .deps/plugin.Tpo -c -o plugin.lo ../../binutils-src/bfd/plugin.c
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../../binutils-src/bfd -DBINDIR=\"/workspace/cross/bin\" -I. -I../../binutils-src/bfd -I../../binutils-src/bfd/../include -DHAVE_i386_elf32_vec -DHAVE_i386_elks_vec -DHAVE_i386_msdos_vec -DHAVE_i386_aout_vec -DHAVE_elf32_le_vec -DHAVE_elf32_be_vec -DHAVE_plugin_vec -W -Wall -Wstrict-prototypes -Wmissing-prototypes -Wshadow -Wstack-usage=262144 -Werror -I../../binutils-src/bfd/../zlib -g -O2 -MT plugin.lo -MD -MP -MF .deps/plugin.Tpo -c ../../binutils-src/bfd/plugin.c -o plugin.o
In function 'load_plugin',
    inlined from 'bfd_plugin_object_p' at ../../binutils-src/bfd/plugin.c:424:53:
../../binutils-src/bfd/plugin.c:403:20: error: 'valid_plugin' may be used uninitialized [-Werror=maybe-uninitialized]
  403 |         has_plugin = valid_plugin;
[...]

PR #1472, opened 2022-12-07
[libc] [tools] Add and install ELKS-specific spec file for gcc-ia16; bump up gcc-ia16 version

tkchia:
A spec file (https://gcc.gnu.org/onlinedocs/gcc/Spec-Files.html) contains spec strings, which say how to transform GCC command line options into program invocations for the actual compiler passes.

The spec strings for the ELKS platform used to be maintained under the `gcc-ia16` source tree, but I am now moving them over here into the ELKS tree.  This should make it easier to tweak ELKS-specific settings for compilation, assembly, & linking passes, without having to rebuild the whole `gcc-ia16` compiler.


tkchia:
I am also bumping up the `gcc-ia16` version to a newer one, which fixes an issue in the compilation of `__attribute__ ((interrupt))` functions (https://github.com/tkchia/gcc-ia16/issues/119).

Thank you!

ghaerr:
Thank you @tkchia, looks good.

For other ELKS developers: although @tkchia is now updating the `gcc-ia16-elf` compiler to a later version, there is not (yet) a need to rebuild the full toolchain to build ELKS. As @tkchia proceeds to make changes and enhancements to the toolchain, there may be further version updates, some of which may require a toolchain rebuild. We will keep you posted as to when this is necessary. On the other hand, for those that have time, now may be a good time to rebuild the toolchain from scratch (the old compiler can be temporarily kept by using `mv cross cross-org; tools/build.sh`). 

For Ubuntu 22.04 users, there is [open toolchain build issue](https://github.com/jbruchon/elks/issues/1270) that would be nice to test against @tkchia's latest version, so that issue can be closed.

ghaerr (approved):

tkchia:
@ghaerr : thank you!

PR #1476, opened 2022-12-10
Manpages: Ethernet interfaces

Mellvik:
This is the manage for the wd ethernet interfaces.

The 3c509 man page is forthcoming. A merge can wait until completed. 
There has to be typos and omissions in these man pages, we'll take them in stride.

-M

### DIFF
--- a/elkscmd/man/man4/3c.4
+++ b/elkscmd/man/man4/3c.4
@@ -1,0 +1,197 @@
+.TH 3C 4
+.SH NAME
+3c0 \- Driver for the 3COM 3C509 ISA Ethernet controller
+.SH SYNOPSIS
+.nf
+3c509
+	/bootopts: 3c0=12,0x320,,0x80
+	/dev/3c0 - major: 9 minor: 2
+.fi
+.SH DESCRIPTION
+The \fB3c0\fP 
+device refers to the ELKS driver for the 3Com 3C509 10Mbps 
+Ethernet interface (aka NIC) running
+on the PC ISA 8 or 16 bit bus. The 
+\fB3C509\fP
+is a 16 bit interface but may also be used in systems with 
+the original PC/XT 8 bit ISA bus. However, this functionality is currently not implemented in 
+the ELKS driver.
+.PP
+The 
+.B 3C509
+is a plug-and-play 
+.I (PnP)
+interface which means that the configuration specified at boot time does not have to match the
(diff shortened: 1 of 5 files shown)


Mellvik:
The latest commit also includes an update to the `man` command to enable man page file/command names to start with a number.

ghaerr:
Looks good, let me know when you're ready for commit.

Mellvik:
Thank you @ghaerr.We're good to go.While we're on man pages, I keep wondering if it's time to get rid of all those pages inherited from full Linux and not relevant to ELKS. As is I suspect they're more confusing than helpful. What's your take?-M12. des. 2022 kl. 16:36 skrev Gregory Haerr ***@***.***>:﻿
Looks good, let me know when you're ready for commit.

ghaerr:
>  if it's time to get rid of all those pages inherited from full Linux and not relevant to ELKS.

The pages are actually from MINIX 3. You're welcome to delete any you think we'll never use - I already took a first pass, but left those that we might use in the future and possibly use some text from. If you like, we can discuss which ones you think we should delete before you submit the PR.

Thank you!

Mellvik:
Admittedly it's a significant undertaking and there are some consideration to make.

For example, the termcap (5) man page is OK content wise but uses a number of commands and register manipulations not supported by ELKS man, and unlikely to be supported in the future. Such pages may be
- dropped, deleted (simple, the content exists elsewhere)
- simplified so that it may be displayed on ELKS (a significant undertaking)
- the ELKS man command may be expanded in such a way that unrecognised commands make it fork off a port of (minix) nroff with a reasonable macro library available which will be slow but doable.

What seems most reasonable to me is the first, just delete, add later if important and then limit to what's important for/supported by ELKS.

As to the cleanup suggested earlier, I think it makes sense to move the Minix manpages to a separate directory structure not being part of an ELKS build, but still available as references for ELKS-specific man pages - like now.


—M

ghaerr:
Hello @Mellvik,

Thanks for the ideas and suggestions for dealing with the set of incorrect, irrelevant, not-yet-implemented, and not-displayed-properly man pages.

I like your idea of moving any of the above-described problem man pages to a separate `elkscmd/man-minix` directory tree such that they aren't actually installed into an ELKS image. They could still be reviewed using the portable `eman` command on macOS or Linux, and then moved back into the production `elkscmd/man/*` directory when updated.

Thank you!

PR #1477, opened 2022-12-10
[nano-X] Clearing text and graphic vram for PC-98 before open

tyama501:
Hello @ghaerr 

As we were talking,
I added codes to clear text and graphic vram when opening the nano-X driver.
I also modified some orders of the code.

This makes not to see graphics of the previous application.

Thank you.


ghaerr:
Thank you @tyama501!

PR #1478, opened 2022-12-12
Improve test framework currently used by libc tests

ccoffing:
- Test macros distilled from Cosmopolitan libc's testlib
- Test runner does some environment setup, teardown, checking
- testlib.[ch] are generic and could be re-used


ccoffing:
This is (working towards) a checkpoint of testing improvements to libc.  I started with the Cosmopolitan assertions, then distilled and iterated, and now feel like I have something a bit more focused.  Still, this is just an iterative step.  Here are some areas I could use some feedback, and some future directions I'm thinking.

- How to organize tests.  I started by creating a file per libc directory.  These files might get large, maybe that's okay.
- Need to convert some tests to be data-driven to reduce code size (the repeated assertion macros cause some bloat; looping over structs scales better).  And/or break into multiple test executables.
- I haven't thought about memory models, to make this more generally useful.  But testlib.[ch] only depend on libc so should be re-usable.
- Tests should auto-register with the test runner, maybe through linker magic.  I haven't yet looked into how to do this; suggestions welcome.  Maybe test registration should include "suite" and "name", and be filterable from the runner, or maybe that's too much bloat.
- More coverage.  I would also like to touch on the syscalls that libc only wraps.
- The test library needs to support creating and cleaning up temporary directories, for tests that want to do IO.
- Not all assertion variations are fully implemented, and some some redundancy can still be reduced.  This will get fully fleshed out as more tests are written.
- Some tests crash or hang elksemu.  Haven't had time to compare to real hardware.  Maybe test bug, maybe elksemu bug, maybe ELKS bug.  Thrill of the chase.
- Floats don't actually print for failed test cases despite __STDIO_PRINT_FLOATS.  Looks like there are multiple build flags to control this...
- I have plans to extend this with some benchmarking support, because I want to be able to easiliy benchmark potential optimizations on real hardware, and maybe it would be cool to be able to track some areas for performance over time to watch for regressions.

toncho11:
So all tests are complied in an executable called "test_libc". And the output of the tests is printed stored where? A file or console?
The "test_libc" is automatically generated only on the HDD images of ELKS? Otherwise there will be no space, right?

ghaerr:
Wow @ccoffing, this is some amazing work! I've only had time for a very quick look this morning. Your tests and test macros look great. I'll have more time this afternoon to go over it and help answer some of your questions.

Thank you!

ghaerr:
Hello @ccoffing,

> How to organize tests. I started by creating a file per libc directory. These files might get large, maybe that's okay.

That's fine. It would seem that separating tests inside a libc directory may only make sense or be desirable if the tested function(s) are quite different from each other, for instance libc/misc.


> Need to convert some tests to be data-driven to reduce code size (the repeated assertion macros cause some bloat; looping over structs scales better). And/or break into multiple test executables.

Given that we probably want to compile and link the tests in small model, at least in the beginning, I think planning for multiple test executables from the start might be a good idea. The executables (and their names) could also be grouped by libc directories to keep things simple.


> I haven't thought about memory models, to make this more generally useful. But testlib.[ch] only depend on libc so should be re-usable.

Almost all ELKS applications are compiled in small model using the cdecl calling convention. The kernel uses medium model extensively, and there are a few (large) application using medium model cdecl. At some later point, it would great to be able to test both medium model cdecl, as well as small model regparmcall, which @tkchia recently contributed (but is mostly untested in libc). An upper-level test Makefile could be passed the memory model and calling convention to produce a set of test binaries in the future.

> Tests should auto-register with the test runner, maybe through linker magic. I haven't yet looked into how to do this; suggestions welcome. Maybe test registration should include "suite" and "name", and be filterable from the runner, or maybe that's too much bloat.

Can you explain what you're thinking for the test runner(s)? Would this be a shell script running on the ELKS machine? If the tests end up being separate binaries, a BASH script (ELKS `ash` which is usually /bin/sh and supports a lot of bash) could work well for that.

> More coverage. I would also like to touch on the syscalls that libc only wraps.

Good idea, those could be covered as part of the libc/system testing. Note that the full list of system calls is auto-generated from (and partially documented) in `elks/arch/i86/kernel/syscall.dat`.

> The test library needs to support creating and cleaning up temporary directories, for tests that want to do IO.

This could be done in C, or with potentially less work with a test runner written in bash.

> Not all assertion variations are fully implemented, and some some redundancy can still be reduced. This will get fully fleshed out as more tests are written.

Great job on distilling test macros from Cosmopolitan! `testlib.h` is fantastic. Note that ELKS currently doesn't support UTF8 natively, but the direct and BIOS console support most CP437 output, other than ASCII control characters that take precedence. So CP437 can probably be defined. I'm working on some other libraries that will provide for both UTF8 and CP437 for non-console output, but that isn't really ready for prime time yet. (That code lives mostly in elkscmd/tui, and is used by the `fm`, `matrix` and `cons` commands. The idea is that ultimately the library code will automatically translate unicode into CP437, UTF8 or ANSI sequences, depending on the output device.


> Some tests crash or hang elksemu. Haven't had time to compare to real hardware. Maybe test bug, maybe elksemu bug, maybe ELKS bug. Thrill of the chase.

Have you started using QEMU (yet) for ELKS? Many of us use it as a de-facto standard, those of us on macOS can't run elksemu. It would be nice to have a test runner that would completely execute on QEMU, as that would allow me and others to quickly help with found bugs and crashes. More thrill of the chase!


> Floats don't actually print for failed test cases despite __STDIO_PRINT_FLOATS. Looks like there are multiple build flags to control this...

Well, there shouldn't be actually, and it ought to "just work". Presently, \_\_STDIO_PRINT_FLOATS declares a global function pointer `__fp_print` initialized to call `dtostr`, versus being NULL. In the file libc/stdio/vfprintf.c, that `__fp_print` is called if not NULL to display floats. The global `__HAS_NO_FLOATS` isn't ever defined for ELKS libc, but is present for those that might want to build a special very small libc for their own purposes.

If you can point me to a specific float print failure case, I'll help track this problem down.


> I have plans to extend this with some benchmarking support, because I want to be able to easiliy benchmark potential optimizations on real hardware, and maybe it would be cool to be able to track some areas for performance over time to watch for regressions.

I had started working on some precision timing in elkscmd/debug/{stacktrace.c,printreg.S} function `rdtsc`, which returns a 64-bit cycle count which would work well, but it uses the RDTSC (Read Time Stamp Counter) instruction, which is only available on 386+ CPUs. However, as an option, requiring a 386+ for benchmarking support shouldn't be a problem, I don't think. My plans for the debug library were to allow for visual stack backtraces and/or runtime profiling with timestamps.

Thank you!

ccoffing:
> Can you explain what you're thinking for the test runner(s)? Would this be a shell script running on the ELKS machine? If the tests end up being separate binaries, a BASH script (ELKS `ash` which is usually /bin/sh and supports a lot of bash) could work well for that.

I have experience with C++ testing frameworks such as Catch2, and external test runners such as CTest (or, in our case, potentially a shell script).  But for this discussion, when I talk about a test runner, I'm thinking about how the C test executable runs the test cases contained within itself.  I think that's the more interesting aspect, and even if there is an external runner that calls all the test executables, each test executable still needs to have some intelligence internally to run more than one test (each of which will likely consist of multiple assertions) and summarize and report them.  This is just a matter of scaling and managing the tests.  (If ELKS were a fully tested 1.0 project, I wouldn't be surprised by 10-20 test executables, handling maybe ~500 test cases, which could be 10,000 assertions.  Hand-wavy order-of-magnitude numbers, but the point is that the shell script would probably only want to run and gather the results from the 10-20 executables.)

I've already got this core loop written, but at the moment telling it about each test is manual.  Cosmopolitan does some linker tricks to store the function pointers.  JART clearly is a wizard with the GNU toolchain, and I need to sprint to see her taillights.

> > The test library needs to support creating and cleaning up temporary directories, for tests that want to do IO.
> 
> This could be done in C, or with potentially less work with a test runner written in bash.

My first inclination is to write this in C.  It's not hard, and keeps me closer to the interfaces to more easily notice inconsistencies and errors (eat your own dog food).

> Have you started using QEMU (yet) for ELKS? Many of us use it as a de-facto standard, those of us on macOS can't run elksemu. It would be nice to have a test runner that would completely execute on QEMU, as that would allow me and others to quickly help with found bugs and crashes. More thrill of the chase!

Absolutely use QEMU.  So far, I have used cppcheck, native gcc + valgrind, clang --analyze, elksemu, qemu.  I have a beautiful NOS 286 AT case that I'm finally building, because it now has a purpose to run ELKS.  The wide variety of tools makes it quick to shake out potential bugs, but now I need to settle down and grind through to find the real (qemu, hardware) test failures.

> I had started working on some precision timing in elkscmd/debug/{stacktrace.c,printreg.S} function `rdtsc`, which returns a 64-bit cycle count which would work well, but it uses the RDTSC

I'll check this out.  I was thinking about doing this "old school" by averaging and backing out the overheads... but... dumb when we actually do have real tools now.  And yet, on second thought, 386 (and up) have much different cycle counts for some instructions.    As just one example, it is painful to see divides and modulo scattered over the code, when each might be 100-150 cycles on an old processor.  Yet memory references are almost free.  Modern processors are the inverse.  So the question is, when doing this low-level of an optimization, what is the target?

Great conversation, thanks.

ghaerr:
Hello @ccoffing,

> My first inclination is to write this in C. 

Go for it, you've got a lot of experience with this. We'll use your design, and you're doing a great job!

> If ELKS were a fully tested 1.0 project, I wouldn't be surprised by 10-20 test executables, handling maybe ~500 test cases, which could be 10,000 assertions. 

The thought of ELKS being fully tested is an admirable and lofty goal. I really like the idea of testing starting with libc, as well as hopefully later the capability of unit-testing various functions within application and kernel code. I plan on giving more thought on how libc as well as other applications could be compiled and run under a desktop OS as well as ELKS. We're already doing that with some newer applications, and the ability to find memory leaks and perform analysis using other compilers would be great. We may need another library other than libc in which to put some of the additional (non-standard) functions useful for testing and debugging, e.g. the elkscmd/debug library. Haven't thought yet how that ought to work, other than the library(s) should probably be built using the same calling conventions and memory models libc supports. If an application can also be compiled under a desktop OS, perhaps it should be as part of the standard build, just for testing.

> So far, I have used cppcheck, native gcc + valgrind, clang --analyze, elksemu, qemu.

That's a nice list, I'll have to check out cppcheck and clang --analyze, haven't used them before. I need to check whether valgrind will run on macOS (my dev OS).

For run-time simulation, I'm also looking hard at [Blink](https://github.com/jart/blink), a spinoff of Blinkenlights that doesn't require Cosmopolitan. It could have the advantage of providing a visual debugger for ELKS applications and possibly the kernel. What will be needed is a program loader (pretty easy) and a system call translator from ELKS to Linux/macOS/Posix (lots of typing for sure). There is also [EMU86](https://github.com/ghaerr/emu86) which can provide an simulated execution of the ELKS kernel. The debug library adds symbol support instead of looking at just addresses. All of this can be useful to just watch an application execute, possibly exposing timing and program loops not otherwise known.

> I have a beautiful NOS 286 AT case that I'm finally building, because it now has a purpose to run ELKS.

Sounds fun. You'll have to send pictures!

> And yet, on second thought, 386 (and up) have much different cycle counts for some instructions. 

Support and optimization for different processor variations was once half-baked into ELKS, and then removed, at least for the time being. Currently, all generated code is "optimized" only for the absolutely basic 8086/8088, including the very first version(s) that were used by the PCjr, that didn't even support long jumps. This is because it is now more important that ELKS run on any x86 and be bug-free and stable, rather than try to be faster for a 386+ variant, etc. There is some configurable kernel code that requires 386+, for instance XMS buffer support in extended memory (1M+).

> As just one example, it is painful to see divides and modulo scattered over the code, when each might be 100-150 cycles on an old processor. 

That makes a good case for benchmark/cycle performance testing. Perhaps we need a list of operations to avoid, and then search the application source tree for them.

The debug library I previously mentioned supports execution profiling (with C symbols) along with displaying a cycle count using the 386 RDTSC instruction. That in turn requires an ia16-elf-gcc modification that hasn't been sent upstream yet. The idea was to allow any executable to display performance benchmarks just by passing a `--ftrace` flag to it (idea [stolen from Cosmopolitan](https://justine.lol/ftrace/)).

> So the question is, when doing this low-level of an optimization, what is the target?

Yes, that is a good question! With benchmark timing added to libc testing, perhaps then experimenting with compiler CPU switches will give some good information. Still, however, all ELKS code currently needs to run on the oldest processors, until performance-enhancing configuration selection(s) are added.

Thank you!

ccoffing:
This is now a reasonable checkpoint.  I disabled a few tests that need more research.  Everything enabled passes under qemu.  Going forward, code and tests can be fixed together, and keep this passing.

image

ccoffing:
Currently installing with the `:other` tag, which I think only installs on the larger images.  Move it to a different tag if that makes more sense.

ghaerr:
Hello @ccoffing,

This is great, thank you for all your efforts!

> Going forward, code and tests can be fixed together, and keep this passing.

I haven't had time to fully look at how you're disabling some tests, but perhaps there might be an option to (temporarily) tag a test as failing, but a way to run all the tests so one can see at once what is failing, without having to recompile. I suppose I sometimes like to see all the problems at once lol. Perhaps something like `test_libc all`? In this way, everything is compiled in, but when `test_libc` is run with no arguments, only the non-fail-marked tests run.

Here's some quick comments on some commented out tests:
- malloc((size_t)-1): Currently, `malloc` only works when its 16-bit unsigned int argument is < 32768. Long story short, `malloc` depends on `sbrk` to get memory, and `sbrk` takes a signed argument whether to increase or decrease the heap brk point. This could possibly all be enhanced, but `malloc(-1)` will never work anyways, as that would require the entirety of the max 64K application data segment to be allocated, which won't ever work since all binaries use some .data for other purposes, (as well the stack has to be included in the 64K limit). In addition, for memory usage reasons, each ELKS binary is only allocated a max data space of 4K, and max stack of 4K; this is fixed at application load time. In order to increase either, the linker `-maout-heap=` and `-maout-stack=` arguments need to be specified at link time. This can be changed after link using the `chmem -h` or `chmem -s` command. You'll probably want to add `-maout-heap=0xffff` to the `test_libc` link line so that it gets a 64K data segment max allocation, and then revise the `malloc` test to ask for something less than 32K bytes for the time being.
- In test/libc/math.c, change `#define __STDIO_PRINT_FLOATS` to just `__STDIO_PRINT_FLOATS;`, after all \#includes, which must include \<stdio.h>. The macro works by initializing a hidden global function pointer to `dtostr`, which drags in floating point printf support.
- regex: not sure on why unresolved symbols, I've added this to my list. There are a couple of regex's in the source tree, I don't think the libc version has ever been tested. I have also been researching a simpler, much smaller regex from V6 that I am thinking of porting.
- Nice work on the string functions. I look forward to filling this out more fully, and then running with performance stats. I have a few ASM versions of string functions that I've been wanting to test to see whether they're faster or not than the C versions. This is complicated by the fact that they are currently 8086-only, which are usually much slower on 386+ systems. Users could have a way to measure libc functions on their specific CPU and allow us to see how various versions fare on different processors.
- (For performance stats, we could possibly implement the 386+ RDTSC cycle count automatically by your test runner, using an argument like `test_libc perf` to turn it on. It could report the cycle count of each test easily as a 64-bit number, which would show some basic timings. As mentioned before, the long long support and RTDSC is in elkscmd/debug. I can supply the routines for you if you'd like.
- `statfs` and `fstatfs` are not yet implemented.

> Currently installing with the :other tag, which I think only installs on the larger images.

Yes, CONFIG_APP_OTHER is usually set for images larger than 1440k, which will bring in test_libc. In your next PR, instead of using `:shutil` which is reserved for shell utilities, lets use `:test`, a new category for any test executables.

Thank you!

PR #1479, opened 2022-12-13
Add missing const to some libc declarations

ccoffing:
Obviously this may help catch bugs, but also makes it easier to use various static analyzers that expect the standard decl.


ghaerr:
Thanks @ccoffing, I've been meaning to do this myself, but you've beat me to it :)

PR #1480, opened 2022-12-13
Manpages

Mellvik:
- Moved Minix man pages not updated for or not relevant to ELKS to a new directory structure. Accidentally the move was done outside git, which makes the commit very noisy.
- Updated a few man pages
- Adjusted output from the `nslookup` command

Also renamed the `miniterm.1` man page back to `term.1` which is what the page is describing. `term`  is a Minix terminal emulator command not currently ported to ELKS. `miniterm` on the other hand is ported and much more powerful, maintained here: https://github.com/jonludlam/xen/blob/master/tools/misc/miniterm/miniterm.c - but unfortunately no manpage.


ghaerr:
@Mellvik, it seems the since you didn't use `git mv` when moving the man pages to `man-minix`, this PR doesn't delete the original entries. That means you will need to run `git rm` on the originals and push, or they will exist after this is committed. 

Given that you will be running `git rm` on each of them, it might be better to just remove the added pages and use `git mv` in the beginning, as then the history will be maintained as it should be. The .c file change and updated man pages look fine. Please advise.

Thank you!

ghaerr:
Hello @Mellvik,

Wow, you have been busy!! These updates look fantastic, especially `hier` (and `bootopts`). The online man pages are really starting to become very informative, thanks to your work.

It looks like `remsync` may still be existing under `man`, can you check whether it also needs a `git mv`?

Thanks, let me know when you are ready for commit.

Mellvik:
Thank you @ghaerr,

I actually decided to delete remsync, thinking it's not interesting in the ELKS context. Then I forgot to do it, mea culpa.

Then again, there are many such pages, so I'll mv that one too.

-M

ghaerr:
The reason I kept remsync is that there is a remsync command being built on ELKS!


On Dec 14, 2022, at 11:04 AM, Helge Skrivervik ***@***.***> wrote:

﻿

Thank you @ghaerr,

I actually decided to delete remsync, thinking it's not interesting in the ELKS context. Then I forgot to do it, mea culpa.

Then again, there are many such pages, so I'll mv that one too.

-M

Mellvik:
﻿
>  The reason I kept remsync is that there is a remsync command being built on ELKS!
>

Wow. Does it do anything? IOW does it work? My judgement may have been too shallow but it occured to me that we don't have the mechanisms that make this and some other minix commands useful.

-M

ghaerr:
Hello @Mellvik,

> The reason I kept remsync is that there is a remsync command being built on ELKS!
> Wow. Does it do anything? IOW does it work? My judgement may have been too shallow but it occured to me that we don't have the mechanisms that make this and some other minix commands useful.

Both `remsync` and it's associated `synctree` are built and included on the 1440k floppies. I've never tried running them, but I would assume they should work. For commands that can be compiled under 16-bit, ELKS does have most of the mechanisms required for them to work. As such, ELKS includes three elkscmd/minix{1,2,3} directories of (possibly untested) commands.

Since both commands are actually built and their associated man pages display properly, I would suggest leaving both remsync.1 and synctree.1 in elkscmd/man1. You might check if there are other MINIX commands being built and leave their man pages also, or at least flag them for updating - I didn't do any checking against this PR.

Thank you!

Mellvik:
Thank you @ghaerr,
this is interesting.
> Both remsync and it's associated synctree are built and included on the 1440k floppies.
> 
NO, and that's why I haven't seen them before. Now looking at what they do I'm getting interested. 

Maybe we have a mechanism here that can be used for more or less automatic ELKS system updates? I've been missing that forever…

I'm thinking running a sync between the ../target directory the the dev system and the ELKS machine, updating whatever is changed after comparing to an exception list.

I'll take closer look and see if/how they work.

In the meanwhile I'll git mv the remsync manpage so you can merge the PR.

—M

ghaerr:
> NO, and that's why I haven't seen them before.

Oops - you're right: elkscmd/Applications has both of them commented out. I'll add them back in for the 2880K floppy and HD distributions. They were removed along with some other applications back when we ran out of space on the 1440K floppy.

> In the meanwhile I'll git mv the remsync manpage so you can merge the PR.

Ok.

PR #1467, opened 2022-12-01
Update ethernet device names & more

Mellvik:
This PR changes all Ethernet device names per discussion in #1445 - which entails minor mods to many system files and the setup.8 man page.

Also included is an updated `fsck` man page, an enhancement to `cp` which adds a -f (force) option, making the `sys` script more useful when used to update an existing installation. The `sys` script has been updated accordingly, as has the `cp` man page.

The `wd` driver has been made fully compatible with the SMC/WD 8013 NIC, including the auto-detection of 8003 vs 8013 types. More testing of the switching between interfaces on a running system is required and ongoing.

Updates to the wiki and other documentation pages will follow when this PR has been merged.

-M

PR #1481, opened 2022-12-16
[net][ne2k] Change default buffer size for 8bit (ne1k) mode

Mellvik:
**Important Ne1k update**
The short version: The assumption that the NE1K is limited to use only 4k of buffer space to run reliably, is wrong and based on a typo. This update changes the ne1k default buffer setting to 8k and makes that the default when 8bit bus is activated - automatically or manually. The change has been extensively tested and eliminates the peculiarities of the forced 16k buffer setting and the very limiting 4k buffer setting. This is a great (and overdue) simplification and performance improvement for the ne1k.

The narrative below is the mostly boring story about how my mistake was discovered. Unless you have too much time on your hands, skip to the summary of changes at the end.

**Background** 
When the ne2k driver was expanded to cover the 8bit ne1k, a lot of time was spent chasing an elusive NIC buffer problem. Eventually a quote from the book Networking and Internetworking with Microcontrollers (Fred Eady, 2004) popped up in a Google search and shed some light on the problem. The Realtek RTL8019AS controller chip (functionally equivalent to the DP8390C) does not support 16k buffer space when running in 8 bit mode. The book says:
_"In 8-bit mode, … the RTL8019AS use a 4 Kb on-chip RAM area despite the RTL8019AS' 16 Kb specification."_ 

Unsurprisingly, when reducing the buffer used to 4k, the ne1k worked fine - with one serious caveat: With no OS/driver level packet buffering in ELKS and a very small on-NIC buffer, the packet loss when under load became unbearable. A reduced MTU alleviated the problem while also reducing the overall throughput of file transfers. Mtu=1000 turned out to be a reasonable compromise.

Given the penalties of the 4k buffer, experimentation with the full 16k continued. It eventually turned out that recovery from the weird buffer errors was possible in most cases, generating many errors (and thus retransmits) but still delivering higher throughput than the 4k setting. After heavy testing, 16k buffer was made the default for the ne1k - with 4k as an option, with mtu=1000 recommended in the latter case.

All testing was done on Compaq portables - 286/12.5 MHz & 386/20MHz, using 2 different ne2k interfaces (for reference) and a recent ne1k from Weird Electronics (yes, interesting name). Also - and as it turns out later not so smart - most of the heavy testing was done with large packet short interval pings.

**New environment, new results**
Fast forward to December 2022: While running all kinds of tests on 'driver/interface mixing' and manpage preparations (#1467) recently, I discovered erratic behaviour from the ne1k while running file transfers (FTP) on a faster system than the Compaqs, the 40MHz AMI 386SX. Interestingly, the errors came when fetching files FROM the ELKS system: No buffer overruns, just a lot of small ack packets. Typically I got 6-7 bad packets (some 10% of the total), then a hard hang requiring power cycling the machine.

Completely repeatable, certainly not acceptable - and hard to diagnose, not the least because the symptoms (errors) were timing sensitive. A printk in the driver's interrupt service routine eliminated the problem (of course with severe consequences for speed). IOW lowering the packet rate made the entire NIC buffer useable. What all errors had in common though, was that they happened in the same part of the buffer - somewhere between the 8k and 10k marks. Interesting but why does it go away when we slow down the rate via printk?

Hitting the wall on testing, which included running similar tests on a ne2k in forced 8bit mode on a Compaq and getting the exact same results, I decided to go back to the source of the 4k limitation, the book mentioned above, which in the meanwhile had made it to my bookshelf from some used-books store online.

Having the full text and taking the time to read the whole chapter (aptly named "Let's Do It Again") revealed the 'secret'  I had missed thus far. The 4k limit is a TYPO. Seriously. 

The quote above is from page 294 in the book. Six pages later, in a figure caption, it says "_.. note that only 8K buffer RAM is available in the RTL8019AS._" What? With considerable scepticism I read it again. It had to be typo. After all, what it says is objectively incorrect. The RTL8019AS has 16k SRAM, not 8K. Discarded. 

Fast forward to page 318 which is where the  bomb exploded: "_The RTL8019AS' datasheet tells us point blank to not exceed 0x60 as a PSTOP register value in 8-bit mode._" PSTOP being the upper endpoint of the buffer. 0x60 is the exact middle of the 16k buffer space which is divided into 256B pages, starting at 0x40, ending at 0x80. Meaning 'do not use the upper half of the buffer'. Really? 8K? So the caption was no typo after all? How did I miss that? 
I went back to the RTL8019AS datasheet (http://www.ethernut.de/pdf/8019asds.pdf), searched for 'PSTOP' and sure enough: Top of page 15 defines PSTOP and says
"_The Page Stop register sets the stop page address of the receive buffer ring. In 8 bit mode the PSTOP register should not exceed to 0x60, in 16 bit mode the PSTOP register should not exceed to 0x80._" [sic]
Loud and clear. Well, clear enough. The language isn't all that good (as in 'exceed to' and more), but I'm being picky trying to share the blame with someone. The truth is I didn't use the Realtek documentation much at all. I used the NS DP8390C datasheet (https://usermanual.wiki/m/abdcdadc75f1281ab4d06ff4c58665a3afeef164405e3755943029d89c6445be.pdf) which doesn't mention this limitation even though its presence in the 8390 has been verified. 

Back to the driver, a minor adjustment to the buffer parameters and more testing: Voila! 'Normal' behaviour (no odd errors requiring special recovery) and decent performance. In addition to fixing the mentioned problems, the 8K buffer gives the ne1k performance on par with (or slightly better than) the wd8003 - as should be expected.

The special recovery code is still in the driver, pending more testing and enabling debug runs with 4/8/16k buffers. 

Summary of changes
- When a ne1k NIC is detected, the NIC buffer is set to 8K (instead of 16k)
- When 8 bit mode is forced via a flag setting, the buffer is also set to 8K
- Buffer size overrides via flag settings still work as before (it's actually possible to force buffer size to 32k, if someone is playing with the flags, it's assumed they know what they're doing (manpage will be updated to that effect)).
- A lower MTU is no longer required for the 'native' NE1K buffer setting.

File transfer speed (FTP) for the ne1k on the 40MHz 386SX is 96KB/s incoming (to /dev/null), 50KB/s outgoing (512b max packet size).

### DIFF
--- a/elks/arch/i86/drivers/net/ne2k-asm.S
+++ b/elks/arch/i86/drivers/net/ne2k-asm.S
@@ -10,8 +10,7 @@
 // Updated by Santiago Hormazabal on Dec 2021:
 //  . 8 bit access if CONFIG_ETH_BYTE_ACCESS is set, using a ne1k patch from
 //    NCommander.
-// apr-2022 (HS) : Support auto detection of 8bit mode, set 4k or 16k buffer size when in 8bit mode
-//	     depending on configuration, 16k is the default, 4k is 'per spec'.
+// apr-2022 (HS) : Support auto detection of 8bit mode, set 8k buffer size when in 8bit mode.
 //	     Rewrote overflow handler, added direct access to many registers from C code
 //	     Cleaned up initialization code, optimized 8bit I/O
 //
@@ -22,7 +21,7 @@
 // 'BOUNDARY' points to the next block to be read from the NIC
 // The two must not be equal. If the ring buffer is empty, BOUNDARY = CURRENT -1
 // So the next block to read is not BOUNDARY but BOUNDARY + 1. Easy - except at
-// the wrap-around point: BOUNDARY may be 80 or 50, while the next block to read is 46.
+// the wrap-around point: BOUNDARY may be 80 (60 if 8bit), while the next block to read is 46.
 // For simplicity (to avoid the ring wraparound logic on every read)
 // the driver uses a variable to hold the next block to read - _ne2k_next_pk.
 // For debugging, this variable is also available to the C part of the driver.
@@ -79,7 +78,7 @@ io_ne2k_reset      = 0x1F	// Really a port, not a register, force HW reset of th
 tx_first           = 0x40
 rx_first           = 0x46
 rx_last_16	   = 0x80
-rx_last_8	   = 0x50	// For 4k in 8 bit mode - per spec. 
+rx_last_8	   = 0x60	// For 8k in 8 bit mode - per spec. 
 
 //-----------------------------------------------------------------------------
 	.data
@@ -99,7 +98,7 @@ _ne2k_is_8bit:
 	.word 0
 
 _ne2k_rx_last:
-	.byte rx_last_16	// default to the 16 bit value
+	.byte rx_last_16	// default to 16K
 
 	.text
 
@@ -210,7 +209,7 @@ dma_write:
 	mov	current,%bx		// setup for far memory xfer
 	mov	TASK_USER_DS(%bx),%ds
 	cld
-	test	$1,%ax		// checking _ne2k_is_8bit
+	test	$0x80,%ax		// checking _ne2k_is_8bit
 	jz	wr_loop_w
 
 	// Byte loop
@@ -287,7 +286,7 @@ buf_local:
 	//mov	net_port,%dx
 	add	$io_ne2k_data_io,%dx
 	cld			// clear direction flag
-	testw	$1,_ne2k_is_8bit
+	testw	$0x80,_ne2k_is_8bit
 	jz	word_loop0
 
 byte_loop:
@@ -662,7 +661,7 @@ ne2k_base_init:
 	//mov	net_port,%dx
 	add	$io_ne2k_data_conf,%dx
 	mov     $0x49,%al	// set word access
-	testw	$1,_ne2k_is_8bit
+	testw	$0x80,_ne2k_is_8bit
 	jz	1f
 	dec     %al	// if in 8bit mode, 
 			// set byte access, data_conf reg = 0x48
@@ -683,6 +682,7 @@ ne2k_base_init:
 //-----------------------------------------------------------------------------
 // NE2K initialization
 // Called from device open
+// Uses AX, DX only
 //-----------------------------------------------------------------------------
 
 	.global ne2k_init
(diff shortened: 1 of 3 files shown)


Mellvik:
Additional notes:
- the changes to `netstat.h` are cosmetic only.
- a possible overflow handling problem in the driver (ne1k mode only) is being investigated.

-M

ghaerr:
Hello @Mellvik,

Wow, super nice description and background on this fix! ELKS should have a blog, your well-written and interesting description would make a great article. I downloaded Fred Eady's book a while back and started reading it, it gets complicated in a hurry. Nice figuring out the typo (finally)!

Some thoughts for the future: perhaps consider using \#defines for the bit-tests (e.g. 0x01 or 0x80) in `_ne2k_is_8bit`, this might make it easier to understand which bit means what in that variable. (I'm finding it hard to tell whether there are multiple bits used in that variable still, or this PR changes it to only a single bit). The ne2k-asm.S file can include a .h file, and portions of the header file which are not \#defines can be protected from the assembler using \#ifndef \_\_ASSEMBLER\_\_.

Thank you!

Mellvik:
Thank you @ghaerr - 

ELKS blog - yes, that sounds like an interesting idea - 'Practical ELKS' or something…
I haven't looked yet, maybe there is a blogging mechanism in github already.

The Eady book is very specialized - oriented towards the microcontroller target, but there is a lot of interesting experiences in there. And even a CD with spec sheets and stuff (code etc.) (that I admittedly haven't opened yet). 

> 
> Some thoughts for the future: perhaps consider using #defines for the bit-tests (e.g. 0x01 or 0x80) in _ne2k_is_8bit, this might make it easier to understand which bit means what in that variable. (I'm finding it hard to tell whether there are multiple bits used in that variable still, or this PR changes it to only a single bit). The ne2k-asm.S file can include a .h file, and portions of the header file which are not #defines can be protected from the assembler using #ifndef __ASSEMBLER__.
> 
Yes, I agree, this is definitely needed. I'll look into that. BTW it's time to rename the _ne2k_is_8bit to something flag-related now that the 8bit flag is just that, one of many.

-M

ghaerr:
> BTW it's time to rename the _ne2k_is_8bit to something flag-related now that the 8bit flag is just that, one of many.

When you do the rename, I suggest removing the '\_' prefix from the name, as it makes source hard to read, and very few other kernel variables are named that way. (I plan to fix some of those also). You'll still probably need the 'ne2k_` prefix, unfortunately as the variable can't be declared static as its shared with the .S file.

Mellvik:
Good idea, I always hated those leading underscores.

And thanks for the hint about the __ASSEMBLER__,
That's very useful.

-M

PR #1483, opened 2022-12-17
Fix strncat

ccoffing:


ccoffing:
found by unit test (will push update to those soon)

ghaerr:
Hello @ccoffing,

Holy heck, what a find, thanks to your test suite! A quick look shows that this means that `kilo`, `tinyirc`, `write`, `sash` history, and some other applications are all buggy in certain cases.

Thank you!

PR #1486, opened 2022-12-23
[cmds/libc/man] Various cleanups

ghaerr:
Various small cleanups from recent PRs, including:

`test_libc` include in CONFIG_APP_OTHER category only.
`remsync` and `synctree` added to distribution for CONFIG_APP_OTHER, including man pages.
`basic` and `sash` call stdio constructor `__io_init_vars()` only once using `__YOINK`.
`stdbool.h` removed (unused) from a couple source files.
Slightly decreased code size in `atexit`.
Fixed a couple `const` compilation warnings in libc.


ghaerr:
Hello @tkchia,

The C library now compiles with a number of warnings `libc/stdio/__io_init.c:24:1: warning: constructor priorities from 0 to 100 are reserved for the implementation`. It would be nice if there were a way to turn that off, in accordance with ELKS libc and kernel not producing any warnings at all. I tried using the documented `-Wno-prio-ctor-dtor` but it seems that's not implemented from upstream. The warning itself is handled in `cross/build/gcc-src/gcc/c-family/c-common.c::get_priority()`.

Thank you!

tkchia:
Hello @ghaerr,

> It would be nice if there were a way to turn that off, in accordance with ELKS libc and kernel not producing any warnings at all.

I believe `-Wno-prio-ctor-dtor` was only added some time after GCC 6.3.0 came out (around GCC 8.x or 9.x?).

I just kind of "back-ported" this option to `gcc-ia16` a few hours ago (https://github.com/tkchia/gcc-ia16/commit/586d4dc432d963a26fda67ceeafc0307349dc955).  It should take effect if you bump up the `gcc-ia16` version number for the ELKS build.  But even if you are on an older `gcc-ia16`, you should be able to specify `-Wno-prio-ctor-dtor` without any issues &mdash; GCC will just harmlessly ignore the option.

Thank you!

ghaerr:
Hello @tkchia,

> I just kind of "back-ported" this option to gcc-ia16 a few hours ago

Wow, thank you!

It is interesting to see how an option is added to gcc. Does this option default to 0, meaning one must use `-Wprio-ctor-dtor` to see the warning we see now, or will we need to add `-Wno-prio-ctor-dtor` in the libc Makefiles?

Given that most libc Makefiles govern many .c files, I was thinking of just adding `#pragma GCC diagnostic ignored "-Wprio-ctor-dtor"` to each libc .c file that used constructors with priority 100 and below. Which way would you think is best?

> It should take effect if you bump up the gcc-ia16 version number for the ELKS build.

We definitely should increase the version number as soon as the weak symbol mechanism you're working with for stdio floating point printf is finished, and possibly for a slight instrumentation mod I would like to enhance ELKS debugging support. Would you be interested in seeing a patch or PR for that (as described elsewhere) if I were to add a new compiler option instead of `-finstrument-functions`? (I was thinking it would be nice to use something like `-finstrument-functions=2` to select a slightly different signature for the cygwin instrument functions but haven't figured out how to do that, since that is all the enhancement requires).

Thank you!

ghaerr:
Hello @tkchia,

> Would you be interested in seeing a patch or PR for that (as described elsewhere) if I were to add a new compiler option instead of -finstrument-functions

I uploaded a patch showing the small change required for this functionality for your consideration in https://github.com/tkchia/gcc-ia16/issues/111#issuecomment-1364102298.

Given the other compiler mods being used to add new functionality to ELKS recently, this would be a great time to have this added.

Thank you!

tkchia:
Hello @ghaerr,

> Does this option default to 0

My understanding of the `c.opt` is that it is enabled on default.

> I was thinking of just adding `#pragma GCC diagnostic ignored "-Wprio-ctor-dtor"` to each libc .c file that used constructors with priority 100 and below. Which way would you think is best?

I am fine with either approach, but I am _slightly_ partial towards using a `#pragma`.  It seems to be a bit clearer and more explicit. :slightly_smiling_face: 

> I uploaded a patch showing the small change required for this functionality for your consideration in https://github.com/tkchia/gcc-ia16/issues/111#issuecomment-1364102298.

Thanks &mdash; let me take a look.

Thank you!

ghaerr:
Hello @tkchia,

> I am slightly partial towards using a #pragma.

Agreed. I will add that after we finish with the compiler enhancements and increment the toolchain version number(s). (Unfortunately, the \#pragma won't ignore the `-Wprio-ctor-dtor` option if it doesn't exist, and produces an unknown pragma warning. That's ok though, I have this added to my list.

Thank you!

PR #1487, opened 2022-12-23
[libc] Initialize environ &/or _program_filename only if needed

tkchia:
This patch allows the C program startup code to skip the initialization (& definition) of the variables `environ` & `_program_filename`, if the program will not actually use these.

### DIFF
--- a/elks/elks-medium.ld
+++ b/elks/elks-medium.ld
@@ -4,6 +4,7 @@ SECTIONS {
 	.text 0x10000 : {
 		/* IA-16 segment start markers. */
 		KEEP (*(".inithead!"))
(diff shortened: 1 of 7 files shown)


ghaerr (approved):
Nice! So the corresponding `.preinit` sections only get pulled in if the linker finds itself looking for `environ` or `_program_filename`. Pretty tricky, and our startup code is getting smaller but more functional than ever!

While on the business of keeping things quite small, I still think the MAXONEXIT being 32 is way overkill - we're not C90 compliant, the compiler never generates `atexit` code automatically, and a review of our sources show that only 2-3 programs are coded to use `atexit` and they call it just once. IMHO, the data used, especially in editors and with our 64K max data segment, would be better available for user program use.

Thank you!

PR #1489, opened 2022-12-25
[kernel,build] Cleanup IRQ and config options, add MK-88 and no hw timer

ghaerr:
Adds support for Russian MK-88 PC (discussed in #1327).
Adds support for systems without hardware timers, and for systems to use BIOS timer, without reprogramming the hardware timer.

Adds the following config options to `make menuconfig`:
```
"MK-88 Hardware (IRQ 3 keyboard)" - CONFIG_HW_MK88
"Use INT 0Fh in idle loop for timer" - CONFIG_TIMER_INT0F (for systems or emulators without timer by simulating IRQ 7)
"Use INT 1Ch from BIOS for timer" - CONFIG_TIMER_INT1C (allow ELKS to run without reprogramming hardware timer)
```

Deletes the following config options:
```
CONFIG_IDLE_HALT (always halt in idle loop until next interrupt to save power)
CONFIG_APP_BYACC
CONFIG_APP_M4
CONFIG_COMPAQ_FAST renamed to CONFIG_HW_COMPAQFAST
```

Cleans up interrupt/irq handling code and adds better comments.
Removes `disable_timer_tick` on IBMPC and PC-98 systems: unneeded because IRQ 0 vector not rewritten until `request_irq` called.

Tested on QEMU.


toncho11:
I tested a 360 kb minix image produced by Github actions for this PR on an IBM 5160 and it loads OK, keyboard is OK.
Date was 2009 though on the computer (there is no clock in the computer). Maybe you have set some default date?

ghaerr:
Hello @toncho11,

Thanks for testing this. I'm thinking the date issue is likely that the `clock` command is reading from a non-existent RTC at startup... Has this changed from previous versions?

Thank you!

PR #1491, opened 2022-12-26
[libc] Fix regparmcall system calls with no arguments

ghaerr:
Fixes regparmcall calling convention bug for system calls with no arguments; initially discussed in https://github.com/jbruchon/elks/pull/1459#issuecomment-1364752365.

The problem was that AX/BX needed to be swapped in the no argument case, as well as the one argument case.

Fixes minor argument count error in `alarm` system call definition.
Adds more debug info to /bin/init, which was used to track down this bug.
Adds commented out CFLBASE definition in elkscmd/Make.defs to turn on regparmcall passing for ELKS applications. Many programs work with the notable exception still of /bin/sh (ash).

Note: programs compiled with regparmcall are **much** smaller than using cdecl! More information and testing coming.

### DIFF
--- a/elks/arch/i86/kernel/syscall.dat
+++ b/elks/arch/i86/kernel/syscall.dat
@@ -45,7 +45,7 @@ setuid		+23	1
 getuid		+24	1	* this gets both uid and euid
 stime		25	2	- this must not exist - even as a libc.
 ptrace		26	4	@ adb/sdb/dbx need this.
-alarm		+27	2
+alarm		+27	1
 fstat		+28	2	 
 pause		29	0	 
 utime		+30	2	 
--- a/elkscmd/Make.defs
+++ b/elkscmd/Make.defs
@@ -75,7 +75,7 @@ ELKS_VSN=$(shell printf '%s.%s.%s-pre%s' $(E_V) | sed 's/-pre$$//')
 ifeq ($(PLATFORM),i86-ELKS)
 	CC=ia16-elf-gcc
 	CFLBASE=-fno-inline -melks-libc -mcmodel=small -mno-segment-relocation-stuff -mtune=i8086 -Wall -Os
-	#CFLBASE += -mregparmcall -march=any_186
+	#CFLBASE += -mregparmcall
 	LD=ia16-elf-gcc
 	LDFLAGS=$(CFLBASE)
 	CHECK=gcc -c -o .null.o -Wall -pedantic
--- a/elkscmd/Make.defs
+++ b/elkscmd/Make.defs
@@ -75,6 +75,7 @@ ELKS_VSN=$(shell printf '%s.%s.%s-pre%s' $(E_V) | sed 's/-pre$$//')
 ifeq ($(PLATFORM),i86-ELKS)
 	CC=ia16-elf-gcc
 	CFLBASE=-fno-inline -melks-libc -mcmodel=small -mno-segment-relocation-stuff -mtune=i8086 -Wall -Os
+	#CFLBASE += -mregparmcall -march=any_186
 	LD=ia16-elf-gcc
 	LDFLAGS=$(CFLBASE)
 	CHECK=gcc -c -o .null.o -Wall -pedantic
--- a/elkscmd/bc/Makefile
+++ b/elkscmd/bc/Makefile
@@ -31,7 +31,7 @@ LEX = flex -I -8
 LOCALFLAGS=-D_POSIX_SOURCE
 
 # Try compiling this program using the `stdcall' calling convention...
-CFLBASE += -mrtd
+#CFLBASE += -mrtd
 
 # For ELKS, bc needs more data segment space than the kernel-given default.
 LDFLAGS += -maout-heap=0xb000
--- a/elkscmd/sys_utils/init.c
+++ b/elkscmd/sys_utils/init.c
@@ -175,9 +175,8 @@ FILE  stderr[1] =
 
 void parseLine(const char* line, void func())
 {
-	char *a[5];
 	int k = 0;
-	char buf[BUFSIZE], *p;
+	char *p, *a[5], buf[BUFSIZE];
 
 	strcpy(buf, line);
 	a[k++] = p = buf;
@@ -186,12 +185,22 @@ void parseLine(const char* line, void func())
 		p++;
 	}
 	if (*p == '#') return;
-	while (k < 4) {
+	while (k < 5) {
 		/* looking for the k-th ':' */
 		while (*p && *p != ':') p++;
+		if (*p == 0) {
+			*p = 0;
+			a[k++] = NULL;
+			continue;
+		}
 		*p = 0;
 		a[k++] = ++p;
 	}
+#if DEBUG
+	for (k=0; k<5; k++) {
+		debug("a[%d]='%s'\n", k, a[k]);
+	}
+#endif
 	func(a);
 }
 
@@ -207,6 +216,7 @@ void scanFile(void func())
 	while (fgets(buf, BUFSIZE, fp)) {
 		if (buf[strlen(buf)-1] == '\n')
 			buf[strlen(buf)-1] = '\0';
+        debug("[buf='%s',%d]\n", buf, strlen(buf));
 		parseLine(buf, func);
 	}
 	fclose(fp);
@@ -261,10 +271,10 @@ pid_t respawn(const char **a)
     int fd;
     char *devtty;
 
-    if (a[3] == NULL) return 1;
+    if (a[3] == NULL || a[3][1] == '\0') return -1;
 
     pid = fork();
-    if (pid == -1) fatalmsg("No fork\r\n");
+    if (pid == -1) fatalmsg("No fork (%d)\r\n", errno);
     if (pid) debug("spawning %d '%s'\r\n", pid, a[3]);
 
     if (0 == pid) {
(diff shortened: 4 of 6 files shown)


tkchia:
@ghaerr : argh, apologies for the mistake.  Thank you!

tkchia:
@ghaerr : incidentally, I think the bug in the `alarm` argument count will show up most clearly under the `cdecl` calling convention, since it will pop the wrong number of arguments off the stack.  This is probably something to keep in mind when we do testing. :thinking: 

Thank you!

tkchia (approved):

ghaerr:
Hello @tkchia,

> I think the bug in the alarm argument count will show up most clearly under the cdecl calling convention, since it will pop the wrong number of arguments off the stack.

Actually, you mean `stdcall` for that error, correct? When using `cdecl`, the compiler pops the arguments using the C declaration (caller pop), rather than the code in syscall0.S for `stdcall`, which pops the arguments before return (callee pop).

It appears that `regparmcall` performs callee pop when the argument count is four or larger. Is there documentation on that? I have ascertained that `regparmcall` uses the following registers in order for parameters: AX, CX, DX (with BX scratch and additional arguments pushed in reverse order).

Also, based on your statement in https://github.com/jbruchon/elks/pull/1459#issuecomment-1364972016, I will remove the `-march=iany_186` from this PR, as ELKS is currently still built to run on the oldest of 8086/8088 machines.

Thank you!

tkchia:
Hello @ghaerr,

> > I think the bug in the alarm argument count will show up most clearly under the cdecl calling convention, since it will pop the wrong number of arguments off the stack.

> Actually, you mean `stdcall` for that error, correct? When using `cdecl`, the compiler pops the arguments using the C declaration (caller pop), rather than the code in syscall0.S for `stdcall`, which pops the arguments before return (callee pop).

Sorry, yes, I misspoke &mdash; I meant `stdcall`.

> It appears that `regparmcall` performs callee pop when the argument count is four or larger. Is there documentation on that? I have ascertained that `regparmcall` uses the following registers in order for parameters: AX, CX, DX (with BX scratch and additional arguments pushed in reverse order).

`gcc-ia16` has an info file which should be installed as `gcc.info` somewhere in the installation directory.

Thank you!

ghaerr:
For those with time to dig deeper into the ongoing problems with `regparmcall`: by uncommenting `CFLBASE += -mregparmcall` in elkscmd/Make.defs, and then `cd elkscmd; make clean; make; cd ..; make`, the entirety of /bin will be compiled with the regparmcall calling convention. /bin/sh still fails, but /bin/sash works, and can be used via logging in via `toor`. There seem to be some issues with exec'ing most commands, but `printenv` works.

What we need is a good visual debugger. I am hoping at some point to port `Blink` to ELKS.

The following shows a typical session:
```
ELKS 0.6.0

login: toor
# printenv
ROOTDEV=/dev/fd0
USER=toor
HOME=/root
SHELL=/bin/sash
TERM=ansi
TZ=GMT0
HISTORY=30
HISTPAGE=20
PATH=/bin
# date
Thu Jan  1 00:00:11 1970
(3)STACK OVER MINSTACK by 52 BYTES
(3)STACK OVER MINSTACK by 52 BYTES
sys_brk(3) fail: brk 594a over by 5336 bytes
```

/bin/sh fails on start, with its error message not being able to find argv[0], the program name. I'm not sure at this point whether the issues are library or compiler related.

The problem with compiling `bc` was fixed by removing the `-mrtd` in its local Makefile, causing confusion when both that and `-mregparmcall` were specified.

Mellvik:
@ghaerr, @tkchia -

I've been following this thread with great interest and I'm inclined to participate in testing.

When following the instructions below and compiling under the are regime for testing, what would we be looking for? Stack screwups, arg screwups, etc.?

It seems to me the changes shouldn't need real HW testing for verification. If it works in QEMU it works, right?

—M

ghaerr:
Hello @Mellvik and @tkchia,

That'd be great to help testing - however for the moment the shells are broken badly, in that /bin/sh crashes on an error message at start and /bin/sash seems to get corrupted after running most programs. As seen above, other programs seem to be getting stack corruption and/or the heap break address isn't getting set properly. After these are fixed, I'm hoping most programs start working, at which time we'd need to just generally use the system to see where other anomalies may lay. You may want to wait for that, as it will make things a lot easier (although the fun sometimes lies in doing object file disassemblies to see exactly what code is being generated and tracing from there!).

The technical change in `-mregparmcall` is that arguments to functions are passed in registers for functions with 1 to 3 non-variadic (non-variable) parameters, and on the stack for 4+. @tkchia modified all ASM files in the C library for this change, and the interface to ELKS system calls was also changed. Thus, we need to test function calls in groups of 0, 1, 2, 3, 4 and 5 parameters to test whether that interface is working, as well as any C library or user program with .S/,.s files. We will need to inspect user programs to find examples of each class of system call and then test that program looking for correctness.

There seems to be a huge benefit is possibly switching to regparmcall, at least for some programs. In my initial testing, for instance, /bin/sh was 3k smaller just from the argument passing changes.

@tkchia, can you clarify whether the `-mregparmcall` is new to the `ia16-elf-gcc` code generation itself, or is the new implementation limited only to the ELKS C library? Are there other specific areas where one might find trouble (code generation, ASM files, system calls, or other)? Also seperate issue: is ES always saved, or can it ever be scratch on function entry? I found a case in irqtab.S where ES is assumed == DS == kernel data segment on function entry.

As part of the debugging, we also need to test each of the certain special fast functions in the C library that have ASM implementations: these include `memcpy`, `memchr`, `strcpy` and `strlen` at the moment. I removed their ASM implementation in libc/asm/config.h for the above testing, but that didn't turn out to be the problem. Also, each of those functions pass @ccoffing's `test_libc` unit testing, so they're probably OK (thank you @ccoffing, your enhancement came in very handy to rule that out with the /bin/init problems). Nonetheless, we plan on adding the ability to swap C vs ASM versions with more testing for upcoming performance measurements.

> It seems to me the changes shouldn't need real HW testing for verification. If it works in QEMU it works, right?

Yes.

Thank you!

tkchia:
Hello @ghaerr,

> @tkchia, can you clarify whether the `-mregparmcall` is new to the `ia16-elf-gcc` code generation itself, or is the new implementation limited only to the ELKS C library?

`-mregparmcall` for the MS-DOS target has been stabilized by around February 2020, and my `libi86` regression tests regularly stress-test it.  My guess is that any problems are likely not due to `gcc-ia16` code generation bugs.

> Are there other specific areas where one might find trouble (code generation, ASM files, system calls, or other)?

I wonder if variadic functions might be causing part of the problem.  For syscalls specifically, there are 3 syscalls `open`, `fcntl`, and `ioctl`, whose corresponding `libc` functions are variadic, but I believe these have already been handled as special cases.  But there might be other variadic functions that are not syscalls that are causing the problem.  Or the actual problem(s) might be something totally different.

> Also seperate issue: is ES always saved, or can it ever be scratch on function entry?

`es` is expected to be saved by the caller, but it is not required to point to any particular place upon function entry.  It is probably best treated as a(nother) call-saved register.

Thank you!

tkchia:
Hello @ghaerr, hello @Mellvik,

This is interesting (and hopefully useful): running `sash` under `elksemu` on Linux quite predictably causes a crash:
```
$ elksemu/elksemu elkscmd/sash/sash -c ls
Segmentation fault (core dumped)
```

> What we need is a good visual debugger. I am hoping at some point to port `Blink` to ELKS.

Agreed.  I do find `elksemu` to be quite helpful, but it is still not very friendly to use for debugging, plus it only runs on Linux/x86 or Linux/x86-64 for now.  It will be nice if `elksemu` or `blink` (or both) can be beefed up to more easily debug ELKS userland programs.

Thank you!

ghaerr:
Hello @tkchia,

Thanks for the report of `sash` crashing `elksemu`. It would be interesting to know where `elksemu` is crashing, as I can't run it on macOS.

I am right in the middle of debugging `ash` (/bin/sh) and have found it's exiting in the main loop just after reading /etc/profile and /root/.profile, where for some reason `open` is being passed a bogus first argument. Interestingly, the main loop (in ash/main.c) uses both `setjmp` and a function `setstackmark(&markstruct)` which are used to mark the stack and roll it back. I have looked at setjmp.S and it seems fine, and the code generated for the call to `setstackmark` which also looks fine, but there is corruption happening nearby. I don't have any simple setjmp test code.

> It will be nice if elksemu or blink (or both) can be beefed up to more easily debug ELKS userland programs.

It's probably going to be lots of work to get blink working, as I only have it almost completing an ELKS boot at this time before it itself crashes with an `abort`. However, I did enhance your [modification](https://github.com/tkchia/reenigne/blob/master/8088/86sim/86sim.cpp) of Andrew Jenner's 8086 simulator, which I have working [here as 86sim](https://github.com/ghaerr/86sim). I wrote a very small disassembler using Jenner's source, which combined with 86sim will load and run ELKS programs, complete with runtime execution disassembly. What it needs is the rest of the system calls fleshed out (I separated the DOS calls from ELKS, and just got enough working to run `banner` and simple programs).

I hope to combine that work with the new symbol-loading code in ELKS debug library to produce a symbolic disassembler. I also have a fork of [EMU86](https://github.com/ghaerr/emu86) which uses .map files for symbolic execution of the kernel. I hope to enhance that using the linker symbols added in the debug library also.  We're almost there!

Thank you!

ghaerr:
@tkchia, It appears `setjmp/longjmp` may be the culprit in `ash` - do you have any other test code for it? The stack seems to be getting trashed in `cmdloop`, which is a called procedure from main only once, but is being entered twice.

[EDIT: `wait4` turned out to be the culprit].

PR #1492, opened 2022-12-28
[libc] Fix regparmcall system calls with 4 and 5 parameters

ghaerr:
This took a while, but finally figured out that all system calls with 4 or 5 parameters were broken when `-mregparmcall` used. There are only a few system calls with 4 or 5 parameters, but both shells used `wait4`, which incorrectly popped the wrong argument count, thus corrupting the stack.

At this time, quick QEMU testing shows many more commands and both shells seemingly working. @Mellvik, now might be a good time to try more testing, by uncommenting the CFLBASE line in elkscmd/Make.defs.

### DIFF
--- a/elks/kernel/exit.c
+++ b/elks/kernel/exit.c
@@ -37,7 +37,8 @@ static void reparent_children(void)
 	}
 }
 
-int sys_wait4(pid_t pid, int *status, int options)
+/* note: 'usage' parameter ignored */
+int sys_wait4(pid_t pid, int *status, int options, void *usage)
 {
 	register struct task_struct *p;
 	int waitagain;
--- a/libc/system/syscall0.S
+++ b/libc/system/syscall0.S
@@ -180,7 +180,7 @@ _syscall_4:
 	xchg %ax,%bx
 	CALL_N_(_syscall)
 	pop %di
-	RET_(2)
+	RET_(8)
 
 _syscall_5:
 	push %di
@@ -193,7 +193,7 @@ _syscall_5:
 	CALL_N_(_syscall)
 	pop %si
 	pop %di
-	RET_(4)
+	RET_(10)
 #else
 # error "unknown calling convention"
 #endif
--- a/libc/system/wait.c
+++ b/libc/system/wait.c
@@ -1,9 +1,8 @@
 #include <sys/types.h>
 #include <sys/wait.h>
-#include <errno.h>
 
 pid_t
 wait(int * status)
 {
-	return wait4(-1, status, 0, (void*)0);
+	return wait4(-1, status, 0, NULL);
 }
--- a/libc/system/wait3.c
+++ b/libc/system/wait3.c
@@ -1,9 +1,10 @@
-#ifdef L_wait3
+#include <sys/types.h>
 #include <sys/wait.h>
 
-int
+#ifdef L_wait3
+pid_t
 wait3(int * status, int opts, struct rusage * usage)
 {
-   return wait4(-1, status, opts, usage);
+   return wait4(-1, status, opts, usage); /* FIXME kernel ignores 'usage' */
 }
 #endif


tkchia (approved):

tkchia:
@ghaerr : that was a rather bad mistake on my part.  Thank you for fixing it!

ghaerr:
Hello @tkchia,

> that was a rather bad mistake on my part. Thank you for fixing it!

I have to confess I'd always assumed the `RET_` macro just emitted a return/pop with the passed byte count, and didn't realize the problem until I finally looked up the definition, then compared this usage with it elsewhere. I can see how this could be easily temporarily forgotten. And I'm glad I spent the time chasing this down, its given me a much better understanding of how the regparmcall calling convention passing works in practice. (I also learned the caller pops any passed arguments, just like stdcall, which further reduces code generated.)

As a result, I've posted this "sticky note" to remind me how everything works when looking through syscall0.S and other generated code:
```
ELKS system calls requires registers set in order of parameters:
    BX, CX, DX, DI, SI


Cdecl calling convention, stack has parms, passed in this order in registers:
    BX, CX, DX, DI, SI
    caller pops all args after call


Regparmcall calling convention, args passed in following registers in order:
    AX, DX, CX, DI, SI
    callee pops args > 3 before return!
```

After having spent considerable time looking at your implementation of `-mregparmcall` for our C library and system calls, I want to say you've done an amazing job of it - I'm particularly impressed with how a few XCHG instructions, and putting the system call number in BX rather than AX, sorts everything out in a bare minimum of code space and execution time. Well done!!

I am extremely impressed with the much smaller amount of code generated for ELKS binaries using regparmcall (subject to the cautions below). I was thinking that perhaps even syscall0.S, which is included in likely every program, could also be trimmed down to only include the code portions for the system call argument counts it actually uses. This may not matter, but we've seen that just a few bytes savings over a disk image of applications can make a difference.

After more testing, we may want to consider making `-mregparmcall` the standard for ELKS applications. My biggest concern will be the following issues, which could pop up should programmers not heed best practices for external function declarations, which it seems regparmcall requires in order to ensure correctness.

During the debugging, I ran into some other issues I have questions about that may be programmer considerations and cautions when using regparmcall compilation:

- When the compiler sees an undefined external function use (or for that matter, an old-style K&R definition with no parameter list), does the compiler assume its cdecl, or will it use regparmcall to call the function? It seems the programmer must be very careful with any undefined external functions when using regparmcall, or the stack may be corrupted on return from calling any of those functions.

- For functions declared variadic, does the compiler tag the function as cdecl, and push all arguments in reverse order, or does it attempt to pass the non-variadicly declared parameters in registers (up to 3)?

Thank you!

tkchia:
Hello @ghaerr,

> I'm particularly impressed with how a few XCHG instructions, and putting the system call number in BX rather than AX, sorts everything out in a bare minimum of code space and execution time. Well done!!

Thanks! :slightly_smiling_face: 

> * When the compiler sees an undefined external function use (or for that matter, an old-style K&R definition with no parameter list), does the compiler assume its cdecl, or will it use regparmcall to call the function? It seems the programmer must be very careful with any undefined external functions when using regparmcall, or the stack may be corrupted on return from calling any of those functions.

If the prevailing calling convention is `regparmcall`, then `gcc-ia16` will continue to assume that the functions use the `regparmcall` calling convention.  Also, functions with K&amp;R declarations are basically assumed to be non-variadic &mdash; i.e. parameters are passed in registers as far as possible.

> * For functions declared variadic, does the compiler tag the function as cdecl, and push all arguments in reverse order, or does it attempt to pass the non-variadicly declared parameters in registers (up to 3)?

The latter: it tries to pass the fixed arguments in registers.  Meanwhile, all the variable arguments are passed on the stack.  Also, any arguments passed on the stack &mdash; whether fixed or variable &mdash; need to be popped by the caller (similar to `cdecl`).

So this pretty much means that most, if not all, variadic functions need to be given proper prototypes, so that the compiler does not produce incorrect code.

Thank you!

ghaerr:
@tkchia: Thanks for the explanations.

> If the prevailing calling convention is 

What is the command line specification for `cdecl`? Also, (we ran into this in the `bc` Makefile): can one re-specify a calling convention, with the last taking precedence? That is, should we convert elkscmd/Make.defs to specify `-mregparmcall`, can a local Makefile override that calling convention? That would be very convenient for local debugging, on certain occasions. The `bc` Makefile specified `-mrtd` and it failed to find the `elks-small.ld` linker script with an error.

Thank you!

tkchia:
Hello @ghaerr,

> What is the command line specification for `cdecl`?

That will be `-mcdecl`, though this is not quite documented in `gcc-ia16` at the moment (I am still not very sure this is a useful thing to have around).

> Also, (we ran into this in the `bc` Makefile): can one re-specify a calling convention, with the last taking precedence?

You can try &mdash; but I do not think (mainline) GCC officially supports this.

> The `bc` Makefile specified `-mrtd` and it failed to find the `elks-small.ld` linker script with an error.

That is kind of strange.

Thank you!

ghaerr:
Hello @tkchia,

> That will be -mcdecl, though this is not quite documented in gcc-ia16 at the moment (I am still not very sure this is a useful thing to have around).

Its usefulness could be to, say, override `CFLBASE += -mregparmcall` in elkcscmd/Make.defs with a local Makefile `-mcdecl`. This could be possibly useful for debugging as CDECL when the default for ELKS applications is REGPARMCALL.

All of this is potentially important should we decide to move to REGPARMCALL as the calling convention for ELKS applications, which I am thinking would be a nice enhancement for v0.7.0 because of the massive amount of disk space saved (75k added extra space on a 1440k floppy!).

> The bc Makefile specified -mrtd and it failed to find the elks-small.ld linker script with an error.
> That is kind of strange.

This only fails with elkscmd/Make.defs specifying `CFLBASE += -mregparmcall` and bc/Makefile specifying `CFLBASE += -mrtd`.

Here is a summary of various calling convention options and their a.out size compiling an arbitrary .c file:
```
ia16-elf-gcc -melks-libc -mtune=i8086 -Os -mcmodel=small foo.c                         # a.out size 2848
ia16-elf-gcc -melks-libc -mtune=i8086 -Os -mcmodel=small -mcdecl foo.c                 # a.out size 2848
ia16-elf-gcc -melks-libc -mtune=i8086 -Os -mcmodel=small -mregparmcal -mcdecl foo.c    # a.out size 2752?
ia16-elf-gcc -melks-libc -mtune=i8086 -Os -mcmodel=small -mreparmcall foo.c            # a.out size 2736
ia16-elf-gcc -melks-libc -mtune=i8086 -Os -mcmodel=small -mcdecl -mreparmcall foo.c    # a.out size 2736
ia16-elf-gcc -melks-libc -mtune=i8086 -Os -mcmodel=small -mrtd foo.c                   # a.out size 2784
ia16-elf-gcc -melks-libc -mtune=i8086 -Os -mcmodel=small -mregparmcall -mrtd foo.c     # error:
  ia16-elf-gcc: error: unable to locate default linker script ‘elks-small.ld’ in the library search paths
```
Without diving too deep, the conclusion seems to be that if Make.defs doesn't specify a calling convention, any can be used to override in a local Makefile, but that if REGPARMCALL were specified in Make.defs, then its is indeterminate what is produced if overriden with CDECL, and an error given if override with STDALL (-mrtd).

BTW, what does `rtd` stand for?

Thank you!

ghaerr:
Hello @tkchia,

Thinking about Make.defs and local Makefiles specifying contradicting calling conventions, I suppose it doesn't really matter whether multiply-specified options might work or not in `ia16-elf-gcc`. An alternative, and probably superior, solution would be to just use `CFLBASE := $(filter-out -mregparmcall $(CFLBASE)) -mcdecl` or something equivalent in order to force a single desired option in a local Makefile.

Thank you!

Issue #1482, opened 2022-12-17
NEC PC-98 enhancements to ELKS


tyama501:
Continued conversation from #1325

tyama501:
Hello @ghaerr ,

It seems that the problem I wrote in https://github.com/jbruchon/elks/issues/1325#issuecomment-1301873959
is not the problem for particular system but rather the serial cross cable is connected or not. 

>I find another problem that I cannot get to "login" prompt if I enabled the serial console or runlevel 3 in inittab for PC-9801RX.
>It gets hang-up before the login.
>Probably the serial driver issue.

It hangs up on PC-9801BX too at the following function if I remove the cable with inittab:3.

serial-pc98.c
static int rs_write(struct tty *tty)
...
    while (tty->outq.len > 0) {
        /* Wait until transmitter hold buffer empty */
        while (!(inb(port->io + UART_LSR) & UART_LSR_THRE))
                ;
        outb((char)tty_outproc(tty), port->io + UART_TX);
        i++;
    }

![PC-9801BX_ELKS_rs_write_IMG_20221217_165116](https://user-images.githubusercontent.com/61556504/208232037-9e268f1f-ddc0-46bb-b52f-063efcdb5d97.jpg)

I need to think a workaround.
Thank you.

tyama501:
One idea is quit using while loop and use for-loop to check empty but the limit may depends on CPU speed.

Is there a good way to timeout checking and abort with couple of seconds?

ghaerr:
Hello @tyama501,

> It hangs up on PC-9801BX too at the following function if I remove the cable with inittab:3.

I am wondering if the PC-98 BIOS or serial chip is somehow holding transmission when the cable is disconnected. You might check any resources on that.

Another idea is to test a different bit in the UART - instead of checking the Transmit Holding Register (UART_LSR_THRE), check the Transmitter Empty bit (UART_LSR_TEMT).

Does your serial cross-cable require that DTR be held high? That is, certain systems that are wired for hardware flow control requires certain pins to be high before the UART will transmit. Cables can "cheat" this by wiring (IIRC) DSR to DTR on the computer side, rather than sending DTR from the receiver. You can read up on this on PC serial cables if not familiar. Currently, ELKS doesn't use HW flow control so the pins should be held high and not sent from the other system.

> One idea is quit using while loop and use for-loop to check empty but the limit may depends on CPU speed.

It seems using a for loop would be exactly the same in this case.

[...]

PR #1497, opened 2022-12-29
[rs] Add timeout to avoid hang-up when serial cable is not connected for PC-98

tyama501:
Hello @ghaerr ,

As we tried on https://github.com/jbruchon/elks/issues/1482#issuecomment-1364638119
I added the timeout.

I also updated .config files to add the followings.
CONFIG_TIMER_INT0F is not set
CONFIG_TIMER_INT1C is not set

Thank you!


ghaerr:
Thanks @tyama501!

PR #1498, opened 2022-12-29
[cmds,build] Various small cleanups

ghaerr:
Fixes `screen` to use 8.3 filenames to allow running on FAT filesystem.

Remove CONFIG_APP_MTOOLS.
Don't assume ES -> kernel data segment in `save_timer_irq`.
Adds commented-out `uname -v` in /etc/profile for optional commit message.


ghaerr:
Hello @ccoffing: I've pushed a fix for your libc regex test, and it passes, thank you!

ccoffing:
> Hello @ccoffing: I've pushed a fix for your libc regex test, and it passes, thank you!

Oh, that's a better solution.  I had toyed with implementing `regerror` in libc itself, but IIRC that somehow caused the grep build to start picking up all these various reg* symbols, and broke the grep compilation.  (I assume it has its own more powerful regex functions, but I haven't even looked yet.)  So it's on the back of my mind that it would be good to proofread the grep build and see why it is fragile.

ghaerr:
Hello @ccoffing,

> that somehow caused the grep build to start picking up all these various reg* symbols, and broke the grep compilation. (I assume it has its own more powerful regex functions, but I haven't even looked yet.) 

Looking at minix1/grep.c, it uses the libc regex functions, and contains its own `regerror`. It seems the design of our regex library is such that the application provides its own implementation of that function, which is nice, since then printf/write etc isn't necessarily dragged in by the library. I hadn't looked at this before doing the same thing in your regex testing code, so I think we're all good!

PR #1499, opened 2022-12-30
[cmds,build] More compilation warning cleanups

ghaerr:


ghaerr:
This brings warnings down quite a bit except for the messes around elkscmd/elvis, elkscmd/levee, some implicit or non-void function declarations, and a bunch of 'use of uninitialized variable' the latter of which may need more thought to properly address.

PR #1501, opened 2022-12-31
test_libc: formatting and fix TODO

ccoffing:


ghaerr:
Thanks @ccoffing. I see you're bothered as much by 4 vs 8 character tabs as I. I haven't got a good solution to this, as the kernel and some applications use both incompatibly. I'm a `vi` user and have `k` and `l` mapped to switch between each to display different files properly. I've come to the conclusion that eliminating tabs entirely may be the best option for the long term... thoughts?

ccoffing:
Personally, I strongly prefer 4 spaces (for multiple reasons).  But what I hate even more than tabs is inconsistency, especially within a single file, because it makes code harder to read.  I only started with tabs because the `coding-style.txt` says to do it for new code, and when in Rome, ...

I use vim, yaifa, and have some mappings to show the formatting, which makes it easier to match existing code styles.

I would happily kill the tabs (and tweak the `coding-style.txt` to document this?)

ccoffing:
> eliminating tabs entirely

:+1:

tyama501:
I also agree to eliminating tabs,
but many files include tabs so I still using for those.

My emacs setting cannot insert tabs so I am copying from other lines...

ghaerr:
> eliminating tabs entirely

Double agree here as well, as some editors will use spaces for 4 char indent, and a tab for 8, causing problems. When tabs are used for left indentation and work for either 4 or 8 char tab settings, we still see miserable lineups for comments on the right side.

I have considered the idea of running the (great) vi `:retab` command on everything, but haven't because of the huge diffs it will cause. So I've been painfully dealing with multiple different tab formats in both the kernel and application code. Ugh!

In general I've been slowly reformatting the kernel and some application code near what I am working on to only use tabs, and always have `:set expandtabs` on. This still causes eyesores when the rest of the file is using 8 character tabs and my changes are stuck at 4. Others with an editor set to 8 will find hard-to-read source. Ugh again!

> Personally, I strongly prefer 4 spaces (for multiple reasons). 

I generally agree with that, but only when spaces are used instead of TAB characters in the file itself. Otherwise, more mess similar to the above.

> I only started with tabs because the coding-style.txt says to do it for new code, and when in Rome, ...

Perhaps it should get modified to read that it the preferred format is 4 character indents for tabs, but always have tabs expanded to spaces (e.g. `:set expandtabs`) when used, except when the file is already formatted otherwise. I worry a bit about other developers that may not have their editor setup to accommodate that, and also don't want to litter the source with vi instructions at the top or bottom of each file.

ghaerr:
> My emacs setting cannot insert tabs so I am copying from other lines...

Geez, I'm doing the same thing to keep tab expansion working in files formatted with all-4 or all-8 tabs instead of spaces.

Perhaps we should create a new rule where any time C function is modified, the entire function (not file) is updated to 4-character indent, no tabs.

Mellvik:
Interesting discussion.

FWIW - I'm all for TABS (and [surprise] I'm vi only] and there are all kinds of tools out there to retab source files to whatever is the preference, so it's hard to see it should be a big deal.

BTW I love small files and I realize the space saving argument went moot 20+ years ago.


-M

PR #1503, opened 2022-12-31
Fix memory leak in cron

ccoffing:


ghaerr:
How did you find this, just by reading code?

Issue #1500, opened 2022-12-30
Boot hangs as result of NE2K driver probe fail with no NIC


toncho11:
![IMG_20221230_185501](https://user-images.githubusercontent.com/1851997/210099463-66e89052-1a2b-473c-af1e-a79e42ac02dc.jpg)

This is on IBM 5160, 64 kb motherboard with 640 kb ram installed booting mimix version of ELKS.

toncho11:
The same works on Pravetz 16 and IBM 5160 256KB version ...

ghaerr:
@toncho11, does this fail in the same place every time, or slightly different?

Also, do you normally compile in support for the network drivers? Just trying to see what might be different that is causing this.

I did make one (what I thought was small) change to the kernel in https://github.com/jbruchon/elks/commit/61fb3b3dda7dd9aa71631145311d6fea7a28ca29 which was tested on QEMU but not real hardware. Perhaps we should roll back prior to that to see if things work.

toncho11:
I always fails on the same place on the 64 kb version of 5160. 
It failed 3 times in a row.
It failed at least one time on the 256 kb version of 5160, but it works in general.

toncho11:
No network card is installed.

ghaerr:
Can you try removing the network drivers by unsetting all `CONFIG_ETH_*`? That will determine if the problem is in the network driver identification code, or possibly a timer interrupt issue that I was working on with @Vutshi.

ghaerr:
I also disabled `disable_timer_tick` in elks/arch/i86/kernel/timer-8254.c, which could be the issue. Remove the `#if NOTNEEDED` around this code to test. Thank you!
```
void disable_timer_tick(void)
 {
 #if NOTNEEDED
     outb (TIMER_MODE0, TIMER_CMDS_PORT);
     outb (0, TIMER_DATA_PORT);
     outb (0, TIMER_DATA_PORT);
 #endif
 }
```

ghaerr:
If you have not tested the kernel boot since Dec 19, that's when @Mellvik last updated the network drivers, and if the kernel is hanging in the same place, it could be in NIC hardware identification code. (Or it could just happen to be a timer interrupt occurs at very nearly the same point, which has also been worked on. That's why it is important to know whether it dies in exactly the same point displayed on the screen or not every time).

ghaerr:
[...]

PR #1508, opened 2023-01-02
[nek2] Add 8390 chip check to probe

ghaerr:
Hello @toncho11, @Mellvik,

I'm posting this to show potential solution to long argument(s) in #1500. Tested on QEMU. Needs testing on both IBM 5160 with no NIC installed, but with both IDE controller installed and absent. Needs testing with real NE2K hardware.

Addresses IBM 5160 incompatibility found in #1500.

NE2K ISA probe routine moved from ASM to C. Adds 8390 chip detect because IBM 5160 fails ISA probe.

New probe routine = 51 bytes. Removed old probe routine = 20 bytes. Net increase in kernel size: 31 bytes. The probe restores the value at `net_port` in the case NIC not present but possibly IDE present. Does not attempt restore of `net_port+0x0d` since IDE controllers only use 8 port addresses.


toncho11:
Thank you @ghaerr!
I could not wait ... (even with my terrible headache) and I tested the 1440 image:

- with no XT-IDE attached: works and n2k is not detected (tested 2 times)
- with XT-IDE attached: works, but a card is detected. It says MAC: ab : ab : ab ...

It does not halt in both cases and ELKS loads until command prompt.

ghaerr:
@toncho11, thanks for the fast testing. Its pretty good news since it fixes your problem, but I was hoping that the 8390 chip identification was better at differentiating from an IDE controller that is sometimes at the same address in older systems. 

Nonetheless, as @Mellvik points out, if the wrong cards are present at configured addresses, the system will have to be reconfigured. The good news is that could then be done by editing `/bootopts` on your booted system, and that the 8390 chip detect doesn't apparently render the BIOS IDE code non-working.

toncho11:
Yes. Thank you!
Actually I need to check if ELKS can access correctly the XT-IDE now.

Maybe not so important, but the N2k driver could check the MAC address for sanity (actually there is some code for that already in ne2k_drv_init) and disable the driver.

```
#if DELETEME
		/* Should not happen since we already have probed the NIC successfully */
		/* May be superflous - candidate for removal */
		if (((prom[0]&0xff) == 0xff) && ((prom[1]&0xff) == 0xff)) {
			printk("%s: No MAC address\n", dev_name);
			err = -1;
			break;
		} 
#endif
```

We could add to check for "ab" as well?

ghaerr:
Hello @toncho11,

> Actually I need to check if ELKS can access correctly the XT-IDE now.

Yes, please do. I assumed that was working since you got to the shell prompt. If the XT-IDE access fails, this check will have to be removed or reprogrammed, as it would be causing another error while trying to fix the first one.

>  but the N2k driver could check the MAC address for sanity 
> We could add to check for "ab" as well?

No, as @Mellvik originally suggested, we don't need to go on a wild goose chase of trying fix things that aren't necessarily broken, and there's a kernel and further testing cost for all of it. The particular contents of your PC at the MAC address could be different for different PCs, etc. I'm not particularly satisfied that this mod can't differentiate between an NE2K and an XT-IDE, but the big issue, which is that ELKS flat hangs on boot, is fixed, provided this fix also works on real NE2K hardware. Also as @Mellvik mentions, fine-tuning an ELKS image to one's hardware should still be done. What this ultimately means is that, although the standard image will run on your IBM XT PC, that same system should ideally be recompiled without the network drivers or turned off via the options available in /bootopts.

Thank you!

toncho11:
I agree :)
I just tested with the 360 KB image because my floppy 1.44 controller seems not to work with XT-IDE. 
Everything is as expected and it works. 

Thank you again!

Mellvik:
@ghaerr,I rest my case. This is not a solution - for a number of reasons, most of which have been voiced already. Adds bloat, fixes nothing, hides the real problem, waste of time.-M2. jan. 2023 kl. 21:49 skrev Gregory Haerr ***@***.***>:﻿Hello @toncho11, @Mellvik,
I'm posting this to show potential solution to long argument(s) in #1500. Tested on QEMU. Needs testing on both IBM 5160 with no NIC installed, but with both IDE controller installed and absent. Needs testing with real NE2K hardware.
Addresses IBM 5160 incompatibility found in #1500.
NE2K ISA probe routine moved from ASM to C. Adds 8390 chip detect because IBM 5160 fails ISA probe.
New probe routine = 51 bytes. Removed old probe routine = 20 bytes. Net increase in kernel size: 31 bytes. The probe restores the value at net_port in the case NIC not present but possibly IDE present. Does not attempt restore of net_port+0x0d since IDE controllers only use 8 port addresses.



You can view, comment on, or merge this pull request online at:
  https://github.com/jbruchon/elks/pull/1508

Commit Summary

  3b99f12  [nek2] Add 8390 chip check to probe


File Changes  (2 files)

  
    M
    elks/arch/i86/drivers/net/ne2k-asm.S
    (2)
  
  
    M
    elks/arch/i86/drivers/net/ne2k.c
    (27)
  


Patch Links:

  https://github.com/jbruchon/elks/pull/1508.patch
  https://github.com/jbruchon/elks/pull/1508.diff

toncho11:
What an inspirational comment @Mellvik !!!

ghaerr:
Hello @toncho11,

I looked at the original NE2K probe routine and see that it doesn't actually perform a raw, read-only ISA probe - instead, it sends a command to the NIC address and then reads the result. I would like to see what an actual raw ISA probe returns, and have pushed another update that will display this at boot. Can you test this on your machine and take a screenshot of the result (with and without your IDE controller)? 

Thank you!

toncho11:
With xt ide attached:
IMG_20230103_193455

toncho11:
No xt ide attached:
IMG_20230103_193829

ghaerr:
Hello @toncho11,

[EDIT: Comment rewritten entirely - I accidentally swapped the above screenshots with regards to XT IDE installed vs not].

Thank you for the testing! 

It seems that an actual raw ISA probe on your machine for a non-existent device returns garbage, as evidenced by the first read returning 88, and reading directly again returning 30. Subsequent reads all returned 30 after an 0x61 command is sent, which also answers why the original NE2K probe fails, as it doesn't attempt a chip identification, and accepts a NIC present as any read value other than 00 or FF. I suspect the reason for this could be non-terminated data lines, or possibly the results of another installed card, as when the XT-IDE is installed, the port reads always return 00.

The 8390 chip detect rejects the (non-existent) NIC, which prevents a hang later in the NIC driver code. Given that your machine appears to return a random value on reading the netport in both raw and the original probe, I think we will need to perform a chip detect, which is working, in order to reject a non-existent NIC.

For the case with the XT-IDE card plugged in, it appears that the controller read always returns 00, which also passes the 8390 chip detect. For this, I propose using the original ISA probe code, which writes a command and rejects both FF and 00 returns, before performing a chip detect.

I will prepare another push with debugging output for you to test.

Thank you!

ghaerr:
@toncho11: Reading a raw probe won't work for this, as testing on QEMU shows it apparently doesn't setup a NIC device until one of its registers is written to. Adding that would fix your system, but break QEMU network emulation. [EDIT: this turns out to not be an issue, as a raw probe won't work for your system either, see above].

ghaerr:
@toncho11,

I am sorry - I completely misread your two posted screenshots - I was reading the XT IDE installed as not and vice versa. This means that everything I concluded in the last two posts after your screenshots is incorrect. I am going to edit those posts, and advise a path forward that will work. Sorry for the confusion!

ghaerr:
Hello @toncho11,

[EDIT: Comment rewritten entirely - I accidentally swapped the above screenshots with regards to XT IDE installed vs not].

Thank you for the testing!

It seems that an actual raw ISA probe on your machine for a non-existent device returns garbage, as evidenced by the first read returning 88, and reading directly again returning 30. Subsequent reads all returned 30 after an 0x61 command is sent, which also answers why the original NE2K probe fails, as it doesn't attempt a chip identification, and accepts a NIC present as any read value other than 00 or FF. I suspect the reason for this could be non-terminated data lines, or possibly the results of another installed card, as when the XT-IDE is installed, the port reads always return 00.

The 8390 chip detect rejects the (non-existent) NIC, which prevents a hang later in the NIC driver code. Given that your machine appears to return a random value on reading the netport in both raw and the original probe, I think we will need to perform a chip detect, which is working, in order to reject a non-existent NIC.

For the case with the XT-IDE card plugged in, it appears that the controller read always returns 00, which also passes the 8390 chip detect. For this, I propose using the original ISA probe code, which writes a command and rejects both FF and 00 returns, before performing a chip detect.

I will prepare another push with debugging output for you to test.

Thank you!

ghaerr:
@toncho11: 

I have pushed (hopefully the last for you to test) commit. This version should reject the NE2K hang when no NIC is installed, (which worked in your last test), but also reject the NE2K positive identification (and no hang) when your XT-IDE controller is installed. There is still debug code included for the values read from the ISA bus included, to both confirm this fixes both cases, or shows us more info for debugging.

Thank you!

toncho11:
With XT IDE attached attached attached attached
IMG_20230106_130510

toncho11:
With NO xt ide attached
IMG_20230106_130944

toncho11:
This is on Amstrad 1640 with no xt ide attached

IMG_20230106_131534

toncho11:
This is on Amstrad 1640 with Xt ide attached:
IMG_20230106_131955

toncho11:
On 486 with hdd and 3com card:
IMG_20230106_132719

toncho11:
I think it could not be better :)
Thank you @ghaerr !

ghaerr:
@toncho11, thank you for the testing. I will write up a summary in #1500 to keep the details in once place.

I notice that on the IBM XT 5160 with no XT IDE attached, the initial ISA probe response (reading from the same port twice resulting in 88 then 30), appears that it's seemingly not random, as the results are the same as your test a few days ago. What other cards are plugged in to the ISA bus on that machine, any?

Also, did the "bioshd(0): track read retry \#1" message at the bottom of the boot screen just start appearing with the latest commit, or have you seen that before?

I am wondering if there is another device, possibly even the floppy controller, that is responding from I/O address 0x300 on your system.

toncho11:
I think eveything is OK.
It is currently tested on 3 different machines.

On the original IBM 5160 (that I have been testing) there is a VGA card, old floppy controller and RAM extension.
I am putting extra effort not to use my (newely built) 1.44 floppy controller on this IBM for these tests.
For me the bioshd(0) is related to the fact that I am using the same 360kb disk all the time. It is an old disk, some problems reading it should not be surprising.

toncho11:
Yet, I confirm that it is not random the (80) (30)

ghaerr:
Do you happen to have a card inserted into Slot 8? [This article](https://www.minuszerodegrees.net/5160/misc/5160_slot_8.htm) talks about some problems with using slot 8, and references it possible use for I/O ports in the "system range" from 0x00 to 0xFF. If there is a card in there, this would explain possibly that card responding at 0x300 instead, which would explain what we're seeing here.

> For me the bioshd(0) is related to the fact that I am using the same 360kb disk all the time. It is an old disk, some problems reading it should not be surprising.

Agreed. I was wondering about how often this occurs, since the same 360kb disk isn't producing the same errors on other machines. If it only is coming recently from the 5160, then that could be because my latest commit doesn't restore a register (for the first time) and the result is the system boots, but always with this error. I could turn that restore back on, and am trying to determine whether to do so or not.

toncho11:
Slot 8 is J8. There is nothing there. I explicitly do not use it.
I just used another 360 kb diskette. 
Again 80 30 30. 
No bioshd(0) track error.

ghaerr:
> Slot 8 is J8. There is nothing there. I explicitly do not use it.

OK, good to know. That then leaves all the information I can find out of the equation, concluding that the XT bus does not return FF on raw ISA probes with no device/card at the I/O address, as later AT systems do. 

> I just used another 360 kb diskette.
> No bioshd(0) track error.t

Great. We don't need to restore values, looks like a marginal diskette as you thought.

I'll write up a conclusion and get a modified version of this without debugging display committed.

toncho11:
The Amstrad 1640 is an XT and it does return FF (from the screenshot). But then the Amstrad is a later version of XT (with faster CPU and built in floppy controller and probably improved BIOS) while well the IBM is the first one.

PR #1511, opened 2023-01-04
[kernel] Introduce precision kernel looping timeouts

ghaerr:
Allows for easy coding of precision kernel timeouts, rather than continuously busy looping or decrementing a variable.

Adds 30ms timeout for IDE Query function, rather than busy looping forever if IDE device not responding (found using blink emulator).

### DIFF
--- a/elks/arch/i86/drivers/block/idequery.c
+++ b/elks/arch/i86/drivers/block/idequery.c
@@ -15,13 +15,13 @@
 #define IDE_DRIVE_ID	0xec	/* IDE command to get access to drive data */
 #define IDE_STATUS	7	/* get drive status */
 #define IDE_ERROR	1
+#define WAIT_READY      (3*HZ/100) /* 3/100 sec = 30msec - should be instantaneous */
 
 #define IDE_DEBUG	0	/* IDE probe debugging */
 
 /* For IDE/ATA access */
 #define STATUS(port)	inb_p(port + IDE_STATUS)
 #define ERROR(port)	inb_p(port + IDE_ERROR)
-#define WAITING(port)	(STATUS(port) & 0x80) == 0x80
 
 #if IDE_DEBUG
 #define debug   printk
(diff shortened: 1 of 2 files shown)


ghaerr:
Hello @tyama501,

Using this method, you should be able to recode the PC-98 serial timeouts to wait a more precise amount of time in the serial driver when attempting output with CTS not ready. I have tested this with the IBM PC serial driver (although this is not needed there), which uses the following code to wait up to 40ms for the UART transmitter ready buffer. 33.3 ms is required for each character transmitted at 300 baud.
```
    while (tty->outq.len > 0) {
        unsigned long timeout = jiffies + 4*HZ/100;     /* 40ms, 300 baud needs 33.3ms */
        /* Wait until transmitter hold buffer empty */
        while (!(INB(port->io + UART_LSR) & UART_LSR_THRE)) {
            if (time_after(jiffies, timeout))    /* waits 40ms max, jiffies updated by hw timer */
                break;
        }
        outb((char)tty_outproc(tty), port->io + UART_TX);
        i++;
    }
```
I suggest using this to overcome your timing problems on different PC-98s while still trying to boot with CTS not connected.

Thank you!

tyama501:
Wow, thank you @ghaerr ,

I will try that.
This can coexist with the keyboard polling timer, right?

ghaerr:
> This can coexist with the keyboard polling timer, right?

Yes, although the keyboard polling might be delayed for the duration of the entire `write` to the serial port. This is very nearly the same as before, which is that the kernel will busy loop, doing nothing else, until the (more precise) timer expires.

Any hardware interrupts, including the timer interrupt, will execute, but the kernel won't switch to another task or schedule any kernel timer callbacks. The keyboard polling timer is setup using a kernel timer, which is executed during `sched()` - and that routine is only allowed to be called when the kernel is not already executing.

This will be interesting to test. Should we feel the need to process kernel timer callbacks during a serial port busy loop, we could replace this new code with a more complex mechanism that sleeps the task for 40 ms, then outputs (discards in the PC-98 case) the serial character anyways. I hadn't thought about the PC-98 requiring the polling keyboard, so that may be an even better longer term solution. This solution at least allows for precise timing on all PC-98s regardless of CPU clock speed.

PR #1513, opened 2023-01-07
[libc] Rewrite libc optional floating support using weak sym linking

ghaerr:
Uses linker weak symbols and `_weaken` and `__YOINK` macros to support C library optional floating point support, as discussed in https://github.com/jbruchon/elks/pull/1469#issuecomment-1374255470.


ghaerr:
Hello @tkchia,

This approach seems to work well, except for in medium model, where the `_weaken`  macro returns a 16-bit value instead of a 32-bit value for the address of a possibly linked in symbol. Can you suggest a change to `_weaken` that for medium model only will return a 32-bit CS:OFFSET value (or 0 when not linked in)? I am not familiar enough with gcc's magic `__asm` output register specifications to make this work. Of course, returning CS:0 when not linked in would not work either.

```
In file included from /Users/greg/net/elks-gh/libc/stdio/_stdio.h:5:0,
                 from /Users/greg/net/elks-gh/libc/stdio/vfprintf.c:28:
/Users/greg/net/elks-gh/libc/stdio/vfprintf.c: In function ‘vfprintf’:
/Users/greg/net/elks-gh/libc/include/asm/yoink.h:25:32: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
 #define _weaken(symbol)       ((const typeof(&(symbol)))_weakaddr(#symbol))
                                ^
/Users/greg/net/elks-gh/libc/stdio/vfprintf.c:251:10: note: in expansion of macro ‘_weake’
      if (_weaken(dtostr))
          ^~~~~~~
```

I'm also realizing that medium model may be the case where your updated linker will be required, since that probably emits the changed R_386_SEG16 handling you spoke of in https://github.com/jbruchon/elks/pull/1469#issuecomment-1362152372, which I didn't see during small model testing.

I have produced this PR mostly for testing; it should probably be held from commit until the toolchain is updated in tools/Makefile.

Thank you!

ghaerr:
For a medium model `_weaken`, I am thinking perhaps something that emitted code like this could work?

```
if (_weaken(sym)) ...

    movw    $sym@OZSEG16,%dx
    movw    $sym,%ax

The compiler would then generate code for calling the function using something like:
(_weaken(sym)()...

    push    %bp
    mov     %bp,%sp
    push    %dx
    push    %ax
    lcall   *-4(%bp)
    add     $4,%sp
```

ghaerr:
This seems to produce pretty good code for `_weaken` in medium model:
```
#define _weakaddr(symbolstr) __extension__ ({   \
    unsigned int a, d;                          \
    asm(".weak\t"  symbolstr "\n"               \
        "\tmov\t$" symbolstr ",%%ax\n"          \
        "\tmov\t$" symbolstr "@OZSEG16,%%dx\n"  \
            : "=a" (a), "=d" (d)                \
            /* no input */ );                   \
    ((unsigned long)d << 16) | a;               \
    })
```
Using a separate`_weaken` macro for medium vs small model would work, assuming that all functions are compiled far or near, respectively. I'm not sure how to approach designing a macro that could check whether a passed function was near or far, and use the appropriate macro.

tkchia:
Hello @ghaerr,

I think you can just say something like
```
#define _weaken(sym)   ({ \
                          extern __typeof__(sym) __weak_sym \
                             __asm(__WEAKEN_STR(sym)) \
                             __attribute__((__weak__)); \
                          __weak_sym; \
                       })
```
This basically creates a new external symbol with the same assembler name and type as the symbol, except it is weak.

Thank you!

ghaerr:
Hello @tkchia,

> This basically creates a new external symbol with the same assembler name and type as the symbol, except it is weak.

Wow, that's a much better solution, thank you! Not having to load the symbol into registers makes the resulting code produced quite small. 

Its too bad there's no way to see weak vs normal symbols in the disassembly output from `ia16-elf-objdump`, but that's ok. 

I've tested your suggestion and it works very well, so I'll update this PR and ask for your final review. I'm also thinking of changing the name `__YOINK` to something more descriptive for ELKS, such as `__FORCE_LINK` or `__LINK_SYMBOL`.

Thank you!

ghaerr:
Hello @tkchia,

Using your suggested definition of `_weaken`, in medium mode, the ASM output produces a symbol modified by `@OZSEG16` similar to the following:
```
    mov $sym@OZSEG16,%ax
```
I'm not familiar with exactly what `ia16-elf-as` actually does for such a symbol, does it just produce the highword (segment) address of `sym` with an R_386_SEG16 relocation, or something else?

Also, does this mechanism require your updated ld.gold fix [don't emit R_386_SEGRELATIVE for weak undef. sym.](https://github.com/tkchia/binutils-ia16/commit/cc5f1a9d7437051c3aae6ec16fca27c3b2d31f6c)? I have tested `_weaken` with small model which works with our exiting toolchain, but have not tried or tested the medium model linked output, just visually inspected a small .s file output. If required, I will bump up the IA-16 binutils with this commit.

Thank you!

tkchia (approved):

tkchia:
Hello @ghaerr,

> ```
>     mov $sym@OZSEG16,%ax
> ```
> 
> I'm not familiar with exactly what `ia16-elf-as` actually does for such a symbol, does it just produce the highword (segment) address of `sym` with an R_386_SEG16 relocation, or something else?

Did you compile the C module with `-melks`?  The correct relocation type in the assembly source should be ...`@SEG`, which will produce a `R_386_SEG16` relocation.

> Also, does this mechanism require your updated ld.gold fix [don't emit R_386_SEGRELATIVE for weak undef. sym.](https://github.com/tkchia/binutils-ia16/commit/cc5f1a9d7437051c3aae6ec16fca27c3b2d31f6c)? I have tested `_weaken` with small model which works with our exiting toolchain, but have not tried or tested the medium model linked output, just visually inspected a small .s file output. If required, I will bump up the IA-16 binutils with this commit.

The patch fixes a bug which happens, if a module tries to take the segment portion (i.e. `R_386_SEG16`) of a weak symbol, and the symbol is _not_ defined.  In this case, the correct behaviour will be to yield an absolute value of 0 (segment:offset = absolute 0:0), without emitting a segment relocation in the output `a.out`.  The buggy linker wrongly emitted a relocation in this case.

Thank you!

tkchia:
Hello @ghaerr,

> Its too bad there's no way to see weak vs normal symbols in the disassembly output from `ia16-elf-objdump`, but that's ok.

The `-t` option to `ia16-elf-objdump` will (separately) output a symbol table &mdash; this will say if any symbols are imported (or exported) as weak symbols.

Thank you!

PR #1516, opened 2023-01-15
[cmds] Fix off-by-one error in xargs for long input lines

ccoffing:


ghaerr:
Thank you @ccoffing, nice find!

PR #1517, opened 2023-01-15
[libc] Add --ftrace option to all commands for function tracing

ghaerr:
Demonstration of some new instrumentation and debugging capabilities recently added to ELKS. 

When enabled, this automatically adds an `--ftrace` command line option to all ELKS programs which displays each function call made with nesting level, and the current stack and max stack used. Currently only displays function address (symbolic display coming shortly).

This option is enabled and can be compiled in to the entire ELKS elkscmd/ build by uncommenting the `CFLBASE += -finstrument-functions-simple` line in elkscmd/Make.defs. It also requires running the latest toolchain, which can be built by running `tools/build.sh`.

The process works by using the new `gcc-ia16` constructor function capability, along with instrumentation calls compiled into every function by the new toolchain. When `-finstrument-functions-simple` is added to the compiler command line, the linker will automatically include from the C library a default version of a constructor routine that is run before `main` that checks the argc/argv command line for `--ftrace` as the first argument, and turns on function tracing while rewriting the remaining argc/argv array if found. The default instrumentation just displays the function address being called and the stack depth. This is immediately useful to determine the maximum stack an application requires, as well get a feel for the nested procedure call depth.

Of course, this capability will be much more useful when function names are taken from a symbol table included in the application binary, which will be the next step.

### DIFF
--- a/elkscmd/basic/Makefile
+++ b/elkscmd/basic/Makefile
@@ -18,6 +18,7 @@ HOST_CFLAGS = -O3
 PRGS = basic
 OBJS = basic.o host.o
 #MAPFILE = -Wl,-Map=basic.map
+#CFLAGS += -finstrument-functions-simple
 
 ifeq ($(CONFIG_ARCH_IBMPC), y)
 OBJS += host-stubs.o
--- a/elkscmd/debug/Makefile
+++ b/elkscmd/debug/Makefile
@@ -14,7 +14,8 @@ HOSTCFLAGS = -O3
 ###############################################################################
 
 PRGS = testsym disasm opcodes
-LIBOBJS += syms.o stacktrace.o printreg.o
+LIBOBJS += instrument.o syms.o
+LIBOBJS += stacktrace.o printreg.o
 
 # disable tail call optimization for better stack traces
 CFLAGS += -fno-optimize-sibling-calls
--- a/elkscmd/debug/instrument.c
+++ b/elkscmd/debug/instrument.c
@@ -1,0 +1,106 @@
+/*
+ * ELKS instrumentation functions for ia16-elf-gcc
+ *
+ * June 2022 Greg Haerr
+ */
+#include <stdio.h>
+#include <string.h>
+#include "stacktrace.h"
+#include "syms.h"
+
+/* turn on for microcycle (CPU cycle/1000) timing info */
+#define HAS_RDTSC       1   /* has RDTSC instruction: requires 386+ CPU */
+
+static char ftrace;
+static int count;
+static unsigned int start_sp;
+static unsigned int max_stack;
+
+/* runs before main and rewrites argc/argv on stack if --ftrace found */
+__attribute__((no_instrument_function,cdecl,constructor(101)))
+static void checkargs(volatile int ac, char **av)
+{
+    char **avp = av + 1;
+
+    if (*avp && !strcmp(*avp, "--ftrace")) {
+        while (*avp) {
+            *avp = *(avp + 1);
+            avp++;
+        }
+        ftrace = 1;
+        ac--;
+    }
+    get_micro_count();      /* init timer base */
+}
+
+/* every function this function calls must also be noinstrument!! */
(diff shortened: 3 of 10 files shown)


tkchia:
Hello @ghaerr,

How exactly are you passing the `argc` and `argv[]` in and out of the constructor function?  Perhaps these are "accidentally" placed on the stack at just the right place for now, but I am not sure the code should depend on such an implementation detail.  (Personally, I also feel a bit jittery about the idea of the C library frobbing the command line arguments without the program's knowledge.)

If you would like to implement a `--ftrace` option, I guess you can merge this PR first, then I will look into how to pass the `argc` and `argv[]` in an out of constructor routines in a more well-defined fashion (e.g. Apple-style `_NSGetArgv ()`).

I would prefer though to implement the tracing setting(s) as an environment variable.  That is, something like `glibc` 's `MALLOC_CHECK_` variable (see `mallopt`(3)).  This should work well:
* `environ[]` would have been initialized by the time the constructor routines kick in.
* And, extra variables in the environment are generally harmless.

Thank you!

ghaerr:
Hello @tkchia,

Thank you for your review.

> How exactly are you passing the argc and argv[] in and out of the constructor function? 

At this point, the arguments are passed in by the (newly) documented push of argc, argv and envp in crt0.S, and the modified argc is passed out by using `volatile int ac` in the constructor function, so as to keep the emitted `decw 6(%bp)` from being optimized out.

> I am not sure the code should depend on such an implementation detail.
> how to pass the argc and argv[] in an out of constructor routines in a more well-defined fashion (e.g. Apple-style _NSGetArgv ()

Ok. Given our recent work trying to keep the crt0.S startup code very small, I thought this a pretty nice albeit a bit tricky solution, since crt0.S and constructors are already very much dependent.  But, the real argc and argv are stored in unnamed stack variables (in contrast to envp which is stored in `environ`), so I see your point - especially since 2/3 of the startup arguments are treated one way, and 1/3 another.

Looking at the Apple `_NSGetArgv()` implementation, it's just a procedure call wrap around static variables for argc, argv, and envp. Looking at Cosmopolitan, a simpler implementation just uses `extern int __argc; extern char **__argv;`, which could be used to name the variables in just the same manner as envp and the existing `extern char **environ`. Given we're already processing envp into `environ` using a .preinit section, I think the same method could work well for argc/argv and keep things quite small.

> (Personally, I also feel a bit jittery about the idea of the C library frobbing the command line arguments without the program's knowledge.)

This idea was taken from Cosmopolitan, where the same technique is used for `--ftrace` and `--strace`. Given that ELKS desperately needs better debugging capabilities, I think the benefit far exceeds its potential downside. I have learned from the Cosmopolitan author how nice it is to be able to rapidly understand the reason(s) for failing programs which allow developers to easily turn on debugging without recompilation. (No, I'm not recommending this be baked into ELKS permanently because of our size constraints, but wow, what a convenient way to better understand a program's execution!)

>  I guess you can merge this PR first

Ok. I will then take a pass at removing the stack variables and keeping them in named globals as described above in a `.preinit` section, for your approval. I would like to keep the startup code as small as possible, per your previous work.

> I would prefer though to implement the tracing setting(s) as an environment variable. 

Good idea, I will add that to this PR as it provides a nice way to turn on debugging for an entire set of applications at once. I would like to keep the `--ftrace` option as well, for quick individual debugging.

> That is, something like glibc 's MALLOC_CHECK_ variable (see mallopt(3)).

Yes, something like that, combined with our now operational weak symbol binding for much stronger memory allocation checking, along with the ability to replace the allocation routine(s) entirely are on my list :)

Thank you!

tkchia:
Hello @ghaerr,

> > (Personally, I also feel a bit jittery about the idea of the C library frobbing the command line arguments without the program's knowledge.)

> This idea was taken from Cosmopolitan, where the same technique is used for `--ftrace` and `--strace`.

Indeed I am aware that Cosmopolitan does this frobbing of command line arguments... but admittedly I still feel a bit jittery about doing so. :eyes: 

Thank you!

PR #1518, opened 2023-01-16
[libc] Move argc/argv from stack to __argc/__argc for constructors

ghaerr:
Hello @tkchia,

Here's a pass at providing a more well-defined method of accessing and/or modifying argc/argv in constructors, as discussed in https://github.com/jbruchon/elks/pull/1517#issuecomment-1384180489.

Since `__argc` is not saved in a stack location so as to allow direct modification by constructor(s), I couldn't see a way to only drag in the saving/restoring of `__argc/__argv` into a `.preinit` section that would be included only when referenced by a constructor or another function.

Thus, this version increases the size of every executable by 16 bytes. Is it worth it? Frankly, I'm not sure, although direct access to `__argc` and `__argv` from other functions outside of `main` is a benefit.  Let me know if you have any better ideas to reduce the size.

Thank you!


tkchia (approved):

tkchia:
Hello @ghaerr,

> Let me know if you have any better ideas to reduce the size.

I am thinking we might be able to do something like this:
```
	push %ax
	push %bx
	push %cx
	// ... .preinit ...
	popw __argc  // ⎤ .preinit
	popw __argv  // ⎦
	// ... more .preinit ...
	// ... .init ...
	pushw __argv  // ⎤ .postinit ?
	pushw __argc  // ⎦
```
where the `popw` 's only happen if a program actually refers to `__argc` and/or `__argv`.  This will introduce a bit of a trade-off:
* binaries will be as small as before in the "normal" case where there is no instrumentation code
* but if instrumentation is added, then the code will be slightly longer than it needs to be (`push` followed by `pop`, rather than a simple `mov`).

If we can further control the placement of the `__argc` and `__argv` variables relative to each other, then we could probably also further shrink the code itself for loading and storing `__argc` and `__argv`.

Thank you!

ghaerr:
Hello @tkchia,

> I am thinking we might be able to do something like this: where the popw 's only happen if a program actually refers to __argc and/or __argv.

That's a really great idea, thank you! And very nice that it keeps the non-instrumented files as small as they were. I don't think it matters much at all that the size is a few bytes larger when instrumentation is included, given that's currently adding ~1500 bytes (plus stdio if not already present).

Since this PR is already open, I'll take a pass at implementing your design, since it gives me more experience dealing with your ingenious linker section management. It seems the solution will require a new `.postinit` section, which I will add to the linker scripts just after `.preinit` and prior to `.inittail`.

> If we can further control the placement of the __argc and __argv variables relative to each other, then we could probably also further shrink the code itself for loading and storing __argc and __argv.

It would be interesting to know what you're thinking - for now I'll code up your first suggested idea so I can keep working on moving the debug/symbol table stuff forward.

Thank you very much!

ghaerr:
Thanks for your help on this @tkchia, it's all done and working great :)

PR #1523, opened 2023-01-18
Fix ash crash checking TERM from linenoise

ccoffing:
When running ash natively on x64 Linux, I can trigger a crash in the `bltinlookup` function due to an overwritten pointer.  16-bit code may or may not crash, but it would be dereferencing freed memory and therefore behave non-deterministically.

Commit 8da6aec933dc1b8e071364b11661c04141f4a85b included this change:
```
 static int isUnsupportedTerm(void) {
-    char *term = getenv("TERM");
+    //char *term = getenv("TERM");
+    char *term = bltinlookup("TERM", 1);    /* getenv won't work here for ELKS*/
     int j;
```

I read "won't work" as "check the shell's current environment, not the one it was started from".  Makes sense, except it appears that `bltinlookup` is only valid to be called in particular contexts.

These environment variables are stored on a stack-like data structure, and by using a hardware watchpoint, I can see that shortly before the crash, this memory location is re-used by the "stack" for some other purpose.  I only loosely understand the code at this point, but I suspect that the allocations are popped off the "stack" during an error, and therefore `bltinlookup()` is not guaranteed valid in all contexts.  For comparison, the rest of ash only calls it from within a builtin command.

And in fact, this theory seems even more likely, since I can only reproduce the crash when I dereference a variable in an error context.

```
$ A=1
$ ecoh $A
ecoh: not found

Program received signal SIGSEGV, Segmentation fault.
varequal (p=p@entry=0x1 <error: Cannot access memory at address 0x1>, q=q@entry=0x55555556825c "TERM") at var.c:649
649		while (*p == *q++) {
(gdb) bt
#0  varequal (p=p@entry=0x1 <error: Cannot access memory at address 0x1>, q=q@entry=0x55555556825c "TERM") at var.c:649
#1  0x0000555555563006 in bltinlookup (name=name@entry=0x55555556825c "TERM", doall=doall@entry=1) at var.c:312
#2  0x000055555556400b in isUnsupportedTerm () at linenoise_elks.c:261
#3  linenoise (prompt=0x5555555681b5 "$ ") at linenoise_elks.c:1218
#4  0x000055555555bf78 in preadbuffer () at input.c:225
#5  0x0000555555560ebc in xxreadtoken () at parser.c:697
#6  0x00005555555610a4 in readtoken () at parser.c:625
#7  0x0000555555561379 in parsecmd (interact=interact@entry=1) at parser.c:143
#8  0x000055555555d7f7 in cmdloop (top=top@entry=1) at main.c:235
#9  0x000055555555da4a in main (argc=1, argv=0x7fffffffe168) at main.c:199
(gdb) up
(gdb) print cmdenviron[0]
$2 = {next = 0x0, text = 0x1 <error: Cannot access memory at address 0x1>}
```

### DIFF
--- a/elkscmd/ash/linenoise_elks.c
+++ b/elkscmd/ash/linenoise_elks.c
@@ -257,8 +257,7 @@ void linenoiseSetMultiLine(int ml) {
 /* Return true if the terminal name is in the list of terminals we know are
  * not able to understand basic escape sequences. */
 static int isUnsupportedTerm(void) {
-    //char *term = getenv("TERM");
-    char *term = bltinlookup("TERM", 1);    /* getenv won't work here for ELKS*/
+    char *term = termval();
     int j;
 
     if (term == NULL) return 0;
--- a/elkscmd/ash/mkbuiltins
+++ b/elkscmd/ash/mkbuiltins
@@ -89,11 +89,11 @@ do
 	shift
 	for fun
 	do
-		echo "	\"$fun\", $i,"
+		echo "	{ \"$fun\", $i },"
 	done
 	i=`expr $i + 1`
 done < $temp
-echo '	NULL, 0
+echo '	{ NULL, 0 }
 };'
 
 exec > builtins.h
--- a/elkscmd/ash/options.c
+++ b/elkscmd/ash/options.c
@@ -175,10 +175,8 @@ options(cmdline) {
 
 
 STATIC void
-setoption(flag, val)
-	char flag;
-	int val;
-	{
+setoption(int flag, int val)
+{
 	register char *p;
 
 	if ((p = strchr(optchar, flag)) == NULL)
--- a/elkscmd/ash/parser.c
+++ b/elkscmd/ash/parser.c
@@ -684,7 +684,7 @@ STATIC int readtoken(void) {
 #define RETURN(token)	return lasttoken = token
 
 STATIC int xxreadtoken(void) {
-	register c;
+	int c;
 
 	if (tokpushback) {
 		tokpushback = 0;
--- a/elkscmd/ash/parser.c
+++ b/elkscmd/ash/parser.c
@@ -104,7 +104,9 @@ STATIC void parsefname __P((void));
 STATIC void parseheredoc __P((void));
 STATIC int readtoken __P((void));
 STATIC int readtoken1 __P((int, char const *, char *, int));
-void attyline __P((void));
+#if ATTY
+STATIC void attyline __P((void));
+#endif
 STATIC int noexpand __P((char *));
 STATIC void synexpect __P((int));
 STATIC void synerror __P((char *));
--- a/elkscmd/ash/show.c
+++ b/elkscmd/ash/show.c
@@ -268,9 +268,8 @@ trace(fmt, a1, a2, a3, a4, a5, a6, a7, a8)
 }
 
 
-trputs(s)
-	char *s;
-	{
+void trputs(char *s)
+{
 	if (tracefile == NULL)
 		return;
 	fputs(s, tracefile);
@@ -279,10 +278,9 @@ trputs(s)
 }
 
 
-trstring(s)
-	char *s;
-	{
-	register char *p;
+void trstring(char *s)
+{
+	char *p;
 	char c;
 
 	if (tracefile == NULL)
--- a/elkscmd/ash/var.c
+++ b/elkscmd/ash/var.c
@@ -79,9 +79,7 @@ struct var vps1;
 struct var vps2;
 struct var vpse;
 struct var vvers;
-#if ATTY
 struct var vterm;
-#endif
 
 const struct varinit varinit[] = {
 #if ATTY
@@ -96,9 +94,7 @@ const struct varinit varinit[] = {
 	 */
 	{&vps2,	VSTRFIXED|VTEXTFIXED,		"PS2=> "},
 	{&vpse,	VSTRFIXED|VTEXTFIXED|VUNSET,	"PSE="},
-#if ATTY
 	{&vterm,	VSTRFIXED|VTEXTFIXED|VUNSET,	"TERM="},
-#endif
 	{NULL,	0,				NULL}
 };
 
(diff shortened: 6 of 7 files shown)


ccoffing:
Tested on x64 Linux + valgrind, and on ELKS on real hardware.

ghaerr:
Hello @ccoffing,

Nice work catching this bug! 

> Commit https://github.com/jbruchon/elks/commit/8da6aec933dc1b8e071364b11661c04141f4a85b included this change:

I hadn't noticed `struct varinit variant[]` potentially already having a TERM variable when I wrote the buggy `builtinlookup` prior fix, nice.

> I suspect that the allocations are popped off the "stack" during an error

The main loop in `ash/main.c::main()` uses `setjmp/longjmp` for error processing, and in addition includes some tricky stack variable reset code using `setstackmark` and `popstackmark`. I suspect this is near where the variables are being wiped. I traced this through for another issue a while back and decided it got complicated quickly, as far as trying to change that overall behavior.

I really like the idea of running as many ELKS programs as possible on the host for testing, and sometimes the command isn't otherwise available and is useful in its own right. We have built this into a few of the elkscmd/\*/Makefiles (compress and basic, for example). Unfortunately, each Makefile does things a little bit differently (some use HOST_CFLAGS, other HOSTCFLAGS, and create `compress.host` vs `hostbasic`, etc).

It would be nice to add the ability to compile `ash` for the host the same way using elkscmd/ash/Makefile. Were you able to use most of the existing Makefile, I'd be interested in adding whatever is needed to create a `hostash`. I'm thinking we could standardize on using HOSTCC/HOSTCFLAGS/HOSTLDFLAGS and adding targets `hostXXX` (instead of XXX.host, like compress currently uses). I'll take a cleanup pass for the HOSTXXX flags issues, and let me know if there's something special that needs to be added to ash/Makefile for host building of it.

Thank you!

ccoffing:
I have been building natively ad-hoc, explorative.  I value sharpening the saw but it's still just so easy to find something to improve (I have a queue of such things to look at already).  I'll look at the existing HOST* flags and see where I can improve that, but if you beat me too it, that's fine too.

I found this because I was thinking: how painful would it be to have to use this shell as my daily driver?  That line of thought quickly leads to triaging bugs, triaging missing features (FreeBSD's sh, evolved from ash, is not a very high bar of usability but still something to consider), and then thinking about how to cut size to compensate.

Might be interesting to do some diffs of ash against MMMM branches (I just made that up:  "modern, mainstream, maintained, minimal".)  I started looking at FreeBSD's last night.  Maybe NetBSD would be even more minimal.  Just see if they've fixed obviously buggy areas; backport those.  See how they did testing (FreeBSD has a collection of shell script snippets and expected output - good approach.)  See if they have conditionalized or modernized chunks of code; follow their lead.  Where it's easy, and doesn't increase code size, follow their lead to decrease divergence.

Regarding decreasing divergence (or simply catching up):  Assume __STDC__.  Assume ANSI prototypes support.  Drop `register`.  Where applicable this would be done by literally copying lines from the adopted pseudo-upstream; otherwise following style.  I would open such PRs if there is agreement.

Another topic is improvements to linenoise.  I might slim and fix some things, but pushing them upstream is probably more selective, lower priority.

ghaerr:
Hello @ccoffing,

> Might be interesting to do some diffs of ash against MMMM branches

I dove down that rabbit hole about a year ago when looking for later-added bash-compatible string substitution that I thought would be useful for some of our more evolved /etc/rc.sys and networking startup scripts. I found the following links to various Almquist shell versions, most of which I looked at:

[Ash (Almquist Shell) Variants](https://www.in-ulm.de/~mascheck/various/ash/)
[Where to find the source code for the Almquist Shell?](https://unix.stackexchange.com/questions/276477/where-to-find-the-source-code-for-the-almquist-shell)
[old-minix ash source](https://github.com/chenshuo/old-minix/tree/master/src/commands/ash)

I ended up taking a hard look at an `ash` vintage 1991-3 or so that was still quite small [attached here](https://github.com/jbruchon/elks/files/10459382/ash2.zip). Its major problem is that the shell arithmetic handling was replaced with a yacc grammar, which I viewed as not desirable for our build. It had a few enhancements to the internal `test` command, which allows for `/bin/test` to be omitted from our floppy distributions, and some of that code I back-ported into our version. My take is that later versions are likely too large for our default /bin/sh, although you're certainly welcome to contribute a modern day shell, should you wish, and we can make a way to have that be configured in.

I mention most of this because of the business of source code cleanup, discussed a bit more below. In some cases, possibly with our ash shell source, it may be desirable to leave it as-is, as diff's with future versions are much easier. Try this with the attached 1993 source and see what you think.

> Regarding decreasing divergence (or simply catching up): Assume STDC. Assume ANSI prototypes support. Drop register. 


Agree on all STDC, ANSI prototypes, register, etc, with the caveat of possibly leaving be certain larger source sets that we may want to diff in the future (this is probably limited to ash, vi/elvis, levee although these three do seem to spit a lot of our remaining compiler warnings).

> I would open such PRs if there is agreement.

That'd be great. I just finished removing most of the pre-ANSI prototype support from the C library include files, there may be more but took a pretty big swipe. Almost all the elkscmd/minix\*/ code is pre-ANSI and needs help. The benefits of compiler argument checking are enormous, and we're considering moving to `-mregparmcall` calling convention, which absolutely requires exact function argument matching. For that reason alone, it may be worth converting `ash` source as well for enhancing correctness - take a look and see what you think.

> Where applicable this would be done by literally copying lines from the adopted pseudo-upstream; otherwise following style.

That could be a good idea for ash, if you find it helps. Of course, for those that might want to diff source, we can always go back to our older git repo versions should we really need/want to.

[EDIT: Although `ash` has history using the up/down arrow keys, I do miss `!`-style history, especially `!!`. Having that would be a nice addition].

> Another topic is improvements to linenoise. I might slim and fix some things

The first linenoise support was contributed pretty much stock from an older version and didn't work very well - I took it and added in ELKS terminal support, function key processing, and rewrote command completion. You're welcome to improve it. Please be sure to test on both the ELKS console as well as (emulated) serial terminal output - we've added code that allows `vi` and other programs to run well on console and serial lines, which use the VT DSR (device status report) to determine column width. The linenoise habit of sending that sequence on every command was turned off, due to some problems with character dropping "noise" (no pun intended).

Thank you!

PR #1526, opened 2023-01-19
[libc] Slightly optimize setting of __argv & __argc variables

tkchia:
This saves about 3 bytes of code space. :neutral_face:


ghaerr:
Hello @tkchia,

Looks great. Thanks to you, our startup runtime size is tighter than ever!

Thank you!

tkchia:
Thanks @ghaerr !

PR #1530, opened 2023-01-21
Fix some typos in docs and comments

ccoffing:


config/Configure.help
```diff
@@ -680,9 +680,8 @@ CONFIG_BLK_DEV_BFD1_720
 CONFIG_BLK_DEV_BFD1_1200
 CONFIG_BLK_DEV_BFD1_1440
 CONFIG_BLK_DEV_BFD1_2880
-CONFIG_CPU_80186
+CONFIG_CPU_8018X
 CONFIG_CPU_80286
-CONFIG_IBMPC_COMPAQ
```
ghaerr, line 685: In the above case, the CONFIG_CPU_80186 change to CONFIG_CPU_8018X should not be made: the "8018X" label in ELKS was added much later and refers to the embedded microcontrollers 80188 and 80186, for which we are using CONFIG_ARCH_8018X. The older original CONFIG_CPU_80186 and CONFIG_CPU_80286 have been removed and are not used at all, as their initial meaning was to configure the compiler to emit opcodes for those CPUs. We've decided against that as ELKS should (at least for now) run on any Intel x86 or compatible chip.

If you'd like, both CONFIG_CPU_80186 and CONFIG_CPU_80286 can be deleted here.

elks/arch/i86/defconfig
```diff
@@ -18,7 +18,7 @@
 # System
 #
 CONFIG_ARCH_IBMPC=y
-# CONFIG_ARCH_8081X is not set
+# CONFIG_ARCH_8018X is not set
```
ghaerr, line 21: Nice catch!!

config/Configure.help
ccoffing: I'll revert the change.  There is still a reference to `CONFIG_CPU_80286` in cputype.S so I'll leave any such cleanup to be done in concert with a code change.

elks/arch/i86/defconfig
ccoffing: I did a quick grep / sed to compare the list of CONFIG_* variables documented vs used in the code, curious if there were typos that caused help text to not appear.  There were a couple (this and the compaq one).

PR #1531, opened 2023-01-21
[libc] Fix fd leak: Close even if flush fails

ccoffing:
POSIX says that if fclose fails, the state of the handle is undefined, so the app is not allowed to try again to close it.  And it's widely known that if you really care about data integrity, you must manually flush and error check first, before closing.  So no matter what, fclose should close.

### DIFF
--- a/libc/stdio/fclose.c
+++ b/libc/stdio/fclose.c
@@ -7,48 +7,48 @@
 int
 fclose (FILE *fp)
 {
-   int   rv = 0;
+    int   rv = 0;
 
-   if (fp == 0)
-   {
-      errno = EINVAL;
-      return EOF;
-   }
-   if (fp->fd != -1)
-   {
-      if (fflush(fp))
-	 return EOF;
+    if (fp == 0)
+    {
+        errno = EINVAL;
+        return EOF;
+    }
+    if (fp->fd != -1)
+    {
+        if (fflush(fp))
+            rv = EOF;
 
-      if (close(fp->fd))
-	 rv = EOF;
-      fp->fd = -1;
-   }
+        if (close(fp->fd))
+            rv = EOF;
+        fp->fd = -1;
+    }
 
-   if (fp->mode & __MODE_FREEBUF)
-   {
-      free(fp->bufstart);
-      fp->mode &= ~__MODE_FREEBUF;
-      fp->bufstart = fp->bufend = 0;
-   }
+    if (fp->mode & __MODE_FREEBUF)
+    {
+        free(fp->bufstart);
+        fp->mode &= ~__MODE_FREEBUF;
+        fp->bufstart = fp->bufend = 0;
+    }
 
-   if (fp->mode & __MODE_FREEFIL)
-   {
-      FILE *prev = 0, *ptr;
-      fp->mode = 0;
+    if (fp->mode & __MODE_FREEFIL)
+    {
+        FILE *prev = 0, *ptr;
+        fp->mode = 0;
 
-      for (ptr = __IO_list; ptr && ptr != fp; ptr = ptr->next)
-	 ;
-      if (ptr == fp)
-      {
-	 if (prev == 0)
-	    __IO_list = fp->next;
-	 else
-	    prev->next = fp->next;
-      }
-      free(fp);
-   }
-   else
-      fp->mode = 0;
+        for (ptr = __IO_list; ptr && ptr != fp; ptr = ptr->next)
+            ;
+        if (ptr == fp)
+        {
+            if (prev == 0)
+                __IO_list = fp->next;
+            else
+                prev->next = fp->next;
+        }
+        free(fp);
+    }
+    else
+        fp->mode = 0;
 
-   return rv;
+    return rv;
 }


ghaerr:
Hello @ccoffing,

Agree with your fix, thanks. However, I am concerned about the total file reformatting for two reasons: 1) I think we're better off moving to a 4 character indent and no tabs, which is the way this file was originally formatted (as is most of the rest of libc/stdio), and 2) it makes it very hard to see any changes made when the entire file is reformatted.

Thus, we probably need to discuss and conclude on a tab formatting convention going forward. Most of libc and the kernel are formatted using an initial 4 spaces, then 4-position indents, but tabs used when the column matches an 8-char tab position. This turns out to make things very hard to read when one uses an editor set for 4-char indents. (That's why I have 'k' and 'l' set to expand to 4 and 8 char tabs so as to be able to see the intended formatting). All that said, I think it is a mistake to convert any majority space-indented files back to tab-indented, and would suggest that only spaces be used throughout if we're going to reformat whole files. Thus, in libc and the kernel, a full file reformat still allows a diff which allows for easy understanding of the reformatting (since the previously tabbed lines are usually anchored within non-reformatted lines) as well as the overall other changes made.

If we applied the above suggestion for reformatting fclose.c, then only 7 lines would change, including your fix.

Thank you!

ccoffing:
I agree with what you're saying.  I'm on the 4-space bandwagon, but only doing it as I touch a function.  But I think you might be looking at the diff backwards.  Or maybe github's coloring of it is confusing the issue.

In `libc/system` (and also in much of `libc/stdio`, as in my other PR), many files are a mix of 3-space indent and tabs.  I applied a consistent 4-space indentation to this function.  I think github is considering the change of a single tab to 8 spaces as inserting 7 "new" characters, and therefore rendering it as dark green (which might be suggestive of a tab?)

It's all 4-space indent.

ghaerr:
>  But I think you might be looking at the diff backwards. Or maybe github's coloring of it is confusing the issue.

Let me pull the source and take a look - perhaps I'm reading the diff incorrectly.

>  I'm on the 4-space bandwagon, but only doing it as I touch a function.

Ok, lets agree on that.

>  many files are a mix of 3-space indent and tabs

Yes, kind of a mess. Ok - I see now: the reformat ends up being hard to read because the entire source was 3 (not 4) character indents AND spaces. So our problem is a bit larger than described - not only do we have tab issues, but also possible reindentations based on non-4 or -8 indents.

I'm still undecided as to the best answer for upgrading our source problem(s): should we take a file from a family of 3-character indented sources and move it all to 4? Certainly, any tabs should be replaced. But are we doing ourselves any favors reformatting just one file to a new format, which now doesn't match the rest? Another lurking issue (now seen in your nice system/opendir.c fix) is that of brace placement: should have the "Whitesmiths" (all braces on their own lines) combined with K&R both in the same file? Or should they be made to match the original, majority, or preferred style?

Just thinking out loud for a moment!

ccoffing:
Appreciate the thinking out loud.

Similarly:  I have some fixes to the utent code.  Minor cleanup of returns, minor de-duplication of some error code paths, DRY, that sort of thing.  But it just sits, because the formatting of that file is so terrible (2 spaces and tabs and weird braces and no spaces around operators) that it hurts to touch it while perpetuating the weird style.

ghaerr:
> But I think you might be looking at the diff backwards. Or maybe github's coloring of it is confusing the issue.
>  I think github is considering the change of a single tab to 8 spaces as inserting 7 "new" characters, and therefore rendering it as dark green (which might be suggestive of a tab?)

Ok - I've finally figured this out, its a GitHub coloring issue: since the file was formatted with 3 spaces, and you've added another (to 4), the diff shows the extra space, which I thought looked like a tab character after leading spaces. It seems there may not be an easy way to determine the difference between 8-indent tabs and spaces, but the good news is that shouldn't matter to us since we're going to be removing tabs in general.

ghaerr:
>  I have some fixes to the utent code. Minor cleanup of returns, minor de-duplication of some error code paths, DRY, that sort of thing. But it just sits, because the formatting of that file is so terrible (2 spaces and tabs and weird braces and no spaces around operators) that it hurts to touch it while perpetuating the weird style.

Geez! Well, I'm not one to want to shun enhancements made by willing, non-paid contributors, just because the source code is 20+ years old and ugly as heck!!! :) 

How about this: when it is felt that a source file needs to be updated (either because its style is absolutely terrible, and/or that one wants to make some enhancements to it), we proceed with a reformatting-only PR with the title something like "Reformat utent.c - no ops changes", which re-tabs and possibly reformats the file entirely (preferably using an `indent.sh` or similar with some basic agreed-on settings). Then, after the reformat is quickly committed, a subsequent PR is submitted with enhancements, etc. In this way, it remains very clear what is reformatting, and what is enhancements. Even if an enhancement is not planned, we have some very ugly files that could use just a reformat for reading purposes alone for which this approach would be useful.

In the case of smaller files, like stdio/fclose.c or system/opendir.c, we could do it in a single PR, although I'd really rather see (in the future) a separate PR with both files reformatted, then changes separately. So you don't have to start over, we can commit both open PRs now, to keep moving things forward.

What do you think?

ghaerr:
In addition, I would go along with the idea that, even though in the case of libc (or the kernel) for instance, file families that are tabbed 2- or 3- spaces be reformatted to 4-, on a per-file modified basis. To do otherwise just promotes our codebase remaining out of date, and there are already too many examples of this, frankly. If we fix the formatting issues when we're in the file, then at least slowly the codebase should come around to being consistent and highly readable. Having any major file reformats occur in a single PR with no other changes allows anyone who cares to carefully inspect the file history when searching backwards for bugs introduced, etc.

ccoffing:
> What do you think?

This all sounds sane to me.

I could even go quickly rewrite my remaining PR, with the reformatting in a first commit, and the actual changes in a separate commit.  Even when those 2 commits are in a single PR, that would make the real changes easier to review, and also would still simplify reviewing the git history.

PR #1532, opened 2023-01-21
[libc] Optimize libc's opendir more like ash's

ccoffing:
Optimize for the common case and do the open first.  This avoids a namei() call; verifying that it is a directory is just a flag check. This also closes the race of interpreting the name twice and possibly getting inconsistent results.

Use a single allocation; avoids a malloc/free.

### DIFF
--- a/elkscmd/test/libc/system.c
+++ b/elkscmd/test/libc/system.c
@@ -49,25 +49,25 @@ TEST_CASE(system_sleep)
 
 TEST_CASE(system_dirent)
 {
-	DIR *d;
-	struct dirent *de;
+    DIR *d;
+    struct dirent *de;
 
-	d = opendir("doesnotexist");
-	EXPECT_EQ(errno, ENOENT);
-	ASSERT_EQ_P(d, NULL);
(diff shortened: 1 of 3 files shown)


ghaerr:
Hello @ccoffing,

Nice work. I completely agree with the `stat` vs `open` then `fstat` and avoiding an additional name() call. While the directory data involved should almost always be buffered, we've found in the past, especially on FAT, that an unneeded `stat` call greatly slowed down `ls` operation in 3+ deep directory structures.

Also nice enhancement on using a single allocation - that's been on my list for a while, as with our allocator(s) its not always a given that the memory will be easily reusable by the application, and the data segment remains tight for larger applications.

With regards to formatting issues discussed in https://github.com/jbruchon/elks/pull/1531#issuecomment-1399288481, let me pull down the revised source to confirm we're on the same page.

Thank you!

ghaerr:
Thanks @ccoffing! Actually, I thought I committed this along with your other fclose PR, but apparently didn't. In the future we can go either this way with both the reformat and enhancement included as separate commits, or as separate PRs.

PR #1534, opened 2023-01-28
ls: fix behavior and crash and opt

ccoffing:
Here are some improvements to ls.

I landed here only because I was perplexed why `ls` from a fresh ELKS install with an empty home directory printed a blank line before the prompt.  UNIX is supposed to be terse.  This is nonstandard behavior, and is confusing because piping it to `wc -l` returns 0 (correct) not 1.

Then noticed that recursive listing would abort on the first permission denied, which is nonstandard.  It should continue where it can, but still report an error.

Then I found some fun stack smashers and optimization opportunities.  Listing deeply nested directories was bad.  I hate excessive strcpy and strlen.  If you already know a length, use it to write more optimal code.

I also see other cross-cutting concerns which I have not addressed.  Long pathnames can cause problems all over (mkdir is terrible).  Username length is not consistently enforced, and can allow stack smashes all over.  etc.

But this is a little better.

### DIFF
--- a/elkscmd/file_utils/ls.c
+++ b/elkscmd/file_utils/ls.c
@@ -578,7 +578,7 @@ int main(int argc, char **argv)
             if (recursive) recursive--;
         }
     } while (files.size || dirs.size);
-    if (!(flags & (LSF_LONG|LSF_ONEPER)))
+    if (!(flags & (LSF_LONG|LSF_ONEPER)) && col)
         fputc('\n', stdout);
     return EXIT_SUCCESS;
 
--- a/elkscmd/file_utils/ls.c
+++ b/elkscmd/file_utils/ls.c
@@ -82,11 +82,11 @@ struct stack
     struct sort *buf;
 };
 
-static int cols = 0, col = 0;
+static int cols, col;
 static int reverse = -1;
-static int sortbytime = 0;
-static int sortbysize = 0;
-static int nosort = 0;
+static int sortbytime;
+static int sortbysize;
+static int nosort;
 static char fmt[16] = "%s";
 
 /* return -1/0/1 based on sign of x */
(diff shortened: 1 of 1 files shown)


elkscmd/file_utils/ls.c
```diff
@@ -444,15 +440,8 @@ static void setfmt(struct stack *pstack, int flags)
...
-    char *p;
-
-    len = strlen(name);
-    p = name + len - 2;
-    if (strcmp(p, "/.") == 0) return 0;
-    p--;
-    if (strcmp(p, "/..") == 0) return 0;
-    return 1;
+    char *p = strrchr(name, '/');
+    return !(p && p[1] == '.' && (!p[2] || p[2] == '.' && !p[2]));
```
ghaerr: I'm having a bit of trouble understanding whether this last test is in fact equivalent to the code it's replacing. Are you sure you don't mean `return !(p && p[1] == '.' && (!p[2] || p[2] == '.' && !p[3]));`? Including parens on the last `&&` test would seem to show that that result would always be false.

Regardless, perhaps there should be a test to ensure `ls` acts the same when executing `not_dotdir()`.

ccoffing: You are right, fixed (and added the parens).

What's your thought on testing standard command line utils?  It makes sense to unit test libc; it's large and exports functions with specified behaviors.  But that's not the case for these utils.  I can only think of writing shell script(s?) to drive them and verify the external behavior.  In fact I did a bit of that when I did the fixes to `tr`.

ghaerr: My concern is mostly that tests have been made where the source code has been changed, since I am only eyeballing the changes, not actually testing. I'm not sure that means having to build test scripts, although I really like what you've done with libc testing as far as providing regression testing for future changes.

In the case of `ls`, due to its design, I have found it somewhat tricky to test, so having on hand even a shell script of various degenerative or improved operational cases (to be visually tried on both FAT and MINIX filesystems) would probably be a good idea. (The kernel directory handling code is vastly different between FAT and MINIX, and we ran into some slow speed problems a while back on heavily nested directories on FAT, for instance). Some of these tests require specially-built disk images though.

That said, building a test framework comparing the output of command line utilities is messy and a lot of work, and probably not worth the effort at this point. Perhaps just including (and updating) a `ls-test.sh` or something like that in the source directory might be a good idea. I have a few of these that I keep offline for `mkfs`, `ramdisk` and other utilities that I used when fixing them that I could submit. If nothing else, the `*.sh` files could at least show what has been tested, which would be nice as we continue to strive for stability and correctness in ELKS.

[EDIT: Perhaps a better location for the any test scripts would be in elkscmd/test/scripts/, with names like `ls.sh`, etc. That would keep the source directories cleaner and the tests together, able to be more easily copied to a disk image.]

Thank you!

ccoffing: I have done further manual testing of these changes, for normal use cases and for deeply nested directories (for which, unmodified ls could hang or infinite loop).  I did the stress testing with Minix but I sanity checked FAT too.  The optimizations are small but measurable (assuming warm cache).
I might think about some test scripts this weekend.
Thanks!

PR #1539, opened 2023-02-09
Mail

ccoffing:
This was a quick "why not", but it does say `No mail.` under elksemu so that is encouraging.

Why did I have to do this to be able to use stdio `rewind` ?  It appears that nothing else has tried to use it previously.
```
-#ifdef L_rewind
```

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -155,6 +155,7 @@ minix3/tee						:minix3					:1200k	:1440k
 minix3/cal						:be-minix3				:1200k	:1440k
 minix3/diff						:be-minix3			:720k
 minix3/find						:be-minix3			:720k
+minix3/mail						:minix3                         :other
 disk_utils/fsck					:diskutil				:1200k	:1440k
 disk_utils/mkfs					:diskutil		:360k
 disk_utils/mkfat				:diskutil		:360k
--- a/elkscmd/man/man1/mail.1
+++ b/elkscmd/man/man1/mail.1
@@ -1,0 +1,124 @@
+.TH MAIL 1
+.SH NAME
+mail \- send and receive electronic mail
+.SH SYNOPSIS
+\fBmail\fR [\fB\-epqr\fR] [\fB\-f\fR \fIfile\fR] 
+.br
+\fBmail\fR [\fB\-dtv\fR] [\fB\-s\fR \fIsubject\fR] \fIuser\fR [...]
+.br
+.de FL
+.TP
+\\fB\\$1\\fR
+\\$2
+..
+.de EX
+.TP 20
+\\fB\\$1\\fR
+# \\$2
+..
+.SH OPTIONS 
+.TP 5
+.B \-e
+# Exit with status TRUE or FALSE to indicate if there is mail in mailbox
+.TP 5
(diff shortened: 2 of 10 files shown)


ccoffing:
Looks like the new stdio tests are failing, so good fishing there too.  Please hold.....

ghaerr:
> Why did I have to do this to be able to use stdio rewind ? It appears that nothing else has tried to use it previously.

LOL, yes it seems you're right - nobody's used `rewind` yet. Your changes removing L_rewind look fine.

> This was a quick "why not", but it does say No mail.

Very cool! I'll have to play with it after you've got it committed. You'll want to add an entry to elkscmd/Applications, perhaps to `:other`, so that it gets copied out to the 2880k+ media, since 1440k is pretty full.

> Looks like the new stdio tests are failing, so good fishing there too. Please hold.....

I haven't had time to look at this, are they failing because of your changes, or for other reasons? Let me know when you're ready to continue.

Thank you!

ccoffing:
Okay, fixed the test.  I had a bug in the test ~~, which I think exposed a minor bug in libc:  fopen as "w" (not "w+") then try to fread, will return 0 bytes, but no error indication.  Not relevant here.~~  (nevermind, I think it was simply a test bug.)

I added the man page from Minix.  The output of "man mail" is a bit garbled, but I haven't changed anything in that regard for this PR.  I'm reading man.c first.

Do all the man pages always get installed?  It would be nice to install (or not) them according to whether the program gets installed.

ghaerr:
Thanks! Does `mail` work? What is required to configure it? (I haven't read the man page yet :)

> The output of "man mail" is a bit garbled, but I haven't changed anything in that regard for this PR. I'm reading man.c first.

The ELKS `man` command (there are two of them, we're using the smaller, simper (and better) one in sys_utils/man), supports a subset of markup commands which aren't really documented fully, but can be found by looking at other man page sources and copying. You can also use `eman mail` (Elks man page) for instance to display an ELKS man page from the elkscmd/man/\* directory on your host system, without having to boot ELKS.

> Do all the man pages always get installed?

Yes, either all of them or none of them, depending on the CONFIG_APP_MAN_PAGES setting. This is because we're pretty short on disk space for most images except 2880k+. We do have an executable binary compressor that I spent quite a bit of time writing, which allows for 30+% compression, but there were complaints that the decompress time took too long on ancient PCs. The distribution mechanism compressed either all or none of the executables. I'm still thinking about a possible better approach where only some of the less-used executables are compressed.

> It would be nice to install (or not) them according to whether the program gets installed.

Yes, that's probably a good idea for section 1 pages, but there's still a lot of other sections that take up lots of space. We could possibly create another file like elkscmd/Applications which could be used for man pages, or perhaps update the installation script to find matching section 1 man pages in elkscmd/Applications and install them. We'll still have a space problem on 360-720k floppies likely, and may need a config option for that. I suppose it comes down to more discussion as to really what should be on each floppy image size we distribute.

Thank you!

tyama501:
1232 disk for PC-98 is also almost full.
(I needed to delete some to debug elksmoria)

Maybe we need optional Application disk image,
besides the basic image in the future.

PR #1540, opened 2023-02-11
UTF8 nits

ccoffing:


ghaerr:
Hello @ccoffing,

> Match UTF8 define used elsewhere (although not set) for consistency

Actually, other than of course ELKS not actually supporting UTF8, this is still a little goofed up: using `UTF8` or `ENABLE_UTF8` is fine, but doesn't match anything elsewhere in ELKS. This leaves an undefined `CP437` case for systems that can display CP 437 such as the ELKS console. As you may have seen elsewhere in the elkscmd/tui directory, an attempt was made to allow Unicode characters that can be converted to CP 437. That allows, for instance, `matrix` to run on the ELKS console as well as a serial terminal, with the TUI library doing the conversion directly to CP 437 for the console, and to UTF8-encoded Unicode for the serial terminal. Thus, what is really needed is likely to move the TUI library UTF8 conversion routines into libc, along with some kernel console driver translation of UTF8 to CP 437. Then we could allow the CP 437 subset expressed in Unicode, rather than code page bytes.

In addition, the sometimes-found `IUTF8` define is supposed to be used to show that UTF8 output is supported on the host, which is usually, but not always, the case on hosts supporting UTF8 output (this define is found in <termios.h>). However, I just noticed that ELKS termios.h defines it (it shouldn't, but is likely included because elkscmd/tui/tty.c uses it without first checking whether its defined). I've made a note of that for future fixing.

Thank you!

Issue #1502, opened 2022-12-31
NX demos fail to start without serial card


toncho11:
I tried all of the NX demos and it says:

- no mouse
- could not open graphics 

Is it possible that the lack of mouse is actually causing the graphics lib to fail?
I am testing using a 1440 image on an IBM 5160 with a VGA card.
I do not have a serial card on this computer.

Also should nxterm require a serial mouse?

toncho11:
Usually the nxdemos require just a VGA card and enough RAM.
There should be at least one demo that is able to run without requiring a mouse.

toncho11:
I am using some old image of ELKS: 5b5567f2 24/05/2022

ghaerr:
> Is it possible that the lack of mouse is actually causing the graphics lib to fail?

NX is not so much a graphics library but instead a windowing environment, which requires a mouse. When the serial open fails in the mouse driver because of no serial port, the failure return causes NX to exit.

The NX demos can be compiled to not use a mouse, as they are on the PC-98 (see nanox/Makefile::drivers/mou_null.o). This is done using the "null" mouse driver.

I'm not sure if it would be a good idea to allow NX to run without a mouse, as users might get confused as to how to exit the program (ESC usually works). Any error messages also would be erased by the graphics clear screen, thus adding to confusion, IMO.

Nano-X itself is pretty much designed to require a mouse, so currently the null mouse driver must be compiled in for systems without a serial port. As stated above, I hesitate a bit to think it a good idea to allow the serial mouse driver to return OK with no mouse.

toncho11:
I just tested on latest build. 
I had to insert my XT-IDE to avoid the network probing bug ... otherwise it halts ... annoying ...

Only the nxlandmine is available in the 1440 now.
Still there should be some program that does not require mouse in order to test that graphics is available on ELKS. 
Maybe calling BASIC with some program ... but currently that is not available. 

I think there can be a simple message "Mouse not detected. Expect problems! Do you wish to continue (Y/N)". This way we keep testing NX even on computers without a mouse.

tyama501:
Oh, I thought nxterm and nxworld also available.

nxclock and nxdemo need to rebuild.

tyama501:
Sorry, yes, the nxlandmine is the only program that need mouse.

ghaerr:
[...]

PR #1542, opened 2023-02-19
Add nxtetris, added Nano-X GrGetNextEventTimeout, other bugfixes

ghaerr:
Adds `nxtetris` and `nxtest` to distribution when CONFIG_APP_OTHER=y set.

`nxtetris` was discussed a bit in https://github.com/jbruchon/elks/issues/871#issuecomment-1435875961, and can run without a mouse, using the s/d/j/k/space keys. The game speed can be increased or decreased using the +/- keys.

`nxtest` allows testing of the graphics drawing speed. It starts out drawing rectangles 10ms apart, but can be speed up or slowed down by using the +/- keys, up to a maximum speed quite faster than the original version.

Here's both files for testing without having to build:
[nxtetris.zip](https://github.com/jbruchon/elks/files/10778301/nxtetris.zip)


toncho11:
@ghaerr Is it possible to hide the mouse pointer? I use only keyboard.

ghaerr:
> Is it possible to hide the mouse pointer? I use only keyboard.

Yes, if you can confirm that you're using the `MOUSE_PORT=none` method of disabling mouse support. Otherwise, the pointer needs to remain visible.

How is the new game speed? I'm interested if it works well on older PCs. I ended up having to implement an event timeout scheme in Nano-X which synchronizes to clock time so that the speed should be the same on all systems.

toncho11:
`export MOUSE_PORT=none` does not work in the sense that the mouse pointer is always there. I am not using the latest build of ELKS though.
The initial speed is fast for me. But I am able to slow it down with "-" and then it is quite playable :)
Thank you!
@tyama501 does it work for you?

ghaerr:
> export MOUSE_PORT=none does not work in the sense that the mouse pointer is always there. 

I'm aware that the mouse pointer will always be visible - my question is: are you using `export MOUSE_PORT=none` in order to disable the mouse when you play the game on your PC? (This feature was requested by you some time back). I can't just remove the mouse pointer without associating it with a mechanism for mouse disablement. Otherwise, users with mice will not see the mouse pointer when they play.

toncho11:
I do not have a mouse attached. Not having a mouse does not block the game from starting with or without `export MOUSE_PORT=none`. So it was not supposed to start without `export MOUSE_PORT=none`? #1502 

So, yes a possible improvement is when `export MOUSE_PORT=none` is to hide the mouse cursor and position the game in the center of the screen.

ghaerr:
> Not having a mouse does not block the game from starting with or without export MOUSE_PORT=none. So it was not supposed to start without export MOUSE_PORT=none?

Yes, that's what I thought was agreed on. I seem to remember that at least one of your systems would not start Nano-X unless we set MOUSE_PORT=none. Nano-X won't be able to open the mouse if there is no serial card at /dev/ttyS0.

> yes a possible improvement is when export MOUSE_PORT=none is to hide the mouse cursor and position the game in the center of the screen.

Ok, will do both.

toncho11:
I see. This computer has indeed a built-in serial card (Amstrad 1640) and that is why it does not require `export MOUSE_PORT=none`. The other one the IBM 5160 requires the `export MOUSE_PORT=none` as I do not usually put a serial card.

PR #1543, opened 2023-02-23
[cmds] Center nxtetris and don't display cursor if no mouse

ghaerr:
As requested in https://github.com/jbruchon/elks/pull/1542#issuecomment-1437640936.

### DIFF
--- a/elkscmd/nano-X/demos/ntetris.c
+++ b/elkscmd/nano-X/demos/ntetris.c
@@ -743,16 +743,19 @@ void new_game(nstate *state)
 
 void init_game(nstate *state)
 {
+	GR_SCREEN_INFO si;
 	//GR_WM_PROPERTIES props;
-	GR_COORD x = MAIN_WINDOW_X_POSITION;
 
 	if(GrOpen() < 0) {
 		printf("Can't open graphics\n");
 		exit(1);
 	}
 
-	state->main_window = GrNewWindow(GR_ROOT_WINDOW_ID, x,
-					MAIN_WINDOW_Y_POSITION,
+	GrGetScreenInfo(&si);
+	GrMoveCursor(si.cols - 16, 0);
+	state->main_window = GrNewWindow(GR_ROOT_WINDOW_ID,
+					(si.cols - MAIN_WINDOW_WIDTH) / 2,
+					(si.rows - MAIN_WINDOW_HEIGHT) / 2,
 					MAIN_WINDOW_WIDTH,
 					MAIN_WINDOW_HEIGHT, 0,
 					MAIN_WINDOW_BACKGROUND_COLOUR, 0);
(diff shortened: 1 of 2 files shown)


toncho11:
In this code: https://github.com/ghaerr/elks/blob/1bacf2613623d01edaf14a39ed0ab9e1550c9efa/elkscmd/nano-X/nanox/srvmain.c#L154 shouldn't be better to check if `mouse_fd>=0` before polling for a mouse event? If "none" is -2 then this will disable the polling and save some time?

ghaerr:
The particular version of `GsSelect()` you point to is for MSDOS, where there are no file descriptors. In MSDOS straight polling is used, which uses the `.Poll()` entry point of the compiled-in driver. ELKS uses the following routine, which is `#if UNIX`.

toncho11:
Yes, I just saw that and I deleted my comment, but you were quick to answer.

toncho11:
If you provide me with an executable I can test it.
Thank you for this update.

ghaerr:
Here you go: 
[nxtetris.zip](https://github.com/jbruchon/elks/files/10819550/nxtetris.zip)

toncho11:
Thanks. 
So the game is centered (which helps a lot on my monitor). 
The mouse pointer is always in the right top part/corner of the screen. 
The mouse pointer is still visible when `export MOUSE_PORT=none`.

It still prints "Can't open graphics" even if the problem is a missing mouse. For example if I do "export MOUSE_PORT=fsdfsdfsdfsdf" it will complain about the mouse, but also about the graphics (as a last message). 

Thank you!

ghaerr:
> The mouse pointer is still visible when export MOUSE_PORT=none.

Is this on your machine that has a working serial port? I'm not sure MOUSE_PORT=none is coded to disable the mouse, just to disable the mouse if a previous open fails. 

> It still prints "Can't open graphics" even if the problem is a missing mouse. 

Yes, that's the "application" displaying its error message after the server (which is linked to the application) has had its say about the mouse.

> For example if I do "export MOUSE_PORT=fsdfsdfsdfsdf" it will complain about the mouse

I suppose if MOUSE_PORT is set to something entirely incorrect, we could just have the server exit, which would exit the application too. The original coding allowed for an application to do something if Nano-X couldn't start, but it seems this isn't the best approach for app/server linked games.

I'll look further into both issues, thanks.

toncho11:
Looking at the code if "none" then the mouse_fd  = -2 and then 
```
if (mouse_fd == -2)
	    GdHideCursor();
```
So the cursor should not be visible (if this is what the hide function does indeed).
Is it ok that the "none" check is performed in the serial driver mou_ser.c. This means that the serial driver is already selected as a default mouse driver (I think there is more than one mouse available?).Just giving you some hints.

ghaerr:
> So the cursor should not visible (if this is what the hide function does indeed).

Yes. That's why I asked whether the machine you tested this on has a serial port. If so, then the cursor likely will remain visible due to the way I think its coded. I'll test further, something is still wrong.

toncho11:
Yes, it is on the machine with the serial port built-in.

toncho11:
Tested. Mouse is invisible on the computer without the serial card.
Thank you!
One thing is that it does not say that "n" is a new game, which can be confusing. And using the arrow keys exits the game.

ghaerr:
> And using the arrow keys exits the game.

Thanks for your testing @toncho11. Yes, the arrow keys now emit a multi-byte escape sequence with ESC at the start, and ESC is currently a quick way to exit any Nano-X program. I've made a note of it, and will check later to see whether we can somehow wire up the arrow keys to work in an expected way.

PR #1544, opened 2023-02-24
fgets boundary bug

ccoffing:
If fgets were passed a size of 0, it would read without bound and overflow the buffer.  I did a quick search of the codebase but didn't see a potential case of this bug (such as fgets in a loop with a dynamic size).

The behavior when size<=1 is poorly specified (see https://en.cppreference.com/w/c/io/fgets ) but elks now matches what glibc does.


ghaerr:
Thanks for looking very closely at our C library Chuck!

PR #1546, opened 2023-02-27
Login fixes

ccoffing:
Here's a fun one.

login

### DIFF
--- a/libc/getent/pwent.c
+++ b/libc/getent/pwent.c
@@ -1,7 +1,7 @@
 /*
  * pwent.c - This file is part of the libc-8086/pwd package for ELKS,
  * Copyright (C) 1995, 1996 Nat Friedman <ndf@linux.mit.edu>.
- * 
+ *
  *  This library is free software; you can redistribute it and/or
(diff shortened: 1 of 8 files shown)


ghaerr:
Hello @ccoffing,

Wow, you've jumped into some of the ugliest (formatted) code in libc - getent! Thanks for cleaning it up. I'd looked at it several times and didn't want to touch it :)

Nice work on the multiple commits for this PR.

Thank you!

libc/getent/pwent.c
```diff
@@ -46,17 +46,19 @@ setpwent(void)
...
 struct passwd *
 getpwent(void)
 {
-    if (pw_fd == -1)
+    if (pw_fd == -1) {
         setpwent();
-    if (pw_fd != -1)
-        return __getpwent(pw_fd);
-    return NULL;
+        if (pw_fd != -1)
```
ghaerr, line 60: Hello @ccoffing,

Just found this cleanup code introduced a new bug, which I've fixed in #1550. This compare should be `if (pw_fd == -1)`.

Thank you!

PR #1548, opened 2023-03-01
Fix incorrect usage of readlink

ccoffing:
@ghaerr the file in fm still has potential buffer overflows as you noted in the file, but at least the output from the readlink is now guaranteed to be properly terminated.

### DIFF
--- a/elkscmd/file_utils/cp.c
+++ b/elkscmd/file_utils/cp.c
@@ -349,7 +349,7 @@ static int do_copies(void)
 		} else if (flags == S_IFLNK) {
 			char lnkname[256];
 
-			int cnt = readlink(inode_build->path, lnkname, sizeof(lnkname));
+			int cnt = readlink(inode_build->path, lnkname, sizeof(lnkname) - 1);
 			if (cnt < 0) {
 				fprintf(stderr, "Can't read symlink "); fflush(stderr);
 				perror(inode_build->path);
--- a/elkscmd/minix2/remsync.c
+++ b/elkscmd/minix2/remsync.c
@@ -189,7 +189,7 @@ char *rdlink(const char *link, off_t size)
 	if (len <= size) {
 		path= allocate(path, (len= size * 2) * sizeof(path[0]));
 	}
-	if ((n= readlink(link, path, len)) == -1) return nil;
+	if ((n= readlink(link, path, len - 1)) == -1) return nil;
 	path[n]= 0;
 	return path;
 }
--- a/elkscmd/tui/realpath.c
+++ b/elkscmd/tui/realpath.c
@@ -39,7 +39,9 @@ static int resolve_path(char *path,char *result,char *pos)
 	    if (lstat(result,&st) < 0) return -1;
 	    if (S_ISLNK(st.st_mode)) {
 		char buf[PATH_MAX];
-		if (readlink(result,buf,sizeof(buf)) < 0) return -1;
+		int n;
+		if ((n = readlink(result,buf,sizeof(buf) - 1)) < 0) return -1;
+		buf[n] = 0;
 		*pos = 0;
 		if (slash) {
 		    *slash = '/';


ghaerr:
Hello @ccoffing,

Thanks for finding more problems. It seems we need a PATH_MAX, perhaps in a new libc <limits.h> (gcc supplies this file, so we might possibly need to `#include_next <limits.h>` in it as well. I think I previously tried `#include_next` and ran into problems though). This would stop the additional problem of arbitrary `lnkname[256]` (or shorter) in cp.c and elsewhere you've seen.

On the subject of arbitrarily truncating buffers (unless PATH_MAX is defined by libc/kernel and used properly), I thought you might find this article interesting: https://nrk.neocities.org/articles/not-a-fan-of-strlcpy. In summary, it's an error to truncate a buffer unless a maximum returned length is known. Thus, your fixes combined with PATH_MAX defined, always using a buffer[PATH_MAX], and passing sizeof(buffer)-1 is required to avoid Undefined Behavior.

Thank you!

ccoffing:
I see there is a `PATH_LEN` in `elkscmd/file_utils/futils.h`.  I assume that is just a somewhat arbitrary "big enough" value, not necessarily a precise limit consistently enforced by the kernel (and really, how can such a thing exist, considering varying filesystems, etc.)  There is some (limited) value in (locally) standardizing on that, because then you have a single symbol to search for when further refactoring the codebase to remove path length limit nastiness.

But yes, even better would be just going for a real `PATH_MAX` in libc.  I will toy with this and see where it goes.

Agreed that silently truncated buffers are a bug.  I recently experienced a kernel boot hang based on config (haven't bisected the bad config yet) so I was `mv`-ing kernels to named backups, and `mv` was happily truncating the names.  I wouldn't want to write a script on top of that behavior.

The real solution to all this bug-prone buffer-sizing and slow copying is improved kernel interfaces, which can manipulate files with respect to another opened file descriptor, eg openat, fstatat, etc.  Are there any technical hurdles to those in the ELKS kernel?

ccoffing:
BTW, that strlcpy link is great.

ghaerr:
> The real solution to all this bug-prone buffer-sizing and slow copying is improved kernel interfaces, which can manipulate files with respect to another opened file descriptor, eg openat, fstatat, etc. Are there any technical hurdles to those in the ELKS kernel?

At this point, I don't think its worth it to add a bunch of new-style openat/fstatat etc routines to the kernel, just to solve this pathname size issue. There are a lot of issues getting all that working, not to mention no applications use any of those calls at this time. Currently, all kernel routines get passed a buffer length and won't overwrite it. With the filesystems we've got, a PATH_MAX of 256, or even 128, (or even 80), would be entirely sufficient. (For stack conservation, I vote for the smaller value. Application stack size requirements add up and we're currently allocating a (large?!) 4K stack per executable, with many system programs operating with less using `-maout-stack=` settings. The problem is resident storage in only 640k memory, there just isn't enough when lots of stuff is running).

Instead, I think a user-mode limit of PATH_MAX (in libc) is the (at least for now) solution - because this would provide for a buffer length that user programs need to use to prevent Undefined Behavior when making system calls or accepting filename input etc. And would also provide a single symbol to search for as you mention, for future enhancements.

> But yes, even better would be just going for a real PATH_MAX in libc. I will toy with this and see where it goes.

[This article](https://stackoverflow.com/questions/9449241/where-is-path-max-defined-in-linux) talks about some issues of using <limits.h>, and if we use this, we need to be very careful with the `#include_next <limits.h>`, as I'm pretty sure I already tried this when writing `fm` and couldn't get it to reliably work. (The new limits.h would go in libc/include).

[This article](https://pubs.opengroup.org/onlinepubs/7908799/xsh/limits.h.html) describes some nice values to have in <limits.h>.

[This article](https://stackoverflow.com/questions/871952/gcc-build-problem-include-next-limits-h) describes some of the issues with using `#include_next` and build issues. Note that our libc uses a specially-prepared `make depend` so that would need to be specially tested.

A possible decent compromise would be to define `PATH_MAX` (and `MAXPATHLEN` and possibly `LINE_MAX`) in <sys/param.h>, which is a BSD-ism, but our libc supports that already. You'll find that sys/param.h already includes `MAXSYMLINKS`, so this could be a good place to put them. Note also the previous article states that POSIX doesn't actually state that <limits.h> has to include `PATH_MAX`. On the other hand, if we get can `#include_next` to reliably work, <limits.h> would be superior.

After adding the new defines, it would be nice to put a list together of the system calls that need to be guarded with PATH_MAX or LINE_MAX, like `readlink`. (Note that `fm`, because of its internal large memory usage and `realloc`, already redefines PATH_MAX to 80, as otherwise it sometimes runs out of memory. This is a good example of why we need to be careful setting PATH_MAX's value).

If you don't want to jump in, let me know, I'm happy to add this to my list.

Thank you!

PR #1551, opened 2023-03-04
[nano-x] Cleanup extra printf after failed GrOpen

ghaerr:
Removes display of "Can't open graphics" when `nxtetris` is opened without MOUSE_PORT=none set on system without mouse.

@toncho11, I think everything should be fixed now with regards to using `export MOUSE_PORT=` as discussed previously.

### DIFF
--- a/elkscmd/nano-X/demos/ntetris.c
+++ b/elkscmd/nano-X/demos/ntetris.c
@@ -744,12 +744,9 @@ void new_game(nstate *state)
 void init_game(nstate *state)
 {
 	GR_SCREEN_INFO si;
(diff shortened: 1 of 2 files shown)


toncho11:
Thank you @ghaerr 
When MOUSE_PORT=none then the mouse cursor should not be visible no matter if a serial card is present or not. Does this sound good to you?

ghaerr:
> When MOUSE_PORT=none then the mouse cursor should not be visible no matter if a serial card is present or not. Does this sound good to you?

Yes, ugh, I forgot to check that. Is that still not working?

PR #1554, opened 2023-03-05
[kernel] Rewrite memcpy in ASM for speed

ghaerr:
Using the recently added P)rofile command in [blink16](https://github.com/ghaerr/blink16), it was found that a significant amount of early time was spent in the kernel `memcpy` routine. Rewriting in ASM for speed. I have not tested the real instruction timings on various x86 processors, but think using `rep movsw` will be faster than the prior C language version.

Here's a screenshot showing `blink16`s profiling of a portion of the kernel boot sequence:

ELKS blink16 memcpy
This fix brings down `memcpy` to less than 0.43% (although `blink16` is not counting `rep movsw` instruction cycles, which means it's still likely quite inaccurate).

### DIFF
--- a/elks/arch/i86/lib/string.S
+++ b/elks/arch/i86/lib/string.S
@@ -108,7 +108,7 @@ memset:
 	shr     $1,%cx         	// copy words
 	rep			// while (cx) cx--, [es:di++] = al
 	stosw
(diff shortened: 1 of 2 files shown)


ghaerr:
Also rewrote `memset` to store words when possible.

PR #1555, opened 2023-03-07
[kernel] [libc] Fix: arguments to kill ( ) syscall were reversed in some places

tkchia:


ghaerr:
Hello @tkchia,

Nice catch. Looking a bit further into this, I notice that `pause()` isn't implemented by the kernel (it's supposed to wait until a signal has been received). That's ok though. However, FYI the man page for `abort` says that if SIGABRT is being caught, that handler is called. It also says that open streams are flushed and closed. It seems neither of these is occurring.

In addition, elks/kernel/sys.c has the following line in it which is also incorrect:
```
extern int sys_kill(sig_t,pid_t);
```
Can you push a fix for that, it's arguments are reversed. (No, there shouldn't be any `extern`s floating around kernel source, but it seems nobody's thought a proper header file to place to put this, so it also isn't being checked against the defined function).

Thank you!

tkchia:
Hello @ghaerr,

> Can you push a fix for that, it's arguments are reversed.

Done.

> However, FYI the man page for `abort` says that if SIGABRT is being caught, that handler is called.

I am quite sure that, when a `SIGABRT` signal is sent, then any signal handler installed for it will be called as a matter of course &mdash; or am I missing something?

> It also says that open streams are flushed and closed. It seems neither of these is occurring.

Regarding the closing of streams, I am guessing the man page (from GNU `libc`) is making extra promises beyond what is strictly needed.  POSIX.1-2017 merely says (https://pubs.opengroup.org/onlinepubs/9699919799/functions/abort.html),

> <sup>[CX]</sup> 🠶 The abnormal termination processing shall include the default actions defined for SIGABRT and may include an attempt to effect [__fclose()__](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fclose.html) on all open streams. 🠴

So it is probably OK for a `libc` implementation to try to close open `stdio` streams; but it is also OK if the `libc` does not try to close them.

Thank you!

tkchia:
Hello @ghaerr,

> I notice that pause() isn't implemented by the kernel (it's supposed to wait until a signal has been received).

On second thought, the call to `pause ()` seems a bit dubious.  It is possible that any `SIGABRT` signal will be processed and handled before `pause ()` is even called, so `pause ()` may end up waiting for _another_ signal, probably forever.  This should probably be fixed.

Thank you!

ghaerr:
Hello @tkchia,

> I am quite sure that, when a SIGABRT signal is sent, then any signal handler installed for it will be called as a matter of course 

What I am referring to is the first line of `abort()`, where SIGABRT processing is set to SIG_DFL, which assures that any signal handler for it will _not_ be called:
```
signal (SIGABRT, SIG_DFL);
```

> On second thought, the call to pause () seems a bit dubious. It is possible that any SIGABRT signal will be processed and handled before pause () is even called, so pause () may end up waiting for another signal, probably forever. This should probably be fixed

I had originally thought that also - but then realized the program would always be terminated by a SIGABRT signal after the first call to `kill(SIGABRT,...)`. This is because kernel `do_signal` is called after processing any system call. 

Actually, looking further into this, I see that `do_signal` processes signals in a low-to-high numbered fashion. That is, should there be another signal waiting to be processed below SIGABRT (SIGINT, SIGWINCH, etc), `do_signal` will process that signal and not the remaining signals. This won't happen now since SIGABRT will always kill the process without calling a registered handler, and `pause` (which incorrectly returns -ENOSYS) allows the kernel to call `do_signal` yet another time.

In summary, a `kill(SIGABRT,...)` should be called prior to  `signal(SIGABRT, SIG_DFL)` to allow a registered handler to execute, and `pause` should probably also be removed since it performs no useful purpose, and allows yet another signal to be processed should `kill(SIGABRT,...)` return. (This still doesn't solve the problem of having multiple queued signals lower than SIGABRT queued, which would require setting all those signals to SIG_DFL prior `kill(SIGABRT,...)`, which has its own problems since each system call would call `do_signal` for any queued signals).

Ignoring the currently hard-to-handle queued signal problems when `abort` is called stalling SIGABRT processing, it seems the following would be a better overall implementation of `abort`:
```
void
abort(void)
{
    /* call registered handler if any, or die if none */
    kill(getpid(), SIGABRT);

    /* no registered handler, die on SIGABRT */
    signal(SIGABRT, SIG_DFL);
    kill(getpid(), SIGABRT);

    /* should not be needed unless multiply queued signals less than SIGABRT present */
    _exit(255);
}
```

Thank you!

ghaerr:
Hello @tkchia,

In my discussion of `do_signal` in the previous post, after reading that routine yet again, I realize I incorrectly stated the operation of `do_signal` when processing a signal with a registered handler: in ELKS, there currently is no signal queuing once a registered handler is called. That is, after building the return user stack frame for a registered signal handler, the entire process signal mask is _cleared_! Thus, currently, there is no ELKS kernel concept of multiple signals remaining to be processed, only during the case of a registered handler being called.

Ultimately, this just means for our case here that if a signal lower than SIGABRT is present at the time of attempting to call the SIGABRT registered handler via `kill(SIGBART,...)`, the SIGABRT handler will _not_ be called, but instead the lower-numbered signal will be processed. Should that signal have a registered handler, the process signal mask will be completely cleared before its execution, and will result in the SIGABRT handler never being called.

In my example provided above, the second `kill` will always terminate the process on SIGABRT, should it be executed. At the end of the day, the current ELKS kernel can't guarantee that a SIGABRT handler will always be called, it seems, and even with my suggested abort rewrite there may be cases where the process aborts via a SIGABRT signal despite a registered handler present in certain race conditions.

Thank you!

PR #1557, opened 2023-03-09
[kernel] Allow multiple signals pending per process, enhance abort(), write test_sigfail test case

ghaerr:
Pursuant to discussion in https://github.com/jbruchon/elks/pull/1555#issuecomment-1460397198, this PR adds `test_sigfail` (available on 2880k+ images) which demonstrates how the ELKS kernel always clears the current process signal mask when calling a registered handler.

To demonstrate, run `test_sigfail`, which registers signal handlers for SIGINT and SIGABRT, then forks, with each process sending SIGINT or SIGABRT to only the first process as fast as possible. One can then type ^C multiple times, producing another SIGINT, which on occasion will either 1) terminate the program, or 2) cause the following kernel message to be displayed:
```
SIGNAL(pid) processing mask 0002, ignoring signal w/mask 0020 // 0002=SIGINT, 0020=SIGABRT
```
(Note the program will display "ABORT failed" continually - this is because `vfork()` doesn't actually share parent and child address space in the current ELKS implementation).

In the first case, the program terminates because ELKS requires re-registration of signal handlers each time a signal is received, and if a SIGINT is received between the time the handler callback is scheduled and the handler is executed, the second SIGINT will cause the kernel to terminate the program.

In the second more rare case, the program has received both SIGINT and SIGABRT signals before being scheduled, and the kernel displays the above message before clearing the remaining signal (in line 74 of signal.c).

In some cases, the kernel will display the following message:
```
SIGNAL(pid) processing mask 0002, ignoring signal w/mask 0002
```
This seems to indicate that the task has received a SIGINT while processing the SIGINT message but before returning to the modified user stack frame, thus indicating that the same signal remains to be processed. This may be why the kernel signal dispatch routine was written to originally clear the signal mask. I have also seen, but now cannot repeat, process STACK OVERFLOW messages when I changed the kernel to _not_ clear the signal mask. So the system needs more testing before changing signal mask processing.

Ctrl-\ (QUIT) can be pressed to terminate the program, sending a SIGQUIT.

The above testing varies radically depending on how many tasks are being scheduled by ELKS, e.g. whether or not an additional terminal may be logged in. It is recommended to run `test_sigfail` using serial console set in /bootopts, so that kernel messages can be seen outside of the ABORT fail messages.

The point of this was to confirm the conclusion in https://github.com/jbruchon/elks/pull/1555#issuecomment-1460397198 that 1) the kernel can't guarantee a registered SIGABRT handler to be called, and that 2) the suggested changes should be made in libc `abort`.

### DIFF
--- a/elks/arch/i86/drivers/char/ntty.c
+++ b/elks/arch/i86/drivers/char/ntty.c
@@ -67,6 +67,8 @@ int tty_intcheck(register struct tty *ttyp, unsigned char key)
     if ((ttyp->termios.c_lflag & ISIG) && ttyp->pgrp) {
 	if (key == ttyp->termios.c_cc[VINTR])
 	    sig = SIGINT;
+	if (key == ttyp->termios.c_cc[VQUIT])
+	    sig = SIGQUIT;
 	if (key == ttyp->termios.c_cc[VSUSP])
 	    sig = SIGTSTP;
 #if DEBUG_EVENT
--- a/elks/arch/i86/kernel/signal.c
+++ b/elks/arch/i86/kernel/signal.c
@@ -58,7 +58,7 @@ int do_signal(void)
 		      (SM_SIGQUIT|SM_SIGILL|SM_SIGABRT|SM_SIGFPE|SM_SIGSEGV|SM_SIGTRAP))
 		    dump_core();
 #endif
-		printk("SIGNAL terminating pid %d\n", currentp->pid);
+		debug_sig("SIGNAL terminating pid %d\n", currentp->pid);
 		do_exit(signr);				/* Default Terminate */
 	    }
 	}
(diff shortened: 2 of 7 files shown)


ghaerr:
The last push fixes the problem of the kernel clearing all signal masks instead of just the current mask. The problem was found to be that with interrupts enabled, the kernel may receive an additional signal of the type it is currently processing, while preparing the stack for return to user mode. The fix disables interrupts before resetting the current signal, until the IRET back to user mode. I am pretty sure this was the reason why stack overflow was occurring during some testing, as it could be possible for the kernel to continually setup user stack frames to return to if not clearing the entire mask.

Testing now shows that the kernel will call a registered signal handler only once for any specific signal received until the process is returned to user mode execution of the signal handler. Not clearing the process signal mask entirely now should guarantee that `abort` will call a registered signal handler and then exit on SIGABRT, once it is rewritten.

tkchia (approved):

tkchia:
Hello @ghaerr,

> The problem was found to be that with interrupts enabled, the kernel may receive an additional signal of the type it is currently processing, while preparing the stack for return to user mode.

Ah, cool. :slightly_smiling_face:  For a moment I was wondering why `do_signal ()` was clearing the appropriate signal bit twice (once at `currentp->signal ^= mask;` and once at `currentp->signal &= ~mask;`).

Thank you!

PR #1558, opened 2023-03-12
Fix UB in libc expandwildcards

ccoffing:
Undefined behavior occurred when a [] or ? did not match and there was not previous a * to backtrack to.

PR #1561, opened 2023-03-13
Fix libc warnings

ccoffing:
Fixing warnings as pointed out by https://github.com/jbruchon/elks/pull/1558#issuecomment-1465241879

I was using the wrong test macro (ASSERT_EQ is for integers; ASSERT_EQ_P is for pointers).

Also added const to some stdio functions, and some other minor warning fixes along the way.

### DIFF
--- a/libc/misc/popen.c
+++ b/libc/misc/popen.c
@@ -3,7 +3,7 @@
 #include <sys/wait.h>
 #include <paths.h>
 
-FILE *popen(char *command, char *rw)
+FILE *popen(const char *command, const char *rw)
 {
     int pipe_fd[2];
     int pid, reading;
--- a/libc/misc/popen.c
+++ b/libc/misc/popen.c
@@ -3,39 +3,42 @@
 #include <sys/wait.h>
 #include <paths.h>
 
-FILE * popen(char *command, char *rw)
+FILE *popen(char *command, char *rw)
 {
-   int pipe_fd[2];
-   int pid, reading;
+    int pipe_fd[2];
+    int pid, reading;
 
-   if( command == NULL || pipe(pipe_fd) < 0 ) return NULL;
-   reading = (rw[0] == 'r');
+    if (command == NULL || pipe(pipe_fd) < 0)
+        return NULL;
+    reading = (rw[0] == 'r');
 
-   pid = vfork();
-   if( pid < 0 ) { close(pipe_fd[0]); close(pipe_fd[1]); return NULL; }
-   if( pid == 0 )
-   {
-      close(pipe_fd[!reading]);
-      close(reading);
-      if( pipe_fd[reading] != reading )
-      {
-	 dup2(pipe_fd[reading], reading);
-         close(pipe_fd[reading]);
-      }
+    pid = vfork();
+    if (pid < 0) {
+        close(pipe_fd[0]);
+        close(pipe_fd[1]);
+        return NULL;
+    }
+    if (pid == 0) {
+        close(pipe_fd[!reading]);
+        close(reading);
+        if (pipe_fd[reading] != reading) {
+            dup2(pipe_fd[reading], reading);
+            close(pipe_fd[reading]);
+        }
 
-      execl(_PATH_BSHELL, "sh", "-c", command, (char*)0);
-      _exit(255);
-   }
+        execl(_PATH_BSHELL, "sh", "-c", command, (char*)0);
+        _exit(255);
+    }
 
-   close(pipe_fd[reading]);
-   return fdopen(pipe_fd[!reading], rw);
+    close(pipe_fd[reading]);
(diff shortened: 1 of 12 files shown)


libc/termcap/tputs.c
```diff
@@ -1,11 +1,8 @@
...
 
-#ifdef __linux__
-speed_t
-#else
-short
-#endif
-	 ospeed;
+#include <sys/types.h>
+
+speed_t ospeed;
```
ghaerr: Hello @ccoffing,

We can't accept this change, as other programs (e.g see elkscmd/elvis/curses.{h,c}) expect `extern short speed`, which seems to be part of the FreeBSD (and macOS) man pages for `tputs(3)`. It seems a bit of a mess, and the Linux man page for tputs has dropped all of this entirely, so perhaps the entire extern declaration should be removed entirely for Linux? (I haven't access for Linux curses.h or term.h to see if this variable exists anymore in Linux).

I think `speed_t` is the type for the non-encoded baud rate values (e.g. 38400 or 115200 which require 32 bits) versus the `ospeed` value which is used as the encoded (termios.c_flag & CBAUD) value (e.g. one of the B9600 etc values) and then used as an index in the baud rate divisor table in the kernel.

Thank you!

libc/misc/popen.c
```diff
@@ -5,37 +5,36 @@
...
-   {
-      close(pipe_fd[!reading]);
-      close(reading);
-      if( pipe_fd[reading] != reading )
-      {
-	 dup2(pipe_fd[reading], reading);
-         close(pipe_fd[reading]);
-      }
+    pid = vfork();
+    if( pid < 0 ) { close(pipe_fd[0]); close(pipe_fd[1]); return NULL; }
```
ghaerr: @ccoffing, when you re-tab functions for readability, what do you think of the idea of also reformatting them to an agreed-on ELKS standard? In this case, the above line in this function uses poor style, IMO. Do you have an indentation program that can be used when retabbing files that have ancient style? I pushed `elks/scripts/indent.sh` which can be used, I suggest that for this file since we're retabbing it entirely anyways. 

The same could also possibly be done for the other files touched for fixups, for instance libc/termcap/tputs.c whose formatting also seems ancient. In that file, it might be a good idea to update function argument decls to ANSI rather than K&R.

libc/termcap/tputs.c
ccoffing: I will just drop the commit with the #include changes.  It's outside the scope of this PR anyway.  It came from a branch I have with misc portability tweaks, so that I can build portions for modern Linux and use valgrind.  I didn't know that about `extern short speed`, thanks.

ghaerr:
Thanks for reformatting popen.c!

libc/misc/popen.c
ccoffing: I like using tools to format.  It's not perfect, but it's certainly nice to have the options checked into the source tree both to drive the tool and as documentation of what the standard even is.

That said, I first improved the formatting by hand (that's what's in the PR just merged) then ran indent using the options in the script (for my education).  The `indent` program on Debian Linux doesn't understand the `-ldi8` option.  But my FreeBSD box does understand it.  And the output differs slightly between the two (besides just due to `ldi`).  Interesting.

ghaerr: > The indent program on Debian Linux doesn't understand the -ldi8 option. But my FreeBSD box does understand it. And the output differs slightly between the two 

Geez, I didn't even think of the problem of multiple versions of indent acting differently with the same command line. I suppose we also need to describe the desired output format somewhere as well, as you mention. (The file `coding-style.txt` in the root directory is out of date, directly from a very old version of the Linux kernel over 20 years ago).

> besides just due to ldi

Actually, the local variable formatting is my least favorite part of the indent.sh script using -ldi8, as it indents the variable names after the type(s), which I'm not sure is great. But on my macOS system, I had to do something, or indent would output all local variables on column 0!

Do we need another version/option for Debian Linux? I'm open to other ideas, I just happened to have an indent spec from some of my other projects handy so I used that. I'm not that familiar with all the tools but I understand there may be some newer tools that are superior.

ccoffing: Regarding `-di8 -ldi8`, if I understand what you're saying, maybe what you want instead is `-di1` (and that looks better to me too).

I realized that the different behavior I was seeing, was due to GNU indent switching the defaults to their own style.  Since your indent.sh does not specify every setting, I was getting a mis-mash of settings.  And worse, GNU indent only has some shared settings with the original indent.  Forget I even confused the issue with GNU indent.

I suggest creating a `indent.pro` in the root directory of the project, with all settings explicitly listed (including, probably, `-di1`).  Then developers can copy it to `~/.indent.pro` or explicitly use `indent -Pindent.pro`.  And format from Mac or a BSD, or original indent.c built on Linux (which is what I did).  Works fine.

PR #1566, opened 2023-03-18
[libc] limits.h with PATH_MAX

ccoffing:
Added PATH_MAX to a new libc limits.h.  The build logs before and after this change have the same warnings.

Grepping for PATH_MAX or NAME_MAX is now a map to a goldmine of buffer overflows, for better or worse.

Is it acceptable to include linuxmt/config.h from limits.h to get some kernel values, or should these just be hardcoded and kept in sync manually?  I could also include linuxmt/fs.h to automatically set OPEN_MAX.  But I suppose there is the risk of namespace pollution.  For the moment, I have not done this due to possible pollution.

Notes on some specific changes:
- I did not update elks/tools/mfs/ and elks/tools/mfsck/ which have NAME_MAX set to 255.  Seems excessive but maybe there is a reason.
- mined claims to truncate the filename, which I changed to proper NAME_MAX, but I think the truncation is buggy anyway.
- mail builds its own short paths, so I left it with its own smaller buffer size.

### DIFF
--- a/elkscmd/ash/autocomplete.c
+++ b/elkscmd/ash/autocomplete.c
@@ -11,6 +11,7 @@
 #include <sys/stat.h>
 #include <stdio.h>
 #include <string.h>
+#include <limits.h>
 #include "linenoise.h"
 #include "output.h"
 
@@ -20,7 +21,7 @@ void completion(const char *buf, linenoiseCompletions *lc)
 	DIR *fp;
 	struct dirent *dp;
 	int count;
-	char dir[128];
+	char dir[PATH_MAX];  /* larger than NAME_MAX since this is user-supplied */
 	char line[LINENOISE_MAX_LINE];
 	char result[LINENOISE_MAX_LINE];
 	char lastentry[LINENOISE_MAX_LINE];
@@ -69,7 +70,7 @@ void completion(const char *buf, linenoiseCompletions *lc)
 		return;
 	if (count == 1) {
 		struct stat st;
-		char path[128];
+		char path[PATH_MAX];
 
 		fmtstr(path, sizeof(path), "%s/%s", dir, lastentry);
 		if (!stat(path, &st)) {
@@ -89,7 +90,7 @@ void completion(const char *buf, linenoiseCompletions *lc)
 	while ((dp = readdir(fp)) != NULL) {
 		if (!strncmp(file, dp->d_name, strlen(file))) {
 			struct stat st;
-			char path[128];
+			char path[PATH_MAX];
 
 			if ((file[0] == '.') || (strcmp(dp->d_name, ".") && strcmp(dp->d_name, ".."))) {
 				strcpy(result, buf);
--- a/elkscmd/ash/redir.c
+++ b/elkscmd/ash/redir.c
@@ -62,7 +62,6 @@ static char sccsid[] = "@(#)redir.c	5.1 (Berkeley) 3/7/91";
 
 
 #define EMPTY -2		/* marks an unused slot in redirtab */
-#define PIPESIZE 4096		/* amount of buffering in a pipe */
 
 
 MKINIT
@@ -232,7 +231,7 @@ openhere(redir)
 		error("Pipe call failed");
 	if (redir->type == NHERE) {
 		len = strlen(redir->nhere.doc->narg.text);
-		if (len <= PIPESIZE) {
+		if (len <= PIPE_BUF) {
 			xwrite(pip[1], redir->nhere.doc->narg.text, len);
 			goto out;
 		}
--- a/libc/include/limits.h
+++ b/libc/include/limits.h
@@ -5,10 +5,13 @@
 #define PATH_MAX 128
 
 /* Maximum number of bytes in a filename, not including terminating null. */
-#define NAME_MAX 14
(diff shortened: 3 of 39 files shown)


ghaerr:
Hello @ccoffing,

Nice, thank you for this, very well done! This is a big step forward in obtaining/proving correctness with our ELKS applications.

> Grepping for PATH_MAX or NAME_MAX is now a map to a goldmine of buffer overflows, for better or worse.

I'd say definitely for the better. A quick look shows that many are unprotected using `strcpy`, but at least now they'll be marked and easy to find/fix in later PRs.

> Is it acceptable to include linuxmt/config.h from limits.h to get some kernel values, or should these just be hardcoded and kept in sync manually?

Including config.h in particular is a problem, as it can unnecessarily cause large rebuilds during development when switching between .config files. A couple of libc files used to include it, and due to the `make depend` that is part of the libc build, this was causing the C library to be rebuilt even when changing .config for output disk image size changes, which I finally tracked down and eliminated.

> I could also include linuxmt/fs.h to automatically set OPEN_MAX.

At this point, I think we're better off not including kernel headers unless absolutely necessary. Those headers do support an `#ifdef __KERNEL__` to hide certain definitions, but aren't complete and still end up allowing access to structures that user programs shouldn't probably see below the defined API layer, as well as namespace pollution as you mention. Some C library files do include kernel headers to avoid structure duplication though. I would suggest we go with manual syncing for now, and try to keep limits.h small. 

You might indicate in limits.h where the value comes from (if anywhere) for easier maintenance.

> I did not update elks/tools/mfs/ and elks/tools/mfsck/ which have NAME_MAX set to 255. Seems excessive but maybe there is a reason.

Agree on no change: these files are used in programs built to run on the host rather than ELKS, so it is easily possible NAME_MAX could be larger than 14.

I have noticed some other problems, I will address these in separate comments.

Thank you!

elkscmd/ash/autocomplete.c
```diff
@@ -20,7 +21,7 @@ void completion(const char *buf, linenoiseCompletions *lc)
 	DIR *fp;
 	struct dirent *dp;
 	int count;
-	char dir[128];
+	char dir[NAME_MAX];
```
ghaerr: It seems that `dir[]` gets its input from sources other than `readdir`, and copying to it isn't yet checked. I would suggest we keep `dir[]` at PATH_MAX for now (along with an explanatory comment) until we can show or know it can be made smaller.

elkscmd/ash/redir.c
```diff
@@ -232,7 +231,7 @@ openhere(redir)
 		error("Pipe call failed");
 	if (redir->type == NHERE) {
 		len = strlen(redir->nhere.doc->narg.text);
-		if (len <= PIPESIZE) {
+		if (len <= PIPE_BUF) {
```
ghaerr, line 234: Nice catch on ELKS' max pipe buffering size. I hadn't realized the shell would fork in the case a 'here document' greater than the max writable pipe buffer. However, the kernel will still accept pipe writes greater than the pipe buffer size by sleeping the process and waiting for the reader to deplete the buffer. I haven't checked enough to see whether this might deadlock the shell or not, assuming that ash is likely also the reader (?). Thus, we should probably test this change with a small and large here document, perhaps in the test suite. 

Thus, this is either pointing out a previously unknown bug in the shell for ELKS, or likely running new untested fork code for the shell. The problem with the shell forking is that its large and there may not be space for the extra data segment (text segment is shared).

libc/include/limits.h
```diff
@@ -0,0 +1,16 @@
+#ifndef __LIMITS_H
+#define __LIMITS_H
+
+/* Maximum number of bytes in a pathname, including the terminating null. */
+#define PATH_MAX 128
+
+/* Maximum number of bytes in a filename, not including terminating null. */
+#define NAME_MAX 14
```
ghaerr: We're going to need NAME_MAX to be 26 for the FAT filesystem. See linuxmt/dirent.h. `(readdir` returns from this structure). ELKS' implementation of the FAT filesystem allows for VFAT (long filename) entries with up to two dir entries (13 bytes each), which ends up allowing for 26 bytes total for FAT (and 14 for MINIX).

libc/include/limits.h
```diff
@@ -0,0 +1,16 @@
...
+
+#define PIPE_BUF 80
+
+#define OPEN_MAX 20
```
ghaerr, line 15: Lets indicate OPEN_MAX comes from linuxmt/fs.h in a comment, along with linuxmt/config.h for PIPE_BUF. I know see why you're talking about including config.h. I think PIPE_BUFSIZ got rammed into config.h as a tunable parameter, but thinking again, it doesn't belong there, and probably ought to be moved to linuxmt/fs.h.

elkscmd/ash/redir.c
ccoffing: I have previously tested very large here documents, and just re-tested ~300 bytes, and I am able to see that it forks.  I have never been able to cause a deadlock.

elkscmd/ash/autocomplete.c
ccoffing: fixed

ghaerr:
Thanks Chuck!

PR #1569, opened 2023-03-22
Fix IRCv3 PING/PONG connection issue

### LINKED ISSUE #1568: TinyIRC does not comply with PING/PONG and thus will not connect to anything on ELKS
**Description**

TinyIRC does not comply with PING/PONG transmissions.
- What is expected ?
Upon login at say, UnrealIRCd TinyIRC should acknoledge the connection by using PING + PONG commands for connection verification and keepalive.
- What is observed ?
It does not at all and thus cannot connect unless the server is old enough to not need this.

**Configuration**
- Version (commit or release):
Built from source yesterday, /2029e799b079c722a6bc4847911eb649021be580
- Emulator or HW platform:
V20, 640K, 64MB CF, VGA.

**How to reproduce ?**
- Does the problem always happen ?
Yes
- List the steps to reproduce the problem.
1. Transfer TinyIRC from compiled sources since it does not get loaded into the HD image.
2. Try to connect say to an UnrealIRCd platform

**Raw data**
- Screenshot(s):
image
- Log file:
None

hiirotsuki:
IRCv3 specification sends an unique token on each PING which it requires be sent back by replying with PONG.

tested on libera.chat (non IRCv3 PING/PONG) and unrealircd (hybrid, one IRCv3 PING packet followed by original PING packets)

Fixes #1568

### DIFF
--- a/elkscmd/inet/tinyirc/tinyirc.c
+++ b/elkscmd/inet/tinyirc/tinyirc.c
@@ -511,7 +511,8 @@ int parsedata()
     int i, found = 0;
 
     if (serverdata[0] == 'P') {
-	sprintf(lineout, "PONG :%s\n", IRCNAME);
+	memcpy(lineout, serverdata, sizeof(serverdata));
+	lineout[1] = 'O';
 	return sendline();
     }
     if (!dumb)


ghaerr:
Wow @Redfoxymoon, thanks a lot! I can't test this but others will. Sounds like we don't need encrypted communications after all.

With regards to the heap space problem also shown in #1568, I'll prepare a separate PR for that.

DutchComputerKid:
Works! TinyIRC has more problems than this but at least people can actually connect and chat now, excellent!
(Yes I tested it on my V20 and it works)

toncho11:
Great! I should test it when I have the time.
Contrary to what I thought it seems IRC is usually unencrypted : https://news.ycombinator.com/item?id=33481692
Thank you!

PR #1572, opened 2023-03-24
[kernel] Remove CAPS_DRIVE_PARMS for NuXT and other 8086 systems

ghaerr:
Fixes BIOS floppy drive problem on NuXT system, discussed in https://github.com/jbruchon/elks/discussions/1370#discussioncomment-5415011.

Also fixes FAT filesystem improperly reading root directory when no /DEV entry exists (which is always the case for DOS disks not created by ELKS make).

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -402,16 +402,14 @@ static unsigned short int INITPROC bioshd_getfdinfo(void)
 	}
     }
 #else
-    /* Use INT 13h function 08h normally if PC/AT or higher*/
-    if (sys_caps & CAP_DRIVE_PARMS) {
-	BD_AX = BIOSHD_DRIVE_PARMS;
-	BD_DX = 0;			/* query floppies only*/
-	BD_ES = BD_DI = BD_SI = 0;	/* guard against BIOS bugs*/
-	if (!call_bios(&bdt))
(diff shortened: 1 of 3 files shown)


ghaerr:
@toncho11, this PR should also run on your older IBM XT systems. I'm going to commit this prior to your testing, as the PR checks whether the BIOS call is allowed, and should still work when the call is not implemented. In the previous version, we only tried calling the function if the CPU was a 286 or newer, which is always the case in PC/AT systems.

[This link](https://stanislavs.org/helppc/int_13.html) talks about the technical issues behind when INT 13h AH=8 get disk params is actually implemented. They are saying it is in fact implemented on older PCs, but is implemented in the replacement hard disk BIOS. Nonetheless, this PR should handle the case of when the function is not present.

toncho11:
Yes, it detects the floppy OK on IBM XT. At least on this version of the BIOS.

toncho11:
Still I would like to note that it takes 7 seconds after "kernel text" and before printing the info for the floppy. 
I think it was not that slow before.

ghaerr:
> Still I would like to note that it takes 7 seconds after "kernel text" and before printing the info for the floppy.

I would guess the reason for that is the this version asks the BIOS for how many floppy drives are attached, and the BIOS attempts to read from the second floppy, causing a delay.

In the future, we could add a "syscaps=" to /bootopts to tell ELKS more exactly what the system capabilities are, but ultimately it's probably better that the system just work on more systems without configuration.

Issue #1589, opened 2023-04-19
Support for UMB memory


mirevall:
Hi,

I discovered elks recently and, as a fan of both very old pc compatibles and linux, it is like a dream come true for me ! Thanks for the great work ! It works wonderfully on my Ericsson Portable PC that has a 3c509 ethernet card.  -> https://imgur.com/a/EOXEkFh
Unfortunately this computer only has 512k of ram, which is not much when using the ethernet daemons, and I am thinking of buying a lo-tech 1Mb card to upgrade it.

I came across issue #709 where you were discussing UMB support with details on how to implement it, and I don't think it was done in the end. I gave it a shot and came up with this -> https://github.com/mirevall/elks
It seams to work well, on 86Box at least, as I don't have any way to test it on real hardware for now. -> https://imgur.com/a/h4ILdU5

Right now the segments to add are hard coded in malloc.c which is of course very temporary, but I don't know the best way to handle this :
- new config file in /etc?
- new entry in /bootops ?
- compile time config ?

To be honest I haven't looked at the feasibility or complexity of any of it, I wanted to have your opinion on this (or even if UMB support is worth adding, but given how little work is required I think it is)

ghaerr:
Hello @mirevall,

Pretty cool running ELKS on your Ericsson Portable PC, along with networking. Thanks for the pictures! I'll bet it is quite low on RAM with everything running.

> I came across issue https://github.com/jbruchon/elks/issues/709 where you were discussing UMB support with details on how to implement it, and I don't think it was done in the end. I gave it a shot and came up with this -> https://github.com/mirevall/elks

No, we never implemented it. Your implementation looks like a great start, as all we really need to do is insert the memory into the kernel mm freelist, which it looks like you've done. I haven't fully studied your code yet, but it seems we'll also need the ability to deallocate RAM back into the freelist, as well as allow any size allocations within a UMB block area. I'm not sure whether your code does that or not yet.

(Another idea would be, that instead of "adding" main memory and then UMB memory into the freelist, we add a contiguous 1M RAM into the freelist, then "subtract" memory by permanently allocating "holes" where the EGA video or BIOS is (but not the UMB memory). In that way, the memory manager itself doesn't have to be rewritten at all, but instead just skips over permanently allocated (EGA/video) segments whilst traversing all of memory. For systems without UMB, the initial freelist can just be `start/end` as is currently passed to `mm_init`).

[...]

PR #1590, opened 2023-04-21
UMB memory support

mirevall:
This pull request adds support for UMB memory in the form of a new line in /bootopts :
```umb=0xC000:0x800,0xD000:0x1000``` for example will allow ELKS to use memory situated from 0xC0000 to 0xC8000 and from 0xD0000 to 0xE0000. 
The usual limitations of UMB still applies : memory needs to actually be here and must not conflict with anything else like bios or video memory.
This should also work to add conventional memory that the bios can't see (if the bios is hard-coded for the specific amount of memory the computer comes with)

This PR also fixes a bug revealed when using the UMB memory (see #1589)

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -141,31 +141,26 @@ void seg_free (segment_s * seg)
 	list_s * i = &_seg_free;
 
 	// Try to merge with previous segment if free
-
 	list_s * p = seg->all.prev;
-	if (&_seg_all != p) {
-		segment_s * prev = structof (p, segment_s, all);
-		if (prev->flags == SEG_FLAG_FREE && prev->base + prev->size == seg->base) {
-			list_remove (&(prev->free));
-			seg_merge (prev, seg);
-			i = _seg_free.prev;
-			seg = prev;
-		} else {
-			seg->flags = SEG_FLAG_FREE;
-			seg->pid = 0;
-		}
+	segment_s * prev = structof (p, segment_s, all);
+	if ((prev->flags == SEG_FLAG_FREE) && (prev->base + prev->size == seg->base)) {
+		list_remove (&(prev->free));
+		seg_merge (prev, seg);
+		i = _seg_free.prev;
+		seg = prev;
+	} else {
+		seg->flags = SEG_FLAG_FREE;
+		seg->pid = 0;
 	}
 
 	// Try to merge with next segment if free
 
 	list_s * n = seg->all.next;
-	if (n != &_seg_all) {
-		segment_s * next = structof (n, segment_s, all);
-		if (next->flags == SEG_FLAG_FREE && seg->base + seg->size == next->base) {
-			list_remove (&(next->free));
-			seg_merge (seg, next);
-			i = _seg_free.prev;
-		}
+	segment_s * next = structof (n, segment_s, all);
+	if ((next->flags == SEG_FLAG_FREE) && (seg->base + seg->size == next->base)) {
+		list_remove (&(next->free));
+		seg_merge (seg, next);
+		i = _seg_free.prev;
 	}
 
 	// Insert to free list head or tail
@@ -343,11 +338,6 @@ void INITPROC mm_init(seg_t start, seg_t end)
 {
 	list_init (&_seg_all);
 	list_init (&_seg_free);
-
-	//temporary
-	seg_add(0xC000, 0xC800);
-	seg_add(0xD000, 0xF000);
-	//!temporary
 
 	seg_add(start, end);
 }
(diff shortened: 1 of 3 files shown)


ghaerr:
Hello @mirevall,

This is some very impressive work adding UMB support that's been wanted for a long time, with so little code. Well done.

Thank you!

toncho11:
So here is  one memory  map:
memory

ghaerr:
@toncho11, the UMB memory tested for this PR, as well as the sample /bootopts command line in #1591, match the Free sections in your memory map.

Thank you!

toncho11:
I added a wiki page: https://github.com/jbruchon/elks/wiki/Access-to-UMB-memory
Feel free to update it.

mirevall:
Thank you @toncho11 for the page !

I didn't know about the MSD software, it seems useful !

I think we need to clarify the difference between free space and usable ram. There is free space in the UMB area, but to use it you need to physically add RAM that maps in this space (using a card like [this one](https://www.lo-tech.co.uk/wiki/Lo-tech_1MB_RAM_Board)) and only then can you specify it in /bootopts and use it. Otherwise ELKS will try to write in a space where there is nothing, and it won't work. 

Some modern hardware/emulators automatically maps memory in the empty areas (I believe qemu does)

Sorry if the description in my PR wasn't very clear, UMB is pretty complicated to explain clearly and English is not my first language.

toncho11:
So not all of the 384kb are physically present?

mirevall:
The 384k are reserved for things other than main memory such as the BIOS rom, BIOS extention roms, video memory, etc... On a standard IBM pc (and most clones), there is no system ram in this space, only in the first 640k of the address space (conventional memory).

So in the spaces maked "free" in the memory map, there is no ram, in fact there is nothing at all. But you have the possibility of adding RAM, since nothing else is using the space.

toncho11:
There is no physical RAM because the BIOS and BIOS extensions are mapped as if they were at these addresses above 640kb?
But then on newer computers such as 286 and 386 there is real RAM?

mirevall:
For compatibility reasons, the bios and the video ram is still between 640k and 1Mb, even on 286 or 386 systems. On those systems, the ram between 640k and 1Mb is hidden so as not to interfere with the bios/video ram, so the processor can't access it at all, even when specifying umb in ELKS.

The umb addition is less useful than it seems, and only works on very specific systems or with add-on cards. Most systems even with multiple megabytes of ram are only able to use 640k of it in ELKS.

tyama501:
Hello @mirevall , @ghaerr , @toncho11 ,

I could increase the free memory from 337KB to 401KB by moving 64KB disk buffer to UMB D000,
on PC-9801UV21(V30).

This machine does not have on-board umb, so I used special freeware to set EMS board as umb prior to boot the ELKS 0.7.0.

Thank you for adding the great feature.

PC-9801UV21_ELKS_UMB

ghaerr:
Hello @tyama501,

That is very cool having the external buffer stuck right into UMB memory.

I am wondering, since there isn't currently a way to specify where the system buffers go (just how many of them) - I am guessing that the reason this works is that the kernel memory allocator uses a "best fit" when selecting a location to allocate. That means that since you have configured for 64 EXT buffers and the UMB is 64k, it selects the UMB location for them. (The UMB configuration does allow for specifying where its memory is, but that's separate from the EXT buffer locations).

To test this, if you were to recompile or set /bootopts with say, 32 EXT buffers, I am wondering whether this would still use half of the UMB memory, or put them somewhere else. Given that the EXT buffer allocation code works by splitting allocations into 64k-chunks, I am guessing that if you allocated 96 buffers, 64 would go to UMB and the 32 would then be right after the kernel data segment.

If this all works, then there may not be a need to specify the location of system buffers. Note that with 0.8.0-dev, the "INITPROC" kernel section of approximately 7-8k bytes is also freed up after kernel init, and can be used for programs. (My system places the always-resident /bin/init there). The init section segment address is shown at boot using `init:13c1` or similar, and the data section is `data:15ee` or similar. Given that the kernel data section is always 64k (1000 hex for segment addresses), `meminfo` would show main memory allocations starting either at `13c1` or `25ee` given the examples.

Thank you!

tyama501:
Hello @ghaerr ,

Here is the meminfo for setting bufs=32 and bufs=96 on the same PC-98 and ELKS 0.7.0.

bufs=32, 32KB is still in the UMB.
PC-9801UV21_ELKS_UMB_32KBbuf

bufs=96, 64KB is in UMB and 32KB in the main memory. 
PC-9801UV21_ELKS_UMB_96KBbuf

Interesting :)

ghaerr:
Hello @tyama501,

Thanks for the report. It appears that the combination of the buffers being allocated in 64k chunks combined with the memory allocator using best fit works quite well for UMB buffers. (I also notice that /bin/init got placed in UMB in the 32k buffers case, which is nice).

So, it seems no action needs to be taken to achieve great results, but it remains important to possibly consider the number of EXT buffers along with the size of any UMB (and possibly /bin/init since its always resident) for best use of memory.

Thank you!

PR #1604, opened 2023-05-22
[8018x] add romprg to flash memory chips thru serial

cocus:
This adds support for a port of the "romprg" project, which enables ELKS to be used as an in-system flash programmer for the same memory that was used to run it. It also provides a PC-side tool that is the "command line tool" for the "romprg" running on the ELKS system.


cocus:
ah, it's failing because it can't find the FCPU of the 8018x... how should I make this thing not even build for this particular target?

ghaerr:
> how should I make this thing not even build for this particular target?

Perhaps add this to elkscmd/Make.install:
```
ifdef CONFIG_ARCH_8018X
ifdef CONFIG_APP_ROMPRG
 	$(MAKE) -C romprg install
endif
endif
```
[EDIT: Oops - you'll also have to remove romprog from SUBDIRS= in elkscmd/Makefile and add it back in using SUBDIRS +=
in the case of 8018X. Another idea would be to adjust the all: entry point in elkscmd/romprog/Makefile].

Lets cleanup elksmcd/romprog/Makefile and remove the stuff from basic such as lines 17, 21-26, etc.

Also, for now I think it best to remove $(HOSTPROGS) from all:, until we find time to test the host build which may not work on macOS etc. You can create that tool my chdir to elkscmd/romprog and `make hostromprog` for now.

cocus:
This actually failed on the main target of the makefile, not on the host version.
```
ia16-elf-gcc -fno-inline -melks-libc -mcmodel=small -mno-segment-relocation-stuff -mtune=i8086 -Wall -Os -Wno-implicit-int -Wno-parentheses  -I/home/runner/work/elks/elks/include -I/home/runner/work/elks/elks/libc/include -I/home/runner/work/elks/elks/elks/include -D__ELKS__ -DELKS_VERSION=\"0.6.0\" -Wno-maybe-uninitialized -c -o main.o main.c
```
This is because this application requires a config value that's only set on 8018x.



> Lets cleanup elksmcd/romprog/Makefile and remove the stuff from basic such as lines 17, 21-26, etc.
> 
> Also, for now I think it best to remove $(HOSTPROGS) from all:, until we find time to test the host build which may not work on macOS etc. You can create that tool my chdir to elkscmd/romprog and `make hostromprog` for now.

Good point!

ghaerr:
See my [EDIT] above; yes, I'm aware of the failure reason, but was suggesting that hostromprog also not be built by default.

cocus:
I set the ROMPRG to be only available for 8018x for now, as the 8018x requirement is really hardcoded into it.

ghaerr:
Thanks @cocus!

PR #1621, opened 2023-07-21
[console] Add DisplayCursor On/Off to AnsiCmd

tyama501:
This PR is adding Cursor On/Off by AnsiCmd as we discussed on https://github.com/jbruchon/elks/discussions/1618#discussioncomment-6406993

I checked with PC-9801BX and it worked.

I also added empty function in console-bios.c and console-direct.c for IBM.

Thank you.


ghaerr:
Thank you @tyama501!

PR #1622, opened 2023-07-23
Cron fixes

ccoffing:
Proofreading pass of cron and crontab.

I will have a follow-up PR at some point of wiring up MAILTO.

### DIFF
--- a/libc/system/sleep.c
+++ b/libc/system/sleep.c
@@ -39,12 +39,14 @@ sleep(unsigned int seconds)
 unsigned int
 sleep(unsigned int seconds)
 {
- struct timeval timeout;
- time_t start = time((void*)0);
- timeout.tv_sec = seconds;
- timeout.tv_usec = 0;
- select(1, NULL, NULL, NULL, &timeout);
- return seconds - (time((void*)0) - start);
+    struct timeval timeout;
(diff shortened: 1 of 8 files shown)


libc/system/sleep.c
```diff
@@ -39,12 +39,14 @@ sleep(unsigned int seconds)
...
- select(1, NULL, NULL, NULL, &timeout);
- return seconds - (time((void*)0) - start);
+    struct timeval timeout;
+    timeout.tv_sec = seconds;
+    timeout.tv_usec = 0;
+    time_t start = time((void*)0);
+    select(1, NULL, NULL, NULL, &timeout);
+    time_t stop = time((void*)0);
+    time_t elapsed = stop <= start ? 0 : stop - start;
+    return elapsed > seconds ? 0 : seconds - elapsed;
```
ghaerr, line 49: I'm not sure of the need for comparing `stop <= start` above: while technically that could happen here once in the year 2106 when the 2^32 time value in seconds wraps, even just using using an `elapsed = stop - start` would result in the function returning 0 correctly, since in that case any wrapped unsigned long elapsed value would be larger than seconds, which is unsigned int.

Perhaps consider changing to `time_t elapsed - time(0) - start` in a later update.

PR #1623, opened 2023-07-23
sync libc's testlib with external repo

ccoffing:
This fixes a few warnings, and makes it a bit more portable outside of just ELKS, and gets the formatting a bit more in line with the ELKS standard.  Also explicitly states that I'd like my original code to be under 0-clause BSD license, which is basically a lawyer-friendly public domain, because I hate the endless proliferation of license notices.

I am working on that 16-bit friendly word processor on the side; this comes from that effort.


ghaerr:
Thank you @ccoffing!

PR #1628, opened 2023-07-29
Install crontab

ccoffing:


ghaerr:
Thanks for your contributions @ccoffing!

I just noticed that `crontab` is rather large (at 14k) for its relative worth... and after this the 1440k floppy image will have just 20k free space. This (and other less-used apps) would be good candidates as compressed executables. (Compressed executables can be 30-40% smaller, but at the time of initial development there were some comments about them being slower to decompress on very slow systems than just doing the disk I/O). There isn't currently a way to specify individual apps to be compressed, it's either all or none - this can be something to consider adding in v0.8.

ccoffing:
I'm always on the lookout for ways to cut bloat (but compressed executables aren't my first choice).

How hard would it be to make the app selection more sophisticated, so that it could select applications not just individually, and not just by image side, but by use case?  For example, "interactive desktop" versus "embedded".

Separately, but related, I am working on adding mail support in a few areas (mail cron job results, and mail notifications in ash) so that there is actually a use for the `mail` port.  This seems like a use case for adding a configuration knob, such as `CONFIG_LOCAL_MAIL`, which can strip out this bloat for those who don't want it.  It works well so far.  And this causes me to look for other cross-cutting features which people might want the ability to trim out with a single configuration knob.

ccoffing:
Oh, I just re-read your comment.  Compressing only less-used apps could make some sense.  But it also just occurred to me that cron and crontab share about half their code, so building a single executable and installing a hard link is another way to save space in this case.

Anyway, yeah, those are all options for 0.8.

ghaerr:
> building a single executable and installing a hard link is another way to save space in this case.

That's a great idea! I hadn't realized they shared so much code. The combined functionality could also be controlled via command line options which would release the requirement for a hard link (a FAT distribution can't support them anyways). If `cron` with no arguments acted like `crontab`, then a new option `--daemon` could be added to run a background cron, for instance. Unfortunately, the man pages don't fit on the 1440k floppy (at least at this point; they might should we compress all executables).

> How hard would it be to make the app selection more sophisticated, so that it could select applications not just individually, and not just by image side, but by use case? For example, "interactive desktop" versus "embedded".

We kind of have that now with configuration options like CONFIG_APP_FILE_UTILS, etc but it still falls short because the options don't take image size into account. And the image-sized options like CONFIG_IMG_FD1440 don't take use case into account. More thought is needed, but at the end of the day, a 1440k distribution is by far the most commonly used for initially trying out ELKS, it seems, so IMO fitting as much as possible in a default build is better.

>  I am working on adding mail support in a few areas (mail cron job results, and mail notifications in ash) so that there is actually a use for the mail port. This seems like a use case for adding a configuration knob, such as CONFIG_LOCAL_MAIL, which can strip out this bloat for those who don't want it.

At this point, I would recommend using CONFIG_APP_MAIL for installation of `mail`, since it more closely aligns with the existing design. However, we have been running into trouble when apps need to be rebuilt based on image options (like CONFIG_APP_anything), which greatly complicates building the system for multiple image sizes. Depending on the included extra size, it still might be better to have mail features included by default in ash and cron, with just a local per-app method of setting internal options (much like ash and sash do now in their .h configuration file). Then everything just works, if one has the mail executable, rather than having to get into source code and rebuild the system.

PR #1629, opened 2023-07-30
Free space on 1440

ccoffing:
Saves about 58k on the 1440k image, saves more on larger images.


ghaerr:
Wow, nice catch on removing the `-O` optimization override of `-Os` in `vi` and saving 7k!!

Everything looks good, but I don't quite understand the savings of 52k on 1440k images, as neither CONFIG_APP_TEST nor CONFIG_APP_MAN_PAGES are set by default in the 1440k standard build (`ipmpc-1440.config`), and before this fix it has only 18k free. Are you able to fit the man pages or tests on a 1440k build?

ghaerr:
@ccoffing: Oops, I get it now: removing `float` saves 51k. Thank you!!

PR #1635, opened 2023-08-03
[release] Update ELKS version number to v0.7.0

ghaerr:
Proposed final commit for v0.7.0.


toncho11:
Looks good. I propose that the next version after the release is called `0.8dev`.

ghaerr:
> I propose that the next version after the release is called 0.8dev.

Would you believe that adding any letters after the version number, and/or not following the strict X.Y.Z numbering scheme requires some major re-work in the ELKS Makefiles? Yes, I plan on doing that, but need to get that work done first and want to see if there are any release issues requiring a v0.7.1. I don't think we're at the point of worrying about maintaining separate release and development branches.

I also have a bunch of new code I'll be posting shortly that adds kernel consistency checking which will greatly help kernel development and testing of dynamic task structures, which I'd like to get done in the next release.

PR #1638, opened 2023-08-06
Better guidance to run slattach.sh as root

floriangit:
When run as user, the error messages from slattach & co are almost always missing my eyes, thus this enhancement to enforce the script to be run as root.


ghaerr:
Thanks @floriangit!

PR #1641, opened 2023-08-08
[kernel] Determine kernel stack usage when allowing async I/O

ghaerr:
The current ELKS kernel tries to save kernel stack space by declaring some of its relatively larger stack variables and buffers as `static`, taking advantage of the fact that when block I/O is guaranteed synchronous, kernel functions that call filesystem code won't block, and thus won't be reentered by another process/system call.

This PR sets up a test capability (setting CONFIG_ASYNCIO) to allocate these stack variables as non-static declarations, and then measure the resulting increased kernel stack usage as a result, using the recently-introduced `kstack` tracing. A shell script is run producing heavy filesystem activity on MINIX and FAT filesystems (the MINIX filesystem doesn't use any larger stack variables/buffers).

The preliminary maximum kernel stack usage results found using static vs non-static for MINIX and FAT are:
```
       static reentrant  filesystem
kstack 386 -> 410        MINIX
kstack 500 -> 598        FAT
```
The largest kernel stack was reported during a `stat` system call:
```
KSTACK(149) sys_stat    max 598 prevmax 382 sysmax 598 (OVERFLOW AT 640)*
```
The preliminary conclusion is that an extra 100 bytes of stack are required to support async I/O, when FAT is used.
Trying to understand the actual maximum kernel stack required per task is tricky, and timing dependent. The current kernel stack allocation is 640 bytes on IBM PC, and 740 on PC-98. The PC-98 seems to require more stack during BIOS calls, and all BIOS calls are made from the kernel stack.

This PR also adds a `-f` option (force no error result) to `rmdir`, as was needed for a test shell script.

### DIFF
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -46,11 +46,6 @@
 
 #include <arch/segment.h>
 
-static struct minix_exec_hdr mh;
-#ifdef CONFIG_EXEC_MMODEL
-static struct elks_supl_hdr esuph;
-#endif
-
 #ifndef __GNUC__
 /* FIXME: evaluates some operands twice */
 #   define add_overflow(a, b, res) \
@@ -74,7 +69,7 @@ static struct elks_supl_hdr esuph;
  * Only IA-16 segment relocations are accepted
  */
 static int relocate(seg_t place_base, lsize_t rsize, segment_s *seg_code,
-		    segment_s *seg_data, struct inode *inode, struct file *filp)
+               segment_s *seg_data, struct inode *inode, struct file *filp, size_t tseg)
 {
     int retval = 0;
     __u16 save_ds = current->t_regs.ds;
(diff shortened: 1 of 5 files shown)


toncho11:
Can async IO be used to speed up the FAT driver?

ghaerr:
> Can async IO be used to speed up the FAT driver?

Well, not the rather slow FAT code directly, but implementing async I/O could increase general responsiveness considerably. 

The problem is that for high compatibility, ELKS uses the machine BIOS for floppy and hard disk I/O, and the BIOS cannot support async I/O. Thus, card or motherboard-specific floppy and/or hard disk drivers must be implemented that directly talk to a machine's specific hardware. While many older floppy controllers are (seemingly?) compatible, few folks are actually using floppy drives instead of solid-state floppy emulators today. This is nonetheless a very interesting subject to continue to look into.

toncho11:
Maybe the BIOS should stay the default, but there should be a kernel compile option to replace the BIOS with a direct access driver. Is it doable on the level of source code? Hmm adding an abstraction over the BIOS or direct access will in itself add a delay.  So all disk kernel calls should be grouped in one file and then there should be a BIOS and several disk versions (for several types of HDDs) of this file?

Issue #1644, opened 2023-08-09
Proposal to discard kernel .fartext.init section after startup


ghaerr:
Hello @tkchia,

When the current kernel is built with CONFIG_FARTEXT_KERNEL=y, two labeled portions of kernel code are treated differently: `INITPROC` for kernel initialization routines, used only once, and `FATPROC` for the FAT filesystem. Both are used to reduce the size of the near code segment, allowing for fast kernel execution most of the time and a larger than 64K total code segment.

The `INITPROC` macro is defined as:
```
#define INITPROC __far __attribute__ ((far_section, noinline, section (".fartext.init")))
```
I am considering the idea of discarding the `.fartext.init` section after kernel initialization, and adding that memory back into the main memory free list, which would then allow for use by application program code or data segments, or external kernel buffers.

In order to do this, the `elks-small.ld` linker script will need some adjustment. I would like to ask your help on this, if possible. What is needed would be to:
- Sort the `.fartext` sections, and place the combined `.fartext.init` as the final section in the kernel code segment. Since the current system physical memory layout is the kernel data segment followed by the kernel code segment(s), this would put the `.fartext.init` section at the highest memory address used, which with some other consideration, would allow it to likely be merged into a larger block of available memory.
- Declare two global symbols, accessible after linking from the kernel, indicating the starting segment and size of the `.fartext.init` section.

With the above, it would be quite easy to just add the unused INITPROC section after initialization to the memory free list. With a little more work, other kernel main memory (e.g. buffer, etc) allocation could be delayed to allow for a contiguous addition to the free list, as described above.

Do you have some suggestions as to how the `elks-small.ld` could be modified for this? Although the same script is used for the kernel and applications programs, I don't think this will be an issue.

Thank you!

tkchia:
Hello @ghaerr,

A while back I crafted a linker script for `newlib-ia16` to allow building of MS-DOS tiny-model terminate-and-stay-resident programs (https://github.com/tkchia/newlib-ia16/blob/bad4dfe895de09aae26271de18acfa99ca286a71/libgloss/ia16/dtr-mt.ld.in); we could probably employ a similar method.

[...]

PR #1645, opened 2023-08-10
[kernel] Update linker script to place INITPROC sections at highest addresses

ghaerr:
First step in allowing for the kernel INIT segment to be discarded after initialization, providing more memory for application programs, or perhaps other system resources.

Using suggestions provided by @tkchia in #1644.

(Also, I was incorrect in the statement that the kernel code segment comes after the data segment: this was reversed some time ago to allow for the kernel data segment size to by dynamically allocated at startup using `SETUP_HEAPSIZE`.) Quick testing shows that the benefit of discarding the kernel INIT code segment is only going to be 5K bytes.

### DIFF
--- a/elks/elks-small.ld
+++ b/elks/elks-small.ld
@@ -47,14 +47,16 @@ SECTIONS {
 		 * IA-16 segment start markers.
 		 */
 		*(".fartext!*" ".fartext.*!")
+		"__start_fartext_init!" = .;
+		"__end_fartext_init!" = .;
 
 		/* Actual segment contents. */
 		/* *(.fartext .fartext$* ".fartext.*[^&]") */
 
 		*(.fartext .fartext.
 		".fartext.[^i]*" ".fartext.i[^n]*" ".fartext.in[^i]*" ".fartext.ini[^t]*")
-		__start_fartext_init = .;
-		*(.fartext .fartext.init ".fartext.init")
+		 __start_fartext_init = .;
+		*(.fartext .fartext.init)
 		__end_fartext_init = .;
 
 		/* IA-16 segment end markers. */
--- a/elks/elks-small.ld
+++ b/elks/elks-small.ld
@@ -53,9 +53,9 @@ SECTIONS {
 
 		*(.fartext .fartext.
 		".fartext.[^i]*" ".fartext.i[^n]*" ".fartext.in[^i]*" ".fartext.ini[^t]*")
-		__start_fartext = .;
+		__start_fartext_init = .;
 		*(.fartext .fartext.init ".fartext.init")
-		__end_fartext = .;
+		__end_fartext_init = .;
 
 		/* IA-16 segment end markers. */
 		*(".fartext&*" ".fartext.*&")
--- a/elks/elks-small.ld
+++ b/elks/elks-small.ld
@@ -49,7 +49,13 @@ SECTIONS {
 		*(".fartext!*" ".fartext.*!")
 
 		/* Actual segment contents. */
-		*(.fartext .fartext$* ".fartext.*[^&]")
+		/* *(.fartext .fartext$* ".fartext.*[^&]") */
+
+		*(.fartext .fartext.
+		".fartext.[^i]*" ".fartext.i[^n]*" ".fartext.in[^i]*" ".fartext.ini[^t]*")
+		__start_fartext = .;
+		*(.fartext .fartext.init ".fartext.init")
+		__end_fartext = .;
 
 		/* IA-16 segment end markers. */
 		*(".fartext&*" ".fartext.*&")


tkchia:
Hello @ghaerr,

For the segment start and segment end markers, you can define the corresponding `!` symbols together with the corresponding `!` sections defining other `!` symbols &mdash; like so:
```
	*(".fartext!*" ".fartext.*!")
	"__start_fartext_init!" = .;
	"__end_fartext_init!" = .;
```
or alternatively:
```
	"__start_fartext_init!" = .;
	"__end_fartext_init!" = .;
	*(".fartext!*" ".fartext.*!")
```
(The `!` segments should have zero length.)

Since the segment:offset form of `__start_fartext_init` will almost certainly need to be of the form `.fartext:`_something_, so `__start_fartext_init!` &mdash; which marks the base of the `__start_fartext_init`'s IA-16 segment &mdash; should point to the beginning of `.fartext`, which will do the correct thing.

Also, I believe `.fartext.init` and `".fartext.init"` are the same pattern?

Thank you!

ghaerr:
Hello @tkchia,

When I quote the symbol values as suggested using:
```
        *(.fartext .fartext.
        ".fartext.[^i]*" ".fartext.i[^n]*" ".fartext.in[^i]*" ".fartext.ini[^t]*")
        /* __start_fartext_init = .;*/
        "__start_fartext_init!" = .;
        *(.fartext .fartext.init ".fartext.init")
        /*__end_fartext_init = .; */
        "__end_fartext_init!" = .;
```
and then use the following in C:
```
extern void INITPROC __start_fartext_init(void);
extern void INITPROC __end_fartext_init(void);
```
I get the following linker errors:
```
init/main.o:function kernel_init: error: unaligned R_386_SEG16 reloc value
init/main.o:function kernel_init: error: undefined reference to '__start_fartext_init'
init/main.o:function kernel_init: error: undefined reference to '__end_fartext_init'
init/main.o:function kernel_init: error: undefined reference to '__start_fartext_init'
init/main.o:function kernel_init: error: undefined reference to '__end_fartext_init'
```
(Uncommenting the two lines in the linker script above still leaves the following error):
```
init/main.o:function kernel_init: error: unaligned R_386_SEG16 reloc value
```
Thus, it seems the linker still wants an actual definition from somewhere else... which is what I eventually did (see below).

I was able to get the entire mechanism working using the previous mechanism, using the originally suggested approach (at the cost of having to declare actual INITPROC functions in C):
```
void INITPROC __start_fartext_init(void) {}
void INITPROC __end_fartext_init(void) {}
```
```
        *(.fartext .fartext.
        ".fartext.[^i]*" ".fartext.i[^n]*" ".fartext.in[^i]*" ".fartext.ini[^t]*")
        __start_fartext_init = .
        *(.fartext .fartext.init ".fartext.init")
        __end_fartext_init = .; 
```
Using this method allowed me to write and complete a working mechanism to discard the unused memory, which is working fine. 

Shall I push that onto this PR for your inspection and review, or do you think it better to try to solve the linker script issue using `!`?

Thank you!

tkchia:
Hello @ghaerr,

Wait, no &mdash; you need to define _both_ `__start_fartext_init!` and `__start_fartext_init`.  The first symbol marks the segment base and the second points to the actual location (base + offset).

You can define `__start_fartext_init!` along with all the other `!` symbols, as I mentioned.

Thank you!

ghaerr:
Hello @tkchia,

> you need to define both __start_fartext_init! and __start_fartext_init. The first symbol marks the segment base and the second points to the actual location (base + offset).

I now understand what the symbols refer to, thanks. However, as mentioned above, when I define both (by uncommenting the two lines posted), I still get the following error:
```
init/main.o:function kernel_init: error: unaligned R_386_SEG16 reloc value
```

I think I may post the working version, for your review. It is unclear why defining both symbols (a total of four), still gives an unaligned R_386_SEG16 error.

Thank you!

tkchia:
Hello @ghaerr,

As I mentioned, try defining the `!` symbols _here_:
```
	*(".fartext!*" ".fartext.*!")
	"__start_fartext_init!" = .;
	"__end_fartext_init!" = .;
```
Or _here_:
```
	*(".fartext!*" ".fartext.*!")
	"__start_fartext_init!" = .;
	"__end_fartext_init!" = .;
```

tkchia:
The `!` symbols all point to the _base_ of the IA-16 segment containing the corresponding non-`!` symbols.  So all the symbols that "belong" to the same IA-16 segment all have `!`' symbols that point to the same place.  E.g. here is part of a linker map file for a simple test program I wrote:
```
.text           0x00010000      0x4b0
 .inithead!     0x00010000        0x0 .../crt0.o
                0x00010000                _start!
 .preinit!      0x00010000        0x0 .../libc.a(environ.o)
 .preinit!      0x00010000        0x0 .../libc.a(argcargv.o)
 .init!         0x00010000        0x0 .../libc.a(do-global-ctors.o)
 .postinit!     0x00010000        0x0 .../libc.a(argcargv.o)
 .inittail!     0x00010000        0x0 .../crt0.o
 .text!         0x00010000        0x0 .../crt0.o
 .text!         0x00010000        0x0 .../cc1emnAz.o
 .fartext.f.elks_fartext_test.c.62242!
                0x00010000        0x0 .../cc1emnAz.o
                0x00010000                hello1!
                0x00010000                hello2!
                0x00010000                hello3!
                0x00010000                main!
 .text!         0x00010000        0x0 .../libc.a(_exit.o)
                0x00010000                _exit!
...
```
and
```
 .inithead      0x00010000       0x11 .../crt0.o
                0x00010000                _start
 .preinit       0x00010011        0x3 .../libc.a(environ.o)
 .preinit       0x00010014        0x8 .../libc.a(argcargv.o)
 .init          0x0001001c       0x11 .../libc.a(do-global-ctors.o)
 .postinit      0x0001002d        0x5 .../libc.a(argcargv.o)
 .inittail      0x00010032        0xc .../crt0.o
 .text          0x0001003e        0x0 .../crt0.o
 .text          0x0001003e        0x0 .../cc1emnAz.o
 .fartext.f.elks_fartext_test.c.62242$
                0x0001003e      0x183 .../cc1emnAz.o
                0x0001003e                hello1
                0x000100a7                hello2
                0x000100d6                hello3
                0x000100f6                main
 .text          0x000101c1        0x6 .../libc.a(_exit.o)
                0x000101c1                _exit
...
```
Thank you!

ghaerr:
Hello @tkchia,

> As I mentioned, try defining the ! symbols here:

Ok, let me work on this a bit more, I now understand the nuance of the `!` symbols all pointing to the same place. Hopefully I can get it working. The advantage of your method is that the symbols are automatic, and don't need additional C definitions (just externs).

Thank you!

ghaerr:
Hello @tkchia,

It is all working, thank you! I will update this PR with the final linker script changes then open another PR with the changes allowing the kernel to release the `.fartext.init` section into the main memory free list. I also removed the extra `".fartext.init"` as suggested. What I have learned is that quoted symbols are the same as unquoted symbols, except for special regex or non-standard symbol character processing.

Thank you!

PR #1656, opened 2023-08-19
[cmds] Port md5 from Minix/NetBSD to ELKS as md5sum

ccoffing:


ghaerr:
Looks good, thanks @ccoffing! Would you like to add an entry into `elkscmd/Applications` so that this gets copied onto the 2880k+ distributions? (Not having any tags like `:1440k` in `Applications` essentially specifies 2880k and HD images). Probably better there, there's only 34k left on 1440k.

ccoffing:
Done.

General question (I could discover this myself so obviously this is a leading question).  Are the size tags in this file "this size and greater", or only that specific size?  It seems strange that some things are installed (for example) only with the ":720k" tag, yet others list multiple size tags.  If it's important, shouldn't it be installed starting with the first image that it fits, and everything larger?

ghaerr:
> Are the size tags in this file "this size and greater", or only that specific size? 

They are interpreted by elkscmd/Make.install to be "this size and greater".

>  yet others list multiple size tags.

Yes, I see that all the `:1200k` tags also have a `:1440k` tag, which is confusing and unnecessary. Thanks for pointing that out! This could be cleaned up (note, the `:128k`tag  is handled differently and is used for ROM builds). IIRC, the 1200k floppy was added separately some time ago and likely was the source of this.

The following in `Make.install` handles the selection and interpretation of "this size and greater":
```
#
# Install apps by image size    
#
ifdef CONFIG_APPS_360K          
    TAGS = :boot|:360k|         
endif

ifdef CONFIG_APPS_720K          
    TAGS = :boot|:360k|:net|:720k|
endif

ifdef CONFIG_APPS_1200K         
    TAGS = :boot|:360k|:net|:720k|:1200k|       
endif

ifdef CONFIG_APPS_1232K         
    TAGS = :boot|:360k|:net|:720k|:1200k|           
endif

ifdef CONFIG_APPS_1440K 
    TAGS = :boot|:360k|:net|:720k|:1200k|:1440k|                
endif

ifdef CONFIG_APPS_2880K         
    # below TAGS relies on :ash rule following :defsash rule in Applications
    TAGS = :*|
    CONFIG_APP_MAN_PAGES=y      
endif
```

ghaerr:
BTW, forgot to mention - nice work in porting `md5sum` and keeping it super small with no printf/stdio! :)

PR #1673, opened 2023-08-27
[kernel] Add physaddr 64k I/O overlap checking to BIOS driver

ghaerr:
Used in case an L1 cache buffer overlaps a 64k physical boundary, a BIOS restriction on floppy I/O because of the DMA hardware. Also checks overlap for HD I/O since not entirely sure of BIOS limitations.

Next step is adding dynamically allocated L1 cache; previous cache never overlapped 64k but could happen after that.

Also added map, unmap and remap counts to ^O buffer statistics.


ghaerr:
Physical address overlap checking added as a result of conversation starting in https://github.com/Mellvik/TLVC/issues/20#issuecomment-1694350505.

ghaerr:
Physical address overlap checking added as a result of conversation starting in https://github.com/Mellvik/TLVC/pull/19#issuecomment-1687069029.

Issue #1692, opened 2023-09-03
VFS can't mount root


cathoderaydude:
Just downloaded the current release, 0.7.0. I'm trying to boot on a Toshiba T1100 Plus with DD drives, so I used fd720-fat.img. When I try to boot, I get "VFS: Unable to open root device." Tried with fd720-minix.img as well, same result. Screenshot below is with minix version. 

I'm using an HD disk, with the density window covered, formatted as 720k, writing image using rawrite under windows. Have been using this same workflow to produce 720k disks, used in this machine, with no trouble, so I don't think it's a media or image writing issue.

![image](https://github.com/ghaerr/elks/assets/46694173/c229e478-f0c1-4ff3-bb29-5eb18aa21d64)

toncho11:
Your system does not say:

`/dev/fd0: 80 cylinders, 2 heads, 9 sectors = 720.0 kb` after network probing. 
This is normally what it says when you use a 720 kb ELKS image.

I am guessing, but it should be something about HD vs DD. It loads to some degree, but then the sectors it expects are not there.

ghaerr:
Hello @cathoderaydude,

Thanks for the screenshot - I can see what's happening: "device_setup: BIOS drive 0x01, root device 0x3a0" - the system is trying to mount the filesystem on drive B, based on the Toshiba BIOS saying that drive 1 is the boot drive (0x3a0 is /dev/fd1 on ELKS).

I'm not sure why this is happening on the T1100. Is there a config that allows setting the boot drive? I imagine it's working fine for DOS. You can force the filesystem on drive A by adding the line `root=fd0` to the /bootopts file (easiest done on FAT).

Are there two floppies on that system? If so, ELKS would mount the filesystem on drive B if present. 

Thank you!

cathoderaydude:
Hey @ghaerr,

Thanks for the quick response! So, this machine does have two FDDs, and I tried booting from both with the same results except that it IDs the device differently. The above screenshot was taken with the disk in drive 2, but if i move it to drive 1, it reports the disk as 0x0, root device 0x380, then fails again. 

[...]

PR #1694, opened 2023-09-04
[kernel] Support Toshiba T1100 BIOS when INT 13 returns 0 floppies

ghaerr:
Solves Toshiba T1100 booting problem reported in #1692.

The T1100 BIOS has an INT 13h BIOS "get drive parms" function, but it always returns 0 drives. When this happens, this enhancement uses the BDA "equipment" byte to get the floppy drive count and sets them to 720k default before probing.

A few additional printk's were added to better see what's happening when BIOS INT13 AH=8 get drive parms fails on other BIOSes.

@cathoderaydude, I've attached two 720k images, can you please test to ensure both drives are now working and post a final screenshot?

Thank you!

[fd720-3.zip](https://github.com/ghaerr/elks/files/12508496/fd720-3.zip)

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -412,10 +412,17 @@ static unsigned short int INITPROC bioshd_getfdinfo(void)
     BD_AX = BIOSHD_DRIVE_PARMS;
     BD_DX = 0;                  /* query floppies only*/
     BD_ES = BD_DI = BD_SI = 0;  /* guard against BIOS bugs*/
-    if (!call_bios(&bdt))
-        ndrives = BD_DX & 0xff; /* floppy drive count*/
-    else
-        printk("fd: no bios get drives, ndrives %d\n", ndrives);
+    if (!call_bios(&bdt)) {
+        int drives = BD_DX & 0xff;      /* floppy drive count */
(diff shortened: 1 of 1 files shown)


toncho11:
Hi @ghaerr ,

I am looking below this line:https://github.com/ghaerr/elks/blob/f3ad774b25d9836f2c62d00a7a38552052284c27/elks/arch/i86/drivers/block/bioshd.c#L417
So if the floppy can not be determined then it is either 1440(AT) or 360 (XT), but in the case of computers that return 0 drives it will be 720 kb to match the Toshiba's 720 kb floppy?

ghaerr:
> So if the floppy can not be determined then it is either 1440(AT) or 360 (XT), but in the case of computers that return 0 drives it will be 720 kb to match the Toshiba's 720 kb floppy?

Yes. But two more steps are taken that may overwrite this initial default: an inserted floppy is first probed, which consists of trying to read the DOS BPB (if FAT) or ELKS BPB (if MINIX) to get the CHS. If none is found, a sector-probing is performed. If that fails, this original default is used. Add to that the BIOS may have its own idea of what the floppy format is. 

This is why the drive format is displayed at various points during ELKS operation. Notice in the last screenshot above the "ELKS bootable" line: this indicates an ELKS BPB was found and that an ELKS-created MINIX floppy was probed. (In this case, it matches the initial guessed parameters).

PR #1708, opened 2023-09-10
[kernel][libc][cmds] Add devname libc function

ghaerr:


ghaerr:
Adds `linuxmt/devnum.h` as kernel header file for macros to access various block devices, in preparation for possible change of minor device numbers in the `bioshd` driver. Removes all hardcoded numbers from kernel source.

Adds `devname` to C library. By default opens /dev and reads directory for dynamic lookup of device names. The old versions contained in df.c and mount.c used hard-coded major/minor device numbers. This can be re-enabled for very slow floppy-only system using USE_FASTVERSION=1 in libc/misc/devname.c.

Rewrites `ttyname`, updates `ps` to use faster version of internal `dev_name` using less space.

PR #1709, opened 2023-09-11
[kernel] Speed up TTY output processing

ghaerr:
Tested on QEMU, this increases `sl` output by 25-50% on already fast systems. Discussed in https://github.com/ghaerr/elks/issues/1619#issuecomment-1712385485.

It seems @Vutshi has a knack for finding deep rabbit holes. Using the operation of `sl`, testing has found that the output speed varies considerably, based on something I can't pinpoint. I've looked at the scheduler, kernel timers, serial console, and all seem operating correctly. What is being observed is that (especially with this enhancement) `sl` runs fast ~0.45 seconds on QEMU on boot (from ~0.73s - 1.0s before). But after running networking, then stopping it, or running /bin/getty on a serial line, then stopping it, or even just typing ^D a few times to login: on serial, the `sl` output time increases 50% to 0.73s or more, then sometimes drops back to speedy again.

This needs testing on real hardware - I'm very interested to learn how much quicker this change gets us, as well as any slowdowns seen when `init 3` is run opening a /bin/getty on /dev/ttyS0.

### DIFF
--- a/elks/arch/i86/drivers/char/ntty.c
+++ b/elks/arch/i86/drivers/char/ntty.c
@@ -212,14 +212,14 @@ void tty_release(struct inode *inode, struct file *file)
  *
  * while (tty->outq.len > 0) {
  *      ch = tty_outproc(tty);
- *      write_char_to_device(device, (char)ch);
+ *      write_char_to_device(device, ch);
  *      cnt++;
  * }
  *
  */
 int tty_outproc(register struct tty *tty)
 {
-    int t_oflag;        /* WARNING: highly arch dependent, termios.c_oflag truncated to 16 bits*/
+    int t_oflag;    /* WARNING: termios.c_oflag truncated to 16 bits */
     int ch;
 
     ch = tty->outq.base[tty->outq.tail];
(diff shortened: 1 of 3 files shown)


Vutshi:
@ghaerr I will test it as soon as I have access to the hardware. Meanwhile, I’ve learned about a new 8086 emulator which claims to be CPU cycle exact. 
https://int10h.org/blog/2023/07/martypc-pc-xt-emulator-raising-the-bar/
I wonder whether it can be useful in our context of performance debugging as it also provides some extensive debugging info.

ghaerr:
>  I’ve learned about a new 8086 emulator which claims to be CPU cycle exact.

I just read about that emulator last week. IIRC it also emulates much of the PC hardware. I think it as well as other non-QEMU emulators would be very useful for ELKS. The problem with many emulators is they can take some time to setup properly. I have more emulators on my long list of things to do, but don't have lots of time to learn new ones. Anyone that has knowledge of a specific emulator or some time to set one up - it would be great to hear from you :)

@tyama501: I have `dosbox` setup and running which I use for testing your PC-98 port, but I have not been able to get it to boot non-FAT floppies (since we're using the its `boot` command running under DOS). Have you found a way to get it to boot an IBM PC MINIX-formattted floppy?

Are there other emulators that we should prioritize? What about Bochs, does anyone use that anymore? I have ongoing concerns about QEMU, as it has a nasty bug/feature that turns off the INT 8 hardware timer interrupt at times (I think too save CPU usage) which wreaks havoc on system timings and ELKS operation in general. It apparently only does this when the serial port is enabled. I would much like to test the `sl` timing issues on another emulator. Although the speed issues described above are likely ELKS internal, I haven't ruled out QEMU timing issues either.

tyama501:
I didn't know DOSBox-X cannot boot non-FAT but I tried now and failed to boot...

PR #1717, opened 2023-09-13
[kernel] Disable track caching for hard drives

ghaerr:
Track caching will only be performed for floppy drives. This will allow for better performance since the floppy track cache won't be invalidated when/if a hard drive access is made. 

This also allows the experimental direct floppy driver to always use DMASEG for simplification.


ghaerr:
This enhancement will be used by further work in #1716.

PR #1719, opened 2023-09-13
[kernel] Cleanup ifdefs and unused code in direct floppy driver

ghaerr:
Removes unused or unneeded ifdefs and outdated code not used in ELKS or TLVC, for understandability and maintainability.
Follow-on from importing direct floppy driver in #1716.

Removes BDEV_SIZE_CHK and most CONFIG_FS_XMS_BUFFER ifdefs.
Removes drive "aliasing" from open code.
Will add `fsync_dev/invalidate_inodes` to `floppy_release` in separate push.

Considering whether to remove all floppy formatting code.
Considering whether `kernel_ds:tmp_floppy_area` could be replaced with `DMASEG:0`, but unsure if DMASEG used separately in driver.

Tested and works with no functional changes in QEMU.

### DIFF
--- a/elks/arch/i86/drivers/block/directfd.c
+++ b/elks/arch/i86/drivers/block/directfd.c
@@ -234,9 +234,7 @@ struct floppy_struct *base_type[4];
  * User-provided type information. current_type points to
  * the respective entry of this array.
  */
-#ifdef HAS_IOCTL
-struct floppy_struct user_params[4];
-#endif
+//struct floppy_struct user_params[4];
 
 /*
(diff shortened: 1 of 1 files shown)


ghaerr:
Adds latest direct driver enhancements from @Mellvik's https://github.com/Mellvik/TLVC/pull/24.

BTW @Mellvik, I'm not sure its OK to not call `fsync_dev(dev)` on the last close to `floppy_release`, since one can't depend on `umount` doing it - the device could be unmounted and accessed just by open/close of /dev/df0, in which case the buffers would not be synced.

PR #1722, opened 2023-09-15
[libc/cmds] Fix time command hour wrapping problem

ghaerr:
Fixes `time` command sometimes displaying incorrect results, reported by @Vutshi in https://github.com/ghaerr/elks/issues/1619#issuecomment-1703364279.

`time` would report incorrect results (e.g. an extremely large number) when a command would execute for a period of time where the number of seconds since Jan , 1970 would "wrap" during a modulo 3600 operation. Long story short, previously `time` used microsecond time calculations with a 32-bit long which was subjected to the improper 3600 second modulo operation approximately every hour GMT. It has been rewritten to use 64-bit arithmetic and is no longer subject to any time modulus.

The `times` C library routine has been removed due to being incorrect, and an alternative method used to get times in `time`.

@Vutshi, `time sl -s` should display properly now, although I don't expect you to actually wait 34 minutes to test it!

### DIFF
--- a/elkscmd/misc_utils/time.c
+++ b/elkscmd/misc_utils/time.c
@@ -1,28 +1,39 @@
 /* time command - modified from BSD 4.2 by Greg Haerr*/
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <signal.h>
 #include <sys/types.h>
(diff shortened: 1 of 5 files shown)


Vutshi:
> although I don't expect you to actually wait 34 minutes to test it!

This shall be done.

PR #1723, opened 2023-09-15
[cmds] Add accurate train timing to sl

ghaerr:
Adds an accurate 40ms "cycle time" between each train re-display, so that the `sl` train should always chug along at a constant speed on all systems (or at least as fast as possible on slower systems). `sl` now calculates a minimum cycle of 40 ms and will add a delay if the last display cycle was quicker than that. This now helps the illusion that all trains have equal conductors (yes, I'm talking about your 8088 systems @Vutshi).

In order to continue to use `sl` as an indicator of how fast or slow your system may actually be without a delay, a new `-f` option was added: `sl -f` will run the trains off the tracks if your system is fast enough (which is the same as `sl` previously with no option).

To summarize our steam powered evolution:
```
sl     # train running at 40ms/frame or slower on all systems
sl -s  # this train emits more ANSI sequences than slow systems can handle
sl -f  # this train doesn't delay or stop for anybody
```

This finalizes all of the reported problems in #1619.

### DIFF
--- a/elkscmd/tui/sl.c
+++ b/elkscmd/tui/sl.c
@@ -95,7 +95,7 @@ void option(char *str)
 
 void cycle(unsigned long usecs)
 {
-    long elapsed;
+    unsigned long elapsed;
     struct timeval now;
     static long start;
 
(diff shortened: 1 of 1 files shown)


Vutshi:
At this stage, ELKS might just happen to possess the most advanced railway system on the planet.

PR #1728, opened 2023-09-20
[kernel] Use outb directly for DMA programming in DF driver

ghaerr:
This uses another one of @Mellvik's ideas - that of dropping the relatively large set of DMA access routines and replacing them with programming the DMA chip directly from the DF driver. This saves ~500 bytes of kernel text and works well because no other devices use or will use DMA. (Even the HD direct driver uses programmed I/O).

The DMA programming is done within the `setup_DMA` routine and borrows from the simplicity found from the Minix OS source.

### DIFF
--- a/elks/arch/i86/drivers/block/Makefile
+++ b/elks/arch/i86/drivers/block/Makefile
@@ -58,7 +58,7 @@ endif
 
 # experimental direct fd support
 ifeq ($(CONFIG_BLK_DEV_FD), y)
-OBJS += directfd.o dma.o
+OBJS += directfd.o
 endif
 
 # experimental (and not working) direct hd support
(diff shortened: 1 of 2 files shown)


Mellvik:
Thank you @ghaerr,

you saved me some time here! :-)

ghaerr:
@Mellvik: I added a FIXME about whether disabling interrupts is actually required during DMA setup. I haven't tested on real hardware so I didn't remove it. It is my understanding that since the kernel process won't ever be switched out and DMA is used nowhere else, the disabling is unnecessary. The only other code that can possibly run are interrupt handling routines which don't touch the DMA controller.

The code generated for the DMA setup is still a bit long and could affect fast serial in or possibly network transfers. It might be something to add to your testing list on real hardware with networking, as the less disabled interrupts the better for lost packets etc. FYI.

Mellvik:
@ghaerr,
the irq disabling was experimental, removing it is most likely fine. I have more of those to look at so I'll add it to the list.

Thanks.

PR #1745, opened 2023-10-07
[direct floppy] Small updates to DF driver

ghaerr:
Don't call original BIOS hardware IRQ 0 timer routine unless configured for BIOS floppy driver (CONFIG_DEV_BLK_BFD). I'm pretty sure this is only useful for BIOS floppy motor timeout, but this hasn't been tested on real hardware. Also assumes that BIOS HD support does not require the timer.

Add adjustable timeout values for motor on, off and command timeout.

Support two floppy drives; this halves the number of /dev entries required for preset floppy format types, should anyone actually ever use that feature.

### DIFF
--- a/elks/arch/i86/drivers/block/blk.h
+++ b/elks/arch/i86/drivers/block/blk.h
@@ -73,7 +73,7 @@ static void floppy_off(int nr);
 
 #define DEVICE_NAME "df"
 #define DEVICE_REQUEST do_fd_request
-#define DEVICE_NR(device) ((device) & 3)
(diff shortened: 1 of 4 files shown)


tyama501:
Hello @ghaerr ,

Does supporting only 2 floppies in blk.h affect BIOS driver too?
PC-98 is supporting 4 floppy drives.

ghaerr:
> Does supporting only 2 floppies in blk.h affect BIOS driver too?

No, the change is only for the direct floppy driver. The BIOS driver supports 4 drives as before.

tyama501:
Thank you!

PR #1749, opened 2023-10-08
support for MacOs brew qemu installation

madejusz:


madejusz:
Adjustment for qemu.sh for MacOs brew support.

ghaerr:
Thank you @madejusz!

PR #1758, opened 2023-11-26
[libc] Fix glitch when building elks-libc on launchpad.net

tkchia:


ghaerr:
Hello @tkchia,

LGTM. Out of curiosity - why is the ~ problem turning up only on launchpad.net?

Thank you!

tkchia:
Hello @ghaerr,

On `launchpad.net` the ELKS `libc` code was being built inside a build directory with a `~` inside the path component due to a pre-release "version number", something like (IIRC) `elks-libc-gcc-ia16-elf-0.8.0~pre-dev`.  The issue did not occur on e.g. my local PC, because I was not building ELKS `libc` in a directory with a `~` in its name.

Thank you!

PR #1762, opened 2023-12-03
[build] Speed up ELKS build

ghaerr:
@and3rson found in https://github.com/ghaerr/elks/pull/1760 that the build process was slowed down through unneeded duplicate setting of variables in `elks/Makefile-rules`.  This PR enhances that for further speedup. One of the biggest problems was the method of setting the variables used for `make dist` which isn't used anymore anyways.


and3rson:
Awesome - the build now takes literally few seconds!

PR #1776, opened 2023-12-24
smaller which

ccoffing:


ghaerr:
Hello @ccoffing,

Nice rewrite. Now that you've rewritten this for size and removed `printf`, why not remove all of stdio (~1500 bytes) as well? Suggest perhaps using the following style defines (modified from elkscmd/file_utils/futils.h as `outstr` or similar), and just call `write` instead using STDOUT_FILENO:
```
#define errstr(str) write(STDERR_FILENO, str, strlen(str))
```
Would you like to do something like that or commit this as is?

Thank you!

ccoffing:
Oh yeah, good catch.  I forget sometimes that `puts` is not lightweight...

PR #1777, opened 2023-12-24
[bioshd] fix fd_types

tyama501:
Hello @ghaerr ,

This fix is for fd_type and the cyliner in
https://github.com/ghaerr/elks/issues/1775

I think #ifdef CONFIG_ARCH_PC98 is better than #ifdef CONFIG_IMG_FD1232 since
1440 kernel for PC-98 can also read 1232 disk.

720KB for PC-98 enhancement will be coming on the next PR.

Thank you.


ghaerr:
> I think #ifdef CONFIG_ARCH_PC98 is better than #ifdef CONFIG_IMG_FD1232 since
1440 kernel for PC-98 can also read 1232 disk.

Agreed, now that PC-98 may support multiple drive types simultaneously. There may be other similar cases that perhaps should be switched. In particular, IIRC, CONFIG_IMG_FD1232 is used to optimize the variable sector size code in the FAT filesystem, producing more code only when CONFIG_IMG_FD1232 is defined. Perhaps a future version of ELKS PC-98 should work for all FD sizes and CONFIG_IMG_FD1232 not be needed anymore?

Thank you!

tyama501:
Actually, ifdef for FD1232/FD1440 are still needed for boot code and the default value for the device address, for now,
since the boot sector is too small to detect and switch both cases.

PR #1778, opened 2023-12-25
Tweaks to allow building on other BSDs

ccoffing:
My machine that can write a floppy is now old enough that Linux doesn't really support it, so I switched it to NetBSD.  Here are the minimal changes that I needed to get a build.

lxconfig still doesn't build because it has a `scroll` variable that collides with the ncurses function.  I don't know how this normally manages to build.  But that can be bypassed by dropping a .config in place.

This expects all the GNU stuff to be installed, and I had to trick the Makefiles to use GNU make instead of BSD's make, by first running `alias make=gmake`.  But all things considered, it was surprisingly easy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ all: .config include/autoconf.h
 	$(MAKE) -C bootblocks all
 	$(MAKE) -C elkscmd all
 	$(MAKE) -C image all
-ifneq ($(shell uname), Darwin)
+ifeq ($(shell uname), Linux)
 	$(MAKE) -C elksemu PREFIX='$(TOPDIR)/cross' elksemu
 endif
 
@@ -28,7 +28,7 @@ clean:
 	$(MAKE) -C bootblocks clean
 	$(MAKE) -C elkscmd clean
 	$(MAKE) -C image clean
-ifneq ($(shell uname), Darwin)
+ifeq ($(shell uname), Linux)
 	$(MAKE) -C elksemu clean
 endif
 	@echo
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This build script is called in main.yml by GitHub Continuous Integration
 # Full build (including the cross tool chain)
--- a/clean.sh
+++ b/clean.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Full clean (except the cross build chain)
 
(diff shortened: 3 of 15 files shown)


ghaerr:
Would you like to take a pass at renaming the ‘scroll’ variable or have this committed now?

Thank you!

ccoffing:
> Would you like to take a pass at renaming the ‘scroll’ variable or have this committed now?

It sure would be nice to have it work right.  I'll fix it.

ccoffing:
This now builds for me on both Linux and NetBSD.  Please check that I didn't goof anything for Mac.

Thanks!

ghaerr:
Hi Chuck,

Looks like you spent some time hacking lxdialog into submission, thanks! 

Have you tested this at all on macOS, or do I need to before commit? Specifically, I'm unsure where/whether macOS gets the `major/minor` macros since there's no `sys/sysmacros.h`? You may need to define that for `#ifdef __APPLE__` using the values from MAJOR/MINOR_SHIFT (which I hacked in long ago on the first macOS port) unless we have a known place for those macros.

My other question is: what's up with the NetBSD compiler regarding local variables/arguments named `scroll`? They are conflicting with a global ncurses function of the same name? Strange, how could that work with much C code, you'd think that would have interfered with lots of other C code a long time ago?!

I guess for shell scripts, `bash` isn't guaranteed to be in `/bin`, huh? But `env` is always at `/usr/bin/env`? I guess its just history for when FreeBSD changed the shebang processing for shell scripts and `/usr/bin/env` was always used... ?

Thank you!

ghaerr:
> do I need to before commit? Specifically, I'm unsure where/whether macOS gets the major/minor macros since there's no sys/sysmacros.h

I just tested on macOS and things work, great! I didn't know that `major()` and `minor()` are included in `sys/types.h` for POSIX.

ccoffing:
> My other question is: what's up with the NetBSD compiler regarding local variables/arguments named `scroll`?

I think I know what was going on there.  The collision happens when attempting to build against curses but not ncurses, so the failure isn't really specific to NetBSD.  In curses, `scroll` is a function, which will collide with a variable of the same name.  ncurses `#define`s `scroll` as a parameterized macro, which doesn't collide.

I only managed to hit that because I had ncurses installed, which bypassed the user-facing error message, but this was before I had added the `pkg-config` check, so it fell back and included `curses.h`.

Obviously this Makefile could use some cleanup:  Part of is insists that ncurses is installed, but other parts of it seem to try to be compatible with BSD curses.  I might look into that...

PR #1779, opened 2023-12-28
[bioshd] 720KB 2DD support for PC-98 2HD/2DD common drive

tyama501:
Hello @ghaerr ,

This is the PR of the 720KB 2DD support for PC-98 as we have talked.

Currently, this is focused on the 720KB(9sectors/track) with device address 0x10 for 2HD/2DD common drive.
640KB(8sectors/track) or device address 0x70 for 2DD sole drive is not supported.
It reads BPB to reject this.

I have tested with PC-9801UV21 and I could read/write 720KB disk.
 
PC-9801UV21_ELKS_2DD

PC-9801UV21_ELKS_2DD_2

Since PC-9801BX with 1440KB drive is broken,
3modes are tested only on the emulator, but it can mount 720KB/1232KB/1440KB from the same kernel.

ELKS_2DD_mount

Thank you!


ghaerr:
Hello @tyama501,

Looks good, thank you!

> 640KB(8sectors/track) or device address 0x70 for 2DD sole drive is not supported.

Are you thinking it would be nice to support 640KB floppies? You're welcome to add for PC-98 if you wish :)

tyama501:
Thank you!

I tried a little about 640KB, but it seems that it is not so easy,
it may need to change read/write command.

PR #1780, opened 2023-12-30
Path cleanups

ccoffing:
Just a collection of path cleanups I have stumbled across.  I have touch-tested the changes and they seem fine, except that screen won't exit (maybe it crashes) but it does that on master too.

So I think this is good as is, but screen could use some proofreading.

### DIFF
--- a/elkscmd/file_utils/split.c
+++ b/elkscmd/file_utils/split.c
@@ -19,6 +19,7 @@
 
 #include <sys/param.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <unistd.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -31,9 +32,6 @@
 #define OK	0			/* okay exit */
 #define YES	1			/* yes/true */
 
-#ifndef MAXPATHLEN
-#define MAXPATHLEN      256
-#endif
 #define MAXBSIZE        2048
 #define	isdecimal(ch)	(((ch) >= '0') && ((ch) <= '9'))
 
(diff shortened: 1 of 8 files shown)


tyama501:
Hi @ccoffing ,

I have been using "screen" on PC-98 and I can exit(kill?) it by Ctrl-a, k.
Can you exit by it?

Thank you.

ccoffing:
> Hi @ccoffing ,
> 
> I have been using "screen" on PC-98 and I can exit(kill?) it by Ctrl-a, k. Can you exit by it?
> 
> Thank you.

Yes, Ctrl-a k causes screen to exit.  However, that returns me to the login prompt, which seems strange (I would have expected to be returned to the shell which I had started screen from.)

But the problem I originally encountered was if I typed `exit` from the last shell in screen.  I get no further output.  The machine seems to hang at that point.

ghaerr:
Thanks @ccoffing.

> However, that returns me to the login prompt, which seems strange (I would have expected to be returned to the shell which I had started screen from.)

> But the problem I originally encountered was if I typed exit from the last shell in screen. I get no further output. The machine seems to hang at that point.

Both of these sound like possible stack or heap overflow. I notice that screen/Makefile already has `-maout-heap=0x7000`, perhaps that should be set to `-maout-stack=8192 -maout-heap=0xffff` to set 8K stack and max heap, to see if that changes anything. The `screen` source should be scrutinized to determine a max stack (default 4K) and minimum heap size.

I've been meaning to add a debug malloc (work in progress) to allow better runtime checking and reporting of these kinds of things...

ghaerr:
`/bin/sh` is also tight on heap, although running `screen` shouldn't affect that issue. I sometimes login as `toor` and run from `/bin/sash` as an alternative for getting more input on these kinds of problems. In general, I believe both `bash` and `sash` to be pretty bug-free, so the stack overflow or memory overwrite is likely in `screen`.

PR #1781, opened 2023-12-31
screen: modernize prototypes

ccoffing:
Hi @ghaerr and @tyama501 ,

I see several ways that screen could be made smaller and improved, but the old function prototypes are so difficult to work with that I decided to modernize the source code first.  I reformatted in the ELKS style (using the indent.pro settings that @ghaerr provided a while back) and then manually converted the K&R prototypes to ANSI.

More focused improvements can now be done in subsequent PRs, which will hopefully be more readable without this noise in the way.

To be sure I didn't break anything in the process, I built master and then disassembled the object files, and then built this branch and disassembled, and then diff'd.  The only difference is that the modernized prototypes caused the compiler to reorder a couple of instructions.  This does not appear to be a functional change.

```
$ ia16-elf-objdump  -d -m i8086 screen.o  > screen.new
$ diff screen.old  screen.new 
3046,3047c3046,3047
<   1b:	8b 56 08             	mov    0x8(%bp),%dx
<   1e:	89 df                	mov    %bx,%di
---
>   1b:	89 df                	mov    %bx,%di
>   1d:	8b 56 08             	mov    0x8(%bp),%dx
3076,3077c3076,3077
<   75:	c7 46 08 01 00       	movw   $0x1,0x8(%bp)
<   7a:	89 5e 0a             	mov    %bx,0xa(%bp)
---
>   75:	89 5e 0a             	mov    %bx,0xa(%bp)
>   78:	c7 46 08 01 00       	movw   $0x1,0x8(%bp)
3193,3194c3193,3194
<  19d:	8b 56 08             	mov    0x8(%bp),%dx
<  1a0:	8b 7e 0a             	mov    0xa(%bp),%di
---
>  19d:	8b 7e 0a             	mov    0xa(%bp),%di
>  1a0:	8b 56 08             	mov    0x8(%bp),%dx
$ diff ansi.old ansi.new
$
```


ghaerr:
Thanks for the ANSI C conversions @ccoffing! Moving source from K&R to ANSI helps a lot for stopping potential bugs with parameter mismatches, for sure.

> To be sure I didn't break anything in the process, I built master and then disassembled the object files, and then built this branch and disassembled, and then diff'd.

Wow, nice showing nothing changed; I don't know why the few instructions have been reordered, but nothing to worry about. 

A small issue is that functions with no parameters, such as:
```
int InitTerm(void);
```
changing to
```
int InitTerm();
```
leave open the possibility of passing arguments to `InitTerm`, which is allowed with ANSI C when no `(void)` used, when the function should in fact never be passed any. Is that your understanding also?

Thank you!

ccoffing:
> A small issue is that functions with no parameters, such as:

Oh, right... my C++ side is showing through.  I'll fix.

PR #1783, opened 2024-01-06
Size reductions

ccoffing:
This is a grab-bag of little optimizations that have accumulated.  It's only a few bytes and few cycles saved here and there, but I'm tired of sitting on them, and combined they cut 4k off the 2.8M media.

I think most of these changes stand on their own merits.  But a few like to `ls` were specifically sought out to bring the binary down to the next smaller filesystem block size.

### DIFF
--- a/elkscmd/ash/linenoise_elks.c
+++ b/elkscmd/ash/linenoise_elks.c
@@ -579,7 +579,6 @@ static void abFree(struct abuf *ab) {
 /* Helper of refreshSingleLine() and refreshMultiLine() to show hints
  * to the right of the prompt. */
 void refreshShowHints(struct abuf *ab, struct linenoiseState *l, int plen) {
-    char seq[64];
     if (hintsCallback && plen+l->len < l->cols) {
         int color = -1, bold = 0;
         char *hint = hintsCallback(l->buf,&color,&bold);
@@ -589,10 +588,11 @@ void refreshShowHints(struct abuf *ab, struct linenoiseState *l, int plen) {
             if (hintlen > hintmaxlen) hintlen = hintmaxlen;
             if (bold == 1 && color == -1) color = 37;
             if (color != -1 || bold != 0)
+            {
+                char seq[64];
                 fmtstr(seq, sizeof(seq), "\033[%d;%d;49m",bold,color);
-            else
-                seq[0] = '\0';
-            abAppend(ab,seq,strlen(seq));
+                abAppend(ab, seq, strlen(seq));
+            }
             abAppend(ab,hint,hintlen);
             if (color != -1 || bold != 0)
                 abAppend(ab,"\033[0m",4);
@@ -627,8 +627,7 @@ static void refreshSingleLine(struct linenoiseState *l) {
 
     abInit(&ab);
     /* Cursor to left edge */
-    strcpy(seq, "\r");
-    abAppend(&ab,seq,strlen(seq));
+    abAppend(&ab,"\r",1);
     /* Write the prompt and the current buffer content */
     abAppend(&ab,l->prompt,strlen(l->prompt));
     if (maskmode == 1) {
@@ -712,8 +711,7 @@ static void refreshMultiLine(struct linenoiseState *l) {
     {
         lndebug("<newline>");
         abAppend(&ab,"\n",1);
-        strcpy(seq, "\r");
-        abAppend(&ab,seq,strlen(seq));
+        abAppend(&ab,"\r",1);
         rows++;
         if (rows > (int)l->maxrows) l->maxrows = rows;
     }
--- a/elkscmd/cron/cron.c
+++ b/elkscmd/cron/cron.c
@@ -185,7 +185,7 @@ securepath(char *path)
     if (stat(path, &sb) != 0)
         fatal("%s: can't stat", path);
     if (stat(".", &sb1) != 0)
-        fatal(".: can't stat");
+        fatal("%s: can't stat", ".");
     if ((sb.st_dev != sb1.st_dev) || (sb.st_ino != sb1.st_ino))
         fatal("cwd not %s", _PATH_CRONDIR);
 }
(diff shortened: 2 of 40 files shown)


ghaerr:
Hello @ccoffing,

Nice work 4K byte savings on the distribution!! :)

I am guessing that some of the literal string splits were designed for use with string merging at link time? Also, pretty tricky work on using a bit vector for testing for the C0 control codes. Fast, though initially hard to understand until I got it. Thanks for leaving in the switch statement as documentation.

I found one line that might be in error, along with a couple other small comments posted separately for your comment.

Thank you!

elkscmd/ash/linenoise_elks.c
```diff
@@ -627,8 +627,7 @@ static void refreshSingleLine(struct linenoiseState *l) {
 
     abInit(&ab);
     /* Cursor to left edge */
-    strcpy(seq, "\r");
-    abAppend(&ab,seq,strlen(seq));
+    abAppend(&ab,"\r",2);
```
ghaerr: Shouldn't this be `abAppend(...,1`)?

elkscmd/cron/cron.c
```diff
@@ -225,7 +225,7 @@ printcrontab(crontab *tab, int nrtab)
                 for (j = 0; j < tabs[i].nrl; j++) {
                     printf("    ");
                     printtrig(&tabs[i].list[j].trigger);
-                    printf("%s", tabs[i].list[j].command);
+                    fputs(tabs[i].list[j].command, stdout);
```
ghaerr: Are you sure that replacing `printf` with `fputs` is actually smaller here? In the case where `printf` is already used, it seems this just drags in another function, rather than saving space as it would if `printf` could actually be replaced in full at link time. Same for a few other cases with printf->fputs substitution in this commit.

elkscmd/cron/crontab.c
```diff
@@ -219,15 +219,15 @@ visual(struct passwd *pwd)
             exit(0);
         seteuid(getuid());
         do {
-            fprintf(stderr, "Do you want to retry the same edit? ");
+            fputs("Do you want to retry the same edit? ", stderr);
```
ghaerr: Same here as commented above.

elkscmd/tui/cons.c
```diff
@@ -55,9 +55,7 @@ static void display(void)
                 fputs(buf, stdout);
             }
         }
-        if (r == LINES - 1)
-            printf("\r");
-        else printf("\n");
+        putc(r == LINES - 1 ? '\r' : '\n', stdout);
```
ghaerr, line 58: Similar to printf->fputs described above, doesn't this just drag in `fputc`, while also including `printf`? I would guess this results in large .o and final codefile ? (Same issue in tty-cp437.c).

elkscmd/cron/cron.c
ccoffing: Right, changing back to printf saves 16 bytes on the binary.  Done.

elkscmd/ash/linenoise_elks.c
ccoffing: Fixed.

elkscmd/cron/crontab.c
ccoffing: Turns out that this doesn't matter (function and binary are same size either way) since something else is already pulling fputs in.  In general, fputs should take fewer cycles when no formatting is required.  But since I don't know how/why fputs is getting pulled in already (maybe it can be trimmed out...) I'll revert this for now.

elkscmd/tui/cons.c
ccoffing: I think my initial motivation here was to save a few CPU cycles by avoiding the unnecessary format string parsing in vfprintf (which, when multiplied across screen redraws, might actually matter).

I just did a test:

Reverting cons.c and tty-cp437.c to printf as before, and doing a clean build:
```
$ du -b target/bin/
1337353	target/bin/
```

Clean build with cons.c and tty-cp437.c putc change as first presented here.  Slightly trading space for time:
```
$ du -b target/bin/
1337369	target/bin/
```

Then converting cons.c and tty-cp437.c to fputs/putc (no printf), and doing a clean build:
```
$ du -b target/bin/
1337321	target/bin/
```

So it seems that avoiding printf where possible in tty code is a [tiny] win, whether the goal is to optimize for space or time.

ghaerr:
After looking a bit more closely at `cons.c` and `tty-cp437.c`, I agree with your original finding that `printf` should be replaced by `fputs`, mainly because `fputs` is already used in both (as is `printf/sprintf`) so primarily a speed gain without pulling in another function. Also, fputs.c calls `putc` so using that costs nothing once `fputs` is used.

However - some executables are linked with $(TINYPRINTF) (elkscmd/lib/tiny_printf.c). In those cases, beware of the substitutions used above because tiny_printf.c only defines a smaller/fast `vprintf` but no `fputs` or `fputc`, so with those programs its more tricky to decrease their size. This could all be fixed by changing a few things in tiny_printf.c but care needs to be taken as it was written only for programs that call printf/sprintf but don't do anything else with stdio.

Thank you!

ccoffing:
> However - some executables are linked with $(TINYPRINTF) (elkscmd/lib/tiny_printf.c). In those cases, beware of the substitutions used above because tiny_printf.c only defines a smaller/fast `vprintf` but no `fputs` or `fputc`, so with those programs 

So regarding TINYPRINTF, it's possible with my changes that some binaries went up in size and some went down (but on average a win).  I'll keep that in mind.  Thanks!

ghaerr:
> regarding TINYPRINTF, it's possible with my changes that some binaries went up in size and some went down

Technically possible, but I don't think any of the programs you changed link with TINYPRINTF. In general, only very small programs that would otherwise use `write` / `errmsg` etc but can't because they need `sprintf` or a `printf` format designator have been linked with TINYPRINTF. It was invented just for that purpose, see elkscmd/rm.c for a good example. 

TINYPRINTF is definitely not compatible with anything that calls `fopen`, and no file I/O other than stdin/stdout.

PR #1785, opened 2024-01-09
Portable build fixes

ccoffing:
I have a branch in which I am toying with building libc and elkscmd more portably.  This gives me the ability to use better tools, including static analyzers and symbolic debuggers.  It's not done yet.  But here are a batch of fixes that have come from it so far.

### DIFF
--- a/elkscmd/ash/error.c
+++ b/elkscmd/ash/error.c
@@ -67,7 +67,6 @@ struct jmploc *handler;
 int exception;
 volatile int suppressint;
 volatile int intpending;
-char *commandname;
 
 
 /*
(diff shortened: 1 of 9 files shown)


ghaerr:
Looks good, thanks!

Its nice being able to use better tools to analyze ELKS source code, I have previously spent some effort trying to build better runtime trace mechanisms in elkscmd/debug for lack of better tools, and printf-debugging remains as a primary method sometimes.

The libc build is tied rather closely to the way that gcc uses multi-libs though, and might bring more issues should it change. Can your build coexist with the existing build system? What about analyzing kernel sources?

ccoffing:
> Can your build coexist with the existing build system? What about analyzing kernel sources?

Too early to know.  I'll keep pushing cleanups that fall out, and then maybe some tweaks to make the build system more consistent in places (which makes it easier to vary the compiler like this), but I don't know if more than that belongs in elks.

PR #1791, opened 2024-01-16
Screen portability fixes

ccoffing:
This modernizes screen a bit more, so it can build on Linux, which makes it a bit easier to shake some bugs out.

I don't think any of this changes actual behavior on ELKS.


ghaerr:
Thanks, nice cleanup!

PR #1794, opened 2024-01-25
Update meta driver file_ops function signatures

ccoffing:
I know this driver doesn't currently work but it's still good to tidy up.


ghaerr:
Thanks for the cleanup! Yes, it's doubtful this driver will ever be used, and it needs lots of work.

PR #1796, opened 2024-02-04
inital commit of elkscmd/beep

floriangit:
Every UNIX needs a beep. So there it is for ELKS.

A beep could be as simple as a `# echo "\a"` (or `putchar('a')`), but this does only work if your TTY is local, i.e. it won't work within a telnet or ssh or qemu session. Thus this solution.

Tested in qemu with the proposed changes in `qemu.sh` and on a real PC-Speaker of my 286 machine.

As for the `#if 1` in the code: I wanted to re-use `inb()` and `outb()` of `io.h`, but these variants did not work for me (no error, but also no sound). If you figure this out, feel free to change without notice (or let me know how to do).

Also: fixed a typo in `LICENSE`.


tyama501:
Hi @floriangit 

Thank you.
Probably PC-98 has different IO address,
so I will findout how to port after it is merged by @ghaerr .

May be I need to add codes with ifdef.

ghaerr:
Hello @floriangit,

Thanks for the PR, the C source looks good. However, I'd like to have it live in elkscmd/sys_utils (as it is system-dependent) rather than its own directory, as that makes things easier to maintain. That would entail moving the two Makefile build lines into elkscmd/sys_utils/Makefile. Would you like to do this, or should I after commit?

Regarding the `#if 1` in beep.c, its OK to remove it rather than bringing in the kernel `arch/io.h` file, I think. I'll take a further look as to why the latter doesn't work, but my first guess is that the port and value arguments are reversed between the two versions - did you notice that?

Hello @tyama501,

> Probably PC-98 has different IO address,

Yes, the `beep` and `silent` functions should be `#ifdef CONFIG_ARCH_IBMPC/PC98` etc, (and also `#include <autoconf.h>`) so that the utility is built correctly across platforms. This will also require rebuilding and an additional `rm elkscmd/sys_utils/beep.o` in `build.sh` for proper automatic image building.

Thank you!

floriangit:
Thanks @ghaerr and @tyama501, since you guys are more experienced I take your ideas and will try to address your suggestions! I'm a bit newbie with the github workflow (but OK with git), so let's see how I can come up with "moving" things around now. If stuck, I'll shout.

I will verify your port<->value reversed theory @ghaerr and check to ignore it to just keep my version and remove the `#if 1` thinggie or just fix it if really reversal is the issue.

Thanks.

ghaerr:
> I'm a bit newbie with the github workflow (but OK with git), so let's see how I can come up with "moving" things around

Sounds good. A `git mv elkscmd/beep/beep.c elkscmd/sys_utils`, followed by a `git push` will work, for example.

floriangit:
> that the port and value arguments are reversed between the two versions - did you notice that?

Yes...now that you're saying it :-) ....fixed!

ghaerr:
Thanks @floriangit, looks great. I'll commit this now and see whether @tyama501 wants to update it for conditional compilation on PC-98 as well.

tyama501:
It seems that PC-98 use TIMER1 of 8253, which is not defined in ports.h yet.
https://github.com/ghaerr/elks/blob/master/elks/include/arch/ports.h

io port is 0x73 or 0x3FDB.

On/Off is controlled by 8255A and
io port is 0x37

I will try it when I have time.

tyama501:
BTW, isn't it fun if it can change frequency and time by option?
To play some beep music:)

It seems that Ubuntu beep can do that.

floriangit:
> if it can change frequency and time by option

Feel free to play around, frequency and durations are just magic numbers in the code ;-)

Maybe add a Japanese かわいい special sound (like the JR lines in Tokyo with those really catchy plays depending on arrival/departure at station). But that's more than 'beep' I guess....

tyama501:
It seems that someone made nice music:)
https://github.com/ShaneMcC/beeps

ghaerr:
> It seems that someone made nice music:)

Feel free to make `beep` compatible with those command line options, looks like quite the music library :)

ghaerr:
> io port is 0x73 or 0x3FDB.
> It seems that PC-98 use TIMER1 of 8253, which is not defined in ports.h yet.

For both PC-98 and IBM PC refinements to `beep`, it would be nice to have all I/O addresses as constants in the `ports.h` file, good idea.

tyama501:
Hello @ghaerr ,

It seems that the "beep" is included in fd1440.img but it is not included in fd1440-fat.img for IBM-PC.
Is this intended?

beep/beep                       :sysutil

ghaerr:
> It seems that the "beep" is included in fd1440.img but it is not included in fd1440-fat.img for IBM-PC.
> beep/beep :sysutil

It looks like the commit didn't include `:1440k` at the end of the line in elkscmd/Applications. I will fix that.

PR #1803, opened 2024-02-07
[build,df] Various minor fixups

ghaerr:
Require CONFIG_ASYNCIO for Direct floppy driver.
Use unsigned char booleans in DF driver.
Cleanup some header files.
Move misc_utils/fdtest to test/other/test_fd.


toncho11:
Is switching to the direct floppy driver DF documented somewhere?

ghaerr:
> Is switching to the direct floppy driver DF documented somewhere?

I am guessing probably not. Would you like to take a pass at it? 

We haven't yet changed the standard configuration to include CONFIG_ASYNCIO, although I think ELKS is ready for that. When the system is compiled with that enabled and the DF driver included, the kernel image is larger, and all that is really required is to set `root=/dev/df0` in /bootopts. For compatibility reasons, I don't think its a good idea to enable the DF driver by default, however.

What other details might need to be documented? Have you played around with it and found more specific benefits that should be mentioned?

toncho11:
> We haven't yet changed the standard configuration to include CONFIG_ASYNCIO

You mean it does not appear in make menu config? Can you add it please? I can make a wiki page. So one might need to remove something to fit on a standard image?


So far I booted Elks on a Zenith Data Systems Multisport 286e with standard image from github and fd.

toncho11:
So there is no menu config option for neither the df nor config_asyncio? Currently it is done in a makefile?

toncho11:
OK. So I made the page: https://github.com/ghaerr/elks/wiki/Direct-Floppy-Driver-DF
I left space for how to compile the kernel with DF enabled.

ghaerr:
> So I made the page: https://github.com/ghaerr/elks/wiki/Direct-Floppy-Driver-DF

Great job, thanks!! I'll add to the section on how to configure the kernel for it.

> So there is no menu config option for neither the df nor config_asyncio?

Yes, there is a menu config option to select the direct floppy disk driver. At this moment, it doesn't automatically turn on CONFIG_ASYNCIO. I will change that so it does.

> So one might need to remove something to fit on a standard image?

I think everything fits, but will double-check building the 360k image to make sure it does, for when we automatically include it in the next release.

ghaerr:
There is currently a `make menuconfig` option to enable Async I/O in Kernel Settings. Also, I checked the build and the kernel and applications fit on a 360k floppy with both async I/O and the direct floppy driver enabled.

PR #1806, opened 2024-02-10
PC-98 beep support

tyama501:
Hello @ghaerr and @floriangit ,

I haven't added option to change the frequency and time,
but I have added the code for PC-98 according to 
https://github.com/ghaerr/elks/pull/1796

I could hear beep from PC-98 NekoProject/W emulator as below.
I also replaced IBM port numbers to defines of ports.h,
but I couldn't check it since my qemu on windows cannot enable speaker.
Could you double check it?

https://github.com/ghaerr/elks/assets/61556504/5932d298-2095-49d0-9d5b-d07a9188a1d8

Thank you!

### DIFF
--- a/elks/include/arch/ports.h
+++ b/elks/include/arch/ports.h
@@ -56,6 +56,10 @@
 #define TIMER_DATA_PORT 0x71		/* data port    */
 #define TIMER_IRQ	0		/* can't change*/
 
+/* bell*/
+#define TIMER1_PORT 0x3FDB		/* timer 1 data port for speaker frequency*/
+#define PORTC_CONTROL 0x37		/* 8255A Port C control*/
+
 /* serial, serial-pc98.c*/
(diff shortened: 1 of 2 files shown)


floriangit:
I've applied your patch locally to test it. Still all beeps fine on qemu (from Debian) as well as on my 286-PC, @tyama501 !
The only nitpick I can find is in `ports.h`: 

+/* bell*/
and
 /* serial, serial-pc98.c*/

missing a white space before the comment ends in `*/`

tyama501:
Hi @floriangit 

Great! Thank you for quick testing!

ghaerr:
Hello @tyama501,

This looks good. When you add the options for frequency and time, perhaps we can also use `#ifdef CONFIG_ARCH_IBMPC` for the IBM PC code, rather then using an `#else` clause. This will allow for nothing to happen for the 8018X port, for instance.

Thank you!

PR #1807, opened 2024-02-10
Screenshot ZDS 286

toncho11:
Elks work well on this ZDS laptop :) equipped with 286 CPU.


ghaerr:
Thanks for the picture @toncho11! That's definitely an old Zenith laptop. I used to have a couple of their older desktops as well.

PR #1808, opened 2024-02-10
[build] Misc fixes

ghaerr:
Swaps `txtetris` for `nxlandmine` on 1440k floppy #1805.
Adds `beep` to 144k floppy https://github.com/ghaerr/elks/pull/1796#issuecomment-1937039976.
Enables CONFIG_ASYNCIO when direct floppy driver selected https://github.com/ghaerr/elks/pull/1803#issuecomment-1937103152.

### DIFF
--- a/elks/arch/i86/drivers/block/config.in
+++ b/elks/arch/i86/drivers/block/config.in
@@ -10,6 +10,9 @@ mainmenu_option next_comment
 	bool '  BIOS preset floppy types'	CONFIG_BLK_DEV_BFD_HARD n
 	bool '  BIOS hard drive support'	CONFIG_BLK_DEV_BHD	y
 	bool '  IDE hard drive CHS probe'	CONFIG_IDE_PROBE	y
+	if [ "$CONFIG_BLK_DEV_FD" == "y" ]; then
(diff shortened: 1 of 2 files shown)


toncho11:
I think you can add builds for the DF driver as well in Github Actions:
* direct-floppy-360-minix-needs-testing
* direct-floppy-1440-minix-needs-testing

ghaerr:
> I think you can add builds for the DF driver as well in Github Actions:

The Github Actions are already quite slow... perhaps just add the DF driver and CONFIG_ASYNCIO into the existing Github Actions builds, as well as the default builds using `make menuconfig`. That should do the same thing.

This will allow a kernel and images to be produced that will work for both BIOS and DF drivers (still defaulting to BIOS for backwards compatibility unless DF enabled in /bootopts), which then might be a good idea to increment our release to 0.8.0?

toncho11:
Yes. If the DF driver is not taking too much kernel space I think it is a very good idea to add it to the default build, so that people can test (by changing the config once they boot the first time). 

Yes, I am always for regular releases. So for me 0.8.0 makes sense. It gives the beat a project needs.

PR #1809, opened 2024-02-15
Printing

ccoffing:
Basic queuing and printing now works.

I tested on a 386SX, both with using the BIOS-defined ports and probing.  lp and lpd both function.

I have an original IBM PC that I will eventually get around to testing on, and I'd like to improve the lp* tools, but that's future.

elks-lp


ghaerr:
Hi @ccoffing,

Wow, very nice major rewrite and cleanup of the lp driver and utilities, thanks!! This was definitely needed. 
Can't wait to try it :)

> I'd like to improve the lp* tools, but that's future.

A very old `pr` that I used to use added a header/footer with date stamp, filename and page number which was nice for printing listings, etc. Perhaps that can be added to `lp` if you get around to it.

> I tested on a 386SX, both with using the BIOS-defined ports and probing.

So when does probing get used now, versus using the BIOS-defined ports? [EDIT: I see now, when !BIOS_PORTS.]

elks/arch/i86/drivers/char/lp.c
```diff
@@ -32,219 +31,155 @@ static struct lp_info ports[LP_PORTS];
...
-    0x2E8, 0, 0,
-    0x2F0, 0, 0,
-    0x2F8, 0, 0,
-    0x3E0, 0, 0,
-    0x3E8, 0, 0
+/* Probe the standard ports. */
+static struct lp_info ports[LP_PORTS] = {
+    {(char*)0x3BC, 0},
+    {(char*)0x378, 0},
+    {(char*)0x278, 0},
```
ghaerr, line 38: I think this older use of casting to pointer was used because originally the argument of `outb_p` required it. Perhaps this should also be cleaned up to use `unsigned short io` in `struct lp_info` instead. I've cleaned up other parts of ELKS that also used pointer<->int casting. Another reason for this sort of thing throughout the older kernel was due to certain optimizations the old BCC compiler would perform on pointers, none of which is beneficial now.

ccoffing:
> So when does probing get used now, versus using the BIOS-defined ports? [EDIT: I see now, when !BIOS_PORTS.]

Right; it's still hardcoded as before.  I manually toggled it both ways to prove each works (at least for my specific system).  Maybe this BIOS_PORTS configuration should be derived from the CONFIG_ARCH_* setting, or perhaps it should be exposed as another tunable under Kernel & Hardware > System.  I assume that some more custom systems would not have the same BIOS and would want to probe instead.

elks/arch/i86/drivers/char/lp.c
ccoffing: Interesting... I remember the warning that required the cast popped up later, and I didn't look closely at why.  Agreed that would be cleaner as `unsigned short io`.  I'll open a PR for that.  Thanks.

PR #1812, opened 2024-02-16
[beep] Add -f and -l option

tyama501:
Hi @ghaerr and @floriangit ,

I added -f option for frequency and -l option for duration.

I could play beginning of Happy Birthday referring to
https://github.com/ShaneMcC/beeps/blob/master/happybirthday.sh

https://github.com/ghaerr/elks/assets/61556504/276889ed-5d4e-4444-b636-fe5fe1846d4a

-n is not supported, so we need to write "beep" in each line of script.
Also, the frequency needs to be integer.

Enjoy.
Thank you!

### DIFF
--- a/elkscmd/sys_utils/beep.c
+++ b/elkscmd/sys_utils/beep.c
@@ -11,7 +11,7 @@
 #include "arch/io.h"
 #include "arch/ports.h"
 
-static void beep(int freq)
+static void beep(unsigned int freq)
 {
 #ifdef CONFIG_ARCH_PC98
     unsigned int d;
@@ -76,10 +76,10 @@ int main(int ac, char **av)
               if(av[i][0] == '-') {
                    switch(av[i][1]) {
                    case 'f':                    /* Frequency */
-                       freq = atoi(&av[i][2]);
+                       freq = atol(&av[i][2]);
                        break;
                    case 'l':                    /* Duration in millisecond */
-                       duration = atoi(&av[i][2]);
+                       duration = atol(&av[i][2]);
                        break;
                    }
               }
--- a/elkscmd/sys_utils/beep.c
+++ b/elkscmd/sys_utils/beep.c
@@ -69,15 +69,28 @@ void beep_signal(int sig)
 int main(int ac, char **av)
 {
     long duration = 333L;
+    int freq = 1000;
+    int i;
     if(ac >= 2) {
-        duration = atoi(av[1]);
+         for(i = 1; i < ac; i++) {
+              if(av[i][0] = '-') {
+                   switch(av[i][1]) {
+                   case 'f':                    /* Frequency */
+                       freq = atoi(&av[i][2]);
+                       break;
+                   case 'l':                    /* Duration in millisecond */
+                       duration = atoi(&av[i][2]);
+                       break;
+                   }
+              }
+         }
     }
 
     signal(SIGKILL, beep_signal);
     signal(SIGINT,  beep_signal);
     signal(SIGTERM, beep_signal);
 
-    beep(1000);
+    beep(freq);
     usleep(duration * 1000L);
     silent();
 
--- a/elkscmd/sys_utils/beep.c
+++ b/elkscmd/sys_utils/beep.c
@@ -73,7 +73,7 @@ int main(int ac, char **av)
     int i;
     if(ac >= 2) {
          for(i = 1; i < ac; i++) {
-              if(av[i][0] = '-') {
+              if(av[i][0] == '-') {
                    switch(av[i][1]) {
                    case 'f':                    /* Frequency */
                        freq = atoi(&av[i][2]);
--- a/elkscmd/sys_utils/beep.c
+++ b/elkscmd/sys_utils/beep.c
@@ -70,29 +70,37 @@ int main(int ac, char **av)
 {
     long duration = 333L;
     long freq = 1000L;
-    int i;
-    if(ac >= 2) {
-         for(i = 1; i < ac; i++) {
-              if(av[i][0] == '-') {
-                   switch(av[i][1]) {
-                   case 'f':                    /* Frequency */
-                       freq = atol(&av[i][2]);
-                       break;
-                   case 'l':                    /* Duration in millisecond */
-                       duration = atol(&av[i][2]);
-                       break;
-                   }
-              }
-         }
-    }
+    int nbeep;
+    int i = 1;
 
-    signal(SIGKILL, beep_signal);
-    signal(SIGINT,  beep_signal);
-    signal(SIGTERM, beep_signal);
+    do {
+        nbeep = 0;
+        while (ac >= 2 && !nbeep) {
+            if(av[i][0] == '-') {
+                switch(av[i][1]) {
+                case 'f':                    /* Frequency */
+                    freq = atol(&av[i][2]);
+                    break;
+                case 'l':                    /* Duration in millisecond */
+                    duration = atol(&av[i][2]);
+                    break;
+                case 'n':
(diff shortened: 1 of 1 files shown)


tyama501:
Oops, Line 76 should be ==.
I will fix it.

elkscmd/sys_utils/beep.c
```diff
@@ -69,15 +69,28 @@ void beep_signal(int sig)
...
     if(ac >= 2) {
-        duration = atoi(av[1]);
+         for(i = 1; i < ac; i++) {
+              if(av[i][0] = '-') {
+                   switch(av[i][1]) {
+                   case 'f':                    /* Frequency */
+                       freq = atoi(&av[i][2]);
+                       break;
+                   case 'l':                    /* Duration in millisecond */
+                       duration = atoi(&av[i][2]);
```
ghaerr: Since `duration` is declared `long`, this should probably use `atol` here. A similar issue could apply with `freq`, which is only valid to 32767. So `beep` and `freq` might want to be declared `beep(unsigned int freq)`. 

In order to keep the code file quite small, one might want to use `atol` in both places, and truncate the long value into an unsigned int for `freq`.

floriangit:
> freq, which is only valid to 32767

Indeed dogs are able to hear those frequencies (I cannot). Of course ELKS users _could_  use beep for their dogs, so :+1:  to better fix it! :-)

tyama501:
Thank you!

elkscmd/sys_utils/beep.c
ghaerr: Now I notice that `freq` is used as following in `beep`:
```
d = 2457600 / freq;
```
which means `freq` will have to be converted to long anyways. Thus `freq` should probably be declared long in both beep and the main code, and also have the above rewritten to:
```
d = 2457600L / freq;
```
The reason for this is that the `ia16-elf-gcc` compiler sometimes does the wrong thing with int constants that are > 65535.

ghaerr:
> -n is not supported, so we need to write "beep" in each line of script.

Why not support `-n` so that the sound scripts can be run unmodified? :)

ghaerr:
> Why not support -n so that the sound scripts can be run unmodified? :)

Oops - I see - they use floating point numbers. Well, the float portion could be ignored, and I think `atol` will properly return just the integer portion of the floating point number.

elkscmd/sys_utils/beep.c
floriangit: > The reason for this is that the ia16-elf-gcc compiler sometimes does the wrong thing with int constants that are bigger than 65535.

Hmm, are you sure about this bigger than 65535? I don't remember now exactly, but in main.c I suffixed L to the duration hardcoded value and to the hardcoded value in usleep. And IIRC there was a necessity to it I did not understand at that time (I thought "oh wait 16bit, that's why....?"), both are well below 65535...sorry I cannot verify my theory right now.

ghaerr: In C, types that are different sizes (e.g. `char`, `int`, `long`) or are different signedness (e.g. `int` vs `unsigned int`) are automatically promoted to 1) the size of the largest sized element in the expression, and 2) to unsigned if any element is unsigned. Types that are passed to functions are also automatically promoted to the defined argument signedness and size.

For "integer" constants, the question is: what is an integer? For `ia16-elf-gcc`, it is `int`, which is 16-bits. This can mean that the compiler silently truncates a long to an int if there are no other elements in the expression, or if it is not being passed to a pre-defined function. This also occurs with signedness: 65535 requires a long, even though an unsigned int could take that value. So is it signed or unsigned? IIRC, the compiler assumes `int` unless otherwise expressed, which would mean that the value gets truncated to 16 bits. This may or may not be what is desired, as 65535 still fits in 16-bits if the value is interpreted as unsigned. However, if an additional element is type `int` and not `long`, then the arithmetic won't be done in 32 bits.

For these reasons, as well as clarity, in the ELKS kernel, most all literal constants > 32767 are suffixed with `L` so as to ensure proper internal and external computation.  In order to know for sure, one can look at the ASM output using something like:
```
$ ia16-elf-objdump -D -r -Mi8086 beep.o
```

floriangit: Thanks @ghaerr !

tyama501:
Hi @ghaerr and @floriangit 

I have added -n option.
Now, I could play the happy birthday without modification.

Here is the sound from PC-9801UV21.

https://github.com/ghaerr/elks/assets/61556504/b0ddc694-ea3b-4f7f-8a78-03706c3cce01

Thank you.

floriangit:
Great stuff @tyama501. We could add in beep.c:2 your contributions if you want. Also beep.c:4 could have an update regarding the supported cmdline options. A small "-h" command line option "usage" with an example to Happy Birthday in main() maybe as well? I can do those cleanups if you want after Greg merged your changes?

Ouh and Happy Birthday NEC-PC98!

ghaerr:
Looks great @tyama501! Let me know if you'd like to merge now, or add anything @floriangit suggested.

tyama501:
Thank you @floriangit 

That would be nice to add option usage.

The happy birthday, I just runned sh script from the web ./happybirth.

@ghaerr , please merge it.

Thanks

tyama501:
I forgot to mention that it seems that the size gets tripled if printf is used.
Are there small printf for usage?

ghaerr:
> I forgot to mention that it seems that the size gets tripled if printf is used.
> Are there small printf for usage?

Thanks for bringing that up - we don't want to use `printf` unless really necessary. There's two ways we can write tiny commands for ELKS, one uses a macro to display messages to stderr using `write` like this (see elkcmd/file_utils/mkdir.c for example):
```
#define errstr(str) write(STDERR_FILENO, str, strlen(str))
...
errmsg("usage: mkdir [-p] directory [...]\n");
// or
errstr(argv[i]); 
```
If `printf` is required for ease of display of formatted output using format strings, then a tiny printf can be used (see elkscmd/lib/tiny_vprintf.c), by adding the following to the link line (see sys_utils/Makefile for example):
```
chmem: chmem.o $(TINYPRINTF)
    $(LD) $(LDFLAGS) -o chmem chmem.o $(TINYPRINTF) $(LDLIBS)
```
I would suggest using `errmsg` it at all possible to keep the executable size very small. In particular, it is nice to keep an executable size <= 1024 bytes, so that only one MINIX block is allocated for the application.

floriangit:
> In particular, it is nice to keep an executable size <= 1024 bytes

The binary exe is now 1040 bytes. When trimming it down to omit the SIGKILL handler (which is handled by the kernel as per POSIX and has no meaning in an application really, since it can never be caught), I can bring it down to 1024 bytes, yaaay.

Thus I'll rather update the "-h" in form of a comments update only in the .c file to play happy birthday and general guidance and remove the SIGKILL stuff.

Unless you object to?

ghaerr:
> The binary exe is now 1040 bytes. When trimming it down to omit the SIGKILL handler

Wow, that did get big in a hurry... the first version was quite a bit smaller. I'm kind of wondering what is taking up so much space?

>  omit the SIGKILL handler (which is handled by the kernel as per POSIX and has no meaning in an application really, since it can never be caught),

Agreed - remove it, as SIGKILL always causes the executable to cease termination, without a signal callback.

> Thus I'll rather update the "-h" in form of a comments update only in the .c file to play happy birthday and general guidance and remove the SIGKILL stuff.

Go ahead. It might be worth taking a look at what's using up the space... this can be done by generating a map file by adding the following to the link line: `-Wl,-Map=beep.map`. `beep_signal` can probably have the `switch` removed and just call `silent` and `exit`.

PR #1813, opened 2024-02-18
[beep] make it round

floriangit:
1) adding author and adding documentation
2) remove the SIGKILL handling, since this will never get caught in userland

We are now exactly with 1024bytes of the executable on x86_16: 0x3c0 for `.text`, 0x10 for `.data` and 0x30 for `.bss`
Main sizes analysed by map-files in `.text` are in almost same sizes:

1. signal handling (also the signal callback is inserted)
2. `atol`
3. `usleep`
4. divisions (`ldiv` from ia16-ldivmodu.o and `udiv`/`umod` from libgcc.a)

IMO removing the switch/case from `beep_signal` would deterior code readability, since now it's easy to catch what is handled and what not. Your mileage may vary, it would save only 16 bytes, though.

Happy beeping!

### DIFF
--- a/elkscmd/sys_utils/beep.c
+++ b/elkscmd/sys_utils/beep.c
@@ -1,7 +1,14 @@
 /* beep - This file is part of the ELKS project
- * authors: Florian Zimmermann et al.
+ * authors: Florian Zimmermann, Takahiro Yamada et al.
  * Free software "as-is"
- * Takes one optional parameter: the beep duration
+ *
+ * By default (no parameters) one beep is output, 
+ * this can be adjusted by using parameters that can be consecutives of each:
+ * -f frequency as digit
+ * -l length as digit in milliseconds
+ * -n (to indicate the next beep)
+ *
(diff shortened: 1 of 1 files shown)


floriangit:
https://github.com/ghaerr/elks/assets/90066055/e4127028-51b7-4c78-9534-2e119409dd85

Cannot beat the quality of a 40-year old PC speaker! :1st_place_medal:

ghaerr:
Thanks for the update!

> removing the switch/case from beep_signal would deterior code readability, since now it's easy to catch what is handled and what not. 

True, but also only the two signals that are specified in calls to `signal` will ever be routed through `beep_signal`.

toncho11:
There can be a C function provided by the compiler or ELKS source code that does the beep. It could be added to https://gitlab.com/tkchia/libi86 of @tkchia . Hmm, the libi86 has been moved from github to gitlab.

toncho11:
Is this a beep function https://gitlab.com/tkchia/libi86/-/blame/master/common/libi86/internal/graph.h#L273 ?

floriangit:
> Is this a beep function https://gitlab.com/tkchia/libi86/-/blame/master/common/libi86/internal/graph.h#L273 ?

Hi @toncho11, I think this is same as I mentioned in the initial commit:
https://github.com/ghaerr/elks/pull/1796

Yes, outputting `\a` will beep, but only if bound to a local TTY.

PR #1822, opened 2024-03-07
[cmds] Remove fsck-dos FAT32 support allowing it to work on floppies

ghaerr:
Since FAT12 and FAT16 cluster sizes are 16-bits, removing support (at least for now) from `fsck-dos` allows for reducing the memory size of `struct fatEntry` by about half. This in turn seems to allow `fsck-dos` to run on 1440k floppies and repair them without running out of memory.

Tested on @tyama501's floppy image posted in https://github.com/ghaerr/elks/discussions/1819#discussioncomment-8696583.

Here's a screenshot of @tyama501's vi-damaged floppy image being repaired which matches the screenshot of the repair working succcesfully with the original fsck_msdos on his host PC (except FAT compare is being skipped):
fsck-dos ok

Obviously we need more testing, but things are looking better now. I would be interested in test cases for more badly damaged FAT filesystems in FAT16 or FAT12 format.

### DIFF
--- a/elkscmd/fsck_dos/Makefile
+++ b/elkscmd/fsck_dos/Makefile
@@ -11,6 +11,8 @@ include $(BASEDIR)/Make.rules
 ###############################################################################
 
 LOCALCFLAGS = -Dlint -DELKS=1
+# remove next line when FAT32 supported
+LOCALCFLAGS += -Wno-shift-count-overflow
 
 PRGS = fsck-dos
 
--- a/elkscmd/fsck_dos/boot.c
+++ b/elkscmd/fsck_dos/boot.c
@@ -232,7 +232,7 @@ readboot(dosfs, boot)
 		boot->ClustMask = CLUST16_MASK;
 	else {
 		pfatal("Filesystem too big (%lu clusters) for non-FAT32 partition",
-		       boot->NumClusters);
+		       (U32)boot->NumClusters);
 		return FSFATAL;
 	}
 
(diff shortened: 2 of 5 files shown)


toncho11:
So it supports only FAT32 now?

tyama501:
Hi @toncho11 
I think it means now only support FAT12 and 16.

Thank you @ghaerr !
I will think about adding 1232K feature,
so that I can test with real hardware.
(signature may not be checked any more since we skipped FAT check)

EDIT:
It seems that the signature check is in readboot so it still exist.
I may be able to test 2DD 720 KB disk which is 512 B sector.

toncho11:
Yes, I see. It is in the name of the PR.
So there should be a warning message that it is not supported if tried on FAT32.

ghaerr:
Hello @tyama501,

> I will think about adding 1232K feature
> It seems that the signature check is in readboot so it still exist.

Yes, but (for now) I am thinking that if you just move the boot block signature check to after where the BPB information is read, then you might be able to check signature only for non-1024-byte sector floppies. (If you want to later add an additional disk read when `boot->BytesPerSec == 1024`, that would check signature for 1232 floppies):
```
    boot->FATsecs = boot->FATsmall;
...
    if ((boot->BytesPerSec != 1024) && (block[510] != 0x55 || block[511] != 0xaa)) {
        pfatal("Invalid signature in boot block: %02x%02x", block[511], block[510]);
                exit(2);
    }
    if(boot->BytesPerSec == 0){
        perror("Not valide BytesPerSec");
        exit(2);
    }
```
Then there may not be a need to rewrite much as it appears that the rest of the program uses `boot->BytesPerSec` instead of a 512 byte constant.

Thank you!

ghaerr:
With more thought, perhaps something like this would work well. This would allow 512-byte and 1024-byte sectors with a (probably) valid BPB, and only check the boot signature on 512-byte sectored floppies. The error message is changed in case a MINIX disk check is attempted:
```
    boot->FATsecs = boot->FATsmall;
...
    if (boot->BytesPerSec != 512 && boot->ByesPerSec != 1024){
        perror("Invalid DOS format, bad bytes per sector");
        exit(2);
    }
    if ((boot->BytesPerSec == 512) && (block[510] != 0x55 || block[511] != 0xaa)) {
        pfatal("Invalid signature in boot block: %02x%02x", block[511], block[510]);
                exit(2);
    }
```

PR #1827, opened 2024-03-24
[kernel] Disallow reschedule during kernel startup

ghaerr:
This problem was identified by @Mellvik in https://github.com/Mellvik/TLVC/pull/51.

If a kernel waits for I/O complete during `bread` at startup, which will currently only happen when an interrupt-driven hard disk driver reads a hard disk partition table, `wait_for_buffer` may try to sleep/reschedule the current task, which will always be the idle task.

This alternative fix to https://github.com/Mellvik/TLVC/pull/51 disallows a task switch during kernel startup (when last allocated process ID will be <= 0), while still checking and printing a warning message should a device driver attempt to sleep from an interrupt handler.

### DIFF
--- a/elks/include/linuxmt/sched.h
+++ b/elks/include/linuxmt/sched.h
@@ -114,6 +114,7 @@ extern __task task[MAX_TASKS];
 
 extern volatile jiff_t jiffies; /* ticks updated by the timer interrupt*/
 extern __ptask current;
+extern pid_t last_pid;
 
 extern struct timeval xtime;
 extern jiff_t xtime_jiffies;
--- a/elks/kernel/fork.c
+++ b/elks/kernel/fork.c
@@ -10,11 +10,11 @@
 
 int task_slots_unused = MAX_TASKS;
 struct task_struct *next_task_slot = task;
+pid_t last_pid = -1;
 
 static pid_t get_pid(void)
 {
     register struct task_struct *p;
-    static pid_t last_pid = -1;
 
     goto startgp;
     do {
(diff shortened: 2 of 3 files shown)


Mellvik (commented):
Great fix, @ghaerr! In order to make sure I understand it completely <pedantic>: It seems to me `last_pid` cannot be 0 (per `get_pid()`), is the '=' in  `last_pid <= 0` intentional or accidental? </pedantic>

ghaerr:
@Mellvik: Actually the test only needs to be `last_pid == 0`, but <= was used because `last_pid` is initialized to -1. 

The potential confusion is that `last_pid` is the last-used PID, not the next PID. So, after sched_init() is called and the idle task created within kfork_proc(), `last_pid` is actually 0. This remains the case until the init process is created with PID 1 at the end of the kernel startup. Should the PID ever wrap via twos-complement to negative, it skips the 0 idle task and starts with 1 again, as you point out.

Mellvik:
Thanks @ghaerr, that completes the picture.
Again, nice fix - and (unsurprisingly) working...

ghaerr:
Thanks @Mellvik! In this case, you did most of the work identifying the problem in the first place. I am imagining the timely process of driving down this rabbit hole starting with removing a single printk that caused the problem!! Of course, that's half the fun of it, and leads one through lots of kernel code one wasn't necessarily that familiar with. Good way to learn the kernel. I certainly hadn't foreseen the problem of calling `bread` in an async kernel from task 0.

PR #1829, opened 2024-03-24
[kernel] Add DF driver to standard IBMPC build

ghaerr:
Adds DF (direct floppy) driver to standard IBM PC build. Defaults to OFF unless `root=df0` set in `/bootopts`.

Sets CONFIG_ASYNCIO default ON (required for direct floppy driver, no effect otherwise).

Change boot version string to `ELKS 0.8.0-pre1` in anticipation of Version 0.8.0 release.


ghaerr:
The PC-98 version should probably stay with CONFIG_ASYNCIO off and no DF driver for now, until it is more tested with PC-98. Given what I've learned about the PC-98 BIOS, it may not be possible to completely move away from it.

toncho11:
I just tried that by uncommenting `root=df0 ` and then ELKS starts and then it halts when trying to mount the root. 
I am no sure if this is a problem with how I edit /bootopts or just uncommenting this line is not enough to switch to df0 driver?

I can never remember the edit commands, there is no man page and no help in the command on a 360 kb image. Practically edit command is absolutely unusable if you do not have a super strong memory to remember the edit commands. Also you can not search on internet because that is not its real name :). So good luck with editing a file.

ghaerr:
Thanks @toncho11 can you send a screenshot of the boot screen?

toncho11:
Or maybe it is for 1440 only? And what was the save and exit commands for edit?

ghaerr:
> how I edit /bootopts or just uncommenting this line is not enough to switch to df0 driver?

Yes, just uncomment this line should work, for 360k as well as 1440k.

The easiest way to edit would be to use a FAT rather than MINIX disk and edit on DOS, at least for now.

toncho11:
Yes, I am on fat. The edit on dos adds something on each line I think \r\n.
I used Norton Commander to just remove the # before root=fd0.

And what was the save and exit commands for edit?

toncho11:
And does it work in QEMU?

ghaerr:
> And does it work in QEMU?

Yes, before the last two PRs. I will try again.

> The edit on dos adds something on each line I think \r\n.

Got it. I'll look to see if that's a problem.

> And what was the save and exit commands for edit?

I think ^S and ^Q.

Please send screenshot so I can see what is happening. Thanks for your testing!

toncho11:
It is ctrl + x, but it still does not switch to df0 and it halts as before.

I see you modified only the 1440 config file last time? I think the driver is not compiled in.

toncho11:
I used the last image from github: https://github.com/ghaerr/elks/actions/runs/8472136457

ghaerr:
Ok the problem is  that there is no DF driver in that image, it seems. Will investigate. The df: driver init is not showing and mount then fails.

ghaerr:
The 360k build should use the same kernel as 1440k config but apparently is not. Will investigate later today, thank you!

ghaerr:
@toncho11, all of the issues you brought up in your testing should be fixed in the GitHub images generated by #1841.

Thank you!

PR #1841, opened 2024-03-30
[config] Update configurations for standard IBM PC build

ghaerr:
Fixes automatic IBM PC build problems identified by @toncho11 in https://github.com/ghaerr/elks/pull/1829#issuecomment-2027407900.

The `build.sh` script used by GitHub CI was not explicitly using `ibmpc-1440.config` to build the default image; instead it was using `make defconfig`, which didn't include the DF driver nor ASYNCIO. This has been fixed.

Also, ELKS now allows CRLF (DOS) line formatting in /bootopts, allowing it to be edited with an MSDOS editor on FAT disks, if necessary.


toncho11:
It all works now root=df0.
I also see that the ls command benefits nicely from file system buffering. 
Thank you @ghaerr !

ghaerr:
Thanks for your testing @toncho11. I'm glad we uncovered and solved a problem with our GitHub automatic builds that has been there from the start.

PR #1851, opened 2024-04-06
[boot] Add media descriptor 0xfe and boot signature in image FD1232 for PC-98

tyama501:
Hello @ghaerr ,

This is the PR for adding the media descriptor 0xfe and the boot signature in image FD1232 for PC-98 as we talked in 
https://github.com/ghaerr/elks/discussions/1819

I could make MS-DOS for PC-98 readable image by adding -m option to mformat as you mentioned.

Also, I have added the boot signature to boot_sect.S.
I was misunderstanding and there were 13 Bytes left in the first 512 Bytes so I could add it without shrink.
Now the current fsck-dos works to the disk without modification.

Thank you!

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -591,6 +591,11 @@ tvram_x:
 end_of_code:
 //------------------------------------------------------------------------------
 
+#if defined(CONFIG_IMG_FD1232)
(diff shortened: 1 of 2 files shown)


ghaerr:
This looks great, thanks @tyama501!

PR #1855, opened 2024-04-07
[cmds] Port ttyclock to ELKS

ghaerr:
First pass at porting `ttyclock` to ELKS. It still needs some work and has no keyboard options or color.

Quick hack at getting ELKS curses to mostly handle ttyclock's code, although true windows/subwindows are not supported.

Also, at 18K, size is huge for ELKS! 

ttyclock


toncho11:
Wow! Bravo! I like it!

PR #1863, opened 2024-04-08
[libc] Add strftime to C library, enhance ttyclock with colors

ghaerr:
Adds `strftime` to C lib which allows `ttyclock -f%F` etc to customize displayed date formats.

Enhance our curses emulation library to more properly handle colors, allowing `ttycolor` to change colors using `1`-`7` keys. Default color is now green (although I kind of prefer normal reverse video white, which can be set using 
`ttyclock -C7`).


tyama501:
Wow, interesting.

I will try compile ELKSmoria with new curses to see what happens.

PR #1864, opened 2024-04-09
[cmds] Port ttypong to ELKS

ghaerr:
As a final tty game before the v0.8.0 release, mostly to test the ELKS curses library (which worked for the most part for this port), `ttypong` is now running on ELKS.

The per-frame output routine had to be rewritten, much like `sl` running extremely slowly as discussed in https://github.com/ghaerr/elks/issues/1619#issuecomment-1703364279 until it was rewritten. ELKS just can't process the huge amount of ANSI escape sequences our mini version of curses spits out.

Tested only on QEMU. The default inter-frame delay is 40000 usecs, which can be changed to be half that using `ttypong 20000` for instance, to speed things up. It would be interesting to see how fast this runs on real hardware.

Unfortunately, at the moment `ttypong` is only available on the 2880k disk image, as the 1440k only has ~10k bytes left, and `ttypong` is 11k. Speaking of which, all these ncurses-based games are huge, compared to what ELKS is used to.

Here are the current sizes of the tty games and `fm`, which also uses curses:
```
-rwxr-xr-x  1 greg  staff  19904 Apr  8 17:45 ttyclock
-rwxr-xr-x  1 greg  staff  11360 Apr  8 18:08 ttypong
-rwxr-xr-x  1 greg  staff   6336 Apr  8 15:19 ttytetris
-rwxr-xr-x  1 greg  staff  21632 Apr  8 17:45 fm
```
ttypong

Use the up/down arrow keys or `h` and `j` to move the racket, and `q` to quit.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -171,6 +171,7 @@ tui/cons                        :tui                            :1440k
 tui/ttyinfo                     :tui                            :1440k
 tui/sl                          :tui                    :1200k
 tui/ttyclock                    :tui                    :1200k
+tui/ttypong                     :tui
 tui/ttytetris                   :tui            :360k
 busyelks/busyelks               :busyelks
 inet/httpd/sample_index.html ::var/www/index.html :net  :1200k
--- a/elkscmd/tui/.gitignore
+++ b/elkscmd/tui/.gitignore
@@ -5,3 +5,5 @@ cons
 ttyinfo
 sl
 ttyclock
+ttytetris
+ttypong
--- a/elkscmd/tui/Makefile
+++ b/elkscmd/tui/Makefile
@@ -10,7 +10,7 @@ include $(BASEDIR)/Make.rules
 
 ###############################################################################
 
-PRGS = fm matrix cons ttyinfo sl ttyclock ttytetris
+PRGS = fm matrix cons ttyinfo sl ttyclock ttypong ttytetris
 
 #PRGS_HOST =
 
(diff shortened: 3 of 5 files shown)


toncho11:
Awesome!

But all these tty apps and games should be on the hdd images where there is enough space, right? It is not so clear looking at elkscmd/Applications file.

ghaerr:
Yes, all programs specified in the elkscmd/Applications file, including the somewhat confusing lines without a `:1440k` (or other) size tag are copied to the HD and 2880k floppy images.

I am thinking about creating a new facility where individual application program files can be compressed on the distribution, which would allow more programs on each floppy size. The drawback is that the kernel decompression slows down the application start, and can be pretty slow on older 8088 machines. Another idea would be a permanent decompressor that could be run by a user for more often-used binaries.

toncho11:
I noticed that you control the racket on the left. Interestingly even if the racket is not there the ball still bounces back in some kind of invisible wall? Can you actually lose the game?

ghaerr:
>  Interestingly even if the racket is not there the ball still bounces back in some kind of invisible wall? 

Yes, that seems to be the way the game was written. Take a look, its a pretty small program, perhaps you can modify it to optionally require a racket hit to bounce back or add other enhancements.

> Can you actually lose the game?

No! I suppose this means its more of a pong demo lol.

Did you run this on your 8088 PC, or just QEMU? I'm interested to see how fast it is on real hardware, and what the sleep usecs needs to be.

toncho11:
Not yet on real 8088, only https://copy.sh/v86/.

PR #1867, opened 2024-04-11
[build] Create disk images with compressed executables by default

ghaerr:
This enhancement sets CONFIG_APPS_COMPRESS=y (default ON) in `ibmpc-1440.config` and adds the ability to turn off compression on individual files by setting `:nocomp` in elkscmd/Applications.

In turn, this opened up 61k in the 360k floppy and 377k in the 1440k floppy, which allows for the following applications to be now available in the default distribution:

Applications now on the 360k floppies:
```
decomp    # (decompress executable if desired)
ttyclock
ttypong
ttytetris
matrix
```
Applications now on the 1440k floppies (in addition to above):
```
mail
tinyirc
advent    # Adventure game
nxclock
disasm
nm
md5sum
test_libc   # Libc test
testsym     # --ftrace test
lib/system.sym # kernel symbol table for disassembler
lib/advent.db  # adventure database
```

Compressed executables can be turned off by unsetting CONFIG_APPS_COMPRESS in .config. This change has not yet been added for PC-98, but could be easily tested by adding COFNFIG_APPS_COMPRESS to pc98-1232.config. A `:1200c` tag has not yet been added to elkscmd/Applications, but may want to be to increase the applications built automatically onto the PC-98 1232k floppy. That can be done after initial testing of PC-98.

The elkscmd/Applications file has two new tags `:360c` and `:1440c` which are used to specify adding files to the 360k and 1440k disk images respectively only when CONFIG_APPS_COMPRESS is set (otherwise, the applications won't fit).

Currently, `/bin/sh`, `ps` and `ls` are set `:nocomp`, meaning they are not compressed. This may not be necessary. Users can test using `decomp /bin/meminfo` etc to decompress files to compare floppy loading speed to decompression speed, if desired.

It has been reported that compressed executables may take a bit more noticable time to decompress on older 8088 systems. However, the ability to have many more applications on the default disk images is currently deemed worthwhile. If a particular program should remain uncompressed that can easily be change using `:nocomp`.


toncho11:
It works. It is not ridiculously slower as I was expecting. Good job!
ttyclock -c for centering does not work, seconds are updated every 4 seconds or every 5 seconds
ttypong and ttytetris are slow. ttytetris feels especially slow.  It is unclear if keyboard input is accepted. I managed to stop them only with ctrl + C.

ghaerr:
> It works. It is not ridiculously slower as I was expecting. Good job!

Thanks for testing @toncho11, and good to hear the compression doesn't cause more problems than its worth.

> ttyclock -c for centering does not work

I'm aware of that, unfortunately a lot of work is required to get that working, but its on my list.

> seconds are updated every 4 seconds or every 5 seconds

Interesting. I will try to figure out what that might be, other than perhaps the whole program running slowly like pong and tetris.

> ttypong and ttytetris are slow. ttytetris feels especially slow. It is unclear if keyboard input is accepted. 

Sounds like the real-time timing isn't working properly, since both tetris and pong rely on a timed wait each display cycle.

It would be informative to try running `ttypong` with a reduced time delay: this can be done by passing a number lower than 40000 as its argument. `ttypong 20000` should attempt to run pong twice as fast as normal. It is unclear whether the issue with all these programs is real-time related, or whether the display cycle is simply taking lots longer than the delay. We had the same problem with `sl` and the delay had to be removed altogether. No delay could be tried using `ttypong 1`.

toncho11:
ttyclock is directly updated +4 seconds for example. From 23 seconds it goes to 28 for example. It does not attempt to do 24 I think. The update itself seems to take around 0.5 seconds. I think it is not calculating properly the delay of 1 second between the updates.

toncho11:
ttypong 1 or 20000 does not make a difference. It is still very slow.
In tetris a piece moves down with a speed of 5 seconds per line. The update (redraw) starts after 5 seconds. The redraw takes between 300 ms and 600 ms for example. I tried tapping on "s" in Tetris but it did not make a difference. I have not check the code but if tetris is redrawing everything instead of the piece that is falling maybe that can explain why it is that slow.

toncho11:
Pressing "q" on Tetris takes 12 seconds to exit which gives the impression that it is not working at all until you figure it out.

ghaerr:
> if tetris is redrawing everything instead of the piece that is falling maybe that can explain why it is that slow.

It is redrawing everything, and that is likely one of the reasons all these programs are so slow. I think your test result synopsis is correct, almost all the programs are running very very slow on 8088, proving to be far less usable than I had hoped. The timing routines don't work because the display refresh cycle times always exceed the time delay in every case. 

Certainly rewriting the ncurses library to not display characters that haven't changed would help, but that's lots of work and also lots of code and data. We already have a full ncurses port we did for the nano editor port, but the program ended up being so large that IIRC only a few lines of text can be edited. 

I will try to think what else might able to be done for "super-slow" aka original IBM PC systems (!).

ghaerr:
Thanks for your testing @toncho11. I now realize that with the "slow" 8088 systems, we have a double whammy - not only does the program have to compute each displayed frame, but then the ANSI escape sequences, including the more complicated color sequences, have to be interpreted by the kernel console driver and then converted into actual text display. 

This conversion is a known problem with `sl` developed and debugged last summer on a Russian 8088. Ultimately it was decided that there wasn't any way to improve the speed of the complex console driver sequence interpretations in ELKS without big design changes. The ELKS (Linux) kernel TTY/console subsystem is basically almost too complicated for the speed the processor has given all that's going on.

toncho11:
Ttyclock is almost ok. Only the seconds do not change one after another :)

toncho11:
In this case ttypong and ttyteris should not be on the 360kb image. Your X nano tetris works much better.

PR #1872, opened 2024-04-15
[builds] Revise Applications list for PC-98 1232k image

ghaerr:
Hello @tyama501,

This PR revises elkscmd/Applications to provide both a compressed and non-compressed selection by image size for PC-98 1232k floppies with your requested applications included.

The `beep`, `nxclock`, `nxlandmine`, `nxterm`, and `nxworld` applications are included on the uncompressed 1232k image, and the compressed image includes the same plus all the applications marked `:1440k` in elkscmd/Applications.

To add files to the 1232k image(s), add either `:1232k` or `:1232c` to elkscmd/Applications for the applications you'd like to add to the automatically built installation, and update pc98-1232.config with one of the configurations below. There is still some space left so feel free to add what you think would be useful.

Here are the two configuration files used for the compressed and uncompressed (-nc) build. Please take a look and see what modifications you might like, and then you're welcome to post a PR with the `pc98-1232.config` changed to whichever one you prefer. This should solve the problem of Selecting Applications by Image Size being discussed in #1870.

[config-1232.zip](https://github.com/ghaerr/elks/files/14982020/config-1232.zip)

Thank you!

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -23,8 +23,8 @@
 #   :192k       Additional apps to fit on 192k rom image
 #   :360k       Minimal set of apps to fit on 360k disks
 #   :360c       Extra apps to add to 360k disks if CONFIG_APPS_COMPRESS set
-#   :720k       Set of apps to fit on 720k disks (includes :net)
-#   :1200k      Set of apps to fit on 1200 disks (includes :net)
+#   :720k       Set of apps to fit on 720k disks
+#   :1200k      Set of apps to fit on 1200k disks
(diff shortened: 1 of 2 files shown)


tyama501:
Thank you!

I will take a look and add/remove if necessary.
As you mentioned I would like to leave some space to add Japanese application and make custom image after elks 0.8.0 release.

ghaerr:
>  I would like to leave some space

IIRC there's around ~60k free in the uncompressed 1232k image, but should be quite a bit more in the compressed image. You can also pre-compress any ELKS programs built outside of the repo using `elks-compress` which is built in the elks/tools/bin directory.

PR #1877, opened 2024-04-17
[build] Free up more space on PC-98 1232k compressed image

ghaerr:
Hello @tyama501,

This PR frees up more space on the compressed 1232k image by moving tee, cal, head, tr, xargs, cut, split, sum, yes, mouse, uuencode and uudecode to the `:1200c` image, which is only created for IBM PC 1200k. The nxclock and sercat programs are moved to the `:1440c` image.

This creates a PC-98 1232k compressed image with 275k free. The above method works because now, the 1232k compressed image does not include programs specified for 1200k compressed, only 1232k compressed. It also removes the `:1440k` programs from the 1232k compressed image.

Hopefully this all makes sense, it seems complicated but basically only the `:1200c` and `:1232c` tags are handled mutually exclusively, rather than cumulatively like the others which include all smaller image tags in increasing image sizes.

If this works out, then we should probably update the `pc98-1232k.config` to specify a compressed image by default. Let me know and I will do that. After that, you can add back programs from `:1200c` if you want them.

Thank you!

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -80,7 +80,7 @@ file_utils/ln                   :fileutil           :720k
 file_utils/ls                   :fileutil       :360k           :128k       :nocomp
 file_utils/rm                   :fileutil       :360k
 file_utils/rmdir                :fileutil       :360k
-file_utils/split                :fileutil                :1200k
+file_utils/split                :fileutil                 :1200c    :1440k
 file_utils/sync                 :fileutil       :360k
 file_utils/touch                :fileutil           :720k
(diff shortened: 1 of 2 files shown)


tyama501:
Hello @ghaerr 

Yes! It worked.

I could include the fonts and the application.
The application is not compressed yet, I will try it later.
Thank you.
We can set the compressed image as default.

fd1232c_nxjtxtv

ghaerr:
> We can set the compressed image as default.

Great, thanks for your testing. I will update the pc98-1232.config in a separate PR.

PR #1882, opened 2024-04-18
[pc98_config] Fix typo of capital Y

tyama501:
Hello @ghaerr ,

It seems that the capital Y cannot read in the menuconfig.
CONFIG_APPS_BY_IMAGESZ=Y

One more thing,
if I use the menuconfig
CONFIG_EXEC_MMODEL duplicate. Why this happen?
image

Thank you.


ghaerr:
Oops, so much for my testing! Thank you @tyama501.

> CONFIG_EXEC_MMODEL duplicate

I'm aware of this problem, but haven't figured a great way to fix it yet. It results from the forced definition in elks/fs/config.in:
```
    bool 'Support compressed executables'  CONFIG_EXEC_COMPRESS       y
    if [ "$CONFIG_EXEC_COMPRESS" == "y" ]; then
        define_bool CONFIG_EXEC_MMODEL y
    fi
    bool 'Support medium memory model'     CONFIG_EXEC_MMODEL         y
```
The `define_bool` is required to force CONFIG_EXEC_MMODEL when compressed executables are set, then apparently the `bool` in the next line sets it again.

I removed the second definition in the committed files because I was trying to keep the configuration files as simple as possible, rather than match `make menuconfig` output exactly. Perhaps that should change?

PR #1886, opened 2024-04-19
[console] Use bg color and invert for PC-98 when bg is not zero

tyama501:
I may debug for white, but PR for the clock.
https://github.com/ghaerr/elks/discussions/1817#discussioncomment-9157071

Thank you.


ghaerr:
Thank you @tyama501!

PR #1890, opened 2024-04-21
Fix a variable logic and add a brace pair

floriangit:
The compiler threw a warning over an unused variable. Looking closer the warning made sense, since a variable is defined, but then only used when a preprocessor macro is set. 
I've tested compilation both with STRICT and without (which is the default).

### DIFF
--- a/elks/tools/mfs/getoptx.c
+++ b/elks/tools/mfs/getoptx.c
@@ -63,8 +63,11 @@ getoptX(int argc, char *const *argv, const char *optstring)
    }
    c = argv[optind][sp];	/* option letter */
 #ifdef STRICT
-   osp = sp++;			/* get ready for next letter */
-#else
+   osp = sp;
+#endif
+   sp++;			/* get ready for next letter */
+
+#ifndef STRICT
    oind = optind;		/* save optind for param test */
 #endif
    if (argv[optind][sp] == '\0')/* end of argument */
--- a/elks/tools/mfs/getoptx.c
+++ b/elks/tools/mfs/getoptx.c
@@ -39,9 +39,10 @@ getoptX(int argc, char *const *argv, const char *optstring)
 /* allowed args, e.g. "ab:c" */
 {
    static int sp = 1;		/* position within argument */
-   register int osp;		/* saved `sp' for param test */
 #ifndef STRICT
    register int oind;		/* saved `optind' for param test */
+#else
+   register int osp;		/* saved `sp' for param test */
 #endif
    register int c;		/* option letter */
    register char *cp;		/* -> option in `optstring' */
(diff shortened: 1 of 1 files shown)


ghaerr:
Hi @floriangit,

It appears this change actually causes `mfs` to busy loop forever... which is why the CI check never stops running.
I assume the build process will timeout on Github at some point, I'm not sure how to stop the infinite loop but am currently looking how too.

Looking at the change, I'm pretty sure the problem is that you introduced an `#ifdef STRICT` around a variable declaration that also incremented a global: `osp = sp++`. This causes sp to remain the same and loop forever.

I suggest you put together a test case so that it can be shown that changed code is known working.

Thank you!

ghaerr:
I was able to cancel the workflow! :)

floriangit:
Hi Greg,

thanks for stopping that strange check job, and sorry for that mishap - on a Sunday! :octopus: 
Based on your comment I looked at it again:
> This causes `sp` to remain the same and loop forever.

Indeed. And `sp` is accessed also outside the STRICT branches. This function really is a rabbit hole! 

Please disregard the patch. For now it seems we cannot get the compiler warning go away without unwinding this function more substantially?

ghaerr:
> thanks for stopping that strange check job, and sorry for that mishap - on a Sunday! 

No worries. I am a bit surprised Github allowed the job to run for almost two hours.

> For now it seems we cannot get the compiler warning go away without unwinding this function more substantially?

My (macOS) compiler doesn't produce any warnings so I can't see the problem. If you can post what the warning you're seeing, we could look into this further.

floriangit:
Thanks, I have a follow up patch now (called: 2nd attempt). And assembly shows my 1st attempt was wrong indeed and the 2nd attempt shows same assembly instructions than the original file (without compiler warning).
I'll write it up in more details (including the exact warning) and come back to you.

ghaerr:
Thanks @floriangit!

floriangit:
Thanks for merging. It's a bit odd that your cross-compiler differs? Should be the same version since it's baked into the ELKS package itself regardless of the host architecture? Anyhow, for the record, the compiler warning on Linux was:

```
getoptx.c: In function ‘getoptX’:
getoptx.c:42:17: warning: variable ‘osp’ set but not used [-Wunused-but-set-variable]
   42 |    register int osp;            /* saved `sp' for param test */
      |                 ^~~

```

ghaerr:
>  It's a bit odd that your cross-compiler differs? Should be the same version since it's baked into the ELKS package itself regardless of the host architecture?

The `mfs` tool which this code is used in is compiled by the host compiler, not the included `ia16-elf-gcc` compiler. I'm running macOS clang 11.0.0 which doesn't emit this warning.

PR #1895, opened 2024-04-27
[build] Remove xxd tool from build requirement

ghaerr:
Removes need for external `xxd` tool from build. Discussed in #1894.


ghaerr:
Oops! Turns out `makeboot` requires a binary of the boot block included within it, thus `xxd` is required after all.

PR #1911, opened 2024-05-31
[libc] First pass at compiling most all of C library with OpenWatcom

ghaerr:
Compiles almost all the ELKS C library using the OpenWatcom C compiler for compact model. Large, medium and small mode are easily built by editing libc/watcom.inc.

This first version is mostly for OpenWatcom compatibility testing, getting the Makefiles adjusted, and further application linking testing. Discussed a bit in #1443 and #1786.

To use, the [OpenWatcom C Toolchain](https://github.com/open-watcom/open-watcom-v2) needs to be cloned and built, and then the `WATCOM` and `WATDIR` environment variables need to be setup by sourcing a script like the following with the full path to the OpenWatcom binary directory (WATCOM) and top-level directory (WATDIR):
```
#!/usr/bin/env bash

# Set up Watcom build environment

export WATCOM=/Users/greg/net/open-watcom-v2/rel/bino64 # for macOS
export WATCOM=/Users/greg/net/open-watcom-v2/rel/binl   # for Linux
export WATDIR=/Users/greg/net/open-watcom-v2

add_path () {
    if [[ ":$PATH:" != *":$1:"* ]]; then
        export PATH="$1:$PATH"
    fi
}

add_path "$WATCOM"

echo PATH set to $PATH
```
After that, to build the C library `libc.lib`:
```
cd $TOPDIR            # ELKS top-level directory
cd libc
make -f watcom.mk     # creates libc.lib
```

### DIFF
--- a/elks/include/arch/cdefs.h
+++ b/elks/include/arch/cdefs.h
@@ -26,6 +26,7 @@
 #define restrict        __restrict
 #define printfesque(n)
 #define noinstrument
+#define __attribute__(n)
 /* force __cdecl calling convention and no register saves in main() arc/argv */
 #pragma aux main "*" modify [ bx cx dx si di ]
 #endif
--- a/libc/.gitignore
+++ b/libc/.gitignore
@@ -1,3 +1,5 @@
 crt0.o
 libc.a
 build-ml/
+*.obj
+*.lib
--- a/libc/Makefile
+++ b/libc/Makefile
@@ -239,4 +239,4 @@ endif
 clean:
 	for DIR in $(SUBDIRS); do $(MAKE) -C $$DIR clean || exit 1; done
 	rm -rf build-ml *.o libc.a
-	$(MAKE) -f Watcom clean
+	$(MAKE) -f watcom.mk clean
--- a/libc/Makefile
+++ b/libc/Makefile
@@ -1,4 +1,4 @@
-# Makefile of ELKS C Library
+# ia16-elf-gcc Makefile for ELKS C Library
 
 ifndef TOPDIR
 $(error TOPDIR is not defined)
@@ -6,7 +6,7 @@ endif
 
 include $(TOPDIR)/Make.defs
 
-include Makefile.inc
+include ia16.inc
 
 # Defines
 
@@ -239,3 +239,4 @@ endif
 clean:
 	for DIR in $(SUBDIRS); do $(MAKE) -C $$DIR clean || exit 1; done
 	rm -rf build-ml *.o libc.a
+	$(MAKE) -f Watcom clean
--- a/libc/Watcom
+++ b/libc/Watcom
@@ -1,0 +1,36 @@
+# OpenWatcom Makefile for ELKS C Library 
+
+include watcom.inc
+
+# \
+	asm	    \
+	gcc	    \
+	math	\
+	crt0.S	\
+
+SUBDIRS =	\
+	ctype	\
+	error	\
+	getent	\
+	malloc	\
+	misc	\
+	net	    \
+	regex	\
+	stdio	\
+	string	\
+	termcap	\
+	termios	\
+	time	\
+	# end of list
+
+.PHONY: all
+all:
+	$(MAKE) -C system -f out.mk COMPILER=watcom LIB=out.lib
+	for DIR in $(SUBDIRS); do $(MAKE) -C $$DIR COMPILER=watcom LIB=out.lib || exit 1; done
+	wlib -n -b -fo libc.lib */*.lib
+
+.PHONY: clean
+clean:
+	rm -f system/*.obj system/*.lib
+	for DIR in $(SUBDIRS); do rm -f $$DIR/*.obj $$DIR/*.lib || exit 1; done
+	rm -f libc.lib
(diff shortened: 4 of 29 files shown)


tyama501:
Hello @ghaerr ,

Fantastic. Does this also need gcc-ia16?
Since I only have OpenWatcom in Ubuntu on qemu and I don't have gcc-ia16 on that yet.

Thank you.

ghaerr:
Hello @tyama501,

> Does this also need gcc-ia16?

No. Except that it does build the ELKS C library within the $TOPDIR/libc directory of the ELKS repo, so the repo must be cloned at present. In addition, the tool `os2toelks` will be required, which is normally built during the ELKS tree tools build. This can be done manually however using:
```
cd $TOPDIR/elks/tools/objtools
cc -W -Wall -O3 -o os2toelks os2toelks.c
```
and then copying `os2toelks` to somewhere in your path.

> I only have OpenWatcom in Ubuntu on qemu and I don't have gcc-ia16 on that yet.

The idea will be that one can compile ELKS programs without requiring `ia16-elf-gcc` at all. Full replacement! Some other tools like the above os2toelks may continue to be necessary though. Building the C library in the ELKS repo will produce a $TOPDIR/libc/libc.lib file that may also have to be copied out. I hope to build some compiler scripts that will automatically run the Watcom C compiler/linker with the correct option along with finding the libc.lib C library so that this can be automated. When that's done I'll have it build the required extra tools (like os2toelks) as well.

Thank you!

toncho11:
Would having this compiler help a bit with porting Doom8088 :) ? Maybe the memory requirements will now be easier: medium, large model, far pointers, etc?

ghaerr:
Yes @toncho11. The OpenWatcom C compiler is an incredible piece of software technology. It has everything we could need as far as supporting small, medium, compact, large and huge models, near/far pointers, fast/specialized register-calling sequences, and the likely ability to run on ELKS itself at some point. It comes with complete source to all tools and its very complete runtime libraries for DOS, OS/2, and 32-bit Linux. I'm quite impressed with it.

That said, there's still quite a bit of work getting the large data model working, and the ASM format is Intel rather than AT&T (ia16-elf-as) compatible. I'm currently in a deep dive trying to get our C library compiled and working in large and compact models. Things are almost working, but the ELKS a.out format isn't suitable for > 64K data or 128K code and will have to be enhanced in order to fully support large data/code. There's also some big issues with floating point, which is that soft floating point emulation will be hard to support. The math library is currently being compiled assuming an 8087 coprocessor. For now, this only affects the `basic` interpreter and floating point number display.

tyama501:
Hello @ghaerr ,

I needed to source the following script that is included in the linux build watcom,
but I could build libc.
(install dir is binl64 for linux 64 and INCLUDE is also needed)

image

Cool!
image

Thank you.

tyama501:
Oh, should not the header in /usr/bin/watcom/lh be used?
It says stddef.h is missing if I don't set that INCLUDE.

toncho11:
> Yes @toncho11. The OpenWatcom C compiler is an incredible piece of software technology. It has everything we could need as far as supporting small, medium, compact, large and huge models, near/far pointers, fast/specialized register-calling sequences, and the likely ability to run on ELKS itself at some point. It comes with complete source to all tools and its very complete runtime libraries for DOS, OS/2, and 32-bit Linux. I'm quite impressed with it.
> 
> That said, there's still quite a bit of work getting the large data model working, and the ASM format is Intel rather than AT&T (ia16-elf-as) compatible. I'm currently in a deep dive trying to get our C library compiled and working in large and compact models. Things are almost working, but the ELKS a.out format isn't suitable for > 64K data or 128K code and will have to be enhanced in order to fully support large data/code. There's also some big issues with floating point, which is that soft floating point emulation will be hard to support. The math library is currently being compiled assuming an 8087 coprocessor. For now, this only affects the `basic` interpreter and floating point number display.

Yes at the end of the https://github.com/FrenkelS/Doom8088 readme it says:

`It's also possible to build a 16-bit version with Watcom: Run setenvwc.bat followed by compwc16.bat.`

toncho11:
Excellent work @ghaerr !!!

ghaerr:
Hello @tyama501,

> It says stddef.h is missing if I don't set that INCLUDE.

Sorry, I'm fixing that right now - long story short it is getting complicated mixing header files between IA16 and Watcom compilers. I should have a PR posted shortly. The WATDIR= environment variable does need to be set though to the base of the OpenWatcom repo (not the bin installation locations).

> Oh, should not the header in /usr/bin/watcom/lh be used?

No, the "lh" headers are for Linux, but they can't be properly used for ELKS because all the Watcom headers for Linux assume 32-bit integer (and other) sizes and so are incorrect. We have to use the DOS headers but the application programs will always use the ELKS headers, with a few exceptions noted below. The INCLUDE variable is set in libc/watcom.inc should not be /usr/bin/watcom/lh, but rather $WATDIR/bld/hdr/dos/h.

That said, some applications will need "standard" headers that are not provided by ELKS C library. These include stddefs.h, stdarg.h, and a few others. The stdint.h header is usable but is redirected by the ELKS stdint.h. I will clarify more on this after I have it all sorted out.

We also cannot use any of the system call wrappers OpenWatcom has already built for DOS, OS/2 or Linux, as none of these will work correctly for ELKS. I'm also working on that now.

Thank you!

PR #1917, opened 2024-06-09
[owc] Add fmemalloc syscall

tyama501:
PR for adding fmemalloc syscall for OpenWatcom regarding
https://github.com/ghaerr/elks/pull/1912#issuecomment-2150346984

Let me know if something is wrong.

Thank you!

### DIFF
--- a/libc/include/malloc.h
+++ b/libc/include/malloc.h
@@ -29,6 +29,9 @@ extern void *(*__alloca_alloc)(size_t);
 #define malloc(x) ((*__alloca_alloc)(x))
 #endif
 
+/* request paras from main memory, returns segment */
+int _fmemalloc(int paras, unsigned short *pseg);
+
 /* alloc from main memory */
 void __far *fmemalloc(unsigned long size);
 
--- a/libc/include/malloc.h
+++ b/libc/include/malloc.h
@@ -29,10 +29,8 @@ extern void *(*__alloca_alloc)(size_t);
 #define malloc(x) ((*__alloca_alloc)(x))
 #endif
 
-/* request paras from main memory, returns segment */
-int _fmemalloc(int paras, unsigned short *pseg);
-
 /* alloc from main memory */
 void __far *fmemalloc(unsigned long size);
+int _fmemalloc(int paras, seg_t *pseg);
 
 #endif
--- a/libc/watcom/syscall/fmemalloc.c
+++ b/libc/watcom/syscall/fmemalloc.c
@@ -1,41 +1,14 @@
 /****************************************************************************
 *
-*                            Open Watcom Project
-*
-* Copyright (c) 2002-2019 The Open Watcom Contributors. All Rights Reserved.
-*    Portions Copyright (c) 1983-2002 Sybase, Inc. All Rights Reserved.
-*
-*  ========================================================================
-*
-*    This file contains Original Code and/or Modifications of Original
-*    Code as defined in and that are subject to the Sybase Open Watcom
-*    Public License version 1.0 (the 'License'). You may not use this file
-*    except in compliance with the License. BY USING THIS FILE YOU AGREE TO
-*    ALL TERMS AND CONDITIONS OF THE LICENSE. A copy of the License is
-*    provided with the Original Code and Modifications, and is also
-*    available at www.sybase.com/developer/opensource.
-*
-*    The Original Code and all software distributed under the License are
-*    distributed on an 'AS IS' basis, WITHOUT WARRANTY OF ANY KIND, EITHER
-*    EXPRESS OR IMPLIED, AND SYBASE AND ALL CONTRIBUTORS HEREBY DISCLAIM
-*    ALL SUCH WARRANTIES, INCLUDING WITHOUT LIMITATION, ANY WARRANTIES OF
-*    MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, QUIET ENJOYMENT OR
-*    NON-INFRINGEMENT. Please see the License for the specific language
-*    governing rights and limitations under the License.
-*
-*  ========================================================================
-*
 * Description:  ELKS _fmemalloc() system call.
 *
 ****************************************************************************/
 
-
 #include <malloc.h>
 #include "watcom/syselks.h"
 
-int _fmemalloc( int __paras, unsigned short *__pseg )
+int _fmemalloc( int __paras, seg_t *__pseg )
 {
-    sys_set_ptr_seg(__pseg);
     sys_setseg(__pseg);
     syscall_res res = sys_call2( SYS_fmemalloc, __paras, __pseg);
     __syscall_return( int, res );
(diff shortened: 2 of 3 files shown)


ghaerr:
Hello @tyama501,

This looks great, nice job!

However, I think there is an error, a bit hard to explain, having to do with my suggestion to use `_sbrk` as an example for `_fmemalloc`. The issue is that `_sbrk` returns a pointer through its second argument (from the kernel), while `_fmemalloc` only returns a segment (`seg_t`, same as `unsigned short`). I will explain directly in the code review area.

Thank you!

libc/include/malloc.h
```diff
@@ -29,6 +29,9 @@ extern void *(*__alloca_alloc)(size_t);
 #define malloc(x) ((*__alloca_alloc)(x))
 #endif
 
+/* request paras from main memory, returns segment */
+int _fmemalloc(int paras, unsigned short *pseg);
```
ghaerr: Lets use `seg_t *` rather than `unsigned short *` to better indicate that the system call returns a segment number.

libc/watcom/syscall/fmemalloc.c
```diff
@@ -0,0 +1,42 @@
...
+*    The Original Code and all software distributed under the License are
+*    distributed on an 'AS IS' basis, WITHOUT WARRANTY OF ANY KIND, EITHER
+*    EXPRESS OR IMPLIED, AND SYBASE AND ALL CONTRIBUTORS HEREBY DISCLAIM
+*    ALL SUCH WARRANTIES, INCLUDING WITHOUT LIMITATION, ANY WARRANTIES OF
+*    MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, QUIET ENJOYMENT OR
+*    NON-INFRINGEMENT. Please see the License for the specific language
+*    governing rights and limitations under the License.
+*
+*  ========================================================================
+*
```
ghaerr: Since this code is actually written by yourself from scratch, and did not come from Open Watcom as a previously-existing system call, lines 2-27 can be deleted.

libc/watcom/syscall/fmemalloc.c
```diff
@@ -0,0 +1,42 @@
...
+****************************************************************************/
+
+
+#include <malloc.h>
+#include "watcom/syselks.h"
+
+int _fmemalloc( int __paras, unsigned short *__pseg )
+{
+    sys_set_ptr_seg(__pseg);
```
ghaerr: There is a bug here calling `sys_set_ptr_seg`, and the line should be deleted. The reason is a bit complicated: in the `_sbrk` code, the argument is `void **__pnewbrk`, which is a pointer to a pointer, unlike here, where we only have a pointer to an unsigned short.

So there are two pointer indirections. The first indirection is used by the kernel, and the `sys_setseg` call sets DS to the segment to be used, and the offset is passed in `(unsigned)__pnewbrk` (see sbrk.c). Thus the kernel gets the DS value from `current->t_regs.ds` and the offset from the (always near) `__pnewbrk` in the kernel.

The problem is that in large model, we actually need a system call that returns a pointer to another pointer. That's the `void *` that the first `*` is pointing to. Now, in large model, that pointer is 4 bytes long, and the offset has been set by the kernel, but the segment portion the kernel knows nothing about. So, the definition of `sys_set_ptr_seg` is:
```
#if defined(__COMPACT__) || defined(__LARGE__)
#define sys_set_ptr_seg(ptr)    \
    (*((unsigned *)(ptr) + 1) = (unsigned long)(ptr) >> 16)
#else
#define sys_set_ptr_seg(ptr)    /* DS already set in return pointer in small/medium */
#endif
```
This bit of magic code sets the segment portion of the second pointer to the segment being passed to the system call (which will almost always be DS or SS).

In our case for `_fmemalloc`, the system call only returns a segment number, not a pointer. Thus, this call writes two bytes of segment in the memory address 2 bytes past the address of the variable the programmer passed to the system call wrapper!

In the small/medium case, pointers are only 2 bytes, so the `sys_set_ptr_seg` does nothing. In all cases, this should not be performed for `_fmemalloc`.

I hope this makes sense!!

libc/watcom/syscall/fmemalloc.c
```diff
@@ -0,0 +1,42 @@
...
+    sys_setseg(__pseg);
+    syscall_res res = sys_call2( SYS_fmemalloc, __paras, __pseg);
```
ghaerr, line 13: I thought the `wdis` code looked correct, but we should write `(unsigned)__pseg` to ensure that only the offset portion of `__pseg` is loaded into registers by the `sys_call2` macro. See sbrk.c for example.

libc/include/malloc.h
tyama501: It is also used for gcc-ia16. Is it ok to modify without ifdef or something?

ghaerr: Yes, the `seg_t` type should be picked up from types.h. This renaming is all part of an ongoing cleanup for ELKS.

libc/watcom/syscall/fmemalloc.c
tyama501: Hmm, difficult:)
I will try to understand, and the fix the codes but is this mean the small model has not been affected?
If it is so, the root cause of the program not working may be different.

Thank you.

ghaerr: > I will try to understand, and the fix the codes but is this mean the small model has not been affected?

True, including the `sys_set_ptr_seg` will have no effect in small model, and thus the root cause could be something else. We are dealing with lots of new variables. That is why I suggest possibly trying to test `fmemalloc` directly in another test program. Nonetheless, `sys_set_ptr_seg` should be removed for this system call.

tyama501:
Hello @gharr,

I tested the fmemalloc with the following test code with the small model and the large model.
256 x 10Bytes allocation, write, and read were OK.

```
#include <stdio.h>
#include <malloc.h>

int main(){

  unsigned char __far *data[10];
  unsigned char __far *test_data;

  for (int i = 0; i < 10; i++) {
    printf("--- test %d ---\n", i);
    data[i] = (unsigned char __far *)fmemalloc(256);
    // write
    test_data = data[i];
    for (int j = 0; j < 256; j++) {
      *test_data = (unsigned char)j;
      test_data++;
    }
    // read
    test_data = data[i];
    for (int j = 0; j < 256; j++) {
      if (*test_data == (unsigned char)j) {
        printf("OK %d\n", *test_data);
      } else {
        printf("NG %d\n", *test_data);
      }
      test_data++;
    }
    getchar();
  }

  return 0;
}
```

Thank you.

ghaerr:
Thank you! Since `fmemalloc` now appears to be working, let me know what you think the next steps might want to be to figure out the problem discussed in https://github.com/ghaerr/elks/pull/1912#issuecomment-2156352406. Perhaps just some simpler version of the program you're trying to port (using `fopen` etc) to more quickly determine why the program is hanging. I can then try to duplicate on my end, without all the Japanese fonts, etc?

tyama501:
Sorry. A little bit late but I made another PR.

I will try it with smaller file to find whether it works or not.
Thank you!

PR #1941, opened 2024-07-27
[watcom] correct some C compiler, Librarian and Linker options

jmalak:
- remove WATDIR variable as useless and correct use of WATCOM variable which must contain path to OpenWatcom installation not to binaries, it is used by OpenWatcom tools to locate some configuration files etc.

- use neutral target for C compiler, option -bnone instead of target specific -bos2 it ensures that compiler don't define any target specific macros or use any other target specific setup, it uses header files with guards for standard DOS definition (more neutral) not for OS/2

- use -bos2 target for linker only, it define output format as OS/2 New Executable, but for ELKS OpenWatcom use should be better to use wcc and wlink instead of owcc driver, wlinker can specify output format without need to specify OS as with owcc

- enable wlinker message 1008 because disabling it mask unavailable libraries, the issue is correctly resolved by wlink option nodefaultlibs which ignore OpenWatcom standard libraries which can be defined in linking modules, it is suppresed by -fnostdlib for C compiler, now -fnostdlib can be removed but it is not necessary

- suppress use of INCLUDE environment variable by C compiler option -x, it ensure that no other path then command line for headers are used

- suppress all headers printing by removing -v option and adding -q option to wlib

### DIFF
--- a/elks/tools/objtools/ewcc
+++ b/elks/tools/objtools/ewcc
@@ -18,29 +18,26 @@ fi
 
 if [ -z "$WATCOM" ]
   then
-    echo "WATCOM= environment variable not set to compiler path"
-    exit
-fi
-
-if [ -z "$WATDIR" ]
-  then
-    echo "WATDIR= environment variable not set to Watcom source directory"
+    echo "WATCOM= environment variable not set"
     exit
 fi
 
 ELKSLIBCINCLUDE=$TOPDIR/libc/include
 ELKSINCLUDE1=$TOPDIR/include
 ELKSINCLUDE2=$TOPDIR/elks/include
-WATCINCLUDE=$WATDIR/bld/hdr/dos/h
+WATCINCLUDE=$WATCOM/h
 
 # owcc options:
+# -bnone                    # no target specific setup
 # -mcmodel={s,m,c,l}        # memory model
 # -march=i86                # 8086 codegen
-# -bos2                     # produce OS/2 exe
 # -std=c99                  # -Wc,-za99
 # -Wc,-zev                  # enable void arithmetic
 # -Wc,-wcd=N                # disable warning N
 # -Wc,-fpi87                # inline 8087 fp
+# -Wc,-x                    # ignore INCLUDE environment variable
+# -fno-stack-check          # don't generate stack check code
+# -fnostdlib                # don't refere standard libraries
 # unused:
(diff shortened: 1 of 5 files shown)


ghaerr:
Hello @jmalak,

Nice to see you here! Thank you very much for reviewing and enhancing the OpenWatcom setup and compilation scripts. Your comments about the new options added are helpful. I have tested the changes and things seem to work very well.

I had wondered about the `-bos2` option for the C compiler, does it just define `__OS2__` for various library code, or are there internal effects related to code generation? The source files this might most affect are some we brought over from OWC that are needed but it was not desired to try to recompile portions of the OpenWatcom C library ASM source (e.g. setjmp, etc). These are located in libc/watcom/asm.

I am thinking we shall set the `-fsigned-char` option as you've hinted in ewcc, so that code is more compatible with gcc.

I originally liked having `owcc -v` as it showed the `wcc` or `wlink` command line for clarity, but it is fine to remove it, now that most research regarding options is likely completed.

 @tyama501, this should fix the INCLUDES=/WATCINCLUDE= change you mentioned [here](https://github.com/ghaerr/elks/issues/1443#issuecomment-2239531123).

Thank you!

jmalak:
Hi @ghaerr ,

I have to commend you for the OpenWatcom configuration, it's nice only minimal changes were necessary. I already checked many projects using OpenWatcom but there it was copied from somewhere and it worked accordingly. Your project shows care and effort to get it right.
It is important to realize that your project is a new operating system, even if it is based on the Linux kernel, and therefore it is not possible to compile the sources either for Linux or for OS/2, even if you use the OS/2 NE format for executable files. I only found one clash with OpenWatcom for MAX_PATH macro, but it is not used frequently only one time in sources and it is overridden in your header correctly. We have several uses of OpenWatcom for embedded systems with their own operating system, e.g. RDOS is also in the official distribution, it is based on WIN32 API and PE executable format.

In the future, it might be better to refactor ewcc to use wcc directly, this allows you to use a lot of options directly without being restricted by the owcc driver, which uses a configuration file with some preferences but only for regular Operating systems to simplify use for user. The same goes for ewlink, it should also use wlink directly instead of owcc, it would be more useful here because it would be possible to easily change the format of the created files in case of extension or change, which owcc simply does not support. I will edit the Wiki to fix some things as we talked and reflect the changes I made. I apologize for my English, so if something is wrong, please correct it.

PR #1943, opened 2024-07-28
Create missing STACK segment in LIBC startup code

jmalak:
STACK segment must be defined in startup code for applications.
Size can be setup by linker, but must exists.
Remove disabling of message 1014 by wlink

### DIFF
--- a/elks/tools/objtools/ewlink
+++ b/elks/tools/objtools/ewlink
@@ -24,17 +24,13 @@ fi
 
 ELKSLIBC=$TOPDIR/libc
 
-# Warning 1014: stack segment not found
-
 LDFLAGS="\
     -bos2                           \
     -s                              \
     -fnostdlib                      \
-    -Wl,disable -Wl,1014            \
     -Wl,option -Wl,start=_start_    \
     -Wl,option -Wl,dosseg           \
     -Wl,option -Wl,nodefaultlibs    \
-    -Wl,option -Wl,stack=0x1000     \
     -Wl,option -Wl,heapsize=0x1000  \
     "
 
--- a/libc/watcom/syscall/crt0.c
+++ b/libc/watcom/syscall/crt0.c
@@ -10,6 +10,16 @@
 #include <unistd.h>
 #include <errno.h>
 
+/*
+ * create STACK segment which is required for applications
+ * size is setup by linker to default size 4096 bytes
+ * or to explicitly defined size by linker directive "OPTION STACK=..."
+ */
+#pragma data_seg("STACK","STACK")
+
+/* return back to default data segment */
+#pragma data_seg()
+
 /* Watcom extern code refs are sym_, extern data refs are _sym */
 
 /* external references created by Watcom C compilation - unused */


ghaerr:
Hello @jmalak,

I am wondering why OWC requires a STACK segment to be defined for applications, things seem to work fine now? This was purposely removed as I wanted the segment definitions to be close as possible to ELKS, where ELKS has no defined stack segment (but does define CODE, DATA and BSS). I figure adding back a STACK segment for OWC compatibility only would be potentially confusing to ELKS developers, if the only reason for this change is to eliminate Warning 1014: Stack segment not found.

In addition, ELKS treats stack size slightly differently than OS/2, DOS or apparently OWC: a stack size of zero means use a default stack size, but this is allocated and the amount determined by the OS, not crt0 startup code. In the present case, I would prefer to keep an explicit stack size in the linker command line (just as heap size is also given, but also has ELKS OS defaults) so that it is very apparent what the stack size is. This has been done in ELKS because we have spent quite a bit of time making application memory size as small as possible and are using the ia16-elf-gcc `-maout-stack=` option to override default. I plan on making further changes with regards to (default) stack requirements as OWC finds itself used more in ELKS, and I work towards a way to get more ELKS applications (compiled automatically without Makefile changes) using OWC.

Finally, if a STACK (or any other) segment is declared, I think it should be in watcom/asm/segments.asm, where all the other segments are declared, rather than crt0.s to avoid confusion. I notice that you have opened #1944 where segments.asm is heavily modified, I will comment on those changes over there.

Thank you!

jmalak:
I understand, I removed default definition for stack size because default stack size allocated by OW linker is exactly same as default value you used on command line. If you have your own stack allocated by ELKS then the stack allocated by wlink should be removed, but OpenWatcom compiler expect stack to be part of DGROUP (near data) by default. In case that ELKS define far stack then it should be compiled with option which expect far stack and stack check routine should be changed to work properly with ELKS stack. I defined stack in crt0 module because it is always linked to program any other modul does not need to be linked. Behaviour should be exactly same as before. I will return back the default stack size to be define on linker command. Please could you give me reference to some documentation how ELKS allocate stack for application to properly use stack by OpenWatcom tools.

It looks like you use far stack then it is only issue with missing -zu compiler option.
I checked libc stack check routine in crt0 and it is OK to work with far stack.
I will wait for your confirmation if my view is correct now and then only adding -zu compiler option should solve the issue.
I will chech rules for OS/2 stack allocation in OW linker and documentation may be there is some issue that it report incorrectly missing stack because OW use for application near stack in DGROUP by default and stack segment always exists. Only DLL use far stack that near stack doesn't exists, but there is no check for stack.

ghaerr:
> In case that ELKS define far stack then it should be compiled with option which expect far stack 

ELKS can't easily use a far stack as the system expects DS == SS in many cases. This could possibly change using OWC but is a limitation using ia16-elf-gcc. Nonetheless, this would be a major change and is not planned for now. Adding DLLs would be very interesting as I can see now how OS/2 and OWC supports them, but they are also not planned and quite complicated.

Also, currently there is no application stack check that occurs, except during kernel entry and for stack size calculation that I wrote for OWC `alloca` to work.

> It looks like you use far stack then it is only issue with missing -zu compiler option.

So we won't be using a far stack. What is -zu compiler option?

> I will chech rules for OS/2 stack allocation in OW linker and documentation may be there is some issue that it report incorrectly missing stack 

My design is using the NE executable format which includes a (non-zero) stack size value which is used by the kernel for stack allocation, but NE format does not require a stack segment, to my knowledge. As you mentioned, we are using OS/2 executable format but not running OS/2, so assumptions don't have to be made and should kept separate. 

DOS linkage seems to require a STACK segment and OS/2 may, but we're just using OS/2 NE format, not OS/2 compatibility. The OWC-produced OBJ files emit named segment references using the SEGDEF16 and other record types, but to my knowledge, SEGDEF16 references to STACK are never emitted. This, combined with the need to order segments discussed elsewhere requires that watcom/asm/segments.asm to specify the DGROUP ordering, but without STACK since it is unused and not wanted.

Note that DGROUP references by OWC in OBJ files don't require knowledge of DGROUP ordering, this is delayed until link time, after having processed a DGROUP segment ordering ASM directive. As previously mentioned, because various tools are also involved inspecting and disassembling OBJ and EXE files, unused segments were removed for both less confusion and a means of error-checking during systems development with a new compiler. That said, certain OWC segments are kept, as I plan to continue to implement some of OWC's very nice runtime and compiler features into ELKS over time. (I am learning a lot reading OWC runtime, super resource!)

In conclusion, I think it best the way it is, with no STACK segment, forced stack size, and disabled 1014 warning message.

> in OW linker and documentation may be there is some issue that it report incorrectly missing stack because OW use for application near stack in DGROUP by default and stack segment always exists.

I would suspect that OW linker emits warning by default since many DOS programs were written in ASM and a stack segment may have been forgotten. Nice to have a warning in that case.

jmalak:
I am surprised by the rejection of the stack segment, because the stack, if it is not supported by the OS, is quite important, because the application must watch for its overflow, and it must be taken into account that it must be set so that it does not overflow. I understand that there is not much talk about a stack segment in Linux, but in a 16-bit segmented architecture it is quite dangerous to hide such things.

image

On the picture above is how OW handle all relevant segments used by C compiler for DOSSEG directive and OS/2 NE executable format. The name of segments can be changed by appropriate compiler options if you need different one.
What is in NE header as stack size and heap size is related to appropriate near segments and calculation of dgroup.
if DGROUP overflow it generates warning and shortens the length of the near heap to overall size be 64 kBytes.
if it is not enough then it is error.
Please check with ELKS memory layout if it is OK or need some changes.

ghaerr:
> I am surprised by the rejection of the stack segment

Thanks for the OW image, now I understand our potential confusion. In OW, this shows the stack is a fixed size and is directly after BSS, with the HEAP after that. In ELKS, the stack is "allocated" fixed size but isn't really, as in ELKS the HEAP is directly after BSS, and the STACK is allocated after the HEAP, on top. Thus, the stack can potentially grow down into unallocated heap area. In summary, the OW CRTL and ELKS stack and heap are in different locations.

Is it true that the GROUP directive specifies the order of segments? Or is it that the linker order segments in the order of their definition outside of a GROUP directive? 

I now see that in the original GROUP directive for OW CRTL, STACK followed BSS but there is not HEAP segment declared, possibly because it is dynamically allocated. Thus I had thought that STACK was at the end of the DGROUP, whereas your picture now shows it to be directly after BSS. So we can say ELKS has a similar issue but in reverse, where the HEAP is directly after BSS (but undeclared as dynamically allocated) and the STACK is after that (undeclared as dynamically allocated).

Thus it would seem that the only reason to have a segment declared in a GROUP directive is if that segment is in a known location or of a fixed size. Neither are the case for ELKS, but STACK is a fixed size and always directly after BSS in OW.

```
ELKS process layout

(high address)
STACK
 |
 V
...
 ^
 |
HEAP
BSS
DATA
CODE
(low address)
```

jmalak:
> > In case that ELKS define far stack then it should be compiled with option which expect far stack
> 
> ELKS can't easily use a far stack as the system expects DS == SS in many cases. This could possibly change using OWC but is a limitation using ia16-elf-gcc. Nonetheless, this would be a major change and is not planned for now. Adding DLLs would be very interesting as I can see now how OS/2 and OWC supports them, but they are also not planned and quite complicated.
> 
> Also, currently there is no application stack check that occurs, except during kernel entry and for stack size calculation that I wrote for OWC `alloca` to work.
> 

It is clear if you use only near stack

> > It looks like you use far stack then it is only issue with missing -zu compiler option.
> 
> So we won't be using a far stack. What is -zu compiler option?
> 
-zu option is about far stack then compiler assume SS != DS and SS is not use as DGROUP in code

> > I will chech rules for OS/2 stack allocation in OW linker and documentation may be there is some issue that it report incorrectly missing stack
> 
> My design is using the NE executable format which includes a (non-zero) stack size value which is used by the kernel for stack allocation, but NE format does not require a stack segment, to my knowledge. As you mentioned, we are using OS/2 executable format but not running OS/2, so assumptions don't have to be made and should kept separate.
> 

It is clear that you use only executable format, it was my question before how looks like ELKS memory layout to better understand potential problems.

> DOS linkage seems to require a STACK segment and OS/2 may, but we're just using OS/2 NE format, not OS/2 compatibility. The OWC-produced OBJ files emit named segment references using the SEGDEF16 and other record types, but to my knowledge, SEGDEF16 references to STACK are never emitted. This, combined with the need to order segments discussed elsewhere requires that watcom/asm/segments.asm to specify the DGROUP ordering, but without STACK since it is unused and not wanted.
> 

It is not about "required" it must exists in any application if it is not handled by OS and linker must included it in linking process because it is part of DGROUP. It is not a number in some header, it represent a memory block which must be taken into acount for linking process.

> Note that DGROUP references by OWC in OBJ files don't require knowledge of DGROUP ordering, this is delayed until link time, after having processed a DGROUP segment ordering ASM directive. As previously mentioned, because various tools are also involved inspecting and disassembling OBJ and EXE files, unused segments were removed for both less confusion and a means of error-checking during systems development with a new compiler. That said, certain OWC segments are kept, as I plan to continue to implement some of OWC's very nice runtime and compiler features into ELKS over time. (I am learning a lot reading OWC runtime, super resource!)
> 
> In conclusion, I think it best the way it is, with no STACK segment, forced stack size, and disabled 1014 warning message.
> 

I don't think this is the right way because ELKS doesn't handle the stack, it just calculates the size of the memory block and allocates it for all the near data including the stack, but the linker has to allocate space on the output virtual address space depending on the stack location and the size that is derived from linking rules (in your case DOSSEG). It looks like the i86-gcc linker is specialized for your format. However, we cannot make any simplifications because we are working for different memory layouts, so the STACK segment represents the size and location in the linker's output virtual address space. In the case of a stack-managed operating system, only the stack size is passed in the header and STACK segment is not included to output virtual address space because it is not task for linker.

> > in OW linker and documentation may be there is some issue that it report incorrectly missing stack because OW use for application near stack in DGROUP by default and stack segment always exists.
> 
> I would suspect that OW linker emits warning by default since many DOS programs were written in ASM and a stack segment may have been forgotten. Nice to have a warning in that case.

No I disagree, it has nothing to do with assembler sources. It is related to my notes on the begining.

jmalak:
> > I am surprised by the rejection of the stack segment
> 
> Thanks for the OW image, now I understand our potential confusion. In OW, this shows the stack is a fixed size and is directly after BSS, with the HEAP after that. In ELKS, the stack is "allocated" fixed size but isn't really, as in ELKS the HEAP is directly after BSS, and the STACK is allocated after the HEAP, on top. Thus, the stack can potentially grow down into unallocated heap area. In summary, the OW CRTL and ELKS stack and heap are in different locations.
> 
> Is it true that the GROUP directive specifies the order of segments? Or is it that the linker order segments in the order of their definition outside of a GROUP directive?
> 

No, GROUP directive has nothing to do with segment ordering. It only define which segments can be addressed by appropriate GROUP. Of cause you can not select segments to GROUP that address of last segment (by address) overflow 64 kB from GROUP begining, standard limitation for 8086 segment.
You can have multiple GROUPS and they can overlap, but each segment can only be in one GROUP.

> I now see that in the original GROUP directive for OW CRTL, STACK followed BSS but there is not HEAP segment declared, possibly because it is dynamically allocated. Thus I had thought that STACK was at the end of the DGROUP, whereas your picture now shows it to be directly after BSS. So we can say ELKS has a similar issue but in reverse, where the HEAP is directly after BSS (but undeclared as dynamically allocated) and the STACK is after that (undeclared as dynamically allocated).
> 
> Thus it would seem that the only reason to have a segment declared in a GROUP directive is if that segment is in a known location or of a fixed size. Neither are the case for ELKS, but STACK is a fixed size and always directly after BSS in OW.
> 
> ```
> ELKS process layout
> 
> (high address)
> STACK
>  |
>  V
> ...
>  ^
>  |
> HEAP
> BSS
> DATA
> CODE
> (low address)
> ```

The Order of segments in OS/2 format for ELKS can be change because OW linker can order segments by custom rules.
In your case it is not necessary, wlink calculate full dgroup size and shortens near heap size that stack size is unchanged.
It has only impact to adress of near heap segment and stack segment how to see it wlink. But it is not expected to have any symbol in stack or heap segment that it should be OK. If you expect ELKS format and calculate address of stack and heap by startup code then no problem because both segments are continuous on the end of memory block.

jmalak:
With regards the ELKS memory layout  and your allocation logic you need to rework OW stack checking a little bit.
OW uses variable stacksize which hold size of stack from header that you need to update it to follow boundary between heap and stack by some logic if near heap is allocated over original stack size.
I am not sure if it is good practice to do it this way. You should be able to guarantie stack size which is more important then near heap.

ghaerr:
>> In conclusion, I think it best the way it is, with no STACK segment, forced stack size, and disabled 1014 warning message.

> I don't think this is the right way because ELKS doesn't handle the stack, it just calculates the size of the memory block and allocates it for all the near data including the stack

Well, the same reasoning could be used for HEAP: the OS calculates the amount to be used, just like stack, but there is no HEAP segment declared. The reason is there doesn't need to be, since as you say there are likely no symbols produced in that segment. OWC's use of STACK directly after BSS makes more reason to include it, while with ELKS it is entirely dynamically allocated and never statically allocated - thus never needs to be a linker segment.

> but the linker has to allocate space on the output virtual address space depending on the stack location and the size that is derived from linking rules (in your case DOSSEG).

You make a point about DOSSEG; since ELKS is using DOSSEG, perhaps we ought to define a STACK segment, even if unused.

> No, GROUP directive has nothing to do with segment ordering. It only define which segments can be addressed by appropriate GROUP.

OK, good to know. I had thought GROUP defined ordering. In that case, I definitely want keep all the segment definitions as they are in watcom/asm/segments.asm, as in that file they are ordered using the SEGMENT or CLASS directives.

> The Order of segments in OS/2 format for ELKS can be change because OW linker can order segments by custom rules.

I am not aware how to tell the OW linker how to order segments. Nonetheless, segments.asm works well.

> But it is not expected to have any symbol in stack or heap segment that it should be OK. If you expect ELKS format and calculate address of stack and heap by startup code then no problem because both segments are continuous on the end of memory block.

Agreed.

> I don't think this is the right way

Ok, after much discussion I am OK with adding back in STACK segment, as it is expect in DOSSEG programs and the original OW ASM segments.asm file defines it. So, go ahead and add it back in, but in segments.asm rather than crt0.S. That will allow us to remove warning 1014 from linker, which also could be good in case somebody uses ewlink to build ASM-only program and forgets stack.

We will also want to add back in `-Wl,option -Wl,stack=0x1000` as discussed. 

In segments.asm, we won't want to have STACK_SIZE equ, just empty segment declaration that is currently commented out, also with GROUP including STACK as last element.

jmalak:
No problem, I think all is clear. I will update this PR, I hope to final version.

Note about segment ordering, now it is control by DOSSEG directive that segments.asm has no impact to ordering, it has sense without DOSSEG directive but then it must be first linked module. It is reason for use DOSSEG that you can have any order of modules, because ordering is made by linker and by exact rules. If you are interested in segment ordering then looks into OW linker documentation for following topics
https://open-watcom.github.io/open-watcom-v2-wikidocs/lguide.html#The_ORDER_Directive
https://open-watcom.github.io/open-watcom-v2-wikidocs/lguide.html#The_SORT_Directive
https://open-watcom.github.io/open-watcom-v2-wikidocs/lguide.html#The_DOSSEG_Option

ghaerr:
> Note about segment ordering, now it is control by DOSSEG directive that segments.asm has no impact to ordering, it has sense without DOSSEG directive but then it must be first linked module.

Wow, very interesting. I didn't realize that DOSSEG is controlling order. I also didn't know DOSSEG had so many segment definitions, I thought it only had CODE, DATA, BSS and STACK (classes). 

I had previously worried about whether segments.asm or any other file could be forced to be first file to wlink, but thought that might be hard to enforce, and then thought segments.asm forced the order. Thank you for the clarification. I still would like to keep the segment definitions in segments.asm (I guess now as documentation only, really), but now I understand why you felt they could be removed without problems. So we will definitely be keeping DOSSEG for now!

> If you are interested in segment ordering then looks into OW linker documentation for following topics

Thanks, I'll take a look.

jmalak:
> > Note about segment ordering, now it is control by DOSSEG directive that segments.asm has no impact to ordering, it has sense without DOSSEG directive but then it must be first linked module.
> 
> Wow, very interesting. I didn't realize that DOSSEG is controlling order. I also didn't know DOSSEG had so many segment definitions, I thought it only had CODE, DATA, BSS and STACK (classes).
> 
> I had previously worried about whether segments.asm or any other file could be forced to be first file to wlink, but thought that might be hard to enforce, and then thought segments.asm forced the order. Thank you for the clarification. I still would like to keep the segment definitions in segments.asm (I guess now as documentation only, really), but now I understand why you felt they could be removed without problems. So we will definitely be keeping DOSSEG for now!
> 

Be carefull DOSSEG doesn't define any segment only switch on DOS memory layout sorting rules for all segments in your executable.

Yes, use DOSSEG simplify linking that is not dependant on modules order and define segment ordering.

Return to segments.asm it was original OW startup code (your crt0.c now) and this code use most of segments therefore here is defined DGROUP group with all segments. Maybe that in very looong history it was used also for segment ordering.

ghaerr:
> Be carefull DOSSEG doesn't define any segment only switch on DOS memory layout sorting rules for all segments in your executable.

Yes, good point. So in reality ELKS doesn't concern itself with segment ordering, but instead is concerned with fully-linked binaries in NE format; ELKS will load and relocate any number of arbitrary sections into memory, and handle the specially-numbered auto data segment. The segment ordering problems are all solved by the OW linker. All of this is part of the reason why I gave up on using outside tools to convert ELF -> OBJ or OBJ -> ELF, then load and relocate, because here to this day in 2024, there is still the requirement for good old *linker magic*!! Using outside tools gets to be quite hard guessing what magic might be used when.

> Maybe that in very looong history it was used also for segment ordering.

Yes, that might be so in old 8086 MASM history. Since there are so many variations of ASM source handling out there, I try to just keep things simple as possible. In this case, I just used the OW segments.asm source with little modifications, thinking this would bring the most compatibility while moving forward quickly.

ghaerr:
> With regards the ELKS memory layout and your allocation logic you need to rework OW stack checking a little bit.
> OW uses variable stacksize which hold size of stack from header that you need to update it to follow boundary between heap and stack by some logic if near heap is allocated over original stack size.
> I am not sure if it is good practice to do it this way. You should be able to guarantie stack size which is more important then near heap.

Yes, understood. ELKS doesn't actually permit or allow changed stack size, and keeps an internal per-task kernel variable that shows where the (max) bottom of stack is.  Since that variable is only in the kernel, ELKS passes the allowed max stack size to application process in DX at startup and this is used to initialize `__stacklow` for `stackavail` function. So everything should work properly.

The kernel of course also keeps the current heap value and won't allow the heap to grow into the allocated stack area.

During each system call, ELKS can be told to check the process SP value and check for stack overflow using the kernel variables. Thus, there is not currently a need for CRTL stack checking, except for the `alloca` function which requires `stackavail` with OW.

PR #1944, opened 2024-07-29
Remove OpenWatcom default symbols and residual code from libc

jmalak:
- remove OpenWatcom default CRTL symbols by -zls option, they are not necessary for ELKS libc
- remove all special segments except init/fini segments used by LIBC code, because this code is not used by ELKS libc
- change entry point **`_start_`** symbol to gcc **`_start`** for compatibility

### DIFF
--- a/elks/tools/objtools/ewcc
+++ b/elks/tools/objtools/ewcc
@@ -33,6 +33,7 @@ WATCINCLUDE=$WATCOM/h
 # -march=i86                # 8086 codegen
 # -std=c99                  # -Wc,-za99
 # -Wc,-zev                  # enable void arithmetic
+# -Wc,-zls                  # remove automatically inserted symbols
 # -Wc,-wcd=N                # disable warning N
 # -Wc,-fpi87                # inline 8087 fp
 # -Wc,-x                    # ignore INCLUDE environment variable
@@ -57,6 +58,7 @@ CCFLAGS="\
     -std=c99                        \
     -Wc,-fpi87                      \
     -Wc,-zev                        \
+    -Wc,-zls                        \
     -Wc,-x                          \
     -fno-stack-check                \
     -fnostdlib                      \
--- a/libc/watcom/asm/fchop87.asm
+++ b/libc/watcom/asm/fchop87.asm
@@ -44,7 +44,7 @@
 include mdef.inc
 include math87.inc
 
-        xrefp           __8087  ; indicate that NDP instructions are present
+;        xrefp           __8087  ; indicate that NDP instructions are present
 
         modstart fchop87
 
--- a/libc/watcom/asm/segments.asm
+++ b/libc/watcom/asm/segments.asm
@@ -25,61 +25,12 @@
 ;*
 ;*  ========================================================================
 ;*
-;* Description:  Watcom C segment ordering and null pointer section offsets
+;* Description:  Watcom C segment ordering for init/fini LIBC
 ;*
 ;*****************************************************************************
 
 
-        name    cstart
-        assume  nothing
-
- DGROUP group _NULL,_AFTERNULL,CONST,STRINGS,_DATA,DATA,XIB,XI,XIE,YIB,YI,YIE,_BSS
-
-; this guarantees that no function pointer will equal NULL
-; (WLINK will keep segment 'BEGTEXT' in front)
-; This segment must be at least 2 bytes in size to avoid confusing the
-; signal function.
-; need a symbol defined here to prevent the dead code elimination from
-; eliminating the segment.
-; (the int 3h is useful for quickly revealing jumps to NULL code pointers)
-
-BEGTEXT  segment word public 'CODE'
-        assume  cs:BEGTEXT
-        int     3
-        nop
-__begtext label byte
-        assume  cs:nothing
-BEGTEXT  ends
-
-_TEXT   segment word public 'CODE'
-
-FAR_DATA segment byte public 'FAR_DATA'
-FAR_DATA ends
-
-        assume  ds:DGROUP
-
-_NULL   segment para public 'BEGDATA'
-__nullarea label word
-        db      'nul'
-        db      0
-        public  __nullarea
-_NULL   ends
-
-_AFTERNULL segment word public 'BEGDATA'
-        ;dw      0                       ; nullchar for string at address 0
-_AFTERNULL ends
-
-CONST   segment word public 'DATA'
-CONST   ends
-
-STRINGS segment word public 'DATA'
-STRINGS ends
-
-_DATA   segment word public 'DATA'
-_DATA   ends
-
-DATA    segment word public 'DATA'
-DATA    ends
+DGROUP  group XIB,XI,XIE,YIB,YI,YIE
 
 XIB     segment word public 'DATA'
 _Start_XI label byte
@@ -107,57 +58,4 @@ _End_YI label byte
         public  "C",_End_YI
 YIE     ends
 
-_BSS    segment word public 'BSS'
-        ;extrn   _edata                  : byte  ; end of DATA (start of BSS)
-        ;extrn   _end                    : byte  ; end of BSS (start of STACK)
-_BSS    ends
-
-;STACK_SIZE      equ     1000h
-;STACK   segment para stack 'STACK'
-        ;db      (STACK_SIZE) dup(?)
-;STACK   ends
-
-        assume  nothing
-        public  _cstart_
-
-        assume  cs:_TEXT
-
-_cstart_ proc near
-        dw      __begtext               ; make sure dead code elimination
-                                        ; doesn't kill BEGTEXT segment
-_cstart_ endp
-
-if 0
-        INIT_VAL        equ 0101h
-        NUM_VAL         equ 16
-NullAssign      db      '*** NULL assignment detected',0
-;       don't touch AL in __exit, it has the return code
-__exit  proc
-        public  "C",__exit
-        push    ax                      ; save return code on stack
-        mov     dx,DGROUP
-        mov     ds,dx
-        cld                             ; check lower region for altered values
-        lea     di,__nullarea           ; set es:di for scan
-        mov     es,dx
-        mov     cx,NUM_VAL
-        mov     ax,INIT_VAL
-        repe    scasw
-        je      L7
-;
-; low memory has been altered
-;
-        pop     bx                      ; restore return code
-        mov     ax,offset NullAssign    ; point to msg
-        mov     dx,cs                   ; . . .
-
-; input: ( char __far *msg, int rc ) always in registers
-;       DX:AX - far pointer to message to print
-;       BX    - exit code
-
-__exit  endp
-endif
-
-_TEXT   ends
-
-        end     _cstart_
+        end
--- a/libc/watcom/syscall/crt0.c
+++ b/libc/watcom/syscall/crt0.c
@@ -10,15 +10,18 @@
 #include <unistd.h>
 #include <errno.h>
 
-/* Watcom extern code refs are sym_, extern data refs are _sym */
+/*
+ * create STACK segment which is required for applications
+ * size is setup by linker to default size 4096 bytes
+ * or to explicitly defined size by linker directive "OPTION STACK=..."
+ */
+#pragma data_seg("STACK","STACK")
 
-/* external references created by Watcom C compilation - unused */
-int _argc;              /* with declaration of main() */
-int _8087;              /* when floating point seen */
+/* return back to default data segment */
+#pragma data_seg()
 
-/* cstart_ is an external reference created by Watcom C, pulls in asm/segments.asm */
-#pragma aux cstart_ "_*" modify [ bx cx dx si di ]
-extern void cstart_(void);
+/* ELKS modified calling convention */
+#pragma aux elkscall __modify [ __bx __cx __dx __si __di ]
 
 static noreturn void sys_exit(int status);
 #pragma aux sys_exit =         \
@@ -33,7 +36,7 @@ noreturn void _exit(int status)
     sys_exit(status);
 }
 
-#pragma aux exit modify [ bx cx dx si di ]
+#pragma aux (elkscall) exit
 noreturn void exit(int status)
 {
     __FiniRtns();
@@ -46,7 +49,7 @@ noreturn void exit(int status)
  *
  * main(ac, av) called by AX, DX (small/medium) or AX, CX:BX (compact/large) model
  */
-#pragma aux main "*" modify [ bx cx dx si di ]
+#pragma aux (elkscall) main "*"
 extern int main(int argc, char **argv);
 
 /* global variables initialized at C startup */
@@ -65,7 +68,7 @@ unsigned int stackavail(void)
     return (_SP() - __stacklow);
 }
 
-#pragma aux premain "*" modify [ bx cx dx si di ]
+#pragma aux (elkscall) premain "*"
 #if defined(__SMALL__) || defined(__MEDIUM__)
 static noreturn void premain(void)
 {
@@ -156,4 +159,5 @@ noreturn static void _crt0(void);
 #endif
 
 /* actual program entry point */
+#pragma aux _start "*"
 noreturn void _start(void) { _crt0(); }
--- a/elks/tools/objtools/ewlink
+++ b/elks/tools/objtools/ewlink
@@ -24,18 +24,15 @@ fi
 
 ELKSLIBC=$TOPDIR/libc
 
-# Warning 1014: stack segment not found
-
 LDFLAGS="\
     -bos2                           \
     -s                              \
     -fnostdlib                      \
-    -Wl,disable -Wl,1014            \
-    -Wl,option -Wl,start=_start_    \
+    -Wl,option -Wl,start=_start     \
     -Wl,option -Wl,dosseg           \
     -Wl,option -Wl,nodefaultlibs    \
-    -Wl,option -Wl,stack=0x1000     \
     -Wl,option -Wl,heapsize=0x1000  \
+    -Wl,library -Wl,$TOPDIR/libc/libc.lib \
     "
 
 while true; do
@@ -71,8 +68,8 @@ if [ "$OUT" == "" ]
     OUT=${PROG%.obj}.os2
 fi
 
-echo owcc $LDFLAGS -o $OUT $@ $ELKSLIBC/libc.lib
-owcc $LDFLAGS -o $OUT $@ $ELKSLIBC/libc.lib
+echo owcc $LDFLAGS -o $OUT $@
+owcc $LDFLAGS -o $OUT $@
 
 # convert to ELKS a.out format
 #os2toelks -f elks -o $OUT $OUT.os2
--- a/libc/watcom.inc
+++ b/libc/watcom.inc
@@ -19,12 +19,14 @@ CARCH =\
     -fno-stack-check                \
     -Wc,-fpi87                      \
     -Wc,-zev                        \
+    -Wc,-zls                        \
     -Wc,-x                          \
     -Wc,-wcd=303                    \
     -fnostdlib                      \
 
 AARCH=\
     -m$(MODEL)                      \
+    -zq                             \
     -0                              \
     -fp0                            \
     -fpi87                          \
@@ -35,7 +37,7 @@ AS=wasm
 
 CFLAGS=$(CARCH) $(INCLUDES) $(DEFINES) -Wall -Os
 ASFLAGS=$(AARCH)
-ARFLAGS_SUB=-n -b -fo
+ARFLAGS_SUB=-n -q -b -fo
 
 %.obj: %.c
 	$(CC) -c $(CFLAGS) -o $*.obj $<


ghaerr:
Hello @jmalak,

Thank you for these enhancements, the changes cover a wider variety of things happening with ELKS. I have a few comments about some of them which I will comment on separately.

Thank you!

libc/watcom/asm/fchop87.asm
```diff
@@ -44,7 +44,7 @@
 include mdef.inc
 include math87.inc
 
-        xrefp           __8087  ; indicate that NDP instructions are present
+;        xrefp           __8087  ; indicate that NDP instructions are present
```
ghaerr, line 47: This is fine for now, as I have not yet figured out how to use OWC floating point with ELKS without requiring a full compile/replacement of the OWC soft floating point and math libraries - which look like a very large effort. (If both the 8086 ASM routines and the OWC math library could be moved easily to libc/watcom/math and built, that would be great).

Currently, OWC applications running on ELKS require a hardware 8087+ in order to run, something that definitely won't work for some of our user base.  I was initially thinking of using `__8087` as a method of knowing whether floating point was used by the program, as my intention was to use weak symbol linking and have the stdio/printf library pull in `dtostr` only when floating point was used, to keep things small. This is work in progress. It is important there be a way to use weak linking when floating point is used in an application. Weak linking is supported by ia16-elf-gcc and ELKS [here](https://github.com/ghaerr/elks/blob/master/libc/include/sys/linksym.h), and I recently added support for weak linking with ELKS to OWC in the same file.

I tried setting up OWC with soft float but noticed that since it's not IEEE 754 compatible, the ELKS math library (libc/math) I ported from elsewhere won't work. Thus I have had to stick with hw float. I see that OWC has very fancy "8087-detection" support where apparently FPU instruction trap and some OS or library code replaces the FPU instruction with a software trap to a soft fp emulator. I deemed this as way too complicated for ELKS (at least at the moment). 

ia16-elf-gcc has automatic soft floating point, which is currently the default and probably what we'd prefer, even in its not IEEE 754 compliant. The demo program for this is [basic](https://github.com/ghaerr/elks/tree/master/elkscmd/basic). It now compiles and runs using hw float but no math library (see elkscmd/basic/host.h).

What are your thoughts on all this?

libc/watcom/asm/segments.asm
```diff
@@ -25,61 +25,12 @@
...
-; need a symbol defined here to prevent the dead code elimination from
-; eliminating the segment.
-; (the int 3h is useful for quickly revealing jumps to NULL code pointers)
-
-BEGTEXT  segment word public 'CODE'
-        assume  cs:BEGTEXT
-        int     3
-        nop
-__begtext label byte
-        assume  cs:nothing
```
ghaerr, line 51: I would prefer to keep the existing BEGTEXT segment as the int3/nop guarantees function pointers won't be NULL, as well as (int)function_ptr > 2 so that signals work (SIG_DFL, SIG_IGN are 0 and 1).

Likewise, the _NULL segment below is required so that zero-valued pointers don't arbitrarily scribble into the application data segment. This happens a lot on 8086 because there is no memory management.

libc/watcom/asm/segments.asm
```diff
@@ -25,61 +25,12 @@
...
-CONST   ends
-
-STRINGS segment word public 'DATA'
-STRINGS ends
-
-_DATA   segment word public 'DATA'
-_DATA   ends
-
-DATA    segment word public 'DATA'
-DATA    ends
```
ghaerr, line 82: The reason for all these segments is that the OBJ files produced by OWC may contain references to them by name, and I thought it important to ensure that there were in fact named segments in this file in order to specify and guarantee the order they are linked. It is quite important that the compilation/linking process controls the ordering of segments, as other ia16-elf-gcc linker script magic sometimes uses them, although I have not implemented them yet. Thus, I would prefer that none of these definitions are removed, including the DGROUP above.

Also, I have some other tools (OBJ file converters, dumpers, etc) that reference these segments. These tools can be seen in elks/tools/objtools.

libc/watcom/asm/segments.asm
```diff
@@ -107,57 +58,4 @@ _End_YI label byte
...
-
-        assume  nothing
-        public  _cstart_
-
-        assume  cs:_TEXT
-
-_cstart_ proc near
-        dw      __begtext               ; make sure dead code elimination
-                                        ; doesn't kill BEGTEXT segment
-_cstart_ endp
```
ghaerr, line 128: Even though `_cstart_` is not longer used (thank you), we need a way too ensure that dead code elimination doesn't remove the `__begtext` reference, so that (int)function pointers > 2. That is why I took advantage of the default OWC `_cstart_` symbol and left this in.

libc/watcom/asm/segments.asm
```diff
@@ -107,57 +58,4 @@ _End_YI label byte
...
-        pop     bx                      ; restore return code
-        mov     ax,offset NullAssign    ; point to msg
-        mov     dx,cs                   ; . . .
-
-; input: ( char __far *msg, int rc ) always in registers
-;       DX:AX - far pointer to message to print
-;       BX    - exit code
-
-__exit  endp
-endif
```
ghaerr, line 159: I left all this code in to remind me of how OWC handles checking whether a NULL pointer has been referenced. I would also like this left in as this is under development but not finished yet.

libc/watcom/syscall/crt0.c
```diff
@@ -10,15 +10,18 @@
...
-int _argc;              /* with declaration of main() */
-int _8087;              /* when floating point seen */
+/* return back to default data segment */
+#pragma data_seg()
 
-/* cstart_ is an external reference created by Watcom C, pulls in asm/segments.asm */
-#pragma aux cstart_ "_*" modify [ bx cx dx si di ]
-extern void cstart_(void);
+/* ELKS modified calling convention */
+#pragma aux elkscall __modify [ __bx __cx __dx __si __di ]
```
ghaerr, line 24: The purpose of this pragma aux modify, as well as the one below for `exit` and `premain` are to eliminate the need for saving registers that OWC otherwise emits. This is not an or the ELKS calling convention but a way to lessen needless code that OWC produces in certain special cases either for functions only called in crt0.s or OS exit, in an attempt to produce a small executable. Thus, I am not sure whether it is a good idea to call this `elkscall` to avoid confusion. Something like `noregsave` might be better.

I have some other questions about how pragma aux modify might be used to remove push bp/mov bp,sp/pop bp sequences in other functions, but we can talk about that later.

libc/watcom/syscall/crt0.c
```diff
@@ -46,7 +49,7 @@ noreturn void exit(int status)
  *
  * main(ac, av) called by AX, DX (small/medium) or AX, CX:BX (compact/large) model
  */
-#pragma aux main "*" modify [ bx cx dx si di ]
+#pragma aux (elkscall) main "*"
```
ghaerr, line 52: The reason for this pragma aux for `main` is actually to keep OWC from automatically renaming the symbol with an underscore and force a `_cdecl` naming convention, rather than an ELKS calling sequence of some kind. The same pragma is used in elks/include/arch/cdefs.h, which with this change now does not match.

The second reason for this pragma is to inhibit register saves in main, but frankly that probably doesn't matter as much, except that its easy to do in the same pragma.

Controlling the process by which `main` is called is important in crt0.s due to the way that we have had to rewrite the argc/argv vector (passed with 2-byte pointers by ELKS) to 4-byte pointers in compact and large models. I have tried to design the whole process so that needless stack pushing or register saves are performed, and instead use the very nice OpenWatcom register calling model, which I am finding is quite cool compared to stack-calling convention(s). (The syscall wrappers are a good example of this, where very little code is generated to convert OWC syscall wrappers to an ELKS system call. This eliminated lots of code that ELKS uses in the ia16-elf-gcc cdecl default calling convention, linked into every program). Overall, I am extremely impressed with the OWC register allocations produced, they are quite different than ia16-elf-gcc, even though in some cases quite a few SS: overrides are seen (but with big benefit allowing DS to be scratch at times).

libc/watcom/syscall/crt0.c
```diff
@@ -156,4 +159,5 @@ noreturn static void _crt0(void);
 #endif
 
 /* actual program entry point */
+#pragma aux _start "*"
```
ghaerr, line 162: Does this pragma do anything? I don't understand what adding it does?

elks/tools/objtools/ewcc
```diff
@@ -33,6 +33,7 @@ WATCINCLUDE=$WATCOM/h
 # -march=i86                # 8086 codegen
 # -std=c99                  # -Wc,-za99
 # -Wc,-zev                  # enable void arithmetic
+# -Wc,-zls                  # remove automatically inserted symbols
```
ghaerr, line 36: Is there a list of symbols OWC might automatically insert? These could be useful in various circumstances for ELKS libc, just like they are in OWC libc.

libc/watcom/asm/fchop87.asm
jmalak: I am not expert on the math and FPU that it is only my opinion. If I understand correctly then now ELKS has no support for FPU. Also support from compiler is needed. I don't know what is possible with ia16-gcc if it is able recognize use of  floating point and generate some symbols to signal linker to load some floating point stuff.
At minimum you need some hardware support for FPU in ELKS to handle FPU exceptions. Too many unclear things for me and also my knowledge in this area is limited. Most complex it is for DOS because there is no fpu support from OS. Other system have some support for FPU smaller or bigger that you don't need handle FPU hw exception you get it from OS. The rest is pure math and is same for all. May be solution to look for existing FPU math implementation (IEEE 754 compliant) and modify it for ELKS FPU stuff in OS which is necessary. OpenWacom use very old implementation still not fully IEEE 754 compliant.

ghaerr: The ELKS kernel doesn't have any FPU support, it catches but ignores any FPU-generated exceptions. The kernel itself uses no floating point.

Application support for FP is soft-only using ia16-elf-gcc, and automatically provided, which is nice. I added a math library which works with the IEEE 754 used by the gcc basic soft float library.

What I would like to see is options for using OWC to compile applications with self-contained soft FP support, but I have not (yet) got that to work, as large OWC ASM library of routines need to be built as part of ELKS libc. Once this is done, likely OWC math library can also be built and problem solved. During this process, it would be nice if the kernel could tell whether an executable used soft vs hw float.

At this point, it is probably not important that ELKS use exception-based hard/soft FP support, although it might be nice to have a separate set of `ewcc` options that allow for compiling an application to require FPU vs soft-only.

jmalak: OW has FPU math which is not depend on host OS and small part which handle exceptions which is OS dependant. Main part is stuff for emulator integration in CRTL but if you always expect FPU HW then it can be abandoned completely. I think that OW code is well divided to part independent on OS and OS specific that can be simply modified for new platform.
There are versions for different OS-es that you could look on how it is done and get some idea how simply implement it on ELKS. Maybe 16-bit OS/2 version could be good starting point. 16-bit QNX version may be also interesting, but 16-bit QNX is more different. It is based on installation emulator to OS if no FPU HW then standard FPU math is used with or without FPU.

libc/watcom/asm/segments.asm
jmalak: I understand, for OW version of libc it is possible to use it because wlink support these BEGTEXT, NULL and AFTERNULL segments and ensure they are on offset 0. But these guards is active only for near code/data.

libc/watcom/syscall/crt0.c
jmalak: > Owner

It changed symbol name to be compatible with gcc generated code now it is  `"_start"` before it was `"_start_"`
#pragma aux is very powerful and this use define mangling of symbol.
`"*"` means use symbol as is defined not use any mangling.
This name will be always same as C symbol even if you change calling convention etc.
by example 
`#pragma aux _start "__my_*"`
output symbol as `"__my__start"` in object file.
star symbol represent C name.
You can simply change mangling for any OW calling convention, but it is danger you need to understend what you are doing.

elks/tools/objtools/ewcc
jmalak: No, but below is list from C compiler source code
```
"__DLLstartw_"
"__DLLstart_"
"_wstartw_"
"_wstart_"
"_cstartw_"
"_cstart_"
"_fltused_80bit_"
"_fltused_"
"_big_code_"
"_small_code_"
"__init_87_emulator"
"__init_387_emulator"
"__old_8087"
"__8087"
"_fltused_"
"__wargc"
"__argc"
"__wargc"
"__argc"
"_wargc"
"_argc"
"__init_default_win"
"__WATCOM_Prelude"
"__p5_profile"
"__new_p5_profile"
```

Take into account that it is full list but they are genearted in dependency of compiled code or target platform, some are specific to architecture etc. Best is to look into compiler code because there is visible condition when it is used. relevant code is in bld/cc/c/cfeinfo.c file.
Anyway now it is suppresed by -zls option that none of this symbols is included in generated code.

libc/watcom/syscall/crt0.c
jmalak: I used first what I have in mind I will change it as you proposed.
Of cause it is possible define it as macro too because #pragma aux fully support macros, but #pragma aux alias is more transparent with single complex definition.

libc/watcom/syscall/crt0.c
jmalak: the main function is defined by C standard and is entry point to C standard environment that it should be handled carefully.
Thanks #pragma aux you can change it completely but caller in startup code must follow you definition of main.
if you want use cdecl calling convention in startup code then you can change calling convention of main by #pragma aux to cdecl standard way as below.
`#pragma aux (__cdecl) main`
It do main cdecl function with all necessary changes parameters on stack name without mangling, changes preserved registers etc.
But don't use this method for other function it is specific for main. For other functions use standard modifier __cdecl in C code.

libc/watcom/asm/segments.asm
jmalak: The wlink use standard rules for ordering segment as is defined for OMF format that put together segments with compatible segment attributes. there is several levels of comparision. The name of segment is mostly not important except some OW special segments that change of segment name don't cause any problem. the main attribute is segment class as 'CODE' 'DATA' STACK' 'BSS' and it is also used by DOSSEG segment ordering. But change segment name has sense if you mixed modules from different toolchains because each use different names and also it depends on target system. In such case you could need some segment name changes etc. it is possible. Also it is possible to use custom segment ordering that you can setup rules for all toolchains modules without change segment names. It is not necessary to use some specialised tools it can be done by wlink configuration.

libc/watcom/asm/segments.asm
jmalak: It is clear I will do necessary change for these guards to works properly.
But be carefull it depends on memory model used for code.
You declared that libc is using large memory model that I will suppose only large memory model or do you expect libc for multiple memory models?

libc/watcom/syscall/crt0.c
jmalak: I looked into crt0 code and it is a little strange because you change standard C function calling convention that it will have a problems with OW compiler.
What sense for such changes if crt0 is specific for OW that can use OW standard behaviour as main calling convention etc without any change of prototype and #pragma aux?
All C function can use OW default calling convention and not need any hack.

ghaerr:
Thank you for your answers, you are a wealth of knowledge about OW and I am learning things seemingly not documented looking at your code. OTOH, there is lots of OW documentation, and I haven't yet read it all :)

This PR changes quite a lot of things and understand that OW support in ELKS is in ongoing development. I would prefer smaller changes and less mass deletions (e.g. in segments.asm) as there are several things happening at once. I am trying to simultaneously learn about OW's features as well as keep original OW CRTL code when possible as I plan on adding other OW CRTL support after reading through many more source files. In general I see no reason to delete lots of code when everything is working perfectly, although I see many examples of enhancements provided with your work. I will comment on these issues separately so that we can more easily get your enhancements committed. In fact, it may be easier to submit new, smaller PRs with things we've agreed on.

All changes/enhancements should be tested with at least 3-4 applications programs (e.g. I use elkscmd/basic as it is larger and easily tested), all memory models are currently supported and everything is working with no problems. I have put a large amount of time into getting OpenWatcom support into ELKS and want to ensure everything stays operational without me having to perform the low-level testing all over again. The process by which OW has been included into ELKS is quite long-winded as I have tried many differently approaches, starting with [owtarget16](https://github.com/pts/owtarget16) and its OS/2 executable to ELKS a.out converter, [then rewriting it heavily](https://github.com/ghaerr/elks/blob/master/elks/tools/objtools/os2toelks.c), to [OBJ file to ELF conversion](https://github.com/pts/minilibc686/blob/master/tools/omf2elf.c) and finally coming to the current approach of producing OS/2 NE format executables and directly loading and relocating them. Thus various code is in different states throughout this multimonth effort to get things working. The latest desire is getting the entire ELKS C library compiled (mostly done) and to keep as much OW source as possible so that pulling in other OW CRTL source will be easy and we don't have to go backwards adding back in things taken out so that everything looks perfect now.

 I will try to comment "approved" or "rejected" so as to make more quick progress. Thank you for your interest in this!

elks/tools/objtools/ewcc
ghaerr: Thanks for the list. Approved. In the future, it may be good to have OW insert `fltused` or similar, but we'll worry about that later.

libc/watcom/syscall/crt0.c
ghaerr: Thanks, I knew about the aux "\*" but continue to get a bit confused as to when an underbar is prefixed or suffixed (appears to be code vs data) and also exactly what "\*" means when aux has underbar prefixed name. I think I understand now. Approved.

libc/watcom/syscall/crt0.c
ghaerr: This is one of the early changes made when trying to optimize function calling to not save registers when not needed, along with some changes to main's signature. They are no longer needed, and now I agree this is over-optimization and main should not have and does not need hack to not save registers and should be treated as normal calling convention. Approved, but please test that elkscmd/basic still compiles and runs:
```
ewcc basic.c
ewcc host.c
ewcc host-stubs.c
ewlink basic.obj host.obj host-stubs.obj
```

Please delete special calling convention for main in both crt0.c as well as include/arch/cdefs.h.

jmalak:
I understand, I don't think that code need big or significant change but rather remove some construct which complicate thing because doesn't use OW default behaviour and introduce other non default solution etc.
I will update PR to something useful to include all your notes.

May be good idea add automation for build OW libc by Github actions and automate testing.
We have a project https://github.com/open-watcom/setup-watcom which was designed by developers to help with OW installation. It completely automates installation of OW into Github Actions that you can use OW in your project simply for build or testing. It is about a few lines which you need to add in Actions nothing strange. It is used by several GitHub projects.

It is pity that you don't ask to add a.out executable format to wlink. It doesn't require big effort because as all OW code it is well separated modular software that only new writer for a.out executable format is necessary. I don't think that a.out format is too complex.

libc/watcom/syscall/crt0.c
ghaerr: Ok, go ahead with `noregsave` pragma aux alias instead of `elkscall` for "reducing unneeded register saves" in crt0.c functions, that will be easier understood to programmers what is being done. Approved.

It is OK to get rid of kluge `cstart_` here, but we have to ensure that watcom/asm/segments.asm always gets pulled in somehow. Previously, it was pulled in because OW defined this symbol. It seems it was never needed to be defined in crt0.c however.

libc/watcom/asm/segments.asm
ghaerr: Rejected. Please leave `if 0` code in as I plan on using it later on.

libc/watcom/asm/segments.asm
ghaerr: For the time being, the OW memory model is set by reading libc/watcom.model. Small, Compact, Medium and Large models are supported and all work.

`_cstart_` can be removed here, but we need a way to ensure that the BEGTEXT segment is included with 2 bytes used so that `signal` function works properly and no function address == 0 or 1. The `__begtext` symbol and usage can be removed if you can think of a better way to force the linker to include the segment, even with dead code optimization is on. If this cannot be done, then we can possibly reference `__begtext` or a better name from crt0.c in stead of using `_cstart_` kluge.

libc/watcom/asm/segments.asm
ghaerr: I want to keep all named segments in segments.asm and not remove any of them. This file is the place for all segment-ordering and segment-related linker special treatment, as well as "documentation" of the application memory layout.

Rejected.

libc/watcom/asm/segments.asm
ghaerr: Please keep BEGTEXT, NULL, AFTERNULL etc as-is. If you think a better way of ensuring BEGTEXT is included than using `__begtext` (and/or `_cstart_`) that is OK. Deletion of segment definition rejected.

It is also OK for only small and medium models to work with near NULL pointer. I am not too worried about creating special handling for far-data programs, unless you have good idea for it.

libc/watcom/asm/fchop87.asm
ghaerr: Approved. We will worry about soft vs hard float support in a separate conversation, thank you.

ghaerr:
>  I don't think that code need big or significant change but rather remove some construct which complicate thing because doesn't use OW default behaviour and introduce other non default solution etc.
I will update PR to something useful to include all your notes.

Agreed, thank you. In general, I approved almost all your enhancements except for deleting segment and if 0 definitions in segments.asm.

> May be good idea add automation for build OW libc by Github actions and automate testing.

Yes, I plan on that. The reason it is not yet done is because I have to setup build of OW repo and that takes a very long time to compile, along with the already long time to compile gcc occasionally. We are already using cache. Also, it is still optional for users to choose OW vs GCC for compiling application programs. Kernel is still compiled using only GCC.

My long term goal, which would be very very cool would be to create OW repo toolchain that would run directly on ELKS itself, using OS/2 NE format. Imagine that, OW running on a new OS as a super-featured compiler and able to compile itself, all applications, and the kernel!!! This would likely require ELKS syscalls to be added to OW repo and ELKS C library compiled separately somehow, but there may be another way.

> We have a project https://github.com/open-watcom/setup-watcom which was designed by developers to help with OW installation.

Thanks, I will take a look at it.

> It is pity that you don't ask to add a.out executable format to wlink.

Well, I considered that but there is a big problem: the a.out executable format cannot handle an arbitrary number of (code or data) segments, which is absolutely required for OW large and medium model support. For that reason alone, I feel it is not worth it to add a.out format to wlink. It has taken me a while to realize exactly what is needed to support large/medium models, and the ELKS a.out format has been extended for ia16-elf-gcc medium model support, but that was not done in general fashion. Thus, the a.out format is still not suitable for OW or any other large model compiler. The a.out format could be extended again, but why bother, when OS/2 format already exists and supports arbitrarily large multi-segment definitions?

Another reason I went with OS/2 NE format is that we could now, in theory, support 16-bit OS/2 DLLs using OW's extensive options and SS != DS support. This is a whole another conversation, but potentially interesting. Unforunately, I do not believe that adding DLLs to ELKS really helps out, as main memory is in short supply and would only become really useful with XMS or EMS memory support for larger programs. The C library is too big to be permanently loaded.

Thank you!

jmalak:
Thanks for your comment.

I worked little bit on integration of OW checks for NULL pointers in code and data to works and there is crucial problem with your crt0.c.
It is impossible by OW C compiler create code label __begintext in C function body in C code segment.
Therefore it is necessary to move initial part of startup code to segments.asm code to be able do all necessary thing correctly and dead code elimination doesn't remove this code.

I propose following changes:
rename existing segments.asm to crt0.asm
rename crt0.c to cstart.c
move initial assembly code (crt0 function) from old crt0.c to new crt0.asm.
It was probably reason for OW to use startup code in assembly module because it can not be done in C code.

I finally understood what you were trying to achieve with construct
**`#pragma aux main "*" modify [ bx cx dx si di ]`**
It is wrong solution. Probably you want to remove useless frames for no returning functions.
For no returning functions OW has two possible solutions
1. C standard compatible **`_Noreturn`** modifier
2. OW own **`#pragma aux func aborts`**

First solution is for external function where C standard solution must be used because it must include frame with parameters passing with regard to target calling convention (non WATCOM).
Second solution is OW which optimize both side callee and caller part of code to minimise overhead.
The reason that you can not use it for external function is that OW compiler also optimize called function.
This is not possible for external functions.
Because libc is generated by OW and for use with OW compiler (no external functions from other toolchains) you can use second solution which do code more efficient.
The solution is simple you need define **`noreturn`** macro for OW compiler to **`__declspec(aborts)`** and you can remove these #pragma aux which is not working generally but only for your case and it is not efficient as OW aborts which also optimized called function.

Maybe better to close this PR and open new one for "Improve integration with OW".

ghaerr:
> I worked little bit on integration of OW checks for NULL pointers in code and data to works and there is crucial problem 

Well, you are now following some of the long road I've been on getting everything working correctly, and there is reason why everything has been done as is (mostly :) )

> I propose following changes:
rename existing segments.asm to crt0.asm
rename crt0.c to cstart.c
move initial assembly code (crt0 function) from old crt0.c to new crt0.asm.

OW support for ELKS is still under my heavily development and I don't want to rename crt0.c or segments.asm to fix a problem that isn't really broken. As mentioned above, there has been a long road to get where we are. Let me explain a bit more my thoughts on this. Everything is working well right now so there isn't a big need to make major changes for small optimizations.

When looking at the owtarget16 project (mentioned above), I liked the fact that a single .c file could be used to handle all application code startup, without the need for messy ASM files. These owtarget16 programs were designed to be *extremely* small, a matter of just a few bytes if required for very small programs. At ELKS, we share the same philosophy that small is beautiful. (Yes, now that C library, atexit processing, stdio initialization, constructors, destructors etc have been added, things aren't super small anymore, but that's the cost of useful C library). But the crt0.c could still be easily trimmed back for very small executables outside the C library for OW ELKS, if desired. This is sometimes important and special cases are already made for GCC ELKS executables to keep the binary sizes down since we boot from floppy around here!! :)

After moving away from owtarget16's approach to producing ELKS executables, since converting OS/2 binaries to a.out format can't work well for large/compact models, I built a crt0.c using the owtarget16 philosophy which handled all aspects of program startup, and then started working on the bigger problem of handling the near pointer argc/argv array created by `exec` that needs to be converted to far pointers, as well as adding constructors/destructors. It is very convenient to have the OS ASM startup code and the later C handling in the same file. Thus, I don't want to move ASM code out of crt0.c.

After learning more about how OW CRTL worked, the segments.asm file was added (after also realizing that there was no way to get around adding other OW ASM files like i4m.asm), which provided the specification for the segments used with the OW toolchain. While it may seem simple for us that now understand it, it took quite some time to figure out what was being used, and like the idea of keeping this information around for other programmers trying to quickly learn about OW support in ELKS. This is why I don't want to rename segments.asm to something different.

Sorry for the long rambling, but I wanted you to know where I'm coming from.

> It is impossible by OW C compiler create code label __begintext in C function body in C code segment.
Therefore it is necessary to move initial part of startup code to segments.asm code to be able do all necessary thing correctly and dead code elimination doesn't remove this code.

Might I remind you that all of the code in segments.asm that you're trying to remove is in fact copied from the original OW segments.asm file? The issue isn't that we need to move the `_crt0` function ASM code into another unrelated file but that dead code elimination requires that a symbol be included in segments.asm such that the original NULL function and data pointer code doesn't get removed. In order to do that, a much simpler approach might be taken - just move `_start` into segments.asm at the very beginning of segment BEGTEXT, then remove `__begintext` and the `dw __begintext` kluge below it.

The startup code in segments.asm could possibly be very simple and look like this:
```
BEGTEXT  segment word public 'CODE'
        assume  cs:BEGTEXT
_start: jmp _crt0               ; jump to C startup code (also prohibit function pointers < 2)
BEGTEXT ends
```
The only problem with this is that in large/medium models, we may need the `jmp` to be a far jump, so it likely needs to be ifdef'd. We don't want to always use a far jump as this requires a relocation entry for every executable. Having a relocation entry for large/medium models is OK since there will likely be tons of them anyways.

[EDIT: of course the other reason why this has become a problem is that a switch is being added to prohibit OWC from emitting special symbols... like `_cstart_`. Sometimes it is good that a compiler emits special symbols, to avoid all this workaround. Do we really need to remove special symbols? We might end up having to leave the option on for floating point detection soon. I do agree that removing `_cstart_` and `_argc` does clean things up, but `_cstart_` is quite useful and eliminates the problem of a near/far jump above.]

> It was probably reason for OW to use startup code in assembly module because it can not be done in C code.

I'm not sure about that - it seems that the OW CRTL has very expansive support for many features and these all require specialized startup, especially when you look at stack/heap/thread startup, etc.

ghaerr:
> Probably you want to remove useless frames for no returning functions.

Yes, I wanted to do that but also stop the large register pushes at function start.

> you need define noreturn macro for OW compiler to __declspec(aborts)

What is the difference between `__declspec(noreturn)` and `__declspec(aborts)`? I see you're changing this in #1946 but don't understand the difference in code generated. Does this change eliminate both the frame generation and additional register saves? (That would be nice).

> and you can remove these #pragma aux which is not working generally but only for your case and it is not efficient as OW aborts which also optimized called function.

Ok, sounds good.

jmalak:
I think it is misunderstanding, I think we agreed on this code to be there and not be removed anyway.
I tried to do code for NULL pointer in data and code workable and there is a main issue that you can not create public label in midle of C function which is required for NULL pointer check. That it is not possible to move this code into C source to be together with front in-line assembly code in C source.
Anyway I think the naming file segments.asm is confusing this code is part of the OW startup code and have nothing with segments. All segments are there because it was single file startup module fully in assembly code.  That's why I suggested renaming it to crt0.asm and move front of startup assembly code (which is in C module as in-line code) to this file. No change in code only moving it to assembly module to be able create labels and refere it. Existing C code would remain in the old crt0. OW inline assembler is not full assembler similary to other toolchains that functionality is limited. C and assembly address space are separated, you can only address global C symbols from assembler. If you don't want the checks for NULL pointers, nothing needs to be changed.
Anyway I checked existing code with dead code elimination and it was removed because there is missing external reference between C and assembly module.

ghaerr:
> I think we agreed on this code to be there and not be removed anyway.

Yes, we agreed to keep segments.asm null pointer allocation code etc, I was talking about your thoughts about removing `__begtext`.

> Anyway I think the naming file segments.asm is confusing this code is part of the OW startup code

Actually it isn't - there is no startup code executed, and the file segments.asm is the same name as in OW source, and mostly copied exactly as-is. Given your comments about segment ordering, this file isn't strictly necessary for OW-tool-generated startup, except we want null pointer check. Thus segments.asm is better name, IMO. It could be removed for size reasons for tiny executables so don't want to name it crt0.asm.

> If you don't want the checks for NULL pointers, nothing needs to be changed.

I do want the NULL pointer regions allocated. I think the real issue is that the desire to add compiler option to remove emitting special symbols is driving this change - is that option worth the changes, given that FPU support may require it to be removed in the future? I have already explained why I don't desire to change filenames at this time.

I'm fine with adding `_start` to segments.asm, but if that won't work, perhaps things should stay the same without adding compiler option? I plan on more research with creating super-tiny executables (this would include special startup code, possibly no constructor/destructors, and a special tiny printf.c in elkscmd/lib/tinyprintf.c, in use now for some small ELKS applications. The GCC constructor/destructor is very much smaller than OW but I'm OK with that since current plan is to use OW for large large-model executables since that is great new feature). Optimizing too much at this early stage in development may cause U-turns in the future as more requirements are fulfilled.

ghaerr:
> but _cstart_ is quite useful and eliminates the problem of a near/far jump above

Another option that could be taken in segments.asm would be to keep compiler option eliminating special symbols, but call `_cstart_` from crt0.c `exit` routine (after exit) to force `__begtext` to be included in segments.asm. In this way, crt0.c controls whether or not segments.asm is included (instead of compiler switch) which allows for good future flexibility in whether or not extra segments.asm null pointer checking is included or not for tiny executables. I kind of like this idea.

jmalak:
No, this issue has nothing to do with removing any symbols it is related to dead code elimination due to mising reference between these modules because code which need this reference is now divided to two modules and in C module you cannot do this.
Again we agreed that these symbols we will not remove only comment it out for future use as you plan it.

jmalak:
I use this PR for discussion only but existing code changes in this PR is nonsense with regards of previous discussion and changes will be minimal.

jmalak:
You are in better possition you handle only single system, but OW library need support all current OS-es that mostly it use general code and somewhere specialised code that not all is optimal. Also handling of errors by standards get some overhead.

PR #1945, opened 2024-07-31
add STACK segment for OW libc

jmalak:
Add **`STACK`** segment to **`segments.asm`** and remove disabling of OW linker message about missing stack
**`ewlink`** - remove compiler option **`-fnostdlib`** it is not used for linker, linker uses directive **`option nodefaultlibs`**

### DIFF
--- a/elks/tools/objtools/ewlink
+++ b/elks/tools/objtools/ewlink
@@ -24,13 +24,9 @@ fi
 
 ELKSLIBC=$TOPDIR/libc
 
-# Warning 1014: stack segment not found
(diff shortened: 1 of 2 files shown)


ghaerr:
Looks good, but don't you want to add STACK as the last definition in the GROUP statement on line 36 of segments.asm? The original OW code had it, and I removed it. It should be present since STACK is part of DGROUP if it is being defined.

jmalak:
Oh sorry I overlooked it. I will correct it.

PR #1946, opened 2024-07-31
ow-libc: use more eficient method #pragma aux aborts for functions wh…

jmalak:
…ich no return

for OW version of libc is possible to use more efficient method by #pragma aux aborts or by __declspec(aborts) to optimize code for no return function this method optimize both side caller code and no return function method with __declspec(noreturn) must be used in case that this function is external (non-WATCOM)

remove some hack which try to to remove calling frame but for special case only, now it works in any situation and for any calling convention

### DIFF
--- a/libc/include/sys/rtinit.h
+++ b/libc/include/sys/rtinit.h
@@ -32,8 +32,6 @@ extern struct _rt_init _Start_YI;
 extern struct _rt_init _End_YI;
 
 /* constructor/destructor functions called before main and after exit */
-#pragma aux __InitRtns modify [ bx cx dx si di ]
-#pragma aux __FiniRtns modify [ bx cx dx si di ]
 extern void __InitRtns(void);
 extern void __FiniRtns(void);
 
--- a/libc/watcom/syscall/crt0.c
+++ b/libc/watcom/syscall/crt0.c
@@ -16,11 +16,7 @@
 int _argc;              /* with declaration of main() */
 int _8087;              /* when floating point seen */
 
-/* cstart_ is an external reference created by Watcom C, pulls in asm/segments.asm */
-#pragma aux cstart_ "_*" modify [ bx cx dx si di ]
-extern void cstart_(void);
-
-static noreturn void sys_exit(int status);
+extern void sys_exit(int status);
 #pragma aux sys_exit =         \
     "xchg ax, bx"           \
     "xor ax, ax"            \
@@ -33,7 +29,6 @@ noreturn void _exit(int status)
     sys_exit(status);
 }
 
-#pragma aux exit modify [ bx cx dx si di ]
 noreturn void exit(int status)
 {
     __FiniRtns();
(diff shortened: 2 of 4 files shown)


ghaerr:
This looks good but I have a couple questions about how things work, without me having to go through and look at `wdis` output. I will post questions separately, thanks.

libc/include/sys/rtinit.h
```diff
@@ -32,8 +32,6 @@ extern struct _rt_init _Start_YI;
 extern struct _rt_init _End_YI;
 
 /* constructor/destructor functions called before main and after exit */
-#pragma aux __InitRtns modify [ bx cx dx si di ]
-#pragma aux __FiniRtns modify [ bx cx dx si di ]
 extern void __InitRtns(void);
```
ghaerr, line 35: Isn't the pragma aux required to tell OWC that the called function doesn't save bx/cx/dx/si/di? We don't want any extra code generated from the calling function (premain) to save registers if it thinks they may be modified by the constructor/destructors. I am trying to keep premain code very efficient and small.

Or is this pragma completely unneeded in this case?

libc/watcom/syscall/crt0.c
```diff
@@ -16,11 +16,7 @@
 int _argc;              /* with declaration of main() */
 int _8087;              /* when floating point seen */
 
-/* cstart_ is an external reference created by Watcom C, pulls in asm/segments.asm */
-#pragma aux cstart_ "_*" modify [ bx cx dx si di ]
-extern void cstart_(void);
-
-static noreturn void sys_exit(int status);
+extern void sys_exit(int status);
```
ghaerr, line 19: I guess what you're saying here is that `noreturn` is ignored since the code is inlined and emitted directly as specified, correct?

What about `static`? Is that also useless since the code is inlined?

libc/include/sys/rtinit.h
jmalak: You broked OW compiler logic by these #pragmas that it do ineficient things. I checked code and now without such strange pragmas is code little bit smaller because compiler replace some far calls by jumps. OW compiler propagate no return on the path that if you broke this path by your pragma then it is less eficient.
You can look on generated code. Now I removed all these strange pragmas everywhere and use __declspec(aborts) for noreturn macro, it is all. Compiler do all work for me and it works anywhere without strange pragmas and for any calling convention because compiler know all details about each function.
__declspec(aborts) switch on optimization to remove useless frames, and it requires also optimization on called function if it is no return, because if you remove frame then you need correct stack offset calculation in called function etc. It is what OW compiler do optimize both sides and then it is more efficien then C standard _Noreturn modifier which can only break flow control but cannot remove frame.

libc/watcom/syscall/crt0.c
jmalak: In-line code cannot be static, OW compiler accept it but it is not fully correct.
Take into account that #pragma aux is template which is instantiated when you use it.
You can have it in header file. If you are interested in such examples look into OW source code in bld/watcom/h/tinyio.h or int10.h or int33.h all these files contains many inline function templates that you can simply include such header file in your C code and use any #pragma aux template function, it include copy of inline code into C code.

ghaerr:
Thanks for your explanations!

PR #1948, opened 2024-08-01
[watcom] Remove OpenWatcom default symbols and residual code (dup #1944)

ghaerr:
Adds @jmalak's enhancements from closed #1944, with a few other modifications. Thank you, @jmalak, what do you think of these changes?

- remove OpenWatcom default CRTL symbols by -zls option, they are not necessary for ELKS libc
- change entry point _`start_` symbol to gcc `_start` for compatibility
- change ewlink to include libc using `-Wl,library` option
- removes forced `__8087` reference for now
- removes `__begtext` kluge used to force BEGINTEXT to be included in link
- Moves default `_start` to BEGINTEXT segment at offset 2

Moving `_start` to watcom/asm/segments.asm allows future plan to allow creation of very small executables without constructors/destructors, null pointer checks or argc/argv pointer rewriting by changing the executable entry point using `-Wl,start=` option and small changes to crt0.c

### DIFF
--- a/elks/tools/objtools/ewcc
+++ b/elks/tools/objtools/ewcc
@@ -33,6 +33,7 @@ WATCINCLUDE=$WATCOM/h
 # -march=i86                # 8086 codegen
 # -std=c99                  # -Wc,-za99
 # -Wc,-zev                  # enable void arithmetic
+# -Wc,-zls                  # remove automatically inserted symbols
 # -Wc,-wcd=N                # disable warning N
 # -Wc,-fpi87                # inline 8087 fp
 # -Wc,-x                    # ignore INCLUDE environment variable
@@ -57,6 +58,7 @@ CCFLAGS="\
     -std=c99                        \
     -Wc,-fpi87                      \
     -Wc,-zev                        \
+    -Wc,-zls                        \
     -Wc,-x                          \
     -fno-stack-check                \
     -fnostdlib                      \
--- a/elks/tools/objtools/ewlink
+++ b/elks/tools/objtools/ewlink
@@ -27,11 +27,12 @@ ELKSLIBC=$TOPDIR/libc
 LDFLAGS="\
     -bos2                           \
     -s                              \
-    -Wl,option -Wl,start=_start_    \
+    -Wl,option -Wl,start=_start     \
     -Wl,option -Wl,dosseg           \
     -Wl,option -Wl,nodefaultlibs    \
     -Wl,option -Wl,stack=0x1000     \
     -Wl,option -Wl,heapsize=0x1000  \
+    -Wl,library -Wl,$TOPDIR/libc/libc.lib \
     "
 
 while true; do
@@ -67,8 +68,8 @@ if [ "$OUT" == "" ]
     OUT=${PROG%.obj}.os2
 fi
 
-echo owcc $LDFLAGS -o $OUT $@ $ELKSLIBC/libc.lib
-owcc $LDFLAGS -o $OUT $@ $ELKSLIBC/libc.lib
+echo owcc $LDFLAGS -o $OUT $@
+owcc $LDFLAGS -o $OUT $@
 
 # convert to ELKS a.out format
 #os2toelks -f elks -o $OUT $OUT.os2
--- a/elks/tools/objtools/ewlink
+++ b/elks/tools/objtools/ewlink
@@ -24,6 +24,12 @@ fi
 
 ELKSLIBC=$TOPDIR/libc
 
+# wlink options for debugging:
+# -Wl,option -Wl,map              # produce mapfile (or use -fm=file.map)
+# -Wl,option -Wl,verbose          # more detail in mapfile
+# -Wl,option -Wl,static           # show static variables in mapfile
+# -Wl,option -Wl,eliminate        # enable dead code elimination
+
 LDFLAGS="\
     -bos2                           \
     -s                              \
--- a/libc/watcom.inc
+++ b/libc/watcom.inc
@@ -19,12 +19,14 @@ CARCH =\
     -fno-stack-check                \
     -Wc,-fpi87                      \
     -Wc,-zev                        \
+    -Wc,-zls                        \
     -Wc,-x                          \
     -Wc,-wcd=303                    \
     -fnostdlib                      \
 
 AARCH=\
     -m$(MODEL)                      \
+    -zq                             \
     -0                              \
     -fp0                            \
     -fpi87                          \
@@ -35,7 +37,7 @@ AS=wasm
 
 CFLAGS=$(CARCH) $(INCLUDES) $(DEFINES) -Wall -Os
 ASFLAGS=$(AARCH)
-ARFLAGS_SUB=-n -b -fo
+ARFLAGS_SUB=-n -q -b -fo
 
 %.obj: %.c
 	$(CC) -c $(CFLAGS) -o $*.obj $<
--- a/libc/watcom/asm/fchop87.asm
+++ b/libc/watcom/asm/fchop87.asm
@@ -44,7 +44,7 @@
 include mdef.inc
 include math87.inc
 
-        xrefp           __8087  ; indicate that NDP instructions are present
+;       xrefp           __8087  ; indicate that NDP instructions are present
 
         modstart fchop87
 
--- a/libc/watcom/asm/segments.asm
+++ b/libc/watcom/asm/segments.asm
@@ -49,15 +49,10 @@ BEGTEXT  segment word public 'CODE'
         assume  cs:BEGTEXT
         int     3
         int     3
-        public _start
-_start:
-if _MODEL and _BIG_CODE
-        extrn _start_crt0:far
-        jmpf _start_crt0
-else
-        extrn _start_crt0:near
-        jmp _start_crt0
-endif
+_start  label   byte
+        public  _start
+        xrefp   _start_crt0
+        dojmp   _start_crt0
         assume  cs:nothing
 BEGTEXT  ends
 
@@ -69,10 +64,10 @@ FAR_DATA ends
         assume  ds:DGROUP
 
 _NULL   segment para public 'BEGDATA'
-__nullarea label word
+;       public  __nullarea
+;__nullarea label word
         db      'nul'
         db      0
-        public  __nullarea
 _NULL   ends
 
 _AFTERNULL segment word public 'BEGDATA'
@@ -119,7 +114,7 @@ YIE     ends
 
 _BSS    segment word public 'BSS'
         ;extrn   _edata                  : byte  ; end of DATA (start of BSS)
-        ;extrn   _end                    : byte  ; end of BSS (start of STACK)
+        ;extrn   _end                    : byte  ; end of BSS  (start of HEAP/STACK)
 _BSS    ends
 
 ;STACK_SIZE      equ     1000h
--- a/libc/watcom/asm/segments.asm
+++ b/libc/watcom/asm/segments.asm
@@ -37,6 +37,9 @@ include mdef.inc
 
 DGROUP group _NULL,_AFTERNULL,CONST,STRINGS,_DATA,DATA,XIB,XI,XIE,YIB,YI,YIE,_BSS,STACK
 
+if (_MODEL and _BIG_CODE) eq 0
+
+; BEGTEXT section used in near code (small and compact) models only
 ; this guarantees that no function pointer will equal NULL
 ; (WLINK will keep segment 'BEGTEXT' in front)
 ; This segment must be at least 2 bytes in size to avoid confusing the
@@ -55,6 +58,8 @@ _start  label   byte
         dojmp   _start_crt0
         assume  cs:nothing
 BEGTEXT  ends
+
+endif
 
 _TEXT   segment word public 'CODE'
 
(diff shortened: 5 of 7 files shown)


jmalak:
I was thinking about strategy that I agree that **`-zls`** is not needed if appropriate symbols will be put into **`segments.asm`** temporarily and when you implement feature then it will be passed to appropriate code.
Due to target is 16-bit there is no issue with symbols for wide characters ( starting with **`_w`** or **`__w`** ).
Because libc is for OpenWatcom then you need not use **`_start`** symbol and you can use OpenWatcom default **`_cstart_`** symbol there is no relation to gcc.
Don't use BEGINTEXT for anything because it is not always used only for BIG code. entry symbol must be in standard code segment.
"change ewlink to include libc using -Wl,library option" be careful if you will use libc for more memory models then you need somehow change it. OpenWatcom put this information about libc model into modules therefore it is not necesary define it only LIBPATH must be defined.
Try to use directive libpath for libc location and extra define libc.lib because you can simply abandone it in the future or if you will somehow handle it.

jmalak:
"removes __begtext kluge used to force BEGINTEXT to be included in link"
It is not possible if you want to check for NULL code pointer because it must be on begining of code and __begtext must be at offset 2.

ghaerr:
> I agree that -zls is not needed if appropriate symbols will be put into segments.asm temporarily and when you implement feature then it will be passed to appropriate code.

After having looked over your original PR and thinking more about it, I think it better to go ahead and not have OWC emit special symbols, just like you thought originally. The reasoning for this is: 1) can remove unneeded `_argc`, `__8087` and `_cstart_` and `__begtext`, which saves space, 2) other developers may find other special symbols being emitted which break ELKS LIBC when link fails, and 3) no need to have special `__8087` handling now when it isn't being used, as it can be added back later after hard/soft FP figured out.

> Due to target is 16-bit there is no issue with symbols for wide characters ( starting with _w or __w ).

Does OWC produce special symbols when wide characters used? I am unsure if you're saying the -zls option affects this.

> Don't use BEGINTEXT for anything because it is not always used only for BIG code. 
> __begtext must be at offset 2.

I'm not following you here. The only guarantee is that the special BEGINTEXT segment will be the first class 'CODE' segment linked in the 'CODE' class, as documented in the DOSSEG section of the wlink manual. Thus (for near code models only, as far code pointers can't be easily checked), having the first two bytes of BEGINTEXT (set to `int 3/int 3`) guarantees that no near function pointer will have an address of 0 or 1. It does not matter what `__begtext` value is, it is merely a symbol in the BEGINTEXT segment in the older kluge code, which guarantees that the BEGINTEXT segment is linked in. The revised code sets the `_start` symbol in BEGINTEXT, so that guarantees BEGINTEXT is linked in and the first C function seen by the linker will never be assigned an address of zero or one.

> entry symbol must be in standard code segment.

The OS/2 NE executable format specifies both CS and IP for the entry symbol - it can be in any 'CODE' class segment. I have tested this PR code and it works in both near and far code models.

> be careful if you will use libc for more memory models

Thanks - I haven't yet figured out how to automatically support all memory models but will likely produce variously-named libc-{model}.lib files that are automatically used in `ewlink`

> Try to use directive libpath for libc location

Is that `-Wl,libpath=` for wlink? That could come in handy, but still have to specify a separate libc.lib file for each model, so `-Wl,library=` will probably be setup as required in ewlink (is what I'm thinking).

jmalak:
I have one idea related to OW C compiler and target definition and definition of include path.
We could use one of OW C compiler feature which little bit simplify cross-compilation to be safer.
We will introduce new OW target **`elks`** and change configuration following way
**`-bt=none`** => **-bt=elks**
add environment variable **`INCLUDE=${WATCOM}/lh`** and **`ELKS_INCLUDE=${WATCOM}/h`**
then you need not handle OW include path anymore.
if you will be using build for **`Linux`** you specify **`-bt=linux`** and OW C compiler will use **`INCLUDE`** variable
if you will be using build for **`ELKS`** you specify **`-bt=elks`** and OW C compiler will use **`ELKS_INCLUDE`** and **`INCLUDE`** variables
It will switch system include path in dependency on target that OW include path need not be handled by command line -I option and is handled by -bt option

ghaerr:
> I have one idea related to OW C compiler and target definition and definition of include path.

Let us think more deeply about that before going ahead with ELKS changes to OW master repo.

To be honest, I don't see good enough reason to add minimal partial support for ELKS (and then possibly confusing older users that have had binary installations from elsewhere installed) to require updated OW binaries or rebuild from source for such a small change. This also confuses things when OW output is OS/2 NE format - ELKS hasn't even released a new version supporting the updated (and currently optional) OS/2 loader, etc. I would say that making OW master repo changes only makes sense if effort were expended to move system call library over to OW repo. But it makes little sense to do that if the ELKS C library is not also automatically built by OW. Too many issues for the moment, I would not want OW master to be required change for developers this early in the game for getting OW for ELKS fully working. We are still in the process of hearing back from ELKS developers about this pre-release and changes that might be useful. Also multi-lib build is not figured out yet, etc.

jmalak:
> > I agree that -zls is not needed if appropriate symbols will be put into segments.asm temporarily and when you implement feature then it will be passed to appropriate code.
> 
> After having looked over your original PR and thinking more about it, I think it better to go ahead and not have OWC emit special symbols, just like you thought originally. The reasoning for this is: 1) can remove unneeded `_argc`, `__8087` and `_cstart_` and `__begtext`, which saves space, 2) other developers may find other special symbols being emitted which break ELKS LIBC when link fails, and 3) no need to have special `__8087` handling now when it isn't being used, as it can be added back later after hard/soft FP figured out.

It is not critical you can change it how it will be better for your strategy any time.

> 
> > Due to target is 16-bit there is no issue with symbols for wide characters ( starting with _w or __w ).
> 
> Does OWC produce special symbols when wide characters used? I am unsure if you're saying the -zls option affects this.
> 

Yes. OW C compiler detect main function if it started by w character then it is wide characters version and then define appropriate symbols. It is used for Windows, but works generaly.

> > Don't use BEGINTEXT for anything because it is not always used only for BIG code.
> > __begtext must be at offset 2.
> 
> I'm not following you here. The only guarantee is that the special BEGINTEXT segment will be the first class 'CODE' segment linked in the 'CODE' class, as documented in the DOSSEG section of the wlink manual. Thus (for near code models only, as far code pointers can't be easily checked), having the first two bytes of BEGINTEXT (set to `int 3/int 3`) guarantees that no near function pointer will have an address of 0 or 1. It does not matter what `__begtext` value is, it is merely a symbol in the BEGINTEXT segment in the older kluge code, which guarantees that the BEGINTEXT segment is linked in. The revised code sets the `_start` symbol in BEGINTEXT, so that guarantees BEGINTEXT is linked in and the first C function seen by the linker will never be assigned an address of zero or one.

I was not exact it must be at offset >= 2
BEGTEXT is specific for OW linker that it is not compatible with any other linker.

> 
> > entry symbol must be in standard code segment.
> 
> The OS/2 NE executable format specifies both CS and IP for the entry symbol - it can be in any 'CODE' class segment. I have tested this PR code and it works in both near and far code models.

Standard code segments I mean any segment of class 'CODE' except BEGTEXT.

> 
> > be careful if you will use libc for more memory models
> 
> Thanks - I haven't yet figured out how to automatically support all memory models but will likely produce variously-named libc-{model}.lib files that are automatically used in `ewlink`
> 
> > Try to use directive libpath for libc location
> 
> Is that `-Wl,libpath=` for wlink? That could come in handy, but still have to specify a separate libc.lib file for each model, so `-Wl,library=` will probably be setup as required in ewlink (is what I'm thinking).

Yes it is about wlink, you need still specify library= and can be changed but libpath= can be fixed.
You can also specify library in C source by #pragma library then it is included by linker automatically.

jmalak:
> > I have one idea related to OW C compiler and target definition and definition of include path.
> 
> Let us think more deeply about that before going ahead with ELKS changes to OW master repo.
> 
> To be honest, I don't see good enough reason to add minimal partial support for ELKS (and then possibly confusing older users that have had binary installations from elsewhere installed) to require updated OW binaries or rebuild from source for such a small change. This also confuses things when OW output is OS/2 NE format - ELKS hasn't even released a new version supporting the updated (and currently optional) OS/2 loader, etc. I would say that making OW master repo changes only makes sense if effort were expended to move system call library over to OW repo. But it makes little sense to do that if the ELKS C library is not also automatically built by OW. Too many issues for the moment, I would not want OW master to be required change for developers this early in the game for getting OW for ELKS fully working. We are still in the process of hearing back from ELKS developers about this pre-release and changes that might be useful. Also multi-lib build is not figured out yet, etc.

It has no impact to binaries it is only configuration of building that is more simplified for cross-compilation.
It is standard OW C compiler feature.

ghaerr:
Thank you for your explanations! :)

> I was not exact it must be at offset >= 2

I plan on committing this, but remain confused about your explanations for the following:

 `__begtext` can be at any offset (including 0 or 1), the former code required it to be anywhere in BEGTEXT, only for the purpose that the `dw __begtext` referenced by `_cstart_` (generated automatically by OWC) drags in the BEGTEXT segment, which drags in the NULL near function pointer allocation of two bytes prohibiting functions being assigned a zero address. It doesn't matter whether BEGTEXT is linked in or not, when wlink sees first C function definition it will assign it address zero if something like this is not done. None of that matters with this PR, as `__begtext` is removed and first assignable symbol in code segment will never be 0 or 1. Only BEGTEXT, not `__begtext` is treated specially by wlink.

[EDIT: are you saying that dead code elimination is handled differently than explanation? I don't know exactly how OW dead code elimination works, but it is not eliminating the `jmp` instruction in BEGINTEXT with this PR.]

> Standard code segments I mean any segment of class 'CODE' except BEGTEXT.

The current PR allows for producing near and far code models and the CS:IP is properly set in the NE executable header, with CS being BEGINTEXT in both cases. This has also been tested on `basic` in small and large models, and everything works. So what is the issue?

jmalak:
In OW BEGINTEXT is used only by big code models because purpose of this segment is check for access NULL pointer, of cause linker link it if it is exists without any restriction. Only note it must be segment aligned (16).
OW dead code elimination run on nonblank segments which are not referenced.

ghaerr:
> In OW BEGINTEXT is used only in big code models

I have tested compiling and linking basic in both small and large models: the BEGTEXT segment is included in both models. Here is map file output for small model:
```
Segment                Class          Group          Address         Size
=======                =====          =====          =======         ====

BEGTEXT                CODE           AUTO           0001:0000       00000005
_TEXT                  CODE           AUTO           0001:0006       00004630
FAR_DATA               FAR_DATA       AUTO           0002:0000       00000000
_NULL                  BEGDATA        DGROUP         0002:0000       00000004
_AFTERNULL             BEGDATA        DGROUP         0002:0004       00000000
...
```
Note that BEGTEXT is assigned the first NE segment number and of course starts at offset 0. Size is 5 corresponding to two int3 instructions followed by 3-byte jump to `_start_crt0`.

Here is large model:
```
Segment                Class          Group          Address         Size
=======                =====          =====          =======         ====

BEGTEXT                CODE           AUTO           0001:0000       00000007
basic_TEXT             CODE           AUTO           0001:0007       00002d78
host_TEXT              CODE           AUTO           0001:2d7f       00000783
host-stubs_TEXT        CODE           AUTO           0001:3502       0000001c
segments_TEXT          CODE           AUTO           0001:351e       00000000
_TEXT                  CODE           AUTO           0001:351e       00000133
strlen-c_TEXT          CODE           AUTO           0001:3651       00000013
...
```
In large model, BEGTEXT is the same except its length is 7, corresponding to the 5-byte far jump to `_start_crt0`.

In both large and small models, the entry point is 0001:0002.

> OW dead code elimination run on nonblank segments which are not referenced.

Ok, that makes sense. So either a segment is included or the whole segment is not included. I was thinking perhaps something fancier with code being partially eliminated within a segment. I suppose this feature is very useful when compiling functions into their own segments, allowing for library code to become smaller while in same C source file.

jmalak:
For testing add following wlink options to ewlink
```
    -Wl,option -Wl,map              \
    -Wl,option -Wl,eliminate        \
    -Wl,option -Wl,static           \
    -Wl,option -Wl,verbose          \
```

It list all static symbols and more details about modules
and use dead code elimination

ghaerr:
> In OW BEGINTEXT is used only by big code models because purpose of this segment is check for access NULL pointer,

Ok, I misunderstood what you were talking about with BEGTEXT - looking at the original OW sources, BEGTEXT is only used in *near code* models (not big code) for purpose of not allocating a NULL function pointer. I pushed a fix adding BIGTEXT only `if ( _MODEL and ( _TINY or _BIG_CODE )) eq 0`, so there is no extra segment being declared unless small or compact model.

> For testing add following wlink options to ewlink

Thanks, these produce a better mapfile. I never saw any change with enabling dead code elimination, but only tested with linking `basic`. Does -Wl,eliminate just turn on eliminating nonblank, non referenced segments, or something more?

Thank you for all your help and comments on this @jmalak!

jmalak:
@ghaerr It only remove unreferenced nonblank segment.
For real use it requires Compiler help.
You need emit each function in unique segment (compiler -zm option) and then it works as you expect.
I understand that it is not simple for you because I give you lot of new info about OW.
It takes time to understand the context, I've been through it myself so I know what it's about. In any case, nothing is unnecessary in OW and everything has its meaning, only one cannot see it at first glance.

PR #1949, opened 2024-08-01
remove nonsense, author used incorrect OpenWatcom configuration

jmalak:
Problem was that author used OS/2 headers instead of Linux headers for building Linux executable. 
OW define O_BINARY for Linux target as 0 and for OS/2 target as 0x200


ghaerr:
Thanks @jmalak!

PR #1960, opened 2024-08-09
actions: add OW libc test build on Ubuntu and OSX

jmalak:
Add automation for ELKS LIBC OW build on 64-bit Ubuntu and OSX
automaticaly install Open Watcom compiler and configure it for host system
next build ELKS CLIB and build ELKS sample

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -4,6 +4,7 @@ on:
   push:
     paths:
       - '.github/workflows/cross.yml'
+      - '!.github/workflows/ow-libc.yml'
       - 'tools/*'
 
 jobs:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,11 +5,13 @@ on:
     paths:
       - '**'
       - '!.github/workflows/cross.yml'
+      - '!.github/workflows/ow-libc.yml'
       - '!tools/*'
   pull_request:
     paths:
       - '**'
       - '!.github/workflows/cross.yml'
+      - '!.github/workflows/ow-libc.yml'
       - '!tools/*'
 
 jobs:
--- a/.github/workflows/ow-libc.yml
+++ b/.github/workflows/ow-libc.yml
@@ -31,10 +31,10 @@ jobs:
         include:
         - runner: ubuntu-latest
           arch: x64
-        - runner: macos-13
-          arch: x64
-        - runner: macos-14
-          arch: arm64
+#        - runner: macos-13
+#          arch: x64
+#        - runner: macos-14
+#          arch: arm64
 
     steps:
       - name: checkout ELKS
--- a/.github/workflows/ow-libc.yml
+++ b/.github/workflows/ow-libc.yml
@@ -1,0 +1,60 @@
+name: ow-libc
+
+on:
+  push:
+    paths:
+      - 'elks/include/**'
+      - 'elks/tools/objtools/**'
+      - 'elkscmd/basic/**'
+      - 'include/**'
+      - 'libc/**'
+      - '.github/workflows/ow-libc.yml'
+  pull_request:
+    paths:
+      - 'elks/include/**'
+      - 'elks/tools/objtools/**'
+      - 'elkscmd/basic/**'
+      - 'include/**'
+      - 'libc/**'
+      - '.github/workflows/ow-libc.yml'
+env: 
+  TOPDIR: ${{ github.workspace }}
+  TOOLDIR: ${{ github.workspace }}/elks/tools/objtools
+  SAMPLEDIR: ${{ github.workspace }}/elkscmd/rootfs_template/root
+
+jobs:
+  build:
+    name: OW LIBC build
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        include:
+        - runner: ubuntu-latest
+          arch: x64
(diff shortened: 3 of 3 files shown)


ghaerr:
This looks pretty cool, but could have issues if it always runs with no means to turn it off.

Is there a way to use caching to get the OW toolchain installed, this automated process adds almost  50% to the time it takes (3m + ~1.5m) for each commit pushed for a PR. Currently, 3m is on the high end and it ends up running twice. The gcc toolchain is built every 30 days or so, and takes 30 minutes, but otherwise is very quick. We need a way that the OW toolchain can be installed that doesn't take 1.5 minutes.

I'm not well-versed on CI, it looks like perhaps these runs only occur if files have changed within the listed paths (mostly libc and include files)? Perhaps 'include/**' isn't needed, as the only file in that directory (autoconf.h) changes on every configuration change, which shouldn't needed. Also perhaps running both macOS builds aren't needed - this isn't being done for the CI ELKS build, only Linux.

I see that when the sample basic.os2 is built, it's copied to the root directory of the image being created. This will be way too large for the 360k floppies being built, as there's no space, and applications are normally added to images using the elkscmd/Applications file. Have you determined whether copying this test file will break the "make images" build which builds all floppy/HD sizes?

jmalak:
I think this is not a problem because the rebuild will only happen if some LIBC source file changes. The OW build is pulled internally from Github, which is not a problem. Of course you can implement a cache, but I don't think it's critical, cache makes sense for frequent rebuilds. You can restrict the dependency (build trigger) in the "on push path" section. Now rebuild OW LIBC is started if any file in following trees are changed
```
      - 'elks/include/**'
      - 'elks/tools/objtools/**'
      - 'elkscmd/basic/**'
      - 'include/**'
      - 'libc/**'
      - '.github/workflows/ow-libc.yml'
```
You can remove some tree if it is not important or replace dependency on tree by explicit files etc.

I did build for 3 platforms but you can limit it to only one by commenting out any from definition
```
    strategy:
      matrix:
        include:
        - runner: ubuntu-latest
          arch: x64
        - runner: macos-13
          arch: x64
        - runner: macos-14
          arch: arm64
```

You can change what you need, you will see that it is not rebuild frequently only if relevant change appears.
I also excluded OW LIBC build yml file from main.yml and cross.yml file dependency that change of this file don't trigger main.yml or cross.yml workflows. It is done by exclusion
      - '!.github/workflows/ow-libc.yml'

You can take this PR as template for your implementation and modify it to follow your requirements.

In case you want to use cache then you will need move the automatic installation and setup of OW to cache creation and add OW configuration of WATCOM and PATH after cache initialization, but take into account that downloading cache can take some time that gain can be smaller then you expect.
Downloading of OW is about 140 Mbytes and unzip it and untar it. Full size (unziped) is about 500Mbytes. I don't know what will be quicker and how much. OW is build on any change and create fixed image each month on first build. It is not same as for gcc that you cannot predict when new version is available.

Anyway I don't know if build of sample executable is OK. I did it based on some notes in git repository, that remove or change what is nonsense.

ghaerr:
Ok @jmalak, thank you for this enhancement. It will be nice to know that OW C library build is checked when libc source or other included files or scripts change. 

We will give it a try. I'm not sure yet about copy of basic.os2 to rootdir, if that is a problem we can comment that out.

tyama501:
Hello @ghaerr 

Is that basic.os2 also copied to 1232 pc-98 build?
I don't think that graphic functions work for pc-98 since it is written in gcc assembler.

ghaerr:
> Is that basic.os2 also copied to 1232 pc-98 build?

I don't think so but I'm not actually sure. The contributed OpenWatcom CI build run is primarily for checking that the ELKS C library and tool scripts still compile and link when any changes are made to the C library or include files. 

The problem with the basic.os2 copy is that I'm not entirely sure exactly which ELKS config file is in use for that build (likely the default which is IBM PC but not exactly ibmpc-1440.config), and thus isn't being built for the PC-98 or 8018X architectures. The other problem is that OpenWatcom-compiled basic does not include a math library (yet), so all math function are missing, and it requires hardware 8087 floating point in order to run.

In addition, its inclusion to an image isn't controlled by the standard elkscmd/Applications file. There isn't any real reason to produce a basic.os2 except for compile & link testing so we'll probably remove the copy from the CI to lessen any confusion.

> I don't think that graphic functions work for pc-98 since it is written in gcc assembler.

That's right, OpenWatcom definitely won't compile the gcc asm directives, it has its own native mechanism which is different.

jmalak:
Yes, take it as starting point. You can simply enhance it by "strategy matrix" add configuration parameters where you can add by example memory model for build etc. that you can simply test multiple build with different configuration. If you use inline assembly then you need to do it conditionally for gcc and OW due to different syntax. it is not hard because OW #pragma aux is used for anything what is related to in-line assembly or symbols properties. If you will have any question with regard #pragma aux then simply create issue in OW GitHub repository and I will help you if I can.

jmalak:
I checked how long it takes and it is about 11 seconds to download and install OW for testing. It confirm my idea that it is not big issue that caching has no much sense for OW. It simplify OW installation steps for you by simple command only. This command can also configure used OW versions or some exact OW build, now it works with latest OW build and OW 2.0 64-bit version.
You can check also that your last changes in ELKS kernel doesn't triggered OW LIBC build that it works OK.

PR #1975, opened 2024-08-22
[watcom] Reinstate auto CI test build of libc.lib

ghaerr:
Runs OpenWatcom CI build without `fixomf` for weak symbol markings, if not present.


toncho11:
Comment for the https://github.com/ghaerr/elks/wiki/Using-OpenWatcom-C-with-ELKS

When you say "Build ELKS" it is in fact only building the libc, not performing a distribution build with images? It is not clear in my opinion. Maybe there can be a line saying that in order to compile applications with OW, the libc needs to be rebuilt with OC and this is what "Build ELKS" does and not an entire application and image build.

ghaerr:
> When you say "Build ELKS" it is in fact only building the libc, not performing a distribution build with images? 

Yes, that's correct.

> It is not clear in my opinion. 

Agreed, thanks. I updated the Wiki page changing "Build ELKS" to "Build the ELKS C Library" or similar in a couple places.

toncho11:
I made a further clarification:

> In order to build programs using Watcom C you need to:
> * first either install OpenWatcom C v2 or binary snapshot, both you can download from https://github.com/open-watcom/open-watcom-v2/releases.
> Next possibility is clone the repo from https://github.com/open-watcom/open-watcom-v2 and build OpenWatcom from this repo.
> * build an OpenWatcom version of the ELKS C library

And I added:

> Building an example program using OpenWatcom

Is wmake (Watcom make) supported for compilation? Is there any advantage using it?

PR #1977, opened 2024-08-26
[bioshd] Implement park for ancient MFM/RLL disks

vkoskiv:
This patch still requires work before merging IMO, but I wanted to file this PR now to gather some feedback.
I'm running two ST-412 hard disks on my IBM 5150+5161 system so I don't ever have to worry about storage again. I noticed that there doesn't seem to be any support in ELKS for parking hard disks of this vintage, so I took a stab at adding it.
There was some [discussion about this in 2020](https://github.com/ghaerr/elks/issues/437#issuecomment-598273782), but I assume most people playing around with ELKS aren't running disks this old :]

It seems to me that the correct way to do this is to park the drive in `bioshd_release()`.
This does work for unmounting non-root disks, but there is a special case for the root drive, where the drive is remounted as read-only instead, and `bioshd_release` never gets called.
So my hack to ensure all drives are parked when shutting down is to just park every drive in `sys_reboot()`, which feels very hacky.

I'd welcome any advice on a better approach, and any nits/comments on my implementation, as I'm not too familiar with writing kernel code.

Tangentially related, I observed that the userland `reboot` and `shutdown` commands don't seem to unmount all mounted volumes. I have a patch that makes them iterate all mounted volumes to unmount them before a shutdown/reboot, but I'll leave that for another PR, and pending more discussion. Shouldn't the kernel sync/unmount mounted filesystems before shutting down? Or is that generally left to the userspace to handle?

### DIFF
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -78,7 +78,7 @@ static int fd_count = 0;                /* number of floppy disks */
 static int hd_count = 0;                /* number of hard disks */
 
 static int access_count[NUM_DRIVES];    /* device open count */
-static struct drive_infot drive_info[NUM_DRIVES];   /* operating drive info */
+struct drive_infot drive_info[NUM_DRIVES];   /* operating drive info */
 static struct drive_infot *cache_drive;
 struct drive_infot *last_drive;         /* set to last drivep-> used in read/write */
 extern struct drive_infot fd_types[];   /* BIOS floppy formats */
--- a/elks/include/linuxmt/biosparm.h
+++ b/elks/include/linuxmt/biosparm.h
@@ -87,5 +87,6 @@ void BFPROC bios_copy_ddpt(void);
 struct drive_infot;
 void BFPROC bios_switch_device98(int target, unsigned int device,
         struct drive_infot *drivep);
+void bios_disk_park_all(void);
 
 #endif
--- a/elks/kernel/sys.c
+++ b/elks/kernel/sys.c
@@ -4,6 +4,7 @@
  *  Copyright (C) 1991, 1992  Linus Torvalds
  */
 
+#include <linuxmt/biosparm.h>
 #include <linuxmt/config.h>
 #include <linuxmt/errno.h>
 #include <linuxmt/sched.h>
@@ -48,16 +49,25 @@ int sys_reboot(unsigned int magic, unsigned int magic_too, int flag)
 		C_A_D = flag;
 		return 0;
 	    case 0x0123:		/* reboot*/
+#ifdef CONFIG_BLK_DEV_BHD
+		bios_disk_park_all();
+#endif
 		hard_reset_now();
 		printk("Reboot failed\n");
 		/* fall through*/
 	    case 0x6789:		/* shutdown*/
 		sys_kill(1, SIGKILL);
 		sys_kill(-1, SIGKILL);
+#ifdef CONFIG_BLK_DEV_BHD
+		bios_disk_park_all();
+#endif
 		printk("System halted\n");
 		do_exit(0);
 		/* no return*/
 	    case 0xDEAD:		/* poweroff*/
+#ifdef CONFIG_BLK_DEV_BHD
+		bios_disk_park_all();
+#endif
 		apm_shutdown_now();
 		printk("APM shutdown failed\n");
 	}
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -222,6 +222,24 @@ int INITPROC bios_gethdinfo(struct drive_infot *drivep) {
     }
     return ndrives;
 }
+
+static void BFPROC bios_disk_park(struct drive_infot *drive)
+{
+    /* int 13 AH=0xC supports only 10 bits for cyl */
+    if (drive->cylinders > 1024 || drive->fdtype != -1) return;
+    BD_AX = 0x0C << 8;
+    BD_CX = ((drive->cylinders & 0xFF) << 8) | ((drive->cylinders & 0x300) >> 2) | 0x1; /* 0x1 = sector */
+    BD_DX = bios_drive_map[drive - drive_info];
+    call_bios(&bdt);
+}
+
+void bios_disk_park_all(void)
+{
+    int i;
+    for (i = 0; i < NUM_DRIVES; ++i)
+        bios_disk_park(&drive_info[i]);
+}
+
 #endif
 
 #ifdef CONFIG_BLK_DEV_BFD_HARD
--- a/elks/arch/i86/drivers/block/rd.c
+++ b/elks/arch/i86/drivers/block/rd.c
@@ -40,7 +40,7 @@ static struct {			/* ramdrive information*/
     int start;			/* starting memory segment*/
     int valid;			/* ramdisk data valid flag*/
     rd_sector_t size;		/* ramdisk size in 512 byte sectors*/
-} drive_info[MAX_DRIVES] = {
+} rd_info[MAX_DRIVES] = {
 #if CONFIG_RAMDISK_SEGMENT
     {0, 1, CONFIG_RAMDISK_SECTORS}
 #endif
@@ -76,13 +76,13 @@ static int rd_open(struct inode *inode, struct file *filp)
     debug("RD: open /dev/rd%d\n", target);
     if (!rd_initialised || target >= MAX_DRIVES
 #if CONFIG_RAMDISK_SEGMENT
-        || !drive_info[target].valid
+        || !rd_info[target].valid
 #endif
                                                 )
         return -ENXIO;
 
     ++access_count[target];
-    inode->i_size = (long)drive_info[target].size << 9;
+    inode->i_size = (long)rd_info[target].size << 9;
     return 0;
 }
 
@@ -115,9 +115,9 @@ static int dealloc(int target)
 {
     int i, j;
 
-    i = drive_info[target].start;
+    i = rd_info[target].start;
     debug("RD: dealloc target %d, index %d, size %d sectors\n",
-	   target, i, drive_info[target].size);
+	   target, i, rd_info[target].size);
     while (i != -1 && rd_segment[i].sectors != 0) {
 	j = i;
 	debug("RD: dealloc purging rd_segment[%d].seg = 0x%x, next index %d, size = %d\n",
@@ -132,8 +132,8 @@ static int dealloc(int target)
 	debug("RD: dealloc status rd_segment[%d].seg = 0x%x, next index %d, size = %d\n",
 	     j, rd_segment[j].seg, rd_segment[j].next, rd_segment[j].sectors);
     }
-    drive_info[target].valid = 0;
-    drive_info[target].size = 0;
+    rd_info[target].valid = 0;
+    rd_info[target].size = 0;
     return 0;
 }
 
@@ -150,10 +150,10 @@ static int rd_ioctl(register struct inode *inode, struct file *file,
     debug("RD: ioctl %d %s\n", target, (cmd ? "kill" : "make"));
     switch (cmd) {
     case RDCREATE:
-	if (drive_info[target].valid)
+	if (rd_info[target].valid)
 	    return -EBUSY;
 
-	drive_info[target].size = 0;
+	rd_info[target].size = 0;
 	k = -1;
 	for (i = 0; i <= (arg - 1) / ((ALLOC_SIZE / 1024) * PARA); i++) {
 		j = find_free_seg();	/* find free place in queue */
@@ -163,7 +163,7 @@ static int rd_ioctl(register struct inode *inode, struct file *file,
 		    return -ENOMEM;
 		}
 		if (i == 0)
-		    drive_info[target].start = j;
+		    rd_info[target].start = j;
 
 		if (i == (arg / ((ALLOC_SIZE / 1024) * PARA)))
 		    /* size in 16 byte pages = (arg % 64) * 64 */
@@ -185,7 +185,7 @@ static int rd_ioctl(register struct inode *inode, struct file *file,
 		/* recalculate size to reflect size in sectors, not pages */
 		size = size / (RD_SECTOR_SIZE / PARA);
 		rd_segment[j].sectors = size;
-		drive_info[target].size += rd_segment[j].sectors;
+		rd_info[target].size += rd_segment[j].sectors;
 		size = (long) rd_segment[j].sectors * RD_SECTOR_SIZE;
 		debug("RD: index %d set to %d sectors, %ld bytes\n",
 		       j, rd_segment[j].sectors, size);
@@ -198,14 +198,14 @@ static int rd_ioctl(register struct inode *inode, struct file *file,
 		    rd_segment[k].next = j;	/* set link to next index */
 		k = j;
 	}
-	drive_info[target].valid = 1;
+	rd_info[target].valid = 1;
 	debug("RD: ramdisk %d created sectors %d index %d bytes %ld\n",
-		target, drive_info[target].size, drive_info[target].start,
-		(long)drive_info[target].size * RD_SECTOR_SIZE);
+		target, rd_info[target].size, rd_info[target].start,
+		(long)rd_info[target].size * RD_SECTOR_SIZE);
 	return 0;
 
     case RDDESTROY:
-	if (drive_info[target].valid) {
+	if (rd_info[target].valid) {
 	    invalidate_inodes(inode->i_rdev);
 	    invalidate_buffers(inode->i_rdev);
 	    dealloc(target);
@@ -241,10 +241,10 @@ static void do_rd_request(void)
 	target = DEVICE_NR(req->rq_dev);
 	debug("RD: %s dev %d sector %d, ", req->rq_cmd == READ? "read": "write",
 		target, start);
-	if (drive_info[target].valid == 0 ||
-                start + req->rq_nr_sectors > drive_info[target].size) {
+	if (rd_info[target].valid == 0 ||
+                start + req->rq_nr_sectors > rd_info[target].size) {
 	    debug("rd%d: sector %d beyond max %d\n",
-		 target, start, drive_info[target].size);
+		 target, start, rd_info[target].size);
 	    end_request(0);
 	    continue;
 	}
@@ -252,7 +252,7 @@ static void do_rd_request(void)
         for (count = 0; count < req->rq_nr_sectors; count++) {
             /* find appropriate memory segment and sector offset*/
             offset = start;
-            index = drive_info[target].start;
+            index = rd_info[target].start;
             debug("index %d, ", index);
             while (offset > rd_segment[index].sectors) {
                 offset -= rd_segment[index].sectors;
@@ -293,7 +293,7 @@ void INITPROC rd_init(void)
 
 #if CONFIG_RAMDISK_SEGMENT
 	printk("rd: %dK ramdisk at %x:0000\n",
-	    drive_info[0].size >> 1, rd_segment[0].seg);
+	    rd_info[0].size >> 1, rd_segment[0].seg);
 
 #if (CONFIG_RAMDISK_SECTORS > 128)
 	/* build segment array for preloaded ramdisks > 64k*/
--- a/elks/include/linuxmt/genhd.h
+++ b/elks/include/linuxmt/genhd.h
@@ -78,6 +78,7 @@ struct gendisk
 
 extern struct drive_infot *last_drive;  /* set to last drivep-> used in read/write */
 extern unsigned char bios_drive_map[];  /* map drive to BIOS drivenum */
+extern struct drive_infot drive_info[];
 
 extern struct gendisk *gendisk_head;    /* linked list of disks */
 


ghaerr:
Hello @vkoskiv,

Thanks for the PR! Your code looks well-written; I'll comment separately on various kernel- and ELKS-related coding issues next to the source lines for better clarity.

> It seems to me that the correct way to do this is to park the drive in bioshd_release().

In general, yes, this is probably much better than calling specialized driver routines from other areas of the kernel. This is likely why the initial build is failing, as we build for IBM PC, PC-98 and embedded 8018x systems, each of which may use different configurations (or none) of he BIOS disk driver.

> Shouldn't the kernel sync/unmount mounted filesystems before shutting down? Or is that generally left to the userspace to handle?
> There was some https://github.com/ghaerr/elks/issues/437#issuecomment-598273782,

The comment from 2020 suggests `shutdown` unmounting disks and then parking, while `reboot` would just unmount without parking. This might be a good way forward, although there are times when a reboot is wanted without having an unmount, as that can cause further I/O which may not be desired, and they should be a way to do that. Perhaps there should be an option to either of those commands that would not unmount or park all disks, etc. We should discuss this more. Keeping the parking and/or multi-filesystem unmount controlled only through command line programs will give the most flexibility.

Agreed that there should be a way to unmount all drives, and then (optionally?) park the drives in the normal case. This could be done by enhancing the `umount` system call and passing a flag to park or not programmatically, which would then get passed to the driver release entry point.

> but there is a special case for the root drive, where the drive is remounted as read-only instead, and bioshd_release never gets called.

If the "park" flag is set for unmount, the remount of root as read-only could be skipped.

> I observed that the userland reboot and shutdown commands don't seem to unmount all mounted volumes. I have a patch that makes them iterate all mounted volumes to unmount them before a shutdown/reboot

Perhaps that should join this PR, but its certainly OK to keep them separate. We can use the `ustatfs` system call in order to enumerate the mounted volumes (see elkscmd/sys_utils/mount.c).

Thank you!

elks/arch/i86/drivers/block/bioshd.c
```diff
@@ -332,15 +332,42 @@ static int bioshd_open(struct inode *inode, struct file *filp)
...
+    asm (
+        "movb $0xC,%%ah\n\t"
+        "movw %0,%%cx\n\t"
+        "movb $0x0,%%dh\n\t"
+        "movb %1,%%dl\n\t"
+        "int  $0x13\n\t"
+        : /* No outputs */
+        : "r" (cyl_sec), "r" (bios_drive)
+        : "ax", "cx", "dx"
+    );
```
ghaerr: Let's use the existing `call_bios` function to execute an INT 13h, instead of adding ASM code within C. See elks/arch/i86/drivers/block/bios.c for examples. The call_bios function itself lives in elks/arch/i86/lib/bios13.S for reference.

elks/arch/i86/drivers/block/bioshd.c
```diff
@@ -332,15 +332,42 @@ static int bioshd_open(struct inode *inode, struct file *filp)
     return 0;
 }
 
+void bioshd_park(struct drive_infot *drivep) {
+    unsigned int cyls = drivep->cylinders;
+    if (cyls > 1024 || drivep->fdtype != -1) return; /* int 13 AH=0xC supports only 10 bits for cyl */
+    unsigned char bios_drive = bios_drive_map[drivep - drive_info];
+    unsigned short cyl_sec = ((cyls & 0xFF) << 8) | ((cyls & 0x300) >> 2) | 0x1; /* 0x1 = sector */
+    asm (
```
ghaerr: For kernel programming, we worry a lot about the stack space any given function uses (as well as chains of function sequences), so it should be as visible as possible. For this reason, we like to see all local variables declared separately at the start of a function, instead of declared when they are needed.

elks/arch/i86/drivers/block/bioshd.c
```diff
@@ -332,15 +332,42 @@ static int bioshd_open(struct inode *inode, struct file *filp)
...
+        "movb %1,%%dl\n\t"
+        "int  $0x13\n\t"
+        : /* No outputs */
+        : "r" (cyl_sec), "r" (bios_drive)
+        : "ax", "cx", "dx"
+    );
+}
+
+void bioshd_park_all(void) {
+    for (int i = 0; i < NUM_DRIVES; ++i)
```
ghaerr: Ditto above for `int i;`.

elks/arch/i86/drivers/block/bioshd.c
```diff
@@ -332,15 +332,42 @@ static int bioshd_open(struct inode *inode, struct file *filp)
...
+        bioshd_park(&drive_info[i]);
+}
+
 static void bioshd_release(struct inode *inode, struct file *filp)
 {
+    register struct drive_infot *drivep;
```
ghaerr: `register` declarations are no longer used - GCC ignores them completely, so go ahead and delete. There are still older sections of the ELKS kernel where register declarations are seen; these cases are from old code still not changed since BCC was the compiler, which definitely required judicious use of the register keyword.

elks/kernel/sys.c
```diff
@@ -24,6 +24,8 @@
 
 static int C_A_D = 1;
 
+void bioshd_park_all(void);
```
ghaerr: No forward declarations are allowed in .c files, all would need to go somewhere in the include/linuxmt or include/arch header files. In this case, the function will likely live as a static function within the BIOSHD driver.

The reason for the build failing is that the code below is calling a specific block driver function which isn't part of the configured build. We'll be better off passing a flag from a user mode application that can find its way to the standard block device interface functions, which will eliminate this problem.

ghaerr:
As I look through the kernel source a bit more, I don't yet see any easy way of passing a flag to the block driver release routine, and some thought needs to go into whether to change the parms of the `umount` syscall. Using an ioctl comes to mind, but that requires a file descriptor, something the umount command line program won't have. I'll continue to think about this a bit more.

ghaerr:
Another way this might be done would be to move the sync, unmount (all), sleep, and park to within the kernel sys_reboot function (calling park like you have designed it, actually). The user programs currently trying to do nearly the same thing include `reboot`, `shutdown` and `poweroff`, all barely different. These could be combined by hard linking to a single executable that calls sys_reboot which already has a flags parameter for the various slightly different functions. The `bioshd_park_all` call could be `#ifdef CONFIG_BLK_DEV_BHD` and thus only called when the BIOS HD driver is linked in and the park flag set. My apologies if you've already analyzed all this and have come to the same conclusion with this PR.

Some advantages of using sys_reboot are 1) we don't build up specialized flag-passing in the kernel for what will be an unused function on almost all systems, 2) the poweroff, reboot, and shutdown programs could be consolidated and save disk space for floppy systems, which is valuable, and 3) some of the current code could be reworked in the reboot function, as, for instance, there are no programs currently using the "change ctrl-alt-del function" flag of the reboot system call. Things would actually become simpler. I also don't think we need separate shutdown and poweroff programs.

Parking the disk(s) only during poweroff/shutdown is probably better than during driver release, because its not certain that the disk should be parked every time its unmounted, as they may be mounted again. What do you think, is it bad to park disks too much, or should this only be done before poweroff?

vkoskiv:
Thank you for the thorough review! I applied some of the suggested fixes, but I'm struggling with `call_bios()` for the time being. I seem to be getting random lockups and now a divide by zero panic on boot whenever the patch for that is applied, even though the corresponding code path isn't even being called.
When I had a `struct biosparms` as a local in `bioshd_park`, the system would just freeze on boot, or print a bunch of dots and capital `F` letters and then freeze. I had a hunch that it might be trashing the stack somehow, so I moved that to file scope as a static variable like it is in `bios.c`, and now it's throwing a divide by zero error instead where it was freezing before.
I'll attach the patch in case there is some obvious issue I just didn't spot, but I'm out of ideas for the time being.

Beyond that issue, I think the approach you suggested in the last comment there makes sense to me. While parking more than necessary doesn't really cause harm beyond the expected mechanical wear the drive encounters in normal use, it also doesn't serve much of a purpose. We only really care about parking when the disk is actually spinning down with the power removed, as the heads crash on the disk surface by design. I doubt any hardware configuration this old would have hot-swappable drives where parking when unmounting would be relevant, so just doing it right before shutting down makes the most sense to me.

Amusingly, Github doesn't support standard git patches as attachments, so here's the patch on pastebin instead: https://pastebin.com/y6ix8x5i

ghaerr:
> now a divide by zero panic on boot whenever the patch for that is applied, even though the corresponding code path isn't even being called.

That's very strange, but I happen to be in the middle of a somewhat nasty bug hunt for a suspected divide-by-zero problem over [here](https://github.com/Mellvik/TLVC/issues/76). I don't yet have a full answer, but I think our new divide-by-zero fault handler works OK. I'll keep you posted on that one.

> Github doesn't support standard git patches as attachments

It might be easier just to do another `git commit -a; git push` to push your changes in another commit, rather than using a force-push. That will allow us to see the changes easier, and we can always squash the merge commit later.

Looking at your change, I don't know why its happening. I am thinking that rather than declaring another `static struct biosparms bdt` in bioshd.c, the entire bioshd_park routine should be moved to bios.c. There's already a struct biosparms bdt over there, and frankly all the other call_bios calls are there too. Try that and see what happens.

> When I had a struct biosparms as a local in bioshd_park, the system would just freeze on boot, or print a bunch of dots and capital F letters and then freeze.

Yes, using any but a very few locals in any function is calling for a stack overflow, long story short the kernel can use upwards of 500+ bytes stack doing disk I/O by the time the fullest stack depth of functions is executed. 

The dots on boot are sector reads of the kernel /linux image by the boot sector, and the 'F' means a failed read for /bootopts on FAT, which is normal for MINIX images. The reason you don't see much else I think is very tricky - also being investigated in the above bug hunt link - it appears that QEMU runs serial and possibly console I/O asynchronously from the code being executed and a hard crash sometimes cuts output short. However, in this case it could just be that for some reason the kernel Image (/linux) is corrupted. I have seen this occur when the kernel is not fully rebuilt. Use `make kclean; make` to ensure it is. You're welcome to post a screenshot of the boot screen. After the 'F', setup.S normally displays tXXXX fXXXX dXXXX for the text, fartext and data segment values before jumping to the kernel itself, which should display "Direct Console..." on the next line.

vkoskiv:
> That's very strange, but I happen to be in the middle of a somewhat nasty bug hunt for a suspected divide-by-zero problem over [here](https://github.com/Mellvik/TLVC/issues/76). I don't yet have a full answer, but I think our new divide-by-zero fault handler works OK. I'll keep you posted on that one.

Very interesting analysis on that issue! Though I'm doing all my testing on my physical 5150, so it's unlikely that the issue is caused by a QEMU bug, at least in my case.

> The dots on boot are sector reads of the kernel /linux image by the boot sector, and the 'F' means a failed read for /bootopts on FAT, which is normal for MINIX images. The reason you don't see much else I think is very tricky - also being investigated in the above bug hunt link - it appears that QEMU runs serial and possibly console I/O asynchronously from the code being executed and a hard crash sometimes cuts output short. However, in this case it could just be that for some reason the kernel Image (/linux) is corrupted. I have seen this occur when the kernel is not fully rebuilt. Use `make kclean; make` to ensure it is. You're welcome to post a screenshot of the boot screen. After the 'F', setup.S normally displays tXXXX fXXXX dXXXX for the text, fartext and data segment values before jumping to the kernel itself, which should display "Direct Console..." on the next line.

I forgot to note that the dots and `F` letters issue happens quite far into kernel init, with a bunch of init output already on screen, then the cursor jumps up to roughly the center of the console output and overwrites a portion of that output with the dot/`F` mess. I guess something goes wrong, and it's jumping back to the initial bootstrap routine, perhaps? Again, this is on my physical 5150. I can grab a 'screenshot' today after work when I do more testing. The freezes/dots issue seemed to be triggered after the second hd*/fd* listing, but I'll confirm that with the screenshot later.

vkoskiv:
> Looking at your change, I don't know why its happening. I am thinking that rather than declaring another `static struct biosparms bdt` in bioshd.c, the entire bioshd_park routine should be moved to bios.c. There's already a struct biosparms bdt over there, and frankly all the other call_bios calls are there too. Try that and see what happens.

I moved the parking implementation to `bios.c` now, and the whole thing seems to work with `call_bios()`.

vkoskiv:
Also renamed to `bios_disk_park()` to better match the other functions in `biosparm.h`

ghaerr:
> I moved the parking implementation to bios.c now, and the whole thing seems to work with call_bios().

I must say, that's absolutely strange. It might be worth diving in to see why two separate static declarations of the same variable (if that's even the real issue) caused this to break, but I'm glad to hear it is now working. I'll take a note of the previous issue for another day.

Extern declarations aren't allowed in .c files, the `extern struct drive_infot drive_info[]` will have to move to biosparms.h, along with the forward decl for bios_disk_park.

Given all the complexity discussed as well as the strange unsolved problem with the multiple static declarations, it might be better to add park capability enhancements in multiple PRs. I'm inclined to go with your original thoughts of parking the drives only in the poweroff/shutdown portions of sys_reboot(). The direct call to bios_disk_park from there can be `#ifdef CONFIG_BLK_DEV_BHD`. You'll also want to remove the park from the bios_release entry point, as our discussion is tending towards the idea of only parking at power off, not when a disk is unmounted. (There are also cases where a drive may be opened and released quickly by various programs, we would not want parking in those cases either).

Let me know when you're closer to finalizing this portion, and I'll turn back on the CI workflow for you.

Thank you!

vkoskiv:
Addressed the forward declare in `sys.c`, it's in `biosparm.h` now like the rest of the `bios.c` stuff. I was wondering if I should mark this as `BFPROC` too, but I wasn't sure what the rationale for using that is. I'm assuming far calls are a bit slower, so seldom used functions are marked with that to save space in `.text`?

vkoskiv:
I'll do those changes you just mentioned too, and then ping you :]

ghaerr:
> I was wondering if I should mark this as BFPROC too,

Yes, nice catch. Mark it BFPROC. The reason is that we're very nearly completely out of .text space so groups of related functions are possibly placed in the single far section based on the configuration. Call speed is rarely an issue for disk-related functions, but FYI calling far from near is quick, but calling near from far requires thunking by the compiler to ensure the `__cdecl` passed arguments stay at their expected offsets.

vkoskiv:
> Extern declarations aren't allowed in .c files, the `extern struct drive_infot drive_info[]` will have to move to biosparms.h, along with the forward decl for bios_disk_park.

Moving it there would require us to include `genhd.h` in `biosparms.h` for the `struct drive_infot` definition. Is that okay to do?

> I'm inclined to go with your original thoughts of parking the drives only in the poweroff/shutdown portions of sys_reboot(). The direct call to bios_disk_park from there can be `#ifdef CONFIG_BLK_DEV_BHD`. You'll also want to remove the park from the bios_release entry point, as our discussion is tending towards the idea of only parking at power off, not when a disk is unmounted.

Agreed, having parking logic in multiple places isn't necessary, and adds unneeded complexity.

ghaerr:
> Moving it there would require us to include genhd.h in biosparms.h for the struct drive_infot definition. Is that okay to do?

No, we're trying to stay away from the complicatedness of having includes within includes. I'm not right looking at it now, perhaps the struct and extern definition should go in genhd.h instead then. If that can work and not be messy, including genhd.h in bios.c if it isn't already is fine.

In summary, its no problem to put the externs in the header file where they are declared, I just kind of jumped at the idea of having it in biosparms.h without much looking.


Sorry this is all more complicated than it should be - but we're getting there!

vkoskiv:
Removing parking from `bioshd.c` also allows us to only expose `bios_disk_park_all()`, and leave `bios_disk_park()` static.
Now, the only place that uses it is `sys_reboot()` in `sys.c`

New finding: marking `bios_disk_park_all()` as `BFPROC` caused a freeze and just now a divide by zero when invoking `shutdown`, but leaving that out seems to make it work.

> Sorry this is all more complicated than it should be - but we're getting there!

I've really enjoyed the thorough feedback and discussion, working on this PR has been quite enjoyable :]

vkoskiv:
Moving the extern `drive_info` to `genhd.h` seems to work, though I did need to rename the conflicting `drive_info` anonymous struct in the ramdisk implementation in `rd`. That is now called `rd_info`, which I think fits a bit better, perhaps. Testing that on hardware, will then push that change too.

vkoskiv:
One more note, earlier I forgot to call `bios_disk_park_all()` in the reset case in `sys_reboot()`, that is now fixed, so the disks get parked on reboot as well.
Still doing some more testing on my hardware, seems to be working quite well now!

vkoskiv:
Gated this thing behind `CONFIG_BLK_DEV_BHD` as well.

ghaerr:
> New finding: marking bios_disk_park_all() as BFPROC caused a freeze and just now a divide by zero when invoking shutdown, 

Wow, another strange finding. Believe it or not, we don't usually have too much of that anymore. (Except for my current ongoing bug hunt in https://github.com/Mellvik/TLVC/issues/76 which interestingly sometimes exhibits this behavior). It appears that when the CS segment gets mangled, this kind of thing happens on ELKS.

However, seeing new Divide by Zero faults could just be that we're now seeing what's actually happening, given that fault handler was very recently added to ELKS.

> but leaving that out seems to make it work.

That's ok for now, but make entirely sure that the reason wasn't that a forward declaration was not BFPROC and the other was, or vice versa. This is actually the reason why forward decls aren't allowed in .c files - too easy to have this error and not reported by the compiler. Needless to say, there should be no warnings produced during kernel compilation.

vkoskiv:
> That's ok for now, but make entirely sure that the reason wasn't that a forward declaration was not BFPROC and the other was, or vice versa. This is actually the reason why forward decls aren't allowed in .c files - too easy to have this error and not reported by the compiler. Needless to say, there should be no warnings produced during kernel compilation.

I changed them both back and forth a few times when testing. That is, I added/removed `BFPROC` for `bios_disk_park_all()` in both `bios.c` and `biosparm.h`

I left `BFPROC` there for the static `bios_disk_park()` in `bios.c`, as that didn't seem to have any adverse effect.

vkoskiv:
As for testing, I've only tested this on my 5150 system so far, and it is working as expected. Any other systems using the `bioshd` driver will now also perform this park, as long as the hard disk has <= 1024 cylinders. I don't foresee this causing any issues, it's just an extra seek for drives that feature built-in park. If there is a way to detect that a given drive requires parking with certainty, then that check could be added to `bioshd_disk_park()`

ghaerr:
> If there is a way to detect that a given drive requires parking with certainty, then that check could be added to bioshd_disk_park()

Yes, I was thinking of just that and what how that might affect other users with unnecessary head movements. It might be worth some research... is there a min or maximum disk size typical of these very old disks? Perhaps only do it for very small hard drives? It's probably not worth writing a bunch of code to query the disk interface, unless it would be simple. Unfortunately, all our new kernel features have to sit core-resident even when not used.

After commit, I plan on testing playing with the BFPROC issue on QEMU or another emulator.

vkoskiv:
I tried to find out if there is a commonly accepted way to detect this, but nothing really came up. As far as I know, all IDE drives autopark, as well as a few of the later MFM/RLL drives.
The implementation of this PR is now complete I think, pending more testing and feedback, of course. Let me know if you find anything. I'll focus on some other neat stuff I have planned in the meantime :]

elks/include/linuxmt/biosparm.h
```diff
@@ -87,5 +87,8 @@ void BFPROC bios_copy_ddpt(void);
 struct drive_infot;
 void BFPROC bios_switch_device98(int target, unsigned int device,
         struct drive_infot *drivep);
+#ifdef CONFIG_BLK_DEV_BHD
+void bios_disk_park_all(void);
+#endif
```
ghaerr: Let's get rid of ifdefs in header files to reduce build dependencies. There's no harm in leaving the forward declaration in, as the linker will always tell you a function is missing anyways.

vkoskiv: Done!

ghaerr:
Thank you @vkoskiv!

tyama501:
Hello @ghaerr and @vkoskiv 

Is this park also called from PC-98 shutdown?
PC-98 does not have int13 so I haven't checked yet but it may have problem.

Reboot,
I haven't supported reboot command for PC-98 yet so it failed previously but it didn't stuck.
I may support reboot by IO control but it would be nice if we can keep it without stuck until then.

Thank you.

ghaerr:
> Is this park also called from PC-98 shutdown?
PC-98 does not have int13 so I haven't checked yet but it may have problem.


Currently, yes if PC-98 defines CONFIG_DEV_BIOS_BHD, which I think it does. This will mean the defines in sys.c will need to change to:
```
#if defined(CONFIG_BLK_DEV_BHD) and !defined(CONFIG_BLK_DEV_BHD)
```
I will make this change, as otherwise the system may crash on reboot/shutdown/poweroff on PC-98.

> I haven't supported reboot command for PC-98 yet so it failed previously but it didn't stuck.

I plan on a simplification/rewrite of the reboot() system call code, and will take this into account, allowing for you to submit a PC98-compatble way of handling reboot.

Thank you!

tyama501:
Hello @ghaerr 

Does the shutdown for PC-98 still call the bios_disk_park_all?
The shutdown is doing sys call,
and sys_reboot in the kernel/sys.c is called right? 
(It seems that the PC-98 emulator does not crash when shutdown but
if it is still calling int13, I would like to add ifdef somewhere.) 

Thank you.

tyama501:
(Or it may be calling int1B with wrong paramaters by call_bios.)

ghaerr:
> Does the shutdown for PC-98 still call the bios_disk_park_all?

Yes, the disk park function is called during the sys_reboot syscall for all platforms that have the BIOS fd/hd driver configured. This performs an INT 13h function 0Ch disk seek function using `call_bios` to the last disk cylinder for IBM PC, and an INT 1Bh for PC-98, which is likely incorrect. I notice that that was coded using a literal 0x0C rather than a define in biosparm.h so it went unnoticed for PC-98 incompatibility.

Is there a BIOS track seek function for PC-98, or should the entire function be commented out on PC-98?

ghaerr:
> (Or it may be calling int1B with wrong paramaters by call_bios.)

Yes - it is calling INT 1Bh function 0Ch for PC-98, since it is using call_bios.

PR #1978, opened 2024-08-28
[cmds] Unmount all volumes in shutdown & reboot

vkoskiv:
Somewhat related to #1977, there was some [discussion of moving this functionality to the kernel](https://github.com/ghaerr/elks/pull/1977#issuecomment-2311504681), but here is the userspace implementation I have currently.

It now unmounts all mounted volumes, instead of only the root volume when you invoke `shutdown` or `reboot`

### DIFF
--- a/elkscmd/sys_utils/reboot.c
+++ b/elkscmd/sys_utils/reboot.c
@@ -15,16 +15,33 @@
 #include <time.h>
 #include <sys/mount.h>
 #include <sys/select.h>
+#include <linuxmt/limits.h>
+
+int try_unmount(dev_t dev)
+{
+	struct statfs statfs;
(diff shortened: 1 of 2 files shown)


ghaerr:
This looks pretty good, thank you. Nice feature that should already have been in place!

I think I'm willing to accept this as-is, although I'm mulling over the idea of moving some of this into the kernel or perhaps better into a single, combined .c file using argv[0] to differentiate between shutdown, poweroff and reboot. The reasoning is that you may notice that for each of the current binaries, they're exactly at 1024 bytes, with reboot unfortunately just larger. In ELKS, a disk block is 1K, so this ends up using 2 more disk blocks per floppy. Our distribution disks are so tight (even with executable compression) we have to worry about that kind of stuff around here lol.

Nonetheless, this is a very nice step in the right direction :)

Thank you!

PR #1980, opened 2024-09-02
[console] Add dual monitor support to console-direct driver

vkoskiv:
I wanted to use two displays at the same time on ELKS, so I rewrote most of the `console-direct` driver to make this happen.

Still very much a WIP, but this is already working quite nicely on my hardware.
IBM folks were clever enough to ensure that I/O ports and vram segments didn't collide between the MDA and CGA cards, so both could be used at once.
Not too many programs in the DOS era supported this. Lotus 123 and Borland's wonderful Turbo Debugger are good examples.
Now ELKS, with its full preemptive multitasking can take advantage of this feature, for some really cool dual-terminal UNIX sessions on even the most ancient IBM hardware.

This PR introduces the concept of an output to the console drivers, and makes use of that in `console-direct.c`, which now has logic for probing, initializing and managing multiple output devices.
Switching between displays is fairly straightforward - Each display gets allocated 0-N consoles, and the active display is indicated by hiding the cursor on the inactive display. Switch with `CTRL+ALT+F<N>`, like before.

Few more things require attention before merging:

- [ ] The disk busy indicator relies on the `VideoSeg` global, which is no more. I'll probably just hook that up so the indicator is displayed on the primary display. Thoughts?
- [ ] I have not tested, or even considered EGA + VGA. As far as I know, EGA should be pretty straightforward, and may even be able to coexist with both a CGA *and* and MDA card, but that needs to be checked. VGA is still //TODO, similarly.
- [ ] I have only tested this on a working MDA + CGA configuration on my 5150+5160 setup. More testing with different hardware configurations is needed.

Below are some photos of the thing in action on my hardware.
IMG_4216
IMG_4217
IMG_4218
IMG_4220


vkoskiv:
It should be noted that the addition of this new stuff to `console-direct.c` bloats up that driver by ~1.7kB, so it might be desirable to move this implementation into a separate driver, and have it as a compile-time option.

```
Before:
> du -b $(find . -name 'console-direct.o')
16504	./elks/arch/i86/drivers/char/console-direct.o
After:
> du -b $(find . -name 'console-direct.o')
18216	./elks/arch/i86/drivers/char/console-direct.o
```

ghaerr:
Wow @vkoskiv this is really amazing! Great work!!

Give me a bit to find some time to see how you've made this work, but I'm super impressed with the screenshots. 

And I have to say, running `matrix` on top of your IBM whilst in the middle of an `vi` editing session on another monitor?! That's really too much, a total classic!!!!  I think I'll hang that picture on my wall :)

I haven't had but two seconds to look at the code, but I'm seeing lots of C->op->var. The double indirection all the time is expensive. How about the idea of just including the `struct output` members directly in `struct console`, or something like that. I realize that design might be a bit different, given that you have identified Consoles and Outputs. If they can be merged into a single struct Console, even at the expense of using a bit more data because of a larger array, that would likely still be preferable to multiple structs and indirections.

The above suggestion would also help for adding a config option for this, as I think the extra 1.7k bytes is too much to force on people that don't have multiple monitors, and, in some configurations, we're within about that much space before running out of our near kernel CS address space! The extra data required in struct Console could be ifdefed out when not configured, along with any unneeded routines.

I'll get back to you shortly with more input, thank you for this amazing contribution!

vkoskiv:
> I haven't had but two seconds to look at the code, but I'm seeing lots of C->op->var. The double indirection all the time is expensive. 

Ah! I had a hunch that that may be the case as I was adding those in, I just didn't look at the generated code yet :]
Merging `struct output` with `struct console` should be quite doable. The main goal was to just get rid of that global state that assumed a single display, and I guess I just figured this might be how this abstraction would be expressed on modern Linux, but of course, modern Linux this is not. Bytes are expensive!

ghaerr:
> Merging struct output with struct console should be quite doable.

Great - please go ahead with that and push another version after you have that working. I should hopefully have more time soon.

If you want to take a pass at a configuration option, a very simple method is to define a file-local config option, say CONFIG_CONSOLE_DUAL, that is defined near the top of console-direct.c. Then you can play with adding (or not) the struct Console variables right during development, and we can then afterwards put the config option into the appropriate place inside the configuration management system.

Turning the config option on and off within the file allows for a simple comparison of text/data sizes. BTW, for that, I would use `ia16-elf-size console-direct.o`. That shows the exact text/data/bss values immediately, which helps.

ghaerr:
> I just didn't look at the generated code yet 

For that, try `ia16-elf-objdump -D -r -Mi8086 console-direct.o`. Its definitely worth getting used to, although a bit complicated at first.

ghaerr:
> a very simple method is to define a file-local config option, say CONFIG_CONSOLE_DUAL, that is defined near the top of console-direct.c.

Now that I see you've gone ahead and implemented this for many of our various console*.c drivers, a better place for the temp config option for the moment would be in console.h.

> The disk busy indicator relies on the VideoSeg global, which is no more. I'll probably just hook that up so the indicator is displayed on the primary display. 

Sure, you can do that provided it doesn't get messy, or just ifdef it out entirely when CONFIG_CONSOLE_DUAL is defined, your choice.

tyama501:
Wow!! This is amazing and very cool!

console.c is used in pc-98 too so let me know if there is build errors:)
Thank you.

toncho11:
Seeing the two apps simultaneously is priceless!
From what I heard the old console driver (current in ELKS) is a bit slow, so maybe this rewrite can also help with that in the future.

ghaerr:
> From what I heard the old console driver (current in ELKS) is a bit slow, so maybe this rewrite can also help with that in the future.

Yes, while the ELKS kernel overhead in processing tons of ANSI escape sequences emitted from `sl` has proved quite slow on an 8088, this enhancement actually might technically make the console just slightly slower. The speed issue isn't just within the ANSI console driver though. But with the double indirection removed, I don't think the proposed changes will make ELKS measurably slower :)

@vkoskiv, you're welcome to run `sl` on your vintage 8088 and report back just how fast/slow it seems. I would be surprised if your code made any noticeable difference.

ghaerr:
Hello @vkoskiv,

I've been thinking more about some ideas for the possible minor (re-)implementation of this PR. As mentioned above, we definitely don't want to use double indirection. And we want as much speed as possible, but certainly the "fixed" values of things like `Width`, `MaxCol` etc would need deferencing via `C->MaxCol` etc (BTW, perhaps don't rename all the member variables, just use the same name in the extended `struct Console` for less confusion with old code).

Since struct Output would be going away, we really only need as may struct Consoles as there are max consoles * monitors, which is 6, and possibly only 4 if MDA hardware doesn't support multiple pages? So we need /dev/tty1 through /dev/tty6, or through /dev/tty4 for a single MDA page. Each struct Console basically maps directly into a /dev/ttyN, just the ordering of the mapping may differ between hardware installed:

In the MDA-only case, /dev/tty1 (though N when more pages) would be allocated, so systems programs can always assume a working /dev/tty1. For the EGA-only case, /dev/tty1 through 3, just like it is now, and for the multiple monitor case, map EGA to /dev/tty1 - 3 and MDA tty4 onwards. This may be how you have it designed now.

For the `glock` global, it may not be needed as a struct Console member, as it's only used by Nano-X for switching between text and graphics modes which MDA doesn't support, so a single global value should suffice.

For EGA+CGA, the EGA consoles would be allocated first, then CGA. For VGA, ELKS treats VGA as EGA and doesn't support any VGA-specific features, so its probably not worth bothering to add options to ELKS to use special (non 0xB800) hardware address for EGA/VGA. I am assuming that your initialization routine can basically differentiate between hardware video memory at B000, B800 and A000 all in some kind of text mode. Are you relying on the BIOS to initialize each monitor's text mode? I haven't looked at that part yet.

I haven't fully studied your code, so some of what I've suggested may not quite be the way you've implemented things, but thought to discuss this more before/while you're refactoring the code.

Thank you!

vkoskiv:
Hey @ghaerr ! Took a few days longer than expected to get back to this, as I fell ill with the flu the very next day after opening this PR. Feeling much better now, so I'm excited to get back to working on this!
I finished up and pushed the patch I was working on earlier. It does what it says on the tin, and gets rid of the `struct output` stuff, and instead includes the necessary bits in `struct console` directly.
That change, plus a few extra days of thinking time, enabled me to simplify the init logic in `console-direct.c` quite a bit. I still need to gate some of the extra stuff behind that `CONFIG_CONSOLE_DUAL` flag, but it's already looking quite good.

> (BTW, perhaps don't rename all the member variables, just use the same name in the extended `struct Console` for less confusion with old code).

Agreed, I'll change those back.

> Since struct Output would be going away, we really only need as may struct Consoles as there are max consoles * monitors, which is 6, and possibly only 4 if MDA hardware doesn't support multiple pages?

The max amount of consoles possible, as far as I can tell, is 5. We can't have multiple CGA cards in the system, as their CRTCs and vram conflict. CGA should support 4 console pages, but `console-direct` in `master` currently only sets up 3 pages, I'm not sure why. I didn't try bumping that up to 4 yet, as there are a lot of other things to test.
So the theoretical max is 5 (4x CGA + 1x MDA), but currently the max is 4 (3x CGA + 1x MDA).
I haven't investigated EGA yet, I don't even know if it can coexist with a CGA card, though I currently suspect that it may.
And yes, you are correct that the 4K of RAM in a MDA card only supports a maximum of 1 80x25 console pages (2k for text, 2k for attribs)


> So we need /dev/tty1 through /dev/tty6, or through /dev/tty4 for a single MDA page. Each struct Console basically maps directly into a /dev/ttyN, just the ordering of the mapping may differ between hardware installed:
> 
> In the MDA-only case, /dev/tty1 (though N when more pages) would be allocated, so systems programs can always assume a working /dev/tty1. For the EGA-only case, /dev/tty1 through 3, just like it is now, and for the multiple monitor case, map EGA to /dev/tty1 - 3 and MDA tty4 onwards. This may be how you have it designed now.

The way that it's designed now is that we ask the BIOS to tell us which display the user has selected as the 'primary' one. This happens with switches on the IBM 5150:

image

and I'm assuming BIOS options on more modern motherboards.

So `/dev/tty1` *always* ends up on that primary display, and the rest are allocated based on availability. So if we only have a CGA card, we get 3 TTYs. CGA + MDA allocates the extra `/dev/tty4` on the MDA display.
I still need to test this, but the logic should also work if I flip the switches to select my MDA card as the primary display, I'd get `/dev/tty1` on the MDA, and `/dev/tty[2,3,4]` on the CGA.
Do you think this makes sense?

> For the `glock` global, it may not be needed as a struct Console member, as it's only used by Nano-X for switching between text and graphics modes which MDA doesn't support, so a single global value should suffice.

Side note, I remember trying the Nano-X stuff earlier, and as far as I can tell, it's not intended to work with CGA or MDA? In any case, I propose a check in the `DCGET_GRAPH` ioctl to return an error code, at least if that console is assigned to a MDA screen that doesn't support graphics.

> I am assuming that your initialization routine can basically differentiate between hardware video memory at B000, B800 and A000 all in some kind of text mode.

Right, so I'm just dealing with the specified video memory addresses and CRTC I/O bases. `probe_crtc()` can check if a CRTC is present, and that info is then used when setting up hardware.


> Are you relying on the BIOS to initialize each monitor's text mode? I haven't looked at that part yet.

No, so the primary source of 'bloat' coming from these changes is that we now have logic for setting up video hardware, as we cannot rely on the BIOS to initialize anything but the primary one specified by the user. `init_output()` handles this. It programs the CRTC to a known good state for 80x25 text, tests and clears the VRAM, and  then enables video at the end. I checked this against what the BIOS does, and it's quite similar. I didn't investigate yet if there is a BIOS interrupt we could call to set up arbitrary adapters.

vkoskiv:
@tyama501 

> Wow!! This is amazing and very cool!

Thanks!

> console.c is used in pc-98 too so let me know if there is build errors:) Thank you.

I've taken care to update `console-direct-pc98.c` and checked that there are no build errors, but I don't have hardware to test that my changes didn't break something at runtime.

vkoskiv:
The bloat of `console-direct.o` was slightly reduced here:

```
Original:
> du -b $(find . -name 'console-direct.o')
16504	./elks/arch/i86/drivers/char/console-direct.o
With struct output:
> du -b $(find . -name 'console-direct.o')
18216	./elks/arch/i86/drivers/char/console-direct.o
Currently:
> du -b $(find . -name 'console-direct.o')
17752	./elks/arch/i86/drivers/char/console-direct.o
```

So 464 bytes smaller, but still 1248 bytes of bloat when compared to how it was originally. I'd expect gating with `CONFIG_CONSOLE_DUAL` will constrain most of this bloat to only when that option is enabled, though.

vkoskiv:
@ghaerr What should we do about the spinning I/O indicator that lives in `arch/i86/kernel/timer.c`?

It relies on the `VideoSeg` global, and it's a bit unclear now which video page on which display that should be. Intuitively I'd just assign the first page of the primary display there, but then the spinner only displays if that page is up on that display.
I think the spinner is cute, but I'd like to see it documented, and perhaps gated behind a config option, so it may be disabled if needed.

vkoskiv:
Cracked open the 5150 to test this with the MDA card selected as primary, and it works as expected. The assignment ends up as:
```
/dev/tty1 -> MDA
/dev/tty2 -> CGA
/dev/tty3 -> CGA
/dev/tty4 -> CGA
```

ghaerr:
> Agreed, I'll change those back.

Having the same member names in struct Console will help, thanks. For those (new) names in struct hw_params, the new member names are probably better. This will be more descriptive and keep the HW changes only in console-direct.c, whereas console.c is shared between the other console drivers/platforms and their member names not changing will help.

> need to gate some of the extra stuff behind that CONFIG_CONSOLE_DUAL flag

Great, probably just as well to starting gating right away, some of my comments will mention more detail about it. 

> but console-direct in master currently only sets up 3 pages, I'm not sure why.

It was probably done a long time ago for space reasons. There are other complicated issues though, for instance the "fake" FAT filesystem /dev emulation is limited as to how many /dev entries there can be. So for now keeping /dev/tty1 - 3 as CGA is a good idea to reduce complications that can be dealt with later.

> The way that it's designed now is that we ask the BIOS to tell us which display the user has selected as the 'primary' one. 

Ok. What BIOS Data Address (BDA 0x40:XXXX) is being used for that info?

> So if we only have a CGA card, we get 3 TTYs. CGA + MDA allocates the extra /dev/tty4 on the MDA display.
> if I flip the switches to select my MDA card as the primary display,
> Do you think this makes sense?

Yes, sounds good.

> I remember trying the Nano-X stuff earlier, and as far as I can tell, it's not intended to work with CGA or MDA?

@tyama501 just added CGA support to Nano-X. MDA is not supported.

>  In any case, I propose a check in the DCGET_GRAPH ioctl to return an error code, at least if that console is assigned to a MDA screen that doesn't support graphics.

Lets not change the logic yet on `glock` or its ioctl issues until we've had a chance to get the first revision of this PR committed and can start testing before/after with Nano-X graphics. What I'd like any new hardware-initialization or code determining monitor types to be fully gated with CONFIG_CONSOLE_DUAL for the first cut. The console driver currently uses the current BIOS BDA video mode to determine whether the console is already running in MDA or CGA/EGA (text) mode, and for compatibility reasons needs to stay that way if this option is not set.

> so the primary source of 'bloat' coming from these changes is that we now have logic for setting up video hardware, as we cannot rely on the BIOS to initialize anything but the primary one specified by the user.

Got it. That's fine, as long as all hardware-specific logic is gated. That will ensure full boot compatibility without the option. (It has taken us a while to achieve booting on almost any 8088 system, which is why we're using only a single BDA video mode for the current startup).

> with the MDA card selected as primary, and it works as expected

Great!

> It relies on the VideoSeg global, and it's a bit unclear now which video page on which display that should be. 

Perhaps leave a VideoSeg global which is set unmodified outside the gate, but set to the video segment of what your new code determines the primary monitor is when CONFIG_CONSOLE_DUAL is defined.

I have some other comments on the last commit, will comment directly on the code for that.

elks/arch/i86/drivers/char/console-direct-pc98.c
```diff
@@ -48,27 +48,34 @@
 
 #define MAXPARMS	28
 
+#define MAX_OUTPUTS 1
+
 struct console;
 typedef struct console Console;
 
 struct console {
+    unsigned char op;
```
ghaerr: Byte-sized struct members should not be placed at the top when not next to another byte member, structures are packed and this cause all subsequent word members to be odd-addressed, which greatly slows down CPU access on 8088.

Move new byte members either to the end of the struct or find another unaligned byte member to stick it next to.

What is "op", it is not clear to me? Is this the effective /dev/ttyN number minus 1?

vkoskiv: op = output, it was a bit more obvious when it was still `Output *op` in the previous iteration. That is now used as the index to the arrays `Visible`, `glock` and `glock_wait`, as we currently have one of each for every output.

elks/arch/i86/drivers/char/console-direct-pc98.c
```diff
@@ -48,27 +48,34 @@
...
+    int vseg_offset;		/* vram offset of vseg for this console page */
 #ifdef CONFIG_EMUL_ANSI
     int savex, savey;		/* saved cursor position */
     unsigned char *parmptr;	/* ptr to params */
     unsigned char params[MAXPARMS];	/* ANSI params */
 #endif
+    unsigned short w;
+    unsigned short max_col;
+    unsigned short h;
+    unsigned short max_row;
```
ghaerr: One has to watch members declared after a member array declaration - these could increase instruction offset from byte to word, increasing code size. We're probably OK here, but give the params[] decl above, perhaps gate these variables and move them above CONFIG_EMUL_ANSI to be safe (MAXPARMS is fairy small I think).

elks/arch/i86/drivers/char/console-direct-pc98.c
ghaerr: The above comment applies to all console*.c drivers; I just happened to notice this in PC-98 code.

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -41,34 +49,86 @@
...
-    unsigned int vseg;          /* video segment for page */
-    int basepage;               /* start of video ram */
+    unsigned int vseg;          /* vram for this console page */
+    int vseg_offset;            /* vram offset of vseg for this console page */
 #ifdef CONFIG_EMUL_ANSI
     int savex, savey;           /* saved cursor position */
     unsigned char *parmptr;     /* ptr to params */
     unsigned char params[MAXPARMS];     /* ANSI params */
 #endif
+    enum OutputType type;
```
ghaerr: I hate to say this, but enum's are problematic. Sometimes an enum is a byte, sometimes its a word. This causes confusion. There are no enums used in the kernel at this time. See my [D-Flat project](https://github.com/ghaerr/dflat/blob/master/system.h#L78) for an example of an enum incompatibility mess between compilers.

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -41,34 +49,86 @@
...
+    unsigned short h;
+    unsigned short crtc_base;
+    unsigned vseg_base;
+    int vseg_bytes;
+    unsigned char init_bytes[16];
+    unsigned char n_init_bytes;
+    unsigned char max_pages;
+    unsigned short page_words;
+    unsigned char is_present;
 };
```
ghaerr, line 72: I'm leaning towards the idea of getting all this hardware-specific structures and code moved out of console-direct.c into a standalone monitor-ibmpc.{c,h} (or something like that). Then call the code as functions when gated.

This will remove quite a bit of added complexity from an already complicated console-direct.c. What do you think?
That also allows you to gets as complicated as needed within that file and have it remain hidden from review by those curious about general console operation.

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -183,54 +246,114 @@ struct tty_ops dircon_ops = {
...
-    CCBase = peekw(0x63, 0x40);
-    PageSizeW = ((unsigned int)peekw(0x4C, 0x40) >> 1);
-
-    VideoSeg = 0xb800;
-    if (peekb(0x49, 0x40) == 7) {
-        VideoSeg = 0xB000;
-        NumConsoles = 1;
-        isMDA = 1;
-    } else {
-        isCGA = peekw(0xA8+2, 0x40) == 0;
```
ghaerr, line 205: As discussed, perhaps leave in the above code when not gated (not using globals though) and call the below code in a separate monitor-ibmpc.c file via a wrapper function. We don't want to change the method of determining the monitor type when DUAL is not configured.

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -183,54 +246,114 @@ struct tty_ops dircon_ops = {
...
-        C->basepage = i * PageSizeW;
-        C->vseg = VideoSeg + (C->basepage >> 3);
-        C->attr = A_DEFAULT;
+static const char *type_string(enum OutputType t)
+{
+    switch (t) {
+    case OT_MDA: return "MDA";
+    case OT_CGA: return "CGA";
+    case OT_EGA: return "EGA";
+    case OT_VGA: return "VGA";
```
ghaerr: No need for code here - the strings can be in a static char * array indexed by OT_xxx, used as a %s in printk rather than calling a function to get a string.

We don't differentiate in ELKS between EGA and VGA so perhaps this also should be gated. Although I very much like this new enhancement, I have been trying to reduce the base requirements for ELKS, not increase them. It has been a long road...

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: Yeah, the vague size always bothered me too. I could just #define the different types, and do it that way.

vkoskiv: Is the bitfield feature supported in `ia16-elf-gcc`? Could make it explicitly one byte with `enum outputType type : 8;` in the struct there?

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -183,54 +246,114 @@ struct tty_ops dircon_ops = {
...
     }
 
     kbd_init();
-
-    printk("Direct console, %s kbd %ux%u"TERM_TYPE"(%d virtual consoles)\n",
-           kbd_name, Width, Height, NumConsoles);
+    printk("boot_crtc: %x\n", Con[0].crtc_base);
+    printk("Direct console %s kbd"TERM_TYPE"(%d screens):\n", kbd_name, screens);
+    for (i = 0; i < NumConsoles; ++i) {
+        printk("/dev/tty%i, %s, %ux%u\n", i + 1, type_string(Con[i].type), Con[i].w, Con[i].h);
```
ghaerr: We can't add more printk lines to the startup boot screen, as it then scrolls past and limits determining what has gone wrong. Suggest moving all extra info inside monitor-ibmpc.c, where if the option is enabled that is less of an issue.

In summary, I would say lets try to keep console-*.c, and especially console-direct.c as little modified as possible, with all the hardware related stuff moved to a new .c file. That has the benefit of leaving the system as close to as it is now when not configured, but allowing you to shoot high when running your configured code. This is because this is likely a very specialized configuration option, after all.

elks/arch/i86/drivers/char/console-direct-pc98.c
ghaerr: Still not obvious what "output" is.

> That is now used as the index to the arrays Visible, glock and glock_wait, as we currently have one of each for every output.

I'm not sure making any of the above glock/wait etc variables into arrays is needed - there's only one graphical screen that can be active per machine at a time. Most of this is extra code, all for nothing. Think simplicity at first - ELKS isn't designed to handle everthing, we are very limited in code/data space. I advise leaving glock/wait etc alone for the time being. They can be modified later when somebody tries to run two Nano-X applications simultaneously (which won't work anyways).

elks/arch/i86/drivers/char/console-direct.c
ghaerr: > Could make it explicitly one byte with enum outputType type : 8; in the struct there?

Definitely no bitfields, please. You're new to the project, but we don't want to make the system dependent on a compiler feature when used in only a single place in the kernel. There's just no need for it, and bitfields, enums and various other C features just complicate things. (Yes, I know both are basic nowadays, but we are strive for simplicity around here. If something can be done without invoking a special compiler feature, that's just one less problem for the future when the compiler is changed or the contributor absent (yes, the ELKS compiler did change about 4 years ago, and we are using some GCC extensions for the time being but I'm trying to reduce them. We have a side project trying to bring in the OpenWatcom C compiler for applications which is working - next step is kernel)).

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: That `boot_crtc: ` message was more for debugging, I'll remove that. Then the rest of that can be squeezed back down to that single line that was there before.

ghaerr: > Then the rest of that can be squeezed back down to that single line that was there before.

Don't get me wrong @vkoskiv, I encourage you to write lots of fun code, have debug messages and try new things. We have a `debug()` macro system in debug.h that is pretty fancy for optionally leaving debug messages intact but not displayed until wanted

In general, if we can get this somewhat complicated PR separating most of your new hardware (and other init?) code into a separate file, then it's OK with me to use a multiple-PR approach to adding or perfecting features. That kind of lets you have your own file to play/experiment/enhance, while keeping the rest of the system stable and relatively unmodified. In that way, you can join the fun and write some interesting code while not having to worry about the state/status/direction of the entire codebase, especially in the beginning.

Just wanted to clarify that! :)

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: That `isCGA` flag is only used in `DisplayCursor` for the CRTC registers `cursor start` and `cursor end`, so it's not really a core part of the display type logic. I'll revisit this when I learn more about EGA and VGA, but I do want to ensure we have a robust setup for detecting this stuff.

ghaerr: I was actually referring to the entire bunch of EGA/MDA/CGA init code above: leave that relatively unmodified, then call your external init routine when configured to enhance things if necessary. (I get that some of that isn't possible perhaps given the new design, but that's the general idea).

Specifically about the isCGA flag, this is a good example of somewhat that started simple but got complicated: we wanted to introduce ESC [ ?25 h/l cursor on/off but then found that it didn't work on CGA. Previously, ELKS knew nothing about CGA and now there's another BIOS BDA dependency on it. 

These "BIOS dependencies" are problematic when running on certain hardware emulators (like emu86, etc) and overall contribute to the problems of porting the kernel to embedded systems, PC-98, and the like. Yes, right now we have a completely separate console-direct98.c file, which isn't optimal, but allows @tyama501 total control over getting things to work since he's the ELKS PC-98 features maintainer. At some point the common portions should be merged into a single console-direct.c driver, with only perhaps a specialized init routine for PC-98. In effect, this is what you are kind of doing. I will look further into this thought as you get your code form closer to finalized.

With regards to EGA vs VGA, same issue: ELKS doesn't presently know the difference. We would need to talk more about why we want to know the difference and what programs that would affect before changing kernel behavior between the two. Just more things to think about!

vkoskiv:
> Ok. What BIOS Data Address (BDA 0x40:XXXX) is being used for that info?

From [here](https://stanislavs.org/helppc/bios_data_area.html):

40:63	word	Base port address for active [6845](https://stanislavs.org/helppc/6845.html) CRT controller
			3B4h = mono, 3D4h = color

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: Very good call! I feel a bit silly now, having written that :D

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: Yeah, I think a file like `6845.c` at least would make sense. And I think some of this probing/setup code could also be marked as `INITPROC`, so we free it after booting, right?

ghaerr: > I think a file like 6845.c

Good idea: crtc-6845.c would match the naming scheme we have in elks/arch/i86/kernel/timer-8254.c etc.

And yes on INITPROC as well, as console_init is already INITPROC. The biggest issue we have with far functions is that the compiler has to generate thunking code when those functions call near functions, which adds time and space. But in the case of INITPROC far functions, that extra code, if any, doesn't matter as much because it is released after boot. Calling far procs from near or far procs don't need thunking code.

elks/arch/i86/drivers/char/console-direct-pc98.c
vkoskiv: Output is the display that console is on, so if C->op == 0, then it's on display 0. My original thinking was that we'd want per-output glock/glock_wait in the case that, say, both screens were running a program that wants the terminal in raw mode, for instance, since that lock also used to guard that.

> I'm not sure making any of the above glock/wait etc variables into arrays is needed - there's only one graphical screen that can be active per machine at a time.

In `console-direct.c` it was done to allow for multiple displays, and I changed it in the other ones because they all share the same `console.c` logic. They aren't really arrays in those, though, since `MAX_OUTPUTS` is defined as `1` in them.

vkoskiv: > Byte-sized struct members should not be placed at the top when not next to another byte member, structures are packed and this cause all subsequent word members to be odd-addressed, which greatly slows down CPU access on 8088.

Very good info! Noted.

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: Totally fair! I'll go with `#define` for this, should work quite nicely.

elks/arch/i86/drivers/char/console-direct-pc98.c
ghaerr: > They aren't really arrays in those, though, since MAX_OUTPUTS is defined as 1 in them.

Well, that doesn't matter - they're declared as arrays when they likely don't need to be, which is confusing, and the "array" is indexed by by compiled code, thus generating more code, even when C->op is zero. My point was that those variables are truly singleton system globals and don't need to be indexed, in any of the drivers, if this code is attempting to lock out graphics access on multiple monitors; it isn't needed.

elks/arch/i86/drivers/char/console-bios.c
```diff
@@ -111,14 +118,14 @@ static void scroll(register Console * C, int n, int x, int y, int xx, int yy)
     int a;
 
     a = C->attr;
-    if (C != Visible) {
+    if (C != Visible[C->op]) {
```
ghaerr: I think here we have a situation similar to glock but in reverse: we don't need a global Visible, it seems (nor a "op" member specially to index it). What we likely need is a "char visible" member that tells whether the given console is active (visible) or not. This member can then be used to determine whether to display the incoming data etc, or just store it, depending on whether the specified console is visible or not. Your driver would have to be responsible for setting the visible member correctly for both displays, or only a single display in the non DUAL case. The driver probably should have been written this way in the first place, except it was using globals for most everything - so another was added.

See what I mean?

ghaerr: In other words, we don't need an 'op' member to index a global outside struct Console *Visible[] array, to get another Console *; instead, we just keep a new Console struct member that allows the routine to perform the output or not, as there isn't a need to actually reference a different Console * than the one passed.

elks/arch/i86/drivers/char/console-direct-pc98.c
vkoskiv: Okay, I'll restore those to how they were. We can obviously revisit them later if needed, and admittedly this change was added 'just in case' to begin with.

ghaerr: I can see the approach you've taken - that of systematically following through with each global or other variable, and naturally creating arrays where there was a singleton, etc - and you've done a good job of it, and it works! 

However, for our kernel, we must always save bytes, data structure extensions, code generated, etc., so one must at all times keep that in mind - IIRC we're within 2-3k bytes of running out of kernel code space! You just happen to have taken on a somewhat complex enhancement as only your second enhancement. I don't normally comment this heavily on PRs - I am impressed with what you've done!

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: Is there a INITPROC equivalent for data that is only needed at init time?

ghaerr: Unfortunately no. I had previously added to my todo list to look at whether GCC can store data in the code segment and then generate CS: overrides to get to it. That INITDATA could then be discarded along with INITPROC code.

We might be able to place the data into its own section with some tricky .section naming and linker scripts, but I still don't know a way to force GCC to reload DS when accessing that data. So basically it'll just be part of the larger implementation with CONFIG_CONSOLE_DUAL.

elks/arch/i86/drivers/char/console-direct-pc98.c
vkoskiv: > I don't normally comment this heavily on PRs - I am impressed with what you've done!

I don't remember the last time I enjoyed working on a PR this much! I really appreciate the thorough and informative feedback, so keep it coming! And don't hesitate to point out even small/trivial stuff, since this is only my second PR for this project, and it did end up a bit larger than I initially scoped :sweat_smile: 
If I missed something, I want to know!

One example is that I changed some instances of
```C
Console *C = Con;
```
to
```C
Console *C = &Con[0];
```

Because to me, that communicates the intent more precisely. But if you or any other maintainer of the project thinks the former way is better in any way, then I'm more than happy to fix that up. I try to avoid style changes when it's not my code I'm working on, but there is a good chance I've missed some spots, and those should be fixed as well.

So yeah, I'm really enjoying this :sunglasses:

elks/arch/i86/drivers/char/console-direct-pc98.c
vkoskiv: Fixed in 6ba7c8a8

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: Fixed in 6ba7c8a8

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: ACK, something for a future PR perhaps!

elks/arch/i86/drivers/char/console-bios.c
ghaerr: With the latest changes, I think I can now see why you're still using a Visible[] array: there are two displays! So that's fine, changes look good.

vkoskiv: I see what you mean, it's fairly useless in most places, and was only kept like this because `console-bios.c`, does some things with it that I didn't fully work out yet as I was working on other things.
I'll see if I can factor it out tomorrow.

vkoskiv: Is the `console-bios.c` required on any platform, by the way? Would removing it entirely be feasible, if it's been mostly replaced by `console-direct.c`?

elks/arch/i86/drivers/char/console-direct-pc98.c
ghaerr: Your recent changes look very nice, thank you. Now that lots of detail is moved out into crtc-6845.c, I am starting to see the whole much more clearly. It's looking very good, there are a few small items I will comment on separately.

I am now also seeing the curious problem of why there were (are) Width vs MaxCol, when one is always literally one less than the other. Same for Height and MaxRow. So I don't like it lol :)

Obviously, getting in to changing these kinds of things can be very prone to human error, and I've already spent too much time solving off-by-one issues in the console driver over the past 4 years. But I have an idea:

First, I suggest we use the old names in struct Console: `MaxRow` instead of max_row, Width instead of w, etc.
That should allow for as few visual changes as possible in the large diff listing. And I definitely look at that stuff.

THEN, start replacing Width and Height (in the far fewer routines that use that versus MaxCol and MaxRow it seems) with a *local* variable: `unsigned int Width = C->MaxRow + 1`. Then just use the local Width rather than C->Width in the routine(s).

This may sound like overkill, but I think it allows us to (almost) prove that the large number of small changes in this PR won't in fact break the console, which will cause (me) to have to jump in and debug it, which I don't want to do! Also, the compiler may optimize and generate better code anyways using this approach, since only one struct member has to be referenced.

What do you think of that idea? I'm looking at routines like ScrollDown for best candidates for this kind of thing. If OTOH you think that it might be easier to swap and replace MaxCol with a local `unsigned int MaxCol = C->Width - 1`,  in console.c, then that's ok too. It's possible that two local variables MaxCol and MaxRow at the top of AnsiCmd will save a lot of code being generated!

elks/arch/i86/drivers/char/console-bios.c
ghaerr: console-bios.c is definitely required - its used on all those PC-like systems that won't run console-direct.c or don't have a 6845. We also have some skillful players in the past, that have embedded ELKS into ROM, but want a full-featured console, and implemented INT 10 functions so that console-bios could run.

We try to keep ELKS dependencies to a bare minimum, as we actually have a wide variety of use cases being one of the few actively worked on, stable 8086 operating systems around.

elks/arch/i86/drivers/char/Makefile
```diff
@@ -63,6 +63,9 @@ else
 ifdef CONFIG_CONSOLE_DIRECT
 OBJS += console-direct.o bell-8254.o
 endif
+ifdef CONFIG_CONSOLE_DUAL
+OBJS += crtc-6845.o
+endif
```
ghaerr, line 67: Lets put this ifdef inside the CONFIG_CONSOLE_DIRECT above, so there's no chance it won't cause a kernel compilation failure if the .config file gets a bit out of sync.

elks/arch/i86/drivers/char/console-bios.c
```diff
@@ -41,27 +41,34 @@
 
 #define MAXPARMS        28
 
+#define MAX_OUTPUTS     1
+
 struct console;
 typedef struct console Console;
 
 struct console {
+    unsigned char op;
```
ghaerr: Still a byte. I recommend making struct console the exact same (top part at least) between it and console-direct.c.

BTW, console-bios.c is also very useful for checking out various BIOS implementations, I use it a lot in certain debugging situations.

elks/arch/i86/drivers/char/console-direct-pc98.c
vkoskiv: I actually had that change at some point to eliminate MaxCol and MaxRow, but I didn't want to stuff in more unrelated changes than necessary 😄

Still need to do that renaming, forgot to include that with the latest patches

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -41,34 +43,45 @@
...
 static struct wait_queue glock_wait;
-static Console Con[MAX_CONSOLES], *Visible;
-static Console *glock;          /* Which console owns the graphics hardware */
-static unsigned short CCBase;   /* 6845 CRTC base I/O address */
-static int Width, MaxCol, Height, MaxRow;
-static int NumConsoles = MAX_CONSOLES;
-static unsigned char isMDA, isCGA;
+static Console *Visible[MAX_OUTPUTS];
+static Console Con[MAX_CONSOLES];
+static int NumConsoles = 0;
```
ghaerr: Remove zero initialization. That will move the data to .bss instead of .data, get automatically zeroed, and reduce the kernel file size by a byte.

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -80,38 +93,38 @@ static void std_char(register Console *, int);
...
-    outb(15, CCBase);
-    outb(Pos, CCBase + 1);
+    outb(14, C->crtc_base);
+    outb(Pos >> 8, C->crtc_base + 1);
+    outb(15, C->crtc_base);
+    outb(Pos, C->crtc_base + 1);
 }
 
-static void DisplayCursor(int onoff)
+static void DisplayCursor(register Console * C, int onoff)
```
ghaerr: Don't add new code with register declarations anywhere. They're completely ignored by GCC and are slowly being removed then a file is updated. I would recommend removing them all here, but the PR is large enough already.

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -183,41 +199,56 @@ struct tty_ops dircon_ops = {
     Console_conout
 };
 
+#if !defined(CONFIG_CONSOLE_DUAL)
 void INITPROC console_init(void)
```
ghaerr, line 203: Use #ifndef rather than #if !defined. Currently for readability, we're only using #if !defined when followed by || or another check not possible with #ifndef.

elks/arch/i86/drivers/char/crtc-6845.c
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+#define CRTC_INDX 0x0
+#define CRTC_DATA 0x1
+#define CRTC_MODE 0x4
+#define CRTC_CSEL 0x5
+#define CRTC_STAT 0x6
+
+/* Check to see if this CRTC is present */
+char INITPROC crtc_probe(unsigned short crtc_base)
```
ghaerr: I can't recommend using `char` either as a return value or an argument: there's too many problems with it. For instance, the compiler will generate `CBW` (convert byte to word) instead of just using AX when using the return value of this function (needlessly).

elks/arch/i86/drivers/char/crtc-6845.c
```diff
@@ -0,0 +1,69 @@
...
+    for (i = 0; i < 100; ++i) {} // TODO: Verify this doesn't get optimized out
+    outb(0x0F, crtc_base + CRTC_INDX);
+    unsigned char value = inb(crtc_base + CRTC_DATA);
+    if (value != test) return 0;
+    outb(0x0F, crtc_base + CRTC_INDX);
+    outb(original, crtc_base + CRTC_DATA);
+    return 1;
+}
+
+int INITPROC crtc_init(unsigned char t)
```
ghaerr: Same here for unsigned char t argument. Using chars as array indices is asking for trouble. Switch statements are especially bug-prone as sign extension occurs and case labels skipped with no warning. Just use int, no need for special typing.

elks/arch/i86/drivers/char/crtc-6845.h
```diff
@@ -0,0 +1,55 @@
...
+static const char *type_string[] = {
+    "MDA",
+    "CGA",
+    "EGA",
+    "VGA",
+};
+#endif
+
+#if defined(CONFIG_CONSOLE_DUAL)
+struct hw_params {
```
ghaerr, line 17: I prefer that structures stay defined regardless of config option. While seemingly useless, this allows for other additions to signal an error should they conflict with another option, even though that option is not selected. So this gate can move to line 29.

elks/arch/i86/drivers/char/crtc-6845.h
```diff
@@ -0,0 +1,55 @@
...
+    unsigned short w;
+    unsigned short h;
+    unsigned short crtc_base;
+    unsigned vseg_base;
+    int vseg_bytes;
+    unsigned char init_bytes[16];
+    unsigned char n_init_bytes;
+    unsigned char max_pages;
+    unsigned short page_words;
```
ghaerr, line 24: Can these two members move above init_bytes for clarity? Will be easier to read. n_init_bytes could also be after page_words and max_pages to make things even clearer, while still keeping the char members together.

elks/arch/i86/kernel/timer.c
```diff
@@ -81,19 +81,20 @@ void timer_tick(int irq, struct pt_regs *regs)
 
 #ifdef CONFIG_CONSOLE_DIRECT
     /* spin timer wheel in upper right of screen*/
-    if (spin_on && !(jiffies & 7)) {
-        static unsigned char wheel[4] = {'-', '\\', '|', '/'};
-        static int c = 0;
+    // FIXME
+    // if (spin_on && !(jiffies & 7)) {
```
ghaerr: I would think this is now easily fixable using Visible[0]->vseg for VideSeg, if Visible[0] is always the primary monitor.

elks/arch/i86/drivers/char/console-direct-pc98.c
ghaerr: > I actually had that change at some point to eliminate MaxCol and MaxRow,

You're getting ahead of me - nice :) I understand about the hesitancy but now things are coming together nicely, thank you!

vkoskiv:
Okay, so I applied the `Width/Height/MaxRow/MaxCol` rename as an interactive rebase, so it applies to this whole branch from the beginning.
Is that okay to do in this project, or are fixup commits preferred instead? Figured I'd check, since not everyone is okay with the 'rewrite history' approach to PRs.
Ignore the two extra force pushes, my rebase left git a bit confused for a moment there.

vkoskiv:
If the rebase approach is okay, I could apply the rest of the fixes like that, too, so we'd end up with 2-3 good commits, and no more extra fixup/feedback commits. Both ways are fine by me.

ghaerr:
I think we're OK with the rebase commits, but git is not my forte (nor do I want it to be!). Agreed there's been probably too many fixup commits to be of much interest to people, so a rewritten history is fine by me for this PR.

I plan on either a squash or merge commit at the end of this; a merge commit would be nice if you want to keep a few good commits showing changes. 

The Width/Height/MaxRow/MaxCol rename is looking good - it is showing exactly what i wanted to see and making it easy to ensure no bugs. The big question now is whether to optimize out Width/Height, or MaxRow/MaxCol. I want to take a slightly closer look, but it seems that if you went with moving MaxRow/MaxCol to locals variables, it could save quite a bit of code generation in the shared console.c routines. As far as speed goes, either way won't matter with the Scroll/Clear routines etc, as they take way longer than what we're concerned with.

ghaerr:
@vkoskiv: I just reviewed our large diff. With regards to the Width/Height/MaxCol/MaxRow we've been talking about: I say you should go ahead and remove MaxCol and MaxRow as members of the Console struct, and replace them at the top of any function which needs them with a local variable of the same name computed from Width or Height. This should lessen the amount of code produced in console.c::AnsiCmd() by quite a bit, and frankly, Width and Height are much more commonly understood than the zero-based equivalent MaxCol/MaxRow. This method also allows us to ensure porting errors to be at a minimum, as the diff will be simple to inspect. And if we're lucky, they'll be some savings as a result of the to balance the added bloat when CONFIG_CONSOLE_DUAL isn't defined.

What do you think about that?

vkoskiv:
@ghaerr Sounds good to me! I have a patch ready for it, I'll just do some checks to ensure I didn't break anything. In the meantime, I addressed most (all?) of the other feedback, mostly as fixup commits to the existing commits.
Ideally I'd like to try squashing the `Remove output struct to reduce indirection` commit into the initial commit to reduce code churn in the git history, I'll see if I can pull that off a bit later.
Really happy with how this PR is turning out, thanks again for the excellent feedback so far. Learned some really useful things!

ghaerr:
Ok @vkoskiv, glad to see you back and get this finished up. Thanks for all the changes. I'll do another code review and then a hopefully final one after the MaxRow/MaxCol elimination. I am anxious to learn how much the code size changes up or down for the console driver when the DUAL option is not selected, given some of these later optimizations.

elks/arch/i86/drivers/char/console-direct.c
```diff
@@ -41,34 +43,45 @@
 
 #define MAXPARMS        28
 
+#ifdef CONFIG_CONSOLE_DUAL
+#define MAX_OUTPUTS     2
+#else
+#define MAX_OUTPUTS     1
+#endif
```
ghaerr, line 50: I'm still having trouble quickly understanding what an "output" is. Can you please rename MAX_OUTPUTS to MAX_DISPLAYS throughout, so that it's obvious?

We're almost there, there are still a couple unresolved conversations outstanding. Can you take a quick look at those?

Thank you!

ghaerr:
> Ideally I'd like to try squashing the Remove output struct to reduce indirection commit into the initial commit to reduce code churn in the git history, I'll see if I can pull that off a bit later.

Don't worry about that if your time is short - I'm happy with the commits you've pushed and plan to do a merge commit.

elks/arch/i86/drivers/char/console-direct.c
vkoskiv: I applied a fixup to the initial commit, so we now have `MAX_OUTPUTS` -> `MAX_DISPLAYS` and `C->op` -> `C->display`

elks/arch/i86/kernel/timer.c
vkoskiv: Fixed, I didn't want to expose the entire `Visible` array, so I'm just setting the existing `VideoSeg` global at init time.

elks/arch/i86/drivers/char/crtc-6845.h
vkoskiv: Done! Looks much nicer now.

elks/arch/i86/drivers/char/crtc-6845.c
vkoskiv: Fixed, and I'll make a mental note of this for the future :]

elks/arch/i86/drivers/char/crtc-6845.c
vkoskiv: Good to know! I need to get into the habit of checking out the generated assembly more often.

elks/arch/i86/drivers/char/Makefile
vkoskiv: Fixed. Not sure why Github doesn't always show the `outdated` tag on these when I push changes.

elks/arch/i86/drivers/char/console-direct.c
ghaerr: Thanks, the C->op rename to C->display was a good idea too. Makes much more immediate sense to me now.

ghaerr:
All of this is looking very nice indeed! The removal of C->MaxCol and C->MaxRow really helped make the diff smaller, which definitely helps regression testing via eyeballs. I am going to fly my drones for a bit but will take a final look after and hopefully get this committed today :)

The build is failing because crtc_init and crtc_probe are undefined - please test without CONFIG_CONSOLE_DUAL defined, that's likely the problem.

vkoskiv:
> The build is failing because crtc_init and crtc_probe are undefined - please test without CONFIG_CONSOLE_DUAL defined, that's likely the problem.

I've done test builds with different sets of options, and it builds fine on my system with `CONFIG_CONSOLE_DUAL` disabled in the kernel config. Not sure what's tripping up the CI build yet :thinking:

ghaerr:
Try running `./buildimages.sh` or `./build.sh auto allimages` in the main directory, it may be related to building the 8018x port. In any case, the latter runs the same stuff that the CI does. You can check

vkoskiv:
I can repro with `./buildimages.sh`. Not sure why, but `autoconf.h` gets `#define CONFIG_CONSOLE_DUAL 1` even though the current `.configure` doesn't even have that config option in it to begin with.

vkoskiv:
I changed the `y` to an `n` in `elks/arch/i86/drivers/char/config.in` for the `CONFIG_CONSOLE_DUAL` line. I don't know why, but that seems to fix the issue.

ghaerr:
@vkoskiv: You've done a great job on this, thank you for the wonderful enhancement! I'm going to commit this now.

I will follow up with a PR to make the change from unsigned short to int for Width and Height, with a few other minor changes, as well as an answer to your above question. I'm anxious to see what the kernel code and data size change for this with DUAL on and off. We can continue discussion on that PR if needed.

Thank you!

PR #1993, opened 2024-09-04
[bootopts] Add heap= option for dynamic sizing of kernel heap

ghaerr:
Here you go @toncho11 and @FrenkelS: this new /bootopts option will allow ELKS to boot up with a minimum of extra kernel resources, allowing for a usable contiguous block of memory ~500Kb for Doom, and more if UMB memory is available.

Adds `heap=` /bootopts option to set the size of the internal kernel heap at startup. By default, ELKS will try to reserve up to 64k of memory for use in its heap allocator for runtime kernel resources. In practice, because of the existing static data from the compiled C routines, this amounts to a ~45k heap. My testing shows that the kernel is stable lowering this to 15k when tasks, buffers, cache, and nodes are also lowered, thus saving ~30k memory, which is then available for user programs, e.g. Doom.

Here's a screenshot of `meminfo` running directly after boot with the reduced resource options in effect:
meminfo

Notice at the bottom right there's 492K of 520k available, with the largest block being 499,888 bytes. If the kernel were to be recompiled with minimal drivers (e.g. no networking), this could be made even larger.

I have modified the default /bootopts to include both the "normal" settings for kernel resources and the example used to generate the display above, to make it easy to play with. Be aware that no checking is done on the heap size passed, and passing a value too low can cause system instability or crashes. I played around and found 15000 as a working lower limit, given the other option values.

Here's the defaults:
```
task=16 buf=64 cache=8 file=64 inode=96 heap=44000
```
And the minimal system:
```
task=6 buf=8 cache=4 file=20 inode=24 heap=15000 n
```
Notice the 'n' at the end of the line, that passes 'n' to /bin/init which means don't run ./etc/rc.sys on startup, which also saves memory fragmentation at the cost of the clock not being set and auto-mounts not processed, etc.

Feel free to play around with these resource settings, depending on what Doom requires.

The startup boot screen is slightly changed: at the bottom right, it now displays the largest contiguous block of memory before running /bin/init, followed by the released "INITPROC" code segment, followed by the UMB memory configured. Since the startup screen is tight on space, This might be shown as "509+9+64K free" which means 509k contiguous main memory plus a 9k free block plus a 64k UMB free block. It also shows the number of tasks, files, inodes and heap for reference.

Here's the changed startup screen (note we've moved to v0.8.0-pre3):
boot

Other enhancements include showing "MEM" instead of "SEG" in `meminfo` for easier understanding, and "CACH" instead of "BUFH" to differentiate cache= L1 buffers from L2 buf= buffers.

### DIFF
--- a/elks/Makefile-rules
+++ b/elks/Makefile-rules
@@ -58,7 +58,7 @@ include $(TOPDIR)/Make.defs
 VERSION     = 0	# (0-255)
 PATCHLEVEL  = 8	# (0-255)
 SUBLEVEL    = 0	# (0-255)
-PRE         = -pre2
+PRE         = -pre3
 
 #########################################################################
 # Specify the architecture we will use.
--- a/elks/arch/i86/kernel/system.c
+++ b/elks/arch/i86/kernel/system.c
@@ -11,57 +11,58 @@
 #include <arch/segment.h>
 #include <arch/io.h>
 
-byte_t sys_caps;		/* system capabilities bits */
+seg_t membase, memend;  /* start and end segment of available main memory */
 unsigned int heapsize;	/* max size of kernel near heap */
+byte_t sys_caps;        /* system capabilities bits */
 
-void INITPROC setup_arch(seg_t *start, seg_t *end)
+unsigned int INITPROC setup_arch(void)
 {
+	unsigned int endbss, heapsegs;
+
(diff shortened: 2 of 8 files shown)


toncho11:
Thank you!!!
This needs to be explained in a wiki page. I will call it "ELKS Minium Memory Boot".

PR #1999, opened 2024-09-05
[kernel] Consolidate reboot, shutdown and poweroff into single program

ghaerr:
Combines three previously separate programs into `shutdown`, then uses hard links to create `reboot` and `poweroff`, reducing disk space while increasing functionality.

The functionality of these programs remains mostly the same, except that `poweroff` now also tries to unmount all mounted filesystems before powering down the system. Combining these almost-identical programs saves about 3-4k disk space.

Rewrites sys_reboot for some simplification, the two magic numbers passed as first arguments are now ignored, and previously unused options are removed.


toncho11:
How are hard links done on Fat?

ghaerr:
> How are hard links done on Fat?

They're not. For FAT, the files have to be copied multiple times. For the smallest image, 360k, we are currently only copying shutdown, instead of all three, just for this reason. But that shouldn't be a big deal not having poweroff or reboot commands on 360k.

Actually, I suppose we could add an option to shutdown/power/reboot to do any of the three, (-s/-p/-r) which could help. The current executable is 1280 bytes, so adding more code won't take away any space.

PR #2003, opened 2024-09-11
[basic cga] CGA Graphic support for IBM-PC ELKS Basic

tyama501:
Hello @ghaerr 

This is the first attempt to add CGA PLOT/DRAW for IBM-PC ELKS Basic as we have been talking on
https://github.com/ghaerr/elks/discussions/1939#discussioncomment-10605658

I have added __ia16__ for the assembler codes and
I have added empty int_10 and memclrw() to the else case for now.

After I merged upstream,
I realized I cannot use seg_t with compile error.
So, I used unsigned int instead.
(Something changed recently?)

Thank you!

### DIFF
--- a/elkscmd/basic/Makefile
+++ b/elkscmd/basic/Makefile
@@ -23,7 +23,8 @@ OBJS = basic.o host.o
 #OBJS += ../debug/instrument.o ../debug/syms.o ../debug/shared.o ../debug/shared-asm.o
 
 ifeq ($(CONFIG_ARCH_IBMPC), y)
-OBJS += host-stubs.o
+#OBJS += host-stubs.o
+OBJS += host-ibmpc.o
 endif
 
 ifeq ($(CONFIG_ARCH_8018X), y)
--- a/elkscmd/basic/Makefile
+++ b/elkscmd/basic/Makefile
@@ -23,7 +23,6 @@ OBJS = basic.o host.o
 #OBJS += ../debug/instrument.o ../debug/syms.o ../debug/shared.o ../debug/shared-asm.o
 
 ifeq ($(CONFIG_ARCH_IBMPC), y)
-#OBJS += host-stubs.o
 OBJS += host-ibmpc.o
 endif
 
--- a/elkscmd/basic/Makefile
+++ b/elkscmd/basic/Makefile
@@ -23,7 +23,7 @@ OBJS = basic.o host.o
 #OBJS += ../debug/instrument.o ../debug/syms.o ../debug/shared.o ../debug/shared-asm.o
 
 ifeq ($(CONFIG_ARCH_IBMPC), y)
-OBJS += host-ibmpc.o
+OBJS += host-ibmpc.o asm-ibmpc.o
 endif
 
 ifeq ($(CONFIG_ARCH_8018X), y)
--- a/elkscmd/basic/asm-ibmpc.S
+++ b/elkscmd/basic/asm-ibmpc.S
@@ -1,0 +1,46 @@
+#define ARG0	2
+#define ARG1	4
+#define ARG2	6
+#define ARG3	8
+#define ARG4	10
+
+	.arch	i8086, nojumps
+	.code16
+	.text
+
+	.global fmemsetw
+	.global int_10
+
+// void fmemsetw(void * off, unsigned int seg, unsigned int val, size_t count)
+
+fmemsetw:
+	mov    %es,%bx
+	mov    %di,%dx
+	mov    %sp,%di
+	mov    ARG2(%di),%ax  // value
+	mov    ARG3(%di),%cx  // byte count
+	les    ARG0(%di),%di  // far pointer
+	cld
+	rep
+	stosw
+	mov    %dx,%di
+	mov    %bx,%es
+	ret
+
+// void int_10(unsigned int ax, unsigned int bx,
+//             unsigned int cx, unsigned int dx)
+
+int_10:
+	push   %bp
+	push   %si
+	push   %di
+	mov    %sp,%di
+	mov    ARG0+6(%di),%ax
+	mov    ARG1+6(%di),%bx
+	mov    ARG2+6(%di),%cx
+	mov    ARG3+6(%di),%dx
+	int    $0x10
+	pop    %di
+	pop    %si
+	pop    %bp
+	ret
--- a/elkscmd/basic/host-8018x.c
+++ b/elkscmd/basic/host-8018x.c
@@ -89,3 +89,6 @@ int host_inpb(int port) {
 int host_inpw(int port) {
     return inw(port);
 }
+
+void host_exit() {
+}
--- a/elkscmd/basic/host-8018x.c
+++ b/elkscmd/basic/host-8018x.c
@@ -89,6 +89,3 @@ int host_inpb(int port) {
 int host_inpw(int port) {
     return inw(port);
 }
-
-void host_exit() {
-}
--- a/elkscmd/basic/host-ibmpc.c
+++ b/elkscmd/basic/host-ibmpc.c
@@ -53,7 +53,7 @@ void int_10(unsigned int ax, unsigned int bx,
                       "pop %ds;");
 }
 
-void memclrw(unsigned int offset, seg_t seg, unsigned int count)
+void memclrw(unsigned int offset, unsigned int seg, unsigned int count)
 {
     __asm__ volatile ("push %ds;"
                       "push %es;"
--- a/elkscmd/basic/host-ibmpc.c
+++ b/elkscmd/basic/host-ibmpc.c
@@ -4,14 +4,7 @@
  */
 #include <stdio.h>
 #include <stdlib.h>
-#if __ia16__
 #include <arch/io.h>
-#else
-#define inb(port)           0
-#define inw(port)           0
-#define outb(value,port)
-#define outw(value,port)
-#endif
 #include "host.h"
 #include "basic.h"
 
--- a/elkscmd/basic/host-ibmpc.c
+++ b/elkscmd/basic/host-ibmpc.c
@@ -11,8 +11,6 @@
 #define VIDEO_05_G320x200   0x0005
 #define VIDEO_03_T80x25     0x0003
 
-extern FILE *outfile;
-
 static int gmode = 0;
 static int exit_on = 0;
 
@@ -93,11 +91,16 @@ int host_analogRead(int pin) {
 void host_pinMode(int pin,int mode) {
 }
 
+static void mode_reset() {
+    if (gmode)
+        int_10(VIDEO_03_T80x25, 0, 0, 0);
+}
+
 void host_mode(int mode) {
     gmode = mode;
 
     if (gmode && !exit_on) {
-        atexit(host_exit);
+        atexit(mode_reset);
         exit_on = 1;
     }
 
@@ -344,9 +347,4 @@ int host_inpb(int port) {
 
 int host_inpw(int port) {
     return inw(port);
-}
-
-void host_exit() {
-    if (gmode)
-        int_10(VIDEO_03_T80x25, 0, 0, 0);
 }
--- a/elkscmd/basic/host-ibmpc.c
+++ b/elkscmd/basic/host-ibmpc.c
@@ -1,0 +1,359 @@
+/*
+ * Architecture Specific routines for CGA
+ * Sep 2024 Takahiro Yamada
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#if __ia16__
+#include <arch/io.h>
+#else
+#define inb(port)           0
+#define inw(port)           0
+#define outb(value,port)
+#define outw(value,port)
+#endif
+#include "host.h"
+#include "basic.h"
+
+#define VIDEO_05_G320x200   0x0005
+#define VIDEO_03_T80x25     0x0003
+
+extern FILE *outfile;
+
+static int gmode = 0;
+static int exit_on = 0;
+
+typedef struct {
+    int x;
+    int y;
+    int fgc;
+    int bgc;
+    int r;
+} xyc_t;
+
+static xyc_t gxyc = {0, 0, 7, 0, 1};
+
+#if __ia16__
+void int_10(unsigned int ax, unsigned int bx,
+            unsigned int cx, unsigned int dx)
+{
+    __asm__ volatile ("push %ds;"
+                      "push %es;"
+                      "push %bp;"
+                      "push %si;"
+                      "push %di;");
+    __asm__ volatile ("int $0x10;"
+                      :
+                      :"a" (ax), "b" (bx), "c" (cx), "d" (dx)
+                      :"memory", "cc");
+    __asm__ volatile ("pop %di;"
+                      "pop %si;"
+                      "pop %bp;"
+                      "pop %es;"
+                      "pop %ds;");
+}
+
+void memclrw(unsigned int offset, seg_t seg, unsigned int count)
+{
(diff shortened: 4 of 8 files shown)


ghaerr:
Hello @tyama501,

In general this looks very good, but there are some changes I'd like to see before commit. I will comment below and in other cases by the code itself.

> I have added ia16 for the assembler codes and
I have added empty int_10 and memclrw() to the else case for now.

In general, one can't assume that `#if !__ia16__` this means Watcom. Thus, better to `#ifdef __WATCOMC__` the unfinished portions of the int_10 and memclrw routines. However - in this case I would recommend removing the blank routines entirely as the Watcom build uses host-stub.c and thus CGA isn't supported anyways. I feel it is better to leave the routines completely out than to add them which do nothing, as this sets up a source of bugs where one forgets the functions are not in fact implemented. The routines can be added in when they are written.

> I realized I cannot use seg_t with compile error.
So, I used unsigned int instead.
(Something changed recently?)

Yes, and unsigned int is fine. During the Doom port to ELKS, it was found that all the kernel types (seg_t, sector_t, etc) were being exposed even though libc because the C library headers include <linuxmt/types.h>. Long story short, both Doom and ELKS uses `sector_t` differently and it was a mess. I made a change that stops all kernel types from leaking out to C library. This change also necessitated a change in `fmemalloc` to use `unsigned int` rather than `seg_t` as well.

elkscmd/basic/Makefile
```diff
@@ -23,7 +23,8 @@ OBJS = basic.o host.o
 #OBJS += ../debug/instrument.o ../debug/syms.o ../debug/shared.o ../debug/shared-asm.o
 
 ifeq ($(CONFIG_ARCH_IBMPC), y)
-OBJS += host-stubs.o
+#OBJS += host-stubs.o
```
ghaerr: Remove this, rather than comment out.

elkscmd/basic/host-8018x.c
```diff
@@ -89,3 +89,6 @@ int host_inpb(int port) {
 int host_inpw(int port) {
     return inw(port);
 }
+
+void host_exit() {
+}
```
ghaerr: Remove all the `host_exit` definitions from all files except host-ibmpc.c. This is because the main basic.c never calls host_exit, it should instead be a static function in host-ibmpc.c only (and probably renamed to something like mode_reset so as to not be confused with the other host_xxx calls required by all host*.c handlers).

elkscmd/basic/host-ibmpc.c
```diff
@@ -0,0 +1,359 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#if __ia16__
+#include <arch/io.h>
+#else
+#define inb(port)           0
+#define inw(port)           0
+#define outb(value,port)
+#define outw(value,port)
+#endif
```
ghaerr: Remove all the "faked" defines and the `#if __ia16__` above it, since the definitions are inaccurate and don't work. This file does not support compilation with Watcom, which uses host-stubs.c instead.

elkscmd/basic/host-ibmpc.c
```diff
@@ -0,0 +1,359 @@
...
+#include "host.h"
+#include "basic.h"
+
+#define VIDEO_05_G320x200   0x0005
+#define VIDEO_03_T80x25     0x0003
+
+extern FILE *outfile;
```
ghaerr: Lets move this (and the same in host-stubs.c) into host.h, since now there are two host-xxx.c files using the extern.

elkscmd/basic/host-ibmpc.c
```diff
@@ -0,0 +1,359 @@
...
+#if __ia16__
+void int_10(unsigned int ax, unsigned int bx,
+            unsigned int cx, unsigned int dx)
+{
+    __asm__ volatile ("push %ds;"
+                      "push %es;"
+                      "push %bp;"
+                      "push %si;"
+                      "push %di;");
+    __asm__ volatile ("int $0x10;"
```
ghaerr: Have you looked at the code generated for this function using `ia16-elf-objdump -D -r -Mi8086 host-ibmpc.o`? 

The first `__asm__` directive is using a very non-standard register save sequence, and I would like to know that gcc isn't also saving the same registers on its own, which it usually does when indicating the loads required on line 47.

Also, saving BP is almost always the job of the compiler, as it will generate code using BP indirection to reference the parameters passed to the function. So I believe it should not be saved explicitly, but you need to look at the code to confirm.

I also believe that SI and DI are automatically saved by both the compiler if needed, as well as the BIOS.

After confirming which registers need to be saved, I then am wondering whether/why DS and ES need to be saved: it is my understanding that INT 10h saves segment register and likely SI and DI as well. Have you determined that these registers need to the saved in order for the code to work?

Finally, the use of a non-standard register save sequence at function prologue breaks our stack backtrace library found in libc/debug/stacktrace.c. The standard register save sequence used by ia16-elf-gcc is to save (and only when needed) SI, DI, BP (then SP->BP). After that, DS and/or ES are pushed, but only when needed. This is a complicated subject but we can talk more about it if you need more information.

elkscmd/basic/host-ibmpc.c
```diff
@@ -0,0 +1,359 @@
...
+                      "pop %ds;");
+}
+
+void memclrw(unsigned int offset, unsigned int seg, unsigned int count)
+{
+    __asm__ volatile ("push %ds;"
+                      "push %es;"
+                      "push %bp;"
+                      "push %si;"
+                      "push %di;");
```
ghaerr: Contrary to above, in this function obviously ES and ES will have to be saved explicitly. But the compiler should generate the normal function prologue, see below.

elkscmd/basic/host-ibmpc.c
```diff
@@ -0,0 +1,359 @@
...
+    __asm__ volatile ("mov %%ax,%%di;"
+                      "mov %%bx,%%es;"
+                      "xor %%ax,%%ax;"
+                      "cld;"
+                      "rep;"
+                      "stosw;"
+                      :
+                      :"a" (offset), "b" (seg), "c" (count)
```
ghaerr: It is better to inform the compiler of the exact registers needed (in this case DI), than to tell the compiler to load them into a general register (AX in this case) and then write ASM that the compiler doesn't know about that loads DI. I think you can use the "D" instead of "a" to load DI directly; try it and look at the disassembled output to be sure.

Also, this routine saves SI for no reason - see above for discussion. If possible we should let the compiler generate the function prologue and epilogue.

Using the GCC ASM constraints mechanism is very tricky, non-portable, and generally I can't recommend it. What ends up happening, which is already happening, is that code written only supports a single compiler and then has to be completely rewritten to support OpenWatcom. I think a better approach would be to move all C-based ASM outside the C file and into a .S or .asm file. I realize that we have then the stupid/big problem of two different ASM formats, but I think some discussion might be useful as to coming up with a way to allow simultaneous support of multiple compilers with a single .S ASM file. 

Unfortunately, doing anything like my suggestion right away would take too much time, so we'll hold on that. But I do like the idea of moving all the ASM code out into separate .S files. They could then be added to the C library and used for other applications, rather than only this one. What do you think of that idea?

If the routine(s) is moved out to a .S file, memclrw should be renamed fmemsetw to be more inline with other C library function names.

elkscmd/basic/host-ibmpc.c
```diff
@@ -0,0 +1,359 @@
...
+
+int host_analogRead(int pin) {
+    return 0;
+}
+
+void host_pinMode(int pin,int mode) {
+}
+
+void host_mode(int mode) {
+    gmode = mode;
```
ghaerr, line 52: I realize this file started by copying from host-stubs.c first, but I notice that we have two distinct C coding styles now going on in the same file with regards to the starting left brace; pick one and change the others, thank you.

elkscmd/basic/host-ibmpc.c
tyama501: Actually PC-98 uses intC5-pc98.S already,
so if you are ok I will make separate .S file.

elkscmd/basic/host-ibmpc.c
tyama501: I have looked the objdump before and yes, I knew bp is saved twice,
but I haven't fixed that yet.
I don't know much about INT 10h and whether it saves ES, DS and others or not,
but maybe you are right.

I will try to write separate .S as you mentioned.

elkscmd/basic/host-ibmpc.c
tyama501: Sorry, I couldn't understand which one you are pointing.

Are you talking about
void int_10()
{
}
and others like
void host_mode(){
}

elkscmd/basic/host-ibmpc.c
tyama501: Do you have preferred file name for the .S file?
I think both int_10 and memclrw would be moved out.

elkscmd/basic/host-ibmpc.c
tyama501: OK.
host-pc98.c also uses host.h but may be that is not big problem.

elkscmd/basic/host-ibmpc.c
ghaerr: > I don't know much about INT 10h and whether it saves ES, DS and others or not,

[Here's a reference site I use a lot](https://stanislavs.org/helppc/int_10.html). They do mention that SI and DI may get trashed on PS/2 so I would save them, but not bother with the segment registers, as unless needed, it will just serve to slow down the graphics that much more.

elkscmd/basic/host-ibmpc.c
ghaerr: > Actually PC-98 uses intC5-pc98.S already,
so if you are ok I will make separate .S file.

Oh, now I see the pc98.S file. I also notice that host-pc98.c has lots of ASM code (also using non-standard prologues etc). But I probably was thinking that PC-98 is special case, which IBM PC is not so much. But ultimately that file is also non-portable, and we're mixing .S and .c files for ASM in the same project for the same platform! Oh well, I suppose that's fine since its' already done.

> Do you have preferred file name for the .S file?

I was thinking just asm-ibmpc.S, and then rename intcC5-pc98.S to asm-pc98.S.

elkscmd/basic/host-ibmpc.c
tyama501: Oh, is this mean only the extern FILE?
or is it also for the #define VIDEO?
(I thought it is also for the define.)

elkscmd/basic/host-ibmpc.c
ghaerr: In general, I think the best way to handle routines that require ASM is to put them in a .S file instead of using nonstandard compiler ASM extension. That also allows the same routines (graphics in this case) to be used again by yourself in other code, or be put in the C library. It also gets one to think about exactly what the parameters should be in a more general fashion, and also to write specific special prologues for very high speed, rather than using or competing against the C compiler generate prologues.

For examples of very fast prologues, see the kernel library function elks/arch/i86/lib/fmemory.S::fmemsetw. Notice how the function prologue is extremely fast by not pushing onto the stack, and that the parameters to the function are such that a LDS or LES instruction can be used to grab both words of a far pointer. Not every ASM function should be written this way, but memory copy routines are expected to run very quickly. For slower routines, using a gcc-compatible function prologue will help because of our symbolic stack tracing (which I hope to make available in an easier fashion sometime soon).

Since we already have a working fmemsetw for the kernel, perhaps just that function should be copied out of that library fmemory.S file and reused for this work, rather than writing another routine? If all the functions were pulled out into separate .S files, then they could more easily be put into the C library under libc/asm. But we will have to wait a bit for that because more work is required to get them to work in medium model (and regparmcall as well). This shows the extended work required to "do it right", but at least we'd be moving in that direction.

elkscmd/basic/host-ibmpc.c
ghaerr: > Oh, is this mean only the extern FILE?

I was only referring to the extern FILE *.

elkscmd/basic/host-ibmpc.c
ghaerr: > Sorry, I couldn't understand which one you are pointing.

Either one, I was just suggesting that all the function start braces should be the same way. My fault, as originally I quickly wrote host-stubs.c and used the first-brace-on-the-same-line style which makes for short code (since they were all empty) but is not very often used coding style.

toncho11:
Would this work on VGA monitor or it is only CGA?

tyama501:
Hi @ghaerr 

I have added asm-ibmpc.S and
copied fmemsetw and also added int_10 for now.

It worked.
I think I have done all you mentioned but if I missed something let me know.

Hi @toncho11 
It is for 320x200, 4 color mode (CGA)
but most ibm compatible pc supports this mode, doesn't it?
(Sorry if it doesn't)

Thank you.

ghaerr:
Hello @tyama501,

This now looks quite nice and very clean, thank you! We are better off with these functions directly in ASM, as they are now quite easy to understand and read, even though we haven't (yet) solved the problem of ASM portability between ia16 and Watcom!

I do have a comment about using ARG0 etc in the int_10 function which I will address with the code.

Thank you!

elkscmd/basic/asm-ibmpc.S
```diff
@@ -0,0 +1,46 @@
...
+int_10:
+	push   %bp
+	push   %si
+	push   %di
+	mov    %sp,%di
+	mov    ARG0+6(%di),%ax
+	mov    ARG1+6(%di),%bx
+	mov    ARG2+6(%di),%cx
+	mov    ARG3+6(%di),%dx
+	int    $0x10
```
ghaerr, line 42: This ASM file is quite clean and nicely done, thank you. Much easier to read and understand than the C extensions, IMO and ready to be used by other programs more easily when needed thank you.

I was initially going to comment about using "ARG0+6" etc in this function, as well as the non-standard prologue and using DI to access the arguments, only to see that (my) fmemsetw function effective does the same thing! So I think everything is fine and will commit now.

Should the int_10 function be moved to a separate file in the future (leaving fmemset in its own fmemset.S file since it is written for very high speed w/non-standard prologue), then I would advise moving back to standard prologue and argument accessing, which will make things far easier if the function has to be used for medium model. This was not done for the original fmemsetw because that routine was specially written for the kernel and would always be accessed as a near procedure, even with some far functions, FYI.

So, when this function is moved into a separate file, or other functions are added, I would suggest using something like the following:
```
#define ARG0  4
#define ARG1  6
...
int_10:
 	push   %bp
        move   %sp,%bp   // Using BP ensures SS: used for access, unlike DI which requires DS==SS.
 	push   %si       // with BP=SP in previous step, any number of registers can be saved without adjusting ARGX
 	push   %di
 	mov    ARG0+FAR_ADJ(%bp),%ax // no +X offset difference per function, only change FAR_ADJ definition with small vs medium model
 	mov    ARG1+FAR_ADJ(%bp),%bx
        ...
```

ghaerr:
> It is for 320x200, 4 color mode (CGA)
but most ibm compatible pc supports this mode, doesn't it?

I uploaded snakecga.bas and started @tyama501's improved `basic snakecga` on QEMU, and it operates correctly :)

@tyama501, you're welcome to upload your final version of snakecga.bas into elkscmd/basic so that others can play with this too, thank you.

elkscmd/basic/asm-ibmpc.S
tyama501: Hi @ghaerr 

Yes, assembling from C uses bp
but I referred to fmemsetw and used the di.
Let's fix it in the future. 

Thank you!

PR #2004, opened 2024-09-12
[build] Use CI actions/upload-artifacts@v4

ghaerr:
Updates CI to use v4 as Github has deprecated v1. This allows the automated CI to continuing operating during push commits to the repository. I received notification of the previous failure due to deprecation when committing #2003.

@tyama501, on PR #2003, did you happen to notice whether the Gitub CI ran the test(s) when you pushed each commit to the PR? I noticed that it did not run on my end as I thought it was supposed to. If it did not run, then I will reenable the 'pull-request:' section of main.yml. I had thought that the 'push:' section would work for non-maintainer pushes as part of a PR, like they do for me.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,103 +45,103 @@ jobs:
         run: ./build.sh auto allimages
 
       - name: upload
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v4
         with:
           name: fd1440.img
           path: image/fd1440.img
 
       - name: upload2
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v4
         with:
           name: fd360-minix.img
           path: image/fd360-minix.img
 
       - name: upload3
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v4
         with:
           name: fd720-minix.img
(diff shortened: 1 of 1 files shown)


tyama501:
I got CI error mail every time I pushes #2003.
I didn't know why it happens.

Thank you.

ghaerr:
> I got CI error mail every time I pushes https://github.com/ghaerr/elks/pull/2003.

Thanks. The reason for that is that Github deprecated some of our CI actions and failed the test run(s). This PR fixes that.

Should you see errors in the future, please post on the corresponding PR so problem can be investigated.

For your PR #2007, are you getting a valid completed test result? On my screen, it shows Merge button but no tests having run. If you are getting a test run and not me, that means I likely have to turn back on the second "PR" test run in main.yml, in addition to the "Push" test run.

tyama501:
I don't know much about CI actions but here is my screen.
image

ghaerr:
Thank you @tyama501. What is happening is that the contributor is getting the notification that their push commit(s) are compiling (or not), which is very useful. What stopped happening is for the administrator or others to see whether the commit checks are working. I will add back in the CI action for "PR" which is the way it used to work. I was just trying to save an extra 3+ minutes each time a commit is pushed. Oh well!

PR #2007, opened 2024-09-13
[basic cga] Add snakcga.bas

tyama501:
Porting from www.quitebasic.com/prj/games/snake
for ELKS BASIC CGA

Hello @ghaerr 

Now, I think it is playable.

BTW
It seems that the file name before the .bas need to be 7 charactors or less.
load "snakecga" for snakecga.bas didn't work.
So, I used the name snakcga.bas.
load "snakcga" works.

Thank you.

snakcga_start

snakcga


ghaerr:
> Now, I think it is playable.

Nice, can't wait to try it. What happened to the square game box that used to display? It appears with your latest screenshot that the full height of the screen is not used anymore?

> It seems that the file name before the .bas need to be 7 charactors or less.

Are you running FAT filesystem, or MINIX? There is an issue with ELKS on FAT for creating files > 8 characters, but 7 characters is possibly something new/different. Overall kind of strange.

For further testing, I renamed "snakecga.bas" to "thisislongfile.bas". This ends up creating a file "testislongfile" on both MINIX and FAT (I think because `mfs` enforces the ELKS 14-character max filename). I then created a MINIX image and everything worked fine running `basic thisislongfile`. However, when I ran the same on FAT, the system crashed!!!

Running `vi thisislongfile` or `cat thisislongfile` worked, only basic is crashing.

Just doing more testing: `cat testtesttest` says No such file or directory. But `cat testtesttest.xxx` crashes ELKS! I've added this to my bug list, but this still doesn't answer why your snakecga.bas had to be renamed - that works on my system. Can you please retest and tell what the result is?

tyama501:
Previously, the square was just a temporary to fit it in the 320x200 and it is not matched to the arrays.
Now the array is reduced from 25x25 of PC-98 640x400 to 25x13 for the 200 line and the box is matched to that.

Thank you for the testing.
I am running on FAT.

I mean using load command on basic.
It says Bad parameter if it is snakecga.

filename_snakecga

ghaerr:
I see now: basic will load snakecga.bas by `basic snakecga`, but using `load snakecga` inside BASIC throws a Bad Parameter error.

I have tracked this down to the following code in basic.c::parseLoadSaveCmd():
```
            char fileName[MAX_PATH_LEN];
            if (strlen(stackGetStr()) >= MAX_IDENT_LEN)
                return ERROR_BAD_PARAMETER;
```
Can you change that code to:
```
            char fileName[MAX_PATH_LEN];
            if (strlen(stackGetStr()) > MAX_PATH_LEN-1)
                return ERROR_BAD_PARAMETER;
```
and check that that solves the problem? That will then allow you to rename back to snakecga.bas with this PR. Please also include the above fix, as that will save a separate PR for me to do.

Thank you!

tyama501:
Thank you @ghaerr ,

I need to go traveling this weekend, so I will try it early next week.

ghaerr:
> I need to go traveling this weekend, so I will try it early next week.

That's OK - I will commit now so you don't have to worry about it. I will fix it in subsequent PR likely with the crash problem we just found.

Thank you!

PR #2020, opened 2024-09-16
makefile: rm -f -> $(RM)

marcin-laszewski:
Simple makefile upgrade: rm -f -> $(RM)

### DIFF
--- a/elkscmd/advent/Makefile
+++ b/elkscmd/advent/Makefile
@@ -43,4 +43,4 @@ install: $(PRGS)
 	$(INSTALL) advent.db $(DESTDIR)/lib
 
 clean:
-	rm -f $(PRGS) $(HOSTPRGS) *.o *.map
+	$(RM) $(PRGS) $(HOSTPRGS) *.o *.map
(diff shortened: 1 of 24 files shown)


ghaerr:
I'm not sure if this change actually helps, or instead possibly introduces more risk, into the ELKS build process. The make RM variable [is not required to be defined by POSIX](https://stackoverflow.com/questions/40907379/makefile-syntax-what-is-rm), and is not defined in our repo. It is entirely unused, except for I notice within the libc/ build. So a portion of the repo is already dependent on it, although that could be easily remedied by defining the RM variable in libc/Makefile. 

On the other hand, [GNU Make](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html) does define the variable by default to `rm -f`. For systems not running GNU make, the variable could be defined elkscmd/Makefile.defs, which would eliminate the risk for elkscmd/*/Makefiles.

Have you tested a full development system rebuild, and on which hosts? In particular I would be hesitant to change this in tools/Makefile. This change would also trigger half-hour automated CI rebuilds for the entire development toolchain.

ghaerr:
Completed via #2028.

PR #2021, opened 2024-09-16
[cmds] Fix off-by-one bug in getty @T command

vkoskiv:
The `@T` command, when placed in /etc/issue, displays the current system time. The shuffling of the date string in `when()` had an off-by-one error, which meant that instead of showing `17:42:00`, it showed `77:42:00`


ghaerr:
Nice catch, and I have been wondering when someone might actually use some of the @ variables in /etc/issue!

PR #2022, opened 2024-09-16
[kernel] Small cleanups, add disasm to 1440k distribution

ghaerr:
Changed `bios_park_disk_all` to do nothing for PC-98. @tyama501, you're welcome to add functionality for PC-98, but unless the PC-98 is running ancient MFM drives, there is no need for auto-park since all new drives are supposed to park automatically.

Removes some unused init code.


tyama501:
Thank you @ghaerr ,

PC-98 has SEEK function but my 9801BX is broken and
I don't have the real hard disk to test.
Other PC-98 I have uses SCSI emulator.

So, it is ok for now.

PR #2063, opened 2024-10-06
[kernel] Fix BIOS track read retry errors under QEMU

### LINKED ISSUE #1119: BIOS track read retry errors when floppy sectors/track differ
As can be seen from the screenshot below, the floppy BIOSHD driver ends up doing a track read retry whenever switching between floppy disks with different sectors per track (e.g. /dev/fd0 1440K and /dev/fd1 720K). This screenshot is using the mount.cfg auto-mount to mount /dev/fd1 at boot time and gets the track read retry error when switching back to reading the boot floppy:
Screen Shot 2022-01-19 at 10 55 59 AM

I'm not sure if this is a new problem with QEMU's BIOS emulation, or happens on real hardware.

Also, BIOSHD is querying the floppy drive type from BIOS and getting the wrong result for 360K floppies. Shown below, it is reporting a 360K /dev/fd1 as 80 cylinders, 2 heads and 15 sectors 1200k. After the ELKS BIOSHD probe, the correct disk size is determined and shown as 40 cylinders, 2 heads and 9 sectors:
Screen Shot 2022-01-19 at 10 29 01 AM
Also not sure if this is just QEMU, or happens on real hardware.

Both these errors end up fixing themselves, but the false reporting of disk attributes is bothersome, and the track retry may have to be modified, as ELKS currently doesn't actually retry track reads, but instead falls back to single sector reads. Changing this to 2 retries also fixes the problem, but not sure whether this is a QEMU-only problem or happens on real hardware. A disk reset causes the entire BIOSHD function to fail, for some reason.

It is not clear how long this behavior has been occurring.

ghaerr:
Fixes #1119.

While this fix has been coded using `RESET_DISK_CHG` for some time in bios.c, a test for running QEMU now makes the fix permanent, with no further `bioshd(0) track read retry` errors being displayed on the console.

Discussed recently in https://github.com/Mellvik/TLVC/pull/88#issuecomment-2386895898, this issue has been previously tested by @Mellvik on real hardware and was found to be a QEMU BIOS issue only. On QEMU, a BIOS disk reset is required when changing I/O (single or multi-sector) between floppy drives, for reasons unknown. This fix will allow for upcoming performance testing on QEMU without being concerned about track or sector I/O retries by automatically performing the required BIOS disk reset when needed.

The `IODELAY` option in bios.c is now enhanced to simulate realistic 1440k and 360k drive delays under QEMU (default OFF). A 1440k drive is assumed to be spinning at 300rpm (200ms/revolution) and 360k at 360rpm (167ms/revolution). A half-revolution delay plus additional time for the number of sectors read/written (~10ms/sector for 1440k and ~20ms for 360k) is performed, simulating average-case real hardware. Soon, this will be compared with the `jiffies` calculations @Mellvik is performing on real hardware in https://github.com/Mellvik/TLVC/pull/88 for comparison and subsequent development of faster-booting disk images.

@Mellvik, feel free to comment on whether my floppy timings make sense for the two drives being simulated.

### DIFF
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -12,10 +12,13 @@
 #include <linuxmt/kernel.h>
 #include <linuxmt/kdev_t.h>
 #include <linuxmt/debug.h>
+#include <linuxmt/sched.h>
 #include <arch/system.h>
+#include <arch/param.h>
 #include "bioshd.h"
 
-#define RESET_DISK_CHG  0       /* =1 to reset BIOS on drive change fixes QEMU retry */
+#define RESET_DISK_CHG  1       /* =1 to reset BIOS on drive change fixes QEMU retry */
+#define IODELAY         0       /* emulate delay for floppy on QEMU */
 
 /*
  * Indices for fd_types array. Note these match the value returned
@@ -105,7 +108,7 @@ int BFPROC bios_disk_rw(unsigned cmd, unsigned num_sectors, unsigned drive,
 
 #if RESET_DISK_CHG
     static unsigned last = 0;
-    if (drive != last) {
+    if (running_qemu && drive != last) {
         bios_disk_reset(1); /* fixes QEMU retry when switching drive types #1119 */
         last = drive;
     }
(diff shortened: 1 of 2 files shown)


Mellvik:
This looks good, @ghaerr. IIRC 3.5in drives are always 300rpm, 5.25in AT drives are always 360 rpm. Older 360k drives are 300rpm too, but not relevant to QEMU.
Looking forward to numbers. Will be really interesting to see how they add up and whether timings can be reliable on QEMU.

Issue #2062, opened 2024-10-04
/bootopts on FAT hard disk


tyama501:
Hello @ghaerr 

I could update my PC-9801RX hard disk (/dev/hdc1) to 0.8.0 (+bug fix).

I took following steps, and it could boot,
but I noticed the bootopts file cannot be recognized.

sys /dev/hdc1
makeboot -f fat.bin (since it needs special boot code)
mount /dev/hdc1 /mnt
cp /bootopts /mnt/

I haven't used bootopts before on the hard disk, so I don't know it could recognize before or not.
I can open the bootopts file and read it.

Where the bootopts should be on the hard disk?
Is there some FAT limitation like it should be within 8 entries? 

![bootopts_on_hd_20241005_032825](https://github.com/user-attachments/assets/6bc3b16f-1bda-4111-a390-57121cfd4768)

Thank you.

ghaerr:
Hello @tyama501,

> Where the bootopts should be on the hard disk?
Is there some FAT limitation like it should be within 8 entries?

Yes, setup.S line 1041 shows that /bootopts needs to be in the first 8 root directory entries on FAT disks (only). However, looking through image/Make.image, I don't see anywhere where this is enforced, including in bin/sys. So that is likely the problem.

In order to check, you can use `ls -laU /mnt` to view the unsorted root directory on your mounted hard drive. Remember that /dev is faked and is not actually present. You may have to delete /tmp or another one of the first files/directories, copy /bootopts, then mkdir the deleted directory again.

Thank you!

tyama501:
I see. Thank you.

I will try it later.
Is there any reason sys does not copy the bootopts?
I think it can avoid issue it copies bootopts after copying LINUX before copying applications.
(If the FAT is initially formatted)

toncho11:
So @tyama501 you tried the installation on an existing ELKS installation (not a formatted disk) and this resulted in /bootopts not being in the top 8 entries?

ghaerr:
I think `sys` should copy /bootopts, but does not since it might not exist. I will look into that change.

tyama501:
Hello @toncho11 

I formatted it but currently the sys does not copy bootopts so I copied it after the install.

tyama501:
It seems that my power supply on PC-98 is broken, and I cannot test with it anymore...
Probably it will occur on the emulator too.

toncho11:
I test on real hardware with MINIX and it is OK.
[...]

PR #2066, opened 2024-10-07
[cmds] Fix sys and makeboot to copy /bootopts and updated required files

ghaerr:
Fixes the initial problem reported by @tyama501 in #2062. `makeboot` now creates /bootopts at the same time that /linux is copied, if it is present. Some other important programs /bin are also now included in the `sys -3` (minimal 360k sys) command.

Took some effort to keep the size of bin/sys within 3K for efficiency.

Tested on MINIX and FAT 360k filesystems on QEMU. 

Haven't looked into @toncho11's reported problem of using `sys` to an HD FAT partition yet (coming).

### DIFF
--- a/elkscmd/rootfs_template/bin/sys
+++ b/elkscmd/rootfs_template/bin/sys
@@ -1,19 +1,17 @@
 # sys - create bootable system
-#
-#set -x
 
 usage()
 {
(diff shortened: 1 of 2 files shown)


toncho11:
I updated my understanding on the issue: https://github.com/ghaerr/elks/issues/2065

PR #2067, opened 2024-10-07
[cmds] Fix mkfs to clear reserved sector corresponding to old MINIX superblock

### LINKED ISSUE #2065: Target and System filesystem must be same type
Continuing from: https://github.com/ghaerr/elks/issues/2062#issuecomment-2395592983

Problem on fat where I get the error: `Target and System filesystem must be same type`
It is here: https://github.com/ghaerr/elks/blob/f752204730e7521f8fcd718fcaf8937bb6301045/elkscmd/sys_utils/makeboot.c#L418

So filesystem type is checked by reading 512 bytes from disk. It is only checking for Minix. This is not transparent in my opinion. I was thinking it is checking partition IDs for example. Why are these 512 bytes not overwritten (with fdisk or mkfat)? Sometimes by chance they get overwritten and problem is hidden. I did fdisk with 'w', mkfat, sync and still I could not overwrite this 512b block ... it should not be a guessing job which step is supposed to overwrite it. In my case it is either fdisk or mkfat not overwriting these 512 bytes  or my new partition is somehow a bit after the old one leaving this 512 bytes untouched (where it was Minix indeed).

ghaerr:
Fixes #2065.

Bug verified and tested using QEMU using updated `mkfat`.

### DIFF
--- a/elkscmd/disk_utils/mkfat.c
+++ b/elkscmd/disk_utils/mkfat.c
@@ -451,6 +451,17 @@ static int fatfs_create_fsinfo_sector(uint32 sector_lba)
     return 1;
 }
 //-----------------------------------------------------------------------------
+// fatfs_zero_reserved_sector: Clear reserved sector corresponding to MINIX superblock
+//-----------------------------------------------------------------------------
+static int fatfs_zero_reserved_sector(uint32 sector_lba)
+{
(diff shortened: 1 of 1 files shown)


toncho11:
Thank you! It works.

PR #2069, opened 2024-10-07
[cmds] Have fdisk check for MBR-capable block device for safety

ghaerr:
Inspired by @Mellvik's https://github.com/Mellvik/TLVC/pull/89. This version checks the minor device number and requires specification of an MBR-capable block device.

Make executable smaller using updated tiny_printf.c.
Rewrites some error message to properly show errno number.
Reformats source code for readability.


toncho11:
I propose this commit code and build for version 8.1.

PR #2071, opened 2024-10-14
[build] Add various options to optimize boot startup times

ghaerr:
This PR sets the stage for measuring and improving boot startup times using the following options:
- Display total elapsed startup time at login: prompt (SHOW_STARTUP)
- Accurately emulate floppy disk I/O track/sector/seek read times on emulator (IODELAY)
- Manage placement of directories and files to improve boot times (Image.all)
- Read extra sector on track split blocks (TRACK_SPLIT_BLK)
- Read extra sector on split blocks (SPLIT_BLK)
- Display track cache hits CH, track reads TR, buffer misses BM, buffer hits L1, L2, and files accessed (DEBUG_CACHE)
- New `mfs addfs` option allowing for adding files separately from standard `mfs genfs -a` filesystem generation
- Creating certain /dev/ directory entries early

@Mellvik, this is the first pass at optimizations talked about in https://github.com/Mellvik/TLVC/pull/88#issuecomment-2408772055.

The IODELAY floppy delay emulation should now accurately emulate track and sector read and seek times for both 1440k and 360k floppies, although it currently assumes 1440k in drive 0 and 360k in drive 1, which is not finalized.

Testing is ongoing, with all options default turned OFF for now. However, early testing is showing a 10 second boot time being halved to just 4.8 seconds for 1440k floppies using `Image.all` to specify startup files and `Make.devices` optimization of /dev creation. Note image creation optimization is currently only implemented for MINIX filesystem generation. 360k floppy boots seem to be improved from 7.2s to 5.5s, although I think the measurements were using 1440k floppy delays. I am also finding that adding TRACK_SPLIT_BLK and SPLIT_BLK (BIOS driver only for now), is not actually helping boot times. More on all that coming. Tests are not running `net start` as I'm trying to compare 1440k and 360k floppy boots at the moment.

The `image/Image.all` file contains a list of directories and files that when turned on via uncommenting the following line in image/Make.images causes the almost 50% speed improvement:
```
#   mfs -v $(VERBOSE) $(TARGET_FILE) addfs Image.all $(DESTDIR)
```

This mechanism works by seperating the previous `mfs genfs` operation into three - `mfs mkfs`, `mfs addfs`, and `mfs genfs -a`, to create, add optimized files, and add all files, respectively to the images.

The eventual implementation will allow for separately-configurable directory and files lists for floppy images; the current `Image.all` will work for any (all) floppy images:
```
# boot
dev 
etc 
bin 
bin/init
etc/inittab
etc/rc.sys
bin/sh 
etc/profile
bin/clock
etc/mount.cfg
bin/date
bin/getty
etc/issue
#etc/net.cfg 
bootopts
linux 
# login
bin/login
etc/passwd
```


Mellvik:
This is great, @ghaerr. I'm looking forward to check out the optimized layouts on physical hardware and compare to the numbers reported before. Our 'endless' (and always interesting) discussions on the matter are delivering real benefits. The `mfs` changes add very welcome flexibility for further experimentation. Thanks.

PR #2073, opened 2024-10-15
[boot] Decrease boot time from 10 to 5 seconds

ghaerr:
On QEMU emulating floppy disk delays, boot time for MINIX images is halved from 10 to 5 seconds.

Adds `image/Image.all` as first directories and files to all MINIX images.
Creates frequently-accessed /dev entries first for less directory searching.
Decrease size of /bin/getty by turning off @-sequence processing by default.
Renames SHOW_STARTUP option to BOOT_TIMER in getty.c.
Cleanup and retab getty.c source.

### DIFF
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -18,7 +18,7 @@
 #include "bioshd.h"
 
 #define RESET_DISK_CHG  1       /* =1 to reset BIOS on drive change fixes QEMU retry */
-#define IODELAY         0       /* emulate delay for floppy on QEMU */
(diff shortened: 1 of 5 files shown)


toncho11:
So it is faster when booting from floppy on minix ? But untested on real hardware?

ghaerr:
> So it is faster when booting from floppy on minix ? But untested on real hardware?

Yes, quite a bit faster. But agreed the system should have more testing, and especially on real hardware. I'll cut a v0.8.1 without it, as discussed in https://github.com/ghaerr/elks/pull/2075#issuecomment-2415652249.

PR #2075, opened 2024-10-15
[docs,build] Update build.sh to stop on error, update docs on mtools, flex, bison

ghaerr:
Enhance `./build.sh` to stop on error (discussed in https://github.com/ghaerr/elks/issues/2042).

Update documentation on mtools, flex, bison (not needed).

Cleans up atoi.c, atol.c.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -3,12 +3,9 @@
 ## Prerequisites
 
 To build ELKS, you need a development environment on Linux or macOS or Windows with [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux), including:
-- flex
-- bison
-- texinfo
 - libncurses5-dev
-- mtools-4.0.23 (for MSDOS/FAT images)
 - compress (for compressed man pages; use `sudo apt-get install ncompress`)
+- texinfo
 
 ## Build steps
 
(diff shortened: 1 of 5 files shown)


toncho11:
Maybe releasing version 8.1 before these boot optimizations is a good idea?

ghaerr:
Why not include the boot layout changes? Going backwards and producing binaries to upload is a lot of work. We might be better off with more frequent releases anyways if I didn't worry about organizing all the changes in a ChangeLog each time. The build process could probably be automated on my end and let Github produce the ChangeLog (I don't prefer its output, but its automated).

With the introduction of #2074 yesterday, it may be possible for Github to automatically build all images with the CI, although that would slow down the individual PR submission/acceptance by quite a bit, which is already 3mins * 2 = 6 mins. (Sounds small but when you have 4-5 PRs in queue, one sits around waiting for quite a bit).

There is no code difference, only the `mfs` program was enhanced to allow specification of the first-used programs to be laid out on the image first. Having floppy boot time cut in half is huge, if it were not included in v0.8.1, then why not just release a v0.8.2 with it immediately afterwards? (There is not any actual boot image optimization code compiled in by default, only measurement tools default OFF).

I was also hoping to solve the problem of not being able to produce PC-98 images on Debian 12.7, for which #2074 is also a fix.

toncho11:
In my mind it is about what is tested. New stuff is untested. For me, even a single line of change can produce a bug.

ghaerr:
> In my mind it is about what is tested. New stuff is untested. For me, even a single line of change can produce a bug.

Agreed, you're right. I've turned off boot optimization in #2078, in preparation for a cut of v0.8.1 without any major changes.

@tyama501, are you OK with cutting v0.8.1 now?

tyama501:
Quick tested v0.8.1 released pc-98 image on the emulator.
Looks great. Thanks!

PR #2099, opened 2024-11-11
[libc] Add abs() function

ghaerr:
Needed for @tyama501's space invaders port in https://github.com/ghaerr/elks/discussions/2093.


tyama501:
Thank you.

I think this also will be solved by this PR.
https://github.com/ghaerr/elks/discussions/1505#discussioncomment-10041160

PR #2101, opened 2024-11-13
[tui] Import curses invaders game from FUZIX repository

tyama501:
Hello @ghaerr 

This is the PR for adding the invaders game in
https://github.com/ghaerr/elks/discussions/2093

I haven't modified the Applications file yet.
Which image should we include this?

Thank you.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -180,6 +180,7 @@ tui/sl                          :tui                    :1200k
 tui/ttyclock                    :tui            :360c   :1200k
 tui/ttypong                     :tui                            :1440k
 tui/ttytetris                   :tui            :360c
+tui/invaders                    :tui                    :1232k  :1440k
 busyelks/busyelks               :busyelks
 inet/httpd/sample_index.html ::var/www/index.html :net
 ktcp/ktcp                       :net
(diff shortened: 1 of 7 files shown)


ghaerr:
Hello @tyama501,

This looks great, thank you!

> Which image should we include this?

I don't think there's much room left on the smaller images, I'm thinking perhaps the same as nxlandmine or nxworld: 1232k and 1440k or 1232c and 1440k? One of the 1232 options should allow it to be present on the distribution PC-98 1232k floppy.

tyama501:
I have added 
tui/invaders                    :tui                    :1232k  :1440k

It's size is about 10kB for compressed and 15kB for uncompressed.
I confirmed the uncompressed file fits in minix 1440k ibm image,
but it is 98% full, so please modify it if necessary.

Thank you.

ghaerr:
Thank you! 10-15k does seem pretty large for an ascii text game, but it can be possibly moved later as we get closer to another release and see what else is contributed. I'll commit this now and then try playing the game!

ghaerr:
It works great!
Screen Shot 2024-11-14 at 8 47 30 AM

My missile firer at the bottom doesn't seem to appear until I hit space or j/k a few times though. Not sure if that's supposed to be the way it works or not.

tyama501:
Thank you.

Yes, I also noticed that.
It happens on pc-98 emu too.

PR #2113, opened 2024-11-21
[kernel] Fix kernel sys_signal handling on multi-code segment executables

ghaerr:
Should fix "SIGNAL sys_signal supplied handler is bad" from https://github.com/ghaerr/elks/issues/1443#issuecomment-2489091235.

The kernel was not properly accepting signal handlers from executables that had multiple code segments - medium model IA16 or OWC large/medium model executables. 

Tested on QEMU for large model OWC executables, but not for multi-code segment large model OWC, which C86 appears to be. @rafael2k, if the message disappears, then the problem is fixed and signal handling should work as expected.


rafael2k:
Thanks @ghaerr! It fixed the issue.

PR #2119, opened 2024-11-29
[tools] Add 'ecc' ELKS C86 compiler driver script

ghaerr:
Adds easy-to-use `ecc` compiler driver for use with @rafael2k's 8086-toolchain which is being discussed in #2112 and requires the changes in https://github.com/rafael2k/8086-toolchain/pull/5.

The initial primary purpose of this script is to allow for quick test preprocessing, compilation, assembly and link of a single .c file to test the C86 toolchain without having to write a Makefile.

For instance, with a complete foo.c program in 8086-toolchain, the files foo.i, foo.asm, foo.o and foo will be created by:
```
cd 8086-toolchain
ecc foo
```
The ecc script lives in ELKS $TOPDIR/elks/tools/bin/ecc. It must first be edited to contain the top directories of the ELKS and C86 installation before it will work. This is documented in the script itself.

Also, currently the script requires running in the 8086-toolchain directory. This is because it assembles the lib86.a library from source each time. This will be automated in the 8086-toolchain master make shortly.

### DIFF
--- a/elks/tools/bin/ecc
+++ b/elks/tools/bin/ecc
@@ -1,0 +1,81 @@
+# ecc - ELKS cc compiler wrapper for C86 (8086-toolchain)
+#
+# This script runs cpp86, c86, nasm and ld86 on the input file.
+# The input file must be specified without its .c extension for now,
+# and the script must be run in the 8086-toolchain/ directory,
+# as it also builds the libc86.a C86 library.
+# This restriction will be removed shortly.
+#
+# Usage: to compile and link foo.c:
+#   cd 8086-toolchain
+#   ecc foo
+#
+# Before using, modify the full paths below to ELKS and C86:
+
+#####################################################################
+
+# Full path to ELKS and C86 installations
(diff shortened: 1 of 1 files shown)


toncho11:
This is is only for cross compile on the host OS?

rafael2k:
While I did not test inside ELKS yet (will do it soon), it should work for both foreign and native.

ghaerr:
The plan is to write a C-based compiler driver that will run on both systems, but its a bit early in the game for that, as we haven't really settled down on a full toolchain flow and there are still lots of issues.

The current `ecc` script is overkill for ELKS and may introduce too much memory overhead, as it involves loading the shell into memory for execution, which is large. There's no header files and no C library available and not much that can be done yet on ELKS anyways. Nonetheless, if you want to try it, I would suggest just using the bottom half of the file, and adding the toolchain to /bin, since the script uses some constructs ASH doesn't support, like `if [[ ...`.

PR #2120, opened 2024-11-29
[c86] Allow C86 to compile using most all ELKS libc header files

ghaerr:
With the recent enhancements made to the [C86 toolchain](https://github.com/rafael2k/8086-toolchain) and the [ELKS C86 compiler driver](https://github.com/ghaerr/elks/pull/2119) `ecc`, most all the ELKS C library header files are now acceptable input to C86, making a major step forward in moving down the path with the next step of beginning to compile the ELKS C library and bringing a toolchain to ELKS.

> which header should I include to use the system calls implemented in the syscall.asm currently in 8086-toolchat repo?

You can now use the standard C headers to access system calls with the 8086 toolchain. I have currently tested the following include files which are accepted:
```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <fcntl.h>
#include <ctype.h>
#include <assert.h>
#include <errno.h>
#include <signal.h>
#include <termios.h>
#include <time.h>
#include <sys/ioctl.h>
#include <sys/stat.h>
```

### DIFF
--- a/elks/include/arch/cdefs.h
+++ b/elks/include/arch/cdefs.h
@@ -41,5 +41,19 @@
 #define __wcnear        __near
 #endif
 
+#ifdef __C86__
+#define __HAS_NO_FLOATS__
+#define __far
+#define noreturn
+#define stdcall
+#define restrict
+#define printfesque(n)
+#define noinstrument
+#define CONSTRUCTOR(fn,pri) void fn(void)
+#define DESTRUCTOR(fn,pri)  void fn(void)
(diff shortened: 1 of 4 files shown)


ghaerr:
Oops @rafael2k, I need to add stdarg.h which I've just written over to 8086-toolchain in order for this to work. I'll submit a PR shorty.

rafael2k:
Yay! Will test as soon as you merge!

PR #2128, opened 2024-12-08
[kernel,libc] Rewrite kernel brk/sbrk and stack_check, fix C library malloc

ghaerr:
Rewrites sys_brk, sys_sbrk and stack_check with better error checking and new messages for heap and stack problems. 

Also fixes a somewhat major problem with C library malloc which caused an extra 1024 bytes to be allocated from the heap for any allocation request >= 1024, which effectively doubled the size of the heap necessary to fulfill the request! 

The C library malloc routine now allocates all requests less than 32 bytes as 32 bytes from the heap for later subdivision, but all larger requests without chunking up to a large (previously 1024 byte) chunk size. In addition, when memory is low, malloc doesn't retry three times which produced multiple failure messages, usually right before a stack overflow message. That has been fixed for the most part, except for when an application uses large stack buffers without stack checking turned on (more coming on that in next steps).

Removes unused CONFIG_EXEC_LOW_STACK option from loader.

These changes were you used to test the 8086 toolchain and resulted in PR https://github.com/rafael2k/8086-toolchain/pull/17.

Now, when applications run out of heap or stack, or the stack pointer moves into the unused heap area, more informative messages are produced by the kernel, to let the developer know what is going on. 

Following are the changed error messages:

When the stack usage grows outside (below) its normally reserved and protected area into the heap, the following messages are produced:

"STACK USING 1131 UNUSED HEAP" when the stack isn't large enough and unused heap area is used, or
"STACK OVERFLOW" when the stack grew into an inuse area of the heap. In this case the process is also terminated.

When a request is made by malloc to get more memory from the heap, and that fails, the message

"SBRK 1024 FAIL, OUT OF HEAP SPACE" is displayed, and malloc returns NULL.

If the stack has already grown into the unused heap, but a malloc allocation request wants some of that heap area, the message 

"SBRK %d FAIL, WOULD OVERWRITE STACK" is displayed.

This may all sound complicated, but in the end more information is displayed, and when the messages are displayed, it gives better information as to whether the program's heap or stack should be increased. This was definitely needed for our upcoming C86 toolchain, as some tools require maximum heap or stack space.

To see detailed SBRK/BRK allocation information on the console, set `debug=1` in /bootopts.

### DIFF
--- a/elks/arch/i86/kernel/process.c
+++ b/elks/arch/i86/kernel/process.c
@@ -43,27 +43,27 @@ int run_init_process_sptr(const char *cmd, char *sptr, int slen)
 }
 
 /*
- * We only need to do this as long as we support old format binaries
- * that grow stack and heap towards each other
+ * Check that SP is within proper range, called before every syscall.
  */
 void stack_check(void)
 {
-    segoff_t end = current->t_endbrk;
+    segoff_t sp = current->t_regs.sp;
+    segoff_t brk = current->t_endbrk;
+    segoff_t stacklow = current->t_begstack - current->t_minstack;
 
-    {
-        /* optional: check stack over min stack*/
-        if (current->t_regs.sp < current->t_begstack - current->t_minstack) {
-          if (current->t_minstack)     /* display if protected stack*/
-            printk("(%P)STACK OUTSIDE PROTECTED LIMIT by %u\n",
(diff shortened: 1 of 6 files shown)


rafael2k:
This really helps to develop the toolchain. Thanks!

PR #2133, opened 2024-12-14
[build] Create 64M MBR HD image with 'make images'

ghaerr:
Adds hd64mbr-minix.img to build.

Requested by @rafael2k. Not tested on real hardware, only on QEMU.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,5 +143,11 @@ jobs:
       - name: upload17
         uses: actions/upload-artifact@v4
         with:
+          name: hd64mbr-minix.img
+          path: image/hd64mbr-minix.img
+
+      - name: upload18
+        uses: actions/upload-artifact@v4
+        with:
           name: fd1232.img
           path: image/fd1232.img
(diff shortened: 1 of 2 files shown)


rafael2k:
Thanks! Will test on real hardware.

rafael2k:
I think there is a problem here. I'm not sure how the partition table is created, but it has the same size of the 32M image, instead of the 64M one.

ghaerr:
>  I'm not sure how the partition table is created, but it has the same size of the 32M image, instead of the 64M one.

Thanks for testing! Let me look into this, I ran into some issues while creating this and thought it was a correct, but I think you're right. Are you getting the partition information from `fdisk`, which is showing the same 32M size for both?

rafael2k:
> > I'm not sure how the partition table is created, but it has the same size of the 32M image, instead of the 64M one.
> 
> Thanks for testing! Let me look into this, I ran into some issues while creating this and thought it was a correct, but I think you're right. Are you getting the partition information from `fdisk`, which is showing the same 32M size for both?

Yes, getting from fdisk.

ghaerr:
Ok, sorry. My original 64M image wouldn't boot from MBR, and I thought I fixed it. Now I know why - setboot isn't changing the MBR properly. I'll post a fix.

PR #2136, opened 2024-12-16
fall back to 8x16 font if 8x14 not present

rafael2k:
There is still a printf I need to remove. Just to get a feedback if this is the way to go.


ghaerr:
Yes, looks good. If this works on both your systems, we should be good to commit with printf removed.

Also - romfont.c is using the typical "K&R" style braces, not the Whitesmith's style you're using. Please update the braces also, thanks.

rafael2k:
I tested here with qemu BIOS (old-style pre-VBE 3.0 standard) and with my recent x86 with VBE 3.0+ VGA BIOS, and all good. Fatty fonts on recent PCs, while the default does not change on anything pre-VBE 3.0 VGA equipped PCs.

rafael2k:
Braces updated. I'm not sure if I should remove the braces from the else case.

rafael2k:
I'll open another PR. I think I messed up with git here.

ghaerr:
> I'll open another PR. 

Yes, looks like you might not have done a git pull before starting. I can sometimes fix that, but this looks a bit beyond my git capabilities.

> I'm not sure if I should remove the braces from the else case.

You're welcome to keep or remove them, but have the else case on the same line as the previous end brace:
`} else {` or `} else`. In these cases its probably better to keep the braces on the else case.

rafael2k:
got it, thanks. Will PR again soon.

PR #2141, opened 2024-12-17
Adding new rom fonts to romfont.h driver in nano-X

rafael2k:
This is in preparation of what I think is to expose font selection through env variable, or something like this, without breaking nothing.


ghaerr:
Did you want to commit this, or are you just holding this open while you experiment with a solution?

rafael2k:
I think for now what we have is safe, and will do the font fallback fine if needed.

Issue #2166, opened 2024-12-31
Can't create, rename or delete long filenames on FAT disks


toncho11:
These were discussed in: https://github.com/ghaerr/elks/issues/2159#issuecomment-2565962085
List:
* there is a problem with the ld linker from the 8086-toolchain. On Minix the toolchain works perfectly, but on FAT it fails
* handling of long filenames in FAT, currently files longer than 8 characters can not be deleted
* FAT is slower than Minix

We should make a difference between FAT16 and FAT32.
I also found this: https://serverfault.com/questions/529786/fat32-limit-on-total-length-of-all-filenames-in-a-directory-combined
It seems people on Windows also have problems. It says: 
> Now, obviously from observation, there seems to be some kind of hardlimit on how many characters can be used for all filenames inside a single directory for FAT32

Maybe it is the support for long filenames in that FAT is causing both the first and second problem.

toncho11:
For example if the hard limit for long filenames on ELKS is much lower then on the Windows implementation? So we need to know what is the number of filename characters per folder for FAT32 for ELKS.

floriangit:
> FAT is slower than Minix

This was my main reason to use Minix, FAT was (not measured, but guessed) 50% slower compared to Minix on real hardware.

ghaerr:
Thanks for opening this issue @toncho11. I suppose you weren't going to let the toolchain on FAT issue be tossed onto the back burner, since your primary real hardware testing is on FAT, right? :)

> there is a problem with the ld linker from the 8086-toolchain. 

I've identified exactly where the "upper level" of this problem occurs in the FAT filesystem: LD86 writes out the chess a.out header, then seeks forward ~6k bytes to a later section in the codefile. Upon trying to write that second section, the kernel goes into a loop with the FAT filesystem allocating all remaining FAT disk, but never attaching it to the file. The LD86 finally return with an out of space write error, and then we have a secondary problem that the toolchain stack sometimes overflows and/or segfaults when the filesystem has run out of space.

I worked on the FAT filesystem part of this problem for way too many hours last night, and came away with nothing. Its a real bugger. I'm regrouping and will dive in again later with hopefully some better ideas.

> handling of long filenames in FAT, currently files longer than 8 characters can not be deleted

[...]

PR #2167, opened 2024-12-31
[kernel] Fix FAT filesystem adding blocks when seeking past EOF

ghaerr:
This PR fixes the recently found issue with the FAT filesystem not properly adding file blocks to a file when seeking past EOF and then writing. The 8086 toolchain linker LD86 does this and would not function properly on FAT filesystems.

Discussed in https://github.com/ghaerr/elks/issues/2159#issuecomment-2565962085 and issued opened with #2166.

The 8086 toolchain now runs properly on FAT and MINIX filesystems.


toncho11:
Wow! Good job! Impressive how fast you fixed that!

PR #2169, opened 2025-01-02
[libc, Watcom C] Fix gettimeofday, ctime and __amalloc debug speed

ghaerr:
This PR is the result of a multi-hour effort diving down a deep rabbit hole opened up by @toncho11's findings that AS86 was taking lots of time in its startup routines, discussed in https://github.com/ghaerr/elks/issues/2159#issuecomment-2567154036.

It turns out that the AS86 startup routines called the malloc memory allocation routine a large number of times, and this in turn led to finding that the real culprit was a bug in the debug \_\_amalloc routine, where the heap was being traced but not displayed, taking lots of execution time. This took a long time to figure out.

However, when turning debug off, the toolchain crashed when CPP86 executed - which took many more hours to figure out. Turns out this was an older bug that I swept under the rug when first found because I couldn't figure how to fix it. Many hours later, it was found that the reasons for the system crash was a buffer overflow in the C library `ctime` routine, and in addition the DS register was being set to 0 in the associated `gettimeofday` routine by the OpenWatcom system call wrapper which incorrectly assumed the first parameter would never be NULL. This set the stage for a crash much later on, when the compiler didn't reload DS and then proceeded to scribble all over memory outside the process. All this was finally found by tracing down CPP86's automatic setting of its internal `__TIME__` and `__DATE__` macros, which haven't even been used yet.

During the deep dive, it was also discovered, at least on QEMU, that a "make" of the toolchain examples runs at half or twice the speed of "time make" depending on the build, which is very strange. This could be QEMU, so it would be nice to see what happens on real hardware between the two invocations. 

As part of tracing that down, and also for more visibility by our testers, AS86 now temporarily displays the assembly pass number when running, with PR https://github.com/ghaerr/8086-toolchain/pull/28. 

That PR also removes the assembler -O switch, and reduces the passes required for chess.c from 5 to 2, at a cost of a few jump instructions not being shorted from 3 bytes to 2.

Other inclusions are adding stddef.h for OWC since the OW include files are no longer required for building OWC programs, including the 8086 toolchain.

The debug amalloc was also cleaned up.


toncho11:
Wow! Impressive! I bet these existed, but people just used the DEBUG without ever finding the real problem. You figured it all! :)

Issue #2157, opened 2024-12-28
Ideas for distributing multiple or larger ELKS disk images


ghaerr:
From the recent and increasing flurry of activity, it seems that ELKS has, (finally!), reached the point where all the good stuff won't fit on one 1440k floppy disk. I'm using the 2880k floppy image for testing the native C86 compiler and have run out of space on that on occasion, having to delete some programs in order to proceed.

Here are some of the larger categories of stuff that have recently become available:
- Doom
- More and more games
- Nano-X graphical programs
- More man pages
- Unit test routines and C library tester (test_libc, test_float, etc)
- C86 native compiler toolchain with all development header files and libraries (this alone won't fit on a 1440k disk)
- Other user-contributed stuff that might not be in the main elkscmd/ Applications file/repo.

I thought to ask the community what might be best - separate "packages" of floppy images perhaps, or a hard disk image with everything on it? Either has its sets of problems: an issue with multiple floppy packages is whether the package disk is bootable, or whether it's just an add-on that must be mounted in order to be accessed, exactly which programs are on it, etc. An HD image with everything on it would be great for, say, a full C86 devkit and/or all games, but only immediately usable with an emulator. Users with hard drives would have to go through more of a hassle to get a system up and running.

Since I primarily run from QEMU, I kind of like the idea of producing an HD image that has everything on it - this would allow for a single image to be used for everything ELKS. Perhaps someone could write a script that would then extract from that into a series of floppy images for later installation onto real hardware - floppies or install sets for a real hard drive. For real hardware, updating a system like this can take a lot of work, and shouldn't be normally performed unless the distribution is stable.

Since Doom and C86 now require the Open Watcom C compiler, there is the additional issue of having that installed and/or built, in order to build a full ELKS system, should one be buildable from the ELKS repo. Given that the ELKS C library will have to be used for both OpenWatcom programs, as well as C86 toolchain, we're probably already well headed in the direction of requiring OWC to be installed to build ELKS "extras" or a full HD distribution.

Thoughts?

floriangit:
For me personally, the only viable interface into my 286-PC is and remains the floppy image. Hence, I like the idea of multiple FD images, one bootable, and the other one with a pink EXTRA label ;). And I guess it would be a nice challenge to replace `bin/sys` finally with a proper installation program (and a proper name, I mean....why 'sys'?), that can handle an interactive "Please insert Disk 2" prompt like it was in good old times. With ASCII arts ... maybe. ;-)

[...]

PR #2181, opened 2025-01-12
Add -p option to mknod.c and use it with sys

floriangit:
This replaces https://github.com/ghaerr/elks/pull/2180 as per your suggestions. Using sys now even on a fully installed medium will now only output around 20 lines which fits to the old terminal sizes (80x25).

### DIFF
--- a/elkscmd/file_utils/mknod.c
+++ b/elkscmd/file_utils/mknod.c
@@ -13,6 +13,19 @@ int main(int argc, char **argv)
 	
 	newmode = 0666 & ~umask(0);
 	
+	if (argv[1] && argv[1][0] == '-' && argv[1][1] == 'p') {
+		/* preserve option */
+		struct stat sb;
+		if(stat(argv[2], &sb) == -1)
+			return 1;
+		if(S_ISCHR(sb.st_mode) && argv[3] && (argv[3][0] == 'c' || argv[3][0] == 'u')) return 1;
+		if(S_ISBLK(sb.st_mode)  && argv[3] && argv[3][0] == 'b') return 1;
+		if(S_ISFIFO(sb.st_mode) && argv[3] && argv[3][0] == 'p') return 1;
+		/* valid node not there yet, so we advance and create it */
+		argc--;
+		argv++;
+	}
+
 	if (argc == 5) {
 		switch(argv[2][0]) {
 		case 'b':
@@ -46,6 +59,6 @@ int main(int argc, char **argv)
 	return 0;
 
 usage:
-	errmsg("usage: mknod device [bcup] major minor\n");
+	errmsg("usage: mknod [-p] device [bcup] major minor\n");
 	return 1;
 }
--- a/elkscmd/file_utils/mknod.c
+++ b/elkscmd/file_utils/mknod.c
@@ -16,11 +16,11 @@ int main(int argc, char **argv)
 	if (argv[1] && argv[1][0] == '-' && argv[1][1] == 'p') {
 		/* preserve option */
 		struct stat sb;
-		if(stat(argv[2], &sb) == -1)
-			return 1;
-		if(S_ISCHR(sb.st_mode) && argv[3] && (argv[3][0] == 'c' || argv[3][0] == 'u')) return 1;
-		if(S_ISBLK(sb.st_mode)  && argv[3] && argv[3][0] == 'b') return 1;
-		if(S_ISFIFO(sb.st_mode) && argv[3] && argv[3][0] == 'p') return 1;
+		if(stat(argv[2], &sb) == 0) {
+			if(S_ISCHR(sb.st_mode) && argv[3] && (argv[3][0] == 'c' || argv[3][0] == 'u')) return 1;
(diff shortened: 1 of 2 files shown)


ghaerr:
Thanks @floriangit, well done!

On a small code formatting note, would you mind changing the new if statements on lines 19-22 to match the existing formatting convention of using a space after the if, to match the rest of the file? Thank you.

floriangit:
> well done!

thanks! Added the spaces now.

ghaerr:
Thanks for enhancing `sys` to work for updating as well as initial installation @floriangit!

floriangit:
one day I will rename `sys`...:P

ghaerr:
It was initially suggested by someone, saying DOS used "sys" ... Obviously we're not emulating DOS here. What possible other names are you thinking?

floriangit:
update, upgrade, install? After all it is now kind of an upgrader for me at least.

floriangit:
then again, I thought it would be too tiresome to rename that now, since one has to know overall ELKS. Because grepping for `sys` to find out where it's used is a very time-consuming thing, I guess.

ghaerr:
I kind of like 'install'. The community has been used to 'sys' for more than a few years, but I would agree 'sys' only means something to folks familiar with DOS. Renaming ideas should probably be mentioned on #2157.

floriangit:
Yes, I thought same. Since it's not (yet) anything like a self-hosted updater, it's still an installer involving two separate media, so install sounds good.

ghaerr:
@floriangit,

I just realized that I think we have the new return value for the mknod -p option incorrect: instead of return 1 for the three cases, I think it needs to be return 0 (which means OK). A return 0 is an exit(0) which is interpreted by the shell as OK. Return 1 will stop shell execution if stop on errors is set (off by default).

To test this, add `set -e` (stop on error) to bin/sys, and try updating. The mknod -p will return 1 and halt execution. And the set -e should stay in the shell script anyways, so that errors aren't ignored.

Please test and let me know what you find, thanks!

floriangit:
My line of thinking was that with -p returning 1 instead of 0: "The user asked me to preserve a file, but he also asked me to overwrite that file, so I indicate 1 instead of 0, so he has a chance to recognize my silent return".
Then again, I do understand your point (All is OK!) and looking at `mkdir -p` returning 0 when directory exists, I agree and will change the return's to 0.

As for `set -e` that's a bit of a mixed bag. In `sys` also makeboot is called, which does _not_ return 0 on success, but rather the FS type, since this is how it's designed. So I cannot globally add `set -e` in `sys` since that will stop the script at that point. I could do it after those lines, but I personally would not overrate `set -e` since it silently returns 1, but gives no warning whatsoever anyhow. The last lines of output of `sys` was:
`System copied`
so if I would not work on it right now, I would fail to see any error....

ghaerr:
> As for set -e that's a bit of a mixed bag. In sys also makeboot is called, which does not return 0 on success, but rather the FS type

Thanks for pointing that out, I forgot about that. 

My current take is that running shell scripts without set -e is unsafe, unless any error is not important. In our case, the user will definitely want to know whether there was success or not in installation. Thus, set -e must be set.

Next, the convention in shell scripts is a command returns "true" (no error) on exit/return 0, while "false" is anything else. A "false" return will stop the script execution if set -e is set. Thus our mknod -p must follow that convention.

HOWEVER, since you've rightly pointed out that our design requires `makeboot` to return an FSTYPE (yes, an earlier kluge since there was otherwise no way to get the mounted filesystem type), then I think we need to set +e before `makeboot` and set -e afterwards - at least until the makeboot returning FSTYPE design is enhanced using another method.

To your earlier points about writing a C program versus a shell script for installation or updating - our current method shows a weak point in the design which would be more easily handled in C. However, given the ease in scripting most of the solution, perhaps we should think of more/better ways to communicate between the kernel and shell scripts for installation issues, then take an overview design pass after we see what more might be required.

floriangit:
Hi @ghaerr,

Agreed first about `return 0` in mknod -p and `set -e` in sys.
I shortly tried to set it after makeboot but it seems to error out somewhere else (as said: alas `set -e` is silent except the error code with $?). Would need to debug.

Which made me step back. (Still agreeing on mknod -p to exit with 0):
`set -e` makes sense to safely execute scripts that can run silently and really need to succeed. But as `sys` is a standalone installer and we don't know whether the user checks its return value with `echo $?` after execution and as there is no error message it makes no sense as-is to me. It's not clear cut for a user. Wrapping `sys` with a `install` script that checks its return and cries "FAILED" is of course an option. Which would mean more bytes, heap and stack (your dilemma? haha). 

I would commit the return 0 for mknod -p next days and wait for thoughts.

The general discussion re. `sys` can continue in the separate thread. Other contributors are much more experienced so maybe something creative comes up there. I don't really mind "the name" that much, since I watched the Dilbert episode of "the name" ;-).

ghaerr:
> we don't know whether the user checks its return value with echo $? after execution and as there is no error message it makes no sense as-is to me. 

See #2186 - the issue isn't what the sys script returns to the user, but rather halting the sys script on any error, but also allowing `makeboot` to return a filesystem type as its return code, and not having that filesystem type stop script execution, but only for that single script statement.

PR #2186, opened 2025-01-16
[cmds] Update 'sys' and 'mknod' to work for updating ELKS

ghaerr:
Enhances the bin/sys shell script and mknod command to update an existing ELKS volume from the boot volume. 

Discussed in https://github.com/ghaerr/elks/pull/2181#issuecomment-2586083027.

@floriangit: I created this PR as there might have been a misunderstanding about the need for set -e/set +e in bin/sys.
(set -e tells the shell to stop execution on an error). This PR fixes `sys` so that any error will stop the script, *except* for the `makeboot` execution, which as you pointed out returns the filesystem type as it's return value. So a set +e is required only before `makeboot`, and otherwise set -e is always set. Should `makeboot` encounter an error, it returns 255, and sys will return 1 (error) to its caller. Otherwise the script now will report "Finished successfully." to the user.

`mknod` was changed to return 0 on successful -p option, as discussed.

Also fixed is "Can't create symlink: File Exists", which is reported by `cp -fR` when copying a full system and encountering the symlink already created on the destination volume. The fix is to unlink the symlink prior to creating it, when the -f option is in effect.

### DIFF
--- a/elkscmd/file_utils/cp.c
+++ b/elkscmd/file_utils/cp.c
@@ -428,6 +428,8 @@ int do_symlink(char *symlnk, char *file)
 {
 	if (opt_verbose) printf("Symlink %s -> %s\n", file, symlnk);
 
+	if (opt_force)
+		unlink(file);
 	if (symlink(symlnk, file) < 0) {
 		fprintf(stderr, "Can't create symlink "); fflush(stderr);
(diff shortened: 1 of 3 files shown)


floriangit:
very smooth updating now! :1st_place_medal:

PR #2201, opened 2025-01-27
[build] Create full development disk from 8086 toolchain on /usr

ghaerr:
The `copyc86.sh` script was updated to create a full 8086-toolchain development image under the /usr directory, which includes the ELKS native binaries, header files, libraries and examples source code. To create a development disk, run `./buildc86.sh` to build the 8086 toolchain (once). Then run `./copyc86.sh` after building a 2880k or HD system, and run make again to add /usr to the image. The full development system takes ~600k, and leaves only ~70 blocks free to do anything on a 2800k floppy. (Those numbers do not include the NASM/NDISASM hogs, which are no longer required).

After booting the ELKS image, one can type:
```
login: root
# cd /usr/src
# make
```
and all examples will build. The structure of the development disk should allow for it to be tarballed or zipped using just /usr. 
The following directories are used:
- /usr/bin - All ELKS toolchain executables
- /usr/include - All header files for the C library
- /usr/lib - Compiled libc86.a C library archive
- /usr/src - Examples directory with source. PATH now includes /usr/bin so typing "make" from here works

While building the development disk, several serious problems were found, where are also fixed in this commit. These include:
- The ability to read data into fmemalloc'd segments owned by the current process. This took a long time to figure out and in retrospect is amazing the toolchain even worked at all. Basically, the previous kernel read/write buffer system call verification code just happened to work, but only when the offsets returned from the `_fmalloc` memory manager happened to be low and seemed to fit within the process's default data segment, even though there were not.
- Solved the above issue by verifying addresses to be either within any of the task structure MAX_SEG segments, *or* within any of the process-allocated `fmemalloc` segments. The latter search is done by traversing the entire main memory allocation list for every process and every system call, which will be too slow. Working on a fast solution.
- Start the free space search in `_fmalloc` at the start of heap, rather than the faster method of starting from the last allocation. Starting at the start of heap is required for cpp86 in order to combine free segments and keep the heap from fragmenting, otherwise cpp86 runs out of memory when preprocessing lots of C header files. It turns out that there may be 8 or 9 levels deep of nested include files within the ELKS C library, and memory use is high during C preprocessing.
- Adds /usr/bin to PATH in /etc/profile to allow running toolchain binaries from /usr/bin without running a shell script first.


toncho11:
I am testing this. 

[tool_chain_user_basic1.zip](https://github.com/user-attachments/files/18563072/tool_chain_user_basic1.zip)

And I can not boot this image in https://copy.sh/v86/ Something is wrong with my setup. I am building MBR images ... and I should not when using https://copy.sh/v86/ 
Can you please have a look? The image should have the new /usr and basic.

toncho11:
It works with 86Box, but not with https://copy.sh/v86/.

toncho11:
Actually I am getting:

```
mformat -V
mformat (GNU mtools) 4.0.23
configured with the following options: enable-xdf disable-vold disable-new-vold disable-debug enable-raw-term
mformat -i /root/elks/image/PC.img -s 8 -h 2 -t 77 -M 1024 -d 2 -L 2 -r 6 -k -N 0 -c 1 -m 0xfe
# Linux has to be the first file for the boot sector loader
rm -f linux; touch linux
mcopy -i /root/elks/image/PC.img  linux ::/linux
rm linux
# Device folder has to be first or second for the 'fake dev' to work
mmd -i /root/elks/image/PC.img ::/dev
for f in $(cd /root/elks/target; find * -name '*'); do \
        if [ -d /root/elks/target/$f -a "$f" != "dev" ]; then mmd -D o -i /root/elks/image/PC.img ::$f || exit 1; fi; \
        if [ -f /root/elks/target/$f ]; then mcopy -i /root/elks/image/PC.img -pmQ -D o /root/elks/target/$f ::$f || exit 1; fi; \
done
Disk full
make[2]: *** [Make.image:123: fatfs] Error 1
make[2]: Leaving directory '/root/elks/image'
make[1]: *** [Makefile:28: image] Error 2
make[1]: Leaving directory '/root/elks/image'
make: *** [Makefile:17: all] Error 2
```

is this related to WSL? My mount is not working in WSL for some reason.

toncho11:
[tool_chain_user_basic4.zip](https://github.com/user-attachments/files/18565978/tool_chain_user_basic4.zip)
I can not boot.
I could not make this work. 86box is restarting when trying to load the above image.

ghaerr:
@toncho11,

I tried both your images with QEMU and neither boot. I really don't know why. Can't you just start with the standard ELKS MINIX 32M image, which is known to work? I'll submit another PR that also builds a c86dev.tar file of the /usr development tree that you can unzip directly into the filesystem either from within ELKS or outside of it, if possible.

If you can't get that to work, I'll add a new 32M image "hd32minix-c86.img" that boots to ELKS and has the 8086 toolchain preinstalled on /usr. Do you want an MBR or non-MBR version of that?

> is this related to WSL? My mount is not working in WSL for some reason.

Unfortunately I know nothing about WSL and don't know why it says "Disk Full" (and prefer to stay a ways away from anything Microsoft develops).

ghaerr:
@toncho11, see #2203 - this has directions for creating a tarball devdisk or a 32M HD image with the C86 development kit preinstalled. Hopefully one of these two solutions will make it easy for you to build and test the devkit. Thank you!

toncho11:
I think you made all the required scripts to add hd32minix-c86.img in github actions. And this will make the whole process of testing and enjoying the tool-chain much easier.

ghaerr:
> you made all the required scripts to add hd32minix-c86.img in github actions.

We can't do that (yet) because the buildc86.sh and copyc86.sh scripts require access to the 8086-toolchain repo, which is currently separate from ELKS. We would need to either combine the repos or pull the toolchain repo over into the ELKS repo and build time to make this work. There are both benefits and disadvantages to keeping the repos separate. I suppose this should be discussed on the C compiler thread?

ghaerr:
Another potential problem of combining the repos or pulling the toolchain over would be that all developers will have to have the OpenWatcom compiler installed. Having OWC is becoming increasingly important though, especially since various new ELKS applications are being built using it, and we should include those applications in our standard distribution.

toncho11:
I think watcom can be added to build.sh script. I mean the binary version. 
It will be downloaded and installed so that it is ready for building the toolchain and all future applications that will needed it.
@rafael2k 

Then the toolchain and other applications will be git cloned in elks/extern/
So the buildc86.sh will download the toolchain in elks/extern/ and build it.
Once built it could be added to an image hd32minix-c86.img

rafael2k:
I like the idea @toncho11. If watcom is integrated to ELKS build system, we can even import the toolchain software to elkscmd. 

Even a new "development tools" tab to menuconfig.
: )

ps: don't build watcom on linux x86_64 - the watcom binary packages work neatly on virtually any linux distribution.

PR #2207, opened 2025-01-31
Update elksemu

BinaryMelodies:
Update the reboot system call


ghaerr:
@BinaryMelodies, thanks for taking the time to submit this sequence of enhancements to elksemu, they are much appreciated, and hopefully will allow for easier maintenance of your RetroLinker project!

PR #2209, opened 2025-01-31
Add system calls for networking support

BinaryMelodies:
This enables elksemu to run ELKS binaries with networking system calls.

There is one significant difference from the standard Linux API: fd_set is contains only 64 file descriptors instead of 1024 descriptors. This is because the ia16-gcc binary I used FD_SETSIZE = 64, but if ELKS currently supports more file descriptors, this can be easily updated.

### DIFF
--- a/elksemu/elks_sys.c
+++ b/elksemu/elks_sys.c
@@ -23,6 +23,7 @@
 #include <dirent.h>
 #include <sys/time.h>
 #include <sys/ptrace.h>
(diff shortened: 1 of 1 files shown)


ghaerr:
The fd_set size should not be an issue at all, since it's currently stored in a single unsigned long, given that 20 file descriptors is the max file descriptors per process (unless specially reconfigured).

PR #2210, opened 2025-01-31
Add support for COMBID executables

BinaryMelodies:
Executables where text and data share the same address space


ghaerr:
Thanks for the COMBID support! We're unlikely to add this to the ELKS kernel since our current toolchains don't support producing it, and desiring to have shared text segments as a default.

PR #2213, opened 2025-02-02
[nanox] Initial changes to build Server/Client mode for ELKS

tyama501:
Hello,

This is the initial PR to build nano-X Server/Client mode for the current ELKS,
as the discussion https://github.com/ghaerr/elks/discussions/1810#discussioncomment-12026927

Makefile has been changed to build the basic applications with nano-X Server/Client mode
if LINK_APP_INTO_SERVER=1 is comment out.

The include headers are changed to use sys/socket.h instead of linuxmt/socket.h.

GR_NAMED_SOCKET is changed to /tmp/nxsock instead of /var/uds.

Timeouts are added. (still need debugs)

I have commented out raise(z); in client.c since there is no raise in our signal.h.
Is this OK?

Thank you.

### DIFF
--- a/elkscmd/nano-X/Makefile
+++ b/elkscmd/nano-X/Makefile
@@ -126,16 +126,18 @@ bin/nxvgatest: demos/demo-vga.o
 # client/server
 #
 nano-X: $(SERVFILES) $(NANOXFILES)
-	$(LD) $(LDFLAGS) $(NANOXFILES) $(SERVFILES) $(NANOLIBS) $(LIBS) -o bin/nano-X $(MATHLIB) $(LDFLAGS)
+	$(LD) $(LDFLAGS) $(NANOXFILES) $(SERVFILES) $(NANOLIBS) $(LIBS) -o bin/nano-X $(MATHLIB) $(LDLIBS)
 
 # only a few demos work in client/server mode
-nxdemos: libnano-X.a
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @tyama501, this looks great, thank you! 

I want to spend a little time looking at it to figure what might be the problem with GrGetNextEventTimeout and then will commit. I will then try to setup to debug the timeout issue we're having.

I think I can add a raise library routine as well.

> GR_NAMED_SOCKET is changed to /tmp/nxsock instead of /var/uds.

Good. I will also attempt a fix at getting UNIX sockets working on FAT filesystems. Since there are no file "mode" bits on the FAT filesystem, we have to resort to some sort of (temporary?) naming kluge in order to get this to work. I'll do this after commit.

Now that you have this running client/server using UNIX sockets, I'm also going to have to remove the almost-identical "Nano-X Sockets" from the ELKS kernel/build. Looking closely, the only difference seems to be using a number rather than a filename to bind a socket. Since the Nano-X Sockets are non-standard and UNIX sockets work (and can be made to work on FAT as well), there is no need for the duplicate socket system.

Also, in writing the new VGA fast line drawing routines for C86 toolchain (which I took from older Nano-X screen drivers written in MASM in 1999 (!) ), I found a few areas where Nano-X drawing should be able to be sped up, possibly significantly. I plan on enhancing that as well.

Thank you for your work on this!

toncho11:
We are going for a similar to Windows 1.0 or GEM UI for ELKS, right :)

PR #2215, opened 2025-02-03
[Nano-X] Fix various problems with multiple client/server Nano-X

ghaerr:
Debug and fix nxtetris problem using GrGetNextTimeout described in https://github.com/ghaerr/elks/discussions/1810#discussioncomment-11873949.

Fixes server error when handling multiple clients.
Ignores ^C instead of exiting (use ESC to exit).

Sets default Nano-X build to client/server.

@tyama501, finding all the problems and getting this working proved to be far more complicated than I thought. Long story short, we can't directly send a timeout to the nano-X server because it is servicing multiple clients - thus when a timeout is required when getting an event, the server is put into a "polling" mode where it times out every 50msecs and then calls back to each client waiting for a timeout to measure whether the elapsed time has been long enough. There was also a major bug where the server would reply to the wrong client in certain cases.

So far, I've been able to get two clients or so working well with the server. Adding more in some cases causes some window remapping problems. So for now I'm staying with just getting one or two clients working well. Go ahead and experiment and see what other problems you might find. I see there are a couple ifdef CONFIG_PC98 in the srvmain.c code which will likely have to removed or changed, as soon as I get the UNIX sockets working on FAT, which is my next step.

After spending all day in this code, I'm seeing there's still a lot of cleanup to be done, but this should be a big step forward.

Here's a video of nxclock running with nxtetris, run using the following script:
```
nano-X & sleep 1; nxclock & exec nxtetris
```

https://github.com/user-attachments/assets/c545c67f-23cf-438b-b83c-a402007e9b2d

### DIFF
--- a/elks/net/unix/af_unix.c
+++ b/elks/net/unix/af_unix.c
@@ -385,6 +385,7 @@ static int unix_write(struct socket *sock, char *ubuf, int size, int nonblock)
     pupd = UN_DATA(sock)->peerupd;	/* safer than sock->conn */
 
     while (!(space = UN_BUF_SPACE(pupd))) {
+	printk("NO SPACE on WRITE pid %d size %d\n", current->pid, size);
 	sock->flags |= SF_NOSPACE;
 
 	if (nonblock)
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -202,9 +202,10 @@ test/libc/test_libc             :test                           :1440c
 test/other/test_fd              :test
 test/other/test_float           :test
 #nano/nano-2.0.6/src/nano       :other
+nano-X/bin/nano-X               :other                              :1440c
 nano-X/bin/nxclock              :other                              :1440c
 nano-X/bin/nxdemo               :other
-nano-X/bin/nxtest               :other
+#nano-X/bin/nxtest               :other
 nano-X/bin/nxtetris             :nanox                  :1200k
 nano-X/bin/nxlandmine           :nanox                   :1232k :1440k
 nano-X/bin/nxterm               :nanox                   :1232k     :1440c
--- a/elkscmd/nano-X/Makefile
+++ b/elkscmd/nano-X/Makefile
@@ -7,6 +7,7 @@ BASEDIR = ..
 include $(BASEDIR)/Makefile-rules
 
 LOCALFLAGS = -DELKS=1 -DUNIX=1 -DDEBUG=1 -I.
+AR = ia16-elf-ar
 MATHLIB =
 
 # nano-X demo programs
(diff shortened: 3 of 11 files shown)


tyama501:
Wow! Great!
Cool, running 2 applications!
Thank you.

toncho11:
Awesome both of you!!! :)

toncho11:
This proves that a desktop app similar to MS Windows 3 is possible. There will be a launcher app (or desktop app) with icons for tetris and clock for example. Clicking on each icon will launch a new process for the corresponding application and draw using the same nano-X. 

`nano-X & sleep 1 & desktop`

toncho11:
What is need to test this?

1) I have to unable unix sockets in the kernel?
2) Are nxtetris and nxclock compiled by default to use nano-X and unix sockets? The github action images do give an error: "cannot bind to named sockets". I use the hdd minix image.

tyama501:
> This proves that a desktop app similar to MS Windows 3 is possible. There will be a launcher app (or desktop app) with icons for tetris and clock for example. Clicking on each icon will launch a new process for the corresponding application and draw using the same nano-X.
> 
> `nano-X & sleep 1 & desktop`

npanel in the microwindow repo is that kind of launcher, although it is not window manager so I think it can launch only one application.
(I have tried it with sdl on linux recently)

ghaerr:
> Are nxtetris and nxclock compiled by default to use nano-X and unix sockets? The github action images do give an error: "cannot bind to named sockets". 

Yes, this commit changed the default built to separate nano-X client and server programs, allowing for multiple clients to connect after the nano-X server is started. I just noticed that the default IBM and PC-98 builds do not have CONFIG_UNIX=y UNIX sockets enabled. I will be fixing that with the next PR so that CI builds will work.

> What is need to test this?

After the next PR, either edit a shell script with the following or type in from the shell:
```
nano-X & sleep 1; nxclock & nxtetris
```

> npanel in the microwindow repo is that kind of launcher, although it is not window manager

Yes, we can port npanel desktop manager, but we need to get the Nano-X Window Manager ported first. I am looking into that.

toncho11:
I started nxterm like that. How do I start nx programs from it? If I just type nxclock it closes nxterm.

ghaerr:
> I started nxterm like that. How do I start nx programs from it? If I just type nxclock it closes nxterm.

We're having lots of technical problems right now - that should work but nxterm is buggy, I just found another bug getting nxworld to run, nxclock clears the background and erases other programs, all sorts of stuff. Client/server is barely standing on two legs. The best we can do at the moment is experiment with running 1 or 2 programs from scratch in a shell script.

I hope to have more fixed, but there's lots of issues it seems. Working on another PR now to fix nxworld and nxclock. Nxterm will have to wait a bit, it crashes when run with nxclock for unknown reasons. Feel free to report other anomaly's! :)

toncho11:
True that nxterm closes when I start nxclock from it.

PR #2216, opened 2025-02-03
[kernel] Update syscall buffer verify FAULT to show bad address

ghaerr:
Adds more verbose error message to FAULT output when large model programs pass a buffer that seemingly does not belong to the process address space. Discussed in https://github.com/ghaerr/elks/issues/2200#issuecomment-2629112920.

Also fixes unsaved CX register in OWC \_\_divmod routine. The C library should be recompiled for Watcom using either "make owc" or ./buildc86.sh.

@rafael2k: It would be worth a try to remove the "return 0" hack in verfy_area and then rerun gzip after recompiling the OW C library. This will then give better output as to what is happening with gzip, especially if run with strace.

### DIFF
--- a/elks/arch/i86/mm/user.c
+++ b/elks/arch/i86/mm/user.c
@@ -34,7 +34,7 @@ int verfy_area(void *p, size_t len)
     if (seg_verify_area(current->pid, current->t_regs.ds, offset))
         return 0;
 
-    printk("FAULT\n");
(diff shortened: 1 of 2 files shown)


rafael2k:
Thanks! I'll debug zlib and try to find what is going on. Already removed the "return 0" before the address check.

PR #2221, opened 2025-02-03
[Nano-X] Rewrite Nano-X Makefiles for easy understanding

ghaerr:
Complete rewrite of the elkscmd/nano-X/Makefile and demos/Makefile to make it super simple to add new packages. Both LINK_APP_INTO_SERVER=y and client/server builds are supported, with the default remaining client/server.

Major cleanup of the source directory, with some clients built with smaller windows so that more stuff can be displayed.

Screen Shot 2025-02-03 at 3 56 45 PM

@tyama501, I haven't done any PC-98 testing but nothing technical has changed with regards to PC-98. Please let me know of any problems you might run into, thanks!


tyama501:
Hello @ghaerr ,

It seems that nano-X is not included in current CI build fd1232.img.
Could you add 1232c to nano-X and nxclock?
(Or I will add when I have another time)

nano-X/bin/nano-X               :nanox                              :1440c
nano-X/bin/nxclock              :nanox                              :1440c

Thank you.

tyama501:
By the way, current nxclock can only display UTC, I may try to display local time..

nxclock.c
now = tv.tv_sec - (60 * tz.tz_minuteswest);

I think it is local time if now = tv.tv_sec

ghaerr:
> Could you add 1232c to nano-X and nxclock?
(Or I will add when I have another time)

Yes, I'll add that ASAP. BTW, I spent many hours yesterday trying to get the Nano-X window manager working on ELKS. Very long story short, our version is 25 years old and lots of issues. So I'm not off trying to get the main Nano-X repo to build for ELKS. It is a lot larger code but may be worth switching to if I can get it running.

> By the way, current nxclock can only display UTC, I may try to display local time..

That should be an extremely simple fix by replacing gmtime() with localtime() in nxclock.c, and setting TZ= in /etc/profile.

PR #2226, opened 2025-02-15
[cmds] Display mouse port being opened on mouse command

ghaerr:
As discussed in https://github.com/ghaerr/microwindows/pull/106 with @toncho11, the `mouse` command will now display the mouse port being used when started.

In addition, Nano-X will display the mouse port on startup. Properly configured, this "message" will display on either /dev/ttyS0 when running on QEMU, or /dev/console otherwise. Because nano-X clears the screen to graphics mode, the mouse message may be erased quickly and not be seen (along with any other error or informative messages). For this reason, it is recommended that emulators set the mouse port to /dev/ttyS1 (along with MOUSE_PORT=/dev/ttyS1) and console output to /dev/ttyS0 so that messages can be seen.

Here's a pre-built image with both fixes for testing.
[fd2880.img.zip](https://github.com/user-attachments/files/18806961/fd2880.img.zip)

### DIFF
--- a/elkscmd/sys_utils/mouse.c
+++ b/elkscmd/sys_utils/mouse.c
@@ -100,6 +100,7 @@ open_mouse(void)
 
 	/* open mouse port*/
 	port = getenv("QEMU")? MOUSE_DEVICE2: MOUSE_DEVICE;
+	printf("Opening mouse on %s\n", port);
 	mouse_fd = open(port, O_EXCL | O_NOCTTY | O_NONBLOCK);
 	if (mouse_fd < 0) {
 		printf("Can't open %s, error %d\n", port, errno);


toncho11:
So /dev/ttyS1 should be COM2 on windows? My emulator uses COM1, COM2 ...

ghaerr:
> So /dev/ttyS1 should be COM2 on windows? My emulator uses COM1, COM2 ...

I'm not sure your question here, but the general idea is to map "COM1" (/dev/ttyS0) to the host Terminal, and "COM2" (/dev/ttyS1) to the emulator MS mouse. Then set MOUSE_PORT=/dev/ttyS1.  If it is not possible to attach COM1 to the host terminal, that is still OK, you can still attach the emulator mouse to COM2, right?

toncho11:
mouse says /dev/ttyS0
I can not change it with MOUSE_PORT=/dev/ttyS1
If I set COM2 in the emulator mouse does not work (because it is not ttyS1)
I can not figure it out.

toncho11:
`MOUSE_PORT=/dev/ttyS1`
or
`export MOUSE_PORT=/dev/ttyS1`

does not work.

PR #2227, opened 2025-02-16
[c86 libc] Add support for signal() for C86 programs

ghaerr:
Adds support for `signal` system call in C86 C library, requested by @rafael2k in https://github.com/ghaerr/8086-toolchain/issues/48.

Using signals in C86 programs is now fully supported, with a caveat: the C handler specified to be called back using `signal` must not be the first function declared in your C program. This is because currently the linker assigns the first function it sees address 0, which happens also to be the value of SIG_DFL (default signal action). If this callback function is the first function in your C program, it will also be assigned address 0.

Thus, the following "zero()" function declaration as the first function is necessary in C86 applications to avoid this problem:
```
#include <signal.h>
#include <stdio.h>
#include <unistd.h>
#include <errno.h>

void zero(void) {}      /* required to keep catch() from having address 0 */
void catch(int sig)
{
    printf("GOT %d\n", sig);
    signal(SIGINT, catch);
}

int main(int ac, char **av)
{
    sighandler_t rv = signal(SIGINT, catch);
    printf("ret %x errno %d\n", rv, errno);

    for (;;)
        sleep(1);
    return 0;
}
```

A later update will provide a linker modification and/or manually supplying the C86 crt0.o file as the first file to ld86, which will solves this issue.

### DIFF
--- a/elks/include/linuxmt/signal.h
+++ b/elks/include/linuxmt/signal.h
@@ -187,9 +187,12 @@ typedef stdcall __far void (*__kern_sighandler_t)(int);
 #pragma GCC diagnostic pop
 #endif
 
-#if defined(__WATCOMC__) || defined(__C86__)
-/* FIXME broken for C86 w/o stdcall and __far */
+#ifdef __WATCOMC__
 typedef void stdcall (__far *__kern_sighandler_t)(int);
+#endif
+
+#ifdef __C86__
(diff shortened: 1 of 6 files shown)


rafael2k:
Thanks @ghaerr ! Re-enabling signal (not first func) in the elks-viewer tools. Any problem I let you know.

PR #2228, opened 2025-02-16
[libc] Rewrite mouse and debug console output to work together

ghaerr:
Both Nano-X and ELKS' mouse programs/driver have been simplified to just work by setting "export MOUSE_PORT=/dev/ttyS0" (or /dev/ttyS1), depending on which serial device the mouse is on. The default line in /bootopts has been updated for COM1 and should work by just uncommenting it.

These changes were made because of the problems discussed by @toncho11 in https://github.com/ghaerr/elks/pull/2226#issuecomment-2661086534. Now, only a single mouse port open is attempted, and the debug console routine was changed to work with the mouse port setting.

@toncho11, this should now work for you. Setting "export MOUSE_PORT=/dev/ttyS0" at the shell prompt, or uncommenting the line in /bootopts should work with 86Box. It seems that 86Box only supports mouse on COM1 and there were also problems in the ELKS debug console code that did not work well with the mouse. The mouse may still work without MOUSE_PORT= being set on your emulator, but definitely will with it set.

Here's a prebuilt image:
[fd2880.img.zip](https://github.com/user-attachments/files/18813375/fd2880.img.zip)

### DIFF
--- a/elkscmd/rootfs_template/bootopts
+++ b/elkscmd/rootfs_template/bootopts
@@ -21,4 +21,4 @@
 #net=ne0
 #debug=1
 #console=ttyS0,19200 3
-#MOUSE_PORT=/dev/ttyS1
+#MOUSE_PORT=/dev/ttyS0
--- a/elkscmd/sys_utils/mouse.c
+++ b/elkscmd/sys_utils/mouse.c
@@ -14,8 +14,8 @@
 #include <fcntl.h>
 #include <termios.h>
 
-#define	MOUSE_DEVICE  "/dev/ttyS0"	/* real hardware mouse tty device*/
-#define	MOUSE_DEVICE2 "/dev/ttyS1"	/* QEMU mouse tty device*/
+#define	MOUSE_PORT    "/dev/ttyS0"  /* default port unless MOUSE_PORT= env specified */
+#define	MOUSE_QEMU    "/dev/ttyS1"  /* default port on QEMU */
 #define	MOUSE_MICROSOFT		1		/* microsoft mouse*/
 #define	MOUSE_PC			0		/* pc/logitech mouse*/
 #define MAX_BYTES	128				/* number of bytes for buffer*/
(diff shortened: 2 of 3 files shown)


toncho11:
Thanks. I will try later.
Please note that I am unable to boot 2880 images. I just do not know how to set a virtual floppy to 2880kb in 86Box. Please put HDD images if possible next time.

toncho11:
Yes, now it works without modifying anything and COM1 selected in the emulator. Thanks. 
It works well on a 486 emulation. 
I noticed that clicking on the X in the Start panel does not exit nicely as with CTRL+A. Maybe it was not designed to. Now it closes but it does not give a usable prompt.

toncho11:
I just tried with COM2 and `export MOUSE_PORT=/dev/ttyS1` and it also worked.

ghaerr:
> I noticed that clicking on the X in the Start panel does not exit nicely as with CTRL+A. Maybe it was not designed to. Now it closes but it does not give a usable prompt.

Thanks! This took a while to figure out: it's not really a hang after exit, but the shell is in a funny terminal mode since it was reading the same keyboard that was set in RAW mode by nano-X. A ^C will exit to a prompt, but PR https://github.com/ghaerr/microwindows/pull/108 fixes the problem of exiting using the close box if no other windows are open, for now.

rafael2k:
I have this here to be able to include string.h with C86:

'''
diff --git a/libc/include/string.h b/libc/include/string.h
index 0eac14ff..cd7285c6 100644
--- a/libc/include/string.h
+++ b/libc/include/string.h
@@ -28,8 +28,10 @@ void * memset(void*, int, size_t);
 int memcmp(const void*, const void*, size_t);
 void * memmove(void*, const void*, size_t);
 
+#if 0
 void __far *fmemset(void __far *buf, int c, size_t l);
 int fmemcmp(void __far *s1, void __far *s2, size_t n);  /* Watcom C only, in ASM */
+#endif
 
 /* Error messages */
 char * strerror(int);
'''
Should I create an issue for it?

PR #2238, opened 2025-02-24
[owc libc] Fix OWC link errors with small and compact models using _weaken

### LINKED ISSUE #2237: [OpenWatcom] Can not link with libc compiled in small memory model
Trying to link software using the small-memory model libc gives me:
```
owcc -c -O2 -mcmodel=s -bnone -march=i86 -std=c99 -fno-stack-check -fnostdlib -Wc,-fpi87 -Wc,-zev -Wc,-zls -Wc,-x -Wc,-wcd=303 -Wall -Wextra -I/home/rafael2k/programs/devel/elks/libc/include -I/home/rafael2k/programs/devel/elks/elks/include -I/home/rafael2k/programs/devel/elks/libc/include/watcom -D__ELKS__ -DNDEBUG -o ppmview.obj ppmview.c
owcc -bos2 -s -Wl,option -Wl,dosseg -Wl,option -Wl,start=_start -Wl,option -Wl,nodefaultlibs -Wl,option -Wl,stack=0x1000 -Wl,option -Wl,heapsize=0x1000 -o ppmview ppmview.obj graphics.obj utils.obj -Wl,library -Wl,/home/rafael2k/programs/devel/elks/libc/libc.lib
Error! E2052: file /home/rafael2k/programs/devel/elks/libc/libc.lib(/home/rafael2k/programs/devel/elks/libc/stdio/vfprintf.c): relocation at 0001:14bf not in the same segment
Error! E2052: file /home/rafael2k/programs/devel/elks/libc/libc.lib(/home/rafael2k/programs/devel/elks/libc/stdio/vfprintf.c): relocation at 0001:15aa not in the same segment
Error: Linker returned a bad status
make: *** [Makefile.owc:51: ppmview] Error 1
```

ps: I build the watcom libc after setting watcom.model to "MODEL=s".

ghaerr:
Fixes #2237.

OWC wlink was failing to link vfprintf.obj in small and compact models due to the mechanism used to make "weak" symbols work using the \_weaken macro, because the OWC compiler generates a far segment relocation entry which causes wlink to fail with a "relocation not in same segment" error.

This all happens because vfprintf attempts to be smart about dragging in the large dtostr and ptostr routines into printf when floating point/ptick display aren't used. One fix would be to always link in dtostr in small model OWC programs, which would make them considerably larger, so instead this fix eliminates automatically including dtostr/ptostr entirely, eliminating automatic floating point display using printf %f/%g or %k (ptick) format specifiers. Instead, the application will need to call dtostr or ptostr explicitly if needed.

Also removes /lib/nxworld.map from the image as Nano-X is now built outside of ELKS in the Microwindows repo, and it is installed from there when built.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -209,7 +209,7 @@ test/other/test_float           :test
 #nano-X/bin/nxlandmine           :nanox                   :1232k :1440k
 #nano-X/bin/nxterm               :nanox                   :1232k     :1440c
 #nano-X/bin/nxworld              :nanox                   :1232c :1440k
-nano-X/bin/nxworld.map ::lib/nxworld.map :nanox          :1232c :1440k
+#nano-X/bin/nxworld.map ::lib/nxworld.map :nanox          :1232c :1440k
 basic/basic                     :basic                  :1200k
 advent/advent                     :other                            :1440c
 advent/advent.db ::lib/advent.db  :other                            :1440c
--- a/libc/stdio/vfprintf.c
+++ b/libc/stdio/vfprintf.c
@@ -42,12 +42,25 @@
  * Greg Haerr enhanced for speed, new features
  */
 
+/*
+ * NOTE: Weak symbols aren't working with OpenWatcom small and compact models
+ * due to compiler creating far segment relocation on _weaken() and link failure.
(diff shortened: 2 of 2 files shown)


rafael2k:
Thanks! Interesting topic about weak symbols and floating point support. Did not know about it.

PR #2241, opened 2025-02-25
Create SECURITY.md

emresandikci:


ghaerr:
Hello @emresandikci,

Thank you for preparing this potential security policy. May I ask, what exactly is the purpose of such a policy for ELKS?

ELKS itself is not secure, and cannot be made secure, since it exists as a 8086 real-mode operating system which unfortunately cannot protect itself due to the lack of any hardware MMU or other device that might signal to the OS that something is doing something that it should not. Any application can read or write to any memory location, disk block, read or write to any hardware I/O port address, etc. 

We cannot provide 48-hour timeline of any fix, being a volunteer organization, and we would prefer that issues of any kind with ELKS be published directly on GitHub, so that all can see, learn and comment on what any concern might be. 

ELKS is primarily built and used for education and having fun on ancient PCs. It should definitely not be used in any environment where security is important, as it has very little.

Should we desire to publish a security policy, the offered policy would need significant rewriting.

Thank you!

emresandikci:
> Hello @emresandikci,
> 
> Thank you for preparing this potential security policy. May I ask, what exactly is the purpose of such a policy for ELKS?
> 
> ELKS itself is not secure, and cannot be made secure, since it exists as a 8086 real-mode operating system which unfortunately cannot protect itself due to the lack of any hardware MMU or other device that might signal to the OS that something is doing something that it should not. Any application can read or write to any memory location, disk block, read or write to any hardware I/O port address, etc. 
> 
> We cannot provide 48-hour timeline of any fix, being a volunteer organization, and we would prefer that issues of any kind with ELKS be published directly on GitHub, so that all can see, learn and comment on what any concern might be. 
> 
> ELKS is primarily built and used for education and having fun on ancient PCs. It should definitely not be used in any environment where security is important, as it has very little.
> 
> Should we desire to publish a security policy, the offered policy would need significant rewriting.
> 
> Thank you


Thanks for the clarification and the detailed explanation! I understand the approach and will go ahead and close the PR.

PR #2243, opened 2025-02-26
[libc] Update libc headers to allow compilation with -std=c99

ghaerr:
Allows ELKS C library and headers to be used with strict C99-compliant ANSI compilers.
Discussed in https://github.com/ghaerr/elks/issues/2239#issuecomment-2683568808.

Protects non-ANSI declarations using \_\_STRICT\_ANSI\_\_. 

C86/CPP86 now defines \_\_STRICT\_ANSI\_\_ since \_\_far is not implemented. `__far` is now not defined to blank for C86 so as to cause an error rather than ignoring the keyword.

@rafael2k, your idea of allowing -std=c99 to be enabled caused finding a few more future problems with ELKS libc using C86: the `fmemalloc`, `fmemfree` and `fmemset` could be called with no error, which would have previously resulted in improper operation, since far pointers are not implemented in C86. 

The C library headers should now be able to be used with ia16-elf-gcc using -std=c99, although I have not tried that on non-ELKS sources. You should also try recompiling elks-viewer and your other projects to ensure that they are also working with the more-restricted C86. Please report any problems. Note that -std=c99 cannot be used with ia16-elf-gcc if any \_\_far data access is required, as this isn't ANSI-compliant.

I have tested the 8086 toolchain build on ELKS and it continues to function properly.

### DIFF
--- a/elks/include/arch/cdefs.h
+++ b/elks/include/arch/cdefs.h
@@ -43,7 +43,6 @@
 
 #ifdef __C86__
 #define __HAS_NO_FLOATS__   1
-#define __far
 #define noreturn
 #define stdcall
 #define restrict
--- a/elks/include/linuxmt/signal.h
+++ b/elks/include/linuxmt/signal.h
@@ -177,6 +177,8 @@ typedef unsigned long sigset_t; /* at least 32 bits */
 
 /* Type of a signal handler within userland.  */
 typedef void (*sighandler_t)(int);
+
+#ifndef __STRICT_ANSI__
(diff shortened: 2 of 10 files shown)


rafael2k:
This is really great, thanks a lot @ghaerr !
I'll try to compile Lua with ia-gcc16 small model in its C99 mode as soon as I have time for it.

No worries, I'll never call fmemalloc/free/set from C86 code. 

I'll now go back to C86, as that C86 bug compiling jpeg decoder is still outstanding.

PR #2246, opened 2025-02-27
[libc] Initial support for OpenWatcom soft floating point

ghaerr:
Adds initial support for software (emulated) floating point to ELKS programs compiled with OpenWatcom C.

Discussed in https://github.com/ghaerr/elks/issues/2239#issuecomment-2683612940.

By default, inline 8087 support is turned off, and `ewcc` uses the `-msoft-float` option (-Wc,-fpc) to produce calls to software floating point routines contained in the C library for `float` and `double` precision floating point.

In order to turn on %f, %e and %g printf output, the following must be included in the application source in a function somewhere, or the dtostr() double-to-string formatting routines will not be linked in, and nothing will be output. Due to technical issues with weak symbols, only medium and large models are currently supported:
```
#include <sys/linksym.h>
#include <stdio.h>
int main(int ac, char **av)
{
    __STDIO_PRINT_FLOATS;  /* link in libc floating point printf support */
    ...
}
```
The above code is not necessary if floating point numbers do not need to be displayed using printf.

There is currently no math library for OWC. OWC documentation states that software emulated floating point is not IEEE 754 compliant; it is unknown exactly what that means.

This is a feature preview, it is *not recommended* to start using floating point in your own programs without special testing. This PR has been tested with ELKS basic, which seems to work, although still no math library. Producing a math library may not be easily possible, depending on learning more about non-IEEE support for emulated floating point, and/or porting the OWC math library vs the ELKS math library.

Programmers should remove any inline hardware 8087 code generation options (e.g. -Wc,-fpi or -Wc,-fpi87) and use `ewcc` or add `-msoft-float` to their OWCC command line.

Support for FP overflow and underflow is non-existent, and may produce entirely incorrect results and/or abort program execution. Fix coming.

The OWC floating point emulation library has some 8087 and 80386 code in it that had to be included - it is not known whether or how well OWC programs will run on older 8086 hardware - this still needs testing. In the future, a `basic.os2` built with OWC will be supplied for testing on 8086 hardware.

### DIFF
--- a/elks/tools/objtools/ewcc
+++ b/elks/tools/objtools/ewcc
@@ -34,16 +34,15 @@ ELKSINCLUDE2=$TOPDIR/libc/include/watcom
 # -Wc,-zev                  # enable void arithmetic
 # -Wc,-zls                  # remove automatically inserted symbols
 # -Wc,-wcd=N                # disable warning N
-# -Wc,-fpi87                # inline 8087 fp
 # -Wc,-x                    # ignore INCLUDE environment variable
 # -Wc,-r                    # save/restore segment registers
 # -fnostdlib                # don't refere standard libraries
 # unused:
 # -fno-stack-check          # don't generate stack check code
 # -ztNum                    # specify far data threshold (default 32767, or 256 if no Num)
+# -Wc,-fpi87                # inline 8087 fp
 # -mhard-emu-float          # -Wc,-fpi (inline 8087 w/emulation)
-# -msoft-float              # -Wc,-fpc
-# -Wc,-fpc                  # non-IEEE soft fp
+# -msoft-float              # -Wc,-fpc (software fp)
 # -fpmath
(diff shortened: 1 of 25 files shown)


toncho11:
Is this related to: https://github.com/rafael2k/lua/issues/4 ?

ghaerr:
This PR allows for OpenWatcom-compiled programs to implement floating point add, subtract, multiply and divide, but there is no math library, so no sin/cos, etc. 

I've been slowly reading through the massive amounts of code and documentation in OpenWatcom with regards to floating point, and at some point I'll probably be able to bring over the math library, but we're not there yet. OpenWatcom's full support for floating point is very complicated, one can read the Users Manual for details.

There may be other issues getting floating point working in Lua (including perhaps its own assumptions of a full math library), but this provides the compiler floating point support with no math library.

Issue #1942, opened 2024-07-28
Use of plain char as signed char in libc is not optimal


jmalak:
With regard to improve OpenWatcom stuff I found one case where libc could be improved get it smaller a little bit.
The libc library is primarily compiled by gcc (16-bit) and it uses signed char (default) for plain char, but OpenWatcom uses unsigned char (default). When I play with OpenWatcom stuff improvement I checked generated code and using of unsigned char for plain text get smaller generated code.
Anyway now it is mismatch between GCC and OpenWatcom that it should be corrected.

Simple solution is switch OpenWatcom to use signed char for plain char, because most of testing was done with executable generated by GCC with signed char type for plain char, but code will be a little bigger.
Unsigned char get a little bit smaller code that it can be switch to use unsigned char also for GCC, but it is not possible without review of code which use plain char because there can be issue with signed/unsigned values not generally but on some places in code.

This issue is rather for future development to optimize size of ELKS libc.

ghaerr:
@jmalak,

I'm not sure how this issue got left open for so long! Very good points and suggestions you have made above. As you can see, #2251 is using your suggestion of a simple solution for now to match ia16-elf-gcc's behavior of plain char being signed.

Thank you!

PR #2251, opened 2025-03-05
[owc libc] Make signed char default for OpenWatcom C

ghaerr:
From suggestion by @jmalak in #1942.

Plain `char` type is now default to `signed` rather than `unsigned` for the C library and all programs compiled with `ewcc`. This should be the default for all programs compiled for ELKS, including programs compiled outside of the ELKS repo.

This is being done so that the C library and other code is also compatible with ia16-elf-gcc, which treats plain `char` as signed.

@rafael2k, you should probably add `-fsigned-char` to the make instructions for your projects (as well as `-msoft-float` for those programs with any floating point if you haven't already). I'll be making the change to 8086-toolchain shortly.


rafael2k:
Thanks @ghaerr, will change. Btw, in C86 plain char is signed or unsigned?

ghaerr:
C86 plain char is signed.

PR #2259, opened 2025-03-16
[cmds] Add paint program for IA16, OWC and C86

ghaerr:
Adds paint program for drawing and saving images on ELKS. Compiles using any of ELKS compiler toolchains including 
IA16, OpenWatcom, and C86.

To run, type `paint` at the ELKS prompt with MOUSE_PORT= set to the mouse serial. QEMU should work out of the box. Some special keys are 'f' for flood fill, 'c' to clear and 'q' to quit; the brush sizes colors and clear/exit are also available on the control panel. The final painted image can be saved as a .bmp file by clicking the 'floppy' icon, but beware, as the image is likely not able to fit on a 1440k disk, as it requires 232k (!).

The original application source came from the cool [TomentPainter](https://github.com/silvematt/TomentPainter) program which was written for SDL2. The application was then rewritten to optionally use floating point if available, and use an enhanced version of the ELKS graphics routines from 8086-toolchain/examples/graphics.c also with the portable event handling routines from there and the mouse/cursor display routines from Nano-X. All were enhanced to run on ia16-elf-gcc, OpenWatcom, or C86. The OpenWatcom graphics ASM code was originally written by @rafael2k.

Currently, the paint application is fairly small, and ranges from 12k-22k depending on compiler and whether floating point emulation is included. It runs standalone, outside of Nano-X.

The plan, suggested originally by @toncho11, will be to add the graphics, event management, and mouse routines into the ELKS C library, supported for all three compilers. 

There's a lot of cool stuff happening behind the scenes that was required to get paint operational. A graphics cursor had to be implemented that follows the mouse input and displays a cursor on the screen. This turned out to be faster than Nano-X, and thus the larger 16x16 cursor is being shown. Various icons are also displayed on the control panel, and it was decided that unlike Nano-X, the images should not be loaded into memory, but read from a file and displayed scanline-by-scanline, in order to allow images in any compiler model and use only a 640 bytes of memory, at the expense of having to open a file each time the image is displayed. (The OWC build is still large model for now, but C86 and IA16 are small model).

ELKS Paint

The program source is in elkscmd/gui, and the IA16 version is automatically built and fits onto the 1440k image. If the WATCOM= or C86= environment variables are set, then typing "make" in that directory will also build the those versions, which are named opaint and cpaint respectiviely. This is useful for compiler portability testing as well as observing code size. A new C86 compiler bug was found and "flood fill" won't compile for the time being.

This project is also a precursor to allow new programs to be added to the ELKS source tree and compiled by either the GCC/IA16, OWC or C86 compilers. As such, three template makefiles (Makefile.gcc, Makefile.owc and Makefile.c86) are included which should be be used as templates to allow for future easy integration. More work coming on that.


ghaerr:
Here's a 1440k MINIX image with the IA16 version compiled as `paint` for testing:
[fd1440.img.zip](https://github.com/user-attachments/files/19275109/fd1440.img.zip)

rafael2k:
Very cool! 

I can see at least one "#ifndef __C86__". Did you find new issues with C86?

ghaerr:
Yes, some code in the otherwise-straightforward render.c::R_Line_FloodFill() function is causing a C86 compiler fault. I haven't chased that down yet, as I wanted to get paint published first.

toncho11:
Awesome! For me the easiest is an HDD image or if you can approve this pull request already.

rafael2k:
Btw, I can see in the Makefile.owc, you are using the large model. I'm also using large model for my elks-viewer using OWC. One suggestion I'd make for 0.9 release - add out-of-the-box support for compilation of all supported memory models in owc for ELKS, and may be keep the current library path for the large model (as a symlink or not), in order not to break software expecting the large model library in the current path.

ghaerr:
@toncho11, here's an HD32 image for you, I'm still working on finalizing the PR for commit.
[hd32-minix.img.zip](https://github.com/user-attachments/files/19275346/hd32-minix.img.zip)

ghaerr:
> add out-of-the-box support for compilation of all supported memory models in owc for ELKS, and may be keep the current library path for the large model

Yes, that's been on my radar for quite a while, and I want to get it done well before the 0.9 release. Although it initially sounds fairly easy, there are some complications with the fact that currently building the library for another model means deleting the \*.obj files for the previous models. This complicates things for debugging, when one needs to use `wdis` in order to confirm code generated, for example. We also have the issue of what the default compilation model should be in the `ewcc` and `ewlink` scripts. Although its easy to just hack a quick make for all models, these other issues still need addressing, and I haven't figure out an answer for those yet.

The other big issue is when exactly should the ELKS tree build the OWC or C86 versions of libraries and/or applications. I'm leaning towards the idea of looking for the WATCOM= or C86= variables set to enable that, which is what is used in this PR, but this also is a bit complicated because that means various environment variables control the output of the build as well as  potentially the image(s) produced, which already is an issue with the Github CL, since C86 isn't available as a Debian package. Perhaps C86 needs to be moved into being a tool in the ELKS tools directory - but I also like the idea of keeping it as an outside repo.

ghaerr:
@toncho11, if the cursor display is too slow on your Amstrad, there is a USE_SMALL_CURSOR define in cursor.c that builds paint using the smaller cursor we had to use for Nano-X. That will be available as soon as I commit this, let me know how fast or slow this is on your real hardware. We can also turn off floating point altogether with a define in render.c, that reduces program size by 7k and the C86 version seems to run pretty well without any floating point, actually.

toncho11:
@ghaerr I think there is no paint in both the floppy and hdd image. In which folder it should be?

ghaerr:
> there is no paint in both the floppy and hdd image. In which folder it should be?

It should be in /bin - did you try just running `paint`?

ghaerr:
The paint program also requires a bunch of images which are in /lib/\*.bmp, as kind of discussed above. These should also be on the images.

rafael2k:
> > add out-of-the-box support for compilation of all supported memory models in owc for ELKS, and may be keep the current library path for the large model
> 
> Yes, that's been on my radar for quite a while, and I want to get it done well before the 0.9 release. Although it initially sounds fairly easy, there are some complications with the fact that currently building the library for another model means deleting the *.obj files for the previous models. This complicates things for debugging, when one needs to use `wdis` in order to confirm code generated, for example. We also have the issue of what the default compilation model should be in the `ewcc` and `ewlink` scripts. Although its easy to just hack a quick make for all models, these other issues still need addressing, and I haven't figure out an answer for those yet.

Got it. Concerning ewcc and ewlink I see them mostly as a reference, and hard-coding to some model seems ok to me. I would think compiling for each model, than cleaning up (just like in C86 for host and elks), or adding some suffix to the object names.

> The other big issue is when exactly should the ELKS tree build the OWC or C86 versions of libraries and/or applications. I'm leaning towards the idea of looking for the WATCOM= or C86= variables set to enable that, which is what is used in this PR, but this also is a bit complicated because that means various environment variables control the output of the build as well as potentially the image(s) produced, which already is an issue with the Github CL, since C86 isn't available as a Debian package. Perhaps C86 needs to be moved into being a tool in the ELKS tools directory - but I also like the idea of keeping it as an outside repo.

If it helps, I can create the "debian" directory for 8086-toolchain package creation.

ps: I also support the idea of moving the tools to ELKS repository.

toncho11:
I had two HDDs in 86box. 
It works!
I am testing on Amstrad 10 Mhz emulation and I see a similar problem with the mouse movement as with Nano-x (before you fixed it). The mouse is very delayed and slow. You need to move is slowly or it is unusable. But once you right click and start to draw it improves and its almost correct. 

Great job!

ghaerr:
>  ewcc and ewlink I see them mostly as a reference, and hard-coding to some model seems ok to me.

Well, they're pretty important wrappers, just like our `ecc` wrapper and come in very handy when you want to quickly compile and link something rather than writing a Makefile. I plan on combining the two and acting more like `ecc` does, where the link is automatic.

> adding some suffix to the object names.

You might notice we're already using .oaj (ia16), .obj (owc) and .ocj (c86) right now - so we need 4 more for the OWC .obj files. I'm trying to keep them to three letters so as to not get out of hand - but yes that's an idea I've thought will be fairly easy to implement.

>  I can create the "debian" directory for 8086-toolchain package creation.

No - frankly I don't want to create a debian package and have to support it for debian distribution, at least not for a while. Nobody uses the compiler but us.

> I also support the idea of moving the tools to ELKS repository.

I'm thinking the best answer would be to use a sub-repo, where the build grabs it from 8086-toolchain and builds it locally. Not sure yet what to do, but we're fast approaching having to do something!

Getting `paint` out will help a lot, that's why I worked to get it running on all three compilers. We also need to think about whether or not you want elks-viewer or your other projects built onto images automatically. So there needs to be a simple way to do it, if we go that direction.

ghaerr:
> The mouse is very delayed and slow. You need to move is slowly or it is unusable. But once you right click and start to draw it improves and its almost correct.

Thanks for the testing. I think what we need is a way to determine how slow a system is, and use a different mouse cursor automatically. The smaller cursor isn't nearly as nice as the large one, so I hesitated on adding the small one as default.

We could add a keyboard key to switch mouse cursors... ?

I am a bit surprised that the mouse seems to act more normally when actually drawing - I'm wondering why that is. It would be interesting to still try rapid mouse movements while drawing. Another idea could be to turn off floating point on the IA16 version, to see what happens. I may make that the default. Unfortunately, we can't turn on or off floating point emulation at run time, only at compile time.

If the mouse is almost unusable on 8088 systems, then we need to address that for sure.

rafael2k:
> Getting `paint` out will help a lot, that's why I worked to get it running on all three compilers. We also need to think about whether or not you want elks-viewer or your other projects built onto images automatically. So there needs to be a simple way to do it, if we go that direction.

I'm looking at it, and thinking if I should port elks-viewer to ia16-gcc. Anyway, I'll implement some target  in the Makefile to copy elks-viewer to the rootfs_template. I think Lua will be nice to have, may be zlib, and in the future bobcat, but I'm not sure if this will be ready soon.

ps: There is a paint.tar inside the commit. Is that supposed to be there?

toncho11:
Ok. @ghaerr you somehow predicted me and wrote the response for the cursor before my comment :)

toncho11:
Can the smaller cursor be set as a command line parameter or env variable?

ghaerr:
> There is a paint.tar inside the commit. Is that supposed to be there?

No - I thought I deleted that, let me check thanks.

> I'm looking at it, and thinking if I should port elks-viewer to ia16-gcc

You'll see that all the code is in graphics.c, which is nice. I looked at your expansive enhancements to graphics.c in elks-viewer, but haven't had time to add all that. I think the next step will be to try to put all this stuff into a graphics portion of libc, which is why I added paint into the "elkscmd/gui" directory, rather than elkscmd/paint, for the time being. The elkscmd/tui curses code also needs placing in libc. Let me think about all that.

ghaerr:
> Can the smaller cursor be set as a command line parameter or env variable?

Yes, good ideas. I'm thinking 'm' or 'M' inside the app for small and large cursor, as well as perhaps -m for small cursor on the command line or perhaps a SMALLCURSOR= environment variable.

> You somehow predicted me and wrote the response for the cursor before my comment :)

LOL! Yes, I have gotten to know you better over these last four years... I can hear @toncho11 now asking "How can this be made easier and where is the documentation for it" now before I finish writing code!! :)

rafael2k:
> > There is a paint.tar inside the commit. Is that supposed to be there?
> 
> No - I thought I deleted that, let me check thanks.
> 
> > I'm looking at it, and thinking if I should port elks-viewer to ia16-gcc
> 
> You'll see that all the code is in graphics.c, which is nice. I looked at your expansive enhancements to graphics.c in elks-viewer, but haven't had time to add all that. I think the next step will be to try to put all this stuff into a graphics portion of libc, which is why I added paint into the "elkscmd/gui" directory, rather than elkscmd/paint, for the time being. The elkscmd/tui curses code also needs placing in libc. Let me think about all that.

I'll port elks-viewer to ia16-gcc asap, so the tools could go to ELKS repo, while I'll also keep elks-viewer repo for development. I'll try to copy the idea / Makefiles from here.

I extended a bit the curses code for bobcat (lynx fork), and not all worked fine, I'm still evolving some bits. I would have some suggestions on the curses support. The problem with lynx I'm having now is that it uses of curses windows for the forms, but I think that can be rewritten to use just a line located fixed at lynx prompt.

ghaerr:
> ... and elks-viewer.  I think Lua will be nice to have, may be zlib, and in the future bobcat

Yes, it would be nice to have all of your ELKS projects built into a distribution somehow. Moving them all into ELKS repo is probably unrealistic, especially for you to enhance and maintain them, possibly for real Linux versions etc. If we had a directory that specified outside repos which were automatically grabbed and built underneath ELKS for a "make dist" or something, which then used a "make install" entry point in the master Makefiles for each sub-repo, then they could be automatically built and installed into /bin, /usr/bin or elsewhere automatically as part of the "make dist" process of building a HD distribution.

rafael2k:
> > ... and elks-viewer.  I think Lua will be nice to have, may be zlib, and in the future bobcat
> 
> Yes, it would be nice to have all of your ELKS projects built into a distribution somehow. Moving them all into ELKS repo is probably unrealistic, especially for you to enhance and maintain them, possibly for real Linux versions etc. If we had a directory that specified outside repos which were automatically grabbed and built underneath ELKS for a "make dist" or something, which then used a "make install" entry point in the master Makefiles for each sub-repo, then they could be automatically built and installed into /bin, /usr/bin or elsewhere automatically as part of the "make dist" process of building a HD distribution.

Indeed. We could provide a Makefile template for all the projects who want to integrate to the distributions build system - that would be nice!

PR #2260, opened 2025-03-17
[ow libc] Build Watcom compact, small, medium and large model libraries

ghaerr:
Now, running `make owc` will build all four compilation models for the ELKS OpenWatcom C libraries.

These four libraries are now named `libc/libc{c,s,m,l}.lib` for compact, small, medium and large models, instead of the old `libc/libc.lib` which was only built for large model.

Internally, the libc build uses some new extensions (.ocj, .osj, .omj and .olj) for each model's .obj file. The previously named C86 .ocj files are now named .o8j. None of this should matter for any user program Makefiles.

`ewcc` and `ewlink` will default to large model, unless libc/watcom.model is changed.

All external projects will have to update the name of the OWC library as above, or a link error will result. Nothing else should have to be changed. (FYI @rafael2k, @tyama501, @toncho11 and @FrenkelS).

Tested on elkscmd/gui (paint) as well as the C86 toolchain, which use OWC. The gui `paint` program now defaults to small model, since large was not actually needed.


toncho11:
Does the example here needs to be updated: https://github.com/ghaerr/elks/wiki/Using-OpenWatcom-C-with-ELKS#building-an-example-program-using-openwatcom to use the correct libc{c,s,m,l}.lib

ghaerr:
Thanks @toncho11, I updated the C library build instructions on the wiki.

rafael2k:
I'm on holidays out home, but as soon as I come back, beginning of January, I'll update my projects using OW, especially Lua, which I will already update to the latest Lua 5.5. 

This is very useful and makes it simple to choose any OW memory model, thanks a lot!

PR #2261, opened 2025-03-17
[cmds] Fix mouse release bug in paint event code

ghaerr:
Fixes occasional problem where mouse button up processing was delayed until after next mouse movement in portable event code used in `paint`. 

Also adds 'm' command to `paint` to switch between 16x16 and 8x8 mouse cursor for slow systems. Discussed with @toncho11 in https://github.com/ghaerr/elks/pull/2259#issuecomment-2727669559.

`paint` main loop moved from polling to hanging on select for all events, thus reducing CPU load for any other applications running.

Event code cleaned up and mouse position returned in all events. This code will be ported back to 8086-toolchain/examples in the future.


toncho11:
Drawing is probably a bit slow so the user tend to keep mouse down and moving. So after releasing the mouse button paint continues to draw for 1 or 2 seconds more on my Amstrad because of the extra buffered events.

Also ctrl+z will be usefull to remove the last mouse down until mouse up (the last drawing).

PR #2267, opened 2025-03-22
[net] Fix network hang in bind() when only single TCP socket open

### LINKED ISSUE #2266: urlget blocks in bind() at second run
When running urlget a second time to a remote http server, it gets blocked in bind(), and system freezes.

It seems that when the ftpd server is running, urlget keeps working as expected after many runs.

I observe the same behavior in other software, like bobcat.

ghaerr:
Fixes #2266. Looks like this was in ktcp for a very long time.


rafael2k:
ps: there is an extra:
"tcpcbs = n;"

PR #2268, opened 2025-03-24
[kernel] Add XMS RAM drive /dev/ssd

ghaerr:
This PR implements a RAM drive based on XMS memory for ELKS applications. 

Based on discussion in https://github.com/ghaerr/elks/issues/2256#issuecomment-2734062479 as well as #2254, the intention is to provide a simple way forward to allow large programs, such as web browsers or text editors to quickly swap internal data to and from disk (XMS ramdisk in this case) without having to add XMS memory read/write routines directly into the application.

This option is configured in menuconfig under Filesystem Support by selecting `Use XMS Memory` (CONFIG_FS_XMS) and then selecting `XMS RAM Disk` (CONFIG_FS_XMS_RAMDISK). XMS system buffers continue to work as before, if selected.

The system doesn't currently check that there is in fact enough XMS memory (memory above 1M address) in the system, so one must be careful not to over-allocate the RAM allocation when running `ramdisk`.

The `ramdisk` utility is used to create the RAM disk (current max 16Mb), here is an example of creating and using a 3MB MINIX ramdisk:
```
ramdisk /dev/ssd make 3072
mkfs /dev/ssd 3072
fsck -lvf /dev/ssd
mount /dev/ssd /mnt 
df /dev/ssd
...
umount /dev/ssd
ramdisk /dev/ssd kill
```
For use in swapping memory quickly to disk, the ramdisk may want to be mounted on /tmp, or wherever is best suited for the programs that will use it.

Note that we are almost out of kernel text address space, so the regular (main memory) RAM disk (CONFIG_BLK_DEV_RAM) should not be selected. This barely leaves room for networking, UNIX sockets, and all the other functions recently being used by OpenWatcom programs. This issue will be fixed in a later PR.


toncho11:
Awesome!
What is "ssd" in this case? "Solid state drive" does not make enough sense.
How `ramdisk /dev/ssd make 3072` knows that /dev/ssd should be linked to XMS memory exactly?
Or "/dev/ssd" is alsways the XMS memory? I think I am just confused by the name. I was expecting /dev/xms I suppose.

toncho11:
Tested. It works. 
@rafael2k 

We can have this in /etc/rc.sys

```
# Try creating ramdisk using XMS memory
/bin/ramdisk /dev/ssd make 1024

if [ $? -eq 0 ]; then
    /bin/fsck -lvf /dev/ssd
    /bin/mount /dev/ssd /tmp/swap
fi
```

toncho11:
We can also have compressed executables cached on /tmp/swap.
We can also have an ISA card based on RaspPI for example that acts as a ramdrive for older 8086 systems for /tmp/swap.

rafael2k:
I'm also very interested in this development, as it is really a portable way to support XMS. I'd cheer for XMS on 286 too using loadall function. Does @Mellvik TLVC support XMS on 286?

Mellvik:
@rafael2k - ELKS and TLVC both support xms on some 286 systems: IFF xms access is available via that system's BIOS - aka 'the int15/87 method'.

PR #2270, opened 2025-03-27
[kernel] Add support for loading kernel in HMA

ghaerr:
Adds the ability to run the ELKS kernel in the high memory area (HMA) which is XMS memory above 1MB, when possible.

Inspired by @Mellvik's kernel HMA implementation in https://github.com/Mellvik/TLVC/pull/159, thanks @Mellvik!

Allowing the kernel to run in HMA frees up ~62k of main memory, allowing programs to use more memory than ever before: where in normal operation the maximum main memory was 478k, using kernel HMA shows free memory of 540k!

There is no compile-time configuration for enabling kernel HMA, instead the option is enabled using `hma=kernel` in /bootopts. This option is default ON for testing with hopes it's permanent subject to more user testing.

Kernel HMA is active when the boot screen shows "Kernel text ffff ...", which shows the kernel .text segment is 0xFFFF, the segment address of HMA memory.

The requirements for kernel HMA are having 64K or more of XMS memory, and the ability to set the "A20 gate". Both of these options are required for XMS buffers and the recent XMS ramdisk enhancement. I have a couple changes and a bit more testing to integrate the XMS ramdisk properly and will then commit this PR.

Running the kernel in HMA is limited to IBM PC only for now; @tyama501, do you know whether PC-98 supports HMA or XMS memory? There may also be a different mechanism for turning the A20 gate on, if it is not done automatically. If the PC-98 supports HMA, it should be straightforward to implement it.

Currently, testing is showing that rebooting the system using `shutdown -r` (new `reboot`)fails when kernel HMA is active. The internal reboot code jumps to FFFF:0000, which also happens to be the HMA segment. From running `hd ffff:0000`, it appears that the setup.S copy of the kernel is overwriting the "ROM" at FFFF:0000 with zeros from the kernel .text segment, rather than keeping the ROM reboot code which resides at the same address.

@Mellvik, I'd be interested to know whether your HMA kernel implementation reboots on real hardware, I would guess it does, but does not on QEMU. 

I plan on looking further into this problem tomorrow before commit.


tyama501:
Wow, nice!

I have been very busy moving my house recently, but I will look it after I settle down.

We have already implemented A20 for PC-98 when we implemented the xms buffer.

https://github.com/ghaerr/elks/blob/master/elks/arch/i86/lib/bios-xms-pc98.S

Thank you.

ghaerr:
Thanks @tyama501 for your quick response, and good luck moving!

I'll add those PC-98 functions into a20.inc so that kernel HMA should work on your systems also. I will finish it and you can test it after you have moved in to your new place and have more time :)

Mellvik:
> @Mellvik, I'd be interested to know whether your HMA kernel implementation reboots on real hardware, I would guess it does, but does not on QEMU.

Yes, I've seen this briefly. It's related to the `reboot` changes in ELKS #1978. I imported the change for TLVC for testing in #158, and it never made it beyond QEMU because of this hang situation. 

Having a working version of `reboot` on all hardware systems, I put the problem on the (wait) list and postponed looking into it.

Mellvik:
> There is no compile-time configuration for enabling kernel HMA, instead the option is enabled using hma=kernel in /bootopts. This option is default ON for testing with hopes it's permanent subject to more user testing.

Wow @ghaerr, /bootopts processing in setup.S. Really nifty - I'll certainly copy that one back to TLVC!

elks/kernel/sys.c
Mellvik, line 1: The `disable_a20_gate()` call will most likely prevent the rest of the code here to be executed if the kernel is loaded high.

Mellvik: OK, I see it was fixed later.

ghaerr:
@tyama501, I have added support for PC-98 for kernel HMA, but we now ask the BIOS how much XMS memory is available, so that HMA, XMS buffers, XMS ramdisk, etc don't run out of XMS space. The IBM PC BIOS function for this is INT 15h, AX=8800h, and returns AX=kbytes of XMS RAM. Can you find the equivalent function for PC-98? I have left this commented out for the time being and supposed 2M XMS for PC-98 in setup.S for now. Thank you!

ghaerr:
>  testing is showing that rebooting the system using shutdown -r (new reboot) fails

> It's related to the reboot changes in ELKS https://github.com/ghaerr/elks/pull/1978. I imported the change for TLVC for testing in https://github.com/ghaerr/elks/pull/158, and it never made it beyond QEMU because of this hang situation.

I debugged this and found the problem is not related to #1978, but in QEMU, the setup.S code was copying the first 16 bytes of segment FFFF on top of  the BIOS ROM and overwriting it. QEMU isn't setting FFFF:0000 read-only and the hard-reset address (used for reboot) is at F000:FFF0 (which is the same as FFFF:0000). Thus the system crashed instead of rebooting. 

The fix is to not copy the 16 bytes into the ROM location when loading kernel HMA, just pushed.

Mellvik:
> > testing is showing that rebooting the system using shutdown -r (new reboot) fails
> 
> > It's related to the reboot changes in ELKS #1978. I imported the change for TLVC for testing in #158, and it never made it beyond QEMU because of this hang situation.
> 
> I debugged this and found the problem is not related to #1978, but in QEMU, the setup.S code was copying the first 16 bytes of segment FFFF on top of the BIOS ROM and overwriting it. QEMU isn't setting FFFF:0000 read-only and the hard-reset address (used for reboot) is at F000:FFF0 (which is the same as FFFF:0000). Thus the system crashed instead of rebooting.
> 
> The fix is to not copy the 16 bytes into the ROM location when loading kernel HMA, just pushed.

Very nice catch @ghaerr! A(nother) QEMU bug really. There have been quite a few over the past year. Rather interesting that it worked with the old `reboot` and not the new one. Hard to make sense of that one. Anyway, a smart fix, Thanks!

ghaerr:
> Rather interesting that it worked with the old reboot and not the new one. Hard to make sense of that one. 

I would imagined it "just happened" to work, since after the "zeroes" are executed at FFFF:0000 for 16 bytes, the kernel \_start code is reentered in crt0.S. Since setup.S passes information to \_start using registers, it just depends on what they are as to whether crt0.S succeeds or not. So it likely wasn't actually rebooting but restarting. That would be my guess, since the `hard_reset_now` code never changed in the reboot code changes.

toncho11:
I tried to boot his new branch and I can not on copy.sh. 
It uses SeaBIOS. I used this floppy build:
[fd1440_minix.zip](https://github.com/user-attachments/files/19494210/fd1440_minix.zip)

toncho11:
Here is a screenshot
image

toncho11:
I did a full rebuild with build.sh and now it seems OK.

ghaerr:
Thanks for testing @toncho11!

ghaerr:
@toncho11, did you happen to notice on copy.sh whether the boot screen message "Kernel text 330" line was that, or "Kernel text ffff"? If 330, then that's a normal boot, with the kernel not in HMA. Your screenshot above shows 330.

I'm curious as to whether copy.sh emulates xms: as described above the kernel will only be located in HMA if XMS memory is available, the A20 gate can be enabled, and hma=kernel is in /bootopts. The config file does not have to specify "Use XMS Memory" (CONFIG_FS_XMS), which is default off.

toncho11:
There is always some problem with these config files. 
I just tested on real 486. And everything seems to work. I can confirm that memory increases from 500kb to 564 kb.
XMS ramdrive also works.

toncho11:
Yes. Copy.sh does not seem to have the 64 kb extra memory.

ghaerr:
Pushed further enhancement to allow CONFIG_FS_XMS=y and CONFIG_FS_XMS_BUFFERS=y set ON by default, since the XMS memory size is now queried, and XMS will be disabled and EXT buffers used when XMS isn't present. I'm thinking of defaulting these two options and also CONFIG_FS_XMS_RAMDISK=y to be default ON for the standard IBM PC and PC-98 builds.

XMS ramdisk integration is now complete and working. The startup XMS boot messages won't repeat regardless of whether XMS buffers and XMS ramdisk are configured. This is fixed from previously where the disabled/error message would repeat from the buffer and ramdisk systems.

Tested PC-98 pretty heavily, everything works. All that's left is to determine how to query the PC-98 BIOS for the XMS memory size. It is currently faked at 4M in setup.S for the time being.

The HMA kernel doesn't require CONFIG_FS_XMS to be enabled, and will function automatically if able when set in /bootopts.

Overall, the system can now probably be configured to automatically take advantage of kernel HMA as well as XMS buffers and ramdisk, and not fail if XMS is not present. This is a nice enhancement from previous operation; the system now tries to automatically take advantage of all memory present, but fails gracefully when memory is not present. The number of XMS buffers to use can be set dynamically using xmsbuf= in /bootopts, and the fallback EXT buffers can be set using buf= in /bootopts.

The only big XMS configuration issue remaining is that some systems will require the "INT 15" (CONFIG_FS_XMS_INT15) option set in order to funtion with XMS. These systems should still at least boot without the option, but can't easily be made to switch from unreal mode to INT 15 mode at runtime.

toncho11:
I also tested with 86box and it works. 
I tried all /bootopts options to get more memory. 
I only get a problem / panic when I enable UMB on 86box. But this is a bit BIOS specific. The problem is there even if no HMA is used.

tyama501:
> Tested PC-98 pretty heavily, everything works. All that's left is to determine how to query the PC-98 BIOS for the XMS memory size. It is currently faked at 4M in setup.S for the time being.
> 

Thank you!
I will check if there is a way to detect it.

ghaerr:
> I only get a problem / panic when I enable UMB on 86box. But this is a bit BIOS specific. 

Yes, different topic but I'm still interested to know what happened. Can you remember the specific panic or post a screenshot? Is there actually memory allocated at the UMB location? If not, yes, we'd likely get a panic.

toncho11:
So both XMS and HMA will be compiled by default.

XMS enabled by default.
HMA only enabled when `hma=kernel`, by default not enabled

We need to make sure that such 360 KB image does work (configured without the `hma=kernel`) - does boot without panic.

ghaerr:
> HMA only enabled when hma=kernel, by default not enabled
We need to make sure that such 360 KB image does work (configured without the hma=kernel) - does boot without panic.

My thoughts were to have HMA enabled by default. If the system doesn't have XMS/HMA, then the kernel will load normally.

The 360k image isn't any different from the others, so if it or the others boot, they all should. A 360k image with hma=kernel set should boot on your Amstrad no problem, even with CONFIG_FS_XMS=y set. 

I plan on making changes to the standard config files in a separate PR though, after a bit more testing.

toncho11:
> > HMA only enabled when hma=kernel, by default not enabled
> > We need to make sure that such 360 KB image does work (configured without the hma=kernel) - does boot without panic.
> 
> My thoughts were to have HMA enabled by default. If the system doesn't have XMS/HMA, then the kernel will load normally.
> 
> The 360k image isn't any different from the others, so if it or the others boot, they all should. A 360k image with hma=kernel set should boot on your Amstrad no problem, even with CONFIG_FS_XMS=y set.
> 
> I plan on making changes to the standard config files in a separate PR though, after a bit more testing.

The computer is different. Some errors will only pop up when the XMS and HMA are really not there. We need to test that the code that checks for XMS and HMA does not halt a 8086 because of some illegal instructions for example.

ghaerr:
> We need to test that the code that checks for XMS and HMA does not halt a 8086 because of some illegal instructions for example.

Yes. I have been very careful with that - the entire XMS/HMA enabling is now based on a single (newer) BIOS call - INT 15h AX=8800h Get Extended Memory Size. This BIOS call should return Carry Set if not implemented, otherwise the size of XMS memory. If not implemented or zero, then neither HMA nor XMS will be turned on, and no attempts at running 386 code (to attempt turning on unreal mode) will be attempted.

So agreed - we need to test the BIOS Get Extended Memory call on Amstradt. For all the emulators out there, they should report XMS size properly, but we'll find out shortly, it sounds like.

toncho11:
> > We need to test that the code that checks for XMS and HMA does not halt a 8086 because of some illegal instructions for example.
> 
> Yes. I have been very careful with that - the entire XMS/HMA enabling is now based on a single (newer) BIOS call - INT 15h AX=8800h Get Extended Memory Size. This BIOS call should return Carry Set if not implemented, otherwise the size of XMS memory. If not implemented or zero, then neither HMA nor XMS will be turned on, and no attempts at running 386 code (to attempt turning on unreal mode) will be attempted.
> 
> So agreed - we need to test the BIOS Get Extended Memory call on Amstrad. For all the emulators out there, they should report XMS size properly, but we'll find out shortly, it sounds like.

If I remember correctly one time we had the XMS enabled by default. There was no check if the XMS was really available and so the system was crashing on 8086.

tyama501:
Hello @ghaerr 

I have heard PC-98 extended memory amount can be obtained from the system memory 
0000:0401h and 0000:0594h.
I will study the detail.

Thank you.

ghaerr:
> I have heard PC-98 extended memory amount can be obtained from the system memory
0000:0401h and 0000:0594h.

Very nice. This agrees with some FreeBSD on PC-98 code I found here: https://web.mit.edu/freebsd/head/sys/boot/pc98/libpc98/biosmem.c:
```
    bios_extmem = *(u_char *)PTOV(0xA1401) * 128 * 1024 +
	*(u_int16_t *)PTOV(0xA1594) * 1024 * 1024;
```

I read elsewhere that the FreeBSD code had copied the low memory area to 0xA1000 for safekeeping, but this shows the proper manipulation of the BIOS memory are to get the extended memory size.

~When running DosBoxX, it seems both these locations are zero, do you know if DosBoxX support XMS emulation, or is that only available through Neko?~

PR #2274, opened 2025-03-30
[kernel] Support XMS unreal mode or INT 15/1F dynamically at runtime

ghaerr:
Removes compile-time CONFIG_FS_XMS_INT15 option and allows for either automatic unreal mode or INT 15/1F operation, determined at XMS startup. Adds about 500 bytes to the kernel.

This PR serves as a framework for the XMS/HMA problems discussed in https://github.com/Mellvik/TLVC/pull/160#issuecomment-2763961450. Further testing is being performed by @Mellvik to ascertain the exact behavior of certain systems (e.g. Compaq 286 and 386 systems) that either can't operate using unreal mode (thus requiring INT 15 block move), or can't have kernel HMA with XMS activated, since its BIOS INT 15 block move disables A20.

This PR combines simultaneous XMS support using either unreal mode or INT 15 block move, while also allowing kernel HMA  at boot, for all CPUs and IBM PC and PC-98 systems. XMS buffers are not turned off by default using /bootopts xmsbuf=0, since we should now be able to determine exactly what XMS support is available (or not) and set up automatically.

The two problem cases, that of a 386 machine not able to use unreal mode, and a 286 or 386 machine requiring INT 15/1F disabling A20 after block moves (thus prohibiting kernel HMA) are separately handled with dummy function calls ForceINT15() and INT15DisablesA20() at the top of xms.c for now.

After implementing combined unreal mode and INT 15/1F support and running testing on the IBM PC version on QEMU and PC-98 on DosBox, I have come up with the following observations that IMO should be able to get everything working well in short order, subject to @Mellvik's testing:
- If no XMS memory is available, neither kernel HMA nor XMS will ever be enabled. This is what happens on 8086 machines.
- If A20 gate is tested operational and hma=kernel, then the kernel is loaded into HMA.
- If the CPU is an 80286 or ForceINT15() is true, INT 15 (or INT 1F for PC-98) will be planned.
- If INT15/1F planned to be used and kernel is HMA but INT15DisablesA20(), XMS will be disabled.
- If INT15/1F not planned to be used, check_unreal_mode() is called, and unreal mode enabled and planned if OK'd.
- XMS is enabled in the planned mode: unreal or INT 15.

Sounds a bit complicated, but isn't really. A bit more complicated is whether INT15 should be forced: yes on Compaq 386 BIOSes, otherwise probably not, and then whether INT 15 disables A20 on block moves: yes on Compaq 286 and 386 BIOSes, but no on PC-98, for example.

All of the above logic is contained in the following rewrite of xms_init():
```
#define ForceINT15()        0       /* =1 to simulate Compaq 386 BIOS requiring INT 15 */
#define INT15DisablesA20()  0       /* =1 when BIOS INT 15 disables A20 */
...
    /* 80286 machines and Compaq BIOSes can't use unreal mode and must use INT 15/1F */
    if (arch_cpu <= 6 || ForceINT15()) {
        if (kernel_cs == 0xffff && INT15DisablesA20()) {
            /* BIOS INT 15 block_move disables A20 on some systems! */
            printk("disabled w/kernel HMA and int 15\n");
            return XMS_DISABLED;
        }
        printk("int 15/1F, ");
        enabled = XMS_INT15;
    } else {
        if (check_unreal_mode() <= 0) {
            printk("disabled, requires 386, ");
            return XMS_DISABLED;
        }
        enable_unreal_mode();
        printk("unreal mode, ");
        enabled = XMS_UNREAL;
    }
```

Finally, it seems that both ForceINT15() and INT15DisablesA20() can be set to 0, unless the system is running a Compaq 286 or 386 BIOS. That is, IBM PC and PC-98 versions are running properly on QEMU and DosBoxX with all variations of HMA and XMS buffers, automatically picking the right XMS mode by using arch_cpu == 80826 to determine INT 15 or unreal mode. QEMU INT 15 and PC-98 INBT 1F block moves don't disable A20, for example. Until @Mellvik tests otherwise, only the Compaq machines seem to have issues with either 386 CPU using unreal mode, or 286 CPU INT 15 disabling A20.

Researching the issue of whether a Compaq BIOS is being used or not, some good news showed up - all Compaq BIOSes have a Machine ID ('COMPAQ') at ROM address F000:FFEA that can be easily tested for during setup.S execution:
Compaq
This is further demonstrated with the Compaq 386 BIOS disassembly available at https://www.pcjs.org/machines/pcx86/compaq/deskpro386/rom/1988-01-28/1988-01-28.asm at the top of ROM:
```
;
;   Offset for storing CPU identifier (low byte) and revision number (high byte)
;
cpu_idrev_offset:
	dw	0x7FBE			; 0000FFE2  BE7F

	db	'G'			; 0000FFE4  47		; Product Family Code
	db	'4'			; 0000FFE5  34		; Point-Release Code
	db	'J '			; 0000FFE6  4A20	; Revision Code
	db	'03'			; 0000FFE8  3033	; BIOS Type Code
	db	'COMPAQ'		; 0000FFEA  434F4D504151; Machine ID

	jmp	0xf000:reset		; 0000FFF0  EA05F900F0

	db	0x20			; 0000FFF5  20
	db	'01/28/88'		; 0000FFF6  30312F32382F3838	; BIOS Revision Date
	db	0xFC			; 0000FFFE  FC		; Machine Type Code
	db	0x98			; 0000FFFF  98		; Checksum
```

My thoughts are that automatic XMS/HMA support can be completed then by having the ForceINT15() and INT15DisablesA20() functions call IsCompaqBIOS() (to be written in setup.S) but keeping the source flexible with the same-named functions should we find other systems that are incompatible with our XMS setup to be easily made to work.


Mellvik:
Thank you @ghaerr,
the good news is that all systems now handle the `check_unreal_mode()` fine, so the auto configure regime suggested will work well for all systems. Also, as pointed out before, I see no reason to assume that compaq is behaving different from other systems and needs special treatment. Being the only systems of their kind tested, it's premature to make assumptions. Case in point, also reported before - my AMI 386SX system behaves exactly like the compaq 386 when using INT15 XMS block move: The A20 is reset after every access.

It may actually still be possible to get the Compaq 386 to work with both HMA and XMS - I'm looking into that.

ghaerr:
> the good news is that all systems now handle the check_unreal_mode() fine, so the auto configure regime suggested will work well for all systems.

Thanks for that testing, that is good to know.

> Also, as pointed out before, I see no reason to assume that compaq is behaving different from other systems and needs special treatment.

> my AMI 386SX system behaves exactly like the compaq 386 when using INT15 XMS block move: The A20 is reset after every access

Got it. I misunderstood the AMI 386SX issue, so agreed my thinking this is Compaq-specific was incorrect. Does the AMI 386 work with unreal mode?

Perhaps many 386 systems disable A20 on INT 15 block moves, but QEMU does not, so some kind of (at least internal) option may be useful for QEMU testing of each of the various XMS/HMA/286/386 options.

> It may actually still be possible to get the Compaq 386 to work with both HMA and XMS

Hmmm. That, along with the AMI disabling A20 on block move, would imply we may not be able to get a fully automatic way of configuring XMS w/HMA. A solution to that could take the form of a /bootopts xms=int15,0 / xms=int15,1 where 1 or 0 specifies whether A20 is disabled by INT 15 or not. PC-98 hardware and QEMU don't disable A20 on INT 15.

This is all potentially a big can of worms, because, on almost all systems, there is in fact a way to enable unreal mode and perform block moves - just don't make any other BIOS calls. That's how Linux works on all of them. I haven't fully checked, but I think the only ELKS BIOS call in regular use is INT 13h disk I/O, which presumably causes the problem with protected-mode BIOSes. For TLVC, or for ELKS configured with the DF driver and no HD, it may be possible to run the kernel w/o INT 15 at all. However, in my initial XMS implementation and testing years ago, I thought I found that the kernel crashed immediately after enabling unreal mode on my Compaq Portable 386. This could be the result of the timer interrupt, possibly?

To me, this is pointing towards creating /bootopts options as described above to be able to quickly test differing configurations, to allow us or the user to get their machine working given the myriad of possible configurations affecting the whole.

Mellvik:
> 
> > my AMI 386SX system behaves exactly like the compaq 386 when using INT15 XMS block move: The A20 is reset after every access
> 
> Got it. I misunderstood the AMI 386SX issue, so agreed my thinking this is Compaq-specific was incorrect. Does the AMI 386 work with unreal mode?

It does - like QEMU it works fine in both modes. 
> 
> Perhaps many 386 systems disable A20 on INT 15 block moves, but QEMU does not, so some kind of (at least internal) option may be useful for QEMU testing of each of the various XMS/HMA/286/386 options.

I think most do. As to QEMU, it is documented somewhere - I forget where - that it boots with A20 on, and never turns it off, certainly not with BIOS block moves (which is strictly a BIOS issue, but that doesn't mean much if the underlying emulator simply has it turned on, period).

> 
> > It may actually still be possible to get the Compaq 386 to work with both HMA and XMS
> 
> Hmmm. That, along with the AMI disabling A20 on block move, would imply we may not be able to get a fully automatic way of configuring XMS w/HMA. A solution to that could take the form of a /bootopts xms=int15,0 / xms=int15,1 where 1 or 0 specifies whether A20 is disabled by INT 15 or not. PC-98 hardware and QEMU don't disable A20 on INT 15.

I my view it's simpler than that. If there is indeed a bootopts switch for this, if set to int15 it means no HMA, since the combination cannot work safely. It means the system will hang if set incorrectly but that will be the case for any variant as I see it.

My plan is to use a hma= setting in bootopts instead, thereby separating xms and hma, possibly with binary codes instead of literals, like 0=off,1=kernel,2=bufheads, 4=kernel and bufheads, the latter if there is enough space on top of the hma-kernel. Part of the last PR, and working fine for the past few days, typically xmsbufs=128 or 64 bufs in low mem.

> 
> This is all potentially a big can of worms, because, on almost all systems, there is in fact a way to enable unreal mode and perform block moves - just don't make any other BIOS calls. That's how Linux works on all of them. I haven't fully checked, but I think the only ELKS BIOS call in regular use is INT 13h disk I/O, which presumably causes the problem with protected-mode BIOSes. For TLVC, or for ELKS configured with the DF driver and no HD, it may be possible to run the kernel w/o INT 15 at all. However, in my initial XMS implementation and testing years ago, I thought I found that the kernel crashed immediately after enabling unreal mode on my Compaq Portable 386. This could be the result of the timer interrupt, possibly?

It could, but it doesn't sound likely. I will spend some time figuring out how far the compaq 386 can be pushed on this. I'm optimistic about the possibility to go all the way. I'm not sure that the theory about any INT call turning off A20 holds, I will verify that along the way. Just to be sure, which timer interrupt are you referring to?

ghaerr:
I'm getting a bit confused here with all the results. Obviously, there's an issue as to whether HMA can be used for the kernel, and also have XMS operational using INT 15. My understanding is that you're saying that with both your Compaq systems, as well as the AMI, this won't work.

The other issue I am thinking we got resolved - that check_unreal_mode crashed the system on some 386's, and/or that your 286 Compaq required INT 15. With this PR having INT 15 and unreal mode always available, arch_cpu is used to force all 286 systems to use INT 15. That's fits with all results, right?

Then, I thought you were saying that *some* 386 systems, Compaq 386 in particular, needed to have INT 15 set, as unreal mode would not work. Correct?

So we need a way of forcing INT 15 (whether using hma=number or xms=int15 etc doesn't matter) for those systems, right?

And for those systems where INT 15 is required, HMA will automatically be turned off, since the combination cannot work safely, right?

> if set to int15 it means no HMA, since the combination cannot work safely. It means the system will hang if set incorrectly but that will be the case for any variant as I see it.

My only addition to your test results is that both QEMU and PC-98 will work with INT 15 and HMA, and a way of allowing this would be nice for testing.

> I'm not sure that the theory about any INT call turning off A20 holds, I will verify that along the way. Just to be sure, which timer interrupt are you referring to?

The hardware timer interrupt IRQ 0. I don't have any idea what is causing the crash when unreal mode is turned on for Compaq 386 systems, I was just guessing. Perhaps their hardware just can't handle it and real Linux couldn't run on the machine anyways.

ghaerr:
Hello @tyama501, I have successfully added support for determining the PC-98 extended memory available, according to your comment https://github.com/ghaerr/elks/pull/2270#issuecomment-2764621788.

I have tested this on DosBox-X, and things work well. The default extended memory size is 15MB, and the total memory size (including the lower 1MB memory) can be set using memsize= in [dosbox]. The maximum XMS memory supported by ELKS is 64MB.

By default, the kernel is loaded into HMA, and XMS buffers and ramdisk are enabled, using INT 1F for block moves, as before. This has been tested on PC-98 w/286, I have not found any way to test this on 386-based PC-98s, if there are any.

Thank you!

Mellvik:
> I'm getting a bit confused here with all the results. Obviously, there's an issue as to whether HMA can be used for the kernel, and also have XMS operational using INT 15. My understanding is that you're saying that with both your Compaq systems, as well as the AMI, this won't work.

Yeah, I get confused too. When things that didn't work, now either work or change behavior, there is a need to go back and question previous assertions. It's time consuming but also rewarding when new opportunities arise.

> 
> The other issue I am thinking we got resolved - that check_unreal_mode crashed the system on some 386's, and/or that your 286 Compaq required INT 15. With this PR having INT 15 and unreal mode always available, arch_cpu is used to force all 286 systems to use INT 15. That's fits with all results, right?

yes, the 286 is covered with this.

> 
> Then, I thought you were saying that _some_ 386 systems, Compaq 386 in particular, needed to have INT 15 set, as unreal mode would not work. Correct?

also true, but that was before it became posdible to entrr unreal kode w/o hanging. Now it' pen again and I'm on it.

> 
> So we need a way of forcing INT 15 (whether using hma=number or xms=int15 etc doesn't matter) for those systems, right?

right, that's important regardless of this particular outcome.

> 
> And for those systems where INT 15 is required, HMA will automatically be turned off, since the combination cannot work safely, right?

yes.

> 
> > if set to int15 it means no HMA, since the combination cannot work safely. It means the system will hang if set incorrectly but that will be the case for any variant as I see it.

yes.

> 
> My only addition to your test results is that both QEMU and PC-98 will work with INT 15 and HMA, and a way of allowing this would be nice for testing.
> 
> > I'm not sure that the theory about any INT call turning off A20 holds, I will verify that along the way. Just to be sure, which timer interrupt are you referring to?

me too, I don't think this is the case.

> 
> The hardware timer interrupt IRQ 0. I don't have any idea what is causing the crash when unreal mode is turned on for Compaq 386 systems, I was just guessing. Perhaps their hardware just can't handle it and real Linux couldn't run on the machine anyways.

I'll be back on that.
Sorry about the spelling, small screen keyboard.

ghaerr:
[EDIT: This post is mostly incorrect and XMS startup redesigned yet again. See next post for details].

> Then, I thought you were saying that some 386 systems, Compaq 386 in particular, needed to have INT 15 set, as unreal mode would not work. Correct?

> also true, but that was before it became posdible to entrr unreal kode w/o hanging. Now it' pen again and I'm on it.

I'm interested to hear the results on entering unreal mode.

Actually, now that we're checking arch_cpu <= 6 (80286 or less) before even calling check_unreal_mode(), I think check_unreal_mode can be removed entirely. I'm not sure of the reason it was hanging your systems, ~(other than perhaps the SMSW instruction triggers a fault in real mode that crashes the system, now that I think of it),~ but all check_unreal_mode does is check we're on a 32-bit CPU, and then check we're not running in V86 mode. Arch_cpu will already be >= 7 in this case, so all it's doing is checking for V86 mode, pretty useless I think.

I may comment it out and run with that for a while, no full-kernel emulators I know of run in V86 mode except Windows DOSBOX which wouldn't boot ELKS anyways.

Thank you for your quick answers. I think I have enough information to continue with this PR. For ELKS, I am thinking of using two options, hma=kernel (for now, bufs maybe later) and xms=int15. With this approach, xms=int15 forces the use of INT 15 for XMS, for any system requiring it. If hma=kernel is also set on a machine like the Compaq's which can't handle it, the solution is to comment out hma=kernel. 

Since /bootopts will be distributed with hma=kernel default but xms=int15 commented out, ~all systems should boot, and load HMA if possible, and then the user can set xms=int15 to enable XMS after  a successful boot~. If that fails, then they'll also have to comment out hma=kernel.

To make it more automatic or informational, we would still need a way to determine if INT 15 disables A20, which would then disable XMS until hma=kernel is removed ~but it wouldn't hang after XMS is enabled~.  [EDIT: Oops - this won't work: a "test" INT 15 that disabled A20 when kernel HMA would crash the system immediately]. ~This could be tested for kernel boot time by performing an INT 15 block transfer, but I'm wondering if its even worth it. So for now I won't do that, and~ Manual removal of hma=kernel in /bootopts (or changing the Int15DisablesA20 define in xms.c to 1 to disable XMS) is required to boot the system.

ghaerr:
Whew, thinking through all the XMS and HMA issues with different machines got me kind of confused as to what design would always boot but also remain easy to understand and remember.

A simpler approach was designed that can guarantee that the ELKS distribution disks will always boot, and then the user can further configure /bootopts for their particular system. 

Since XMS and HMA support are now fully compiled in, successful booting on every system out there requires XMS off, unless enabled by /bootopts. Kernel HMA can remain default on.

The redesign uses three options in /bootopts:
```
            xms=on      # XMS on, uses unreal mode unless 286 which will automatically use INT 15
            xms=int15   # XMS on, force INT 15 for all (286 and 386+) systems
            hma=kernel  # kernel in HMA, may need to be removed for some xms=int15 systems
```
On the distribution disks, only `hma=kernel` will be set, since that can be fully automatically controlled.
- In general, for most all 286 and 386+ systems, to enable XMS use `xms=on`. This works for PC-98 and QEMU as well.
- For those 386+ systems that requires INT 15 support, use `xms=int15`. 
- Some systems disable the A20 gate when INT 15 is used. For these systems, comment out `hma=kernel`.

The new addition of `xms=on` also allows for easy testing of system internals and device driver non-XMS code paths without turning off CONFIG_FS_XMS.

For current testing, both `hma=kernel` and `xms=on` will be active in /bootopts. On the final distribution disks, `xms=on` will be commented out, enabling guaranteed booting on all systems.

The /dev/ssd XMS ramdisk now shows the remaining memory available in Kbytes after XMS buffer allocations, for more easily creating the maximum size ramdisk using `ramdisk /dev/ssd make Kbytes`.

@Mellvik: with the above design, I think check_unreal_mode() can be deleted, saving more space. It'll never get executed by 8086 or 286 CPUs, and now it's only benefit is automatic prevention of enabling XMS in 386 V86 mode, and XMS now defaults to being disabled anyways. What do you think?

rafael2k:
This is really quite a development! Thank you!!

Mellvik:
> > Then, I thought you were saying that some 386 systems, Compaq 386 in particular, needed to have INT 15 set, as unreal mode would not work. Correct?

That is correct and that was before. By now we have a lot more insight into the workings of a20 switching in particular which opens up for more retailed 'research' into what/where the actual problems are.
 
> 
> > also true, but that was before it became posdible to entrr unreal kode w/o hanging. Now it' pen again and I'm on it.
> 
> I'm interested to hear the results on entering unreal mode.

As is, and before I start experimenting, the Compaq 386 returns OK from check_unreal_mode and hangs when accessing anything above 1M, indicating that A20 is off. So I'm moving the enable_a20 call around looking for a 'hit'. Since we know that HMA is available when not touching XMS, our a20 code is indeed working.

> 
> Actually, now that we're checking arch_cpu <= 6 (80286 or less) before even calling check_unreal_mode(), I think check_unreal_mode can be removed entirely. I'm not sure of the reason it was hanging your systems, ~(other than perhaps the SMSW instruction triggers a fault in real mode that crashes the system, now that I think of it),~ but all check_unreal_mode does is check we're on a 32-bit CPU, and then check we're not running in V86 mode. Arch_cpu will already be >= 7 in this case, so all it's doing is checking for V86 mode, pretty useless I think.

I forget exactly what the world looked like back then, so let it be bygones. And I agree with you - the current situation allows a lot of simplifications. Unless someone is tempted to work on unreal mode on the 286, with load all and all it entails, sticking with cpu=7+ is safe and simple.

ghaerr:
> and hangs when accessing anything above 1M, indicating that A20 is off.

I wasn't aware our previous "system crashes" could actually be a hang instead. Are you thinking the Compaq hardware might actually halt in a "bus hang" situation when trying to access XMS memory with A20 off? I would have thought the process would instead just access zero-wrapped memory, but I guess that's what an 8086 would do, not a 286/386 with a larger address bus.

Mellvik:
> > and hangs when accessing anything above 1M, indicating that A20 is off.
> 
> I wasn't aware our previous "system crashes" could actually be a hang instead. Are you thinking the Compaq hardware might actually halt in a "bus hang" situation when trying to access XMS memory with A20 off? I would have thought the process would instead just access zero-wrapped memory, but I guess that's what an 8086 would do, not a 286/386 with a larger address bus.

Some more meat on this bone:
The way it looks after a short dive into the matter today is that a20 is fine. The point at which it hangs is the fist instruction using 32bit registers, iow the `addr32 rep movsw`.  Which leads me to suspect that the ASIC which handles memory mgmt is blocking 32 bit memory accesses. Finding out how to get around that may be hard, maybe possible.

ghaerr:
> The point at which it hangs is the fist instruction using 32bit registers, iow the addr32 rep movsw.

You mean in the linear32_fmemcpyw function?? That function should never be called int XMS INT 15 mode, it is only used for  copying memory in unreal mode. Perhaps the issue is our XMS code is calling this function somehow when it should not be?!

Mellvik:
The purpose of this exercise is to investigate why we haven't been able to get the compaq 386 to run in unreal mode. Iow we're running unreal mode.

ghaerr:
> The purpose of this exercise is to investigate why we haven't been able to get the compaq 386 to run in unreal mode.

I see. I thought we already assumed the Compaq BIOS was running in protected mode, i.e. using its own GDT. Our unreal mode only uses a temporary GDT to expand the shadow limit registers to allow addr32 operations with an extended offset. If the Compaq BIOS is in fact running in protected mode, I would guess that any or all entry/exits would have to switch to and back from protected mode, in order to keep the "emulation" of normal operation compatible when in real mode. Those switches would necessarily have to reload a GDT, which could cause our linear32_fmemcpyw to scribble anywheres via segment wrapping. However we only reset the limit registers, not the base registers (which one would think would have to be zero for proper real mode operation). Agreed something is fishy here.

Let me drag back out the [Compaq 386 BIOS disassembly](https://www.pcjs.org/machines/pcx86/compaq/deskpro386/rom/1988-01-28/1988-01-28.asm) to see whether anything can be learned from that regarding protected mode and selector activity within the BIOS.

ghaerr:
@Mellvik,

The Compaq 386 BIOS disassembly is a long read, but I think I'm getting somewhere with my first deeper dive. 

Two big issues: it seems no, the BIOS is *not* running in protected mode, and does not switch back and forth. (That would be quite slow, come to think about it). Instead, it appears that an internal GDT is used in order to access extended memory. But that's not unusual - actually it's the only way to access it, and all BIOS INT 15 handlers have to set up a GDT.

The Compaq difference is that apparently its extended memory doesn't start at 1M and go up, it starts at 16Mb and goes down. At boot, the Compaq BIOS copies its ROM BIOS down to F000. 

So the first big issue is that Compaq memory at 1M may not actually exist - it's up at 16M growing down. Does your Compaq have 16MB RAM? The BIOS INT 15 calls will remap or possibly bank switch the extended memory request to the actual memory location, so our unreal mode won't access it. I now remember playing with this, which is where this define came from in my older xms.c:
```
//#define XMS_START_ADDR  0x00FA0000L   /* 15.6M (Compaq with only 1M ram) */
```
So it does not seem that my previous guess was correct at all - there is no issue with multiple GDT's nor the ability to run in unreal mode. The issue is "where is the memory"? This might be able to made to work if we allocate XMS memory downwards, instead of upwards.

However, there is a another potential problem, check this comment out in the source:
```
;
;   When we arrive here, the A20 line has been disabled; on most systems, that would
;   mean that the ROM's GDT would only be accessible at the "low" ROM address (%0F0730),
;   not the "high" address (%FF0730).  But fortunately, A20 management on COMPAQ
;   DeskPros affects wrap-around only from the 1st to the 2nd megabyte; no other address
;   range is affected.
```
 Our assumptions about gating A20 opening up the entire address space may or may not be correct. We may have to run the special CEMM code (Compaq extended memory manager?) in order to "bank switch" it in. I am guessing that based on the following comment in the source:
```
;   Finally, the DeskPro 386 has a "RAM Relocation" feature that allows 128Kb of RAM at
;   %00FE0000 through %00FFFFFF to be mapped to %000E0000 through %000FFFFF, effectively
;   replacing the ROM in the first megabyte with write-protected RAM; the top 64Kb of that
;   RAM must first be initialized with the 64Kb at %000F0000 prior to remapping.  It's also
;   possible to copy external ROMs from %000C0000 through %000EFFFF into the bottom 64Kb of
;   that RAM, but this is only done for ROMs known to contain relocatable code; eg, a COMPAQ
;   Video Graphics Controller (VGC) Board.
;
;   Every DeskPro 386 system must have a MINIMUM of 1Mb of RAM, of which either 256Kb,
;   512Kb, or 640Kb can be physically mapped as conventional memory (at the bottom of the
;   first megabyte), with the remainder (either 768Kb, 512Kb, or 384Kb) physically mapped
;   to the top of the 16th megabyte (ending at address %00FFFFFF), the last 128Kb of which
;   is used by the "RAM Relocation" feature.  The remaining memory immediately below that
;   128Kb (ie, below %00FE0000) can only be accessed by special system software, such as CEMM.
;
;   COMPAQ refers to that remaining memory as "Compaq Built-in Memory".
```

Perhaps recompiling TLVC using the 0x00FA0000L XMS_START_ADDR and very few XMS buffers will allow the system to work using the upper memory bank, before we worry about whether bank switching is in effect or not. I don't have any other information on CEMM at the moment.

Mellvik:
Thank you @ghaerr,
This is good stuff. I need to look at that in the morning. I tried FA000 and 20000 earlier today w/o success. We also need to keep in mind that since hma works, there is memory at 10000.

ghaerr:
> I tried FA000 and 20000 earlier today w/o success.

Perhaps try F9000, with only 64 XMS buffers?

Are you finding the crash immediately after enable_real_mode, or xms_init actually returning and then hanging/crashing in linear32_fmemcpyw at a later point?

> We also need to keep in mind that since hma works, there is memory at 10000.

Great point! Perhaps uncomment the printk in xms_alloc to be sure of the allocation, and try just 32 XMS buffers, starting at 10000 with no kernel HMA.

If all these fail, I would guess the issue is that even though the BIOS isn't switching GDTs every entry, there's still something causing our unreal mode shadow registers to be reset. If they're being reset to FFFF (instead of FFFFFFFF), the 286 and 386+ processors, unlike the 8086, [throw a GP fault](https://stackoverflow.com/questions/17786357/segment-size-in-x86-real-mode). This could be what is happening, if you're seeing the system hang on the addr32 rep movsw instruction.

I'll read the referenced GP fault article to see whether anything can be done to prove that's what's happening.

Mellvik:
For the record , the machine has 10mb total ram.
As mentioned before, it stops dead at the first attempt to use 32bit registers, quite possibly a protection fault - as there may not be any memory where we’re looking. Come to think of it, hma working doesn’t really tell much as we’re in real mode. What may be the case is that the right gdt mapping is something completely off the wall compared to our assumptions ? Is there a way to read the ‘active’ gdt?

ghaerr:
> hma working doesn’t really tell much as we’re in real mode. 

Well, we're always in real mode except for possibly during INT 15, even when testing the effects of a previous protected mode entry/exit to set the segment shadow registers, otherwise known as unreal mode.

I believe HMA working tells us that the memory at 1M Is present. The verify_a20 routine explicitly writes a word at FFFF:0010 and reads it back. My understanding is that the 286/386+ never wraps segments, which is exactly why the a20 gate was added, to give older programs the ability to simulate 8086 wrap behavior, which was present in some DOS programs.

> the machine has 10mb total ram.

Given that the docs supposedly tell us the real RAM is at 16MB going downwards, this means to me that the BIOS, CEMM code, or something has previously mapped one of the Compaq memory segments down from high memory to 1M. Perhaps this should be investigated by reading the Compaq BIOS and/or CEMM manuals.

> What may be the case is that the right gdt mapping is something completely off the wall compared to our assumptions ? Is there a way to read the ‘active’ gdt?

There is the SGDT instruction, which will store the physical address of the active GDT. Docs say that's a privileged instruction, so I don't think it'll work in real mode, but you might try it. If it works, it should store the address as the address of 'gdt_ptr' in unreal.S.

When you say "off the wall", I'm not sure if makes any difference, since for any real mode to function at all, the base address for each segment register have to be 0 and the limits would alway shave to be at least FFFF. So then the question is - what happens when the CPU executes an addr32 memory access? I believe if the PA is <= 1M, it references that lower memory and works. If >= 1M, I believe there is no GP fault.

The reason I believe there is no GP fault is that real mode doesn't have a GP fault, so there is not a defined interrupt vector to call. The protected mode GP fault vector is 0Dh, which which overlaps IRQ 5 in real mode. That is because IRQ 0 is mapped by the PIC to the INT 8 vector in real mode (reference [IRQ 1 keyboard producing EXC 9 in protected mode due to improper PIC vector reprogramming](https://wiki.osdev.org/I_Can%27t_Get_Interrupts_Working#I'm_receiving_EXC9_instead_of_IRQ1_when_striking_a_key?!). Effectively this means that a GP fault would call the HD disk interrupt vector. Pretty sure this is not happening.

Since I'm now fairly confident we're not getting an interrupt or fault in real mode as a result of accessing XMS, I'm back to your first thought that the system is actually halting, possibly in a bus hang state, due to the Compaq memory remapping hardware. Do we need to look into the schematics or tech ref on that, perhaps?

I conclude with the following questions, if you're interested:

How do you know the system is actually hanging or halting on the addr32 rep movsw instruction? Are you using a printk or something after that, and assuming the rep mows doesn't complete? Does removing the "REP" and executing a single movsw stop the hang? It's just a buffer copy, after all, we would see the system continue and crash while trying to start a program. Where exactly is the kernel hanging, after the banner or before?

Given that HMA works and you're saying the system continues after enable_unreal_mode right up to the first linear32_fmemcpyw, it seems to me perhaps unreal mode could work, but perhaps for only the first 128K of memory at 1M.
 
Can you try running XMS with just 8 XMS buffers enabled, with or without HMA? That would eliminate the possibly of system crashes due to XMS buffers outside the first 128K bytes causing this problem. If that's the case, then we've solved the problem - the Compaq XMS memory needs to be mapped/unmapped using their CEMM tool instead of assuming its all mapped contiguously.

ghaerr:
@Mellvik,

I found the reference we need for Compaq systems, assuming that much of what's in it is also applicable to the Compaq 386 you're testing. In particular, the way the memory is laid out is discussed, the special memory remapping registers are documented, the cache controller is discussed which can problems when reading non-existent memory, and finally everything you'd ever want to know about the BIOS is documented, including a big list of 80386 processor malfunctions that should be worked around. 

I'm talking about the [COMPAQ DeskPro 386/25 Technical Reference Guide (Volume I, Aug 1988)](https://www.pcjs.org/machines/pcx86/compaq/deskpro386/). The link to the Vol 1 PDF is at the bottom of the page.

Some quick highlights:
- Page 2-15 Memory cache description
- Page 3-5 640k base memory description. With this normally used switch setting, the upper 384K of the lower 1M memory is mapped to the highest addresses at just below 16M. 
- Page 3-10 RAM relocation register: controls whether the uppermost 128K RAM is also addressable at E000:0000, as well as whether it is write-protected. The Compaq BIOS copies the ROM BIOS into RAM and then its shared between both these locations in normal operation.
The expansion memory doesn't start at 16M going downwards, it starts at 1M going upwards, except for the base 1M, which is described above.

- Page 4-9 warns about reading non-existent memory, but doesn't mention any hang issues, just cache issues.
- Page 4-10 is the long list of 80386 programming anomalies. In particular, Item 7 "Wrong register size in string instructions mixed 16/32-bit addressing" shows a special 386 bug workaround that I'd already added to linear32_fmemcpyw. However, looking at my routine in unreal.S, I've also commented out CLI/STI pairs with the comment that they need to be added should 32-bit extended registers be used in an interrupt routine. These may have been tested when I wrote the code and tested on Compaq, I can't recall. But since the Compaq BIOS may use 32-bit registers, these won't be saved across interrupt invocations unless CLI/STI used, so all the CLI/STI pairs should be uncommented for further testing.
- Page 4-23 Item 14 "Reload of the GDT register when entering protected mode" is potentially problematic - it's basically saying the BIOS has to reload GDT if it enters protected mode, which will surely invalidate our unreal mode shadow registers.
- Page 3-10 also says that one should not fiddle with the RAM memory remapping register directly, but always use INT 15 block move. It sounds likely that INT 15 may be used internally for LIM and/or EMS management.

In conclusion, there's lots of potential pitfalls for understanding exactly why unreal mode isn't working on Compaq, especially because of the extensive memory (re)mapping and cache features. This is exacerbated by early problems with the 80386 chips used in that era of machines as well.

We may have to perform more workarounds (e.g. reloading our local GDT on interrupt returns or before our memory copies) in order to get unreal mode working. We could reload our GDT before every linear32_fmemcpyw call, that would be pretty interesting. To do that, just 'call enable_unreal_mode` in the prologue of each linear32_ function in unreal.S.

Overall, the Compaq documentation is pretty amazing, and has given some new ideas to try, and seems to set straight our confusion about the XMS memory layout.

ghaerr:
@Mellvik, uncommenting the CLI/STI pairs was a little more complicated than I thought, since the original code tried to save the interrupt state using PUSHF and restore w/POPF. However, this doesn't work because the function arguments are then at the wrong offset. So I rewrote it to work on QEMU by using just CLI/STI and also calling enable_unreal_mode before each function. This diff would be interesting to try on Compaq, as it essentially enables unreal mode before each access using any addr32 instruction prefixes and also disables interrupts. It would be pretty cool if it worked!
```
diff --git a/elks/arch/i86/lib/unreal.S b/elks/arch/i86/lib/unreal.S
index 6a3fc91d..e621afd9 100644
--- a/elks/arch/i86/lib/unreal.S
+++ b/elks/arch/i86/lib/unreal.S
@@ -110,8 +110,7 @@ linear32_peekw:
    call   setgpf
    mov    %si,%dx

-   //pushf               // save interrupt status
-   //cli                 // uncomment if extended registers used in interrupt routines
+   cli                 // uncomment if extended registers used in interrupt routines
    xorl   %esi,%esi
    mov    %sp,%si

@@ -125,7 +124,7 @@ linear32_peekw:

    cld
    addr32 lodsw          // AX = [DS:ESI++]
-   //popf                // restore interrupt status
+   sti                // restore interrupt status

    mov    %dx,%si
    mov    %ss,%bx
@@ -140,12 +139,12 @@ linear32_peekw:
 #          Trashes EBX, ECX, ESI and EDI without saving.
 #
 linear32_fmemcpyw:
+   call   enable_unreal_mode
    push   %es
    mov    %si,%ax
    mov    %di,%dx

-   //pushf               // save interrupt status
-   //cli                 // uncomment if extended registers used in interrupt routines
+   cli                 // uncomment if extended registers used in interrupt routines
    xorl   %esi,%esi
    mov    %sp,%si

@@ -170,7 +169,7 @@ linear32_fmemcpyw:
    addr32 rep movsw      // word [ES:EDI++] <- [DS:ESI++], ECX times
    addr32 nop            // 80386 B1 step chip bug on address size mixing

-   //popf                // restore interrupt status
+   sti                // restore interrupt status

    mov    %ax,%si
    mov    %dx,%di
@@ -186,12 +185,12 @@ linear32_fmemcpyw:
 #          Trashes EBX, ECX, ESI and EDI without saving.
 #
 linear32_fmemcpyb:
+   call   enable_unreal_mode
    push   %es
    mov    %si,%ax
    mov    %di,%dx

-   //pushf               // save interrupt status
-   //cli                 // uncomment if extended registers used in interrupt routines
+   cli                 // uncomment if extended registers used in interrupt routines
    xorl   %esi,%esi
    mov    %sp,%si

@@ -221,7 +220,7 @@ linear32_fmemcpyb:
    addr32 rep movsb      // byte [ES:EDI++] <- [DS:ESI++], ECX times
    addr32 nop            // 80386 B1 step chip bug on address size mixing

-   //popf                // restore interrupt status
+   sti                   // restore interrupt status

    mov    %ax,%si
    mov    %dx,%di
@@ -236,12 +235,12 @@ linear32_fmemcpyb:
 #          Trashes EAX, EBX, ECX, ESI and EDI without saving.
 #
 linear32_fmemset:
+   call   enable_unreal_mode
    push   %si
    push   %es
    mov    %di,%dx

-   //pushf               // save interrupt status
-   //cli                 // uncomment if extended registers used in interrupt routines
+   cli                 // uncomment if extended registers used in interrupt routines
    xorl   %esi,%esi
    mov    %sp,%si

@@ -270,7 +269,7 @@ linear32_fmemset:
    addr32 rep stosb      // byte [ES:EDI++] <- AL, ECX times
    addr32 nop            // 80386 B1 step chip bug on address size mixing

-   //popf                // restore interrupt status
+   sti                // restore interrupt status

    mov    %dx,%di
    mov    %ss,%ax
```

ghaerr:
Finally found an article describing [Compaq's CEMM (Compaq Expanded Memory Manager)](https://www.os2museum.com/wp/original-cemm-unearthed/). This describes how CEMM in fact used nothing other than 80386 V86 virtual 8086 emulation mode (!) in order to page-map (i.e. *not* copy) memory above 1M into the EMS or UMB area under 1M.

More technical details on how this worked is described in [this article about EMM386](https://www.os2museum.com/wp/the-importance-of-emm386/) which is an evolution of CEMM, eventually used everywhere for running any 8088 or 8086 software on a 386. And it used V86 mode to do it - CEMM/EMM386 was actually a protected mode OS loaded as a DOS driver. It could stay loaded and when inactive, the machine ran in real mode, but when active, took over and the entire machine, DOS, and BIOS were essentially running in V86 protected mode, but allowing 8086 DOS programs to access lots of memory that acted like bank-switching but was actually page-mapped for super high speeds.

A couple last tidbits from the Compaq Tech Ref: Compaq states outright that INT 15 enables then disables A20 afterwards. It also mentions that interrupts are disabled for the duration of the INT 15 call, which for us means that network or async disk or serial activity could suffer and drop packets or get data overruns, depending on how long the block move actually takes.

This is all very interesting because it tells me that the Compaq BIOS can definitely be made to run with at least V86 protected mode outside the BIOS, which gets me to think there is a way to get ELKS to natively access XMS memory as well. Of course, it is possible that CEMM and EMM386 used INT 15 for memory copies and/or enter/exit protected mode, but that seems somewhat unlikely, given the wide variety of BIOS variations from back then.

Mellvik:
That's a shipload of great stuff, @ghaerr , much appreciated. I'm offline for a few days, will dive back to this rather exciting 'deepdive' this weekend, possibly earlier.

Mellvik:
It is indeed fascinating to finally get some understanding of what goes on under the hood here. And like you I'm pretty certain we'll be able to get the compaq working in unreal mode.  It's going to take some time to work this out - I'm still devouring your posts and I'm still away from my physical systems (hospitalized actually). 

Using the resources you've found (thanks, really great, and I have all the time in the world to browse them) and the physical machine we'll get there and learn a lot in the process.

what I was thinking when I asked about reading the gdt was that the bios is using either unreal or protected mode both when mapping memory and doing the xms block transfers. This seems to confirmed by your findings. This also means - as you allude to - that gdt mappings have been set up by the bios and that we may be trashing that when setting unreal mode.

also, the 'off the wall' reference was based on my understanding that in real mode, the mapping registers are not used at all, which means that we won't notice if a gdt entry is bad until we actually execute a reg32 instruction. Also, I know that's exactly where the code hangs because I'd already removed the `rep` to find out. So there is just a single `adds32 movsw` and a return, and it hangs. Further, adding enable_unreal_mode inside each call to linear32_fmem.... doesn't change anything.

the way I understand the low level basics before diving further into the docs is that
- the remapping of the ram from 640k to 1m is physical, requiring A23 to be set for that part of the 1st meg. Using a register for this as you found to be the case, is smart, and simplifies the change when...
- ... flipping a main board dipswitch changes main memory from 640 to 512k. The 'freed' memory is immediately added to the block mapped high, now totalling 512k.
- While the upper 128k of this is referred to as system memory and holds a copy of the bios, the rest is available as xms, meaning that even a 1M system, which was not unusual in the day, had xms.
- The bios copy, while unavailable from real mode gets used when doing INT15 block transfers by (I thought) switching to protected mode, actually V86 mode which makes sense.
- The 2nd Meg of RAM, also on the main board, is physically addressed from 1M and up. That allows for HMA  access from real mode. Adding more, up to a max of 10M, follows on top of that. The two blocks (9M low, 256k high) are probably mapped into a consecutive block via the gdt for use as xms (and counts into what's reported from the XMS size bios call).

At the end of the day though, what's important is how to get access to this entire block of memory via unreal mode. And I speculate, still without having completely devoured your analysis and the new docs: if the necessary gdt registers are set already, then  enable_unreal_mode may be unnecessary, possibly detrimental. That would be my first test when getting back to real hardware. Like - enable unreal_mode, then do a int15 block transfer. If it fails, we know that enable_unreal_mode screwed up the gdt. And next - skip the enable_unreal call and attempt to do 'unreal' transfers at various addresses (with 10M available, 80000 would be a reasonable place to start).

Your excepts from the BIOS code about BIOS memory mapping and a20 switching applying only to the first meg above 1M are really interesting. The latter not immediately applicable to anything as I see it, but useful to know when things down the line don't make sense. The former is news - they actually map the RAM copy back down to the ROM memory space. I'm not sure that's the case for the portable, but there is nothing in the doc I have that says otherwise. In fact what is says may be construed that way. What particularly fascinates me with this is that if we find out how they do it, we can map more of that RAM back down to the UMB area and get more conventional memory that way, including extending base memory from 640 to 702k, maybe 734k  (which I'm doing now on an XT, but that's a different story - and very different hardware).

Anyway - more to come. It may take another couple of days before I'm back to the hardware. In the meanwhile, there is plenty time to study.

Mellvik:
@ghaerr, the BIOS source is indeed very interesting read, in a number of ways. The tech ref too, but I've spent most of my the time on the bios source this far. The reading sends me back to previous posts, and it would be useful to clarify a few things.

- The BIOS does indeed switch between real and protected modes a lot, so messing with the GDT has to have consequences, as mentioned in one of your posts. Our previous interaction about dumping and looking at the gdt may be a good idea, not only to find the first available entry, but to evaluate how it would affect (or be affected by) the previous entries. (More about that below)
- It also seems important to keep in mind that the GDT & co are completely ignored in real mode, so the HMA memory has to be at physical 1M. As noted in the previous post I think all add-on memory is physically located upwards from that.
- Great work on the cli/sti changes in the unreal code, I'll get that into my version before more testing.
- > Page 4-23 Item 14 "Reload of the GDT register when entering protected mode" is potentially problematic - it's basically saying the BIOS has to reload GDT if it enters protected mode, which will surely invalidate our unreal mode shadow registers.

Yes, the BIOS does just that - in INT15/0x89, which is being used internally. I'm confident there is a way to modify the gdt structure it uses to preserve our changes, otherwise having that call would make moderate sense at best. Actually the ?EMM sources you have been looking at may have some indications if we don't find them in ch4 of the tech notes.

More later.

ghaerr:
> The BIOS does indeed switch between real and protected modes a lot

For which calls/services does the BIOS switch, do you have a list? Of course INT 15 will be switching modes into and out of the BIOS call, the GDT to use for mapping and moving memory in protected mode is passed to INT 15 in ES:BX. I am wondering whether the BIOS switches to protected moe for any service calls other than those requiring memory to be mapped/moved.

> Like - enable unreal_mode, then do a int15 block transfer. If it fails, we know that enable_unreal_mode 

Yes, that will be interesting, but I think it will work: the block move GDT is passed to INT 15, which the BIOS then uses with the LGDT instruction just before turning on the PE bit in CR0 to enable the required address mapping for the block move. This test should be done without XMS turned on, and only using a single enable_unreal_mode before the INT 15. Then it fails, it would seem to indicate that just the code in enable_unreal_mode is the problem.

Looking ahead on that - I now see that we reset the FS and GS registers in protected mode, although they are restored. There could be an issue there with the BIOS, although I can't right now think of any.

> This also means - as you allude to - that gdt mappings have been set up by the bios and that we may be trashing that when setting unreal mode.

There's a very high likelihood that the segment limit and descriptor bits will be different, thus "trashed", since our kernel GDT and the BIOS GDT are undoubtably different. A new GDT has to be specified when entering either mode (see below). The issue of having to reload the GDT was also listed in the Tech Manual errata, so there'll be no case where a previous GDT is assumed (and perhaps we shouldn't either - which is why a call to enable_unreal_mode before each linear32_ call might be necessary). The question is when does the BIOS switch to protected mode other than INT 15.

"Unreal" mode and protected mode are enabled in exactly the same way - by executing LGDT with a new GDT and then setting PE in CR0. The difference is that when the PE bit is cleared, which re-enters real mode, the previous GDT limits are still in effect in "shadow" registers in the 386+ CPU. This is then called unreal mode, which is real mode using shadowed limits that are set to > 64K.

> It also seems important to keep in mind that the GDT & co are completely ignored in real mode,

Well, not quite - the previous GDT limits are stored in shadow registers and when in real mode they are still in effect when using the addr32 prefix - this is unreal mode. There may be different limits depending on how many selectors are specified in the GDT, and I'm not sure exactly which limit is used for the ES and DS registers (the only ones we care about) but I'm guessing they would be those according to the selector currently in use in each of them.

> I'm confident there is a way to modify the gdt structure it uses to preserve our changes, otherwise having that call would make moderate sense at best. 

I don't think there's much chance in knowing or learning about any GDT that might be passed both internally or externally to the BIOS helping us, since they change internally and we depend on our own. That's why I wanted to know the results of enable_unreal_mode in linear32_ (which you said didn't work), since that function loads our own GDT, sets PE, then unsets PE to return to real mode. IMO the real question is why is it hanging on our first addr32 mov instruction. Trying that with interrupts disabled may provide an answer if the BIOS is switching modes on us as a result of hardware interrupt.

>  so the HMA memory has to be at physical 1M. As noted in the previous post I think all add-on memory is physically located upwards from that.

Agreed. The tech manual has basically removed suspicions that there is real memory at 16MB; it is reflected from < 1MB back up to 16MB by dip switches as we've been both talking about.

> Actually the ?EMM sources you have been looking 

Unfortunately I have not found any sources to CEMM nor EMM386; that would be very interesting indeed. Especially since ELKS could possibly actually run in V86 mode underneath a protected mode microkernel doing page-mapping, something until now I'd never considered.

However, I don't think the BIOS is doing any memory mapping on the fly, as both CEMM and EM386 were DOS .SYS files which were ultimately controlled by application programs making VCPI or DPMI calls. But I hope we find out soon!

Mellvik:
> > The BIOS does indeed switch between real and protected modes a lot
> 
> For which calls/services does the BIOS switch, do you have a list? Of course INT 15 will be switching modes into and out of the BIOS call, the GDT to use for mapping and moving memory in protected mode is passed to INT 15 in ES:BX. I am wondering whether the BIOS switches to protected moe for any service calls other than those requiring memory to be mapped/moved.

I didn't go that far yet, the point here was to ascertain that that's what it's doing. I need somthing better than my iPad to do the full map, and I agree with you, that will be an interesting piece of the full picture.

> 
> > Like - enable unreal_mode, then do a int15 block transfer. If it fails, we know that enable_unreal_mode
> 
> Yes, that will be interesting, but I think it will work: the block move GDT is passed to INT 15, which the BIOS then uses with the LGDT instruction just before turning on the PE bit in CR0 to enable the required address mapping for the block move. This test should be done without XMS turned on, and only using a single enable_unreal_mode before the INT 15. Then it fails, it would seem to indicate that just the code in enable_unreal_mode is the problem.

yes, having read more of the sources it seems evident that the bios is doing an almost complete setup every time. I've mentioned extracting the gdt to have a look at it, which may still be interesting, but looking at the bios source shoud provide the same info (and possibly more understanding). In the code, where it prepares for an int15 call to enter protected mode, it does the entire setup, probably in several places. I'm heading back into that as soon as I get the chance.

> 
> > This also means - as you allude to - that gdt mappings have been set up by the bios and that we may be trashing that when setting unreal mode.
> 
> There's a very high likelihood that the segment limit and descriptor bits will be different, thus "trashed", since our kernel GDT and the BIOS GDT are undoubtably different. A new GDT has to be specified when entering either mode (see below). The issue of having to reload the GDT was also listed in the Tech Manual errata, so there'll be no case where a previous GDT is assumed (and perhaps we shouldn't either - which is why a call to enable_unreal_mode before each linear32_ call might be necessary). 

I agree, and needing an enable_… call ahead of each linear32_ call makes sense even though it didn't change anything when I tested. We'll soon figure out why.

> The question is when does the BIOS switch to protected mode other than INT 15.

I plan to map that out when I get to a mac.

> 
> > It also seems important to keep in mind that the GDT & co are completely ignored in real mode,
> 
> Well, not quite - the previous GDT limits are stored in shadow registers and when in real mode they are still in effect when using the addr32 prefix - this is unreal mode. 

Yes, I know. I wasn't including unreal in that statement. My point was that HMA access is real 'real', and uses physical addressing only.

> 
> > I'm confident there is a way to modify the gdt structure it uses to preserve our changes, otherwise having that call would make moderate sense at best.
> 
> I don't think there's much chance in knowing or learning about any GDT that might be passed both internally or externally to the BIOS helping us, since they change internally and we depend on our own. 

I agree, knowing what we know now, that idea doesn't make sense.

> That's why I wanted to know the results of enable_unreal_mode in linear32_ (which you said didn't work), since that function loads our own GDT, sets PE, then unsets PE to return to real mode. IMO the real question is why is it hanging on our first addr32 mov instruction. Trying that with interrupts disabled may provide an answer if the BIOS is switching modes on us as a result of hardware interrupt.

At least that will eliminate a possible source of error. Logically, processing time for the enable-unreal function should be too short to be hit by an interrupt every time, but there are still a number of unknowns here.

> 
> > Actually the ?EMM sources you have been looking
> 
> Unfortunately I have not found any sources to CEMM nor EMM386; that would be very interesting indeed. Especially since ELKS could possibly actually run in V86 mode underneath a protected mode microkernel doing page-mapping, something until now I'd never considered.

Sorry, I misread your previous post about that. You were referring to an article, not sources. Having browsed the bios source, which is obviously a disassembly with added comments and a lot of nice processing, it would surprising if noone has attacked either cemm or emm the same way. Then of course there are the drdos and freedos versions.

ghaerr:
> IMO the real question is why is it hanging on our first addr32 mov instruction.

> Also, I know that's exactly where the code hangs because I'd already removed the rep to find out. So there is just a single adds32 movsw and a return, and it hangs. 

What do you mean "and a return"? Are you saying that the hang is after the RET instruction already coded at the end of linear32_fmemcpyw, or hangs directly at the ADDR32 movsw? Or did you recode linear32_fmemcpy to just perform a addr32 movsw then pop the saved registers and execute a RET to a printk above? If so, I would make sure that routine doesn't crash even without the movsw.

Yes it would be nice to know exactly where it is hanging. This could also possibly be accomplished with inline ASM.

Another test might be to try accessing RAM < 1M using the very same linear32_fmemcpy and the add32 movsw for a few bytes. This would then not use any XMS memory and by all rights in real mode with any shadow limit should work.

There is one other possibility I hadn't considered, however unlikely: that the CPU is actually in [Virtual 8086 VM](https://wiki.osdev.org/Virtual_8086_Mode). This would be pretty unlikely since setting PE in CR0 is not allowed, but it is possible there's an exception being thrown on addr32 movsw that then crashes from no handler.

I suppose if you know for sure the hang is on that very instruction that the shadow limit registers are somehow being reset by the BIOS to invalid values for us and a bus error or other hardware hang is occurring when we try to access using a selector from our now-invalid GDT.

I'm looking forward to your deep dive :)

Mellvik:
> > IMO the real question is why is it hanging on our first addr32 mov instruction.
> 
> > Also, I know that's exactly where the code hangs because I'd already removed the rep to find out. So there is just a single adds32 movsw and a return, and it hangs.
> 
> What do you mean "and a return"? 

Don't take that literally.

> 
> Another test might be to try accessing RAM < 1M using the very same linear32_fmemcpy and the add32 movsw for a few bytes. This would then not use any XMS memory and by all rights in real mode with any shadow limit should work.

that's a good idea. My plan is to stay away the 100000-200000 area for now to make sure a20 gating is out of the picture. Adding < 100000 is good.

> 
> There is one other possibility I hadn't considered, however unlikely: that the CPU is actually in [Virtual 8086 VM](https://wiki.osdev.org/Virtual_8086_Mode). This would be pretty unlikely since setting PE in CR0 is not allowed, but it is possible there's an exception being thrown on addr32 movsw that then crashes from no handler.

Agreed, this seems unlikely since the bios (as far as I can tell) doesn't use that mode. 

> 
> I suppose if you know for sure the hang is on that very instruction that the shadow limit registers are somehow being reset by the BIOS to invalid values for us and a bus error or other hardware hang is occurring when we try to access using a selector from our now-invalid GDT.

I'm thinking it will be interesting to dump the gdt just before and just after the call to enable_unreal to check that. I keep asking myself why it works on the other machines (the SX and qemu, that is), and not this one, and if it isn't bios related, then there is something about memory mapping still not understood. Comparing gdt dumps from working and non-working systems may also be interesting. Also on the list is a close look at how the bios prepares the gdt before an int15 call to enter protected mode. The latter can be done without access to the hw, which is still the case.

> 
> I'm looking forward to your deep dive :)

An interesting journey indeed. When I came across the int-call to setup the reset-return vector from protected mode on the 286 (in the compaq tech ref), `loadall` started to wag its tail in the back of my head - again (not that `loadall` requires a reset, it doesnkt, it was just a trigger). Likewise for the potential use of V86 mode for TLVC/ELKS, as you too mentioned. That could be some real fun on a late night - if not necessarily all that useful, but educating.

Mellvik:
@ghaerr,
I finally found the reason why the Compaq portable doesn't seem to work in unreal mode. It has been quite a journey, educating as usual, and a lot of time wasted, but that's the nature of debugging, isn't it? I spent hours on the Deskpro BIOS, eventually determining that what I read couldn't possibly match what I'm seeing on the Portable, so I disassembled the Portable 386 BIOS too, and spent quite some time with it. Again, educating, and while it didn't solve the problem, it provided background to ask the right questions.

I finally have code (still to be cleaned) that make unreal mode work across all systems. The problem is this: As is, the `enable_unreal_mode()` routine does nothing. As it turns out, flipping the PE bit does enable, but not turn on protected mode. Until the prefetch queue has been emptied and most importantly, the CS changed, the processor stays in real mode. This is why we never - in the BIOS sources for example  - see a switch to/from protected mode without an ljmp immediately following it. The ljmp takes care of both. And the ljmp was missing from our enable_unreal_mode routine.

The question is: Why does this work in some cases, like on QEMU and my AMI386SX system? The answer seems to be lucky circumstances. QEMU is easy. I commented out the calls to enable_unreal_mode completely, and xms still works just fine. What's presumably happening is that the SEA BIOS is using a full, flat GDT entry for DS and ES, and leaves those in the segment descriptor cache. Since our memory transfer routines expect that simple mapping, everything is fine. And A20 is always on anyway on QEMU.
The explanation for the AMI386 system is likely similar, but on this system behaviour without calling enable_unreal was different, and before I figured out why, I got the new enable_unreal working, so I left it.

IOW, getting the long jumps into place, the Compaq 386 runs unreal like a charm. I have yet to demonstrate it with HMA kernel, but there is not reason it should not work.
 
Code coming on TLVC when I've cleaned it.

ghaerr:
> I finally found the reason why the Compaq portable doesn't seem to work in unreal mode. 

Wow! Absolutely extraordinary work debugging this issue @Mellvik!!! Very impressive!!!!!! Amazing!!!

Your above writeup sheds new, bright light on "unreal mode" in a way I kind of thought I understood but didn't fully: unreal mode isn't a "mode" at all, it's a misnomer - all 386+ CPUs always use the shadow segment limit registers values on add32r prefix instructions, and their values are the same as whatever they were previously set to from the last time the CPU was in protected mode. There is no "unreal mode" - it's the way the 386 always works in real mode.

The BIOS on any machine with > 1M memory will undoubtably have to enter protected mode at startup, just to run the memory tests. And unless the BIOS explicitly resets them to 64K before exiting protected mode, they could be any value, but seem to usually be set to large flat linear segments. This also means that real mode operation using addr32 prefixes is inherently unreliable if any BIOS call is made that changes to protected mode. The good news is that it seems the disk and console BIOS services don't.

> Until the prefetch queue has been emptied and most importantly, the CS changed, the processor stays in real mode. 

Very nicely said. I wish I had previously read that somewhere. I now remember articles discussing the need to execute an LJMP immediately after "entering" protected mode, but never understood what you have empirically found - that setting PE doesn't in fact immediately enter protected mode. I also believed, from reading certain articles, that there was such a thing as a separate "unreal mode". Thank you for this enlightenment!

> QEMU is easy. I commented out the calls to enable_unreal_mode completely, and xms still works just fine. 

Unbelievable. But now makes perfect sense.

> The explanation for the AMI386 system is likely similar,

I actually tested my original XMS implementation on real hardware. I'm sure it is as you're thinking - the AMI386 and any other machine w/XMS memory will have previously entered protected mode and most setup flat descriptors for DS and ES.

> The answer seems to be lucky circumstances. 

Incredible work @Mellvik. It's a wakeup call, and amazing that something can "work" on most all emulators as well as many real hardware BIOSes but not actually be correct at all.

> Code coming on TLVC when I've cleaned it.

I'm thinking 'enable_unreal_mode' should be renamed, since even with an LJMP, there isn't actually an unreal mode. Perhaps something like 'set_shadow_limits' or something like that would be a better description of what is actually required.

Finally - your findings now release the need to support both "unreal mode" and INT 15/1F dynamically, which was the subject of this PR. The INT 15/1F support can be compiled in, without an option, for PC-98 systems that use INT 1F, and otherwise left out. However, given this new information, it's possible that PC-98 can work using the shadow limit registers properly set. If so, the entire INT 15/1F support can be thrown out (or left in for future manual testing but outside user configuration).

The ability to dynamically switch between unreal mode and INT 15 handling costs 1.5k near .text, which is IMO too much to leave in if not absolutely needed.

Thank you!!!

Mellvik:
> 
> Your above writeup sheds new, bright light on "unreal mode" in a way I kind of thought I understood but didn't fully: unreal mode isn't a "mode" at all, it's a misnomer - all 386+ CPUs always use the shadow segment limit registers values on add32r prefix instructions, and their values are the same as whatever they were previously set to from the last time the CPU was in protected mode. There is no "unreal mode" - it's the way the 386 always works in real mode.

I haven't thought about it that way, but of course you're right. It's no 'mode', it's more like a setting. That said, I'm voting for keeping the name since it is well known and (not so well?) understood. There are likely many strong opinions about such a thing out there,  keeping the name eliminates noise ...

> 
> The BIOS on any machine with > 1M memory will undoubtably have to enter protected mode at startup, just to run the memory tests. And unless the BIOS explicitly resets them to 64K before exiting protected mode, they could be any value, but seem to usually be set to large flat linear segments. This also means that real mode operation using addr32 prefixes is inherently unreliable if any BIOS call is made that changes to protected mode. 

Right, having spent far too much time reading BIOSes this past week, and noting that this will vary between vendors, protected mode is used at boot time for memory test and for rearranging ROM content (the COMPAQs may be extreme in that with their  juggling of both ROM and RAM). Apart from that, what I've seen is the block mem move INT15 and the enter protected mode INT15. 

> The good news is that it seems the disk and console BIOS services don't.

That's right. Actually I think it's pretty safe to use unreal_mode for whatever it may be good for now that we have a complete understanding. If really paranoid, 'enable_unreal_mode' (!) could be called for every transaction, or much simpler, look at the return from `sgdt` to check that the address is 'ours', not 'their'. 

> 
> > Until the prefetch queue has been emptied and most importantly, the CS changed, the processor stays in real mode.
> 
> Very nicely said. I wish I had previously read that somewhere. I now remember articles discussing the need to execute an LJMP immediately after "entering" protected mode, but never understood what you have empirically found - that setting PE doesn't in fact immediately enter protected mode. I also believed, from reading certain articles, that there was such a thing as a separate "unreal mode". Thank you for this enlightenment!

The same goes for leaving protected mode. Until the CS reg has been changed, we're still in protected mode. Now, with relocatable code, creating the `ljmp` for the return to real mode is tricky. It's either `ljmp immediate:immedaite` as in `ljmp 0x1d0,0x1234`, the first being the new CS, the latter is the new IP. But we cannot hardcode the CS value since it depends (say, HMA). 
The second form is `ljmp *memloc`, where memloc holds the two word-values, easy except we've changed the DS, so we cannot get to it. I ended up with (ugh!) self-modifying code, the alternative - unless I'm missing something - is to have a fixed location in memory for it.

> 
> Finally - your findings now release the need to support both "unreal mode" and INT 15/1F dynamically, which was the subject of this PR. The INT 15/1F support can be compiled in, without an option, for PC-98 systems that use INT 1F, and otherwise left out. However, given this new information, it's possible that PC-98 can work using the shadow limit registers properly set. If so, the entire INT 15/1F support can be thrown out (or left in for future manual testing but outside user configuration).
> 
> The ability to dynamically switch between unreal mode and INT 15 handling costs 1.5k near .text, which is IMO too much to leave in if not absolutely needed.

I agree. The way it looks from my perspective is that INT15 XMS access must be a CONFIG option, useful for 286 systems only. And if enabled, the unreal XMS (and possibly HMA, but that's very small) code can be left out. Also, in order to provide the smallest possible kernels for XT class systems, it must be possible to CONFIG out most/all of the HMA and XMS stuff. Anyway - a lot simpler than today.

> 
> Thank you!!!

And thanks for the discussion.

ghaerr:
> creating the ljmp for the return to real mode is tricky.

Although DS and ES segments are changed, I think SS is still valid. Try something like the following, where the existing CS and a known IP are pushed on the stack, then LRET to set them:
```
        push %cs
        mov $1f,%ax
        push %ax
        lret
1:
```

Mellvik:
> > creating the ljmp for the return to real mode is tricky.
> 
> Although DS and ES segments are changed, I think SS is still valid. Try something like the following, where the existing CS and a known IP are pushed on the stack, then LRET to set them:
> 
> ```
>         push %cs
>         mov $1f,%ax
>         push %ax
>         lret
> 1:
> ```

No sh..., I tested that and couldn't get it to work. Now it works. I must have forgotten the '$'. Thanks- some more testing and a PR ( accidentally pushed the code into the currently open PR, it's being reverted and turned into a separate push).

PR #2276, opened 2025-04-01
Improve paint line drawing

Vutshi:
Implement proper Bresenham's algorithm. See results in https://github.com/ghaerr/elks/issues/2269#issuecomment-2768918792


ghaerr:
I just noticed previous floating point \#defines and \#include code on lines 7-18 are still included, but never used below. 

Can you delete those, all that is completely unneeded now that floating point is no longer used.

Vutshi:
Done.

PR #2277, opened 2025-04-05
[cmds] Speed up Paint code drawing on slow 8088 systems

ghaerr:
Speeds up drawing in Paint program considerably, discussed in https://github.com/ghaerr/elks/issues/2269#issuecomment-2779789406.

This turned out to be an interesting exercise, and an ia16-elf-gcc compiler bug was found in the process of trying to get it to not translate (x << 6) + (x << 4) to (x * 80) and emit a MUL instruction. On the 8088, MUL instructions are very slow and using left shifts will be much faster.

What was found with regards to ia16-elf-gcc is:
- The default code generation option -Os (optimize for size) is not well suited for 8088 code generation. Specifically, it generates small code with no regard to speed.
- Replacing the global optimization option for paint to -O3 (or -O2 or -O1) caused the compiler to emit code which crashed paint when drawing. This is still being investigated.
- Trying to use the GCC \_\_attribute\_\_((0))) to turn off optimization for a single function generated very sloppy code which was deemed too slow.
- When optimizing in the default -Os (small size) mode, GCC actually replaces (x << 6) + (x << 4) with (x * 80) and generates a MUL instruction.
- In the same -Os default optimization, writing (x / 8) actually generates a DIV instruction (!!!). Coding (x >> 3) generates a right shift, which was nice to see.

After playing way too long trying to get GCC to work without potentially putting each graphics function in a separate file, it was finally decided that translating the entire C86 ASM language vga-4pp.s fast draw routines from AS86 format to GCC AS format was the way to go, and it now works very well.

As a result of conversations with @Vutshi and @dbalsom in #2269, the following changes were made to greatly increase the speed of paint on 8088. This has been tested with QEMU slowing it down considerably by adding `-singlestep -icount 8,align` to the QEMU command line. Soon I hope to get [MartyPC](https://github.com/dbalsom/martypc) running on my macOS laptop for cycle-accurate 8088 emulation.
- Rewrite the VGA 4-planar C86 ASM routines in vga-4bpp.s to GCC assembly in vga-ia16.S.
- Rewrite the cursor show/hide routines so that only the masked cursor bits are saved and restored. This sped up cursor drawing by about 50%.
- Remove MUL and DIV instructions from the main loop event_wait_timeout routine.
- Replace previous y * 80 code with (y << 6) + (y << 4) in C and the new vga-ia16.S code. vga-4bpp.s not done yet (for C86).
- Slightly rewrote and got working @Vutshi's C drawhline routine. This is where I later found the GCC bug when compiling in -O3 mode. This routine is commented out now since it has been replaced with fast ASM code in vga-ia16.S.
- Add a mouse acceleration filter which allows the mouse to speed up when moved rapidly. Seems to work well but needs to be more throughly tested on different speed systems.
- Removed x >= 0 and y >= 0 clipping checks in render.c.
- Rewrote R_ClearCanvas and R_DrawPalette to use drawhline for ia16.

### DIFF
--- a/elkscmd/gui/Makefile
+++ b/elkscmd/gui/Makefile
@@ -7,7 +7,7 @@ ALL += owc
 endif
 
 ifdef C86
-ALL += c86
+#ALL += c86
 endif
 
 all: $(ALL)
--- a/elkscmd/gui/Makefile.gcc
+++ b/elkscmd/gui/Makefile.gcc
@@ -26,11 +26,15 @@ BINDIR = .
 LOCALFLAGS =
 PROG = $(BINDIR)/paint
 SRCS = app.c gui.c input.c render.c event.c mouse.c graphics.c drawbmp.c cursor.c
+OBJS += vga-ia16.oaj
 
 all: $(PROG)
 
 $(PROG): $(OBJS)
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
+
+vga-ia16.oaj: vga-ia16.o
+	mv vga-ia16.o vga-ia16.oaj
 
 clean:
 	rm -f $(PROG) *.oaj
--- a/elkscmd/gui/app.c
+++ b/elkscmd/gui/app.c
@@ -7,8 +7,6 @@
 #include "graphics.h"
 #include "mouse.h"
 
-#define LIBPATH     "/lib/"     // path to images
-
 // --------------------------------------------
 // Definition of Globals
 // --------------------------------------------
--- a/elkscmd/gui/app.h
+++ b/elkscmd/gui/app.h
@@ -7,6 +7,8 @@
 
 #define PALETTE_WIDTH 148
 
+#define LIBPATH     "/lib/"     // path to images
+
 // Boolean Data Type
 typedef enum boolean_e
 {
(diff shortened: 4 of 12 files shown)


Vutshi:
@ghaerr this is very fast on 8088!
Is there a reason to use ifdefs in `render.c`? It seems to be more convenient to define all compiler dependent functions in `graphics.c` and expose only graphics primitives `drawpixel`, `drawhline`, `drawvline`, and `fill_rect`. This would simplify the paint rendering code.

ghaerr:
> this is very fast on 8088!

That is very good to hear, thank you for your testing! I think the combo of moving to ASM draw routines and removing the cursor hide/show overhead made the biggest differences. The jury is still out on the 8-instruction SHL/ADD sequence to replace MUL, but that could be helping too.

> Is there a reason to use ifdefs in render.c?

It got to be late last night and I didn't have time for cleanup. I plan on doing just what you're suggesting and will have a follow-on PR today to clean things up. There was a lot of experimentation going on finding out where the bottlenecks were.

Please let me know what other slowdowns you are experiencing, as well as what you think about the new mouse acceleration feature.

toncho11:
Sounds great for all the 8086/8088 I have :) I am a collector :) maybe you got that already.

I wonder if the bug(s) is affecting Micro Windows. No, Micro Windows is OWC currently?

ghaerr:
> I wonder if the bug(s) is affecting Micro Windows. No, Micro Windows is OWC currently?

Yes, many of these speedups apply to Microwindows as well, I plan on updating it also.

Nano-X is currently being built using ia16-elf-gcc in small model. There has not been a port to OWC as of yet, as it hasn't been needed due to not requiring large model.

PR #2193, opened 2025-01-20
Added support for circle for CGA / IBM PC in Basic

toncho11:
Also a few comments and a renamed file.
Linked to: https://github.com/ghaerr/elks/issues/2188#issuecomment-2601649569

PR #2278, opened 2025-04-05
[cmds] Add drawvhline and cleanup paint code, get C86 version working again

ghaerr:
Discussed in https://github.com/ghaerr/elks/pull/2277#issuecomment-2780822278.

Cleans up graphics.c and render.c code.
Adds drawvline graphics primitive (in ASM for IA16 and C86).
Adds drawhline in C for OpenWatcom.
Renames vga-4bpp.s to vga-c86.s.

Replaces some expressions too complicated for C86 (at the moment) in render.c to get C86 version running again.

Paint now compiles and runs quickly with the ia16, c86 and Watcom compilers!


Vutshi:
@ghaerr, with so many compilers available, am I missing out on performance by only using ia16?

ghaerr:
> with so many compilers available, am I missing out on performance by only using ia16?

No. The different compiler support is for reasons other than performance, that are otherwise not available. 

Ia16 optimization is extremely good, with the possible exceptions that I just found using -Os, which optimizes for size rather than speed. But this likely only matters in certain speed-critical loops, the last of which may be R_DrawCircle - I haven't studied the output of  that function yet, and it seems a bit slow, although this is only with large circles and likely from drawpixel overhead.

OpenWatcom uses registers rather than the stack for function calls, which can help, but other optimizations aren't as good as ia16. C86 only tries to optimize on a per-statement/expression basis, and in general doesn't optimize much.

The purpose of supporting our three compilers are:
- Ia16 for the kernel and almost all application programs, including graphics. (Nano-X uses ia16 also)
- OpenWatcom for large data model, unavailable with the other compilers. This allows arbitrarily large code and data size, and uses the OS/2 executable format loaded by the kernel.
- C86 for native compilation, this compiler supports compiling and linking on ELKS itself, as well as cross-compilation.

Keeping the graphics routines and now paint portable across the compilers helps me to better understand our new C86 compiler, as well move forward with the plan of adding basic graphics routines to our C library, which is used as the library for all three compilers, which is pretty cool!

Vutshi:
> R_DrawCircle - I haven't studied the output of that function yet

I am replacing this function with a much better Bresenham's circle

ghaerr:
> I am replacing this function with a much better Bresenham's circle

What may be needed for real speed improvement would be filling the circle using calls to drawhline, rather than drawing pixel-by-pixel. Either that, or using a in-memory circle pixmap and a fast blit for drawing (this is what the normal Nano-X does, but blit gets complicated with larger images). Nano-X has a fast blit written in C, but it's not converted to ASM yet. I don't really know whether its worth the effort to speed circle drawing up this way or not.

I'm done modifying internal Paint routines for now (until your next problem), so feel free to add enhancements. If I see any improvements that can be made to graphics.c or the mouse routines, I'll mention it before putting up a PR.

Thank you!

Vutshi:
> What may be needed for real speed improvement would be filling the circle using calls to drawhline

Yes, it is using drawhlines and it is free of multiplications. However, the main point is that Bresenham’s circle is an optimal rasterization of an ideal circle. What we have now is an approximation.

toncho11:
I added a circle draw in #2193 that uses the midpoint circle algorithm. It might easily be updated for fill as well I think. This algorithm is all fixed point.

Vutshi:
These are two names of essentially one circle drawing algorithm.

Vutshi:
@ghaerr I suspect that `vga_drawhline` is doing something wrong.

Circles made with `vga_drawhline`:
vga_hline

Circles made with simple `drawhline`:
```
void drawhline(int x1, int x2, int y, int c)
{
    while (x1 <= x2)
        drawpixel(x1++, y, c);
}
```
simple_hline

ghaerr:
> I suspect that vga_drawhline is doing something wrong.

Oh geez. Not sure what the problem is, and the current code is only using vga_drawhline in very limited circumstances: fillrect.

Why don't you go ahead and create your circle drawing enhancement/PR using a call to a locally-named C drawhline (using draw pixel), so that it fully works. I'll then commit it and use that to debug vga_drawhline. We have a C86 and IA16 version of vga_drawhline which I will use first to see if it was a translation error on my part, or something deeper.

This should work for you since we're not using vga_drawhline anywhere else (yet).

ghaerr:
> Circles made with vga_drawhline:

It almost looks as though it is drawing a square, poking out the left side, as though the left side of the "circle" is not getting masked off, but getting drawn anyways. The right side of the circle is always a circle and smooth, indicating that the line segment is being masked off on the right hand side.

The way the VGA drawhline works is that it cuts the horizontal line segment into three parts, corresponding to three possible bytes of RAM, any of which could be not used: the left side, of which the high bits are supposed to be masked off, the middle, of which all bits are written, and the right side, of which the lower bits are supposed to be masked off.

It might be interesting to try this with various circle sizes, to see whether the idea of left bits not being masked is true. If the circle is <= 8 bits, then a single byte is used for the masking. Where the error occurs tells us where in drawhline the problem is.

Vutshi:
I made a PR. Good night)

PR #2279, opened 2025-04-05
Paint update 2

Vutshi:
New circle drawing function.
Some cleanups to make sure that we do not draw on the side panel.


ghaerr:
Thanks @Vutshi I fixed my vga_drawhline bug in #2280.

PR #2280, opened 2025-04-06
[cmds] Fix vga_drawhline in Paint

ghaerr:
Fixes error in IA16 vga_drawhline found in https://github.com/ghaerr/elks/pull/2278#issuecomment-2781111983.

The problem was a single ASM line with reversed arguments (you have to thank Intel vs AT&T for that)!!

@Vutshi, I had to remove the mouse clipping (when drawing) you added into event.c, as when selecting various colors one after another in the control panel, after a bit of doing just that the entire screen would get filled with the color selected. It seems that this is the same problem you were trying to fix when moving the cursor off the canvas when drawing, when using the new R_DrawDisk function.

I think this is a new bug that has to do with R_DrawDisk, as I'd never seen it before, but not sure.

In any case, we don't want to/can't keep the mouse clipping code in event.c anyways, as it contains app.h and application globals (drawing, altduawing). The event.c code is meant to be a library routine, so I'd like to ask that you clip the mouse coordinates and fix the problem in input.c during mouse down processing, instead of in the shared app main event code. I haven't figured out exactly why the problem is, but just commented it out for the time being. With it commented out, the buttons never fail but you can get in trouble when mouse down drawing going off the canvas.

On another note, I am closely comparing the circle code and there are definitely differences. TBH, I'm not sure the new code looks better - the 4th pixel size in particular was very round in the before version, and now seems to have points on it and kind of looks like a diamond (see screenshots below, you may want download them and quickly switch between them with mac Preview "spacebar" in order to really check them out). I can see that the single pixel circle is now correct whereas in the old version it was a bit larger. Nonetheless, the new code draws a circle that is a little bit larger (you need to quickly switch between them to see it).

The new circle looks are fine with me, I just wanted to bring this up, in case you handn't already noticed it. I did notice that the new draw code is much faster, undoubtedly from drawing using drawhline, is that the main benefit of this rewrite?

I had already written the above when I realized that there must be a strict way that circles have to be rasterized if they are to easily use drawhline to draw them, rather than the very slow original drawpixel method.

Before
before

After
after

Thank you for your enhancements!

### DIFF
--- a/elkscmd/gui/event.c
+++ b/elkscmd/gui/event.c
@@ -115,13 +115,13 @@ out:
                     posy += y;
                     if (posx < 0) posx = 0;
                     if (posy < 0) posy = 0;
-                    if ((drawing || altdrawing)){
-                        if (posx >= SCREEN_WIDTH) posx = SCREEN_WIDTH - 1;
-                        if (posy >= SCREEN_HEIGHT) posy = SCREEN_HEIGHT - 1;
-                    } else {
+                    //if ((drawing || altdrawing)){
+                        //if (posx >= SCREEN_WIDTH) posx = SCREEN_WIDTH - 1;
+                        //if (posy >= SCREEN_HEIGHT) posy = SCREEN_HEIGHT - 1;
+                    //} else {
                         if (posx >= SCREENWIDTH) posx = SCREENWIDTH - 1;
                         if (posy >= SCREENHEIGHT) posy = SCREENHEIGHT - 1;
-                    }
+                    //}
                     event->x = posx;
                     event->y = posy;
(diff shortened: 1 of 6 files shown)


Vutshi:
> TBH, I'm not sure the new code looks better - the 4th pixel size in particular was very round in the before version, and now seems to have points on it and kind of looks like a diamond

I didn’t say the new circles look nice, I said they are the raster circles closest to the ideal one :)

There are no better circles **centered** on a pixel. I’ll make an illustration later. The previous circles were not, like e.g. 2x2 block of pixels. 

UPDATE: Diamonds are a small circle’s best friend
Circles_r0_4

PR #2281, opened 2025-04-06
[cmds] Add palette cache to paint bmp image drawing for speed

ghaerr:
While playing around trying to find some fast planar 4 VGA blit code, I ran across SVGALib which had what seemed to be a complicated but possibly fast 16-color buffer to VGA planar vga_drawscanline blit routine. 

I ended up porting the routine and getting it working on ia16, and during the process found that most of the time being spent displaying our BMP images is actually in converting the BMP palette index into a standard VGA color. This is very time-consuming because the BMP image bytes are just internal palette indices, and then the palette entry itself needs to be compared with our Paint palette, and uses a least-squared distance algorithm to determine the VGA color, which is based off a standard VGA palette.

This PR adds a palette-cache to the BMP image loader which ends up speeding up image display by at least 10x. I'm not sure whether the large vga_drawscanline routine is actually helping much yet, but I've left that in and will attempt to speed that up also in another PR.

What would be nice is a BMP image creator tool that converts the image to the VGA standard (or our) palette, and then stores that palette in the same order, using only 16 entries rather than 256 in the BMP palette storage location. This would allow us to speed up display without having to cache nor run any color-matching algorithm on the BMP. Another option would be to write a BMP converter that would input a .bmp and output the same converted. But for now we're doing it at runtime and its pretty fast.

Note the cache only works for 256-color (8 bpp) images, not 24-bit RGB BMP files. Currently, the Paint bmp files are all 256-color 8bpp so it works well.

The vga_drawscanline routine is only implemented for the IA16 compiler for now, but with caching working with all three compilers, all Paint versions are now displaying images much faster than before.

### DIFF
--- a/elkscmd/gui/Makefile.gcc
+++ b/elkscmd/gui/Makefile.gcc
@@ -26,6 +26,7 @@ BINDIR = .
 LOCALFLAGS =
 PROG = $(BINDIR)/paint
 SRCS = app.c gui.c input.c render.c event.c mouse.c graphics.c drawbmp.c cursor.c
+SRCS += drawscanline.c
 OBJS += vga-ia16.oaj
 
 all: $(PROG)
--- a/elkscmd/gui/drawbmp.c
+++ b/elkscmd/gui/drawbmp.c
@@ -163,6 +163,7 @@ draw_bmp(char *path, int x, int y)
     BMPFILEHEAD bmpf;
     static unsigned char imagebits[640*4];  /* static so no stack usage */
     static struct palette bmppal[256];
+    static int cache[256];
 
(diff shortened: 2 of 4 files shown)


Vutshi:
> What would be nice is a BMP image creator tool that converts the image to the VGA standard (or our) palette, and then stores that palette in the same order, using only 16 entries rather than 256 in the BMP palette storage location.

I wrote a BMP RLE8 encoding tool in Python for compressing the icons. It should be easy to make the necessary changes.

toncho11:
https://hackaday.com/2025/04/05/a-look-inside-macpaint/

PR #2282, opened 2025-04-06
Fix for painting outside canvas

Vutshi:
Fix for painting outside canvas.

### DIFF
--- a/elkscmd/gui/render.c
+++ b/elkscmd/gui/render.c
@@ -1,7 +1,7 @@
 #include <stdlib.h>
 #include "app.h"
 #include "render.h"
-#include "event.h"
+// #include "event.h"
 #include "graphics.h"
 
 #define FLOOD_FILL_STACK    100
@@ -11,7 +11,7 @@
 // ----------------------------------------------------
 void R_ClearCanvas(void)
 {
-    fillrect(0, 0, SCREEN_WIDTH-1, SCREEN_HEIGHT-1, BLACK);
+    fillrect(0, 0, SCREEN_WIDTH, SCREEN_HEIGHT-1, BLACK);
 }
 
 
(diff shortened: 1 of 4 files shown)


Vutshi:
I messed up something as it shows some previous commits as part of this PR, but the actual changes in files are correct.

elkscmd/gui/render.c
```diff
@@ -11,7 +11,7 @@
 // ----------------------------------------------------
 void R_ClearCanvas(void)
 {
-    fillrect(0, 0, SCREEN_WIDTH-1, SCREEN_HEIGHT-1, BLACK);
+    fillrect(0, 0, SCREEN_WIDTH, SCREEN_HEIGHT-1, BLACK);
```
ghaerr, line 14: I still don't quite understand this change, this will clear from 0 up to and including SCREEN_WIDTH wide. Isn't that 1 too many? The canvas mouse X clipping above is to SCREEN_WIDTH-1, for instance. This isn't clearing the white dividing line though, so I guess it is working. I tried the previous version and it didn't seem to have an issue though.

elkscmd/gui/render.c
```diff
@@ -1,7 +1,7 @@
 #include <stdlib.h>
 #include "app.h"
 #include "render.h"
-#include "event.h"
+// #include "event.h"
```
ghaerr, line 4: On your next PR, go ahead and take this out.

elkscmd/gui/render.c
Vutshi: Apparently we draw vertical white line at screen_width + 1. 
Btw, why don’t we name it canvas_width as opposed to screenwidth?

elkscmd/gui/render.c
Vutshi: Ok.

elkscmd/gui/render.c
ghaerr: > why don’t we name it canvas_width as opposed to screenwidth?

That's a great idea, especially considering that SCREEN_WIDTH is so close to SCREENWIDTH (the actual graphical screen width), which were globals from the original graphics.c I pulled over from 8086-toolchain/examples. Go ahead and make that change to with your further enhancements, thank you.

> Apparently we draw vertical white line at screen_width + 1.

I see. That probably means we can't actually draw a single pixel brush at SCREEN_WIDTH, since it's clipped.

PR #2285, opened 2025-04-07
Paint circle drawing

Vutshi:
- Add (hardware accelerated!!! 😉) circle drawing
- Cleanup button generation with macros only

XOR works nicely even on 8088
Screenshot 2025-04-07 at 15 56 43


ghaerr:
Wow! Perfect circles and fast hardware XOR drawing, very nice!!!

It's nice how it worked out that the hardware SET vs XOR is able to be decoupled from the drawing primitives; this works because the routines themselves don't touch the VGA Data Rotate Register, allowing upper level calls to modify their functionality without adding tedious APIs.

Thank you for this work, that's a beautiful screenshot and kind of amazing it's all running on ELKS and 8088!!

PR #2291, opened 2025-04-10
Update Paint icons

Vutshi:
The icons now use the correct CGA palette (`drawbmp.c` line 79) and include all 16 colors. The exception is the logo file, `paint.bmp`, which includes only the necessary colors to reduce the file size to 2042 bytes.


ghaerr:
Out of curiosity, did you find it a better idea, or more convenient, to include all 16 colors in the embedded palette when only 3-4 colors are used in the image?

Are these now produced with your Python image converter, or did you manage to get Image Magic or another tool to produce them natively?

Vutshi:
> Are these now produced with your Python image converter, or did you manage to get Image Magic or another tool to produce them natively?

I use my own Python-based BMP converter. It seems that both RLE8—and especially RLE4—have become so obscure that no modern tools can encode them efficiently. For instance, ImageMagick's RLE8 implementation is quite naive; it doesn't even utilize Absolute Mode. Additionally, I haven’t found a way to restrict ImageMagick to only use the necessary colors in the palette.

I suspect that the last program capable of producing well-compressed RLE BMPs is Microsoft Paint. I even installed it under Wine, and it runs fine—right up until I try to save as BMP with RLE compression. At that point, it crashes. So unfortunately, I haven’t been able to compare my encoder's output directly against MS Paint’s.

I do have one test BMP file that achieves slightly better compression than my current Python converter, so there's still room for improvement.

> Out of curiosity, did you find it a better idea, or more convenient, to include all 16 colors in the embedded palette when only 3-4 colors are used in the image?

For a converter, it doesn’t matter—it can handle both. What I’m more curious about is whether `drawbmp` can draw faster if it knows ahead of time that the palette doesn’t need any conversion.

ghaerr:
> ImageMagick's RLE8 implementation is quite naive; it doesn't even utilize Absolute Mode.

What does it actually do then, just set the RLE8 bit and then not actually do any compression? Our drawbmp.c decoder handles absolute lengths, but doesn't implement relative (delta?) lengths. Considering that I wrote this back in 1999 and using Windows to produce the BMPs, I suspect even Windows never produced relative mode RLE BMPs, and I never encountered one.

> I haven’t found a way to restrict ImageMagick to only use the necessary colors in the palette.

You mean always outputs a palette with the RGB colors used in the drawing and can't be forced to any preset palette, or that it always outputs a full-sized palette according to the bits-per-pixel and won't shorten it to the entries actually used?

> right up until I try to save as BMP with RLE compression. At that point, it crashes.

I'm wondering if the Wine guys skipped that implementation also?!

> What I’m more curious about is whether drawbmp can draw faster if it knows ahead of time that the palette doesn’t need any conversion.

I had originally thought just that - that we would produce a special CGA palette based BMP that would then allow the drawbmp code to skip the palette entries and just use the image data's palette indices directly for speed. But then I figured out that it the actual speed bottleneck in drawbmp was that it was calling find_nearest_color for *every* index - which for a 30x30 icon would be 900 times. So a quick cache was added, where find_nearest_color is only called once when finding a previously non-cached index. Given our 3-4 color icons, now find_nearest_color is only called 3 or 4 times per icon, so its very fast, and can only be called max 16 times per image file.

Thus, I don't currently see much of any benefit from assuming a known fixed palette, and the downside would be our image files would not be standard BMPs anymore.

I have recently added drawscanline code attempting to draw an image scanline by scanline, rather than pixel by pixel. So far, due to VGA's planar complexity, I'm not seeing much speed increase, although I think it is faster on real hardware. I hope to be able to use a cycle-accurate emulator like MartyPC to measure any improvement more realistically.

Regarding speed, I am researching using a better compiler optimization than -Os (ELKS default). I have come to realize the default -Os optimize for size emits code that is very slow. I originally tried -O2 and -O3 and it crashed paint, and then the problem went away, so something strange is happening, but both options seem to greatly increase the paint speed when working. I recently added some of our speed improvements to Microwindows, and [someone mentioned](https://github.com/ghaerr/microwindows/pull/123#issuecomment-2793704981) some additional compiler options that have a very good speed for size tradeoff. I hope to get this tested and then introduce it to Paint and all of ELKS.

Vutshi:
> > ImageMagick's RLE8 implementation is quite naive; it doesn't even utilize Absolute Mode.
> 
> What does it actually do then, just set the RLE8 bit and then not actually do any compression?

BMP RLE has three modes: RLE, Absolute, and Delta. The RLE mode compresses repeating pixels, if pixels are random then RLE introduces overhead leading to an anti-compression. Absolute mode is a mode for storing such random pixels untouched and uncompressed with a minimal overhead. Delta is some other crazy mode. The art of RLE compression is to find the right balance between RLE and Absolute Modes.

Image Magick just does RLE mode everywhere.

Vutshi:
> Thus, I don't currently see much of any benefit from assuming a known fixed palette, and the downside would be our image files would not be standard BMPs anymore.
> 
> I have recently added drawscanline code attempting to draw an image scanline by scanline, rather than pixel by pixel.

What if the `drawbmp` function had a flag to indicate whether the palette should be read from the BMP file or if the standard CGA palette should be used instead?

Then, a benchmark that draws all 8 BMP files under different settings and logs the elapsed time to the console could help reveal any performance differences.

ghaerr:
> What if the drawbmp function had a flag 

Well, before the caching code was added, displaying the paint.bmp icon which is 128x128 pixels, took 128 times 128 = 16384 calls to find_nearest_color to display. Now it takes a grand total of 3:
```
'/lib/paint.bmp' bpp 8 comp 1 pal 16 128x128
cache 8
cache 7
cache 0
```
That's why it was so slow. The find_nearest_color function is fast, but calling it 16,000 times for paint.bmp made drawing slow. 

Here's the cache code:
```
        case 8:
            for (w = 0; w < width; w++) {
                unsigned int c;
                struct palette *pal;
                if ((c = cache[image[w]]) == 0xffff) {
                    pal = &bmppal[image[w]];
                    c = find_nearest_color(pal->r, pal->g, pal->b);
                    cache[image[w]] = c;
                    __dprintf("cache %d\n", c);
                }
                image[w] = c;
            }
            vga_drawscanline(image, x, y+h, width);
            break;
```


> Then, a benchmark that draws all 8 BMP files under different settings and logs the elapsed time to the console could help reveal any performance differences.

You're welcome to play with it and can use our new high resolution "p-tick" (0.838ns) timer routines in libc/debug/prectime.c if you want, but you'd only be saving three calls to find_nearest_color() at this point:
```
#include <prectimer.h>

init_ptime();

unsigned long start = get_ptime();
... // do something
unsigned long stop = get_ptime();
printf("elapsed time: %lk\n", stop-start); // automatically displays in usecs, msecs or secs
```

PR #2292, opened 2025-04-10
[paint] Increase speed by changing ia16-elf-gcc -Os to -O2+

ghaerr:
This is an experimental update for paint, changing the optimization flags for the ia16-elf-gcc compiled version.

[Other research](http://git.musl-libc.org/cgit/musl/commit/?id=b90841e2583237a4132bbbd74752e0e9563660cd) has shown that GCC's -Os optimize for size produced some unwanted optimizations, especially those converting left shifts to MUL instructions, which this group of optimization settings overrides, while also apparently producing pretty small binaries.

These new optimizations are meant to increase speed as much as possible for slow 8088 CPUs while keeping the code size increase small.

In paint's case, the size only increased 400 bytes to 15,616 bytes.

I was previously having problems with paint crashing when compiling with raw -O2 or -O3 but for unknown reasons that problem has gone away. For this reason I am calling this change experimental, and upon further evaluation and testing results will consider it permanent and possibly used for the of ELKS applications and possibly the kernel.

### DIFF
--- a/elkscmd/gui/Makefile.gcc
+++ b/elkscmd/gui/Makefile.gcc
@@ -9,10 +9,21 @@ CC = ia16-elf-gcc
 LD = ia16-elf-gcc
 CLBASE = -melks-libc -mtune=i8086 -mcmodel=small -mno-segment-relocation-stuff
 CLBASE += -fno-inline -fno-builtin-printf -fno-builtin-fprintf
+OPTFLAGS = -O2
+OPTFLAGS += -fno-align-jumps
+OPTFLAGS += -fno-align-functions
+OPTFLAGS += -fno-align-loops
+OPTFLAGS += -fno-align-labels
+OPTFLAGS += -fira-region=one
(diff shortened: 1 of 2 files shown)


asiekierka:
Since you're optimizing for the 8088, have you tried `-mtune=i8088` in place of `-mtune=i8086`?

There is a hardcoded default in gcc-ia16, and that is that `O2` uses processor cycles for its instruction cost model, while `Os` uses code size only. As such, the role of `-mtune` becomes somewhat more important (though tuning for the 8-bit bus is applied even with `Os`).

ghaerr:
> have you tried -mtune=i8088 in place of -mtune=i8086?

Thanks @asiekierka, I actually wasn't aware there was an 8088 tuning option! It sounds like gcc-ia16 uses a separate table for opcode cycles for 8088 than 8086, which is great. I'll take a further look at this and the code that's emitted.

PR #2295, opened 2025-04-12
Paint filled shapes

Vutshi:
Add filled rectangle and circle drawing on right-click.


Vutshi:
@ghaerr, how much performance gain do you think we could get by drawing the cursor using XOR on the 8088?

Right now, updating rectangle with XOR actually feels faster than moving a small cursor in MartyPC. I'm wondering if XORing the cursor might make it feel more responsive.

ghaerr:
> how much performance gain do you think we could get by drawing the cursor using XOR on the 8088?

With the recent cursor speed enhancement, displaying a 16x16 cursor draws just the mask pixels, which are about 1/3 to 1/2 of the total so 256/2 = 128 pixels. When moved, the cursor needs to be hidden (=128 pixel writes from saved memory), then be shown (=128 reads to save screen contents for next move plus another 128 writes to display). So 128 reads and 256 writes = 384 total. 

However, looking at the readpixel routine in graphics.c, the EGA/VGA hardware requires a for loop reading 4 planes to be read for each pixel, it's actually 4 times slower reading than writing. So it looks like a cursor move takes (4 read+2 write)*128 = 768 memory operations per change.

An XOR draw doesn't use any reads and only a single write, since the XOR is, as you said, accelerated in hardware. So an XOR cursor change is 1 write to hide the old cursor, and 1 write to show the current cursor = 256 mem operations.

So my guess would be that an XOR cursor would be 768/256 = 3x faster than the current cursor - lots faster, so we should try it :)

Are you thinking we should try a large or small X, similar to the mask in the small cursor? Would you like to take a pass at it, or should I try it?

ghaerr:
Regarding all the "Merge pull requests" showing up in your PRs: I think the reason for this may be that you are not doing a "git checkout master; git pull upstream master; git push" after every PR. This reads from the ELKS master (here) and then pushes to your Github repo. Google "add git upstream" for the exact git remote add command to add the "upstream" variable to be set to this repo.

It's no big deal, except that I have to Squash your commits in order to get rid of those commits, so your PRs end up as a single commit, which you may or may not want.

Vutshi:
> Are you thinking we should try a large or small X, similar to the mask in the small cursor? Would you like to take a pass at it, or should I try it?

My understanding of the cursor code is very bad at the moment. When I tried to make 4x4 pixel cursor only half was showing:). I think it is better if you try it.

ghaerr:
Thanks for your continued enhancements @Vutshi!!

PR #2298, opened 2025-04-15
[paint] Rewrite 16x16 cursor from hexadecimal to visual, 7x7 to 8x8

ghaerr:
Converts the previous hexadecimal representation of the 16x16 cursor to visual, using the MASK() macro.

It was also discovered that the previous MASK macro only created a 7x7 cursor, so the small cursor was redrawn slightly larger as 8x8 and uses the mask() macro.


Vutshi:
I was really puzzled when I first saw this 7x7 cursor—I thought I was missing something. :)

PR #2301, opened 2025-04-16
Paint open bmp

Vutshi:
Now Paint can open bmp file:
`paint <bmp_filename>`
Screenshot 2025-04-16 at 18 07 15

### DIFF
--- a/elkscmd/gui/app.c
+++ b/elkscmd/gui/app.c
@@ -1,4 +1,5 @@
 #include <stdlib.h>
+#include <string.h>
 #include "app.h"
 #include "input.h"
 #include "render.h"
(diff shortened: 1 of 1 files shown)


ghaerr:
You just had to fill in one of those antlers, didn't you?

We're looking like a competitor to Photoshop now! Should we charge for it?

Vutshi:
> You just had to fill in one of those antlers, didn't you?

Yes, it feels amazing:)

> We're looking like a competitor to Photoshop now! Should we charge for it?

Not just yet—we’re still missing light pen support. Rumor has it @dbalsom is working on that for MartyPC! ;)

PR #2302, opened 2025-04-17
[paint] Draw XOR cursor 8 pixels at a time

ghaerr:
@Vutshi asked:

> I thought that a small cursor line is just one byte per plane—maybe two if the cursor position isn’t 8-pixel aligned—so what could be faster than writing a byte directly to memory?

Answer: Not much. And so a much faster cursor display routine was written. This version uses a maximum of two VGA memory writes per 8 bits of the cursor image and never draws pixel-by-pixel. Any 16x16 cursor should be wildly faster on real hardware.

The routine could be made even faster for 16x16 cursors, but I wanted to have this first round tested and get more feedback, as it may be finally fast enough to keep up on 8088. I think we should be able to switch to a larger (16x16 or slightly less) cursor as standard and not have to even worry about an 8x8 cursor using the fast draw code, since the draw time is now not dependent on the number of cursor mask one-bits anymore.

I will commit this now for testing, and plan on some cleanup code later, as there are still some issues seen while testing with the Watcom and C86 compilers. I temporarily added back in the original 16x16 cursor for speed comparisons with XOR drawing. The GCC version is running well and uses @Vutshi's new 11x16 cursor.

As a side note, I think we may be able to use the same-style routine to greatly increase the speed of drawing BMP scanlines. More on that later.

### DIFF
--- a/elkscmd/gui/cursor.c
+++ b/elkscmd/gui/cursor.c
@@ -18,8 +18,8 @@
 #include "mouse.h"
 #include "vgalib.h"
 
-#define USE_SMALL_CURSOR    0       /* =1 small cursor for slow systems*/
 #define USE_XOR_CURSOR      1       /* =1 use XOR drawing for slow systems*/
+#define USE_VGA_DRAWCURSOR  defined(__ia16__)
 
 #define MWMAX_CURSOR_SIZE   16      /* maximum cursor x and y size*/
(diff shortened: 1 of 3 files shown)


Vutshi:
It is fast!
Small bug: the cursor now wraps around—when you move it past the right edge of the screen, it reappears on the left.

PR #2303, opened 2025-04-17
Paint fix flood fill on 'f' key

Vutshi:
'f' command didn't work correctly with the new mode switching logic.

### DIFF
--- a/elkscmd/gui/input.c
+++ b/elkscmd/gui/input.c
@@ -105,8 +105,10 @@ void I_HandleInput(void)
                     break;
 
                     case 'f':
-                        current_mode = mode_Fill;
-                        current_state = state_Finalize;
+                        current_color = currentMainColor;
+                        hidecursor();
+                        R_LineFloodFill(omx, omy, current_color, readpixel(mx, my));
+                        showcursor();
                     break;
 
                     case 'm':
--- a/elkscmd/gui/input.c
+++ b/elkscmd/gui/input.c
@@ -105,9 +105,8 @@ void I_HandleInput(void)
                     break;
 
                     case 'f':
-                        current_color = currentMainColor;
                         hidecursor();
-                        R_LineFloodFill(omx, omy, current_color, readpixel(mx, my));
+                        R_LineFloodFill(omx, omy, currentMainColor, readpixel(mx, my));
                         showcursor();
                     break;
 
--- a/elkscmd/gui/input.c
+++ b/elkscmd/gui/input.c
@@ -32,6 +32,8 @@ void I_HandleInput(void)
 
                 mx = event.x;
                 my = event.y;
+
+                hidecursor();
                 if (mx < CANVAS_WIDTH){
                     current_color = currentMainColor;
                     switch (current_mode){
@@ -46,7 +48,6 @@ void I_HandleInput(void)
                             startX = mx;
                             startY = my;
                             lastRadius = 0;
-                            hidecursor();
                             set_op(0x18);    // turn on XOR drawing
                             goto break_mousedown;
 
@@ -55,7 +56,6 @@ void I_HandleInput(void)
                             AltFinalize = (event.button == BUTTON_R);
                             startX = mx;
                             startY = my;
-                            hidecursor();
                             set_op(0x18);    // turn on XOR drawing
                             goto break_mousedown;
 
@@ -65,7 +65,6 @@ void I_HandleInput(void)
                             break;
                     }
                 }
-                hidecursor();
 
             break_mousedown:
             break;


Vutshi:
@ghaerr, sometimes Paint only responds to a key press after the mouse is moved, but other times it reacts immediately. Do you know why that happens?

ghaerr:
> sometimes Paint only responds to a key press after the mouse is moved, but other times it reacts immediately. 

I've been wondering whether you've noticed that or not. I've seen it for a while, and thought I had a handle on it, but the fix I made seemed to introduce other problems. I'll up the priority again and try to fix it.

I'll also fix the wrapping cursor problem and some more code speedups now that I know it's working well for you. Are you thinking we might be able to get rid of having to keep a smaller cursor for speed, and just use a standard larger cursor?

Vutshi:
> I've been wondering whether you've noticed that or not. I've seen it for a while, and thought I had a handle on it, but the fix I made seemed to introduce other problems. I'll up the priority again and try to fix it.

This behaviour is very strange. It seems like the keys reaction is postponed depending on the position of the mouse cursor. If I shift it by a pixel or two it starts to work normally, another shift by pixel or two and keyboard is ignored again.

I inserted print to console in `event.c`
```
       73         event->keychar = buf[0];
       74         event->type = EVT_KEYCHAR;
                __dprintf("%s: key pressed\n", buf);
       75         goto out;
```
The same thing happens: if cursor is in an unlucky position the pressed keys are accumulated somewhere and then are printed altogether on mouse move. Furthermore, the `tab` character is printed as shown in the last five lines below:

```
m: key pressed
mD: key pressed
mD: key pressed
hD: key pressed
gD: key pressed
yD: key pressed
m	: key pressed
gD	: key pressed
hD	: key pressed
tD	: key pressed
rD	: key pressed
```
How anything is added to `buf` which is defined as length 1?

ghaerr:
Agreed it's strange. Are you sure its related to mouse cursor position, or just that you can move the cursor a little each time and then it happens again? It seems to me the latter, but I'm in the middle of tightening the new cursor code and found one of our GCC set_op macros was not being interpreted correctly and trashing a register, causing bad cursor display. It is possible that any other set_op mixed in with code is having the same effect, which only occurs when 'char' variables are also  used in the function. I will test my code as well as your cursor position idea shortly as soon as I have it fully working to see if this might be the issue. We have the same issue with Watcom though so not likely.

The problem you are having with displaying buf[0] is that buf[1] is not NUL terminated; that means that printf will keep on going and print trash sometimes, depending on what's on the stack. To properly display, you'd need to specify %c, or increase buf to buf[2] and set buf[1] = '\0';

Vutshi:
New data point. Polling events fixes the problem of keyboard block, I use this in `input.c`:
```
// if (event_wait_timeout(&event, EV_BLOCK))
    if (event_wait_timeout(&event, EV_POLL))
```

ghaerr:
> Polling events fixes the problem of keyboard block

Geez, maybe its a kernel bug. Glad you found at least a temp workaround. I'll add that to the code I'm working on and submit it, thanks.

Running Paint without blocking means that the CPU will be spending tons of time in paint, rather than doing something else for another "user". But I don't think that's too big a concern for this workaround, since likely nothing else is running on the system. When networking, this could be an issue, but we'll leave that for later and I'll write this up as a possible kernel bug pending more investigation.

Vutshi:
> Are you sure it's related to mouse cursor position, or just that you can move the cursor a little each time and then it happens again? It seems to me the latter

You are right, it doesn't depend on absolute cursor position. I tried different starting mouse positions by changing `posx` in `int event_open(void)` and it doesn't affect anything. In the beginning keyboard always works.

It must be related to mouse motion events.

Vutshi:
@ghaerr , I was looking into `mouse.c` and its assembler and found two `mul` operations in `filter_relative`. I couldn't get rid of them by doing << 1 manually:
```
0000009c <filter_relative.constprop.0>:
  9c:	57                   	push   %di
  9d:	89 e3                	mov    %sp,%bx
  9f:	8b 4f 04             	mov    0x4(%bx),%cx
  a2:	8b 7f 06             	mov    0x6(%bx),%di
  a5:	bb 01 00             	mov    $0x1,%bx
  a8:	85 c9                	test   %cx,%cx
  aa:	7d 05                	jge    b1 <filter_relative.constprop.0+0x15>
  ac:	f7 d9                	neg    %cx
  ae:	8d 5f fe             	lea    -0x2(%bx),%bx
  b1:	83 f9 05             	cmp    $0x5,%cx
  b4:	7e 05                	jle    bb <filter_relative.constprop.0+0x1f>
  b6:	d1 e1                	shl    %cx
  b8:	83 c1 fb             	add    $0xfffb,%cx
  bb:	89 c8                	mov    %cx,%ax
  bd:	f7 e3                 **mul    %bx**      
  bf:	89 c1                	mov    %ax,%cx
  c1:	bb 01 00             	mov    $0x1,%bx
  c4:	85 ff                	test   %di,%di
  c6:	7d 05                	jge    cd <filter_relative.constprop.0+0x31>
  c8:	f7 df                	neg    %di
  ca:	8d 5f fe             	lea    -0x2(%bx),%bx
  cd:	83 ff 05             	cmp    $0x5,%di
  d0:	7e 09                	jle    db <filter_relative.constprop.0+0x3f>
  d2:	89 f8                	mov    %di,%ax
  d4:	d1 e0                	shl    %ax
  d6:	83 c0 fb             	add    $0xfffb,%ax
  d9:	89 c7                	mov    %ax,%di
  db:	89 0e 12 00          	mov    %cx,0x12
			dd: R_386_16	.bss
			dd: R_386_SUB16	.bss!
  df:	89 f8                	mov    %di,%ax
  e1:	f7 e3                 **mul    %bx**
  e3:	a3 10 00             	mov    %ax,0x10
			e4: R_386_16	.bss
			e4: R_386_SUB16	.bss!
  e6:	5f                   	pop    %di
  e7:	c3                   	ret
```

ghaerr:
> found two mul operations in filter_relative.

Yes, those are from the `x *= sign` and `y *= sign` statements, they won't get optimized out. We'd have to rewrite the code a bit to get rid of them. I'll look further into it. I'm still running on QEMU but given the latest XOR speed, I was actually thinking of removing the filter_relative entirely, as it seems the mouse acceleration might now be too much at 2. It was three from Microwindows, but its cursor code is quite a bit slower right now.

You can test acceleration by just changing SCALE to 1 to see if you like that better. If so, we'll set filter_relative to be a compiler define inside mouse.c.

PR #2307, opened 2025-04-18
Initial WonderSwan port

asiekierka:
oops

`sash` does run! However, little else runs. And a WonderSwan Color or SwanCrystal is required; it does not work on the original WonderSwan due to insufficient RAM.

also on WonderSwan Color

TODO (but not necessarily merge blockers):

- [x] Disable building `clock` and `nano-X` for the WonderSwan target
  - `clock` can be ported to use on-cartridge RTC. `nano-X`, uh,
- [x] Expand build system to create a proper WonderSwan ROM
  - I typically use Python for scripts like this, but that hasn't been introduced to the codebase yet, so I welcome suggestions
  - 1 MiB ROM file, the ROMFS has to be at `0x40000` in the file (or wherever it's configured), the kernel image has to be at `0xE0000` (or wherever it's configured), and there has to be a [ROM footer](https://ws.nesdev.org/wiki/ROM_header) at `0xFFFF0`.
- [ ] More extensive testing of `/dev/ttyS0`
  - The emulators I have on hand all have a variety of issues that seem to prevent more extensive testing.
- [x] Using `-march=v30mz -mtune=v30mz` to build the kernel
  - Given the improvements are moderately decent, should this be extended to applications?

There's a lot that could be done (adding support for 4x8 or 4x6 font displays for an effective 56x18 or 56x24 on-device resolution, adding support for more ANSI attribute features than just foreground; binding the console's buttons to at least *some* operations - or even adding a virtual on-screen keyboard), and user RAM bank switching exploration could allow running more than just `sash`, but I think this is a good proof of concept. I will be evaluating some of those later.

Note that I haven't yet reviewed the code extensively, as I'm going to be busy tomorrow...


ghaerr:
Hello @asiekierka,

Wow! Very nicely done, impressive work. :)

Most everything looks fantastic. I do have a few small comments and possible minor changes I'll post alongside the code. After discussing those, we can get this committed, so you can more easily continue.

Thanks for the photo! How fast does the Swan scroll text?

> sash does run! However, little else runs.

Any ideas why? Out of memory, crashing, error messages?

>  Expand build system to create a proper WonderSwan ROM
> I typically use Python for scripts like this, but that hasn't been introduced to the codebase yet

You're welcome to use your favorite scripting language to produce a build, but I'm not sure I want to add a Python dependency for the general build, which means the Swan build won't likely get added to the master buildimages.sh script, but users instead would need to have their own Python etc installed after configuring with swan.config. I don't see that as a big deal for now, really.

Added scripts could go under elks/scripts/ or scripts/. I don't think any of those (shell) scripts are used now.

> Restoring CONFIG_CPU_* and using -march=v30mz -mtune=v30mz throughout

I've spent years removing way too many CONFIG_ options (there were FAR too many to reliably maintain), so not sure whether using CONFIG_CPU_* is the way to go for your optimizations, versus adding `-march=v30mz -mtune=v30mz` etc into elks/Makefile-rules under an ifdef  CONFIG_ARCH_SWAN for the time being. 

I haven't yet had time to investigate the -mtune=i8088 you previously suggested, but plan to do so, after which we can revisit this. The reason for not bringing back a full-blown CONFIG_CPU_* set of options at this time is that a decision was made to ensure that ELKS can boot and run on any IBM PC right out of the box, whereas adding differing CPU options for the IBM PC target brings with it more maintenance and porting work, usually with very little real gain. We're not a production target and most code optimizations, (except for graphics) just don't matter much, and those that do are written in assembly. That said, anyone should be able to change to their own optimization settings by editing elks/Makefile-rules or something similar should they want to.

> Disable building clock and nano-X for the WonderSwan target

This can most simply be done for the time being by adding ifdef CONFIG_ARCH_SWAN lines to the various elkscmd/*/Makefiles. I can see that the for instance elks/arch/i86/drivers/char/Makefile is already a mess, we'll have to clean this up at some point.

Thank you!

elks/arch/i86/boot/setup.S
```diff
@@ -262,6 +262,16 @@ no_blob:
...
 start_os:
+#ifdef CONFIG_ARCH_SWAN
+        // Force color mode
+	mov	$0x80,%al
+	out	%al,$0x60
+
+	// Enable UART serial
+	mov	$0xE0,%al
+	out	%al,$0xB3
+#endif
```
ghaerr, line 269: I would prefer to move this ifdef code into the arch-specific arch_set_initseg, but I can see that you can't because the hello message is displayed prior to the call. So we'll leave this hear for now until I think of whether to move the hello message there also or some other solution. It seems like making a call to arch_set_initseg immediately after clearing the setup data segment would be best for handling arch-specific system initialization.

elks/arch/i86/drivers/char/console-font-8x8.S
```diff
@@ -0,0 +1,174 @@
+	.section .text
+	.global font_8x8
+font_8x8:
+	.byte 0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x7E,0x81,0xA5,0x81
```
ghaerr: Can you rename this file console-font-8x8-swan.S? That will make it very apparent where it's used.

elks/arch/i86/kernel/irqtab.S
```diff
@@ -239,10 +239,30 @@ ret_from_syscall:
...
 */
 updct:
+#ifdef CONFIG_ARCH_SWAN
+//
+//	Acknowledge interrupt
+//
+//	Note that the WonderSwan RTC/UART interrupts cannot be acknowledged
+//	without processing them, which is why interrupts are not re-enabled
+//	for IRQ work on this platform.
+//
```
ghaerr: I was wondering why the STI/CLI was removed below until I saw this comment. Running interrupt handlers with interrupts disabled may or may not be the best solution for Swan, as there are assumptions made in various areas that the interrupt handler has interrupts enabled.

I'm not sure what "interrupts cannot be acknowledged without processing them" means exactly? Will another interrupt come flying in after the OUT 0xB6 below, or something else? Anything that would help interrupt handlers run with interrupts enabled will help efficiency as well as system design constraints.

In particular, various interrupt handlers or other parts of the kernel may be written with just clr_irq()/set_irq() instead of save_flags()/clr_irq()/restore_flags(); If written using the former rather than the latter interrupts could get inadvertently reenabled when the Swan needs then off. This is likely easily handled in the interrupt handlers, but more complicated in the kernel, as some kernel routines are called back at interrupt time, but never schedule() except under very controlled circumstances.

elks/arch/i86/kernel/irqtab.S
```diff
@@ -254,7 +274,9 @@ updct:
         pop     %bx
 
         pop     %ax             // Saved IRQ
+#ifndef CONFIG_ARCH_SWAN
         cli                     // Disable interrupts to avoid reentering ISR
+#endif
```
ghaerr: This CLI may still want to stay as a result of the potential issues with the kernel possibly having enabled interrupts inadvertently (for Swan only). The CLI is very important here because otherwise there is the possibility of overflowing the kernel stack. I would suggest leaving this one in, just in case.

elks/arch/i86/kernel/timer.c
```diff
@@ -79,6 +79,15 @@ void timer_tick(int irq, struct pt_regs *regs)
...
 
+#ifdef CONFIG_ARCH_SWAN
+    /* spin timer segments */
+    if (spin_on && !(jiffies & 7)) {
+        static unsigned char wheel[4] = {0x00, 0x20, 0x10, 0x08};
+        static unsigned char c = 0;
+
+        outb((inb(0x15) & 0x07) | wheel[c++ & 0x03], 0x15);
+    }
+#else
```
ghaerr: Do you really want this in here? This little bit of fancy is only called from the BIOS block device driver, so I'm wondering whether it will ever run on Swan. It's purpose is to spin a visual while waiting for very slow disk I/O.

If you think that Swan will not use this feature or have any very slow I/O, I would remove it and the similar code below in spin_timer().

elks/include/linuxmt/config.h
```diff
@@ -90,6 +90,23 @@
...
+#define SETUP_MEM_KBYTES        128     /* base memory in 1K bytes */
+#define SETUP_XMS_KBYTES        0       /* xms memory in 1K bytes */
+#define SETUP_ROOT_DEV          0x0600  /* root device ROMFS */
+#define SETUP_ELKS_FLAGS        0       /* flags for root device type */
+#define SETUP_PART_OFFSETLO     0       /* partition offset low word */
+#define SETUP_PART_OFFSETHI     0       /* partition offset high word */
+#define SYS_CAPS                0       /* no XT/AT capabilities */
+#define UTS_MACHINE             "swan"
+#define SETUP_HEAPSIZE          32768
+#define CONFIG_MEM_SEGMENT      0x1000
```
ghaerr: Since CONFIG_MEM_SEGMENT is brand new, please add a comment after it explaining its purpose, something like /* start segment for application memory heap */ or similar.

elks/arch/i86/drivers/char/console-font-8x8.S
asiekierka: My initial thinking was that future "graphics mode as text mode" drivers could use the same file, making it a bit more generic; but sure.

elks/arch/i86/boot/setup.S
asiekierka: I can move *one* of them there, as `enabling color mode` (to enable 64 KiB of RAM, as opposed to 16) is only required later in the initialization process.

elks/arch/i86/kernel/irqtab.S
asiekierka: > I'm not sure what "interrupts cannot be acknowledged without processing them" means exactly? Will another interrupt come flying in after the OUT 0xB6 below, or something else?

The WonderSwan has two types of interrupts, and two subtypes of hardware interrupts:

- [CPU interrupts](https://ws.nesdev.org/wiki/NEC_V30MZ_interrupts), of which there are six (INT 0-5); those are the standard 80186 ones, but the invalid opcode and ESC interrupts are not implemented.
- [Hardware interrupts](https://ws.nesdev.org/wiki/Interrupts), of which there are eight (mapped on the ELKS port to INT 8-15); in these, we further distinguish two types:
  - Edge-triggered, which are requested by the hardware on the edge of a condition: that covers all the timer-related interrupts, as well as the key press interrupt. These will be cleared by simple acknowledgement.
  - Level-triggered, which are requested by the hardware constantly while a condition is true; that covers the UART and cartridge interrupts. These can only be cleared by clearing *the condition* causing the interrupt, and *then* acknowledging the interrupt.

The problem is with the level-triggered interrupts; in particular, the "UART RX ready" interrupt will fire as long as the (one-byte) RX queue is not serviced by reading the value out, so it will fire immediately after the first `sti` in that interrupt handler. There are a few solutions I thought of for this:

- The one I implemented, which is "simply disable interrupts until the IRQ handler runs".
- Only disable interrupts for level-triggered interrupt numbers; this will limit the scope of disabled interrupts to the UART driver.
- Clear the "interrupt enable" bit for the given hardware interrupt, then expect the IRQ handler to re-set it. This will also do the trick.

In addition, interrupt handling is slightly slow on the WonderSwan, requiring about 30-35 cycles of bringup between an interrupt is requested and user interrupt code starts executing. This is a limitation of the V30MZ CPU architecture, presumably because they microcoded the whole thing. This informed my initial guess at a solution above.

What do you think?

asiekierka:
> You're welcome to use your favorite scripting language to produce a build, but I'm not sure I want to add a Python dependency for the general build, which means the Swan build won't likely get added to the master buildimages.sh script

That's why I intended to ask what scripting languages are already depended on (if there are any other than the shell); I'm flexible. Sorry for not being clear enough.

> The reason for not bringing back a full-blown CONFIG_CPU_* set of options at this time is that a decision was made to ensure that ELKS can boot and run on any IBM PC right out of the box, whereas adding differing CPU options for the IBM PC target brings with it more maintenance and porting work, usually with very little real gain.

True; while supporting `-mtune=...` configuration options without `-march=...` would still allow users to compile code *optimized* for their specific machine, it would add maintenance hassle for little real gain. However, with the addition of non-IBM platforms which have fixed CPUs (and especially one with very different optimization strategies like the Swan), I feel your suggested solution is fine with me.

asiekierka:
I checked: making the kernel build using `-march=v30mz -mtune=v30mz` on `swan` reduces the image size by 1456 bytes. As the kernel is platform-specific, I feel there's no issue in adding that in this PR already. The question is - should I extend this to applications as well, given that some of them compile in an platform-specific manner regardless?

elks/arch/i86/kernel/irqtab.S
ghaerr: > What do you think?

It is perhaps a bit tricky since anything we're talking about is just more overhead, decreasing overall throughput if not done correctly.

My first thoughts would be since the real problem is only the UART with its level-triggered interrupt, it's probably not worth writing mainline code that tries to differentiate the interrupt and clear it specially (option 3 or 2). Since it's better for ELKS that we continue with the assumption that interrupt handlers run with interrupts enabled, especially when calling back into core kernel routines, I would seriously consider the following:
- Don't enable interrupts in mainline interrupt ASM code before calling C handler.
- Specifically enable interrupts as first thing in handler using set_irq() in all but serial UART interrupt. This allows the rest of the routine to run normally, as far as the rest of the kernel is concerned, as well as prioritized interrupt handling.
- Enable interrupts in the serial handler immediately after servicing the RX queue.

I would presume going the above route would mean that each Swan interrupt handler would also call a function (or GCC asm macro) to acknowledge the interrupt when appropriate. Overall, this means that the Swan port deviates from a standard ELKS port only by not enabling interrupts before the specific C handler, and having each handler enable interrupts ASAP and also acknowledge them internally, unless it makes sense to combine the interrupt acknowledgement code after calling the C handler, like the normal ELKS case.

asiekierka: > Specifically enable interrupts as first thing in handler using set_irq() in all but serial UART interrupt. This allows the rest of the routine to run normally, as far as the rest of the kernel is concerned, as well as prioritized interrupt handling.

Since we have to do a platform-specific patch with this approach anyway, I think it would be a better idea to add a Swan specific `ack_hw_irq([0-7])` which performs the acknowledgement write right before (and including) `set_irq()`. This would allow removing the `PUSH/POP/MOV/SHL` dance from the main interrupt handler, which would save about ten cycles per interrupt.

ghaerr:
> That's why I intended to ask what scripting languages are already depended on (if there are any other than the shell)

Only awk. I'm a big believer in "smaller is better" especially here at ELKS, and that goes for complicated build scripts as well. But I understand when it gets complicated for specific builds, although in those cases we've written a C program and stuck it in the elks/tools/ directory. Whenever a new, cool tool is required, it ends up with users asking questions and problems porting the tool. An interesting sideline, we're dependent `mtools` a DOS FAT filesystem creator. Long story short, newer versions of that package broke earlier features for our old images, and that then escalated into sometimes not being able to find the older specific build on Github CI Ubuntu versions, etc, etc. If you can do it with a bash shell script, all the better for us IMO.

ghaerr:
> making the kernel build using -march=v30mz -mtune=v30mz on swan reduces the image size by 1456 bytes.

Wow. I would have to say that sounds huge. Any ideas as to what's changed? Is that just tons of ENTER/LEAVE instead of longer prologues, or what do you think?

> should I extend this to applications as well, given that some of them compile in an platform-specific manner regardless?

I say yes, you should extend this to applications, but not because they're already platform-specific; instead because you're ROM-based and every byte counts. The Swan build is already likely to be a custom build, so it doesn't really matter whether the applications are built differently, since they'll never run on PC platforms. Our issue with the problems resulting from platform-specific application builds have to do with floppy images being traded around and kernel APIs, etc.

asiekierka:
> instead because you're ROM-based and every byte counts.

To be fair, there's a *lot* of ROM space. The trivial image layout allows for anywhere between 384 and 704 KiB of ROMFS (depending on subtargets, which I'll get to adding at a later date), and if I adjust the ROMFS driver to support banking, this can easily go up to about 8-16 MiB (that's the size modern flash cartridges typically support for this platform). So, that's not necessarily a good reason.

elks/arch/i86/kernel/irqtab.S
ghaerr: > I think it would be a better idea to add a Swan specific ack_hw_irq([0-7])

That sounds like a good idea, go for it. How about just ack_irq([0-7])? Stick the forward declaration in elks/include/arch/irq.h and label it specific to Swan.

ghaerr:
> To be fair, there's a lot of ROM space

> So, that's not necessarily a good reason.

In that case, don't change the application compilation tunings. There's then no really good reason for it and it's just one less thing changed to not worry about later.

ghaerr:
> given that some of them compile in an platform-specific manner regardless?

And to that point, we don't really want platform-specific applications at all. Currently, the reasons we've got them are 1) easier to write a single simple application than construct an new kernel API for just one need but uses kernel address space which is in extremely short supply, 2) very specific hardware platform variants, like RTC clock adaptors used once e.g. `clock` time setting, and 3) the OS has no hardware protection anyways so no real point in writing a special system call for say, I/O address IN/OUT instructions.

asiekierka:
There, that should be everything. After these patches, the `swan` target builds an `image/rom.wsc` file that can be ran in an emulator of choice (though getting keyboard input via the UART set up is a bit of a challenge).

asiekierka:
FYI, I commented out some features in `sash` and got it to be small enough to run `ps` and `meminfo`:

```
# ps
  PID   GRP  TTY USER STAT CPU  HEAP  FREE   SIZE COMMAND
    1     0      root    S   2  1204  6778  37280 /bin/sash
    9     0      root    R   0     0  2192   8816 ps
# meminfo
  HEAP   TYPE  SIZE    SEG   TYPE    SIZE  CNT  NAME
  1BCE   MEM     16   1000   CSEG   20432    1  /bin/sash
  1BEA   TASK 13280
  4FD6   INOD  6720
  6A22   FILE   896
  6DAE   CACH  8192
  8DBA   BUFH   128
  8E46   TTY     80
  8EA2   TTY     80
  8EFE   MEM     16   14FD   DSEG   16848    1  /bin/sash
  8F1A   MEM     16   191A   free   16848    0
  8F36   MEM     16   1D37   CSEG    3472    1  meminfo
  8F52   MEM     16   1E10   DSEG    2320    1  meminfo
  8F6E   MEM     16   1EA1   free    5616    0
  8F8A   free  3128
  Heap/free   32768/ 3128 Total mem   65536
  Memory usage   64KB total,   42KB used,   22KB free
```

I wonder what's up with `/bin/sash` having two `DSEG`-sized sections before the area of memory dedicated for `meminfo`. I'd look into what it is reserved for, but I have to dash for now...

ghaerr:
> I wonder what's up with /bin/sash having two DSEG-sized sections before the area of memory dedicated for meminfo

The second DSEG is actually free memory, although it's exactly the same size as sash's DSEG. It's the result of sash's `fork` duplicating the data segment before calling exec on meminfo, whose memory is then allocated before the new process's old data segment is released. A good reminder that in UNIX/Linux, a shell needs twice its data segment in order to run another program. 

Thanks for the Swan contribution and making the requested changes!

ghaerr:
@asiekierka,

You'll want to do a git pull before proceeding, I had to fix the following error which broke the CI build:
```
diff --git a/image/Make.defs b/image/Make.defs
index 207f4221..44235cd5 100644
--- a/image/Make.defs
+++ b/image/Make.defs
@@ -135,5 +135,5 @@ endif
 
 # override default image name if NAME specified
 ifdef NAME
-TARGET_FILE = $(IMG_DIR)/$(NAME).$(TARGET_EXT)
+TARGET_FILE = $(IMG_DIR)/$(NAME)$(TARGET_EXT)
 endif
```

In the future, you can test the CI make by running `./buildimages.sh` or `./buildimages.sh fast` in ELKS root.

toncho11:
@asiekierka do you need a special cartridge to run it on WS color?

asiekierka:
@toncho11 Yes. You need a reflashable cartridge with a generous amount of SRAM (at least 64 KiB for now, 256 KiB in the future) and at least 512 KiB of ROM space.

Vutshi:
@asiekierka, inspired by your success I found even more challenging porting target:
Vtech
The display is the size of two Paint buttons—64x32 pixels :)

asiekierka:
I appreciate your suggestion, but is that even an 8086?

Vutshi:
> I appreciate your suggestion, but is that even an 8086?

I have no idea. Just got it from my neighbors today :)

Vutshi:
> > I appreciate your suggestion, but is that even an 8086?
> 
> I have no idea. Just got it from my neighbors today :)

it is made in 2017

PR #2311, opened 2025-04-20
[kernel] Store syscall table in ROM on ROM platforms

asiekierka:
Since the system call table is already in memory on ROM platforms, and it doesn't appear to be written to at runtime, there's no need to copy it from the text segment to the data segment.

This saves about 416 bytes of RAM on those platforms at the expense of having to process the `cs:` segment modifier as part of a syscall.


ghaerr:
Another nice saving of RAM for ROM systems, thank you!

PR #2312, opened 2025-04-20
[kernel] Place temporary stacks in BSS over data

asiekierka:
This reduces the kernel near data size by 640 bytes, which reduces the Image size a little on both disk and ROM platforms.


ghaerr:
Thanks @asiekierka, nice catch!

Issue #2293, opened 2025-04-11
PC-98 image does not boot on Neko Project 21/W emulator


tyama501:
Hello @ghaerr ,

Nice to see great improvements recently :) Than you.
I still don't have my build environment in the new house, (mainly because of my PC for the build is still in the box) 
but I downloaded fd1232.img and tried on the Neko Project 21/w emulator.
Unfortunately, it seems that it won't boot on NP21/w although it boots on DOSBox-X.

If the hard disk is not attached,
it stops here.

![Image](https://github.com/user-attachments/assets/2d0815a8-b4f6-43e2-9128-ceb95b7c8e08)

If the IDE hard disk is attached,
it stops here.

![Image](https://github.com/user-attachments/assets/ad387959-86aa-4785-a0d3-ca2aab66a28d)

Current memory setting is 640KB + 3MB.

Do you have any idea how should we debug this?
It might related to the memory issue,
but I wonder the hard disk issue is coming from the same issue or not.

Thank you.

ghaerr:
Hello @tyama501,

> it seems that it won't boot on NP21/w although it boots on DOSBox-X.

That's good news it boots on DosBox-X, this could mean that either XMS or HMA isn't well supported on Neko, or it might be something to do with the hard drive. When a hard disk is seen, early on the kernel reads the MBR to get the partition table, and that means the XMS buffer system is used, so XMS is still possibly related. 

> Do you have any idea how should we debug this?

Can you edit /bootopts on the boot disk?

I would suggest taking two steps: commenting out xms=on in /bootopts will turn off XMS. If that boots, then it would seem that XMS memory, or at least INT 1F BIOS block move isn't supported in Neko. However, HMA also uses memory above 1M, so you may want to comment out hma=kernel as well. If commenting that out, then perhaps the A20 line isn't supported. If this is so, then we'll need you to write some ASM in setup.S to setup A20 for HMA booting.

With both lines commented out, the system should hopefully boot, I don't think we've made any other changes that might affect this.

Thank you!

tyama501:
Hello @ghaerr 

Thank you,
I will try that later.
Neko is known as xms well supported and I have implemented the xms buffer to elks with the int1F in it before.

Actually DosBox-X int1F has been fixed referred to NP2 recently.
https://github.com/joncampbell123/dosbox-x/issues/5463

I haven't used that DosBox-X yet, so I may try that as well.

ghaerr:
[...]

PR #2318, opened 2025-04-24
[kernel] Fix variety of PC-98 issues when using XMS

ghaerr:
This PR fixes or provides PC-98 system enhancements for the following issues found and described by @tyama501 and @drachen6jp in #2293:

- Fix boot crashes seen on various hardware with extended memory (XMS) sizes
- Handle possible XMS memory hole issue between 15MiB and 16MiB on some systems. If a hardware memory hole exists, by BIOS byte 401h indicating < 15MiB XMS memory present below 16MiB, any upper XMS memory >= 16MiB indicated by BIOS word 594h will be ignored. This effectively means that only the XMS memory below any memory hole will be used
- Automatically detect CPU type of 286 vs 386, and selecting either INT 1F or unreal mode for XMS, respectively
- Automatically disable XMS when on 286 CPU and HMA kernel loaded. This is required until the INT 1F block move function is moved to lower memory and the A20 gate is reenabled, coming in next PR. XMS can be enabled in this situation by removing hma=kernel in /bootopts
- Shows CPU type in boot screen: 6=80286, 7=80386+

PC-98 BIOS memory locations used for XMS calculations:
    0:501h, 3bit size - 1 in 128k of memory in 1-1M range
    0:401h, byte size of memory in 128k of memory in 1M-16M range
    0:594h, word size of memory in 1024k of memory in 16M-4GB range

Tested only on DosBox-X. It seems that DosBox-X always keeps A20 gate enabled after INT 1F, unlike real hardware; this will make moving block-move out of HMA kernel hard to test. 

Testing all XMS modes (xms=on, xms=int15) is needed on 286 and 386 real hardware, as well as on Neko 21/W. This PR enables the first use of unreal mode on any PC-98 system.

@tyama501: I now understand what you are saying in your last issue with portions of the filesystem disappearing when the system is booted on XMS on 286. Unfortunately, I cannot duplicate this problem on DosBox-X. It is possible this has something to do with INT 1F call on 286, versus 386 or emulator? Can you duplicate this on another emulator? Perhaps @drachen6jp can test this problem on his 286 or 386 hardware?

Also, is Neko 21/W or another emulator available for macOS that I might be able to setup and use? @drachen6jp I see you are using a private build of QEMU for PC-98, very cool. Could that be compiled for macOS, that would be very nice for testing!

### DIFF
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -1247,7 +1247,7 @@ checkhma:
 	mov	%ax,%ds
 	mov	xms_kbytes,%ax		//. must have xms size >= 64k
 	cmp	$64,%ax
-	jl	1f
+	jc	1f
 	mov	hma_kernel,%al		// and hma=kernel in /bootopts
 	test	%al,%al
 	jz	1f
@@ -1312,7 +1312,7 @@ arch_get_mem:
 	push %bx
 	xor %ax,%ax
 	mov %ax,%es
-	mov $0x501,%bx
+	mov $0x501,%bx                  // 3 bit size - 1 in 128k of 0-1M mem range
 	mov %es:(%bx),%cl
 	and $7,%cl
 	inc %cl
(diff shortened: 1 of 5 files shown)


tyama501:
Hello @ghaerr 

Thank you!
I will look it closely later.

There is Neko emulator NP2Kai 
which I am using on the Debian linux.
It shows almost same behavior as NP21/W.
It seems there is a brew instruction for macOS
and SDL.

https://github.com/AZO234/NP2kai?tab=readme-ov-file#sdl

tyama501:
https://github.com/AZO234/NP2kai?tab=readme-ov-file#install-tools-3

tyama501:
https://github.com/AZO234/NP2kai?tab=readme-ov-file#build-3

tyama501:
(Hmm, links may not work. Need to open SDL tab on the top page

toncho11:
When is the int15 required?

ghaerr:
Int 15 is no longer required. I've left it compiled in for regression testing and for emulators that don't support the semi-documented LOADALL instruction. For instance, I can now use PCem to emulate LOADALL for IBM PC, but not (yet) for PC-98. For PC-98 on 80286, I still have to use INT 15 on PCem or DosBox-X for testing.

Otherwise, ELKS XMS support is totally automatic when CONFIG_FS_XMS is set, and the proper mode of operation, i.e. unreal mode (for 80386+) or LOADALL (for 80286) is used when xms=on in /bootopts. To force INT 15 on real hardware, use xms=int15. Technically, if an 8086 had XMS, ELKS would also use INT 15 for that system.

toncho11:
I updated: https://github.com/ghaerr/elks/wiki/Access-to-XMS-memory

PR #2321, opened 2025-04-24
Fix flood fill stack overflow

Vutshi:
@ghaerr you were right about the off-by-one error. I increased the stack a little bit as well.


ghaerr:
Thanks for testing and fixing this!

PR #2324, opened 2025-04-29
286 Loadall block_move

drachen6jp:
thanks @ghaerr 
I've checked this code.with my machine and 286 emulators.

your question
1.
why did't I use "db0xf 0xb"? 
ia16-elf-as refuse this pattern.

2. why call set_a20_enable?
my mistake. Loadall set new DS. but I reset old DS. so after movsw push ds and pop ds.

3.
mov dx,37
mov al,6
out dx,al

this is PC-98 beep code. I 've not deleted this debug code.

4.meminfo doesn't run.
menuconfig  Memory character devices was not set.sorry.

thank you!

### DIFF
--- a/elks/arch/i86/lib/fmemory.S
+++ b/elks/arch/i86/lib/fmemory.S
@@ -242,10 +242,6 @@ rep	stosw			#800h-866h clear
 	mov	-6(%bp),%ax
 	add	$2,%ax
 	mov	%ax,%es:(0x1a)	#new IP
-	mov	%ax,%bx
-	sub	$7,%bx
-	mov	$0x050f,%ax		#loadall
-	mov	%ax,%cs:(%bx)
 
 	mov	%ds,%es:(0x1e)	#not realaddress
 	mov	%ss,%es:(0x20)	#
@@ -259,10 +255,8 @@ rep	stosw			#800h-866h clear
 	mov	%dx,%es:(0x30)
 	mov	%cx,%es:(0x32)
 	mov	%ax,%es:(0x34)
-#db 0xf,db 0x5			#ia16-elf-as "no such instruction" error!!
-#syscall(0xf,0x5)			#ia16-elf-as "not supported on i8086" error
-	nop			#loadall
-	nop
+.word 0x050f				#loadall for direct binary
+
 modechange2:
 	call	retaddr	#3byte
 #this address is stacked [bp-6]
@@ -273,9 +267,11 @@ retaddr:
 rep	movsw				#word transfer
 	call	modechange2		#retf "cs back to 20bit segment" this is for 32bit CPU noneed?
 	pop	%ax
-	mov	%ax,%ss
+	mov	%ax,%ss		#SS 20bit segment
 	push	%ds
 	pop	%ds			#used 24bit DS back to 20bit segment
+	push	%es
+	pop	%es			#used 24bit ES back to 20bit segment
 	sti
 
 	pop	%bp
--- a/elks/arch/i86/lib/fmemory.S
+++ b/elks/arch/i86/lib/fmemory.S
@@ -20,6 +20,8 @@
 	.global fmemcmpb
 	.global fmemcmpw
 
+	.global loadall_block_move
+
 // void fmemcpyb (void * dst_off, seg_t dst_seg, void * src_off, seg_t src_seg,
 //		size_t count)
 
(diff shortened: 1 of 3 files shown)


drachen6jp:
I have changed very important 3 define address in config.h both ibm-pc and pc-98.
and order was also changed.
this change  was not tested enough.

my test is Pcem (286) and nekoproject2(286) and PC-9801DX2 at little time.

thank you.

and thanks to @tyama501 
xms.c was changed again. thank you and sorry.

elks/arch/i86/lib/fmemory.S
```diff
@@ -159,3 +161,126 @@ fmemcmpw_exit:
...
+	mov	%es,%es:(0x24)	#set 20bit segment
+	mov	%di,%es:(0x26)
+	mov	%si,%es:(0x28)
+	mov	%bp,%es:(0x2a)
+	mov	%sp,%es:(0x2c)
+	mov	%bx,%es:(0x2e)
+	mov	%dx,%es:(0x30)
+	mov	%cx,%es:(0x32)
+	mov	%ax,%es:(0x34)
+#db 0xf,db 0x5			#ia16-elf-as "no such instruction" error!!
```
asiekierka: `.arch i286` is required to support instructions added by the 286. `i8086` only allows compiling code that works correctly on an 8086-class CPU.

tyama501:
Hi @drachen6jp ,

The latest codes above worked on the PC-9801RX, intel286.
I could boot it with the LOADALL and hma both enabled.
Great! Thank you.

IMG_20250429_175950

IMG_20250429_180004

SSD ramdisk seems still not working in my system, but that may be my EMS board issue.
I will create another ticket if needed.

ghaerr:
@drachen6jp thank you for your changes. 

I would still like to see if .byte works instead of using self-modifying code, can you try that? Unfortunately, I am currently unable to test any LOADALL instruction, but I may try to setup pcem since you have tested with that.

Also, I still don't understand how DS is saved across entire block_move function call. I think PUSH DS/POP DS will reset segment base and limit cache register but this doesn't return DS to original value, correct? Or do I not understand the routine? Ia16-elf-gcc requires DS and ES to be fully (along with SI, DI and BP) saved across function calls. I may be incorrect but wanted to get your verification of how the routine works to save DS across full function.

Is source code available for custom build of QEMU for PC-98 somewhere? That would be very interesting if there is a QEMU version that will emulate PC-98 and LOADALL.

Thank you!

drachen6jp:
@ghaerr thank you for your reply.
It is obvious that direct binary value (0x0f 0x05) specified code is better than self-rewriting code, so that is very welcome.
@asiekierka
I will test your advise.(.arch i286)thank you.

Now, of the values that LOADALL receives
from 84eh is the actual address calculation source.
The 8-byte number from 81eh is supposed to be the same as a mere label.

For example
When stosw is executed
The contents of the ax register are put into [es:di].
When the real address is obtained from es:di, the contents of the es register is not used (as in protected mode?). I believe that when the real address is obtained from es:di, it is calculated by getting a numerical value from the ES descriptor.

At the end of the loadall code
push ds(I get just a label (original DS) that I had assigned in line 250)
pop ds (get a label and write it back to the actual calculation, including the DS descriptor)
in succession to erase the DSDescriptor used by rep movsw (this is what I wanted to do with the previous call to enable_a20gate).
I expect that the DS for the 20-bit segment at the time of the call will be reinserted into the DSdescriptor again.

For a quick check, delete lines 250-253 that I added to fmemory.s, or delete the combination of push ss on line 236 and pop ax on line 275 and mov %ax,%ss on line 276, which might help me understand, so I'll try it here.

but it is hard to test without a real 286.

“custom build of QEMU for PC-98” This is only in my HDD. It is not open to the public because it is full of debugging codes. Moreover, the BIOSROM set of a specific model is necessary. Sorry.

There is another work of the author who wrote qemu/9821. This one also requires the BIOSROM set of the specific model.
http://takeda-toshiya.my.coocan.jp/

Thank you!

Translated with DeepL.com (free version)

ghaerr:
Hello @drachen6jp,

> At the end of the loadall code
push ds(I get just a label (original DS) that I had assigned in line 250)
pop ds (get a label and write it back to the actual calculation, including the DS descriptor)

Ok, I think I see now, thank you. Since original DS is set in line 250, after LOADALL, DS will be same value.
The DS base and limit will be set to desired values by code in lines 190-195.
Then, after REP MOVSW, PUSH DS/POP DS will reset base and limit cache, but keep same DS.

So I think no change is needed after all, now that I understand. Thank you, very amazing code. :)

> For a quick check, delete lines 250-253 that I added to fmemory.s, or delete the combination of push ss on line 236 and pop ax on line 275 and mov %ax,%ss on line 276,

I think those lines are fine and OK, 250-253 are needed in order to keep the values of segment registers the same and then reset SS and then DS are reset to 20-bit in 275-278. So all looks good.

ES is finally reset after STI and POP ES. I wonder whether difference between DS being restored before STI and ES after STI, but probably OK since any interrupt handler would likely reload ES before using it.

Thus I think all code in loadall_block_move is OK now that I understand it, except it would be nice to see if you can use .byte to eliminate self-modifying code. Also try xms.c change to see what you think, and after we should be ready for commit.

Thank you!

drachen6jp:
.word can be used.
and I wrote anothoer  couple of push es pop es before sti.

PC-286VE(PC-98 compatible 286 12MHz 4Mbyte) & PCem are tested.
both run LOADALL good.
thank you.

ghaerr:
Hello @drachen6jp,

> .word can be used.
and I wrote anothoer couple of push es pop es before sti.

Ok, that looks good, thank you. I think I now fully understand how the whole routine works.

> PC-286VE(PC-98 compatible 286 12MHz 4Mbyte) & PCem are tested.
both run LOADALL good.

Since everything is working on real hardware and PCem, I think I should commit this now, even though I want to make other changes. I can make other changes in xms.c and other modifications to simplify ELKS INT 15 handling afterwards. I will explain that in next PR.

I do not have PCem running on my system yet. Are you running the version on Github: https://github.com/sarah-walker-pcem/pcem or a different version?

Do you have a copy of the PC-98 ROM that is required for PCem to run the 286 emulation for PC-98? I see a [collection of ROMs here](https://github.com/BaRRaKudaRain/PCem-ROMs) but don't see PC-98 ROM anywhere.

If you could post PC-98 ROM and config file, that would help so that I can emulate LOADALL solution when developing changes required for better ELKS throughput using modified INT 15 routine. 

If OK with you, I will commit now, then make other changes. I hope you can test them for me until I get PCem running. Does that sound OK to you?

Thank you!

tyama501:
Hi @ghaerr 

I think he cannot post ROMs
since it might be still illegal in Japan law.

Thanks.

drachen6jp:
>Now, of the values that LOADALL receives
from 84eh is the actual address calculation source.
The 8-byte number from 81eh is supposed to be the same as a mere label.

oops.not 84eh but 836h.

I do pcem on gentoo linux with  emerge system.and I got samsung 286 PC ROM from www.(actually this is illegal action in Japan now.)
Japanese law still refuse  right of fair use. The rights of author are protected in law.

as @tyama501 said I cannot post PC-98 ROM set to www. sorry.

ghaerr:
Hello @drachen6jp and @tyama501,

> I think he cannot post ROMs
> this is illegal action in Japan now.)

That is fine, it just means I cannot use PCem to emulate PC-98 for now. I will have to use another emulator. PC-98 ROMs are not required to emulate PC-98 system, depending on type of emulator. 

I will look at DosBox-X, which currently runs PC-98 version but LOADALL is undefined opcode. Perhaps a newer version implements LOADALL, I will check on that. [Here is a link](https://github.com/sarah-walker-pcem/pcem/blob/bd1283b91282d522617ac6d29eb0f271ded83ffc/includes/private/cpu/x86_ops_misc.h#L827) to PCem source code to emulate LOADALL, it is very informative. This shows how another emulator might be enhanced to support this opcode.

So I will wait until further testing of config.h DEF_OPTSEG not changed before commit. I left previous comment which now I think does not require changing low memory in order to implement LOADALL block_move on ELKS.

Thank you!

ghaerr:
Thank you @drachen6jp for all your work and testing on this! I will commit this now. This very interesting work of yours replacing INT 1F/15 block move with LOADALL for 80286 is a very nice enhancement to ELKS, and I have learned a lot in the process. It is also good news that we do not have to rearrange low memory in order to work around LOADALL requiring 800h-865h physical addresses.

This PR completes the discussion originally started in https://github.com/ghaerr/elks/pull/2320#issuecomment-2831558669.

I will be working on a followup PR that performs the following additions:
- Documents LOADALL memory area format in source file(s)
- Allows for selecting original INT 1F/15 BIOS calls even when LOADALL and 286 available; this is to allow regression testing of the following features:
- Implement xms_fmemset using LOADALL, since currently a slow workaround is required, used when a zeroed buffer is required in XMS memory, in elks/fs/buffer.c::zero_buffer()
- Implement int15_fmemcpyb instead of int15_fmemcpyw that accepts byte rather than word count; this will be used within xms_fmemcpyb where lots of extra code is required to work around word-copy-only INT 15/1F block moves. (See xms.c::xms_fmemcpyb()). An odd-number byte copy is required when a filesystem read/write is performed from a buffer that happens to be in XMS memory, rather than L1 cache.

I [opened an issue](https://github.com/joncampbell123/dosbox-x/issues/1184) in DosBox-X to see what it might take to get that emulator to emulate the LOADALL opcode. I will still work to try to find an emulator that can emulate PC-98 and LOADALL on macOS.

Thank you!

ghaerr:
Hello @drachen6jp,

I managed to compile PCem for my macOS. I can emulate AMI BIOS and have some IBM ROMs. Do you know how to setup boot from floppy? I see in configs/test.cfg my ELKS HD image:
```
hdc_fn = /Users/greg/net/elks-gh/image/hd32-minix.img
```
But I cannot get HD to boot due to IDE problems, Do you know how to specify floppy image in configuration file?

Getting this to work will allow me to test LOADALL on 286, which is the same for PC-98.

I have attached my configuration files: 
[PCem_configs.zip](https://github.com/user-attachments/files/20026901/PCem_configs.zip)

Thank you!

PR #2325, opened 2025-05-02
Paint: Flood Fill improvement

Vutshi:
Add FrontFill algorithm (on mouse click)
Add benchmark_mode flag
Fix canvas bounds check in R_LineFloodFill (keep it on 'f' button)

### DIFF
--- a/elkscmd/gui/app.c
+++ b/elkscmd/gui/app.c
@@ -10,6 +10,8 @@
 #include "vgalib.h"
 #include "mouse.h"
 
+#define BENCHMARK_PAINT 0  // Set to 1 to enable benchmark mode (draw and exit immediately)
+
 // --------------------------------------------
 // Definition of Globals
 // --------------------------------------------
@@ -147,7 +149,8 @@ void A_GameLoop(void)
             case mode_Fill:
                 if(mx <= CANVAS_WIDTH){
                     hidecursor();
-                    R_LineFloodFill(omx, omy, current_color, readpixel(mx, my));
+                    int maxCap = R_FrontFill(omx, omy, current_color, readpixel(mx, my));
+                    __dprintf("Peak stack size: %d\n", maxCap);
                     showcursor();
                     current_state = state_Idle;
                 }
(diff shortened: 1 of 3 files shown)


Vutshi:
The front_fill function performs a flood fill on a 2D canvas, starting from an initial point (x0, y0), replacing all connected pixels of a target color with a new_color. It uses a scanline approach, processing the image row by row by tracking horizontal segments of pixels.

S.V. Burtsev, Ye.P. Kuzmin,
_An efficient flood-filling algorithm_,
[Computers & Graphics, Volume 17, 5, 549-561 (1993)](https://doi.org/10.1016/0097-8493(93)90006-U)

Key Components:

1. Data Structures:

- Segment: Represents a horizontal run of pixels with left (xl) and right (xr) bounds.
- SegList: Groups segments on a specific row (y) with a direction (dir: +1 for down, -1 for up) and a list of segments (s).

2. Helper Functions:

- expand and expand_right: Finds the maximal horizontal run of target pixels around a point (x, y).
- fill: Paints all pixels in a SegList’s segments with new_color.
- link: Generates a new SegList for the next row (E.y + E.dir), finding segments vertically adjacent to those in E.s.
- diff: Computes segments in Ep not overlapping with Ed’s segments (expanded by 1 pixel on both sides), identifying new areas to explore in the opposite direction.
- merge: Combines two SegLists on the same row and direction into one, efficiently merging sorted, non-overlapping segment lists.

3. Main Algorithm (front_fill):

- Initialization: Starts by expanding around (x0, y0) to form an initial segment, creating E_plus (downward) and E_minus (upward) SegLists, and filling the initial segment.
- Stacks: Uses two stacks, S_plus and S_minus, assigned to active (Sa) and passive (Sp) roles, swapping them to balance workload.
- Loop:
      - Pops a SegList (E) from Sa.
      - Links to the next row’s segments (Ep), fills them, and computes the difference (Em) as new frontiers in the opposite direction.
      - Merges Ep into Sa if on the same row as the top of Sa, otherwise pushes it.
      - Pushes Em to Sp.
      - Swaps Sa and Sp if Sp has more segments, balancing processing.

- Return: The maximum capacity of Sa encountered, this is a temporary diagnostic

ghaerr:
Nice catch on CANVAS_HEIGHT in R_LineFloodFill! I'm surprised that bug made it this long :)

Pretty amazing work and code on the new fill algorithm - wow! I'm traveling and haven't had time to review it, but look forward to understanding it. I will also be very surprised if that code makes it through the C86 compiler, but that's ok if it does not.

Thank you!

PR #2326, opened 2025-05-03
[kernel] Document and small enhancements to new LOADALL code

ghaerr:
This PR documents the fields and use of @drachen6jp's newly committed LOADALL implementation for the 80286 processor. Also, some small changes are made to the function based on previous discussions in #2324 and https://github.com/Mellvik/TLVC/pull/165.

The following untested modifications were made to the original submission (will be tested as soon as PCem for IBM PC is running):
- Move CLI closer to execution of actual LOADALL. This may be able to be moved further still.
- Add CLD before REP STOS
- Use 92h for data segment access byte (removes preset Accessed bit); this will be changed in other kernel code for consistency between code and data access bytes and their Accessed bit, although it doesn't actually matter.
- Remove unneeded zeroing DI after clearing LOADALL memory.
- Change loadall_block_move to be function type void, as no error return is checked/possible.

Since the time this PR was coded, @Mellvik has come up with additional enhancements backed by real hardware testing - these will be added in an additional commit. Thanks @Mellvik!

### DIFF
--- a/elks/arch/i86/lib/fmemory.S
+++ b/elks/arch/i86/lib/fmemory.S
@@ -162,11 +162,83 @@ fmemcmpw_exit:
 	mov    %bx,%es
 	ret
 
+// void loadall_block_move(struct gdt_table *gdtp, size_t words)
+//
+// LOADALL Memory Format (0800h - 0865h)
+// Physical     Size in
+// Address	    Words   Associated CPU Register
+// 800          3       unused
+// 806          1       MSW
+// 808          7       unused
+// 816          1       TR (Task Register)
+// 818          1       Flag Word
+// 81A          1       IP
+// 81C          1       LDT (Local Descriptor Table)
+// 81E          1       DS
+// 820          1       SS
+// 822          1       CS
+// 824          1       ES
+// 826          1       DI
+// 828          1       SI
+// 82A          1       BP
+// 82C          1       SP
+// 82E          1       BX
+// 830          1       DX
+// 832          1       CX
+// 834          1       AX
+// 836          3       ES Descriptor Cache
+//  	836         1       ES base low  15:0
+//  	838         1       ES base high 23:0 & access byte
+//  	83A         1       ES limit     15:0
+// 83C          3       CS Descriptor Cache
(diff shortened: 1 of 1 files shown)


ghaerr:
Second commit removes dynamic IP calculation in favor of fast immediate mode setting of IP by @Mellvik.

Also changes CLI to operate with interrupts disabled around the entire routine!

My original modification moved CLI closer to LOADALL, as most kernel code is protected from reentry by the \_irqit routine. However, I then realized that this block move function could be called at interrupt time from the directfd.c floppy disk driver via xms_fmemcpyw. This means that the kernel may be in the middle of a normal buffer XMS copy operation when the direct floppy driver gets an interrupt, in which case this routine, including the initial zeroing of the LOADALL buffer, would destroy the operation in progress. @Mellvik, you might want to play with that - I am betting that after lots of direct floppy I/O mixed with multitasked application filesystem buffer activity, data will get corrupted.

Registers SI/DI are saved in the function prologue, but not saved across LOADALL. AX, BX, CX and DX are now not saved across the function call (or LOADALL), as they are scratch using ia16-elf-gcc's ABI.

ghaerr:
Hello @drachen6jp,

> I managed to compile PCem for my macOS. I can emulate AMI BIOS and have some IBM ROMs. Do you know how to setup boot from floppy?

I was able to figure out how to mount a floppy on PCem: right click on the screen when running. This then writes the configuration file.

I am now able to emulate LOADALL on IBM PC! 

This also showed I introduced a one-line error in this PR, which causes boot failure. I forgot to save CX across LOADALL, since that is used as word count for REP MOVS. The following line needs to be uncommented in order to work in fmemory.S:
```
        mov     %cx,%es:(0x32) // <--- uncomment this line
```
I will submit a PR fixing this shortly. 

Thank you!

drachen6jp:
I am honored that you turn to me.

https://github.com/sarah-walker-pcem/pcem/blob/dev/docs/pcem.man.1

 also pcem --help in console.
 pcem --load_drive_a_image image/fd1440.img

and thanks to repair and to improve my code.
May the 4th be with you.

PR #2327, opened 2025-05-04
[kernel] XMS enhancements and fixes around new LOADALL block move

ghaerr:
Fixes LOADALL CX not saved in last PR https://github.com/ghaerr/elks/pull/2326#issuecomment-2849126837.

Adds several new enhancements / bug fixes as a result of our new LOADALL implementation:
- Disables interrupts at start of int15_fmemcpyw since xms_fmemcpyw, which calls it, is called at interrupt time by the direct floppy driver. Looking into getting rid of static GDT data structure, but too large to place on kernel stack as an array[8].
- Always reenables interrupts on return from BIOS 15/1F calls; previously only did on error return for some reason?
- Allow xms=int15 to force INT 15/1F even on 80286 systems. Otherwise, xms=on will use LOADALL for 286, and unreal mode for 386 CPUs.
- Rewrote loadall_block_move function prologue for speed, remove multiple push/pop of segment registers.
- Fixed auto-disabling XMS when INT 15/1F specified w/kernel HMA, since INT 15/1F disables A20 gate.
- Changed GDT access bytes to standard 0x92 for data segments for consistency throughout kernel and LOADALL code.
- Added internal XMS_LOADALL xms_enabled state when running LOADALL block move.

Tested on pcEM for IBM PC only. PC-98 should work fine but needs testing. Hopefully @drachen6jp and @tyama501 can test PC-98 on real hardware.

### DIFF
--- a/elks/arch/i86/lib/bios15-ibm.S
+++ b/elks/arch/i86/lib/bios15-ibm.S
@@ -41,8 +41,8 @@ block_move:
 
 	mov	$0x87,%ah	# BIOS Block Move
 	int	$0x15
-	jnc	1f
 	sti			# ensure we've got interrupts
+	jnc	1f
 	mov	$-1,%ax		# fail return AX < 0
 	jmp	2f
 1:	xor	%ax,%ax		# success return AX = 0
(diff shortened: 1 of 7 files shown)


drachen6jp:
boot ok with hma and XMS.
but don't run ramdisk command,both pc/at and pc98.

https://github.com/ghaerr/elks/pull/2329
afiter fixed cx problem.
but mount /dev/ssd could't success. thank you

ghaerr:
Thank you for testing. I'll take a look at ramdisk and /dev/ssd.

PR #2328, opened 2025-05-04
[kernel] Rename loadall_block_move to loadall_block_op, move to loadall.S

ghaerr:
Moves loadall_block_move to its own source file loadall.S; no code changes.
Speeds up loadall_block_move in xms.c by not clearing large GDT array only required by INT 15/1F.


ghaerr:
Renames loadall_block_move to loadall_block_op. Takes byte instead of word count and future operation code for block move versus block clear using same routine.

PR #2329, opened 2025-05-04
[kernel] Change int15_fmemcpy to accept byte rather than word count

ghaerr:
Next part of evolving previous INT 15/1F word-count routines to work using byte-counts with the LOADALL block move routine. This allows greater efficiency within the kernel and eventual removal of some special hacks required for INT 15.

PC-98 BIOS INT 1F does allow for byte count block moves, but since LOADALL now working, probably not worth updating.

In general, INT 15/1F support is only being kept for backwards compatibility and regression testing when LOADALL can't be used, for instance in some emulators that don't emulate the undocumented opcode.

Changes include:
- int15_fmemcpyw renamed to int15_fmemcpy and takes byte instead of word count.
- block_move renamed to bios_block_movew and made void function.

Tested on PCem for IBM PC platform only.

### DIFF
--- a/elks/arch/i86/lib/bios15-ibm.S
+++ b/elks/arch/i86/lib/bios15-ibm.S
@@ -2,7 +2,7 @@
 # BIOS INT 15H Block Move and A20 enable for IBM PC
 # Nov 2021 Greg Haerr
 #
-# int block_move(struct gdt_table *gdtp, size_t words)
+# int bios_block_movew(struct gdt_table *gdtp, size_t words)
 #
 #	Move words to/from anywhere in physical memory using BIOS INT 15h AH=87h.
 #	Normally used to move data to/from extended memory (> 1M) because
@@ -18,17 +18,17 @@
 	.code16
 	.text
 
-	.global	block_move
+	.global	bios_block_movew
 	.global	enable_a20_gate
 
 # IBM PC A20 gate functions shared with setup.S
 #include "a20-ibm.inc"
 
 #
-# int block_move(struct gdt_table *gdtp, size_t words)
+# int bios_block_movew(struct gdt_table *gdtp, size_t words)
 # Uses BIOS INT 15h AH=87h Block Move
 # NOTE: BIOS disables A20 during call, won't work with HMA kernel
-block_move:
+bios_block_movew:
 	push	%es
 	push	%si
 	push	%bp
--- a/elks/arch/i86/lib/bios15-ibm.S
+++ b/elks/arch/i86/lib/bios15-ibm.S
@@ -2,17 +2,16 @@
 # BIOS INT 15H Block Move and A20 enable for IBM PC
 # Nov 2021 Greg Haerr
 #
-# int bios_block_movew(struct gdt_table *gdtp, size_t words)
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @drachen6jp, @tyama501 and @Mellvik,

> boot ok with hma and XMS.
but don't run ramdisk command,both pc/at and pc98.
https://github.com/ghaerr/elks/pull/2329
afiter fixed cx problem.
but mount /dev/ssd could't success. thank you

I can make ramdisk and /dev/ssd run with very small (40k) ramdisk using this script, which I call 'zz':
```
set -x
set -e
ramdisk /dev/ssd make 40
mkfs /dev/ssd 40
sync
fsck -lvf /dev/ssd
mount /dev/ssd /mnt
cp /bin/ls /mnt
ls -l /mnt
/mnt/ls -l
sync
df /dev/ssd
umount /dev/ssd
fsck -lvf /dev/ssd
```
When I using xms=on (LOADALL on 286) and PCem, even reverting back to #2324 (your first commit of LOADALL), but also including this PR, funny things happen:
```
cat zz
```
Displays the contents of the script file, but hangs on last line. ^C will then return to shell prompt.
When executing using ./zz, script runs and ramdisk works but system crash. This is the same on #2324 as well as now (#2329).

Can you try it, or show where /dev/ssd and ramdisk are failing, as well as "cat zz" whether it hangs on last line? I don't know if this is a problem in PCem emulator or real hardware. When using xms=int15, everything works OK.

I am currently unable to test with larger ramdisk because PCem is limiting the amount of XMS memory right now. I will look at that and fix later after this issue.

Very strange!

Thank you!

drachen6jp:
I could do ramdisk command with 40k bytes OK.
but 2000k bytes NG.
GqK4PYVbsAA6575


and pcem test.cfg
mem_size = 1536
you can modify this value for memory test.

with unreal mode test, 3000k byte OK.

good day!

ghaerr:
> I could do ramdisk command with 40k bytes OK.
but 2000k bytes NG.

Can you show the boot screen, I would like to see both the "xms: 2048k" and "ssd: xxxK avail" messages to verify what the kernel thinks the XMS memory is. Does the problem occur with both HMA kernel and without?

> and pcem test.cfg
mem_size = 1536
you can modify this value for memory test.

Yes, I have tried that, but then the "[286] Award 286 Clone" BIOS gives a "CMOS memory not matching NVR" error or something like that. The largest I have been able to run is 1536K, which is 1M lower memory and 512K XMS. It is still usually good enough for testing, but perhaps there is LOADALL error with larger byte counts or addresses? I will double check the GDT setup in the loadall_block_op function (or int15_fmemcpy also) to see if the source/dest address calculation may be the problem.

ghaerr:
Also @drachen6jp, your screenshots show something is definitely wrong with `ramdisk`: it is supposed to display the size of the ramdisk created, instead is displays nothing. Here is what it should look like:
ramdisk

The ramdisk make is failing with no error message, that is why later `mkfs` fails: there is no /dev/ssd ramdisk, it is 0K in size.

Something is definitely strange going on when LOADALL is used, we have this problem, and we also have problem with `cat file` not working.

Can you try running xms=int15 on real PC-98 to see if that works OK?

PR #2330, opened 2025-05-05
[kernel] Fix application errors when running XMS LOADALL

ghaerr:
Fixes system stability and application problems seen when running certain programs when XMS LOADALL is enabled. In particular, this fixes the `cat` problem mentioned in https://github.com/ghaerr/elks/pull/2329#issuecomment-2850070989, and should also fix the `ramdisk` problem reported by @drachen6jp in https://github.com/ghaerr/elks/pull/2329#issuecomment-2850206403 and by @tyama501 previously.

It took a while to track this down, but what was happening is that even though the LOADALL buffer at physical address 0800-0865 is within DEF_OPTSEG (the 1K /bootopts buffer loaded at boot time), after kernel initialization the entire DEF_OPTSEG and REL_INITSEG (512 byte setup.S data area) were being released back to the kernel main memory manager. That memory was then allocated to program loading and execution on demand.

The freed memory usually totaled 1536 bytes. Then, when small programs such as cat (928 bytes) or ramdisk (1392 bytes) ran, they were sometimes placed in this 1536 byte segment. If/when the kernel did any XMS I/O during the period when the application was running, the LOADALL loadall_block_op routine would write to its fixed buffer, which is right in the middle of the running application! We were sometimes seeing crashes, but other times just strange results, like not exiting, since the exit() code just happened to be linked into the scribbled location(s). I'm very glad to have finally found this.

The `ramdisk` problem has not been tested since I can't yet emulate > 512K XMS on 80286 PCem, but given the reported result of not issuing an error or the size created, I think this is the same problem.

The fix is to "carve out" the LOADALL buffer's 66h bytes (70h when rounded to paragraph) from the released memory. Here's a screenshot showing two segments, 768 and 656 bytes at 0050 and 0087, just 112 bytes short of the originally released amount. So LOADALL is only costing 112 bytes of application RAM, and it's only carved out when the system is running XMS LOADALL:

Screen Shot 2025-05-05 at 9 52 25 PM

@Mellvik: I'm not sure whether your fork releases this memory like ELKS based on your comment https://github.com/Mellvik/TLVC/pull/165#issuecomment-2848688954, if so you may the same problem!

### DIFF
--- a/elks/include/linuxmt/config.h
+++ b/elks/include/linuxmt/config.h
@@ -175,6 +175,7 @@
 #ifdef CONFIG_ARCH_IBMPC
 #define OPTSEGSZ        0x400       /* max size of /bootopts file (1024 bytes max) */
 #define DEF_OPTSEG      0x50        /* 0x400 bytes boot options at lowest usable ram */
+#define LOADALL_SEG     0x80        /* LOADALL buffer from 0x800-0x865 on 80286 CPU */
 #define REL_INITSEG     0x90        /* 0x200 bytes setup data */
 #define DMASEG          0xB0        /* start of floppy sector buffer */
 #define REL_SYSSEG      DMASEGEND   /* kernel code segment */
@@ -184,6 +185,7 @@
 #ifdef CONFIG_ARCH_PC98
 #define OPTSEGSZ        0x400       /* max size of /bootopts file (1024 bytes max) */
 #define DEF_OPTSEG      0x60        /* 0x400 bytes boot options at lowest usable ram */
+#define LOADALL_SEG     0x80        /* LOADALL buffer from 0x800-0x865 on 80286 CPU */
 #define REL_INITSEG     0xA0        /* 0x200 bytes setup data */
 #define DMASEG          0xC0        /* start of floppy sector buffer */
 #define REL_SYSSEG      DMASEGEND   /* kernel code segment */
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -279,6 +279,12 @@ static void INITPROC do_init_task(void)
 #ifdef CONFIG_BOOTOPTS
     /* Release options parsing buffers and setup data seg */
     heap_add(&opts, sizeof(opts));
+#ifdef CONFIG_FS_XMS
+    if (xms_enabled == XMS_LOADALL) {
+        seg_add(DEF_OPTSEG, 0x80);  /* carve out LOADALL buf 0x800-0x865 from release! */
+        seg_add(0x87, DMASEG);
+    } else  /* fall through */
+#endif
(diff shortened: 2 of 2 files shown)


drachen6jp:
good.
thanks @ghaerr.

I've checked this improve with pcem & nekoproject2.
both can run "ramdisk /dev/ssd make 4000"

by the way,pcem with award 286, I could do with many memory system
image

my nvr ctrl alt help setting(.cfg mem_size = 8192) is
image

ghaerr:
Glad it's all working!

Can you post your Award .nvr file? PCem doesn't work well with ctrl/alt key combinations on MacOS so I can't change the memory value. 8M would be great. 

Thank you!

drachen6jp:
[award_286.zip](https://github.com/user-attachments/files/20053374/award_286.zip)

ghaerr:
> [award_286.zip](https://github.com/user-attachments/files/20053374/award_286.zip)

Thank you. Despite configuration file saying 8192K, that configuration only allows for 5120K XMS, as apparently Award BIOS requires 2048K minimum for EMS!

Even with running BIOS Setup, the system still says Equipment Configuration Error (but still works):

Screen Shot 2025-05-06 at 8 55 48 AM
Screen Shot 2025-05-06 at 8 52 18 AM

I am not able to make BIOS boot error message go away, but I think things still OK.

I am able to create 4000K ramdisk, so now I can test what you are doing also, thank you.

BTW, is "cpu_speed = 6" the emulated CPU speed? Can that be increased so boot is faster?

Mellvik:
> @Mellvik: I'm not sure whether your fork releases this memory like ELKS based on your comment https://github.com/Mellvik/TLVC/pull/165#issuecomment-2848688954, if so you may the same problem!

Thanks for the warning, @ghaerr.  It's different in TLVC where OPTSEG overlaps with the floppy bounce buffer, but it could still have been a problem. It was addressed from the start though, by conditionally moving OPTSEG and everythign else from 0x50 to 0x90 --- which brought up the problem we discussed yesterday (different thread) about OPTSEG being different in the boot block and the kernel, causing bootopts to not be found by `setup`. 

With low memory now being CONFIGuration dependent, this may become a more visible problem than before, so I decided to change the minix bootblock to save the OPTSEG value to SETUP_DATA. Then, the setup code can move  `bootopts` to where the kernel (and setup itself) expects it. Minimal code changes.

drachen6jp:
sorry for slow reply.
https://github.com/sarah-walker-pcem/pcem/blob/dev/src/cpu/cpu_tables.c
cpu=6 in .cfg means 286 speed 25Mhz on the table "CPU cpus_286[]"

another work ,meminfo XMS size. It is wonderful news.thank you!

PR #2332, opened 2025-05-06
[cmds] Add XMS total, used and free memory to meminfo display

ghaerr:
XMS memory usage is now displayed in meminfo:
meminfo


Mellvik:
Very nice @ghaerr - admirable simplicity.

PR #2333, opened 2025-05-06
Paint: hardware-accelerated flood fill

Vutshi:
Use VGA Read Mode 1 for hardware-accelerated color comparison to optimize flood fill operations. Add segment expansion functions that leverage VGA color compare features for improved performance with bit-level optimizations.

### DIFF
--- a/elkscmd/gui/graphics.c
+++ b/elkscmd/gui/graphics.c
@@ -154,7 +154,7 @@ void vga_cmp8_init(int target_color) {
     set_color_dont_care(0x0F);
 }
 
-unsigned short cmp8(int x, int y, int target_color) {
+unsigned short cmp8(int x, int y) {
     /* Calculate offset in video memory */
     unsigned int offset = (y<<6) + (y<<4) + (x >> 3);
     return asm_getbyte(offset);
--- a/elkscmd/gui/graphics.c
+++ b/elkscmd/gui/graphics.c
@@ -143,7 +143,7 @@ void fillrect(int x1, int y1, int x2, int y2, int c)
 }
 
 /* initialize VGA to Read Mode 1 & set compare/color mask */
-static void vga_cmp8_init(int target_color) {
+void vga_cmp8_init(int target_color) {
     /* Set Graphics Mode Register to read mode 1 (bit 3 = 1) */
     set_write_mode(8);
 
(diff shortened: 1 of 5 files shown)


ghaerr:
Very nice work using VGA hardware acceleration to greatly increase drawing speed! Your work is definitely showing how closer attention to the hardware can eliminate what we might have once thought were permanent bottlenecks.

I'm thinking perhaps this approach might also be able to be used to increase the speed of non-XOR cursor drawing, for certain filled cursor shapes. I haven't figured out how that might work yet, though.

elkscmd/gui/graphics.c
```diff
@@ -142,6 +142,24 @@ void fillrect(int x1, int y1, int x2, int y2, int c)
...
+    set_write_mode(8);
+
+    /* Set Color Compare Register to target color (4 bits) */
+    set_color_compare(target_color & 0x0F);
+
+    /* Set Color Don't Care Register to 0x0F (include all planes) */
+    set_color_dont_care(0x0F);
+}
+
+unsigned short cmp8(int x, int y, int target_color) {
```
ghaerr: Do you want to remove the unused 'target_color' as a parameter here? A slight speedup may result by not passing it when called also.

Vutshi: Right, I forgot about this one. Initially I was doing initialization inside cmp8 and it costs ~2 seconds in the benchmark. I’ll remove it.

Vutshi: cmp8 function is pretty silly at the moment maybe it is better to directly read the byte, offset_y calculation can be saved this way

ghaerr: > offset_y calculation can be saved this way

I was thinking along the same lines myself, but the y offset calc isn't that expensive. But you may be able to track both x and y in a single offset in the calling code. I see a bunch of x << 3 etc that might not be needed, etc.

Vutshi:
another `0.4` second saved: `0m21.690s`

elkscmd/gui/graphics.c
Vutshi: I’ll try this optimization this evening. By the way, you’re active at an unusual hour—are you traveling far from home?

ghaerr:
Looks good, I'll commit now.

I would say that if in the upper level you calculated a  premultiplied y offset for each line, and then tracked x via offset only, that would eliminate lots of x >> 3 and x << 3, and cmp8 would only have to add the x + y offsets together to get the byte.

> are you traveling far from home?

I've been in Paris for the last week, working on my French lol. Will be around for another week! :)

Vutshi:
> I've been in Paris for the last week, working on my French lol. Will be around for another week! :)

I see—have fun! The concentration of ELKS enthusiasts in Europe has become dangerously high, so I’ll be flying to China tomorrow to restore some balance 🙂.

Meanwhile, as you brush up on your French, I’ve got a fresh case of ELKS panic attack for you.

PR #2334, opened 2025-05-07
Paint: small flood fill optimization

Vutshi:
Replace cmp8 with a direct byte read optimizing offset calculation.
It saves 3 seconds in the Paint benchmark: `0m18.600s`.

### DIFF
--- a/elkscmd/gui/graphics.c
+++ b/elkscmd/gui/graphics.c
@@ -154,12 +154,6 @@ void vga_cmp8_init(int target_color) {
     set_color_dont_care(0x0F);
 }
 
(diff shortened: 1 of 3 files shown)


ghaerr:
Amazing this mod saves 3 seconds, but then again I guess when a large portion of the screen is being filled, small speedups become large speedups.

We need to track down the same speed issue(s) with `sl` sometime! All this is giving me hope there may be lots of other speedups possible by realizing that it's the code that gets called thousands of times that needs to be looked hard at and optimized. I'm hoping to get a kernel/application profiler running sometime soon, it's been on the back burner...

PR #2345, opened 2025-06-29
SOLO/86 platform support (including ATA-CF driver)

fhendrikx:
Changes to support the Solo/86 platform.

Changes include:
- Solo/86 Platform support
- ATA-CF driver (and the generic ATA module)
- CLI tools updated for Solo/86

Please note that BIOSHD_MAJOR definition is changed to ATA_MAJOR (but we leave in place an alias so things keep working).


ghaerr:
Hello @fhendrikx,

Very nicely done, thank you! 

I have just a very few questions/comments which I'll post next to the corresponding code, and we should be good to go for commit.

elks/arch/i86/drivers/block/ata-cf.c
```diff
@@ -0,0 +1,206 @@
...
+
+
+/**********************************************************************
+ * device initialisation
+ **********************************************************************/
+
+jiff_t ata_cf_timeout;
+
+char ata_cf_initialized;
+sector_t ata_cf_num_sects[NUM_DRIVES];     /* max # sectors on SSD device */
```
ghaerr: These three variable declarations aren't used in any other file, right? Can you please make them static? Thanks.

elks/arch/i86/drivers/block/ata-cf.c
```diff
@@ -0,0 +1,206 @@
...
+
+void INITPROC ata_cf_init(void)
+{
+    sector_t sectors;
+    int i;
+
+    // register device
+
+    if (register_blkdev(MAJOR_NR, DEVICE_NAME, &ata_cf_fops) == 0)
+        blk_dev[MAJOR_NR].request_fn = DEVICE_REQUEST;
```
ghaerr, line 61: If register_blkdev fails, it might be better to just return, rather than to continue into driver initialization with ata_reset etc. However, the only reason register_blkdev can fail is if the MAJOR_NR is already in use, which it would seem could only happen if BIOSHD was also configured, so this isn't really much of an issue.

elks/arch/i86/drivers/block/ata.c
```diff
@@ -0,0 +1,322 @@
...
+
+/**********************************************************************
+ * ATA support functions
+ **********************************************************************/
+
+/**
+ * Delays some amount of time (not exactly microseconds)
+ */
+static void delay_us(uint16_t n)
+{
```
ghaerr, line 42: Let's change this declaration to `unsigned short`, as well as on line 182, and then remove include <stdint.h> above.  At this time, I'd rather keep stdint.h out of kernel includes, and currently the use of unsigned short is quite prevalent within ELKS source and must be 16-bits wide.

Keeping stdint.h would introduce a new standard header file dependency into the kernel build, with the exception of some OS/2 headers files which are shared with external tools.

elks/arch/i86/lib/bios-none.S
```diff
@@ -0,0 +1,13 @@
...
+
+#include <linuxmt/config.h>
+
+	.arch	i8086, nojumps
+	.code16
+	.text
+	.global	call_bios
+
+call_bios:
+	ret
```
ghaerr: Do we have to have this file and the null call_bios function? My concern is other places in the kernel calling this function with SOLO/86 configured and not knowing the function isn't implemented. I can't tell exactly where this may be required to be linked in without compiling this PR prior to commit, so I thought I'd ask. 

I would hope that without the BIOSHD driver compiled in there would be no or very few calls to call_bios, but I'm guessing that's not the case. I would agree that it's probably not a great idea to ifdef out all the calls elsewhere, but we should probably discuss where these are just to make sure this makes sense.

elkscmd/nano-X/Makefile
```diff
@@ -33,6 +33,19 @@ DRIVERS += drivers/ramfont.o drivers/X6x13.o
...
+
+ifeq ($(CONFIG_HW_VGA), y)
+DRIVERS += drivers/scr_bios.o drivers/vgaplan4.o
+DRIVERS += drivers/romfont.o drivers/elksutilasm.o
+else
+DRIVERS += drivers/scr_cga.o drivers/vgaplan4_cga.o
+DRIVERS += drivers/ramfont.o drivers/X6x13.o
+endif
+endif
+
```
ghaerr: Is this version of Nano-X working with Solo/86 with these Makefile changes? We've recently deprecated the entire Nano-X source tree within ELKS in favor of having got the main Microwindows tree running well on ELKS, with lots of superior capabilities, including multi-window/multi-client support. So it's not really necessary or desirable to use this ancient version. However, not a big deal, since this subtree will likely be removed in the near future.

ghaerr:
The above posts conclude my comments, thank you.

fhendrikx:
Hi @ghaerr
The changes you requested have been made. I also found an edge case where I was getting IO errors when copying large from one CF card to the other, so have improved the robustness of the ATA handling by adding more status checks and delays.

ghaerr (approved):
The changes look good, thanks!

ghaerr:
I'm going to commit this now. I did look at the ATA driver itself in more detail and found potentially dangerous stack overflow issue, which I will comment on separately. However, it's likely OK for now so this can be addressed later.

Thank you very much for this very nice enhancement! Your ATA driver is quite clean and I'm thinking could possibly be used as the basis for a direct access HD driver not using any BIOS. I'd like to separate out the existing partition handling code from the BIOSHD driver so it could be used with this and an HD ATA driver more easily.

Out of curiosity, does the ATA spec deal with interrupt-complete notifications on read/write commands, or is that handled on a per-system basis? Since all ATA I/O is PIO, I'm really not sure how much advantage an interrupt-driven HD driver might bring on a single processor system with kernel buffering, but on hard drives there will always be rotational delays that wouldn't have to be waited on within the driver.

elks/arch/i86/drivers/block/ata.c
```diff
@@ -0,0 +1,359 @@
...
+
+
+/**
+ * initialise an ATA device
+ *
+ * drive : physical drive number (0 or 1)
+ */
+sector_t ata_init(unsigned int drive)
+{
+    unsigned short buffer[ATA_SECTOR_SIZE/2];
```
ghaerr, line 251: FYI: declaring any large buffers like this on the stack is usually strictly disallowed, since the kernel stack pointer is either within the 640-700 byte KSTACK_BYTES section of the currently running task structure, or, for interrupt handlers called when already executing kernel code or any functions called by them, on the fixed 512-byte ISTACK_BYTES interrupt stack.

For now, this is only working because ata_init is called only during kernel startup, on the idle task's stack, prior to any task switching, but not necessarily prior to any interrupt handler activity (e.g. direct floppy I/O (not BIOS I/O)).
Since the idle task stack is 640-700 bytes depending on CONFIG_ASYNCIO and this buffer is 512 bytes and Solo/86 isn't configured for direct floppy, it all works.

For an example of a workaround using a buffer allocated from the kernel heap, see heap_alloc() in idequery.c. That code happens to be performing the same Identify function in some old messy code dealing with IDE and CHS identification.

I'm OK with leaving this as is, with a warning comment, or you could add heap_alloc/heap_free in your next driver revision.

elks/arch/i86/drivers/block/ata.c
```diff
@@ -0,0 +1,359 @@
...
+ * read from an ATA device
+ *
+ * drive : physical drive number (0 or 1)
+ * sector: sector number
+ * buffer: __far pointer to buffer containing 512 bytes of space
+ */
+int ata_read(unsigned int drive, sector_t sector, char *buf, ramdesc_t seg)
+{
+    unsigned char __far *buffer = _MK_FP((seg_t)seg, (unsigned)buf);
+    unsigned char status;
```
ghaerr, line 279: FYI: This code is also OK, but would fail in the case that CONFIG_XMS were enabled. For Solo/86, no issue since that won't happen, but in the general case, the ramdesc_t type becomes a 32-bit linear physical address and can't be used with \_MK_FP; instead the xms_fmemcpy functions are used. For PIO read/write, a new set of wrapper functions would be required, which gets a bit complicated depending on unreal/loadall/int15 modes etc. All this would be dealt with later if this driver were turned into a full ATA HD device driver for PCs.

ghaerr:
I also just noticed that [the CI build failed](https://github.com/ghaerr/elks/actions/runs/15978003032/job/45065383661); I had CI turned off on pushes (but not PRs) so we didn't see this earlier. I'll fix this in a commit shortly.

fhendrikx:
> Out of curiosity, does the ATA spec deal with interrupt-complete notifications on read/write commands, or is that handled on a per-system basis? Since all ATA I/O is PIO, I'm really not sure how much advantage an interrupt-driven HD driver might bring on a single processor system with kernel buffering, but on hard drives there will always be rotational delays that wouldn't have to be waited on within the driver.

Yes, there are definitely interrupt driven notifications as part of the spec. You could also use DMA to drive the transfers instead of PIO.

elks/arch/i86/drivers/block/ata.c
fhendrikx: I'll do another PR in the coming few days/weeks to fix this issue. TBH I didn't consider it all... too many years spent on projects where heap size wasn't a real issue. :)

elks/arch/i86/drivers/block/ata.c
fhendrikx: Sounds like a plan.

fhendrikx:
> I also just noticed that [the CI build failed](https://github.com/ghaerr/elks/actions/runs/15978003032/job/45065383661); I had CI turned off on pushes (but not PRs) so we didn't see this earlier. I'll fix this in a commit shortly.

Thanks.. sorry, wasn't aware of that.. my builds here at home have been good.

fhendrikx:
> I'm going to commit this now. I did look at the ATA driver itself in more detail and found potentially dangerous stack overflow issue, which I will comment on separately. However, it's likely OK for now so this can be addressed later.

OK, thanks

> Thank you very much for this very nice enhancement! Your ATA driver is quite clean and I'm thinking could possibly be used as the basis for a direct access HD driver not using any BIOS. I'd like to separate out the existing partition handling code from the BIOSHD driver so it could be used with this and an HD ATA driver more easily.

Thank you for the support. Let me know if you want any help; happy to help.

ghaerr:
> I also just noticed that [the CI build failed](https://github.com/ghaerr/elks/actions/runs/15978003032/job/45065383661)

Fixed in https://github.com/ghaerr/elks/commit/22dcd692bc921b322a2bbd90750550d9d146c17f. The problem was that CONFIG_BLK_DEV_ATA_CF defaulted to 'y' in config.in, which won't work when the BIOSHD driver is configured with either CONFIG_BLK_DEV_BFD or CONFIG_BLK_DEV_BHD. This is because they share the same major device number, but have separate I/O request routines. 

Thus for now, the ATA CF driver isn't easily tested outside of Solo/86 unless BIOSHD is turned off and the direct floppy driver is on. I'll think about how to best rectify this situation so that @toncho11 can play with the ATA CF driver on the Amstrad.

fhendrikx:
> Thus for now, the ATA CF driver isn't easily tested outside of Solo/86 unless BIOSHD is turned off and the direct floppy driver is on. I'll think about how to best rectify this situation so that @toncho11 can play with the ATA CF driver on the Amstrad.

Sounds good.

BTW Will have a PR for the heap alloc stuff to you later today.

PR #2346, opened 2025-07-01
ATA: heap_alloc update

fhendrikx:
As requested, move away from using the stack to hold the buffer and allocate a buffer on the heap.


ghaerr:
Thanks @fhendrikx!

PR #2349, opened 2025-07-06
Added support for different base IO port address for ATA

toncho11:
XT-IDE usually uses 0x300 instead of the standard 0x1F0.

### DIFF
--- a/elks/include/arch/ata.h
+++ b/elks/include/arch/ata.h
@@ -5,16 +5,18 @@
 
 #ifdef CONFIG_ARCH_IBMPC
 
-#define ATA_PORT_DATA   0x1F0
-#define ATA_PORT_ERR    0x1F1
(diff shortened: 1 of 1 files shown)


ghaerr:
I thought @fhendrikx said he already made this change...  Can you or @fhendrikx use ATA_BASE_PORT as previously suggested, so as to match other naming conventions already in use?

fhendrikx:
> I thought @fhendrikx said he already made this change... Can you or @fhendrikx use ATA_BASE_PORT as previously suggested, so as to match other naming conventions already in use?

Yep, I've already got that on my branch, of course, this has not yet pushed.

ghaerr:
> I've already got that on my branch, of course, this has not yet pushed.

Ok, I'll close this now then, thanks.

PR #2350, opened 2025-07-07
ATA changes

fhendrikx:
- Move ATA detection to ata_cf_open function
- Update ATA sanity checks
- Fix outw() usage bug in 8bit write

### DIFF
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -300,11 +300,11 @@ void ata_reset(void)
     ata_delay();
 
 
-    // controller 8-bit transfer mode
+    // controller 8-bit transfer is enabled for 8086/80186 systems
 
     use_8bitmode = 0;
 
-    if (1)                  // TODO: Detect 8086 CPU
+    if (arch_cpu < 5)
     {
         use_8bitmode = ata_set8bitmode();
     }
@@ -337,29 +337,29 @@ sector_t ata_init(unsigned int drive)
         // ATA LBA sector total (MSB << 16, LSB)
         total = (sector_t)buffer[ATA_INFO_SECT_HI] << 16 | buffer[ATA_INFO_SECT_LO];
 
-        printk("cf%d: blocks=%7lu size=%7luK CHS=%2u,%2u,%2u ",
-            drive, total, total >> 1, buffer[ATA_INFO_CYLS], buffer[ATA_INFO_HEADS],
-            buffer[ATA_INFO_SPT]);
+        printk("cf%d: %7luK CHS %2u,%2u,%2u SZ %x ",
+            drive, total >> 1, buffer[ATA_INFO_CYLS], buffer[ATA_INFO_HEADS],
+            buffer[ATA_INFO_SPT], buffer[ATA_INFO_SECT_SZ]);
 
         // ATA version
         if ((buffer[ATA_INFO_VER_MAJ] != 0) && (buffer[ATA_INFO_VER_MAJ] != 0xFFFF))
         {
             // dump out version word (don't bother decoding it)
 
-            printk("version=%x ", buffer[ATA_INFO_VER_MAJ]);
+            printk("VER %x ", buffer[ATA_INFO_VER_MAJ]);
         }
 
         // Sanity check
         if ((buffer[ATA_INFO_CYLS] == 0 || buffer[ATA_INFO_CYLS] > 0x7F00) &&
-            (buffer[ATA_INFO_HEADS] == 0 || buffer[ATA_INFO_HEADS] > 0x7F00) &&
-            (buffer[ATA_INFO_SPT] == 0 || buffer[ATA_INFO_SPT] > 0x7F00))
+            (buffer[ATA_INFO_HEADS] == 0 || buffer[ATA_INFO_HEADS] > 16) &&
+            (buffer[ATA_INFO_SPT] == 0 || buffer[ATA_INFO_SPT] > 63))
         {
             printk("ATA drive not present");
             //total = 0;
         }
         else if (! (buffer[ATA_INFO_CAPS] & ATA_CAPS_LBA))      // ATA LBA support?
         {
-            printk("no LBA");
+            printk("LBA not detected");
             //total = 0;
         }
 
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -352,8 +352,8 @@ sector_t ata_init(unsigned int drive)
         }
 
         // Sanity check
-        if ((buffer[ATA_INFO_CYLS] == 0 || buffer[ATA_INFO_CYLS] > 0x7F00) &&
-            (buffer[ATA_INFO_HEADS] == 0 || buffer[ATA_INFO_HEADS] > 16) &&
+        if ((buffer[ATA_INFO_CYLS] == 0 || buffer[ATA_INFO_CYLS] > 0x7F00) ||
+            (buffer[ATA_INFO_HEADS] == 0 || buffer[ATA_INFO_HEADS] > 16) ||
             (buffer[ATA_INFO_SPT] == 0 || buffer[ATA_INFO_SPT] > 63))
         {
             printk("ATA drive not present");
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -33,6 +33,7 @@
 
 static char use_8bitmode;
 
+
 /**********************************************************************
  * ATA support functions
  **********************************************************************/
@@ -126,6 +127,7 @@ static unsigned char ata_select(unsigned int drive, unsigned int cmd, unsigned l
     return (status);
 }
 
+
 /**********************************************************************
  * ATA functions
  **********************************************************************/
@@ -141,6 +143,7 @@ static int ata_set8bitmode(void)
 
     outb(0x01, ATA_PORT_FEAT);
     outb(ATA_CMD_FEAT, ATA_PORT_CMD);
+
 
     // wait for drive to be not-busy
 
@@ -279,6 +282,9 @@ void ata_reset(void)
 {
     unsigned char select = 0xA0;
 
+
+    // controller reset
+
     outb(select, ATA_PORT_DRVH);
 
     ata_delay();
@@ -293,8 +299,15 @@ void ata_reset(void)
 
     ata_delay();
 
-    // always attempt 8-bit transfer mode for now for testing
-    use_8bitmode = ata_set8bitmode();
+
+    // controller 8-bit transfer mode
+
+    use_8bitmode = 0;
+
+    if (1)                  // TODO: Detect 8086 CPU
+    {
+        use_8bitmode = ata_set8bitmode();
+    }
 }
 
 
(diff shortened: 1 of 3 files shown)


fhendrikx:
I put blocks= back into the CF printk, as I needed this number for mkfat. I also added a version= thing to dump out the major version field. This needs decoding, but is useful to see how "late" a device might be.

The ATA base port is currently set to 0x300.

Finally, 8-bit mode is still enabled by default, but we should consider putting in a 8086 detection step.

ghaerr:
Initial review looks great, thank you!

> Finally, 8-bit mode is still enabled by default, but we should consider putting in a 8086 detection step.

I didn't realize that Solo/86 uses an 80286. You can use "if (arch_cpu < 5)" to enable 8-bit mode for 8086/80186 systems (80286 == 6, 80386 == 7).

So is this all working on Solo/86? I'll probably commit in the morning.

Where did you come up with the <= 0x7F00 values for CHS validity testing, have you ever seen heads > 16 or sectors > 63?

What about ATA_INFO_SECT_SZ, was that not 512 on your system?

fhendrikx:
> I didn't realize that Solo/86 uses an 80286. You can use "if (arch_cpu < 5)" to enable 8-bit mode for 8086/80186 systems (80286 == 6, 80386 == 7).

Super, will add that. 

> So is this all working on Solo/86? I'll probably commit in the morning.

Yep, no errors reported, looking good.

> Where did you come up with the <= 0x7F00 values for CHS validity testing, have you ever seen heads > 16 or sectors > 63?

I thought I had, but I could be wrong. I'll double check tomorrow. In any case, I trawled other kernels and that was the magic number. 

> What about ATA_INFO_SECT_SZ, was that not 512 on your system?

It was, but the field is deprecated. It doesn't seem to be a test that's used. 

I'll do a wrap up tomorrow to the PR. Will confirm we're all good to go.

fhendrikx:
Morning, 

- Re-checked INFO_HEADS and INFO_SPT, these are indeed always below 16 and 63, so those values have been used.
- Re-checked INFO_SECT_SZ, and I'm getting 0x200 (512), as expected, but also 0x240 (576). Very odd, but explains why no one is using it?
- I've updated printk as requested
- However, I've left in place the SZ and VER outputs, as these are useful to gather more information.
- CPU arch check added. Anything below 5 and we're into 8-bit mode.

I also did some timing tests to compare 8 vs 16-bit modes:

16-bit mode:
- mkfat -fat32 /dev/hda 3903480: 107 seconds 
- cp /bin/* /mnt/: 8 seconds

8-bit mode:
- mkfat -fat32 /dev/hda 3903480: 107 seconds 
- cp /bin/* /mnt/: 7 seconds

So, at least on this kernel, there is no real difference in performance. 

It would be interesting to compile ELKS for the 286, and see if that makes a difference.

elks/arch/i86/drivers/block/ata.c
```diff
@@ -323,27 +337,34 @@ sector_t ata_init(unsigned int drive)
...
-            printk(" ATA drive not present");
+            // dump out version word (don't bother decoding it)
+
+            printk("VER %x ", buffer[ATA_INFO_VER_MAJ]);
+        }
+
+        // Sanity check
+        if ((buffer[ATA_INFO_CYLS] == 0 || buffer[ATA_INFO_CYLS] > 0x7F00) &&
+            (buffer[ATA_INFO_HEADS] == 0 || buffer[ATA_INFO_HEADS] > 16) &&
+            (buffer[ATA_INFO_SPT] == 0 || buffer[ATA_INFO_SPT] > 63))
```
ghaerr, line 357: Shouldn't these sanity checks all be logical OR (||), not AND (&&)? Any one of invalid C/H/S and we're disallowing the device, right?

ghaerr:
Great job, thanks for all your testing and the enhanced code. With the exception of the sanity check issue above, looks good, let me know when you're ready for commit.

> Re-checked INFO_SECT_SZ, and I'm getting 0x200 (512), as expected, but also 0x240 (576). Very odd, but explains why no one is using it?

That is kind of crazy - was this the value that changed depending on 8- vs 16-bit transfer mode?!

> I also did some timing tests to compare 8 vs 16-bit modes:

Pretty interesting that 16-bit is even *slower*? Is that repeatable? I'm wondering why that would be... Still OK by me to keep 16-bit mode for non-8086 systems as you have it though.

> It would be interesting to compile ELKS for the 286, and see if that makes a difference.

The kernel isn't recompiled differently based on CPU type, we produce the same kernel for all systems at this point. Are you suggesting something else?

elks/arch/i86/drivers/block/ata.c
fhendrikx: Err, yep, sorry :)

fhendrikx:
> Great job, thanks for all your testing and the enhanced code. With the exception of the sanity check issue above, looks good, let me know when you're ready for commit.

All good to go.

> > Re-checked INFO_SECT_SZ, and I'm getting 0x200 (512), as expected, but also 0x240 (576). Very odd, but explains why no one is using it?
> 
> That is kind of crazy - was this the value that changed depending on 8- vs 16-bit transfer mode?!

Looking back at my screenshots, the ancient 16MB CF card was reporting SECT_SZ 0x2020 in 8-bit mode. I'm guessing that the field had other uses, or was vendor-specific. Hence, avoid using it.

... and yes, this card also reports CHS 490,2,32 in 16-bit mode, and CHS 512,32,12849 (!!) in 8-bit mode. That was why I changed it from 63 to 32512 (0x7F00)... that said the card also drops LBA support in 8-bit mode, so the kernel was ignoring it anyway.

> Pretty interesting that 16-bit is even _slower_? Is that repeatable? I'm wondering why that would be... Still OK by me to keep 16-bit mode for non-8086 systems as you have it though.

I didn't repeat the timing tests in 8-bit mode, but they were pretty consistent in 16-bit mode.
 
> The kernel isn't recompiled differently based on CPU type, we produce the same kernel for all systems at this point. Are you suggesting something else?

No, I just had visions of a .286 directive, like in some assemblers.  :)

ghaerr:
> and yes, this card also reports CHS 490,2,32 in 16-bit mode, and CHS 512,32,12849 (!!) in 8-bit mode. That was why I changed it from 63 to 32512 (0x7F00)... that said the card also drops LBA support in 8-bit mode, so the kernel was ignoring it anyway.

Thanks for the details - agreed we need to keep an eye on various CF cards, sounds like there's quite a variety out there. Glad we're showing CHS, SZ and VER. I don't know that much about CF tech, does the each chip contain the ATA controller within the CF card, or is the controller in the USB/other mounting device? 

> I just had visions of a .286 directive, like in some assemblers. :)

Yes, there kind of used to be such a thing... but the difference for most kernel operations was similar to your ATA CF 8- vs 16-bit testing - just not that much difference to worry about maintaining separate kernel versions for.

Committing this now, thank you!

ghaerr:
Geez, quick testing w/QEMU and latest version doesn't work anymore:
Screenshot 2025-07-07 at 4 35 32 PM

I'll continue to investigate... not sure why CHS values completely different on same ATA drive. Any thoughts? It didn't look like hardly anything was changed in this area? Have you seen anything like this?

fhendrikx:
> Thanks for the details - agreed we need to keep an eye on various CF cards, sounds like there's quite a variety out there. Glad we're showing CHS, SZ and VER. I don't know that much about CF tech, does the each chip contain the ATA controller within the CF card, or is the controller in the USB/other mounting device?

No, I understand there is a single ATA controller at 0x1F0 or 0x300 (or 0x40 for Solo/86), and then each CF card implements some part of the standard itself. 

> Yes, there kind of used to be such a thing... but the difference for most kernel operations was similar to your ATA CF 8- vs 16-bit testing - just not that much difference to worry about maintaining separate kernel versions for.

Agreed, doesn't seem to be a big issue.
 
> Committing this now, thank you!

👍

fhendrikx:
> I'll continue to investigate... not sure why CHS values completely different on same ATA drive. Any thoughts? It didn't look like hardly anything was changed in this area? Have you seen anything like this?

Urgh, The values look fine in the initial detection at the top, then get rubbish results in the open call.

No, I didn't see this behaviour, it was all very consistent.

fhendrikx:
Presumably we didn't get the rubbish second set of results previously, because there was no ATA detect call in open, only the initial detection at boot.

ghaerr:
Something extremely strange is going on w/QEMU, one can call ata_init(0) multiple times in ata_cf_init, but when called from ata_cf_open, it fails. So it's not as simple as just ata_identify failing on second call. Very strange. Continuing to debug.

ghaerr:
Found it: ATA_BASE_PORT was set to 0x300, not 0x1F0!

Not sure what to do about this yet, but thinking of adding dynamic handling so that it works on XT's using 8-bit mode and XT-IDE port 0x300 and 0x1F0 otherwise. Thoughts?

fhendrikx:
> Found it: ATA_BASE_PORT was set to 0x300, not 0x1F0!

ah, of course!

> Not sure what to do about this yet, but thinking of adding dynamic handling so that it works on XT's using 8-bit mode and XT-IDE port 0x300 and 0x1F0 otherwise. Thoughts?

We probably have a number of options for automation:
- Detect machine type, and assume it will be using XT-IDE (e.g. 8086 and less)
- Detect XT-IDE explicitly
- Detect ATA controller starting at 0x1F0, then 0x170, and then 0x300 (.. and then possibly, moving onto the other XT-IDE addresses: 0x320, 0x360, 0x380, 0x3A0, 0x3C0, 0x3E0).

I think something should work out nicely. I can have a poke around, see if I can get some kind of ATA controller detection working.

ghaerr:
> We probably have a number of options for automation:

Thanks for your ideas. I've basically chosen your first thought (use 8086 and less as meaning PC/XT) in https://github.com/ghaerr/elks/pull/2351, which will work for @toncho11's testing and remains simple.

Later we might want to detect XT-IDE is possible, but it might not be worth it at this point, lets see how/whether this runs on the Amstrad.

> see if I can get some kind of ATA controller detection working.

I'm interested to hear what you might come up with, but frankly remain worried about using anything remotely complicated given all the differing results we've already seen from ATA controllers on Solo/86 and QEMU! 

The whole driver needs more testing on lots of CF cards, and then perhaps a move to ATA hard drives. But since the ELKS BIOS already handles hard drives quite well, there's not too much use except for custom systems like your Solo/86. So we're probably good if PR #2351 works on @toncho11's Amstrad.

fhendrikx:
> Thanks for your ideas. I've basically chosen your first thought (use 8086 and less as meaning PC/XT) in #2351, which will work for @toncho11's testing and remains simple.

Agreed.
 
> Later we might want to detect XT-IDE is possible, but it might not be worth it at this point, lets see how/whether this runs on the Amstrad.

Agreed.

> I'm interested to hear what you might come up with, but frankly remain worried about using anything remotely complicated given all the differing results we've already seen from ATA controllers on Solo/86 and QEMU!

Yeah, I might have a go; will send you a PR if I can get something reliable working.
 
> The whole driver needs more testing on lots of CF cards, and then perhaps a move to ATA hard drives. But since the ELKS BIOS already handles hard drives quite well, there's not too much use except for custom systems like your Solo/86. So we're probably good if PR #2351 works on @toncho11's Amstrad.

:)

PR #2352, opened 2025-07-12
[kernel] Add XMS support to ATA driver, fix ATA drive recognition 

ghaerr:
The ATA-CF driver now supports compilation with CONFIG_FS_XMS enabled, in which case an extra copy is required from the ATA I/O port to a main memory buffer if the system buffer is in XMS. When XMS is disabled, no extra copy is performed.

XMS code only tested on QEMU; updated driver not tested on Solo/86.

Also fixes a CF card recognition problem seen in QEMU previously showing both CF cards present when not actually present. While rewriting the ATA driver function error return code to kernel standard, this problem went away. I think the issue was that the functions previously mixed returning a negative kernel error code, zero and ATA status register bits. The function return value numbers are now all kernel standard, returning negative numbers on errors, zero on success, and not returning status register bits.

It should be noted that block device drivers don't actually ever return any I/O error code to kernel upper layers or to the user, so the only usefulness in tracking I/O error numbers is for driver understanding and debugging.

Various repeated code throughout the driver was combined to reduce kernel code size (e.g. wait while busy, getting ATA status, etc).


toncho11:
Nice catch for the control port!

PR #2353, opened 2025-07-14
ATA driver: improved performance

fhendrikx:
Hi @ghaerr 

I disassembled the code in the ATA driver and found that for the 16-bit code, GCC was doing a lookup for each buffer[i] mention (hence, 2 in the 16-bit code).

So, I've switched to using an incremental pointer. The old code was 10 instructions, while this improved version is 5 instructions. The performance improvements are as follows:

## 8-bit original
106s (format FAT32; 4GB)
24s   (write 2Mb file)
11s    (read 2Mb file)

## 8-bit improved
102s
23s
10s

## 16-bit original
106s
24s
11s

## 16-bit improved
92s (thats 14 seconds saved)
22s
10s

I suspect the changes would be even more obvious if I had tested even larger files.

### DIFF
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -263,7 +263,7 @@ static int ata_identify(unsigned int drive, char *buf)
     {
         for (i = 0; i < ATA_SECTOR_SIZE; i++)
         {
-            buf[i] = inb(ATA_PORT_DATA);
+            *buf++ = inb(ATA_PORT_DATA);
         }
     }
     else
@@ -272,8 +272,8 @@ static int ata_identify(unsigned int drive, char *buf)
         {
             word = inw(ATA_PORT_DATA);
 
-            buf[i+0] = (unsigned char) (word & 0xFF);
-            buf[i+1] = (unsigned char) (word >> 8);
+            *buf++ = (unsigned char) (word & 0xFF);
+            *buf++ = (unsigned char) (word >> 8);
         }
     }
 
@@ -429,7 +429,7 @@ int ata_read(unsigned int drive, sector_t sector, char *buf, ramdesc_t seg)
     {
         for (i = 0; i < ATA_SECTOR_SIZE; i++)
         {
-            buffer[i] = inb(ATA_PORT_DATA);
+            *buffer++ = inb(ATA_PORT_DATA);
         }
     }
     else
@@ -438,8 +438,8 @@ int ata_read(unsigned int drive, sector_t sector, char *buf, ramdesc_t seg)
         {
             word = inw(ATA_PORT_DATA);
 
-            buffer[i+0] = (unsigned char) (word & 0xFF);
-            buffer[i+1] = (unsigned char) (word >> 8);
+            *buffer++ = (unsigned char) (word & 0xFF);
+            *buffer++ = (unsigned char) (word >> 8);
         }
     }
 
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @fhendrikx,

Wow, nice work finding and measuring this large bottleneck! I've been disassembling various routines as well, and with all the heavy optimizations already being performed by GCC within this driver I'm surprised it didn't automatically handle this one.

I had been thinking of using REP INS/OUT for fast I/O here - given your findings, I'd say it'd be worth trying for even further improvement. This can probably be done pretty easily in C with the asm extension, but given the new code which uses a far buffer, DS will have to be saved and reloaded first, somewhat complicating the asm code.

I have kept #2352 outstanding awaiting XTIDE test results, I'd like to commit it first before this PR. That will likely cause merge conflicts since the port I/O has been moved into generalized routines, so you may have to fix and resubmit. I'll commit #2352 today so you can more easily move this forward.

Thank you!

elks/arch/i86/drivers/block/ata.c
```diff
@@ -486,14 +486,14 @@ int ata_write(unsigned int drive, sector_t sector, char *buf, ramdesc_t seg)
...
-            outb(buffer[i], ATA_PORT_DATA);
+            outb(*buffer++, ATA_PORT_DATA);
         }
     }
     else
     {
         for (i = 0; i < ATA_SECTOR_SIZE; i+=2)
         {
-            word = buffer[i+0] | buffer[i+1] << 8;
+            word = *buffer++ | *buffer++ << 8;
```
ghaerr, line 496: While this seems to work, I can't recommend using two auto increments in the same expression, given the compiler's freedom to reorder expressions. I don't know exactly what the C standard has to say about this.

If you use an auto variable for the first 8-bits, the compiler will usually optimize the extra store out. Another option is to directly access 16-bits from the buffer each loop iteration.

fhendrikx:
> Wow, nice work finding and measuring this large bottleneck! I've been disassembling various routines as well, and with all the heavy optimizations already being performed by GCC within this driver I'm surprised it didn't automatically handle this one.

Me too.
 
> I had been thinking of using REP INS/OUT for fast I/O here - given your findings, I'd say it'd be worth trying for even further improvement. This can probably be done pretty easily in C with the asm extension, but given the new code which uses a far buffer, DS will have to be saved and reloaded first, somewhat complicating the asm code.

Yep, agreed. I thought I would try the simple optimisations first.

> I have kept #2352 outstanding awaiting XTIDE test results, I'd like to commit it first before this PR. That will likely cause merge conflicts since the port I/O has been moved into generalized routines, so you may have to fix and resubmit. I'll commit #2352 today so you can more easily move this forward.

Of course. Will adjust it as the changes come through.

elks/arch/i86/drivers/block/ata.c
fhendrikx: Yeah, I did have my doubts when I typed it... but forgot about it after. Good catch.

I'll change that when I update the PR. :)

toncho11:
In general maybe the current performance of the direct CF driver is close to the BIOS one (I have no tested). The BIOS routines are optimized in assembler I suppose.
Having our own driver means that the responsibility for the performance now shifts to this driver (and only to it).
From what I know - less function calls, inlines and some re-writes in assembly will help squeeze the max performance.
But for now we also need to appreciate the fact that it works :).

A fast driver will allow a fast swap file support in the future.

ghaerr:
Hi @fhendrikx,

I ended up having to rewrite the read/write I/O based on newly-learned XTIDE information in PR #2359, and that involved getting right in the middle of this PR's changes. As a result, I ended up adding your enhanced performance code to that PR as well. As such, I'm closing this now.

Thank you!

fhendrikx:
No worries, and thank you.

PR #2354, opened 2025-07-15
[kernel] Support dynamic port setup for ATA, XTIDE, XTCF and SOLO/86 in ATA CF driver

ghaerr:
Discussed in https://github.com/ghaerr/elks/pull/2351#issuecomment-3070623294 and https://github.com/ghaerr/elks/pull/2351#issuecomment-3071293816 as a result of the ATA driver still not working on @toncho11's Amstrad PC/XT, which has a LoTech XT-CF-lite v4 compatible card. It was determined that the ATA base and control register port addresses as well as the register offset port I/O address mappings are not the same between standard ATA, XTIDE and XTCF controllers. The Solo/86 acts like an XTCF controller but with yet another set of port base addresses.

This PR enhances the ATA CF driver to support dynamically adjusting the port setup for the following supported cards and platforms:
```
            ATA             XTIDE           XTCF                SOLO/86
BASE        0x1F0           0x300           0x300               0x40
BASE+reg    0x1F0+reg       0x300+reg       0x300+(reg<<1)      0x40+(reg<<1)

CTRL        BASE+0x200+reg  BASE+0x08+reg   BASE+0x10+(reg<<1)  BASE+0x10+(reg<<1)
CTRL+6      0x3F6           0x30E           0x31C               0x5C
DEVCTRL     BASE+0x206      BASE+0x0E       BASE+0x1C           BASE+0x1C
```
For the time being, an XTCF controller is assumed when running on a PC/XT, and ATA otherwise, unless configured for SOLO/86. XTIDE controllers should work, but are not yet auto-configured.

The busy-wait time has been decreased 10x for non-read/write commands, as a result of @toncho11's finding that the initial ATA probe took longer than a minute. The current version uses busy-looping; an update is planned to use a kernel real time timer for determining delays based on the XTCF and ATA specs. Also, read/write sector retries are not currently operational.

@toncho11, lets have @fhendrikx test this before you do. I'm fairly confident it should work on both systems, but let's wait to be sure to minimize wasted time. I have thoroughly checked the dynamic port addresses for all supported platforms.

### DIFF
--- a/elks/arch/i86/drivers/block/ata-cf.c
+++ b/elks/arch/i86/drivers/block/ata-cf.c
@@ -161,8 +161,10 @@ static void do_ata_cf_request(void)
                 debug_blk("cf%d: reading sector %lu\n", drive, start);
                 ret = ata_read(drive, start, buf, req->rq_seg);
             }
-            if (ret != 0)           /* I/O error */
+            if (ret != 0) {         /* I/O error */
+                printk("cf%d: I/O error %d\n", drive, ret);
                 break;
+            }
             start++;
             buf += ATA_SECTOR_SIZE;
         }
--- a/elks/arch/i86/drivers/block/ata-cf.c
+++ b/elks/arch/i86/drivers/block/ata-cf.c
@@ -162,7 +162,7 @@ static void do_ata_cf_request(void)
                 ret = ata_read(drive, start, buf, req->rq_seg);
             }
             if (ret != 0) {         /* I/O error */
-                printk("cf%d: I/O error %d\n", drive, ret);
+                printk("cf%d: I/O error %d cmd %d\n", drive, ret, req->rq_cmd);
                 break;
             }
             start++;
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -75,7 +75,7 @@ static unsigned char INB(int reg)
     return inb(BASE(reg));
 }
 
-/* output byte from translated register number */
+/* output byte to port from translated register number */
 /* FIXME: compiler bug if 'unsigned int byte' declared below. Not debugged yet. */
 static void OUTB(unsigned int byte, int reg)
 {
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -41,7 +41,7 @@
 #include <arch/io.h>
 #include <arch/ata.h>
 
-/* wait loop counts while busy waiting (FIXME use jiffies for accuracy) */
+/* wait loop counts while busy waiting (FIXME: use jiffies for accuracy) */
 #define SHORT_WAIT  500
 #define LONG_WAIT   5000        /* 14s wait required for some writes */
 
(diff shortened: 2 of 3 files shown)


fhendrikx:
I'll grab this and test it later tonight.

toncho11:
It is working!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
I could not mount a fat formatted CF card with fat16 made from Windows, but an ELKS one with mkfat yes.
I am able to mount and write files.

0x300h/31C

Would it be possible to boot from it?

ghaerr:
> It is working!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Wow, nice! Thanks for your faith in my work :)

> I could not mount a fat formatted CF card with fat16 made from Windows, but an ELKS one with mkfat yes.

[This article](https://minuszerodegrees.net/xtide/XT-IDE%20-%20Problems.htm) discusses common problems associated with getting CF cards working, and recommends that the formatting be done on the target system. Glad you got it working!

> Would it be possible to boot from it?

Yes - I think so. You would have to directly copy an HD image to it, for now, the one without partitions. See the article above for issues. I'm not sure it will work yet, as the system will be booting using the XTIDE BIOS, not ELKS, and also partitions are not supported. [EDIT: I'm not sure whether XTIDE supports booting from non-MBR (non-partitioned) drives.]

When/if the CF boot image comes up, it will be ELKS using the BIOSHD driver to access the root filesystem. By accessing /dev/cf0 you would then be accessing the very same hard drive so you must not mount it, but instead just dump the contents for testing using `hd /dev/cf0 | more`.

See my discussion [here](https://github.com/ghaerr/elks/pull/2351#issuecomment-3071158199) for more thoughts.

Now that this is working, I will move forward with testing a system that can be booted from CF, and then use the XTCF driver to access the root filesystem. This could be tricky to test, since both the machine BIOS, then ELKS boot, then ELKS kernel all  need to be configured correctly to work.

ghaerr:
> It is working!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

What is very cool about this is that ELKS doesn't use or need the XTIDE BIOS at all. So we can support XTCF cards without having to have a ROM chip, although I think most do. However, in order to boot from CF, the system definitely needs XTIDE BIOS, but is not used after boot.

For non-PC/XT systems, the same ATA driver should also work with CF drives. This has been tested on Solo/86, but not yet on other 286/386 systems. The current ATA driver should auto-configure itself to port 0x1F0 (standard ATA port) when accessing  /dev/cf0.

I'm not sure what all this means as far as accessing non-CF cards, like SD or USB-based flash. I suspect those use the much more complex USB software stack inside the BIOS, instead of an ATA interface. Does anyone have more information on this kind of thing?

fhendrikx:
> Now that this is working, I will move forward with testing a system that can be booted from CF, and then use the XTCF driver to access the root filesystem. This could be tricky to test, since both the machine BIOS, then ELKS boot, then ELKS kernel all need to be configured correctly to work.

Let me know if I can help here... I don't have a way to boot ELKS yet, other than through the ROM image.

fhendrikx:
> I'm not sure what all this means as far as accessing non-CF cards, like SD or USB-based flash. I suspect those use the much more complex USB software stack inside the BIOS, instead of an ATA interface. Does anyone have more information on this kind of thing?

I understand that SD cards have a different access method, and USB is different again. That said, we now have a base in place for ATA devices. There is also a nice little CF-card shaped SD card adaptor I have here... will try that out too, since that would be useful.

fhendrikx:
@ghaerr I've sent you a PR to fix the Solo/86 support.

PR #2357, opened 2025-07-16
[kernel] Add partition support to ATA CF driver

ghaerr:
The ATA CF driver now supports drive partitions for ATA, XTIDE, XTCF and SOLO86 configurations. For now, the media (usually CF card) must be present at boot time in order to recognize the card or partitions.

For now, /dev/cf0 and /dev/cf1 continue to correspond to the non-partitioned (flat/raw) device. New device names in the style of /dev/hda1 etc have been created: /dev/cfa1-8 for up to 8 partitions on the first device, and /dev/cfb1-4 for up to 4 partitions on the second CF device. This may change in the future to just be able to use /dev/hda1 etc instead of the /dev/cfa1 names. They are being kept separate so that testing can be done by using the BIOSHD driver simultaneously, even on the dame ATA drive (tried on QEMU only).

Tested on QEMU for up to 8 partitions, although seeing some problems with drives with "extended" partitions (i.e. more than 4 partitions).

`fdisk` is not yet supported for creating or changing partitions on ATA CF, but will be shortly.

Booting from ATA CF could possibly work but may not correctly determine the boot partition. I'm interested in more testing and experimentation on booting from ATA CF cards. In general, support for setting the root device to the ATA CF driver is not done yet, although booting from ATA CF to another root device should work. More on this coming.


toncho11:
One test can be: create a partition, install ELKS with fdisk and sys, boot with BIOS, change root=cf0-1 in /bootopts.

ghaerr:
Lots of stuff added:
- Booting from ATA CF driver now fully supported. Of course, hardware requires BIOS CF boot support to load kernel first
- Can configure system with ATA CF and BIOS for testing, or only ATA CF
- When booting from ATA CF with BIOS driver linked in (standard), use root=/dev/cfa1 etc to force mounting root filesystem
- Correct ATA CF boot partition calculated when no BIOS driver present, otherwise BIOS drive takes precedencei
- ATA driver will auto-configure for standard ATA or CF/XTCF. This means ATA driver can be used as direct HD or direct CF driver, replacing the BIOS entirely. No need to link in BIOS driver if no floppies present. ATA can also be linked with direct floppy driver for systems without any BIOS.
- Block devices renamed from /dev/cf0 and /dev/cf1 to raw devices /dev/cfa and /dev/cfb
- Partition names added /dev/cfa1, /dev/cfa2, ... /dev/cfa7, /dev/cfb1, ... almost all CF cards are MBR partitioned so use these
- Solo/86 still uses /dev/hda, /dev/hda1, /dev/hdb, etc
- Mount works with /dev/cfa1 etc.
- Fdisk works using /dev/cfa etc for reading partition table
- Extended partion fixes: MBR partitions are numbered 1-4, extended partitions 5-7
- Empty partition no longer created on first MBR extended partition
- Various block device open bugs fixed in ATA and BIOS drivers

Questions? Please ask; sounds a bit complicated but most of it works automatically.

Tested extensively on QEMU; feeling pretty confident partitions will work on real hardware and booting from ATA CF.

toncho11:
Looks very good!

elks/arch/i86/drivers/block/ata.c
```diff
@@ -401,23 +414,29 @@ sector_t ata_init(unsigned int drive)
         // Sanity check
         if ((buffer[ATA_INFO_CYLS] == 0 || buffer[ATA_INFO_CYLS] > 0x7F00) ||
             (buffer[ATA_INFO_HEADS] == 0 || buffer[ATA_INFO_HEADS] > 16) ||
-            (buffer[ATA_INFO_SPT] == 0 || buffer[ATA_INFO_SPT] > 63))
+            (buffer[ATA_INFO_SPT] == 0 || buffer[ATA_INFO_SPT] > 63) ||
+            (buffer[ATA_INFO_SECT_SZ] != ATA_SECTOR_SIZE))
```
fhendrikx: Just a reminder that sector size != 512 does occur on valid cards.

ghaerr: Hmmm, does that mean that reading a "sector" with an ATA read command doesn't return ATA_SECTOR_SIZE (=512) bytes? If so, that's a big problem. Or are you just saying that the sector size sometimes returns something funny even though LBA addressing always works using our assumed 512-bytes-per-sector scheme?

fhendrikx:
@ghaerr You got really keen :)  Nice work

elks/arch/i86/drivers/block/ata.c
fhendrikx: The latter; the card I'm thinking of returns something odd for sector size, but works just fine at 512. So I was just ignoring the sector size.

ghaerr: Ok. Thanks for the code review. I'll remove that check.

toncho11:
I am having problems with partitions. 
I created a new partition with both hda and cfa with fdisk, but then both sys and mkfat can not access it.
When I create a partition I should start from cylinder 0 or 1?
I am using a fat floppy disk. It gives an error that it can not open /dev/hda1 or /dev/cfa1.

ghaerr:
Did this problem happen before, or are you trying something new?

> I am using a fat floppy disk. It gives an error that it can not open /dev/hda1 or /dev/cfa1.

What do you mean floppy disk? Floppies aren't supported by the /dev/hda or /dev/cfa driver, nor are partitions supported on floppy disks.

If you post what commands you're running w/arguments, I can try to duplicate it over here.

toncho11:
I mean I boot from a floppy disk 360kb  and I try to install ELKS on my XT IDE + CF card as explained in the documentation:
https://github.com/ghaerr/elks/wiki#Installation

I tried mkfat -fat32 /dev/hda1 and mkfat -fat32 /dev/cfa1 after creating a partition, but it says it can not open it. Neither sys is able to access /dev/hda1 or /dev/cfa1

ghaerr:
> I tried mkfat -fat32 /dev/hda1 and mkfat -fat32 /dev/cfa1 after creating a partition, but it says it can not open it. 

Ok - I see. So you started with a non-partitioned CF card, then ran fdisk to create a partition on it. 

I think what is happening is that the BIOSHD and ATA CF drivers both only read the partition table on boot, but don't update the internal partition table after the fdisk, so the driver open fails. This probably needs to be changed from the way it is right now. Were you able to do this kind of thing before? I didn't see any code that would have allowed this to work.

I think a temp workaround would be after running fdisk, run sync, then reboot. When the system comes up, the boot screen should show the new partition you just created, and then mkfat should be able to open /dev/cfa1. Please try that.

toncho11:
I managed to even boot from the /cfa1

Here are some observations:
- I had a working cfa1 / hda1 after simply booting, reinforcing the idea that hda1/cfa1 are not updated after fdisk
- at line hda(n1,n2) there is a big pause 5 seconds that was non existent before (kind of annoying)
- this is a new card that I have never used for ELKS before
- the first time with sys I could not boot. I used cfa1 for the install.
- the second time I installed on hda1 and added the -M option
- after successful boot I changed the root=cfa1 and it is working ok

toncho11:
The delay on the hda(n1,n2) line is indeed 15 seconds. Then it says "RETRY 1 ..."

ghaerr:
>  I managed to even boot from the /cfa1

Wow, nice! There was a lot of code written to support that :)

> I had a working cfa1 / hda1 after simply booting, reinforcing the idea that hda1/cfa1 are not updated after fdisk

I'm thinking of ways to have the ELKS partition table automatically updated - the BIOSHD driver hasn't ever updated the partition table either, but CF cards are likely much more frequently reformatted. The potential downsides to an automatic partition update is when other partitions of the CF/HD are in use, or is mounted as the root filesystem.

> at line hda(n1,n2) there is a big pause 5 seconds that was non existent before (kind of annoying)

> The delay on the hda(n1,n2) line is indeed 15 seconds. Then it says "RETRY 1 ..."

I was thinking this isn't the hda() line, but the line after: the cfa: line, where the ATA CF driver is stuck in a busy loop trying to identify a non-existent drive. It could also be the XTIDE BIOS, not sure.

Can you post the exact display of the "RETRY 1 ..." line? I don't see any RETRY all-caps displays in the kernel.

I will rewrite the ATA CF driver timeout code, I suspect this could be the issue.

BTW - now that you can boot a CF card directly, can you easily create that card from your (Windows) PC? What is missing? It would be nice to be able to create a full bootable CF system directly from your desktop PC. Let me know what else needs to be done.

> the first time with sys I could not boot. I used cfa1 for the install.
> the second time I installed on hda1 and added the -M option

So you're saying here that cfa1 didn't work, but hda1 did? That would mean the ATA CF driver might have an issue. Did you use -M the first try on CF as well? That could also be the issue.

Using fdisk on either hda1 or cfa1 still for now requires a sync; reboot to re-read the partition table.

Thanks for your report and testing. I'll do some more over here.

ghaerr:
> at line hda(n1,n2) there is a big pause 5 seconds that was non existent before (kind of annoying)
> The delay on the hda(n1,n2) line is indeed 15 seconds. Then it says "RETRY 1 ..."

> I was thinking this isn't the hda() line, but the line after: the cfa: line

It would be nice to have a photograph of the boot screen. I'm thinking maybe this wait delay is AFTER the initial hda: and cfa: lines which identify the basic non-partitioned, full drive size (which should be the same).


If that is true, then it appears that the hda1 partition is calculated quickly, but the cfa1 partition is taking a long time. OR, if the cursor is still on the hda1 partition line (not a new line), then the hda2 partition code is taking a long time, and it is your machine BIOS giving the RETRY message. I need to know how many partitions are on the drive as well as see the 'fdisk' partition listing using 'p'. A bit complicated. I will do more testing over here.

> there is a big pause 5 seconds that was non existent before

Is this for a hard drive, or for the CF card using the BIOSHD driver? I don't fully understand what is going on.

toncho11:
So this the 15 second pause and after it. I am talking about one pause only.

ghaerr:
Thanks for the screenshots, that helps a lot. So hda is your internal HD and cfa is the CF. the "cmd 0 retry #1 ..." message is produced by our BIOSHD driver and indicates that the system BIOS returned an error trying to read a sector, and retried it, which is what's taking a while.

The sector read which had to be retried is CHS 0/0/1, which is the MBR. I hate to say it, but it seems that your HD boot sector may in fact be failing! Is this an old drive?  Did you recently fdisk that drive (which rewrites the partition table in the MBR)?

ghaerr:
You might try booting again with no CF card inserted... this should tell you for sure that the problem lies on your HD. The wait is occurring prior to the hda1(...) which indicates that it is occurring prior to displaying the HD partition table, which has nothing to do with CF. The CHS retry error confirms this.

toncho11:
Wait ... there is only an XTCF ISA card. There is no hdd.
hda = cfa (one physical media)
You mean to boot from floppy without the CF card inserted? No. This will give nothing.
So what I noticed is that the number of cylinders is different. Check the screenshots: 979 for hda and 980 and for cba. As this is the same media I was expecting them to be the same.

This single CF card is not used very much. It was on dos before and I have rarely used it. 
I just received a new CF card that should be almost identical.  Both are 256 MB.

toncho11:
On the old CF card (that I use for testing here) I lost my fat partition a few days ago. It got corrupted when accessed on WIndows.

ghaerr:
> Wait ... there is only an XTCF ISA card. There is no hdd.

Ok, that's what I thought, then I got confused (again). Easy to do with all this emulation going on. What is happening then is that both hda and cfa are the same card. But hda is being accessed through ELKS BIOSHD which then is through your XTIDE Universal BIOS (XUB). XUB is giving the "BIOS" read error on the MBR. There may be some special reason that XUB does that, I'm not sure yet.

> So what I noticed is that the number of cylinders is different. Check the screenshots: 979 for hda and 980 and for cba.

That's yet another BIOS issue: the C value is underreported by 1 by most BIOSes. That's not the issue here.

I can't emulate the XTIDE BIOS here, but its possible we can under 86Box.

ghaerr:
Can you run fdisk p)print option and take a picture of that? That might help debug perhaps a bad MBR that is confusing XUB.

toncho11:
So hda1 should not start at cylinder 0?

ghaerr:
The partition table looks good.

> So hda1 should not start at cylinder 0?

On some systems, they like it to start on CHS 1/0/1 rather than CHS 0/1/1, but that should not matter. Who knows why the XUB didn't like it.

However, fdisk didn't get a retry on (re)reading the MBR on /dev/hda... strange that ELKS is getting the BIOS retry only at startup. 

This is the first time we've ever supported booting off of CF, so there may be another issue though.

I have made other changes, you're not running the latest commit, I can see that from the cf: setting 8 bit mode message. The updated version says cfa: ... I'm not sure this has much to do with it.

Is it easy to create new CF images from Windows now, or still time-consuming?

ghaerr:
Looking hard at the boot screens, I just noticed something else which could be the problem: hda: is first initialized by BIOSHD, then cfa: by our ATA CF driver.

The ATA CF code reprograms the (same) CF controller, and XUB doesn't know about that. So there is the possibility that then, AFTER the cfa: init, the BIOSHD driver goes back and does more I/O: specifically to read the MBR using XUB. That's exactly when the error occurs.

What is possibly needed is a way to turn off the BIOSHD driver entirely and only use ATA CF. It's confusing anyway for the user to see the hda stuff (just like I was) and think it's a separate drive.

Let me think about a way to do that. I suppose a /bootopts option is needed since we don't want to recompile with the BIOSHD driver removed - too much work going back and forth for testing for sure.

ghaerr:
@toncho11, the ability to disable BIOS HD was added in PR #2364 using `disable=hda` in /bootopts. You might try that the next time you test booting from the CF card, it might remove the long wait you were seeing. If that works, then it shows that the ATA CF driver was interfering with the ELKS BIOS driver using the XTIDE Universal BIOS when access /dev/hda1, which is actually the same CF card as /dev/cfa using the ATA CF driver.

If that doesn't work, we can keep looking for an answer. The disable= line is also useful for disabling the ATA CF driver when an ethernet NIC card uses the same address, although we've recently added a probe function to the ATA CF driver that should prevent any problems.

PR #2359, opened 2025-07-18
[kernel] Add proper XTIDE 8-bit I/O handling to ATA CF driver

ghaerr:
Adds corrected handling for 8-bit XTIDE I/O to ATA CF driver, learned from https://github.com/86Box/86Box/discussions/5797#discussioncomment-13762737 and the [XTIDE BIOS](https://github.com/JayesonLS/xtideuniversalbios/tree/master) source.

Both XTCF and XTIDE require 8-bit I/O, but XTIDE doesn't use a Set 8-bit Feature controller command like XTCF but instead uses the data port+8 to read/write the high byte of word transfers from a controller hardware latch.

For now, the XTIDE uses the same linear register file as ATA, instead of the XTCF register file with offsets SHL 1. We will see whether any other changes are required when tested on real hardware.

For now, XTCF mode is automatically set when running on 8088 or 80186 CPUs. To set XTIDE operation, the ATA CF driver needs to be recompiled with the following line changed in ata.c:
```
static int mode = MODE_XTIDE;         /* change this to force a particular controller */
```
A /bootopts option may be added at a later point to select XTIDE dynamically at boot, along with possible settings for the base I/O address.

Also adds @fhendrikx's higher speed I/O from PR #2353.

### DIFF
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -393,10 +393,10 @@ void ata_reset(void)
     {
         if (arch_cpu <= CPU_80186)
         {
-            if (mode == MODE_XTCF)
-                xfer_mode = XFER_8XTCF;
-            else
+            if (mode == MODE_XTIDE)
                 xfer_mode = XFER_8XTIDE;
+            else
+                xfer_mode = XFER_8XTCF;
         }
         else
             xfer_mode = XFER_16BIT;
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -50,10 +50,15 @@
 #define MODE_SOLO86 3           /* XTCF at ports 0x40/0x5C, 16-bit I/O */
 #define AUTO        (-1)        /* use MODE_ATA for PC/AT (286+), MODE_XTCF on PC/XT */
 
+/* hardware I/O port xfer modes */
+#define XFER_16BIT  0           /* standard 16-bit ATA I/O */
+#define XFER_8XTCF  1           /* XTCF set 8-bit feature cmd, then xfer lo then hi */
+#define XFER_8XTIDE 2           /* XTIDE 8-bit xfer hi at data+8 then lo at data+0 */
+
 /* configurable options - may have to be changed and kernel recompiled*/
 
 static int mode = AUTO;         /* change this to force a particular controller */
-static int use_8bitmode = AUTO; /* =0 16-bit xfer, =1 force 8-bit xfer, AUTO=automatic */
+static int xfer_mode = AUTO;    /* change this to force a particular I/O xfer method */
 
 /* default base port (change if required)   ATA, XTIDE,  XTCF, SOLO86 */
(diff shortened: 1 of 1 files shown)


toncho11:
So what will my XTIDE + IDE to SD card adapter? The IDE to SD should be  a standard IDE device on the XTIDE card.

It should be the above: `static int mode = MODE_XTIDE; `?

ghaerr:
> So what will my XTIDE + IDE to SD card adapter? The IDE to SD should be a standard IDE device on the XTIDE card.

> It should be the above: static int mode = MODE_XTIDE; ?

Yes. I'd like to write some auto-probing software but haven't figured out how to do that yet. 

What's the brand of your XTIDE card?

toncho11:
The one with the SD IDE adapter that I want to try is: Monotech XT‑IDE Deluxe / Rev 5.2.
There is a jumper called "IDE address" and it is clearly set to default 0x300.

ghaerr:
Let me see if I can learn about that card a bit to make sure it'll likely work, so you don't waste time trying it.

ghaerr:
> The one with the SD IDE adapter that I want to try is: Monotech XT‑IDE Deluxe / Rev 5.2.

Alright, I think I've finally got enough information on XTIDE to get a driver working for that card. Long story short, XTIDE changed the register layout from v1 to [v2](https://minuszerodegrees.net/xtide/rev_2/XT-IDE%20Rev%202%20-%20general.htm). In particular [high speed mode](https://minuszerodegrees.net/xtide/XT-IDE%20-%20Hi-speed%20mode.htm) was added, which changed the register file port layout to allow high speed "16-bit" I/O using the capability of writing a word to two consecutive port address which almost all 8088 system busses support (we use the same "trick" in our fast VGA routines):
XT-IDE - Register map

Right now, the ATA CF driver implements XTIDE v1. It looks like your MonoTech card is XTIDE v2, so I'm going to have to add yet another operating mode to our ATA CF driver. Some cards have a jumper option for high-speed mode or not, effectively implementing a v1 compatible mode.

I'll post an updated driver shortly.

toncho11:
I have a jumper for high-speed mode indeed.

toncho11:
The high speed mode should be an option. The card is not marked well, but I think the jumper is to off currently.
I vaguely remember that it did not work on some machines.

toncho11:
So now I should use `xtide=2` for my XT-IDE + SD card?

```
xtide=0    standard ATA (default for PC/AT (80286+) systems)
xtide=1    XTIDE rev 1 (pre 2012)
xtide=2    XTIDE rev 2 (now standard for XTIDE)
xtide=3    XTCF (automatically set for 8088/8086 systems
```

ghaerr:
> So now I should use xtide=2 for my XT-IDE + SD card?

> I have a jumper for high-speed mode indeed.

> The high speed mode should be an option.

The difference between XTIDE rev 1 and rev 2 is the high speed option, as documented above. The high speed option swaps address lines A3 and A0 on the card, which allows a using single 16-bit IN/OUT instruction for I/O, increasing speed. This option requires a change in the driver, so you should set xtide=1 for rev 1 or xtide=2 for rev 2, depending on which way your jumper is set.

PR #2364, opened 2025-07-21
[kernel] Add disable= /bootopts option to disable ATA CF, BIOS FD/HD or DF drivers

ghaerr:
Allows disabling compiled-in drivers for testing, or for when drivers may try to talk to the same hardware. Discussed in https://github.com/ghaerr/elks/pull/2357#issuecomment-3092540943.

/bootopts now parses a comma separated list of block device names to disable them at boot. The complete list of allowable drivers is:
```
disable=hda,cfa,df0,fd0
```
Note that there are two options for the BIOS driver: hda turns off the hard disk, and fd0 the floppy. df0 is the direct DF driver and cfa is the ATA CF driver.

This line should be edited to only include only drivers that might be causing problems. Leaving in the entire list will end up disabling all bootable block devices, and the boot will fail when trying to mount the root device.

The most common use of this option will likely be to either turn off the BIOS hard disk driver when booting from CF card. To do this, one would use:
```
disable=hda
root=cfa
```

Since the ATA CF driver will likely be compiled in, one might want to turn it off when not desired, and just use the BIOS HD driver instead:
```
disable=cfa
```

### DIFF
--- a/elks/arch/i86/drivers/block/ata-cf.c
+++ b/elks/arch/i86/drivers/block/ata-cf.c
@@ -12,6 +12,7 @@
 #include <linuxmt/kernel.h>
 #include <linuxmt/errno.h>
 #include <linuxmt/genhd.h>
+#include <linuxmt/devnum.h>
 #include <linuxmt/debug.h>
 #include <arch/ata.h>
 
@@ -66,6 +67,10 @@ struct gendisk * INITPROC ata_cf_init(void)
 
     // register device
 
+    if (dev_disabled(DEV_CFA)) {
+        printk("cfa: disabled\n");
+        return NULL;
+    }
     if (register_blkdev(MAJOR_NR, DEVICE_NAME, &ata_cf_fops))
         return NULL;
     blk_dev[MAJOR_NR].request_fn = DEVICE_REQUEST;
--- a/elks/arch/i86/drivers/block/bioshd.c
+++ b/elks/arch/i86/drivers/block/bioshd.c
@@ -43,8 +43,9 @@
 #include <linuxmt/string.h>
 #include <linuxmt/mm.h>
 #include <linuxmt/memory.h>
-#include <linuxmt/debug.h>
 #include <linuxmt/timer.h>
+#include <linuxmt/devnum.h>
+#include <linuxmt/debug.h>
 
 #include <arch/io.h>
 #include <arch/segment.h>
@@ -74,8 +75,8 @@ struct elks_disk_parms {
 } __attribute__((packed));
 
 static int bioshd_initialized = 0;
-static int fd_count = 0;                /* number of floppy disks */
-static int hd_count = 0;                /* number of hard disks */
+static int fd_count;                    /* number of floppy disks */
+static int hd_count;                    /* number of hard disks */
 
 static int access_count[NUM_DRIVES];    /* device open count */
 struct drive_infot drive_info[NUM_DRIVES];   /* operating drive info */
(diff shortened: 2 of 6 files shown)


toncho11:
Xtide=3 works but disabled=hda crashes

toncho11:
It was terrible experience.
- Win32 imager did write correctly the hdd32 fat image to the cf card. I suppose because I did not set anywhere the CHS
- you need to restart after fdisk
- sys /dev/hda1 will not upgrade your distribution .... I already knew that, but tried anyway. It copies the linux file in a way that it does not allow it too boot. It says "bad header ..."
- also the cp command when booted from floppy and trying to copy between to partitions mounted as /mnt and /mnt2 will actually try to copy the files first on the floppy and then on the destination partition? It told there is no space ... which is the case on the floppy disk, not the destination partition. 

As I said - terrible experience. I forgot to mention that I could not mount my fat32 partition ... until I deleted /tmp and restarted. I also got heap errors at some point :)

toncho11:
Also it is not clear which is the root drive from the boot messages.
Wait I disabled the hda, but did not set the root=cfa1. That must be the problem.  I need all:

```
xtide=3
disabled=hda
root=cfa1
```

ghaerr:
> It was terrible experience.

Sorry. I would say there is quite a bit of things being tested here at the same time.

> Xtide=3 works but disabled=hda crashes

The reason for that, as shown in the screenshot, is that without root=cfa, but with disabled=hda, the kernel can't mount the root filesystem /dev/hda (device 0x300). This was mentioned in the PR above.

> you need to restart after fdisk

I'm working on a fix for that.

> sys /dev/hda1 will not upgrade your distribution .... I already knew that, but tried anyway. It copies the linux file in a way that it does not allow it to boot. It says "bad header ..."

I'm not quite sure the reason for this... /linux needs to be in the first few directory entries of the / root directory, but the message you're getting with "bad header" is different than that.

> also the cp command when booted from floppy and trying to copy between to partitions mounted as /mnt and /mnt2 will actually try to copy the files first on the floppy and then on the destination partition? It told there is no space ... which is the case on the floppy disk, not the destination partition.

Can you explain this a bit more? I don't quite understand: what do you mean "copy the files first on the floppy?" What devices are mounted on /mnt and /mnt2?

> Also it is not clear which is the root drive from the boot messages.

"Unable to open root device 0x300" is /dev/hda. That can be seen by doing a ls -l /dev. I'll see if I can get a better message displayed in text rather than numbers.

> Wait I disabled the hda, but did not set the root=cfa1. That must be the problem. I need all:

Yes, when you disable=hda, you must root=cfa, since the system is default booting from HDD.

BTW, the only reason you "must" disable hda is to test whether that stopped the BIOS retry message and long wait at boot you reported earlier. Did it?

ghaerr:
> sys /dev/hda1 will not upgrade your distribution .... I already knew that, but tried anyway. 

Are you running `sys` right after running fdisk and rebooting? That will be required, as `makeboot` is run by `sys` and creates /linux and /bootopts. I'll test this to see what might be happening.

Thanks for your testing, sorry you are very frustrated.

toncho11:
So it works! I changed the root=cfa1 on Windows and it works now.
No pause. disabled=hda managed to remove the pause.

ghaerr:
Ok, so configuring with new stuff is frustrating, but now you figured it out and the pause is gone, disabled=hda works, and overall you can boot from cfa1. Very nice!

What still needs work seems to be:
- fdisk shouldn't require reboot afterwards (working on that)
- sys after fdisk/reboot doesn't work (how did you end up working around that problem?)
- "Can't mount root 0x300" message should show /dev/ name (working on that)

Thank you for testing!

toncho11:
mnt is /hda1
mn2 is /hda2
And it told me not enough space even if /hda2 is empty 60mb
I am on fa32 for hda1 and hda2
floppy is fat16 I suppose.

`cp /mnt/bin/* /mnt2/bin`

ghaerr:
> And it told me not enough space even if /hda2 is empty 60mb

Can you run `df`? Does that show proper free space?

> floppy is fat16 I suppose.
> cp /mnt/bin/* /mnt2/bin

So the failure has nothing to do with floppy, but fails when copying from hda1 to hda2? Is either of them a CF card? I thought you were running only CF?

toncho11:
I mean cfa1 and cfa2. But at the time I was using the bios driver so hda.
Never mind this. I think the problem could be only when we boot from floppy. Any way. Never mind.

toncho11:
Yes, when mounting the root there is a lot of information, but not which is the root in /dev.

ghaerr:
> Never mind this. I think the problem could be only when we boot from floppy. 

Ok, I'll drop that issue for now. Report it again if you have cp issues.

> Yes, when mounting the root there is a lot of information, but not which is the root in /dev.

Got it. Will put a change to report what root device name in /dev is so one doesn't have to remember numbers.

toncho11:
And actually you can not switch to direct floppy access from /bootopts ? I mean without setting the root?

ghaerr:
> And actually you can not switch to direct floppy access from /bootopts ? I mean without setting the root?

Yes, you should be able to just use /dev/df0 (not fd0, but df0) to access the floppy using the non-BIOS direct FD driver.

ghaerr:
To boot from direct floppy, use root=df0 in /bootopts.

toncho11:
I got it. df0 is indeed always compiled and accessible.
Thank you! :)

ghaerr:
So what we now have is two BIOS drivers, hda and fd0, which use ELKS BIOS to access the HDD and FD. However, if you're running XTIDE Universal BIOS, hda could also access the CF card. In that case, we probably want to disable=hda.

We then have two direct drivers, cfa and df0. 

All can be booted from using root=, and any can be disabled using disable=

> df0 is indeed always compiled and accessible.

Yes - and now, so will hda, cfa, and fd0 as well. That's the fancy disable= option was added, which takes comma separated device names if wanted.

ghaerr:
My plan is to ship ELKS with all the drivers (BIOS hda and fd0), ATA CF (cfa) and direct FD (df0) compiled in. By default, the old BIOS driver will be used, but root=, xtide= and disable= can be used to select your own dynamic configuration.

ghaerr:
>  I forgot to mention that I could not mount my fat32 partition ... until I deleted /tmp and restarted. I also got heap errors at some point :)

I don't quite understand this error. If you end up going through this sequence again, please report the exact errors and I'll try to get it fixed.

toncho11:
All clear :)

PR #2368, opened 2025-07-22
small build fixes

hexadec1mal:
fix parallel make
fix env.sh
fix compilation with new host gcc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 *.o
 *.a
 *.map
+*.sym
 
 #
 # Normal output and testing dirs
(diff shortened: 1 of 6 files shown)


ghaerr:
Hello @hexadec1mal,

Thanks for your enhancements, the make and gcc updates look good.

Can you please revert the change to image/Image.all and its associated change in image/Make.image? While you're correct that all of target/ is added to the image, the Image.all file is used to order the image for faster boot from real floppies, based on extensive analysis of disk sectors used at boot. This was performed and discussed in #2071.

Thank you!

hexadec1mal:
@ghaerr done

PR #2375, opened 2025-07-29
[kernel] Fix major speed bug in ATA CF driver

ghaerr:
Fixes major speed problem in ATA CF driver, first identified by @toncho11 through testing and unrelenting intuition that something was definitely wrong, in https://github.com/ghaerr/elks/pull/2370#issuecomment-3124456438.

The ATA CF driver was running two times slower than the XTIDE Universal BIOS, as discussed at length in #2370. The problem was solved after setting up the [PCem Emulator](https://github.com/sarah-walker-pcem/pcem) and then modifying it to support XTCF 8-bit I/O while emulating XTIDE v1. While this allowed actual emulation of the insb/outsb macros for XTCF without requiring @toncho11's hardware, only after using ELKS' [precision timing routines](https://github.com/ghaerr/elks/pull/1962) within the driver was the speed issue identified.

The speed issue was that, after writing the drive select byte to the ATA controller, a 10ms delay was performed, which replaced the original 400ns delay. The 10ms delay was enhanced to use the kernel timer which ticks at 10ms resolution. This ended up adding 10-20ms delay per I/O request, which ended up adding tons of time to all ATA CF I/O.

After more research into the ATA spec and the [OSDev Wiki](https://wiki.osdev.org/ATA_PIO_Mode#400ns_delays), it was determined that no select delay at all is actually required by any hardware, since the BSY status is checked prior to sending the drive select command. Removing the wait allowed the driver to run approximately twice as fast.

The driver delay was not limited to XTIDE or XTCF operation, it was also seen using QEMU's 16-bit ATA emulation, where the root=cfa1 boot time decreased from 5.39 secs to 0.58 secs, ten times faster.

The PCem emulated XTCF boot time decreased from 6.64 secs to 2.24 secs, and a `dd` copying 5000 sectors decreased from 2:55 minutes to 42 seconds.

The ATA CF driver is now quicker than the BIOS XUB driver. This probably means we don't have to update the insb/outsb macros even though a slight speed improvement might still result.

Fixes included in this PR are:
- eliminate 10-20ms delay before every command
- eliminate any delay after sector reads
- adding EMUL_XTCF test harness to allow quick reconfiguration for testing on PCem
- alternate revert for previous env.sh fixup for systems without `realpath` (macOS Catalina)

Fixes to be included in another PR:
- add error check to sector write routine
- possibly implement multi-sector I/O
- possibly improve ins/outs macros

Thank you @toncho11 for your testing and not giving up, allowing this to be identified and fixed!

### DIFF
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -52,6 +52,7 @@
 #include <linuxmt/heap.h>
 #include <linuxmt/memory.h>
 #include <linuxmt/debug.h>
+#include <linuxmt/prectimer.h>
 #include <arch/io.h>
 
 /* hardware controller access modes, override using xtide= in /bootopts */
@@ -70,6 +71,7 @@
 
 /* configurable options */
 #define FASTIO          1       /* =1 to use ASM in/out instructions for I/O */
+#define EMUL_XTCF       0       /* =1 to force XTCF xfer mode for use with PCem */
 
 static int xfer_mode = AUTO;    /* change this to force a particular I/O xfer method */
 
@@ -245,6 +247,7 @@ static void ATPROC write_ioport(int port, unsigned char __far *buffer, size_t co
 /**
  * ATA try setting 8-bit transfer mode, may be unimplemented by controller
  */
+#pragma GCC diagnostic ignored "-Wunused-function"
 static int ATPROC ata_set8bitmode(void)
 {
     int error;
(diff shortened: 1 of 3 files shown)


toncho11:
Thank you for fixing it so fast!!! That was impressive! And thank you for adapting [PCem](https://pcem-emulator.co.uk/), this saved me a lot of time and efforts.

Bravo!

toncho11:
So the ata_select is executed on every workflow (wait, send command, wait, read/write ...) before read or write. So it was a 10ms delay for each 512 bytes of read or write (instead of the original 400ns). And now there is none. Great :)

Issue #2377, opened 2025-08-02
Drive information CHS for hdb is duplicated of hda


tyama501:
Hello @ghaerr 

I have noticed the drive information CHS for hdb is duplicated of hda with the latest build.
Were there some changes here recently?

Here is the message when hda is 19MB and hdb is 250MB.
It says hda: 19M CHS 300, 8, 17, hdb : 19M CHS 300, 8, 17.

<img width="698" height="523" alt="Image" src="https://github.com/user-attachments/assets/f2daa09b-617d-4cbd-a507-7426ce92ce31" />

If I switch the order of these drives,
then it says hda: 250M CHS 3765, 8, 17,  hdb: 250M CHS 3765, 8, 17.

<img width="698" height="523" alt="Image" src="https://github.com/user-attachments/assets/3c452ec8-e6b1-44d8-a12d-a8bfa84188d5" />

I can mount partitions of these drives, so it seems that the issue is only for the messages.

Thank you.

tyama501:
I have confirmed it does not occur on the last December build.

ghaerr:
Hello @tyama501,

> I have noticed the drive information CHS for hdb is duplicated of hda with the latest build.
Were there some changes here recently?


Yes, when the partition code was generalized to support the ATA CF driver, a new routine in genhd.c::show_drive_info() was written which displays the drive info:
```
void GENPROC show_drive_info(struct drive_infot *drivep, const char *name, int drive,
    int count, const char *eol)
{
    unsigned long size;
    char *unit;
    static char UNITS[4] = "KMGT"; 

    for (; count; count--, drive++) {
        if (drivep->cylinders != 0) {
            unit = UNITS;
            size = (unsigned long)drivep->sectors * 5;  /* 0.1 kB units */
            if (drivep->sector_size == 1024)
                size <<= 1;
            size *= ((unsigned long) drivep->cylinders) * drivep->heads;

            /* Select appropriate unit */
            while (size > 99999 && unit[1]) {
                debug("DBG: Size = %lu (%X/%X)\n", size, *unit, unit[1]);
                size += 512;
                size /= 1024U;
                unit++;
            }
            debug("DBG: Size = %lu (%X/%X)\n",size,*unit,unit[1]);
            printk("%s%c: %4lu%c CHS %3u,%2d,%d%s",
                name, drive + (drivep->fdtype < 0? 'a' : '0'), (size/10), *unit,
[...]

PR #2378, opened 2025-08-02
[genhd] Add incrementing drivep in show_drive_info

tyama501:
Hello @ghaerr 

Yes, the fix for #2377 worked.
Thank you!

image


ghaerr:
Thanks @tyama501. BTW, in the `uname -a` above, what is the graphics triangular character displayed in the date field?

tyama501:
It occurs when I build it with Japanese supported terminal and the build date has kanji.
It does not occur if I build it with English terminal.

ghaerr:
I see. Does that display proper kanji only on PC-98 console, or also used to display using UTF-8 or japanese byte encoding on serial terminal?

tyama501:
Well, I haven't try that on the serial console but it might be able to display it.
It should display 月 which means month.
I will try it when I have time.

PR #2388, opened 2025-09-20
[basic] fix 16 color mode setting for pc-98

tyama501:
Hello @ghaerr 

This is the similar PR that has been made in scr_pc98 of nano-X.
Wrong parameter is fixed.

Thank you.


ghaerr:
Thank you @tyama501 !

PR #2394, opened 2025-10-10
[cmds] Update ecalc

HelleBenjamin:
Hi,
I've made some enhancements to the ecalc. Here are the changes in a nutshell:
- Added bitwise operations, such as OR and AND.
- Added a help message.
- Made the program include only on 1440k floppy images and larger.


ghaerr:
Hello @HelleBenjamin,

Thank you for the enhancements. 

For command help messages, in order to keep our disk image usage low, it is recommended that all help messaging be moved to a man page (in your case, in elkscmd/man/man1/ecalc.1), instead of embedding help within the executable. This is now the case for all ELKS executables. To display a man page on ELKS, use `man cp`, on the host, use `eman cp`. Copyright messages also don't need to be embedded in the executables. For executables, the smaller the better!

HelleBenjamin:
Alright. I'll do those in the next version. Thanks for the feedback.

PR #2397, opened 2025-10-11
[cmds] Update ecalc

HelleBenjamin:
Hi,
I've made some changes, such as moving the help message into a man page and cleaning up the code overall (mostly strings), thus making the executable smaller. I also added support for using previous result with the keyword "ans".

### DIFF
--- a/elkscmd/man/man1/ecalc.1
+++ b/elkscmd/man/man1/ecalc.1
@@ -1,0 +1,52 @@
+.TH ecalc 1
+.SH NAME
+ecalc \- simple integer calculator for ELKS
+.SH SYNOPSIS
+.B ecalc
+.SH DESCRIPTION
+.I ecalc
+is a simple terminal integer calculator for ELKS. It uses 16-bit signed integers(-32768 to 32767).
+.SH OPERATORS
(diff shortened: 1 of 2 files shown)


ghaerr:
Nice, and thanks for the man page.

tyama501:
Hello @ghaerr,

I have compiled 1440k pc-98 image, but it seems that ecalc is not included,
although I can see :1440k in the Applications list. 
https://github.com/ghaerr/elks/blob/master/elkscmd/Applications
Do you know why this happen?
 
image

Thank you.

ghaerr:
> I have compiled 1440k pc-98 image, but it seems that ecalc is not included,

That is strange, I can't duplicate this over here, even when using stock pc98-1440.config. 

Can you check end of image build listing to see if following is occurring:
```
...
install misc_utils/mined to bin/edit
install misc_utils/sleep to bin
install misc_utils/tty to bin
install misc_utils/uuencode to bin
install misc_utils/uudecode to bin
install misc_utils/ecalc to bin <-- looking for this line
install elvis/elvis to bin/vi
install minix1/banner to bin
...
```

Issue #2407, opened 2025-10-19
Default hda / cfa


toncho11:
I just tried the latest ELKS on https://copy.sh/v86/ and there is this delay I observed previously when both hda and cfa drivers are querying the hardware one after another. The delay can make people think that ELKS has halted. 

Out of many solutions could be the one where the cfa driver is disabled. It is the safest bet for both real hardware and emulators. In the same time not advertising the cfa driver is a pity, considering the effort invested in it. Maybe a simple message such as: "Direct ATA and XTIDE are ELKS supported with cfa, check wiki."

ghaerr:
> The delay can make people think that ELKS has halted.

Looking at the current ATA CF driver, what seems to be happening is that a "probe" is made at kernel startup to determine whether an ATA controller is present when the ATA CF driver is enabled. The probe is very quick but and is likely failing, since copy.v86 doesn't emulate it. 

However, the driver (for now) purposely ignores the failed return value and goes ahead with sending an ATA IDENTIFY command to the controller, which I see uses a 10 second timeout. The 10 seconds was recommended by some ATA documentation is is likely way too long. Apparently on some CF cards it can take seconds in some cases to get a valid sector (?).

I left the ATA IDENTIFY in the driver since we were in development, and there may have been a case where ATA was present but the probe failed, and we didn't bother going further. So now, we could either just return immediately on probe fail, or reduce the 10 second read time to a much lower value for only the IDENTIFY command - perhaps 1 second? Reducing the time to 1 second would likely not affect working ATA CF systems and would only result in a 1-2 second delay on boot. OTOH just returning from a failed probe immediately would result in no 1-2 second boot delay.

Which do you think we should do?

> In the same time not advertising the cfa driver is a pity, considering the effort invested in it. 

Yes - I think it would be better to take one of the two options above than disable the driver. I'm leaning towards just a 1 second delay on ATA IDENTIFY and trying that out.

toncho11:
The cfa driver is not used until it is configured - until a proper switch to cfa is configured.
I do not know how reliable the probe is. Maybe we should stop there.
The thing is that people are not actually aware what cfa is and we lose time detecting it. But then this is done for other things such as network cards for example.
Problem is also that once the driver is in active use we use the same probe/identify and then in this case the proper detection becomes more important?
I am for my original proposition or 2 second cfa check. A 2 second cfa check with additional message such as "Direct disk ATA detected/failed".

ghaerr:
[...]

PR #2408, opened 2025-10-19
[kernel] Reduce wait time for ATA CF drive identification to 1 sec

ghaerr:
Discussed in #2407. 

Should reduce boot delays when ATA CF driver compiled in and no CF or HD found from 10 seconds to 2 seconds max. This could possibly be reduced further after testing on real hardware with various real hard drives, but it is unknown whether ATA HD controller setup information is saved on disk or in controller firmware.

Also clarifies ATA driver startup messages when controller present and not found.

ATA controller not found:
cfa probe fail

Controller found and one drive/card recognized:
cfa only

### DIFF
--- a/elks/arch/i86/drivers/block/ata.c
+++ b/elks/arch/i86/drivers/block/ata.c
@@ -95,6 +95,7 @@ static unsigned int xlate_XTIDEv2[8] = {
 
 /* wait loop 10ms ticks while busy waiting */
 #define WAIT_50MS   (5*HZ/100)      /* 5/100 sec = 50ms while busy */
+#define WAIT_1SEC   (1*HZ)          /* 1 sec wait for identify */
 #define WAIT_10SEC  (10*HZ)         /* 10 sec wait for read/write */
 
 /* end of configurable options */
@@ -307,7 +308,7 @@ static int ATPROC ata_select(unsigned int drive, unsigned int cmd, unsigned long
 static int ATPROC ata_cmd(unsigned int drive, unsigned int cmd, unsigned long sector,
     unsigned int count)
 {
-    int error;
+    int error, wait;
     unsigned char status;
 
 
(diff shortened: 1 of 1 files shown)


toncho11:
I tried latest build with copy.86 and it is a 30 seconds delay. It was supposed to be 2-3 seconds, right?

ghaerr:
> I tried latest build with copy.86 and it is a 30 seconds delay. It was supposed to be 2-3 seconds, right?

Yes. Let me go back to the drawing board on this and test it better before resubmitting. Thank you for taking the time to test.

ghaerr:
@toncho11 

> I tried latest build with copy.86 and it is a 30 seconds delay. It was supposed to be 2-3 seconds, right?

I added debug code to the PR and it is operating properly. It seems there is something strange going on with the copy.86 emulator with regards to the timeouts and/or ATA registers. Is the 30 second delay shorter than the original delay you saw? I can't find your original post on the subject.

I'm not sure what the next step is. BTW the ATA CF driver will use standard ATA port 1F0h (not 300h) when using xtide=0. I'm not sure whether this part of the issue with your outstanding problem on 86Box emulation.

ghaerr:
@toncho11,

I uploaded the latest commit (1440k MINIX floppy image) to copy.86 (image/fd1440.img and booted: the boot took 2.36 seconds. This version has a display for the ATA CF wait calls (see 'wait(100)` = 1 second) below. So I'm not sure what is happening with your version taking 30 seconds? It seems there might be some other factor at play here.

Screenshot 2025-10-20 at 10 19 18 AM

Thank you!

ghaerr:
@toncho11 update: 

After testing more heavily on copy.86, it appears the ATA CF driver is working properly with minimal wait, but the BIOSHD driver is hanging for 30 seconds after displaying "hda: (0, 63504)". I suspect this may be the older IDE probe. Looking further into it.

ghaerr:
@toncho11 This should be fixed now with #2411. Here's an hd32-minix.img you can use for testing so you don't have to rebuild.

[hd32-minix.img.zip](https://github.com/user-attachments/files/23006002/hd32-minix.img.zip)

PR #2410, opened 2025-10-20
ハイレゾ98対応部がメインですが FD1200対応も目玉です

drachen6jp:
再度やってみました。お手数おかけします

宛先間違えました 消します


ghaerr:
Hello @drachen6jp,

Thank you for your PR. It looks like your PR is from an older tree, but this is OK, even though PR is showing large number of commits, end result will only affect PC-98 code and files, so this is not a problem and will use a squash commit.

You are welcome to resubmit. I would ask the following changes be made:
- for README.md: add PC98 changes to middle or end of file, not at top (or change README.md in separate PR).
- Rewrite Japanese comments in English, if possible
- In setup.S, don't delete blank lines outside of PC98 code, for better tracking with other projects
- Best not to add new `#ifdef CONFIG_IMG_xxx` types in bioshd.c, this requires recompilation for each image type. Currently this is only done for 1232K floppies and PC98 build, rather than 3 separate PC98 builds for new 1200k and 720k floppy types. Is it possible to add 1200K and 720K floppy support without ifdef, or only within CONFIG_ARCH_PC98?

We can discuss more after PR resubmitted.

Thank you!

drachen6jp:
Thank you for your reply.
I am talking about these changes with Mr.tyama501.
if this problem was end,I will post this code again with your advice.

sorry and thank you.

tyama501:
Hello @ghaerr and @drachen6jp ,

I will cherry-pick (and modify a little) for FD1200 support and reboot functions once @drachen6jp code is settled to make separate PRs. Maybe it is better to keep high-resolution PC-98 support in the fork branch for a while since it seems that there are huge number of differences, although it is interesting to support high resolution nano-X in the future.

Thank you.

ghaerr:
Hello @tyama501,

I like the idea of moving in changes with separate PRs, that will be easier to integrate. Go ahead and have yourself or @drachen6jp submit PRs, that will be nice to get @drachen6jp's enhancements into master!

> although it is interesting to support high resolution nano-X in the future.

That would be great! I didn't realize that the high resolution enhancements worked for Nano-X, I had assumed it was just the text console.

Thank you!

tyama501:
Hello @ghaerr ,

Sorry, I meant the nano-X is future plan.
I think the current modification of @drachen6jp is just for the text console and memory difference (it has 768kB memory!) for bootoptions. But I don't think it is hard to support nano-X if it uses 4-planes. (I don't know)

I will try to keep his name when cherry-pick.

ghaerr:
> I will try to keep his name when cherry-pick.

I think the author can be changed for a single commit by using the --author option:
```
git commit --author="John Doe <john@doe.org>"
```
You may try using that option before you `git push ...` for the PR. I am not sure yet how to change to GitHub name @drachen6jp for example. It might be as simple as
```
git commit --author=@drachen6jp
```
but I have not tested that.

PR #2411, opened 2025-10-20
[kernel] Issue BIOSHD disk reset when ATA CF driver present on boot

ghaerr:
Finally fixes extended boot delay issue discussed in https://github.com/ghaerr/elks/pull/2408#issuecomment-3422276867.

When the ATA CF driver is initialized, the BIOSHD driver was hanging on the next I/O request, which happened to be in the MBR partition handling code. Apparently both drivers talking to the same ATA hardware caused the BIOS code to wait 30 seconds before returning an I/O error, then retrying. Issuing a BIOS disk reset beforehand eliminates this problem.

### DIFF
--- a/elks/arch/i86/drivers/block/ll_rw_blk.c
+++ b/elks/arch/i86/drivers/block/ll_rw_blk.c
@@ -20,6 +20,7 @@
 #include <linuxmt/init.h>
 #include <linuxmt/mm.h>
 #include <linuxmt/ioctl.h>
+#include <linuxmt/biosparm.h>
 #include <linuxmt/debug.h>
 
 #include <arch/system.h>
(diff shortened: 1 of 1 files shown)


toncho11:
Tested quickly, seems OK now.
Thank you.

toncho11:
I have not tested using the cfa driver, just booting as image provided.
What are the chances that the reset from hda driver will interfere with the cfa when configured to be used?

I just tried xtide=0 in copy.86 and it was fine.

ghaerr:
> What are the chances that the reset from hda driver will interfere with the cfa when configured to be used?

That's a pretty good question, I am assuming the problem is that the BIOS likely caches some ATA hardware settings and needs a reset in order to prevent a retry, which then internally does a reset. So I tested some more with copy.86 and it all seems to be working well. 

I am guessing that copy.86 uses a real BIOS which emulates the potential problem better than QEMU, whose BIOS is written in C. QEMU has not shown any retry behaviour, so I hadn't seen it previously. I'll be testing on copy.86 more frequently now.

> I just tried xtide=0 in copy.86 and it was fine.

Great! I think we're good to go, another ATA CF problem solved. Thank you!

PR #2414, opened 2025-10-22
delete some ifdef options

drachen6jp:
PC-98 type FD1200 can read and write.
thank you!


ghaerr:
Although this looks like a work in progress, I like the idea of trying to remove all ifdef CONFIG_IMG_FD1232, CONFIG_IMG_1440 or CONFIG_FD720 from the BIOSHD driver. 

It would be much better to determine dynamically what type drives the system has installed, preferably with a single routine, which after calling can both setup the bios_drive_map[] tables as well as handle differences between operation. Is this possible?

It would also be very nice to remove these same ifdefs in the boot_sect.S boot sector. Is there enough room to do so? There is possibly extra code space in the MINIX boot sector, but FAT boot sector is very tight.

PR #2415, opened 2025-10-22
Elks hirezo98

drachen6jp:
some TVRAM point change to 0xE000.
Floppy: MFM1024(FD1232) and MFM512(FD1200 FD1440 FD720) both readable.

### DIFF
--- a/bootblocks/boot_sect.S
+++ b/bootblocks/boot_sect.S
@@ -23,8 +23,9 @@
 #define SECTOR_SIZE	1024
 #else
 #define SECTOR_SIZE	512
+#define MEMTOPSEG	0x8000		// 512K memory
 #endif
-#define MEMTOPSEG	0xA000		// 640K memory
+
 
 #else
 // IBM PC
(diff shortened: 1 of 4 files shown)


ghaerr:
Hello @drachen6jp,

Thank you for submitting this!

It appears this PR was trying to add a number of different enhancements, is this correct?:
- Boot with 512K memory instead of 640K
- Calculate total main memory using better method
- Turn on or off NMI and Parity check at startup?
- Turn on or off NMI and Parity check when XMS/A20 enabled?
- Use TVRAM address (high resolution?) for some systems
- Dynamically select FD1232 or FD1440 drive in setup.S code without ifdefs

Given all the changes, perhaps good idea to separate some changes into separate PRs? I won't have the ability to test some of these changes, and can only use DosBox-X for testing, not Neko or real hardware. It would also be nice to know the specific PC-98 model numbers that support is being added for.

As discussed in #2414, I like the idea of removing CONFIG_IMG_xxx ifdefs if possible, although some remain in this PR. It would be nice to use a single callable routine that the kernel or setup.S can use to determine which floppy format is supported.

Does the NMI/Parity change fix @tyama501's problem of system hang when running hma=kernel and XMS LOADALL?

Thank you!

PR #2416, opened 2025-10-22
[reset] Adding reboot function by shutdown -r to PC-98

tyama501:
Hello @ghaerr and @drachen6jp 

This is the PR to add the shutdown -r reboot function for PC-98 by @drachen6jp .

I have cherry-picked his commit, then I have added reset-pc98.c and moved the function from the reset-stub.c.

I have confirmed it work with NP2 emulator.

Thank you!

### DIFF
--- a/elks/arch/i86/kernel/Makefile
+++ b/elks/arch/i86/kernel/Makefile
@@ -44,7 +44,7 @@ OBJS += divzero.o
 endif
 
 ifeq ($(CONFIG_ARCH_PC98), y)
-OBJS += irq-8259.o reset-stubs.o timer-8254.o
+OBJS += irq-8259.o reset-pc98.o timer-8254.o
 endif
 
 ifeq ($(CONFIG_ARCH_8018X), y)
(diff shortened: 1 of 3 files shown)


ghaerr:
Hello @tyama501,

Is there a chance you could rewrite using more readable C macros, something similar to reset-swan.c? This would allow the possible use of additional defines to use as port numbers or data constants as self-documentation. This would allow much easier understanding of what is being done. Something like:
```
#define PORT_CTRL  0x37  /* control port for ... */
#define PORT_xxxx  0xF0  /* special port for ... */

    outb(0x0B, PORT_CTRL);    /* send ... to ... control */
    outb(0x0F, PORT_CTRL);
    if (inb(0xF0) & 2)       /* V30 */
        outb(0x07, 0xF0);    /* send .. to ... control */
    else outb(0x00, 0xF0);
    asm("ljmp $0xFFFF,$0\n");
```
I realize that reset-ism.c is written all ASM, but the reason for that is that INT 15h is executed, for which there are no predefined macros to load registers beforehand. 

Thank you!

tyama501:
Hello @ghaerr 

I have rewritten it.

Thank you.

ghaerr:
Nice, thank you!

PR #2419, opened 2025-10-24
[bios] Supporting FD1200 mount on PC-98

tyama501:
Hello @ghaerr and @drachen6jp 

This PR is based on the bios.c code of @drachen6jp that I have cherry-picked to mount FD1200 on PC-98.

I have reverted ifdefs and some modifications that are not related to FD1200, 512Byte/sector, device address 90 feature,
since we are in the release phase of elks 0.9 and
I think it is better to postpone heavy modifications until the release.
(I might have reasons for the ifdefs. One of those are to indicate the kernel(boot) is compiled for 1232K or 1440K on the boot messages.)

I have added FORCE_PROBE = 1 in the bioshd.c and some minor changes,
since I have found bios_switch_device98 with the probe always needed when FD1232 <-> FD1200.

I have confirmed on the NP2.
image

To boot from FD1200, further changes to boot_sect and Setup are needed.
I will try to separate the FD1200 feature of those from other enhancement @drachen6jp made.

Thank you!

### DIFF
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -47,11 +47,7 @@ struct drive_infot fd_types[] = {   /* AT/PS2 BIOS reported floppy formats*/
 #ifdef CONFIG_ARCH_PC98
 unsigned char bios_drive_map[MAX_DRIVES] = {
     0xA0, 0xA1, 0xA2, 0xA3,             /* hda, hdb */
-#ifdef CONFIG_IMG_FD1232
     0x90, 0x91, 0x92, 0x93              /* fd0, fd1 */
-#else
-    0x30, 0x31, 0x32, 0x33              /* fd0, fd1 */
-#endif
 };
 #else
 unsigned char bios_drive_map[MAX_DRIVES] = {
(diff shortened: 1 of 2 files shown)


ghaerr:
Thank you!

> I might have reasons for the ifdefs. One of those are to indicate the kernel(boot) is compiled for 1232K or 1440K on the boot messages.

Hopefully it will not matter if the kernel is compiled for 1232K or 1440K (since there would be no ifdefs) - the kernel could work with either drive type, depending on what it finds in the drive or system. The kernel could also display a message at boot or probe time alerting the user to which type of floppy is found, if that is what is wanted?

tyama501:
Hello @ghaerr 

Thank you.

The reason why I want to keep indicating the difference of 1232k and 1440k on boot message is that currently 1440k boot sector 512B is too tight to be common with 1232k version. (It seems that @drachen6jp showed 1232k can use 1K boot sector space but on the pc-98 real hardware 1440k can only use 512Byte)

tyama501:
Oh I understand what you mentioned.
Yes it can be found in the probe time.

ghaerr:
Hello @tyama501,

> The reason why I want to keep indicating the difference of 1232k and 1440k on boot message is that currently 1440k boot sector 512B is too tight to be common with 1232k version.

I see. So a user needs to be able to determine whether the kernel matches the boot sector for PC-98 1232K vs 1440K floppy images? 

> It seems that @drachen6jp showed 1232k can use 1K boot sector space but on the pc-98 real hardware 1440k can only use 512Byte

If the smaller 512-byte boot sector is completely full, then, yes, we may need a special boot sector for 1232K. I would still prefer the PC98 kernel to work dynamically with any floppy image. If only boot sectors are different, then it may be possible that the `makeboot` program could copy an appropriate PC98 boot sector from /lib or compiled into itself to the target image, if required. 

If the kernel is also special-cased per floppy image type, then it won't be easily possible to create a new bootable system image unless the current system image matches the target device, which may cause problems between 1232K floppies and 512-byte HD images.

Getting back to "floppy type displayed at boot", perhaps that isn't needed as much as having a way the to make `sys` problem just "work" without user having to worry about it. The system /lib directory could contain different boot sectors for PC-98 and sys/makeboot modified take are of it automatically or with a command line option.

What are your thoughts?

Thank you!

tyama501:
That is good idea but many PC-98 (unofficially) 
support FD1200 512B sector, so maybe we don't need that immediately.

I think I can remove ifdefs (again),
I will do that later.
Thank you!

PR #2420, opened 2025-10-25
[boot] Add FD1200 boot for PC-98

tyama501:
Hello @ghaerr and @drachen6jp 

Just with a single line modification, I could boot it from FD1200.

image

It is nice if we can have fd1200 image for PC-98 when ELKS 0.9 release,
since  fd1200 and hdd can have common boot code, so that it can install to the hdd just using sys without makeboot fat.bin.
(I have confirmed it)

EDIT: old PC-98 cannot use fd1440.

Thank you.


ghaerr:
> It is nice if we can have fd1200 image for PC-98 when ELKS 0.9 release,

I will add that to the standard floppy distribution image set.

Thank you!

PR #2422, opened 2025-10-26
[build] Add PC98 1200K floppy build to buildimage.sh standard distribution

ghaerr:
Requested by @tyama501 in #2420.

The PC98 1200K (compressed) floppy distribution includes applications in elkscmd/Applications with :1200k and :1200c tags.


tyama501:
Thank you!

PR #2423, opened 2025-10-28
[cmds] Minor fixes to Paint

ghaerr:
While porting some of the Paint graphics code to [GFX](https://github.com/ghaerr/gfx), it was noticed that when drawing lines in Paint, the first point is always drawn twice and the last point never drawn. Also, the draw rectangle routine draws the corners twice. This PR fixes that, for @Vutshi's perfection!

A bug in the IA16 and C86 versions for drawvline is also fixed, when the end point is previous to the starting point.


Vutshi:
Great to see Paint getting closer to perfection!

Someday I’ll be back with a path-tracing 3D renderer for ELKS and the good old 8088. 😄

PR #2425, opened 2025-10-28
[cmds] Build fsck-dos.os2 and paint.os2 when WATCOM environment installed

ghaerr:
This is the first step in automatically building some OpenWatcom applications already in the ELKS applications tree.

When OpenWatcom C has been configured for ELKS by executing the `libc/wcenv.sh` script, the WATCOM= environment variable is set and the OWC tools are added to PATH.

When WATCOM is defined, an enhanced FAT filesystem checker is built using OWC huge model `fsck-dos.os2`, which allows for checking FAT16 HD images without running out of memory, and an experimental `paint.os2` OWC version of Paint is built, for testing graphics using OWC. In addition, when C86 is defined from executing `c86env.sh` in 8086-toolchain, the C86 version of Paint `paint.c86` is built, testing the C86 compiler and development library by building a larger application. 

All these programs are copied directly to /bin in the target image (elkscmd/rootfs_template/bin). The next step will be integrating external Nano-X into the build process.

### DIFF
--- a/elkscmd/fsck_dos/.gitignore
+++ b/elkscmd/fsck_dos/.gitignore
@@ -1,1 +1,2 @@
 fsck-dos
+fsck-dos.os2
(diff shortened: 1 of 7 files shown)


toncho11:
Great for the Nano X builds. It is needed for release 9.

PR #2427, opened 2025-10-31
Remove ifdef CONFIG_IMG_FD1232 in bios.c again

tyama501:
Hello @ghaerr and @drachen6jp ,

I have removed ifdef CONFIG_IMG_FD1232 in bios.c again.
I have also added some clear for CX and DX before call bios.

I now trying to remove ifdef CONFIG_IMG_FD1232 in setup.S with the similar way @drachen6jp did in his elks_hirezo98 fork,
using int1B, service 0x5A, but I'm getting trouble booting. I need some more time.
It might be related to moving codes to the architecture specific region in the file but I'm not sure yet.

Thank you.

### DIFF
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -91,6 +91,8 @@ int BFPROC bios_disk_rw(unsigned cmd, unsigned num_sectors, unsigned drive,
     else {
         if ((0xF0 & drive) == 0x90) {
             BD_AX = 0x5a00|drive;
+            BD_CX = 0;
+            BD_DX = 0;
             call_bios(&bdt);
             BD_AX = cmd | drive;
             if((BD_CX & 0x300)==0x200) goto notMFM1024;
@@ -420,6 +422,8 @@ void BFPROC bios_switch_device98(int target, unsigned int device,
         *drivep = fd_types[FD720];
     else if (device == 0x90) {
         BD_AX = 0x5a00|device|(bios_drive_map[target + DRIVE_FD0] & 0x0F);
+        BD_CX = 0;
+        BD_DX = 0;
         call_bios(&bdt);
         if((BD_CX & 0x300)==0x300)
          *drivep = fd_types[FD1232];
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -313,9 +313,13 @@ int INITPROC bios_getfdinfo(struct drive_infot *drivep)
 #ifdef CONFIG_ARCH_PC98
     for (drive = 0; drive < 4; drive++) {
         if (peekb(0x55C,0) & (1 << drive)) {
-            bios_drive_map[DRIVE_FD0 + drive] = drive + 0x90;
-            *drivep = fd_types[FD1232];
-
+            if ((peekb(0x584,0) & 0xF0) == 0x30) {
+                bios_drive_map[DRIVE_FD0 + drive] = drive + 0x30;
+                *drivep = fd_types[FD1440];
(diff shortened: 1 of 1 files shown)


ghaerr:
> I need some more time.

That's ok - would you like this PR committed now, or should it wait?

> I have removed ifdef CONFIG_IMG_FD1232 in bios.c again.

Does this code still work with 1440k drives? I don't see how *drivep is updated for 1440k drives? It seems there may want to be a single place (a new function or bios init code?) where the drive types and bios_drive_map are all updated to the proper values for either 1200k, 1232k or 1440k. Is this possible?

tyama501:
Hmm, looking at the code, it may get trouble in the bios_conv_bios_drive() to get the correct minor number to boot FD1440 from fd other than fd0. Is this your concern?
I don't have FD1440 bootable pc98 since it is broken so I can only check on the emulator but I will check the case to boot from fd1 later.

Thank you.

elks/arch/i86/drivers/block/bios.c
```diff
@@ -322,13 +313,9 @@ int INITPROC bios_getfdinfo(struct drive_infot *drivep)
...
     for (drive = 0; drive < 4; drive++) {
         if (peekb(0x55C,0) & (1 << drive)) {
-#ifdef CONFIG_IMG_FD1232
             bios_drive_map[DRIVE_FD0 + drive] = drive + 0x90;
             *drivep = fd_types[FD1232];
-#else
-            bios_drive_map[DRIVE_FD0 + drive] = drive + 0x30;
-            *drivep = fd_types[FD1440];
-#endif
+
```
ghaerr: > Is this your concern?

I guess I don't understand the code well enough? My concern is that previous to this change, bios_drive_map[] is set to 0x90... for 1232K drives, and 0x30... otherwise (lines 51-53). But with this change above, it appears that bios_drive_map[] is always set to 0x90... and never to 0x30... 

I don't know exactly what the byte at 0:0x55C details, but it seems that it indicates whether a drive is 1232K or not, is that correct? If so, then shouldn't the above code at line 329 be an else statement that sets the drive type to 0x30... and 1440k? Or does the byte just indicate whether a floppy drive is present or not? If that is the case, then where does the new code set bios_drive_map[] properly for whatever drive type is installed on each PC98 floppy?

Also, I don't understand when the PC98 BIOS requires 0x90 vs 0x30 for FD1200 disks. Is this documented somewhere in the code, or in the ELKS project for reference?

tyama501: I think your concern and my concern are the same.
I will fix the getfdinfo.

For FD1200 the same device address as FD1232 is used which is 0x90.
That is why @drachen6jp used int 1B service 0x5A to get sector size information.
Thank you.

tyama501:
Hello 

I have modified bios_getfdinfo.
I have confirmed FD1440 can boot it from fd1 after the modification.

Before the modification
Screenshot from 2025-11-02 01-01-54

After the modification
Screenshot from 2025-11-02 01-37-43

I will make another PR if setup.S is worked.

Thank you.

ghaerr:
> I have confirmed FD1440 can boot it from fd1 after the modification.

Thank you, I now understand the code much better.

I can commit now, but do you think that the lines 292-295 (presetting drivep[0-4]) are still necessary? It seems that in your new `for` loop, all four drives are queried and drivep set again to the values required. It looks like they could be removed. You can do this in another PR if required.

ghaerr:
> but do you think that the lines 292-295 (presetting drivep[0-4]) are still necessary? 

Oops - my mistake - I see that these lines are necessary in the case BFD_HARD case where drives are hard-set to default values. No change needed!

PR #2428, opened 2025-10-31
Make USER_FLAGS a config parameter

swausd:
USER_FLAGS is used in process.c/arch_setup_user_stack() to prepare the processor status word (PSW) for a context switch. Currently bits 15-14 are set to 0 which conflicts with the NEC V25 CPU. To make this configurable USER_FLAGS is moved to config.h

see https://github.com/ghaerr/elks/issues/2424

### DIFF
--- a/elks/arch/i86/kernel/process.c
+++ b/elks/arch/i86/kernel/process.c
@@ -88,9 +88,6 @@ void kfork_proc(void (*addr)())
  *  Build a user return stack for exec*(). This is quite easy,
  *  especially as our syscall entry doesnt use the user stack.
  */
-
-#define USER_FLAGS 0x3200               /* IPL 3, interrupt enabled */
-
 void put_ustack(register struct task_struct *t,int off,int val)
 {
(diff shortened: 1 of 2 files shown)


ghaerr:
Hi @swausd,

Thank you for this PR. As discussed in https://github.com/ghaerr/elks/issues/2424#issuecomment-3458078109, lets leave the USER_FLAGS in its original location; just change the value to 0xF200 and I think we'll be good to go for all systems. I can then commit this.

The top-level config file isn't really the place for this kind of thing, and after studying the NEC documentation, I can see they've worked hard to maintain 8086 compatibility by only using unused bits in the PSW.

swausd:
OK. Thank you!

swausd:
Can be closed then...

ghaerr:
> I created a fork of your repository and will try (never did this before) to setup a pull request with this change as a first step.

BTW, you can just make the change in the checked-out branch and do another "git push" and things should update properly over here.

ghaerr:
> Can be closed then...

If you haven't deleted your branch, just re-open this issue, make the change, then git push. This should work.

PR #2429, opened 2025-10-31
Change USER_FLAGS so it is compatible with NEC V25 microcontroller

swausd:
USER_FLAGS is used in process.c/arch_setup_user_stack() to prepare the processor status word (PSW) for a context switch. Currently bits 15-14 are set to '0' which conflicts with the NEC V25 CPU (switches to undefined register bank on next contrext switch). As these bits are not used for other CPUs it can be set to '1'.

Refernce: https://github.com/ghaerr/elks/issues/2424


ghaerr:
Thank you!

PR #2440, opened 2025-11-04
[cmds] Enhance ibmpc BASIC to use EGA or VGA graphics mode

ghaerr:
Enhances IBM PC version of ELKS `basic` to use 16-color EGA 640x350 graphics (when EGAMODE environment variable set), otherwise use VGA 640x480. This allows programs such as elkscmd/basic/test98.bas (originally written for PC-98) to also run on IBM PC BASIC.

The 16 colors are still not compatible between IBM PC and PC98, see screenshots below.

Also replaces host_plot() function with the much smaller draw_line() routine from ELKS Paint, which reduces the size of `basic` by 1K bytes on disk.

IBM PC `basic test98.bas`:
IBMPC basic

PC-98 `basic test98.bas`:
PC98 basic

Note that the PC-98 allows ASCII text to display in graphics mode (see "BASIC" and "260-STOP" in above screenshot), while the IBM PC EGA/VGA does not. There isn't much that can be done about this. Typing ^D to exit basic, or "mode 0" to exit graphics mode, will revert to EGA/VGA text mode.

### DIFF
--- a/elkscmd/basic/host-ibmpc.c
+++ b/elkscmd/basic/host-ibmpc.c
@@ -72,10 +72,8 @@ void host_mode(int mode) {
 
 void host_cls() {
 
-    if (gmode) {
-        fmemsetw(0, 0xB800, 0, 4000);
-        fmemsetw(0, 0xBA00, 0, 4000);
-    }
+    if (gmode)
+        fmemsetw(0, 0xA000, 0, 19200);  /* 640*480/8 bytes VGA ram */
     else
         fprintf(outfile, "\033[H\033[2J");
 }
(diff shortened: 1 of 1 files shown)


toncho11:
So no CGA for IBM PC?

ghaerr:
> So no CGA for IBM PC?

I'm thinking of extending the `MODE n` command to accept a BIOS graphics mode number, with MODE 1 meaning default (EGA or VGA). That would easily allow CGA operation. The bigger problem I've found is that it seems IBM PC CGA, EGA or VGA programs aren't compatible with PC-98, and don't display properly. Not sure what to do about that, and it doesn't help that BASIC uses a flipped Y-coordinate axis.

toncho11:
I see - complicated. I am updating here https://github.com/ghaerr/elks/wiki/Coding-games-for-ELKS#3-using-basic - the Basic section when there are new changes,

PR #2444, opened 2025-11-07
[cmds] Remove build directory from Makefile

BalakeKarbon:
elkscmd/Makefile no longer references the build directory, since Makefile.owc does not exist in elkscmd/build. This fixes errors that were preventing the build.


ghaerr:
Thanks for point this out, I had forgotten to push the new basic/Makefile.owc. Do a git pull and things should work now.

BalakeKarbon:
Thank you I was not sure what was intended.

PR #2446, opened 2025-11-08
[kernel] Issue BIOS disk reset only when ATA CF driver configured

ghaerr:
Discussed in https://github.com/ghaerr/elks/issues/2398#issuecomment-3506754501 as likely problem for new BIOS retry errors on PC-98 system.

Should fix original PR #2411 by only resetting BIOS using bios_drive_map[] when **both** ATA CF driver and BIOS HD driver present. While still not optimal, this should fix PC98 issue with BIOS retries since ATA CF driver is not present on PC98.

It appears the original fix #2411 may not work on all systems, more testing on real hardware system may be required.


ghaerr:
Closing as not needed.

PR #2453, opened 2025-11-13
Support for hardware SPI on NEC V25 port

swausd:
ssd-sd block driver now supports hardware SPI  on NEC V25 MCU or software SPI (bit-banged) on 8018x and NEC V25. Added new option `Interface SD card with hardware SPI` in `make menuconfig` if `NEC V25` system is selected. The bit banged driver should only be used if the serial port 0 is not available.

The new hardware driver use SPI Mode 3 (hardware limit of V25) on serial port 0. Despite SD cards usually require SPI mode 0, the hardware driver in mode 3 works with all the different SD cards i have (3 different brands, two noname and sandisk compact extreme).

Clock is configurable by setting prescaler value in the driver code spi-hw-necv25.S. Default CPU `clock / 2 `which is the max clock usable. For my current system that is 5 MHz. With the bit-banged driver it is about 128KHz.

I introduced two new assembler functions for block read/write and changed ssd-sd.c accordingly.
Minix filesystem should be used as this is much faster than FAT16.

For connecting the SD card and port usage see spi-hw-nev25.S and sp-necv25.S.


ghaerr:
Thank you @swausd, fantastic work as always! This is very nice having both software and hardware SPI drivers in ELKS! :)

PR #2454, opened 2025-11-14
romprg for the NEC V25 port

swausd:
Minor changes to the nice 8018x romprg code from @cocus to get it working on my NEC V25 system.  Just a "quick and dirty" job, as there is only a small audience for this tool. Made a "how to" for my own reference. May be helpful for others too, so I added it. I can verify that the original version can be used unmodified for SST39SF040 flash ROMs.


ghaerr:
Thanks for testing out this utility and porting it to another platform along with some bug fixes. Cool!

elkscmd/romprg/main.c
```diff
@@ -50,16 +53,84 @@ static const unsigned int baudrate_compares[] = {
     (CONFIG_8018X_FCPU * 1000000UL / (500000UL * 8UL)) - 1, /* X5 = B500000 */
 };
 
+void set_baud_rate(uint8_t idx) {
+    outw(0x8000 | baudrate_compares[idx], PCB_B0CMP); // 38400bps
```
cocus, line 57: due to the fact that this is being called with an index of 17, the 38400bps is not true anymore. I think it has jumped to 115200 as well as on the V25

swausd: @cocus, yes you are right. I use 115200 baud. The comment should be deleted.

cocus: that's fine :) I never tried that baudrate, not sure why though. probably the error rate was high. I forgot :)

swausd: No errors for me. I use 20 cm to a TTL to USB converter.

cocus: no no, I meant the error rate due to the baudrate being too off from the actual 115200, due to the crystal I've used on my SBC

swausd: That's why I use 22.1184 MHz

ghaerr: @cocus, are you saying the 8018X (and possibly V25) don't map to the baud rate indices setup by termios.h in ELKS? 

The IBM PC serial handler in serial-8250.c sets the baud rate using a divisor, and takes two steps to compute a baud rate index:
```
    /* set baud rate divisor, first lower, then higher byte */
    cflags = port->tty->termios.c_cflag & CBAUD;
    if (cflags & CBAUDEX)
        cflags = B38400 + (cflags & 03);
    divisor = divisors[cflags];
```
The termios.h value for B38400 is 0000017, which is actually 15 decimal. 17 decimal would correspond to B115200.

cocus: right, that's probably why I couldn't get 115200 reliable on mine, I've used a 20MHz crystal

cocus: > @cocus, are you saying the 8018X (and possibly V25) don't map to the baud rate indices setup by termios.h in ELKS?
> 
> The IBM PC serial handler in serial-8250.c sets the baud rate using a divisor, and takes two steps to compute a baud rate index:
> 
> ```
>     /* set baud rate divisor, first lower, then higher byte */
>     cflags = port->tty->termios.c_cflag & CBAUD;
>     if (cflags & CBAUDEX)
>         cflags = B38400 + (cflags & 03);
>     divisor = divisors[cflags];
> ```
> 
> The termios.h value for B38400 is 0000017, which is actually 15 decimal. 17 decimal would correspond to B115200.

Right, but in [here](https://github.com/ghaerr/elks/pull/2454/files#diff-5d7a28fba742da907442c4e488c207b29df336d08f191a534a63a080565d7b85R133) decimal 17 is being used. 15 was used before

swausd: added a pull request to change the comment...

cocus: thank you! it was a small nit pick :) It'll probably bite me in the future

PR #2455, opened 2025-11-14
Changed erronous comment to make clear 115200 baud are used

swausd:
@ghaerr 
changed the wrong comment and added a correct one. Sorry for the error.


ghaerr:
Discussed in #2454.

PR #2459, opened 2025-11-16
[kernel] Move CONFIG_FAST_IRQ4 to config.h

ghaerr:
Moves CONFIG_FAST_IRQ4 and CONFIG_FAST_IRQ3 out of ports.h into config.h.

Cleanup of various files associated with fast serial handler. 

Doesn't (yet) add TTY signal processing, as discussed in https://github.com/ghaerr/elks/issues/2457#issuecomment-3538899648. Since the fast serial handler runs on any stack and DS != SS, calling most kernel functions from the handler isn't easily possible; still working on a good solution.

### DIFF
--- a/elks/arch/i86/drivers/char/serfast.S
+++ b/elks/arch/i86/drivers/char/serfast.S
@@ -8,7 +8,6 @@
 // 25 June 2020 Greg Haerr
 //
 #include <linuxmt/config.h>
-#include <arch/ports.h>
 	.code16
 	.text
 
@@ -41,7 +40,7 @@ _irq_com1:
 	pop	%cx
 	pop	%bx
 	pop	%ax
-	add	$4,%sp		// skip the trampoline DS:*irq
+	add	$4,%sp			// skip the trampoline DS:*irq
 	iret
 #endif
 
@@ -74,6 +73,6 @@ _irq_com2:
 	pop	%cx
 	pop	%bx
 	pop	%ax
-	add	$4,%sp		// skip the trampoline DS:*irq
+	add	$4,%sp			// skip the trampoline DS:*irq
 	iret
 #endif
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -206,39 +206,17 @@ static int rs_write(struct tty *tty)
     return i;
 }
 
-#if defined(CONFIG_FAST_IRQ4) || defined(CONFIG_FAST_IRQ3)
-/* called from timer interrupt - check ring buffer and wakeup waiting processes*/
-void rs_pump(void)
-{
-    struct serial_info *sp;
-    struct ch_queue *q;
-
-#ifdef CONFIG_FAST_IRQ4
-    sp = &ports[0];
-    q = &sp->tty->inq;
-
-    if (sp->tty->usecount && q->len)
-        wake_up(&q->wait);
-#endif
-#ifdef CONFIG_FAST_IRQ3
-    sp = &ports[1];
-    q = &sp->tty->inq;
-
-    if (sp->tty->usecount && q->len)
-        wake_up(&q->wait);
-#endif
-}
-#endif
-
-#ifdef CONFIG_FAST_IRQ4
+#if defined(CONFIG_FAST_IRQ4)
 /*
  * Fast serial driver for slower machines. Should work up to 38400 baud.
- * No ISIG (tty signal interrupt) handling for shells, used for fast SLIP transfer.
+ * No ISIG (tty signal interrupt) handling for shells, used for fast
+ * SLIP transfer or arrow key input on slow systems.
  *
- * Specially-coded fast C interrupt handler, called from asm irq_com[12] after saving
- * scratch registers AX,BX,CX,DX & DS and setting DS to kernel data segment.
- * NOTE: no parameters can be passed, nor any code written which
- * emits code using SP or BP addressing, as SS is not set and not guaranteed to equal DS.
+ * Specially-coded fast C interrupt handler, called from asm _irq_com1/2 after
+ * saving scratch registers AX,BX,CX,DX & DS and setting DS to kernel data segment.
+ * NOTE: no parameters can be passed, nor any code written which emits code using
+ * SP or BP addressing, as SS is not set and not guaranteed to equal DS.
+ *
  * Use 'ia16-elfk-objdump -D -r -Mi8086 serial.o' to look at code generated.
  */
 extern void _irq_com1(int irq, struct pt_regs *regs);
@@ -259,7 +237,7 @@ void fast_com1_irq(void)
 }
 #endif
 
-#ifdef CONFIG_FAST_IRQ3
+#if defined(CONFIG_FAST_IRQ3)
 extern void _irq_com2(int irq, struct pt_regs *regs);
 void fast_com2_irq(void)
 {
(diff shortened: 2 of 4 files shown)


ghaerr:
@Mellvik 
I added the capability to check for SIGINT while running the fast serial interrupt handlers. It's limited to having to assume that ^C = VINTR due to the handler running DS != SS and not being able to allow the compiler to generate any data references using BP, since BP uses the SS stack segment. Nonetheless, it appears to work well and allows you to use ^C at the shell prompts which is very useful.

Here's the diff, you might want to try this on your real hardware with CONFIG_FAST_IRQ4. I recommend only running one fast serial handler for speed considerations because of the rs_pump requirement every 10ms. 
```
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -28,7 +28,8 @@ struct serial_info {
     unsigned char mcr;
     unsigned int  divisor;
     struct tty *tty;
-    int pad1, pad2, pad3;       // round out to 16 bytes for faster addressing of ports[]
+    int intrchar;       /* used by fast handler for ^C SIGINT processing */
+    int pad2, pad3;     /* round out to 16 bytes for faster addressing of ports[] */
 };
 
 /* flags*/
@@ -234,6 +235,9 @@ void fast_com1_irq(void)
             q->head = 0;
         q->len++;
     }
+    /* unfortunately, can't add more specifics w/o compiler generating BP accesses */
+    if (c == 03)                        /* must assume VINTR = ^C */
+        sp->intrchar = c;
 }
 #endif
 
@@ -261,12 +265,21 @@ void fast_com2_irq(void)
 void rs_pump(void)
 {
     struct serial_info *sp;
+    struct tty *ttyp;
     struct ch_queue *q;
 
 #if defined(CONFIG_FAST_IRQ4)
     sp = &ports[0];
     q = &sp->tty->inq;
 
+    if (sp->intrchar) {
+        ttyp = sp->tty;
+        if ((ttyp->termios.c_lflag & ISIG) && ttyp->pgrp) {
+            if (sp->intrchar == ttyp->termios.c_cc[VINTR])
+                kill_pg(ttyp->pgrp, SIGINT, 1);
+        }
+        sp->intrchar = 0;
+    }
     if (sp->tty->usecount && q->len)
         wake_up(&q->wait);
 #endif
@@ -274,6 +287,14 @@ void rs_pump(void)
     sp = &ports[1];
     q = &sp->tty->inq;
 
+    if (sp->intrchar) {
+        ttyp = sp->tty;
+        if ((ttyp->termios.c_lflag & ISIG) && ttyp->pgrp) {
+            if (sp->intrchar == ttyp->termios.c_cc[VINTR])
+                kill_pg(ttyp->pgrp, SIGINT, 1);
+        }
+        sp->intrchar = 0;
+    }
     if (sp->tty->usecount && q->len)
         wake_up(&q->wait);
 #endif
```
Let me know if you'd like to see this posted as a PR, thanks!

Mellvik:
Wow @ghaerr - this is great, much appreciated. You're racing ahead like an untamed mustang leaving the donkeys (that includes me for the time being) in the dust, facing an impossible catch-up challenge :-). 

I'll dive into this tomorrow.

Thanks.
M

Mellvik:
@ghaerr, 
this looks good. No champagne but linenoise editing now works fine, at least for the duration of this short test, still on the XT/8MHz - 19200 bps.

Pasting into the window is not perfect but improved - and very confusing in that the data loss is baud rate independent. Exactly the same bytes are lost at 19200 and 1200 bps - and rates between. Only at rates above 19200, the pattern changes (more lost bytes). Is this what you expected?

```
v20-16# cat > xx
From 10.0.2.2 icmpseq=124 Destinatio Host Unreachable
rom 10.0.2.2 icmp_eq=125 Destination ost Unreachable
Frm 10.0.2.2 icmp_se=126 Destination Hst Unreachable
From10.0.2.2 icmp_seq=27 Destination Hos Unreachable
From 0.0.2.2 icmp_seq=18 Destination Host nreachable
From 100.2.2 icmp_seq=129Destination Host Ureachable
From 10..2.2 icmp_seq=130 Dstination Host Unrachable
From 10.0..2 icmp_seq=131 Detination Host Unrechable
From 10.0.2. icmp_seq=132 Destnation Host Unreacable
From 10.0.2.2icmp_seq=133 Destiation Host Unreachale
From 10.0.2.2 imp_seq=134 Destinaion Host Unreachabe
From 10.0.2.2 icp_seq=135 Destinatin Host UnreachableFrom 10.0.2.2 icmpseq=136 Dest
inatio Host Unreachable
Fom 10.0.2.2 icmp_sq=137 Destination ost Unreachable
Frm 10.0.2.2 icmp_se=138 Destination Hot Unreachable
From10.0.2.2 icmp_seq=39 Destination Hos Unreachable
From 0.0.2.2 icmp_seq=14 Destination Host nreachable
From 100.2.2 icmp_seq=141Destination Host Ureachable
From 10.02.2 icmp_seq=142 Dstination Host Unrachable
From 10.0..2 icmp_seq=143 Detination Host Unreahable
From 10.0.2. icmp_seq=144 Destnation Host Unreacable
From 10.0.2.2icmp_seq=145 Destintion Host Unreachale
From 10.0.2.2 imp_seq=146 Destinaion Host Unreachabl
From 10.0.2.2 icm_seq=147 Destinatin Host Unreachable
```

^C works as expected - and I'm fine with this, this is a great functional improvement for serial on slow machines.

ghaerr:
Thank you @Mellvik for the testing!! Is this from ELKS or TLVC?

> No champagne but linenoise editing now works fine

> ^C works as expected - and I'm fine with this, this is a great functional improvement for serial on slow machines.

Very good to know! Glad it's all still working, I wrote all this many years ago for the SLIP driver on my Compaq Portable. Yesterday I wrote the signal processing, glad that's working too.

> very confusing in that the data loss is baud rate independent. Exactly the same bytes are lost at 19200 and 1200 bps
> Only at rates above 19200, the pattern changes (more lost bytes). Is this what you expected?

Yes, this fits my mental model. I'm thinking that what you're seeing is that the fast serial IRQ handler is stuffing all the bytes into its large buffer up to 19200 baud, with no receive data overrun. But remember we still have the IRQ 0 and possibly some other interrupts that occur that keep interrupts disabled for too long (in this case, likely for almost two received characters) due to exactly how the UART input mechanism works. So when that long interrupt disable naturally occurs, you're seeing a regular data loss that is independent of baud rate. This IMO also confirms why you saw so much data loss at any baud rate previously - the serial IRQ4 interrupt was taking way too long, period.

When the baud rate is 38400+, it seems that even the low overhead of the fast serial driver has become too much - and data is lost.

The reason the fast serial driver works well with no other modifications is because the serial driver was written long ago to assume that it might be taking input from SLIP or pasted data, so it allocates IIRC 1024 bytes. The key is to have a large enough buffer so that it can gulp down all the streaming input with no processing, then process it all later after the interrupt barge has completed. While dedicated 1k bytes to a serial input queue is excessive for normal non-SLIP use, I struggled with how to implement this without yet more /bootopts options etc, so just left it as is - and things just work.

All the more reason to take a very hard look at all device drivers, and disable interrupts ONLY when absolutely necessary. We can discuss each case individually if you want - just open a TLVC issue and let me know.

Thanks for your quick testing and glad to hear this is resolved. I'll have to update the Wiki to let people know how to configure their slower systems.

PR #2462, opened 2025-11-16
[net] Move various WD NIC driver routines to far NICPROC

ghaerr:
Saves 1104 bytes of kernel near (main) code segment by moving some WD NIC driver routines to new NICPROC section.

Not tested, but corresponding changes to code generated were inspected for correctness. In general, ia16-elf-gcc handles near <-> far calling very well, one just needs to be aware of when trampoline code is generated, which is when FAR routines call NEAR routines. When NEAR calls FAR, or FAR cals FAR, no trampoline code is executed.

@Mellvik, this should be straightforward to add to TLVC as well.

### DIFF
--- a/elks/arch/i86/drivers/net/wd.c
+++ b/elks/arch/i86/drivers/net/wd.c
@@ -212,11 +212,11 @@ static unsigned int verbose;
 static unsigned char current_rx_page;
 static struct netif_stat netif_stat;
 
-static word_t wd_rx_stat(void);
-static word_t wd_tx_stat(void);
+static word_t NICPROC wd_rx_stat(void);
+static word_t NICPROC wd_tx_stat(void);
 static void wd_int(int irq, struct pt_regs * regs);
-static void fmemcpy(void *, seg_t, void *, seg_t, size_t, int);
-static void wd_stop(void);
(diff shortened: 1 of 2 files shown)


Mellvik:
That's interesting, @ghaerr. What's your estimate on the price of this change in terms of performance?

Thanks.

ghaerr:
> What's your estimate on the price of this change in terms of performance?

I would estimate almost-zero. We're only talking about the overhead of a two-byte longer LCALL instruction and the overhead of PUSH CS on execution, when calling NEAR to FAR or FAR to FAR. When calling FAR to NEAR, there's the trampoline overhead which involves a bit more execution time by a slight rearrangement of the stack so that the NEAR called procedure sees the arguments at the correct offset. But in the WD driver, I heavily looked through the code generated, and every single (and the only) FAR to NEAR calls were to printk! How slow is printk compared to a NICPROC function call? I would estimate 500x slower.

Also, FAR calls aren't running with disabled interrupts so interrupts aren't slowed down in the slightest. Given the size of an ETH packet (compared to say, a received UART single byte) I would say that you couldn't ever tell the difference using NICPROC, other than having tons of NEAR kernel code space made available for other better reasons.

Changing NICPROC here lowered my kernel from 65200 bytes to 64096!

Conclusion: use NICPROC everywhere you possibly can.

PR #2466, opened 2025-11-17
[net] Fix tcp bad checksum error on localhost connections

### LINKED ISSUE #2465: Telnet connection to localhost displays "BAD CHECKSUM" error
Started seeing these TCP error messages after (possibly for the first time since ktcp was updated) connecting to localhost using telnet. Since `telnetd` was recently updated, I reverted that commit with no change.

These errors occur immediately after connection, then max retries are exceeded and the connection is closed on the first character typed. This only occurs on local connections, I am not seeing it happen when connecting outside the QEMU network to the host.

@Mellvik, any ideas as to why this might be? I can roll back the ktcp modifications, but thought you'd probably have a better initial idea of why this might be happening.

Screenshot on QEMU:

Image

ghaerr:
Fixes #2465.

Was able to track down the problem: the following now working code statements in tcp_output.c were swapped in the TLVC version of ktcp:
```
    add_for_retrans(cb, th, len, &apair);
    ip_sendpacket((unsigned char *)th, len, &apair, cb);
```
I'm not exactly sure why this works, but I had noticed this swap in the TLVC ktcp enhancement, and not knowing the reason for it. I swapped the statements back to how I had originally witten them and the TCP checksum error went away. The debug output was showing that certain TCP packets were not getting ACKed. I think this might have something to do with the TCP or IP buffers possibly being written into by ip_sendpacket (e.g. checksum?) that was causing problems with add_for_retrans?

What do you think @Mellvik, what was the reason for swapping these statements in TLVC?

BTW, you might try `telnet localhost` on TLVC to see if that works, I would bet it might not.

Also fixes some debug printfs that were using long format descriptors for unsigned int values.

### DIFF
--- a/elkscmd/ktcp/tcp_output.c
+++ b/elkscmd/ktcp/tcp_output.c
@@ -306,10 +306,8 @@ void add_for_retrans(struct tcpcb_s *cb, struct tcphdr_s *th, __u16 len,
 
 void tcp_reoutput(struct tcp_retrans_list_s *n)
 {
-    //unsigned int datalen = n->len - TCP_DATAOFF(&n->tcphdr[0]);
-
-    //if (datalen < n->cb->rcv_wnd)		/* don't record retry if not in recv window*/
     /*  I don't understand, a retrans is per definition within the rcv window */
+    //if (datalen < n->cb->rcv_wnd)		/* don't record retry if not in recv window*/
 	n->retrans_num++;
     n->rto <<= 2;				/* quadruple retrans timeout*/
     if (n->rto > TCP_RETRANS_MAXWAIT)		/* limit retransmit timeouts to 4 seconds*/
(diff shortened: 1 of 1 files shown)


Mellvik:
Thanks @ghaerr, great catch. 
I don't remember the reason for the change/swap - most likely an idea about performance, getting the packet to the NIC asap, then do the retrans 'admin' while the packet gets on the wire (or at least into the NIC queue if there is one). Like you point out, there are 'hidden' dependencies here that come back to bite unexpectedly. 

And yes, I do get the intermittent bad checksum messages when telnetting, and looking into my notes it's on the buglist. You just saved me the effort, thanks.

ghaerr:
> I don't remember the reason for the change/swap

Thanks for the comments. You might want to add `telnet localhost` to your regression testing. I use it a lot, since I'm not running on real hardware. Of course, QEMU works for NE2K; I wish it emulated other NICs.

>  I do get the intermittent bad checksum messages when telnetting

Interesting. On localhost, it starts immediately. I hadn't seen any errors talking to macOS from QEMU.

Mellvik:
Yes, I really should use localhost regularly for testing. It is obviously using different data paths useful for tracking down problems.

Telnet: sorry about that the confusion, I meant 'telnet localhost'. Like I said, I had seen the error before, and only in the localhost context. That's why I ended up forgetting about it.

Thanks.

Mellvik:
> I'm not running on real hardware. Of course, QEMU works for NE2K; I wish it emulated other NICs.

This reminds me - I really need to recheck VirtualBox with TLVC now that Lance is supported. That could be a useful new venue for testing.

I will post a discussion issue on TLVC if it turns into something.

PR #2469, opened 2025-11-17
[NEC V25] Add support for DS3231 I2C Realtime Clock

swausd:
DS3231 RTC modules with battery backup are used in the Arduino cosmos, are cheap and readily available. To use these modules, a serial I2C bus interface is needed. This PR implements a very simple bit-banged interface for use with the DS3231. It is specific to the NEC V25 ELKS port and is only useful to setup the software clock from the DS3231 at startup.

This is no ELKS device driver - just a simple interface for use with the clock app! May be the I2C interface can be used for other devices too, but I did not test this, nor do I see a use case for now. If someone wanta to try this, the code has to be changed into a device driver to serialize access!

Added the necessary parts to sys-utils/clock app, which already has many #ifdefs. The specific DS3231 and I2C code was separated to own source files sys-utils/ds3231.c/h and sys-utils/i2c-ll.S/h. The Makefile had to be modified too, so these NEC V25 specific files are only used for the V25 build.

While working on the I2C part, I noticed a major pitfall. The configuration and direction registers of the NEC V25 parallel ports are "write-only" and can only be used with byte-access. So I had to rework my existing SPI drivers and moved the initialization of the parallel ports P1/P2 to my startup code in the ROM-monitor, from where I start ELKS. The used initialization values are now documented in necv25.h and are used to rewrite the configuration of the I2C SDA pin in i2c-ll.S This is a rather tricky work around, but I think it is ok for now.

I tested all combinations of sw/hw SPI drivers with the I2C code and it worked. I also run elks/buildimages.sh and everything looked ok to me.

Oh ... and I struggled a little bit with git again. I reverted one of the commits to change the commit note because of a typing error. Not the most elegant way to do this....

### DIFF
--- a/elkscmd/sys_utils/ds3231.c
+++ b/elkscmd/sys_utils/ds3231.c
@@ -1,0 +1,68 @@
+// DS3231 RTC device driver
+//
+// 14. Nov. 2025 swausd
+
+#include "ds3231.h"
+#include "i2c-ll.h"
+
+
+// Initialize the I2C bus
+void ds3231_init(void)
+{
+   i2c_init();
+
+   write_rtc(REG_CNTRL, 0x00);
+   write_rtc(REG_STAT,  0x08);
+}
+
+
+// Read from a DS3231 register
+unsigned char ds3231_read(unsigned char clock_reg)
+{
+   unsigned char val = 0;
+
+   i2c_start();
+
+   i2c_write(DS3231_WRITE);
+
+   if (i2c_ack())
+   {
+      i2c_write(clock_reg);
+
+      if (i2c_ack())
+      {
+         i2c_start();
+         i2c_write(DS3231_READ);
+
+         if (i2c_ack())
+            val = i2c_read();
+      }
+   }
+
+   i2c_nak();
+   i2c_stop();
+
+   return val;
+}
+
+
+// Write to a DS3231 register
+void ds3231_write(unsigned char clock_reg, unsigned char val)
+{
+   i2c_start();
+
+   i2c_write(DS3231_WRITE);
+
+   if (i2c_ack())
+   {
+      i2c_write(clock_reg);
+
+      if (i2c_ack())
+      {
+         i2c_write(val);
+         i2c_ack();
+      }
+   }
+
+   i2c_stop();
+}
--- a/elkscmd/sys_utils/ds3231.c
+++ b/elkscmd/sys_utils/ds3231.c
@@ -11,8 +11,8 @@ void ds3231_init(void)
 {
    i2c_init();
 
-   write_rtc(REG_CNTRL, 0x00);
-   write_rtc(REG_STAT,  0x08);
+   ds3231_write(REG_CNTRL, 0x00);
+   ds3231_write(REG_STAT,  0x08);
 }
 
 
--- a/elks/arch/i86/drivers/block/spi-hw-necv25.S
+++ b/elks/arch/i86/drivers/block/spi-hw-necv25.S
@@ -17,6 +17,11 @@
  * serial has to be serial 0
  * /SCK0 connected with /CTS0, Tx clock (out) to Rx clock (in)
  * configure baud rate by setting BR_PRSCLR to desired value
+ *
+ **************************************************************************
+ * Attention: Direction and functions of port pins have to be set in a 
+ * central position (BIOS or startup code). See definitions in necv25.h
+ **************************************************************************
  */
 
 #include <arch/necv25.h>
@@ -62,11 +67,6 @@ spi_init_ll:
     movb    $0x47, (NEC_STIC0)               // STIC0: clear Tx ready flag
     movb    $0xc0, (NEC_SCM0)                // SCM0:  synchronous mode, enable Rx/Tx, external Rx clock
 
-    clr1    PIN_CS, NEC_PMC2                 // set to IO, not special function
-    clr1    PIN_CS, NEC_PM2                  // set to output
-    set1    PIN_CS, NEC_P2                   // de-select SPI
-    set1    PIN_SCK0, NEC_PMC1               // set to special function
-    clr1    PIN_SCK0, NEC_PM1                // set to output
     set1    PIN_CS, NEC_P2                   // de-select SPI
 
     pop     %ds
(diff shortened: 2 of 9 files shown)


ghaerr:
Hello @swausd,

This looks great, thanks! It's nice having an I2C implementation, even if it is specific to your V25 hardware. 

And yes, the clock's a mess, but this is the best way to handle it for now. It would be nice to get the I2C and DS3231 out and into some library code someday, along with the other specialized clock stuff, but that code will likely never be seen or used by other than clock anyways, it seems. By all rights, the kernel should be handling all the clock stuff anyways instead of applications talking directly to hardware, but it would be even worse having to include specific clock code in the kernel only to be used just once at boot. So all of this is fine.

I have a couple comments on ia16-elf-gcc's treatment of char/unsigned char parameters to functions, I'll leave them separately.

I struggle with git too occasionally, especially when going outside a normal workflow. The good news is that a merge commit doesn't affect any files whose changes were reverted, so don't worry about it. Sometimes it's far easier to let the merge commit sort it all out than to start over!

In the future in Makefiles, I would prefer the non-negative form `ifeq ($(CONFIG_ARCH_NECV25), y)` rather than ifneq, since the lines immediately below are more closely associated with the if condition. If you like, you can change that in another commit, as I'm going to make it easy and just commit this now!

Thank you!

swausd:
Hi @ghaerr,
thank you for the commit and the feedback. I used the negative form in the Makefile to leave the most needed variant up front. The V25 is a very special case, after all.
I will change this with a commit tomorrow.

elkscmd/sys_utils/ds3231.c
```diff
@@ -0,0 +1,68 @@
...
+{
+   i2c_init();
+
+   ds3231_write(REG_CNTRL, 0x00);
+   ds3231_write(REG_STAT,  0x08);
+}
+
+
+// Read from a DS3231 register
+unsigned char ds3231_read(unsigned char clock_reg)
```
ghaerr, line 20: Just a warning on `ia6-elf-gcc` and char or unsigned char parameters, the compiler is known to emit buggy code generating code to pass the parameter. See the 128 of drivers/block/ata.c for more details, and you can use `ia16-elf-objdump -D -r -Mi8086` to inspect the code generated. The bad code is in the calling routine, not the function using the char parameters, so it's hard to find unless looking at all calling cases.

Same issue for line 50 below.


To work around this, I highly recommend just changing the declaration(s) to int or unsigned int. There is never any additional code generated, as a MOVW will be used, and in some cases less code, since sometimes a MOVB AL; XOR AH,AH is generated before a PUSHW AX. In other cases, completely incorrect code is generated and the program doesn't work.

I also don't recommend using char/unsigned char return values. Although I haven't yet found an actual compiler bug, I get the feeling that the 8086 version of GCC just isn't fully tested with 8-bit access.



If interested, I can use the ata.c code to (again) inspect exactly what the compiler bug is, for our education.

swausd: Thanks for the hint! I will look into this tomorrow and change from char to int. No problem.

swausd: Hi @ghaerr 
I changed  (and tested) my code to use unsigned int and will commit the changes  it in a few minutes. Tried to see the compiler problem you described in ata.c by changing the definition of the OUTB in line 128 and comparing the resulting calls of OUTB in the assembler listing. But frankly, I could not see what could be wrong.

> If interested, I can use the ata.c code to (again) inspect exactly what the compiler bug is, for our education.

So if it is not too much trouble for you, I would really like to see exactly what you have found.

Thank you!

ghaerr: > But frankly, I could not see what could be wrong.

I looked more into the undocumented compiler bug last night, and it appears that the required trigger requires a more complex expression, one that apparently isn't seen when just passing byte parameters in the simple case. 

What I did find was that the compiler generated a lot code than what one would not expect when byte parameters are used. All byte parameters are still pushed using PUSHW, so instructions are generated to clear the high-order byte. This sometimes gets more complicated when the parameter is in a register that doesn't allow clearing the high half (e.g. SI or SI).  It appears the compiler doesn't like leaving a "junk" byte on the stack (for a parameter at least) so it feels the need to clear it. Internally, that seems to trigger entirely different compiler code paths which end up generating vastly different code and register usage.

After looking through this, I firmly believe in more than before that there is no benefit in passing byte parameters, for most code, with ia16-elf-gcc. It's not like the old days of C compilers on the 8080 where this was a big deal!!

> if it is not too much trouble for you, I would really like to see exactly what you have found.

I spent about a half hour looking at it, but didn't have the time to understand exactly when the problem is occurring, due to the nature of the compiler generating vastly different code for ata.c from a single change of unsigned int to unsigned char. I would like to trace this down though, the problem occurs in ata.c only when the compiler happens to see an "unsigned long >> 16" which I'm guessing changes the register allocations enough to trigger the bug.

The bug is documented in ata.c with the first two FIXME comments in the file.

If you're still interested in getting more of a taste for this issue, try doing the following: save the original ata.o, then create a copy of ata.c with a one line change in OUTB taking an unsigned char parameter on line 129, and make a new ata.o, rename it as ata-char.o. Then run:
```
$ ia16-elf-objdump -D -r -Mi8086 ata.o.> file1
$ ia16-elf-objdump -D -r -Mi8086 ata-char.o > file2
$ diff file1 file2
```
You'll see the compiler generates different code for functions not even related to OUTB or ata_cmd!! So it's messy. I found the original bug because the ATA driver was failing, and I put a printk in OUTB and found it, after lots of debugging. The next step would be isolating ata_cmd and OUTB into a separate file, but I tried that last night and it got to be lots of work. So one will need to gradually reduce ata.c while still keeping an eye on the obdump output.

swausd: Hi @ghaerr,
Thank you for this long explanation and your invested time! I think I have to read this over again some more times, to understand ;-) But you are right, there is no need to save space on the stack as the 8088/8086 are 16 bit CPUs after all.

So I will remember to stay at 16 bit.

ghaerr: Agreed. I've found that to really know what's going on with an extreme optimizing compiler like ia16-elf-gcc, one has to look at the code generated all the time. For this, I've got an alias which I use constantly:
```
alias odd='ia16-elf-objdump -D -r -Mi8086'
```
This helps particularly to ensure that embedded ASM statements behave as intended.

swausd: I already tried to find out how to get an assembly listing from the compiler and had no luck. Now I know how - thanks to you!

PR #2470, opened 2025-11-18
Clean up

swausd:
Changed the conditional building of the clock app to "positive" decision in the Makefile. Changed "unsigned char" to "unsigned int" for parameters and return values of the DS3231 and I2C interface to work araound compiler problems.


ghaerr:
Discussed more in https://github.com/ghaerr/elks/pull/2469#discussion_r2534925479. Thank you!

PR #2472, opened 2025-11-19
Fix build and gitignore

hexadec1mal:


elks/tools/elkscmds/Makefile
```diff
@@ -53,7 +53,7 @@ all: $(CLEANME)
 	$(CC) -o $@ $(CFLAGS) -D__far= $^
 
 ../bin/disasm: $(DISASM)
-	$(CC) -o $@ $(CFLAGS) -I. -D__far= -Wno-int-to-void-pointer-cast -Wno-format $(DISASM)
+	$(CC) -o $@ $(CFLAGS) -I. -D__far= -std=gnu99 $(DISASM)
```
ghaerr: Can you leave the previous -W warnings in place and also add -std=gnu99? Unfortunately on macOS (where there is no GCC, only CLANG) removing the -W warnings causes warnings to be displayed, which -std=gnu99 doesn't effect.

I don't have Linux: is -std=gnu99 required to remove warnings? I would think that the -W warnings would also work on Linux, but I might be wrong. 

Thank you!

hexadec1mal: GCC 15 updated default standard to c23, which enabled boolean types by default:
```
error: ‘bool’ cannot be defined via ‘typedef’
note: ‘bool’ is a keyword with ‘-std=c23’ onwards

error: cannot use keyword ‘false’ as enumeration constant
note: ‘false’ is a keyword with ‘-std=c23’ onwards
```
-std=gnu99 is to fix that

-Wno-format don't remove any warnings in these files, so I removed it
-Wno-int-to-void-pointer-cast is a clang only warning, which is an alias to -Wno-int-to-pointer-cast, so I replaced it with that
-Wno-pointer-to-int-cast is to remove another warning

ghaerr: > -Wno-format don't remove any warnings in these files, so I removed it

That's incorrect - did you test this on macOS? 

Please add back in -Wno-format, as CLANG still gives warnings:
```
gcc -o ../bin/disasm -Wall -I/Users/greg/net/elks-gh/include -I/Users/greg/net/elks-gh/elks/include -D__KERNEL__ -I. -D__far= -Wno-int-to-pointer-cast -Wno-pointer-to-int-cast -std=gnu99 /Users/greg/net/elks-gh/elkscmd/debug/disasm.c /Users/greg/net/elks-gh/elkscmd/debug/dis.c /Users/greg/net/elks-gh/libc/debug/syms.c
/Users/greg/net/elks-gh/elkscmd/debug/dis.c:91:48: warning: format specifies type
      'unsigned short' but the argument has type 'int' [-Wformat]
        if (!f_asmout) printf("%04hx:%04hx  ", cs, ip);
                               ~~~~~           ^~
                               %04x
/Users/greg/net/elks-gh/elkscmd/debug/dis.c:91:52: warning: format specifies type
      'unsigned short' but the argument has type 'int' [-Wformat]
        if (!f_asmout) printf("%04hx:%04hx  ", cs, ip);
                                     ~~~~~         ^~
                                     %04x
/Users/greg/net/elks-gh/elkscmd/debug/dis.c:130:42: warning: format specifies type
      'unsigned short' but the argument has type 'int' [-Wformat]
        if (!f_asmout) printf("%04hx  ", ip);
                               ~~~~~     ^~
                               %04x
/Users/greg/net/elks-gh/elkscmd/debug/dis.c:148:42: warning: format specifies type
      'unsigned short' but the argument has type 'int' [-Wformat]
        if (!f_asmout) printf("%04hx  ", ip);
                               ~~~~~     ^~
                               %04x
4 warnings generated.
```
It is better that you just add -std=gnu99 if you can't test on macOS. I still don't understand why in fact that is being added. What is it doing? What warnings are you seeing? -std=gnu99 is not portable.

hexadec1mal: -std=gnu99 is to fix build with gcc 15

ghaerr: Ok, I see that now. Thank you!

ghaerr: It looks like disasm.c is the ultimate culprit here with its use of C23 reserved words. I'll fix that, thank you for pointing this out!

PR #2474, opened 2025-11-22
[bios] Add bios_check_sector98() and removed bios call to check sector size from bios_disk_rw for PC-98 hardware.

tyama501:
Hello @ghaerr 

This is a PR for https://github.com/ghaerr/elks/issues/2398#issuecomment-3538918975

This worked well on my hardware,
expect detecting ELKS boot parameter when boot FD1200 on the 3rd FD drive.
It goes to probe.
I may change something if I found the reason,
but I will leave it for now and go back to the original issue.

Thank you!


ghaerr:
Thanks @tyama501 I'm glad you got this working and can still use the probe routine!

Issue #2457, opened 2025-11-15
Arrow keys don't work on ash/vi on slow systems from serial terminal


ghaerr:
Described in https://github.com/ghaerr/elks/issues/2456#issuecomment-3536442776 as VTIME doesn't work on slow systems.

Looking at the source to `ash`, VTIME is not used for function/arrow key parsing. Instead, single character I/O is setup using VMIN=1, VTIME=0. This should mean that the kernel waits indefinitely for a single character, from linenoise code.

In ash, when ESC is seen, two separate `read` calls are made, meaning ash won't timeout on the ESC, but waits for two more characters. If these characters are [ followed by D, for instance, the left arrow is interpreted.

@Mellvik, can you test /bin/sh on your slow system? At the ash prompt, type sdf, then press ESC, then [, then D. Does the cursor move over to the left? If so, type x, the result should be sdxf. Does this occur? If so, then the problem is in the scancode keyboard driver, where the BIOS scancode for arrow key is translated to an ANSI sequence. If this method of emulating an arrow key doesn't work, then there is possibly something wrong with the VMIN tty code (not VTIME).

I will investigate the method `vi` uses for arrow key parsing shortly.

ghaerr:
@Mellvik,

I have tested ELKS using the method described above on the (supposed to be) cycle-accurate MartyPC emulator. I can't get the arrow keys to fail with ash. Can you test this on a machine running ELKS vs TLVC? I want to make sure we're not chasing down a problem in TLVC that has been fixed in ELKS.

Also, I checked how vi works, and it uses a more complicated method that ends up using `alarm` in order to return from a `read` hanging (otherwise waiting for the remaining characters following an ESC in an ANSI escape sequence). So it could be that the issue(s) in vi are a separate problem. We will see after getting ash working.

Thank you!

ghaerr:
I just noticed that the kernel may operate differently than described when reading from a serial driver rather than the console.

@Mellvik can you verify whether arrow keys work on console, but not on serial?

Mellvik:
Will do, @ghaerr. 
I believe this is really hard to replicate on any emulator. The baud rate has to be 'real' as in physical 9600 or lower (19200 will cause lost characters on these systems).

As to what actually gets on the screen when using the arrow buttons on `ash`, it varies - typically the ESC is gone/ignored and the CAP letter indicating the actual ANSI command is shown. In `vi` it's more chaotic since any letter and control char is (usually) a command.

[...]

PR #2475, opened 2025-11-23
[kernel] Enable divide by zero handler on swan target

asiekierka:


ghaerr:
Hello @asiekierka, thanks for this update.

BTW, not sure with how much detail you've been following the conversation in #2457, but it now seems by my analysis that it should/might be possible to generate an interrupt stack overflow by receiving steady streaming serial input at a high baud rate on the Swan device, due to the PIC EOI being sent prior to the end of the serial input handler. This would be due to stacked register saves occurring on the interrupt stack while still not finished with the previous interrupt handlers execution. Can you by chance test a very high baud rate to see if this happens? I would be interested to know. We also have a way to measure the interrupt stack usage (although that takes yet more time to execute as well).

Thank you!

asiekierka:
> but it now seems by my analysis that it should/might be possible to generate an interrupt stack overflow by receiving steady streaming serial input at a high baud rate on the Swan device, due to the PIC EOI being sent prior to the end of the serial input handler.

- The driver can't go any faster than 38400 baud, and given the V30MZ's instruction speeds place it in the same league as a ~6-8 MHz 286, I'm not *too* worried about it. Since I developed it, I discovered a way to enable a 192000 (yes, 192 thousand) baud rate on the hardware, but it's not yet exposed in the ELKS driver.
- However, you are correct; the interrupt stack can, in theory, overflow with the way it currently operates. 
- Either way, the port is still in a state where it can barely run `sash`, let alone programs. I've been tinkering with adding support for bank switching (that is, giving each process its own 64 KiB area of memory in `0x10000 - 0x1FFFF`, and switching between them as part of the task switcher), but it's a little tricky, not to mention requiring a few fairly invasive patches depending on which syscalls are adjusted to split banks. I can go into more detail, if you're curious.

ghaerr:
> the interrupt stack can, in theory, overflow with the way it currently operates.

Sounds like we're on the same page, thanks for your response on that.

> The port is still in a state where it can barely run sash, let alone programs.

Too little free RAM?

> but it's a little tricky, not to mention requiring a few fairly invasive patches depending on which syscalls are adjusted to split banks. 

I would bet so. I would recommend starting on your own fork for invasive patches, but am happy to help discuss any architectural or other issues in any case.

> I can go into more detail, if you're curious.

Yes, I'd like to hear about your design. Go ahead and open a separate issue for that - I'm curious how you're thinking about it.

PR #2479, opened 2025-11-24
[mm, exec] Allow text segments pointing directly to ROMFS

asiekierka:
The changes proved surprisingly small and independent, so here goes:

- `seg_alloc_fixed()` has been added, complete with a notion of "fixed segments" (that is, memory segments _not_ managed by the allocator). These can be distinguished by their linked list pointers being NULL. (Maybe it should be renamed to `seg_create_fixed()`?)
- `seg_free()` has been modified to support deallocating segments created by `seg_alloc_fixed()`.
- `exec()` has been modified to use the text segment directly when possible. It tries to be conservative about determining when that is the case:
    - The file system on which the inode is located has to be of the ROMFS type.
    - The text segment has to be stored at a multiple of 16 in the file. This is, thankfully, the case with the standard MINIX header.
    - The header has to be the standard MINIX header. This conveniently excludes cases of relocatable and compressed executables - none of which can be supported by this method.

I've tested this with `emu86.sh`, and it appears to work fine, with `ps -l` correctly showing the `CSEG` as pointing to ROM for processes.

The benefits are twofold:

- Any program loaded from ROM uses less memory - it's reduced by the size of its text segment. This can even be useful for setups where user programs are stored on a hard drive, if, say, the shell is kept in ROM.
- Avoiding a RAM allocation and copy means programs technically load slightly faster, though I expect the difference to be imperceptible.

### DIFF
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -112,6 +112,25 @@ static void seg_merge (segment_s * s1, segment_s * s2)
     heap_free (s2);
 }
 
+#ifdef CONFIG_ROMFS_FS
+
+// Allocate fixed segment for romfs (outside of memory allocator)
+
+segment_s * seg_alloc_romfs (seg_t base, segext_t size, word_t type)
+{
+    segment_s * seg = heap_alloc(sizeof(segment_s),
+        HEAP_TAG_SEG | HEAP_TAG_CLEAR);
+    if (!seg)
+        return seg;
+
+    seg->base = base;
+    seg->size = size;
+    seg->flags = SEG_FLAG_USED | type;
+    seg->ref_count = 1;
+    return seg;
+}
+
+#endif
 
 // Allocate segment
 
@@ -129,6 +148,15 @@ segment_s * seg_alloc (segext_t size, word_t type)
 
 void seg_free (segment_s * seg)
 {
+#ifdef CONFIG_ROMFS_FS
+    // Handle segments created outside of the memory allocator
+    // (via seg_alloc_romfs).
+    if (seg->all.prev == NULL && seg->all.next == NULL) {
+        heap_free(seg);
+        return;
+    }
+#endif
+
     // Free segment will be inserted to free list:
     //   - tail if merged to previous or next free segment
     //   - head if still alone to increase 'exact hit'
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -352,6 +352,26 @@ static int FARPROC execve_aout(struct inode *inode, struct file *filp,
     /*
      *      Looks good. Get the memory we need
      */
+
+#ifdef CONFIG_ROMFS_FS
+    if (filp->f_inode->i_sb->s_type->type == FST_ROMFS
+        && mh.hlen == EXEC_MINIX_HDR_SIZE
+    ) {
+        /* Point the code segment directly to in-memory ROMFS. This runs text
+         * segments directly from ROM, as opposed to making copies of them in
+         * RAM. Not supported for compressed or relocatable binaries. */
+
+        seg_code = seg_alloc_romfs(
+            filp->f_inode->u.romfs.seg + (filp->f_pos >> 4),
+            bytes_to_paras(mh.tseg), SEG_FLAG_CSEG);
+        if (!seg_code) goto error_exec3;
+
+        filp->f_pos += (size_t)mh.tseg;
+#ifdef CONFIG_EXEC_MMODEL
+        need_reloc_code = 0;
+#endif
+    } else
+#endif
     if (!seg_code) {
         bytes = (size_t)mh.tseg;
         paras = bytes_to_paras(bytes);
--- a/elks/include/linuxmt/mm.h
+++ b/elks/include/linuxmt/mm.h
@@ -55,6 +55,7 @@ int fs_memcmp(const void *,const void *,size_t);
 
 /* Memory allocation */
 
+segment_s * seg_alloc_romfs (seg_t, segext_t, word_t);
 segment_s * seg_alloc (segext_t, word_t);
 void seg_free (segment_s *);
 segment_s * seg_get (segment_s *);
--- a/elks/fs/config.in
+++ b/elks/fs/config.in
@@ -11,7 +11,7 @@ mainmenu_option next_comment
 
 	if [ "$CONFIG_ROMFS_FS" = "y" ]; then
 		hex 'Base in ROM (paragraphs)' CONFIG_ROMFS_BASE 0x8000
-		fi
+	fi
 
 	bool 'FAT filesystem' CONFIG_FS_FAT 'n'
 


ghaerr:
Hello @asiekierka,

Nice work. Pretty tight code and you've chosen a nice design with a lot of benefits for our ROMFS users.

I've been traveling for the past 5 days and now am out of the country for the next 8, so bear with me, it seems when it rains, it pours, and I may not have much time! We've got active development going on with the new V25 port, but I think we can squeeze this one in now anyways since there's really so few changes for such a nice enhancement. Nonetheless, I'd still like a few changes; I'll leave those separately by the code affected.

Thank you!

config/Configure.help
```diff
@@ -557,6 +557,10 @@ CONFIG_ROMFS_FS
 CONFIG_ROMFS_BASE
   Base of the file system in memory (in paragraphs).
   Typically 0x8000 for a 512K ROM in a 1 MB address space.
+CONFIG_ROMFS_FS_DIRECT_CSEG
+  Saying y to this option runs text segments directly from ROM, as
+  opposed to making copies of them in RAM. Not supported for
+  compressed or relocatable binaries.
```
ghaerr: I'm of the belief that having everything "just work" with fewer options is better. Thus, I'd like to use CONFIG_ROMFS_FS as the controlling define for this enhancement. You might want to move this nice descriptive comment over to above the new code in exec.c.

asiekierka: That is not a problem to change, but I'm concerned about a situation in which someone, for whichever reason, explicitly relies on the functionality of the text segment technically being writable on a standard ELKS setup, given the lack of memory protection.

elks/arch/i86/mm/malloc.c
```diff
@@ -112,6 +112,20 @@ static void seg_merge (segment_s * s1, segment_s * s2)
...
+    segment_s * seg = heap_alloc(sizeof(segment_s), HEAP_TAG_SEG | HEAP_TAG_CLEAR);
+    if (!seg)
+        return seg;
+
+    seg->base = base;
+    seg->size = size;
+    seg->flags = SEG_FLAG_USED | type;
+    seg->ref_count = 1;
+    return seg;
+}
```
ghaerr, line 131: Please wrap this new function with ifdef CONFIG_ROMFS_FS.

elks/arch/i86/mm/malloc.c
```diff
@@ -129,6 +143,13 @@ segment_s * seg_alloc (segext_t size, word_t type)
 
 void seg_free (segment_s * seg)
 {
+    // Handle segments created outside of the memory allocator
+    // (via seg_alloc_fixed).
+    if (seg->all.prev == NULL && seg->all.next == NULL) {
+        heap_free(seg);
+        return;
+    }
+
```
ghaerr, line 159: Lets ifdef this CONFIG_ROMFS_FS also.

elks/fs/exec.c
```diff
@@ -352,6 +352,27 @@ static int FARPROC execve_aout(struct inode *inode, struct file *filp,
     /*
      *      Looks good. Get the memory we need
      */
+
+#if defined(CONFIG_ROMFS_FS) && defined(CONFIG_ROMFS_FS_DIRECT_CSEG)
+    if (filp->f_inode && filp->f_inode->i_sb
+        && filp->f_inode->i_sb->s_type->type == FST_ROMFS
+        && !(filp->f_pos & 15)
+        && mh.hlen == EXEC_MINIX_HDR_SIZE
+    ) {
```
ghaerr, line 359: There'll never be a case where filp->f_inode is NULL since the header's already been read, and likewise for i_sb, so the two tests in line 357 can be deleted. Also, if I'm not mistaken filp->f_pos will always be a multiple of 16 on ROMFS, won't it? Can you please check that in romfs.c or with an operating kernel?

asiekierka: I'm worried about removing the `!(filp->f_pos & 15)` check as I'm worried someone might be surprised that changing `EXEC_MINIX_HDR_SIZE` for whichever reason could lead to a code crash in this configuration; however, I can propose instead replacing it by a `!(EXEC_MINIX_HDR_SIZE & 15)` check, which would be optimized out by the compiler and cover the same scenario.

elks/fs/exec.c
```diff
@@ -352,6 +352,27 @@ static int FARPROC execve_aout(struct inode *inode, struct file *filp,
...
+        // Point the code segment directly to in-memory ROMFS.
+        seg_code = seg_alloc_fixed(
+            filp->f_inode->u.romfs.seg + (filp->f_pos >> 4),
+            bytes_to_paras(mh.tseg),
+            SEG_FLAG_CSEG
+        );
```
ghaerr: Please reformat as many arguments as will fit on an 80-90 char line, as this is the way the rest of the kernel is formatted (or at least is supposed to be).

elks/include/linuxmt/mm.h
```diff
@@ -55,6 +55,7 @@ int fs_memcmp(const void *,const void *,size_t);
 
 /* Memory allocation */
 
+segment_s * seg_alloc_fixed (seg_t, segext_t, word_t);
```
ghaerr: How about the idea of `seg_alloc_romfs` since this function only works with ROMFS code segments?

asiekierka: It can, in theory, be used for any memory location placed outside of the memory manager's jurisdiction - while it's true that upstream probably won't see uses beyond ROMFS code segments, the function itself can be used in a more general fashion. I'm not sure I want to introduce file system nomenclature into memory manager code - it feels like going against separating concerns.

config/Configure.help
ghaerr: > explicitly relies on the functionality of the text segment technically being writable on a standard ELKS setup,

The code segment is already shared when possible across processes, so code relying on writing the code segment breaks that assumption and would break when run simultaneously. 

My direction with ELKS is simplicity - we have too few users knowledgeable enough to set all the proper options as it is, and I'm trying to get to a point where there's such a thing as a standard distribution which works very well out of the box for a majority of users. The less options the better - I've spent the last five years removing them and consider that progress.

asiekierka: Okay then.

(My personal position is that such issues should be resolved by better configuration wizards, not by removing options, but it has to also be said that fewer options means fewer debugging headaches and reduced code complexity, and it's not my labor I'd be arguing about, so...)

elks/fs/exec.c
ghaerr: > changing EXEC_MINIX_HDR_SIZE for whichever reason

I really do appreciate your concerns for every case - but EXEC_MINIX_HDR_SIZE will never change without breaking all sorts of other code, there are already three added options on top of it. The a.out format in general requires paragraph-aligned sections for text, fartext and data.

We may have a slight difference in philosophy - with no hardware checking things will go to hell quickly when major assumptions like this are broken, so I'm not worried about it. Instead, readability and understandability of kernel code is more important - so that other developers can understand ELKS without having to read tons of modern-day-style checks that seem to bloat so much code these days. I want developers to be able to read less code and understand more...

elks/include/linuxmt/mm.h
ghaerr: I'm running out of time to comment - I had originally thought we were keeping these mods in your fork for the time being. To me , seg_alloc_fixed didn't mean anything, whereas seg_alloc_romfs immediately tells me what's going on. All of this may want to be changed or upgraded after you've had more change to implement all your planned Swan enhancements, I'm just trying to get this done ASAP while I still have some time while traveling.

Ultimately, our naming issues come down to whose fork this stuff is in and how maintains it - I have to maintain the ELKS main fork and I'm unable to even test the Swan code, unfortunately. But this code can be tested, and I use the same method emu86.sh.

> I'm not sure I want to introduce file system nomenclature into memory manager code - it feels like going against separating concerns.

But you're adding a function that won't work without romfs compiled in!!

asiekierka: I mean, it would work without romfs compiled in, it just wouldn't be used for anything without romfs compiled in - effectively becoming unused or dead code. But sure, it's really a nitpick.

For testing, I use an emulator myself... I could probably contribute some emulation scripts for the platform at a later date.

asiekierka:
That should be everything. It's 11:30 PM where I live, so I'll be going to bed now - thank you for your patience. If there's anything else, I'll respond to it tomorrow.

ghaerr:
Thank you @asiekierka for this super enhancement and some great code. Thanks for your patience with me also, I appreciate it!

ghaerr:
@asiekierka 

> It can, in theory, be used for any memory location placed outside of the memory manager's jurisdiction
> the function itself can be used in a more general fashion. 

Now that I have had a bit of time to better review this function - you're right. I got caught up looking at the use of the 
`filp->f_inode->u.romfs.seg + (filp->f_pos >> 4)` code and was thinking romfs only. So I would have to agree with your original statement and design. If you'd like to change this function name back to seg_alloc_fixed, that's probably a better name. Sorry! :)

ghaerr:
@asiekierka 
Have you tried testing this PR with exec's current handling of shared text segments? This is performed with the following code, very early in exec():
```
    /* Look for the binary in memory */
    seg_code = 0;
    currentp = &task[0];
    do { 
        if ((currentp->state <= TASK_STOPPED) && (currentp->t_inode == inode)) {
            debug("EXEC found copy\n");
            seg_code = currentp->mm[SEG_CODE];
            break;
        }   
    } while (++currentp < &task[max_tasks]);
```
Then, later, *after* your check for ROMFS, the else case of `if (!seg_code)` increments the text segment use count using seg_get() (which will never get executed with your new code):
```
    } else {
        seg_get (seg_code);
#ifdef CONFIG_EXEC_MMODEL
        filp->f_pos += esuph.esh_compr_tseg? esuph.esh_compr_tseg: (size_t)mh.tseg;
        filp->f_pos += esuph.esh_compr_ftseg? esuph.esh_compr_ftseg: (size_t)esuph.esh_ftseg;
        need_reloc_code = 0;
#else
        filp->f_pos += (size_t)mh.tseg;
#endif
    }
```
It would seem to me that it would be better to keep the original method of shared text consistent, rather than allocating another segment entry using seg_alloc_romfs/seg_alloc_fixed the way you're doing it now. I think this would be easily fixed though by moving your new code's if statement `if (filp->f_inode->i_sb->s_type->type == FST_ROMFS ...` down to below the `if (!seg_code) {` statement on line 375. 

In this way, if seg_code == 0 in the original code, *then* you check for FST_ROMFS, otherwise, the else case of the original if statement will handle the seg_get and segment use count  increment properly. If you have to use a goto to skip some code after the seg_alloc_romfs/fixed case, that's ok since much less code is duplicated.

Also, this method corrects the not-so-good "dangling" else just inside the \#endif on line 373 which is no longer needed.

Does this all make sense? At the end of the day, we want a single code segment descriptor per shared-text processes, not two like we now have; the code is simpler, with less duplication.

asiekierka:
> Does this all make sense?

Yes! The main reason I didn't spot this is because my original implementation of this did not have `seg_alloc_fixed()`, and instead had a somewhat more separated code path for handling executables from ROMFS; then, when I refactored it to use that, I failed to spot the opportunity. Oops!

I'll get around to it.

swausd:
Hi @asiekierka ,

this change has two side effects! 

1. In system programming of the ROM with elkscmd/romprg can't be used any more, because reprogramming not only changes the ROM containing the code segment of romprg but also blocks the ROM for the time of erasing/writing. 

2. ROM usually is noticeable slower than RAM so it may be wanted to get the code segments copied from ROM to RAM for faster execution.

So it would be nice to have a config option to turn this new feature on/off.

ghaerr:
Hello @swausd,

> So it would be nice to have a config option to turn this new feature on/off.

@asiekierka's original PR included a config option to enable/disable this feature, which I shot down trying to keep options to a minimum. I'm wondering whether it would make sense to flag executables with a mode bit (e.g. romprg or others), or just a global option?

ghaerr:
>  but also blocks the ROM for the time of erasing/writing.

How is that happening, does `romprg` open the ROM device for writing, which is then failing? (I haven't looked much at the manner in which romprg works).

Let's discuss a few ways both options above might be solved: should this be a kernel compile-time option, or perhaps a dynamic option using `sysctl`, or perhaps individual application mode bits? What are you thinking @swausd, that the option isn't useful for V25, but only for those systems with very little RAM?

asiekierka:
> How is that happening, does romprg open the ROM device for writing, which is then failing? (I haven't looked much at the manner in which romprg works).

Well, if romprg erases the entire ROM, it would also erase its own text segment, no?

I don't like the idea of using an application mode bit here, unless more users declare interest in such mixed operation. A configuration option would be easy to restore, and there's `#ifdef`s around most of the code anyway.

swausd:
Hi @ghaerr ,

in system programming of the ROM needs full access to the ROM chip (not an ELKS device driver) and blocks the whole ROM for a "very long time". During this time no code can be executed nor data read from the whole ROM. That's why @cocus had to block the whole OS for using romprg with a final reboot of the system.

I think a compile time option is OK.

swausd:
Code execution from ROM is an interesting option for **all** systems with ROMFS. But during development, in system programming is very helpful, especially for those tinkerers with no external programming device.

asiekierka:
I think any kind of global setting would work fine here - be it a sysctl, a boot option, or a compile-time flag. However, a compile-time flag is worse: if you have a kernel built with the option enabled, there is no way to `romprg` it into one with the option disabled!

swausd:
> if you have a kernel built with the option enabled, there is no way to romprg it into one with the option disabled!

Right!

swausd:
> a boot option

Boot options usually don't work with ROMFS.

ghaerr:
> Boot options usually don't work with ROMFS.

That's right. And `sysctl` might be tricky, since sysctl can only run after boot, and thus many programs may already be running with their code segments mapped to ROM. 

Glad we're discussing this, I'm sure a good solution will be found.

ghaerr:
@swausd,

Given @asiekierka's comment about a compile-time option resulting in not being able to romprg a new kernel with the option disabled, and using sysctl being very hard to implement, I'm back to considering using a file mode bit to resolve this issue.

If a mode bit (or combination) was used to disable running that executable from ROM, then setting that bit on `romprg` would allow it to run from RAM and thus reprogram ROM while executing. There might still be some extra code required in romprg to potentially disable other applications from running, unless all other applications are sleeping for input.

What do you think of this idea - it would take only one extra line of code in exec.c, and would also allow specifiable applications to be disabled from running from ROM, if desired in a given distribution, while conserving RAM for all other applications otherwise.

swausd:
Hi @ghaerr ,

> What do you think of this idea 

I don't know if I have an overview of the overall impact of this change, but if it's just a line of code, we should give it a try. Here's what romprg needs just for completeness:

1. Exclusive access to the whole ROM is needed, even if only parts of it are programmed.
2. Full control of ROM access timing is required, so there should be no rescheduling or interrupts. This is prevented by disabling all interrupts.
3. After programming the ROM, ELKS has to reboot, as no assumption on ROM contents can be made by the system or other programs.

Just give me a notice when you are ready and I will test romprg.

ghaerr:
> Just give me a notice when you are ready and I will test romprg.

I'm working on it - turns out ROMFS doesn't actually store Linux-compatible file mode bits in ROM, even though it uses a full word to store the inode type. I'm looking into rewriting that.

asiekierka:
Given it's only needed by one specific executable, and it needs it in every situation, wouldn't it be easier to add a way to add a mark in the MINIX header (or use a supplementary header, which already effectively does that) that says "this program cannot run from ROM"?

ghaerr:
>  add a mark in the MINIX header (or use a supplementary header, which already effectively does that)

Possibly easier, yes. But modifying executable formats brings in a host of other potential issues with outside emulators, etc. The idea wasn't just kluging romprog to work, which could be done in a variety of ways, but rather allowing user-specifiable programs to run in RAM for speed, requested by @swausd. The easiest approach would be (close to your original idea) just adding yet another configuration option in config.h to enable/disable running from ROM entirely.

Finding out that ROMFS doesn't actually provide real UNIX mode bits should probably be fixed separately regardless. It appears the original reason for that might have been for ease in writing tools/mkromfs. Since a full word is allocated in ROM anyways, that's not a good enough reason anymore IMO.

swausd:
Hi @ghaerr and @asiekierka ,

on systems, which have an additional filesystem, like minix on a SD card on my system, romprg could be started from this filesystem as workaround. Gives time for a more generic solution later on, if desired.

ghaerr:
@swausd
> Gives time for a more generic solution later on, if desired.

I've implemented a one-line solution  in #2494 for those that want to run ELKS in minimal RAM environments, but still allow `romprg` to have its code segment copied to RAM while everything else runs direct from ROM using the 01000 sticky bit. 

This was very straightforward with minimal code once the ROMFS mode bit handling was rewritten. I realize though that for systems with lots of RAM like yours, not having a compile time option means slower execution for everything. Thus, I'm thinking that adding an option such as CONFIG_ROMFS_EXEC=y to enable ROM code segment execution, originally contributed by @asiekierka but initially rejected by myself, would be a good idea after all. Let me know what you think.

swausd:
@ghaerr ,

> I realize though that for systems with lots of RAM like yours, not having a compile time option means slower execution for everything. 

That is true for fast CPUs and slow ROM/FLASH with the need for wait states. On my system, the CPU is rather slow, so no wait states for FLASH necessary. As it is now, fine tuning using sticky bit gives full control over which apps/CSEGs to move to RAM and which to leave in ROM. I see no need for a global CONFIG_ROMFS_EXEC compile option.

Thanks for asking for my opinion!

PR #2482, opened 2025-11-25
[cmds] Fix more not handling wrapping lines correctly on terminals sh…

asiekierka:
…orter than 80 columns

The swan has 28 or 56 columns, depending on configuration.


ghaerr:
Thank you!

PR #2483, opened 2025-11-26
[kernel] moved not working debug()

swausd:
Moved in sys2.c 
`debug("(%P)sys_alarm` %d\n", secs);
from setalarm() to sys_alarm()


ghaerr:
Thanks for the fix!

PR #2488, opened 2025-11-27
[kernel] More accurate swan model reporting

asiekierka:
Also used the opportunity to move a bit of arch-specific code out of `init/main.c` and into `arch/`.


ghaerr:
Thank you @asiekierka!

PR #2489, opened 2025-11-27
[kernel] Fix handling of shared text segments when pointing to ROM

asiekierka:
As discussed, this PR:

- renames `seg_alloc_romfs` back to `seg_alloc_fixed`,
- fixes handling of shared text segments when a program from ROMFS is loaded.

It does not tackle the configuration issue yet, however.


ghaerr:
Thank you!

PR #2494, opened 2025-12-02
[romfs] Implement all inode mode bits in ROMFS

ghaerr:
Discussed in more detail in https://github.com/ghaerr/elks/pull/2479#issuecomment-3594079102.

Rewrites tools/mkromfs to transfer source file's mode bits for most file types, allowing finer-grained Read/eXecute modes on files or directories in a ROMFS filesystem. Any Any Write mode bit is masked off on file, directory, or symlinks and OR-ed in to character and block device special files.

Adding in full inode mode-bit capability then allowed the use of the "sticky" (i.e. S_ISVTX 01000) bit to be specified in any ROMFS file. 

For the time being, image/Make.images sets the sticky bit on /bin/romprg (if present), which should also solve the problem of requiring `romprg` to always be allocated its code segment in RAM, so that the ROM can be reprogrammed at runtime.

@swausd, let me know how this works for you. Should you need modifications to `romprg`, go ahead and submit them in a separate PR.

### DIFF
--- a/elks/fs/exec.c
+++ b/elks/fs/exec.c
@@ -359,6 +359,7 @@ static int FARPROC execve_aout(struct inode *inode, struct file *filp,
         retval = -ENOMEM;
 #ifdef CONFIG_ROMFS_FS
         if (filp->f_inode->i_sb->s_type->type == FST_ROMFS
+            && !(filp->f_inode->i_mode & S_ISVTX)
             && mh.hlen == EXEC_MINIX_HDR_SIZE) {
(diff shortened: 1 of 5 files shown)


swausd:
Hi @ghaerr 

>@swausd, let me know how this works for you. 

Perfect! Thank you. Amazing!  With ktcp and sh from ROM there now is significant more RAM space free. And romprg now also working. That is really nice. Thank you also to @asiekierka!

> Should you need modifications to romprg, go ahead and submit them in a separate PR.

I see no need for changes to romprg.

PR #2496, opened 2025-12-04
[boot] Fix MINIX boot when /linux disk location is > 32MB on hard drives

ghaerr:
Original code and discussion from https://github.com/Mellvik/TLVC/pull/210 and https://github.com/Mellvik/TLVC/issues/209.

When /linux is located past block number 32767 in MINIX filesystems, the boot would fail since the sector number exceeded 16 bits. This PR fixes that. Obviously, only seen in HD filesystems > 32M in size.

Uses code originally developed by @Mellvik for TLVC, thank you!

During porting to ELKS, the following additional changes were made:
- Adds proper FAT BPB entry for 2880k floppies in FAT boot.
- Saves another byte in boot code by loading BX instead of BL then BH.
- Adds even more ifdefs to code to support PC-98 boot, of which the boot code is entirely out of space. This means that this  HD boot enhancement is not available for PC-98 systems The boot code is getting ridiculously hard to read and probably should be split into separate files.

Also adds code from TLVC ignoring soft errors on floppy reads (error 11h) on MFM drives.

Tested for MINIX and FAT boots in QEMU.

@tyama501, I have tested the PC-98 1232k version on DosBox, and built the 1440k boot. I think everything should still work, but you might try building and testing on Neko if/when you have time. Essentially, the CONFIG_ARCH_PC98 ifdefs in this PR remove the enhancement, since there is zero code space available in the PC98 1232k floppy boot!


tyama501:
I have confirmed fd1440/fd1232/fd1200 works on Neko.

Thank you.

PR #2497, opened 2025-12-05
[ktcp] recover from system hanging after lost TCP packet

swausd:
Discussed here https://github.com/ghaerr/elks/issues/2487

Error description: After loosing a TCP packet in a telnet session the ELKS system hangs. Cause for the packet loss was a router with defect hardware/software. 

With the patch created by @Mellvik makes the system recover after packet loss. It was tested by me,  @swausd.

For testing @Mellvik created a new "packet loss simulation" in deveth.c

### DIFF
--- a/elkscmd/ktcp/deveth.c
+++ b/elkscmd/ktcp/deveth.c
@@ -143,6 +143,15 @@ void eth_sendpacket(unsigned char *packet, int len, eth_addr_t eth_addr)
 	memcpy(ipll->ll_eth_src, eth_local_addr, 6);
 	ipll->ll_type_len = 0x08; //FIXME what is 0x0800
 
+//#define FORCE_MISSING_PACKET 128        /* 128 and 512 are good values for debugging */
+
+#if FORCE_MISSING_PACKET        /* For debugging: Create errors */
+        static failcnt = 0;
+        if (failcnt++ > FORCE_MISSING_PACKET)
+                failcnt = 0;
+        else
+#endif
+
 	eth_write((unsigned char *)ipll, sizeof(struct ip_ll) + len);
 }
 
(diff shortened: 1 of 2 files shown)


elkscmd/ktcp/deveth.c
```diff
@@ -143,6 +143,15 @@ void eth_sendpacket(unsigned char *packet, int len, eth_addr_t eth_addr)
 	memcpy(ipll->ll_eth_src, eth_local_addr, 6);
 	ipll->ll_type_len = 0x08; //FIXME what is 0x0800
 
+//#define FORCE_MISSING_PACKET 128        /* 128 and 512 are good values for debugging */
```
Mellvik: @swausd - you may want to remove the comment-slashes and set the value of FORCE_MISSING_PACKET to 0 instead.

elkscmd/ktcp/deveth.c
```diff
@@ -143,6 +143,15 @@ void eth_sendpacket(unsigned char *packet, int len, eth_addr_t eth_addr)
...
 
+#define FORCE_MISSING_PACKET 0  /* 128 and 512 are good values for debugging */
+
+#if FORCE_MISSING_PACKET        /* For debugging: Create errors */
+        static failcnt = 0;
+        if (failcnt++ > FORCE_MISSING_PACKET)
+                failcnt = 0;
+        else
+#endif
+
```
ghaerr: Can you move the FORCE_MISSING_PACKET to the top of the file, just under the \#includes? That'll allow others to more easily discern it's configurable.

Also, please add "int" to the static declaration - many compilers now complain about implicit int. Is the (tab) formatting matching on the new code? A sad reality is that ELKS is littered with the disastrous 4-character space indentation, then single 8-character TAB for the next indentation, causing (at least to me) huge grief. It appears the indentation may not match. (I've actually gone to the trouble of mapping 4 special keys in `vi`, two to set 4 vs 8 character tabs, and another 2 for expand/noexpand tabs. Without that, I'm unable to read half the source. I'm working on fixing all that to use spaces only *no* tabs, but have left ktcp unchanged for easier diffs with TLVC).

Finally, please delete line 154, as it may confound the reader consider the dangling else on the line above.

swausd: @ghaerr made the changes.

PR #2506, opened 2025-12-08
[kernel] Implement bottom half interrupt handlers!

ghaerr:
This PR implements Linux 2.0-style "bottom-half" or "softirq" interrupt handlers into the ELKS kernel. This is a major enhancement, that with proper reprogramming of various device drivers, allows interrupt handlers to be effectively split in half, where the "top" portion runs under the "original" interrupt context for extreme hardware time-sensitive tasks, while less important code can be deferred for execution outside of the 8259 PIC (or other) interrupt priority, which effectively disallows  other hardware devices to be serviced promptly.

The "bottom half" deferred execution handler names by convention end with \_bh, and this PR implements the timer_bh and serial_bh bottom halves for the IRQ 0 hardware timer and IRQ 4 and 3 serial receive interrupt. To do this, the original timer_irq and rs_fast_irq4 (with CONFIG_FAST_IRQ4 now default enabled) routines are halved. The timer_bh bottom half increments jiffies, then calls `mark_bh(TIMER_BH)` which schedules timer_bh for later execution outside the IRQ 0 interrupt. During kernel init, `init_bh(TIMER_BH, timer_bh)` is called to set the function to call when marked.

The backend timer_bh routine, usually run directly after the IRQ 0 EOI (but not always, its complicated) runs with interrupts enabled, just like the interrupt handler still does, but is guaranteed non-reentrant by the kernel. For timer_bh, it runs the kernel timer lists, the CPU usage calculations for `ps`, "pumps" the received serial characters that are queued by the fast IRQ4 interrupt to wake_up waiting tasks, and runs the disk I/O spinwheel. All of these were previously run *during* the PIC's IRQ 0 interrupt, which effectively disabled any other interrupts for quite a long period every 10ms. Now, the timer_bh runs after the IRQ 0 EOI, but remains interruptible (but not reschedule-able) so that fast serial input or network packet interrupts can execute immediately, interrupting the bottom half.

Essentially, the bottom half portion of any interrupt handler now runs outside of interrupt context, but still runs in kernel context, which, just like the kernel, is non-reentrant (except for top-half interrupt handlers, of course, except they're also protected from reentry by the PIC).

Sound complicated? It is. But for the driver writer, it's not too bad, as an old-style handler can be split into two, servicing the hardware (as quick as possible into a queue/buffer), then scheduling the backend to run with mark_bh. It is very important to run the wake_up() etc calls in the bottom half, as wake_up is very time consuming.

A quick discussion of bottom halves in Linux Kernel Internals 2.0 is [here (page 205, Section 7.2.4)](http://www.staroceans.org/kernel-and-driver/Addison%20Wesley%20-%20Linux%20Kernel%20Internals%202Nd%20Ed.pdf). Another [article](https://thinkty.net/general/2024/04/29/bottom_half.html) is available. Note that Linux implements many different types of "bottom halves" - SoftIRQs, Tasklets, Task Queues, Work Queues, and Threaded Interrupts. I've only implemented SoftIRQs, which are very basic but all we need for ELKS.

Also discussed a bit in https://github.com/Mellvik/TLVC/pull/211#issuecomment-3620566893, https://github.com/ghaerr/elks/issues/2457#issuecomment-3551620305,  https://github.com/ghaerr/elks/issues/2457#issuecomment-3551620305 and https://github.com/ghaerr/elks/pull/2502#issuecomment-3623082246.

Right now, include/arch/irq.h defines the following bottom halves:
```
enum {
    TIMER_BH = 0,
    SERIAL_BH,
    MAX_SOFTIRQ
};
```
Which means that a developer may need to add a new type for their own device driver(s), such as NETWORK_IRQ. All this does is allocate another entry in an array of function pointer to call when mark_bh is called. The new function `do_bottom_half` in elks/arch/i86/kernel/softirq.c then loops through the active_bh (mask bits) to see which, if any, bottom halves to call.

@swausd, look at the IBM PC 8250 serial driver in serial-8250.c::rs_fast_irq4(). While a bit complicated since it's called directly from an interrupt vector outside of \_irqit (for speed) and runs with SS != DS, it effectively just queues a received character during serial interrupt pre-EOI period, then the timer interrupt bottom half (every 10ms) ends up calling serial_bh that calls wake_up. This can be done slightly differently, as noted in the driver, where mark_bh(SERIAL_BH) is called which will then run the wake_up directly after the hardware interrupt. Let me know if you have questions.

@Mellvik, this will be a little more complicated for the network drivers, since after reviewing them, I notice that wake_up and other admin stuff is called potentially in multiple places in the interrupt handler. So you'll have to think about how or whether you can easily split them into two parts. Note this is entirely optional though, and the NIC interrupt is much lower priority than serial or timer, so you might get away with doing nothing, since the serial and timer interrupts now run much quicker.

Enough theory for now. I've tested this only on QEMU, and ~think~hope that it will work well on real hardware. I'd like to see if either of you can test this PR on real hardware before I commit, just in case, since going backwards will be harder. I've tested it pretty hard throughout the day and it seems pretty solid, since the basic hardware interrupt path remains unchanged.

Finally, this PR fixes a terribly broken divide-by-zero handler, which now works from both kernel and user mode (displays DIVIDE FAULT and kills the program), and also introduces an NMI handler (displays NMI FAULT).

Feel free to ask questions.

### DIFF
--- a/elks/arch/i86/drivers/char/serfast.S
+++ b/elks/arch/i86/drivers/char/serfast.S
@@ -1,9 +1,9 @@
-// Fast custom serial input interrupt routines for ELKS
+// First part of top half of fast serial interrupt handlers for ELKS
 //
-// runs on any stack and skips all ELKS overhead
-// must run with interrupts disabled as could interrupt user, kernel or interrupt stack
-// reads character into ring buffer
-// timer interrupt runs serial_bh() which checks for non-zero queue and calls wake_up
+// Runs on any stack and skips ELKS _irqit stack switching overhead.
+// Must run with interrupts disabled as could interrupt user, kernel or interrupt stack.
+// Calls the second part of the top half (rs_fast_irq4) to read and queue received byte.
+// The bottom half (serial_bh) runs later, which processes queue and calls wake_up.
 //
 // 25 June 2020 Greg Haerr
 //
@@ -13,7 +13,7 @@
 
 #ifdef CONFIG_FAST_IRQ4
 //
-// fast ttyS0 interrupt routine, called by CALLF within dynamic handler
+// Entry for ttyS0 top half interrupt handler, called by CALLF within dynamic handler
 //
 	.extern	rs_fast_irq4
 	.global asm_fast_irq4
@@ -29,7 +29,7 @@ asm_fast_irq4:
 	mov	%sp,%bx
 	mov	%ss:12(%bx),%ds
 
-	call	rs_fast_irq4		// call special C interrupt routine
+	call	rs_fast_irq4		// call special 2nd part of top half C handler
 					// which doesn't use any SS/SP/BP addressing
 
 	mov	$0x20,%al		// EOI on primary controller
@@ -46,7 +46,7 @@ asm_fast_irq4:
 
 #ifdef CONFIG_FAST_IRQ3
 //
-// fast ttyS1 interrupt routine, called by CALLF within dynamic handler
+// Entry for ttyS1 top half interrupt handler, called by CALLF within dynamic handler
 //
 	.extern	rs_fast_irq3
 	.global	asm_fast_irq3
@@ -62,7 +62,7 @@ asm_fast_irq3:
 	mov	%sp,%bx
 	mov	%ss:12(%bx),%ds
 
-	call	rs_fast_irq3		// call special C interrupt routine
+	call	rs_fast_irq3		// call special 2nd part of top half C handler
 					// which doesn't use any SS/SP/BP addressing
 
 	mov	$0x20,%al		// EOI on primary controller
--- a/elks/arch/i86/drivers/char/serfast.S
+++ b/elks/arch/i86/drivers/char/serfast.S
@@ -3,7 +3,7 @@
 // runs on any stack and skips all ELKS overhead
 // must run with interrupts disabled as could interrupt user, kernel or interrupt stack
 // reads character into ring buffer
-// timer interrupt runs rs_pump() which checks for non-zero queue and calls wake_up
+// timer interrupt runs serial_bh() which checks for non-zero queue and calls wake_up
 //
 // 25 June 2020 Greg Haerr
 //
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -250,6 +250,13 @@ void rs_fast_irq4(void)
     /* unfortunately, can't add more specifics w/o compiler generating BP accesses */
     if (c == 03)                        /* assumes VINTR = ^C and byte queued anyways */
         sp->intrchar = c;
+    /*
+     * This will mark the serial bottom half serial_bh to run very soon after
+     * this interrupt, which then calls wake_up() on every character received.
+     * For faster baud rates and less data loss, don't run bottom half here,
+     * but call serial_bh directly from timer bottom half in timer_bh.
+     */
+    //mark_bh(SERIAL_BH);
 }
 #endif
 
@@ -272,6 +279,7 @@ void rs_fast_irq3(void)
     /* unfortunately, can't add more specifics w/o compiler generating BP accesses */
     if (c == 03)                        /* assumes VINTR = ^C and byte queued anyways */
         sp->intrchar = c;
+    //mark_bh(SERIAL_BH);               /* see comment in rs_fast_irq4 */
 }
 #endif
 
@@ -298,8 +306,8 @@ static void pump_port(struct serial_info *sp)
     }
 }
 
-/* called from timer interrupt - check ring buffer and wakeup waiting processes */
-void rs_pump(void)
+/* serial interrupt bottom half - check ring buffer and wakeup waiting processes */
+void serial_bh(void)
 {
 #if defined(CONFIG_FAST_IRQ4)
     pump_port(&ports[0]);
@@ -372,12 +380,14 @@ static int rs_open(struct tty *tty)
 #if defined(CONFIG_FAST_IRQ4)
     case 4:
         port->intrchar = 0;
+        init_bh(SERIAL_BH, serial_bh);
         err = request_irq(port->irq, (irq_handler) asm_fast_irq4, INT_SPECIFIC);
         break;
 #endif
 #if defined(CONFIG_FAST_IRQ3)
     case 3:
         port->intrchar = 0;
+        init_bh(SERIAL_BH, serial_bh);
         err = request_irq(port->irq, (irq_handler) asm_fast_irq3, INT_SPECIFIC);
         break;
 #endif
--- a/elks/arch/i86/drivers/char/serial-8250.c
+++ b/elks/arch/i86/drivers/char/serial-8250.c
@@ -219,18 +219,28 @@ static int rs_write(struct tty *tty)
 
 #if defined(CONFIG_FAST_IRQ4)
 /*
- * Second half of fast serial input interrupt handler for slower machines.
- * Should work up to 38400 baud.
- * Incomplete tty signal handling, will generate SIGINT when VINTR = ^C.
+ * Serial interrupt top half. This top half actually consists of two parts.
+ * The first part of the top half of the handler is asm_fast_irq4 in serfast.S,
+ * and the function below is the second part of the top half, rs_fast_irq4.
+ *
+ * The first part asm_fast_irq4 is called directly from the IRQ 4 interrupt
+ * vector, bypassing the normal kernel stack switch code in _irqit. That code
+ * runs with interrupts disabled and saves registers AX,BX,CX,DX,DS, and
+ * sets DS to the kernel data segment. The stack segment is not changed,
+ * so SS != DS, then it calls the second part rs_fast_irq4 C code below.
+ *
+ * NOTE: Since the compiler emits BP addressing for parameters, no parameters can
+ * be passed, nor any code written which emits code using SP or BP addressing, as
+ * SS is not changed and not guaranteed to be equal to DS.
+ * Use 'ia16-elf-objdump -D -r -Mi8086 serial-8250.o' to look at code generated.
+ *
+ * Thus, the rs_fast_irq4 function below is a specially-coded C top half interrupt
+ * handler, subject to the above limitations, which just reads and queues the
+ * UART byte received, with the serial_bh bottom half run after the IRQ 4 EOI.
+ * As a result, it should handle speeds up to 38400 baud.
+ *
+ * Incomplete tty signal handling, generates SIGINT when VINTR = ^C.
  * Useful for fast SLIP transfer or arrow key input on slow systems.
- *
- * Specially-coded C interrupt handler, called from asm_fast_irq[43] with
- * interrupts disabled after saving scratch registers AX,BX,CX,DX & DS and
- * setting DS to kernel data segment.
- * NOTE: no parameters can be passed, nor any code written which emits code using
- * SP or BP addressing, as SS is not set and not guaranteed to equal DS.
- *
- * Use 'ia16-elfk-objdump -D -r -Mi8086 serial.o' to look at code generated.
  */
 extern void asm_fast_irq4(int irq, struct pt_regs *regs);   /* initial entry point */
 
@@ -251,8 +261,8 @@ void rs_fast_irq4(void)
     if (c == 03)                        /* assumes VINTR = ^C and byte queued anyways */
         sp->intrchar = c;
     /*
-     * This will mark the serial bottom half serial_bh to run very soon after
-     * this interrupt, which then calls wake_up() on every character received.
+     * This will mark the serial bottom half serial_bh to run after an _irqit
+     * interrupt, which then calls wake_up() on every character received.
      * For faster baud rates and less data loss, don't run bottom half here,
      * but call serial_bh directly from timer bottom half in timer_bh.
      */
@@ -296,7 +306,7 @@ static void pump_port(struct serial_info *sp)
             sp->intrchar = 0;
         }
         if (q->len == 1) {
-            /* attempt process (again) of VINTR, also VQUIT and ^P-^N debug chars */
+            /* attempt process (again) of VINTR, also VQUIT and ^N^O^P debug chars */
             if (tty_intcheck(ttyp, chq_peekch(q))) {
                 (void)chq_getch(q);     /* discard received character */
                 return;
--- a/elks/arch/i86/kernel/Makefile
+++ b/elks/arch/i86/kernel/Makefile
@@ -39,16 +39,15 @@ OBJS  = strace.o irq.o irqtab.o process.o \
 		entry.o signal.o timer.o softirq.o
 
 ifeq ($(CONFIG_ARCH_IBMPC), y)
-OBJS += irq-8259.o reset-ibm.o timer-8254.o nmi.o
-#OBJS += divzero.o
+OBJS += irq-8259.o reset-ibm.o timer-8254.o divzero.o nmi.o
 endif
 
 ifeq ($(CONFIG_ARCH_PC98), y)
-OBJS += irq-8259.o reset-pc98.o timer-8254.o nmi.o
+OBJS += irq-8259.o reset-pc98.o timer-8254.o divzero.o nmi.o
 endif
 
 ifeq ($(CONFIG_ARCH_SOLO86), y)
-OBJS += irq-solo86.o reset-stubs.o timer-8254.o nmi.o
+OBJS += irq-solo86.o reset-stubs.o timer-8254.o divzero.o nmi.o
 endif
 
 ifeq ($(CONFIG_ARCH_8018X), y)
@@ -60,7 +59,7 @@ OBJS += irq-necv25.o reset-stubs.o timer-necv25.o
 endif
 
 ifeq ($(CONFIG_ARCH_SWAN), y)
-OBJS += irq-swan.o reset-swan.o timer-swan.o system-swan.o nmi.o
+OBJS += irq-swan.o reset-swan.o timer-swan.o system-swan.o divzero.o nmi.o
 else
 OBJS += system.o
 endif
--- a/elks/arch/i86/kernel/Makefile
+++ b/elks/arch/i86/kernel/Makefile
@@ -36,16 +36,19 @@ include $(BASEDIR)/Makefile-rules
 # Objects to be compiled.
 
 OBJS  = strace.o irq.o irqtab.o process.o \
-		entry.o signal.o timer.o
+		entry.o signal.o timer.o softirq.o
 
 ifeq ($(CONFIG_ARCH_IBMPC), y)
-OBJS += irq-8259.o reset-ibm.o timer-8254.o
-OBJS += divzero.o
+OBJS += irq-8259.o reset-ibm.o timer-8254.o nmi.o
+#OBJS += divzero.o
 endif
 
 ifeq ($(CONFIG_ARCH_PC98), y)
-OBJS += irq-8259.o reset-pc98.o timer-8254.o
-OBJS += nmi.o
+OBJS += irq-8259.o reset-pc98.o timer-8254.o nmi.o
+endif
+
+ifeq ($(CONFIG_ARCH_SOLO86), y)
+OBJS += irq-solo86.o reset-stubs.o timer-8254.o nmi.o
 endif
 
 ifeq ($(CONFIG_ARCH_8018X), y)
@@ -57,15 +60,9 @@ OBJS += irq-necv25.o reset-stubs.o timer-necv25.o
 endif
 
 ifeq ($(CONFIG_ARCH_SWAN), y)
-OBJS += irq-swan.o reset-swan.o system-swan.o timer-swan.o
-OBJS += divzero.o
+OBJS += irq-swan.o reset-swan.o timer-swan.o system-swan.o nmi.o
 else
 OBJS += system.o
-endif
-
-ifeq ($(CONFIG_ARCH_SOLO86), y)
-OBJS += irq-solo86.o reset-stubs.o timer-8254.o
-OBJS += divzero.o
 endif
 
 #########################################################################
--- a/elks/arch/i86/kernel/divzero.c
+++ b/elks/arch/i86/kernel/divzero.c
@@ -4,21 +4,19 @@
 
 /*
  * Divide by zero and divide overflow exception handler
- * NOTE: This handler is currently unused, see div0_handler_panic in irqtab.S
  *
  * 19 Aug 24 Greg Haerr
  */
 
-static char div0msg[] = { "Divide fault\n" };
+#define DEBUG       0   /* =1 for CS:IP detail of fault */
 
-void div0_handler(int i, struct pt_regs *regs)
+void div0_handler(int irq, struct pt_regs *regs)
 {
-    /* divide by 0 from nested interrupt or idle task means kernel code was executing */
     if (intr_count > 1 /*|| current->t_regs.ss == kernel_ds*/) {
         /*
-         * Trap from kernel code.
+         * Trap from kernel code or idle task.
          *
-         * Not panicing at this point involves determining the CS:IP of the
+         * Not calling panic at this point involves determining the CS:IP of the
          * faulting instruction, then doing two different things depending on
          * whether the CPU is 8088/8086/V20/V30 or 286/386+: the former trap
          * pushes the CS:IP following the DIV, while the latter pushes CS:IP of
@@ -26,20 +24,23 @@ void div0_handler(int i, struct pt_regs *regs)
          * all forms of DIV, while allowing either would introduce undefined
          * behaviour as a result of an incorrect calcuation. So we panic.
          */
-#if 0
+
+#if DEBUG
         struct uregs __far *sys_stack;
         sys_stack = _MK_FP(regs->ss, regs->sp);
-        printk("Div0 at CS:IP %x:%x\n", sys_stack->cs, sys_stack->ip);
+        printk("CS:IP %04x:%04x\n", sys_stack->cs, sys_stack->ip);
 #endif
-        panic(div0msg);
-    } else {
-        /* For user mode faults, display error message and kill the process */
-        printk(div0msg);
-#if 0
+
+        panic("DIVIDE FAULT\n");
+    } else {    /* For user mode faults, display error message and kill the process */
+        printk("DIVIDE FAULT\n");
+
+#if DEBUG
         struct uregs __far *user_stack;
         user_stack = _MK_FP(current->t_regs.ss, current->t_regs.sp);
-        printk("Div0 at CS:IP %x:%x\n", user_stack->cs, user_stack->ip);
+        printk("CS:IP %04x:%04x\n", user_stack->cs, user_stack->ip);
 #endif
+
         sys_kill(current->pid, SIGABRT);    /* no SIGFPE so send SIGABRT for now */
     }
 }
--- a/elks/arch/i86/kernel/divzero.c
+++ b/elks/arch/i86/kernel/divzero.c
@@ -4,6 +4,7 @@
 
 /*
  * Divide by zero and divide overflow exception handler
+ * NOTE: This handler is currently unused, see div0_handler_panic in irqtab.S
  *
  * 19 Aug 24 Greg Haerr
  */
--- a/elks/arch/i86/kernel/irq.c
+++ b/elks/arch/i86/kernel/irq.c
@@ -120,24 +120,15 @@ int free_irq(int irq)
  */
 void INITPROC irq_init(void)
 {
-    /* use INT 0x80h for system calls */
-    int_handler_add(IDX_SYSCALL, 0x80, _irqit);
+    int_handler_add(IDX_SYSCALL, 0x80, _irqit); /* INT 80 for system calls */
 
 #if defined(CONFIG_ARCH_IBMPC) || defined(CONFIG_ARCH_PC98) || \
     defined(CONFIG_ARCH_SOLO86) || defined(CONFIG_ARCH_SWAN)
 
-#if 1
-    /* install direct panic-only DIV fault handler until known that
-     * the _irqit version doesn't overwrite the stack
-     */
-    int_handler_add(IDX_DIVZERO, 0x00, div0_handler_panic);
-#else
-    /* catch INT 0 divide by zero/divide overflow hardware fault */
-    irq_action[IDX_DIVZERO] = div0_handler;
+    irq_action[IDX_DIVZERO] = div0_handler;     /* INT 0 divide by 0/divide overflow */
     int_handler_add(IDX_DIVZERO, 0x00, _irqit);
-#endif
 
-    irq_action[IDX_NMI] = nmi_handler;              /* catch NMI (INT 2) */
+    irq_action[IDX_NMI] = nmi_handler;          /* INT 2 non-maskable interrupt */
     int_handler_add(IDX_NMI, 0x02, _irqit);
 #endif
 
--- a/elks/arch/i86/kernel/irq.c
+++ b/elks/arch/i86/kernel/irq.c
@@ -123,22 +123,21 @@ void INITPROC irq_init(void)
     /* use INT 0x80h for system calls */
     int_handler_add(IDX_SYSCALL, 0x80, _irqit);
 
-#if defined(CONFIG_ARCH_IBMPC) || defined(CONFIG_ARCH_SWAN)
-    /* catch INT 0 divide by zero/divide overflow hardware fault */
+#if defined(CONFIG_ARCH_IBMPC) || defined(CONFIG_ARCH_PC98) || \
+    defined(CONFIG_ARCH_SOLO86) || defined(CONFIG_ARCH_SWAN)
+
 #if 1
     /* install direct panic-only DIV fault handler until known that
      * the _irqit version doesn't overwrite the stack
      */
     int_handler_add(IDX_DIVZERO, 0x00, div0_handler_panic);
 #else
+    /* catch INT 0 divide by zero/divide overflow hardware fault */
     irq_action[IDX_DIVZERO] = div0_handler;
     int_handler_add(IDX_DIVZERO, 0x00, _irqit);
 #endif
-#endif
 
-#if defined(CONFIG_ARCH_PC98)
-    /* catch NMI */
-    irq_action[IDX_NMI] = nmi_handler;
+    irq_action[IDX_NMI] = nmi_handler;              /* catch NMI (INT 2) */
     int_handler_add(IDX_NMI, 0x02, _irqit);
 #endif
 
@@ -165,6 +164,7 @@ void INITPROC irq_init(void)
     if (request_irq(TIMER_IRQ, timer_tick, INT_GENERIC))
         panic("Unable to get timer");
 
+    init_bh(TIMER_BH, timer_bh);
     enable_timer_tick();        /* reprogram timer for 100 HZ */
 #endif
 }
--- a/elks/arch/i86/kernel/irqtab.S
+++ b/elks/arch/i86/kernel/irqtab.S
@@ -100,6 +100,8 @@ int_vector_set:
         .extern trace_begin
         .extern trace_end
         .extern panic
+        .extern bh_active
+        .extern do_bottom_half
 
         .global _irqit
 _irqit:
@@ -332,14 +334,42 @@ was_trap:
 //
 //      Look at rescheduling
 //
-        cmpw    $1,intr_count
+        cmpw    $1,intr_count   // Syscall or interrupted user mode code?
         jne     restore_regs    // No
 //      cmp     $0,_need_resched // Schedule needed ?
 //      je      restore_regs    // No
 //
 // This path will return directly to user space
 //
-        sti                     // Enable interrupts to help fast devices
+
+//
+// Check for any active bottom-half handlers and if we're able to run them now.
+//  Bottom half handlers can run now only if this interrupt is from user mode,
+//  otherwise handlers will be run during next call to schedule().
+//  Interrupts from the idle task have to return now since they're running on
+//  the interrupt stack and can't be switched. On return the idle task will call
+//  schedule() in its main loop, which then runs the bottom half handlers.
+//
+//  At this time, intr_count values mean:
+//  intr_count==0   not possible
+//  intr_count==1   syscall or interrupted user code (using process's kernel stack)
+//  intr_count==2   interrupted kernel code or idle task (using interrupt stack)
+//  intr_count>2    interrupted handler code (using interrupt stack)
+//  The idle task always runs at intr_count==1, since it is actually kernel code.
+//  This means that interrupts of idle task or kernel run on the interrupt stack
+//  and thus calling schedule() cannot be allowed. This also means that
+//  task rescheduling doesn't happen on return from a timer interrupt from the idle task.
+//  Instead, the idle task is resumed and it calls schedule which calls bottom half.
+//
+        cmpw    $0,bh_active    // Any active bottom halfs?
+        je      1f              // No
+        //incw    intr_count    // Don't rerun bottom half if interrupted
+        sti                     // interrupts enabled but handlers not reentrant
+        call    do_bottom_half
+        //cli
+        //decw    intr_count
+
+1:      sti                     // Enable interrupts to help fast devices
         call    schedule        // Task switch
         call    do_signal       // Check signals
         cli
@@ -401,6 +431,10 @@ setsp:
         pop     %ax
         mov     %ax,%sp
         jmp     *%bx
+
+        .global getsp
+getsp:  mov     %sp,%ax
+        ret
 
 // Halt - wait for next interrupt to save CPU power
         .global idle_halt
--- a/elks/arch/i86/kernel/irqtab.S
+++ b/elks/arch/i86/kernel/irqtab.S
@@ -334,7 +334,7 @@ was_trap:
 //
 //      Look at rescheduling
 //
-        cmpw    $1,intr_count   // Syscall or interrupted user mode code?
+        cmpw    $1,intr_count   // Interrupted user mode code?
         jne     restore_regs    // No
 //      cmp     $0,_need_resched // Schedule needed ?
 //      je      restore_regs    // No
--- a/elks/arch/i86/kernel/irqtab.S
+++ b/elks/arch/i86/kernel/irqtab.S
@@ -352,7 +352,7 @@ was_trap:
 //
 //  At this time, intr_count values mean:
 //  intr_count==0   not possible
-//  intr_count==1   syscall or interrupted user code (using process's kernel stack)
+//  intr_count==1   interrupted user code (using process's kernel stack)
 //  intr_count==2   interrupted kernel code or idle task (using interrupt stack)
 //  intr_count>2    interrupted handler code (using interrupt stack)
 //  The idle task always runs at intr_count==1, since it is actually kernel code.
@@ -432,8 +432,10 @@ setsp:
         mov     %ax,%sp
         jmp     *%bx
 
+// short *getsp(void) - get stack pointer
         .global getsp
-getsp:  mov     %sp,%ax
+getsp:
+        mov     %sp,%ax
         ret
 
 // Halt - wait for next interrupt to save CPU power
@@ -441,35 +443,6 @@ getsp:  mov     %sp,%ax
 idle_halt:
         hlt
         ret
-
-        .global div0_handler_panic
-// Divide Fault hander - just panic for now
-div0_handler_panic:
-        push    %ax                     // save regs, uses 4+4+10 bytes of current stack
-        push    %bx
-        push    %cx
-        push    %dx
-        push    %ds
-
-        // Recover kernel data segment
-        // Was pushed by the CALLF of the dynamic handler
-        mov     %sp,%bx
-        mov     %ss:12(%bx),%ds
-
-        mov     $dmsg,%ax
-        push    %ax
-        call    panic
-        pop     %ax
-1:      hlt
-        jmp     1b
-
-//      pop     %ds                     // restore regs
-//      pop     %dx
-//      pop     %cx
-//      pop     %bx
-//      pop     %ax
-//      add     $4,%sp                  // skip the trampoline lcall
-//      iret
 
         .data
         .global intr_count
--- a/elks/arch/i86/kernel/irqtab.S
+++ b/elks/arch/i86/kernel/irqtab.S
@@ -462,7 +462,6 @@ intr_count:                     // stacked interrupts count. Start with 1
 #ifdef CHECK_SS
 pmsg:   .ascii "Running unknown code\0"
 #endif
-dmsg:   .ascii  "DIVIDE FAULT\0"
 
         .bss
         .p2align 1
--- a/elks/arch/i86/kernel/nmi.c
+++ b/elks/arch/i86/kernel/nmi.c
@@ -6,7 +6,7 @@
  * NMI handler
  */
 
-void nmi_handler(int i, struct pt_regs *regs)
+void nmi_handler(int irq, struct pt_regs *regs)
 {
     printk("NMI FAULT\n");
 
--- a/elks/arch/i86/kernel/nmi.c
+++ b/elks/arch/i86/kernel/nmi.c
@@ -1,15 +1,14 @@
 #include <linuxmt/config.h>
+#include <linuxmt/kernel.h>
 #include <arch/io.h>
 
 /*
  * NMI handler
  */
 
-static char nmi_msg[] = { "NMI occurred\n" };
-
 void nmi_handler(int i, struct pt_regs *regs)
 {
-    printk(nmi_msg);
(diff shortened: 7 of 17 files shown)


swausd:
Hi @ghaerr, sounds like an interesting major change in concept of interrupt handling. I will look into it starting this afternoon.
And thanks for the two external links to further information! Much to learn…

swausd:
Hi @ghaerr ,

I'm struggling to understand the flow of control in TH/BH drivers. The comment **Second half...**  in serial-8250.c at line 220 is a little distracting.

```
#if defined(CONFIG_FAST_IRQ4)
/*
 * Second half of fast serial input interrupt handler for slower machines.
.
.
extern void asm_fast_irq4(int irq, struct pt_regs *regs);   /* initial entry point */

void rs_fast_irq4(void)
{
```
Is `rs_fast_irq4` the `second half` of the `top half` of interrupt serial handling, while `asm_fast_irq4` is the `first half` of the `top half`?

ghaerr:
@swausd 
The fast serial driver is a bit more complicated than breaking an interrupt driver in half. The basic (new) idea is that the "top half" is called by the (original) request_irq call, and runs as a normal interrupt handler. Its purpose, rs_fast_irq4 in this instance, is to perform hardware data acquisition by reading the UART, queuing the character, and then calling mark_bh to flag the bottom half to run later (which performs the wakeups and tty_intcheck, etc).

In the fast serial driver only, instead of running the "normal" serial bottom half serial_bh after *every* serial interrupt, the serial bottom half serial_bh is called directly from timer.c::timer_bh. That is, every 10ms the timer_bh is run, and that then deals with *all* received serial characters at once, rather than calling wake_up every serial interrupt in the bottom half.

I hope that doesn't sound too complicated.

The simpler version could be that you uncomment the mark_bh in rs_fast_irq4, and comment out the call to serial_bh in timer_bh. For your slow system, you'll want to model the way is now: just write a top half normal serial interrupt handler, and have the timer bottom half call wakeup by calling serial_bh directly.

> Is rs_fast_irq4 the second half of the top half of interrupt serial handling, while asm_fast_irq4 is the first half of the top half?

Yes - but that's only because running the serial interrupt all the way through \_irqit is just too slow, so a special first half of the top half was written from scratch: asm_fast_irq4 skips \_irqit, and thus doesn't switch stacks to the kernel stack. Because of this, the C code in the second part of the top half won't work if the compiler generates any code that accesses through the BP register, since the 8086 uses SS: as the default segment register in that case (which != DS).

At the end of the day, you should be able to use the exact same code as asm_fast_irq4 (renamed) for your V25 first part of the top half, and the exact same code as rs_fast_irq4 (renamed) as the top half, with only the UART inb changed. The serial_bh bottom half can probably remain very similar as well.

We should probably use the nomenclature of "first part" and "second part" of the first half of the serial driver, since saying first half of the first half is confusing, to say the least. The fast serial driver is the result of a lot of effort and normally this complexity is not required at all. It was written to allow very high frequency input, far more than running an interrupt through \_irqit will every be able to handle.

ghaerr:
> The comment Second half... in serial-8250.c at line 220 is a little distracting.

I'll update that comment to say "Second part of the top half...", thank you.

ghaerr:
BTW @swausd, you should be able to run the code in this PR without change on your V25; the bottom half interrupt handler operation is not mandatory. That is, all previous interrupt handlers can continue to run as-is without registering and marking bottom halves to run - the system should operate just as before. I would be interested to hear whether your system would boot and run without modification. The timer interrupt handler has been halved, but the 8250 serial driver isn't used on the V25, of course.

swausd:
Hi @ghaerr ,

I successfully changed my console-serial-necv25.c to use the new irq-driver concept in a first step (without understanding everything). Arrow keys are working in the shell history up to 57600 baud - even when moving left/right fast. Nice!
This it what I did:

In console-serial-necv25.c
1. Removed  tty_intcheck() and chq_addch() from irq_rx()
2. Added character queuing like in rs_fast_irq4() in irq_rx()
3. Added mark_bh(SERIAL_BH) in irq_rx()
4. Added  init_bh(SERIAL_BH, serial_bh) in sercon_open()
5. Added pump_port() and serial_bh()
6. Added intrchar to struct serial_info
--
7. Activated chq_peekch() in chqueueu.c
8.  Activated serial_bh() in timer_bh() in timer.c

I now have a working system again. Tomorrow I will try to get deeper into details to understand what I did ;-) and to try to optimize it.

Thank you for this huge increase in usability!

swausd:
@ghaerr 

> the system should operate just as before. I would be interested to hear whether your system would boot and run without modification. 

Yes my system did run unmodified using this PR!

ghaerr:
> Arrow keys are working in the shell history up to 57600 baud - even when moving left/right fast.

> Yes my system did run unmodified using this PR!


Very nice!!! Thank you for your quick testing. This means I think I can commit this PR without problem, so we can both continue to move forward. I am very happy to hear the much higher baud rate you're getting without errors. It would be interested to try sending a large batch of serial input (up to 1024 bytes, the default size of the serial input queue) to see if there's any drops then. That would be effectively testing the 10ms timer interrupt and its (now) background processing, to see if that gets in the way (it shouldn't).

IIRC your V25 has the serial IRQ at the highest priority, so the above timer interrupt probably isn't an issue. With the IBM PC architecture, there's a potential ongoing issue with the fact that \_irqit calls the original BIOS IRQ 0 timer vector every 5th call through. This is because DOS systems run at 55ms timer and ELKS is 10ms, and is primarily used to turn off the floppy drive motors by the BIOS when using the ELKS BIOS floppy driver. That's more pre-EOI overhead for IBM PC that I haven't figured a way to remove (other than running the direct FD driver, of course).

>This it what I did:

Super! I would say that you've figured a lot out, even if you think you don't fully understand it. The good news is one doesn't have to understand exactly how/when bottom halves are run in order to write a BH handler. Each step you've taken looks completely correct. And now you've got what I think is the fastest serial input possible - since the byte queuing is all handled in the serial top half, and no wakeup processing called until max 10ms later in the timer interrupt, and then only a single wakeup for all the received bytes. I'd be interested to learn how fast you can actually receive characters before dropping. With the V25 serial interrupt being higher priority than the timer, you may be able to burst-mode serial without a single timer interrupt until stopping. It will be interesting to play with this.

> Activated chq_peekch() in chqueueu.c

I'll add that routine as standard so it won't have to be ifdef'd.

Definitely post a PR with your changes, I'll be committing this PR shortly, thank you!

BTW, I have a plan for the system determining for us when the input interrupt frequency exceeds the system's ability to handle it, accounting for both top and bottom half execution. I'll be putting that together in a subsequent PR.

swausd:
Hi @ghaerr ,

I have some questions:
1. Why start `serial_bh()` in `timer_bh()` for FAST_IRQ and not using `mark_bh()` ?
2. Why only catch VINTR in rs_fast_irq4/3() and not VSUSP and VQUIT as well?
3. Why call tty_intcheck() two times - at least for VINTR ?

I implemented FAST_IRQ for my NECV25 system according to rs_fast_irq4/3() in serail-8250.c. This is a little bit faster as my TH/BH implementation without FAST_IRQ. Now the cursor keys are working in the shell with 115200 baud. **One** line cut&paste in the terminal emulation works as well, too.

> I'd be interested to learn how fast you can actually receive characters before dropping. 

This question is difficult to answer. On a system with (for example continuous network) load, characters **can** get lost, if entered via cut&paste (my system doesn't have a serial fifo).  That wouldn't bother me, as I rather like a responsive system under "normal" condition - that is manual typing.

I think in addition to "no receive fifo", the console echo (serial out of received characters) is disturbing serial reception on an idle system.

So how do you like me to measure `how fast you can actually receive`?

ghaerr:
Hello @swausd,

> Why start serial_bh() in timer_bh() for FAST_IRQ and not using mark_bh() ?

Originally because I thought it faster to just call the serial BH handler directly. It might be easier to call mark_bh(SERIAL_BH), but this would case 1) a separate run of a bottom half handler to run, involving another for loop in do_bottom_half, and 2) could cause the same for loop for an unregistered handler (doing nothing) if there is no serial driver compiled in. 

Your idea does have the benefit of not requiring a preset serial_bh name for the serial BH handler. It is probably better design to use mark_bh; but why add constant unnecessary overhead every 10ms?

> Why only catch VINTR in rs_fast_irq4/3() and not VSUSP and VQUIT as well?

I ran into a problem with the GCC code being emitted in rs_fast_irq4: the compiler emitted a PUSH BP which allowed it to allocate stack space for local variables, then used -2(BP) etc to access them. The problem is that rs_fast_irq4 runs with SS != DS, and SS could be anything, so it won't work. The `sp->intrchar` is somewhat of a trick, which allows guaranteed VINTR signals, even when q->len > 1 (unlikely on keyboard though).

Also, I wanted the top half to run as fast as possible, without spending valuable time checking very rarely used characters such as VSUSP and VQUIT.

> Why call tty_intcheck() two times - at least for VINTR ?

This is more of the trick code described above, attempting to guarantee processing of VINTR when q->len != 1. You're welcome to rewrite that code; the idea was to remove the character if it's been processed by tty_intcheck from the input queue.

Using the trick allows for not having to search through every character received in the bottom half handler, especially since VSUSP and VQUIT are very rarely pressed keystrokes.

> On a system with (for example continuous network) load, characters can get lost, if entered via cut&paste (my system doesn't have a serial fifo).

The NIC drivers don't (yet) have a BH hander, so they might be using excessive interrupt time. Will your system lose characters during cut&paste without network activity running? That test would determine if serial + timer interrupt bottom halves have solved the problem they were invented for.

> I think in addition to "no receive fifo", the console echo (serial out of received characters) is disturbing serial reception on an idle system.

Yes - if those console printk's occur during an interrupt, then that can't be easily fixed. Currently, even normal kernel printk's will probably drop serial chars since the normal kernel code will always run to completion before any bottom half handlers are called. I plan on adding a debug kernel precision timer to measure this shortly.

ghaerr:
> the console echo (serial out of received characters) is disturbing serial reception on an idle system.

Actually, even though the kernel console echo code prevents the serial bottom half from running, it should not be stopping the top half from running - so characters should not be dropped. We might need to look more into this.


> So how do you like me to measure how fast you can actually receive?

Using cut&past to send 1000 characters, perhaps text with column numbers to a `cat > file`, would be interesting. I am looking for the highest baud rate setting that allows the text to be received with no errors. You can also run `sercat` on the serial terminal which will set raw mode and no TTY handling, to see if that changes anything.

swausd:
Hi @ghaerr ,

thank you for this quick response!

> I ran into a problem with the GCC code being emitted in rs_fast_irq4: the compiler emitted a PUSH BP which allowed it to allocate stack space for local variables, then used -2(BP) etc to access them.

I modified the `intrchar` part in my FAST_IRQ code to
```

 if ((c == 03) || (c == 26) || (c == 28))    /* assumes VINTR = ^C, VSUSP = ^Z, VQUIT = ^\ and byte queued anyways */
        sp->intrchar = c;
```

and looked at the asm code produced. No use of SP or BP. May be the compiler doesn't try to optimize because of my lengthy inline asm code.
 
> Also, I wanted the top half to run as fast as possible, without spending valuable time checking very rarely used characters such as VSUSP and VQUIT.

That makes sense. I added VSUSP and VQUIT because I couldn't escape otherwise from a `cat /dev/tty1 /mnt/test.txt` Tried to test when characters get lost. But I think terminal echo is disturbing...

 >>   So how do you like me to measure how fast you can actually receive?

>Using cut&past to send 100 characters, perhaps text with column numbers to a cat > file, would be interesting. I am looking for the highest baud rate setting that allows the text to be received with no errors. You can also run sercat on the serial terminal which will set raw mode and no TTY handling, to see if that changes anything.

I will look into this tomorrow.

ghaerr:
> I added VSUSP and VQUIT because I couldn't escape otherwise from a cat /dev/tty1 /mnt/test.txt

Hmmm. You should be able to stop the cat using ^C/VINTR. Is that not working?

> if ((c == 03) || (c == 26) || (c == 28))

Not a bad idea. I was thinking before trying to handle very rarely used characters, start with the serial top half being as fast as possible to determine max baud rate, then add this sort of stuff back in afterwards knowing you can test whether it decreases max baud rate reception with no data loss.

> But I think terminal echo is disturbing...

Agreed that terminal echo is adding a variable we don't need at this point.

Try running `sercat > file` on the serial line. That will turn off TTY echo and redirect to the file. That program was written specifically to test high speed input.

swausd:
Hi @ghaerr ,

> Hmmm. You should be able to stop the cat using ^C/VINTR. Is that not working?

No, I get ` Terminal Stop` and the system hangs. With ^\ I get `Quit`, the system closes the file and returns to the shell.

ghaerr:
> I get  Terminal Stop and the system hangs. With ^\ I get Quit, the system closes the file and returns to the shell.

Something is screwed up. I just tested this program on the (direct) console:
```
#include <unistd.h>

int main(int ac, char **av)
{
    volatile int i = 0;

    for (;;) {
        if ((++i & 0x3FFF) == 0)
            write(1, "X", 1);
    }
}
```
When I type ^Z, the output stops, but the system appears hung. After that, I type ^C, and "Terminal Stop" appears, but the shell is still frozen. Another shell is operational, so the system isn't hung, but the shell seems to be. Meanwhile, ^Z should stop the program and return to the shell immediately, not wait for a SIGINT/^C. 

The fact that you're getting Terminal Stop on VINTR/^C means either your new driver has a small bug in it (saving ^Z instead of ^C, or perhaps not clearing it?), or perhaps the TTY/kernel bug above.

I'm investigating. I'm pretty sure ^Z worked prior to BH interrupts, not sure yet. The proper operation is a shell prompt, then `fg` typed will resume the stopped process when wanted.

When the `write` above is replaced by `continue`, even stranger things happen, and the shell enters an infinite loop with `ps` showing 100% CPU.

ghaerr:
Hello @swausd,


> No, I get  Terminal Stop and the system hangs.

Trying to duplicate that problem led me into a rabbit hole of finding out that terminal-stopped jobs basically didn't work in ELKS. That's now fixed in https://github.com/ghaerr/elks/pull/2509. I'm glad you found this!

> Something is screwed up. 

All that is fixed in #2509.

> You should be able to stop the cat using ^C/VINTR. Is that not working?

However, I can't get any ^C/VINTR keystroke to produce a "Terminal Stop" action unless typing ^Z. So I'm pretty sure this is a bug in your serial driver, please check. I tested pretty thoroughly with both the IBM PC serial driver and IBM PC console. The good news is ^Z handling is now fixed regardless.

If you want to post your new V25 changes as a prelim PR, I can help look at it if you can't find the ^Z/^C mixup issue.

> Using cut&past to send 1000 characters, perhaps text with column numbers to a cat > file

I mistyped the above, I was suggesting sending 1000 bytes quickly, not 100, to test the serial driver for character loss. The default serial receive interrupt buffer is 1024 bytes.

Thank you!

swausd:
Hi @ghaerr ,

> However, I can't get any ^C/VINTR keystroke to produce a "Terminal Stop" action unless typing ^Z.

Sorry for leading you in the wrong direction and deep into a rabbit hole.

You are right! I typed ^Z, because I thought this would be "End Of File". But nothing happened. Then I typed ^C and "Terminal Stop" was returned. Then ^C didn't break the session.

Now I will try your last changes...

swausd:
Hi @ghaerr ,

Tried your changes for VINTR = ^C, VSUSP = ^Z, VQUIT = ^\ and they now work as expected. I noticed I had a typing error above in `cat /dev/tty1 /mnt/test.txt`, it should be with redirected output `cat /dev/tty1 >/mnt/test.txt`

Thank you for the changes and sorry again for pointing you in the wrong direction!

> If you want to post your new V25 changes as a prelim PR

I will make a PR with my changes, but first I have some cleanup to do - like setting the default baud rate, which you also already solved for me!

swausd:
Hi @ghaerr ,

>  You can also run sercat on the serial terminal which will set raw mode and no TTY handling

I did a simple test to check  the speed of my FAST_IRQ implementation on an idle ELKS system using 
`sercat /dev/tty1 >/mnt/file.txt`
 Sending a text file with max. speed (without char or EOL delay) using raw file send of GTKTerm. Hoping that the I/O buffering is used before writing to SD card mounted to /mnt. I only counted the chars up to the first error. This is what I get:

38400 baud: about 10.000 chars, then I think writing to SD card finally happened.
57600 baud: about 430 chars.
115200 baud: exactly 64 chars with 3 runs.

I am now using 115200 baud as default baud rate for my console. This is a nice speed for output and save enough on input with my typing.

ghaerr:
Hello @swausd,

Thanks for the testing. I was kind of hoping for burst rates higher than 38400, but even when I first developed the fast serial driver for my Compaq Portable never was able see reliable input with higher baud rates. Even then, the clock tick would intervene sometimes with data loss, which likely is prevented now with bottom half handling. For UARTs without a FIFO, I'd say we're doing pretty good!

I am thinking that what is happening at higher baud rates, instead of the system crashing via interrupt stack overflow, etc, we're seeing that the system flat can't keep up with the input and the chqueue routine finally just refuses to add the character to the input buffer. That is, the system maintains total control running the serial interrupt handler, timer interrupt, timer and serial bottom halves and application read/writes until the buffer is full. It doesn't appear we're getting a UART receiver overrun. Nice!

> Hoping that the I/O buffering is used before writing to SD card mounted to /mnt. 

Your SD card driver isn't interrupt driven, right? Does it need interrupts disabled for larger periods? 

> Sorry for leading you in the wrong direction and deep into a rabbit hole.

No worries at all! I enjoy the deep dives; in this case the process stop/continue mechanism was enhanced, opening the door for application profiling via a debugger or other interesting uses. I am bit surprised users haven't got more confused about ^Z usage given its usage as DOS EOF.

swausd:
Hi @ghaerr ,

> and the chqueue routine finally just refuses to add the character to the input buffer

I don't think that chqueue is full on my system. May be I should test this. I think the serial port HW is overrun (should that check too in the ISR) and gets out of sync. It doesn't have a fifo after all. Unfortunately I don't have a serial ISA card with fifo at hand to test this on my system. I'm going on the lookout to find one...

> Your SD card driver isn't interrupt driven, right?

Yes, it is **not** interrupt driven. The low level driver **doesn't** disable interrupts. It is not necessary as only atomic asm instructions are used and the SPI clock doesn't have to be stable.

ghaerr:
>  I think the serial port HW is overrun (should that check too in the ISR) and gets out of sync. It doesn't have a fifo after all.

Yes, you might check receiver overrun in the ISR, somehow (remember you can't call panic or much else since SS != DS). Perhaps just replace the received character with an '@'.

The reason I think you're not likely getting a receiver overrun is that there are two major portions of the timing we're concerned with: the time that interrupts are disabled, and everything else. The way we have the system built now, either the receiver interrupt routine takes too long at a certain baud rate, and loses the 2nd or third byte immediately, or, it has time to queue the character. Then we have the occasional timer interrupt, which on your system is lower priority, but otherwise would run for a small amount of time with interrupts disabled to increment jiffies. This would then show as loss of characters every 10ms. We'll need to divide out the baud rate to calculate time per character to see if this might have an effect.

Finally, there is the time remaining in the system with interrupts enabled. During that time, all bottom halves, kernel code, and application code executes. None of this (except very small critical sections) should affect the serial interrupt from queuing a character. 

Thus, I was thinking that the serial interrupt routine on your system is fast enough to service 115200 baud without receiver overrun, probably also while the timer interrupt is occuring, and thus everything runs well until the system just doesn't have enough time left for actual throughput by running enough software to empty the input buffer's 1024 bytes before it gets filled by the interrupt queue.

> The low level driver doesn't disable interrupts.

In that case, there should be no issues writing to the SD card while receiving high speed data. Have you seen dropouts while writing SD?

swausd:
> Have you seen dropouts while writing SD?

Currently I don't have any indicator signaling SD card writes.

ghaerr:
> Currently I don't have any indicator signaling SD card writes.

If, for testing only, you slightly rewrite `sercat` so that it only reads 256 bytes at a time, then writes, then calls `sync()`, that should flush the buffers, all while continuing to receive UART data at high speed.

ghaerr:
> 57600 baud: about 430 chars.
115200 baud: exactly 64 chars with 3 runs.

> the system flat can't keep up with the input and the chqueue routine finally just refuses to add the character to the input buffer. 

I'm continuing to think about this and must say I'm getting a bit confused. Under my theory, the system queues the received character if there's room, otherwise it does not. Thus, at super high baud rates and no receiver overrun, one should see, at 115200 baud, the first 64 characters received before the queue got full and input discarded (no other characters dropped). Is this what you're seeing? We should not be seeing characters past the first 64 in the sent data stream, at all.

By my calculations, at 115200 baud, it takes 10000 / 115200 = 0.0868ms (= 86.8 microseconds) per character. The entire 1000 byte buffer would be sent in 86.8ms, or approximately 9 clock ticks. I'm working on a way to record the time required to execute an \_irqit interrupt, but measuring the fast serial irq may be more difficult. `sercat` could possibly use the `get_ptime` precision timer to determine how long it is taking, but I'm not sure all the precise analysis would be worth much.

swausd:
Hi @ghaerr ,

tomorrow I will try to get some insight if/when uart overruns happen.

ghaerr:
> 38400 baud: about 10.000 chars, then I think writing to SD card finally happened.

If there is no interrupt-driven block device driver I/O compiled into your system, then the following section of code in ll_rw_blk.c may be able to run without disabling interrupts:
```
static struct request *get_request_wait(int n, kdev_t dev)
{   
    struct request *req;
    
    clr_irq();
    req = get_request(n, dev);
    set_irq();
    if (req) return req;
...
```
The same goes for `add_request`.

It is possible that the fast serial interrupt handler can't handle this block device request queue management code being executed with interrupts disabled. ELKS was enhanced to allow async I/O and is normally configured for it, even though CONFIG_ASYNCIO is disabled for the NECV25 port. I'm thinking these critical sections may also be able to be removed in the nonasync I/O case.

Not that any of this is super important, but it would be interesting to be able to build an ELKS system that allowed for super high speed serial input in certain configurations...

Mellvik:
@swausd @ghaerr - I'm following your exchanges and experiments with keen interest. It will take some time until I can experiment with this on TLVC. In the meanwhile - some observations:

- if pure raw input speed measurements is the goal, why not use /dev/null?
- apropos sd cards and interrupts: note that many ISA cards allowing CF and SD cards to look like IDE disks to the system, don't support interrupts at all. 
- given that the floppy driver is completely asynch, it would be interesting to see how that works as recipient of data in a situation like this.
- a dropped character counter would be useful at the chqueue level - obviously along with a way to read it, but hd will do for now.

On the subject of NICs and the new interrupt architecture, that will be interesting- in a positive way. The 'trad' driver architecture is very simple to adapt since the ISR doesn't do much except flagging data avail or data sent. 

All 3 elks ethernet drivers currently adhere to this structure. Since I'm still working on the wd driver adjustments, I'll take a look at it from this perspective while at it.

ghaerr:
@Mellvik 

> given that the floppy driver is completely asynch, it would be interesting to see how that works as recipient of data in a situation like this.

Yes - a quick look shows that interrupts aren't disabled for very long except in the floppy reset routine, where it's a disaster and a brute-force delay is used. Luckily, that likely doesn't happen very often. I see even in my DF driver that interrupts are off during DMA setup, which is only required if the DMA controller is used by another driver. I'll probably undo those. Nonetheless, I don't have easy access to real hardware, so it will be interesting to hear about your DF testing on TLVC when you get this enhancement ported, the DF driver looks better than I had imagined.

>  The 'trad' driver architecture is very simple to adapt since the ISR doesn't do much except flagging data avail or data sent.

I took a quick look earlier and thought that the NIC drivers spend all their time in dma_read/dma_write with interrupts disabled, but those routines are only called by the upper level read/write entry points, right? Why do those reads have to occur with interrupts disabled - just in case another RX/TX NIC interrupt occurs? It would seem to me the NIC card designers would have engineered away to safely read/write the data without having to disable interrupts, for fast servers, etc.

The NE2K driver has three separate calls to wake_up, one for RX, one for TX, and another I don't understand. Those are very expensive, but operate with interrupts enabled, but still at a lower priority pre-EOI IRQ number, so serial and timer interrupts are still effectively enabled (unless installed at IRQ 2, of course).

The other thing I noticed were quite a few printk's in the ISR routines. Keep in mind no bottom half handlers can run until the kernel is entirely finished with any top-half handler, so we might need to think about moving those printk's elsewhere. This also might not be a big deal for the NIC driver themselves, since the printk's slow them down just as much in either case. But the serial and timer bottom halves won't run until all top half handlers terminate, *and* any interrupted kernel code finishes as well.

Mellvik:
> I took a quick look earlier and thought that the NIC drivers spend all their time in dma_read/dma_write with interrupts disabled, but those routines are only called by the upper level read/write entry points, right? 

The 'remove_dma' stuff applies to the ne2k driver only. 

> Why do those reads have to occur with interrupts disabled - just in case another RX/TX NIC interrupt occurs? It would seem to me the NIC card designers would have engineered away to safely read/write the data without having to disable interrupts, for fast servers, etc.

With some tweaking, it's not required. There are - currently, in the old ne2k driver - a couple of situations that require protection: An ioctl may come in requesting the HW address. It's using the dma_read to get that. In practice it never happens (only ktcp uses it and only when it starts), and for a possible future rare occurrence, the protection can be put in the ioctl function instead even if that would make it longer. The other case (IIRC) is (or was, I don't remember if this is still the case) that the transmit can or could be in progress (initiated by a kerne write) when an interrupt happens which in itself is not a problem, but it is assumed that the 'remote dma' process needs to complete before the NIC does other stuff.  This protection - if required - is easy to do locally without blocking system interrupts.

> The NE2K driver has three separate calls to wake_up, one for RX, one for TX, and another I don't understand. Those are very expensive, but operate with interrupts enabled, but still at a lower priority pre-EOI IRQ number, so serial and timer interrupts are still effectively enabled (unless installed at IRQ 2, of course).

The wake_ups are unavoidable of course and given their cost they should be in the BH part of the new structure if at all possible. The way I understand things at this point, that would be OK. The third wake_up is in the overflow handler, which after cleaning up may decide there is more data available for read.

> The other thing I noticed were quite a few printk's in the ISR routines. Keep in mind no bottom half handlers can run until the kernel is entirely finished with any top-half handler, so we might need to think about moving those printk's elsewhere. This also might not be a big deal for the NIC driver themselves, since the printk's slow them down just as much in either case. But the serial and timer bottom halves won't run until all top half handlers terminate, and any interrupted kernel code finishes as well.

Good point, and like you I don't think this is a problem, at least not for the 'regular' (informational) messages. Debugging output may be different but not normally very time-critical.

swausd:
@ghaerr and @Mellvik ,

a short interim report of my results from this morning. Used the current ELKS repository, system setup to 115200 baud.

1. Blocked pump_port() in serial_bh() and character queuing in my sercon_fast_irq1_necv25() uart receiver ISR. So ELKS doesn't see any console input. ELKS thus waits at the login prompt for input that is not coming. Added serial echo of received chars and flag for uart errors (echos error flag with bit 7 set high) in sercon_fast_irq1_necv25(). Then send the file elkscmd/basic/eliza.bas (has line numbers so errors can be watched easy) via GTKTerm to my ELKS system.

I can see now, that the uart gets overrun-errors after about 800 characters. This can be repeated and so I conclude has nothing to do with ELKS interrupts.

Fun fact: Setting up the sending uart for 2 stop bits, gets the echo of the 9.5Kb file back to GTKTerm OK! So my conclusion is: sender and receiver get out of sync. I am using a "USB to TTL serial" thingy, because my PC doesn't have a serial port. I tried 3 different types of USB dongles -  one of them a sparkfun with an original FTDI FT232 - but that doesn't change anything. So I setup my PC for 2 stop bits from now on.

2. Re-enabled pump_port() and character queuing for the rest of this report. Serial echo is kept enabled.

4.  Following @Mellvik advise of using /dev/zero with `sercat /dev/tty1 >/dev/zero`. Serial echo still without errors, when sending eliza.bas.

5. Mounted SD card with minix filesystem and tried  `sercat /dev/tty1 >/mnt/x.txt`. The first 80 chars are OK, the rest is totally screwed up with most of the chars lost. The direct uart echo is still active and looking OK! So there are no uart overrun-errors! This can be repeated!

To be continued after a break and walk with the dog...

Mellvik:
@swausd -
great testing - I'll let @ghaerr comment on most of it.

> Following @Mellvik advise of using /dev/zero with sercat /dev/tty1 >/dev/zero. Serial echo still without errors, when sending eliza.bas.

While this probably works, be aware the `/dev/zero` is NOT `/dev/null`. The latter is the 'black hole' I was referring to, writing to it means just dumping the data. `/dev/zero` on the other hand, is used for input, and will deliver zeros forever. So, like a case we had just a week or so ago - we needed a 33MB file fast, content unimportant, size is the only thing that counts. So we read 33,000 1k blocks from /dev/zero and save in a file. `dd if=/dev/zero bs=2b of=bigzerofile` -- that's it.

What I'm getting from your test is - correct me if I'm wrong - you can dump a really big file into the serial port at 115200 bps - with echo - and no loss at all if you use `/dev/null`??

swausd:
Hi @Mellvik ,

thanks for pointing out my error with /dev/zero !

>  you can dump a really big file into the serial port at 115200 bps - with echo - and no loss at all if you use /dev/null?

Yes - if I setup the sending PC to 2 stop bits.

swausd:
Hi @ghaerr and  @Mellvik ,

OK, one mystery solved. @ghaerr posted above

> The default serial receive interrupt buffer is 1024 bytes.

But that is not true for tty console. INQ_SIZE and OUTQ_SIZE is only 80 chars long. This is not enough, as in one tick (10 ms) we get about 100 chars. I raised the buffers for a test to 1024. We then get a chqueue overrun after receiving  1196 chars. 

This I have to research a bit further...

ghaerr:
@swausd 

> that is not true for tty console. INQ_SIZE and OUTQ_SIZE is only 80 chars long.

Wow, that's a hell of a find! I had mistakenly been thinking that your serial console driver was somehow based off the standard serial driver's use of `tty_allocq(tty, RSINQ_SIZE, RSOUTQ_SIZE)`. You'll want to replace the call to ttystd_open in your console driver to incrementing the use count and allocating RS_INQSIZE bytes directly, which will fix this problem.

> my conclusion is: sender and receiver get out of sync. 

Another great finding! I had not considered the hardware connection itself as the issue. Do you suppose the clock frequencies are just a bit off? I've made a note ensure that fast serial senders should set two stop bits for reliability.

With the above info and fixes, it seems that your seemingly slow V25 system can reliably receive continuous serial input at 11200 baud. That's incredible! The new bottom half handler design seems to be definitely paying off! Thank you for all your great testing on this. It sounds to me like the serial receive handler is now working quite well, and should have no more issues after setting the input buffer to 1024.

> Mounted SD card with minix filesystem 
> The first 80 chars are OK, the rest is totally screwed up with most of the chars lost. 

At first glance, this sounds like a possible big problem in the SSD driver? Or do you think somewhere else? It seems your testing has shown with sercat echo that all characters are received correctly without overrun or dropout. It is unlikely that sercat itself isn't getting the data (but possible). That then leaves activity after sercat `write`. 

It might be worth testing the SSD driver using `cp` etc, or possibly faking up the buffer contents in `sercat` (with all 'A's for example) and seeing whether that data gets to the SD or not. This should be able to tell whether we've got a lower level buffer or block driver issue, or if sercat isn't getting data correctly.

ghaerr:
@Mellvik 

> currently, in the old ne2k driver - a couple of situations that require protection: An ioctl may come in requesting the HW address.

I don't understand how the "incoming" ioctl requires protection: the ioctl execution will never interrupt any other kernel code. It gets executed only after the application executing ioctl syscall, which can only occur when no kernel code is executing. Isn't the mac address already stored in netif_stat.mac_addr? 

If the mac address were dynamically required and not saved, then yes, with the previous design a hardware interrupt may interrupt the ioctl and attempt to perform a dma_read on top of a previous dma_read. But it still looks to me like the protection is not required even in the old design, since the mac addr is saved.

With the new bottom half design, even if the mac address was dynamically retrieved using dma_read, the bottom half handler will never interrupt other kernel (ioctl) code, so dma_read itself can likely run with interrupts enabled. Overall, it seems the bottom half handler design should allow big simplification to the NIC drivers.

> we needed a 33MB file fast, content unimportant, size is the only thing that counts. So we read 33,000 1k blocks from /dev/zero and save in a file. dd if=/dev/zero bs=2b of=bigzerofile -- that's it.

BTW, something you may not be aware of: while using /dev/zero on ELKS itself will always generate actually allocated blocks of zeros, using /dev/zero with `mfs`, for instance (or Linux), will just increase inode file size, but *not* actually allocate a block of zeros. This means that if expecting disk I/O to get a block of data, no I/O will occur, and instead a memset-zero buffer will be prepared internally and handed to the application. I ran into this when using `mfs` and found that /linux wasn't placed where I thought it would be, instead it remained very early in the filesystem image.

> While this probably works, be aware the /dev/zero is NOT /dev/null. 

Good point. But yes, /dev/zero is exactly the same as /dev/null on write.



> The third wake_up is in the overflow handler, 

My suggestion is that instead of calling `wake_up(&rxwait)` possibly twice in the routine, it only gets called once.

swausd:
Hi @ghaerr ,

> I had mistakenly been thinking that your serial console driver was somehow based off the standard serial driver's use of tty_allocq(tty, RSINQ_SIZE, RSOUTQ_SIZE). 

@cocus based the console driver on ntty and there INQ_SIZE and OUTQ_SIZE is set to 80.

> Do you suppose the clock frequencies are just a bit off?

To not have this problem, I used a 22.118400 Mhz clock. But I don't know what the USB dongles are doing. Don't they have to have a 25 MHz clock? I think these dongles are off frequency. When I have nothing else to to, may be I will measure the output of these dongles.

>> The first 80 chars are OK, the rest is totally screwed up with most of the chars lost.

> At first glance, this sounds like a possible big problem in the SSD driver?

No this 80 chars was exactly the serial input buffer size. With increased buffer size to 2048 I can write 2673 OK chars to SD card until it gets screwed up. But this is still less than the disk IO buffers, so I don't think that I have concurrent SD card access right now.

I think the problem that is left with serial input is, ELKS can fill up the chqueue, but can't get the queue emptied quick enough, because the CPU time is consumed with serial RX interrupts. So we get a ring buffer overflow even with a 2048 chars long serial RX buffer.

To test concurrent SD card writing, I will disable the disk IO buffers.

ghaerr:
@swausd 

> With increased buffer size to 2048 I can write 2673 OK chars to SD card until it gets screwed up. 

>  ELKS can fill up the chqueue, but can't get the queue emptied quick enough, because the CPU time is consumed with serial RX interrupts. So we get a ring buffer overflow even with a 2048 chars long serial RX buffer.

Yes, I see and agree. So what is happening is that we've got serial input working quite well, but even though sercat is writing say, 1024 or 2048 or 4096 bytes at a time (although likely far, far less), the data is only going to system buffers, and not actually written through the SSD driver. When the SSD driver is finally called to sync buffers, too much time is taken and the serial receive buffer overflows.

Thus we have a different actual system throughput problem, should we decide we really need to solve this.

One way would be to decrease the disk I/O buffers, but that's not a good idea, since the MINIX filesystem needs at least 3-4 for various technical reasons to prevent internal deadlock.

Another way would be to determine whether there's actually enough time to write a single 1024 byte buffer, and/or have a serial input buffer large enough to handle written just one block. This could be arranged by rewriting sercat: first, have it buffer received characters in its own 1024 byte buffer, and only issue a `write` at exactly 1024 bytes. Then execute a `sync` directly after the write. (Also execute `sync` on sercat start).

Since the sercat termios sets VMIN to 255 (max possible) and VTIME to 1, the TTY driver code will try to return 255 bytes each read in burst mode reception. It might be interesting to check if that is in fact happening. Otherwise, much extra cycles are spent writing data to system buffers when sercat could just be buffering them, saving time.

Frankly, the existing 4096 byte buffer in sercat is of little use, since it's very much larger than the serial input buffer. With using a single `write` of 1k bytes when possible followed by sync, we should be able to easily determine the smallest serial input buffer size required for this system. That value could then be allocated specifically in the console driver.

swausd:
Hi @ghaerr ,

I am afraid, I don't understand what is your goal or use case. I am quite happy with this BH-handling of interrupts. We can catch serial burst of 1000 chars at 115200 baud - that is much more than I want to cut & paste on a command line.

For file transfer over serial line I would always want a block orientated protocol with software handshake, automatic retry and secured, 100% delivery like kermit or xyz-modem. 
sercat is nice for testing serial action, but I wouldn't want it to use for file transfer.

Can you explain what you have in mind?

To optimise sercat I think, it would be best to buffer less and get received characters out fast.

ghaerr:
@swausd 

>  I don't understand what is your goal or use case. I am quite happy with this BH-handling of interrupts. 

Oh - I was just commenting on how we might improve on overall system throughput, if we want to. I am also quite happy with the results of our new BH interrupt design.

> Can you explain what you have in mind?

I was thinking it might be interesting to try to increase overall throughput with sercat, just for the fun of it.

> To optimise sercat I think, it would be best to buffer less and get received characters out fast.

Actually, I was thinking the opposite, since partial (e.g. less than 1024 byte) `writes` can't be written to SD, they must go to system buffers. By buffering to exactly 1024 bytes, writing 1024 bytes and then calling sync, it would seem that, while the data would initially be written to a system buffer, that system buffer would then be written to SD immediately thereafter, while the RX interrupt is busy filling up another 1024 (or larger, or smaller?) input queue. Then the cycle repeats. In this manner, less time is spent processing data downstream from sercat until a full block is received and then force-written to SD.

> sercat is nice for testing serial action, but I wouldn't want it to use for file transfer.

Agreed!

swausd:
Hi @ghaerr ,

> Actually, I was thinking the opposite, since partial (e.g. less than 1024 byte) writes can't be written to SD, they must go to system buffers. 

There has to be something left to do for tomorrow ;-)

Mellvik:
> > currently, in the old ne2k driver - a couple of situations that require protection: An ioctl may come in requesting the HW address.
> 
> I don't understand how the "incoming" ioctl requires protection: the ioctl execution will never interrupt any other kernel code. It gets executed only after the application executing ioctl syscall, which can only occur when no kernel code is executing. Isn't the mac address already stored in netif_stat.mac_addr?

First, the IOCTL is guaranteed to return the HW Mac address, the one stored in the card, which may be different from the one in use. Second, you're right, for the ELKS ne2k driver, this is superfluous, and a leftover from when the driver had a single read buffer. For the TLVC version it's different because all data moves to/from the NIC happen in the ISR.

> 
> With the new bottom half design, even if the mac address was dynamically retrieved using dma_read, the bottom half handler will never interrupt other kernel (ioctl) code, so dma_read itself can likely run with interrupts enabled. Overall, it seems the bottom half handler design should allow big simplification to the NIC drivers.

I'm not seeing the big simplifications yet but I certainly see the advantages - and looking forward to 'hit the road' so to speak. L ooking at these and other drivers with a fresh perspective - and like I mentioned before, more/fuller understanding of the total picture, is bound to deliver significant improvements. 

> 
> > we needed a 33MB file fast, content unimportant, size is the only thing that counts. So we read 33,000 1k blocks from /dev/zero and save in a file. dd if=/dev/zero bs=2b of=bigzerofile -- that's it.
> 
> BTW, something you may not be aware of: while using /dev/zero on ELKS itself will always generate actually allocated blocks of zeros, using /dev/zero with `mfs`, for instance (or Linux), will just increase inode file size, but _not_ actually allocate a block of zeros. This means that if expecting disk I/O to get a block of data, no I/O will occur, and instead a memset-zero buffer will be prepared internally and handed to the application. I ran into this when using `mfs` and found that /linux wasn't placed where I thought it would be, instead it remained very early in the filesystem image.

Hmmm, you almost (?) lost me on this one. My expectation from /dev/zero is that a stream of zeros is delivered, and I would be very surprised indeed if that generated physical read activity. Then again, this is about context, isn't it? You example expected/required something else.

> 
> > While this probably works, be aware the /dev/zero is NOT /dev/null.
> 
> Good point. But yes, /dev/zero is exactly the same as /dev/null on write.

On our systems (possibly even most), yes but not a guaranteed behaviour - education only :-)

> 
> > The third wake_up is in the overflow handler,
> 
> My suggestion is that instead of calling `wake_up(&rxwait)` possibly twice in the routine, it only gets called once.

The two instances cannot be called in the same loop. OTOH, it's a point. If a RX interrupt has already been serviced and a new one is generated in the meanwhile (creating a new loop in the ISR), wake_up() is going to be called more than once, which is useless - and negative now that its cost has came into focus! A very rare situation but still one I'll consider when redoing these drivers.


Thanks.

ghaerr:
> Hmmm, you almost (?) lost me on this one. My expectation from /dev/zero is that a stream of zeros is delivered

Sorry for my confusion, I wasn't really talking about how /dev/zero works. What I thought you might not realize (I had forgotten) has to do with using zeroes for allocating file blocks in a MINIX filesystem. When using Linux or `mfs` to create a file of all zeroes on a MINIX filesystem, neither system will actually allocate blocks in the target MINIX filesystem; the free space does not change. For that reason, I now use /dev/random, to ensure blocks are allocated. I was trying to speed up the process of moving /linux beyond 32MB for your 64MB boot bug and wanted to build the filesystem on the desktop, since copying 32M of data on ELKS is so slow, and it didn't work.

In fact, one can actually create a file of zeroes on a MINIX filesystem larger than the size of the MINIX filesystem itself (!), using the following, since zero blocks aren't actually allocated:
```
# Add 70MB /zeroes file onto existing 1440k MINIX floppy image
$ dd if=/dev/zero of=zeroes bs=2b count=70000
$ echo zeroes > filelist
$ mfs -fv image/fd1440.img adds filelist .
$ mfs -f image/fd1440img ls -l
drwxr-xr-x 11   1   0,  0     240 Dec 10 2025 20:05:52 .
drwxr-xr-x 11   1   0,  0     240 Dec 10 2025 20:05:52 ..
drwxr-xr-x  2   2   0,  0    1024 Dec 10 2025 20:05:52 dev
drwxr-xr-x  2   3   0,  0     224 Dec 10 2025 20:05:52 etc
drwxr-xr-x  2   4   0,  0    2320 Dec 10 2025 20:05:52 bin
-rw-r--r--  1  15   0,  0     666 Dec 10 2025 15:14:52 bootopts
-rwxr-xr-x  1  16   0,  0  116136 Dec 10 2025 20:05:52 linux
drwxr-xr-x  2  19   0,  0     384 Dec 10 2025 20:05:52 root
drwxr-xr-x  2  20   0,  0      32 Dec 10 2025 20:05:52 home
-rw-r--r--  1  21   0,  0   10244 Dec 04 2025 13:50:45 .DS_Store
drwxr-xr-x  3 165   0,  0      48 Dec 10 2025 20:05:53 var
drwxr-xr-x  2 168   0,  0      32 Dec 10 2025 20:05:53 mnt
drwxr-xr-x  9 191   0,  0     304 Dec 10 2025 20:05:53 lib
drwxrwxrwx  2 366   0,  0      32 Dec 10 2025 20:05:53 tmp
-rw-r--r--  1 429   0,  0 71680000 Dec 11 2025 11:10:15 zeroes
```
(BTW, ignore the macOS \.DS_Store file on the floppy above - I found that problem the other day when I realized that somehow my disk images were seemingly getting smaller, not realizing they were being created by macOS Finder. Fix coming to `mfs` for that).

Mellvik:
> When using Linux or mfs to create a file of all zeroes on a MINIX filesystem, neither system will actually allocate blocks in the target MINIX filesystem; the free space does not change. 

Thanks for clarifying that @ghaerr , I realized later that I had misunderstood. The `mfs`  reference should have been more than enough to send my head to the host rather than staying on the 'target'. 

Fixing `mfs`- thanks, it's a group of tools I (and I guess most of us) take for granted without considering the fact that they are key parts of the system.

Mellvik:
> Actually, I was thinking the opposite, since partial (e.g. less than 1024 byte) writes can't be written to SD, they must go to system buffers. By buffering to exactly 1024 bytes, writing 1024 bytes and then calling sync, it would seem that, while the data would initially be written to a system buffer, that system buffer would then be written to SD immediately thereafter, while the RX interrupt is busy filling up another 1024 (or larger, or smaller?) input queue. Then the cycle repeats. In this manner, less time is spent processing data downstream from sercat until a full block is received and then force-written to SD.

It seems to me - again watching from the sidelines, that unless the storage driver blocks interrupts a lot, this should work really well up to the point where the system capacity limit is reached. So - as far as purpose is concerned - it seems you're heading towards a tool or a script that can determine the capacity limit (a metric) of the system. Primarily for serial input, but that also becomes a metric for overall system response and certainly a benchmark value for measuring system level performance improvements - both very valuable.

Eventually I'll get to test the new interrupt architecture on a system with 4 serial lines, all of them with FIFOs unfortunately, but it might still be an interesting experiment. A different system has 2 non-fifo serial lines and is quite slow (V20/8MHz), another fun setting to push. Later.

swausd:
@ghaerr and @Mellvik 

OK, i have finally done, what I should have done much earlier: **checking for system serial receive buffer overflow (chqueue)**.

Fact is, my system now catches all serial chars - even at 115200 baud - thanks to the BH IRQ handling, but my system is not fast enough to pass the received characters  along to waiting processes, even if the buffer size is very large, like 4096 bytes.

swausd:
At 19200 baud I successfully copied a 750Kb text file multiple times with `sercat /dev/tty1 >/mnt/x1.txt` from my PC to the ELKS systems SD card without errors. The sending PC was configured for **1 stop bit**. Any higher baud rate produces a `chqueue` overflow eventually.

swausd:
Above I reported, that I get UART receiver hardware overruns at 115200 baud if only using 1 stop bits.

>> Do you suppose the clock frequencies are just a bit off?

> To not have this problem, I used a 22.118400 Mhz clock. But I don't know what the USB dongles are doing. Don't they have to have a 25 MHz clock? I think these dongles are off frequency. When I have nothing else to to, may be I will measure the output of these dongles.

I looked at the bit length with my oscilloscope and they looked the same, and in spec,  on the PC and NEC V25 side. But I only used a rather static setup for measuring. I can't verify if this is always the case during a longer transaction, but don't think that this is a problem.

When comparing the UART description in NEC V25 manual with other, dedicated UART chips, there is the impression, that NEC used a rather simple implementation. May be, it not only looks simple but in fact is simple. 

So with higher baud rates use 2 stop bits on the sender side with NEC V25 internal UARTs on the receiving end.

Mellvik:
@swausd , this is a lot of great testing - and BTW I find it rather convenient to have access to someone with not only an oscilloscope but the ability to use it .

I find the 'seemingly out of sync' problem (or rather, observation) very interesting. In the past, and unless my memory is totally off, I have used USB serial converters at speeds much higher than 115200 connected to the likes of Raspberry PI 2 and higher. Working reliably. Which would indicate - as @swausd is now suggesting - that the clock frequencies are indeed correct.

The reason the addition of an extra stopbit makes a difference may be as simple as the extra delay it adds, giving the venerable UART time to handle speeds that weren't considered (or conceivable) when they were made. I found this word of wisdom from the net enlightening in that regard:

```
Extra stop bits can be a useful way to add a little extra receive processing time, especially at high baud rates and/or using soft UART, where time is required to process the received byte.

Where speed is tight, and your UART only offers division ratios in powers of 2, adding an extra stopbit can be an option to give a less drastic reduction in speed than the next lowest baudrate.

I believe this is may one reason the DMX512 standard specifies 2 stopbits.

Another situation where they can be useful is if you have devices forwarding a data stream without any buffering or packetisation - small differences in clock rates between nodes and finite sampling granularity can cause errors to occur as data is received and retransmitted by a number of nodes in a chain, but if the data is sent with 2 stopbits and the receivers are set to one stopbit, it adds enough margin to accommodate these errors and leave at least one valid stopbit period for nodes far down the chain to receive reliably.

I have also enountered a situation where a very long cable run caused some asymmetry in the rise and fall times, resulting in inadequate stopbit length - sending 2 stopbits and having the receiver only require one fixed this.
```
https://electronics.stackexchange.com/questions/29945/one-or-two-uart-stop-bits
 
Also keep in mind that an extra bit in each 'frame' reduces the effective 'byte-rate' significantly.

When all this has settled down and a new level of serial performance has been established, adding hardware flow control would be fun ... In my case, most serial connections are through a terminal server (telnet in, serial out) with full RS232 support.

swausd:
@Mellvik 

> The reason the addition of an extra stopbit makes a difference may be as simple as the extra delay

I understand that with dedicated UART chips, the clock usually should be 16x the baud rate to do "oversampling" for edge detection. With the NEC V25 internal UART in my situation this may not be the case. 

By the way, I overclock my CPU heavily. NEC never advertised this CPU for such high clock rates or UART speeds, so I am happy its working stable somehow.

> Also keep in mind that an extra bit in each 'frame' reduces the effective 'byte-rate' significantly.

Yes that's right. The clock rate is reduced by about 10% so it is not 11520 chars/s (10 bits per char) but "only" max. 10472 chars/s (11 bits per char).

ghaerr:
Very interesting information on 1 vs 2 stop bits.

> Any higher baud rate produces a chqueue overflow eventually.

I found the `sercat -v` option might be useful in determining how fast a system might be wastefully processing received data before having enough to write a system buffer. When -v is specified, sercat will display the number of bytes received each `read`. It might be interesting to run this at 115200 baud with 1 and stop bits, to see the difference.

> At 19200 baud I successfully copied a 750Kb text file multiple times 

It sounds like this is likely the limit of your system to continuously process serial data, without a packet oriented protocol.

PR #2507, opened 2025-12-08
[ktcp] Don't delete retransmit buffers when retrans memory full

ghaerr:
This PR adds @Mellvik's TLVC enhancement where the retransmit buffer was deleted when retrans memory was getting full on an active connection, causing a hang if any retransmission was needed.

As described in https://github.com/Mellvik/TLVC/pull/211, when the retransmit buffer memory is nearing capacity, an -ERESTARTSYS is already returned to the sending application to retry the write 100ms later.

Other changes in this PR updates ktcp with descriptive comments from TLVC ktcp. Thanks @Mellvik!

This concludes the changes made in ELKS and TLVC for https://github.com/ghaerr/elks/issues/2487, and our ktcp should now be in sync with TLVC.

@swausd, when you find time, please test this latest version on your hardware, as the final fix for #2487.

### DIFF
--- a/elkscmd/ktcp/deveth.c
+++ b/elkscmd/ktcp/deveth.c
@@ -146,12 +146,13 @@ void eth_sendpacket(unsigned char *packet, int len, eth_addr_t eth_addr)
 	ipll->ll_type_len = 0x08; //FIXME what is 0x0800
 
 #if FORCE_MISSING_PACKET        /* For debugging: Create errors */
-    static int failcnt = 0;
-    if (failcnt++ > FORCE_MISSING_PACKET)
-        failcnt = 0;
-    else
+	static int failcnt = 0;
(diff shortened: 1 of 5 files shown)


swausd:
Hi @ghaerr and @Mellvik ,

> please test this latest version on your hardware, as the final fix for https://github.com/ghaerr/elks/issues/2487.

I tested with my old switch, which loses packets from time to time, with telnet and my counter app with very short packets. With the current ktcp I can reproduce the error in a real life situation (FORCE_MISSING_PACKET = 0). I get 99-100 TCP DUP ACKs, then a timeout of about 4-6 seconds until ELKS retransmits the lost packet. Then ktcp continues. 

After 50.000 packets there where 15 retransmits.

So I think the problem is fixed and the issue https://github.com/ghaerr/elks/issues/2487#issue-3672608378 can be closed.

Issue #2508, opened 2025-12-09
How to set serial tty default baud rate


swausd:
Hi @ghaerr ,

I want to set the default baud rate of my serial console to 115200 baud. My monitor baud rate is also 115200 and I want a smooth change to ELKS without losing any ELKS boot messages.

Did this in ntty.c with
```

/* default termios, set at init time, not reset at open*/
struct termios def_vals = {
    BRKINT|ICRNL,                                       /* c_iflag*/
    OPOST|ONLCR,                                        /* c_oflag*/
#ifdef CONFIG_ARCH_NECV25
    (tcflag_t) (B115200 | CS8),                         /* c_cflag*/
#else
    (tcflag_t) (B9600 | CS8),                           /* c_cflag*/
#endif
```
But I think this is not the best way to do it. Any suggestions how to do it better?

ghaerr:
Yes, use #2510 and add a CONFIG_DEF_BAUD B115200 in the NECV25 section of config.h.

PR #2510, opened 2025-12-10
[kernel] Set default baud rate using CONFIG_DEF_BAUD

ghaerr:
Answers question in #2508.

Add CONFIG_DEF_BAUD in config.h to set default baud rate.


swausd:
@Hi @ghaerr ,

Just posted my question "after work" and waking up this morning, I see you already solved my "problem". 

Thank you very much!

PR #2520, opened 2025-12-12
[kernel] Allow stopped jobs to be killed, update sash kill/SIGTERM handling

ghaerr:
Previously, all of a stopped process' received signals were cleared when restarted via SIGCONT, which prevented sending SIGKILL via `kill -9 pid`. Now, only the SIGINT and SIGQUIT signals are cleared (as they may be sent to the process group by the TTY device), along with SIGSTOP and SIGTSTP. This allows all other signals to accumulate while the process is stopped, and then acted on immediately when continued.

### DIFF
--- a/elks/arch/i86/kernel/signal.c
+++ b/elks/arch/i86/kernel/signal.c
@@ -27,6 +27,7 @@ void do_signal(void)
     unsigned int signr;
     sigset_t mask;
 
+top:
     signr = 1;
     mask = (sigset_t)1;
     while (current->signal) {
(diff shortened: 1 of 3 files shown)


ghaerr:
Also allows specifying negative pid for process group in `sash` builtin `kill` command, like /bin/kill.

`sash` now ignores SIGTERM (like `ash`) allowing a SIGTERM to be sent to a either shell's process group using 
`kill -15 -pgrp` and not also killing the shell.

PR #2522, opened 2025-12-12
[NECV25] changed a little bit more...

swausd:
Missed deleting line 333 (now 330)  from @ghaerr comment.


ghaerr:
Ha! I was wondering about that! I don't see much point in using else after returns, so glad to see this :)

PR #2524, opened 2025-12-12
[elkscmd/inet] netcat-style network utility

sepen:
I have added a new lightweight **netcat** implementation for **ELKS**.
The goal was to provide a simple TCP client/server tool, similar in spirit to the classic Unix netcat, but written from scratch to fit the limitations of **ELKS**: small memory, no fork/exec overhead, and minimal libc use.

It supports two basic modes:
- **Client mode**: `netcat <host> <port>`: It resolves the hostname using **ELKS**’s `in_gethostbyname()`, connects, and forwards data between stdin/stdout and the TCP socket.
- **Listen mode**: `netcat -l <port>`: It binds and accepts a single incoming TCP connection.

Instead of adding command-line options, I chose to keep things simple and use environment variables for optional behavior:
- `NETCAT_DEBUG=1`: Enables debug messages showing resolve/connect/select steps.
- `NETCAT_MAXBYTES=<n>`: Stops after sending/receiving a total of n bytes, useful for tests or scripted interactions.

The program is intentionally minimal, but still very usable for testing services, making simple TCP connections, debugging servers, or piping data through the network. 

If desired, we can extend it later (timeouts, NETCAT_EXEC, etc.), but for now I think this provides a small and reliable **netcat** suited to **ELKS**.

**Features not included** (by design)

To keep the program small and reliable for ELKS:

- No UDP mode
- No exec feature (-e, -c)
- No port scanning
- No telnet option negotiation
- No IPv6
- No proxy support (SOCKS / HTTP-CONNECT / TLS)
- No timing/interval options (-w, -i)
- No verbose/hexdump modes (only simple NETCAT_DEBUG=1)
- No multiple / parallel connections

**Screenshot**

Screenshot 2025-12-12 at 00 50 30


ghaerr:
Wow @sepen this is fantastic, very well done! Thank you, I will commit now so I can start playing with it :)

PR #2532, opened 2025-12-16
[NECV25] Support for 2nd console on "Serial 0" for NEC V25 port

swausd:
Usually, the internal MCU UART "Serial 0" is configured as a hardware SPI to connect an SD card. If slower software SPI is configured, the now free "Serial 0" is set up as a second serial console.

This is the preparation to work on some xyz-modem app to get files down-/up-loadable via serial port. With only one serial port this is nearly impossible to test.

### DIFF
--- a/elks/arch/i86/drivers/char/console-serial-necv25.c
+++ b/elks/arch/i86/drivers/char/console-serial-necv25.c
@@ -4,9 +4,11 @@
  * This file contains code used for the embedded NEC V25 family only
  * Based on Santiago Hormazabal console-serial-8018x.c
  *
- * Uses the MCU internal serial 1 port as tty console.
- * Serial 0 is used as SPI for SD card interface and cannot be used
- * for serial communication.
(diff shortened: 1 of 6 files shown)


ghaerr:
Looks good @swausd, thank you!

> This is the preparation to work on some xyz-modem app to get files down-/up-loadable via serial port. 

NIce. BTW @tyama501 is working on an ELKS kermit, I believe.

> With only one serial port this is nearly impossible to test.

Actually, I thought some versions of XMODEM allowed it run and receive on the tty it was started from, using a command line option?

swausd:
> BTW @tyama501 is working on an ELKS kermit, I believe.

I have seen that, but I am only fan of kermit the frog ;-)

>>  With only one serial port this is nearly impossible to test.

> Actually, I thought some versions of XMODEM allowed it run and receive on the tty it was started from, using a command line option?

That is true and I am aiming for that. But testing with only one serial port is a PITA.

PR #2560, opened 2026-01-01
[kernel] Move syscall vector to .text, reduce number of unused entries

ghaerr:
The syscall vector is moved to .text from .data, operating the same way ROM kernels always have. This frees up 414 bytes in the kernel .data section.

Additionally, the networking and fmemalloc/fmemfree system calls from 198-207 are renumbered 75-83, freeing up 492 bytes .data when CONFIG_TRACE tracing is compiled in, and reduces the size of the syscall vector by 246 bytes.

NOTE: Renumbering the high 10 syscalls requires recompilation of the ELKS, OWC and C86 C libraries, as well as applications using networking or fmemalloc/fmemfree. This is all done automatically when building the system via `make clean; make`, followed by `make owc` or `make c86` when configured for OWC or C86.


ghaerr:
This change will require the following apps to be relinked in order to run: Nano-X, Doom, 8086-toolchain, C86/OWC libraries, ngIRCd, elkirc, and possibly Lua.

Even though this introduces incompatibility with previously linked networking applications and programs using far memory, I think it worth getting this done prior to release v0.9, to save space in the longer term and reduce the large current NUL gap in the system call list. 

My hope is to mitigate most these factors by introducing a mechanism to build known external programs along with the v0.9 release and including the working and tested binaries in a full ELKS distribution, for the first time.

tyama501:
My Japanese text viewers also using fmemalloc.
Although may be I am only one using it but I will release another build after this merge.

tyama501:
Hello @ghaerr 

I'm not sure this is related to this PR,
but I got undefined symbols in the elkscmd/gui
after doing make c86.

image

I have rebuild 8086 toolchain before this.
make and make owc worked well.
Do you have any idea why this happen?

Thank you.

ghaerr:
Hello @tyama501,

I am guessing something might still be incorrect about building C86 support from scratch, or that the ELKS build is trying to build C86 elkscmd/gui too early. Can you try running
```
$ ./buildc86.sh
```
and see if that fixes it? It appears that libc/libc86.a may be out of date. If that is the case, perhaps running
```
$ cd libc
$ make -f c86.mk clean
$ make -f c86.mk
```
will help.

Thank you!

tyama501:
I'm not sure why but ./buildc86.sh worked!

Although paint is not supported for PC-98, thank you :)

ghaerr:
> I'm not sure why but ./buildc86.sh worked!

Good to know, something is wrong with `make c86`, I've made a note of it.

> Although paint is not supported for PC-98

Should you ever want to look further into it, the graphics operations are contained in elkscmd/gui/vgalib.h and graphics.c. However, there are certain aspects of paint that are dependent on VGA hardware. We can discuss that should you become interested in porting it. Since your work has been in successfully porting Nano-X to PC-98, this probably isn't necessary though.

ghaerr:
@tyama501,

> something is wrong with make c86, I've made a note of it.

The C86 build of paint.c86 in elkscmd/gui is only supposed to occur when the following is set in .config. Can you check your .config to see what is the setting?
```
CONFIG_APPS_C86=y
```
The same for building OpenWatcom applications, the following should be set (or unset if not desired):
```
CONFIG_APPS_WATCOM=y
```

tyama501:
Hello @ghaerr 

I have rebuilt c86 and copied to the latest elks,
but something is still wrong.

If I do make,
meminfo is called...
and get errors.

image

Is this related to the system call difference?
Is it working on your environment?
(If you are aware this it is ok)

Thank you.

ghaerr:
No, I don’t think this is related to syscall changes. The C86 compiler libraries execute meminfo when no memory is available, to show the amount of free memory.

Have you changed the amount of memory available for PC98 recently? This sounds like there is not enough main memory to complete the compilation. I did test this, but will test again with syscall change.

On Jan 11, 2026, at 10:03 PM, Takahiro Yamada ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/61556504?s=20&v=4]tyama501 left a comment (ghaerr/elks#2560)<https://github.com/ghaerr/elks/pull/2560#issuecomment-3736893458>

Hello @ghaerr<https://github.com/ghaerr>

I have rebuilt c86 and copied to the latest elks,
but something is still wrong.

If I do make,
meminfo is called...
and get errors.

image.png (view on web)<https://github.com/user-attachments/assets/ce1fd559-644c-47fb-b6f4-a98454680ef8>

Is this related to the system call difference?
Is it working on your environment?
(If you are aware this it is ok)

Thank you.

tyama501:
No, I haven't change any conventional memory amount.

This occurs both on the emulator and the real PC.

Thank you.

ghaerr:
Hello @tyama501,

I have tested the C86 build on QEMU and it seems to be working fine, as long as there is disk space available. Are you testing on emulator with HDD? I am not sure if I can get DosBoxX to accept an HDD for me to test PC98.

Your previous screen shot is showing 338K free, which seems to say this isn't about free memory.

Have you rebuilt the entire C86 chain (again)? Please try the following to see if it fixes things:
```
$ ./buildc86.sh
$ ./copyc86.sh
$ make
```
If it still errors, is this happening immediately upon `make` or after completion of several compiles? Additional screenshots would help showing the error. This sounds like it could be to do with syscalls changing but not updating C86, as the C86 compiler uses the fmemalloc function, whose syscall number has changed.

Thank you!

tyama501:
Hello @ghaerr 

I have done that several times, 
but I will look again when I have another time.

One thing I remember is when running it with elks user account not root on the elks,
it says creating /dev/console fails.
I don't understand what that means, but 
that also strange.

Thank you.

ghaerr:
Can you give me more information on exactly when the build fails? Failing immediately after the first make is very different that failing after compiling a few source files. You can screenshot while it's displaying if easier, before meminfo display.

I think /dev/console create failure is bad permissions on /dev/console for non-root users. The \_\_dprintf debug printf code tries to open /dev/console to give certain error messages, and a quick look using ls -l shows /dev/console as mode 600. I will try to fix that, or you can chmod it and try again to confirm.

ghaerr:
@tyama501, another option to try for C86 error is just building a single program: `make chess` rather than `make`. Try that and see what happens, thanks!

tyama501:
Hello @ghaerr ,

It seem that I needed to do
make owclean
and
make c86clean
before doing buildc86.

Now, it worked.
Thank you.

ghaerr:
@tyama501, thank you for letting me know! This means the binaries or libraries were not rebuilt correctly following the syscall vector changes.

I will modify buildc86.sh to clean the OWC libraries also, I think this is the reason for your problem.

Thank you!

PR #2564, opened 2026-01-07
[ekermit] Reduce SCANSIZ since count is 16bits int

tyama501:
Hello @ghaerr 

There was a warning
elksio.c: In function ‘fileinfo’:
elksio.c:513:19: warning: comparison is always true due to limited range of data type [-Wtype-limits]
      while (count < SCANSIZ && !isbinary) { /* Scan this much */

because count is 16bits int in our system.

I have reduced SCANSIZ to eliminate this warning.

Thank you.

### DIFF
--- a/elkscmd/ekermit/elksio.c
+++ b/elkscmd/ekermit/elksio.c
@@ -464,7 +464,7 @@ openfile(struct k_data * k, UCHAR * s, int mode) {
 */
 #ifdef F_SCAN
 #define SCANBUF 1024
-#define SCANSIZ 49152
+#define SCANSIZ 32767
 #endif /* F_SCAN */
 
 ULONG


tyama501:
I have confirmed CAT still can be detected as binary file with this SCAN.

ghaerr:
Although I quickly committed this without looking at it much, I now realize that your proposed fix is in error. The problem is that when SCANSIZ is 32767 and `count` a signed `int`, when count increments past 32767, it's value turns into -32768, which is still smaller than 32767 (SCANSIZ). Thus, this introduces undefined behavior.

A proper fix would entail either turning off F_SCAN (does ekermit actually require knowing whether file is actually binary or not? It should always transfer files as though they are binary, IMO), *or* setting SCANSIZ to a value that won't overflow, so that when incremented, its value is still a positive integer.

Note that count is always incremented by max SCANBUF, so SCANSIZ can be no larger than (32767 - SCANBUF - 1). 

In general, on ELKS, "scanning" a file to determine whether its binary is quite slow, especially on floppy. That is why I propose possibly forcing all transfers to be binary, and then turning off F_SCAN (and fixing SCANSIZ also).

tyama501:
Hello @ghaerr 

Well I think that is intended.
If control characters other than LF etc is not detected untill 37676 and it reaches the 37677 then it is text file.
If it is finded or EOF comes then it is binary file and breaks.

tyama501:
Oh, maybe you mean count += n makes problem.

tyama501:
I will check why binary or text need to be set. It is also selectable in the option.

ghaerr:
> maybe you mean count += n makes problem.

Yes. When count += n is executed, count can become < 0 and continue in loop when SCANSIZ is 32767. This is why SCANSIZ can't be larger than 32767-SCANBUF-1. (Actually that might not be exact number required, depending on read() return value and exact file size. Perhaps better to set SCANSIZ to very small number like 1024, since then scan would be very quick since very few binary files would appear like text files in first 1K bytes.

> I will check why binary or text need to be set. It is also selectable in the option.

I suspect the reason is that text files can get automatic LF <-> CRLF (UNIX <-> DOS) conversion, which was useful in the old days. For ELKS, scanning 32k files on disk is very slow and there is usually no need for file type conversion. I propose always forcing binary mode transfers and \#undef F_SCAN so that file transfer is quicker.

PR #2566, opened 2026-01-09
[net] Add Interrupt Process and Abort Output telnet protocol handling

ghaerr:
Enhances telnet and telnetd to properly handle telnet protocol "Interrupt Process" (^C) and "Abort Output" (^O). Previously, these characters were sent to the remote system. Now, ^C sends telnet IAC IP, which is then converted to ^C by telnetd. ^O sends IAC AO, which currently does nothing in telnetd.

Telnet continues to discard output while ^C or ^O processing is ongoing. Typing any character will reenable output, as will also after a period of network input timeout.

The telnet protocol and how other systems work, particularly GNU inetutils, is discussed in detail in https://github.com/Mellvik/TLVC/pull/215.

@Mellvik, these are my minimal changes based on our long discussion. Everything seems to be working well testing using QEMU and ELKS localhost, and macOS. However, strangely enough, my debug shows that macOS never sends a DM (DataMark) response. Thus, it seems a network timeout is required in order to make this work. I have debug code writing "TO" for timeout and "DM" when DataMark received, and DM is never displayed.

### DIFF
--- a/elkscmd/inet/telnet.c
+++ b/elkscmd/inet/telnet.c
@@ -32,7 +32,7 @@
 #define IAC         255
 #define IAC_SE      240
 #define IAC_NOP     241
-#define IAC_DataMark 242
+#define IAC_DM      242
 #define IAC_BRK     243
 #define IAC_IP      244
 #define IAC_AO      245
@@ -89,6 +89,30 @@ finish()
     exit(0);
 }
 
+void sendcmd(int cmd)
+{
+    unsigned char reply[3];
+
+    reply[0] = IAC;
+    reply[1] = cmd;
+    write(tcp_fd, reply, 2);
+}
+
+int iscmdchar(int c)
+{
+    if (c == CTRL('C')) {
+        sendcmd(IAC_IP);
+        discard = 1;
+        return 1;
+    }
+    else if (c == CTRL('O')) {
+        sendcmd(IAC_AO);
+        discard = 1;
+        return 1;
+    }
+    return 0;
+}
+
 void
 read_keyboard(void)
 {
@@ -100,15 +124,10 @@ read_keyboard(void)
         fprintf(stderr, "\nSession terminated\n");
         finish();
     }
-#if DISABLED
-    if (buffer[0] == CTRL('C'))
-        discard = 1;
-    else
-#endif
-    if (buffer[0] == CTRL('O')) {
-        discard ^= 1;
+
+    if (iscmdchar(buffer[0] & 255))
         return;
-    }
+
     count = write(tcp_fd, buffer, count);
     if (count < 0) {
         perror("Connection closed");
@@ -245,8 +264,10 @@ main(int argc, char **argv)
 
         n = select(tcp_fd + 1, &fdset, NULL, NULL, &tv);
         if (n == 0) {
-            if (discard)
+            if (discard) {
                 write(tcp_fd, "\r", 1);
+                write(1, "TO", 2);
+            }
             discard = 0;
             continue;
         }
@@ -457,8 +478,12 @@ process_opt(char *bp, int count)
     switch (command) {
     case IAC_NOP:
         break;
-    case IAC_DataMark:
-        debug("got DataMark\n");
+    case IAC_DM:
+        debug("got DM\n");
+        if (discard) {
+            write(1, "DM", 2);
+            discard = 0;
+        }
         break;
     case IAC_BRK:
         debug("got BRK\n");
--- a/elkscmd/inet/telnet.c
+++ b/elkscmd/inet/telnet.c
@@ -105,7 +105,7 @@ int iscmdchar(int c)
         discard = 1;
         return 1;
     }
-    else if (c == CTRL('O')) {
+    if (c == CTRL('O')) {
         sendcmd(IAC_AO);
         discard = 1;
         return 1;
--- a/elkscmd/inet/telnet.c
+++ b/elkscmd/inet/telnet.c
@@ -127,6 +127,7 @@ read_keyboard(void)
 
     if (iscmdchar(buffer[0] & 255))
         return;
+    discard = 0;
 
     count = write(tcp_fd, buffer, count);
     if (count < 0) {
@@ -267,8 +268,8 @@ main(int argc, char **argv)
             if (discard) {
                 write(tcp_fd, "\r", 1);
                 write(1, "TO", 2);
+                discard = 0;
             }
-            discard = 0;
             continue;
         }
         if (n < 0) {
--- a/elkscmd/inet/telnet.c
+++ b/elkscmd/inet/telnet.c
@@ -263,11 +263,11 @@ main(int argc, char **argv)
         tv.tv_sec = 0;
         tv.tv_usec = 100000L;   /* 100ms */
 
-        n = select(tcp_fd + 1, &fdset, NULL, NULL, &tv);
+        n = select(tcp_fd + 1, &fdset, NULL, NULL, discard? &tv: NULL);
         if (n == 0) {
             if (discard) {
+                debug("TO");
                 write(tcp_fd, "\r", 1);
-                write(1, "TO", 2);
                 discard = 0;
             }
             continue;
@@ -481,10 +481,7 @@ process_opt(char *bp, int count)
         break;
     case IAC_DM:
         debug("got DM\n");
-        if (discard) {
-            write(1, "DM", 2);
-            discard = 0;
-        }
+        discard = 0;
         break;
     case IAC_BRK:
         debug("got BRK\n");
--- a/elkscmd/inet/telnet.c
+++ b/elkscmd/inet/telnet.c
@@ -25,7 +25,7 @@
 #define BUFSIZE     1500
 
 #define debug(...)
-//#define debug     printf
+//#define debug     __dprintf
 //#define RAWTELNET             /* test mode for raw telnet without IAC */
 
 /* telnet protocol */
@@ -89,16 +89,16 @@ finish()
     exit(0);
 }
 
-void sendcmd(int cmd)
+static void sendcmd(int cmd)
 {
-    unsigned char reply[3];
+    unsigned char reply[2];
 
     reply[0] = IAC;
     reply[1] = cmd;
     write(tcp_fd, reply, 2);
 }
 
-int iscmdchar(int c)
+static int iscmdchar(int c)
 {
     if (c == CTRL('C')) {
         sendcmd(IAC_IP);
@@ -266,7 +266,7 @@ main(int argc, char **argv)
         n = select(tcp_fd + 1, &fdset, NULL, NULL, discard? &tv: NULL);
         if (n == 0) {
             if (discard) {
-                debug("TO");
+                debug("TO\n");
                 write(tcp_fd, "\r", 1);
                 discard = 0;
             }
--- a/elkscmd/inet/telnet.c
+++ b/elkscmd/inet/telnet.c
@@ -261,7 +261,7 @@ main(int argc, char **argv)
         FD_SET(0, &fdset);
         FD_SET(tcp_fd, &fdset);
         tv.tv_sec = 0;
-        tv.tv_usec = 100000L;   /* 100ms */
+        tv.tv_usec = 500000L;   /* 500ms */
 
         n = select(tcp_fd + 1, &fdset, NULL, NULL, discard? &tv: NULL);
         if (n == 0) {
--- a/elkscmd/inet/telopt.c
+++ b/elkscmd/inet/telopt.c
@@ -146,6 +146,8 @@ tel_in(int fdout, int telout, char *buffer, int len)
             case WONT:
             case DO:
             case DONT:
+            case IP:
+            case AO:
                 InState = IN_IAC2;
                 ThisOpt = c;
                 break;
@@ -156,8 +158,6 @@ tel_in(int fdout, int telout, char *buffer, int len)
             case SE:
             case NOP:
             case BREAK:
-            case IP:
-            case AO:
             case AYT:
             case EC:
             case EL:
@@ -186,6 +186,12 @@ tel_in(int fdout, int telout, char *buffer, int len)
                 break;
             case DONT:
                 dodont(c);
+                break;
+            case IP:
+                write(fdout, "\003", 1);
+                break;
+            case AO:
+                /* no action for Abort Output */
                 break;
             }
             break;
--- a/elkscmd/inet/telopt.c
+++ b/elkscmd/inet/telopt.c
@@ -146,13 +146,19 @@ tel_in(int fdout, int telout, char *buffer, int len)
             case WONT:
             case DO:
             case DONT:
-            case IP:
-            case AO:
                 InState = IN_IAC2;
                 ThisOpt = c;
                 break;
             case SB:
                 InState = IN_SB;
+                break;
+            case IP:
+                InState = IN_DATA;
+                write(fdout, "\003", 1);
+                break;
+            case AO:
+                InState = IN_DATA;
+                /* no action for Abort Output */
                 break;
             case EOR:
             case SE:
@@ -186,12 +192,6 @@ tel_in(int fdout, int telout, char *buffer, int len)
                 break;
             case DONT:
                 dodont(c);
-                break;
-            case IP:
-                write(fdout, "\003", 1);
-                break;
-            case AO:
-                /* no action for Abort Output */
                 break;
             }
             break;
--- a/elkscmd/inet/telopt.c
+++ b/elkscmd/inet/telopt.c
@@ -90,7 +90,7 @@ tel_opt(int fdout, int what, int option)
         break;
     }
     if (len > 0)
-        (void)write(fdout, buf, len);
+        write(fdout, buf, len);
 }
 
 static void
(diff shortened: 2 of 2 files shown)


Mellvik:
Nice @ghaerr.

> However, strangely enough, my debug shows that macOS never sends a DM (DataMark) response. Thus, it seems a network timeout is required in order to make this work. I have debug code writing "TO" for timeout and "DM" when DataMark received, and DM is never displayed.

Interesting - you definitely have a different telnetd (probably telnet/telnetd pair) than I have (I'm using home-brew). Your comment sent me testing the MacOS versions again, and (there is no end to it) I made a set of new discoveries. Including the fact that there is a 'DO TIMING MARK/WILL TIMING MARK/WONT TIMING MARK' (RFC 860) that may be useful in order to synchronize between client and server. Also I'm surprised to find that the macOS telnetd server initiates option negotiations all the time, every return, every ^C. Very chatty to say the least (see below). Other servers I've tested don't do this.

What I immediately found **valuable** in this is the observation that when the telnet server processes an IAC command that is supposed to return a DM, it does not return that immediately, but waits - until <something> - and then sends it. What that <something> is, is unclear and probably needs a peek into the sources. What it looks like though, is that the DM is held back **until there is more output to send back**. Then the DM is pushed ahead of that data. If this is the case it seems really smart because it solves the lingering data problem (discard after IAC_IP). It is also possible that the server is programmed to wait to see if there is a DO TIMING MARK coming, which apparently is supposed to be handled before the DM:
```
macOS% ^CSENT IAC IP                  <---------- typed ^C
SENT DO TIMING MARK
RCVD WILL TIMING MARK
RCVD IAC DMARK
                      <------ this is the newline echoed from the server after the ^C
more options processing... 

```
[ In the example I'm connecting  the macOS telnet client to localhost, then issue the `set options` command to telnet to see what's going on. ]

This is of course speculation until verified from the sources, but I find it really interesting. I tested this (that is, moving the DM reply) on TLVC `telnetd` (no TIMING MARK), it works well - as would be expected. I never thought of that before.

When MacOS connects to a Big Linux server, the frequent options negotiations are gone and the Time Mark is nowhere to be seen.

elkscmd/inet/telopt.c
```diff
@@ -146,14 +146,20 @@ tel_in(int fdout, int telout, char *buffer, int len)
...
-            case AO:
                 InState = IN_IAC2;
                 ThisOpt = c;
                 break;
             case SB:
                 InState = IN_SB;
                 break;
+            case IP:
+                InState = IN_DATA;
+                write(fdout, "\003", 1);
```
Mellvik: I'm assuming the syntax "\003" will create a string occupying 2 bytes. A literal 3 would also be two bytes (int), right? What about '\03'?

elkscmd/inet/telnet.c
```diff
@@ -105,7 +105,7 @@ int iscmdchar(int c)
         discard = 1;
         return 1;
     }
-    else if (c == CTRL('O')) {
+    if (c == CTRL('O')) {
```
Mellvik, line 108: This is effectively disabling ^O for most hosts, is that the purpose?

elkscmd/inet/telopt.c
ghaerr: A "\003" will add two bytes somewhere in the data segment: 3, 0, and returns the address of the start byte. `write` then writes the first byte only - exactly what's wanted.

A literal 3 is an 'int' and if cast to (char *) would return whatever is at address 3 in the data segment (this would be a big error).
A '\03' is a 'char' and if cast to (char *)(int) would also return whatever is at address 3 in the data segment (error).

Mellvik: Sorry, I obviously wasn't thinking for a moment (?), I guess the extra byte is less than an initialized char on the stack.

ghaerr:
> and (there is no end to it) I made a set of new discoveries. 

There really is no end to the madness of telnet protocol implementations - you've got that right! Seriously, I'm pretty sure that the deeper you go, the more telnet will end up like GNU telnet, and having to implement an output ring buffer in order to meet the spec. BTW, I also saw in the GNU source that it negotiates the same additional TELNET options like you're seeing, but I didn't dig into it because my plan is not to implement a standards-based telnet - far too big for ELKS.

> when the telnet server processes an IAC command that is supposed to return a DM, it does not return that immediately, but waits - until - and then sends it.

Yes - I think I kind of thought something like that after reviewing the GNU sources. All select I/O is time-stamped and put in a  huge input and output ring buffer to be processed and managed at a later time. Very complicated. I imagine they're doing all this to fully meet "the spec".

> If this is the case it seems really smart because it solves the lingering data problem (discard after IAC_IP).

No doubt perfection can be achieved with lots of code. 

This PR seems to work extremely well with almost no overhead. telnet and telnetd now handle ^C -> IAC IP and ^O -> IAC AO properly and a small timeout manages to (in my case) handle discarding output from both satisfactorily, since having any output discarded from our previous version is a huge step forward. I won't be implementing a full TELNET protocol, as I don't see much benefit in trying to achieve perfection from a ^C typed occasionally.

> This is effectively disabling ^O for most hosts, is that the purpose?

Well, looking at GNU telnet, it used VDISCARD (^O) to abort output by sending IAC AO, which is where I got this idea from. In other words, rather than passing ^O (discard output) passthrough to the remote system, I'm implementing the GNU telnet approach of handling ^O locally and sending AO.  While my version doesn't check termios.c_cc[VDISCARD] like GNU does, the result would be the same - VDISCARD is used to locally override. Note that from my previous discussion on your thread, one can use a telnet setting to not handle special characters locally, in which case the ^O would be interpreted by the local OS if VDISCARD is set locally to ^O. If not, then it would be passed through to the remote telnetd unfiltered.

In summary, for my systems, DM is never seen, and this PR discards output very well, all things considered, for both ^C and ^O. I will leave the DM processing in ("discard = 0") for later testing with other hosts, but won't likely try to setup fancy telnet negotiations for such things, as then this would all have to be added to ELKS telnetd... and the question is: for what, really? I suppose the answer is whether one wants a "spec" telnet, or just wants Interrupt Process to discard output and "mostly" work in "all" cases.

elkscmd/inet/telopt.c
ghaerr: Yes, you could say "char c = '\03'" then write(1, &c, 1), but don't forget then you're adding code to initialize the variable c, which will certainly be more than a single byte of code.

Mellvik:
> > when the telnet server processes an IAC command that is supposed to return a DM, it does not return that immediately, but waits - until - and then sends it.
> 
> Yes - I think I kind of thought something like that after reviewing the GNU sources. All select I/O is time-stamped and put in a huge input and output ring buffer to be processed and managed at a later time. Very complicated. I imagine they're doing all this to fully meet "the spec".
> 
> > If this is the case it seems really smart because it solves the lingering data problem (discard after IAC_IP).
> 
> No doubt perfection can be achieved with lots of code.

Adding a static state variable doesn't really count as 'lots of code', does it? As I recall the goal was to add functional ^C and ^O support, not perfection.

> 
> This PR seems to work extremely well with almost no overhead. telnet and telnetd now handle ^C -> IAC IP and ^O -> IAC AO properly and a small timeout manages to (in my case) handle discarding output from both satisfactorily, since having any output discarded from our previous version is a huge step forward.
> 
> > This is effectively disabling ^O for most hosts, is that the purpose?
> 
> Well, looking at GNU telnet, it used VDISCARD (^O) to abort output by sending IAC AO, which is where I got this idea from. In other words, rather than passing ^O (discard output) passthrough to the remote system, I'm implementing the GNU telnet approach of handling ^O locally and sending AO. 

As it turns out, Linux (my testbed is Ubuntu latest) doesn't support discard/^O in the tty drive or any other context that I've found. It's present in the configurations (like `stty -a` and `display` at the `telnet` prompt). Linux `telnet` simply discards (!) ^O, does not turn it into AO and does not pass it along. Finally, in `telnetd`, and AO simply returns a DM, nothing else. So the GNU telnet approach turns out to be to ignore ^O, really disappointing. Going back to the old Linux version on my RapberryPi 2, it's the same.

On the Mac it's different - supported in the tty driver and in telnet.  AO (in telnet) returns a DM and nothing else, like Linux.

So in order for telnet ^O (discard) support to be meaningful, it either has to be implemented locally in the telnet client (which I am doing) or passed on to the server, which will work with MacOS (possibly other Unix) hosts, but not with Linux servers.

> 
> In summary, for my systems, DM is never seen, 

It would be interesting to know what systems that is, I'd like to add them to my test cases! All systems I can get my hands on, emit DMs at the same (and expected) places.

ghaerr:
> No doubt perfection can be achieved with lots of code.
> Adding a static state variable doesn't really count as 'lots of code', does it?

Oops - I was referring to the perfection achieved with the GNU telnet with lots of code!

> As I recall the goal was to add functional ^C and ^O support, not perfection.

Yes. I mentioned this because I thought you might have said you were still working on eliminating some "tail end" data being received after ^C from certain telnetd systems... is that still the case? Does the dm_flag fix in your own telnetd finally remove that data from being displayed, or does it still sometimes happen with some systems?

> Linux (my testbed is Ubuntu latest) doesn't support discard/^O in the tty drive

Interesting - so ^O/discard is macOS only after all. I just looked this up and apparently ^O/discard is not Posix and the Linux termios(7) man page also says it's not supported.

> So the GNU telnet approach turns out to be to ignore ^O, really disappointing. 

Well, the GNU telnet source does show that after sending an AO, which is does on ^O, auto flush is called:
``` 
  else if (c == termFlushChar)
    {
      xmitAO ();		/* Transmit Abort Output */
      return 0;
    }
```
```
void
xmitAO (void)
{
  NET2ADD (IAC, AO);
  printoption ("SENT", IAC, AO);
  if (autoflush)
    {
      doflush ();
    }
}
```
I'm not sure whether this contradicts what you're seeing, or whether the auto flush mechanism doesn't seem to be doing much on the Linux telnet? I haven't studied the GNU telnetd source (yet) but perhaps it doesn't do anything on AO, sends no DM, and the GNU telnet ring buffer finally decides to start displaying data after a short timeout?

> On the Mac it's different - supported in the tty driver and in telnet. AO (in telnet) returns a DM and nothing else, like Linux.

Oh I see - Linux returns DM on AO, but doesn't actually flush any data. I think I understand now. Yes, so we have perform discard ourselves in the local telnet client.

> In summary, for my systems, DM is never seen,

> It would be interesting to know what systems that is, I'd like to add them to my test cases! All systems I can get my hands on, emit DMs at the same (and expected) places.

My telnetd seems to be from homebrew, from 2018. I don't have it running under automatic start, instead I start it on port 2424 using:
```
$ telnetd -debug 2424 &
```
I've attached it here so you can play with it. I've never seen a DM from it, but it does try to send DO TIMING-MARK, which of course our telnet respond with WONT. If you can get it to send a DM, that'd be great, since then I could test the DM handling in this PR. Nonetheless, this PR still seems to work quite well with this macOS telnetd and the updated ELKS telnetd with regards to discarding data appropriately.
[telnetd-63.zip](https://github.com/user-attachments/files/24538166/telnetd-63.zip)

ghaerr:
> So the GNU telnet approach turns out to be to ignore ^O, really disappointing.

I looked further at the GNU telnetd code and found it does actually "try" to flush its (PTY) buffers, reinitializes its terminal buffer, try inserting the "AO" character into the PTY stream (not sure whether that's ^O?) and then sends DM:
```
	      /*
	       * Abort Output
	       */
	    case AO:
	      {
		DEBUG (debug_options, 1, printoption ("td: recv IAC", c));
		ptyflush ();	/* half-hearted */
		init_termbuf ();

		if (slctab[SLC_AO].sptr
		    && *slctab[SLC_AO].sptr != (cc_t) (_POSIX_VDISABLE))
		  pty_output_byte (*slctab[SLC_AO].sptr);

		netclear ();	/* clear buffer back */
		net_output_data ("%c%c", IAC, DM);
		set_neturg ();
		DEBUG (debug_options, 1, printoption ("td: send IAC", DM));
		break;
	      }
```
It appears what you're saying is that either this code doesn't result in much of any output discarding, *or* perhaps this isn't the code running on Linux telnetd?

Mellvik:
> ... I mentioned this because I thought you might have said you were still working on eliminating some "tail end" data being received after ^C from certain telnetd systems... is that still the case? 

Not exactly. I said I still didn't quite understand why that data was still showing up, but that I'd leave it alone. It's the same between Linux systems and MacOs systems, there is always some 'leftovers'  part of the time.

> Does the dm_flag fix in your own telnetd finally remove that data from being displayed, or does it still sometimes happen with some systems?

No, it's like above - some data showing part of the time. But only a few like 3-8 lines typically. Good enough.

> 
> > Linux (my testbed is Ubuntu latest) doesn't support discard/^O in the tty drive
> 
> Interesting - so ^O/discard is macOS only after all. I just looked this up and apparently ^O/discard is not Posix and the Linux termios(7) man page also says it's not supported.

No, what I'm observing is that Linux inherited the termios structure (whatever the name is) from Unix and decided to keep the discard flag without implementing it. It has been there on all Unix systems I've known and I'm pretty sure it still is, Solaris, NetBSD, FreeBSD, MacOS etc.

> 
> > So the GNU telnet approach turns out to be to ignore ^O, really disappointing.
> 
> Well, the GNU telnet source does show that after sending an AO, which is does on ^O, auto flush is called:
> 
> ```
>   else if (c == termFlushChar)
>     {
>       xmitAO ();		/* Transmit Abort Output */
>       return 0;
>     }
> ```
> 
>  ```
> void
> xmitAO (void)
> {
>   NET2ADD (IAC, AO);
>   printoption ("SENT", IAC, AO);
>   if (autoflush)
>     {
>       doflush ();
>     }
> }
> ```
> 
> I'm not sure whether this contradicts what you're seeing, or whether the auto flush mechanism doesn't seem to be doing much on the Linux telnet? I haven't studied the GNU telnetd source (yet) but perhaps it doesn't do anything on AO, sends no DM, and the GNU telnet ring buffer finally decides to start displaying data after a short timeout?

I'm not sure either. autoflush works fine (with IAC_IP that is) on all systems as far as I can tell. Discard does not, and AO has no effect that I can detect except returning DM - not on Linux, not on MacOS. As to the GNU telnet/telnetd implementation, it does indeed seem like the practical observations contradict the code. It may be possible that the GNU implementation is complete in the sense that it - at least in the discard case - checks whether a certain function is implemented locally and follows that. Just a guess. After all it's supposed to work out of the box on just about any OS.

> 
> > On the Mac it's different - supported in the tty driver and in telnet. AO (in telnet) returns a DM and nothing else, like Linux.
> 
> Oh I see - Linux returns DM on AO, but doesn't actually flush any data. I think I understand now. Yes, so we have perform discard ourselves in the local telnet client.

Right. And MacOS does the same, AO has no effect. BTW, observing how differently the homebrew telnet/d implementation is from Linux (GNU) I'm guessing it comes from netBSD or FreeBSD - a completely different origin and history. 

> 
> > In summary, for my systems, DM is never seen,
> 
> > It would be interesting to know what systems that is, I'd like to add them to my test cases! All systems I can get my hands on, emit DMs at the same (and expected) places.
> 
> My telnetd seems to be from homebrew, from 2018. I don't have it running under automatic start, instead I start it on port 2424 using:
> 
> ```
> $ telnetd -debug 2424 &
> ```
> 
> I've attached it here so you can play with it. I've never seen a DM from it, but it does try to send DO TIMING-MARK, which of course our telnet respond with WONT. If you can get it to send a DM, that'd be great, since then I could test the DM handling in this PR. Nonetheless, this PR still seems to work quite well with this macOS telnetd and the updated ELKS telnetd with regards to discarding data appropriately. [telnetd-63.zip](https://github.com/user-attachments/files/24538166/telnetd-63.zip)

Interesting. I'll take a look at that. What happens if you ask brew to update telnetd?

As I write I did a `strings /opt/homebrew/sbin/telnetd` on my system, showing that it does come from FreeBSD and doesn't seem to have been updated for more than 20 years:
```
➜  TLVC git:(networking) ✗ strings /opt/homebrew/sbin/telnetd 
@(#)PROGRAM:telnetd  PROJECT:remote_cmds-
$FreeBSD: src/contrib/telnet/telnetd/termstat.c,v 1.13 2004/07/28 05:37:18 kan Exp $
$FreeBSD: src/contrib/telnet/telnetd/telnetd.c,v 1.29 2006/09/26 21:46:11 ru Exp $
$FreeBSD: src/contrib/telnet/telnetd/slc.c,v 1.9 2003/05/04 02:54:49 obrien Exp $
"9HYal
 #(&
$FreeBSD: src/contrib/telnet/telnetd/utility.c,v 1.13 2003/05/04 02:54:49 obrien Exp $
"%('!G#
)?IS
JJJJJJJ")
4;JJ
)0=$FreeBSD: src/contrib/telnet/telnetd/state.c,v 1.14 2003/05/04 02:54:49 obrien Exp $
"%(+.14$FreeBSD: src/contrib/telnet/telnetd/sys_term.c,v 1.18 2003/05/04 02:54:49 obrien Exp $
$FreeBSD: src/contrib/telnet/telnetd/global.c,v 1.6 2003/05/04 02:54:49 obrien Exp $
$FreeBSD: src/crypto/telnet/libtelnet/getent.c,v 1.2.6.2 2002/04/13 10:59:07 markm Exp $
...
```

ghaerr:
Committing this version now with no DM support in telnetd. Pending ongoing discussion and investigations in https://github.com/Mellvik/TLVC/pull/217, will add DM support in subsequent PR when able to test with GNU inetutils telnetd.

PR #2572, opened 2026-01-12
[ash] Add fg command for simple job control to ash

ghaerr:
Adds `fg` command to ash for simple job control. Use in conjunction with ^Z (terminal stop) and `jobs` command to restart stopped jobs into foreground.

ELKS `ash` doesn't implement full job control. In particular, job control doesn't work well or at all with programs that are stopped while waiting for input, due to problems related to simultaneous reading of stdin by the shell and child programs. This implementation of `fg` essentially just reads the list of stopped jobs and executes a `kill -CONT <pid>`. Use `jobs` to list all stopped or background jobs.

@Mellvik This is the fg command enhancement you requested (the previously described ash enhancements must also be included).

Screenshot showing sample `out` program (output's X's) being run and stopped twice, then restarted using fg:
Screenshot 2026-01-12 at 4 01 47 PM

### DIFF
--- a/elkscmd/ash/builtins.table
+++ b/elkscmd/ash/builtins.table
@@ -61,7 +61,7 @@ execcmd 	exec
 exitcmd 	exit
 exportcmd	export readonly
 exprcmd 	expr test [
-fgcmd -j	fg
+fgcmd		fg
 getoptscmd	getopts
 hashcmd 	hash
 jobidcmd -j	jobid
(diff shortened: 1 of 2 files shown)


Mellvik:
Nice, @ghaerr, this is good. an extra treat (!) would be to add '%' syntax to the command, (`fg %2`). Its may not make sense at this point but let me ask anyway: Is `fg` w/o parameters supported? Not having '+' and '-' available in the jobs listing, would `fg` alone simply take the highest job #?

ghaerr:
@Mellvik, all your requested treats have been added in #2574. 

`fg` without parameters will reference the most recently-run or accessed stopped job. 

So everything basically now works as expected.

PR #2581, opened 2026-01-17
[CI] Build Nano-X with Github Continuous Integration

ghaerr:
This PR allows automatic generation of ELKS HDD image with Nano X applications integrated. Nano-X is now built from an external repository on every PR.

Replaces @toncho11's original request in #2562 and adds the ELKS build of Nano-X from [The Nano-X Window System](https://github.com/ghaerr/microwindows) to the 32M and 64M HD MINIX and FAT images. 

Nano-X is not included on any of the FD images (including 2880k floppy recently, unfortunately) as the distribution is too large. I would like to get the 2880k floppy distribution down a few hundred blocks so that Nano-X can be included, but have not yet had the time to figure out what to remove.

@toncho11, let me know how this works for you, thanks!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -122,15 +122,6 @@ jobs:
           name: fd2880-fat.img
           path: image/fd2880-fat.img
 
-      - name: upload1232
-        uses: actions/upload-artifact@v4
-        with:
-          name: fd1232.img
-          path: image/fd1232.img
-
-      - name: buildnx
-        run: ./buildnx.sh
-
       - name: upload14
         uses: actions/upload-artifact@v4
         with:
@@ -154,3 +145,9 @@ jobs:
         with:
           name: hd64mbr-minix.img
           path: image/hd64mbr-minix.img
+
+      - name: upload18
+        uses: actions/upload-artifact@v4
+        with:
+          name: fd1232.img
+          path: image/fd1232.img
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -122,6 +122,15 @@ jobs:
           name: fd2880-fat.img
           path: image/fd2880-fat.img
 
+      - name: upload1232
+        uses: actions/upload-artifact@v4
+        with:
+          name: fd1232.img
+          path: image/fd1232.img
+
+      - name: buildnx
+        run: ./buildnx.sh
+
       - name: upload14
         uses: actions/upload-artifact@v4
         with:
@@ -145,9 +154,3 @@ jobs:
         with:
           name: hd64mbr-minix.img
           path: image/hd64mbr-minix.img
-
-      - name: upload18
-        uses: actions/upload-artifact@v4
-        with:
-          name: fd1232.img
-          path: image/fd1232.img
--- a/build.sh
+++ b/build.sh
@@ -66,10 +66,18 @@ make -j1 all || clean_exit 5
 # Possibly build all images
 
 if [ "$2" = "allimages" ]; then
-	echo "Building all images..."
+	echo "Building FD images..."
 	cd image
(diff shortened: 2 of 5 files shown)


toncho11:
Looks very good!  nxdsktop works on both the minix and fat images. 
Awesome! :)

toncho11:
Maybe there could be a PC-98 HDD (minix or fat) version with Nano X? This will require both ELKS and Nano X to be complied for it. Just an idea.

tyama501:
Actually pc-98 cannot boot current MBR FAT HDD nor minix HDD.
User need to take out files from the HDD image by emulators or some image editor then copy files using FD or serial port to PC-98 harddisk.

tyama501:
Is it possible that nano-x and nxstart remain in fd image?

ghaerr:
> Is it possible that nano-x and nxstart remain in fd image?

Yes, but a bit more work needs to be done to enable that. With the addition of `nxdsktop` and other enhancements, all the Nano-X binaries no longer fit on even a 2880k floppy. I would like to see most NX apps on 2880k, and just nano-x, nxterm and nxstart on 1440k (or nxdsktop, etc).

This PR introduced the first step of building a separate repository, which for now automatically copies all its binaries to /bin. I am thinking of adding an ExtApplications (or similar named) file that will allow selecting which binaries from which remote repositories will be copied onto which images.

The process is a little complicated because Nano-X (and possibly other repositories) may have to be specially built for PC-98 as well, just like ELKS is.

The next step will be to add ExtApplications, then add more external repository compilations and specific platform builds. This will then enable us to finally release v0.9 with a much larger set of updated applications.

ghaerr:
> Actually pc-98 cannot boot current MBR FAT HDD nor minix HDD.

@tyama501, what tools or modifications would be required to create a SCSI-compatible HDD image for PC-98?

tyama501:
Thank  you @ghaerr 

There are still some spaces for fd1232c since I haven't included much applications in it (to keep some nano-X).
I have been working for window managers for rev0.9 so it would be nice to have at least nano-x, nxterm, nxstart, and some others if possible.

I need to find tools that open source license and can be built in linux to create PC-98 partitions,
but I am not sure there are one.
(I am using some dos tools)

tyama501:
Even the bootable HDD image can be created, copying files to the real HDD on the real PC-98 still be needed since the image cannot be copied directly to the real HDD.

tyama501:
FYI,

currently about 500KB free in fd1232c.

image

Thank you.

ghaerr:
> I need to find tools that open source license and can be built in linux to create PC-98 partitions,

I think tools/setboot.c should be able to be somewhat easily modified to create PC-98 partitions, since it already creates DOS MBR partitions. IIRC there is not much difference between the two, details can be found in drivers/block/genhd.c.

> Even the bootable HDD image can be created, copying files to the real HDD on the real PC-98 still be needed since the image cannot be copied directly to the real HDD.

I see. I was thinking perhaps building PC-98 HDD would be good for emulator testing of larger application distribution, but maybe it's not worth the effort since image cannot be copied directly.

> currently about 500KB free in fd1232c.

Thanks, that's more than I thought. I will add fd1232c to ExtApplications.

ghaerr:
> currently about 500KB free in fd1232c.

I just tested the 2880k build again, and if nxworld, /lib/nxworld.map and the sample \*.jpgs are removed, all the other Nano-X applications will fit, including nxdsktop and its companion programs (but not the sample images). That fits in ~300k, so I will start with that and the work to see what will fit on uncompressed 1440k floppy.

tyama501:
That's great. Thank you.

toncho11:
I booted from CD https://github.com/ghaerr/elks/discussions/2583 in an emulator.

Issue #2575, opened 2026-01-14
NEC V25 hangs on IN/OUT instructions in ATA CF driver


swausd:
Hi @ghaerr,

I modified my NEC V25 system and now I can use it as CPU card on a passive ISA backplane using only 8 Bit slots. Works fine with a NE200 NIC in 8 bit mode. I added this XT-IDE Rev. 4 card from a german ebay site:

https://www.ebay.com/itm/226138176425?_skw=xt-ide&itmmeta=01KEYMPV9NCA5FDF0Z8E51Z8XW&hash=item34a6e2bba9:g:j7cAAOSwqHFmPTAL

The card has no dip-switches, but solder points on the back where options can be configured by cutting traces or adding solder. I only did one change from standard configuration to disabled the flash, because it would be in conflict with my memory map and it is not used anyways.

The card is used with an IDE to CF adapter and I am using an old 64MB Sandisk CF. This config works OK in another environment in 8 bit mode.

To escape IO address conflicts the NIC is configured to IO 0x200 and the XT-IDE is in standard config IO 0x300. I configure the ATA driver to 

    ata_mode = MODE_XTIDEv2;
    xfer_mode = XFER_8_XTCF;

and on first sight it seems to work. I can copy and read files but sometimes the system hangs. This can be reproduced doing this with the current ELKS:

```
ELKS Setup INT f002 START
16K ext buffers, 8K cache, 1 req hdrs
eth: ne0 at 200, irq 2, (ne1k) MAC 00:4f:4c:02:4f:6b, flags 0x80
ssd: 2013265920K disk
cfa: 8-bit xfer on
cfa:   61M CHS 490, 8,32 62720K VER 0/576 xtide=2,1
cfb: ATA at 300/307 xtide=2,1 not found (-22)
cfa:(0,125440) cfa1:(1,125439) 
NECV25 machine, cpu 3, syscaps 0, 512K base ram, 16 tasks, 64 files, 96 inodes
ELKS 0.9.0-dev (51504 text, 0 ftext, 4768 data, 4816 bss, 55950 heap)
Kernel text e062 data 80 end 1080 top 8000 446+0+0K free
VFS: Mounted root device /dev/rom (0600) romfs filesystem.
Running /etc/rc.sys script
Wed Jan 14 16:59:02 2026

ELKS 0.9.0-dev

[0.68 secs] login: root
# mkfs /dev/cfa 16384
5461 inodes
16384 blocks
Firstdatazone=176 (176)
[...]

PR #2582, opened 2026-01-17
[NECV25] New IBRK exception handler, inclusion of DIVZERO and NMI interrupt handler, current NEC V25 config

swausd:
This PR adds a NEC V25 specific exception handler discussed here #2575. If the exception occurres, it is caught by the handler and the triggering IO instruction is re-executed. There is also a new config option in config.h to enable/disable the output of an associated message.

Also some configuration changes for integrating XT-IDE and NE2000 cards in the same NEC V25 system and some cleanups are included..


ghaerr:
Nicely done, especially the IBRK handler setting IBRK to restart the instruction! Were you able to actually test that? 

It seems the NECV25 doesn't increment IP and requires the IBRK handler to skip over the offending instruction, if IBRK not set in the handler? I suppose that'd be even harder, since it would require special decoding instructions to determine exactly what instruction was used for intended I/O...

elks/include/arch/irq.h
```diff
@@ -29,6 +34,9 @@ void idle_halt(void);
 
 void div0_handler(int irq, struct pt_regs *regs);
 void nmi_handler(int irq, struct pt_regs *regs);
+#ifdef CONFIG_ARCH_NECV25
+void ibrk_handler(int irq, struct pt_regs *regs);
+#endif
```
ghaerr: I may remove this ifdef and just add a comment that ibrk_handler is NECV25 specific, to reduce ifdef overload. Same with lines 9,12,14 - even though N_IRQS would waste two bytes on non-V25 systems, it's cleaner to not have configuration-specific ifdefs in the header files unless really necessary.

elks/arch/i86/kernel/ibrk-necv25.c
```diff
@@ -0,0 +1,32 @@
...
+ * about to excecute, an INT-13 interrupt is activated. Setting
+ * the IBRK here re-excecutes the IO instruction on return from interrupt.
+ */
+
+void ibrk_handler(int irq, struct pt_regs *regs)
+{
+    if (intr_count > 1 /*|| current->t_regs.ss == kernel_ds*/) {
+        struct uregs __far *sys_stack;
+        sys_stack = _MK_FP(regs->ss, regs->sp);
+#if IBRK_VERBOSE_MODE == 1
```
ghaerr: Pretty cool handler showing exactly where the problem is, then restarting the instruction!

I would suggest a simpler ifdef here too, just `#if IBRK_VERBOSE_MODE` which will work when == 1 and not when == 0, since multiple modes won't be used. This method of using `#if` is used in lots of other ELKS kernel code also.

swausd:
> Nicely done, 

With your support and the already prepared divzero handler, this was easy. So thank you very much for that!

> especially the IBRK handler setting IBRK to restart the instruction! Were you able to actually test that?

Yes, I have tested it and it did work.

> It seems the NECV25 doesn't increment IP and requires the IBRK handler to skip over the offending instruction, if IBRK not set in the handler? 

Yes that's the special handling of the emulation mode. Otherwise I probably would have skipped the restart.

I prepare the changes you mentioned above and submit in just a minute...

ghaerr:
Looks fantastic! Thank you for this nice update, it's a pleasure working with you. I especially like the instruction restart on V25 I/O, very cool. I only wish the 8086 didn't increment IP before a DIV 0 fault (and then that decision reversed in later processors!).

PR #2588, opened 2026-01-20
[cmds] Add elks-viewer image viewers as external application

rafael2k:
Add support for building elks-viewer (BMP, PPM, JPG image viewers) as an external application using OpenWatcom C compiler. Source repository: https://github.com/rafael2k/elks-viewer

Could you please check if I got the idea right?

ps: I'm still compiling latest ELKS (many months without compiling it!) - so code still untested!
ps2: I don't quite remember if there was any missing feature when compiling with C86, this is why I submitted the MR using OWC

### DIFF
--- a/buildext.sh
+++ b/buildext.sh
@@ -19,6 +19,7 @@
 #       c86_elkscmd     C86             Some elkscmd/ programs compiled by C86
 #       doom            OpenWatcom      Doom for ELKS
 #       ngircd_elks     OpenWatcom      IRC daemon for ELKS
+#       elks_viewer     OpenWatcom      Image viewers (BMP, PPM, JPG) for ELKS
 #
 # Some projects may require prerequisites.
 # To only build the C86 toolchain, use './buildext.sh owc_libc c86_toolchain'
@@ -197,6 +198,20 @@ elkirc()
     echo "elkirc build complete"
 }
 
+elks_viewer()
+{
+    echo "Building elks-viewer..."
+    cd $TOPDIR/extapps
+    if [ ! -d elks-viewer ] ; then
+        git clone https://github.com/rafael2k/elks-viewer
+    fi
+    cd elks-viewer
+    git pull
+    make -f Makefile.owc clean
+    make -f Makefile.owc
+    echo "elks-viewer build complete"
+}
+
 # build all extapps repos
 make_all()
 {
@@ -210,6 +225,7 @@ make_all()
         c86_elkscmd
         doom
         ngircd_elks
+        elks_viewer
     fi
 }
 
--- a/elkscmd/ExtApplications
+++ b/elkscmd/ExtApplications
@@ -51,6 +51,10 @@ ngircd-elks/ngircd.os2                              :ngircd
 ngircd-elks/ngircd.conf ::etc/ngircd.conf           :ngircd
 elkirc/elkirc                                       :elkirc             :2880k
 elkirc/man/elkirc.1 ::lib/man1/elkirc.1             :elkirc             :2880k
+# elks-viewer image viewers
+elks-viewer/jpgview                                 :elksviewer         :2880k
+elks-viewer/ppmview                                 :elksviewer         :2880k
+elks-viewer/bmpview                                 :elksviewer         :2880k
 # sample ELKS OWC/C86 files
 ../elkscmd/basic/basic.os2                          :owc                :2880k
 ../elkscmd/fsck_dos/fsck-dos.os2                    :owc                :2880k


ghaerr:
Looks great, thanks!

> I'm still compiling latest ELKS (many months without compiling it!) - so code still untested!

Have you tested this build using `./buildext elks-viewer` on your host system? It would also be good to make sure the binaries function as intended afterwards too. Please let me know, then I'll commit.

rafael2k:
It built!
```
...
elks-viewer build complete
Buildext.sh has completed successfully.
Target image is in 'image' folder.
Build script has completed successfully.
```

Tomorrow I test!

rafael2k:
Dumb question - how to automatically copy the compiled external software to the HD image (set in menuconfig)?

ghaerr:
> how to automatically copy the compiled external software to the HD image (set in menuconfig)?

When you run `cd image; make images`, it should add the files in ExtApplications to all images that match the :2880k (or whatever image size). If just running "make", then ensure that your .config has CONFIG_APPS_BY_IMAGESZ=y and that the appropriate CONFIG_APPS_2880K (or whatever) option is set.

[EDIT: Oh - are you building an HD image and the above is not working? A new CONFG_APPS_HD option was added and used in images/Makefile. It seems this may need to be added to elkscmd/config.in. I'll add that now.]

ghaerr:
@rafael2k, CONFIG_APPS_HD should be fixed in commit 5a77d5feeb2bb9b16e6f3efbd556c64834c4be31.

rafael2k:
I need to "make clean", then make again right? 
Just make in elks root and make in image/ did not included the binaries.

ghaerr:
> I need to "make clean", then make again right?

I'm not sure which of the two above discussed options you're using. Check that CONFIG_APPS_HD is in .config, that should then work. You can use `make kimage` which is much faster than make clean; make, if only image size is changed.

ghaerr:
You can also `cd image; make hd32-minix` which will always create the hd32-minix.img with any flagged programs :2880k or larger from ExtApplications.

rafael2k:
Adding CONFIG_APPS_HD worked! All good.
There was a couple of jpgs in the /root:
image
image

EDIT: These jpegs in /root were handy, but are they suposed to be there?

rafael2k:
btw, we need to update file at some point.
: )
image

ghaerr:
> These jpegs in /root were handy, but are they suposed to be there?

They are there temporarily as an easy way for @toncho11's to test `nxdsktop`, but should be removed or moved elsewhere before the final v0.9 distribution.

> we need to update file at some point.

Thanks, I'll fix that.

toncho11:
@ghaerr I did not think of the 2 images as temporary. They are showcase images and in opinion they should always be there. I lost a lot of time checking the license of both exactly so that they can stay. We can use images from Pexels.com. We could put them in /root/images or /root/mypics? Modern OS have folders such as MyPictures, MyMusic, etc. These images will be used by both elksviewer apps and [nxjepg](https://github.com/toncho11/microwindows/blob/start_menu/src/demos/nanox/nxjpeg.c). @rafael2k there is new image viewer for Nano X called [nxjpeg](https://github.com/toncho11/microwindows/blob/start_menu/src/demos/nanox/nxjpeg.c) that uses your version of picojpeg. You can learn more about it in the source file [nxjpeg](https://github.com/toncho11/microwindows/blob/start_menu/src/demos/nanox/nxjpeg.c). Nxjpeg needs to be improved and some small problems fixed, but it works well with no crashes. It is max 16 colors, which complicates things a lot.

rafael2k:
This is pretty cool @toncho11! I did not know about nxjpeg! picojpeg is a pretty neat jpeg implementation.

EDIT: about the location of the pictures, I have no strong feelings. I have some demo pictures here:
https://github.com/rafael2k/elks-viewer/tree/master/images
I like the racecar. The jpeg, for example, is here: https://github.com/rafael2k/elks-viewer/blob/master/images/racecar.jpg

rafael2k:
btw, one thing that I did not implement (yet) in elks-viewer, but would improve a lot the subjective quality of images when using 16 colors mode is use dithering.

ghaerr:
> I did not think of the 2 images as temporary. They are showcase images and in opinion they should always be there.

Yes, I meant temporarily in /root, as users won't expect to have arbitrary data installed in their home directories. But for now and possibly the first distribution, this should be OK and allow more visibility into ELKS capabilities, while we figure out a longer term home for this kind of stuff.

> We could put them in /root/images or /root/mypics? 

I don't think a user or root home directory is the best place. /lib comes to mind though, or possibly /demos. Creating a new directory has implications for floppy users where space is at a premium. For now, we'll keep things in /root and wait to see what users (and ourselves) think after using some of the many contributed programs :)

PR #2594, opened 2026-01-21
[cmds] Add Lua interpreter as external application

rafael2k:
Add support for building Lua 5.5 interpreter as external
application using OpenWatcom C compiler.

Source repository: https://github.com/rafael2k/lua
Includes drawcube.lua sample in /root.

### DIFF
--- a/buildext.sh
+++ b/buildext.sh
@@ -20,6 +20,7 @@
 #       doom            OpenWatcom      Doom for ELKS
 #       ngircd_elks     OpenWatcom      IRC daemon for ELKS
 #       elks_viewer     OpenWatcom      Image viewers (BMP, PPM, JPG) for ELKS
+#       lua             OpenWatcom      Lua 5.5 interpreter
 #
 # Some projects may require prerequisites.
 # To only build the C86 toolchain, use './buildext.sh owc_libc c86_toolchain'
@@ -212,6 +213,20 @@ elks_viewer()
     echo "elks-viewer build complete"
 }
 
+lua()
+{
+    echo "Building Lua 5.5..."
+    cd $TOPDIR/extapps
+    if [ ! -d lua ] ; then
+        git clone https://github.com/rafael2k/lua lua
+    fi
+    cd lua
+    git pull
+    make -f Makefile.elks clean
+    make -f Makefile.elks
+    echo "Lua 5.5 build complete"
+}
+
 # build all extapps repos
 make_all()
 {
@@ -226,6 +241,7 @@ make_all()
         doom
         ngircd_elks
         elks_viewer
+        lua
     fi
 }
 
--- a/elkscmd/ExtApplications
+++ b/elkscmd/ExtApplications
@@ -55,6 +55,9 @@ elkirc/man/elkirc.1 ::lib/man1/elkirc.1             :elkirc             :2880k
 elks-viewer/jpgview                                 :elksviewer         :2880k
 elks-viewer/ppmview                                 :elksviewer         :2880k
 elks-viewer/bmpview                                 :elksviewer         :2880k
+# Lua interpreter
+lua/lua                                             :lua                :2880k
+lua/testes/drawcube.lua ::root/drawcube.lua         :lua                :2880k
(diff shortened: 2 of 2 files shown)


rafael2k:
Tbh, I'm unsure about using Lua 5.1 as default... it seems to have some kind of memory leak with drawcube. Lua 5.5 is working so much smoother (and is the latest release, with 14 years of improvements since 5.1...).

ghaerr:
Why do you we need seperate lua 5.1 and lua 5.5 versions? Does ELKS really require multiple versions of a lua interpreter? I'm not aware of anyone actually using LUA for anything except a demo, currently. (I'm also not aware of the apparently huge difference between lua versions, but this would seem to only be important for pre-written applications, right? Otherwise why not distribute the latest working lua version and leave it at that?)

The issues involved with compiling every version add up - every commit has to (re)compile all the software, and ultimately slows down the process of submitting PRs. We're already at the 3.5 minute mark per commit.

rafael2k:
Let's @toncho11 also carry his tests. I was thinking about keeping just the old Lua 5.1, because it is smaller. But Lua 5.5 seems to be much more stable and faster. Give us a couple of days to decide.
: )

ghaerr:
> I'm unsure about using Lua 5.1 as default... it seems to have some kind of memory leak with drawcube

> Lua 5.5 is working so much smoother

> But Lua 5.5 seems to be much more stable and faster. 

IMO, you've just answered your (and my) question. Lets drop Lua 5.1 and just distribute and compile 5.5 only. Compilation times will be quicker, and we can respond when people actually start using the software, rather than confusingly shipping three versions.

> the old Lua 5.1, because it is smaller. 

All this is going on HDD only, so that matters not. But if its buggy, we shouldn't ship it at all at this time.

> Give us a couple of days to decide.

Ok. I'll commit this and then we'll decide on a single Lua interpreter, how's that?

ghaerr:
The purpose of our initial distribution should not be to just ship "whatever we have", but to ship quality software that shows off ELKS. Buggy software should not be distributed, just because we invented a distribution method.

rafael2k:
Agreed. But as you said... 
99% chance we'll keep with the latest greatest Lua 5.5. Lets hear from @toncho11 too.

rafael2k:
> The purpose of our initial distribution should not be to just ship "whatever we have", but to ship quality software that shows off ELKS. Buggy software should not be distributed, just because we invented a distribution method.

Totally agree. I'm carrying some tests here.

rafael2k:
Btw, I wrote in Lua mailing that we ported Lua to ELKS, and a doctoral student of Roberto (Lua's author) told me he still has in his room a IBM 5150! When I go to Rio I'll try to sneak in and install ELKS + Lua (master) for him, if he allows me.
: )

rafael2k:
OK, I pulled the trigger. Lets keep just Lua 5.5, unless there is a good reason to revisit this topic. It is working fine and stable. After some tests, I'm confident this is the way to go.

ghaerr:
Thanks for re-testing and checking out your Lua versions. Should you want to update or change this in the future, that's always possible, of course.

ghaerr:
Hi @rafael2k,

The 2880k disk image just ran out of space with this PR: https://github.com/ghaerr/elks/actions/runs/21228257792/job/61080755009.

I've fixed it with f529d7dc6bfc3c22fb29235aa4dc6ba7a7672e8d.

All the extra OWC apps will have to be on the HDD images only, at least for now.

Thank you!

rafael2k:
> Hi @rafael2k,
> 
> The 2880k disk image just ran out of space with this PR: https://github.com/ghaerr/elks/actions/runs/21228257792/job/61080755009.
> 
> I've fixed it with [f529d7d](https://github.com/ghaerr/elks/commit/f529d7dc6bfc3c22fb29235aa4dc6ba7a7672e8d).
> 
> All the extra OWC apps will have to be on the HDD images only, at least for now.
> 
> Thank you!

HDD is the way to go! If anyone does not have HDD, at least a second floppy drive will be needed anyway to place all tools one might need for development.

PR #2595, opened 2026-01-21
[cmds] Add Bobcat web browser as external application

rafael2k:
Add support for building Bobcat (a Lynx fork) text-mode web browser as external application using OpenWatcom C compiler.

Source repository: https://github.com/rafael2k/bobcat

### DIFF
--- a/elkscmd/ExtApplications
+++ b/elkscmd/ExtApplications
@@ -61,6 +61,8 @@ elks-viewer/ppmview                                 :elksviewer
 elks-viewer/bmpview                                 :elksviewer
 lua/lua                                             :lua
 lua/testes/drawcube.lua ::root/drawcube.lua         :lua
+bobcat/src/bobcat                                   :bobcat
+bobcat/lynx.cfg ::etc/lynx.cfg                      :bobcat
 # C86 devkit and header files
 8086-toolchain/elks-bin/make        ::usr/bin/make
 8086-toolchain/elks-bin/cpp         ::usr/bin
--- a/buildext.sh
+++ b/buildext.sh
@@ -21,6 +21,7 @@
 #       ngircd_elks     OpenWatcom      IRC daemon for ELKS
 #       elks_viewer     OpenWatcom      Image viewers (BMP, PPM, JPG) for ELKS
 #       lua             OpenWatcom      Lua 5.5 interpreter
+#       bobcat          OpenWatcom      Bobcat web browser (Lynx fork)
 #
 # Some projects may require prerequisites.
 # To only build the C86 toolchain, use './buildext.sh owc_libc c86_toolchain'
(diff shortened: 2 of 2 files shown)


elkscmd/ExtApplications
```diff
@@ -58,6 +58,8 @@ elks-viewer/bmpview                                 :elksviewer         :2880k
 # Lua interpreter
 lua/lua                                             :lua                :2880k
 lua/testes/drawcube.lua ::root/drawcube.lua         :lua                :2880k
+# Bobcat web browser (Lynx fork)
+bobcat/src/bobcat                                   :bobcat             :2880k
```
ghaerr: Sorry @rafael2k, your previous PR broke the 2880k build, we're out of space. Can you change this by removing the `# Bobcat ...` comment as well as the `:2880k` setting on the install?

rafael2k: Lets also remove Lua from the 2880k build?

ghaerr: That's already done, do a git pull before proceeding.

rafael2k:
Btw, I'm fixing the links navigation, so we'll have a pretty working WWW browser.
image

ghaerr:
> I'm fixing the links navigation, so we'll have a pretty working WWW browser.

Nice! Can't wait to try it! :)

ghaerr:
@rafael2k, I can commit this after the branch conflict is resolved, thanks.

rafael2k:
I tried to fix the navigation artifacts, but I'm not there yet. Nevertheless, networking works and simple sites render fine. It is possible to access the links, just the rendering of the highlighted text of the link gets to the wrong coordinates...
: /

ghaerr:
Glad you made it through the git mountain!! :) 

> I tryed to fix the navigation artifacts, but I'm not there yet.

That's ok, the good news is that bobcat will now automatically update on ELKS after commits to your bobcat repo. I'll follow it so I can see what you're doing.

ghaerr:
>  just the rendering of the highlighted text of the link gets to the wrong coordinates...

Actually, bobcat seems more unusable than before... just moving the arrow keys or trying to access google.com does not work:
Screenshot 2026-01-21 at 6 32 54 PM

rafael2k:
OK, yesterday I tried some stuff, did not work. Later today after work I'll revert all the mess I tried.

ghaerr:
Hello @rafael2k, can you please revert the previous cursor-fix changes to the bobcat repo? I would like to take a look and determine how suitable it is for the upcoming v0.9 distribution. The most recent changes render the application pretty much unusable.

rafael2k:
Reverted. Indeed... I tried AI, but of no help... No training data of modern ELKS software (OWC, etc) for AI yet.

rafael2k:
I have a feeling that if fixing bobcat turns out too painful, we might write a simple html browser from scratch.

ghaerr:
Thanks @rafael2k, I'll take another look at how it runs and comment back.

Since Bobcat does work on other systems, I would hope the issue is some kind of mixup with curses. However, if Bobcat depends on curses "windows", meaning auto-save of underlying screen contents while a rectangular area of the screen is replaced with new text, then that will require a much larger implementation of our tiny ELKS curses, which probably isn't going to get written very easily.

ghaerr:
@rafael2k, just tested bobcat again. It's better than it was, but does seem to scroll up a line possibly. But the bigger problem is that it almost always immediately runs out of heap, and then doesn't accept any keyboard input at all - until a ^C is typed, and the program exits.

Were you ever able to get it to display any web page? What URL actually works?

rafael2k:
I test with this one:
http://info.cern.ch/

If we had C++, we could work on https://github.com/jhhoward/MicroWeb

PR #2598, opened 2026-01-23
[kernel] Add preliminary support for PS/2 mouse driver /dev/psaux

ghaerr:
Requested and discussed in https://github.com/ghaerr/microwindows/issues/109#issuecomment-3790802841.

This PR takes @toncho11's initial PS/2 mouse driver and adapts it into the TTY infrastructure in ELKS. The driver is not tested on any real hardware nor emulator. The mouse driver is accessible from /dev/psaux and likely only supports the base 3-byte PS/2 protocol. I have tested the basic operation of the TTY infrastructure around the driver, which works. I doubt the driver works as-is though.

There are likely serious issues with this driver, noted within the source with FIXME comments. These include the following:
- PS/2 controller status bit 1 should be checked before writing data to it
- IRQ 12 is not likely enabled using the required 8042 command sequence
- Opening /dev/psaux without a PS/2 controller will likely hang the system
- IRQ 12 interrupts need to be disabled on the controller on mouse close or the system may crash

These will be further researched by @toncho11 and tested on copy.sh/v86.

For now, the driver is always compiled and included on IBM PC builds by having the CONFIG_MOUSE_PS2 define set in linuxmt/config.h.

There is further work to do in Nano-X and ELKS to automatically access this driver when MOUSE_TYPE=ps2. I will be adding those changes either within this PR or as follow ups. The ELKS `mouse` program will be included allowing for easier debugging by enabling a user to see mouse activity using ASCII print statements rather than inside a GUI where that is much harder.

### DIFF
--- a/elks/arch/i86/drivers/char/Makefile
+++ b/elks/arch/i86/drivers/char/Makefile
@@ -128,6 +128,10 @@ ifdef CONFIG_CHAR_DEV_RS
 	endif
 endif
 
+ifdef CONFIG_ARCH_IBMPC
+  OBJS += mouse-ps2.o
+endif
+
 ifdef CONFIG_CHAR_DEV_CGATEXT
 	OBJS += cgatext.o
 endif
--- a/elks/arch/i86/drivers/char/console.h
+++ b/elks/arch/i86/drivers/char/console.h
@@ -2,7 +2,6 @@
 
 void Console_conin(unsigned char Key);
 void Console_conout(dev_t dev, int Ch);
-extern struct tty ttys[];
 
 void kbd_init(void);
 extern char kbd_name[];
--- a/elks/arch/i86/drivers/char/mouse-ps2.c
+++ b/elks/arch/i86/drivers/char/mouse-ps2.c
@@ -1,0 +1,120 @@
+/*
+ * Minimal PS/2 mouse driver for ELKS
+ *
+ * 23 Jan 2026 Original version by @toncho11, adapted by Greg Haerr
+ */
+
+#include <linuxmt/config.h>
+#include <linuxmt/wait.h>
+#include <linuxmt/chqueue.h>
+#include <linuxmt/sched.h>
+#include <linuxmt/errno.h>
+#include <linuxmt/ntty.h>
+#include <arch/io.h>
+#include <arch/irq.h>
+
+#ifdef CONFIG_MOUSE_PS2
+
+#define IRQ_MOUSE       12      /* mouse interrupt, may conflict with networking */
+
+/* i8042 ports */
+#define I8042_DATA      0x60
+#define I8042_STATUS    0x64
+#define I8042_CMD       0x64
+
+/* status bits */
+#define ST_OBF          0x01
+#define ST_AUXDATA      0x20
+
+/* Minimal mouse enable (no reset, BIOS-friendly) */
+static void ps2_mouse_enable(void)
+{
+    /* FIXME controller ready (status bit 1) should be read before each write!! */
+    /* Enable AUX (mouse) device */
+    outb(0xA8, I8042_CMD);
+
+    /* FIXME unclear whether the following enables IRQ 12 */
+    /* Enable data reporting */
+    outb(0xD4, I8042_CMD);   /* next byte to mouse */
+    outb(0xF4, I8042_DATA);  /* enable streaming */
+
+    /* FIXME this could cause OS hang if no PS/2 controller present */
+    /* Eat ACK (0xFA) */
+    while (!(inb(I8042_STATUS) & ST_OBF))
+        ;
+    inb(I8042_DATA);
+}
+
+/* IRQ handler */
+static void ps2_irq(int irq, struct pt_regs *regs)
+{
+    struct ch_queue *q = &ttys[NR_CONSOLES + NR_PTYS + NR_SERIAL].inq;
+    unsigned char c;
+
+    /* Read all available mouse bytes */
+    while ((inb(I8042_STATUS) & (ST_OBF | ST_AUXDATA)) == (ST_OBF | ST_AUXDATA)) {
+        c = inb(I8042_DATA);
+        chq_addch_nowakeup(q, c);
+    }
+
+    if (q->len)
+        wake_up(&q->wait);
+}
+
+static int ps2_open(struct tty *tty)
+{
+    int err;
+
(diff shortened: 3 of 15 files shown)


ghaerr:
The ELKS `mouse` command as well as the GUI `paint` program have both been updated to use this experimental PS/2 kernel mouse driver on /dev/psaux, when MOUSE_TYPE=ps2 in the environment. 

Using `mouse` will greatly aid the ability to debug the kernel, since it can be run in a shell script or `MOUSE_TYPE=ps2 mouse` to display mouse movement in ASCII on the console, rather than inside a window system. Moving the mouse should display the button status and X,Y coordinates.

The Nano-X mouse driver has also been updated to automatically use the /dev/psaux port when MOUSE_TYPE=ps2. It already supported PS/2 mice so the rest of that driver is unchanged.

rafael2k:
Wow, I think my modernish Thinkpad trackpad is PS2! I might be able to test this in real hardware. I'll double check

tyama501:
Hello @ghaerr 

I have tried fd1440-fat.img on FabGL/PCEmulator with PS2 mouse,
but it says can't open mouse.
Do I need any other setting?

IMG_20260125_161848

Thank you.

toncho11:
@tyama501 we must be sure that FabGL/PCEmulator provides / exposes the PS2 hardware (8042 controller) to ELKS. If you ever need a different board: https://www.olimex.com/Products/Retro-Computers/ESP32-SBC-FabGL/open-source-hardware

toncho11:
I just saw that actually @ghaerr improved the driver. I personally am unable to test or work on it for the moment.

tyama501:
Yes, I have used the latest one.

ghaerr:
@tyama501 Are you running the standard IBM PC configuration, or PC-98? The /dev/psaux mouse is only compiled in on IBM PC right now. Also, @toncho11 is correct that the PS/2 mouse is actually implemented within the 8042 controller, so that must be emulated as well.

To try the PS/2 mouse on other architectures, add \#define CONFIG_MOUSE_PS2 in config.h.

I can't figure out errno 22 (=EINVAL), the only place that is in the driver is in `ioctl`. Are you running the direct console, or something else? You might try making the following change to comment out the -EINVAL return in ioctl in the mouse driver to see if that makes any difference:
```
diff --git a/elks/arch/i86/drivers/char/mouse-ps2.c b/elks/arch/i86/drivers/char/mouse-ps2.c
index 8f9a6088..3e8659aa 100644
--- a/elks/arch/i86/drivers/char/mouse-ps2.c
+++ b/elks/arch/i86/drivers/char/mouse-ps2.c
@@ -144,7 +144,8 @@ static int ps2_ioctl(struct tty *tty, int cmd, char *arg)
     case TCSETSF:
         return 0;
     }
-    return -EINVAL;
+    //return -EINVAL;
+    return 0;
 }
 
```
Also add the line  to see the ioctl being sent to mouse driver:
```
    printk("ioctl %04x\n", cmd);
    switch (cmd) {
```
You might also check `ls -l /dev/psaux` to make sure it exists as char device 4,32.

Thank you!

tyama501:
I am using IBM fd1440-fat from github actions.

I assume the mouse is attached to the i8042.
It works on dos.

FabGL/examples/VGA/PCEmulator/machine.cpp at master · fdivitto/FabGL https://share.google/hCzL1IfeS3IEMt5bI

It seems it also uses INT.

I will the char dev although it is existing.

Thank you.

ghaerr:
@tyama501 Can you send me a picture of the boot screen? I have tracked down the -EINVAL response on open and it appears it only occurs as the result of the inability for the PS/2 mouse driver to grab IRQ 12 for the mouse. It could be that the emulator thinks your system is PC/XT instead of PX/AT, in which case sys_caps does not have IRQ8TO15 capability.

This will show in the "PC/XT class cpu YY" or "PC/AT class CPU YY" line at boot. I would like to know whether your system shows PC/XT, PC/AT and the class cpu YY value.

ghaerr:
@tyama501 I'm getting ahead of your test results, but looking at FabGL source it appears they emulate an 8086 CPU (https://github.com/fdivitto/FabGL/blob/master/src/emudevs/i8086.cpp) based on code from 8086tiny. They also emulate two 8259 PICs, enabling the to have interrupts 8-15. Normally, the PC/XT (8086-only) only has a single 8259 PIC, and ELKS assumes if the CPU is < 80286 (i.e. 8086) then the ability to use IRQs 8-15 is not possible. Meanwhile, FabGL also uses IRQ 12 within the 8042 emulation (https://github.com/fdivitto/FabGL/blob/master/src/emudevs/i8042.cpp) for the PS/2 mouse.

In order to make this work, we can either change ELKS to always allow IRQs 8-15 even when it finds an 8086 processor, *or* we could implement a syscaps= option in /bootopts, which would allow overriding the default sys_caps, which is 0 for 8086 and CAPS_ALL for 80286+.  What do you think?

Here are the system capabilities that now settable at compile-time only in config.h, which override the rule above:
```
/*
 * System capabilities - configurable for ROM or custom installations.
 * Normally, all capabilities will be set if arch_cpu >= CPU_80286 (PC/AT),
 * except when SYS_CAPS is defined for custom installations or emulations.
 */
#define CAP_PC_AT       (CAP_IRQ8TO15|CAP_IRQ2MAP9)      /* PC/AT capabilities */
#define CAP_KBD_LEDS    0x04    /* has keyboard LEDs */
#define CAP_HD_IDE      0x08    /* can do hard drive IDE probes */
#define CAP_IRQ8TO15    0x10    /* has IRQ 8 through 15 */
#define CAP_IRQ2MAP9    0x20    /* map IRQ 2 to 9 */
#define CAP_ALL         0xFF    /* all capabilities if PC/AT only */
```

ghaerr:
@tyama501, you should be able to test directly from the GitHub Actions downloads from the above PR, which aims to automatically fix this issue on FabGL and other emulators that implement a PS/2 mouse on IRQ 12 while emulating an 8086 CPU.

tyama501:
Thank you! I will test again late this week.

tyama501:
Hello @ghaerr 

The ps2 mouse on the FabGL worked with the latest https://github.com/ghaerr/elks/actions/runs/21502723483


https://github.com/user-attachments/assets/02c6fef1-6e3e-48ed-8066-b5ce3c6b525b

Thank you!

PR #2599, opened 2026-01-25
[kernel] Fix PS/2 mouse driver from testing on Bochs emulator

ghaerr:
This PR adds working PS/2 mouse support to the kernel driver first added in #2598 and discussed in https://github.com/ghaerr/microwindows/issues/109.

I was able to get the Bochs emulator running which supports PS/2 mice, which allowed for getting the driver working. Only tested on Bochs for now using `mouse`, not real hardware, copy.sh/v86 or Nano-X, but all should work.

### DIFF
--- a/elks/arch/i86/drivers/char/mouse-ps2.c
+++ b/elks/arch/i86/drivers/char/mouse-ps2.c
@@ -15,34 +15,72 @@
 
 #ifdef CONFIG_MOUSE_PS2
 
-#define IRQ_MOUSE       12      /* mouse interrupt, may conflict with networking */
+#define IRQ_MOUSE       12          /* mouse interrupt, may conflict with networking */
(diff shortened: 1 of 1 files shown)


toncho11:
It proves we are on the right path already :)

toncho11:
@ghaerr @rafael2k @tyama501 

I just tested quickly by setting `export MOUSE_TYE=ps2` on the last build and everything worked on [copy.86](https://copy.sh/v86)!!!
Like nxdsktop, paint and mouse. I was expecting more testing would be needed. 

One thing that does not work is when you exit the application. It seems the keyboard is blocked. You get to the prompt, but you can not type.

Using nxdsktop and paint on [copy.86](https://copy.sh/v86) was a real joy! It was very fluent ... and fast. :) Just the way it should be.

PR #2602, opened 2026-01-27
Improved PS2 code

toncho11:
Keyboard lock on application exit seems fixed.

PS/2 handling safe by splitting status bits, draining any pending controller output, and enabling/disabling only the mouse IRQ via read-modify-write - avoids keyboard lock.

### DIFF
--- a/elks/arch/i86/drivers/char/mouse-ps2.c
+++ b/elks/arch/i86/drivers/char/mouse-ps2.c
@@ -26,10 +26,12 @@
 #define COMMAND         0x64
 
 /* 8042 commands using outb() */
-#define WRITE_CTRLR     0x60        /* send command to controller */
-#define WRITE_MOUSE     0xd4        /* send command to aux device (mouse) */
-#define DISABLE_AUX     0xa7        /* disable aux (mouse) port */
-#define ENABLE_AUX      0xa8        /* enable aux (mouse) port */
+#define WRITE_CTRLR      0x60        /* send command to controller */
+#define WRITE_MOUSE      0xd4        /* send command to aux device (mouse) */
+#define DISABLE_AUX      0xa7        /* disable aux (mouse) port */
+#define ENABLE_AUX       0xa8        /* enable aux (mouse) port */
+#define GET_CMD_BYTE     0x20        /* read command byte */
+#define ENABLE_MOUSE_IRQ 0x02        /* enable mouse irq 12 */
 
 /* commands using WRITE_CTRLR/write_controller() */
 #define ENABLE_INTS     0x47        /* enable interrupts */
@@ -63,20 +65,20 @@ static void poll_aux_status(void)
     }
 }
 
-static unsigned char read_ccb(void)
+static unsigned int read_ccb(void)
 {
     poll_aux_status();
-    outb_p(0x20, COMMAND);          /* read command byte */
+    outb_p(GET_CMD_BYTE, COMMAND);          /* read command byte */
     poll_aux_status();
-    return inb_p(DATA);
+    return (unsigned int)inb_p(DATA);
 }
 
-static void write_ccb(unsigned char ccb)
+static void write_ccb(unsigned int ccb)
 {
     poll_aux_status();
     outb_p(WRITE_CTRLR, COMMAND);   /* 0x60 */
     poll_aux_status();
-    outb_p(ccb, DATA);
+    outb_p((unsigned char)ccb, DATA);
 }
 
 /* write command to mouse */
@@ -99,27 +101,27 @@ static void write_controller(int cmd)
 
 static void ps2_mouse_disable(void)
 {
-    unsigned char ccb;
+    unsigned int ccb;
 
-    write_mouse(DISABLE_AUX_DEV);       /* stop mouse packets */
+    write_mouse(DISABLE_AUX_DEV);        /* stop mouse packets */
 
     ccb = read_ccb();
-    ccb &= (unsigned char)~0x02;        /* clear IRQ12 enable ONLY */
+    ccb &= ~ENABLE_MOUSE_IRQ;            /* clear IRQ12 enable ONLY */
     write_ccb(ccb);
 
     poll_aux_status();
-    outb_p(DISABLE_AUX, COMMAND);
+    outb_p(DISABLE_AUX, COMMAND);        /* disable aux (mouse) port */
     poll_aux_status();
 }
 
 static void ps2_mouse_enable(void)
 {
-    unsigned char ccb;
+    unsigned int ccb;
 
     poll_aux_status();
-    outb_p(ENABLE_AUX, COMMAND);
+    outb_p(ENABLE_AUX, COMMAND);        /* enable aux (mouse) port */
 
-    write_mouse(ENABLE_AUX_DEV);
+    write_mouse(ENABLE_AUX_DEV);        /* enable mouse (aux device) */
 
     ccb = read_ccb();
     ccb |= 0x02;                        /* set IRQ12 enable ONLY */
@@ -136,9 +138,9 @@ static void ps2_irq(int irq, struct pt_regs *regs)
 
     /* Read all available mouse bytes */
     while ((inb(STATUS) & (OBUF_FULL | AUXDATA)) == (OBUF_FULL | AUXDATA)) {
-		c = inb(DATA);
-		chq_addch_nowakeup(q, c);
-	}
+        c = inb(DATA);
+        chq_addch_nowakeup(q, c);
+    }
 
     if (q->len)
         wake_up(&q->wait);
--- a/elks/arch/i86/drivers/char/mouse-ps2.c
+++ b/elks/arch/i86/drivers/char/mouse-ps2.c
@@ -1,7 +1,8 @@
 /*
- * Minimal PS/2 mouse driver for ELKS
+ * PS/2 mouse driver for ELKS
  *
  * 23 Jan 2026 Original version by Anton Andreev, adapted by Greg Haerr
+ * No support for extensions IntelliMouse PS/2 (ImPS/2) and IntelliMouse Explorer PS/2 (IMEX / Explorer). 
  */
 
 #include <linuxmt/config.h>
@@ -40,7 +41,7 @@
 
 /* controller status bits */
 #define IBUF_FULL       0x02        /* input buffer to device full */
-#define OBF             0x01        /* output buffer full */
+#define OBUF_FULL       0x01        /* output buffer full */
 #define AUXDATA         0x20        /* mouse data */
 
 static void poll_aux_status(void)
@@ -51,7 +52,7 @@ static void poll_aux_status(void)
         unsigned char st = inb_p(STATUS);
 
         /* Drain any pending output byte (kbd or mouse) */
-        if (st & OBF) {
+        if (st & OBUF_FULL) {
 			(void)inb_p(DATA);
 			continue;
 		}
@@ -134,7 +135,7 @@ static void ps2_irq(int irq, struct pt_regs *regs)
     unsigned char c;
 
     /* Read all available mouse bytes */
-    while ((inb(STATUS) & (OBF | AUXDATA)) == (OBF | AUXDATA)) {
+    while ((inb(STATUS) & (OBUF_FULL | AUXDATA)) == (OBUF_FULL | AUXDATA)) {
 		c = inb(DATA);
 		chq_addch_nowakeup(q, c);
 	}
--- a/elks/arch/i86/drivers/char/mouse-ps2.c
+++ b/elks/arch/i86/drivers/char/mouse-ps2.c
@@ -1,7 +1,7 @@
 /*
  * Minimal PS/2 mouse driver for ELKS
  *
- * 23 Jan 2026 Original version by @toncho11, adapted by Greg Haerr
+ * 23 Jan 2026 Original version by Anton Andreev, adapted by Greg Haerr
  */
 
 #include <linuxmt/config.h>
@@ -40,17 +40,42 @@
 
 /* controller status bits */
 #define IBUF_FULL       0x02        /* input buffer to device full */
-#define OBUF_FULL       0x21        /* aux (mouse) output and output buffer full */
+#define OBF             0x01        /* output buffer full */
+#define AUXDATA         0x20        /* mouse data */
 
 static void poll_aux_status(void)
 {
-    int retries = 0;
+    int retries = MAX_RETRIES;
 
-    while ((inb(STATUS) & 0x03) && retries < MAX_RETRIES) {
-        if ((inb_p(STATUS) & OBUF_FULL) == OBUF_FULL)
-            inb_p(DATA);
-        retries++;
+    while (retries--) {
+        unsigned char st = inb_p(STATUS);
+
+        /* Drain any pending output byte (kbd or mouse) */
+        if (st & OBF) {
+			(void)inb_p(DATA);
+			continue;
+		}
+
+        /* Input buffer clear and no output pending -> ready */
+        if (!(st & IBUF_FULL))
+            return;
     }
+}
+
+static unsigned char read_ccb(void)
+{
+    poll_aux_status();
+    outb_p(0x20, COMMAND);          /* read command byte */
+    poll_aux_status();
+    return inb_p(DATA);
+}
+
+static void write_ccb(unsigned char ccb)
+{
+    poll_aux_status();
+    outb_p(WRITE_CTRLR, COMMAND);   /* 0x60 */
+    poll_aux_status();
+    outb_p(ccb, DATA);
 }
 
 /* write command to mouse */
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @toncho11,

It seems the reason for the keyboard lock is that the 8042 command byte contents can't be assumed from the bit definitions, but should be read then set instead? Here's the flag values I used to get this working originally (from FabGL), which where then combined into DISABLE_INTS and ENABLE_INTS (which is used in the Linux 2.0 PS/2 mouse driver but apparently doesn't work on copy.sh/v86):
```
// Command byte bits
#define CMDBYTE_ENABLE_KEYBOARD_IRQ           0x01  // 0 : 1 = Keyboard output buffer full causes interrupt (IRQ 1)
#define CMDBYTE_ENABLE_MOUSE_IRQ              0x02  // 1 : 1 = Mouse output buffer full causes interrupt (IRQ 12)
#define CMDBYTE_SYSFLAG                       0x04  // 2 : 1 = System flag after successful controller self-test
#define CMDBYTE_UNUSED1                       0x08  // 3 : unused (must be 0)
#define CMDBYTE_DISABLE_KEYBOARD              0x10  // 4 : 1 = Disable keyboard by forcing the keyboard clock low
#define CMDBYTE_DISABLE_MOUSE                 0x20  // 5 : 1 = Disable mouse by forcing the mouse serial clock line low
#define CMDBYTE_STD_SCAN_CONVERSION           0x40  // 6 : 1 = Standard Scan conversion
#define CMDBYTE_UNUSED2                       0x80  // 7 : unused (must be 0)
```

Glad the driver is working, and that you've found a way to stop disabling the keyboard, even if we don't fully understand why. 

There are a number of other minor problems with this PR (back to using hex constants, removed comments, excessive code, GCC bugs) that I will comment on separately.

Thank you!

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -40,19 +41,44 @@
...
+        /* Input buffer clear and no output pending -> ready */
+        if (!(st & IBUF_FULL))
+            return;
     }
 }
 
+static unsigned char read_ccb(void)
+{
+    poll_aux_status();
+    outb_p(0x20, COMMAND);          /* read command byte */
```
ghaerr: Let's not use reintroduce hex constants. Use GET_CMD_BYTE and define above in the 8042 commands  section.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -73,18 +99,32 @@ static void write_controller(int cmd)
 
 static void ps2_mouse_disable(void)
 {
-    write_controller(DISABLE_INTS);     /* disable controller interrupts */
+    unsigned char ccb;
+
+    write_mouse(DISABLE_AUX_DEV);       /* stop mouse packets */
+
+    ccb = read_ccb();
+    ccb &= (unsigned char)~0x02;        /* clear IRQ12 enable ONLY */
```
ghaerr: Same here. Use ENABLE_MOUSE_IRQ in a new command byte bits section. An (unsigned char) cast is unnecessary and should be removed.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -73,18 +99,32 @@ static void write_controller(int cmd)
...
 {
+    unsigned char ccb;
+
     poll_aux_status();
-    outb_p(ENABLE_AUX, COMMAND);        /* enable aux (mouse) port */
-    write_mouse(ENABLE_AUX_DEV);        /* enable mouse (aux device) */
-    write_controller(ENABLE_INTS);      /* enable controller interrupts */
+    outb_p(ENABLE_AUX, COMMAND);
+
+    write_mouse(ENABLE_AUX_DEV);
```
ghaerr: Why were the "enable aux (mouse) port" and  "enable mouse (aux device)" comments removed? Please add back, as well as line 111 above. I spent quite a bit of time researching exactly what these are supposed to be doing.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -40,19 +41,44 @@
...
+        /* Drain any pending output byte (kbd or mouse) */
+        if (st & OBUF_FULL) {
+			(void)inb_p(DATA);
+			continue;
+		}
+
+        /* Input buffer clear and no output pending -> ready */
+        if (!(st & IBUF_FULL))
+            return;
     }
```
ghaerr, line 63: Why were lines 49-52 rewritten? Substituting "OBUF_FULL" in the original code with "OBUF_FULL | AUXDATA" above does the same thing with a lot less code.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -40,19 +41,44 @@
...
+static unsigned char read_ccb(void)
+{
+    poll_aux_status();
+    outb_p(0x20, COMMAND);          /* read command byte */
+    poll_aux_status();
+    return inb_p(DATA);
+}
+
+static void write_ccb(unsigned char ccb)
+{
```
ghaerr, line 75: Please rewrite using parameter declared as "unsigned int ccb". As discussed elsewhere, ia16-elf-gcc has bugs calling routines with unsigned char parameters. Using unsigned int avoids that and actually generates better code.

I recommend the same thing on the return type for read_ccb above: use unsigned int. This will avoid bugs if read_ccb is called as a function parameter and also generates better code.

The reason for the code improvement is that many 8086 instructions don't have byte get/set equivalents, and extra code has to be generated to move a byte into a normal "full-width" 16-bit registers, which is then typically pushed using PUSHW when calling functions.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -95,10 +135,10 @@ static void ps2_irq(int irq, struct pt_regs *regs)
...
 
     /* Read all available mouse bytes */
-    while ((inb(STATUS) & OBUF_FULL) == OBUF_FULL) {
-        c = inb(DATA);
-        chq_addch_nowakeup(q, c);
-    }
+    while ((inb(STATUS) & (OBUF_FULL | AUXDATA)) == (OBUF_FULL | AUXDATA)) {
+		c = inb(DATA);
+		chq_addch_nowakeup(q, c);
+	}
```
ghaerr: Lines 139-141 show no change: are you using TABS? Just checking, the file should contain no TAB characters, we're now using spaces due to the problem of 4 vs 8 char tab settings on editors.

toncho11: Done.

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: Done.

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: Done.

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: Done.

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: Because OBUF_FULL was doing two different jobs at once, and that was potentially unsafe. In the original code, OBUF_FULL mixed: “output buffer full” (bit 0), and “this byte is from the mouse” (bit 5), so it was used for both during controller setup (where any pending byte must be drained) and in the IRQ handler (where only mouse bytes must be consumed). Rewriting these lines separated those responsibilities.

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: Done.

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: Now it is spaces.

toncho11:
Correct read-modify-write CCB handling means reading the existing 8042 command byte and changing only the required bit (e.x. mouse IRQ enable).  Writing fixed values (the whole byte) can accidentally disable the keyboard.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -25,10 +26,12 @@
...
-#define WRITE_CTRLR     0x60        /* send command to controller */
-#define WRITE_MOUSE     0xd4        /* send command to aux device (mouse) */
-#define DISABLE_AUX     0xa7        /* disable aux (mouse) port */
-#define ENABLE_AUX      0xa8        /* enable aux (mouse) port */
+#define WRITE_CTRLR      0x60        /* send command to controller */
+#define WRITE_MOUSE      0xd4        /* send command to aux device (mouse) */
+#define DISABLE_AUX      0xa7        /* disable aux (mouse) port */
+#define ENABLE_AUX       0xa8        /* enable aux (mouse) port */
+#define GET_CMD_BYTE     0x20        /* read command byte */
+#define ENABLE_MOUSE_IRQ 0x02        /* enable mouse irq 12 */
```
ghaerr: As indicated prior, please place ENABLE_MOUSE_IRQ under a new section called "command byte status bits". The above list is 8042 commands only. Also put GET_CMD_BYTE in sorted order like the rest are, and reset changed hex columns and comments to proper tab stops.

elks/arch/i86/drivers/char/mouse-ps2.c
ghaerr: Ok, I see - my bad. I was thinking you wanted to wait until both OBUF_FULL and AUXDATA were 0. ~Instead, you're saying you want to wait until both OBUF_FULL and IBUF_FULL are 0 (no input or output), right?~

~In that case, the routine would be much smaller and faster with just the second inb_p in the original removed:~
```
   /* drain any pending input or output */
   while ((inb(STATUS) & (OBUF_FULL|IBUF_FULL)) && retries < MAX_RETRIES)  
        retries++;
```
Sorry (again)! I was having trouble understanding this routine. Keep it as you have rewritten it, thank you.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -73,18 +101,32 @@ static void write_controller(int cmd)
...
+    unsigned int ccb;
+
     poll_aux_status();
     outb_p(ENABLE_AUX, COMMAND);        /* enable aux (mouse) port */
+
     write_mouse(ENABLE_AUX_DEV);        /* enable mouse (aux device) */
-    write_controller(ENABLE_INTS);      /* enable controller interrupts */
+
+    ccb = read_ccb();
+    ccb |= 0x02;                        /* set IRQ12 enable ONLY */
```
ghaerr: Please don't use hex constants - replace 0x02 with ENABLE_MOUSE_IRQ.

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -73,18 +101,32 @@ static void write_controller(int cmd)
...
+
+    write_mouse(DISABLE_AUX_DEV);        /* stop mouse packets */
+
+    ccb = read_ccb();
+    ccb &= ~ENABLE_MOUSE_IRQ;            /* clear IRQ12 enable ONLY */
+    write_ccb(ccb);
+
     poll_aux_status();
-    outb_p(DISABLE_AUX, COMMAND);       /* disable aux (mouse) port */
+    outb_p(DISABLE_AUX, COMMAND);        /* disable aux (mouse) port */
```
ghaerr, line 102: Can you double-check that your editor is tab-stopping at 8 spaces? It appears, unless I'm mistaken, that 104-113 now have comments at 9 rather than 8. All comments should start at the same column for all functions. Sorry for the nit-pick, perhaps this is just GitHub showing things incorrectly?

elks/arch/i86/drivers/char/mouse-ps2.c
```diff
@@ -40,19 +43,44 @@
...
+    poll_aux_status();
+    return (unsigned int)inb_p(DATA);
+}
+
+static void write_ccb(unsigned int ccb)
+{
+    poll_aux_status();
+    outb_p(WRITE_CTRLR, COMMAND);   /* 0x60 */
+    poll_aux_status();
+    outb_p((unsigned char)ccb, DATA);
```
ghaerr: Please remove the (unsigned char) cast above, as well as on line 73. Neither are needed, both `inb` and `outb` macros only read/write 8 bits (i.e. unsigned char) and don't need casts. If they were improperly written and returned signed char, the proper fix would be to the macro and never to all the C code calling them. And even if on line 73 `inb` didn't return 8 bits, casting to unsigned int does nothing. The compiler knows the return value is an unsigned int, since the function is declared as returning unsigned int. 

A good C casting rule is: *never use casts unless they are absolutely necessary.*

ghaerr:
> Correct read-modify-write CCB handling means reading the existing 8042 command byte and changing only the required bit (e.x. mouse IRQ enable). 

Yes, I get that, but programming the 8042 seems very complicated since there are three different destinations to send commands to:
- 8042 controller itself (labelled "8042 commands" in the source. These are "commands" sent directly to the COMMAND port of the 8042.
- bytes sent to set the "command byte" of the controller. This is done by `write_controller` as well as seemingly duplicated exactly with a new routine called `write_ccb`. (Please remove `write_controller` since it is now unused, to avoid duplicated code and more confusion).
- bytes sent to the mouse "controller". This is done by `write_mouse`.

So you can see, given that this PR duplicates `write_controller` with `write_ccb` and nobody even noticed,  things are very confusing. I have been trying hard to figure out exactly how the system works, especially since we're dealing with testing only on an emulator and not real hardware - so we don't really know if what we're writing works or not.

> Writing fixed values (the whole byte) can accidentally disable the keyboard.

Well - yes, agreed. But looking at the "command byte" bit definitions and the previous values of DISABLE_INTS (0x47) and ENABLE_INTS (0x65), bits 0 & 1 and 4 & 5 are properly set regarding disable/enable keyboard or mouse:
```
// Command byte bits
#define CMDBYTE_ENABLE_KEYBOARD_IRQ           0x01  // 0 : 1 = Keyboard output buffer full causes interrupt (IRQ 1)
#define CMDBYTE_ENABLE_MOUSE_IRQ              0x02  // 1 : 1 = Mouse output buffer full causes interrupt (IRQ 12)
#define CMDBYTE_SYSFLAG                       0x04  // 2 : 1 = System flag after successful controller self-test
#define CMDBYTE_UNUSED1                       0x08  // 3 : unused (must be 0)
#define CMDBYTE_DISABLE_KEYBOARD              0x10  // 4 : 1 = Disable keyboard by forcing the keyboard clock low
#define CMDBYTE_DISABLE_MOUSE                 0x20  // 5 : 1 = Disable mouse by forcing the mouse serial clock line low
#define CMDBYTE_STD_SCAN_CONVERSION           0x40  // 6 : 1 = Standard Scan conversion
#define CMDBYTE_UNUSED2                       0x80  // 7 : unused (must be 0)
```
Thus, while I agree it is definitely a better design to just read the command byte then set or reset a particular bit only, I remain unsatisfied that we don't really know why this worked. Perhaps we need to printk the value of read_ccb() so we know and understand what's going on. The ENABLE_INTS/DISABLE_INTS code came directly from Linux 2.0. Perhaps the command byte bits (documented above?) have changed since those days? Not sure, I don't have any definitive documentation on the 8042 or its ROM programming. I did find this [8042 ROM disassembly](https://github.com/Halicery/8042) but haven't had time to study it yet. I suspect this is part of the difference between real hardware, various emulators, and the various reference PS/2 mouse drivers out there.

elks/arch/i86/drivers/char/mouse-ps2.c
ghaerr: It appears that when you wrote `write_ccb` that this routine, `write_controller` was already present and contained exactly the same statements. Since write_controller is no longer used, since DISABLE_INTS and ENABLE_INTS are no longer used, this function should be deleted. 

The compiler should have given a warning about an unused static function - did it?

toncho11:
I do not have time for this.

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: I could not understand about the sorting and the tabs stops!

elks/arch/i86/drivers/char/mouse-ps2.c
toncho11: Same, I do not have tab stopping editor at 8 spaces.

ghaerr:
No worries @toncho11. I have been debugging on copy.sh/v86 and have found the actual problem for the keyboard not working after the mouse driver close. More to be reported later today.

toncho11:
So should I complete the code above?

ghaerr:
> So should I complete the code above?

No, not yet. I think I have found a separate solution. The problem had nothing to do with setting/resetting the mouse IRQ bits individually.

toncho11:
@ghaerr I envy you how you always have the time to read all the documentation and even look in [8042 ROM disassembly](https://github.com/Halicery/8042). For me it is just not possible to devote so much time with my work and everything. Also please note that as I do not regularly submit PRs it is not possible for me to know the tabs, spaces, gcc quirks etc. Maybe next time you can do these inside the PR directly. Anyway, that was my attempt to help with the driver.

toncho11:
If your version is better, then you can close this PR.

toncho11:
I came to the conclusion that ELKS is not a project where I can contribute to the level I would like to.

ghaerr:
@toncho11,

After your running out of time to work on this, I had started with PR #2605 to further probe my concerns about exactly how the driver started working. This is explained in more detail in #2605. That research caused me to work on yet more routine renaming to best clarify how the driver actually works. So I will commit #2605 and close this PR. I have included your routines that allowed the driver to properly function, thank you.

> For me it is just not possible to devote so much time with my work and everything. Also please note that as I do not regularly submit PRs it is not possible for me to know the tabs, spaces, gcc quirks etc. Maybe next time you can do these inside the PR directly. Anyway, that was my attempt to help with the driver.

I feel I probably over-nitpicked the formatting and some quirks in this PR. For that, I apologize. I had just prior spent quite a lot of time renaming routines and defines, and reformatting the driver, mostly to better understand myself the workings of the 8042. I appreciate your work getting the PS/2 driver working on copy.sh/v86. I probably should have just pushed some fixes directly into the PR, but I don't like to do that without asking, which I will do next time. The good news is that we now have what seems to be a fully working PS/2 mouse driver for ELKS, thanks to your initial and continuing efforts. I had hoped to stay out of the driver initially, but then got pulled in and it did take a while to finally get my head around how the 8042 actually operates - I have learned a lot.

Thank you!

PR #2607, opened 2026-02-02
[kernel] Rewrite C fast serial driver top half in ASM for speed and size

ghaerr:
Follow-on to #2604. The four almost-duplicated C fast serial top-half interrupt handlers are combined and rewritten in ASM in serfast.S. 

Inspired by trying to combine the C handlers in https://github.com/Mellvik/TLVC/pull/225. While originally having most of the serial interrupt handler written in C was thought to be a good idea, having 4 near-duplicate routines because of GCC compiler limitations finally tipped the scale in favor of just rewriting a single routine in assembly to handle all four com ports.

The serial_info, tty and ch_queue structs were modified to place the ASM-accessed members at the top for code size and (relatively) known offsets.

Saves ~224 bytes kernel code space and should be marginally faster than previous C code.

Tested on QEMU.

### DIFF
--- a/elks/arch/i86/drivers/char/serfast.S
+++ b/elks/arch/i86/drivers/char/serfast.S
@@ -1,114 +1,130 @@
-// First part of top half of fast serial interrupt handlers for ELKS
+// Top half of fast serial interrupt handlers for ELKS
 //
 // Runs on any stack and skips ELKS _irqit stack switching overhead.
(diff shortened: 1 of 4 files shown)


Mellvik:
Great @ghaerr - I was kind of hoping to trigger just that. Very compact. I'll port this back to TLVC now and test on real hardware. Should be good. Thanks.

PR #2609, opened 2026-02-05
[ExtApp] Add ELKSmoria Roguelike game

tyama501:
Hello @ghaerr 

I have added ELKSmoria as ExtApp.

I have confirmed the buildext.sh and making minix hdd IBM image work.

Thank you!

### DIFF
--- a/buildext.sh
+++ b/buildext.sh
@@ -23,6 +23,7 @@
 #       lua             OpenWatcom      Lua 5.5 interpreter
 #       bobcat          OpenWatcom      Bobcat web browser (Lynx fork)
 #       kilomacs        OpenWatcom      Kilo-based editor with Emacs-style keybindings
+#       elksmoria       OpenWatcom      Roguelike game
 #
 # Some projects may require prerequisites.
 # To only build the C86 toolchain, use './buildext.sh owc_libc c86_toolchain'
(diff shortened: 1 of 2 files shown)


ghaerr:
Hello @tyama501,

I was thinking it might be a better idea to install this new game as just `moria`, rather then elksmori.os2 or elksmori. This will be easier for users to understand what it is. I don't think 'elks' needs to be in the name, since everyone knows they're on ELKS.

If that sounds OK, can you also make changes so that the extra moria files are placed in /lib/moria, rather than /lib/morifiles? That will also be more apparent to users what the files are. Actually, I think it might even be better to have the files live directly under /lib instead of a subdirectory, but I don't know the best names for the files. Something like /lib/moria.\* would then keep the disk hierarchy flatter but also understandable. These are all ideas to try to keep things simple.

Thank you!

tyama501:
Hello @ghaerr 

Maybe I can do that but I need to modify source code to change lib filename and locations.

It is easy to change the directory name so I can change it to lib/moria but can I keep the lib filenames?

Thank you.

ghaerr:
> It is easy to change the directory name so I can change it to lib/moria but can I keep the lib filenames?

Ok, do that for now. If you change the source code, ExtApplications can be updated afterwards.

tyama501:
I have fixed that.

ghaerr:
Thanks @tyama501. In retrospect, now that I look at /lib, it's probably better to use a directory under /lib for applications that require more than one library file, like you're doing. I notice that `paint` has lots of .bmp files in /lib, I'll move them to /lib/paint.

Issue #2552, opened 2025-12-24
MACOS build: configure: error: libmpfr not found or uses a different ABI (including static vs shared).


anchorz:
./build.sh ends up with the error message. I have no idea how to fix it. normally the library will be installed as part of the build step. Is there a chance I can get some help?

sorry, i cant delete this issue anymore. but there are several issues, that blocks me from compiling the toolchain. after deleting everthing and start from clean build, the error message is different:
In file included from ../../gcc-src/zlib/zutil.c:10:
In file included from ../../gcc-src/zlib/gzguts.h:21:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/stdio.h:61:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/_stdio.h:318:7: error: expected ')'
  318 | FILE    *fdopen(int, const char *) __DARWIN_ALIAS_STARTING(__MAC_10_6, __IPHONE_2_0, __DARWIN_ALIAS(fdopen));
      |          ^
../../gcc-src/zlib/zutil.h:128:33: note: expanded from macro 'fdopen'
  128 | #        define fdopen(fd,mode) NULL /* No fdopen() */

3 warnings and 3 errors generated.

i cant get the tool chain compiled.

anchorz:
i tried gcc-ia16 as well:
*** Configuration arm-apple-darwin24.5.0 not supported

ghaerr:
Hello @anchorz,

I'm sorry about the troubles you're having. I use macOS myself but x86 not arm. 

> i tried gcc-ia16 as well:

You mean from a different repo? Which one and what exactly did you try? The purpose of ./build.sh is pretty much to compile gcc-elf-ia16. We don't actually maintain ia16-elf-gcc here, but instead use build.sh to avoid having to use an external build for the compiler.


I see earlier you got an error libmpfr not found, but above in your edited post the compilation is failing in zlib with some header.
It would seem the issue is that various gcc sources are not compatible with macOS ARM header files, but I can't be sure unless I can see the entire build log - can you post that?

[...]

PR #2616, opened 2026-02-12
[build] Fix build problems on macOS ARM machines

ghaerr:
This PR attempts to fix all previously-reported problems building ELKS, the ia16-elf-gcc cross-compiler, and mtools-4.0.23 on macOS ARM-based systems.

Inspired by @Mellvik's findings and fixes in https://github.com/Mellvik/TLVC/pull/227, with more issues reported in #2552.

Also updates gcc-ia16 and binutils-ia16 to the latest Github versions, suggested by @asiekierka in #2309.

It is not necessary for developer to update the toolchain, although some small fixes are included, detailed below. To upgrade the toolchain, run `git pull; tools/build.sh`; this will take ~9 minutes on macOS ARM and ~20 minutes on most x86 systems.

The following changes are included in this PR:
- Fix macOS toolchain build failures in gcc/zlib, binutils/zlib, gcc/host_hooks, gcc fibonacci C++ and C locale issues
- Fix macOS mtools-4.0.23 build within ELKS
- Update gcc-ia16 and binutils-ia16 to latest versions
- Default don't build external apps in ./build.sh for users
- Add better usage instructions in build.sh
- Clarify error message when `make` run without TOPDIR= set
- Update qemu.sh for QEMU v10
- Some minor non-ELKS-specific bug fixes in gcc 
- Some improvements to instruction cost analysis for code generation by @asiekierka


ghaerr:
@asiekierka: BTW, after upgrading gcc-ia16 and binutils-ia16, many, but not all, of the compiled applications are exactly 16 bytes larger. The kernel .text segment is 32 bytes larger. I'm not sure whether this might have to do with instruction optimizations behaving differently with -Os, disabling the IRA SPA optimizer (apparently only used in far segments), or something to do with binutils changes. The 16 bytes could point to a linker tool, but the only two binutils changes are GAS "--segelf-32" fixes which wouldn't see to have any effect. Of course, with more instructions emitted, the .text rounding would increase to the next 16-byte boundary.

I'm not sure it's worth tracking down, but I found it interesting. Do you have any thoughts on this?

PR #2631, opened 2026-03-16
[NECV25] Add ISA wait state handling and a minor change to tools/mfs

swausd:
1. To use NECV25 on the ISA bus handling of ISA-Signal IOCHRDY has been added.

2. Minor change to mfs: A line feed was added to the usage message


ghaerr:
Thanks @swausd and nice catch on the mfs usage line!

PR #2645, opened 2026-04-19
Fix error to enable set_a20 and reset_a20 when USE_HIMEM_AT is configured

swausd:
With `xor %ah, %ah` only reset_a20 can be selected.


ghaerr:
Hello @swausd, nice catch! Glad this fixed your PS/2 keyboard problem!

PR #2648, opened 2026-04-23
[net] Use MIME type `application/octet-stream` for unknown file types in httpd

SuperMaxusa:
According to [RFC 1945 (section 7.2.1)](https://datatracker.ietf.org/doc/html/rfc1945#section-7.2.1), `application/octet-stream` is used when the data type is unknown (or as a generic MIME type).


ghaerr:
Thanks @SuperMaxusa!

PR #2653, opened 2026-04-26
[kernel] Fix sys_wait4 not returning -ECHILD immediately in certain cases

ghaerr:
Since it was recently found that the IBM PC ELKS v0.9.0 images contain the wrong binary for Nano-X, a version increment to v0.9.1 will have to be made soon and new images produced. Since another wait-oriented kernel reparenting bug was also fixed recently in #2639, I thought to include a few other important kernel fixes recently found by @Vutshi's exploration into AI tools.

This PR fixes the `wait4` syscall issue found by @Vutshi's ChatGPT Plus and Claude kernel audit in #2646. The primary issue was that `wait4` would wait rather than return -ECHILD when a process ID passed that was not a child of the calling process. 

This fix includes my modified code from Claude's proposed fix, as it was much less messy than that proposed by ChatGPT Plus.

Claude also pointed out another issue which is that in the case of the parent only having stopped or zombie children which do not match the passed PID, an extra sleep/wakeup cycle is performed. This inefficiency will be handled at a later time, and there is no regression test yet built for it.

The regression tests used to verify correctness, provided by ChatGPT and Claude, are also add here in elkscmd/test/syscall/. These are not yet compiled and distributed on images, but will be after more regression tests are seen. Both tests initially showed failure and now show success.

### DIFF
--- a/elks/kernel/exit.c
+++ b/elks/kernel/exit.c
@@ -45,9 +45,25 @@ static void FARPROC reparent_children(void)
 int sys_wait4(pid_t pid, int *status, int options, void *usage)
 {
     register struct task_struct *p;
-    int waitagain;
+    int waitagain, has_child;
 
     debug_wait("WAIT(%P) for %d opts %x\n", pid, options);
+
+    /*
+     * If a specific pid is requested, verify it is actually our child first,
+     * otherwise return -ECHILD immediately.
+     */
+    if ((int)pid > 0) {
+        has_child = 0;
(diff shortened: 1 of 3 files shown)


ghaerr:
Note to implementors: an additional PR #2668 contains further fixes required as a result of applying this PR. In particular, the oft-used `telnetd` and `ftpd` daemons will fail without the additional fixes in #2668.

PR #2657, opened 2026-04-26
[kernel] Fix sys_gettimeofday microsecond normalization

ghaerr:
Fixes typo in microsecond wrap from 10M to 1M. Found by ChatGPT Plus in #2646. Very low risk.

### DIFF
--- a/elks/kernel/time.c
+++ b/elks/kernel/time.c
@@ -110,8 +110,8 @@ int sys_gettimeofday(register struct timeval *tv, struct timezone *tz)
         now = jiffies();
         tmp_tv.tv_sec = xtime.tv_sec + (now - xtime_jiffies) / HZ;
         tmp_tv.tv_usec = xtime.tv_usec + ((now - xtime_jiffies) % HZ) * (1000000L / HZ);
-        if (tmp_tv.tv_usec >= 10000000L)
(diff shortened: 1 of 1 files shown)


Vutshi:
@ghaerr, why don't we need to increment seconds during the microsecond wrap?

ghaerr:
Well, this isn't the actual code that increments time - that happens with `jiffies` in the HW timer interrupt. This code is just normalizing tv_usec after subtracting out the start jiffies, stored in xtime_jiffies. 

Admittedly this is a bit tricky code, and I think my use of the word 'wrap' above is not correct. At this point, I'm only fixing the obvious error pointed out by ChatGPT of using 10M rather than 1M for the normalization. I think the way to think about what is happening here is that the computed values are correct, but the normalization is required in the case where the initial xtime_jiffies.tv_usec value happens to cause wrap when added to the already-subtracted `now - xtime_jiffies`. In other words, the `now - xtime_jiffies` is always the correct jiffies time value, but because of exactly what time it was when sys_settimeofday was called, xtime_jiffies.tv_usec may overflow when converting the time back to a timeval struct, which does not affect the already-calculated actual number of jiffies, only when converting back to tv_sec/tv_usec.

Issue #2664, opened 2026-04-30
ELKS Digger


Vutshi:
Here is a preview of the [Digger Remastered](https://www.digger.org/index.html) port for ELKS.

https://github.com/user-attachments/assets/4a0d5d1f-eae8-433c-9e7f-da5b4f8b8abe

A binary is attached below for brave testers:

[digger.zip](https://github.com/user-attachments/files/27245601/digger.zip)

## Input mapping

```text
Arrow keys       -> movement
WASD             -> movement
Space            -> fire
Ctrl-Space/NUL   -> fire fallback
Ctrl-A           -> fire fallback
p / P            -> pause
q / Q            -> quit
Esc              -> quit
Ctrl-C           -> quit in raw mode
n / N            -> preserved for title-screen player-count toggle
```

ghaerr:
This is great, you're on a roll @Vutshi! 

It'd be great to get this on the v0.9.1 distribution, if you're interested. That could be done with a PR from you into elkscmd/, or it could be built from your repo using buildext.sh.

Vutshi:
It still needs a few things to be fixed. Keys are not working properly, apparently input.c doesn’t detect long press of a key. There is no sound yet. 
The rest seems to work well, it is pretty fast already and Chatty promised me some asm for hot code to make it perfect.

Vutshi:
@ghaerr is there an example of a program with sound in ELKS?

ghaerr:
> is there an example of a program with sound in ELKS?

Sure is, very basic sounds is in elkscmd/beep.c. But far more interesting is [this PCFX code](https://github.com/FrenkelS/elksdoom/pull/5/changes) I found in elksdoom. To use the latter requires a front end waveform generator (Muse?) which I haven't been able to find.

Vutshi:
> Sure is, very basic sounds is in `elkscmd/beep.c`

It works! And it sounds wonderfully terrible :)

https://github.com/user-attachments/assets/79d07316-1d65-4e33-89db-9459ade2cebc

Vutshi:
Finally we have the first approximation of DOS sound. The tune is recognisable!

https://github.com/user-attachments/assets/741afebb-3e19-48ba-8a7e-30f5cf85bb69

Vutshi:
Now we have QEMU playable ELKS Digger. Previously it was too fast. Attached are ELKS image with `digger` in `/home` folder and the standalone binary.

[...]

PR #2672, opened 2026-05-04
[cmds] Add ELKS Digger game to buildext.sh

ghaerr:
Adds ELKS Digger game recently ported by @Vutshi in #2664 to distribution images, starting with 2880k floppy and larger images.

@Vutshi, I followed your repo instructions and added digtitle.bmp to /bin (next to /bin/digger) but I'm only seeing the score screen on startup. Is this correct?

It would be nice to have Digger search /lib for digtitle.bmp rather than (or in addition to) /bin, which keeps images out of /bin. If you make that change in your repo, all we have to do is modify ExtApplications of the new location using an added `::lib/digtitle.bmp` after the source filename.

### DIFF
--- a/buildext.sh
+++ b/buildext.sh
@@ -15,6 +15,7 @@
 #       dcc             ia6-elf-gcc     DCC self-compiling C compiler for ELKS
 #       dflat           ia16-elf-gcc    D-Flat TUI memopad/library
 #       elkirc          ia16-elf-gcc    IRC for ELKS
+#       elksdigger      ia16-elf-gcc    Digger for ELKS
 #       owc_libc        OpenWatcom      ELKS C Library compiled by OWC
 #       owc_elkscmd     OpenWatcom      Some elkscmd/ programs compiled by OWC
 #       c86_toolchain   OpenWatcom/C86  C86 Toolchain, header files and examples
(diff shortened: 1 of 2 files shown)


Vutshi:
> I'm only seeing the score screen on startup. Is this correct?

Yes, that is correct. The score screen is only shown on startup and after game over.

The BMP image is honestly a bit silly anyway — it is mostly just a red frame around the score table and the Windmill Software title. I removed it from the game binary because I was originally fighting to squeeze everything into the small memory model. The game runs fine without it; the image is really just there to preserve a bit of the authentic atmosphere for discerning players.

ghaerr:
I'm not seeing any red frame, not seeing any .bmp display, that was my question about whether that's supposed to be before the score screen. So the BMP display doesn't seem to be working. Do we have to have the .bmp in /bin? I would rather it be in /lib than in the distribution /bin directory. I can change this to /lib in the ExtApplications now but in either case it doesn't seem to display on my end.
Screenshot 2026-05-04 at 11 39 30 AM

Vutshi:
This is strange. I will investigate.

ghaerr:
> This is strange. I will investigate.

Ok. I will change ExtApplications to install digtitle.bmp into /lib and commit this now. When you make the changes to fix this, it will automatically update on the buildext.sh build. If you end up deleting digtitle.bmp, we can remove it from ExtApplications, thanks.

Vutshi:
Yes, do this. Digger works fine anyway and I will fix the bitmap issue later.

Vutshi:
@ghaerr, does speaker output work for you in QEMU?

ghaerr:
> does speaker output work for you in QEMU?

No, I've never gotten sound to work on QEMU.

Vutshi:
@ghaerr 
> I'm not seeing any red frame, not seeing any .bmp display

This should be fixed now. Digger has now absolute path to `/lib/digtitle.bmp`

PR #2674, opened 2026-05-05
Add experimental PC speaker sequencer

Vutshi:
This PR adds a small experimental PC speaker sequencer for ELKS. See discussion in #2664.

The existing `KIOCSOUND` interface can set a single PIT channel-2 tone, but music playback from user space requires repeated timing-sensitive calls. This patch adds a compact event sequencer: user programs queue `{ divisor, ticks, flags }` events, and the kernel advances them using the existing timer infrastructure.

Design goals:

- keep PIT timer 0 kernel-owned
- use PIT channel 2 for square-wave tone generation
- avoid PCM, PWM, mixing, or high-rate interrupt work
- use event-boundary scheduling rather than a 100 Hz playback loop
- keep the kernel state small and static
- preserve existing bell / `KIOCSOUND` behavior
- stop speaker output on stop/flush and process exit where practical

The sequencer is intended for simple monophonic music/SFX streams on 8086/186 systems. It is not a general audio subsystem.

A small test program, `pcspkpop`, is included. It queues a compact Popcorn melody through the new ioctl and can also be used as a basic regression test for the sequencer.

## Short usage instructions

```c
#include <fcntl.h>
#include <sys/ioctl.h>
#include <linuxmt/pcspk.h>

int fd;
struct pcspk_event evs[2];
struct pcspk_seq seq;

fd = open("/dev/console", O_WRONLY);

/* A4 for 50 ticks, then rest for 10 ticks */
evs[0].divisor = 2712;
evs[0].ticks = 50;
evs[0].flags = PCSPK_F_TONE;
evs[0].priority = 0;

evs[1].divisor = 0;
evs[1].ticks = 10;
evs[1].flags = PCSPK_F_REST;
evs[1].priority = 0;

seq.events = evs;
seq.count = 2;
seq.rate_hz = 0;
seq.flags = PCSPK_F_FLUSH;

ioctl(fd, KIOCSNDSEQ, &seq);
```

To stop playback:

```c
struct pcspk_seq seq;

seq.events = 0;
seq.count = 0;
seq.rate_hz = 0;
seq.flags = PCSPK_F_STOP | PCSPK_F_FLUSH;

ioctl(fd, KIOCSNDSEQ, &seq);
```

### DIFF
--- a/elks/arch/i86/drivers/char/Makefile
+++ b/elks/arch/i86/drivers/char/Makefile
@@ -129,7 +129,7 @@ ifdef CONFIG_CHAR_DEV_RS
 endif
 
 ifdef CONFIG_ARCH_IBMPC
-  OBJS += mouse-ps2.o
+  OBJS += mouse-ps2.o pcspkseq.o
 endif
 
 ifdef CONFIG_CHAR_DEV_CGATEXT
--- a/elks/arch/i86/drivers/char/bell-8254.c
+++ b/elks/arch/i86/drivers/char/bell-8254.c
@@ -8,6 +8,7 @@
  */
 
 #include <linuxmt/config.h>
+#include <linuxmt/pcspk.h>
 #include <arch/ports.h>
 #include <arch/io.h>
 #include <arch/irq.h>
@@ -20,6 +21,9 @@
  */
 void soundp(unsigned period)
 {
+#ifdef CONFIG_ARCH_IBMPC
+    pcspk_seq_stop();
+#endif
     clr_irq();
     outb(inb(SPEAKER_PORT) | 0x03, SPEAKER_PORT);
     outb(0xB6, TIMER_CMDS_PORT);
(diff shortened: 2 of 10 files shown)


ghaerr:
Thanks @Vutshi. My plan is to commit this then immediately start making changes, as previously discussed in #2664. I will post PR(s) so you can preview.

The changes will likely temporarily break your own version(s) of Digger, but should be easily updated. In particular, I'm thinking of changing all pcspk\_ entry points, variables, defines etc to audio\_. The entire feature will be enabled via a new config option something like CONFIG_AUDIO, and sys_utils/pcspkpop.c renamed and likely moved to elkscmd/test/other/test_audio.c. The aforementioned interrupt flags and stack issues will also be addressed. I'll try to stay out of the actual speaker output code and leave in the kernel data segment ring buffer for now.

Hopefully I'll be able to test with the your or @SuperMaxusa's QEMU command line suggestions for getting audio working. After that it can be thrown back into your hands for further enhancements to itself or Digger.

PR #2675, opened 2026-05-06
[kernel] Cleanup kernel PC speaker audio driver

ghaerr:
Changes made to PR #2674 contents to finalize initial support for kernel audio originally submitted by @Vutshi.

Changes include:
- Entire functionality enabled with CONFIG_AUDIO config setting, which is by enabled by default for IBM PC builds. If the system uses an 8254 programmable timer, this can likely be easily ported to non-IBM PC systems as well.
- Replace save_flags/clr_irq/restore_flags with clr_irq/set_irq as interrupts are always enabled on entry to kernel timer callbacks and ioctl syscall.
- Rename audio driver from pcspkseq.c to audio-8254.c.
- Move audio test program to elkscmd/test/other/test_audio.c.
- Rename PCSPK\_ and pcspk\_ defines and function names to AUDIO_/audio\_.
- Removed linuxmt/pcspk.h and moved contents to linuxmt/kd.h.

The total kernel code increase is now 736 bytes. Tested on QEMU with the following line added to `qemu.sh`:
```
AUDIO="-audiodev coreaudio,id=audio0 -machine pcspk-audiodev=audio0"
```

The test program seems to work well but I note that typing on the keyboard during audio output disrupts the output a bit. This will should be further looked into as it's unclear whether the keyboard interrupt is disrupting kernel timer callbacks, or whether other programs or kernel routines are being run instead. 

@Vutshi, hopefully it is fairly straightforward to make any needed changes to digger. The structure names and flag names are changed from PCSPKR/pcspkr to AUDIO/audio as described above but nothing else was changed from the API standpoint. Also, I noticed that test_audio uses the KIOCSOUND ioctl as a fallback and converts the audio sequences if the kernel doesn't support KIOCSNDSEQ, which is pretty cool! Please go ahead and upgrade your digger repo to use this functionality if it isn't already present, and the CI will automatically update the ELKS builds with it.


toncho11:
@ghaerr @Vutshi 

https://www.youtube.com/watch?v=RLAtZ_9ZMD4

Can you try this one: 

```
/* megalo.c - ELKS PC speaker sequencer: Megalovania main loop */

#include <sys/ioctl.h>
#include <errno.h>
#include <fcntl.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <linuxmt/kd.h>

/* PIT Divisors for Megalovania notes */
#define REST 0
#define D4   4063
#define C4   4560
#define B3   4831
#define BB3  5118
#define F4   3417
#define G4   3044
#define GS4  2873
#define A4   2712
#define D5   2031

struct note {
    unsigned short div;
    unsigned char dur;
};

static int seqfd = -1;
static volatile int stop_requested;

static void stop_sound(void) {
    struct audio_seq s;
    if (seqfd < 0) return;
    s.events = 0; s.count = 0; s.rate_hz = 0;
    s.flags = AUDIO_SEQ_F_STOP | AUDIO_SEQ_F_FLUSH;
    ioctl(seqfd, KIOCSNDSEQ, &s);
    ioctl(seqfd, KIOCSOUND, 0);
}

static void on_signal(int sig) { stop_requested = 1; }

static unsigned short tempo_ticks(unsigned char dur, int tempo) {
    unsigned long v = (unsigned long)dur * 3000UL;
    v = (v + (unsigned)tempo / 2) / (unsigned)tempo;
    return (v < 1) ? 1 : (unsigned short)v;
}

static int build_events(struct audio_event *out, struct note *in, int nnotes, int tempo) {
    int i, n = 0;
    unsigned short t, gap;
    for (i = 0; i < nnotes; i++) {
        t = tempo_ticks(in[i].dur, tempo);
        gap = (t > 2) ? 1 : 0; /* Tight gap for fast tempo */
        if (in[i].div) {
            out[n].divisor = in[i].div;
            out[n].ticks = t - gap;
            out[n].flags = AUDIO_F_TONE;
            n++;
        }
        if (gap || !in[i].div) {
            out[n].divisor = 0;
            out[n].ticks = gap ? gap : t;
            out[n].flags = AUDIO_F_REST;
            n++;
        }
    }
    return n;
}

int main(int argc, char **argv) {
    /* The famous Megalovania opening riff */
    static struct note melody[] = {
        /* Bar 1 (Root D) */
        {D4,1}, {D4,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1},
        /* Bar 2 (Root C) */
        {C4,1}, {C4,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1},
        /* Bar 3 (Root B) */
        {B3,1}, {B3,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1},
        /* Bar 4 (Root Bb) */
        {BB3,1}, {BB3,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1}
    };

    static struct audio_event events[128];
    int nnotes = sizeof(melody) / sizeof(melody[0]);
    int nevents, tempo = 240; /* High energy tempo */
    struct audio_seq s;

    signal(SIGINT, on_signal);
    seqfd = open("/dev/tty", 0);
    if (seqfd < 0) seqfd = 0;

    nevents = build_events(events, melody, nnotes, tempo);

    printf("Playing Megalovania... Press Ctrl+C to stop.\n");

    while (!stop_requested) {
        s.events = events;
        s.count = nevents;
        s.rate_hz = 0;
        s.flags = AUDIO_SEQ_F_FLUSH;
        
        /* Send to kernel sequencer */
        if (ioctl(seqfd, KIOCSNDSEQ, &s) < 0) {
            perror("ioctl");
            break;
        }
        
        /* Wait while the loop plays (approx 4 seconds per loop at 240 BPM) */
        sleep(4); 
    }

    stop_sound();
    return 0;
}
```

Vutshi:
@toncho11, here we go:

https://github.com/user-attachments/assets/e4913bf2-ed17-4231-b6cf-b718c6e060ac

a bit of frequency wobbling is some kind of MartyPC issue on macOS.

toncho11:
Thanks!

This is a longer version:

```
/* megalo.c - ELKS PC speaker sequencer: Megalovania Full Loop */

#include <sys/ioctl.h>
#include <errno.h>
#include <fcntl.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <linuxmt/kd.h>

/* PIT Divisors */
#define REST 0
#define D4   4063
#define C4   4560
#define B3   4831
#define BB3  5118
#define F4   3417
#define G4   3044
#define GS4  2873
#define A4   2712
#define D5   2031
#define F5   1708
#define G5   1521
#define GS5  1436
#define A5   1356
#define C6   1140

struct note {
    unsigned short div;
    unsigned char dur;
};

static int seqfd = -1;
static volatile int stop_requested;

static void stop_sound(void) {
    struct audio_seq s;
    if (seqfd < 0) return;
    s.events = 0; s.count = 0; s.rate_hz = 0;
    s.flags = AUDIO_SEQ_F_STOP | AUDIO_SEQ_F_FLUSH;
    ioctl(seqfd, KIOCSNDSEQ, &s);
    ioctl(seqfd, KIOCSOUND, 0);
}

static void on_signal(int sig) { stop_requested = 1; }

static unsigned short tempo_ticks(unsigned char dur, int tempo) {
    unsigned long v = (unsigned long)dur * 3000UL;
    v = (v + (unsigned)tempo / 2) / (unsigned)tempo;
    return (v < 1) ? 1 : (unsigned short)v;
}

static int build_events(struct audio_event *out, struct note *in, int nnotes, int tempo) {
    int i, n = 0;
    unsigned short t, gap;
    for (i = 0; i < nnotes; i++) {
        t = tempo_ticks(in[i].dur, tempo);
        gap = (t > 1) ? 1 : 0; 
        if (in[i].div) {
            out[n].divisor = in[i].div;
            out[n].ticks = t - gap;
            out[n].flags = AUDIO_F_TONE;
            n++;
        }
        if (gap || !in[i].div) {
            out[n].divisor = 0;
            out[n].ticks = gap ? gap : t;
            out[n].flags = AUDIO_F_REST;
            n++;
        }
    }
    return n;
}

int main(int argc, char **argv) {
    static struct note melody[] = {
        /* Bars 1-4: The Classic Drop */
        {D4,1}, {D4,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1},
        {C4,1}, {C4,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1},
        {B3,1}, {B3,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1},
        {BB3,1}, {BB3,1}, {D5,2}, {A4,2}, {REST,1}, {GS4,2}, {G4,2}, {F4,2}, {D4,1}, {F4,1}, {G4,1},

        /* Bars 5-8: The Climbing Bridge */
        {F5,2}, {F5,1}, {F5,2}, {F5,1}, {F5,2}, {D5,2}, {D5,2}, 
        {F5,2}, {F5,1}, {F5,2}, {G5,1}, {GS5,2}, {G5,2}, {F5,2}, {D5,1}, {F5,1}, {G5,1},
        {F5,2}, {F5,1}, {F5,2}, {G5,1}, {GS5,1}, {A5,1}, {C6,2}, {A5,4},
        {D5,2}, {D5,2}, {D5,2}, {A5,2}, {G5,4}
    };

    static struct audio_event events[256];
    int nnotes = sizeof(melody) / sizeof(melody[0]);
    int nevents, tempo = 240;
    struct audio_seq s;

    signal(SIGINT, on_signal);
    seqfd = open("/dev/tty", 0);
    if (seqfd < 0) seqfd = 0;

    nevents = build_events(events, melody, nnotes, tempo);

    printf("Playing Full Megalovania... (Ctrl+C to stop)\n");

    while (!stop_requested) {
        s.events = events;
        s.count = nevents;
        s.flags = AUDIO_SEQ_F_FLUSH;
        ioctl(seqfd, KIOCSNDSEQ, &s);
        
        /* The full sequence takes about 8 seconds at this tempo */
        sleep(8); 
    }

    stop_sound();
    return 0;
}
```

toncho11:
@ghaerr Can we increase the slots from 32 to 256 in the kernel? A 256 slot version should not take a lot of memory.

This code benefits from a 256 slot version. It uses Density Modulation (a form of Pulse Width Modulation) to "sculpt" the sound. By breaking each note into 16 micro-instructions, it simulates the behavior of an instrument with a physical string - hitting hard and then fading out.

```
#include <sys/ioctl.h>
#include <fcntl.h>
#include <signal.h>
#include <stdio.h>
#include <unistd.h>
#include <linuxmt/kd.h>

/* PIT Divisors */
#define D4   4063
#define C4   4560
#define B3   4831
#define BB3  5118
#define F4   3417
#define G4   3044
#define GS4  2873
#define A4   2712
#define D5   2031

static int seqfd = -1;

void stop_sound(void) {
    struct audio_seq s = {0, 0, 0, AUDIO_SEQ_F_STOP | AUDIO_SEQ_F_FLUSH};
    ioctl(seqfd, KIOCSNDSEQ, &s);
}

/* 
 * Advanced 16-event ADSR Pluck. 
 * This creates a sharp "attack" followed by a vibrating "fade".
 */
int build_adsr_note(struct audio_event *out, unsigned short div) {
    if (div == 0) { 
        out[0] = (struct audio_event){0, 16, AUDIO_F_REST};
        return 1;
    }
    int n = 0;
    // --- ATTACK (4 ticks) ---
    out[n++] = (struct audio_event){div, 4, AUDIO_F_TONE}; 
    
    // --- SUSTAIN/VIBRATO (8 ticks) ---
    // We alternate tone and rest to create a "thinner" textured sound
    for(int i=0; i<4; i++) {
        out[n++] = (struct audio_event){div, 1, AUDIO_F_TONE};
        out[n++] = (struct audio_event){0,   1, AUDIO_F_REST};
    }

    // --- RELEASE (4 ticks) ---
    out[n++] = (struct audio_event){div, 1, AUDIO_F_TONE};
    out[n++] = (struct audio_event){0,   3, AUDIO_F_REST};
    
    return n;
}

int main() {
    // The 4 bars of the Megalovania opening riff
    unsigned short melody[] = {
        D4, D4, D5, A4, 0, GS4, G4, F4, D4, F4, G4,
        C4, C4, D5, A4, 0, GS4, G4, F4, D4, F4, G4,
        B3, B3, D5, A4, 0, GS4, G4, F4, D4, F4, G4,
        BB3, BB3, D5, A4, 0, GS4, G4, F4, D4, F4, G4
    };
    
    // 44 notes * 16 events = 704 events total.
    // Even with a 256 buffer, we still send in chunks, 
    // but the playback will be perfectly smooth.
    static struct audio_event event_buffer[1024];
    struct audio_seq s;

    seqfd = open("/dev/tty", 0);
    if (seqfd < 0) seqfd = 0;

    printf("Megalovania 256-Buffer Edition\n");
    printf("Using 16-event ADSR sculpting per note...\n");

    while (1) {
        int total_events = 0;
        for (int i = 0; i < 44; i++) {
            total_events += build_adsr_note(&event_buffer[total_events], melody[i]);
        }

        /* 
         * With a 256-slot kernel buffer, this call will dump a huge 
         * amount of data into the driver. The kernel will handle the 
         * rapid pulsing in the background.
         */
        s.events = event_buffer;
        s.count = total_events; 
        s.flags = AUDIO_SEQ_F_FLUSH;
        
        ioctl(seqfd, KIOCSNDSEQ, &s);
        
        // Sleep for the duration of the riff (approx 4-5 seconds)
        sleep(5); 
    }
    return 0;
}
```

Vutshi:
Sounds funky

https://github.com/user-attachments/assets/bb9f0b0c-16b7-4af8-9c4f-55bbbd43bfd7

In principle one should be able to stream the music even using buffer size 8 or 16.

toncho11:
Here is a more complete version. Let's call it **synthpi**.
If this works I will make a PR, but it really needs the 256 slot buffer.

```
#include <sys/ioctl.h>
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>
#include <termios.h>
#include <linuxmt/kd.h>

/* PIT Divisors for the C-Major Scale */
#define C4  4560
#define D4  4063
#define E4  3620
#define F4  3417
#define G4  3044
#define A4  2712
#define B4  2415
#define C5  2280

/* 
 * The Synthesizer Engine: 11-event ADSR Pluck 
 * This simulates a piano-like strike and fade.
 */
int build_pluck(struct audio_event *out, unsigned short div) {
    int n = 0;
    // 1. ATTACK: Solid tone (High volume strike)
    out[n++] = (struct audio_event){div, 4, AUDIO_F_TONE};
    
    // 2. SUSTAIN: 50% density (Perceived half-volume ring)
    for(int i=0; i<3; i++) {
        out[n++] = (struct audio_event){div, 1, AUDIO_F_TONE};
        out[n++] = (struct audio_event){0,   1, AUDIO_F_REST};
    }
    
    // 3. RELEASE: 25% density (Quiet fade-out)
    out[n++] = (struct audio_event){div, 1, AUDIO_F_TONE};
    out[n++] = (struct audio_event){0,   5, AUDIO_F_REST};
    
    return n;
}

void print_ui() {
    printf("\033[2J\033[H"); // Clear screen
    printf("=== ELKS Rudimentary Keyboard Synthesizer ===\n\n");
    printf("  [ A ] [ S ] [ D ] [ F ] [ G ] [ H ] [ J ] [ K ]\n");
    printf("    C4    D4    E4    F4    G4    A4    B4    C5\n\n");
    printf("Press keys to play. Press 'Q' to quit.\n");
    printf("---------------------------------------------\n");
}

int main() {
    int fd = open("/dev/tty", O_RDWR);
    char c;
    struct termios oldt, newt;
    struct audio_event events[16];
    struct audio_seq s;

    if (fd < 0) fd = 0;

    /* Terminal Setup: Enable Raw Mode for instant key response */
    tcgetattr(fd, &oldt);
    newt = oldt;
    newt.c_lflag &= ~(ICANON | ECHO);
    tcsetattr(fd, TCSANOW, &newt);

    print_ui();

    while (1) {
        if (read(fd, &c, 1) <= 0) continue;
        if (c == 'q' || c == 'Q') break;

        unsigned short note_div = 0;

        switch (c) {
            case 'a': case 'A': note_div = C4; break;
            case 's': case 'S': note_div = D4; break;
            case 'd': case 'D': note_div = E4; break;
            case 'f': case 'F': note_div = F4; break;
            case 'g': case 'G': note_div = G4; break;
            case 'h': case 'H': note_div = A4; break;
            case 'j': case 'J': note_div = B4; break;
            case 'k': case 'K': note_div = C5; break;
            default: note_div = 0; break;
        }

        if (note_div > 0) {
            int count = build_pluck(events, note_div);
            
            s.events = events;
            s.count = count;

            /* 
             * --- CRITICAL PERFORMANCE NOTE ---
             * CURRENT (32-slot buffer): We use AUDIO_SEQ_F_FLUSH to clear the 
             * small buffer so the next note starts instantly. This cuts off 
             * the previous note's fade-out (staccato effect).
             * 
             * FUTURE (256-slot buffer) without FLUSH
             * This will allow notes to queue up. We should hear the "release tails" 
             * of previous notes continuing to ring while we play new ones, 
             * creating a richer, organic piano sound.
             */
            s.flags = AUDIO_SEQ_F_FLUSH; 
            
            ioctl(fd, KIOCSNDSEQ, &s);
            printf("\rPlaying: %c (Div: %u)  ", c, note_div);
            fflush(stdout);
        }
    }

    /* Reset speaker and terminal */
    struct audio_seq stop = {0, 0, 0, AUDIO_SEQ_F_STOP | AUDIO_SEQ_F_FLUSH};
    ioctl(fd, KIOCSNDSEQ, &stop);
    tcsetattr(fd, TCSANOW, &oldt);
    printf("\n\nSynth powered down.\n");
    return 0;
}
```

ghaerr:
> Can we increase the slots from 32 to 256 in the kernel? A 256 slot version should not take a lot of memory.

> In principle one should be able to stream the music even using buffer size 8 or 16.

I'm not understanding why the slot increase is needed? @Vutshi's design only requires a small buffer since the application supplying the sound data loops calling ioctl continuously, and the kernel buffer is only required to be larger than the max time it might take for the application to be round-robin rescheduled, which in many cases would likely be the only application running, so small should work.

> By breaking each note into 16 micro-instructions, 

Is a larger buffer required for dynamic modulation or something else? For how long a period does the dynamic modulation actually have to occur? [EDIT: see below, I think I figured out the dynamic modulation is done in user mode, but don't actually know the truly required slot count, since each slot has its own time component].

Currently, audio isn't used really at all except for the single digger program. 32 slots take up 160 bytes kernel space, 256 would use 1280, which would be considered a very large static buffer for the kernel data segment - a very scarce resource. If truly required, it would probably have to be moved to a dynamic allocation, which then might require opening a /dev/audio device etc, thereby complicating the implementation and using up even more kernel memory. 

Another choice would be to have the application pass the data from user space and eliminate the kernel slots entirely; I have advocated for this from the start, but the current implementation is fairy simple and there's only a single ELKS program ported to use audio.

> Sounds funky

What are the reasons this sounds inferior to existing audio implementation, when it is supposed to sound better? [EDIT: see below for untested 256 slot recommendation?].

> If this works I will make a PR, but it really needs the 256 slot buffer.
```
            /* 
             * --- CRITICAL PERFORMANCE NOTE ---
             * CURRENT (32-slot buffer): We use AUDIO_SEQ_F_FLUSH to clear the 
             * small buffer so the next note starts instantly. This cuts off 
             * the previous note's fade-out (staccato effect).
             * 
             * FUTURE (256-slot buffer) without FLUSH
             * This will allow notes to queue up. We should hear the "release tails" 
             * of previous notes continuing to ring while we play new ones, 
             * creating a richer, organic piano sound.
             */
```
I just noticed this. I would suggest rebuilding the kernel yourself and testing with a 256 slot buffer to see whether this helps or not. As stated above, I have concerns about the requirement for using a large dedicated kernel buffer, if it's just for a sample audio program. We should seriously consider moving to a user-supplied buffer to mitigate these concerns. I still don't quite understand why the kernel itself needs 256 slots, since the dynamic modulation is done in user space and uses the same audio_event/audio_seq structures to actually produce the sound from a kernel timer.

[EDIT: I just noticed that the synth program is not using the suggested/required method of continuously calling ioctl to feed the audio data into a small kernel buffer. I think doing so should remove the requirement of a large slot count. If this code is all AI-generated the prompts should include looking at @Vutshi's digger and/or test_audio program both of which have more complex handling of ioctl queuing as well as fallback playback code for the same audio data using the original KIOCSOUND ioctl when the kernel doesn't have the new KIOCSNDSEQ implementation present].

That said, we'll need to all agree on what audio engine we require in the kernel, versus C library routines for work that can be done outside the kernel, prior to sending the raw data in a syscall. I can already see the need for C library routines to ensure various AI-suggested implementations of audio generators don't start from scratch every time differently].

In general, I will probably defer to @Vutshi for now for audio engine decisions, since the digger project is still under construction and seems to be getting enhanced rapidly.

ghaerr:
> This will allow notes to queue up. We should hear the "release tails" 
 > of previous notes continuing to ring while we play new ones, 

I don't think this is the way the kernel sound engine works: each sequence of audio data is played for its specified duration and frequency and then discarded. The audio events are not merged or continued in any way after playing/processing.

Even though the AI suggested using AUDIO_SEQ_F_FLUSH to start a new sound (which also violates the concept of continuously writing the ioctl queue allowing a small slot count), it would seem to me the "dynamic modulation", "instant start", and "release tails" can all be implemented by better pre-processing of the original audio sequence data *before* then sending the larger volume of processed audio sequence events to the kernel using the recommended continuous writing and small queue size.

Thus I am thinking the issue here isn't changing the kernel audio engine but better user-mode processing of the audio sequence data giving a smoother perceived output using more audio events fed continuously as described.

toncho11:
@ghaerr What I wanted to do was to use some hacks and squeeze the maximum of the pc speaker. Understand quickly how it works. To make it sound more natural that it was designed for. 
Anyway.

ghaerr:
> use some hacks and squeeze the maximum of the pc speaker

I agree! Didn't mean to go on about it I think I was posting as I was figuring it out. I think it would be great to have better sound, and it seems that could be done with your idea of dynamic modulation - it requires more code on the app side, rather than a kernel change, now that I think I understand it. If you can get that to work, we can add that as a C library routine so that better sounds are easily generated from simple data.

toncho11:
The 11 or 16-slot idea is a technique that divides a single musical note into a sequence of tiny time fragments, allowing the computer to simulate volume and fading by rapidly toggling the speaker on and off to create different "densities" of sound.

ghaerr:
@Vutshi and @toncho11:

I went ahead and spent some time compiling both synthpi.c as well as the earlier density modulation example program, both written by @toncho11's AI. In addition, I increased the slots to 256 in the kernel audio driver. I wanted to make sure my untested analysis wasn't in error.

Here are the results of the tests between 32- and 256-slot kernel audio and the two example programs:
- No difference in sound output in the synthpi.c program.
- The sound output in the original dynamic modulation example program didn't change either, except that the output played for a longer period of time.

The conclusion is that adding more slots to the kernel doesn't actually make the example density modulation programs work, despite source code comments to the contrary. The longer play time for the first example program is consistent with the fact that more kernel slots available allow acceptance of more audio events in a single ioctl, given that neither example continuously sends ioctl audio events as is otherwise required.

My guess is that the AI confused breaking up each note into 11 or 16 time "slots" with the audio event "slots" in the kernel. I still think it possible that the proposed ADSR technique could be implemented in user mode code if wanted.

toncho11:
For me it was that a  256-slot will alow for less IO operations, preventing some sound glitches in the case when a tone has sarted to occupy much more slots. 

Thank you for testing. Does the synthpi.c sounds good? I am not able to test right now. 

It fakes piano-like fading volume by rapidly alternating tone and silence, using less and less tone density after the initial loud attack.

toncho11:
So the idea is:

Attack:   100% tone density  -> loud strike
Sustain:   50% tone density  -> quieter ring
Release: ~17% tone density   -> very quiet fade-out

Basically:

more tone time = louder
more rest time = quieter

It is an attenuated ADSR (Attack => Decay => Sustain => Release) trick.

ghaerr:
> preventing some sound glitches in the case when the tone has sarted to occupy much more slots.

Each queued sound is played for the duration of a timer set by the kernel scheduler which runs a high priority, so the number of slots is highly unlikely to affect sound output, unless possibly quite a few other heavy number crunching applications are running simultaneously. With the sound application supposed to be running in a tight loop sending ioctl's, that's not an issue to worry too much about at present.

> Does the synthpi.c sounds good? I

No, it's pretty glitchy sounding, not very good.

> Attack: 100% tone density -> loud strike
Sustain: 50% tone density -> quieter ring
Release: ~17% tone density -> very quiet fade-out

Yes, I get that. But the user program isn't accomplishing that, it seems.

> It fakes piano-like fading volume by rapidly alternating tone and silence,

I would guess that programming the silence in is greatly contributing to the glitchiness of the sound heard, but not sure.  I haven't studied how well different algorithms change amplitudes but continue to sound good. Perhaps even more time slices are needed?

It is possible that a base interval of much less than ELKS' 10ms hardware timer be required for chopping the sounds into smaller fragments. I don't know audio that well.

toncho11:
I see the idea is difficult to implement. It seems 10 ms is too slow for fake amplitude/volume control. I do not see any solutions right now.

Thank you.

PR #2695, opened 2026-06-09
Implement parallel builds

djohanse:
I know previous attempts at making parallel builds in elks have been largely unsuccessful. So I understand @ghaerr if you choose not to take this one.

That being said...

Had opencode review the makefiles and evaluate what would make sense to build in parallel.  I've been using these changes for a few days and they seem to be working well.  

Tested on Mac and linux.  On my linux PC the make after the config menu selections runs in almost no time (12 cores in parallel).

Some notes:

1. if terminal size was less than 19 lines by 80 columns the config menu would not appear.  It would just continue with the build. Modified to display error message if screen size is too small.  Also provided a --config flag and alternate instructions on how to use a pre-existing config file if the user cannot resize their screen and wants to skip the config menu and just use a pre-existing config.

2. Port 8021 conflicts with launchd com.apple.ftp-proxy on macOS 26.5.  Modified qemu.sh to use 2121 instead of 8021 (just picked 2121 at random).

3. Originally on Mac it was only running on the efficiency cores.  That's the reason for the performance core specific line in line 26 of env.sh

4. I've really only tested nominal builds.  If there are any good ways to stress-test this please let me know.

5. Tested builds of the cross-compiler as well as the "everyday" build.


ghaerr:
@djohanse, thanks for submitting this. Yes, I get very concerned about changing the make system, as I've tried hard over the years to allow it to work on (almost) everyone's system. I'm not real keen on having to support lots of specific system issues all over again.

That said, some of your changes are quite straightforward, (e.g. menuconfig fixes); perhaps these fixes should be applied in smaller chunks. I can do that with separate commits, or you can. But let's talk a bit more first.

Is the P-core issue only relevant on macOS ARM machines, or also x86?

I haven't had time to fully review, but can you explain exactly why this version allows for safe parallel builds with `make -j`? I only see a couple unbundlings of `make -C` and don't fully understand it. There are definitely race issues with the kernel build creating new files like include/compiler-generated.h, include/arch/asm-offsets.h, etc. I don't see how they're handled here, as I don't understand exactly how/when `make` allows for parallel builds using multiple `$(MAKE) -C` within multiple targets within a Makefile.

I like the idea of setting MAKEFLAGS outside the build (in env.sh) instead of internal to the Makefiles. The internal -j1 that your PR removes had to be included because of the above race issues. So likely more testing and understanding is required. This one is tough since it was written by AI and is thus prone to errors.

djohanse:
> @djohanse, thanks for submitting this. Yes, I get very concerned about changing the make system, as I've tried hard over the years to allow it to work on (almost) everyone's system. I'm not real keen on having to support lots of specific system issues all over again.
> 
> That said, some of your changes are quite straightforward, (e.g. menuconfig fixes); perhaps these fixes should be applied in smaller chunks. I can do that with separate commits, or you can. But let's talk a bit more first.
> 

If you want to merge the "low-hanging fruit" (terminal size issue and 8021 port issue) in separate commits I'm good with that. 

> Is the P-core issue only relevant on macOS ARM machines, or also x86?
> 

MacOS ARM machines only to the best of my knowledge.

> I haven't had time to fully review, but can you explain exactly why this version allows for safe parallel builds with `make -j`? I only see a couple unbundlings of `make -C` and don't fully understand it. There are definitely race issues with the kernel build creating new files like include/compiler-generated.h, include/arch/asm-offsets.h, etc. I don't see how they're handled here, as I don't understand exactly how/when `make` allows for parallel builds using multiple `$(MAKE) -C` within multiple targets within a Makefile.
> 

Been thinking about this.  As you allude to below, having AI doing the work requires increased review and testing. A lot of the time the first thing it outputs doesn't work.  But it offers unprecedented opportunities for rapid development. And it usually gets to a good answer eventually.

That being said, let's take a step back.  Ignoring the proposed changes in this branch for the moment, I'd like to try to understand, reproduce, and if possible fix the underlying issue.  

Going back to the master branch, it looks like the -j1 flag was added specifically for when building the different disk images.  

As a starting point, I had openCode do an analysis and propose a solution.  

Does the following seem reasonable?   Also does this reflect what you were seeing with the race/parallel builds?

Note I don't plan to make code changes until we have a plan...

Conversation with open code (includes my questions and its responses):
====================================================================================
(djohanse question): checked out master.  are there race problems in the current build/makefile system with build allimages without the -j1

+ Thought: 2.1s
✱Grep "allimages|images" in . (11 matches)
+ Thought: 4.6s
→Read image/Makefile 
+ Thought: 4.9s
Line 44 has .NOTPARALLEL: images images-minix images-fat images-hd — this serializes all image builds even under -jN. Without it, parallel builds would race since every image target modifies the same $(DESTDIR) template directory.
But the bigger concern is the elkscmd install phase: Make.install iterates Applications and copies binaries into $(DESTDIR)/bin. 

With -jN on the top-level elkscmd target, multiple application sub-makes could run install concurrently, all writing into the same $(DESTDIR) tree — no .NOTPARALLEL protection there.
On master (no MAKEFLAGS set, defaulting to -j1), neither issue arises.
▣  Build · Big Pickle · 41.0s

(djohanse question):can this be fixed to not race and also support parallel builds?


+ Thought: 4.8s
 Let me study the install system to understand the race points.
→Read elkscmd/Make.install 
+ Thought: 13.3s
Yes, it can be fixed. Two race points:
1. elkscmd build — Already safe. Each subdirectory compiles independently into its own objects. The install phase (Make.install) is a single while read shell loop processing one app at a time, so no parallelism there.
2. Image builds (image/Makefile) — This is the real problem. Every image target (fd1440-minix, fd2880-fat, etc.) writes into the same $(DESTDIR) (set to $(TOPDIR)/target). Running two in parallel corrupts each other's template. That's why line 44 has .NOTPARALLEL: images images-minix images-fat images-hd.
The fix: give each image its own staging directory instead of sharing DESTDIR. For example, in Make.image change:
- DESTDIR = $(TOPDIR)/target
+ DESTDIR = $(TOPDIR)/target-$(NAME)
Then .NOTPARALLEL can be removed and make -jN images would build every image variant in parallel — each has its own isolated template. The tradeoff is more disk space during builds.
▣  Build · Big Pickle · 39.3s


(djohanse question): what does this mean: Running two in parallel corrupts each other's template

+ Thought: 2.6s
Two parallel $(MAKE) -f Make.image processes both do:
1. Copy rootfs_template/ → target/
2. Install compiled binaries into target/bin/
3. Generate filesystem image from target/
If they run concurrently:
- Process A copies template, then starts installing apps
- Process B re-copies the template, wiping out A's installed binaries
- Process A installs more files
- Process B's filesystem snapshot includes some of A's files but is also missing others
- Process A's filesystem snapshot gets B's template copy instead of its own
Both end up with incomplete or mixed-up root filesystems — missing binaries, wrong files, or even inconsistent metadata. It's not disk corruption, just a garbled build output.



> I like the idea of setting MAKEFLAGS outside the build (in env.sh) instead of internal to the Makefiles. The internal -j1 that your PR removes had to be included because of the above race issues. So likely more testing and understanding is required. This one is tough since it was written by AI and is thus prone to errors.

ghaerr:
@djohanse:

I've been reading up on exactly how parallel make works, and have a better understand now of what needs to be done.

I now believe we should work towards getting the ELKS build working using parallel makes, but IMO this should be done in 4-5 stages: the first stage should be getting the kernel build itself race-free, as mentioned above regarding .h files being created from scratch during the process, which must be created before other make jobs can run. The C library build is complicated, since it builds a tree deeper used by the ia16-elf-gcc compiler itself, and there are some definite tricky areas that currently cause C source files to not be compiled when changed when running make from elks/libc rather than the top-level. Unfortunately, the C library takes a significant time to build, so this remains important but problematic.

The C library needs to be built before the ELKS kernel can build, and a method is required for syncing this. After the C library and kernel are built, it should be pretty straightforward to build elkscmd/, since most (except elkscmd/lib) are independent. This may be an area that could be parallelized before the kernel, although executed after the kernel build.

As your AI points out, the distribution images can't be parallelized without creating multiple temp directories, which, for other reasons, isn't currently a good idea. So the distribution images (built with "cd images; make" likely will be built linearly. Finally, the ./build.sh (and ./buildimages.sh and ./buildext.sh) need to be parallelized, which can probably be most easily done after all the other build processes are sped up.

So, lots to do. I don't want to take the chance of causing arbitrary build failures for lots of users due to lack of testing, so this probably needs to be done in phases.

> As a starting point, I had openCode do an analysis and propose a solution.
Does the following seem reasonable? 

Unfortunately, not really. My comments above are mostly why.

> Also does this reflect what you were seeing with the race/parallel builds?

No, it misses it entirely. (The kernel build has race issues, the image build specifically uses .NOTPARALLEL to stop races).

My summary of AI results: misses the major kernel race issue(s), generally agree with elkscmd/ build being generally safe (but don't agree with fancy make functions that may not exist on all makes), and disagree that suggested image building fix will work since many other files are created in the same working directory which will also overwrite each other, by the make target rules.

This PR also contains other enhancements/fixes which should be discussed further, so lets leave this open for more discussion.

djohanse:
Re "the first stage should be getting the kernel build itself race-free"

Some questions.  What are the characteristics of the race condition (e.g how did you realize/determine these races exist)?  How would I identify a correct vs incorrect build?  Is there a  way to consistently reproduce a bad build created by the race condition?

ghaerr:
> What are the characteristics of the race condition (e.g how did you realize/determine these races exist)? 

A while back, in some earlier parallel testing, the kernel build failed due to a header file (e.g. asm-offsets.h or possibly compiler-generated.h) being required by the build before it was generated. This is race condition. I haven't had time to get more detail lately, but this kind of thing can all likely be fixed with proper prequisite rules, plus better understanding of exactly how and when `make` allocates parallel jobs. It won't spawn a new job unless it feels all prerequisites have been completed, of course, but its a bit complicated with the sub-make considerations.

> Is there a way to consistently reproduce a bad build created by the race condition?

I haven't found a consistent way, but this could possibly be more tested on your machine by running just the kernel build after a `make clean` that clears out the compiler-generated headers. This testing is complicated by the fact that `make clean` also cleans the C library, which takes a long time to build. So it's a bit of a pain at the moment. 
[EDIT: A "cd elks; make -j8" might work around the C library clean issue].


 Perhaps the C library should be attacked first, but this is also complicated by ELKS now using the same directory for GCC, OWC and C86 compiler builds of the C library.

ghaerr:
@djohanse: I have made some progress on getting parallel make working and will be posting a PR shortly, after a bit more testing.

ghaerr:
@djohanse,

Thanks for your work on parallel builds! Much of your code and ideas went into #2701. I think I tracked down the kernel build race problem(s) discussed above.

I've made a note to consider the AI's thoughts on building images concurrently, but that usually isn't performed on most builds.

I'm going to close this now. I will add the hostfwd qemu.sh fix in another commit and we can discuss ./build.sh --config here separately if desired. While a nice option, I think it might be better to use `./build.sh --config <file>`, although overwriting .config is somewhat dangerous for the user.

Thank you!

PR #2699, opened 2026-06-12
fix net stop

djohanse:
1. Fix grep error on net stop which was causing net stop to not actually stop the network (this is fixed)

Screenshot 2026-06-11 at 10 56 54 PM

2. modify qemu.sh to use port 2121 instead of 8021.  8021 conflicts with an ftp proxy server in macOS 26.5

3. Net show shows status of net processes:

Screenshot 2026-06-11 at 10 59 54 PM

4. Mods so serial console works better (easier for copy and paste)

Screenshot 2026-06-11 at 11 03 05 PM

### DIFF
--- a/elkscmd/rootfs_template/bin/net
+++ b/elkscmd/rootfs_template/bin/net
@@ -69,7 +69,15 @@ start_network()
 stop_network()
 {
 	echo "Stopping network"
-	kill $(ps | grep "ktcp|telnet|httpd|ftpd" | cut -c 1-5) > /dev/null 2>&1
+	ps > /tmp/psout
+	while read pid cmd
+	do
+		case $cmd in
+		*ktcp*|*telnet*|*httpd*|*ftpd*)
+			kill $pid > /dev/null 2>&1
+			;;
+		esac
+	done < /tmp/psout
 	custom_stop_network
 }
 
@@ -89,7 +97,16 @@ restart)
 	start_network ;;
 show)
 	echo -n ip $localip gateway $gateway mask $netmask $link
-	if test "$link" = "slip"; then echo "" $baud $device; else echo; fi ;;
+	if test "$link" = "slip"; then echo "" $baud $device; else echo; fi
+	running="[stopped]"
+	ps > /tmp/psout
+	while read pid cmd
+	do
+		case $cmd in
+		*ktcp*) running="[running]" ;;
+		esac
+	done < /tmp/psout
+	echo "ktcp $running" ;;
 *) usage ;;
 esac
 
--- a/elkscmd/rootfs_template/etc/inittab
+++ b/elkscmd/rootfs_template/etc/inittab
@@ -1,4 +1,4 @@
-id:1:initdefault:
+id:3:initdefault:
 
 si::sysinit:/etc/rc.sys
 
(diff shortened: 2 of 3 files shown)


elkscmd/rootfs_template/bin/net
```diff
@@ -69,7 +69,15 @@ start_network()
...
-	kill $(ps | grep "ktcp|telnet|httpd|ftpd" | cut -c 1-5) > /dev/null 2>&1
+	ps > /tmp/psout
+	while read pid cmd
+	do
+		case $cmd in
+		*ktcp*|*telnet*|*httpd*|*ftpd*)
+			kill $pid > /dev/null 2>&1
+			;;
+		esac
+	done < /tmp/psout
```
ghaerr, line 80: I have a couple issues with this change: where is `grep`? It's supposed to be on all distribution disk images, so this change should not be needed. Second issue is that, in general, it's not a good idea to have this script write a file (/tmp/psout) since that both slows down the floppy disk activity and also dirties the floppy media, both unnecessary. The same issue for below with the "net show" enhancement.

elkscmd/rootfs_template/etc/inittab
```diff
@@ -1,4 +1,4 @@
-id:1:initdefault:
+id:3:initdefault:
```
ghaerr, line 1: This change isn't well suited for distribution images, since Nano-X and programs that depend on the mouse then won't work out of the box. Best to keep these kinds of changes for your personal system, if wanted.

qemu.sh
```diff
@@ -143,5 +141,5 @@ fi
 
 echo "Using QEMU: $QEMU $ACCEL"
 exec $QEMU $ACCEL $DEBUG $AUDIO $CONSOLE -nodefaults -name ELKS -machine isapc -cpu 486,tsc -m 8M \
-$KEYBOARD $QDISPLAY -vga std -rtc base=utc $SERIAL \
+$KEYBOARD $QDISPLAY -vga std -boot order=a -rtc base=utc $SERIAL \
```
ghaerr, line 144: Does `-boot order=a` force floppy boot when an HD image is also specified? If so, this change is best kept in the (commented out) IMAGE= lines at the top of the file, rather than in the main command line, for greater flexibility in qemu IMAGE= options.

elkscmd/rootfs_template/bin/net
djohanse: That's bizarre.  I pulled master and did builds yesterday and grep wasn't there.  It's there tonight.  Ok let's throw out these changes you're not in favor of.  I think that leaves port 2121 and serial terminal.  Very minor changes.  Do you want me to modify this PR or do you want to make the changes and I'll cancel the PR.

ghaerr: Let's just hold for a second and think through what we want changed. I'm writing a response to your parallel make issues now. In general, smaller PRs that only contain a single issue fixed/enhanced, etc is always better, which avoids these issues.

qemu.sh
djohanse: So all that's left is the port 2121.  Will withdraw this PR.  Still learning though; appreciate the explanations and patience.

djohanse:
rejected

PR #2703, opened 2026-06-17
fix traceroute and ifconfig bugs

djohanse:
Couple bug fixes.  

1. ifconfig had the interface name hardcoded to ne0.

2. for in_resolve, if you did:

ping -c 1 google.com
traceroute google.com
ping -c 1 google.com
traceroute google.com

the second traceroute  would fail. 

Now can consistently call ping and/or traceroute successfully

### DIFF
--- a/libc/net/in_resolv.c
+++ b/libc/net/in_resolv.c
@@ -159,7 +159,10 @@ ipaddr_t in_resolv(const char *hostname, char *server, int *ancount)
 		signal(SIGALRM, old);
 		close(fd);
 
-		if (rc < sizeof(struct DNS_HEADER) + sizeof(struct RR))
+		/* must check rc <= 0: read() returning -1 promotes to unsigned
+		 * in the comparison below, making -1 appear > 28 and skipping
+		 * the retry, then parsing the stale request buffer as a response */
+		if (rc <= 0 || (unsigned)rc < sizeof(struct DNS_HEADER) + sizeof(struct RR))
 			continue;
 
 		dns = (struct DNS_HEADER *)buf;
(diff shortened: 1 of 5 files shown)


libc/net/in_resolv.c
```diff
@@ -159,7 +159,10 @@ ipaddr_t in_resolv(const char *hostname, char *server, int *ancount)
 		signal(SIGALRM, old);
 		close(fd);
 
-		if (rc < sizeof(struct DNS_HEADER) + sizeof(struct RR))
+		/* must check rc <= 0: read() returning -1 promotes to unsigned
+		 * in the comparison below, making -1 appear > 28 and skipping
+		 * the retry, then parsing the stale request buffer as a response */
+		if (rc <= 0 || (unsigned)rc < sizeof(struct DNS_HEADER) + sizeof(struct RR))
```
ghaerr: Yes, comparing an int with unsigned `sizeof` can be tricky. But this pattern is the same as other code you (or your AI) wrote in netconf.c lines 47-51: wouldn't it be simpler and more efficient to stay with that pattern and use a single comparison, instead of two:
```
if (rc < (int)(sizeof(struct DNS_HEADER) + sizeof(struct RR)))
```

djohanse: Yeah I think I like that better.  Will fix this evening when I get a chance (unless you want to make the change)

ghaerr: I'll wait for your change, thank you.

djohanse: fixed

ghaerr:
Thanks @djohanse!

PR #2712, opened 2026-06-21
ssd-xms.c build crash

djohanse:
Got a crash while building.  Had open code investigate and then manually confirmed fix.

### Steps to produce error:

Starting from master:

git checkout master
cp elks/arch/i86/defconfig .config

Then edit .config to enable:
CONFIG_FS_XMS=y
CONFIG_FS_XMS_RAMDISK=y
Then:
source env.sh && make -C elks clean && make -C elks

Reason: (according to open code): In C89, undeclared globals were implicitly int (just a warning), but the ia16-elf-gcc toolchain with -melks -mcmodel=small rejects them as hard errors.)

### build result after changes:

...
ia16-elf-ar rcs mm.a malloc.o user.o memcpyfs.o xms.o
ia16-elf-ar rcs akernel.a strace.o irq.o irqtab.o process.o entry.o signal.o timer.o softirq.o irq-8259.o reset-ibm.o timer-8254.o divzero.o nmi.o system
(cd ../.. ; ia16-elf-ld.gold  -M  -T /Users/derekjohansen/djelksclone/elks/elks/elks-small.ld \
		arch/i86/boot/crt0.o \
		init/main.o '-(' fs/fs.a kernel/kernel.a lib/lib.a net/net.a fs/minix/minixfs.a fs/msdos/msdos.a arch/i86/kernel/akernel.a arch/i86/lib/l
		-o arch/i86/boot/system > arch/i86/boot/system-full.map ; \
		ia16-elf-nm arch/i86/boot/system | sed -e '/&/d; /!/d' | sort > \
			arch/i86/boot/system.map; \
		elf2elks --symfile arch/i86/boot/system.sym arch/i86/boot/system)
tools/build arg_ignored boot/setup boot/system > boot/Image
Root device is (3, 128)
Setup is 1576 bytes (4 sectors).
Kernel is 101952 B (57456 B code, 27552 B fartext, 9296 B data and 7648 B bss)
System is 100200 B (0x1877 paras)
Load segment start 0x1400 end 0x1ba7Load segment start 0x1400 end 0x1ba7

### DIFF
--- a/elks/arch/i86/drivers/block/ssd-xms.c
+++ b/elks/arch/i86/drivers/block/ssd-xms.c
@@ -25,6 +25,8 @@
 /* current implementation requires no other XMS allocations other than XMS buffers */
 static ramdesc_t     xms_ram_base;      /* ramdisk XMS memory start address */
 static unsigned  int xms_ram_size;      /* ramdisk size in Kbytes */
+extern int xms_enabled;
+extern unsigned int xms_alloc_ptr;
 
 /* initialize SSD device */
 sector_t ssddev_init(void)


ghaerr:
Taking the steps to reproduce the problem results in a different result than listed over here:
```
arch/i86/drivers/block/blk_drv.a(ll_rw_blk.o):function blk_dev_init: error: undefined reference to 'ssd_init'
arch/i86/drivers/block/blk_drv.a(ll_rw_blk.o):function blk_dev_init: error: undefined reference to 'ssd_init!'
arch/i86/drivers/block/blk_drv.a(ll_rw_blk.o):function blk_dev_init: error: undefined reference to 'ssd_init!'
ia16-elf-nm: 'arch/i86/boot/system': No such file
elf2elks: error: cannot open input file `arch/i86/boot/system': No such file or directory
make[2]: *** [boot/system] Error 1
make[1]: *** [Image] Error 2
make: *** [elks] Error 2
```
This is the result of ssd_init() called in ll_rw_block.c but sms-ssd.c not being compiled. Since ssd-sms.c isn't being compiled, I'm not seeing the error you're trying to fix.

Nonetheless, the extern globals in your PR fix are declared in include/linxumt/memory.h, so including that file is the proper solution. However, I would like to be able to reproduce your problem before doing so.

> cp elks/arch/i86/defconfig .config

Unfortunately, `defconfig` isn't a working config file - although it possibly should be, it just contains the defaults for the  `make menuconfig` etc configuration. Various config options are missing, and ultimately, that's the reason for this issue; there are some configurations that produce build errors. However, none of those configurations should be produced (even after copying defconfig) if `make menuconfig` is then ran and immediately exited with changes saved.

I'm trying to track down getting the build error you're seeing (but not posted), and haven't been able to do that yet.

> Reason: (according to open code): In C89, undeclared globals were implicitly int (just a warning)

While untyped variables are assumed to be type int, every C compiler nowadays, including C89, requires globals to be declared, even in some cases as noted without a type (which defaults implicitly to int).

ghaerr:
> source env.sh && make -C elks clean && make -C elks

Also, running `make -C elks` is not an approved method of building the kernel. This error may be the result of not including 'all' in the make, since Makefile-rules in included in elks/Makefile before 'all:'. In general, it is recommended to use `make clean; make` from the top level, or for slightly quicker, `make kclean; make kimage` to just rebuild the kernel.

djohanse:
> > source env.sh && make -C elks clean && make -C elks
> 
> Also, running `make -C elks` is not an approved method of building the kernel. This error may be the result of not including 'all' in the make, since Makefile-rules in included in elks/Makefile before 'all:'. In general, it is recommended to use `make clean; make` from the top level, or for slightly quicker, `make kclean; make kimage` to just rebuild the kernel.

I typically don't build using non-prescribed methods for this very reason.  There's a million ways to do something wrong and waste everybody's time.

What I was actually trying to was figure out how to select an fd2880 image from the config menu(Never did figure that out though).  I'm running out of room on fd1440.  .  But when I was building I saw this error and tried to reverse-engineer how to reproduce it.

Seems likely it was a combination of stale files and incorrect process.  Apologies.  I try to do my due diligence; I'm loathe to waste peoples time.

So going to put this on the shelf and call it a red herring unless/until I can reliably reproduce it with the correct process.

ghaerr:
> Apologies. I try to do my due diligence; 

No worries - I'm happy to try to help you learn the system.

> likely it was a combination of stale files

I think that might actually be the case.

> What I was actually trying to was figure out how to select an fd2880 image from the config menu

`make menuconfig` -> Target Image -> should produce this screen, from which you select FD2880.

Screenshot 2026-06-21 at 1 56 10 PM

Issue #2587, opened 2026-01-20
Mkfat needs reboot after fdisk partition, then crash running sys -M on BIOS HD


bastetfurry:
**Description**
- What is expected ?

mkfat formating the harddisk partition as outlaid in the wiki.
 
- What is observed ?

Failing with: (Had to manually type this, no copy_n_paste from the emulator)
bioshd: sector 0 access beyond partition (-1,0)
bioshd: I/O read error dev 0301 lba sector 0

Works after a reboot.

**Configuration**
- Version (commit or release): Release v0.8.1
- Emulator or HW platform: Emulator, 86Box v5.0

**How to reproduce ?**
- List the steps to reproduce the problem.

PX XT with 7.16 MHz, 640 KB, installed FPU, CGA, PC/XT FDC, PC/XT XTIDE, 3½" 720k, 5¼" 360k, 255 MByte HDD image (your average CF card you would use in an XT)

So someone with write access to the wiki should add that one needs to reboot the system after doing the fdisk dance.

bastetfurry:
And sys just crashed the system... if someone wants to dig trough this, i attached my config file for 86Box.

[XT.cfg.txt](https://github.com/user-attachments/files/24744468/XT.cfg.txt)

<img width="648" height="608" alt="Image" src="https://github.com/user-attachments/assets/8b09b71c-1851-4916-8966-4f1f8e75acfb" />

ghaerr:
I'm not really sure if you're saying that `mkfat` worked, but then required a reboot and is not documented, or whether `mkfat` failed and did not work. I have also heard of some problems with ELKS running on 86Box, there may be a compatibility issue. I don't have 86Box available on my macOS laptop to further test, I am sorry.

The screenshot seems to show a write-protected device, was it just these three errors or did that occur for many sectors?

It might be worth running a much later version than v0.8.1, which is over a year old - there has been much done on the system since then. Images can be downloaded from any commit, available automatically by GitHub CI.

I can help with more debugging information if a photo is provided of the boot screen, as well as exactly when `mkfat` or `fdisk` is failing.

bastetfurry:
I will try a newer version then first, if that too fails i will write down every step. Or better yet, if you have the time, we make a video call and i share my screen. :)

bastetfurry:
After a bit of fiddling i got it to at least segfault and not outright crash on me. ^^'

[...]

PR #2714, opened 2026-06-22
[kernel] Warn when cylinder count too large for BIOS HD driver, increase kstack

ghaerr:
This PR fixes all the issues found and discussed in #2587.

When using the ELKS BIOS disk driver, the maximum number of cylinders that can be used for CHS addressing is 1024 (0-1023). Previously, no check was made for cylinder overflow/truncation when performing I/O to/from hard drives, and the cylinder number was truncated and I/O performed, silently causing filesystem or user data corruption.

The ATA/CF driver (/dev/cfa) uses LBA rather than CHS addressing, and must be used when accessing large SSD or hard drives.

This PR implements a warning a boot, `CYL > 1024, must use /dev/cfa`, letting the user know to use /dev/cfa rather than /dev/hda to eliminate otherwise possibly silent errors. Note that a large hard drive may be able to be booted from /dev/hda if /linux and programs run from /bin are all in the early portions (i.e. cylinder numbers < 1024) of the disk/SSD.

BIOS HD I/O past the 1024th cylinder (> 1023) now give I/O errors to stop any data corruption.

During the testing of the above, a problem which previously wasn't able to be duplicated was finally duplicated and found: during `sys -M /dev/hda1` execution, the `cp -R` command was crashing due to stack underflow and terminated. The real cause was finally found to be kernel stack overflow! What was happening is that the system was booted on a FAT floppy; the FAT filesystem is known to use a large amount of kernel stack. Then, the `sys` script executed `cp` which recursively copied files from one FAT filesystem to another, which overflowed the kernel stack of `cp`, going downwards through the task structure by 14 bytes and overwriting the t_begstack task structure member, which holds the stack base address. Then, during regular system call integrity checking in `check_ustack`, called before each syscall in \_irqit, the process SP appeared to be outside the stack base location and the process was terminated with a SIGSEGV.

After testing with CONFIG_TRACE on, the kernel stack was increased by 30 bytes (times 16 tasks = 480 byte extra permanent kernel heap usage), which should allow for enough stack to operate. @Mon0x4A, it would be nice if you could test this on your real hardware.

Finally, this PR implements a full-time "quick" kernel stack check after ever syscall, to catch any similar (future) kernel stack overflows immediately before returning from the syscall to user mode. This will make it easier to spot in the future, which will be solved by increasing kernel stack even more.

The `sys` script adds the verbose -v option to `cp` to see more easily what is being copied during system copying.

### DIFF
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -122,6 +122,9 @@ notMFM1024:
     BD_BP = offset;
 #else
 
+    if (drivep->cylinders > 1023)       /* return special error on cylinder overflow */
+        return -1;
+
 #if RESET_DISK_CHG
     static unsigned last = 0;
     if (running_qemu && drive != last) {
--- a/elks/arch/i86/drivers/block/bios.c
+++ b/elks/arch/i86/drivers/block/bios.c
@@ -122,7 +122,8 @@ notMFM1024:
     BD_BP = offset;
 #else
 
-    if (drivep->cylinders > 1023)       /* return special error on cylinder overflow */
+    /* return special error on BIOS cylinder overflow */
+    if (cylinder > 1023 && drivep->cylinders > 1023)
         return -1;
 
 #if RESET_DISK_CHG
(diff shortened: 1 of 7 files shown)


Mon0x4A:
@ghaerr Testing results for this PR have been posted in a comment in the original issue: https://github.com/ghaerr/elks/issues/2587#issuecomment-4773740490

Issue #2718, opened 2026-07-01
80286 protected-mode (CONFIG_286_PMODE): prototype boots to interactive shell - interest and guidance?


duzenko:
I've been prototyping an 80286 protected-mode HAL for ELKS and have it booting to an **interactive serial shell** with `fork`+`exec` of small-model userland (QEMU/KVM). Before investing in a proper PR I wanted to gauge interest and ask how you'd want it approached.

**Prototype branch / full diff:** https://github.com/duzenko/elks/pull/1

**Disclosure:** this is AI-assisted (Claude), with me reviewing/driving. Happy to follow whatever policy you have on AI-assisted contributions.

**What works (`CONFIG_286_PMODE=y`):** PM entry -> medium-model ftext far-calls -> IDT + live hardware IRQs -> serial console -> direct floppy driver with real 8237 DMA -> FAT root mount -> exec /bin/sh -> interactive shell -> fork+exec (e.g. `ls` runs and lists /bin).

**Approach:** the existing `seg_t` / far-memory abstraction is the HAL boundary - `seg_t` carries a GDT selector instead of a paragraph; `SEG_HANDLE`/`fmemcpy`/`peek`/`poke` resolve per-mode, so `fs/`, `kernel/`, and `net/` are almost entirely untouched. ~27 files: 3 new (seg286.h, pm286.c, pm_enter.S) plus machine-layer edits and a few spots in the portable kernel that reach through the abstraction (exec.c, fork.c, signal.c, buffer.c).

**Known limitations:** requires the XMS group off, L1-only buffers, CONSOLE_SERIAL; medium-model userland exec not yet supported (needs CONFIG_EXEC_MMODEL + PM-aware relocations).

**Questions:**
1. Is 286 protected-mode support something you'd want upstream, or does it cut across existing/planned design?
2. If so, how would you want it structured (it would clearly need splitting)?
3. Your stance on AI-assisted contributions?

Apologies if this duplicates existing discussion - happy to be pointed at prior work.

ghaerr:
Hello @duzenko,

[...]

PR #2719, opened 2026-07-02
[kernel] Guard fast serial IRQ handler against spurious QEMU interrupts

duzenko:
Please find the AI-generated PR description below.
Human note: this was originally a part of 286 branch, that I think belongs in mainline if correct.
Apparently there are/were three code paths (rs_irq/rs_fast_com1/serfast.S)
The asm path (the only one live) is lacking the uart_rx guard present in serial-8250.c
This complicates kernel debugging over serial port in my 286 port.
Let me know if this needs a better fix or it's not supposed to be fixed at all for some reason.

## Problem

On QEMU, at a ttyS0 login shell, the next command typed after a pipeline
frequently vanishes — no output, no error, next prompt appears:

    # ls /etc | wc -l
         12
    # echo ONE          <- silently discarded
    # echo TWO
    TWO

Reproduced on a stock ibmpc-1440 kernel (unmodified master). It is a race,
so it may take 2-3 pipelines to show. Repro: change /etc/inittab default
runlevel to 3 (getty on ttyS0), boot
`qemu-system-i386 -m 4M -nographic -drive file=fd1440.img,format=raw,if=floppy -boot a`,
log in on serial, run a pipeline, then any command.

## Cause

serfast.S's rx loop is a do-while: it reads UART_RX *before* checking LSR
data-ready. QEMU generates spurious IRQ4s (the disabled C reference handler
rs_irq() in serial-8250.c guards this exact case, with the comment "QEMU may
interrupt w/no data"). On a spurious interrupt the handler reads the empty RX
register and queues the resulting 0x00 into the tty input queue. A kernel-side
trace of the shell's read(0) shows a lone NUL arriving before the next
command's characters — ash discards the poisoned line silently. Pipelines
trigger it reliably because the children's tty writes generate extra
interrupt traffic.

Real 8250 hardware presumably never raises IRQ4 without data, so this only
bites emulator users (untested on pcem/86Box/real hardware).

## Fix

Add the same data-ready check rs_irq() has to the asm handler entry: read
LSR, and skip the receive loop when DR is clear (5 instructions).

Verified on QEMU/KVM: with the guard, post-pipeline commands execute
normally on both a stock real-mode kernel and the 286-PM fork (#2718),
and the stdin trace shows no NUL injection.

---
Found while stress-testing the 286 protected-mode fork (#2718). Diagnosis
and patch were Claude-assisted, per the discussion there; the change mirrors
the existing C reference handler and I'm happy to rework it however you
prefer.

### DIFF
--- a/elks/arch/i86/drivers/char/serfast.S
+++ b/elks/arch/i86/drivers/char/serfast.S
@@ -64,6 +64,11 @@ common_entry:
         mov     %cx,%bx                 // bx = sp = &ports[n]
         mov     (%bx),%si               // si = q = &sp->tty->inq
         mov     0x4(%bx),%dx            // dx = sp->io + UART_RX
+        add     $5,%dx                  // dx = sp->io + UART_LSR
+        in      (%dx),%al
+        sub     $5,%dx                  // dx = sp->io + UART_RX
+        test    $1,%al                  // data ready? QEMU may interrupt w/no data:
+        jz      4f                      // reading empty RX would queue a NUL as input
 0:      in      (%dx),%al               // do {
         mov     %al,%dl                 // dl = c = INB(sp->io+UART_RX)
         mov     (%si),%ax               // ax = q->len


ghaerr:
Hello @duzenko,

Thanks for the bug report. I understand it and can see why this could be happening, although I'm not able to get this to repeat here on my end. As the report details, this problem was previously seen in the older C serial interrupt handler because QEMU (apparently still) incorrectly generates spurious receive interrupts.

When rewriting the handler in ASM, I was hoping to achieve max throughput by not having to add five extra instructions to test the UART line status register for data received, but looks like this is necessary. I hope to somehow optimize this later.

Thank you!

duzenko:
> although I'm not able to get this to repeat here on my end

Do you run the commands from keyboard or feed a shell script via the serial port?

ghaerr:
I was typing at the keyboard - perhaps that was the issue. No worries, the bug needed fixing, and there's no telling whether real hardware may generate a spurious interrupt, so I'm good with this fix. Thanks.

PR #2722, opened 2026-07-03
[kernel] Implement 286 protected-mode kernel (CONFIG_286_PMODE)

duzenko:
@ghaerr Here's the requested PR per your comment in the fork
TBH I'd prefer to fix the kmem issue first in the branch but happy either way.

AI description below:

Builds on the pmode refactor (#2721, now in master). Adds a 286 protected-mode HAL: the kernel runs in 16-bit PM with a GDT/IDT, segments become GDT selectors, and the memory manager hands out descriptors — while the mainline real-mode path is untouched (everything behind `CONFIG_286_PMODE`, off by default).

Submitting the whole POC here per @ghaerr's request, for landing in master as a shared base.

**Status:** boots to a shell in 286 PM on QEMU/KVM — boot → directfd DMA → FAT root → init → getty → login → shell. Multi-user works including telnet login (ne2k → ktcp → telnetd → pty). Verified on both the full config and a trimmed ~60K single-segment kernel. All userland runs at ring 0 for now.

**What's here**
- **Mode bring-up:** `pm286.c` (GDT/IDT, real→PM switch), `pm_enter.S`, `seg286.h`.
- **Segment abstraction:** `segment.h`/`config.h` — `SEL_*`/`SEG_*`; `KERNEL_CS/DS`, `LINADDR`, `TRACKSEG` resolve to selectors in PM and paragraphs in real mode, so most code stays ifdef-free.
- **Memory manager:** `malloc.c` — `seg_pm_attach`/`detach` allocate a GDT descriptor per segment (`seg->base` = selector, `seg->para` = physical); no-ops in real mode.
- **IRQ:** `irq.c` (IDT gates via the data-as-code alias), `irq-8259.c`, `irqtab.S`.
- **Drivers/boot:** `directfd.c` (DMA via `desc_base`/selectors), `crt0.S`/`setup.S`, `exec.c`, `process.c`, `main.c`.
- **Build:** `ibmpc-286pm-small.config`, Makefile hooks, `.gitattributes`.

**Known / deferred (per discussion)**
- `/dev/kmem` PM decode (`mem.c`) is a partial bias-strip for ps/meminfo's task-table reads; the argv/`COMMAND` case isn't invertible with selectors. Left in for now — glad to drop it, since you noted you may leave it out until the selector-aware protocol lands.
- `mkimg.sh` included per your note; a test-harness convenience (injects a fresh kernel into a prebuilt image), not needed for from-scratch builds.
- procfs / ps rewrite deferred (agreed — after 0.9.2 if at all).
- Backlog in `PMODE-TODO.md` (ring 3, >1MB/XMS, INITPROC-in-small-model, ifdef audit).

Rebases cleanly on master — picked up your `ecdf027d` malloc fix; our matching external-selector edits folded into it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Kernel config files are parsed by shell/awk build tooling that anchors on
+# LF; keep them LF in the working tree even when core.autocrlf is set (Windows).
+*.config text eol=lf
--- a/PMODE-TODO.md
+++ b/PMODE-TODO.md
@@ -1,0 +1,84 @@
+# 286 protected-mode port — backlog
+
+Status: full multi-user works on QEMU/KVM — boot → directfd DMA → FAT root →
+init → /etc/rc.sys → getty → login → shell; meminfo/ps/uptime/uname/df/date
+work. Proven on both the full (far-text) config and a trimmed 55K
+single-segment (far-text-off) kernel. All userland currently runs at ring 0.
+
+## Bugs / broken
+
+- ~~Shell pipes wedge the shell~~ FIXED upstream (PR #2719, merged): not
+  pipes at all — serfast.S read UART_RX before checking LSR_DR, so QEMU's
+  spurious serial IRQs queued a NUL into the tty input, and ash silently
+  discarded the poisoned next command line. Entry guard added, matching the
+  C handler. Confirmed on stock real-mode ELKS (not a PM-port artifact);
+  the fix now arrives via upstream master.
+- ~~telnetd faults EXC=0005 at startup~~ RESOLVED: another fragmentation-era
+  phantom — the pty driver needed no PM fix at all. With CONFIG_PSEUDO_TTY
+  enabled, **telnet login into 286 PM works end-to-end** (client → QEMU
+  hostfwd → ne2k → ktcp → telnetd → pty → login → shell) on the small
+  single-segment kernel (text 60624). The original roadmap goal is complete.
+
+## Degraded / deferred
+
+- **ps COMMAND column: blank or kernel-string garbage.** Cmdline reads pack
+  another process's SS as (seg<<4)+off — not invertible with selectors. The
+  kmem_read bounds check (16ad9d65) rejects positions outside the
+  (KERNEL_DS<<4)+64K window, but with KERNEL_DS=0x10 the window starts at
+  linear 0x100, so many selector-packed positions land INSIDE it and read
+  valid-but-wrong kernel data — pty-session processes' SPs collide, so ps
+  prints kernel rodata strings as COMMAND (cosmetic; system unaffected).
+  Real fix: selector-aware /dev/kmem protocol (e.g. ioctl taking selector +
+  offset) + rebuilt ps — blocked on userland rebuilds.
+- ~~Userland cannot be rebuilt~~ FIXED: `make -C libc all && make -C libc
+  install DESTDIR=/usr` installs elkslibc into the PPA toolchain
+  (/usr/ia16-elf/lib/elkslibc) and provides the runtime specs that
+  `-melks-libc` needs; apps then build normally (verified: ps links and
+  runs). Unblocks: /proc-based ps rewrite, test programs, kmem protocol
+  work.
+- **procfs (CONFIG_PROC_FS) prototype landed** (41725883): /proc/uptime +
+  /proc/pid/{stat,cmd}, automounted, ~2.5K text, correct command names in
+  PM via kernel-side selector reads. Next: rewrite ps to read /proc
+  (kills the kmem COMMAND garbage + the struct-version coupling), then
+  the upstream design issue (user is raising it).
+- **Medium-model user executables** (FTEXT != 0): need CONFIG_EXEC_MMODEL +
+  selector-aware relocate() + a per-process second code descriptor. No known
+  binary on the image needs it; telnet path is all small-model.
+- **/bootopts not found** on hand-built images unless the file sits in the
+  first 8 root directory entries (setup.S scans only those). mkimg.sh omits
+  /bootopts; defaults work. Robust fix: read /bootopts post-root-mount.
+
+## Design debt
+
+- **Ring 3 / protection.** Everything runs at DPL0, so PM currently adds no
+  memory protection. Needs DESC_U* descriptors, a TSS for the ring3→ring0
+  stack switch, and _irqit/restore_regs rework for the hardware ring switch.
+- **Extended memory (>1MB).** XMS group is disabled; kernel uses only the
+  639K conventional pool (it is tight: kstack bump once OOM'd buffer_init).
+  On 286, >1MB access in PM is natural via descriptors — a PM-native buffer
+  cache above 1MB would relieve the main pool.
+- **#ifdef reduction / upstream convergence** (ghaerr, issue #2718): replace
+  raw CONFIG_286_PMODE ifdefs with mainline-friendly abstractions in the
+  KERNEL_DS/BIOSSEG/VIDEOSEG style. Done so far: KERNEL_CS/KERNEL_DS in
+  mem.c. Audit remaining ifdef sites.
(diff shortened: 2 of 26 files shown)


ghaerr:
> TBH I'd prefer to fix the kmem issue first in the branch but happy either way.

Can you describe a bit more what that fix would be? I'm trying to keep things somewhat separated so that we don't have too many changes in one PR. I haven't fully reviewed your kmem changes in this PR but was willing to accept them on the basis they worked. Do we need a lot more changes to make `ps` work? If so, I would rather they go in a separate PR, since some/many of those changes would be in the ps/meminfo applications. Right?

> the argv/COMMAND case isn't invertible with selectors. Left in for now — glad to drop it, since you noted you may leave it out until the selector-aware protocol lands.

Had lots going on the last few hours, haven't had time to review the actual problem relating to the argv inspection w/selectors. So I'm not yet much help in suggesting a fix or knowing what needs to be done.

duzenko:
I've just heard about kmem from AI first time this week. But I will reply as human rather than copy pasting the AI generated 'reasoning'.

> Can you describe a bit more what that fix would be?

The real fix for user-level prorgams reading from kernel memory in protected mode? Not read kernel memory at all.
The changes that you see in the PR is 'translating' from what `ps` expects to read in `kmem` vs. what it would (crash trying to) read from it in PM. It almost works - when it's readin `kmem` only. But then it also tries to read from other process memory too - specifically `argv` and it's just too much effort to try and fix.

So my honest opinion - add `procfs` first _before_ this PR - real-mode compatible, optional, cheap because kmem is compiled out, already tested in both modes. That will also remove the need to change `mem.c` at all.

Or, suggest your alternative fix.

The problem with this PR in this master state is that in PM it breaks programs that read from other programs memory - argv reading by `ps` is one but possibly not only case.
Technically it's not breakage as it does not affect real mode - but it makes the PR incomplete/half-working.

I'm not familiar with how `ps` hunts to running process `argv` but it does involve segment-to-linear computations. I assume part of alternative fix is returning SS of each process as paragraph instead of selector but AI for some reason shows aversion to this suggestion.

ghaerr:
> add procfs first before this PR - real-mode compatible, optional, cheap because kmem is compiled out, 

I already mentioned there are other issues with having to mount filesystems to do something like run `ps`, along with other concerns. So that's out for now. 

There are a number of other programs and C library routines that read /dev/kmem, so it's not optional at the moment. Admittedly, things could be done with a better API, but were purposely designed to keep the kernel size down. The kernel is within ~1500 bytes of running out of code space for certain platforms/options. Your comments bring up good points though, and I now feel that there at least ought to be C library wrapper routines that hide the details of how these things work. Then, if you have to have procfs, you could also write an alternative C library wrapper routine that then makes rewriting ps/meminfo yet again unnecessary. And one of the reasons I wanted to get this PM base port started in master is to identify the non-standard or kluge areas of ELKS so that it can be improved. But that needs to be done with knowledge of all the platforms involved and the fact there are competing technical interests amongst ELKS ports, especially for running in very small (e.g. 256k) systems. (A recent video review of ELKS mentioned they couldn't get ELKS to run in 256k, which was then fixed: ELKS will boot unmodified in a PC w/256K RAM from the distribution disk image).

This is just one of what will be ongoing issues that different people might call a "working" PM port. There's other reasons to keep /dev/kmem and the whole discussion gets larger than this PR because you're not concerned with running ELKS on various other platforms, including ROM, where the VFS filesystem code may not be included. Best to let you run free in your fork, then you won't be impeded; I'm just trying to get a base version running so that I and others can start thinking more clearly about the issues involved with a full PM port, as previously discussed.

> Technically it's not breakage as it does not affect real mode - but it makes the PR incomplete/half-working.

I already stated we wouldn't try to add everything in one PR to get a "full" PM port working. Many other things don't work in the PM port at this point, including EXT buffers, decent-sized kernel free heap, XMS, INITPROC-freed kernel sections, etc etc. I see the static 4K GDT residing in the kernel data segment as a major problem, it should probably be allocated from main memory at boot along with the 2k IDT. That can all come at a later time.

> it's just too much effort to try and fix.

Not really, now that I think I (finally) understand the issue.

> I'm not familiar with how ps hunts to running process argv 

`ps' just gets the segment address for the process data segment from the task structure (which is read from /dev/kmem), then calculates the beginning stack location, then seeks there and reads the argv command line from /dev/kmem. The only problem is that the task structure now contains selectors rather than segments for the task data segment. *All* the other code in ps/meminfo remains the same, no big rewrite is required, and its not too much effort.

> Or, suggest your alternative fix.

A potential solution is simple: make the physical address of a process's data segment be retrievable; in real mode, this is always the saved SS, since SS == DS for application execution. 

> part of alternative fix is returning SS of each process as paragraph instead of selector 

Yes - each process's data segment (e.g. dataseg->phys) could be stored in the task structure, or the offset of the global GDT in the kernel data segment could be returned, allowing for any selector to be converted (with a library routine for simplicity). There are issues with each approach, and I'd like to think about it a bit.

> It almost works - when it's readin kmem only.

Yes, this is a little strange, in that /dev/kmem isn't actually kernel memory, its physical memory.

Summary: I'll take a look at the /dev/kmem fixes and also push some way to get the task data segment physical address. This will then allow for ps/meminfo to be easily made working, which can be done immediately after this PR by your AI if desired (or I can do it).

Thank you!

duzenko:
@ghaerr 
Sorry for the extra round trip
AI did figure out a clean fix after all
Please see the 2nd commit - it includes ps/meminfo source code too
I can split it to a separate PR if you prefer that way

ghaerr:
> Please see the 2nd commit
I can split it to a separate PR if you prefer that way

Let me take a look and I'll comment after, thanks.

ghaerr:
> Let me take a look and I'll comment after, thanks.

I don't fully understand it yet, but I see now that the issue involves the 32-bit "seek offset" sent to /dev/kmem, and its interpretation as a real mode seg:offset versus a selector:offset in protected mode. The code is relatively uncomplicated, it appears we don't need to worry about stuffing a physical address of the task data segment at all, as the selector for that segment is now passed as the upper 16-bits of the seek offset. Kind of ingenious in a way, but then I may have forgotten that /dev/kmem was interpreting the file position as a seg:offset rather than a straight linear address into physical memory.

It needs some cleanup, but probably should leave it in this PR so things work for you right after commit. I will plan on committing after some more inspection.

Thank you!

duzenko:
> then I may have forgotten that /dev/kmem was interpreting the file position as a seg:offset rather than a straight linear address into physical memory.

No, encoding did change and new kernel and clients need each other to work
We might want to switch encoding with a special `ioctl` call in new clients

ghaerr:
> No, encoding did change and new kernel and clients need each other to work

I just figured that out. This means I now need to look at any/all other elkscmd/ commands and C library code that reads /dev/kmem to make sure they still work.

> We might want to switch encoding with a special ioctl call in new clients

I don't want to add that complexity at this time, things need to remain simple.

> TBH I'd prefer to fix the kmem issue first in the branch but happy either way.

I like the kmem fix, but frankly it might be better to split the PM vs kmem/ps/meminfo changes into two PRs at this point. Otherwise I've got to look through all of libc/ and elkscmd/ and check things, thus complicating this first PR, something I wanted to avoid.

I was pretty much ready to commit this first PR, but with /dev/kmem access changes, committing this now makes `ps` work for real mode/PM, but breaks all other programs that use /dev/kmem - not good, really.

ghaerr:
> [with this change] ps works for real mode/PM, but breaks all other programs that use /dev/kmem

OK, I think I've figured out a way to interpret the /dev/kmem file offset specially for `ps` and `meminfo` (as selector:offset or segment:offset) and keep compatibility with the normal interpretation as a linear offset for all other programs. This would be similar to your `ioctl` idea but using an `open` flag since ioctl would require additional state.

In the interests of moving this forward so I can start testing, I plan on committing this now, followed by an additional PR along with some additional cleanups as mentioned previously.

PR #2723, opened 2026-07-04
[kernel] Cleanup protected mode kernel code

ghaerr:
Follow-on PR to #2722 cleans up various aspects of new protected mode kernel.

Rewrites and simplifies /dev/kmem device driver to be backwards compatible unless new O_ALT flag used in open().

A few more changes will be pushed in a subsequent commit.


duzenko:
Why `CONFIG_APPS_COMPRESS ` disabled in ibmpc-286pm-small.config? Programs on the typical ibm-pc image will not start without it.
Or the entire program files need uncompressing in the image first?

ghaerr:
> Programs on the typical ibm-pc image will not start without it

What do you mean? They all run, the option only saves space on the target image, but takes much longer to create.

duzenko:
You mean you re-generate the entire image on each run including re-compressing programs? Why?

ghaerr:
See image/Make.image for details on how the images are built. The target image directory tree is built from scratch each time since elkscmd/rootfs_template/ may have changed; the applications themselves in elkscmd/ use normal Makefile rules and aren't rebuilt, but re-installed using `install` each build since some may have more complex installation requirements.

CONFIG_APPS_COMPRESS was added to save space on floppies and is required for some distributions, but in general users have complained that the decompression time from floppy isn't worth it on 8088 machines.

PR #2725, opened 2026-07-04
286pm ext buffers

duzenko:
This closes one of the protected mode gaps - EXT buffer.
Two commits: one fixes it in the conventional memory and the other moves it to high memory.
Can be split in two PRs or used for research of alternative solution.
No hurry to review.

### DIFF
--- a/elks/arch/i86/boot/setup.S
+++ b/elks/arch/i86/boot/setup.S
@@ -931,6 +931,12 @@ arch_set_initseg:
 	jc	1f
 	mov	%ax,xms_kbytes
 1:
+#ifdef CONFIG_286_PMODE
+	// PM reaches extended memory (>1MB) through GDT descriptors, so A20 must be
+	// enabled or a >1MB physical access wraps back into low memory and corrupts it.
+	// The HMA path only enables A20 for hma=kernel; do it unconditionally for PM.
+	call	enable_a20_gate
+#endif
 #ifdef CONFIG_BOOTOPTS
 	call	bootopts	// load /bootopts into DEF_OPTSEG (0050:0000)
 	call    hmabootopts	// check /bootopts for hma=kernel
--- a/elks/arch/i86/mm/pm286.c
+++ b/elks/arch/i86/mm/pm286.c
@@ -11,6 +11,8 @@
 #include <linuxmt/config.h>
 #include <linuxmt/types.h>
 #include <linuxmt/kernel.h>
+#include <linuxmt/memory.h>     /* setupw() -> SETUP_XMS_KBYTES */
+#include <arch/io.h>            /* inb/outb for the A20 gate */
 #include <arch/seg286.h>
 
 #ifdef CONFIG_286_PMODE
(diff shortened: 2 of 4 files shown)


ghaerr:
Hello @duzenko,

Let's keep this open for a while, while the initial PM port is better integrated, and more time can be spent reflecting on proper kernel design for it and additional enhancements like this for PM. This PR contains all sorts of problems: ifdefs, duplicated allocation routines, non-portable A20 code, etc.

I potentially like the idea of using AI for research ideas, but in general think building a kernel using AI isn't a great idea, unless you're just doing it for yourself. The tool excels at reading entire source trees so it can quickly produce a to-the-point solution that can work, but as you can see, almost always just produces \#ifdefs shoe-horning its tightly-defined solution into the source tree. Very kludgy and a maintenance nightmare. Proper kernel design will take a lot more than that. For instance, this PR contains completely non-portable inb/outb instructions to enable the A20 gate right in the middle of a buffer routine (very bad design), despite the fact that there's already very specialized and highly tested A20 code already available elsewhere in the system (that analyzation path was likely eliminated early on until the tool later realized it required A20 and didn't perform another top-down analysis, etc).

I imagine the day is here where non-programmers can create large enhancements quite quickly, but from the code I'm seeing, it appears to be just more and more code piled onto a system. For already-large 32- and 64-bit systems with large flat address spaces and seemingly infinite space for kernel size and applications, this isn't a big problem, but has the potential to turn into spaghetti code, almost requiring an AI to maintain it. We will see how that works out in the years to come, for other systems. For ELKS, the problem is in reverse, where resources are extremely limited, and lots of thought as well as a good philosophy has to be present in order to have the system work well, but remaining within its (16-bit) limitations. Do a lot with very little, definitely old-skool. IMO, this philosophy should not change just because the system gets protected mode operation and suddenly has tons more space to operate in.

For non-mission-critical, or ordinary applications written with AI, I think the problem is much less severe, as the application serves as the entire container for any problems described above. This in many ways is similar to all applications in the wild: the range of design goodness covers the entire span from ugly to beautiful, well-designed to terrible, but applications are quite compartmentalized and relatively easy to replace (at least for command line apps).

In summary, this PR gives some potentially interesting ideas for how to extend the kernel buffer system to XMS completely contained within ifdefs. An actual solution will require a rethink of the buffer system such that no ifdefs are required and appropriate wrapper functions are linked, depending on the configured options and architecture.

Thank you!

ghaerr:
Hello @duzenko,

Thank you for posting this. 

> as long I'm not the one reviewing its "output" :)

As you could probably tell from my post above, it's easy to get overwhelmed trying to keep up with AI output, especially in PR form :) 

While AI's actual code may sometimes not be very fitting for its intended purpose, I'm finding that they can still be thought provoking. In this case, I've realized we should be able to fairly easily be able to use most all of the existing ELKS XMS code in our new PM mode for buffers and other XMS access, like RAM disks.

Currently, ELKS XMS uses 80386 'unreal' mode or the 80286 LOADALL instruction to gain CPU access to XMS memory. For a PM kernel a GDT and protected mode is setup and enabled directly after boot, but the XMS initialization could still be configured and executed, and then use the standard XMS execution path for XMS buffer access. The magic then happens in an enhancement to `xms_fmemcpy`, where currently `linear32_fmemcpy` is used. `linear32_fmemcpy` works in 386+ unreal mode by previously setting a GDT descriptor to all of main and XMS memory, then using the passed 'long offset' to rapidly copy a memory block using 'addr32' prefixes to 16-bit instructions, which is very fast.

For the PM kernel, a replacement `pm_linear32_fmemcpy` or possibly `int15_fmemcpy`  could update a single GDT entry and use its selector to quickly access any memory location, which would also be very fast. The existing 286 XMS code uses `int15_fmemcpy` which updates an internal GDT before calling BIOS INT 15h. This same routine could be mostly used for PM, and then do an internal fast block copy rather than calling INT 15, which transitions from real mode <-> PM and back.

Protected mode 80286 segments are limited to 64K bytes, while PM on 386+ allows for 4G. While using a single selector for all of memory could be appealing, that's potentially less secure, can't work the same way on 80286, and wouldn't use the same path as the existing XMS code, where the XMS_INT15 path already updates a (different) GDT.

The above method is probably better than this AI-PR's solution, which allocates as many selectors as there are EXT or XMS buffers. You may be aware that ELKS can be configured to use 2500 XMS buffers, so that'd be a huge drain on the `desc_alloc` GDT allocator requiring 20K bytes for the GDT in the kernel data segment (which was originally set to 512 selectors max). (BTW, the latest cleanup decreased MAX_GDT_ENTRIES from 512 to 100, which reduced the kernel data segment requirement from 4K to 800 bytes, and it seems to work well running networking, etc).

I'll start work on integrating ELKS XMS to PM. Perhaps 0.9.2 should be released beforehand. I enjoy writing kernel enhancements by hand, but appreciate your posting AI "ideas" (even as code or PRs) to start the process of old-fashioned human thought on a subject.

Thank you!

ghaerr:
Completed in #2729.

PR #2729, opened 2026-07-07
[kernel] Implement EXT buffers and XMS in PM kernel

ghaerr:
Enhance PM kernel to allow operating with CONFIG_FS_XMS and CONFIG_FS_EXTERNAL_BUFFER options set.

By default, the PM kernel operates with XMS buffers, unless xms=off in /bootopts. In that case, a fallback to EXT buffers is implemented, although this requires allocating as many selectors as there are EXT buffers. MAX_GDT_ENTRIES may have to be incremented from 129 in limits.h.

The XMS ram buffer is also implemented. XMS and MINIX filesystem image are now set by default in ibmpc-pmode.config.

In general, the PM kernel should always specify CONFIG_FS_XMS, since XMS memory is almost always available on protected mode capable machines. If EXT buffer operation is perferred, a new option xms=off can be set in /bootopts to force EXT buffer operation.

It was discovered that the ATA/CF driver doesn't work with XMS enabled in the PM kernel. This is now fixed in an additional commit to this PR. The PM kernel will then be operational with XMS or EXT buffers with the direct floppy and ATA/CF hard disk driver. After that, #2725 can be closed as completed.

The next step will be more cleanup in the CONFIG_286_PMODE configuration, including some renaming since the option actually works on 80286, 80386 and later Intel 86 machines, when keeping segment sizes <= 64K. This allows the PM kernel to be used on most all non-8088/8086 IBM PCs. I think, but have not yet tested, that the PM kernel should also work on PC-98 (with the exception that the PC-98 BIOS can't be called from protected mode, so the direct FD driver may have to be enhanced to work on PC-98 hardware).


ghaerr:
A number of boot-related fixed segments assigned in config.h were renamed in this PR, to make it easier track the changes required for selectors required for various segments, as well as easier understanding of the kernel. Now all start with the SEG\_ prefix. 

The following changed:
DEF_OPTSEG -> SEG_OPTSEG (/bootopts options segment)
REL_INITSEG -> SEG_INITSEG (setup.S data segment for `setupb/setupw` calls)
REL_SYSSEG ->SEG_SYSSEG (relocated load address of kernel by setup.S)

PM kernel now works with DF and ATA/CF drivers. Regression testing for IBMPC, PC-98 and ROM builds on QEMU and EMU86.

PR #2732, opened 2026-07-08
Elks parallel compress

djohanse:
Opencode implementation of running elks-compress in parallel.  I know compress only takes a few seconds to run so not a great ROI but worth a look

Sanity check that elks builds/runs:

Screenshot 2026-07-07 at 11 11 52 PM
Screenshot 2026-07-07 at 11 14 03 PM


ghaerr:
Thanks @djohanse, this speeds up the compressed builds by quite a bit!

PR #2735, opened 2026-07-10
[kernel] Fix BIOSDATA descriptor limit (bytes, not paragraphs) — PM t…

### LINKED ISSUE #2734: PM kernel triple-faults under QEMU/KVM on current master (boots fine under TCG)
**Description**
Expected: Master build runs in QEMU with KVM (nested VT-x)
Observed: not only QEMU crashes, but takes the entire Linux host with it
Image

**Configuration**
- current master
- QEMU in AlmaLinux docker container

**How to reproduce ?**
- Does the problem always happen ? yes
- List the steps to reproduce the problem. see below

**Raw data**
- Screenshot(s): see above
- Log file:

**Additional information**
AI generated

> The protected-mode kernel (CONFIG_286_PMODE=y) built from current master (2c8c71a9) triple-faults during early boot when run under QEMU with -accel kvm, but boots normally under -accel tcg. A PM build from a pre-PM-rework base boots under the same KVM setup, so this looks like a regression introduced by the recent PM series (IDT→0:0, XMS/EXT in PM, descriptor rework, fartext).
> 
> Environment
> - Kernel: master @ 2c8c71a9, config CONFIG_286_PMODE=y, IBMPC target, CONFIG_BLK_DEV_FD (directfd), scancode keyboard, direct+serial console.
> - Image: standard 1440K FAT floppy via mkimg.sh.
> - QEMU: qemu-system-i386 (both accels tried).
> - KVM host: Intel, kvm_intel unrestricted_guest=Y. (Observed nested inside VirtualBox with nested-VT-x on; there the guest triple-fault surfaces as a VBox "Guru Meditation" VINF_EM_TRIPLE_FAULT.)
> 
> Steps to reproduce
> 1. Build a PM kernel from master: CONFIG_286_PMODE=y, make.
> 2. mkimg.sh <Image> pm.img 3 → 1440K FAT floppy.
> 3. Boot under TCG:
> qemu-system-i386 -accel tcg -m 4M -nographic -drive file=pm.img,format=raw,if=floppy -boot a
> → boots to login normally.
> 4. Boot the same image under KVM:
> qemu-system-i386 -accel kvm -m 4M -nographic -drive file=pm.img,format=raw,if=floppy -boot a
> → triple-faults during early boot (guest resets; under nested-VBox this crashes the host VM).
> 
> Expected: boots to login under -accel kvm, same as -accel tcg.
> 
> Actual: early-boot triple fault under -accel kvm only.
> 
> Additional data
> - A PM build on the pre-PM-rework base (before the pmode.c/IDT-to-0:0/XMS-in-PM series) boots under this same KVM without issue → points the regression at the recent PM commits (candidates: 5e2d5236 "0:0", 55a11b3c "EXT buffers + XMS in PM", d6e480de/c929cc58descriptor/A20 rework, fe4d4170 PM fartext).
> - Fault occurs before any console output on the KVM path, sred — consistent with an early PM-entry / IDT / descriptorissue that accurate CPU emulation (KVM) enforces but TCG tolerates.
> 
> Question for maintainer
> - Does current PM master boot for you under bare-metal -acctriple-faults there too, a git bisect across the PM seriesagainst a KVM boot should pinpoint it. If it's KVM-clean on your setup, the issue may be the IDT-at-0:0 / real-mode-IVT overlap being handled differently under KVM.
> 
>

duzenko:
…riple-faults under VT-x

Human note: supposed to fix https://github.com/ghaerr/elks/issues/2734

AI description below:

> Summary
> 
> The protected-mode kernel triple-faults during early boot under hardware virtualization (VirtualBox / QEMU-KVM / VT-x), while booting normally under QEMU-TCG. Root cause is a one-line unit mismatch: a GDT descriptor is given a limit in paragraphs where bytes are now expected.
> 
> Root cause
> 
> Commit d6e480de ("Prepare for > 64K future selector limits") changed desc_set() to take a byte limit instead of a paragraph count, and updated the callers accordingly — except GDT_BIOSDATA in gdt_init(), which still wraps its size in BYTES_PARA(256):
> 
> desc_set(MK_SEL(GDT_BIOSDATA, ...), SEG_BIOSDATA << 4, BYTES_PARA(256), DESC_KDATA);
> 
> BYTES_PARA(256) evaluates to 256 >> 4 = 16, so the BIOS Data Area segment gets a 16-byte limit instead of 256.
> 
> console_init() reads the 80-column word at BDA offset 0x40:0x4A via peekb(0x4A, GDT_BIOSDATA). Offset 0x4A (74) is past the 16-byte limit:
> 
> - On hardware VT-x, the segment limit is enforced → #GP → the kernel triple-faults right after ttyS0 ..., before the PC/AT class cpu banner.
> - On QEMU TCG, segment limits aren't checked, so the read silently returns the correct byte and the kernel boots — which is why this only manifests on hardware-virtualized targets.
> 
> Fix
> 
> Pass the limit in bytes, consistent with the other descriptors updated in d6e480de:
> 
> desc_set(MK_SEL(GDT_BIOSDATA, ...), SEG_BIOSDATA << 4, 256, DESC_KDATA);
> 
> How it was found
> 
> git bisect between a5b7da49 (good) and 2c8c71a9 (bad), boot-testing each step in a VirtualBox VM on VT-x → first-bad-commit d6e480de. Fault frame decoded to peekb called from console_init with selector GDT_BIOSDATA, offset 0x4A.
> 
> Testing
> 
> - Before: current master → VirtualBox Guru Meditation (VINF_EM_TRIPLE_FAULT) / KVM triple fault; TCG boots.
> - After: boots to login on VirtualBox VT-x (mounts root, runs /etc/rc.sys); TCG unaffected.
> - GDT_BIOSDATA was the only remaining BYTES_PARA()-wrapped descriptor limit in gdt_init().

### DIFF
--- a/elks/arch/i86/mm/pmode.c
+++ b/elks/arch/i86/mm/pmode.c
@@ -161,8 +161,15 @@ void gdt_init(void)
     desc_set(MK_SEL(GDT_KCODE,  SEL_GDT, SEL_RPL0),
         (addr_t)kernel_cs << 4, (unsigned)_endtext, DESC_KCODE);
 
+    /* Full 64K limit: gdt_init runs before setup_arch() computes membase,
+     * so (membase - kernel_ds) << 4 here would be garbage (membase still 0),
+     * truncating the limit and #GP-faulting the first heap access above it
+     * on hardware that enforces limits (VT-x/KVM; QEMU TCG doesn't check).
+     * setup_arch() extends the kernel data segment to a maximum of 64K for
+     * the near heap anyway, matching the KDATA_EXEC alias below.
+     */
     desc_set(MK_SEL(GDT_KDATA,  SEL_GDT, SEL_RPL0),
-        data_base, (membase - kernel_ds) << 4, DESC_KDATA);
+        data_base, 65536L, DESC_KDATA);
 
     if (kernel_ftext) {
         desc_set(MK_SEL(GDT_KFTEXT, SEL_GDT, SEL_RPL0),
--- a/elks/arch/i86/mm/pmode.c
+++ b/elks/arch/i86/mm/pmode.c
@@ -190,7 +190,7 @@ void gdt_init(void)
     desc_set(MK_SEL(GDT_OPTSEG, SEL_GDT, SEL_RPL0), SEG_OPTSEG << 4, 1024, DESC_KDATA);
 
     /* BIOS data area */
-    desc_set(MK_SEL(GDT_BIOSDATA, SEL_GDT, SEL_RPL0), SEG_BIOSDATA << 4, BYTES_PARA(256),
+    desc_set(MK_SEL(GDT_BIOSDATA, SEL_GDT, SEL_RPL0), SEG_BIOSDATA << 4, 256,
         DESC_KDATA);
 
     /* text video memory */


duzenko:
Regarding the 2nd commit:

> gdt_init() sizes KDATA from membase, which setup_arch() hasn't set yet — 16-bit math truncates the limit to ~37K and the heap #GPs on VT-x/KVM. Use the full 64K.

ghaerr:
Hello @duzenko,

Thank you so much for testing and finding this!! I am sorry if my recent heavy integration-related changes caused you to spend a lot of time tracking these down. I really appreciate your having found the solution, especially since I now realize my version/OS of QEMU can't check descriptor limits.

> On hardware VT-x, the segment limit is enforced → #GP → the kernel triple-faults 

I am not yet understanding why a triple-fault occurs, given your initial code went to some lengths to prevent a triple fault by catching all exception vectors and redirecting them to a fault handler which displays the fault info. My C rewrite of the same was also tested and should be doing the same... a triple fault should never occur, instead an EXC= panic should display. 

> triple-faults during early boot when run under QEMU with -accel kvm, but boots normally under -accel tcg. 

Unfortunately, I am unable to run (on macOS) with -accel kvm, and running the equivalent of -accel tcg (-singlestep) doesn't fault. This means that my QEMU setup is not a good emulator for development and testing of PM, and this why  my errors in the last few commits went undetected.

I will try getting VirtualBox installed to ensure better regression testing.

> gdt_init() sizes KDATA from membase, which setup_arch() hasn't set yet

Oops, great catch. I am thinking gdt_init can possibly be enhanced to run in two separate passes, and also move most of that code into an INITPROC segment so it can be discarded after initialization. I will work on that.

Regarding the move to allowing PM descriptor limits to be set to exact byte sizes, I wanted to better ensure system integrity by setting all kernel descriptors to their exact limit. For example, the IVT is now 129 entries long (=1032 bytes) which isn't a paragraph boundary. Likewise for the kernel data segment, which may not always be 64K. I missed thinking about the fact that membase isn't initialized yet, but will also fix this later. 

Thank you!

duzenko:
@ghaerr I'm afraid you need hardware x86 MMU for these crashes to occur. Your Mac has Apple CPU?
How about renting a (cheap) x86 hosting with root and virtualization?

ghaerr:
> I'm afraid you need hardware x86 MMU for these crashes to occur. Your Mac has Apple CPU?

My Mac has an x86 CPU, but -accel kvm isn't supposed on macOS.

I have other emulators but haven't set them up for testing PM yet. I also want to understand why this most recent problem triple-faulted rather than was caught using the exception handler vectors. A triple-fault can occur if no vector is set, and then the vector which handles that case is also not set (which should not be the case). I will read up more on it and also try other emulators.

duzenko:
AI says that a (second) crash happens inside the exception handler

> `-accel kvm` is Linux-only, but on an Intel Mac QEMU supports **`-accel hvf`** (Hypervisor.framework) — that's real hardware virtualization and enforces descriptor limits exactly like KVM/VT-x does. So you should be able to reproduce this class locally with just `qemu-system-i386 -accel hvf ...`. (Alternatively 86Box/PCem emulate a real 286/386 cycle-accurately including limit checks, with no hardware dependency at all — they work on ARM Macs too.)
> 
> On why it triple-faulted rather than being caught: the exception vectors *were* set — the #GP was delivered and the handler ran. The escalation came from two things conspiring, both fixed in this PR:
> 
> 1. The #GP fired inside `console_init()`, and the fault handler's output path goes through the console code — which itself reads the BIOS Data Area through the same broken 16-byte `GDT_BIOSDATA` descriptor. So the handler re-faulted, that handler re-faulted, and so on: an unbounded #GP recursion, each level pushing another exception frame plus handler frames onto the stack.
> 
> 2. Each recursion level eats stack. When SP eventually wrapped below 0 to 0xFFFE, the next exception-frame **push** landed above the KDATA limit — which was *also* broken at the time (the `membase` truncation from the second commit, ~0x9470 instead of 64K). A fault while pushing an exception frame can't be delivered on that stack → #DF; delivering #DF needs the same stack → fails again → triple fault.
> 
> So it's not that a vector was missing — it's that the fault *reporting* path depended on the two broken descriptors, and once the stack wrapped, exception delivery itself became impossible. That's also why post-fix the later, unrelated #GPs print clean `panic: EXC 000d ...` lines: with sane descriptors the handler completes on the first try.
>

ghaerr:
> AI says that a (second) crash happens inside the exception handler

Thanks for the explanation, yes, I think that is likely what was happening. I want to further explore changes so that the PM kernel exception reporter always works. More coming on that when I have a bit more time.

Thank you!

PR #2737, opened 2026-07-10
CI: boot-smoke-test ELKS under qemu-kvm (hardware VT-x)

duzenko:
Following up on your comment that you can't test MMU crashes on MacOS, AI suggested to run build + run to login tests on github infrastructure that seemingly runs on x86 and supports VT-x. It seems to correctly fail on master, caused by hidden bugs in PM (see run reports in my fork).

AI description below:

Summary

Adds a workflow that builds ELKS and boots it under qemu-kvm (hardware VT-x via the runner's nested virtualization), grading the result from the serial console.

QEMU TCG does not enforce segment descriptor limits; KVM — like real silicon — does. The bug class fixed in #2735 (GDT_BIOSDATA/GDT_KDATA limits: boot triple-faults on VirtualBox/KVM while TCG boots fine) is therefore invisible to TCG-based testing, and to any macOS/ARM host. Runner-hosted KVM was verified to enforce limits identically to direct VT-x.

What it does

- Reuses the same cached cross/ toolchain as main.yml (first run builds it, ~20 min; cached runs ~5 min).
- Matrix over two configs: ibmpc-1440 (real-mode baseline) and ibmpc-pmode (286 protected mode — the limit-enforcement-sensitive build).
- Uncomments the console=ttyS0,19200 3 bootopts line in the test image so a serial getty makes the outcome observable.
- Boots image/fd1440.img with qemu-system-i386 -accel kvm and grades the serial log:
  - login: → pass
  - panic → fail
  - 150 s of neither → fail (hang, or triple-fault reset loop)
- Serial log is printed in the job output and uploaded as an artifact on failure.

First-run results (on my fork)

- ibmpc-1440: boots to login under KVM on a GitHub runner.
- ibmpc-pmode: fails at root mount — this is the pre-existing PM+XMS-under-hardware-virt issue (boot reaches df0: floppy type 1.44M after the xms banner, then dies; same image boots to login with xms=off). The job is marked continue-on-error so it reports as a warning rather than blocking, with a comment to flip it to gating once that issue is fixed. I can file the issue with the serial evidence separately.

Notes

- /dev/kvm needs an explicit chmod on the runners; the udev-rule-only recipe races (first run demonstrated it).
- Triple faults under -no-reboot leave a nearly-empty serial log — still a red X via the timeout path, which is the point.

### DIFF
--- a/.github/workflows/kvm-boot-test.yml
+++ b/.github/workflows/kvm-boot-test.yml
@@ -1,0 +1,114 @@
+name: kvm-boot-test
+
+# Boot-smoke-tests ELKS under qemu-kvm (hardware VT-x via the runner's
+# nested virtualization).  Unlike QEMU TCG, KVM enforces segment descriptor
+# limits and faults exactly like real silicon - the bug class behind
+# PR #2735 (GDT_BIOSDATA / GDT_KDATA limits) is invisible on TCG and only
+# manifests here.  Pass = serial console reaches "login:"; fail = kernel
+# panic or boot timeout.
+
+on:
+  push:
+    paths:
+      - '**'
+      - '!.github/workflows/cross.yml'
+      - '!tools/*'
+  pull_request:
+
+jobs:
+  boot:
+    runs-on: ubuntu-22.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - ibmpc-1440.config    # real-mode baseline
+          - ibmpc-pmode.config   # 286 protected mode - the VT-x-sensitive build
+
+    # ibmpc-pmode currently fails at root mount under hardware
+    # virtualization with XMS enabled (pre-existing issue, invisible on
+    # TCG).  Report it as a warning, not a red X, until that is fixed -
+    # then remove this to make the job gating.
+    continue-on-error: ${{ matrix.config == 'ibmpc-pmode.config' }}
+
+    env:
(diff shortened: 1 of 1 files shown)


ghaerr:
This looks potentially interesting for learning but I don't want to add another CI build since we're already ~4 minutes already for each PR push and commit. Also, https://github.com/ghaerr/elks/pull/2735#issuecomment-4938864590 shows a method for macOS QEMU to be used to actually generate an exception on descriptor limits exceeded.

I'll leave this open for now and play around with QEMU options a bit more to determine how well PM can be tested with the options suggested in #2735.

duzenko:
Can be closed - no problem
I should have raised issue first anyway
I run this CI in my fork so it will fail on upstream merges too

ghaerr:
> I run this CI in my fork so it will fail on upstream merges too

Ok, I will close this now. I have copied this file for use in the future.

 I'm thinking of re-enabling running the CI for pull requests as well as just pushes, which allows users opening PRs to see a full CI run result. Currently, they don't and some merged PRs end up breaking the build. Running CI on pushes and PRs doubles the time from 4 minutes to 8 minutes for my PRs and pushes, so I had turned it off back when I was producing lots of daily PRs and commits. I had thought we had got the CI build time down under 2 minutes with the recent parallel make enhancement, but with the requirement to build OpenWatcom libraries and the maintained PC-98 platform, we're still just under ~4 minutes.

PR #2738, opened 2026-07-10
[kernel] Fix XMS PM copy descriptor limits (byte length, not word count)

duzenko:
This is another attempt to fix the master PM crash by AI.
This time it's more localized and grounded. It actually fixed the CI build to login test.

AI description:

xms_fmemcpyw() allocates a temporary GDT selector to reach an XMS (>1MB linear) operand, then copies `count` WORDS through it:

    sel_src = desc_alloc(src_seg, count, DESC_KDATA);   /* limit = count bytes */
    ...
    fmemcpyw(dst_off, dst_seg, src_off, src_seg, count); /* copies 2*count bytes */

The descriptor limit is passed `count`, but every caller passes a word count (BLOCK_SIZE/2, sector_size>>1, ...) and the copy touches 2*count bytes starting at the far pointer's offset.  So the selector is up to half the size the access needs, and the far pointer offset is ignored entirely.  xms_fmemcpyb()/xms_fmemset() have the same offset omission (without the word-doubling).

On QEMU TCG segment limits aren't checked, so this is invisible.  On hardware-assisted virtualization (VirtualBox VT-x, QEMU/KVM) the limit is enforced: at root mount the first block read out of the XMS floppy track cache

    xms_fmemcpyw(req->rq_buffer, req->rq_seg, cache_offset, df_cache_seg,
                 BLOCK_SIZE/2);       /* directfd.c */

reads 1024 bytes at cache_offset (up to ~8K into the 9K cache) through a selector limited to 512 bytes -> #GP.  The PM fault reporter then faults again on the same path -> triple fault, so ibmpc-pmode with XMS enabled resets/Guru-Meditates right after "df0: floppy type 1.44M" and never mounts root.  xms=off avoids it only because it removes the XMS selector from the path.

Set the limit to cover the whole access: offset + (count << 1) for the word copy, offset + count for the byte copy and memset.

Same one-line unit-mismatch family as the GDT_BIOSDATA and GDT_KDATA descriptor-limit fixes.

Tested: ibmpc-pmode with XMS on now boots to login (mounts root, runs /etc/rc.sys) on VirtualBox VT-x and under QEMU/KVM, where it previously triple-faulted; QEMU TCG unaffected.

### DIFF
--- a/elks/arch/i86/mm/xms.c
+++ b/elks/arch/i86/mm/xms.c
@@ -122,7 +122,13 @@ ramdesc_t xms_alloc(unsigned int kbytes)
 
 #ifdef CONFIG_286_PMODE
 
-/* copy words between XMS and far memory using dynamic PM selector */
+/* copy words between XMS and far memory using dynamic PM selector.
+ * The temp selector's byte limit must cover the whole access: the far
+ * pointer's offset plus (count words << 1) bytes.  count is a WORD count,
+ * so the limit is offset + (count << 1), NOT count (which would be half a
+ * word-copy's length and fault on the first access past it on hardware
+ * that enforces segment limits - QEMU TCG does not, hardware VT-x/KVM do).
+ */
 void xms_fmemcpyw(void *dst_off, ramdesc_t dst_seg, void *src_off, ramdesc_t src_seg,
 		size_t count)
 {
@@ -132,11 +138,11 @@ void xms_fmemcpyw(void *dst_off, ramdesc_t dst_seg, void *src_off, ramdesc_t src
         dst_seg, dst_off, src_seg, src_off, count);
 
     if (src_seg >> 16) {
-        sel_src = desc_alloc(src_seg, count, DESC_KDATA);
+        sel_src = desc_alloc(src_seg, (size_t)src_off + (count << 1), DESC_KDATA);
         src_seg = sel_src;
     }
     if (dst_seg >> 16) {
-        sel_dst = desc_alloc(dst_seg, count, DESC_KDATA);
+        sel_dst = desc_alloc(dst_seg, (size_t)dst_off + (count << 1), DESC_KDATA);
         dst_seg = sel_dst;
     }
 	fmemcpyw(dst_off, (seg_t)dst_seg, src_off, (seg_t)src_seg, count);
@@ -154,11 +160,11 @@ void xms_fmemcpyb(void *dst_off, ramdesc_t dst_seg, void *src_off, ramdesc_t src
         dst_seg, dst_off, src_seg, src_off, count);
 
     if (src_seg >> 16) {
-        sel_src = desc_alloc(src_seg, count, DESC_KDATA);
+        sel_src = desc_alloc(src_seg, (size_t)src_off + count, DESC_KDATA);
         src_seg = sel_src;
     }
     if (dst_seg >> 16) {
-        sel_dst = desc_alloc(dst_seg, count, DESC_KDATA);
+        sel_dst = desc_alloc(dst_seg, (size_t)dst_off + count, DESC_KDATA);
         dst_seg = sel_dst;
     }
 	fmemcpyb(dst_off, (seg_t)dst_seg, src_off, (seg_t)src_seg, count);
@@ -174,7 +180,7 @@ void xms_fmemset(void *dst_off, ramdesc_t dst_seg, size_t count)
     debug("xms_fmemset(%08lx:%04x count %u\n", dst_seg, dst_off, count);
 
     if (dst_seg >> 16) {
-        sel_dst = desc_alloc(dst_seg, count, DESC_KDATA);
+        sel_dst = desc_alloc(dst_seg, (size_t)dst_off + count, DESC_KDATA);
         dst_seg = sel_dst;
     }
 	fmemsetb(dst_off, (seg_t)dst_seg, 0, count);


ghaerr:
Hello @duzenko,

Thank you for another good catch. The AI's proposed solution was still slightly inaccurate: it created a selector valid from base + 0 -> base + offset + (count << 1). This, while fixing the fault from having too small a limit, also suffers from allowing potential access all the way from base + 0 to base + offset plus count << 1 bytes, when only count << 1 bytes from the src/dst + offset address is required.

This additional issue is addressed and fixed in 95bb89d2125ee5a02dcd21efcf59a2dfb482a8bb.

I've managed to get QEMU to reproduce this failure by running -accel hvf, which was suggested in https://github.com/ghaerr/elks/pull/2735#issuecomment-4938864590 and works on macOS, thanks. Hopefully this will allow me to produce much less buggy code due to lack of actual and proper testing of descriptor emulation.

> reads 1024 bytes at cache_offset (up to ~8K into the 9K cache) through a selector limited to 512 bytes -> #GP. The PM fault reporter then faults again on the same path -> triple fault

The PM fault report is faulting again, producing a CPU reset. This was explained previously, and here, as due to stack overflow in https://github.com/ghaerr/elks/pull/2735#issuecomment-4938864590 as the reason for past triple-faults.  I don't understand why the fault reporter here is likely triple-faulting, since the previous issue was the fault reporter required a working direct console (since it couldn't access BIOS data area). However, here, that is no longer the case. The reporter should be redirecting its output to ser_putc and output on the serial port (See `kputchar` in printk.c).

So I still would like to understand why the triple fault here. I, like your original code, would like to have the exception handler work and display the fault info at any step in the boot process. If your AI can answer this, that would help. Otherwise I plan on continuing to look into getting the fault reporter working at any time after boot.

Thank you!

duzenko:
Here's AI reply to your last question:

The triple-fault here isn't the console/BIOS, and it isn't stack overflow — the
fault entry runs the reporter with a **stale DS**.

`pm_fault_vector` calls the C reporter without reloading DS:

```
    pushw $vec
    call  pm_exception_handler
```

`pm_exception_handler()` → `panic("EXC %04x ...")` reads the `.rodata` format
string and runs `printk`/`vsprintf`, all DS-relative. So the reporter only
survives if DS still points at the kernel data segment when the fault hit. Two
cases behave very differently:

1. **Fault on a bad segment *load*** (loading a paragraph as a selector, a bad
   RPL, etc.): the `mov %ds,<sel>` itself faults, so DS keeps its previous,
   valid KERNEL_DS value → the reporter runs and you get a clean `EXC=` line.
   I hit exactly this recently — a #GP from loading a paragraph-as-selector
   printed `panic: EXC 000d CODE 1850 CS 0008 IP .... HALTED` on the serial
   port with no trouble.

2. **Limit/protection violation while DS already holds a foreign descriptor** —
   e.g. `xms_fmemcpy` has loaded the temporary XMS selector (base >1MB, small
   limit) into DS for the copy, and the copy runs past that limit. The fault now
   fires with **DS = the XMS operand**. The reporter's first DS access — reading
   the panic format string, a low kernel-data offset — lands outside that
   descriptor's base+limit → a second #GP → double fault → and the exception
   gate delivery re-faults the same way → triple fault → reset. You never see a
   byte, because it dies *reading its own format string*, before anything
   reaches `ser_putc`.

That's why the serial redirect doesn't help (the fault is upstream of the
output), and why it isn't stack overflow: SS stays the flat kernel data
selector, so both the CPU's exception-frame push and the handler's `&arg` frame
read work fine — only DS is wrong.

Fix: at the top of the fault entry, **save the faulting DS/ES, then reload them**
to the kernel data selector before the C handler runs. Saving them first matters
— for this bug class the faulting DS (the >1MB XMS descriptor) is the key clue,
so the reporter should display it, not discard it. Pushing a segment register is
safe even when it's a bad/limited selector (it pushes the selector value via SS,
it doesn't dereference it):

```
        pushw   $vec
        jmp     pm_fault_common
        ...
pm_fault_common:
        push    %ds            ; preserve faulting DS for the report
        push    %es            ; preserve faulting ES
        push    $SEL_KDATA     ; make the reporter's own DS/ES valid
        pop     %ds            ; (pop reads from SS, not the broken DS)
        push    $SEL_KDATA
        pop     %es
        call    pm_exception_handler
```

The saved DS/ES are now on the stack as part of the frame, so
`pm_exception_handler` can add them to the `EXC=` line (adjust its `&arg`
indexing + format string for the two extra words). Then the reporter both
survives any DS/ES state *and* shows what they were — at any point in boot. For
the XMS case that line would immediately show DS pointing at a >1MB descriptor.

ghaerr:
Hi @duzenko,

> the fault entry runs the reporter with a stale DS.

Ah, of course - the exception handler isn't being run through \_irqit (which it should, but this code is supposed to work prior to most of kernel init). \_irqit will setup the kernel SS and DS == SS which is necessary for properly calling almost all kernel routines, including `panic` or `printk`. 

I must admit I'm finding things interesting dealing with the speed with which AI both produces correct and incorrect explanations, and am also contemplating which is better: actually knowing something well enough to write it in the kernel, or being able to very quickly get a complex answer which is correct most/some of the time. I believe the former, but the latter can be very helpful. So thank you for this AI answer; it both gives me a good idea of how to fully handle the issue, as well as reminds me I need to think a bit more when writing or re-writing code I didn't write originally myself. But that's the fun of it - deeper learning of how things work through experience.

I'm also finding it nice that the PM kernel can use (emulated) hardware to better identify and then fault over buggy portions of the kernel. Just having and experimenting with a PM kernel makes the real mode kernel safer as it exposes more potential problem areas for consideration.

I'll use your AI suggestion above to think through how to get the kernel init fault-catching code to always report without triple-faults, and then also how to transition to a more proper fault-catcher which would run through \_irqit to catch and report faults during normal kernel and user mode operations after system initialization. User-mode faults should be able to be handled by killing the process, much like the DIV0 handler currently, instead of calling panic and halting the system.

Thank you!

PR #2739, opened 2026-07-11
Elks dhcp implementation

djohanse:
Implementation of DHCP.  Configures user's ip address dynamically.

### Internal, net connection

Screenshot 2026-07-11 at 8 12 18 AM

### internal, net connection, telnet

Screenshot 2026-07-11 at 8 14 27 AM

### internal, net connection, ping
Screenshot 2026-07-11 at 8 17 07 AM

### internal, net connection, ftp
Screenshot 2026-07-11 at 8 14 51 AM

### external, net connection, telnet
Screenshot 2026-07-11 at 8 18 32 AM

### external, net connection, ftp
Screenshot 2026-07-11 at 8 18 01 AM

### internal, tap connection
Screenshot 2026-07-11 at 8 24 37 AM

### internal, tap connection, ftp
Screenshot 2026-07-11 at 8 25 15 AM

### internal, tap connection, ping
Screenshot 2026-07-11 at 8 27 48 AM

### internal, tap connection, telnet
Screenshot 2026-07-11 at 8 47 18 AM

### external, tap connection, telnet
Screenshot 2026-07-11 at 8 49 35 AM

### external, tap connection, ftp
Screenshot 2026-07-11 at 8 31 29 AM

### net connection 
Screenshot 2026-07-11 at 8 55 13 AM

### tap connection
Screenshot 2026-07-11 at 8 59 39 AM

Screenshot 2026-07-11 at 9 00 35 AM


ghaerr:
Hello @djohanse,

Wow, an implementation of the important DHCP protocol, along with regression testing, thank you!

In general, this all looks nicely done. The beginnings of a UDP implementation without the full overhead of running through sockets is probably a good idea, but then DHCP itself has to live core-resident in ktcp whether it's used or not.

I do have some small concerns and nits, but they're pretty small given the utility of this enhancement, so I think I'll commit now and possibly do some minor source cleanup afterwards, or perhaps you can do them if you have time.

A growing concern is or will be the size of resident ktcp - of course the code size is limited to 64K max and is always core resident when networking is running. I think having UDP and DHCP greatly outweigh this, but we may want the DHCP option compile-time configurable. Another issue that you may not be testing for and sometimes pops up is ktcp's data segment size ultimately limits the number of active connections. This is mostly dynamic but static strings are something to consider making as small as possible.

I'll list some other smaller concerns by the code, for your possible future consideration in an add-on PR, or I can clean up after commit. 

Thank you for this enhancement, ELKS networking is definitely taking a big leap forward thanks to your contributions :)

elkscmd/ktcp/ktcp.c
```diff
@@ -245,9 +254,15 @@ int main(int argc,char **argv)
...
-    netmask_ip = in_gethostbyname(optind < argc? argv[optind++]: default_netmask);
+    if (!dhcp_enabled) {
+	local_ip = in_gethostbyname(optind < argc? argv[optind++]: default_ip);
+	gateway_ip = in_gethostbyname(optind < argc? argv[optind++]: default_gateway);
+	netmask_ip = in_gethostbyname(optind < argc? argv[optind++]: default_netmask);
+    } else {
+	local_ip = 0;
+	gateway_ip = 0;
+	netmask_ip = 0;
+    }
```
ghaerr, line 265: This else clause can be deleted to save code, as these global variables are guaranteed to be set to 0 at program start.

elkscmd/ktcp/udp.h
```diff
@@ -0,0 +1,36 @@
+#ifndef UDP_H
+#define UDP_H
+
+#include <stdint.h>
+
```
ghaerr, line 5: While stdint.h is pretty standard, the kernel itself and all the rest of ktcp uses the ELKS (non-standard) definitions like \_\_u16 instead of uint16_t. Thus, at least for internal definitions to match the rest of ktcp, I think this should be removed and the below udphdr_s and udp_sock structures use \_\_u16 for consistency.

I notice that dhcp.c is written also using stdint.h, but can possibly relax on requiring its rewrite without using stdint.h, since it is more of an "internal application" using the internal UDP without sockets. I don't think it worth the trouble of implementing UDP into sockets at this time.

elkscmd/man/man8/dhcp.8
```diff
@@ -0,0 +1,135 @@
...
+.RB [ yes
+.IR | no ]
+.SH DESCRIPTION
+.B DHCP
+is a built-in subsystem of the ktcp TCP/IP daemon.
+When enabled, ktcp automatically negotiates an IP address, netmask, and default
+gateway using the DHCP protocol at startup.
+.PP
+DHCP is configured in
+.BR /etc/eth0.cfg .
```
ghaerr, line 15: Can you rewrite this entire man page to be more accurate for ELKS? There is no /etc/eth0.cfg etc, and such references should be deleted and/or made more accurate with your actual implementation.

elkscmd/rootfs_template/bin/net
```diff
@@ -36,14 +36,26 @@ start_network()
...
+			ktcp="ktcp -b -p cslip -s $baud -l $device $localip $gateway $netmask"
+		fi
 		;;
 	ne0|wd0|3c0)
-		ktcp="ktcp -b -p $link $mtu $localip $gateway $netmask"
+		if test "$dhcp" = "yes"; then
+			ktcp="ktcp -b -D -p $link $mtu"
+		else
+			ktcp="ktcp -b -p $link $mtu $localip $gateway $netmask"
+		fi
```
ghaerr, line 58: All this duplication of possible ktcp command lines could be made much more readable by having a single command line prepared and then leaving some shell variables blank, rather than repeating command lines all over the place, which is harder to maintain.

Another idea would be to possibly have ktcp automatically start in "DHCP mode" if it is only passed a few arguments, and localip/gateway/netmask omitted. This would then allow for -D to be removed, and the system would tend to "just work", while also leaving older ELKS configurations compatible.

elkscmd/rootfs_template/etc/net.cfg
```diff
@@ -3,9 +3,13 @@
...
-if test "$LOCALIP" != ""; then localip=$LOCALIP; else localip=10.0.2.15; fi
-gateway=10.0.2.2
-netmask=255.255.255.0
+# Uncomment the following line to use DHCP instead of static config:
+#dhcp=yes
+if test "$dhcp" != "yes"; then
+    if test "$LOCALIP" != ""; then localip=$LOCALIP; else localip=10.0.2.15; fi
+    gateway=10.0.2.2
+    netmask=255.255.255.0
+fi
```
ghaerr, line 12: This might also be cleaned up, where we set LOCALIP=10.0.2.15 in /bootopts and thereby eliminate the test for $LOCALIP.

How is gateway set for DHCP use, does it still need to be set somewhere? This perhaps should also move to /bootopts, so that the system quickly either runs in DHCP mode or not, depending on /bootopts, rather than editing /etc/net.cfg.

A final nit is that the size of /etc/net.cfg is 992 bytes, which is <= 1024. This then saves a full extra 1K on distribution floppies, for which there is very little space for all our cool features you're adding. The above modification sets /etc/net.cfg > 1024, using more space.

elkscmd/ktcp/udp.c
```diff
@@ -0,0 +1,83 @@
...
+
+    for (i = 0; i < MAX_UDP_SOCKS; i++) {
+	if (udp_socks[i].active && udp_socks[i].local_port == dest) {
+	    udp_socks[i].callback(iph, src, data, datalen);
+	    netstats.udprcvcnt++;
+	    return;
+	}
+    }
+    netstats.udpnoportcnt++;
+    icmp_send_port_unreachable(iph);
```
ghaerr, line 82: Also, just noticed after commit the following warning:
```
udp.c:82:5: warning: implicit declaration of function 'icmp_send_port_unreachable' [-Wimplicit-function-declaration]
```
This likely requires \#include "icmp.h" in the includes list of this file.

ghaerr:
> This then saves a full extra 1K on distribution floppies, for which there is very little space for all our cool features you're adding.

Looks like the worry about running out of space has already come true: the CI build after commit just broke, we've run out of space on the 1440k floppy! This is likely not only because of larger /etc/net.cfg etc but also ktcp itself is larger:

> image 1K blocks: boot 1, super 1, inodemap 1, filemap 1, inode 8, data 1428
inodes inuse 19, free 236, overhead 1, total 256
blocks inuse 201, free 1227, overhead 12, total 1440
mfs  /home/runner/work/elks/elks/image/fd1440-minix.img genfs -a -1 -n14 -i256 -s1440 /home/runner/work/elks/elks/target
mfs: /dev already exists, no change made
mfs: /bin already exists, no change made
mfs: /etc already exists, no change made
No free slots in bitmap found
make[1]: *** [Make.image:93: minixfs] Error 255

We probably want to keep DHCP included in ktcp for the time being for others to test with, so I'll decide what else we can remove from the 1440k floppy to accommodate all our good stuff. I'll do that in a bit, let me know if you have time to make any of the other above changes, or if I should do it.

djohanse:
> > This then saves a full extra 1K on distribution floppies, for which there is very little space for all our cool features you're adding.
> 
> Looks like the worry about running out of space has already come true: the CI build after commit just broke, we've run out of space on the 1440k floppy! This is likely not only because of larger /etc/net.cfg etc but also ktcp itself is larger:
> 
> > image 1K blocks: boot 1, super 1, inodemap 1, filemap 1, inode 8, data 1428
> > inodes inuse 19, free 236, overhead 1, total 256
> > blocks inuse 201, free 1227, overhead 12, total 1440
> > mfs  /home/runner/work/elks/elks/image/fd1440-minix.img genfs -a -1 -n14 -i256 -s1440 /home/runner/work/elks/elks/target
> > mfs: /dev already exists, no change made
> > mfs: /bin already exists, no change made
> > mfs: /etc already exists, no change made
> > No free slots in bitmap found
> > make[1]: *** [Make.image:93: minixfs] Error 255
> 
> We probably want to keep DHCP included in ktcp for the time being for others to test with, so I'll decide what else we can remove from the 1440k floppy to accommodate all our good stuff. I'll do that in a bit, let me know if you have time to make any of the other above changes, or if I should do it.

Hello @ghaerr,

I will not be able to get back to working on this for a couple days so if you want to go ahead and make the changes that sounds good to me.  Otherwise I can get back to it late Sunday night at the earliest.

I will also review/rework the man page.

ghaerr:
@djohanse: I've reworked the /etc/net.cfg and /bin/net scripts somewhat to more easily integrate your DHCP client in #2740. The decreased script sizes now allow the 1440k image to be built, which is nice.

Most everything requested above has been handled, except the dhcp(8) man page.

PR #2743, opened 2026-07-13
Dhcp readme cleanup

djohanse:
Correct and clean up dhcp readme


ghaerr:
Thanks! I think we're going to need a separate ktcp and DHCP man page. I'll work on that.

Issue #2720, opened 2026-07-02
Considerations for procfs vs /dev/kmem for ps and meminfo


duzenko:
You must have noticed in my 286 PR that some of the ifdef's were in the char memory driver.
Even with best effort, it's not enough to let programs like `ps` or `meminfo` things like process name/cmd line.
So AI's suggestion is to replace/augment kmem with `/proc/` file system.
https://github.com/ghaerr/elks/compare/master...duzenko:elks:procfs
It raises a few questions.

- (me) Whether procfs should stay optional/off-by-default (hence the ps fallback) or become standard. I think it's nice to have but it's also 2-3KB of kernel size.
- (AI) The procfs_read_super unconditional-success issue (can hijack root mount if FAT isn't in file_systems[]) — should reject real block devices.
- (AI) Direction for the remaining kmem readers (meminfo needs a /proc/meminfo; getty/ktcp read uptime).
- (per your PR review) The struct segment linear-base / MEM_SEL2LINEAR idea as an optional hardening of the kmem fallback for PM. So segment's physical address member could make all reader work via `kmem` in theory.

ghaerr:
> AI's suggestion is to replace/augment kmem with /proc/ file system.

Yes, I noticed that in the PMODE-TODO.md document. 

I don't think its a good idea to implement /proc in ELKS early on (or possibly ever), just so that `ps` and possibly `meminfo` can work: the kernel is already almost too large, and the extra dead weight internally just for running `ps` is not worth it. There will be other ways we can solve this problem. The current 286 PM port already uses 6K+ extra kernel data segment which is going to have to be rearchitected at some point, should this port gain legs.

As usual, the AI is suggesting fixes/enhancements faster than I/we can process them. I am going to leave all the issues related to getting `ps` to work as-is for the moment and concentrate on understanding whether the ELKS kernel can be made to work in protected mode on not only 286 processors, but 386+ as well. I am renaming this issue so that all such issues can be discussed here, along with implementation suggestions and considerations.

The first order of importance is fully understanding which CPU features are being used for the current POC PM implementation, with the idea of widening its implementation to 386+ CPU's. As Claude's comment in https://github.com/duzenko/elks/pull/1#issuecomment-4869904362 details, we can use GDT and IDT's, execute in ring 0, but no 32-bit registers or instructions. I think this can be made to work, but more research is needed.

[...]

PR #2744, opened 2026-07-13
Elks dmesg

djohanse:
Implementation of dmesg.

A bit different tack from the network applications.  But I did some work in my real life with linux device drivers years ago.  So while AI wrote this, I am familiar with linux device drivers and I do understand it.

Screenshot 2026-07-13 at 12 25 21 AM


ghaerr:
Hello @djohanse,

Thanks for posting this. I've been considering how to implement dmesg for some time, it's more complicated than it seems due to limits imposed by the 8086 architecture and all the other things people want from a kernel.

I'll start with a summary with details to follow: While this implementation is naively "proper", it's use of kernel code and data is *way* too large for something used as infrequently as `dmesg`. In particular, there's no way we can statically allocate 2K of kernel data for such a function.

The problem here is similar to that discussed in #2720, where the proposal was made to implement a Linux-style procfs (/proc) for use by `ps` and `meminfo`, which "only" used 2-3k kernel space. While convenient to write things as though we're running in a 32-bit code and data address space where the resource limits basically won't ever be reached, when ELKS reaches its extremely tight 64K kernel data segment limit or 128K code segment limit, no new features *at all* can be added. In the ps/meminfo case, a solution was finally devised that changed 4-5 lines of kernel code and about the same in ps.c and meminfo.c, thus allowing it to work in the PM kernel without rearchitecting major sections of the kernel and adding a bunch of new code.

In this case, `dmesg` is essentially a tiny program that implements `cat /dev/kmsg`, while the resident kernel has to keep a resident 2K data buffer and almost-fully-duplicated lib/chqueue.c queue/dequeue for a utility that most will never use. The solution needs to be the opposite: the kernel needs an extremely minimal implementation, while the user mode application bears the brunt of accessing and decoding the kernel data. I know this goes against the grain for "proper" design, but the same proper design only works for large systems that can tolerate bloat via large address spaces and hardware memory installed.

More details on what will be required for `dmesg` in ELKS:
- No static kernel data. The data will have to be stored in otherwise-unused main memory. This is a bit complicated since we don't know exactly which address to use when `start_kernel` is executed, unless setup.S figures that out during the kernel load process.
- Implementing a new special file requires more than is expected: e.g. for FAT filesystems, the entire /dev directory is "faked" and implemented in kernel code via fs/msdos/inode.c, requiring yet another entry in an already tight mapping; otherwise /dev/kmsg won't work. Same problem for ROMFS filesystems, where inodes are in short supply.
- Since extra main memory may be used for the dmesg implementation, there needs to be a way to turn it off, otherwise extremely large programs like `elksdoom` won't run, which need absolutely all of main memory with smaller task=, file= etc in /bootopts. Thus, /bootopts may need to specify whether to keep a dmesg buffer and that needs to be specially read by setup.S as well.
- A good place for the buffer may be at the very top of main memory, less a fixed size for the buffer. This would then be outside all the heap_alloc and seg_alloc functions, which would probably be OK unless we wanted to dynamically allocate the buffer size or eliminate it, etc.
- The lib/chqueue.c functions won't work for far buffers, a very small implementation would be required, with minimal stack usage, since printk/kputchar typically use the very last bit of task kernel stack when running FAT filesystem code and contributes to kernel stack overflow, since printk itself is reentrant and called from some interrupt handler routines. ASM might be the best solution for the queuing routine.
- Enqueue/dequeue routines need to be protected from re-entrancy since printk and thus kputchar are re-entrant.
- All dequeuing, scroll-control like `more`, etc handled by user mode `dmesg` program instead of residing in kernel. This would likely only require the far memory address of the queue routine variables followed by the circular queue itself, all at the top of physical memory. Then a single ioctl could return this address, similar to the way that MEM_GETJIFFADDR etc work.

A reviewer of ELKS tried typing `dmesg` on video recently, to no avail. If dmesg were present, it might also allow rethinking the way that the kernel sometimes now displays debug data which in too many cases scrolls past the running application too quickly to read, so it would be a nice enhancement. Other options, like the current ^P kernel debug toggle, as well as the ^O/^N displays for inode and buffer output, could also be redirected to the dmesg buffer.

What is the benefit you're looking to gain from dmesg?

Thank you!

Mellvik:
This PR caught my interest because I've been thinking about a cheap (resource wise) way to implement `syslog` for a while. Which is what this is really about unless I'm missing the point completely. `dmesg` itself is nothing more than a specialized version of `tail` displaying whole or part of the syslog buffer. 

`syslog`, which IIRC has been around since 4.2BSD - extremely useful for debugging and maintenance - is normally implemented as a service (daemon). Minimal code, very flexible and configurable - but it does take a task slot and does require either a named pipe or a Unix socket. Which is where the 'cost' consideration comes in: If enabling either of those adds significant weight to the kernel, it may be a no-go right there. Then there would obviously be a cost involved in making kernel messages get duplicated to syslog.

In addition to pointing out the obvious, that `dmesg` is not or should not be the issue, there are two important points here:
- System log messages - regardless of mechanism - are useless without a time stamp.
- If saving boot messages is the key purpose, there are very easy ways to do it at minimal cost. Like increasing the console scroll buffer to all 4 virtual screens and then have `init` save it to - say, `/tmp/bootlog` before starting user processes.

djohanse:
> This PR caught my interest because I've been thinking about a cheap (resource wise) way to implement `syslog` for a while. Which is what this is really about unless I'm missing the point completely. 

You are correct.  My thinking was that one very practical application of a dmesg/syslogd type implementation would be for someone who wants to run ELKS headless.  This would give them a way to log in and review the status of their system. 

>`dmesg` itself is nothing more than a specialized version of `tail` displaying whole or part of the syslog buffer.
> 
> `syslog`, which IIRC has been around since 4.2BSD - extremely useful for debugging and maintenance - is normally implemented as a service (daemon). Minimal code, very flexible and configurable - but it does take a task slot and does require either a named pipe or a Unix socket. Which is where the 'cost' consideration comes in: If enabling either of those adds significant weight to the kernel, it may be a no-go right there. Then there would obviously be a cost involved in making kernel messages get duplicated to syslog.
> 
> In addition to pointing out the obvious, that `dmesg` is not or should not be the issue, there are two important points here:
> 
> * System log messages - regardless of mechanism - are useless without a time stamp.

You raise a good point.

> * If saving boot messages is the key purpose, there are very easy ways to do it at minimal cost. Like increasing the console scroll buffer to all 4 virtual screens and then have `init` save it to - say, `/tmp/bootlog` before starting user processes.

@ghaerr Thank you for the detailed explanation of everything it would take to implement this.  I had a gut feeling there was more to it than I expected.

But it has spawned interest.  I think this is a topic/project that is worth further discussion.  Since this will take some further design/discussion/thinking, I am ok with closing this PR so it doesn't linger and possibly moving this topic to the discussions or issues board?

Please advise on how you would like me to proceed.

Thanks!

ghaerr:
My take is that the implementation of course depends on what's trying to be accomplished: review just kernel boot messages vs  the ability to review post-boot kernel `printk` messages, vs a full syslog where applications themselves can log to a system log, etc. Reviewing just boot messages could work with a small fixed buffer, while post-boot messages might require a larger or circular buffer, so a configurable buffer length would be a nice option. Full syslog will require a new system call, etc.

For all the above options, I believe it a good idea to split the collection of messages from the displaying of messages, this allows for a minimal implementation while adding features later for more complex display or storage of messages. As described above, the collection mechanism needs be very small, fast and very little stack usage, as it may be called during interrupt time, and moving all display functionality outside the kernel keeps the kernel small and allows the collection mechanism to be remain architecture-independent and display more suited to particular system requirements.

> increasing the console scroll buffer to all 4 virtual screens and then have init save it 

IMO not a good idea, since that wouldn't work for any non-IBM PC system (e.g. PC-98, ROM, Swan, NECV25, headless, etc), would not work with systems booted with an MDA or CGA card on IBM PC, and then drags /bin/init into cleaning up the mess, which also wouldn't work when init= specified a standalone shell in /bootopts, etc.

All that said, having a way to review kernel printk messages would come in handy in many debugging scenarios, as well as just being able to see the boot messages without having to reboot. The same program that reads the circular buffer could also potentially store it, allowing for previously-seen review while keeping a small circular buffer, as well be run as a daemon, etc. Time-stamping is very useful, but complicates the review of messages, since many kernel messages are hand-crafted to 80 columns, thus a horizontal scroll mechanism would likely be required, unless a keystroke were used to turn on/off the timestamp and clip to 80 columns.

> possibly moving this topic to the discussions or issues board?

I don't know if a PR can be moved to an issue, but that would be nice. We can easily refer back here for details already discussed. You're welcome to leave this open a while longer for further comments, then close it and/or re-open as an issue for further discussion with more thoughts. I've made a note to think more about the collection implementation. Perhaps the kernel implementation should be done first, followed later by an application for display of the information collected, which can be enhanced over time.

Mellvik:
Interesting perspective @ghaerr - I agree with you on the general case. As to collecting and saving boot messages, we're seemingly on completely different tangents.
> 
> > increasing the console scroll buffer to all 4 virtual screens and then have init save it
> 
> IMO not a good idea, since that wouldn't work for any non-IBM PC system (e.g. PC-98, ROM, Swan, NECV25, headless, etc), would not work with systems booted with an MDA or CGA card on IBM PC, 

Most importantly, the resource is there, available and would work on just about any PC ever made. At boot time, in real mode, CGA was and is the standard - possibly even today).  The exception, the few MGA and Hercules systems still in operation, are too few and far between to be a showstopper.  Also, if any potential improvement or addition to ELKS should be gauged by it's compatibility with/availability on non-PC platforms, the future would be very limited, wouldn't it?

Secondly it's completely uncomplicated - the scrolling is there already, the buffer too.

> ... and then drags /bin/init into cleaning up the mess, which also wouldn't work when init= specified a standalone shell in /bootopts, etc.

What's the problem? BTW I meant `init/main` but since you mention it, using `/bin/init` to do it may be even better. And there is no mess to clean up, is there? Unless calling `memset` to clear the video pages after saving them is a mess. If booting into a shell, there is a lot that isn't working 'normally' - it's a choice made for a reason with known sacrifices. And if the functionality is important, it's easy to fix.

My point is that if such a feature would be useful, it's a very low hanging fruit.

ghaerr:
> if any potential improvement or addition to ELKS should be gauged by it's compatibility with/availability on non-PC platforms, the future would be very limited, wouldn't it?

Actually I think quite the opposite - not limited, by design. Portable implementations increase the usability of the project as a whole, thus bringing a feature to all users of the project.  After some discussion, our kernel dmesg was implemented without having any dependencies on IBM PC platforms, in a total of 44 bytes resident kernel code and 64 bytes INITPROC reusable code. It works on all platforms and is hardware independent.

Mellvik:
Nice work, @djohanse and @ghaerr. And many very pertinent observations about AI code versus 'real' code.

> Portable implementations increase the usability of the project as a whole, thus bringing a feature to all users of the project.

I don't think anyone would disagree with that. And it's not what your initial comment said. What you implied was that anything that cannot work on all platforms mentioned is uninteresting. I find that unwise given the fact that more than 99.99% of the potential 'takers' would be using a PC compatible platform. Or maybe I just read you wrong.

PR #2751, opened 2026-07-14
[build] Add optional GEM Desktop external application

parabyte:
## Summary

- add a default-off IBM PC real-mode GEM Desktop external application pinned to public tag v2026.07.14-core.1
- build the native AES/VDI owner and original Desktop client with the ELKS GNU ia16 toolchain
- install exactly two executables and four original Desktop resources into hard-disk images
- fail closed when IBM PC, real mode, or CONFIG_GEM_TRAP requirements are not met
- document build, runtime ordering, GPL provenance, and the direct resource path

Public source: https://github.com/parabyte/Gem-elks
Pinned commit: 9797037be09f1a064e63d8c03843c35b8f2d947a

## Dependencies

This draft depends on #2745 for the GEM INT EF broker and gemctl ABI. Runtime qualification also used #2747 for true vfork, #2746 for task-owned graphics access, and #2748 for medium-model signal delivery. #2749 is useful for supervised direct startup but is not required for manual startup.

## Validation

- anonymous HTTPS clone and pinned detached checkout passed
- buildext completed from the public tag
- ELKS model gates passed for both 01 03 executables
- source gate passed 151 files
- 8086 disassembly gate passed 127 objects
- retained-function audit passed with zero duplicate strong symbols and 91.35% core GEM lineage
- image installer staged exactly /bin/gemaes, /bin/gemdesk, and four GEMSYS assets; all six byte comparisons passed
- config visibility matrix and install-time dependency failures passed
- original asset SHA-256 manifest passed
- final diff, path, commit metadata, and public-source privacy scans passed

Binary SHA-256 values:

- gemaes: 159bbb34de4c597237fdff253c29a3d1e6e834e6d9c365409d695eaf1eb6fbe4
- gemdesk: 255098a800e10f129716694d961ca651197be0fcc104e4ead1ac5206bd57dd7a

## Licensing

The retained FreeGEM/OpenGEM core and original runtime assets carry their GPL version 2 release notices. ELKS adaptation work is GPL-2.0-only, with historical notices and pinned provenance retained. The DRI Toolkit Clock-derived material is not included.

### DIFF
--- a/Documentation/Readme.md
+++ b/Documentation/Readme.md
@@ -21,6 +21,8 @@ This section contains some HOWTOs to help using ELKS.
 
 * [The Kilo editor](https://htmlpreview.github.io/?https://github.com/jbruchon/elks/master/Documentation/html/user/kilo.html)
 
+* [GEM Desktop on ELKS](https://raw.githubusercontent.com/ghaerr/elks/master/Documentation/text/gem.txt)
+
 * [Writing applications for ELKS in C](https://htmlpreview.github.io/?https://github.com/jbruchon/elks/blob/master/Documentation/html/user/writing_apps_in_C.html)
 
 * [Writing applications for ELKS in assembler](https://htmlpreview.github.io/?https://github.com/jbruchon/elks/blob/master/Documentation/html/user/writing_apps_in_assembler.html)
--- a/Documentation/text/gem.txt
+++ b/Documentation/text/gem.txt
@@ -1,0 +1,40 @@
+GEM Desktop for ELKS
+====================
+
+The optional GEM Desktop external application is intended for IBM PC compatible
+8088 and 8086 systems using an ELKS hard disk image.  It runs in real mode and
+requires CONFIG_GEM_TRAP=y.  It is unavailable when CONFIG_286_PMODE=y or when
+another ELKS architecture is selected.
+
+The GEM AES, VDI and Desktop sources and direct runtime resources are derived
+from the GPL-2.0 FreeGEM/OpenGEM source release; original copyright and GPL
+notices are retained.  ELKS-specific adaptation sources in Gem-elks are
+distributed under GPL-2.0-only.  See the external repository's LICENSE and
+provenance manifest for exact source revisions and asset hashes.
+
+The build is pinned to the v2026.07.14-core.1 tag at:
+
+    https://github.com/parabyte/Gem-elks
+
+After configuring ELKS, build the pinned source with:
+
+    ./buildext.sh elksgem
+
+Then enable "GEM Desktop (hard disk only)" in the Userland configuration and
+build the hard disk image normally.  The installation contains:
+
+    /bin/gemaes
+    /bin/gemdesk
+    /GEMAPPS/GEMSYS/DESKTOP.RSC
+    /GEMAPPS/GEMSYS/DESKHI.ICN
+    /GEMAPPS/GEMSYS/DESKLO.ICN
+    /GEMAPPS/GEMSYS/DESKTOP.INF
+
+Start /bin/gemaes first, then start /bin/gemdesk from a second shell or from a
+separate supervised startup entry.  gemaes must remain running as the shared
+AES and VDI owner while gemdesk uses the kernel GEM broker.  gemdesk cannot be
+started before the resident owner.
+
+The four data files are original GEM resources.  They are installed unchanged
+and parsed directly on ELKS; no resource conversion program is used on the
+target system.
--- a/buildext.sh
+++ b/buildext.sh
@@ -14,6 +14,7 @@
 #       microwindows_pc98 ia16-elf-gcc  Nano-X Graphical Windowing Environment for PC-98
 #       dcc             ia6-elf-gcc     DCC self-compiling C compiler for ELKS
 #       dflat           ia16-elf-gcc    D-Flat TUI memopad/library
+#       elksgem         ia16-elf-gcc    GEM Desktop for ELKS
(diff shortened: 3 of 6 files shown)


ghaerr:
Wow, 7 PRs in one hour, 600+ lines of existing kernel code file mods and another 800+ lines of new kernel code. It will definitely take a while to review this, let alone figure out whether it should be committed, changes made, and how it will be maintained.

Since there are so many open PRs, I thought to start here and ask a few questions, then move to other PRs, in the order of: 1) importance to existing ELKS kernel and applications, and 2) acceptance of the GEM graphics desktop into ELKS.

To start, most of the PRs (except MFM) look like they're generated from a template: who wrote all this stuff, and who/what wrote the PRs? Where are all the regression tests that are being run for the heavy kernel mods?

I would say that medium model signal handling is important and merits more immediate consideration. 

While I agree that a `vfork` that doesn't copy the caller's data segment would be good when memory is tight, why does GEM need vfork? Will `fork` work, just slightly slower and use more RAM? Also, why can't the existing (untested) `fvork` implementation in fork.c be used by just changing \#if 1 to \#if 0: 
```
pid_t sys_vfork(void)
{
#if 1
    return do_fork(0);
#else
```

Each PR with significant changes will need a suite of regression tests in order to prove it's addition doesn't affect the existing stability of the kernel.

I notice that lots of kernel code is added to workaround having the compiler emit a MUL or occasional DIV instruction: why? These very infrequent cases bloat kernel code for no real gain, IMO. We can get further into that on a case-by-case basis once I understand your reasoning.

How can the MFM driver be tested? What emulators are available for such a driver? Also, adding 40+ new /dev entries is definitely problematic, for lots of reasons.

Finally, can you describe what the GEM desktop actually does? I don't see any screenshots on your fork. Given that the propose kernel modification entail 1300+ lines, I'd like to understand a lot more about what GEM brings to ELKS, how GEM applications are written, that sort of thing. Is there a terminal emulator?

All said, this is a huge set of modifications that will certainly have to wait until after a 0.9.2 or later to be committed. I also think we can likely engineer a better way to allow the implementation of the GEM INT EF broker in an application rather than the kernel; or at the very least having GEM files completely outside core kernel files and routines.

parabyte:
Basically i will rethink this approach, 

I was going for a genuine Gem desktop experience, I took things a little too far.

ghaerr:
> I was going for a genuine Gem desktop experience, I took things a little too far.

There's nothing wrong with a "genuine" GEM desktop experience, graphically speaking, I would say. It would be nice to see what it actually looks like, I've seen pictures years ago but never actually run it.

And there's nothing directly wrong with continuing the apparently original implementation of INT EF, etc: but there may be ways to allow the "resident broker" to in fact be an ELKS process that traps INT EF (yes, initially outside the kernel for starters), and communicates with GEM applications. I don't know the internal GEM architecture to comment further. In the Nano-X environment, for instance, one can build the applications linked directly with the Nano-X server, thus eliminating any special need for inter-application or application-to-server specialized communications. This might be a worth first step in getting GEM running on ELKS: everything linked into each GEM app. It seems there is only a single app anyways, right: the GEM desktop. Is that a single app, or does it run multiple applications?

HOWEVER, depending on whether you're trying to "literally run GEM on ELKS", or whether you just "want the look of GEM on ELKS", a different approach could be taken: modifying the GEM development tools or application libraries such that single GEM applications might run without requiring the ELKS kernel itself to directly handle GEM requests. In either case, we might be able to engineer kernel "exec" and "exit" callouts that use another INT to communicate with a rewritten GEM trap handler running in user mode.... that kind of thing.

In summary, what I would like, if you're interested, would be to keep the ELKS kernel very small and stable, while possibly engineering a way for larger subsystems with specialized code (taking advantage in some cases of the fact that there is no MMU or real protection) and allowing large specialized subsystems to run *on* ELKS. This direction would keep a very large part of the required per-platform specialized modifications outside the kernel and in the user-mode application space.

All that said, there may be good reasons to add vfork, graphics task-locking, etc.

Can you still answer my above questions so I don't have to repeat them on each of the individual other PRs? Thanks!

ghaerr:
Hello @parabyte,

> lots of kernel code is added to workaround having the compiler emit a MUL or occasional DIV instruction: why? These very infrequent cases bloat kernel code for no real gain, IMO.

My comment above was written hastily after reading the following comment in #2747, the `vfork` enhancement:
```
   /*
     * Do not form &task[max_tasks] here.  Scaling that variable array index
     * makes the 16-bit compiler emit MUL on an original 8088/8086.  A bounded
     * countdown and one structure-pointer increment per slot use only cheap
     * word decrement, compare and add instructions.
     */
```
After more inspection, and especially after looking at the definition of `for_each_task` in sched.h:
```
#define for_each_task(p) \
        for (p = &task[0] ; p!=&task[max_tasks]; p++ )
```
I realized that my comment was basically wildly incorrect: every instance of kernel code using that macro to iterate the task array (and there are 12 of them) generates code using a MUL instruction; *and* the MUL is executed *every* time through the loop! Thus in most instances the 8086 will execute 16 very slow MUL instructions during any task inspection loop. The wake_up function, which uses the same macro, is called from interrupt handlers all the time, making the system slow during interrupt processing as well.

So, the `vfork` code comment turned out to be very useful, as `for_each_task` can be defined using an preset variable for '&task[max_tasks]' which will not only speed the system up but reduce code in every instance. I plan to push a PR with that fix shortly.

Thank you! I had other comments on your PRs but I see you've closed all of them. I have a slightly rewritten vfork implementation based on some other code comments that I also hope to publish shortly after more testing.

PR #2752, opened 2026-07-15
Add XT MFM hard disk driver

parabyte:
## Summary

- add a direct XT WD1002/Xebec command-block MFM hard disk driver
- support WD1002A-WX1 and Seagate ST11/FileCard controller profiles
- add `/dev/mfma` through `/dev/mfmd`, partition scanning, root-device names, and `mfm=slow|fast`
- document geometry overrides, DMA ownership, recovery behavior, and optional maintenance ioctls

## Implementation

Command, status, sense, and parameter bytes use programmed I/O. Sector payloads use PC/XT 8237 DMA channel 3 through a fixed 1024-byte buffer below the first 64 KiB boundary. Controller completion remains polled with IRQ delivery masked.

The driver is limited to 8086 word and byte operations. The existing ELKS two-word `sector_t` ABI is isolated at block and disk-format boundaries; CHS capacity uses explicit word carry handling and no wide multiplication helper.

## Validation

- clean IBM PC real-mode kernel and boot-image build with `ia16-elf-gcc -mtune=i8086`
- Amstrad 1640DD With serdrive to host elks
- created `/mfmproof`, synchronized it, issued a hard reset, and read `MFMREADWRITEPASS` from the guest after remount
- verified the post-reset VHD with ELKS `mfs` and `mfsck`
- compiled the optional maintenance-ioctl configuration without warnings from `mfmhd.c`
- built an MFM-disabled kernel regression configuration
- compiled the MFM-only protected-mode descriptor path
- inspected `mfmhd.o`: no 80186-or-newer instructions, FPU instructions, or wide multiply/divide helper references


ghaerr:
Hello @parabyte,

Thank you for this driver! Is there a way this can be tested with an emulator, or has this been used/tested on real hardware only? Is this the same MFM driver that was used in your (now-removed) elks-enhanced fork?

Who wrote this driver, and how well and how long has it been tested? It seems a very large amount of code is required, I wasn't aware MFM drives were so very complicated. I notice there are quite a few functions in the driver that use excessive stack space with tons of variables declared, including multiple char arrays, etc (e.g. mfm_cmd_len). This excessive stack usage is extremely dangerous on ELKS, as, depending on whether a FAT filesystem is in use at the time of the block device call, kernel stack overflow can easily happen. I haven't studied the driver enough to calculate exactly how much stack each function may use, and whether the function calls might get stacked, adding to the kernel stack requirement. Is this an ASYNC_IO driver, or synchronous?

Other than the above, things look OK, but have some questions/considerations: 
- I've worked to try to keep tons of less-used CONFIG_ options out of the .config file: do we really need the 4x CONFIG_MFMHD_DRIVE0_CYLINDER etc in .config? Unless there's a really good case for requiring them in every config file, I would prefer to see these set in the driver source file itself. What have you found, does each MFM drive require its own override? Same for slow profile, it's already settable in /bootopts, why have another .config option. I can't tell if this was written by AI with every case trying to be handled, or whether it's human crafted after actual needs.
- Do we really need 32 new /dev entries for this driver? Inode space is at a premium, and many systems may boot from floppy. Are people really using all 8 partitions of all 4 drives, or would this as a much shorter list, or commented out. Note that even our BIOS HD driver /dev/hdXX list is much shorter, for the same reasons.
- I am thinking of a better way to code up `root_dev_name` that doesn't require lots of entries, but that's not really a problem for your PR. This table information should probably be provided within the driver, but I haven't figured a way to do that easily yet.
- I don't think we need three different option name variants for `mfm=` in /bootopts, a single integer reduces kernel code size. I would probably change that to `mfmhd_slow_profile = (int)simple_strtol(line+3, 10`.
- `jiffies` will want to be accessed as `jiffies()`, since grabbing a volatile 32-bit long may get interrupted and return an incorrect value.

Out of curiosity, does the BIOS itself on these old machines work with MFM drives, or is this special driver required for it?

Thank you!

parabyte:
let me rework this driver,  and i have used it on a real controller. 

I will rework it when i have access to real machine

PR #2755, opened 2026-07-17
head, tail, uniq: use long counts to avoid 16-bit overflow

duzenko:
@ghaerr 
Can we please have 'long' versions of these utils? It would make quick benchmarking in CLI so much easier.

image

AI text below:

On ia16 int is 16 bits, so counts above 32767 overflowed: head -n and tail -n/-c/+N parsed with atoi into int, and uniq -c wrapped its duplicate counter. wc/grep/du/sum/cksum already use long.

Verified on ELKS: yes | head -40000 | wc -l now prints 40000 (was empty), uniq -c prints 40000 (was negative), and tail +79000c skips correctly (byte-exact output).

### DIFF
--- a/elkscmd/minix1/uniq.c
+++ b/elkscmd/minix1/uniq.c
@@ -65,10 +65,10 @@ static int equal(char *s1, char *s2)
   return !strcmp(skip(s1), skip(s2));
 }
 
-static void show(char *line, int count)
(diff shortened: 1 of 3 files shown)


ghaerr:
Thanks for the enhancements!

Issue #2741, opened 2026-07-12
Add AMD PCnet/LANCE Ethernet driver (VirtualBox default NIC, QEMU -device pcnet)


duzenko:
Hi, I have a PR-ready branch that implements driver for PCnet-family network cards.
This is mainly to support VirtualBox elks VMs. The big problem I have on Windows that when elks triple faults it typically brings down the Linux VM that hosts QEMU. And QEMU on Windows is not using VT-x, so software emulation does not crash on PM exceptions. So this driver would allow small separate elks VMs running with memory protection enforcement.
I understand that you're happy with QEMU / pcbox emulation however on Windows it's more practical to use VT-x direct software, such as VirtualBox/VMware.
There are also hypothetical physical PCs with these network cards, but that has secondary purpose.

Should I raise a PR for review or you're not interest in more AI-generated code at this point and I'll leave it in my fork?

### AI text below (ignore the github-generated links, that's just `#1`):

Summary

ELKS has no driver for the AMD PCnet family, which is VirtualBox's default emulated NIC (PCnet-PCI II / PCnet-FAST III) and is also available in QEMU via -device pcnet. So ELKS currently has no out-of-the-box networking under VirtualBox. I'd like to add a driver for it.

The device

- AMD PCnet-PCI family: Am79C970 / 970A / 973 (PCI id 1022:2000), plus ISA LANCE boards (Am7990/79C960, NE1500/NE2100 class) which share the same programming model.
- Driven in 16-bit LANCE mode (SWSTYLE 0): register access through the RAP/RDP I/O ports, with descriptor rings and buffers in main memory addressed by 24-bit physical addresses — fully 8086-compatible.
- The only 32-bit code is the optional PCI configuration probe (mechanism #1, ports 0xCF8/0xCFC), guarded to 386+. On 8086/286 the PCI scan is skipped and an ISA board is configured with le0=<irq>,<port> in /bootopts, so the driver still works there.

Integration

Registered as a new ETH_LANCE device (le0), following the existing wd/el3 driver framework. Touches netstat.h, eth.c, main.c (netif_parms), init.h, ports.h, net/config.in and the net Makefile. net=le0 in /bootopts and net start le0 work; DHCP (when LOCALIP is unset) leases correctly.

Testing

Verified with a full telnet login round-trip (login → shell → command → exit):
- Real mode and 286 protected mode
- QEMU (-device pcnet) and on real VirtualBox PCnet-FAST III (Am79C973)
[...]

PR #2756, opened 2026-07-17
[net] Add AMD PCnet/LANCE Ethernet driver

duzenko:
Per suggestion here: https://github.com/ghaerr/elks/issues/2741#issuecomment-5006819155.

See #2741 for original full description of driver capabilities.

AI text below:

Supports the PCnet-PCI family (Am79C970/970A/973, as emulated by VirtualBox and QEMU '-device pcnet') and, via /bootopts, ISA LANCE boards which share the same programming model. Driven in 16-bit LANCE mode (SWSTYLE 0), 8086-compatible; the optional PCI config probe uses 32-bit port I/O and is skipped below a 386.

4-descriptor TX ring with per-slot lazily allocated buffers. Since VirtualBox (like real hardware) completes TX asynchronously, a fast sender can catch the ring full: ktcp's eth_write now waits ~1ms once and retries instead of dropping the frame and stalling on the 3s TCP RTO. Measured on VirtualBox/VT-x: telnet yes-flood 15.6MB/10s, zero stalls, time meminfo 0.01s (3+s before).

### DIFF
--- a/elks/arch/i86/drivers/char/eth.c
+++ b/elks/arch/i86/drivers/char/eth.c
@@ -15,6 +15,7 @@
 extern struct file_operations ne2k_fops;    /* 0 CONFIG_ETH_NE2K */
 extern struct file_operations wd_fops;      /* 1 CONFIG_ETH_WD */
 extern struct file_operations el3_fops;     /* 2 CONFIG_ETH_EL3 */
+extern struct file_operations pcnet_fops;   /* 4 CONFIG_ETH_PCNET */
 
 struct eth eths[MAX_ETHS];
 
@@ -109,5 +110,9 @@ void INITPROC eth_init(void)
 #ifdef CONFIG_ETH_EL3
     eths[ETH_EL3].ops = &el3_fops;
     el3_drv_init();
+#endif
+#ifdef CONFIG_ETH_PCNET
+    eths[ETH_LANCE].ops = &pcnet_fops;
+    pcnet_drv_init();
 #endif
 }
--- a/elks/arch/i86/drivers/net/Makefile
+++ b/elks/arch/i86/drivers/net/Makefile
@@ -37,6 +37,9 @@ endif
 ifeq ($(CONFIG_ETH_EL3), y)
 OBJS += el3-asm.o el3.o
 endif
+ifeq ($(CONFIG_ETH_PCNET), y)
+OBJS += pcnet.o pcnet-asm.o
+endif
 
 all: net_drv.a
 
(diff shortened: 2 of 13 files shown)


ghaerr:
Hi @duzenko , can you delete the following whole source-file changes, they are older than the current ELKS master:
- pmode.S, pmode.c, dhcp.8

Also please delete the deveth.c enhancement from this PR, I would like that in a separate PR for tracking purposes, and will be creating that myself.

In the interests of moving forward, I plan on committing the LANCE driver and then making some minor mods. These may include calling/configuring it as CONFIG_ETH_LANCE etc rather than PCNET (or possibly vice-versa), to be consistent with ETH_LANCE and other defines and filenames. But that will be done after commit. This should allow you to run the PM kernel on VB with full speed networking.

Thank you!

duzenko:
Requested files reset.
Formal note regarding the 386 PCI initialization code: I completely understand and support the no 32 bit code rule for elks. This one exception is to confirm that rule - it's an externally enforced VM compatibility necessity, isolated to a specific VM driver.
As soon as someone discovers how to use VB/VMware with 16-bit ISA cards, I'll be happy to replace it with that.

ghaerr:
> Requested files reset.

Looks good, I plan on committing this later this afternoon/evening when I have a bit more time to make small mods, as well as create the deveth retry fix.

> I completely understand and support the no 32 bit code rule 

Agreed I would not have accepted this had it not been contained and protected for 386+ PCI initialization only within a driver. However, given that certain cards in more modern PCs (e.g. 386+ CPUs) have more modern methods of self-configuration (i.e. PCI that require 32-bit CPU access), I figure why not support them, providing that ELKS itself nor the driver after configuration is complete require 286 or 386+ code. Hopefully after commit others with an ISA Lance card can test the driver and we can get more familiar with real mode and real hardware operation. I also plan on installing VB myself for network and other testing - QEMU alone has its own problems and BTW there are some major bugs that require the use of (-singlestep or -accel tcg,one-insn-per-tb=on due to unfixed bugs in the TCG JIT code generator).

Other issues with 386+ code is that none of our current tools (`disasm`, etc) support non-8086 code. However, we do have 386+ code for initializing "unreal" mode required for use by XMS, as well as 286-specific code implementing the undocumented LOADALL instruction for XMS on 286. So it's not without precedence, even within the kernel.

Frankly, I haven't had but hardly any time to study this driver. Given that its implementation is required for VB that emulates networking, I view this as a big enhancement for more accurately testing the regular kernel, the PM kernel, and ELKS networking in greater variety, something that will definitely help ELKS and its users in the longer term. I look forward to studying the driver more as well as PCI in general.

Thank you!

PR #2759, opened 2026-07-18
Elks dmesg implementation take 2

djohanse:
This pr addresses @ghaerr review comments for PR #2744 -- the initial attempt at a dmesg implementation.

Enqueue/dequeue functions are reentrant/thread safe
Buffer size is configurable via dmesg= parameter in bootopts  
Buffer is stored at top of memory
Uses ioctl instead of /dev/kmsg
User mode dmesg implements display function
Minimal kernel-side code

Note that opencode estimates this implementation takes about 200 bytes of kernel space. Which I believe is about 10% of what's left.  Assuming this implementation checks all the other boxes from the previous PR comments, would it have enough value for end users to justify its cost in kernel space?

Test artifacts:

cat /bootopts, dmesg=2048:

Screenshot 2026-07-17 at 10 35 32 PM

dmesg:
Screenshot 2026-07-17 at 10 36 02 PM

cat /bootopts, dmesg=20

Screenshot 2026-07-17 at 10 39 07 PM

dmesg (bottom of screenshot):

Screenshot 2026-07-17 at 10 41 20 PM

cat /bootopts, dmesg commented out:

Screenshot 2026-07-17 at 10 43 49 PM

dmesg (bottom of screenshot):

Screenshot 2026-07-17 at 10 44 18 PM


ghaerr:
Hello @djohanse,

Well, you're getting closer.

> would it have enough value for end users to justify its cost in kernel space?

In order to get a feature as conceptually simple as `dmesg` into the ELKS kernel, we need to think *small*, and this small feature should as unobtrusive as possible, both in terms of its affecting current readability of source code, as well that the implementation be very small. Both need work - this feature can be implemented in the kernel with two global variables, a C structure, a single queuing function written in C and a couple other lines of code; no ifdefs, no messy volatile variables, no ASM, no extra source files or extra headers. It will be very fast and very small.

I think you should consider taking your current AI-generated kernel code, adding a few functions by hand, and then deleting a lot of it. We already have character queuing functions in the kernel for use as a working template. More code can be moved to the user-mode dmesg.c, and AI can be used there if wanted. But kernel code needs hand crafting. If you're interested, I will post a code review on what needs changing for this to get committed.

Thank you!

djohanse:
> Hello @djohanse,
> 
> Well, you're getting closer.
> 
> > would it have enough value for end users to justify its cost in kernel space?
> 
> In order to get a feature as conceptually simple as `dmesg` into the ELKS kernel, we need to think _small_, and this small feature should as unobtrusive as possible, both in terms of its affecting current readability of source code, as well that the implementation be very small. Both need work - this feature can be implemented in the kernel with two global variables, a C structure, a single queuing function written in C and a couple other lines of code; no ifdefs, no messy volatile variables, no ASM, no extra source files or extra headers. It will be very fast and very small.
> 
> I think you should consider taking your current AI-generated kernel code, adding a few functions by hand, and then deleting a lot of it. We already have character queuing functions in the kernel for use as a working template. More code can be moved to the user-mode dmesg.c, and AI can be used there if wanted. But kernel code needs hand crafting. If you're interested, I will post a code review on what needs changing for this to get committed.
> 
> Thank you!

Hello @ghaerr,

I'm definitely interested.  Too invested in this to walk away now…

ghaerr:
Allright, here you go, here's basically the entire kernel implementation in a few lines:
```
unsigned int dmseg;     /* size of dmesg buffer in K */
seg_t   dmesg_seg;      /* segment or selector of dmesg circular queue */

struct dmesg_queue {
    unsigned int     len;    /* # chars in queue */
    unsigned int     size;   /* queue size */
    unsigned int     head;
    unsigned int     tail;
    unsigned char   base[]; /* queue data follows */
};

void dmesg_addch(int c)
{
    struct dmesg_queue __far *q = _MK_FP(dmesg_seg, 0);

    clr_irq();
    q->base[q->head] = c;
    if (++q->head >= q->size)
        q->head = 0;
    if (q->len < q->size)
        q->len++;
    set_irq();
}

void INITPROC dmesg_init(void)
{
    struct dmesg_queue __far *q;

    if (dmesg) {
        if (dmesg > 63) dmesg = 63;
        dmesg_seg = memend - (dmesg << 6);
        memend -= (dmesg << 6);

        q = _MK_FP(dmesg_seg, 0); 
        q->size = (dmesg << 10) - 4 * sizeof(int);
        q->len = q->head = q->tail = 0;
    }   
}       
```
[EDIT: I corrected a couple errors above, code is untested but should work now.]

Get rid of all the ifdefs, remove configuration option, delete extra header files. Put dmesg_addch in printk.c, put dmesg_queue in kernel.h, call dmesg_init inside early_kernel_init.

Delete the ASM, delete /dev/kmsg stuff, delete all the MEM_READKMSG stuff (that will be moved into dmesg.c user app), rename MEM_GETKMSG to MEM_GETDMESG and return dmesg_seg (basically like you have that).

You'll notice we don't call some things dmesg then others kmsg, etc, that's all confusing. Simple variable names, simple struct member names; AI hurts readability by adding underscores and ifdefs everywhere, no good. Humans do much better with simplicity, until handling complexity really matters.

Creation through deletion! I love it. You'll see that dmesg=2 now specifies 2 Kbytes, which eliminates more unnecessary calculations. In kputchar, just say "if (dmesg_seg) dmesg_addch(c)".

Then slightly rewrite dmesg.c user app to read the dmesg queue directly, rather than having the kernel transfer all the bytes, etc. BTW, dmesg should likely *not* update the tail pointer, just let it show the circular queue contents, unless perhaps an option is passed to "read and remove" from the queue. This allows `dmesg` to be run multiple times before say `dmesg -c` runs to clear the queue.

Hopefully this all makes sense! If you'd like more input, please let me know.

ghaerr:
@djohanse In the code above, for convenience I used the tested chqueue code from lib/chqueue.c; that ring buffer drops characters when the queue is full. For dmesg, we'll just keep adding characters and drop the oldest character when the queue is full. I've adjusted the code to reflect that.

elkscmd/sys_utils/dmesg.c
```diff
@@ -0,0 +1,62 @@
...
+    n = kr.count;
+    {
+        char *p = buf;
+        while (n > 0) {
+            int w = write(1, p, n);
+            if (w <= 0) break;
+            p += w;
+            n -= w;
+        }
+    }
```
ghaerr: To read the dmesg queue, construct a far pointer using the returned MEM_GETDMESG segment or selector. ~~You'll then want to have two while loops, one from the tail to the current head, and another from the head to the tail. These loops will read up to MIN(q->size, DMESG_BUFSIZE) bytes from the queue, then write them to FILE_STDOUT. Which order they run in will depend on whether tail > head or not.~~ In that way, the fewest number of `write` syscalls are required to display any size dmesg queue (up to max 64K bytes, of course).

[EDIT: Actually as I think about this more, if the dmesg queue is going to be truly circular, then we don't need a tail pointer at all. The head pointer always points to the next character position to be added, and the `dmesg` program will read the q->len bytes before q->head, with reverse-wrapping using q->size and q->len to find the start character location of the valid q->len bytes in the buffer to display. This might be a good use of your AI if you don't want to write that yourself. To clear the buffer, just set q->len = 0.]

Something like the following might work in dmesg.c:
```
#define DMESG_BUFSIZ    2048
static char localbuf[DMESG_BUFSIZ];

void dmesg_write(unsigned char __far *farbuf, unsigned int count)
{
    unsigned int n;

    while (count) {
        n = MIN(count, DMESG_BUFSIZ);
        fmemcpy(localbuf, farbuf, n);
        write(STDOUT_FILENO, localbuf, n);
        count -= n;
    }
}

void dmesg_read(void)
{
    unsigned int tail, n;
    struct dmesg_queue __far *q = _MK_FP(dmesg_seg, 0);

    if (!q->len)
        return;
    tail = (q->head - q->len) % q->size;    /* start read pointer */
    n = q->size - tail;                     /* bytes to end of buffer */
    dmesg_write(&q->base[tail], n);
    n = q->len - n;                         /* bytes from start of buffer */
    dmesg_write(&q->base[0], n);
}
```

elkscmd/sys_utils/dmesg.c
```diff
@@ -0,0 +1,62 @@
...
+int main(void)
+{
+    int fd;
+    unsigned int kmsg_seg;
+    struct kmsg_read kr;
+    int n;
+
+    fd = open("/dev/kmem", O_RDONLY);
+    if (fd < 0) {
+        write(2, "dmesg: cannot open /dev/kmem\n", 29);
```
ghaerr: Use an `errmsg` macro like the following so that humans don't have to count characters when they want to maintain the code by changing the message:
```
```

Error strings can be verbose, but recommend keeping them small so that the entire `dmesg` binary fits in 1K. This will allow it to be put on smaller floppy images.

elkscmd/sys_utils/dmesg.c
```diff
@@ -0,0 +1,62 @@
...
+        return 1;
+    }
+
+    if (ioctl(fd, MEM_GETKMSG, &kmsg_seg) < 0) {
+        write(2, "dmesg: kernel message log not enabled\n", 38);
```
ghaerr: This check is no longer needed, as the ioctl MEM_GETDMESG will always work.

elkscmd/sys_utils/dmesg.c
```diff
@@ -0,0 +1,62 @@
...
+        close(fd);
+        return 1;
+    }
+
+    if (kmsg_seg == 0) {
+        write(2, "dmesg: kmsg buffer not configured (add dmesg= to /bootopts)\n", 60);
```
ghaerr: Rename everything "dmesg*", saying "kmsg buffer" to a user is confusing. Same for variable names in this program IMO.

elkscmd/rootfs_template/bin/sys
```diff
@@ -93,6 +94,7 @@ copy_bin_files()
 	cp /bin/cp	$MNT/bin
 	cp /bin/date	$MNT/bin
 	cp /bin/df	$MNT/bin
+	cp /bin/dmesg	$MNT/bin
```
ghaerr: Delete this as `dmesg` won't likely be on the 360k image, which is what this section of /bin/sys copies.

elkscmd/sys_utils/dmesg.c
djohanse: Hello @ghaerr,

getting compile errors.  Looks like fmemcpy* and _MK_FP are not available in user space?

djohanse: could use lseek/read which would keep everything in user space

ghaerr: The \_MK_FP macro is defined in a kernel header, like its also used in similar programs like ps.c and meminfo.c:
```
#define _MK_FP(seg,off) ((void __far *) ((((unsigned long)(seg)) << 16) | ((unsigned int)(off))))
```
`fmemcpy` is basically `memcpy` with far pointers:
```
void fmemcpy(char __far *dst, char __far *src, unsigned int count)
{
    while (cnt--)
        *dst++ = *src++;
}
```


> could use lseek/read which would keep everything in user space

Yes, you could alternatively `read` /dev/kmem, just like ps and meminfo. You'll want to set the `lseek` offset the same way, which trickily uses a \_MK_LP macro for the long offset just like the above example, which then also works for real mode segments and protected mode selectors. See `memread` in elkscmd/sys_utils/ps.c for an example. 
```
#define _MK_LP(seg,off) ((unsigned long)((((unsigned long)(seg)) << 16) | ((unsigned int)(off))))
```

djohanse:
Updated per review comments.  On the kernel side, incorporated review comments as noted.  On the user side, I did have open code rewrite dmesg.c using ps.c as an example.

dmesg=2
Screenshot 2026-07-24 at 12 21 52 AM

dmesg=0
Screenshot 2026-07-24 at 12 24 25 AM

dmesg commented out
Screenshot 2026-07-24 at 12 26 46 AM

dmesg=75
Screenshot 2026-07-24 at 12 30 17 AM

ghaerr:
Hello @djohanse,

Thanks for completing this! The kernel portion looks good. Last I checked, the resident kernel implementation using this method was just over 100 bytes. Shows the difference hand crafting a micro version can accomplish, huh?

I believe there may be an issue with the dmseg.c user program overflowing `buf` when q->len > sizeof(buf), it doesn't look like there's any code to handle that.  I suspect that the initial kernel messages queue used length never exceeds 2048, so the wrap was never tested, regardless of total queue size.

An easy way to produce more kernel printk output is to enable CONFIG_TRACE=y then type ^N or ^O to display the kernel inode or buffer debug lists. I'll commit this now and test that.

There are a few other cleanups that can be done, as well as some changes to allow this to work in the experimental PM kernel, which requires using a selector rather than segment in q->seg; I'll post a PR with those changes for your information.

Thank you!

PR #2761, opened 2026-07-20
[kernel] Execute user programs in XMS memory using PM kernel

ghaerr:
ELKS is now no longer limited to running user programs in the lower 640K of memory!

Using the (experimental) protected mode kernel and running on a 286 or 386+ CPU, XMS memory above 1M is now usable for the code and data segments of user programs, where previously the system would run out of memory.

This PR implements a first-pass at adding XMS memory to the internal main memory manager, allowing the kernel to allocate from main or XMS memory for user program execution. XMS memory is still used for buffer allocations, and the amount remaining will be usable for program execution.

~~In this PR, only 1M of XMS memory is accessible for program execution, due to technical issues storing memory segment sizes larger than 1M (the size is stored in paragraphs in a 16-bit word and would overflow). An update allowing use of all XMS memory will come after more design considerations.~~ All of XMS memory is now available for   user program execution.

User program binaries compiled with either ia16-elf-gcc or OpenWatcom should work unmodified, provided that the program doesn't attempt to perform segment arithmetic internally. Most small and medium model programs will all run, while large model programs using normal \_\_far pointers should also work, unless special code is written to attempt to access segment offsets > 64K or use `fmemalloc` to allocate > 64K. The ELKS PM kernel allocates a single selector for each real mode segment in most cases, and an paragraph count cannot be directly added to a selector, unlike in real mode where it is permitted by the CPU. Up to 64K of offset can be used with any selector, just like real mode segments.

For ease of testing, this first commit contains code making the kernel think the size of main memory is only 256K, which forces XMS allocation soon after login. Then, most user programs will be running in XMS, also allowing deeper testing of the memory manager, which was updated to handle 24-bits of memory (=15MB XMS).

Some cleanup was done in a couple drivers changing segext_t types to segoff_t, for correctness. The first commit also contains lots of debug code that will be removed.

### DIFF
--- a/elks/arch/i86/drivers/char/console-direct.c
+++ b/elks/arch/i86/drivers/char/console-direct.c
@@ -67,7 +67,7 @@ struct console {
     void (*fsm)(struct console *, int);
     seg_t vseg;                 /* current render target: video page seg when foreground,
                                    ram_seg when backgrounded in RAM-buffer mode */
-    segext_t vseg_offset;       /* offset into vseg for buffer ram calculations */
+    segoff_t vseg_offset;       /* offset into vseg for buffer ram calculations */
     unsigned int crtc_offset;   /* CRTC start address (chars) for this VC's video page;
                                    only meaningful when this VC is video-page-backed */
     unsigned short crtc_base;   /* 6845 CRTC base I/O address */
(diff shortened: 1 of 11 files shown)


ghaerr:
This implementation is technically discussed in more detail in https://github.com/ghaerr/elks/issues/2718#issuecomment-5013000423.

Issue #2764, opened 2026-07-22
sleep() is too unprecise


toncho11:
sleep() takes microseconds as argument, but it can be as unprecise as 20ms?

I have tried 1ms, so sleep(1000), but I get more. It seems a 10 ms timer quantum plus one extra 10 ms tick for every positive usleep() call.

Also gettimeofday() is a bit better, but also in the 10ms range of error.

So there is actually no way to have a 1-2 ms sleep on ELKS?

ghaerr:
> sleep() takes microseconds as argument

`usleep` takes microseconds, `sleep` takes seconds.


> So there is actually no way to have a 1-2 ms sleep on ELKS?

Not by actually sleeping, since the IBM PC hardware clock interrupt fires only once every 10ms.

If busy-looping is an option, your application can try using the same technique as the ELKS precision library timer uses - actually reading the hardware Programmable Interval Timer countdown register, which ticks every 0.838 microseconds. This is a little complicated, see libc/debug/prectimer.c for details. Basically, you would compile your program setup for precision timer, then busy loop the CPU until the PIT countdown elapsed. It's all complicated since the PIT countdown wraps, but the upper-level `get_ptime` handles that.

Note that busy-looping still isn't guaranteed, since your application may get pre-empted in the next 10ms timer interval (or not).

toncho11:
I was using only usleep.

Consider exposing the precision timer from libc/debug/prectimer.c as a small public libc API, for example:

int ptime_init(void);
unsigned long ptime_us(void);

ptime_us() would return a monotonic microsecond timestamp and handle PIT wrap internally.

I need it for:
- calculate accurate dt in a callback loop
- implement short 1–4 ms busy-wait pacing
- use absolute frame deadlines to avoid timing drift

ghaerr:
> exposing the precision timer from libc/debug/prectimer.c as a small public libc API, for example:

The precision timer is already a public API, no need for a separate `ptime_us`, you can write that using a three-line while loop using `get_ptime` for however you need it. For your own uses, just get the initial ptick count with get_ptime, then loop calling get_time gain until ptick > "the delay you're looking for". Its no big deal for your own application programs to use it, but would not be a great idea to add a busy-loop to a library, as users may not realize they're grabbing the CPU, and it isn't guaranteed to be accurate.

toncho11:
OK, but it is not exposed for OWC libc and I am using OWC to compile LuaGFX.
[...]

PR #2768, opened 2026-07-23
[owlibc] Add precision timer library to ELKS OpenWatcom C library

ghaerr:
Adds `init_ptime` and `get_ptime` into ELKS' C library for OpenWatcom.

Requested by @toncho11 in #2764.

@toncho11: this is untested, but I have reviewed the ASM output from the compiled prectimer.c and it looks good, as far as having had to add the ASM rmacros/routines that weren't previously present to allow this to be compiled. Recompile the entire C library with OWC and hopefully you should be good to go.

### DIFF
--- a/elks/include/arch/io.h
+++ b/elks/include/arch/io.h
@@ -128,4 +128,16 @@
 
 #endif /* __ia16__ */
 
+#ifdef __WATCOMC__
+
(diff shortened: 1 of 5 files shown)


toncho11:
Thank you!

PR #2769, opened 2026-07-23
PCnet ISA default

duzenko:
In the original PR comments I pledged to revert the 32-bit code as soon as an alternative is discovered.
Soon enough I learned that a particular VirtualBox PCnet model is in fact an ISA-style device. It does not show in the PCI config and needs/supports ISA-like configuration.
So I think I owe this driver rewrite:
- PCI config asm with 32-bit instructions - removed (can be restored later from history if need arises)
- ISA VBox default config - hardcoded
- bootopts override for QEMU/physical cards - retained
- stability tests - basic passed, general kTCP issues under investigation

QEMU note: PCnet on it seems to be PCI-only, and does not work out of the box (needs bus master activation via 32-bit port). Workaround for now - enable network boot in BIOS that initializes the card in real mode and it keeps working in elks afterwards, at the cost of a few seconds of boot time. 
Alternatively I can do try BIOS INT voodoo if in-elks activation is preferred.


ghaerr:
TBH, I'm not sure this is really an enhancement worth committing. The reasons are 1) the 32-bit code now is fully protected from execution and allows for dynamic configuration, which is well worth it IMO, 2) waiting 2 seconds for QEMU is a non-starter, and 3) just changing LANCE_PORT from 0 to 0x300 and LANCE_IRQ to 3 in ports.h has exactly the same effect, without losing anything gained from the first two reasons (or just editing /bootopts le0= to the values wanted).

We already have 80286 and 80386 code in the kernel, for XMS support, and now are semi-supporting a PM kernel which requires >= 80286, which has definite advantages should you get C86 compiling itself with it, etc.

So why delete 80386 code from this driver and lose dynamic networking capability?

duzenko:
I'm fine either way, I just assumed we only ever accept absolutely necessary 386 code, and PCI config section is strictly not necessary any more.

As for the 2 second wait, QEMU elks has ne2k NIC by default, and there is no wait in that config. And I can try to add BIOS initialization instead of 386 assembler for QEMU + PCnet.

So which way? Close PR / BIOS init on qemu / others ?

ghaerr:
Thank you for concern about 386 instructions. I like being able to quickly test the PM kernel with QEMU, and think a direct auto-configuration would be superior to using the BIOS. Thus, I would say your PCnet driver is already quite good, so go ahead and close this PR, thank you.

PR #2772, opened 2026-07-26
implementation of logout

djohanse:
### login user1

Screenshot 2026-07-25 at 10 30 48 PM

### user1 has read-only permission to /etc/passwd

Screenshot 2026-07-25 at 10 31 46 PM

### user1 processes:
Screenshot 2026-07-25 at 10 32 29 PM

### logout.  user1 processes stopped:
Screenshot 2026-07-25 at 10 33 48 PM

### root has write access to /etc/passwd:
Screenshot 2026-07-25 at 10 34 26 PM


ghaerr:
Hello @djohanse,

This `logout` program just sends a SIGHUP to all processes associated with a given user ID, right?

Reading the system task list using /dev/kmem and then sending SIGHUP to all matching process IDs is a pretty hacky way to do that, and isn't really correct; since there's no way to stop multitasking, the system task array may change while `logout` is executing and then it sends `kill` signals to the wrong processes. Thus, I'd like to explore other ways to do this.

The `kill` syscall will send a SIGHUP signal to all processes associated with a PID's process group if the PID is passed as a negative number (e.g. -PID). I think this is what you want, right? A process group is slightly different than every task running under a PID, but excludes those processes that have detached themselves using `setsid` (set session ID), which will usually exclude daemons started (e.g. ftpd, etc). I think this is more closely associated with what you want in `logout` anyways, right?

The current user can always just send SIGHUP to all its current processes by passing PID=0 to kill.

Here's the associated sys_kill code for your reference:
```
    if ((int)pid < 1)
        return kill_pg((!pid ? current->pgrp : -pid), sig, 0);
```

To find the login shell (or more accurately, the controlling process of the process group) of a user, what we need to do is loop through the system process list looking for a matching PID and also PID == PGRP. Then send a single kill to that process group with -PID.

What do you think, will this accomplish your goal for `logout`? 

The logout.c program should probably use the `errmsg` macro as used in dmesg.c to decrease the size of the system program; otherwise, using `fprintf`, especially without tiny_printf.c, drags in lots of the stdio library. Using `errmsg` results in the smallest possible size, and is useful when the caller only otherwise uses constant strings or an occasional %s format specifier.

Thank you!

djohanse:
Updated per review comments.

1. root logs out user1

Screenshot 2026-08-02 at 11 32 16 AM


Screenshot 2026-08-02 at 11 32 35 AM



2 user1 logs out self

Screenshot 2026-08-02 at 11 34 10 AM


Screenshot 2026-08-02 at 11 34 36 AM

ghaerr:
Hello @djohanse,

Looks great, thanks for the changes.

ghaerr:
After commit, I find the following warning during compilation of logout.c:
```
logout.c: In function 'main':
logout.c:70:9: warning: implicit declaration of function 'ioctl' [-Wimplicit-function-declaration]
     if (ioctl(fd, MEM_GETDS, &ds) < 0 ||
         ^~~~~
```
In the future, please check to see there are no compilation warnings. I've fixed this with \#include <sys/ioctl.h> in 36fedc6704929e1964be2f9adfea5de2ec774e50.

djohanse:
> After commit, I find the following warning during compilation of logout.c:
> 
> ```
> logout.c: In function 'main':
> logout.c:70:9: warning: implicit declaration of function 'ioctl' [-Wimplicit-function-declaration]
>      if (ioctl(fd, MEM_GETDS, &ds) < 0 ||
>          ^~~~~
> ```
> 
> In the future, please check to see there are no compilation warnings. I've fixed this with #include <sys/ioctl.h> in [36fedc6](https://github.com/ghaerr/elks/commit/36fedc6704929e1964be2f9adfea5de2ec774e50).

Will do, sorry about that

PR #2773, opened 2026-07-26
[App] Add dmesg and screen to fd1232c

tyama501:
Hello @ghaerr and @djohanse 

I have confirmed that dmesg works on PC-98.

Screenshot_20260726_140749

I have added 1232c label to the dmesg on the application list.

Also I have been using the screen since PC-98 does not have the switching console ability.
So I put it back to 1232c image.

Thank you.

### DIFF
--- a/elkscmd/Applications
+++ b/elkscmd/Applications
@@ -104,8 +104,8 @@ sys_utils/console               :sysutil                :1200k
 sys_utils/beep                  :sysutil                :1200k
 sys_utils/decomp                :sysutil         :360c              :1440k
 sys_utils/sysctl                :sysutil                :1200k
-sys_utils/dmesg                 :sysutil                            :1440k
-screen/screen                   :screen                             :1440c
+sys_utils/dmesg                 :sysutil                :1232c      :1440k
+screen/screen                   :screen                 :1232c      :1440c
 cron/cron                       :cron                   :1200k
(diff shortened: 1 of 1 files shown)


ghaerr:
Hello @tyama501,

I remembered that you have been using `screen` for multiple displays after I submitted my change, sorry about that!

> PC-98 does not have the switching console ability.

Are you aware of the contribution in #2671 and #2716 for the IBM PC direct console which allows the use of RAM-backed consoles? This was written primarily to accommodate up to 8 consoles on the IBM PC, instead of the 4 (or less) that will fit in the EGA/CGA/MDA display memory. It has been updated since its initial contribution to support allocating console from main memory using `seg_alloc` rather than `heap_alloc`.

I am thinking you might be able to modify that code and use it in your PC-98 direct console driver to implement multiple consoles, all backed up in RAM, and finally implement /dev/tty2-tty4 or more on PC-98 without requiring the use of `screen`.

If interested, you might look at the VC_USE_RAM_BUFFER part of direct-console.c. The idea would be to always have `UseRambuf` on, which effectively allocates screen buffers out of main memory, then uses `fmemcpy` to copy them onto the single physical console when the user types Alt-F1/2, etc.

Thank you!

tyama501:
Thank you!
I will look it.

Issue #2767, opened 2026-07-23
ftpd: passive mode fails on QEMU when the guest is bridged


duzenko:
Sorry, but another edge case which may or may not merit your attention.
Is there any room for flexibility in qemu/ftp use case? I'm trying to do some stress testing on kTCP and noticed that elks ftp makes certain assumptions.
My AI agent found and explained how to work around it but I'd think we might want to tighten the logic here anyway.
Specifically: additionally to discovering the QEMU hypervisor, ftpd should also check that the local IP address is in the typical NAT subnet. If not behind NAT, then don't activate that special qemu mode.

PR to follow, pending your general approvement. Expected scope: one/few lines of code around https://github.com/ghaerr/elks/blob/0ad19e9f9330db09ef7d80fded82517210fbbe0e/elkscmd/inet/ftpd.c#L705

AI text below:

**Description**
- What is expected ? With a networked ELKS guest that has a routable IP (bridged, not slirp), an FTP client using passive mode should be able to upload/download — the same as it does under VirtualBox, and the same as active mode does under QEMU.
- What is observed ? Under QEMU, passive transfers fail: ftpd's PASV reply advertises 127.0.0.1 and a remapped port (8041+N) that only works behind QEMU slirp + hostfwd. On a bridge there is no hostfwd, so the client connects to a port nothing is listening on (ftpd actually listens on 49821+N). Active mode works; the identical image under VirtualBox works in passive mode.

**Configuration**
- Version (commit or release): upstream master 0ad19e9f (2026-07-23); elkscmd/inet/ftpd.c + elks/init/main.c are stock.
- Emulator or HW platform: QEMU 11.0.1, ne2k_isa, NIC attached to a host bridge/TAP (routable IP, e.g. 192.168.52.7) instead of -netdev user (slirp). Contrast host: VirtualBox (no QEMU fw_cfg), passive works.

**How to reproduce ?**
- Does the problem always happen? Yes — deterministic whenever a QEMU guest with a bridged/routable IP is contacted by a passive-mode FTP client.
- Steps:
  a. Boot a networked ELKS image under QEMU with the NIC bridged (TAP), so the guest gets a routable IP from a real DHCP server (not 10.0.2.15).
  b. From another host on that segment: curl -T file "ftp://<guest-ip>/x" (curl defaults to passive).
  c. Transfer fails on the data connection. curl -P - ... (active) succeeds. Same image under VirtualBox succeeds in passive.

**Raw data**
[...]

PR #2778, opened 2026-07-29
Remove qemu hack from ftp/ftpd

djohanse:
Removes "QEMU hack" from ftp and ftpd.  Adds -n parameter to ftpd for specifying ip address advertised in passive mode.  Adds -P parameter for specifying port range (needed for slirp passive mode)

Test artifacts: 

### ftpd configured for slirp, ports 49152-49153.  Note you must also set up corresponding port forwarding in qemu.sh

Screenshot 2026-07-28 at 9 33 47 PM

### slirp, elks internal, ftp localhost, passive mode:
Screenshot 2026-07-28 at 9 36 18 PM

### slirp, elks internal, ftp localhost, active mode not supported
note this is correct behavior.  I think I previously said or inferred that active mode worked.  It does not and is not expected to for slirp connection.  I was mistaken.
Also the "no such command" was due to a typo.  The ftp server responds with the "use passive mode" message.

Screenshot 2026-07-28 at 9 40 28 PM

### slirp, elks internal, ftp 127.0.0.1
Screenshot 2026-07-28 at 9 42 41 PM

### slirp, elks internal, ftp 10.0.2.15

Screenshot 2026-07-28 at 9 44 02 PM

### configure external server for slirp:
Screenshot 2026-07-28 at 9 48 16 PM

### ftp to slirp gateway (10.0.2.2):
Screenshot 2026-07-28 at 9 50 17 PM

### Mac ftp server also does not support active mode with slirp connection

Screenshot 2026-07-28 at 9 52 27 PM

### inbound to elks, active mode not supported::

Screenshot 2026-07-28 at 9 57 46 PM

### passive mode works:
Screenshot 2026-07-28 at 9 59 44 PM

### Configure ftpd for tap connection:

Screenshot 2026-07-28 at 10 05 40 PM

### ftp localhost, passive mode

Screenshot 2026-07-28 at 10 06 52 PM

### ftp localhost, active mode
Screenshot 2026-07-28 at 10 08 01 PM

### Inbound to elks, active:

Screenshot 2026-07-28 at 10 15 08 PM

### Inbound to elks, passive:
Screenshot 2026-07-28 at 10 15 56 PM

### Configure external server:
Screenshot 2026-07-28 at 10 18 17 PM

### Passive connection:
Screenshot 2026-07-28 at 10 19 15 PM

### Active connection:
Screenshot 2026-07-28 at 10 19 50 PM

### And curl:
Screenshot 2026-07-28 at 10 23 57 PM

### shows ls / before and after curl:
Screenshot 2026-07-28 at 10 24 15 PM

### DIFF
--- a/elks/arch/i86/drivers/block/ssd-xms.c
+++ b/elks/arch/i86/drivers/block/ssd-xms.c
@@ -25,6 +25,7 @@
 /* current implementation requires no other XMS allocations other than XMS buffers */
 static ramdesc_t     xms_ram_base;      /* ramdisk XMS memory start address */
 static unsigned  int xms_ram_size;      /* ramdisk size in Kbytes */
+static segment_s *   xms_seg;           /* saved segment_s for PM dealloc */
 
 /* initialize SSD device */
 sector_t ssddev_init(void)
@@ -43,9 +44,16 @@ int ssddev_ioctl(struct inode *inode, struct file *file,
         debug_blk("SSD: ioctl make %d\n", arg);
         if (xms_ram_size)
             return -EBUSY;
+#ifdef CONFIG_286_PMODE
+        xms_seg = seg_alloc_xms((selext_t)arg << 6, SEG_FLAG_RAMDSK);
+        if (!xms_seg)
+            return -ENOMEM;
+        xms_ram_base = xms_seg->addr << 4;  /* NOTE: selector in seg->base not used */
+#else
         xms_ram_base = xms_alloc(arg);      /* in Kbytes */
         if (!xms_ram_base)
             return -ENOMEM;
+#endif
         xms_ram_size = arg;
         ssd_num_sects = arg << 1;
 
@@ -62,8 +70,13 @@ int ssddev_ioctl(struct inode *inode, struct file *file,
         if (xms_ram_size) {
             invalidate_inodes(inode->i_rdev);
             invalidate_buffers(inode->i_rdev);
+#ifdef CONFIG_286_PMODE
+            seg_free(xms_seg);
+#else
             xms_alloc_ptr -= xms_ram_size;      /* NOTE: no xms_free yet */
+#endif
             xms_ram_size = 0;
+            ssd_num_sects = 0;
             ssd_initialized = 0;
             return 0;
         }
--- a/elks/arch/i86/mm/malloc.c
+++ b/elks/arch/i86/mm/malloc.c
@@ -35,7 +35,18 @@ static byte_t seg_pm_access(word_t type)
 /* allocate selector for passed segment_s structure, 0 = GDT full or limit exceeded */
 static int seg_pm_attach(segment_s *seg, word_t type)
 {
-    seg->base = desc_alloc(seg->addr << 4, seg->size << 4, seg_pm_access(type));
+    selext_t limit = seg->size << 4;
+
+    /* Allow non-XMS selector limit of max 64K for now for automatic 80286 compatibility.
+     * User mode fmemalloc allocations are also limited to 64K max segments
+     * unless 32-bit instructions are used within the application.
+     */
+    if (!(type & SEG_FLAG_XMS) && (limit >> 16) > 1) {
+        printk("seg_alloc: %08lx limit > 64K\n", limit);
+        return 0;
+    }
+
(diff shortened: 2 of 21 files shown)


ghaerr:
Hello @djohanse,

Wow, this is super nicely-done, thank you! And the testing looks wonderful, very much appreciated.

Your code shows a true marvel at what can be done through "creation through destruction", deleting absolutely all of the hack qemu code, made possible by your enhancement to `accept` allowing address/port information through `getsockname` instead. Great job!!!

I assume the new ftpd= line in /etc/net.cfg is required for local PASV operation, correct? For that reason, the -n and -P options are required, right? Will 127.0.0.1 always likely be required for -n? But for bridge operation, both options would be left out? I am trying to find if there's a way to decrease the size of the ftpd command line, as the current length ends up wrapping on `ps` past 80 columns. [EDIT: I suppose a better way would be to just fix `ps` instead to limit output at column 80].

I think the next step should be to commit this as-is, then have @duzenko determine whether this allows his issue in #2767 to be fixed? For that, the `ftpd="ftpd -d -n 127.0.0.1 -P 49152:49153"` would be changed back to `ftpd="ftpd -d"`, correct?

Out of curiosity, why was -n picked for the passive mode IP address? I am thinking it might be easier to remember to use -p instead, which then goes right along with -P for the min:max port numbers. Or possibly if -n/-p is specified without a following address, 127.0.0.1 is assumed.

Let me know what you think, I am ready to commit this on your approval.

Thank you!

duzenko:
I support it as is because my upcoming kTCP PR may conflict with this one if unmerged.
However, you might want to check if the following AI nitpicking makes sense (it also rambled about EPSV, but that's too much even for me):

> 1. CMD_PORT gated on pasv_ip is too broad.
> case CMD_PORT:
>     if (pasv_ip) { send_reply(502, "Active mode not available, use passive mode"); break; }
> 1. Having a masquerade address and being unable to do active mode are independent. Active mode has the server dial out to the client, which is the direction NAT permits — so -n shouldn't disable it. This conflates configuration with capability and should be its own flag, if it's needed at all.
> 2. -P and -n don't bounds-check their argument. ftpd -P with nothing following does argc--; argv++; then dereferences argv[0] past the end. One-line guard each.
> 3. Hardcoded loopback constant in tcpdev_bind: db->addr.sin_addr.s_addr == htonl(0x7f000001) — should be INADDR_LOOPBACK, and it matches only 127.0.0.1 while ip_route() loops back all of 127.x.

ghaerr:
> I support it as is because my upcoming kTCP PR may conflict with this one if unmerged.

Thanks for your quick response. I will commit this now. I am also trying to get #2776 merged, although it has a few changes I wanted to see. I'm thinking of then committing that now, and making the few small changes myself, rather than waiting. I mention this because I want to cleanup requiring one-line changes in /bin/net, /etc/rc.sys and ktcp.c for NIC device names which aren't really useful or needed. Thus there may be more changes to ktcp.c coming very shortly, but only in these "admin" areas.

> you might want to check if the following AI nitpicking makes sense 

@djohanse, if you'd like to make further changes as suggested by @duzenko's AI, go ahead in a subsequent PR.

ghaerr:
@duzenko, if this PR fixes #2767, please go ahead and close that issue, thanks.

ghaerr:
@djohanse, are the following lines still required in qemu.sh:

```
hostfwd=tcp::8041-:49821,\
hostfwd=tcp::8042-:49822,\
hostfwd=tcp::8043-:49823,\
hostfwd=tcp::8044-:49824,\
hostfwd=tcp::8045-:49825,\
hostfwd=tcp::8046-:49826,\
hostfwd=tcp::8047-:49827,\
hostfwd=tcp::8048-:49828,\
hostfwd=tcp::8049-:49829,\
```

djohanse:
> @djohanse, are the following lines still required in qemu.sh:
> 
> ```
> hostfwd=tcp::8041-:49821,\
> hostfwd=tcp::8042-:49822,\
> hostfwd=tcp::8043-:49823,\
> hostfwd=tcp::8044-:49824,\
> hostfwd=tcp::8045-:49825,\
> hostfwd=tcp::8046-:49826,\
> hostfwd=tcp::8047-:49827,\
> hostfwd=tcp::8048-:49828,\
> hostfwd=tcp::8049-:49829,\
> ```

No, we should remove them

PR #2782, opened 2026-07-30
write - bugfix, recepient username is printed instead of sender username

venomix666:
I found this bug in write on Fuzix, so I thought I should upstream the fix here as well.

I have not been able to test the fix on elks as write is not built currently.


ghaerr:
Thank you!

PR #2783, opened 2026-07-30
[console] Enable ANSI emulation on the Amstrad IGA console

parabyte:
## Problem

Selecting the Amstrad IGA console (`CONFIG_CONSOLE_AMSTRAD_IGA`) disables ANSI
escape handling entirely, so escape sequences are printed to the screen as
literal text.

The most visible symptom is tab completion. `ash` redraws the edited line using
CSI sequences, so on a PC1640 completing `ls /bi` produces:

```
# ls /bi<TAB>
^[[10C/bin/^[[0K
```

instead of moving the cursor forward and clearing to end of line. `clear(1)`,
`vi` and any coloured output are affected the same way — this is not specific to
completion, it is every escape sequence.

## Cause

`console-direct-amstrad.c` shares `console.c` with the direct and BIOS console
drivers, and `console.c` is where the escape parser lives:

```c
#ifdef CONFIG_EMUL_ANSI
    case ESC:
        C->fsm = esc_char;
        break;
#endif
```

But `config.in` only offers `CONFIG_EMUL_ANSI` for the direct and BIOS drivers:

```sh
if [[ "$CONFIG_CONSOLE_DIRECT" = "y" || "$CONFIG_CONSOLE_BIOS" = "y" ]]; then
        int  '  Number of virtual consoles (1-8)' CONFIG_CONSOLE_MAX 3
        bool 'Console ANSI terminal emulation'  CONFIG_EMUL_ANSI  y
fi
```

With the Amstrad console selected the question is never asked, so the symbol is
never defined and the `case ESC:` arm is compiled out. `autoconf.h` ends up with
`CONFIG_CONSOLE_AMSTRAD_IGA 1` and no `CONFIG_EMUL_ANSI` at all.

## Fix

Add `CONFIG_CONSOLE_AMSTRAD_IGA` to that test.

`ntty.h` needs the same addition, otherwise `MAX_CONSOLES` falls through to the
hardcoded 3 and `CONFIG_CONSOLE_MAX` would be offered by the configurator but
silently ignored.

## Testing

Built for a PC1640 and run under 86Box with the `pc1640` machine profile
(8086, 640K, MFM disk).

* Before: `ls /bi<TAB>` displays `^[[10C/bin/^[[0K`
* After: `ls /bi<TAB>` completes to `ls /bin/` with the cursor correctly
  positioned, and `clear(1)` clears the screen

Kernel size grows by 1424 bytes on this configuration, which is the escape
parser now being compiled in.

Note that a config change here does not by itself trigger a rebuild of the
console objects — `autoconf.h` regenerates from `.config`, so the object files
must be removed or touched to see the difference.


ghaerr:
Oops, yes this slipped through the cracks on my cleanup commit.

ghaerr:
@parabyte I also added the keyboard type to Amdstrad config in 8e671dd0a72205a08228b0127b373ef6e33bc2cd.

PR #2784, opened 2026-07-31
[net] Display Network down instead of Unknown host before net start

ghaerr:
`ping` called in_gethostbyname before creating socket, which fails with "Unknown host" rather than "Network down" error message.

Also cleans up other unused code in ping.c.

`traceroute` suffers from the same problem, but wasn't as easy to fix. `ifconfig` should also be fixed. Network programs should use the standard error message "Network is down" displayed automatically by `perror("progname")` rather than using "progname: ktcp not running" (except for after connect()). That is, attempt to create a socket first and error with `perror`, then call in_gethostbyname(), to get more accurate error messages.

@djohanse Is this the preferred display for `ping` ms output? It would seem superior to show the elapsed time rather than system time displayed in milliseconds, which ends up being a super long number:
Screenshot 2026-07-31 at 11 01 50 AM

### DIFF
--- a/elkscmd/inet/ping.c
+++ b/elkscmd/inet/ping.c
@@ -55,14 +55,13 @@ static unsigned long get_ms(void)
 
 static int ping_via_ktcp(ipaddr_t target, char *hostname, int count, int interval, int s)
 {
-	unsigned char buf[2 + sizeof(struct icmp_echo_request_s)];
-	struct icmp_echo_reply_s reply;
 	int seq, sent, recv;
 	unsigned long rtt, min_rtt, max_rtt, total_rtt;
 	unsigned short id = getpid();
+	unsigned char buf[2 + sizeof(struct icmp_echo_request_s)];
+	struct icmp_echo_reply_s reply;
 
-	printf("PING %s (%s): 12 data bytes\n",
-	       hostname, in_ntoa(target));
+	printf("PING %s (%s): 12 data bytes\n", hostname, in_ntoa(target));
 
 	sent = 0;
 	recv = 0;
@@ -121,8 +120,8 @@ static int ping_via_ktcp(ipaddr_t target, char *hostname, int count, int interva
 					if (rtt < min_rtt) min_rtt = rtt;
 					if (rtt > max_rtt) max_rtt = rtt;
 					total_rtt += rtt;
-				printf("12 bytes from %s: icmp_seq=%d ttl=%d time=%lu ms\n",
-				       in_ntoa(target), seq, reply.ttl, rtt);
+					printf("12 bytes from %s: icmp_seq=%d ttl=%d time=%lu ms\n",
+						in_ntoa(target), seq, reply.ttl, rtt);
 				} else {
 					printf("Request timeout for icmp_seq %d\n", seq);
 				}
@@ -164,35 +163,27 @@ static int ping_via_ktcp(ipaddr_t target, char *hostname, int count, int interva
 
 static void usage(void)
 {
-	printf("Usage: ping [-c count] [-I local_ip] [-i interval] host\n");
+	printf("Usage: ping [-c count] [-i interval] host\n");
 	exit(1);
 }
 
 int main(int argc, char **argv)
 {
-	ipaddr_t local_ip = 0, target_ip;
+	ipaddr_t target_ip;
 	int count = -1;
 	int interval = DEFAULT_INTERVAL;
-	int ch;
+	int s, ch;
+	struct sockaddr_in local, rem;
 
 	signal(SIGINT, sigint);
 
-	{
-		char *hostname = getenv("HOSTNAME");
-		if (hostname)
-			local_ip = in_gethostbyname(hostname);
-	}
-
-	while ((ch = getopt(argc, argv, "c:i:I:")) != -1) {
+	while ((ch = getopt(argc, argv, "c:i:")) != -1) {
 		switch (ch) {
 		case 'c':
 			count = atoi(optarg);
 			break;
 		case 'i':
 			interval = atoi(optarg) * 1000;
-			break;
-		case 'I':
-			local_ip = in_gethostbyname(optarg);
 			break;
 		default:
 			usage();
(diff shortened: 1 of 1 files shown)


djohanse:
No that's not what I would expect.  This is what I see from yesterday's master:

Screenshot 2026-07-31 at 9 08 06 PM

ghaerr:
> No that's not what I would expect. This is what I see from yesterday's master:

Thanks, I'll check deeper into it. Perhaps this was because I was testing on the PM kernel. I will check and report back.

ghaerr:
>  I will check and report back.

Very trange. Same results on real mode testing:
Screenshot 2026-07-31 at 8 11 32 PM

I suppose this PR broke it? I don't get it, since none of the timing code was touched. I'll keep digging.

ghaerr:
@djohanse, it appears the problem is with the line:
```
                    rtt = now - ntohl(reply.timestamp);
```
in ping.c. The ntohl(reply.timestamp) value is huge amount larger than the  req->timestamp we sent out, and when subtracted from `now` produces a negative long which is then displayed as a large unsigned value. 

Not sure why this is, could this be the particular router I'm using, or is there perhaps an issue with the reference units used in the echo request/reply timestamp?

djohanse:
The current master looks fine for me:

Screenshot 2026-07-31 at 11 52 36 PM

This is open codes initial thoughts.  Note that if something is just plain wrong we'll need to keep asking until he gets smarter and figures it out.  But lets start here:

Verdict: not a byte-order bug, and not a reference-units problem — it's the peer (almost certainly the affected user's router) not echoing the ICMP payload. I traced the full chain in current main and it's consistent.

Why the code is correct:
Send: ping.c:87 req->timestamp = htonl(tstart) → ktcp icmp.c:45 *payload = timestamp writes the swapped value, which on little-endian ia16 lands on the wire as tstart in network order. Reply: ktcp icmp.c:159 LE-reads the echoed 4 bytes back to bswap(tstart) (netconf.c:163), and ping.c:119 ntohl() restores tstart. rtt = now - tstart ≈ small. The structs match on both sides (12-byte request, 6-byte reply), and netconf_set_extra is genuinely wired in (tcpdev.c:408).
Units are fine too: the echo payload is opaque per RFC 792 — ms is fine because send and read use the same get_ms(), so units cancel on a verbatim echo.

Why it fails for that user: the displayed values drift ~1000 ms/sec — that's now advancing while ntohl(reply.timestamp) stays constant. A correctly-echoing peer makes reply.timestamp track tstart and rtt stays ~1ms. A constant reply.timestamp means the reply's 4 payload bytes are not your timestamp — the router/firewall answered the echo from its CPU without copying the data field (a common RFC 792 violation on NAT gateways and cheap SoC routers).


Confirm it with a differential test:
1. ping 127.0.0.1 — ktcp's own echo path (icmp.c:138-155) echoes verbatim. Normal times here → stack is fine, router is the culprit.
2. If loopback is also broken → the affected user is running an old ktcp image: before 207f776e ("Implement ping and ifconfig", in main since June 3), netconf_set_extra was never called, the payload was dropped, and the static icmp_req.timestamp was stale/zero — that produces the exact same symptom.
3. Capture to prove it: use ELKS's new tcpdump capture (netconf_capture_packet) or a PC-side tcpdump -i <if> icmp and compare the 4 payload bytes in request vs reply.
Two latent robustness notes (not the cause, but relevant): icmp.c:159 reads 4 payload bytes without checking len, so a truncated reply yields whatever follows in the buffer; and get_ms() (tv_sec*1000) wraps 32-bit in 2026 — benign for correct echoes, but if a peer substitutes a seconds-scale value, the wrapped subtraction produces exactly this garbage

djohanse:
Also FYI

is there a fix/workaround for this router issue?

+ Thought: 20.4s
Yes. Two levels:

Workaround (no code change): ping hosts that echo the payload properly. Loopback (127.0.0.1, answered by ktcp itself) always works, and most real hosts echo verbatim. The failure is specific to the router's own echo replies — pinging a host through the router usually works, because NAT/ICMP forwarding passes the data field through untouched while only the router's CPU-generated replies skip it. So the affected user should test ping 127.0.0.1 (confirms stack OK) and a host beyond the router.


Real fix (1-line code change): stop trusting the echoed payload for RTT. The client already records its own send time — tstart at ping.c:78 — so compute RTT locally instead of from the reply timestamp:
rtt = now - tstart;          // ping.c:119, instead of now - ntohl(reply.timestamp)

This is exactly how Linux/macOS ping times round trips (local send/receive clocks); it makes RTT immune to any peer that mangles or drops the payload. With single-flight pings (one outstanding request, pending_icmp_cb), reply.timestamp isn't even needed for correlation, so the payload field becomes irrelevant for timing.

Optional hardening while you're there: ktcp forwards any echo reply to the pending client without checking id/seq (icmp.c:156-165), so a stray or NAT-id-rewritten reply is currently counted as a success — harmless here but worth noting.

djohanse:
Note in a follow-up question open code decided that "optional hardening" would do more harm than good.  I no longer have the transcript unfortunately

ghaerr:
Thanks for the information. Pinging localhost works fine. I'm pretty sure the problem is either receiving a truncated reply (see below, we probably should check rather than returning garbage), or the router is inserting garbage into the timestamp field. I'd bet its probably the former, and likely due to my specific router.

> latent robustness notes (not the cause, but relevant): icmp.c:159 reads 4 payload bytes without checking len, so a truncated reply yields whatever follows in the buffer

This should probably be fixed in the ktcp ICMP code, but not sure exactly what should be done, except perhaps pre-filling the buffer with zeros before copying any reply.

> compute RTT locally instead of from the reply timestamp

> This is exactly how Linux/macOS ping times round trips (local send/receive clocks); it makes RTT immune to any peer that mangles or drops the payload.

Computing RTT locally is a better solution. I've made the suggested change in 5b8ce20a3255f941c60da54edd8f982f028688ed - this fixed the problem. 

> "optional hardening" would do more harm than good. 

Agreed, no need to have our ICMP get into sequence number checking for fixing `ping`.

Thanks for your help!

PR #2785, opened 2026-08-01
MFM driver: controller fault recovery, optional IRQ 5 completion, 8237 Extended Write option

parabyte:
Three improvements to the XT MFM hard disk driver, developed against a WD1002A-WX1 and tested functional on a real Amstrad PC1640 DD.

## Controller reset on phase fault

A handshake fault left the WD1002A-WX1 wedged with SELECT asserted (status `0xCB`, drive light stuck on): every later command failed, a reprobe reported "no XT MFM drives found" on a perfectly good disk, and a fault part way through a write left the sector half written. `mfmhd_phase_fault()` now pulses the reset port and waits for SELECT to drop. Observed live on the hardware: status went from `0xCB` straight back to `0xC0` and the retried command succeeded. The command still fails and the caller still retries as before — the reset only ensures the retry meets a sane controller.

## Optional interrupt-driven completion, IRQ 5 (`mfm=` bit 3)

The controller has always been able to assert IRQ on command completion (`MFM_CTL_IRQEN`, bit 1 of the mask register) but the driver only ever span on the status port, burning the CPU for the whole of every transfer. With `mfm=` bit 3 set, the task sleeps until the completion interrupt arrives on IRQ 5, the PC/XT fixed-disk line.

It is opt-in rather than default because this is the driver the system boots from: if a board does not drive the line, an unbootable machine would be the failure mode. Without the bit, behaviour is exactly the previous polled code, and the poll loop is retained underneath as a backstop even when the bit is set — a missing interrupt costs latency rather than a hang. The IRQ 5 selection was verified against the board's jumper.

## `dmaxw=1` bootopts option: 8237 Extended Write

The Amstrad PC1512/1640 clock their 8237 at 4 MHz for a five-clock 1.25 µs bus cycle on channels 1–3, and the ROS initialises the controller to Late Write (PC1640 Technical Reference, section 1.5) — a shorter write strobe than 8-bit peripherals were designed against, so this controller misses its DMA handshake and the driver ends up falling back to PIO. `dmaxw=1` sets the 8237's Extended Write command bit once at driver init, asserting the strobe a clock earlier so it is held longer. Off by default: machines with standard timing have no reason to alter a working controller, and the option is harmless where enabled unnecessarily.

## Testing

All three changes are running on real hardware: an Amstrad PC1640 DD booting ELKS from the WD1002A-WX1 as its root disk, with `dmaxw=1` in use and the filesystem fsck-clean after sustained load (including a full 20 MB image write over FTP). The build was also verified against current master with the MFM driver enabled.

### DIFF
--- a/elks/init/main.c
+++ b/elks/init/main.c
@@ -46,6 +46,7 @@ struct netif_parms netif_parms[MAX_ETHS] = {
     { ULTRA_IRQ, ULTRA_PORT, ULTRA_RAM, ULTRA_FLAGS },
     { LANCE_IRQ, LANCE_PORT, 0, LANCE_FLAGS },
 };
+int dma_extwrite_opt;           /* /bootopts dmaxw=1: 8237 Extended Write */
 int mfmhd_slow_profile;         /* /bootopts mfm= bit 0: slow controller timing */
 int mfmhd_pio;                  /* /bootopts mfm= bit 1: PIO sector transfers */
 int mfmhd_trace;                /* /bootopts mfm= bit 2: driver request tracing */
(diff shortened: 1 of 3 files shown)


elks/init/main.c
```diff
@@ -640,6 +641,16 @@ static int INITPROC parse_options(void)
...
+        /*
+         * dmaxw=1 switches the 8237 to Extended Write.  Off by default: only
+         * machines whose DMA timing is tighter than 8-bit cards expect need
+         * it, notably the Amstrad PC1512/1640.  Harmless elsewhere, but there
+         * is no reason to alter a working controller's timing.
+         */
+        if (!strncmp(line,"dmaxw=",6)) {
+            dma_extwrite_opt = (int)simple_strtol(line+6, 0);
+            continue;
+        }
```
ghaerr, line 653: Please remove this extra option from main.c/bootopts and instead use bit 4 of mdm_opts to indicate 8237 Extended Write. This is because this is only used by the MFMHD driver.

PR #2787, opened 2026-08-01
MFM driver: controller fault recovery, optional IRQ 5 completion, 8237 Extended Write option

parabyte:
Resubmission of #2785 — that PR was auto-closed when its source fork was deleted.

Three improvements to the XT MFM hard disk driver, developed against a WD1002A-WX1 and tested functional on a real Amstrad PC1640 DD.

## Controller reset on phase fault

A handshake fault left the WD1002A-WX1 wedged with SELECT asserted (status `0xCB`, drive light stuck on): every later command failed, a reprobe reported "no XT MFM drives found" on a perfectly good disk, and a fault part way through a write left the sector half written. `mfmhd_phase_fault()` now pulses the reset port and waits for SELECT to drop. Observed live on the hardware: status went from `0xCB` straight back to `0xC0` and the retried command succeeded. The command still fails and the caller still retries as before — the reset only ensures the retry meets a sane controller.

## Optional interrupt-driven completion, IRQ 5 (`mfm=` bit 3)

The controller has always been able to assert IRQ on command completion (`MFM_CTL_IRQEN`, bit 1 of the mask register) but the driver only ever span on the status port, burning the CPU for the whole of every transfer. With `mfm=` bit 3 set, the task sleeps until the completion interrupt arrives on IRQ 5, the PC/XT fixed-disk line.

It is opt-in rather than default because this is the driver the system boots from: if a board does not drive the line, an unbootable machine would be the failure mode. Without the bit, behaviour is exactly the previous polled code, and the poll loop is retained underneath as a backstop even when the bit is set — a missing interrupt costs latency rather than a hang. The IRQ 5 selection was verified against the board's jumper.

## `mfm=` bit 4: 8237 Extended Write

The Amstrad PC1512/1640 clock their 8237 at 4 MHz for a five-clock 1.25 µs bus cycle on channels 1–3, and the ROS initialises the controller to Late Write (PC1640 Technical Reference, section 1.5) — a shorter write strobe than 8-bit peripherals were designed against, so this controller misses its DMA handshake and the driver ends up falling back to PIO. `mfm=` bit 4 (mfm=16) sets the 8237's Extended Write command bit once at driver init, asserting the strobe a clock earlier so it is held longer. Off by default: machines with standard timing have no reason to alter a working controller, and the option is harmless where enabled unnecessarily.

Per review, this is selected with `mfm=` bit 4 rather than a separate `dmaxw=` /bootopts option, since only this driver uses it — main.c is no longer touched, and dma.h now carries only the `DMA1_CMD_EXTWRITE` register bit definition, with the Amstrad-specific timing discussion moved into the driver. The `mfm=` bits are documented in Documentation/text/mfmhd.txt.

## Testing

All three changes are running on real hardware: an Amstrad PC1640 DD booting ELKS from the WD1002A-WX1 as its root disk, with Extended Write in use and the filesystem fsck-clean after sustained load (including a full 20 MB image write over FTP). The build was also verified against current master with the MFM driver enabled.

### DIFF
--- a/elks/include/arch/dma.h
+++ b/elks/include/arch/dma.h
@@ -77,6 +77,21 @@
 
 /* DMA controller registers */
 #define DMA1_CMD_REG		0x08	/* command register (w) */
+/*
+ * Command register bit 5 picks the write-strobe timing: 0 is Late Write, 1 is
+ * Extended Write, which asserts the strobe a clock earlier and so holds it for
+ * longer.  It exists for peripherals that cannot meet the short pulse.
+ *
+ * The Amstrad PC1512/1640 need it.  Their 8237 is clocked at 4MHz and takes a
+ * five-clock 1.25us bus cycle on channels 1-3 (PC1640 Technical Reference
+ * 1.5), where a real XT clocks the part near 2.39MHz for a four-clock cycle of
+ * about 1.68us - so the Amstrad completes each transfer roughly 25% faster
+ * than 8-bit cards were designed against, and the ROS firmware additionally
+ * initialises the controller to Late Write (1.5.2), the shorter of the two.
+ * The result is dropped or mistimed bytes: distorted audio on the sound card,
+ * and a hard disk controller that misses its handshake altogether.
+ */
+#define DMA1_CMD_EXTWRITE	0x20	/* extended write for slow peripherals */
 #define DMA1_STAT_REG		0x08	/* status register (r) */
 #define DMA1_REQ_REG            0x09	/* request register (w) */
 #define DMA1_MASK_REG		0x0A	/* single-channel mask (w) */
--- a/elks/include/arch/dma.h
+++ b/elks/include/arch/dma.h
@@ -77,21 +77,7 @@
 
 /* DMA controller registers */
 #define DMA1_CMD_REG		0x08	/* command register (w) */
-/*
- * Command register bit 5 picks the write-strobe timing: 0 is Late Write, 1 is
- * Extended Write, which asserts the strobe a clock earlier and so holds it for
- * longer.  It exists for peripherals that cannot meet the short pulse.
- *
- * The Amstrad PC1512/1640 need it.  Their 8237 is clocked at 4MHz and takes a
- * five-clock 1.25us bus cycle on channels 1-3 (PC1640 Technical Reference
- * 1.5), where a real XT clocks the part near 2.39MHz for a four-clock cycle of
- * about 1.68us - so the Amstrad completes each transfer roughly 25% faster
- * than 8-bit cards were designed against, and the ROS firmware additionally
- * initialises the controller to Late Write (1.5.2), the shorter of the two.
- * The result is dropped or mistimed bytes: distorted audio on the sound card,
- * and a hard disk controller that misses its handshake altogether.
- */
-#define DMA1_CMD_EXTWRITE	0x20	/* extended write for slow peripherals */
+#define DMA1_CMD_EXTWRITE	0x20	/* cmd bit 5: extended write strobe */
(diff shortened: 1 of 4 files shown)


elks/init/main.c
```diff
@@ -640,6 +641,16 @@ static int INITPROC parse_options(void)
...
+        /*
+         * dmaxw=1 switches the 8237 to Extended Write.  Off by default: only
+         * machines whose DMA timing is tighter than 8-bit cards expect need
+         * it, notably the Amstrad PC1512/1640.  Harmless elsewhere, but there
+         * is no reason to alter a working controller's timing.
+         */
+        if (!strncmp(line,"dmaxw=",6)) {
+            dma_extwrite_opt = (int)simple_strtol(line+6, 0);
+            continue;
+        }
```
ghaerr: Please remove this extra option from main.c/bootopts and instead use bit 4 of mdm_opts to indicate 8237 Extended Write. This is because this is only used by the MFMHD driver.

elks/include/arch/dma.h
```diff
@@ -77,6 +77,21 @@
...
+ * The Amstrad PC1512/1640 need it.  Their 8237 is clocked at 4MHz and takes a
+ * five-clock 1.25us bus cycle on channels 1-3 (PC1640 Technical Reference
+ * 1.5), where a real XT clocks the part near 2.39MHz for a four-clock cycle of
+ * about 1.68us - so the Amstrad completes each transfer roughly 25% faster
+ * than 8-bit cards were designed against, and the ROS firmware additionally
+ * initialises the controller to Late Write (1.5.2), the shorter of the two.
+ * The result is dropped or mistimed bytes: distorted audio on the sound card,
+ * and a hard disk controller that misses its handshake altogether.
+ */
+#define DMA1_CMD_EXTWRITE	0x20	/* extended write for slow peripherals */
```
ghaerr: Move the above system-specific long comment into the driver, where it is better seen and explains operation there.

elks/init/main.c
parabyte: Done in 88d3da9d — dmaxw= and its main.c plumbing are removed; Extended Write is now selected with mfm= bit 4 (mfm=16), parsed in the driver alongside the other bits. main.c is no longer touched by this PR.

elks/include/arch/dma.h
parabyte: Done in 88d3da9d — the timing discussion now sits in the driver next to the new mfm= bit 4 flag; dma.h keeps only the DMA1_CMD_EXTWRITE register bit definition. Also documented mfm= bits 2-4 in Documentation/text/mfmhd.txt.

ghaerr:
Thanks. Change from draft to final PR and I'll commit.

ghaerr:
@parabyte Hopefully you can see what I'm looking for in these large drivers, etc. By keeping all your special operations within the driver file(s) and out of mainline, I don't have to worry about kernel maintenance and integrity when your driver is not compiled in. Likewise for source files, by keeping all the extra options out of mainline, the kernel isn't bogged down with stuff that only affects the driver. As I've indicated in your DSP PR, sometimes we need to actually enhance the kernel by adding a new structure (e.g. isa_conf), which then removes all the extra ifdef crap from the kernel code. I don't mind adding drivers, but definitely remain concerned about kernel simplicity and correctness. During the whole process, it is important to observe sensible naming across config options, driver names, and variables so as to keep ELKS as desirable as it has become - IMO this is because of all the work that's been put in to make it understandable.

The same applies to configuration options in .config files: let's not add tons of new options until the driver involved actually gets used by (many) others. Otherwise, the configuration itself becomes quite burdensome to the average user, with tons of options that are rarely used. I deleted tons of options over the last five years and I like to think the config files are easier to understand now too.

PR #2791, opened 2026-08-04
MFM driver: take port, irq and flags from mfm= in /bootopts

parabyte:
Now that struct isa_conf is in, mfm= takes the same irq,port,ram,flags form as the NIC and audio options, parsed through parse_isaopts.

The port, irq and flag bits come from /bootopts with the defaults in ports.h as MFM_PORT, MFM_IRQ and MFM_FLAGS. CONFIG_MFMHD_PORT is gone from block/config.in, so moving the controller between 0x320 and 0x324 no longer needs a kernel rebuild. CONFIG_MFMHD_SEAGATE stays, since it selects command profiles with #if rather than a value.

The mfm= bits are unchanged in meaning, just named: MFMF_SLOW, MFMF_PIO, MFMF_TRACE, MFMF_IRQ and MFMF_EXTWRITE in the driver. Nothing was removed. PIO, the automatic fall back to PIO after a failed DMA transfer, tracing, interrupt completion and extended write all behave as before, as do the source tunables at the top of the file. mfm=off skips the driver the same way sb=off and ne0=off do.

Three stale globals in main.c are also gone: mfmhd_slow_profile, mfmhd_pio and mfmhd_trace were declared there as well as in the driver, and only the driver used them. Those and the two others are static now.

Documented in Documentation/text/mfmhd.txt and Configure.help.

Builds and runs on the Amstrad PC1640 with a WD1002A-WX1 at 0x320.

Head parking is a separate PR to follow.

### DIFF
--- a/Documentation/text/mfmhd.txt
+++ b/Documentation/text/mfmhd.txt
@@ -13,22 +13,29 @@ Hardware resources
 
 One controller with up to two drives is supported.  The controller normally
 occupies I/O ports 0x320 through 0x323; jumper W4 can select 0x324 instead
-(set CONFIG_MFMHD_PORT to match).
+(set mfm= in /bootopts to match).
 
 Command, status, sense, and controller-parameter bytes use programmed I/O.
 Sector payloads use 8237 DMA channel 3, the standard IBM XT hard-disk DMA
 channel.  The driver reserves a fixed 1024-byte buffer below the first 64 KiB
 physical boundary, so an 8237 transfer cannot wrap.  Normal transfers are
 therefore limited to two 512-byte sectors per command.  The controller IRQ is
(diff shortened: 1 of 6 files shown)


parabyte:
I did remove pio support, but i did a little google and forum read, and it turns out some quebec controllers have faulty interrupt handling inside the devices option rom. so i decided to leave that in.

ghaerr:
@parabyte,

Very nice cleanup!!

> A latent bug went with it. crtc_base was read from the BIOS data area before the mode set, so it described the console being displaced rather than the mode about to run, and a mono mode then wrote 6845 registers at 0x3D4

Nice find, but I don't see it? Are you fixing that in another PR?

Thank you

PR #2792, opened 2026-08-04
MFM driver: park heads on shutdown

parabyte:
Parks the MFM drives on shutdown, which the driver did not do.

sys_reboot seeks each drive to its last cylinder, the landing zone on an ST-506, so an unpowered head settles off the data area. The call sits next to the existing bios_disk_park_all under its own ifdef.

It goes through the BIOS seek rather than the controller's own SEEK command block. That is smaller, and on any board whose option ROM is enabled it is the same seek the BIOS driver already performs. Worth saying plainly though: it depends on that ROM hooking INT 13h. A card strapped with its ROM off is exactly the case CONFIG_BLK_DEV_MFMHD exists for, and there BD_DX = 0x80 + drive addresses whatever the system BIOS thinks drive 80h is. If you would rather it used MFMHD_CMD_SEEK, which needs no ROM, say so and I will swap it, it is about ten lines more.

Two notes on reusing the BIOS interface from a driver: bdt is bios.c's own static and the BD_ macros name it by identifier, so this declares its own struct biosparms, and zeroes it because a local starts as stack garbage where the static did not.

Tested on an Amstrad PC1640 with a WD1002A-WX1. Rebooted through the new path, the machine came back, the filesystem mounted clean and the disk reads normally. What that does not show is the heads actually moving, which is not observable from software.

Touches mfmhd.c, as does #2791. They edit different parts of the file so they should not conflict, but whichever lands second may want a look.

### DIFF
--- a/elks/kernel/sys.c
+++ b/elks/kernel/sys.c
@@ -19,6 +19,7 @@
 #include <arch/segment.h>
 #include <arch/system.h>
 #include <arch/io.h>
+#include <arch/mfmhd.h>
 
 /*
  * reboot doesn't sync: do that yourself before calling this.
@@ -30,6 +31,9 @@ int sys_reboot(unsigned int magic, unsigned int magic_too, int flag)
 
 #ifdef CONFIG_BLK_DEV_BHD
     bios_disk_park_all();
+#endif
+#ifdef CONFIG_BLK_DEV_MFMHD
+    mfmhd_park_all();
 #endif
     switch(flag) {
     case RB_REBOOT:
--- a/elks/arch/i86/drivers/block/mfmhd.c
+++ b/elks/arch/i86/drivers/block/mfmhd.c
@@ -61,8 +61,7 @@
 #include <arch/segment.h>
 #include <arch/system.h>
 #include <arch/divmod.h>
-
-#include "mfmhd.h"
+#include <arch/mfmhd.h>
 
 #define MAJOR_NR        MFMHD_MAJOR
 #include "blk.h"
@@ -75,7 +74,7 @@
 #define MFM_HD_JUMPER           2
 #define MFM_HD_CONTROL          3
 
-/* Hardware status register bits, mirrored in mfmhd.h for user tools. */
+/* Hardware status register bits, mirrored in arch/mfmhd.h for user tools. */
 #define MFM_STAT_READY          MFMHD_STAT_READY
 #define MFM_STAT_INPUT          MFMHD_STAT_INPUT
 #define MFM_STAT_COMMAND        MFMHD_STAT_COMMAND
@@ -108,6 +107,7 @@
 #define MFM_CMD_TEST_DRIVE_READY        MFMHD_CMD_TEST_DRIVE_READY
 #define MFM_CMD_RECALIBRATE             MFMHD_CMD_RECALIBRATE
 #define MFM_CMD_SENSE                   MFMHD_CMD_READ_STATUS
+#define MFM_CMD_SEEK                    MFMHD_CMD_SEEK
 #define MFM_CMD_READ_SECTORS            MFMHD_CMD_READ_SECTORS
 #define MFM_CMD_WRITE_SECTORS           MFMHD_CMD_WRITE_SECTORS
 #define MFM_CMD_INIT_DRIVE_PARAMETERS   MFMHD_CMD_INIT_DRIVE_PARAMETERS
@@ -1116,6 +1116,7 @@ mfmhd_command_wait_ticks(unsigned char op)
 {
     switch (op) {
     case MFM_CMD_RECALIBRATE:
+    case MFM_CMD_SEEK:              /* a park is a full stroke, 3ms a step */
     case MFM_CMD_READ_SECTORS:
     case MFM_CMD_WRITE_SECTORS:
     case MFM_CMD_INIT_DRIVE_PARAMETERS:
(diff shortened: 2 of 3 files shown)


parabyte:
i am unsure, as i often use my mfm drive as a secondary drive with the mfm drives option rom disabled, 
and a bios derived drive as my primary when i am doing data archiving.

ghaerr:
@parabyte,

> If you would rather it used MFMHD_CMD_SEEK, which needs no ROM, say so 

Given the trouble you've gone to to implement the MFMHD driver completely without BIOS calls, I say it's a shame to have to pull a BIOS call just to park the heads. Frankly, the MFM driver is already huge, so I don't see much issue with making it a bit larger to command the heads to park directly. Also, getting away from the BIOS has other benefits - in our experimental protected mode kernel, no BIOS calls are allowed, so this BIOS call would invalidate using this driver in the PM kernel, since the kernel would panic on invalid interrupt just as the BIOS call were attempted.

So I would suggest not using BIOS.

I have other ideas about how to decrease the size of the MFM driver, but it doesn't really matter that much, I think.

Speaking of when to park the heads, I'm not really happy about the kluge I wrote having a special call to park all HD heads in the shutdown routine - perhaps it would be better to park the heads in the driver release (close) routine. What do you think? I am pretty sure that the root block device is released before shutdown, but there's a chance it isn't. So we don't need to worry about that now. I'm also unsure whether parking the heads between close and a subsequent open would be a good idea or not anyways - whether parking the heads is hard on the drive if done frequently.

There's another problem with the mfmhd_park_all() declaration, I'll comment on that by the source code.

elks/include/linuxmt/genhd.h
```diff
@@ -103,6 +103,7 @@ extern unsigned char bios_drive_map[];  /* map drive to BIOS drivenum */
 extern int boot_partition;              /* MBR boot partition, if any */
 
 void GENPROC init_partitions(struct gendisk *dev);
+void mfmhd_park_all(void);
```
ghaerr: Unfortunately, we don't want a particular driver declaration in this header file. It belongs in mfmhd.h. This means that you're probably going to have to move mfmhd.h to elks/include/arch and make the other required fixups to .c files that include it.

You can then include arch/mfmhd.h in kernel/sys.c.

parabyte:
Swapped to MFMHD_CMD_SEEK, no BIOS. It costs 20 bytes of text in mfmhd.o and the kernel image comes out the same size, so the ROM independence is close to free. Your PM kernel point settles it anyway.

Going through the controller turned up a bug in what I first posted. Park now walks controller units rather than BIOS drive numbers, and a FileCard 20 is one drive spread over two platters with an actuator each, so the old single BIOS seek parked one head and left the other loaded.

One thing not in the original diff. MFM_CMD_SEEK was falling into the default two second phase timeout. A park is a full stroke, around 1000 cylinders at 3ms a step, so it would have timed out and dragged a resync in behind it. Seek now shares recalibrate's disk timeout.

mfmhd.h moved to include/arch/mfmhd.h with the guard renamed, mfmhd.c includes it as <arch/mfmhd.h>, and the declaration lives there under __KERNEL__ rather than in genhd.h. Nothing else in the tree included the old path, so there were no other fixups. genhd.h is untouched now.

On parking in release rather than at shutdown, I would rather leave it where it is. A full stroke on every close means one each time the root is remounted or a tool closes the device, and that is wear for nothing.

Builds clean. I have not rerun it on the PC1640 yet, will do that and report back.

elks/kernel/sys.c
```diff
@@ -19,6 +19,9 @@
 #include <arch/segment.h>
 #include <arch/system.h>
 #include <arch/io.h>
+#ifdef CONFIG_BLK_DEV_MFMHD
```
ghaerr: Remove this ifdef; not necessary and obscures readability.

parabyte:
Ifdef around the include dropped, it is unconditional now. The declaration costs nothing when the driver is not built and only the call needs guarding.

Built both ways to be sure: with CONFIG_BLK_DEV_MFMHD on, and with it off after regenerating autoconf.h. Both link clean, and the image drops from 129744 to 117120 with the driver out, so the header really is inert when it is not wanted.

genhd.h stays untouched and mfmhd.h is in include/arch as before.

ghaerr:
Thanks!

ghaerr:
Hello @parabyte,

In the above committed code, 'MFMHD_PORT' is used, but not defined anywhere in the source base. This means that the driver could not have actually been compiled and tested with the existing source tree!

I just happened to find this, as I was thinking about my previous statement about bit-rot: that each of the driver files gets compiled, so there is not need to ifdef the entire source file for the option. It turns out that is incorrect, since both drivers/block/Makefile and drivers/char/Makefile don't actually compile a given driver unless their config option is set in .config. During the process of changing the Makefile to compile every driver, just to see what would happen, I found this.

I will fix this in my upcoming PR.

In the meantime, I can't really figure out how this code could have compiled on your machine? Previously, the MFMHD port value was set in .config, but it was set via CONFIG_MFMHD_PORT, not MFMHD_PORT.

All the more reason for me to somewhat radically change the build mechanism for ELKS drivers: by removing the ifdefs in the driver Makefiles, we should be able to compile all possible drivers for each platform, then let the kernel link process automatically pull in any required drivers through undefined symbols in the kernel and the driver.a libraries.

PR #2793, opened 2026-08-04
Do not artificially restrict ftp active mode if -n ip address specified

djohanse:
Do not artificially restrict ftp active mode if -n ip address is specified


ghaerr:
Thanks @djohanse. Do you want this merged now, or wait for @duzenko's testing?

djohanse:
I think we want this change regardless.  So I'll say merge now

PR #2798, opened 2026-08-05
[kernel] Use himem.sys method to enable A20 for greater 8042 compatibility

ghaerr:
After research and discussion in #2790, this changes the A20 enable mechanism to use the method employed by DOS himem.sys, which should allow for more compatibility with the various implementations of the 8042 keyboard controller.

As also discussed in #2790, the BIOS Data Area location 40:96 bit 4 is checked to see whether an 101/102 enhanced keyboard is installed, and enables CAP_KBD_LEDS so that the LEDS are processed during keypresses.

@EC1842, can you please test this to verify it works on your hardware?


EC1842:
> can you please test this to verify it works on your hardware?

I'm away from home right now, but I can check it in the next three days.

Issue #2781, opened 2026-07-30
kTCP wedged by RST


duzenko:
**Description**
- What is expected ? kTCP handles basic connectivity reliably
- What is observed ? heap leaks and hangups on telnet + FTP

**Configuration**
- Version (commit or release): master build
- Emulator or HW platform: QEMU ne2k kvm

**How to reproduce ?**
- Does the problem always happen ? even the minimal case - usually (leak/hang on first killed ftp upload)
- List the steps to reproduce the problem

As easy as 
```
timeout 1 bash -c "yes | curl -s --max-time 5 -T - 'ftp://$IP/k.txt' -u root:" >/dev/null 2>&1
```
The killed FTP upload triggers an RST which then leads to kTCP panic and then wedge. Note that on FDD images the timeout must be tuned down so that ftp upload writes ~50KB and does not fill the floppy and causes another bug.

**Raw data**
- Screenshot(s):
<img width="720" height="400" alt="Image" src="https://github.com/user-attachments/assets/471da111-9be5-4165-9f7c-cb3fa6c4cb1d" />

- Log file:

[issue-stock-serial.log](https://github.com/user-attachments/files/30536070/issue-stock-serial.log)

**Additional information**
The vibe-coded fix is at https://github.com/duzenko/elks/commit/38bd6f46f7ba3d6e22437328c769c3d329e88c38.
To review/understand it would strictly require a person with the kernel/kTCP communication knowledge, who I'm not regretfully.
It probably does something wrong/stupid, but still fixes the symptom and thus demonstrates the problem.
Note: I can still reproduce another wedge on a PM build in VBox with faster PCnet card, but QEMU is pretty stable with the fix.

ghaerr:
Hello @duzenko,

Geez, sorry, I completely missed seeing this. I'll take a look and report back, thanks!

ghaerr:
I'm trying to reproduce this on QEMU: I'm running macOS and it doesn't have `timeout`, so I'm trying to think of a way to force a RST to be transmitted. Any other ideas?

> The vibe-coded fix is at https://github.com/duzenko/elks/commit/38bd6f46f7ba3d6e22437328c769c3d329e88c38.

I've reviewed the fix, and I think it looks pretty good, but want to play with it a bit when I can get a RST to reproduce the bug.

> ftp upload writes ~50KB and does not fill the floppy and causes another bug.

Does `ftpd` not return an error when it runs out of space? I can look into that also.

[...]

PR #2801, opened 2026-08-06
[net] ktcp rework: memory, throughput, UDP sockets

parabyte:
Rework of ktcp and the kernel socket shim. I've spent the last 20 hours or so, on and off, in ktcp.

## Why

Four things bound the stack, in the order they bite:

1. **~7 connections machine-wide.** `tcpcb_new()` allocated 86 + 4380 bytes per connection out of a 33792 byte heap, shared by telnetd, ftpd, httpd and everything else.
2. **~13 KB of ktcp's 16.8 KB BSS was unusable.** `ipbuf`, `tcpbuf` and tcpdev's `sbuf` were all sized off `CB_NORMAL_BUFSIZ` — the per-connection *ring* size — rather than off the MTU. There's no IP reassembly and the MTU is capped at 1500, so most of that was dead weight sitting where heap should be.
3. **A deadlock that wedged the whole machine.** `tcpdev_read()` on a socket whose control block had gone printed a message and returned without replying. The caller stayed asleep holding the global `rwlock` in `af_inet.c`, so every socket on the machine stopped. This is what had been hanging my box: all services would accept connections and then transfer nothing. It reproduces reliably by killing a server while a client is attached.
4. **UDP was unreachable from userspace.** `udp.c` existed but said "not accessible through sockets"; `sys_socket()` rejected `SOCK_DGRAM`, and there were no `sendto`/`recvfrom` syscalls.

## What changed

**Buffer sizing.** `TCPDEV_BUFSIZ`, `IP_BUFSIZ` and `CB_NORMAL_BUFSIZ` are now three independent definitions sized from what each actually holds, instead of all deriving from the receive window. The wire buffers come from `MAX_PACKET_ETH`; the ring stays a per-connection tuning dial. That frees enough heap to raise the connection ceiling from about 7 to 14 measured, and to 16 with a smaller `SO_RCVBUF`, at which point the fd limit binds rather than memory.

**`SO_RCVBUF` is honoured on accepted sockets.** `tcp_listen()` used to hardcode the ring size, so a bulk receiver and an idle game client cost the same. `SO_ACCEPT_BUFSIZ_TINY/SMALL/BULK` are suggested values; the semantics are that `SO_RCVBUF` on a socket that goes on to `listen()` sets the ring of the connections it *accepts*, since the listening socket never receives anything itself.

**Ring copies use `memcpy`.** Both were byte-at-a-time loops with a wrap test per byte; they're now at most two `memcpy` calls split at the wrap point. These run on every received segment and every application read.

**The deadlock fix.** A read on a socket with no control block now answers with `-EPIPE` instead of returning silently. Connect on a refused socket likewise notifies rather than leaving the caller asleep.

**Embryonic connection reaping.** Unaccepted control blocks are now timed out, so an unattended port can't exhaust the heap.

**UDP sockets.** `sendto`/`recvfrom` syscalls, `TDC_SENDTO`/`TDC_RECVFROM` tcpdev opcodes carrying the peer address, and `udp.c` promoted from a DHCP callback table to a real demultiplexer with BSS most-specific-wins matching, per-socket receive queues, length validation and ICMP error delivery. DHCP still registers the same way and is unaffected.

The syscalls take five arguments, not six — ELKS syscalls carry no more than five — so `addrlen` is dropped at the kernel boundary and fixed at `sizeof(struct sockaddr_in)`. libc restores the POSIX six-argument shape. There's precedent for this in `getsocknam`.

## Measured on the hardware

Real Amstrad PC1640, 8086:

| | |
|---|---|
| TCP | **20463 B/s** |
| UDP | lossless to **16000 B/s**, saturating around 20500 B/s |
| Concurrent connections | **7 → 14** (16 with a smaller `SO_RCVBUF`, then fd-limited) |

Same figures on QEMU and on 86Box configured as a PC1640, using the same instrument.

**UDP works, but TCP currently outperforms it.** I haven't chased why yet. I suspect the per-datagram path through tcpdev has more fixed cost than the TCP path amortises over a segment, but that's a guess and I'd rather say so than dress it up.

The clearest practical result: with the old stack, audio receive topped out at **8000 Hz**. With this one I can max the driver out at **20000 Hz** 8-bit PCM and send it continuously with zero issues.

Also, my game server daemon used to hang the box after a while. Since the `-EPIPE` fix I haven't reproduced it.

## Notes

`netstat` gains UDP packet and drop counters.

I've deliberately kept this to the stack itself and the changes needed to build it. The applications that would benefit from the new `SO_ACCEPT_BUFSIZ_*` values aren't touched here.

There's also a `libc/asm/memcpy-s.S` change I've left out of this PR — it copies words instead of bytes, which is about 26% cheaper on an 8088 and speeds the ring copies above. It's a global libc change rather than networking, so it seemed better as its own PR.

### DIFF
--- a/elks/arch/i86/kernel/syscall.dat
+++ b/elks/arch/i86/kernel/syscall.dat
@@ -106,6 +106,12 @@ setsockopt	+80	5	= CONFIG_SOCKET
 getsocknam	+81	4	= CONFIG_SOCKET
 fmemalloc	+82	2	*
 fmemfree	+83	1	*
+# datagram send/recv. POSIX takes 6 args; ELKS syscalls take at most 5
+# (bx,cx,dx,di,si - libc/system/syscall0.inc), so addrlen is dropped and
+# fixed at sizeof(struct sockaddr_in). libc restores the POSIX shape,
+# exactly as getsocknam does for getsockname/getpeername.
+_sendto	+84	5	= CONFIG_SOCKET
+_recvfrom	+85	5	= CONFIG_SOCKET
 #
 # Name			No	Args	Flag&comment
 #
--- a/elks/include/linuxmt/errno.h
+++ b/elks/include/linuxmt/errno.h
@@ -75,6 +75,13 @@
 #define ESERVERERR      129     /* Server error */
 #define ERESTARTSYS     512     /* Restart system call*/
 
+/* datagram sockets need these; moved out of the UNUSED block below when
+ * sendto/recvfrom were added. Remember etc/perror needs matching strings. */
+#define EDESTADDRREQ    89      /* Destination address required */
+#define EMSGSIZE        90      /* Message too long */
+#define EPROTOTYPE      91      /* Protocol wrong type for socket */
+#define EOPNOTSUPP      95      /* Operation not supported on transport endpoint */
+
 #if UNUSED
 /* when any of these added in, add string to etc/perror as well */
 #define EWOULDBLOCK     41      /* Operation would block */
@@ -124,13 +131,9 @@
 #define ESTRPIPE        86      /* Streams pipe error */
 #define EUSERS          87      /* Too many users */
 
-#define EDESTADDRREQ    89      /* Destination address required */
-#define EMSGSIZE        90      /* Message too long */
-#define EPROTOTYPE      91      /* Protocol wrong type for socket */
 #define ENOPROTOOPT     92      /* Protocol not available */
 #define EPROTONOSUPPORT 93      /* Protocol not supported */
 #define ESOCKTNOSUPPORT 94      /* Socket type not supported */
-#define EOPNOTSUPP      95      /* Operation not supported on transport endpoint */
 #define EPFNOSUPPORT    96      /* Protocol family not supported */
 #define EAFNOSUPPORT    97      /* Address family not supported by protocol */
 
--- a/elks/include/linuxmt/net.h
+++ b/elks/include/linuxmt/net.h
@@ -74,6 +74,7 @@ struct proto_ops {
 #define SF_RST_ON_CLOSE	(1 << 4) /* inet */
 #define SF_REUSE_ADDR	(1 << 5) /* inet */
 #define SF_CONNECT	(1 << 6) /* inet */
+#define SF_DGRAM	(1 << 7) /* inet: SOCK_DGRAM socket (bit 7 was free) */
 
 struct net_proto {
     const char *name;		/* Protocol name */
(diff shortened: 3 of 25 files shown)


parabyte:
sorry, same thing here as 2802, i ran the code through chatgpt to write the comments because i was being lazy and it padded them out with essays. ive been through the lot and cut them right down, about 240 lines of comment gone, no code touched. ill write my own comments from now on

ghaerr:
Hello @parabyte,

Thank you for preparing this (huge) proposal. There is a lot to be looked at, understood and learned here, and it looks like you've identified several, if not many, important bug fixes as well as achieved consider throughput increase, even while admitting you're not sure why. However, there is no way I can accept a single PR with as many changes this makes to core kernel routines and the user land TCP/UP stack, especially when written by AI tool(s). There are simply too many changes at once to ensure correctness, and at the end of the day, I am responsible to ensure ELKS continues to work.

> ive been through the lot and cut them right down, about 240 lines of comment gone, no code touched. 

I wish you wouldn't do that - this PR is obviously written by AI. I was right in the middle of typing up comments about the PR when it changed on me. When an AI tool *initially* writes code, it is convenient to have it explained in detail. This particular PR has way too much to swallow at once, so having comments to understand what is what helps for an initial review.

I will write the rest of my review later, when the dust settles. The short version is that each bug fix will have to be a separate PR and separately tested with regression tests, to ensure correctness outside what is being fixed. The "wedged" fix, also written by AI, is proposed in #2781, for which I have reviewed and am expecting a PR. Your PR shows the hang issue may be broader than just when an RST is received. All this needs considerable attention by humans, AI tools are welcome, but regression tests will have to be made. I have to be able to reproduce bug fix scenarios to see the fix in action, as well as consider outside effects.

After the (many more?) bugs have been fixed in ELKS TCP, we can consider UDP. Due to the fact we're already out of near kernel space, and not all kernels are built FARPROC, any such enhancement will have to be a config option.

I will continue to study the bugs identified by your work, so that we can work on a path forward to correct them. I don't use AI, instead work the old-fashioned way and have to take time to understand any fixes or enhancements within the larger contexts of ELKS, so this will take some time, and will need discussions by others involved in ELKS networking.

Thank you!

parabyte:
understood

ghaerr:
I have noted 14 potential networking bug fixes in this PR, along with an additional 8 issues for discussion. That's in addition to the UDP implementation, which is now being discussed #2807. 

I plan on detailing these bugs/issues here, then organizing their implementation via separate PRs, along with regression testing and finding ways to reproduce each issue. These should probably be prioritized according to severity, fixing more serious problems first. This is all taking some time, as I have to fully understand each problem and its proposed resolution.

Hopefully this will keep the risks of introducing more problems to our networking stack to a minimum while improving it.

PR #2804, opened 2026-08-08
audio: move the adpcm expander to its own file shared by play and audiorcv

parabyte:
went with your suggestion, no define no makefile switch, the expander just lives in adpcm.c and adpcm.h and both play and audiorcv link it, always on. play gains -4 the same as audiorcv. builds clean both programs.

### DIFF
--- a/elkscmd/audio/Makefile
+++ b/elkscmd/audio/Makefile
@@ -11,13 +11,13 @@ all: $(PRGS)
 install: $(PRGS)
 	$(INSTALL) $(PRGS) $(DESTDIR)/bin
 
-play: play.o $(TINYPRINTF)
+play: play.o adpcm.o $(TINYPRINTF)
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
 audiomix: audiomix.o $(TINYPRINTF)
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
-audiorcv: audiorcv.o $(TINYPRINTF)
+audiorcv: audiorcv.o adpcm.o $(TINYPRINTF)
 	$(LD) $(LDFLAGS) -maout-heap=256 -maout-stack=1024 -o $@ $^ $(LDLIBS)
 
 clean:
--- a/elkscmd/audio/adpcm.c
+++ b/elkscmd/audio/adpcm.c
@@ -1,0 +1,74 @@
+/* creative 4 bit adpcm unpacker, shared by play and audiorcv */
+
+#include "adpcm.h"
+
+#define ADPCM_STATES    4
+
+/* all the sums done ahead of time so the 8086 just looks up and adds */
+
+static const signed char adpcm_delta[ADPCM_STATES * 16] = {
+    0, 1, 2, 3, 4, 5, 6, 7, 0, -1, -2, -3, -4, -5, -6, -7,
+    1, 3, 5, 7, 9, 11, 13, 15, -1, -3, -5, -7, -9, -11, -13, -15,
+    2, 6, 10, 14, 18, 22, 26, 30, -2, -6, -10, -14, -18, -22, -26, -30,
+    4, 12, 20, 28, 36, 44, 52, 60, -4, -12, -20, -28, -36, -44, -52, -60
+};
+
+static const unsigned char adpcm_next[ADPCM_STATES * 16] = {
+    0, 0, 0, 0, 0, 1, 1, 1, 0, 0, 0, 0, 0, 1, 1, 1,
+    0, 1, 1, 1, 1, 2, 2, 2, 0, 1, 1, 1, 1, 2, 2, 2,
+    1, 2, 2, 2, 2, 3, 3, 3, 1, 2, 2, 2, 2, 3, 3, 3,
+    2, 3, 3, 3, 3, 3, 3, 3, 2, 3, 3, 3, 3, 3, 3, 3
+};
+
+static int adpcm_ref = 128;
+static int adpcm_state;
+static int adpcm_primed;
+
+void adpcm_reset(void)
+{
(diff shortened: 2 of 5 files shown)


ghaerr:
Hello @parabyte,

Why are we removing support for ADPCM? IMO, it should probably be kept since `audiorcv` decodes ADPCM into PCM format beforehand, which is suitable for /dev/dsp (or possibly our internal speaker PCM support). If you're trying to decrease the size of `audiorcv`, which may be a good idea, perhaps the ADPCM support should be configurable via an `#define ADPCM 0` (default OFF) in audiorcv.c?

> hiya ghaeer, 

Actually ghaerr, thanks.

parabyte:
have you tried the adpcm?

it is inspired from a arduino project.

was my reference.

its vaguely useful on something like a 8086 clocked at 7 or so mhz like my amstrad.

not so on a original ibm, only verified via emulator not on real hardware.

so yes it absolutely needs to be gated behind a #define

The adpcm does have some interesting use cases, on higher end xt's and 286's.

ghaerr:
> have you tried the adpcm?

No, I am unable to emulate any sound blaster audio on macOS. Do you know of an emulator that can do this?

> so yes it absolutely needs to be gated behind a #define

> The adpcm does have some interesting use cases, on higher end xt's and 286's.

It seems there is usefulness in ADPCM then. So I can't quite tell whether this should be gated with a define in audiorcv.c, or just left alone.

If you think it should be gated, then add a \#define ADPCM or USE_ADPCM defaulted to on or off to this PR, and I'll commit it. Since `audiorcv` is a resident daemon, it may be better smaller, unless you think DosBox or 86Box will be used to generate ADPCM data, in which case perhaps it should always be left in, and this PR closed. For those users worried about resident daemon sizes, they can always just edit /etc/net.cfg, as they already have to in order to start audiorcv during net start anyways.

parabyte:
86box is your answer, but recommend the github source version

86box does not emulate the opti so you will only need standard sb

parabyte:
my whole point of adding in adpcm was cause of the networking throughput issue, that you have now resolved.

When i get to my workstation later, i shall look at this,

ghaerr:
@parabyte,

This is better, but frankly, it contains a ridiculous number of ifdef ADPCMs. In general, I don't recommend using ifdef for include files (e.g. \#include "adpcm.h"), nor for usage() lines. Having ifdefs impairs the readability of the source.

For most ELKS programs, rather than putting explaining all the command options in the usage() function, writing a man page is preferred. This keeps the size of the programs and disk images smaller. If wondering about a command from the desktop, `eman <program>` can be used automatically, for when not running HD images w/man pages.

Additionally, the vast majority of the remaining ifdefs could be deleted if you left in the global adpcm in, the '4' parsing in  getopts() etc, and then just displayed and error message when either admpc_reset or admpc_expand were called. That also allows getting rid of funky Makefiles that add or subtract .o files based on an option likely no one will ever play with anyways.

All that said, you don't need to do everything above, but I wanted to let you know what I think about massive ifdefs coming into ELKS - not preferred.

parabyte:
let me rethink this!

ghaerr:
> let me rethink this!

One idea might be just to remove the ADPCM option(s) entirely and all their associated stuff - I like the idea of keeping a separate decoder for adpcm to be shared between play and audiorcv, but just leave it ON at all times. In this way, all the mess disappears, and we don't have to maintain duplicated code between play and audiorcv.c.

The idea here would be get it working and designed properly first, then optimize later. I think you now have the right design by sharing a function in a separate file, but no need to optimize this into ADPCM Makefile and .c options, because frankly very very few people will even use it, given its dependency on the /dev/dsp hardware SB driver anyways.

My $0.02.

parabyte:
iv never written a man page, so need to read up on this! "excuse the bad pun"

otherwise i believe i have addressed the issues

parabyte:
> > let me rethink this!
> 
> One idea might be just to remove the ADPCM option(s) entirely and all their associated stuff - I like the idea of keeping a separate decoder for adpcm to be shared between play and audiorcv, but just leave it ON at all times. In this way, all the mess disappears, and we don't have to maintain duplicated code between play and audiorcv.c.
> 
> The idea here would be get it working and designed properly first, then optimize later. I think you now have the right design by sharing a function in a separate file, but no need to optimize this into ADPCM Makefile and .c options, because frankly very very few people will even use it, given its dependency on the /dev/dsp hardware SB driver anyways.
> 
> My $0.02.

I am hopeful other people will port other drivers over.

But also i am under the impression most 90's era 16bit cards can do sb emulation mode. some like my opti need to be told to go into this mode.

ghaerr:
This looks fine, except for one small issue in the Makefile, I will point that out by the source. Thank you.

> iv never written a man page, so need to read up on this!

Actually, creating a man page using AI would be the way to go, it could read other man pages to understand how, and then the source code to learn what the options are...

That's ok, though, man pages are optional.

parabyte:
> This looks fine, except for one small issue in the Makefile, I will point that out by the source. Thank you.
> 
> > iv never written a man page, so need to read up on this!
> 
> Actually, creating a man page using AI would be the way to go, it could read other man pages to understand how, and then the source code to learn what the options are...

unfortunately i caused enough damage with ai! 

Also i wish to learn this new skill.
I will put together a man page over the next week,

ghaerr:
> except for one small issue in the Makefile,

Oops - my bad. Everything looks great, thanks!

ghaerr:
> Also i wish to learn this new skill.

An easy way is to just look at a few man pages in elkscmd/man/man1/. You can use `eman` from the host to display them, which makes it easy to see what the dot-commands to in the man page files. The man pages aren't included on the floppy distributions since they take up too much space, so you have to run `man` from an HD distribution to see them in ELKS, unless you use `eman`.

PR #2805, opened 2026-08-08
[cmds] Implementation of sudo

djohanse:
Implementation of sudo command

user1 in wheel group so has sudo , user 2 not in wheel group, sudo denied

Screenshot 2026-08-07 at 10 05 28 PM


elkscmd/Make.install
```diff
@@ -228,6 +228,7 @@ install:
 		fi \
 	done < Filelist
 	rm Filelist
+	chmod 4755 $(DESTDIR)/bin/sudo 2>/dev/null || true
```
ghaerr, line 231: Hello @djohanse,

Command-specific installation doesn't belong here, please move to associated Makefile install: target. Eliminate 2>/dev/null  and true, we want to see output and stop the build if error.

djohanse:
So here is openCodes proposed rework and explanation.  Not sure its better:

Screenshot 2026-08-08 at 2 23 18 PM

Re: moving the sudo chmod out of Make.install
The command-specific logic is gone from Make.install — the chmod 4755 now lives in sys_utils/Makefile (its install-sudo target), with the 2>/dev/null || true removed so any chmod failure aborts the build. The single remaining line in Make.install is only the wiring that triggers it:
$(MAKE) -C sys_utils install-sudo
That line has to be there, and can't be removed or routed through the existing subdir mechanism, for these reasons:
1. The setuid bit can't be set earlier in the chain. The generic install used by the Applications filelist loop (Make.install:226) strips the setuid/setgid bits (verified: source 4755 → installed 0755). So the chmod must be applied to the copy in $(DESTDIR)/bin/sudo after the filelist loop — which is exactly where the hook sits.
2. There is no other app-agnostic trigger point. The image build (Make.image:49) calls Make.install as a single black box; the whole $(DESTDIR) tree is then copied into the filesystem image verbatim (genfs/mcopy), which is how the setuid bit set at install time reaches the image. Nothing between Make.install and image generation knows about per-app post-install steps.
3. The existing per-subdir mechanism doesn't fit. Make.install has an INSTDIRS loop (:271) for this purpose, but it only ever calls $(MAKE) -C $$DIR install. sys_utils' install target installs all of its $(PRGS) (init, min_init, who, ...), which would over-install those binaries into every image regardless of the Applications tag filter — including ones intentionally disabled there. The install-sudo target exists precisely so the image flow runs only the chmod.
So: the app-specific logic is confined to the app's Makefile as requested; Make.install retains only a one-line delegation, with failures no longer suppressed.

ghaerr:
> Not sure its better:

I would agree, it's even more special case handling and complexity. I hadn't thought of the problem of the host `install` losing the setuid bits, but I'm not surprised.

Since we don't really have a method to handle this, lets just leave the chmod line where it is, but don't ignore errors. That is, change the line to: `chmod 4755 $(DESTDIR)/bin/sudo`. This should always work, and we want to be notified if it doesn't.

I'm OK for now, adding new commands into 360k/360c floppy, but in general there's very little space left and when full, these new commands will likely be moved to larger floppy images, FYI.

With all that's being added to both the kernel and cmds, we are definitely going to permanently run out of floppy and kernel code/data space for anything new at some point, although I'm not sure exactly when that'll happen. Application commands are not nearly as big an issue as kernel modifications, but it all adds up and must be managed carefully. 

For example,I have been looking at adding $(TINYPRINTF) to the link lines of ping, ifconfig, traceroute and tcpdump to reduce their disk image size. I think they will continue to work with no other modifications, but have not tested yet. You might want to read the comments at the top of elkscmd/lib/tiny_printf.c to see which printf format designators it implements and which each of the programs use (I have), and submit a PR adding TINYPRINTF to elkscmd/inet/Makefile. Currently, ftp and ftpd can't use it because of fgets and popen. This is a good example of re-looking at submissions to ensure things are absolutely as small as possible to delay the inevitable!

All this said, your recently added `ping`, `curl`, ftp/ftpd (and other) enhancements are wonderful, thank you. I've also found myself using `dmesg` to be quite convenient to check what options the kernel booted with too :)

ghaerr:
@djohanse,

I realized with my last comment that since the chmod is in the main elkscmd/ application installation script, there may be architectures where /bin/sudo isn't in fact installed. Thus, we will have to leave the "|| true" in. Because of that, I am going to commit this now. Sorry for my confusion!

PR #2808, opened 2026-08-09
Add tinyprintf linker flag to new network utilities to reduce code size

djohanse:
Add $TINYPRINT linker flag to new network utilities per discussion in #2805

ftp was interesting.  With this change to add the $TINYPRINTF linker flag, initially I was not seeing the password prompt in testing until I logged out of the ftp session.  The ftp.c change in this PR is the AI-deduced "fix".

AI-generated explanation:
```
Why  ftp fix works:
- printf("Password: ") + fflush(stdout) writes the prompt immediately (tiny's fflush flushes bufout), instead of it being stuck in buferr until exit.
- getpass("") still does the termios echo-off + /dev/tty read, just prints nothing.
```

AI generated reasoning as to why the warning in tiny_vfprint is too restrictive:

```
Why is tiny_print comment too restrictive?

The comment at tiny_vfprintf.c:11 says "Don't use with fopen (stdout, stderr only)" — but that's overly conservative, and the ftp fix earlier proved it. Here's why:
The tiny functions are actually FILE-agnostic. Look at what they touch:
- fflush(fp) → just write(fp->fd, fp->bufstart, fp->bufpos - fp->bufstart) (tiny_vfprintf.c:73-75)
- fputc(ch, fp) → only fp->bufpos, fp->bufend, calls fflush (line 83-91)
- vfprintf(op, ...) → only passes op down to fputc
All of these fields (fd, bufpos, bufstart, bufend) are standard libc FILE members that fopen() initializes identically to the static stdout/stderr here. So FILE *fp = fopen("x","w"); fprintf(fp,...) works fine — as demonstrated when getpass()'s fopen("/dev/tty") worked under tiny.
What the warning is really about (the legitimate core): mixing tiny's write path with libc's buffered read path on the same stream. libc stdio does "flush/switch buffer on read↔write mode change" logic; tiny's fflush ignores __MODE_READING/__MODE_WRITING and won't discard libc's read-ahead when you switch from fgetc to fprintf on one FILE. That's a real hazard — but only for mixed read+write streams.
So the accurate restriction is:
Don't mix tiny's output with libc's input buffering on the same stream; write-only fopen'd files are fine.
The stated blanket ban is too strong because it forbids the perfectly safe write-only case (exactly the getpass/ftp scenario). The correct fix for the comment would be to narrow it to "no read/write mode-switching on the same stream" (and note tiny forces line-buffering on every FILE, so bulk-write perf on fopen'd files is worse than full-buffer libc).
```

ftp/ftpd with fix:

Screenshot 2026-08-09 at 3 03 14 PM

Other tests:

ifconfig, ping, traceroute

Screenshot 2026-08-09 at 2 29 40 PM

nslookup, tcpdump
Screenshot 2026-08-09 at 2 38 01 PM

curl
Screenshot 2026-08-09 at 2 41 48 PM

tiny irc 
Screenshot 2026-08-09 at 2 49 46 PM

tiny irc 2

Screenshot 2026-08-09 at 2 53 09 PM

### DIFF
--- a/elkscmd/inet/Makefile
+++ b/elkscmd/inet/Makefile
@@ -40,16 +40,16 @@ netstat: netstat.o $(TINYPRINTF)
 nslookup: nslookup.o
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
-ping: ping.o
+ping: ping.o $(TINYPRINTF)
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
-ifconfig: ifconfig.o
+ifconfig: ifconfig.o $(TINYPRINTF)
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
-tcpdump: tcpdump.o
+tcpdump: tcpdump.o $(TINYPRINTF)
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
-traceroute: traceroute.o
+traceroute: traceroute.o $(TINYPRINTF)
 	$(LD) $(LDFLAGS) -o $@ $^ $(LDLIBS)
 
 telnet: telnet.o $(TINYPRINTF)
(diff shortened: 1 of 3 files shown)


ghaerr:
Thanks @djohanse, I just compared binary file sizes and this saves quite a few thousand bytes on the distribution images!

And good to know that my comment in tiny_printf.c is too conservative - this opens up the possibilities of using tiny printf in other programs as well. I've made a note of it. It is a bit tricky since one has to ensure any arbitrary program source doesn't mix read/write buffering on the same stream, as well as of course only using the format strings present in tiny printf.

PR #2812, opened 2026-08-11
[cmds] Add GEM Desktop as an external application

parabyte:


ghaerr:
Would there be a way to rename /GEMAPPS/GEMSYS  to lib/gemsys or something like that, so that GEM doesn't have to live in an all-capital root directory entry, but instead its assets live in /lib/gemsys or something similar and hidden? All ELKS applications are suggested to use /lib to store all their stuff, under a single directory - in this case, gemsys. It would be nice if it weren't capitalized, but we can live with it capitalized if hidden in /lib.

The comment lines 48-49 should be removed in ExtApplications, comments are not necessary or expected.

I'll test the gem build using your buildext.sh and see how things work in the meantime.

> Install: select the :gem tag (e.g. CONFIG_APP_TAGS=:gem).

FYI, a tags line isn't required, all ExtApplication files that exist will be written during an HD install, and if they are not present, "SKIPPED" will be displayed, but no error like what occurs in Applications file.

ghaerr:
@parabyte,

Wow! This is very cool indeed! I tested the build and it compiles up and runs!!

GEM on ELKS (run 'gem'):
Screenshot 2026-08-11 at 2 38 52 PM

However, I can't get it to do anything... Does the mouse work? [See here](https://github.com/ghaerr/microwindows/blob/master/src/drivers/mou_ser.c) for the mouse driver for Nano-X, check out particular the open code, it deals with MOUSE_PORT= environment variable, as well as MOUSE_TYPE= variables which support ms, pc and ps/2 mice. It would be nice if the mouse setup were the same as Nano-X. Feel free to take from the Nano-X mouse driver if you want. I default the MOUSE_PORT=/dev/ttyS1 in /bootopts since that allows me to use /dev/ttyS0 serial port to be redirected to stdout on the terminal for debug output through \__dprintf.

I also couldn't get the keyboard to do anything. Is there special key/alt- handling to get to the menu? I tried F1 and Alt and the View dropdown worked once, but I couldn't get the menu tabs to move left/right. 

It would also be nice top hardcode an "exit" key into the application: I use ^A for Nano-X for the time being. This allows people to get out of the application with a single keystroke for testing. I couldn't get out of `gem` at all, and had to quit QEMU.

The graphics are impressive, nice job! :) I'm going to commit this now so that others can play with GEM on the standard ELKS HD distribution images.

Thank you!

parabyte:
there was a few bugs! no doubt 100 more, input IO seems to be fully functional and graphics drivers are in a happier state!

decided to drop ps2 pointer device support, cause i figured if users see ps2 mouse driver here then they expect ps2 support from elks!!

at some point in the future, i will try and squeeze more of gem into near.

ghaerr:
> decided to drop ps2 pointer device support, cause i figured if users see ps2 mouse driver here then they expect ps2 support from elks!!

Not sure why you'd do that, ELKS supports ps/2 mice.

> input IO seems to be fully functional 

As previously mentioned, I can't get any keyboard or mouse I/O to work at all. GEM displays a desktop screen and I can't get it to do anything at all.

parabyte:
okay removal of ps/2 pointer support was a dumb mistake now you pointed that out, 
I still have the driver in my local source tree.

I have only tested this on a real xt machine. 

to give the poor thing a rest, i might fire up qemu this evening see what is happening there with the non responsive desktop.

Alot of users i am guessing will want to play in a emulator as xt's are getting hard to come by nowadays.

PR #2813, opened 2026-08-11
Elks parallel elkscmd

djohanse:
OpenCode implementation of building elkscmd in parallel


ghaerr:
Hello @djohanse,

I would rather not accept this change, for the following reasons:
- Almost all subdirectories in elkscmd/  are already built in parallel
- The changes are much harder to read, understand and maintain using the tricky %- make rules
- This hard-to-understand change probably results in 1 sec difference in compile times
- The c86 and owc targets are rarely used and primarily added for experimental testing of a couple elkscmd/ source files by separate compilers (and produce .c86 and .os2 extensions).

djohanse:
Hello @ghaerr.

if I could persuade you that there may be more benefit than you're thinking:

```
Master [branch]  ia16 target is a strictly serial for-loop over ~30 subdirs (for DIR in $(IA16DIRS); do $(MAKE) -C $$DIR all ...). No two subdirs ever build at the same time.
Within each subdir, compiles are already parallel — e.g. busyelks lists cmd/*.o as independent targets, so make -j (inherited via the jobserver) parallelizes its file compiles.
So the reviewer's reason is only partially accurate: parallelism already exists inside each subdirectory, but not across subdirectories
```

and we can reduce the complexity by only working with ia16 and ignoring the experimental builds:



Screenshot 2026-08-10 at 10 10 23 PM

would you be interested? Or should I just move on?

Thanks!

ghaerr:
> if I could persuade you that there may be more benefit than you're thinking:

Actually that would be your AI tool trying to persuading us. There *may* be some benefit(s). However:

> would you be interested? 

May I be frank?

What's happening, with rapidly increasing frequency over the last 4-5 months, is that AI "analyses" have been submitted  as Issues, Discussions, and PRs. This then takes me, a single human with one mind and limited time, lots and lots of time to read through usually very interesting reams, reams, and more reams of AI-generated code, essays, very long descriptive comments, audits, summaries, proposed action plans, tables, and more, usually generated by a human who typed a few phrases into an AI tool over the course of an hour or two. It's me versus a data center. Almost all the input I find very interesting, some extremely clever, some entirely incorrect, many ifdef-d to the hilt and extremely short-sighted, prone to deal only with a specific prompt I can't see, and thus requiring tons of my time to evaluate. In many cases, the submitter does not fully understand the submission. In some cases, the submitter doesn't say AI wrote it, and still doesn't understand it.

What this project needs more of is humanity, and less AI. IMHO, the project itself isn't made better by AI. We're not trying to (re)build a system using tools to write code the submitter doesn't understand, to make a system better in ways not understood. There's no rush, we don't have to ship anything, but we all want want to learn, I think. The ELKS project is about going back to simpler days, to days where a human can in fact fully understand a system, and through full understanding, contribute to (slowly) (re)build a vintage environment using what we humans have learned over decades into the present day. It's fun. But using AI alone to "develop" ELKS is not the answer. We must couple our own knowledge with AI, if we must use AI at all. I would go so far as to say that using too much AI steals from ourselves, in this project.

ELKS is a small system and should remain small, and remain closely understood by people (not machines). it should be simply written so that others can easily understand it. Makefiles, C code, ifdefs, the simpler the better.


Please don't take this the wrong way - some submissions of AI-written applications programs have been extremely useful, especially the recent networking applications - but they're coupled with human knowledge, rather than just AI analysis. I don't (necessarily) mind that a tool was used to (completely?) build an application, since they're easily written by AI and can be easily removed, but so far I have not been impressed with AI kernel code, with the big exception of reviewing (but not committing) bug fixes. The AI bug fixes usually don't get it fully right but can sometimes capture the essence of a good solution, although in some ways I feel we might be cheating since neither myself nor the submitter fully developed the answer. But that's OK, I guess, as long as it doesn't go to excess. I fear we may be nearing the limit, but I'm not ready to make a definitive statement about AI in ELKS yet. 

\</diatribe>


> Or should I just move on?

I see from the last post that yes, there is some room for improvement by looking at how all the directories might be compiled simultaneously using an AI-generated idea. But how important is this? How do we know it's correct? Compiling an elkscmd/ subdirectory that uses an analogous file to the shared tiny_printf.o could possibly be attempted linked before the .o is generated? I don't know, I haven't time to check all the elkscmd/ directories, who knows whether AI considered it, there may be other dependencies, Makefiles aren't perfect, etc. It's just not that important, the risk vs reward is minimal, and it's yet more stuff to do. Right now, I'm shoulders deep in producing ELKS networking proven bug fixes from just two AI PRs produced two weeks ago that produced 25+ issues and fixes for consideration. These are more important than possibly breaking the build with an AI-generated solution that might not be fully tested, etc, etc, etc. Finally, it is unknown how much time this proposed solution saves each build.

I have made a note to consider this PR in more detail after I've had time to review it more fully. But that's pretty far down the list right now.

ghaerr:
@djohanse I am having trouble understanding how the 2nd proposed change runs more parallel builds than the original, but then there are many things about parallel make I may not get yet.

In the original code, a for loop is used to run make -C:
```
for DIR in $(IA16DIRS); do $(MAKE) -C $$DIR all || exit 1; done
```

In the proposed enhancement, the targets are listed on the left of a rule, which then runs make -C for each of them:
```
$(IA16DIRS): 
    $(MAKE) -C $@ all
```

How is it that the latter make -C executions run quicker than the former make -C using a for loop? That is, doesn't 
`make -C` wait for completion of the subdirectory parallel build in both cases before executing the next make -C?

ghaerr:
OK I understand now: in the former case, a parallel make -C is started in each subdirectory, sequentially. In the latter, the top level make will start parallel make -C simultaneously in each subdirectory.

Let's assume  -j16 and there's an average of 25 tasks (make targets) in each subdirectory. In the former case, the single make -C quickly becomes task-bound, but as those tasks complete, the system will start waiting when the number of jobs left to complete per directory is less than the CPUs available. In the latter case, this also occurs, but there may (or may not) be other subdirectory make -C's waiting on another parallel make token. So for elkscmd/, it really depends on how many average tasks there are per subdirectory make in order to gain improvement. Most of elkscmd/ have a large number of tasks per subdirectory, so I would not expect much improvement.

On my system, the existing elkscmd/ make takes 8.75 to 9.5 seconds. So even with a 50% improvement (unlikely I would say), we'd save 5 seconds of build time.

I'll measure this just for the heck of it.

ghaerr:
> I'll measure this just for the heck of it.

Allright @djohanse, the results are in! I tested (again) both as-is and with your proposed enhancement using `time make` after a `make clean`:

Current: 10.15 secs
Proposed 11.26 secs

The AI solution takes longer than the existing Makefile. Glad this was analyzed and tested by a human :) 

Not exactly sure why proposed Makefile takes longer, but guessing that the newly proposed mechanism started a bunch of make -C's quickly, but then got bogged down running out of parallel tokens because of the many open make's, while the current version starts just one make -C initially, then lots more jobs are able to be completed within the subdirectory, allowing the next make -C to start more quickly. Or maybe it's just a different ordering of subdirectories, I don't know.

I am very glad I took the time to refute the AI's proposal, as once again, it was merely an untested idea that didn't work.

djohanse:
You are correct.  If I had to submit this in the corporate world I would
have had to show benchmarks and checksums.  Not fair to put that burden on
you.

Hope your doing well.

